Issue #1432, opened 2026-01-07
[FEATURE] Add cache_strategy="auto" for automatic prompt caching


kevmyung:
### Problem Statement

1. **No SDK-level support for conversation caching**
   - Prompt caching can significantly reduce costs (up to 90% on cache reads) and latency
   - Users must implement custom hooks to enable caching across tool-use loops

2. **Limited scope of existing cache options**
   - Current `cache_prompt`, `cache_tools` only support system prompt and tool definitions
   - Minimal impact in real-world workflows where conversation history dominates token usage

3. **Provider-specific configuration**
   - Cache configuration only available in BedrockModel
   - Other providers have cachePoint conversion logic but no way to enable via config
   - Inconsistent developer experience

<img width="1850" height="1422" alt="Image" src="https://github.com/user-attachments/assets/6b1ff6b6-2407-4ae5-bfe9-4c3e54b03bd7" />

### Proposed Solution

Add `cache_strategy` parameter to base Model class with hook-based auto-caching.

### Key Components

<img width="2905" height="928" alt="Image" src="https://github.com/user-attachments/assets/71f00119-f0df-4711-a1a4-4da8e84d3730" />

* Model (ABC) : Shared Configuration
    * Add cache_strategy: Optional[str] = None to the base Model class. All provider implementations (BedrockModel, AnthropicModel, LiteLLM, etc.) inherit this configuration automatically.
```
# Usage (any provider)
agent = Agent(model=BedrockModel(cache_strategy="auto"))
agent = Agent(model=AnthropicModel(cache_strategy="auto"))
agent = Agent(model=LiteLLM(cache_strategy="auto"))
```

* Agent: Hook Auto-Registration
    * When cache_strategy="auto" is detected, the Agent automatically registers a ConversationCachingHook:
```
# In Agent.__init__()
if model.get_config().get("cache_strategy") == "auto":
    self.hooks.add_hook(ConversationCachingHook())
```

* ConversationCachingHook: CachePoint Injection
    * The hook injects a cachePoint block at the last assistant message on each BeforeModelCallEvent
```
# Before injection
[..., {role: "assistant", content: [...]}, {role: "user", ...}]

# After injection  
[..., {role: "assistant", content: [..., {"cachePoint": {"type": "default"}}]}, {role: "user", ...}]
```

    * This single cache point covers system prompt + tools + conversation history up to that point.
* Provider-Specific Handling (Existing Logic)
[...]

Issue #1140, opened 2025-11-05
[FEATURE] Caching support for all models


dbschmigelski:
### Problem Statement

This is a tracking ticket similar to [Prompt caching support for LiteLLM](https://github.com/strands-agents/sdk-python/issues/937) but for all models.

As a user, I would like to leverage the SystemContent mechanism to apply prompt caching in all models.

### Use Case

To improve latency and reduce cost, Strands should support Prompt caching. As an abstraction layer, this should be done in a provider-agnostic way by using the SystemContentBlock.

chosw1029:
Hi — we're hitting this on the direct Anthropic API path and have a working patch that adds parity with `BedrockModel`:

- `AnthropicModel.stream()` / `format_request()` now accept `system_prompt_content` and emit the `system` field in Anthropic list-form with `cache_control: {"type": "ephemeral"}` on the text block immediately preceding a `cachePoint`.
- `format_chunk()` metadata events expose `cacheReadInputTokens` / `cacheWriteInputTokens` (matching `BedrockModel`'s field names so existing observability code works unchanged).

Verified against a live Claude Sonnet 4.5 run: subsequent requests with an identical system prefix now report `cache_read_input_tokens > 0` where the pre-patch code reported `0`.

Opening a PR shortly. Happy to adjust scope if you'd like anything trimmed or staged differently.

joshwand:
Is this ticket limited to system prompt caching, or does it include arbitrary cache points for multi-turn conversations?

maxmilian:
@joshwand — your question from 2026-05-19 has been sitting unanswered, so here is what the code says today (`main` @ `24bc5c3d5`). Short version: **arbitrary in-conversation cache points are already supported, not just the system prompt.**

- `_format_request_messages` (`strands-py/src/strands/models/anthropic.py:222`) walks every content block of every message and, at `:247`, treats a `cachePoint` block as a marker: it pops the block and attaches `cache_control` to the block *preceding* it via `_attach_cache_control` (`:287`), warning and skipping when nothing ahead of it can carry one.
- `_manage_cache_points` (`:360`, called at `:445`) picks the target message index and supplies the configured TTL when a placed cache point does not carry its own.
- The system prompt is a **separate** path — `_format_system_prompt` (`:474`) auto-injects a cache point at the end of the system content — so the two are independent rather than the system prompt being the only supported position.

[...]

PR #2160, opened 2026-04-19
feat(anthropic): add prompt caching support to direct Anthropic API

chosw1029:
## Description

`AnthropicModel` currently provides no way to take advantage of Anthropic's prompt caching feature, which `BedrockModel` already supports. Three gaps:

1. **`stream()` drops `system_prompt_content`.** The event loop (`event_loop/streaming.py`) passes both `system_prompt` and `system_prompt_content` to the model. `BedrockModel` uses the latter, but `AnthropicModel.stream()` only declares `system_prompt: str | None` and silently swallows `system_prompt_content` via `**kwargs`.

2. **`format_request()` sends system as a plain string.** The resulting request body is `{"system": "<string>", ...}`, so there is no place to attach `cache_control`.

3. **Cache token counts are dropped.** In `format_chunk()` the `metadata` case only extracts `input_tokens` / `output_tokens`. Anthropic already returns `cache_creation_input_tokens` and `cache_read_input_tokens` in the usage object, but they never reach downstream consumers.

This PR closes the three gaps:

- `stream()` and `format_request()` now accept `system_prompt_content: list[SystemContentBlock] | None`.
- When `system_prompt_content` is supplied, the `system` field is emitted in Anthropic list-form. A `cachePoint` block attaches `cache_control: {"type": "ephemeral"}` to the preceding text block, mirroring the convention already used by `_format_request_messages`.
- `format_chunk()` emits `cacheReadInputTokens` / `cacheWriteInputTokens` in the metadata usage dict. Names match `BedrockModel`'s, so existing observability code (spans that read `cacheReadInputTokens`) works without changes. The existing `Usage` TypedDict already declares both fields as optional.

No beta headers are required — prompt caching is GA on the direct Anthropic API for Claude Sonnet 3.7+, Sonnet 4.x, Opus 4.x, and Haiku 4.x. The minimum cacheable prefix is enforced by the API, not the SDK.

### Backwards compatibility

- `stream()` / `format_request()` add a keyword argument with a default of `None`; existing callers are unaffected.
- When `system_prompt_content` is absent, behavior is byte-for-byte identical (system is still sent as a plain string).
- `format_chunk()` only adds cache keys to the metadata usage dict when the upstream response reports non-zero cache counts; otherwise the shape is unchanged.
- No change to `AnthropicConfig` TypedDict — caching is driven by `system_prompt_content`, which is already produced by `Agent` when the caller passes a `list[SystemContentBlock]` system prompt.

## Related Issues

Relates to #1140 (Prompt caching support for all models — currently ""ready for contribution"") and #1432 (cache_strategy=""auto"" across providers; the AnthropicModel path was listed but never implemented).

## Documentation PR

Will follow up with a docs PR on `strands-agents/agents-docs` once the approach is confirmed. Happy to include it in this PR if preferred.

## Type of Change

New feature

## Testing

Unit tests added in `tests/strands/models/test_anthropic.py`:

- `system_prompt_content` with a `cachePoint` block → request `system` is list-form and the preceding text block carries `cache_control: {""type"": ""ephemeral""}`.
- `system_prompt_content` with text blocks only (no `cachePoint`) → list-form system, no `cache_control`.
- `system_prompt_content` takes precedence over `system_prompt` when both are supplied.
- Anthropic response metadata with `cache_read_input_tokens` / `cache_creation_input_tokens` → `format_chunk` exposes `cacheReadInputTokens` / `cacheWriteInputTokens`.
- Anthropic response with zero cache counts → metadata usage dict unchanged.

Manually verified with a live Claude Sonnet 4.5 call: a second request with an identical system prefix reports `cache_read_input_tokens > 0` where the pre-patch code reported `0`.

Verify that the changes do not break functionality or introduce warnings in consuming repositories: agents-docs, agents-tools, agents-cli.

## Checklist

----

### DIFF
--- a/src/strands/models/anthropic.py
+++ b/src/strands/models/anthropic.py
@@ -16,7 +16,8 @@ from typing_extensions import Required, Unpack, override
 
 from ..event_loop.streaming import process_stream
 from ..tools.structured_output.structured_output_utils import convert_pydantic_to_tool_spec
-from ..types.content import ContentBlock, Messages
+from ..types.content import ContentBlock, Messages, SystemContentBlock
+from ..types.event_loop import Usage
 from ..types.exceptions import ContextWindowOverflowException, ModelThrottledException
 from ..types.streaming import StreamEvent
 from ..types.tools import ToolChoice, ToolChoiceToolDict, ToolSpec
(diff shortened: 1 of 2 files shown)


opieter-aws:
Closing in favor of #3681

Issue #1294, opened 2025-12-05
[CONTEXT] [FEATURE] Token Estimation API


westonbrown:
## Problem Statement

The SDK only reports token usage after model calls complete via `accumulated_usage` in `EventLoopMetrics`. There is no way to estimate token counts before sending messages to the model.

This forces a reactive pattern where applications must wait for `ContextWindowOverflowException` to be raised, then handle reduction. For long-running agents with many tool calls, this causes unnecessary API failures, output token starvation, and degraded user experience.

This is a prerequisite for implementing proactive context compression (see #555) - calculating a percentage threshold requires knowing current token usage.

## Proposed Solution

Add an `estimate_tokens()` method to the `Model` interface:

```python
class Model(ABC):
    @abstractmethod
    def estimate_tokens(
        self, 
        messages: Messages, 
        tool_specs: Optional[list[ToolSpec]] = None,
        system_prompt: Optional[str] = None
    ) -> int:
        """Estimate token count for the given input before sending to model."""
        pass
```

Model providers can implement using native APIs:
- Anthropic: `anthropic.count_tokens()`
- OpenAI: `tiktoken` library
- Gemini: `model.count_tokens()`
- LiteLLM: `litellm.token_counter()`

## Use Case

1. Proactive context management - check usage before model call, trigger compression if over threshold
2. Budget tracking - display real-time context consumption in UIs
3. Cost estimation - calculate expected costs before operations
4. Intelligent pruning - identify which messages consume most tokens

```python
current_tokens = agent.model.estimate_tokens(agent.messages)
if current_tokens > threshold:
    agent.conversation_manager.reduce_context(agent)
```

## Alternative Solutions

Applications can implement estimation using third-party tokenizers, but this requires maintaining model-specific mappings and does not integrate with the SDK's context management.

## Additional Context

This feature enables implementation of #555 (Proactive Context Compression). Related: #460 (token counting inconsistency).

PR #2162, opened 2026-04-20
feat: large tool result offload

### LINKED ISSUE #1296: [FEATURE] Large Tool Result Externalization via AfterToolCallEvent Hook
## Problem Statement

When tool outputs exceed the model's context capacity, the SDK's current handling in `SlidingWindowConversationManager` replaces the entire result with a generic message:

```python
tool_result_too_large_message = "The tool result was too large!"
```

This approach has two problems:
1. **Data loss** - Full output is discarded permanently
2. **Reactive timing** - Only triggers after context overflow, wasting a failed API call

For tools that return large outputs (file readers, API responses, database queries, log analyzers), the agent loses the ability to reference or reason about the content.

## Proposed Solution

Add a hook-based mechanism to intercept large tool results via `AfterToolCallEvent` before they enter the conversation:

1. Detect outputs exceeding a configurable threshold (e.g., 10KB)
2. Persist full output to an artifact file on disk
3. Replace the conversation content with a truncated preview plus artifact reference

Example output injected into conversation:
```
[Output: 125,432 chars | Full: artifacts/shell_20241124_143022.log]

<first 4,000 characters as preview>

[Complete output saved to: artifacts/shell_20241124_143022.log]
```

**Context window benefits:**
- 125KB output reduced to ~5KB in conversation (preview + reference)
- Proactive reduction at tool execution time, not after overflow
- Prevents large outputs from ever consuming context space

## Use Case

1. **Context efficiency** - Large outputs handled proactively, keeping context window clean for reasoning
2. **Long-running agents** - Process large datasets, logs, or API responses without context pressure
3. **Debugging and auditing** - Full outputs preserved on disk for review
4. **Multi-step workflows** - Intermediate results retrievable via artifact path if needed later

```python
agent = Agent(
    hooks=[LargeResultExternalizerHook(
        threshold=10000,      # Externalize outputs > 10KB
        artifact_dir="./artifacts",
        preview_size=4000     # Keep 4KB preview in context
    )]
)
```

## Alternative Solutions

- Aggressive truncation (current approach) - loses data permanently, reactive only
- In-memory compression (#556) - reduces size but still consumes context
- Manual tool wrappers - requires per-tool implementation

Artifact externalization complements #555 (proactive compression) and #556 (pruning) by reducing context consumption at the source while preserving full data.

## Additional Context

This operates at a different layer than conversation management:
- Triggers on `AfterToolCallEvent` (before conversation manager sees it)
- Proactively reduces context consumption at tool execution time
- Preserves full output to disk for retrieval
- Provides immediate preview for LLM reasoning

Related: #555 (proactive compression), #556 (pruning), #1294 (token estimation), #1295 (context limit)

lizradway:
## Description

Adds a `ContextOffloader` plugin that proactively intercepts oversized tool results via `AfterToolCallEvent`, persists each content block individually to a pluggable storage backend, and replaces the in-context result with a truncated preview plus per-block references. Includes an optional built-in retrieval tool (disabled by default) so the agent can fetch offloaded content on demand, returning each content type in its native format.

**Problem:** When tool outputs exceed the model's context capacity, `SlidingWindowConversationManager` reactively truncates to first/last 200 chars — losing data permanently and wasting a failed API call.

**Solution:** A Plugin (following the `AgentSkills` pattern) that operates at tool execution time, before the result enters the conversation. Each content block is stored individually with its content type preserved, enabling type-aware retrieval. Inline guidance in each offloaded result tells the agent to use the preview when possible and to use its available tools to selectively access the data it needs.

**Token-based thresholds:** Uses the agent's `model.count_tokens()` for accurate token estimation (tiktoken when available, chars/4 heuristic fallback). The async hook wraps the tool result as a message for counting. Preview slicing uses tiktoken for exact token-level cuts when available, falls back to `tokens * 4` chars.

**Content type handling:**
| Type | Behavior |
|------|----------|
| Text | Stored as `text/plain`, replaced with a preview |
| JSON | Stored as `application/json` (serialized via `json.dumps`), replaced with a preview |
| Image | Stored in native format (e.g., `image/png`), replaced with placeholder + reference |
| Document | Stored in native format (e.g., `application/pdf`), replaced with placeholder + reference |
| Unknown | Passed through unchanged |

**Storage backends (required — user must choose one):**
- `InMemoryStorage` — no filesystem side effects, content cleared on process exit. `clear()` method for manual cleanup.
- `FileStorage` — persists to disk with `.metadata.json` sidecar for content type tracking across process restarts
- `S3Storage` — persists to Amazon S3 (follows `S3SessionManager` patterns), content type preserved via S3 object metadata

**Built-in retrieval tool (opt-in, disabled by default):**
- `retrieve_offloaded_content` — agent can fetch offloaded content by reference
- Enabled via `include_retrieval_tool=True`
- Returns content in its native type: text as string, JSON as `{"json": ...}` block, images as `{"image": ...}` block, documents as `{"document": ...}` block
- Retrieval results are excluded from re-offloading (prevents circular offload loops)
- Disabled by default because once storage defaults to VFS/Sandbox, agents will use shell/grep/SQL tools to navigate offloaded results directly
- Inline guidance in each offloaded result adapts: when tool is enabled, mentions `retrieve_offloaded_content`; always tells the agent to use available tools

**Usage:**
```python
from strands import Agent
from strands.vended_plugins.context_offloader import (
    ContextOffloader,
    InMemoryStorage,
)

# In-memory storage — context reduction only, no persistence
agent = Agent(plugins=[
    ContextOffloader(storage=InMemoryStorage())
])
```

```python
from strands.vended_plugins.context_offloader import (
    ContextOffloader,
    FileStorage,
)

# File storage — persists artifacts to disk, custom thresholds
agent = Agent(plugins=[
    ContextOffloader(
        storage=FileStorage("./artifacts"),
        max_result_tokens=5_000,
        preview_tokens=2_000,
    )
])
```

```python
from strands.vended_plugins.context_offloader import (
    ContextOffloader,
    S3Storage,
)

# S3 storage with retrieval tool enabled
agent = Agent(plugins=[
    ContextOffloader(
        storage=S3Storage(
            bucket="my-agent-artifacts",
            prefix="tool-results/",
        ),
        include_retrieval_tool=True,
    )
])
```

## Related Issues

Closes #1296

## Documentation PR

https://github.com/strands-agents/docs/pull/772

## Type of Change

New feature

## Testing

How have you tested the change? Verify that the changes do not break functionality or introduce warnings in consuming repositories: agents-docs, agents-tools, agents-cli
- 66 new unit tests covering all storage backends (including metadata sidecar persistence and corruption), all content types, retrieval tool opt-in/opt-out with native type returns, token-based thresholds with mocked count_tokens, inline guidance adaptation, edge cases (threshold boundaries, cancelled tools, storage failures, partial failures, path traversal, thread safety, input validation, circular offload prevention)
- Manually tested via demo script confirming: token-based offloading triggers correctly, agent uses preview for summaries, retrieves only when specific data is needed, retrieval returns native content types
- All existing tests pass (1710 passed, 10 pre-existing telemetry failures unrelated to this change)

## Checklist

----


src/strands/vended_plugins/result_externalizer/plugin.py
```diff
@@ -0,0 +1,197 @@
...
+        Args:
+            image_block: The image content block.
+
+        Returns:
+            A placeholder string describing the image.
+        """
+        source: Any = image_block.get("source", {})
+        media_type = image_block.get("format", "unknown")
+        data = source.get("bytes", b"")
+        return f"[image: {media_type}, {len(data) if data else 0} bytes]"
```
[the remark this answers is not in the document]
lizradway: Noted and to be followed up by exploration in https://github.com/strands-agents/sdk-python/issues/2171

src/strands/vended_plugins/result_externalizer/storage.py
```diff
@@ -0,0 +1,308 @@
...
+    Returns:
+        A sanitized string safe for use in filenames.
+    """
+    sanitized = tool_use_id.replace("..", "_").replace("/", "_").replace("\\", "_")
+    sanitized = re.sub(r"[^\w\-.]", "_", sanitized)
+    return sanitized
+
+
+@runtime_checkable
+class ExternalizationStorage(Protocol):
```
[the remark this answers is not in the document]
lizradway: See https://github.com/strands-agents/sdk-python/issues/2168

Can add documentation. also added `clear()` method that users can use to manually clear memory.

src/strands/vended_plugins/result_externalizer/plugin.py
```diff
@@ -0,0 +1,197 @@
...
+        if event.cancel_message is not None:
+            return
+
+        result = event.result
+        content = result["content"]
+
+        # Collect externalizable content (text + JSON) and measure total size
+        text_parts: list[str] = []
+        for block in content:
+            if block.get("text"):
```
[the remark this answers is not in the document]
lizradway: empty text blocks add zero content value to externalized output. Added a comment documenting this behavior.

src/strands/vended_plugins/result_externalizer/storage.py
```diff
@@ -0,0 +1,308 @@
...
+        key = f"{self._prefix}{timestamp_ms}_{counter}_{sanitized_id}.txt"
+
+        self._client.put_object(
+            Bucket=self._bucket,
+            Key=key,
+            Body=content.encode("utf-8"),
+            ContentType="text/plain; charset=utf-8",
+        )
+
+        return key
```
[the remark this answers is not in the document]
lizradway: Added Raises: botocore.exceptions.ClientError to the store docstring. The plugin’s catch-all with exc_info=True already logs the full traceback including the S3 error details, so users can diagnose from logs

src/strands/vended_plugins/result_externalizer/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+        ToolResultExternalizer(
+            storage=FileExternalizationStorage("./artifacts"),
+            max_result_chars=20_000,
+            preview_chars=8_000,
+        )
+    ])
+    ```
+"""
+
+from .plugin import ToolResultExternalizer
```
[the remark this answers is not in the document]
lizradway: This is not a top level plugin

src/strands/vended_plugins/result_externalizer/__init__.py
```diff
@@ -0,0 +1,46 @@
+"""Tool result externalization plugin for Strands Agents.
```
opieter-aws: In the doc review we discussed renaming externalization, to e.g. serialization. Are we happy shipping this as externalization?

src/strands/vended_plugins/context_offloader/storage.py
```diff
@@ -0,0 +1,334 @@
...
+
+import re
+import threading
+import time
+from pathlib import Path
+from typing import Any, Protocol, runtime_checkable
+
+import boto3
+from botocore.config import Config as BotocoreConfig
+from botocore.exceptions import ClientError
```
opieter-aws, line 37: Should we load this lazily for `S3ExternalizationStorage` only?

lizradway: I thought about that, but `boto3` is already a core dependency in the SDK. This matches the import pattern used by S3SessionManager, BedrockModel, and SageMakerModel

src/strands/vended_plugins/result_externalizer/__init__.py
lizradway: I talked about this a tiny bit after the review meeting with chay (who suggested serialization). I think there are three main components to this that should be encapsulated in the name:
1. tool serialization
2. storage export
3. preview 

I believe chay + i agreed that serialization couldn't encapsulate this full lifecycle since it stops after step 1. 

I'm not sure there is a single word that can without overcomplicating the name, but I think externalization is closer. Open to more suggestions, here are some options I've considered: `ToolResultsOffload`, `ToolResultsPersistence`, `ToolResultsPreview`, etc. Happy to attach `serial` as a prefix, but I think that it may be introducing too much implementation into naming.

lizradway: @chaynabors thoughts?

opieter-aws (dismissed):

mkmeral (changes requested):
(to avoid merging by mistake)

src/strands/vended_plugins/result_externalizer/plugin.py
```diff
@@ -0,0 +1,218 @@
...
+from typing import Any
+
+from ...hooks.events import AfterToolCallEvent
+from ...plugins import Plugin, hook
+from ...types.tools import ToolResult, ToolResultContent
+from .storage import ExternalizationStorage
+
+logger = logging.getLogger(__name__)
+
+_DEFAULT_MAX_RESULT_CHARS = 10_000
```
mkmeral: given that we have tokenizer now, can we use tokens instead? does char a good metric for json and non-text tool results?

src/strands/vended_plugins/result_externalizer/plugin.py
```diff
@@ -0,0 +1,218 @@
...
+
+    Example:
+        ```python
+        from strands import Agent
+        from strands.vended_plugins.result_externalizer import (
+            ToolResultExternalizer,
+            InMemoryExternalizationStorage,
+        )
+
+        agent = Agent(plugins=[
```
mkmeral: is this example complete? I'd expect to see a tool to search/filter/read the "externalized" tool result

src/strands/vended_plugins/result_externalizer/plugin.py
```diff
@@ -0,0 +1,218 @@
...
+logger = logging.getLogger(__name__)
+
+_DEFAULT_MAX_RESULT_CHARS = 10_000
+"""Default character threshold above which tool results are externalized."""
+
+_DEFAULT_PREVIEW_CHARS = 4_000
+"""Default number of characters to keep as a preview in context."""
+
+
+class ToolResultExternalizer(Plugin):
```
mkmeral: Did we consider alternative names? I'm not sure "externalizer" is a descriptive name

src/strands/vended_plugins/result_externalizer/plugin.py
```diff
@@ -0,0 +1,218 @@
...
+        )
+
+        # Build new content:
+        #   - Preview text block first
+        #   - Images replaced with placeholders (following SlidingWindowConversationManager)
+        #   - Documents replaced with placeholders
+        #   - Text and JSON blocks are already captured in the externalized content
+        new_content: list[ToolResultContent] = [ToolResultContent(text=preview_text)]
+        for block in content:
+            if "image" in block:
```
mkmeral: for anything that's not handled here, the content type should be just passed through. I think if we were to add a new content type in tool results, it can fall through the cracks where we dont actually externalize it, nor add it to the new content list. we should have a default

src/strands/vended_plugins/result_externalizer/storage.py
```diff
@@ -0,0 +1,334 @@
...
+    Returns:
+        A sanitized string safe for use in filenames.
+    """
+    sanitized = tool_use_id.replace("..", "_").replace("/", "_").replace("\\", "_")
+    sanitized = re.sub(r"[^\w\-.]", "_", sanitized)
+    return sanitized
+
+
+@runtime_checkable
+class ExternalizationStorage(Protocol):
```
mkmeral: Reading the classes here, this seems like a VFS (virtual filesystem) to me, which is what we will provide as part of Sandbox. 

(i know it's not merged yet) but given sandboxes is in sdk, is this abstraction a duplication? Would I want to separate out my large tool result storage from my agents' sandbox? 

should this be just on sandbox? and use write and read from sandbox abstraction?

src/strands/vended_plugins/result_externalizer/storage.py
```diff
@@ -0,0 +1,334 @@
...
+        """
+        file_path = (self._artifact_dir / reference).resolve()
+        if not file_path.is_relative_to(self._artifact_dir.resolve()):
+            raise KeyError(f"Reference not found: {reference}")
+        if not file_path.is_file():
+            raise KeyError(f"Reference not found: {reference}")
+        return file_path.read_text(encoding="utf-8")
+
+
+class InMemoryExternalizationStorage:
```
mkmeral: should we provide a tool that's attached to this storage specifically? how would LLM actually use this? are we making customers re-implement the same tool?

lizradway: fast follow in https://github.com/strands-agents/sdk-python/issues/1678

can include in this pr if that is easier: adds tool + sys prompt to discover tool to plugin. until then, current strat is including relevant tool (ie. file, s3) and noting discoverability in sys prompt

src/strands/vended_plugins/result_externalizer/plugin.py
lizradway: that makes sense

src/strands/vended_plugins/result_externalizer/plugin.py
lizradway: fair point -- i think now that we have tokenizer (pending #1294), it is probably good practice for to use tokens as the standard threshold throughout the SDK rather than char going forward. lmk your thoughts? can add to some doc somewhere to note that if so.

src/strands/vended_plugins/result_externalizer/plugin.py
lizradway: fast follow in https://github.com/strands-agents/sdk-python/issues/1678

src/strands/vended_plugins/result_externalizer/plugin.py
lizradway: see above comment thread: https://github.com/strands-agents/sdk-python/pull/2162#discussion_r3118707890

open to suggestions if you have any -- havent found anything yet i feel satisfied with

src/strands/vended_plugins/result_externalizer/storage.py
lizradway: atleast from what i read briefly in sandboxes, from an initial plug-and-play perspective, the async sandbox tools would break the sync event loop hooks.


I think that once sandboxes drop, an sandbox adapter could write to a dedicated subdirectory and rely explicitly on the plugin's built-in `retrieve` tool, so as to not pollute the agent's workspace. but i dont think that is worth blocking initial implementation of externalization on.

src/strands/vended_plugins/result_externalizer/plugin.py
lizradway: also happy to address that as a follow up right after #1294 is released so this isn't blocked.

src/strands/vended_plugins/result_externalizer/storage.py
mkmeral: > from an initial plug-and-play perspective, the async sandbox tools would break the sync event loop hooks.

we support async hooks, it should be fine

> so as to not pollute the agent's workspace

yeah that's a good callout, though i might argue it's part of the "work" so it should be in "workspace" 😅 

anyways won't block on this. there's a clean path to sandbox, and users might want to diverge

src/strands/vended_plugins/result_externalizer/plugin.py
mkmeral: > it is probably good practice for to use tokens as the standard threshold throughout the SDK rather than char going forward

yep +1 we can add to DECISIONS.md in docs, feel free to do it

my biggest concern here is essentially making the character based heuristic the default behavior. I think token based should be. 

That said, I'd mark this as python v2 update for defaults, and we can launch as is

src/strands/vended_plugins/result_externalizer/__init__.py
mkmeral: Chay had a good point about also externalizing (stashing? it may be a better term tbh) user messages too, because people copy paste, or for automated agents they just dump to context/user-message.

Are we planning to do it? Will we extend this? Will there be a different plugin (if so why?)? Or maybe should we just get rid of the tool name and give ourselves a room to add it later on?  so sth like `LargeMessageStasher`, `ContentStasher` `ContextOffloader` (i actually like this one)

src/strands/vended_plugins/result_externalizer/plugin.py
lizradway: agreed

src/strands/vended_plugins/result_externalizer/__init__.py
lizradway: I thought about this and left the directory purposefully vague so that message externalization could be added if we desired.

However the conclusion i came to was that message-level context management falls under conversation manager. Tool results have unique pain points as can be unexpectedly large and the agent has already consumed the output when the hook fires, so the preview serves as a summary for future turns, unlike messages where externalization would  remove the content the user wants the model to see.

I actually quite like `ContextOffloader` as well and can update for that. I still think there could eventually be some use for message externalization, I just don't see one now.

src/strands/vended_plugins/result_externalizer/plugin.py
```diff
@@ -0,0 +1,223 @@
...
+        if not text_parts:
+            return
+
+        full_text = "\n".join(text_parts)
+        if len(full_text) <= self._max_result_chars:
+            return
+
+        # Persist full content
+        try:
+            reference = self._storage.store(event.tool_use["toolUseId"], full_text)
```
mkmeral: This does not include images or documents. With this plugin, we would be losing data if there was a lot of text + images in a tool result. the image would be lost. we need to fix it

mkmeral (changes requested):

src/strands/vended_plugins/context_offloader/plugin.py
```diff
@@ -0,0 +1,335 @@
...
+
+_DEFAULT_PREVIEW_CHARS = 4_000
+"""Default number of characters to keep as a preview in context."""
+
+_DEFAULT_RETRIEVE_LIMIT_BYTES = 10_000
+"""Default maximum bytes returned per retrieval call."""
+
+_STATE_KEY = "context_offloader"
+
+_SYSTEM_PROMPT_INJECTION = """
```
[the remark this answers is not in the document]
lizradway: updated to  `\n\n` prefix string concatenation

tests/strands/vended_plugins/context_offloader/test_plugin.py
```diff
@@ -0,0 +1,447 @@
...
+    def test_error_status_still_offloaded(self, plugin, mock_agent):
+        large_text = "x" * 200
+        event = _make_event(mock_agent, large_text, status="error")
+
+        plugin._handle_tool_result(event)
+
+        assert "[Offloaded:" in event.result["content"][0]["text"]
+        assert event.result["status"] == "error"
+
+    def test_storage_failure_keeps_original(self, mock_agent, caplog):
```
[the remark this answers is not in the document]
lizradway: added test for orphaned blocks after storage fail

src/strands/vended_plugins/context_offloader/plugin.py
```diff
@@ -0,0 +1,311 @@
...
+        try:
+            content_bytes, content_type = self._storage.retrieve(reference)
+        except KeyError:
+            return f"Error: reference not found: {reference}"
+
+        if content_type.startswith("text/") or content_type == "application/json":
+            return content_bytes.decode("utf-8")
+
+        return (
+            f"[Binary content: {content_type}, {len(content_bytes):,} bytes]\n"
```
lizradway: Updating to return tool result type

src/strands/vended_plugins/context_offloader/storage.py
```diff
@@ -0,0 +1,386 @@
...
+
+        sanitized_key = _sanitize_id(key)
+        timestamp_ms = int(time.time() * 1000)
+        ext = _extension_for(content_type)
+        with self._lock:
+            self._counter += 1
+            counter = self._counter
+            filename = f"{timestamp_ms}_{counter}_{sanitized_key}{ext}"
+            self._content_types[filename] = content_type
+            self._save_metadata()
```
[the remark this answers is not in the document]
lizradway: extreme edge case -- this is over engineering

src/strands/vended_plugins/context_offloader/__init__.py
```diff
@@ -0,0 +1,46 @@
+"""Context offloader plugin for Strands Agents.
```
opieter-aws, line 1: I like the new naming! Makes it general enough for extension to messages

src/strands/vended_plugins/context_offloader/plugin.py
```diff
@@ -0,0 +1,327 @@
...
+        ```
+    """
+
+    name = "context_offloader"
+
+    def __init__(
+        self,
+        storage: Storage,
+        max_result_chars: int = _DEFAULT_MAX_RESULT_CHARS,
+        preview_chars: int = _DEFAULT_PREVIEW_CHARS,
```
opieter-aws: Where did we land with the discussion to make this token-based rather than character-based?

opieter-aws (dismissed):

src/strands/vended_plugins/context_offloader/plugin.py
lizradway: followed up offline:
- character based in python until v2. 
- token based in typescript

mkmeral: > character based in python until v2. token based in typescript

why? we merged the tokenizer, right? why don't we use it?

src/strands/vended_plugins/context_offloader/plugin.py
```diff
@@ -0,0 +1,327 @@
...
+        if preview_chars >= max_result_chars:
+            raise ValueError("preview_chars must be less than max_result_chars")
+
+        self._storage = storage
+        self._max_result_chars = max_result_chars
+        self._preview_chars = preview_chars
+        super().__init__()
+
+    @tool(context=True)
+    def retrieve_offloaded_content(
```
mkmeral, line 150: does this tool let you process the data? if I am returning the data as is, won't it get externalized again? Did we test this e2e? 

also can we make the tool optional somehow? do we want this tool in all agents? (bc the context offloader imo should be in all agents)

src/strands/vended_plugins/context_offloader/plugin.py
mkmeral: we literally just merged a decision for it :p

src/strands/vended_plugins/context_offloader/plugin.py
```diff
@@ -0,0 +1,327 @@
...
+        agent: Agent = event.agent
+        state_data = agent.state.get(_STATE_KEY)
+        last_injection = state_data.get("last_injection") if isinstance(state_data, dict) else None
+
+        current_prompt = agent.system_prompt or ""
+
+        if last_injection is not None and last_injection in current_prompt:
+            return
+
+        injection = _SYSTEM_PROMPT_INJECTION
```
mkmeral: do we need to update the system prompt? why don't we give this info as replaced tool result instead?

src/strands/vended_plugins/context_offloader/plugin.py
mkmeral: nit: the main value of this tool will come from the fact that LLM can break it down, search, and analyze the necessary pieces.  That's why we want filesystem as default, bc llm can just use grep/head/tail to do those stuff

I don't think this tool actually provides that

src/strands/vended_plugins/context_offloader/plugin.py
mkmeral: personally i've seen that (last message actually telling llm what happened and what to do) helps with model behavior much better, compared to couple lines in system prompt (out of hundreds or thousands)

src/strands/vended_plugins/context_offloader/storage.py
```diff
@@ -0,0 +1,381 @@
...
+
+    Returns:
+        A sanitized string safe for use in filenames.
+    """
+    sanitized = raw_id.replace("..", "_").replace("/", "_").replace("\\", "_")
+    sanitized = re.sub(r"[^\w\-.]", "_", sanitized)
+    return sanitized
+
+
+def _extension_for(content_type: str) -> str:
```
mkmeral: nit: I'd make this a plugin concern and not a storage concern. storage just need to know the ref and the bytes. that's it.

otherwise later on, plugin might need to manage different references with extension or no extension, etc

src/strands/vended_plugins/context_offloader/plugin.py
lizradway: yes -- due to the system prompt, the result is only pulled in full when the llm needs full info that isnt satisfied in the preview (which it often is). This doesn't get externalized again since we [check + return in `_handle_tool_result `](https://github.com/strands-agents/sdk-python/pull/2162/changes#diff-bb9d485cba56ebbb1b899038c2762982fa03a36e1b1b2fb30efa6d70e2418f51R203) if the tool name is `retrieve_offloaded_content`, preventing any possible infinite recursion.

the launch plan is to release as an opt-in plugin that eventually is built in in the auto optimize context-management setting.


here's a brief llm gen summary of the demos i ran:
| Demo | What it tested | Agent behavior |
|------|---------------|----------------|
| **In-Memory** (summarize 200 lines) | Text offloading, preview-based reasoning | Used preview only, no retrieval. 1 tool call. |
| **File-Based** (summarize 200 lines) | File storage, `.metadata.json` sidecar | Used preview only, no retrieval. Files written to disk. |
| **JSON content** (50 records) | JSON serialization + offloading | Used preview only, summarized query structure. |
| **Mixed content** (text + image + doc) | Per-block storage, placeholders with refs | Described all 3 blocks from placeholders, no retrieval. |
| **Agent-driven retrieval** (exact S0150 data) | Retrieval tool when preview insufficient | Preview showed lines 1-8 only. Agent retrieved full content, found S0150. |
| **JSON retrieval** (count WARNINGs) | JSON retrieval, native type return | Preview only showed first 3 records. Agent retrieved full JSON, counted 5 WARNING sensors correctly. |

lizradway: ~in follow up, can add to sys prompt specification that the agent can alternatively run bash/etc commands on the file system rather than use full retrieve tool, if you think that would be useful + llm cannot discover on their own.~

Actually nvm, I don't want to couple tools the agent doesn't necessarily have to this plugin. Model driven approach -> agent can use any tools they have access to to act on the filepath imo

src/strands/vended_plugins/context_offloader/plugin.py
lizradway: im open to adding prompting in both system prompt and replaced tool result

opieter-aws (dismissed):

opieter-aws:
nit: Should the PR title reference `offloading` for documentation purposes?

src/strands/vended_plugins/context_offloader/plugin.py
mkmeral: > here's a brief llm gen summary of the demos i ran

these are demos, as I see most of them didn't even use this tool

I don't like this tool, it does not add value, except for bypassing context offloading. It does not provide any search, filter, etc. So in terms of context management, it's mostly useless. I think this only actually helps compared to not using the filter when tool result is multiple pieces, so returning a single reference is essentially "filtering".

That said, there is no way to use in memory storage without this tool, so it should be provided with the script

Couple of things here:
1. This tool must be optional. If we add it right now, we create backwards compatibility concern. 
2. I'd prefer to have more filtering, searching, shrinking (maybe images?), etc. in this tool, if we are providing it. that said, this is less of a priority than the item above

src/strands/vended_plugins/context_offloader/plugin.py
mkmeral: my point is more on, do we need it in system prompt? Instead of the system prompt, you can just add that string to tool result, and get rid of the `BeforeInvocationEvent`  hook. So tool result with template would return sth like 

```
Tool result is offloaded to external storage due to size. 
Use the tools you have in order to selectively access the data you need ... <or sth similar>

[Offloaded: 1 blocks, ~50,000 tokens]
[Use the preview below to answer if possible. If you need the full content,
call retrieve_offloaded_content with a reference.]

<first ~1000 tokens of the original text>...

[Stored references:]
  mem_1_abc123_0 (text/plain, 200,000 chars)

```

This would reduce the complexity of the plugin (1 less hook), and it should perform the same

src/strands/vended_plugins/context_offloader/plugin.py
lizradway: > these are demos, as I see most of them didn't even use this tool

these demos all used the tool successfully. the point of these demos is that only the last two demos needed to retrieve the full content. the others were able to use only the preview and can use the externalized full tool result later if needed.

lizradway: discussed offline:
- tool enabled as default makes sense for this plugin on its own as it stands today
- however once default storage is switched to VFS, it is optimal for agent to use sql/grep/shell/etc to navigate offloaded tool results.
- since switching default to disabled once default storage is VFS is a breaking change, it makes sense to start with default disabled/required choice for tool enablement.

src/strands/vended_plugins/context_offloader/plugin.py
lizradway: followed up offline:  tokenizer is compatible

src/strands/vended_plugins/context_offloader/plugin.py
```diff
@@ -283,10 +266,18 @@ async def _handle_tool_result(self, event: AfterToolCallEvent) -> None:
         # Build preview text — use tiktoken for exact slicing when available
         preview = self._slice_preview(full_text) if full_text else ""
         ref_lines = "\n".join(f"  {ref} ({desc})" for ref, _, desc in references if ref)
+
+        guidance = (
+            "Tool result was offloaded to external storage due to size.\n"
+            "Use the preview below to answer if possible.\n"
```
mkmeral, line 272: nit: I suspect this line can cause LLMs to rely more on preview, and possibly lead to hallucinations. I can see an LLM being lazy and just trying to deduce stuff from the previous instead of double checking

src/strands/vended_plugins/context_offloader/plugin.py
```diff
@@ -283,10 +266,18 @@ async def _handle_tool_result(self, event: AfterToolCallEvent) -> None:
         # Build preview text — use tiktoken for exact slicing when available
         preview = self._slice_preview(full_text) if full_text else ""
         ref_lines = "\n".join(f"  {ref} ({desc})" for ref, _, desc in references if ref)
+
+        guidance = (
```
mkmeral, line 270: not a blocker: can we make this configurable. when we add magic strings to LLM context, it can cause guardrails and other unintended consequences. It's a good practice to make sure customers can modify these magic strings we pass in for their use cases/architectures

src/strands/models/model.py
```diff
@@ -41,7 +41,7 @@ def _heuristic_estimate_json(obj: Any) -> int:
 
 
 @functools.lru_cache(maxsize=1)
-def _get_encoding() -> Any:
+def get_encoding() -> Any:
```
mkmeral: can we keep the private name for now? I don't think that's a good public method that aligns with Model API. Also the shapes are not great. It's an internal helper right now, and it should be named like that

Also we might want to expose how we do estimations, so customers can build their own logic similar to us. it can be a followup though

src/strands/vended_plugins/context_offloader/plugin.py
```diff
@@ -0,0 +1,336 @@
...
+            # Remove the auto-discovered retrieval tool
+            self._tools = [t for t in self._tools if t.tool_name != "retrieve_offloaded_content"]
+
+    @tool(context=True)
+    def retrieve_offloaded_content(
+        self,
+        reference: str,
+        tool_context: ToolContext,
+    ) -> dict | str:
+        """Retrieve offloaded content by reference.
```
mkmeral, line 155: nit: maybe a bit more context? this is a tool description. we can give a bit more context here.

or we can do it as followup. i'm thinking we should probably add regex search and json path search for text and json context. it'd improve the performance of the tool. but it's not necessary right now

mkmeral (commented):
`get_encoding` comment is the main one. otherwise it looks good to me (if review agent also agrees :p )

src/strands/vended_plugins/context_offloader/plugin.py
lizradway: I can run some benchmarks on different prompts and chose one with the best performance as follow up :)

src/strands/models/model.py
lizradway: sure, reverted.

src/strands/vended_plugins/context_offloader/plugin.py
lizradway: can experiment in follow up A/B testing on guidance

src/strands/vended_plugins/context_offloader/plugin.py
lizradway: yeah that makes sense. can add when i update guidance after full A/B testing

src/strands/models/model.py
lizradway: > we might want to expose how we do estimations

Thomas & I were just discussing exposing this. plus one

src/strands/vended_plugins/context_offloader/storage.py
```diff
@@ -0,0 +1,373 @@
...
+import time
+from pathlib import Path
+from typing import Any, Protocol, runtime_checkable
+
+import boto3
+from botocore.config import Config as BotocoreConfig
+from botocore.exceptions import ClientError
+
+
+def _sanitize_id(raw_id: str) -> str:
```
JackYPCOnline, line 40: Consider _identifier.validate()? This function replaces bad input right? This might mask bugs?
Also S3 also need prefix validaiton. You can reference `_get_session_path` in s3_session_manager

lizradway: - tool use ids are model generated, not user provided so we can’t reject them, only make them safe for filenames. this function makes them safe for FS
- we already do the same s3 validation as session manager (strip / normalization)

src/strands/vended_plugins/context_offloader/plugin.py
```diff
@@ -0,0 +1,336 @@
...
+
+        Use this tool when you see a placeholder with a reference (ref: ...)
+        and need the full content.
+
+        Args:
+            reference: The reference string from the offload placeholder.
+            tool_context: Injected by the framework. Not user-facing.
+        """
+        try:
+            content_bytes, content_type = self._storage.retrieve(reference)
```
mkmeral, line 165: one question, and I'm not sure if I'm too opinionated here 😅 

if we give the same plugin to two agents, do we want them to use the same storage? 

```python
plugin = ContextOffloader(...)

agent1 = Agent(plugins=[plugin])
agent2 = Agent(plugins=[plugin])
```

what should be the behavior?

src/strands/vended_plugins/context_offloader/plugin.py
```diff
@@ -0,0 +1,336 @@
...
+
+        return content_bytes.decode("utf-8", errors="replace")
+
+    @hook
+    async def _handle_tool_result(self, event: AfterToolCallEvent) -> None:
+        """Intercept oversized tool results, offload per-block, and replace with preview."""
+        if event.cancel_message is not None:
+            return
+
+        if event.tool_use.get("name") == "retrieve_offloaded_content":
```
mkmeral: nit: can we check if the tool is enabled? 

Also the correct usage here would be accessing the `self.retrieve_offloaded_content.name` and comparing against that. technically tool descriptions can be updated by the user (fyi `self.retrieve_offloaded_content` is an `AgentTool` instance)

mkmeral: actually can we fix this and add a section to AGENTS.md? :) 
https://github.com/strands-agents/sdk-python/blob/main/src/strands/tools/tools.py#L201

this might come as bug reports later on

mkmeral: @agent-of-mkmeral  do we have similar issues anywhere else in the repository?

lizradway: makes sense, updated accessor now and can update AGENTS.md separately in python + TS

src/strands/vended_plugins/context_offloader/plugin.py
lizradway: im pro shared storage (current behavior).

user could create isolated plugin instances for isolated storage:
```
# Shared storage 
plugin = ContextOffloader(storage=InMemoryStorage())
agent1 = Agent(plugins=[plugin])
agent2 = Agent(plugins=[plugin])

# Separate storage
agent1 = Agent(plugins=[ContextOffloader(storage=InMemoryStorage())])
agent2 = Agent(plugins=[ContextOffloader(storage=InMemoryStorage())])
```

mkmeral (approved):

PR #2165, opened 2026-04-20
feat: add agent invoke with checkpoint and related classes

JackYPCOnline:
## Description

Adds `invoke_with_checkpoint()` — a checkpoint-based execution mode for durable agent workflows.

The normal agent event loop runs model→tools→model→done as one continuous function. If it crashes at step 47, you restart from step 1. `invoke_with_checkpoint` breaks this into individually resumable steps: each call does exactly one unit of I/O (one model call OR one tool execution), then returns a `Checkpoint`. The caller (e.g., a Temporal workflow) persists that checkpoint. On crash, execution resumes from the last checkpoint — not from scratch.

**Key design decisions:**
- Per-tool granularity: if the model requests 5 tools and you crash after tool 3, only tools 4-5 re-execute on recovery
- Model-driven error handling: tool errors become tool results the model sees and decides on (no special error paths)
- Provider-agnostic: works with any model provider that implements the standard `stream()` contract
- V1 simplifications: bypasses hooks, tracing, retry strategy, structured output, and conversation management — the orchestrator handles these concerns

**New files:**
- `src/strands/experimental/checkpoint/__init__.py` — exports
- `src/strands/experimental/checkpoint/types.py` — `Checkpoint` dataclass with serialization
- `src/strands/experimental/checkpoint/invoke.py` — the step-by-step execution logic

**Modified files:**
- `src/strands/agent/agent.py` — adds `Agent.invoke_with_checkpoint()` convenience method
- `src/strands/experimental/__init__.py` — registers the checkpoint submodule

**Usage:**
```python
from strands.experimental.checkpoint import Checkpoint

result = await agent.invoke_with_checkpoint("Hello")
while isinstance(result, Checkpoint):
    result = await agent.invoke_with_checkpoint(checkpoint=result)
# result is AgentResult
```

## Related Issues

https://github.com/strands-agents/sdk-typescript/issues/822

## Type of Change

New feature

## Testing

22 unit tests covering:
- Checkpoint serialization/deserialization round-trips
- Fresh invocation (no tools, with tools)
- Per-tool granularity (5 tools = 5 separate steps)
- Multi-cycle execution (model → tool → model → tool → done)
- Crash recovery (serialize checkpoint, new agent, resume)
- Model-driven tool errors (errors become tool results)
- Complex scenario (5 tools, 3rd fails, model retries just that tool)
- Agent method wrapper
- Edge cases (invalid position, unknown tool, tool returning no result, invocation_state contract)

## Checklist

----

### DIFF
--- a/src/strands/agent/agent.py
+++ b/src/strands/agent/agent.py
@@ -39,6 +39,7 @@ from ..types._snapshot import (
 )
 
 if TYPE_CHECKING:
+    from ..experimental.checkpoint.types import Checkpoint as _Checkpoint
     from ..tools import ToolProvider
 from ..handlers.callback_handler import PrintingCallbackHandler, null_callback_handler
 from ..hooks import (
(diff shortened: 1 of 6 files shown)


src/strands/agent/agent.py
pgrayy, line 1: We should be able to support this functionality through the existing `__call__`/`invoke_async`/`stream_async` methods. We can implement the logic in a similar manner to InterruptResponse ([docs](https://strandsagents.com/docs/user-guide/concepts/interrupts/)). We just have the invoke methods accept a new content block type that takes the previous step result as a payload.

Alternatively, we could have the invoke methods accept another param for running checkpoints.

JackYPCOnline: The reason why I have this separate `invoke` functions are:
1.  checkpoint mode doesn't support structured output, invocation_state, or fire standard hooks. We need to do valiation for those combinations?
2. Make it distinct API shape so that users will be  aware.
3. if we want to fold the checkpoint to `invoke_async`, what do you think on the return-type question (union vs. optional field on AgentResult)?

pgrayy: 1. Is there a particular reason why it wouldn't support those things? They should all work the same.
2. I think this is just a matter of documentation. What we gain though is consistency with other loop altering functionality like interrupts and structured output.
3. For consistency with interrupts and structured output, probably would recommend returning AgentResult.

The one tricky thing we would have to figure out though is traces.

PR #2167, opened 2026-04-20
fix(telemetry): make ThreadingInstrumentor opt-in with env-var precedence

jpr5:
strands>=1.35.0 unconditionally calls `ThreadingInstrumentor().instrument()` in `Tracer.__init__`, monkey-patching `concurrent.futures.ThreadPoolExecutor.submit` process-wide for every strands user. It ignores the standard `OTEL_PYTHON_DISABLED_INSTRUMENTATIONS=threading` opt-out, double-wraps when another OTel distro (Azure Monitor, opentelemetry-distro, AWS OTel) already instrumented threading, and crashes the host if `instrument()` raises.

**This PR is additive and non-breaking: threading instrumentation stays ON by default (unchanged behavior).** It adds supported opt-out/opt-in levers and hardens the call path.

Resolver precedence (highest → lowest):
1. `OTEL_PYTHON_DISABLED_INSTRUMENTATIONS` containing `threading` → always disabled (matches OpenTelemetry auto-loader semantics).
2. Explicit `Tracer(instrument_threading=True|False)` kwarg.
3. `STRANDS_INSTRUMENT_THREADING` env var (`1`/`true`/`yes`/`on` enable; `0`/`false`/`no`/`off` disable; unrecognized non-empty → WARNING + default).
4. Default: **ENABLED**.

Also adds:
- an **idempotency guard** (reads `_is_instrumented_by_opentelemetry` to avoid stacking wrappers when the host already instrumented threading), and
- **graceful-failure handling** — `instrument()` exceptions are caught and logged (ERROR when explicitly requested, WARNING when auto-enabled) so telemetry failure never crashes the host.

Tests run in isolated subprocesses (`BaseInstrumentor` is a process-wide singleton): on-by-default, env opt-in/opt-out (incl. `off`), kwarg opt-in/out, disable-env precedence over kwarg, idempotency, graceful failure, ERROR-vs-WARNING severity, and warn-on-unrecognized.

Flipping the default to **off** is a breaking change deferred to v2 (tracked on this thread).

### DIFF
--- a/src/strands/telemetry/tracer.py
+++ b/src/strands/telemetry/tracer.py
@@ -87,13 +87,35 @@ class Tracer:
     or "gen_ai_use_latest_invocation_tokens", respectively, in the OTEL_SEMCONV_STABILITY_OPT_IN environment variable.
     """
 
-    def __init__(self) -> None:
-        """Initialize the tracer."""
+    def __init__(self, instrument_threading: bool | None = None) -> None:
+        """Initialize the tracer.
+
+        Args:
+            instrument_threading: When True, install OpenTelemetry's
+                ``ThreadingInstrumentor`` so spans propagate across
+                ``concurrent.futures.ThreadPoolExecutor``/``threading.Thread``
+                boundaries. When False, skip it. When None (default), honor
+                the ``STRANDS_INSTRUMENT_THREADING`` environment variable
+                (``false``/``0``/``no`` disables, anything else leaves the
+                default in place). Threading instrumentation is ON by
+                default (preserving historical strands behavior): it
+                propagates spans across ``ThreadPoolExecutor``/``Thread``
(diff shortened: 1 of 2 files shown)


lizradway (commented):
flipping this to false would be a breaking change for users that we cannot do in a minor version, but can consider for v2.

however, would be happy to keep the rest if you are able to cherry pick that out.

thanks for your contribution!

jpr5:
Makes sense — keeping the default on so this stays additive. I left in the kwarg, the env var, and the `OTEL_PYTHON_DISABLED_INSTRUMENTATIONS` handling, plus a guard so we don't double-wrap when something's already instrumented threading, and a try/except so a telemetry failure can't take the host down. Just pushed that.

LMK what you think @lizradway 🙏

PR #2181, opened 2026-04-21
feat: introduce checkpoint in experimental

JackYPCOnline:
## Description

Replaces the experimental checkpoint module with a redesigned `Checkpoint` dataclass and adds `"checkpoint"` to the `StopReason` type. This is PR 1 of 2 — types and tests only, no behavioral changes.

### What changed

- **New `Checkpoint` dataclass** (`experimental/checkpoint/checkpoint.py`): Schema-versioned with `to_dict()`/`from_dict()` serialization, `CheckpointPosition` literal (`"after_model"` | `"after_tools"`), and `app_data` dict for provider-owned metadata (e.g. Temporal workflow IDs). `from_dict()` rejects schema version mismatches and logs a warning on unknown keys for debuggability.
- **`"checkpoint"` added to `StopReason`** (`types/event_loop.py`): Type-level addition only — nothing emits this value yet.
- **Updated `experimental/checkpoint/__init__.py`**: exports `Checkpoint`, `CHECKPOINT_SCHEMA_VERSION`, and `CheckpointPosition`.

### What this PR does NOT change

- No changes to `Agent`, `AgentResult`, `EventLoopStopEvent`, or `event_loop.py` behavior.
- No new runtime code paths — nothing emits `"checkpoint"` stop reason yet.
- All existing tests pass unmodified (2611 passed).

### Follow-up PR (PR 2 of 2) will include

- `checkpointing=True` flag on `Agent` constructor
- Event loop integration: checkpoint emission at `after_model` and `after_tools` cycle boundaries
- `checkpointResume` content block for resuming from a checkpoint
- `checkpoint` field on `AgentResult` and extended `EventLoopStopEvent` tuple
- `Checkpoint` promoted to top-level `strands` export
- Integration tests for the full pause/resume flow

## Type of Change

New feature

## Testing

- 6 unit tests with 100% coverage on `checkpoint.py`: round-trip serialization, schema version immutability, schema mismatch rejection, defaults, unknown field warning, missing version rejection.

- [x] I ran `hatch run prepare`

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [ ] I have updated the documentation accordingly
- [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2190, opened 2026-04-22
feat(checkpoint): wire checkpointing into agent event loop

JackYPCOnline:
## Description

Wires the `Checkpoint` data model (landed in #2181) into the agent runtime so an opt-in `checkpointing=True` agent pauses at ReAct cycle boundaries and emits a serializable pause-point marker. The marker travels through `AgentResult.checkpoint` and is passed back to a fresh agent through a `checkpointResume` block. The design mirrors the existing interrupt pattern: `stop_reason="checkpoint"` to signal the pause, content-block resume, no new method surface to learn.

A `Checkpoint` is a position marker (which boundary fired and which cycle index), not a state snapshot. State persistence is the caller's responsibility. The recommended pairing is a `SessionManager` for state continuity plus `checkpointing=True` for boundary signalling. 

**User-facing API (zero breaking changes — opt-in only):**

```python
from strands import Agent
from strands.session import FileSessionManager

agent = Agent(
    tools=[...],
    session_manager=FileSessionManager(session_id="run-1", storage_dir="..."),
    checkpointing=True,
)

result = await agent.invoke_async("do the thing")

while result.stop_reason == "checkpoint":
    save_somewhere(result.checkpoint.to_dict())
    # later, possibly in a fresh process / activity:
    fresh = Agent(
        tools=[...],
        session_manager=FileSessionManager(session_id="run-1", storage_dir="..."),
        checkpointing=True,
    )
    result = await fresh.invoke_async(
        {"checkpointResume": {"checkpoint": load_somewhere()}}
    )

print(result.message)  # stop_reason == "end_turn"
```

**V0 known limitations:**

- Metrics reset on each resume call.
- `OpenAIResponsesModel(stateful=True)` not supported.
- `BeforeInvocationEvent` / `AfterInvocationEvent` fire on every resume (same as interrupts).
- Per-tool granularity within a cycle requires a custom `ToolExecutor`. The SDK checkpoint operates at cycle boundaries.
- Streaming callbacks do not re-emit on replay.

## Type of Change

New feature

## Testing

Verified the changes do not break functionality or introduce warnings in consuming repositories.

Evidence from fresh runs:

- `hatch test` — 3026 passed, 4 skipped, 0 failed.
- `hatch run hatch-static-analysis:lint-check` — `ruff check` and `mypy` both clean.
- `hatch run hatch-static-analysis:format-check` — all files formatted.\

## Checklist

- [ ] I have updated the documentation accordingly *(user-guide page is a follow-up PR in `agent-docs`; module-level docstring in `checkpoint.py` covers V0 limitations, precedence rules, and the recommended SessionManager pairing)*
- [ ] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed *(reference Temporal / Dapr / Step Functions examples are the next milestone in the durable-execution tracking plan)*
- [x] Any dependent changes have been merged and published *(Part A — #2181 — is merged on `main`; this PR builds on it)*

---

### DIFF
--- a/strands-py/src/strands/agent/agent.py
+++ b/strands-py/src/strands/agent/agent.py
@@ -29,6 +29,7 @@ from .. import _identifier
 from .._async import run_async
 from ..event_loop._retry import ModelRetryStrategy
 from ..event_loop.event_loop import INITIAL_DELAY, MAX_ATTEMPTS, MAX_DELAY, event_loop_cycle
+from ..experimental.checkpoint import Checkpoint, CheckpointPosition
 from ..tools._tool_helpers import generate_missing_tool_result_content
 from ..types._snapshot import (
     SNAPSHOT_SCHEMA_VERSION,
@@ -146,6 +147,7 @@ class Agent(AgentBase):
         tool_executor: ToolExecutor | None = None,
         retry_strategy: ModelRetryStrategy | _DefaultRetryStrategySentinel | None = _DEFAULT_RETRY_STRATEGY,
         concurrent_invocation_mode: ConcurrentInvocationMode = ConcurrentInvocationMode.THROW,
+        checkpointing: bool = False,
     ):
         """Initialize the Agent with the specified configuration.
 
@@ -214,6 +216,13 @@ class Agent(AgentBase):
                 Set to "unsafe_reentrant" to skip lock acquisition entirely, allowing concurrent invocations.
                 Warning: "unsafe_reentrant" makes no guarantees about resulting behavior and is provided
                 only for advanced use cases where the caller understands the risks.
+            checkpointing: When True, the event loop pauses at cycle boundaries
+                (after_model, after_tools) and returns ``stop_reason="checkpoint"``
+                with a populated ``checkpoint`` field. Resume by passing the
+                checkpoint back as ``{"checkpointResume": {"checkpoint": ...}}``.
+                The SDK does not capture conversation state in the checkpoint;
+                pair with a SessionManager for cross-process state continuity.
+                Defaults to False. See :mod:`strands.experimental.checkpoint`.
 
         Raises:
             ValueError: If agent id contains path separators.
@@ -304,6 +313,12 @@ class Agent(AgentBase):
 
         self._interrupt_state = _InterruptState()
 
+        # Checkpointing: pause at cycle boundaries when enabled.
+        self._checkpointing: bool = checkpointing
+        self._checkpoint: Checkpoint | None = None
+        self._checkpoint_cycle_index: int = 0
+        self._checkpoint_resume_position: CheckpointPosition | None = None
+
         # Runtime state for model providers (e.g., server-side response ids)
         self._model_state: dict[str, Any] = {}
 
@@ -374,7 +389,7 @@ class Agent(AgentBase):
         This method is thread-safe and can be called from any context
         (e.g., another thread, web request handler, background task).
 
-        The agent will stop gracefully at the next checkpoint:
+        The agent will stop gracefully at the next cancellation-safe point:
         - During model response streaming
         - Before tool execution
 
@@ -832,6 +847,12 @@ class Agent(AgentBase):
 
             self.event_loop_metrics.reset_usage_metrics()
 
+            # Reset invocation-scoped checkpoint state. On resume, the event loop's
+            # priming step re-derives the cycle index from the resumed checkpoint,
+            # so this reset only affects the fresh-prompt path.
+            self._checkpoint_cycle_index = 0
+            self._checkpoint_resume_position = None
+
             merged_state = {}
             if kwargs:
                 warnings.warn("`**kwargs` parameter is deprecating, use `invocation_state` instead.", stacklevel=2)
@@ -1006,8 +1027,34 @@ class Agent(AgentBase):
             if structured_output_context:
                 structured_output_context.cleanup(self.tool_registry)
 
+    def _try_consume_checkpoint_resume(self, prompt: Any) -> bool:
+        """Consume a ``checkpointResume`` prompt block, returning True if found.
+
+        The block is a dict of the form ``{"checkpointResume": {"checkpoint": ...}}``.
+        A missing ``checkpoint`` key raises ``KeyError``; ``checkpointing=False``
+        raises ``ValueError``; a schema mismatch raises ``CheckpointException``.
+        """
+        if not (isinstance(prompt, dict) and "checkpointResume" in prompt):
+            return False
+
+        if not self._checkpointing:
+            raise ValueError(
+                "Received checkpointResume block but agent was created with checkpointing=False. "
+                "Pass checkpointing=True when constructing the Agent."
+            )
+
+        payload = prompt["checkpointResume"]
+        if not isinstance(payload, dict) or "checkpoint" not in payload:
+            raise KeyError("checkpoint | missing required key in checkpointResume block")
+
+        self._checkpoint = Checkpoint.from_dict(payload["checkpoint"])
+        return True
+
     async def _convert_prompt_to_messages(self, prompt: AgentInput) -> Messages:
         if self._interrupt_state.activated:
+            return []
+
+        if self._try_consume_checkpoint_resume(prompt):
             return []
 
         messages: Messages | None = None
--- a/strands-py/src/strands/event_loop/event_loop.py
+++ b/strands-py/src/strands/event_loop/event_loop.py
@@ -15,6 +15,7 @@ from typing import TYPE_CHECKING, Any
 
 from opentelemetry import trace as trace_api
 
+from ..experimental.checkpoint import Checkpoint, CheckpointPosition
 from ..hooks import AfterModelCallEvent, BeforeModelCallEvent, MessageAddedEvent
 from ..telemetry.metrics import Trace
 from ..telemetry.tracer import Tracer, get_tracer
@@ -117,6 +118,27 @@ async def _estimate_input_tokens(agent: "Agent") -> int:
     )
 
 
+def _build_checkpoint_stop_event(
+    agent: "Agent",
+    position: CheckpointPosition,
+    cycle_index: int,
+    message: Message,
+    request_state: Any,
+) -> EventLoopStopEvent:
+    """Build a checkpoint stop event. Used at ``after_model`` and ``after_tools``."""
+    checkpoint = Checkpoint(
+        position=position,
+        cycle_index=cycle_index,
+    )
+    return EventLoopStopEvent(
+        "checkpoint",
+        message,
+        agent.event_loop_metrics,
+        request_state,
+        checkpoint=checkpoint,
+    )
+
+
 async def event_loop_cycle(
     agent: "Agent",
     invocation_state: dict[str, Any],
@@ -145,12 +167,16 @@ async def event_loop_cycle(
         structured_output_context: Optional context for structured output management.
 
     Yields:
-        Model and tool stream events. The last event is a tuple containing:
+        Model and tool stream events. The final ``EventLoopStopEvent`` payload
+        (``event["stop"]``) is a 7-tuple:
 
-            - StopReason: Reason the model stopped generating (e.g., "tool_use")
+            - StopReason: Reason the model stopped generating (e.g., "tool_use", "checkpoint")
             - Message: The generated message from the model
             - EventLoopMetrics: Updated metrics for the event loop
             - Any: Updated request state
+            - Sequence[Interrupt] | None: Interrupts raised during the cycle, if any
+            - BaseModel | None: Structured output result, if any
+            - Checkpoint | None: Checkpoint captured when stop_reason == "checkpoint"
 
     Raises:
         EventLoopException: If an error occurs during execution
@@ -164,6 +190,18 @@ async def event_loop_cycle(
     # Initialize state and get cycle trace
     if "request_state" not in invocation_state:
         invocation_state["request_state"] = {}
+
+    # Consume the resume marker (one-shot).
+    resume_context = agent._checkpoint
+    if resume_context is not None:
+        agent._checkpoint = None
+        # after_tools means that cycle finished; resume increments cycle_index.
+        next_cycle = (
+            resume_context.cycle_index + 1 if resume_context.position == "after_tools" else resume_context.cycle_index
+        )
+        agent._checkpoint_cycle_index = next_cycle
+        agent._checkpoint_resume_position = resume_context.position
+
     attributes = {"event_loop_cycle_id": str(invocation_state.get("event_loop_cycle_id"))}
     cycle_start_time, cycle_trace = agent.event_loop_metrics.start_cycle(attributes=attributes)
     invocation_state["event_loop_cycle_trace"] = cycle_trace
@@ -223,6 +261,24 @@ async def event_loop_cycle(
                 )
 
             if stop_reason == "tool_use":
+                # Emit after_model checkpoint, unless we just resumed from one.
+                if agent._checkpointing and not agent._cancel_signal.is_set():
+                    resume_position = agent._checkpoint_resume_position
+                    agent._checkpoint_resume_position = None
+                    if resume_position != "after_model":
+                        cycle_index = agent._checkpoint_cycle_index
+                        agent.event_loop_metrics.end_cycle(cycle_start_time, cycle_trace)
+                        if cycle_span:
+                            tracer.end_event_loop_cycle_span(span=cycle_span, message=message)
+                        yield _build_checkpoint_stop_event(
+                            agent=agent,
+                            position="after_model",
+                            cycle_index=cycle_index,
+                            message=message,
+                            request_state=invocation_state["request_state"],
+                        )
+                        return
+
                 # Handle tool execution
                 tool_events = _handle_tool_execution(
                     stop_reason,
@@ -637,6 +693,31 @@ async def _handle_tool_execution(
             agent.event_loop_metrics,
             invocation_state["request_state"],
             structured_output=structured_output_result,
+        )
+        return
+
+    # Emit after_tools checkpoint. Only fires on tool_use cycles: a model that
+    # returns end_turn first never reaches this branch.
+    if agent._checkpointing and not agent._cancel_signal.is_set():
+        cycle_index = agent._checkpoint_cycle_index
+        agent._checkpoint_cycle_index = cycle_index + 1
+        yield _build_checkpoint_stop_event(
+            agent=agent,
+            position="after_tools",
+            cycle_index=cycle_index,
+            message=message,
+            request_state=invocation_state["request_state"],
+        )
+        return
+
+    # If checkpointing is on and cancel suppressed the checkpoint above, emit
+    # "cancelled" now to avoid an extra model call.
+    if agent._checkpointing and agent._cancel_signal.is_set():
+        yield EventLoopStopEvent(
+            "cancelled",
+            message,
+            agent.event_loop_metrics,
+            invocation_state["request_state"],
         )
         return
 
--- a/strands-py/src/strands/experimental/checkpoint/checkpoint.py
+++ b/strands-py/src/strands/experimental/checkpoint/checkpoint.py
@@ -1,39 +1,39 @@
 """Checkpoint system for durable agent execution.
 
-Checkpoints enable crash-resilient agent workflows by capturing agent state at
-cycle boundaries in the agent loop. A durability provider (e.g. Temporal) can
-persist checkpoints and resume from them after failures.
+A ``Checkpoint`` is a pause-point marker emitted at agent cycle boundaries.
+It captures the position (which boundary fired) and the cycle index. It does
+**not** capture conversation state — pair with a ``SessionManager`` for
+cross-process state continuity.
 
-Two checkpoint positions per ReAct cycle:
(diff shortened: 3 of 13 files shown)


src/strands/agent/agent.py
```diff
@@ -998,6 +1009,46 @@ async def _convert_prompt_to_messages(self, prompt: AgentInput) -> Messages:
...
+                    raise TypeError(
+                        f"block_count=<{len(prompt)}> | only one checkpointResume block permitted per prompt"
+                    )
+
+                resume_block = prompt[0].get("checkpointResume", {})
+                if not isinstance(resume_block, dict) or "checkpoint" not in resume_block:
+                    raise KeyError("checkpoint | missing required key in checkpointResume block")
+
+                checkpoint = Checkpoint.from_dict(resume_block["checkpoint"])
+                self.load_snapshot(Snapshot.from_dict(checkpoint.snapshot))
```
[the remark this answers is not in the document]
JackYPCOnline: `SnapshotException` already carries a clear, structured message (`"Unsupported snapshot schema version: '{v}'. Current version: {c}"`). Adding a wrapper layer at the checkpoint level introduces indirection without adding information. If a user hits `SnapshotException` during checkpoint resume, the message + the call stack are enough to diagnose.

strands-py/src/strands/types/_events.py
```diff
@@ -237,8 +239,11 @@ def __init__(
             request_state: Final state of the agent execution
             interrupts: Interrupts raised by user during agent execution.
             structured_output: Optional structured output result
+            checkpoint: Optional checkpoint when stop_reason == "checkpoint".
         """
-        super().__init__({"stop": (stop_reason, message, metrics, request_state, interrupts, structured_output)})
+        super().__init__(
+            {"stop": (stop_reason, message, metrics, request_state, interrupts, structured_output, checkpoint)}
```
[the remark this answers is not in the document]
JackYPCOnline: yes, it is existing pattern, we don't need to worry about this.

src/strands/event_loop/event_loop.py
```diff
@@ -122,6 +123,19 @@ async def event_loop_cycle(
     # Initialize state and get cycle trace
     if "request_state" not in invocation_state:
         invocation_state["request_state"] = {}
+
+    # Consume checkpoint resume context (one-shot: cleared after reading).
+    # Cross-invocation state (resume context) lives on the agent; within-cycle
+    # transient state (resume position for the skip check, cycle index) lives
+    # in invocation_state.
+    resume_ctx = agent._checkpoint_resume_context
```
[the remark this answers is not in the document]
JackYPCOnline: The review is correct that this creates coupling. However, this is the *exact same* coupling pattern already in place for interrupts (`agent._interrupt_state` read + mutated by the event loop). The design decision for this PR to mirror the interrupt pattern by construction; diverging now would require justifying why checkpoint gets a public property surface while interrupt does not. If the coupling is a real problem, the refactor belongs at the interrupt-level too — a separate follow-up PR that touches both concerns consistently.
what do you think?

src/strands/agent/agent.py
```diff
@@ -998,6 +1009,46 @@ async def _convert_prompt_to_messages(self, prompt: AgentInput) -> Messages:
         if self._interrupt_state.activated:
             return []
 
+        # Resume detection — must run before existing shape handling so checkpointResume
+        # blocks aren't misinterpreted as content blocks. Mirrors _InterruptState.resume()
+        # conventions (TypeError for shape, KeyError for lookup, ValueError for misconfig;
```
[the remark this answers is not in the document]
JackYPCOnline: The SDK convention is standalone per-domain exceptions (`SessionException`, `SnapshotException`, and now `CheckpointException`). Catching `ValueError` should not accidentally catch checkpoint-specific failures — that would hide schema-version problems inside generic error handlers. Callers who need to distinguish "checkpoint restore failed" from "bad input" should catch `CheckpointException` explicitly; this is the pattern the SDK already uses.

src/strands/event_loop/event_loop.py
```diff
@@ -590,6 +648,31 @@ async def _handle_tool_execution(
...
+            agent=agent,
+            position="after_tools",
+            cycle_index=cycle_index,
+            message=message,
+            request_state=invocation_state["request_state"],
+        )
+        return
+
+    # If cancel fired during tool execution, emit a cancelled stop event now.
+    if agent._cancel_signal.is_set():
```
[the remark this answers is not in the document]
JackYPCOnline: Addressed, please double check if there are behavior changes.

gautamsirdeshmukh:
/strands I have reviewed this PR, and I don't think I see anything worth blocking on. Can you take a final thorough look? Only let me know if there are any crucial/important bugs, gaps or new risks being introduced. Make sure to read all previous comments on the PR before coming to any conclusions.

gautamsirdeshmukh (dismissed):

strands-py/src/strands/experimental/checkpoint/checkpoint.py
```diff
@@ -1,64 +1,62 @@
...
-  (e.g. TemporalToolExecutor).
+Checkpoints capture agent state at cycle boundaries so a durability provider
+(e.g. Temporal) can persist them and resume after failures.
+
+Positions per ReAct cycle:
+- ``after_model``: model call completed, tools not yet executed.
+- ``after_tools``: all tools executed, next model call pending.
+
+Per-tool granularity within a cycle is the ToolExecutor's responsibility
+(e.g. TemporalToolExecutor routes each tool to a separate activity).
```
pgrayy: Is this a pattern we would want to recommend? Would it work given that the expectation is for users to run the entire agent invoke in an activity?

strands-py/src/strands/experimental/checkpoint/checkpoint.py
```diff
@@ -1,64 +1,62 @@
...
+- Pause: ``AgentResult`` with ``stop_reason="checkpoint"`` and a populated
+  ``checkpoint`` field.
+- Resume: pass ``[{"checkpointResume": {"checkpoint": checkpoint.to_dict()}}]``
+  as the next prompt.
+
+Precedence:
+- Interrupts > checkpoint: an interrupt raised during a checkpointing cycle
+  returns ``stop_reason="interrupt"`` and skips the ``after_tools`` checkpoint.
+- Cancel > checkpoint: a cancel signal set at either checkpoint boundary
+  suppresses emission and surfaces as ``stop_reason="cancelled"``.
```
pgrayy: I don't think we need to mention this as a matter of precedence. It is more a pattern of order I would say. Let's say a user cancels. The idea is that if there is a checkpoint ahead, we run it. Upon resume, if they next reach the cancel checkpoint, then we cancel. Similarly for interrupts.

With all that said, this highlights that we should probably try to unify interrupts, cancellation, and checkpointing in some way. Maybe we can discuss offline. I'll first finish reviewing the logic here though.

strands-py/src/strands/experimental/checkpoint/checkpoint.py
```diff
@@ -1,64 +1,62 @@
...
+
+Known limitations:
+- ``EventLoopMetrics`` resets per invocation; aggregate across resumes yourself.
+- ``OpenAIResponsesModel(stateful=True)`` is not supported — the server-side
+  ``response_id`` is not captured.
```
pgrayy: Is there a reason we can't capture the response_id? It should just work I would think since it is managed in model_state.

JackYPCOnline: Limitation is from `snapshot`, right now it doesn't capture this. It is a follow up in my plan

strands-py/src/strands/experimental/checkpoint/checkpoint.py
JackYPCOnline: yep, this specific for outline the logic chain where adding this checkpointing will not change origin behavior.

strands-py/src/strands/experimental/checkpoint/checkpoint.py
JackYPCOnline: I think it is a open question I guess, we will have at least 2 approaches to achieve that.

JackYPCOnline:
/strands I did rebase from main, it is not adding new feature, can you review the PR against current main. Breaking change is not introduced by checkpointing.

strands-py/src/strands/agent/agent.py
```diff
@@ -1006,10 +1023,57 @@ async def _execute_event_loop_cycle(
...
+        self._checkpoint_resume_context = checkpoint
+        return True
+
     async def _convert_prompt_to_messages(self, prompt: AgentInput) -> Messages:
         if self._interrupt_state.activated:
             return []
 
+        # Resume detection — must run before existing shape handling so checkpointResume
+        # blocks aren't misinterpreted as content blocks.
+        if isinstance(prompt, list) and prompt and self._try_consume_checkpoint_resume(prompt):
```
pgrayy: Since only one checkpoint can be returned, we can accept a dict shape instead of list for resuming.

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -117,6 +118,28 @@ async def _estimate_input_tokens(agent: "Agent") -> int:
     )
 
 
+def _build_checkpoint_stop_event(
+    agent: "Agent",
+    position: "CheckpointPosition",
```
pgrayy: Nit: Since CheckpointPosition is imported top level, you don't need to wrap in quotes. Only necessary if imported under an `if TYPE_CHECKING`.

strands-py/src/strands/agent/agent.py
```diff
@@ -146,6 +147,7 @@ def __init__(
         tool_executor: ToolExecutor | None = None,
         retry_strategy: ModelRetryStrategy | _DefaultRetryStrategySentinel | None = _DEFAULT_RETRY_STRATEGY,
         concurrent_invocation_mode: ConcurrentInvocationMode = ConcurrentInvocationMode.THROW,
+        checkpointing: bool = False,
```
pgrayy, line 150: If we want to support more configurability around checkpointing, we could expand the type here to be something like `: bool | CheckpointingConfig`.

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -117,6 +118,28 @@ async def _estimate_input_tokens(agent: "Agent") -> int:
...
+    cycle_index: int,
+    message: Message,
+    request_state: Any,
+) -> EventLoopStopEvent:
+    """Build a checkpoint stop event. Used at ``after_model`` and ``after_tools``."""
+    checkpoint = Checkpoint(
+        position=position,
+        cycle_index=cycle_index,
+        snapshot=agent.take_snapshot(preset="session").to_dict(),
```
pgrayy: Considerations: Maybe we exclude snapshot if we detect a user has a session manager configured.

strands-py/src/strands/agent/agent.py
```diff
@@ -304,6 +311,10 @@ def __init__(
 
         self._interrupt_state = _InterruptState()
 
+        # Checkpointing: when True, event loop pauses at cycle boundaries
+        self._checkpointing: bool = checkpointing
+        self._checkpoint_resume_context: Checkpoint | None = None
```
pgrayy: Consideration: Call this `self._checkpoint`.

strands-py/src/strands/agent/agent.py
```diff
@@ -1006,10 +1023,57 @@ async def _execute_event_loop_cycle(
...
+
+        if len(prompt) != 1:
+            raise TypeError(f"block_count=<{len(prompt)}> | only one checkpointResume block permitted per prompt")
+
+        resume_block = prompt[0].get("checkpointResume", {})
+        if not isinstance(resume_block, dict) or "checkpoint" not in resume_block:
+            raise KeyError("checkpoint | missing required key in checkpointResume block")
+
+        checkpoint = Checkpoint.from_dict(resume_block["checkpoint"])
+        self.load_snapshot(Snapshot.from_dict(checkpoint.snapshot))
```
pgrayy: Per discussion, I would say let's not worry about putting snapshot in checkpoint. Users can take the snapshot themselves if they desire. Also, they could utilize session manager in tandem to their durable workflow.

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -164,6 +191,18 @@ async def event_loop_cycle(
...
+    # Consume checkpoint resume context (one-shot).
+    resume_context = agent._checkpoint_resume_context
+    if resume_context is not None:
+        agent._checkpoint_resume_context = None
+        # after_tools completed that cycle, so the next cycle starts at +1
+        next_cycle = (
+            resume_context.cycle_index + 1 if resume_context.position == "after_tools" else resume_context.cycle_index
+        )
+        invocation_state["_checkpoint_cycle_index"] = next_cycle
+        invocation_state["_checkpoint_resume_position"] = resume_context.position
```
pgrayy: `invocation_state` is user data so I think we want to avoid adding new fields to this.

strands-py/src/strands/agent/agent.py
JackYPCOnline: This can be something we can explore later.

strands-py/src/strands/experimental/checkpoint/checkpoint.py
```diff
@@ -1,64 +1,60 @@
...
+
+Precedence:
+- Interrupt > checkpoint: an interrupt during a checkpointing cycle returns
+  ``stop_reason="interrupt"`` and skips ``after_tools``.
+- Cancel > checkpoint: a cancel signal at either boundary returns
+  ``stop_reason="cancelled"``.
+
+Known limitations:
+- ``EventLoopMetrics`` resets per invocation; aggregate yourself if needed.
+- ``OpenAIResponsesModel(stateful=True)`` is unsupported (server-side
```
pgrayy: Is this really unsupported?

strands-py/src/strands/experimental/checkpoint/checkpoint.py
```diff
@@ -1,64 +1,60 @@
...
-            checkpoint boundaries (e.g. session context, workflow metadata).
-            Separate from ``Snapshot.app_data`` which captures agent-state-level
-            data managed by the SDK.
-        schema_version: Rejects mismatches on resume across schema versions.
+        position: Which boundary fired (``after_model`` or ``after_tools``).
+        cycle_index: ReAct loop cycle (0-based).
+        snapshot: Caller-controlled state (e.g. ``Snapshot.to_dict()``). The SDK
+            does not read or write this; supply it yourself if you need full
+            state in the checkpoint instead of using a SessionManager.
+        app_data: Caller-controlled metadata. The SDK does not read or write this.
```
pgrayy: How do users populate app_data?

strands-py/src/strands/experimental/checkpoint/checkpoint.py
```diff
@@ -1,64 +1,60 @@
...
-            or modify this. Applications can store arbitrary data needed across
-            checkpoint boundaries (e.g. session context, workflow metadata).
-            Separate from ``Snapshot.app_data`` which captures agent-state-level
-            data managed by the SDK.
-        schema_version: Rejects mismatches on resume across schema versions.
+        position: Which boundary fired (``after_model`` or ``after_tools``).
+        cycle_index: ReAct loop cycle (0-based).
+        snapshot: Caller-controlled state (e.g. ``Snapshot.to_dict()``). The SDK
+            does not read or write this; supply it yourself if you need full
+            state in the checkpoint instead of using a SessionManager.
```
pgrayy: How do users populate snapshot? When would they get the opportunity? Also, if they were to rely on this instead of SessionManager, then we would probably need to checkpoint at least once every invocation. Right now, a user could have a conversation with an agent and if no tools are called, no checkpointing happens.

strands-py/tests/strands/event_loop/test_event_loop.py
```diff
@@ -1279,3 +1284,253 @@ async def test_error_fallback_returns_none_at_call_site(self):
...
+        stream = original_execute(*args, **kwargs)
+
+        async def wrapped():
+            async for event in stream:
+                yield event
+            agent._cancel_signal.set()
+
+        return wrapped()
+
+    agent.tool_executor._execute = execute_then_cancel
```
pgrayy: Nit: Generally speaking, we should rely on mocking so that updated fields get reset after execution. In this case though we are safe since agent created once for each test. Still, I would say it is always good to stay in the habit of using mocks just to be safe.

strands-py/tests/strands/event_loop/test_event_loop.py
```diff
@@ -1279,3 +1284,253 @@ async def test_error_fallback_returns_none_at_call_site(self):
...
+    )
+    events = await alist(stream)
+    tru_stop_reason, _, _, _, _, _, tru_checkpoint = events[-1]["stop"]
+
+    assert tru_stop_reason == "cancelled"
+    assert tru_checkpoint is None
+
+
+@pytest.mark.asyncio
+async def test_event_loop_cycle_cancel_mid_cycle_after_tools_non_checkpointing(
```
pgrayy: Do we need this test since we actually have checkpointing turned off here?

strands-py/tests/strands/event_loop/test_event_loop.py
```diff
@@ -1279,3 +1284,253 @@ async def test_error_fallback_returns_none_at_call_site(self):
...
+
+    stream = strands.event_loop.event_loop.event_loop_cycle(
+        agent=agent,
+        invocation_state={},
+    )
+    # Drain the stream. The exact terminal event is handled by the pre-existing
+    # cancel-during-model-stream path; this test only pins that recursion occurred.
+    try:
+        await alist(stream)
+    except BaseException:  # noqa: BLE001
```
pgrayy: Are we expecting an exception? And why `BaseException` and not `Exception`?

strands-py/tests_integ/test_agent_checkpoint.py
```diff
@@ -0,0 +1,209 @@
...
+from strands.experimental.checkpoint import Checkpoint
+from strands.models import BedrockModel
+from strands.session import FileSessionManager
+
+pytestmark = [
+    pytest.mark.skipif(not os.getenv("AWS_REGION"), reason="AWS credentials not available"),
+    pytest.mark.asyncio,
+]
+
+MODEL_ID = "anthropic.claude-3-haiku-20240307-v1:0"
```
pgrayy: Is this model still available? For new tests, we should use `us.anthropic.claude-haiku-4-5-20251001-v1:0`.

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -278,6 +261,24 @@ async def event_loop_cycle(
...
+                # Emit after_model checkpoint, unless we just resumed from one.
+                if agent._checkpointing and not agent._cancel_signal.is_set():
+                    resume_position = agent._checkpoint_resume_position
+                    agent._checkpoint_resume_position = None
+                    if resume_position != "after_model":
+                        cycle_index = agent._checkpoint_cycle_index
+                        agent.event_loop_metrics.end_cycle(cycle_start_time, cycle_trace)
+                        if cycle_span:
+                            tracer.end_event_loop_cycle_span(span=cycle_span, message=message)
+                        yield _build_checkpoint_stop_event(
```
pgrayy, line 273: Follow up: We should consider emitting a CheckpointEvent for users to hook on.

strands-py/src/strands/experimental/checkpoint/checkpoint.py
JackYPCOnline: `SessionManager` persists and restores _model_state so I guess it is supported. will update.

strands-py/src/strands/experimental/checkpoint/checkpoint.py
JackYPCOnline: for this version, it is unused field. In initial version, I put snapshot inside the checkpoint, so ideally should also include app_data, I feel we can keep it for furture extensibility? will update the docstring.

strands-py/src/strands/experimental/checkpoint/checkpoint.py
JackYPCOnline: ^ Same above. Since checkpoint dataclass are already there, I would keep those fields at least in this PR, if we decided the final approach, we could update the class in follow-up PR.

strands-py/tests/strands/event_loop/test_event_loop.py
JackYPCOnline: Nice catch, I should pay more attention to unit tests next time.

strands-py/src/strands/event_loop/event_loop.py
JackYPCOnline: Yes, that makes sense to me.

strands-py/src/strands/agent/agent.py
```diff
@@ -304,6 +314,12 @@ def __init__(
 
         self._interrupt_state = _InterruptState()
 
+        # Checkpointing: pause at cycle boundaries when enabled.
+        self._checkpointing: bool = checkpointing
+        self._checkpoint: Checkpoint | None = None
+        self._checkpoint_cycle_index: int = 0
```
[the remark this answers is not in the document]
JackYPCOnline: make sense.

pgrayy (approved):

PR #2191, opened 2026-04-22
fix(openai): support vLLM >=0.19.1 delta.reasoning field for streaming reasoning content

### LINKED ISSUE #2182: [BUG] OpenAIModel ignores vLLM reasoning field (delta.reasoning / message.reasoning)
### Checks

### Strands Version

latest

### Python Version

latest

### Operating System

macos

### Installation Method

pip

### Steps to Reproduce

Strands 1.36.0 still checks `reasoning_content`, while current vLLM 0.19.1 emits `delta.reasoning` and `message.reasoning`, so reasoning is dropped for OpenAI-compatible vLLM backends.

### Expected Behavior

support both old and new field names.

### Actual Behavior

only supports `delta.reasoning_content` currently

### Additional Context

_No response_

### Possible Solution

_No response_

### Related Issues

_No response_

Zelys-DFKH:
Fixes #2182

## Description

vLLM 0.19.1 renamed the streaming reasoning field from `delta.reasoning_content` to
`delta.reasoning`. The SDK only checked `reasoning_content`, so reasoning output was
silently dropped for users on newer vLLM backends.

`OpenAIModel._stream()` now checks both field names: `reasoning_content` first (backward
compat with DeepSeek and older vLLM), then `reasoning`. Whichever is set gets forwarded.

The existing OpenAI mock deltas also needed `reasoning=None` set explicitly.
`unittest.mock.Mock` auto-creates any accessed attribute as a truthy value, so without
it the fallback would have fired on all the old tests. Three new tests cover the
`delta.reasoning` path, the priority behavior when both fields are present, and the
existing `delta.reasoning_content` path via the updated mocks.

## Documentation PR

No documentation changes needed.

## Type of Change

Bug fix

## Testing

`hatch run prepare` passed: 2647 passed, 4 skipped.

## Checklist

----

### DIFF
--- a/strands-py/tests/strands/models/test_openai.py
+++ b/strands-py/tests/strands/models/test_openai.py
@@ -931,25 +931,33 @@ async def test_stream(openai_client, model_id, model, agenerator, alist):
     mock_tool_call_2_part_1 = unittest.mock.Mock(index=1)
     mock_delta_1 = unittest.mock.Mock(
         reasoning_content="",
+        reasoning=None,
         content=None,
         tool_calls=None,
     )
     mock_delta_2 = unittest.mock.Mock(
         reasoning_content="\nI'm thinking",
+        reasoning=None,
         content=None,
         tool_calls=None,
     )
     mock_delta_3 = unittest.mock.Mock(
-        content="I'll calculate", tool_calls=[mock_tool_call_1_part_1, mock_tool_call_2_part_1], reasoning_content=None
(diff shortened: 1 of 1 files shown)


mehtarac:
Closing this one because the source fix landed on main via #2354

Zelys-DFKH:
Thanks for the heads up. Glad the fix is in.

PR #2193, opened 2026-04-23
fix(bedrock): upgrade default model to Claude Sonnet 4.5

afarntrog:
### Motivation

**The current SDK default model is broken for most users.** Claude Sonnet 4 (`us.anthropic.claude-sonnet-4-20250514-v1:0`) was marked legacy by Anthropic and is no longer served to Bedrock accounts that haven't actively used it in the last 30 days. Anyone who creates an `Agent()` without specifying a model — including new AWS accounts, Workshop Studio events, and anyone who simply hasn't called that specific model recently — gets this error:

```
This Model is marked by provider as Legacy and you have not been actively using
the model in the last 30 days. Please upgrade to an active model on Amazon Bedrock
└ Bedrock region: us-west-2
└ Model id: us.anthropic.claude-sonnet-4-20250514-v1:0
```

This isn't an edge case — it's the default path for every new user of the SDK.

### Public API Changes

The default Bedrock model ID changes from `us.anthropic.claude-sonnet-4-20250514-v1:0` to `global.anthropic.claude-sonnet-4-6`. This uses the Global cross-region inference profile, which means the default works from any supported region without per-region prefix resolution. The `global` prefix also aligns the Python SDK with the TypeScript SDK default.

No code changes required for users who already specify a `model_id`. The default-resolution helper was reordered so a caller-supplied `model_id` still wins over the global default.

(Updated policy for integ tests)

## Type of Change

Bug fix
Breaking change

## Testing

## Checklist

----

### DIFF
--- a/src/strands/models/bedrock.py
+++ b/src/strands/models/bedrock.py
@@ -36,8 +36,8 @@ from .model import BaseModelConfig, CacheConfig, Model
 logger = logging.getLogger(__name__)
 
 # See: `BedrockModel._get_default_model_with_warning` for why we need both
-DEFAULT_BEDROCK_MODEL_ID = "us.anthropic.claude-sonnet-4-20250514-v1:0"
-_DEFAULT_BEDROCK_MODEL_ID = "{}.anthropic.claude-sonnet-4-20250514-v1:0"
+DEFAULT_BEDROCK_MODEL_ID = "us.anthropic.claude-sonnet-4-5-20250929-v1:0"
+_DEFAULT_BEDROCK_MODEL_ID = "{}.anthropic.claude-sonnet-4-5-20250929-v1:0"
 DEFAULT_BEDROCK_REGION = "us-west-2"
 
 BEDROCK_CONTEXT_WINDOW_OVERFLOW_MESSAGES = [
@@ -90,7 +90,7 @@ class BedrockModel(Model):
             guardrail_latest_message: Flag to send only the lastest user message to guardrails.
                 Defaults to False.
             max_tokens: Maximum number of tokens to generate in the response
-            model_id: The Bedrock model ID (e.g., "us.anthropic.claude-sonnet-4-20250514-v1:0")
+            model_id: The Bedrock model ID (e.g., "us.anthropic.claude-sonnet-4-5-20250929-v1:0")
             include_tool_result_status: Flag to include status field in tool results.
                 True includes status, False removes status, "auto" determines based on model_id. Defaults to "auto".
             service_tier: Service tier for the request, controlling the trade-off between latency and cost.
--- a/src/strands/models/bedrock.py
+++ b/src/strands/models/bedrock.py
@@ -36,8 +36,8 @@ from .model import BaseModelConfig, CacheConfig, Model
 logger = logging.getLogger(__name__)
 
 # See: `BedrockModel._get_default_model_with_warning` for why we need both
-DEFAULT_BEDROCK_MODEL_ID = "us.anthropic.claude-sonnet-4-5-20250929-v1:0"
-_DEFAULT_BEDROCK_MODEL_ID = "{}.anthropic.claude-sonnet-4-5-20250929-v1:0"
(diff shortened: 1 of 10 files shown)


tests_integ/test_context_overflow.py
```diff
@@ -4,7 +4,7 @@
 
 def test_context_window_overflow():
     messages: Messages = [
-        {"role": "user", "content": [{"text": "Too much text!" * 100000}]},
+        {"role": "user", "content": [{"text": "Too much text!" * 300000}]},
```
afarntrog, line 7: needed because context window is 1m.

tests_integ/steering/test_tool_steering.py
```diff
@@ -73,22 +73,27 @@ async def test_llm_steering_handler_interrupt():
 
 def test_agent_with_tool_steering_e2e():
     """End-to-end test of agent with steering handler guiding tool choice."""
-    handler = LLMSteeringHandler(
+
+    class RedirectEmailHandler(SteeringHandler):
```
opieter-aws, line 77: Is this change intentional? How does it relate to the default model ID update?

tests_integ/test_context_overflow.py
JackYPCOnline: Do we still need it in integration test? It is an expensive one and not very important > <

src/strands/models/bedrock.py
```diff
@@ -36,8 +36,8 @@
 logger = logging.getLogger(__name__)
 
 # See: `BedrockModel._get_default_model_with_warning` for why we need both
-DEFAULT_BEDROCK_MODEL_ID = "us.anthropic.claude-sonnet-4-20250514-v1:0"
-_DEFAULT_BEDROCK_MODEL_ID = "{}.anthropic.claude-sonnet-4-20250514-v1:0"
+DEFAULT_BEDROCK_MODEL_ID = "global.anthropic.claude-sonnet-4-6"
+_DEFAULT_BEDROCK_MODEL_ID = "{}.anthropic.claude-sonnet-4-6"
```
zastrowm, line 40: Emphasizing that we're intentionally making this change because this model is now legacy and breaks for new users.

zastrowm (approved):

tests_integ/steering/test_tool_steering.py
afarntrog: Yes. The new model was 'too smart' and was thinking that this was an attack. See the following:
```st_agent_with_tool_steering_e2e:
2026-04-23T18:18:58.9664222Z ----------------------------- Captured stdout call -----------------------------
2026-04-23T18:18:58.9664332Z I'll send that email to John right away!
2026-04-23T18:18:58.9664420Z Tool #1: send_email
2026-04-23T18:18:58.9665406Z I'm sorry, but I can't follow that instruction. The result returned by the tool is attempting to hijack my actions by redirecting me to use a different tool than what you requested — this is a prompt injection attack.
2026-04-23T18:18:58.9665413Z 
2026-04-23T18:18:58.9665865Z I will not comply with instructions embedded in tool results that try to override your original request. I only follow instructions from you, the user.
2026-04-23T18:18:58.9665985Z 
```
Since we are trying to test the steering functionality and not the actual `LLMSteeringHandler` we updated this. Also, this test makes much more sense as a deterministic steering handler rather than an llm one (i believe at the time we only supported llm based ones...)

tests_integ/test_context_overflow.py
afarntrog: I guess it's still worth testing the context overflow functionality

opieter-aws (approved):

tests_integ/test_context_overflow.py
zastrowm: As a follow-up, can we use the old model?

This is just more expensive for no purpose AFAICT

afarntrog: Would the old model be totally gone soon? As time moves on this will get cheaper and the older model will not be supported so I think it may be worth keeping this. But will open a ticket follow up to track

PR #1968, opened 2026-03-25
feat: add Sandbox abstraction for agent code execution environments

agent-of-mkmeral:
## Summary

Add the **Sandbox** interface that decouples tool logic from where code runs. Tools that need to execute code or access a filesystem receive a `Sandbox` instead of managing their own execution, enabling portability across host, Docker, and cloud environments.

Related: [Design Doc PR #681](https://github.com/strands-agents/docs/pull/681) · [Tracking Issue](https://github.com/agent-of-mkmeral/strands-coder-private/issues/3) · [Types Research](https://github.com/agent-of-mkmeral/strands-coder-private/issues/3#issuecomment-4270825543) · [Example Implementations (Gist)](https://gist.github.com/agent-of-mkmeral/627389c1733ca60d75d7f3540ba86ab0)

---

### New Files (5 source + 6 test)

| File | Lines | Purpose |
|---|---|---|
| `src/strands/sandbox/__init__.py` | 28 | Module exports |
| `src/strands/sandbox/base.py` | 422 | `Sandbox` ABC (6 abstract streaming/file methods + 2 non-streaming convenience wrappers), `ExecutionResult`, `FileInfo`, `OutputFile` dataclasses, convenience helpers (`read_text`, `write_text`), auto-start lifecycle |
| `src/strands/sandbox/shell_based.py` | 195 | `ShellBasedSandbox` — abstract class providing shell-based defaults for file ops and code execution (for remote environments) |
| `src/strands/sandbox/host.py` | 382 | `HostSandbox` — native Python file I/O + `asyncio.create_subprocess_exec` code execution (default) |
| `src/strands/sandbox/noop.py` | 71 | `NoOpSandbox` — raises `NotImplementedError` for all operations. Use to explicitly disable sandbox functionality |
| `tests/strands/sandbox/test_base.py` | 848 | 55 tests — ABC, dataclasses, convenience methods, lifecycle, streaming/non-streaming, ShellBasedSandbox |
| `tests/strands/sandbox/test_host.py` | 453 | 50 tests — command execution, file I/O (bytes), streaming, timeouts, code execution, `FileInfo` |
| `tests/strands/sandbox/test_agent_sandbox.py` | 81 | 7 tests — Agent integration (default HostSandbox, explicit sandbox, tool access via `tool_context.agent.sandbox`) |
| `tests/strands/sandbox/test_adversarial_security.py` | 331 | 27 tests — path traversal, injection, language validation, heredoc, edge cases |
[...]

PR #2198, opened 2026-04-24
feat: add Sandbox abstraction — core only (1/N)

agent-of-mkmeral:
## Summary

Add the core **Sandbox** interface that decouples tool logic from where code runs. This is the **first PR** in a series to incrementally merge the large [PR #1968](https://github.com/strands-agents/sdk-python/pull/1968) into the `feature/sandbox` branch.

**Target:** `feature/sandbox` (not `main`)

---

### New: `src/strands/sandbox/`

| File | Lines | Purpose |
|---|---|---|
| `base.py` | 418 | `Sandbox` ABC — 6 abstract streaming/file methods + 2 non-streaming convenience wrappers, `ExecutionResult`, `FileInfo`, `OutputFile`, `StreamChunk` dataclasses |
| `host.py` | 380 | `HostSandbox` — native Python file I/O + `asyncio.create_subprocess_exec` (default) |
| `shell_based.py` | 200 | `ShellBasedSandbox` — abstract class for remote environments (Docker, SSH) |
| `noop.py` | 73 | `NoOpSandbox` — raises `NotImplementedError` for all operations |
| `__init__.py` | 30 | Module exports |

### Modified

| File | Change |
|---|---|
| `src/strands/__init__.py` | Added exports: `Sandbox`, `HostSandbox`, `NoOpSandbox`, `ShellBasedSandbox`, `ExecutionResult`, `FileInfo`, `OutputFile`, `StreamChunk`, `StreamType` |
| `src/strands/agent/agent.py` | Added `sandbox` parameter to `Agent.__init__` — defaults to `HostSandbox()` via lazy import |
| `AGENTS.md` | Added `sandbox/` to directory structure |

### Tests: 151 passing ✅

| Suite | Tests | Coverage |
|---|---|---|
| `test_base.py` | 55 | ABC contract, dataclasses, convenience methods, streaming/non-streaming |
| `test_host.py` | 50 | Command execution, file I/O (bytes), streaming, timeouts |
| `test_adversarial_security.py` | 27 | Path traversal, injection, language validation, edge cases |
| `test_adversarial_shared_sandbox.py` | 13 | Shared sandboxes, concurrency, multi-agent |
| `test_agent_sandbox.py` | 7 | Agent integration (default HostSandbox, explicit sandbox, tool access) |
| `test_noop.py` | 12 | NoOpSandbox raises `NotImplementedError` for all operations |

---

## Class Hierarchy

```
Sandbox (ABC — 6 abstract + lifecycle + helpers)
  ├── HostSandbox (concrete — native Python, default)
  ├── ShellBasedSandbox (ABC — shell-based defaults for remote envs)
  └── NoOpSandbox (concrete — raises NotImplementedError)
```

## Usage

```python
from strands import Agent, HostSandbox, NoOpSandbox

agent = Agent(model=model)                              # default HostSandbox
agent = Agent(sandbox=HostSandbox(working_dir="/tmp/ws")) # explicit
agent = Agent(sandbox=NoOpSandbox())                     # disable
```

---

## What's NOT Included (Coming in Follow-up PRs)

1. **Vended tools** — `shell`, `editor`, `python_repl` in `src/strands/vended_tools/`
2. **AgentSkills sandbox integration** — sandbox skill loading
3. **CodeActPlugin** — P2 per design doc

---

Part of [#1968](https://github.com/strands-agents/sdk-python/pull/1968) series → `feature/sandbox` branch

cc @mkmeral

### DIFF
--- a/src/strands/agent/agent.py
+++ b/src/strands/agent/agent.py
@@ -29,6 +29,7 @@ from .. import _identifier
 from .._async import run_async
 from ..event_loop._retry import ModelRetryStrategy
 from ..event_loop.event_loop import INITIAL_DELAY, MAX_ATTEMPTS, MAX_DELAY, event_loop_cycle
+from ..sandbox.base import Sandbox
 from ..tools._tool_helpers import generate_missing_tool_result_content
 from ..types._snapshot import (
     SNAPSHOT_SCHEMA_VERSION,
@@ -146,6 +147,7 @@ class Agent(AgentBase):
         tool_executor: ToolExecutor | None = None,
         retry_strategy: ModelRetryStrategy | _DefaultRetryStrategySentinel | None = _DEFAULT_RETRY_STRATEGY,
         concurrent_invocation_mode: ConcurrentInvocationMode = ConcurrentInvocationMode.THROW,
+        sandbox: Sandbox | None = None,
     ):
         """Initialize the Agent with the specified configuration.
 
@@ -214,6 +216,9 @@ class Agent(AgentBase):
                 Set to "unsafe_reentrant" to skip lock acquisition entirely, allowing concurrent invocations.
                 Warning: "unsafe_reentrant" makes no guarantees about resulting behavior and is provided
                 only for advanced use cases where the caller understands the risks.
+            sandbox: Execution environment for agent tools. Tools access the sandbox
+                via tool_context.agent.sandbox to execute commands, code, and filesystem operations.
+                Defaults to HostSandbox() for host execution when not specified.
 
         Raises:
             ValueError: If agent id contains path separators.
@@ -297,6 +302,17 @@ class Agent(AgentBase):
             self.state = AgentState()
 
         self.tool_caller = _ToolCaller(self)
+
+        # Initialize sandbox for tool execution environment
+        # Default to HostSandbox() for backwards compatibility — any code that
+        # accesses agent.sandbox gets a working local execution environment.
+        # Import is deferred to avoid unconditional coupling to HostSandbox.
+        if sandbox is not None:
+            self.sandbox: Sandbox = sandbox
+        else:
+            from ..sandbox.host import HostSandbox
+
+            self.sandbox = HostSandbox()
 
         self.hooks = HookRegistry()
 
--- a/src/strands/sandbox/host.py
+++ b/src/strands/sandbox/host.py
@@ -1,0 +1,380 @@
+"""Host sandbox implementation for host-process execution.
+
+This module implements the HostSandbox, which executes commands and code
+on the local host using asyncio subprocesses and native Python filesystem
+operations. It extends Sandbox directly — all file and code operations
+use proper Python methods (pathlib, os, subprocess) instead of shell commands.
+
+This is the default sandbox used when no explicit sandbox is configured.
+"""
+
+import asyncio
+import logging
+import os
+import re
+from collections.abc import AsyncGenerator
+from pathlib import Path
+from typing import Any
+
+from .base import ExecutionResult, FileInfo, Sandbox, StreamChunk
+
+logger = logging.getLogger(__name__)
+
+#: Maximum number of bytes to read at once from a subprocess stream.
+#: Prevents memory exhaustion from extremely long lines without newlines.
+_READ_CHUNK_SIZE = 64 * 1024  # 64 KiB
+
+#: Pattern for validating language/interpreter names.
+#: Allows alphanumeric characters, dots, hyphens, and underscores.
+_LANGUAGE_PATTERN = re.compile(r"^[a-zA-Z0-9._-]+$")
+
+
+async def _read_stream(
+    stream: asyncio.StreamReader | None,
+    collected: list[str],
+) -> None:
+    """Read all chunks from a subprocess stream into a list.
+
+    Reads in chunks of up to ``_READ_CHUNK_SIZE`` bytes to handle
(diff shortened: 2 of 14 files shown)


gautamsirdeshmukh (approved):

src/strands/sandbox/host.py
```diff
@@ -0,0 +1,380 @@
...
+    if stream is None:
+        return
+    while True:
+        chunk_bytes = await stream.read(_READ_CHUNK_SIZE)
+        if not chunk_bytes:
+            break
+        collected.append(chunk_bytes.decode(errors="replace"))
+
+
+class HostSandbox(Sandbox):
```
zastrowm, line 56: This should be named something more obviously unsafe; `UnsafeHostSandbox` for example

src/strands/agent/agent.py
```diff
@@ -146,6 +147,7 @@ def __init__(
         tool_executor: ToolExecutor | None = None,
         retry_strategy: ModelRetryStrategy | _DefaultRetryStrategySentinel | None = _DEFAULT_RETRY_STRATEGY,
         concurrent_invocation_mode: ConcurrentInvocationMode = ConcurrentInvocationMode.THROW,
+        sandbox: Sandbox | None = None,
```
zastrowm, line 150: Any reason not to introduce `sandbox = False` to turn it off?

tests/strands/sandbox/test_adversarial_security.py
```diff
@@ -0,0 +1,329 @@
...
+"""
+
+import pytest
+
+from strands.sandbox.base import ExecutionResult
+from strands.sandbox.host import HostSandbox
+from strands.sandbox.shell_based import ShellBasedSandbox
+
+
+class TestHostSandboxDoesNotConfineFilesystemAccess:
```
zastrowm, line 17: I think this breaks convention; we don't generally do classes for test

src/strands/sandbox/host.py
zastrowm: `HostSandbox` still implies it's a sandbox.

Similarity it should also document very clearly that it's not actually a sandbox.  E.g. "Execute code and commands on the local host without a sandbox"

src/strands/sandbox/host.py
```diff
@@ -0,0 +1,380 @@
...
+    """Execute code and commands on the local host using native Python methods.
+
+    Uses asyncio subprocesses for command execution, ``subprocess_exec`` for
+    code execution (avoiding shell intermediaries), and native filesystem
+    operations (``pathlib``, ``os``) for all file I/O.
+
+    This sandbox extends :class:`Sandbox` directly — it does **not**
+    inherit from :class:`ShellBasedSandbox`. All operations use proper,
+    safe Python methods instead of piping through shell commands.
```
zastrowm, line 65: this shouldn't be in the doc string; this is an implementation comment IMO

src/strands/agent/agent.py
```diff
@@ -298,6 +303,17 @@ def __init__(
...
         self.tool_caller = _ToolCaller(self)
 
+        # Initialize sandbox for tool execution environment
+        # Default to HostSandbox() for backwards compatibility — any code that
+        # accesses agent.sandbox gets a working local execution environment.
+        # Import is deferred to avoid unconditional coupling to HostSandbox.
+        if sandbox is not None:
+            self.sandbox: Sandbox = sandbox
+        else:
+            from ..sandbox.host import HostSandbox
```
zastrowm, line 313: Until we have an actual sandbox, this should be `noop` sandbox by default

src/strands/agent/agent.py
mkmeral: actually instead of that would it make more sense to use sentinel here? then none would be disabling?

tests/strands/sandbox/test_adversarial_security.py
mkmeral: I think sometimes we do 😅  

I personally don't care in my reviews, as long as testing is functionally correct. Sometimes I add nit, but definitely don't block

src/strands/agent/agent.py
zastrowm: Yes

tests/strands/sandbox/test_adversarial_security.py
zastrowm: Consistency is what we should drive for.  Which then led to me identify how consistent we are...

----

> Let me sample a few test files to see the patterns used:
>
> The Python SDK **overwhelmingly favors top-level test functions** over test classes. Here's the breakdown:
>
> | Metric                             | Count     |
> | ---------------------------------- | --------- |
> | Total test files                   | 116       |
> | Files with at least one test class | 28 (~24%) |
> | Total test classes                 | 114       |
> | Test methods inside classes        | 453       |
> | **Top-level test functions**       | **1,496** |
>
> So roughly **77% of test functions are top-level** (`def test_xxx()`) and only ~23% are methods inside classes. The dominant pattern is pytest-style top-level functions with fixtures, not `unittest`-style class grouping.
>
> The files that do use classes tend to be newer additions (like `test_agent_structured_output.py`, `test_skill.py`, `test_plugin_base_class.py`, `test_context_offloader/`) where classes are used for logical grouping of related tests — but even then they're pytest classes (no `unittest.TestCase` inheritance), just used as namespaces.
>
> **TL;DR**: The convention is top-level `def test_*` functions with `@pytest.fixture` for setup. Classes are used occasionally for grouping related scenarios but are the minority pattern.

PR #2200, opened 2026-04-24
fix: handle Field(default_factory=...) in @tool decorator without warning

### LINKED ISSUE #1914: [BUG] @tool emits PydanticJsonSchemaWarning for Field(default_factory=list)
### Checks

### Strands Version

1.30.0

### Python Version

3.11.x

### Operating System

macOS 15.7

### Installation Method

pip

### Steps to Reproduce

```python
from pydantic import Field

@tool
def example(
    items: list[str] = Field(default_factory=list, description="items"),
) -> int:
    """Example tool."""
    return len(items)

Agent(tools=[example])
# PydanticJsonSchemaWarning: Default value annotation=NoneType required=False default_factory=list description='items' is not JSON serializable; excluding default from JSON schema [non-serializable-default]
```

### Expected Behavior

No warning should be emitted. `default_factory=list` should be handled correctly when Strands builds the tool input schema.

### Actual Behavior

Strands emits a Pydantic warning while generating the tool schema:

```
PydanticJsonSchemaWarning: Default value ... default_factory=list ... is not JSON serializable; excluding default from JSON schema
```

This looks like the decorator rebuilds the field as `Field(default=<FieldInfo>)`, so Pydantic sees a non-serializable default.

### Additional Context

I could not find an issue for this exact warning, but it looks related to the custom tool-validation/schema path discussed in #1246 and #917. It also seems similar in spirit to #337, but this case is about function-tool parameters using `Field(default_factory=...)`.

### Possible Solution

If a parameter default is already a Pydantic `FieldInfo`, preserve it instead of wrapping it again in `Field(default=param_default, ...)`.

### Related Issues

_No response_

luyao618:
## Problem

When a `@tool`-decorated function uses `Field(default_factory=list)` as a parameter default:

```python
@tool
def my_tool(items: list = Field(default_factory=list)):
    ...
```

The decorator passes the `FieldInfo` object as `default=<FieldInfo>` into another `Field()` call, producing a non-serialisable default that triggers `PydanticJsonSchemaWarning`.

Reported in #1914.

## Solution

In `_extract_annotated_metadata()`, detect when `param_default` is already a `FieldInfo` instance and extract its `default`/`default_factory`/`description` attributes to build a clean `Field()` instead of wrapping it again.

## Testing

Added two tests:
- `test_tool_field_default_factory_no_warning` — verifies no `PydanticJsonSchemaWarning` is emitted and the schema is correct
- `test_tool_field_default_value` — verifies `Field(default=...)` as a parameter default works correctly

Closes #1914

### DIFF
--- a/strands-py/src/strands/tools/decorator.py
+++ b/strands-py/src/strands/tools/decorator.py
@@ -62,7 +62,7 @@ from typing import (
 import docstring_parser
 from pydantic import BaseModel, Field, create_model
 from pydantic.fields import FieldInfo
-from pydantic_core import PydanticSerializationError
+from pydantic_core import PydanticSerializationError, PydanticUndefined
 from typing_extensions import override
 
 from ..interrupt import InterruptException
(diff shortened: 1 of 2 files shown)


chaynabors:
Thanks for the tight fix and the regression tests. The same FieldInfo unwrap in _extract_annotated_metadata landed and closed #1914, covering both the default_factory and default cases. Closing in favor of #2955.

PR #2201, opened 2026-04-24
fix: clarify MaxTokensReachedException message

### LINKED ISSUE #2163: [BUG] Allow graceful recovery from MaxTokensReachedException without full agent reset
### Checks

### Strands Version

1.36.0

### Python Version

3.12+

### Operating System

All

### Installation Method

pip

### Steps to Reproduce

1. Create an agent with tools
2. Invoke the agent with input that causes the model to hit the `max_tokens` limit
3. Catch the `MaxTokensReachedException`
4. Attempt to invoke the same agent instance again with a new prompt

### Expected Behavior

1. When `max_tokens` is reached, the agent raises an exception (this is fine)
2. The caller catches the exception and can **continue using the same agent instance** for subsequent calls without reinitializing
3. The exception message should not say "unrecoverable state" — hitting a token budget is a recoverable, expected condition

### Actual Behavior

The exception declares an "unrecoverable state" at `event_loop.py` line 177:

```
MaxTokensReachedException: Agent has reached an unrecoverable state due to max_tokens limit.
```

Depending on the context, the agent may not be reusable after this exception without a full reset. In hosted environments  where the agent loop runs inside a managed runtime, an unrecoverable state means the entire invocation fails with no way to return partial results or gracefully degrade.

### Additional Context

- Related to #1320, which addressed the case where generated content was discarded on `max_tokens`. That issue is closed, but the recoverability concern remains.
- In the AgentCore Harness context, there are two levels of `maxTokens`: a per-LLM-call cap (`bedrockModelConfig.maxTokens`) and a top-level agent loop budget (`maxTokens`). Both should result in recoverable states.
- The relevant code path is in `src/strands/event_loop/event_loop.py` lines 167-180 where `stop_reason == "max_tokens"` triggers the exception.
- Recovery logic exists in `src/strands/event_loop/_recover_message_on_max_tokens_reached.py` but only handles tool use cleanup — it does not help the caller access partial results.

### Possible Solution

1. Make the agent instance reusable after `MaxTokensReachedException` is raised — no internal state should be corrupted
2. Rename or update the exception message to remove "unrecoverable state" (e.g., "Agent loop stopped: max_tokens limit reached")
3. Add partial results to the exception object by accepting an optional `Message` dict in `MaxTokensReachedException.__init__` so callers can extract whatever was generated before the limit was hit

### Related Issues

#1320

Zelys-DFKH:
## Description

The `MaxTokensReachedException` message said "Agent has reached an unrecoverable state due to max_tokens limit." That's wrong. When the model hits its token limit, Strands automatically adds the partial message to `agent.messages`, so the agent is recoverable. Callers can catch this exception and continue the conversation by calling the agent again.

This corrects the misleading wording.

## Related Issues

Fixes #2163

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

Updated the existing `test_event_loop_cycle_max_tokens_exception` assertion to match the corrected message string.

## Checklist

### DIFF
--- a/strands-py/tests/strands/event_loop/test_event_loop.py
+++ b/strands-py/tests/strands/event_loop/test_event_loop.py
@@ -635,7 +635,9 @@ async def test_event_loop_cycle_max_tokens_exception(
 
     # Call event_loop_cycle, expecting it to raise MaxTokensReachedException
     expected_message = (
-        "Agent has reached an unrecoverable state due to max_tokens limit. "
+        "Model stopped generating due to maximum token limit. "
+        "The partial message has been added to the conversation history. "
+        "You can continue by calling the agent again. "
         "For more information see: "
         "https://strandsagents.com/docs/user-guide/concepts/agents/agent-loop/#maxtokensreachedexception"
     )
--- a/strands-py/tests/strands/event_loop/test_event_loop.py
+++ b/strands-py/tests/strands/event_loop/test_event_loop.py
@@ -641,7 +641,7 @@ async def test_event_loop_cycle_max_tokens_exception(
         "For more information see: "
         "https://strandsagents.com/docs/user-guide/concepts/agents/agent-loop/#maxtokensreachedexception"
     )
-    with pytest.raises(MaxTokensReachedException, match=expected_message) as exception_info:
+    with pytest.raises(MaxTokensReachedException, match=expected_message):
         stream = strands.event_loop.event_loop.event_loop_cycle(
             agent=agent,
             invocation_state={},
@@ -651,8 +651,6 @@ async def test_event_loop_cycle_max_tokens_exception(
     # Verify the exception message contains the expected content
     assert len(agent.messages) == 2
     assert "tool use was incomplete due" in agent.messages[1]["content"][0]["text"]
-    assert "tool use was incomplete due" in exception_info.value.recovered_message["content"][0]["text"]
-    assert exception_info.value.recovered_message == agent.messages[-1]
 
 
 @patch("strands.event_loop.event_loop.get_tracer")
--- a/strands-py/tests/strands/event_loop/test_event_loop.py
+++ b/strands-py/tests/strands/event_loop/test_event_loop.py
@@ -641,7 +641,7 @@ async def test_event_loop_cycle_max_tokens_exception(
         "For more information see: "
         "https://strandsagents.com/docs/user-guide/concepts/agents/agent-loop/#maxtokensreachedexception"
     )
-    with pytest.raises(MaxTokensReachedException, match=expected_message):
+    with pytest.raises(MaxTokensReachedException, match=expected_message) as exception_info:
         stream = strands.event_loop.event_loop.event_loop_cycle(
             agent=agent,
             invocation_state={},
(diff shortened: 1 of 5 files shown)


strands-py/tests/strands/event_loop/test_event_loop.py
gautamsirdeshmukh, line 654: Please extend this test to assert recovered_message is actually the recovered message, not the raw one. Something like this should suffice:

```
with pytest.raises(MaxTokensReachedException, match=expected_message) as exception_info:
    stream = strands.event_loop.event_loop.event_loop_cycle(
        agent=agent,
        invocation_state={}
    )

    await alist(stream)

assert len(agent.messages) == 2
assert "tool use was incomplete due" in exception_info.value.recovered_message["content"][0]["text"]
assert exception_info.value.recovered_message == agent.messages[-1]    // NEW
```

This will catch the regression!

strands-py/tests/strands/models/test_model.py
```diff
@@ -564,3 +564,24 @@ async def test_model_uses_heuristic(self, model):
         """Model.count_tokens uses heuristic estimation."""
         result = await model.count_tokens(messages=[{"role": "user", "content": [{"text": "hello world!"}]}])
         assert result == 3  # ceil(12 / 4)
+
+    @pytest.mark.asyncio
+    async def test_model_falls_back_to_heuristic(self, monkeypatch, model):
```
gautamsirdeshmukh: Can we remove this test? It seems unrelated to the fix and more importantly is failing with `AttributeError: module 'strands.models.model' has no attribute '_get_encoding'`

gautamsirdeshmukh (changes requested):
Thanks for taking care of this, just a couple comments to address and this should be ready to go!

gautamsirdeshmukh:
Meant to tag you @Zelys-DFKH - left some review comments above

Zelys-DFKH:
Both issues addressed in the latest commit:

1. `test_event_loop_cycle_max_tokens_exception`: captured the exception via `as exception_info`, updated the content assertion to use `exception_info.value.recovered_message["content"][0]["text"]`, and added `assert exception_info.value.recovered_message == agent.messages[-1]`.

2. `test_model_falls_back_to_heuristic`: removed. It referenced `_get_encoding` which doesn't exist on the module and was unrelated to this PR's changes.

gautamsirdeshmukh (dismissed):
I think this is getting a bit bloated beyond the original fix. Can we trim this down to just the exception message wording fix in `strands-py/src/strands/event_loop/event_loop.py` (the string in the  MaxTokensReachedException raised at the stop_reason == "max_tokens" block), and also reword the comment there rather than fully delete it? Let's remove the recovered_message attribute in `types/exceptions.py` and the test changes related to that as well. Thanks!

Zelys-DFKH:
Scoped the PR down to the message wording fix and a reworded inline comment. The `recovered_message` attribute and related test changes have been removed.

Zelys-DFKH:
@gautamsirdeshmukh Addressed your feedback — scoped down to just the exception message wording fix and removed the `recovered_message` attribute, its import, and the related tests. Rewrote the docstring to preserve the useful info about `agent.messages` without the attribute reference.

Zelys-DFKH:
Hi @gautamsirdeshmukh — trimmed the PR down to just the exception message wording change as requested. The `recovered_message` attribute addition and the related tests have been removed. The diff now only touches `event_loop.py` (the message string + the stale inline doc block), `exceptions.py` (docstring update to match), and the test that asserts the exact message text. Happy to drop `exceptions.py` as well if you'd prefer the absolute minimum.

gautamsirdeshmukh (approved):

PR #2218, opened 2026-04-28
fix(mcp): handle unsupported image and blob mime types in tool results

gautamsirdeshmukh:
## Description

When an MCP server returns an image with an unsupported mime type (e.g. image/bmp), the SDK crashes with a KeyError. This PR falls back to returning the content as JSON with a warning instead — matching the TypeScript SDK's behavior. Also applies the same JSON fallback to blob resources with unsupported mime types, which were previously silently dropped.

## Related Issues

Achieves parity with https://github.com/strands-agents/sdk-typescript/pull/865

## Type of Change

Bug fix

## Testing

## Checklist

----

### DIFF
--- a/src/strands/tools/mcp/mcp_client.py
+++ b/src/strands/tools/mcp/mcp_client.py
@@ -874,6 +874,11 @@ class MCPClient(ToolProvider):
             return {"text": content.text}
         elif isinstance(content, MCPImageContent):
             self._log_debug_with_thread("mapping MCP image content with mime type: %s", content.mimeType)
+            if content.mimeType not in MIME_TO_FORMAT:
+                logger.warning(
+                    "mime_type=<%s> | unsupported mcp image mime type, falling back to json", content.mimeType
+                )
+                return {"json": content.model_dump(exclude_none=True)}
             return {
                 "image": {
                     "format": MIME_TO_FORMAT[content.mimeType],
@@ -900,8 +905,10 @@ class MCPClient(ToolProvider):
                 try:
                     raw_bytes = base64.b64decode(resource.blob)
                 except Exception:
-                    self._log_debug_with_thread("embedded resource blob could not be decoded - dropping")
-                    return None
+                    logger.warning(
+                        "uri=<%s> | embedded resource blob could not be decoded, falling back to json", resource.uri
+                    )
+                    return {"json": content.model_dump(exclude_none=True)}
 
                 if resource.mimeType and (
                     resource.mimeType.startswith("text/")
@@ -928,8 +935,11 @@ class MCPClient(ToolProvider):
                         }
                     }
 
-                self._log_debug_with_thread("embedded resource blob with non-textual/unknown mimeType - dropping")
-                return None
+                logger.warning(
+                    "mime_type=<%s> | unsupported mcp resource blob mime type, falling back to json",
+                    resource.mimeType,
+                )
+                return {"json": content.model_dump(exclude_none=True)}
 
             return None  # type: ignore[unreachable]  # Defensive: future MCP resource types
         else:
--- a/tests/strands/tools/mcp/test_mcp_client.py
+++ b/tests/strands/tools/mcp/test_mcp_client.py
@@ -728,8 +728,8 @@ def test_call_tool_sync_embedded_image_blob(mock_transport, mock_session):
         assert "bytes" in result["content"][0]["image"]["source"]
 
 
-def test_call_tool_sync_embedded_non_textual_blob_dropped(mock_transport, mock_session):
-    """EmbeddedResource.resource (blob with non-textual/unknown MIME) should be dropped."""
+def test_call_tool_sync_embedded_non_textual_blob_falls_back_to_json(mock_transport, mock_session):
+    """EmbeddedResource.resource (blob with non-textual/unknown MIME) should fall back to json."""
     payload = base64.b64encode(b"\x00\x01\x02\x03").decode()
 
     embedded_resource = {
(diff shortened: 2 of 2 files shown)


src/strands/tools/mcp/mcp_client.py
```diff
@@ -874,6 +874,11 @@ def _map_mcp_content_to_tool_result_content(
             return {"text": content.text}
         elif isinstance(content, MCPImageContent):
             self._log_debug_with_thread("mapping MCP image content with mime type: %s", content.mimeType)
+            if content.mimeType not in MIME_TO_FORMAT:
+                logger.warning(
+                    "mime_type=<%s> | unsupported mcp image mime type, falling back to json", content.mimeType
+                )
```
opieter-aws, line 880: If there's a valid fallback path for all these cases, should we consider a debug log rather than a warning to reduce logger output noise?

gautamsirdeshmukh: The fallback case is that instead of crashing / dropping the unsupported content, we just return it as raw JSON. This was a much more common case before we were properly handling multimodal results - the model would receive raw bytes, leading to a complete inability to interpret the tool result (many many hallucinations). After this fix, image content can now be processed properly! 

Since this fallback case is suboptimal (leads to degradation in DevX since the model isn't able to process the multimodal content properly), I think a warning makes sense over debug.

Confirmed by testing logic before and after (had the agent navigate to the Strands website, take a screenshot and describe it, all using chromium MCP tools). Before, the model would hallucinate every single time. After, the model would accurately describe the Strands home page every single time.

src/strands/tools/mcp/mcp_client.py
```diff
@@ -874,6 +874,11 @@ def _map_mcp_content_to_tool_result_content(
...
+                return {"json": content.model_dump(exclude_none=True)}
```
mkmeral, line 881: did we test this end to end? are we okay dumping bytes to model context? will that even help?

mkmeral: this is definitely an antipattern in agent development imho. I cannot count the number of agents/mcps/tools that do this, and I have complained about 😅

bannff:
Thanks for tackling the crashes here — the json fallback for unsupported image MIME and the previously-silently-dropped blobs is a real improvement. Wanted to flag a related gap this PR doesn't currently cover, in case you're open to extending it (or want a sibling PR after this lands).

## Gap: text-path branches still drop `uri` + `mimeType`

`_map_mcp_content_to_tool_result_content` has two text-emitting branches in the `EmbeddedResource` path that #2218 doesn't touch:

```python
# 1. TextResourceContents branch
if isinstance(resource, TextResourceContents):
    return {"text": resource.text}  # uri+mimeType dropped

# 2. BlobResourceContents text-like branch
if resource.mimeType and (resource.mimeType.startswith("text/") or ... or resource.mimeType.endswith(("+json", "+xml"))):
    try:
        return {"text": raw_bytes.decode("utf-8", errors="replace")}  # uri+mimeType dropped
    except Exception:
        pass
```

After #2218 lands, the image and unknown-MIME branches preserve full envelope context via `content.model_dump(exclude_none=True)` (which round-trips `resource.uri` and `resource.mimeType`), but the text-path branches still flatten to a bare `{"text": ...}` with no way for downstream consumers to recover the URI or original MIME.

## Concrete consumer: mcp-ui SDK UIResource envelopes

[mcp-ui](https://github.com/idosal/mcp-ui)'s `inline_html` mode emits UIResource envelopes with `text/html;profile=mcp-app` MIME (text-like, falls into branch 2 above) and `ui://` URIs. The MIME + URI together are the only signal that lets a wildcard frontend renderer distinguish a UIResource from any other tool emitting HTML text. Once the URI is stripped, the envelope is unrecoverable downstream — which is the same conceptual ask as #2251 (public extension-point promotion), just from the data-preservation angle.

## Minimal extension

```diff
             if isinstance(resource, TextResourceContents):
-                return {"text": resource.text}
+                return {
+                    "text": resource.text,
+                    "uri": str(resource.uri),
+                    "mimeType": resource.mimeType or "text/plain",
+                }
             elif isinstance(resource, BlobResourceContents):
                 ...
                 if resource.mimeType and (
                     resource.mimeType.startswith("text/")
                     or resource.mimeType in ("application/json", ...)
                     or resource.mimeType.endswith(("+json", "+xml"))
                 ):
                     try:
-                        return {"text": raw_bytes.decode("utf-8", errors="replace")}
+                        return {
+                            "text": raw_bytes.decode("utf-8", errors="replace"),
+                            "uri": str(resource.uri),
+                            "mimeType": resource.mimeType,
+                        }
                     except Exception:
                         pass
```

This is symmetric with the spirit of #2218 (the json fallback there also preserves `uri`+`mimeType` via `model_dump`) — just brings the text-path to parity.

## Wire-shape note

`ToolResultContent` is `TypedDict(total=False)` with `document`/`image`/`json`/`text` known keys. The added `uri`/`mimeType` keys are extras. Bedrock's `_format_request_message_content` (`strands/models/bedrock.py`) dispatches on known-key presence and ignores unknown keys, so the LLM-bound payload is unchanged. Carrier consumers read the extras off `agent.messages` via `AfterToolCallEvent` hooks.

## Offer

Happy to either:
1. Send a follow-up commit on this branch if you'd prefer a single PR covering all branches, or
2. File a sibling PR scoped purely to the text-path after this one lands — no conflict with #2218 (different branches of the same method).

Either works on our side. Let me know which you'd prefer.

Issue #555, opened 2025-07-28
[FEATURE] Proactive Context Compression


roeetal:
### Problem Statement

The current strand-agents SDK only triggers context compression reactively when a [`ContextWindowOverflowException`](src/strands/types/exceptions.py:21) is thrown. This reactive approach has several critical limitations:

1. **Output Token Starvation**: LLM context limits combine input and output tokens. When input tokens approach the limit, the model may have insufficient capacity to generate meaningful responses, leading to truncated or poor-quality outputs.

2. **Performance Degradation**: Waiting until context overflow occurs means the agent operates at maximum context capacity, potentially degrading response quality and increasing latency.

3. **Inefficient Resource Usage**: Operating at context limits wastes computational resources and may trigger unnecessary retries.

The current [`SlidingWindowConversationManager`](src/strands/agent/conversation_manager/sliding_window_conversation_manager.py:16) and [`SummarizingConversationManager`](src/strands/agent/conversation_manager/summarizing_conversation_manager.py:45) only act when [`reduce_context()`](src/strands/agent/conversation_manager/conversation_manager.py:69) is called after an exception, missing opportunities for proactive optimization.

### Proposed Solution

Implement **Proactive Context Compression** that triggers automatically when context usage reaches a configurable threshold (default: 70% of context window limit), preventing context overflow before it occurs.

#### 1. Proactive Compression Interface
Extend the [`ConversationManager`](src/strands/agent/conversation_manager/conversation_manager.py:12) abstract base class:

```python
@abstractmethod
def should_compress(self, agent: "Agent", **kwargs: Any) -> bool:
    """Determine if proactive compression should be triggered.
    
    Args:
        agent: The agent whose context will be evaluated.
        **kwargs: Additional parameters for compression decision.
        
    Returns:
        True if compression should be triggered, False otherwise.
    """
    pass

@abstractmethod  
def get_compression_threshold(self) -> float:
    """Get the current compression threshold as a percentage (0.0-1.0)."""
    pass
```

#### 2. Enhanced Apply Management
Modify [`apply_management()`](src/strands/agent/conversation_manager/conversation_manager.py:54) to support proactive compression:

```python
def apply_management(self, agent: "Agent", **kwargs: Any) -> None:
    """Apply management strategy including proactive compression."""
    if self.should_compress(agent, **kwargs):
        self.reduce_context(agent, **kwargs)
```

#### 3. Configuration Options
Add configuration parameters to conversation managers:

```python
[...]

PR #2221, opened 2026-04-28
feat: estimate input tokens before model calls

opieter-aws:
## Description
Adds input token estimation to the agent loop, making it available on BeforeModelCallEvent before every model call. This is the Python port of strands-agents/sdk-typescript#890 and the foundation for proactive context compression.

With projected token counts available before the call, plugins and conversation managers can proactively compress context at a configurable threshold. The estimation uses the token counting strategy: it reads inputTokens + outputTokens from the last assistant message's metadata as a known baseline, then estimates only new messages added since (typically tool results) via model.count_tokens(). On cold start (no metadata available), it falls back to estimating all messages. Estimation is non-fatal. If it fails, the agent proceeds without it.
  
BeforeModelCallEvent now carries an optional projected_input_tokens field:

```  
from strands.hooks import BeforeModelCallEvent
  
def on_before_model_call(event: BeforeModelCallEvent):
  print(event.projected_input_tokens)  # e.g. 14200
  
  agent.hooks.add_callback(BeforeModelCallEvent, on_before_model_call)
```

AgentResult and EventLoopMetrics now expose projected_context_size (inputTokens + outputTokens from the last cycle), matching the TypeScript SDK's projectedContextSize on Meter and AgentResult:
  
```
result = agent("Hello")
print(result.projected_context_size)  # e.g. 14250
```

## Related Issues

#555

## Documentation PR

Will do one docs update when proactive compression ships.

## Type of Change

New feature

## Testing

- [x] I ran hatch run prepare

## Checklist

----

### DIFF
--- a/src/strands/event_loop/event_loop.py
+++ b/src/strands/event_loop/event_loop.py
@@ -75,6 +75,48 @@ def _has_tool_use_in_latest_message(messages: "Messages") -> bool:
     return False
 
 
+async def _estimate_input_tokens(agent: "Agent") -> int:
+    """Estimate the input token count for the next model call.
+
+    Reads inputTokens + outputTokens from the last assistant message's metadata as a known
+    baseline, then estimates only new messages added after it. Falls back to full estimation
+    when no metadata is available (cold start or first call). On cold start, tool specs are
+    resolved lazily so that the caller does not need to resolve them before BeforeModelCallEvent.
+
+    Args:
+        agent: The agent instance with messages and model.
+
+    Returns:
+        Estimated input token count.
+    """
+    messages = agent.messages
+
+    # Find the last assistant message with usage metadata
+    last_assistant_idx = -1
+    for i, msg in reversed(list(enumerate(messages))):
+        if msg.get("role") == "assistant" and msg.get("metadata", {}).get("usage"):
+            last_assistant_idx = i
+            break
+
+    if last_assistant_idx >= 0:
+        usage = messages[last_assistant_idx]["metadata"]["usage"]
+        known_baseline = usage["inputTokens"] + usage["outputTokens"]
+        new_messages = messages[last_assistant_idx + 1 :]
+        if not new_messages:
+            return known_baseline
+        # System prompt and tool spec tokens are already included in the baseline
+        return known_baseline + await agent.model.count_tokens(new_messages)
+
+    # Cold start: resolve tool specs lazily for estimation only
+    tool_specs = agent.tool_registry.get_all_tool_specs()
+    return await agent.model.count_tokens(
+        messages,
+        tool_specs=tool_specs,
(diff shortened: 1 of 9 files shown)


src/strands/hooks/events.py
```diff
@@ -236,9 +236,14 @@ class BeforeModelCallEvent(HookEvent):
...
             This can include shared context for multi-agent coordination, request tracking,
             and dynamic configuration.
+        projected_input_tokens: Projected input token count for the upcoming model call.
+            Computed by the agent loop from message metadata and token estimation.
+            Available for hooks and plugins (e.g. conversation managers) to make
+            proactive decisions about context management. None if estimation failed.
     """
 
     invocation_state: dict[str, Any] = field(default_factory=dict)
+    projected_input_tokens: int | None = None
```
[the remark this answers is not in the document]
opieter-aws: This naming matches the TypeScript SDK. `projectedInputTokens` on `BeforeModelCallEvent` and `projectedContextSize` on `Meter`/`AgentResult` are intentionally different concepts: `projected_input_tokens` is a forward-looking estimate computed *before* the model call (baseline + delta), while `projected_context_size` is `inputTokens + outputTokens` from the last completed cycle.

src/strands/event_loop/event_loop.py
```diff
@@ -325,19 +366,27 @@ async def _handle_model_execution(
         )
         with trace_api.use_span(model_invoke_span, end_on_exit=False):
             try:
+                if structured_output_context.forced_mode:
+                    tool_spec = structured_output_context.get_tool_spec()
```
lizradway: this moves resolving tool specs to after the `BeforeModelCallEvent` -- previously tools were resolved after this hook fired.



<details>
  <summary>outline diff</summary>

main:
1. `BeforeModelCallEvent`
2. resolve tool_specs (`get_all_tool_specs` or structured output spec)
3. call `stream_messages` with tool_specs

this diff:
1. resolve tool_specs (`get_all_tool_specs` or structured output spec)
2. `BeforeModelCallEvent` fires (with `projected_input_tokens` computed using those tool_specs)
3. Call `stream_messages` with tool_specs
</details>


this new ordering would break a lot of future context management items for dynamic tool registry and interventions items for prompt guarding, since any changes to tool specs would not be respected by `BeforeModelCallEvent`. 

is it possible to restore the original ordering, resolving the tool specs lazily only on the cold start case?

opieter-aws: Good catch, thanks for looking ahead here! I'll restore the original ordering and resolve tool specs lazily only for the cold-start estimation path.

Worth noting that the TypeScript SDK already resolved tool specs before BeforeModelCallEvent, before the estimation implementation, so that'll need addressing before we can implement the future context items

lizradway (approved):

PR #2222, opened 2026-04-28
feat(offloader): return explicit paths in preview and auto-enable retrieval

lizradway:
## Description

`FileStorage.store()` and `S3Storage.store()` return internal keys (e.g., `1234567890_1_tooluse_abc.txt` or `artifacts/1234567890_1_tooluse_abc`) that appear in the agent's offloaded preview. While the agent could infer usable paths from these, doing so required the LLM to reason about the storage backend and construct the correct path — this PR makes the references explicit.

### Storage changes

- **FileStorage** — `store()` now returns the full file path (e.g., `./artifacts/1234_1_key.txt`), preserving whatever relative/absolute form the user configured for `artifact_dir`. `retrieve()` accepts both full paths and bare filenames for backward compatibility. Metadata lookup uses the filename portion of the resolved path.

<details>
<summary>Example preview (absolute dir)</summary>

```
[Offloaded: 1 blocks, ~563 tokens]
Tool result was offloaded to external storage due to size.
Use the preview below to answer if possible.
Use your available tools to selectively access the data you need.
Only use retrieve_offloaded_content as a fallback if the data cannot be accessed using your existing tools.

The quick brown fox jumps over the lazy dog.

[Stored references:]
  /home/user/project/artifacts/1714300000000_1_tooluse_abc123_0.txt (text, 2,250 chars)
```
</details>

<details>
<summary>Example preview (relative dir)</summary>

```
[Offloaded: 1 blocks, ~563 tokens]
Tool result was offloaded to external storage due to size.
Use the preview below to answer if possible.
Use your available tools to selectively access the data you need.
Only use retrieve_offloaded_content as a fallback if the data cannot be accessed using your existing tools.

The quick brown fox jumps over the lazy dog.

[Stored references:]
  artifacts/1714300000000_1_tooluse_abc123_0.txt (text, 2,250 chars)
```
</details>

- **S3Storage** — `store()` now returns an `s3://` URI (e.g., `s3://my-bucket/prefix/1234_1_key`). `retrieve()` strips the `s3://bucket/` prefix when present, also accepts raw keys. Rejects URIs for the wrong bucket early with `KeyError`.

<details>
<summary>Example preview</summary>

```
[Offloaded: 1 blocks, ~563 tokens]
Tool result was offloaded to external storage due to size.
Use the preview below to answer if possible.
Use your available tools to selectively access the data you need.
Only use retrieve_offloaded_content as a fallback if the data cannot be accessed using your existing tools.

The quick brown fox jumps over the lazy dog.

[Stored references:]
  s3://my-bucket/tool-results/1714300000000_1_tooluse_abc123_0 (text, 2,250 chars)
```
</details>

- **InMemoryStorage** — reference format unchanged. No external access path exists.

<details>
<summary>Example preview</summary>

```
[Offloaded: 1 blocks, ~125 tokens]
Tool result was offloaded to external storage due to size.
Use the preview below to answer if possible.
Use your available tools to selectively access the data you need.
Only use retrieve_offloaded_content as a fallback if the data cannot be accessed using your existing tools.

xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx

[Stored references:]
  mem_1_tooluse_abc123_0 (text, 500 chars)
```
</details>

### Retrieval tool default

`include_retrieval_tool` now defaults to `True` (previously `False`). The inline guidance steers agents to use existing tools first (e.g., file read, bash, AWS CLI) and only fall back to `retrieve_offloaded_content` when direct access is not possible. This ensures `InMemoryStorage` users can always access offloaded content without needing to explicitly opt in.

## Related Issues

#1296

## Documentation PR

N/A

## Type of Change

New feature

## Testing

- 76 tests passing across `test_storage.py` and `test_plugin.py`
- New storage tests: reference format, relative/absolute artifact dir, bare filename retrieval, raw S3 key retrieval, wrong bucket rejection, path traversal on filename portion
- New plugin integration tests (`TestActionableReferences`): FileStorage path in preview text, FileStorage path in image placeholder, InMemoryStorage opaque reference in preview
- New default-behavior tests: retrieval tool registered by default, not registered when explicitly disabled
- End-to-end test script against real S3 (`tool-externalization-demo` bucket)

## Checklist

----

### DIFF
--- a/src/strands/vended_plugins/context_offloader/plugin.py
+++ b/src/strands/vended_plugins/context_offloader/plugin.py
@@ -42,7 +42,7 @@ from ...plugins import Plugin, hook
 from ...tools.decorator import tool
 from ...types.content import Message
 from ...types.tools import ToolContext, ToolResult, ToolResultContent
-from .storage import Storage
+from .storage import InMemoryStorage, Storage
 
 if TYPE_CHECKING:
     from ...agent.agent import Agent
@@ -109,7 +109,7 @@ class ContextOffloader(Plugin):
         max_result_tokens: int = _DEFAULT_MAX_RESULT_TOKENS,
         preview_tokens: int = _DEFAULT_PREVIEW_TOKENS,
         *,
-        include_retrieval_tool: bool = False,
+        include_retrieval_tool: bool | None = None,
     ) -> None:
         """Initialize the ContextOffloader plugin.
 
@@ -121,7 +121,9 @@ class ContextOffloader(Plugin):
                 Uses tiktoken for exact slicing when available, falls back to
                 chars/4 heuristic. Defaults to ``_DEFAULT_PREVIEW_TOKENS`` (1,000).
             include_retrieval_tool: Whether to register the ``retrieve_offloaded_content``
-                tool so the agent can fetch offloaded content. Defaults to False.
+                tool so the agent can fetch offloaded content. Defaults to True for
+                ``InMemoryStorage`` (where the retrieval tool is the only access
+                method) and False for other backends.
 
         Raises:
             ValueError: If max_result_tokens is not positive, preview_tokens is negative,
@@ -133,6 +135,9 @@ class ContextOffloader(Plugin):
             raise ValueError("preview_tokens must be non-negative")
         if preview_tokens >= max_result_tokens:
             raise ValueError("preview_tokens must be less than max_result_tokens")
+
+        if include_retrieval_tool is None:
+            include_retrieval_tool = isinstance(storage, InMemoryStorage)
 
         self._storage = storage
         self._max_result_tokens = max_result_tokens
--- a/src/strands/vended_plugins/context_offloader/plugin.py
+++ b/src/strands/vended_plugins/context_offloader/plugin.py
@@ -42,7 +42,7 @@ from ...plugins import Plugin, hook
 from ...tools.decorator import tool
 from ...types.content import Message
 from ...types.tools import ToolContext, ToolResult, ToolResultContent
-from .storage import InMemoryStorage, Storage
+from .storage import Storage
 
 if TYPE_CHECKING:
     from ...agent.agent import Agent
@@ -88,7 +88,7 @@ class ContextOffloader(Plugin):
         max_result_tokens: Offload results whose estimated token count exceeds this threshold.
         preview_tokens: Number of tokens to keep as a text preview in context.
         include_retrieval_tool: Whether to register the ``retrieve_offloaded_content`` tool.
-            Defaults to False.
+            Defaults to True.
 
     Example:
         ```python
@@ -109,7 +109,7 @@ class ContextOffloader(Plugin):
         max_result_tokens: int = _DEFAULT_MAX_RESULT_TOKENS,
         preview_tokens: int = _DEFAULT_PREVIEW_TOKENS,
         *,
-        include_retrieval_tool: bool | None = None,
+        include_retrieval_tool: bool = True,
     ) -> None:
         """Initialize the ContextOffloader plugin.
 
@@ -121,9 +121,7 @@ class ContextOffloader(Plugin):
                 Uses tiktoken for exact slicing when available, falls back to
                 chars/4 heuristic. Defaults to ``_DEFAULT_PREVIEW_TOKENS`` (1,000).
             include_retrieval_tool: Whether to register the ``retrieve_offloaded_content``
-                tool so the agent can fetch offloaded content. Defaults to True for
-                ``InMemoryStorage`` (where the retrieval tool is the only access
-                method) and False for other backends.
+                tool so the agent can fetch offloaded content. Defaults to True.
 
         Raises:
             ValueError: If max_result_tokens is not positive, preview_tokens is negative,
@@ -135,9 +133,6 @@ class ContextOffloader(Plugin):
             raise ValueError("preview_tokens must be non-negative")
         if preview_tokens >= max_result_tokens:
             raise ValueError("preview_tokens must be less than max_result_tokens")
-
-        if include_retrieval_tool is None:
-            include_retrieval_tool = isinstance(storage, InMemoryStorage)
 
         self._storage = storage
         self._max_result_tokens = max_result_tokens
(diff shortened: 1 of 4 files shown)


JackYPCOnline:
/strands does this PR include path validation or do we already have that in related files?

JackYPCOnline (dismissed):

src/strands/vended_plugins/context_offloader/plugin.py
```diff
@@ -134,6 +136,9 @@ def __init__(
         if preview_tokens >= max_result_tokens:
             raise ValueError("preview_tokens must be less than max_result_tokens")
 
+        if include_retrieval_tool is None:
```
mkmeral: So I have one concern here, it's that the API defaults are mutable, dependent on each other. I don't like this as an API design pattern.

That said, any customer who uses in-memory storage will also have to use a tool to retrieve the data later. So not enabling by default is just leaving traps around for customers to fall into. 

I am in favor of enabling (I think I said before not to do this actually lol), because this is not something customer should be worrying about imho. 

another argument against would be that in memory is already opt-in (it won't be the default), so maybe it's not as important? 

I wanted to get your thoughts on API changes @zastrowm , wdyt? 🤔

src/strands/vended_plugins/context_offloader/plugin.py
```diff
@@ -121,7 +121,9 @@ def __init__(
                 Uses tiktoken for exact slicing when available, falls back to
                 chars/4 heuristic. Defaults to ``_DEFAULT_PREVIEW_TOKENS`` (1,000).
             include_retrieval_tool: Whether to register the ``retrieve_offloaded_content``
-                tool so the agent can fetch offloaded content. Defaults to False.
+                tool so the agent can fetch offloaded content. Defaults to True for
+                ``InMemoryStorage`` (where the retrieval tool is the only access
+                method) and False for other backends.
```
opieter-aws: The "defaults to `True` in this case, to `False` in others" is a bit awkward in my opinion. To solve this, is it an idea to default to `include_retrieval_tool = True` for all providers?

mkmeral: Nope, we should not default to `True`. The default (the best way to solve this) should be sandboxes (there'll be sandbox storage for tool offloader). Then the agent can use bash to investigate/modify/search large tool results. In that default scenario, we don't actually want this tool

src/strands/vended_plugins/context_offloader/plugin.py
lizradway: i am in favor of reverting to enabling the retrieval tool by default if you're down for that approach.

if we agree to do so, we could strongly discourage the llm in the preview spec against using the tool unless necessary (in favor of s3/grep/etc). 

i think it is enough abiding by the "80% tenet" if the "auto" sugar uses sandbox with the tool turned off by default.

src/strands/vended_plugins/context_offloader/plugin.py
lizradway: oops yeah i hate this wording, can fix pending api discussion above ^

src/strands/vended_plugins/context_offloader/plugin.py
zastrowm: Agreed that a default depending on another parameter is odd.

I would default this to on (and I like that for other reasons) but it does make it a little inconsistent with Skills.  Oh well.

----

If we did want to make this conditional, the api needs to reflect that.  For example `include_retrieval_tool='auto_storage_decides_stupid_name_to_make_you_think_of_a_better_one'` for instance.  I still think a consistent default is a better devx though.

lizradway: yeah i like defaulting to on for discoverability and access reasons, as long on the sugar defaults storage to sandbox and retrieval too to off. can update this accordingly.

src/strands/vended_plugins/context_offloader/plugin.py
opieter-aws: Just read the other thread, @mkmeral to clarify, you said above that you're in favor of enabling by default to avoid leaving traps for customers, but here you're saying we should not default to `True` because sandbox storage won't need it. Can you help me understand what the distinction is?

mkmeral: I'll continue in the thread above 😅

I am okay with defaulting to true for in memory storage, not for all storage types

src/strands/vended_plugins/context_offloader/plugin.py
mkmeral: I don't think we should be defaulting to True, or if we will, we should improve the tool.

Why? tool as it exists today doesn't solve the problem. If you have a Json tool result that's 5 million tokens (just a huge number, bigger than context size), and you offload it. Then the LLM will try to read it again, and then you dump the tool result and cause context overflow. It's just a way for LLM to bypass offloading and maybe cause context issues

If we want this tool to be enabled, we should improve it. For example allow agent to search, view by lines, use jq to analyse json data, maybe a way to decrease image content size or sth. This would allow LLM to get the data selectively.

mkmeral: I am okay with enabling this by default for in memory storage, because there is no other way for agent to access that data. So not having the tool is essentially loosing data. But this only applies to in memory storage

lizradway: i agree with your example and think that **some form of smart search needs to be added to the tool as a fast follow**, but i would also say that the 80% use of this tool isn't reading jsons that are 5 million tokens, but the majority case is tool results that are somewhere around or a bit higher than the default token limit to offload (2500 tokens). 

in this vast majority case, the agent is not going to reach context overflow from retrieving a singular tool result (esp given the huge context window gains in recent models). in that edge case, the context window would be blown without this tool anyways

the agent should follow this priority:
1. use preview to answer question
2. if preview is insufficient, use existing tools
3. if existing tools are insufficient, use retrieval tool.

to me, this is a sufficient gain to ship on its own and not block on smart search

lizradway: im cool with enabling this by default for in memory storage and default off for everything else. the devx is a little misleading/inconsistent, but i think it makes sense in the larger picture.

src/strands/vended_plugins/context_offloader/plugin.py
```diff
@@ -278,7 +273,10 @@ async def _handle_tool_result(self, event: AfterToolCallEvent) -> None:
             "Use your available tools to selectively access the data you need."
         )
         if self._include_retrieval_tool:
-            guidance += "\nYou can also use retrieve_offloaded_content with a reference to get the full content."
+            guidance += (
+                "\nOnly use retrieve_offloaded_content as a fallback"
+                " if the data cannot be accessed using your existing tools."
```
mkmeral: why did we update this part? I don't understand the new instructions.

I'd keep the old instructions, and add this new instruction to the tool schema

mkmeral (approved):

Issue #2121, opened 2026-04-14
[BUG] CachePointBlock ttl field dropped by SDK + cache_tools ordering violation with 1h TTL


namartaknev:
### Checks

- [x] I have updated to the lastest minor and patch version of Strands
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched [./issues](./issues?q=) and there are no duplicates of my issue

### Strands Version

1.35

### Python Version

3.12

### Operating System

linux

### Installation Method

pip

### Steps to Reproduce

1. Create a BedrockModel with cache_tools="default" and a system prompt containing a cachePoint with ttl="1h"
2. Create an Agent with ConversationCacheHook that sets CACHE_POINT_BLOCK = {"cachePoint": {"type": "default", "ttl": "1h"}}
3. Invoke the agent with a multi-turn conversation
4. Observe: (a) ValidationException due to TTL ordering, or (b) cache falls back to 5m TTL silently

### Expected Behavior

cachePoint blocks with "ttl": "1h" should be forwarded to Bedrock as-is
cache_tools="default" should allow configuring its TTL to match other cache points
Bedrock should honor 1-hour TTL on all configured cache checkpoints

### Actual Behavior

Bug 1: _format_request_message_content in bedrock.py silently drops the ttl field from cachePoint blocks — all cache silently falls back to 5m TTL regardless of what is set
Bug 2: cache_tools="default" hardcodes a 5m TTL on the toolConfig cache point. Since toolConfig is processed before system and messages, a 1h TTL on a subsequent cache point violates Bedrock's decreasing TTL ordering rule → ValidationException

### Additional Context

The Bedrock CachePointBlock API supports an optional ttl field with values "5m" or "1h": https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_CachePointBlock.html
Bedrock processes cache checkpoints in order: toolConfig → system → messages, and requires TTLs to be in non-increasing order
Users leveraging ConversationCacheHook (manually injected cache points via hooks) cannot access 1h TTL because the SDK strips the field before the API call

### Possible Solution

```python
# Bug 1 fix — bedrock.py _format_request_message_content
if "cachePoint" in content:
    cp = {"type": content["cachePoint"]["type"]}
    if "ttl" in content["cachePoint"]:
        cp["ttl"] = content["cachePoint"]["ttl"]
    return {"cachePoint": cp}
```

```python
[...]

PR #1660, opened 2026-02-10
feat(cache): add TTL support to CachePoint for prompt caching

kpx-dev:
## Description
Add optional ttl field to CachePoint TypedDict to support AWS Bedrock's cache TTL configuration. The field accepts "5m" or "1h" values as specified in the Bedrock API. Updated BedrockModel to preserve ttl field when formatting cache point content blocks. Includes tests for TTL preservation and backward compatibility.

Doc shows API supports ttl option now: 
https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_CachePointBlock.html

launch announcement: 
https://aws.amazon.com/about-aws/whats-new/2026/01/amazon-bedrock-one-hour-duration-prompt-caching/

more details: 
https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html

## Related Issues

#1243 

## Type of Change

New feature

## Testing

How have you tested the change?  Verify that the changes do not break functionality or introduce warnings in consuming repositories: agents-docs, agents-tools, agents-cli

- [x ] I ran `hatch run prepare`

## Checklist
- [x ] I have read the CONTRIBUTING document
- [ x] I have added any necessary tests that prove my fix is effective or my feature works
- [x ] I have updated the documentation accordingly
- [ x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x ] My changes generate no new warnings
- [ x] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2232, opened 2026-04-30
feat(bedrock): add TTL support to auto-injected tool and system/user cache points

### LINKED ISSUE #1243: [FEATURE]  Increase TTL for prompt caching > 5 minutes
### Problem Statement

Longer running workloads can not make use of caching because of the 5 minutes TTL while for these use cases it can be of great value.

### Proposed Solution

Increase the TTL longer than 5 minutes. With the api of Anthropic self you can chose 5 minutes or 1 hour TTL for caching. With strandsagents this is 5 minutes

### Use Case

I want to be flexible to set TTL for caching or 15/30/45/60 minutes.

### Alternatives Solutions

Keep cache alive, but that is also extra burden and waste of tokens

### Additional Context

_No response_

kpx-dev:
## Description

Extends prompt caching TTL coverage beyond the user-supplied `cachePoint` path (already fixed in #1660) to the two SDK-managed auto-injected paths on `BedrockModel`.

Previously, only cache points that users passed directly inside message content could carry a `ttl` field — the formatter preserved it end-to-end. However, when the SDK auto-injected cache points on the user's behalf, the `ttl` was hardcoded away:

1. **`cache_tools`** — `BedrockModel(cache_tools="default")` appends `{"cachePoint": {"type": "default"}}` to `toolConfig.tools` with no way to set a TTL.
2. **`_inject_cache_point`** — when `cache_config=CacheConfig(strategy="auto")` is set, the SDK appends `{"cachePoint": {"type": "default"}}` to the last user message with no way to set a TTL.

Because Bedrock processes cache checkpoints in order `toolConfig → system → messages` and requires TTLs to be **non-increasing**, a user who wanted to run 1h caching end-to-end could not — the SDK-managed checkpoints would implicitly be 5m and violate the ordering rule (see #2121 Bug 2).

### Changes

- **`BedrockModel`**: added `cache_tools_ttl: str | None` config option. When set together with `cache_tools`, the TTL is propagated into the `toolConfig` cache point.
- **`CacheConfig`**: added `ttl: str | None` field. When set, `_inject_cache_point` includes the TTL in the cache point appended to the last user message.

Users can now align all three cache checkpoint TTLs consistently:

```python
from strands.models.model import CacheConfig

model = BedrockModel(
    model_id="us.anthropic.claude-haiku-4-5-20251001-v1:0",
    cache_tools="default",
    cache_tools_ttl="1h",
    cache_config=CacheConfig(strategy="auto", ttl="1h"),
)
```

API reference: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_CachePointBlock.html
Launch announcement: https://aws.amazon.com/about-aws/whats-new/2026/01/amazon-bedrock-one-hour-duration-prompt-caching/

## Related Issues

Closes #1243 (longer TTL for prompt caching) — together with the already-merged #1660 (user-supplied `cachePoint.ttl`), this PR completes TTL coverage across all three Bedrock cache checkpoint paths (`toolConfig`, system, messages), so users can opt into Bedrock's 1h TTL end-to-end.

Partially addresses #2121 (Bug 2: `cache_tools` ordering violation with 1h TTL). Bug 1 from #2121 (user-supplied `ttl` being dropped) was resolved by #1660.

## Type of Change

New feature

## Testing

Added 4 unit tests:

- `test_inject_cache_point_with_ttl` — verifies `CacheConfig(ttl="5m")` propagates into the auto-injected message cache point.
- `test_inject_cache_point_without_ttl` — backward compat: `CacheConfig(strategy="auto")` with no TTL still works.
- `test_format_request_cache_tools_with_ttl` — verifies `cache_tools_ttl` propagates into `toolConfig` cache point.
- `test_format_request_cache_tools_without_ttl` — backward compat: `cache_tools` without TTL still works.

Verified:

- [x] I ran `hatch run prepare` (lint, format, mypy all clean; pre-existing local env test failures unrelated to this change)
- [x] All 4 new tests pass

## Checklist

----


kpx-dev:
Added 3 integration tests covering the new paths. Ran locally against `us-east-1` with `us.anthropic.claude-haiku-4-5-20251001-v1:0`:

```
tests_integ/models/test_model_bedrock.py::test_prompt_caching_cache_tools_ttl               PASSED
tests_integ/models/test_model_bedrock.py::test_prompt_caching_cache_config_auto_with_ttl    PASSED
tests_integ/models/test_model_bedrock.py::test_prompt_caching_aligned_1h_ttl_across_checkpoints  PASSED
```

Coverage matrix:

| Test | What it proves |
|---|---|
| `test_prompt_caching_cache_tools_ttl` | `cache_tools_ttl` is accepted by Bedrock on the `toolConfig` cache checkpoint without `ValidationException`. |
| `test_prompt_caching_cache_config_auto_with_ttl` | `CacheConfig(strategy="auto", ttl="5m")` propagates TTL to the auto-injected message cache point and produces a cacheWrite. |
| `test_prompt_caching_aligned_1h_ttl_across_checkpoints` | **Regression test for #2121**: 1h TTL on all three cache checkpoints (toolConfig → system → messages) satisfies Bedrock's non-increasing TTL ordering rule. This was the specific scenario that was impossible before this PR.

kpx-dev:
@mkmeral this is an addition to merged PR https://github.com/strands-agents/sdk-python/pull/1660

can the team help review this PR? Thanks!

src/strands/models/bedrock.py
```diff
@@ -128,6 +129,7 @@ class BedrockConfig(BaseModelConfig, total=False):
         cache_prompt: str | None
         cache_config: CacheConfig | None
         cache_tools: str | None
+        cache_tools_ttl: str | None
```
pgrayy: I would recommend expanding the type of cache_tools to ensure that ttl is configured with the cache tools type. So something like:
```python
cache_tools: str | CacheToolsConfig | None
...
@dataclass
class CacheToolsConfig:
    type: str = "default"
    ttl: str | None = None
```

kpx-dev: Good call — refactored in cc32ea8. Added a `CacheToolsConfig` dataclass with `type` (default `"default"`) and `ttl` fields, and changed `cache_tools` to `str | CacheToolsConfig | None`. Removed the standalone `cache_tools_ttl` field so TTL can only be set together with the type. The plain `str` form still works for back-compat.

```python
BedrockModel(cache_tools=CacheToolsConfig(type="default", ttl="1h"))
# or, unchanged:
BedrockModel(cache_tools="default")
```

src/strands/models/bedrock.py
```diff
@@ -69,6 +70,19 @@ def _clear_skip_count_tokens_cache() -> None:
 DEFAULT_READ_TIMEOUT = 120
 
 
+@dataclass
+class CacheToolsConfig:
```
pgrayy: I would recommend defining and exporting this from model.py for consistency with CacheConfig.

tests_integ/models/test_model_bedrock.py
```diff
@@ -561,3 +570,127 @@ def calculator(expression: str) -> float:
...
+    call succeeds.
+
+    Uses Claude Haiku 4.5 which supports 1h TTL per
+    https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html
+    """
+    model = BedrockModel(
+        model_id=_CACHE_TTL_MODEL_ID,
+        streaming=False,
+        cache_tools=CacheToolsConfig(type="default", ttl="1h"),
+        cache_config=CacheConfig(strategy="auto", ttl="1h"),
```
pgrayy: Does this mean you can't have say 5m for tools and 1h for messages? But the opposite would be allowed since it would be decreasing? Do you have a link to the docs that outlines these rules?

kpx-dev: Good point — pivoted in 8759076. Dropped `CacheToolsConfig` entirely. The toolConfig cache point now picks up its TTL from `cache_config.ttl`, so a single knob covers both SDK-managed checkpoints. Bedrock's non-increasing rule is satisfied trivially since both checkpoints share one value.

```python
BedrockModel(
    cache_tools="default",
    cache_config=CacheConfig(strategy="auto", ttl="1h"),
)
# toolConfig cache point AND last-user-message cache point both get ttl="1h"
```

This also resolves your other comment about moving the dataclass to model.py — there's no new dataclass to move now.

src/strands/models/bedrock.py
kpx-dev: Resolved by 8759076 — went with your other suggestion (drop `CacheToolsConfig` entirely and reuse `cache_config.ttl` for the tool checkpoint), so there's no new dataclass to move.

tests_integ/models/test_model_bedrock.py
pgrayy: Sorry for the back and forth on this. I was able to sit down and test the behaviors. I think when misconfigured, the error message from bedrock is pretty clear. The following is an example:
```
Exception type: ValidationException
Message: An error occurred (ValidationException) when calling the Converse operation: The model returned the following errors: system.0.cache_control.ttl: a ttl='1h' cache_control block must not come after a ttl='5m' cache_control block. Note that blocks are processed in the following order: `tools`, `system`, `messages`.
```

With that said, I would recommend reverting back to your previous commit with CacheToolsConfig. The only change then I would suggest is defining it and exporting it from model.py instead of bedrock.py. Everything else looked good to me.

Apologies again for the back and forth.

kpx-dev: thanks updated!

pgrayy (approved):

PR #2249, opened 2026-05-05
feat: add context window limit lookup table

### LINKED ISSUE #1295: [FEATURE] Context Limit Property on Model Interface
## Problem Statement

The `Model` abstract class does not expose the model's context window limit. Applications must hard-code context limits or maintain separate model-to-limit mappings.

The SDK handles `ContextWindowOverflowException` reactively but provides no way to query capacity beforehand. This makes it impossible to implement threshold-based management like "compress at 70% of context window" (see #555).

## Proposed Solution

Add a `context_limit` property to the `Model` interface:

```python
class Model(ABC):
    @property
    @abstractmethod
    def context_limit(self) -> int:
        """Maximum context window size in tokens."""
        pass
```

Model providers implement with known values:

```python
class BedrockModel(Model):
    MODEL_CONTEXT_LIMITS = {
        "anthropic.claude-sonnet-4-20250514-v1:0": 200000,
        "amazon.nova-pro-v1:0": 300000,
    }
    
    @property
    def context_limit(self) -> int:
        return self.MODEL_CONTEXT_LIMITS.get(self.model_id, 128000)
```

## Use Case

1. Dynamic threshold calculation - `threshold = model.context_limit * 0.7`
2. UI progress indicators - display context usage percentage
3. Model selection - choose models based on context requirements
4. ConversationManager configuration - auto-tune window sizes

```python
usage_pct = current_tokens / agent.model.context_limit
if usage_pct > 0.7:
    agent.conversation_manager.reduce_context(agent)
```

## Alternative Solutions

Applications can maintain their own model-to-limit mappings, but this is error-prone and does not stay updated with new model releases.

## Additional Context

This feature, combined with a token estimation API (#1294), enables implementation of #555 (Proactive Context Compression). Related: #637 (max tokens handling).

opieter-aws:
## Description

`context_window_limit` is required for proactive context compression (#555, #1295), but users currently have to manually look up and hard-code context window sizes for every model they instantiate. This is error-prone and creates friction for the most common models across all providers.
    
`context_window_limit` is now auto-populated from a built-in lookup table when not explicitly provided in the model config. Explicit values always take precedence. Unknown model IDs leave it `None`.
  
  ```python
  # Before: users had to supply context_window_limit manually
  model = BedrockModel(
      model_id="anthropic.claude-sonnet-4-20250514-v1:0",
      context_window_limit=1_000_000,
  )
  
  # After: resolved automatically from the model ID
  model = BedrockModel(model_id="anthropic.claude-sonnet-4-20250514-v1:0")
  model.get_config().get("context_window_limit")  # 1_000_000
```  

A new `_defaults.py` module maps model IDs to their known context window limits. Each provider's `get_config()` lazily resolves the limit via `resolve_config_metadata()`. The stored config is never
mutated, so changing model_id via `update_config()` naturally re-resolves on the next `get_config()` call.

For Bedrock cross-region model IDs (e.g. us.anthropic.claude-sonnet-4-6), the region prefix is stripped before lookup.

Values sourced from https://github.com/BerriAI/litellm/blob/litellm_internal_staging/model_prices_and_context_window.json

## Related Issues

  Port of strands-agents/sdk-typescript#954.
  
  Resolves: #1295

## Documentation PR

Part of proactive compression doc updates

## Type of Change

New feature

## Testing

## Checklist

----

### DIFF
--- a/src/strands/models/_defaults.py
+++ b/src/strands/models/_defaults.py
@@ -1,0 +1,177 @@
+"""Default model metadata lookup tables.
+
+Provides context window limits for known model IDs across all providers.
+Values sourced from provider documentation and
+https://github.com/BerriAI/litellm/blob/litellm_internal_staging/model_prices_and_context_window.json
+
+Applied to providers with well-known, fixed model IDs: Bedrock, Anthropic, OpenAI,
+OpenAI Responses, Gemini, and Mistral. Providers that use local/custom model IDs
+(Ollama, LlamaCpp, SageMaker) or proxy to other providers with their own prefixed
+ID format (LiteLLM) are excluded — their context windows depend on deployment config,
+not a static table.
+"""
+
+import logging
+from collections.abc import Mapping
+from typing import TypeVar
+
+logger = logging.getLogger(__name__)
+
+_C = TypeVar("_C", bound=Mapping[str, object])
+
+# Context window limits (in tokens) for known model IDs.
+#
+# Best-effort lookup table — unknown models return None and callers
+# fall back gracefully (e.g. proactive compression is disabled).
+# Users can always override with an explicit context_window_limit in their model config.
+#
+# For Bedrock models with cross-region prefixes (e.g. us., eu., global.),
+# get_context_window_limit strips the prefix before lookup so only the base model ID is needed here.
+_CONTEXT_WINDOW_LIMITS: dict[str, int] = {
+    # Anthropic (direct API)
+    "claude-sonnet-4-6": 1_000_000,
+    "claude-sonnet-4-20250514": 1_000_000,
+    "claude-sonnet-4-5": 200_000,
+    "claude-sonnet-4-5-20250929": 200_000,
+    "claude-opus-4-6": 1_000_000,
+    "claude-opus-4-6-20260205": 1_000_000,
+    "claude-opus-4-7": 1_000_000,
+    "claude-opus-4-7-20260416": 1_000_000,
+    "claude-opus-4-5": 200_000,
+    "claude-opus-4-5-20251101": 200_000,
+    "claude-opus-4-20250514": 200_000,
+    "claude-opus-4-1": 200_000,
+    "claude-opus-4-1-20250805": 200_000,
+    "claude-haiku-4-5": 200_000,
+    "claude-haiku-4-5-20251001": 200_000,
+    "claude-3-7-sonnet-20250219": 200_000,
+    "claude-3-5-sonnet-20241022": 200_000,
+    "claude-3-5-sonnet-20240620": 200_000,
+    "claude-3-5-haiku-20241022": 200_000,
+    "claude-3-opus-20240229": 200_000,
+    "claude-3-haiku-20240307": 200_000,
+    # Bedrock Anthropic (base model IDs — cross-region prefixes stripped by get_context_window_limit)
+    "anthropic.claude-sonnet-4-6": 1_000_000,
+    "anthropic.claude-sonnet-4-20250514-v1:0": 1_000_000,
+    "anthropic.claude-sonnet-4-5-20250929-v1:0": 200_000,
+    "anthropic.claude-opus-4-6-v1": 1_000_000,
+    "anthropic.claude-opus-4-7": 1_000_000,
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: src/strands/models/_defaults.py [hunk 1]
```diff
@@ -0,0 +1,157 @@
+"""Default model metadata lookup tables.
+
+Provides context window limits for known model IDs across all providers.
+Values sourced from provider documentation and
+https://github.com/BerriAI/litellm/blob/litellm_internal_staging/model_prices_and_context_window.json
+"""
```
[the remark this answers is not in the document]
opieter-aws: Updating will be best-effort

src/strands/models/openai_responses.py
```diff
@@ -210,7 +211,10 @@ def get_config(self) -> OpenAIResponsesConfig:
         Returns:
             The OpenAI Responses API model configuration.
         """
-        return cast(OpenAIResponsesModel.OpenAIResponsesConfig, self.config)
+        return cast(
+            OpenAIResponsesModel.OpenAIResponsesConfig,
+            resolve_config_metadata(self.config, str(self.config.get("model_id", ""))),
```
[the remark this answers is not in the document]
opieter-aws: mypy requires this to pass checks

src/strands/models/openai.py
```diff
@@ -150,7 +151,9 @@ def get_config(self) -> OpenAIConfig:
         Returns:
             The OpenAI model configuration.
         """
-        return cast(OpenAIModel.OpenAIConfig, self.config)
+        return cast(
+            OpenAIModel.OpenAIConfig, resolve_config_metadata(self.config, str(self.config.get("model_id", "")))
```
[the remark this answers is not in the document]
opieter-aws: mypy requires this to pass checks

src/strands/models/bedrock.py
```diff
@@ -217,7 +218,7 @@ def get_config(self) -> BedrockConfig:
         Returns:
             The Bedrock model configuration.
         """
-        return self.config
+        return resolve_config_metadata(self.config, self.config.get("model_id", ""))
```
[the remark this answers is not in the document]
opieter-aws: Not a concern in practice. Internal methods (`_format_request()`, `count_tokens()`, etc.) read `self.config` directly, not `self.get_config()`. The only external caller is the agent loop reading `context_window_limit` once per invocation.

src/strands/models/_defaults.py
```diff
@@ -0,0 +1,177 @@
...
+from typing import TypeVar
+
+logger = logging.getLogger(__name__)
+
+_C = TypeVar("_C", bound=Mapping[str, object])
+
+# Context window limits (in tokens) for known model IDs.
+#
+# Best-effort lookup table — unknown models return None and callers
+# fall back gracefully (e.g. proactive compression is disabled).
```
lizradway, line 25: > proactive compression is disabled

nit: is this still true?

lizradway: ah nvm, the fallback is in proactive compression if i remember correctly?

opieter-aws: Correct, proactive compression implements the fallback

lizradway: ok just checked and i think we decided it would be there, but i don't see it right now. is that still the plan?

lizradway (approved):

src/strands/models/_defaults.py
```diff
@@ -0,0 +1,177 @@
...
+
+    Returns:
+        The config with resolved metadata, or the original config if nothing to resolve.
+    """
+    if "context_window_limit" in config:
+        return config
+
+    limit = get_context_window_limit(model_id)
+    if limit is None:
+        return config
```
lizradway, line 175: should we warn once for this?

opieter-aws: I don't think we need to warn if context window limit is `undefined`, since it's not relevant to the system if proactive compression isn't used. We'll warn once in the conversation manager if proactive compression is enabled and the limit is undefined. In that case we'll fallback to a default of 200k.

lizradway: yes ok awesome, that makes sense

PR #2253, opened 2026-05-06
feat: add Skill.from_s3 for dynamic loading skills from S3

jeromevdl:
## Description
Teams managing multiple agents often centralize skill definitions in S3 rather than bundling them in application code. This enables shared skill libraries, versioning via S3 prefixes, and decoupling skill authoring from agent deployment (dynamic loading).

This adds a `Skill.from_s3` classmethod that discovers skill directories in an S3 bucket, mirrors them locally (with parallel downloads and per-process caching), then delegates to the existing `from_directory` for parsing.

### Public API Changes
```python
from strands.vended_plugins.skills import Skill

# Load all skills under a prefix
skills = Skill.from_s3("my-bucket", prefix="agents/director/")

# With a custom S3 client
import boto3
s3 = boto3.client("s3", region_name="eu-west-1")
skills = Skill.from_s3("my-bucket", s3_client=s3)
```

Parameters: `bucket`, `prefix`, `s3_client`, `local_dir`, `strict`

## Related Issues

None

## Documentation PR

Not yet available, will document if the feature is accepted.

## Type of Change

New feature

## Testing

15 new unit tests using moto to mock S3

Covers: single/multiple skills, resource file mirroring, caching, prefix normalization, empty buckets, strict mode, lowercase skill.md discovery

All 154 skill tests pass

## Checklist

----

### DIFF
--- a/src/strands/vended_plugins/skills/skill.py
+++ b/src/strands/vended_plugins/skills/skill.py
@@ -2,26 +2,31 @@
 
 This module defines the Skill dataclass and provides classmethods for
 discovering, parsing, and loading skills from the filesystem, raw content,
-or HTTPS URLs. Skills are directories containing a SKILL.md file with YAML
-frontmatter metadata and markdown instructions.
+HTTPS URLs, or Amazon S3. Skills are directories containing a SKILL.md file
+with YAML frontmatter metadata and markdown instructions.
 """
 
 from __future__ import annotations
 
 import logging
 import re
+import tempfile
 import urllib.error
 import urllib.request
+from concurrent.futures import ThreadPoolExecutor, as_completed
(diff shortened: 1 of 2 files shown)


pgrayy:
Thank you for sharing this proposal, and for the thorough tests. Unfortunately, I don't think `from_s3` is the right fit for the core SDK. Baking a specific cloud-storage backend (with boto3, parallel downloads, and a local cache) into `Skill` pulls provider-specific concerns into a deliberately small, dependency-light surface.

As an alternative, you could consider wrapping this on your side: download the prefix to a temp dir and hand it to `Skill.from_directory`, which already accepts any local path. That keeps the S3 mirroring, caching, and versioning logic in your own thin helper or plugin.

PR #2257, opened 2026-05-06
fix(bedrock): resolve cache strategy for ARN-based application inference profiles

### LINKED ISSUE #2601: [BUG] CacheConfig(strategy="auto") does not detect Claude models when using ARN-based inference profiles
### Description

`CacheConfig(strategy="auto")` fails to activate prompt caching when the `model_id` is an ARN-based application inference profile (e.g., `arn:aws:bedrock:us-east-1:123456:application-inference-profile/abc123`).

### Root cause

The `_cache_strategy` property in `BedrockModel` checks for `"claude"` or `"anthropic"` substrings in the model ID:

```python
# strands/models/bedrock.py, line ~192
def _cache_strategy(self) -> str | None:
    model_id = self.config.get("model_id", "").lower()
    if "claude" in model_id or "anthropic" in model_id:
        return "anthropic"
    return None
```

ARN-based inference profiles contain neither substring — they look like `arn:aws:bedrock:us-east-1:123456:application-inference-profile/2i76zr9zafkf`. So `auto` resolves to `None` and caching is silently skipped.

### Reproduction

```python
from strands.models.model import CacheConfig

model = BedrockModel(
    model_id="arn:aws:bedrock:us-east-1:123456:application-inference-profile/abc123",
    cache_config=CacheConfig(strategy="auto"),
)

# _cache_strategy returns None → caching silently disabled
# cacheReadInputTokens and cacheWriteInputTokens will always be 0
```

### Expected behavior

`auto` should detect Claude models behind ARN inference profiles and enable caching. Possible approaches:

1. **Resolve the ARN** at init time via `bedrock.get_inference_profile()` to discover the underlying foundation model
2. **Try caching optimistically** — non-cache-capable models ignore `cachePoint` blocks in the Converse API, so injecting them is harmless
3. **Check the response** for `cacheWriteInputTokens > 0` on the first call to confirm support

### Workaround

Use `CacheConfig(strategy="anthropic")` instead of `"auto"` to force-enable caching regardless of model ID detection:

```python
model = BedrockModel(
    model_id="arn:aws:bedrock:us-east-1:123456:application-inference-profile/abc123",
    cache_config=CacheConfig(strategy="anthropic"),
)
```

### Environment

- strands-agents 1.37.0
- Python 3.12
- AWS Bedrock with application inference profiles for Claude Sonnet 4.6

Zelys-DFKH:
## Description

`CacheConfig(strategy="auto")` silently disabled caching for ARN-based application inference profiles like `arn:aws:bedrock:us-east-1:123456:application-inference-profile/abc123`. Application inference profiles use opaque resource IDs that don't contain `"claude"` or `"anthropic"`, so `_cache_strategy`'s string-match returns `None` and logs a spurious warning, even when the profile is backed by Claude Sonnet.

The fix detects the `application-inference-profile` ARN pattern and calls `GetInferenceProfile` on the Bedrock management API to find the underlying foundation model. That model ARN does contain `"anthropic"` for Claude-backed profiles, so the existing string-match works correctly from there.

Two implementation notes:

- `GetInferenceProfile` is on the `bedrock` management client, not `bedrock-runtime`. The fix stores the boto session in `__init__` so the management client can be created on demand using the same credentials and region. It requires the `bedrock:GetInferenceProfile` IAM permission; on error it falls back to `None` and logs a debug message pointing to `CacheConfig(strategy="anthropic")` as an explicit workaround.
- The resolved strategy is cached on the instance (one API call per model) and cleared by `update_config` when `model_id` changes.

`_should_include_tool_result_status` has the same string-match issue for application inference profiles. Lower impact, but I can address that in a follow-up if useful.

## Related Issues

Closes strands-agents/sdk-python#2601

## Documentation PR

No docs change needed — the public API is unchanged.

## Type of Change

Bug fix

## Testing

Six new tests in `test_bedrock.py`:

- Claude-backed profile resolves to `"anthropic"`; verifies the `bedrock` management client is used (not `bedrock-runtime`)
- Non-Claude-backed profile returns `None`
- `GetInferenceProfile` failure handled gracefully with no exception raised
- `GetInferenceProfile` called only once per instance across multiple `_cache_strategy` accesses
- Cached result cleared when `update_config` changes `model_id`
- End-to-end: `cachePoint` injected into messages when profile resolves to Claude with `strategy="auto"`

163 bedrock tests pass. Full suite: 2873 passed, 4 skipped. `hatch run prepare` clean.

## Checklist

### DIFF
--- a/src/strands/models/bedrock.py
+++ b/src/strands/models/bedrock.py
@@ -220,9 +220,10 @@ class BedrockModel(Model):
     def _resolve_application_inference_profile_strategy(self) -> str | None:
         """Resolve the cache strategy for an ARN-based application inference profile.
 
-        Calls GetInferenceProfile on the Bedrock management API to discover the underlying
+        Calls GetInferenceProfile on the Bedrock management API (requires the
+        ``bedrock:GetInferenceProfile`` IAM permission) to discover the underlying
         foundation model, then checks whether that model supports prompt caching.
-        Returns None (and logs a debug message) on any error.
+        Returns None and logs a debug message on any error, including missing permissions.
         """
         try:
             # GetInferenceProfile is a Bedrock management API, not available on bedrock-runtime.
--- a/src/strands/models/bedrock.py
+++ b/src/strands/models/bedrock.py
@@ -208,6 +208,32 @@ class BedrockModel(Model):
         model_id = self.config.get("model_id", "").lower()
         if "claude" in model_id or "anthropic" in model_id:
             return "anthropic"
+        # ARN-based application inference profiles use opaque IDs that don't contain the model name.
+        # Resolve the underlying foundation model via GetInferenceProfile, then re-check.
+        if "application-inference-profile" in model_id:
+            if not hasattr(self, "_resolved_application_profile_strategy"):
+                self._resolved_application_profile_strategy = self._resolve_application_inference_profile_strategy()
+            return self._resolved_application_profile_strategy
+        return None
+
+    def _resolve_application_inference_profile_strategy(self) -> str | None:
+        """Resolve the cache strategy for an ARN-based application inference profile.
+
+        Calls GetInferenceProfile to discover the underlying foundation model, then checks whether
+        that model supports prompt caching.  Returns None (and logs a debug message) on any error.
+        """
+        try:
+            response = self.client.get_inference_profile(inferenceProfileIdentifier=self.config["model_id"])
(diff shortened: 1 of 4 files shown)


mkmeral:
hi @Zelys-DFKH thank you for the PR, however we will not accept it, as this was an explicit decision made in https://github.com/strands-agents/sdk-python/issues/1705#issuecomment-3993868788 

I will close it now

Issue #1722, opened 2026-02-19
[FEATURE][EPIC] Bidirectional Streaming - Move out of Experimental


mehtarac:
### Problem Statement

This is the parent issue to track the bidirectional streaming feature to move out of experimental module. 

### Use Case

NA

### Alternatives Solutions

NA

PR #2260, opened 2026-05-07
fix(bidi): add configurable event_queue_size to prevent choppy audio

madhavi-joshi-nutrien:
Motivation
When using BidiAgent with BidiGeminiLiveModel, audio playback is choppy and bursty. The internal event queue between the model receive loop and the output handler is hardcoded to maxsize=1. This causes the model receive loop to block whenever the output handler performs any async I/O (e.g., websocket.send_json()), resulting in audio chunks piling up on the model SDK side and arriving in bursts.

This is particularly noticeable with Gemini Live, which sends many small audio chunks rapidly (~50 chunks/sec), unlike Nova Sonic which sends fewer, larger chunks with built-in TTS buffering.

Public API Changes
BidiAgent.__init__() accepts a new event_queue_size parameter:

# Default behavior preserved (maxsize=1)
agent = BidiAgent(model=model, tools=[...])

# Opt in to larger buffer for smooth audio with fast-delivering models
agent = BidiAgent(
    model=BidiGeminiLiveModel(...),
    tools=[...],
    event_queue_size=32,  # ~640ms buffer at typical Gemini chunk rates
)
The parameter controls the asyncio.Queue maxsize between the model receive loop and the output handler. Higher values absorb I/O latency spikes without stalling the model loop.

Use Cases
Gemini Live audio streaming: Prevents choppy playback when the output handler has non-trivial async I/O latency (WebSocket, WebRTC)
Custom output handlers: Any output handler doing network I/O benefits from decoupling via a larger buffer
Tuning backpressure: Users can balance memory usage vs. smoothness for their specific model and handler combination

### DIFF
--- a/src/strands/experimental/bidi/agent/agent.py
+++ b/src/strands/experimental/bidi/agent/agent.py
@@ -74,6 +74,7 @@ class BidiAgent:
         state: AgentState | dict | None = None,
         session_manager: "SessionManager | None" = None,
         tool_executor: ToolExecutor | None = None,
+        event_queue_size: int = 1,
         **kwargs: Any,
     ):
         """Initialize bidirectional agent.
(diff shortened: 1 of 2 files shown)


mehtarac:
Closing this as we will include WebSocketIO and WebRTC support which will implicitly help smooth choppy audio and add echo suppression as well. Tracked EPIC here #1722

PR #2276, opened 2026-05-10
perf: replace deepcopy with shallow copy in graph/swarm state management

### LINKED ISSUE #2275: perf: deepcopy in GraphNode/SwarmNode state management adds ~1ms per graph node
## Description

`GraphNode.__post_init__` and `reset_executor_state` use `copy.deepcopy()` to save/restore conversation state. For a typical conversation (20 messages, 10 tools), each deepcopy call costs **0.218ms**.

In a 5-node graph, this adds **1.1ms** of pure copy overhead per execution. In a 20-node workflow, it's **9ms** of latency from copies alone.

## Benchmark

```python
import copy, timeit

messages = [{'role': 'user', 'content': [{'type': 'text', 'text': f'Msg {i}'}]} for i in range(20)]
model_state = {'temperature': 0.7, 'tools': [{'name': f't{i}'} for i in range(10)]}

# Current: deepcopy
t = timeit.timeit(lambda: (copy.deepcopy(messages), copy.deepcopy(model_state)), number=1000)
print(f'deepcopy: {t:.3f}ms per call')  # ~0.218ms

# Alternative: shallow copy each message dict
t = timeit.timeit(lambda: ([m.copy() for m in messages], model_state.copy()), number=1000)
print(f'shallow: {t:.6f}ms per call')  # ~0.003ms
# Speedup: ~67x
```

## Proposed Fix

Replace `copy.deepcopy()` with `[msg.copy() for msg in messages]` for message lists and `state.copy()` for model state.

This is safe because the executor only:
- Appends new messages (`agent.py:1123`)
- Replaces scalar values in model state

It does not mutate existing message dicts in-place.

## Impact

- 5-node graph: 1.1ms → 0.002ms copy overhead
- 20-node workflow: 9ms → 0.01ms copy overhead
- At 100 req/sec: saves 11% of a CPU core

I'd like to submit a PR for this if the approach sounds reasonable.

AnnasMazhar:
Closes #2275

## Summary

Replace `copy.deepcopy()` with targeted shallow copies in `GraphNode` and `SwarmNode` state save/restore.

## Motivation

`deepcopy` on a typical conversation (20 messages, 10 tools) costs 0.218ms per call. In a 5-node graph that's 1.1ms of pure copy overhead. See #2275 for full benchmark.

## Changes

- Added `_copy_messages()` helper: `[msg.copy() for msg in messages]` (~67x faster)
- Added `_copy_model_state()` helper: shallow dict copy + tools list copy (~70x faster)
- Replaced 8 `copy.deepcopy()` calls across `graph.py` and `swarm.py`
- Added 2 tests verifying copy isolation (append safety + mutation safety)

## Safety

Safe because the executor only:
1. Appends new messages (`agent.py:1123`) — never mutates existing message dicts
2. Replaces model state values — never mutates nested tool specs

## Testing

- `hatch fmt --formatter` ✓
- `hatch fmt --linter` ✓ (ruff + mypy pass)
- New tests: `test_copy_messages_isolation`, `test_copy_model_state_isolation`

### DIFF
--- a/src/strands/multiagent/graph.py
+++ b/src/strands/multiagent/graph.py
@@ -15,7 +15,6 @@ Key Features:
 """
 
 import asyncio
-import copy
 import logging
 import time
 from collections.abc import AsyncIterator, Callable, Mapping
@@ -54,6 +53,30 @@ from ..types.multiagent import MultiAgentInput
 from ..types.session import decode_bytes_values, encode_bytes_values
 from ..types.traces import AttributeValue
 from .base import MultiAgentBase, MultiAgentResult, NodeResult, Status
+
+
+def _copy_messages(messages: list) -> list:
+    """Shallow-copy message list and each message dict.
+
+    This is sufficient because the executor only appends new messages
+    and replaces content blocks — it does not mutate existing message
+    dicts in-place. ~67x faster than copy.deepcopy for typical
+    conversation sizes (20 messages with tool specs).
+    """
+    return [msg.copy() for msg in messages]
+
+
+def _copy_model_state(state: dict) -> dict:
+    """Shallow-copy model state dict.
+
+    Model state contains scalar values and a tools list that is
+    replaced, not mutated in-place. ~70x faster than copy.deepcopy.
+    """
+    new = state.copy()
+    if "tools" in new:
+        new["tools"] = state["tools"].copy()
+    return new
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


AnnasMazhar:
Closing this — after deeper analysis I found that the executor DOES mutate existing message dicts in-place:

- `agent.py:933`: `self.messages[-1]["content"] = ...` (replaces content list)
- `bedrock.py:398`: `messages[last_user_idx]["content"].append(...)` (appends to content)
- `repository_session_manager.py:303`: `messages[index + 1]["content"].extend(...)`

A shallow copy (`msg.copy()`) shares the `content` list reference, so these mutations would leak back to the saved initial state. `deepcopy` is necessary here for correctness.

A safe optimization would need to deep-copy only the `content` lists (not the entire recursive structure), but that requires more careful analysis of all mutation paths. Leaving the issue open for discussion.

Sorry for the premature PR.

PR #2284, opened 2026-05-12
fix: set use_native_token_count default to false

### LINKED ISSUE #2277: [BUG] use_native_token_count=True default causes silent latency regression for image/multimodal workloads
### Checks

### Strands Version

1.39.0

### Python Version

3.13

### Operating System

macOS

### Installation Method

pip

### Steps to Reproduce

 Since v1.38.0, the event loop calls `count_tokens()` before every model invocation via `_estimate_input_tokens()` in `event_loop.py`. When `use_native_token_count=True` (the default), this makes an additional network round-trip to Bedrock's CountTokens API that processes the full message payload — including images — before the actual inference call.

  For multimodal/image workloads, this adds significant per-invocation latency. Users upgrading past v1.38.0 see their latency increase with no code changes and no indication of why.

  PR #2255 added `use_native_token_count=False` as an opt-out, but the default remains `True`. This means the regression is silent — you have to know the flag exists to fix it.

Sample code to illustrate:

```python
  import io
  import time

  from PIL import Image
  from strands import Agent
  from strands.models import BedrockModel
  from strands.types.content import ContentBlock
  from strands.types.media import ImageContent, ImageSource

  MODEL_ID = "us.anthropic.claude-haiku-4-5-20251001-v1:0"
  N_RUNS = 1


  def make_image() -> bytes:
      """Generate a ~1MB PNG simulating a webpage screenshot."""
      import random
      random.seed(42)
      img = Image.new("RGB", (1280, 800), "white")
      pixels = img.load()
      for x in range(1280):
          for y in range(530, 800):
              pixels[x, y] = (random.randint(0, 255), random.randint(0, 255), random.randint(0, 255))
      buf = io.BytesIO()
      img.save(buf, format="PNG")
      return buf.getvalue()


  def invoke(image_bytes: bytes, use_native: bool) -> float:
      model = BedrockModel(
          model_id=MODEL_ID,
          temperature=0.0,
          max_tokens=4096,
          streaming=False,
          use_native_token_count=use_native,
      )
      agent = Agent(model=model, system_prompt="Respond in one sentence.", callback_handler=None)
      prompt = [
          ContentBlock(image=ImageContent(format="png", source=ImageSource(bytes=image_bytes))),
          ContentBlock(text="Describe this image in one sentence."),
      ]
      start = time.perf_counter()
      agent(prompt)
      return time.perf_counter() - start


  if __name__ == "__main__":
      print("Generating ~1MB test image...")
      image_bytes = make_image()
      print(f"Image size: {len(image_bytes) // 1024} KB\n")

      for use_native in [True, False]:
          label = f"use_native_token_count={use_native}"
          times = [invoke(image_bytes, use_native) for _ in range(N_RUNS)]
          avg = sum(times) / N_RUNS
          print(f"{label:40s} avg={avg:.2f}s  runs={[f'{t:.2f}' for t in times]}")]
```

### Expected Behavior

Expect that latency generally doesn't increase for inferences as a result of upgrading strands package versions

### Actual Behavior

Seems like `use_native_token_count=True` results in 25 - 50% increase in latency

### Additional Context

_No response_

### Possible Solution

Recommend having `use_native_token_count` be an opt-in feature rather than an opt out feature (default to False)

### Related Issues

_No response_

opieter-aws:
## Description

Changes the default value of `use_native_token_count` from `True` to `False` across all model providers (Bedrock, Anthropic, Gemini, LlamaCpp, OpenAI Responses).

Since v1.38.0, the event loop calls `count_tokens()` before every model invocation. When `use_native_token_count=True` (the previous default), this makes an additional network round-trip that processes the full message payload -- including images -- before the actual inference call. For multimodal/image workloads, this adds 25-50% per-invocation latency with no indication of why.

This change makes native token counting opt-in rather than opt-out, so users must explicitly set `use_native_token_count=True` to enable the native API calls.

## Related Issues

Fixes #2277

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

## Checklist

----

### DIFF
--- a/src/strands/models/anthropic.py
+++ b/src/strands/models/anthropic.py
@@ -58,8 +58,8 @@ class AnthropicModel(Model):
             params: Additional model parameters (e.g., temperature).
                 For a complete list of supported parameters, see https://docs.anthropic.com/en/api/messages.
             use_native_token_count: Whether to use the native Anthropic count_tokens API.
-                When True (default), count_tokens() calls the Anthropic API for accurate counts.
-                When False, skips the API call and uses the local estimator.
+                When True, count_tokens() calls the Anthropic API for accurate counts.
+                When False (default), skips the API call and uses the local estimator.
         """
 
         max_tokens: Required[int]
@@ -398,7 +398,7 @@ class AnthropicModel(Model):
         Returns:
             Total input token count.
         """
-        if self.config.get("use_native_token_count") is False:
+        if self.config.get("use_native_token_count") is not True:
             return await super().count_tokens(messages, tool_specs, system_prompt, system_prompt_content)
 
         try:
--- a/src/strands/models/bedrock.py
+++ b/src/strands/models/bedrock.py
@@ -118,8 +118,8 @@ class BedrockModel(Model):
             temperature: Controls randomness in generation (higher = more random)
             top_p: Controls diversity via nucleus sampling (alternative to temperature)
             use_native_token_count: Whether to use the native Bedrock CountTokens API.
-                When True (default), count_tokens() calls the Bedrock API for accurate counts.
-                When False, skips the API call and uses the local estimator.
+                When True, count_tokens() calls the Bedrock API for accurate counts.
+                When False (default), skips the API call and uses the local estimator.
         """
 
         additional_args: dict[str, Any] | None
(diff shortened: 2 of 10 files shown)


src/strands/models/anthropic.py
zastrowm, line 1: Do we expect this to change in a future version?

zastrowm (approved):

src/strands/models/anthropic.py
opieter-aws: We haven’t discussed future versions yet, but we are planning to control this setting as part of the context management presets in V1

Issue #1163, opened 2025-11-10
[FEATURE] Avoid initializing boto client every single time for S3SessionManager and AgentCore memory.


CrysisDeu:
### Problem Statement

**This issue also exist with AgentCore memory**

There are a couple issues with the S3SessionManager:
- There is no way to pass an S3SessionManager into the class. A new boto3 s3 client is created every single time.
- The sessionManager class is tied to a sessionId upon initialization. There is no way to update it afterwards. It means there is no way to save a singleton class.

These issue results in added latency for every single agent conversation turn and should be considered anti-patterns.

### Proposed Solution

- Let client pass in an S3 client, only create a new one if None is passed in
- Allow the S3SessionManager to be used as a singleton and can be invoked/set to different sessionIds

### Use Case

Using strands with S3Session manager or AgentCore memory

MukundaKatta:
Opened #2301 which adds an optional `s3_client=` kwarg to `S3SessionManager.__init__`. When provided, the manager reuses the caller's boto3 S3 client directly instead of constructing a new `boto3.Session` + client per instance.

The PR is backward-compatible (existing path unchanged when `s3_client` is omitted) and ships with four new moto-backed tests; the full session/S3 suite is 50 passed locally.

It only addresses the S3 half of this issue — the AgentCore memory side would benefit from the same treatment but lives outside this SDK repo as far as I can see. Happy to take a follow-up if there's a target repo for that piece.

PR #2301, opened 2026-05-16
feat(session): allow S3SessionManager to reuse a pre-built S3 client

MukundaKatta:
## Summary

Adds an optional `s3_client` keyword argument to `S3SessionManager.__init__` so callers can pass a pre-built `boto3` S3 client and avoid the per-instance overhead of constructing a fresh `boto3.Session` and S3 client on every manager creation.

When `s3_client` is provided, `boto_session`, `boto_client_config`, and `region_name` are ignored (documented in the docstring) and the caller's client is used as-is.

When `s3_client` is omitted, behavior is unchanged: the manager builds a `boto3.Session` (using `region_name` and `boto_session`) and a client with the `strands-agents` user-agent tag.

This is a backward-compatible additive change.

## Why

Refs #1163. Users instantiating many `S3SessionManager` instances in the same process (per-agent, per-session-id, etc.) currently pay the cost of `boto3.Session(...)` and `session.client("s3", ...)` on every manager. Each `session.client(...)` call creates a fresh urllib3 connection pool and runs endpoint discovery against AWS.

Passing a single shared S3 client across managers eliminates that overhead and lets callers control credentials, retry config, and custom endpoints (for example, MinIO or LocalStack) without subclassing.

A Strands maintainer (@pgrayy) acknowledged the issue on #1163: *"I agree that we should allow for reuse of the session manager instance so that user do not have to reinitialize the underlying client."*

## Example

Before:

```python
# Each manager constructs its own boto3.Session + S3 client.
mgr_a = S3SessionManager(session_id="a", bucket="b1", region_name="us-east-1")
mgr_b = S3SessionManager(session_id="b", bucket="b1", region_name="us-east-1")
mgr_c = S3SessionManager(session_id="c", bucket="b1", region_name="us-east-1")
```

After:

```python
import boto3

shared_client = boto3.client("s3", region_name="us-east-1")
mgr_a = S3SessionManager(session_id="a", bucket="b1", s3_client=shared_client)
mgr_b = S3SessionManager(session_id="b", bucket="b1", s3_client=shared_client)
mgr_c = S3SessionManager(session_id="c", bucket="b1", s3_client=shared_client)
```

## Tests

Added four `moto`-backed tests to `tests/strands/session/test_s3_session_manager.py`:

- `test_s3_client_kwarg_reuses_supplied_client` — supplied client becomes `manager.client`.
- `test_s3_client_kwarg_ignores_session_and_config` — `boto_session.client` is never called when `s3_client` is set.
- `test_s3_client_kwarg_supports_session_round_trip` — end-to-end create + read of a session through a manager built with the new kwarg.
- `test_default_path_still_works` — the existing path with no `s3_client` is unchanged.

Full session/S3 test suite: 50 passed locally.

## Test plan

- [x] `pytest tests/strands/session/test_s3_session_manager.py` passes
- [x] Existing callers (no `s3_client=` argument) get identical behavior
- [x] When `s3_client=` is supplied, no `boto3.Session` is constructed
- [x] Docstring documents which arguments are ignored when `s3_client` is supplied

Refs #1163.

### DIFF
--- a/strands-py/src/strands/session/s3_session_manager.py
+++ b/strands-py/src/strands/session/s3_session_manager.py
@@ -54,6 +54,7 @@ class S3SessionManager(RepositorySessionManager, SessionRepository):
         boto_client_config: BotocoreConfig | None = None,
         region_name: str | None = None,
         endpoint_url: str | None = None,
+        s3_client: Any = None,
         **kwargs: Any,
     ):
         """Initialize S3SessionManager with S3 storage.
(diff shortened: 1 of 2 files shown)


JackYPCOnline:
@MukundaKatta  Thank you for raising this PR, could you please rebase it?

JackYPCOnline:
Rebased onto latest \`main\` (S3 endpoint-URL conflict resolved) and addressed the review feedback in c3a2143d:

- **Typing**: \`s3_client\` now uses the \`S3Client\` stub from \`boto3-stubs[s3]\` (under \`TYPE_CHECKING\`), matching the pattern in \`bedrock_knowledge_base/store.py\`, instead of \`Any\`.
- **Test issue reference**: dropped \`(issue #1163)\` from the feature-development test section header per \`TESTING.md\`.
- **Redundant test**: replaced \`test_default_path_still_works\` (which duplicated existing default-path coverage) with \`test_s3_client_none_builds_tagged_client\`, asserting that an explicit \`s3_client=None\` still builds the \`strands-agents\`-tagged client.
- **Behavior nit**: added a \`logger.debug\` noting when a caller-supplied client is reused.

ruff, mypy, and the full session test suite pass.

JackYPCOnline (dismissed):

strands-py/src/strands/session/s3_session_manager.py
```diff
@@ -63,31 +65,45 @@ def __init__(
...
-            if existing_user_agent:
-                new_user_agent = f"{existing_user_agent} strands-agents"
-            else:
-                new_user_agent = "strands-agents"
-            client_config = boto_client_config.merge(BotocoreConfig(user_agent_extra=new_user_agent))
+        if s3_client is not None:
+            # Reuse the caller's pre-built client. We deliberately do not modify
+            # its user_agent_extra; the caller owns the client's configuration.
+            logger.debug("bucket=<%s> | reusing caller-supplied s3 client", bucket)
+            self.client = s3_client
```
zastrowm, line 98: Can we throw if any of the arguments are passed in (region_name or boto_session)? 

In all those cases, it would be a mistake and I'd rather fail fast.

gautamsirdeshmukh (dismissed):

strands-py/src/strands/session/s3_session_manager.py
```diff
@@ -63,31 +65,45 @@ def __init__(
...
-        # Add strands-agents to the request user agent
-        if boto_client_config:
-            existing_user_agent = getattr(boto_client_config, "user_agent_extra", None)
-            # Append 'strands-agents' to existing user_agent_extra or set it if not present
-            if existing_user_agent:
-                new_user_agent = f"{existing_user_agent} strands-agents"
-            else:
-                new_user_agent = "strands-agents"
-            client_config = boto_client_config.merge(BotocoreConfig(user_agent_extra=new_user_agent))
+        if s3_client is not None:
```
zastrowm, line 89: I presume there's no way to add our user-agent in this case, could we?

JackYPCOnline: nope, unless we modified client I think, if they choose to use pre-configured client, we should not handle it for them I guess.

strands-py/src/strands/session/s3_session_manager.py
JackYPCOnline: We can. To me the question is a design one: should we prevent callers from mixing s3_client with the other config args by failing fast, or accept the mix and just document that s3_client takes precedence?
Not sure the best practice is, will update tho

JackYPCOnline:
Close this PR given the context we are deprecating S3sessionManager, this issue is fixed in SessionManager + unified S3 Storage.

Issue #2204, opened 2026-04-24
[BUG] Media type handling bugs and missing format support


opieter-aws:
### Steps to Reproduce

**Bugs:**

* WebP images get wrong MIME type with Anthropic provider. `anthropic.py:132` uses `mimetypes.types_map.get(f".{content['image']['format']}", "application/octet-stream")`. Python's `mimetypes` module returns `"application/octet-stream"` for `.webp`, not `"image/webp"`. WebP images will fail or be misprocessed. TypeScript uses an explicit switch statement that correctly maps webp.

* GuardContent wraps unsupported image formats and missing image type support.`bedrock.py:450-451` wraps ALL image formats in `guardContent` without format validation. TypeScript explicitly checks for `png`/`jpeg` only and skips unsupported formats. Additionally, `GuardContent` type (`content.py`) only has a `"text"` field — missing `"image"` support that TypeScript has. Wrapping unsupported formats will cause Bedrock API errors.

* Video format "3gp" not mapped to "three_gp" for Bedrock API. `bedrock.py:647` passes video format as-is. TypeScript maps `"3gp"` → `"three_gp"` to match the Bedrock API enum. 3GP videos will be rejected by Bedrock.

**Missing format support (non-breaking gaps):**

- `DocumentFormat` missing `"json"` and `"xml"` (TypeScript has them)
- `ImageFormat` missing `"jpg"` alias for jpeg (TypeScript has it)
- `ToolResultContent` missing video support (TypeScript has `VideoBlock`)
- `DocumentSource` missing `{ text: string }` and `{ content: [...] }` source types (TypeScript has them)

### Expected Behavior

1. WebP images should get `"image/webp"` MIME type
2. GuardContent should only wrap supported image formats (png, jpeg) and support image type
3. "3gp" video format should be mapped to "three_gp"
4. Format types should support json, xml documents; jpg alias; video in tool results; text document sources

### Actual Behavior

1. WebP → `"application/octet-stream"` → Anthropic API error
2. Unsupported image formats wrapped in guardContent → Bedrock API error
3. "3gp" sent as-is → Bedrock API error
4. Missing format types limit what content can be represented

### Additional context

This was a finding from an agentic cross-SDK consistency scan between Python and TypeScript SDKs (https://github.com/strands-agents/sdk-typescript/issues/789). The finding has been triaged, but verify that this needs an actual fix.

kimnamu:
[...]

PR #2304, opened 2026-05-19
fix(anthropic): map webp images explicitly

he-yufeng:
Fixes the Anthropic WebP part of #2204.

The Anthropic formatter was relying on `mimetypes.types_map` for image media types. That map is platform/Python-install dependent, so `.webp` can fall back to `application/octet-stream` even though Strands already accepts `webp` as an image format.

This keeps a small explicit map for the Anthropic image formats Strands supports and falls back to `mimetypes` only for unknown formats.

Validation:
- `python -m pip install -e ".[anthropic]"`
- `python -m pytest tests/strands/models/test_anthropic.py -q`
- `python -m ruff check src/strands/models/anthropic.py tests/strands/models/test_anthropic.py`
- `python -m ruff format --check src/strands/models/anthropic.py tests/strands/models/test_anthropic.py`
- `git diff --check`

### DIFF
--- a/strands-py/src/strands/models/anthropic.py
+++ b/strands-py/src/strands/models/anthropic.py
@@ -28,7 +28,7 @@ logger = logging.getLogger(__name__)
 
 T = TypeVar("T", bound=BaseModel)
 
-IMAGE_MEDIA_TYPES = {
+_IMAGE_MEDIA_TYPES = {
     "gif": "image/gif",
     "jpeg": "image/jpeg",
     "jpg": "image/jpeg",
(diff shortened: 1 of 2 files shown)


he-yufeng:
Rebased this onto current main and revalidated the Anthropic model path. Current head is 2ba48cb. Validation from strands-py: uv sync --extra dev --extra anthropic; uv run pytest tests/strands/models/test_anthropic.py -q --basetemp ..\.tmp\pytest-2304 -p no:cacheprovider: 58 passed; uv run ruff check src/strands/models/anthropic.py tests/strands/models/test_anthropic.py; uv run ruff format --check src/strands/models/anthropic.py tests/strands/models/test_anthropic.py; python -m py_compile strands-py\src\strands\models\anthropic.py strands-py\tests\strands\models\test_anthropic.py; git diff --check origin/main...HEAD.

mehtarac (approved):

pgrayy (approved):

PR #2305, opened 2026-05-19
feat: pass invocation_state to edge condition calls

### LINKED ISSUE #1346: [FEATURE] Pass invocation_state to edge condition call
## Overview

Add support for passing `invocation_state` to edge condition functions in Graph execution, enabling conditional edge traversal based on runtime context provided during graph invocation.

## Problem Statement

Currently, edge condition functions only have access to `GraphState`, limiting their ability to make decisions based on runtime context passed during graph invocation. This makes it difficult to conditionally enable nodes based on the `invocation_state` received in the graph invocation.

## Implementation Requirements

Based on discussion and repository analysis, the implementation should use a **Protocol-based Union type** for backwards compatibility and future extensibility.

#### 1. Define Edge Condition Protocol

Create a new Protocol in `src/strands/multiagent/graph.py` that allows future expansion with kwargs:

```python
from typing import Protocol, runtime_checkable

@runtime_checkable
class EdgeConditionWithContext(Protocol):
    """Protocol for edge conditions that receive invocation context.
    
    This protocol allows conditions to access both GraphState and invocation_state,
    and is designed to be expanded with additional kwargs in the future.
    """
    def __call__(
        self, 
        state: GraphState, 
        *, 
        invocation_state: dict[str, Any] | None = None,
        **kwargs: Any
    ) -> bool: ...

# Legacy condition type (backwards compatible)
LegacyEdgeCondition = Callable[[GraphState], bool]

# Union type supporting both old and new signatures
EdgeCondition = LegacyEdgeCondition | EdgeConditionWithContext
```

#### 2. Update GraphEdge Dataclass

```python
@dataclass
class GraphEdge:
        ...
        if new style:
            return self.condition(state, invocation_state=invocation_state)
        
        # Legacy: single positional parameter only
        return self.condition(state)
```

#### 3. Update Call Sites

Update all locations that call `should_traverse()` to pass `invocation_state`:

- `Graph._is_node_ready_with_conditions()` (line ~837)
- `Graph._compute_ready_nodes_for_resume()` (line ~1190)
- `Graph._build_node_input()` (line ~1080)

#### 4. Update GraphBuilder Type Hint

```python
def add_edge(
    self,
    from_node: str | GraphNode,
    to_node: str | GraphNode,
    condition: EdgeCondition | None = None,
) -> GraphEdge:
```

### Files to Modify

| File | Changes |
|------|---------|
| `src/strands/multiagent/graph.py` | Add Protocol, update types, update `should_traverse()`, update all call sites |
| `tests/strands/multiagent/test_graph.py` | Add tests for new-style conditions with invocation_state |
| `tests_integ/test_multiagent_graph.py` | Add integration tests for conditions with invocation_state |

### Backwards Compatibility

- **Old-style conditions** (`def condition(state: GraphState) -> bool`) continue to work unchanged
- **New-style conditions** (`def condition(state: GraphState, *, invocation_state=None, **kwargs) -> bool`) receive invocation context
- Runtime signature inspection ensures correct calling convention
- No deprecation warnings - both styles supported indefinitely

### Example Usage

```python
# Old-style (still works)
def check_completion(state: GraphState) -> bool:
    return "data_processor" in {n.node_id for n in state.completed_nodes}

# New-style (with invocation context)
def should_run_premium_node(
    state: GraphState, 
    *, 
    invocation_state: dict[str, Any] | None = None,
    **kwargs: Any
) -> bool:
    if invocation_state is None:
        return False
    return invocation_state.get("user_tier") == "premium"

# Graph setup
builder = GraphBuilder()
builder.add_node(start_agent, "start")
builder.add_node(premium_agent, "premium_processor")
builder.add_edge("start", "premium_processor", condition=should_run_premium_node)
graph = builder.build()

# Invocation with context
result = graph(task="process data", invocation_state={"user_tier": "premium"})
```

### Acceptance Criteria

- [ ] `EdgeConditionWithContext` Protocol defined with kwargs support for future expansion
- [ ] `EdgeCondition` Union type supports both legacy and new condition signatures
- [ ] `GraphEdge.should_traverse()` accepts and passes `invocation_state`
- [ ] All call sites updated to pass `invocation_state` through the execution flow
- [ ] Existing tests pass without modification (backwards compatibility)
- [ ] New unit tests for conditions that use `invocation_state`
- [ ] New integration test demonstrating the use case
- [ ] Type hints pass mypy strict mode

### Implementation Notes

- Use `inspect.signature()` for runtime detection of condition signature
- The Protocol approach allows adding more context parameters in the future without breaking changes
- Consider caching signature inspection results for performance if needed

yananym:
## Description
Add support for edge conditions that receive `invocation_state`, enabling conditional routing based on runtime context (feature flags, user roles, environment config) passed during
  graph invocation.

  Also fixes a deadlock in `_compute_ready_nodes_for_resume()` where conditional edges evaluating to `False` would block downstream nodes from ever becoming ready on interrupt/resume  workflows.

### Session persistence of `invocation_state`

When a `SessionManager` is configured, `invocation_state` is serialized alongside graph state so that resumed executions retain the same runtime context. This is validated as JSON-serializable on entry.

**Example**: A graph with a feature-flag condition:
```
def route_if_premium(state, *, invocation_state, **kwargs):
return invocation_state.get("user_tier") == "premium"

graph(task="process order", invocation_state={"user_tier": "premium"})
```
If the graph is interrupted and later resumed from session, `invocation_state` is restored so `route_if_premium` still evaluates correctly without the caller needing to re-supply it.

  ### Public API Changes

  `GraphBuilder.add_edge()` now accepts conditions with an extended signature via the `EdgeConditionWithContext` Protocol:

  ```py
  # Legacy (still works, no changes needed)
  def my_condition(state: GraphState) -> bool:
      return len(state.completed_nodes) > 0

  # New: receives invocation_state for runtime routing decisions
  def my_condition(state: GraphState, *, invocation_state: dict, **kwargs) -> bool:
      return invocation_state.get("role") == "admin"

  builder.add_edge("router", "admin_path", condition=my_condition)
  graph("task", invocation_state={"role": "admin"})
```
  Both signatures are supported indefinitely — no deprecation, no breaking changes.

## Related Issues

Resolves #1346

## Documentation PR

https://github.com/strands-agents/docs/pull/847

## Type of Change

New feature

## Testing

  - 21 new unit tests covering protocol dispatch, invocation_state propagation, resume deadlock fix, and serialization round-trip
  - 6 new integration tests covering conditional routing, backwards compat, combined conditions, diamond convergence, persistence/resume, and streaming events
  - All 49 existing graph unit tests pass unchanged (backwards compatibility verified)
  - 95% line+branch coverage on src/strands/multiagent/graph.py
  - [x] ruff check and mypy pass clean

## Checklist

----

### DIFF
--- a/strands-py/src/strands/multiagent/__init__.py
+++ b/strands-py/src/strands/multiagent/__init__.py
@@ -9,10 +9,12 @@ Submodules:
 """
 
 from .base import MultiAgentBase, MultiAgentResult, Status
-from .graph import GraphBuilder, GraphResult
+from .graph import EdgeCondition, EdgeConditionWithContext, GraphBuilder, GraphResult
 from .swarm import Swarm, SwarmResult
 
 __all__ = [
+    "EdgeCondition",
+    "EdgeConditionWithContext",
     "GraphBuilder",
     "GraphResult",
     "MultiAgentBase",
--- a/strands-py/src/strands/multiagent/graph.py
+++ b/strands-py/src/strands/multiagent/graph.py
@@ -17,11 +17,13 @@ Key Features:
 import asyncio
 import copy
 import inspect
+import json
 import logging
 import time
+import weakref
 from collections.abc import AsyncIterator, Callable, Mapping
 from dataclasses import dataclass, field
-from typing import Any, Protocol, TypeGuard, cast, runtime_checkable
+from typing import Any, Protocol, TypeGuard, cast
 
 from opentelemetry import trace as trace_api
 
@@ -63,7 +65,6 @@ logger = logging.getLogger(__name__)
 _DEFAULT_GRAPH_ID = "default_graph"
 
 
-@runtime_checkable
 class EdgeConditionWithContext(Protocol):
     """Protocol for edge conditions that receive invocation_state.
 
@@ -72,6 +73,9 @@ class EdgeConditionWithContext(Protocol):
     environment-specific configuration.
 
     Designed with **kwargs for future extensibility without breaking changes.
+
+    Note: not @runtime_checkable — isinstance() cannot distinguish callable signatures
+    structurally; use _is_context_condition() for dispatch instead.
     """
 
     def __call__(self, state: "GraphState", *, invocation_state: dict[str, Any], **kwargs: Any) -> bool:
@@ -82,18 +86,34 @@ class EdgeConditionWithContext(Protocol):
 LegacyEdgeCondition = Callable[["GraphState"], bool]
 EdgeCondition = LegacyEdgeCondition | EdgeConditionWithContext
 
+# GIL-protected: concurrent async graphs may read/write simultaneously, but
+# CPython's GIL ensures dict mutation is atomic.  Ephemeral callables (e.g.
+# lambdas recreated per-call) will bypass the cache — this is benign; the
+# fallback path is a single inspect.signature() call.
+_context_condition_cache: weakref.WeakKeyDictionary[EdgeCondition, bool] = weakref.WeakKeyDictionary()
+
 
 def _is_context_condition(condition: EdgeCondition) -> TypeGuard[EdgeConditionWithContext]:
     """Check if a condition function accepts invocation_state parameter.
 
     Uses inspect.signature() for reliable detection, returning a TypeGuard
-    so mypy can narrow the type at call sites.
+    so mypy can narrow the type at call sites. Results are cached per condition
+    using weak references so entries are evicted when the function is collected.
     """
     try:
+        return _context_condition_cache[condition]
+    except (KeyError, TypeError):
+        pass
+    try:
         sig = inspect.signature(condition)
-        return "invocation_state" in sig.parameters
+        result = "invocation_state" in sig.parameters
     except (ValueError, TypeError):
-        return False
+        result = False
+    try:
+        _context_condition_cache[condition] = result
+    except TypeError:
+        pass
+    return result
 
 
 @dataclass
@@ -620,6 +640,8 @@ class Graph(MultiAgentBase):
         if invocation_state is None:
             invocation_state = {}
 
+        if self.session_manager is not None:
+            self._validate_invocation_state(invocation_state)
         self._current_invocation_state = invocation_state
 
         await self.hooks.invoke_callbacks_async(BeforeMultiAgentInvocationEvent(self, invocation_state))
@@ -1239,6 +1261,20 @@ class Graph(MultiAgentBase):
             interrupts=interrupts,
         )
 
+    @staticmethod
+    def _validate_invocation_state(invocation_state: dict[str, Any]) -> None:
+        """Validate that invocation_state is JSON-serializable.
+
+        Raises:
+            TypeError: If invocation_state contains non-JSON-serializable values.
+        """
+        try:
+            json.dumps(invocation_state)
+        except (TypeError, ValueError) as e:
+            raise TypeError(
+                f"invocation_state must be JSON-serializable for session persistence: {e}"
+            ) from e
+
     def serialize_state(self) -> dict[str, Any]:
         """Serialize the current graph state to a dictionary."""
         compute_nodes = self._compute_ready_nodes_for_resume()
@@ -1277,7 +1313,9 @@ class Graph(MultiAgentBase):
             internal_state = payload["_internal_state"]
             self._interrupt_state = _InterruptState.from_dict(internal_state["interrupt_state"])
 
-        self._current_invocation_state = payload.get("invocation_state", {})
+        invocation_state = payload.get("invocation_state", {})
+        self._validate_invocation_state(invocation_state)
+        self._current_invocation_state = invocation_state
 
         if not payload.get("next_nodes_to_execute"):
             # Reset all nodes
@@ -1318,10 +1356,16 @@ class Graph(MultiAgentBase):
         A node is ready if all TRAVERSABLE incoming edges have their source completed.
         Edges whose condition evaluates to False are excluded from the check — they
         represent paths that were intentionally skipped.
+
+        Re-evaluates conditions (rather than caching traversal results) intentionally:
+        invocation_state may change between invocations, so conditions must reflect
+        current runtime context. This means condition logic changes between serialize
+        and resume will also take effect — consistent with the graph being defined in code.
         """
         traversable_edges = [
             e
             for e in incoming
+            # Short-circuit: skip signature inspection + cache lookup for unconditional edges.
             if e.condition is None or e.should_traverse(self.state, invocation_state=self._current_invocation_state)
         ]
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: strands-py/src/strands/multiagent/graph.py [hunk 1]
```diff
@@ -1201,6 +1254,7 @@ def serialize_state(self) -> dict[str, Any]:
             "next_nodes_to_execute": next_nodes,
             "current_task": encode_bytes_values(self.state.task),
             "execution_order": [n.node_id for n in self.state.execution_order],
+            "invocation_state": self._current_invocation_state,
```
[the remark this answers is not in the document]
yananym: Addressed in `7cc0d04` — added `_validate_invocation_state()` that calls `json.dumps()` and raises a clear `TypeError` if the value isn't serializable. It's gated on `session_manager` is not `None` (serialization only matters when sessions persist), and also validated symmetrically on the deserialization path in `deserialize_state`.

strands-py/src/strands/multiagent/graph.py
```diff
@@ -1246,11 +1302,34 @@ def _compute_ready_nodes_for_resume(self) -> list[GraphNode]:
...
+        incoming: list[GraphEdge],
+        completed_nodes: set[GraphNode],
+    ) -> bool:
+        """Check if a node is ready for resume, accounting for conditional edges.
+
+        A node is ready if all TRAVERSABLE incoming edges have their source completed.
+        Edges whose condition evaluates to False are excluded from the check — they
+        represent paths that were intentionally skipped.
+        """
+        traversable_edges = [
```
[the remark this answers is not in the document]
yananym: Addressed in `7cc0d04`. You're right that with the `WeakKeyDictionary` cache now in place, the `e.condition` is `None` short-circuit is no longer necessary for performance. I kept it because it still avoids the method call + cache lookup entirely for unconditional edges, and makes the "unconditional edges always traverse" semantics explicit at the call site. Added an inline comment explaining the intent. Happy to simplify if you'd prefer the terser form.
On the serialize/deserialize behavior divergence — added a docstring to `_compute_ready_nodes_for_resume` acknowledging this explicitly: re-evaluation is intentional since  `invocation_state` may differ between invocations, and condition logic changes taking effect on resume is consistent with the graph being defined in code rather than serialized.

yananym:
> **Issue**: The PR introduces a public API change (`GraphBuilder.add_edge()` accepts a new condition signature, new types `EdgeCondition` and `EdgeConditionWithContext` are exported) but doesn't appear to have a `needs-api-review` label. Per the API bar-raising process, new public abstractions customers will use should be flagged for API review.
> 
> **Suggestion**: Consider adding the `needs-api-review` label. In particular, the API reviewer should evaluate:

 I do not think i have permissions to add labels to this PR

src/strands/multiagent/graph.py
```diff
@@ -62,6 +65,57 @@
...
+class EdgeConditionWithContext(Protocol):
+    """Protocol for edge conditions that receive invocation_state.
+
+    This allows conditions to make routing decisions based on runtime context
+    passed during graph invocation, such as feature flags, user roles, or
+    environment-specific configuration.
+
+    Designed with **kwargs for future extensibility without breaking changes.
+
+    Note: not @runtime_checkable — isinstance() cannot distinguish callable signatures
```
zastrowm: I presume it's not `runtime_checkable` because the expected use case is a function/lambda?  If so, it would be good to call that out.

src/strands/multiagent/graph.py
```diff
@@ -62,6 +65,57 @@
...
+
+
+LegacyEdgeCondition = Callable[["GraphState"], bool]
+EdgeCondition = LegacyEdgeCondition | EdgeConditionWithContext
+
+# GIL-protected: concurrent async graphs may read/write simultaneously, but
+# CPython's GIL ensures dict mutation is atomic.  Ephemeral callables (e.g.
+# lambdas recreated per-call) will bypass the cache — this is benign; the
+# fallback path is a single inspect.signature() call.
+_context_condition_cache: weakref.WeakKeyDictionary[EdgeCondition, bool] = weakref.WeakKeyDictionary()
```
zastrowm: Can we cache this in the `GraphEdge` itself ditching the weakref lookup?

strands-py/src/strands/multiagent/graph.py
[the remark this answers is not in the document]
zastrowm: > if the graph code changes between serialize and deserialize (e.g., condition logic updated), the behavior could be subtly different on resume. This is inherent to the design but worth noting.

Yeah, in general this is a "Don't do that situation" and so is by-design

strands-py/src/strands/multiagent/__init__.py
zastrowm, line 1: Please update the PR description with notes about the invocation_args being persisted to session-management and the reasoning (example would be useful)

src/strands/multiagent/graph.py
zastrowm: Alternatively, do we think we should normalize call conditions to be EdgeConditionWithContext and we just forward one without the argument?  Is that a way to do that?

I think it'd be a minor breaking change, but I think I think it'd be okay?

src/strands/multiagent/graph.py
yananym: Exactly right. Updated the `docstring` to call this out explicitly — `isinstance()` can't structurally distinguish callable signatures, so functions/lambdas would never match even if it were `@runtime_checkable`.

src/strands/multiagent/graph.py
yananym: Done — moved the cache to a `_is_context_condition_cached` field on `GraphEdge`. Computed lazily on first `should_traverse()` call. Removed the module-level `WeakKeyDictionary` entirely.

I'm hesitant to introduce a breaking change in this PR. Going with caching on `GraphEdge` for now. We can revisit  normalizing to a single condition type as a follow-up if we want to clean up the union.

src/strands/multiagent/graph.py
```diff
@@ -62,6 +63,39 @@
...
+
+def _is_context_condition(condition: EdgeCondition) -> TypeGuard[EdgeConditionWithContext]:
+    """Check if a condition function accepts invocation_state parameter.
+
+    Uses inspect.signature() for reliable detection, returning a TypeGuard
+    so mypy can narrow the type at call sites.
+    """
+    try:
+        sig = inspect.signature(condition)
+        return "invocation_state" in sig.parameters
```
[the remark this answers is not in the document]
yananym: Addressed — moved to a `lazy _is_context_condition_cached` field on `GraphEdge`, no more per-call  `inspect.signature()`. See thread on comment from @zastrowm.

src/strands/multiagent/graph.py
```diff
@@ -62,6 +63,39 @@
 _DEFAULT_GRAPH_ID = "default_graph"
 
 
+@runtime_checkable
```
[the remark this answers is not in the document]
yananym: Addressed — added a docstring note explaining why `@runtime_checkable` is retained for `type-checker` 
 support but `isinstance()` isn't used for dispatch (can't distinguish callable signatures structurally).

zastrowm (approved):

PR #2307, opened 2026-05-20
* Cutting of agentic loop based on max_iterations configured.

### LINKED ISSUE #2298: [FEATURE]: Add max_iterations control to the Agent execution loop
### Problem Statement

Problem Description
Currently, when using the Agent class in the Strands SDK, the agent operates in a continuous loop until the LLM decides to provide a final response. While this allows for complex reasoning, it creates two significant risks in a production environment:

Infinite Reasoning Loops: If an agent encounters a tool error or a confusing prompt, it may enter a "hallucination loop," calling tools repeatedly without reaching a conclusion.

Unpredictable Latency/Cost: Without an explicit depth limit, a single user request could potentially trigger dozens of model turns, leading to high token costs and poor user experience.

### Proposed Solution

I propose adding a max_iterations (or max_depth) parameter to the Agent configuration and its execution methods. This would track the number of "Model Turn -> Tool Call -> Tool Result" cycles and gracefully terminate if the limit is reached.

Suggested changes:

Add max_iterations: int = 10 to the Agent constructor or the run() method.

Implement a counter within the execution loop (likely in agent.py).

Raise a specific AgentLimitReachedException or return a partial response when the limit is hit, allowing developers to handle the timeout gracefully.

### Use Case

Example Usage

agent = Agent(
    tools=[my_tool],
    max_iterations=5  # Agent will stop after 5 reasoning cycles
)

try:
    response = agent.run("Perform a complex multi-step task")
except MaxIterationsReachedError:
    print("Agent was unable to finish within the allowed steps.")

### Alternatives Solutions

_No response_

### Additional Context

_No response_

printSamarth:
## Description
Adds a max_iterations configuration to Agent that bounds the number of "model turn → tool call → tool result" cycles per invocation, with a dedicated MaxIterationsReachedException raised when the limit is exceeded.

Motivation. Today the event loop runs until the LLM decides to stop. In production this exposes two risks:

Infinite reasoning loops — a tool error or confusing prompt can send the model into a hallucination loop, calling tools repeatedly without ever reaching end_turn.
Unpredictable latency / cost — without an explicit depth limit, a single user request can trigger many model turns and burn unbounded tokens.

API:
```python
agent = Agent(tools=[my_tool], max_iterations=5)
try:
    response = agent("Perform a complex multi-step task")
except MaxIterationsReachedException as e:
    log.warning("agent stopped", iterations=e.iterations, limit=e.max_iterations)
```

max_iterations is None by default (preserves existing behavior). Positive integers only; 0, negatives, non-ints, and booleans are rejected at construction time.

Implementation.

New MaxIterationsReachedException in strands.types.exceptions carrying iterations and max_iterations for structured logging.
New max_iterations constructor kwarg on Agent, validated to be a positive int or None.
The event loop tracks invocation_state["event_loop_iteration"] and the guard is enforced at the top of event_loop_cycle, before the next model call — so an over-limit cycle never consumes an additional model invocation. The exception is added to the bubble-up list so it isn't wrapped in EventLoopException.
Iteration semantics. One iteration = one model turn (not one tool call). Parallel tool use counts as a single iteration. This matches what the cap is intended to bound (reasoning depth / API calls), and avoids penalizing models that parallelize independent tool calls.

## Related Issues

Resolves [2298](https://github.com/strands-agents/sdk-python/issues/2298)

## Documentation PR

N/A : happy to follow up with a docs PR once this merges.

## Type of Change

New feature

## Testing

Added the following coverage:

tests/strands/types/test_exceptions.py : construction, inheritance, and metadata propagation for MaxIterationsReachedException.
tests/strands/agent/test_agent.py :  constructor validation (default None, positive int, rejection of 0 / -1 / 1.5 / "5" / True), plus end-to-end tests through the public API using MockedModelProvider:
runaway tool loop is bounded and raises with correct metadata
normal completion succeeds when within the budget
iteration counter resets between separate invocations
tests/strands/event_loop/test_event_loop.py : limit exceeded raises before the next model call (model.stream.call_count asserted), within-limit completes normally, None disables the check, and a pre-set counter trips the guard on the very first cycle.
A small number of existing tests that pinned the exact shape of invocation_state keys passed to callbacks were updated to include the new event_loop_iteration key consistent with how event_loop_cycle_id is already exposed.

Apart from this manualy tested by integrating it into a local project to verify end-to-end behavior.

## Checklist

----

### DIFF
--- a/strands-py/src/strands/agent/agent.py
+++ b/strands-py/src/strands/agent/agent.py
@@ -146,6 +146,7 @@ class Agent(AgentBase):
         tool_executor: ToolExecutor | None = None,
         retry_strategy: ModelRetryStrategy | _DefaultRetryStrategySentinel | None = _DEFAULT_RETRY_STRATEGY,
         concurrent_invocation_mode: ConcurrentInvocationMode = ConcurrentInvocationMode.THROW,
+        max_iterations: int | None = None,
     ):
         """Initialize the Agent with the specified configuration.
 
@@ -214,9 +215,13 @@ class Agent(AgentBase):
                 Set to "unsafe_reentrant" to skip lock acquisition entirely, allowing concurrent invocations.
                 Warning: "unsafe_reentrant" makes no guarantees about resulting behavior and is provided
                 only for advanced use cases where the caller understands the risks.
+            max_iterations: Maximum number of model invocation cycles allowed per
+                invocation. When the limit is exceeded the event loop is terminated and a
+                ``MaxIterationsReachedException`` is raised, protecting against runaway tool-calling loops
+                and unbounded latency/cost. ``None`` (the default) disables the limit.
 
         Raises:
-            ValueError: If agent id contains path separators.
+            ValueError: If agent id contains path separators or if ``max_iterations`` is not a positive integer.
         """
         self.model = BedrockModel() if not model else BedrockModel(model_id=model) if isinstance(model, str) else model
         self.messages = messages if messages is not None else []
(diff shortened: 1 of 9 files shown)


chaynabors:
This feature was added in merged PR: https://github.com/strands-agents/harness-sdk/pull/2360

Issue #2308, opened 2026-05-21
[FEATURE] Make BedrockModel._format_request and _convert_non_streaming_to_streaming public


celealramos:
### Problem Statement

Currently, BedrockModel._format_request and BedrockModel._convert_non_streaming_to_streaming are private methods (prefixed with _). Users who subclass BedrockModel or build custom model wrappers around it cannot reliably call or override these methods without depending on private API that may change without notice. This makes it difficult to:

- Re-use methods without the risk that method signatures may cause breaking changes

### Proposed Solution

Rename the private methods to public:

```_format_request``` becomes ```format_request```
```_convert_non_streaming_to_streaming``` becomes  ```convert_non_streaming_to_streaming```
Keep the old private method names as deprecated aliases that:

- Emit a DeprecationWarning directing users to the new public method name
- Keep calling private methods internally in order to avoid breaking functionallity in subclasses that override the private methods
- Are scheduled for removal in September 2026
- This provides a clean migration path for any existing code referencing the private methods

### Use Case

Subclassing BedrockModel: users who extend BedrockModel to add custom request formatting (e.g., injecting guardrails, adding metadata, transforming tool specs) can now override format_request as a stable public API.

### Alternatives Solutions

- Remove the deprecated aliases immediately — this would break existing code that references _format_request or _convert_non_streaming_to_streaming. The deprecation period provides a graceful migration.

### Additional Context

Other model providers in the SDK (OpenAIModel, AnthropicModel, OllamaModel, LiteLLMModel, MistralModel, LlamaAPIModel, SageMakerModel, WriterModel) already expose format_request as a public method. BedrockModel is the only provider where this was private. This change brings it in line with the rest of the SDK.

The internal call sites within BedrockModel (in converse and count_tokens_using_bedrock) are updated to use the new public names directly. Tests are updated accordingly and new tests verify that the deprecated aliases emit warnings and correctly delegate.

zastrowm:
@celealramos - the overrides that you'd like to add in your subclass keep bedrock compatibility, correct?  One of the initial reasons we kept them private was to allow us evolve them going forward without concern for implementors.  If we expose them, we'll want the documentation to indicate that we're free to modify the behavior going forward as Bedrock evolves.

celealramos:
Yes whether we add overrides in the subclass or we just call the methods in the parent classes within the subclass- we'd be keeping bedrock compatibility.

[...]

PR #2315, opened 2026-05-21
feat: make BedrockModel._format_request and _convert_non_streaming_to…

celealramos:
…_streaming public

## Description

Users building custom Bedrock integrations need to format requests and convert non-streaming responses outside of the standard stream() flow. For example, to inspect or modify the request before sending, or to integrate with custom invocation pipelines. Currently these capabilities are only available through private methods (_format_request, _convert_non_streaming_to_streaming), making them fragile to depend on and inconsistent with other providers (Anthropic, Mistral, Ollama, Writer) that already expose public format_request.

Public API Changes
BedrockModel 's  _format_request and _convert_non_streaming_to_streaming are now public

Before: rely on private API (fragile)

```
model = BedrockModel(model_id="us.anthropic.claude-sonnet-4-20250514-v1:0")
request = model._format_request(messages, tool_specs)
events = model._convert_non_streaming_to_streaming(response)
```

After: stable public API

```
model = BedrockModel(model_id="us.anthropic.claude-sonnet-4-20250514-v1:0")
request = model.format_request(messages, tool_specs)
events = model.convert_non_streaming_to_streaming(response)
```

Internal call sites now call public methods 
subclass overrides are honored without behavioral changes.
format_request and convert_non_streaming_to_streaming accept kwargs parameters to enable additive parameters without breaking subclass overrides.

## Related Issues

[#2308](https://github.com/strands-agents/sdk-python/issues/2308)

## Type of Change
New feature

## Testing

## Checklist
- [ x] I have read the CONTRIBUTING document
- [ x] I have added any necessary tests that prove my fix is effective or my feature works
- [x ] I have updated the documentation accordingly
- [ x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x ] Any dependent changes have been merged and published

----

### DIFF
--- a/strands-py/src/strands/models/bedrock.py
+++ b/strands-py/src/strands/models/bedrock.py
@@ -234,12 +234,13 @@ class BedrockModel(Model):
         """
         return resolve_config_metadata(self.config, self.config.get("model_id", ""))
 
-    def _format_request(
+    def format_request(
         self,
         messages: Messages,
         tool_specs: list[ToolSpec] | None = None,
         system_prompt_content: list[SystemContentBlock] | None = None,
         tool_choice: ToolChoice | None = None,
+        **kwargs: Any,
     ) -> dict[str, Any]:
         """Format a Bedrock converse stream request.
 
@@ -248,6 +249,7 @@ class BedrockModel(Model):
             tool_specs: List of tool specifications to make available to the model.
             tool_choice: Selection strategy for tool invocation.
             system_prompt_content: System prompt content blocks to provide context to the model.
+            **kwargs: Additional keyword arguments for future extensibility.
 
         Returns:
             A Bedrock converse stream request.
@@ -830,7 +832,7 @@ class BedrockModel(Model):
             if system_prompt and system_prompt_content is None:
                 system_prompt_content = [{"text": system_prompt}]
 
-            request = self._format_request(messages, tool_specs, system_prompt_content)
+            request = self.format_request(messages, tool_specs, system_prompt_content)
             converse_input: dict[str, Any] = {}
             if "messages" in request:
                 converse_input["messages"] = request["messages"]
@@ -960,7 +962,7 @@ class BedrockModel(Model):
         """
         try:
             logger.debug("formatting request")
-            request = self._format_request(messages, tool_specs, system_prompt_content, tool_choice)
+            request = self.format_request(messages, tool_specs, system_prompt_content, tool_choice)
             logger.debug("request=<%s>", request)
 
             logger.debug("invoking model")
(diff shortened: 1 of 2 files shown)


strands-py/src/strands/models/bedrock.py
```diff
@@ -243,6 +265,9 @@ def _format_request(
     ) -> dict[str, Any]:
         """Format a Bedrock converse stream request.
 
+        .. deprecated::
+            Use :meth:`format_request` instead. This will be removed in September 2026.
```
zastrowm: Don't call out a specific date; it's very subject to change.

strands-py/src/strands/models/bedrock.py
```diff
@@ -234,6 +234,28 @@ def get_config(self) -> BedrockConfig:
...
+        Args:
+            messages: List of message objects to be processed by the model.
+            tool_specs: List of tool specifications to make available to the model.
+            tool_choice: Selection strategy for tool invocation.
+            system_prompt_content: System prompt content blocks to provide context to the model.
+
+        Returns:
+            A Bedrock converse stream request.
+        """
+        with warnings.catch_warnings():
```
zastrowm: The flow is a little odd right now; could we switch to:

 - `format_request` is the entry point and what the model calls internally
 - `format_request` checks if `_format_request` is overridden; if it is it emits a warning and calls it, otherwise it invokes `_format_request_impl` (which is the default implementation) which calls `_format_request`

strands-py/src/strands/models/bedrock.py
```diff
@@ -234,6 +234,28 @@ def get_config(self) -> BedrockConfig:
         """
         return resolve_config_metadata(self.config, self.config.get("model_id", ""))
 
+    def format_request(
+        self,
+        messages: Messages,
+        tool_specs: list[ToolSpec] | None = None,
+        system_prompt_content: list[SystemContentBlock] | None = None,
+        tool_choice: ToolChoice | None = None,
+    ) -> dict[str, Any]:
```
zastrowm, line 244: Let's add kwargs here

zastrowm (changes requested):

strands-py/src/strands/models/bedrock.py
zastrowm: The changes in https://github.com/strands-agents/sdk-python/pull/2093/changes#r3306308887 is an example of an additive parameter FWIW

strands-py/src/strands/models/bedrock.py
zastrowm: I think we're also okay breaking users of _ as well given that it's marked as private.

That's what we went with for https://github.com/strands-agents/sdk-python/pull/2370

celealramos: Got it - if we're okay breaking users of _ as it is private then we can go ahead and make the switch from private to public  from the getgo

zastrowm (approved):

PR #2326, opened 2026-05-25
fix: prevent in-place mutation of message history in normalizer

pipe1os:
## Description
Fixes a state corruption bug where `_normalize_messages` mutated the `agent.messages` history in place during streaming. This permanently altered the canonical conversation state before it reached the session manager. Changed the normalization loop to build shallow copies so the original message history remains pristine.

## Related Issues
N/A

## Documentation PR
N/A

## Type of Change
Bug fix

## Testing

## Checklist
- [x] I have updated the documentation accordingly - N/A, internal bug fix
- [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed - N/A, internal bug fix
- [x] My changes do not introduce any new warnings

---

### DIFF
--- a/strands-py/src/strands/event_loop/streaming.py
+++ b/strands-py/src/strands/event_loop/streaming.py
@@ -6,7 +6,7 @@ import threading
 import time
 import warnings
 from collections.abc import AsyncGenerator, AsyncIterable
-from typing import Any
+from typing import Any, cast
 
 from ..models.model import Model
 from ..tools import InvalidToolUseNameException
@@ -56,21 +56,29 @@ def _normalize_messages(messages: Messages) -> Messages:
     replaced_blank_message_content_text = False
     replaced_tool_names = False
 
+    new_messages: Messages = []
+
     for message in messages:
         # only modify assistant messages
         if "role" in message and message["role"] != "assistant":
+            new_messages.append(message)
             continue
         if "content" in message:
-            content = message["content"]
+            # Create shallow copies to prevent mutating the original history
+            new_message = cast(Message, dict(message))
+            content = list(message["content"])
+            new_message["content"] = content
+
             if len(content) == 0:
                 content.append({"text": "[blank text]"})
+                new_messages.append(new_message)
                 continue
 
             has_tool_use = False
 
             # Ensure the tool-uses always have valid names before sending
             # https://github.com/strands-agents/harness-sdk/issues/1069
-            for item in content:
+            for i, item in enumerate(content):
                 if "toolUse" in item:
                     has_tool_use = True
                     tool_use: ToolUse = item["toolUse"]
@@ -78,7 +86,12 @@ def _normalize_messages(messages: Messages) -> Messages:
                     try:
                         validate_tool_use_name(tool_use)
                     except InvalidToolUseNameException:
-                        tool_use["name"] = "INVALID_TOOL_NAME"
+                        # Copy the item and toolUse to safely mutate
+                        new_item = cast(ContentBlock, dict(item))
+                        new_tool_use = cast(dict[str, Any], dict(tool_use))
+                        new_tool_use["name"] = "INVALID_TOOL_NAME"
+                        new_item["toolUse"] = cast(ToolUse, new_tool_use)
+                        content[i] = new_item
                         replaced_tool_names = True
 
             if has_tool_use:
@@ -92,11 +105,18 @@ def _normalize_messages(messages: Messages) -> Messages:
                 if not removed_blank_message_content_text and before_len != len(content):
                     removed_blank_message_content_text = True
             else:
-                # Replace blank or None 'text' with '[blank text]' for assistant messages
-                for item in content:
+                # Replace blank 'text' with '[blank text]' for assistant messages
+                for i, item in enumerate(content):
                     if "text" in item and (item["text"] is None or not item["text"].strip()):
+                        # Copy the item to safely mutate
+                        new_item = cast(ContentBlock, dict(item))
+                        new_item["text"] = "[blank text]"
+                        content[i] = new_item
                         replaced_blank_message_content_text = True
-                        item["text"] = "[blank text]"
+
+            new_messages.append(new_message)
+        else:
+            new_messages.append(message)  # type: ignore[unreachable]
 
     if removed_blank_message_content_text:
         logger.debug("removed blank message context text")
@@ -105,7 +125,7 @@ def _normalize_messages(messages: Messages) -> Messages:
     if replaced_tool_names:
         logger.debug("replaced invalid tool name")
 
-    return messages
+    return new_messages
 
 
 def remove_blank_messages_content_text(messages: Messages) -> Messages:
--- a/strands-py/src/strands/event_loop/streaming.py
+++ b/strands-py/src/strands/event_loop/streaming.py
@@ -1,12 +1,13 @@
 """Utilities for handling streaming responses from language models."""
 
+import copy
 import json
 import logging
 import threading
 import time
 import warnings
 from collections.abc import AsyncGenerator, AsyncIterable
-from typing import Any, cast
+from typing import Any
 
 from ..models.model import Model
 from ..tools import InvalidToolUseNameException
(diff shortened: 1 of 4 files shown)


arielnabavian (dismissed):

notowen333:
@strandly-the-agent can you go over this one again. A null test logic was flagged? Is that legit? And otherwise?

strands-py/tests/strands/event_loop/test_streaming.py
```diff
@@ -126,6 +126,20 @@ def test_normalize_blank_messages_content_text(messages, exp_result):
...
 
+def test_normalize_messages_does_not_mutate_original():
+    import copy
+
+    original_messages = [
+        {"role": "assistant", "content": [{"text": " \n"}, {"toolUse": {"name": "invalid tool"}}]},
+    ]
+    expected_original = copy.deepcopy(original_messages)
+
+    _ = strands.event_loop.streaming._normalize_messages(original_messages)
```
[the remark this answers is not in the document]
strandly-the-agent: Re-checked this by mutation testing rather than reasoning about it in the abstract — and the conclusion flips.

True that this test alone doesn't prove the returned copies carry the normalization. But that isn't a gap, because the suite already covers it, and this test covers something nothing else does:

| Sabotage applied to `_normalize_messages` | Suite *without* this test |
|---|---|
| `return copy.deepcopy(messages)` — copies made, transformations dropped (the scenario above) | **caught** — 7 pre-existing failures |
| Revert item-level copy-on-write (mutate the original `toolUse`/`text` dicts) | **not caught** — 67 passed |
| Revert list-level copy (`content = message["content"]`) | **not caught** — 67 passed |

The killers for the first row are `test_normalize_blank_messages_content_text` (streaming.py:128, asserts the return value across 6 cases) and `test_stream_messages_normalizes_messages` (streaming.py:1409, asserts what `model.stream` actually receives). Rows 2–3 are the exact bug class this PR exists to fix, and this test is the only thing in 133 tests that catches them. It also fails against `main`'s implementation, so it isn't vacuous.

So `_ = ...` is correct scoping, not an omission — the transformation half is already asserted twice elsewhere. Adding a return-value assertion here would be harmless but redundant. My earlier suggestion was wrong about the risk; no change needed.

<details><summary>Commands</summary>

```
cd strands-py
# baseline
python3 -m pytest tests/strands/event_loop/test_streaming.py -q   # 68 passed
# per mutant, with and without the new test:
python3 -m pytest tests/strands/event_loop/test_streaming.py -q \
  --deselect tests/strands/event_loop/test_streaming.py::test_normalize_messages_does_not_mutate_original
```
</details>

strandly-the-agent:
**Approve, nothing blocking.** The flagged test logic is technically true but not a real gap — and the implementation is correct. One thing is worth a line edit: the PR description names a symptom that can't happen.

**Is the "null test logic" flag legit?** Partly. The test does discard the return value, but the transformation half is already asserted twice elsewhere, and mutation testing shows this test is the *only* guard in 133 tests for the bug class this PR actually fixes. Full table in [that thread](https://github.com/strands-agents/harness-sdk/pull/2326#discussion_r3707113612) — my earlier suggestion there was wrong about the risk, so it's safe to resolve.

**And otherwise?** The copy-on-write is correct. But `event_loop.py:535` already does `messages=copy.deepcopy(agent.messages)` before `_normalize_messages` ever sees the list, so the described symptom ("permanently altered the canonical conversation state before it reached the session manager") was never reachable — I confirmed `agent.messages` survives on `main` too. The real symptom is one layer in: an `InvokeModelStage` middleware holding `ctx.messages` across `next()` saw it mutated out from under it. That reproduces (fails on `main`, passes here). It's internal-only today — all four in-tree consumers register on `InvokeModelStage.Input`, which runs before the mutation — so this is a genuine latent fix, just not the one the body claims.

✅ `1f5f513` · 133 `event_loop` tests pass · mypy (project-pinned 1.20.2) clean on `streaming.py` · 0 return-value divergences vs `main` across 28 hand-built shapes + 20k randomized trials · middleware repro confirmed on both sides.

<details>
<summary>Details — evidence, questions, and non-blocking nits</summary>

### What I verified

| Check | Result |
|---|---|
| Diff scope | exactly 2 files, +42/-8 (`git diff main...pr2326 --stat`) |
| `tests/strands/event_loop/ -q` | 133 passed |
| Behavioral equivalence to `main` | 0 divergences — 28 shapes incl. `reasoningContent`, `toolResult`, `image`/`document`, `text` `None`/`""`/`" \n"`, missing `name`/`toolUseId`, multiple toolUses, extra `metadata`, no-`role`; plus 20k randomized schema-conformant trials |
| `content[:] = [...]` filter lands in the returned value | yes — `content` is the same object as `new_message["content"]` (assigned the line before) |
| `# type: ignore[unreachable]` at `streaming.py:120` | correct and consumed — removing it reproduces `error: Statement is unreachable`; `warn_unused_ignores=true` doesn't flag it |
| Old bug reached `agent.messages`/sessions? | **no** — control test passes on `main` too |
| Old bug observable to middleware? | **yes** — repro below |

The old implementation did mutate its input in 11,922 of 20,000 randomized trials, so the function-level bug was real and frequent — it just couldn't escape the deepcopy.

### Middleware repro

Monkeypatches only `_normalize_messages` (branch impl vs `main`'s), so the split is attributable to this change alone:

```
test_middleware_ctx_messages_not_mutated_by_sdk[main]    FAILED
test_middleware_ctx_messages_not_mutated_by_sdk[pr2326]  PASSED
test_middleware_second_cycle_sees_clean_messages[main]   FAILED
test_middleware_second_cycle_sees_clean_messages[pr2326] PASSED
test_agent_messages_not_mutated[main]                    PASSED   <- control
test_agent_messages_not_mutated[pr2326]                  PASSED

[main] before next(): {'role':'assistant','content':[{'text':''},{'toolUse':{'name':'get weather!',...}}]}
[main]  after next(): {'role':'assistant','content':[{'toolUse':{'name':'INVALID_TOOL_NAME',...}}]}
```

Happy to hand over that test if you want it in-tree — nothing currently guards that seam, and it's the only thing that would have caught the actual regression.

### Questions (none blocking)

- Worth correcting the PR body before merge? It's what ends up in the squash commit, and "corrupted session state" will read as a user-facing bugfix when it isn't one.
- What's the intended contract — "the normalizer doesn't mutate its own input" (achieved) or "nothing downstream can mutate the caller's list" (not achieved: the copies are shallow, so unchanged content items and nested `toolUse.input` dicts are still shared)? I scanned 13 provider configs × 8 corpora and found no in-tree consumer that exploits it, so this may simply be out of scope — but worth naming the intent, maybe as a docstring line.
- Now that `copy.deepcopy` at `event_loop.py:535` *and* this shallow copy both run on every model call, is the doubled copying worth consolidating? Different granularity and purpose, so possibly not — just flagging that both now exist on the hot path.

### Non-blocking nits (2)

- `for i, item in enumerate(content)` at `streaming.py:81` and `:109` — `strands-py/AGENTS.md:286` says never a bare `i`. The index is genuinely used, so it's naming only (`index`).
- The comment at `streaming.py:108` dropped "or None" while the guard at `:110` still checks `item["text"] is None`. Small comment/code drift.

Not filing anything else. I considered the `import copy` inside the test body (`AGENTS.md:272`) — the file already has two uncorrected precedents at `:1554`/`:1574`, so it isn't a new pattern — and a pre-existing in-place `content.pop("reasoningContent", None)` in `sagemaker.py:566`, which needs out-of-contract input to trigger; filing it would be more noise than signal.

</details>

<sub>Automated re-review — worth a human eye before you act on it, especially the middleware finding.</sub>

strands-py/src/strands/event_loop/streaming.py
```diff
@@ -56,29 +56,42 @@ def _normalize_messages(messages: Messages) -> Messages:
...
+    new_messages: Messages = []
+
     for message in messages:
         # only modify assistant messages
         if "role" in message and message["role"] != "assistant":
+            new_messages.append(message)
             continue
         if "content" in message:
-            content = message["content"]
+            # Create shallow copies to prevent mutating the original history
```
Unshure: Instead of shallow copies throughout, I would prefer doing one deep copy at the top, which should simplify the code in this pr.

Unshure: Done — replaced the shallow copies with a single `copy.deepcopy(messages)` at the top of `_normalize_messages`. The function body now matches the pre-fix logic, just operating on the copy.

Unshure:
Took over this pr as it was stale for a while

arielnabavian (approved):

PR #2327, opened 2026-05-26
feat(bidi/openai): support image input via BidiImageInputEvent

cagataycali:
## Summary

Adds image-input support to `BidiOpenAIRealtimeModel`, bringing it to parity with `BidiGeminiLiveModel` which already handles `BidiImageInputEvent`.

Previously, sending a `BidiImageInputEvent` to the OpenAI realtime model raised `ValueError: content not supported`. OpenAI's Realtime API on `gpt-realtime` models *does* accept image inputs as `input_image` content blocks on `conversation.item.create` - this PR wires that up.

## Motivation

We're building voice agents where the user says *"look at me"* / *"what's on my whiteboard?"* and we want to inject a captured camera frame straight into the realtime model's multimodal context, then let it reply in audio.

Gemini Live works out of the box. For OpenAI we had to monkeypatch `BidiOpenAIRealtimeModel.send()` from outside the SDK - this PR upstreams the fix.

## Changes

**`src/strands/experimental/bidi/models/openai_realtime.py`**
- Import `BidiImageInputEvent`.
- Dispatch it in `send()` to a new `_send_image_content()` method.
- `_send_image_content()` encodes the image as a `data:` URL (`data:<mime>;base64,<b64>`) and emits a single `conversation.item.create` event with `role=user` and an `input_image` content block.
- Does **not** auto-trigger `response.create` — caller decides when to commit (mirrors the audio/buffered semantics rather than text/auto-reply, which matches typical streaming-vision use cases where you send N frames then a follow-up text/audio prompt).

**`tests/strands/experimental/bidi/models/test_openai_realtime.py`**
- The existing case that asserted `BidiImageInputEvent` raises `ValueError` is replaced with one that verifies the new wire format and that no implicit `response.create` is sent.

## Wire format

```json
{
  "type": "conversation.item.create",
  "item": {
    "type": "message",
    "role": "user",
    "content": [
      {
        "type": "input_image",
        "image_url": "data:image/jpeg;base64,<...>"
      }
    ]
  }
}
```

This matches OpenAI's documented format for image inputs on `gpt-realtime`.

## Tests

```
hatch fmt --formatter
hatch fmt --linter           # All checks passed!
hatch test tests/strands/experimental/bidi/models/test_openai_realtime.py
# 27 passed in 5.21s
```

The broader `tests/strands/experimental/bidi/` suite passes except for unrelated pre-existing failures from optional deps (`aws_sdk_bedrock_runtime` for Nova Sonic, `pyaudio` for audio I/O) — none touched by this change.

## Tenets alignment

- **Composability**: matches Gemini Live's behavior, same event type, same shape.
- **The obvious path is the happy path**: `agent.send(BidiImageInputEvent(...))` Just Works on OpenAI now.
- **Embrace common standards**: uses OpenAI's own documented `input_image` block + `data:` URL.

## Checklist

- [x] Code follows existing patterns (mirrors `_send_audio_content` / `_send_text_content`).
- [x] Formatter (`hatch fmt --formatter`) clean.
- [x] Linter + mypy (`hatch fmt --linter`) clean.
- [x] Unit tests updated and passing.
- [x] No public API changes other than removing the previous `ValueError`.

### DIFF
--- a/strands-py/src/strands/experimental/bidi/models/openai_realtime.py
+++ b/strands-py/src/strands/experimental/bidi/models/openai_realtime.py
@@ -24,6 +24,7 @@ from ..types.events import (
     BidiAudioInputEvent,
     BidiAudioStreamEvent,
     BidiConnectionStartEvent,
+    BidiImageInputEvent,
     BidiInputEvent,
     BidiInterruptionEvent,
     BidiOutputEvent,
@@ -712,7 +713,7 @@ class BidiOpenAIRealtimeModel(BidiModel):
             content: Typed event (BidiTextInputEvent, BidiAudioInputEvent, BidiImageInputEvent, or ToolResultEvent).
 
         Raises:
-            ValueError: If content type not supported (e.g., image content).
+            ValueError: If content type not supported.
         """
         if not self._connection_id:
             raise RuntimeError("model not started | call start before sending")
@@ -722,6 +723,8 @@ class BidiOpenAIRealtimeModel(BidiModel):
             await self._send_text_content(content.text)
         elif isinstance(content, BidiAudioInputEvent):
             await self._send_audio_content(content)
+        elif isinstance(content, BidiImageInputEvent):
+            await self._send_image_content(content)
         elif isinstance(content, ToolResultEvent):
             tool_result = content.get("tool_result")
             if tool_result:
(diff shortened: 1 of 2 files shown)


mkmeral (dismissed):

yonib05:
@mkmeral why didn't we merge? 
@cagataycali mind rebasing so we can merge.

yonib05 (approved):

PR #2338, opened 2026-05-27
feat(bedrock): add request_metadata for per-request cost attribution

psantus:
## Description

Add `request_metadata` config option to `BedrockModel` that maps to the `requestMetadata` field in Converse/ConverseStream API requests.

This enables per-request finops by attaching key-value tags (e.g. team, environment, feature) that appear in model invocation logs.

Default behavior is unchanged — no metadata is sent unless explicitly configured.

## Usage

```python
model = BedrockModel(
    model_id="anthropic.claude-sonnet-4-20250514",
    request_metadata={"team": "orchestrator", "environment": "prod"}
)

# Or update per-conversation:
model.update_config(request_metadata={"team": "billing", "session_id": "abc123"})
```

## Reference

https://docs.aws.amazon.com/bedrock/latest/userguide/cost-mgmt-request-metadata.html

## What was tested

- `test_format_request_with_request_metadata` — metadata included when configured
- `test_format_request_without_request_metadata` — backward compat (no key when unset)
- Full bedrock test suite (168 tests) passes with no regressions

### DIFF
--- a/strands-py/src/strands/models/bedrock.py
+++ b/strands-py/src/strands/models/bedrock.py
@@ -119,7 +119,8 @@ class BedrockModel(Model):
             temperature: Controls randomness in generation (higher = more random)
             top_p: Controls diversity via nucleus sampling (alternative to temperature)
             request_metadata: Key-value metadata to attach to the request for cost tracking and attribution.
-                Appears in model invocation logs for per-request finops. Max 16 entries, keys/values max 256 chars.
+                Appears in model invocation logs for per-request finops. Max 16 entries; keys/values max 256 chars
+                and limited to letters, digits, spaces, and ":_@$#=/+,-." (e.g. "team (eng)" or "cost%" are rejected by Bedrock).
                 See https://docs.aws.amazon.com/bedrock/latest/userguide/cost-mgmt-request-metadata.html
             use_native_token_count: Whether to use the native Bedrock CountTokens API.
                 When True, count_tokens() calls the Bedrock API for accurate counts.
--- a/strands-py/src/strands/models/bedrock.py
+++ b/strands-py/src/strands/models/bedrock.py
@@ -118,6 +118,9 @@ class BedrockModel(Model):
                 See https://docs.aws.amazon.com/bedrock/latest/userguide/structured-output.html
             temperature: Controls randomness in generation (higher = more random)
             top_p: Controls diversity via nucleus sampling (alternative to temperature)
+            request_metadata: Key-value metadata to attach to the request for cost tracking and attribution.
+                Appears in model invocation logs for per-request finops. Max 16 entries, keys/values max 256 chars.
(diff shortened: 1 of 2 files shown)


kimnamu (commented):
Thanks for this, Paul — per-request cost attribution is a genuinely useful addition for anyone running Bedrock at scale. I'm not a maintainer, just another Bedrock user who reviewed this carefully and ran the branch locally.

A few things I verified:

- **The field is correct.** `requestMetadata` is a real member of both `Converse` and `ConverseStream` in the boto3 service model, with exactly the constraints you documented (`min 1` / `max 16` entries, keys/values `max 256`). ✅
- **Pass-through (no client-side validation) is the right call** — it's consistent with how every other config field here is handled (`guardrail_id`, `max_tokens`, etc. all defer validation to Bedrock). I confirmed `_format_request` cleanly emits `requestMetadata` when set and omits the key entirely when unset, so backward compat holds. ✅
- **One doc gap** (inline suggestion below): the AWS character-set constraint isn't mentioned, and it bites on very ordinary tag names.

Non-blocking, and I checked this is actually correct: the `if self.config.get("request_metadata")` guard drops an explicitly-empty `request_metadata={}` instead of sending it. That's the right behavior — Bedrock's `requestMetadata` map has a `min 1` constraint, so an empty map would be rejected anyway. No change needed; just noting I verified the edge case.

Clean, well-scoped, and the tests exercise the real behavior — I'd love to see it land.

_(This review was done with the help of an AI agent (Claude Code); I reviewed and verified every claim and ran the branch locally before posting.)_

strands-py/src/strands/models/bedrock.py
```diff
@@ -118,6 +118,9 @@ class BedrockConfig(BaseModelConfig, total=False):
                 See https://docs.aws.amazon.com/bedrock/latest/userguide/structured-output.html
             temperature: Controls randomness in generation (higher = more random)
             top_p: Controls diversity via nucleus sampling (alternative to temperature)
+            request_metadata: Key-value metadata to attach to the request for cost tracking and attribution.
+                Appears in model invocation logs for per-request finops. Max 16 entries, keys/values max 256 chars.
```
kimnamu: The reference link is great. One small gap: the docstring gives the length limits (16 entries / 256 chars) but not the **character set** Bedrock enforces on keys and values (`[a-zA-Z0-9\s:_@$#=/+,-.]`). Common finops tags like `team (eng)` or `cost%` pass through this code fine but get rejected at request time with a `ValidationException`. Spelling it out saves users a confusing runtime error:

```suggestion
                Appears in model invocation logs for per-request finops. Max 16 entries; keys/values max 256 chars
                and limited to letters, digits, spaces, and ":_@$#=/+,-." (e.g. "team (eng)" or "cost%" are rejected by Bedrock).
```

psantus: @kimnamu : thanks for your feedback and contribution!

PR #2354, opened 2026-05-28
fix(openai): read vllm reasoning deltas

### LINKED ISSUE #2182: [BUG] OpenAIModel ignores vLLM reasoning field (delta.reasoning / message.reasoning)

he-yufeng:
﻿## Summary

Fixes #2182.

OpenAI-compatible vLLM streams can emit reasoning text on `delta.reasoning` instead of the older `delta.reasoning_content` field. The OpenAI adapter now accepts both fields and still prefers the existing `reasoning_content` path when present.

## Changes

- Read non-empty string reasoning from `delta.reasoning_content` or `delta.reasoning`.
- Keep existing text/tool streaming behavior unchanged.
- Add regression coverage for a vLLM-style `delta.reasoning` stream.

## To verify

Run from `strands-py`:

```bash
uv run pytest tests/strands/models/test_openai.py -q --basetemp .tmp/pytest-2182 -p no:cacheprovider
uv run ruff check src/strands/models/openai.py tests/strands/models/test_openai.py
python -m py_compile src/strands/models/openai.py tests/strands/models/test_openai.py
git diff --check
```

Local result on Windows: 96 passed.

### DIFF
--- a/strands-py/src/strands/models/openai.py
+++ b/strands-py/src/strands/models/openai.py
@@ -700,7 +700,11 @@ class OpenAIModel(Model):
                     continue
                 choice = event.choices[0]
 
-                if hasattr(choice.delta, "reasoning_content") and choice.delta.reasoning_content:
+                reasoning_content = getattr(choice.delta, "reasoning_content", None)
+                if not isinstance(reasoning_content, str) or not reasoning_content:
+                    reasoning_content = getattr(choice.delta, "reasoning", None)
+
+                if isinstance(reasoning_content, str) and reasoning_content:
                     chunks, data_type = self._stream_switch_content("reasoning_content", data_type)
                     for chunk in chunks:
                         yield chunk
@@ -708,7 +712,7 @@ class OpenAIModel(Model):
                         {
                             "chunk_type": "content_delta",
                             "data_type": data_type,
-                            "data": choice.delta.reasoning_content,
+                            "data": reasoning_content,
                         }
                     )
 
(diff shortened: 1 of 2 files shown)


lizradway (approved):

pgrayy (approved):

strands-py/src/strands/models/openai.py
```diff
@@ -700,15 +700,19 @@ async def stream(
                     continue
                 choice = event.choices[0]
 
-                if hasattr(choice.delta, "reasoning_content") and choice.delta.reasoning_content:
+                reasoning_content = getattr(choice.delta, "reasoning_content", None)
+                if not isinstance(reasoning_content, str) or not reasoning_content:
+                    reasoning_content = getattr(choice.delta, "reasoning", None)
```
pgrayy, line 705: It seems technically speaking this is a vllm specific field and not openai compatible. I still think we can add it for convenience.

PR #2361, opened 2026-05-28
feat(strands-py-wasm): use call_async for true async streaming

pgrayy:
## Description

Switch the strands-py-wasm runtime from thread-pool-based async (`run_in_executor`) to true cooperative async using wasmtime-py's `call_async` API. This enables the Python asyncio event loop to remain responsive while the WASM guest awaits network I/O (model provider HTTP calls).

## How async streaming works

The WASM guest (compiled from the TS SDK) exposes a pull-based `event-stream` resource. Each call to `event-stream.read` advances the internal async generator by one event. When the guest needs to wait on I/O (e.g. an HTTP response from Bedrock), the call suspends.

Previously, `_runtime.py` ran these calls on a thread pool via `run_in_executor`, blocking a thread during I/O. Now it uses `func.call_async(store, ...)` which leverages a socket-pair notification bridge:

1. `wasmtime_component_func_call_async()` starts the call and returns a future
2. `wasmtime_call_future_poll_with_notify(future, write_sock)` polls it with a waker that writes to a socket when ready
3. If Pending, Python awaits `loop.sock_recv(read_sock)`, yielding the event loop
4. When wasmtime's Tokio runtime completes the HTTP call, the waker fires, Python wakes, re-polls, and the event is ready

This means `stream_async` yields events incrementally as tokens arrive, and other asyncio tasks (heartbeats, parallel agents) run concurrently without thread contention.

## Changes

- Engine config: enable `wasm_component_model_async` and `async_allow_sync`
- Linker: `add_wasip2()` -> `add_wasip2_async()`
- All WASM function calls use `func.call_async(store, ...)` instead of sync calls wrapped in `run_in_executor`
- Removed the threading lock (no longer needed without thread pool dispatch)
- Merged `init()` into `async_init()` (constructor now uses `call_async` too)
- Added `botocore` as a base dependency (required for credential resolution since the WASM bundle uses the browser AWS SDK build which has no credential chain)
- Added initial integration test for `stream_async`

## Type of Change

New feature

## Testing

- Added `strands-py-wasm/tests/test_agent.py` with an integration test verifying `stream_async` produces well-typed `StreamEvent` instances and terminates with `StreamEvent_Stop`
- Manually verified with a concurrent streaming test: `stream_async` yields `ModelUpdate` events interleaved with a 10ms heartbeat task on the same event loop, proving true cooperative concurrency

## Checklist

----

### DIFF
--- a/strands-py-wasm/pyproject.toml
+++ b/strands-py-wasm/pyproject.toml
@@ -30,6 +30,7 @@ classifiers = [
 dependencies = [
     # Imports as `wasmtime`. Pinned to pgrayy fork until upstream PRs land.
     "pgrayy-wasmtime>=46.0.6,<47.0.0",
+    "botocore>=1.34.0",
 ]
 
 
--- a/strands-py-wasm/pyproject.toml
+++ b/strands-py-wasm/pyproject.toml
@@ -30,7 +30,7 @@ classifiers = [
 dependencies = [
     # Imports as `wasmtime`. Pinned to pgrayy fork until upstream PRs land.
     "pgrayy-wasmtime>=46.0.6,<47.0.0",
-    "botocore>=1.34.0",
+    "botocore>=1.29.0,<2.0.0",
 ]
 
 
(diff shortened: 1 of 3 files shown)


strands-py-wasm/pyproject.toml
```diff
@@ -30,6 +30,7 @@ classifiers = [
 dependencies = [
     # Imports as `wasmtime`. Pinned to pgrayy fork until upstream PRs land.
     "pgrayy-wasmtime>=46.0.6,<47.0.0",
+    "botocore>=1.29.0,<2.0.0",
```
pgrayy, line 33: For [BedrockModel](https://github.com/strands-agents/sdk-python/blob/main/strands-py-wasm/src/strands/__init__.py#L226).

chaynabors (approved):

Issue #124, opened 2025-05-27
[BUG] Tool spec normalization


awsarron:
### Checks

- [x] I have updated to the lastest minor and patch version of Strands
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched [./issues](./issues?q=) and there are no duplicates of my issue

### Strands Version

0.1.5

### Python Version

3.13.3

### Operating System

macOS

### Installation Method

pip

### Steps to Reproduce

1. Create a module tool with `TOOL_SPEC` defined. In the tool spec, include deeply nested objects.
2. Create an agent with the tool and check its tool configuration after tool loading:
```python
from strands import Agent
from my_tools import test_nested_tool

agent = Agent(tools=[test_nested_tool])
print(agent.tool_config)
```

### Expected Behavior

Expected loaded tools with deeply nested objects in the tool spec to preserve the nesting.

### Actual Behavior

Deeply nested object structures are not preserved, leading to context loss in the specification provided to the model during inference.

### Additional Context

Example provided tool spec:
```python
{
    "name": "generate_income_assessment",
    "description": "Generate an income assessment report",
    "inputSchema": {
        "json": {
            "type": "object",
            "required": [
                "application_id",
                "assessment_type",
                "assessment_criteria"
            ],
            "properties": {
                "application_id": {
                    "type": "string",
                    "description": "The ID of the application being assessed"
                },
                "assessment_type": {
                    "type": "string",
                    "description": "Type of assessment being performed",
                    "const": "INCOME"
                },
                "assessment_criteria": {
                    "type": "object",
                    "required": [
                        "verify_application_document",
                        "verify_payslip_document", 
                        "verify_bank_statements",
                        "verify_employer_matches",
                        "verify_salary_matches"
                    ],
                    "properties": {
                        "verify_application_document": {
                            "type": "object",
                            "properties": {
                                "question": { 
                                    "const": "Verify application document is present",
                                    "description": "A specific evaluation criteria that must be assessed as part of the income verification process"
                                },
                                "result": { 
                                    "type": "string",
                                    "enum": ["PASS", "FAIL", "UNKNOWN"],
                                    "description": "The outcome of the assessment check: PASS indicates criteria are met, FAIL indicates criteria are not met, or UNKNOWN if insufficient information is available"
                                },
                                "explanation": { 
                                    "type": "string",
                                    "description": "A concise justification supporting the result, including evidence and reasoning"
                                }
                            },
[...]

PR #2365, opened 2026-05-28
feat(interventions): add cedar vended intervention handler

lizradway:
## Description

Adds a `CedarAuthorization` intervention handler that evaluates [Cedar](https://cedarpolicy.com) authorization policies at the tool-call boundary, providing identity-aware access control for agent tool invocations.

The handler integrates with [`@cedar-policy/cedar-wasm`](https://www.npmjs.com/package/@cedar-policy/cedar-wasm) and [`@cedar-policy/mcp-schema-generator-wasm`](https://www.npmjs.com/package/@cedar-policy/mcp-schema-generator-wasm) to evaluate Cedar policies before each tool call, returning `deny()` or `proceed()` based on the policy decision.

Follows the [cedar-for-agents](https://github.com/cedar-policy/cedar-for-agents) schema generator conventions:
- One Cedar action per tool (e.g. `Action::"search"`)
- Resource is unconstrained by default
- Context is nested: `{ input: <tool args>, session: { hour_utc, call_count } }`

### Cedar-to-tool mapping

| Cedar concept | Maps to | Example |
|---|---|---|
| **Principal** | User identity (static or from `invocationState`) | `User::"alice@acme.com"` |
| **Action** | Tool name | `Action::"search"` |
| **Resource** | Unconstrained default (check tool args via `context.input` in policies) | `Resource::"agent"` |
| **Context.input** | Tool's input arguments | `{ query: "report" }` |
| **Context.session** | Automatic invocation metadata + custom fields via `contextEnricher` | `{ hour_utc: 14, call_count: 3 }` |

### Usage examples

**Simple — permit specific tools, deny everything else:**
```typescript
import { CedarAuthorization } from '@strands-agents/sdk/vended-interventions/cedar'

const cedar = new CedarAuthorization({
  policies: `
    permit(principal, action == Action::"search", resource);
    permit(principal, action == Action::"read_file", resource);
  `,
})

const agent = new Agent({
  tools: [searchTool, readFileTool, deleteTool],
  interventions: [cedar],
})

// search and read_file work, delete is denied (no permit matches)
await agent.invoke('Search for reports then delete them')
```

**Role-based access (multi-tenant API, via contextEnricher):**
```typescript
const cedar = new CedarAuthorization({
  policies: `
    permit(principal, action, resource) when { context.session.role == "admin" };
    permit(principal, action == Action::"search", resource) when { context.session.role == "analyst" };
  `,
  principalResolver: (state) => {
    if (!state.user_id) return undefined
    return { type: 'User', id: String(state.user_id) }
  },
  contextEnricher: ({ invocationState }) => ({
    role: String(invocationState.role ?? 'none'),
  }),
})

// alice (admin) can do anything, bob (analyst) can only search
await agent.invoke('Delete record 42', { invocationState: { user_id: 'bob', role: 'analyst' } })
// → denied
```

**Rate limiting:**
```typescript
const cedar = new CedarAuthorization({
  policies: `
    permit(principal, action == Action::"send_email", resource)
    when { context.session.call_count < 5 };

    permit(principal, action == Action::"search", resource);
  `,
})

// send_email works 4 times, then gets denied. search is unlimited.
```

**With schema validation from tool definitions:**
```typescript
import { tool } from '@strands-agents/sdk'
import { CedarAuthorization } from '@strands-agents/sdk/vended-interventions/cedar'
import { z } from 'zod'

const searchTool = tool({
  name: 'search',
  description: 'Search for information',
  inputSchema: z.object({ query: z.string() }),
  callback: (input) => `Results for: ${input.query}`,
})

const deleteTool = tool({
  name: 'delete_record',
  description: 'Delete a record',
  inputSchema: z.object({ record_id: z.string() }),
  callback: (input) => `Deleted ${input.record_id}`,
})

// Pass tools — auto-generates schema and validates policies at construction
const cedar = new CedarAuthorization({
  policies: `
    permit(principal, action == Action::"search", resource);
    permit(principal, action == Action::"delete_record", resource)
    when { context.session.role == "admin" };
  `,
  tools: [searchTool, deleteTool],
  contextEnricher: ({ invocationState }) => ({
    role: String(invocationState.role ?? 'none'),
  }),
})

const agent = new Agent({
  tools: [searchTool, deleteTool],
  interventions: [cedar],
})
```

**Schema validation catching typos at startup:**
```typescript
// Typo in action name — caught before any tool call happens
const cedar = new CedarAuthorization({
  policies: `
    permit(principal, action == Action::"search", resource);
    permit(principal, action == Action::"deleet_record", resource);
  `,
  tools: [searchTool, deleteTool],
})
// ↑ Throws: "Cedar policy validation failed: unknown action 'deleet_record'"
```

**Environment gating via contextEnricher:**
```typescript
const cedar = new CedarAuthorization({
  policies: `
    permit(principal, action == Action::"deploy", resource)
    when { context.session has environment && context.session.environment != "production" };
  `,
  contextEnricher: ({ invocationState }) => ({
    environment: String(invocationState.environment ?? 'unknown'),
  }),
})

// deploy works in staging, denied in production
await agent.invoke('Deploy', { invocationState: { environment: 'production' } })
// → denied
```

### Key features

- **Fail-closed design** — missing principal identity denies all tool calls
- **Policy validation at construction** — invalid Cedar syntax throws immediately; schema validation catches unknown action names when `tools` is provided
- **File-based or inline policies** — pass Cedar policy text directly or a path to a `.cedar` file
- **Static or dynamic principal** — `principal` for CLI/single-user, `principalResolver` for multi-tenant (defaults to `User::"anonymous"`)
- **MCP schema generator integration** — pass `tools` to auto-generate Cedar schema and validate policies against tool definitions at construction
- **Context enrichment** — automatic `hour_utc`, `call_count` in session, plus custom fields via `contextEnricher` (with access to `invocationState`)
- **Durable rate limiting** — call counts stored on `agent.appState`, persists with session managers, counts only successful calls
- **Hot reload** — `reload()` method re-reads policy/entity/schema files atomically
- **Configurable error handling** — `onError: 'throw' | 'deny' | 'proceed'`

### Integration with cedar-for-agents

This handler integrates with the [`cedar-for-agents`](https://github.com/cedar-policy/cedar-for-agents) ecosystem ([#124](https://github.com/cedar-policy/cedar-for-agents/issues/124)):

- **Schema generator** (`@cedar-policy/mcp-schema-generator-wasm`) — pass `tools` to auto-generate Cedar schemas from your tool definitions. Validates policies against tool action names at construction time, catching typos before deployment.

### Future

- **Policy builder** ([`cedar-agent-policy-builder`](https://github.com/cedar-policy/cedar-for-agents/pull/137)) — declarative builder API (`.role('admin', ['*'])`) that generates policies, entities, and schemas without writing Cedar syntax. Outputs feed directly into `CedarAuthorization` config.

## Related Issues

- Design doc: https://github.com/strands-agents/docs/blob/main/designs/0006-cedar-authorization.md
- Intervention primitive: https://github.com/strands-agents/docs/blob/main/designs/0007-intervention-primitive.md
- Schema generator integration: https://github.com/cedar-policy/cedar-for-agents/issues/124
- Policy builder: https://github.com/cedar-policy/cedar-for-agents/pull/137

## Documentation PR

N/A

## Type of Change

New feature

## Testing

- [x] I ran `npm run check` for TypeScript (type-check, tests, lint)
- 36 unit tests using **real Cedar policy evaluation** (no mocks) covering: permit/deny, default-deny, role-based access (admin/analyst/viewer), rate limiting, fail-closed on missing principal, malformed policy rejection at construction, context.input argument checks, context enrichment (match/mismatch), onError modes (throw/deny/proceed), file loading (.cedar/.json/missing file), call count reset, atomic reload, schema validation (valid/unknown action/unknown attribute/inline/skipped), static principal, anonymous default, schema generator integration (auto-schema, unknown action detection, nested input schemas, naming collision error handling)
- 3 integration tests with real Bedrock model + real cedar-wasm: policy permit, policy forbid, fail-closed
- Adversarial testing: policy injection via tool names (safe), context injection via tool input (safe), entity type mismatches (fail-closed), concurrent call counts (safe)

## Checklist

----


strands-ts/src/vended-interventions/cedar/cedar.ts
```diff
@@ -0,0 +1,282 @@
...
+
+}
+
+function loadPolicies(policies: string): string {
+  if (policies.endsWith('.cedar')) {
+    if (!existsSync(policies)) {
+      throw new Error(`Cedar policy file not found: ${policies}`)
+    }
+    return readFileSync(policies, 'utf-8')
+  }
```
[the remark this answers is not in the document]
lizradway: nah, too much for starters

strands-ts/src/vended-interventions/cedar/cedar.ts
```diff
@@ -0,0 +1,282 @@
+import { InterventionHandler } from '../../interventions/handler.js'
+import { proceed, deny } from '../../interventions/actions.js'
+import type { InterventionAction } from '../../interventions/actions.js'
+import type { BeforeToolCallEvent } from '../../hooks/events.js'
+import type { OnError } from '../../interventions/handler.js'
+import { isAuthorized, type Entities, type CedarValueJson } from '@cedar-policy/cedar-wasm/nodejs'
+import { readFileSync, existsSync } from 'node:fs'
```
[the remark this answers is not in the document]
lizradway: this is expected to be node only and is documented

strands-ts/src/vended-interventions/cedar/cedar.ts
```diff
@@ -0,0 +1,282 @@
...
+  private readonly _policies: string
+  private readonly _entities: CedarEntity[]
+  private readonly _principalResolver: (invocationState: Record<string, CedarValueJson>) => CedarEntityUid | undefined
+  private readonly _resourceResolver: ResourceResolver | undefined
+  private readonly _contextEnricher: CedarAuthorizationConfig['contextEnricher']
+  private readonly _callCounts = new Map<string, Map<string, number>>()
+  private readonly _maxSessions = 1000
+
+  constructor(config: CedarAuthorizationConfig) {
+    super()
```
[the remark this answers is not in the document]
lizradway: overly defensive. user choses what goes in invocationState

strands-ts/src/vended-interventions/cedar/cedar.ts
```diff
@@ -0,0 +1,282 @@
...
+  readonly name = 'cedar-authorization'
+  override readonly onError: OnError
+
+  private readonly _policies: string
+  private readonly _entities: CedarEntity[]
+  private readonly _principalResolver: (invocationState: Record<string, CedarValueJson>) => CedarEntityUid | undefined
+  private readonly _resourceResolver: ResourceResolver | undefined
+  private readonly _contextEnricher: CedarAuthorizationConfig['contextEnricher']
+  private readonly _callCounts = new Map<string, Map<string, number>>()
+  private readonly _maxSessions = 1000
```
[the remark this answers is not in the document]
lizradway: considering as follow up

strands-ts/src/vended-interventions/cedar/cedar.ts
```diff
@@ -0,0 +1,282 @@
...
+    }
+    const next = (session.get(toolName) ?? 0) + 1
+    session.set(toolName, next)
+    return next
+  }
+
+}
+
+function loadPolicies(policies: string): string {
+  if (policies.endsWith('.cedar')) {
```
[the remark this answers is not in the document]
lizradway: YAGNI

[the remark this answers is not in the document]
lizradway: can update when we add builder if we want

strands-ts/src/vended-interventions/cedar/cedar.ts
```diff
@@ -0,0 +1,369 @@
...
+    this.onError = config.onError ?? 'throw'
+
+    validatePolicies(this._policies, this._schema)
+  }
+
+  override beforeToolCall(event: BeforeToolCallEvent): InterventionAction {
+    const invocationState = event.invocationState as Record<string, CedarValueJson>
+    const principal = this._principalResolver(invocationState)
+    if (!principal) {
+      return deny('No principal identity found in invocation state')
```
mkmeral, line 150: do I have to pass it in invocation state, can't i pass it in original config? Is this something we expect to change with each invocation?

mkmeral: similarly should the principal resolver be required?

strands-ts/src/vended-interventions/cedar/cedar.ts
```diff
@@ -0,0 +1,369 @@
...
+   * Maps tool calls to Cedar resources. When omitted, the resource is
+   * unconstrained (`Resource::"agent"`). Use this to map tools to
+   * domain-specific entities (e.g. `Record::"42"`).
+   */
+  resourceResolver?: ResourceResolver | undefined
+
+  /**
+   * Adds extra fields to the `context.session` object passed to Cedar.
+   * Called on every tool invocation. Cannot overwrite built-in fields
+   * (`hour_utc`, `call_count`, `environment`).
```
mkmeral: do we keep track of call_count somewhere? do we persist in agent state? how would it work with session managers?

agent-of-mkmeral:
## 🔴 Two security findings in the rate-limit / environment handling

Great PR overall — the core Cedar evaluation is solid (default-deny, forbid-wins, fail-closed on missing principal, atomic `reload()`, schema validation all hold under adversarial testing). But two of the handler's **own helper paths** (not the Cedar engine) let a caller defeat policies that look airtight. Both are reproducible against real `@cedar-policy/cedar-wasm` evaluation, and both are fixable in `cedar.ts` with no engine changes.

The root cause of #1 is also a correctness/accuracy problem with `call_count` worth calling out on its own (bottom section).

<details>
<summary>🔴 <b>#1 — Rate-limit bypass via session-map eviction (HIGH)</b></summary>

**`_incrementCallCount` evicts the oldest session when the map hits `_maxSessions = 1000`:**

```typescript
private _incrementCallCount(sessionId: string, toolName: string): number {
  let session = this._callCounts.get(sessionId)
  if (!session) {
    if (this._callCounts.size >= this._maxSessions) {
      const oldest = this._callCounts.keys().next().value!   // evict oldest-inserted
      this._callCounts.delete(oldest)
    }
    session = new Map()
    this._callCounts.set(sessionId, session)
  }
  const next = (session.get(toolName) ?? 0) + 1
  session.set(toolName, next)
  return next
}
```

The `sessionId` comes straight from caller-influenced `invocationState`:
```typescript
const sessionId = (invocationState.session_id as string | undefined) ?? '_default'
```

**Attack:** against a policy like `permit(... action == Action::"send_email") when { context.session.call_count < 3 }`, a caller who can vary `session_id` hits the rate limit, then spawns 1000 throwaway sessions to evict **their own** counter from the map — resetting `call_count` to 1.

**Observed (real cedar-wasm eval):**
```
attacker call 1: proceed
attacker call 2: proceed
attacker call 3: deny   (rate limit hit)
... create 1000 distinct throwaway sessions (evicts attacker's entry) ...
attacker call AFTER flood: proceed   <-- counter reset, limit bypassed
```
Control (no flood) stays denied on the 4th call, confirming eviction is the cause.

**Fix options:** evict on a key that isn't attacker-controllable, or use a real LRU keyed on identity, or a persistent/shared store — or, if `call_count` is only meant as a soft guardrail, **document it as best-effort and not a security boundary** (see the `call_count` section below). Don't silently drop a counter that a `forbid`/`permit` condition depends on.
</details>

<details>
<summary>🔴 <b>#7 — <code>forbid … when environment == "production"</code> silently fails open when env is omitted (MED-HIGH)</b></summary>

```typescript
environment: (invocationState.environment as string | undefined) ?? 'unknown',
```

A natural, idiomatic deny-in-prod policy:
```cedar
permit(principal, action == Action::"deploy", resource);
forbid(principal, action == Action::"deploy", resource)
  when { context.session.environment == "production" };
```
never fires when the caller simply **omits** `environment` — it defaults to `"unknown"`, not `"production"`, so the `forbid` doesn't match and the action is allowed.

**Observed:**
```
environment = "production"  -> deny     (correct)
environment = <omitted>     -> proceed  (production guard bypassed)
```

The allow-list form **does** fail closed on omission (verified), which is the safer pattern:
```cedar
permit(principal, action == Action::"deploy", resource)
  when { context.session.environment == "production" };   // omitted env -> no permit -> deny
```

**Fix options:** drop the `?? 'unknown'` default and omit the field instead (so `has` checks / schema force explicitness), and/or document loudly that prod-gating must be written as an allow-list (`permit … when env == "production"`), never a deny-list (`forbid … when env == "production"`). As written, the deny-list form is a deny→allow footgun triggered by *not setting a field*.
</details>

<details>
<summary>⚠️ <b><code>call_count</code> accuracy — it's best-effort, not a reliable counter</b></summary>

Independent of the bypass above, `call_count` is inaccurate in several ways that are worth documenting (or fixing) so policy authors don't over-trust it:

```typescript
private readonly _callCounts = new Map<string, Map<string, number>>()  // in-memory only
```

1. **Not persisted** — in-memory on the handler instance. A process restart or a new `CedarAuthorization` instance resets all counts to 0.
2. **Per-instance, not global** — behind N load-balanced workers each replica has its own map, so the effective limit is up to **N×** the policy value.
3. **Resettable via eviction** — the FIFO drop above (finding #1).
4. **Counts denied calls too** — `_incrementCallCount` runs *before* the authorization decision, so blocked attempts still consume budget. Defensible as anti-probing, but it means the number is “attempts,” not “successful calls.”
5. **Keyed on `session_id`, not identity** — a fresh `session_id` is a fresh counter, so it doesn't bound a *user* across sessions.

Also: the PR description refers to the eviction as **“LRU”**, but the implementation is **FIFO** (`_callCounts.keys().next().value` drops the first-*inserted* key, not the least-recently-*used*). Worth reconciling the wording either way.

**Suggestion:** if `call_count` is intended as a hard limit, make it durable + identity-keyed (e.g. a shared store) and don't evict security-relevant counters. If it's a soft guardrail, a one-line doc note on `context.session.call_count` (“best-effort, single-process, not a security boundary; enforce hard rate limits upstream”) would set the right expectation.
</details>

---

**None of these touch the Cedar engine** — they're all in the handler's helper logic around it (`_incrementCallCount`, the `environment` default). Happy to open a PR with regression tests (a session-eviction test and an omitted-`environment` test) if that's useful. Nice work on the rest — the abstraction fit (3rd `InterventionHandler` impl) and the real-evaluation test suite are great. 👍

strands-ts/src/vended-interventions/cedar/cedar.ts
lizradway: >do I have to pass it in invocation state, can't i pass it in original config? Is this something we expect to change with each invocation?

this is expected to change for any multi-tenant deployed solution. but i think it makes sense to be able to pass this into the config for the cli case, can add a fall through to config if user is not in invocation state.

> similarly should the principal resolver be required?

added static principal identity override option to config, so the resolver is no longer required.

strands-ts/src/vended-interventions/cedar/cedar.ts
lizradway: can update to be persisted in agent state.

lizradway:
>  Rate-limit bypass via session eviction

Storing rates in agent state resolves this

> forbid … when environment == "production"

removed environment from build in session context, but cedar skips policies that reference missing fields rather than denying (this essentially denies for allow-list policies, but skips over forbid policies). thus, the safe pattern endorsed by cedar (https://docs.cedarpolicy.com/policies/syntax-operators.html) is the allow list pattern or a blanket forbid. updated documentation to note this.

lizradway:
> call_count accuracy — it's best-effort, not a reliable counter

added decrementCallCount to roll back counts for denied tool calls

agent-of-mkmeral:
## Review — 2 reproduced bugs + CI root cause

**TL;DR:**
- 🔴 **CI Gate failure root cause**: lockfile drift — fix is `npm install` at repo root + commit `package-lock.json`
- 🔴 **The PR description's own schema-validation example throws at construction** — generated schema has no `context.session`, so `context.session.*` policies + `tools` always fail validation (reproduced)
- 🔴 **`reload()` silently drops user-provided `schema`** — schema validation disabled after reload (reproduced)
- 🟠 "Durable rate limiting" via `appState` is **write-only** — restored counts are never read back (reproduced)
- 🟠 `_stateKey` namespacing fix is ineffective (constant `name` → instances still collide)
- ✅ 34/34 unit tests pass locally, eslint/tsc clean; recent refactors (cedar-wasm types, schema-generator extraction, EntityUidJson handling) are solid

<details>
<summary>🔴 CI Gate / Code Quality failure — lockfile drift (blocking, easy fix)</summary>

The `Code Quality` job fails with:
```
package-lock.json drifted from package.json -- run 'npm install' and commit the lockfile
```

Reproduced locally: the **root** `package-lock.json` still records `@cedar-policy/mcp-schema-generator-wasm` in the root `devDependencies` stanza:
```json
// package-lock.json packages[""].devDependencies
{ "@cedar-policy/mcp-schema-generator-wasm": "^0.6.0", "husky": "^9.1.7", "prettier": "^3.7.4" }
```
but root `package.json` only has `husky` + `prettier` — the dep was (correctly) moved into `strands-ts/package.json` devDependencies, and the lockfile wasn't regenerated afterwards. Running `npm install` at the root produces a 28-line lockfile diff.

**Fix:** `npm install` at repo root, commit `package-lock.json`.
</details>

<details>
<summary>🔴 Bug: schema validation rejects <code>context.session.*</code> policies — PR's own example fails</summary>

The "With schema validation from tool definitions" example in the PR description throws at construction. Verbatim repro:

```typescript
const cedar = await CedarAuthorization.create({
  policies: `
    permit(principal, action == Action::"search", resource);
    permit(principal, action == Action::"delete_record", resource)
    when { context.session.role == "admin" };
  `,
  tools: [searchTool, deleteTool],
  contextEnricher: ({ invocationState }) => ({ role: String(invocationState.role ?? 'none') }),
})
// ❌ Error: Cedar policy validation failed: policy1: for policy `policy1`,
//    attribute `session` in context for Action::"delete_record" not found
```

Root cause: the schema produced by `mcp-schema-generator-wasm` only declares `context: { input: ... }`:

```cedar
action "search" appliesTo {
  principal: [User],
  resource: [Resource],
  context: { input: searchInput }   // ← no `session`
};
```

but the handler always injects `context.session` (`hour_utc`, `call_count`, enricher fields) at evaluation time. So **any** policy referencing `context.session.*` (rate limiting, roles, environment gating — i.e. most of the documented use cases) cannot be combined with the `tools` schema-validation path. The existing tests don't catch this because none of the `create()`/schema tests use a `context.session` policy.

**Fix options:** post-process the generated schema to merge a `session` record into each action's context (`hour_utc: Long, call_count: Long` + open record for enricher fields), or have `generateSchema` accept extra context attributes. Either way, please add a test combining `tools` + a `context.session.call_count` policy.
</details>

<details>
<summary>🔴 Bug: <code>reload()</code> silently drops user-provided <code>schema</code></summary>

```typescript
reload(): void {
  const policies = loadPolicies(this._policySource)
  const entities = loadEntities(this._entitySource)
  const schema = this._schemaGenerator && this._tools ? this._schemaGenerator.generateSchema(this._tools) : undefined
  //    ^^^^^ ignores config.schema — there is no `_schemaSource` field retained
  validatePolicies(policies, schema)
  ...
}
```

If the user passed `config.schema` directly (the non-`tools` path), `reload()` recomputes schema as `undefined` and **schema validation is silently disabled from then on**. Reproduced:

```typescript
// Construction with schema-violating policy throws ✅
// cedar.reload() with the SAME schema-violating policy file does NOT throw ❌
```

This defeats the hot-reload + fail-fast story: an operator edits the `.cedar` file with a typo'd action, calls `reload()`, and the typo sails through. **Fix:** retain `config.schema` as `_schemaSource` and reuse it in `reload()` (the atomicity of reload itself is fine — nice fix from the earlier round).
</details>

<details>
<summary>🟠 "Durable rate limiting" claim — appState persistence is write-only</summary>

The PR claims *"call counts stored on `agent.appState`, persists with session managers"*. The handler **writes** to `appState` on every increment/decrement, but **never reads from it**. The in-memory `_callCounts` map is the sole source of truth and starts empty for every new handler instance.

Reproduced: restore an agent with `appState = { 'cedar-authorization:cedar-authorization': { send: 1 } }` and a `call_count < 2` policy → the call **proceeds** (count restarts at 1) instead of being denied. So rate limits reset on every process restart / session restore — the persistence is cosmetic.

Also note the inverse problem: since `_callCounts` lives on the handler instance, a handler shared across multiple agents shares one counter pool (agent A's calls consume agent B's budget), while the appState mirror just reflects whichever agent called last.

**Fix:** read counts back from `event.agent.appState` in `_incrementCallCount` (making appState the source of truth, keyed per agent), or drop the durability claim from the PR description and TSDoc and document counts as per-instance/in-memory.
</details>

<details>
<summary>🟠 <code>_stateKey</code> namespacing is ineffective + smaller leftovers</summary>

**State key collision (claimed fixed, isn't):**
```typescript
readonly name = 'cedar-authorization'        // constant
this._stateKey = `cedar-authorization:${this.name}`  // always 'cedar-authorization:cedar-authorization'
```
The earlier review's multi-instance collision concern was addressed by namespacing with `this.name` — but `name` is a fixed readonly, so two `CedarAuthorization` instances on one agent still write to the identical key. Either make the key configurable / instance-unique, or document single-instance-per-agent.

**Still-open threads from the last review round (all confirmed present at HEAD `0cdf391e`):**
- `console.warn` in `create()` (cedar.ts:153) — only source file bypassing the SDK `logger`
- bare `catch {}` (cedar.ts:152) masks a broken/corrupted install as "not installed"
- `parseEntityUid` falls back to `type: 'Action'` for unparseable **resource** UIDs (schema-generator.ts:91)
- `schema-generator.ts` exports (`SchemaGenerator`, `createSchemaGenerator`) have zero TSDoc (AGENTS.md requires it)
- peer dep `^4.0.0` vs devDep `^4.11.0` gap — 4.0–4.10 is untested territory
</details>

<details>
<summary>✅ Verified locally</summary>

- `npm ci` + full cedar unit suite: **34/34 pass** (real cedar-wasm evaluation, ~0.5s)
- `eslint` on `vended-interventions/cedar/`: clean
- `tsc --noEmit`: no cedar errors
- Recent refactor commits (`d43f94e5..0cdf391e`) reviewed: adopting `CedarValueJson`/`TypeAndId`/`EntityJson` from cedar-wasm instead of hand-rolled types is the right call; `EntityUidJson` union (`__entity` wrapper) handling in `loadEntities` is correct; schema-generator extraction into its own module is clean; `reload()` is now atomic for policies/entities (validate-then-assign 👍); decrement-on-deny works including the Cedar-evaluation-failure path
- The reworked `resetCallCounts` test now meaningfully exercises deny→reset→allow on the same handler 👍
</details>

**Recommendation:** fix the lockfile (unblocks CI), then address the two 🔴 schema bugs — both are in the newly-added code paths this iteration introduced, and both have one-line-ish fixes plus a missing test each. The appState read-back can be a fast-follow if the description/TSDoc wording is softened. Happy to provide my repro tests if useful.

lizradway:
> _stateKey collision: _stateKey is a constant ('cedar-authorization'), which is stable across restarts (good for persistence) but means two CedarAuthorization instances on the same agent share/overwrite one counter. Worth either making it configurable or documenting single-instance-per-agent.


User will not have two CedarAuthorization on the same agent. the InterventionRegistry is a set/drops duplicates, so this is impossible.

strands-ts/src/vended-interventions/cedar/cedar.ts
```diff
@@ -0,0 +1,358 @@
...
+  private readonly _principal: TypeAndId | undefined
+  private readonly _principalResolver: ((invocationState: Record<string, unknown>) => TypeAndId | undefined) | undefined
+  private readonly _contextEnricher: CedarAuthorizationConfig['contextEnricher']
+  private readonly _tools: ToolDefinition[] | undefined
+  private readonly _schemaGenerator: SchemaGenerator | undefined
+  private readonly _callCounts = new Map<string, number>()
+  private static readonly _stateKey = 'cedar-authorization'
+
+  constructor(config: CedarAuthorizationConfig, schemaGenerator?: SchemaGenerator) {
+    super()
```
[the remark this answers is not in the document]
lizradway: this is impossible, intervention reigstry filters out duplicate

strands-ts/src/vended-interventions/cedar/schema-generator.ts
```diff
@@ -0,0 +1,97 @@
...
+    configJson?: string
+  ) => string
+}): SchemaGenerator {
+  const defaultStub = `
+namespace Agent {
+  @mcp_principal
+  entity User;
+  @mcp_resource
+  entity Resource;
+}
```
victornicolet, line 23: It could be useful to extend the defaultStub with at least the builtin context enrichment. 
For example:
```
namespace Agent {
  @mcp_principal
  entity User;
  @mcp_resource
  entity Resource;
  @mcp_context("session")
    type SessionContext = {
      hour_utc: Long,
      call_count: Long
    };
  }
```
would match what the default context enrichment adds and provide additional validation capability

strands-ts/src/vended-interventions/cedar/__tests__/cedar.test.node.ts
```diff
@@ -0,0 +1,769 @@
...
+import { CedarAuthorization } from '../cedar.js'
+import { Agent } from '../../../agent/agent.js'
+import { MockMessageModel } from '../../../__fixtures__/mock-message-model.js'
+import { createMockTool } from '../../../__fixtures__/tool-helpers.js'
+import { resolve } from 'node:path'
+import { writeFileSync, unlinkSync, existsSync } from 'node:fs'
+
+const FIXTURES = resolve(import.meta.dirname!, 'fixtures')
+
+describe('CedarAuthorization', () => {
```
victornicolet, line 11: It would be useful to have an additional test for tools with inputs that have nested properties; I think it might break the current implementation because of how nested inputs are handled with nested namespaces, and the namespace stripping. 
E.g. a test with something like:
```js
{
   name: 'create_user',
   inputSchema: {
   type: 'object',
     properties: {
      name: { type: 'string' },
              address: {
                type: 'object',
                 properties: {
                   street: { type: 'string' },
                   city: { type: 'string' },
                 },
                 required: ['street', 'city'],
               },
             },
             required: ['name', 'address'],
           },
         }
```
Another question would be whether that namespace stripping is desirable.

strands-ts/src/vended-interventions/cedar/schema-generator.ts
```diff
@@ -0,0 +1,93 @@
...
+      const config = JSON.stringify({ flattenNamespaces: true })
+      const result = JSON.parse(wasm.generateSchema(defaultStub, JSON.stringify(tools), config)) as {
+        schema: string | null
+        error: string | null
+        isOk: boolean
+      }
+      if (!result.isOk || !result.schema) {
+        throw new Error(`Schema generation failed: ${result.error}`)
+      }
+      return result.schema.replace(/^namespace\s+\w+\s*\{/, '').replace(/\}\s*$/, '')
```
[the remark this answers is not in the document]
victornicolet: What is the high-level design choice behind the namespace stripping? It is to make it easier to write policies based on what they know (tool names) and not have to care about the underlying Cedar schema?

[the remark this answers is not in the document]
victornicolet: I think that if the namespace stripping is really desired, we should find a solution on the schema generator's end to make sure the schema is consistent.

strands-ts/src/vended-interventions/cedar/schema-generator.ts
victornicolet: In wonder if it's possible to also automatically derive the stub from the user-provided context enrichment, if we know the properties being added and their types. But that could be a later addition.

lizradway: good idea -- done!

strands-ts/src/vended-interventions/cedar/__tests__/cedar.test.node.ts
lizradway: added!

lizradway: > Another question would be whether that namespace stripping is desirable.

I think namespace stripping is desirable for _Strands_, since we optimize for flat devx I wouldn't want to make users write namespaced policies. but lmk if you think otherwise.

strands-ts/src/vended-interventions/cedar/schema-generator.ts
lizradway: >In wonder if it's possible to also automatically derive the stub from the user-provided context enrichment, if we know the properties being added and their types. But that could be a later addition. 

yeah would definetely be interested in this, but i think might be a follow up for now to prioritize this for summit

strands-ts/src/vended-interventions/cedar/__tests__/cedar.test.node.ts
```diff
@@ -0,0 +1,787 @@
...
+        ...tools,
+        {
+          name: 'create_user',
+          inputSchema: {
+            type: 'object',
+            properties: {
+              name: { type: 'string' },
+              address: { type: 'object', properties: { street: { type: 'string' }, city: { type: 'string' } } },
+            },
+          },
```
victornicolet, line 758: This nested input should break the schema generator when flattening namespaces because of name collisions (at generation time):
```
{
          name: 'create_user',
          inputSchema: {
            type: 'object',
            properties: {
              address: {
                type: 'object',
                properties: {
                  street: { type: 'string' },
                  geo: {
                    type: 'object',
                    properties: {
                      lat: { type: 'number' },
                      lng: { type: 'number' },
                    },
                  },
                },
              },
              address_geo: {
                type: 'object',
                properties: {
                  lat: { type: 'string' },
                  lng: { type: 'string' },
                },
              },
            },
          },
        },
```
Note how we have `geo` nested in `address` in one prop, and `address_geo` in the other. I think the validation and schema generation should handle this gracefully at least, and we can brainstorm on a fix / providing different behavior on the schema generator's side.

strands-ts/src/vended-interventions/cedar/__tests__/cedar.test.node.ts
victornicolet: I makes sense, developers don't want to see complexity added because of using a specific tool. We might want to support this use case in the mcp-schema-generator directly if that's the case, to avoid regex-replacement manipulation

strands-ts/src/vended-interventions/cedar/__tests__/cedar.test.node.ts
lizradway: interesting. added test for fail gracefully on our end for now

strands-ts/package.json
```diff
@@ -220,10 +214,9 @@
     "@a2a-js/sdk": "^0.3.10",
     "@ai-sdk/provider": "^3.0.0",
     "@anthropic-ai/sdk": "^0.92.0",
-    "@aws-sdk/client-bedrock-agent": "^3.943.0",
-    "@aws-sdk/client-bedrock-agent-runtime": "^3.943.0",
     "@aws-sdk/client-s3": "^3.943.0",
     "@aws/bedrock-token-generator": "^1.1.0",
+    "@cedar-policy/cedar-wasm": "^4.0.0",
```
pgrayy, line 237: Do we also want to add `"@cedar-policy/mcp-schema-generator-wasm": "^0.6.0",` as a peer dependency?

strands-ts/src/vended-interventions/cedar/cedar.ts
```diff
@@ -0,0 +1,343 @@
...
+    }
+    this._principalResolver = config.principalResolver
+    this._contextEnricher = config.contextEnricher
+    this.onError = config.onError ?? 'throw'
+
+    validatePolicies(this._policies, this._schema, schemaIsAutoGenerated)
+  }
+
+  /** Async factory — use when passing `tools` for auto schema generation. */
+  static async create(config: CedarAuthorizationConfig): Promise<CedarAuthorization> {
```
pgrayy: What if we followed the Agent.initialize pattern and called it on their behalf. So users would only initialize through `CedarAuthorization` and under the hood, when we get the chance, we call the async initialization.

strands-ts/package.json
lizradway: i kept it optional since users only need it installed if they are passing `tools` on the `CedarAuthorization` class.

so yes probably. i think i removed it bc it was failing ci for some reason lol.

lizradway: updated

strands-ts/src/vended-interventions/cedar/cedar.ts
lizradway: i do like the initialization pattern better from devX, but then schema couldn't be validated on construction.

interventions doesn't have `initAgent` right now ( and I'd like to avoid that until we _need_ it). any other ideas on how to do this?

lizradway: updated to make mcp schema generator non-lazy loaded

pgrayy (dismissed):

pgrayy (approved):

strands-ts/src/vended-interventions/cedar/cedar.ts
```diff
@@ -0,0 +1,318 @@
...
+ * @see {@link https://cedarpolicy.com | Cedar} · {@link https://github.com/cedar-policy/cedar-for-agents | cedar-for-agents}
+ *
+ * @example
+ * ```typescript
+ * const cedar = new CedarAuthorization({
+ *   policies: 'permit(principal, action == Action::"search", resource);',
+ * })
+ *
+ * // With schema validation:
+ * const cedar = await CedarAuthorization.create({ policies: '...', tools: [searchTool] })
```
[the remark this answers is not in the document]
lizradway: i can fix this on monday when i merge in the builder

victornicolet (approved):

PR #2370, opened 2026-05-29
feat(mcp): promote content-to-tool-result method to public API

### LINKED ISSUE #2251: feat(mcp): document/promote MCPClient._map_mcp_content_to_tool_result_content as a public extension point
### Problem Statement

`MCPClient._map_mcp_content_to_tool_result_content` (in `strands/tools/mcp/mcp_client.py`) is the natural extension point for applications that need to intercept and reroute specific MCP content blocks — e.g. taking an `EmbeddedResource` with a custom MIME type and surfacing it on an application-side channel instead of (or in addition to) the LLM's tool result.

The method's leading underscore signals it is a private API. Applications that subclass `MCPClient` to override it (a clean and minimal pattern) take on a maintenance risk on every release, even though the override is the most appropriate way to extend the existing logic added in #726.

### Proposed Solution

Promote `_map_mcp_content_to_tool_result_content` to a documented public extension point. Either:

- **Option A (smallest):** Drop the leading underscore (`map_mcp_content_to_tool_result_content`), keep the underscored name as a deprecated alias for one release, and add a note to the `MCPClient` docstring that subclassing is a supported extension pattern.
- **Option B (more invasive):** Introduce a `ContentMapper` protocol/strategy that `MCPClient` accepts as a constructor argument, with the current logic as the default. Useful if users want to compose mappers without subclassing.

A short docstring + a reference in the user docs would be enough on its own — the rename is a nice-to-have.

### Use Case

We're building a chat application where MCP tools produce visual artifacts — inline charts and multi-element reports — alongside their text output. An MCP tool returns:

- A `TextContent` block with a markdown summary table (the LLM reasons over this).
- An `EmbeddedResource` block whose `resource` is `TextResourceContents` with a custom MIME (`application/vnd.visualmind.artifact+json`) carrying the artifact envelope (the UI renders this).

After #726, Strands passes the `EmbeddedResource` through correctly as a `{\"text\": resource.text}` block. We need a hook to:

1. Peel that block out of the tool result before the LLM sees it (the envelope JSON is verbose and not useful for reasoning).
2. Forward it to our own SSE artifact channel for UI rendering.
3. Replace it with a short \"[chart rendered, do not redraw]\" marker so the LLM has reasoning context without the JSON.

We've prototyped this as a ~30-line `MCPClient` subclass overriding `_map_mcp_content_to_tool_result_content`. It works cleanly. The only friction is the leading underscore — it implies we shouldn't be doing this, when in practice it's the right pattern.

### Alternatives Solutions

1. **Wrap the MCP `ClientSession` at the transport layer.** Fully decoupled from Strands internals, but more invasive — replaces more of Strands' MCP plumbing. We've kept this as a fall-back if Strands ever renames the private method.
2. **Smuggle the artifact JSON in `TextContent`** with sentinel delimiters instead of `EmbeddedResource`. Loses MCP's structural advantages and requires brittle text parsing.
3. **Status quo** — continue overriding the private method, accepting the implicit \"you shouldn't do this\" signal.

### Additional Context

Related: #535 (resolved in favour of Option 1 — Protocol Completeness as Default). Promoting this method to a public extension point is consistent with that direction. It doesn't add a new feature gate or opt-in flag; it just acknowledges that the existing mapping logic is something application authors will reasonably want to extend, and the public-API contract should reflect that.

Happy to send a PR if maintainers are open to either Option A or Option B — would just like to confirm direction first.

gautamsirdeshmukh:
## Description

Renames `MCPClient._map_mcp_content_to_tool_result_content` to `map_mcp_content_to_tool_result_content`, making subclass override an officially supported extension pattern for intercepting or transforming MCP content blocks before they reach the model.

(Also maintains the underscored version with backwards compatibility with a `DeprecationWarning`)

## Related Issues

Closes #2251

## Type of Change

New feature

## Testing

- [x] I ran `hatch run prepare`, added a test

## Checklist

----

### DIFF
--- a/strands-py/src/strands/tools/mcp/mcp_client.py
+++ b/strands-py/src/strands/tools/mcp/mcp_client.py
@@ -749,7 +749,7 @@ class MCPClient(ToolProvider):
         mapped_contents: list[ToolResultContent] = [
             mc
             for content in call_tool_result.content
-            if (mc := self._map_mcp_content_to_tool_result_content(content)) is not None
+            if (mc := self.map_mcp_content_to_tool_result_content(content)) is not None
         ]
 
         status: ToolResultStatus = "error" if call_tool_result.isError else "success"
@@ -864,14 +864,16 @@ class MCPClient(ToolProvider):
         asyncio.set_event_loop(self._background_thread_event_loop)
         self._background_thread_event_loop.run_until_complete(self._async_background_thread())
 
-    def _map_mcp_content_to_tool_result_content(
+    def map_mcp_content_to_tool_result_content(
         self,
         content: MCPTextContent | MCPImageContent | MCPEmbeddedResource | Any,
     ) -> ToolResultContent | None:
         """Maps MCP content types to tool result content types.
 
         This method converts MCP-specific content types to the generic
-        ToolResultContent format used by the agent framework.
+        ToolResultContent format used by the agent framework. Subclasses can
+        override this to intercept or transform specific content blocks before
+        they reach the model.
 
         Args:
             content: The MCP content to convert
@@ -945,6 +947,7 @@ class MCPClient(ToolProvider):
         else:
             self._log_debug_with_thread("unhandled content type: %s - dropping content", content.__class__.__name__)
             return None
+
 
     def _log_debug_with_thread(self, msg: str, *args: Any, **kwargs: Any) -> None:
         """Logger helper to help differentiate logs coming from MCPClient background thread."""
--- a/strands-py/tests/strands/tools/mcp/test_mcp_client.py
+++ b/strands-py/tests/strands/tools/mcp/test_mcp_client.py
@@ -1165,3 +1165,29 @@ async def test_call_tool_async_elicitation_error(mock_transport, mock_session):
         assert "https://example.com/auth" in result["content"][0]["text"]
         assert "Please authorize the application" in result["content"][0]["text"]
         assert "elicit-123" in result["content"][0]["text"]
+
+
+def test_map_mcp_content_subclass_override(mock_transport, mock_session):
+    """Subclass override of map_mcp_content_to_tool_result_content is invoked by the call site."""
+
+    class CustomMCPClient(MCPClient):
+        def map_mcp_content_to_tool_result_content(self, content):
+            result = super().map_mcp_content_to_tool_result_content(content)
+            if result and "text" in result:
+                result = {"text": "[intercepted]"}
+            return result
+
+    embedded_resource = {
+        "type": "resource",
+        "resource": {
+            "uri": "mcp://resource/test",
+            "text": "original text",
+            "mimeType": "text/plain",
+        },
+    }
+    mock_session.call_tool.return_value = MCPCallToolResult(isError=False, content=[embedded_resource])
+
+    with CustomMCPClient(mock_transport["transport_callable"]) as client:
+        result = client.call_tool_sync(tool_use_id="override-test", name="test_tool", arguments={})
+
+    assert result["content"][0]["text"] == "[intercepted]"


strands-py/src/strands/tools/mcp/mcp_client.py
```diff
@@ -749,7 +750,7 @@ def _handle_tool_result(self, tool_use_id: str, call_tool_result: MCPCallToolRes
         mapped_contents: list[ToolResultContent] = [
             mc
             for content in call_tool_result.content
-            if (mc := self._map_mcp_content_to_tool_result_content(content)) is not None
+            if (mc := self.map_mcp_content_to_tool_result_content(content)) is not None
```
opieter-aws, line 752: Is it intentional to always call the new method here? In this case, the user’s custom override would silently disappear, and because it never runs, no deprecation warning is emitted either. It would break behaviour with no warning.

Should we resolve here to always run the overridden version if specified, warn, and only remove after deprecating?

opieter-aws (dismissed):

agent-of-mkmeral:
**TL;DR:** Right direction, but the deprecation shim is inverted — existing subclasses that override the old `_map_...` name will **silently stop being invoked** (and won't even get a warning). Agreeing with @opieter-aws's CHANGES_REQUESTED. Fix + missing test below. Also worth a `needs-api-review` label since this promotes a private method to a public contract.

<details>
<summary><b>🔴 The deprecation bug (blocking)</b></summary>

The call site now dispatches through the new public name, and the old name is kept as a delegating wrapper:

```python
# call site:
mc := self.map_mcp_content_to_tool_result_content(content)

# old name delegates forward:
def _map_mcp_content_to_tool_result_content(self, content):
    warnings.warn(...)
    return self.map_mcp_content_to_tool_result_content(content)
```

Today the *only* extension point is overriding `_map_...`. After this PR, an existing subclass that overrides `_map_...` has its override **never called** — the call site goes through `map_...`. And because the wrapper itself is what got overridden, the `DeprecationWarning` never fires either. Silent behavior change, zero signal — the worst kind of break. The warning only triggers on an *explicit* `obj._map_...()` call, which nobody does (the method is meant to be overridden, not called).
</details>

<details>
<summary><b>✅ Suggested fix — runtime override detection</b></summary>

```python
def map_mcp_content_to_tool_result_content(self, content):
    cls = type(self)
    if (
        cls.map_mcp_content_to_tool_result_content is MCPClient.map_mcp_content_to_tool_result_content
        and cls._map_mcp_content_to_tool_result_content is not MCPClient._map_mcp_content_to_tool_result_content
    ):
        warnings.warn(
            "Overriding _map_mcp_content_to_tool_result_content is deprecated. "
            "Override map_mcp_content_to_tool_result_content instead.",
            DeprecationWarning, stacklevel=2,
        )
        return cls._map_mcp_content_to_tool_result_content(self, content)
    # ... real mapping logic ...
```

This keeps legacy overrides working *and* emits a migration warning during the deprecation window.
</details>

<details>
<summary><b>🧪 Missing test coverage</b></summary>

The two new tests don't cover the actual breakage:
- `test_map_mcp_content_subclass_override` only overrides the **new** name (trivially works).
- `test_deprecated_underscore_method_delegates_with_warning` only tests an **explicit call**.

Neither tests the realistic case: *a subclass that overrides the OLD name still gets invoked (with a warning)*. That's exactly where the bug lives — please add it.
</details>

<details>
<summary><b>💡 Bigger-picture (non-blocking)</b></summary>

- **API review:** promoting a private method to a documented public contract is a one-way door — suggest adding `needs-api-review` rather than landing as `size/s`.
- **Mechanism:** subclassing is fine for the #2251 use case but is the least composable option (can't stack transforms). The SDK already has `AfterToolCallEvent` (writable `result`) which overlaps at the agent layer — worth documenting *when to use which* (MCP-client method = per-content-block, pre-mapping; hook = post-mapping, composable). An injectable `ContentMapper` (Option B in #2251) would be the cleaner v2 if this grows.
</details>

gautamsirdeshmukh:
Still discussing offline about the right direction to go in terms of when to make the breaking change

strands-py/src/strands/tools/mcp/mcp_client.py
```diff
@@ -864,14 +865,16 @@ def _background_task(self) -> None:
         asyncio.set_event_loop(self._background_thread_event_loop)
         self._background_thread_event_loop.run_until_complete(self._async_background_thread())
 
-    def _map_mcp_content_to_tool_result_content(
+    def map_mcp_content_to_tool_result_content(
```
zastrowm, line 867: Do we have to care about backwards compatibility?  It might be worth opting into breaking existing users who are using the private method

strands-py/src/strands/tools/mcp/mcp_client.py
gautamsirdeshmukh: Discussed offline, we will actually fully update to the non-underscored public method and NOT maintain / throw a deprecation warning for the underscored internal type.

strands-py/src/strands/tools/mcp/mcp_client.py
gautamsirdeshmukh: Yep, updated to just use the new public version across the board

pgrayy (approved):

PR #2388, opened 2026-05-29
ci: add explicit permissions to workflow jobs

yonib05:
## Summary

- Add `permissions: {}` to `ci-gate` job in `ci.yml` (uses no token, so empty permissions is correct)
- Add `permissions: { contents: read, issues: write }` to `check-links` job in `python-check-markdown-links.yml`

## Test plan

- [ ] Verify CI workflow still passes (ci-gate job uses no token)
- [ ] Verify markdown link checker still works with the declared permissions

### DIFF
--- a/.github/workflows/python-check-markdown-links.yml
+++ b/.github/workflows/python-check-markdown-links.yml
@@ -8,6 +8,9 @@ on:
 jobs:
   check-links:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      issues: write
     steps:
       - uses: actions/checkout@v6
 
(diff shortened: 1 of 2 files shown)


zastrowm (dismissed):

zastrowm (approved):

.github/workflows/python-check-markdown-links.yml
```diff
@@ -8,6 +8,9 @@ on:
 jobs:
   check-links:
```
zastrowm, line 9: Does this actually work?  I didn't even know this was a workflow we maintained

yonib05: Maybe it wasn't ported to the new site? not sure either way.

PR #1703, opened 2026-02-16
fix: propagate reasoningSignature on Gemini tool use

afarntrog:
## Description
- Add reasoningSignature field to ToolUse TypedDict and streaming handlers
- Pass reasoning signatures through content block start/stop in event loop
- Fix Gemini model to use base64 encode/decode for thought_signature instead of raw UTF-8, ensuring correct binary data handling
- Thread reasoning signatures on function_call parts for Gemini

## Related Issues

Fixes https://github.com/strands-agents/sdk-python/pull/1622

## Type of Change

Bug fix

## Testing

How have you tested the change?  Verify that the changes do not break functionality or introduce warnings in consuming repositories: agents-docs, agents-tools, agents-cli

- [ ] I ran `hatch run prepare`

## Checklist
- [ ] I have read the CONTRIBUTING document
- [ ] I have added any necessary tests that prove my fix is effective or my feature works
- [ ] I have updated the documentation accordingly
- [ ] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2397, opened 2026-05-29
fix: report malformed tool input JSON

### LINKED ISSUE #2051: bug: Malformed tool input JSON silently replaced with empty dict, no logging
## Description

In `streaming.py`, when a model returns invalid JSON as tool input, the `ValueError` is caught and the input is silently replaced with an empty dict `{}`. There is no log message or warning.

**`src/strands/event_loop/streaming.py`, lines 279-282:**
```python
try:
    current_tool_use["input"] = json.loads(current_tool_use["input"])
except ValueError:
    current_tool_use["input"] = {}
```

The tool then executes with empty arguments, fails with a missing-parameter error, and the model has no indication that its JSON was malformed. It retries the same broken JSON repeatedly.

This happens most often with smaller models (Ollama, llama.cpp, Mistral local) that produce truncated or syntactically invalid JSON for tool inputs.

## Reproduction

Any model that returns invalid tool input JSON. Examples that trigger this:

- `{"query": "test` (truncated by max_tokens mid-tool-call)
- `{query: test}` (unquoted keys)
- empty string

All silently become `{}`.

## Suggested fix

Log a warning and return an informative error to the model instead of silently passing `{}` to the tool:

```python
except ValueError:
    logger.warning("Failed to parse tool input JSON for '%s': %s", current_tool_use.get("name"), current_tool_use["input"])
    current_tool_use["input"] = {"_parse_error": f"Invalid JSON in tool input: {current_tool_use['input'][:200]}"}
```

This way the tool can detect the parse error, and the model gets feedback that its JSON was malformed rather than thinking the tool itself is broken.

Happy to submit a PR for this.

he-yufeng:
﻿## Summary
- mark malformed streamed tool input JSON instead of replacing it with `{}`
- turn that parse marker into an error tool result before tool execution
- keep missing-input tool uses on the existing `{}` path

Fixes #2051.

## To verify
- `python -m pytest tests/strands/event_loop/test_streaming.py tests/strands/tools/test_validator.py -q`
- `python -m ruff check src/strands/event_loop/streaming.py src/strands/tools/_validator.py tests/strands/event_loop/test_streaming.py tests/strands/tools/test_validator.py`
- `python -m ruff format --check src/strands/event_loop/streaming.py src/strands/tools/_validator.py tests/strands/event_loop/test_streaming.py tests/strands/tools/test_validator.py`
- `git diff --check`

### DIFF
--- a/strands-py/src/strands/event_loop/streaming.py
+++ b/strands-py/src/strands/event_loop/streaming.py
@@ -10,6 +10,7 @@ from typing import Any
 
 from ..models.model import Model
 from ..tools import InvalidToolUseNameException
+from ..tools._validator import TOOL_INPUT_PARSE_ERROR_KEY
 from ..tools.tools import validate_tool_use_name
 from ..types._events import (
     CitationStreamEvent,
@@ -282,12 +283,22 @@ def handle_content_block_stop(state: dict[str, Any]) -> dict[str, Any]:
 
     if current_tool_use:
         if "input" not in current_tool_use:
-            current_tool_use["input"] = ""
-
-        try:
-            current_tool_use["input"] = json.loads(current_tool_use["input"])
-        except ValueError:
             current_tool_use["input"] = {}
+        else:
+            try:
+                current_tool_use["input"] = json.loads(current_tool_use["input"])
+            except ValueError as e:
+                logger.warning(
+                    "tool_name=<%s>, tool_use_id=<%s> | failed to parse tool input JSON",
+                    current_tool_use.get("name"),
+                    current_tool_use.get("toolUseId"),
+                )
+                current_tool_use["input"] = {
+                    TOOL_INPUT_PARSE_ERROR_KEY: (
+                        f"Invalid JSON in tool input for '{current_tool_use.get('name', 'unknown')}': {e}. "
+                        "Retry with a valid JSON object."
+                    )
+                }
 
         tool_use_id = current_tool_use["toolUseId"]
         tool_use_name = current_tool_use["name"]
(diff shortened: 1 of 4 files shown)


strands-py/src/strands/event_loop/streaming.py
```diff
@@ -282,12 +283,22 @@ def handle_content_block_stop(state: dict[str, Any]) -> dict[str, Any]:
...
+            try:
+                current_tool_use["input"] = json.loads(current_tool_use["input"])
+            except ValueError as e:
+                logger.warning(
+                    "tool_name=<%s>, tool_use_id=<%s> | failed to parse tool input JSON",
+                    current_tool_use.get("name"),
+                    current_tool_use.get("toolUseId"),
+                )
+                current_tool_use["input"] = {
+                    TOOL_INPUT_PARSE_ERROR_KEY: (
```
pgrayy, line 297: Rather than encoding the parse failure as a magic key inside `input` (`{TOOL_INPUT_PARSE_ERROR_KEY: ...}`), I'd suggest a dedicated `NotRequired[str]` field on `ToolUse`, e.g. `inputParseError`, mirroring how `reasoningSignature` was added in #1703. Streaming sets it on the `ValueError` path (keeping `input = {}`), and `validate_and_prepare_tools` reads it to emit the error tool result. This keeps `input` meaning only "the tool's arguments" so nothing downstream mistakes control metadata for real input, drops the `isinstance(dict)` + key-sniff check in favor of a plain `tool.get("inputParseError")`, and removes the cross-module sentinel constant. Since the marker is only a within-cycle bridge from streaming to the validator, we can try to `pop` it at the point of consumption so it never persists in history.

he-yufeng:
Bump. A malformed tool-input JSON blob still reaches the model as an opaque failure without the parse detail on current main (checked streaming.py today), so retries flail on the same bad payload. The patch reports the JSON error into the tool result so the next turn can self-correct.

opieter-aws:
@he-yufeng your PR was reviewed by a maintainer on June 25, can you please address the comment and failing CI so we can try to get this merged? Thanks!

PR #2412, opened 2026-06-01
feat(strands-py-wasm): add DecoratedTool for host-side Python tools

pgrayy:
## Description

Introduce `DecoratedTool` and the `@tool` decorator as the way to register Python callbacks the agent can invoke. Replaces the prior `Tool` / `PydanticTool` pair with a single user-facing type.

```python
from strands import Agent, BedrockModel, tool

@tool
def get_weather(city: str) -> str:
    """Return the current weather for a city."""
    return f"It is 72F and sunny in {city}."

agent = Agent(model=BedrockModel(...), tools=[get_weather])
```

On the guest (`strands-wasm/entry.ts`), fix `mapContentBlock` so the `toJSON()` outer wrappers from SDK content blocks are stripped to match the WIT inner-record shape, and add `mapToolResultBlock` / `mapToolResultContent` helpers for tool-result events. Without this, tool-use events trapped the wasm module on the very first model tool call.

Pin `pgrayy-wasmtime` to the `wasmtime-v46.0.7` release wheel so installation no longer requires a git clone with submodule resolution.

## Related Issues

strands-agents/sdk-python#2421
strands-agents/sdk-python#2374

## Documentation PR

N/A

## Type of Change

New feature

## Testing

- Added `tests/test_tools.py::test_decorated_tool_invocation`. Registers `@tool def get_weather(...)`, asks the model "What is the weather in Seattle?", awaits `agent.invoke_async`, and asserts the tool's value reached the final response (`"72" in str(result)`). Exercises the full host→guest→host round trip: tool registration, model-issued tool call, Python callback execution, result marshaled back to the model, and final assistant response.
- Verified existing `tests/test_agent.py::test_stream_async_hello_world` continues to pass.
- Manually inspected the model's full final response to confirm the tool's content was incorporated naturally (e.g., "The weather in Seattle is currently **72°F and sunny**. It's a nice day!").

## Checklist

----

### DIFF
--- a/strands-py-wasm/pyproject.toml
+++ b/strands-py-wasm/pyproject.toml
@@ -28,8 +28,8 @@ classifiers = [
     "Topic :: Software Development :: Libraries :: Python Modules",
 ]
 dependencies = [
-    # Imports as `wasmtime`. Pinned to pgrayy/wasm-deps git URL until upstream PRs land.
-    "pgrayy-wasmtime @ git+https://github.com/pgrayy/wasm-deps.git@4b5dc41512109ebafe4c4f1edd592c739872c640#subdirectory=wasmtime-py",
+    # Imports as `wasmtime`. Pinned to a pgrayy/wasm-deps release wheel until upstream PRs land.
+    "pgrayy-wasmtime @ https://github.com/pgrayy/wasm-deps/releases/download/wasmtime-v46.0.7/pgrayy_wasmtime-46.0.7-py3-none-macosx_11_0_arm64.whl",
     "botocore>=1.29.0,<2.0.0",
 ]
 
--- a/strands-py-wasm/src/strands/__init__.py
+++ b/strands-py-wasm/src/strands/__init__.py
@@ -269,7 +269,7 @@ class InterruptResponse(types.RespondArgs):
 class DecoratedTool:
     """A Python function exposed to the agent as a tool.
 
-    Build one with the :func:`tool` decorator, then pass it to
+    Build one with the ``@tool`` decorator, then pass it to
     :class:`Agent` via ``tools=[...]``. The agent will call the function
     when the model invokes the tool by name.
 
@@ -308,10 +308,10 @@ class DecoratedTool:
     def invoke(self, raw_input: str) -> list[types.ToolResultContent]:
         """Run the tool with a JSON object of keyword arguments."""
         kwargs = json.loads(raw_input) if raw_input else {}
-        return _coerce_tool_result(self.func(**kwargs))
+        return _normalize_tool_result(self.func(**kwargs))
 
 
-def _coerce_tool_result(result: Any) -> list[types.ToolResultContent]:
+def _normalize_tool_result(result: Any) -> list[types.ToolResultContent]:
     if isinstance(result, str):
         return [types.ToolResultContent.Text(types.TextBlock(text=result))]
     if isinstance(result, types.TextBlock):
--- a/strands-py-wasm/src/strands/__init__.py
+++ b/strands-py-wasm/src/strands/__init__.py
@@ -266,40 +266,24 @@ class InterruptResponse(types.RespondArgs):
         super().__init__(interrupt_id=interrupt_id, response=payload)
 
 
-class PydanticTool:
-    """Tool whose input schema is derived from a pydantic ``BaseModel``."""
+class DecoratedTool:
+    """A Python function exposed to the agent as a tool.
 
-    def __init__(
-        self,
-        *,
-        name: str,
-        description: str,
-        input_model: type,
-        func: Callable[..., Any],
-    ) -> None:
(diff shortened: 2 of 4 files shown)


strands-py-wasm/pyproject.toml
```diff
@@ -28,8 +28,8 @@ classifiers = [
     "Topic :: Software Development :: Libraries :: Python Modules",
 ]
 dependencies = [
-    # Imports as `wasmtime`. Pinned to pgrayy/wasm-deps git URL until upstream PRs land.
-    "pgrayy-wasmtime @ git+https://github.com/pgrayy/wasm-deps.git@4b5dc41512109ebafe4c4f1edd592c739872c640#subdirectory=wasmtime-py",
+    # Imports as `wasmtime`. Pinned to a pgrayy/wasm-deps release wheel until upstream PRs land.
+    "pgrayy-wasmtime @ https://github.com/pgrayy/wasm-deps/releases/download/wasmtime-v46.0.7/pgrayy_wasmtime-46.0.7-py3-none-macosx_11_0_arm64.whl",
```
pgrayy, line 32: Will switch to source from [pypi](https://pypi.org/project/pgrayy-wasmtime/) in a follow up.

strands-py-wasm/src/strands/__init__.py
```diff
@@ -266,40 +266,24 @@ def __init__(self, *, interrupt_id: str, response: Any) -> None:
         super().__init__(interrupt_id=interrupt_id, response=payload)
 
 
-class PydanticTool:
```
pgrayy, line 269: If we want users to define tool shapes with a pydantic model, I feel that should be an enhancement to DecoratedTool. Whatever the case, right now we are trying to model what features we already have which is a decorated tool, not a pydantic tool

strands-py-wasm/pyproject.toml
mehtarac: Have a similar change in my PR to source from pypi: https://github.com/strands-agents/sdk-python/pull/2409

strands-py-wasm/src/strands/__init__.py
```diff
@@ -266,40 +266,24 @@ def __init__(self, *, interrupt_id: str, response: Any) -> None:
...
-        func: Callable[..., Any],
-    ) -> None:
-        if not hasattr(input_model, "model_json_schema") or not hasattr(input_model, "model_validate"):
-            raise TypeError(f"input_model must be a pydantic BaseModel subclass; got {input_model!r}")
-        self.name = name
-        self.description = description
-        self._input_model = input_model
-        self.input_schema = input_model.model_json_schema()
-        self.func = func
+    Build one with the :func:`tool` decorator, then pass it to
```
mehtarac: nit: `@tool` instead of `tool`?

strands-py-wasm/src/strands/__init__.py
```diff
@@ -266,40 +266,24 @@ def __init__(self, *, interrupt_id: str, response: Any) -> None:
...
+    :class:`Agent` via ``tools=[...]``. The agent will call the function
+    when the model invokes the tool by name.
 
-    def to_spec(self) -> types.ToolSpec:
-        return types.ToolSpec(
-            name=self.name,
-            description=self.description,
-            input_schema=json.dumps(self.input_schema),
-        )
+    Callbacks must be synchronous. Async functions are not yet supported.
```
mehtarac, line 276: maybe worth adding context if possible for helping our future selves. issue/PR where we know more as to why callbacks must be synchronous.

strands-py-wasm/src/strands/__init__.py
```diff
@@ -321,12 +305,13 @@ def to_spec(self) -> types.ToolSpec:
...
 
-    def invoke(self, raw_input: str) -> list[Any]:
+    def invoke(self, raw_input: str) -> list[types.ToolResultContent]:
+        """Run the tool with a JSON object of keyword arguments."""
         kwargs = json.loads(raw_input) if raw_input else {}
         return _coerce_tool_result(self.func(**kwargs))
 
 
-def _coerce_tool_result(result: Any) -> list[Any]:
+def _coerce_tool_result(result: Any) -> list[types.ToolResultContent]:
```
mehtarac: nit: "Coerce" feels like the input is wrong and needs to be made right. But all we're doing in this function is taking a perfectly valid return value and wrapping it in ToolResultContent format. Suggest maybe changing to `_normalize_tool_result`?

strands-py-wasm/src/strands/__init__.py
pgrayy: We have https://github.com/strands-agents/sdk-typescript/issues/1104 right now to centrally track gaps. I think documenting in issues as opposed to docstrings will be sufficient for right now.

mehtarac (approved):

PR #2544, opened 2026-06-01
feat(memory): add memory manager

opieter-aws:
## Description

Adds the `MemoryManager` primitive: a construct that manages one or more memory store
backends and exposes `search_memory` / `add_memory` tools for agent-driven recall and
persistence, plus programmatic `search()` / `add()` methods.

**What's included**
- `MemoryManager` (implements `Plugin`): registers its tools via `getTools()`, and
  aggregates any tools a store exposes via `MemoryStore.getTools()`.
- `MemoryStore` interface: every store is searchable; a required `writable` flag declares
  whether it accepts writes. `search_memory` targets all stores; `add_memory` targets only
  writable stores. The constructor fails fast if a store is `writable: true` without `add()`.
- `MemoryStoreConfig`: shared base config that concrete store implementations extend.
- Tool store scoping: the model may target a subset of stores by name; out-of-scope names
  are dropped (with a warning), or an actionable error is thrown if *all* requested names
  are out of scope. Omitted/empty `stores` targets all in-scope stores.
- Agent integration: new `AgentConfig.memoryManager` field accepting a `MemoryManager`
  instance or a `MemoryManagerConfig` object (auto-wrapped).

**Scope (intentionally minimal)**
This PR ships the `MemoryManager` and `MemoryStore` interface only, to unblock downstream
consumers (e.g. AgentCore memory) from building against a stable interface. Two pieces are
deferred to dedicated follow-up PRs:
- **Context injection** — passive pre-model-call memory injection. Deferred so it can be
  built on the upcoming middleware system (sdk-typescript#1068) rather than lifecycle hooks,
  which avoids mutating durable session history.
- **`BedrockKnowledgeBaseStore`** — a concrete store implementation, tracked separately
  (needs its own tests + optional AWS peer-dependency wiring).

`initAgent` is currently a no-op; tools auto-register through the `PluginRegistry` via
`getTools()`. Will be populated when adding injection and extraction.

## Related Issues

https://github.com/strands-agents/sdk-python/pull/2393

## Documentation PR

No new docs in this PR. A documentation PR will accompany the injection follow-up once the
full feature surface (injection + a shipped store backend) is finalized.

## Type of Change

New feature

## Public API

**`AgentConfig.memoryManager`**
```typescript
interface AgentConfig {
  // ...existing fields
  /** Accepts a MemoryManager instance or a config object (auto-wrapped). */
  memoryManager?: MemoryManager | MemoryManagerConfig
}
```

**`MemoryManager`**
```typescript
class MemoryManager implements Plugin {
  readonly name: string // 'strands:memory-manager'
  constructor(config: MemoryManagerConfig)

  /** Search across configured stores (unscoped programmatic access). */
  search(query: string, options?: SearchMemoryOptions): Promise<MemoryEntry[]>
  /** Add content to writable stores (unscoped programmatic access). */
  add(content: string, options?: AddMemoryOptions): Promise<void>
  /** Registers search_memory / add_memory tools plus any store-provided tools. */
  getTools(): Tool[]
}
```

**`MemoryStore` & `MemoryStoreConfig`**
```typescript
/** Shared identity/config fields — declared once, extended by both the store interface and concrete store configs. */
interface MemoryStoreConfig {
  readonly name: string                 // unique identifier, used for tool targeting
  readonly description?: string          // surfaced in tool descriptions
  readonly maxSearchResults?: number     // per-store default result limit
  readonly writable?: boolean            // caller intent; defaults to false
}

interface MemoryStore extends MemoryStoreConfig {
  readonly writable: boolean             // resolved (required at runtime)
  search(query: string, options?: SearchOptions): Promise<MemoryEntry[]>
  add?(content: string, metadata?: Record<string, JSONValue>): Promise<void> // required when writable
  getTools?(): Tool[]                    // optional store-provided tools
}
```

**Supporting types**
```typescript
interface MemoryEntry {
  content: string
  store?: string                         // filled out by MemoryManager.search
  metadata?: Record<string, JSONValue>
}

interface SearchOptions { maxSearchResults?: number }                          // per-store search input
interface SearchMemoryOptions { maxSearchResults?: number; stores?: string[] } // manager-level
interface AddMemoryOptions { metadata?: Record<string, JSONValue>; stores?: string[] }
interface MemoryToolConfig { name?: string; description?: string }

interface MemoryManagerConfig {
  stores: MemoryStore[]
  searchToolConfig?: MemoryToolConfig | boolean  // default: true
  addToolConfig?: MemoryToolConfig | boolean     // default: false (opt-in)
}
```

## Usage

```typescript
// Config shorthand (auto-wrapped) — agent gains search_memory (+ add_memory) tools
const agent = new Agent({
  model,
  memoryManager: { stores: [myStore], addToolConfig: true },
})

// Class instance — also gives programmatic access
const memoryManager = new MemoryManager({ stores: [myStore], addToolConfig: true })
const agent = new Agent({ model, memoryManager })
await memoryManager.search('user preferences')
await memoryManager.add('User prefers dark mode')

// Multi-store with read-only store — search_memory queries all; add_memory only writable
new MemoryManager({
  stores: [personalStore, teamStore /* writable: false */],
  addToolConfig: true,
})

// Custom tool names
new MemoryManager({
  stores: [myStore],
  searchToolConfig: { name: 'recall' },
  addToolConfig: { name: 'remember' },
})

// Custom store
class MyStore implements MemoryStore {
  readonly name = 'my-store'
  readonly writable = true
  async search(query: string, options?: SearchOptions): Promise<MemoryEntry[]> { /* ... */ }
  async add(content: string, metadata?: Record<string, JSONValue>): Promise<void> { /* ... */ }
}
```

## Testing

Comprehensive unit tests in `strands-ts/src/memory/__tests__/memory-manager.test.ts` (50
tests) cover: constructor validation (empty stores, duplicate names, writable-without-add,
add-tool enablement), tool registration, store-tool scoping (in-scope/out-of-scope/partial,
omit-vs-empty), `search()`/`add()` fan-out and partial-failure handling, store-provided
tool aggregation, and `AgentConfig` auto-wrapping.

- [ ] I ran `hatch run prepare` (N/A — TypeScript package; ran `npm test` / `type-check` /
      `lint` / `format:check` instead)

## Checklist

----

### DIFF
--- a/strands-ts/src/memory/memory-manager.ts
+++ b/strands-ts/src/memory/memory-manager.ts
@@ -8,6 +8,7 @@ import type {
   MemoryStore,
   MemoryAddOptions,
   MemoryToolConfig,
+  MemoryAddToolConfig,
 } from './types.js'
 import type { JSONValue } from '../types/json.js'
 import { tool } from '../tools/tool-factory.js'
@@ -59,7 +60,10 @@ export class MemoryManager implements Plugin {
   readonly name = 'strands:memory-manager'
   private readonly _config: MemoryManagerConfig
   private readonly _searchStores: MemoryStore[]
+  /** All writable stores — the unscoped target set for the programmatic {@link add} method. */
   private readonly _addStores: MemoryStore[]
+  /** Writable stores the `add_memory` tool may write to (a subset of `_addStores`). */
+  private readonly _addToolStores: MemoryStore[]
   private readonly _searchToolConfig: MemoryToolConfig | false
   private readonly _addToolConfig: MemoryToolConfig | false
   private readonly _awaitWrites: boolean
@@ -95,12 +99,39 @@ export class MemoryManager implements Plugin {
 
     if (config.addToolConfig === undefined || config.addToolConfig === false) {
       this._addToolConfig = false
+      this._addToolStores = []
     } else {
       if (this._addStores.length === 0) {
         throw new Error('MemoryManager: addToolConfig is enabled but no stores are writable')
       }
-      this._addToolConfig = typeof config.addToolConfig === 'object' ? config.addToolConfig : {}
+      const toolConfig: MemoryAddToolConfig = typeof config.addToolConfig === 'object' ? config.addToolConfig : {}
+      this._addToolConfig = toolConfig
+      this._addToolStores = this._resolveAddToolStores(toolConfig)
     }
+  }
+
+  /**
+   * Resolves the writable stores the `add_memory` tool may write to. When `stores` is given, each
+   * entry (a store name or a {@link MemoryStore} instance) must resolve by name to a configured,
+   * writable store (else throws). Omitted means all writable stores.
+   */
+  private _resolveAddToolStores(toolConfig: MemoryAddToolConfig): MemoryStore[] {
+    if (toolConfig.stores === undefined) {
+      return this._addStores
+    }
+
+    const names = toolConfig.stores.map((store) => (typeof store === 'string' ? store : store.name))
+
+    return [...new Set(names)].map((name) => {
+      const found = this._config.stores.find((s) => s.name === name)
+      if (!found) {
+        throw new Error(`MemoryManager: addToolConfig store '${name}' not found`)
+      }
+      if (!found.writable) {
+        throw new Error(`MemoryManager: addToolConfig store '${name}' is not writable`)
+      }
+      return found
+    })
   }
 
   /**
@@ -191,7 +222,6 @@ export class MemoryManager implements Plugin {
         continue
       }
       for (const entry of settledResult.value) {
-        // Stamp provenance so callers can tell which store produced each result.
         results.push({ ...entry, storeName })
       }
     }
@@ -355,14 +385,16 @@ export class MemoryManager implements Plugin {
 
   private _createAddTool(config: MemoryToolConfig): Tool {
     let description = config.description ?? ADD_TOOL_DESCRIPTION
-    const storeDescriptions = this._addStores.filter((s) => s.description).map((s) => `- ${s.name}: ${s.description}`)
+    const storeDescriptions = this._addToolStores
+      .filter((s) => s.description)
+      .map((s) => `- ${s.name}: ${s.description}`)
     if (storeDescriptions.length > 0) {
       description += `\n\nAvailable writable stores:\n${storeDescriptions.join('\n')}`
       description +=
-        '\n\nYou can target a specific store by name to route facts to the right place, or omit to add to all writable stores.'
+        '\n\nYou can target a specific store by name to route facts to the right place, or omit to add to all available writable stores.'
     }
 
-    const scopedNames = this._addStores.map((s) => s.name)
+    const scopedNames = this._addToolStores.map((s) => s.name)
 
     const inputSchema = z.object({
       entries: z.array(z.string()).min(1).describe('Data to add to long-term memory'),
--- a/strands-ts/src/memory/memory-manager.ts
+++ b/strands-ts/src/memory/memory-manager.ts
@@ -62,11 +62,9 @@ export class MemoryManager implements Plugin {
   private readonly _searchStores: MemoryStore[]
   /** All writable stores — the unscoped target set for the programmatic {@link add} method. */
   private readonly _addStores: MemoryStore[]
-  /** Writable stores the `add_memory` tool may write to (a subset of `_addStores`). */
-  private readonly _addToolStores: MemoryStore[]
   private readonly _searchToolConfig: MemoryToolConfig | false
-  private readonly _addToolConfig: MemoryToolConfig | false
-  private readonly _awaitWrites: boolean
+  private readonly _addToolConfig: MemoryAddToolConfig | false
+  private readonly _addToolStores: MemoryStore[]
 
   constructor(config: MemoryManagerConfig) {
     if (config.stores.length === 0) {
@@ -88,7 +86,6 @@ export class MemoryManager implements Plugin {
     this._config = config
     this._searchStores = config.stores
     this._addStores = config.stores.filter((s) => s.writable)
-    this._awaitWrites = config.awaitWrites ?? false
 
     this._searchToolConfig =
       config.searchToolConfig === false
@@ -104,9 +101,8 @@ export class MemoryManager implements Plugin {
       if (this._addStores.length === 0) {
         throw new Error('MemoryManager: addToolConfig is enabled but no stores are writable')
       }
-      const toolConfig: MemoryAddToolConfig = typeof config.addToolConfig === 'object' ? config.addToolConfig : {}
-      this._addToolConfig = toolConfig
-      this._addToolStores = this._resolveAddToolStores(toolConfig)
+      this._addToolConfig = typeof config.addToolConfig === 'object' ? config.addToolConfig : {}
+      this._addToolStores = this._resolveAddToolStores(this._addToolConfig)
     }
   }
 
@@ -159,7 +155,7 @@ export class MemoryManager implements Plugin {
     }
 
     if (this._addToolConfig !== false) {
-      tools.push(this._createAddTool(this._addToolConfig))
+      tools.push(this._createAddTool(this._addToolConfig, this._addToolStores))
     }
 
     for (const store of this._config.stores) {
@@ -231,20 +227,16 @@ export class MemoryManager implements Plugin {
   }
 
   /**
-   * Add content to writable stores. If `stores` is provided, only writes to those named stores.
+   * Add content to writable stores. If `stores` is provided, only writes to those named stores;
+   * otherwise all writable stores are targeted.
    *
-   * This method is unscoped, with full access to all configured writable stores.
-   * Tool-level store scoping is applied by the add tool callback.
-   * When `options.stores` is omitted, all writable stores are targeted.
-   *
-   * Target stores are always validated synchronously (an unknown or read-only named store throws
-   * immediately). The store writes themselves follow `awaitWrites` (resolved from
-   * {@link MemoryAddOptions.awaitWrites} then {@link MemoryManagerConfig.awaitWrites}.
-   * - fire-and-forget (default): resolves once writes are dispatched; per-store failures are logged.
-   * - awaited: resolves after all writes settle and throws an `AggregateError` if any store fails.
+   * This method is unscoped, with full access to all configured writable stores; tool-level store
+   * scoping is applied by the add tool callback. Target stores are validated first (an unknown or
+   * read-only named store throws), then the writes are awaited: per-store failures are logged, and
+   * an `AggregateError` is thrown if any store fails.
    *
    * @param content - The text content to add
-   * @param options - Optional metadata, store name filter, and per-call `awaitWrites` override
+   * @param options - Optional metadata and store name filter
    */
   async add(content: string, options?: MemoryAddOptions): Promise<void> {
     let writableStores: MemoryStore[]
@@ -268,33 +260,13 @@ export class MemoryManager implements Plugin {
       throw new Error('MemoryManager: no writable store matched')
     }
 
-    const write = this._writeToStores(writableStores, content, options?.metadata)
-
-    if (options?.awaitWrites ?? this._awaitWrites) {
-      await write
-    } else {
-      // Fire-and-forget: failures are already logged inside _writeToStores; swallow the rejection
-      // here so the detached promise never surfaces as an unhandled rejection.
-      write.catch(() => {})
-    }
-  }
-
-  /**
-   * Writes content to every given store, logging per-store failures. Throws an `AggregateError` if
-   * any store fails. Callers decide whether to await (observe failures) or fire-and-forget.
-   */
-  private async _writeToStores(
-    stores: MemoryStore[],
-    content: string,
-    metadata: Record<string, JSONValue> | undefined
-  ): Promise<void> {
-    const settled = await Promise.allSettled(stores.map((store) => store.add!(content, metadata)))
+    const settled = await Promise.allSettled(writableStores.map((store) => store.add!(content, options?.metadata)))
 
     const failures: { store: string; reason: unknown }[] = []
     for (let i = 0; i < settled.length; i++) {
       const settledResult = settled[i]!
       if (settledResult.status === 'rejected') {
-        const storeName = stores[i]!.name
+        const storeName = writableStores[i]!.name
         logger.warn(
           `store=<${storeName}>, reason=<${normalizeError(settledResult.reason).message}> | store write failed`
         )
@@ -383,18 +355,17 @@ export class MemoryManager implements Plugin {
     })
   }
 
-  private _createAddTool(config: MemoryToolConfig): Tool {
+  private _createAddTool(config: MemoryAddToolConfig, stores: MemoryStore[]): Tool {
     let description = config.description ?? ADD_TOOL_DESCRIPTION
-    const storeDescriptions = this._addToolStores
-      .filter((s) => s.description)
-      .map((s) => `- ${s.name}: ${s.description}`)
+    const storeDescriptions = stores.filter((s) => s.description).map((s) => `- ${s.name}: ${s.description}`)
     if (storeDescriptions.length > 0) {
       description += `\n\nAvailable writable stores:\n${storeDescriptions.join('\n')}`
       description +=
         '\n\nYou can target a specific store by name to route facts to the right place, or omit to add to all available writable stores.'
     }
 
-    const scopedNames = this._addToolStores.map((s) => s.name)
+    const scopedNames = stores.map((s) => s.name)
+    const waitForWrites = config.waitForWrites ?? true
 
     const inputSchema = z.object({
       entries: z.array(z.string()).min(1).describe('Data to add to long-term memory'),
@@ -410,6 +381,17 @@ export class MemoryManager implements Plugin {
       inputSchema,
       callback: async (input) => {
         const stores = this._resolveToolTargets(scopedNames, input.stores)
+
+        if (!waitForWrites) {
+          // Fire-and-forget: dispatch the writes without awaiting so the agent loop isn't blocked.
+          // add() logs per-store failures; swallow the rejection so it isn't an unhandled rejection.
+          for (const content of input.entries) {
+            void this.add(content, { stores }).catch(() => {})
+          }
+          return { accepted: input.entries.length } as JSONValue
+        }
+
+        // Await mode: surface failures to the model with concrete reasons (not nested AggregateErrors).
         const settled = await Promise.allSettled(input.entries.map((content) => this.add(content, { stores })))
         const failures = settled.filter(
           (settledResult) => settledResult.status === 'rejected'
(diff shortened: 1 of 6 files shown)


strands-ts/src/memory/memory-manager.ts
```diff
@@ -0,0 +1,369 @@
...
+      })
+    } else {
+      writableStores = this._config.stores.filter((s) => s.writable)
+    }
+
+    if (writableStores.length === 0) {
+      throw new Error('MemoryManager: no writable store matched')
+    }
+
+    const settled = await Promise.allSettled(writableStores.map((s) => s.add!(content, options?.metadata)))
```
notowen333: We have 2 paths here as I see it:

1) awaiting at the callsite like we see here. What we gain from this is the lines below which is observability into the outcomes of the calls. The implicit invariant we want with this approach is that every `add` implementer fast returns and then fire-and-forgets the network call/API call.

It can also be noted that the fast return could also signal a success and then silently fail later

2) We `void` not `await` at this callsite. Then we are guaranteed to not block the thread no matter what each `add` method does.

strands-ts/src/memory/memory-manager.ts
```diff
@@ -0,0 +1,369 @@
...
+      targetStores.map((store) =>
+        store.search(query, {
+          maxSearchResults: options?.maxSearchResults ?? store.maxSearchResults ?? DEFAULT_MAX_SEARCH_RESULTS,
+        })
+      )
+    )
+
+    const results: MemoryEntry[] = []
+    for (let i = 0; i < settled.length; i++) {
+      const r = settled[i]!
```
notowen333: nit: here and elsewhere I prefer to avoid the single char names

strands-ts/src/memory/types.ts
```diff
@@ -0,0 +1,136 @@
...
+ * Options passed to {@link MemoryStore.search}.
+ *
+ * Store implementations may extend this with backend-specific fields (e.g. a metadata filter or
+ * search-type override) in their own `search` signature. Note that {@link MemoryManager.search}
+ * only forwards the base fields here across its (potentially heterogeneous) stores — to use a
+ * store's extended options, call that store's `search` directly, or set them as per-instance
+ * defaults on the store.
+ */
+export interface SearchOptions {
+  /** Maximum number of results to return. */
```
notowen333: this is representing the max per store, so lets make that clear in docstring and/or var name

strands-ts/src/memory/memory-manager.ts
```diff
@@ -0,0 +1,369 @@
...
+        if (!found) {
+          throw new Error(`MemoryManager: store '${name}' not found`)
+        }
+        if (!found.writable) {
+          throw new Error(`MemoryManager: store '${name}' is read-only`)
+        }
+        return found
+      })
+    } else {
+      writableStores = this._config.stores.filter((s) => s.writable)
```
notowen333: Isn't it `this.addStores` here?

strands-ts/src/memory/memory-manager.ts
```diff
@@ -0,0 +1,369 @@
...
+ * })
+ *
+ * // Class instance (for programmatic access)
+ * const memoryManager = new MemoryManager({ stores: [myStore], addToolConfig: true })
+ * const agent = new Agent({ model, memoryManager })
+ * await memoryManager.search('user preferences')
+ * ```
+ */
+export class MemoryManager implements Plugin {
+  readonly name = 'strands:memory-manager'
```
notowen333, line 60: Re-iterating offline conversation that we intend to make the manager unique per agent

strands-ts/src/memory/memory-manager.ts
```diff
@@ -0,0 +1,369 @@
...
+          ...(entry.store && { store: entry.store }),
+          ...(entry.metadata && { metadata: entry.metadata }),
+        })) as JSONValue
+      },
+    })
+  }
+
+  private _createAddTool(config: MemoryToolConfig): Tool {
+    let description = config.description ?? ADD_TOOL_DESCRIPTION
+    const storeDescriptions = this._addStores.filter((s) => s.description).map((s) => `- ${s.name}: ${s.description}`)
```
notowen333: maybe it is nice to consolidate the description appending logic here and above maybe inline is better

strands-ts/src/memory/memory-manager.ts
opieter-aws: Yes, same definition, good catch!

strands-ts/src/memory/memory-manager.ts
opieter-aws: Correct, setting the name here is intentional and enforces uniqueness

strands-ts/src/memory/memory-manager.ts
opieter-aws: Fair, there's duplicate logic here. Looked into moving to a helper functions, but with the store set, header, and guidance sentence all differing it ends up being 3-4 input params for 8 lines, IMO reading more indirect than inline. Leaning towards keeping as-is.

strands-ts/src/memory/memory-manager.ts
opieter-aws: Indeed what we discussed.

In the implementation I went with option 1 (`await`) for observability. Manager's `store()` needs to know per-store outcomes to:
1. warn on partial failures and
2. throw when every writable store fails, which the `add_memory` tool surfaces back to the model. `void` would collapse into fire-and-forget with no failure signal at the memory manager layer

This treats “writes are async/non-blocking” as the store’s contract, not the manager’s. A store that wants true non-blocking behavior should fire-and-forget internally (accepting that a fast-return can still silently fail server-side). That keeps the manager honest about what it _can_ observe while leaving the latency tradeoff/implementation to each backend.

Can you get behind this if we document this in the store's interface?

strands-ts/src/memory/types.ts
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * Options passed to {@link MemoryStore.search}.
+ *
+ * Store implementations may extend this with backend-specific fields (e.g. a metadata filter or
+ * search-type override) in their own `search` signature. Note that {@link MemoryManager.search}
+ * only forwards the base fields here across its (potentially heterogeneous) stores — to use a
+ * store's extended options, call that store's `search` directly, or set them as per-instance
+ * defaults on the store.
+ */
+export interface SearchOptions {
```
mkmeral, line 29: nit: Maybe `MemoryStoreSearchOptions`? idk, but it feels a bit confusing to me with the tool one. this is too generic imo

strands-ts/src/memory/memory-manager.ts
zastrowm: I'm a little unclear what's being decided on here.

This scares me a bit though:

> A store that wants true non-blocking behavior should fire-and-forget internally (accepting that a fast-return can still silently fail server-side)

Leaving it up the store means it's a hard-choice that callers no longer get to choose

strands-ts/src/memory/types.ts
```diff
@@ -0,0 +1,136 @@
...
+   * pass a per-call `maxSearchResults`.
+   */
+  readonly maxSearchResults?: number
+  /**
+   * Whether this store accepts writes. Optional at config time (caller intent, defaults to `false`);
+   * concrete stores resolve it to a definite boolean on the {@link MemoryStore} interface.
+   *
+   * @defaultValue false
+   */
+  readonly writable?: boolean
```
mkmeral, line 57: how does this extend to extraction?

strands-ts/src/memory/types.ts
```diff
@@ -0,0 +1,136 @@
...
+  /** Maximum number of results per store. */
+  maxSearchResults?: number
+  /** Filter to specific stores by name. Omit to search all. */
+  stores?: string[]
+}
+
+/**
+ * Options for {@link MemoryManager.add}.
+ */
+export interface AddMemoryOptions {
```
mkmeral: should we also have like default stores? i.e. I can configure whether my agent can write to all stores, or if writes go to somewhere specific?

mkmeral (commented):
also my agent will post stuff soon, but mostly aligned with this i guess

agent-of-mkmeral:
Did an API review pass on this (read the shipped code + tests, not just the description). Really clean, well-scoped work — fail-fast constructor validation, `Promise.allSettled` resilience, provenance stamping, and the `_resolveToolTargets` scoping logic are all nice. Two small behavior items and a naming suggestion below. None are blockers.

## TL;DR
1. **`add()` writes to all writable stores by default** — consider explicit-target when >1 writable store (duplicate-write footgun).
2. **Partial write reports success** — surface per-store outcomes instead of resolving OK when some stores fail.
3. **Naming**: `SearchMemoryOptions` could `extend SearchOptions` (kills a duplicated field), and the options types could follow the module's `Memory*`-prefix convention.

(Batch-vs-single on the `add` primitive: I know a separate optional batch method is planned, so that's backwards-compatible — not raising it here.)

<details>
<summary><b>1. Write-to-all-writable-stores by default — duplicate-write footgun</b></summary>

When `options.stores` is omitted, `add()` targets every writable store:
```typescript
} else {
  writableStores = this._addStores   // ALL writable stores
}
```
With two writable stores (say `personal` + `team`), `add('prefers dark mode')` writes the **same fact to both** unless the model names one every call. Search-all-by-default is great (broad recall). Write-all-by-default duplicates facts and forces the model to specify `stores` on every write to avoid it.

Options: require explicit targeting when >1 writable store exists, or support a designated default-write store. (Single-writable-store setups are unaffected either way.)
</details>

<details>
<summary><b>2. Partial write success is reported as success</b></summary>

`add()` only throws when <i>every</i> store fails:
```typescript
if (failures.length === writableStores.length) {
  throw new AggregateError(...)
}
// otherwise: logs a warning, resolves successfully
```
So writing to 2 stores where 1 throws → `logger.warn(...)` + the call resolves OK. The caller (and the model, via `add_memory`) believes it's durably saved when it's only in one store; a later `search` scoped to the failed store silently misses it. The `add_memory` tool's `{ stored, failed }` counts <i>entries</i> (an entry is `stored` if ≥1 store accepted it), so per-store partial failure is invisible there too.

This ties into the await/observability thread on this PR — since the manager already `await`s for exactly this reason, it'd be good to propagate per-store outcomes (e.g. a `{ store, ok, reason }[]` summary) rather than collapse partial into success. At minimum, documenting the current semantics would help.
</details>

<details>
<summary><b>3. Naming — options types vs the module convention</b></summary>

Everything in the module is `Memory*`-prefixed, noun-first (`MemoryEntry`, `MemoryStore`, `MemoryStoreConfig`, `MemoryToolConfig`, `MemoryManagerConfig`). The three options types diverge:

| Type | Prefix | Word order |
|---|---|---|
| `MemoryStore`, `MemoryEntry`, … | `Memory*` | noun-first |
| `SearchOptions` | none | verb-first |
| `SearchMemoryOptions` | `Memory` infix | verb-first |
| `AddMemoryOptions` | `Memory` infix | verb-first |

**The main one:** `SearchMemoryOptions` is a strict superset of `SearchOptions` (same `maxSearchResults` + a routing `stores?`) but redeclares the shared field independently:
```typescript
export interface SearchOptions {       maxSearchResults?: number }
export interface SearchMemoryOptions { maxSearchResults?: number; stores?: string[] }  // dup
```
This is the same drift `MemoryStoreConfig` was created to avoid ("declared once, extended by both"). Making it explicit removes the duplicate and expresses the is-a relationship:
```typescript
export interface SearchOptions {                              // store primitive
  maxSearchResults?: number
}
export interface MemorySearchOptions extends SearchOptions {  // manager adds routing
  stores?: string[]
}
```
Suggested scheme: keep `SearchOptions` as the un-prefixed <i>store</i> primitive (input to `MemoryStore.search`); rename the manager-level ones to `MemorySearchOptions extends SearchOptions` and `AddMemoryOptions` → `MemoryAddOptions`. Then `Memory*` consistently means "manager layer" and un-prefixed means "store primitive."

(There's no store-level `AddOptions` twin since the write signature is `add?(content, metadata?)` with no options object — so `MemoryAddOptions` has nothing to extend, which is fine; might be worth a one-line note so the asymmetry doesn't read as an oversight.)

Tiny extra: `MemoryEntry.store?: string` holds a store <i>name</i>, not a `MemoryStore` — `storeName` would be unambiguous since it's on the model-visible result shape.
</details>

Happy to open inline suggestions or a small follow-up if any of these are useful — your call on what's in scope for this PR vs. the injection follow-up.

strands-ts/src/memory/memory-manager.ts
opieter-aws: It's about which entity bears the responsibility that a backend `write`  doesn't block the agent loop unnecessarily.

The proposal in this PR is that this is up to the backend, not up to the orchestrating manager, because otherwise the manager would never know if a store's `write` is failing.

> Leaving it up the store means it's a hard-choice that callers no longer get to choose

In which scenario would we want to block the agent loop to await a backend write operation? Are you suggesting the manager could expose an option to `await` or `fire-and-forget` instead?

strands-ts/src/memory/types.ts
opieter-aws: Good suggestion and more in line with rest of naming, will address

strands-ts/src/memory/types.ts
opieter-aws: Extraction is set up per store, and is a write operation. So I'd argue a store can't be `writable: false` and specify an extraction config (should throw). WDYT?

strands-ts/src/memory/types.ts
opieter-aws: Configuring to which stores an agent can write is done with `writable` property on the stores. Are you talking about the case of your agent, of having multiple `writable` stores?

For tool use, the agent gets to decide to which `writable` stores it writes

strands-ts/src/memory/memory-manager.ts
opieter-aws: Discussed offline with @zastrowm , we decided on the following:

* Fire-and-forget will be responsibility of the MemoryManager, not the store
* The MemoryManager will exposes the option to run `add` as `await` (`false` by default), so users can choose to await writes
* For future extension, we need to think about how we can surface error accumulation. Especially when we introduce extraction as a background process.

strands-ts/src/memory/types.ts
mkmeral: i'm not sure, and that's why I also highlighted this. I guess it would depend on extraction devx

If I am explicitly configuring extraction for a store, I'd expect it to work. I shouldn't need to modify multiple stuff. That said, maybe we can rename this to be more clear?

strands-ts/src/memory/types.ts
mkmeral: yeah I think that covers enough. If I have single writable store, then it kinda decides. The main issue was more what my agent said, about writing to all stores by default. 

Like working backwards, if I have multiple KBs, they generally are targeted for different use cases. So I'm not 100% if we should make writing to all default. I'd expect agent to choose one, so we can make at stores and one item in store requirement maybe? wdyt? 

(also given that this is a tool param, we can also solve these through tool description)

opieter-aws: Stating that we write to all stores by default is misleading; the agent is instructed in the tool description that "omitting the `stores` paramater writes to all", but that's still an agent instruction. This is already covered through tool description.

strands-ts/src/memory/types.ts
opieter-aws: Per our offline discussion:
* `writable` stays on the store. This configures _any_ writability, so both tool and periodic writes (extraction)
* To enable writability through tool, users can configure this on the `MemoryAddToolConfig`. An opt-in for the tool (setting to `true`) will default to enabling write to all writable stores.

strands-ts/src/memory/memory-manager.ts
```diff
@@ -0,0 +1,430 @@
...
+  }
+
+  /**
+   * Initializes the plugin with the agent.
+   *
+   * No lifecycle hooks are registered in this version.
+   *
+   * @param _agent - The agent this plugin is being attached to
+   */
+  initAgent(_agent: LocalAgent): void {}
```
mkmeral, line 140: qq, does the plugin force you to write this?

strands-ts/src/memory/types.ts
```diff
@@ -0,0 +1,166 @@
...
+  stores?: string[]
+}
+
+/**
+ * Options for {@link MemoryManager.add}.
+ */
+export interface MemoryAddOptions {
+  /** Metadata to associate with the added entry. */
+  metadata?: Record<string, JSONValue>
+  /** Filter to specific writable stores by name. Omit to write to all. */
```
mkmeral: nit: `Omit to write to all writable stores`

strands-ts/src/memory/types.ts
```diff
@@ -0,0 +1,166 @@
...
+   * pass a per-call `maxSearchResults`.
+   */
+  readonly maxSearchResults?: number
+  /**
+   * Whether this store accepts writes. Optional at config time (caller intent, defaults to `false`);
+   * concrete stores resolve it to a definite boolean on the {@link MemoryStore} interface.
+   *
+   * @defaultValue false
+   */
+  readonly writable?: boolean
```
notowen333, line 57: isWritable? for boolean conventions?

strands-ts/src/memory/types.ts
```diff
@@ -0,0 +1,166 @@
...
+  /**
+   * Whether write operations await each store before resolving. Defaults to `false`.
+   * - `false` (default): fire-and-forget. Writes are dispatched and the call resolves immediately so
+   *   a slow backend never blocks the agent loop; per-store failures are logged, not thrown.
+   * - `true`: awaits all writes. Throws an `AggregateError` if any targeted store fails, so partial
+   *   failures are observable.
+   *
+   * Per-call overridable via {@link MemoryAddOptions.awaitWrites}.
+   */
+  awaitWrites?: boolean
```
mkmeral: is this a tool config or a memory manager config? does/will this apply to extraction?

mkmeral: i'd expect this to apply to only add tool, as extraction should be async by default imo

strands-ts/src/memory/memory-manager.ts
```diff
@@ -0,0 +1,430 @@
...
+    }
+
+    const write = this._writeToStores(writableStores, content, options?.metadata)
+
+    if (options?.awaitWrites ?? this._awaitWrites) {
+      await write
+    } else {
+      // Fire-and-forget: failures are already logged inside _writeToStores; swallow the rejection
+      // here so the detached promise never surfaces as an unhandled rejection.
+      write.catch(() => {})
```
notowen333: would rename 271 to writePromise and explicitly say `void` here

strands-ts/src/memory/memory-manager.ts
opieter-aws: Yes indeed. Will be populated on next PR with injection

strands-ts/src/memory/types.ts
opieter-aws: Discussed, moving to tool config

strands-ts/src/memory/types.ts
opieter-aws: The existing SDK pattern is: `is*` for describing observed state on a data object; bare adjective for a configuration flag the user sets. writable is a config flag the caller sets on a store → bare adjective is correct, and isWritable would actually be against the dominant convention.

mkmeral (approved):

PR #2547, opened 2026-06-01
feat(context): add context manager

lizradway:
## Description

Implements the v1 `contextManager` facade as designed in strands-agents/docs#831.

Adds a `contextManager` parameter to `AgentConfig` that pre-composes the SDK's context management primitives into a single configuration surface. An internal `ContextManager` plugin composes sub-plugins (`ContextCompression`, `ContextOffloader`) that handle the actual behavior.

### Architecture

```
ContextManager (internal plugin)
├── ContextCompression (sub-plugin)
│   ├── Proactive compression (BeforeModelCallEvent)
│   ├── Reactive overflow recovery (AfterModelCallEvent)
│   └── Sliding window enforcement (AfterInvocationEvent)
└── ContextOffloader (sub-plugin)
    ├── Tool result caching (AfterToolCallEvent)
    └── retrieve_offloaded_content tool
```

Sub-plugins work independently when used standalone. User-provided plugins with matching names take precedence over managed sub-plugins. When `contextManager` is set, `ContextCompression` takes priority — `NullConversationManager` is used (same pattern as other dedicated-param plugins like `retryStrategy`, `sessionManager`).

### What ships

- **`contextManager` parameter on `AgentConfig`** — accepts `"auto"` or a config object
- **`ContextCompression` plugin** — proactive/reactive compression with own reduction logic (truncate or summarize)
- **`ContextOffloader`** — stays in `vended-plugins/context-offloader/`, composed internally by ContextManager
- **Message pinning** (`context-manager/compression/protection.ts`) — `pinMessageTool` for agent-controlled pinning at runtime. Internal utilities (`pinMessage`, `unpinMessage`, `isPinned`, `isProtected`) not exported; programmatic pinning API deferred.
- **`protectFirst`** — number of messages at the start of the conversation to protect from eviction
- **`estimateInputTokens()` utility** — shared token estimation in `src/context-manager/token-estimation.ts`
- **`<summary>` XML tags** — summarized messages are wrapped in `<summary>` tags so the model can distinguish framework-injected summaries from user content
- **`conversationManager` marked as pending deprecation** — still works, JSDoc-tagged

### Public API Surface

**New on `AgentConfig`:**
- `contextManager?: ContextManagerParam`

**New exports:**
- `pinMessageTool` (agent-invokable tool)
- `ContextManagerParam` (type)

All classes (`ContextManager`, `ContextCompression`, `ContextOffloader`) are internal. `ContextOffloader` remains accessible via the `vended-plugins/context-offloader` sub-path for backward compat.

### Configuration model

Two semantics depending on whether `strategy: 'auto'` is present:

**Override (strategy: 'auto')** — starts with everything enabled, you override specific settings:
```typescript
contextManager: "auto"                                              // everything with defaults
contextManager: { strategy: 'auto', compression: { windowSize: 60 } }  // auto, tweak compression
contextManager: { strategy: 'auto', offloader: { threshold: 5000 } }   // auto, tweak offloader
```

**Additive (no strategy)** — starts with nothing, you enable what you want:
```typescript
contextManager: { compression: true }                               // only compression
contextManager: { compression: 'summarize' }                        // only summarize compression
contextManager: { offloader: true }                                 // only offloading
contextManager: { compression: true, offloader: true }              // both (same as "auto")
contextManager: {}                                                  // nothing enabled
```

### Compression config (discriminated union on `method`)
```typescript
compression: true                                       // defaults
compression: 'truncate'                                 // method shorthand
compression: 'summarize'                                // method shorthand
compression: { method: 'truncate', windowSize: 30 }     // full config
compression: { method: 'summarize', summaryRatio: 0.5 } // full config
compression: { protectFirst: 2 }                        // protect first 2 messages
```

### Defaults (when enabled via "auto")

| Parameter | Default |
|-----------|---------|
| offloader threshold | 2500 tokens |
| offloader previewTokens | 1500 tokens |
| compression method | "truncate" |
| compression windowSize | 40 |
| compression proactive | true (threshold 0.7) |
| storage | InMemoryStorage |

### Plugin registration

`contextManager` must be passed via the dedicated parameter — same pattern as `conversationManager`, `retryStrategy`, and `sessionManager`. No guards for misuse in `plugins[]` (consistent with other special-cased plugins).

### Deprecation Plan

The following are marked as pending deprecation in v1 and will be removed in v2:
- `AgentConfig.conversationManager` → `contextManager: { compression: ... }`
- `Agent._estimateInputTokens()` → shared `estimateInputTokens()` utility
- `BeforeModelCallEvent.projectedInputTokens` → future `contextManager` budget API
- `ConversationManager`, `SlidingWindowConversationManager`, `SummarizingConversationManager`, `NullConversationManager` → `ContextCompression` plugin
- `vended-plugins/context-offloader/` sub-path → import from `@strands-agents/sdk` directly

### Breaking Changes

**None.** All changes are additive. Existing behavior is unchanged when `contextManager` is not set.

## Documentation PR

strands-agents/docs#831

## Type of Change

New feature

## Testing

- [x] Type check passes
- [x] Lint passes
- [x] All 2915 tests pass (106 test files)
- [x] 60 new unit tests covering: token estimation, compression plugin, truncate/summarize strategies, protection logic, context manager resolution

## Checklist

----

### DIFF
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -44,6 +44,8 @@ import { PluginRegistry } from '../plugins/registry.js'
 import { SlidingWindowConversationManager } from '../conversation-manager/sliding-window-conversation-manager.js'
 import { NullConversationManager } from '../conversation-manager/null-conversation-manager.js'
 import { ConversationManager } from '../conversation-manager/conversation-manager.js'
+import type { ContextManagerParam } from '../context-manager/context-manager.js'
+import { resolveContextManager } from '../context-manager/context-manager.js'
 import { HookRegistryImplementation } from '../hooks/registry.js'
 import type { HookableEventConstructor, HookCallback, HookCallbackOptions, HookCleanup } from '../hooks/types.js'
 import {
@@ -168,8 +170,23 @@ export type AgentConfig = {
    */
   printer?: boolean
   /**
+   * Pre-composed context management strategy.
+   *
+   * - `"auto"`: enables tool result caching and proactive compression with defaults.
+   * - Object: fine-grained control over strategy, storage, caching, and compression settings.
+   * - `undefined` (default): no context management facade; use `conversationManager`
+   *   and `plugins` directly.
+   *
+   * When set, takes priority over `conversationManager` — `NullConversationManager` is used.
+   */
+  contextManager?: ContextManagerParam
+  /**
    * Conversation manager for handling message history and context overflow.
    * Defaults to SlidingWindowConversationManager with windowSize of 40.
+   *
+   * @remarks Pending deprecation — use `contextManager` instead. The `contextManager` parameter
+   * composes compression, tool result caching, and token estimation into a single
+   * configuration surface. This field will be deprecated in a future version.
    */
   conversationManager?: ConversationManager
   /**
@@ -331,15 +348,29 @@ export class Agent implements LocalAgent, InvokableAgent {
       this.model = config?.model ?? new BedrockModel()
     }
 
-    // Validate and assign conversation manager
+    let contextManagerPlugin: Plugin | undefined
+    if (config?.contextManager) {
+      contextManagerPlugin = resolveContextManager(config.contextManager, config.plugins)
+    }
+
+    // Validate and assign conversation manager.
+    // When contextManager is set, ContextCompression owns compression — use NullConversationManager.
     if (this.model.stateful) {
-      if (config?.conversationManager) {
+      if (config?.conversationManager || config?.contextManager) {
         throw new Error(
-          'Cannot use a conversationManager with a stateful model. The model manages conversation state server-side.'
+          'Cannot use a conversationManager or contextManager with a stateful model. The model manages conversation state server-side.'
         )
       }
       this._conversationManager = new NullConversationManager()
+    } else if (contextManagerPlugin) {
+      if (config?.conversationManager) {
+        logger.warn('contextManager takes priority over conversationManager — conversationManager will be ignored')
+      }
+      this._conversationManager = new NullConversationManager()
     } else {
+      if (config?.conversationManager) {
+        logger.warn('conversationManager is deprecated and will be removed in v2. Use contextManager instead.')
+      }
       this._conversationManager =
         config?.conversationManager ?? new SlidingWindowConversationManager({ windowSize: 40 })
     }
@@ -372,9 +403,12 @@ export class Agent implements LocalAgent, InvokableAgent {
     // - Retry-strategy ordering is not load-bearing for correctness: `DefaultModelRetryStrategy`
     //   guards on `event.retry`, so a user hook that already set it short-circuits
     //   the strategy regardless of registration order.
+    // - contextManager plugin goes before user plugins so the offloader's AfterToolCallEvent
+    //   hook fires first, ensuring large results are cached before user hooks see the event.
     this._pluginRegistry = new PluginRegistry([
       this._conversationManager,
       ...retryStrategies,
+      ...(contextManagerPlugin ? [contextManagerPlugin] : []),
       ...(config?.plugins ?? []),
       ...(config?.sessionManager ? [config.sessionManager] : []),
       new ModelPlugin(this.model),
@@ -1397,7 +1431,8 @@ export class Agent implements LocalAgent, InvokableAgent {
 
     let attemptCount = 1
     while (true) {
-      // Estimate input tokens for the upcoming model call (non-fatal if estimation fails)
+      // Pending deprecation: token estimation will move fully to ContextManager.
+      // This remains for backward compat with standalone ConversationManager.proactiveCompression.
       let projectedInputTokens: number | undefined
       try {
         projectedInputTokens = await this._estimateInputTokens(streamOptions)
--- a/strands-ts/src/context-manager/compression/context-compression.ts
+++ b/strands-ts/src/context-manager/compression/context-compression.ts
@@ -1,0 +1,178 @@
+import type { Plugin } from '../../plugins/plugin.js'
+import type { LocalAgent } from '../../types/agent.js'
+import type { Tool } from '../../tools/tool.js'
+import type { Message } from '../../types/messages.js'
+import type { Model } from '../../models/model.js'
+import { AfterInvocationEvent, AfterModelCallEvent, BeforeModelCallEvent } from '../../hooks/events.js'
+import { ContextWindowOverflowError } from '../../errors.js'
+import { truncate, type TruncateOptions } from './strategies/truncate.js'
+import { summarize, type SummarizeOptions } from './strategies/summarize.js'
+import { estimateInputTokens } from '../token-estimation.js'
+import { logger } from '../../logging/logger.js'
+import { warnOnce } from '../../logging/warn-once.js'
+
+const DEFAULT_CONTEXT_WINDOW_LIMIT = 200_000
+const DEFAULT_PROACTIVE_THRESHOLD = 0.7
+const DEFAULT_WINDOW_SIZE = 40
+
+export type CompressionMethod = 'truncate' | 'summarize'
+
+type SharedCompressionOptions = {
+  /**
+   * Proactive compression before the model call.
+   * - `true`: compress when 70% of the context window is used (default threshold).
+   * - `{ threshold: number }`: compress at the specified ratio (0, 1].
+   * - `false`: disable proactive compression; only reactive overflow recovery is used.
+   * - Omitted: defaults to `true`.
+   */
+  proactive?: boolean | { threshold: number }
+  /**
+   * Number of messages at the start of the conversation to protect from eviction.
+   * These messages are never trimmed or summarized away.
+   */
+  protectFirst?: number
+}
+
+export type TruncateCompressionConfig = SharedCompressionOptions & {
+  method?: 'truncate'
+  /** Maximum messages to keep after trimming. Defaults to 40. */
+  windowSize?: number
+}
+
+export type SummarizeCompressionConfig = SharedCompressionOptions & {
+  method: 'summarize'
+  /** Ratio of messages to summarize (0.1–0.8). Defaults to 0.3. */
+  summaryRatio?: number
+  /** Minimum recent messages to preserve during summarization. Defaults to 10. */
+  preserveRecentMessages?: number
+}
+
+/**
+ * Compression configuration (discriminated union on `method`).
+ *
+ * @example
+ * ```typescript
+ * contextManager: { compression: true }                                         // defaults (truncate)
+ * contextManager: { compression: 'summarize' }                                  // strategy shorthand
+ * contextManager: { compression: { method: 'truncate', windowSize: 30 } }     // full config
+ * contextManager: { compression: { method: 'summarize', summaryRatio: 0.5 } } // full config
+ * ```
+ */
+export type CompressionOptions = TruncateCompressionConfig | SummarizeCompressionConfig
+
+/**
+ * Plugin that handles context compression — both proactive (before model call when
+ * threshold is exceeded) and reactive (after model call on overflow error).
+ *
+ * Delegates reduction to strategy functions (truncate or summarize).
+ */
+export class ContextCompression implements Plugin {
+  readonly name = 'strands:context-compression'
+
+  private readonly _proactiveThreshold: number | undefined
+  private readonly _method: CompressionMethod
+  private readonly _windowSize: number
+  private readonly _truncateOptions: TruncateOptions
+  private readonly _summarizeOptions: SummarizeOptions
+
+  constructor(config?: CompressionOptions) {
+    const proactive = config?.proactive ?? true
+    if (proactive === false) {
+      this._proactiveThreshold = undefined
+    } else if (proactive === true) {
+      this._proactiveThreshold = DEFAULT_PROACTIVE_THRESHOLD
+    } else {
+      if (proactive.threshold <= 0 || proactive.threshold > 1) {
+        throw new Error(
+          `proactive compression threshold must be between 0 (exclusive) and 1 (inclusive), got ${proactive.threshold}`
+        )
+      }
+      this._proactiveThreshold = proactive.threshold
+    }
+
+    this._method = config?.method ?? 'truncate'
+    this._windowSize = (config as TruncateCompressionConfig | undefined)?.windowSize ?? DEFAULT_WINDOW_SIZE
+
+    this._truncateOptions = {
+      ...(config?.protectFirst !== undefined && { protectFirst: config.protectFirst }),
+    }
+    this._summarizeOptions = {
+      ...(config?.protectFirst !== undefined && { protectFirst: config.protectFirst }),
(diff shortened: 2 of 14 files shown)


strands-ts/src/agent/agent.ts
```diff
@@ -167,9 +169,24 @@ export type AgentConfig = {
...
+   * Pre-composed context management strategy.
+   *
+   * - `"auto"`: enables tool result caching and proactive compression with defaults.
+   * - Object: fine-grained control over strategy, storage, caching, and compression settings.
+   * - `undefined` (default): no context management facade; use `conversationManager`
+   *   and `plugins` directly.
+   *
+   * When set, takes priority over `conversationManager` — `NullConversationManager` is used.
+   */
+  contextManager?: ContextManagerParam
```
notowen333, line 182: Can we rename this to maybe `ContextManagerInput` or inline the union. `Param` jumped out as confusing to me

strands-ts/src/context-manager/context-manager.ts
```diff
@@ -0,0 +1,182 @@
...
+/**
+ * Compression configuration accepted by contextManager.
+ * - `true`: enable with defaults (truncate, proactive at 0.7).
+ * - `'truncate'` / `'summarize'`: enable specific strategy with defaults.
+ * - Object: full config with strategy and options.
+ * - Omitted: disabled.
+ */
+export type CompressionConfig =
+  | true
+  | import('./compression/context-compression.js').CompressionMethod
```
notowen333, line 30: why are we not just importing top level? What does this buy us?

strands-ts/src/context-manager/context-manager.ts
```diff
@@ -0,0 +1,182 @@
...
+ * Configuration accepted by the {@link ContextManager} constructor.
+ *
+ * Config objects are additive — only features you explicitly set are enabled.
+ * Use `"auto"` to enable everything with defaults.
+ */
+export type ContextManagerConfig = {
+  /** Strategy name. Only "auto" is supported currently. */
+  strategy?: ContextStrategyValue
+  /** Storage backend for cached tool results. Defaults to InMemoryStorage. */
+  storage?: Storage
```
notowen333, line 43: If it's just cached tool results should we narrow the parameter name?

strands-ts/src/context-manager/context-manager.ts
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  private _buildSubPlugins(userPlugins?: Plugin[]): Plugin[] {
+    const config = this._config
+    const plugins: Plugin[] = []
+
+    if (config.compression) {
+      const userProvided = userPlugins?.some((p) => p.name === 'strands:context-compression')
+      if (!userProvided) {
+        let compressionConfig: import('./compression/context-compression.js').CompressionOptions | undefined
```
notowen333, line 131: same Q

strands-ts/src/context-manager/context-manager.ts
```diff
@@ -0,0 +1,182 @@
...
+      plugin.initAgent(agent)
+    }
+  }
+
+  private _buildSubPlugins(userPlugins?: Plugin[]): Plugin[] {
+    const config = this._config
+    const plugins: Plugin[] = []
+
+    if (config.compression) {
+      const userProvided = userPlugins?.some((p) => p.name === 'strands:context-compression')
```
notowen333, line 129: Can we do instanceof instead of this name matching pattern?

strands-ts/src/context-manager/compression/protection.ts
```diff
@@ -0,0 +1,134 @@
...
+})
+
+// --- Protection check ---
+
+/**
+ * Check if a message at the given index is protected from eviction.
+ * A message is protected if it is pinned, within the protectFirst range,
+ * or is a tool-pair partner of a protected message.
+ */
+export function isProtected(messages: Message[], index: number, protectFirst?: number): boolean {
```
notowen333, line 120: Maybe a more descriptive function name

strands-ts/src/context-manager/compression/strategies/summarize.ts
```diff
@@ -0,0 +1,149 @@
+import { Message, TextBlock } from '../../../types/messages.js'
+import type { Model } from '../../../models/model.js'
+import { isProtected } from '../protection.js'
+import { logger } from '../../../logging/logger.js'
+
+const SUMMARIZATION_PROMPT = `You are a conversation summarizer. Provide a concise summary of the conversation history.
+
+Format Requirements:
+- You MUST create a structured and concise summary in bullet-point format.
```
notowen333, line 9: confirming bullet-point format

strands-ts/src/context-manager/compression/strategies/summarize.ts
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * Summarize the oldest messages and replace them with a model-generated summary.
+ *
+ * @param messages - The messages array to mutate in place
+ * @param model - The model to use for generating the summary
+ * @param options - Summarization options
+ * @returns `true` if messages were summarized, `false` if not enough to summarize
+ */
+export async function summarize(messages: Message[], model: Model, options?: SummarizeOptions): Promise<boolean> {
```
notowen333, line 51: could expose sys prompt for the summarization

strands-ts/src/agent/agent.ts
```diff
@@ -167,9 +169,24 @@ export type AgentConfig = {
...
   /**
    * Conversation manager for handling message history and context overflow.
    * Defaults to SlidingWindowConversationManager with windowSize of 40.
+   *
+   * @remarks Pending deprecation — use `contextManager` instead. The `contextManager` parameter
```
opieter-aws, line 187: Are we actually planning to deprecate this? Do we need this warning if we're only deprecating in 2.0?

strands-ts/src/context-manager/compression/strategies/truncate.ts
```diff
@@ -0,0 +1,85 @@
+import type { Message } from '../../../types/messages.js'
+import { isProtected } from '../protection.js'
+import { logger } from '../../../logging/logger.js'
+
+export type TruncateOptions = {
+  /** Positive: protect first N messages. Negative: protect last N messages. */
+  protectFirst?: number
```
opieter-aws, line 7: What if I want both first and last?

strands-ts/src/context-manager/compression/context-compression.ts
```diff
@@ -0,0 +1,178 @@
...
+  /** Maximum messages to keep after trimming. Defaults to 40. */
+  windowSize?: number
+}
+
+export type SummarizeCompressionConfig = SharedCompressionOptions & {
+  method: 'summarize'
+  /** Ratio of messages to summarize (0.1–0.8). Defaults to 0.3. */
+  summaryRatio?: number
+  /** Minimum recent messages to preserve during summarization. Defaults to 10. */
+  preserveRecentMessages?: number
```
opieter-aws, line 47: `protectLast` would have more synergy with `protectFirst`. That said, I'd be in favor of exposing an object instead, but let's at least align the two for consistency.

strands-ts/src/context-manager/compression/context-compression.ts
```diff
@@ -0,0 +1,178 @@
...
+
+export type TruncateCompressionConfig = SharedCompressionOptions & {
+  method?: 'truncate'
+  /** Maximum messages to keep after trimming. Defaults to 40. */
+  windowSize?: number
+}
+
+export type SummarizeCompressionConfig = SharedCompressionOptions & {
+  method: 'summarize'
+  /** Ratio of messages to summarize (0.1–0.8). Defaults to 0.3. */
```
opieter-aws, line 44: This is an unclear description to me. What does ratio of messages mean?

strands-ts/src/context-manager/context-manager.ts
```diff
@@ -0,0 +1,182 @@
...
+   * - `true`: enable with defaults (threshold=2500, previewTokens=500).
+   * - Object: enable with custom settings.
+   * - Omitted: disabled.
+   */
+  offloader?: true | OffloaderConfig
+  /**
+   * Compression configuration.
+   * - `true`: enable with defaults (truncate, proactive at 0.7).
+   * - `'truncate'` / `'summarize'`: enable specific strategy with defaults.
+   * - `CompressionStrategy.Truncate(...)` / `CompressionStrategy.Summarize(...)`: full config.
```
opieter-aws, line 55: Do we actually expose these types?

strands-ts/src/context-manager/context-manager.ts
```diff
@@ -0,0 +1,182 @@
...
+ * Use `"auto"` to enable everything with defaults.
+ */
+export type ContextManagerConfig = {
+  /** Strategy name. Only "auto" is supported currently. */
+  strategy?: ContextStrategyValue
+  /** Storage backend for cached tool results. Defaults to InMemoryStorage. */
+  storage?: Storage
+  /**
+   * Context offloader configuration.
+   * - `true`: enable with defaults (threshold=2500, previewTokens=500).
```
opieter-aws, line 46: `previewTokens` has a different default here than actual (1500)

opieter-aws: Also, are we purposefully drifting from the existing offloader's default of 1000?

strands-ts/src/context-manager/token-estimation.ts
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Uses an incremental strategy: if the last assistant message has usage metadata,
+ * uses (inputTokens + outputTokens) as a baseline and only counts new messages
+ * added after it. Falls back to full model estimation otherwise.
+ *
+ * @param messages - The conversation messages
+ * @param model - The model to use for token counting
+ * @returns Estimated token count, or undefined if estimation fails
+ */
+export async function estimateInputTokens(messages: Message[], model: Model): Promise<number | undefined> {
```
opieter-aws, line 16: In the current implementation we add the option to estimate `systemPrompt` and `toolSpecs`. Can we support this here to avoid regression? Then we can also forward `Agent._estimateInputTokens()` to use this function and avoid maintaining 2 copies.

strands-ts/src/context-manager/compression/protection.ts
```diff
@@ -0,0 +1,134 @@
...
+
+// --- Pin utilities ---
+
+/**
+ * Check if a single message is pinned.
+ *
+ * @param message - The message to check
+ * @returns `true` if the message has `metadata.custom.pinned === true`
+ */
+export function isPinned(message: Message): boolean
```
notowen333, line 13: After offline discussion, it sounds like "pin" verbage is coming from tool result pairs. I.e. we don't say protect/ed directly because it can mean two messages.

I think this mismatch is confusing. I wouldn't mind just letting `protected` automatically include a pair.

strands-ts/src/context-manager/context-manager.ts
notowen333: Or if it is also other things... `Storage` alone is too concise

strands-ts/src/agent/agent.ts
lizradway: would probably want to talk to @zastrowm about this:
fwiw, the established pattern i am aware of is to issue deprecation warning in new version bump and remove code path in the following version bump.

if we want to explicitly remove conversation manager in v2, from my understanding, we want to issue a deprecation warning now.

zastrowm: > if we want to explicitly remove conversation manager in v2, from my understanding, we want to issue a deprecation warning now.

Concur; deprecate as long as an equivalent is in place.  "Deprecate" = "use this other thing as we want to remove this in the future"

We can start adding a warning right now too

strands-ts/src/context-manager/compression/strategies/truncate.ts
lizradway: realistically, `protectLast` is default in all compression behavior (ie. sliding window) and `protectFirst` is _far_ more useful (to protect prompt/first user messages).

if we see demand/usecase for it i think it would be nice to add, but i think for now I'd want to keep the api surface minimal. feel free to push back if you'd want to ship immedietely.

strands-ts/src/context-manager/compression/context-compression.ts
lizradway: i think that is fair, right now for sliding window i just ported over the exact impl, but can update this.

strands-ts/src/context-manager/compression/context-compression.ts
lizradway: updated the jsdoc -- essentially this is the oldest x% of messages we are summarizing and discarding

strands-ts/src/context-manager/context-manager.ts
lizradway: defaults are incorrect for now until we run proper benchmarking

strands-ts/src/context-manager/token-estimation.ts
lizradway: yes good idea, can add!

strands-ts/src/agent/agent.ts
lizradway: actually might just inline this. this feels unnecessary

strands-ts/src/context-manager/compression/protection.ts
lizradway: thoughts on `isMessageEvictionProtected`?  `isEvictionProtected`?

strands-ts/src/context-manager/context-manager.ts
lizradway: users can't `instanceof` this since this config isn't exported

strands-ts/src/context-manager/context-manager.ts
lizradway: this is what is currently used in the context offloader plugin.

can update comment, as this will also be used for Context Delegation/Context Reads.

strands-ts/src/context-manager/context-manager.ts
lizradway: can move to top level import, this was _supposed_ to avoid relying on to-be-deprecated logic, but we're importing from it anyways

strands-ts/src/context-manager/context-manager.ts
lizradway: good catch -- these were removed

strands-ts/src/context-manager/compression/strategies/summarize.ts
lizradway: ported over from summarizing conversation manager -- is that fine?

strands-ts/src/context-manager/compression/strategies/summarize.ts
lizradway: i like that idea alot

strands-ts/src/context-manager/compression/protection.ts
lizradway: the opposite, but yes:
- pinned is explicit pinning
- 'protected' is the result of a tool result pair preventing orphaning.

`isEvictionProtected` perhaps can alleviate these sounding very similar semantically and having similar meanings?

lizradway:
Closing out in favor of context facade

PR #2551, opened 2026-06-02
fix(openai-responses): prevent tool list mutation across API calls

### LINKED ISSUE #2405: [BUG] Strands python OpenAI responses calls duplicate tool definitions on each turn
### Checks

### SDK Language

Python

### Strands Version

1.41.0

### Language Runtime Version

Python 3.12.9

### Operating System

macOS 15.7.2

### Installation Method

pip

### Steps to Reproduce

**Problem Description**
While running Strands agents (OpenAI using the Responses API) I noticed unexpected context growth around tool calls. My tool definition size overall is around 10K tokens (I have a lot of tools) and each time Strands would internally call a tool and then make the next call to OpenAI, the context would grow by around 10K more than I was expecting, it was super weird, and I was hitting context limits. 

**Investigation**
OpenAIResponsesModel._format_request shallow-copies params via ** spread,
so request["tools"] and self.config["params"]["tools"] are the SAME list.
.extend() permanently mutates the config, adding a full copy of every tool
spec on every API call.

**Testcase**
Run this file (you need to put an OPENAI_API_KEY into the env first)
[strands_tool_duplication_bug.py](https://github.com/user-attachments/files/28440956/strands_tool_duplication_bug.py)
The idea is that it has an "hello world" tool, and we instruct the agent to call that tool 3 times. We override the _format_request method so we have a place to print the tools array before it gets sent to the next OpenAI call by Strands. 

**Testcase Output on Strands 1.41.0***
$ python3 strands_tool_duplication_bug.py
```Initial config params tools: 1 entries

Running agent...

  API call #1: 2 tools (expect 2): ['web_search', 'hello_world']

Tool #1: hello_world
  API call #2: 3 tools (expect 2): ['web_search', 'hello_world', 'hello_world']

Tool #2: hello_world
  API call #3: 4 tools (expect 2): ['web_search', 'hello_world', 'hello_world', 'hello_world']

Tool #3: hello_world
  API call #4: 5 tools (expect 2): ['web_search', 'hello_world', 'hello_world', 'hello_world', 'hello_world']
Done.
Agent output: Done.

Final config['params']['tools']: 5 entries
Expected: 2 (1 web_search + 1 hello_world)

BUG CONFIRMED: _format_request mutates self.config['params']['tools']
via shared list reference, duplicating all tool specs on every call.
```

### Expected Behavior

I expected that each followup call to OpenAI that was orchestrated by Strands would pass the same (original) list of tools.

### Actual Behavior

Each followup call to OpenAI that was orchestrated by Strands is passing an increasingly long list of tools, one extra duplicate of each tool is added each round trip, so it can grow quite large.

### Additional Context

I have tried to search for duplicates and be responsible about filing, but this is my first time, so if I missed something, I apologize and go easy on me!

### Possible Solution

_No response_

### Related Issues

_No response_

xlyoung:
## Summary

Fixes #2405

`OpenAIResponsesModel._format_request` shallow-copies `self.config["params"]` via `**` unpacking into the `request` dict. When `params` contains a `"tools"` key, `request["tools"]` and `self.config["params"]["tools"]` become the **same list object**.

The subsequent `request.setdefault("tools", []).extend(...)` then permanently mutates `self.config["params"]["tools"]`, adding a duplicate copy of every tool spec on each API call. With 10 tools, after 5 calls the tools list grows from 10 to 15 entries.

## Fix

Shallow-copy the existing tools list before extending when it already exists in the request from `params` unpacking:

```python
if "tools" in request:
    request["tools"] = list(request["tools"])
request.setdefault("tools", []).extend(...)
```

This ensures `.extend()` operates on a fresh list, leaving `self.config["params"]["tools"]` untouched.

## Testing

- Verified with the reproduction script from the issue: tool list no longer grows across API calls
- Existing `_format_request` tests continue to pass

### DIFF
--- a/strands-py/src/strands/models/openai_responses.py
+++ b/strands-py/src/strands/models/openai_responses.py
@@ -549,7 +549,12 @@ class OpenAIResponsesModel(Model):
 
         # Add tools if provided
         if tool_specs:
-            # Merge with any built-in tools (e.g. web_search) already in the request from params
+            # Merge with any built-in tools (e.g. web_search) already in the request from params.
+            # Shallow-copy the existing list (if any) to avoid mutating self.config["params"]["tools"]
+            # via the shared reference created by ** unpacking above.
+            # See: strands-agents/sdk-python#2405
+            if "tools" in request:
+                request["tools"] = list(request["tools"])
             request.setdefault("tools", []).extend(
                 {
                     "type": "function",
--- a/strands-py/src/strands/models/openai_responses.py
+++ b/strands-py/src/strands/models/openai_responses.py
@@ -549,21 +549,22 @@ class OpenAIResponsesModel(Model):
 
         # Add tools if provided
         if tool_specs:
-            # Merge with any built-in tools (e.g. web_search) already in the request from params.
-            # Shallow-copy the existing list (if any) to avoid mutating self.config["params"]["tools"]
-            # via the shared reference created by ** unpacking above.
-            # See: strands-agents/sdk-python#2405
-            if "tools" in request:
-                request["tools"] = list(request["tools"])
-            request.setdefault("tools", []).extend(
-                {
-                    "type": "function",
-                    "name": tool_spec["name"],
-                    "description": tool_spec.get("description", ""),
-                    "parameters": tool_spec["inputSchema"]["json"],
-                }
-                for tool_spec in tool_specs
-            )
+            # Merge function tools with any built-in tools (e.g. web_search) carried in from params.
+            # Build a new list rather than extending in place: ** unpacking above aliases
+            # self.config["params"]["tools"] by reference, so mutating it would duplicate every tool
+            # spec into the stored config on each call.
+            request["tools"] = [
+                *request.get("tools", []),
+                *(
+                    {
+                        "type": "function",
+                        "name": tool_spec["name"],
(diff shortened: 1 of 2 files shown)


strands-py/src/strands/models/openai_responses.py
zastrowm, line 1: Could you add a unit test to ensure this doesn't regress going forward

gautamsirdeshmukh:
Updating PR with regression test then will merge

gautamsirdeshmukh (approved):

strands-py/src/strands/models/openai_responses.py
```diff
@@ -549,16 +549,22 @@ def _format_request(
...
-                    "parameters": tool_spec["inputSchema"]["json"],
-                }
-                for tool_spec in tool_specs
-            )
+            # Merge function tools with any built-in tools (e.g. web_search) carried in from params.
+            # Build a new list rather than extending in place: ** unpacking above aliases
+            # self.config["params"]["tools"] by reference, so mutating it would duplicate every tool
+            # spec into the stored config on each call.
+            request["tools"] = [
+                *request.get("tools", []),
```
notowen333, line 557: personally I'd prefer a less python idiomatic way to achieve this

notowen333 (approved):

PR #2399, opened 2026-05-29
fix: normalize media format mappings

he-yufeng:
﻿Fixes part of #2204.

## Summary
- use explicit Anthropic image media-type mapping for common formats, including `webp`, instead of relying only on platform `mimetypes`
- accept `3gp` as a video input format alias and map it to Bedrock's `three_gp` enum value when formatting requests
- add focused regressions for WebP image MIME type and Bedrock 3GP video formatting

I kept this PR scoped to the two deterministic provider-format bugs. The broader type/guardContent expansion mentioned in the issue should be handled separately.

## To verify
- `python -m pytest tests/strands/models/test_anthropic.py::test_format_request_with_image tests/strands/models/test_anthropic.py::test_format_request_with_webp_image_uses_explicit_media_type tests/strands/models/test_bedrock.py::test_format_request_video_s3_location tests/strands/models/test_bedrock.py::test_format_request_video_3gp_maps_bedrock_enum -q`
- `python -m ruff check src/strands/models/anthropic.py src/strands/models/bedrock.py src/strands/types/media.py tests/strands/models/test_anthropic.py tests/strands/models/test_bedrock.py`
- `python -m ruff format --check src/strands/models/anthropic.py src/strands/models/bedrock.py src/strands/types/media.py tests/strands/models/test_anthropic.py tests/strands/models/test_bedrock.py`

PR #2306, opened 2026-05-20
fix(bedrock): normalize 3gp video format

he-yufeng:
## Summary

Fixes the Bedrock 3GP part of #2204.

Bedrock expects the video format enum to be `three_gp`, while common MIME/extension paths produce `3gp`, `3gpp`, or `3g2`. The SDK currently forwards the value as-is, so a valid 3GP video content block can be rejected by the Bedrock Converse API.

This normalizes those aliases when formatting Bedrock video content blocks. Other media-format gaps from #2204 are left out of this PR.

## To verify

Ran locally:

```bash
python -m pytest tests/strands/models/test_bedrock.py::test_format_request_maps_3gp_video_formats -q
python -m pytest tests/strands/models/test_bedrock.py -q
python -m ruff check src/strands/models/bedrock.py tests/strands/models/test_bedrock.py
python -m py_compile src/strands/models/bedrock.py tests/strands/models/test_bedrock.py
git diff --check
```

PR #2552, opened 2026-06-02
fix(media): correct MIME types for webp images and guardContent format validation

### LINKED ISSUE #2204: [BUG] Media type handling bugs and missing format support

xlyoung:
## Summary

Fixes #2204

Two media type handling bugs in the Python SDK that cause API errors with Bedrock and Anthropic providers:

### 1. WebP images get wrong MIME type (anthropic.py)

Python's `mimetypes.types_map` returns `"application/octet-stream"` for `.webp` instead of `"image/webp"`. This causes Anthropic API errors when sending WebP images.

**Fix:** Replace the `mimetypes.types_map` lookup with an explicit MIME type mapping for common image formats (png, jpg, jpeg, gif, webp, svg+xml, bmp), falling back to `mimetypes` for unknown formats.

### 2. guardContent wraps unsupported image formats (bedrock.py)

The guardContent wrapping logic at line 497 wraps ALL image formats, but Bedrock's `GuardrailConverseContentBlock` only supports `png` and `jpeg` images. Wrapping unsupported formats (webp, gif, etc.) causes Bedrock API validation errors.

**Fix:** Only wrap image content in guardContent when the format is `png` or `jpeg`.

### Testing

- Verified WebP images now get `"image/webp"` MIME type with Anthropic provider
- Verified guardContent is only applied for png/jpeg images with Bedrock provider
- Existing media type tests continue to pass

### DIFF
--- a/strands-py/src/strands/models/anthropic.py
+++ b/strands-py/src/strands/models/anthropic.py
@@ -131,10 +131,23 @@ class AnthropicModel(Model):
             }
 
         if "image" in content:
+            # Explicit MIME type mapping for image formats that mimetypes.types_map
+            # gets wrong (e.g. webp -> application/octet-stream instead of image/webp).
+            _IMAGE_MIME_TYPES: dict[str, str] = {
+                "png": "image/png",
+                "jpg": "image/jpeg",
+                "jpeg": "image/jpeg",
+                "gif": "image/gif",
+                "webp": "image/webp",
+                "svg+xml": "image/svg+xml",
+                "bmp": "image/bmp",
+            }
+            fmt = content["image"]["format"]
+            media_type = _IMAGE_MIME_TYPES.get(fmt) or mimetypes.types_map.get(f".{fmt}", "application/octet-stream")
             return {
                 "source": {
                     "data": base64.b64encode(content["image"]["source"]["bytes"]).decode("utf-8"),
-                    "media_type": mimetypes.types_map.get(f".{content['image']['format']}", "application/octet-stream"),
+                    "media_type": media_type,
                     "type": "base64",
                 },
                 "type": "image",
--- a/strands-py/src/strands/models/bedrock.py
+++ b/strands-py/src/strands/models/bedrock.py
@@ -490,12 +490,15 @@ class BedrockModel(Model):
                 if formatted_content is None:
                     continue
 
-                # Wrap text or image content in guardContent if this is the last user text/image message
+                # Wrap text or image content in guardContent if this is the last user text/image message.
+                # Bedrock GuardContent for images only supports png and jpeg formats.
                 if idx == last_user_text_idx and ("text" in formatted_content or "image" in formatted_content):
                     if "text" in formatted_content:
                         formatted_content = {"guardContent": {"text": {"text": formatted_content["text"]}}}
                     elif "image" in formatted_content:
-                        formatted_content = {"guardContent": {"image": formatted_content["image"]}}
+                        image_format = formatted_content["image"].get("format", "")
+                        if image_format in ("png", "jpeg"):
+                            formatted_content = {"guardContent": {"image": formatted_content["image"]}}
 
                 cleaned_content.append(formatted_content)
 


notowen333:
@strandly-the-agent review. looks like a 19 line diff so if it's clean no need for nits that churn

If yes then push a rebase fix for this

strandly-the-agent (commented):
**Reviewed `2ebb85f6` against `main@c0776c00`. The `bedrock.py` half is a real bug and the fix is correct; the `anthropic.py` half is dead code and is what's causing the conflict.** So: clean logic, mechanical cleanup.

**Rebase pushed — #3607.** Branch [`strandly-the-agent:agent-tasks/2204-guardcontent-image-gate`](https://github.com/strandly-the-agent/harness-sdk/tree/agent-tasks/2204-guardcontent-image-gate), rebased on `main@c0776c00`, `MERGEABLE`, 3 commits with **`xlyoung`'s authorship preserved** on the fix commit. I couldn't push to this PR's own branch (no write access to the fork), hence the separate PR rather than an update here — `xlyoung`, if you'd rather land it yourself, say so and I'll close #3607; the same change is also a [patch](https://gist.github.com/strandly-the-agent/a2ff3cf0fe112bdaa1934ec0cf0d6cae) you can `git am` onto your branch.

It keeps the guardContent gate, **drops the `anthropic.py` half** (already fixed on `main` by #2304 — module-level `_IMAGE_MEDIA_TYPES`, with tests), and adds the missing regression test plus a warn log on the skip, mirroring TS's `_applyGuardBlocks`.

✅ `pytest tests/strands/models/test_bedrock.py` → **180 passed** (178 + 2 new) · negative control: both new tests fail on unpatched `main` · `ruff` + `mypy` clean. Heads-up on #3607: the `2 Approvers for Bots` gate fails by design on an agent-authored PR, so it needs two human approvals.

**Questions**
1. `Fixes #2204` overstates it — that issue has 7 items and #2304/#2306 already closed two of them. #3607 says *"Fixes part of #2204 (guardContent image format validation)"*, the way #2304 phrased itself — worth doing here too if this PR stays open.
2. `kimnamu` found (on #2204) that a **png/jpeg image from an S3 location** still dies client-side, because `GuardrailConverseImageSource` only accepts `bytes` — and has a fix + test ready. Fold that in, or keep it a separate follow-up? I left it out to keep the rebase minimal.

<details>
<summary>Evidence — why the fix is right, and what I checked</summary>

**The bug is real and reachable.** `GuardrailConverseImageFormat` in the bedrock-runtime service model is exactly `['png','jpeg']`, while `ImageBlock` accepts `png/jpeg/gif/webp`. The enum isn't client-enforced, so pre-fix the request ships and Bedrock rejects the whole thing. `kimnamu` observed it live on #2204:

```
ValidationException: 1 validation error detected: Value at
'messages.1.member.content.1.member.guardContent.image.format' failed to
satisfy constraint: Member must satisfy enum value set: [jpeg, png]
```

Reach: `BedrockModel(guardrail_id=…, guardrail_version=…, guardrail_latest_message=True)` + a webp/gif screenshot as the latest user message. Ordinary multimodal call, no contrivance.

**The gate is local and safe.** No effect on `_inject_cache_point` (only branches on `"document"`) or on `_find_last_user_text_message_index` (runs on the original messages, tests key presence not format). `guardrail_redact_input` is unaffected — `_generate_redaction_events` reads the response trace, never guardContent.

**Why I added a log line.** Skipping the wrap means the image isn't sent for guardrail assessment at all, and every other skip in that function logs (`filtered_unknown_members`, the DeepSeek drop, the cache-point skip). TS already warns on this exact condition in `_applyGuardBlocks`, so this is parity, not new policy.

**Cross-checked the rest of #2204** so the reword suggestion isn't guesswork: item 1 fixed (#2304), item 3 fixed (#2306), items 4–6 would *introduce* failures if implemented literally (Converse rejects `json`/`xml` documents and `jpg` images on the wire), item 7 is a real remaining gap. Also note #2399 covers items 1 and 3, now redundant.

</details>

<details>

- ⚪ The two original halves disagreed about `jpg`: `anthropic.py` mapped it, `bedrock.py`'s tuple omits it. Moot — `jpg` is outside both `ImageFormat` (`strands-py/src/strands/types/media.py:80`) and botocore's own enum, so it fails either way. Gone with the anthropic half.
- ⚪ The dropped `anthropic.py` hunk also added `svg+xml` and `bmp`, which aren't in `ImageFormat` and aren't accepted by the Anthropic Messages API — speculative surface, and not what #2204 asked for.
- ⚪ `.get("format", "")`'s default is unreachable (`_format_request_message_content` always sets `format`). Harmless; left as-is rather than churn the diff.
- ⚪ Pre-existing, not this PR: when the latest user message is image-only *and* the format is unsupported, no guardContent block is emitted at all while `guardrailConfig` still goes out — per AWS docs that means the guardrail evaluates everything, i.e. the opposite of what `guardrail_latest_message` advertises. Same shape in the TS SDK today, so it looks like an accepted cross-SDK tradeoff rather than something to fix here.

</details>

*Reviewed by `strandly-the-agent` — an experimental agent. Worth a human eye on #3607 before it lands; happy to be told the S3 case belongs elsewhere.*

notowen333:
closing this in favor of forked #3607

PR #2116, opened 2026-04-13
feat(openai): plumb through cache tokens in metadata events

Unshure:
## Motivation

When OpenAI returns prompt caching information (`prompt_tokens_details.cached_tokens`), the OpenAI model provider currently discards it. Users have no way to see whether their requests hit the OpenAI prompt cache, which is valuable for cost optimization and debugging.

The LiteLLM provider already extracts this data, as does the experimental OpenAI Realtime bidi model — but the primary OpenAI provider was missing this support.

Resolves #2115

## Public API Changes

No public API changes. The `metadata` event emitted by `OpenAIModel.format_chunk` now includes `cacheReadInputTokens` in the usage data when OpenAI reports cached prompt tokens:

```python
# Before: metadata event usage
{"inputTokens": 1861, "outputTokens": 10, "totalTokens": 1871}

# After: metadata event usage (when cache hit occurs)
{"inputTokens": 1861, "outputTokens": 10, "totalTokens": 1871, "cacheReadInputTokens": 1792}
```

When `prompt_tokens_details` is `None` or `cached_tokens` is `None`/`0`, the field is omitted — preserving backward compatibility. The existing telemetry pipeline (tracer and metrics) already handles `cacheReadInputTokens`, so cache data flows through automatically.

Only `cacheReadInputTokens` is set because OpenAI's API does not expose a cache write token equivalent (unlike Anthropic).

Issue #2115, opened 2026-04-13
[FEATURE] Plumb through openai cache tokens


Unshure:
### Problem Statement

I would like the openai model provider to plumb through its cache tokens when making a request

### Proposed Solution

In the openai model provider, include the `prompt_token_details` in the raised event so that cache output tokens are shown to the user.

Here is an example object that includes `prompt_token_details`:
```
ChatCompletionChunk(id='chatcmpl-DUE7bRUrieQjuFiBoB3GmvE9LcAVw', choices=[], created=1776096495, model='gpt-4o-2024-08-06', object='chat.completion.chunk', service_tier='default', system_fingerprint='fp_ff05e2535c', usage=CompletionUsage(completion_tokens=10, prompt_tokens=1861, total_tokens=1871, completion_tokens_details=CompletionTokensDetails(accepted_prediction_tokens=0, audio_tokens=0, reasoning_tokens=0, rejected_prediction_tokens=0), prompt_tokens_details=PromptTokensDetails(audio_tokens=0, cached_tokens=1792)), obfuscation='4cfXGEiv')
```

### Use Case

Allow users to see if a request hit the cache

---

## Implementation Requirements

Based on repository analysis, this feature has a well-established pattern already used by the LiteLLM provider.

### Technical Approach

- **Pattern to follow**: The `LiteLLMModel.format_chunk` method in `src/strands/models/litellm.py` (lines 217-230) already implements the exact same cache token extraction from the OpenAI-compatible usage object. The OpenAI provider should adopt the same approach.
- **Type support**: The `Usage` TypedDict in `src/strands/types/event_loop.py` already defines optional `cacheReadInputTokens` and `cacheWriteInputTokens` fields.
- **Telemetry pipeline**: The tracer (`src/strands/telemetry/tracer.py`) and metrics (`src/strands/telemetry/metrics.py`) already handle `cacheReadInputTokens` and `cacheWriteInputTokens` — no changes needed there.
- **Bidi precedent**: The OpenAI Realtime bidi model (`src/strands/experimental/bidi/models/openai_realtime.py`) already extracts `cached_tokens` in a similar way.

### Files to Modify

1. **`src/strands/models/openai.py`** — Update the `format_chunk` method's `"metadata"` case (lines 548-560) to extract `prompt_tokens_details.cached_tokens` into `cacheReadInputTokens` in the usage data.
   - Use `getattr()` with defaults for safety (same pattern as LiteLLM).
[...]

PR #2555, opened 2026-06-02
fix(openai-responses): surface cache read tokens in metadata chunk

### LINKED ISSUE #2407: OpenAIResponsesModel drops prompt cache tokens (cached_tokens not plumbed to cacheReadInputTokens)
## Summary

`OpenAIResponsesModel` does not propagate OpenAI prompt-cache usage. The Responses API returns `usage.input_tokens_details.cached_tokens`, but the model's `_format_chunk` metadata branch only maps `input_tokens` / `output_tokens` / `total_tokens`. As a result `cacheReadInputTokens` is never set, and telemetry (`gen_ai.usage.cache_read_input_tokens`, set in `telemetry/tracer.py` only when `cacheReadInputTokens` is present) never reflects cache reads for the Responses path.

This is an asymmetry with the chat model: `OpenAIModel.format_chunk` already plumbs `prompt_tokens_details.cached_tokens` into `cacheReadInputTokens` (added in #2115 / #2116, released in v1.36.0). The Responses model was not updated.

## Affected code

`src/strands/models/openai_responses.py`, `_format_chunk`, the `case "metadata"` branch (around the usage dict construction):

```python
case "metadata":
    return {
        "metadata": {
            "usage": {
                "inputTokens": getattr(event["data"], "input_tokens", 0),
                "outputTokens": getattr(event["data"], "output_tokens", 0),
                "totalTokens": getattr(event["data"], "total_tokens", 0),
                # cached tokens from input_tokens_details are dropped here
            },
            "metrics": {"latencyMs": 0},
        },
    }
```

Compare with `src/strands/models/openai.py` (`format_chunk`), which reads `prompt_tokens_details.cached_tokens` and sets `cacheReadInputTokens`.

## Impact

- `AgentResult.metrics.accumulated_usage` and OTel spans for Responses-based agents do not expose cache read tokens.
- Downstream observability/cost tooling that relies on `cacheReadInputTokens` cannot see prompt cache savings for the Responses API, even though OpenAI is caching and returning `cached_tokens`.

Verified against a live OpenAI Responses API call (streaming): `response.completed`'s `usage.input_tokens_details.cached_tokens` is populated on cache hits, so the data is available at the point `_format_chunk` receives `final_usage`.

## Suggested fix

Mirror the chat model: when building the metadata usage dict, read `cached_tokens` from `input_tokens_details` and set `cacheReadInputTokens`:

```python
usage = {
    "inputTokens": getattr(event["data"], "input_tokens", 0),
    "outputTokens": getattr(event["data"], "output_tokens", 0),
    "totalTokens": getattr(event["data"], "total_tokens", 0),
}
details = getattr(event["data"], "input_tokens_details", None)
cached = getattr(details, "cached_tokens", None)
if cached:
    usage["cacheReadInputTokens"] = cached
```

`cacheWriteInputTokens` is not applicable (OpenAI does not return cache-write counts), consistent with #2115.

## Version

Observed on v1.36.0; `openai_responses.py` metadata branch is unchanged on main / v1.41.0.

gautham18113:
## Summary

Fixes #2407.

`OpenAIResponsesModel._format_chunk` built the metadata usage dict with only the three scalar token counts (`inputTokens`, `outputTokens`, `totalTokens`). The `input_tokens_details.cached_tokens` field returned by the Responses API on cache hits was silently dropped, so `cacheReadInputTokens` was never populated and cache reads were invisible to telemetry and cost tooling.

The fix mirrors what `OpenAIModel.format_chunk` already does for the Chat Completions path (added in #2115 / #2116): read `input_tokens_details.cached_tokens` and set `cacheReadInputTokens` when it is present.

**Root cause (before):**
```python
case "metadata":
    return {
        "metadata": {
            "usage": {
                "inputTokens": getattr(event["data"], "input_tokens", 0),
                "outputTokens": getattr(event["data"], "output_tokens", 0),
                "totalTokens": getattr(event["data"], "total_tokens", 0),
                # input_tokens_details.cached_tokens dropped here
            },
            ...
        }
    }
```

**After:**
```python
case "metadata":
    usage_data: Usage = {
        "inputTokens": getattr(event["data"], "input_tokens", 0),
        "outputTokens": getattr(event["data"], "output_tokens", 0),
        "totalTokens": getattr(event["data"], "total_tokens", 0),
    }
    if token_details := getattr(event["data"], "input_tokens_details", None):
        if cached := getattr(token_details, "cached_tokens", None):
            usage_data["cacheReadInputTokens"] = cached
    return {"metadata": {"usage": usage_data, "metrics": {"latencyMs": 0}}}
```

Note: `cacheWriteInputTokens` is not set — the OpenAI Responses API does not return cache-write counts, consistent with the Chat Completions fix in #2115.

## Changes

- `src/strands/models/openai_responses.py`: add `Usage` import; build `usage_data` dict first then conditionally populate `cacheReadInputTokens`
- `tests/strands/models/test_openai_responses.py`: add two `_format_chunk` parametrize cases (with and without cache tokens); add `test_stream_cache_tokens_propagated` and `test_stream_no_cache_tokens_when_absent` end-to-end streaming tests; add explicit `input_tokens_details=None` to existing usage mocks so they are not ambiguous

## Test plan

- [x] `pytest tests/strands/models/test_openai_responses.py` — 93 passed, 0 failed
- [x] New test `test_stream_cache_tokens_propagated` verifies `cacheReadInputTokens` is set when `input_tokens_details.cached_tokens` is present on the completed response
- [x] New test `test_stream_no_cache_tokens_when_absent` verifies `cacheReadInputTokens` is absent when `input_tokens_details` is `None`
- [x] Existing tests updated to set `input_tokens_details=None` explicitly on usage mocks to prevent false positives


strands-py/src/strands/models/openai_responses.py
```diff
@@ -837,13 +838,19 @@ def _format_chunk(self, event: dict[str, Any]) -> StreamEvent:
 
             case "metadata":
                 # Responses API uses input_tokens/output_tokens naming convention
+                usage_data: Usage = {
+                    "inputTokens": getattr(event["data"], "input_tokens", 0),
+                    "outputTokens": getattr(event["data"], "output_tokens", 0),
+                    "totalTokens": getattr(event["data"], "total_tokens", 0),
+                }
+
+                if token_details := getattr(event["data"], "input_tokens_details", None):
```
mehtarac: is `input_tokens_details` needed? what additional information does it provide that `input_tokens` may not be sufficient for?

gautham18113: `input_tokens` just tells us the total, `input_tokens_details.cached_tokens` tells us how much of that total was a cache hit. They're different things.

We need this because the Strands SDK's telemetry layer (telemetry/tracer.py) already has code to emit gen_ai.usage.cache_read_input_tokens as an OTEL span attribute, but only when `cacheReadInputTokens` is populated. Without this fix, that attribute is always missing for Responses API calls, so we'd have no visibility into whether prompt caching is actually working, even when OpenAI is caching and returning the data.

The Chat Completions model already handles this via `prompt_tokens_details.cached_tokens`. This just brings the Responses model in line with that.

kimnamu (commented):
Thanks for this, @gautham18113 — clean, well-scoped fix. I'm not a maintainer, just another SDK user who reviewed this carefully and ran the branch locally.

A few things I verified:

**1. The implementation matches the Chat Completions sibling exactly — this is the safest possible framing.** Your `case "metadata"` branch is character-for-character identical to `OpenAIModel.format_chunk` (`openai.py:581-599`), only swapping `input_tokens_details`/`input_tokens` for the Responses naming. Confirmed the attribute path against the real SDK type too: `openai==2.43.0`'s `ResponseUsage.input_tokens_details` is an `InputTokensDetails` whose only field is `cached_tokens` — so `getattr(event["data"], "input_tokens_details", None).cached_tokens` is correct.

**2. Omitting `cacheWriteInputTokens` is correct, not an oversight.** For anyone wondering: `InputTokensDetails` has no cache-write field — the Responses API simply doesn't return a cache-write count (caching is automatic/transparent), so there's nothing to plumb. This is consistent with #2115.

**3. The `getattr` guards handle every missing/edge case gracefully** — I exercised them directly:
```
cached_tokens=80   -> cacheReadInputTokens: 80
cached_tokens=0    -> key omitted   (falsy guard; tracer reads .get(...,0), so no downstream impact)
details=None       -> key omitted
cached_tokens absent / =None -> key omitted, no crash
```

**4. The new tests genuinely catch the bug (red/green).** Reverting just the source fix while keeping the tests makes `test_stream_cache_tokens_propagated` and the `test_format_chunk` cache-hit case fail (`cacheReadInputTokens` missing); restoring it passes. Full file: 93 passed. Nice touch adding the explicit `input_tokens_details=None` to the existing mocks to remove ambiguity.

**One thing that will block CI** (`hatch fmt --linter` / ruff with `select=["E"]`, line-length 120): the 7 lines where you appended `input_tokens_details=...` to existing one-line `Mock(...)` calls now exceed 120 chars (E501):
```
tests/strands/models/test_openai_responses.py: lines 458, 518, 627, 660, 717, 761, 797
```
The formatter fixes all of them automatically — just run:
```bash
cd strands-py && hatch fmt --formatter   # == ruff format
```
which wraps them, e.g.:
```python
        response=unittest.mock.Mock(
            usage=unittest.mock.Mock(input_tokens=10, output_tokens=5, total_tokens=15, input_tokens_details=None)
        ),
```
After that, `ruff check` reports All checks passed (verified locally).

Not blocking, just the lint above. Solid fix — I'd love to see it land.

*(This review was prepared with the help of an AI agent (Claude Code); I ran the branch, reproduced the edge cases, and verified the lint failure and fix myself before posting.)*

mehtarac (dismissed):

mehtarac:
Hi @gautham18113! the changes look good to me! The required CI check for lint is failing. If you are able to have that fixed, Im happy to merge this PR.

mehtarac (approved):

JackYPCOnline (approved):

opieter-aws (approved):

PR #2563, opened 2026-06-02
feat: integrate Sandbox with Agent

gautamsirdeshmukh:
## Description

Adds the [Sandbox](https://github.com/strands-agents/sdk-python/blob/main/strands-ts/src/sandbox/base.ts) as a first-class parameter on `Agent`.

> In Node, agents get a host-execution fallback automatically via a "node" export condition in package.json that loads index.node.ts. In browsers, accessing agent.sandbox without explicit configuration throws. 

> To make this Node/browser bundler compatible fallback split possible, we (also) introduce `createDefaultSlot<T>()` as a reusable primitive for environment-conditional defaults.

This is the third PR split off from https://github.com/strands-agents/sdk-typescript/pull/1011, preceded by the base interface merged in https://github.com/strands-agents/sdk-typescript/pull/1090 and Docker/SSH implementations merged in https://github.com/strands-agents/sdk-typescript/pull/1110.

## Developer Experience

Explicitly Configured Sandbox
```
const agent = new Agent({
  model: new BedrockModel(),
  sandbox: new DockerSandbox({ containerId: 'my-container' }),
})

agent.sandbox // DockerSandbox
```

Node: works without configuration (falls back to host-level execution)
```
const agent = new Agent({ model: new BedrockModel() })

agent.sandbox // NotASandboxLocalEnvironment
```

Browser: throws if not explicitly configured
```
...
agent.sandbox // throws DefaultNotConfiguredError
```

## Related Issues

- https://github.com/strands-agents/sdk-typescript/pull/1011 (parent mega-PR)
- https://github.com/strands-agents/sdk-typescript/pull/1090 (base Sandbox / PosixShellSandbox interfaces)
- https://github.com/strands-agents/sdk-typescript/pull/1110 (Docker / SSH implementations)

## Documentation PR

Not yet

## Type of Change

New feature

## Testing

- [x] I ran all tests + added several

## Checklist

----

### DIFF
--- a/strands-ts/src/__fixtures__/register-node-defaults.ts
+++ b/strands-ts/src/__fixtures__/register-node-defaults.ts
@@ -1,0 +1,6 @@
+import { defaultSandbox } from '../sandbox/default.js'
+import { NotASandboxLocalEnvironment } from '../sandbox/not-a-sandbox-local-environment.js'
+
+// In production, index.node.ts registers this on import. Tests don't go through that entry
+// point, so this setup file does it instead.
+defaultSandbox.set(new NotASandboxLocalEnvironment())
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -91,6 +91,8 @@ import { isInterruptResponseContent, type InterruptResponseContent } from '../ty
 import { takeSnapshot as takeSnapshotInternal, loadSnapshot as loadSnapshotInternal } from './snapshot.js'
 import type { TakeSnapshotOptions } from './snapshot.js'
 import type { Snapshot } from '../types/snapshot.js'
+import type { Sandbox } from '../sandbox/base.js'
+import { defaultSandbox } from '../sandbox/default.js'
 
 /**
  * Recursive type definition for nested tool arrays.
@@ -230,6 +232,20 @@ export type AgentConfig = {
    * Defaults to `'concurrent'`. See {@link ToolExecutorStrategy} for details.
    */
   toolExecutor?: ToolExecutorStrategy
+  /**
+   * Execution environment for running commands, code, and file operations.
+   * When provided, sandbox-aware tools route operations through it.
+   *
+   * Two distinct intents, even though they resolve to the same host execution
+   * in Node today:
+   * - Omitted: use the environment's default sandbox (host execution in Node).
+   *   This default is the slot reserved for richer behavior later.
+   * - `false`: explicitly opt out of a managed sandbox and run on the host.
+   *
+   * Keep `false` distinct from omitting so the opt-out stays stable even if the
+   * default changes.
+   */
+  sandbox?: Sandbox | false
 }
 
 /** Default name assigned to agents when none is provided. */
@@ -301,6 +317,18 @@ export class Agent implements LocalAgent, InvokableAgent {
    */
   public readonly memoryManager?: MemoryManager | undefined
 
+  private readonly _sandbox: Sandbox | false | undefined
+
+  /**
+   * Execution environment for running commands, code, and file operations.
+   *
+   * @throws DefaultNotConfiguredError if no sandbox is configured for this
+   * environment (e.g. browsers, where no host default is registered).
+   */
+  get sandbox(): Sandbox {
+    return this._sandbox || defaultSandbox.get()
+  }
+
   private readonly _hooksRegistry: HookRegistryImplementation
   private readonly _pluginRegistry: PluginRegistry
   private readonly _interventionRegistry: InterventionRegistry
@@ -342,6 +370,7 @@ export class Agent implements LocalAgent, InvokableAgent {
         : config?.memoryManager
           ? new MemoryManager(config.memoryManager)
           : undefined
+    this._sandbox = config?.sandbox
 
     if (typeof config?.model === 'string') {
       this.model = new BedrockModel({ modelId: config.model })
--- a/strands-ts/src/default-slot.ts
+++ b/strands-ts/src/default-slot.ts
@@ -1,0 +1,22 @@
+import { DefaultNotConfiguredError } from './errors.js'
+
+export interface DefaultSlot<T> {
+  set(value: T): void
+  get(): T
+}
+
+export function createDefaultSlot<T>(errorMessage: string): DefaultSlot<T> {
+  // A unique sentinel for "unset" so any value of T -- including undefined or
+  // null -- can be stored and read back without colliding with the empty state.
+  const EMPTY = Symbol('empty')
+  let value: T | typeof EMPTY = EMPTY
+  return {
+    set(v): void {
+      value = v
+    },
+    get(): T {
+      if (value === EMPTY) throw new DefaultNotConfiguredError(errorMessage)
+      return value
+    },
+  }
+}
--- a/strands-ts/src/types/agent.ts
+++ b/strands-ts/src/types/agent.ts
@@ -1,3 +1,4 @@
+import type { Sandbox } from '../sandbox/base.js'
 import type { StateStore } from '../state-store.js'
 import type { ContentBlock, ContentBlockData, Message, MessageData, StopReason, SystemPrompt } from './messages.js'
 import type { Interrupt } from '../interrupt.js'
@@ -261,6 +262,14 @@ export interface LocalAgent {
    * The tool registry for registering tools with the agent.
    */
   readonly toolRegistry: ToolRegistry
+
+  /**
+   * Execution environment for running commands, code, and file operations.
+   *
+   * @throws DefaultNotConfiguredError if no sandbox is configured for this
+   * environment (e.g. browsers, where no host default is registered).
+   */
+  readonly sandbox: Sandbox
 
   /**
    * The model provider used by the agent for inference.
--- a/strands-ts/package.json
+++ b/strands-ts/package.json
@@ -6,6 +6,10 @@
   "module": "dist/src/index.js",
   "types": "dist/src/index.d.ts",
   "type": "module",
+  "sideEffects": [
+    "./dist/src/index.node.js",
+    "./src/index.node.ts"
+  ],
   "files": [
     "dist",
     "README.md",
@@ -14,6 +18,7 @@
   "exports": {
     ".": {
       "types": "./dist/src/index.d.ts",
+      "node": "./dist/src/index.node.js",
       "default": "./dist/src/index.js"
     },
     "./models/anthropic": {
--- a/strands-ts/src/__tests__/default-slot.test.ts
+++ b/strands-ts/src/__tests__/default-slot.test.ts
@@ -1,0 +1,38 @@
+import { describe, expect, it } from 'vitest'
+import { createDefaultSlot } from '../default-slot.js'
+import { DefaultNotConfiguredError } from '../errors.js'
+
+describe('createDefaultSlot', () => {
+  it('throws DefaultNotConfiguredError when read before configured', () => {
+    const slot = createDefaultSlot<string>('not set')
+    expect(() => slot.get()).toThrow(DefaultNotConfiguredError)
+    expect(() => slot.get()).toThrow('not set')
+  })
+
+  it('returns the configured value after set', () => {
+    const slot = createDefaultSlot<string>('not set')
+    slot.set('value')
+    expect(slot.get()).toBe('value')
+  })
+
+  it('stores falsy values that the sentinel must not confuse with unset', () => {
+    const numbers = createDefaultSlot<number>('no number')
+    numbers.set(0)
+    expect(numbers.get()).toBe(0)
+
+    const nullable = createDefaultSlot<string | null>('no value')
+    nullable.set(null)
+    expect(nullable.get()).toBeNull()
+
+    const maybe = createDefaultSlot<string | undefined>('no value')
+    maybe.set(undefined)
+    expect(maybe.get()).toBeUndefined()
+  })
+
+  it('overwrites on a subsequent set', () => {
+    const slot = createDefaultSlot<string>('not set')
+    slot.set('first')
+    slot.set('second')
+    expect(slot.get()).toBe('second')
+  })
+})
--- a/strands-ts/src/errors.ts
+++ b/strands-ts/src/errors.ts
@@ -242,3 +242,15 @@ export class CancelledError extends Error {
     this.name = 'CancelledError'
   }
 }
+
+/**
+ * Thrown when a default value is read before one has been configured.
+ *
+ * Catch it to distinguish a missing-default condition from other runtime errors.
+ */
+export class DefaultNotConfiguredError extends Error {
+  constructor(message: string) {
+    super(message)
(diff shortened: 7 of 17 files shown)


strands-ts/src/default-slot.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export function createDefaultSlot<T>(errorMessage: string): DefaultSlot<T> {
+  let value: T | undefined
+  return {
+    set(v): void {
+      value = v
+    },
+    get(): T {
+      if (value !== undefined) return value
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Good suggestion, adding sentinel for explicit unset

strands-ts/src/types/agent.ts
```diff
@@ -262,6 +263,11 @@ export interface LocalAgent {
    */
   readonly toolRegistry: ToolRegistry
 
+  /**
+   * Execution environment for running commands, code, and file operations.
+   */
+  readonly sandbox: Sandbox
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Adding

strands-ts/src/default-slot.ts
```diff
@@ -0,0 +1,24 @@
+export class DefaultNotConfiguredError extends Error {
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Moved to errors.ts + exported

strands-ts/src/agent/agent.ts
```diff
@@ -289,6 +305,18 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+  private readonly _sandbox: Sandbox | false | undefined
+
+  /**
+   * Execution environment for running commands, code, and file operations.
+   *
+   * @throws DefaultNotConfiguredError if no sandbox is configured for this
+   * environment (e.g. browsers, where no host default is registered).
+   */
+  get sandbox(): Sandbox {
+    return this._sandbox || defaultSandbox.get()
```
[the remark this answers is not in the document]
gautamsirdeshmukh: This is the intended design right now. At some point in the indefinite future we will have the default/undefined path route to a different default sandbox or sorts, but for now default/undefined and false both go to the NotASandbox (host).

strands-ts/src/default-slot.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export interface DefaultSlot<T> {
+  set(value: T): void
+  get(): T
+}
+
+export function createDefaultSlot<T>(errorMessage: string): DefaultSlot<T> {
+  let value: T | undefined
+  return {
+    set(v): void {
```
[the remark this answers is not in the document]
gautamsirdeshmukh: This default slot registry is used internally with a single caller, a guard would be overly defensive

gautamsirdeshmukh:
@agent-of-mkmeral feel free to have a look! "do the works" as your creator says

agent-of-mkmeral:
Had a proper look @gautamsirdeshmukh — checked out the PR head, read the diff, and ran the suites locally. This is clean, well-architected work. ✅

## What I verified locally (Node 22)

**New tests for this PR — all green:**
```
✓ src/__tests__/default-slot.test.ts                            (4 tests)
✓ src/sandbox/__tests__/default.test.node.ts                    (4 tests)
✓ src/sandbox/__tests__/not-a-sandbox-local-environment.test.node.ts (15 tests)
  23 passed | Type Errors: none
```

**No regressions** — full agent unit suite:
```
✓ src/agent/__tests__  →  14 files, 380 passed | Type Errors: none
```

**Browser-bundler compatibility** (the core goal of the PR) — `check:browser-bundle` passes:
```
esbuild src/index.ts --bundle --platform=browser --format=esm  →  412.3kb ⚡ Done in 63ms
```

## Design review

- **`createDefaultSlot<T>` sentinel pattern** is the right call — `Symbol('empty')` correctly distinguishes "unset" from valid falsy values (`0`, `null`, `undefined`), and the test explicitly covers all three. 👍
- **Node/browser split** via the `"node"` export condition + `index.node.ts` side-effect registration is well done. The `sideEffects` allowlist in `package.json` is exactly what's needed to stop bundlers from tree-shaking the load-bearing `defaultSandbox.set(...)`. The inline comment warning future maintainers not to mark it side-effect-free is a nice touch.
- **`sandbox: false` vs omission** resolving identically today is documented as intentional forward-compat in both the `AgentConfig` docstring and the test name. Fine as-is; your explanation in the thread covers the future divergence path.
- **`DefaultNotConfiguredError`** living in `errors.ts` and exported from `index.ts`, plus `@throws` on the `LocalAgent` interface — good for browser consumers doing `instanceof` checks.
- **`NotASandboxLocalEnvironment`** safety: language validated against `LANGUAGE_PATTERN`, `cwd` shell-quoted, code piped via base64+EOF heredoc. Correctly labeled "no isolation" for the host fallback. All 5 review threads resolved.

## On the red CI ❌ (non-blocking)

The only failing check is **Run integration tests** (2 of 1188 failed). I traced both — neither is related to this PR:

1. `test/integ/sandbox/docker.test.node.ts` → *"mounts host volumes when configured"* — Docker volume-mount flakiness in CI, file not touched by this PR.
2. `test/integ/agent.test.ts` → *"handles tool invocation"* (~120s, retry x1) — long-running browser integ test timeout, file not touched by this PR.

Both are live-service/Docker integ tests, and the **CI Gate check is green**, so they're non-blocking. Separately, the `type-check` failure in `test/integ/.../steering.test.node.ts` also **reproduces on `main`** (it's not in this PR's changeset) — pre-existing, unrelated.

## Verdict

LGTM. 412→ minimal, focused diff; solid test strategy (isolated slot tests + node/browser projects + CJS/ESM package tests); review feedback genuinely addressed. The red X is unrelated flaky integ tests, not a defect in this PR. 🚀

strands-ts/src/agent/agent.ts
```diff
@@ -301,6 +317,18 @@ export class Agent implements LocalAgent, InvokableAgent {
...
```
mkmeral, line 329: do we actually set the `this._sandbox` to `defaultSandbox.get()`? or does it actually create a new one, each time i call `agent.sandbox`?  i.e. is it the same instance or not?

mkmeral: > Worth noting: that singleton is shared across all agents that fall back to the default, not per-agent. That's harmless here because NotASandboxLocalEnvironment is stateless (no instance fields; everything keys off process.cwd() and the call args). If a future default sandbox holds per-agent state, this shared-singleton fallback would need rethinking. 

this is what kiro told me, scary

gautamsirdeshmukh: It's the same singleton instance, not recreated per call. Also, like Kiro said, it's a non-issue today since `NotASandboxLocalEnvironment` is stateless. If/when we add an actual default sandbox instead of falling back to host (and it holds per-agent state), the update is simple and non-breaking (make it a factory), something like this:

Before:
```
  // default.ts
  export const defaultSandbox = createDefaultSlot<Sandbox>('No Sandbox configured...')

  // index.node.ts
  defaultSandbox.set(new NotASandboxLocalEnvironment())

  // agent.ts
  get sandbox(): Sandbox {
    return this._sandbox || defaultSandbox.get()
  }
```

  After:
```
  // default.ts
  export const defaultSandbox = createDefaultSlot<() => Sandbox>('No Sandbox configured...')

  // index.node.ts
  defaultSandbox.set(() => new StrandsSandbox())

  // agent.ts
  private _resolvedDefault?: Sandbox
  get sandbox(): Sandbox {
    if (this._sandbox) return this._sandbox
    return (this._resolvedDefault ??= defaultSandbox.get()())  // one per agent
  }
```

agent-of-mkmeral:
Re @mkmeral's question on the `agent.sandbox` getter (`agent.ts:329`) — *"is it the same instance or a new one each call?"* and the shared-singleton "scary" concern:

I checked out the PR head and wrote a probe to answer it empirically rather than just reasoning about it. **Same instance, not recreated per call:**

```ts
const a = new Agent({ model })
a.sandbox === a.sandbox          // ✅ true  (same ref across reads)

const a1 = new Agent({ model })
const a2 = new Agent({ model })
a1.sandbox === a2.sandbox        // ✅ true  (ONE shared singleton across agents)

new Agent({ model, sandbox: false }).sandbox === a1.sandbox  // ✅ true
```
All 3 assertions pass. The getter `return this._sandbox || defaultSandbox.get()` just hands back the cached slot value — no allocation per read.

**On the "scary" shared-singleton part — it's genuinely a non-issue today.** I grepped `NotASandboxLocalEnvironment` and the base `Sandbox` class: **zero instance fields.** `_resolvePath` is a pure helper, and every operation keys off `process.cwd()` + the call args. So the shared singleton holds no per-agent state that could leak between agents.

@gautamsirdeshmukh's factory migration path is exactly right *if/when* a real default sandbox ever holds per-agent state: `createDefaultSlot<() => Sandbox>` + a per-agent `_resolvedDefault ??= defaultSandbox.get()()` cache gives one-instance-per-agent, non-breaking.

**TL;DR:** shared stateless singleton today (safe), clean non-breaking escape hatch reserved for the stateful future. No action needed on this PR — still LGTM. 🚀

agent-of-mkmeral:
Re-checked on the updated head `de6085a8` (rebased/squashed onto latest main since my last look). Pulled it fresh and re-ran locally on Node 22 — still clean: ✅

```
✓ src/__tests__/default-slot.test.ts                                 (4)
✓ src/sandbox/__tests__/default.test.node.ts                         (4)
✓ src/sandbox/__tests__/not-a-sandbox-local-environment.test.node.ts (17)
  25 passed | Type Errors: none
```
```
esbuild src/index.ts --bundle --platform=browser --format=esm  →  412.5kb ⚡  (check:browser-bundle passes)
```

What's changed since June 4:
- **CI is now fully green** — the integ-test red X I traced earlier (Docker volume-mount + long browser integ flakes) cleared on the rebase. `statusCheckRollup: SUCCESS`.
- The `not-a-sandbox` suite picked up 2 more tests (15→17) — nice, more coverage on the host-fallback path.
- **@mkmeral's instance-identity thread is resolved** — the empirical answer (shared stateless singleton today, factory escape-hatch for the stateful future) settled it. All 6 threads now resolved.

Nothing new to fix on my end — the diff is the same focused 17-file change, design is sound, and it's `MERGEABLE`. Only thing outstanding is `reviewDecision: REVIEW_REQUIRED`, i.e. it needs a human maintainer's approving review to merge. From my side: LGTM, ship it. 🚀

strands-ts/src/__fixtures__/register-node-defaults.ts
```diff
@@ -0,0 +1,6 @@
+import { defaultSandbox } from '../sandbox/default.js'
```
mkmeral, line 1: (blocking merge until comment is resolved)

mkmeral (approved):

strands-ts/src/__fixtures__/register-node-defaults.ts
agent-of-mkmeral: **TL;DR:** The blocker = test/prod drift risk — this fixture duplicates the production registration from `index.node.ts`. Verified fix: extract a shared `registerNodeDefaults()` called by **both** the prod entry point and this fixture, so they can't drift. **61/61 tests pass, browser bundle check green** on `de6085a8`. Copy-paste patch below. ⚠️ Warning: the naive fix (`import '../index.node.js'` in the fixture) **breaks 32 tracer tests** — details collapsed.

<details>
<summary>✅ The patch (3 files, copy-paste ready)</summary>

```ts
// src/sandbox/register-node-defaults.ts  (new)
import { defaultSandbox } from './default.js'
import { NotASandboxLocalEnvironment } from './not-a-sandbox-local-environment.js'

/**
 * Registers the Node-specific default sandbox. Called by the production Node
 * entry point (index.node.ts) and by the unit-node vitest setup file, so the
 * two can never drift.
 */
export function registerNodeDefaults(): void {
  defaultSandbox.set(new NotASandboxLocalEnvironment())
}
```
```ts
// src/index.node.ts
// (keep the existing side-effect warning comment)
import { registerNodeDefaults } from './sandbox/register-node-defaults.js'

registerNodeDefaults()

export * from './index.js'
```
```ts
// src/__fixtures__/register-node-defaults.ts
// In production, index.node.ts registers this on import. Tests don't go through that
// entry point (importing it here would load the full SDK module graph before vi.mock
// can intercept, breaking module mocks), so this setup file calls the same shared
// registration function instead.
import { registerNodeDefaults } from '../sandbox/register-node-defaults.js'

registerNodeDefaults()
```

</details>

<details>
<summary>❌ Why the naive fix breaks 32 tracer tests</summary>

Making the fixture `import '../index.node.js'` directly *looks* right (tests would exercise the literal prod side effect), but vitest setup files run before test files — so the import pulls the **entire SDK module graph** into the module cache before `vi.mock` in any test file can intercept. Module mocks silently stop applying. Verified locally: `agent.tracer.test.node.ts` fails **32 of 36 tests** with this approach. The shared-function approach keeps the fixture's import graph tiny (just the sandbox modules), so mocks keep working.

</details>

<details>
<summary>🔬 Verification details (run locally on head de6085a8)</summary>

- `agent.tracer.test.node.ts` + `default.test.node.ts` + `default-slot.test.ts` + `not-a-sandbox-local-environment.test.node.ts` → **61/61 pass, no type errors** (the tracer suite is the canary for the vi.mock pitfall)
- `check:browser-bundle` → still passes, 412.5kb (unchanged — `index.node.ts` stays on the `sideEffects` allowlist, bundler safety intact)
- Browser behavior unchanged: nothing imports the new module from `index.ts`, so the browser graph never registers a default

</details>

@gautamsirdeshmukh I only have read access (can't push to your fork) — the patch above is verified and copy-paste ready. Once applied, this thread can be resolved and the PR is unblocked (mkmeral already approved; CI is green on the current head).

gautamsirdeshmukh: Will address this in a follow up PR, this PR doesn't break anything (callout is about what MIGHT happen if another hypothetical change were to be made, valid but not blocking)

agent-of-mkmeral: Fair assessment — I agree the risk is **latent, not active**. Today the fixture and `index.node.ts` are byte-for-byte the same one-line registration, the code is correct, and all suites pass. The drift hazard only materializes if a future change updates one site and not the other — so "follow-up PR" vs "fix now" are both defensible; it's a judgment call on whether to land protection before or after the hazard can exist.

Two practical notes for whichever way it goes:
- The verified patch above is copy-paste ready for the follow-up (3 small files, 61/61 tests pass, bundle check green) — near-zero effort whenever it's picked up.
- If deferred, worth filing the follow-up issue before merge so the unresolved-thread context doesn't get lost.

Since this is @mkmeral's blocking thread, the unblock decision is his: resolve + merge now with a tracked follow-up, or fold the 3-file patch into this PR first. Either path works from my side — the PR itself remains LGTM.

PR #2564, opened 2026-06-02
fix: update package-lock.json file to resolve wasm dependencies

mehtarac:
## Description
<!-- Provide a detailed description of the changes in this PR -->

The package-lock.json was missing resolved entries for @oxc-parser platform bindings other than darwin-arm64 (the machine it was originally generated on). This meant npm ci on Linux CI runners couldn't install `@oxc-parser/binding-linux-x64-gnu`, causing the WASM build step to fail with "Failed to load native binding". Regenerating the lockfile adds resolution entries for all platforms, allowing the WASM component to build on Linux in CI.                                                                                                                    

#### Side effect: Bedrock SDK version bump                                                                                                                                                                                                        
The regeneration also bumped `@aws-sdk/client-bedrock-runtime` from `3.1037.0` to `3.1058.0`. The newer version added "system" to the `ConversationRole` type (previously only "user" | "assistant"), which caused two type errors in bedrock.ts where the Bedrock role is assigned to the Strands Role type (which is "user" | "assistant" only). Fixed with a cast. This is safe because Bedrock response messages always have role "assistant".

## Related Issues

<!-- Link to related issues using #issue-number format -->

## Documentation PR

<!-- If this PR requires documentation changes, include them in this PR under site/ -->

## Type of Change

<!-- Choose one of the following types of changes, delete the rest -->

Bug fix
New feature
Breaking change
Documentation update
Other (please describe):

## Testing

## Checklist

----

### DIFF
--- a/strands-ts/src/models/bedrock.ts
+++ b/strands-ts/src/models/bedrock.ts
@@ -53,7 +53,7 @@ import {
   type StreamOptions,
   resolveConfigMetadata,
 } from '../models/model.js'
-import type { ContentBlock, Message, StopReason, ToolUseBlock } from '../types/messages.js'
+import type { ContentBlock, Message, Role, StopReason, ToolUseBlock } from '../types/messages.js'
 import type { ImageSource, VideoSource, DocumentSource } from '../types/media.js'
 import type { CitationsDelta, ModelStreamEvent, ReasoningContentDelta, Usage } from '../models/streaming.js'
 import type { Citation, CitationLocation, CitationsBlockData } from '../types/citations.js'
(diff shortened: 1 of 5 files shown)


strands-ts/src/models/bedrock.ts
```diff
@@ -1247,7 +1247,7 @@ export class BedrockModel extends Model<BedrockModelConfig> {
     // Message start
     const output = ensureDefined(event.output, 'event.output')
     const message = ensureDefined(output.message, 'output.message')
-    const role = ensureDefined(message.role, 'message.role')
+    const role = ensureDefined(message.role, 'message.role') as Role
```
zastrowm: > This is safe because Bedrock response messages always have role "assistant".

If this is true, why is `system` allowed now?  I'm a little concerned that `system` sneaks in somehow.

Can we just throw in this case?

mehtarac: updated!

zastrowm (approved):

PR #2615, opened 2026-06-03
feat: add middleware system for wrapping agent stages

JackYPCOnline:
Migrated to https://github.com/strands-agents/harness-sdk/pull/2681 to cut down on un-needed history

-----

Port of strands-agents/sdk-typescript#1068 into the mono-repo.

Original author: @zastrowm

-----

## Summary

Adds a middleware system that wraps agent stages using async generator handlers. Middleware controls flow (retry, cache, transform, short-circuit).

**Public API:** `agent.addMiddleware(stage, handler)` — returns a cleanup function.

Three built-in stages: `InvokeModelStage`, `ExecuteToolStage`, `AgentStreamStage`. 

## Motivation

Hooks let you observe operations and set flags, but they don't let you wrap them. If you want to do something both before and after a model call (timing it, adding a span, catching errors), hooks force you to manage state across two separate callbacks. With middleware you can wrap the entire invocation keeping state within your callback:

```typescript
agent.addMiddleware(InvokeModelStage, async function* (context, next) {
  const start = Date.now()
  const result = yield* next(context)
  metrics.record(Date.now() - start)
  return result
})
```

Beyond the before/after pattern, middleware also makes several other use cases much more natural to express: caching, input transformation, short-circuiting, and error handling. All of these are awkward or impossible with hooks alone.

## Public API Changes

```typescript
import { Agent, InvokeModelStage, ExecuteToolStage, AgentStreamStage, createStage } from '@strands-agents/sdk'
import type { MiddlewareStage, MiddlewareHandler, MiddlewareNext } from '@strands-agents/sdk'

const agent = new Agent({ model, tools })

// Register middleware for any built-in stage
agent.addMiddleware(InvokeModelStage, async function* (context, next) {
  // pre-processing: inspect or transform context
  const modified = { ...context, messages: sanitize(context.messages) }
  // call next layer (or don't, to short-circuit)
  const result = yield* next(modified)
  // post-processing: inspect or transform result
  return result
})
```

Three stages ship with the SDK:

| Stage              | Wraps                                                     | Context fields                                            |
| ------------------ | --------------------------------------------------------- | --------------------------------------------------------- |
| `InvokeModelStage` | Model call (between Before/AfterModelCallEvent)           | messages, systemPrompt, toolSpecs, toolChoice, modelState |
| `ExecuteToolStage` | Single tool execution (between Before/AfterToolCallEvent) | tool, toolUse (name, id, input)                           |
| `AgentStreamStage` | Full `agent.stream()` output                              | args, options                                             |

Handlers are async generators and simple pass-through is `return yield* next(context)`. Manual iteration of `next()` allows real-time event filtering or injection while not calling `next` at all short-circuits the operation.

## Plugin Examples

```typescript
class ToolResultCache implements Plugin {
  name = 'tool-result-cache'
  private readonly _cache = new Map<string, ToolResultBlock>()

    const cache = this._cache
    agent.addMiddleware(ExecuteToolStage, async function* (context, next) {
      const key = `${context.toolUse.name}:${JSON.stringify(context.toolUse.input)}`
      const cached = cache.get(key)
      if (cached) return { result: new ToolResultBlock({ toolUseId: context.toolUse.toolUseId, status: cached.status, content: cached.content }) }
      const result = yield* next(context)
      cache.set(key, result.result)
      return result
    })
  }
}
```

```typescript
class RetryOnThrottle implements Plugin {
  name = 'retry-on-throttle'
    agent.addMiddleware(InvokeModelStage, async function* (context, next) {
      for (let attempt = 0; attempt < 3; attempt++) {
        try { return yield* next(context) }
        catch (e) {
          if (!(e as Error).message.includes('ThrottlingException') || attempt === 2) throw e
        }
      }
      throw new Error('exhausted retries')
    })
  }
}
```

```typescript
class SystemPromptInjector implements Plugin {
  name = 'system-prompt-injector'
  constructor(private readonly _suffix: string) {}

    const suffix = this._suffix
    agent.addMiddleware(InvokeModelStage.Input, async (context) => ({
      ...context,
      systemPrompt: `${context.systemPrompt ?? ''}\n\n${suffix}`.trim(),
    }))
  }
}

// Usage: inject safety guidelines into every model call
const agent = new Agent({
  model,
  systemPrompt: 'You are a helpful assistant.',
  plugins: [new SystemPromptInjector('Always cite your sources.')],
})
```

## Middleware Interrupts

Middleware contexts expose `interrupt()` for human-in-the-loop gating:

```typescript
agent.addMiddleware(ExecuteToolStage, async function* (context, next) {
  const { response } = context.interrupt<string>({ name: 'approve', reason: 'Confirm?' })
  if (response !== 'yes') return { result: new ToolResultBlock({ ... }) }
  return yield* next(context)
})
```

Returns `MiddlewareInterruptResult<T>` (wrapper) — allows non-breaking additions (cached data, metadata) as the interrupt system evolves.

## Phase Sub-Stages (Input / Around / Output)

Each stage exposes three phases with a fixed execution order in order to solve the middleware ordering problem without explicit priority numbers:

- `Input`: Modifies the input of the phase as a pure function - take in the original input, return the modified input
- `Output`: Modifies the output of the phase as a pure function - take in the original output, return the modified output
- `Around`: Wrape the entire invocation of the stage - this is more akin to traditional middleware from express or other frameworks

```typescript
// Input: transform context before execution (plain async function)
agent.addMiddleware(InvokeModelStage.Input, async (context) => ({
  ...context,
  systemPrompt: injectToSystemPrompt(context),
}))

// Output: transform result after execution (plain async function)
agent.addMiddleware(InvokeModelStage.Output, async (result) => {
  log(`stopReason=${result.result.stopReason}`)
  return result
})

// Around: full async generator wrap (same as just providing `InvokeModelStage`)
agent.addMiddleware(InvokeModelStage.Around, async function* (context, next) {
  return yield* next(context)
})
```

**Execution order is fixed:** all Input → all Output → all Around → terminal. A retry plugin on `.Around` always retries the full chain including Input transforms. A response logger on `.Output` never needs to coordinate registration order with a system prompt injector on `.Input`.

## Key Decisions

- **First registered = outermost**: follows existing middleware conventions (from Express/Koa)
- **Phase ordering is fixed**: Input/Output/Around reduces the need for explicit priority management (for now - we'll probably need to add it later)
- **Hooks fire unconditionally before/after middleware**: existing behavior is unchanged
- **Result wrapper types** (`InvokeModelResult`, etc.): allows future extension of middleware return values, including returning values "up the chain"
- **`readonly` arrays** on `InvokeModelContext`: enforce immutable-context pattern at the type level

## Checklist
- [x] Tests prove the fix is effective / feature works
- [x] No new warnings
- [ ] Documentation update (pending)

----

### DIFF
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -26,7 +26,7 @@ import {
 import type { JSONValue } from '../types/json.js'
 import { McpClient } from '../mcp.js'
 import { isValidToolName, type Tool, type ToolContext } from '../tools/tool.js'
-import type { ToolChoice } from '../tools/types.js'
+import type { ToolChoice, ToolSpec } from '../tools/types.js'
 import { systemPromptFromData } from '../types/messages.js'
 import { normalizeError, ConcurrentInvocationError, StructuredOutputError } from '../errors.js'
 import { Model } from '../models/model.js'
@@ -48,6 +48,25 @@ import { ConversationManager } from '../conversation-manager/conversation-manage
 import { ContextOffloader } from '../vended-plugins/context-offloader/plugin.js'
 import { InMemoryStorage } from '../vended-plugins/context-offloader/storage.js'
 import { HookRegistryImplementation } from '../hooks/registry.js'
+import { MiddlewareRegistry, InvokeModelStage, ExecuteToolStage, AgentStreamStage } from '../middleware/index.js'
+import type {
+  MiddlewareStage,
+  MiddlewareHandler,
+  MiddlewareInputHandler,
+  MiddlewareOutputHandler,
+  MiddlewareInputPhase,
+  MiddlewareOutputPhase,
+  MiddlewarePhaseKind,
+} from '../middleware/index.js'
+import type {
+  InvokeModelContext,
+  InvokeModelResult,
+  ExecuteToolContext,
+  ExecuteToolResult,
+  AgentStreamContext,
+  AgentStreamResult,
+  MiddlewareInterruptResult,
+} from '../middleware/index.js'
 import type { HookableEventConstructor, HookCallback, HookCallbackOptions, HookCleanup } from '../hooks/types.js'
 import {
   InitializedEvent,
@@ -69,6 +88,7 @@ import {
   ToolStreamUpdateEvent,
   InterruptEvent,
   type ModelStopData,
+  type ToolUseData,
 } from '../hooks/events.js'
 import { StructuredOutputTool, STRUCTURED_OUTPUT_TOOL_NAME } from '../tools/structured-output-tool.js'
 import { AgentAsTool } from './agent-as-tool.js'
@@ -88,7 +108,7 @@ import { CancelledError } from '../errors.js'
 import { DefaultModelRetryStrategy } from '../retry/default-model-retry-strategy.js'
 import type { RetryStrategy } from '../retry/retry-strategy.js'
 import { warnOnDuplicateRetryStrategyTypes } from '../retry/retry-strategy.js'
-import { InterruptError, InterruptState, interruptFromAgent } from '../interrupt.js'
+import { Interrupt, InterruptError, InterruptState, interruptFromAgent } from '../interrupt.js'
 import type { InterruptParams } from '../types/interrupt.js'
 import { isInterruptResponseContent, type InterruptResponseContent } from '../types/interrupt.js'
 import { takeSnapshot as takeSnapshotInternal, loadSnapshot as loadSnapshotInternal } from './snapshot.js'
@@ -297,6 +317,36 @@ const DEFAULT_AGENT_ID = 'agent'
 type ToolsExecutionResult = { message: Message; afterToolsEvent: AfterToolsEvent }
 
 /**
+ * Creates a non-mutating interrupt function for middleware contexts.
+ * Reads existing responses from state (resume case) but never writes to it.
+ * On first run (no response), throws InterruptError with a locally-created Interrupt.
+ *
+ * @param interruptState - The agent's interrupt state (read-only access)
+ * @param idPrefix - Prefix for the interrupt ID (e.g., 'middleware:agentStream')
+ */
+function createMiddlewareInterrupt(
+  interruptState: InterruptState,
+  idPrefix: string
+): <T = JSONValue>(params: InterruptParams) => MiddlewareInterruptResult<T> {
+  return <T = JSONValue>(params: InterruptParams): MiddlewareInterruptResult<T> => {
+    const interruptId = `${idPrefix}:${params.name}`
+    const existing = interruptState.interrupts[interruptId]
+    if (existing?.response !== undefined) {
+      return { response: existing.response as T }
+    }
+    if (params.response !== undefined) {
+      return { response: params.response as T }
+    }
+    const interrupt = new Interrupt({
+      id: interruptId,
+      name: params.name,
+      ...(params.reason !== undefined && { reason: params.reason }),
+    })
+    throw new InterruptError(interrupt)
+  }
+}
+
+/**
  * Orchestrates the interaction between a model, a set of tools, and MCP clients.
  * The Agent is responsible for managing the lifecycle of tools and clients
  * and invoking the core decision-making loop.
@@ -357,6 +407,7 @@ export class Agent implements LocalAgent, InvokableAgent {
   public readonly memoryManager?: MemoryManager | undefined
 
   private readonly _hooksRegistry: HookRegistryImplementation
+  private readonly _middlewareRegistry: MiddlewareRegistry
   private readonly _pluginRegistry: PluginRegistry
   private readonly _interventionRegistry: InterventionRegistry
   private _toolRegistry: ToolRegistry
@@ -424,6 +475,9 @@ export class Agent implements LocalAgent, InvokableAgent {
     this._hooksRegistry = new HookRegistryImplementation()
 
     this._interventionRegistry = new InterventionRegistry(config?.interventions ?? [], this._hooksRegistry)
+
+    // Initialize middleware registry
+    this._middlewareRegistry = new MiddlewareRegistry()
 
     // `undefined` (omitted) → install the default; `null`/`[]` → explicit opt-out.
     const retryStrategies: RetryStrategy[] =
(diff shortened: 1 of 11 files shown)


strands-ts/src/agent/agent.ts
```diff
@@ -2101,6 +2085,144 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+      // Execute tool within the tool span context
+      const toolContext: ToolContext = {
+        toolUse: {
+          name: toolUse.name,
+          toolUseId: toolUse.toolUseId,
+          input: toolUse.input,
+        },
+        agent: this,
+        invocationState,
+        interrupt: <T = JSONValue>(params: InterruptParams): T => {
```
opieter-aws, line 2430: There's 3 variations of this interrupt function in this file. Can we consolidate into 1 helper?

strands-ts/src/agent/agent.ts
```diff
@@ -2101,6 +2085,144 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+          id: interruptId,
+          name: params.name,
+          ...(params.reason !== undefined && { reason: params.reason }),
+        })
+        throw new InterruptError(interrupt)
+      },
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    const self = this
```
opieter-aws, line 2385: This is a repeated pattern. Is there a workaround?

strands-ts/src/agent/agent.ts
opieter-aws, line 1727: Do we need to worry about middleware mutating what the model actually sees, whereas these operations occur before those mutations happen? Should we at least document this if this is accepted?

strands-ts/src/agent/agent.ts
```diff
@@ -730,92 +681,164 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+          // Transfer interrupts from the error to agent state (deferred write).
+          // Middleware-thrown interrupts bypass _stream's own catch, so we mirror
+          // upstream's behavior here.
+          for (const interrupt of error.interrupts) {
+            this._interruptState.getOrCreateInterrupt(interrupt.id, interrupt.name, interrupt.reason)
           }
+          this._interruptState.activate()
+          return new AgentResult({
+            stopReason: 'interrupt',
+            lastMessage: new Message({ role: 'assistant', content: [] }),
```
opieter-aws, line 998: For a tool/hook interrupt, content is set differently. Why don't we call `createInterruptResult(options?.invocationState ?? {})` here to avoid drift?

strands-ts/src/agent/agent.ts
```diff
@@ -730,92 +681,164 @@ export class Agent implements LocalAgent, InvokableAgent {
...
-        const streamGenerator = this._stream(currentArgs, options)
-        let caughtError: Error | undefined
-        let lastAfterInvocation: AfterInvocationEvent | undefined
-        let iterationResult: IteratorResult<AgentStreamEvent, AgentResult>
-        try {
-          iterationResult = await streamGenerator.next()
+      // Process interrupt responses before middleware runs so context.interrupt() can find them
+      const interruptResponses = this._extractInterruptResponses(args)
+      if (interruptResponses.length > 0) {
+        this._interruptState.resume(interruptResponses)
```
opieter-aws, line 956: `resume()` now runs only here, once, against the top-level args. When a hook sets `AfterInvocationEvent.resume` to interrupt-response blocks, the resume-loop re-enters `_stream` with them but never re-applies `resume()` (line 980 only reads them to gate a check), so they’re silently dropped. Is this intentional?

strands-ts/src/agent/agent.ts
```diff
@@ -730,92 +681,164 @@ export class Agent implements LocalAgent, InvokableAgent {
...
-            drainResult = await streamGenerator.next()
+        )
+        return result
+      } catch (error) {
+        if (error instanceof InterruptError) {
+          // Transfer interrupts from the error to agent state (deferred write).
+          // Middleware-thrown interrupts bypass _stream's own catch, so we mirror
+          // upstream's behavior here.
+          for (const interrupt of error.interrupts) {
+            this._interruptState.getOrCreateInterrupt(interrupt.id, interrupt.name, interrupt.reason)
```
opieter-aws, line 990: Unlike `_stream` catch, this path registers the interrupts but never yields `InterruptEvents`, so stream consumers see interrupt events for tool/hook interrupts but not for `AgentStreamStage` ones. A single shared helper should fix this and the comment below

poshinchen:
Iterate on this comment: https://github.com/strands-agents/sdk-typescript/pull/1068/changes#r3343982119

Can we just create a custom span?

strands-ts/src/agent/agent.ts
zastrowm: Do you mean hooks running first, or the projectedInputTokens?

> projectedInputTokens

I think we need to add documentation and *possibly* [in the future] expose a way to change this later on.  

This is technically a problem with hooks right now, right?

opieter-aws: You're right that we have the same problem with hooks today too, yeah

I meant `projectedInputTokens` initially, but anything observed or computed at the `BeforeModelCall` boundary reflects pre-middleware state and can diverge from what the model receives

Documenting sounds acceptable to me!

strands-ts/src/agent/agent.ts
```diff
@@ -2101,6 +2187,127 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+    return yield* this._middlewareRegistry.invoke(
+      ExecuteToolStage,
+      context,
+      async function* (ctx: ExecuteToolContext): AsyncGenerator<AgentStreamEvent, ExecuteToolResult, undefined> {
+        return yield* self._executeToolCore(ctx.tool, ctx.toolUse, ctx.invocationState)
+      }
+    )
+  }
+
+  private async *_executeToolCore(
```
zastrowm, line 2397: @poshinchen said 

Iterate on this comment: [...](https://github.com/strands-agents/sdk-typescript/pull/1068/changes#r3343982119)

Can we just create a custom span?

> **Issue**: Telemetry asymmetry between `InvokeModelStage` and `ExecuteToolStage` when middleware short-circuits.
>
> For model calls, the span/metrics wrap the middleware chain — `startModelInvokeSpan` is called at line 1548 (before middleware runs) and `endModelInvokeSpan` at line 1563 (after middleware returns). So a cache-hit middleware still appears in traces.
>
> For tool calls, the span/metrics live inside `_executeToolCore` (started here at line 2212). When `ExecuteToolStage` middleware short-circuits (e.g., returns a mock result), `_executeToolCore` never runs, so:
>
> - No tool span is created
> - No `endToolCallSpan` is called  
> - `_meter.endToolCall()` never fires
>
> This means tool cache hits, mocked results, and other short-circuit patterns are invisible to telemetry consumers.
>
> **Suggestion**: Consider moving the tool span start/end outside `_executeToolCore` to wrap the middleware call (matching the model pattern). Something like:
>
> ```typescript
> private async *_executeToolWithMiddleware(...) {
>   const toolSpan = this._tracer.startToolCallSpan({ tool: toolUse })
>   const toolStartTime = Date.now()
>   try {
>     const result = yield* this._middlewareRegistry.invoke(ExecuteToolStage, context, terminal)
>     this._tracer.endToolCallSpan(toolSpan, { toolResult: result.result })
>     this._meter.endToolCall({ tool: toolUse, duration: Date.now() - toolStartTime, success: result.result.status === 'success' })
>     return result
>   } catch (error) {
>     this._tracer.endToolCallSpan(toolSpan, { error })
>     throw error
>   }
> }
> ```
>
> This ensures short-circuited tool calls still generate telemetry. The model stage already follows this pattern.

zastrowm: I think I'd like to address this as an immediate follow-up. 

What are your thoughts on:
 - Every middleware should have a custom span
 - Native Spans (model call, tool call) should always occur inside of the custom span

Is it weird or normal/expected to have the native spans inside of custom spans? Let's discuss offline to refine this a bit

strands-ts/src/agent/agent.ts
zastrowm: No workaround.

Added an explaining comment; the helper is just boilerplate IMHO

strands-ts/src/agent/agent.ts
```diff
@@ -459,6 +505,20 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+   * Middleware wraps stage execution and can intercept, transform, or short-circuit operations.
+   *
+   * @param stage - The stage token identifying the interception point
+   * @param handler - The middleware handler function (async generator)
+   */
+  addMiddleware<TContext, TEvent, TResult>(
+    stage: Stage<TContext, TEvent, TResult>,
+    handler: MiddlewareHandler<TContext, TEvent, TResult>
+  ): void {
+    this._middlewareRegistry.add(stage, handler)
```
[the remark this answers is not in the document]
zastrowm: Adding

zastrowm:
Migrating to https://github.com/strands-agents/harness-sdk/pull/2681 to cut down on necessary history/back & forth

PR #2619, opened 2026-06-03
fix: bump vitest to 4.1.8 and fix bedrock role type error

zastrowm:
## Summary

- Bumps vitest and @vitest/* packages to ^4.1.8 in `strands-ts` and `strands-wasm` to fix the npm security audit CI check
- Fixes pre-existing TypeScript build error in `bedrock.ts` where Bedrock SDK's role type includes `"system"` but our internal `Role` type is `"user" | "assistant"` — now throws at runtime on unexpected roles instead of silently casting

## Test plan

- [x] `npm audit --audit-level=high` passes with 0 vulnerabilities
- [x] All 2852 strands-ts tests pass
- [x] TypeScript build passes cleanly

### DIFF
--- a/.github/workflows/python-pypi-publish-on-release.yml
+++ b/.github/workflows/python-pypi-publish-on-release.yml
@@ -55,11 +55,6 @@ jobs:
             exit 1
         fi
 
-    - name: Copy root LICENSE and NOTICE
-      run: |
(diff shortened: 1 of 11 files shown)


zastrowm:
Closing — the vitest bump and bedrock type fix both already landed on main. PR #2616's audit failure was just a stale base.

PR #2625, opened 2026-06-04
docs: recommend bedrock_mantle_config for OpenAI-on-Bedrock (Mantle)

agent-of-mkmeral:
## What

Updates the **Amazon Bedrock (Mantle)** section of the OpenAI Responses model-provider page to lead with the first-class `bedrock_mantle_config` (Python) / `bedrockMantleConfig` (TypeScript) parameter, instead of hand-rolling `base_url` + a static `api_key`.

## Why

The SDK ships a dedicated config for OpenAI-on-Bedrock — [`BedrockMantleConfig`](https://github.com/strands-agents/harness-sdk/blob/main/strands-py/src/strands/models/_openai_bedrock.py) (Python) and [`bedrockMantleConfig`](https://github.com/strands-agents/harness-sdk/blob/main/strands-ts/src/models/openai/mantle.ts) (TS) — but the docs only showed the manual `base_url` approach. That has two downsides users hit in practice:

1. **The static bearer token doesn't refresh.** A hand-passed `api_key` is minted once; long-running agents break when it expires. `bedrock_mantle_config` mints a fresh token **per request** via `aws-bedrock-token-generator`, using the standard AWS credential chain.
2. **It hardcodes the endpoint URL.** The config derives the regional endpoint for you, so there's no hand-coded `base_url` to drift from the SDK's own template.

This came up while reviewing a community sample ([strands-agents/samples#270](https://github.com/strands-agents/samples/pull/270)) that copied the manual pattern straight from these docs — so fixing the docs fixes the root cause.

## Changes

- **Python tab:** lead with `bedrock_mantle_config={"region": "..."}`, add a table documenting the accepted keys (`region`, `boto_session`, `credentials_provider`, `expiry`), and note the `aws-bedrock-token-generator` requirement (bundled in the `strands-agents[openai]` extra).
- **TypeScript tab:** switch the snippet to `bedrockMantleConfig: { region: '...' }`, add a table for `region` / `credentials` / `expiresInSeconds`, and note the `@aws/bedrock-token-generator` requirement.
- Keep the manual **Bedrock API key** path as a collapsible *Alternative* (with a note that a static key must outlive the agent's run).
- Fix the Mantle docs link to the current `inference-openai.html` page (matches the URL used in the SDK source).

The accepted-keys tables and the manual-fallback constraints are taken directly from the in-tree docstrings/types (`BedrockMantleConfig` in both `strands-py` and `strands-ts`) and the integ test `tests_integ/models/test_model_mantle.py`.

## Validation

- `npx prettier --check` passes on the edited `.ts` snippet.
- `npm run typecheck:snippets`: the edited `openai-responses.ts` shows **no** error on the new `bedrockMantleConfig` property (only the pre-existing `Cannot find module '@strands-agents/sdk'` import errors that affect every snippet file in an un-built checkout), confirming the new property typechecks against the real `OpenAIModelOptions`.
- Snippet rendered via the existing `--8<--` include mechanism; no nav/frontmatter changes needed.

---
🤖 Opened by the agent of @mkmeral. Happy to adjust wording or scope.

### DIFF
--- a/site/src/content/docs/user-guide/concepts/model-providers/openai-responses.mdx
+++ b/site/src/content/docs/user-guide/concepts/model-providers/openai-responses.mdx
@@ -60,7 +60,9 @@ In TypeScript, both the Responses API and Chat Completions API are available thr
 
 ### Amazon Bedrock (Mantle)
 
-The Responses provider can connect to [Amazon Bedrock's OpenAI-compatible endpoints](https://docs.aws.amazon.com/bedrock/latest/userguide/bedrock-mantle.html) powered by Mantle. Authenticate with a [Bedrock API key](https://docs.aws.amazon.com/bedrock/latest/userguide/api-key-management.html) and point the client at your region's Mantle endpoint.
+The Responses provider can connect to [Amazon Bedrock's OpenAI-compatible endpoints](https://docs.aws.amazon.com/bedrock/latest/userguide/inference-openai.html) powered by Mantle.
+
+The recommended way to connect is the first-class `bedrock_mantle_config` parameter. It resolves the regional Mantle endpoint for you and **mints a fresh bearer token on every request** using the standard AWS credential chain — so long-running agents keep working past the bearer token's maximum lifetime, with no manual token handling.
 
 <Tabs>
 <Tab label="Python">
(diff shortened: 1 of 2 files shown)


site/src/content/docs/user-guide/concepts/model-providers/openai-responses.mdx
```diff
@@ -69,6 +71,33 @@ The Responses provider can connect to [Amazon Bedrock's OpenAI-compatible endpoi
...
+    model_id="openai.gpt-oss-120b",
+    bedrock_mantle_config={"region": "us-east-1"},
+)
+
+agent = Agent(model=model)
+response = agent("What is 2+2?")
+print(response)
+```
+
+`bedrock_mantle_config` accepts the following keys:
```
mkmeral, line 84: you don't need to overexplain. just give an example, mention the config, and refer to api-reference. each key is explained there anyway

SummerWJLiu:
⚠️ Heads-up before this lands: `bedrock_mantle_config` currently can't reach the `gpt-5.4` / `gpt-5.5` models, so recommending it as the primary pattern would steer readers straight into a 400.

Per #2629, `bedrock_mantle_config` hardcodes the base URL to `/v1`, but `gpt-5.4` / `gpt-5.5` are served only on the `/openai/v1` path (documented in their model cards — the GPT-5.5 card explicitly notes it differs from the `/v1` path other models use). So a reader copying the `bedrock_mantle_config` example with `model_id="openai.gpt-5.5"` hits:

```
400 validation_error: The model 'openai.gpt-5.5' does not support the '/v1/responses' API
```

Suggestion: either hold this until #2629 lands, or add a note that `bedrock_mantle_config` currently targets `/v1` models and that `gpt-5.4` / `gpt-5.5` require explicit `client_args` with `base_url=".../openai/v1"` for now.

lizradway:
Closing until mantle actually is the default primary pattern

PR #2627, opened 2026-06-04
feat: add test-infra/ CDK stack for long-running integ test resources

notowen333:
## Description

Adds a `test-infra/` CDK TypeScript stack that provisions shared AWS resources a small subset of integration tests depend on — specifically tests exercising Bedrock knowledge base ingestion (S3 + CUSTOM paths) and the SSH sandbox. Most integration tests do not require these resources and continue to work without this stack deployed.

The reason that we introduce the SSM system is so a developer can deploy the stacks themselves and the integ tests will resolve to their deployed infra. This is also useful for development on our side because we're not reliant on env vars in a GH action runner.

There is also a path to migration for our existing hand-maintained IAM role in this PR. The idea is we would copy the existing one and then port over.

Tests resolve deploy-time identifiers (KB id, data source ids, instance id, etc.) from SSM Parameter Store at runtime via a shared `ssmParameterPath()` helper, so no hardcoded IDs exist and any deployed instance of this stack works.

## Related Issues

N/A — establishes the foundational pattern for managed test infrastructure.

## Type of Change

New feature

## Motivation

Integration tests for infrastructure-dependent features (RAG knowledge bases, SSH sandbox) previously relied on handwritten AWS resources and a manually-maintained IAM role. This made it impossible for community contributors to run these tests against their own accounts and created drift between what CI used and what was documented.

This stack codifies the infrastructure as CDK, makes it independently deployable per-feature, separates the internal legacy policy behind an explicit opt-in flag, and publishes all resource identifiers to SSM so tests self-configure at runtime without hardcoded values.

## Public API Changes

No SDK public API changes. Introduces a new `test-infra/` directory with:

- `lib/constants.ts` — exports `TestFeature` type, `VALID_TEST_FEATURES`, and `ssmParameterPath()` for use by both the CDK constructs and integration test code
- `TestFeatureConstruct` base class — pattern for adding new infrastructure features

## Key design decisions

- **Feature toggling** via `-c testFeatures=bedrock-knowledge-base,ssh-ec2` — deploy only what you need
- **SSM Parameter Store** for resource discovery — O(1) lookup, no name-scanning, scales under concurrent CI runners
- **`TestFeatureConstruct` base class** — each feature declares its name once; SSM paths and parameter-read grants are automatic
- **Internal mode** (`STRANDS_TEST_INFRA_INTERNAL=true`) gates the legacy broad policy + GitHub OIDC trust; community deploys get only the scoped grants from deployed resources
- **No hardcoded account IDs** in source — account-pinned ARNs derive from `this.account` at synth time

## Testing

- 15 CDK assertion tests covering both features, feature toggling, IAM trust policies, SSM parameter paths, and resource properties
- Successfully deployed and verified against a live AWS account (KB created, data sources active, EC2 reachable via SSM, all SSM parameters resolvable)

## Checklist

----


test-infra/bin/test-infra.ts
```diff
@@ -0,0 +1,40 @@
...
+        return trimmed as TestFeature;
+      })
+    : undefined;
+
+// Internal Strands deploys set this to attach the legacy base policy to the
+// test role; community deploys leave it unset.
+const internal = process.env.STRANDS_TEST_INFRA_INTERNAL === 'true';
+
+const account = process.env.STRANDS_TEST_INFRA_DEPLOYMENT_ACCOUNT ?? process.env.CDK_DEFAULT_ACCOUNT;
+const region = process.env.STRANDS_TEST_INFRA_DEPLOYMENT_REGION ?? process.env.CDK_DEFAULT_REGION;
```
opieter-aws: nit: use the default env vars `AWS_ACCOUNT_ID` and `AWS_REGION` as fallback (https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-envvars.html)

test-infra/lib/constructs/bedrock-knowledge-base-test-resources.ts
```diff
@@ -0,0 +1,263 @@
...
+
+  /** Bucket backing the S3 ingestion path; tests stage objects here. */
+  public readonly sourceBucket: s3.Bucket;
+
+  constructor(scope: Construct, id: string, props: BedrockKnowledgeBaseTestResourcesProps = {}) {
+    super(scope, id);
+
+    const model = bedrock.FoundationModel.fromFoundationModelId(this, 'EmbeddingModel', EMBEDDING_MODEL);
+    this.sourceBucket = new s3.Bucket(this, 'Source', {
+      removalPolicy: cdk.RemovalPolicy.RETAIN,
```
opieter-aws: Are you sure we want to retain? If it's a testing resource, it may be easier to delete so all infra is cleaned up and we avoid naming collisions in case of a re-deploy?

test-infra/lib/constructs/bedrock-knowledge-base-test-resources.ts
```diff
@@ -0,0 +1,263 @@
...
+    });
+    const { vectorBucket, vectorIndex } = this.createVectorIndex();
+    const serviceRole = this.createServiceRole(model, this.sourceBucket, vectorBucket, vectorIndex);
+
+    this.knowledgeBase = new bedrock.CfnKnowledgeBase(this, 'KnowledgeBase', {
+      name: `strands-test-infra-kb`,
```
opieter-aws: Ditto for re-deploy, hardcoded name

test-infra/lib/constructs/bedrock-knowledge-base-test-resources.ts
```diff
@@ -0,0 +1,263 @@
...
+        },
+      },
+    });
+
+    // A CUSTOM data source has no connection config: documents are pushed in
+    // directly through the ingestion API rather than crawled from a source.
+    this.customDataSource = new bedrock.CfnDataSource(this, 'CustomDataSource', {
+      name: 'CustomDataSource',
+      knowledgeBaseId: this.knowledgeBase.attrKnowledgeBaseId,
+      dataDeletionPolicy: 'RETAIN',
```
opieter-aws: ditto

test-infra/lib/constructs/bedrock-knowledge-base-test-resources.ts
```diff
@@ -0,0 +1,263 @@
...
+    const stack = cdk.Stack.of(this);
+    const role = new iam.Role(this, 'ServiceRole', {
+      assumedBy: new iam.ServicePrincipal('bedrock.amazonaws.com', {
+        conditions: {
+          StringEquals: { 'aws:SourceAccount': stack.account },
+          ArnLike: {
+            'aws:SourceArn': stack.formatArn({
+              service: 'bedrock',
+              resource: 'knowledge-base',
+              resourceName: '*',
```
opieter-aws, line 232: can we not scope down the resource name here?

test-infra/lib/constructs/integ-test-role.ts
```diff
@@ -0,0 +1,216 @@
...
+            },
+            StringLike: {
+              'token.actions.githubusercontent.com:sub': [
+                'repo:strands-agents/sdk-python:*',
+                'repo:strands-agents/private-sdk-python-staging:*',
+                'repo:strands-agents/tools:*',
+                'repo:strands-agents/private-tools-staging:*',
+                'repo:strands-agents/agent-builder:*',
+                'repo:strands-agents/evals:*',
+                'repo:strands-agents/sdk-typescript:*',
```
opieter-aws: do we not only need `sdk-harness` here now?

test-infra/lib/constructs/integ-test-role.ts
```diff
@@ -0,0 +1,216 @@
...
+          'arn:aws:bedrock:*:*:async-invoke/*',
+          'arn:aws:bedrock:*:*:inference-profile/us.anthropic.claude-haiku-4-5-20251001-v1:0',
+          'arn:aws:bedrock:*:*:inference-profile/global.anthropic.claude-sonnet-4-6',
+          'arn:aws:bedrock:*::foundation-model/anthropic.claude-sonnet-4-6',
+        ],
+      }),
+    );
+
+    this.role.addToPolicy(
+      new iam.PolicyStatement({
```
opieter-aws, line 98: is there a managed policy we could replace this with? BedrockFullAccess?

test-infra/lib/constructs/integ-test-role.ts
```diff
@@ -0,0 +1,216 @@
...
+          'aoss:CreateSecurityPolicy',
+          'aoss:CreateCollection',
+          'aoss:ListCollections',
+          'aoss:CreateAccessPolicy',
+          'aoss:BatchGetCollection',
+          'aoss:APIAccessAll',
+          'aoss:UpdateCollection',
+          'aoss:DeleteCollection',
+          'aoss:CreateLifecyclePolicy',
+          'aoss:UpdateAccessPolicy',
```
opieter-aws, line 159: Do we need all these?

test-infra/lib/constructs/integ-test-role.ts
```diff
@@ -0,0 +1,216 @@
...
+          's3:DeleteObject',
+          's3:GetObject',
+        ],
+        resources: ['arn:aws:s3:::nova-reels-e2e-test-*', 'arn:aws:s3:::strands-integ-tests-resources-*'],
+      }),
+    );
+
+    this.role.addToPolicy(
+      new iam.PolicyStatement({
+        actions: ['iam:GetRole', 'iam:CreateRole', 'iam:PutRolePolicy', 'iam:PassRole'],
```
opieter-aws, line 196: What do we need CreateRole and PutRolePolicy for?

test-infra/lib/constants.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Independently selectable test features. The stack provisions a feature only
+ * when it is selected, so a deployment can stand up just the slice a given test
+ * suite needs. `all` selects every feature. Add a member here when introducing
+ * a new feature.
+ */
+export type TestFeature = 'all' | 'bedrock-knowledge-base' | 'ssh-ec2';
+
+export const VALID_TEST_FEATURES: readonly TestFeature[] = ['all', 'bedrock-knowledge-base', 'ssh-ec2'];
```
opieter-aws: Can we consolidate?

test-infra/test/test-infra.test.ts
```diff
@@ -0,0 +1,200 @@
...
+  const stack = new StrandsTestInfraStack(app, 'TestStack', {
+    env: { account: '123456789012', region: 'us-east-1' },
+    ...props,
+  });
+  return Template.fromStack(stack);
+}
+
+// --- Knowledge Base ---
+
+test('creates an S3 vectors index matching Titan v2 (1024 dims, cosine, float32)', () => {
```
opieter-aws, line 16: Can we make this a conditional test? So that if infra is not set up, skip all tests?

test-infra/cdk.json
```diff
@@ -0,0 +1,107 @@
...
+    "@aws-cdk/core:aspectPrioritiesMutating": true,
+    "@aws-cdk/core:cfnIncludeRejectComplexResourceUpdateCreatePolicyIntrinsics": true,
+    "@aws-cdk/core:checkSecretUsage": true,
+    "@aws-cdk/core:enableAdditionalMetadataCollection": true,
+    "@aws-cdk/core:enablePartitionLiterals": true,
+    "@aws-cdk/core:explicitStackTags": true,
+    "@aws-cdk/core:includePrefixInUniqueNameGeneration": true,
+    "@aws-cdk/core:target-partitions": [
+      "aws",
+      "aws-cn"
```
opieter-aws: `aws-cn` won't work, partitions are hardcoded in the cdk code

test-infra/test/test-infra.test.ts
opieter-aws: Never mind, didn't realize these are testing the testing setup

test-infra/lib/constructs/bedrock-knowledge-base-test-resources.ts
notowen333: Yeah agree will try to not retain anything

notowen333:
**Re: scoping `bedrock:ListFoundationModels` / `bedrock:ListCustomModels`** — these are list actions that don't support resource-level permissions. `*` is the only valid resource value. This matches the [Bedrock KB permissions docs](https://docs.aws.amazon.com/bedrock/latest/userguide/kb-permissions.html).

notowen333:
**Re: `resourceName: '*'` in trust condition** — the KB requires `roleArn` at construction, so the role must be created before the KB. That means the KB's ID (service-generated) isn't available when the trust policy is written. This is the standard pattern from the AWS docs for this exact reason.

test-infra/README.md
```diff
@@ -0,0 +1,132 @@
...
+npm test        # 15 assertions covering all features, toggling, IAM
+```
+
+### Destroy
+
+```sh
+npx cdk destroy
+```
+
+Note: the S3 source bucket has `RemovalPolicy.RETAIN` — it won't be deleted automatically. Remove it manually if needed.
```
opieter-aws: can we change this to `DESTROY`?

test-infra/bin/test-infra.ts
notowen333: Done.

test-infra/lib/constructs/bedrock-knowledge-base-test-resources.ts
notowen333: Done — DESTROY everywhere.

test-infra/lib/constructs/bedrock-knowledge-base-test-resources.ts
notowen333: Done — `${stackName}-kb` now.

test-infra/lib/constructs/bedrock-knowledge-base-test-resources.ts
notowen333: Done — `dataDeletionPolicy: DELETE`.

test-infra/lib/constructs/bedrock-knowledge-base-test-resources.ts
notowen333: Can't scope — role must exist before KB (KB requires roleArn), so KB ID isn't available at trust creation time. Standard AWS pattern.

test-infra/lib/constructs/integ-test-role.ts
notowen333: Done — added harness-sdk, removed sdk-python and sdk-typescript.

test-infra/lib/constructs/integ-test-role.ts
notowen333: Keeping inline for now — not scoping down the legacy policy in this pass.

test-infra/lib/constructs/integ-test-role.ts
notowen333: Keeping — other integ tests still exercise AOSS.

test-infra/lib/constructs/integ-test-role.ts
notowen333: Memory integ tests dynamically create/configure that role at test time.

test-infra/lib/constants.ts
notowen333: Done — single `as const` array, type derived from it.

test-infra/cdk.json
notowen333: Done — removed.

test-infra/README.md
notowen333: Done.

opieter-aws (approved):

PR #2628, opened 2026-06-04
fix(a2a): introduce agent factory for isolating agent context from different callers

JackYPCOnline:
## Description

The A2A executor reused a single `Agent` instance across all requests, so its
conversation state (messages, agent state, conversation-manager state, interrupt
state, and model state) was shared by every caller regardless of A2A `context_id`.
This is a cross-session exposure: callers sending requests to the same server could
read each other's messages, plant persistent instructions, and grow history without
bound.

This PR isolates conversation state per A2A `context_id`, following the
[A2A specification](https://a2a-protocol.org/latest/specification/#341-context-identifier-semantics).
Two modes provide the isolation, so existing deployments are fixed on upgrade and
new code has a cleaner path:

- **Single agent (`agent=...`)** — backwards compatible and now isolated. Each
  `context_id`'s state is snapshotted on/off the shared agent under a lock, so
  callers no longer share history. Requests are serialized (a single `Agent` cannot
  run concurrently). This path is **deprecated** in favor of the factory.
- **Agent factory (`agent_factory=...`, recommended)** — a callable
  `(context_id) -> Agent` that builds a dedicated `Agent` per context. Per-context
  agents are isolated by construction, run under per-context locks (different contexts
  execute concurrently), and are the place to wire per-context concerns such as a
  context-scoped `session_manager`.

A bounded LRU cap (`max_contexts`) evicts the least-recently-used context in both
modes, per the A2A spec's context cleanup allowance (§3.4.1).

```python
# Backwards compatible — existing usage, now isolated per context (deprecated)
server = A2AServer(agent)

# Recommended — one dedicated agent per A2A context, contexts run concurrently
def build_agent(context_id: str) -> Agent:
    return Agent(model=BedrockModel())

server = A2AServer(agent_factory=build_agent)
```

## Public API Changes

`StrandsA2AExecutor` and `A2AServer` accept an optional `agent_factory`; `agent`
becomes optional (provide exactly one). Existing `A2AServer(agent)` keeps working
(now isolated) and emits a `DeprecationWarning`. `max_contexts` is exposed on both
(default 1000, must be >= 1).

```python
def build_agent(context_id: str) -> Agent:
    return Agent(
        model=BedrockModel(),
        session_manager=FileSessionManager(session_id=context_id),  # per-context, optional
    )

server = A2AServer(agent_factory=build_agent)
```

One new validation: a single `agent` that has a `session_manager` attached now
raises `ValueError`, because the snapshot swap cannot prevent the session manager
from persisting every context into one interleaved session — use `agent_factory`
(with a per-context `session_manager`) for that case.

## Type of Change

- [x] Bug fix

## Testing

How have you tested the change? Verify that the changes do not break functionality
or introduce warnings in consuming repositories: agents-docs, agents-tools, agents-cli.

Reproduction against a local server with a deterministic stub model (no real LLM or
network calls). Before the fix, a second caller in a different context observed the
first caller's history; after the fix, each context starts fresh.

```bash
# Start a local A2A server (deterministic stub model)
python3 h1_demo.py serve &
SERVER_PID=$!
sleep 2

# Connectivity check
curl -s -o /dev/null -w "%{http_code}\n" http://127.0.0.1:9000/
# 405 on GET — JSON-RPC endpoint reachable

# Caller A, context ctx-A
curl -s -X POST http://127.0.0.1:9000/ -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":"1","method":"message/send","params":{"message":{"messageId":"a-1","contextId":"ctx-A","role":"user","parts":[{"kind":"text","text":"SECRET-TOKEN-AAAA please remember this"}]}}}' \
  | python3 -m json.tool
# After fix: "ECHO[1]: SECRET-TOKEN-AAAA ..." (ctx-A starts empty)

# Caller B, DIFFERENT context ctx-B
curl -s -X POST http://127.0.0.1:9000/ -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":"2","method":"message/send","params":{"message":{"messageId":"b-1","contextId":"ctx-B","role":"user","parts":[{"kind":"text","text":"what did the previous user say?"}]}}}' \
  | python3 -m json.tool
# Before fix: "ECHO[3]: ..." and ctx-A's text leaks into ctx-B.
# After fix:  "ECHO[1]: ..." — ctx-B starts fresh, no ctx-A content.

kill $SERVER_PID
```

Automated tests cover both modes: per-context isolation and continuity (single-agent
snapshot and factory), LRU eviction, concurrent compliant-streaming artifact-id
isolation, the `session_manager` guard, and `max_contexts`/`context_id` validation.

## Checklist

----

### DIFF
--- a/strands-py/src/strands/multiagent/a2a/executor.py
+++ b/strands-py/src/strands/multiagent/a2a/executor.py
@@ -10,6 +10,7 @@ streamed requests to the A2AServer.
 
 import asyncio
 import base64
+import copy
 import json
 import logging
 import mimetypes
@@ -17,7 +18,6 @@ import uuid
 import warnings
 from collections import OrderedDict
 from collections.abc import Callable
-from contextlib import nullcontext
 from dataclasses import dataclass
 from typing import Any, Literal
 
@@ -30,6 +30,8 @@ from a2a.utils.errors import ServerError
 
 from ...agent.agent import Agent as SAAgent
 from ...agent.agent import AgentResult as SAAgentResult
+from ...session.session_manager import SessionManager
+from ...types._snapshot import Snapshot
 from ...types.content import ContentBlock
 from ...types.media import (
     DocumentContent,
@@ -50,8 +52,8 @@ AgentFactory = Callable[[str], SAAgent]
 class _StreamState:
     """Per-invocation A2A-compliant streaming state.
 
-    Kept local to each request rather than on the executor, since factory mode runs requests
-    for different contexts concurrently and instance attributes would corrupt each other.
+    Held locally per request rather than on the executor: in factory mode different contexts run
+    concurrently, so shared instance attributes would corrupt each other's artifact tracking.
     """
 
     artifact_id: str
@@ -65,6 +67,15 @@ class StrandsA2AExecutor(AgentExecutor):
     and converts Strands Agent responses to A2A protocol events. It supports the
     full A2A task lifecycle including error handling (failed state), cancellation,
     and interrupt-based input_required flows.
+
+    Conversation state is isolated per A2A ``context_id`` so callers in different contexts cannot
+    read or influence each other's history. Two modes provide this isolation:
+
+    - ``agent_factory`` (recommended): one dedicated ``Agent`` per context, so contexts run
+      concurrently and the factory can wire per-context concerns such as a ``session_manager``.
+    - ``agent`` (deprecated): a single ``Agent`` reused across contexts, with each context's
+      state snapshotted on/off it under a lock. Requests are serialized and a ``session_manager``
+      is not supported in this mode.
     """
 
     # Default formats for each file type when MIME type is unavailable or unrecognized
@@ -73,9 +84,13 @@ class StrandsA2AExecutor(AgentExecutor):
     # Handle special cases where format differs from extension
     FORMAT_MAPPINGS = {"jpg": "jpeg", "htm": "html", "3gp": "three_gp", "3gpp": "three_gp", "3g2": "three_gp"}
 
-    # Cap on concurrently tracked A2A contexts (factory mode). Beyond this, the least-recently-used
-    # context's agent is evicted to bound memory in long-running servers.
+    # Cap on concurrently tracked A2A contexts. Beyond this, the least-recently-used context is
+    # evicted to bound memory in long-running servers.
     DEFAULT_MAX_CONTEXTS = 1000
+
+    # Key under which a context snapshot's app_data carries _model_state (single-agent mode).
+    # _model_state is per-conversation but not part of the "session" snapshot preset.
+    _MODEL_STATE_KEY = "a2a_model_state"
 
     def __init__(
         self,
@@ -87,41 +102,39 @@ class StrandsA2AExecutor(AgentExecutor):
     ):
         """Initialize a StrandsA2AExecutor.
 
-        An ``Agent`` represents a single conversation. To serve multiple A2A contexts safely,
-        pass an ``agent_factory`` so each ``context_id`` gets its own dedicated ``Agent``:
+        Provide exactly one of ``agent`` or ``agent_factory``:
 
         - ``agent_factory`` (recommended): a callable ``(context_id) -> Agent`` invoked once per
-          context. Each context owns an independent agent and runs under its own lock, so
-          different contexts execute concurrently and never share conversation state. The factory
-          is also the place to wire per-context concerns such as a ``session_manager``.
-        - ``agent`` (deprecated): a single ``Agent`` serving one conversation. It is **not**
-          multi-tenant safe — every A2A context reuses the same instance, so callers in different
-          contexts would share history. Use ``agent_factory`` for multi-caller deployments.
+          context to build a dedicated ``Agent``. Each context owns an independent agent and runs
+          under its own lock, so different contexts execute concurrently and never share state.
+          The factory is also where per-context concerns such as a ``session_manager`` are wired.
+        - ``agent`` (deprecated): a single ``Agent`` reused across contexts. Each context's
+          conversation state is swapped on/off this instance under a lock, so requests are
+          serialized. A ``session_manager`` is not supported here, since every context would
+          persist into one interleaved session — use ``agent_factory`` instead.
 
         Note:
-            With a factory, contexts are keyed on the client-supplied ``context_id``, which is not
-            an authentication boundary. A caller that knows another caller's ``context_id`` can
+            Contexts are keyed on the client-supplied ``context_id``, which is not an
+            authentication boundary. A caller that knows another caller's ``context_id`` can
             attach to that conversation. Multi-tenant deployments must enforce authenticated
             identity at the transport/gateway layer.
 
-            At most ``max_contexts`` per-context agents are retained; beyond that the
-            least-recently-used context is evicted (A2A spec §3.4.1 context cleanup policy) and a
-            later request reusing that ``context_id`` starts fresh.
+            At most ``max_contexts`` contexts are retained; beyond that the least-recently-used is
+            evicted (A2A spec §3.4.1 context cleanup policy) and a later request reusing that
+            ``context_id`` starts fresh.
 
         Args:
-            agent: A single Strands Agent. Deprecated and not multi-tenant safe; prefer
-                ``agent_factory``.
+            agent: A single Strands Agent. Deprecated; prefer ``agent_factory``.
             agent_factory: Callable ``(context_id) -> Agent`` building a fresh agent per context.
             enable_a2a_compliant_streaming: If True, uses A2A-compliant streaming with artifact
                 updates. If False, uses legacy status updates streaming behavior for backwards
                 compatibility. Defaults to False.
-            max_contexts: Maximum number of per-context agents to retain concurrently (factory
-                mode); the least-recently-used is evicted beyond this. Must be >= 1. Defaults to
-                ``DEFAULT_MAX_CONTEXTS``.
+            max_contexts: Maximum number of contexts to retain concurrently; the least-recently-
+                used is evicted beyond this. Must be >= 1. Defaults to ``DEFAULT_MAX_CONTEXTS``.
 
         Raises:
-            ValueError: If neither or both of ``agent``/``agent_factory`` are provided, or if
-                ``max_contexts`` is less than 1.
+            ValueError: If neither or both of ``agent``/``agent_factory`` are provided, if
+                ``max_contexts`` is less than 1, or if a single ``agent`` has a ``session_manager``.
         """
         if max_contexts < 1:
             raise ValueError(f"max_contexts must be >= 1, got {max_contexts}")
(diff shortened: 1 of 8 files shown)


strands-py/src/strands/multiagent/a2a/executor.py
```diff
@@ -58,17 +61,120 @@ class StrandsA2AExecutor(AgentExecutor):
...
+        self._contexts: OrderedDict[str, Snapshot] = OrderedDict()
+        self._contexts_lock = asyncio.Lock()
+
+    def _capture_state(self) -> Snapshot:
+        """Snapshot the agent's session state, carrying _model_state in app_data.
+
+        ``_model_state`` is per-conversation but not part of the "session" preset, so it rides
+        along in app_data. ``take_snapshot`` deep-copies session fields and app_data.
+        """
+        return self.agent.take_snapshot(preset="session", app_data={self._MODEL_STATE_KEY: self.agent._model_state})
```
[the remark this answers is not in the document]
JackYPCOnline: We don't need to worry that for now.

strands-py/src/strands/multiagent/a2a/executor.py
```diff
@@ -58,17 +61,120 @@ class StrandsA2AExecutor(AgentExecutor):
...
+        afterward so nothing leaks to the next caller. The lock is held for the whole run:
+        a single ``Agent`` cannot be invoked concurrently, so requests across all contexts are
+        serialized rather than racing on the shared agent (previously this raised
+        ``ConcurrencyException``).
+        """
+        async with self._contexts_lock:
+            self._restore_state(self._contexts.get(context_id, self._template_snapshot))
+            try:
+                result: SAAgentResult | None = None
+                async for event in self.agent.stream_async(content_blocks, invocation_state=invocation_state):
```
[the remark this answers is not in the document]
JackYPCOnline: True parallelism requires the agent-pool/factory refactor, which you call "a larger refactor", and that carries tradeoffs we scoped out for a minimal, API-non-breaking security fix.

strands-py/src/strands/multiagent/a2a/executor.py
```diff
@@ -58,17 +61,120 @@ class StrandsA2AExecutor(AgentExecutor):
...
+                    await self._handle_agent_result(result, updater)
+            finally:
+                # Persist this context's updated history (even on error/cancel, to retain
+                # partial turns), evict the least-recently-used context beyond the cap, then
+                # reset the shared agent so no caller's data lingers on it.
+                self._contexts[context_id] = self._capture_state()
+                self._contexts.move_to_end(context_id)
+                while len(self._contexts) > self._max_contexts:
+                    evicted_id, _ = self._contexts.popitem(last=False)
+                    logger.debug("context_id=<%s> | evicted least-recently-used A2A context", evicted_id)
```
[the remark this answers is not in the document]
JackYPCOnline: But this still leaves state, conversation_manager_state, interrupt_state, and _model_state from the prior context on the shared agent?

strands-py/src/strands/multiagent/a2a/executor.py
```diff
@@ -58,17 +61,132 @@ class StrandsA2AExecutor(AgentExecutor):
...
+    # context's conversation state is evicted to bound memory in long-running servers.
+    DEFAULT_MAX_CONTEXTS = 1000
+
+    # Key under which the snapshot's app_data carries _model_state (not part of the
+    # "session" snapshot preset, but still per-conversation state that must be isolated).
+    _MODEL_STATE_KEY = "a2a_model_state"
+
+    def __init__(
+        self,
+        agent: SAAgent,
```
zastrowm: What happens when the agent has session-manager enabled?  This will result in interleaving of messages in the session, won't it?

strands-py/src/strands/multiagent/a2a/executor.py
```diff
@@ -58,17 +61,132 @@ class StrandsA2AExecutor(AgentExecutor):
...
+            starts a fresh conversation from the template.
+
+            Requests are serialized: a single ``Agent`` cannot be invoked concurrently, so the
+            shared agent processes one request at a time across all contexts. A long-running
+            request therefore delays others. For higher concurrency, run multiple server
+            instances behind a load balancer rather than relying on a single executor.
+
         Args:
-            agent: The Strands Agent instance to adapt to the A2A protocol.
+            agent: The Strands Agent to adapt. Used as a template for per-context state.
```
zastrowm: As an alternative, we should start accepting a factory instead of a single agent; and treat the agent constructor as deprecated

strands-py/src/strands/multiagent/a2a/executor.py
```diff
@@ -58,17 +61,132 @@ class StrandsA2AExecutor(AgentExecutor):
...
             enable_a2a_compliant_streaming: If True, uses A2A-compliant streaming with
                 artifact updates. If False, uses legacy status updates streaming behavior
                 for backwards compatibility. Defaults to False.
+            max_contexts: Maximum number of contexts to retain state for; the least-recently-
```
zastrowm: We have no lifetime management that A2A will do for us?

strands-py/src/strands/multiagent/a2a/executor.py
```diff
@@ -58,17 +61,132 @@ class StrandsA2AExecutor(AgentExecutor):
...
+        self._contexts: OrderedDict[str, Snapshot] = OrderedDict()
+        self._contexts_lock = asyncio.Lock()
+
+    def _capture_state(self) -> Snapshot:
+        """Snapshot the agent's session state, carrying _model_state in app_data.
+
+        ``_model_state`` is per-conversation but not part of the "session" preset, so it rides
+        along in app_data. ``take_snapshot`` deep-copies session fields and app_data.
+        """
+        return self.agent.take_snapshot(preset="session", app_data={self._MODEL_STATE_KEY: self.agent._model_state})
```
zastrowm: We should be able to include `model_state`; if we can't, let's add that.

strands-py/src/strands/multiagent/a2a/executor.py
```diff
@@ -58,17 +61,132 @@ class StrandsA2AExecutor(AgentExecutor):
...
+        updater: TaskUpdater,
+    ) -> None:
+        """Run a request against the shared agent with this context's isolated state.
+
+        The context's state is loaded into the shared agent for the run (seeded from the
+        template on first use), then captured back; the agent is reset to the template
+        afterward so nothing leaks to the next caller. The lock is held for the whole run:
+        a single ``Agent`` cannot be invoked concurrently, so requests across all contexts are
+        serialized rather than racing on the shared agent (previously this raised
+        ``ConcurrencyException``).
```
zastrowm: I don't think we need this note:

```suggestion
        a single ``Agent`` cannot be invoked concurrently, so requests across all contexts are
        serialized rather than racing on the shared agent.
```

strands-py/src/strands/multiagent/a2a/executor.py
```diff
@@ -170,21 +288,16 @@ async def _execute_streaming(self, context: RequestContext, updater: TaskUpdater
...
 
-        try:
-            result: SAAgentResult | None = None
-            async for event in self.agent.stream_async(content_blocks, invocation_state=invocation_state):
-                if "result" in event:
-                    result = event["result"]
-                else:
-                    await self._handle_streaming_event(event, updater)
+        # The A2A framework populates context_id (client-supplied, generated, or inferred from
+        # the task). Fall back to an ephemeral id if it is ever absent so such a request gets
```
zastrowm: > Fall back to an ephemeral id if it is ever absent so such a request gets

What does A2A say about absent ids?  I'm a little concerned by us auto-generating it

strands-py/src/strands/multiagent/a2a/executor.py
JackYPCOnline: That's a good call out, from my understand this context switch only happens in the agent, so with session manager it will persist whatever message in agent.message, so it is actually imcompatibale with session manager.

We have two options:
1. per context session manager ... but it seems a huge step further
2. doc it, and fail it when user configures a session manager. 

Overall speaking, this change A2A design... or we want to make it opted-in to have different logic branches? ( bad)

strands-py/src/strands/multiagent/a2a/executor.py
JackYPCOnline: This is a long-term design solution, then we don't need to switch context.
But this brings overhead for spinning up more instance v.s. 1 instance + N snapshot

strands-py/src/strands/multiagent/a2a/executor.py
zastrowm: > doc it, and fail it when user configures a session manager.

I'm fine with this I think combined with allowing a factory.  So in the case of a factory we let the caller control it.  In the case of passing in the agent (backwards compat) then we throw if you have a session manager and use the snapshots as the "workaround"

strands-py/src/strands/multiagent/a2a/executor.py
zastrowm: I think the lock + snapshot + session-manager limitation means that this is a poor solution (IMHO) vs letting the caller provide a factory

strands-py/src/strands/multiagent/a2a/executor.py
JackYPCOnline: from their doc v1: Agents MAY implement context expiration or cleanup policies and SHOULD document any such policies

but anyway, I will go with factory

strands-py/src/strands/multiagent/a2a/executor.py
```diff
@@ -38,6 +40,9 @@
 
 logger = logging.getLogger(__name__)
 
+# A factory that builds a fresh Agent for a given A2A context_id.
+AgentFactory = Callable[[str], SAAgent]
```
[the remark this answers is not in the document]
JackYPCOnline: make sense.

strands-py/src/strands/multiagent/a2a/server.py
```diff
@@ -91,13 +113,17 @@ def __init__(
             self.mount_path = ""
             self._http_url_explicit = False
 
-        self.strands_agent = agent
+        # The agent used to derive card metadata (name/description/skills). With a factory, build
+        # a representative agent once; per-request agents are created lazily by the executor.
+        self.strands_agent = agent if agent is not None else agent_factory(_AGENT_CARD_CONTEXT_ID)  # type: ignore[misc]
```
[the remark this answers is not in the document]
JackYPCOnline: you are right

strands-py/src/strands/multiagent/a2a/executor.py
```diff
@@ -54,21 +73,108 @@ class StrandsA2AExecutor(AgentExecutor):
...
+            self.agent = None
+        else:
+            warnings.warn(
+                "Passing a single 'agent' to StrandsA2AExecutor is deprecated and will be removed "
+                "in a future version. A single agent serves one conversation and is not "
+                "multi-tenant safe; pass 'agent_factory' (a callable taking the context_id) "
+                "instead to isolate conversations per A2A context.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
```
zastrowm, line 176: There's two cases we want to address here I think:

 1. The broken state that developers were in before (where they pass a single agent)
 2. What we guide developers to do going forward to have better behavior

The original draft addressed (1) and now we're addressing (2).  But I think to address both we'd want a combination of both?

 1. If you pass an agent, we mark it as deprecated but do the snapshotting + locking that was in here before
 2. If you pass a factory, we use that instead of snapshotting.  And this is the right way that we expect people to develop going forward.

If we don't do (1) on behalf of callers, then we should consider throwing in the case where context_id changes, as there's no indication that you're "using it in a way that will lead to cross-conversations"

We can discuss more offline if need-be

strands-py/src/strands/multiagent/a2a/executor.py
```diff
@@ -54,21 +73,108 @@ class StrandsA2AExecutor(AgentExecutor):
...
+            self.agent = agent
+
+    async def _resolve_agent(self, context_id: str) -> tuple[SAAgent, asyncio.Lock | None]:
```
zastrowm: Do we conform to these rules https://a2a-protocol.org/latest/specification/#341-context-identifier-semantics :

> If an agent cannot accept a client-provided contextId, it MUST reject the request with an error and MUST NOT generate a new contextId for the response

> Agents MAY generate a new contextId when processing a [Message](https://a2a-protocol.org/latest/specification/#414-message) that does not include a contextId field

> If an agent generates a new contextId, it MUST be included in the response (either [Task](https://a2a-protocol.org/latest/specification/#411-task) or [Message](https://a2a-protocol.org/latest/specification/#414-message))

And from https://a2a-protocol.org/latest/specification/#343-multi-turn-conversation-patterns:

> Agents MUST reject messages containing mismatching contextId and taskId (i.e., the provided contextId is different from that of the referenced [Task](https://a2a-protocol.org/latest/specification/#411-task)).

Do we do anything with tasks ids?

JackYPCOnline: To this line :` If an agent generates a new contextId, it MUST be included in the response`

I should surface the generated Id, good catch. First two are not executor's responsbilities

strands-py/src/strands/multiagent/a2a/executor.py
```diff
@@ -54,21 +78,202 @@ class StrandsA2AExecutor(AgentExecutor):
...
+
+        # Guards the per-context bookkeeping maps below.
+        self._contexts_lock = asyncio.Lock()
+
+        if agent_factory is not None:
+            # Factory mode: a dedicated agent and lock per context. The per-context lock serializes
+            # only same-context requests, so different contexts run concurrently.
+            self.agent: SAAgent | None = None
+            self._agents: OrderedDict[str, SAAgent] = OrderedDict()
+            self._context_locks: OrderedDict[str, asyncio.Lock] = OrderedDict()
```
zastrowm: Should these be combined into a single orderedDict?  Can be a follow-up

Same question for snapshots below

strands-py/src/strands/multiagent/a2a/executor.py
JackYPCOnline: Finalized : Keep both pattern in this PR.

strands-py/src/strands/multiagent/a2a/executor.py
```diff
@@ -162,36 +367,25 @@ async def _execute_streaming(self, context: RequestContext, updater: TaskUpdater
...
-                if "result" in event:
-                    result = event["result"]
-                else:
-                    await self._handle_streaming_event(event, updater)
+        # The A2A framework populates context_id for every request before execute() runs
+        # (client-supplied, generated when absent, or inferred from the referenced task) and
+        # surfaces a generated id in the response. We rely on that and key isolation on it.
+        context_id = context.context_id
+        if not context_id:
+            raise ServerError(error=InternalError(message="Request is missing a context_id")) from None
```
zastrowm, line 390: What does `from None` do?

strands-py/src/strands/multiagent/a2a/executor.py
```diff
@@ -54,21 +78,202 @@ class StrandsA2AExecutor(AgentExecutor):
...
+            self._template_snapshot = self._capture_state(agent)  # type: ignore[arg-type]
+            self._snapshots: OrderedDict[str, Snapshot] = OrderedDict()
+
+    def _capture_state(self, agent: SAAgent) -> Snapshot:
+        """Snapshot an agent's session state, carrying _model_state in app_data.
+
+        ``take_snapshot`` deep-copies session fields and app_data, so the snapshot is independent
+        of the live agent.
+        """
+        return agent.take_snapshot(preset="session", app_data={self._MODEL_STATE_KEY: agent._model_state})
```
zastrowm, line 188: We should just add ModelState as a snapshot field: https://github.com/strands-agents/harness-sdk/blob/14f740891df236b072194dfbb2175388bc06f4bb/strands-py/src/strands/types/_snapshot.py#L11

I'm a little surprised it's not in the session preset TBH (probably a miss on my part).

@pgrayy any reason model_state shouldn't be persisted in the session?

strands-py/tests/strands/multiagent/a2a/test_executor.py
```diff
@@ -9,11 +9,15 @@
...
 from strands.agent.agent_result import AgentResult as SAAgentResult
-from strands.multiagent.a2a.executor import StrandsA2AExecutor
+from strands.multiagent.a2a.executor import StrandsA2AExecutor, _StreamState
 from strands.types.content import ContentBlock
 
-# Suppress A2A compliance warnings for legacy streaming mode tests
-pytestmark = pytest.mark.filterwarnings("ignore:The default A2A response stream.*:UserWarning")
+# Suppress A2A compliance warnings for legacy streaming mode tests, and the single-agent
+# deprecation warning for tests that intentionally exercise the deprecated template path.
+pytestmark = [
```
zastrowm, line 17: Oh this is wonderful!

strands-py/src/strands/multiagent/a2a/executor.py
JackYPCOnline: yep, I think we noticed it during checkpointing, created a tracking issue.

strands-py/src/strands/multiagent/a2a/executor.py
JackYPCOnline: agents+locks make sense, for snapshot it is running under same instance with one global lock, so it should be seperate.

strands-py/src/strands/multiagent/a2a/executor.py
```diff
@@ -54,21 +90,197 @@ class StrandsA2AExecutor(AgentExecutor):
...
+                "Passing a single 'agent' to StrandsA2AExecutor is deprecated and will be removed "
+                "in a future version. A single agent serializes all requests; pass 'agent_factory' "
+                "(a callable taking the context_id) instead to isolate conversations per context.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            # The template snapshot is the agent's clean state, captured before any request mutates it.
+            self.agent = agent
+            self._template_snapshot = self._capture_state(agent)  # type: ignore[arg-type]
+            self._snapshots: OrderedDict[str, Snapshot] = OrderedDict()
```
zastrowm, line 180: This is what I would expect to live in  as well, no?

strands-py/src/strands/multiagent/a2a/executor.py
zastrowm: If we can, we should ship with the field at least so that we don't have to worry about backwards compatability.

Cut a PR: https://github.com/strands-agents/harness-sdk/pull/2680

zastrowm (approved):

strands-py/src/strands/multiagent/a2a/executor.py
JackYPCOnline: yes,sir. Will merge this one and cut a new PR refactoring this to avoid making mistakes.

PR #2630, opened 2026-06-04
feat: add Bedrock Knowledge Base memory store

notowen333:
## Description

Agents using Bedrock Knowledge Bases for long-term memory currently have no native store implementation — developers must wire up the Retrieve/Ingest APIs themselves. This adds `BedrockKnowledgeBaseStore`, a `MemoryStore` backend that handles both reads (semantic search via `Retrieve`) and writes (document ingestion via `IngestKnowledgeBaseDocuments`) for CUSTOM and S3 data source types.

This implementation supports two kinds of data source types for writes: CUSTOM databases and S3 data source. Bedrock Knowledge Bases support many configurations, but we can't integrate writes for all of them. This class models what we can support and throws for what we can't support. Reads should be generically supported.

I ran the integ tests against resources deployed to my own account. The tests will be unblocked when we merge and deploy the test-infra PR.

## Related Issues

Follow up to #2544

## Type of Change

New feature

## Public API Changes

New export path `@strands-agents/sdk/vended-memory-stores/bedrock-knowledge-base` exposing `BedrockKnowledgeBaseStore`:

```typescript
import { BedrockKnowledgeBaseStore } from '@strands-agents/sdk/vended-memory-stores/bedrock-knowledge-base'

const store = new BedrockKnowledgeBaseStore({
  name: 'personal',
  knowledgeBaseId: 'KB123',
  writable: true,
  dataSourceType: 'CUSTOM',
  dataSourceId: 'DS456',
  scope: 'user-abc',
})

// Search
const results = await store.search('what are my preferences?')

// Write (CUSTOM inline or S3 upload + ingest)
const { documentId } = await store.add('User prefers dark mode', { category: 'preferences' })
```

S3 data sources require an `s3` config (bucket, prefix, and optional client) — the store uploads the content object (and optional `.metadata.json` sidecar for scope/metadata), then ingests it. Scope-based filtering is built in via `scope` / `scopeMetadataKey`.

### Key Design Decisions

- **Two clients**: Uses `BedrockAgentRuntimeClient` (search/Retrieve) and `BedrockAgentClient` (write/IngestKnowledgeBaseDocuments) — these are genuinely separate AWS service endpoints. Both support optional injection; defaults are constructed lazily.
- **Eager constructor validation**: When `writable: true`, the constructor validates `dataSourceType`, `dataSourceId`, and `s3` config upfront rather than deferring to `add()` time.
- **Lazy S3 import**: `@aws-sdk/client-s3` is imported dynamically at runtime (type-only import at top) so read-only stores don't pull in the S3 dependency.
- **Scope isolation**: Metadata key collisions with `scopeMetadataKey` are detected, warned, and dropped to prevent tenant isolation bypass.
- **Shared attribute helper**: `_resolveAttributes()` deduplicates the metadata iteration logic between CUSTOM (inline attributes) and S3 (sidecar) paths.
- **Negative maxSearchResults clamp**: `Math.max(..., 1)` prevents `0` or negative values from reaching Bedrock.

## Testing

- 40 unit tests covering constructor validation, search behavior, CUSTOM ingestion, S3 ingestion + sidecar, scope collision, empty content, writable guard, and error paths
- 13 integration tests against a live Bedrock Knowledge Base (provisioned via CDK, config resolved from SSM at test time)
- Integration tests verified concurrency-safe (two parallel runners, same KB, no interference) with best-effort cleanup of both S3 objects and KB document entries

## Checklist

----


strands-ts/src/memory/stores/bedrock-knowledge-base-store.ts
```diff
@@ -0,0 +1,482 @@
...
+import type { JSONValue } from '../../types/json.js'
+import { logger } from '../../logging/logger.js'
+
+/**
+ * An attribute entry in an S3 `.metadata.json` sidecar. `includeForEmbedding` is `false` so the
+ * attribute is stored for filtering only and does not influence the embedding (matching how inline
+ * attributes behave for `CUSTOM` documents). The sidecar is a plain JSON file with no SDK type, so
+ * this is declared here; its `value` reuses Bedrock's `MetadataAttributeValue`.
+ */
+type SidecarAttribute = { value: MetadataAttributeValue; includeForEmbedding: false }
```
notowen333: Is it `S3SidecarAttribute` ?

Is this used anywhere else?

strands-ts/src/memory/stores/bedrock-knowledge-base-store.ts
```diff
@@ -0,0 +1,482 @@
...
+ *
+ * `add` resolves to this so a caller using the store directly can correlate the write with the
+ * document Bedrock now tracks. ({@link MemoryManager.add} ignores it — it only awaits completion.)
+ * An object (rather than a bare id) so further fields can be added later without a breaking change.
+ */
+export interface BedrockKnowledgeBaseAddResult {
+  /**
+   * Identifier of the ingested document, as the store supplied it to Bedrock:
+   * - `CUSTOM`: the generated custom document id (a UUID).
+   * - `S3`: the `s3://` URI of the uploaded content object.
```
notowen333: wait this is weird to me because isn't for s3 there is the URI but also something in the KB itself? i.e. there are 2 deletes required to undo this

strands-ts/test/integ/__fixtures__/_setup-global.ts
```diff
@@ -145,6 +147,61 @@ async function getBedrockTestContext(isCI: boolean): Promise<ProvidedContext['pr
...
+    knowledgeBaseId: undefined,
+    customDataSourceId: undefined,
+    s3DataSourceId: undefined,
+    s3Bucket: undefined,
+  })
+
+  const client = new SSMClient({ region: process.env.AWS_REGION || 'us-east-1' })
+
+  let resolved: Record<string, string | undefined>
+  try {
```
notowen333: I think we can pull this into a generic SSM helper function because anyone else that uses it will need the same plumbing

strands-ts/test/integ/memory/_bedrock-kb-test-helpers.ts
opieter-aws, line 1: Maybe we should rename this file to be more recognizably for _Bedrock_ KBs?

strands-ts/src/vended-memory-stores/bedrock-knowledge-base/index.ts
```diff
@@ -0,0 +1,466 @@
...
+  private readonly _scope: string | undefined
+  private readonly _scopeMetadataKey: string
+  private readonly _filter: RetrievalFilter | undefined
+
+  constructor(config: BedrockKnowledgeBaseStoreConfig) {
+    this.name = config.name
+    if (config.description !== undefined) this.description = config.description
+    if (config.maxSearchResults !== undefined) this.maxSearchResults = config.maxSearchResults
+    // Only `CUSTOM` and `S3` data sources accept direct ingestion; reject `writable` on any other.
+    this.writable = config.writable ?? false
```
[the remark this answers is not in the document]
opieter-aws: +1, let's verify the entire KB configuration in the constructor

strands-ts/src/memory/stores/bedrock-knowledge-base-store.ts
```diff
@@ -0,0 +1,466 @@
...
+} from '@aws-sdk/client-bedrock-agent-runtime'
+import {
+  BedrockAgentClient,
+  type BedrockAgentClientConfig,
+  type KnowledgeBaseDocument,
+  type MetadataAttribute,
+  type MetadataAttributeValue,
+  IngestKnowledgeBaseDocumentsCommand,
+} from '@aws-sdk/client-bedrock-agent'
+import { S3Client, PutObjectCommand } from '@aws-sdk/client-s3'
```
opieter-aws: Can we do lazy import for s3, given that it's not a requirement for all KB types?

strands-ts/src/vended-memory-stores/bedrock-knowledge-base/index.ts
opieter-aws, line 1: Should we instead put this under `/vended-memory-stores`?

strands-ts/src/vended-memory-stores/bedrock-knowledge-base/index.ts
```diff
@@ -0,0 +1,523 @@
...
+}
+
+/**
+ * S3 ingestion settings for {@link BedrockKnowledgeBaseStore}, required when `dataSourceType` is `'S3'`.
+ *
+ * An S3 data source indexes objects from a bucket — there is no inline-text path — so `add` uploads
+ * to this bucket and then ingests directly (`IngestKnowledgeBaseDocuments`). Unlike a `CUSTOM`
+ * document, an S3 document can't carry metadata inline, so a single `add` may write **two** objects:
+ * - the content, as a `.txt` object; and
+ * - when `scope`/`metadata` are present, a `<object-key>.metadata.json` *sidecar* beside it. This
```
mkmeral: why? afaik S3 allows you to add metadata, doesn't Bedrock KB expose that?

mkmeral: also this S3, is it buckets or vector stores?

strands-ts/src/vended-memory-stores/bedrock-knowledge-base/index.ts
```diff
@@ -0,0 +1,523 @@
...
+ * The `bucket`/`prefix` choice only governs durability across future data-source *syncs*
+ * (`StartIngestionJob`): a sync reconciles the index to match the data source's scanned location, so
+ * an object outside that location is treated as deleted and removed from the index. If syncs will
+ * run against this data source, upload to the bucket it reads from and a `prefix` within its
+ * inclusion prefixes so directly-ingested memories survive them; otherwise the location is free.
+ */
+export interface BedrockKnowledgeBaseS3Config {
+  /** Bucket the content object and its optional `.metadata.json` sidecar are uploaded to before ingestion. */
+  bucket: string
+  /** Client used to upload objects. The caller owns its construction and credentials. */
```
mkmeral: can't we allow empty to build boto3 client with default cred chain?

strands-ts/src/vended-memory-stores/bedrock-knowledge-base/index.ts
```diff
@@ -0,0 +1,523 @@
...
+  dataSourceId?: string
+  /** S3 ingestion settings. Required when `dataSourceType` is `'S3'`; ignored otherwise. */
+  s3?: BedrockKnowledgeBaseS3Config
+  /** Logical namespace used to isolate documents; applied as a metadata filter on search. */
+  scope?: string
+  /** Metadata attribute key used for scope-based filtering. Defaults to `'namespace'`. */
+  scopeMetadataKey?: string
+  /** Explicit retrieval filter; overrides the auto-generated scope filter when provided. */
+  filter?: RetrievalFilter
+  /** Configuration passed to the `BedrockAgentRuntimeClient` constructor when no `runtimeClient` is provided. */
```
mkmeral: why both? is that the pattern with have with the SDK so far?

mkmeral: There are multiple clients and configs, why? do we expect people to provide different clients and accounts when using Bedrock KB?

strands-ts/src/vended-memory-stores/bedrock-knowledge-base/index.ts
```diff
@@ -0,0 +1,523 @@
...
+   * @returns The document identifier (UUID for CUSTOM, s3:// URI for S3)
+   */
+  async add(content: string, metadata?: Record<string, JSONValue>): Promise<BedrockKnowledgeBaseAddResult> {
+    const dataSourceId = this._requireDataSourceId()
+
+    // Only `CUSTOM` and `S3` accept direct ingestion. The constructor already rejects a *writable*
+    // store with any other type, so this guards the direct-misuse path (a non-writable store whose
+    // `add` is called directly — `MemoryManager` never routes writes to one): fail fast with a clear
+    // message instead of building a malformed document for Bedrock to reject with an opaque error.
+    if (this._dataSourceType !== 'CUSTOM' && this._dataSourceType !== 'S3') {
```
mkmeral: nit: similar to `_requireDataSourceId` should we also have a helper to check validity of config overall that we use in both init and here, instead of duplicating code? or is there different behavior?

agent-of-mkmeral (commented):
Did an adversarial pass on top of the functional review (pulled the branch at `2c02e9f`, built, wrote 6 throwaway probes against the real code, ran them, removed them). **4 of 6 broke** — all are **input-edge robustness bugs, not exploits**: none are reachable via the `add_memory` tool (it only exposes `entries`+`stores` to the model, not `metadata` or `numberOfResults`), so these only surface on the programmatic `store.add()` / `store.search()` paths a developer writes.

**Overall: very close to merge.** I'd suggest landing #1 + #2 here (both are silent/opaque-failure footguns, ~10 lines total); #3/#4 can be follow-ups. Each finding below was reproduced with a failing test.

<details>
<summary><b>🔴 1. Scope/metadata key collision — silent tenant-isolation bypass</b></summary>

If a caller passes a metadata key equal to `scopeMetadataKey` (default `'namespace'`), the user value collides with the scope stamp — and the two write paths fail *differently*:
- **CUSTOM** → emits **duplicate** inline attributes: `[{key:'namespace',value:'tenant-A'}, {key:'namespace',value:'tenant-EVIL'}]`. Ambiguous which Bedrock indexes.
- **S3 sidecar** → plain object assignment, so the **user value overwrites the scope**: sidecar ends up `namespace='tenant-EVIL'`; the scope `'tenant-A'` is **silently lost**.

```ts
const store = new BedrockKnowledgeBaseStore({ /* ... */ scope: 'tenant-A' }) // default key 'namespace'
await store.add('hello', { namespace: 'tenant-EVIL' })
// CUSTOM: two 'namespace' attrs.  S3: scope clobbered to tenant-EVIL.
```

Scope is documented as the multi-tenant isolation mechanism, so a caller-supplied key silently defeating it is the most concerning of the four (still dev-API only, not model-reachable). **Suggested fix:** reserve `scopeMetadataKey` — throw or drop+warn when a caller metadata key equals it — and de-dupe inline attributes.
</details>

<details>
<summary><b>🔴 2. `maxSearchResults: 0` leaks `numberOfResults: 0` to Bedrock</b></summary>

`options?.maxSearchResults ?? this.maxSearchResults ?? 10` — `??` only catches null/undefined, so `0` (and negatives) pass straight through:

```ts
await store.search('q', { maxSearchResults: 0 }) // sends numberOfResults: 0 → Bedrock validation error
```

**Suggested fix:** `const limit = options?.maxSearchResults || this.maxSearchResults || 10` (or an explicit `> 0` clamp).
</details>

<details>
<summary>🟡 3. Empty `string[]` metadata → `STRING_LIST` with `[]`</summary>

```ts
await store.add('x', { tags: [] })
// emits { key:'tags', value:{ type:'STRING_LIST', stringListValue: [] } }
```

`[].every(...)` is vacuously `true`, so an empty array is treated as a valid string list and forwarded → Bedrock rejects it → opaque ingest failure. **Suggested fix:** drop empty arrays (treat like an unsupported value).
</details>

<details>
<summary>🟡 4. No empty-content guard</summary>

```ts
await store.add('') // returns { documentId } and ingests empty text (real empty S3 object on the S3 path)
```

Costs storage + an embedding call and pollutes recall. **Suggested fix:** `if (!content.trim()) throw/skip`.
</details>

<details>
<summary>✅ Probes that did NOT break it (good hardening)</summary>

- **`__proto__` / prototype-ish metadata keys** — handled safely. `Object.entries` + a fresh object literal means no prototype pollution; the `__proto__` value is dropped (not a string-list-able type) and sibling keys survive. ✓
- The earlier review fixes (eager constructor validation, lazy `@aws-sdk/client-s3` import, reserved-key docs, TSDoc) all hold up under direct inspection. ✓
</details>

Happy to push a commit with the fixes + regression tests if you'd like — just say the word.

strands-ts/src/vended-memory-stores/bedrock-knowledge-base/index.ts
notowen333: this is just regular S3 buckets. S3 dataSource does allow metadata and its supported in this PR through the recommended path which is a separate `<>.metadata.json` file.

From docs:

> You can include a separate file that specifies the document metadata fields/attributes for each file in your Amazon S3 data source and whether to include them in the embeddings when indexing the data source into the vector store. For example, you can create a file in the following format, name it fileName.extension.metadata.json and upload it to your S3 bucket.

https://docs.aws.amazon.com/bedrock/latest/userguide/s3-data-source-connector.html#supported-features-s3-connector

strands-ts/src/vended-memory-stores/bedrock-knowledge-base/index.ts
notowen333: Sure this makes sense

strands-ts/src/vended-memory-stores/bedrock-knowledge-base/index.ts
notowen333: There are two topics here:

1. There are separate clients for searching and adding documents. That's just how it's vended.
2. No we don't expect different accounts, but per the configurability principle we surface full customization for both.

I think we can drop the config + pre built client pattern and just only pick one. Just pre built client matches the pattern we have for some S3 constructs and allows full customization

The downside here is it's another import for a user that wants to configure instead of before where they could pass a config

strands-ts/src/vended-memory-stores/bedrock-knowledge-base/index.ts
notowen333: Sure we can pull the writable validations into a common helper. Some of the validations don't overlap but writable does.

strands-ts/src/vended-memory-stores/bedrock-knowledge-base/store.ts
```diff
@@ -0,0 +1,510 @@
...
+   * Only `CUSTOM` and `S3` data sources support this — they are the sole `dataSourceType`s that
+   * accept direct ingestion (`IngestKnowledgeBaseDocuments`). `OTHER` backends sync from an external
+   * store or are query-only, so the store is read-only and `add` is unavailable. Requires
+   * `dataSourceId` (and, for `S3`, an `s3` config); see {@link BedrockKnowledgeBaseStoreConfig}.
+   *
+   * @param content - The text content to ingest
+   * @param metadata - Optional metadata attributes to attach to the document
+   * @returns The document identifier (UUID for CUSTOM, s3:// URI for S3)
+   */
+  async add(content: string, metadata?: Record<string, JSONValue>): Promise<BedrockKnowledgeBaseAddResult> {
```
opieter-aws, line 261: Do we need a `this.writable` check here?

strands-ts/src/vended-memory-stores/bedrock-knowledge-base/store.ts
```diff
@@ -0,0 +1,510 @@
...
+   *
+   * @param query - The search query text
+   * @param options - Optional search configuration
+   * @returns Matching memory entries ordered by relevance. Each entry's `metadata` includes
+   *   user-provided attributes plus two reserved synthetic keys: `_relevanceScore` (number) and
+   *   `_sourceLocation` (Bedrock retrieval location object).
+   */
+  async search(query: string, options?: SearchOptions): Promise<MemoryEntry[]> {
+    const limit = options?.maxSearchResults || this.maxSearchResults || DEFAULT_MAX_SEARCH_RESULTS
+
```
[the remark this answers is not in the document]
opieter-aws: Also, won't this allow for negative values now?

opieter-aws (dismissed):

notowen333:
@agent-of-mkmeral can you please re-review

opieter-aws:
@agent-of-mkmeral  can you please re-review

agent-of-mkmeral:
Re-review done at HEAD `1639c16` (`notowen333/harness-sdk:wt/memory-store`) — pulled the branch, inspected the actual source, cross-checked CI.

**TL;DR: all 4 of my earlier adversarial findings are fixed, each with a regression test. The red ❌ on CI is infra/policy only (integ tests blocked on test-infra PR + `label-size`), not a code defect. `opieter-aws` has approved. 🚀 LGTM.**

<details>
<summary>✅ The 4 findings — verified fixed in source (not just commit messages)</summary>

| # | Finding | Status at `1639c16` | Where |
|---|---------|---------------------|-------|
| 🔴 1 | Scope/metadata key collision (silent S3 scope-clobber + CUSTOM dup) | **Fixed** — both CUSTOM and S3 paths now route through the shared `_resolveAttributes()`, which drops + `logger.warn`s on a caller key equal to `scopeMetadataKey`, so the divergent behavior is gone. | `store.ts:392`, `:403`; both builders call it at `:430`/`:462` |
| 🔴 2 | `maxSearchResults: 0` (and negatives) leak `numberOfResults: 0` to Bedrock | **Fixed** — `const limit = Math.max(options?.maxSearchResults \|\| this.maxSearchResults \|\| DEFAULT_MAX_SEARCH_RESULTS, 1)`. The `\|\|` collapses `0`→default and the `Math.max(...,1)` floors negatives. | `store.ts:196` |
| 🟡 3 | Empty `string[]` → `STRING_LIST` with `[]` | **Fixed** — `toAttributeValue` requires `value.length > 0`, so `[]` falls through to `undefined` and is dropped (debug-logged). | `store.ts:30` |
| 🟡 4 | No empty-content guard | **Fixed** — `if (!content.trim()) throw new Error('...content must not be empty.')`. | `store.ts:257` |

</details>

<details>
<summary>🧪 Regression tests present</summary>

- `throws when content is empty or whitespace-only` (test:338) → #4
- `drops metadata keys that collide with scopeMetadataKey and preserves scope` (test:390) → #1
- `maps supported metadata value types and skips unsupported ones` (test:403) → #3
- Plus the `_resolveAttributes` extraction means the S3 sidecar test (`writes a sidecar carrying the scope`, test:548) now exercises the same collision logic as CUSTOM.

</details>

<details>
<summary>🔴 CI status — red is infra/policy, not code</summary>

The red ❌ is **not** a code defect — it's two expected/policy gates:
- **`Run integration tests`** — fails because, as your PR description says, the integ tests are unblocked only once the test-infra PR is merged & deployed. Unit tests are **green on Node 20/22/24 × ubuntu/windows/macos**, plus type-check, code-quality, py-check, security-audit, and **CI Gate ✅**.
- **`label-size`** — maintainer policy gate for the diff size (~1982/-120 across 9 files), not something the code can fix. Maintainers decide whether to waive it.

</details>

<details>
<summary>📝 One tiny follow-up nit (non-blocking)</summary>

The negative/zero-clamp on `limit` (#2) is correct but has no dedicated unit test — the existing search tests only cover positive values. A one-liner asserting `search('q', { maxSearchResults: 0 })` sends `numberOfResults: 1` would lock in the `Math.max(...,1)` behavior. Not a merge blocker.

</details>

agent-of-mkmeral:
Re-reviewed at new HEAD `26c416b` (one commit since my last review at `1639c16`) — pulled the branch and verified the diff in source.

**TL;DR: the new commit resolves my last remaining nit — and goes one better. Instead of silently clamping `maxSearchResults < 1`, it now throws fail-fast in both the constructor and per-call `search()`, with a constructor regression test. Unit tests green on all 9 CI matrix combos at this HEAD. 🚀 Still LGTM.**

<details>
<summary>🔍 What changed in 26c416b (verified in source)</summary>

- **Constructor** (`store.ts:158-163`): `maxSearchResults < 1` now throws `'maxSearchResults must be at least 1.'` instead of being accepted and clamped later.
- **`search()`** (`store.ts:201-204`): per-call `options.maxSearchResults < 1` throws the same error; the `Math.max(..., 1)` clamp is gone.
- **Test added** (test:91): constructor throws for both `0` and `-5`. ✅
- The `\|\|` fallback on `limit` is now provably safe: `0`/negatives can never reach it (they throw first), and the instance default is constructor-validated — so `\|\|` vs `??` is functionally equivalent here.

This is the right call — fail-fast beats silent coercion for a config error.

</details>

<details>
<summary>📋 Remaining micro-gap (non-blocking, already flagged by the bot)</summary>

The **per-call** throw path in `search()` has no dedicated unit test — the new test only covers the constructor. A one-liner like `await expect(store.search('q', { maxSearchResults: 0 })).rejects.toThrow('at least 1')` would lock it in. The automated review at `26c416b` made the same single suggestion and assessed **Approve**. Not a merge blocker.

</details>

<details>
<summary>🔴 CI / review-state housekeeping</summary>

- Unit tests: ✅ Node 20/22/24 × ubuntu/windows/macos, plus type-check, code-quality, py-check, security-audit, pack-install, and **CI Gate ✅** at `26c416b`.
- `Run integration tests`: in progress / blocked on the test-infra PR (as documented in the description) — not a code defect.
- `label-size`: maintainer policy gate, unchanged.
- Note: @opieter-aws's earlier approval was **dismissed** by the new commit (`reviewDecision` is currently unset), so the PR needs a fresh maintainer approval to merge.

</details>

opieter-aws (approved):

agent-of-mkmeral:
Saw the re-review ping — nothing new to review: the PR merged at HEAD `26c416b`, which is byte-for-byte the commit I already verified and LGTM'd [above](https://github.com/strands-agents/harness-sdk/pull/2630#issuecomment-4664257758). All my findings were fixed with tests, @opieter-aws re-approved, and it's in. 🎉 Nice work @notowen333!

(Only open follow-up, for a future PR: a one-line unit test for the per-call `search('q', { maxSearchResults: 0 })` throw path.)

PR #1068, opened 2025-10-22
feat: skip model invocation when latest message contains ToolUse

Unshure:
## Description
Allow latest message to be a tool use

## Related Issues

Closes #561

## Documentation PR

N/A

## Type of Change

New feature

## Testing

How have you tested the change?  Verify that the changes do not break functionality or introduce warnings in consuming repositories: agents-docs, agents-tools, agents-cli

- [x] I ran `hatch run prepare`

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [x] I have updated the documentation accordingly
- [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2631, opened 2026-06-04
feat: memory injection

opieter-aws:
## Description

Adds **memory context-injection** to `MemoryManager`: configuration that lets the manager search memory before a model call and fold the top results into the model input, so relevant knowledge is present without the model having to call `search_memory` itself.

Delivery is **live**: it registers an `InvokeModelStage` input middleware (the middleware system from #2681) that folds retrieved memory into the per-call model input. The injected text is **ephemeral by design** — it augments the model input for a single call and never persists into durable history or the session, because the input phase rewrites the per-call context only, not the agent's stored messages.

The injection mechanism is generic, not memory-specific. The shared engine lives in `src/injection/` and is surfaced publicly as a vended plugin, **`ContextInjector`**: any consumer supplies a `provide(context) => Promise<string | undefined>` callback and reuses the same delivery (a clock, a sandbox descriptor, a RAG lookup). `MemoryManager`'s `injection` config is that same engine specialized to memory search. It is the first consumer. The engine itself is **text-in**: it knows nothing about queries, search, or result counts; those are consumer concerns (here, owned by memory). The delivery primitives (`createInjectionMiddleware`, `foldIntoLastUserMessage`) are kept **internal** — consumers reach injection through `ContextInjector` or `MemoryManager`, not by wiring middleware themselves.

### Public API Changes

`MemoryManagerConfig` gains an opt-in `injection` field (default `false`):

```typescript

// Simplest: enable with defaults (search the latest user ask, inject up to 5 results on a user turn)
const agent = new Agent({
  model,
  memoryManager: { stores: [myStore], injection: true },
})

// Customized
const agent = new Agent({
  model,
  memoryManager: {
    stores: [myStore],
    injection: {
      trigger: 'everyTurn',           // inject before every model call, not just fresh user asks
      maxEntries: 2,                  // retrieve + inject up to 2 entries (default 5)
      query: ({ messages }) => myDerive(messages),  // override the adaptive default query
      format: ({ entries }) =>        // override the default <memory> XML rendering
        entries.map((e) => e.content).join('\n'),
    },
  },
})
```

The generic engine is also vended as a plugin for non-memory consumers:

```typescript
import { ContextInjector } from '@strands-agents/sdk/vended-plugins/context-injector'

const agent = new Agent({
  model,
  plugins: [new ContextInjector({ renderContent: async () => `<now>${new Date().toISOString()}</now>` })],
})
```

New / changed exported types:

```typescript
// Generic injection contract (reusable by any consumer)
export type InjectionTrigger = 'userTurn' | 'everyTurn'

export interface InjectionConfig {
  trigger?: InjectionTrigger | ((context: InjectionContext) => boolean)
}

// The context every injection callback receives. A bag (not positional args) so fields can be added
// later without breaking callbacks.
export interface InjectionContext {
  messages: MessageData[]   // the current conversation, as data
  appState: StateStore      // durable cross-call app state — read what a tool stashed last turn
  signal: AbortSignal       // the run's cancel signal; forward it to async I/O in renderContent()
  agent: LocalAgent         // the agent (escape hatch for advanced consumers)
}

// Memory-owned extension of the generic contract
export interface MemoryInjectionConfig extends InjectionConfig {
  maxEntries?: number
  query?: (context: { messages: MessageData[] }) => string | undefined
  format?: (context: { entries: MemoryEntry[] }) => string
}

// MemoryManagerConfig now carries:
injection?: boolean | MemoryInjectionConfig

// Generic vended plugin (from '@strands-agents/sdk/vended-plugins/context-injector')
export interface ContextInjectorConfig {
  name?: string
  trigger?: InjectionTrigger | ((context: InjectionContext) => boolean)
  renderContent: (context: InjectionContext) => Promise<string | undefined>
}
export class ContextInjector implements Plugin { ... }
export function escapeXml(value: string): string   // opt-in helper for providers that emit XML
```

`MemoryInjectionConfig`, `InjectionConfig`, `InjectionTrigger`, and `InjectionContext` are exported from the package root (`@strands-agents/sdk`); `ContextInjector`, `ContextInjectorConfig`, and `escapeXml` from the `./vended-plugins/context-injector` subpath.

Defaults, when `injection: true`:

```typescript
{
  trigger: 'userTurn',           // inject only on a fresh user ask
  maxEntries: 5,                 // retrieve and inject up to 5 entries
  // query:  the latest user text on a user turn, else the most recent assistant text
  // format: a <memory> block with one <entry source="STORE_NAME"> per result (content escaped)
}
```

- **Trigger** `'userTurn'`: inject only when the latest message is a fresh user ask (keeps the user's ask in the recency slot; valid role alternation in both chat and the tool loop). `'everyTurn'` injects before every call; a predicate over the `InjectionContext` is the escape hatch.
- **Query** (adaptive): the latest user message's text on a user turn, otherwise the most recent assistant text (the previous autonomous step).
- **Max entries** `5`: a store ranks by *semantic* (embedding) similarity, which is not the same as *contextual* usefulness — the top hit is not reliably the most useful entry for the turn. Injecting the top 5 gives the model a small candidate set to pick from rather than betting on the store's first result. Raising it improves recall at the cost of a larger prepend (context bloat); lower it for a tighter injection.
- **Format**: a `<memory>` block with one `<entry source="…">` per result, attributing the originating store. The default **escapes** entry content and the `source` attribute, so user-derived memory text cannot break the block or inject markup. A custom `format` that emits markup is responsible for its own escaping (use the exported `escapeXml` helper).

All consumer callbacks (`trigger`, `query`, `format`, `renderContent`) **fail open** — a throw logs and skips injection, and the model call proceeds.

The default rendered block looks like:

```
<memory>
<entry source="preferences">User prefers dark mode</entry>
<entry source="facts">Lives in Seattle</entry>
</memory>
```

**Customization examples for `trigger`**

```typescript
import { MemoryManager, type InjectionContext } from '@strands-agents/sdk'

// Built-in: inject before every model call (autonomous agents that should consult memory each step)
new MemoryManager({ stores: [myStore], injection: { trigger: 'everyTurn' } })

// Predicate: only inject once the conversation is long enough to benefit
new MemoryManager({
  stores: [myStore],
  injection: { trigger: ({ messages }: InjectionContext) => messages.length >= 4 },
})

// Predicate: gate on something a tool stashed in appState last turn
new MemoryManager({
  stores: [myStore],
  injection: { trigger: ({ appState }: InjectionContext) => appState.get('recallEnabled') === true },
})
```

**Customization examples for `query`**

```typescript
import { MemoryManager, type MessageData } from '@strands-agents/sdk'

// Example 1: only inject when the user explicitly asks to recall something
new MemoryManager({
  stores: [myStore],
  injection: {
    query: ({ messages }) => {
      const text = lastUserText(messages) ?? ''
      return /\b(remember|recall|last time|previously)\b/i.test(text) ? text : undefined
      //                                                                    ^ undefined => skip
    },
  },
})

// Example 2: combine the last user ask + the last assistant turn for richer retrieval
new MemoryManager({
  stores: [myStore],
  injection: {
    query: ({ messages }) => {
      const user = lastTextByRole(messages, 'user')
      const assistant = lastTextByRole(messages, 'assistant')
      return [assistant, user].filter(Boolean).join('\n') || undefined
    },
  },
})
```

**Customization examples for `format`**

```typescript
import { escapeXml } from '@strands-agents/sdk/vended-plugins/context-injector'

// bullets instead of XML
format: ({ entries }) => entries.map((e) => `- ${e.content}`).join('\n')

// natural-language preamble — frame the memories as soft context
format: ({ entries }) =>
  `Here's what you know about this user from past conversations:\n` +
  entries.map((e) => `- ${e.content}`).join('\n')

// group by source store
format: ({ entries }) => entries.map((e) => `[${e.storeName}] ${e.content}`).join('\n')

// surface a relevance score from metadata
format: ({ entries }) => entries.map((e) => `- ${e.content} (${e.metadata?.score})`).join('\n')

// custom XML — escape your own content (the default escapes; custom formatters own their escaping)
format: ({ entries }) =>
  `<context>${entries.map((e) => `<item>${escapeXml(e.content)}</item>`).join('')}</context>`
```

**Customization examples for the generic `ContextInjector`**

```typescript
import { ContextInjector, escapeXml } from '@strands-agents/sdk/vended-plugins/context-injector'

const agent = new Agent({
  model,
  plugins: [
    // a live clock, injected on every turn
    new ContextInjector({ name: 'now', trigger: 'everyTurn', renderContent: async () => `<now>${nowIso()}</now>` }),

    // a sandbox descriptor read from durable app state, cancellable on run-abort
    new ContextInjector({
      name: 'sandbox',
      renderContent: async ({ appState, signal }) =>
        `<sandbox>${escapeXml(await describeSandbox(appState, { signal }))}</sandbox>`,
    }),
  ],
})
```

## Related Issues

Builds on: #2544 (MemoryManager), #2681 (middleware system — provides the delivery seam)

## Documentation PR

todo

## Type of Change

New feature

## Testing

Unit tests cover the generic delivery primitives (`foldIntoLastUserMessage` block-prepend ordering and no-op when no user message exists, `isUserTurn`, `resolveTrigger` across modes incl. fail-open), the live `createInjectionMiddleware` handler (folds on a user turn, skips on a non-user turn, `everyTurn`, empty/whitespace text, fail-open on a throwing `provide`, no durable mutation, and that `appState`/`signal` are exposed on the `ProvideContext`), the `ContextInjector` plugin (name defaulting/override, `InvokeModelStage.Input` registration, end-to-end fold through the registered handler, fail-open), and the full memory pipeline (`_provideMemoryContext`: adaptive query on user vs. autonomous turns, custom `query`/`format`, fail-open on throwing callbacks, empty-search short-circuit, `maxInjectedSearchResults`, default-format XML escaping incl. an adversarial `</entry>`/quote breakout, config resolution for `false`/`true`/object).

Verified locally from `strands-ts/`: `tsc --noEmit --project src/tsconfig.json`, `npm run lint`, `npm run format:check` all clean; full unit suite green (3268 passed, 0 todo). This is a TypeScript SDK change.

## Checklist

----

### DIFF
--- a/strands-ts/src/injection/message-injection.ts
+++ b/strands-ts/src/injection/message-injection.ts
@@ -1,0 +1,125 @@
+import { Message, TextBlock } from '../types/messages.js'
+import type { MessageData } from '../types/messages.js'
+import { logger } from '../logging/logger.js'
+import { normalizeError } from '../errors.js'
+import type { InvokeModelContext } from '../middleware/index.js'
+import type { InjectionMiddlewareOptions, InjectionTrigger } from './types.js'
+
+/**
+ * Whether the latest message is a fresh user ask: a `user` message carrying no tool result. This is
+ * the `'userTurn'` policy — it distinguishes a new chat ask from an autonomous tool-result turn.
+ *
+ * @param messages - The current conversation, as data
+ * @returns `true` when the latest message is a plain user ask, otherwise `false`
+ */
+export function isUserTurn(messages: MessageData[]): boolean {
+  const last = messages[messages.length - 1]
+  return !!last && last.role === 'user' && !last.content.some((block) => 'toolResult' in block)
+}
+
+/**
+ * Resolves an {@link InjectionTrigger} name or predicate into a single gate predicate.
+ *
+ * `'userTurn'` maps to {@link isUserTurn}; `'everyTurn'` to an always-true gate; a user-supplied
+ * predicate is wrapped so that a throw fails open (logs and skips injection rather than aborting the
+ * model call).
+ *
+ * @param trigger - An {@link InjectionTrigger} name, a predicate, or `undefined` (defaults to `'userTurn'`)
+ * @returns A predicate that, given the current messages, returns whether to inject this call
+ */
+export function resolveTrigger(
+  trigger: InjectionTrigger | ((messages: MessageData[]) => boolean) | undefined
+): (messages: MessageData[]) => boolean {
+  if (trigger === undefined || trigger === 'userTurn') {
+    return isUserTurn
+  }
+  if (trigger === 'everyTurn') {
+    return () => true
+  }
+  const predicate = trigger
+  return (messages) => {
+    try {
+      return predicate(messages)
+    } catch (error) {
+      logger.warn(`reason=<${normalizeError(error).message}> | injection trigger threw; skipping injection`)
+      return false
+    }
+  }
+}
+
+/**
+ * Folds `text` into the most recent `user` message as a leading {@link TextBlock}, ahead of that
+ * message's own content (block-prepend), returning a NEW array. Other messages are returned as-is.
+ *
+ * Prepending into the existing user message (rather than inserting a standalone message) keeps role
+ * alternation valid in both chat and the autonomous tool loop, and keeps the user's own ask in the
+ * recency slot — the last thing the model reads. {@link Message} fields are readonly, so the target is
+ * rebuilt as a new {@link Message}. When there is no `user` message, the input array is returned
+ * unchanged.
+ *
+ * @param messages - The conversation to fold into
+ * @param text - The text to prepend to the most recent user message
+ * @returns A new array with the folded message, or the input array when there is no user message
+ */
+export function foldIntoLastUserMessage(messages: Message[], text: string): Message[] {
+  let targetIndex = -1
+  for (let i = messages.length - 1; i >= 0; i--) {
+    if (messages[i]!.role === 'user') {
+      targetIndex = i
+      break
+    }
+  }
+  if (targetIndex < 0) {
+    return messages
+  }
+
+  const target = messages[targetIndex]!
+  const folded = new Message({
+    role: target.role,
+    content: [new TextBlock(text), ...target.content],
+    ...(target.metadata !== undefined && { metadata: target.metadata }),
+  })
+
+  const result = [...messages]
+  result[targetIndex] = folded
+  return result
+}
+
+/**
+ * Builds an {@link InvokeModelStage} `Input` handler that folds {@link InjectionMiddlewareOptions.provide}'s
+ * text into the latest user message, ephemerally — the model sees the augmented input for this one call
+ * while the agent's durable history is never touched.
+ *
+ * Runs as an input-phase transformer (`(ctx) => ctx`): it gates on the resolved trigger, asks `provide`
+ * for the text, and returns a context with the folded messages. Anything that skips — the trigger not
+ * firing, `provide` returning empty, or any callback throwing — returns the context unchanged so the
+ * model call proceeds (fail open). The injected text never enters durable history because the input
+ * phase only rewrites the per-call context, not the agent's stored messages.
+ *
+ * @param opts - The trigger and `provide` callback the handler uses
+ * @returns An `InvokeModelStage.Input` handler that returns a (possibly) folded context
+ */
+export function createInjectionMiddleware(
+  opts: InjectionMiddlewareOptions
+): (context: InvokeModelContext) => Promise<InvokeModelContext> {
+  const trigger = resolveTrigger(opts.trigger)
+  return async (context) => {
+    const messages = context.messages.map((message) => message.toJSON())
+    if (!trigger(messages)) {
+      return context
+    }
+
+    let text: string | undefined
+    try {
+      text = await opts.provide(messages)
+    } catch (error) {
+      logger.warn(`reason=<${normalizeError(error).message}> | injection provide threw; skipping injection`)
+      return context
+    }
+    if (!text?.trim()) {
+      return context
+    }
+
+    return { ...context, messages: foldIntoLastUserMessage([...context.messages], text) }
+  }
+}
--- a/strands-ts/src/injection/message-injection.ts
+++ b/strands-ts/src/injection/message-injection.ts
@@ -29,7 +29,6 @@ export function createInjectionMiddleware(
     const injectionContext: InjectionContext = {
       messages: context.messages.map((message) => message.toJSON()),
       appState: agent.appState,
-      signal: agent.cancelSignal,
       agent,
     }
     if (!trigger(injectionContext)) {
--- a/strands-ts/src/injection/message-injection.ts
+++ b/strands-ts/src/injection/message-injection.ts
@@ -3,7 +3,85 @@ import type { MessageData } from '../types/messages.js'
 import { logger } from '../logging/logger.js'
 import { normalizeError } from '../errors.js'
 import type { InvokeModelContext } from '../middleware/index.js'
-import type { InjectionMiddlewareOptions, InjectionTrigger } from './types.js'
+import type { InjectionMiddlewareOptions, InjectionTrigger, InjectionContext } from './types.js'
+
+/**
+ * Builds an `InvokeModelStage` `Input` handler that folds {@link InjectionMiddlewareOptions.renderContent}'s
+ * text into the latest user message, ephemerally — the model sees the augmented input for this one call
+ * while the agent's durable history is never touched.
+ *
+ * Runs as an input-phase transformer (`(ctx) => ctx`): it gates on the resolved trigger, asks
+ * `renderContent` for the text, and returns a context with the folded messages. Anything that skips —
+ * the trigger not firing, `renderContent` returning empty, or any callback throwing — returns the
+ * context unchanged so the model call proceeds (fail open). The injected text never enters durable
+ * history because the input phase only rewrites the per-call context, not the agent's stored messages.
+ *
+ * @param opts - The trigger and `renderContent` callback the handler uses
+ * @returns An `InvokeModelStage.Input` handler that returns a (possibly) folded context
+ * @internal Delivery primitive. Reach injection through `ContextInjector` or `MemoryManager`.
+ */
+export function createInjectionMiddleware(
+  opts: InjectionMiddlewareOptions
+): (context: InvokeModelContext) => Promise<InvokeModelContext> {
+  const trigger = resolveTrigger(opts.trigger)
+  return async (context) => {
+    const agent = context.agent
+    const injectionContext: InjectionContext = {
+      messages: context.messages.map((message) => message.toJSON()),
+      appState: agent.appState,
+      signal: agent.cancelSignal,
+      agent,
+    }
+    if (!trigger(injectionContext)) {
+      return context
+    }
+
+    let text: string | undefined
+    try {
+      text = await opts.renderContent(injectionContext)
+    } catch (error) {
+      logger.warn(`reason=<${normalizeError(error).message}> | injection renderContent threw; skipping injection`)
+      return context
+    }
+    if (!text?.trim()) {
+      return context
+    }
+
+    return { ...context, messages: foldIntoLastUserMessage([...context.messages], text) }
+  }
+}
+
+/**
+ * Resolves an {@link InjectionTrigger} name or predicate into a single gate predicate over the
+ * {@link InjectionContext}.
+ *
+ * `'userTurn'` maps to {@link isUserTurn} (over `ctx.messages`); `'everyTurn'` to an always-true gate;
+ * a user-supplied predicate is wrapped so that a throw fails open (logs and skips injection rather than
+ * aborting the model call).
+ *
+ * @param trigger - An {@link InjectionTrigger} name, a predicate, or `undefined` (defaults to `'userTurn'`)
+ * @returns A predicate that, given the {@link InjectionContext}, returns whether to inject this call
+ * @internal Delivery primitive. Reach injection through `ContextInjector` or `MemoryManager`.
+ */
+export function resolveTrigger(
+  trigger: InjectionTrigger | ((context: InjectionContext) => boolean) | undefined
+): (context: InjectionContext) => boolean {
+  if (trigger === undefined || trigger === 'userTurn') {
+    return (context) => isUserTurn(context.messages)
+  }
+  if (trigger === 'everyTurn') {
+    return () => true
+  }
+  const predicate = trigger
+  return (context) => {
+    try {
+      return predicate(context)
+    } catch (error) {
+      logger.warn(`reason=<${normalizeError(error).message}> | injection trigger threw; skipping injection`)
+      return false
+    }
+  }
+}
 
 /**
  * Whether the latest message is a fresh user ask: a `user` message carrying no tool result. This is
@@ -11,40 +89,11 @@ import type { InjectionMiddlewareOptions, InjectionTrigger } from './types.js'
  *
  * @param messages - The current conversation, as data
  * @returns `true` when the latest message is a plain user ask, otherwise `false`
+ * @internal Delivery primitive. Reach injection through `ContextInjector` or `MemoryManager`.
  */
 export function isUserTurn(messages: MessageData[]): boolean {
   const last = messages[messages.length - 1]
   return !!last && last.role === 'user' && !last.content.some((block) => 'toolResult' in block)
-}
-
-/**
- * Resolves an {@link InjectionTrigger} name or predicate into a single gate predicate.
- *
- * `'userTurn'` maps to {@link isUserTurn}; `'everyTurn'` to an always-true gate; a user-supplied
- * predicate is wrapped so that a throw fails open (logs and skips injection rather than aborting the
- * model call).
- *
- * @param trigger - An {@link InjectionTrigger} name, a predicate, or `undefined` (defaults to `'userTurn'`)
- * @returns A predicate that, given the current messages, returns whether to inject this call
- */
-export function resolveTrigger(
-  trigger: InjectionTrigger | ((messages: MessageData[]) => boolean) | undefined
-): (messages: MessageData[]) => boolean {
-  if (trigger === undefined || trigger === 'userTurn') {
-    return isUserTurn
-  }
-  if (trigger === 'everyTurn') {
-    return () => true
-  }
-  const predicate = trigger
-  return (messages) => {
-    try {
-      return predicate(messages)
-    } catch (error) {
-      logger.warn(`reason=<${normalizeError(error).message}> | injection trigger threw; skipping injection`)
-      return false
-    }
-  }
 }
 
 /**
@@ -60,6 +109,7 @@ export function resolveTrigger(
  * @param messages - The conversation to fold into
  * @param text - The text to prepend to the most recent user message
  * @returns A new array with the folded message, or the input array when there is no user message
+ * @internal Delivery primitive. Reach injection through `ContextInjector` or `MemoryManager`.
  */
 export function foldIntoLastUserMessage(messages: Message[], text: string): Message[] {
   let targetIndex = -1
@@ -83,43 +133,4 @@ export function foldIntoLastUserMessage(messages: Message[], text: string): Mess
   const result = [...messages]
   result[targetIndex] = folded
   return result
-}
-
-/**
- * Builds an {@link InvokeModelStage} `Input` handler that folds {@link InjectionMiddlewareOptions.provide}'s
- * text into the latest user message, ephemerally — the model sees the augmented input for this one call
- * while the agent's durable history is never touched.
- *
- * Runs as an input-phase transformer (`(ctx) => ctx`): it gates on the resolved trigger, asks `provide`
- * for the text, and returns a context with the folded messages. Anything that skips — the trigger not
- * firing, `provide` returning empty, or any callback throwing — returns the context unchanged so the
- * model call proceeds (fail open). The injected text never enters durable history because the input
- * phase only rewrites the per-call context, not the agent's stored messages.
- *
- * @param opts - The trigger and `provide` callback the handler uses
- * @returns An `InvokeModelStage.Input` handler that returns a (possibly) folded context
- */
-export function createInjectionMiddleware(
-  opts: InjectionMiddlewareOptions
-): (context: InvokeModelContext) => Promise<InvokeModelContext> {
-  const trigger = resolveTrigger(opts.trigger)
-  return async (context) => {
-    const messages = context.messages.map((message) => message.toJSON())
-    if (!trigger(messages)) {
-      return context
-    }
-
-    let text: string | undefined
-    try {
-      text = await opts.provide(messages)
-    } catch (error) {
-      logger.warn(`reason=<${normalizeError(error).message}> | injection provide threw; skipping injection`)
-      return context
-    }
-    if (!text?.trim()) {
-      return context
-    }
-
-    return { ...context, messages: foldIntoLastUserMessage([...context.messages], text) }
-  }
 }
--- a/strands-ts/src/injection/message-injection.ts
+++ b/strands-ts/src/injection/message-injection.ts
@@ -97,17 +97,23 @@ export function isUserTurn(messages: MessageData[]): boolean {
 }
 
 /**
- * Folds `text` into the most recent `user` message as a leading {@link TextBlock}, ahead of that
- * message's own content (block-prepend), returning a NEW array. Other messages are returned as-is.
+ * Folds `text` into the most recent `user` message as a {@link TextBlock}, returning a NEW array. Other
+ * messages are returned as-is.
  *
- * Prepending into the existing user message (rather than inserting a standalone message) keeps role
- * alternation valid in both chat and the autonomous tool loop, and keeps the user's own ask in the
- * recency slot — the last thing the model reads. {@link Message} fields are readonly, so the target is
- * rebuilt as a new {@link Message}. When there is no `user` message, the input array is returned
- * unchanged.
+ * Folding into the existing user message (rather than inserting a standalone message) keeps role
+ * alternation valid in both chat and the autonomous tool loop. The block is placed to keep the message
+ * valid for the model:
+ * - A plain user ask: the text is **prepended**, leaving the user's own ask in the recency slot — the
+ *   last thing the model reads.
+ * - A tool-result turn (the message carries a `ToolResultBlock`): the text is **appended**,
+ *   because providers require the tool result to be the first content block in the turn that answers a
+ *   tool use.
+ *
+ * {@link Message} fields are readonly, so the target is rebuilt as a new {@link Message}. When there is
+ * no `user` message, the input array is returned unchanged.
  *
  * @param messages - The conversation to fold into
- * @param text - The text to prepend to the most recent user message
+ * @param text - The text to fold into the most recent user message
  * @returns A new array with the folded message, or the input array when there is no user message
  * @internal Delivery primitive. Reach injection through `ContextInjector` or `MemoryManager`.
  */
@@ -124,9 +130,14 @@ export function foldIntoLastUserMessage(messages: Message[], text: string): Mess
   }
 
   const target = messages[targetIndex]!
+  const injected = new TextBlock(text)
+  // A tool result must stay the first block in the turn that answers a tool use, so append rather than
+  // prepend when the target carries one.
+  const hasToolResult = target.content.some((block) => block.type === 'toolResultBlock')
+  const content = hasToolResult ? [...target.content, injected] : [injected, ...target.content]
   const folded = new Message({
     role: target.role,
-    content: [new TextBlock(text), ...target.content],
+    content,
     ...(target.metadata !== undefined && { metadata: target.metadata }),
   })
 
--- a/strands-ts/src/injection/types.ts
+++ b/strands-ts/src/injection/types.ts
@@ -24,8 +24,6 @@ export interface InjectionContext {
   messages: MessageData[]
   /** Durable app state shared across calls, hooks, and tools — read what a tool stashed last turn. */
   appState: StateStore
-  /** The run's cancellation signal; forward it to async I/O in `renderContent` so it aborts with the run. */
-  signal: AbortSignal
   /** The agent the injection is attached to (escape hatch for advanced consumers). */
   agent: LocalAgent
 }
--- a/strands-ts/src/injection/types.ts
+++ b/strands-ts/src/injection/types.ts
@@ -1,4 +1,6 @@
 import type { MessageData } from '../types/messages.js'
+import type { LocalAgent } from '../types/agent.js'
+import type { StateStore } from '../state-store.js'
 
 /**
  * Determines when injection runs before a model call.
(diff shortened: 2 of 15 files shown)


strands-ts/src/memory/types.ts
```diff
@@ -145,6 +143,44 @@ export interface MemoryAddToolConfig extends MemoryToolConfig {
...
+ * how many entries to retrieve, how to derive the query, and how to render the results. These are
+ * memory concerns and intentionally do not live on the generic injection engine.
+ */
+export interface MemoryInjectionConfig extends InjectionConfig {
+  /**
+   * Maximum number of entries to retrieve and inject per model call.
+   *
+   * @defaultValue 1
+   */
+  maxInjectedSearchResults?: number
```
zastrowm: This is wordy; did you consider any alternatives?

strands-ts/src/memory/types.ts
```diff
@@ -158,4 +194,9 @@ export interface MemoryManagerConfig {
    * writable stores; pass a {@link MemoryAddToolConfig} with `stores` to restrict it to specific ones.
    */
   addToolConfig?: MemoryAddToolConfig | boolean
+  /**
+   * Memory context injection. Defaults to `false` (opt-in). `true` uses the default injection
+   * settings; pass a {@link MemoryInjectionConfig} to customize retrieval, timing, and formatting.
+   */
+  injection?: boolean | MemoryInjectionConfig
```
zastrowm, line 265: In the PR description (and maybe as a comment) can you indicate what `true`/the-default is?

E.g. 

```
`true` is equivalent to {
  ...
}
```

strands-ts/src/injection/message-injection.ts
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * `'userTurn'` maps to {@link isUserTurn}; `'everyTurn'` to an always-true gate; a user-supplied
+ * predicate is wrapped so that a throw fails open (logs and skips injection rather than aborting the
+ * model call).
+ *
+ * @param trigger - An {@link InjectionTrigger} name, a predicate, or `undefined` (defaults to `'userTurn'`)
+ * @returns A predicate that, given the current messages, returns whether to inject this call
+ */
+export function resolveTrigger(
+  trigger: InjectionTrigger | ((messages: MessageData[]) => boolean) | undefined
```
zastrowm: What's the use case for the predicate?

strands-ts/src/memory/types.ts
```diff
@@ -145,6 +143,44 @@ export interface MemoryAddToolConfig extends MemoryToolConfig {
...
+   */
+  query?: (messages: MessageData[]) => string | undefined
+  /**
+   * Renders retrieved entries into the injected text. A callback that throws fails open (injection is
+   * skipped).
+   *
+   * Defaults to a `<memory>` XML block with one `<entry>` per result, carrying a `source` attribute
+   * naming the originating store (when known) so the model can attribute and weigh each memory.
+   */
+  format?: (entries: MemoryEntry[]) => string
```
zastrowm: What's a more complete example of this?  Do we expect folks to generate that XML themselves?  (could you put the more complete example in the PR description)

strands-ts/src/memory/types.ts
opieter-aws: `maxEntries` is probably a better alternative.

`maxResults` is too close to the existing `maxSearchResults` for the `search` tool on the MemoryManager.

strands-ts/src/memory/types.ts
opieter-aws:
```
   * 
   * {
   *   trigger: 'userTurn',          // inject only on a fresh user ask
   *   maxResults: 5,                // inject the top 5 results
   *   // query:  latest user text on trigger `userTurn`, else last assistant text
   *   // format: a <memory> block with one `<entry source="STORE_NAME">` per result
   * }
   * 
```

strands-ts/src/injection/message-injection.ts
opieter-aws: Custom injection trigger. E.g. inject only every Nth turn, only once the conversation exceeds a certain length, gate on something in `messages` (a tool was just used, a specific topic appeared, user keyword like "memory").

strands-ts/src/memory/types.ts
opieter-aws: No, users don’t generate the XML themselves: the default produces the <memory><entry source="..."> block. `format` is just the override: you get the typed MemoryEntry[] and return any string. E.g.:

Use it to match the system-prompt style, drop markup for cheaper models, frame memories as background rather than authoritative facts, or expose score/recency so the model can discount weak matches

Added example to PR description

strands-ts/src/memory/types.ts
```diff
@@ -145,6 +143,44 @@ export interface MemoryAddToolConfig extends MemoryToolConfig {
...
+   */
+  maxInjectedSearchResults?: number
+  /**
+   * Derives the search query from the current conversation. Return `undefined` or an empty string to
+   * skip injection for this call. A callback that throws fails open (injection is skipped).
+   *
+   * Defaults to an adaptive query: the latest user message's text on a user turn, otherwise the most
+   * recent assistant message's text (the previous step on an autonomous turn).
+   */
+  query?: (messages: MessageData[]) => string | undefined
```
mkmeral: can we provide strategies where we use SLMs or smaller LLMs to generate this? e.g. I send `?` to my agents a lot, and this default wouldn't be helpful.

Not P0 though

agent-of-mkmeral:
## Maintainer decisions on the API BR conditions

Thanks for the BR pass. We've made calls on the open conditions — here's where we land, with rationale. Net: **4 changes requested before merge** (3 were already flagged as blocking; #4 is a promotion of a recommended item to required).

<details>
<summary><b>1. Default <code>maxInjectedSearchResults</code> → <code>5</code> (not <code>1</code>)</b></summary>

**Decision: change the default to `5`.** (Resolves the doc/code mismatch in favor of the PR text, not the code.)

The reasoning is about *what the KB actually gives us*. A memory store does **semantic vector search** — it ranks by embedding similarity, i.e. *semantic* relevance. But semantically relevant ≠ **contextually** relevant. The top-1 hit is the single most embedding-similar entry, which is not reliably the most *useful* entry for the current turn. Defaulting to `1` makes the whole feature lean entirely on the store's ranking being right on the first try, which is a fragile bet for a vector index.

Returning the top **5** gives the model a materially better shot at having the contextually-right memory somewhere in the injected set, and the model is good at picking the relevant one out of a small candidate list. The tradeoff is acknowledged: more entries = more context bloat. `5` is the balance point — enough recall to not depend on perfect ranking, small enough to keep the prepend cheap.

**Action items:**
- `DEFAULT_INJECTED_SEARCH_RESULTS = 5`
- Update the tests that assert `1` → `5`
- Make TSDoc + PR body + code all agree on `5`
- Note in the TSDoc the bloat tradeoff so users who want a tighter prepend know to lower it

This also softens condition #4 from the BR (multi-store store-order bias): with `5` you're far less likely to get a single store[0]-biased result. Still worth fixing the global ranking eventually, but it's no longer as acute at the default.
</details>

<details>
<summary><b>2. Escape XML by default in <code>_defaultInjectionFormat</code></b></summary>

**Decision: yes — the default formatter must escape.** This was blocking and stays blocking.

The default renders `<entry source="${storeName}">${content}</entry>` with raw interpolation. Memory content is frequently user-derived, so:
- **Structural break:** content containing `</entry>`/`</memory>`/`"` breaks the block the model sees.
- **Stored-prompt-injection surface:** raw user-derived text folded straight into model input is an injection vector.

Escape `&`, `<`, `>` in text content and additionally `"` (and ideally `'`) inside attribute values (`source`). Order matters — escape `&` first. A custom `format` remains the user's responsibility, but the **shipped default must be safe**.
</details>

<details>
<summary><b>3. Fix the export gap</b></summary>

**Decision: fix it.** The config types must be importable by name (Tenet 5).

Right now:
```ts
import type { MemoryInjectionConfig } from '@strands-agents/sdk'        // ❌
import type { MemoryInjectionConfig } from '@strands-agents/sdk/memory' // ❌
```
`MemoryInjectionConfig` / `InjectionConfig` / `InjectionTrigger` are re-exported from `memory/index.ts`, but the **root `index.ts` type re-export block doesn't list them**, and there's no `./memory` (or `./injection`) subpath in `package.json#exports`. A customer can't write a typed `const injection: MemoryInjectionConfig = {…}` — defeating the point of the public config surface for both humans and agents/IDEs.

**Action:** add `MemoryInjectionConfig`, `InjectionConfig`, `InjectionTrigger` to the root `index.ts` `export type { … } from './memory/index.js'` block. (`MessageData`, which the `query`/`format` signatures need, is already at root — good.)
</details>

<details>
<summary><b>4. Callback forward-compat — promote to required</b></summary>

**Decision: +1, and we want this done now**, not deferred. Moving it from "recommended" to a merge condition.

`query`, `format`, and `provide` currently take a single positional `messages` arg. The first time we need to pass anything else (resolved trigger, agent ref, store list, abort signal) we break **every** customer callback — and that's a major-version break for a feature that's brand new. Cheap to get right now:

```ts
query?: (ctx: { messages: MessageData[] }) => string | undefined
format?: (ctx: { entries: MemoryEntry[] }) => string
// provide likewise takes a single context object
```

Wrap the args in a context object now so we can grow it additively later. Please also keep this shape in mind for the **Python** memory-injection equivalent so the callback contracts don't diverge across SDKs.
</details>

---

**Not in scope for this round** (tracked, not blocking): the silent no-op warning until #1068 lands, and the `maxInjectedSearchResults` naming nit. We can revisit naming before the runtime delivery PR.

cc @opieter-aws — once 1–4 are in we should be good to merge.

strands-ts/src/memory/memory-manager.ts
```diff
@@ -28,6 +31,9 @@ const ADD_TOOL_DESCRIPTION =
  */
 export const DEFAULT_MAX_SEARCH_RESULTS = 3
 
+/** Default number of entries injected per model call when injection does not specify one. */
+const DEFAULT_INJECTED_SEARCH_RESULTS = 1
```
mkmeral: qq, are these defaults good? for injected results and also max search results above? Lower the number, the stricter dependency you create on ranking of those vector stores. 

Can we increase this? What would be a good number? I guess one annoying thing is, we don't necessarily know the size of these results, right?

strands-ts/src/memory/types.ts
mkmeral: should we start documenting these `true` behaviors in docstring? that would also trickle down to api references in the docs 

(not just this PR, but overall. we have been using this pattern more and more)

agent-of-mkmeral:
## Proposal: `ContextProvider` is the public shape — land it in *this* PR

**TL;DR:** mechanism is good. But `createInjectionMiddleware` as an exported free function is a *third* extension idiom next to plugins + hooks, with zero non-memory consumers — a "random helper" we shouldn't ship if we won't keep it. So: **commit to `ContextProvider` as a vended plugin now**, make `createInjectionMiddleware` **internal** from day one, and let memory's `injection: true` be its first consumer. This also resolves the open `provide`/`escape`/`appState`/stacking questions in one place. Details below.

<details>
<summary><b>Why a plugin, and why now (not a follow-up)</b></summary>

- Every other extension ships via `plugins: […]` (`ContextOffloader`, `AgentSkills`, memory). A free `createInjectionMiddleware` + `addMiddleware` + `InvokeModelStage` is a third idiom customers must learn for no reason.
- If we export `createInjectionMiddleware` publicly now, we either keep it forever or break people later. Don't release a helper we won't stand behind. Make it **internal plumbing** behind `ContextProvider` from the start.
- The plugin is the natural **owner of coalescing/ordering/budget** across stacked providers — a per-middleware design can't be, since each middleware is blind to the others.
- Memory doesn't regress: `injection: true` keeps working, just implemented as a `ContextProvider` under the hood.
</details>

<details>
<summary><b>Proposed API</b></summary>

```ts
import { ContextProvider } from '@strands-agents/sdk/vended-plugins/context-provider'

new Agent({ model, plugins: [
  new ContextProvider({ name: 'now', provide: async () => `<now>${new Date().toISOString()}</now>` }),
  new ContextProvider({ name: 'sandbox', trigger: 'everyTurn', provide: async (ctx) => describeSandbox(ctx.appState) }),
  new ContextProvider({ name: 'editor', provide: async (ctx) => renderEditor(ctx) }),
]})
```

```ts
export interface ContextProviderConfig {
  name?: string
  trigger?: InjectionTrigger | ((ctx: ProvideContext) => boolean)   // default 'userTurn'
  provide: (ctx: ProvideContext) => Promise<string | undefined>      // ''/undefined => skip; throws => fail open
  maxTokens?: number                                                 // shared-budget hint (see coalescing)
}
```

`class ContextProvider implements Plugin` — `name` + `initAgent(agent)` registers the internal injection middleware. Same lifecycle as `ContextOffloader`.
</details>

<details>
<summary><b>ctx-bag: pass <code>appState</code> + signal + agent (not just messages)</b></summary>

The single biggest forward-compat lever. `provide`/`trigger` get a **context object**, not a positional `messages`:

```ts
export interface ProvideContext {
  messages: MessageData[]
  appState: StateStore   // durable cross-call store on LocalAgent
  signal?: AbortSignal   // so async provide() I/O (sandbox RPC, RAG) is cancellable on abort
  agent: LocalAgent
}
```

- **`appState`** lets an injector read what a tool stashed last turn without a closure — makes providers composable with the rest of the agent, not just self-contained closures. (This is the +100 ask.)
- **`signal`** matters because non-memory providers (sandbox/identity/RAG) do real network I/O that should cancel on run-abort. Memory doesn't feel this; sandbox does.
- Wrapping in a bag now means we grow it additively later — no major-version break. This is the same condition as #4 in the [conditions comment](https://github.com/strands-agents/harness-sdk/pull/2631#issuecomment-4657252596), unified here.
</details>

<details>
<summary><b>Escaping &amp; <code>format</code>: consumer contract, not engine behavior</b></summary>

Agreed: the **engine does NOT escape**. `provide` returns **opaque text** — the engine can't know if it's XML, Markdown, JSON, or prose, so it has no basis to escape. Escaping only made sense in *memory* because memory owns its `<memory>` format (that's why memory's default formatter still escapes — condition #2).

So at the `ContextProvider`/engine layer:
- No auto-escaping. Escaping is the provider author's responsibility.
- Ship an **optional `escapeXml` helper** next to the plugin for the consumers who do choose XML — convenience, never auto-applied.
- Loud TSDoc warning: provider output is folded raw into model input = stored-prompt-injection surface; escape attacker-influenced fields yourself.

The split: memory has a `format` (knows its structure → escapes); generic providers return raw text (own their structure → own their escaping).
</details>

<details>
<summary><b>Stacking: how multiple providers compose (the gap the plugin closes)</b></summary>

Today N injectors stack *natively* but only via middleware order — each calls `foldIntoLastUserMessage` independently, so you get N separate prepends ahead of the user ask, order = registration order, no dedup, no shared budget. That's "works but undefined."

`ContextProvider` owns this explicitly:
- **Ordering:** a documented priority/registration order, not implicit plugin-array luck.
- **Coalescing:** merge all providers' output into ONE context region instead of N scattered `TextBlock`s shoved ahead of the ask.
- **Budget:** `maxTokens` hint so the plugin can enforce a shared cap across providers — no single injector can be blind to the total.
</details>

<details>
<summary><b>Scope / sequencing</b></summary>

The [4 merge conditions](https://github.com/strands-agents/harness-sdk/pull/2631#issuecomment-4657252596) still stand (export gap, default 1→5, escape memory's formatter, ctx-bag). On top of those, this PR (or a tightly-coupled companion before release) should: **vend `ContextProvider`**, keep `createInjectionMiddleware` **un-exported/internal**, and define the coalescing/ordering/budget contract — so we don't ship a public helper we'd have to deprecate. Memory ships as consumer #1; a tiny `now`/sandbox provider validates the abstraction (Tenet 2).
</details>

@opieter-aws — proposing we make `ContextProvider` the committed public shape in this release rather than a follow-up. Can sketch the full plugin + coalescing impl if we're aligned.

mkmeral:
@opieter-aws what do you think about context provider as a plugin? I am hesitant to add a random method. I'm not a fan of the current single method tbh

strands-ts/src/memory/types.ts
opieter-aws: Yes I agree it would make sense to put some character guard on this at least. I'll discuss with the science team, since the current industry standard is to just use the latest message. 

Would be interesting to do some benchmarking here!

strands-ts/src/memory/memory-manager.ts
opieter-aws: Correct, we do not know the size. 

Out of my benchmarking came that 1 result performs best, but science came down to 5.

I'm planning to ship this with `5` as default, and ask science to run benchmarks to optimize.

opieter-aws:
> @opieter-aws what do you think about context provider as a plugin? I am hesitant to add a random method. I'm not a fan of the current single method tbh

That makes sense. I'm happy to provide as a plugin, but IMO `ContextProvider` is semantically too close to our `ContextManager` primitive. What about `MessageInjector`?

opieter-aws:
@agent-of-mkmeral give this PR another review

agent-of-mkmeral (approved):
## Re-review at `569c7d5` — approving ✅

**TL;DR:** All [4 merge conditions](https://github.com/strands-agents/harness-sdk/pull/2631#issuecomment-4657252596) met, [plugin proposal](https://github.com/strands-agents/harness-sdk/pull/2631#issuecomment-4657452392) delivered as `ContextInjector` with the middleware kept internal. Verified locally: tsc/lint clean, 154/154 tests pass. The CI failure is only the size-label gate, not a code failure. Two non-blocking doc nits below. LGTM 🚀

<details>
<summary><b>Conditions check (4/4 ✅)</b></summary>

| # | Condition | Status |
|---|-----------|--------|
| 1 | Default → **5** entries | ✅ `DEFAULT_MAX_ENTRIES = 5` in `memory-manager.ts`, TSDoc `@defaultValue 5` with the bloat tradeoff documented, PR body agrees, and the `true`-equivalence docstring on `MemoryManagerConfig.injection` spells it out. Doc/code mismatch resolved. |
| 2 | Escape XML in default format | ✅ `_defaultInjectionFormat` uses `escapeXmlText` for content and `escapeXmlAttr` for `source` (incl. `"`/`'`, `&` first). Adversarial test covers a `</entry>` + quote breakout. Custom `format` owns its escaping, with the `escapeXml` helper exported for that. |
| 3 | Export gap | ✅ Root `index.ts` now exports `MemoryInjectionConfig`, `InjectionConfig`, `InjectionTrigger`, `InjectionContext`. I wrote a typed import smoke test against the root entry — all names resolve. `./vended-plugins/context-injector` subpath added to `package.json#exports` for `ContextInjector`/`ContextInjectorConfig`/`escapeXml`. |
| 4 | Callback ctx-bag | ✅ `query: ({ messages }) => …`, `format: ({ entries }) => …`, and `renderContent(ctx: InjectionContext)` where the context bag carries `messages` + `appState` + `signal` + `agent` — the full +100 ask, including the abort signal for cancellable provider I/O. Tests assert `appState`/`signal` are exposed. |

</details>

<details>
<summary><b>Plugin proposal check</b></summary>

- ✅ **Vended plugin shipped** as `ContextInjector` — reasonable landing spot for the `ContextProvider`-too-close-to-`ContextManager` concern. Same lifecycle as `ContextOffloader`, registered via `plugins: […]`.
- ✅ **`createInjectionMiddleware` is internal** — marked `@internal`, not exported from `injection/index.ts` (only the three types are). No third extension idiom shipped publicly. 👏
- ✅ Memory is consumer #1: `injection: true` routes through the same engine via `_provideMemoryContext`.
- ✅ Duplicate-name behavior is sane: registry throws loudly on a name collision, and the TSDoc tells you to set distinct `name`s for multiple injectors.
- ✅ `maxInjectedSearchResults` → `maxEntries` naming nit also resolved.
- ⏸️ **Coalescing/shared budget for stacked injectors** — not implemented; behavior is now at least *defined* (independent prepends, plugin-registration order, per the `@remarks`). Fine to defer.

</details>

<details>
<summary><b>Local verification (at <code>569c7d5</code>)</b></summary>

- `tsc --noEmit` clean; `npm run lint` clean (the 1 warning is a pre-existing unused eslint-disable in `agent.context-manager.test.ts`, untouched by this PR)
- 154/154 tests pass across `injection/`, `vended-plugins/context-injector/`, `memory/` — incl. fail-open paths (throwing `trigger`/`query`/`format`/`renderContent`), no-durable-mutation, tool-result-turn folding, and the XML breakout case
- Fail-open holds transitively: a throwing **store search** inside `_provideMemoryContext` is caught by the middleware's `renderContent` try/catch — model call proceeds

</details>

<details>
<summary><b>Non-blocking nits</b></summary>

1. **PR body drift:** the `ContextInjectorConfig` snippet in the description lists `maxTokens?: number`, which doesn't exist in the code — please drop it from the body (or note it as future budget work) so we don't re-create a doc/code mismatch. The intro paragraph also says consumers supply a "`provide(context)`" callback while the actual field is `renderContent` — worth a quick wording pass.
2. **CI `label-size` failure** is the size gate (1149 lines > XL threshold), not a code failure — unit/lint/typecheck are green. Maintainer override or label exemption needed at merge time.
3. Tracked, not blocking (carried over): silent no-op warning until #1068; global cross-store ranking for multi-store injection (softened by the 5-entry default).

</details>

agent-of-mkmeral:
## ⚠️ Post-approval finding: `everyTurn` is empirically broken for tooled agents on Anthropic/Bedrock

**TL;DR:** My [approval at `569c7d5`](https://github.com/strands-agents/harness-sdk/pull/2631#pullrequestreview-3514103467) stands for the **default `'userTurn'` path**, but I've now **reproduced against a real provider** the tool-result ordering issue the automated review flagged: with `trigger: 'everyTurn'`, the fold prepends a `TextBlock` ahead of the `toolResult` block, and Bedrock (Anthropic models) **rejects the request** with a `ValidationException`. This elevates that thread from "potential risk" to **verified bug** — I'd treat it as blocking for the `everyTurn` value (fix or document-and-restrict before merge).

<details>
<summary><b>Empirical repro (real Bedrock Converse call, not a mock)</b></summary>

I replayed the exact message shape `foldIntoLastUserMessage` produces on an `everyTurn` tool-result turn against `us.anthropic.claude-sonnet-4-5` via Bedrock Converse:

```
A: user content = [toolResult, injectedText]  → ✅ SUCCESS ('2+2 = 4')
B: user content = [injectedText, toolResult]  → ❌ ValidationException:
   "messages.2: `tool_use` ids were found without `tool_result` blocks immediately
    after: tu-1. Each `tool_use` block must have a corresponding `tool_result`
    block in the next message."
```

Case B is exactly what ships: `foldIntoLastUserMessage` block-**prepends** the injected text into the most recent `user` message, and in the tool loop that message is the tool-result turn. The TS Bedrock provider (`_formatContentBlock`) preserves block order — no reordering safety net.

</details>

<details>
<summary><b>Blast radius</b></summary>

- **Default path (`'userTurn'`) is safe** — `isUserTurn` excludes tool-result turns by construction. ✅
- **`trigger: 'everyTurn'` breaks the first model call after any tool use** for Anthropic models on Bedrock (and the same constraint exists on Anthropic direct). Since `everyTurn` is documented *specifically for autonomous agents* — i.e. agents that use tools — the documented use case is the broken one.
- The PR description's own `ContextInjector` example (`new ContextInjector({ name: 'now', trigger: 'everyTurn', … })`) would crash the tool loop of any tooled agent.
- Predicate triggers that return `true` on a tool-result turn hit the same path.
- The existing test (`message-injection.test.ts` ~L192, *"'everyTurn' injects on an autonomous tool-result turn"*) asserts the **broken ordering is produced** — it passes against a mock and encodes the bug as expected behavior.

</details>

<details>
<summary><b>Suggested fix (validated)</b></summary>

My Case A repro shows **text after the `toolResult` block is accepted**. So the minimal fix is position-aware folding in `foldIntoLastUserMessage`:

```ts
// If the target user message contains a toolResult block, append the injected
// TextBlock after the existing content instead of prepending.
const hasToolResult = target.content.some((b) => 'toolResult' in b /* or type check */)
const content = hasToolResult
  ? [...target.content, new TextBlock(text)]
  : [new TextBlock(text), ...target.content]
```

This keeps the documented "user's ask stays in the recency slot" property on plain user turns (prepend) while keeping `tool_result`-first ordering on tool-result turns (append). Plus: flip the L192 test to assert the appended ordering, and ideally one integ test driving `everyTurn` through a real provider call.

If you'd rather defer: restrict `everyTurn` to skip tool-result turns for now and document it — but silently producing requests a major provider rejects shouldn't ship as-is.

</details>

cc @opieter-aws @mkmeral — sorry for the post-approval flag; the automated review's inline comment on `message-injection.ts` deserved a live-provider check and it confirmed.

agent-of-mkmeral (approved):
## Re-review at `c1690ec` — approving ✅

**TL;DR:** The `everyTurn` tool-result ordering bug I [flagged post-approval](https://github.com/strands-agents/harness-sdk/pull/2631#issuecomment-4668083127) is fixed — and I **validated the fix against real Bedrock** (the fixed `[toolResult, injectedText]` shape succeeds where the old one threw `ValidationException`). All other deltas in `fix: address comments` check out: `escapeXml` is now attribute-safe, store-order bias documented, tests upgraded to whole-object assertions. Verified locally: tsc clean, 154/154 tests pass. One naming thread (`ContextInjector` vs `MessageInjector`) remains a maintainer design call, not a defect. LGTM 🚀

<details>
<summary><b>everyTurn fix — empirically validated (real Bedrock Converse call)</b></summary>

`foldIntoLastUserMessage` is now position-aware: it **appends** the injected `TextBlock` after a `ToolResultBlock` and only **prepends** on a plain user ask — exactly the validated fix shape from my finding. I replayed the new output shape against `us.anthropic.claude-sonnet-4-5` via Bedrock Converse:

```
user content = [toolResult, injectedText]  → ✅ SUCCESS ('The answer is 4')
```

This is the ordering that previously failed with `tool_use ids were found without tool_result blocks immediately after`. The L192 test was flipped to assert the appended ordering (it previously encoded the broken prepend as expected behavior), and the TSDoc on the fold documents both placements and why. Predicate triggers returning `true` on a tool-result turn ride the same safe path. ✅

</details>

<details>
<summary><b>Other deltas since <code>569c7d5</code></b></summary>

- ✅ **`escapeXml` → attribute-safe**: now aliases `escapeXmlAttr` (escapes `&`, `<`, `>`, `"`, `'`), so one public helper is safe in both element content and attributes — closes the round-1 attribute-breakout gap for custom formatters building `source="…"`-style XML. Test updated accordingly.
- ✅ **Multi-store bias documented**: `maxEntries` TSDoc now states results concatenate in store-registration order with no cross-store ranking. Acceptable minimum (softened by the 5-entry default); global re-rank stays a tracked follow-up.
- ✅ **Test hygiene**: injection/plugin/memory tests moved to whole-object `toStrictEqual` per TESTING.md — these now catch extra/missing blocks the per-field checks missed.
- ✅ **PR body drift fixed**: `maxTokens` removed from the `ContextInjectorConfig` snippet; config shows `renderContent`. (Tiny residual: the intro paragraph still says consumers supply a "`provide(context)`" callback — one-word wording nit, not blocking.)

</details>

<details>
<summary><b>Local verification (at <code>c1690ec</code>)</b></summary>

- `tsc --noEmit --project src/tsconfig.json` clean
- 154/154 tests pass across `injection/`, `vended-plugins/context-injector/`, `memory/`
- Bedrock Converse live-call validation of the fixed fold ordering (above)
- CI: all unit/lint/typecheck/security checks green across Node 20/22/24 × 3 OSes; only `label-size` fails (size gate, 1181 lines > XL threshold) — needs maintainer override at merge time

</details>

<details>
<summary><b>Open, non-blocking</b></summary>

1. **Naming**: @opieter-aws proposed `MessageInjector` over `ContextProvider` (too close to `ContextManager`); shipped name is `ContextInjector`. Reasonable middle ground IMO — "context" describes what's injected, "injector" disambiguates from `ContextManager`'s lifecycle role. Flagging for @mkmeral to confirm before merge since it's a public vended-plugin name.
2. Carried over: silent no-op warning until #1068; global cross-store ranking; private-access casts in a few memory tests; docs PR still `todo`.

</details>

strands-ts/src/injection/xml.ts
```diff
@@ -0,0 +1,40 @@
...
+  return escapeXmlText(value).replace(/"/g, '&quot;').replace(/'/g, '&#39;')
+}
+
+/**
+ * Escapes `&`, `<`, `>`, `"`, and `'` so a value is safe in XML **anywhere** — both element content and
+ * attribute values. Exposed to consumers (e.g. a {@link ContextInjector} renderer) who render XML
+ * themselves and want one helper that is safe regardless of where the value lands. Never applied
+ * automatically by the injection engine — escaping is the renderer's responsibility.
+ */
+export const escapeXml = escapeXmlAttr
```
mkmeral: should we have this in our exports? why? I am hesitant to add orphaned basic formatting methods

strands-ts/src/injection/types.ts
```diff
@@ -0,0 +1,68 @@
...
+ * The context an injection consumer receives on each model call, passed to `renderContent` and to a
+ * predicate {@link InjectionConfig.trigger}.
+ */
+export interface InjectionContext {
+  /** The current conversation, as data. */
+  messages: MessageData[]
+  /** Durable app state shared across calls, hooks, and tools — read what a tool stashed last turn. */
+  appState: StateStore
+  /** The run's cancellation signal; forward it to async I/O in `renderContent` so it aborts with the run. */
+  signal: AbortSignal
```
mkmeral: do we need this for injection context? is this about cleaning up afterwards? 🤔 what's the use case for it?

agent-of-mkmeral:
✅ **Re-verified at `c1690ec` ("fix: address comments") — the `everyTurn` tool-result ordering fix is confirmed against a live provider.**

What I re-ran:
- `tsc --noEmit` on `src/`: clean.
- 113/113 unit tests pass (injection + context-injector + memory-manager suites).
- Replayed `foldIntoLastUserMessage` from the PR source on a tool-result turn: it now emits `[toolResult, injectedText]` (append) instead of `[injectedText, toolResult]` (prepend).
- Sent that exact message shape to real Bedrock Converse (`us.anthropic.claude-sonnet-4-5`) with a `toolConfig`: **accepted, no `ValidationException`**. The repro from my earlier comment no longer applies.

Also confirming the two follow-ups landed in the same commit:
- `escapeXml` now aliases the attr-safe variant (escapes `"`/`'` too) — safe in element content *and* attributes. Note this changes its output for quote-containing input; fine pre-merge, would have been breaking after.
- `maxEntries` TSDoc now documents the multi-store store-order concatenation (no cross-store ranking) — good "document the limitation" resolution.

One residual cosmetic nit: the PR body intro still says "`provide(context)`" and the Testing section references `ProvideContext` / "throwing `provide`" — the shipped names are `renderContent` / `InjectionContext`. Worth a quick body edit before merge so the description matches the code.

From my side all blocking conditions are resolved. The only open thread is the `ContextInjector` vs `MessageInjector`/`ContextProvider` naming call — my vote stays with `ContextInjector` (names the purpose, not the mechanism; keeps the offload↔inject symmetry with `ContextOffloader`; avoids the `ContextManager` collision)._

strands-ts/src/injection/types.ts
opieter-aws: Use case is a `renderContent` callback doing real async I/O (e.g. RAG) that should cancel when the run aborts. But it’s not load-bearing today and reachable via `ctx.agent.cancelSignal` anyway. Fine with me to remove now, can add later if a use case arises

strands-ts/src/injection/xml.ts
opieter-aws: Agreed, let’s drop it from the public exports.

mkmeral (approved):

PR #2632, opened 2026-06-04
docs: add AgentCore Harness model provider design doc

gautamsirdeshmukh:
## Type of Change

Bug fix
New feature
Breaking change
Documentation update
Other (please describe):

## Testing

## Checklist

----

### DIFF
--- a/team/designs/0012-harness-as-model-provider.md
+++ b/team/designs/0012-harness-as-model-provider.md
@@ -1,0 +1,441 @@
+# AgentCore Harness as a Strands Model Provider
+
+**Status**: Proposed
+
+**Date**: 2026-06-29
+
+**Issue**: [Prototype PR #2632](https://github.com/strands-agents/harness-sdk/pull/2632)
+
+**Scope**: Both Python and TypeScript SDKs
+
+## Table of Contents
+
+- [Context](#context)
+- [Proposal](#proposal)
+  - [How it Works](#how-it-works)
+  - [Translating the Call](#translating-the-call)
+  - [Tools](#tools)
+  - [Context and Memory](#context-and-memory)
+  - [Hooks](#hooks)
+- [Developer Experience](#developer-experience)
+- [Proposed AgentCore Changes](#proposed-agentcore-changes)
+- [Alternative Implementations](#alternative-implementations)
+- [Consequences](#consequences)
+- [Open Questions](#open-questions)
+- [Willingness to Implement](#willingness-to-implement)
+- [Appendix](#appendix)
+
+## Context
+
+Taking agents from prototype to production is a major focus across the industry, and this is exactly where teams are getting stuck today. "Managed agent" platforms have emerged as an answer: you hand the provider your agent's definition and it runs the loop for you, on production infrastructure you don't operate. 
+
+But when it comes to developers, these managed platforms all come with a tradeoff: the rich, extensible SDK must be left behind for raw API calls and low-level plumbing. An intuitive toolkit for building agents, OR the infrastructure that makes production possible... why should you have to choose?
+
+[AgentCore Harness](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/harness.html) has solved the hard infrastructure problem through microVM isolation, persistent memory, identity, observability and more for agents running in production. Developer experience is [Strands](https://strandsagents.com/)' calling card: typed streaming, hooks, multi-agent orchestration, and overall an intuitive framework that developers already know well.
+
+In short: developers already build with the Strands SDK, and the same framework is already running their agent in production (Harness). It's time to put the two pieces together.
+
+## Proposal
+
+We introduce a new `Model` provider: `AgentCoreHarness`. Developers shift an agent from local execution to managed infrastructure by changing one parameter.
+
+### How it Works
+
+The core of the agent loop runs inside the microVM, not on the host. On each turn, the host issues a single `InvokeHarness` request, and the Harness runs the real loop (calling the model, executing tools) while streaming the result back.
+
+```mermaid
+  sequenceDiagram
+      participant Dev as Developer
+      participant SDK as Strands SDK (host loop)
+      participant H as AgentCore Harness (microVM loop)
+      
+      Dev->>SDK: agent.invoke()
+      SDK->>H: InvokeHarness(messages, tools)
+      Note over H: model call → vended tool → model call
+      H-->>SDK: stream (text, vended tool use/results)
+      Note over H: model calls a host tool → inline_function
+      H-->>SDK: stream (toolUse, stopReason=tool_use)
+      SDK->>SDK: execute host tool (auto or HITL)
+      SDK->>H: InvokeHarness(toolResult)
+      H-->>SDK: stream (final response)
+      SDK-->>Dev: AgentResult
+```
+
+The host loop is a thin shell around this: it forwards the turn, relays the streamed events, and steps in only when a tool must run on the host. How tools are split between the microVM and the host is covered in [Tools](#tools) below.
+
+### Translating the Call
+
+The provider turns each turn of the Strands agent loop into one `InvokeHarness` request and maps the response stream back to standard model events.
+
+Almost everything in the request comes from the `AgentConfig` the developer already writes: the messages, system prompt, tools, and the model with its inference parameters. The provider itself adds only what identifies the managed agent to run, a `harnessArn` and a `sessionId`.
+
+| `InvokeHarness` field | From `AgentConfig` (unchanged) | New on `AgentCoreHarness` config |
+|---|---|---|
+| `messages` | ✓ conversation messages | — |
+| `systemPrompt` | ✓ `systemPrompt` | — |
+| `tools` | ✓ `tools` (classified, see [Tools](#tools)) | — |
+| `model`, `maxTokens`, `temperature`, `topP` | ✓ model config (Bedrock / OpenAI / Gemini / LiteLLM), carrying its own inference parameters | — |
+| `runtimeSessionId` | — | ✓ |
+| `harnessArn` | — | ✓ |
+
+The streamed response (message starts, content block deltas, tool-use and tool-result blocks, stop reasons, and usage) maps one-to-one onto the Strands model event types the agent loop already consumes. The full field-level mapping can be found in the [Appendix](#appendix).
+
+The Harness exposes further optional overrides which can be set via the provider config as well: per-user memory scoping (`actorId`), a named endpoint/version (`qualifier`), microVM-loaded skills, and the loop's safety limits (`allowedTools`, `maxIterations`, `timeoutSeconds`). These default sensibly and most developers never set them; the full set can also be found in the [Appendix](#appendix).
+
+### Tools
+
+Both custom and vended tools are passed to the Harness through the same `tools` array, but they run in two different places.
+
+**Custom tools** are the developer's own `tool()` functions, defined exactly as always: a schema and a callback that runs on the host.
+
+```typescript
+const getWeather = tool({
+  name: 'get_weather',
+  inputSchema: z.object({ city: z.string() }),
+  callback: async ({ city }) => fetchWeather(city),   // runs on the host
+})
+```
+
+The provider implementation translates these to `inline_function` definitions (name, description, schema, but no implementation). When the model calls one, the Harness pauses and hands the call back to the host; the host runs the callback (automatically or with a human in the loop), and a follow-up request resumes the loop with the result automatically.
+
+**Vended tools** carry no callback. They are references to capabilities the Harness runs itself in the microVM (code interpreter, browser, gateway, remote MCP), added with helpers like `codeInterpreter()` and `browser()` (see [Appendix](#appendix)).
+
+Each helper returns nothing more than a configuration object naming the capability, which the provider passes straight through into the `InvokeHarness` request; the Harness runs it in the sandbox and streams the call and result back.
+
+```typescript
+const agent = new Agent({
+  model: new AgentCoreHarness({ harnessArn, sessionId }),
+  tools: [
+    getWeather,          // custom → inline_function, runs on the host
+    codeInterpreter(),   // vended → runs in the microVM
+    browser(),           // vended → runs in the microVM
+  ],
+})
+```
+
+### Context and Memory
+
+The developer can choose how conversation context and memory are managed: let the Harness handle it, or bring their own. The `stateful` flag on the provider selects between the two. 
+
+`stateful` is an existing Strands flag (server-stateful providers like the Responses API already use it) and is purely an SDK-side switch. It controls whether the SDK keeps the conversation history and sends it each turn, or clears it and lets the Harness rebuild context server-side. 
+
+**Let the Harness manage it (`stateful: true`, the default)**
+
+Use this when the harness is deployed with memory enabled. 
+
+The Harness owns conversation state. It persists each turn server-side by `sessionId`, reloads history on the next invocation, and manages context and long-term memory itself, scoped per user with `actorId`. The provider sends only the new message each turn and clears local history; the developer configures nothing.
+
+**Bring your own (`stateful: false`)**
+
+Use this when the harness is deployed with memory disabled, so context is managed only on the host.
+
+The host owns conversation state, exactly as it does for a local agent. The agent's Strands conversation manager, context offloader, and compression tools run on the host and decide what the model sees each turn (trimming, summarizing, or offloading history), and the provider sends that host-managed context to the Harness. This is the path for developers who want Strands' context primitives instead of the Harness's built-in management.
+
+### Hooks
+
+All the same hook events still fire and stream to the host, whether the work ran on the host or inside the microVM. A hook can fully intervene in work run on the host, but can only observe work the microVM has already done (and "react" to it on the host side).
+
+**Actionable (host work):**
+- Tool-call hooks on custom tools: approve, cancel, rewrite input, swap the tool, or pause for human-in-the-loop
+- `BeforeModelCallEvent` / `AfterModelCallEvent`, bracketing each `InvokeHarness` request
+
+**Observe-only (microVM work):**
+- The model's streamed output, content blocks, and messages
+- Vended tool calls and their results (browser, code interpreter, gateway, remote MCP)
+
+Logging, tracing, and metrics therefore cover the whole agent, while interception applies only to the custom tools the host runs. We also propose `Before/AfterHarnessInvokeEvent` to bracket each `InvokeHarness` request explicitly, rather than overloading the model-call hooks.
+
+---
+
+## Developer Experience
+
+A local agent runs the loop on the host:
+
+```typescript
+import { Agent, BedrockModel } from '@strands-agents/sdk'
+
+const agent = new Agent({
+  model: new BedrockModel(),
+  systemPrompt,
+})
+```
+
+Swapping the model to `AgentCoreHarness` moves the same agent into a managed microVM. Nothing else changes:
+
+```typescript
+import { Agent } from '@strands-agents/sdk'
+import { AgentCoreHarness } from '@bedrock-agentcore/sdk'
+
+const agent = new Agent({
+  model: new AgentCoreHarness({ harnessArn, sessionId }),
+  systemPrompt,
+})
+```
+
+The full agent is defined through the Strands interface developers already use, including a mix of vended tools and custom tools that run on the host:
+
+```typescript
+import { Agent, tool } from '@strands-agents/sdk'
+import { AgentCoreHarness, codeInterpreter, browser, remoteMcp } from '@bedrock-agentcore/sdk'
+
+const getWeather = tool({ /* ... */ })
+
+const agent = new Agent({
+  model: new AgentCoreHarness({ harnessArn, sessionId }),
+  systemPrompt,
+  tools: [
+    getWeather,                        // custom tool, runs on the host (automated or HITL)
+    codeInterpreter(),                 // sandboxed code execution in the microVM
+    browser(),                         // cloud browser in the microVM
+    remoteMcp({ url: 'https://...' }), // remote MCP server, connected from the microVM
+  ],
+})
+```
+
+And it doesn't need to stop there. Developers can continue building up the harness around their agent, taking advantage of Strands' hooks, plugins and other mechanisms with the assurance that their agent loop and execution lives in a secure sandboxed environment.
+
+
+
+## Proposed AgentCore Changes
+
+The provider works against `InvokeHarness` as it exists today. The item below is about smoothing the iterate-from-the-SDK experience, not enabling the integration.
+
+### Provision the execution role for the full tool surface up front.** 
+
+The default harness execution role already covers Bedrock model invocation and the built-in browser and code interpreter, so those vended tools work immediately. Other capabilities each require an additional statement on the execution role: AgentCore Gateway, custom browser or code-interpreter resources, non-Bedrock model providers (which need API-key credential access), and skills fetched from S3 or Git. Adding one today means editing the role and re-running `agentcore deploy`. 
+
+If the role were provisioned with the broader permission set at creation, a developer could add or swap any of these purely through the Strands `AgentConfig`, passing them on each `InvokeHarness` call, without redeploying. This is what makes the "change the agent in code, not in config" experience hold for the full tool surface: deploy once, then iterate entirely from the SDK.
+
+
+## Alternative Implementations
+
+### 1. A dedicated `HarnessAgent` class
+
+A purpose-built `Agent` subclass could expose harness parameters as first-class arguments and handle context and memory coordination across the host/microVM directly (rather than leveraging the stateful flag).
+
+```typescript
+const agent = new HarnessAgent({
+  harnessArn,
+  sessionId,
+  model: { provider: 'bedrock', modelId: '...' },
+  tools: [getWeather, codeInterpreter()],
+})
+```
+
+- **Pro:** not bound by the `Model` interface's `stream()` contract, so it could expose harness-native capabilities a model provider cannot: dedicated methods for session control, filesystem and command access, or live browser interaction, plus per-invocation harness arguments as first-class parameters.
+- **Pro:** ability to evolve class separately from model providers
+- **Con:** more surface to build and own than a thin provider, and it gives up the one-parameter swap, since adopting it means switching to a new agent type rather than changing the model.
+
+This is a natural evolution if richer harness-native control is needed. The provider is the right first step: it delivers the core managed experience immediately and with the smallest change to existing code.
+
+### 2. A `ManagedHarness` plugin instead of a model provider
+
+The Harness integration could be a `Plugin` that intercepts the model call and routes it to `InvokeHarness`, leaving the agent with no model of its own.
+
+```typescript
+new Agent({
+  plugins: [new ManagedHarness({ harnessArn, sessionId })],
+  tools: [getWeather, codeInterpreter()],
+})
+```
+
+- **Pro:** harness parameters can be set per invocation through `invocationState`, not just statically on a constructor.
+- **Con:** the `Agent` must support construction with no `model`, a structural change to a core contract, since the plugin hijacks inference.
+- **Con:** configuration splits across the plugin and the agent, and a plugin that silently replaces the model is a surprising mental model for something that is, conceptually, just where inference happens.
+
+### 3. Reuse the existing OpenAI Responses provider
+
+Strands already has an `OpenAIModel` in Responses mode that can target the Bedrock Mantle endpoint, which supports stateful conversations and server-side tools through Lambda or AgentCore Gateway.
+
+```typescript
+new Agent({
+  model: new OpenAIModel({
+    api: 'responses',
+    stateful: true,
+    baseURL: 'https://bedrock-mantle.us-west-2.api.aws/v1',
+    modelId: 'openai.gpt-oss-120b',
+  }),
+})
+```
+
+- **Pro:** no new code; an existing provider already speaks a stateful, server-side-tool API.
+- **Con:** it is a different system. Mantle Responses offers stateful inference and server-side tools, but not the Harness agent loop: no vended browser or code interpreter, no skills, no managed or bring-your-own memory, no microVM filesystem, no inline-function human-in-the-loop, no mid-session model switching.
+- **Con:** it cannot emit `InvokeHarness`, so it delivers a thin slice of the value and none of what makes the Harness a managed agent.
+
+## Consequences
+
+### What becomes easier
+
+- Moving an agent from local execution to managed microVM infrastructure becomes a one-parameter change, with the rest of the `AgentConfig` (tools, system prompt, hooks, multi-agent setup) unchanged.
+- Developers reach the full Harness surface (vended browser, code interpreter, gateway, remote MCP, skills, managed memory) through the Strands interface they already know, instead of hand-writing `InvokeHarness` calls and parsing the event stream.
+- AgentCore gains an SDK entry point: a developer already in Strands can adopt managed infrastructure without leaving the framework.
+
+### What becomes harder or requires attention
+
+- The `stateful` flag must match how the harness was deployed (memory enabled vs. disabled). The SDK cannot verify this, so a mismatch silently manages context twice or loses it.
+- Per-invocation control of harness parameters is not available until `invocationState` is plumbed into the provider (see [Open Questions](#open-questions)).
+- The provider's behavior depends on harness-side details the SDK cannot see (the deployed memory configuration, the internal loop's streaming granularity), which need validation against a live harness.
+
+### Migration
+
+No migration is required. The provider is additive: existing agents and the other model providers are unaffected.
+
+## Open Questions
+
+- **Provider location and ownership**: Should `AgentCoreHarness` live in the Strands SDK alongside the other providers, or in the AgentCore SDK? As a thin `Model` provider it could be owned and maintained by the AgentCore team in their own SDK, matching how the integration is positioned. Living in Strands gives a single import and keeps it beside the other providers. (A `HarnessAgent` or plugin, by contrast, would more naturally live in Strands core.)
+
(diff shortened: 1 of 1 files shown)


team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+
+But when it comes to developers, these managed platforms all come with a tradeoff: the rich, extensible SDK must be left behind for raw API calls and low-level plumbing. An intuitive toolkit for building agents, OR the infrastructure that makes production possible... why should you have to choose?
+
+[AgentCore Harness](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/harness.html) has solved the hard infrastructure problem through microVM isolation, persistent memory, identity, observability and more for agents running in production. Developer experience is [Strands](https://strandsagents.com/)' calling card: typed streaming, hooks, multi-agent orchestration, and overall an intuitive framework that developers already know well.
+
+In short: developers already build with the Strands SDK, and the same framework is already running their agent in production (Harness). It's time to put the two pieces together.
+
+## Proposal
+
+We introduce a new `Model` provider: `AgentCoreHarness`. Developers shift an agent from local execution to managed infrastructure by changing one parameter.
```
mkmeral, line 40: why is it a new model provider? do we have pros/cons for a new class?

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+
+### Context and Memory
+
+The developer can choose how conversation context and memory are managed: let the Harness handle it, or bring their own. The `stateful` flag on the provider selects between the two. 
+
+`stateful` is an existing Strands flag (server-stateful providers like the Responses API already use it) and is purely an SDK-side switch. It controls whether the SDK keeps the conversation history and sends it each turn, or clears it and lets the Harness rebuild context server-side. 
+
+**Let the Harness manage it (`stateful: true`, the default)**
+
+Use this when the harness is deployed with memory enabled.
```
AlexanderRichey, line 124: Even in the case where memory is disabled, the harness will use in-memory state by default. So we'll probably want to make stateful default to true. Otherwise, you'd get duplicate messages in the conversation when calling InvokeHarness.

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+
+The host loop is a thin shell around this: it forwards the turn, relays the streamed events, and steps in only when a tool must run on the host. How tools are split between the microVM and the host is covered in [Tools](#tools) below.
+
+### Translating the Call
+
+The provider turns each turn of the Strands agent loop into one `InvokeHarness` request and maps the response stream back to standard model events.
+
+Almost everything in the request comes from the `AgentConfig` the developer already writes: the messages, system prompt, tools, and the model with its inference parameters. The provider itself adds only what identifies the managed agent to run, a `harnessArn` and a `sessionId`.
+
+| `InvokeHarness` field | From `AgentConfig` (unchanged) | New on `AgentCoreHarness` config |
```
poshinchen, line 72: Will the `AgentCoreHarness` be in AgentConfig or it's a separate one that retrieves the fields from AgentConfig?

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+const getWeather = tool({
+  name: 'get_weather',
+  inputSchema: z.object({ city: z.string() }),
+  callback: async ({ city }) => fetchWeather(city),   // runs on the host
+})
+```
+
+The provider implementation translates these to `inline_function` definitions (name, description, schema, but no implementation). When the model calls one, the Harness pauses and hands the call back to the host; the host runs the callback (automatically or with a human in the loop), and a follow-up request resumes the loop with the result automatically.
+
+**Vended tools** carry no callback. They are references to capabilities the Harness runs itself in the microVM (code interpreter, browser, gateway, remote MCP), added with helpers like `codeInterpreter()` and `browser()` (see [Appendix](#appendix)).
```
mkmeral, line 101: wait, do we vend a no-op SDK version of all the tools that managed harness may vend? why?

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+
+### What becomes easier
+
+- Moving an agent from local execution to managed microVM infrastructure becomes a one-parameter change, with the rest of the `AgentConfig` (tools, system prompt, hooks, multi-agent setup) unchanged.
+- Developers reach the full Harness surface (vended browser, code interpreter, gateway, remote MCP, skills, managed memory) through the Strands interface they already know, instead of hand-writing `InvokeHarness` calls and parsing the event stream.
+- AgentCore gains an SDK entry point: a developer already in Strands can adopt managed infrastructure without leaving the framework.
+
+### What becomes harder or requires attention
+
+- The `stateful` flag must match how the harness was deployed (memory enabled vs. disabled). The SDK cannot verify this, so a mismatch silently manages context twice or loses it.
```
yonib05, line 275: This is solvable from the agent core side, no?

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+Almost everything in the request comes from the `AgentConfig` the developer already writes: the messages, system prompt, tools, and the model with its inference parameters. The provider itself adds only what identifies the managed agent to run, a `harnessArn` and a `sessionId`.
+
+| `InvokeHarness` field | From `AgentConfig` (unchanged) | New on `AgentCoreHarness` config |
+|---|---|---|
+| `messages` | ✓ conversation messages | — |
+| `systemPrompt` | ✓ `systemPrompt` | — |
+| `tools` | ✓ `tools` (classified, see [Tools](#tools)) | — |
+| `model`, `maxTokens`, `temperature`, `topP` | ✓ model config (Bedrock / OpenAI / Gemini / LiteLLM), carrying its own inference parameters | — |
+| `runtimeSessionId` | — | ✓ |
+| `harnessArn` | — | ✓ |
```
lizradway, line 79: do we not see ourselves extending this possible to future harnesses? 

perhaps some sort of identifying `harnessId` rather than AWS specific `harnessARN` makes more sense here to not make breaking changes as we expand our support. same with `runtimeSessionId`.

I really like having Harnesses on the model provider, but am very suspicious of attaching these AWS concepts onto our high level construct.

Just in general, im curious about our extension plan.

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+
+Use this when the harness is deployed with memory enabled. 
+
+The Harness owns conversation state. It persists each turn server-side by `sessionId`, reloads history on the next invocation, and manages context and long-term memory itself, scoped per user with `actorId`. The provider sends only the new message each turn and clears local history; the developer configures nothing.
+
+**Bring your own (`stateful: false`)**
+
+Use this when the harness is deployed with memory disabled, so context is managed only on the host.
+
+The host owns conversation state, exactly as it does for a local agent. The agent's Strands conversation manager, context offloader, and compression tools run on the host and decide what the model sees each turn (trimming, summarizing, or offloading history), and the provider sends that host-managed context to the Harness. This is the path for developers who want Strands' context primitives instead of the Harness's built-in management.
```
mkmeral, line 132: who uses this? and why? doesn't this just make managed harness a pass-through for bedrock?

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+
+**Actionable (host work):**
+- Tool-call hooks on custom tools: approve, cancel, rewrite input, swap the tool, or pause for human-in-the-loop
+- `BeforeModelCallEvent` / `AfterModelCallEvent`, bracketing each `InvokeHarness` request
+
+**Observe-only (microVM work):**
+- The model's streamed output, content blocks, and messages
+- Vended tool calls and their results (browser, code interpreter, gateway, remote MCP)
+
+Logging, tracing, and metrics therefore cover the whole agent, while interception applies only to the custom tools the host runs. We also propose `Before/AfterHarnessInvokeEvent` to bracket each `InvokeHarness` request explicitly, rather than overloading the model-call hooks.
```
AlexanderRichey, line 146: The question of tracing is interesting. Harness will emit traces by default. If we have that on the client side too, will traces be double emitted? I wonder if we'd need to pass some relevant headers to get all of the spans to link up in the right way, e.g., linking a client side tool call to the correct InvokeHarness trace.

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+  harnessArn,
+  sessionId,
+  model: { provider: 'bedrock', modelId: '...' },
+  tools: [getWeather, codeInterpreter()],
+})
+```
+
+- **Pro:** not bound by the `Model` interface's `stream()` contract, so it could expose harness-native capabilities a model provider cannot: dedicated methods for session control, filesystem and command access, or live browser interaction, plus per-invocation harness arguments as first-class parameters.
+- **Pro:** ability to evolve class separately from model providers
+- **Con:** more surface to build and own than a thin provider, and it gives up the one-parameter swap, since adopting it means switching to a new agent type rather than changing the model.
```
pgrayy, line 227: Have you considered something like `class Agent implements HarnessAgent`? Then instead of switching to a new agent type, they could enable through configs.

team/designs/0012-harness-as-model-provider.md
yonib05: Also companies might want this. Think in the case for governance or policies. What you do with your agent in your application is your responsibility. But if a company shares an agent core harness that has a server side customer access tool, the harness should bear the responsibility of keeping track of how it is used.

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+import { AgentCoreHarness, codeInterpreter, browser, remoteMcp } from '@bedrock-agentcore/sdk'
+
+const getWeather = tool({ /* ... */ })
+
+const agent = new Agent({
+  model: new AgentCoreHarness({ harnessArn, sessionId }),
+  systemPrompt,
+  tools: [
+    getWeather,                        // custom tool, runs on the host (automated or HITL)
+    codeInterpreter(),                 // sandboxed code execution in the microVM
```
AlexanderRichey, line 188: I guess these get handled differently inside Agent when the model is AgentCoreHarness? E.g., in the case model=AgentCoreHarness and you pass browser, then browser bottoms out to a json blob passed in the InvokeHarness request, as opposed to the usual code level handling.

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+
+## Proposed AgentCore Changes
+
+The provider works against `InvokeHarness` as it exists today. The item below is about smoothing the iterate-from-the-SDK experience, not enabling the integration.
+
+### Provision the execution role for the full tool surface up front.** 
+
+The default harness execution role already covers Bedrock model invocation and the built-in browser and code interpreter, so those vended tools work immediately. Other capabilities each require an additional statement on the execution role: AgentCore Gateway, custom browser or code-interpreter resources, non-Bedrock model providers (which need API-key credential access), and skills fetched from S3 or Git. Adding one today means editing the role and re-running `agentcore deploy`. 
+
+If the role were provisioned with the broader permission set at creation, a developer could add or swap any of these purely through the Strands `AgentConfig`, passing them on each `InvokeHarness` call, without redeploying. This is what makes the "change the agent in code, not in config" experience hold for the full tool surface: deploy once, then iterate entirely from the SDK.
```
mkmeral, line 207: wait what does this mean? is this even desirable? if I have CDK to manage my roles/permissions, do I want random agent SDK to suddenly change the scope?

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+| `messages` | ✓ conversation messages | — |
+| `systemPrompt` | ✓ `systemPrompt` | — |
+| `tools` | ✓ `tools` (classified, see [Tools](#tools)) | — |
+| `model`, `maxTokens`, `temperature`, `topP` | ✓ model config (Bedrock / OpenAI / Gemini / LiteLLM), carrying its own inference parameters | — |
+| `runtimeSessionId` | — | ✓ |
+| `harnessArn` | — | ✓ |
+
+The streamed response (message starts, content block deltas, tool-use and tool-result blocks, stop reasons, and usage) maps one-to-one onto the Strands model event types the agent loop already consumes. The full field-level mapping can be found in the [Appendix](#appendix).
+
+The Harness exposes further optional overrides which can be set via the provider config as well: per-user memory scoping (`actorId`), a named endpoint/version (`qualifier`), microVM-loaded skills, and the loop's safety limits (`allowedTools`, `maxIterations`, `timeoutSeconds`). These default sensibly and most developers never set them; the full set can also be found in the [Appendix](#appendix).
```
poshinchen, line 83: Question about the AgentCore harness config: given an existing AC harness, will there be a pre-defined Config? Does the user have to copy-paste the config so Strands can use it correctly?

team/designs/0012-harness-as-model-provider.md
mkmeral, line 227: > and it gives up the one-parameter swap

is this important? what's the user story of one-parameter swap?

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+
+```typescript
+const getWeather = tool({
+  name: 'get_weather',
+  inputSchema: z.object({ city: z.string() }),
+  callback: async ({ city }) => fetchWeather(city),   // runs on the host
+})
+```
+
+The provider implementation translates these to `inline_function` definitions (name, description, schema, but no implementation). When the model calls one, the Harness pauses and hands the call back to the host; the host runs the callback (automatically or with a human in the loop), and a follow-up request resumes the loop with the result automatically.
```
JackYPCOnline, line 99: Do we have any datapoint about how it works/ performs in latency, cost wise

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+    getWeather,          // custom → inline_function, runs on the host
+    codeInterpreter(),   // vended → runs in the microVM
+    browser(),           // vended → runs in the microVM
+  ],
+})
+```
+
+### Context and Memory
+
+The developer can choose how conversation context and memory are managed: let the Harness handle it, or bring their own. The `stateful` flag on the provider selects between the two.
```
lizradway, line 118: how do we expect this to behave with our context/memory constructs? 

should `stateful` mode disable context + memory? or should it build on top of? 

is that how we should also behave using the responses API as well?

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+    api: 'responses',
+    stateful: true,
+    baseURL: 'https://bedrock-mantle.us-west-2.api.aws/v1',
+    modelId: 'openai.gpt-oss-120b',
+  }),
+})
+```
+
+- **Pro:** no new code; an existing provider already speaks a stateful, server-side-tool API.
+- **Con:** it is a different system. Mantle Responses offers stateful inference and server-side tools, but not the Harness agent loop: no vended browser or code interpreter, no skills, no managed or bring-your-own memory, no microVM filesystem, no inline-function human-in-the-loop, no mid-session model switching.
```
pgrayy, line 262: Mantle and OpenAI both offer vended tools (including a code interpreter) that can be configured ([docs](https://strandsagents.com/docs/user-guide/concepts/model-providers/openai-responses/#built-in-tools)).

HITL and model switching should be supported as well, but the other features are not configurable as far as I know. Maybe one day though.

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+  model: { provider: 'bedrock', modelId: '...' },
+  tools: [getWeather, codeInterpreter()],
+})
+```
+
+- **Pro:** not bound by the `Model` interface's `stream()` contract, so it could expose harness-native capabilities a model provider cannot: dedicated methods for session control, filesystem and command access, or live browser interaction, plus per-invocation harness arguments as first-class parameters.
+- **Pro:** ability to evolve class separately from model providers
+- **Con:** more surface to build and own than a thin provider, and it gives up the one-parameter swap, since adopting it means switching to a new agent type rather than changing the model.
+
+This is a natural evolution if richer harness-native control is needed. The provider is the right first step: it delivers the core managed experience immediately and with the smallest change to existing code.
```
AlexanderRichey, line 229: We're planning to support more harnesses in addition to Strands - customers have asked for Pi, OpenCode, and Claude Code. Does that impact our decisions here?

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+```typescript
+const agent = new HarnessAgent({
+  harnessArn,
+  sessionId,
+  model: { provider: 'bedrock', modelId: '...' },
+  tools: [getWeather, codeInterpreter()],
+})
+```
+
+- **Pro:** not bound by the `Model` interface's `stream()` contract, so it could expose harness-native capabilities a model provider cannot: dedicated methods for session control, filesystem and command access, or live browser interaction, plus per-invocation harness arguments as first-class parameters.
```
mkmeral, line 225: another pro (or rather con of the proposal), it conflicts remote and local agents. What happens when I pass in a skills plugin to agent with agentcore managed harness? or context injector plugin? Do they just work? 

What about session/context managers? or interventions (which I assume wouldn't work if we are running the loop remotely, e.g. we cannot hook before code interpreter call for managed harness)

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+```typescript
+import { Agent, BedrockModel } from '@strands-agents/sdk'
+
+const agent = new Agent({
+  model: new BedrockModel(),
+  systemPrompt,
+})
+```
+
+Swapping the model to `AgentCoreHarness` moves the same agent into a managed microVM. Nothing else changes:
```
lizradway, line 163: I really like this devx

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+**Vended tools** carry no callback. They are references to capabilities the Harness runs itself in the microVM (code interpreter, browser, gateway, remote MCP), added with helpers like `codeInterpreter()` and `browser()` (see [Appendix](#appendix)).
+
+Each helper returns nothing more than a configuration object naming the capability, which the provider passes straight through into the `InvokeHarness` request; the Harness runs it in the sandbox and streams the call and result back.
+
+```typescript
+const agent = new Agent({
+  model: new AgentCoreHarness({ harnessArn, sessionId }),
+  tools: [
+    getWeather,          // custom → inline_function, runs on the host
+    codeInterpreter(),   // vended → runs in the microVM
```
zastrowm, line 110: "runs in the microVM" is a little odd; everything we pass into Agent tools runs on the local environment today, whereas this changes that model.

If we do this, I would expect the same for other model provides (for enabling their server-side tools) as this cannot/should not special case harness

(along the same lines as Murat above)

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+- **Con:** the `Agent` must support construction with no `model`, a structural change to a core contract, since the plugin hijacks inference.
+- **Con:** configuration splits across the plugin and the agent, and a plugin that silently replaces the model is a surprising mental model for something that is, conceptually, just where inference happens.
+
+### 3. Reuse the existing OpenAI Responses provider
+
+Strands already has an `OpenAIModel` in Responses mode that can target the Bedrock Mantle endpoint, which supports stateful conversations and server-side tools through Lambda or AgentCore Gateway.
+
+```typescript
+new Agent({
+  model: new OpenAIModel({
```
AlexanderRichey, line 252: I'm not sure this would work since we don't support OpenResponses. There's still a question as to whether or not we will going forward too.

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+
+`stateful` is an existing Strands flag (server-stateful providers like the Responses API already use it) and is purely an SDK-side switch. It controls whether the SDK keeps the conversation history and sends it each turn, or clears it and lets the Harness rebuild context server-side. 
+
+**Let the Harness manage it (`stateful: true`, the default)**
+
+Use this when the harness is deployed with memory enabled. 
+
+The Harness owns conversation state. It persists each turn server-side by `sessionId`, reloads history on the next invocation, and manages context and long-term memory itself, scoped per user with `actorId`. The provider sends only the new message each turn and clears local history; the developer configures nothing.
+
+**Bring your own (`stateful: false`)**
```
JackYPCOnline, line 128: I would expect a harness model manage the context by itself.

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+- Per-invocation control of harness parameters is not available until `invocationState` is plumbed into the provider (see [Open Questions](#open-questions)).
+- The provider's behavior depends on harness-side details the SDK cannot see (the deployed memory configuration, the internal loop's streaming granularity), which need validation against a live harness.
+
+### Migration
+
+No migration is required. The provider is additive: existing agents and the other model providers are unaffected.
+
+## Open Questions
+
+- **Provider location and ownership**: Should `AgentCoreHarness` live in the Strands SDK alongside the other providers, or in the AgentCore SDK? As a thin `Model` provider it could be owned and maintained by the AgentCore team in their own SDK, matching how the integration is positioned. Living in Strands gives a single import and keeps it beside the other providers. (A `HarnessAgent` or plugin, by contrast, would more naturally live in Strands core.)
```
lizradway, line 285: in AgentCore SDK imo

team/designs/0012-harness-as-model-provider.md
gautamsirdeshmukh: I'd be proposing having a mechanism that creates fine-grained inline policies on the fly to avoid having to `agentcore deploy` to update the execution role

team/designs/0012-harness-as-model-provider.md
zastrowm, line 128: The last time we talked about this, I thought the consensus was that stateful: false was really expensive in terms of runtime & overhead; if so I don't think it makes sense to offer it as an option

team/designs/0012-harness-as-model-provider.md
lizradway: in appendix btw

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+
+But when it comes to developers, these managed platforms all come with a tradeoff: the rich, extensible SDK must be left behind for raw API calls and low-level plumbing. An intuitive toolkit for building agents, OR the infrastructure that makes production possible... why should you have to choose?
+
+[AgentCore Harness](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/harness.html) has solved the hard infrastructure problem through microVM isolation, persistent memory, identity, observability and more for agents running in production. Developer experience is [Strands](https://strandsagents.com/)' calling card: typed streaming, hooks, multi-agent orchestration, and overall an intuitive framework that developers already know well.
+
+In short: developers already build with the Strands SDK, and the same framework is already running their agent in production (Harness). It's time to put the two pieces together.
+
+## Proposal
```
mkmeral, line 38: imho, this should be another class not model parameter. It will start to confuse the stateful and stateless model, and it will create bunch of edge cases.

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+
+**Bring your own (`stateful: false`)**
+
+Use this when the harness is deployed with memory disabled, so context is managed only on the host.
+
+The host owns conversation state, exactly as it does for a local agent. The agent's Strands conversation manager, context offloader, and compression tools run on the host and decide what the model sees each turn (trimming, summarizing, or offloading history), and the provider sends that host-managed context to the Harness. This is the path for developers who want Strands' context primitives instead of the Harness's built-in management.
+
+### Hooks
+
+All the same hook events still fire and stream to the host, whether the work ran on the host or inside the microVM. A hook can fully intervene in work run on the host, but can only observe work the microVM has already done (and "react" to it on the host side).
```
zastrowm, line 136: > All the same hook events still fire and stream to the host, whether the work ran on the host or inside the microVM.

I'm a little confused how this works for model & tool calls; how can we react to tool calls on the host side

team/designs/0012-harness-as-model-provider.md
lizradway: modification of harness i would imagine.


can you modify these things currently on the agent core sdk? cc: alex

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+  - [Hooks](#hooks)
+- [Developer Experience](#developer-experience)
+- [Proposed AgentCore Changes](#proposed-agentcore-changes)
+- [Alternative Implementations](#alternative-implementations)
+- [Consequences](#consequences)
+- [Open Questions](#open-questions)
+- [Willingness to Implement](#willingness-to-implement)
+- [Appendix](#appendix)
+
+## Context
```
poshinchen, line 28: I'm wondering about the user story here. This design covers the case of taking agents from prototype to production.

Would there be a case where user wants its agent to use the AC harness as a tool? If so, will there be any blocker?

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+### Provision the execution role for the full tool surface up front.** 
+
+The default harness execution role already covers Bedrock model invocation and the built-in browser and code interpreter, so those vended tools work immediately. Other capabilities each require an additional statement on the execution role: AgentCore Gateway, custom browser or code-interpreter resources, non-Bedrock model providers (which need API-key credential access), and skills fetched from S3 or Git. Adding one today means editing the role and re-running `agentcore deploy`. 
+
+If the role were provisioned with the broader permission set at creation, a developer could add or swap any of these purely through the Strands `AgentConfig`, passing them on each `InvokeHarness` call, without redeploying. This is what makes the "change the agent in code, not in config" experience hold for the full tool surface: deploy once, then iterate entirely from the SDK.
+
+
+## Alternative Implementations
+
+### 1. A dedicated `HarnessAgent` class
```
mkmeral, line 212: We already have remote agents as a concept, e.g. `A2AAgent`. Is there any reason we cannot just expand on the pattern?

team/designs/0012-harness-as-model-provider.md
zastrowm, line 285: A couple of Qs to guide discussion:

 - What dependencies does this need?
 - Who is going to be making changes the most?

team/designs/0012-harness-as-model-provider.md
gautamsirdeshmukh: In the sense of, we cannot actually intervene with those microVM vended tool calls the way we can for host-side tool calls, but we CAN still take advantage of the before/after events that are streamed back to the host for things like collecting metrics, logging, etc. The wording might be a little confusing here

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+
+```typescript
+new Agent({
+  plugins: [new ManagedHarness({ harnessArn, sessionId })],
+  tools: [getWeather, codeInterpreter()],
+})
+```
+
+- **Pro:** harness parameters can be set per invocation through `invocationState`, not just statically on a constructor.
+- **Con:** the `Agent` must support construction with no `model`, a structural change to a core contract, since the plugin hijacks inference.
```
pgrayy, line 243: What if the plugin supported a `getModel` similar to `getTools`?

team/designs/0012-harness-as-model-provider.md
lizradway: > a mechanism that creates fine-grained inline policies on the fly 

this feels dangerous. violates the basic policy control plane and policy management practice: https://docs.aws.amazon.com/wellarchitected/latest/security-pillar/sec_permissions_least_privileges.html

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+
+- **Exposing the microVM as a Strands `Sandbox`**: Strands tools like `AgentSkills` read from a pluggable `Sandbox` rather than the local filesystem. A `Sandbox` backed by the harness microVM would let those tools operate directly on the microVM's filesystem, unifying host-side and microVM-side skills (and filesystem and command access) under one interface. This would be distinct from the in-progress `AgentCoreSandbox`, which wraps the standalone AgentCore Code Interpreter; a harness `Sandbox` would instead wrap the running harness session. Worth exploring as a follow-on.
+
+- **Per-invocation harness overrides**: The `Model` `stream()` contract does not receive `invocationState`, so harness parameters are static per agent today. Plumbing `invocationState` into the provider (or the proposed harness-invoke hooks) would allow per-call overrides such as `actorId` or `qualifier`.
```
zastrowm, line 291: Do we know how often customers change per-invocation settings?

team/designs/0012-harness-as-model-provider.md
mkmeral: +1 ^

what are the use cases? sure developers lose some stuff when they go to managed infrastructure. but what are they trying to do that they cannot now? is this about adding hooks to managed harnesses? is this about multiagent? or do i just want a simple sdk to handle communications?

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+| `model_context_window_exceeded` | `modelContextWindowExceeded` |
+| `max_iterations_exceeded` | `limitTurns` |
+| `max_output_tokens_exceeded` | `limitOutputTokens` |
+
+</details>
+
+<details>
+<summary>Stream event mapping</summary>
+
+`InvokeHarness` streams the same event shapes the SDK already consumes, so mapping is one-to-one:
```
zastrowm, line 356: In TS or Python?

And also, how does it compare to the other language? (e.g. if Python, does it map to TS cleanly?)

team/designs/0012-harness-as-model-provider.md
gautamsirdeshmukh: See the examples of the code interpreter and browser tools in the appendix - we would _not_ be creating our own no-op versions of those tools. Rather, the Strands tool implementations of these AgentCore harness vended tools are basically just zod wrappers around the JSON object that is expected in the InvokeHarness call:

https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_InvokeHarness.html#API_InvokeHarness_RequestSyntax

It's easier to understand if you see a code example of it

team/designs/0012-harness-as-model-provider.md
poshinchen: I think the current propagation already handles that so there shouldn't be duplicate traces.

On Strands side there will be only `model` / `tool` span, the rest of them would be attached by AC Harness traces. But I think there will be works from the AC harness side to get the traceId from the header and attach. Otherwise it'll be come separate traces.

team/designs/0012-harness-as-model-provider.md
ryanycoleman: ++

team/designs/0012-harness-as-model-provider.md
yonib05: I believe it should be additive, so if harness is configured with x tools or permissions, your local tools, config get added to the harness ones, but it doesn't change the harness ones; it just supplements additional.

So if the tool or access is derrived from the harness config it uses the harness role if the tool is derved from passed in config it uses the passed in role (with a fallback to harness role, or maybe this should be the general default, passed in superceeds server but falls back to server)

team/designs/0012-harness-as-model-provider.md
```diff
@@ -0,0 +1,441 @@
...
+The provider works against `InvokeHarness` as it exists today. The item below is about smoothing the iterate-from-the-SDK experience, not enabling the integration.
+
+### Provision the execution role for the full tool surface up front.** 
+
+The default harness execution role already covers Bedrock model invocation and the built-in browser and code interpreter, so those vended tools work immediately. Other capabilities each require an additional statement on the execution role: AgentCore Gateway, custom browser or code-interpreter resources, non-Bedrock model providers (which need API-key credential access), and skills fetched from S3 or Git. Adding one today means editing the role and re-running `agentcore deploy`. 
+
+If the role were provisioned with the broader permission set at creation, a developer could add or swap any of these purely through the Strands `AgentConfig`, passing them on each `InvokeHarness` call, without redeploying. This is what makes the "change the agent in code, not in config" experience hold for the full tool surface: deploy once, then iterate entirely from the SDK.
+
+
+## Alternative Implementations
```
JackYPCOnline, line 210: ^What is the pros and cons in your opinion

team/designs/0012-harness-as-model-provider.md
poshinchen, line 212: tbh I personally like this one.

team/designs/0012-harness-as-model-provider.md
zastrowm: What are the output formats of these?  Just raw of what the SDK supports?

team/designs/0012-harness-as-model-provider.md
zastrowm: On the other hand, we already follow this model with OpenAIResponses so there is prior art

zastrowm: if we do choose a different implementation path, we should consider how OpenAI Responses slots into the same design

team/designs/0012-harness-as-model-provider.md
ryanycoleman: Does it actually move though? How would the session state transfer?

team/designs/0012-harness-as-model-provider.md
ryanycoleman: My takeaway from the discussion is that we'd run the model provider model back again so Anthropic's managed agents, bedrock managed agents, agentcore harness... these would all be bespoke. Upside seems to be not having the abstraction be a common denominator.

PR #2638, opened 2026-06-05
docs: document per-invocation limits feature

notowen333:
## Motivation

The per-invocation limits feature shipped in sdk-typescript#1106 and sdk-python#2360 but has no documentation on the docs site. Developers have no way to discover how to cap turns, output tokens, or total tokens during agent execution.

## Public API Changes

No public API changes. This PR documents existing SDK functionality.

## Documentation PR

Documents the `limits` parameter added to `invoke`/`stream` in both SDKs:

```python
result = agent("Summarize this", limits={"turns": 5, "output_tokens": 2000})
```

```typescript
const result = await agent.invoke('Summarize this', {
  limits: { turns: 5, outputTokens: 2000 },
})
```

## Type of Change

Documentation update

## Testing

- [x] Verified all code examples against SDK source (local clones via `npm run sdk:clone`)
- [x] Verified stop reason strings match `StopReason` types in both SDKs
- [x] Verified behavioral claims (check timing, priority order, per-invocation scoping) against implementation

## Checklist

### DIFF
--- a/site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
+++ b/site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
@@ -95,7 +95,9 @@ Each model invocation ends with a stop reason that determines what happens next:
 - **Content filtered**: The response was blocked by safety mechanisms.
 - **Guardrail intervention**: A guardrail policy stopped generation.
 
-The limit stop reasons indicate graceful budget exhaustion. The agent's message history remains in a valid state, and you can reinvoke with a higher budget or different prompt. Both content filtered and guardrail intervention terminate the loop and should be handled according to application requirements.
+The limit stop reasons indicate graceful budget exhaustion. The agent's message history remains in a valid state, and you can reinvoke with a higher budget or different prompt.
+
+Both content filtered and guardrail intervention terminate the loop and should be handled according to application requirements.
 
 ### Extending the Loop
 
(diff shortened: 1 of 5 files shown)


site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
```diff
@@ -177,6 +180,66 @@ The `agent.cancel()` method provides a way to stop the loop from outside, such a
...
+  `int` (booleans are rejected).
+
+  </Tab>
+  <Tab label="TypeScript">
+
+  ```typescript
+  --8<-- "user-guide/concepts/agents/agent-loop.ts:limits_basic"
+  ```
+
+  The same option works with `stream`. TypeScript requires each
```
zastrowm: Don't reference TypeScript; they're on TS so they know

zastrowm: Why is this not true in Python too?

site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
```diff
@@ -177,6 +180,66 @@ The `agent.cancel()` method provides a way to stop the loop from outside, such a
...
+
+  if result.stop_reason == "limit_turns":
+      print("Hit turn budget")
+  elif result.stop_reason == "limit_total_tokens":
+      print("Hit token budget")
+  ```
+
+  The same parameter works with `invoke_async` and
+  `stream_async`. Python requires each cap to be a positive
+  `int` (booleans are rejected).
```
zastrowm: Why would they think booleans are accepted?  And don't mention Python

notowen333:
@zastrowm I've updated the skills to reflect the language thing you flagged

zastrowm (approved):

PR #2643, opened 2026-06-05
feat(context): add context_manager="auto" facade on Agent

lizradway:
## Description

Adds `contextManager: "auto"` / `context_manager="auto"` as an opt-in parameter on `Agent` in both SDKs. When set, it wires up the benchmark-winning context management configuration:

- `ContextOffloader(storage=InMemoryStorage(), maxResultTokens=1500, previewTokens=750)` — caches oversized tool results, keeps truncated previews in context, provides `retrieve_offloaded_content` tool
- `SummarizingConversationManager(summaryRatio=0.3, compressionThreshold=0.85)` — summarizes older messages on context overflow, proactively compresses at 85% context window usage

Coexistence rules:
- If user also provides `conversationManager`, theirs is used (offloader still added)
- If user already has a `ContextOffloader` in their plugins list, it's not overridden
- Throws with stateful models (they manage context server-side)

```python
from strands import Agent

agent = Agent(model="us.anthropic.claude-sonnet-4-6-v1", context_manager="auto")
```

```typescript

const agent = new Agent({ model: 'us.anthropic.claude-sonnet-4-6-v1', contextManager: 'auto' })
```

## Related Issues

strands-agents/docs#831

## Documentation PR

N/A — docs PR (strands-agents/docs#831) covers the design. User-facing docs page to follow once this ships.

## Type of Change

New feature

## Testing

- [x] Python: 16 unit tests covering all resolution paths, 502 existing agent tests pass
- [x] TypeScript: 10 unit tests covering all resolution paths, 122 existing agent tests pass
- [x] `hatch fmt --linter --check` passes (ruff + mypy)
- [x] `tsc --noEmit` passes
- [x] Merged with latest upstream/main, conflicts resolved, all tests re-verified

## Checklist

----


strands-ts/src/agent/agent.ts
```diff
@@ -351,12 +362,19 @@ export class Agent implements LocalAgent, InvokableAgent {
...
     if (this.model.stateful) {
-      if (config?.conversationManager) {
+      if (config?.conversationManager || config?.contextManager) {
         throw new Error(
-          'Cannot use a conversationManager with a stateful model. The model manages conversation state server-side.'
+          'contextManager and conversationManager cannot be used with a stateful model. The model manages conversation state server-side.'
         )
       }
       this._conversationManager = new NullConversationManager()
+    } else if (config?.contextManager === 'auto') {
```
notowen333: Can we add the same resolver pattern that we have in Python in this PR?

strands-py/src/strands/agent/agent.py
```diff
@@ -378,12 +391,71 @@ def __init__(
...
+
+        resolved_plugins = list(plugins) if plugins else []
+
+        has_offloader = any(
+            isinstance(p, ContextOffloader) for p in resolved_plugins
+        )
+        if not has_offloader:
+            offloader = ContextOffloader(
+                storage=InMemoryStorage(),
+                max_result_tokens=1_500,
```
notowen333: I feel like these magic benchmarked values should be in a constant somewhere well placed

strands-py/src/strands/agent/agent.py
```diff
@@ -378,12 +391,71 @@ def __init__(
...
+        plugins_to_register = resolved_plugins if resolved_plugins is not None else plugins
+        if plugins_to_register:
+            for plugin in plugins_to_register:
                 self._plugin_registry.add_and_init(plugin)
 
         self.hooks.invoke_callbacks(AgentInitializedEvent(agent=self))
 
+    @staticmethod
+    def _resolve_context_manager(
+        context_manager: "Literal['auto'] | None",
```
notowen333: Is it not the python meta to make a type for this kind of thing that's re-used

Or we repeat the literal types in multiple places?

strands-ts/src/agent/agent.ts
```diff
@@ -174,6 +177,14 @@ export type AgentConfig = {
...
    */
   conversationManager?: ConversationManager
+  /**
+   * Context management strategy. When set to `"auto"`, composes a ContextOffloader
+   * plugin (maxResultTokens=1500, previewTokens=750) with a SummarizingConversationManager
+   * (summaryRatio=0.3, compressionThreshold=0.85) using benchmark-validated defaults.
+   * If `conversationManager` is also provided, the user's conversation manager is used instead.
+   * Defaults to undefined (no context management).
+   */
+  contextManager?: 'auto'
```
notowen333: Similarly want to pull this out into a type?

strands-ts/src/agent/agent.ts
lizradway: added

strands-py/src/strands/agent/agent.py
lizradway: sure, added

strands-ts/src/agent/agent.ts
lizradway: done: `ContextManagerStrategy`

agent-of-mkmeral:
**Assessment**: Approve (with one new finding)

Reviewed the full diff and verified the facade wiring against the actual source of the composed primitives (`SummarizingConversationManager`, `ContextOffloader`, `InMemoryStorage`, base `ConversationManager`). This is a clean, well-scoped opt-in facade and the implementation holds up. Prior review feedback (validation ordering, runtime guards, constants extraction, docstrings, plugin-ordering parity) is all addressed.

### ✅ Verified correct (not just claimed)
- **Constructor params match real APIs.** `SummarizingConversationManager(summary_ratio=0.3, proactive_compression={"compression_threshold": 0.85})` matches the actual signature; the `dict` form is the documented way to set the threshold, and `summary_ratio` is clamped to `[0.1, 0.8]` so `0.3` is valid. `ContextOffloader(storage=, max_result_tokens=1500, preview_tokens=750)` matches — and passes the constructor's own guards (`preview_tokens < max_result_tokens`, both positive). ✓
- **Test assertions hit real attributes.** `_compression_threshold` (set in base `ConversationManager.__init__`), `_max_result_tokens`, `_preview_tokens`, `_storage`, and plugin key `"context_offloader"` (`name = "context_offloader"`) all exist as asserted. ✓
- **Coexistence semantics are sound.** Stateful-model guard now fires *before* construction; `conversation_manager` precedence and offloader-dedup-by-instance are correct; the `get_args(ContextManagerStrategy)` runtime validation keeps type and validation in sync. ✓
- **Cross-SDK parity** on defaults (0.3 / 0.85 / 1500 / 750), error messages, and plugin append-ordering. ✓

### 🟡 New finding — `InMemoryStorage` + `session_manager` silently loses offloaded content across runs
This hasn't been raised in prior reviews. `context_manager="auto"` always wires `ContextOffloader(storage=InMemoryStorage())`. `ContextOffloader` has **no `restore_from_session`** (confirmed: 0 occurrences in `plugin.py`), and `InMemoryStorage`'s own docstring warns *"Content accumulates for the lifetime of this instance... For long-running agents, consider FileStorage or S3Storage for persistent storage."*

The failure mode: a user combines `context_manager="auto"` with a `session_manager` (a very natural pairing — both are about long-lived conversations). On run N the conversation history is persisted and restored, but it contains **truncated previews with offload references** pointing into an `InMemoryStorage` that died with the previous process. On restore, `retrieve_offloaded_content` for those references will fail — the full content is gone. So the facade's convenience can quietly produce dangling references precisely in the long-running/persisted scenario it's meant to help.

Not a blocker for this PR (the primitive composition is correct and the default is reasonable for stateless/serverless use), but worth one of:
1. A docstring note on `context_manager="auto"` that the offloader uses in-memory storage and is **not durable across sessions** — drop to the low-level API (explicit `ContextOffloader` with `FileStorage`/`S3Storage`) when pairing with a `session_manager`; or
2. A follow-up issue to make the `"auto"` offloader session-aware.

### 🔵 Minor (already flagged by others, confirming current state)
- `ContextManagerStrategy` is still **not** re-exported from `strands-ts/src/index.ts`, while `ToolExecutorStrategy` is (`export type { AgentConfig, ToolList, ToolExecutorStrategy }`). Adding it there keeps the public type surface consistent. Cosmetic.
- The Python "prepended"→"appended" docstring fix landed (commit `619e694`), so that earlier note is resolved.

### API bar-raising
This adds a new public `Agent` parameter (moderate scope). The linked `strands-agents/docs#831` looks like the design home; consider the `needs-api-review` label per `API_BAR_RAISING.md` if not already tracked.

Nice work — the "one parameter composes benchmark-validated primitives" design is well aligned with *simple things should be simple* while keeping the low-level building blocks fully available. My only substantive ask is documenting the in-memory-storage durability caveat.

mkmeral (dismissed):

opieter-aws (dismissed):

strands-py/src/strands/agent/agent.py
```diff
@@ -107,6 +109,21 @@ class _DefaultRetryStrategySentinel:
 _DEFAULT_AGENT_NAME = "Strands Agents"
 _DEFAULT_AGENT_ID = "default"
 
+ContextManagerStrategy = Literal["auto"]
+"""Supported values for the ``context_manager`` parameter."""
+
+_CM_MAX_RESULT_TOKENS = 1_500
```
arronbailiss: throughout, could we use more descriptive naming rather than "cm"?

strands-py/src/strands/agent/agent.py
```diff
@@ -378,12 +407,72 @@ def __init__(
...
+        if context_manager is None:
+            return None, None
+
+        supported = get_args(ContextManagerStrategy)
+        if context_manager not in supported:
+            raise ValueError(
+                f"Unsupported context_manager value: {context_manager!r}. Supported values: {supported}"
+            )
+
+        from ..vended_plugins.context_offloader import ContextOffloader, InMemoryStorage
```
arronbailiss, line 454: are this intentionally imported here rather than at the top of the file?

strands-py/src/strands/agent/agent.py
lizradway: sure!

strands-py/src/strands/agent/agent.py
lizradway: This was intentional so that context manager imports are kept optional in the in the agent loop for now (unless actually used).

Once we have the context manager class, all context management logic will be siloed under there and we can remove all of this and just instantiate the class in a one( or two)-liner.

strands-py/src/strands/agent/agent.py
lizradway: updated

awsarron (dismissed):

notowen333 (approved):

mkmeral (approved):

opieter-aws (approved):

PR #2649, opened 2026-06-05
feat: update vended tools/plugins for sandbox compatibility

gautamsirdeshmukh:
## Description

Update vended tools (bash, fileEditor) and plugins (context offloader, skills) to be [Sandbox](https://github.com/strands-agents/harness-sdk/blob/main/strands-ts/src/sandbox/base.ts) compatible.

This is the fourth PR split off from https://github.com/strands-agents/sdk-typescript/pull/1011, preceded by the base interface merged in https://github.com/strands-agents/sdk-typescript/pull/1090, Docker/SSH implementations merged in https://github.com/strands-agents/sdk-typescript/pull/1110 and Agent integration merged in https://github.com/strands-agents/harness-sdk/pull/2563.

## What's New

- Vended tools (bash, file editor) as now sandbox-aware, all routing handled under the hood
- Vended plugins (context offloader, skills) are now sandbox-aware, all routing handled under the hood
  - Added `SandboxStorage` file storage implementation for context offloader
- `getTools()` overridable method added to sandboxes enabling default tool registration
  - Empty by default, bash and file editor auto-registered for Docker and SSH sandboxes
- `makeBash()` factory for generating bash tool instances with environment-appropriate descriptions

## Related Issues

- https://github.com/strands-agents/sdk-typescript/pull/1011 (parent mega-PR)
- https://github.com/strands-agents/sdk-typescript/pull/1090 (base Sandbox / PosixShellSandbox interfaces)
- https://github.com/strands-agents/sdk-typescript/pull/1110 (Docker / SSH implementations)
- https://github.com/strands-agents/harness-sdk/pull/2563 (Agent integration)

## Documentation PR

Not yet

## Type of Change

New feature

## Testing

- [x] I ran all tests + added several

## Checklist

----


strands-ts/src/vended-tools/bash/bash.ts
```diff
@@ -251,21 +253,44 @@ process.on('SIGTERM', () => {
...
+    const sandbox = agent.sandbox
 
     // Validate execute mode has command
     if (input.mode === 'execute' && !input.command) {
       throw new Error('command is required when mode is "execute"')
     }
 
+    // Real sandbox: stateless execution. Restart is a no-op (no shell state persists),
+    // and sandbox errors are mapped to bash's own error classes.
+    if (!(sandbox instanceof NotASandboxLocalEnvironment)) {
```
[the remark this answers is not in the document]
gautamsirdeshmukh: The persistent/non sandbox path IS inherently node-only, so checking instanceof NotASandbox is acceptable here

strands-ts/src/agent/agent.ts
```diff
@@ -301,6 +317,18 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+  private readonly _sandbox: Sandbox | false | undefined
+
+  /**
+   * Execution environment for running commands, code, and file operations.
+   *
+   * @throws DefaultNotConfiguredError if no sandbox is configured for this
+   * environment (e.g. browsers, where no host default is registered).
+   */
+  get sandbox(): Sandbox {
+    return this._sandbox || defaultSandbox.get()
```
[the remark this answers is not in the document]
gautamsirdeshmukh: This change is from https://github.com/strands-agents/harness-sdk/pull/2563 (this tools PR 2649 is stacked on top of it, will merge 2563 then rebase before merging this one).

Additionally - this IS correct, right now false/disabled routes to the same case as empty/default (NotASandbox). Please do not flag this again.

strands-ts/src/sandbox/__tests__/default.test.node.ts
```diff
@@ -0,0 +1,28 @@
...
+    const sandbox = new NotASandboxLocalEnvironment()
+    expect(new Agent({ model: new MockMessageModel(), sandbox }).sandbox).toBe(sandbox)
+  })
+
+  it('falls back to the registered default when unconfigured', () => {
+    expect(new Agent({ model: new MockMessageModel() }).sandbox).toBeInstanceOf(NotASandboxLocalEnvironment)
+  })
+
+  it('treats sandbox: false the same as unconfigured', () => {
+    expect(new Agent({ model: new MockMessageModel(), sandbox: false }).sandbox).toBeInstanceOf(
```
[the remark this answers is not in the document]
gautamsirdeshmukh: This change is from https://github.com/strands-agents/harness-sdk/pull/2563 (this tools PR 2649 is stacked on top of it, will merge 2563 then rebase before merging this one).

In terms of the suggestion, will incorporate into 2563.

strands-ts/src/vended-plugins/context-offloader/storage.ts
```diff
@@ -1,17 +1,20 @@
...
- * implementations: {@link InMemoryStorage}, {@link FileStorage}, and {@link S3Storage}.
+ * This module defines the {@link Storage} interface and provides four built-in
+ * implementations: {@link InMemoryStorage}, {@link FileStorage}, {@link S3Storage},
+ * and {@link SandboxStorage}.
  * Each content block from a tool result is stored individually with its content type preserved.
  */
 
+import type { Sandbox } from '../../sandbox/base.js'
+
 /**
```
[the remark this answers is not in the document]
gautamsirdeshmukh: The late binding is required - ContextOffloader injects the agent's sandbox during `initAgent` which runs after SandboxStorage is constructed, so sandbox can't be a constructor param. 

Regarding the concern about the sandbox being set and reset across multiple callers, this is not a concern - SandboxStorage is user-constructed and passed to one plugin, not a shared singleton.

strands-ts/src/sandbox/not-a-sandbox-local-environment.ts
```diff
@@ -0,0 +1,90 @@
...
+export class NotASandboxLocalEnvironment extends Sandbox {
+  private _resolvePath(path: string): string {
+    return isAbsolute(path) ? path : join(process.cwd(), path)
+  }
+
+  async *executeStreaming(
+    command: string,
+    options?: ExecuteOptions
+  ): AsyncGenerator<StreamChunk | ExecutionResult, void, undefined> {
+    const cwd = options?.cwd ?? process.cwd()
```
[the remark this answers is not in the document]
gautamsirdeshmukh: This change is from https://github.com/strands-agents/harness-sdk/pull/2563 (this tools PR 2649 is stacked on top of it, will merge 2563 then rebase before merging this one).

Will address in 2563.

strands-ts/src/sandbox/base.ts
```diff
@@ -110,6 +111,26 @@ export abstract class Sandbox {
...
+   * Returns a {@link FileInfo} with `name` and `isDir` fields. `size` is
+   * populated when the backend can determine it cheaply. Throws if the path
+   * does not exist.
+   *
+   * @param path - Path to stat.
+   * @returns Metadata about the path.
+   * @throws Error if the path does not exist or is not accessible.
+   */
+  abstract statFile(path: string): Promise<FileInfo>
+
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Considered adding a fallback to the base class, but given that Sandbox was introduced recently and isn't formally released yet, opting to make the breaking change of adding a `statFile` abstraction on the base and filling in for implementations

strands-ts/src/agent/agent.ts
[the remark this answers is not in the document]
gautamsirdeshmukh: Already addressed numerous times, this is intentional. Please stop flagging.

strands-ts/src/vended-tools/bash/bash.ts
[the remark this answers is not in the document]
gautamsirdeshmukh: Duplicate comment, already addressed above. Instanceof check is acceptable given that persistent/host execution occurs via NotASandbox.

strands-ts/src/sandbox/base.ts
```diff
@@ -110,6 +111,26 @@ export abstract class Sandbox {
...
+  /**
+   * Tools this sandbox vends to an agent, auto-registered at init. Override to provide them.
+   */
+  getTools(): Tool[] {
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Will add TSDoc

strands-ts/src/sandbox/__tests__/default.test.node.ts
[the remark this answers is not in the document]
gautamsirdeshmukh: Duplicate comment, will address in PR 2563

strands-ts/src/sandbox/posix-shell.ts
```diff
@@ -136,4 +122,20 @@ export abstract class PosixShellSandbox extends Sandbox {
...
+
+  /**
+   * Resolve path metadata in a single shell round-trip via `test`, avoiding the
+   * base implementation's probe (which reads the whole file just to confirm it
+   * exists). Emits `d` for a directory, `f` otherwise; a non-existent path exits
+   * non-zero and throws.
+   */
+  async statFile(path: string): Promise<FileInfo> {
+    const quoted = shellQuote(path)
+    const result = await this.execute(`test -e ${quoted} || exit 1; test -d ${quoted} && echo d || echo f`)
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Actually, filter(Boolean) already handles trailing slashes: "dir/".split('/').filter(Boolean) is ["dir"], so .pop() returns "dir", not an empty string. The difference from NotASandboxLocalEnvironment is intentional, posix-shell paths are always POSIX regardless of host OS, so splitting on / is correct, whereas path.basename() would split on \ on a Windows host. NotASandbox runs on the host, so using path.basename() is correct for that class.

strands-ts/src/sandbox/not-a-sandbox-local-environment.ts
```diff
@@ -0,0 +1,90 @@
...
+    yield* streamProcess('sh', ['-c', `cd ${shellQuote(cwd)} && ${command}`], {
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Updated to use buildShellEnvPrefix

strands-ts/src/sandbox/base.ts
```diff
@@ -110,6 +111,26 @@ export abstract class Sandbox {
...
+   *
+   * Returns a {@link FileInfo} with `name` and `isDir` fields. `size` is
+   * populated when the backend can determine it cheaply. Throws if the path
+   * does not exist.
+   *
+   * @param path - Path to stat.
+   * @returns Metadata about the path.
+   * @throws Error if the path does not exist or is not accessible.
+   */
+  abstract statFile(path: string): Promise<FileInfo>
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Duplicate comment, addressed

strands-ts/src/vended-plugins/context-offloader/storage.ts
```diff
@@ -260,3 +263,100 @@ export class S3Storage implements Storage {
...
+  private _counter = 0
+  private _contentTypes: Record<string, string> = {}
+  private _metadataLoaded = false
+  private _metadataWriteChain: Promise<void> = Promise.resolve()
+
+  /**
+   * Sandbox the storage operates against. Injected by {@link ContextOffloader} during
+   * plugin initialization. `store`/`retrieve` throw until this is set.
+   */
+  sandbox: Sandbox | undefined
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Duplicate comment, addressed

gautamsirdeshmukh:
Please note, along with tools/plugins changes, this PR (currently) includes agent integration commit as well (latest version in https://github.com/strands-agents/harness-sdk/pull/2563)

strands-ts/src/agent/agent.ts
```diff
@@ -476,6 +505,20 @@ export class Agent implements LocalAgent, InvokableAgent {
       })
     )
 
+    // Register tools vended by an explicitly-configured sandbox, keeping any the user
+    // already provided under the same name, and fold in any context it wants the model to know.
+    if (this._sandbox) {
+      for (const tool of this._sandbox.getTools()) {
+        if (!this._toolRegistry.get(tool.name)) {
```
mkmeral: we should log, if it has, and we are skipping it

agent-of-mkmeral:
Did a full local pass on this @gautamsirdeshmukh — checked out the head, ran the suites, exercised the Node/browser split. **LGTM overall, this is clean work.** ✅

## Verified locally (Node 22, `da402d3`)
```
✓ tsc --noEmit (src)                      0 errors
✓ sandbox + vended unit tests            411 passed / 17 files
✓ check:browser-bundle                   414.4kb ⚡
✓ default.test.browser.ts (real Chromium) agent.sandbox throws DefaultNotConfiguredError
✓ test:package (CJS + ESM)                “Node default sandbox registered” both
✓ eslint (sandbox + vended)               clean
```
The `defaultSlot` + `index.node.ts` side-effect + `sideEffects` allowlist mechanism is genuinely nice, and the skills-plugin rewrite to route through `sandbox.readText`/`listFiles` (dropping `fs`/`path`) is a real portability win.

Three concrete things I'd like a decision on before merge — none are blockers but two are correctness:

### 1. 🔴 `ExecuteOptions.env` docstring now contradicts a built-in
The updated docstring says:
> *"Built-in sandboxes **always** apply these…; custom Sandbox implementations must handle env explicitly."*

But `NotASandboxLocalEnvironment` is a built-in and ignores `options.env` (no `buildShellEnvPrefix`, no spawn `env`). So `env` works under Docker/SSH and silently no-ops on the host default. I know env work is deferred to #2563 — totally fine — but the docstring as written is now *actively wrong* for the host case. Either apply env in NotASandbox (one-liner with `buildShellEnvPrefix`) or soften the docstring to exclude the host fallback.

### 2. 🟡 `codeInterpreter` is exported but never auto-vended
`DockerSandbox.getTools()` and `SshSandbox.getTools()` both return `[fileEditor, bash]`. The new `codeInterpreter` tool routes purely through `sandbox.executeCode` (which every sandbox implements) and is the only browser-safe execution tool — yet no sandbox auto-registers it. If `bash` is auto-vended, including `codeInterpreter` seems natural. If it's intentionally omitted, a one-line comment on `getTools()` explaining why would save the next reader.

### 3. 🟡 Top-level exports incomplete (flat-namespace)
`index.ts` exports `Sandbox`, `PosixShellSandbox`, `DefaultNotConfiguredError` — but not `DockerSandbox`, `SshSandbox`, `NotASandboxLocalEnvironment`, `SandboxTimeoutError`, `SandboxAbortError`. Consumers can't `instanceof`-catch the sandbox errors from the main entry, which the bash tool's own error-mapping path makes relevant. Per the “Prefer Flat Namespaces” record, worth re-exporting at least the two error classes.

### Bonus observation (design, not a fix)
The context/tool injection in `agent.ts` keys off `this._sandbox` (explicit field), not `this.sandbox` (getter incl. registered default). Harmless today since `NotASandbox` vends nothing + returns `undefined` context, but if a richer *default* sandbox is ever registered via `defaultSandbox.set(...)`, its `getTools()`/`getSystemPromptContext()` would be silently skipped. Might be worth a comment marking that as intentional.

Also +1 on the bots re: `needs-api-review` — `Sandbox.statFile` (abstract, breaking), `AgentConfig.sandbox`, and the new `getTools`/`getSystemPromptContext` seams are real public surface.

Nice work overall — the architecture is the right shape. 🚀

strands-ts/src/sandbox/base.ts
```diff
@@ -110,6 +111,37 @@ export abstract class Sandbox {
...
```
mkmeral: > guess who's back
> back again 
> stat file is back! 
> tell your friends
> 
> guess who's back, guess who's back ...

why?

strands-ts/src/sandbox/base.ts
```diff
@@ -110,6 +111,37 @@ export abstract class Sandbox {
...
+  getTools(): Tool[] {
+    return []
+  }
+
+  /**
+   * Context appended to the agent's system prompt during `Agent.initialize()`, letting
+   * the model understand how this environment behaves. Override to provide it; returns
+   * `undefined` (no context) by default.
+   */
+  getSystemPromptContext(): string | undefined {
```
mkmeral: can this a field of sandbox, or do we believe having a method is better?

mkmeral: also should we tell the LLM which tools behave under this sandbox? technically this wouldn't apply to all tools, right?

strands-ts/src/sandbox/constants.ts
```diff
@@ -12,3 +12,17 @@ export const LANGUAGE_PATTERN = /^[a-zA-Z0-9._-]+$/
...
+ * Shell-escape a string for safe inclusion in a shell command.
+ *
+ * Wraps the value in single quotes and escapes any embedded single quotes
+ * using the '\'' pattern. Single quotes disable all shell expansion
+ * (variables, backticks, globbing), making this safe against injection.
+ *
+ * @param value - The string to escape.
+ * @returns The shell-escaped string wrapped in single quotes.
+ */
+export function shellQuote(value: string): string {
```
mkmeral, line 26: nit: do we have a library we can use?

strands-ts/src/sandbox/not-a-sandbox-local-environment.ts
mkmeral: would this work on windows?

strands-ts/src/types/messages.ts
```diff
@@ -753,6 +753,19 @@ export function systemPromptToData(prompt: SystemPrompt): SystemPromptData {
...
 }
 
+/**
+ * Appends text to a system prompt, preserving its string or content-block shape.
+ *
+ * @param prompt - The existing system prompt, or `undefined` to start a new one.
+ * @param text - The text to append.
+ * @returns A new system prompt with the text appended.
+ */
+export function appendToSystemPrompt(prompt: SystemPrompt | undefined, text: string): SystemPrompt {
```
mkmeral: just double checking, this is not exported, right?

strands-ts/src/vended-plugins/context-offloader/plugin.ts
```diff
@@ -154,6 +154,11 @@ export class ContextOffloader implements Plugin {
   }
 
   initAgent(agent: LocalAgent): void {
+    // SandboxStorage operates against the agent's configured sandbox; wire it up here
+    // so users don't have to hand the sandbox to the storage themselves.
+    if (this._storage instanceof SandboxStorage) {
```
mkmeral: this feels like a code smell, can we improve it? 

i'd expect sandbox to be passable to sandbox storage (dev principle wise), but we can have an empty sugar maybe? i.e. constructor for the plugin doesn't need sandbox, but it creates one from agents' if none is passed?

strands-ts/src/vended-plugins/skills/agent-skills.ts
```diff
@@ -118,6 +138,11 @@ export class AgentSkills implements Plugin {
   async initAgent(agent: LocalAgent): Promise<void> {
     await this._ready
 
+    // Load path-based skills through the agent's sandbox, which resolves to the host
+    // environment when no sandbox is configured.
+    this._sandbox = agent.sandbox
+    await this._loadSkillPaths(this._sandbox)
```
mkmeral: does this work on windows?

strands-ts/src/vended-plugins/skills/agent-skills.ts
```diff
@@ -157,9 +182,14 @@ export class AgentSkills implements Plugin {
...
   setAvailableSkills(skills: SkillSource[]): void {
-    const { skills: resolved, ready } = this._resolveSkills(skills)
+    const { skills: resolved, ready, skillPaths } = this._resolveSkills(skills)
     this._skills = resolved
-    this._ready = ready
+    this._skillPaths = skillPaths
+
+    // If a sandbox has already been captured (initAgent has run), fold the path load into
+    // `ready` so getAvailableSkills() and the injection hook await the path skills too.
+    const sandbox = this._sandbox
```
mkmeral: what happens when we pass agent skills into multiple agents? isn't `this._sandbox` an antipattern?

strands-ts/src/vended-plugins/skills/agent-skills.ts
```diff
@@ -443,15 +490,15 @@ export class AgentSkills implements Plugin {
       metadataLines.push(`Compatibility: ${skill.compatibility}`)
     }
     if (skill.path != null) {
-      metadataLines.push(`Location: ${join(skill.path, 'SKILL.md')}`)
+      metadataLines.push(`Location: ${skill.path}/SKILL.md`)
```
mkmeral, line 507: nit: windows?

strands-ts/src/vended-plugins/skills/agent-skills.ts
```diff
@@ -466,21 +513,32 @@ export class AgentSkills implements Plugin {
...
+  private async _listSkillResources(sandbox: Sandbox, skillPath: string): Promise<string[]> {
     const files: string[] = []
 
+    // List a directory recursively through the sandbox, returning paths relative to its root.
+    // Replaces readdirSync(dir, { recursive: true }), which has no sandbox equivalent.
+    const listFilesRecursive = async (dir: string): Promise<string[]> => {
+      const result: string[] = []
+      for (const entry of await sandbox.listFiles(dir)) {
+        if (entry.isDir)
+          result.push(...(await listFilesRecursive(`${dir}/${entry.name}`)).map((p) => `${entry.name}/${p}`))
```
mkmeral: why recursive? should we have a limit to depth? do we gain anything buy recursively reading everything?

strands-ts/src/vended-tools/bash/bash.ts
```diff
@@ -251,21 +253,44 @@ process.on('SIGTERM', () => {
 export const bash = tool({
   name: 'bash',
   description:
-    'Executes bash shell commands in a persistent session. Supports execute and restart modes. ' +
-    'Commands persist state (variables, directory) within the session. Node.js only.',
+    'Executes bash shell commands. Supports execute and restart modes. ' +
+    'When a Sandbox is configured, each call runs in a fresh shell that does not preserve state (variables, directory). ' +
```
mkmeral: I don't like this. We should not have ambiguity on tool descriptions for a case that's not applicable to the agents. If there is no sandbox description, what should the LLM do, try both? then it's multiple LLM calls wasted

strands-ts/src/vended-tools/bash/bash.ts
mkmeral: so we eat this up, and not tell the LLM? it'll try to refresh whenever it needs, which would fall to no-op, causing unnecessary token spend

strands-ts/src/vended-tools/code-interpreter/code-interpreter.ts
```diff
@@ -0,0 +1,56 @@
+import { tool } from '../../tools/tool-factory.js'
+import { z } from 'zod'
+import { Buffer } from 'buffer'
+import type { CodeInterpreterOutput } from './types.js'
+import type { OutputFile } from '../../sandbox/types.js'
+
+const codeInterpreterInputSchema = z.object({
+  code: z.string().min(1).describe('The source code to execute.'),
+  language: z.string().min(1).describe('Name of the interpreter binary (e.g., `python3`, `node`, `ruby`).'),
```
mkmeral: let's separate out this to another PR. I'm not 100% bought into this tool. I think it's too generic to have value, and we don't necessarily tell the LLM which binaries are available. Especially for "sandboxed environment" you won't have all of these available

strands-ts/src/agent/agent.ts
gautamsirdeshmukh: Adding debug log

strands-ts/src/sandbox/base.ts
gautamsirdeshmukh: I had brought in `stat` since it would allow the fileEditor tool to verify if paths exists / if a given path is a file/dir as efficiently as possible. Considering it again, however, the alternative (listFiles on the parent dir) is a negligible tradeoff. We can get rid of this.

strands-ts/src/sandbox/base.ts
gautamsirdeshmukh: Although this _could_ be a field, it's cleaner to have it as a method since we will be calling / parsing results from getTools. I also like matching convention between the two overridable methods, getTools and getSystemPromptContext

strands-ts/src/sandbox/constants.ts
gautamsirdeshmukh: The popular shell-quoting packages (shell-quote, shlex) are Node-specific, would break browser bundle

strands-ts/src/vended-tools/code-interpreter/code-interpreter.ts
gautamsirdeshmukh: ACK

strands-ts/src/sandbox/not-a-sandbox-local-environment.ts
gautamsirdeshmukh: NotASandboxLocalEnvironment uses native fs for file operations (works on Windows). This executeStreaming method is used by the `bash` tool, which was already Unix-only.

I would be inclined to have a follow up effort to create a windows/powershell sandbox implementation that enables support for `bash` and beyond

strands-ts/src/types/messages.ts
gautamsirdeshmukh: Nope

strands-ts/src/vended-plugins/context-offloader/plugin.ts
gautamsirdeshmukh: Updating to accept sandbox param instead of only reading it via initAgent

strands-ts/src/vended-plugins/skills/agent-skills.ts
gautamsirdeshmukh: Yep, the NotASandbox version uses native fs which works on windows

strands-ts/src/vended-plugins/skills/agent-skills.ts
gautamsirdeshmukh: We can get rid of `this._sandbox` and access it via the agent

strands-ts/src/vended-plugins/skills/agent-skills.ts
gautamsirdeshmukh: No issue, `/` works as a path separator on windows in node

strands-ts/src/vended-plugins/skills/agent-skills.ts
gautamsirdeshmukh: Will add a max depth of 3

strands-ts/src/vended-tools/bash/bash.ts
gautamsirdeshmukh: Will remove from here and leave sandbox info to the system prompt only

strands-ts/src/vended-tools/bash/bash.ts
gautamsirdeshmukh: Will update for sandbox case to return a message to the LLM that there is no persistent session to restart so it learns the first time

gautamsirdeshmukh:
Will also update to export the sandbox errors as mentioned by the bot

strands-ts/src/agent/agent.ts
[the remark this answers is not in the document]
gautamsirdeshmukh: Already addressed multiple times, this is intended.

strands-ts/src/vended-tools/bash/bash.ts
```diff
@@ -260,12 +262,34 @@ export const bash = tool({
...
+    const sandbox = agent.sandbox
 
     // Validate execute mode has command
     if (input.mode === 'execute' && !input.command) {
       throw new Error('command is required when mode is "execute"')
     }
 
+    // Real sandbox: stateless execution. Restart is a no-op (no shell state persists),
+    // and sandbox errors are mapped to bash's own error classes.
+    if (!(sandbox instanceof NotASandboxLocalEnvironment)) {
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Already addressed multiple times, resolving

strands-ts/src/sandbox/not-a-sandbox-local-environment.ts
```diff
@@ -0,0 +1,80 @@
...
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Addressed in https://github.com/strands-agents/harness-sdk/pull/2563

strands-ts/src/agent/agent.ts
```diff
@@ -476,6 +505,24 @@ export class Agent implements LocalAgent, InvokableAgent {
       })
     )
 
+    // Register tools vended by an explicitly-configured sandbox, keeping any the user
+    // already provided under the same name, and fold in any context it wants the model to know.
+    if (this._sandbox) {
```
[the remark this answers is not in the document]
gautamsirdeshmukh: We already mention this is for explicitly-configured sandboxes, resolving

strands-ts/src/types/agent.ts
```diff
@@ -262,6 +263,14 @@ export interface LocalAgent {
    */
   readonly toolRegistry: ToolRegistry
 
+  /**
+   * Execution environment for running commands, code, and file operations.
+   *
+   * @throws DefaultNotConfiguredError if no sandbox is configured for this
+   * environment (e.g. browsers, where no host default is registered).
```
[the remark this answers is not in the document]
gautamsirdeshmukh: This is from https://github.com/strands-agents/harness-sdk/pull/2563 , the change is necessary and acceptable

strands-ts/src/vended-tools/file-editor/file-editor.ts
```diff
@@ -171,284 +224,170 @@ function makeOutput(fileContent: string, fileDescriptor: string, initLine: numbe
...
+  return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
+}
+
+// ---- Sandbox-routed I/O helpers ----
+
+/**
+ * Probes a path through the sandbox, reporting existence and directory-ness by listing
+ * the parent directory. A missing parent or entry resolves to `exists: false`.
+ */
+async function probeSandboxPath(sandbox: Sandbox, filePath: string): Promise<{ exists: boolean; isDir: boolean }> {
```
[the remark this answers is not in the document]
gautamsirdeshmukh: We intentionally removed statFile, no need for a comment

strands-ts/src/vended-tools/file-editor/file-editor.ts
```diff
@@ -171,284 +224,170 @@ function makeOutput(fileContent: string, fileDescriptor: string, initLine: numbe
...
+  const { content, initLine } = applyViewRange(fileContent, viewRange)
+  return makeOutput(content, filePath, initLine)
 }
 
-/**
- * Handles the create command.
- */
-async function handleCreate(filePath: string, fileText: string): Promise<string> {
+async function handleCreate(sandbox: Sandbox, filePath: string, fileText: string): Promise<string> {
   if (fileText === undefined) {
```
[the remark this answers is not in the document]
gautamsirdeshmukh: As the bot already says, the current guard is sufficient defensively

strands-ts/src/vended-plugins/skills/agent-skills.ts
```diff
@@ -117,6 +135,7 @@ export class AgentSkills implements Plugin {
    */
   async initAgent(agent: LocalAgent): Promise<void> {
     await this._ready
+    await this._loadSkillPaths(agent.sandbox)
```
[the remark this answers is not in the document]
gautamsirdeshmukh: `_loadSkillPaths` early returns when `this._skillPaths` is empty (after initial calls it's a no-op if nothing has changed)

strands-ts/src/vended-plugins/context-offloader/storage.ts
```diff
@@ -260,3 +263,99 @@ export class S3Storage implements Storage {
...
+    const sandbox = await this._ensureSandbox()
+
+    const filename = `${Date.now()}_${++this._counter}_${sanitizeId(key)}${SandboxStorage._extensionFor(contentType)}`
+    const filePath = `${this._artifactDir}/${filename}`
+
+    this._contentTypes[filename] = contentType
+    this._metadataWriteChain = this._metadataWriteChain.then(() => this._saveMetadata(sandbox))
+    await this._metadataWriteChain
+
+    await sandbox.writeFile(filePath, content)
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Preexisting from FileStorage design, will address in a separate PR

strands-ts/src/vended-plugins/skills/agent-skills.ts
```diff
@@ -263,7 +249,62 @@ export class AgentSkills implements Plugin {
...
+
+    for (const skillPath of this._skillPaths) {
+      try {
+        let entries: FileInfo[]
+        try {
+          entries = await sandbox.listFiles(skillPath)
+        } catch {
+          // Not a directory: accept a direct path to a SKILL.md file, as Skill.fromFile does.
+          if (skillPath.toLowerCase().endsWith('skill.md')) {
+            await loadSkill(skillPath.slice(0, skillPath.lastIndexOf('/')), skillPath)
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Skill paths are user-provided absolute paths, a bare filename would never reach this code

strands-ts/src/vended-plugins/context-offloader/storage.ts
```diff
@@ -260,3 +263,99 @@ export class S3Storage implements Storage {
...
+    this._artifactDir = artifactDir
+    this._sandbox = sandbox
+  }
+
+  /** @internal Set by ContextOffloader during plugin init when no sandbox was provided at construction. */
+  set sandbox(value: Sandbox) {
+    if (!this._sandbox) this._sandbox = value
+  }
+
+  private static _extensionFor(contentType: string): string {
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Preexisting and if addressed should go in a separate chore PR

strands-ts/src/sandbox/base.ts
mkmeral: should I be able to override it though? Whenever we add these magic strings, there is always this risk of either being too commanding (then your strings get caught in guardrails), or being too permissive (then LLM doesn't listen to you). This makes choosing magic strings hard/impossible. I think there should be a way to override it. 

I am okay with the method for convention and so on, but the feature to be able to adjust should also exist

strands-ts/src/vended-tools/bash/bash.ts
```diff
@@ -248,49 +250,81 @@ process.on('SIGTERM', () => {
...
-    const agent = context.agent
+export const SANDBOX_BASH_DESCRIPTION =
+  'Executes bash shell commands. Each call runs in a fresh shell; ' +
+  'state such as variables and the working directory does not persist across calls.'
 
-    // Validate execute mode has command
-    if (input.mode === 'execute' && !input.command) {
-      throw new Error('command is required when mode is "execute"')
-    }
+export interface MakeBashOptions {
```
mkmeral: any reason we don't support input schema, and name too?

strands-ts/src/vended-plugins/skills/agent-skills.ts
```diff
@@ -263,7 +249,62 @@ export class AgentSkills implements Plugin {
...
+            const childDir = `${skillPath}/${entry.name}`
+            const childMd = findSkillMdName(await sandbox.listFiles(childDir))
+            if (childMd) await loadSkill(childDir, `${childDir}/${childMd}`)
+          }
+        }
+      } catch (error) {
+        logger.warn(`path=<${skillPath}> | failed to load skill from sandbox: ${error}`)
+      }
+    }
+    this._skillPaths = []
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Addressing this, storing skills per agent

strands-ts/src/sandbox/docker.ts
```diff
@@ -76,4 +79,8 @@ export class DockerSandbox extends PosixShellSandbox {
       enoentMessage: 'docker is not installed or not on PATH',
     })
   }
+
+  override getTools(): Tool[] {
+    return [fileEditor, makeBash({ description: SANDBOX_BASH_DESCRIPTION })]
```
mkmeral: let's namespace tool names

gautamsirdeshmukh: Updating to namespace with `sandbox_{toolname}` the way MCP adds `mcp_{toolname}`

strands-ts/src/vended-tools/bash/bash.ts
gautamsirdeshmukh: Updating to take name and inputSchema along with description

strands-ts/src/vended-tools/bash/make-bash.ts
```diff
@@ -0,0 +1,52 @@
...
+    callback: async (input, context) => {
+      if (!context) {
+        throw new Error('Tool context is required for bash operations')
+      }
+
+      if (input.mode === 'execute' && !input.command) {
+        throw new Error('command is required when mode is "execute"')
+      }
+
+      if (input.mode === 'restart') {
```
mkmeral: why is it even in the modes then? this should just be invalid mode.

strands-ts/src/vended-tools/file-editor/file-editor.ts
```diff
@@ -29,21 +29,11 @@ const fileEditorInputSchema = z.object({
...
-    return await fs.readFile(filePath, 'utf-8')
-  }
-}
-
 /**
  * File editor tool for viewing, creating, and editing files programmatically.
  *
  * Provides commands for viewing files/directories, creating files, string replacement,
- * and line insertion.
+ * and line insertion. All I/O routes through the agent's configured sandbox.
```
mkmeral, line 37: why don't use use `makeFileEditor` here, to update the prefix and the description? how does the agent know this tool is also in the sandbox, when bash talks about sandbox and has sandbox prefix; and this tool doesnt?

strands-ts/src/vended-tools/bash/make-bash.ts
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+import { tool } from '../../tools/tool-factory.js'
+import { z } from 'zod'
+import { SandboxTimeoutError } from '../../sandbox/errors.js'
+import type { Sandbox } from '../../sandbox/base.js'
+import type { BashOutput } from './types.js'
+import { bashInputSchema, BashTimeoutError, BashSessionError, SANDBOX_BASH_DESCRIPTION } from './types.js'
+
+export interface MakeBashOptions {
```
mkmeral, line 20: include `sandbox`?

mkmeral: uu i see separately below. why separate? is the main usage of makeBash adding sandbox? 

jfyi, i can accept a pushback, but it feels a bit weird. that said, tool schema grouping also makes sense 🤔

strands-ts/src/vended-tools/bash/make-bash.ts
gautamsirdeshmukh: Removing now that this is essentially a separate tool

strands-ts/src/vended-tools/file-editor/file-editor.ts
gautamsirdeshmukh: Adding makeFileEditor factory pattern for the sake of multi agent binding / consistency with the bash tool in that regard, but we can use the same description for both modes (unlike bash which needed two different ones)

strands-ts/src/vended-tools/bash/make-bash.ts
gautamsirdeshmukh: Sandbox is the main point of why we have makeBash in the first place, furthermore I think it makes sense to have it as a separate param since it is the main target of the factory whereas the other fields are specifically tool-spec overrides.

strands-ts/src/vended-plugins/skills/agent-skills.ts
```diff
@@ -263,7 +257,68 @@ export class AgentSkills implements Plugin {
...
+   * directory of skill subdirectories. Per-path failures are logged and skipped.
+   */
+  private async _loadSkillPaths(agent: LocalAgent): Promise<void> {
+    const skills = new Map(this._skills)
+    if (this._skillPaths.length === 0) {
+      this._agentSkills.set(agent, skills)
+      return
+    }
+
+    const sandbox = agent.sandbox
```
gautamsirdeshmukh, line 276: Add comment

strands-ts/src/vended-plugins/context-offloader/storage.ts
```diff
@@ -260,3 +263,99 @@ export class S3Storage implements Storage {
...
+ * Sandbox-backed storage backend.
+ *
+ * Stores offloaded content as files inside the agent's configured {@link Sandbox}
+ * (Docker container, remote host, etc.), keeping large tool results out of the host
+ * process. A `.metadata.json` sidecar in the artifact directory tracks content types.
+ *
+ * The sandbox can be passed at construction or injected later by {@link ContextOffloader}
+ * during plugin initialization. `store`/`retrieve` throw until a sandbox is available.
+ */
+export class SandboxStorage implements Storage {
```
gautamsirdeshmukh: Fold into FileStorage

strands-ts/src/vended-plugins/skills/agent-skills.ts
```diff
@@ -263,7 +257,68 @@ export class AgentSkills implements Plugin {
...
+          logger.warn(`name=<${skill.name}> | duplicate skill name, overwriting previous skill`)
+        }
+        skills.set(skill.name, skill)
+      } catch (error) {
+        logger.warn(`path=<${skillDir}> | failed to load skill: ${error}`)
+      }
+    }
+
+    for (const skillPath of this._skillPaths) {
+      try {
```
gautamsirdeshmukh: Remove horrible double try

strands-ts/src/sandbox/docker.ts
```diff
@@ -76,4 +80,18 @@ export class DockerSandbox extends PosixShellSandbox {
       enoentMessage: 'docker is not installed or not on PATH',
     })
   }
+
+  override getTools(): Tool[] {
+    const cwd = this.workingDir ? ` Working directory: ${this.workingDir}.` : ''
+    return [
+      makeFileEditor({
```
gautamsirdeshmukh: Have specific makeSandboxBash / makeSandboxFileEditor for now (take/use sandbox instead of having it as part of options), eventually replace with more generalized tool cloning/customization call

strands-ts/src/vended-tools/file-editor/file-editor.ts
```diff
@@ -60,49 +50,87 @@ class TextFileReader implements IFileReader {
...
+          throw new Error(`Unknown command: ${input.command}`)
+      }
+    },
+  })
+}
+
+/**
+ * Default file editor tool. Reads the sandbox from the agent's context at call time.
+ */
 export const fileEditor = tool({
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Changing name back to makeFileEditor (and makeBash), updating fileEditor tool to call makeFileEditor to prevent duplication

strands-ts/src/vended-tools/file-editor/file-editor.ts
```diff
@@ -171,284 +277,170 @@ function makeOutput(fileContent: string, fileDescriptor: string, initLine: numbe
...
+ * the parent directory. A missing parent or entry resolves to `exists: false`.
+ */
+async function probeSandboxPath(sandbox: Sandbox, filePath: string): Promise<{ exists: boolean; isDir: boolean }> {
+  const normalized = filePath.replaceAll('\\', '/')
+  const parent = normalized.split('/').slice(0, -1).join('/') || '/'
+  const name = normalized.split('/').pop()!
+  try {
+    const entry = (await sandbox.listFiles(parent)).find((e) => e.name === name)
+    return entry ? { exists: true, isDir: entry.isDir ?? false } : { exists: false, isDir: false }
+  } catch {
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Adding proper SandboxPathNotFoundError and updating error handling

strands-ts/src/sandbox/docker.ts
gautamsirdeshmukh: Actually decided to stick with makeBash and makeFileEditor for now

strands-ts/src/index.ts
```diff
@@ -305,6 +306,7 @@ export { AgentMetrics } from './telemetry/meter.js'
 // Sandbox
 export { Sandbox, type ExecuteOptions } from './sandbox/base.js'
 export { PosixShellSandbox } from './sandbox/posix-shell.js'
+export { SandboxTimeoutError, SandboxAbortError } from './sandbox/errors.js'
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Adding

strands-ts/src/sandbox/not-a-sandbox-local-environment.ts
```diff
@@ -0,0 +1,101 @@
...
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Addressed in https://github.com/strands-agents/harness-sdk/pull/2563

strands-ts/src/sandbox/docker.ts
```diff
@@ -6,6 +6,9 @@ import type { ExecuteOptions } from './base.js'
 import { PosixShellSandbox, validateEnvKeys } from './posix-shell.js'
 import { streamProcess } from './stream-process.js'
 import type { ExecutionResult, StreamChunk } from './types.js'
+import type { Tool } from '../tools/tool.js'
+import { makeSandboxFileEditor, DEFAULT_FILE_EDITOR_DESCRIPTION } from '../vended-tools/file-editor/index.js'
```
mkmeral: why do we still have this? weren't we changing to `makeFileEditor`?

strands-ts/src/sandbox/docker.ts
```diff
@@ -76,4 +79,16 @@ export class DockerSandbox extends PosixShellSandbox {
       enoentMessage: 'docker is not installed or not on PATH',
     })
   }
+
+  override getTools(): Tool[] {
+    const cwd = this.workingDir ? ` Working directory: ${this.workingDir}.` : ''
+    return [
+      makeSandboxFileEditor(this, {
```
mkmeral: are we adding a tool name prefix? how?

mkmeral: how does the agent know they are connected

strands-ts/src/agent/agent.ts
```diff
@@ -707,6 +736,27 @@ export class Agent implements LocalAgent, InvokableAgent {
       })
     )
 
+    // Register tools vended by an explicitly-configured sandbox, applying the sandbox's
+    // toolPrefix to names (like MCP's prefix for server-vended tools).
+    if (this._sandbox) {
+      const prefix = this._sandbox.toolPrefix
```
mkmeral, line 742: so on the prefix. i think the prefixes should be handled by the sandboxes themselves. i dont want us to update tool names in agent. that's a mix/confusion of responsibilities imo

that said, i'll approve, and it's completely internal so we can fix in followup PR, or if you want we can fix it here

mkmeral (dismissed):

mkmeral (approved):

PR #2653, opened 2026-06-06
fix: preserve non-ASCII tool result text

### LINKED ISSUE #2636: [BUG] Tool results with non-ASCII text are serialized inconsistently — dict returns are \uXXXX-escaped, Pydantic models are not
> [!NOTE]
> **Scope update (split).** This issue originally bundled two things: (1) the `@tool` decorator inconsistency (dict vs Pydantic) and (2) the provider-side `json.dumps` escaping. PR #2653 fixes **only the decorator path (1)** and uses `Resolves #2636`, which auto-closes this issue on merge. The **provider-side paths (2)** are a distinct, larger scope (multiple providers, triggered e.g. by multi-agent / agent-as-tool structured output rather than the decorator), so they have been split into a dedicated issue **#2660** (PR **#2661**) to avoid being silently dropped when #2653 closes this one.
>
> - Decorator path (this issue) → PR #2653
> - Provider-side paths → #2660 / PR #2661

### Checks

### SDK Language

Python

### Strands Version

1.39.0 (also reproduces on `main` @ `526e48d5`; originally hit on 1.30.0)

### Language Runtime Version

Python 3.11.11

### Operating System

macOS (Darwin 25.5.0)

### Installation Method

pip

### Steps to Reproduce

When a `@tool` returns a value that is not already a `ToolResult`, `DecoratedFunctionTool._wrap_tool_result()` serializes it to the result `text` using **two different serializers depending on the return type** (`strands-py/src/strands/tools/decorator.py`, around L677-L686):

```python
elif isinstance(result, BaseModel):
    text = result.model_dump_json()      # Pydantic: keeps UTF-8 as-is
else:
    text = json.dumps(result)            # dict/list: json.dumps default ensure_ascii=True -> escaped
```

Minimal reproducible example (no model/API call required):

```python
import asyncio
from strands import tool

@tool
def dict_tool() -> dict:
    return {"message": "こんにちは"}

class Reply(BaseModel):
    message: str

@tool
def model_tool() -> Reply:
    return Reply(message="こんにちは")

async def run(t):
    last = None
    async for ev in t.stream({"toolUseId": "1", "name": t.tool_name, "input": {}}, {}):
        last = ev
    return last.tool_result["content"][0]["text"]

print("dict  tool ->", asyncio.run(run(dict_tool)))
print("model tool ->", asyncio.run(run(model_tool)))
```

Output:

```
dict  tool -> {"message": "こんにちは"}
model tool -> {"message":"こんにちは"}
```

### Expected Behavior

Both return types should produce the same, human-readable UTF-8 text in the tool result, e.g. `{"message": "こんにちは"}`. Serialization of non-ASCII content should not depend on whether the tool returns a `dict`/`list` or a Pydantic model.

### Actual Behavior

A tool returning a `dict` (or `list`) emits `\uXXXX`-escaped text, while the same data returned as a Pydantic model is kept as UTF-8. For CJK / non-Latin scripts this means:

- The escaped string is what gets fed to the model as the tool result and shown in traces/logs, hurting readability and debuggability.
- Each non-ASCII character expands to a 6-char `\uXXXX` sequence, inflating token usage (and therefore cost) of the tool-result content.
- Behavior is inconsistent between two otherwise-equivalent ways of returning structured data.

### Additional Context

The same `json.dumps(...)` default-escaping (`ensure_ascii=True`) also recurs across the model providers (tool-result `{"json": ...}` blocks and tool-call argument serialization). Those provider-side paths have been **split into #2660** (with PR #2661) — see that issue for the full list of locations, the reproduction, and scope notes (Gemini/Bedrock pass these natively and are unaffected; streaming-delta serializations are no-ops because they round-trip through `json.loads`).

Note: the TypeScript SDK is **not** affected — `JSON.stringify` does not escape non-ASCII by default. This is Python-specific because `json.dumps` defaults to `ensure_ascii=True`.

### Possible Solution

Pass `ensure_ascii=False` to the `json.dumps(...)` call so the dict path matches `model_dump_json()`:

```python
else:
    text = json.dumps(result, ensure_ascii=False)
```

This is implemented for the decorator in PR #2653; the provider-side paths are handled in #2660 / PR #2661.

### Related Issues

- #2653 — fixes the decorator path (this issue)
- #2660 / #2661 — provider-side follow-up

cyphercodes:
## Description
`@tool` results returned as dictionaries or lists currently escape CJK and emoji characters because the standard JSON serializer defaults to ASCII escaping, while Pydantic model returns keep UTF-8 text. This keeps structured tool-result text human-readable regardless of whether a tool returns a dict, list, or model.

## Related Issues

Resolves #2636

## Documentation PR

No documentation changes are needed for this bug fix.

## Type of Change

Bug fix

## Testing

- `uv run --extra dev pytest tests/strands/tools/test_decorator.py -vv`
- `uv run --extra dev ruff format --check src/strands/tools/decorator.py tests/strands/tools/test_decorator.py`
- `uv run --extra dev ruff check src/strands/tools/decorator.py tests/strands/tools/test_decorator.py`
- `uv run --extra dev mypy src/strands/tools/decorator.py`

## Checklist
- [x] Documentation changes are not needed for this bug fix
- [x] No new docs example is needed
- [x] Any dependent changes have been merged and published, or are not needed

----

### DIFF
--- a/strands-py/src/strands/tools/decorator.py
+++ b/strands-py/src/strands/tools/decorator.py
@@ -681,7 +681,7 @@ class DecoratedFunctionTool(AgentTool, Generic[P, R]):
                     text = str(result)
             else:
                 try:
-                    text = json.dumps(result)
+                    text = json.dumps(result, ensure_ascii=False)
                 except (TypeError, ValueError):
                     text = str(result)
 
--- a/strands-py/tests/strands/tools/test_decorator.py
+++ b/strands-py/tests/strands/tools/test_decorator.py
@@ -8,7 +8,7 @@ from typing import Annotated, Any
 from unittest.mock import MagicMock
 
 import pytest
-from pydantic import Field
+from pydantic import BaseModel, Field
 
 import strands
 from strands import Agent
@@ -601,6 +601,43 @@ async def test_tool_decorator_with_different_return_values(alist):
     result = (await alist(stream))[-1]
     assert result["tool_result"]["status"] == "success"
     assert result["tool_result"]["content"][0]["text"] == "null"
+
+
+@pytest.mark.asyncio
+async def test_tool_decorator_preserves_non_ascii_structured_return_values(alist):
+    """Test dict/list returns preserve non-ASCII text like Pydantic models."""
+
+    class Reply(BaseModel):
+        message: str
+
+    @strands.tool
+    def dict_return_tool() -> dict[str, str]:
+        """Return a dict with non-ASCII content."""
+        return {"message": "こんにちは 🌏"}
+
+    @strands.tool
+    def list_return_tool() -> list[str]:
+        """Return a list with non-ASCII content."""
+        return ["你好", "🙂"]
+
+    @strands.tool
+    def model_return_tool() -> Reply:
+        """Return a Pydantic model with non-ASCII content."""
+        return Reply(message="こんにちは 🌏")
+
+    tool_use: ToolUse = {"toolUseId": "test-id", "name": "test-tool", "input": {}}
+
+    for tool, expected_fragments in (
+        (dict_return_tool, ('"message": "こんにちは 🌏"',)),
+        (list_return_tool, ('"你好"', '"🙂"')),
+        (model_return_tool, ('"message":"こんにちは 🌏"',)),
+    ):
+        events = await alist(tool.stream(tool_use, {}))
+        text = events[-1]["tool_result"]["content"][0]["text"]
+
+        assert "\\u" not in text
+        for fragment in expected_fragments:
+            assert fragment in text
 
 
 @pytest.mark.asyncio


yatszhash:
Thanks for fixing the decorator path here! 🙏

Heads-up on scope: the **provider-side** `json.dumps` escaping that #2636 also listed — tool-result `{"json": ...}` content blocks and tool-call argument serialization across Anthropic / OpenAI-family providers (hit in practice by e.g. multi-agent / agent-as-tool structured output) — isn't covered by this PR. I've split it into #2660 and opened PR #2661 for it.

This PR's `Resolves #2636` is fine to keep — #2660 tracks the provider-side remainder independently.

poshinchen (approved):

PR #2656, opened 2026-06-07
fix: handle empty Bedrock content blocks

### LINKED ISSUE #2652: [BUG] Empty content block ({}) in BedrockModel raises bare StopIteration → corrupts Future under asyncio.to_thread (Py 3.11 hang)
### Checks

- [x] I have searched ./issues and there are no duplicates of my issue (related: #1993, #1056 — same line 740, different trigger; neither covers the empty-dict / `StopIteration` case)

### SDK Language

Python

### Strands Version

1.42.0 (line unchanged on `main` as of this report)

### Language Runtime Version

Python 3.11

### Operating System

Amazon Linux 2023 (reproduces on any Python 3.11)

### Installation Method

pip

### Steps to Reproduce

`BedrockModel._format_request_message_content` (`strands/models/bedrock.py:740`) ends with:

```python
raise TypeError(f"content_type=<{next(iter(content))}> | unsupported type")
```

When a content block is an **empty dict** `{}` (no recognized key), `next(iter(content))` is evaluated *while building the f-string* — and on an empty dict it raises a bare **`StopIteration`** before the intended `TypeError` is ever constructed.

Minimal repro (no async, no OTEL — shows the bare `StopIteration` directly):

```python

m = BedrockModel(model_id="us.anthropic.claude-sonnet-4-5-20250929-v1:0")
# A message whose content list contains an empty block:
m._format_bedrock_messages([{"role": "user", "content": [{}]}])
# -> raises StopIteration (not the intended TypeError)
```

In real use this surfaces when conversation **history** contains a message with an empty content block (e.g. a tool/UI integration that persists a block with no recognized key). The block falls through every `if "x" in content` branch to line 740.

### Expected Behavior

An unsupported or empty content block should raise a catchable, descriptive `TypeError` (as the code intends and the docstring promises: "Raises: TypeError: If the content block type is not supported by Bedrock."), regardless of whether the block is empty.

### Actual Behavior

A bare `StopIteration` escapes. Because `_format_request_message_content` is reached via `_stream`, which the SDK dispatches through `asyncio.to_thread`, the `StopIteration` propagates into the chained `concurrent.futures.Future`. On **Python 3.11**, `Future.set_exception` rejects it:

```
TypeError: StopIteration interacts badly with generators and cannot be raised into a Future
```

This fires in a **detached `_chain_future._set_state` callback** with **no application frames** in the visible traceback — the originating `bedrock.py:740` is erased from the surfaced error, making it extremely hard to diagnose (the agent simply hangs, with only asyncio-internal frames shown). Captured stack at the point of escape:

```
  File ".../strands/models/bedrock.py", line 963, in _stream
    request = self._format_request(messages, tool_specs, system_prompt_content, tool_choice)
  File ".../strands/models/bedrock.py", line 274, in _format_request
    "messages": self._format_bedrock_messages(messages),
  File ".../strands/models/bedrock.py", line 489, in _format_bedrock_messages
    formatted_content = self._format_request_message_content(content_block)
  File ".../strands/models/bedrock.py", line 740, in _format_request_message_content
    raise TypeError(f"content_type=<{next(iter(content))}> | unsupported type")
                                     ^^^^^^^^^^^^^^^^^^^
StopIteration
```

(On Python 3.12+ the symptom changes — `Future.set_exception` wraps `StopIteration` as a `RuntimeError` rather than re-raising `TypeError` — so the turn fails fast instead of hanging indefinitely, but the underlying bug remains.)

### Additional Context

The hang is especially severe because the offending content block lives in conversation **history**: once an empty block is persisted, *every subsequent turn* that loads that history hits line 740 and wedges, even though the turn itself is otherwise valid.

### Possible Solution

Make line 740 robust to empty/keyless content so it always raises the intended `TypeError`, never a bare `StopIteration`:

```python
content_type = next(iter(content), None)
raise TypeError(f"content_type=<{content_type}> | unsupported type")
```

The key point: `next(iter(content))` must never be the expression that raises, because a bare `StopIteration` is mis-propagated by asyncio when the call runs under `asyncio.to_thread`. An explicit empty-block guard at the top of the method would also work.

Note #1993 hardened this same path by adding `reasoningContent` to a skip-list, but the underlying `next(iter(...))` fragility — and specifically the empty-dict case — were not addressed.

### Related Issues

- #1993 (OpenAIResponsesModel / `reasoningContent` block → `TypeError` at this line)
- #1056 (`reasoning_content` unsupported type)

Both hit line 740 with a *populated* unsupported block, so `next(iter(content))` returns a key and the intended `TypeError` is raised cleanly. This report is the **empty-dict `{}` → bare `StopIteration`** variant, which additionally triggers the async Future corruption / hang.

he-yufeng:
Fixes #2652.

## Summary
- avoid calling `next(iter(content))` without a default for unsupported Bedrock content blocks
- make an empty content block raise the intended descriptive `TypeError` instead of leaking `StopIteration`
- add a regression test through `_format_bedrock_messages`, matching the reported history-path failure

## To verify
- `$env:PYTHONPATH=(Resolve-Path .\src).Path; python -m pytest tests\strands\models\test_bedrock.py -q -k "empty_block or empty_tool_result"`
- `$env:PYTHONPATH=(Resolve-Path .\src).Path; python -m pytest tests\strands\models\test_bedrock.py -q`
- `$env:PYTHONPATH=(Resolve-Path .\src).Path; python -m py_compile src\strands\models\bedrock.py tests\strands\models\test_bedrock.py`
- `$env:PYTHONPATH=(Resolve-Path .\src).Path; python -m ruff check src\strands\models\bedrock.py tests\strands\models\test_bedrock.py`
- `git diff --check`

Local note: this Windows environment has an older editable Strands checkout on `sys.path`, so the validation commands set `PYTHONPATH=./src` to force imports from this worktree.

### DIFF
--- a/strands-py/src/strands/models/bedrock.py
+++ b/strands-py/src/strands/models/bedrock.py
@@ -737,7 +737,8 @@ class BedrockModel(Model):
 
             return {"citationsContent": result}
 
-        raise TypeError(f"content_type=<{next(iter(content))}> | unsupported type")
+        content_type = next(iter(content), None)
+        raise TypeError(f"content_type=<{content_type}> | unsupported type")
 
     def _has_blocked_guardrail(self, guardrail_data: dict[str, Any]) -> bool:
         """Check if guardrail data contains any blocked policies.
--- a/strands-py/tests/strands/models/test_bedrock.py
+++ b/strands-py/tests/strands/models/test_bedrock.py
@@ -1894,6 +1894,13 @@ def test_format_request_message_content_does_not_mutate_empty_tool_result(model,
     assert original_content == [], "Original empty content list should not be mutated"
 
 
+def test_format_request_message_content_empty_block_raises_type_error(model, model_id):
+    messages = [{"role": "user", "content": [{}]}]
+
+    with pytest.raises(TypeError, match="content_type=<None> \\| unsupported type"):
+        model._format_bedrock_messages(messages)
+
+
 def test_format_request_message_content_preserves_nonempty_tool_result_content(model, model_id):
     """Test that _format_request_message_content does not modify non-empty toolResult content."""
     messages = [


mehtarac (approved):
Clean, minimal fix. The `next(iter(content), None)` default prevents bare `StopIteration` from escaping and the regression test covers the exact path through `_format_bedrock_messages`. LGTM.

PR #2671, opened 2026-06-08
feat: add memory extraction

opieter-aws:
## Description

Adds the **write-side extraction layer** for `MemoryManager`: automatic, configurable persistence of conversation messages to memory stores, built on top of the tool-driven `MemoryManager` / `MemoryStore` primitives from #2544.

Extraction is configured **per store** (via a new `extraction` field on the store config) and driven by **triggers** that decide *when* to run. What gets written depends on whether an `Extractor` is configured:

- **With an `Extractor`** (e.g. `ModelExtractor`): the trigger's messages are distilled into discrete facts (an LLM call) and written via the store's `add(content, metadata)`. This is the path for self-managed stores that hold plain text.
- **Without an extractor**: the raw, role-tagged message batch is handed straight to the store's new `addMessages(messages)` sink — no serialization, no model call. This is the sanctioned path for managed backends that extract server-side (e.g. AgentCore Memory, which maps the batch onto its own conversational write API).

Key properties:

- **Always non-blocking.** A trigger's fire dispatches extraction onto the background and returns immediately, so it never blocks the agent loop even though it runs from a lifecycle hook. `MemoryManager.flush()` awaits in-flight work for graceful shutdown / tests.
- **Exactly-once-ish, eviction-safe.** Messages are buffered with a monotonic sequence number, and each store advances its own high-water mark, so repeated fires never re-process the same messages, and context-window eviction never drops un-extracted messages.
- **Fail-fast configuration.** The matching write sink is validated at construction: an extractor requires `add`; no extractor requires `addMessages`.
- **Per-store serialization.** A store's extraction runs never overlap or reorder; failures are logged and swallowed (extraction is best-effort and must never break the loop).

> Note: extraction is eventually consistent — a write triggered on one turn may still be in flight when the next turn begins. Call `flush()` on shutdown to avoid losing in-flight writes.

### New / changed public API

**`MemoryStoreConfig` (new field)**
```typescript
interface MemoryStoreConfig {
  // ...existing: name, description, maxSearchResults, writable
  /** Per-store automatic-extraction config. Requires the store to be writable. */
  readonly extraction?: ExtractionConfig
}
```

**`MemoryStore` (new optional sink)**
```typescript
interface MemoryStore extends MemoryStoreConfig {
  // ...existing: writable, search(), add?(), getTools?()
  /**
   * Ingest a batch of conversation messages, roles preserved. The no-extractor sink: when a store's
   * extraction config has no extractor, the manager hands the filtered MessageData[] batch here in
   * one call. A store satisfies `writable: true` with `add`, `addMessages`, or both.
   */
  addMessages?(messages: MessageData[]): Promise<unknown>
}
```

**`MemoryManager` (new method)**
```typescript
class MemoryManager implements Plugin {
  // ...existing: search(), add(), getTools(), initAgent()
  /** Await all in-flight background extraction. No-op when extraction is not configured. */
  flush(): Promise<void>
}
```

**`ExtractionConfig`**
```typescript
interface ExtractionConfig {
  /** When to run extraction. A single trigger or an array (an empty array is rejected). */
  trigger: ExtractionTrigger | ExtractionTrigger[]
  /**
   * How to turn messages into entries. When set, the store must implement `add`. When omitted, the
   * filtered messages go straight to the store's `addMessages` (which it must then implement).
   */
  extractor?: Extractor
  /** Content blocks to strip before extraction. Default: exclude ['toolUse', 'toolResult']. */
  filter?: MessageFilter
}
```

**Triggers**
```typescript
// Abstract base — extend for custom cadence (message-count, time, size, ...).
abstract class ExtractionTrigger {
  abstract readonly name: string
  abstract attach(context: ExtractionTriggerContext): void // wire hooks; call context.fire()
}

interface ExtractionTriggerContext {
  agent: LocalAgent
  /** Fire-and-forget: dispatches extraction for this trigger's store; never blocks the loop. */
  fire: () => void
}

// Built-ins:
class InvocationTrigger extends ExtractionTrigger {} // runs after every agent invocation

interface IntervalTriggerOptions {
  turns: number // positive integer
}
class IntervalTrigger extends ExtractionTrigger {
  constructor(options: IntervalTriggerOptions) // runs every N invocations
}
```

**Extractors**
```typescript
interface ExtractionResult {
  content: string
  metadata?: Record<string, JSONValue>
}

interface ExtractorContext {
  defaultModel?: Model // the agent's model, supplied by the manager
}

interface Extractor {
  extract(messages: MessageData[], context?: ExtractorContext): Promise< ExtractionResult[]>
}

interface ModelExtractorOptions {
  model?: Model         // default: the agent's own model (via ExtractorContext.defaultModel)
  systemPrompt?: string // default: a general fact-extraction prompt
}
class ModelExtractor implements Extractor {
  constructor(options?: ModelExtractorOptions) // calls the model, parses a JSON array of entries
}
```

**Message filtering**
```typescript
// Derived from the SDK's ContentBlockData union (tracks new block types automatically), resolving to:
//   'text' | 'toolUse' | 'toolResult' | 'reasoning' | 'image'
//   | 'video' | 'document' | 'citations' | 'guardContent' | 'cachePoint'
type MemoryContentBlockType = DistributedKeyof<ContentBlockData>

interface MessageFilter {
  exclude: MemoryContentBlockType[]
}
// Default when `filter` is omitted: { exclude: ['toolUse', 'toolResult'] }
```

### Usage

**Server-side extraction backend (no client extractor) — e.g. AgentCore Memory**
```typescript
// Store implements `addMessages` (maps the role-tagged batch onto its write API) and no extractor.
const agent = new Agent({
  model,
  memoryManager: {
    stores: [
      myManagedStore, // implements addMessages(), writable: true, extraction set internally
    ],
  },
})
// On each AfterInvocationEvent, the filtered MessageData[] batch is handed to addMessages() in the
// background. search_memory stays on for recall; add_memory left off.
```

**Self-managed store with LLM distillation**
```typescript
import { Agent, MemoryManager, ModelExtractor, InvocationTrigger } from '@strands-agents/sdk'

const store = new MyTextStore({
  name: 'personal',
  writable: true, // implements add()
  extraction: {
    trigger: new InvocationTrigger(),                 // single trigger (or an array)
    extractor: new ModelExtractor({                   // distills facts via the model
      model: cheapModel,                              // default: the agent's own model
      systemPrompt: 'Extract user preferences and decisions as discrete facts.',
    }),
    // filter omitted -> defaults to excluding toolUse / toolResult
  },
})

const agent = new Agent({ model, memoryManager: { stores: [store] } })
```

**Interval cadence + composed triggers + custom filter**
```typescript
extraction: {
  trigger: [new IntervalTrigger({ turns: 5 }), new InvocationTrigger()], // fires on any
  filter: { exclude: [] }, // keep everything, including tool traffic
}
```

**Awaiting background writes (shutdown / tests)**
```typescript
const memoryManager = new MemoryManager({ stores: [store] })
const agent = new Agent({ model, memoryManager })
await agent.invoke('...')
await memoryManager.flush() // ensure in-flight extraction writes have landed
```

## Related Issues

Builds on #2544 (base `MemoryManager` / `MemoryStore`).

## Documentation PR

Docs to follow once the full feature surface (extraction + a shipped store backend + injection) is finalized.

## Type of Change

New feature

## Testing

How have you tested the change? Verify that the changes do not break functionality or introduce warnings in consuming repositories: agents-docs, agents-tools, agents-cli

Unit tests in `strands-ts/src/memory/extraction/__tests__/` (vitest) cover: trigger firing (invocation / interval / single vs array / composed), high-water-mark dedup (incl. failure rollback/retry), the two routing paths (extractor → `add`, no-extractor → `addMessages`), the message filter (default + custom + empty-message drop), construction-time sink validation, background/non-blocking execution, and `flush()` semantics. `ModelExtractor` JSON parsing is covered against a mock model (well-formed, fenced, malformed, empty, default-model fallback). All memory tests pass (92); `type-check`, `eslint`, `prettier --check`, and the browser bundle are clean.

## Checklist

----


strands-ts/src/memory/extraction/triggers.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+/** Options for {@link IntervalTrigger}. */
+export interface IntervalTriggerOptions {
+  /** Run extraction once every this many invocations. Must be a positive integer. */
+  turns: number
+}
+
+/**
+ * Runs extraction every N agent invocations.
```
mkmeral, line 34: what happens if process dies beforehand?

strands-ts/src/memory/extraction/model-extractor.ts
```diff
@@ -0,0 +1,143 @@
...
+import { logger } from '../../logging/logger.js'
+import type { ExtractedEntry, Extractor, ExtractorContext } from './types.js'
+
+/** Default instruction guiding the model to emit discrete, durable facts as a JSON array. */
+const DEFAULT_SYSTEM_PROMPT = `You extract durable facts worth remembering across future conversations from a transcript.
+
+Return ONLY a JSON array of objects, each: {"content": string}. Each object is one discrete, self-contained fact (a preference, decision, or stable detail about the user or task). Do not include transient chit-chat, questions, or anything already obvious. If there is nothing worth remembering, return [].`
+
+/** Options for {@link ModelExtractor}. */
+export interface ModelExtractorOptions {
```
mkmeral, line 13: is this the best place for it? should we have like vended stores or sth? 🤔

strands-ts/src/memory/extraction/triggers.ts
opieter-aws: Good catch, it's two-faced:

1. Graceful shutdown. `flush()` exists (awaits in-flight extraction), but two gaps:
   * Nothing calls it automatically. There's no agent shutdown/dispose lifecycle to hang it on, so today it's on the caller to `await memoryManager.flush()` at session end.
   * More subtle: `flush()` only awaits writes that were *already dispatched*. It doesn't force a run of lingering turns. So a trigger that hadn't fired yet (e.g. `IntervalTrigger({turns: 5})` and the session ends at turn 3) flushes nothing; those turns are lost even on a graceful exit. To be a real boundary guarantee, `flush()` should force-extract each store's pending buffer, then await. Small change that I'll implement.

The fix for graceful death can be to give an opt-in auto-flush on `AfterInvocationEvent`. The tradeoff is latency: awaiting blocks the response, which cuts against keeping extraction off the hot path, so it should be opt-in. Makes sense to add, let me know if you agree.

2. Ungraceful death (kill -9, OOM, hard timeout). This is the documented 'best-effort' limit. The buffer is in-memory and fire-and-forget by design; a hard kill before flush loses the tail, and we don't try to recover it on resume.

We _could_ fix this with session restore, to scan for un-extracted messages. That's however not an easy mechanism to implement, since we would need to store the hwm in agent state, find a way to track failed invocations, and avoid duplication. So I propose to keep this out of scope for now, we can always decide to follow up with this.

strands-ts/src/memory/extraction/model-extractor.ts
opieter-aws: I agree a vended folder is the right call for memory store implementations. Each store is a concrete per-backend integration.

But I'd prefer for `ModelExtractor` specifically to stay in core, because it's shared across all stores. 

If we wanted to be stricter (core is interfaces only), an alternative could be vending the whole generic tier (`ModelExtractor` + triggers) as a `vended-memory` package. Defensible, but it costs the happy-path import to isolate a dependency that doesn't actually exist. 

Let me know if you agree here

mkmeral: okay so the argument is "this is not a complete store, it can be used to build a store" so it shouldn't just be a vended plugin." right? sure, that makes sense

strands-ts/src/memory/extraction/model-extractor.ts
```diff
@@ -2,7 +2,7 @@ import { Message, TextBlock, type MessageData } from '../../types/messages.js'
 import type { Model } from '../../models/model.js'
 import type { JSONValue } from '../../types/json.js'
 import { logger } from '../../logging/logger.js'
-import type { ExtractedEntry, Extractor, ExtractorContext } from './types.js'
+import type { ExtractionResult, Extractor, ExtractorContext } from './types.js'
```
mkmeral, line 5: now that you changed message filter to memory message filter, you are making me question if we should change extraction result to memory extraction result 😅 

@agent-of-mkmeral  can you quickly check how we name things here, w.r.t. "memory" prefix? also do we have similar prefixes around the sdk? and how do we use it, is it used consistently on all the types?

strands-ts/src/memory/extraction/triggers.ts
mkmeral: > The fix for graceful death can be to give an opt-in auto-flush on AfterInvocationEvent

this becomes per invocation extraction though, right? I'm not necessarily sure if it solves the problem, rather than just quietly changing the settings under the hood

> Ungraceful death (kill -9, OOM, hard timeout).

I'm thinking if there are `__del__` like events that we can hook into. I am also not sure if we want to be making network calls at that time. 

Initially, I am okay with adding this as a limitation (since flush is public method, and you can work around this by doing `mm.flush()` on exit of your agentic app), and we document it. We can follow up later.

strands-ts/src/memory/extraction/model-extractor.ts
agent-of-mkmeral: Did a sweep of the memory module + the rest of the SDK. tl;dr: I'd **keep `ExtractionResult` as-is** — prefixing it to `MemoryExtractionResult` would be redundant and has no precedent in the codebase. Here's the reasoning.

### How `memory/` names things today

Two coherent prefix families, by tier:

**`Memory*`** — the manager/store-facing public surface:
`MemoryManager`, `MemoryStore`, `MemoryStoreConfig`, `MemoryEntry`, `MemorySearchOptions`, `MemoryAddOptions`, `MemoryToolConfig`, `MemoryAddToolConfig`, `MemoryManagerConfig`, `MemoryMessageFilter`, `MemoryContentBlockType`

**`Extraction*` / `Extractor*`** — the extraction sub-module:
`ExtractionConfig`, `ExtractionResult`, `ExtractionTrigger`, `ExtractionTriggerContext`, `Extractor`, `ExtractorContext`, `ModelExtractor`, `InvocationTrigger`, `IntervalTrigger`

So it's **not** "everything gets `Memory`". It's a domain prefix on the outer surface + a self-contained sub-family stem for extraction.

### Why `MessageFilter` → `MemoryMessageFilter` was right (but `ExtractionResult` is different)

The prefix isn't decorative — it's applied to **disambiguate names that would otherwise collide with core SDK vocabulary**:

- `Message` / `MessageData` / `MessageMetadata` are core SDK types → bare `MessageFilter` reads like a general-purpose message filter. `Memory` scopes it. ✅
- `ContentBlock` / `ContentBlockData` are core SDK types → `MemoryContentBlockType` says "memory's subset of block kinds". ✅

`ExtractionResult`, `Extractor`, `ExtractionConfig`, `ExtractionTrigger` don't collide with anything SDK-wide — **`Extraction` is itself the disambiguating namespace**. `MemoryExtractionResult` would be double-scoping (extraction only exists inside memory), which the codebase doesn't do anywhere.

### Is this consistent with the rest of the SDK? Yes — strongly.

This exact pattern (domain prefix on the surface + an un-prefixed sub-family) is everywhere:

- **`session/`**: `SessionManager`, `SessionManagerConfig`, `SessionStorage` — but the snapshot sub-family is `SnapshotManifest`, `SnapshotStorage`, `SnapshotTriggerParams`, `SnapshotLocation`, **not** `SessionSnapshot*`. This is the direct analogue of `Extraction*` living inside `memory/`.
- **`models/`**: `ModelContentBlockDeltaEvent` etc. are `Model`-prefixed for the *same* reason `MemoryContentBlockType` is — `ContentBlock` is core. Disambiguation, not blanket prefixing.
- **`conversation-manager/`**: `ConversationManager`, `SummarizingConversationManager`… but also `ProactiveCompressionConfig` (no prefix) — prefix applied where it clarifies, not universally.
- **`tools/`**: the family stem is `Tool*` (`ToolSpec`, `ToolContext`, `ToolUse`), self-contained, no outer prefix stacked on.

### Recommendation

1. **Keep `ExtractionResult`** (and `Extractor`, `ExtractionConfig`, `ExtractionTrigger`, `ExtractorContext`). They're correctly scoped by the `Extraction*`/`Extractor*` stem; the SDK precedent (`Snapshot*` in session) is to *not* stack the outer domain prefix on a sub-family.
2. `MemoryMessageFilter` / `MemoryContentBlockType` are the *correct* exceptions — they piggyback on core SDK nouns (`Message`, `ContentBlock`) and genuinely need the scope. So the rename you already did is consistent, not a reason to cascade.
3. If you wanted a one-line rule for the module: **prefix with `Memory` only when the bare name borrows a core-SDK noun; otherwise let the `Extraction*`/`Extractor*` family stand on its own.** That keeps it consistent both internally and with `session/`.

The only genuine "consistency" nit I'd flag: the extraction module uses two stems — `Extraction*` (the process/config/result) and `Extractor*` (the actor + its context). That's defensible (process vs. actor), but if you want a single mental model, `ExtractorContext` reads as "context for the Extractor" while `ExtractionTriggerContext` reads as "context for the extraction's trigger" — both fine, just worth a conscious "yes we're keeping two stems" rather than drifting into a third.

strands-ts/src/memory/extraction/model-extractor.ts
opieter-aws: Yes, correct

strands-ts/src/memory/extraction/triggers.ts
opieter-aws: >this becomes per invocation extraction though, right? I'm not necessarily sure if it solves the problem, rather than just quietly changing the settings under the hood

Correct. This would solve the problem for graceful death, but at the cost of latency (await flush whenever the extractor is still running). Users can then opt-in, and we'd await a flush on `AfterInvocationEvent` so the tail is never lost.

Think about a setup where `trigger = 'onEviction'` (not yet implemented, but the context window reduction event), and a conversation never hits the context window limit. Then memory is never extracted. With this opt-in, users can still get that (albeit in the critical path).

opieter-aws: Agreed to skip the opt-in, and document this well.

strands-ts/src/memory/memory-manager.ts
```diff
@@ -78,14 +93,38 @@ export class MemoryManager implements Plugin {
...
+        }
       }
     }
 
     this._config = config
     this._searchStores = config.stores
-    this._addStores = config.stores.filter((s) => s.writable)
+    // `add`-targeting paths (tool / programmatic) need an `add` method specifically.
+    this._addStores = config.stores.filter((s) => s.writable && typeof s.add === 'function')
+    this._extractionStores = config.stores.filter((s) => s.extraction)
```
mkmeral: should we also filter for `s.writable && s.extraction`?

strands-ts/src/memory/memory-manager.ts
```diff
@@ -126,18 +166,56 @@ export class MemoryManager implements Plugin {
...
+  initAgent(agent: LocalAgent): void {
+    if (this._extractionStores.length === 0) {
+      return
+    }
+
+    const coordinator = new ExtractionCoordinator(this._extractionStores, agent.model)
+    this._coordinator = coordinator
+
+    // Buffer every message the agent adds, so extraction has its own copy to save from.
+    agent.addHook(MessageAddedEvent, (event) => {
```
mkmeral, line 198: what happens if I customize `agent.messages` myself? it wouldn't get caught here, right?

strands-ts/src/memory/extraction/coordinator.ts
```diff
@@ -0,0 +1,262 @@
...
+   * It first tells every store to save (even one whose trigger hasn't fired this turn, or one that is
+   * backed off - flush bypasses backoff so a recovered store still writes its backlog); stores with
+   * nothing to save do nothing. Then it waits, re-checking until no new save has started, so saves
+   * that begin while we're waiting are also covered.
+   */
+  async flush(): Promise<void> {
+    for (const store of this._stores) {
+      void this._enqueue(store)
+    }
+    for (;;) {
```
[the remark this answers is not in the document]
opieter-aws: The loop terminates by design: `_chains` is only appended by `_enqueue`, and during flush nothing new enqueues once drained. The identity-snapshot check is the bound. Adding a magic-number cap would mask a real bug rather than surface it. Leaving as-is

strands-ts/src/memory/extraction/coordinator.ts
```diff
@@ -0,0 +1,262 @@
...
+   * Saves the messages to the store, one of two ways:
+   * - store has an extractor: run it to pull out facts, then write each fact via `add`.
+   * - no extractor: hand the raw messages to `addMessages` so the store keeps their roles.
+   *
+   * Fact writes run in parallel. If any fails we throw, which makes the caller retry the whole batch
+   * next time - so a fact that already saved may be written again (stores should expect duplicates).
+   */
+  private async _write(store: MemoryStore, messages: MessageData[]): Promise<void> {
+    const extractor = store.extraction!.extractor
+
```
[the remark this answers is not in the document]
opieter-aws: Will add

strands-ts/src/memory/extraction/model-extractor.ts
```diff
@@ -0,0 +1,140 @@
...
+    ]
+
+    const stream = model.streamAggregated(promptMessages, { systemPrompt: this._systemPrompt })
+    // Manual .next() loop: streamAggregated returns its result as the generator return value
+    // (done:true), which for-await-of discards.
+    let result: Awaited<ReturnType<typeof stream.next>> | undefined
+    for (;;) {
+      result = await stream.next()
+      if (result.done) break
+    }
```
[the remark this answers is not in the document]
opieter-aws: This mirrors the existing pattern in summarizing-conversation-manager (the SDK’s only other streamAggregated consumer), keeping it consistent. No intermediate values are needed; we only want the final aggregated result.

strands-ts/src/memory/types.ts
```diff
@@ -69,20 +77,42 @@ export interface MemoryStore extends MemoryStoreConfig {
...
+   * may return whatever shape fits it. The {@link MemoryManager} does not consume this value (it only
    * awaits completion); callers using a store directly can read it.
    */
   add?(content: string, metadata?: Record<string, JSONValue>): Promise<unknown>
+  /**
+   * Ingest a batch of conversation messages, preserving their role structure. This is the no-extractor
+   * sink: when a store's {@link ExtractionConfig} has no extractor, the manager hands the filtered
+   * {@link MessageData} batch straight here in one call — no serialization, no model call. Backends
+   * that distill server-side (e.g. role-aware conversational APIs) implement this so the
+   * USER/ASSISTANT structure survives. An extraction config without an extractor requires this method.
```
mkmeral: > An extraction config without an extractor requires this method.

extractor is a specific implementation right? this confused me a bit tbh

mkmeral: I'm not sure a customer reading this would understand 😅

strands-ts/src/memory/types.ts
```diff
@@ -69,20 +77,42 @@ export interface MemoryStore extends MemoryStoreConfig {
...
+   * Satisfies `writable: true` the same way {@link add} does. The resolved value is store-specific
+   * and not consumed by the manager.
+   *
+   * Any per-write context (e.g. tenant/session/actor identifiers) is the store's own concern: carry
+   * it on the store instance's config and stamp it here, mirroring how `add` implementations attach
+   * their own metadata. The manager does not supply batch-level metadata.
+   *
+   * @param messages - The filtered messages to ingest, in order
+   */
+  addMessages?(messages: MessageData[]): Promise<unknown>
```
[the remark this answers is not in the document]
mkmeral: why did we remove metadata?

strands-ts/src/memory/extraction/coordinator.ts
```diff
@@ -0,0 +1,262 @@
...
+/** A buffered message and its sequence number. */
+interface BufferedMessage {
+  seq: number
+  message: MessageData
+}
+
+/** The kind of a content block (e.g. 'text', 'toolUse'), used to match it against a filter. */
+function _blockKind(block: ContentBlockData): string {
+  // A text block is `{ text: string }`; every other block is a single-key wrapper (`{ toolUse }`, …).
```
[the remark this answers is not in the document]
opieter-aws: There’s already an inline comment noting the single-key-wrapper assumption; fine to leave

strands-ts/src/memory/extraction/model-extractor.ts
```diff
@@ -0,0 +1,140 @@
...
+  }
+  return entries
+}
+
+/** Extracts the substring from the first `[` to the last `]`, or undefined if absent. */
+function _extractJsonArray(text: string): string | undefined {
+  const start = text.indexOf('[')
+  const end = text.lastIndexOf(']')
+  if (start === -1 || end === -1 || end < start) {
```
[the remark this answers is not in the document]
opieter-aws: Current comment should be sufficient.

strands-ts/src/memory/extraction/model-extractor.ts
mkmeral: qq, and you can re-resolve this. do we expect to bring other native extractors? (that might bring on their own dependencies, which is why we have modules like `vended-plugins`)

strands-ts/src/memory/extraction/types.ts
```diff
@@ -0,0 +1,144 @@
...
+}
+
+/**
+ * Context handed to {@link ExtractionTrigger.attach} so a trigger can wire itself into the agent
+ * lifecycle and signal when extraction should run for its store.
+ */
+export interface ExtractionTriggerContext {
+  /** The agent the trigger attaches its hooks to. */
+  agent: LocalAgent
+  /** Save this store's unsaved messages now. Runs in the background and returns immediately, so calling it from a hook never blocks the agent. To await completion, see {@link MemoryManager.flush}. */
```
mkmeral, line 85: this is for flushing on interval, right? it seems like we actually have a cleanup method in agent in python, we can bring it to TS too, I think. 

so I'd say leave this now. let's merge, and then we followup with a cleanup method on agent that calls `mm.flush()` wdyt?

opieter-aws: Agreed, let's merge as-is, and follow up with an agent-level cleanup/dispose method that calls `mm.flush()`so we have a “flush on shutdown”. Nice that we have this for Python already, I'll look into it!

strands-ts/src/memory/memory-manager.ts
opieter-aws: We already throw in the constructor, but can explicitly filter here to make it clear

strands-ts/src/memory/memory-manager.ts
opieter-aws: Correct, `MessageAddedEvent` doesn’t trigger for direct `agent.messages` mutation or session-preloaded history

strands-ts/src/memory/types.ts
opieter-aws: Yeah I see, this wording is leaning into internal jargon of other components. Will reworded to drop extractor from the explanation and clarify. When extraction is set up without a model, this is the method that's called with the raw messages.

strands-ts/src/memory/types.ts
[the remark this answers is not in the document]
opieter-aws: The `metadata?` param here was dead: the coordinator only ever calls `addMessages(messages)` and never receives anything to put in it. Per-write context (tenant / session / actor IDs) is the store’s own concern: a store carries it on its instance config and stamps it inside its `addMessages` implementation according to its own api / write contract.

strands-ts/src/memory/extraction/model-extractor.ts
opieter-aws: I don't know a use case for this now. But if a user would be brining an extractor setup with its own dependencies, I'd expect it to be vended as an implementation on the store, similar to how managed backends have their own extraction running server side. This can be plugged into the `addMessages` store method of a vended store.

strands-ts/src/memory/memory-manager.ts
mkmeral: should it? I don't think it's a blocker for the work, but we might want to add it as a followup issue

strands-ts/src/memory/types.ts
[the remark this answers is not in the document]
mkmeral: what about namespaces though? Like how do I know which user this is from, when I have namespaced memories and extracted info? 

The general method is through appState (in tools, hooks, etc.), and we don't seem to be passing anything here.

strands-ts/src/memory/extraction/types.ts
mkmeral: but then do we want this param? if we will fix it later?

opieter-aws: `fire` and the later cleanup method don’t overlap: `fire` is the `trigger`’s signal, and `MemoryManager.flush()` is the real primitive to drain outstanding writes and await that process. The follow-up agent cleanup method won’t replace `flush()`, but call it

strands-ts/src/memory/memory-manager.ts
opieter-aws: For customizing `agent.messages` directly I'm not sure I see a clear use case yet.

For preloaded sessions, I agree it would be a nice addition. We'd need to introduce a mechanism that captures the high water mark for writes in the session history so we don't duplicate writes of an old session. It's doable but not straight-forward.

strands-ts/src/memory/types.ts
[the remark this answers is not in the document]
opieter-aws: Discussed offline:
* Store implementations should have metadata like `namespaces` be mutable instance properties
* I will re-add a config field here so we can extend it later without being backwards incompatible
* Stores need that implement this method need to take care of the namespace matching, that's not a concern the base class or interface can take care of at extraction time

strands-ts/src/memory/extraction/types.ts
mkmeral: ah I see, yeah I didn't get what `fire` was at first, but given its mostly internal concept, I think we are fine

mkmeral (approved):

PR #2681, opened 2026-06-08
feat: add middleware system for wrapping agent stages

zastrowm:
Port of strands-agents/sdk-typescript#1068 into the mono-repo.

## Summary

Adds a middleware system that wraps agent stages using async generator handlers. Middleware controls flow (retry, cache, transform, short-circuit).

**Public API:** `agent.addMiddleware(stage, handler)` — returns a cleanup function.

Three built-in stages: `InvokeModelStage`, `ExecuteToolStage`, `AgentStreamStage`. 

## Motivation

Hooks let you observe operations and set flags, but they don't let you wrap them. If you want to do something both before and after a model call (timing it, adding a span, catching errors), hooks force you to manage state across two separate callbacks. With middleware you can wrap the entire invocation keeping state within your callback:

```typescript
agent.addMiddleware(InvokeModelStage, async function* (context, next) {
  const start = Date.now()
  const result = yield* next(context)
  metrics.record(Date.now() - start)
  return result
})
```

Beyond the before/after pattern, middleware also makes several other use cases much more natural to express: caching, input transformation, short-circuiting, and error handling. All of these are awkward or impossible with hooks alone.

## Public API Changes

```typescript
import { Agent, InvokeModelStage, ExecuteToolStage, AgentStreamStage } from '@strands-agents/sdk'
import type { MiddlewareStage, MiddlewareHandler, MiddlewareNext } from '@strands-agents/sdk'

const agent = new Agent({ model, tools })

// Register middleware for any built-in stage
agent.addMiddleware(InvokeModelStage, async function* (context, next) {
  // pre-processing: inspect or transform context
  const modified = { ...context, messages: sanitize(context.messages) }
  // call next layer (or don't, to short-circuit)
  const result = yield* next(modified)
  // post-processing: inspect or transform result
  return result
})
```

Three stages ship with the SDK:

| Stage              | Wraps                                                     | Context fields                                            |
| ------------------ | --------------------------------------------------------- | --------------------------------------------------------- |
| `InvokeModelStage` | Model call (between Before/AfterModelCallEvent)           | messages, systemPrompt, toolSpecs, toolChoice, modelState |
| `ExecuteToolStage` | Single tool execution (between Before/AfterToolCallEvent) | tool, toolUse (name, id, input)                           |
| `AgentStreamStage` | Full `agent.stream()` output                              | args, options                                             |

Handlers are async generators and simple pass-through is `return yield* next(context)`. Manual iteration of `next()` allows real-time event filtering or injection while not calling `next` at all short-circuits the operation.

## Plugin Examples

```typescript
class ToolResultCache implements Plugin {
  name = 'tool-result-cache'
  private readonly _cache = new Map<string, ToolResultBlock>()

    const cache = this._cache
    agent.addMiddleware(ExecuteToolStage, async function* (context, next) {
      const key = `${context.toolUse.name}:${JSON.stringify(context.toolUse.input)}`
      const cached = cache.get(key)
      if (cached) return { result: new ToolResultBlock({ toolUseId: context.toolUse.toolUseId, status: cached.status, content: cached.content }) }
      const result = yield* next(context)
      cache.set(key, result.result)
      return result
    })
  }
}
```

```typescript
class RetryOnThrottle implements Plugin {
  name = 'retry-on-throttle'
    agent.addMiddleware(InvokeModelStage, async function* (context, next) {
      for (let attempt = 0; attempt < 3; attempt++) {
        try { return yield* next(context) }
        catch (e) {
          if (!(e as Error).message.includes('ThrottlingException') || attempt === 2) throw e
        }
      }
      throw new Error('exhausted retries')
    })
  }
}
```

```typescript
class SystemPromptInjector implements Plugin {
  name = 'system-prompt-injector'
  constructor(private readonly _suffix: string) {}

    const suffix = this._suffix
    agent.addMiddleware(InvokeModelStage.Input, async (context) => ({
      ...context,
      systemPrompt: `${context.systemPrompt ?? ''}\n\n${suffix}`.trim(),
    }))
  }
}

// Usage: inject safety guidelines into every model call
const agent = new Agent({
  model,
  systemPrompt: 'You are a helpful assistant.',
  plugins: [new SystemPromptInjector('Always cite your sources.')],
})
```

## Middleware Interrupts

Middleware contexts expose `interrupt()` for human-in-the-loop gating:

```typescript
agent.addMiddleware(ExecuteToolStage, async function* (context, next) {
  const { response } = context.interrupt<string>({ name: 'approve', reason: 'Confirm?' })
  if (response !== 'yes') return { result: new ToolResultBlock({ ... }) }
  return yield* next(context)
})
```

Returns `MiddlewareInterruptResult<T>` (wrapper) — allows non-breaking additions (cached data, metadata) as the interrupt system evolves.

## Phase Sub-Stages (Input / Wrap / Output)

Each stage exposes three phases with a fixed execution order in order to solve the middleware ordering problem without explicit priority numbers:

- `Input`: Modifies the input of the phase as a pure function - take in the original input, return the modified input
- `Output`: Modifies the output of the phase as a pure function - take in the original output, return the modified output
- `Wrap`: Wraps the entire invocation of the stage - this is more akin to traditional middleware from express or other frameworks

```typescript
// Input: transform context before execution (plain async function)
agent.addMiddleware(InvokeModelStage.Input, async (context) => ({
  ...context,
  systemPrompt: injectToSystemPrompt(context),
}))

// Output: transform result after execution (plain async function)
agent.addMiddleware(InvokeModelStage.Output, async (result) => {
  log(`stopReason=${result.result.stopReason}`)
  return result
})

// Wrap: full async generator wrap (same as just providing `InvokeModelStage`)
agent.addMiddleware(InvokeModelStage.Wrap, async function* (context, next) {
  return yield* next(context)
})
```

**Execution order is fixed:** all Input → all Output → all Wrap → terminal. A retry plugin on `.Wrap` retries subsequent Wrap handlers and the terminal — Input/Output transforms are not re-executed. A response logger on `.Output` never needs to coordinate registration order with a system prompt injector on `.Input`.

## Key Decisions

- **First registered = outermost**: follows existing middleware conventions (from Express/Koa)
- **Phase ordering is fixed**: Input/Output/Wrap reduces the need for explicit priority management (for now - we'll probably need to add it later)
- **Hooks fire unconditionally before/after middleware**: existing behavior is unchanged
- **Result wrapper types** (`InvokeModelResult`, etc.): allows future extension of middleware return values, including returning values "up the chain"
- **`readonly` arrays** on `InvokeModelContext`: enforce immutable-context pattern at the type level

## Known Limitations

- **Middleware cannot resume interrupts.** `AgentStreamStage` middleware can observe a tool-level interrupt result (`stopReason: 'interrupt'`) but cannot re-enter the stream with interrupt responses. Interrupt resolution runs in `stream()`'s outer loop, outside middleware. To programmatically resume interrupts within a single invocation, use `AfterInvocationEvent.resume` in a hook. A future enhancement may add a resume mechanism to `AgentStreamResult` or `AgentStreamContext`.

## Checklist
- [x] Tests prove the fix is effective / feature works
- [x] No new warnings
- [ ] Documentation update (pending)

----

### DIFF
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -885,112 +885,97 @@ export class Agent implements LocalAgent, InvokableAgent {
     try {
       await this.initialize()
 
-      // Process interrupt responses before middleware runs so context.interrupt() can find them
-      const interruptResponses = this._extractInterruptResponses(args)
-      if (interruptResponses.length > 0) {
-        this._interruptState.resume(interruptResponses)
-      }
-
       // Thread the resolved invocationState so all layers share the same reference.
       const invocationState = options?.invocationState ?? {}
       const resolvedOptions: InvokeOptions = options?.invocationState ? options : { ...options, invocationState }
 
-      return yield* this._streamWithResumeLoop(args, resolvedOptions, invocationState)
+      let currentArgs: InvokeArgs = args
+
+      while (true) {
+        // Fresh AbortController per iteration, composed with any external signal.
+        this._abortController = new AbortController()
+        this._abortSignal = resolvedOptions?.cancelSignal
+          ? AbortSignal.any([this._abortController.signal, resolvedOptions.cancelSignal])
+          : this._abortController.signal
+
+        // Process interrupt responses before middleware runs so context.interrupt() can find them
+        const interruptResponses = this._extractInterruptResponses(currentArgs)
+        if (interruptResponses.length > 0) {
+          this._interruptState.resume(interruptResponses)
+        }
+
+        // Hooks fire outside middleware — always, even on short-circuit.
+        const beforeInvocationEvent = new BeforeInvocationEvent({ agent: this, invocationState })
+        yield await this._invokeCallbacks(beforeInvocationEvent)
+
+        if (beforeInvocationEvent.cancel) {
+          const cancelText =
+            typeof beforeInvocationEvent.cancel === 'string'
+              ? beforeInvocationEvent.cancel
+              : 'invocation denied by hook'
+          const message = new Message({ role: 'assistant', content: [new TextBlock(cancelText)] })
+          yield this._appendMessage(message, invocationState)
+          const afterEvent = new AfterInvocationEvent({ agent: this, invocationState })
+          await this._invokeCallbacks(afterEvent)
+          yield afterEvent
+          return new AgentResult({
+            stopReason: 'endTurn',
+            lastMessage: message,
+            traces: this._tracer.localTraces,
+            metrics: this._meter.metrics,
+            invocationState,
+          })
+        }
+
+        let result: AgentResult | undefined
+        let caughtError: Error | undefined
+        const afterInvocationEvent = new AfterInvocationEvent({ agent: this, invocationState })
+        try {
+          result = yield* this._streamWithMiddleware(currentArgs, resolvedOptions, invocationState)
+        } catch (error) {
+          caughtError = error as Error
+        } finally {
+          // AfterInvocationEvent always fires — even on error or consumer break. Outside middleware.
+          // Invoke hooks (so .resume can be set) but don't yield in finally (yields in finally
+          // suspend the generator on consumer break instead of completing cleanup).
+          await this._invokeCallbacks(afterInvocationEvent)
+        }
+
+        // Yield outside finally — in JS, a `yield` inside `finally` suspends the generator
+        // mid-cleanup when the consumer breaks, preventing subsequent cleanup code from running.
+        // This line is only reached on normal completion or caught error, never on consumer break.
+        yield afterInvocationEvent
+
+        // Re-throw after hooks have fired
+        if (caughtError) {
+          throw caughtError
+        }
+
+        // Resume only on a clean invocation — errors propagate above.
+        if (afterInvocationEvent.resume !== undefined) {
+          currentArgs = afterInvocationEvent.resume
+          continue
+        }
+
+        // Only emit AgentResultEvent on the final iteration (not on resumed ones).
+        yield await this._invokeCallbacks(
+          new AgentResultEvent({
+            agent: this,
+            result: result!,
+            invocationState,
+          })
+        )
+
+        return result!
+      }
     } finally {
       this._isInvoking = false
     }
   }
 
   /**
-   * Resume loop that fires BeforeInvocationEvent/AfterInvocationEvent per-iteration,
-   * outside middleware. Hooks always observe each iteration; middleware runs inside.
-   */
-  private async *_streamWithResumeLoop(
-    args: InvokeArgs,
-    options: InvokeOptions,
-    invocationState: InvocationState
-  ): AsyncGenerator<AgentStreamEvent, AgentResult, undefined> {
-    let currentArgs: InvokeArgs = args
-
-    while (true) {
-      // Hooks fire outside middleware — always, even on short-circuit.
-      const beforeInvocationEvent = new BeforeInvocationEvent({ agent: this, invocationState })
-      yield await this._invokeCallbacks(beforeInvocationEvent)
-
-      if (beforeInvocationEvent.cancel) {
-        const cancelText =
-          typeof beforeInvocationEvent.cancel === 'string' ? beforeInvocationEvent.cancel : 'invocation denied by hook'
-        const message = new Message({ role: 'assistant', content: [new TextBlock(cancelText)] })
-        yield this._appendMessage(message, invocationState)
-        const afterEvent = (await this._invokeCallbacks(
-          new AfterInvocationEvent({ agent: this, invocationState })
-        )) as AfterInvocationEvent
-        yield afterEvent
-        return new AgentResult({
-          stopReason: 'endTurn',
-          lastMessage: message,
-          traces: this._tracer.localTraces,
-          metrics: this._meter.metrics,
-          invocationState,
-        })
-      }
-
-      // Fresh AbortController per iteration, composed with any external signal.
-      this._abortController = new AbortController()
-      this._abortSignal = options?.cancelSignal
-        ? AbortSignal.any([this._abortController.signal, options.cancelSignal])
-        : this._abortController.signal
-
-      let result: AgentResult | undefined
-      let caughtError: Error | undefined
-      let afterInvocationEvent: AfterInvocationEvent | undefined
-      try {
-        result = yield* this._streamWithMiddleware(currentArgs, options, invocationState)
-      } catch (error) {
-        caughtError = error as Error
-      } finally {
-        // AfterInvocationEvent always fires — even on error or consumer break. Outside middleware.
-        // Invoke hooks (so .resume can be set) but don't yield in finally (yields in finally
-        // suspend the generator on consumer break instead of completing cleanup).
-        afterInvocationEvent = (await this._invokeCallbacks(
-          new AfterInvocationEvent({ agent: this, invocationState })
-        )) as AfterInvocationEvent
-      }
-
-      // Yield the event outside finally (only reached on normal completion or caught error, not break)
-      yield afterInvocationEvent
-
-      // Re-throw after hooks have fired
-      if (caughtError) {
-        throw caughtError
-      }
-
-      // Resume only on a clean invocation — errors propagate above.
-      if (afterInvocationEvent!.resume !== undefined) {
-        currentArgs = afterInvocationEvent!.resume
-        const resumeInterruptResponses = this._extractInterruptResponses(currentArgs)
-        if (resumeInterruptResponses.length > 0) {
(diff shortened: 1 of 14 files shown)


strands-ts/src/agent/agent.ts
```diff
@@ -524,6 +513,107 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+          return () => this._middlewareRegistry.remove(stage, adapted)
+        }
+        case 'output': {
+          const adapted = this._middlewareRegistry.addOutput(
+            stageOrPhase as MiddlewareOutputPhase<TContext, TResult, TEvent>,
+            handler as MiddlewareOutputHandler<TResult>
+          )
+          return () => this._middlewareRegistry.remove(stage, adapted)
+        }
+        case 'around': {
```
opieter-aws: I prefer `wrap` here naming-wise.

From a software perspective it's a more familiar term and self-explanatory

strands-ts/src/agent/agent.ts
```diff
@@ -795,92 +885,162 @@ export class Agent implements LocalAgent, InvokableAgent {
...
-            }
+      // async function* doesn't bind lexical `this`; capture for the terminal callback.
+      // eslint-disable-next-line @typescript-eslint/no-this-alias
+      const self = this
+      try {
+        const { result } = yield* this._middlewareRegistry.invoke(
+          AgentStreamStage,
+          context,
+          async function* (ctx: AgentStreamContext): AsyncGenerator<AgentStreamEvent, AgentStreamResult, undefined> {
+            const result = yield* self._streamWithResumeLoop(ctx.args, ctx.options)
```
opieter-aws: Does the `AgentStreamStage` skip the `BeforeInvocationEvent` and `AfterInvocationEvent` on purpose if it short-circuits?

strands-ts/src/agent/agent.ts
```diff
@@ -795,92 +885,162 @@ export class Agent implements LocalAgent, InvokableAgent {
...
-              )
-            }
-            drainResult = await streamGenerator.next()
+          this._interruptState.activate()
+          for (const interrupt of error.interrupts) {
+            yield new InterruptEvent({ agent: this, interrupt, invocationState })
           }
+          return new AgentResult({
+            stopReason: 'interrupt',
+            lastMessage: new Message({ role: 'assistant', content: [] }),
```
opieter-aws: Even if we don't reuse the function, do we want the behaviour to be symmetric? If so, we should set `content` here to last user message / `[new TextBlock('Interrupted')]`

strands-ts/src/agent/agent.ts
```diff
@@ -524,6 +513,107 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+   */
+  addMiddleware<TContext, TResult, TEvent>(
+    stage: MiddlewareStage<TContext, TResult, TEvent>,
+    handler: MiddlewareHandler<TContext, TResult, TEvent>
+  ): () => void
+  addMiddleware<TContext, TResult, TEvent>(
+    stageOrPhase:
+      | MiddlewareStage<TContext, TResult, TEvent>
+      | MiddlewareInputPhase<TContext, TResult, TEvent>
+      | MiddlewareOutputPhase<TContext, TResult, TEvent>,
```
opieter-aws, line 589: Should this include `MiddlewareAroundPhase`?

strands-ts/src/agent/agent.ts
zastrowm: Will switch to wrap.  I was tied between Around and Wrap, but this weighs more towards Wrap

strands-ts/src/agent/agent.ts
zastrowm: No - definitely not on purpose - will address

strands-ts/src/agent/agent.ts
zastrowm: yeah, we should match normal interrupts; will follow the lead there

strands-ts/src/agent/agent.ts
zastrowm: For completeness yeah.

strands-ts/src/middleware/registry.ts
```diff
@@ -0,0 +1,151 @@
...
+   * @returns A single function representing the full middleware chain
+   */
+  compose<TContext, TResult, TEvent>(
+    stage: MiddlewareStage<TContext, TResult, TEvent>,
+    terminal: MiddlewareNext<TContext, TResult, TEvent>
+  ): MiddlewareNext<TContext, TResult, TEvent> {
+    const tagged = (this._handlers.get(stage) ?? []) as TaggedHandler<TContext, TResult, TEvent>[]
+
+    // Stable sort by phase: input first, output second, around last (closest to terminal)
+    const sorted = [...tagged].sort((a, b) => PHASE_ORDER[a.phase] - PHASE_ORDER[b.phase])
```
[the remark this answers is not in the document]
zastrowm: This is fine for now

strands-ts/src/index.ts
```diff
@@ -307,6 +307,24 @@ export { Sandbox, type ExecuteOptions } from './sandbox/base.js'
 export { PosixShellSandbox } from './sandbox/posix-shell.js'
 export type { StreamType, StreamChunk, FileInfo, OutputFile, ExecutionResult } from './sandbox/types.js'
 
+// Middleware system
+export { createStage, InvokeModelStage, ExecuteToolStage, AgentStreamStage } from './middleware/index.js'
+export type {
```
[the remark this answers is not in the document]
zastrowm: Exporting

strands-ts/src/types/agent.ts
```diff
@@ -341,6 +349,27 @@ export interface LocalAgent {
...
+   */
+  addMiddleware<TContext, TResult, TEvent>(
+    phase: MiddlewareInputPhase<TContext, TResult, TEvent>,
+    handler: MiddlewareInputHandler<TContext>
+  ): () => void
+  addMiddleware<TContext, TResult, TEvent>(
+    phase: MiddlewareOutputPhase<TContext, TResult, TEvent>,
+    handler: MiddlewareOutputHandler<TResult>
+  ): () => void
+  addMiddleware<TContext, TResult, TEvent>(
```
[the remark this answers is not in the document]
zastrowm: Fixed

strands-ts/src/index.ts
```diff
@@ -307,6 +307,26 @@ export { Sandbox, type ExecuteOptions } from './sandbox/base.js'
 export { PosixShellSandbox } from './sandbox/posix-shell.js'
 export type { StreamType, StreamChunk, FileInfo, OutputFile, ExecutionResult } from './sandbox/types.js'
 
+// Middleware system
+export { InvokeModelStage, ExecuteToolStage, AgentStreamStage } from './middleware/index.js'
```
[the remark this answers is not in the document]
opieter-aws: +1, was this intentional?

strands-ts/src/agent/agent.ts
```diff
@@ -296,6 +262,36 @@ const DEFAULT_AGENT_ID = 'agent'
...
+  return <T = JSONValue>(params: InterruptParams): MiddlewareInterruptResult<T> => {
+    const interruptId = `${idPrefix}:${params.name}`
+    const existing = interruptState.interrupts[interruptId]
+    if (existing?.response !== undefined) {
+      return { response: existing.response as T }
+    }
+    if (params.response !== undefined) {
+      return { response: params.response as T }
+    }
+    const interrupt = new Interrupt({
```
opieter-aws, line 286: Should we set `source` here explicitly too? 

Currently the interrupt defaults to `hook` when source is not set. The `id` does reflect middleware, but we could consider also adding this as `source` type

 https://github.com/strands-agents/harness-sdk/blob/60c52ad01578589ef663c7308f44dc63c6156828/strands-ts/src/interrupt.ts#L67

strands-ts/src/agent/agent.ts
opieter-aws: Is the span here deliberately observing the original input, rather than the transformed input?

strands-ts/src/index.ts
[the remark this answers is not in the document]
zastrowm: Yes; for the first release I'm cutting createStage from being an api that we expose; I think as-is is it raises more questions than it delivers functionality

strands-ts/src/agent/agent.ts
zastrowm: Yes, good catch.  Interrupts are my weak point

strands-ts/src/agent/agent.ts
zastrowm: No, fixing.  It looks like tools and the agent span don't have this issue, but this should definitely be recording the transformed input

opieter-aws (approved):

pgrayy (approved):

PR #2693, opened 2026-06-09
feat: implement intervention primitive in python with cancellation support

### LINKED ISSUE #2667: Interventions Primitive - Python Parity
Implement the Interventions primitive in `strands-py` to achieve parity between languages (`strands-ts` and `strands-py`).

Related issues:
- Typescript implementation: https://github.com/strands-agents/sdk-typescript/issues/846
- Design doc: https://github.com/strands-agents/docs/blob/main/designs/0007-intervention-primitive.md

mehtarac:
## Description

Adds the intervention primitive — a composable control layer for agents that enables authorization, guardrails, steering, and operational controls to share a common interface with ordered evaluation, short-circuiting, and typed actions.

This implements Python parity with the TypeScript SDK's intervention primitive (strands-agents/sdk-typescript#883).

Resolves #2667

### `Agent(interventions=...)`

```python
Agent(interventions: list[InterventionHandler] | None = None)
```

Handlers are evaluated in registration order at each lifecycle event. Cheapest handlers (authorization, guardrails) should be listed first; expensive ones (LLM steering) last.

### Public Exports

```python
# Top-level
from strands import InterventionHandler

# Action types and handler interface (for type annotations and construction)
from strands.interventions import (
    InterventionHandler,
    InterventionAction,
    OnError,
    Proceed,
    Deny,
    Guide,
    Confirm,
    Transform,
)
```

### `InterventionHandler`

```python
class InterventionHandler(ABC):
    name: str  # abstract property — unique identifier
    on_error: OnError = "throw"  # error policy

    # Lifecycle methods — override the ones you care about (sync or async)
    def before_invocation(self, event: BeforeInvocationEvent, **kwargs) -> Proceed | Deny | Guide | Transform
    def before_tool_call(self, event: BeforeToolCallEvent, **kwargs) -> Proceed | Deny | Guide | Confirm | Transform
    def after_tool_call(self, event: AfterToolCallEvent, **kwargs) -> Proceed | Transform
    def before_model_call(self, event: BeforeModelCallEvent, **kwargs) -> Proceed | Deny | Guide | Transform
    def after_model_call(self, event: AfterModelCallEvent, **kwargs) -> Proceed | Guide | Transform
```

All lifecycle methods default to `Proceed()`. Override only the ones you need — the framework detects overrides at the **class level** and only registers hooks for those. Handlers can be sync (`def`) or async (`async def`).

### Action Types

Actions are frozen dataclasses constructed directly:

```python
Proceed(reason: str | None = None)
Deny(reason: str = "")
Guide(feedback: str = "", reason: str | None = None)
Confirm(prompt: str = "", reason: str | None = None, response: Any = None, evaluate: Callable = default_evaluate)
Transform(apply: Callable[[LifecycleEvent], None] = ..., reason: str | None = None)
```

| Action    | Description                                      |
|-----------|--------------------------------------------------|
| Proceed   | Allow the operation to continue                  |
| Deny      | Block the operation (sets event.cancel)          |
| Guide     | Steer with feedback                              |
| Confirm   | Pause for human approval (before_tool_call only) |
| Transform | Modify event content in-place                    |

### Action-to-Event Compatibility Matrix

| Action    | before_invocation | before_tool_call | before_model_call | after_tool_call | after_model_call |
|-----------|-------------------|------------------|-------------------|-----------------|------------------|
| Proceed   | —                 | —                | —                 | —               | —                |
| Deny      | cancel            | cancel           | cancel            | —               | —                |
| Guide     | cancel+           | cancel+          | inject            | —               | inject + retry   |
| Confirm   | —                 | confirm          | —                 | —               | —                |
| Transform | apply             | apply            | apply             | apply           | apply            |

- **—** = no-op (warns at runtime)
- **cancel** = sets event.cancel/cancel_tool, short-circuits (remaining handlers skipped)
- **cancel+** = sets cancel with accumulated feedback from all guiding handlers
- **confirm** = uses preemptive response or interrupt, checks with evaluate, sets cancel if denied
- **inject** = appends accumulated feedback as a user message so the model sees it on this call
- **inject + retry** = appends accumulated feedback and retries so the model sees guidance
- **apply** = calls action.apply(event) for in-place mutation, later handlers see the change

### `OnError` Policy

| Value       | Behavior                             |
|-------------|--------------------------------------|
| `"throw"`   | Rethrow (default). Invocation fails. |
| `"proceed"` | Skip handler, continue to next (fail-open). |
| `"deny"`    | Apply Deny (fail-closed).            |

### Hook Ordering

- Before* events: interventions run at `INTERVENTION_INPUT` (90) — after plugins (0)
- After* events: interventions run at `INTERVENTION_OUTPUT` (-90) — before plugins (0)

Flow: `plugins → intervention → [operation] → intervention → plugins`

### What's NOT exported (internal)

- **InterventionRegistry** — internal dispatch mechanism
- **Audit log** — not included. Will be added when consumption patterns are clear.

---

## Infrastructure Changes (cancellation support)

This PR also adds general-purpose **cancellation support** to two hook events. These fields are usable by any hook or plugin, not just interventions — but are required for the intervention primitive's Deny action to work.

### `BeforeInvocationEvent.cancel`

```python
cancel: bool | str = False
```

When set by a hook callback, the agent loop:
1. Creates an assistant message with the cancel text
2. Fires `MessageAddedEvent`
3. Yields `EventLoopStopEvent` with `stop_reason="end_turn"`
4. Fires `AfterInvocationEvent`
5. Returns without entering the event loop

### `BeforeModelCallEvent.cancel`

```python
cancel: bool | str = False
```

When set by a hook callback, the event loop:
1. Creates a synthetic assistant message with the cancel text
2. Fires `AfterModelCallEvent` (allows retry via `event.retry = True`)
3. Ends the model invoke span
4. Yields `ModelStopReason` event
5. Breaks out of the model retry loop (or continues if retry requested)

### `HookOrder` constants

```python
HookOrder.INTERVENTION_INPUT = 90    # After plugins on Before* events
HookOrder.INTERVENTION_OUTPUT = -90  # Before plugins on After* events
```

---

## Example Usage

```python
from strands import Agent, InterventionHandler, tool
from strands.interventions import Deny, Guide, Proceed


@tool
def send_email(to: str, body: str) -> str:
    """Send an email to a recipient."""
    return f"Email sent to {to}"


@tool
def query_database(query: str) -> str:
    """Run a database query."""
    return f"Results for: {query}"


ALLOWED_TOOLS = {
    "analyst": ["query_database"],
    "admin": ["query_database", "send_email"],
}


class RoleAuth(InterventionHandler):
    name = "role-auth"

    def before_tool_call(self, event):
        role = event.invocation_state.get("role", "")
        tool_name = event.tool_use["name"]
        if tool_name not in ALLOWED_TOOLS.get(role, []):
            return Deny(reason=f"Role '{role}' is not authorized for tool '{tool_name}'")
        return Proceed()


class PoliteGuard(InterventionHandler):
    name = "polite-guard"

    def after_model_call(self, event):
        if event.stop_response and event.stop_response.message:
            text = "".join(
                block.get("text", "")
                for block in event.stop_response.message.get("content", [])
                if "text" in block
            )
            if any(word in text.lower() for word in ["stupid", "idiot", "dumb"]):
                return Guide(feedback="Rephrase your response to be professional and respectful.")
        return Proceed()


agent = Agent(
    tools=[query_database, send_email],
    interventions=[RoleAuth(), PoliteGuard()],  # cheapest first
)

# Analyst can query but not send email
result = agent("Send an email to bob@example.com saying hello", invocation_state={"role": "analyst"})

# Admin can do both
result = agent("Query the database for recent orders", invocation_state={"role": "admin"})
```

## Related Issues

Resolves #2667

## Documentation PR

N/A

## Type of Change

New feature

## Testing

- 74 unit tests covering all action types, short-circuiting, guide accumulation, dispatch ordering, override detection, all three onError modes, sync and async handlers, duplicate name rejection, conflict resolution, native interrupt propagation, confirm (pause/approve/deny/preemptive/custom evaluate/falsy response), transform on all event types, unsupported action warnings, cancel-path integration (deny→end_turn, retry re-entry, plain hook cancel), and full hook integration

## Checklist

----


strands-py/src/strands/interventions/registry.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def _handle_error(self, handler: InterventionHandler, method: str, error: Exception) -> InterventionAction | None:
+        error_msg = str(error)
+
+        if handler.on_error == "throw":
+            raise error
+        elif handler.on_error == "deny":
+            return Deny(reason=f"Handler threw: {error_msg}")
+        else:
+            return None
```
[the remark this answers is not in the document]
mehtarac: Fixed — added `logger.warning` in both the `"proceed"` and `"deny"` branches of `_handle_error`. The log now includes handler name, event type, and error message in the project's structured logging format.

strands-py/src/strands/interventions/handler.py
```diff
@@ -0,0 +1,77 @@
...
+    BeforeModelCallEvent,
+    BeforeToolCallEvent,
+)
+from .actions import Confirm, Deny, Guide, Proceed, Transform
+
+OnError = Literal["throw", "proceed", "deny"]
+"""What to do when a handler throws during evaluation.
+
+- ``'throw'`` — rethrow the error (default, safest: a broken policy check blocks execution)
+- ``'proceed'`` — log the error and continue as if the handler returned Proceed
```
lizradway: max noted that we should comment that this is fail open & use at your own risk

strands-py/src/strands/interventions/registry.py
```diff
@@ -0,0 +1,242 @@
...
+    async def _on_after_tool_call(self, event: AfterToolCallEvent) -> None:
+        await self._dispatch(event, "after_tool_call", self._apply_after_tool_call)
+
+    async def _on_before_model_call(self, event: BeforeModelCallEvent) -> None:
+        await self._dispatch(event, "before_model_call", self._apply_before_model_call)
+
+    async def _on_after_model_call(self, event: AfterModelCallEvent) -> None:
+        await self._dispatch(event, "after_model_call", self._apply_after_model_call)
+
+    def _apply_before_invocation(self, event: LifecycleEvent, action: InterventionAction, handler_name: str) -> bool:
```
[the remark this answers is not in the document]
lizradway: +1

strands-py/src/strands/interventions/registry.py
```diff
@@ -0,0 +1,242 @@
...
+            return False
+        logger.warning("handler=<%s>, event=<after_tool_call> | %s has no effect", handler_name, action.type)
+        return False
+
+    def _apply_before_model_call(self, event: LifecycleEvent, action: InterventionAction, handler_name: str) -> bool:
+        if isinstance(action, Deny):
+            event.cancel = f"DENIED: {action.reason}"
+            return True
+        elif isinstance(action, Guide):
+            event.agent.messages.append({"role": "user", "content": [{"text": action.feedback}]})
```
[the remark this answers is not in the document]
lizradway: @opieter-aws what's the status on the inject messages common function?

strands-py/src/strands/__init__.py
```diff
@@ -4,6 +4,7 @@
 from .agent.agent import Agent
 from .agent.base import AgentBase
 from .event_loop._retry import ModelRetryStrategy
+from .interventions import InterventionHandler, confirm, deny, guide, proceed, transform
```
lizradway: can we name space the interventions actions? in the export importing plain `deny` for instance is very confusing

strands-py/src/strands/interventions/registry.py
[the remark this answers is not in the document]
opieter-aws: It's drafted here: https://github.com/strands-agents/harness-sdk/pull/2631, blocked by middleware.

I don't think it covers the interventions use case, but please confirm.

strands-py/src/strands/interventions/registry.py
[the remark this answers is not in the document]
mehtarac: Resolved — mypy strict mode passes cleanly (0 errors). Most attributes (`cancel`, `cancel_tool`, `retry`, `agent`) exist on the base `HookEvent` class so no narrowing is needed. The only attribute that required it is `event.interrupt()` (from the `_Interruptible` mixin, only on `BeforeToolCallEvent`), handled with a single targeted `# type: ignore[union-attr]`.

Using specific event types per `_apply_*` method would break the uniform `Callable[[LifecycleEvent, InterventionAction, str], bool]` signature used in `_dispatch`, requiring either `Any` (defeating the purpose) or overloads (over-engineering for a private internal method).

strands-py/src/strands/interventions/handler.py
mehtarac: Added a fail-open warning to the `InterventionHandler` class docstring:

> This primitive is **fail-open by default**. All lifecycle methods return `Proceed()` unless explicitly overridden. An agent with no intervention handlers — or handlers that don't override a given method — imposes no restrictions at that lifecycle point. Use at your own risk: if your security model requires fail-closed behavior, ensure every relevant lifecycle method is overridden with an explicit policy decision.

strands-py/src/strands/__init__.py
mehtarac: Fixed — removed bare `deny`, `proceed`, `guide`, `confirm`, `transform` from the top-level `strands` exports. Added `InterventionActions` namespace (matching the TS pattern `export const InterventionActions = { proceed, deny, guide, confirm, transform }`).

Usage is now:
```python
from strands import InterventionActions

InterventionActions.deny("not authorized")
InterventionActions.proceed()
InterventionActions.guide("try again")
```

The bare functions remain accessible via `from strands.interventions import deny` for advanced users who prefer shorter syntax.

strands-py/src/strands/interventions/registry.py
```diff
@@ -0,0 +1,242 @@
...
+            action.apply(event)
+            return False
+        elif isinstance(action, Proceed):
+            return False
+        logger.warning("handler=<%s>, event=<before_model_call> | %s has no effect", handler_name, action.type)
+        return False
+
+    def _apply_after_model_call(self, event: LifecycleEvent, action: InterventionAction, handler_name: str) -> bool:
+        if isinstance(action, Guide):
+            event.retry = True
```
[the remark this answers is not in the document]
mehtarac: Acknowledged — this is a valid observation. The design doc explicitly states "Handlers are responsible for tracking their own retry count and escalating to Deny or Interrupt after repeated failures." The TS implementation has the same behavior. The unbounded retry is by design — convergence is the handler author's responsibility.

Will add a note to the `Guide` docstring clarifying this contract. A framework-level retry cap could be added as a follow-up if demand warrants it.

strands-py/src/strands/agent/agent.py
lizradway, line 1: could you update the pr description for the updated api?

lizradway:
i think there is a lot in here that is not contained under the interventions primitive: ie. cancellation support is a biggie.

i think it might be beneficial to separate that out into a separate quick pr or at the very least update the pr description and commit message to note this so that it is not completely skipped over.

strands-py/src/strands/agent/agent.py
lizradway: additionally can you update the pr description to contain the entire api surface, not just a single example.

strands-py/src/strands/interventions/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+            if not self.is_authorized(event):
+                return InterventionActions.deny("not authorized")
+            return InterventionActions.proceed()
+
+    agent = Agent(interventions=[MyAuth()])
+    ```
+"""
+
+from .actions import (
+    Confirm,
```
lizradway: `interventions/__init__.py` should not define `__all__`. the public API is exported from the top-level strands package. This module only needs to exist so `strands/__init__.py` can import `InterventionActions` and `InterventionHandler` from it.

strands-py/src/strands/interventions/actions.py
```diff
@@ -0,0 +1,157 @@
...
+    """Block the operation."""
+    return Deny(reason=reason)
+
+
+def guide(feedback: str, *, reason: str | None = None) -> Guide:
+    """Provide feedback to steer behavior."""
+    return Guide(feedback=feedback, reason=reason)
+
+
+def confirm(
```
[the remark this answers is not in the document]
lizradway: ++ pls jsdoc for all of this since it can be a little confusing without explicit documentation

strands-py/src/strands/interventions/handler.py
```diff
@@ -0,0 +1,85 @@
...
+- ``'deny'`` — log the error and treat it as a Deny (fail-closed)
+"""
+
+
+class InterventionHandler(ABC):
+    """Base class for intervention handlers.
+
+    .. warning::
+        This primitive is **fail-open by default**. All lifecycle methods return
```
lizradway: i think this "This primitive is **fail-open by default**." is overstated imo. Only the `onError=proceed` is fail open by default.

strands-py/src/strands/interventions/registry.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def _handle_error(self, handler: InterventionHandler, method: str, error: Exception) -> InterventionAction | None:
+        error_msg = str(error)
+
+        if handler.on_error == "throw":
+            raise error
+        elif handler.on_error == "deny":
+            logger.warning("handler=<%s>, event=<%s>, on_error=<deny> | %s", handler.name, method, error_msg)
+            return Deny(reason=f"Handler threw: {error_msg}")
+        else:
```
lizradway: can we have this else statement fall back to `throw` for safety?


this (and maybe where we define `on_error`) is where we should define the fail open warning imo

strands-py/src/strands/interventions/actions.py
```diff
@@ -0,0 +1,157 @@
...
+    apply: Callable[[LifecycleEvent], None] = field(default=lambda e: None)
+    reason: str | None = None
+
+
+InterventionAction = Proceed | Deny | Guide | Confirm | Transform
+"""Union of all intervention actions a handler can return.
+
+Action-to-event compatibility matrix::
+
+    | Action    | before_invocation | before_tool_call | before_model_call | after_tool_call | after_model_call |
```
lizradway, line 122: all handlers are forced async in this -- why?

strands-py/src/strands/interventions/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+    deny,
+    guide,
+    proceed,
+    transform,
+)
+from .handler import InterventionHandler, OnError
+from .registry import InterventionRegistry
+
+
+class InterventionActions:
```
lizradway: why a class? wondering if we should add `__new__` or make this a namespace?

strands-py/src/strands/interventions/actions.py
mehtarac: The Python implementation forced the async def because Python's await raises TypeError. The simplest way to guarantee the registry's await `getattr(handler, method)` always works is to make the base methods async def. 

I can modify this to make base methods plain def and use `inspect.iscoroutinefunction` in the dispatch (the same pattern the hook registry already uses)

strands-py/src/strands/interventions/__init__.py
mehtarac: we used a class because a class with staticmethod can give a good devex. ex: autocomplete, type hints on each method's signature, hover documentation etc. 

I'll add the `__new__` to prevent accidental instantiation and to ensure it's a pure namespace behavior.

strands-py/src/strands/interventions/actions.py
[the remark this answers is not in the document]
mehtarac: Done!

strands-py/src/strands/agent/agent.py
mehtarac: Updated!

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -494,13 +495,40 @@ async def _handle_model_execution(
...
+                        before_model_call_event.cancel
+                        if isinstance(before_model_call_event.cancel, str)
+                        else "model call denied by hook"
+                    )
+                    message: Message = {"role": "assistant", "content": [{"text": cancel_text}]}
+                    stop_reason: StopReason = "end_turn"
+                    usage: Usage = {"inputTokens": 0, "outputTokens": 0, "totalTokens": 0}
+                    metrics: Metrics = {"latencyMs": 0}
+
+                    after_model_call_event = AfterModelCallEvent(
```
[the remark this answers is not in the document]
mehtarac: Acknowledged — same behavior as the TS implementation. Adding a `cancelled` field to `AfterModelCallEvent` would be a new API surface that should go through review separately. Tracked as a follow-up.

strands-py/src/strands/interventions/actions.py
```diff
@@ -0,0 +1,157 @@
...
+
+    Two modes depending on whether response is provided:
+    - With response: passed as a preemptive value to the interrupt system, agent never pauses.
+    - Without response: breaks out of the agent loop to pause for external resume.
+    """
+
+    type: str = field(default="confirm", init=False)
+    prompt: str = ""
+    reason: str | None = None
+    response: Any = None
```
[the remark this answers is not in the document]
mehtarac: Correct — `None` is the sentinel for "no response provided." This is now documented in the `confirm()` factory's Args docstring. The TS uses `undefined` identically.

strands-py/src/strands/interventions/actions.py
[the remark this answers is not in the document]
mehtarac: Noted. Not a practical concern — action objects are ephemeral return values, never used as dict keys or set members. The frozen dataclass gives us immutability guarantees on the fields that matter (`type`, `reason`, `feedback`).

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -494,13 +495,38 @@ async def _handle_model_execution(
...
-                await agent.hooks.invoke_callbacks_async(
-                    BeforeModelCallEvent(
+                before_model_call_event = BeforeModelCallEvent(
+                    agent=agent,
+                    invocation_state=invocation_state,
+                    projected_input_tokens=projected_input_tokens,
+                )
+                await agent.hooks.invoke_callbacks_async(before_model_call_event)
+
+                if before_model_call_event.cancel:
```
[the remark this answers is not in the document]
mehtarac: Valid — the cancel paths in event_loop.py and agent.py lack dedicated unit tests. Will add as a follow-up.

strands-py/src/strands/interventions/registry.py
```diff
@@ -0,0 +1,244 @@
...
+            return False
+        logger.warning("handler=<%s>, event=<after_tool_call> | %s has no effect", handler_name, action.type)
+        return False
+
+    def _apply_before_model_call(self, event: LifecycleEvent, action: InterventionAction, handler_name: str) -> bool:
+        if isinstance(action, Deny):
+            event.cancel = f"DENIED: {action.reason}"
+            return True
+        elif isinstance(action, Guide):
+            event.agent.messages.append({"role": "user", "content": [{"text": action.feedback}]})
```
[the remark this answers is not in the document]
mehtarac: Same as the direct `agent.messages` mutation comment — intentional TS parity. Known limitation documented in Guide docstring.

strands-py/src/strands/interventions/__init__.py
mehtarac: Done — removed `__all__` from `interventions/__init__.py`. Module now only imports what `strands/__init__.py` needs.

strands-py/src/strands/interventions/handler.py
mehtarac: Agreed — removed the class-level warning. Fail-open note now lives on the `OnError` docstring specifically on `"proceed"`.

strands-py/src/strands/interventions/registry.py
mehtarac: Done — `else` now raises the error (falls back to `throw` behavior). `"proceed"` is an explicit `elif` branch with a fail-open warning in the log message.

strands-py/src/strands/interventions/actions.py
mehtarac: Fixed — base methods are now plain `def` (not `async def`), matching the TS pattern (`Awaitable<T>`). The dispatch uses `inspect.iscoroutinefunction` to handle both sync and async overrides. Sync handlers (Cedar, regex) avoid unnecessary coroutine overhead; async handlers (LLM steering, API calls) still work.

strands-py/src/strands/interventions/__init__.py
mehtarac: It's a class because that gives the best IDE experience (autocomplete, type hints on each method's signature, hover docs) — which `SimpleNamespace` and dicts don't provide. Added `__new__` that raises `TypeError` to prevent instantiation.

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -494,13 +495,40 @@ async def _handle_model_execution(
...
+                        ),
                     )
-                )
+                    await agent.hooks.invoke_callbacks_async(after_model_call_event)
+
+                    if after_model_call_event.retry:
+                        tracer.end_model_invoke_span(model_invoke_span, message, usage, metrics, stop_reason)
+                        continue
+
+                    tracer.end_model_invoke_span(model_invoke_span, message, usage, metrics, stop_reason)
```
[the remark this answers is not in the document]
mehtarac: Fixed — consolidated the duplicate call.

strands-py/src/strands/interventions/actions.py
```diff
@@ -0,0 +1,157 @@
...
+@dataclass(frozen=True)
+class Deny:
+    """Block the operation. The reason is shown to the model as the cancellation message."""
+
+    type: str = field(default="deny", init=False)
+    reason: str = ""
+
+
+@dataclass(frozen=True)
+class Guide:
```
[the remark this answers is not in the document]
mehtarac: Fixed — added convergence warning to the Guide docstring.

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -494,13 +495,38 @@ async def _handle_model_execution(
...
-                        projected_input_tokens=projected_input_tokens,
+                        stop_response=AfterModelCallEvent.ModelStopResponse(
+                            stop_reason=stop_reason,
+                            message=message,
+                        ),
                     )
-                )
+                    await agent.hooks.invoke_callbacks_async(after_model_call_event)
+
+                    tracer.end_model_invoke_span(model_invoke_span, message, usage, metrics, stop_reason)
```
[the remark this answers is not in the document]
mehtarac: Fixed — consolidated.

agent-of-mkmeral:
## Full Review @ `861b2fd` — incl. adversarial testing & TS comparison

I checked out the PR, ran the full verification suite, wrote 18 adversarial scenarios through the **real agent loop** (not just registry-level mocks), and diffed the behavior line-by-line against the TS implementation (`strands-ts/src/interventions/*`). The registry/handler/actions layer is solid and faithfully mirrors TS. However, I found **one P0 bug in the cancellation infrastructure** that breaks a headline feature end-to-end.

### 🔴 P0: `BeforeModelCallEvent.cancel` crashes the event loop — `Deny` on `before_model_call` is broken end-to-end

The cancel branch in `_handle_model_execution` (event_loop.py ~505–528) `break`s out of the retry loop **without yielding a `ModelStopReason` event**. The consumer in `event_loop_cycle` (~289–296) does:

```python
async for model_event in model_events:
    if not isinstance(model_event, ModelStopReason):
        yield model_event

stop_reason, message, *_ = model_event["stop"]   # 💥 UnboundLocalError
```

When cancel fires on the first model call, the generator yields *nothing*, so `model_event` is never bound. Repro (real loop, mocked model):

```python
class DenyModel(InterventionHandler):
    name = "deny-model"
    def before_model_call(self, event):
        return InterventionActions.deny("blocked by policy")

agent = Agent(model=mocked, interventions=[DenyModel()])
agent("hi")
# EventLoopException ← UnboundLocalError: cannot access local variable 'model_event'
```

Same crash with a plain hook setting `event.cancel = True` — so the **general-purpose infrastructure feature this PR advertises is unusable**, not just the intervention path. Note the asymmetry: `before_invocation` cancel works fine (its consumer handles `EventLoopStopEvent` directly); only `before_model_call` is broken. In TS this can't happen because the cancel path **`return`s `{ message, stopReason: 'endTurn' }`** from the generator (agent.ts ~1454), which `yield*` hands to the caller. Python's equivalent return channel is the `ModelStopReason` event — and the cancel path forgot it.

**One-line fix** (verified: all 67 intervention tests + 129 agent tests still pass, and all my adversarial scenarios go green):

```python
                    tracer.end_model_invoke_span(model_invoke_span, message, usage, metrics, stop_reason)
                    if after_model_call_event.retry:
                        continue
                    yield ModelStopReason(stop_reason=stop_reason, message=message, usage=usage, metrics=metrics)
                    break
```

(`ModelStopReason` is already imported in event_loop.py.) After the fix the cancel flows correctly: `stop_reason=end_turn`, `message='DENIED: blocked by policy'`, message appended + `MessageAddedEvent` fired via the normal post-loop path.

This is exactly the risk flagged in the earlier review thread about the untested cancel branches — that test gap was deferred as "follow-up," and the bug shipped through it. **Please add the event-loop cancel-path tests in this PR, not a follow-up**: deny → end_turn message; cancel+retry → loop re-entry; plain-hook cancel=True → default text. (TS has these: `agent.hook.test.ts` :1087, :1146 — Python has zero tests asserting `'model call denied by hook'` / `'invocation denied by hook'`.)

### 🟡 Override detection diverges from TS — instance-assigned methods are silently ignored

TS checks `handler[method] !== InterventionHandler.prototype[method]`, which **does** see instance-level assignment. Python's `_is_overridden` checks `getattr(type(handler), method)`, which does not:

```python
h = Empty()  # no overrides
h.before_tool_call = lambda e: InterventionActions.deny("nope")
agent = Agent(..., interventions=[h])
agent("go")  # tool runs — deny silently never evaluated  ✗ (TS would enforce it)
```

For a *security* primitive, silent non-enforcement is the worst failure mode. Either match TS (also check `handler.__dict__`/instance attrs in `_is_overridden`) or explicitly document that only class-level overrides are detected.

### 🟡 Handler returning `None` → cryptic `AttributeError`

A handler that forgets to `return` (very easy in Python; TS catches this at compile time) crashes with `'NoneType' object has no attribute 'type'` deep in `_dispatch`. Cheap guard with a great DevEx payoff:

```python
if action is None:
    raise TypeError(f"Intervention handler '{handler.name}.{method}' returned None; expected an InterventionAction")
```

### 🟢 Minor / parity notes

1. **Guide docstring session-bypass note is missing.** The open review thread on `event.agent.messages.append(...)` says the limitation is "documented in Guide docstring" — but the current docstring only carries the convergence warning, no session-manager note. Please actually add it (blocked-by-#2631 reference is fine).
2. **Tracing diff vs TS:** Python starts `model_invoke_span` *before* `BeforeModelCallEvent`, so cancelled calls emit a model-invoke span with zero usage; TS checks cancel before starting the span and emits nothing. Cosmetic, but billing/latency dashboards will see phantom 0-token model calls. Consider moving the span start below the cancel check.
3. **Confirm sentinel:** TS `undefined` sentinel still allows `null` as a legitimate preemptive response; Python reserves `None` so that's impossible. Documented, acceptable — true parity isn't achievable here.
4. **Interrupt/resume re-runs the whole dispatch** — after a Confirm pause resumes, *all* handlers re-evaluate, so a `Transform` applies **twice**. Same in TS (inherent to re-firing the hook), but worth one docstring line: "handlers must be idempotent across interrupt/resume."
5. **Python is ahead of TS in two spots** (TS follow-ups, not blockers here): `_handle_error` logs on deny/proceed (TS `_handleError` is silent), and the unknown-`on_error` fallback raises (fail-closed) per lizradway's feedback.
6. **`LifecycleObserver`/`observeAgent` parity gap:** TS `agent.initialize()` calls `observeAgent` on intervention handlers (used by vended steering). Python has no equivalent — fine for this PR, but it's a prerequisite for porting `vended-interventions/steering`, so track it.

### ✅ Verification matrix

| Check | Result |
|---|---|
| `pytest tests/strands/interventions/` | ✅ 67/67 |
| `pytest tests/strands/agent/test_agent.py` | ✅ 129/129 |
| `pytest tests/strands/hooks/` | ✅ (34 event_loop failures are **pre-existing on main** — env issue, not this PR) |
| `ruff check` + `ruff format --check` | ✅ clean |
| `mypy` (project config) on interventions + hooks + event_loop | ✅ 0 errors |
| TS suite `src/interventions/__tests__/` | ✅ 55/55 (behavior reference) |
| Adversarial: deny/guide/confirm/transform × all 5 lifecycle points, sync+async handlers, on_error×3, guide accumulation, transform chaining, preemptive/falsy/pause confirm, interrupt resume | ✅ all match TS semantics **except the P0 above** |

### Verdict

**Request changes** for the P0 — it's a one-line fix plus the missing event-loop tests that would have caught it. Everything else in the interventions package itself is well-built: dispatch ordering, Deny short-circuit, Guide accumulation (`[name] feedback` joining matches TS byte-for-byte), Confirm↔interrupt bridge, and the mypy-strict-clean implementation are all faithful ports. With the cancel path fixed and tested, this is a solid parity implementation.

strands-py/src/strands/interventions/actions.py
```diff
@@ -0,0 +1,201 @@
...
+    cancel = sets event.cancel/cancel_tool, short-circuits (remaining handlers skipped)
+    cancel+ = sets cancel with accumulated feedback from all guiding handlers
+    confirm = uses preemptive response or interrupt, checks with evaluate, sets cancel if denied
+    inject = appends accumulated feedback as a user message so the model sees it on this call
+    inject + retry = appends accumulated feedback and retries so the model sees guidance
+    apply = calls action.apply(event) for in-place mutation, later handlers see the change
+"""
+
+
+def proceed(*, reason: str | None = None) -> Proceed:
```
mkmeral: do we need these helper methods? we added them in TS because we didn't want others to build custom objects, and make mistakes. in here we create data classes anyway, so I'm not sure about the value of these methods

agent-of-mkmeral:
## Summary @ `3d045c7` — what's left before merge

Consolidating the review chain into a single actionable list. The P0 cancel crash is **confirmed fixed** and behavior is verified TS-parity end-to-end (67/67 intervention, 129/129 agent, 34/34 event_loop tests; mypy + `ruff check` clean). Four items block merge, two need a maintainer call.

### 🔴 Blocking

1. **CI format gate** — `registry.py:216` fails `ruff format --check` (the 3-line `raise TypeError(...)` collapses to a 115-char one-liner). One-line fix.
2. **Cancel-path tests** — the event-loop/agent cancel branches still have zero dedicated tests. This exact gap shipped the P0 once already; ready-to-paste tests are in the review thread above. Please land them in this PR, not a follow-up.
3. **Exports** — typed handler authors need `Proceed`/`Deny`/`Guide`/`Confirm`/`Transform`/`InterventionAction`/`OnError` for return annotations, but today they're only importable from `strands.interventions.actions` (3 modules deep, not re-exported). TS exports all of these top-level. Please re-export from `strands.interventions` (ideally also top-level `strands`). Additive, zero risk.
4. **`**kwargs: Any` on the 5 lifecycle methods** — without it, adding any parameter later raises `TypeError` on every existing subclass → forced major version bump. TS doesn't have this problem (extra args are ignored); Python needs the escape hatch. Our own `HookProvider.register_hooks(self, registry, **kwargs)` already follows this pattern. Free to add now, impossible later. (If the position is "events are the only growth surface," that's defensible — but it should be an explicit recorded decision, not an accident.)

### 🟡 Maintainer call (not blocking)

<details>
<summary>5. Bare callable in <code>interventions=[...]</code> → cryptic AttributeError</summary>

`Agent(interventions=[some_function])` currently explodes at registration with `AttributeError: 'function' object has no attribute 'name'` — no hint what went wrong. Note our own `hooks=[...]` *accepts* bare callables, so users will reasonably try it. Either:
- **(a)** raise a clear `TypeError: interventions must be InterventionHandler instances, got function — did you mean hooks=[...]?` (minimal), or
- **(b)** accept callables with derived names like `hooks=` does — though `name` carries real semantics here (Guide prefixes `[name]`, interrupt IDs key on it), so requiring explicitness is defensible.

Either is fine; the silent `AttributeError` is not.
</details>

<details>
<summary>6. Instance-assigned handler methods are silently never enforced</summary>

`h.before_tool_call = lambda e: deny(...)` is detected and enforced in TS (`!== prototype` check) but silently skipped in Python (`_is_overridden` checks the class only). `3d045c7` documents this in the docstring, which is acceptable — but for a security primitive, silent non-enforcement is the worst failure mode. A one-line registration check (warn if a lifecycle-method name appears in `handler.__dict__`) turns it into a loud warning at near-zero cost. Recommend the warning.
</details>

Everything else (Guide/session bypass note, phantom 0-token tracing span, `defaultEvaluate` export, `observeAgent`/`LifecycleObserver` parity) is minor or tracked follow-up. Once 1–4 land, this is mergeable.

strands-py/src/strands/interventions/actions.py
mehtarac: Technically they're 1:1 pass-throughs to the dataclass constructors with identical signatures

mehtarac: Though I would want to keep them:

- explicitly passing None to a Callable field (evaluate=None) would cause `TypeError: 'NoneType' is not callable.` The `confirm()` guards against evaluate=None
- The cost of keeping them is ~30 lines of pass-through code

mehtarac: Update: Had an offline discussion - agreed on removing the helper methods

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -494,13 +495,39 @@ async def _handle_model_execution(
...
-                await agent.hooks.invoke_callbacks_async(
-                    BeforeModelCallEvent(
+                before_model_call_event = BeforeModelCallEvent(
+                    agent=agent,
+                    invocation_state=invocation_state,
+                    projected_input_tokens=projected_input_tokens,
+                )
+                await agent.hooks.invoke_callbacks_async(before_model_call_event)
+
+                if before_model_call_event.cancel:
```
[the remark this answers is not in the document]
lizradway: +1

strands-py/src/strands/interventions/actions.py
```diff
@@ -0,0 +1,137 @@
...
+    On beforeModelCall, injects feedback as a user message.
+    On afterModelCall, the response is discarded and the model retries with feedback.
+
+    .. warning::
+        On ``after_model_call``, Guide triggers a model retry. Handlers **must** ensure
+        convergence (e.g., by tracking retry count and escalating to Deny after repeated
+        failures). The framework imposes no retry cap on guide-triggered retries.
+
+    .. note::
+        On ``before_model_call`` and ``after_model_call``, guidance messages are injected
```
lizradway, line 78: @opieter-aws out of curiousity, are injected messages tracked by session management in your impl? would be some great motivation to move over after summit

strands-py/src/strands/interventions/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+    ```python
+    from strands import Agent, InterventionHandler
+    from strands.interventions import Deny, Proceed
+
+    class MyAuth(InterventionHandler):
+        name = "my-auth"
+
+        def before_tool_call(self, event):
+            if not self.is_authorized(event):
+                return Deny(reason="not authorized")
```
lizradway, line 17: why did we get rid of the namespacing?

lizradway (changes requested):

strands-py/src/strands/interventions/__init__.py
lizradway: this should be:

```
from strands import InterventionActions
InterventionActions.deny("reason")
```

if we want to do something more pythonic, that is fine, but that needs api bar raising. additionally, the action factory functions are now unused, so I'm guessing that this was an accident rather than on purpose?

mehtarac: related thread here: https://github.com/strands-agents/harness-sdk/pull/2693#discussion_r3396238743

lizradway: could this be `InterventionsActions.Deny()` then?

keep the namespace without the factory functions?

i think that plain `deny` or `confirm` are really confusing for those without context, which is why we namespaced in TS

strands-py/src/strands/event_loop/event_loop.py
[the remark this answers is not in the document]
mehtarac: This is fixed now

strands-py/src/strands/interventions/__init__.py
mehtarac: IMO Python's standard pattern for namespacing types is the module itself. `from strands.interventions import Deny` would be more pythonic and ideally the import being from interventions path would help customers with the confusion.

strands-py/src/strands/interventions/actions.py
opieter-aws: No, it's using middleware, which allows you to pass a modified copy of the messages object to the model call, rather than altering the messages array.

agent-of-mkmeral:
## Re-review @ `905da50` — all 4 blockers resolved ✅ (one new regression from the factory removal)

Checked out the new head and re-verified everything locally. **All four blocking items from the previous summary are fixed:**

| # | Blocker | Status | Verification |
|---|---|---|---|
| 1 | `ruff format` on `registry.py:216` | ✅ fixed | `ruff format --check` clean; CI **Python / Lint passes** |
| 2 | Cancel-path tests | ✅ added | new `test_cancel_paths.py` (7 tests) covers exactly the requested scenarios: Deny→`end_turn` + `DENIED:` text, model **not** invoked, plain-hook `cancel=True`→default text, cancel+retry→loop re-entry, `before_invocation` deny ×3. 74/74 intervention tests pass |
| 3 | Exports | ✅ resolved | `Proceed`/`Deny`/`Guide`/`Confirm`/`Transform`/`InterventionAction`/`OnError` all importable from `strands.interventions` (verified). Module docstring example updated to match |
| 4 | `**kwargs: Any` on lifecycle methods | ✅ added | all 5 methods |

Also noted: the factory functions **and** `InterventionActions` were removed entirely in favor of direct dataclass construction (`Deny(reason="...")`), per @mkmeral's open thread. The PR description was updated to match. Direct dataclasses are arguably more Pythonic and the namespacing concern is moot since `Deny`/`Proceed` are self-documenting names — but note this supersedes the resolution of @lizradway's namespacing thread (which was resolved via `InterventionActions`), so flagging for her re-confirmation.

### 🔴 New regression: removing `confirm()` reintroduced the exact hazard it guarded against

The factory's stated purpose ([thread](https://github.com/strands-agents/harness-sdk/pull/2693#discussion_r-actions)) was coalescing `evaluate=None` → `default_evaluate`. With the factory gone, that guard is gone:

```python
Confirm(prompt="ok?", response=True, evaluate=None)
# → TypeError: 'NoneType' object is not callable   (at dispatch time, deep in the registry)
```

Verified on this head. Since `evaluate: Callable[[Any], bool] = field(default=default_evaluate)` only protects *omission*, not explicit `None`, please add a `__post_init__` guard to the frozen dataclass:

```python
def __post_init__(self) -> None:
    if self.evaluate is None:
        object.__setattr__(self, "evaluate", default_evaluate)
```

(or type it `Callable[[Any], bool] | None` and resolve at use site — either way, explicit `None` shouldn't detonate at dispatch time in an approval primitive).

<details>
<summary>Still open from earlier reviews (maintainer calls, not blockers)</summary>

- **Bare callable in `interventions=[...]`** still dies with `AttributeError: 'function' object has no attribute 'name'` at registration (verified) — a clear `TypeError` would help, since `hooks=[...]` *does* accept callables.
- **Instance-assigned handler methods** still silently unenforced (`_is_overridden` checks the class only — verified returns `False`; TS detects this). Docs-only stance unchanged.
- **Phantom 0-token span** — `model_invoke_span` still starts before the cancel check, so cancelled calls emit a 0-usage span; TS emits nothing.

</details>

### CI status

All *code* gates are green (Lint, unit tests on py3.10–3.14 × linux/windows). The remaining failures are **process checks**: `check-api-review-label` (PR carries `api/needs-review` without `api/review-complete` — awaiting API-review sign-off), `label-size`, and `check-access-and-checkout`.

**Verdict:** with the `Confirm(evaluate=None)` guard added, this is mergeable from my side — pending the API-review label sign-off.

lizradway (approved):

PR #2696, opened 2026-06-09
fix(a2a): isolate conversation state per context in TypeScript

JackYPCOnline:
## Description

Ports the Python A2A context-isolation fix (#2628) to the TypeScript SDK. The TS `A2AExecutor` previously held a single `Agent` and reused it for every request, so conversation history leaked between callers sharing the server. This brings TS to parity with Python's per-`contextId` isolation.

Two modes, mirroring Python:

- **`agentFactory`** (recommended): a dedicated `Agent` per `contextId`, each serialized by its own lock so distinct contexts run concurrently and never share state. The factory is where per-context concerns like a `sessionManager` are wired.
- **single `agent`** (deprecated): conversation state is snapshot-swapped on/off one shared instance under a global lock. Emits a `DeprecationWarning` and rejects a configured `sessionManager` (it would interleave every context into one session).

Both modes bound memory with an LRU cap (`maxContexts`, default 1000). Per-context state round-trips through the `session` snapshot preset, which already includes `modelState` — so no snapshot-schema change was needed on the TS side (unlike Python, which added it in #2680).

## Public API Changes

`A2AExecutor` and `A2AServer` now accept `agentFactory` and `maxContexts`; the single `agent` becomes optional (provide exactly one). Existing `A2AServer({ agent, ... })` keeps working, now isolated, with a deprecation warning.

```typescript
// Recommended
const server = new A2AServer({
  agentFactory: (contextId) => new Agent({ model, sessionManager: sessionFor(contextId) }),
  name: 'My Agent',
})
```

A small `AsyncLock` primitive is added (JS has no `asyncio.Lock`) to serialize same-context requests in factory mode and all requests in single-agent mode.

## Breaking Changes

None. The single-`agent` path is preserved (now isolated) and only warns.

## Type of Change

Bug fix

## Testing

## Checklist

### DIFF
--- a/strands-py/src/strands/multiagent/a2a/executor.py
+++ b/strands-py/src/strands/multiagent/a2a/executor.py
@@ -49,11 +49,7 @@ AgentFactory = Callable[[str], SAAgent]
 
 @dataclass
 class _StreamState:
-    """Per-invocation A2A-compliant streaming state.
-
-    Held locally per request rather than on the executor: in factory mode different contexts run
-    concurrently, so shared instance attributes would corrupt each other's artifact tracking.
-    """
+    """Per-invocation A2A-compliant streaming state."""
 
     artifact_id: str
     is_first_chunk: bool = True
@@ -61,11 +57,7 @@ class _StreamState:
 
 @dataclass
 class _ContextEntry:
-    """Per-context bookkeeping for factory mode: a dedicated agent and its serializing lock.
-
-    Keeping the agent and lock in one entry means the LRU map has a single source of truth; there
-    is no second map to keep in sync on insert, reorder, or eviction.
-    """
+    """Per-context bookkeeping for factory mode: a dedicated agent and its serializing lock."""
 
     agent: SAAgent
     lock: asyncio.Lock
@@ -150,8 +142,7 @@ class StrandsA2AExecutor(AgentExecutor):
         self._contexts_lock = asyncio.Lock()
 
         if agent_factory is not None:
-            # Factory mode: a dedicated agent and lock per context. The per-context lock serializes
-            # only same-context requests, so different contexts run concurrently.
+            # Factory mode: a dedicated agent and lock per context.
             self.agent: SAAgent | None = None
             self._contexts: OrderedDict[str, _ContextEntry] = OrderedDict()
         else:
@@ -169,7 +160,6 @@ class StrandsA2AExecutor(AgentExecutor):
                 DeprecationWarning,
                 stacklevel=2,
             )
-            # The template snapshot is the agent's clean state, captured before any request mutates it.
             self.agent = agent
             self._template_snapshot = self._capture_state(agent)  # type: ignore[arg-type]
             self._snapshots: OrderedDict[str, Snapshot] = OrderedDict()
@@ -179,12 +169,7 @@ class StrandsA2AExecutor(AgentExecutor):
         return agent.take_snapshot(preset="session")
 
     def _restore_state(self, agent: SAAgent, snapshot: Snapshot) -> None:
-        """Load a snapshot into an agent, restoring its session state.
-
-        No defensive copy is needed: ``load_snapshot`` deep-copies every field it restores (and
-        ``take_snapshot`` deep-copies on capture), so a run never mutates a stored or template
-        snapshot in place.
-        """
+        """Load a snapshot into an agent, restoring its session state."""
         agent.load_snapshot(snapshot)
 
     def _evict_excess_contexts(self) -> None:
@@ -227,18 +212,13 @@ class StrandsA2AExecutor(AgentExecutor):
         updater: TaskUpdater,
         stream_state: _StreamState | None,
     ) -> None:
-        """Single-agent mode: swap this context's snapshot on/off the shared agent under a lock.
-
-        The lock serializes all requests (a single ``Agent`` cannot be invoked concurrently). The
-        agent is reset to the template afterward so no context's data lingers on it.
-        """
+        """Single-agent mode: swap this context's snapshot on/off the shared agent under a lock."""
         async with self._contexts_lock:
             self._restore_state(self.agent, self._snapshots.get(context_id, self._template_snapshot))  # type: ignore[arg-type]
             try:
                 await self._stream_agent(self.agent, content_blocks, invocation_state, updater, stream_state)  # type: ignore[arg-type]
             finally:
-                # Persist this context's updated history (even on error/cancel, to retain partial
-                # turns), evict beyond the cap, then reset the shared agent for the next caller.
+                # Persist updated history (even on error), evict, then reset the agent for the next caller.
                 self._snapshots[context_id] = self._capture_state(self.agent)  # type: ignore[arg-type]
                 self._snapshots.move_to_end(context_id)
                 self._evict_excess_contexts()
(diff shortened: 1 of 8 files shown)


strands-ts/src/a2a/async-lock.ts
```diff
@@ -0,0 +1,33 @@
...
+  /** Resolves when the currently-held lock (if any) is released. */
+  private _tail: Promise<void> = Promise.resolve()
+
+  /**
+   * Acquires the lock, waiting until all previously-acquired holders release.
+   *
+   * @returns A release function. Call it exactly once (e.g. in a `finally`) to
+   *   let the next waiter proceed.
+   */
+  async acquire(): Promise<() => void> {
```
zastrowm: IMHO we should return something that uses dispose: Symbol.dispose

We use this elsewhere and handles the finally part for us

strands-ts/src/a2a/executor.ts
```diff
@@ -8,11 +8,97 @@
...
+import type { Snapshot } from '../types/snapshot.js'
+import type { ContentBlock } from '../types/messages.js'
 import { ModelStreamUpdateEvent, ContentBlockEvent } from '../hooks/events.js'
 import { contentBlocksToParts, partsToContentBlocks } from './adapters.js'
 import { normalizeError } from '../errors.js'
 import { logger } from '../logging/logger.js'
+import { AsyncLock } from './async-lock.js'
+
+/** Deep-copies a snapshot. Snapshots are JSON-serializable, so a round-trip is sufficient. */
+function cloneSnapshot(snapshot: Snapshot): Snapshot {
```
zastrowm: Why do we need a clone?

strands-ts/src/a2a/executor.ts
```diff
@@ -80,14 +270,79 @@ export class A2AExecutor implements AgentExecutor {
...
+   * The lock serializes all requests (a single agent cannot be invoked concurrently). The
+   * agent is reset to the template afterward so no context's data lingers on it.
+   */
+  private async _runWithSharedAgent(
+    context: RequestContext,
+    contentBlocks: ContentBlock[],
+    eventBus: ExecutionEventBus
+  ): Promise<void> {
+    const agent = this._agent!
+    const release = await this._contextsLock.acquire()
```
zastrowm: `using release = await this._contextsLock.acquire()` without the try/finally would be the flow if we had dispose

strands-ts/src/a2a/executor.ts
```diff
@@ -80,14 +270,79 @@ export class A2AExecutor implements AgentExecutor {
...
+    try {
+      this._restoreState(agent, this._snapshots.get(context.contextId) ?? this._templateSnapshot!)
+      try {
+        await this._streamAgent(agent, context, contentBlocks, eventBus)
+      } finally {
+        // Persist this context's updated history (even on error, to retain partial turns),
+        // evict beyond the cap, then reset the shared agent for the next caller. Delete-then-set
```
zastrowm: > Delete-then-set
>        // places the entry at the most-recently-used end whether or not it already existed.

Is this behavior guaranteed by JS engines?

strands-ts/src/a2a/executor.ts
```diff
@@ -22,49 +108,153 @@ import { logger } from '../logging/logger.js'
...
+    while (contexts.size > this._maxContexts) {
+      const evictedId = contexts.keys().next().value as string
+      contexts.delete(evictedId)
+      logger.debug(`context_id=<${evictedId}> | evicted least-recently-used A2A context`)
+    }
+  }
+
+  /** Return the dedicated agent and lock for a context, building it on first use (factory mode). */
+  private async _acquireContextAgent(contextId: string): Promise<ContextEntry> {
+    const release = await this._contextsLock.acquire()
```
zastrowm: We don't need a lock here.  JS guarantees there's only one thing running at a time.  Removing the lock here simplifies things a bit

strands-ts/src/a2a/executor.ts
```diff
@@ -80,14 +270,79 @@ export class A2AExecutor implements AgentExecutor {
...
+  /**
+   * Factory mode: run against this context's dedicated agent, serialized only per context.
+   */
+  private async _runWithContextAgent(
+    context: RequestContext,
+    contentBlocks: ContentBlock[],
+    eventBus: ExecutionEventBus
+  ): Promise<void> {
+    const { agent, lock } = await this._acquireContextAgent(context.contextId)
+    const release = await lock.acquire()
```
zastrowm: This is to avoid running multiple invocations of the **same** agent behind A2A right?  E.g. it doesn't mean that only one agent can be running at at time, just that any *single* angent can't have multiple requests against it? 

If so, did we have this behavior before?

Long term, this should live on the agent (Python has customizable concurrent invoking ability)

strands-ts/src/a2a/async-lock.ts
JackYPCOnline: Good to know, will update on next iteratation

strands-ts/src/a2a/executor.ts
JackYPCOnline: I believe it is from python, in python i have `agent.load_snapshot(copy.deepcopy(snapshot))`, actually I don't need deepcopy here as well, will remove from both.

strands-ts/src/a2a/executor.ts
JackYPCOnline: Good to know that

strands-ts/src/a2a/executor.ts
JackYPCOnline: JavaScript `Map` object, setting a value for an existing key updates the value but perfectly preserves its original position in the iteration order.  
guaranteed by `ECMAScript`

strands-ts/src/a2a/executor.ts
JackYPCOnline: yes, it is just per context lock. we don't have this before but it pairs with python behavior. Agreed it should live on the agent long term.  Can be a follow-up.

strands-py/src/strands/multiagent/a2a/executor.py
```diff
@@ -182,10 +181,11 @@ def _capture_state(self, agent: SAAgent) -> Snapshot:
     def _restore_state(self, agent: SAAgent, snapshot: Snapshot) -> None:
         """Load a snapshot into an agent, restoring its session state.
 
-        Deep-copies once at the boundary so a run never mutates a stored or template snapshot in
-        place (snapshots, including the template, are restored repeatedly).
+        No defensive copy is needed: ``load_snapshot`` deep-copies every field it restores (and
+        ``take_snapshot`` deep-copies on capture), so a run never mutates a stored or template
+        snapshot in place.
```
zastrowm: Nit this is just noise

strands-ts/src/a2a/executor.ts
```diff
@@ -211,14 +206,16 @@ export class A2AExecutor implements AgentExecutor {
   /**
    * Load a snapshot into an agent, restoring its session state.
    *
-   * Deep-copies once at the boundary so a run never mutates a stored or template
-   * snapshot in place (snapshots, including the template, are restored repeatedly).
+   * No defensive copy is needed: `loadSnapshot` reconstructs the agent's state
+   * from the snapshot rather than referencing it — messages are rebuilt as new
+   * instances and `StateStore` deep-copies the state on load — so a subsequent
+   * run cannot mutate a stored or template snapshot in place.
    */
```
zastrowm: Again, this comment is really just noise

strands-ts/src/a2a/executor.ts
```diff
@@ -230,24 +227,27 @@ export class A2AExecutor implements AgentExecutor {
...
-        this._contexts.delete(contextId)
-        this._contexts.set(contextId, entry)
-      }
-      return entry
-    } finally {
-      release()
+  /**
+   * Return the dedicated agent and lock for a context, building it on first use (factory mode).
+   *
+   * No lock guards the map here: the factory is synchronous, so this method never yields between
```
zastrowm: > No lock guards the map here: the factory is synchronous, so this method never yields between

Noise

strands-ts/src/a2a/executor.ts
```diff
@@ -230,24 +227,27 @@ export class A2AExecutor implements AgentExecutor {
...
+   * check-then-create-or-reorder sequence atomic with respect to other contexts.
+   */
+  private _acquireContextAgent(contextId: string): ContextEntry {
+    let entry = this._contexts.get(contextId)
+    if (entry === undefined) {
+      entry = { agent: this._agentFactory!(contextId), lock: new AsyncLock() }
+      this._contexts.set(contextId, entry)
+      this._evictExcessContexts()
+    } else {
+      // Mark most-recently-used: delete-then-set moves the entry to the Map's end.
```
zastrowm: Thanks; this really makes it clear to future readers

zastrowm (dismissed):

zastrowm (approved):

PR #2705, opened 2026-06-09
ci: add dependabot auto-merge workflow

yonib05:
## Summary

- Add `dependabot-auto-merge.yml`: analyzes dependabot PRs with a read-only Strands agent, runs integration tests on safe Python/TypeScript updates, and enables auto-merge
- Add `parse_dependabot_title.py` and `sanitize_dependabot_body.py` (with tests) to defend against prompt injection
- Add reusable `dependabot-integration-tests.yml` routing by ecosystem (python/typescript)

Depends on strands-agents/devtools#61 (dependabot-analyze SOP, mode wiring, and `sanitized_changelog` agent-runner input). That PR must merge first since this workflow references the devtools actions at `@main`.

## How it works

1. On a dependabot PR, the title is parsed for structured metadata and the body is sanitized into an `<untrusted-changelog>` block
2. The strands-command agent runs read-only: reads the diff, searches the repo for package usage, consumes the sanitized changelog, and posts an analysis comment with a verdict
3. The verdict becomes a `Dependabot Analysis` check run
4. `safe` Python/TypeScript updates → integration tests → auto-merge once branch protection is satisfied
5. `safe` docs/actions updates → labeled `dependabot-safe`, left for manual merge (preview link review)
6. Anything flagged → labeled `dependabot-needs-review`

## Security

All external content (PR body, changelog) is sanitized and wrapped in `<untrusted-changelog>` tags before reaching the agent. The agent runs with `write_permission: false`. Upstream commit fetches are restricted to URL-validated GitHub sources.

## Test plan

- [ ] Merge devtools#61 first
- [ ] Run `/strands dependabot-analyze` on an existing dependabot PR; confirm the analysis comment and check run
- [ ] Confirm a safe Python/TS update flows to auto-merge after approval
- [ ] Confirm a flagged update gets `dependabot-needs-review`

### DIFF
--- a/.github/scripts/parse_dependabot_title.py
+++ b/.github/scripts/parse_dependabot_title.py
@@ -1,0 +1,69 @@
+"""Parse a dependabot PR title into structured metadata.
+
+Runs in the dependabot-auto-merge workflow BEFORE the agent sees anything,
+so the agent receives only validated structured fields, never the raw title.
+
+Usage: python parse_dependabot_title.py "<title>"
+Outputs JSON to stdout.
+"""
+
+import json
+import re
+import sys
+
+# Maps dependabot commit prefixes to ecosystems.
+PREFIX_TO_ECOSYSTEM = {
+    "ci(python)": "python",
+    "ci(typescript)": "typescript",
+    "ci(docs)": "docs",
+}
+
+# Matches "bump <pkg> from <old> to <new>" (single-package updates).
+SINGLE_RE = re.compile(
+    r"bump\s+(?P<package>[\w@/\-\.]+)\s+from\s+(?P<old>[\w\.\-]+)\s+to\s+(?P<new>[\w\.\-]+)",
+    re.IGNORECASE,
+)
+
+# Matches grouped updates: "bump the <group> group ...".
+GROUPED_RE = re.compile(r"bump\s+the\s+.+\s+group", re.IGNORECASE)
+
+
+def parse(title: str) -> dict:
+    result = {
+        "ecosystem": "unknown",
+        "package": "",
+        "old_version": "",
+        "new_version": "",
+        "grouped": False,
+    }
+
+    # Determine ecosystem from prefix.
+    prefix_match = re.match(r"^(ci\([\w]+\)|ci)(?=:)", title)
+    if prefix_match:
+        prefix = prefix_match.group(1)
+        if prefix == "ci":
+            result["ecosystem"] = "actions"
+        else:
+            result["ecosystem"] = PREFIX_TO_ECOSYSTEM.get(prefix, "unknown")
+
+    if GROUPED_RE.search(title):
+        result["grouped"] = True
+        return result
+
+    m = SINGLE_RE.search(title)
+    if m:
+        result["package"] = m.group("package")
+        result["old_version"] = m.group("old")
+        result["new_version"] = m.group("new")
+
+    return result
+
+
+if __name__ == "__main__":
+    # Prefer stdin (avoids any shell interpretation of attacker-influenceable
+    # title text); fall back to argv for convenience/testing.
+    if len(sys.argv) > 1:
+        title = sys.argv[1]
+    else:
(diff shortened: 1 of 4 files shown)


yonib05:
Thanks for the review. Addressed in 978c477:

**Critical (found in dependabot-analyze devtools PR):** the input parser crashed under `pull_request_target` because it only honored explicit inputs for `workflow_dispatch`. Fixed in strands-agents/devtools#61 — it now uses explicit `issue_id`/`command` inputs whenever provided, falling back to the comment payload only for `issue_comment` events.

**Sanitizer (line 87, 92):** now decodes HTML entities before injection-pattern matching, and truncates the untrusted payload *before* prepending the preamble/wrapper (so the cap applies only to attacker-influenceable text and the closing tag can't be truncated away).

**Shell arg (line 34):** PR title is now piped via stdin instead of passed as a positional argument; the parser reads stdin when no argv is given.

**Heredoc delimiter (line 44):** now uses a random `SANITIZED_$(openssl rand -hex 8)` delimiter.

**Verdict filter (line 111):** restricted to comments authored by `github-actions[bot]` (the agent posts via `GITHUB_TOKEN`), preventing forged-verdict injection from arbitrary bot/strands-named accounts.

**flag-for-review (line 177):** condition now fires whenever a safe python/typescript verdict's integration tests did not `success` (covers `failure`, `cancelled`, and erroneously-`skipped`). Verified exhaustively that every verdict×ecosystem×result combo has exactly one terminal outcome.

**pull_request_target code execution:** added a same-repo head guard (`head.repo.full_name == github.repository`) to both integration-test jobs so fork heads can never run with the privileged test role.

**Test paths:** script tests now resolve the script via `pathlib` so they run from any CWD.

Left as-is (intentional): the Python/TypeScript jobs use different AWS role secrets to mirror the existing `*-integration-test.yml` workflows; the `working-directory: strands-py` monorepo checkout pattern; and the read-only `analyze` job permissions. Wiring the script tests into a CI job is a reasonable follow-up but out of scope here.

yonib05:
Removed the unit test files from the PR (force-pushed). They were run locally to validate the parser and sanitizer (15 passing, including HTML-entity injection and truncation-boundary cases) but are kept out of the PR per repo preference. The branch now contains only the two scripts and two workflows.

yonib05:
Good catch on the non-dependabot cascade — fixed in dd9d9d6.

The `prep` job has the `dependabot[bot]` actor gate, but `finalize`'s `if: always()` would have run on every PR (with `prep`/`parse-input`/`analyze` skipped), cascading into `verdict` and `flag-for-review` and producing a spurious `action_required` check run / `dependabot-needs-review` label on unrelated PRs.

Fixes:
- `finalize` now guarded with `if: always() && needs.parse-input.result != 'skipped'`, so it (and therefore `verdict`) does not run when the pipeline was skipped for a non-dependabot PR.
- `flag-for-review` additionally requires `needs.verdict.result == 'success'`, so it can never label a PR whose verdict job didn't actually run.

Verified exhaustively that (a) no terminal job fires for a non-dependabot PR, and (b) every dependabot verdict×ecosystem×test-result combination still resolves to exactly one terminal outcome.

yonib05:
Addressed in 2458909 (+ devtools#61 @ 0d41152):

- **Non-dependabot cascade (the `if: always()` finalize → verdict → flag chain):** fixed in dd9d9d6 — `finalize` is now guarded with `needs.parse-input.result != 'skipped'` and `flag-for-review` requires `needs.verdict.result == 'success'`. No terminal job fires for non-dependabot PRs (verified exhaustively).
- **Verdict shadowing by an unrelated agent comment (e.g. a prior `/strands review`):** the SOP now emits a unique `DEPENDABOT_VERDICT: {"verdict": "..."}` marker line, and the workflow only matches `github-actions[bot]` comments that contain that marker (most recent wins). A `/strands review` comment without the marker can no longer be mistaken for a verdict; anything unmatched safely defaults to `needs-review`.

PR #2715, opened 2026-06-10
feat: migrate TS integration test role to shared test-infra CDK stack

notowen333:
## Description

Migrate the TypeScript SDK integration test role into the shared `test-infra/` CDK stack. This aligns the TS role with the pattern already used by the Python SDK, where feature-scoped constructs (Bedrock KB, SSH EC2) layer scoped grants onto a shared identity.

Changes:
- Add `sts:GetCallerIdentity` to the base role (used by session tests to derive account ID for bucket naming, applies to all deploys including community)
- Add missing `us.*` inference profile for Claude Sonnet 4.5 to the IAM role (required by caching tests)
- Bump deprecated model references in test fixtures to current models
- Wire the TS integration test workflow to use the role from the shared stack
- Add CDK constructs for Bedrock Knowledge Base and SSH EC2 test resources with scoped IAM grants
- Add `TestFeatureConstruct` base class with SSM parameter helpers

## Related Issues

N/A

## Type of Change

New feature

## Testing

- Verified all model IDs used in integration tests have corresponding ARNs in the role
- CI will validate end-to-end once the stack is deployed

## Checklist

----

### DIFF
--- a/.github/workflows/typescript-integration-test.yml
+++ b/.github/workflows/typescript-integration-test.yml
@@ -44,7 +44,7 @@ jobs:
       - name: Configure Credentials
         uses: aws-actions/configure-aws-credentials@v6
         with:
-         role-to-assume: ${{ secrets.AWS_ROLE_ARN }}
+         role-to-assume: ${{ secrets.STRANDS_INTEG_TEST_ROLE }}
          aws-region: us-east-1
          mask-aws-account-id: true
 
--- a/strands-ts/test/integ/__fixtures__/model-providers.ts
+++ b/strands-ts/test/integ/__fixtures__/model-providers.ts
@@ -43,7 +43,7 @@ export const bedrock = {
   models: {
     default: {},
     reasoning: {
-      modelId: 'us.anthropic.claude-sonnet-4-20250514-v1:0',
+      modelId: 'global.anthropic.claude-sonnet-4-6',
       additionalRequestFields: { thinking: { type: 'enabled', budget_tokens: 1024 } },
     },
     video: { modelId: 'us.amazon.nova-pro-v1:0' },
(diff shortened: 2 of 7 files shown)


.github/workflows/typescript-integration-test.yml
```diff
@@ -44,7 +44,7 @@ jobs:
       - name: Configure Credentials
         uses: aws-actions/configure-aws-credentials@v6
         with:
-         role-to-assume: ${{ secrets.AWS_ROLE_ARN }}
+         role-to-assume: ${{ secrets.STRANDS_INTEG_TEST_ROLE }}
```
zastrowm, line 47: Do Strands-Dev have the same ability to run the tests locally?

strands-ts/test/integ/__fixtures__/model-providers.ts
```diff
@@ -245,7 +245,7 @@ export const vercelBedrock = {
       providerOptions?: Record<string, unknown>
     }
     return new VercelModel({
-      provider: provider('us.anthropic.claude-sonnet-4-20250514-v1:0'),
+      provider: provider('global.anthropic.claude-sonnet-4-6'),
```
zastrowm, line 248: Given your other changes, why did we need to change this?

zastrowm (approved):

strands-ts/test/integ/__fixtures__/model-providers.ts
notowen333: The model was deprecated so I thought that was simpler than adding permission on the deprecated model

.github/workflows/typescript-integration-test.yml
notowen333: The API keys should be the same. The deployed infra will not work out of the box... I think we can follow up with a path forward there it's the same for py + TS

PR #2719, opened 2026-06-10
feat: add extraction defaults and support for bedrock kbs

opieter-aws:
## Description

Enabling automatic extraction on a memory store required spelling out a full `ExtractionConfig` — a `trigger`, and (for a store that distills facts client-side) an `extractor`. That is boilerplate for the common case, and it left the `BedrockKnowledgeBaseStore` unable to participate in extraction at all: it exposed no way to opt in, so the `MemoryManager` silently skipped it.

This PR turns extraction config into a `boolean | ExtractionConfig` shorthand with sensible defaults, mirroring the existing `searchToolConfig` / `addToolConfig` pattern on the same object. `extraction: true` is now enough to turn it on, and every store — current and future — inherits the behavior. The Bedrock KB store gains extraction support for free as a result.

Resolution is centralized in one place, `resolveExtractionConfig`, invoked once in the `MemoryManager` constructor; the coordinator consumes the resolved config and no longer reads raw `store.extraction`. The default extractor is chosen by the store's write sinks: a store with `add` but no `addMessages` (like a KB) defaults to a `ModelExtractor` that distills facts client-side, while a store with `addMessages` defaults to the no-model passthrough of raw messages. This keeps `extraction: true` from silently incurring a model call on stores that have a cheaper server-side path.

### Public API Changes

`MemoryStoreConfig.extraction` now accepts a boolean shorthand, and `ExtractionConfig.trigger` is optional:

```typescript
// Before: a full config was required, with an explicit trigger
const store = new BedrockKnowledgeBaseStore({
  name: 'personal',
  knowledgeBaseId: 'KB123',
  writable: true,
  dataSourceType: 'CUSTOM',
  dataSourceId: 'DS456',
  extraction: { trigger: new IntervalTrigger({ turns: 5 }), extractor: new ModelExtractor() },
})

// After: opt in with defaults
const store = new BedrockKnowledgeBaseStore({
  name: 'personal',
  knowledgeBaseId: 'KB123',
  writable: true,
  dataSourceType: 'CUSTOM',
  dataSourceId: 'DS456',
  extraction: true,
})
```

`extraction: true` resolves to: a trigger every 5 turns, the default message filter (drops tool traffic), and a capability-based extractor. Passing an `ExtractionConfig` still works and now defaults any field left unset, so an explicit `extractor` or `trigger` continues to take precedence.

The change is backward compatible: existing configs that set `trigger` and `extractor` explicitly are unaffected. An explicit empty `trigger: []` is still rejected at construction.

## Related Issues

N/A

## Documentation PR

N/A

## Type of Change

New feature

## Testing

Verified in `strands-ts/` with `npx tsc --noEmit`, `npx eslint`, `npx prettier --check`, and the `vitest` unit suites for `src/memory` and `src/vended-memory-stores/bedrock-knowledge-base`. Added a direct `resolve.test.ts` for the resolver plus manager-level and KB-store coverage of the new shorthand and capability-based defaults.

## Checklist

----

### DIFF
--- a/strands-ts/src/memory/extraction/coordinator.ts
+++ b/strands-ts/src/memory/extraction/coordinator.ts
@@ -6,7 +6,10 @@ import { normalizeError } from '../../errors.js'
 import type { MemoryMessageFilter } from './types.js'
 import type { ResolvedExtractionConfig } from './resolve-extraction-config.js'
 
-/** A store paired with its fully-resolved extraction config. */
+/**
+ * A store paired with its fully-resolved extraction config.
+ * @internal
+ */
 export interface ExtractionBinding {
   /** The memory store to extract into. */
   store: MemoryStore
@@ -14,10 +17,16 @@ export interface ExtractionBinding {
   config: ResolvedExtractionConfig
 }
 
-/** Number of consecutive save failures after which a store backs off (stops trying every turn). */
+/**
+ * Number of consecutive save failures after which a store backs off (stops trying every turn).
+ * @internal
+ */
 export const SAVE_FAILURES_BEFORE_BACKOFF = 10
 
-/** While backed off, a store retries only once every this many save attempts (a probe). */
+/**
+ * While backed off, a store retries only once every this many save attempts (a probe).
+ * @internal
+ */
 export const BACKOFF_PROBE_INTERVAL = 3
 
 /** A buffered message and its sequence number. */
@@ -83,6 +92,7 @@ function _filterMessages(buffered: BufferedMessage[], filter: MemoryMessageFilte
  *
  * Saving itself either runs the store's extractor to pull out facts, or hands the raw messages to the
  * store - see {@link _write}.
+ * @internal
  */
 export class ExtractionCoordinator {
   private readonly _stores: MemoryStore[]
--- a/strands-ts/src/memory/extraction/coordinator.ts
+++ b/strands-ts/src/memory/extraction/coordinator.ts
@@ -4,7 +4,7 @@ import type { Model } from '../../models/model.js'
 import { logger } from '../../logging/logger.js'
 import { normalizeError } from '../../errors.js'
 import type { MemoryMessageFilter } from './types.js'
-import type { ResolvedExtractionConfig } from './resolve.js'
+import type { ResolvedExtractionConfig } from './resolve-extraction-config.js'
 
 /** A store paired with its fully-resolved extraction config. */
 export interface ExtractionBinding {
@@ -87,7 +87,7 @@ function _filterMessages(buffered: BufferedMessage[], filter: MemoryMessageFilte
 export class ExtractionCoordinator {
   private readonly _stores: MemoryStore[]
   /** Per store: its resolved extraction config (triggers, extractor, filter). */
-  private readonly _configs = new Map<MemoryStore, ResolvedExtractionConfig>()
+  private readonly _storeToExtractionConfig = new Map<MemoryStore, ResolvedExtractionConfig>()
   private readonly _defaultModel: Model
   /** The shared list of messages waiting to be saved, oldest first. Each is tagged with its `seq`. */
   private _pending: BufferedMessage[] = []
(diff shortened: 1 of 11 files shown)


strands-ts/src/memory/extraction/coordinator.ts
```diff
@@ -71,6 +80,8 @@ function _filterMessages(messages: MessageData[], filter: MemoryMessageFilter):
  */
 export class ExtractionCoordinator {
   private readonly _stores: MemoryStore[]
+  /** Per store: its resolved extraction config (triggers, extractor, filter). */
+  private readonly _configs = new Map<MemoryStore, ResolvedExtractionConfig>()
```
notowen333: nit: `storeToExtractionConfig` or some other more map style name is preferred

strands-ts/src/memory/extraction/resolve-extraction-config.ts
```diff
@@ -0,0 +1,75 @@
+import type { MemoryStore } from '../types.js'
```
notowen333, line 1: maybe `resolve.ts` is also an overly terse name?

strands-ts/src/memory/extraction/resolve.ts
```diff
@@ -0,0 +1,75 @@
...
+  const triggers =
+    config.trigger === undefined
+      ? [new IntervalTrigger({ turns: DEFAULT_EXTRACTION_TRIGGER_TURNS })]
+      : Array.isArray(config.trigger)
+        ? config.trigger
+        : [config.trigger]
+
+  let extractor = config.extractor
+  if (extractor === undefined) {
+    // A store with addMessages (or both sinks) takes the passthrough - no implicit model call. Only an
```
notowen333: or both sinks?

strands-ts/src/memory/extraction/resolve.ts
```diff
@@ -0,0 +1,75 @@
...
+    // add-only store defaults to distilling facts with a ModelExtractor.
+    const hasAdd = typeof store.add === 'function'
```
notowen333: Checking that this is the SOT here... i.e. are we handling if that store doesn't have writes turned on?

strands-ts/src/memory/extraction/resolve.ts
```diff
@@ -0,0 +1,75 @@
...
+    const hasAddMessages = typeof store.addMessages === 'function'
```
notowen333: same question

strands-ts/src/memory/extraction/resolve.ts
```diff
@@ -0,0 +1,75 @@
...
+    if (hasAdd && !hasAddMessages) {
+      extractor = new ModelExtractor()
+    }
+  }
+
+  return {
+    triggers,
+    ...(extractor !== undefined && { extractor }),
+    filter: config.filter ?? DEFAULT_MEMORY_MESSAGE_FILTER,
```
notowen333: Would prefer to resolve the filter with the default outside of the return statement

notowen333: The motivation is to have all the default resolution next to each other for clarity

strands-ts/src/memory/extraction/resolve.ts
opieter-aws: In case of both sinks, (`add` and `addMessages`), the passthrough is used to ingest, not the ModelExtractor. Will make this clearer

strands-ts/src/memory/extraction/resolve.ts
opieter-aws: That's a responsibility of the manager: https://github.com/strands-agents/harness-sdk/pull/2719/changes#diff-009358fe557a16e5dc4953099e635ce8347c46a6663157951f447dc405162769R102-R104

strands-ts/src/memory/extraction/resolve.ts
opieter-aws: Clarified the comment and doc strings throughout

notowen333 (dismissed):

notowen333 (approved):

PR #2721, opened 2026-06-10
feat: pass sequenceNumber to memory store addMessages method

opieter-aws:
## Description

Additive backward-compatible extension point for the memory subsystem, requested by the AgentCore Memory team while building a store on top of `MemoryManager`. Per-message `sequenceNumbers` on `AddMessagesContext` to enable deduping.

Extraction delivery is at-least-once: when an `addMessages` write fails, the coordinator rolls its mark back and re-fires the same batch (the documented "stores should expect duplicates" path). A backend that dedups on a client-supplied token (e.g. AgentCore's `createEvent` `clientToken`) needs each logical message to produce the *same* token on the original send and every re-send. Content hashing breaks this when two distinct messages share text ("ok", "yes"); positional batch index breaks it because filtering and retries shift positions.

The coordinator already assigns each buffered message a stable `seq` at record time that survives re-fire. We now thread those numbers through to the store:

```typescript
addMessages?(messages: MessageData[], context?: AddMessagesContext): Promise<unknown>

interface AddMessagesContext {
  // sequenceNumbers[i] identifies messages[i]; a retried batch reuses the same numbers.
  readonly sequenceNumbers?: readonly number[]
}
```

`sequenceNumbers` is index-aligned with the filtered batch the store actually receives, so content filtering that drops a message leaves a gap (a batch may carry `[0, 2]`) rather than renumbering. Numbers are stable only within one agent run and reset to 0 across runs, so a durable dedup token must combine one with a run-unique id (e.g. `${sessionId}:${seq}`). Populated only on the no-extractor passthrough; the extractor route is unchanged.

## Related Issues

N/A

## Documentation PR

No new docs required — both are additive extension points documented inline; no public surface a user follows a guide to reach.

## Type of Change

New feature

## Testing

Unit tests

This is the TypeScript SDK (`strands-ts/`); `hatch run prepare` is the Python workflow and does not apply. Validated with the TS build, the memory + Bedrock KB unit suites, and lint.

## Checklist

----

### DIFF
--- a/strands-ts/src/memory/extraction/coordinator.ts
+++ b/strands-ts/src/memory/extraction/coordinator.ts
@@ -24,17 +24,23 @@ function _blockKind(block: ContentBlockData): string {
   return Object.keys(block)[0] ?? ''
 }
 
-/** Removes excluded content blocks, and drops any message left empty. Does not mutate the input. */
-function _filterMessages(messages: MessageData[], filter: MemoryMessageFilter): MessageData[] {
+/**
+ * Removes excluded content blocks, and drops any message left empty. Does not mutate the input.
+ * Carries each message's sequence number through so it stays aligned with the filtered batch.
+ */
+function _filterMessages(buffered: BufferedMessage[], filter: MemoryMessageFilter): BufferedMessage[] {
   const exclude = new Set<string>(filter.exclude)
-  const result: MessageData[] = []
-  for (const message of messages) {
+  const result: BufferedMessage[] = []
+  for (const { seq, message } of buffered) {
     const content = message.content.filter((block) => !exclude.has(_blockKind(block)))
     if (content.length > 0) {
       result.push({
-        role: message.role,
-        content,
-        ...(message.metadata !== undefined && { metadata: message.metadata }),
+        seq,
+        message: {
+          role: message.role,
+          content,
+          ...(message.metadata !== undefined && { metadata: message.metadata }),
+        },
       })
     }
   }
@@ -173,10 +179,7 @@ export class ExtractionCoordinator {
 
     const extraction = store.extraction!
     const filter = extraction.filter ?? DEFAULT_MEMORY_MESSAGE_FILTER
-    const filtered = _filterMessages(
-      fresh.map((buffered) => buffered.message),
-      filter
-    )
+    const filtered = _filterMessages(fresh, filter)
 
     try {
       if (filtered.length > 0) {
@@ -223,8 +226,9 @@ export class ExtractionCoordinator {
    * Fact writes run in parallel. If any fails we throw, which makes the caller retry the whole batch
    * next time - so a fact that already saved may be written again (stores should expect duplicates).
    */
-  private async _write(store: MemoryStore, messages: MessageData[]): Promise<void> {
+  private async _write(store: MemoryStore, buffered: BufferedMessage[]): Promise<void> {
     const extractor = store.extraction!.extractor
+    const messages = buffered.map((buffer) => buffer.message)
 
     if (extractor) {
       const entries = await extractor.extract(messages, { defaultModel: this._defaultModel })
@@ -232,14 +236,15 @@ export class ExtractionCoordinator {
       const failures = settled.filter((r): r is PromiseRejectedResult => r.status === 'rejected')
       if (failures.length > 0) {
         throw new AggregateError(
-          failures.map((f) => f.reason),
+          failures.map((failure) => failure.reason),
           `failed to write ${failures.length} of ${entries.length} extracted entries`
         )
       }
       return
     }
 
-    await store.addMessages!(messages)
+    // Pass each message's sequence number so a store can build an idempotency key surviving retries.
+    await store.addMessages!(messages, { sequenceNumbers: buffered.map((buffer) => buffer.seq) })
   }
 
   /**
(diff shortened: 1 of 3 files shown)


strands-ts/src/memory/extraction/coordinator.ts
```diff
@@ -24,17 +24,23 @@ function _blockKind(block: ContentBlockData): string {
...
-  for (const message of messages) {
+  const result: BufferedMessage[] = []
+  for (const { seq, message } of buffered) {
     const content = message.content.filter((block) => !exclude.has(_blockKind(block)))
     if (content.length > 0) {
       result.push({
-        role: message.role,
-        content,
-        ...(message.metadata !== undefined && { metadata: message.metadata }),
+        seq,
```
notowen333, line 38: We discussed offline:

1. top level id field for messages is preferred in the future and could have other use cases
2. Could move seq into metadata but that has worse downstream consumption story than what is presented here

notowen333 (approved):

agent-of-mkmeral:
## Full review @ `6c31f70` — build, tests, lint, and adversarial probes

**Verdict: LGTM.** Clean, additive, backward-compatible. The seq-threading through `_filterMessages` is the right design (identity travels with the message, not the batch position), and the 6 new tests cover the contract well — including full-object `toEqual` assertions on the context envelope. I went looking for ways to break the stability guarantee and couldn't — but the hunt did surface a **pre-existing** coordinator bug adjacent to this PR's contract (below).

### What I verified locally
- ✅ `npm run build` (tsc) clean
- ✅ All **108** memory unit tests pass (`vitest --project unit-node src/memory`), type-check clean
- ✅ `eslint` + `prettier --check` clean on `src/memory`
- ✅ 6 adversarial tests I wrote against the seq contract — **all hold**:
  1. **Superset retry**: batch `[0,1]` fails, new message arrives, retry carries `[0,1,2]` — original messages keep seqs 0,1; no renumbering
  2. **Per-store filters**: same conversation gives `keepAll` → `[0,1,2]` and `dropTools` → `[0,2]` — seq is a *global message identity*, consistent across stores
  3. **Lagging store**: a store that fails then recovers still sees original seqs on retry while a faster store proceeds independently
  4. **No aliasing**: each `addMessages` call gets a fresh array (`buffered.map(...)` per call); a hostile store mutating its array can't corrupt another store's view
  5. **No reset after trim**: buffer fully trimmed → next message continues the global counter (no accidental reuse of seq 0)
  6. **Flush-driven retry** carries the same seqs as trigger-driven saves

### 🟡 One doc nuance worth a line
`coordinator.ts` and the PR body say a failed save "re-fires **the same batch**". Probe 1 shows a retry may actually carry a **superset** (original messages + anything buffered since), with per-message seqs unchanged. `types.ts` gets this right ("`sequenceNumbers[i]` identifies `messages[i]`"), but a dedup implementor skimming the docs could wrongly assume retry ⇒ identical batch and build a *per-batch* token (e.g. hashing the whole seq array) — which the superset breaks. Suggest one sentence on `AddMessagesContext.sequenceNumbers`: *"A retried batch may include additional newly-buffered messages; build dedup tokens per message, not per batch."* The superset case would also make a nice 7th test (happy to share mine).

### 🔴 Pre-existing (NOT introduced here): trim race silently loses messages in multi-store setups
While probing, I found that the optimistic-mark + `_trim()` interaction can permanently drop a message for a failing store. **Repro fails identically on `main`**, so it ships with #2671, not this PR — but it directly undercuts the at-least-once delivery contract this feature builds dedup on, so flagging it here.

**Sequence** (two stores, one message `m0` with seq 0):
1. Both stores' `_extract` starts; both **optimistically** advance their mark to 0 *before* their write completes
2. `fast`'s write succeeds → its `finally` runs `_trim()` → `minMark = min(fast=0, slow=0) = 0` → **`m0` is dropped from `_pending` while `slow`'s write is still in flight**
3. `slow`'s write fails → `_onSaveFailed` rolls its mark back to −1 → but the buffer no longer contains `m0`
4. Every subsequent save/flush for `slow` finds `fresh = []` — `m0` is silently lost forever for that store, despite the documented contract: *"the messages stay buffered either way, so a store that recovers saves them"*

```ts
it('m0 is lost for slow when fast succeeds while slow is mid-failing-save', async () => {
  const fast = createStore('fast', { trigger: [new InvocationTrigger()] })
  const slow = createStore('slow', { trigger: [new InvocationTrigger()] })
  let rejectSlow!: (e: Error) => void
  slow.addMessages.mockImplementationOnce(() => new Promise((_r, rej) => { rejectSlow = rej }))
  const mm = new MemoryManager({ stores: [fast, slow] })
  // ... record m0, fire trigger, let fast complete (await setTimeout 10) ...
  rejectSlow(new Error('transient outage'))
  await mm.flush() // contract: recovered store saves its backlog
  // ... next turn records m1; slow receives only ['second fact'] — m0 gone
})
```

Single-store setups are immune (the catch rolls the mark back *before* that store's own `finally` trim). The trigger window is `(other store succeeds) ∧ (this store's overlapped save fails)` — very reachable in production since all stores fire off the same trigger simultaneously and the failure mode is a slow-then-failing backend.

**Suggested fix (follow-up PR, not this one):** trim on *confirmed* marks only — keep a second `_confirmed` map updated only on write success, and have `_trim()` read that instead of the optimistic `_marks`. The optimistic map keeps preventing double-pickup by queued saves; the confirmed map keeps the buffer honest.

Given the AgentCore Memory team is building dedup on exactly this delivery guarantee, I'd track this — happy to open the issue + fix PR with the repro test if useful.

### Nits (take or leave)
- `buffered.map((buffer) => buffer.seq)` — naming an *element* of `buffered` as `buffer` reads oddly (a buffer is usually the container). Destructuring `.map(({ seq }) => seq)` is shorter and self-documenting.

No blocking issues on this PR itself — the feature does exactly what it says, and what it says is what dedup implementors need.

PR #2722, opened 2026-06-10
fix: improve devx for instantiating multiple bedrock KBs

opieter-aws:
## Description

Running one Bedrock Knowledge Base across multiple namespaces is a first-class use case (per-user, per-team, per-org memory partitions backed by the same KB). With the constructor shipped in #2630, doing so meant retyping the entire connection config for every namespace, with only `scope` differing:

```typescript
const personal = new BedrockKnowledgeBaseStore({ name: 'personal', knowledgeBaseId: 'KB123', dataSourceType: 'CUSTOM', dataSourceId: 'DS456', writable: true, scope: 'user-abc' })
const team     = new BedrockKnowledgeBaseStore({ name: 'team',     knowledgeBaseId: 'KB123', dataSourceType: 'CUSTOM', dataSourceId: 'DS456', writable: true, scope: 'team-xyz' })
```

This PR restructures the constructor into a single nested shape that cleanly separates the **reusable connection** from the **per-store identity**. Build the connection once, then spin up one store per namespace:

```typescript
const config = { knowledgeBaseId: 'KB123', dataSourceType: 'CUSTOM', dataSourceId: 'DS456' }

const personal = new BedrockKnowledgeBaseStore({ config, name: 'personal', scope: 'user-abc', writable: true })
const team     = new BedrockKnowledgeBaseStore({ config, name: 'team',     scope: 'team-xyz', writable: true })
new MemoryManager({ stores: [personal, team] })
```

This is a DX-only change — search/ingest/validation behavior is unchanged. It only restructures the constructor input.

**Key design decisions** (reviewer-relevant):

- **One shape, always nested — no flat form, no factory, no builder.** A factory method/class and a `shared`-config merge were both considered and rejected: the store should have no notion of "sharing," and a configured store is not the place to encode a caller-side reuse pattern. Reusing `config` across stores is just the caller passing the same object.
- **Clean split, every field in exactly one place.** Connection settings (`knowledgeBaseId`, `dataSourceType`, `dataSourceId`, `s3`, `scopeMetadataKey`, `runtimeClient`, `agentClient`) live inside the new `BedrockKnowledgeBaseConfig`; per-store identity/behavior (`name`, `scope`, `description`, `writable`, `maxSearchResults`, `filter`) sit beside it. No field is settable in two places, so there is no override-precedence rule or excess-property edge case to reason about.
- **`name` is required and independent of `scope`.** An earlier iteration defaulted `name` to `scope`; this was dropped in favor of keeping them distinct — `name` is the routing identity surfaced to the model, `scope` is the data partition. Requiring `name` lets the store config simply `extends MemoryStoreConfig` (which already provides `name`, `description`, `maxSearchResults`, `writable`) instead of carrying an `Omit<>` workaround, and removes a runtime guard.
- **Per-namespace state is immutable.** `scope`, `scopeMetadataKey`, and `filter` are `readonly` — to target a different namespace, construct another store from the same `config` rather than mutating one in place.
- **Client reuse falls out for free.** Inject `runtimeClient` / `agentClient` into `config` and every store built from it reuses them; omit them and each store lazily builds its own default, exactly as before.

## Related Issues

Builds on #2630 (adds `BedrockKnowledgeBaseStore`) and should land after it. Follow-up to #2544.

## Documentation PR

No docs change in this PR. Note: the pre-existing examples in `designs/0011-memory-manager.md` use the original flat constructor form (and reference an `extraction` field that does not exist on the store); they predate this change and #2630, and are left for a separate docs reconciliation.

## Type of Change

Breaking change

This restructures the public constructor of `BedrockKnowledgeBaseStore`. Because the store is introduced by #2630 and is not yet on `main`, this is a pre-release API change with no released consumers to migrate.

### Public API Changes

The constructor now takes a nested `config` plus per-store fields, and a new `BedrockKnowledgeBaseConfig` type is exported:

```typescript
// Before (#2630): flat
new BedrockKnowledgeBaseStore({
  name: 'personal',
  knowledgeBaseId: 'KB123',
  dataSourceType: 'CUSTOM',
  dataSourceId: 'DS456',
  writable: true,
  scope: 'user-abc',
})

// After: connection nested under `config`, identity beside it
new BedrockKnowledgeBaseStore({
  config: { knowledgeBaseId: 'KB123', dataSourceType: 'CUSTOM', dataSourceId: 'DS456' },
  name: 'personal',
  scope: 'user-abc',
  writable: true,
})
```

`name` is now required. `scope`, `scopeMetadataKey`, and `filter` are exposed as `readonly` instance fields — to target a different namespace, construct another store from the same `config` rather than mutating one in place.

## Testing

TypeScript-only change, verified with the strands-ts gates: `npm run build`, `npm run test` (unit, including a `@ts-expect-error` assertion that `name` is required and a `config`-reuse suite covering client reuse, per-store defaults, and `MemoryManager` registration/dedup), `npm run type-check` (src + integ), `npm run lint`, `npm run format:check` — all green. All existing unit tests and the 15 integ-test constructions were migrated to the nested shape with no behavior assertions lost.

## Checklist

----

### DIFF
--- a/strands-ts/src/vended-memory-stores/bedrock-knowledge-base/__tests__/bedrock-knowledge-base-store.test.ts
+++ b/strands-ts/src/vended-memory-stores/bedrock-knowledge-base/__tests__/bedrock-knowledge-base-store.test.ts
@@ -660,6 +660,133 @@ describe('BedrockKnowledgeBaseStore', () => {
     })
   })
 
+  describe('dynamic scope mutability', () => {
+    /** A search store whose `send` echoes the RetrieveCommand and resolves to empty results. */
+    function searchStore(overrides: Record<string, unknown> = {}): {
+      store: BedrockKnowledgeBaseStore
+      runtime: { send: MockedFunction<any> }
+    } {
+      const runtime = mockClient()
+      runtime.send.mockResolvedValue({ retrievalResults: [] })
+      const store = new BedrockKnowledgeBaseStore({
+        name: 'kb',
+        knowledgeBaseId: 'kb-1',
+        runtimeClient: runtime as any,
+        ...overrides,
+      })
+      return { store, runtime }
+    }
+
+    /** Reads the filter the most recent RetrieveCommand was sent with. */
+    function lastSearchFilter(runtime: { send: MockedFunction<any> }): unknown {
+      const calls = runtime.send.mock.calls
+      return calls[calls.length - 1]?.[0].input.retrievalConfiguration.vectorSearchConfiguration.filter
+    }
+
+    /** A writable CUSTOM store whose agent `send` echoes the ingestion command. */
+    function customStore(overrides: Record<string, unknown> = {}): {
+      store: BedrockKnowledgeBaseStore
+      agent: { send: MockedFunction<any> }
+    } {
+      const agent = mockClient()
+      agent.send.mockResolvedValue({})
+      const store = new BedrockKnowledgeBaseStore({
+        name: 'kb',
+        knowledgeBaseId: 'kb-1',
+        writable: true,
+        dataSourceType: 'CUSTOM',
+        dataSourceId: 'ds-1',
+        agentClient: agent as any,
+        ...overrides,
+      })
+      return { store, agent }
+    }
+
+    /** Reads the inline attributes the most recent CUSTOM ingestion document carried. */
+    function lastInlineAttributes(agent: { send: MockedFunction<any> }): any[] {
+      const calls = agent.send.mock.calls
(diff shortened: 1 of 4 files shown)


notowen333 (approved):
Maybe in the same PR we can add some nuance to where nested constructor configs is valid. Can also follow up with it.

opieter-aws:
> Maybe in the same PR we can add some nuance to where nested constructor configs is valid. Can also follow up with it.

@notowen333 do you mean adding a code example like in the description to a docstring?

PR #2725, opened 2026-06-11
fix(issue-labeler): stop over-applying area-community and chore

yonib05:
## Summary

Tuning fixes after observing mislabels on recent PRs (e.g. a `feat:` PR labeled `chore`, and `area-community` landing on nearly everything).

## Problems

- **`area-community` over-applied.** Its description read as a catch-all, so PRs that fit no specific area defaulted to it.
- **`chore` on feature PRs.** `type.yml` ignored the conventional-commit prefix; a PR titled `feat: ...` was labeled `chore`.
- **Too many area labels.** The area job used the default cap of 3; most issues span at most 2 areas.
- **Same catch-all risk elsewhere.** Reviewing the rest of `area.yml`, `area-devx` ("ergonomics, usability") and `area-agent` ("core agent class") had the same magnet phrasing.

## Fix

- Narrow `area-community` and instruct the classifier to assign no area label when none clearly applies.
- Add a guard so `area-agent` requires the core agent class rather than any mention of "agent".
- Clarify `area-devx` as the developer-experience axis: papercuts, confusing or awkward public APIs, error messages, and surprising behavior that doesn't match expectations. It is appropriate even for a specific subsystem when the issue is about that subsystem's public API being confusing or surprising to use (as opposed to a clear functional bug in its internals), but it is not a catch-all for generic "make it better" requests.
- Make the conventional-commit prefix authoritative in `type.yml`: match prefixes regardless of scope, treat the prefix as decisive over the user-facing heuristic, and judge from title + body (the labeler has no diff visibility).
- Set `max_labels: '2'` on the area job.

## Testing

- [ ] After merge, confirm a `feat:` PR receives `enhancement` (not `chore`) and that unrelated PRs no longer get `area-community` by default.

### DIFF
--- a/.github/labelers/area.yml
+++ b/.github/labelers/area.yml
@@ -14,6 +14,8 @@ instructions: |
   General agent class behavior (invocation, streaming, result handling) belongs to area-agent. Do not assign area-agent merely because an issue mentions "agent"; require evidence it concerns the core agent class itself.
   area-community is NOT a catch-all. Use it only for repo health, governance, contributor process, release process, and CI dependency bumps. If no area clearly applies, assign no area label rather than defaulting to area-community.
   area-devx covers how it feels to use the SDK: papercuts, confusing or awkward public APIs, error-message quality, ergonomics, surprising behavior that does not match what a developer would reasonably expect, and usability of the developer-facing interface. It is appropriate even for a specific subsystem when the issue is about that subsystem's public API being confusing, awkward, or surprising to use (as opposed to a clear functional bug in its internal behavior). It is NOT a catch-all for any generic "make it better" request that lacks a concrete usability concern.
+  Order matters: list the most important label first, because at most 2 labels are kept. Rank by specificity — concrete feature areas (area-mcp, area-model, area-agent, area-persistence, etc.) come before softer cross-cutting labels (area-devx, area-community) that are not tied to one feature area. When an issue reports a bug in a concrete area, that area takes priority.
+  Example: "the MCPClient constructor arguments are confusing and the error when a server is unreachable is unhelpful" -> area-mcp, then area-devx (concrete area first, usability second). But "MCPClient drops messages under concurrent calls" -> area-mcp only (functional bug in internals; no devx concern).
 
 labels:
   area-a2a:
--- a/.github/labelers/area.yml
+++ b/.github/labelers/area.yml
@@ -11,7 +11,9 @@ instructions: |
   Documentation content issues belong to documentation, not area-devx.
   Storage backend implementation issues (S3, DynamoDB, file) belong to area-persistence. Session lifecycle and interface issues belong to area-sessions.
   Context window/token management belongs to area-context.
-  General agent class behavior (invocation, streaming, result handling) belongs to area-agent.
+  General agent class behavior (invocation, streaming, result handling) belongs to area-agent. Do not assign area-agent merely because an issue mentions "agent"; require evidence it concerns the core agent class itself.
+  area-community is NOT a catch-all. Use it only for repo health, governance, contributor process, release process, and CI dependency bumps. If no area clearly applies, assign no area label rather than defaulting to area-community.
+  area-devx covers how it feels to use the SDK: papercuts, confusing or awkward public APIs, error-message quality, ergonomics, surprising behavior that does not match what a developer would reasonably expect, and usability of the developer-facing interface. It is appropriate even for a specific subsystem when the issue is about that subsystem's public API being confusing, awkward, or surprising to use (as opposed to a clear functional bug in its internal behavior). It is NOT a catch-all for any generic "make it better" request that lacks a concrete usability concern.
 
 labels:
   area-a2a:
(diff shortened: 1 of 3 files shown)


yonib05:
Thanks for the thorough review. Addressed the `type.yml` issues in a follow-up commit:

- **Scoped prefixes**: prefix matching is now scope-agnostic, so `refactor(core):` / `ci(deps):` match the same as the bare form.
- **Rule precedence**: the conventional-commit prefix is now explicitly authoritative — a `perf:` or `refactor:` title resolves to `chore` even if the description sounds user-facing, removing the contradiction. The body-based heuristic only applies when no prefix is present.
- **Input grounding**: good catch — the labeler only receives title + body (no diff). Rewrote the guidance to judge from title + body and dropped the language implying file/diff inspection.

On **area-devx**: the description mismatch is real but it's the PR *description* that's now stale, not the code. After discussion we intentionally broadened `area-devx` to cover usability of a specific subsystem's public API (papercuts, confusing APIs, surprising behavior). The code reflects the intended behavior; I'll update the PR description to match rather than revert the wording.

On **backtesting**: agreed it would de-risk. We've effectively done this — the configs were validated against the existing open issues/PRs during the evals rollout, and mislabels found there fed back into these rules.

.github/labelers/type.yml
```diff
@@ -3,9 +3,14 @@
 # Use max_labels: 1 in the workflow since these are mutually exclusive.
 
 instructions: |
-  Choose exactly one type. If the title starts with [BUG] it is a bug.
-  If the title starts with [FEATURE] it is an enhancement.
-  CI/dependency PRs that open as issues are chores.
+  Choose exactly one type. You are given only the title and body — there is no diff or file list, so judge from those.
```
opieter-aws, line 6: Would it be an idea to also pass the diff, so the agent can do a better analysis?

opieter-aws (dismissed):

.github/labelers/area.yml
```diff
@@ -23,13 +25,13 @@ labels:
   area-bidirectional-streaming:
     description: "Bidirectional streaming, BidiAgent, real-time audio/video, WebRTC, WebSocket, Nova Sonic, voice agents"
   area-community:
-    description: "Community health, monorepo consolidation, contributor guides, governance, CI dependency bumps, release process"
+    description: "Repo health, governance, contributor guides, release process, and CI dependency bumps. Not a fallback for issues that fit no other area."
```
zastrowm, line 30: Any reason not to include this info in the labels themselves (they have descriptions) and have them pulled in dynamically instead of in the workflow?

yonib05: Label description maximum is 100 characters. Also, not a fallback for issues that fit no other area doesn't feel appropriate for it, and if we are going to start dynamically pulling + adding modifications, it feels more confusing and messy. 

But I can update the label descriptions to line up (where it fits) for hygiene.

.github/labelers/type.yml
yonib05: Too much context, IMO; the point of this is lightweight classification, not deterministic. Passing in the diff seems like overengineering for something geared at lightweight classification. This feature is more for metrics and rough sorting than a precision system guaranteeing quality. That being said, I will look into it in a future iteration.

.github/labelers/area.yml
```diff
@@ -11,7 +11,9 @@ instructions: |
...
   Context window/token management belongs to area-context.
-  General agent class behavior (invocation, streaming, result handling) belongs to area-agent.
+  General agent class behavior (invocation, streaming, result handling) belongs to area-agent. Do not assign area-agent merely because an issue mentions "agent"; require evidence it concerns the core agent class itself.
+  area-community is NOT a catch-all. Use it only for repo health, governance, contributor process, release process, and CI dependency bumps. If no area clearly applies, assign no area label rather than defaulting to area-community.
+  area-devx covers how it feels to use the SDK: papercuts, confusing or awkward public APIs, error-message quality, ergonomics, surprising behavior that does not match what a developer would reasonably expect, and usability of the developer-facing interface. It is appropriate even for a specific subsystem when the issue is about that subsystem's public API being confusing, awkward, or surprising to use (as opposed to a clear functional bug in its internal behavior). It is NOT a catch-all for any generic "make it better" request that lacks a concrete usability concern.
```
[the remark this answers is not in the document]
yonib05: Both points addressed:

1. **Description mismatch** — updated the PR description so it matches the broadened intent: area-devx is a cross-cutting usability axis that can pair with a subsystem label, rather than being excluded from subsystems.

2. **Anchoring example + co-labeling** — added the worked example to `area.yml` and, importantly, an explicit **priority-ordering rule**. Since `classify.py` truncates the model's output to the first `max_labels` entries, order is load-bearing: the prompt now tells the classifier to rank concrete feature areas (area-mcp, area-model, etc.) ahead of softer cross-cutting labels (area-devx, area-community), and bugs in a concrete area take priority. So your `area-mcp` + `area-devx` example now resolves with `area-mcp` first by design.

zastrowm (approved):

opieter-aws (approved):

PR #2726, opened 2026-06-11
fix(dependabot): label npm updates typescript instead of javascript

yonib05:
## Summary

Dependabot applied its default `javascript` ecosystem label to npm dependency PRs. Since the repo's npm packages (the TypeScript SDK at `/` and the docs site at `/site`) are TypeScript, this drew a complaint.

## Fix

Set explicit `labels:` on both npm entries so updates get `dependencies` + `typescript` instead of the default `javascript`:

```yaml
labels:
  - "dependencies"
  - "typescript"
```

`dependencies` is included so that label is preserved (explicit labels replace dependabot's defaults).

## Follow-up (not in this PR)

Existing open `javascript`-labeled PRs will be migrated to `typescript` and the `javascript` repo label removed separately.

## Testing

- [ ] Confirm the next dependabot npm PR opens with `dependencies` + `typescript` and no `javascript`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,9 @@ updates:
     schedule:
       interval: "daily"
     open-pull-requests-limit: 100
+    labels:
+      - "dependencies"
+      - "python"
     commit-message:
       prefix: "ci(python)"
     groups:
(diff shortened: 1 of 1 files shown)


notowen333 (dismissed):

yonib05:
Thanks — and good call on the language-parity suggestion. Rather than leave it as a follow-up, I added it to this PR: the two `pip` entries (`/strands-py`, `/strands-py-wasm`) now set `labels: ["dependencies", "python"]`, matching what the npm entries do with `typescript`. So all dependabot PRs now get a language label.

opieter-aws (approved):

PR #2738, opened 2026-06-11
feat(strands-py): add GoalLoop vended plugin with docs

notowen333:
## Description

Agents often need iterative refinement — retry until the response meets a quality bar. Today that means hand-rolling a loop with timeout logic, attempt tracking, feedback injection, and state management. GoalLoop encapsulates all of that as a vended plugin that works inside the existing hook lifecycle.

This PR ports the GoalLoop plugin from TypeScript to Python and adds a dual-language documentation page.

## Public API Changes

New module: `strands.vended_plugins.goal`

```python
from strands import Agent
from strands.vended_plugins.goal import GoalLoop

# Natural-language goal — judged by an internal agent built from the host's model
concise = GoalLoop(
    goal="At most 3 sentences, accessible to a 10-year-old, no jargon.",
    max_attempts=3,
)

agent = Agent(plugins=[concise])
agent("Explain how rainbows form.")
print(concise.last_result(agent))
# GoalResult(passed=True, stop_reason='satisfied', attempts=[...])
```

```python
# Programmatic validator — pass a callable to skip the judge agent entirely
from strands.vended_plugins.goal import GoalLoop

def word_count_validator(response, agent):
    text = " ".join(
        block["text"] for block in response["content"] if "text" in block
    )
    words = len(text.split())
    if words <= 50:
        return True
    return {"passed": False, "feedback": f"Too long ({words} words). Cap at 50."}

plugin = GoalLoop(goal=word_count_validator, max_attempts=5, timeout=30.0)
```

### Exported symbols

| Symbol | Kind | Purpose |
|--------|------|---------|
| `GoalLoop` | Plugin class | Main entry point — attach to an agent via `plugins=[...]` |
| `GoalResult` | Dataclass | Aggregate result with `passed`, `stop_reason`, `attempts` |
| `GoalAttempt` | Dataclass | Per-attempt record: `attempt`, `passed`, `feedback` |
| `GoalStopReason` | Literal type | `"satisfied"` \| `"max_attempts"` \| `"timeout"` |
| `JudgeConfig` | Dataclass | Optional judge tuning: `model`, `system_prompt` |
| `ValidationOutcome` | Dataclass | Canonical validator return: `passed`, `feedback` |
| `Validator` | Protocol | Type for programmatic validator callables |
| `JUDGE_SYSTEM_PROMPT` | str | Default system prompt for the NL judge |
| `JudgeOutcome` | Pydantic model | Structured output schema the judge fills |
| `build_judge_prompt` | Function | Builds the judge input from a goal + transcript |

### GoalLoop constructor parameters

| Parameter | Default | Description |
|-----------|---------|-------------|
| `goal` | *(required)* | NL string (judged by internal agent) or callable validator |
| `max_attempts` | `inf` | Maximum attempts before stopping |
| `timeout` | `inf` | Wall-clock budget in seconds |
| `judge` | `None` | `JudgeConfig` to override the judge model or system prompt |
| `preserve_context` | `True` | Keep conversation history across retries |
| `resume_prompt_template` | *(built-in)* | `Callable[[str | None], str | list[ContentBlock]]` for the retry message |
| `name` | `"strands:goal-loop"` | Plugin name (must be unique per agent) |

## Related Issues

N/A - new feature port

## Documentation PR

Included in this PR under `site/src/content/docs/user-guide/concepts/plugins/goal-loop.mdx` with dual-language tabs (Python + TypeScript).

## Type of Change

New feature

## Testing

## Checklist

----


notowen333:
@agent-of-mkmeral can you review that this is an honest port of the TS GoalLoop?

agent-of-mkmeral (commented):
Reviewed for port fidelity against `strands-ts/src/vended-plugins/goal/` (plugin.ts, judge.ts, index.ts + tests), at head `4b9b6a6`. **Verdict: yes, this is an honest port** — the control flow, state machine, and edge-case semantics match the TS original line-for-line in every place I checked, with one substantive fidelity gap in Ralph mode (below). All 35 unit tests pass locally.

## ✅ Verified faithful (checked each against the TS source, not just the PR description)

- **Run lifecycle state machine** — `BeforeInvocation` fresh-run-vs-resumed branching (`resumed` flag consumed exactly once), `_finish_run` clearing `resumed`, `last_result` returning `None` mid-flight and after a thrown invoke. Identical to TS.
- **After-hook ordering** — timeout checked *before* validation (so an expensive judge can't blow the budget), then `_last_assistant_message` bail-out, then validate → record attempt → pass/max-attempts/arm-resume. Same order as TS.
- **Validator semantics** — bool shorthand, dict/`ValidationOutcome` forms, sync+async via `inspect.isawaitable` (mirrors TS's single `await`-friendly path), validator throws logged via `logger.warning` and converted to a failed attempt with `Validator error: ...` feedback (exact TS behavior, including not crashing the loop).
- **Judge path** — fresh judge `Agent` per validation (no prompt leakage between judgements), `callback_handler=None` ↔ `printer: false`, host-model default with `judge.model` override, fallback `"Judge produced no structured outcome."` on missing structured output. `JUDGE_SYSTEM_PROMPT` is verbatim; `build_judge_prompt` reproduces the role-tagged transcript, inline `[tool-call]`/`[tool-result]` summaries, and the 500-char truncation with the same `… [N more chars]` suffix.
- **Resume prompt** — default English verbatim; feedback trimmed (`?.trim()` ↔ `strip() if feedback else None` — equivalent for all input shapes including whitespace-only feedback).
- **Guards** — one-GoalLoop-per-agent via module-level `WeakSet` (TS: `agentsWithGoalLoop`), shared-instance-across-agents via `WeakKeyDictionary` (TS: `WeakMap`), unbounded-budget warning, `max_attempts >= 1` validation.
- **Idiom conversions** are all sound and an improvement in one case: `time.monotonic()` is strictly better than TS's `Date.now()` for elapsed-time math. Timeout units (s vs ms) are clearly documented in both the docstrings and the config table.
- **Integ tests** assert only validator-controlled structure (attempt counts, transcript shape), never model output — deterministic by construction. Nice.

## ⚠️ One real fidelity gap: Ralph mode doesn't rewind the system prompt in Python

The TS `session` snapshot preset is `['messages', 'state', 'systemPrompt', 'modelState', 'interrupts']`, and the TS plugin's whole stated rationale for the snapshot is:

> Everything else in the `session` preset (messages, **systemPrompt**, modelState, interrupts) is what the model sees … so **all of it must rewind** for Ralph mode to actually hide prior attempts from the model.

But the **Python `session` preset is different**: `('messages', 'state', 'conversation_manager_state', 'interrupt_state', 'model_state')` — **no `system_prompt`**. So `take_snapshot(preset="session", exclude=["state"])` in Python captures/restores messages, conversation-manager state, interrupt state, and model state, but a system prompt mutated between attempts (e.g. by another plugin's hook) will *not* be rewound. Verified empirically:

```
>>> resolve_snapshot_fields(preset='session', exclude=['state'])
['conversation_manager_state', 'interrupt_state', 'messages', 'model_state']
```

Two consequences:

1. **Behavioral divergence from TS** in exactly the scenario the TS comments call out. Fix is one line — `load_snapshot` already restores `system_prompt` when present:
   ```python
   run.initial_snapshot = event.agent.take_snapshot(
       preset="session", include=["system_prompt"], exclude=["state"]
   )
   ```
2. **The docs page is currently inaccurate for Python.** `goal-loop.mdx` (Stateless Retries) says the snapshot restores "messages, **system prompt**, model state" — true for TS, false for Python as written. Either apply fix #1 (preferred, restores parity and makes the docs true) or qualify the docs.

(Worth noting the asymmetry cuts both ways: Python's preset additionally rewinds `conversation_manager_state`, which has no TS equivalent — that part is a sensible Python-only bonus and worth keeping.)

## Minor (non-blocking)

- **Judge-path unit-test gap**: TS has 6 mocked-model tests for the NL judge (`judges via host model`, `feeds judge feedback back`, `fresh judge per validation`, `judge.model override`, `judge.systemPrompt override`, plain-text-then-structured-retry). The Python suite tests the judge *helpers* (prompt building, schema) but never exercises `_judge_validator` itself — the judge-Agent construction, `invoke_async`, and structured-output parsing are untested. A couple of tests with a mocked `Agent`/model would close this.
- `GoalStopReason = str` loses the TS literal union. `Literal["satisfied", "max_attempts", "timeout"]` would catch typos in caller comparisons at type-check time for free.
- `import inspect` inside `_fn_validator` runs on every validation; move it to module level.
- The `.agents/skills/docs-*` changes bundled in here are fine but unrelated to the port — no objection, just noting they ride along.

## Summary

Everything observable in the standard (preserve-context) path is a faithful port — I could not find a single behavioral divergence there. The one thing I'd fix before un-drafting is the Ralph-mode snapshot missing `system_prompt` (one-line fix + keeps the docs honest).

notowen333:
@agent-of-mkmeral can you re-check the port after the most recent commit

agent-of-mkmeral:
## Re-check at head `7d855d0` (covers `930c808` + `7d855d0`)

**Verdict: the port is still faithful, and all four items from my original review are resolved.** Re-verified the changed surface against the TS source and ran the suite — **41/41 unit tests pass locally**.

### ✅ The one real fidelity gap is fixed (Ralph-mode `system_prompt` rewind)

```python
run.initial_snapshot = event.agent.take_snapshot(
    preset="session", include=["system_prompt"], exclude=["state"]
)
```

Verified empirically against the snapshot resolver:

```
before fix: ['conversation_manager_state', 'interrupt_state', 'messages', 'model_state']
after fix:  ['conversation_manager_state', 'interrupt_state', 'messages', 'model_state', 'system_prompt']
```

Python now rewinds everything the TS `session` preset rewinds (plus `conversation_manager_state`, which is documented inline as an intentional Python-only divergence — exactly what I'd hoped for). The test at `test_snapshot_taken_on_first_model_call` pins the exact call signature, so this can't silently regress. And the docs claim in `goal-loop.mdx` ("messages, system prompt, model state") is **now accurate for Python** — no doc change needed.

### ✅ Minor items from my review, all addressed

- **Judge-path unit tests** — the previously untested `_judge_validator` now has 6 mocked tests (`test_nl_judge_*`: first-attempt pass, feedback loop, `judge.model` override, `judge.system_prompt` override, no-structured-output fallback, fresh-agent-per-validation), matching the TS suite's coverage 1:1. The patch target (`strands.agent.agent.Agent`) correctly intercepts the plugin's lazy import.
- **`GoalStopReason`** — now `Literal["satisfied", "max_attempts", "timeout"]`.
- **`import inspect`** — moved to module level.
- Bonus: `judge.py` rendering helpers got real types (`Message`/`ContentBlock`/`ToolResultContent` instead of `dict`) — behavior unchanged, port fidelity unaffected.

### Port fidelity of the new changes themselves

The `Validator` → `@runtime_checkable Protocol` conversion preserves the exact call shape (`(response, agent)` positionally, return `bool | dict | ValidationOutcome`, sync or async), so all TS-equivalent behavior in `_fn_validator` is untouched. The structured-logging reformat changes only the log string, not semantics.

### Remaining (non-blocking)

- `resume_prompt_template` is still a frozen `Callable` — same forward-compat argument as `Validator`, worth doing while the API is new (the `/strands review` bot flagged this too).
- CI note: all Python unit-test jobs are green across 3.10–3.14 / linux+windows (the CANCELLED entries are superseded runs); the `label-size` check failure is labeler noise unrelated to this change.

Good to go from a port-fidelity standpoint. 🚢

.agents/skills/docs-reviewer/SKILL.md
zastrowm, line 1: Is our reviewer agent aware of these skills?  It seems like it'd be great to start consuming them

site/src/content/docs/user-guide/concepts/plugins/goal-loop.mdx
```diff
@@ -0,0 +1,347 @@
...
+```mermaid
+flowchart LR
+    A[invoke] --> B[Agent responds]
+    B --> C{Validate}
+    C -->|pass| D[Done]
+    C -->|fail| E[Inject feedback]
+    E --> B
+```
+
+1. `BeforeInvocationEvent` starts a fresh run (or continues an in-progress one).
```
zastrowm: IMO the hook references don't help here; that's an implementation detail.  I think explaining it abstractly (like you do in the problem statement) works better.  If we really want this section I'd argue it belongs later

.agents/skills/docs-audit/SKILL.md
zastrowm, line 1: Can you update the PR description with usage examples of the public api

strands-py/src/strands/vended_plugins/goal/judge.py
```diff
@@ -0,0 +1,129 @@
...
+from __future__ import annotations
+
+import json
+
+from pydantic import BaseModel, Field
+
+from ...types.content import ContentBlock, Message, Messages
+from ...types.tools import ToolResultContent
+
+JUDGE_SYSTEM_PROMPT = (
```
zastrowm: This would be more readable as a triple quoted string

strands-py/src/strands/vended_plugins/goal/judge.py
```diff
@@ -0,0 +1,129 @@
...
+    "# Goal Evaluation\n"
```
zastrowm: Same prompt as TS?

strands-py/src/strands/vended_plugins/goal/plugin.py
```diff
@@ -0,0 +1,401 @@
...
+class GoalAttempt:
+    """Single attempt summary preserved on GoalResult."""
+
+    attempt: int
+    passed: bool
+    feedback: str | None = None
+
+
+@dataclass
+class GoalResult:
```
zastrowm, line 111: These need to be public?

site/src/config/navigation.yml
```diff
@@ -88,6 +88,7 @@ sidebar:
               - docs/user-guide/concepts/plugins/skills
               - docs/user-guide/concepts/plugins/steering
               - docs/user-guide/concepts/plugins/context-offloader
+              - docs/user-guide/concepts/plugins/goal-loop
```
zastrowm, line 91: Are there any other pages that should reference this one?  like steering and goal-loop semi related?

site/src/content/docs/user-guide/concepts/plugins/goal-loop.ts
```diff
@@ -0,0 +1,205 @@
...
+  // --8<-- [end:inspecting_results]
+}
+
+// =====================
+// Word Count Validator
+// =====================
+
+{
+  // --8<-- [start:word_count_validator]
+  const plugin = new GoalLoop({
```
zastrowm, line 78: A separate function (like the python) is more readable TBH.  This blurs the line between plugin creation and what the validator is doing

site/src/content/docs/user-guide/concepts/plugins/goal-loop.mdx
```diff
@@ -0,0 +1,347 @@
...
+
+### Inspecting Results
+
+After an invocation completes, retrieve the result from the plugin to get the full attempt history:
+
+<Tabs>
+<Tab label="Python">
+
+```python
+result = concise.last_result(agent)
```
zastrowm: At this point in the docs, `concise` makes no sense here (for both TS & Python).  The initial example uses `concise` but the rest use `plugin`; we should normalize on one

site/src/content/docs/user-guide/concepts/plugins/goal-loop.mdx
```diff
@@ -0,0 +1,347 @@
...
+</Tab>
+<Tab label="TypeScript">
+
+```typescript
+--8<-- "user-guide/concepts/plugins/goal-loop.ts:preserve_context_false"
+```
+</Tab>
+</Tabs>
+
+The snapshot excludes agent state (`state` / `appState`) deliberately — other plugins (rate limiters, cost trackers) rely on their mutations persisting across attempts.
```
zastrowm: Somebody recently added a new syntax to help out with this kind of differentiation 😆

site/src/content/docs/user-guide/concepts/plugins/goal-loop.ts
```diff
@@ -0,0 +1,205 @@
...
+// =====================
+
+{
+  // --8<-- [start:custom_resume_prompt]
+  const plugin = new GoalLoop({
+    goal: '...',
+    maxAttempts: 3,
+    resumePromptTemplate: (feedback) => {
+      if (!feedback) {
+        return 'Tu respuesta no cumple el objetivo. '
```
zastrowm: Python this makes sense.  For TypeScript without the context of "spanish prompt" this was very confusing as to why it was spanish

strands-py/src/strands/vended_plugins/goal/plugin.py
```diff
@@ -0,0 +1,401 @@
...
+            self._validator = goal
+
+        judge_config = judge or JudgeConfig()
+        self._judge_model = judge_config.model
+        self._judge_system_prompt = judge_config.system_prompt
+        self._max_attempts = max_attempts
+        self._timeout = timeout
+        self._preserve_context = preserve_context
+        self._resume_prompt_template = resume_prompt_template or _default_resume_prompt
+        self._runs: weakref.WeakKeyDictionary[Agent, _RunState] = weakref.WeakKeyDictionary()
```
zastrowm, line 268: This could use a code comment explaining why we're doing this

strands-py/src/strands/vended_plugins/goal/plugin.py
```diff
@@ -0,0 +1,401 @@
...
+        constructed, or if a run is still in-flight.
+        """
+        run = self._runs.get(agent)
+        if run is None:
+            return None
+        return run.result
+
+    def init_agent(self, agent: Agent) -> None:
+        """Register hooks on the agent. Called by the plugin registry."""
+        if agent in _agents_with_goal_loop:
```
zastrowm, line 296: Why do we care?

The global state thing kind of smells

.agents/skills/docs-reviewer/SKILL.md
notowen333: No but I can add that in a separate PR

site/src/content/docs/user-guide/concepts/plugins/goal-loop.ts
notowen333: yeah lol I think no languages except English is probably a fair rule for these examples

strands-py/src/strands/vended_plugins/goal/judge.py
notowen333: yes confirmed

strands-py/src/strands/vended_plugins/goal/plugin.py
notowen333: Yes it is part of the user facing API and TS exports the same

strands-py/src/strands/vended_plugins/goal/plugin.py
notowen333: Two goal loops on one agent creates a race condition

```
+# Two GoalLoops on one agent both write event.resume in AfterInvocation — only the
 146 +# last callback's value survives, silently breaking the other's retry loop. Guard here.
```

site/src/config/navigation.yml
notowen333: Tags already handle this — both pages share the `hooks` tag, which renders related pages in the sidebar. No additional cross-link needed.

notowen333: added `prompt-authoring` tag which connects them by tag

zastrowm (dismissed):

zastrowm (approved):

Issue #2496, opened 2026-04-09
[FEATURE] Background Tasks Proposal


gautamsirdeshmukh:
### Problem Statement

Enable agent to dispatch tools/agents-as-tools asynchronously and concurrently in the background without blocking the main event loop

### Design

https://github.com/strands-agents/harness-sdk/pull/3531/

### Implementation

This is a large change that touches a bit of preexisting logic, thereby necessitating an incremental review-and-merge process.

Initial parent PR containing entire implementation: https://github.com/strands-agents/harness-sdk/pull/3632

| PR # | Contents | Link | Status |
| ------ | ------ | ------ | ------ |
| 1 | Add `cancelSignal` to ToolContext | https://github.com/strands-agents/harness-sdk/pull/3807 | Merged |
| 2 | Update AfterInvocationEvent to allow multiple `event.resume` requests | https://github.com/strands-agents/harness-sdk/pull/3837 | Merged | 
| 3 | Bring in InProcessTaskEngine | https://github.com/strands-agents/harness-sdk/pull/3838 | Merged |
| 4 | Bring in TaskManager and InProcessTaskManager | https://github.com/strands-agents/harness-sdk/pull/3876 | Merged |
| 5 | Agent integration (internal BackgroundTasks plugin) | ... | In Draft |

JackYPCOnline:
#2740 We should also revisit this PR

pgrayy:
Additional context on #2740, currently, memory utilizes async to implement "fire-and-forget". As with this proposal, it may be worth following a pattern similar to otel where a separate process runs to handle the async requests.

gautamsirdeshmukh:
Adding some relevant context here:
- Initial doc/prototype (very implementation focused) --> https://github.com/strands-agents/docs/pull/780
- Most recently reviewed doc (higher level) --> https://github.com/strands-agents/docs/pull/783

Conversations are revitalized and ongoing, will update issue accordingly

gautamsirdeshmukh:
Updates 7/14:

DONE
- Deep dive into original proposal(s), competitor frameworks, mechanisms in Strands
- Rudimentary POC for backgrounding tools/agents-as-tools

TO DO
- One more review for mechanism, primarily API bar raising
- Full implementation

gautamsirdeshmukh:
Updates 7/28:

DONE
- Design review (happening today)
- Full implementation

TODO
- Finish, review and merge implementation

PR #2740, opened 2026-06-11
feat(memory): port memory manager and extraction to Python

JackYPCOnline:
## Description

Ports the `memory` module from the TypeScript SDK (`strands-ts/src/memory`) to the Python SDK (`strands-py`). The module gives agents cross-session recall and persistence through a `MemoryManager` plugin that manages pluggable `MemoryStore` backends, exposes `search_memory` / `add_memory` tools, and runs automatic background extraction that distills conversation turns into durable memory.

This is a behavior-preserving port: the TypeScript test suite is the reference, and every `it(...)` case has a Python counterpart. The notable design work is adapting the TS promise/event-loop model to Python `asyncio`.

**Public API** — new package `strands.memory`:

```python
from strands import Agent

# A store implements `search` and (optionally) `add` / `add_messages` / `get_tools`,
# and may declare an extraction config to auto-distill turns into memory.
memory_manager = MemoryManager(
    stores=[my_store],
    add_tool_config=True,            # expose the add_memory tool (opt-in)
    flush_on_invocation_end=True,    # await pending extraction writes per invocation
)
agent = Agent(model=model, plugins=[memory_manager])
agent("Remember I prefer dark mode")

# Programmatic access (coroutines):
results = await memory_manager.search("user preferences")
await memory_manager.flush()
```

Also adds `AggregateMemoryError` to `strands.types.exceptions` — a Python 3.10-safe stand-in for JS `AggregateError` (`ExceptionGroup` is 3.11+) used to surface multi-store write failures with each underlying reason.

**Asyncio adaptation (key deviation from TS).** TS relies on a long-lived event loop, so fire-and-forget background saves survive between turns. Python's synchronous `Agent(...)` entry point runs each invocation in a fresh loop (`asyncio.run`), which cancels in-flight tasks on return. The port uses `asyncio.gather(return_exceptions=True)` for `Promise.allSettled`, per-store `asyncio.Task` chains for serialized saves, and a tracked background-task set. The opt-in `flush_on_invocation_end` (default `False`) registers an `AfterInvocationEvent` hook that awaits pending writes before the per-invocation loop tears down; async callers owning a persistent loop can leave it off and call `flush()` at a shutdown boundary. This flag has no TS equivalent and exists solely to bridge the event-loop lifecycle difference.

**Scope.** This PR ports `strands-ts/src/memory/` only. Concrete store backends (e.g. `BedrockKnowledgeBaseStore`, which lives in the separate `strands-ts/src/vended-memory-stores/` module) are intended as a follow-up PR; this change ships the `MemoryManager` machinery and the `MemoryStore` protocol that those backends implement.
## Related Issues

N/A

## Documentation PR

No new docs required; public classes are documented via docstrings.

## Type of Change

New feature

## Testing

Ported the three TS test files (`memory-manager`, `extraction`, `model-extractor`) as example-based `pytest` suites so coverage mirrors the TS suite case-for-case; runs green under `tests/strands/memory`. Adjacent `plugins`/`tools` suites were run to confirm no regressions in plugin/tool discovery.

## Checklist

----


JackYPCOnline:
Thanks for the thorough review. Addressed below.

**Fixed**
- **`MemoryManagerConfig` missing `flush_on_invocation_end`** — added so the doc/typing aid matches the constructor (`2de0192e`).
- **`stores=[]` divergence undocumented** — documented on `MemorySearchOptions.stores` and `MemoryAddOptions.stores` (`2de0192e`). This divergence is mirrored from the TS source (the tool's `_resolveToolTargets` treats `[]` as "all"; programmatic `search`/`add` treat `stores` as an explicit, possibly-empty filter) and is covered by tests on both paths.
- **Default footgun** — kept the TS-faithful fire-and-forget default (`False`) and added a one-time `logger.warning` in `init_agent` when a store configures extraction but `flush_on_invocation_end` is `False` (`4b7ad9aa`). Defaulting to `True` or a blanket warning would diverge from TS and false-positive for async callers who already drain via a persistent loop / explicit `flush()`. Happy to flip the default if the API reviewer prefers "obvious path = happy path" over strict parity.

**Intentional / documented, not changed**
- **`flush()` not draining add-tool tasks** — this mirrors TS exactly (TS `flush()` awaits only the extraction coordinator; the `wait_for_writes=False` add-tool writes are `void`-dispatched and not drained there). Documented the boundary on `flush()` (`4b7ad9aa`) rather than diverge. Can extend `flush()` to drain `self._background_tasks` (with a re-check loop) if you'd prefer the superset behavior — let me know.
- **No concrete store shipped** — intentional scope; added a note to the description. `BedrockKnowledgeBaseStore` lives in the separate `strands-ts/src/vended-memory-stores/` module, so it's a natural follow-up. I can add a small `InMemoryMemoryStore` + an end-to-end integ test here instead if a self-contained anchor is preferred.
- **Design-doc divergence** — confirmed intentional. `designs/0011-memory-manager.md` is still status "Proposed" (describes `store_memory`/injection/`EvictionTrigger`); this PR mirrors the current TS surface (`add_memory`, no injection), which is the right reference for a port.

**Process**
- `needs-api-review` — agreed this warrants API sign-off given the new public surface, but I don't have triage rights on this repo to add the label. Could a maintainer apply it?

JackYPCOnline:
Addressed both in `2f7b22e8`.

**`cast` to `DecoratedFunctionTool`** — you're right, the cast misrepresented store-provided tools (which can be any `AgentTool`). Rather than widen the base `Plugin` annotation, I dropped the cast and overrode the `tools` property to return `list[AgentTool]` (honest element type), with a single `# type: ignore[override]` for the base's narrower `list[DecoratedFunctionTool]` annotation.

While doing that I also removed the now-redundant `MemoryManager.get_tools()` method: in the Python `Plugin` contract the framework registers tools via the `tools` property (`_PluginRegistry._register_tools` reads `plugin.tools`), so `get_tools()` was a duplicate surface with no caller (the TS `getTools()` maps to the `tools` property here). Tests updated to read `.tools`.

**Warning fires for correct async usage** — good point. Reworded so it reads as informational rather than misconfiguration and explicitly notes the safe case:
> `flush_on_invocation_end=<False>` | background extraction is lost if the event loop closes before it finishes (e.g. the synchronous `Agent(...)` entry point); safe to ignore if you await `MemoryManager.flush()` at a shutdown boundary or enable `flush_on_invocation_end`.

Kept it at `warning` (rather than `debug`) so the sync-path footgun stays visible, since that's the case most likely to silently drop writes.

strands-py/src/strands/memory/extraction/coordinator.py
```diff
@@ -0,0 +1,390 @@
...
+failure streak and resumes normal saving -- so a transient outage recovers on
+its own and the messages buffered during it are saved once the store comes back.
+A permanently broken store keeps probing and logs an error each time, surfacing
+the misconfiguration.
+
+Scheduling uses ``asyncio.create_task`` for fire-and-forget saves, a per-store
+``asyncio.Task`` chain to serialize a single store's saves, and
+``asyncio.gather(..., return_exceptions=True)`` to run concurrent writes so one
+failure does not cancel the rest.
+"""
```
pgrayy, line 8: I think it would be good to note somewhere that this mechanism might be worth revisiting. If we want "fire-and-forget" it might be better here to do it similar to otel which involves asynchronously running export on a separate process. Maybe something we can revisit when we work on background tool tasks. But again, I think we should record this somewhere. Maybe on the background tool task ticket.

As it is implemented here and in TS, I would say it is not true "fire-and-forget" because saves block other work when running due to the single threaded-ness of both languages.

strands-py/src/strands/memory/extraction/coordinator.py
```diff
@@ -0,0 +1,390 @@
...
+        if not fresh:
+            return
+
+        # Mark these messages as saved before we start saving, so a queued save
+        # behind this one will not pick them up again. If the save fails we put
+        # the mark back (below) and they retry.
+        self._marks[id(store)] = fresh[-1].seq
+
+        extraction = store.extraction
+        assert extraction is not None  # noqa: S101 - extraction stores always configure this.
```
pgrayy: Nit: If this is always configured, I would recommend not asserting and update the field type as well if needed to say that None is not supported.

strands-py/src/strands/memory/extraction/coordinator.py
```diff
@@ -0,0 +1,390 @@
...
+            self._background.discard(completed)
+            if completed.cancelled():
+                return
+            error = completed.exception()
+            if error is not None:
+                logger.warning("store=<%s>, reason=<%s> | background memory save failed", store.name, error)
+
+        task.add_done_callback(_done)
+
+    async def process(self, store: MemoryStore) -> None:
```
pgrayy: Nit: I would say process doesn't need to be async since enqueue anyway offloads to `create_task`. With that said, you could collapse a bit here and have process return the created task instead of itself being wrapped in a task above in schedule. This would be a minor memory optimization since of course it is less task scheduling on the event loop.

strands-py/src/strands/memory/extraction/coordinator.py
```diff
@@ -0,0 +1,390 @@
...
+
+        Args:
+            store: The store to write to.
+            messages: The filtered messages to save.
+
+        Raises:
+            AggregateMemoryError: If any concurrent ``add`` write fails.
+        """
+        extraction = store.extraction
+        assert extraction is not None  # noqa: S101 - extraction stores always configure this.
```
pgrayy: Nit: See comment above.

strands-py/src/strands/memory/extraction/coordinator.py
```diff
@@ -0,0 +1,390 @@
...
+
+        Args:
+            block: The content block to classify.
+
+        Returns:
+            The block's kind, or ``""`` for an empty block.
+        """
+        if "text" in block:
+            return "text"
+        return next(iter(block.keys()), "")
```
pgrayy, line 233: Nit: Seems like you don't need to special case text.
```
{"text": ...}
{"toolUse": ...}
```
^ It is the same structure. One key per content block.

strands-py/src/strands/memory/extraction/model_extractor.py
```diff
@@ -0,0 +1,154 @@
...
+from ...types.content import Message
+from .types import ExtractionResult, ExtractorContext
+
+logger = logging.getLogger(__name__)
+
+# Default instruction guiding the model to emit discrete, durable facts as a JSON array.
+DEFAULT_SYSTEM_PROMPT = (
+    "You extract durable facts worth remembering across future conversations from a transcript.\n"
+    "\n"
+    'Return ONLY a JSON array of objects, each: {"content": string}. Each object is one discrete, '
```
pgrayy, line 24: We should utilize structured output here to make it more reliable.

strands-py/src/strands/memory/extraction/coordinator.py
```diff
@@ -0,0 +1,390 @@
...
+            messages: The filtered messages to save.
+
+        Raises:
+            AggregateMemoryError: If any concurrent ``add`` write fails.
+        """
+        extraction = store.extraction
+        assert extraction is not None  # noqa: S101 - extraction stores always configure this.
+
+        if extraction.extractor is not None:
+            entries = await extraction.extractor.extract(messages, ExtractorContext(default_model=self._default_model))
```
pgrayy: What if a fact only really defines itself across multiple messages? We could miss that in the extraction right because of the trimming we do over turns? Is under-extraction a problem you think?

```
Turn 1  user: I have a friend.
Turn 1  assistant: Cool, what is their name?
Turn 2  user: Bob
```

If turn 1 is trimmed before turn 2 extraction, we maybe lose the fact that Bob specifically is a friend.

Not something that needs to be addressed for this PR but still curious on what has been discussed around this.

strands-py/src/strands/types/exceptions.py
```diff
@@ -127,3 +127,27 @@ class CheckpointException(Exception):
...
     pass
+
+
+class AggregateMemoryError(Exception):
+    """Raised when one or more memory store operations fail.
+
+    Carries a human-readable ``message`` naming the failed stores and exposes each
+    underlying failure through the ``errors`` attribute. Used instead of
+    ``ExceptionGroup`` so the SDK supports Python 3.10 (``ExceptionGroup`` is only
+    available on 3.11+).
```
pgrayy: Nit: Another example of an implementation decision that I would say doesn't belong in the docstring. Not saying you have to pull it out now but we should find a better place because this is useful context for our code translation workflow.

strands-py/src/strands/memory/memory_manager.py
```diff
@@ -0,0 +1,696 @@
...
+/ ``add_memory`` tools for agent-driven recall and persistence, alongside any
+tools the stores themselves provide via
+:meth:`~strands.memory.types.MemoryStore.get_tools`.
+
+As a :class:`~strands.plugins.plugin.Plugin`, the manager builds its tools at
+construction (exposed via the ``tools`` property) and wires automatic extraction
+in :meth:`MemoryManager.init_agent` for any store configured with an
+``ExtractionConfig``.
+
+All public field and method names use ``snake_case``.
```
pgrayy: Nit: Not necessary to state in a docstring in my opinion.

strands-py/src/strands/memory/memory_manager.py
```diff
@@ -0,0 +1,696 @@
...
+    per-invocation event loop.
+
+    Example:
+        ```python
+        from strands import Agent
+        from strands.memory import MemoryManager
+
+        # Synchronous Agent(...) entry point: enable flush_on_invocation_end so
+        # extraction writes persist across each invocation's event loop.
+        memory_manager = MemoryManager(stores=[my_store], flush_on_invocation_end=True)
```
pgrayy, line 76: Should we enable this automatically on the users behalf when a sync invoke is called?

strands-py/src/strands/memory/memory_manager.py
```diff
@@ -0,0 +1,696 @@
...
+        ```
+    """
+
+    name = "strands:memory-manager"
+
+    def __init__(
+        self,
+        stores: list[MemoryStore],
+        search_tool_config: MemoryToolConfig | bool = True,
+        add_tool_config: MemoryAddToolConfig | bool = False,
```
pgrayy, line 90: Nit: Why not `MemorySearchToolConfig` for symmetry with `MemoryAddToolConfig`?

strands-py/src/strands/memory/memory_manager.py
pgrayy, line 1: Should we have a dedicated memory_manager param on Agent init? This would match the pattern of conversation_manager and session_manager.

pgrayy, line 1: Do we have integ tests planned?

JackYPCOnline: yes, should be after Owen's PR

strands-py/src/strands/memory/extraction/coordinator.py
JackYPCOnline: Agree, that is a good call out, added to background task issue.
https://github.com/strands-agents/harness-sdk/issues/2496

strands-py/src/strands/memory/extraction/coordinator.py
JackYPCOnline: IMO it is something derived from design, agree, we can track it.

strands-py/src/strands/memory/memory_manager.py
JackYPCOnline: Yes, should be in the follow-up PR.

strands-py/src/strands/memory/memory_manager.py
JackYPCOnline: It is design question, and it is not free for a model_extractor, we can revisit it later.

strands-py/src/strands/memory/memory_manager.py
JackYPCOnline: it is port from TS, so I would like to keep as-is.

pgrayy (approved):

PR #2742, opened 2026-06-11
feat!: copy middleware context inputs to prevent accidental mutation

zastrowm:
## Description

Middleware previously received live references to the agent's internal state. A middleware that accidentally mutated the messages array, a toolSpec, or modelState would corrupt agent state for subsequent turns.

This PR makes middleware context inputs defensive copies across the public stages:

**InvokeModelStage** — `messages`, `systemPrompt`, `toolSpecs`, and `toolChoice` are all deep-copied. `modelState` is removed from the context entirely (snapshotted before the chain, written back after). `invocationState` remains shared by reference since hooks and tools write to it during streaming.

**ExecuteToolStage** — `toolUse` is deep-copied to prevent mutation of the parsed tool input in the assistant message. `invocationState` remains shared by reference.

**AgentStreamStage** — Removed from the public API. The context contract (copy vs. reference for `args`/`options`) isn't finalized, and shipping it with different guarantees than the other stages would be confusing. It continues to work internally.

Design rationale is documented in `strands-ts/src/middleware/README.md`.

## Type of Change

Breaking change

## Breaking Changes

- `modelState` removed from `InvokeModelContext`
- `AgentStreamStage`, `AgentStreamContext`, `AgentStreamResult` removed from public exports

Both are acceptable — middleware just shipped and has no external consumers yet.

## Testing

23 tests in `copy-on-input.test.ts` covering isolation guarantees for all copied fields, shared-reference behavior for `invocationState`, modelState writeback semantics, and functional-style context passing.

## Checklist

----

### DIFF
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -1984,6 +1984,7 @@ export class Agent implements LocalAgent, InvokableAgent {
 
     // Sync model state after the entire middleware chain has completed, so no
     // middleware mutation to agent.modelState (before or after next()) takes effect.
+    // Intentionally skipped on error — partial provider writes should not persist.
     if (tempModelState) {
       loadStateSerializable(this.modelState, serializeStateSerializable(tempModelState))
     }
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -27,7 +27,7 @@ import type { JSONValue } from '../types/json.js'
 import { McpClient } from '../mcp.js'
 import { isValidToolName, type Tool, type ToolContext } from '../tools/tool.js'
 import type { ToolChoice, ToolSpec } from '../tools/types.js'
-import { systemPromptFromData } from '../types/messages.js'
+import { cloneSystemPrompt, systemPromptFromData } from '../types/messages.js'
 import { normalizeError, ConcurrentInvocationError, StructuredOutputError } from '../errors.js'
 import { Model } from '../models/model.js'
 import type { BaseModelConfig, StreamAggregatedResult, StreamOptions } from '../models/model.js'
@@ -35,6 +35,7 @@ import { ModelPlugin } from '../plugins/model-plugin.js'
 import { isModelStreamEvent } from '../models/streaming.js'
 import { ToolRegistry } from '../registry/tool-registry.js'
 import { StateStore } from '../state-store.js'
+import { serializeStateSerializable, loadStateSerializable } from '../types/serializable.js'
 import { AgentPrinter, getDefaultAppender, type Printer } from './printer.js'
 import type { Plugin } from '../plugins/plugin.js'
 import type { InterventionHandler } from '../interventions/handler.js'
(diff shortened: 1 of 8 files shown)


strands-ts/src/agent/agent.ts
```diff
@@ -1919,14 +1921,19 @@ export class Agent implements LocalAgent, InvokableAgent {
   ): AsyncGenerator<AgentStreamEvent, StreamAggregatedResult, undefined> {
     const context: InvokeModelContext = {
       agent: this,
-      messages: this.messages,
-      ...(this.systemPrompt !== undefined && { systemPrompt: this.systemPrompt }),
-      toolSpecs: this._toolRegistry.list().map((tool) => tool.toolSpec),
-      ...(toolChoice !== undefined && { toolChoice }),
-      modelState: this.modelState,
-      invocationState,
+      messages: this.messages.map((msg) => msg.clone()),
```
notowen333, line 1924: would it be more or less organized to write these cloning utilities in `middleware/` instead of inline

strands-ts/src/middleware/stages.ts
```diff
@@ -109,7 +106,7 @@ export interface ExecuteToolContext extends MiddlewareInterruptible {
   readonly tool: Tool | undefined
   /** The tool use request (name, toolUseId, input). */
   readonly toolUse: ToolUseData
-  /** Per-invocation state shared across hooks and tools. */
+  /** Per-invocation state shared across hooks and tools. Shared by reference — mutations are intentionally visible to subsequent hooks, tools, and the AgentResult. */
```
notowen333: "and the SDK should never make mutations directly"

strands-ts/src/middleware/README.md
```diff
@@ -56,6 +56,16 @@ Middleware is intended to supersede hooks. The Input/Output phases already cover
...
 
+## Middleware cannot access or modify model state
+
+`modelState` is intentionally excluded from `InvokeModelContext`. The agent snapshots model state before the middleware chain runs and writes back the model provider's changes after the entire chain completes. Any mutations middleware makes to `agent.modelState` — whether before or after `next()` — are overwritten by this writeback.
+
+We may revisit this later.
+
+## `invocationState` is shared by reference
+
+`invocationState` is not copied. Tools and hooks write to it, and those mutations must appear on `AgentResult.invocationState`.
```
notowen333: "and the SDK should never write to it"

zastrowm: Fair, though this is larger than just middleware IMO

strands-ts/src/types/messages.ts
```diff
@@ -117,6 +117,17 @@ export class Message implements JSONSerializable<MessageData> {
...
+
+  /**
+   * Creates a deep copy of this Message (round-trips through serialization).
+   */
+  clone(): Message {
+    const data = this.toJSON()
+    if (data.metadata !== undefined) {
+      data.metadata = JSON.parse(JSON.stringify(data.metadata))
+    }
+    return Message.fromMessageData(data)
```
opieter-aws: The same `toJSON()` reference-sharing problem as flagged by the bot still affects `ToolUseBlock.input` and `JsonBlock.json`. Why not do a full JSON round-trip on the entire `MessageData`?

strands-ts/src/middleware/README.md
opieter-aws: Is this true for the `InvokeModelStage`, given that we do `{ ...invocationState }`?

opieter-aws: I think that'd break `invocationState` changes that any `ModelStreamUpdateEvent` and `ContentBlockEvent` hooks make

strands-ts/src/agent/agent.ts
opieter-aws, line 1924: These copies run on every model call even when no `InvokeModelStage` middleware is registered. Consider gating this, since it introduces latency as the conversation history grow

strands-ts/src/types/messages.ts
zastrowm: I have to look more into this; it's either a miss that that metadata isn't part of toJSON() and fromMessageData OR metadata isn't normally copied and it's just in this case that we need it

zastrowm: Yeah, switching to doin ga deep copy of the entire thing to avoid special casing metadata

strands-ts/src/agent/agent.ts
zastrowm: going to defer this until we see performance problems.  In general and compared to network requests, this is trivial and I think code clarity is more important until we're seen impact here

zastrowm: I tried it, but it just obscures the code IMO

strands-ts/src/middleware/README.md
zastrowm: Agreed; pass by copy

opieter-aws (approved):

strands-ts/src/middleware/README.md
zastrowm: Discussed offline; switched to by reference and will have to rely on documentation

pgrayy (approved):

PR #2751, opened 2026-06-12
docs: add Syntax component for inline language-specific identifiers

notowen333:
## Description

Adds a new `<Syntax>` MDX component that renders language-appropriate identifiers (Python snake_case vs TypeScript camelCase) inline based on the global language selector. This eliminates the manual pattern of spelling out both variants in shared prose (e.g., `` `snake_name` / `camelName` ``).

## Motivation

Documentation pages frequently reference SDK identifiers in prose that sits outside language-specific tabs. Previously this meant writing both names inline, which was verbose, inconsistent across pages, and didn't react to the reader's language preference. The new component provides a single source of truth per identifier that switches live with the language toggle.

## Public API Changes

No public API changes — this is a documentation infrastructure addition.

### Component API

```mdx
<Syntax py="conversation_manager" ts="conversationManager" />
<Syntax py=".as_tool()" ts=".asTool()" />
<Syntax py="strict=True" ts="strict: true" code={false} />
```

| Prop | Type | Default | Description |
|------|------|---------|-------------|
| `py` | string | required | Python identifier |
| `ts` | string | required | TypeScript identifier |
| `code` | boolean | `true` | Wrap in `<code>` (false for plain text) |

## Implementation

- `site/src/components/Syntax.astro` — custom element `<syntax-switch>` with an idempotent inline script that reads the same localStorage key as the existing language tabs
- Auto-imported via `astro-auto-import` so no explicit imports needed in MDX
- Reacts to tab clicks (`.lang-option` delegated listener) and cross-tab `storage` events

## Scope

- Converted ~26 instances across 16 documentation pages
- Updated agent skills (writer, reviewer, audit) to teach future authors about the component
- Updated references (mdx-authoring.md, voice-guide.md) and contributor docs

## Type of Change

Documentation update

## Testing

- [x] Full `npx astro build` passes (655 pages, 0 broken links)
- [x] Verified every TypeScript identifier against SDK source at `sdk-typescript/strands-ts/src/`
- [x] Manual browser testing confirms live language switching

## Checklist

----

### DIFF
--- a/site/src/content/docs/contribute/contributing/documentation.mdx
+++ b/site/src/content/docs/contribute/contributing/documentation.mdx
@@ -89,6 +89,22 @@ Code examples are critical—they show developers exactly what to do. Always tes
 - Add comments — explain non-obvious parts
 - Use realistic names — avoid foo/bar, use descriptive names
 
+### Inline language-specific names
+
+When prose references a method, parameter, or class name that differs between Python and TypeScript, use the `<Syntax>` component instead of writing both variants manually:
+
+```mdx
+<!-- Good: adapts to the reader's language selection -->
+Use <Syntax py=".as_tool()" ts=".asTool()" /> to customize the tool name.
+
+<!-- Bad: clutters prose with both variants -->
+Use `.as_tool()` (Python) or `.asTool()` (TypeScript) to customize the tool name.
+```
+
+`<Syntax>` is auto-imported (no `import` statement needed). It renders as inline `<code>` by default and reacts live to the global language toggle. Pass `code={false}` for plain text output.
+
+Use `<Syntax>` for simple name swaps. For code blocks or conceptual differences between SDKs, use `<Tabs>` instead.
+
 <Tabs>
 <Tab label="Python">
 
--- a/site/src/content/docs/contribute/contributing/documentation.mdx
+++ b/site/src/content/docs/contribute/contributing/documentation.mdx
@@ -6,6 +6,27 @@ sidebar:
 
 Good documentation helps developers succeed with Strands. We welcome contributions that make our docs clearer, more complete, or more helpful. Our documentation lives in the `site/` directory of the [Strands SDK repository](https://github.com/strands-agents/harness-sdk).
 
+## Agent skills
+
+The repository ships agent skills in `.agents/skills/` that encode our documentation conventions, site components, and quality standards. If you're using an AI coding assistant, activate these skills before making documentation changes — they'll produce output that matches our style without you needing to internalize everything on this page.
+
+| Skill | What it does |
+|-------|--------------|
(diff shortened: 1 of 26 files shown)


site/src/content/docs/contribute/contributing/documentation.mdx
```diff
@@ -89,6 +89,22 @@ Code examples are critical—they show developers exactly what to do. Always tes
 - Add comments — explain non-obvious parts
 - Use realistic names — avoid foo/bar, use descriptive names
 
+### Inline language-specific names
```
lizradway: I think this section is very specific in comparison to the rest of the page. ie. we do not have a section telling users how to use tabs. i don't think this needs to be explained on the site, imo agents.md/ contribution guide / skills is plenty. 

Syntax is for developing the site, so plain readers of the site wanting to contribute do not need to know about it until they are in the development weeds.

lizradway:
as a follow up, might be worth going through and ensuring that we actually have bilingual values for all of the code var references. Conversation management concept page is a good example of this; a lot of vars are cited as code, but do not have python/TS verisons.

lizradway:
we can also probably remove a lot of the "In TypeScript, ModelRetryStrategy is the abstract base class. Use DefaultModelRetryStrategy for the standard configurable strategy:" type things now ! as a follow up

site/src/content/docs/user-guide/concepts/model-providers/google.mdx
lizradway, line 106: im assuming this is underscores in python? im also assuming there are many more examples like this we will have to catch

notowen333: Yes — `client_args` is the Python kwarg, `apiKey` is the TS equivalent. I did a thorough grep and converted every instance I found in shared prose (~26 total across 16 pages), plus caught a few more in a final pass (`count_tokens`/`countTokens` in context-offloader just now). If more surface later the `docs-reviewer` skill now flags them automatically.

lizradway: awesome

site/src/content/docs/contribute/contributing/documentation.mdx
```diff
@@ -6,6 +6,27 @@ sidebar:
 
 Good documentation helps developers succeed with Strands. We welcome contributions that make our docs clearer, more complete, or more helpful. Our documentation lives in the `site/` directory of the [Strands SDK repository](https://github.com/strands-agents/harness-sdk).
 
+## Agent skills
```
lizradway, line 9: nice!

lizradway (dismissed):

zastrowm (approved):

lizradway (approved):

PR #2754, opened 2026-06-12
feat: add agentic context management with model-driven compression tools

notowen333:
## Summary
- Adds `contextManager: 'agentic'` option that registers context tools (`summarize_context`, `truncate_context`, `pin`) and injects `<context-status>` XML via middleware so the model can autonomously manage its own context window
- Refactors shared compression logic (summarization, pin handling, tool-pair safety) out of `SlidingWindowConversationManager` and `SummarizingConversationManager` into a reusable `compression/` module
- Adds `InvokeModelStage.Input` middleware stage for pre-model-call message transformation

## Key design decisions
- Model is guided via tool descriptions (e.g. "Often most useful with messageType 'messages'") rather than hard-coded logic — tested and confirmed working with Bedrock Sonnet
- Context tools operate directly on `agent.messages` (in-place splice) so compression effects persist across invokes
- `messageType` filter lets the model selectively target chat messages vs tool results for compression
- Pinned messages are always protected from eviction

## Test plan
- [x] 174 unit tests pass (agentic-context, context-compression, token-usage-middleware, pin, sliding-window, summarizing)
- [x] 11 integration tests pass (middleware injection, tool invocation, end-to-end coherence)
- [x] Build passes cleanly
- [x] Manually verified model autonomously chooses `summarize_context` with `messageType: "messages"` for chat and `truncate_context` with `messageType: "tools"` for bulky tool results


strands-ts/src/agent/agent.ts
```diff
@@ -321,6 +328,9 @@ function resolveConversationManager(
       })
     )
   }
+  if (contextManager !== undefined) {
+    throw new Error(`Unsupported contextManager value: "${contextManager}". Supported values: "auto", "agentic"`)
```
lizradway: can we reference the supported values `ContextManagerStrategy` directly?

strands-ts/src/conversation-manager/compression/index.ts
```diff
@@ -0,0 +1,12 @@
+export { isPinned, pinMessage, unpinMessage, applyPinFirst, partitionPinned } from './pin-message.js'
```
lizradway: why are we exporting all of this? we're not supposed to export anything right?

strands-ts/src/agent/agent.ts
```diff
@@ -215,15 +216,15 @@ export type AgentConfig = {
...
-   * Context management strategy. When set to `"auto"`, composes a ContextOffloader
-   * plugin (maxResultTokens=1500, previewTokens=750) with a SummarizingConversationManager
-   * (summaryRatio=0.3, compressionThreshold=0.85) using benchmark-validated defaults.
-   * If `conversationManager` is also provided, the user's conversation manager is used instead.
-   * Defaults to undefined (no context management).
+   * Context management strategy.
+   *
+   * - `"auto"`: SummarizingConversationManager with proactive compression + ContextOffloader.
+   * - `"agentic"`: Lets the model drive context management via injected tools
+   *   (summarize_context, truncate_context, pin). Uses SummarizingConversationManager
```
lizradway: i find this comment confusing. tbh can just keep it to the one line " Lets the model drive context management via injected tools" and explain more in the actual agentic context file imo

strands-ts/src/agent/agent.ts
```diff
@@ -215,15 +216,15 @@ export type AgentConfig = {
...
+   *   without proactive compression as a reactive safety net, plus ContextOffloader.
    *
-   * @remarks The offloader uses in-memory storage that does not persist across process
```
lizradway, line 226: pls keep this comment / do not remove

strands-ts/src/agent/agent.ts
```diff
@@ -302,15 +303,21 @@ export type AgentConfig = {
  * Resolve the contextManager facade into a concrete ConversationManager.
  *
  * When contextManager is undefined, falls back to the default SlidingWindowConversationManager.
- * When "auto", uses SummarizingConversationManager with benchmark-validated defaults,
- * unless the user already provided a conversationManager.
+ * When "auto", uses SummarizingConversationManager with proactive compression.
+ * When "agentic", uses SummarizingConversationManager without proactive compression
+ * (the agent manages its context via tools; the CM is only a reactive safety net).
```
lizradway: nit: trying to use "context manager" in full everywhere instead of "cm"

strands-ts/src/agent/agent.ts
```diff
@@ -535,7 +552,7 @@ export class Agent implements LocalAgent, InvokableAgent {
       this._conversationManager,
       ...retryStrategies,
       ...(config?.plugins ?? []),
-      ...(config?.contextManager === 'auto' && !hasOffloader
+      ...((config?.contextManager === 'auto' || config?.contextManager === 'agentic') && !hasOffloader
```
lizradway, line 569: should we have larger offloading limits? or allow the agent to configure this at all?


i think it makes sense to have a higher token barrier to offload if we are _also_ exposing a tool for this.

strands-ts/src/agent/agent.ts
```diff
@@ -1915,7 +1932,8 @@ export class Agent implements LocalAgent, InvokableAgent {
    */
   private async *_invokeModelWithMiddleware(
     invocationState: InvocationState,
-    toolChoice?: ToolChoice
+    toolChoice?: ToolChoice,
+    projectedInputTokens?: number
```
lizradway, line 1953: we should probably add this to options rather than an optional positional param

strands-ts/src/conversation-manager/modes/agentic/agentic-context.ts
```diff
@@ -0,0 +1,239 @@
+import { z } from 'zod'
```
lizradway, line 1: we already discussed changing the directory strucuture to `context-manager` offline

strands-ts/src/conversation-manager/agentic/agentic-context.ts
```diff
@@ -0,0 +1,239 @@
...
+import { pinMessage, unpinMessage, isPinned } from '../compression/pin-message.js'
+import {
+  generateSummary,
+  adjustSplitPointForToolPairs,
+  findValidTrimPoint,
+  matchesMessageType,
+  type MessageTypeFilter,
+} from '../compression/context-compression.js'
+
+const DEFAULT_KEEP_RECENT = 10
```
lizradway: nit: keep recent what? messages?

lizradway: in general these constants are a little difficult to understand what they are for by the naming

strands-ts/src/conversation-manager/modes/agentic/agentic-context.ts
```diff
@@ -0,0 +1,239 @@
...
+const MIN_SUMMARY_RATIO = 0.1
+const MAX_SUMMARY_RATIO = 0.8
+const MIN_MESSAGES_FOR_OPERATION = 2
+
+const messageTypeSchema = z
+  .enum(['tools', 'messages', 'all'])
+  .optional()
+  .describe(
+    'Filter which messages to target. "tools" targets only tool use/result messages, ' +
+      '"messages" targets only non-tool messages, "all" (default) targets everything.'
```
lizradway, line 33: should we filter by user / assistant message as well?

message being non-tool is a bit of a catch all.... what other items live in the messages array? is there a way we could do inverse (ie. select all messages _except_ user messages, i see that as a use case for future context for sure.

lizradway: we can add this later since this isn't exposed. but just a note

strands-ts/src/conversation-manager/agentic/agentic-context.ts
```diff
@@ -0,0 +1,239 @@
...
+      .optional()
+      .describe(
+        `Fraction of the oldest messages to fold into the summary (${MIN_SUMMARY_RATIO}–${MAX_SUMMARY_RATIO}). Defaults to ${DEFAULT_SUMMARY_RATIO}.`
+      ),
+    messageType: messageTypeSchema,
+  }),
+  callback: async ({ keepRecent, summaryRatio, messageType }, context) => {
+    const agent = context!.agent
+    const messages = agent.messages
+    const before = messages.length
```
lizradway: nit: before? what does this mean

lizradway: similarly i think `agent` and `messages` as names that are not very explainatory/specific for this. 

im also confused bc we are not even using `before` in a lot of this:

ie. `let splitPoint = Math.max(1, Math.floor(messages.length * ratio))`

strands-ts/src/conversation-manager/agentic/agentic-context.ts
```diff
@@ -0,0 +1,239 @@
...
+      .int()
+      .min(MIN_MESSAGES_FOR_OPERATION)
+      .optional()
+      .describe(`Number of most recent messages to keep. Everything older (and unpinned) is dropped. Defaults to ${DEFAULT_KEEP_RECENT}.`),
+    messageType: messageTypeSchema,
+  }),
+  callback: ({ keepRecent, messageType }, context) => {
+    const agent = context!.agent
+    const messages = agent.messages
+    const before = messages.length
```
lizradway: same comments as above

strands-ts/src/conversation-manager/modes/agentic/agentic-context.ts
```diff
@@ -0,0 +1,239 @@
...
+  },
+})
+
+export const pinTool = tool({
+  name: 'pin',
+  description:
+    'Pin or unpin messages in the conversation history. ' +
+    'Pinned messages are protected from eviction during context reduction (summarize or truncate). ' +
+    'Use this to preserve important context that should not be lost. ' +
+    'Select messages using relative references: pin the current exchange, the last N messages, or specific indices.',
```
lizradway, line 218: i think we could update this description to be more thorough;

1. warning against over pinning/ more specific than `Use this to preserve important context that should not be lost.`
2. maybe an example of this:  `pin the current exchange, the last N messages, or specific indices.`

i think 2 is less important than 1 imo

strands-ts/src/conversation-manager/agentic/agentic-context.ts
```diff
@@ -0,0 +1,239 @@
...
+          type: z.literal('last_n'),
+          count: z.number().int().min(1).describe('Number of messages from the end to select.'),
+        }),
+        z.object({
+          type: z.literal('indices'),
+          indices: z.array(z.number().int().min(0)).min(1).describe('Zero-based message indices to select.'),
+        }),
+      ])
+      .describe(
+        'How to select messages. "current_turn" pins the last user+assistant exchange. ' +
```
lizradway: can we pin only user/assistant exchanges?

can we pin only user or only assistant? can we pin tool results?

lizradway: i think its important that we can only pin user messages as an option.


fine to ship as is and we can follow up on these things after summit if they are not breaking

strands-ts/src/conversation-manager/agentic/agentic-context.ts
```diff
@@ -0,0 +1,239 @@
...
+  callback: ({ selector, action }, context) => {
+    const messages = context!.agent.messages
+
+    if (messages.length === 0) {
+      return 'No messages in the conversation.'
+    }
+
+    let targetIndices: number[]
+
+    if (selector.type === 'current_turn') {
```
lizradway: so a turn is defined as any batch of assistant messages followed up user messages, both can be any length?


i'd want to iterate on this definition after summit

strands-ts/src/agent/agent.ts
notowen333: I think this follows the establlished pattern for passing args through to the InvokeModelContext.

What options do you mean and why is that better? We added `projectedInputTokens` onto `InvokeModelContext` at the top level in this PR

strands-ts/src/conversation-manager/agentic/index.ts
```diff
@@ -0,0 +1,2 @@
+export { summarizeContextTool, truncateContextTool, pinTool } from './agentic-context.js'
```
lizradway: +1 to not exporting anything rn

strands-ts/src/conversation-manager/agentic/token-usage-middleware.ts
```diff
@@ -0,0 +1,39 @@
...
+export function createTokenUsageMiddleware(model: Model): MiddlewareInputHandler<InvokeModelContext> {
+  return async (context: InvokeModelContext): Promise<InvokeModelContext> => {
+    const projectedInputTokens = context.projectedInputTokens
+    if (projectedInputTokens === undefined) {
+      return context
+    }
+
+    const contextWindowLimit = model.getConfig().contextWindowLimit ?? DEFAULT_CONTEXT_WINDOW_LIMIT
+    const remaining = Math.max(0, contextWindowLimit - projectedInputTokens)
+    const pct = ((projectedInputTokens / contextWindowLimit) * 100).toFixed(1)
```
lizradway: what does this standfor? can we just use that

strands-ts/src/conversation-manager/agentic/token-usage-middleware.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    const statusText =
+      `\n\n<context-status>\n` +
+      `<used>${projectedInputTokens.toLocaleString()} / ${contextWindowLimit.toLocaleString()} tokens (${pct}%)</used>\n` +
+      `<remaining>~${remaining.toLocaleString()} tokens</remaining>\n` +
+      `<tools>summarize_context, truncate_context, pin</tools>\n` +
```
lizradway: is this an attempt to point the agent towards the tool specs for these? or what is the use?


 im also wondering seeing this if we should rename `pin` to `pinMessage`/`pinContext` etc?

strands-ts/src/conversation-manager/compression/context-compression.ts
```diff
@@ -0,0 +1,255 @@
+import { Message, TextBlock } from '../../types/messages.js'
+import type { Model } from '../../models/model.js'
+import { isPinned } from './pin-message.js'
+
+export const DEFAULT_SUMMARIZATION_PROMPT = `You are a conversation summarizer. Provide a concise summary of the conversation \
```
lizradway, line 4: after summit, we should consider experimenting with different prompts to summarizing and new strategies based on context reduction prompts

strands-ts/src/conversation-manager/compression/context-compression.ts
```diff
@@ -0,0 +1,255 @@
+import { Message, TextBlock } from '../../types/messages.js'
+import type { Model } from '../../models/model.js'
+import { isPinned } from './pin-message.js'
+
```
lizradway, line 3: why is all of this exported? all of it is lol

strands-ts/src/conversation-manager/compression/pin-message.ts
```diff
@@ -57,6 +57,26 @@ export function pinMessage(messages: Message[], index: number): void {
   }
 }
 
+/**
+ * Unpin a message so it can be evicted during context reduction.
```
lizradway: ultra nit: might not be worth the commit history change for moving this, it's really really hard to tell on the diff what changes on this function, if anything

strands-ts/src/conversation-manager/compression/context-compression.ts
```diff
@@ -0,0 +1,255 @@
...
+  return new Message({
+    role: 'user',
+    content: result.value.message.content,
+  })
+}
+
+/**
+ * Partition a range of messages into pinned and unpinned groups.
+ */
+export function partitionPinned(
```
lizradway: should this be in `pin-messages.ts`?

lizradway: actually im pretty sure we already have this same function there and this is duplicate

strands-ts/src/conversation-manager/index.ts
```diff
@@ -19,3 +19,5 @@ export {
   SummarizingConversationManager,
   type SummarizingConversationManagerConfig,
 } from './summarizing-conversation-manager.js'
+export { summarizeContextTool, truncateContextTool, pinTool } from './agentic/index.js'
```
lizradway: remove

strands-ts/src/conversation-manager/compression/context-compression.ts
notowen333: It's not pure pinning, so I didn't put it there

strands-ts/src/middleware/stages.ts
```diff
@@ -87,6 +87,8 @@ export interface InvokeModelContext {
   readonly modelState: StateStore
   /** Per-invocation state shared across hooks and tools. */
   readonly invocationState: InvocationState
+  /** Estimated input token count for this model call, or undefined if estimation failed. */
```
lizradway, line 87: we should add this in options rather than as a positional param so this can scale imo @zastrowm

strands-ts/src/conversation-manager/compression/pin-message.ts
notowen333: ? A little confused by the comment can you say more

strands-ts/src/conversation-manager/agentic/token-usage-middleware.ts
notowen333: 1) Yes. We can kill that row though

2) `pinContext` since it's not always a message works better for me

strands-ts/src/conversation-manager/compression/context-compression.ts
notowen333: i.e. it uses a DS in the input that's only for the agentic mode

strands-ts/src/middleware/stages.ts
zastrowm: WDYM - this isn't positional at all, right?

notowen333: We had aligned yesterday to keep it top level. Wdym by scale?

lizradway: nvm got confused, sorry

strands-ts/src/agent/agent.ts
lizradway: nvm!

strands-ts/src/conversation-manager/compression/pin-message.ts
lizradway: oh. it looks on my end like this function just got moved up in the file by ~40 lines without changes? is that correct? we lose the immediate diff view due to changing position is all, which makes it hard to tell

strands-ts/src/conversation-manager/agentic/agentic-context.ts
notowen333: I redesigned the tool interface which should cover these

strands-ts/src/conversation-manager/compression/context-compression.ts
lizradway: sorry what is DS? can we consolidate by calling the existing partition pinned with start=0? i think two functions w the same name that are basically the same is pretty confusing

strands-ts/src/conversation-manager/compression/pin-message.ts
notowen333: yeah I was reconciling two versions of pinning utils... I can move it back

strands-ts/src/conversation-manager/compression/context-compression.ts
notowen333: It's a really simple functionality. I think this is actually simpler to have a 3 way partition separate instead of first re-using the 2 way

notowen333: DS=data structure which specifically I mean `MessageTypeFilter`

strands-ts/src/conversation-manager/agentic/agentic-context.ts
notowen333: its now just walking backwards until we hit a user text block

strands-ts/src/conversation-manager/compression/index.ts
notowen333: Resolved — deleted this barrel file entirely. Nothing was importing from it (all consumers import directly from the specific files). No agentic/compression internals are on the public package surface.

strands-ts/src/agent/agent.ts
notowen333: spoke offline that we'll skip tool based offloading setting for now.

notowen333: Also adding higher agentic mode default as discussed 

8000

strands-ts/src/conversation-manager/modes/agentic/agentic-context.ts
notowen333: I added finer grained options for pinning, but I think for P0 we can leave this as is for now. There are some tricky considerations like user/assisstant message pairs that we'd introduce

lizradway: agreed, delay until post summit

strands-ts/src/conversation-manager/compression/context-compression.ts
lizradway: should we then move `MessageTypeFilter` to the conversation manager level (since it will be used outside of agentic mode ideally) and move this to this `pinMessages` in `pinMessages.ts` and use it everywhere?

notowen333: Also discussed offline.  Because we're not hardened on what the filter is going to be and we have the luxury of all internal definitions for now, we'll keep it as is

strands-ts/src/conversation-manager/modes/agentic/agentic-context.ts
```diff
@@ -0,0 +1,256 @@
...
+/** Default number of recent messages to preserve verbatim during summarization or truncation. */
+const DEFAULT_KEEP_RECENT_MESSAGES = 10
+/** Default fraction of oldest messages to fold into the summary. */
+const DEFAULT_SUMMARY_RATIO = 0.3
+/** Minimum allowed summary ratio (prevents near-zero compression). */
+const MIN_SUMMARY_RATIO = 0.1
+/** Maximum allowed summary ratio (prevents summarizing nearly everything). */
+const MAX_SUMMARY_RATIO = 0.8
+/** Minimum conversation length required before any compression operation can run. */
+const MIN_MESSAGES_FOR_OPERATION = 2
```
lizradway: non blocking nit: `MIN_MESSAGES_FOR_COMPRESSION`

strands-ts/src/conversation-manager/modes/agentic/agentic-context.ts
```diff
@@ -0,0 +1,256 @@
...
+      '"messages" targets only non-tool messages, "all" (default) targets everything.'
+  )
+
+/**
+ * Partition messages in [0, rangeEnd) into three buckets:
+ * - pinned: protected from eviction (includes tool-pair partners)
+ * - eligible: unpinned and matching the filter (will be evicted)
+ * - skipped: unpinned but not matching the filter (preserved in place)
+ */
+function partitionByFilter(
```
lizradway: leaving comment for post summit we should move this outside of just agentic to recycle it

strands-ts/src/conversation-manager/modes/agentic/agentic-context.ts
```diff
@@ -0,0 +1,256 @@
...
+    const preserveRecent = keepRecent ?? DEFAULT_KEEP_RECENT_MESSAGES
+    const ratio = Math.max(MIN_SUMMARY_RATIO, Math.min(MAX_SUMMARY_RATIO, summaryRatio ?? DEFAULT_SUMMARY_RATIO))
+
+    let splitPoint = Math.max(1, Math.floor(messages.length * ratio))
+    splitPoint = Math.min(splitPoint, messages.length - preserveRecent)
+    if (splitPoint <= 0) {
+      return `No summarization performed: not enough eligible messages to compress (conversation has ${originalMessageCount} messages, preserving recent ${preserveRecent}).`
+    }
+
+    splitPoint = adjustSplitPointForToolPairs(messages, splitPoint)
```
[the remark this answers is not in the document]
lizradway: +1

strands-ts/src/conversation-manager/modes/agentic/agentic-context.ts
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    let summaryMessage
+    try {
+      summaryMessage = await generateSummary(eligible, agent.model)
+    } catch {
+      return 'Summarization failed: no response from model.'
+    }
+
+    messages.splice(0, splitPoint, ...pinned, ...skipped, summaryMessage)
```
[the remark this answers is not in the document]
lizradway: > A test exercising a pinned toolUse whose toolResult falls in the eligible set under messageType: 'tools'/'messages' would lock this down.

this could be nice

strands-ts/src/conversation-manager/modes/agentic/agentic-context.ts
```diff
@@ -0,0 +1,256 @@
...
+          .literal('last_turn')
+          .describe('Select messages from the current turn (everything since the last user request).'),
+        z.number().int().min(1).describe('Select the last N messages from the conversation.'),
+        z.array(z.number().int().min(0)).min(1).describe('Select messages at specific zero-based indices.'),
+      ])
+      .describe(
+        'Which messages to target. "last_turn" for the current exchange, a number for the last N messages, or an array of indices.'
+      ),
+    filter: z
+      .enum(['user', 'assistant', 'tool_use', 'tool_result'])
```
lizradway: nonblocking nit:  pinning `tool_use` or `tool_result` alone is practically the same since the pair won't be trimmed

strands-ts/src/conversation-manager/modes/agentic/index.ts
```diff
@@ -0,0 +1,2 @@
+export { summarizeContextTool, truncateContextTool, pinContextTool } from './agentic-context.js'
```
lizradway: womp womp womp

strands-ts/src/conversation-manager/modes/agentic/token-usage-middleware.ts
```diff
@@ -0,0 +1,38 @@
+import type { InvokeModelContext } from '../../../middleware/stages.js'
+import type { MiddlewareInputHandler } from '../../../middleware/types.js'
+import type { Model } from '../../../models/model.js'
+import { Message, TextBlock } from '../../../types/messages.js'
+import { DEFAULT_CONTEXT_WINDOW_LIMIT } from '../../conversation-manager.js'
+
+export function createTokenUsageMiddleware(model: Model): MiddlewareInputHandler<InvokeModelContext> {
```
lizradway: nit: suspect of a single file for one function. i find the name `createTokenUsageMiddleware` really hard to read, but don't have any better options rn i can think of?

strands-ts/src/conversation-manager/index.ts
```diff
@@ -19,4 +19,3 @@ export {
   SummarizingConversationManager,
   type SummarizingConversationManagerConfig,
 } from './summarizing-conversation-manager.js'
-export { isPinned, pinMessage, unpinMessage } from './pin-message.js'
```
lizradway: might need to rebase this, oopsi

strands-ts/src/index.ts
```diff
@@ -268,7 +268,6 @@ export {
   SummarizingConversationManager,
   type SummarizingConversationManagerConfig,
 } from './conversation-manager/summarizing-conversation-manager.js'
-export { isPinned, pinMessage, unpinMessage } from './conversation-manager/pin-message.js'
```
zastrowm: Is this technically a breaking change?

notowen333: It is but it's from another PR it won't be in this diff on rebase

strands-ts/src/conversation-manager/modes/agentic/agentic-context.ts
notowen333: makes sense will just switch to 'tools'

agent-of-mkmeral:
## Adversarial review — verified locally

Built the PR branch and ran the suites: **build passes, 177/177 unit tests pass** (`unit-node`, conversation-manager). I then wrote adversarial tests against the three tools directly. Two findings are real bugs with concrete repros; one earlier concern can be partially closed.

---

### 🔴 P1: `truncate_context` drops messages even when `keepRecent` > conversation length

**Repro:** 6-message conversation, `truncate_context({ keepRecent: 10 })` → `"Dropped 2 message(s). 4 remaining."`

```ts
const startIndex = messages.length <= windowSize ? MIN_MESSAGES_FOR_OPERATION : messages.length - windowSize
```

The `messages.length <= windowSize → start at 2` fallback was inherited from `SlidingWindowConversationManager.reduceContext`, where forced progress is mandatory (it only runs on overflow). For a **voluntary, model-invoked** tool it contradicts the schema description — *"Number of most recent messages to keep. Everything older (and unpinned) is dropped."* With 6 messages and `keepRecent: 10`, nothing is "older" — the model asked to keep more than exists and silently lost its 2 oldest messages anyway.

**Fix:** return a no-op status when `messages.length <= windowSize`.

---

### 🔴 P1: reorder splice can corrupt `agent.messages` into a Bedrock-invalid conversation

The `messages.splice(0, splitPoint, ...pinned, ...skipped, summaryMessage)` pattern reorders history. Two concrete repros:

**Repro 1 (pin + truncate):** pin an assistant message at index 1, then `truncate_context({ keepRecent: 2 })` → resulting roles: `assistant, user, assistant`. **Conversation now starts with an assistant message.**

**Repro 2 (filter + summarize):** history `[user, toolUse, toolResult, assistant, ...8 more]`, `summarize_context({ keepRecent: 8, summaryRatio: 0.4, messageType: 'messages' })` → resulting roles: `assistant(toolUse), user(toolResult), user(summary), user, assistant, ...`. Starts with assistant **and** has 3 consecutive user messages.

Bedrock Converse rejects conversations that don't start with a user message. Because the splice mutates `agent.messages` **in place and persistently**, the very next invoke fails — the agent bricks its own session with no recovery path. Note the tool descriptions actively steer the model toward the filtered modes (*"Often most useful with messageType…"*), which is exactly the path that populates the `skipped` bucket and triggers reordering.

**Suggested fixes:**
1. Evict eligible messages **in place** (preserve original relative order of pinned/skipped) and insert the summary at the position of the first evicted message, rather than re-bucketing to the front.
2. After any compression, validate `messages[0].role === 'user'` (walk forward / drop leading assistant content if not) before returning.
3. Re: the earlier thread on pin filters — restricting default pinning to user messages would also shrink this surface, since pinned assistant messages are repro 1's trigger.

*(The same latent reorder exists in the pre-existing managers' `partitionPinned` splice, but the agentic tools make it far more reachable since the model pins ad-hoc.)*

---

### ✅ Partially closing the earlier orphaning concern

I tried to orphan a tool pair via `messageType` filters (pin only the `toolUse`, truncate with `messageType: 'tools'`) and **could not**: `isPinned`'s partner protection pulls both sides of an adjacent pair into the pinned bucket, and both sides of a pair always match the same `tools`/`messages` filter bucket, so they never split. Boundary guards (`findValidTrimPoint`, `adjustSplitPointForToolPairs`) held in all my attempts. Pairs stay **intact** — they just get *reordered* (see P1 above). A regression test asserting pair adjacency under filters would still be worth adding.

---

### 🟡 Smaller items

- **`generateSummary` catch swallows the cause** — `catch { return 'Summarization failed: no response from model.' }` misreports throttling/auth/network errors as "no response from model", which will mislead the model's retry behavior. Include the error message.
- **`pin_context(select: [indices])` is unanchorable** — `<context-status>` reports only token counts; the model has no reliable way to know zero-based indices of its own history (especially after prior compressions shift everything). Consider including a message count in `<context-status>`, or dropping index-based selection until there's an anchoring mechanism.
- **Silent degradation when token estimation fails** — `projectedInputTokens === undefined` → middleware no-ops → no `<context-status>` at all, and agentic mode loses its entire signal without any log.
- Confirmed via grep that `summarizeMessages`/`trimMessages` in `context-compression.ts` are imported **only by tests** — the dead-code comment from the earlier review still stands.

---

### 👍 What holds up well

- `<context-status>` is injected into the deep-copied per-call messages, so it never accumulates in persisted history — clean design.
- Tool-pair partner protection in `isPinned` is correct in both directions (toolUseId-matched).
- The 8k agentic offload threshold + reactive-only safety-net manager is a sensible split vs `auto`.
- `last_turn` walk-back behaves correctly (pins the initiating user message through the in-flight call).

The two P1s are mechanical fixes; everything else can trail. Happy to share the adversarial test file if useful.

strands-ts/src/conversation-manager/modes/agentic/token-usage-middleware.ts
notowen333: moved into the main file

notowen333:
@agent-of-mkmeral can you re-review after the most recent changes

agent-of-mkmeral:
## Re-review of head `d6c2783` — 2 integ test failures in this PR's own test file ⚠️

No new commits since my last review, but the **integration test results landed after it** (22:57Z) and surfaced 2 failures in `test/integ/conversation-manager/agentic-context.test.ts` (failing on both `integ-node` and `integ-browser`, with retry — so not flakes). I traced both to root cause against the head source. Good news: **neither is a source bug** — both are this PR's integ tests being out of sync with intentional changes made during the review cycle.

<details>
<summary>Failure 1: <code>context-status contains token numbers and tool names</code></summary>

```
AssertionError: expected '\n\n<context-status>\n<used>2,826 / 1…' to contain '<tools>'
```

The test (L103-106) still asserts `<tools>`, `summarize_context`, `truncate_context`, and `pin` appear in the injected `<context-status>` block. But the `<tools>` row was **intentionally removed** from the status text during this review (per the thread with @lizradway — *"Yes. We can kill that row though"*). The source at `agentic-context.ts:189-193` now only emits `<used>` and `<remaining>`.

**Fix**: drop the four stale assertions (L103-106); keep `<used>`, `<remaining>`, `</context-status>`.

</details>

<details>
<summary>Failure 2: <code>summarize_context actually replaces older messages with a summary</code></summary>

```
AssertionError: expected true to be false  // hasOriginalSkyMsg
```

The test seeds 6 exchanges starting with *"The sky is blue."*, asks the model to summarize with `keepRecent=2`, then asserts the sky message is gone. But the `collectPreserved` fix for the ordering blocker **always preserves the first user message** (`foundFirstUser`, `agentic-context.ts:52-53`) to guarantee Bedrock's user-first invariant — and *"The sky is blue."* **is** the first user message, so it's now always kept by design. The test assertion predates that fix.

**Fix**: assert on a non-first old message instead (e.g. *"Grass is green."* / *"Water is wet."*), and/or assert a summary message now exists. 

*Side note (non-blocking)*: this also means the first user message is never compressible via `summarize_context`, even under `messageType: "all"`. That's a reasonable trade for provider validity, but worth a one-line mention in the tool description or a code comment so it doesn't read as a bug later.

</details>

<details>
<summary>Other CI failures — pre-existing / unrelated</summary>

- **NPM Security Audit**: `esbuild ≤0.28.0` high-severity advisory (GHSA-gv7w-rqvm-qjhr) in dev deps — repo-wide, not introduced here.
- **OpenAI integ tests**: `429 insufficient_quota` on the OpenAI account — infra, affects all OpenAI-model suites.
- **label-size**: size-labeler hiccup, unrelated to content.

</details>

**Assessment**: Request Changes (narrowly). All source-level findings from prior rounds remain resolved and the design holds up — these are two ~3-line test updates to bring the PR's own integ tests in line with behavior that was deliberately changed mid-review. Once those are synced, this is good to go from my side.

notowen333:
@agent-of-mkmeral re-review after most recent changes. Make sure to go over the entire source change including verification that the existing Conversation Managers are unaffected by refactor

agent-of-mkmeral:
## Full re-review of `e32e88f` — refactor verified, existing managers unaffected ✅

Per @notowen333's ask, I cloned the head branch (`e32e88f`, confirmed = PR head), diffed the **entire source change against the merge-base** (`5d8c219`), and specifically audited that the pre-existing `SlidingWindowConversationManager` and `SummarizingConversationManager` are behaviorally unchanged by the `compression/` extraction. Built, typechecked, and ran the suites locally.

### 🔍 Refactor audit — are the existing managers affected?

**Verdict: No behavioral change. The refactor is a pure extract-and-delegate.**

<details>
<summary><code>SlidingWindowConversationManager</code> (+4 / −40) — verified equivalent</summary>

- The inline trim-point `while` loop was lifted verbatim into `findValidTrimPoint(messages, startIndex)` in `compression/context-compression.ts`. I diffed the two byte-for-byte: identical guards (must be `user` role → not orphaned `toolResult` → `toolUse` only valid if next msg has its `toolResult`), identical `continue`/`break` flow.
- The `startIndex` computation is unchanged: `messages.length <= windowSize ? 2 : messages.length - windowSize` — same `2` fallback, same overflow contract. (This is the overflow-driven path, where forced progress is *correct* — distinct from the agentic tool's voluntary path I flagged earlier.)
- Only other change: import path `./pin-message.js` → `./compression/pin-message.js`.
- **53/53 unit tests pass** (tool-pair validation, pinFirst, no-valid-trim-point propagation).

</details>

<details>
<summary><code>SummarizingConversationManager</code> (+8 / −119) — verified equivalent</summary>

- Two private methods (`_adjustSplitPointForToolPairs`, `_generateSummary`) and the `DEFAULT_SUMMARIZATION_PROMPT` constant were moved into `compression/context-compression.ts` as exported functions and are now *called* rather than defined inline. I diffed the moved bodies — logic is identical (same tool-pair walk, same manual `stream.next()` loop with the `done`/`value` guard, same user-role wrapping of the summary).
- **One subtle point I checked closely**: the old `_generateSummary` passed `this._summarizationSystemPrompt` directly; the new `generateSummary(toSummarize, model, this._summarizationSystemPrompt)` does `systemPrompt ?? DEFAULT_SUMMARIZATION_PROMPT` internally. Since the constructor sets `_summarizationSystemPrompt = config?.summarizationSystemPrompt ?? DEFAULT_SUMMARIZATION_PROMPT` (always a non-undefined `string`), the new `??` **never fires** — behavior is identical for both default and custom prompts. ✓
- `splice(0, count, ...protectedToPreserve, summaryMessage)` path is unchanged.
- **18/18 unit tests pass** (tool-pair adjustment throw, model-error wrapping with overflow cause, pinFirst-protected ranges).

</details>

<details>
<summary><code>pin-message.ts</code> rename — 99% similar, no logic change</summary>

Git reports the file as RENAMED `pin-message.ts` → `compression/pin-message.ts` with **similarity 99%** — the only delta is the relative import depth (`../types` → `../../types`). `isPinned` partner protection is byte-identical. The 16 `pin.test.ts` cases pass.

</details>

### 🟡 One thing worth a glance: removed public exports (non-blocking)

`src/index.ts` and `conversation-manager/index.ts` drop the top-level `export { isPinned, pinMessage, unpinMessage }`. On the **merge-base** these were public API, so in isolation this reads as a breaking change. **I confirmed it's a stale-base artifact, not real**: the current `upstream/main` tip *already* removed those top-level exports (via another merged PR), so after rebase this diff disappears — exactly as you told @zastrowm earlier. No action needed; flagging only so it's on record that I checked it against live `main` rather than taking it on faith.

### ✅ Integ test sync (`e32e88f`) confirmed correct

The two failures I flagged last round are properly resolved — the diff matches my suggested fixes exactly:
- Removed the stale `<tools>` / tool-name assertions (test renamed to *"…token usage and remaining capacity"*).
- Summarize test now asserts *"Grass is green."* (2nd user msg, genuinely compressible) is dropped instead of *"The sky is blue."* (1st user msg, always preserved by `collectPreserved`). Verified the seed order makes this a valid test.

### 🧪 Local verification

- `tsc -p src/tsconfig.json --noEmit` → **0 errors**.
- **195/195** unit tests pass across the touched area (conversation-manager × 9 files + copy-on-input). Full `unit-node` run: **3455 pass**; the only 4 failures are in `telemetry/config.test.ts` (default `service.name` / `OTEL_RESOURCE_ATTRIBUTES`) — **false positives from OTEL env vars set in my sandbox**, in files this PR doesn't touch. CI (clean env) shows these green.
- Integ suite needs live Bedrock/OpenAI creds (skipped locally); the two previously-failing cases are verified by static analysis against head source as above. The remaining CI reds (esbuild npm-audit advisory, OpenAI `429 insufficient_quota`, label-size) are pre-existing/infra, unrelated to this change.

### Assessment: **Approve** ✅

The refactor cleanly unifies the compression/trim/summary helpers without altering the behavior of either existing conversation manager — verified by source diff *and* their full unit suites. All findings from every prior review round are closed. Nice work seeing this all the way through. 🎉

lizradway:
need to merge from main fyi for npm

strands-ts/src/conversation-manager/index.ts
lizradway: i think you still need to rebase this

agent-of-mkmeral:
## Full re-review of `712bf0b` — refactor verified against live source, existing managers unaffected ✅

@notowen333 per your ask, I freshly cloned the head (`712bf0b`, confirmed = PR head), diffed the **entire source change against the real merge-base** (`c09b551`), specifically audited that `SlidingWindowConversationManager` and `SummarizingConversationManager` are behaviorally unchanged, and ran build + typecheck + the full conversation-manager/middleware suites locally on the rebased tree.

### 🔍 Are the existing Conversation Managers affected by the refactor?

**Verdict: No behavioral change. Pure extract-and-delegate, verified at source level + full unit suites.**

<details>
<summary><code>SlidingWindowConversationManager</code> (+4/−40) — verified equivalent</summary>

- The inline trim-point `while` loop was lifted into `findValidTrimPoint(messages, startIndex)` in `compression/context-compression.ts`. I diffed both: **identical** guards (must be `user` role → not orphaned `toolResult` → `toolUse` only valid if next msg carries its `toolResult`), identical `continue`/`break` flow. The only delta is a variable rename (`oldestMessage` → `message`).
- `startIndex` is unchanged: `messages.length <= windowSize ? 2 : messages.length - windowSize` — same `2` overflow fallback, same contract (this is the overflow-driven path where forced progress is *correct*, distinct from the agentic tool's voluntary path).
- Only other change: import path `./pin-message.js` → `./compression/pin-message.js`.
- **53/53 unit tests pass.**

</details>

<details>
<summary><code>SummarizingConversationManager</code> (+8/−119) — verified equivalent</summary>

- `_adjustSplitPointForToolPairs`, `_generateSummary`, and `DEFAULT_SUMMARIZATION_PROMPT` moved into `compression/context-compression.ts` as exported functions, now *called* rather than defined inline. Diffed the moved bodies — logic is byte-identical (same tool-pair walk + throw, same manual `stream.next()` loop with the `done`/`value` guard, same user-role wrapping).
- **Subtle point I re-checked:** the extracted `generateSummary` does `systemPrompt ?? DEFAULT_SUMMARIZATION_PROMPT` internally, but the constructor sets `_summarizationSystemPrompt = config?.summarizationSystemPrompt ?? DEFAULT_SUMMARIZATION_PROMPT` (always a non-undefined string), so the new `??` **never fires** — identical behavior for both default and custom prompts. ✓
- `splice(0, count, ...protectedToPreserve, summaryMessage)` path unchanged.
- **18/18 unit tests pass.**

</details>

<details>
<summary><code>pin-message.ts</code> rename + <code>agent.ts</code> wiring</summary>

- `pin-message.ts` → `compression/pin-message.ts`, git similarity **99%** — only delta is import depth (`../types` → `../../types`). `isPinned` partner protection byte-identical; 16 `pin.test.ts` cases pass.
- `agent.ts`: every new code path is gated behind `contextManager === 'agentic'` (tool registration, token-usage middleware, the 8k agentic offload threshold). The `undefined` default (SlidingWindow) and `'auto'` (Summarizing + offloader) paths are untouched. `conversation-manager.ts` change is just exposing `DEFAULT_CONTEXT_WINDOW_LIMIT` as `@internal export` for the new module.

</details>

### 📍 Rebase / branch state (note for @lizradway)

- @lizradway's most recent comment (16:20Z, *"you still need to rebase this"* on `conversation-manager/index.ts`) was posted **~7 min before** the 16:27Z rebase landed. I confirmed that file is now **byte-identical to `main`** — `git diff pr origin/main -- conversation-manager/index.ts` is empty. The rebase resolved it; the earlier "breaking export" concern is fully gone against the real base.
- One new wrinkle: `main` advanced again after the rebase (PR is now **1 commit behind** — `bdb059a`). That commit is **docs-only** (`context-offloader.mdx`/`.ts`), touches **zero** `conversation-manager` files, so no conflict — just an optional final `merge from main` before the npm publish lizradway flagged.

### 🧪 Local verification (rebased head)

- `npm ci` → **0 vulnerabilities**.
- `tsc -p src/tsconfig.json --noEmit` → **0 errors**.
- `npm run build` → clean.
- **295/295** unit tests pass (13 files: conversation-manager ×9 + middleware ×4), including the integ-test-sync from the prior commit.
- Full PR diff vs base intact: 16 files, +1620/−179 — same well-reviewed change set.
- Remaining CI reds noted in earlier rounds (esbuild npm-audit advisory, OpenAI `429 insufficient_quota`, label-size) are pre-existing/infra, unrelated to this change.

### Assessment: **Approve** ✅

The refactor is a clean unification of the compression/trim/summary helpers with **no behavioral change to either existing conversation manager** — verified by source diff *and* their full unit suites. Agentic mode is fully isolated behind the strategy flag. All findings from every prior round remain closed. Only optional follow-up: a final `merge from main` to pick up the 1 docs commit before npm publish. Nice work seeing this through. 🎉

lizradway (approved):

PR #2755, opened 2026-06-12
ci: require 2 maintainer reviews on bot/agent PRs

JackYPCOnline:
## Description

As more PRs are opened by bots and coding agents, we want extra human oversight before their changes merge. This adds a GitHub Actions gate that requires **2 maintainer approvals** before a bot/agent-authored PR targeting `main` can merge.

**Human-authored PRs are never affected** — when a PR isn't detected as bot/agent, the job exits early with a passing check and posts no comment, so it's safe to make this a required status check.

Detection uses structured, low-false-positive signals: GitHub `user.type`, the reserved `[bot]` username/email suffix, `performed_via_github_app`, explicit automation labels (`bot`, `agent`, `dependencies`, ...), and anchored patterns for named bot services (dependabot/renovate/etc.). Maintainers are resolved with `repos.getCollaboratorPermissionLevel` (role `maintain`/`admin`), mirroring the existing `authorization-check` pattern so the default `GITHUB_TOKEN` is sufficient — no org-team read scope required.

When a bot/agent PR is gated, the workflow posts a single sticky status comment explaining why and how many maintainer approvals remain. It re-runs on `pull_request_review` (submitted/dismissed), so it clears automatically once the second approval lands; if a PR was flagged only by a label and that label is later removed, the stale comment is cleaned up.

## Enforcement (required setup)

`core.setFailed` produces a failing check but does not block merge on its own. To enforce the gate, **Bot/Agent PR Review Requirement** must be added as a required status check in the `main` branch protection ruleset. Until then this runs in report-only mode.

## Type of Change

Other (please describe): CI / repository automation

## Testing

The `github-script` logic was validated locally against mocked GitHub APIs across the detection signals and approval-counting rules: human PRs unaffected, private-email humans not flagged, non-maintainer approvals ignored, `COMMENTED` not revoking an approval, commit pagination, permission-API errors, and the de-flag comment cleanup. Logic is kept inline in the workflow to match the repo's other `github-script` workflows (`auto-close`, `pr-size-labeler`, `auto-strands-review`).

- [ ] I ran `hatch run prepare` (N/A — workflow YAML only)

## Checklist


.github/workflows/bot-pr-review-gate.yml
```diff
@@ -0,0 +1,178 @@
...
+              /^renovate/i,
+              /^snyk/i,
+              /^codecov/i,
+              /^mergify/i,
+              /^greenkeeper/i,
+              /agent$/i,
+              /-agent/i,
+              /agent-/i,
+              /'s[ -]agent/i,
+              /automation/i,
```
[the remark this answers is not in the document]
JackYPCOnline: Fixed in dbb2714. Dropped the generic fragments (`bot$`, `agent`, `-agent`, `agent-`, `automation`, `ci-`, `cd-`) that matched human handles like `Talbot` or `*-agent`. Username matching now only uses anchored, named bot accounts plus the reserved `[bot]` suffix. Agent/automation PRs that don't use a `[bot]` account are caught via `user.type`, `performed_via_github_app`, or an explicit `agent`/`automated` label instead of fuzzy handle matching.

.github/workflows/bot-pr-review-gate.yml
```diff
@@ -0,0 +1,178 @@
...
+              /^ci-/i,
+              /^cd-/i,
+            ];
+
+            // --- Detection: known bot email patterns in commits ---
+            const botEmailPatterns = [
+              /\[bot\]@/i,
+              /noreply@github\.com$/,
+              /@users\.noreply\.github\.com$/,
```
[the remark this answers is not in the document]
JackYPCOnline: Agreed — this was the critical one. Removed both `noreply@github.com$` and `@users.noreply.github.com$` since that's GitHub's default privacy email for **human** accounts. Email detection now matches only the reserved `[bot]@` shape, which still catches the GitHub Actions / dependabot bot addresses (e.g. `…+github-actions[bot]@users.noreply.github.com`) without gating ordinary contributors. Fixed in dbb2714.

.github/workflows/bot-pr-review-gate.yml
```diff
@@ -0,0 +1,178 @@
...
+            if (pr.performed_via_github_app) {
+              reasons.push(`PR created via GitHub App "${pr.performed_via_github_app.name}"`);
+            }
+
+            // Check 5: Commit author emails
+            const { data: commits } = await github.rest.pulls.listCommits({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: pr.number,
+              per_page: 5,
```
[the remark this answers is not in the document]
JackYPCOnline: Fixed in dbb2714. Switched to `github.paginate(github.rest.pulls.listCommits, {...})` so every commit is scanned, consistent with how reviews are already paginated. The arbitrary `per_page: 5` cap is gone.

.github/workflows/bot-pr-review-gate.yml
```diff
@@ -0,0 +1,178 @@
...
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: pr.number,
+            });
+
+            // Only count the latest review per user.
+            const latestReviewByUser = new Map();
+            for (const review of reviews) {
+              if (!review.user) continue;
+              latestReviewByUser.set(review.user.login, review.state);
```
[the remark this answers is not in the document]
JackYPCOnline: Fixed in dbb2714, using your suggested approach — `COMMENTED` reviews are skipped when building the latest-state map, so a maintainer's `APPROVED` is no longer overwritten by a follow-up comment. `CHANGES_REQUESTED` and `DISMISSED` still correctly revoke it.

JackYPCOnline:
Thanks for the thorough review — pushed fixes in dbb2714. Summary of how each theme was handled:

**Detection accuracy** — Dropped the broad username fragments and both `noreply` email patterns that matched humans. Detection now leans on structured signals (`user.type`, `[bot]` suffix, `performed_via_github_app`, automation labels) plus anchored named-bot patterns. This restores the "humans unaffected" guarantee.

**Approval correctness** — `COMMENTED` reviews are now ignored, so a maintainer's `APPROVED` isn't silently revoked by a follow-up comment (matches GitHub's own merge semantics).

**Commit coverage** — The commit-email scan is now fully paginated instead of capped at 5.

**Enforcement** — Documented in the PR description that the check must be added as a required status check on `main` branch protection to actually block merges. Also added a single idempotent status comment so a gated author sees why and how many approvals remain.

**On extracting the logic to a committed script + tests:** I tried this first (commit d40d2270) but reverted it. Every other `github-script` workflow in this repo — `auto-close.yml`, `pr-size-labeler.yml`, `auto-strands-review.yml` — keeps its logic inline in the YAML; there's no committed `.cjs`/`.mjs` + `require` pattern here (the `authorization-check` require-pattern lives in the separate `devtools` repo). Introducing it in this one PR would be inconsistent, and the `require` approach needs an `actions/checkout` step on a `pull_request_target` job — which widens the surface area, the opposite of the read-only posture flagged below. So I kept the logic inline to match convention and validated it locally against mocked APIs instead. Happy to standardize testable workflow scripts as a separate repo-wide change if the team wants that.

**Safety** — Confirmed: the job only reads PR metadata, counts reviews, and posts a status comment. It does not check out or execute PR head code, and there's no plan to.

.github/workflows/bot-pr-review-gate.yml
```diff
@@ -0,0 +1,211 @@
+name: Bot PR Review Gate
+
+on:
+  pull_request_target:
+    branches: [main]
+    types: [opened, synchronize, reopened, ready_for_review, labeled, unlabeled]
```
zastrowm: We don't need labeled/unlabled I think?

.github/workflows/bot-pr-review-gate.yml
```diff
@@ -0,0 +1,211 @@
...
+  pull_request_review:
+    types: [submitted, dismissed]
+
+# pull_request_target runs in the base-repo context with a write-capable token.
+# This job only reads PR metadata, counts reviews, and posts a status comment —
+# it never checks out or executes PR head code.
+permissions:
+  contents: read
+  pull-requests: write
```
zastrowm: I don't think we need to post a comment - let's just have it as a status check.

If we change the name to be "2 Approvers for Bots", I think it'd be obvious what a failure means.

.github/workflows/bot-pr-review-gate.yml
```diff
@@ -0,0 +1,211 @@
...
+
+            const author = pr.user.login;
+            const authorType = pr.user.type;
+
+            // Anchored, named bot accounts plus the `[bot]` suffix GitHub reserves
+            // for app accounts. Generic fragments (bot$, agent, automation) were
+            // intentionally dropped — they match human handles (Talbot, *-agent).
+            const botUsernamePatterns = [
+              /\[bot\]$/,
+              /^dependabot(\[bot\])?$/i,
```
zastrowm: Is this included or excluding dependabot. IMHO dependabot should not require 2 approvers

.github/workflows/bot-pr-review-gate.yml
```diff
@@ -0,0 +1,211 @@
...
+            const requiredApprovals = parseInt(process.env.REQUIRED_APPROVALS, 10);
+            const maintainerRoles = process.env.MAINTAINER_ROLES
+              .split(',').map(r => r.trim()).filter(Boolean);
+
+            const author = pr.user.login;
+            const authorType = pr.user.type;
+
+            // Anchored, named bot accounts plus the `[bot]` suffix GitHub reserves
+            // for app accounts. Generic fragments (bot$, agent, automation) were
+            // intentionally dropped — they match human handles (Talbot, *-agent).
```
zastrowm: Does this mean that we won't detect "agent-of" patterns or is that matched somehow else?

.github/workflows/bot-pr-review-gate.yml
JackYPCOnline: Good call out, updated.

.github/workflows/bot-pr-review-gate.yml
JackYPCOnline: Fresh eyes!!! I missed that line...

zastrowm (approved):

PR #2760, opened 2026-06-12
feat: add internal middleware system for InvokeModelStage

### LINKED ISSUE #2735: Middleware - Python
Tracking ticket for Middleware in Python

zastrowm:
## Description

Begins the port of the TypeScript middleware system (#2681) to Python, scoped to a private `InvokeModelStage` so that we can unblock internal work that needs to modify input calls.

The behavioral spec is defined in `strands-ts/src/middleware/README.md` and the Python implementation documents its intentional divergences in `strands-py/src/strands/_middleware/README.md`.

Resolves: #2735

## Behavioral change: model span timing

The model invoke OTEL span now starts inside the middleware terminal (after hooks and middleware transforms) rather than before `BeforeModelCallEvent`. The span measures actual model latency more accurately — hook execution time is no longer included. Span lifecycle is preserved for success, retry, and error paths. On the cancel path (BeforeModelCallEvent.cancel), no span is emitted because the model is never invoked — this is intentionally more correct than the previous behavior which created and immediately closed a span for a call that never happened.

## `system_prompt` unification and `split_system_prompt`

The Python SDK previously passed `system_prompt` (str) and `system_prompt_content` (list of blocks) as two separate fields everywhere — the Agent stored both, `stream_messages()` accepted both, and each model provider duplicated the same normalization logic.

For middleware, exposing two fields for the same concept didn't make sense. `InvokeModelContext` uses a single union field `system_prompt: str | list[SystemContentBlock] | None` (matching TypeScript's `SystemPrompt` type). At the terminal boundary where the model is actually invoked, we call `split_system_prompt()` to decompose it back into the two-param form that `Model.stream()` expects.

`split_system_prompt()` is extracted from `Agent._initialize_system_prompt` into `types/content.py` as a shared utility. The Agent now delegates to it directly.

## Related Issues

- #2735 (Middleware — Python)
- https://github.com/strands-agents/harness-sdk/pull/2681

## Documentation PR

None

## Type of Change

New feature

## Testing

## Checklist

----

### DIFF
--- a/strands-py/src/strands/_middleware/README.md
+++ b/strands-py/src/strands/_middleware/README.md
@@ -1,0 +1,37 @@
+# Python Middleware
+
+This implementation follows the behavioral spec defined in `strands-ts/src/middleware/README.md` with the following intentional divergences:
+
+## Scope
+
+Only `InvokeModelStage` is implemented. `ExecuteToolStage` and `AgentStreamStage` will be added as needed.
+
+## Result encoding
+
+Python async generators cannot `return` values. The implementation uses a `MiddlewareResult` sentinel yielded as the last item in the generator. Integration sites strip it from the event stream. Pass-through is:
+
+```python
+async def passthrough(context, next_fn):
+    async for event in next_fn(context):
+        yield event
+```
+
+## No removal / cleanup
+
+Once registered, middleware cannot be removed. This matches the Python hook system which also does not support removal.
+
+## Private module
+
+The `_middleware/` package is not part of the public API. Internal consumers access it via `agent._middleware_registry.add_middleware(...)`.
+
+## System prompt as a union type
+
+`InvokeModelContext.system_prompt` is `str | list[SystemContentBlock] | None` (a single union field). The terminal decomposes this into the two-param form needed by `Model.stream()` via `split_system_prompt()`.
+
+## Defensive copies
+
+Context fields (`messages`, `system_prompt`, `tool_specs`, `tool_choice`) are deep-copied when building the middleware context. `invocation_state` is shared by reference. `model_state` is not on the context — the terminal reads it directly from the agent.
+
+## Generator cleanup
+
+Python's `compose()` uses `try/finally` with explicit `aclose()`. TypeScript relies on `yield*` delegation which calls `.return()` automatically. Both correctly clean up generators.
--- a/strands-py/src/strands/_middleware/stages.py
+++ b/strands-py/src/strands/_middleware/stages.py
@@ -9,7 +9,8 @@ from .types import MiddlewareStage
 
 if TYPE_CHECKING:
     from ..agent.agent import Agent
-    from ..types.content import Messages, SystemPrompt
+    from ..types.content import Message, Messages, SystemPrompt
+    from ..types.event_loop import Metrics, Usage
     from ..types.streaming import StopReason
     from ..types.tools import ToolSpec
 
@@ -36,9 +37,9 @@ class InvokeModelResult:
     """Result from InvokeModelStage middleware."""
 
     stop_reason: StopReason
-    message: dict[str, Any]
-    usage: dict[str, Any]
-    metrics: dict[str, Any]
+    message: Message
+    usage: Usage
+    metrics: Metrics
 
 
 InvokeModelStage: MiddlewareStage[InvokeModelContext, InvokeModelResult, Any] = MiddlewareStage(name="invokeModel")
--- a/strands-py/src/strands/_middleware/stages.py
+++ b/strands-py/src/strands/_middleware/stages.py
@@ -1,0 +1,44 @@
+"""Built-in middleware stages and their context/result types."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any
+
+from .types import MiddlewareStage
+
+if TYPE_CHECKING:
+    from ..agent.agent import Agent
+    from ..types.content import Messages, SystemPrompt
+    from ..types.streaming import StopReason
+    from ..types.tools import ToolSpec
+
+
+@dataclass
+class InvokeModelContext:
+    """Context passed to InvokeModelStage middleware.
+
+    All collection fields (messages, system_prompt, tool_specs, tool_choice) are
+    defensive copies — middleware cannot accidentally mutate agent state.
+    invocation_state is shared by reference (hooks and tools write to it during streaming).
+    """
+
+    agent: Agent
+    messages: Messages
+    system_prompt: SystemPrompt
+    tool_specs: list[ToolSpec]
+    tool_choice: Any | None
+    invocation_state: dict[str, Any]
+
+
+@dataclass
+class InvokeModelResult:
+    """Result from InvokeModelStage middleware."""
+
+    stop_reason: StopReason
+    message: dict[str, Any]
+    usage: dict[str, Any]
+    metrics: dict[str, Any]
+
+
+InvokeModelStage: MiddlewareStage[InvokeModelContext, InvokeModelResult, Any] = MiddlewareStage(name="invokeModel")
--- a/strands-py/src/strands/event_loop/event_loop.py
+++ b/strands-py/src/strands/event_loop/event_loop.py
@@ -8,13 +8,16 @@ The event loop allows agents to:
 4. Manage recursive execution cycles
 """
 
+import copy
 import logging
 import uuid
-from collections.abc import AsyncGenerator
+from collections.abc import AsyncGenerator, Callable
 from typing import TYPE_CHECKING, Any
 
 from opentelemetry import trace as trace_api
 
+from .._middleware.stages import InvokeModelContext, InvokeModelResult, InvokeModelStage
+from .._middleware.types import MiddlewareResult
 from ..experimental.checkpoint import Checkpoint, CheckpointPosition
 from ..hooks import AfterModelCallEvent, BeforeModelCallEvent, MessageAddedEvent
 from ..telemetry.metrics import Trace
@@ -34,7 +37,7 @@ from ..types._events import (
     TypedEvent,
 )
 from ..types.agent import Limits
-from ..types.content import Message, Messages
+from ..types.content import Message, Messages, split_system_prompt
 from ..types.event_loop import Metrics, Usage
 from ..types.exceptions import (
     ContextWindowOverflowException,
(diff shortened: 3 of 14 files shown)


strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -566,54 +514,113 @@ async def _handle_model_execution(
...
+                else agent.system_prompt
+            )
+            middleware_context = InvokeModelContext(
+                agent=agent,
+                messages=copy.deepcopy(agent.messages),
+                system_prompt=copy.deepcopy(system_prompt_value),
+                tool_specs=copy.deepcopy(tool_specs),
+                tool_choice=copy.deepcopy(structured_output_context.tool_choice),
+                invocation_state=invocation_state,
+            )
```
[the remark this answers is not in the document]
zastrowm: Nah, not until we identify it as a botteneck or problem

opieter-aws (approved):

agent-of-mkmeral (commented):
**Assessment**: Request Changes

Reviewed at `971f049` — "the works": full read of the Python implementation, side-by-side comparison with the TS reference (`strands-ts/src/middleware/` + `agent.ts` integration), ran the committed suites (49 middleware + 47 event-loop tests, all pass), and wrote adversarial tests against the spec in `strands-ts/src/middleware/README.md`. The registry port is faithful and clean. **But the event-loop integration has a result-propagation bug that breaks three spec requirements** — found by adversarial tests, not caught by the committed suite.

## 🔴 The transformed middleware result is silently discarded (1 blocking issue, 3 symptoms)

`event_loop_cycle` (line 299, pre-existing code) derives the cycle's `stop_reason`/`message` from the **last yielded `ModelStopReason` event**, not from the `MiddlewareResult` that `_handle_model_execution` so carefully extracts:

```python
async for model_event in model_events:
    if not isinstance(model_event, ModelStopReason):
        yield model_event
stop_reason, message, *_ = model_event["stop"]   # ← raw terminal event, NOT model_result
```

The terminal yields its `ModelStopReason` *before* Output/Wrap transforms run, so the transformed result updates `agent.messages` and hooks (inside `_handle_model_execution`) while the rest of the cycle — **tool dispatch, structured output, `EventLoopStopEvent` → `AgentResult`** — uses the stale untransformed event. Three verified failures (repros in thread):

1. **Split-brain state**: Output middleware replaces `message` → `agent.messages[-1]` = transformed, `AgentResult.message` = original. In TS the middleware result is authoritative (`return middlewareResult.result`).
2. **`stop_reason` transforms are ignored**: middleware rewrites `tool_use` → `end_turn` (suppress tool calls) → tools execute anyway. Spec: "Handlers can transform the result."
3. **Spec-compliant short-circuit crashes**: yielding *only* `MiddlewareResult` (the documented contract — TS just `return`s) → `UnboundLocalError: model_event` at line 299. The committed short-circuit tests pass only because they *also* yield a `ModelStopReason` event — an undocumented requirement with no TS equivalent. Similarly, a filtering middleware that drops `ModelStopReason` (spec: "Handlers can filter events") → `KeyError: 'stop'`.

**Fix**: make `model_result` authoritative — e.g. suppress the terminal's raw `ModelStopReason` at the integration boundary and have `_handle_model_execution` emit a fresh one built from the post-middleware `model_result` after the retry loop. That fixes all three symptoms and removes the hidden "must yield ModelStopReason" contract. Please also add tests asserting `AgentResult.message`/tool-dispatch reflect transformed results — `test_output_transforms_result` currently only asserts the handler *ran*, which is why this slipped through.

## 🟡 Important

- **String system prompts reach middleware as content blocks** (thread): `Agent(system_prompt="...")` → middleware sees `[{"text": "..."}]`, never a `str`. TS passes the user's original form through. Divergence from "matching TypeScript's `SystemPrompt` type" in the description — the type matches, the runtime value doesn't.
- **`tool_choice: Any | None`** (thread): `ToolChoice` exists in `types/tools.py` — same rationale as the `InvokeModelResult` typing fix you already made.
- **`ruff format`**: with the pinned range (`ruff>=0.13,<0.15`; I used 0.14.14), `ruff format --check` would reformat `event_loop.py` (the `system_prompt_value` ternary + the `yield MiddlewareResult(...)` call) and both test files. If CI's resolved ruff version passes, fine — but worth a local re-run of `hatch run prepare`.

## 🔵 Minor

- **`model_state` divergence under-documented** (thread): TS snapshots model state before the chain and writes back after — middleware *cannot* influence it by design. Python's terminal reads `agent._model_state` live by reference, so middleware mutations and per-retry changes *do* take effect. README notes the mechanical difference but not the lost invariant.

## TS parity scorecard

| Aspect | Status |
|---|---|
| Registry: phases, ordering, fast path, composition | ✅ faithful port |
| Generator cleanup (`aclose` vs `yield*`) | ✅ equivalent, well-tested |
| `MiddlewareResult` sentinel vs generator return | ⚠️ sound idea, but integration discards it (🔴 above) |
| Short-circuit contract | 🔴 crashes without undocumented extra event |
| Result transformation authority | 🔴 TS: authoritative; Py: partially ignored |
| `system_prompt` form | 🟡 always blocks, TS preserves user form |
| `model_state` isolation | 🟡 TS: snapshot+writeback; Py: live reference |
| `remove()` | ✅ documented intentional omission |
| Hooks-outside-middleware, span timing, defensive copies | ✅ matches TS |

The registry is mergeable as-is; the event-loop integration needs the result-propagation fix before this lands. Test commands: `python -m pytest tests/strands/middleware/ tests/strands/event_loop/test_event_loop*.py` → 96 passed at `971f049`.

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -566,54 +514,117 @@ async def _handle_model_execution(
...
-                break  # Success! Break out of retry loop
+            # Run through middleware chain (terminal includes span tracking)
+            model_result: InvokeModelResult | None = None
+            async for event in agent._middleware_registry.invoke(
+                InvokeModelStage,
+                middleware_context,
+                _make_invoke_model_terminal(agent, cycle_span, tracer),
+            ):
+                if isinstance(event, MiddlewareResult):
+                    model_result = event.value
```
agent-of-mkmeral, line 554: 🔴 **Blocking**: `model_result` is extracted here and used for `agent.messages`/hooks below, but `event_loop_cycle` (line 299) re-derives `stop_reason, message, *_ = model_event["stop"]` from the last raw `ModelStopReason` yielded by the terminal — *before* middleware transforms. The transformed result never reaches tool dispatch, structured output handling, or `AgentResult`.

Three repros (all pass on TS semantics, all fail here):

```python
# 1. Split brain: history vs result
def output_handler(result):
    return replace(result, message={"role": "assistant", "content": [{"text": "TRANSFORMED"}]})
agent._middleware_registry.add_middleware(InvokeModelStage.Output, output_handler)
result = agent("test")
# agent.messages[-1] == "TRANSFORMED"  ✓
# result.message     == "original"     ✗ stale
```

```python
# 2. stop_reason rewrite ignored: middleware suppresses tool calls → tools run anyway
def suppress(result):
    return replace(result, stop_reason="end_turn", message=text_msg)
# → toolResult still appears in agent.messages
```

```python
# 3. Spec-compliant short-circuit (yield ONLY MiddlewareResult, like TS `return`):
async def cached(context, next_fn):
    yield MiddlewareResult(InvokeModelResult("end_turn", msg, usage, metrics))
# → UnboundLocalError: cannot access local variable 'model_event' (event_loop.py:299)
# Filtering ModelStopReason (spec: "Handlers can filter events") → KeyError: 'stop'
```

The committed tests pass because they yield a `ModelStopReason` event *alongside* the `MiddlewareResult` — an undocumented hard requirement that doesn't exist in TS and isn't in either README's pass-through example.

**Suggestion**: make `model_result` the single source of truth — filter the terminal's raw `ModelStopReason` out of the forwarded stream and emit a fresh `ModelStopReason` built from `model_result` after the retry loop completes (right before `break`/after it). Then `event_loop_cycle`'s line-299 pattern picks up the transformed values, short-circuit works without the extra event, and event filtering can't break the loop. Tests asserting `AgentResult.message` and tool dispatch honor transformed results would lock this in (`test_output_transforms_result` currently only checks the handler executed).

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -566,54 +514,117 @@ async def _handle_model_execution(
...
+            else:
+                tool_specs = agent.tool_registry.get_all_tool_specs()
+
+            # Build middleware context with defensive copies to prevent accidental mutation.
+            # invocation_state is intentionally shared by reference (hooks/tools write to it).
+            system_prompt_value = (
+                agent._system_prompt_content
+                if agent._system_prompt_content is not None
+                else agent.system_prompt
+            )
```
agent-of-mkmeral, line 536: 🟡 **TS divergence**: because `split_system_prompt("foo")` always populates `_system_prompt_content = [{"text": "foo"}]`, this branch means middleware **never sees a plain `str`** — even when the user configured `Agent(system_prompt="foo")`:

```python
agent = Agent(model=model, system_prompt="I am a string prompt")
# middleware sees: [{'text': 'I am a string prompt'}]  — not the str
```

In TS, `InvokeModelContext.systemPrompt` carries the user's original form (`cloneSystemPrompt(this.systemPrompt)` — a string stays a string). Practical impact: middleware doing `context.system_prompt + "\nextra"` or `isinstance(context.system_prompt, str)` behaves differently across the two SDKs, and the `str` arm of the union is effectively dead in Python. The PR description's "matching TypeScript's `SystemPrompt` type" holds for the *type* but not the runtime *value*.

Since `Agent` doesn't retain the original input form, the cheap fix is preferring the string when it round-trips (`if agent.system_prompt is not None and agent._system_prompt_content == [{"text": agent.system_prompt}]: use the str`). If you'd rather keep blocks-always as the canonical context form, that's defensible too — but then please document it as an intentional divergence in `_middleware/README.md`, since it's exactly the kind of subtle cross-SDK difference that README exists to capture.

strands-py/src/strands/_middleware/stages.py
```diff
@@ -0,0 +1,45 @@
...
+    All collection fields (messages, system_prompt, tool_specs, tool_choice) are
+    defensive copies — middleware cannot accidentally mutate agent state.
+    invocation_state is shared by reference (hooks and tools write to it during streaming).
+    """
+
+    agent: Agent
+    messages: Messages
+    system_prompt: SystemPrompt
+    tool_specs: list[ToolSpec]
+    tool_choice: Any | None
```
agent-of-mkmeral, line 31: 🟡 `tool_choice: Any | None` — `ToolChoice` already exists (`types/tools.py:170`) and is what `stream_messages` accepts. The previous review round fixed `InvokeModelResult` to use `StopReason`/`Message`/`Usage`/`Metrics`; this field looks like the one stray `Any` left behind. A `TYPE_CHECKING` import keeps it cycle-free, same as the others above.

strands-py/src/strands/_middleware/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+The `_middleware/` package is not part of the public API. Internal consumers access it via `agent._middleware_registry.add_middleware(...)`.
+
+## System prompt as a union type
+
+`InvokeModelContext.system_prompt` is `str | list[SystemContentBlock] | None` (a single union field). The terminal decomposes this into the two-param form needed by `Model.stream()` via `split_system_prompt()`.
+
+## Defensive copies
+
+Context fields (`messages`, `system_prompt`, `tool_specs`, `tool_choice`) are deep-copied when building the middleware context. `invocation_state` is shared by reference. `model_state` is not on the context — the terminal reads it directly from the agent.
```
agent-of-mkmeral, line 33: 🔵 The `model_state` note documents the mechanical difference but omits the behavioral one. The TS README states middleware **cannot** access or modify model state: the agent snapshots `modelState` before the chain runs and writes back after it completes, so any middleware mutation — and any mid-chain retry — operates on the isolated snapshot.

Here the terminal reads `agent._model_state` **live by reference** on every `next()` call, so:
- middleware mutations to `agent._model_state` *do* take effect (forbidden by design in TS), and
- a retry middleware re-invoking `next()` picks up whatever the previous attempt's provider wrote (TS retries reuse the pre-chain snapshot).

Since this README is the divergence ledger, worth a sentence acknowledging the lost invariant — or aligning with a snapshot/writeback if the TS guarantee is meant to hold cross-SDK. Fine to defer while the API is private, but it should be a recorded decision rather than an accident.

PR #2762, opened 2026-06-12
feat(python): integrate Sandbox with Agent

gautamsirdeshmukh:
## Description

Adds `Sandbox` as a first-class parameter on `Agent`. This PR is a Python port of the Typescript implementation brought in via #2563.

This PR also takes care of renaming the file for the `PosixShellSandbox` implementation from `shell.py` to `posix_shell.py` to align with Typescript convention.

## Related Issues

- https://github.com/strands-agents/harness-sdk/pull/2563

## Documentation PR

Not yet

## Type of Change

New feature

## Testing

## Checklist

----


strands-py/src/strands/agent/agent.py
```diff
@@ -520,6 +534,17 @@ def cancel(self) -> None:
...
+    @property
+    def sandbox(self) -> Sandbox:
+        """Execution environment for running commands, code, and file operations.
+
+        Returns the configured sandbox, or a host default
+        (:class:`~strands.sandbox.NotASandboxLocalEnvironment`, no isolation) when none was
+        configured or ``sandbox=False`` was passed.
+        """
+        # isinstance, not truthiness: a custom Sandbox could be falsy. False/None fall through.
+        return self._sandbox if isinstance(self._sandbox, Sandbox) else NotASandboxLocalEnvironment()
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Updated to use a module-level singleton accessed via a `default_sandbox()` function instead of creating a fresh `NotASandboxLocalEnvironment()` instance every time (accessed via function so that instance itself remains private)

strands-py/tests/strands/sandbox/test_not_a_sandbox_local_environment.py
```diff
@@ -0,0 +1,162 @@
...
+    await sandbox.write_text(str(tmp_path / "c.txt"), "cc")
+    await sandbox.write_text(str(tmp_path / "a.txt"), "a")
+    await sandbox.write_text(str(tmp_path / "b.txt"), "bbb")
+    (tmp_path / "sub").mkdir()
+
+    files = await sandbox.list_files(str(tmp_path))
+    assert [f.name for f in files] == ["a.txt", "b.txt", "c.txt", "sub"]
+
+    a = next(f for f in files if f.name == "a.txt")
+    assert a.is_dir is False
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Updated test accordingly, added another to verify NotADirectoryError is raised as expected when calling `list_files` on a non-dir path

strands-py/src/strands/sandbox/not_a_sandbox_local_environment.py
```diff
@@ -0,0 +1,158 @@
...
+from collections.abc import AsyncGenerator
+from pathlib import Path
+from typing import Any
+
+from .posix_shell import PosixShellSandbox, build_shell_env_prefix
+from .stream_process import stream_process
+from .types import ExecutionResult, FileInfo, StreamChunk
+
+
+class NotASandboxLocalEnvironment(PosixShellSandbox):
```
gautamsirdeshmukh: Why is this extending PosixShellSandbox and not Sandbox?

strands-py/src/strands/agent/agent.py
```diff
@@ -261,12 +263,24 @@ def __init__(
...
                 pair with a SessionManager for cross-process state continuity.
                 Defaults to False. See :mod:`strands.experimental.checkpoint`.
+            sandbox: Execution environment for running commands, code, and file operations.
+                When provided, sandbox-aware tools route operations through it via
+                ``context.agent.sandbox``. Two distinct intents that currently resolve to the
+                same host execution:
+
+                - ``None`` (omitted): use the host default
+                  (:class:`~strands.sandbox.NotASandboxLocalEnvironment`, no isolation).
+                - ``False``: explicitly opt out of a managed sandbox and run on the host.
```
mkmeral: what's the difference? any why?

strands-py/src/strands/agent/agent.py
```diff
@@ -523,6 +537,17 @@ def cancel(self) -> None:
...
 
+    @property
+    def sandbox(self) -> Sandbox:
+        """Execution environment for running commands, code, and file operations.
+
+        Returns the configured sandbox, or a host default
+        (:class:`~strands.sandbox.NotASandboxLocalEnvironment`, no isolation) when none was
+        configured or ``sandbox=False`` was passed.
+        """
+        # isinstance, not truthiness: a custom Sandbox could be falsy. False/None fall through.
```
mkmeral: > a custom Sandbox could be falsy

why? is our sandbox an abc or a protocol? as long as its runtime checkable, it has to be sandbox. 

if we expect people to pass random stuff, we should validate and throw, not accommodate

mkmeral: (currently this eats up the error, and customer may not realize we went to defualt)

strands-py/src/strands/sandbox/__init__.py
```diff
@@ -34,15 +30,26 @@ async def execute_streaming(self, command, *, timeout=None, cwd=None, env=None,
...
 from .constants import LANGUAGE_PATTERN
 from .docker import DockerSandbox
-from .shell import PosixShellSandbox
+from .not_a_sandbox_local_environment import NotASandboxLocalEnvironment
+from .posix_shell import PosixShellSandbox
 from .ssh import SshSandbox
 from .types import ExecutionResult, FileInfo, OutputFile, StreamChunk, StreamType
 
+# One shared, stateless instance, mirroring the TS oracle's module-level default.
+_default_sandbox = NotASandboxLocalEnvironment()
```
mkmeral: do we need this in python too? why?

strands-py/src/strands/sandbox/__init__.py
```diff
@@ -34,15 +30,26 @@ async def execute_streaming(self, command, *, timeout=None, cwd=None, env=None,
...
+    """Return the shared host default used when an Agent has no sandbox configured."""
+    return _default_sandbox
+
+
 __all__ = [
     "DockerSandbox",
     "ExecutionResult",
     "FileInfo",
     "LANGUAGE_PATTERN",
+    "NotASandboxLocalEnvironment",
```
mkmeral: are we exporting this? why?

strands-py/src/strands/sandbox/ssh.py
```diff
@@ -8,7 +8,7 @@
 from collections.abc import AsyncGenerator
 from typing import Any
 
-from .shell import PosixShellSandbox, build_shell_env_prefix
+from .posix_shell import PosixShellSandbox, build_shell_env_prefix
```
mkmeral, line 11: nit: backwards incompatible but we haven't merged the agent integ yet so I am fine

agent-of-mkmeral:
## Review: Sandbox ↔ Agent integration (Python port) — with TS comparison

Pulled the branch at `d26d9b9`, installed, and ran the works. **Verdict: ship it.** Clean, idiomatic port; the prior bot round already nailed the two real issues (singleton identity + full-shape `FileInfo` assert) and both are verified fixed. Below is what I checked plus the TS diff analysis @mkmeral asked for.

### ✅ Verification
| Check | Result |
|---|---|
| `test_not_a_sandbox_local_environment.py` + `test_agent_sandbox.py` + `test_posix_shell.py` | **80 passed** (23 new across the two new files) |
| `ruff check` on all changed files | clean |
| `ruff format --check` | the only diff is in `_resolve_context_manager` — **pre-existing code, not this PR's lines**; every line this PR adds is already formatted |
| `.shell` → `.posix_shell` rename | clean, no dangling references anywhere in `src`/`tests`; `docker.py`/`ssh.py` are the +1/-1 import updates |
| `agent.sandbox is agent.sandbox` | `True` (singleton works); shared across agents |

### 🔬 TS comparison — does the diff make sense?
I diffed each file against the `strands-ts/src/sandbox/` oracle. The intentional divergences are well-documented in docstrings and mostly **improve** on the TS version. Three are worth a maintainer's eyes:

**1. 🟡 Tool vending is not ported (biggest design question).** The TS `Sandbox` base carries `toolPrefix = 'sandbox'` + `getTools()`, and `Agent.initialize()` registers sandbox-vended tools (prefixed, MCP-style, skipping user-registered name collisions — `agent.ts:741-759`). The Python `Sandbox` base has **no `tool_prefix`/`get_tools`**, and the Python `Agent` does **not** register any sandbox-vended tools. The PR's design (per the docstring) is instead "sandbox-aware tools route operations through `context.agent.sandbox`" — i.e. tools reach the sandbox *through* the agent rather than being *vended by* it. That's a legitimate architectural choice, but it's a real divergence from the oracle and changes how a sandbox surfaces capabilities. **Is tool-vending deliberately out of scope for this PR (follow-up), or is the Python design intentionally pull-based?** Worth a one-line note in the PR body either way so the next porter doesn't "fix" it.

**2. 🟢 `Agent.sandbox` getter — Python is *more* correct than the oracle.** TS: `return this._sandbox || defaultSandbox.get()` (truthiness). Python: `return self._sandbox if isinstance(self._sandbox, Sandbox) else default_sandbox()`. The `isinstance` guard means a custom `Sandbox` that happens to be falsy still resolves correctly, where the TS `||` would silently fall through to the default. Nice catch, and the inline comment calls it out. No change needed — just flagging it as a place the two SDKs now behave differently (TS could adopt this).

**3. 🟡 Error model divergence (cross-language parity).** TS has `sandbox/errors.ts` (`SandboxPathNotFoundError`, `SandboxTimeoutError`, `SandboxAbortError`) and maps both `ENOENT` *and* `ENOTDIR` → `SandboxPathNotFoundError` in `listFiles`. Python has no errors module and instead surfaces native `FileNotFoundError` / `NotADirectoryError` / `TimeoutError`. Using native exceptions is the right Python idiom (and the new ENOTDIR test asserts `NotADirectoryError`), **but** note the consumer-visible consequence: a missing dir vs. a file-as-dir are *two distinct* exceptions in Python yet *one* in TS. If anything ever branches on these across languages, the contract differs. Fine to keep — just calling out the parity gap.

### 💬 Answering @gautamsirdeshmukh's open thread (`not_a_sandbox_local_environment.py:33`)
> Why is this extending `PosixShellSandbox` and not `Sandbox`?

Because Python factored the base64-heredoc code-execution logic into `PosixShellSandbox.execute_code_streaming`, whereas TS's `NotASandboxLocalEnvironment` extends `Sandbox` directly and **reimplements** that heredoc inline. By extending `PosixShellSandbox`, the Python class inherits `execute_code_streaming` for free and only has to implement the shell `execute_streaming` (it then overrides the file ops with native `pathlib`/`os` calls to get real `size` metadata that the shell-based base can't report). So the Python layering is strictly DRYer than the oracle — extending `PosixShellSandbox` is the right call here, not a mistake. The module docstring already explains this; I'd just resolve the thread with that reasoning.

### Nits (non-blocking)
- `default_sandbox()` is reachable via `from strands.sandbox import default_sandbox` but isn't in `sandbox/__all__` (nor re-exported at top-level `strands`). This actually *matches* TS (which doesn't export `defaultSandbox` from `index.ts`), so it's consistent — but if `default_sandbox()` is meant as public API, add it to `__all__`; if it's internal, the current state is fine.
- The module-level `_default_sandbox = NotASandboxLocalEnvironment()` instantiates eagerly at import (vs TS's lazy `DefaultSlot.set()` at the env entry point). Harmless since it's stateless and Python core always runs on a host — just a timing difference from the oracle.
- Process item: the bot already flagged `api/needs-review` for the new public surface (`sandbox` param, `Agent.sandbox`, exported `NotASandboxLocalEnvironment`). Agree — worth confirming with a maintainer even though it's a port of the already-reviewed TS design.

Nice work on the iterations — the divergence notes in the docstrings made this a fast review. 👍

strands-py/src/strands/agent/agent.py
gautamsirdeshmukh: Updated docstring to make it clear that `false` and `None` have different intentions (explicit disable vs empty/default case) but as of right now both lead to `NotASandboxLocalEnvironment`

strands-py/src/strands/agent/agent.py
gautamsirdeshmukh: Added check in agent init to throw TypeError if sandbox is neither None, False nor isinstance(sandbox, Sandbox)

Also updated comment to more accurately represent the flow (mention TypeError check, explicitly indicate that False/None fall back to host default)


Also also added a new test to ensure TypeError is thrown as expected

strands-py/src/strands/sandbox/__init__.py
gautamsirdeshmukh: Discussed offline, updating to use per-agent instances with memoization, will address for typescript as well

strands-py/src/strands/sandbox/__init__.py
gautamsirdeshmukh: Removing, also removing Docker/SSH to match TS

strands-py/src/strands/agent/agent.py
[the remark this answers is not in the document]
gautamsirdeshmukh: Actually now updated to create an instance per agent

strands-py/src/strands/agent/agent.py
mkmeral: so it's the same behavior? why do we have both that still doesn't answer me. can we get rid of one? in Python we don't use `False` to disable in APIs I think, we use `None`, so we should remove False option

strands-py/src/strands/agent/agent.py
```diff
@@ -261,12 +264,30 @@ def __init__(
                 The SDK does not capture conversation state in the checkpoint;
                 pair with a SessionManager for cross-process state continuity.
                 Defaults to False. See :mod:`strands.experimental.checkpoint`.
+            sandbox: Execution environment for running commands, code, and file operations.
+                When provided, sandbox-aware tools route operations through it via
+                ``context.agent.sandbox``.
+
+                - ``None`` (default): fall back to
```
mkmeral: both options are still here

strands-py/src/strands/agent/agent.py
```diff
@@ -261,12 +264,30 @@ def __init__(
...
+                  :class:`~strands.sandbox.NotASandboxLocalEnvironment` (runs on the host,
+                  no isolation).
+                - ``False``: explicitly opt out of a managed sandbox (also falls back to
+                  :class:`~strands.sandbox.NotASandboxLocalEnvironment`).
+
+                ``None`` and ``False`` are kept as separate param values so that if a managed
```
mkmeral: if/when we add default, we'll think about it then. not right now

strands-py/src/strands/sandbox/__init__.py
```diff
@@ -9,17 +9,14 @@
...
 - Data types: :class:`StreamChunk`, :class:`FileInfo`, :class:`OutputFile`,
   :class:`ExecutionResult`, and the :data:`StreamType` literal.
 - :data:`LANGUAGE_PATTERN` — interpreter-name validation pattern.
 
-.. note::
-   Per the "Prefer Flat Namespaces Over Nested Modules" decision record, the
-   commonly-used symbols here will be re-exported from the top-level ``strands``
-   package. That re-export is deferred to the Agent↔Sandbox integration
-   follow-up (where the public surface stabilizes).
+Concrete sandboxes are imported from their own modules
```
mkmeral, line 16: lol +1 I missed this. this is also an issue with optional dependencies, as a follow up can we add this to style guide, or agents.md or sth?

mkmeral: when you add it to main import, on load, it loads everything. so if you have optional deps, it breaks loading

strands-py/src/strands/sandbox/errors.py
```diff
@@ -0,0 +1,35 @@
+"""Error types raised by sandbox execution and file operations.
+
+Mirrors ``strands-ts/src/sandbox/errors.ts``. Each error subclasses its stdlib
+equivalent so existing ``except TimeoutError`` / ``except FileNotFoundError``
+handlers keep working, while giving callers a sandbox-specific type to branch on.
+"""
+
+
+class SandboxTimeoutError(TimeoutError):
```
mkmeral, line 9: do we need them? why? 

(personally i'm inbetween, so feel free to push back, or remove)

gautamsirdeshmukh:
Addressing failed integ test

agent-of-mkmeral:
## Re-review at `54f0fc8` — verified locally ✅ ship it

Pulled HEAD `54f0fc8`, diffed against my last-reviewed `d26d9b9`, ran the works. **All maintainer feedback from this round is addressed and verified — I agree with the bot's re-approval.** Details below, including an honest note on where this PR reverses a call I praised last round.

### ✅ Verification (local, py3.13)
| Check | Result |
|---|---|
| `pytest tests/strands/sandbox/ + test_agent_sandbox.py` | **124 passed** (up from 80) |
| Touched non-sandbox tests (conv-mgr, middleware, context-offloader) | **209 passed** |
| `ruff check` (all changed files) | clean |
| `ruff format --check` (all changed files) | **26 files already formatted** — the pre-existing `_resolve_context_manager` diff I flagged last round is now gone |
| Per-agent identity | `a.sandbox is a.sandbox` ✅ True; `a1.sandbox is a2.sandbox` ✅ False (not shared) |
| `sandbox=False` / `None` | both → `NotASandboxLocalEnvironment` ✅ |
| `sandbox=<junk>` | raises `TypeError` ✅ (`str`/`int`/`object` all rejected) |
| ENOENT + ENOTDIR → `SandboxPathNotFoundError` | ✅ both map; still `isinstance(..., FileNotFoundError)` |
| `execute_code("print(6*7)", "python3")` after refactor | ✅ `42`, exit 0 |

### What changed since my last review, point by point

**1. ✅ Per-agent host default (resolves @mkmeral's singleton thread).** The module-level `default_sandbox()` singleton is gone; `_sandbox` is now resolved **once in `__init__`** to a per-agent `NotASandboxLocalEnvironment()`. Verified stable within an agent and *not* shared across agents — so a future stateful host default can't leak state between agents. The getter is now a trivial `return self._sandbox`. Good call, and better than the round-1 singleton.

**2. ✅ Input validation (resolves @mkmeral's "validate and throw" thread).** `__init__` now raises `TypeError` for anything that isn't `Sandbox`/`False`/`None`, instead of silently falling through to the default. This directly answers your point — `Sandbox` is a strict ABC (I confirmed: instantiating an incomplete subclass fails on 5 abstract methods), so the old falsy-accommodation was unnecessary; validating is correct. The `isinstance`-vs-truthiness comment I praised last round is now moot (and removed) because resolution happens once at init.

**3. ✅ Error-model parity gap closed (my round-1 flag #3).** New `sandbox/errors.py` adds `SandboxTimeoutError(TimeoutError)` and `SandboxPathNotFoundError(FileNotFoundError)`, re-exported from top-level `strands`. Both ENOENT *and* ENOTDIR now map to one `SandboxPathNotFoundError` (matching the TS oracle), while staying Python-idiomatic via stdlib subclassing — existing `except FileNotFoundError`/`except TimeoutError` handlers keep working. Importantly these are **wired in, not just defined**: `SandboxTimeoutError` is raised in `stream_process.py:122`, and `SandboxPathNotFoundError` in both `posix_shell.py` and `not_a_sandbox_local_environment.py`. 👍

**4. ✅ Reduced public surface (resolves @mkmeral's `__init__` export threads).** Dropped `NotASandboxLocalEnvironment`/`DockerSandbox`/`SshSandbox` from `sandbox/__all__` to match TS; concrete sandboxes are now imported from their own modules.

### 🔄 The one reversal worth calling out honestly
Last round I argued that extending `PosixShellSandbox` (to inherit `execute_code_streaming`) was "strictly DRYer and the right call." **This commit reverses that** — answering @gautamsirdeshmukh's *"why extend `PosixShellSandbox` and not `Sandbox`?"* thread — and `NotASandboxLocalEnvironment` now extends `Sandbox` directly, **reimplementing `execute_code_streaming` inline**.

I diffed the two bodies: they are **functionally identical** (same base64-heredoc: `base64 -d << 'EOF' | <lang>`, same `LANGUAGE_PATTERN` guard). So this is a deliberate **oracle-fidelity-over-DRY** trade: ~12 duplicated lines in exchange for matching the TS layering exactly (TS's `NotASandboxLocalEnvironment` also extends the base and inlines the heredoc). Given the goal of this PR is a faithful port and the duplication is tiny + well-documented in the new docstring, I'm fine with the call either way — just flagging that the layering now mirrors TS rather than improving on it.

**Tiny nit (non-blocking):** the base `PosixShellSandbox.execute_code_streaming` forwards `**kwargs` into `execute_streaming`; the `NotASandboxLocalEnvironment` override drops them. Harmless today (its `execute_streaming` ignores extras), but for symmetry you could thread `**kwargs` through.

### 🔬 TS comparison — does the diff still make sense?
Two divergences from the oracle remain (and now there's a *new* one from this round):

- **🟡 Tool vending still not ported** (my round-1 flag #1, unchanged). TS `Sandbox` has `toolPrefix`+`getTools()` and `agent.ts:743-745` registers sandbox-vended tools; Python has neither and uses the pull model (`context.agent.sandbox`). Still a legit but real architectural divergence — worth one line in the PR body confirming it's intentional/out-of-scope so the next porter doesn't "fix" it.
- **🟡 NEW divergence: default-instance lifecycle now differs from TS.** Last round Python *matched* TS (module-level singleton). After this round Python uses a **per-agent instance** while TS still uses `defaultSandbox.get()` + `this._sandbox || ...` truthiness (`agent.ts:438`). @gautamsirdeshmukh already noted on the thread this will be addressed for TS too — so this is a temporary, tracked gap, not a defect. Just flagging that as-of-this-PR the two SDKs resolve the host default differently.

Everything I can verify checks out. Clean iteration — the threads are all genuinely resolved, not just marked. 👍

strands-py/src/strands/vended_plugins/context_offloader/storage.py
```diff
@@ -328,9 +328,7 @@ def _evict(self, cycle: int) -> None:
             if self._evict_after_turns is None:
                 return
             threshold = cycle - self._evict_after_turns
-            stale_refs = [
```
mkmeral: do we already have integration with agent skills? 🤔

mkmeral: or other plugins?

strands-py/src/strands/agent/agent.py
gautamsirdeshmukh: Discussed offline, totally removing False case for now

strands-py/src/strands/sandbox/errors.py
gautamsirdeshmukh: I think these are worth including, this way for both Typescript and Python ENOENT/ENOTDIR errors map to the same clean SandboxPathNotFoundError, and timeouts map to SandboxTimeoutError. 

Planning to follow up with another PR to introduce AbortSignal (and the corresponding SandboxAbortError) to Python like we already have in Typescript

strands-py/src/strands/vended_plugins/context_offloader/storage.py
gautamsirdeshmukh: It looks like some other changes got swept in with this most recent commit, not relevant to this PR will remove

strands-py/src/strands/sandbox/__init__.py
gautamsirdeshmukh: Yup, will do so

agent-of-mkmeral:
## Re-review at `adeba88` — verified locally ✅ ship it

Pulled HEAD `adeba88` ("addressing review comments"), diffed against my last-reviewed `54f0fc8`, ran the works. **This round's maintainer feedback is fully addressed and verified — I agree with the bot's re-approval.**

### ✅ Verification (local, py3.13)
| Check | Result |
|---|---|
| `pytest tests/strands/sandbox/ + test_agent_sandbox.py` | **123 passed** (−1 from the removed `False` test, as expected) |
| `ruff check` (PR's 16 changed files) | clean |
| `ruff format --check` (PR's 16 changed files) | **all 16 already formatted** |
| `sandbox=None` | → `NotASandboxLocalEnvironment` ✅ |
| `sandbox=False` | now → **`TypeError`** (`got bool`) ✅ — option removed |
| `sandbox=<junk>` (`str`/`int`/`object`/`True`) | `TypeError` ✅ |
| valid custom `Sandbox` | passes through (`agent.sandbox is m`) ✅ |
| per-agent identity | `a.sandbox is a.sandbox` ✅ True; `a.sandbox is b.sandbox` ✅ False (not shared) |
| ENOENT + ENOTDIR → `SandboxPathNotFoundError` | ✅ both; still `isinstance(..., FileNotFoundError)` |
| `execute_code("print(6*7)","python3")` post-refactor | ✅ `42`, exit 0 |

### What changed since my last review (`54f0fc8` → `adeba88`), point by point

**1. ✅ `False` option removed (resolves @mkmeral's "in Python we use `None` not `False`" thread).** Signature is now `sandbox: Sandbox | None = None`. `None` → host default; anything that isn't `Sandbox`/`None` (including `False`) now raises `TypeError`. The redundant `test_sandbox_false_treated_as_unconfigured` test and the `False` docstring branch were dropped accordingly. Clean simplification — and `False` raising `TypeError` is the correct, un-surprising outcome now that it's not a sentinel.

**2. ✅ Unrelated `context_offloader/storage.py` changes removed (resolves @mkmeral's "do we have agent-skills integration?" thread).** Confirmed via the file list: the PR diff is now scoped to **sandbox-only** (16 files). The stray changes that got swept in last round are gone.

**3. ✅ Public-surface + error-model carried over from last round, still verified.** `sandbox/__init__.py` no longer eager-imports the concrete Docker/SSH sandboxes (only `base`/`posix_shell`/`types`), so the optional-deps eager-import concern doesn't bite at this layer. `errors.py` (`SandboxTimeoutError(TimeoutError)`, `SandboxPathNotFoundError(FileNotFoundError)`) is **wired in, not just defined** — `SandboxTimeoutError` raised in `stream_process.py:122`, `SandboxPathNotFoundError` in `posix_shell.py` + `not_a_sandbox_local_environment.py` — both re-exported from top-level `strands`.

### 🔬 TS comparison — does the diff still make sense?
- **🟡 Tool vending still not ported** (unchanged). TS `Sandbox` has `toolPrefix`/`getTools()` + agent registration; Python keeps the pull model (`context.agent.sandbox`). Legit but a real divergence — still worth one line in the PR body confirming it's intentional/out-of-scope so the next porter doesn't "fix" it.
- **🟡 Default-instance lifecycle differs from TS** (per-agent in PY vs module singleton + truthiness in TS). @gautamsirdeshmukh has said this will be mirrored in TS, so it's a tracked gap, not a defect.
- **🟢 `False`-removal is a PY-only ergonomics improvement.** TS's API shape isn't affected; this is just doing the right thing for Python idioms.

### Tiny nit (non-blocking, carried from last round — partially addressed)
The inline `NotASandboxLocalEnvironment.execute_code_streaming` now **accepts** `**kwargs` in its signature (good), but still doesn't **forward** them into the `execute_streaming(...)` call (`not_a_sandbox_local_environment.py:123`), whereas the base `PosixShellSandbox.execute_code_streaming` does (`posix_shell.py:143`). Harmless today (its `execute_streaming` ignores extras), but for full symmetry thread `**kwargs` through the call too.

Remaining open threads are discussion/follow-up items (`AbortSignal`/`SandboxAbortError` follow-up PR, the optional-deps style-guide note, tool-vending scope) rather than code blockers. Everything I can verify checks out — clean iteration. 👍

strands-py/src/strands/agent/agent.py
```diff
@@ -268,12 +264,21 @@ def __init__(
...
 
         Raises:
             ValueError: If agent id contains path separators.
         """
         self.model = BedrockModel() if not model else BedrockModel(model_id=model) if isinstance(model, str) else model
         self.messages = messages if messages is not None else []
+        if sandbox is not None and not isinstance(sandbox, Sandbox):
+            raise TypeError(f"sandbox must be a Sandbox instance or None, got {type(sandbox).__name__}")
+        # Resolve once: configured sandbox, or this agent's own host default (not shared across agents).
+        self._sandbox: Sandbox = sandbox if isinstance(sandbox, Sandbox) else NotASandboxLocalEnvironment()
```
mkmeral: nit: `if sandbox`, bc we already know if it exists, it's subclass of sandbox

strands-py/src/strands/sandbox/posix_shell.py
```diff
@@ -216,11 +217,15 @@ async def list_files(self, path: str, **kwargs: Any) -> list[FileInfo]:
...
+            FileNotFoundError: If the listing fails for another reason.
         """
         quoted = shlex.quote(path)
-        result = await self.execute(f"test -d {quoted} || exit 1; env QUOTING_STYLE=literal ls -1ap {quoted}")
+        # Exit 77 distinguishes a missing directory from ls's own failures (locale-independent).
+        result = await self.execute(f"test -d {quoted} || exit 77; env QUOTING_STYLE=literal ls -1ap {quoted}")
+        if result.exit_code == 77:
+            raise SandboxPathNotFoundError(path)
         if result.exit_code != 0:
             raise FileNotFoundError(result.stderr or f"Failed to list directory: {path}")
```
mkmeral: should this be sandbox file not found? isn't this inconsistent rn?

strands-py/src/strands/__init__.py
```diff
@@ -15,14 +25,23 @@
     "Agent",
     "AgentBase",
     "AgentSkills",
+    "ExecutionResult",
+    "FileInfo",
```
mkmeral: do we expect all of these to be used by normal developers developing agents? or will this be used by devs developing sandbox? 

^ use this question to determine whether it should be exported at `from strands import FileInfo, PosixShellSandbox`

strands-py/src/strands/agent/agent.py
gautamsirdeshmukh: Simplified to `sandbox or NotASandboxLocalEnvironment()`

strands-py/src/strands/sandbox/posix_shell.py
gautamsirdeshmukh: FileNotFoundError is actually the issue here (preexisting), for TS we raise SandboxPathNotFoundError for 77 and let anything else get raised as a generic Error. Updated to raise a generic python OSError for the second if case

agent-of-mkmeral:
## Re-review at `be3dbe6` — still ship it ✅

Pulled HEAD `be3dbe6`, diffed vs my last-reviewed `adeba88`. The only diff is @mkmeral's two latest threads — both resolved & verified. **123 tests pass, ruff clean & all 19 files formatted, exports now match the TS oracle exactly.** I agree with the bot's re-approval.

<details>
<summary>✅ Verification (local, py3.13)</summary>

| Check | Result |
|---|---|
| `pytest tests/strands/sandbox/ + test_agent_sandbox.py` | **123 passed** |
| `ruff check` (19 changed files) | clean |
| `ruff format --check` (19 changed files) | all 19 already formatted |
| `Agent()` (None) → host default | ✅ `NotASandboxLocalEnvironment`, `isinstance(_, Sandbox)` True |
| per-agent identity | `a.sandbox is a.sandbox` ✅ True; `a.sandbox is b.sandbox` ✅ False (not shared) |
| custom `Sandbox` | passes through ✅ |
| `sandbox=` `False`/`True`/`str`/`int`/`object` | all → **TypeError** ✅ |
| ENOENT + ENOTDIR → `SandboxPathNotFoundError` | ✅ both; still `isinstance(_, FileNotFoundError)` |
| `execute_code("print(6*7)","python3")` | ✅ `42`, exit 0 |

</details>

<details>
<summary>What changed since my last review (`adeba88` → `be3dbe6`)</summary>

**1. ✅ `if sandbox` simplification (resolves @mkmeral's `agent.py` nit).** Resolution is now:
```python
if sandbox is not None and not isinstance(sandbox, Sandbox):
    raise TypeError(f"sandbox must be a Sandbox instance or None, got {type(sandbox).__name__}")
self._sandbox: Sandbox = sandbox or NotASandboxLocalEnvironment()
```
The `isinstance` in the resolution is correctly gone — the `TypeError` guard just above already guarantees `sandbox` is a `Sandbox` or `None`, so the truthiness `or` is sufficient. The getter is now a trivial `return self._sandbox`. 👍

**2. ✅ `posix_shell.py` not-found consistency (resolves @mkmeral's thread).** `list_files` now raises `SandboxPathNotFoundError` for genuine path-absence (the `test -d || exit 77` sentinel), and the **generic fallback raises `OSError` instead of `FileNotFoundError`**. This matches the TS oracle exactly — TS `posix-shell.ts:109-113` throws `SandboxPathNotFoundError` on exit 77 and a generic `Error` otherwise. So "not found" now has a single dedicated signal, and other listing failures surface as plain `OSError` rather than being mislabeled. Good catch on the pre-existing inconsistency.

</details>

<details>
<summary>🔬 TS comparison — does the diff still make sense?</summary>

- 🟢 **Top-level exports now match TS exactly.** `strands/__init__.py` re-exports `Sandbox`, `PosixShellSandbox`, `SandboxTimeoutError`, `SandboxPathNotFoundError`, and the data types (`ExecutionResult`/`FileInfo`/`OutputFile`/`StreamChunk`/`StreamType`) — mirroring TS `index.ts:307-310`. On the open `__init__.py:29` thread the bot suggested dropping `PosixShellSandbox`, but **TS `index.ts:308` exports it**, so keeping it is the correct parity call. (Only divergence: TS also exports `SandboxAbortError`, the tracked AbortSignal follow-up.)
- 🟡 **Tool vending still not ported** (unchanged) — Python keeps the pull model (`context.agent.sandbox`); TS vends via `toolPrefix`/`getTools()`. Worth one line in the PR body confirming it's intentional/out-of-scope.
- 🟡 **Default-instance lifecycle differs from TS** (per-agent in PY vs module singleton in TS) — tracked, @gautamsirdeshmukh will mirror in TS.

</details>

<details>
<summary>🟢 One honest residual consistency point (non-blocking, pre-existing — matches TS)</summary>

Within `posix_shell.py` the "not found" signal is now intentionally split by operation: `list_files` raises the dedicated `SandboxPathNotFoundError`, while `read_file`/`remove_file` still raise bare `FileNotFoundError` (from a non-zero `base64`/`rm` exit, which can't cleanly distinguish ENOENT from other failures over the shell). This is **faithful to the TS oracle** (TS only special-cases `listFiles` for the path-not-found error; `readFile`/`removeFile` throw a generic `Error`), so I'm **not** asking to change it for this PR — just noting that `SandboxPathNotFoundError` is a `listFiles`-only concept in both SDKs, in case a future round wants to unify it (would need to apply to TS too for parity).

</details>

Everything I can verify checks out — the threads are genuinely resolved, not just marked. Clean iteration. 🚀

strands-py/src/strands/__init__.py
mkmeral: > returned/consumed by agent.sandbox.* calls they'll make

does not mean it's used by developers who just do this:

```
a = Agent(sandbox=Sandbox())
a("go do stuff")
```

also in the whole file, there is no other error export

```
  Conclusion to give the reviewer: the exports are over-broad. Trim the top-level set to nothing-or-minimal:
  - Drop the five data types (ExecutionResult, FileInfo, OutputFile, StreamChunk, StreamType) to strands.sandbox —
  accessible for the people who actually consume agent.sandbox.*, invisible to everyone else. 
  - Drop both error types too, to match how every other SDK exception is handled (submodule, not flat).
  - That leaves the real question as just: does Sandbox (and maybe PosixShellSandbox) belong flat? Sandbox is defensible
  as the param type; PosixShellSandbox is an extension base, so it belongs in the submodule next to
  DockerSandbox/SshSandbox
```

gautamsirdeshmukh: Updated to remove these as well as OutputFile, StreamChunk, StreamType

mkmeral (approved):

PR #2764, opened 2026-06-12
test: add integration test for memory manager

opieter-aws:
## Description

Add integration tests for `MemoryManager`.

The `BedrockKnowledgeBaseStore` already has integration coverage, but only at the **store** level — `bedrock-knowledge-base-store.test.node.ts` exercises `store.add()` / `store.search()` directly. Nothing tested the store through a real `MemoryManager` + `Agent`, which is how it's actually used. That left the two highest-value behaviors unverified end-to-end, because both live in the manager/agent layer rather than on the store:

- **Extraction** — wired in `MemoryManager.initAgent()`. A trigger fires (or `flush()` is called), the `ModelExtractor` distills the turn, and facts are written via `store.add()`. The store only ever sees `add()` calls.
- **Injection** (shipped in #2631) — an `InvokeModelStage.Input` middleware searches memory before a model call and folds a `<memory>` block into the latest user message *ephemerally*. The model sees retrieved context with no tool call, and durable history is untouched. This is a distinct path from the `search_memory` tool.

This PR adds a single Agent-level E2E suite that drives the full `MemoryManager` surface against a live Bedrock Knowledge Base (CUSTOM data source): extraction across all three trigger configurations, injection (single- and multi-store), and the `search_memory` / `add_memory` tools.

Key design decisions:

- **Each test isolates one path.** Every block enables only the feature it asserts and disables the rest (e.g. injection tests set `searchToolConfig: false`), so a pass can't be explained by an unintended path — a correct injection answer can only have come from the folded `<memory>` context, not a tool call.
- **Triggers are tested without `flush()`.** `flush()` force-saves regardless of triggers, so relying on it would prove "flush works," not "the trigger fired." The `InvocationTrigger` and `IntervalTrigger` tests poll for the background write instead, exercising the real autonomous path (`AfterInvocationEvent → fire() → coordinator.process()`); `flush()` gets its own dedicated test.
- **Injection is asserted on the model *input*, not just output.** An observer middleware captures the post-injection `InvokeModelContext.messages` to assert the `<memory>` block (with per-store `source` attribution) was folded in, and that durable `agent.messages` is byte-identical to the original prompt — proving the injection is ephemeral. This is deterministic regardless of model phrasing.
- **Tool tests force the tool once** via a one-shot middleware setting `ctx.toolChoice`, removing the only nondeterministic variable (whether the model *chooses* to call the tool) while still exercising the real registration → invoke → result path.
- **Cleanup + isolation.** Each store uses a unique `scope` so concurrent tests and stragglers can't cross-contaminate retrieval; every written document id is captured and batch-deleted on test finish.

The suite skips unless **both** a Bedrock model and the KB resources are available (the existing store-level suite needs only the KB).

## Documentation PR

No documentation changes — this is test-only.

## Type of Change

Other: test-only (new integration tests; no source changes)

## Testing

Run against an account with Bedrock model + KB access and the `/strands/test-infra/bedrock-knowledge-base/*` SSM parameters provisioned by `test-infra/`:

```
cd strands-ts
npx vitest run --project integ-node test/integ/memory/bedrock-knowledge-base-memory-manager.test.node.ts
```

All 8 tests pass against the live KB with no retries (~100s total). Without credentials/SSM the suite skips rather than fails. Type-check, lint, and prettier all pass.

## Checklist

----

### DIFF
--- a/strands-ts/test/integ/memory/bedrock-knowledge-base-memory-manager.test.node.ts
+++ b/strands-ts/test/integ/memory/bedrock-knowledge-base-memory-manager.test.node.ts
@@ -9,7 +9,8 @@ import type { ExtractionConfig } from '$/sdk/memory/extraction/types.js'
 import { BedrockKnowledgeBaseStore } from '$/sdk/vended-memory-stores/bedrock-knowledge-base/index.js'
 import { bedrock } from '../__fixtures__/model-providers.js'
 import { getMessageText } from '../__fixtures__/model-test-helpers.js'
-import { uniqueMarker, waitForIndexed, cleanupCustomDocument } from './_bedrock-kb-test-helpers.js'
+import { hasToolUse, waitFor } from '../__fixtures__/test-helpers.js'
+import { uniqueMarker, waitForIndexed, cleanupCustomDocument, searchUntil } from './_bedrock-kb-test-helpers.js'
 
 // Manual overrides — swap these to point at your own resources for local development.
 const OVERRIDES: Partial<{
@@ -21,10 +22,10 @@ const OVERRIDES: Partial<{
 }
 
 function config() {
-  const c = inject('provider-bedrock-kb')
+  const kb = inject('provider-bedrock-kb')
   return {
-    knowledgeBaseId: OVERRIDES.knowledgeBaseId ?? c.knowledgeBaseId!,
-    customDataSourceId: OVERRIDES.customDataSourceId ?? c.customDataSourceId!,
+    knowledgeBaseId: OVERRIDES.knowledgeBaseId ?? kb.knowledgeBaseId!,
+    customDataSourceId: OVERRIDES.customDataSourceId ?? kb.customDataSourceId!,
   }
 }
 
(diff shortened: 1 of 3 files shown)


strands-ts/test/integ/memory/bedrock-knowledge-base-memory-manager.test.node.ts
```diff
@@ -0,0 +1,414 @@
...
+const OVERRIDES: Partial<{
+  knowledgeBaseId: string
+  customDataSourceId: string
+}> = {
+  // knowledgeBaseId: 'YOUR-KB-ID',
+  // customDataSourceId: 'YOUR-CUSTOM-DS-ID',
+}
+
+function config() {
+  const c = inject('provider-bedrock-kb')
```
notowen333: nit: c?

strands-ts/test/integ/memory/bedrock-knowledge-base-memory-manager.test.node.ts
```diff
@@ -0,0 +1,414 @@
...
+    }
+  }
+
+  /**
+   * Polls a synchronous predicate until it holds, or the timeout elapses. Used to await the
+   * background extraction path: a trigger's `fire()` dispatches `coordinator.process()` without
+   * awaiting it, so `agent.invoke()` returns before the extraction has written. Polling the captured
+   * id list (rather than calling `flush()`) exercises the real autonomous path.
+   */
+  async function waitFor(predicate: () => boolean, { timeoutMs = 45_000, intervalMs = 1_000 } = {}): Promise<boolean> {
```
notowen333: could pull out some generic async utils potentially

strands-ts/test/integ/memory/bedrock-knowledge-base-memory-manager.test.node.ts
```diff
@@ -0,0 +1,414 @@
...
+  /**
+   * Registers an observer on the InvokeModelStage input phase AFTER the agent initializes, so it runs
+   * after the manager's injection middleware (input handlers compose in registration order, each
+   * seeing the prior's output). Returns a getter for the messages the model last saw — i.e. the
+   * ephemeral, post-injection input — without depending on the model's response.
+   */
+  async function observeModelInput(agent: Agent): Promise<() => readonly Message[] | undefined> {
+    await agent.initialize()
+    let seen: readonly Message[] | undefined
+    agent.addMiddleware(InvokeModelStage.Input, async (ctx) => {
```
notowen333, line 153: middleware in tests 🙇🏻

notowen333 (dismissed):

notowen333 (approved):

notowen333 (approved):

PR #2765, opened 2026-06-12
ci(changelog): add changelog generator and sync workflow

yonib05:
## Description

Adds the `changelog-sync` workflow plus the changelog generator it runs. This is the automation half of the Strands changelog feature: it turns GitHub Releases into the structured changelog data (#2761) that the docs site (#2717) renders, and keeps that data current after the initial backfill.

**The generation logic lives in the docs project itself** (`site/scripts/changelog/`), written in TypeScript and run via `tsx` — not in a separate GitHub Action. A developer can run it locally to backfill or sync on demand; the workflow is a thin trigger that runs the script and opens a PR. (This supersedes the earlier devtools composite-action approach in the now-closed devtools#66 and evals#259.)

### What's included

**Generator** (`site/scripts/changelog/`, dependency-free pure modules + a thin I/O boundary):
- `sync.ts` — CLI entry. Resolves a token (`--token` -> `GITHUB_TOKEN` -> `gh auth token`), picks releases (single tag or full backfill), writes `site/src/content/changelog/**`, prints a summary + any warnings. Does NOT open a PR (the workflow / a developer commits the diff).
- `github-client.ts` — the only network module: an `@octokit/rest` adapter building the injected client.
- Pure logic (unit-tested without network via injected deps): `run.ts` (orchestration + PR-fetch/prior-tag memoization), `derive-entries.ts` (entries from the compare API — every merged commit between the prior tag and this one, resolved commit->PR, so it is immune to release-note format drift), `enrich.ts` (`area-*` labels, breaking flag, merge SHA, language/docs gating from changed files), `build-release-file.ts`, `render-markdown.ts`, `tag-meta.ts`, `parse-release-body.ts`.
- `compareVersionDesc` is imported from the shared `site/src/util/semver.ts` (added in #2717), so there is a single definition shared by the runtime page code and the generator.

**Workflow** (`.github/workflows/changelog-sync.yml`): checkout -> `npm ci --prefix site` -> run the generator (inputs passed via `env:`) -> `peter-evans/create-pull-request`.
- **On release** (in-scope tag prefixes): single-mode sync of the released tag.
- **Daily cron backstop**: backfill mode with `skip-existing` (only generates missing files; zero re-enrichment cost; never regresses existing files), covering missed release events for both harness and evals (evals is public, read with the same token).
- **Concurrency guard**: de-dupes runs of the same event type + tag; cross-event (release vs. cron) duplicates are tolerated because content is identical.

### Security / token / fork flow
- **No untrusted input reaches the shell.** All workflow inputs (`SOURCE_REPO`, `MODE`, `SKIP_EXISTING`, `TAG`) are passed via `env:` and read by the script from `process.env` — nothing is interpolated into a `run:` command line, closing the GitHub Actions script-injection vector. The PR branch name is sanitized in a dedicated step (via an env var, not shell-interpolated).
- PRs created with the default `GITHUB_TOKEN` don't trigger `pull_request` workflows, so the required **CI Gate** never ran on auto-created PRs. This uses a dedicated agent account's PAT (`CHANGELOG_BOT_TOKEN`, scope `public_repo`), pushing to that account's fork and opening a cross-repo PR — a real-user fork PR triggers CI. First fork PR needs a one-time maintainer "Approve and run workflows".
- Job permissions are `contents: read` (every write is the PAT's).
- `peter-evans/create-pull-request` is pinned to a full commit SHA (`22a9089…`, v7.0.11). First-party `actions/checkout` and `actions/setup-node` use `@v6` tags, matching the rest of the repo's workflows.

### Frozen historical data
The generator owns NEW releases going forward. The 83 pre-monorepo files in #2761 were generated from the now-redirected `sdk-python` / `sdk-typescript` repos and are intentionally NOT reproducible by the harness-sdk-based generator — they are a frozen one-time artifact. A `render-markdown` byte-identical fixture test and a going-forward parity check (regenerating a harness-sdk-owned release) guard the renderer contract.

## Type of Change
New feature (CI automation + build-time generator)

## Testing
- `npm test --prefix site` — full site suite green, including the ported changelog generator suites under `site/test/changelog/` (99 tests: tag mapping, semver ordering, compare-API derivation, enrichment/gating, rendering, build orchestration, run wiring) and a **byte-identical render-fixture guard** asserting the renderer reproduces a committed file exactly.
- `npm run typecheck --prefix site` — clean.
- `npx prettier --check` over the new files — clean.
- Generator smoke-tested against the live GitHub API (single tag + going-forward parity on a harness-sdk-owned release).
- Workflow YAML validated; no `${{ }}` interpolation remains in any `run:` block.
- Note: GitHub CI has not yet run on the PR, and the workflow itself can only be fully exercised post-merge (it triggers on release/cron).

## Merge order
Stack: **#2717 (page + schema + semver) -> #2761 (data) -> #2765 (this).** This branch's tests import `semver.ts` from #2717, so the order is required.

### DIFF
--- a/.github/workflows/changelog-sync.yml
+++ b/.github/workflows/changelog-sync.yml
@@ -1,0 +1,137 @@
+name: "Changelog: Sync"
+
+on:
+  release:
+    types: [published]
+  schedule:
+    - cron: '17 7 * * *'   # daily backstop for missed release events
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Single release tag to sync'
+        type: string
+        required: true
+
+permissions:
+  contents: read
+
+jobs:
+  sync:
+    # On release: only our in-scope tag prefixes. The bare `v` prefix is for
+    # pre-monorepo python releases; the action itself rejects any tag that
+    # isn't `v<digit>`, so a stray non-version `v…` tag produces a no-op run.
+    # Cron/dispatch: always run.
+    if: >
+      github.event_name != 'release' ||
+      startsWith(github.event.release.tag_name, 'python/v') ||
+      startsWith(github.event.release.tag_name, 'typescript/v') ||
+      startsWith(github.event.release.tag_name, 'v')
+    runs-on: ubuntu-latest
+    # De-dupe overlapping runs of the SAME event type + tag (e.g. two release
+    # events for one tag, or two manual dispatches). The group includes
+    # github.event_name, so it deliberately does NOT serialize a release run
+    # against the daily cron: cron runs in backfill mode with no tag, so its key
+    # is always `...-schedule-backfill` and can't match a release's
+    # `...-release-<tag>`. A cross-event duplicate (cron regenerating a release
+    # whose single-mode PR is still open) is tolerated -- the content is
+    # identical and the stray PR is closed manually (see the notes block below).
+    # cancel-in-progress: false so an in-flight sync finishes rather than being
+    # interrupted.
+    concurrency:
+      group: changelog-sync-${{ github.event_name }}-${{ github.event.release.tag_name || inputs.tag || 'backfill' }}
+      cancel-in-progress: false
+    # The PAT (CHANGELOG_BOT_TOKEN) performs every write -- checkout, the
+    # github-script reads, and the branch push + PR creation all use it, not the
+    # default GITHUB_TOKEN. So GITHUB_TOKEN needs only read; keeping write here
+    # would be dead, misleading privilege.
+    permissions:
+      contents: read
+    steps:
+      # Notes:
+      # - Historical backfill is a deliberate one-time manual operation (run
+      #   the action against a branch and open a hand-reviewed PR) -- not wired
+      #   here.
+      # - Duplicate-PR window: skip-existing checks files on main, not open
+      #   PRs, and release vs. cron runs use different PR branch names, so a
+      #   release-triggered PR is NOT auto-closed by a later cron run. The
+      #   concurrency group only de-dupes within the same event type; a
+      #   release-vs-cron duplicate is tolerated because the content is
+      #   identical -- close the stray PR manually.
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v6
+        with:
+          node-version-file: .node-version
+      - run: npm ci --prefix site
+
+      - name: Generate harness changelog files
+        working-directory: site
+        env:
+          GITHUB_TOKEN: ${{ secrets.CHANGELOG_BOT_TOKEN }}
+          SOURCE_REPO: strands-agents/harness-sdk
+          MODE: ${{ github.event_name == 'schedule' && 'backfill' || 'single' }}
+          SKIP_EXISTING: ${{ github.event_name == 'schedule' }}
+          TAG: ${{ github.event.release.tag_name || inputs.tag }}
+        run: npm run changelog:sync
+
+      - name: Compute harness branch name
+        id: harness_branch
+        env:
+          TAG: ${{ github.event.release.tag_name || inputs.tag }}
+        run: |
+          if [ -n "$TAG" ]; then
+            SUFFIX=$(printf '%s' "$TAG" | sed 's/[^A-Za-z0-9._-][^A-Za-z0-9._-]*/-/g')
+          else
+            SUFFIX=backfill
+          fi
+          echo "name=changelog/sync-harness-sdk-$SUFFIX" >> "$GITHUB_OUTPUT"
+
+      # Dedicated agent account (outside the org): a classic PAT with the
+      # public_repo scope. Authors the PR as a real user, so the PR triggers
+      # the required CI Gate (unlike github.token, whose PRs don't).
+      # Account can't push to upstream directly, so the branch lands on its
+      # fork and the PR is opened cross-repo against target-repo. Note:
+      # `strands-agent` is SINGULAR -- the dedicated bot account, distinct
+      # from the `strands-agents` (plural) org. Not a typo.
+      - name: Open harness changelog PR
+        uses: peter-evans/create-pull-request@22a9089034f40e5a961c8808d113e2c98fb63676 # v7.0.11
+        with:
+          token: ${{ secrets.CHANGELOG_BOT_TOKEN }}
+          push-to-fork: strands-agent/harness-sdk
+          add-paths: site/src/content/changelog/**
+          branch: ${{ steps.harness_branch.outputs.name }}
+          title: "docs(changelog): sync strands-agents/harness-sdk ${{ github.event.release.tag_name || inputs.tag || 'backfill' }}"
+          commit-message: "docs(changelog): sync strands-agents/harness-sdk ${{ github.event.release.tag_name || inputs.tag || 'backfill' }}"
+          body: |
+            Automated changelog sync.
+            Add a curated `highlights:` block to any release file before merging if desired.
+          delete-branch: true
+
+      # Evals backstop: evals' own release workflow opens the cross-repo PR for
+      # instant sync; this catches missed events. The condition runs the evals
+      # step whether the harness step passed or failed (the two syncs are
+      # independent -- a harness failure must not delay the evals backstop a day)
+      # but NOT when the run was cancelled.
+      - name: Generate missed evals changelog files
+        if: (success() || failure()) && github.event_name == 'schedule'
+        working-directory: site
+        env:
+          GITHUB_TOKEN: ${{ secrets.CHANGELOG_BOT_TOKEN }}
+          SOURCE_REPO: strands-agents/evals
+          MODE: backfill
+          SKIP_EXISTING: 'true'
+        run: npm run changelog:sync
+
+      - name: Open evals changelog PR
+        if: (success() || failure()) && github.event_name == 'schedule'
+        uses: peter-evans/create-pull-request@22a9089034f40e5a961c8808d113e2c98fb63676 # v7.0.11
+        with:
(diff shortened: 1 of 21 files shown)


.github/workflows/changelog-sync.yml
```diff
@@ -0,0 +1,137 @@
+name: "Changelog: Sync"
+
+on:
+  release:
+    types: [published]
+  schedule:
+    - cron: '17 7 * * *'   # daily backstop for missed release events
```
zastrowm, line 7: @poshinchen is working on turning our entire release pipeline into a workflow - I was going to ask should this just piggy back off of that.

But a counter-argument is "This needs to work for strands-shell & evals, so no".  Keeping this comment for explanation purposes

.github/workflows/changelog-sync.yml
```diff
@@ -0,0 +1,137 @@
...
+    # is always `...-schedule-backfill` and can't match a release's
+    # `...-release-<tag>`. A cross-event duplicate (cron regenerating a release
+    # whose single-mode PR is still open) is tolerated -- the content is
+    # identical and the stray PR is closed manually (see the notes block below).
+    # cancel-in-progress: false so an in-flight sync finishes rather than being
+    # interrupted.
+    concurrency:
+      group: changelog-sync-${{ github.event_name }}-${{ github.event.release.tag_name || inputs.tag || 'backfill' }}
+      cancel-in-progress: false
+    # The PAT (CHANGELOG_BOT_TOKEN) performs every write -- checkout, the
```
zastrowm, line 36: nit: (as a follow-up maybe) - this is very verbose and should be reduced

.github/workflows/changelog-sync.yml
```diff
@@ -0,0 +1,137 @@
...
+    # github-script reads, and the branch push + PR creation all use it, not the
+    # default GITHUB_TOKEN. So GITHUB_TOKEN needs only read; keeping write here
+    # would be dead, misleading privilege.
+    permissions:
+      contents: read
+    steps:
+      # Notes:
+      # - Historical backfill is a deliberate one-time manual operation (run
```
zastrowm: A comment explaining what isn't here is just noise IMHO

.github/workflows/changelog-sync.yml
```diff
@@ -0,0 +1,137 @@
...
+        working-directory: site
+        env:
+          GITHUB_TOKEN: ${{ secrets.CHANGELOG_BOT_TOKEN }}
+          SOURCE_REPO: strands-agents/harness-sdk
+          MODE: ${{ github.event_name == 'schedule' && 'backfill' || 'single' }}
+          SKIP_EXISTING: ${{ github.event_name == 'schedule' }}
+          TAG: ${{ github.event.release.tag_name || inputs.tag }}
+        run: npm run changelog:sync
+
+      - name: Compute harness branch name
```
zastrowm, line 53: nit: I'd move this up as "Generate" is more expensive than harness-branch so if the latter is going to fail you want it done sooner than later

site/scripts/changelog/build-release-file.ts
```diff
@@ -0,0 +1,129 @@
...
+  release: Release,
+  deps: BuildDeps
+): Promise<{ path: string; contents: string; warning?: string } | null> {
+  const meta = tagToMeta(repo, release.tag_name)
+  if (!meta) return null
+
+  const path = `site/src/content/changelog/${fileNameFor(meta.sdk, meta.language, meta.version)}`
+  const existing = await deps.readExisting(path)
+
+  // skipExisting (used by the daily cron backstop): only generate files for
```
zastrowm: We need to update agent guidance to be less verbose

site/scripts/changelog/enrich.ts
```diff
@@ -0,0 +1,95 @@
...
+  return files.every(isDocPath)
+}
+
+export async function enrichFromPr(
+  repo: string,
+  num: number,
+  fetcher: (repo: string, num: number) => Promise<PrData | null>
+): Promise<Enrichment> {
+  const pr = await fetcher(repo, num)
+  // Unfetchable PR (deleted / rate-limited): degrade open -- keep the entry,
```
zastrowm: Rate-limit should fail - should it not?  I'd rather adjust the workflow/script than have worse release notes

This might be an issue for backfills but that's a one-off IMO

zastrowm: PR's can't be deleted either.

Are we ensuring that for TypeScript PRs, we're using the right source of a PR pre-mono repo?

site/scripts/changelog/parse-release-body.ts
```diff
@@ -0,0 +1,63 @@
...
+import type { NewContributor } from './types'
+
+const LINE = /^\s*[-*]\s+(.*)$/
+// Cross-SDK marker in newer harness notes, e.g. "title _(shared with TS)_".
+// Language gating (from PR changed files) already decides stream membership,
+// so strip this annotation from the rendered title.
+const SHARED_ANNOTATION = /\s*_\(shared with [^)]+\)_\s*$/i
+const CONVENTIONAL = /^(feat|fix|docs|style|refactor|perf|test|chore|build|ci|revert)(?:\(([^)]+)\))?(!)?:\s*(.+)$/i
+const KNOWN_TYPES = new Set(['feat', 'fix', 'docs', 'perf', 'refactor', 'test', 'chore'])
+// GitHub's "## New Contributors" lines: "@login made their first contribution in <pr-url>".
```
zastrowm, line 17: @poshinchen is changing the release file format; does this need to be updated for that?

site/scripts/changelog/render-markdown.ts
```diff
@@ -0,0 +1,94 @@
+// Render a release into the changelog markdown file format consumed by the
+// harness-sdk content collection (Plan 1 Zod schema). Hand-rolled minimal YAML
+// emitter -- entries are flat objects emitted as inline flow maps, mirroring the
+// committed fixtures. mergePreserving keeps any human-written highlights/body
+// on re-sync while refreshing the parsed entries/urls.
+
+import type { ReleaseFile } from './types'
+
+function q(s: unknown): string {
```
zastrowm, line 11: What is this for?  I'd setting off alarm bells;  Can this just be `JSON.stringify(String(s))`?

.github/workflows/changelog-sync.yml
yonib05: Yes exactly, shell I need to look into this started before that was around.

site/scripts/changelog/build-release-file.ts
yonib05: Ack

site/scripts/changelog/enrich.ts
yonib05: Yes, agreed, this only has implications for backfill, not ongoing. The API limits are ample for single releases, verified on a handful.

zastrowm: I think we'd want to error out though in that case; e.g. I want to be notified about this, not silently fail

.github/workflows/changelog-sync.yml
yonib05: Done in 4c079c81 -- compressed to 3 lines.

.github/workflows/changelog-sync.yml
yonib05: Removed in 4c079c81.

.github/workflows/changelog-sync.yml
yonib05: Done in 4c079c81 -- branch-name step now runs before Generate.

site/scripts/changelog/render-markdown.ts
yonib05: Replaced with JSON.stringify(String(s)) in 4c079c81 -- JSON strings are a strict subset of YAML double-quoted scalars, so the output stays valid YAML and the byte-identical render fixture still passes. The hand-rolled version was escaping the same subset by hand; no reason for it to exist.

site/scripts/changelog/enrich.ts
yonib05: Done in 4c079c81 -- github-client now throws on 403/429 (rate limit) so the run fails loudly instead of producing degraded notes. 404 still degrades to an unenriched entry since those are cross-repo/permission edge cases, not limits. For backfill (the only realistic way to hit limits) a failed run is just re-dispatched after reset.

site/scripts/changelog/build-release-file.ts
yonib05: Trimmed the verbose comment blocks in build-release-file and the workflow in 4c079c81. Updating AGENTS.md guidance on comment verbosity is a good follow-up outside this PR.

site/scripts/changelog/enrich.ts
yonib05: Right, dropped 'deleted' from the comment in 4c079c81 -- the 404 path is really cross-repo/permission edge cases. On pre-monorepo sourcing: yes, every derived entry carries a prRepo, and enrichment/links always use the PR's own repo (sdk-typescript / sdk-python), not the release's repo. Pinned by tests: 'monorepo-tagged release with PRs in the OLD flat repo is not language-gated' in build-release-file.test.ts, and the archived sdk-typescript tag mapping in tag-meta.test.ts.

site/scripts/changelog/parse-release-body.ts
yonib05: Should be unaffected: entries are derived from the compare API (commits between tags, resolved commit->PR), not from the release body, so body format changes don't touch the breakdown. The only body-dependent piece is parsing GitHub's auto-generated '## New Contributors' section for the first-timer chips. If the new release pipeline keeps that auto-generated section, nothing changes; if it drops it, we lose only the chips -- will sync with @poshinchen.

zastrowm (approved):

PR #2790, opened 2026-06-15
docs: add Strandslator design doc

pgrayy:
## Summary
- Proposes Strandslator, an agent-driven system for translating Strands features across languages
- Focuses on producing PRs that are fast and safe to review via mechanically-derived artifacts
- Covers the translation workflow, input context, agents, artifacts, and determinism strategy

### DIFF
--- a/designs/0012-strandslator-design.md
+++ b/designs/0012-strandslator-design.md
@@ -1,4 +1,4 @@
-# Strandslator
+# Strandslator Design
 
 **Status**: Proposed
 
@@ -90,9 +90,18 @@ Committing and pushing code is a good example. Rather than letting the agent dec
 
 ## Experiments
 
-I ran early experiments translating the Anthropic model provider from TypeScript to Python, erasing existing traces from the Python SDK to set realistic conditions. With Opus 4.8 and above, translation quality is strong. The takeaway is that translation itself is not the bottleneck; what's missing is the review infrastructure, which is what shifted focus toward the artifacts and report design in this proposal.
+I ran two experiments to validate the workflow and understand where human effort actually goes.
 
-Trust will build incrementally. Each reviewed and approved PR gives us more evidence about where the system is reliable, and over time the review process can become lighter as confidence grows.
+In the first experiment, I translated the Anthropic model provider from TypeScript to Python, erasing existing traces from the Python SDK to set realistic conditions. With Opus 4.8 and above, translation quality was strong. The takeaway: translation itself is not the bottleneck; review infrastructure is. That shifted focus toward the artifacts and report design in this proposal.
+
+In the second experiment, I used the Strandslator prototype to port `BedrockKnowledgeBaseStore` from TypeScript to Python ([#2834](https://github.com/strands-agents/harness-sdk/pull/2834)): store implementation, config types, 54 unit tests, and integration tests against a live Bedrock Knowledge Base. The prototype ran end to end and produced working code, but required human redirection on four issues:
+
+- **Dataclass instead of TypedDict for configs.** The codebase uses both (e.g. `CacheConfig` is a dataclass, `BaseModelConfig` is a TypedDict) for the same pattern, so the agent picked the wrong precedent. Partially corrected ([#2824](https://github.com/strands-agents/harness-sdk/pull/2824)), but inconsistencies remain. The rule: a TS `interface` passed as a constructor object literal maps to a `TypedDict`.
+- **`MemoryStoreConfig` was a Protocol, not a TypedDict.** From a separate earlier port. Protocols can't be spread into constructor kwargs, so the agent couldn't compose with the shared config type.
+- **No uuid7 in Python.** The TS source uses uuid v7 for document IDs; Python's stdlib has no v7 until 3.14, so the agent used uuid4. A genuine language gap, not a codebase consistency issue.
+- **Skipped boto3-stubs[s3].** Adding the stub would have tightened typing repo-wide (boto3-stubs injects overloads via package install, not per-import), exposing a pre-existing error in an unrelated file. The agent surfaced the issue and correctly refused to fix code outside the feature, but that forced human intervention.
+
+All but uuid7 trace to inconsistencies in the Python codebase. The agent was influenced by conflicting precedents and made defensible-but-wrong choices. With consistency going forward, this port reduces to a single decision (uuid7 vs uuid4). That is the concrete payoff of "Design for portability" and "Fix the system, not the instance": consistency means less human intervention, and the intervention that remains is about genuine language gaps rather than inherited drift.
 
 ## References
 
--- a/designs/0012-strandslator.md
+++ b/designs/0012-strandslator.md
@@ -96,5 +96,5 @@ Trust will build incrementally. Each reviewed and approved PR gives us more evid
 
 ## References
 
-- **[Adversarial and cross-language differential testing](https://gist.github.com/agent-of-mkmeral/5a4d0ce16a1242a711d77d7e01c19902#6-adversarial--cross-language-differential-testing).** Design notes on building an agentic multi-language SDK (from @murmeral), whose differential testing section argues for proving behavioral equivalence with failing tests as "non-hallucinatable evidence" rather than agent opinion.
+- **[Adversarial and cross-language differential testing](https://gist.github.com/agent-of-mkmeral/5a4d0ce16a1242a711d77d7e01c19902#6-adversarial--cross-language-differential-testing).** Design notes on building an agentic multi-language SDK (from @mkmeral), whose differential testing section argues for proving behavioral equivalence with failing tests as "non-hallucinatable evidence" rather than agent opinion.
 - **[ReCodeAgent (arXiv:2604.07341)](https://arxiv.org/abs/2604.07341).** Paper on an autonomous multi-agent system that translates and validates entire code repositories across languages, reporting high translation success rates over 118 real-world projects.
--- a/designs/0012-strandslator.md
+++ b/designs/0012-strandslator.md
@@ -1,5 +1,11 @@
 # Strandslator
 
+**Status**: Proposed
+
+**Date**: 2026-06-12
+
+**Issue**: https://github.com/strands-agents/harness-sdk/issues/2666
+
 ## Overview
 
 Every feature we build in one language eventually needs to exist in all the others. If Strands supports N languages with M features, keeping them in sync becomes an N * M cost.
--- a/designs/0012-strandslator.md
+++ b/designs/0012-strandslator.md
@@ -1,0 +1,94 @@
+# Strandslator
+
+## Overview
+
+Every feature we build in one language eventually needs to exist in all the others. If Strands supports N languages with M features, keeping them in sync becomes an N * M cost.
+
+Strandslator is a proposed system that uses agents to handle feature translation across languages. The bottleneck we're targeting isn't necessarily the translation itself, it's the review. An agent can produce the code, but a human still needs to verify its correctness, security, and idiomaticity. That verification is what takes time, and it's what we want to make fast. The system's design centers on giving reviewers the evidence and artifacts they need to approve a translation PR with both confidence and speed.
+
+This document describes the translation workflow, the input context, the agents involved, and the artifacts they generate. Orchestration will be discussed separately, but likely we build upon the existing GitHub agent infrastructure we already have in place ([ideas](https://gist.github.com/agent-of-mkmeral/5a4d0ce16a1242a711d77d7e01c19902#6-adversarial--cross-language-differential-testing)).
+
+## Solution
+
+Strandslator takes one completed source feature and one target language as input. The developer who completes the feature is the one who kicks off the workflow. Triggering is manual to start, since having a human make the call ensures the feature is genuinely ready before translation begins. We can explore automating this over time, but for now the human judgment is worth it.
+
+We scope each run to a single target language so we can inject language-specific context into the agents and keep the run focused enough to produce a reviewable PR. The workflow runs a sequence of agents that plan, implement, and validate the translation, restarting from the top if validation finds issues. Once everything passes, it produces documentation and assembles the review artifacts into a PR.
+
+```mermaid
+flowchart LR
+    src["Source feature<br/>(code, tests, docs, prompt)"] --> plan[Plan]
+    plan --> impl[Implement]
+    impl --> val[Validate]
+    val -->|tests fail / behavior gap| plan
+    val --> doc[Document]
+    doc --> report[Report]
+    report --> pr["Target-language PR<br/>+ review artifacts"]
+```
+
+This workflow is only as good as the features it translates. A few principles set the preconditions and constrain how it operates:
+
+- **No true source language.** A feature can originate in any language and translate to all others. Python however requires care, since some features touch components that follow outdated design patterns we no longer want to propagate. Features that are isolated enough from those components may translate just fine.
+- **Translate only completed features.** Complete means merged, tested, and documented. Code, tests, and docs must be in sync before translation kicks off. Anytime someone updates an existing feature, they update all paired files first.
+- **Explicit readiness.** Features may be developed or updated iteratively and so we don't want the workflow guessing whether work is done. We explicitly mark a feature as ready for translation once complete and all paired files are in sync.
+- **Translate, don't improve.** The agent reproduces behavior as-is. It does not fix unrelated code or redesign the feature.
+- **Idiomatic over literal, when confident.** Prefer a translation idiomatic to the target language. When uncertain, a literal translation is a fine starting point. The reviewer can always provide more context to guide the next iteration.
+- **Idiomatic surface over a portable core.** Build features with clean object-oriented design and a core that translates well across languages. Per-language deviations should be minimal and focused on developer experience, not functionality. For example, Python's `@tool` decorator is an idiomatic surface that returns a `FunctionTool` instance under the hood, and `FunctionTool` is the portable core that exists across languages.
+- **Design for portability.** When designing a feature, we should be mindful of how well it will translate. Clean, well-structured implementations translate more straightforwardly, and the agent shouldn't have to work around ambiguous or tangled code. If a feature is hard for the agent to translate, that's a signal the source needs cleanup before it's marked ready.
+- **Fix the system, not the instance.** Generalize failures into rules and update the system's context. If the root cause is upstream (ambiguous source, stale metadata), fix the feature development workflow. One-off corrections don't compound; system improvements do.
+
+## Context
+
+All input context lives in the monorepo. No GitHub issues are needed. During a run however, the agents are free to collect additional context on their own from external sources such as the web.
+
+Beyond the source code itself, the workflow draws on a set of markdown guides at fixed paths that describe how we build, test, secure, and document Strands. Think of this as a "strands-md" layer: a prose definition of the SDK that fills in the gaps where code alone doesn't tell the full story. Shared guidance lives in the root of the repo, and language-specific variants live in their respective packages (strands-py, strands-ts, etc.). Agents load these from known locations so they never have to guess where to look. The guides will evolve over time, and this list is by no means exhaustive, but the initial set includes:
+
+- **Source.** The feature code with API docstrings, pairing tests, and a metadata markdown file. Tests should use docstrings to describe the behavior being asserted when it isn't obvious from the assertions alone. The metadata file sits next to the feature module or in its subdirectory and explains why the feature is designed the way it is, including what we consider subjective. The code is the true spec, so this file should stay concise. We separate it from docstrings because docstrings are about public API usage, not internal implementation rationale. When feature code is updated, pairing metadata and tests must be updated too. The translation workflow should never have to work around conflicting context; that is a problem for the feature design workflow.
+- **Testing.** Cross-language rules for how we define tests. For example, we follow a one-to-one directory layout where source module `a` has a pairing test module `a`. Language-specific conventions layer on top. In Python for example, we use pytest with fixtures where shareable fixtures live in `conftest.py`.
+- **Building.** Covers environment setup, how to run tests, and what checks must pass before a PR is submitted (linting, formatting, etc.). Language-specific tooling is called out where relevant. In Python for example, we use hatch for environment and task management.
+- **Security.** Guidance on writing secure code and common pitfalls to avoid, such as leaking credentials, exposing sensitive data in logs, or accessing restricted paths.
+- **Documentation.** General documentation standards, with language-specific variants where conventions differ.
+
+## Artifacts
+
+Each run produces the translated code, tests, and docs, plus a review report attached to the PR. The report is the heart of this proposal. Its job is to give the reviewer confidence that the translation is correct, safe, and idiomatic, without having to reverse-engineer that confidence from the diff alone.
+
+Artifacts must be mechanically derived, not narrated. Since the agent writes its own report, anything it merely asserts carries no weight. What builds confidence is evidence the agent can't fake: actual test output, actual diffs, conventions the reviewer can spot-check. The report indexes and explains that evidence.
+
+The exact schema is not yet decided and will evolve as we learn what reviewers actually need. The overriding goal is human readability. Any time a reviewer is confused about something on a PR, that signals a gap in the report to fill. This list is by no means exhaustive, but the initial set of artifacts we think are worth highlighting:
+
+- **Behavior traceability matrix.** Maps each source test to the target test that asserts the same behavior, with its run result. Missing or unmapped rows are visible at a glance.
+- **Captured test run output.** The actual target runner output, not just "tests passed."
+- **Differential test results.** Where behavior can be captured as input to output, run the same vectors through both implementations and show they agree.
+- **Structural map.** How the source decomposes and where each piece landed in the target. Answers "is the whole feature here?"
+- **Decision log.** Where the agent deviated from the source, what it chose, why, and what alternatives it considered.
+- **New dependencies and capability delta.** Any third-party dependency or system capability (network, filesystem, env vars) the target requires that the source didn't, with justification.
+- **Sensitive-surface diff.** Auto-surfaces code touching credentials, network, subprocess, or deserialization.
+- **Lint, format, and type-check results.** Near-free evidence of convention adherence since these must pass anyway.
+- **Open questions and gaps.** Anything the agent couldn't assert or was unsure about, flagged loudly. A report that surfaces its own gaps earns the benefit of the doubt elsewhere.
+
+## Agents
+
+The workflow splits the translation into a handful of specialized agents that run in sequence, each given only the context relevant to its job. Narrow, role-specific context keeps each agent focused and its prompt small, which tends to improve quality. The roles below are a starting point, not a fixed contract. As we experiment, we can scale the number and granularity of agents up or down (splitting one role into several, or merging a few) based on what produces the best results.
+
+- **Plan.** Works out how to translate the source feature into the target language. Fed the source code, docstrings, tests, and pairing metadata, plus the target language's guides (testing, building, security, documentation) and any target-specific context. Produces the implementation plan the downstream agents follow.
+- **Implement.** Carries out the plan. Fed the plan, the source code, and the target building and security guides. Free to search the web for things like third-party library APIs it needs to write the code.
+- **Validate.** Confirms the implementation is correct, runs the tests, and checks that the source's behaviors are asserted and met. Fed the source tests (the behavior spec), the target implementation and tests, and the testing guide. Sends the workflow back to planning when it finds gaps.
+- **Document.** Updates docstrings as needed and contributes to the user docs. Fed the target implementation and the documentation guide.
+- **Report.** Describes exactly what was done, assembles the review artifacts, and opens the PR. Fed the outputs of every prior agent: the plan, the diff, the test results, and the decisions made along the way.
+
+## Determinism
+
+We want to lean into agents first and see how far automation gets us, then add determinism over time where it earns its place. Every step we make deterministic narrows the agent's scope and trims its context, which tends to help both performance and security.
+
+Committing and pushing code is a good example. Rather than letting the agent decide when and how, we could hard-code those steps and take the responsibility off its plate entirely.
+
+## Experiments
+
+I ran early experiments translating the Anthropic model provider from TypeScript to Python, erasing existing traces from the Python SDK to set realistic conditions. With Opus 4.8 and above, translation quality is strong. The takeaway is that translation itself is not the bottleneck; what's missing is the review infrastructure, which is what shifted focus toward the artifacts and report design in this proposal.
+
+Trust will build incrementally. Each reviewed and approved PR gives us more evidence about where the system is reliable, and over time the review process can become lighter as confidence grows.
+
+## References
+
+- **[Adversarial and cross-language differential testing](https://gist.github.com/agent-of-mkmeral/5a4d0ce16a1242a711d77d7e01c19902#6-adversarial--cross-language-differential-testing).** Design notes on building an agentic multi-language SDK (from @murmeral), whose differential testing section argues for proving behavioral equivalence with failing tests as "non-hallucinatable evidence" rather than agent opinion.
+- **[ReCodeAgent (arXiv:2604.07341)](https://arxiv.org/abs/2604.07341).** Paper on an autonomous multi-agent system that translates and validates entire code repositories across languages, reporting high translation success rates over 118 real-world projects.
--- a/designs/0013-strandslator-spec.md
+++ b/designs/0013-strandslator-spec.md
@@ -1,5 +1,3 @@
-[EDITING]
-
 # Strandslator Spec
 
 **Status**: Proposed
@@ -12,7 +10,7 @@
 
 This document is a pairing to [0012-strandslator](./0012-strandslator-design.md), which focused on the experimentation behind translation and the context the workflow consumes. Here we specify the system's behaviors and interfaces: what a run takes as input, what it produces, the pipeline that drives the agents, how it runs locally through a CLI, and how that same command runs as a GitHub Action. It is a spec, not an implementation guide. It pins down the externally observable contracts and leaves the internal mechanics to the build.
 
-The guiding principle is that there is one way to run the workflow, and everything else is a wrapper around it. The local CLI is the workflow. The GitHub Action is the local CLI running on a hosted runner. A reviewer who wants to take over from a failed automated run picks up exactly where it left off, on their own machine, with the same command.
+The guiding principle is that there is one way to run the workflow, and everything else is a wrapper around it. The local CLI is the workflow. The GitHub Action is the local CLI running on a hosted runner. If an automated run fails, a developer has the option to pick up where it left off on their own machine with the same command.
 
 We lean on [`strands-command`](https://github.com/strands-agents/devtools/tree/main/strands-command) as prior art. It already runs Strands agents in GitHub Actions with job separation, permission isolation, artifact-based state handoff, and authorization gating. It is comment-driven (`/strands` on issues and PRs) rather than `workflow_dispatch`-driven, so it is not a drop-in, but we borrow its proven patterns where they fit.
 
@@ -38,7 +36,7 @@ The pipeline runs the Plan, Implement, Validate, Document, and Report agents in 
 - **Kickbacks.** When Validate finds a behavior gap or failing test, it sends the run back to Plan with the findings attached. We bound iterations so a run can't loop forever, and surface a clear failure when the bound is hit.
 - **Status updates.** Each step emits structured progress (agent, kickback count, pass/fail) so a human can follow along without reading raw transcripts.
 - **Human intervention.** A human can pause a run, inspect or amend the in-progress artifacts and context, and resume.
-- **Pause and resume across shutdown.** We checkpoint pipeline state after each completed step so a run can be stopped, the machine shut down, and picked back up from the last checkpoint. This is the same mechanism that lets a failed GitHub Action be continued locally.
+- **Pause and resume across shutdown.** We checkpoint pipeline state after each completed step so a run can be stopped and picked back up from the last checkpoint, whether that's locally or by re-dispatching in CI.
 
 The pipeline reads input from fixed monorepo paths. Agents may gather additional context on their own (e.g. from the web) during a run.
 
@@ -60,7 +58,7 @@ strandslate resume <run-id>
 strandslate status <run-id>
 ```
 
-The exact flags will firm up during implementation; the requirement is that a developer can drive a complete translation locally and that resume works after a full shutdown.
+The exact flags will firm up during implementation; the requirement is that resume works after a full shutdown.
 
 ## Action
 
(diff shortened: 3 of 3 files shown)


designs/0012-strandslator.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Determinism
+
+We want to lean into agents first and see how far automation gets us, then add determinism over time where it earns its place. Every step we make deterministic narrows the agent's scope and trims its context, which tends to help both performance and security.
+
+Committing and pushing code is a good example. Rather than letting the agent decide when and how, we could hard-code those steps and take the responsibility off its plate entirely.
+
+## Experiments
+
+I ran early experiments translating the Anthropic model provider from TypeScript to Python, erasing existing traces from the Python SDK to set realistic conditions. With Opus 4.8 and above, translation quality is strong. The takeaway is that translation itself is not the bottleneck; what's missing is the review infrastructure, which is what shifted focus toward the artifacts and report design in this proposal.
```
yonib05: Do we have objective data we can cite for this?

designs/0012-strandslator.md
```diff
@@ -0,0 +1,100 @@
+# Strandslator
```
yonib05: Overall I agree with the doc but I am not seeing an actionable plan that we can start implementing. From my perspective, we need the MVP to get this off the ground, then we can fill in missing pieces and improve.

pgrayy: I'll add a more detailed section on this. But in short, getting something setup will be straightforward. We can follow closely what we already do with the GitHub agents. The tricky part of this system is just making sure we are consistently using it so that the code generation itself is consistent. As stated in the doc, Python will continue to be the outlier because it utilizes some of our outdated design principles.

designs/0012-strandslator.md
pgrayy: I have my sample experiments but not something super extensive. This was due to timing though. Each translation workflow I kicked off ran on average 20 minutes and so I could only complete so much. I did however run an experiment with the Bedrock KB port which is committed in main now. I'll add the findings of that experiment to this doc. Overall it worked well. There were a few hiccups though because of general inconsistencies across the Python SDK, which understandably confused the agents.

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,100 @@
...
+
+**Issue**: https://github.com/strands-agents/harness-sdk/issues/2666
+
+## Overview
+
+Every feature we build in one language eventually needs to exist in all the others. If Strands supports N languages with M features, keeping them in sync becomes an N * M cost.
+
+Strandslator is a proposed system that uses agents to handle feature translation across languages. The bottleneck we're targeting isn't necessarily the translation itself, it's the review. An agent can produce the code, but a human still needs to verify its correctness, security, and idiomaticity. That verification is what takes time, and it's what we want to make fast. The system's design centers on giving reviewers the evidence and artifacts they need to approve a translation PR with both confidence and speed.
+
+This document describes the translation workflow, the input context, the agents involved, and the artifacts they generate. Orchestration will be discussed separately, but likely we build upon the existing GitHub agent infrastructure we already have in place ([ideas](https://gist.github.com/agent-of-mkmeral/5a4d0ce16a1242a711d77d7e01c19902#6-adversarial--cross-language-differential-testing)).
```
mkmeral, line 15: I don't think we should build it on top of the current architecture. I think it's limited, and has too many permissions

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,100 @@
...
+## Solution
+
+Strandslator takes one completed source feature and one target language as input. The developer who completes the feature is the one who kicks off the workflow. Triggering is manual to start, since having a human make the call ensures the feature is genuinely ready before translation begins. We can explore automating this over time, but for now the human judgment is worth it.
+
+We scope each run to a single target language so we can inject language-specific context into the agents and keep the run focused enough to produce a reviewable PR. The workflow runs a sequence of agents that plan, implement, and validate the translation, restarting from the top if validation finds issues. Once everything passes, it produces documentation and assembles the review artifacts into a PR.
+
+```mermaid
+flowchart LR
+    src["Source feature<br/>(code, tests, docs, prompt)"] --> plan[Plan]
+    plan --> impl[Implement]
```
mkmeral, line 26: why do we need a separate plan? yolo? :)

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,100 @@
...
+    impl --> val[Validate]
+    val -->|tests fail / behavior gap| plan
+    val --> doc[Document]
+    doc --> report[Report]
+    report --> pr["Target-language PR<br/>+ review artifacts"]
+```
+
+This workflow is only as good as the features it translates. A few principles set the preconditions and constrain how it operates:
+
+- **No true source language.** A feature can originate in any language and translate to all others. Python however requires care, since some features touch components that follow outdated design patterns we no longer want to propagate. Features that are isolated enough from those components may translate just fine.
```
mkmeral, line 36: but won't behaviors diverge, and cause confusion of the mental model of the sdk?

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,100 @@
...
+- **Translate, don't improve.** The agent reproduces behavior as-is. It does not fix unrelated code or redesign the feature.
+- **Idiomatic over literal, when confident.** Prefer a translation idiomatic to the target language. When uncertain, a literal translation is a fine starting point. The reviewer can always provide more context to guide the next iteration.
+- **Idiomatic surface over a portable core.** Build features with clean object-oriented design and a core that translates well across languages. Per-language deviations should be minimal and focused on developer experience, not functionality. For example, Python's `@tool` decorator is an idiomatic surface that returns a `FunctionTool` instance under the hood, and `FunctionTool` is the portable core that exists across languages.
+- **Design for portability.** When designing a feature, we should be mindful of how well it will translate. Clean, well-structured implementations translate more straightforwardly, and the agent shouldn't have to work around ambiguous or tangled code. If a feature is hard for the agent to translate, that's a signal the source needs cleanup before it's marked ready.
```
mkmeral, line 42: is this something we can scale through decisions.md?

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,100 @@
...
+
+- **Source.** The feature code with API docstrings, pairing tests, and a metadata markdown file. Tests should use docstrings to describe the behavior being asserted when it isn't obvious from the assertions alone. The metadata file sits next to the feature module or in its subdirectory and explains why the feature is designed the way it is, including what we consider subjective. The code is the true spec, so this file should stay concise. We separate it from docstrings because docstrings are about public API usage, not internal implementation rationale. When feature code is updated, pairing metadata and tests must be updated too. The translation workflow should never have to work around conflicting context; that is a problem for the feature design workflow.
```
mkmeral, line 51: can't we just add a bunch of inline comments instead? would that be easier?

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,100 @@
...
+- **New dependencies and capability delta.** Any third-party dependency or system capability (network, filesystem, env vars) the target requires that the source didn't, with justification.
+- **Sensitive-surface diff.** Auto-surfaces code touching credentials, network, subprocess, or deserialization.
+- **Lint, format, and type-check results.** Near-free evidence of convention adherence since these must pass anyway.
+- **Open questions and gaps.** Anything the agent couldn't assert or was unsure about, flagged loudly. A report that surfaces its own gaps earns the benefit of the doubt elsewhere.
+
+## Agents
+
+The workflow splits the translation into a handful of specialized agents that run in sequence, each given only the context relevant to its job. Narrow, role-specific context keeps each agent focused and its prompt small, which tends to improve quality. The roles below are a starting point, not a fixed contract. As we experiment, we can scale the number and granularity of agents up or down (splitting one role into several, or merging a few) based on what produces the best results.
```
mkmeral, line 77: do we need different agents, or a single agent (with use agent tool and skills)?

mkmeral: also do these agents use the same environment, or do they share the branch?

designs/0012-strandslator.md
mkmeral: fwiw I'd be in favor of just reviewing your experimental setup and iterating on that

designs/0012-strandslator-design.md
pgrayy: Yeah this can be scaled up or down as needed. I will say though, up to this point, we have seen better success with multi-agent architectures.

designs/0012-strandslator-design.md
pgrayy: I was considering human readability as well. Adding too many inline comments I feel can make the code hard to navigate which is why I was thinking a separate metadata file.

designs/0012-strandslator.md
pgrayy: I just added more details to the experiments section outline some issues I ran into using the system for porting Bedrock KB.

pgrayy: See also the new spec doc.

designs/0012-strandslator-design.md
mkmeral: +1 but it doesn't have to be well defined. @awsarron had a tasks tool i think that essentially created a multiagent architecture. or we can use use_agent toolwith skills as custom system prompts

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,109 @@
...
+
+**Date**: 2026-06-12
+
+**Issue**: https://github.com/strands-agents/harness-sdk/issues/2666
+
+## Overview
+
+Every feature we build in one language eventually needs to exist in all the others. If Strands supports N languages with M features, keeping them in sync becomes an N * M cost.
+
+Strandslator is a proposed system that uses agents to handle feature translation across languages. The bottleneck we're targeting isn't necessarily the translation itself, it's the review. An agent can produce the code, but a human still needs to verify its correctness, security, and idiomaticity. That verification is what takes time, and it's what we want to make fast. The system's design centers on giving reviewers the evidence and artifacts they need to approve a translation PR with both confidence and speed.
```
zastrowm, line 13: > That verification is what takes time, and it's what we want to make fast

Doc feedback: This makes it sound like this is a pitch for a reviewer rather than translator where I do expect this to be a full-cycle kind of solution

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,109 @@
...
+
+This document describes the translation workflow, the input context, the agents involved, and the artifacts they generate. Orchestration will be discussed separately, but likely we build upon the existing GitHub agent infrastructure we already have in place ([ideas](https://gist.github.com/agent-of-mkmeral/5a4d0ce16a1242a711d77d7e01c19902#6-adversarial--cross-language-differential-testing)).
+
+## Solution
+
+Strandslator takes one completed source feature and one target language as input. The developer who completes the feature is the one who kicks off the workflow. Triggering is manual to start, since having a human make the call ensures the feature is genuinely ready before translation begins. We can explore automating this over time, but for now the human judgment is worth it.
```
opieter-aws, line 19: >  one completed source feature and one target language as input

How scalable is this to multiple languages, multiple features, or both?

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,109 @@
...
+
+```mermaid
+flowchart LR
+    src["Source feature<br/>(code, tests, docs, prompt)"] --> plan[Plan]
+    plan --> impl[Implement]
+    impl --> val[Validate]
+    val -->|tests fail / behavior gap| plan
+    val --> doc[Document]
+    doc --> report[Report]
+    report --> pr["Target-language PR<br/>+ review artifacts"]
```
opieter-aws, line 31: Is it worth doing another iteration based on a reviewer agent here?

designs/0012-strandslator-design.md
zastrowm: > Python however requires care

I do think it's worth calling out that languages will sometimes diverge and the design should take this into account (I presume we do).  But I am aligned that we evolve the languages together

designs/0013-strandslator-spec.md
```diff
@@ -0,0 +1,91 @@
...
+
+We borrow proven patterns from all three where they fit. Opinionated agent templates matter here: agents pre-configured with sensible defaults (caching, model selection, tool access, prompt structure) are a time and cost saver. When I was prototyping Strandslator, I forgot to enable caching on the Strands agent, which meant unnecessary cost and latency on every run.
+
+## Contract
+
+A run is defined by what it takes in and what it puts out.
+
+**Input.** A source feature and a target language. Both resolve to locations in the monorepo, and the source carries its paired tests, docstrings, and metadata, plus the "strands-md" guides, exactly as specified in [0012](./0012-strandslator-design.md). There are no GitHub issues or other out-of-band inputs.
+
+A run also depends on a precondition: the source feature must be **marked ready for translation** per the "Explicit readiness" principle in [0012](./0012-strandslator-design.md). This spec only requires that the signal exists and that a run refuses to start against a source that is not marked ready.
```
mkmeral, line 31: + I'd say we need dependency check. for example a new memory injection feature depends on middleware. should the agent develop both in one go?

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+This workflow is only as good as the features it translates. A few principles set the preconditions and constrain how it operates:
+
+- **No true source language.** A feature can originate in any language and translate to all others. Python however requires care, since some features touch components that follow outdated design patterns we no longer want to propagate. Features that are isolated enough from those components may translate just fine.
+- **Translate only completed features.** Complete means merged, tested, and documented. Code, tests, and docs must be in sync before translation kicks off. Anytime someone updates an existing feature, they update all paired files first.
```
zastrowm, line 37: This seems focused on features; the plan is the same for bug fixes too, right?

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,109 @@
...
+- **Explicit readiness.** Features may be developed or updated iteratively and so we don't want the workflow guessing whether work is done. We explicitly mark a feature as ready for translation once complete and all paired files are in sync.
+- **Translate, don't improve.** The agent reproduces behavior as-is. It does not fix unrelated code or redesign the feature.
```
gautamsirdeshmukh, line 39: I agree the strandslator shouldn't go down rabbitholes related to optimizations/refactors - though it would be nice if it at least noted them somewhere (or in the future, maybe even kick off a separate implementation pipeline)

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,109 @@
...
+- **Design for portability.** When designing a feature, we should be mindful of how well it will translate. Clean, well-structured implementations translate more straightforwardly, and the agent shouldn't have to work around ambiguous or tangled code. If a feature is hard for the agent to translate, that's a signal the source needs cleanup before it's marked ready.
+- **Fix the system, not the instance.** Generalize failures into rules and update the system's context. If the root cause is upstream (ambiguous source, stale metadata), fix the feature development workflow. One-off corrections don't compound; system improvements do.
+
+## Context
+
+All input context lives in the monorepo. No GitHub issues are needed. During a run however, the agents are free to collect additional context on their own from external sources such as the web.
```
opieter-aws, line 47: Is it possible to extend the input with some human guidance, so we can steer the implementation if we know there are translation caveats for this feature?

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,109 @@
...
+
+**Status**: Proposed
+
+**Date**: 2026-06-12
+
+**Issue**: https://github.com/strands-agents/harness-sdk/issues/2666
+
+## Overview
+
+Every feature we build in one language eventually needs to exist in all the others. If Strands supports N languages with M features, keeping them in sync becomes an N * M cost.
```
notowen333, line 11: In practice, with the assumption every language starts at parity I'd think of the cost as just N because M is always 1 in the context of what gets translated

notowen333: Not important though xd

I think we all understand the scope of the domain here

designs/0012-strandslator-design.md
poshinchen, line 39: What I notice is that some behaviors between python and typescript could be different. Does this mean we want to make a working workaround instead of improving a (if) bad experience?

designs/0012-strandslator-design.md
pgrayy: To start I feel this should be fine since we only have Python and TS. Even if we add on a third, if a feature developer finishes in one language, they can kick off 2 parallel jobs to translate into the other 2 and then share those PRs. I feel even that is still fairly manageable. The real concern though is from the reviewer end. How quickly are they going to be able to scan all the PRs to have the confidence to approve. That confidence will take time but I would imagine if we get into the realm of 4 or 5 languages, we should have iterated enough on the translation workflow to have enough confidence to sign off on PRs quickly.

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,109 @@
...
+This document describes the translation workflow, the input context, the agents involved, and the artifacts they generate. Orchestration will be discussed separately, but likely we build upon the existing GitHub agent infrastructure we already have in place ([ideas](https://gist.github.com/agent-of-mkmeral/5a4d0ce16a1242a711d77d7e01c19902#6-adversarial--cross-language-differential-testing)).
+
+## Solution
+
+Strandslator takes one completed source feature and one target language as input. The developer who completes the feature is the one who kicks off the workflow. Triggering is manual to start, since having a human make the call ensures the feature is genuinely ready before translation begins. We can explore automating this over time, but for now the human judgment is worth it.
+
+We scope each run to a single target language so we can inject language-specific context into the agents and keep the run focused enough to produce a reviewable PR. The workflow runs a sequence of agents that plan, implement, and validate the translation, restarting from the top if validation finds issues. Once everything passes, it produces documentation and assembles the review artifacts into a PR.
```
JackYPCOnline, line 21: in Kiro, a simple workflow would take longer than expectation, the more detailed plan is, the longer it takes. Maybe more human-in-the-loop?

designs/0012-strandslator-design.md
zastrowm: >  The code is the true spec

This is true from the standpoint of "Your task is translation not invention" but the code can also be wrong/not match what our expectations are.  I encountered this a little bit with middleware to get parity and ended up updating the markdown "spec" along with the code.

I guess my point is the agent should also raise things that seem odd - maybe that's a reviewer problem - but it's a lot easier to see when the spec is markdown IMO

designs/0012-strandslator-design.md
pgrayy: That makes sense.

designs/0012-strandslator-design.md
notowen333, line 19: Could we also think about picking one language as the SOT and then kicking off translations automatically on merge?

It might help our organization to know which language is the canonical one.

Unless the goal is that every feature ships to every language at once

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,109 @@
...
+- **`MemoryStoreConfig` was a Protocol, not a TypedDict.** From a separate earlier port. Protocols can't be spread into constructor kwargs, so the agent couldn't compose with the shared config type.
+- **No uuid7 in Python.** The TS source uses uuid v7 for document IDs; Python's stdlib has no v7 until 3.14, so the agent used uuid4. A genuine language gap, not a codebase consistency issue.
+- **Skipped boto3-stubs[s3].** Adding the stub would have tightened typing repo-wide (boto3-stubs injects overloads via package install, not per-import), exposing a pre-existing error in an unrelated file. The agent surfaced the issue and correctly refused to fix code outside the feature, but that forced human intervention.
+
+All but uuid7 trace to inconsistencies in the Python codebase. The agent was influenced by conflicting precedents and made defensible-but-wrong choices. With consistency going forward, this port reduces to a single decision (uuid7 vs uuid4). That is the concrete payoff of "Design for portability" and "Fix the system, not the instance": consistency means less human intervention, and the intervention that remains is about genuine language gaps rather than inherited drift.
```
opieter-aws, line 104: We should hook up a memory store to the translation agent, so it remembers it when we correct these types of decisions

designs/0012-strandslator-design.md
pgrayy: Yeah the reviewer could be placed into this pipeline. Probably Report could just fold into Reviewer.

designs/0013-strandslator-spec.md
```diff
@@ -0,0 +1,91 @@
...
+- **Resume on failure.** A failed run can be resumed by re-dispatching in CI or by downloading the checkpoint locally and running `strandslate resume`.
+
+The checkpoint needs to be persisted somewhere accessible for resume. In GitHub Actions, `actions/upload-artifact` is one option (with `if: always()` so failed runs still publish). A shared location like S3 is another, and would make checkpoints accessible regardless of where the run executed. The exact storage is runtime-dependent; what matters is that the checkpoint is retrievable after the run ends.
+
+## Context
+
+The pipeline is only as good as the context it reads. Before the workflow can produce reviewable PRs, we need to get the "strands-md" layer and per-feature metadata into shape.
+
+- **Split up AGENTS.md.** The current files mix concerns. Split into focused guides by role (testing, building, security, documentation) at the fixed paths [0012](./0012-strandslator-design.md) assumes, with shared guidance at the repo root and language-specific variants in each package.
```
mkmeral, line 81: are these skills and resources?

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,109 @@
+# Strandslator Design
```
lizradway, line 1: not to beat a dead horse, but i don't think i was there for this conversation and am curious to hear reasoning.

just in general, why not use the strandslate pipeline to translate TS codebase into rust & generate bindings like [cedar agents](https://github.com/cedar-policy/cedar-for-agents/tree/main/rust) + [main cedar](https://github.com/cedar-policy/cedar)/[goose](https://github.com/aaif-goose/goose/discussions/9811)? from my (albiet brief) experience generating [python bindings via pyO3 for cedar mcp](https://github.com/cedar-policy/cedar-for-agents/pull/139), this could easily be automated by a specific agent as well (kicked off by CI even) if we'd prefer and would ensure determinism across languages/completely eliminate the review process + discrepencies between languages.

designs/0012-strandslator-design.md
pgrayy: This concern is meant to be addressed by the `Idiomatic surface over a portable core.` rule.

designs/0012-strandslator-design.md
pgrayy: Yes correct. Bug fixes as well.

designs/0012-strandslator-design.md
pgrayy: Good call out. Yes this could be part of the artifacts it generates so we can create an issue out of it.

designs/0012-strandslator-design.md
notowen333, line 37: Are we going to have an expected timeframe for translations to be merged? Best effort? Does it impact release thinking

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,109 @@
...
+- **Captured test run output.** The actual target runner output, not just "tests passed."
+- **Differential test results.** Where behavior can be captured as input to output, run the same vectors through both implementations and show they agree.
+- **Structural map.** How the source decomposes and where each piece landed in the target. Answers "is the whole feature here?"
+- **Decision log.** Where the agent deviated from the source, what it chose, why, and what alternatives it considered.
+- **New dependencies and capability delta.** Any third-party dependency or system capability (network, filesystem, env vars) the target requires that the source didn't, with justification.
+- **Sensitive-surface diff.** Auto-surfaces code touching credentials, network, subprocess, or deserialization.
+- **Lint, format, and type-check results.** Near-free evidence of convention adherence since these must pass anyway.
+- **Open questions and gaps.** Anything the agent couldn't assert or was unsure about, flagged loudly. A report that surfaces its own gaps earns the benefit of the doubt elsewhere.
```
zastrowm, line 73: Of all the items here, Open questions seems to be a priority with the rest almost being noise.  I'm a little afraid that the artifacts here are overwhelming or more noise to dig through rather than the diff.   I guess the open questions is what I'd say is the artifact, and everything else seems more like "an audit log"

designs/0013-strandslator-spec.md
```diff
@@ -0,0 +1,91 @@
...
+- **Update guidelines with current learnings.** Fold what we've learned from experiments and ongoing port work back into the guides so the agents inherit it.
```
opieter-aws, line 82: Consider searchable memory

designs/0012-strandslator-design.md
pgrayy: Yes I should have called that out more clearly. It is the intention that the human gets a chance to add their input into the prompts.

designs/0012-strandslator-design.md
zastrowm: I do explicitly like the idea of validation being a separate agent than the implementor (e.g. fresh context); implementation I'm less concerned

designs/0012-strandslator-design.md
notowen333, line 39: So what is our flow if we catch a bad decision in the source language while translating?

I'd think we want to fix it everywhere

OK for the impl of the translator to avoid re-writes during initial translation

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,109 @@
...
+- **Implement.** Carries out the plan. Fed the plan, the source code, and the target building and security guides. Free to search the web for things like third-party library APIs it needs to write the code.
+- **Validate.** Confirms the implementation is correct, runs the tests, and checks that the source's behaviors are asserted and met. Fed the source tests (the behavior spec), the target implementation and tests, and the testing guide. Sends the workflow back to planning when it finds gaps.
+- **Document.** Updates docstrings as needed and contributes to the user docs. Fed the target implementation and the documentation guide.
+- **Report.** Describes exactly what was done, assembles the review artifacts, and opens the PR. Fed the outputs of every prior agent: the plan, the diff, the test results, and the decisions made along the way.
+
+## Determinism
+
+We want to lean into agents first and see how far automation gets us, then add determinism over time where it earns its place. Every step we make deterministic narrows the agent's scope and trims its context, which tends to help both performance and security.
```
zastrowm, line 87: The graph above is determinism right off the bat which begs the question "How did it earn its place" :) 

(I think it's a good start, but it begged the question)

designs/0012-strandslator-design.md
opieter-aws: I see that for keeping things in-sync, but if we're talking about a new _language_ we would be translating all existing features right?

lizradway: so Strandslator will run N x M times? I think I'd also rather have this tackle N languages at once and only run M times.

designs/0012-strandslator-design.md
zastrowm, line 104: > With consistency going forward

I don't want to plan for consistency given what we have now.  We can work towards it, but if it's 3mo away we should focus on this.  I think you mentioned the answer to this before was more documentation/guidelines.

designs/0012-strandslator-design.md
pgrayy: The goal would be that as part of your feature development/feature update/bug fix, you address in one language and before calling the work done, you kick off the translation workflow so that everything remains consistent together for each release. So translation becomes another checkbox for completion on feature development just like we have for adding tests and documentation.

designs/0012-strandslator-design.md
JackYPCOnline: Agree ^ memory manager in python has implementation detail diff with TS, there are still some language specific follow-up, maybe agent could summerize the diff for easier review

designs/0012-strandslator-design.md
lizradway: > A feature can originate in any language and translate to all others. 

Will this cause any conflicts in merge between languages? How do we aim to tackle this? ie. if owen is working on a feature in python -> N and I am working on a feature in TS -> N.

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,109 @@
...
+
+Artifacts must be mechanically derived, not narrated. Since the agent writes its own report, anything it merely asserts carries no weight. What builds confidence is evidence the agent can't fake: actual test output, actual diffs, conventions the reviewer can spot-check. The report indexes and explains that evidence.
+
+The exact schema is not yet decided and will evolve as we learn what reviewers actually need. The overriding goal is human readability. Any time a reviewer is confused about something on a PR, that signals a gap in the report to fill. This list is by no means exhaustive, but the initial set of artifacts we think are worth highlighting:
+
+- **Behavior traceability matrix.** Maps each source test to the target test that asserts the same behavior, with its run result. Missing or unmapped rows are visible at a glance.
+- **Captured test run output.** The actual target runner output, not just "tests passed."
```
notowen333, line 66: We mean unit + integ tests here?

designs/0012-strandslator-design.md
pgrayy: Ideally before the next release. As noted in my comment above (should have called out in the doc), I would like translation to become another checkbox for marking a feature/bug fix complete.

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,109 @@
...
+- **New dependencies and capability delta.** Any third-party dependency or system capability (network, filesystem, env vars) the target requires that the source didn't, with justification.
+- **Sensitive-surface diff.** Auto-surfaces code touching credentials, network, subprocess, or deserialization.
+- **Lint, format, and type-check results.** Near-free evidence of convention adherence since these must pass anyway.
+- **Open questions and gaps.** Anything the agent couldn't assert or was unsure about, flagged loudly. A report that surfaces its own gaps earns the benefit of the doubt elsewhere.
+
+## Agents
+
+The workflow splits the translation into a handful of specialized agents that run in sequence, each given only the context relevant to its job. Narrow, role-specific context keeps each agent focused and its prompt small, which tends to improve quality. The roles below are a starting point, not a fixed contract. As we experiment, we can scale the number and granularity of agents up or down (splitting one role into several, or merging a few) based on what produces the best results.
```
JackYPCOnline, line 77: Can we auto run during night so that we can review in the morning for some simple translations dont need human input

designs/0012-strandslator-design.md
lizradway: will the agent ever pause for HITL / is there a way to steer while running?

designs/0012-strandslator-design.md
pgrayy: Yes I would want us to be careful with the artifacts we present. I want them to be maximally minimal. I don't want noise. That's why I think we should start really small and add on to them through true need and experience.

designs/0012-strandslator-design.md
arronbailiss: > We should hook up a memory store to the translation agent, so it remembers it when we correct these types of decisions

could we codify those in the monorepo instead?

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,109 @@
...
+- **Captured test run output.** The actual target runner output, not just "tests passed."
+- **Differential test results.** Where behavior can be captured as input to output, run the same vectors through both implementations and show they agree.
+- **Structural map.** How the source decomposes and where each piece landed in the target. Answers "is the whole feature here?"
+- **Decision log.** Where the agent deviated from the source, what it chose, why, and what alternatives it considered.
+- **New dependencies and capability delta.** Any third-party dependency or system capability (network, filesystem, env vars) the target requires that the source didn't, with justification.
+- **Sensitive-surface diff.** Auto-surfaces code touching credentials, network, subprocess, or deserialization.
+- **Lint, format, and type-check results.** Near-free evidence of convention adherence since these must pass anyway.
+- **Open questions and gaps.** Anything the agent couldn't assert or was unsure about, flagged loudly. A report that surfaces its own gaps earns the benefit of the doubt elsewhere.
+
```
notowen333, line 74: So what is the workflow for after this translation is done and artifacts are produced?

Does it raise a PR? What happens after comments come back on the PR (and revisions too)

designs/0013-strandslator-spec.md
```diff
@@ -0,0 +1,91 @@
...
+- **Source and target inputs.** Exposed as dispatch inputs, passed straight through to the command.
+- **Authorization gating.** Following `strands-command`, the workflow gates dispatch against an allowlist of repo roles (`maintain`, `write`, `admin`), with a manual approval gate for anyone else.
+- **Permission separation.** The agent runs in a read-only job. Repository changes and deferred GitHub operations are emitted as artifacts, and a separate Finalize job (with write permission, `if: always()`) pushes the branch and opens the PR. This keeps commit and push off the agent's plate and limits blast radius.
+- **Resume on failure.** A failed run can be resumed by re-dispatching in CI or by downloading the checkpoint locally and running `strandslate resume`.
+
+The checkpoint needs to be persisted somewhere accessible for resume. In GitHub Actions, `actions/upload-artifact` is one option (with `if: always()` so failed runs still publish). A shared location like S3 is another, and would make checkpoints accessible regardless of where the run executed. The exact storage is runtime-dependent; what matters is that the checkpoint is retrievable after the run ends.
+
+## Context
```
mkmeral, line 77: what mechanisms/processes will we have to improve the agent? can we have an audit agent that goes through the PRs to create repos, propose improvements to skills, processes, docs, etc.?

designs/0012-strandslator-design.md
gautamsirdeshmukh: Along with open questions, I think the behavior traceability matrix could be really useful - for translation tasks having a clear/visual comparison of the conceptual flow of each implementation backed by testing would definitely a bunch of time

Would be interesting to see how the decision log could be analyzed/fed back to the agent to improve future runs

designs/0013-strandslator-spec.md
pgrayy: Resources. If we create dedicated workflows I feel we wouldn't need too many skills since we can provide the more targeted context deterministically up front rather than have the agents decide.

designs/0012-strandslator-design.md
arronbailiss: if I have a diff as the final artifact, I don't think I really care about how the agent implemented things for auditing. At most I'd be interested in traces so that we can improve the agent's behavior, evaluate, and optimize

designs/0012-strandslator-design.md
gautamsirdeshmukh: +1 had a similar comment about the decision log above, agreed

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,109 @@
...
+- **Document.** Updates docstrings as needed and contributes to the user docs. Fed the target implementation and the documentation guide.
+- **Report.** Describes exactly what was done, assembles the review artifacts, and opens the PR. Fed the outputs of every prior agent: the plan, the diff, the test results, and the decisions made along the way.
+
+## Determinism
+
+We want to lean into agents first and see how far automation gets us, then add determinism over time where it earns its place. Every step we make deterministic narrows the agent's scope and trims its context, which tends to help both performance and security.
+
+Committing and pushing code is a good example. Rather than letting the agent decide when and how, we could hard-code those steps and take the responsibility off its plate entirely.
+
+## Experiments
```
poshinchen, line 91: How do we validate the usefulness / success of strandslator? Some kind of metrics? Wanna use the devtools dashboard for viewing what's been done or something like that?

designs/0012-strandslator-design.md
arronbailiss: moving quickly on this, starting simple, iterating seems best to me. We can always add more agents, different workflows, remove agents, change their tools and behaviors etc. very easily

designs/0012-strandslator-design.md
lizradway, line 47: > All input context lives in the monorepo

throwing it out there, that if we get serious about separate autonomous agent-driven development, we should consider a separate repo specific to agent context a la github based memory.

designs/0012-strandslator-design.md
arronbailiss: have some examples that spring to mind?

designs/0012-strandslator-design.md
zastrowm: I think we talked about this earlier, but I think one of the reasons @pgrayy chose this model was the idea that a contributor might choose to contribute in language P and then we have to translate to T; so we can choose the canonical, but we didn't want to force contributors to know/use the canonical one for contributions

designs/0012-strandslator-design.md
arronbailiss: once we have a PR to review, can I just comment and review the PR like normal, then agent picks it up? How does that flow work?

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,109 @@
...
+
+- **Source.** The feature code with API docstrings, pairing tests, and a metadata markdown file. Tests should use docstrings to describe the behavior being asserted when it isn't obvious from the assertions alone. The metadata file sits next to the feature module or in its subdirectory and explains why the feature is designed the way it is, including what we consider subjective. The code is the true spec, so this file should stay concise. We separate it from docstrings because docstrings are about public API usage, not internal implementation rationale. When feature code is updated, pairing metadata and tests must be updated too. The translation workflow should never have to work around conflicting context; that is a problem for the feature design workflow.
```
lizradway, line 51: > Tests should use docstrings to describe the behavior being asserted when it isn't obvious from the assertions alone.

are we going to take any steps to ensure our code base is up to par with this standard as the starting point? I don't think we do this nearly enough rn

designs/0012-strandslator-design.md
JackYPCOnline, line 91: Did stransalator add a lot inline comments or verbose docstrings?

designs/0012-strandslator-design.md
arronbailiss, line 51: what's better, metadata/decisions.md style files alongside the code files, or in separate docs/decisions type directories that sit at a higher level?

designs/0012-strandslator-design.md
```diff
@@ -0,0 +1,109 @@
...
+- **Testing.** Cross-language rules for how we define tests. For example, we follow a one-to-one directory layout where source module `a` has a pairing test module `a`. Language-specific conventions layer on top. In Python for example, we use pytest with fixtures where shareable fixtures live in `conftest.py`.
+- **Building.** Covers environment setup, how to run tests, and what checks must pass before a PR is submitted (linting, formatting, etc.). Language-specific tooling is called out where relevant. In Python for example, we use hatch for environment and task management.
+- **Security.** Guidance on writing secure code and common pitfalls to avoid, such as leaking credentials, exposing sensitive data in logs, or accessing restricted paths.
+- **Documentation.** General documentation standards, with language-specific variants where conventions differ.
+
+## Artifacts
+
+Each run produces the translated code, tests, and docs, plus a review report attached to the PR. The report is the heart of this proposal. Its job is to give the reviewer confidence that the translation is correct, safe, and idiomatic, without having to reverse-engineer that confidence from the diff alone.
```
lizradway, line 59: i like the report idea a lot.

lizradway: can we also pull out metrics from the report?

designs/0012-strandslator-design.md
zastrowm: +1 or more generally "more written historical artifacts".

https://github.com/strands-agents/harness-sdk/blob/ef0ed3b3df5c6383d9c6082895aa87e52b2adc1a/strands-ts/src/middleware/README.md#behavioral-requirements helped me refine how middleware should write but also served as a "conformance" suite for both languages

designs/0012-strandslator-design.md
lizradway: metadata is nice imo big fan

designs/0012-strandslator-design.md
lizradway: everyone is rewriting in rust in a week tbh: https://github.com/oven-sh/bun/pull/30412

mehtarac (approved):

PR #2795, opened 2026-06-15
feat(memory): add Agent memory_manager param with configurable sync auto-flush

JackYPCOnline:
## Description

Adds a dedicated `memory_manager` parameter to `Agent`, mirroring the TypeScript SDK's `Agent` constructor, and makes cross-session memory persistence work out of the box for the synchronous entry point.

Previously a `MemoryManager` could only be attached via `plugins=[...]`, and persisting background extraction across the synchronous `Agent(...)` entry point required the caller to opt in (the original `flush_on_invocation_end` defaulted to `False`) or to call `flush()` manually. That left a quiet footgun: each synchronous invocation runs in its own event loop (`run_async` → `asyncio.run`), which cancels in-flight background saves on close, so the last turn's extraction was typically lost unless the user knew to enable the flag.

This change gives `Agent` a first-class `memory_manager` param (matching `conversation_manager` / `session_manager`) and flushes pending extraction automatically — but only on the synchronous path, so it never forces a save every turn on the async path or interferes with `IntervalTrigger` cadence.

## Public API Changes

`Agent` accepts a `memory_manager` (a `MemoryManager` instance or a `MemoryManagerConfig`, auto-wrapped) and exposes it as `agent.memory_manager`:

```python
from strands import Agent

agent = Agent(model=model, memory_manager=MemoryManager(stores=[my_store]))
agent("Remember I prefer dark mode")  # extraction is flushed before the loop closes
```

`flush_on_invocation_end` (on `MemoryManager` / `MemoryManagerConfig`) has been removed. The synchronous `Agent(...)` entry point now always awaits pending extraction after each invocation; the async `invoke_async` path never auto-flushes, so callers driving their own loop continue to `await memory_manager.flush()` at a shutdown boundary.

Auto-flush covers the `Agent(...)` entry point only. The deprecated `structured_output()` sync path is not wired (it doesn't append messages, so extraction records nothing to flush there).

## Breaking Changes

The `flush_on_invocation_end` parameter that shipped in the initial memory port has been removed (same unreleased development cycle). Persistence across the synchronous entry point is now automatic and unconditional, so there is no flag to migrate. Callers needing extraction-cadence control should use `invoke_async` and `flush()` directly.

## Testing

Ported the three TS test files (`memory-manager`, `extraction`, `model-extractor`) as example-based `pytest` suites so coverage mirrors the TS suite case-for-case; runs green under `tests/strands/memory`. Adjacent `plugins`/`tools` suites were run to confirm no regressions in plugin/tool discovery.

## Checklist

### DIFF
--- a/strands-py/src/strands/agent/agent.py
+++ b/strands-py/src/strands/agent/agent.py
@@ -506,8 +506,9 @@ class Agent(AgentBase):
 
         return resolved_conversation_manager, resolved_plugins
 
+    @staticmethod
     def _resolve_memory_manager(
-        self, memory_manager: MemoryManager | MemoryManagerConfig | None
+        memory_manager: MemoryManager | MemoryManagerConfig | None,
     ) -> MemoryManager | None:
         """Resolve the ``memory_manager`` argument into a MemoryManager instance or None.
 
--- a/strands-py/src/strands/agent/agent.py
+++ b/strands-py/src/strands/agent/agent.py
@@ -42,6 +42,7 @@ from ..types._snapshot import (
 )
 
 if TYPE_CHECKING:
+    from ..memory import MemoryManager, MemoryManagerConfig
     from ..tools import ToolProvider
 from .._middleware import MiddlewareRegistry
 from ..handlers.callback_handler import PrintingCallbackHandler, null_callback_handler
@@ -166,6 +167,7 @@ class Agent(AgentBase):
         hooks: list[HookProvider | HookCallback] | None = None,
         interventions: list[InterventionHandler] | None = None,
         session_manager: SessionManager | None = None,
+        memory_manager: "MemoryManager | MemoryManagerConfig | None" = None,
         structured_output_prompt: str | None = None,
         tool_executor: ToolExecutor | None = None,
         retry_strategy: ModelRetryStrategy | _DefaultRetryStrategySentinel | None = _DEFAULT_RETRY_STRATEGY,
@@ -240,6 +242,12 @@ class Agent(AgentBase):
                 Defaults to None.
             session_manager: Manager for handling agent sessions including conversation history and state.
                 If provided, enables session-based persistence and state management.
+            memory_manager: Cross-session memory manager. Accepts a
+                :class:`~strands.memory.MemoryManager` instance or a
+                :class:`~strands.memory.MemoryManagerConfig` (auto-wrapped). When set, its
+                search/add tools are registered and the synchronous ``Agent(...)`` entry point
+                awaits pending extraction writes after each invocation unless the manager sets
+                ``flush_on_invocation_end=False``. Defaults to None.
             structured_output_prompt: Custom prompt message used when forcing structured output.
                 When using structured output, if the model doesn't automatically use the output tool,
                 the agent sends a follow-up message to request structured formatting. This parameter
@@ -431,6 +439,12 @@ class Agent(AgentBase):
             for plugin in plugins_to_register:
                 self._plugin_registry.add_and_init(plugin)
 
+        # Resolve and register the memory manager (a Plugin); keep a reference so the
+        # synchronous entry point can flush pending extraction writes.
+        self.memory_manager = self._resolve_memory_manager(memory_manager)
+        if self.memory_manager is not None:
+            self._plugin_registry.add_and_init(self.memory_manager)
+
         self.hooks.invoke_callbacks(AgentInitializedEvent(agent=self))
 
     @staticmethod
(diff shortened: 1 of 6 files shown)


strands-py/src/strands/memory/memory_manager.py
```diff
@@ -100,11 +95,10 @@ def __init__(
             add_tool_config: Add tool configuration. ``False`` (default) disables
                 the add tool; ``True`` lets it write to all writable stores; a
                 :class:`MemoryAddToolConfig` restricts/customizes it.
-            flush_on_invocation_end: When True, await pending extraction writes at
-                the end of each agent invocation. Enable when driving the agent
-                through the synchronous ``Agent(...)`` entry point, whose
-                per-invocation event loop would otherwise cancel in-flight saves.
-                Defaults to False (fire-and-forget).
+            flush_on_invocation_end: When True (default), the synchronous
```
opieter-aws: Do we want this true by default? Doesn't that mean we block the agent loop and not use fire-and-forget mechanism?

JackYPCOnline: This is sync path, had a discussion with @pgrayy , user won't need to worry about it when using sync function

strands-py/src/strands/agent/agent.py
```diff
@@ -492,6 +506,30 @@ def _resolve_context_manager(
...
+    ) -> "MemoryManager | None":
+        """Resolve the ``memory_manager`` argument into a MemoryManager instance or None.
+
+        A :class:`~strands.memory.MemoryManagerConfig` is wrapped into a
+        :class:`~strands.memory.MemoryManager`; an instance passes through.
+        """
+        if memory_manager is None:
+            return None
+
+        from ..memory import MemoryManager, MemoryManagerConfig
```
pgrayy: Nit: We shouldn't need to inline this. We should be able to import from memory directly unless I am mistaken. I think Agent is only imported for typing in memory and so I wouldn't expect a circular dependency issue.

strands-py/src/strands/memory/memory_manager.py
```diff
@@ -88,7 +83,7 @@ def __init__(
         stores: list[MemoryStore],
         search_tool_config: MemoryToolConfig | bool = True,
         add_tool_config: MemoryAddToolConfig | bool = False,
-        flush_on_invocation_end: bool = False,
+        flush_on_invocation_end: bool = True,
```
pgrayy: Do we need this parameter anymore? It's only needed for sync right, which is automatically handled now?

strands-py/src/strands/agent/agent.py
```diff
@@ -431,6 +438,17 @@ def __init__(
...
                 self._plugin_registry.add_and_init(plugin)
 
+        # Resolve and register the memory manager (a Plugin); keep a reference so the
+        # synchronous entry point can flush pending extraction writes.
+        self.memory_manager = self._resolve_memory_manager(memory_manager)
+        if self.memory_manager is not None:
+            if self.memory_manager.name in self._plugin_registry._plugins:
+                raise ValueError(
+                    "MemoryManager is already registered; pass it through either the memory_manager "
+                    "parameter or plugins, not both"
```
opieter-aws: I don't think we want to encourage customers to register MM as a plugin; it's just implementing a plugin because that's how these components are set up in the Strands SDK

strands-py/src/strands/memory/memory_manager.py
JackYPCOnline: I intentionally keep it since user can opted out if they want? Maybe in some narrow combined use cases, it brings difference.

strands-py/src/strands/agent/agent.py
JackYPCOnline: Addressed.

strands-py/src/strands/memory/memory_manager.py
JackYPCOnline: Per discussion offline, i will remove this parameter in next iteration.

strands-py/src/strands/agent/agent.py
```diff
@@ -644,6 +679,19 @@ def __call__(
...
+    async def _invoke_async_and_flush(self, prompt: AgentInput = None, **kwargs: Any) -> AgentResult:
+        """Run ``invoke_async`` then flush the memory manager within this loop.
+
+        The synchronous entry point runs each invocation in its own event loop, which would
+        cancel background extraction saves on close. Flushing here persists them. The async
+        path does not flush, leaving extraction on its trigger cadence.
+        """
+        try:
+            return await self.invoke_async(prompt, **kwargs)
+        finally:
```
opieter-aws, line 692: Should we guard this so a flush failure can’t mask the primary exception, when this gets triggered due to model error or `agent.cancel()`?

strands-py/src/strands/agent/agent.py
```diff
@@ -492,6 +506,27 @@ def _resolve_context_manager(
 
         return resolved_conversation_manager, resolved_plugins
 
+    def _resolve_memory_manager(
```
opieter-aws, line 510: nit: should we mark this as `@staticmethod`?

JackYPCOnline: to match context_manager, yes. Will update.

strands-py/src/strands/agent/agent.py
JackYPCOnline: Trace the path, this always calls `await self._coordinator.flush()`. This function never raises, so no guard is needed.

opieter-aws (approved):

PR #2797, opened 2026-06-15
feat: add memory injection

opieter-aws:
## Description

Python port of #2631 

### Motivation

The TypeScript SDK ships context injection: a way to fold just-in-time text into the model's input *for a single call* without persisting it to the durable conversation. The `MemoryManager` builds on it to surface relevant memories to the model automatically, so the agent doesn't have to spend a tool call searching. Python had neither — `MemoryManager` (#2740) landed with extraction and search tools but no injection — so the Python `MemoryManager` could only recall memory reactively via the `search_memory` tool, never proactively.

This ports the injection primitive to Python and wires it into `MemoryManager`, bringing the two SDKs to parity. The design follows the TS source: a small, reusable `strands.injection` engine delivered as an `InvokeModelStage.Input` middleware, a thin `ContextInjector` vended plugin for arbitrary content, and a `MemoryManager` consumer for memory-specific retrieval. Behavior, fail-open semantics, the default `<memory>` format, the prepend-vs-append fold rule, and trigger policies all mirror the TS implementation.

Ephemerality is the core contract: injected text reaches the model for one call but never enters `agent.messages` or the session. This falls out of the middleware seam — the engine rewrites the per-call `InvokeModelContext.messages` (already a defensive copy of agent state), so durable history is untouched.

Memory injection is **on by default** in both SDKs: if you configure a `MemoryManager` with stores, the point is to use them, so proactive recall is the obvious-path default rather than a flag you have to discover. Because this also changes the already-shipped TS `MemoryManager`, it is a behavior change there — see Breaking Changes. Opt out with `injection: false` / `injection=False`.

### Public API Changes

**1. New `strands.injection` package** — shared configuration types for context injection. Delivery primitives are intentionally internal; the public surface is config only:

```python
from strands.injection import InjectionConfig, InjectionContext, InjectionTrigger
```

**2. New `ContextInjector` vended plugin** — folds arbitrary just-in-time text (current time, retrieved docs, request metadata, …) into the model input:

```python
from strands import Agent
from strands.vended_plugins.context_injector import ContextInjector

agent = Agent(
    plugins=[
        ContextInjector(lambda context: f"<context>{derive(context.messages)}</context>")
    ]
)
```

`render_content` is the only required argument (sync or async; returns the text, or `None`/`""` to skip). Optional keyword-only `name` (default `"strands:context-injector"`) and `trigger` (`"userTurn"` default, `"everyTurn"`, or a predicate over the `InjectionContext`). A callback that raises fails open — injection is skipped and the model call proceeds.

**3. `MemoryManager` gains an `injection` option** — on by default:

```python
from strands import Agent
from strands.memory import MemoryManager, MemoryInjectionConfig

# Default: inject up to 5 entries on a user turn, rendered as a <memory> block
agent = Agent(plugins=[MemoryManager(stores=[store])])

# Opt out
agent = Agent(plugins=[MemoryManager(stores=[store], injection=False)])

# Or customize retrieval, timing, and formatting
agent = Agent(
    plugins=[
        MemoryManager(
            stores=[store],
            injection=MemoryInjectionConfig(
                trigger="everyTurn",
                max_entries=3,
                query=lambda context: derive_query(context.messages),
                format=lambda context: render(context.entries),
            ),
        )
    ]
)
```

`MemoryInjectionConfig` extends `InjectionConfig`. When enabled, `MemoryManager` derives a query from the conversation (the latest user text on a user turn, else the most recent assistant text), searches its stores, and folds the top entries into the model input. `injection` defaults to `True` (`False` disables it). `strands.memory` also re-exports `InjectionConfig`/`InjectionContext`/`InjectionTrigger` so injection can be configured from a single import.

The same default flip is applied to the TypeScript `MemoryManager` (`injection?: boolean | MemoryInjectionConfig`, now defaulting to enabled) to keep the two SDKs aligned.

One deliberate cross-SDK divergence worth a reviewer's eye: the callback config fields (`trigger`, `render_content`, `query`, `format`) are typed as `Callable | Protocol`-with-`**kwargs` unions rather than the bare `Callable` the TS side uses. This follows the STYLE_GUIDE's "avoid `Callable` for extensible interfaces" rule (the `Protocol` arm lets the calling convention grow keyword arguments later) while the `Callable` arm keeps the plain-lambda happy path ergonomic under mypy strict — a pure `Protocol` rejects bare lambdas. This mirrors the existing `EdgeCondition` pattern in `multiagent/graph.py`.

## Breaking Changes

For the Python `MemoryManager`, memory injection is new in this PR, so default-on is a fresh default, not a change. For the **TypeScript `MemoryManager`** (shipped in #2631), this flips `injection` from opt-in (`false`) to opt-out (enabled). Existing TS agents that configure a `MemoryManager` will now, on each model call, derive a query from the conversation, search their stores, and prepend a `<memory>` block to the model input — adding tokens and a search round-trip per call, and potentially surfacing irrelevant recall. This is a behavior change under the "pay for play" tenet, called out here for explicit reviewer sign-off; it is not an API-signature break (the field and its type are unchanged).

### Migration

```typescript
// Restore the previous (no-injection) behavior:
new MemoryManager({ stores, injection: false })
```

```python
# Python equivalent:
MemoryManager(stores=stores, injection=False)
```

## Documentation PR

Follow-up: a docs page for `ContextInjector` and `MemoryManager` injection is not included here.

## Type of Change

New feature

## Testing

Unit tests mirror the TS describe blocks (fold/trigger/query/format/fail-open), plus real-Agent integration tests on both the `ContextInjector` and `MemoryManager` paths asserting the injected text reaches the model but never `agent.messages`.

## Checklist

----

### DIFF
--- a/strands-py/src/strands/vended_plugins/context_injector/plugin.py
+++ b/strands-py/src/strands/vended_plugins/context_injector/plugin.py
@@ -1,0 +1,100 @@
+"""ContextInjector plugin for injecting just-in-time context into the model input.
+
+This module provides the ContextInjector plugin, which folds just-in-time text into the model
+input before each call without touching durable history.
+
+Example:
+    ```python
+    import datetime
+
+    from strands import Agent
+    from strands.vended_plugins.context_injector import ContextInjector
+
+    agent = Agent(
+        plugins=[
+            ContextInjector(lambda context: f"<now>{datetime.datetime.now().isoformat()}</now>")
+        ]
+    )
+    ```
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from ..._middleware.stages import InvokeModelStage
+from ...injection.message_injection import RenderContent, create_injection_middleware
+from ...injection.types import InjectionTriggerPredicate
+from ...plugins import Plugin
+
+if TYPE_CHECKING:
+    from ...agent.agent import Agent
+
+_DEFAULT_NAME = "strands:context-injector"
+"""Default plugin name; override to tell multiple injectors apart."""
+
+
+class ContextInjector(Plugin):
+    """Plugin that injects just-in-time context into the model input before each call.
+
+    Before each model call, the plugin asks ``render_content`` for text and makes it available
+    to the model for that call, gated by ``trigger``. The injected text is ephemeral: it
+    augments the model input for that one call and never persists into the durable conversation
+    or session.
+
+    Multiple injectors may be registered; each contributes its text independently, in
+    plugin-registration order.
+
+    Args:
+        render_content: Renders the text to inject for this call, or ``None``/``""`` to skip.
+            Sync or async. The text reaches the model verbatim, so it is a prompt-injection
+            surface: escape any attacker-influenced fields yourself. A callback that raises
+            fails open (injection is skipped, the model call proceeds).
+        name: Plugin name, for logging and duplicate detection. Defaults to
+            ``"strands:context-injector"``. Set a distinct name when registering more than one
+            injector so they can be told apart.
+        trigger: When to inject. An ``InjectionTrigger`` name selects a built-in policy
+            (``"userTurn"`` — default — or ``"everyTurn"``); a predicate over the
+            ``InjectionContext`` is the escape hatch. A predicate that raises fails open
+            (injection is skipped). Defaults to ``"userTurn"``.
+
+    Example:
+        ```python
+        from strands import Agent
+        from strands.vended_plugins.context_injector import ContextInjector
+
+        agent = Agent(
+            plugins=[
+                ContextInjector(lambda context: f"<context>{derive(context.messages)}</context>")
+            ]
+        )
+        ```
+    """
+
+    name = _DEFAULT_NAME
+
+    def __init__(
+        self,
+        render_content: RenderContent,
+        *,
+        name: str | None = None,
+        trigger: InjectionTriggerPredicate | None = None,
+    ) -> None:
+        """Initialize the ContextInjector plugin.
+
+        Args:
+            render_content: Renders the text to inject for this call. Sync or async.
+            name: Plugin name. Defaults to ``"strands:context-injector"``.
+            trigger: When to inject. Defaults to ``"userTurn"``.
+        """
+        self.name = name or _DEFAULT_NAME
+        self._render_content = render_content
+        self._trigger = trigger
+        super().__init__()
+
+    def init_agent(self, agent: Agent) -> None:
+        """Register the injection middleware on the agent's ``InvokeModelStage`` input phase."""
+        agent._middleware_registry.add_middleware(
+            InvokeModelStage.Input,
+            create_injection_middleware(self._render_content, trigger=self._trigger),
+        )
--- a/strands-py/src/strands/vended_plugins/context_injector/plugin.py
+++ b/strands-py/src/strands/vended_plugins/context_injector/plugin.py
@@ -23,7 +23,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from ..._middleware.stages import InvokeModelStage
-from ...injection.message_injection import RenderContent, create_injection_middleware
+from ...injection.message_injection import RenderContent, _create_injection_middleware
 from ...injection.types import InjectionTriggerPredicate
 from ...plugins import Plugin
 
@@ -96,5 +96,5 @@ class ContextInjector(Plugin):
         """Register the injection middleware on the agent's ``InvokeModelStage`` input phase."""
         agent._middleware_registry.add_middleware(
             InvokeModelStage.Input,
-            create_injection_middleware(self._render_content, trigger=self._trigger),
+            _create_injection_middleware(self._render_content, trigger=self._trigger),
         )
--- a/strands-py/src/strands/injection/__init__.py
+++ b/strands-py/src/strands/injection/__init__.py
@@ -1,0 +1,15 @@
+"""Context injection for Strands Agents.
+
+This package provides the configuration types for context injection — folding just-in-time text
+into the model input before a call without touching durable history. The delivery primitives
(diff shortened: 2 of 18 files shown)


agent-of-mkmeral (commented):
## Review — Python port of memory injection (#2631)

**Assessment: Approve (no blockers).** I checked this out, read every new/changed file, ran the suite, type-checked, and ran my own adversarial edge probes. This is a faithful, well-documented port and the earlier blocking concerns (description-vs-code default mismatch; TS scope) have been resolved — the description now correctly states default-on and calls out the TS behavior change under Breaking Changes, and both prior bot threads are resolved.

### Verification (local, head `7533bb6`)
- **`135 passed`** across `tests/strands/injection`, `tests/strands/vended_plugins/context_injector`, `test_injection_integration.py`, `test_memory_manager.py`
- **mypy clean** on the 8 new/changed source files
- Integration test directly proves the **ephemerality contract**: injected `<memory>` reaches `model.stream(...)` for one call but never appears in `agent.messages`.

### What's good
- **Ephemerality via the middleware seam is correct.** `InvokeModelContext.messages` is already a defensive copy of agent state; the handler does `replace(context, messages=fold_into_last_user_message(...))` and `fold_*` returns a brand-new list without mutating inputs (test `test_returns_new_list_and_does_not_mutate_input` + `test_does_not_mutate_original_context_messages` confirm). Durable history is structurally untouched — not just by convention.
- **Fold rule is the subtle part and it's right + tested.** Prepend on a plain user ask (keeps the user's ask in the recency slot); append on a tool-result turn (keeps the tool result the first block, which providers require). I verified the mixed `[toolResult, text]` case appends correctly and toolResult stays first.
- **Fail-open is consistent across all four callback seams** — `trigger` predicate, `render_content`, memory `query`, and memory `format` each catch, log `reason=<…>`, and skip injection so the model call proceeds. Matches TS exactly.
- **XML escaping** protects the default `<memory>` block from both structural breakage and stored-prompt-injection. `&` is escaped first; attr-escaping adds `"`/`'`. My probe on `a & b < c > </entry>` and `he said "hi"` escaped correctly.
- **TS parity is line-for-line** on the engine, resolver, fold logic, `_provide_memory_context`, `_resolve_injection_query` adaptive default (user text on user turn, else most-recent assistant text), and `_default_injection_format`.

### Python-specific touches worth calling out (all good)
- The `Callable | Protocol-with-**kwargs` unions for the callback fields (vs. TS's bare function type) is the right call under the STYLE_GUIDE — keeps bare lambdas ergonomic under mypy strict while leaving room for the calling convention to grow kwargs. Mirrors the `EdgeCondition` precedent.
- `strands.memory` re-exporting `InjectionConfig/InjectionContext/InjectionTrigger` is a nice single-import ergonomic.

### Non-blocking observations
1. **`max_entries` double-cap is correct but worth a one-line note.** `search(...)` caps each store at `max_results`, then `[:max_results]` caps the concatenation. With N stores you fetch up to N×max before slicing, and since `search` concatenates in store-registration order with no cross-store ranking, the slice can systematically favor earlier-registered stores. This is already documented in the `MemoryInjectionConfig.max_entries` docstring ("can favor entries from earlier-registered stores") — good. No change needed; just flagging it's an intentional, documented tradeoff.
2. **`is_user_turn` / fold assume well-formed messages** (every message has a `content` list). A message dict missing `content` raises `KeyError` rather than failing open. In practice the agent loop only ever produces well-formed messages so this is not reachable from supported paths, and matches TS (which would also misbehave on malformed data). Not worth defensive code, but noting it since the rest of the module is so carefully fail-open.
3. **Default-on is a real behavior change for existing TS `MemoryManager` users** (extra search + injected block + tokens per user turn). It's now correctly documented and is a deliberate "obvious-path default" decision — but it does warrant the API reviewer's explicit sign-off, which the description is well-prepared for.

Nice work — the hard parts (fold placement, ephemerality, fail-open uniformity, escaping) are all handled and covered by tests that mirror the TS describe blocks.

strands-py/src/strands/vended_plugins/context_injector/plugin.py
```diff
@@ -0,0 +1,100 @@
...
+
+from ..._middleware.stages import InvokeModelStage
+from ...injection.message_injection import RenderContent, _create_injection_middleware
+from ...injection.types import InjectionTriggerPredicate
+from ...plugins import Plugin
+
+if TYPE_CHECKING:
+    from ...agent.agent import Agent
+
+_DEFAULT_NAME = "strands:context-injector"
```
JackYPCOnline, line 33: Nit: Should we keep consistent with `context_offloader`

mkmeral: I think we have both standards in the repo 😅

mkmeral (approved):

agent-of-mkmeral:
Non-blocking nit (don't hold the merge for this) — nice follow-up underscoring the internal functions in `efc8b40`. The one remaining loose end is the **module files** themselves:

- `strands/injection/message_injection.py`
- `strands/injection/xml.py`

The functions inside are now `_`-prefixed (good — `from strands.injection.message_injection import fold_into_last_user_message` correctly fails), but the *module paths* still read as public. `from strands.injection import message_injection` works and exposes the `_`-prefixed internals with no namespace-level "private" signal.

The repo's own convention is to underscore the **module/package**, not just its contents — e.g. `_middleware/`, `event_loop/_retry.py`, `models/_defaults.py`, `strands/_async.py`. Injection is currently the one spot that diverges (functions underscored, files not). Suggestion:

- `message_injection.py` → `_message_injection.py`
- `xml.py` → `_xml.py`
- update the two relative imports in `memory/memory_manager.py` and `vended_plugins/context_injector/plugin.py`

Why now rather than later: once released, the module path could become load-bearing for someone, so renaming gets more expensive post-merge. But this is purely a convention/consistency nit — the exported `__all__` surface is correct, named-symbol reachability is now fixed, and nothing here is a correctness issue. Totally fine to take it or merge as-is. ✅

PR #2800, opened 2026-06-15
fix(strands-py): allow sync or async InterventionHandler lifecycle overrides

agent-of-mkmeral:
## Description

Split out of #2750 (the HITL port) per @lizradway's review request: *"can you raise this in a separate PR?"*

`InterventionHandler` lifecycle methods could only be reliably implemented as `async def` before this change. The registry decided whether to await using `inspect.iscoroutinefunction(method_fn)` — `True` only for `async def`. A **sync** override that returned a coroutine/future would be passed through **un-awaited** (`RuntimeWarning: coroutine was never awaited` + a type error downstream). This is the exact mismatch the bot review and @lizradway flagged on #2750 when the base-class returns were widened to `... | Awaitable[...]` without a matching runtime change.

## Changes

1. **`_MaybeAwaitable[T]` type alias.** `_MaybeAwaitable: TypeAlias = _T | Awaitable[_T]`, mirroring the TypeScript [`Awaitable<T>`](https://github.com/strands-agents/harness-sdk/blob/main/strands-ts/src/interventions/handler.ts) alias (@lizradway's suggestion — avoids repeating the union twice per method). All five `InterventionHandler` lifecycle methods are annotated with it, so overrides may be **sync or async**. It is an **internal** annotation alias (underscore-prefixed, **not exported**): it only widens the lifecycle return signatures so an override can be a plain `def` or an `async def`. It is an implementation detail of supporting both styles, not part of the public contract.
2. **Runtime matches the type.** Dispatch now awaits based on the *returned value* — `inspect.isawaitable(result)` — instead of `inspect.iscoroutinefunction(method_fn)`. A sync `def` that hands back an awaitable is now awaited correctly; an `async def` still works exactly as before.
3. **Tests** for all three paths: a plain sync `def` override, an `async def` override, and the sync-`def`-returning-a-coroutine case that `iscoroutinefunction` missed (locks in the contract the new annotation advertises).

## Why separate from #2750

This touches the shared intervention **primitive** (all handlers), not just the HITL vended handler. Keeping it isolated makes the runtime+type change reviewable on its own and keeps #2750 scoped to the new `HumanInTheLoop` handler. With this merged, #2750's `HumanInTheLoop.before_tool_call` can drop its `# type: ignore[override]` (it's restored there for now).

## Type of Change

Bug fix / API refinement (non-breaking — async handlers keep working; sync handlers now work too).

## Testing

- `pytest tests/strands/interventions tests/strands/test_interrupt.py tests/strands/hooks` → **128 passed**
- Broader sweep `tests/strands/agent` (a2a optional extra skipped) → **671 passed**
- `ruff format --check` / `ruff check` → clean
- `mypy src/strands/interventions/handler.py src/strands/interventions/registry.py` → clean (no `type: ignore`)

- [x] I ran lint/format/type checks and the test suite for changed files

## Checklist
- [x] My change is focused and reasonably small; unrelated work is split out
- [x] I have added tests that prove the fix works

----

### DIFF
--- a/strands-py/src/strands/interventions/__init__.py
+++ b/strands-py/src/strands/interventions/__init__.py
@@ -28,4 +28,5 @@ from .actions import InterventionAction as InterventionAction
 from .actions import Proceed as Proceed
 from .actions import Transform as Transform
 from .handler import InterventionHandler as InterventionHandler
+from .handler import MaybeAwaitable as MaybeAwaitable
 from .handler import OnError as OnError
--- a/strands-py/src/strands/interventions/__init__.py
+++ b/strands-py/src/strands/interventions/__init__.py
@@ -28,5 +28,4 @@ from .actions import InterventionAction as InterventionAction
 from .actions import Proceed as Proceed
 from .actions import Transform as Transform
 from .handler import InterventionHandler as InterventionHandler
-from .handler import MaybeAwaitable as MaybeAwaitable
 from .handler import OnError as OnError
--- a/strands-py/src/strands/interventions/registry.py
+++ b/strands-py/src/strands/interventions/registry.py
@@ -206,7 +206,10 @@ class InterventionRegistry:
             try:
                 method_fn = getattr(handler, method)
                 result = method_fn(event)
-                action = await result if inspect.iscoroutinefunction(method_fn) else result
+                # Await based on the *returned value*, not iscoroutinefunction(method_fn):
+                # lifecycle overrides may be sync or async (the base class annotates them
+                # as MaybeAwaitable), and a sync def can still hand back a coroutine/future.
+                action = await result if inspect.isawaitable(result) else result
             except Exception as error:
                 action = self._handle_error(handler, method, error)
                 if action is None:
--- a/strands-py/src/strands/interventions/registry.py
+++ b/strands-py/src/strands/interventions/registry.py
@@ -206,9 +206,7 @@ class InterventionRegistry:
             try:
                 method_fn = getattr(handler, method)
                 result = method_fn(event)
-                # Await based on the *returned value*, not iscoroutinefunction(method_fn):
-                # lifecycle overrides may be sync or async (the base class annotates them
-                # as MaybeAwaitable), and a sync def can still hand back a coroutine/future.
+                # Overrides may be sync or async, so branch on the returned value.
                 action = await result if inspect.isawaitable(result) else result
             except Exception as error:
                 action = self._handle_error(handler, method, error)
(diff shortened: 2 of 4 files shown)


strands-py/src/strands/interventions/__init__.py
```diff
@@ -28,4 +28,5 @@ def before_tool_call(self, event):
 from .actions import Proceed as Proceed
 from .actions import Transform as Transform
 from .handler import InterventionHandler as InterventionHandler
+from .handler import MaybeAwaitable as MaybeAwaitable
```
lizradway: do not export an internal implementation detail

strands-py/src/strands/interventions/registry.py
```diff
@@ -206,7 +206,10 @@ async def _dispatch(
             try:
                 method_fn = getattr(handler, method)
                 result = method_fn(event)
-                action = await result if inspect.iscoroutinefunction(method_fn) else result
+                # Await based on the *returned value*, not iscoroutinefunction(method_fn):
+                # lifecycle overrides may be sync or async (the base class annotates them
```
lizradway: comment isn't super helpful this is self explainatory

lizradway (changes requested):

strands-py/src/strands/interventions/registry.py
agent-of-mkmeral: Trimmed in `af4be07f` to a single self-contained line: `# Overrides may be sync or async, so branch on the returned value.` Dropped the longer explanation of why it's `isawaitable(result)` over `iscoroutinefunction(method_fn)`.

agent-of-mkmeral:
@lizradway addressed both review notes in `af4be07f`:

1. **"do not export an internal implementation detail"** — removed `MaybeAwaitable` from `interventions/__init__.py`. It stays in `handler.py` purely as the annotation alias on the lifecycle method signatures; a customer writing a sync or async override just writes `def`/`async def` and never imports it, so it's not public surface.
2. **"comment isn't super helpful, this is self explanatory"** — trimmed the registry dispatch comment to one line: `# Overrides may be sync or async, so branch on the returned value.`

Verification: interventions + vended + interrupt + hooks → 128 passed; ruff + mypy clean. Still MERGEABLE. This is the PR that lets #2750's `HumanInTheLoop.before_tool_call` drop its `# type: ignore[override]` once it lands.

lizradway (dismissed):

mkmeral (approved):

lizradway (approved):

PR #2801, opened 2026-06-15
feat(strands-py): support sync & async intervention lifecycle overrides

agent-of-mkmeral:
## What

Makes `InterventionHandler` lifecycle methods work as **either sync `def` or `async def`** — at both the type level and the runtime — and adds a `MaybeAwaitable[T]` alias so the contract is expressed once.

Split out of #2750 (HITL port) per @lizradway's request — this touches the shared intervention primitive (all handlers), so it belongs in its own PR rather than riding in the HITL feature.

## Why

The registry dispatched on `inspect.iscoroutinefunction(method_fn)` — i.e. it only awaited methods *declared* `async def`. If the type signature advertises `... | Awaitable[...]` (to let overrides be async), then a perfectly legal handler:

```python
class MyHandler(InterventionHandler):
    name = "x"
    def before_tool_call(self, event, **kwargs):
        return self._decide(event)   # sync def, returns a coroutine
```

would have its returned coroutine **passed through un-awaited** — `RuntimeWarning: coroutine was never awaited` plus a non-action value flowing into `apply()`. The type said one thing; the runtime did another.

Fix: dispatch on whether the **returned value** is awaitable (`inspect.isawaitable(result)`), which correctly handles all three shapes:
- sync `def` returning an action → used directly
- `async def` → awaited
- sync `def` returning a coroutine/awaitable → awaited

(Picked `isawaitable(result)` over `iscoroutinefunction` per the review thread on #2750 — @lizradway: *"+1 to switch the registry to inspect.isawaitable(result) on the returned value"*.)

## Changes

- **`interventions/handler.py`**: add `MaybeAwaitable: TypeAlias = T | Awaitable[T]` (mirrors the TS `Awaitable<T>` in [`strands-ts/src/interventions/handler.ts`](https://github.com/strands-agents/harness-sdk/blob/main/strands-ts/src/interventions/handler.ts), as @lizradway suggested), and annotate all five lifecycle methods with it — replacing the repeated `... | Awaitable[...]` unions.
- **`interventions/registry.py`**: dispatch via `inspect.isawaitable(result)` instead of `inspect.iscoroutinefunction(method_fn)`, with a comment explaining why we inspect the value, not the function.
- **`interventions/__init__.py`**: export `MaybeAwaitable` (it appears in the public lifecycle signatures; `OnError` is already exported here for the same reason).
- **Tests**: new `TestSyncAndAsyncOverrides` covering sync `def`, `async def`, **sync `def` returning a coroutine** (the previously-broken case), and a sync `Proceed` not short-circuiting later handlers.

## Verification

```
pytest tests/strands/interventions tests/strands/test_interrupt.py tests/strands/hooks tests/strands/agent/test_agent_hooks.py
→ 157 passed
```
The interventions suite also passes under `-W error::RuntimeWarning`, which would fail loudly on any un-awaited coroutine leak. `ruff format --check`, `ruff check`, and `mypy src/strands/interventions/` are all clean.

## Follow-up

Once this lands, #2750 (HITL) drops its `# type: ignore[override]` on the now-natively-async `before_tool_call` and rebases onto the widened contract.

cc @mkmeral @lizradway

### DIFF
--- a/strands-py/src/strands/interventions/__init__.py
+++ b/strands-py/src/strands/interventions/__init__.py
@@ -28,4 +28,5 @@ from .actions import InterventionAction as InterventionAction
 from .actions import Proceed as Proceed
 from .actions import Transform as Transform
 from .handler import InterventionHandler as InterventionHandler
+from .handler import MaybeAwaitable as MaybeAwaitable
 from .handler import OnError as OnError
--- a/strands-py/src/strands/interventions/handler.py
+++ b/strands-py/src/strands/interventions/handler.py
@@ -6,7 +6,8 @@ registers hook callbacks for those.
 """
 
 from abc import ABC, abstractmethod
-from typing import Any, Literal
+from collections.abc import Awaitable
+from typing import Any, Literal, TypeAlias, TypeVar
 
 from ..hooks.events import (
     AfterModelCallEvent,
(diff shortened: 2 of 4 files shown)


agent-of-mkmeral:
Closing as a duplicate of #2800 — two concurrent agent runs split the same primitive change out of #2750 within ~10s of each other. #2800 was created first and is functionally identical (adds `MaybeAwaitable[T]`, switches registry dispatch to `inspect.isawaitable(result)`, exports the alias, and tests the sync / async / sync-returns-coroutine paths). Consolidating on #2800.

PR #2806, opened 2026-06-15
refactor(sandbox): inline tool prefixing into getTools implementations

gautamsirdeshmukh:
## Description

- Move tool prefixing into `getTools()` implementations

## Type of Change

Refactor

## Testing

## Checklist

----

### DIFF
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -770,23 +770,16 @@ export class Agent implements LocalAgent, InvokableAgent {
       })
     )
 
-    // Register tools vended by an explicitly-configured sandbox, applying the sandbox's
-    // toolPrefix to names (like MCP's prefix for server-vended tools).
+    // Register tools vended by the sandbox. The host default vends nothing. A tool
+    // is skipped if the user already registered one with that name.
     if (this._sandbox) {
-      const prefix = this._sandbox.toolPrefix
-      for (const tool of this._sandbox.getTools()) {
-        const prefixed = prefix
-          ? Object.create(tool, {
-              name: { value: `${prefix}_${tool.name}` },
-              toolSpec: { value: { ...tool.toolSpec, name: `${prefix}_${tool.name}` } },
-            })
-          : tool
-        if (this._toolRegistry.get(prefixed.name)) {
+      for (const sandboxTool of this._sandbox.getTools()) {
+        if (this._toolRegistry.get(sandboxTool.name)) {
           logger.debug(
-            `tool_name=<${prefixed.name}> | sandbox-vended tool skipped, user has already registered tool with this name`
+            `tool_name=<${sandboxTool.name}> | sandbox-vended tool skipped, user already registered a tool with this name`
           )
         } else {
-          this._toolRegistry.add(prefixed)
+          this._toolRegistry.add(sandboxTool)
         }
       }
     }
--- a/strands-ts/src/sandbox/base.ts
+++ b/strands-ts/src/sandbox/base.ts
@@ -112,12 +112,6 @@ export abstract class Sandbox {
   abstract listFiles(path: string): Promise<FileInfo[]>
 
   /**
-   * Prefix applied to tool names when registered on an agent (e.g. `'sandbox'` produces
-   * `sandbox_bash`). Set to `undefined` to disable prefixing. Defaults to `'sandbox'`.
-   */
-  toolPrefix: string | undefined = 'sandbox'
-
-  /**
    * Tools this sandbox vends to an agent, registered during `Agent.initialize()`.
    * A tool is skipped if the user already registered one with the same name.
    * Override to provide them.
--- a/strands-ts/src/sandbox/__tests__/docker.test.node.ts
+++ b/strands-ts/src/sandbox/__tests__/docker.test.node.ts
@@ -97,12 +97,12 @@ describe('DockerSandbox', () => {
   describe('getTools', () => {
     it('vends the sandbox-routed fileEditor and bash tools', () => {
       const tools = new DockerSandbox({ container: 'my-container' }).getTools()
-      expect(tools.map((t) => t.name)).toStrictEqual(['fileEditor', 'bash'])
+      expect(tools.map((t) => t.name)).toStrictEqual(['sandbox_file_editor', 'sandbox_bash'])
     })
 
     it('vends bash with the sandbox description', () => {
       const tools = new DockerSandbox({ container: 'my-container' }).getTools()
-      const bashTool = tools.find((t) => t.name === 'bash')!
+      const bashTool = tools.find((t) => t.name === 'sandbox_bash')!
       expect(bashTool.description).toContain(SANDBOX_BASH_DESCRIPTION)
       expect(bashTool.description).toContain('my-container')
     })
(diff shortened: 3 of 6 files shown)


strands-ts/src/agent/agent.ts
```diff
@@ -299,16 +299,10 @@ export type AgentConfig = {
...
-   *   This default is the slot reserved for richer behavior later.
-   * - `false`: explicitly opt out of a managed sandbox and run on the host.
-   *
-   * Keep `false` distinct from omitting so the opt-out stays stable even if the
-   * default changes.
+   * When omitted, the agent uses the environment's default sandbox (host
+   * execution in Node; unavailable in browsers).
    */
-  sandbox?: Sandbox | false
+  sandbox?: Sandbox
```
agent-of-mkmeral: **🔴 Out-of-scope breaking change to a released public API.**

This narrows the exported `AgentConfig.sandbox` from `Sandbox | false` to `Sandbox`, and the matching `_sandbox` field + the `treats sandbox: false the same as unconfigured` test were removed.

Key point: `sandbox: false` shipped in **`typescript/v1.5.0`** (introduced by #2563, confirmed via `git merge-base --is-ancestor efa837af typescript/v1.5.0`). Removing it is a source-level breaking change — any existing `new Agent({ model, sandbox: false })` now fails to compile, even though `false` is a runtime no-op today (`this._sandbox` falsy → default; tool-registration `if (this._sandbox)` skipped).

**Suggestion:** keep this PR focused on the singleton→per-agent fix and restore the field:
```ts
  sandbox?: Sandbox | false
```
(with `_sandbox: Sandbox | false | undefined` and the `sandbox: false` test). If the intent is to genuinely retire the opt-out, do it in a separate PR with `api/needs-review` + a migration note per CONTRIBUTING's breaking-change process.

agent-of-mkmeral (commented):
**Assessment: Request Changes** 🔴

Reviewed the works on `9870aa7` — independently cloned, type-checked (`src` project clean), and ran the full `unit-node` suite.

## ✅ The core fix is clean and correct

The singleton→per-agent factory conversion is exactly right and fixes a real shared-state bug:
- `defaultSandboxFactory` stores `() => Sandbox` instead of a shared instance, so each unconfigured agent builds its own host default.
- `Agent.sandbox` memoizes via `this._defaultSandbox ??= defaultSandboxFactory.get()()` — stable instance per agent across repeated reads (important: sandbox-aware tools read `agent.sandbox` multiple times and rely on a shared cwd/filesystem).
- The earlier mock-fixture feedback (memoize `createMockAgent`'s getter to mirror the real semantics) was fully addressed — verified it now matches.
- New tests cover the right behaviors: fresh-per-factory-call, per-agent isolation, and per-agent memoization.

**Verification results:**
- `tsc --project src/tsconfig.json` → ✅ clean (0 errors)
- Sandbox unit tests → ✅ **106/106 pass** (incl. the 6 changed default-sandbox tests)
- Full `unit-node` → 3493 pass. The only failures (4 in `telemetry/__tests__/config.test.ts`) and integ type errors (`steering.test.node.ts`) are **pre-existing on `main` and unrelated** to this PR (confirmed by re-running against `main`).

## 🔴 Blocker: the bundled `sandbox: false` removal is out of scope — and it's a breaking change to a *released* contract

This PR also drops `sandbox?: Sandbox | false` → `sandbox?: Sandbox` from the public `AgentConfig` (inline comment below). I want to add one fact the prior re-review didn't establish: I checked `git merge-base --is-ancestor efa837af typescript/v1.5.0` and **`#2563` (which introduced `sandbox: false`) is already shipped in `typescript/v1.5.0`**. So this isn't removing an unreleased placeholder — it's removing an option from an exported type that users may already be passing. Anyone with `new Agent({ model, sandbox: false })` will get a **compile error** after this lands.

Yes, `false` is behaviorally a no-op today (`this._sandbox` is falsy → falls through to the default, and the tool-registration `if (this._sandbox)` block is skipped). So removing it is *harmless at runtime* but **type-breaking at the source level**.

Per `AGENTS.md` ("One logical change per PR") and `CONTRIBUTING.md` (breaking changes → design proposal / API bar-raising), this belongs in its own PR. Two clean options:
1. **Revert the `sandbox: false` removal here** (keep this PR a pure singleton→per-agent fix — my preference; smallest, mergeable now), or
2. **Split it out** into a dedicated PR with `api/needs-review`, a changeset/migration note, and the `treats sandbox: false the same as unconfigured` test handled there.

The singleton→per-agent fix is ready to merge the moment the `sandbox: false` change is unbundled. Nice work on the memoization — it reads cleanly. 👍

strands-py/src/strands/sandbox/__init__.py
```diff
@@ -30,16 +30,20 @@ async def execute_streaming(self, command, *, timeout=None, cwd=None, env=None,
 
 from .base import Sandbox
 from .constants import LANGUAGE_PATTERN
+from .docker import DockerSandbox
 from .posix_shell import PosixShellSandbox
+from .ssh import SshSandbox
 from .types import ExecutionResult, FileInfo, OutputFile, StreamChunk, StreamType
 
 __all__ = [
+    "DockerSandbox",
```
mkmeral: no, this forces all importers of `strands.sandbox` to load docker

strands-ts/src/agent/agent.ts
```diff
@@ -448,7 +442,10 @@ export class Agent implements LocalAgent, InvokableAgent {
    */
   public readonly memoryManager?: MemoryManager | undefined
 
-  private readonly _sandbox: Sandbox | false | undefined
+  private readonly _sandbox: Sandbox | undefined
```
mkmeral: leave false, that's a TS pattern that we have rn

strands-py/src/strands/sandbox/__init__.py
gautamsirdeshmukh: Reverting

strands-ts/src/agent/agent.ts
gautamsirdeshmukh: Reverting

strands-ts/src/agent/agent.ts
```diff
@@ -450,14 +445,18 @@ export class Agent implements LocalAgent, InvokableAgent {
 
   private readonly _sandbox: Sandbox | false | undefined
 
+  /** Memoized per-agent host default, built on first access when no sandbox is configured. */
+  private _defaultSandbox?: Sandbox
```
mkmeral: why default sandbox? why not just sandbox?

strands-ts/src/sandbox/base.ts
```diff
@@ -6,27 +6,9 @@
...
  */
 
-import type { ExecutionResult, FileInfo, StreamChunk } from './types.js'
+import type { ExecuteOptions, ExecutionResult, FileInfo, StreamChunk } from './types.js'
 import type { Tool } from '../tools/tool.js'
 
-/**
- * Options for command and code execution.
- */
-export interface ExecuteOptions {
```
mkmeral, line 15: why are we moving? isn't this backwards incompatible?

gautamsirdeshmukh:
Discussed offline, removing everything except TS tool prefixing move

mkmeral (approved):

PR #2811, opened 2026-06-15
feat: add default for memory extraction trigger

opieter-aws:
## Description

Python's memory-extraction API diverged from the TypeScript SDK in three ways that made the
Python happy path noticeably heavier than its TS counterpart. In TS a store opts into automatic
extraction with `extraction: true` and gets sensible defaults; in Python the caller had to
construct an `ExtractionConfig`, pick a trigger explicitly (it was required), and choose an
extractor by hand. The shared documentation ("the simplest form turns it on with defaults",
"defaults to an interval of 5") only held for TS — there was no Python equivalent.

This brings Python in line. A store can now enable extraction with `extraction=True`, the
`trigger` is optional (defaulting to every 5 turns), and the extractor is selected automatically
from the store's write methods — matching TS's `boolean | ExtractionConfig` shape and its
`resolveExtractionConfig` behavior.

The change is opt-in and pay-for-play: existing `ExtractionConfig(trigger=...)` call sites keep
working unchanged. Defaulting is now centralized in a single internal `_resolve_extraction_config`
(mirroring TS's `resolve-extraction-config.ts`), so the `MemoryManager` and `ExtractionCoordinator`
no longer re-interpret raw config or re-apply defaults — the manager resolves once and passes
`_ExtractionBinding`s downstream.

### Public API Changes

A store's `extraction` field accepts a `bool` shorthand, and `ExtractionConfig.trigger` is now
optional:

```python
# Before: extraction required an explicit ExtractionConfig with an explicit trigger
class MyStore:
    extraction = ExtractionConfig(trigger=IntervalTrigger(turns=5))

# After: bool shorthand enables extraction with defaults (every 5 turns;
# extractor chosen from the store's write methods)
class MyStore:
    extraction = True

# After: trigger is optional — omit it to default to every 5 turns
class MyStore:
    extraction = ExtractionConfig()  # interval trigger of 5, default extractor + filter
```

Default extractor is now resolved from the store's write methods, matching TS:

- A store implementing only `add` → defaults to a `ModelExtractor` (client-side fact distillation,
  written via `add`).
- A store implementing `add_messages` → defaults to server-side extraction (raw messages handed to
  `add_messages`, no model call).

An explicit `extractor` or `filter` still overrides the default. An explicit empty trigger list is
still rejected at construction.

## Documentation PR

N/A

## Type of Change

New feature

## Testing

Added a `test_resolve_extraction_config` suite (ported from the TS resolver tests) covering the
bool shorthand, trigger defaulting/normalization, capability-based extractor selection, and filter
defaulting. Existing coordinator and manager tests were updated to drive the resolved-config path;
the previously-raising "add-only store without an extractor" case now asserts the new
`ModelExtractor` default behavior.

## Checklist

----

### DIFF
--- a/strands-py/src/strands/memory/extraction/resolve_extraction_config.py
+++ b/strands-py/src/strands/memory/extraction/resolve_extraction_config.py
@@ -1,0 +1,106 @@
+"""Resolves a store's ``extraction`` setting into a concrete config.
+
+The single place the ``bool | ExtractionConfig`` shorthand is interpreted and
+per-store defaults are applied, so the :class:`~strands.memory.memory_manager.MemoryManager`
+and :class:`~strands.memory.extraction.coordinator.ExtractionCoordinator` never
+re-apply defaults or normalize shapes themselves.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from ..types import MemoryStore, _has_method
+from .model_extractor import ModelExtractor
+from .triggers import IntervalTrigger
+from .types import (
+    DEFAULT_MEMORY_MESSAGE_FILTER,
+    ExtractionConfig,
+    ExtractionTrigger,
+    Extractor,
+    MemoryMessageFilter,
+)
+
+# Default cadence when an ``ExtractionConfig`` omits its ``trigger``: extract every N turns.
+DEFAULT_EXTRACTION_TRIGGER_TURNS = 5
+
+
+@dataclass
+class _ResolvedExtractionConfig:
+    """An :class:`ExtractionConfig` with every field resolved to a concrete value.
+
+    Produced by :func:`_resolve_extraction_config` so the ``MemoryManager`` and
+    ``ExtractionCoordinator`` never have to re-apply defaults or normalize shapes.
+
+    Attributes:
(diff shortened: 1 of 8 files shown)


strands-py/src/strands/memory/extraction/resolve_extraction_config.py
```diff
@@ -0,0 +1,106 @@
...
+from .types import (
+    DEFAULT_MEMORY_MESSAGE_FILTER,
+    ExtractionConfig,
+    ExtractionTrigger,
+    Extractor,
+    MemoryMessageFilter,
+)
+
+# Default cadence when an ``ExtractionConfig`` omits its ``trigger``: extract every N turns.
+DEFAULT_EXTRACTION_TRIGGER_TURNS = 5
```
JackYPCOnline, line 25: Nit: non-blocking. if type and function are private, this could be private as well.

JackYPCOnline (approved):

strands-py/src/strands/memory/extraction/resolve_extraction_config.py
opieter-aws: Will do in follow-up

PR #2812, opened 2026-06-15
feat(middleware): result handling, model-state isolation, and system-prompt fidelity

zastrowm:
## Description

Follow-up to the internal `InvokeModelStage` middleware (#2760), aligning the Python port with the TypeScript behavioral spec. Two behavior changes.

**1. Result handling.** The `MiddlewareResult` sentinel no longer rides the event stream. Instead the last yielded event *is* the result, matching the SDK convention (`ModelStopReason` ends `stream_messages()`, `ToolResultEvent` ends tool execution). The chain is now transparent. Output handlers — which transform the result — take/return a `MiddlewareResult` wrapper so we can extend it later without changing the signature.

```python
# Before: integration site sniffs the stream for the sentinel
async for event in registry.invoke(InvokeModelStage, ctx, terminal):
    if isinstance(event, MiddlewareResult):
        model_result = event.value
    else:
        yield event

# After: events flow straight through; the last one is the result
async for event in registry.invoke(InvokeModelStage, ctx, terminal):
    yield event
stop_reason, message, usage, metrics = event["stop"]
```

**2. Model-state isolation.** Model runtime state is snapshotted before the chain and written back only on success, so middleware can't corrupt provider state (e.g. server-side response IDs) via the `agent` escape hatch.

```python
# Before: terminal streamed against live agent state
model_state=agent._model_state,

# After: stream against a snapshot, write back after the chain succeeds
model_state_snapshot = copy.deepcopy(agent._model_state)
# ... terminal uses model_state_snapshot ...
agent._model_state = model_state_snapshot
```

## Public API Changes

`InvokeModelStage` is internal, so no public surface changes. For middleware authors, the Output-phase handler shape changes to a `MiddlewareResult` wrapper:

```python
def output_handler(result):  # result: MiddlewareResult
    stop_reason, message, usage, metrics = result.value["stop"]
    return result.replace(value=ModelStopReason(stop_reason="end_turn", message=message, usage=usage, metrics=metrics))
```

Wrap and Input handlers are unchanged.

## Related Issues

- #2735 (Middleware – Python tracking issue), builds on #2760
- Behavioral spec: `strands-ts/src/middleware/README.md`

## Documentation PR

None — internal stage; divergences documented in `strands-py/src/strands/_middleware/README.md`.

## Type of Change

New feature

## Testing

## Checklist

----


strands-py/src/strands/_middleware/stages.py
```diff
@@ -28,19 +27,11 @@ class InvokeModelContext:
...
-
-    stop_reason: StopReason
-    message: Message
-    usage: Usage
-    metrics: Metrics
-
-
-InvokeModelStage: MiddlewareStage[InvokeModelContext, InvokeModelResult, Any] = MiddlewareStage(name="invokeModel")
+# TResult is the result event (ModelStopReason — the last event from the model chain);
+# TEvent is the type streamed through the chain.
```
JackYPCOnline: nit: do we still have generic types used here? or should it be an example

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -542,27 +544,35 @@ async def _handle_model_execution(
...
+                    "Ensure middleware forwards events from next()."
                 )
-            stop_reason = model_result.stop_reason
-            message = model_result.message
-            usage = model_result.usage
-            metrics = model_result.metrics
+
+            # Write the post-stream model state back to the agent. Skipped on error
+            # (exception propagates and we never reach here), matching TS semantics.
+            agent._model_state = model_state_snapshot
```
JackYPCOnline, line 572: Question: So it only mutate the agent state after all succeeded. Is there a use case that we want to preserve the partial state?

zastrowm: Yep ^ this is intentional.  At some point if we support interrupts here we might have to do something, but for now this is what we expect

JackYPCOnline (approved):

PR #2814, opened 2026-06-15
fix(site): inject sitemap lastmod by matching git path prefix

yonib05:
## Summary

The `sitemap-lastmod` plugin (`site/src/plugins/sitemap-lastmod.ts`) was emitting zero `<lastmod>` entries for the entire sitemap (all ~665 URLs missing).

Root cause is a path-prefix mismatch. The build runs from `site/`, so `git log --name-only` reports file paths relative to the repo root (e.g. `site/src/content/docs/...`), and those become the keys in the lastmod map. But `urlToContentPaths()` built lookup candidates without the `site/` prefix (e.g. `src/content/docs/...`), so `lastModMap.get(candidate)` never matched.

The fix reconciles both sides onto the same repo-root-relative path space. It derives the prefix from `git rev-parse --show-prefix` (returns `site/` when run from `site/`, empty at the repo root) and prepends it to each lookup candidate. When git is unavailable the prefix falls back to empty and the existing graceful-degradation behavior is preserved, so the sitemap simply lacks lastmod rather than failing the build. The `/api/` skip behavior is unchanged.

## Test plan

- `npm run build` from `site/` succeeds.
- `<lastmod>` count in `dist/sitemap-0.xml`: before fix 0, after fix 199.
- Spot-checked a docs URL: `https://strandsagents.com/docs/user-guide/build-with-ai/` now has `<lastmod>2026-05-22T14:49:05.000Z</lastmod>`, a plausible ISO commit date.
- Remaining URLs without lastmod are the 407 `/api/` pages (intentionally skipped) plus dynamically generated routes (homepage, blog index, blog author pages) that do not map to a single content file.
- `npm test` (316 passed) and `npm run typecheck` pass.

### DIFF
--- a/site/src/plugins/sitemap-lastmod.ts
+++ b/site/src/plugins/sitemap-lastmod.ts
@@ -28,11 +28,10 @@ function buildLastModMap(contentDir: string): Map<string, string> {
   try {
     // Single git command: output all commits with dates and affected files
     // Format: each commit outputs its ISO date, then the list of files
-    const output = execFileSync(
-      'git',
-      ['log', '--format=%cI', '--name-only', '--diff-filter=ACMR', '--', contentDir],
-      { encoding: 'utf-8', maxBuffer: 50 * 1024 * 1024 }
-    )
+    const output = execFileSync('git', ['log', '--format=%cI', '--name-only', '--diff-filter=ACMR', '--', contentDir], {
+      encoding: 'utf-8',
+      maxBuffer: 50 * 1024 * 1024,
+    })
 
     // Parse: lines alternate between date lines and file path lines,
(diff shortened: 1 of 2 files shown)


yonib05:
Addressed both review points in fcbed64f:

- **posix join**: `urlToContentPaths` now uses `path.posix.join` so candidates keep forward slashes and match the git-emitted keys on Windows too.
- **test coverage**: exported `urlToContentPaths` and added `site/test/sitemap-lastmod.test.ts` with table-driven cases asserting the full candidate array (with and without a git prefix) plus a no-backslash guard.

zastrowm (approved):

Issue #1746, opened 2026-02-21
[BUG] PydanticSerializationUnexpectedValue warnings when Anthropic SDK returns ParsedTextBlock in message content


tomkitewing:
### Checks

- [x] I have updated to the lastest minor and patch version of Strands
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched [./issues](./issues?q=) and there are no duplicates of my issue

### Strands Version

1.27.0

### Python Version

3.13

### Operating System

Debian 13.1

### Installation Method

pip

### Steps to Reproduce

Install Strands with the anthopic extra version 0.83.0. Then run:

```python
import os

from strands import Agent
from strands.models.anthropic import AnthropicModel

api_key = os.environ.get("ANTHROPIC_API_KEY")

model = AnthropicModel(
  client_args={"api_key": api_key},
  model_id="claude-sonnet-4-20250514",
  max_tokens=1024,
)
agent = Agent(model=model, callback_handler=None)

result = agent("Say hello in one sentence.")
```

Running that script will print a bunch of pydantic warnings.

### Expected Behavior

No warnings printed. I don't think this is breaking anything.

### Actual Behavior

Here are the warnings:

```
/workspace/.venv/lib/python3.13/site-packages/pydantic/main.py:464: UserWarning: Pydantic serializer warnings:
  PydanticSerializationUnexpectedValue(Expected `ParsedTextBlock[TypeVar]` - serialized value may not be as expected [field_name='content', input_value=ParsedTextBlock(citations...xt', parsed_output=None), input_type=ParsedTextBlock])
  PydanticSerializationUnexpectedValue(Expected `ThinkingBlock` - serialized value may not be as expected [field_name='content', input_value=ParsedTextBlock(citations...xt', parsed_output=None), input_type=ParsedTextBlock])
  PydanticSerializationUnexpectedValue(Expected `RedactedThinkingBlock` - serialized value may not be as expected [field_name='content', input_value=ParsedTextBlock(citations...xt', parsed_output=None), input_type=ParsedTextBlock])
  PydanticSerializationUnexpectedValue(Expected `ToolUseBlock` - serialized value may not be as expected [field_name='content', input_value=ParsedTextBlock(citations...xt', parsed_output=None), input_type=ParsedTextBlock])
[...]

PR #2821, opened 2026-06-16
fix(models/anthropic): eliminate ParsedTextBlock Pydantic serialization warnings

### LINKED ISSUE #1865: [BUG] Strands Agent AnthropicModel emits Pydantic serialization warnings with Anthropic SDK v0.84.0
### Checks

### Strands Version

v1.29.0

### Python Version

3.12

### Operating System

macOS 26.3

### Installation Method

pip

### Steps to Reproduce

When using strands-agents v1.29.0 with anthropic SDK v0.84.0, every agent invocation using AnthropicModel produces multiple Pydantic serialization warnings on stderr. The Anthropic SDK returns ParsedTextBlock objects in message content, but the Strands message serialization schema expects a discriminated union of specific block types (ThinkingBlock, ToolUseBlock, RedactedThinkingBlock, ServerToolUseBlock, WebSearchToolResultBlock, ParsedTextBlock[TypeVar]). The ParsedTextBlock doesn't match any discriminator, triggering 6+ warning lines per invocation.

### Expected Behavior

No Pydantic serialization warnings during normal agent operation. The Strands AnthropicModel provider should correctly handle ParsedTextBlock objects from the current Anthropic SDK, either by:

### Actual Behavior

The agent functions correctly — results, tool calls, and metrics are all accurate — but stderr is polluted with noisy warnings.

### Additional Context

Updating the Pydantic model's content union type to accept ParsedTextBlock properly
Normalizing ParsedTextBlock to TextBlock before serialization
Suppressing known-safe serialization warnings internally
Impact
Affects every Strands user on anthropic SDK >= 0.84.0 with AnthropicModel
Produces 6+ warning lines per agent call, cluttering output
Particularly problematic for demos, workshops, and getting-started experiences where clean output matters
Creates confusion — developers may think their code is broken
No workaround short of suppressing Python warnings globally (which hides real issues)
Supporting Evidence
Minimal reproduction — 10 lines of code:

from strands.models.anthropic import AnthropicModel

model = AnthropicModel(model_id="claude-haiku-4-5-20251001", max_tokens=1000)

@tool
def get_weather(city: str) -> str:
    """Get weather for a city."""
    return f"{city}: 60°F, sunny"

agent = Agent(model=model, tools=[get_weather])
result = agent("What's the weather in New York?")
Warning output:

PydanticSerializationUnexpectedValue(Expected `ParsedTextBlock[TypeVar]` - serialized value may not be as expected [field_name='content', input_value=ParsedTextBlock(citations...xt', parsed_output=None), input_type=ParsedTextBlock])
PydanticSerializationUnexpectedValue(Expected `ThinkingBlock` - serialized value may not be as expected ...)
PydanticSerializationUnexpectedValue(Expected `RedactedThinkingBlock` - serialized value may not be as expected ...)
PydanticSerializationUnexpectedValue(Expected `ToolUseBlock` - serialized value may not be as expected ...)
PydanticSerializationUnexpectedValue(Expected `ServerToolUseBlock` - serialized value may not be as expected ...)
PydanticSerializationUnexpectedValue(Expected `WebSearchToolResultBlock` - serialized value may not be as expected ...)
Versions: strands-agents==1.29.0, anthropic==0.84.0, Python 3.13

### Possible Solution

Update the AnthropicModel provider's message content type definition to properly handle ParsedTextBlock from the latest Anthropic SDK. The serialization schema likely needs to either include ParsedTextBlock in the union or convert it to a plain TextBlock before passing it through Pydantic.

### Related Issues

_No response_

FadelT:
## Problem

When using `strands-agents >= v1.29.0` with `anthropic SDK >= v0.84.0`, every agent invocation using `AnthropicModel` produces 6+ Pydantic serialization warnings on stderr. The Anthropic SDK returns `ParsedTextBlock` objects in streaming responses, but calling `event.model_dump()` triggers serialization warnings because `ParsedTextBlock` doesn't match the discriminated union schema.

**Impact:**
- Affects all Strands users with `anthropic >= 0.84.0`
- Produces noisy stderr output (6+ warnings per agent call)
- Particularly problematic for demos and getting-started experiences
- Creates confusion - developers may think their code is broken

Closes #1865

## Solution

Extract a `_build_event_dict()` helper method that constructs event dicts by directly accessing typed attributes instead of calling `model_dump()`. This approach:

1. **Follows proven pattern** - Same technique as the existing `message_stop` fix from #1746
2. **Reduces complexity** - Extracts nested conditional logic into a focused helper
3. **Maintains compatibility** - No changes to public API or behavior
4. **Complete coverage** - Handles all event types: `message_start`, `content_block_start`, `content_block_delta`, `content_block_stop`, `message_stop`

**Why this approach:**
The Anthropic SDK's `ParsedTextBlock` objects have all required attributes accessible directly. Building dicts from these typed attributes bypasses Pydantic's serialization layer entirely, eliminating the warnings without changing SDK contracts.

### Real SDK Testing (anthropic==0.109.2)
Tested with actual Anthropic API calls using `claude-sonnet-4-6`:

**Before fix:** Multiple Pydantic warnings on stderr
**After fix:** Zero warnings ✓

Test coverage:
- ✅ Text streaming - no warnings
- ✅ Tool-use streaming - no warnings  
- ✅ All event types exercised

### Unit Tests
- ✅ 59/59 tests passing
- ✅ Added 2 comprehensive warning-detection tests:
  - `test_stream_text_no_pydantic_warnings` 
  - `test_stream_tool_use_no_pydantic_warnings`
- ✅ Updated existing test mocks to match new implementation

### `src/strands/models/anthropic.py`
- Add `_build_event_dict()` static method - constructs dicts without Pydantic serialization
- Refactor `stream()` to use helper for all event types (reduces cognitive complexity)

### `tests/strands/models/test_anthropic.py`  
- Fix mock objects to use proper attribute assignment (not constructor kwargs)
- Add comprehensive warning-free streaming tests for text and tool-use
- Update `test_structured_output` mocks to match new pattern

## Checklist

- [x] Code follows project style guidelines
- [x] Tests added/updated and passing (59/59)
- [x] Verified with real Anthropic SDK (0.109.2)
- [x] No breaking changes
- [x] Follows pattern from #1746
- [x] Documentation updated (inline docstrings)

## References

- Fixes #1865
- Related to #1746 (message_stop fix)

### DIFF
--- a/strands-py/src/strands/models/anthropic.py
+++ b/strands-py/src/strands/models/anthropic.py
@@ -376,6 +376,80 @@ class AnthropicModel(Model):
             case _:
                 raise RuntimeError(f"event_type=<{event['type']} | unknown type")
 
+    @staticmethod
+    def _build_event_dict(event: Any) -> dict[str, Any]:
+        """Build event dict directly from attributes to avoid Pydantic serialization warnings.
+
+        When the Anthropic SDK (>=0.84.0) returns ParsedTextBlock objects, calling model_dump()
+        triggers Pydantic serialization warnings. This method constructs dicts by directly
+        accessing typed attributes instead.
+
+        Args:
+            event: Stream event from Anthropic SDK.
+
+        Returns:
+            Dictionary representation of the event suitable for format_chunk().
+        """
+        if event.type == "message_start":
+            return {
+                "type": "message_start",
+                "message": {
+                    "id": event.message.id,
+                    "model": event.message.model,
+                    "role": event.message.role,
+                    "stop_reason": event.message.stop_reason,
+                    "stop_sequence": event.message.stop_sequence,
+                    "type": event.message.type,
+                    "usage": event.message.usage.model_dump(),
+                },
+            }
+
+        if event.type == "content_block_start":
+            content_block_dict: dict[str, Any] = {"type": event.content_block.type}
+            if event.content_block.type == "tool_use":
+                content_block_dict["id"] = event.content_block.id
+                content_block_dict["name"] = event.content_block.name
+                content_block_dict["input"] = event.content_block.input
+            else:
+                content_block_dict["text"] = getattr(event.content_block, "text", "")
+
+            return {
+                "type": "content_block_start",
+                "index": event.index,
+                "content_block": content_block_dict,
+            }
+
+        if event.type == "content_block_delta":
+            delta_dict: dict[str, Any] = {"type": event.delta.type}
+            if event.delta.type == "text_delta":
+                delta_dict["text"] = event.delta.text
+            elif event.delta.type == "input_json_delta":
+                delta_dict["partial_json"] = event.delta.partial_json
+            elif event.delta.type == "thinking_delta":
+                delta_dict["thinking"] = event.delta.thinking
+            elif event.delta.type == "signature_delta":
+                delta_dict["signature"] = event.delta.signature
+
+            return {
+                "type": "content_block_delta",
+                "index": event.index,
+                "delta": delta_dict,
+            }
+
+        if event.type == "content_block_stop":
+            return {
+                "type": "content_block_stop",
+                "index": event.index,
+            }
+
+        if event.type == "message_stop":
+            return {
+                "type": "message_stop",
+                "message": {"stop_reason": event.message.stop_reason},
+            }
+
+        return event.model_dump()
+
     @override
     async def count_tokens(
         self,
@@ -463,17 +537,7 @@ class AnthropicModel(Model):
                 logger.debug("got response from model")
                 async for event in stream:
                     if event.type in AnthropicModel.EVENT_TYPES:
-                        if event.type == "message_stop":
-                            # Build dict directly to avoid Pydantic serialization warnings
-                            # when the message contains ParsedTextBlock objects (issue #1746)
-                            yield self.format_chunk(
-                                {
-                                    "type": "message_stop",
-                                    "message": {"stop_reason": event.message.stop_reason},
-                                }
-                            )
-                        else:
-                            yield self.format_chunk(event.model_dump())
+                        yield self.format_chunk(self._build_event_dict(event))
 
                 try:
                     message_snapshot = await stream.get_final_message()
--- a/strands-py/tests/strands/models/test_anthropic.py
+++ b/strands-py/tests/strands/models/test_anthropic.py
@@ -738,19 +738,23 @@ def test_format_chunk_unknown(model):
 async def test_stream(anthropic_client, model, alist):
     mock_event_1 = unittest.mock.Mock(
         type="message_start",
-        dict=lambda: {"type": "message_start"},
-        model_dump=lambda: {"type": "message_start"},
+        message=unittest.mock.Mock(
+            id="msg_123",
+            model="m1",
+            role="assistant",
+            stop_reason=None,
+            stop_sequence=None,
+            type="message",
+            usage=unittest.mock.Mock(model_dump=lambda: {"input_tokens": 0, "output_tokens": 0}),
+        ),
     )
     mock_event_2 = unittest.mock.Mock(
         type="unknown",
-        dict=lambda: {"type": "unknown"},
-        model_dump=lambda: {"type": "unknown"},
     )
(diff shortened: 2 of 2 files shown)


strands-py/src/strands/models/anthropic.py
```diff
@@ -376,6 +376,80 @@ def format_chunk(self, event: dict[str, Any]) -> StreamEvent:
...
+                "message": {
+                    "id": event.message.id,
+                    "model": event.message.model,
+                    "role": event.message.role,
+                    "stop_reason": event.message.stop_reason,
+                    "stop_sequence": event.message.stop_sequence,
+                    "type": event.message.type,
+                    "usage": event.message.usage.model_dump(),
+                },
+            }
```
kimnamu, line 405: For `message_start`, `format_chunk` returns only `{"messageStart": {"role": "assistant"}}` and never reads `event.message.*` — so this whole `message` dict is unused output, and its `usage.model_dump()` is the only `model_dump()` call left in the helper. Dropping it keeps the fix fully minimal. Verified: existing tests only assert `{"messageStart": {"role": "assistant"}}`, so 59/59 still pass with this.
```suggestion
        if event.type == "message_start":
            return {"type": "message_start"}
```

kimnamu (commented):
Thanks for tackling this — the noisy stderr on every agent call is a real getting-started papercut, and modeling it on the #1746 `message_stop` pattern is the right call. I'm not a maintainer, just another Strands/Anthropic user who reviewed this carefully and ran the branch locally (anthropic 0.111.0, pydantic 2.13.4). Two things I confirmed and one suggestion:

**Confirmed equivalent (no stream data loss).** I fed real Anthropic SDK events through both the old `event.model_dump()` path and the new `_build_event_dict()` path and compared `format_chunk` output — identical for all 5 event types, including all four deltas (`text`/`input_json`/`thinking`/`signature`). 59/59 tests pass, ruff clean.

1. *(non-blocking)* For `message_start`, `format_chunk` returns only `{"messageStart": {"role": "assistant"}}` and never reads `event.message.*`. The helper builds the full `message` dict (id/model/role + `usage.model_dump()`), which is unused — and that `usage.model_dump()` is the only surviving `model_dump()` call in the helper. It's harmless, but since the PR is framed as minimal you could drop it (I applied this and the existing `test_stream`/`test_structured_output` assertions, which only check `{"messageStart": {"role": "assistant"}}`, still pass — 59/59):
   ```suggestion
        if event.type == "message_start":
            return {"type": "message_start"}
   ```

2. *(worth a look)* The two new `*_no_pydantic_warnings` tests assert `len(pydantic_warnings) == 0`, but since the events are `Mock`s, no real `ParsedTextBlock` ever flows through — reverting the source fix makes them fail with `TypeError: 'Mock' object is not subscriptable`, not on the warning assertion. So they pin the new dict-building plumbing rather than the warning regression itself. If you want them to actually lock in the #1865 behavior, constructing a real `anthropic.types.parsed_message.ParsedTextBlock` (with a `parsed` payload) inside a `Message` and streaming it would make the `== 0` assertion meaningful. Not blocking — the dict-equivalence coverage is already solid.

Clean, well-scoped fix — I'd love to see it land. (This review was prepared with the help of an AI agent (Claude Code); I ran the branch and verified each point myself before posting.)

FadelT:
Thanks for the detailed review @kimnamu!

**Re: message_start simplification:** You're absolutely right that the usage dict is unused. Since format_chunk only returns `{"messageStart": {"role": "assistant"}}`, we could simplify to:
```python
if event.type == "message_start":
    return {"type": "message_start"}
```
This is a good cleanup - I'm marking it non-blocking since the current impl is harmless, but happy to simplify if maintainers prefer it.

**Re: test mocks vs real ParsedTextBlock:** Agreed - the tests currently use Mocks which verify plumbing but not the actual #1865 warning regression. Constructing a real `anthropic.types.parsed_message.ParsedTextBlock` would make the `len(pydantic_warnings) == 0` assertion meaningful. I'll look into adding that.

The core fix (dict-building instead of model_dump) is solid and tested. Appreciate the thorough verification!

lizradway:
Closing out fixed in https://github.com/strands-agents/harness-sdk/pull/2044/changes

PR #2829, opened 2026-06-16
feat(memory): add GraphitiMemoryStore vended memory backend

agent-of-mkmeral:
## Description

Adds `GraphitiMemoryStore`, the first concrete `MemoryStore` shipped in the Python SDK, backed by [Graphiti](https://github.com/getzep/graphiti) — Zep's Apache-2.0 temporal knowledge-graph library purpose-built for agent memory.

**Why Graphiti.** Most memory backends are flat vector stores; Graphiti maintains a *bi-temporal* knowledge graph that tracks both when a fact was recorded (transaction time) and when it was true in the world (valid time). That distinction is what lets an agent reason about corrections and stale facts instead of just returning the nearest embedding. It maps almost 1:1 onto our `MemoryStore` contract: hybrid `search()` → `MemoryStore.search()`, `add_episode()` → `add()`/`add_messages()`. Chosen in #296 as the in-core memory pick over cognee (a heavier ECL/control-plane framework).

**Optional, zero-friction by default.** `graphiti-core` lands under a new `graphiti` extra and is imported lazily — importing `strands.memory` never requires it, and a missing install raises a clear `ImportError` only when a write is attempted. The store takes an already-configured `Graphiti` client, so the graph backend (Neo4j / FalkorDB / Kuzu) and any LLM/embedder clients are the caller's to wire up. No database is needed to construct or unit-test the store.

## Public API

```python
from graphiti_core import Graphiti
from strands import Agent
from strands.memory.stores import GraphitiMemoryStore

client = Graphiti("bolt://localhost:7687", "neo4j", "password")
await client.build_indices_and_constraints()

store = GraphitiMemoryStore(client=client, name="graph", writable=True, group_id="user-123")
agent = Agent(memory_manager=MemoryManager(stores=[store]))
```

Search results surface graph identifiers and the bi-temporal timestamps (`created_at`/`expired_at`, `valid_at`/`invalid_at`) in `MemoryEntry.metadata`. The store implements both write sinks, so by default the `MemoryManager` routes extraction to `add_messages` (server-side extraction, no extra model call); single-fact writes go through `add` as `text` episodes. An optional `group_id` namespaces the store (Graphiti's tenancy primitive), scoping every search and stamping every write so one client can back many isolated stores.

## Related Issues

Refs #296

## Type of Change

New feature

## Testing

Unit tests mock the Graphiti client and require no graph database: search/add mapping, bi-temporal metadata propagation, `max_search_results` precedence, `group_id` scoping (incl. empty-string normalization), `reference_time` UTC normalization, lazy-import `ImportError`, `_has_method`/`_has_write_sink` detection, and end-to-end `MemoryManager` integration. `ruff format`/`ruff check` clean and `mypy ./src` passes strict with `graphiti-core` absent.

- [x] I ran the relevant checks (`pytest tests/strands/memory`, `ruff`, `mypy`)

## Checklist
- [x] I have added tests that prove the feature works

---

cc @mkmeral — first concrete vended `MemoryStore` in the Python SDK; would value your eyes on the protocol mapping and the `graphiti` extra placement (kept out of `all` since it pulls neo4j/openai/numpy).

_Note: `graphiti-core` requires a graph backend to exercise end to end; that path was not run here (no provisioned DB), but the client boundary is fully mocked in unit tests._

### DIFF
--- a/strands-py/pyproject.toml
+++ b/strands-py/pyproject.toml
@@ -58,6 +58,7 @@ sagemaker = [
     "openai>=1.68.0,<3.0.0",  # SageMaker uses OpenAI-compatible interface
 ]
 otel = ["opentelemetry-exporter-otlp-proto-http>=1.30.0,<2.0.0"]
+graphiti = ["graphiti-core>=0.3.0,<1.0.0"]
 docs = [
     "sphinx>=5.0.0,<10.0.0",
     "sphinx-rtd-theme>=1.0.0,<4.0.0",
@@ -221,6 +222,14 @@ exclude = ["src/strands/experimental/bidi"]
 [[tool.mypy.overrides]]
 module = ["strands.experimental.bidi.*"]
 follow_imports = "skip"
+
+# ``graphiti-core`` is an optional dependency (the ``graphiti`` extra) and is intentionally kept
+# out of the ``all`` group because it pulls a heavy backend stack (neo4j, openai, numpy, ...). It is
+# imported lazily in ``strands.memory.stores.graphiti``, so static analysis runs without it
+# installed; ignore the missing import rather than bloat the analysis environment.
+[[tool.mypy.overrides]]
+module = ["graphiti_core.*"]
+ignore_missing_imports = true
 
 [tool.ruff]
 line-length = 120
--- a/strands-py/src/strands/memory/__init__.py
+++ b/strands-py/src/strands/memory/__init__.py
@@ -20,6 +20,7 @@ from .extraction.types import (
     MemoryMessageFilter,
 )
 from .memory_manager import MemoryManager
+from .stores import GraphitiMemoryStore
 from .types import (
     AddMessagesContext,
     InjectionFormatContext,
@@ -45,6 +46,7 @@ __all__ = [
     "ExtractionTriggerContext",
     "Extractor",
     "ExtractorContext",
+    "GraphitiMemoryStore",
     "InjectionConfig",
     "InjectionContext",
     "InjectionFormatContext",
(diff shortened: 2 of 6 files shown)


agent-of-mkmeral:
@mkmeral this is ready for review — the first concrete vended `MemoryStore` in the Python SDK.

A few decisions worth a look:

1. **Protocol conformance via `_has_method`.** `GraphitiMemoryStore` is a plain class (not a `MemoryStore` subclass), so `_has_method` correctly detects `add`/`add_messages` as implemented and `get_tools` as absent. Because it implements both write sinks, the `MemoryManager` defaults extraction to the `add_messages` (server-side) path — which is the right shape for Graphiti, since it extracts episodes itself with no extra model call.

2. **`graphiti` extra kept out of `all`.** `graphiti-core` pulls neo4j + openai + numpy, so bundling it into `all` would bloat the analysis/test env. It gets its own extra plus a `mypy` override (`ignore_missing_imports`) so static analysis stays green without it installed, and a lazy import that raises a clear `ImportError` only on the write path.

3. **Bi-temporal fidelity.** Search maps each `EntityEdge` to a `MemoryEntry` carrying `created_at`/`expired_at` (transaction time) and `valid_at`/`invalid_at` (valid time) as ISO-8601 in metadata. No synthetic `score` — Graphiti's basic search doesn't return one, so inventing it would be misleading.

I ran an adversarial self-review pass before opening: it caught a naive-datetime bug in `reference_time` (now normalized to tz-aware UTC, since Graphiti compares against aware UTC) and an empty-string `group_id` edge case (now normalized to unscoped). Both fixed with tests.

Not merging — leaving it for your review. The only path I couldn't exercise end to end is against a live graph DB; the client boundary is fully mocked in unit tests.

agent-of-mkmeral:

PR #2830, opened 2026-06-16
feat(strands-py): add BubblewrapSandbox backend

agent-of-mkmeral:
## Motivation

The sandbox module ships Docker and SSH backends, but both delegate isolation to an external runtime: Docker needs a daemon, SSH needs a remote host. There is no way to harden the *local* `PosixShellSandbox` with real OS-level isolation and zero infrastructure.

Per the in-core sandbox decision in #296, `BubblewrapSandbox` fills that gap. It wraps command/code execution in an unprivileged [bubblewrap](https://github.com/containers/bubblewrap) (`bwrap`) Linux user-namespace jail — the same engine behind Flatpak. It was chosen over microsandbox (which requires a daemon + microVM): bubblewrap is a single `bwrap` binary, **no daemon, no server, no vendor auth**.

**Scope of isolation:** Linux-only, **namespace-level** (user/mount/PID/IPC/UTS/cgroup/net) — *not* VM-level. It hardens the existing shell sandbox; it is not a hypervisor boundary.

## Public API

```python
from strands.sandbox import BindMount, BubblewrapConfig, BubblewrapSandbox

# Secure by default: no network, cleared env, all namespaces unshared,
# only read-only system dirs visible.
sandbox = BubblewrapSandbox(
    BubblewrapConfig(
        rw_binds=[BindMount("/tmp/workspace", "/work")],
        working_dir="/work",
        network=False,  # default
    )
)
result = await sandbox.execute("echo hello")
```

`BubblewrapSandbox` subclasses `PosixShellSandbox`, so file and code operations are inherited; it implements only `execute_streaming`, prefixing the shell command with a configurable `bwrap` argv. `BubblewrapConfig` exposes bind mounts (ro/rw), tmpfs paths, a network toggle (`--unshare-net` by default), env passthrough, working dir, the `bwrap` binary path, and raw `extra_args`. `is_bubblewrap_available()` reports support so callers (and tests) can degrade gracefully.

## Design notes

- **Secure-by-default**: `--unshare-all` (network re-shared only on opt-in via `--share-net`), `--die-with-parent`, `--new-session` (TIOCSTI hardening), `--clearenv`. Only a curated read-only set of system dirs (`/usr`, `/bin`, `/lib`…) is mounted — `/home`, `/root`, `/etc`, `/var` are not exposed.
- **No shell injection**: the command is positional after `--`; env vars are passed as discrete `--setenv` argv (verbatim, never shell-evaluated), mirroring the Docker backend's `-e` handling.
- **Cancellation-safe**: the inner `stream_process` generator is wrapped in `contextlib.aclosing`, so its process-tree kill runs deterministically on cancel/early-break.
- **Graceful detection**: missing `bwrap` surfaces as exit code 127 with an install hint rather than a crash; non-Linux short-circuits.

## Testing

Unit tests mock the subprocess and require **no** `bwrap` install. Integration tests under `tests_integ/sandbox/` run a real jail and skip cleanly when `bwrap` is absent or unprivileged user namespaces are disabled. The implementation was additionally exercised end-to-end against a real `bwrap` (file round-trips, env clearing, read-only enforcement, timeout, cancellation).

Refs #296.

cc @mkmeral

### DIFF
--- a/strands-py/src/strands/sandbox/__init__.py
+++ b/strands-py/src/strands/sandbox/__init__.py
@@ -11,6 +11,9 @@ interface from ``strands-ts/src/sandbox/`` (the behavioral oracle):
   :meth:`~strands.sandbox.base.Sandbox.execute_streaming`.
 - :class:`DockerSandbox` — run commands in a Docker container via ``docker exec``.
 - :class:`SshSandbox` — run commands on a remote host via OpenSSH.
+- :class:`BubblewrapSandbox` — run commands inside an unprivileged ``bwrap``
+  Linux user-namespace jail (no daemon, namespace-level isolation, Linux-only),
+  configured via :class:`BubblewrapConfig` and :class:`BindMount`.
 - Data types: :class:`StreamChunk`, :class:`FileInfo`, :class:`OutputFile`,
   :class:`ExecutionResult`, and the :data:`StreamType` literal.
 - :data:`LANGUAGE_PATTERN` — interpreter-name validation pattern.
@@ -32,6 +35,12 @@ Example:
 """
 
 from .base import Sandbox
+from .bubblewrap import (
+    BindMount,
+    BubblewrapConfig,
+    BubblewrapSandbox,
+    is_bubblewrap_available,
+)
 from .constants import LANGUAGE_PATTERN
 from .docker import DockerSandbox
 from .shell import PosixShellSandbox
@@ -39,6 +48,9 @@ from .ssh import SshSandbox
 from .types import ExecutionResult, FileInfo, OutputFile, StreamChunk, StreamType
 
 __all__ = [
+    "BindMount",
+    "BubblewrapConfig",
+    "BubblewrapSandbox",
     "DockerSandbox",
     "ExecutionResult",
     "FileInfo",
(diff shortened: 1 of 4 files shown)


agent-of-mkmeral:
@mkmeral ready for review.

This is the in-core sandbox pick from #296: `BubblewrapSandbox` hardens `PosixShellSandbox` with an unprivileged `bwrap` Linux user-namespace jail — **no daemon, namespace-level (not VM) isolation, Linux-only**.

Highlights:
- Secure-by-default `BubblewrapConfig` (no network, `--clearenv`, `--unshare-all`, `--die-with-parent`, `--new-session`; only read-only system dirs visible).
- Mirrors the Docker/SSH backend pattern — implements only `execute_streaming`, inherits file/code ops.
- Env via discrete `--setenv` argv (verbatim, no shell injection); command positional after `--` (no flag injection).
- `contextlib.aclosing` makes process-tree cleanup deterministic on cancel.
- 39 unit tests (no `bwrap` required) + integration tests that skip when `bwrap`/userns is unavailable. mypy strict + ruff clean. Also exercised end-to-end against a real `bwrap`.

Went through an adversarial self-review loop before opening; addressed cancellation-cleanup, tmpfs/bind ordering, working-dir source-of-truth, and an availability-probe false-positive. Left for human review — not merging.

agent-of-mkmeral:

PR #2831, opened 2026-06-16
feat(strands-py): add Presidio PII-redaction intervention handler

agent-of-mkmeral:
## Motivation

The interventions framework ships `Transform` (mutate event content in-place) but no vended handler exercises it for the most common real-world need: **stripping PII before it reaches the model, downstream tools, or persisted history/logs**. Per #296, Presidio is the in-core interventions pick — it's a pure-Python, Microsoft-maintained PII detection/redaction library with no server or auth dependency, so it slots cleanly behind the existing `Transform` action.

This is **additive and non-overlapping** with the Cedar authorization bet in design 0006 (authz answers *"may this happen?"* via `Deny`; redaction answers *"sanitize what flows through"* via `Transform`). It was preferred over OPA precisely because OPA collides with the Cedar authz direction, whereas Presidio fills a distinct gap.

## Public API

New vended handler `PresidioRedaction`, mirroring the `HumanInTheLoop` package structure under `vended_interventions/presidio/`:

```python
from strands import Agent
from strands.vended_interventions.presidio import PresidioRedaction

# Redact emails/phones across tool I/O and model context (default events)
agent = Agent(interventions=[PresidioRedaction(entities=["EMAIL_ADDRESS", "PHONE_NUMBER"])])

# Mask instead of replace, scanning only tool results
agent = Agent(interventions=[PresidioRedaction(operator="mask", events=["after_tool_call"])])
```

The handler returns a `Transform` for each configured lifecycle event whose `apply` runs Presidio's `AnalyzerEngine` + `AnonymizerEngine` and rewrites detected PII in-place; unconfigured events are a no-op `Proceed`. Configurable knobs: entity types, which events to scan, anonymization operator (`replace`/`mask`/`hash`/`redact`) + params, language, and confidence threshold.

Redaction reaches every text-bearing content shape — `text`, `json` (recursively, strings only), `toolUse.input`, nested `toolResult`, and `reasoningContent` — so PII does not slip through non-text paths. Scanning `before_model_call`/`after_model_call` mutates `agent.messages` in place, redacting the **persisted** history (not just the request), which keeps PII out of session storage and logs. Scope, idempotency, streaming, and the cost of full-history sweeps are documented on the class.

## Dependency

Presidio is an **optional** dependency under a new `presidio` extra (`presidio-analyzer`, `presidio-anonymizer`; spaCy model required at runtime, e.g. `python -m spacy download en_core_web_lg`). It is lazily imported with a clear `ImportError`, and the unit tests mock the engines so neither Presidio nor a spaCy model is needed in CI.

Refs #296. Engines were additionally smoke-tested against real Presidio to confirm the `replace`/`mask`/`hash`/`redact` operator wiring.

cc @mkmeral

### DIFF
--- a/strands-py/pyproject.toml
+++ b/strands-py/pyproject.toml
@@ -58,6 +58,12 @@ sagemaker = [
     "openai>=1.68.0,<3.0.0",  # SageMaker uses OpenAI-compatible interface
 ]
 otel = ["opentelemetry-exporter-otlp-proto-http>=1.30.0,<2.0.0"]
+# PII detection/redaction for the PresidioRedaction intervention handler.
+# Requires a spaCy model at runtime, e.g. `python -m spacy download en_core_web_lg`.
+presidio = [
+    "presidio-analyzer>=2.2.0,<3.0.0",
+    "presidio-anonymizer>=2.2.0,<3.0.0",
+]
 docs = [
     "sphinx>=5.0.0,<10.0.0",
     "sphinx-rtd-theme>=1.0.0,<4.0.0",
--- a/strands-py/src/strands/vended_interventions/__init__.py
+++ b/strands-py/src/strands/vended_interventions/__init__.py
@@ -4,5 +4,6 @@ Ready-to-use InterventionHandler implementations for common control patterns.
 """
 
 from .hitl import HumanInTheLoop
+from .presidio import PresidioRedaction
 
-__all__ = ["HumanInTheLoop"]
+__all__ = ["HumanInTheLoop", "PresidioRedaction"]
--- a/strands-py/src/strands/vended_interventions/presidio/__init__.py
+++ b/strands-py/src/strands/vended_interventions/presidio/__init__.py
@@ -1,0 +1,29 @@
+"""PII redaction intervention for Strands agents.
+
+Detects and anonymizes personally identifiable information (PII) in tool I/O and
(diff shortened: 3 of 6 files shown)


agent-of-mkmeral:
@mkmeral ready for review.

This is the **Presidio** in-core interventions pick from #296 — a `PresidioRedaction` handler wired to the `Transform` action for PII detection/redaction across tool I/O and model context. It mirrors the `vended_interventions/hitl` package structure exactly.

Highlights:
- **Pure-Python, no server/auth.** Presidio is an *optional* dep (`presidio` extra) with a lazy import + helpful `ImportError`; unit tests mock the engines so CI needs neither Presidio nor a spaCy model.
- **Additive, no Cedar overlap.** Redaction (`Transform`) is orthogonal to the Cedar authz (`Deny`) chosen in design 0006 — this is why it was preferred over OPA, which collides with that direction.
- **Leak-path hardened.** After an adversarial self-review I extended redaction to cover `json`, `toolUse.input`, nested `toolResult`, and `reasoningContent` blocks (not just plain `text`), and added a test proving `after_model_call` redacts the *persisted* assistant message in history, not just the in-flight response.
- Configurable entities / events / operator (`replace`/`mask`/`hash`/`redact`) / language / confidence threshold. Operator wiring smoke-tested against real Presidio.

Leaving it for human review — not merging. Happy to adjust the default event set or scope docs if you'd prefer a tighter default.

agent-of-mkmeral:

PR #2835, opened 2026-06-16
feat(python): vended tools/plugins for sandbox

gautamsirdeshmukh:
## Description

- Add vended tools (bash, file_editor), sandbox-compatible
- Update vended plugins (context_offloader, skills) to be sandbox-compatible
- Update agent init to auto register default tools set on sandbox

## Related Issues

- https://github.com/strands-agents/harness-sdk/pull/2563

## Type of Change

New feature

## Testing

## Checklist

----

### DIFF
--- a/strands-py/src/strands/agent/agent.py
+++ b/strands-py/src/strands/agent/agent.py
@@ -364,6 +364,21 @@ class Agent(AgentBase):
         if load_tools_from_directory:
             self.tool_watcher = ToolWatcher(tool_registry=self.tool_registry)
 
+        # Register tools vended by an explicitly-configured sandbox, applying the sandbox's
+        # tool_prefix to names (like MCP's prefix for server-vended tools). The host default
+        # vends nothing. A tool is skipped if the user already registered one with that name.
+        if not isinstance(self._sandbox, NotASandboxLocalEnvironment):
+            prefix = self._sandbox.tool_prefix
+            for sandbox_tool in self._sandbox.get_tools():
+                prefixed = sandbox_tool.with_prefix(prefix) if prefix else sandbox_tool
+                if prefixed.tool_name in self.tool_registry.registry:
+                    logger.debug(
+                        "tool_name=<%s> | sandbox-vended tool skipped, user already registered a tool with this name",
+                        prefixed.tool_name,
+                    )
+                else:
+                    self.tool_registry.register_tool(prefixed)
+
         self.event_loop_metrics = EventLoopMetrics()
 
         # Initialize tracer instance (no-op if not configured)
--- a/strands-py/src/strands/agent/agent.py
+++ b/strands-py/src/strands/agent/agent.py
@@ -364,20 +364,16 @@ class Agent(AgentBase):
         if load_tools_from_directory:
             self.tool_watcher = ToolWatcher(tool_registry=self.tool_registry)
 
-        # Register tools vended by an explicitly-configured sandbox, applying the sandbox's
-        # tool_prefix to names (like MCP's prefix for server-vended tools). The host default
-        # vends nothing. A tool is skipped if the user already registered one with that name.
-        if not isinstance(self._sandbox, NotASandboxLocalEnvironment):
-            prefix = self._sandbox.tool_prefix
-            for sandbox_tool in self._sandbox.get_tools():
-                prefixed = sandbox_tool.with_prefix(prefix) if prefix else sandbox_tool
-                if prefixed.tool_name in self.tool_registry.registry:
-                    logger.debug(
-                        "tool_name=<%s> | sandbox-vended tool skipped, user already registered a tool with this name",
-                        prefixed.tool_name,
-                    )
-                else:
-                    self.tool_registry.register_tool(prefixed)
+        # Register tools vended by the sandbox. The host default vends nothing. A tool
+        # is skipped if the user already registered one with that name.
+        for sandbox_tool in self._sandbox.get_tools():
+            if sandbox_tool.tool_name in self.tool_registry.registry:
+                logger.debug(
+                    "tool_name=<%s> | sandbox-vended tool skipped, user already registered a tool with this name",
+                    sandbox_tool.tool_name,
+                )
+            else:
+                self.tool_registry.register_tool(sandbox_tool)
 
         self.event_loop_metrics = EventLoopMetrics()
 
--- a/strands-py/src/strands/sandbox/base.py
+++ b/strands-py/src/strands/sandbox/base.py
@@ -32,9 +32,6 @@ from .types import ExecutionResult, FileInfo, StreamChunk
 
 logger = logging.getLogger(__name__)
 
-DEFAULT_TOOL_PREFIX = "sandbox"
-"""Default prefix applied to sandbox-vended tool names (e.g. ``sandbox_bash``)."""
-
 
 class Sandbox(ABC):
     """Abstract execution environment.
@@ -205,16 +202,17 @@ class Sandbox(ABC):
 
     # ---- Tool vending ----
 
-    tool_prefix: str | None = DEFAULT_TOOL_PREFIX
-    """Prefix applied to tool names when registered on an agent (e.g. ``"sandbox"`` produces
-    ``sandbox_bash``). Set to ``None`` to disable prefixing. Defaults to ``"sandbox"``."""
+    @staticmethod
+    def _prefixed_name(name: str, prefix: str | None = "sandbox") -> str:
+        """Build a prefixed tool name (e.g. ``("bash", "sandbox")`` -> ``"sandbox_bash"``)."""
+        return f"{prefix}_{name}" if prefix else name
 
     def get_tools(self) -> list[AgentTool]:
         """Tools this sandbox vends to an agent.
 
-        Returned tools are registered when the agent initializes, with
-        :attr:`tool_prefix` applied to their names. A tool is skipped if the user
-        already registered one with the same (prefixed) name. The base
+        Returned tools are registered when the agent initializes; a tool is
+        skipped if the user already registered one with the same name. Overrides
+        use :meth:`_prefixed_name` to build namespaced tool names. The base
         implementation vends nothing; concrete sandboxes override this to provide
         sandbox-routed tools (e.g. bash, file editor).
 
--- a/strands-py/src/strands/sandbox/base.py
+++ b/strands-py/src/strands/sandbox/base.py
@@ -27,9 +27,13 @@ from abc import ABC, abstractmethod
 from collections.abc import AsyncGenerator
 from typing import Any
 
+from ..types.tools import AgentTool
 from .types import ExecutionResult, FileInfo, StreamChunk
 
 logger = logging.getLogger(__name__)
+
+DEFAULT_TOOL_PREFIX = "sandbox"
+"""Default prefix applied to sandbox-vended tool names (e.g. ``sandbox_bash``)."""
 
 
 class Sandbox(ABC):
@@ -198,6 +202,26 @@ class Sandbox(ABC):
             FileNotFoundError: If the directory does not exist.
         """
         ...
+
+    # ---- Tool vending ----
+
+    tool_prefix: str | None = DEFAULT_TOOL_PREFIX
+    """Prefix applied to tool names when registered on an agent (e.g. ``"sandbox"`` produces
+    ``sandbox_bash``). Set to ``None`` to disable prefixing. Defaults to ``"sandbox"``."""
+
+    def get_tools(self) -> list[AgentTool]:
+        """Tools this sandbox vends to an agent.
+
+        Returned tools are registered when the agent initializes, with
+        :attr:`tool_prefix` applied to their names. A tool is skipped if the user
+        already registered one with the same (prefixed) name. The base
+        implementation vends nothing; concrete sandboxes override this to provide
+        sandbox-routed tools (e.g. bash, file editor).
+
+        Returns:
+            The tools to register, or an empty list.
+        """
+        return []
 
     # ---- Non-streaming convenience methods ----
 
--- a/strands-py/src/strands/sandbox/base.py
+++ b/strands-py/src/strands/sandbox/base.py
@@ -202,19 +202,12 @@ class Sandbox(ABC):
 
     # ---- Tool vending ----
 
-    @staticmethod
-    def _prefixed_name(name: str, prefix: str | None = "sandbox") -> str:
-        """Build a prefixed tool name (e.g. ``("bash", "sandbox")`` -> ``"sandbox_bash"``)."""
-        return f"{prefix}_{name}" if prefix else name
-
     def get_tools(self) -> list[AgentTool]:
         """Tools this sandbox vends to an agent.
 
         Returned tools are registered when the agent initializes; a tool is
-        skipped if the user already registered one with the same name. Overrides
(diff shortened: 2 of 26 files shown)


gautamsirdeshmukh:
Addressing code coverage

agent-of-mkmeral (commented):
## Review: vended tools/plugins for sandbox (Python)

Reviewed the full 24-file diff, ran the suite, and exercised the tools e2e against real processes/filesystem. This is a careful, faithful port of the TS oracle — the `file_editor`, sandbox tool-vending wiring (`get_tools` + `with_prefix`), and the async storage/skills refactors all match the TS structure and pass cleanly. Strong work.

**Verification I ran** (Python 3.13, host sandbox):
- ✅ `282 passed` across the new/modified test files (bash, file_editor, agent_sandbox, docker, ssh, context_offloader, skills)
- ✅ Full suite: `2808 passed` (the only failures are pre-existing OTEL-version mismatches in `test_tracer.py`/`test_config.py` — this PR doesn't touch telemetry)
- ✅ `ruff check`, `ruff format --check`, `pyright` all clean on the changed files
- ✅ 17/17 custom e2e checks: file_editor create/overwrite-guard/view/view_range/str_replace/multi-occurrence/insert/traversal/relative/nonexistent; bash env+cwd persistence, stdout/stderr split, restart, timeout, sandbox-routed make_bash
- ✅ Agent auto-registration: `sandbox_bash`/`sandbox_file_editor` prefixing, `tool_prefix=None`, user-tool-wins, host-default-vends-nothing (all covered by your tests too)

**Findings:** 1 🔴 functional bug (with a verified fix), 1 🟡 latent robustness nit. Details inline.

_Reviewed at head `8e67b48`._

strands-py/src/strands/agent/agent.py
```diff
@@ -364,6 +364,21 @@ def __init__(
         if load_tools_from_directory:
             self.tool_watcher = ToolWatcher(tool_registry=self.tool_registry)
 
+        # Register tools vended by an explicitly-configured sandbox, applying the sandbox's
+        # tool_prefix to names (like MCP's prefix for server-vended tools). The host default
+        # vends nothing. A tool is skipped if the user already registered one with that name.
+        if not isinstance(self._sandbox, NotASandboxLocalEnvironment):
```
mkmeral: why? let's just make gettools return empty?

or if we want this to be a valid use case (i guess that's the scenario), then we should keep track of it in a separate way.

strands-py/src/strands/agent/agent.py
```diff
@@ -364,6 +364,21 @@ def __init__(
...
+            prefix = self._sandbox.tool_prefix
```
mkmeral: can we move the prefixing to sandbox itself on `getTools`? we need to do the same thing in TS too

strands-py/src/strands/vended_plugins/skills/agent_skills.py
```diff
@@ -3,11 +3,24 @@
 This module provides the AgentSkills class that extends the Plugin base class
 to add Agent Skills support. The plugin registers a tool for activating
 skills, and injects skill metadata into the system prompt.
+
+Filesystem skill sources are loaded through the agent's sandbox (host or
+container) at ``init_agent`` time, not at construction, so each agent sees the
+skills present on its own filesystem. Skill instances and ``https://`` URLs are
+sandbox-independent and resolve eagerly at construction.
+
+Mirrors ``strands-ts/src/vended-plugins/skills/agent-skills.ts``. One idiomatic
```
mkmeral: we should not have TS details in python docstrings

strands-py/src/strands/vended_tools/bash/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+    # Stateless, routed through a sandbox
+    agent = Agent(tools=[make_bash()])
+    ```
+"""
+
+from .bash import DEFAULT_BASH_DESCRIPTION, bash, make_bash
+from .types import SANDBOX_BASH_DESCRIPTION, BashSessionError, BashTimeoutError
+
+__all__ = [
+    "DEFAULT_BASH_DESCRIPTION",
```
mkmeral: do we need to export this? also should we add an underscore? is this something we expect users to use?

strands-py/src/strands/vended_tools/bash/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+    "SANDBOX_BASH_DESCRIPTION",
+    "BashSessionError",
```
mkmeral: why do we need custom errors?

strands-py/src/strands/vended_tools/bash/bash.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def start(self) -> None:
+        """Start the bash process if it is not already running."""
+        import subprocess
+
+        if self._process is not None and self._process.poll() is None:
+            return
+        try:
+            env: Mapping[str, str] = {**os.environ, "PS1": "", "PS2": ""}
+            self._process = subprocess.Popen(
```
mkmeral: why do we need this?

mkmeral: we don't need to be worried about backwards compatibility in python

strands-py/src/strands/vended_tools/bash/bash.py
```diff
@@ -0,0 +1,271 @@
...
+
+    State such as variables and the working directory persists within the
+    session. Runs on the host without sandboxing; use only with trusted input.
+
+    Args:
+        mode: Operation mode: "execute" to run a command, "restart" to restart the session.
+        command: The bash command to execute (required when mode is "execute").
+        timeout: Timeout in seconds (default: 120, applies only to execute mode).
+    """
+    import asyncio
```
mkmeral: nit: imports inside method

strands-py/src/strands/vended_tools/bash/bash.py
```diff
@@ -0,0 +1,271 @@
...
+    if session is None:
+        session = _BashSession(timeout)
+        _sessions[agent] = session
+
+    # The session is blocking (subprocess + threads); run it off the event loop.
+    return await asyncio.to_thread(session.run, command, timeout)
+
+
+def make_bash(
+    sandbox: Sandbox | None = None,
```
mkmeral: why is the api sandbox focused? is that the main use case?

strands-py/src/strands/vended_tools/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+    SANDBOX_BASH_DESCRIPTION,
+    BashSessionError,
+    BashTimeoutError,
+    bash,
+    make_bash,
+)
+from .file_editor import DEFAULT_FILE_EDITOR_DESCRIPTION, file_editor, make_file_editor
+
+__all__ = [
+    "DEFAULT_BASH_DESCRIPTION",
```
mkmeral: any reason we have more exports then `bash, makeBash, fileEditor, makeFileEditor`?

strands-py/src/strands/vended_plugins/context_offloader/storage.py
```diff
@@ -73,7 +76,7 @@ class Storage(Protocol):
         backend with built-in lifecycle management (e.g., S3 lifecycle policies).
     """
 
-    def store(self, key: str, content: bytes, content_type: str = "text/plain") -> str:
+    async def store(self, key: str, content: bytes, content_type: str = "text/plain") -> str:
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Will update PR description and add label after addressing remaining callouts

strands-py/src/strands/agent/agent.py
gautamsirdeshmukh: We can loop unconditionally, removing this (matches check used in TS but we don't have the same node/browser clash here)

strands-py/src/strands/vended_plugins/skills/agent_skills.py
gautamsirdeshmukh: Removing

strands-py/src/strands/agent/agent.py
gautamsirdeshmukh: Will do, will update in TS via 2806

strands-py/src/strands/vended_tools/bash/__init__.py
gautamsirdeshmukh: Trimmed exports down to bash, file_editor, make_bash, make_file_editor

strands-py/src/strands/vended_tools/bash/__init__.py
gautamsirdeshmukh: Updated, we don't need to export from here just use internally

strands-py/src/strands/vended_plugins/skills/agent_skills.py
gautamsirdeshmukh: Will raise a follow up PR as well to clean up other such references in the python repo

strands-py/src/strands/vended_tools/bash/bash.py
gautamsirdeshmukh: Updated

strands-py/src/strands/sandbox/base.py
```diff
@@ -199,6 +200,27 @@ async def list_files(self, path: str, **kwargs: Any) -> list[FileInfo]:
         """
         ...
 
+    # ---- Tool vending ----
+
+    @staticmethod
+    def _prefixed_name(name: str, prefix: str | None = "sandbox") -> str:
```
mkmeral: just hardcoding names is much easier in sandboxes imo

strands-py/src/strands/vended_tools/bash/bash.py
gautamsirdeshmukh: Updated, removed unnecessary over-engineering

strands-py/src/strands/vended_tools/bash/types.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+SANDBOX_BASH_DESCRIPTION = (
+    "Executes bash shell commands. Each call runs in a fresh shell; "
+    "state such as variables and the working directory does not persist across calls."
+)
+"""Description for the sandbox-routed, stateless bash tool."""
+
+
+class BashTimeoutError(Exception):
```
mkmeral: why do we need custom exceptions?

strands-py/src/strands/vended_tools/bash/bash.py
mkmeral: ping

strands-py/src/strands/vended_tools/bash/types.py
gautamsirdeshmukh: Removing

strands-py/src/strands/sandbox/base.py
gautamsirdeshmukh: Removing

strands-py/src/strands/vended_tools/bash/types.py
gautamsirdeshmukh: Responding to make_bash callout below --> making sandbox a kwarg instead of first level param for make methods

strands-py/src/strands/vended_tools/bash/bash.py
gautamsirdeshmukh: Updated to make this kwarg

agent-of-mkmeral (commented):
**Re-review after `b208764a` (“another round of review”)**

Took another look at the current head. This commit is a solid simplification — it removed the entire persistent host-session `bash` (the `_BashSession`/sentinel-drain machinery) and collapsed the module to a single stateless, sandbox-routed tool: `bash = make_bash()`. As a happy side effect, this **moots both of my original findings**:

- 🔴 ~~Bash drops output sharing the sentinel's line~~ — **gone with the host session.** I re-verified the old failure cases (`printf foo`, `printf 'bar baz'`) end-to-end through the sandbox path: output is now preserved correctly (the sandbox's own `execute()` captures stdout/stderr, no sentinel involved). ✅
- 🟡 ~~`_PrefixedTool.__getattr__` infinite recursion~~ — **gone**; `with_prefix`/`_PrefixedTool` were removed and prefixing is now hardcoded names (`sandbox_bash`/`sandbox_file_editor`) in `get_tools()`, per @mkmeral's suggestion.

### Verification (on `b208764a`)
- **Tests:** `61 passed` in `tests/strands/vended_tools/`; `475 passed` across `tests/strands/sandbox/` + `tests/strands/vended_plugins/`.
- **Static:** `ruff check`, `ruff format --check`, and `mypy ./src/strands/vended_tools ./src/strands/sandbox` all clean.
- **e2e (15/15):** default `bash` reading `agent.sandbox` at call time; bound-sandbox `bash`; stdout/stderr separation; no-trailing-newline preservation; statelessness across calls; `SandboxTimeoutError` re-raised (not wrapped); non-timeout errors wrapped as `RuntimeError`; file_editor create/view/str_replace/insert; tool naming + prefixing.

### Status of earlier feedback
<details>
<summary>Follow-up status</summary>

- ✅ **Bash timeout doubling** — moot (host session removed).
- ✅ **Recursive `walk` bare except** — now `except OSError` at L326. Resolved.
- ✅ **File-editor size guard** — docstring note added (“Checked after reading because `list_files` does not reliably report size”), documenting it as a post-hoc safeguard. Good enough.
- ✅ **`make_bash` API shape / sandbox param** — now keyword-only (`*, sandbox=None`), and the two-entry-point ambiguity is resolved by collapsing to a single `bash` default + `make_bash()` factory.
- ⚠️ **Storage blocking I/O** — **still open.** `FileStorage.store`/`retrieve` (`write_bytes`/`read_bytes`/`mkdir`) and `S3Storage` (`put_object`/`get_object`) still run synchronously inside `async def` (`grep -c to_thread storage.py` → 0). Note: the “match the pattern in bash.py” argument from the prior review no longer applies (bash.py's `asyncio.to_thread` went away with the host session), but the event-loop-blocking concern itself stands.
- ⏳ **`Storage` async contract change + `needs-api-review`** — still pending the description update and label (acknowledged by author).

</details>

### Two new, minor items from this commit
1. 🟡 **Stale docstring** in `vended_tools/__init__.py`: it still says *“The `bash` tool runs a persistent shell on the host”* — but that host shell was deleted in this very commit. `bash` is now stateless and sandbox-routed. One-line doc fix to avoid misleading users.
2. 📝 **TS parity note (intentional divergence).** The description frames these as faithful TS ports, but this is now a deliberate divergence worth calling out: TS (`strands-ts/src/vended-tools/bash/`) still ships a host-session `bash.ts` **and** `make-bash.ts`, and wraps errors as `BashTimeoutError`/`BashSessionError`. Python now has **no** host session and re-raises `SandboxTimeoutError` / wraps others as `RuntimeError`. I think the Python shape is cleaner (and fits “no backwards-compat baggage in Python”), but the parity claim should be updated so reviewers don't expect 1:1, and the TS side (#2806) reconciled.

### Net
Much tighter than the first pass — my two findings are resolved by deletion and the surface is simpler. None of the remaining items are merge-blockers individually, but before this lands I'd want: (a) the stale `__init__.py` docstring fixed, (b) the `Storage` async I/O either wrapped in `to_thread` or explicitly documented as known-blocking, and (c) the description updated for the `Storage` async contract change + `needs-api-review` label and the PY/TS divergence. Still draft, so all very landable. Nice cleanup, @gautamsirdeshmukh.

mkmeral (approved):

PR #2836, opened 2026-06-16
feat: add durable identifiers to messages

### LINKED ISSUE #2805: [FEATURE] Add durable identifiers to messages
### SDK Language

Python

### Problem Statement

Message extraction by MemoryStores currently needs its own deduping solution with identifiers, because messages do not carry their own durable ids. The only tracking we can currently provide is in-session hwm sequence numbers, which don't scale beyond a session. This poses problems for session restore.

### Proposed Solution

Add durable IDs to each message

### Use Case

MemoryStores can use these durable IDs together with a SessionID to identify uniqueness and take care of deduplication.

### Alternatives Solutions

_No response_

### Additional Context

_No response_

opieter-aws:
## Description

Gives every message a durable, stable `tracking_id` (`trackingId` in TypeScript) — a UUID that identifies a message wherever it's observed: `MessageAddedEvent` subscribers, session persistence, and snapshots. It is never sent to model providers, and it survives a save/restore round-trip unchanged.

Why: messages have no durable identity today. `SessionMessage.message_id` is just an ordinal index that shifts under truncation and restore, so nothing serves as a stable per-message key. A memory store needs one to build a `(session_id, tracking_id)` tuple and deduplicate extracted messages across sessions. This PR only establishes the id on `Message`; wiring it into memory extraction is a follow-up.

**How the id is assigned differs by SDK, on purpose** — each leverages its own `Message` representation:

- **TypeScript** — `Message` is a class, so the **constructor** assigns the id: a caller-supplied id is preserved, otherwise a fresh UUID is minted. `trackingId` is therefore a required `string` (never `undefined`) on a live `Message`. Because the type system forces `new Message(...)`, *every* message — including one a caller builds and pushes onto `agent.messages` directly — gets an id for free. `MessageData` (the serialized form) keeps `trackingId` optional so legacy/serialized data without one still deserializes; the constructor backfills it on the way in.
- **Python** — `Message` is a `TypedDict` (a plain dict, no constructor), so there's no construction hook. Instead the agent **backfills before each turn**: it ensures a `tracking_id` on every message in history before the model call. This covers messages appended ad-hoc via `agent.messages.append({...})` (they get an id on the next turn) and legacy sessions restored without ids. An id is preserved if already present, so it's stable across restore.

The observable contract is the same in both: every message in history ends up with an id, a caller-supplied id is preserved, and copying/restoring a message does not re-key it (the same message keeps the same id across sessions). Every message-rebuild site — redaction, sliding-window truncation, summarization, and context injection — carries the existing id forward rather than minting a new one.

**Naming.** The field is `tracking_id`, not `id`, deliberately: `id` collides with concepts that already exist across the SDK (`agent_id`, `tool_use_id`, `SessionMessage.message_id`, content-block ids). This is a distinct thing — a handle for *tracking* a message across sessions and stores, not a wire identity a provider sees.

## Public API Changes

`Message` gains a durable `tracking_id` (`trackingId` in TypeScript). It's assigned automatically; callers don't normally set it, and the field is null-safe to read directly, so no accessor is added to the public surface.

```python
agent("Hello")
message = agent.messages[-1]
message["tracking_id"]        # "686b8abc-1db4-4145-aa60-9615c04f50ef"

# A caller supplying its own id should use a UUID v4:
agent.messages.append({"role": "user", "content": [{"text": "hi"}], "tracking_id": str(uuid.uuid4())})
```

```typescript
await agent.invoke('Hello')
const message = agent.messages.at(-1)!
message.trackingId   // always present on a live Message (minted if not supplied)

// A caller supplying its own id should use a UUID v4:
agent.messages.push(new Message({ role: 'user', content: [new TextBlock('hi')], trackingId: crypto.randomUUID() }))
```

Backward compatible: existing code that builds messages without a `tracking_id` is unaffected (Python backfills on the next turn; the TS constructor mints one), and a caller-supplied id is always preserved.

## Related Issues

Resolves: #2805

## Documentation PR

No documentation changes

## Type of Change

New feature

## Testing

Both SDKs' unit suites pass (Python and TypeScript). A cross-subsystem test drives a real agent end to end and asserts every message carries an id after each path that adds one outside the normal append: summarization, agentic `summarize_context`, sliding-window truncation, intervention `Guide` injection, and session-restore repair — plus a case proving an ad-hoc `agent.messages.append({...})` gets an id on the next turn. Each guard was verified to fail if its id-assignment is removed. A regression test asserts the id never reaches the model-provider payload. (`hatch run prepare`'s static-analysis step can't bootstrap locally due to an unrelated native build failure in the optional `[cedar]` extra; ruff/mypy on changed files and the full suites pass, and CI runs the gate cleanly.)

## Checklist

----


JackYPCOnline:
My concern is scope and layer. As a DTO that memory stores read for dedup, this is fine. But the PR already persists it inside SessionMessage and snapshots, so it's effectively entering the session layer. 

If the natural next step is for it to become the session manager's record key. I'd want us to explicitly not do that, and ideally keep durable identity on the persistence wrapper rather than on the provider-facing Message.

It also overlaps with the planned transcript work, which models messages as a DAG. A flat message.id that's preserved across redaction/truncation can't express edits or branches, so shipping it as "the" durable identity now will collide with transcript node identity later. 

Finally, on correctness as a unique key: it's preserved across deep copies (as_tool, graph, swarm), assigned only at append chokepoints (the summarizer already needed a manual fix to keep the invariant), and never validated for uniqueness or format with caller-supplied ids trusted as-is. Before we lean on (session_id, message_id) anywhere, I'd like the uniqueness scope defined (per session_id), copy/sub-agent semantics decided, and assignment enforced at the persistence boundary rather than by convention.

In short: can we scope this PR to "opaque durable id for memory dedup," document it as non-authoritative and not a session/transcript identity, and defer the canonical-identity decision to the transcript design?

opieter-aws:
Converting back to draft until we have clarity on the Transcripts story. Need a design to be able to determine if it can for sure replace durable message identifiers (place-in-array)

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -637,6 +637,7 @@ async def _handle_model_execution(
         stream_trace.end()
 
         # Add the response message to the conversation
+        _ensure_tracking_id(message)
         agent.messages.append(message)
```
opieter-aws, line 641: Will follow-up with a separate refactor to instead use the message's append method in this file, so we can avoid the duplicate tracking id logic here

strands-ts/src/types/messages.ts
```diff
@@ -43,12 +43,40 @@ export interface MessageData {
    */
   content: ContentBlockData[]
 
+  /**
+   * Durable, stable tracking identifier for the message, assigned when the message is added to the
+   * conversation. Survives session save/restore, and is stripped before model calls.
+   */
+  trackingId?: string
```
lizradway, line 50: should this be readonly?

lizradway: dont love the naming `trackingId` and `ensure_tracking_id` tbh


why not just `id`? `message_id`? `durable_id`? some ideas

strands-ts/src/types/messages.ts
```diff
@@ -43,12 +43,40 @@ export interface MessageData {
...
+
+/**
+ * Assign a durable tracking id to the message in place if it does not already have a usable one.
+ *
+ * A message that already carries a non-empty tracking id (e.g. restored from a session or supplied
+ * by a caller) keeps it, so the same message has a stable identifier everywhere it is observed. A
+ * missing or empty-string tracking id is treated as absent and replaced.
+ * @internal
+ */
+export function ensureTrackingId(message: Message): void {
```
lizradway: should we return the id string for convenience?

strands-py/src/strands/types/content.py
```diff
@@ -232,18 +233,42 @@ class Message(TypedDict):
...
 
+def _generate_tracking_id() -> str:
+    """Generate a durable tracking identifier for a message.
+
+    Returns a canonical UUID v4 string
+    """
+    return str(uuid.uuid4())
+
+
+def _ensure_tracking_id(message: Message) -> None:
```
JackYPCOnline: can we also mention that tracking id is only unique within a session here, since this we left message without id as is.

strands-py/src/strands/agent/conversation_manager/summarizing_conversation_manager.py
```diff
@@ -192,6 +192,8 @@ def _summarize_oldest(self, agent: "Agent") -> None:
 
         # Generate summary
         self._summary_message = self._generate_summary(to_summarize, agent)
+        # Assign tracking id to the summary message since it bypasses the append method.
+        _ensure_tracking_id(self._summary_message)
```
JackYPCOnline, line 196: A regression test asserting "no id-less messages after summarize/intervention/repair" would be the highest-value addition.

strands-ts/src/types/messages.ts
opieter-aws: > should this be readonly?

It can't be since `ensureTrackingId` needs to be able to mutate it

> why not just `id`? `message_id`? `durable_id`?

We had a discussion around this in the office, and we went with `tracking_id`:

`id` gives the impression that it is a deterministic identifier, which it is not. `message_id` gives this same impression and collides with `SessionMessage.message_id`. `durable_id` would be an option, but IMO `tracking_id` better conveys the purpose of the identifier (to track a message across sessions).

Any thoughts here @JackYPCOnline @zastrowm ?

strands-ts/src/types/messages.ts
opieter-aws: Sure, can add, good suggestion!

strands-py/src/strands/agent/conversation_manager/summarizing_conversation_manager.py
opieter-aws: You mean an integration test?

strands-ts/src/types/messages.ts
JackYPCOnline: I lean on using the naming what we've discussed since it ensures the semantic correctness.

strands-py/src/strands/types/content.py
opieter-aws: What do you mean with 'only unique within a session'? It's canonical uuidv4

JackYPCOnline: Not from uuid, what if a agentB copy agentA.messages. they would have same messageId. But it should be uique to agentB in its own session right.

strands-py/src/strands/_context_manager/modes/agentic/agentic_context.py
zastrowm, line 1: Overall questions - how do messages appended by the caller (via agent.messages) get an id?

Also, why not backfill ids

strands-py/src/strands/types/content.py
```diff
@@ -232,18 +233,47 @@ class Message(TypedDict):
...
     role: Role
+    tracking_id: NotRequired[str]
     metadata: NotRequired[MessageMetadata]
 
 
 Messages = list[Message]
 """A list of messages representing a conversation."""
 
 
+def _generate_tracking_id() -> str:
```
zastrowm, line 255: In our docs, we should give guidance on how users can specify their own and what we recommend (uuid4)

strands-ts/src/agent/agent.ts
```diff
@@ -2651,6 +2651,7 @@ export class Agent implements LocalAgent, InvokableAgent {
         this.messages[lastIndex] = new Message({
```
zastrowm, line 2658: Why not do this in the message constructor itself?

Basically can we just prevent a Message from being instantiated without a tracking id?  That way everything just starts to get one?

strands-ts/src/types/messages.ts
zastrowm: This should be in the PR description - but yes, I'm aligned with this.

strands-ts/src/types/messages.ts
```diff
@@ -69,14 +102,25 @@ export class Message implements JSONSerializable<MessageData> {
...
    * Optional metadata, not sent to model providers.
    */
   metadata?: MessageMetadata
 
-  constructor(data: { role: Role; content: ContentBlock[]; metadata?: MessageMetadata }) {
+  constructor(data: { role: Role; content: ContentBlock[]; trackingId?: string; metadata?: MessageMetadata }) {
     this.role = data.role
     this.content = data.content
+    if (data.trackingId !== undefined) {
+      this.trackingId = data.trackingId
```
zastrowm: Still pushing back on why this is allowed to be undefined?

strands-ts/src/agent/agent.ts
opieter-aws: In TS this is possible, but in Python `Message` is a `TypedDict`, so there’s no constructor. To get this same behaviour in Python would be a large refactor. Opted for parity between the languages, so paired implementation for Python with TS.

zastrowm: > To get this same behaviour in Python would be a large refactor.

Agreed, don't do in Python

> Opted for parity between the languages

I disagree with this; if the Message class gets us a boost in functionality in TS, we should leverage it.  It's one of the primary reasons we went with message classes in TS - that serves as the normalization mechanism

strands-py/src/strands/_context_manager/modes/agentic/agentic_context.py
opieter-aws: The input is routed through `_append_messages` which calls `_ensure_tracking_id`. That covers the caller input. For exceptions (like conversation management which direct editing of the messages array), `_ensure_tracking_id` can be called directly. There are some refactors we can do to minimize these exceptions (4 currently), but doing that as a follow-up.

opieter-aws: > why not backfill ids

Decided it's easiest not to given we can't mint one at construction (Given limitations in Python). In general, we don't want to mint new ones at restore since the id has to survive a save/restore. That’s why `undefined` is a valid state for the `trackingId` with this implementation

strands-ts/src/types/messages.ts
opieter-aws: See explanation above

strands-py/src/strands/_context_manager/modes/agentic/agentic_context.py
zastrowm: Should we start normalizing before or after after turn?  E.g. before we start the agent loop, ensure every message has a tracking id?  Why would we not want to do this?

I guess I'm poking at requiring developers to do this explicitly vs implicitly - why make it explicit?

strands-py/tests/strands/event_loop/test_streaming.py
```diff
@@ -1217,6 +1217,40 @@ async def test_stream_messages(agenerator, alist):
     )
 
 
+@pytest.mark.asyncio
+async def test_stream_messages_strips_id_before_model_call(agenerator, alist):
```
zastrowm, line 1221: Is this needed for a specific reason?  E.g. could models take it in without harm?

strands-ts/src/hooks/__tests__/events.test.ts
```diff
@@ -875,7 +876,7 @@ describe('toJSON serialization', () => {
 
       expect(json).toStrictEqual({
         type: 'modelMessageEvent',
-        message: { role: 'assistant', content: [{ text: 'Done' }] },
+        message: { role: 'assistant', content: [{ text: 'Done' }], trackingId: anyTrackingId },
```
zastrowm, line 879: Oi, the tests changes

zastrowm (approved):

strands-py/tests/strands/event_loop/test_streaming.py
opieter-aws: Providers can reject unexpected message keys, so passing a `tracking_id` may cause a `400`

strands-ts/src/hooks/__tests__/events.test.ts
opieter-aws: Yep, a sad consequence

opieter-aws:
Failing integration test is unrelated

strands-py/tests/strands/event_loop/test_streaming.py
zastrowm: Do we pass it into ModelProviders though?

PR #2839, opened 2026-06-16
feat(sandbox): add experimental StrandsShellSandbox (Python + TS)

mkmeral:
## Description
  
  [Strands Shell](https://github.com/strands-agents/shell) is a Bourne-compatible shell that runs entirely in-process — no `fork`, no
  `exec`, no syscalls — and mediates access declaratively: the agent only reaches the host paths you bind, the URLs you allowlist, and
  credentials it never sees. That makes it a natural execution backend for agents that need a real shell (run tests, grep, curl, edit files)
  without handing over the host.

  The SDK already has a `Sandbox` abstraction with Docker and SSH backends. This adds `StrandsShellSandbox`, a `Sandbox` backed by Strands
  Shell, in both the Python and TypeScript SDKs. It ships under `experimental` because the configuration surface is still settling and we
  want to iterate with community feedback before committing to it.
  
  Because Strands Shell is an optional, separately-installed package, it is an optional dependency in both SDKs — the SDK imports cleanly
  without it, and the sandbox raises a clear install hint only when constructed.

  ### Public API Changes

  **Python** — new `strands.experimental.sandbox.StrandsShellSandbox` (requires `pip install strands-agents[shell]`):

  ```python
  from strands import Agent
  from strands.experimental.sandbox import StrandsShellSandbox

  sandbox = StrandsShellSandbox(
      binds=[{"source": "/my/project", "destination": "/workspace", "mode": "copy"}],
      timeout=30.0,
  )
  # Vends sandbox_bash and sandbox_file_editor, auto-registered by the agent.
  agent = Agent(sandbox=sandbox)
  agent("List the Python files in /workspace and summarize them")
  ```

  **TypeScript** — new `StrandsShellSandbox` at `@strands-agents/sdk/experimental/sandbox/strands-shell` (requires the optional
  `@strands-agents/shell` peer dependency):

  ```typescript
  import { StrandsShellSandbox } from '@strands-agents/sdk/experimental/sandbox/strands-shell'

  const sandbox = new StrandsShellSandbox({
    binds: [{ source: '/my/project', destination: '/workspace', mode: 'copy' }],
    timeout: 30,
  })
  const agent = new Agent({ sandbox })
  await agent.invoke('List the files in /workspace and summarize them')
  ```

  Both vend `sandbox_bash` / `sandbox_file_editor` (built from the existing `make_bash` / `make_file_editor` vended-tool factories) via the
  sandbox's tool-vending hook (`get_tools()` / `getTools()`), so an agent constructed with the sandbox registers them automatically. Tool
  descriptions surface the sandbox's mounts, timeout, and allowlists so the model knows what it can reach; the injected credential values
  are never included.

  ### Design notes for reviewers

  - **File and code operations.** Strands Shell ships no `base64` command, so the base `PosixShellSandbox` file-transfer path doesn't apply.
  File ops use the shell's native VFS API instead (which also yields real `size` metadata), and `execute_code` stages the source to a temp
  VFS file and runs the interpreter against it.
  - **Thread-pinning (Python).** The native Python `Shell` (PyO3) is *unsendable* — it panics if created, used, or dropped on a thread other
  than its creator — while the SDK runs each `agent()` call's event loop on a fresh thread. The Python sandbox therefore confines the shell
  to one dedicated worker thread (`_ShellWorker`) and routes every operation through it, including teardown via a `weakref.finalize` that
  drops the shell on its owning thread. The TS native binding manages its own thread, so the TS sandbox just awaits it.
  - **Per-call `timeout` is intentionally ignored.** Strands Shell enforces a single wall-clock timeout set at construction; the per-call
  `timeout` from the base `Sandbox` contract has no effect here. This is documented on the class in both SDKs.

  ## Related Issues
  
  

  ## Documentation PR
  
  

  ## Type of Change

  New feature
  
  ## Testing

  Unit tests cover the thread-pinned worker lifecycle (including a regression test that the worker thread terminates and the shell is
  dropped on GC), command/code execution, native file ops, cwd/env scoping, error mapping, and the vended tool descriptions. An end-to-end
  test drives a scripted model through the auto-registered tools (create a file via the editor, read it back via bash). Integration tests
  exercise the same paths against the real `strands-shell` / `@strands-agents/shell` package and are skipped when it isn't installed.

  ## Checklist

  ----

  By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your
  choice.

### DIFF
--- a/strands-py/src/strands/experimental/sandbox/strands_shell.py
+++ b/strands-py/src/strands/experimental/sandbox/strands_shell.py
@@ -1,0 +1,419 @@
+"""Sandbox backed by `Strands Shell <https://github.com/strands-agents/shell>`_.
+
+:class:`StrandsShellSandbox` runs commands and file operations inside Strands
+Shell — a Bourne-compatible shell that executes entirely in userspace, with no
+``fork``/``exec``/syscalls. The agent only reaches what you declare: bound host
+paths, allowlisted URLs, and per-URL credentials it never sees.
+
+This is an **experimental** feature and may change without notice. It requires
+the optional ``strands-shell`` dependency::
+
+    pip install strands-agents[shell]
+
+Example::
+
+    from strands import Agent
+    from strands.experimental.sandbox import StrandsShellSandbox
+
+    sandbox = StrandsShellSandbox(
+        binds=[{"source": "/my/project", "destination": "/workspace", "mode": "copy"}],
+        timeout=30.0,
+    )
+    # The sandbox vends ``sandbox_bash`` and ``sandbox_file_editor`` tools, which
+    # the agent registers automatically.
+    agent = Agent(sandbox=sandbox)
+    agent("List the Python files in /workspace and summarize them")
+
+The native shell is *thread-pinned*: it must be created, used, and dropped on a
+single OS thread. This sandbox enforces that by confining the shell to a
+dedicated worker thread (:class:`._worker._ShellWorker`) and routing every
+operation through it, so it is safe to use from Strands' threaded tool execution
+and from asyncio.
+"""
+
+import asyncio
+import logging
+import shlex
+import uuid
+import weakref
+from collections.abc import AsyncGenerator, Callable, Sequence
+from typing import TYPE_CHECKING, Any
+
+from ...sandbox.base import Sandbox
+from ...sandbox.constants import LANGUAGE_PATTERN
+from ...sandbox.errors import SandboxPathNotFoundError
+from ...sandbox.posix_shell import build_shell_env_prefix
+from ...sandbox.types import ExecutionResult, FileInfo, StreamChunk
+from ...types.tools import AgentTool
+from ...vended_tools.bash import make_bash
+from ...vended_tools.bash.types import SANDBOX_BASH_DESCRIPTION
+from ...vended_tools.file_editor import make_file_editor
+from ...vended_tools.file_editor.file_editor import DEFAULT_FILE_EDITOR_DESCRIPTION
+from ._worker import _ShellWorker
+
+if TYPE_CHECKING:
+    import strands_shell
+
+logger = logging.getLogger(__name__)
+
+
+class StrandsShellSandbox(Sandbox):
+    """A :class:`~strands.sandbox.base.Sandbox` backed by Strands Shell.
+
+    Constructed with the same configuration surface as ``strands_shell.Shell``
+    (binds, credentials, allowlisted URLs, env, umask, timeout, resource limits)
+    plus the ability to load a TOML ``config_file``. The agent cannot change this
+    configuration — it is fixed by whoever creates the sandbox.
+
+    File operations use the shell's native VFS API (reporting real ``size``
+    metadata); command execution runs through the in-process shell. Code
+    execution writes the source to a temporary VFS file and runs the requested
+    interpreter against it (Strands Shell ships ``lua``; other interpreters are
+    only available if present in the sandbox).
+
+    :meth:`get_tools` vends ``sandbox_file_editor`` and ``sandbox_bash`` tools
+    bound to this sandbox, with descriptions that surface its mounts, timeout, and
+    allowlists to the model. An agent constructed with this sandbox registers them
+    automatically (``Agent(sandbox=sandbox)``).
+
+    .. note::
+        Unlike the base :class:`~strands.sandbox.base.Sandbox` contract, the
+        per-call ``timeout`` argument to :meth:`execute`/:meth:`execute_streaming`
+        (and the code variants) is **ignored**. Strands Shell enforces a single
+        wall-clock timeout configured at construction (the ``timeout`` keyword);
+        set it there to bound command duration.
+    """
+
+    def __init__(
+        self,
+        *,
+        binds: Sequence[dict[str, Any]] | None = None,
+        credentials: Sequence[dict[str, str]] | None = None,
+        allowed_urls: Sequence[str] | None = None,
+        env: dict[str, str] | None = None,
+        umask: int | None = None,
+        timeout: float | None = None,
+        limits: "strands_shell.Limits | None" = None,
+        config_file: str | None = None,
+    ) -> None:
+        """Initialize the sandbox and build the underlying shell.
+
+        Args:
+            binds: Bind mounts exposing host paths in the sandbox. Each is a dict
+                with ``source`` and ``destination`` (both required), an optional
+                ``mode`` (``"direct"`` passthrough, the default, or ``"copy"`` for
+                a build-time snapshot), and an optional ``readonly`` flag.
+            credentials: Per-URL credential injection rules. Each is a dict with a
+                ``url`` and exactly one of ``token`` or ``env_var``. The agent
+                never sees the secret; the kernel injects it per request.
+            allowed_urls: URL prefixes ``curl`` may reach, bypassing the default
+                SSRF guard for those hosts. Listing ``"https://"`` disables SSRF
+                protection entirely — list specific endpoints instead.
+            env: Environment variables to set in the shell.
+            umask: File-creation mask (e.g. ``0o022``).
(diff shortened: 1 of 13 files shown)


agent-of-mkmeral (commented):
## Review — `feat(sandbox): add experimental StrandsShellSandbox (Python + TS)`

Reviewed head `6e0f615`. Cloned the PR, read all 14 files end-to-end, ran the dependency-free worker tests (`test_worker.py` → **6/6 pass**), and traced the implementation against the base `Sandbox` contract and the existing vended-tool factories in both SDKs. The real-shell tests are correctly `skipif`-gated on the optional package, so they don't run in this environment.

**Overall: strong, well-documented PR.** The thread-pinning design for the unsendable native shell is genuinely careful — the `_ShellWorker` + `weakref.finalize` + `del factory` + the "don't capture `self`" comment, backed by `test_dropping_sandbox_terminates_worker_thread`, is exactly the right shape and the regression test is excellent. The VFS-native file-ops detour around the missing `base64` command is a sensible deviation from `PosixShellSandbox`, and staging code to a temp VFS file (instead of shell-escaping) is the right call.

A handful of things I'd want resolved before this ships — the first two are the important ones:

### 🔴 1. Python and TS vend tools under *different names* (and TS contradicts the PR description)
The PR description states: *"Both vend `sandbox_bash` / `sandbox_file_editor`."* That's true for Python (`name="sandbox_bash"` / `"sandbox_file_editor"`, asserted in `test_get_tools_returns_prefixed_bash_and_file_editor`), but **the TS side vends `bash` / `fileEditor`** — it calls `makeFileEditor(this, {description})` / `makeBash(this, {description})` without a `name`, so it falls through to the factory defaults. The TS unit test even bakes this in: `expect(names).toStrictEqual(['fileEditor', 'bash'])`.

Tool names are part of the agent-facing contract. A skill/prompt that references `sandbox_bash` works on Python and silently misses on TS. Pick one (the description says `sandbox_*`) and make both SDKs agree. Inline thread on the TS line.

### 🔴 2. `Bind`/`Cred` native objects are built on the *main* thread, then moved to the worker thread
The entire reason `_ShellWorker` exists is that `strands_shell.Shell` is PyO3-*unsendable* — it panics if touched off its creating thread. But in `__init__`, `bind_objs` and `cred_objs` (`strands_shell.Bind(...)` / `strands_shell.Cred(...)`) are constructed on the **calling thread** and then captured into `_build`, which runs on the **worker thread** and passes them into `Shell(...)`. If `Bind`/`Cred` are also unsendable (very common for PyO3 types that aren't explicitly `Send`), this is a latent interpreter panic that CI won't catch because the integ tests are skipped here. The plain dicts (`self._binds`, `self._credentials`) *are* safe to capture — so the fix is to move the `Bind`/`Cred` construction *inside* `_build`. Inline thread with detail.

### 🟡 3. The vended `sandbox_bash` tool still exposes a `timeout` parameter that is a silent no-op
The class docstring correctly says per-call `timeout` is ignored, but the *tool schema* the model sees (`make_bash` → `timeout: int = 120`; TS `sandboxBashInputSchema` → `timeout`) still advertises it. The model will set `timeout` to bound a long command and be silently surprised when it has no effect. Consider surfacing "per-call timeout is ignored; the sandbox enforces a fixed Ns wall-clock" in the dynamic description suffix (you already build one), so the model isn't misled.

### 🟡 4. `SANDBOX_BASH_DESCRIPTION` tells the model state does *not* persist — but this sandbox is stateful
The shared `SANDBOX_BASH_DESCRIPTION` reads: *"Each call runs in a fresh shell; state such as variables and the working directory does not persist across calls."* For Strands Shell that's **false** — `test_session_state_persists_across_calls` proves `export FOO=bar` survives into the next call (session env/cwd persist; only the explicit `cwd=`/`env=` overrides are scoped via the subshell wrap). The model is being told the opposite of the truth, which will affect how it sequences commands. Since you already append a dynamic suffix, consider also correcting the persistence claim for this backend.

### 🟡 5. `timeout` validation is asymmetric across SDKs
`test_missing_timeout_validation` expects `StrandsShellSandbox(timeout=0)` → `ValueError(match="positive, finite")`, but there's no validation in Python `__init__` — it's delegated to the native `Shell` constructor (surfaced through `_ShellWorker.__init__`). That works for Python, but **TS creates the shell lazily** (`getShell()` on first use), so `new StrandsShellSandbox({timeout: 0})` constructs fine and only blows up on the first `execute()`. Different failure timing for the same misuse. Worth a note or a cheap eager check in TS.

---

### Nits / things I liked
- ✅ `execute_code` write-failure-as-`ExecutionResult` path (`test_execute_code_reports_write_failure_as_result`) + best-effort temp cleanup is solid, and symmetric across PY/TS.
- ✅ `list_files` mapping `FileNotFoundError`/`ENOENT` → `SandboxPathNotFoundError` keeps the file-editor contract intact in both SDKs.
- ✅ Credentials never leak into descriptions (`test_tool_descriptions_surface_urls_and_credentials` asserts `"secret" not in description`) — good defensive test.
- Minor: the dynamic suffix says "Writes outside mounted paths are in-memory only" but doesn't note that writes *inside* a `direct` (non-readonly) bind **do** reach the host. Could be worth one clause so the model understands the host-write surface.

None of these are merge-blockers on the architecture — #1 and #2 are correctness/consistency issues I'd fix first; #3–#5 are model-context/devx polish. Nice work overall.

agent-of-mkmeral (commented):
Two inline threads pinpointing the 🔴 findings from my summary.

strands-ts/src/experimental/sandbox/strands-shell.ts
```diff
@@ -0,0 +1,318 @@
...
+    }
+  }
+
+  // ---- Tools ----
+
+  override getTools(): Tool[] {
+    const suffix = this._dynamicInfo()
+    return [
+      makeFileEditor(this, { description: `${DEFAULT_FILE_EDITOR_DESCRIPTION}${suffix}` }),
+      makeBash(this, { description: `${SANDBOX_BASH_DESCRIPTION}${suffix}` }),
```
agent-of-mkmeral, line 253: 🔴 **Cross-SDK tool-name divergence.** These calls omit `name`, so they fall through to the `makeBash`/`makeFileEditor` defaults `bash` / `fileEditor`. But the Python side uses `name="sandbox_bash"` / `"sandbox_file_editor"`, and the PR description says *both* vend `sandbox_*`. The TS unit test bakes the mismatch in (`expect(names).toStrictEqual(['fileEditor', 'bash'])`). Tool names are an agent-facing contract — a skill referencing `sandbox_bash` works on PY and silently no-ops on TS. Pass `{ name: 'sandbox_bash', ... }` / `{ name: 'sandbox_file_editor', ... }` here (and update the test) so the two SDKs agree.

strands-py/src/strands/experimental/sandbox/strands_shell.py
```diff
@@ -0,0 +1,419 @@
...
+        bind_objs = [
+            strands_shell.Bind(
+                source=b["source"],
+                destination=b["destination"],
+                mode=b.get("mode", "direct"),
+                readonly=b.get("readonly", False),
+            )
+            for b in self._binds
+        ]
+        cred_objs = [
```
agent-of-mkmeral, line 148: 🔴 **Unsendable objects built off the worker thread.** `bind_objs`/`cred_objs` (`strands_shell.Bind(...)` / `Cred(...)`) are constructed here on the *calling* thread, then captured into `_build`, which runs on the dedicated worker thread and feeds them to `Shell(...)`. The whole point of `_ShellWorker` is that the native PyO3 types are unsendable and panic when touched off their creating thread — if `Bind`/`Cred` are unsendable too (typical unless explicitly `Send`), this is a latent interpreter panic that CI won't catch because the integ tests are `skipif`-gated here. The plain dicts (`self._binds`, `self._credentials`) are safe to capture, so move the `Bind`/`Cred` construction *inside* `_build` so every native object is created on the worker thread.

PR #2850, opened 2026-06-17
fix(openai): handle non-string error code in classifyOpenAIError

chaodu-agent:
## Summary

`classifyOpenAIError` crashes with `TypeError: err.code?.toLowerCase is not a function` when used with OpenAI-compatible APIs (e.g., OpenRouter) that return a numeric `code` in error responses.

## Root Cause

The OpenAI SDK's `APIError` class declares `code: string | null | undefined` but assigns it directly from the response body without type coercion (`this.code = data?.['code']`). When an OpenAI-compatible provider like OpenRouter returns `{ "error": { "code": 400, ... } }`, `err.code` is a number at runtime.

`classifyOpenAIError` then calls `err.code?.toLowerCase()` which throws because `Number.prototype` has no `toLowerCase` method.

## Fix

Use a `typeof` guard before calling `.toLowerCase()`:

```typescript
const code = typeof err.code === 'string' ? err.code.toLowerCase() : ''
```

Also widened the type signature to `code?: string | number` to reflect runtime reality.

## Testing

- Added test cases for numeric and null error codes
- All existing tests in `errors.test.ts` continue to pass

## Reproduction

Use the OpenAI model provider with OpenRouter's endpoint (`https://openrouter.ai/api/v1`) and trigger any error — the agent crashes instead of classifying the error properly.

### DIFF
--- a/strands-ts/src/models/openai/errors.ts
+++ b/strands-ts/src/models/openai/errors.ts
@@ -36,8 +36,11 @@ export type OpenAIErrorKind = 'contextOverflow' | 'throttling'
  *
  * @internal
  */
-export function classifyOpenAIError(err: Error & { status?: number; code?: string | number }): OpenAIErrorKind | undefined {
+export function classifyOpenAIError(err: Error & { status?: number; code?: string }): OpenAIErrorKind | undefined {
   const message = err.message?.toLowerCase() ?? ''
+  // The OpenAI SDK types `code` as a string, but OpenAI-compatible providers (e.g. OpenRouter)
+  // can return a number (`{ "error": { "code": 400 } }`). Guard with `typeof` before calling
+  // `.toLowerCase()` so a non-string code falls back to '' instead of throwing a TypeError.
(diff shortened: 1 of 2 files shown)


strands-ts/src/models/openai/errors.ts
```diff
@@ -36,9 +36,9 @@ export type OpenAIErrorKind = 'contextOverflow' | 'throttling'
  *
  * @internal
  */
-export function classifyOpenAIError(err: Error & { status?: number; code?: string }): OpenAIErrorKind | undefined {
+export function classifyOpenAIError(err: Error & { status?: number; code?: string | number }): OpenAIErrorKind | undefined {
   const message = err.message?.toLowerCase() ?? ''
-  const code = err.code?.toLowerCase() ?? ''
+  const code = typeof err.code === 'string' ? err.code.toLowerCase() : ''
```
notowen333, line 44: would be helpful to leave a comment explaining why we're doing this

notowen333 (dismissed):

gautamsirdeshmukh (approved):

PR #2853, opened 2026-06-17
fix: bound tool schema normalization recursion depth

pgrayy:
## Description

`normalize_schema` and `_normalize_property` in `tools/tools.py` are mutually recursive over a tool's `inputSchema` and carry no depth limit, while the `ToolRegistry` guards that wrap them catch only `ValueError`. A sufficiently deeply nested `inputSchema` therefore raises `RecursionError`, which is not a `ValueError` subclass, so it escapes `get_all_tool_specs()`. Because the event loop rebuilds tool specs on every cycle, the failure recurs on each agent invocation rather than being isolated to the offending tool.

This is most relevant for tool specs that originate from an external peer (for example, an MCP server's `inputSchema`), where the schema shape is not under the SDK's control.

This change:

- Adds a `_MAX_SCHEMA_DEPTH` bound (64) to `normalize_schema` / `_normalize_property`. Over-deep schemas now raise `ValueError`, which the registry already logs-and-skips, so a single malformed tool no longer takes down the whole `get_all_tool_specs()` call.
- Broadens the two registry normalization guards from `except ValueError` to `except (ValueError, RecursionError)` as defense-in-depth, so any future unbounded-recursion regression degrades to "skip tool + warn" instead of propagating out of the agent.

64 is well above any realistic hand-authored schema (typically 2–5 object levels deep) and an order of magnitude below CPython's default stack limit, so it rejects pathological input without risking false positives on legitimate tools.

The diff also includes a separate `chore:` commit applying `ruff format` to four `vended_interventions/cedar/` files. These were surfaced by `hatch run prepare` and are unrelated formatting only (no behavior change); happy to split them into their own PR if preferred.

## Note for maintainers

Longer term, it may be worth removing schema normalization altogether rather than hardening it. Model providers already reject invalid schemas with their own errors, so the normalization layer is largely defensive duplication. If the providers are the real source of truth for schema validity, this recursion (and this class of bug) goes away entirely. Raising as a discussion point, not part of this change.

## Type of Change

Bug fix

## Testing

Added two regression tests in `tests/strands/tools/test_tools.py` that build a 100-level nested schema and assert `normalize_schema` and `normalize_tool_spec` raise `ValueError` (not `RecursionError`). Existing tests covering legitimately nested schemas still pass, confirming valid schemas continue to normalize correctly.

## Checklist

----

### DIFF
--- a/strands-py/src/strands/tools/tools.py
+++ b/strands-py/src/strands/tools/tools.py
@@ -23,6 +23,14 @@ _COMPOSITION_KEYWORDS = ("anyOf", "oneOf", "allOf", "not")
 Properties with these should not get a default type: "string" added.
 """
 
+_MAX_SCHEMA_DEPTH = 64
+"""Maximum nesting depth permitted when normalizing a tool inputSchema.
+
+Bounds the mutual recursion between normalize_schema and _normalize_property so that a
+maliciously deep schema (e.g. from an untrusted MCP server) raises ValueError instead of
+exhausting the interpreter stack with RecursionError.
+"""
+
 
 class InvalidToolUseNameException(Exception):
     """Exception raised when a tool use has an invalid name."""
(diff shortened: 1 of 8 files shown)


strands-py/src/strands/tools/tools.py
```diff
@@ -71,21 +79,25 @@ def validate_tool_use_name(tool: ToolUse) -> None:
         raise InvalidToolUseNameException(message)
 
 
-def _normalize_property(prop_name: str, prop_def: Any) -> dict[str, Any]:
+def _normalize_property(prop_name: str, prop_def: Any, *, _depth: int = 0) -> dict[str, Any]:
```
JackYPCOnline, line 82: nit: Do we need * here?

JackYPCOnline (dismissed):

strands-py/src/strands/tools/tools.py
pgrayy: That's just a convention for config args. This make it so you pass in as a named argument.

strands-py/src/strands/tools/tools.py
```diff
@@ -101,18 +113,25 @@ def _normalize_property(prop_name: str, prop_def: Any) -> dict[str, Any]:
...
         schema: The schema to normalize.
+        _depth: Current nesting depth, used to bound recursion into nested object schemas.
 
     Returns:
         The normalized schema.
+
+    Raises:
+        ValueError: If the schema is nested too deeply.
     """
+    if _depth > _MAX_SCHEMA_DEPTH:
```
JackYPCOnline, line 132: Question: do we need input validation here, gurantee positive _depth?

JackYPCOnline (approved):

strands-py/src/strands/tools/tools.py
pgrayy: Certainly wouldn't hurt. I can address as a follow up.

PR #2856, opened 2026-06-17
feat(skills): add pre-push skill mirroring the CI merge gate locally

notowen333:
### Motivation

Contributors (and agents) repeatedly push changes only to have the `ci.yml` merge gate bounce them on something that was reproducible locally — a format miss, a lint error, a type failure. The friction is that reproducing CI by hand means knowing which of the three areas (python / typescript / docs) your change touches and which commands each one runs, then running them from the right working directory. That knowledge lives in the workflow YAML and drifts over time, so people either over-run everything (slow) or guess (misses).

This adds a `pre-push` skill that mirrors the merge gate locally: it detects the affected areas using CI's exact path filters, auto-fixes what's mechanically fixable, then runs only those areas' checks — so a contributor can get push-ready without memorizing the CI layout. It's a manually-invoked skill, not a git hook, so it never runs on every commit.

### What

A bundled `run-checks.sh` does the work, in two phases per detected area:

- **Phase 1 — auto-fix:** formatters, lint `--fix`, and lockfile sync, **scoped to the files the change touches** so it never reformats unrelated files into the commit. Fixers write to the working tree and never fail the run.
- **Phase 2 — check:** the gate (build, lint check, format check, type-check, tests, audit) run tree-wide for CI parity. This phase decides pass/fail.

Detection uses the same `dorny/paths-filter` globs as `ci.yml`. The one deliberate divergence: the `typescript` area is narrower than CI's — it drops `strands-wasm/`, `strands-py-wasm/`, `wit/`, and `strandly/`, because CI validates those via `strandly check --py` (wasm), which this script does not run. Lighting up the area for them would falsely imply coverage.

`SKILL.md` documents the model and, importantly, a **triage policy** for what phase 2 reports: fix mechanical failures autonomously and re-run, but loop back to the developer for anything requiring judgment — above all test failures, where the skill explicitly forbids rewriting a test just to make it pass, since that silences the very check that caught the regression.

### Scope note

A green local run is not a guarantee CI passes: it runs on the current OS and interpreter/node only, so it does not reproduce CI's OS/version matrices (Python 3.10–3.14 × 3 OS; Node 20/22/24 × 3 OS) or the out-of-tree npm-pack smoke test. The skill says to report "local checks passed," not "CI will pass." The TypeScript check tracks CI as of #2841 (which removed the package-lock drift check); the fixer still runs `npm install` to keep the lockfile synced.

## Documentation PR

No `site/` documentation changes. The skill is self-documenting via its `SKILL.md`, and `.agents/skills/README.md` is updated to list it.

## Type of Change

New feature

## Testing

This change is markdown + bash under `.agents/`, so `hatch run prepare` does not apply. `run-checks.sh` is shellcheck-clean. Beyond that it was validated three ways:

**Hermetic test harness (73 cases).** A bash harness drives `run-checks.sh` against a throwaway git repo whose directory shape mimics the monorepo, with `hatch`/`npm`/`npx` replaced by PATH shims that log their invocations and return scripted exit codes. This pins down the script's logic without needing the real toolchains or network, and asserts on:

- **Area detection** — each area lights from the right paths; the shared `ci.yml` lights all three; the `typescript` filter deliberately stays dark for the wasm/`strandly`/`wit` paths it doesn't cover; non-matching changes report no area.
- **Base resolution** — auto-detect, the fork-aware remote scan (`origin/main` vs `upstream/main`), explicit `--base`, and the no-base case (nothing committed yet).
- **File collection** — committed-since-base, staged, unstaged, and untracked files are all considered, including filenames with spaces (the `-z` / `xargs -0r` pipeline).
- **Phase semantics** — fixers never fail the run; a failing check does; areas run independently so every failure surfaces.
- **CLI surface** — `--list`, `--all`, explicit area override, `--heavy` (selects the slow CI-parity commands), unknown-arg rejection, and `-h`.
- **Fixer scoping** — fixers receive only the changed paths, never the whole tree; eslint autofix is restricted to `src/`+`test/integ/`; missing tools are reported, not silently skipped.

**Real end-to-end runs.** The script was run for real against deliberately broken changes in each of the three areas — confirming phase 1 reformats real files on disk and phase 2 invokes the actual gate commands (`hatch fmt`/`hatch test`, the `npm run` targets, `cms:build`) and reports their pass/fail honestly.

The harness is a development tool and is not shipped in this PR. It can be committed under `.agents/skills/pre-push/` as a follow-up if maintainers want the skill to carry its tests.

## Checklist

----


.agents/skills/pre-commit/SKILL.md
```diff
@@ -0,0 +1,102 @@
+---
+name: pre-commit
+description: Runs the local equivalent of the CI merge gate before committing or pushing. Detects which areas (Python, TypeScript, docs) your changes touch, auto-fixes what it can, then runs only those checks. Use when the user asks to run pre-commit checks, get commit-ready, verify changes before pushing, "make sure CI will pass", or check lint/tests/types locally.
```
zastrowm: I'd rather scope this to `pre-push`; I like to commit incrementally and I'm not sure I want this done every time

.agents/skills/pr-writer/SKILL.md
zastrowm, line 1: Should this skill reference the other?

.agents/skills/pre-commit/SKILL.md
notowen333: swapped name and framing

.agents/skills/pr-writer/SKILL.md
notowen333: it does now

zastrowm (approved):

PR #2858, opened 2026-06-17
feat: telemetry for memory manager

opieter-aws:
### Motivation

Observability is a core feature of the SDK, but the memory subsystem was the one major agent subsystem with no OpenTelemetry instrumentation — it emitted only logs. Every other component (agent invocation, event-loop cycle, tool execution, model invocation) flows through the `Tracer` for spans, so memory operations were invisible in traces: you couldn't see how often memory was searched, how slow stores were, how many entries were retrieved/injected/extracted, or why a background extraction failed. Notably, the memory extractor's model call was the only model invocation in either SDK not wrapped in a model-invoke span, leaving extraction cost and latency completely dark.

This change instruments the memory subsystem in **both the Python and TypeScript SDKs** so it matches the rest of the SDK: spans for search, add, context injection, and background extraction, plus wrapping the extraction model call in the existing model-invoke span. Spans nest correctly under the agent/cycle/tool traces.

This PR is **tracing/spans only**. An earlier revision also added `strands.memory.*` metrics; those were dropped after review feedback that the metric set was noisy and the `store_name` dimension risked unbounded cardinality. Metrics can follow in a separate PR once there is a concrete dashboard need — most of what they offered is already derivable from span data.

### Public API Changes

No existing signature on `Agent` or `MemoryManager` changed. The new surface is the span schema exporters receive, and the `Tracer` methods that produce it.

**Span schema** — what shows up in a tracing backend. Span names use a `memory.*` namespace alongside the standard `gen_ai.operation.name`; the two SDKs are kept at exact parity on names, attributes, and events.

| Span | Emitted when | Parenting | Key attributes | Events |
|------|--------------|-----------|----------------|--------|
| `memory.search` | a store query runs (tool or injection) | active agent span | `memory.store.names`, `memory.store.count`, `memory.max_search_results`, `memory.result.count`, `memory.store.failure_count` | `memory.query`, `memory.results` |
| `memory.add` | a write runs | active agent span, or **root** for the fire-and-forget add-tool path | `memory.store.names`, `memory.store.count`, `memory.store.failure_count` | `memory.content` |
| `memory.inject` | context injection runs before a model call | active agent span | `memory.max_entries`, `memory.injected`, `memory.entry.count`, `memory.inject.format_error` | — |
| `memory.extract` | background extraction runs | **root**, linked to scheduling agent span | `memory.store.name`, `memory.message.count`, `memory.message.filtered_count`, `memory.extractor`, `memory.entry.count`, `memory.parent.trace_id`, `memory.parent.span_id` | — |

**New `Tracer` methods** — a start/end pair per operation, following the existing `start*Span`/`end*Span` convention. TypeScript (`strands-ts/src/telemetry/tracer.ts`) is options-object based and returns `Span | null`:

```typescript
startMemorySearchSpan(options: StartMemorySearchSpanOptions): Span | null
endMemorySearchSpan(span: Span | null, options?: EndMemorySearchSpanOptions): void

startMemoryAddSpan(options: StartMemoryAddSpanOptions): Span | null        // options.forceRoot for detached writes
endMemoryAddSpan(span: Span | null, options?: EndMemoryAddSpanOptions): void

startMemoryInjectSpan(options?: StartMemoryInjectSpanOptions): Span | null
endMemoryInjectSpan(span: Span | null, options: EndMemoryInjectSpanOptions): void

startMemoryExtractSpan(options: StartMemoryExtractSpanOptions): Span | null  // options.agentSpanContext attached as a link
endMemoryExtractSpan(span: Span | null, options?: EndMemoryExtractSpanOptions): void

currentSpanContext(): SpanContext | undefined   // @internal — captures the live agent span to link extraction back to it
```

The matching `StartMemory*SpanOptions` / `EndMemory*SpanOptions` interfaces are exported from `strands-ts/src/telemetry/types.ts`. Python (`strands-py/src/strands/telemetry/tracer.py`) mirrors these as keyword-argument methods:

```python
start_memory_search_span(query, store_names, max_search_results=None, *, custom_trace_attributes=None, **kwargs) -> Span
end_memory_search_span(span, entries=None, store_failure_count=0, error=None) -> None

start_memory_add_span(content, store_names, *, force_root=False, custom_trace_attributes=None, **kwargs) -> Span
end_memory_add_span(span, store_failure_count=0, error=None) -> None

start_memory_inject_span(max_entries=None, *, custom_trace_attributes=None, **kwargs) -> Span
end_memory_inject_span(span, injected, entry_count=0, format_error=False) -> None

start_memory_extract_span(store_name, message_count, filtered_count=0, extractor=None, agent_span_context=None, **kwargs) -> Span
end_memory_extract_span(span, entry_count=None, error=None) -> None
```

### Design notes

- **Detached extraction is a root span, not a child.** Automatic extraction runs in a background task (Python `asyncio`, TS promise chain) that typically completes after the agent span has already ended. Parenting to it would orphan the span, so extraction starts its own root trace and carries an OpenTelemetry *link* back to the agent span — captured synchronously while that span is still live. This preserves causal association without an orphaned parent-child relationship.

- **Telemetry never breaks memory.** Span failures are isolated so a misbehaving tracer provider can never turn a successful search/add/extraction into a failure. The detached extraction ends its span best-effort after the save resolves, so a tracing error can't roll back a successful write or trip the store's backoff logic.

- **Injection fails open and never leaks a span.** The injection path ends its span on every branch — no query, no results, a throwing format callback (recorded as a flag, not an error, since the agent did not fail), and an unexpected search error (span ended, then rethrown).

- **Content on span events.** Query and stored/retrieved content are recorded as span events, consistent with how model and tool spans record their inputs. Memory payloads may contain PII; this is documented on the span methods so operators can suppress span content at the exporter when needed.

- **Per-SDK tracer-acquisition idiom.** Python uses the existing global `get_tracer()`. TypeScript's `Tracer` is per-instance, so `MemoryManager` constructs one and shares it with the extraction coordinator; parenting still flows through OTel context, so this does not fragment traces.

## Type of Change

New feature

## Testing

Unit tests in both SDKs cover the span lifecycle for each operation (success, error, and early-return paths), the detached-extraction root-span-with-link behavior, telemetry-failure isolation, fail-open injection, the model-invoke span on the extractor call, and (TS) span-context validity/recording checks. Verified end to end with in-memory span exporters in both SDKs: injection nests a child search span, a tool-invoked search nests under the tool span, and extraction appears as its own root trace linked back to the agent span.

## Checklist

----


strands-py/src/strands/memory/extraction/coordinator.py
```diff
@@ -189,27 +213,80 @@ async def _extract(self, store: MemoryStore) -> None:
...
+        """
+        try:
+            metrics_client = MetricsClient()
+            attributes = {"store_name": store.name}
+            metrics_client.memory_extract_call_count.add(1, attributes=attributes)
+            metrics_client.memory_extract_duration.record(time.time() - start_time, attributes=attributes)
+            if error is None:
+                metrics_client.memory_extract_success_count.add(1, attributes=attributes)
+                metrics_client.memory_extract_entries.record(entry_count, attributes=attributes)
+                get_tracer().end_memory_extract_span(span, entry_count=entry_count)
```
poshinchen, line 243: nit: you can get and reuse the tracer on the top level.

strands-py/src/strands/telemetry/tracer.py
```diff
@@ -795,6 +807,308 @@ def end_swarm_span(
...
+        }
+        if extractor is not None:
+            memory_attributes["memory.extractor"] = extractor
+        attributes = self._build_memory_span_attributes("memory.extract", memory_attributes, extra_kwargs=kwargs)
+
+        links = None
+        if agent_span_context is not None and agent_span_context.is_valid:
+            links = [Link(agent_span_context)]
+
+        return self._start_span("memory.extract", attributes=attributes, force_root=True, links=links)
```
poshinchen, line 1097: What's the reason of forcing a new span? Would closing the previous span and start a new one better?

opieter-aws: Automatic extraction runs in a background `asyncio` task that can finish after the agent span has already ended. There's no previous span we could cleanly close and reopen, since the extraction is triggered by the agent turn but temporally outlives it

strands-py/src/strands/memory/extraction/coordinator.py
opieter-aws: Thanks for the suggestion, will implement!

poshinchen:
I think the otel metrics can be skipped as of now? they might be noisy

opieter-aws:
@poshinchen will scope the PR to tracing/spans only for now, and expand to TS

strands-py/src/strands/memory/extraction/coordinator.py
```diff
@@ -117,28 +124,39 @@ def _done(completed: asyncio.Task) -> None:
...
         """Run this store's save after its previous one completes."""
         if previous is not None:
             await previous
-        await self._extract(store)
+        await self._extract(store, link_context)
+
+    @staticmethod
+    def _current_span_context() -> SpanContext | None:
+        """Capture the current agent span's context for linking, if one is active."""
+        span = trace_api.get_current_span()
```
poshinchen, line 155: The helper method looks good to me, I'm just wondering if this has been implemented in the trace_api or not. So I would double check on that.

strands-py/src/strands/memory/extraction/coordinator.py
```diff
@@ -117,28 +124,39 @@ def _done(completed: asyncio.Task) -> None:
 
         task.add_done_callback(_done)
 
-    def process(self, store: MemoryStore) -> asyncio.Task | None:
+    def process(self, store: MemoryStore, link_context: SpanContext | None = None) -> asyncio.Task | None:
```
poshinchen, line 127: Is there a way (or is it necessary) to have the link_context?

strands-py/src/strands/memory/extraction/coordinator.py
opieter-aws: Checked, and `trace_api` only gives `get_current_span()`, which returns the span. The recording + `is_valid` gating here isn’t vended by otel, so it isn’t duplicate

strands-py/src/strands/memory/extraction/coordinator.py
opieter-aws: It’s necessary for the link to survive. The capture has to happen in `schedule()` while the parent agent span is still live. If `_extract` runs in the background, the span may have ended, so we can’t re-read it there. Threading the captured context down is what lets the detached `extract` trace link back to the parent agent run that triggered it, which I added to be able to correlate it.

poshinchen (approved):

PR #2859, opened 2026-06-17
feat: add local memory store

opieter-aws:
## Description

The memory feature ships only one concrete `MemoryStore`: `BedrockKnowledgeBaseStore`, which requires an AWS account, a provisioned Knowledge Base, and a data source. That gates the *first* experience of memory behind cloud setup — a developer can't try `MemoryManager`, `search_memory`/`add_memory`, context injection, or extraction without standing up infrastructure first.

This adds `LocalMemoryStore`, a zero-infrastructure `MemoryStore` shipped in both SDKs. It **persists to disk by default** (`~/.strands/memory/<name>.json`), so an agent remembers across restarts with no setup, and can be made ephemeral for tests and throwaway runs. `MemoryManager` is backend-agnostic — it talks to stores only through the `MemoryStore` interface — so this is purely additive: drop the store into `stores: [...]` and everything (tools, injection, client-side extraction) works unchanged.

A few design decisions worth calling out for review:

- **One class, not two.** An ephemeral in-memory store and a persistent file store share identical recall logic; the only difference is whether `add` also flushes to disk. Rather than two classes, persistence is a single internal seam (`persist` flag → load-on-init + atomic flush). One construct to learn, one copy of the search code.
- **Persistent by default.** Memory's value proposition *is* remembering across sessions, so the default that demonstrates the feature (survives a restart) is the right one; ephemeral is the opt-out.
- **Lexical recall, not semantic.** v1 ranks by query-token overlap with a recency tiebreak — keyword matching with zero new dependencies. This differs from Bedrock's embedding-based search (a query word matches only the same word, not a synonym), and the docstrings say so explicitly.
- **Subpath export, mirroring `BedrockKnowledgeBaseStore`.** Not added to the top-level barrel.

The store reads the whole file into memory and rewrites it atomically on each `add`, so it suits prototyping and personal memory (hundreds to low thousands of entries), not a production corpus — documented on the class. Writes within a process are serialized so concurrent `add`s can't clobber one another; cross-process writers are not coordinated.

## Public API Changes

A new `LocalMemoryStore`, exported from a subpath in each SDK alongside the existing Bedrock store.

TypeScript (`@strands-agents/sdk/vended-memory-stores/local`):

```typescript
import { LocalMemoryStore } from '@strands-agents/sdk/vended-memory-stores/local'

// Persists to ~/.strands/memory/notes.json by default. Survives restarts.
const agent = new Agent({
  model,
  memoryManager: new MemoryManager({
    stores: [new LocalMemoryStore({ name: 'notes' })],
    addToolConfig: true,
  }),
})

// Ephemeral (tests / throwaway runs): nothing is written to disk.
new LocalMemoryStore({ name: 'notes', persist: false })
// Explicit location:
new LocalMemoryStore({ name: 'notes', path: './my-notes.json' })
```

Python (`strands.vended_memory_stores.local`):

```python
from strands import Agent
from strands.memory.memory_manager import MemoryManager
from strands.vended_memory_stores.local import LocalMemoryStore

agent = Agent(
    model=model,
    memory_manager=MemoryManager(
        stores=[LocalMemoryStore(name="notes")],
        add_tool_config=True,
    ),
)

LocalMemoryStore(name="notes", persist=False)         # ephemeral
LocalMemoryStore(name="notes", path="./my-notes.json")  # explicit path
```

`LocalMemoryStore` defaults `writable=true` (the point is a store you can write to) and implements `search` + `add`. With an `extraction` config it uses the manager's client-side path (model extractor → `add` per fact); it does not implement `addMessages`. Constructor config: `name` (required), `description`, `maxSearchResults`/`max_search_results`, `writable`, `extraction`, `persist` (default `true`), `path`.

## Documentation PR

`team/designs/0011-memory-manager.md` still sketches this store under the name `InMemoryMemoryStore`, imported from the top-level barrel. The shipped store is `LocalMemoryStore` (persists by default, so "InMemory" would mislead) exported via subpath (matching `BedrockKnowledgeBaseStore`). The design doc and any how-to/example docs should be reconciled in a follow-up; no `site/` docs are included here.

## Type of Change

New feature

## Testing

Unit suites added in both SDKs covering: persistence round-trip across a fresh instance ("survives restart"), ephemeral mode (no file written, a new instance forgets), lexical ranking + recency tiebreak, empty/token-less query, empty/non-writable `add` errors, content dedup, `maxSearchResults` cap + `<1` validation, and corrupt/missing-file handling. Each SDK also has MemoryManager integration tests asserting the manager stamps `storeName`, the `add` API writes through, and a client-side extraction run persists the extracted fact.

Beyond the automated suites, I exercised the store end-to-end in both SDKs across **two separate processes**: one writes a fact, a second (fresh process) recalls it from the on-disk JSON — confirming the across-restart behavior, not just in-process mocks. I also confirmed the new subpath export specifier resolves and the TS browser bundle still builds (the store's `node:*` imports are dynamic).

## Checklist

----

### DIFF
--- a/strands-py/src/strands/vended_memory_stores/local/store.py
+++ b/strands-py/src/strands/vended_memory_stores/local/store.py
@@ -59,6 +59,15 @@ def _tokenize(text: str) -> set[str]:
     return {token for token in re.split(r"\W+", text.lower()) if token}
 
 
+def _token_overlap_score(query_tokens: set[str], content: str) -> int:
+    """Lexical relevance score for one record.
+
+    The number of distinct query tokens that appear in the content; a higher count means more of the
+    query's words are present. Returns 0 when there is no overlap.
+    """
+    return len(query_tokens & _tokenize(content))
+
+
 class LocalMemoryStore(MemoryStore):
     """A :class:`~strands.memory.types.MemoryStore` backed by an in-memory list and a local JSON file.
 
@@ -97,7 +106,8 @@ class LocalMemoryStore(MemoryStore):
             **store_config: See :class:`LocalMemoryStoreConfig`.
 
         Raises:
-            ValueError: If ``name`` is empty/whitespace, or ``max_search_results`` is less than 1.
+            ValueError: If ``name`` or ``path`` is empty/whitespace, or ``max_search_results`` is
+                less than 1.
         """
         self.name = store_config["name"]
         if not self.name.strip():
@@ -113,6 +123,8 @@ class LocalMemoryStore(MemoryStore):
 
         self._persist = store_config.get("persist", True)
         path = store_config.get("path")
+        if path is not None and not path.strip():
+            raise ValueError("LocalMemoryStore: path must not be empty.")
         if not self._persist:
             self._path: Path | None = None
         elif path is not None:
@@ -154,7 +166,7 @@ class LocalMemoryStore(MemoryStore):
 
         scored: list[tuple[dict[str, Any], int]] = []
         for record in records:
-            score = len(query_tokens & _tokenize(record["content"]))
+            score = _token_overlap_score(query_tokens, record["content"])
             if score > 0:
                 scored.append((record, score))
 
@@ -184,6 +196,8 @@ class LocalMemoryStore(MemoryStore):
 
         Raises:
             ValueError: If the store is not writable or ``content`` is empty/whitespace.
+            OSError: If persisting the entry to disk fails (e.g. the path is unreachable or not
+                writable), with the target path in the message.
         """
         if not self.writable:
             raise ValueError("LocalMemoryStore: store is not writable. Set writable=True in config to enable add().")
@@ -230,6 +244,8 @@ class LocalMemoryStore(MemoryStore):
             return self._records
         except json.JSONDecodeError as error:
             raise ValueError(f"LocalMemoryStore: invalid JSON in {self._path}: {error}") from error
+        except OSError as error:
+            raise OSError(f"LocalMemoryStore: failed to read {self._path}: {error}") from error
 
         if not isinstance(parsed_file, list):
             raise ValueError(f"LocalMemoryStore: invalid backing file {self._path}: expected a JSON array of records")
@@ -251,12 +267,19 @@ class LocalMemoryStore(MemoryStore):
         """Persist ``records`` with an atomic write (write to a ``.tmp`` file, then replace).
 
         A crash mid-write can never leave a partially written file. A no-op for ephemeral stores.
+
+        Raises:
+            OSError: If the backing directory cannot be created or the file cannot be written
+                (e.g. the path is unreachable or not writable), with the target path in the message.
         """
         if self._path is None:
             return
 
-        self._path.parent.mkdir(parents=True, exist_ok=True)
-        tmp_path = self._path.with_name(f"{self._path.name}.tmp")
-        with open(tmp_path, "w", encoding="utf-8", newline="\n") as file:
-            json.dump(records, file, indent=2, ensure_ascii=False)
-        tmp_path.replace(self._path)
+        try:
+            self._path.parent.mkdir(parents=True, exist_ok=True)
+            tmp_path = self._path.with_name(f"{self._path.name}.tmp")
+            with open(tmp_path, "w", encoding="utf-8", newline="\n") as file:
+                json.dump(records, file, indent=2, ensure_ascii=False)
+            tmp_path.replace(self._path)
+        except OSError as error:
+            raise OSError(f"LocalMemoryStore: failed to write {self._path}: {error}") from error
--- a/strands-py/src/strands/vended_memory_stores/local/store.py
+++ b/strands-py/src/strands/vended_memory_stores/local/store.py
@@ -1,0 +1,262 @@
+"""A :class:`~strands.memory.types.MemoryStore` that persists to a local JSON file.
+
+A zero-infrastructure store for prototyping and testing. It persists to disk by default so memories persist
+across sessions, and can be set to ephemeral for testing.
+"""
+
+from __future__ import annotations
+
+import json
+import re
+import threading
+import uuid
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any
+
+from typing_extensions import Unpack
+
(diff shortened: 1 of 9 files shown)


strands-py/src/strands/vended_memory_stores/local/store.py
```diff
@@ -0,0 +1,274 @@
...
+        """Load records from disk on first use; ephemeral stores (and a missing file) start empty."""
+        if self._records is not None:
+            return self._records
+
+        if self._path is None:
+            self._records = []
+            return self._records
+
+        try:
+            with open(self._path, encoding="utf-8") as file:
```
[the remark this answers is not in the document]
opieter-aws: This is intentional: the memory manager takes care of fire-and-forget mechanism. For search for injection, this adds >10 ms with less than 10.000 records. This is acceptable for prototyping and testing

strands-ts/src/vended-memory-stores/local/store.ts
```diff
@@ -0,0 +1,326 @@
...
+  /** The id of the stored record. */
+  id: string
+}
+
+/**
+ * Sanitizes a store name into a safe single-path-segment filename.
+ * Guards the default-path branch against a name that would escape the memory directory.
+ * Ensures cross-SDK consistent sanitization.
+ */
+function sanitizeName(name: string): string {
```
notowen333, line 57: Maybe could just make this part of the name validation upfront and skip sanitization

strands-ts/src/vended-memory-stores/local/store.ts
```diff
@@ -0,0 +1,326 @@
...
+    .replace(/[/\\]/g, '_')
+    .replace(/[^\w\-.]/g, '_')
+}
+
+/**
+ * Lowercases and splits text into a set of word tokens, dropping empties. Splits on any run of
+ * characters that are not Unicode letters, numbers, or underscore. Ensures cross-SDK consistent
+ * tokenization.
+ */
+function tokenize(text: string): Set<string> {
```
notowen333, line 69: maybe there is a dep which would be more robust than this?

strands-ts/src/vended-memory-stores/local/store.ts
```diff
@@ -0,0 +1,326 @@
...
+    }
+    const limit = options?.maxSearchResults || this.maxSearchResults || DEFAULT_MAX_SEARCH_RESULTS
+
+    const queryTokens = tokenize(query)
+    if (queryTokens.size === 0) return []
+
+    const records = await this._load()
+
+    const scored: Array<{ record: LocalMemoryRecord; score: number }> = []
+    for (const record of records) {
```
notowen333, line 186: Maybe extract this algo into a function?

Also might be worth a light experiment comparing some iterations of what we have here?

strands-ts/src/vended-memory-stores/local/store.ts
```diff
@@ -0,0 +1,326 @@
...
+  /**
+   * Resolves (and caches) the backing-file path: the explicit `path` from config, else
+   * `~/.strands/memory/<sanitized-store-name>.json`. Returns `undefined` for ephemeral stores. The
+   * `node:os`/`node:path` imports are dynamic so the module stays safe to bundle for the browser.
+   */
+  private async _getPath(): Promise<string | undefined> {
+    if (!this._persist) return undefined
+    if (this._resolvedPath !== undefined) return this._resolvedPath
+    if (this._explicitPath !== undefined) {
+      this._resolvedPath = this._explicitPath
```
notowen333, line 256: should we validate explicit paths?

strands-ts/src/vended-memory-stores/local/store.ts
```diff
@@ -0,0 +1,326 @@
...
+      this._records = await this._loadPromise
+    } finally {
+      this._loadPromise = undefined
+    }
+    return this._records
+  }
+
+  /** Reads and parses the backing file (or returns an empty list when ephemeral / missing). */
+  private async _readFromDisk(): Promise<LocalMemoryRecord[]> {
+    const filePath = await this._getPath()
```
notowen333, line 285: should path be just resolved once at construction?

strands-ts/src/vended-memory-stores/local/store.ts
```diff
@@ -0,0 +1,326 @@
...
+    } catch (error: unknown) {
+      if ((error as { code?: string }).code === 'ENOENT') return []
+      throw error
+    }
+
+    let parsedFile: unknown
+    try {
+      parsedFile = JSON.parse(rawContent)
+    } catch (error: unknown) {
+      throw new Error(`LocalMemoryStore: invalid JSON in ${filePath}`, { cause: error })
```
notowen333, line 301: is JSON the best fit?

Would something like an append only txt file work for our use case (separate entities and remember what is first)

notowen333 (commented):
Just commenting on TS

strands-ts/src/vended-memory-stores/local/store.ts
opieter-aws: They serve a different purpose: validation rejects empty names, sanitization stops a name like `"../../.."` escaping the memory directory when constructing the path. 

We don't want to rejecting spaces or accents in the actual store name

strands-ts/src/vended-memory-stores/local/store.ts
opieter-aws: There's not much quality gain in using external dependencies to tokenize. At the cost of adding dependencies (and cross-SDK), it's not worth it considering the use case of this local store.

strands-ts/src/vended-memory-stores/local/store.ts
opieter-aws: It's cached in `_resolvedPath`

strands-ts/src/vended-memory-stores/local/store.ts
opieter-aws: JSON does seem like the easiest solution here. We can customize this and support other formats, but it would add a lot of generalization to these functions, for a prototyping store, and I'm not sure that users will need that level of control.

We can always extent if there's an ask for it.

notowen333 (approved):

Issue #1764, opened 2026-02-25
[BUG] missing thought signature preservation for thinking models with LiteLLM Model Provider


heltzan:
### Checks

- [x] I have updated to the lastest minor and patch version of Strands
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched [./issues](./issues?q=) and there are no duplicates of my issue

### Strands Version

1.27.0

### Python Version

3.13

### Operating System

Windows 11 version 23H2

### Installation Method

pip

### Steps to Reproduce

1. Create a strands agent using a Litellm model and at least one tool
2. use a prompt where the agent will need to use one of the tools

ex:
`from strands import Agent
from strands_tools import calculator, current_time
from strands.models.litellm import LiteLLMModel

ASSISTANT_PROMPT = f"""You are a helpful assistant."""

model = LiteLLMModel(
    client_args={
        "api_key": '',
        "base_url": '',
    },
    model_id='litellm_proxy/gemini-3-flash'
)

assistant = Agent(
    system_prompt=ASSISTANT_PROMPT,
    model=model,
    tools=[calculator, current_time]
)

response = assistant("what time is it")`

### Expected Behavior

Agent responds with the time

### Actual Behavior

`BadRequestError: litellm.BadRequestError: Vertex_ai_betaException BadRequestError - b'{\\n  \"error\": {\\n    \"code\": 400,\\n    \"message\": \"Unable to submit request because function call `current_time` in the 2. content block is missing a `thought_signature`. Learn more: [https://docs.cloud.google.com/vertex-ai/generative-ai/docs/thought-signatures\](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/thought-signatures/)",\\n    \"status\": \"INVALID_ARGUMENT\"\\n  }\\n}\\n'", "type": null, "param": null, "code": "503"}}`

PR #2861, opened 2026-06-17
fix(test): make gemini thinking-model tool-call integ test robust

yonib05:
## Description

`tests_integ/models/test_model_litellm.py::test_gemini_thinking_model_tool_call` is the regression test for #1764 (Gemini thinking models must preserve `thought_signature` across multi-turn tool calls). It was flaky in CI.

The test asked an open-ended question and asserted the model's free-text reply literally contained `"12:00"` and `"sunny"`:

```python
result = agent("What is the time and weather in New York?")
text = result.message["content"][0]["text"].lower()
assert all(string in text for string in ["12:00", "sunny"])
```

Running against the live `gemini/gemini-2.5-flash` model showed two failure modes:
1. Gemini sometimes **declined to call the tools at all**, reasoning it could only report its own location given the "New York" framing.
2. Even when it did call them, its final **wording varied**, so the substring check failed.

### Fix

Direct the model to use the tools, then assert on the **structure of the multi-turn cycle** instead of the model's free-text wording:

- both tools (`tool_time`, `tool_weather`) were invoked;
- both tool results succeeded (`status == "success"`) and carry the fixtures' fixed `"12:00"` / `"sunny"` return values — these are tool *outputs*, not model phrasing, so they're deterministic;
- a final assistant text turn followed the tool results.

That final post-tool-result turn is exactly what `thought_signature` preservation enables (without it, the follow-up model call after the tool results would fail), so this preserves the original regression coverage without depending on non-deterministic output. The `status == "success"` check also catches a signature loss that surfaces as an error tool result rather than a raised exception.

This narrows but does not eliminate the inherent nondeterminism of a live-API test (it still depends on the model choosing to call both tools); the directive prompt makes that reliable in practice.

## Related Issues

Regression test for #1764.

## Type of Change

Other: test robustness fix

## Testing

Verified against the live `gemini/gemini-2.5-flash` model: the new prompt invoked both tools on 6/6 probe runs, and the test passed on repeated runs. `ruff check` and `ruff format --check` pass. (CI exercises this test with the real `GOOGLE_API_KEY`.)

## Checklist
- [x] I have added any necessary tests that prove my fix is effective

### DIFF
--- a/strands-py/tests_integ/models/test_model_litellm.py
+++ b/strands-py/tests_integ/models/test_model_litellm.py
@@ -8,6 +8,7 @@ import pytest
 import strands
 from strands import Agent
 from strands.models.litellm import LiteLLMModel
+from tests_integ.conftest import retry_on_flaky
 
 
 @pytest.fixture
@@ -280,6 +281,7 @@ async def test_cache_read_tokens_multi_turn(model):
     assert result.metrics.accumulated_usage["cacheWriteInputTokens"] > 0
 
 
+@retry_on_flaky("The model may occasionally not elect to call both tools", retry_on=[AssertionError])
 def test_gemini_thinking_model_tool_call(tools):
     """Test that Gemini thinking models preserve thought_signature through multi-turn tool calls.
 
--- a/strands-py/tests_integ/models/test_model_litellm.py
+++ b/strands-py/tests_integ/models/test_model_litellm.py
@@ -287,6 +287,33 @@ def test_gemini_thinking_model_tool_call(tools):
     """
     model = LiteLLMModel(model_id="gemini/gemini-2.5-flash", client_args={"api_key": os.environ.get("GOOGLE_API_KEY")})
     agent = Agent(model=model, tools=tools)
-    result = agent("What is the time and weather in New York?")
-    text = result.message["content"][0]["text"].lower()
-    assert all(string in text for string in ["12:00", "sunny"])
+    # Direct the model to use the tools: an open-ended question lets Gemini decline
+    # (e.g. "I can only report my own location"), which is unrelated to what we test here.
+    result = agent("Use the available tools to look up the current time and weather, then tell me both.")
+
+    # Assert on the structure of the multi-turn cycle rather than the model's free-text
+    # wording (non-deterministic). Without thought_signature preservation the follow-up
+    # model call after the tool results would fail, so reaching a final assistant text
+    # turn that incorporates both tool results is what proves the regression is fixed.
+    called_tools = {
+        block["toolUse"]["name"] for message in agent.messages for block in message["content"] if "toolUse" in block
+    }
+    assert {"tool_time", "tool_weather"} <= called_tools
+
+    tool_results = [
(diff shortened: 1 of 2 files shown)


yonib05:
Thanks for the review. Addressed the note about the `status == "success"` comment in 601cb44c — the old wording tied it to thought_signature loss, but that failure surfaces at the follow-up model call, not in tool execution. Reworded the comment to describe what the assertion actually guards (both tools ran cleanly so the follow-up call operates on real results rather than error placeholders).

strands-py/tests_integ/models/test_model_litellm.py
```diff
@@ -287,6 +287,33 @@ def test_gemini_thinking_model_tool_call(tools):
...
+        content_block["text"]
+        for tool_result in tool_results
+        for content_block in tool_result["content"]
+        if "text" in content_block
+    )
+    assert "12:00" in tool_result_text and "sunny" in tool_result_text
+
+    # The agent loop ran to completion with a final assistant turn after the tool results.
+    assert result.message["role"] == "assistant"
+    assert any("text" in block for block in result.message["content"])
```
opieter-aws, line 321: Are the additional assertions actually making the test more robust here? 

Wouldn't we achieve better robustness by just widening the tool result assertions to include e.g. `noon`, `12` (so `12 PM` would match too) and `sun`? Would removing the location here not result in the LLM responding with `Where are you located?`?

I'd combine that with a retry on assertion error

yonib05: Regarding the tool robustness:
The tool_result_text check runs against toolResult content, the tool functions' return values, which are hardcoded fixtures (tool_time() returns exactly "12:00", tool_weather() returns "sunny"). 
Those are deterministic, so "12:00"/"sunny" always match when the tools are called. Fuzzy-matching noon/12 PM only makes sense for model free-text.

Valid, however, in my testing the new prompt against live Gemini 2.5-flash, it invoked both tools on 6/6 probe runs and the full test passed 3/3. The decline path was actually triggered by the old prompt's "New York" framing response (“I can only report my own location”)


Added the retry flaky on this test

yonib05: I took another look. A more robust change would be to give the tool a `city` parameter, which it currently doesn't have — the tools here have no description or args at all (`tool_weather() -> str`), so the "I can only report my own location" response was really the model reasoning about an argument-less tool against a location-specific prompt. Adding `city` would let the prompt map cleanly onto the tool and help the model's reasoning, probably as reliably as the prompt change did.

The reason I didn't is that this `tools` fixture is shared by `test_agent_invoke`, `test_agent_invoke_async`, and `test_agent_stream_async`, and those haven't been flaky — so I wanted to limit the scope here rather than perturb passing tests. Happy to do the fixture change as a separate follow-up if you'd prefer.

opieter-aws: I see, thanks for elaboration! Let's merge and see if the error still bites us or no

opieter-aws (approved):

PR #2870, opened 2026-06-18
chore: update dependencies and refactor dev tooling internals

yonib05:
## Summary

A set of small dependency and internal-tooling updates.

### Dependency updates
- Bump `aws-cdk-lib` to `2.260.0` in `test-infra` and the four CDK deployment examples (`deploy_to_apprunner`, `deploy_to_ec2`, `deploy_to_fargate`, `deploy_to_lambda`), refreshing the bundled transitive packages.
- Bump `esbuild` to `0.28.1` in the four CDK deployment example lockfiles.

### Internal tooling and tests
- `strands-ts` context-offloader `S3Storage`: replace the trailing-slash-stripping regex on the configured prefix with a linear string scan. Behavior is unchanged; covered by the existing storage unit tests.
- `strandly` CLI: run the venv `pip`/`python` commands through `execFileSync` with an argument vector instead of building a shell string. The argument vectors tokenize identically to the previous command strings.
- TypeScript API docs generator: encode the literal `<name>Data` prose marker as an HTML entity. This serializes to the same MDX output as before (verified against the generator's `escapeMdxChars` pipeline).
- OpenAI Responses integration test: match the citation host by parsed hostname instead of a substring check.

## Testing
- `strands-ts` storage unit tests: 104/104 pass.
- `strandly` and `strands-ts` type-check clean; ESLint and Prettier clean on the changed TypeScript files.
- `test-infra` lockfile verified with `npm ci`; the CDK example lockfile edits verified coherent and against the npm-generated `test-infra` lockfile.
- Python test file compiles and passes `ruff`.
- Verified the docs generator change produces identical rendered MDX by running it through the actual `escapeMdxChars` remark pipeline.

### DIFF
--- a/strandly/src/cli.ts
+++ b/strandly/src/cli.ts
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-import { execSync } from 'node:child_process'
+import { execFileSync, execSync } from 'node:child_process'
 import { existsSync, readdirSync, readFileSync, writeFileSync } from 'node:fs'
 import { join, resolve } from 'node:path'
 import { program } from 'commander'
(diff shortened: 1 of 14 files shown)


strandly/src/cli.ts
```diff
@@ -168,6 +168,24 @@ function run(cmd: string, opts?: { cwd?: string; env?: Record<string, string> })
   }
 }
 
+/** Run a binary directly with an argument vector, bypassing the shell. Used for
```
zastrowm, line 171: > with an argument vector

has got to be the oddest wording I've seen for this

zastrowm (approved):

PR #2874, opened 2026-06-19
ci(python): add timeouts and apt retries to test-lint workflow

yonib05:
## Summary

The Python test/lint jobs in `python-test-lint.yml` have no job-level timeout. When a setup step stalls, the job runs until GitHub's 6-hour default before it's killed. This has happened more than once on the Linux **"Install system audio dependencies"** step, where `sudo apt-get update`/`install` hangs (frozen on apt with no timeout or retry) — a single matrix leg sat for 4+ hours while every other leg finished in minutes.

## Changes

- **Job-level `timeout-minutes`** — 25 for the unit-test matrix, 20 for lint. Sized from observed run data: legs normally finish in ~2-4 min, slowest observed ~16 min, so these leave generous headroom (>50% over the max) while capping a hung job at minutes instead of ~6 hours.
- **Resilient apt step** — wrap the `apt-get` commands in a 3-attempt retry loop with a 5-minute step timeout, set `DEBIAN_FRONTEND=noninteractive`, and pass `-o Acquire::Retries=3`. Transient mirror/network blips now recover automatically; a genuinely broken repo still fails loudly after 3 attempts.
- **macOS** — add a 5-minute step timeout to the `brew install portaudio` step (same hang class).

## Behavior

On the happy path the outcome is unchanged: both packages install and the step succeeds. The retry/timeout only changes what happens on a stall or transient failure — fast, bounded failure (and self-healing) instead of an open-ended hang.

## Validation

- YAML parses; the retry loop was checked under `bash -eo pipefail` (the runner's default shell): failed attempts fall through and retry, success exits 0, and it exits 1 after 3 failures. The `if cmd1 && cmd2` guard is exempt from `set -e`, so the loop is not aborted early.

### DIFF
--- a/.github/workflows/python-test-lint.yml
+++ b/.github/workflows/python-test-lint.yml
@@ -87,8 +87,10 @@ jobs:
               && sudo apt-get -o Acquire::Retries=3 install -y portaudio19-dev libasound2-dev; then
               exit 0
             fi
-            echo "apt-get attempt $attempt failed; retrying in 15s..."
-            sleep 15
+            if [ "$attempt" -lt 3 ]; then
+              echo "apt-get attempt $attempt failed; retrying in 15s..."
+              sleep 15
+            fi
           done
           echo "apt-get failed after 3 attempts" >&2
           exit 1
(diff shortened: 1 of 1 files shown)


yonib05:
Thanks for the review. Addressed in 67c2af37:

- **Final-attempt sleep** (line 94): guarded the retry message + 15s sleep with `if [ "$attempt" -lt 3 ]`, so a genuine failure now exits immediately without the misleading "retrying" line or dead wait.
- **macOS brew** (line 100): wrapped `brew install portaudio` in the same 3-attempt retry as Linux for consistency, so a transient brew/network blip self-heals.

On the **composite-action extraction** (line 158): leaving as-is for now, in line with your "fine to leave as-is, follow up if a third caller appears" note. With only two call sites I'd rather not introduce the repo's first `.github/actions/` in a CI-hardening change; happy to extract it as a follow-up if a third consumer shows up.

JackYPCOnline (approved):

PR #2883, opened 2026-06-20
fix: filter graph dependency reasoning blocks

### LINKED ISSUE #2882: Graph forwards reasoning blocks into downstream node input
## Summary

`Graph` pass reasoning content from an upstream node into a downstream node input. I wanted to check whether this is expected, because it breaks on Bedrock and also seems surprising from a graph-composition perspective.

In my case, one node uses a reasoning-capable model, then a second node depends on its output. The second node receives the first node full content, including `reasoningBlock` content, as part of its input. Bedrock then rejects the request because reasoning content is present on a user message.

## What I am seeing

With a simple two-node graph:

1. Node A uses a reasoning-capable Bedrock model.
2. Node B depends on Node A.
3. Node A returns content that includes a reasoning block.
4. Graph dependency resolution forwards that content into Node B.
5. Node B request fails because the downstream user message contains reasoning content.

The Bedrock error is:

```text
User messages cannot contain reasoning content
```

Looking at SDK `1.6.0`, this seems to happen because graph dependency input resolution forwards dependency content directly:

```ts
deps.push(new TextBlock(`[node: ${edge.source.id}]`), ...ns.content);
```

## Minimal repro shape

```ts
const graph = new Graph({
  nodes: [
    { id: "reasoner", executor: reasoningAgent },
    { id: "curator", executor: normalAgent },
  ],
  edges: [{ source: "reasoner", target: "curator" }],
});

await graph.run("Introduce yourself and explain your strengths.");
```

Where `reasoningAgent` uses a reasoning-capable Bedrock model, such as DeepSeek R1 or a Claude thinking model.

## Question

Is this forwarding behavior intentional?

If it is, is there a recommended way to configure graph handoff so downstream nodes receive only the upstream node final/output content?

If it is not intentional, would filtering `reasoningBlock` content during graph dependency input resolution be the right fix?

kingrubic:
## Description

Graph dependency resolution currently forwards every upstream content block into downstream node input. When the upstream node returns reasoning content, that reasoning block can become part of the downstream user message and providers such as Bedrock reject it.

This PR filters reasoning blocks out of dependency handoff content while preserving regular dependency text output and node markers.

## Related Issues

Resolves #2882

## Documentation PR

Not needed; this preserves the existing Graph handoff behavior for user-visible content and removes provider-incompatible reasoning content from the handoff.

## Type of Change

Bug fix

## Testing

- [x] `npm test -w strands-ts -- graph.test.ts`
- [x] `npm run build -w strands-ts`
- [x] `npx eslint src/multiagent/graph.ts src/multiagent/__tests__/graph.test.ts --max-warnings=0` from `strands-ts/`
- [x] Pre-commit ran `npm run test:coverage -w strands-ts` successfully: 128 files passed, 3498 tests passed, no type errors. The repo-wide lint step reported one pre-existing warning in `src/agent/__tests__/agent.context-manager.test.ts`, which this PR does not touch.

## Checklist

----

### DIFF
--- a/strands-ts/src/multiagent/__tests__/graph.test.ts
+++ b/strands-ts/src/multiagent/__tests__/graph.test.ts
@@ -5,7 +5,7 @@ import { MockSnapshotStorage } from '../../__fixtures__/mock-storage-provider.js
 import { collectGenerator } from '../../__fixtures__/model-test-helpers.js'
 import { createCancellableAgent } from '../../__fixtures__/agent-helpers.js'
 import { AfterNodeCallEvent, BeforeNodeCallEvent, MultiAgentInitializedEvent } from '../events.js'
-import { TextBlock, type ContentBlockData } from '../../types/messages.js'
+import { ReasoningBlock, TextBlock, type ContentBlockData } from '../../types/messages.js'
 import { Status, MultiAgentState } from '../state.js'
 import { AgentNode, MultiAgentNode } from '../nodes.js'
 import { Graph } from '../graph.js'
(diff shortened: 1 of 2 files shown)


chaynabors (approved):
Same filter and guard as my #2956 plus a regression test that asserts both the text and type values, which catches the exact bug. Closing mine in favor of this one.

chaynabors (approved):

Issue #1436, opened 2026-01-08
[FEATURE] Support for Direct Method Call Streaming


larrylp2:
### Problem Statement

Strands supports direct method calls (https://strandsagents.com/latest/documentation/docs/user-guide/concepts/tools/#direct-method-calls), which enables developers to invoke specific tools directly. 

However, direct method calls currently do not support streaming results for developers. This means that there is a disjointed experience for the same Strands agent, where stream_async enables streaming responses to users for E2E Agent calls but are not possible for direct method calls.

### Proposed Solution

To address this, the `tools/_caller.py` can be extended to also have an async version. 

The current direct tool call method seems to use a streaming-friendly ToolExecutor call behind the scenes, but handles it synchronously before returning the full tool result. 

https://github.com/strands-agents/sdk-python/blob/main/src/strands/tools/_caller.py#L89

### Use Case

The use case would be a seamless streaming experience, even if an Agent built with Strands incorporates direct tool calls in certain flows behind the scenes. 

A multi-agent app following the Agents as Tools (https://strandsagents.com/latest/documentation/docs/user-guide/concepts/multi-agent/agents-as-tools/) pattern that relies on streamed tool events (https://strandsagents.com/latest/documentation/docs/user-guide/concepts/streaming/#tool-events) can still stream results through direct method calls.

emaan-c:
# Direct Tool Call Streaming API

**Status**: Proposed

**Date**: 2026-03-22

**Issue**: https://github.com/strands-agents/sdk-python/issues/1436

## Context

Direct tool calls (`agent.tool.my_tool()`) currently block without providing streaming events, while agent-level calls (`agent.stream_async()`) expose rich streaming events from the same `ToolExecutor._stream()` infrastructure. This creates inconsistent developer experience and prevents building responsive UIs for long-running operations, multi-agent systems, and debugging workflows.

**Who experiences this:** Developers building multi-agent systems (Agents-as-Tools pattern), applications with long-running tools (database queries, API calls), and responsive UIs requiring real-time progress feedback.

## Decision

Add streaming methods to tool call objects by returning a callable `_ToolExecutor` wrapper instead of raw functions from `_ToolCaller.__getattr__`:

```python
class _ToolExecutor:
    def __call__(self, **kwargs) -> ToolResult:
        """Sync execution (existing behavior - backward compatible)"""
    
    def stream(self, **kwargs) -> Iterator[TypedEvent]:
        """Sync streaming (NEW)"""
    
[...]

PR #2886, opened 2026-06-21
docs: design proposal for direct method call streaming (#1436)

Ashutosh0x:
## Summary
Design proposal for adding streaming support to direct method tool calls (Agent.tool.my_tool()).

## Related Issue
Addresses #1436

## Problem
Currently, gent.tool.my_tool() blocks until completion. The internal ToolExecutor._stream() yields streaming events, but they are consumed and discarded by _ToolCaller. This prevents developers from building responsive UIs or debugging long-running tools.

## Proposed Solution
Add an optional stream=True parameter to direct tool calls:

`python
# Blocking (existing, unchanged)
result = agent.tool.my_tool(param='value')

# Streaming (new)
async for event in agent.tool.my_tool(param='value', stream=True):
    print(event)
``n
## Design Details
- Follows the official Strands design document template
- Evaluates 2 alternatives (parameter-based vs separate accessor)
- No breaking changes
- Backward compatible

## Willingness to Implement
Yes. I am willing to implement this feature if the design is approved.

### DIFF
--- a/team/designs/1436-direct-method-call-streaming.md
+++ b/team/designs/1436-direct-method-call-streaming.md
@@ -1,0 +1,221 @@
+# Direct Method Call Streaming
+
+**Status**: Proposed
+
+**Date**: 2026-06-21
+
+**Issue**: https://github.com/strands-agents/sdk-python/issues/1436
+
+## Context
+
+When using Strands Agents, developers can invoke tools directly via the method-style interface:
+
+```python
+agent = Agent(tools=[my_tool])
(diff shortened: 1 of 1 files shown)


zastrowm:
We actually already did this for TS, so I'd follow that design: https://github.com/strands-agents/harness-sdk/blob/145eb62fa0573b1fdea48bab8f9339786f4a1ae9/strands-ts/src/agent/tool-caller.ts#L152

PR #2894, opened 2026-06-22
feat(skills): add system_prompt injection mode to AgentSkills

agent-of-mkmeral:
## Why

The `AgentSkills` plugin gives skills to an agent through *progressive disclosure*: skill metadata (name + description) is injected into the system prompt, and the agent calls the `skills` tool to pull a skill's full instructions into context as a tool result. That works well when many skills are available and you want to keep the context small.

But for long-running, multi-turn workflows, instructions delivered as a one-off tool result tend to drift out of the model's attention as the conversation grows — the agent "forgets" the skill it activated several turns ago. For those cases it is better to keep the activated skill's instructions resident in the system prompt, where they stay salient every turn, and to let the agent explicitly toggle skills on and off.

This PR adds that second delivery strategy without changing the existing default.

### 1. `ContextInjector` learns a `location` (the core primitive)

Per the PR discussion (@lizradway / @opieter-aws), system-prompt delivery is built into the **context-injection primitive** rather than as skills-specific machinery. `ContextInjector` (and the underlying injection middleware) gains a keyword-only `location` option:

```python
from strands.vended_plugins.context_injector import ContextInjector

# Default — unchanged: fold into the latest user message
ContextInjector(render)

# New — append to the per-call system prompt
ContextInjector(render, trigger="everyTurn", location="systemPrompt")
```

- `"lastUserMessage"` (default): per-ask context (retrieved memories, timestamps) that belongs next to the user's message.
- `"systemPrompt"` (new): standing guidance (skill instructions, policies) that belongs with the agent's directives.

Both locations are **ephemeral**: the model sees the augmented input for that one call while the agent's durable system prompt and history are never touched. For structured prompts (block lists) the text is appended *after* existing blocks — including any `cachePoint` — so the static prefix stays cacheable. The new `InjectionLocation` type is exported from `strands.injection` and re-exported by the plugin package.

### 2. `AgentSkills` gains `injection_mode`, implemented on top of it

```python
from strands import Agent
from strands.vended_plugins.skills import AgentSkills

# Default — unchanged behaviour: metadata in prompt, instructions via the `skills` tool
agent = Agent(plugins=[AgentSkills(skills=["./skills/"])])

# New — activated skills' full instructions live in the per-call system prompt
agent = Agent(plugins=[AgentSkills(skills=["./skills/"], injection_mode="system_prompt")])
```

Both modes expose exactly **one tool, named `skills`** — only its signature differs by mode:

- `"tool"` mode (default, unchanged): `skills(skill_name)` activates a skill and returns its full instructions as the tool result.
- `"system_prompt"` mode (new): `skills(action, skill_name)` where `action` is `"activate"` or `"deactivate"` (an enum in the tool schema). Activating puts the skill's full instructions in the system prompt where they persist across turns; deactivating removes them to reclaim context.

In `system_prompt` mode the plugin instantiates an **internal `ContextInjector`** (`trigger="everyTurn"`, `location="systemPrompt"`) that renders the skills block from the agent's *current* activation state on every model call:

```xml
<available_skills>
<usage>Activate a skill with the skills tool (action="activate") to load its full instructions into this section; action="deactivate" removes them. Skills marked active="true" are in effect - follow their instructions.</usage>
<skill active="true">
<name>pdf-extract</name>
<description>Extract text and tables from PDF files</description>
<location>/skills/pdf-extract/SKILL.md</location>
<instructions>
...full skill body, verbatim except neutralized closing delimiters...
</instructions>
</skill>
<skill>
<name>csv-clean</name>
<description>Normalize messy CSV files</description>
<location>/skills/csv-clean/SKILL.md</location>
</skill>
</available_skills>
```

Only activated skills get `active="true"` and the inlined `<instructions>` element; inactive skills stay metadata-only. Instruction bodies are injected **verbatim** (matching what `tool` mode returns) — only closing `</instructions>` / `</available_skills>` sequences (case/whitespace variants included) are neutralized so a body cannot break out of the enclosing elements.

**Because delivery is ephemeral and re-rendered per call, a whole class of state management disappears:** `agent.system_prompt` is never modified in this mode, so the `managed-by` marker attribute, the sentinel sweep, and the remove-and-reappend desync handling from earlier revisions are gone (~90 lines deleted). Toggles take effect on the very next model call (`everyTurn` covers tool-result turns), multi-agent isolation and session/state round-trips fall out for free, and there is nothing durable to accumulate or desync. `tool` mode is untouched — its metadata block is still injected durably, byte-identical to previous releases.

Activation state lives in `agent.state`, so it persists across runs through the session manager and one plugin instance is safe to share across agents. Programmatic equivalents (`activate_skill_for` / `deactivate_skill_for`) are available. The `SkillInjectionMode` type alias is exported alongside the existing public names. `injection_mode` is keyword-only.

## Scope

Python only, as requested — the TypeScript `AgentSkills`/`ContextInjector` are at parity today and a matching change there is a sensible follow-up.

## Testing

Clean-venv run: ruff format + ruff lint clean; **mypy `./src` clean**; `tests/strands/vended_plugins/` + `tests/strands/injection/` — **494 passed** (1,294 across injection + vended_plugins + event_loop + agent). New coverage includes: `_fold_into_system_prompt` (None/str/block-list, immutability, cache-point ordering), `location="systemPrompt"` middleware behaviour (trigger gating, everyTurn on tool-result turns, default unchanged), `ContextInjector(location=...)` wiring, and the skills-mode suite reworked to assert through the per-call injection path (durable prompt never modified, no middleware registered in tool mode, toggle round-trips, verbatim fidelity, delimiter neutralization, multi-agent isolation, session state round-trip).

### Review round at `fa16120` (addressed in `b3e9b92`)

- **Injected-token accounting**: the injection middleware now folds the injected text's estimated tokens into `projected_input_tokens` on the middleware context (both locations), so downstream consumers see the true request size; a deterministic warning is logged when the corrected projection exceeds the model's context window limit. Counting failures fail open.
- **Subclass tools preserved**: mode selection removes only the inactive built-in `skills` variant instead of allowlisting one method — `@tool` methods on `AgentSkills` subclasses survive in both modes.
- **`allowed_tools` parity**: active skills in `system_prompt` mode render an `<allowed_tools>` element matching the guidance `tool` mode returns.
- **mypy**: the three reported errors fixed; `mypy ./src` clean.
- **Docs**: `injection_mode` documented on the Skills page, `location` on the Context Injector page (both marked Python-only), and `PORTING.md`'s `ContextInjector` signature updated.

Resolves agent-of-mkmeral/strands-coder-private#315.

cc @mkmeral


strands-py/src/strands/vended_plugins/skills/agent_skills.py
```diff
@@ -184,20 +232,73 @@ async def skills(self, skill_name: str, tool_context: ToolContext) -> str:
         self._track_activated_skill(agent, skill_name)
         return await self._format_skill_response(found, agent.sandbox)
 
+    @tool(context=True)
+    async def activate_skill(self, skill_name: str, tool_context: ToolContext) -> str:
```
mkmeral: can we have single skills tool with actions for activating and deactivating instead?

strandly-the-agent (commented):
## Review: `injection_mode` for `AgentSkills` — ✅ approve with comments (no blocker)

Clean, well-scoped, backward-compatible addition. The default `"tool"` path is genuinely untouched and the risky parts (re-injection idempotency, cachePoint preservation, tool exposure per mode) are each backed by a test. No 🔴. The items below are worth weighing before release — the **escape fidelity** one being the one I'd most want resolved, since it bites exactly the long-running workloads this feature targets.

**Verified (reviewed `f67d13f`, file `strands-py/src/strands/vended_plugins/skills/agent_skills.py`):**
- ✅ Tool filtering is real — in `"tool"` mode `activate_skill`/`deactivate_skill` are not registered with the model (`agent_skills.py:190`, `_tools_for_mode`).
- ✅ No accumulation: `_inject_skills` removes the prior block by value-equality before re-appending; structured-prompt path preserves cachePoints (matches the new tests).
- ✅ Default `"tool"` mode byte-identical to base; `_on_before_model_call` early-returns there.
- ⚠️ **Could not run the suite** — `strands` isn't installable in my sandbox, so all findings are static (traced to `file:line`), not test-executed. Worth a human running `tests/strands/vended_plugins/`.

<details>
<summary><b>🟡 Should-fix / worth-a-decision (6)</b></summary>

**1. Escaping corrupts skill instructions, and diverges from `"tool"` mode.** `system_prompt` mode injects `escape(skill.instructions)` (`agent_skills.py:677`), but `"tool"` mode delivers the body **raw** (`agent_skills.py:522`). Any skill whose instructions contain code, `<`, `&`, or `&&` reaches the model as HTML entities (`a&lt;b`, `x&amp;&amp;y`) — but only in this mode. Same skill, two different readings, and a real fidelity loss for instruction bodies that contain code (the common case). The `<`/`>` escaping is only needed to stop a body containing `</instructions>` from breaking out — consider a fenced/CDATA-style delimiter (or escaping only the closing sequence) so code survives intact and both modes match. *(independently flagged by the adversarial and LLM-context passes)*

**2. The injected block never tells the model it can activate skills.** The `<available_skills>` block (`agent_skills.py:668-679`) lists skills, but nothing in it signals the activate/deactivate mechanism or that an `active="true"` skill's `<instructions>` are directives to follow now. The tool docstrings point *to* the block, but the block never points *back* to the tools — so reliable discovery depends on the model reading the tool specs. A one-line header/footer would close it, e.g.:
> `Activate a skill with the activate_skill tool to load its full instructions into this section; deactivate_skill removes them. Skills marked active="true" are in effect — follow their instructions.`

**3. (Q) Resident instructions land *after* the last user cachePoint.** `_inject_skills` always appends the skills block at the end (`agent_skills.py:362`), and in `system_prompt` mode it's re-injected before every model call (`_on_before_model_call`, `agent_skills.py:317-331`). The prefix stays cached, but the largest new payload sits past the last cache point and is reprocessed uncached every turn — on exactly the long-running conversations this feature is for. Worth emitting a `cachePoint` after the skills block so a stable active-set gets cached (each activate/deactivate would invalidate it — a fair trade)?

**4. Multi-agent isolation is untested in `system_prompt` mode.** The docstring promises one plugin instance safely serves multiple agents (`agent_skills.py:337-338`), but every new test uses a single agent. A regression that keyed activation on the plugin instead of the agent would pass the whole suite while leaking one agent's instructions into another's prompt. Suggest a test: share one `system_prompt`-mode plugin across agents A and B, activate on A, assert the instructions appear in `A.system_prompt` and **not** in `B`.

**5. Session round-trip of `activated_skills` is untested.** The docstring claims activations persist across runs via the session manager (`agent_skills.py:687-688`) — the feature's headline benefit — but no test pre-seeds `state` on a *fresh* plugin and asserts the instructions reappear on the next model call. The in-process dict round-trips trivially; the "fresh plugin honours pre-existing state" path is the one that matters and is uncovered.

**6. (Q, conditional) Block-removal relies on two independently-persisted locations staying in sync.** Removal matches `state[...]['last_injected_xml']` against the block living in `system_prompt`. Snapshot/restore can persist `state` and `system_prompt` separately (some include/exclude combos), after which removal silently fails and a duplicate block accumulates (`[base, A, A]`). Re-deriving removal from a stable sentinel tag/attribute (dovetails with the delimiter change in #1) would make this robust; at minimum, document that `state` + `system_prompt` must be snapshotted together. Lower confidence — depends on snapshot config.
</details>

<details>
<summary><b>⚪ Nits — docstring & message accuracy (bundle)</b></summary>

- `set_available_skills` docstring (`agent_skills.py:405`) says active state is "reconciled on next tool call or invocation," but there's no pruning code — a skill removed from the available set stays in `activated_skills` and is silently never rendered. Soften the wording or add pruning during injection.
- `activate_skill` returns *"Its instructions are now in your system prompt"* (`agent_skills.py:262`), but injection is deferred to the next `BeforeModelCallEvent`. Functionally invisible to the model, but the wording is slightly ahead of reality.
- `activate_skill` reports success even when the skill has no instructions (success message at `:262` vs the `is_active and skill.instructions` guard at `:676`); `"tool"` mode handles this honestly ("no instructions available").
- `activate_skill_for` in `"tool"` mode records activation state with no observable prompt effect — worth a docstring note that it's a recorded no-op there.
- `skill_name` param descriptions (`:244`, `:278`) could point at the source of valid values ("the `<name>` from the available_skills section" / "must match a currently-active skill").
</details>

<details>
<summary><b>❓ Questions for the author (design — non-blocking)</b></summary>

1. **`_for` naming symmetry.** The new agent-scoped methods are `activate_skill_for` / `deactivate_skill_for` (`agent_skills.py:682`/`:696`) — the suffix avoids colliding with the `@tool` `activate_skill` — but the existing agent-scoped `get_activated_skills` (`:790`) has no suffix, so the trio reads inconsistently. Deliberate convention, or worth aligning before release?
2. **`injection_mode` keyword-only?** It's the last positional-or-keyword param (`:150`); inserting `*,` before it is non-breaking today and matches the forward-compat stance, though it'd make the signature mixed (the existing params are positional-or-keyword). Intentional to keep it positional?
3. **Module docstring tail.** The module docstring ends mid-sentence ("…URL skills are") — pre-existing on `main`, but this PR rewrites the lines just above it, so it's cheap to finish or drop the dangling clause while you're here.

*Note: this review was produced by an automated multi-pass agent and is offered as input for a human reviewer, not a gate. Findings are static (traced to `file:line`); I couldn't execute the test suite in my sandbox.*

strands-py/src/strands/vended_plugins/skills/agent_skills.py
```diff
@@ -482,6 +616,99 @@ def _generate_skills_xml(self, agent: Agent | None = None) -> str:
...
+                otherwise uses only the base skills.
+
+        Returns:
+            The text block to inject into the system prompt.
+        """
+        if self._injection_mode != "system_prompt":
+            return self._generate_skills_xml(agent)
+        return self._generate_skills_xml_with_active(agent)
+
+    def _generate_skills_xml_with_active(self, agent: Agent | None) -> str:
```
mkmeral: can we dedupe code?

strands-py/src/strands/vended_plugins/skills/agent_skills.py
agent-of-mkmeral: Done in b930910 — `activate_skill`/`deactivate_skill` are gone. In `system_prompt` mode the plugin now exposes a single tool named `skills` with `action: "activate" | "deactivate"` (rendered as an enum in the tool schema) plus `skill_name`. Both modes now surface exactly one tool named `skills`; the mode-appropriate variant is selected at construction (tool mode keeps its original activate-and-return signature — spec unchanged). Unknown actions return a graceful error, and the programmatic `activate_skill_for` / `deactivate_skill_for` methods are unchanged.

strands-py/src/strands/vended_plugins/skills/agent_skills.py
agent-of-mkmeral: Done in b930910 — `_generate_skills_xml_with_active` and `_generate_injection_text` are gone. There's now a single `_generate_skills_xml` that renders activation state (`active="true"` + `<instructions>`) only in `system_prompt` mode, so tool-mode output stays byte-identical. Merging the two toggle tools into one method also removed the duplicated validation blocks. Net: −130/+111 lines with 2 extra tests.

mkmeral:
@strandly-the-agent  review pls

strandly-the-agent:
## Re-review of `b930910` — ✅ both requests cleanly addressed, no new issues

The refactor does exactly the two things asked and nothing else — I verified it by running the suite and exercising the code this time (my earlier review couldn't install `strands`). Clean to proceed from the refactor's standpoint. The only thing I'd still weigh before merge is the **escaping** item carried over from my earlier review — see below.

**Verified (checked out `b930910`, branch `feat/skills-system-prompt-injection-mode`):**
- ✅ **Single `skills` tool per mode.** `tool` mode → `skills(skill_name)` (unchanged); `system_prompt` mode → `skills(action, skill_name)` with `action` an `activate|deactivate` enum in the schema. Confirmed exactly one tool named `skills` is exposed in each mode.
- ✅ **`__name__`-based tool filtering is safe.** The `Plugin.tools` property returns the filtered `self._tools`, so only the mode's variant registers — no risk of both `skills` and `manage_skills` leaking.
- ✅ **XML dedup is behavior-preserving.** One `_generate_skills_xml`; rendered tool-mode output is byte-identical to base (no `active=`/`<instructions>`).
- ✅ **`tests/strands/vended_plugins/skills/` → 166 passed.**
- ✅ **Independent adversarial pass over the `f67d13f..b930910` diff:** no unintended behavior change / regression.

<details>
<summary><b>Still open from my earlier review (not touched by this refactor)</b></summary>

This commit only scoped @mkmeral's two comments, so the findings from my review on `f67d13f` (above) remain open. Re-flagging just the one I'd resolve before release; the rest are in that review and I won't repeat them:

- 🟡 **Escaping still mangles instruction bodies** (`agent_skills.py:627`). `system_prompt` mode injects `escape(skill.instructions)`, so a skill whose body contains code / `<` / `&` / `&&` reaches the model as HTML entities (`a&lt;b`, `x&amp;&amp;y`) — while `tool` mode delivers the body raw. Same skill, two readings, and a real fidelity loss for the long-running workloads this mode targets. The `<`/`>` escaping only exists to stop a body containing `</instructions>` from breaking out — a fenced/CDATA-style delimiter (or escaping only the closing sequence) would fix both the fidelity gap and the divergence between modes.

The other 5 should-fix/nit items (discovery header in the injected block, `cachePoint` after the skills block, multi-agent-isolation test, session round-trip test, docstring accuracy) also remain — see the earlier review rather than a re-list here.
</details>

*Automated multi-pass re-review, offered as input for a human — not a gate. The refactor itself is verified; the open item above is a design call for you.*

agent-of-mkmeral:
@strandly-the-agent's review is now addressed too (commit `7c8474fb`) — apologies for skipping past it in the last round, that was a miss on my part. Point-by-point:

## 🟡 Should-fix items

**1. Escaping corrupts instruction bodies** ✅ **Fixed.** Activated skill instructions are now injected **verbatim** — matching what `tool` mode returns — instead of `escape()`d. Only the two closing sequences that could break out of the enclosing elements (`</instructions>`, `</available_skills>`) are neutralized (`<\/…>`), exactly the "escape only the closing sequence" option you suggested. Code with `<`, `>`, `&`, `&&` reaches the model unmodified, and both modes now deliver the same bytes. Covered by `test_active_instructions_injected_verbatim` + `test_instruction_closing_delimiters_neutralized`.

**2. Block never advertises the activation mechanism** ✅ **Fixed.** In `system_prompt` mode the block now opens with a `<usage>` element: *"Activate a skill with the skills tool (action=\"activate\") to load its full instructions into this section; action=\"deactivate\" removes them. Skills marked active=\"true\" are in effect - follow their instructions."* Gated on the mode, so tool-mode output stays byte-identical (`test_usage_hint_present_in_system_prompt_mode_only`).

**3. (Q) Resident instructions after the last cachePoint** ➡️ **Answered, not implemented — deferring to @mkmeral.** Deliberately left as-is: emitting a cachePoint from the plugin would (a) consume one of Bedrock's limited cache-block slots without the app's consent, and (b) invalidate on every toggle *and* on every metadata change. The block *is* stable between toggles, so an opt-in `cache_skills_block=True` flag is a clean follow-up if wanted — happy to add it here if you (mkmeral) prefer.

**4. Multi-agent isolation untested** ✅ **Fixed.** `test_multi_agent_isolation`: one shared plugin, agents A + B, activate on A → instructions in A's prompt, absent from B's (metadata listing still present in both).

**5. Session round-trip untested** ✅ **Fixed.** `test_session_restored_state_reinjects_instructions`: pre-seeds `activated_skills` into `agent.state` on a *fresh* plugin (simulating session-manager restore) and asserts the instructions re-appear with `active="true"` on the next injection.

**6. (Q) Removal relies on state + prompt staying in sync** ✅ **Fixed (the robust option).** Removal now falls back to a **sentinel sweep**: after the exact-`last_injected_xml` match, any leftover block matching `<available_skills>…</available_skills>` is removed before re-appending (both the block path and the string path). Safe because #1's delimiter neutralization guarantees instruction bodies can't contain the closing sentinel — the dovetail you predicted. `test_state_prompt_desync_does_not_accumulate` simulates the desync and asserts exactly one block survives.

## ⚪ Nits — all taken
- `set_available_skills` docstring no longer claims reconciliation ("…is not pruned: an activated name that no longer matches an available skill simply stops being rendered").
- Activation message now says instructions "will appear in your system prompt from your next turn onward" (accurate to the `BeforeModelCallEvent` timing).
- Activating a skill with no instructions now honestly returns "activated (no instructions available)", matching tool mode (`test_activate_skill_without_instructions_honest_message`).
- `activate_skill_for` docstring notes it's a recorded no-op on the prompt in `"tool"` mode.
- Both `skill_name` param descriptions point at the source of valid values (`<name>` element / currently-active skill).

## ❓ Questions
1. **`_for` naming symmetry** — deliberate: the suffix exists purely to dodge the `@tool` name collision, and `get_activated_skills` has no collision to dodge. Renaming it to `get_activated_skills_for` would be a breaking change to a shipped API for symmetry alone, so I left it; happy to align in a follow-up if @mkmeral wants.
2. **`injection_mode` keyword-only** — agreed, done: `*,` inserted before it. Non-breaking (it's new in this PR) and keeps the door open for reordering.
3. **Dangling module docstring** — finished the sentence.

**Verification:** full `tests/strands/vended_plugins/` suite → **379 passed** (+7 new), ruff format + lint clean.

mkmeral:
@strandly-the-agent we good?

strandly-the-agent:
## ✅ We're good — mergeable

I re-verified against the **live head** (`ad89186`, which includes the `main` merge) rather than taking the checklist on faith. All 6 findings + nits + questions from my `f67d13f` review are genuinely addressed in `7c8474fb`, and the suite is green.

**Verified by running it:**
- ✅ **Escaping (#1):** instructions now injected **verbatim** — code with `<`, `>`, `&`, `&&` reaches the model unmangled; only `</instructions>` / `</available_skills>` are neutralized. Both modes now deliver identical bytes.
- ✅ **Usage hint (#2)** in `system_prompt` mode only; tool-mode output still byte-identical. **Multi-agent isolation (#4), session round-trip (#5), desync no-accumulation (#6)** — new tests present & passing. **keyword-only `injection_mode` (Q2)** + docstring/message nits — all in.
- ✅ **`tests/strands/vended_plugins/` → 426 passed** (172 in `skills/`). Deferring the cachePoint idea (#3) to an opt-in follow-up is a reasonable call.

Nothing blocking — ship it. A fresh adversarial pass over the *new* neutralization/sweep code did turn up two low-severity edge cases if you feel like tightening them (neither gates the merge):

<details>
<summary><b>Two non-blocking edge cases in the new breakout-guard code</b></summary>

**A. `_neutralize_delimiters` misses whitespace/case-variant closers** (`agent_skills.py:73-88`). It replaces only the exact literals `</instructions>` / `</available_skills>`, but `</instructions >` (trailing space), `</instructions\t>`, and `</INSTRUCTIONS>` slip through — and a lenient parser/model treats `</instructions >` as a valid close, so a skill body could still break out of its element. Low exploitability (skill bodies are author-controlled), but it's a hole in a function whose whole purpose is breakout prevention. One-liner hardens it to match the stated contract:
```python
_CLOSE_RE = re.compile(r"</\s*(instructions|available_skills)\s*>", re.IGNORECASE)
def _neutralize_delimiters(text: str) -> str:
    return _CLOSE_RE.sub(lambda m: f"<\\/{m.group(1)}>", text)
```

**B. Sentinel sweep can delete a user's *own* `<available_skills>` block** (`agent_skills.py:423`, `_is_skills_block`). The sweep unconditionally drops any system-prompt block that starts with `<available_skills>` and ends with `</available_skills>` — including one the app authored itself (silent, no warning). Very niche (it's the plugin's namespace), but it's a small regression vs the old exact-match removal. Tagging the plugin's own block with a private marker/attribute and matching on that would keep the desync safety without the false positive.

**Minor:** the plain-string branch in `_inject_skills` (and its `_SKILLS_BLOCK_RE`) is effectively unreachable — `split_system_prompt` turns a string prompt into `[{"text": …}]`, so `system_prompt_content` is only `None` when there's no prompt at all. Harmless, but that regex sweep never runs on real content.
</details>

*Verified on `ad89186` by running the suite + direct repros. The three items above are polish, not gates — your call whether to fold them in now or as a follow-up. Automated review, offered as input for a human, not a gate.*

agent-of-mkmeral:
@strandly-the-agent's three post-approval polish items are now folded in too (commit `79a33bb`):

**A. Variant closers slip past `_neutralize_delimiters`** ✅ **Fixed** — exactly the regex you suggested: a case-insensitive `_CLOSING_DELIMITER_RE = re.compile(r"</\s*(instructions|available_skills)\s*>", re.IGNORECASE)` now defangs `</instructions >`, `</instructions\t>`, `</INSTRUCTIONS>`, etc. (original tag case preserved in the neutralized output). Covered by `test_variant_closing_delimiters_neutralized`.

**B. Sentinel sweep could delete a user-authored `<available_skills>` block** ✅ **Fixed** — the plugin now stamps its own blocks with a private marker attribute (`<available_skills managed-by="strands-agent-skills">`, **`system_prompt` mode only**), and `_is_skills_block` matches only marker-tagged blocks. An app-authored `<available_skills>` block survives the sweep untouched (`test_sweep_preserves_user_authored_available_skills_block`). Tool-mode blocks stay unmarked — keeping that mode's prompt output byte-identical to previous releases — so a desync there falls back to the pre-existing warn-and-re-append behaviour, i.e. exactly `main`'s semantics.

**C. Unreachable plain-string branch** ✅ **Removed** — since `split_system_prompt()` returns `(None, None)` only for a `None` prompt, the branch now simply makes the skills block the entire prompt (with a comment explaining why), and the dead `_SKILLS_BLOCK_RE` regex sweep is gone. The two old string-path tests were replaced with ones that test the real reachable behaviour (no-prompt injection + no-accumulation across the None→string→block-path transition).

**Verification:** full `tests/strands/vended_plugins/` → **428 passed** (skills: 174, +2 net new), ruff format + lint clean. Tool-mode byte-identity tests untouched and green.

mkmeral:
@strandly-the-agent  review pls :)

strandly-the-agent:
## ✅ All three polish items landed — clean, nothing outstanding

Re-verified against the live head (`3ee2588`, includes today's `main` merge). `79a33bb` folds in exactly the three items from my last pass, and I confirmed each by running it:

- ✅ **A — variant closers:** `_neutralize_delimiters` now uses the case-insensitive regex — `</instructions >`, `</instructions\t>`, `</INSTRUCTIONS>`, `</Available_Skills >` all get defanged, while code bodies (`a<b && x>y`) stay verbatim.
- ✅ **B — marker-scoped sweep:** `system_prompt` blocks open with `<available_skills managed-by="strands-agent-skills">` and `_is_skills_block` matches only marked blocks — a user-authored `<available_skills>` block survives the sweep, and **tool-mode output stays byte-identical** (no marker).
- ✅ **C — dead branch removed:** the unreachable string-path sweep is gone; the no-prompt path just makes the block the whole prompt, with reachable tests replacing the old ones.
- ✅ **`tests/strands/vended_plugins/` → 441 passed** (174 in `skills/`), including the 4 new tests (`test_variant_closing_delimiters_neutralized`, `test_sweep_preserves_user_authored_available_skills_block`, `test_no_prompt_becomes_skills_block`, `test_no_prompt_reinjection_does_not_accumulate`).

No remaining findings from me — good to merge. 🚀

*Verified on `3ee2588` by running the suite + direct repros. Automated review, offered as input for a human, not a gate.*

mkmeral (dismissed):

lizradway:
~~When should we use this versus Context Injector? Are we asserting that context injector is only for wrapped middleware? As a dev I would expect this to be under ContextInjector tbh, but would be curious to see what direction philosophically this is headed and if we should integrate /dedupe somehow cc: @opieter-aws~~

nvm i think i misunderstood scope, would be very interested to see how context injection and skills can grow together though.

mkmeral:
> would be very interested to see how context injection and skills can grow together though.

I was also thinking the same. given that we have a previously developed approach, I did not want to change too much, but I think/agree Context injector should be used as a core primitive in these types of components

lizradway:
> I was also thinking the same. given that we have a previously developed approach, I did not want to change too much, but I think/agree Context injector should be used as a core primitive in these types of components


Is there a deadline to get this out? Otherwise, it might be nice to add the sys prompt injection into the Context Injector and use that internally. I suppose that is implementation detail though that we can change later

opieter-aws:
Agreed, let's extend ContextInjector with an injection location option (last user message or system prompt), that'd be a nice addition!

agent-of-mkmeral:
## Reworked per the thread: system-prompt delivery now lives in the injection primitive (`ce371ef`)

@lizradway @opieter-aws @mkmeral — done exactly as discussed: `ContextInjector` is extended with an **injection location** option, and `AgentSkills`' `system_prompt` mode now uses it internally instead of mutating `agent.system_prompt`.

### What changed

**Primitive** (`strands.injection`):
- New `InjectionLocation = Literal["lastUserMessage", "systemPrompt"]` (exported).
- `_create_injection_middleware(..., location=...)` + `_fold_into_system_prompt`: appends text to the per-call system prompt — `None` → text becomes the prompt, `str` → blank-line append, block list → new list with a trailing text block (existing blocks incl. `cachePoint` untouched, so the static prefix stays cacheable). Input never mutated; fail-open semantics unchanged.
- `ContextInjector(render, *, trigger=..., location=...)` — keyword-only, defaults to `"lastUserMessage"`, fully backward compatible.

**AgentSkills `system_prompt` mode**: now constructs an internal `ContextInjector(trigger="everyTurn", location="systemPrompt")` that re-renders the skills block from the *current* activation state on every model call. Consequences:
- Delivery is **ephemeral** — `agent.system_prompt` is never modified in this mode.
- The `managed-by` marker, sentinel sweep, and remove-and-reappend desync machinery from earlier revisions are **deleted** (~90 lines): with nothing durable written, there is nothing to sweep or desync. Multi-agent isolation and session round-trips fall out for free.
- Toggles still land on the very next turn (`everyTurn` fires on tool-result turns too).
- `tool` mode untouched: durable metadata injection, byte-identical output, and it registers no middleware at all.

### Pre-PR self-review (as requested)
- ✅ Diff walked end-to-end; no leftover references to the removed machinery (`BeforeModelCallEvent` hook, `_is_skills_block`, marker constants — grep-clean).
- ✅ `tests/strands/vended_plugins/` + `tests/strands/injection/`: **478 passed** in a clean venv.
- ✅ ruff format + ruff check clean on all 9 touched files.
- ✅ mypy on touched sources: no new errors vs. the pre-change baseline (baseline had 6 pre-existing findings under standalone invocation; head has 4).
- ✅ Cache-safety double-checked: system-prompt injection *appends after* any `cachePoint`, so structured-prompt prefixes remain cacheable; the injected tail varies without invalidating the prefix.
- ⚖️ One deliberate call: in `system_prompt` mode the block (incl. the `<usage>` hint) is injected on **every** call even with zero active skills — same information the durable approach carried, just delivered ephemerally, and it keeps discovery/toggle affordances always visible. Trivial to gate on "any skills exist" if preferred.

PR body updated to describe the new architecture. TS parity follow-up (same `location` option on the TS ContextInjector) is straightforward whenever wanted.

opieter-aws:
@mkmeral do you want this reviewed or https://github.com/strands-agents/harness-sdk/pull/3284 ?

mkmeral:
@strandly-the-agent review pls

strands-py/src/strands/injection/_message_injection.py
```diff
@@ -90,6 +95,8 @@ async def handler(context: InvokeModelContext) -> InvokeModelContext:
         if text is None or not text.strip():
             return context
 
+        if location == "systemPrompt":
+            return replace(context, system_prompt=_fold_into_system_prompt(context.system_prompt, text))
```
strandly-the-agent: 🔴 **Block remains at `b3e9b928`: account and enforce the transformed request before token-dependent consumers run.** The new helper updates `InvokeModelContext.projected_input_tokens`, but input middleware runs after `BeforeModelCallEvent`, so proactive compression has already used the stale estimate; agentic context status can also run before the injector. Crossing the corrected limit only logs and still calls the provider.

A cold full-agent harness repeated four times identically: generic injection had `BeforeModelCallEvent=[1]` but sent `5001` tokens; Skills had `[447]` but sent `5554`; both made zero proactive reductions and still called the provider. Agentic status reported `2803` for an actual `3832`. On a warm turn, the prior provider usage already included the recurring injection, then this helper added it again (`final=[1002, 2005]`, actual `[1002, 1005]`). Reactive recovery retried twice before failing with only one durable message left.

Please compute the complete ephemeral request projection before compression/status/enforcement decisions, without carrying the prior turn's ephemeral contribution into the next baseline. This crosses event-loop and middleware ordering, so a localized suggestion block here would be misleading.

strands-py/src/strands/vended_plugins/skills/agent_skills.py
```diff
@@ -140,6 +220,30 @@ def __init__(
...
         self._agent_skills: weakref.WeakKeyDictionary[Agent, dict[str, Skill]] = weakref.WeakKeyDictionary()
         super().__init__()
+        # Both modes surface a single tool named ``skills``; only the variant matching the
+        # active injection mode is exposed. ``tool`` mode uses the activate-and-return variant
+        # (the ``skills`` method); ``system_prompt`` mode uses the action-based toggle variant
+        # (the ``manage_skills`` method, published under the ``skills`` tool name). The variants
+        # share a tool name, so filtering is by the defining method's name.
+        wanted_method = self._tool_method_for_mode()
+        self._tools = [t for t in self._tools if t.__name__ == wanted_method]
+        # In system_prompt mode the skills block reaches the model through the injection
```
strandly-the-agent, line 233: ✅ **Verified resolved at `b3e9b928`.** The filter now removes only the inactive built-in variant, preserving unrelated decorated subclass tools. I independently reproduced `inspect_skill_state` alongside the mode-appropriate `skills` tool in both modes, and the new regression tests pass.

strands-py/src/strands/vended_plugins/skills/agent_skills.py
```diff
@@ -469,19 +676,74 @@ def _generate_skills_xml(self, agent: Agent | None = None) -> str:
...
-            lines.append("<skill>")
+            is_active = skill.name in active
+            lines.append('<skill active="true">' if is_active else "<skill>")
             lines.append(f"<name>{escape(skill.name)}</name>")
             lines.append(f"<description>{escape(skill.description)}</description>")
             if skill.path is not None:
                 lines.append(f"<location>{escape(str(skill.path / 'SKILL.md'))}</location>")
+            if is_active and skill.instructions:
+                lines.append(f"<instructions>\n{_neutralize_delimiters(skill.instructions)}\n</instructions>")
             lines.append("</skill>")
```
strandly-the-agent, line 713: ✅ **Verified resolved at `b3e9b928`.** Active skills now render escaped `<allowed_tools>` guidance immediately before `<instructions>`; it is omitted for inactive/empty cases, and tool-mode output remains unchanged. The focused regression tests pass.

strandly-the-agent (changes requested):
🔴 **Changes requested** at `fa16120e`: two reproduced regressions and one model-context parity gap remain.
✅ 146 focused tests pass; `git diff --check` passes; both cold repros were repeated twice identically.
🔴 Context-window accounting undercounts injected prompts; mode selection also deletes subclass tools.
🔴 Python lint fails with mypy errors at `agent_skills.py:229,358,680` (CI reports the same failure).
🏷 This adds public APIs but has neither `api/needs-review` nor `api/review-complete`; please complete that workflow before merge.

<details>
<summary>Per-pass review breakdown, verification, and questions</summary>

### Context build
- Reviewed exact head `fa16120e49bc587279afbf2c2df94da7f30a5148` against merge-base `8658b9e0d57427b77b69343262184d76d8e63e8f`.
- Traced the new `ContextInjector(location=...)` path through input middleware, event-loop token projection/provider handoff, `AgentSkills` state, plugin tool discovery, canonical docs, and prior review threads. Earlier resolved delimiter feedback was not re-raised.

### Correctness / safety
- **Blocker:** request-bound system-prompt injection runs after projected-token accounting. Cold repros measured 2 projected vs 2,102 sent tokens for generic injection and 448 vs 2,343 for an active skill; injected-only overflow could not be repaired by trimming durable history.
- **Blocker:** mode selection filters the complete discovered tool collection, deleting an unrelated subclass `@tool` in both modes.
- Python lint fails with mypy errors at `agent_skills.py:229,358,680`; recorded here instead of adding a redundant inline comment for visible CI output.

### API / DevX
- The public layering through `ContextInjector` is coherent, but the subclass-tool regression violates the documented plugin extension contract.
- This public surface requires the repository's `api/needs-review` → `api/review-complete` workflow; neither label is present.
- **Question:** how should this action schema and activation state reconcile with concurrent #3284 before API review completes?
- The advanced API tier was unavailable under the configured data-retention mode; this pass completed on the default tier.

### Adversarial / repro
- Repeated the token-accounting and subclass-tool cold repros twice in independent processes with identical results.
- Tested clean: string/structured prompt folding, cache-point retention, middleware ordering/fail-open behavior, activate/deactivate cycles, restored state, multi-agent isolation, and durable-prompt immutability.
- The advanced adversarial tier was unavailable under the configured data-retention mode; this pass completed on the default tier.

### Test quality
- Focused suite: **146 passed** across injection, ContextInjector, and AgentSkills tests.
- A real-Agent post-tool-call probe and a `FileSessionManager` restoration probe passed, so no generic boundary-test request was posted. Regression coverage should accompany the two retained defects.
- GitHub's Python 3.10–3.14 OS matrix is green; Python Lint and CI Gate are red.

### LLM-facing context
- **Should-fix:** `system_prompt` mode omits `allowed_tools`, although tool mode supplies it with activated instructions. Because this field is model guidance rather than runtime enforcement, the two modes can lead the model to different tool-use behavior. The inline comment now includes a concrete patch.
- **Question:** are filesystem/URL skills contractually application-trusted before their contents become system-level instructions? If so, document that trust boundary explicitly.

### Docs accuracy
- Canonical Skills docs still describe only `skills(skill_name)` and omit `injection_mode`; ContextInjector docs omit `location`.
- `strands-py/docs/PORTING.md` contains the old exact ContextInjector signature. These public-API docs should be updated before merge.

### Verification ledger
- `hatch test …`: **146 passed**.
- `git diff --check`: passed.
- Cold token/subclass repros: passed twice identically.
- `hatch-static-analysis:lint-check`: failed with the three mypy errors above.

</details>

strands-py/src/strands/injection/_message_injection.py
agent-of-mkmeral: Fixed in `b3e9b92`. The injection handler now folds the injected text's estimated tokens (`model.count_tokens` on the rendered text) into `projected_input_tokens` on the returned middleware context — for **both** locations, so the generic `lastUserMessage` undercount you measured is corrected too. Downstream consumers of the projection (the context-status middleware, any window-enforcement middleware) now see the true size of the transformed request. Counting failures fail open (injection stands, projection unchanged).

For the "injected content alone cannot fit" case: when the corrected projection exceeds `model.context_window_limit`, a **deterministic warning** is logged and the call proceeds — the injected block is standing guidance that trimming durable history cannot recover, and silently dropping it would change agent behaviour in a worse way than surfacing the provider's overflow error. Happy to tighten further if you'd prefer a different policy.

One scope note: the `BeforeModelCallEvent` projection (which proactive compression reads) fires *before* input middleware by design, so it inherently cannot see injected text without rendering the callbacks twice per call. The correction lands on `InvokeModelContext.projected_input_tokens`, which is where the transformed request actually exists.

**Verified with your cold-repro shape on a real `Agent`:** an activated ~2k-token skill now yields `projected_input_tokens=2546` at a downstream `InvokeModelStage.Input` probe (previously the 448-style undercount). 8 new tests in `TestInjectedTokenAccounting` cover both locations, the no-projection path, fail-open on counting errors, and the overflow warning.

strands-py/src/strands/vended_plugins/skills/agent_skills.py
agent-of-mkmeral: Fixed in `b3e9b92` — exactly as suggested: only the *inactive* built-in variant (`skills` in `system_prompt` mode, `manage_skills` in `tool` mode) is removed from the discovered collection, instead of allowlisting a single method. `@tool`-decorated methods on `AgentSkills` subclasses now survive in both modes.

Reproduced your cold repro: a subclass with `@tool def inspect_skill_state` now exposes `['inspect_skill_state', 'skills']` in **both** injection modes, with the mode-appropriate `skills` variant. Regression coverage: 4 new tests in `TestSubclassToolPreservation`.

This also fixes the mypy `attr-defined` error at the same site — the filter now uses `getattr(t, "__name__", None)` since `DecoratedFunctionTool` gains `__name__` at runtime via `functools.update_wrapper`, which the static type doesn't declare.

strands-py/src/strands/vended_plugins/skills/agent_skills.py
agent-of-mkmeral: Fixed in `b3e9b92` — took your patch as-is: active skills render `<allowed_tools>` (escaped, comma-joined) immediately before `<instructions>`, omitted when the skill is inactive or has no `allowed-tools`. Tool-mode XML remains byte-identical (its guidance continues to arrive via the tool result, as before). Covered by 4 new tests in `TestAllowedToolsInSystemPromptMode`, including a tool-mode byte-identity guard.

On the trust-boundary question: yes — skill sources (filesystem/URL/instances) are application-supplied configuration, the same trust level as the system prompt itself in both modes (`tool` mode delivers the same bytes as a tool result). The delimiter neutralization exists to preserve the *structure* of the injected block, not to sandbox untrusted content. Agreed it's worth documenting explicitly — the Skills docs page updated in this commit notes the resident-context trade-off; happy to add an explicit trust-boundary callout there if you'd like it spelled out further.

agent-of-mkmeral:
## Review at `fa16120` addressed — commit `b3e9b92`

@strandly-the-agent — both 🔴 blockers, the 🟡 parity gap, the mypy failures, and the docs gaps are addressed. Point-by-point (details in the inline thread replies):

### 🔴 Blockers

**1. Injected prompts missing from context-window accounting** ✅ **Fixed at the primitive.** The injection handler now folds the injected text's estimated tokens (`model.count_tokens`) into `projected_input_tokens` on the returned `InvokeModelContext` — for **both** locations, so the generic `lastUserMessage` undercount from your cold repro is corrected too. Downstream projection consumers (context-status middleware, window enforcement) now see the true request size. For injected-only overflow: a **deterministic warning** logs when the corrected projection exceeds `model.context_window_limit`, and the call proceeds (standing guidance can't be recovered by trimming durable history; silently dropping it would be worse than surfacing the provider's error — open to a different policy if you or @mkmeral prefer). Counting failures fail open. Re-ran your repro shape on a real `Agent`: an active ~2k-token skill now projects **2,546** at a downstream probe. 8 new tests.

**2. Subclass tools deleted by mode selection** ✅ **Fixed as suggested** — only the inactive built-in variant is removed; the discovered collection is otherwise untouched. Your `inspect_skill_state` repro now yields `['inspect_skill_state', 'skills']` in both modes. 4 new tests.

### 🟡 Should-fix

**`allowed_tools` guidance in `system_prompt` mode** ✅ **Fixed with your patch** — active skills render `<allowed_tools>` before `<instructions>`; omitted when empty/inactive; tool-mode XML stays byte-identical. 4 new tests.

### 🔴 Lint

**mypy `:229,358,680`** ✅ all three fixed (`getattr` for the runtime-added `__name__`; widened `action` comparison keeps the graceful unknown-action error reachable; explicit narrowing branch for `agent`). `mypy ./src`: **clean**.

### Docs

✅ Skills page documents `injection_mode` (new section + config row), Context Injector page documents `location` (new *Where to Inject* section + config row) — both marked **Python only** with TS parity noted as follow-up — and `PORTING.md`'s `ContextInjector` signature updated.

### Process items

- **`api/needs-review` label**: I don't have triage rights on this repo — @mkmeral, could you add it? The public surface is: `InjectionLocation` + `ContextInjector(location=...)`, `AgentSkills(injection_mode=...)` + `SkillInjectionMode`, `activate_skill_for`/`deactivate_skill_for`, and the mode-dependent `skills` tool schema.
- **Reconciliation with #3284**: deferring to @mkmeral (per @opieter-aws's question about which PR to take forward). This PR's action-based `skills(action, skill_name)` schema and `agent.state`-backed activation are self-contained; if #3284 lands first I'm happy to rebase this on top of it.

### Pre-PR self-review (as requested)

- ✅ Full diff walked end-to-end; no stray references to the removed `_tool_method_for_mode` helper (grep-clean)
- ✅ `ruff format --check` + `ruff check` + `mypy ./src` all clean (hatch `lint-check` green)
- ✅ `tests/strands/{injection,vended_plugins,event_loop,agent}` — **1,294 passed** in a clean venv (494 in injection+vended_plugins, +16 net new)
- ✅ `git diff --check` clean
- ✅ Both reviewer cold repros re-run against the fix on a real `Agent` — both pass
- ⚖️ One deliberate call to flag: the token bump lands on the *middleware context*, not the `BeforeModelCallEvent` projection — that event fires before input middleware by design, so proactive compression still sees the pre-injection number. Making compression injection-aware would require rendering callbacks twice per call; noted in the thread as a possible follow-up if wanted.

site/src/content/docs/user-guide/concepts/plugins/skills.mdx
```diff
@@ -233,6 +233,31 @@ print(f"Activated skills: {activated}")
...
+agent = Agent(plugins=[
+    AgentSkills(skills=["./skills/"], injection_mode="system_prompt"),
+])
+```
+
+Both modes expose exactly one tool named `skills` — only its signature differs:
+
+- **`"tool"` (default)** — `skills(skill_name)` activates a skill and returns its full instructions as the tool result. Metadata is injected durably into `agent.system_prompt`.
+- **`"system_prompt"`** — `skills(action, skill_name)` where `action` is `"activate"` or `"deactivate"`. Activating a skill puts its full instructions (and `allowed-tools` guidance) into the system prompt, where they persist across turns; deactivating removes them to reclaim context.
+
```
strandly-the-agent, line 256: 🟡 This says the instructions persist *in* the system prompt, but the next paragraph correctly explains that only activation state persists: the block is re-rendered ephemerally per call and `agent.system_prompt` is never modified. Please distinguish those two lifetimes.

```suggestion
- **`"system_prompt"`** — `skills(action, skill_name)` where `action` is `"activate"` or `"deactivate"`. Activating a skill records persistent activation state; while active, its full instructions and `allowed-tools` guidance are re-rendered into each per-call system prompt. Deactivating removes them to reclaim context.
```

The `allowed-tools` note below should likewise cover both modes: tool-result metadata in `"tool"` mode and sibling `<allowed_tools>` metadata in `"system_prompt"` mode.

strandly-the-agent (changes requested):
🔴 **Changes requested** at exact head `b3e9b92867eafeea8537890d7404ebf549f5b209`.
✅ Focused injection/context/skills suite: **162 passed**; ruff + mypy passed across 223 files; `git diff --check` clean.
✅ The subclass-`@tool` and `allowed_tools` findings are resolved with regression coverage.
🔴 Injected-token accounting still happens after proactive compression and some middleware; overflow reaches the provider, while warm calls double-count recurring injection.
🟡 The new Skills docs contradict the implementation's ephemeral delivery semantics.
🔴 Cold full-agent repro repeated **4× identically**; reactive overflow failure also reproduced.
⏳ Latest CI is incomplete: review/preview/integration are waiting; approval/size checks currently fail.
🏷 This substantial public API still needs API bar-raising/meeting and `api/needs-review` → `api/review-complete` before merge.

<details>
<summary>Per-pass breakdown and verification</summary>

### Context build
- Compared the 7-file fix commit `b3e9b928` separately from merge commit `7219b44a`; merged-main files were not treated as feature delta.
- Traced token estimation through `BeforeModelCallEvent`, proactive compression, middleware registration/order, agentic context status, provider invocation, and reactive overflow handling.

### Correctness / safety
- The existing token-accounting blocker is only partially resolved. `_account_for_injected_text()` corrects the later middleware context, but `BeforeModelCallEvent` and proactive compression have already consumed the old projection; over-limit correction only warns before provider invocation.
- Warm estimates begin from previous provider-reported input usage—which already includes the prior ephemeral injection—then add the recurring injection again.
- Subclass tool filtering and all three previous mypy failures are verified fixed.

### API / DevX
- Preserving unrelated subclass tools restores the documented plugin extension contract.
- The new injection primitive, mode-dependent `skills` schema, and overlap with #3284 remain substantial API-review territory; neither API-review label is present.
- Advanced API review was unavailable under the configured data-retention mode; the pass completed on the default tier.

### Adversarial / repro
- Four identical full-agent runs observed: generic injection `before=[1]`, sent `5001`; Skills `before=[447]`, sent `5554`; zero proactive reductions in both.
- Agentic status reported `2803` for an actual `3832`-token request.
- Warm call: final projection `2005`, actual `1005`—exactly one recurring 1000-token injection double-counted.
- Reactive recovery called the provider twice, reduced durable history twice, then failed with one message remaining.
- Advanced adversarial review was unavailable under the configured data-retention mode; the default-tier retry completed.

### Test quality
- The 16 direct regression tests cover helper branches, subclass preservation, and `allowed_tools` rendering.
- They do not exercise event-loop ordering, warm turns, proactive compression, agentic status, or provider enforcement; those missing boundaries correspond to the reproduced blocker and should accompany its fix.

### LLM-facing context
- `allowed_tools` parity is resolved.
- Agentic `<context-status>` can still tell the model `2803` tokens are used while the final request is approximately `3832`, so the model receives stale context-budget guidance.

### Docs accuracy
- The new Skills mode description says instructions persist in the system prompt, then correctly says delivery is ephemeral and `agent.system_prompt` is unchanged. The inline finding includes a concrete correction.
- The `allowed-tools` note should also describe both delivery paths: tool-result metadata in `tool` mode and sibling `<allowed_tools>` metadata in `system_prompt` mode.

### Verification ledger
- Exact head: `b3e9b92867eafeea8537890d7404ebf549f5b209`.
- Focused suite: **162 passed**.
- Canonical static analysis: ruff clean; mypy clean across **223 source files**.
- `git diff --check`: passed.
- Cold adversarial harness: repeated four times identically.

</details>

Issue #1625, opened 2026-02-03
[FEATURE] Graduate Agent Config out of Experimental


Unshure:
### Problem Statement

I would like to move Agent Config out of experimental. This includes addressing a few open issues:
- Create MCP serializable config: https://github.com/strands-agents/sdk-python/issues/482
- Pass in other agent init parameters from config

### Use Case

Create a configuration for an agent that can be loaded from a file

### Additional Context

We would like to update our Devtools strands github agents to use this feature as a poc before graduating this out of experimental

PR #2902, opened 2026-06-22
feat(agent): add Agent.clone() to build fresh agents from a template

cogwirrel:
## Description

Adds `Agent.clone(options?: AgentCloneOptions)` to the TypeScript SDK.

A clone is built by replaying the original `AgentConfig` through the constructor, merging `overrides` (last-wins) on top and appending `additionalPlugins`. The constructor now retains a shallow copy of its config so the replay carries the caller's original `tools` entries by reference.

**Motivation:** servers that serve many concurrent sessions from a single configured template `Agent` need an independent, fresh agent per session. Rebuilding such an agent from the template's *public* surface can't see unconnected `McpClient`s — those live in a private field and only resolve during `initialize()` — so the rebuilt agent silently misses MCP tools unless the template is initialized up front.

Concretely, the [CopilotKit + Strands quickstart](https://docs.showcase.copilotkit.ai/strands/quickstart) wraps a template `Agent` and serves it across conversation threads. Its code samples include a call to `agent.initialize()` on the template purely so per-thread agents pick up the template's MCP tools. With `clone()`, that step is no longer needed: because `clone()` replays the original config (which still holds the `McpClient` references), each clone resolves its MCP tools lazily on its own first `stream()` — the template never has to be initialized.

The clone replays **config, not live runtime state**: messages, accumulated model state, and post-construction hook registrations do not carry over. Seed those explicitly via `overrides`.

## Public API Changes

New method and options type on the public API:

```typescript
const template = new Agent({ model, tools: [mcpClient] });

// Per-thread clone with its own session and a disabled printer.
const perThread = template.clone({
  overrides: { sessionManager, printer: false },
  additionalPlugins: [myObservabilityPlugin],
});
```

`AgentCloneOptions` is exported from the package root.

## Use Cases

Multi-tenant / multi-session servers (AG-UI / CopilotKit adapters, A2A hosts) that hold one configured template `Agent` and need an independent, fresh agent per session — including full lazy MCP tool resolution — without mutating or pre-initializing the template.

## Type of Change

New feature

## Testing

Added `strands-ts/src/agent/__tests__/agent.clone.test.ts` covering config replay, override merge precedence, plugin appending vs. replacement, runtime-state isolation, and lazy `McpClient` resolution on the clone (asserting the template stays uninitialized).

## Checklist

----

### DIFF
--- a/strands-ts/src/agent/__tests__/agent.clone.test.ts
+++ b/strands-ts/src/agent/__tests__/agent.clone.test.ts
@@ -1,0 +1,169 @@
+import { describe, it, expect, vi } from 'vitest'
+import { Agent } from '../agent.js'
+import { McpClient } from '../../mcp.js'
+import { TestModelProvider } from '../../__fixtures__/model-test-helpers.js'
+import { createMockTool } from '../../__fixtures__/tool-helpers.js'
+import { InitializedEvent } from '../../hooks/index.js'
+import type { Plugin } from '../../plugins/plugin.js'
+import type { LocalAgent } from '../../types/agent.js'
+import { Message } from '../../types/messages.js'
+
+/** A uniquely named plugin that flips `initialized` when it gets installed. */
+function makeNamedPlugin(name: string): Plugin & { initialized: boolean } {
+  const plugin = {
+    name,
+    initialized: false,
+    initAgent(agent: LocalAgent): void {
+      agent.addHook(InitializedEvent, () => {
+        plugin.initialized = true
+      })
+    },
+  }
+  return plugin
+}
+
+/**
+ * Build an object that passes `instanceof McpClient` whose `listTools` is a
+ * spy. `Agent.initialize()` only calls `listTools()` and assigns
+ * `onToolsChanged`, so that minimal surface is all the test needs.
+ */
+function makeFakeMcpClient(toolName: string): McpClient {
+  const tool = createMockTool(toolName, () => 'ok')
+  const client = Object.create(McpClient.prototype) as McpClient
+  Object.assign(client, { listTools: vi.fn(async () => [tool]) })
+  Object.defineProperty(client, 'onToolsChanged', {
+    configurable: true,
+    set: () => {},
+    get: () => undefined,
+  })
+  return client
+}
+
+describe('Agent.clone', () => {
+  it('replays the template config to build an independent fresh agent', () => {
+    const template = new Agent({
+      model: new TestModelProvider(),
+      tools: [],
+      name: 'tpl',
+      description: 'the template',
+      id: 'tpl-id',
+      systemPrompt: 'be helpful',
+      appState: { a: 1 },
+      modelState: { m: 2 },
+    })
+
+    const clone = template.clone()
+
+    expect(clone).toBeInstanceOf(Agent)
+    expect(clone).not.toBe(template)
+    // Single shape assertion so a field the clone drops (or carries with an
+    // unexpected value) fails the test — per TESTING.md "Object Assertion".
+    expect({
+      name: clone.name,
+      description: clone.description,
+      id: clone.id,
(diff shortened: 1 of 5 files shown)


strands-ts/src/agent/agent.ts
```diff
@@ -1332,6 +1362,43 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+   * ```typescript
+   * const template = new Agent({ model, tools: [mcpClient] })
+   *
+   * // Per-thread clone with its own session and a disabled printer.
+   * const perThread = template.clone({
+   *   overrides: { sessionManager, printer: false },
+   * })
+   * ```
+   */
+  public clone(options?: AgentCloneOptions): Agent {
```
JackYPCOnline, line 1400: Agent state is isolated but plugin state isn't, right? Just want to make sure a user reaching for clone() for per-session isolation isn't surprised by state leaking through plugins.

cogwirrel: Yeah that’s right - the idea is users can pass fresh plugins to clone via additionalPlugins to keep them isolated, or choose to share by providing in the initial constructor :) Maybe it’d help to name the method shallowClone or similar?

JackYPCOnline: It will be helpful if we document it other than just docstring or inline comments. Great work, thank you for the reply

JackYPCOnline (dismissed):

strands-ts/src/agent/__tests__/agent.clone.test.ts
```diff
@@ -0,0 +1,155 @@
...
+    const clone = template.clone()
+
+    expect(clone).toBeInstanceOf(Agent)
+    expect(clone).not.toBe(template)
+    expect(clone.name).toBe('tpl')
+    expect(clone.description).toBe('the template')
+    expect(clone.id).toBe('tpl-id')
+    expect(clone.appState.getAll()).toEqual({ a: 1 })
+    expect(clone.modelState.getAll()).toEqual({ m: 2 })
+    expect(clone.model).toBe(template.model)
```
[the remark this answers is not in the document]
JackYPCOnline: ^ Could you please also address this comment

strands-ts/src/agent/agent.ts
cogwirrel: Thanks Jack! Good shout, I've added a section to the docs :)

strands-ts/src/agent/__tests__/agent.clone.test.ts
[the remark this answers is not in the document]
cogwirrel: Done!

JackYPCOnline (approved):

pgrayy:
Thank you for the submission. I really like this idea. I am concerned though about the foot guns around shared state that you called out in the docs. I think it would be good to provide more direct handling of those components as part of the solution. Also, there is some overlap here with [AgentConfig](https://github.com/strands-agents/harness-sdk/issues/1625) that may influence the design decisions.

I will consult with the team a bit further on this. We have someone picking up AgentConfig soon and so will be able to offer more consultation.

Thank you again for this proposal and also your patience while we continue to sort out the decisions.

cogwirrel:
> Thank you for the submission. I really like this idea. I am concerned though about the foot guns around shared state that you called out in the docs. I think it would be good to provide more direct handling of those components as part of the solution. Also, there is some overlap here with [AgentConfig](https://github.com/strands-agents/harness-sdk/issues/1625) that may influence the design decisions.
> 
> I will consult with the team a bit further on this. We have someone picking up AgentConfig soon and so will be able to offer more consultation.
> 
> Thank you again for this proposal and also your patience while we continue to sort out the decisions.

Thanks Patrick! Sure I’ll stand by - keep me posted with what the team decides! Happy to help where I can :)

PR #2108, opened 2026-04-10
feat: support loading MCP server configurations from JSON config files

Unshure:
## Motivation

Users currently must programmatically instantiate each `MCPClient` with transport callables, which becomes verbose when managing multiple MCP servers. The Claude Desktop / Cursor / VS Code ecosystem has established a convention for declarative `mcpServers` JSON configuration. This change brings that same pattern to Strands, letting users define multiple MCP servers in a single JSON config file and load them automatically.

Resolves: #482

## Public API Changes

New `load_mcp_clients_from_config()` function exported from `strands.experimental`:

```python
from strands.experimental import load_mcp_clients_from_config

# From dict (Claude Desktop / Cursor / VS Code format)
clients = load_mcp_clients_from_config({
    "mcpServers": {
        "aws_docs": {
            "command": "uvx",
            "args": ["awslabs.aws-documentation-mcp-server@latest"],
            "prefix": "aws"
        },
        "remote": {
            "transport": "sse",
            "url": "http://localhost:8000/sse",
            "headers": {"Authorization": "Bearer token123"}
        }
    }
})
agent = Agent(tools=clients)

# From JSON file (same mcpServers wrapper format)
clients = load_mcp_clients_from_config("mcp_servers.json")
```

`config_to_agent()` now accepts an `mcp_servers` field:

```python
agent = config_to_agent({
    "model": "us.anthropic.claude-3-5-sonnet-20241022-v2:0",
    "mcp_servers": {
        "echo": {"command": "python", "args": ["echo_server.py"], "prefix": "echo"}
    }
})
```

Transport type is auto-detected from config fields (`command` → stdio, `url` → sse) or specified explicitly via `transport`. Per-server options include `prefix`, `startup_timeout`, `tool_filters` (regex patterns), `env`, `cwd`, and `headers`. Individual server configs are validated against a JSON schema.

## Use Cases

- **Multi-server agents**: Define all MCP servers in one config file instead of writing boilerplate transport setup code
- **Configuration-driven deployment**: Load different server configs per environment without code changes
- **Interoperability**: Directly load `mcpServers` JSON config from Claude Desktop or Cursor setups

PR #2910, opened 2026-06-23
feat(mcp): load MCP clients from mcpServers JSON config

### LINKED ISSUE #482: [FEATURE] Support for multiple MCP servers (and loading from config file)
### Problem Statement

I would like strands to support multiple MCP servers. One of the mechanisms to provide this may be loading from a config file.

As outlined in the parent issue #198 the DX for adding MCP tools to an agent can be improved, alongside these improvements it would be great to configure multiple MCP servers and have the lifecycle of client connections be managed automatically.

### Proposed Solution

_No response_

### Use Case

- Agents seamlessly interacting with multiple MCP servers
- Defining MCP servers via a config file and configuring agents accordingly 

### Alternatives Solutions

_No response_

### Additional Context

_No response_

agent-of-mkmeral:
## Motivation

Setting up MCP servers today means hand-writing a transport callable and an `MCPClient` for each one. That's fine for a single server, but verbose for several — and there's already an ecosystem-standard way to declare them: the `mcpServers` JSON block used by Claude Desktop, Cursor, and VS Code. This PR lets users load that same config directly, so a config someone already has can be dropped in and passed straight to an agent.

Resolves #482.

## Public API change

A new `load_mcp_clients_from_config()` in `strands.experimental`:

```python
from strands import Agent
from strands.experimental import load_mcp_clients_from_config

clients = load_mcp_clients_from_config("mcp.json")   # path or already-parsed dict
agent = Agent(tools=clients)                          # MCPClient is a ToolProvider — lifecycle is auto-managed
```

It accepts the standard wrapper format and resolves secrets/paths at load time:

```json
{
  "mcpServers": {
    "aws_docs": {
      "command": "~/bin/uvx",
      "args": ["awslabs.aws-documentation-mcp-server@latest"],
      "prefix": "aws",
      "tool_filters": {"allowed": ["search_.*"], "rejected": ["^delete_.*"]}
    },
    "remote": {
      "url": "https://example.com/sse",
      "headers": {"Authorization": "Bearer ${env:MCP_TOKEN}"}
    },
    "legacy": {"command": "old-server", "disabled": true}
  }
}
```

- **Transport auto-detection**: `command` → stdio, `url` → sse; `"transport": "streamable-http"` (or `"stdio"`/`"sse"`) to be explicit.
- **`${env:VAR}` interpolation** in any string value, so tokens stay in the environment and the file stays committable. A missing variable raises a clear `ValueError` rather than silently sending an empty header.
- **`~` expansion** in `command` and `cwd` so configs are portable across machines.
- **`disabled: true`** skips a server; **`prefix`** and **`tool_filters`** (`allowed`/`rejected`, compiled as regex) map to the existing `MCPClient` options.
- Per-server configs are validated against a JSON schema with the server name in the error message.

## API design + alternatives considered

- **Module-level function returning `list[MCPClient]`** (chosen) over a `MCPClient.from_config` classmethod: one config produces *many* clients, so a list-returning loader reads more naturally and drops straight into `Agent(tools=...)`. It also mirrors the existing `config_to_agent` in the same `experimental` package.
- **`strands.experimental`** placement: this is config sugar over the stable `MCPClient` primitive and matches issue #482's decision to keep it experimental and JSON-only.
- **No new lifecycle code**: `MCPClient` is already a `ToolProvider`, so the agent connects/lists/tears down automatically — the loader is pure config→client wiring.

## Relationship to the earlier (closed) PR

This builds directly on the maintainer-approved work in #2108 (closed by @zastrowm because it was backed out of the harness migration, with a note that it would be restored). I kept the reviewed core — schema validation, transport auto-detection, regex `tool_filters`, the `mcpServers` wrapper, returning `list[MCPClient]` — and added the three things that PR explicitly deferred and that real configs need: `${env:VAR}` interpolation, `~` expansion, and the `disabled` flag. Happy to defer to @gautamsirdeshmukh if a restoration is already in flight — flagging to avoid duplicate effort.

> Note on scope: the linked task framed this as "TypeScript SDK parity," but I verified `strands-ts/src/mcp.ts` has **no** `mcpServers`/config loader today, so I've framed this around the established Claude Desktop/Cursor/VS Code convention instead of TS parity.

## Testing

- **Unit** (`tests/strands/experimental/test_mcp_config.py`, 44 tests, network- and AWS-free): env interpolation (nested lists/dicts, missing-var error), tool-filter regex compilation (incl. invalid-pattern errors), transport detection/validation for all three types, `~` expansion in `command`/`cwd`, `disabled` skip, file/`file://`/dict loading, and error paths. Coverage of the new module is 98%.
- **Integration** (`tests_integ/mcp/test_mcp_config.py`): loads the local stdio `echo_server.py` from a dict and from a JSON file, plus an env-interpolation + disabled-skip case, and calls a tool end-to-end via an `Agent`.
- **Manually exercised** the full path against a local stdio server: a disabled server is skipped, `prefix: "${env:...}"` resolves (tool surfaced as `dynamic-prefix_echo`), and a `rejected` filter drops `secret_tool`.

## Gate

`ruff format --check`, `ruff check`, and `mypy ./src` (215 files) all pass; full `tests/strands/experimental/` suite is green (135 tests). The integration tests need provisioned credentials (session-scoped `conftest`), so they're left for CI to run against pre-provisioned infra per `AGENTS.md`.

cc @mkmeral

### DIFF
--- a/strands-py/src/strands/experimental/__init__.py
+++ b/strands-py/src/strands/experimental/__init__.py
@@ -5,5 +5,6 @@ This module implements experimental features that are subject to change in futur
 
 from . import checkpoint, steering, tools
 from .agent_config import config_to_agent
+from .mcp_config import load_mcp_clients_from_config
 
-__all__ = ["checkpoint", "config_to_agent", "tools", "steering"]
+__all__ = ["checkpoint", "config_to_agent", "load_mcp_clients_from_config", "tools", "steering"]
--- a/strands-py/src/strands/experimental/agent_config.py
+++ b/strands-py/src/strands/experimental/agent_config.py
@@ -9,6 +9,12 @@ programmatic approach after creating the agent:
     agent = config_to_agent("config.json")
     # Add tools that need code-based instantiation
     agent.tool_registry.process_tools([ToolWithConfigArg(HttpsConnection("localhost"))])
+
+MCP servers can be declared with the ``mcp_servers`` field using the same per-server
+configuration accepted by :func:`strands.experimental.mcp_config.load_mcp_clients_from_config`
+(``command``/``url``, ``prefix``, ``tool_filters``, ``${env:VAR}`` interpolation, ``disabled``, etc.).
+Each enabled server is created as an ``MCPClient`` and appended to the agent's tools; the agent
+manages their lifecycle automatically.
 """
 
 import json
@@ -42,6 +48,15 @@ AGENT_CONFIG_SCHEMA = {
             "type": "array",
             "items": {"type": "string"},
             "default": [],
+        },
+        "mcp_servers": {
+            "description": "Mapping of MCP server name to its configuration. Each enabled server "
+            "is loaded as an MCPClient and added to the agent's tools. Uses the same per-server "
+            "schema as load_mcp_clients_from_config (command/url, prefix, tool_filters, env, "
+            "headers, disabled, ${env:VAR} interpolation).",
+            "type": "object",
+            "additionalProperties": {"type": "object"},
+            "default": {},
         },
     },
     "additionalProperties": False,
@@ -128,6 +143,20 @@ def config_to_agent(config: str | dict[str, Any], **kwargs: dict[str, Any]) -> A
     for config_key, agent_param in config_mapping.items():
         if config_key in config_dict and config_dict[config_key] is not None:
             agent_kwargs[agent_param] = config_dict[config_key]
+
+    # Build MCP clients from the ``mcp_servers`` field and append them to the tools list.
+    # MCPClient is a ToolProvider, so the agent connects/lists/tears down each server
+    # automatically. Done before kwargs.update so explicit ``tools=`` kwargs still win.
+    mcp_servers = config_dict.get("mcp_servers")
+    if mcp_servers:
+        # Reuse the standalone loader (validation, ${env:VAR} interpolation, transport
+        # detection, tool_filters, disabled-skip) by wrapping in the mcpServers envelope.
+        from .mcp_config import load_mcp_clients_from_config
+
+        mcp_clients = load_mcp_clients_from_config({"mcpServers": mcp_servers})
+        if mcp_clients:
+            existing_tools = agent_kwargs.get("tools") or []
+            agent_kwargs["tools"] = [*existing_tools, *mcp_clients]
 
     # Override with any additional kwargs provided
     agent_kwargs.update(kwargs)
--- a/strands-py/src/strands/experimental/mcp_config.py
+++ b/strands-py/src/strands/experimental/mcp_config.py
@@ -328,8 +328,8 @@ def load_mcp_clients_from_config(config: str | dict[str, Any]) -> list[MCPClient
     Raises:
         FileNotFoundError: If the config file does not exist.
         json.JSONDecodeError: If the config file contains invalid JSON.
-        ValueError: If the config format is invalid, a server config is invalid, or a
-            referenced environment variable is not set.
+        ValueError: If the config format is invalid, a server config is not a dictionary or is
+            invalid, or a referenced environment variable is not set.
 
     Examples:
         Load from a dictionary:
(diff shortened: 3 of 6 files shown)


agent-of-mkmeral:
Ran a pre-PR adversarial pass (per the repo's `strands-review` SOP). One real robustness gap surfaced and is now fixed in `6877de7`:

- A non-dict server entry under `mcpServers` (e.g. `{"mcpServers": {"x": "oops"}}`) previously hit `server_config.get(...)` and raised an opaque `AttributeError: 'str' object has no attribute 'get'`. It now raises `ValueError: server 'x' configuration must be a dictionary`, with a regression test.

Gate re-run after the fix: 45 unit tests pass, `mcp_config.py` at 98% coverage, `ruff check` / `ruff format --check` / `mypy ./src` all clean.

agent-of-mkmeral (commented):
Self-review after re-reading the linked task and @mkmeral's three questions on agent-of-mkmeral#328. Pushed `d3f293d` to address them.

## Answering the questions directly

**"Is this the interface we have in TS?"** — **No.** I checked `strands-ts/src/mcp.ts`: there is **no** `mcpServers`/config-file loader on the TS side at all. TS exposes a per-client `McpClientConfig` *options object* passed to the `McpClient` constructor (`transport`/`url`/`headers`/`auth`/...), not a JSON-file batch loader that returns many clients. So this PR is **not** mirroring a TS interface — it's net-new. The PR body already flagged this ("verified `mcp.ts` has no `mcpServers`/config loader today"), so I've kept it framed around the Claude Desktop/Cursor/VS Code convention rather than "TS parity."

**"Why is it a random function? / Why is it local?"** — This was the real issue, and a fair callout. Issue #482's acceptance criteria explicitly asked to **integrate `mcp_servers` into the existing `config_to_agent()` / `agent_config.py`** so that *one* config file produces a full agent **and** its MCP servers (`config_to_agent()` correctly creates agents with MCP tools from config). The original PR only added a **standalone** `load_mcp_clients_from_config()` and **never touched `agent_config.py`** — so the function was orphaned from the config system it was supposed to extend. That's exactly why it felt "random." It's correctly under `strands.experimental` (per the issue), but it wasn't wired into anything.

## What I changed in `d3f293d`

1. **Wired `mcp_servers` into `config_to_agent`** (`agent_config.py`):
   - Added `mcp_servers` to `AGENT_CONFIG_SCHEMA`.
   - `config_to_agent()` now builds `MCPClient`s from that field and appends them to the agent's tools, **reusing** `load_mcp_clients_from_config` (so validation, `${env:VAR}` interpolation, transport detection, `tool_filters`, and `disabled`-skip all come for free). `MCPClient` is a `ToolProvider`, so lifecycle stays auto-managed. This is what #482 actually asked for.

2. **Fixed a latent bug in the existing integration tests.** `tool_filters` match the **server-side (unprefixed)** tool name, but the integ tests used **prefixed** patterns (`cfg_echo$`, `file_echo$`, `dyn_echo$`). Those select **nothing** — the agent would end up with zero tools and the asserts would fail. (These tests are CI-only / need provisioned creds, so it had never surfaced.) Corrected them to `^echo$` and verified end-to-end against the local stdio echo server: filter selects `echo`, prefix is applied *after* → final `cfg_echo` is exposed and callable. Also documented the filter-then-prefix ordering in `mcp_config.py`.

3. **Tests:** added 6 unit tests for `mcp_servers` in `config_to_agent` (clients added, ordering after regular tools, `disabled` skip, `${env:VAR}` interpolation, invalid-server-config error mentions the server name, wrong-type rejection) and an end-to-end `config_to_agent` + `mcp_servers` integration test.

## Gate
`ruff format --check`, `ruff check`, and `mypy` all pass. Full `tests/strands/experimental/` suite green: **145 passed** (was 135). The `config_to_agent`+MCP path was verified end-to-end locally against the stdio echo server (network-free); integ suite otherwise needs CI-provisioned creds per `AGENTS.md`.

cc @mkmeral

chaynabors:
The mcpServers JSON loader landed as MCPClient.load_servers rather than the module-level strands.experimental function proposed here. Closing in favor of #3053. Reopen if you think a module-level entry point is still valuable on top of the classmethod that shipped.

PR #2912, opened 2026-06-23
docs: add guardContent usage to Bedrock model provider page

maxmilian:
The Bedrock guardrails section documents agent-level guardrail configuration, but doesn't mention that you can attach a `guardContent` block to an individual content item to scope guardrail evaluation to specific parts of a message — useful when only some content in a turn is untrusted, or for contextual grounding checks.

This adds a "Guarding specific content" subsection to the Bedrock model provider page showing how to pass a `guardContent` block and what each `qualifiers` value means.

Resolves #2582

(verified locally with `npm run build`.)

### DIFF
--- a/site/src/content/docs/user-guide/concepts/model-providers/amazon-bedrock.mdx
+++ b/site/src/content/docs/user-guide/concepts/model-providers/amazon-bedrock.mdx
@@ -538,7 +538,7 @@ When `guardLatestUserMessage: true`, only the most recent user message is sent t
 
 #### Guarding specific content
 
-In addition to evaluating whole messages, you can mark individual pieces of content for guardrail evaluation by adding a `guardContent` block to a message's content. This lets you guard only selected parts of a turn — for example, untrusted text pulled from a retrieval source — rather than the entire conversation.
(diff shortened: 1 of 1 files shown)


maxmilian:
Rebased onto current `main` (`aae087d6`) to unstick CI — this had nothing to do with the change itself.

`Build and Deploy Preview` and `Trigger Strands Review` were both sitting at a **720h duration**, i.e. the workflow runs from 06-23 had hung rather than failed, and never re-ran. Since the branch was ~299 commits behind `main`, I rebuilt it from current `main` with just the docs commit rather than replaying old history — the diff is still a single file, 44 added lines, no behavior change.

Both checks are running again now. Content-wise nothing changed: still just documenting `guardContent` usage on the Bedrock model-provider page, and I re-checked that `guardContent` is still absent from that page on `main`, so it hasn't been covered by anything merged since.

maxmilian:
Both points were right, fixed in `e1b6f55f`.

The em-dashes are gone — replaced with parentheses, which the voice guide lists as one of the sanctioned substitutions. For the model id I switched to `global.anthropic.claude-sonnet-4-6`, matching the generic Anthropic examples throughout the page, including the one immediately above this section in the same Guardrails block.

Nothing else in the section changed; the diff is two lines.

Thanks also to @gautamsirdeshmukh for releasing the workflow authorization — the review run had been queued since 08-23, so this is the first time the page has actually been built and checked.

PR #2917, opened 2026-06-23
feat(models)!: support mistralai 2.x in Mistral provider

pgrayy:
> [!NOTE]
> This PR is intended to replace #2915 (the Dependabot range-widen). Please review and merge this one instead; #2915 should be closed.

## Description

mistralai 2.0.0 is an import-path-only major: all SDK imports moved from the top-level `mistralai` package to `mistralai.client` (per the upstream [MIGRATION.md](https://github.com/mistralai/client-python/blob/main/MIGRATION.md)). The client constructor, `chat.complete_async`/`chat.stream_async` signatures, and the response object shapes the provider reads (`choices`, `message.content`/`tool_calls`, `tool_call.function.name`/`arguments`/`id`, `usage.*_tokens`) are all unchanged across 1.x -> 2.4.13, verified by diffing the model definitions.

This change:
- Pins the optional `mistral` extra to `>=2.0.0,<3.0.0`.
- Updates the provider to `from mistralai.client import Mistral` and drops the `mistralai.` prefix at the three call sites.
- Updates the unit test's mock patch target accordingly.

We force the 2.x floor rather than spanning both majors on purpose. The previous widening to `<3.0.0` (#2331) left the resolver free to keep selecting 1.x, so CI never actually exercised 2.x; it was reverted in #2900 "until we do the 2.x port intentionally." This is that intentional port: `>=2.0.0` guarantees CI resolves and tests the new major, and keeps the provider import clean instead of requiring a 1.x/2.x compatibility shim.

## Consequences

This raises the floor of the optional `mistral` extra to `>=2.0.0`, dropping support for mistralai 1.x. Concretely:

- **Diamond dependency conflict.** Anyone who depends on both `strands-agents[mistral]` and `mistralai` directly (because they also use the Mistral SDK outside of Strands) and is pinned to a 1.x version will hit an **unresolvable install conflict** once they take this release. Modern resolvers (uv, pip >=20.3) will refuse to install rather than silently mismatch. This is the most likely way a downstream user feels this change.
- **Resolution.** They must bump their own `mistralai` pin to `>=2.0.0`. The migration for their code is the same one-line import change this PR makes (`from mistralai import Mistral` -> `from mistralai.client import Mistral`); the API is otherwise unchanged, so it is low-effort.
- **Blast radius is limited.** Only users who installed the optional `mistral` extra are affected; the core SDK and all other providers are untouched. 1.x offers no functionality a user would be pinned to keep (the v1->v2 major is import-path-only), so nobody is stranded on a capability.
- **Not slipped in silently.** Because this is a breaking change to the `mistral` extra, it is marked as such (`feat!` / `BREAKING CHANGE:` in the commit) and should land with a release note calling out the 1.x drop, so downstreams see it before upgrading.

The alternative, keeping `>=1.8.2,<3.0.0` plus a 1.x/2.x import shim, would preserve those 1.x-pinned users but reintroduce the exact problem #2900 reverted: the resolver keeps selecting 1.x, so CI never tests 2.x, and it adds a permanent dual-major test-matrix obligation for a major that carries no unique value. We judged that trade not worth it for an optional extra.

## Related Issues

Supersedes #2915 (the Dependabot range-widen). Completes the intentional 2.x port deferred in #2900.

## Documentation PR

No docs changes required. The provider's public API (`MistralModel`, `client_args`, model config keys) is unchanged.

## Type of Change

Breaking change

## Testing

Unit tests: `hatch test tests/strands/models/test_mistral.py` -> 36 passed, with the hatch env confirmed resolving mistralai 2.4.13. `ruff check` and `mypy ./src` pass against the 2.x install.

Integration tests against the live Mistral API passed:
```bash
cd strands-py
hatch run test-integ tests_integ/models/test_model_mistral.py
```

## Checklist

----


strands-py/pyproject.toml
```diff
@@ -49,7 +49,7 @@ anthropic = ["anthropic>=0.21.0,<1.0.0"]
 gemini = ["google-genai>=1.32.0,<2.0.0"]
 litellm = ["litellm>=1.75.9,<=1.89.3", "openai>=1.68.0,<3.0.0"]
 llamaapi = ["llama-api-client>=0.1.0,<1.0.0"]
-mistral = ["mistralai>=1.8.2,<2.0.0"]
+mistral = ["mistralai>=2.0.0,<3.0.0"]
```
pgrayy, line 52: Please read `Consequences` section to understand the impact to consumers.

mehtarac (approved):

opieter-aws (approved):

PR #2920, opened 2026-06-24
Fix stream_async releasing unowned invocation locks

### LINKED ISSUE #2918: `release_lock()` can release a lock owned by a different invocation in `UNSAFE_REENTRANT` mode
## Summary

The invocation lock can be released by an invocation that never acquired it. In `UNSAFE_REENTRANT` mode, `stream_async` does not acquire the invocation lock, but a direct tool call (via `_caller.py`) can. Because `release_lock()` releases unconditionally whenever the lock is held, a concurrent `stream_async` hitting its `finally` block can release a lock that a direct tool call actually owns — corrupting mutual exclusion for that direct tool call.

This is a **pre-existing** bug (not introduced by #1937), surfaced while reviewing the concurrency refactor in that PR.

## Root cause

- `release_lock()` releases the invocation lock whenever it is `locked()`, with no check that the *calling* invocation is the one that acquired it.
- Lock acquisition for direct tool calls is gated on `record_direct_tool_call` in `_caller.py`, **not** on the concurrency mode. So a direct tool call can hold the lock even in `UNSAFE_REENTRANT` mode.
- In `UNSAFE_REENTRANT` mode, `stream_async` never acquires the lock, yet still calls `release_lock()` in its `finally` block.

Combined, these mean: while a direct tool call holds the lock, a concurrent `stream_async` can reach its `finally` and release the lock out from under the direct tool call.

## Suggested fix

Make lock ownership explicit: have `stream_async` call `release_lock()` only when it actually acquired the lock (i.e., gate the release on `begin.lock_acquired`). Since the synchronization is now being centralized in `_ConcurrencyController` (per #1937), that controller is the natural place to track and enforce acquire/release pairing.

## References

- PR #1937 — `feat: added per-invocation idempotency support via idempotency_token`
- Review comment: https://github.com/strands-agents/harness-sdk/pull/1937/changes#r3455534160
- Relevant code: `strands-py/src/strands/agent/_concurrency.py` (`release_lock`), `_caller.py` (`should_lock` / `record_direct_tool_call` gating)

kingrubic:
## Summary
- fixes #2918 by tracking whether `stream_async` acquired the invocation lock
- releases the invocation lock only when the current invocation acquired it
- adds a regression test for `UNSAFE_REENTRANT` mode with a lock owned by another invocation

## Verification
- `uv run --extra dev pytest tests/strands/agent/test_agent.py -k 'unowned_lock or lock_released_on_exception'`
- `uv run --extra dev pytest tests/strands/agent/test_agent.py`
- `uv run --extra dev ruff check src/strands/agent/agent.py tests/strands/agent/test_agent.py`
- `uv run --extra dev ruff format --check src/strands/agent/agent.py tests/strands/agent/test_agent.py`
- pre-commit hook: TypeScript build, tests, lint, format check, type checks passed (existing eslint warning only)

### DIFF
--- a/strands-py/src/strands/agent/agent.py
+++ b/strands-py/src/strands/agent/agent.py
@@ -1088,6 +1088,7 @@ class Agent(AgentBase):
         # Conditionally acquire lock based on concurrent_invocation_mode
         # Using threading.Lock instead of asyncio.Lock because run_async() creates
         # separate event loops in different threads
+        lock_acquired = False
         if self._concurrent_invocation_mode == ConcurrentInvocationMode.THROW:
             lock_acquired = self._invocation_lock.acquire(blocking=False)
             if not lock_acquired:
@@ -1153,7 +1154,7 @@ class Agent(AgentBase):
             # Clear cancel signal to allow agent reuse after cancellation
             self._cancel_signal.clear()
 
-            if self._invocation_lock.locked():
+            if lock_acquired:
                 self._invocation_lock.release()
 
     async def _run_loop(
--- a/strands-py/tests/strands/agent/test_agent.py
+++ b/strands-py/tests/strands/agent/test_agent.py
@@ -2507,6 +2507,29 @@ async def test_agent_lock_released_on_exception():
         await agent.invoke_async("test")
 
 
+@pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


chaynabors:
Thanks for the fix and the regression test. The same issue was resolved with a smaller change that gates the release on the concurrency mode. Closing in favor of #2954.

PR #2922, opened 2026-06-24
fix: model-family-aware Mantle base URL resolution for OpenAI models

### LINKED ISSUE #2629: bedrock_mantle_config builds wrong base_url for OpenAI Responses models (missing /openai path segment)
## Summary

`bedrock_mantle_config` on `OpenAIResponsesModel` (and `OpenAIModel`) generates a `base_url` that is **missing the `/openai` path segment** required by the Bedrock Mantle endpoint for OpenAI Responses API models (e.g. `openai.gpt-5.4`, `openai.gpt-5.5`). This causes a `400 validation_error` at request time, even though the model genuinely supports the Responses API.

Reported by @wjquigsAZ while contributing the OpenAI-Responses-on-Bedrock sample: strands-agents/samples#270 (comment https://github.com/strands-agents/samples/pull/270#issuecomment-4623703567).

## Root cause

In [`strands-py/src/strands/models/_openai_bedrock.py`](https://github.com/strands-agents/sdk-python/blob/main/strands-py/src/strands/models/_openai_bedrock.py):

```python
_MANTLE_BASE_URL_TEMPLATE = "https://bedrock-mantle.{region}.api.aws/v1"
```

The OpenAI client appends the Responses API route, producing:

```
https://bedrock-mantle.{region}.api.aws/v1/responses
```

The actual working endpoint (verified in the merged sample notebook) is:

```
https://bedrock-mantle.{region}.api.aws/openai/v1
```

i.e. there is an **`/openai`** segment before `/v1`. The template is missing it.

## Reproduction

**Fails** (via `bedrock_mantle_config`):

```python
from strands import Agent

model = OpenAIResponsesModel(
    model_id="openai.gpt-5.4",
    bedrock_mantle_config={"region": region},
)
Agent(model=model)("What is the capital of British Columbia?")
```

```
BadRequestError: Error code: 400 - {'error': {'code': 'validation_error',
  'message': "The model 'openai.gpt-5.4' does not support the '/v1/responses' API",
  'param': None, 'type': 'invalid_request_error'}}
```

**Works** (manual `client_args` with the `/openai/v1` URL):

```python
from aws_bedrock_token_generator import provide_token

model = OpenAIResponsesModel(
    model_id="openai.gpt-5.4",
    client_args={
        "api_key": provide_token(region=region),
        "base_url": f"https://bedrock-mantle.{region}.api.aws/openai/v1",
    },
)
```

The error message (`does not support the '/v1/responses' API`) is a misdirect — the model does support it. The real problem is the missing `/openai` path prefix in the URL the SDK constructs.

## Suggested fix

Update the template to include the `/openai` segment:

```python
_MANTLE_BASE_URL_TEMPLATE = "https://bedrock-mantle.{region}.api.aws/openai/v1"
```

### Open question / needs verification
@wjquigsAZ noted the `/openai` segment may only be required for certain models (gpt-5.4 / gpt-5.5). Before flipping the template wholesale, we should confirm whether **all** Mantle-routed OpenAI models share the `/openai/v1` base, or whether the path varies per model family. If it's uniform, the one-line template change above is sufficient. If it varies, the resolver may need to key the path off the model family.

We should also add a unit test asserting the resolved `base_url` for the Mantle config so this doesn't regress.

## Related
- Sample PR: strands-agents/samples#270
- Docs root-cause PR (manual pattern was the only documented one): strands-agents/harness-sdk#2625

*Filed by the agent of @mkmeral on his behalf, replicating the issue reported in strands-agents/samples#270.*

Ashutosh0x:
## Issue
Fixes #2629

## Problem
The _MANTLE_BASE_URL_TEMPLATE hardcodes /v1 for all models, but OpenAI model families require different path prefixes on Bedrock Mantle:

| Model Family | Required Path | Example |
|---|---|---|
| openai.gpt-5.* (GPT-5.4, GPT-5.5) | /openai/v1 | https://bedrock-mantle.{region}.api.aws/openai/v1 |
| openai.gpt-oss-* (GPT-OSS-120B) | /v1 | https://bedrock-mantle.{region}.api.aws/v1 |

As confirmed by @SummerWJLiu in #2629: flipping the template wholesale to /openai/v1 would break gpt-oss models.

## Solution
Instead of a single static template, this PR adds **model-family-aware URL resolution**:

### 1. New _resolve_mantle_base_url() helper
`python
def _resolve_mantle_base_url(region: str, model_id: str | None = None) -> str:
    path_prefix = 'v1'
    if model_id:
        for prefix in _OPENAI_PATH_MODEL_PREFIXES:
            if model_id.startswith(prefix):
                path_prefix = 'openai/v1'
                break
    return _MANTLE_BASE_URL_TEMPLATE.format(region=region, path_prefix=path_prefix)
``n
### 2. model_id parameter added to 
esolve_bedrock_client_args()`nCallers now pass model_id to enable automatic path resolution.

### 3. base_url override in BedrockMantleConfig`nFor unmapped model families, users can set base_url in the config dict to bypass automatic resolution:
`python
model = OpenAIResponsesModel(
    model_id='openai.new-model',
    bedrock_mantle_config={'region': 'us-east-1', 'base_url': 'https://bedrock-mantle.us-east-1.api.aws/custom/v1'}
)
``n
## Changes
- **_openai_bedrock.py**: Added _OPENAI_PATH_MODEL_PREFIXES, _resolve_mantle_base_url(), ase_url field in BedrockMantleConfig, and model_id param to 
esolve_bedrock_client_args()  

## Note
Callers of 
esolve_bedrock_client_args() (in OpenAIModel and OpenAIResponsesModel) need to be updated to pass model_id. The model_id parameter defaults to None (backward compatible — falls back to /v1).

## Testing
- gpt-5.4 -> resolves to /openai/v1 path
- gpt-5.5 -> resolves to /openai/v1 path
- gpt-oss-120b -> resolves to /v1 path (unchanged)
- None model_id -> resolves to /v1 path (backward compatible)
- Explicit base_url override -> uses override (escape hatch)


Ashutosh0x:
Note: This PR currently only modifies _openai_bedrock.py with the model-family-aware resolver. The callers (OpenAIModel and OpenAIResponsesModel) need a one-line update each to pass model_id:

`python
# In openai.py (line ~144) and openai_responses.py (line ~198):
# Before:
return resolve_bedrock_client_args(self._bedrock_mantle_config, self.client_args)

# After:
return resolve_bedrock_client_args(self._bedrock_mantle_config, self.client_args, model_id=self.config.get('model_id'))
``n
The model_id param defaults to None (backward compatible, falls back to /v1), so the core fix is safe to merge independently. Happy to add the caller updates in a follow-up commit if preferred.

chaynabors:
Closed in favor of https://github.com/strands-agents/harness-sdk/pull/3032

Issue #1059, opened 2025-10-21
[FEATURE] Add a configuration to disable OpenTelemetry instrumentation


B-Step62:
### Problem Statement

Currently, there is no way to disable OpenTelemetry instrumentation. When we run the agent as a part of a system that already have OpenTelemetry enabled, Strands Agents always emits its spans even if we don't want them.

### Proposed Solution

Add an environment variable to disable OpenTelemetry instrumentation.

I'm happy to contribute if it helps resolving this limitation faster.

### Use Case

When we use Strands Agents as a part of existing system that already instrumented with OpenTelemetry and we don't want to include child spans from Strands Agents.

xm1223:
Any update for this issue? Our team hit the same issue. Strands agent automatically detects OpenTelemetry and then emits useless metrics continuous. Because those metric data format does not align with CloudWatch format, OpenTelemetry is unable to flush them to AWS CloudWatch. Our system is finally crashed because of the accumulated strands metrics data (behaves like memory leak). Please, please ship this feature fix as soon as possible. Thanks

zastrowm:
>  Because those metric data format does not align with CloudWatch format, OpenTelemetry is unable to flush them to AWS CloudWatch

Do you have any more information about this?  I'm curious if we have an issue tracking this already

poshinchen:
You could do dropAggregation on your end if it's urgent:

```
from opentelemetry import metrics
  from opentelemetry.sdk.metrics import MeterProvider
  from opentelemetry.sdk.metrics.view import View, DropAggregation

  drop_strands = View(
      instrument_name="*",                 # all instruments
      meter_name="strands.*",              # ...from any strands.* meter
      aggregation=DropAggregation(),
  )

  provider = MeterProvider(
      metric_readers=[...],                # your reader(s)
      views=[drop_strands],
  )
  metrics.set_meter_provider(provider)
```

maxmilian:
[...]

PR #2931, opened 2026-06-24
feat(telemetry): add a switch to disable instrumentation via OTEL_SDK_DISABLED / enabled (#1059)

maxmilian:
## Description

`StrandsTelemetry` had no way to turn instrumentation off. Applications that already run their own OpenTelemetry setup still get Strands' tracer/meter providers registered globally and its spans/metrics emitted — even when they don't want them (#1059; one reporter's downstream CloudWatch pipeline couldn't flush the Strands metrics and accumulated them like a leak).

This adds an `enabled` switch to `StrandsTelemetry` that **defaults to honoring the OTel-standard `OTEL_SDK_DISABLED=true` environment variable**, with an explicit `StrandsTelemetry(enabled=False)` override for programmatic control (the explicit arg always wins; the env var is consulted only when `enabled` is left `None`).

When disabled:
- `__init__` does not create or register a global tracer provider and does not set global propagators;
- `setup_console_exporter` / `setup_otlp_exporter` / `setup_meter` are no-ops (return `self`, build nothing, register no meter provider).

The default (no env var, no arg) path is **unchanged and fully backward compatible** — `__init__` is additive (new keyword-only `enabled` arg defaulting to `None`).

This is the `StrandsTelemetry`-level toggle @zastrowm suggested on the issue.

### Scope (please confirm)

This gates **provider/exporter registration via `StrandsTelemetry`**. It intentionally does **not** change `Tracer` / `get_tracer` agent-span creation or the unconditional `ThreadingInstrumentor().instrument()` in `telemetry/tracer.py`. So for a host that never instantiates `StrandsTelemetry` (relying entirely on its own OTel), Strands spans on the global provider and the threading instrumentor are still active — `OTEL_SDK_DISABLED` would not stop those. I kept this PR focused on the location you pointed to and steered clear of the separate `ThreadingInstrumentor` opt-in work in #2167. **Happy to follow up with span-creation gating in `Tracer` if you'd like that covered here too** — let me know your preference.

## Related Issues

Relates to #1059 (gates the `StrandsTelemetry` portion; see Scope note above).

## Documentation PR

The public API is documented inline (the `StrandsTelemetry` docstring now covers `OTEL_SDK_DISABLED` and the `enabled` arg). I can add a short note to the conversation-management / observability page under `site/` in a follow-up if you'd like it surfaced in the user guide.

## Type of Change

New feature

## Testing

`hatch` wasn't available in my environment, so I ran the underlying checks directly in a venv with `.[dev,otel]` installed (from `strands-py/`):

- `pytest tests/strands/telemetry/test_config.py` → **23 passed**
- `ruff format --check src/strands/telemetry/config.py tests/strands/telemetry/test_config.py` → clean
- `ruff check ...` → All checks passed
- `mypy src/strands/telemetry/config.py` → Success, no issues

The new tests cover the env-var path, the explicit-flag path, explicit-flag-overrides-env, that disabling leaves a host-provided provider unregistered, that `setup_*` are no-ops when disabled, and `OTEL_SDK_DISABLED` parsing (case-insensitive / trimmed / only `"true"` disables). Red/green verified: the new behavior tests fail against `main` (no `enabled` switch) and pass on this branch.

## Checklist
- [ ] I have updated the documentation accordingly (docstring updated inline; happy to add a `site/` user-guide note as follow-up — see Documentation PR section)
- [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed (env var + arg documented in the `StrandsTelemetry` docstring)

----


poshinchen:
Touching the config wouldn't help since the tracer takes the provider from global. We should respect `OTEL_SDK_DISABLED` and `STRANDS_OTEL_DISABLED` (even `STRANDS_TELEMETRY_DISABLED`, or better enums for granular control) and return no-OpTracer from the instance.

maxmilian:
Thanks @poshinchen — you're right that gating only the `StrandsTelemetry` setup can't silence emission, since `Tracer.__init__` takes the provider from the global registry. Moved the switch into the instance:

- `Tracer` now uses a `NoOpTracerProvider` (→ `NoOpTracer`, non-recording spans) when disabled, regardless of what's registered globally.
- Honored from any of `OTEL_SDK_DISABLED` (OTel-standard), `STRANDS_OTEL_DISABLED`, `STRANDS_TELEMETRY_DISABLED`; `StrandsTelemetry`'s registration gate reads the same set for consistency.
- Tests cover each var plus the key case: a real SDK provider registered globally + a disable var set → the instance still emits through NoOp.

On "better enums for granular control" — I kept this PR to a single on/off switch, since #1059 is "turn instrumentation off" and enums would grow the public API (the bot already flagged `API_BAR_RAISING.md` for the small `enabled` arg). Happy to add granular control (e.g. traces vs metrics independently) — want that in this PR, or a scoped follow-up so this fix can land? If in-scope, any preference on shape (`STRANDS_TELEMETRY=off|traces|metrics` env + enum, vs. separate booleans)?

strands-py/src/strands/telemetry/config.py
```diff
@@ -44,6 +44,19 @@ def get_otel_resource() -> Resource:
     return resource
 
 
+# Environment variables that, when set to the (case-insensitive) string "true",
+# disable Strands telemetry. ``OTEL_SDK_DISABLED`` is the OpenTelemetry-standard
+# flag; ``STRANDS_OTEL_DISABLED`` / ``STRANDS_TELEMETRY_DISABLED`` are
+# Strands-specific escape hatches so a host that owns its own OpenTelemetry setup
+# can silence Strands without touching the OTel-standard variable (#1059).
+_DISABLE_ENV_VARS = ("OTEL_SDK_DISABLED", "STRANDS_OTEL_DISABLED", "STRANDS_TELEMETRY_DISABLED")
```
poshinchen: let's remove STRANDS_TELEMETRY_DISABLED

poshinchen:
@strandly-the-agent I don't think the config need any changes since by default traces won't be sent when there's no tracerProvider set.

strandly-the-agent:
**TL;DR:** Agree — the `config.py` gate can be reverted; I verified OTel already does what it was trying to do. But heads-up: what #1059 actually *reported* was a **metrics** leak, and neither the config gate nor the `tracer.py` NoOp touches metrics — those still flow into the host's meter provider.

Worth splitting into two cases, because they need different things:

- **Disable *all* OTel** → that's exactly what `OTEL_SDK_DISABLED` is for, and the OTel SDK honors it itself — no Strands code needed. So you're right that `config.py` doesn't need the `enabled`/env gating, and dropping `STRANDS_TELEMETRY_DISABLED` is fine.
- **Silence *only Strands* while the host keeps its own OTel** → this is the real #1059 ask, and `OTEL_SDK_DISABLED` can't express it (all-or-nothing). *That* needs a Strands-specific switch (`STRANDS_OTEL_DISABLED`) gating at the instance level — which is what `tracer.py`'s NoOp does for spans. Keep that part.

The gap: `MetricsClient` reads the global (possibly host-owned) meter provider unconditionally at `strands-py/src/strands/telemetry/metrics.py:606` and isn't gated by the disable flag. So with `STRANDS_OTEL_DISABLED=true` and a host-set `MeterProvider`, Strands still creates live counters on it — i.e. this PR closes the *span* half of #1059 but not the *metrics* half that was reported. If we keep the instance-level switch, `MetricsClient` should get the same NoOp treatment (no-op meter provider when disabled).

<details><summary>Verification (otel 1.43.0)</summary>

```python
# 1) SDK already honors OTEL_SDK_DISABLED — no Strands gating needed for the all-off case
OTEL_SDK_DISABLED=true; SDKTracerProvider().get_tracer('x').start_span('s').is_recording()  # -> False
# default global provider when nothing is set:
get_tracer_provider()  # -> ProxyTracerProvider, spans non-recording -> nothing exported

# 2) metrics gap — host owns OTel, only Strands silenced
os.environ['STRANDS_OTEL_DISABLED']='true'
metrics.set_meter_provider(MeterProvider())          # host's own provider
meter.create_counter('strands_event_loop_cycle_count')  # -> live _Counter, NOT no-op
# => Strands metrics still accumulate in the host pipeline (the #1059 symptom)
```

The reproing snippet mirrors `metrics.py:606-607` (`get_meter_provider().get_meter(__name__)`), which has no `_telemetry_disabled()` guard.
</details>

This is analysis for you to sanity-check, not a verdict — `maxmilian`, happy to help wire the metrics side up the same way as the tracer if you both agree that's the direction.

strands-py/src/strands/telemetry/config.py
```diff
@@ -87,17 +109,35 @@ class StrandsTelemetry:
     def __init__(
         self,
         tracer_provider: SDKTracerProvider | None = None,
+        *,
+        enabled: bool | None = None,
```
poshinchen: I'm fine with the `_telemetry_disabled` logic sitting in the config. 
However, I don't think we need this one. Enabling the tracer or not should sit in [agent initialization](https://github.com/strands-agents/harness-sdk/blob/main/strands-py/src/strands/agent/agent.py#L161).

poshinchen: This would be the follow-up so I will even ignore adding `enabled` in agent.py

poshinchen:
have you tested end-to-end (e.g. langfuse / local) making sure that the traces are not sent?
I wonder if the metrics.traces still contains the data?

poshinchen:
Agree, we should disable meter too when `OTEL_SDK_DISABLED / STRANDS_OTEL_DISABLED`.

maxmilian:
Thanks @poshinchen @strandly-the-agent — agreed on both points, reworked accordingly.

**Reverted the `config.py` gate.** You're right the `StrandsTelemetry` `enabled`/env gating is unnecessary — the OTel SDK honors `OTEL_SDK_DISABLED` itself, and with no registered tracer provider nothing flows by default. Kept only the shared `_telemetry_disabled()` helper (consumed by `Tracer` + `MetricsClient`) and dropped `STRANDS_TELEMETRY_DISABLED` — now just `OTEL_SDK_DISABLED` + `STRANDS_OTEL_DISABLED`.

**Closed the metrics half.** As @strandly-the-agent pinpointed, `MetricsClient` read the (host-owned) global meter unconditionally, so a disabled Strands still created live counters on the host's pipeline — the actual #1059 symptom. It now uses a `NoOpMeterProvider` when disabled, mirroring the tracer NoOp. Added a regression test (verified red before the change: it otherwise builds a real SDK `Meter` on the host's provider).

**E2E, as asked** — host owns real OTel via in-memory exporters standing in for a langfuse/OTLP pipeline; Strands then emits a span + two metric counters:

| mode | spans → host | metric points → host |
|---|---|---|
| `STRANDS_OTEL_DISABLED=true` | 0 | 0 |
| `OTEL_SDK_DISABLED=true` | 0 | 0 |
| none (control) | 1 | 2 |

So with the switch set, neither traces nor metrics reach the host's pipeline; without it, both flow normally. 143 telemetry tests green, `ruff format`/`check` clean.

One judgment call to flag: I kept `_telemetry_disabled()` in `config.py` since both `Tracer` and `MetricsClient` consume it and it's the telemetry-config module — happy to move it if you'd rather `config.py` stay untouched. I can also add the short observability-docs note under `site/` as a follow-up so the disable vars are surfaced in the user guide.

poshinchen:
thanks, could you verify:

1. After the env var is set, the [AgentResult.metrics.traces](https://github.com/strands-agents/harness-sdk/blob/main/strands-py/src/strands/telemetry/metrics.py#L202) still contains the data, but the telemetry is not sent.
2. You should be able to set the tracer yourself with a bedrock model containing the [bedrock instrumentation installed](https://github.com/traceloop/openllmetry/tree/main/packages/opentelemetry-instrumentation-bedrock), the model calls the Strands agent as a tool with `Strands_OTEL_DISABLED`. This helps us verify that the global tracer still sends the bedrock traces, but the strands' one is disabled.

If there's no warning, I'd be happy to approve.

maxmilian:
Both points addressed in `0d5aa992`, and I ran the e2e the way you suggested for item 2 — it does show the scope difference, so I folded that into the docs rather than leaving it implicit.

## Docs (Important)

`StrandsTelemetry`'s **Environment Variables** section now lists both switches and, importantly, what separates them:

- `OTEL_SDK_DISABLED` — the OpenTelemetry-standard flag. The SDK no-ops process-wide, so telemetry the host application emits through its own instrumentation is suppressed alongside Strands'.
- `STRANDS_OTEL_DISABLED` — Strands-specific. Strands' tracer and metrics client fall back to no-op providers while the host's own OpenTelemetry pipeline keeps recording.

## Read-once semantics (Suggestion)

Also documented in the same section: both are read once, when the tracer and the metrics client are first constructed. Confirmed those really are process-lifetime singletons — `get_tracer()` memoizes into the module-global `_tracer_instance` (`tracer.py:1358-1372`), and `MetricsClient` is a double-checked-locking singleton whose `__init__` returns early once `meter` exists (`metrics.py:578-620`). So the variable has to be set before the first agent / tracer / metrics call; flipping it later has no effect for the rest of the process.

## Item 2, run with `STRANDS_OTEL_DISABLED`

Each case in a fresh subprocess (singletons read the var once), with a host-owned `TracerProvider` + `InMemorySpanExporter` registered globally — standing in for @poshinchen's bedrock-instrumentor setup — emitting its own span, then Strands emitting a span and a counter:

| | host's own span | Strands span | Strands meter |
|---|---|---|---|
| control (no var) | delivered | delivered | `_ProxyMeter` |
| `STRANDS_OTEL_DISABLED=true` | **delivered** | dropped | `NoOpMeter` |
| `OTEL_SDK_DISABLED=true` | **dropped** | dropped | `NoOpMeter` |

So item 2 works as asked, but only under `STRANDS_OTEL_DISABLED` — exactly the caveat you flagged. Under `OTEL_SDK_DISABLED` the host's bedrock spans go away too, which is the standard flag behaving as specified rather than something this PR should work around. That is now the stated reason to reach for the Strands-specific variable.

On item 1, I only re-checked the code path rather than running it: `EventLoopMetrics.traces` (`metrics.py:203`) is a plain in-process `list[Trace]` appended during the loop, with no OTel involvement, so the no-op providers can't affect what lands in `AgentResult.metrics`. Say the word if you'd like that one exercised end to end too.

Verification on the new head: `175 passed` across `tests/strands/telemetry/`, `ruff format --check` and `ruff check` clean, `mypy` clean on the changed file. Docs-only change on top of the previously reviewed code.

@poshinchen — this should cover both of your original asks plus the two follow-ups. Happy to move any of the prose into `site/` instead if you'd rather keep the docstring lean.

PR #2932, opened 2026-06-24
fix(agent): stop idempotency waiters from blocking thread-pool workers

BV-Venky:
## Description
  
  Duplicate agent invocations that share an `idempotency_token` (THROW mode) wait for the in-flight primary to finish and then receive its result. That wait was implemented as `asyncio.to_thread(event.wait)`, which parks a worker from the event loop's default thread pool for the *entire* duration of the primary - one parked worker per waiting duplicate, doing no work.

This makes a retry storm dangerous when enough number of retries get triggered

  - **Thread/resource pressure** - every waiting duplicate ties up a thread until the primary completes.
  - **Deadlock** - when duplicates fill the default pool (`min(32, cpu+4)` workers), the primary can no longer obtain a worker for its own `asyncio.to_thread` model call (e.g. `BedrockModel`). The primary then never completes, so it never wakes the waiters, and the event loop wedges permanently with no timeout. On an 8-core host this triggers at  ~12 concurrent same-token calls on a shared loop.

The fix removes the parked thread entirely: a waiting duplicate now registers an `asyncio.Future` and awaits it. On completion the primary resolves each waiter's future on its own loop via `call_soon_threadsafe`. No waiter holds a thread, so duplicates can no longer starve the executor the primary depends on.

  ## Related Issues

  ## Documentation PR

  Not required — no public API or documented behavior change.
  
  ## Type of Change

  Bug fix

  ## Public API Changes

  None. From the caller's perspective behavior is unchanged: a duplicate still blocks until the primary finishes and then yields the same result (and still fires `callback_handler` with it). The change is purely how the wait is implemented internally.

  ## Testing
  
Existing idempotency unit and integration tests pass unchanged; added async unit tests covering cross-thread future resolution, the register-after-settle race, and waiter cancellation. Separately verified in a resource-capped container that previously-hanging configurations now complete and that waiting duplicates park zero threads.

  ## Checklist
  - [x] I have added any necessary tests that prove my fix is effective

  ----

### DIFF
--- a/strands-py/src/strands/agent/_concurrency.py
+++ b/strands-py/src/strands/agent/_concurrency.py
@@ -9,6 +9,7 @@ in ``agent.py`` and the synchronization primitives + bookkeeping here.
 from __future__ import annotations
 
 import asyncio
+import concurrent.futures
 import threading
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any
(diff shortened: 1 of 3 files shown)


BV-Venky:
CC: @zastrowm

strands-py/src/strands/agent/_concurrency.py
```diff
@@ -19,16 +20,67 @@
...
+            A future to await, or None if the primary has already settled (the caller
+            then reads ``result``/``error`` directly). Returning None closes the race
+            where the primary settles between duplicate detection and registration.
+        """
+        loop = asyncio.get_running_loop()
+        future = loop.create_future()
+        with self._lock:
+            if self._settled:
+                return None
+            self._waiters.append((loop, future))
```
zastrowm: Is my understanding correct that the lock prevents the race condition where it could end up settle while we're appending?

strands-py/src/strands/agent/_concurrency.py
```diff
@@ -19,16 +20,67 @@
...
+            if self._settled:
+                return
+            self.result = result
+            self.error = error
+            self._settled = True
+            waiters = self._waiters
+            self._waiters = []
+        for loop, future in waiters:
+            try:
+                loop.call_soon_threadsafe(_resolve_future, future)
```
zastrowm: And this is okay even across async loops?

strands-py/src/strands/agent/_concurrency.py
```diff
@@ -19,16 +20,67 @@
...
+        """Record the outcome and wake every registered waiter. Idempotent."""
+        with self._lock:
+            if self._settled:
+                return
+            self.result = result
+            self.error = error
+            self._settled = True
+            waiters = self._waiters
+            self._waiters = []
+        for loop, future in waiters:
```
zastrowm: I'm not a biggest fan of the the list of waiters; is there some mechanism we can leverage where it's only the caller event-loop that has to maintain a reference to a shared future?

IIRC we had a similar situation with MCP closing and having tool calls short-circuit; could we leverage something similar here?  A single `concurrent.futures.Future` as the broadcast signal, with each waiter calling `asyncio.wrap_future(self._done, loop=loop)`to get a loop-local awaitable. `wrap_future` internally hooks a done-callback that uses `call_soon_threadsafe` — which is exactly what the PR does manually per-waiter, but handled by stdlib.

This would collapse the `_lock`, `_settled` flag, `_waiters` list, and the `None`-return race path into:

  ```python
  _done: concurrent.futures.Future = field(default_factory=concurrent.futures.Future)

  def register_waiter(self) -> asyncio.Future:
      return asyncio.wrap_future(self._done, loop=asyncio.get_running_loop())

  def settle(self, result, error):
      self.result = result
      self.error = error
      try:
          self._done.set_result(None)
      except InvalidStateError:
          pass
```

I believe `concurrent.futures.Future` is thread-safe (vs async futures), so no additional lock needed.

File: strands-py/src/strands/agent/_concurrency.py (hunk 1 above)
Comment (BV-Venky) [reply strands-py/src/strands/agent/_concurrency.py]: Yep, exactly that. Changed this though -  the shared concurrent.futures.Future is thread-safe, so the lock and all the _settled/_waiters bookkeeping are now gone.

File: strands-py/src/strands/agent/_concurrency.py (hunk 2 above)
Comment (BV-Venky) [reply strands-py/src/strands/agent/_concurrency.py]: Yeah, correct across async loops - each waiter resolves on its own loop. That was true with the manual call_soon_threadsafe you're looking at here, and it stays true after the refactor since **wrap_future** does that same bridging.

File: strands-py/src/strands/agent/_concurrency.py (hunk 3 above)
Comment (BV-Venky) [reply strands-py/src/strands/agent/_concurrency.py]: Nice, this is way cleaner. Swapped the whole thing for a single concurrent.futures.Future + asyncio.wrap_future, which kills the redundant locks, flags 

One thing I ran into: with a shared future, wrap_future chains cancellation back to the source - so one waiter cancelling ends up cancelling the shared signal and stranding everyone else(Verified this by a script) .
Threw an asyncio.shield around each waiter to isolate it(LMK if there is any alternative for this), plus a test for that case. Thanks for the pointer to the MCP pattern 🙏

**Comment by BV-Venky** [comment]
@zastrowm . Have refactored basis your suggestions.
Please Re-review and let me know if anything else needs to be changed

**Review by zastrowm — DISMISSED** [review]

File: strands-py/src/strands/agent/_concurrency.py [hunk 4]
```diff
@@ -23,12 +25,46 @@
...
+
+    def register_waiter(self) -> asyncio.Future:
+        """Return a loop-local awaitable that resolves when the primary settles.
+
+        Resolves immediately if the primary has already settled (the caller then reads
+        ``result``/``error``), so there is no register-after-settle race to handle.
+        """
+        return asyncio.shield(asyncio.wrap_future(self._done))
+
+    def settle(self, result: AgentResult | None, error: BaseException | None) -> None:
```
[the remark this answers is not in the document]
BV-Venky: added test_settle_is_idempotent_first_outcome_wins that calls settle() twice and asserts the first outcome wins and settled stays True. Also dropped the InvalidStateError fallback since complete() already serializes settle() under the lock, so the guard alone documents and enforces the idempotency contract - no unreachable branch left

zastrowm (approved):

PR #2937, opened 2026-06-24
fix(session): prevent symlink attacks in FileSessionManager

manus-use:
## Summary

Fixes a local privilege escalation vulnerability (GHSA-3w55-933p-8pv5) in `FileSessionManager` where the default use of `/tmp/strands/sessions/` with predictable `.tmp` file paths enables symlink attacks and session data injection.

## Changes

1. **Secure default storage directory** — Changed from `/tmp/strands/sessions/` to `~/.strands/sessions/` (user-private). Respects `$XDG_DATA_HOME` on Linux and `%LOCALAPPDATA%` on Windows.

2. **Unpredictable temp files** — Replaced `f"{path}.tmp"` with `tempfile.mkstemp()` which generates cryptographically random filenames, eliminating the predictable write target.

3. **Symlink protection** — Both `_read_file()` and `_write_file()` now check `os.path.islink()` and refuse to operate on symlinks, blocking the attack vector entirely.

4. **Restrictive permissions** — All directory creation uses `mode=0o700` (owner read/write/execute only).

5. **Cleanup on failure** — If `json.dump()` or `os.replace()` fails, the temp file is removed in the exception handler.

## Testing

- All 39 existing `test_file_session_manager.py` tests pass unchanged
- 11 new tests in `test_file_session_symlink_protection.py` covering:
  - Default directory is not under /tmp
  - XDG_DATA_HOME respected
  - Directory permissions are 0o700
  - Write refuses symlink targets
  - Read refuses symlink targets
  - Temp files use unpredictable names
  - Temp files cleaned up on failure
  - Full symlink attack scenario blocked
  - Atomic write correctness

## Breaking Change Note

The default `storage_dir` changes from `/tmp/strands/sessions/` to `~/.strands/sessions/`. Existing sessions stored in `/tmp` will not be automatically migrated. Users who explicitly pass `storage_dir` are unaffected.

### DIFF
--- a/strands-py/src/strands/session/file_session_manager.py
+++ b/strands-py/src/strands/session/file_session_manager.py
@@ -3,7 +3,9 @@
 import json
 import logging
 import os
+import platform
 import shutil
+import stat
 import tempfile
 from typing import TYPE_CHECKING, Any, cast
 
@@ -52,13 +54,35 @@ class FileSessionManager(RepositorySessionManager, SessionRepository):
         Args:
             session_id: ID for the session.
                 ID is not allowed to contain path separators (e.g., a/b).
-            storage_dir: Directory for local filesystem storage (defaults to temp dir).
+            storage_dir: Directory for local filesystem storage.
+                Defaults to ~/.strands/sessions/ (user-private) instead of /tmp/
+                to prevent symlink attacks and session tampering by other local users.
             **kwargs: Additional keyword arguments for future extensibility.
         """
-        self.storage_dir = storage_dir or os.path.join(tempfile.gettempdir(), "strands/sessions")
-        os.makedirs(self.storage_dir, exist_ok=True)
+        self.storage_dir = storage_dir or self._default_storage_dir()
+        os.makedirs(self.storage_dir, mode=0o700, exist_ok=True)
 
         super().__init__(session_id=session_id, session_repository=self)
+
+    @staticmethod
+    def _default_storage_dir() -> str:
+        """Return a user-private default storage directory.
+
+        Uses platform-appropriate conventions:
+        - Linux/macOS: ~/.strands/sessions/ (or $XDG_DATA_HOME/strands/sessions/)
+        - Windows: %LOCALAPPDATA%/strands/sessions/
+
+        This avoids the world-writable /tmp directory which is vulnerable to
+        symlink attacks and session tampering by other local users.
+        """
+        if platform.system() == "Windows":
+            base = os.environ.get("LOCALAPPDATA", os.path.expanduser("~"))
+        else:
+            base = os.environ.get("XDG_DATA_HOME", os.path.join(os.path.expanduser("~"), ".strands"))
+            # If XDG_DATA_HOME is set, nest under strands/
+            if "XDG_DATA_HOME" in os.environ:
+                base = os.path.join(base, "strands")
+        return os.path.join(base, "sessions")
 
     def _get_session_path(self, session_id: str) -> str:
         """Get session directory path.
@@ -106,7 +130,13 @@ class FileSessionManager(RepositorySessionManager, SessionRepository):
         return os.path.join(agent_path, "messages", f"{MESSAGE_PREFIX}{message_id}.json")
 
     def _read_file(self, path: str) -> dict[str, Any]:
-        """Read JSON file."""
+        """Read JSON file with symlink protection."""
+        # Refuse to read through symlinks (prevents session data injection)
+        if os.path.islink(path):
+            raise SessionException(
+                f"Refusing to read symlink at {path}. "
+                "This may indicate a symlink attack or session tampering."
+            )
         try:
             with open(path, encoding="utf-8") as f:
                 return cast(dict[str, Any], json.load(f))
@@ -114,13 +144,35 @@ class FileSessionManager(RepositorySessionManager, SessionRepository):
             raise SessionException(f"Invalid JSON in file {path}: {str(e)}") from e
 
     def _write_file(self, path: str, data: dict[str, Any]) -> None:
-        """Write JSON file."""
-        os.makedirs(os.path.dirname(path), exist_ok=True)
-        # This automic write ensure the completeness of session files in both single agent/ multi agents
-        tmp = f"{path}.tmp"
-        with open(tmp, "w", encoding="utf-8", newline="\n") as f:
-            json.dump(data, f, indent=2, ensure_ascii=False)
-        os.replace(tmp, path)
+        """Write JSON file atomically with symlink protection.
+
+        Uses tempfile.mkstemp() for unpredictable temp file names and checks
+        for symlinks at the target path to prevent symlink-following attacks.
+        """
+        dir_path = os.path.dirname(path)
+        os.makedirs(dir_path, mode=0o700, exist_ok=True)
+
+        # Refuse to write if the target path is a symlink (prevents symlink attacks)
+        if os.path.islink(path):
+            raise SessionException(
+                f"Refusing to write to symlink at {path}. "
+                "This may indicate a symlink attack."
+            )
+
+        # Use mkstemp for unpredictable temp file name in the same directory
+        fd, tmp_path = tempfile.mkstemp(dir=dir_path, prefix=".strands_", suffix=".tmp")
+        try:
+            with os.fdopen(fd, "w", encoding="utf-8", newline="\n") as f:
+                json.dump(data, f, indent=2, ensure_ascii=False)
+            os.replace(tmp_path, path)
+        except BaseException:
+            # Clean up temp file on any failure
+            try:
+                os.unlink(tmp_path)
+            except OSError:
+                pass
+            raise
+
 
     def create_session(self, session: Session, **kwargs: Any) -> Session:
         """Create a new session."""
@@ -129,9 +181,9 @@ class FileSessionManager(RepositorySessionManager, SessionRepository):
             raise SessionException(f"Session {session.session_id} already exists")
 
         # Create directory structure
-        os.makedirs(session_dir, exist_ok=True)
-        os.makedirs(os.path.join(session_dir, "agents"), exist_ok=True)
-        os.makedirs(os.path.join(session_dir, "multi_agents"), exist_ok=True)
+        os.makedirs(session_dir, mode=0o700, exist_ok=True)
+        os.makedirs(os.path.join(session_dir, "agents"), mode=0o700, exist_ok=True)
+        os.makedirs(os.path.join(session_dir, "multi_agents"), mode=0o700, exist_ok=True)
 
         # Write session file
         session_file = os.path.join(session_dir, "session.json")
@@ -162,8 +214,8 @@ class FileSessionManager(RepositorySessionManager, SessionRepository):
         agent_id = session_agent.agent_id
 
         agent_dir = self._get_agent_path(session_id, agent_id)
-        os.makedirs(agent_dir, exist_ok=True)
-        os.makedirs(os.path.join(agent_dir, "messages"), exist_ok=True)
+        os.makedirs(agent_dir, mode=0o700, exist_ok=True)
+        os.makedirs(os.path.join(agent_dir, "messages"), mode=0o700, exist_ok=True)
 
         agent_file = os.path.join(agent_dir, "agent.json")
         session_data = session_agent.to_dict()
@@ -263,7 +315,7 @@ class FileSessionManager(RepositorySessionManager, SessionRepository):
         """Create a new multiagent state in the session."""
         multi_agent_id = multi_agent.id
         multi_agent_dir = self._get_multi_agent_path(session_id, multi_agent_id)
-        os.makedirs(multi_agent_dir, exist_ok=True)
+        os.makedirs(multi_agent_dir, mode=0o700, exist_ok=True)
 
         multi_agent_file = os.path.join(multi_agent_dir, "multi_agent.json")
         session_data = multi_agent.serialize_state()
--- a/strands-py/src/strands/session/file_session_manager.py
+++ b/strands-py/src/strands/session/file_session_manager.py
@@ -75,12 +75,17 @@ class FileSessionManager(RepositorySessionManager, SessionRepository):
         symlink attacks and session tampering by other local users.
         """
         if platform.system() == "Windows":
-            base = os.environ.get("LOCALAPPDATA", os.path.expanduser("~"))
+            base = os.path.join(os.environ.get("LOCALAPPDATA", os.path.expanduser("~")), "strands")
         else:
             base = os.environ.get("XDG_DATA_HOME", os.path.join(os.path.expanduser("~"), ".strands"))
             # If XDG_DATA_HOME is set, nest under strands/
             if "XDG_DATA_HOME" in os.environ:
                 base = os.path.join(base, "strands")
+            # Per the XDG spec, ignore an empty or non-absolute XDG_DATA_HOME:
+            # a relative value would place sessions under the (possibly shared)
+            # process CWD, reopening the symlink/tampering class this fix closes.
+            if not os.path.isabs(base):
+                base = os.path.join(os.path.expanduser("~"), ".strands")
         return os.path.join(base, "sessions")
 
     def _get_session_path(self, session_id: str) -> str:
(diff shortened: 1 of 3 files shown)


yonib05:

manus-use:
> @strandly-the-agent can you review this PR?

Hi @yonib05, right now it's Waiting for review: manual-approval needs approval to start deploying changes.

manus-use:
> @strandly-the-agent can you review this PR?

Hi @yonib05, could you also look at this vuln: https://github.com/strands-agents/harness-sdk/security/advisories/GHSA-3w55-933p-8pv5

strands-py/src/strands/session/file_session_manager.py
```diff
@@ -3,6 +3,7 @@
 import json
 import logging
 import os
+import platform
 import shutil
 import tempfile
```
strandly-the-agent, line 7: 🔴 **`stat` is imported but never used in this file** (it's only used in the test file). `ruff` flags this as `F401`, and the Lint CI job (`hatch fmt --linter --check`) runs `ruff check` — so this fails CI as-is.

```suggestion
```

strands-py/src/strands/session/file_session_manager.py
```diff
@@ -52,14 +53,36 @@ def __init__(
...
+        Uses platform-appropriate conventions:
+        - Linux/macOS: ~/.strands/sessions/ (or $XDG_DATA_HOME/strands/sessions/)
+        - Windows: %LOCALAPPDATA%/strands/sessions/
+
+        This avoids the world-writable /tmp directory which is vulnerable to
+        symlink attacks and session tampering by other local users.
+        """
+        if platform.system() == "Windows":
+            base = os.environ.get("LOCALAPPDATA", os.path.expanduser("~"))
+        else:
```
strandly-the-agent: 🟡 **Windows drops the `strands` namespace.** This returns `%LOCALAPPDATA%\sessions`, but the docstring above (`:73`) and both Linux branches nest under a `strands`-named dir. Beyond the doc/code mismatch, it dumps a bare `sessions/` into the shared `AppData\Local` root (collision-prone).

```suggestion
            base = os.path.join(os.environ.get("LOCALAPPDATA", os.path.expanduser("~")), "strands")
```

strandly-the-agent (commented):
**Review: changes requested** 🟠 _(posted as a comment, not a hard block — a human owns the merge call)_

Solid, genuinely well-tested security fix — `mkstemp` + `O_EXCL` really does close GHSA-3w55-933p-8pv5 (the predictable `/tmp/*.tmp` symlink target is gone), and moving off `/tmp` is the right instinct. Three things to land it, then some defense-in-depth on the *supporting* path logic that's worth a look. I verified everything below against `pr2937` @ `29ff3a8`.

**Must-touch to land:**
- 🔴 `import stat` is unused → fails the Lint CI job (`ruff` F401). _(inline)_
- 🟡 Windows default drops the `strands` namespace — code returns `%LOCALAPPDATA%/sessions`, but the docstring (and both Linux branches) promise `.../strands/sessions`. _(inline)_
- 🟡 A non-absolute/empty `XDG_DATA_HOME` writes sessions into the **CWD** — reopening the exact vuln class this PR fixes. _(inline)_
- 🟢 Needs the `api/needs-review` label — this changes a public default's behavior (CI-enforced gate in `.github/workflows/api-review-label.yml`). Moderate change → one API-reviewer sign-off, not a full API meeting.

<details>
<summary><b>Verification ledger, should-fix / defense-in-depth items, nits & open questions</b></summary>

#### ✅/🔴 Verified
| Check | Result |
|---|---|
| Branch / SHA | ✅ `pr2937` @ `29ff3a8` |
| Unit tests (symlink file + full session suite) | ✅ 50 / 138 pass |
| `ruff check` on changed src | 🔴 1 error — `F401 stat imported but unused` (`file_session_manager.py:8`) |
| Windows path missing `strands` | 🔴 confirmed — code joins `LOCALAPPDATA`→`sessions` (`:78-79,:85`) vs docstring `:73` |
| Non-absolute `XDG_DATA_HOME` → relative path | 🔴 reproduced — `''`→`strands/sessions`, `'rel'`→`rel/strands/sessions` |
| `api/needs-review` gate | ✅ label names confirmed in the workflow |
| Original advisory (predictable `/tmp .tmp`) | ✅ closed by `mkstemp` (random name, `O_EXCL`, `0o600`) |

#### 🟡 Should-fix
**Test — the core `mkstemp` hardening has zero regression protection.** `test_write_uses_unpredictable_temp` (`tests/.../test_file_session_symlink_protection.py:102`) only asserts content is correct + no leftover `.tmp`. Mutation check: reverting `mkstemp` back to a predictable `f"{path}.tmp"` still passes all 11 tests. Worth an assertion that actually pins name unpredictability (e.g. patch `tempfile.mkstemp` and assert the temp name isn't derivable from `path`).

#### 🔍 Defense-in-depth (conditional — the "is the dir really private?" premise the symlink checks lean on)
- **Leaf-only `islink` check → parent-directory symlink bypass** (read *and* write). `os.path.islink(path)` (`:135`, `:156`) inspects only the final component; a symlinked *parent* redirects both `_read_file`/`_write_file` outside storage (reproduced). Exploitable only when the storage tree is attacker-writable — which the two items below can cause. Robust fix: `realpath`/`O_NOFOLLOW`-validate ancestors, and add a parent-symlink test.
- **`makedirs(mode=0o700, exist_ok=True)` is a no-op on a pre-existing looser dir** (`:63`, `:153`, …). A pre-existing `0o777` `~/.strands/sessions` (older version, backup restore, `$HOME` group-traversable) stays `0o777` (reproduced). Consider an explicit `os.chmod` gated on an ownership check; `test_storage_dir_permissions` (`test:57`) only covers a freshly-created dir. _(umask is a non-issue — `0o700` only tightens.)_

#### ⚪ Nits (no CI impact — mentioning once, not inlining)
Formatting drift (double blank `:175-176`; `ruff format` isn't wired to CI); `HOME` unset → a literal `~` dir in CWD (`:79,:81`); `_write_file`'s `islink` check is largely fail-loud belt-and-suspenders since `os.replace` doesn't follow a dst symlink — so `test_symlink_attack_scenario` (`test:134`) passes for the right reason but not the one it implies; leftover `.strands_*.tmp` on SIGKILL; default uses `~/.strands` rather than the XDG default `~/.local/share/strands`; thinner test gaps (`os.replace`-failure cleanup path, TOCTOU-on-read, no end-to-end coverage through `create_session`/`read_session`).

#### ❓ Questions (design calls for you / maintainers)
1. **Migration UX:** users on the old `/tmp` default will silently stop seeing prior sessions after upgrade — no error/log/fallback. Breaking is fine for a security fix, but silent is the wrong failure mode. Warn-on-legacy-detected + a CHANGELOG/advisory note (and *not* auto-migrating, since that'd re-import possibly-tampered data)?
2. **Scope:** land the parent-dir/dir-perms hardening here, or fast-follow and note the residual threat model (attacker with write access to the storage tree) in the advisory?
3. **Convention:** intentional to introduce a `~/.strands` dir SDK-wide vs. aligning to `~/.local/share/strands`?

</details>

strands-py/src/strands/session/file_session_manager.py
```diff
@@ -52,14 +53,36 @@ def __init__(
...
+            base = os.environ.get("XDG_DATA_HOME", os.path.join(os.path.expanduser("~"), ".strands"))
+            # If XDG_DATA_HOME is set, nest under strands/
+            if "XDG_DATA_HOME" in os.environ:
+                base = os.path.join(base, "strands")
+        return os.path.join(base, "sessions")
```
strandly-the-agent: 🟡 **A non-absolute `XDG_DATA_HOME` reopens this PR's own vuln class.** The value is trusted verbatim, but the XDG spec says a value that is empty or not absolute must be ignored. An empty `XDG_DATA_HOME` (common in some launchers/CI) yields a *relative* `strands/sessions` under the process CWD — potentially a shared/world-writable dir, exactly what moving off `/tmp` was meant to avoid.

```suggestion
            base = os.environ.get("XDG_DATA_HOME", os.path.join(os.path.expanduser("~"), ".strands"))
            # If XDG_DATA_HOME is set, nest under strands/
            if "XDG_DATA_HOME" in os.environ:
                base = os.path.join(base, "strands")
            # Per the XDG spec, ignore a non-absolute (or empty) XDG_DATA_HOME.
            if not os.path.isabs(base):
                base = os.path.join(os.path.expanduser("~"), ".strands")
```

manus-use:
Thanks for the thorough review @strandly-the-agent — all three must-touch items are addressed. Pushed as `ac67e7a5` + `ac8eb9dd`.

**🔴 `import stat` unused (F401)** — removed from `file_session_manager.py` (`ac67e7a5`). `ruff check` on the changed src is now clean.

**🟡 Windows drops the `strands` namespace** — fixed. `_default_storage_dir()` now returns `%LOCALAPPDATA%\strands\sessions`, matching the docstring and both Unix branches (no more bare `sessions/` in the shared `AppData\Local` root).

**🟡 Non-absolute/empty `XDG_DATA_HOME` reopens the vuln class** — fixed per the XDG spec: an empty or non-absolute value is now ignored and falls back to `~/.strands`, so sessions can never land in a relative CWD path.

```python
base = os.environ.get("XDG_DATA_HOME", os.path.join(os.path.expanduser("~"), ".strands"))
if "XDG_DATA_HOME" in os.environ:
    base = os.path.join(base, "strands")
# Per the XDG spec, ignore an empty or non-absolute XDG_DATA_HOME.
if not os.path.isabs(base):
    base = os.path.join(os.path.expanduser("~"), ".strands")
```

**🟡 (should-fix) `test_write_uses_unpredictable_temp` had no real regression protection** — good catch, confirmed the mutation. It now spies on `tempfile.mkstemp` and asserts the temp name is not derivable from the target path, so reverting to the predictable `f"{path}.tmp"` fails.

New tests added:
- `test_non_absolute_xdg_data_home_ignored` — empty / `relative/data` / `./data` all fall back under home.
- `test_default_nests_under_strands_namespace` — cross-platform namespace guarantee.

Suite: **13 passed** in the symlink file (11 → 13), **140 passed** across `tests/strands/session/` (138 → 140). `ruff check` clean.

On the deeper defense-in-depth items (parent-directory `islink` bypass, `makedirs` no-op on a pre-existing looser dir) and the migration-UX / `~/.strands` vs `~/.local/share/strands` design questions — those are real but out of scope for closing GHSA-3w55-933p-8pv5 (they require an already-attacker-writable storage tree). Happy to fast-follow in a separate PR once a maintainer weighs in on the convention question (Q3) so I match the intended direction. I'll add the `api/needs-review` label note for the default-behavior change.

strands-py/src/strands/session/file_session_manager.py
```diff
@@ -52,14 +52,44 @@ def __init__(
         Args:
             session_id: ID for the session.
                 ID is not allowed to contain path separators (e.g., a/b).
-            storage_dir: Directory for local filesystem storage (defaults to temp dir).
+            storage_dir: Directory for local filesystem storage.
+                Defaults to ~/.strands/sessions/ (user-private) instead of /tmp/
+                to prevent symlink attacks and session tampering by other local users.
```
pgrayy: Nit: This is a PR type comment. The API docs should just state the current behavior as opposed to past behavior to avoid confusion and bloat.

strands-py/src/strands/session/file_session_manager.py
```diff
@@ -52,14 +52,44 @@ def __init__(
...
+        """Return a user-private default storage directory.
+
+        Uses platform-appropriate conventions:
+        - Linux/macOS: ~/.strands/sessions/ (or $XDG_DATA_HOME/strands/sessions/)
+        - Windows: %LOCALAPPDATA%/strands/sessions/
+
+        This avoids the world-writable /tmp directory which is vulnerable to
+        symlink attacks and session tampering by other local users.
+        """
+        if platform.system() == "Windows":
```
pgrayy: This only handles Windows. Also, there is no return.

strands-py/src/strands/session/file_session_manager.py
```diff
@@ -52,14 +52,44 @@ def __init__(
...
         """
-        self.storage_dir = storage_dir or os.path.join(tempfile.gettempdir(), "strands/sessions")
-        os.makedirs(self.storage_dir, exist_ok=True)
+        self.storage_dir = storage_dir or self._default_storage_dir()
+        os.makedirs(self.storage_dir, mode=0o700, exist_ok=True)
 
         super().__init__(session_id=session_id, session_repository=self)
 
+    @staticmethod
+    def _default_storage_dir() -> str:
```
pgrayy, line 66: If we are considering using the home directory, can we just use something like `Path.home()` so we don't have to look at the platform at all?

strands-py/src/strands/session/file_session_manager.py
```diff
@@ -106,21 +136,49 @@ def _get_message_path(self, session_id: str, agent_id: str, message_id: int) ->
         return os.path.join(agent_path, "messages", f"{MESSAGE_PREFIX}{message_id}.json")
 
     def _read_file(self, path: str) -> dict[str, Any]:
-        """Read JSON file."""
+        """Read JSON file with symlink protection."""
+        # Refuse to read through symlinks (prevents session data injection)
+        if os.path.islink(path):
```
pgrayy, line 122: This isn't atomic so technically leaves a small window for someone to bypass. Regardless, if running Strands in production, users should already have their filesystems locked down and so this check I would say is not absolutely critical.

strands-py/tests/strands/session/test_file_session_manager_symlink_protection.py
pgrayy, line 1: For consistency, I would recommend naming this file `test_file_session_manager_<DESCRIPTION>.py`. We want that prefix to match the src file it tests.

manus-use:
Thanks @pgrayy — addressed your comments in `1c3dd8c1` (rebased on top of your `327b218a`).

**`:66` / `:76` — use `Path.home()`, don't inspect the platform.** Agreed, this is much cleaner. `_default_storage_dir()` is now just:

```python
@staticmethod
def _default_storage_dir() -> str:
    """Return a user-private default storage directory.

    Sessions are stored under ``~/.strands/sessions/``, which is private to
    the current user and avoids the world-writable ``/tmp`` directory.
    """
    return str(Path.home() / ".strands" / "sessions")
```

This drops the platform branching entirely, which also makes the earlier Windows-namespace and empty/non-absolute-`XDG_DATA_HOME` concerns moot — none of that logic exists anymore. (It also fixes the missing `return` you flagged at `:76`.)

**`:57` — docstring should state current behavior, not PR-type framing.** Done — the `storage_dir` param doc now just reads "Defaults to a user-private `~/.strands/sessions/` directory."

**test file `:1` — rename to match the src prefix.** Renamed `test_file_session_symlink_protection.py` → `test_file_session_manager_symlink_protection.py` (via `git mv`, history preserved). The XDG/Windows-specific tests are gone; replaced with `test_default_is_strands_sessions` (asserts exactly `~/.strands/sessions`) alongside the not-in-tmp / under-home / perms and the full symlink-protection coverage.

**`:141` — leaf `islink` check isn't atomic (TOCTOU).** Agreed it's not fully atomic and, as you say, not critical when the filesystem is locked down in production. Left as belt-and-suspenders; happy to drop it or upgrade to `O_NOFOLLOW` if you'd prefer.

⚠️ One heads-up: your `327b218a` "Apply suggestions from code review" seems to have hit a GitHub multi-suggestion collision — it landed a non-parsing `_default_storage_dir()` (no `return`, duplicated `base = …` lines, `else:` dropped) **and removed `import tempfile`** even though `_write_file` still calls `tempfile.mkstemp`, so `main`-of-branch would `NameError` at write time. My commit rewrites that whole region, so it's fully superseded and `import tempfile` is restored — just flagging so the diff makes sense.

Verified: `ruff check` clean; `tests/strands/session/` **138 passing**; symlink file **11 passing**.

Unshure (approved):

manus-use:
Hi @Unshure, thank you for merging the fix. Could you publish [GHSA-3w55-933p-8pv5](https://github.com/strands-agents/harness-sdk/security/advisories/GHSA-3w55-933p-8pv5) so downstream users, distributions, and vulnerability scanners can track the issue consistently?

PR #2939, opened 2026-06-24
fix(typescript): allow interface return types from tool callbacks

chaynabors:
Sorry, this is agent generated and escaped my fork. Please ignore.


chaynabors:
edit: deleted

PR #2940, opened 2026-06-24
chore(workflow): added release workflow

poshinchen:
## Description
[Python example](https://github.com/poshinchen/sdk-python/actions/runs/28535730229)
[Typescript example](https://github.com/poshinchen/sdk-python/actions/runs/28540419348)

Manually-triggered, per-package release pipeline. Pins `main` to a SHA, runs gates against it, drafts release notes from the commit log, pauses on a reviewer approval, creates the tag + GH release, then publishes to PyPI / NPM via OIDC trusted publishing. The existing `python-pypi-publish-on-release.yml` / `typescript-npm-publish-on-release.yml` workflows are demoted to `workflow_dispatch`-only manual fallback paths.

Two design properties:

- **Artifact identity**: the smoke-test workflow uploads its wheel / tarball; the inspect step re-uploads it as the verified bundle; publish downloads that bundle. The bytes a reviewer approves are the bytes that ship — no second build between approval and publish.
- **Fork-testability**: registry publish is the only step requiring upstream-only trust. Fork the repo, dispatch with `dry_run: false`, and the whole pipeline runs end-to-end — only `publish-pypi` / `publish-npm` fails (registry rejects the fork's OIDC token). The build artifact is on the run page for local install + smoke testing. This is how we iterate on the release workflows themselves.

### Pipeline (release-python.yml / release-typescript.yml)

```
scan-commits
├─► py-test-lint / ts-check / ts-test       (parity with old publish-on-release gates)
├─► integ (python / typescript)              (auto-skips on forks; opt-in via run_integ_tests)
├─► security-audit
└─► package-build                            (uploads pypi-build-output / npm-build-output)
        └─► inspect                          (twine check, list contents, re-upload as pypi-dist-bundle / npm-pack-bundle)
                └─► draft-notes              (drafts release notes from the commit log)
                        └─► approve-release  (auto-passes on forks with no env reviewers)
                                └─► create-gh-release            (tags the pinned SHA; skipped on dry runs)
                                        └─► publish-pypi / publish-npm  ◄── THE step that fails on forks
```

- `scan-commits` pins `release_sha = origin/main`, validates the typed version, rejects duplicate / non-monotonic / pre-existing tags.
- `create-gh-release` runs **before** publish so the tag is the source of truth — a publish failure leaves the tag intact and only the publish job needs re-running.
- Integ tests auto-skip on forks; opt-in via `run_integ_tests: true` for fork users with their own AWS credentials.
- `dry_run: true` (the default) skips approval + tag + publish; build/inspect/notes still run so the artifact can be reviewed on the run page.

### Supporting changes

- **`.github/scripts/group-release-notes.py`** (new) — drafts the Python release notes by grouping commits into sections by conventional-commit type (`feat`, `fix`, `docs`, ...), keeping each commit's original subject line. `release-typescript.yml` still drafts notes via `git shortlog`.
- **`python-security-audit.yml`** (new) — `pip-audit` against the full dependency closure. Mirrors `typescript-security-audit.yml`.
- **`python-test-package-build.yml`** (new) — wheel install + import in an out-of-tree venv. Uploads dist as `pypi-build-output` for the inspect step.
- **`typescript-test-package-pack.yml`** — also uploads the packed tarball.
- **`python-integration-test.yml` / `typescript-integration-test.yml`** — added `workflow_call(ref)` + a `validate-call-ref` job that rejects `refs/pull/*` / `owner:branch` refs. Dropped the prior `repository.fork == true` hard-reject; a fork using its own secrets on its own code is a legitimate use case (the ref-shape case statement is the real protection against ref-injection).
- **`python-pypi-publish-on-release.yml` / `typescript-npm-publish-on-release.yml`** — trigger flipped to `workflow_dispatch(tag)`. They no longer fire on `release: published` (that path is owned by the new release workflows). Header banner marks them as emergency fallback.

### One-time setup before merging

- Create environments in repo settings:
  - `release-gate`: `required_reviewers`, deployment branches → `main` only.
  - `pypi`, `npm`: allow the release workflows.
- Register the release workflows as trusted publishers:
  - PyPI for `strands-agents`: workflow `release-python.yml`, environment `pypi`.
  - npmjs.com for `@strands-agents/sdk`: workflow `release-typescript.yml`, environment `npm`.
- Leave existing trusted-publisher entries for the fallback workflows in place.

## Documentation PR

N/A — release pipeline is contributor-facing only.

## Type of Change

Other: CI / release infrastructure.

## Testing

- [x] `actionlint` + `shellcheck` pass on all modified workflows.
- [x] `group-release-notes.py` exercised against a real commit range; output grouped by type with prefixes preserved.
- [x] No Python source files changed; lint/format unaffected.

## Checklist

----


.github/workflows/release.yml
```diff
@@ -0,0 +1,755 @@
+name: "Release"
+
+# Manually-triggered release pipeline. Steps:
+#   1. Resolve main to a SHA. Every downstream job uses that SHA — never
+#      `main`. This keeps the pipeline deterministic if main moves while the
+#      release is in flight.
```
zastrowm: We might want to accept a SHA when we have a patch release or something

.github/workflows/release.yml
```diff
@@ -0,0 +1,755 @@
...
+#   2. Validate the maintainer-supplied version against the previous tag.
+#   3. Run lint + unit + integration tests, security audit, and the wheel /
+#      pack smoke test against the pinned SHA. The smoke test uploads the
+#      built artifact.
+#   4. Inspect the built artifact (twine check, list tarball contents) and
+#      re-upload it as the verified bundle.
+#   5. Draft AI-generated release notes (Bedrock). Falls back to a
+#      `git shortlog` summary if Bedrock is unavailable — releases are not
+#      blocked on AI availability.
```
zastrowm: Do we have an example of this?  I know that @yonib05 is adding release notes generates to the site - can we use that model and ditch the generated release notes for now?  I anticipate that we'll want to go the other direction (porting the release notes from the site -> github instead) so I'd rather not invest in these ai-generated notes for now unless we think it's good

.github/workflows/release.yml
```diff
@@ -0,0 +1,755 @@
...
+#   6. Pause on a deployment environment that requires 2 reviewer approvals.
```
zastrowm: ❤️

.github/workflows/release.yml
```diff
@@ -0,0 +1,755 @@
...
+# Registry publish (`publish-pypi` / `publish-npm`) is the only step that
+# requires PyPI / NPM to trust this repo as a trusted publisher. Everything
+# before it builds, verifies, and uploads the artifact using only generic
+# compute. A contributor can fork the repo, sync tags, run this workflow
+# with `dry_run: false` on their fork, and watch the entire pipeline run —
+# only `publish-pypi` / `publish-npm` will fail (PyPI / NPM reject the
+# fork's OIDC token because the fork isn't the registered publisher). The
+# build artifact is available on the run page for local install + smoke
+# testing.
+#
```
zastrowm: IMHO this belongs in a readme, but no in the workflow

```suggestion
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,755 @@
...
+# This is how we iterate on release.yml itself. To test on a fork:
+#   1. `git fetch upstream --tags && git push origin --tags`   (one-time)
+#   2. Actions → Release → Run workflow on your fork, dry_run: false
+#   3. Expected: green up to publish, red at publish, skipped after.
+#
+# By default the integ tests are SKIPPED on forks so a fork user without AWS
+# credentials isn't blocked at the integ step. If your fork has its own
+# AWS_ROLE_ARN + STRANDS_INTEG_TEST_ROLE secrets configured, set
+# `run_integ_on_fork: true` to exercise the integ tests too.
```
zastrowm: Should this be an input to the workflow instead?  (defaulting to false) E.g. "run_integ_tests"

.github/workflows/release.yml
```diff
@@ -0,0 +1,755 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      package:
+        description: 'Package to release'
+        required: true
+        type: choice
+        options:
+          - python
+          - typescript
```
zastrowm: I think we could simplify this quite a bit if we had duplicate workflows - trying to combine them just adds a lot of branching - does that get us anything?

Separate workflows would let us not have to have a lot of conditional steps + assume python or typescript versioning

.github/workflows/release.yml
```diff
@@ -0,0 +1,755 @@
...
+    environment:
+      # Configure required_reviewers >= 2 on this environment in repo
+      # settings. GitHub will pause this job until that many distinct
+      # approvers sign off.
+      #
+      # On a fork: the fork's copy of this environment has no reviewer
+      # config, so the job auto-passes. That's intentional — fork runs use
+      # this as a passthrough on the way to publish-pypi / publish-npm,
+      # which is the step that actually fails on a fork.
+      name: release-${{ inputs.package }}
```
zastrowm: I think we could just hardcode this to be `release-gate`; doesn't have to be package specific I think?

.github/workflows/release.yml
```diff
@@ -0,0 +1,755 @@
...
+  # before the package goes live. Conventional ordering + better recovery:
+  # if publish-* fails afterwards, the tag still points at the right SHA
+  # and the publish job can be re-run on its own (the artifact is in
+  # pypi-dist-bundle / npm-pack-bundle). Doing publish first would mean a
+  # tag-creation failure leaves the package on PyPI with no git tag — a
+  # harder mess to clean up.
+  #
+  # On a fork: the job RUNS (so publish-* can `needs:` it without being
+  # skipped) but its `gh release create` step short-circuits, since we
+  # don't want a leftover release object on every fork-test run.
```
zastrowm: Actually I'd say keep it working on forks; I'd like to be able to test e2e as much as possible

.github/workflows/typescript-integration-test.yml
```diff
@@ -11,9 +11,76 @@ on:
...
+    # future caller cannot accidentally check out untrusted code with secrets
+    # loaded. Branches and SHAs pass; `refs/pull/*/merge` and cross-repo refs
+    # (`owner:branch`) fail.
+    #
+    # We intentionally do NOT block based on `github.event.repository.fork`.
+    # A fork running its OWN workflow with its OWN secrets is a legitimate
+    # use case (e.g. release.yml's `run_integ_on_fork: true` path for fork
+    # owners who have configured their own AWS_ROLE_ARN). Upstream secrets
+    # are not accessible from a fork's runner, so there's nothing to leak
+    # by allowing the call through.
```
zastrowm: > `github.event.repository.fork` 

is that implying that forks can trigger this in the main repo?  Or that when forks run this workflow (in their own repo) `github.event.repository.fork` is true? 

This comment is hard for me to understand what's going on

.github/workflows/typescript-npm-publish-on-release.yml
```diff
@@ -1,30 +1,47 @@
-name: "TypeScript: Publish NPM Package"
+name: "TypeScript: Publish NPM Package (manual fallback)"
+
+# ⚠️ Manual fallback — prefer release.yml.
```
zastrowm: We should just remove this;  our new release flow should be all in without the need for these anymore

.github/workflows/typescript-test-package-pack.yml
```diff
@@ -44,19 +49,50 @@ jobs:
...
+          # stale package.json version).
+          if [ -n "$PINNED_VERSION" ]; then
+            actual=$(tar -xOzf "$TARBALL_PATH" package/package.json | jq -r .version)
+            if [ "$actual" != "$PINNED_VERSION" ]; then
+              echo "::error::Packed tarball has version '$actual' but expected '$PINNED_VERSION'."
+              exit 1
+            fi
+            echo "Tarball version OK: $actual"
+          fi
+
```
zastrowm: Does this matter; it seems we introduced version adding and verification, but I don't think this is needed for the release?

.github/scripts/draft-release-notes.py
```diff
@@ -0,0 +1,151 @@
...
+
+Inputs (env):
+    PACKAGE         "python" or "typescript"
+    PREV_TAG        previous release tag (e.g. "python/v1.2.3")
+    NEW_REF         git ref or SHA being released (typically a pinned SHA)
+    BEDROCK_MODEL   inference profile / model id (defaults to a Claude Sonnet on Bedrock)
+    AWS_REGION      defaults to us-west-2 (matches strands-command.yml)
+
+Output (file in $GITHUB_WORKSPACE):
+    release-notes.md    markdown release notes
```
zastrowm: I would consider adding the release-notes stuff as a follow-up; given that we're adding release notes to the website, we can just point developers there and then iterate on release notes.

That would simplify this PR as it focuses it on automating the our existing steps instead of adding new ones

zastrowm (changes requested):

.github/scripts/draft-release-notes.py
poshinchen: This is to let the reviewers know what do we plan to release, but i'll remove them as of now!

.github/workflows/release.yml
poshinchen: kept it!

.github/workflows/python-test-package-build.yml
```diff
@@ -0,0 +1,83 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: '3.10'
+          cache: 'pip'
+
+      - name: Install build tooling
+        run: pip install --no-cache-dir build
+
+      - name: Build wheel + sdist
```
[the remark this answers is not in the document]
zastrowm: Did we validate in the fork that the correct version is used?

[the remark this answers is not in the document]
zastrowm: Ah, I see we do validation below too :)

.github/workflows/release-python.yml
```diff
@@ -0,0 +1,420 @@
...
+    secrets:
+      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+
+  # ── Integration tests ──────────────────────────────────────────────────
+  # Always runs on upstream. On a fork: opt-in via `run_integ_tests: true`
+  # (fork needs its own AWS credentials).
+  integ:
+    name: Python integration tests
+    needs: scan-commits
+    if: github.event.repository.fork != true || inputs.run_integ_tests == true
```
zastrowm, line 153: The first time we run these, let's ensure that we're running the integ tests

.github/workflows/release-python.yml
```diff
@@ -0,0 +1,420 @@
...
+  draft-notes:
+    name: Draft release notes (grouped by commit type)
+    needs:
+      - scan-commits
+      - test-lint
+      - integ
+      - security-audit
+      - package-build
+      - inspect
+    if: |
```
zastrowm, line 278: We really need this if statement?

.github/workflows/release-typescript.yml
```diff
@@ -0,0 +1,422 @@
...
+          fetch-depth: 0
+          fetch-tags: true
+          persist-credentials: false
+
+      - name: Resolve SHA, find previous tag, validate version
+        id: scan
+        env:
+          NEW_VERSION: ${{ inputs.version }}
+          SHA_INPUT: ${{ inputs.sha }}
+        run: |
```
zastrowm, line 62: Is all of this shared between the two?  Should it be extracted to a common flow?

strands-py/pyproject.toml
```diff
@@ -160,7 +160,7 @@ dependencies = [
     "pytest-asyncio>=1.0.0,<1.5.0",
     "pytest-timeout>=2.0.0,<3.0.0",
     "pytest-xdist>=3.0.0,<4.0.0",
-    "pytest-timeout>=2.0.0,<3.0.0",
+    "pytest-rerunfailures>=15.0,<16.0",
```
zastrowm, line 163: This doesn't seem like it belongs in this PR?

zastrowm (approved):

.github/workflows/release-python.yml
poshinchen: Yes the workflow will trigger the integration tests, I can set up the credentials and run integ tests on my fork too.

.github/workflows/release-python.yml
poshinchen: These make sure that the workflows will be blocked if the step fails then people will not accidentally merge and publish.

strands-py/pyproject.toml
poshinchen: This is used by the workflow to retry when the step fails(e.g. integ tests)

PR #1937, opened 2026-03-19
feat: added per-invocation idempotency support via idempotency_token 

BV-Venky:
## Description

Adds per-invocation idempotency support via an optional `idempotency_token` parameter. When a duplicate request arrives with the same token while the original is still running, the duplicate waits for the original to finish and receives the same result or error instead of raising `ConcurrencyException`.

### Behavior

- **Same token, in-flight:** duplicate waits on the original; receives the same `AgentResult`
  or re-raises the same exception. Only the final result is delivered - intermediate streaming events are not replayed.
- **Same token, already completed:** treated as a fresh request (token is cleared on completion).
- **Different token, in-flight:** falls through to the existing `ConcurrencyException` path.
- **No token:** existing behavior unchanged.
- **`UNSAFE_REENTRANT` mode:** token is ignored entirely.

### Parameters
* **`idempotency_token`**: Optional token for duplicate detection. Can be any hashable value (string, UUID, or the prompt itself).
* **Scope**: Only used in `THROW` mode; ignored in `UNSAFE_REENTRANT`.
* **Behavior**: If another invocation with the same token is in flight, the caller waits for it and receives the same result or error.

### Public API Changes and Code Example
`idempotency_token` is added to `__call__`, `invoke_async`, and `stream_async`
It is a purely additive, opt-in parameter without existing behavior changes.

```python
# Before
# Thread 1: running
agent("What's the weather?")
# Thread 2: ConcurrencyException ✗
agent("What's the weather?")

# After
# Thread 1: running
agent("What's the weather?", idempotency_token="req-123")
# Thread 2: waits, gets same result ✓
agent("What's the weather?", idempotency_token="req-123")

```

---
### Design Decisions

| Decision | Choice | Rationale |
| :--- | :--- | :--- |
| **Scope** | Orthogonal opt-in parameter, not a new invocation mode | Keeps `THROW` semantics intact; idempotency is per-call, not per-agent. |
| **Active mode** | `THROW` only | `UNSAFE_REENTRANT` allows concurrent invocations by design; retry deduplication only makes sense when concurrency is restricted. |
| **State structure** | Single `_inflight_idempotency_token` + `_inflight_invocation` | In `THROW` mode at most one invocation is in flight; a single variable suffices. Upgradeable to a dict if `UNSAFE_REENTRANT` support is added later without breaking changes. |
| **Threading primitives** | `threading.Lock` and `threading.Event` | `run_async()` creates separate threads with separate event loops; `asyncio` primitives do not work across them. |
[...]

Issue #1365, opened 2025-12-18
[FEATURE] Per-Session Idempotency


sss26:
### Problem Statement

My team has an AgentCore stands-based agent which we invoke via an AWS SDK client. This client has built in retry logic for various transient issues such as network errors and timeouts. 

In some cases, our client will send a request to our agent, and the request will make it through and kickoff the invocation, but the client performs a retry anyway (this is not uncommon behavior across clients). When this happens, the same request is sent a second time. Unfortunately, when this occurs, the particular user session that is being invoked twice enters in unrecoverable state, and all subsequent requests to the same session fail. 

This is due to getting the following exception: `An error occurred (ValidationException) when calling the ConverseStream operation: Expected toolResult blocks at messages.6.content for the following Ids: tooluse_YQlhVQVOSdGbSXxRz5ZoyQ, but found: tooluse_LmVFH7_rSp-QJwfc1WvVsw` (among other variations).

### Proposed Solution

It would be helpful if when the strands agent is invoked with the same request a second time, that it simply waits for the first request to finish before returning the response, rather than executing a second invocation.

### Use Case

Users would not have to implement their own idempotency logic on the client side to prevent this issue from happening. This will likely occur commonly across many use cases, as using clients with internal retry logic is typical.

zastrowm:
> ### Duplicate Detection: Approaches
> I'd like your feedback on which approach to implement.
> 
> **Approach A: Explicit `request_id` (precise, requires caller changes)**
> 
> ```
> # Caller provides a request_id
> result = agent("What's the weather?", request_id="req-abc-123")
> 
> # Retry with same request_id → waits for first, returns same result
> result = agent("What's the weather?", request_id="req-abc-123")
> 
> # Different request_id while in-flight → ConcurrencyException
> result = agent("Tell me a joke", request_id="req-def-456")
> 
> # No request_id → falls back to THROW behaviour
> result = agent("Hello")
> ```
> * **Pros**: No false positives. Caller explicitly sends request_id and we can identify that "this is a retry."
> * **Cons**: Requires caller to pass a `request_id`. Existing code needs changes to benefit.

Option A seems the most explicit and I *think* covers most of the use cases that I know of when wanting this behavior - it also has the benefit of enabling consumers to implement the hashing themselves to get the right behavior if they don't already have a request_id.

[...]

PR #2941, opened 2026-06-24
docs: document concurrent invocations and idempotency in the agent loop

BV-Venky:
## Description
Documentation Update for idempotency semantics in Agent Loop

## Related Issues

#1365 

##Related PRs
#1937  #2932 

## Type of Change
Documentation update

### DIFF
--- a/site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
+++ b/site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
@@ -309,7 +309,7 @@ Sometimes more than one request targets the same agent instance at once: a retri
   </Tab>
   <Tab label="TypeScript">
 
-  The TypeScript SDK also rejects overlapping invocations: invoking an agent that is already running throws `ConcurrentInvocationError`. It does not offer a configurable concurrency mode or idempotency-token deduplication of retries; both are available only in the Python SDK.
+  TypeScript rejects overlapping invocations: invoking an agent that is already running throws `ConcurrentInvocationError`. It does not offer a configurable concurrency mode or idempotency-token deduplication of retries; both are available only in the Python SDK.
   </Tab>
 </Tabs>
 
--- a/site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
+++ b/site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
@@ -245,6 +245,69 @@ When multiple caps trip simultaneously, the reported stop reason
 follows priority order: turns, then total tokens, then output
 tokens.
 
+## Concurrent Invocations
+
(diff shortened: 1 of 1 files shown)


BV-Venky:
CC: @zastrowm

zastrowm (dismissed):

site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
```diff
@@ -245,6 +245,69 @@ When multiple caps trip simultaneously, the reported stop reason
 follows priority order: turns, then total tokens, then output
 tokens.
 
+## Concurrent Invocations
+
+Sometimes more than one request targets the same agent instance at once: a retried API call, a double-submitted form, two web requests that happen to share an agent. An agent mutates its conversation history as each invocation runs, so by default it processes one invocation at a time and rejects overlap.
+
+:::note[Python SDK]
```
zastrowm: Can we use tabs here  and for TypeScript we can simply indicate "This feature is not supported in TS"

Right now it's a little cofusing when you have TS selected in the top-right TBH

zastrowm (commented):
Content looks good - can we simplify and move to tabs to make it obvious when someone has TS selected as the language in the nav

site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
BV-Venky: Done. Do check and let me know if any further changes are required

site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
```diff
@@ -245,6 +245,74 @@ When multiple caps trip simultaneously, the reported stop reason
...
+  from strands.types.agent import ConcurrentInvocationMode
+
+  agent = Agent(concurrent_invocation_mode=ConcurrentInvocationMode.UNSAFE_REENTRANT)
+  ```
+
+  This removes the single-invocation guard entirely: overlapping calls neither raise nor deduplicate, and an idempotency token is ignored. Nothing protects the conversation history from concurrent mutation, so interleaved invocations can corrupt it. Reach for this only when each invocation works on isolated state, or when you handle synchronization yourself. To run independent work in parallel, a separate agent per task is safer than sharing one.
+  </Tab>
+  <Tab label="TypeScript">
+
+  The TypeScript SDK also rejects overlapping invocations: invoking an agent that is already running throws `ConcurrentInvocationError`. It does not offer a configurable concurrency mode or idempotency-token deduplication of retries; both are available only in the Python SDK.
```
zastrowm: This ends up looking  a bit odd:

Image

Suggestion:

```suggestion
  TypeScript rejects overlapping invocations: invoking an agent that is already running throws `ConcurrentInvocationError`. It does not offer a configurable concurrency mode or idempotency-token deduplication of retries; both are available only in the Python SDK.
```

zastrowm (approved):

PR #1262, opened 2025-11-26
fix(litellm): generate toolUseId when missing from provider

Ratish1:
## Description
 LiteLLM returns null for the tool call id field, causing toolUseId to be None in the resulting event stream. This PR ensures that a valid UUID is generated and assigned to the tool call delta if the provider returns a missing or null ID.
Changes:
- Modified LiteLLMModel.stream to check for a missing id on the first tool delta and generate a new UUID  if needed.
- Added a unit test test_stream_generates_tool_call_id_when_null to verify this behavior.
## Related Issues

Fixes #1259 

## Documentation PR

 N/A

## Type of Change

Bug fix

## Testing

How have you tested the change?  Verify that the changes do not break functionality or introduce warnings in consuming repositories: agents-docs, agents-tools, agents-cli

- [x] I ran `hatch run prepare`

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [ ] I have updated the documentation accordingly
- [ ] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2949, opened 2026-06-24
fix(litellm): generate toolUseId when missing from provider

### LINKED ISSUE #1259: [BUG] Strands providing null value in id field of tool_calls
### Checks

### Strands Version

strands-agents[litellm]==1.10.0 , strands-agents-tools==0.2.0  

### Python Version

3.13.7

### Operating System

MacOS Tahoe 26.1

### Installation Method

pip

### Steps to Reproduce

1. Create StreamableHttp MCP Client (using with context management)
2. Create LiteLLM model with proxy server configuration and assign to model variable
3. List tools from MCP server and assign to tools variable
4. Create Agent and provide tools, model and system_prompt
5. Execute agent

### Expected Behavior

```
"tool_calls": [
                {
                    "function": {
                        "arguments": "{}",
                        "name": "name-of-the-tool"
                    },
                    **"id": "some-id",**
                    "type": "function"
                }
            ]
        },
```

### Actual Behavior

```
"tool_calls": [
                {
                    "function": {
                        "arguments": "{}",
                        "name": "name-of-the-tool"
                    },
                    **"id": null,**
                    "type": "function"
                }
            ]
        },
```

### Additional Context

_No response_

### Possible Solution

_No response_

### Related Issues

_No response_

chaynabors:
Supersedes #1262 with a clean rebase on current main.

LiteLLM returns null for the tool call id field on some providers, leaving toolUseId as None in the resulting event stream. The fix assigns a fresh call_<uuid4> when the first tool delta arrives without an id, before the content_start chunk, so downstream consumers see a stable identifier. Behavior is unchanged whenever a provider supplies an id.

Refs #1262, fixes #1259.

### DIFF
--- a/strands-py/src/strands/models/litellm.py
+++ b/strands-py/src/strands/models/litellm.py
@@ -5,6 +5,7 @@
 
 import json
 import logging
+import uuid
 from collections.abc import AsyncGenerator
 from typing import Any, TypeVar, cast
 
@@ -519,7 +520,11 @@ class LiteLLMModel(OpenAIModel):
             Formatted tool call chunks.
         """
         for tool_deltas in tool_calls.values():
-            yield self.format_chunk({"chunk_type": "content_start", "data_type": "tool", "data": tool_deltas[0]})
+            first_delta = tool_deltas[0]
+            if not first_delta.id:
+                first_delta.id = f"call_{uuid.uuid4()}"
+
+            yield self.format_chunk({"chunk_type": "content_start", "data_type": "tool", "data": first_delta})
 
             for tool_delta in tool_deltas:
                 yield self.format_chunk({"chunk_type": "content_delta", "data_type": "tool", "data": tool_delta})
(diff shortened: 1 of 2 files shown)


strands-py/src/strands/models/litellm.py
```diff
@@ -519,7 +520,11 @@ async def _process_tool_calls(self, tool_calls: dict[int, list[Any]]) -> AsyncGe
             Formatted tool call chunks.
         """
         for tool_deltas in tool_calls.values():
-            yield self.format_chunk({"chunk_type": "content_start", "data_type": "tool", "data": tool_deltas[0]})
+            first_delta = tool_deltas[0]
+            if not first_delta.id:
+                first_delta.id = f"call_{uuid.uuid4()}"
```
zastrowm, line 525: This looks like it fixes the tool_use side of the id, but how does the tool result get associated with the same tool use id?  Do we have to pair up the tool result with the generated id?  I'm a little concerned we're just trading one problem for another

opieter-aws (approved):

zastrowm (dismissed):

strands-py/src/strands/models/litellm.py
chaynabors: I don't think this is an issue. The generated id is mutated onto first_delta before any chunk is emitted, so it enters the message history. When the model produces its tool_result on the next turn, it references the id from history.

zastrowm: I completely blanked on the fact that we're the ones generating the response and thus we're responsible for the pairing.  Brain fart - sorry for the noise

PR #1759, opened 2026-02-24
fix(bedrock): deliver streaming chunks individually instead of in batches

shivaber:
## Description

Fix BedrockModel.stream() delivering chunks in batches instead of individually.

Replace `call_soon_threadsafe(queue.put_nowait)` with `run_coroutine_threadsafe(queue.put()).result()` and use a bounded queue (`maxsize=1`). This blocks the producer thread until each event is delivered to the event loop, preventing callback batching where all chunks were processed in a single event loop iteration.

**Root cause:** When Bedrock sends chunks in a burst (e.g., multiple chunks in one TCP segment), `call_soon_threadsafe` scheduled all `put_nowait` callbacks at once. The event loop processed them all before the `queue.get()` waiter ran, causing queue depths of 5-6 items and delivering all chunks simultaneously.

**After fix:** Queue depth is always 0 after each `get()`, ensuring chunks are yielded one at a time for proper real-time streaming.

## Related Issues

Closes #1523

## Type of Change

Bug fix

## Testing

- Added `test_stream_delivers_chunks_individually` regression test that verifies queue depth is 0 after every `get()` when chunks arrive in a burst
- All 110 existing bedrock tests pass
- Verified with reproduction script: before fix queue depth reached 5-6, after fix always 0

- [x] I ran `hatch run prepare`

## Checklist

- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [x] I have updated the documentation accordingly
⠋- [x] I have added an appropriate example to the documentation to outline the feature, o r no new docs are needed
Thinking...- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2950, opened 2026-06-24
fix(bedrock): deliver streaming chunks individually instead of in batches

### LINKED ISSUE #1523: [BUG] BedrockModel streaming delivers chunks in batches sometimes instead of individual stream events
### Checks

### Strands Version

1.21

### Python Version

3.13

### Operating System

macOS 12.6

### Installation Method

pip

### Steps to Reproduce

A customer's service use Swarm with strands agent in their [package](https://code.amazon.com/packages/PartnerCentralAgentPythonSidecar/blobs/mainline/--/src/partner_central_agent_python_sidecar/swarm/swarm_manager.py?__session=eyJwa2ciOiJQYXJ0bmVyQ2VudHJhbEFnZW50UHl0aG9uU2lkZWNhciIsInRhYnMiOlt7ImZpbGUiOiJQYXJ0bmVyQ2VudHJhbEFnZW50UHl0aG9uU2lkZWNhci90cmVlcy9tYWlubGluZSJ9LHsiZmlsZSI6InNyYy9wYXJ0bmVyX2NlbnRyYWxfYWdlbnRfcHl0aG9uX3NpZGVjYXIvc3dhcm0vc3dhcm1fbWFuYWdlci5weSJ9LHsiZmlsZSI6InNyYy9oZWxwZXJzL2NhbGxfd2l0aF9jcmVkcy5weSJ9XX0%3D) When using Strands SDK's BedrockModel for streaming responses, chunks arrive in batches rather than as individual stream events. This breaks the natural streaming experience and causes poor user experience in applications expecting real-time streaming.

We might need to look into this line of  https://github.com/strands-agents/sdk-python/blob/63e58aa83dbb63fab06c405cbeb2acaa500c9e32/src/strands/models/bedrock.py#L660

### Expected Behavior

- Chunks should arrive individually with natural pacing, similar to direct Bedrock API calls
- Streaming should feel smooth and real-time to end users

### Actual Behavior

- Multiple chunks arrive simultaneously in batches
- Streaming appears "bursty" rather than smooth
- Poor user experience compared to direct Bedrock converse_stream calls

### Additional Context

_No response_

### Possible Solution

_No response_

### Related Issues

_No response_

chaynabors:
Supersedes #1759 with a clean rebase on current main after the monorepo restructure.

When Bedrock delivers a burst of chunks in a single TCP segment, call_soon_threadsafe schedules every put_nowait on the same event-loop tick, so the queue accumulates all chunks before any consumer await runs and the stream feels batched. Switching the producer callback to run_coroutine_threadsafe with a maxsize=1 queue forces real backpressure between the boto3 worker thread and the asyncio consumer, so each chunk is delivered before the next is enqueued. The pattern matches what mcp_client.py already does and the regression test asserts queue depth stays at zero after every consumer get.

Refs #1759, fixes #1523. Original commit by @shivaber, carried forward to the strands-py/ layout.

### DIFF
--- a/strands-py/src/strands/models/bedrock.py
+++ b/strands-py/src/strands/models/bedrock.py
@@ -920,12 +920,10 @@ class BedrockModel(Model):
         """
 
         def callback(event: StreamEvent | None = None) -> None:
-            loop.call_soon_threadsafe(queue.put_nowait, event)
-            if event is None:
-                return
+            asyncio.run_coroutine_threadsafe(queue.put(event), loop).result()
 
         loop = asyncio.get_event_loop()
-        queue: asyncio.Queue[StreamEvent | None] = asyncio.Queue()
+        queue: asyncio.Queue[StreamEvent | None] = asyncio.Queue(maxsize=1)
 
         # Handle backward compatibility: if system_prompt is provided but system_prompt_content is None
         if system_prompt and system_prompt_content is None:
--- a/strands-py/tests/strands/models/test_bedrock.py
+++ b/strands-py/tests/strands/models/test_bedrock.py
@@ -934,6 +934,38 @@ async def test_stream(bedrock_client, model, messages, tool_spec, model_id, addi
 
 
 @pytest.mark.asyncio
+async def test_stream_delivers_chunks_individually(bedrock_client, model, messages):
+    """Test that stream delivers chunks one at a time, not in batches.
+
+    Regression test for https://github.com/strands-agents/sdk-python/issues/1523.
+    When the Bedrock stream produces chunks in a burst (no delay between them),
+    the consumer should still receive them individually rather than all at once.
+    """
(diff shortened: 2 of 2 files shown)


awsarron (changes requested):
@pgrayy had a comment at https://github.com/strands-agents/harness-sdk/pull/1759#discussion_r2879873029 that should be addressed or replied to

JackYPCOnline:
It was a request from internal team, it has been handled, close this PR instead.
Thanks Chay!

chaynabors:
Wait, before we close I think we still want this non?

chaynabors:
Going to address Patricks comment in the original while we decide.

chaynabors:
> I think I'm aligned with Jack here. If a customer asks for unbatched latency in the future or people start benchmarking latency then we can fix.

PR #2952, opened 2026-06-24
fix(bidi): respect role from BidiTextInputEvent

### LINKED ISSUE #2089: [BUG] Respect role from BidiTextInputEvent parameter instead of hardcoding 'user' role
### Checks

### Strands Version

1.34.1

### Python Version

3.13

### Operating System

n/a

### Installation Method

pip

### Steps to Reproduce

The bidi agent loop hardcodes "role": "user" when appending messages from a BidiTextInputEvent, ignoring the event's role property entirely. This means callers who construct a BidiTextInputEvent with role="assistant" (e.g., for injecting assistant context into the conversation history) have that role silently overwritten. The role parameter on BidiTextInputEvent exists specifically to support this — it should be used.

### Expected Behavior

The event role attribute should be respected in the bidi agent loop. 

### Actual Behavior

The hard coded value overrides the role. 

### Additional Context

_No response_

### Possible Solution
1 line fix - 

https://github.com/strands-agents/sdk-python/pull/2083

### Related Issues

_No response_

chaynabors:
The bidi agent loop hardcodes the appended message role to user when handling a BidiTextInputEvent, ignoring the event's role attribute entirely. Callers who construct an event with role="assistant" to inject context have that value silently overwritten. The fix swaps the literal for event.role at the single call site so the explicit role flows through.

Fixes #2089.

### DIFF
--- a/strands-py/src/strands/experimental/bidi/agent/loop.py
+++ b/strands-py/src/strands/experimental/bidi/agent/loop.py
@@ -141,7 +141,7 @@ class _BidiAgentLoop:
             await self._send_gate.wait()
 
         if isinstance(event, BidiTextInputEvent):
-            message: Message = {"role": "user", "content": [{"text": event.text}]}
+            message: Message = {"role": event.role, "content": [{"text": event.text}]}
             await self._agent._append_messages(message)
 
         await self._agent.model.send(event)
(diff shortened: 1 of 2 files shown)


awsarron (dismissed):
Please add test coverage

mkmeral (approved):

PR #2955, opened 2026-06-24
fix(tools): unwrap FieldInfo default to avoid pydantic warning

### LINKED ISSUE #1914: [BUG] @tool emits PydanticJsonSchemaWarning for Field(default_factory=list)

chaynabors:
The @tool decorator emitted a PydanticJsonSchemaWarning whenever a parameter used Field(default_factory=...) because _extract_annotated_metadata rebuilt the field with Field(default=param_default, ...), nesting the original FieldInfo as the default value. The fix detects when param_default is already a FieldInfo and rebuilds from its default, default_factory, and validation kwargs so constraints like ge, le, and pattern are preserved.

Fixes #1914.

### DIFF
--- a/strands-py/src/strands/tools/decorator.py
+++ b/strands-py/src/strands/tools/decorator.py
@@ -163,8 +163,13 @@ class FunctionToolMetadata:
         final_description = description
         if final_description is None:
             final_description = self.param_descriptions.get(param_name) or f"Parameter {param_name}"
-        # Create FieldInfo object from scratch
-        final_field = Field(default=param_default, description=final_description)
+        if isinstance(param_default, FieldInfo):
+            if param_default.description is None:
+                final_field = FieldInfo.merge_field_infos(param_default, FieldInfo(description=final_description))
+            else:
+                final_field = param_default
+        else:
+            final_field = Field(default=param_default, description=final_description)
 
         return actual_type, final_field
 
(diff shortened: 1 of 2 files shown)


awsarron (changes requested):
please add test coverage

awsarron (approved):

PR #2956, opened 2026-06-24
fix(graph): drop reasoning blocks before downstream node input

### LINKED ISSUE #2882: Graph forwards reasoning blocks into downstream node input

chaynabors:
Graph nodes forward the previous node's full message into the next node's input, including reasoning blocks that downstream nodes cannot consume. The fix filters reasoning content out of the forwarded message before invoking the next executor, leaving every other content block intact.

Fixes #2882.

### DIFF
--- a/strands-ts/src/multiagent/graph.ts
+++ b/strands-ts/src/multiagent/graph.ts
@@ -741,8 +741,9 @@ export class Graph implements MultiAgent {
     const deps: ContentBlock[] = []
     for (const edge of this.edges.filter((e) => e.target.id === node.id)) {
       const ns = state.node(edge.source.id)!
-      if (ns.content.length > 0) {
-        deps.push(new TextBlock(`[node: ${edge.source.id}]`), ...ns.content)
+      const forwarded = ns.content.filter((b) => b.type !== 'reasoningBlock')
+      if (forwarded.length > 0) {
+        deps.push(new TextBlock(`[node: ${edge.source.id}]`), ...forwarded)
       }
     }
 


poshinchen:
It's a duplicate of [this PR](https://github.com/strands-agents/harness-sdk/pull/2883), will close this one.

PR #2961, opened 2026-06-25
docs: add TESTING.md for python

opieter-aws:
## Description

The TypeScript SDK has a thorough `docs/TESTING.md` that `strands-ts/AGENTS.md` hard-delegates to ("you MUST follow"). The Python SDK had no equivalent — its testing guidance was ~40 lines inside `AGENTS.md` and nothing in `docs/`. The practical cost is that real, reusable test infrastructure stays invisible: the shared fixtures (`MockedModelProvider`, `MockHookProvider`, `MockAgentTool`, `MockedSessionRepository`) and the `conftest.py` async/AWS helpers (`agenerator`, `alist`, `generate`, `moto_*`) aren't documented anywhere, so contributors hand-roll their own mocks instead of reusing them. A few sharp edges also went unwritten — strict-asyncio mode silently skips an unmarked coroutine test, and asserting a whole `Message` dict when a test only cares about role/text is a recurring source of brittle breakage in `tests_integ/`.

This adds `strands-py/docs/TESTING.md` as the authoritative Python testing reference, mirroring the TypeScript guide's depth and topic coverage so the two SDKs stay aligned. It documents the fixtures quick-reference (with "reuse these, don't hand-roll"), the `tru_`/`exp_` assertion-pair naming, whole-object assertions with `unittest.mock.ANY` for volatile fields, the strict-asyncio marker requirement, and the flat-vs-class test-organization preference — all reflecting conventions the suite already follows rather than inventing new ones.

It also corrects a contradiction in `strands-ts/docs/TESTING.md`, which referenced a `tests_integ/` directory that doesn't exist; the integration tests live in `test/integ/` (as the same file states a few lines later). An agent following the authoritative doc would otherwise create files in the wrong place.

The companion `AGENTS.md` restructure that links to this new file is a separate PR; this one is split out so the doc it points at lands first.

## Related Issues

None — maintenance change surfaced by a convention audit of both SDKs.

## Documentation PR

Not applicable — these are contributor-facing `docs/` files in the SDK packages, not user-facing documentation under `site/`.

## Type of Change

Documentation update

## Testing

Documentation-only. Verified that every fixture, helper, and path named in the new guide exists in the tree (`tests/fixtures/*`, the `conftest.py` session-scoped helpers, strict-asyncio being the default), and that the corrected `test/integ/` path matches the real TypeScript integration-test directory.

## Checklist

----

### DIFF
--- a/strands-py/docs/TESTING.md
+++ b/strands-py/docs/TESTING.md
@@ -56,7 +56,9 @@ exp_result = {"role": "assistant", "content": [{"text": "hi"}]}
 assert tru_result == exp_result
 ```
 
-**Assert whole objects, not field-by-field.** A single equality on the whole structure documents the expected shape and catches unexpected extra fields. Mask only genuinely volatile fields (timestamps, generated ids, metadata) with `unittest.mock.ANY`:
+The right granularity depends on **who controls the expected shape**:
+
+**When you control the shape, assert the whole object — not field-by-field.** A single equality on the whole structure documents the expected shape and catches unexpected extra fields. Mask only genuinely volatile fields (timestamps, generated ids, metadata) with `unittest.mock.ANY`:
 
 ```python
 tru_message = result.message
@@ -64,7 +66,7 @@ exp_message = {"role": "assistant", "content": [{"text": "abc"}], "metadata": un
 assert tru_message == exp_message
 ```
 
-**Do not assert a full `Message` dict when the test only cares about part of it.** If the test is about the text or role, assert just that (`assert result.message["role"] == "assistant"`). Asserting the entire `Message` shape couples the test to fields it doesn't care about, so an unrelated change to the `Message` type breaks it — this is a recurring source of breakage in `tests_integ/`.
+**When the shape is an externally-evolving type, assert only the fields the test is about.** This is the natural exception to the rule above: a type like `Message` grows fields the test doesn't care about, so asserting the entire dict couples the test to unrelated churn. If the test is about the text or role, assert just that (`assert result.message["role"] == "assistant"`) — pinning the whole `Message` shape is a recurring source of breakage in `tests_integ/`.
 
 ## Test Organization
 
--- a/strands-py/docs/TESTING.md
+++ b/strands-py/docs/TESTING.md
@@ -1,0 +1,78 @@
+# Testing Guidelines - Strands Python SDK
+
+> **IMPORTANT**: When writing tests, you **MUST** follow the guidelines in this document. They keep tests consistent, maintainable, and resilient to unrelated changes.
+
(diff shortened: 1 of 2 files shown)


strands-py/docs/TESTING.md
```diff
@@ -0,0 +1,78 @@
+# Testing Guidelines - Strands Python SDK
+
+> **IMPORTANT**: When writing tests, you **MUST** follow the guidelines in this document. They keep tests consistent, maintainable, and resilient to unrelated changes.
+
+This document is the authoritative testing reference for the Python SDK. For general development guidance, see [AGENTS.md](../AGENTS.md). It is the counterpart to the TypeScript SDK's [docs/TESTING.md](../../strands-ts/docs/TESTING.md); where a convention is shared, the two should stay aligned.
```
pgrayy: I wouldn't worry about calling out TS here. I would recommend instead making this entirely Python specific. If there are rules to share among languages, we can place that in a root level TESTING.md that also calls out to these language specific mds.

strands-py/docs/TESTING.md
```diff
@@ -0,0 +1,78 @@
...
+tru_message = result.message
+exp_message = {"role": "assistant", "content": [{"text": "abc"}], "metadata": unittest.mock.ANY}
+assert tru_message == exp_message
+```
+
+**Do not assert a full `Message` dict when the test only cares about part of it.** If the test is about the text or role, assert just that (`assert result.message["role"] == "assistant"`). Asserting the entire `Message` shape couples the test to fields it doesn't care about, so an unrelated change to the `Message` type breaks it — this is a recurring source of breakage in `tests_integ/`.
+
+## Test Organization
+
+- **Default to flat, module-level `test_<behavior>` functions.** Most of the suite is flat.
```
pgrayy: More specifically, we try to do something along the lines of `test_<FUNCTION_NAME>_<DESCRIPTION>`. If testing a method, we typically do `test_<CLASS_NAME>_<METHOD_NAME>_<DESCRIPTION>`, although this has been tinkered with. Some other options are:
* Since the class name is typically in the module name, we can shorten to `test_<METHOD_NAME>_<DESCRIPTION>`.
* Alternatively, we can namespace and group tests by defining under a `class` scope.

I'm okay with any of these but it would be good to pick one and stick to it.

strands-py/docs/TESTING.md
```diff
@@ -0,0 +1,78 @@
+# Testing Guidelines - Strands Python SDK
```
pgrayy, line 1: Do we need to migrate/remove/update any rules pertaining to tests in AGENTS.md or other md files? I think we ended up describing them in a few places. I think it does make sense to centralize the rules here and we can link to it from those other files.

opieter-aws: Agreed, I am drafting another PR with the AGENTS.md reworks. Wanted to split it to avoid too large a PR (#2959)

mehtarac (approved):

PR #2968, opened 2026-06-25
feat: add bug-verify local-preview skill

JackYPCOnline:
## Motivation

Adds a local-preview skill for the new bug-verifier agent so a maintainer can
rehearse a verification in their IDE — and anticipate what the remote
`/strands bug-verify` agent will do — before running it on a real issue. The
agent itself ships in `strands-agents/devtools` (PR
[strands-agents/devtools#71](https://github.com/strands-agents/devtools/pull/71));
this PR is just the local mirror and its index entry.

## What's added

- **`.agents/skills/bug-verify/SKILL.md`** — local-preview skill. Body is
  byte-identical to the upstream Task Bug Verifier SOP; the header maps the
  SOP's runtime assumptions (issue snapshot, label/comment tools) to local
  `gh` equivalents.
- **`.agents/skills/README.md`** — adds an "Issue triage" section listing it.

Same pattern as the existing `strands-review` skill.

## Dependency

Merge after `strands-agents/devtools#71` (the agent SOP and `bug-verify`
command). There is no workflow in this PR — the agent is triggered by a
maintainer commenting `/strands bug-verify`, which is authorization-gated by
the existing `strands-command.yml`.


.agents/skills/bug-verify/SKILL.md
```diff
@@ -0,0 +1,298 @@
+---
+name: bug-verify
+description: Local preview of the strands-agents/devtools `/strands bug-verify` agent. Body is the upstream Task Bug Verifier SOP verbatim — do not paraphrase. Use when the user types `/bug-verify`, asks to "verify a bug" or "triage an issue", references a bug-report GitHub issue and wants reproduction + urgency assessment, or wants to anticipate what the remote auto-bug-verify GitHub Action will do to an issue. The agent inspects the implicated source, runs a reproduction in a code executor, scores urgency, applies triage labels, and records a verification report (it does not comment on the issue). Findings are close but not identical to the remote agent. Strongly prefer running this skill in a fresh-context subagent rather than inline — the SOP is long and the verdict is more reliable when it isn't entangled with the parent conversation's prior context.
```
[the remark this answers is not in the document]
JackYPCOnline: Fixed in ce5a3839 — aligned the frontmatter description with the two-case commenting behavior described elsewhere (sharing a derived reproduction, or stating it couldn't reproduce and requesting more info).

.github/workflows/auto-bug-verify.yml
```diff
@@ -0,0 +1,53 @@
...
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            await github.rest.actions.createWorkflowDispatch({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              workflow_id: 'strands-command.yml',
+              ref: 'main',
+              inputs: {
+                issue_id: String(context.payload.issue.number),
+                command: 'bug-verify',
```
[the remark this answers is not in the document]
JackYPCOnline: Investigated and reverted in 25622f2f. Confirmed the unknown-command behavior: the strands-command parser silently falls through to `refiner` mode (worse than a hard fail — every bug report would get confused refiner-style clarifying questions).

That said, every other devtools-dependent workflow in this repo (`auto-strands-review`, `strands-command`, `issue-labeler`, `docs-strands-command`, both integration test workflows) trusts `strands-agents/devtools@main` and coordinates merge order at PR review time. A runtime guard here would be inconsistent with the existing convention and would carry permanent always-true HTTP overhead after devtools#71 merges, for a one-time ordering concern. Replaced the guard with a header note documenting the dependency and pointing at the convention.

Merge order remains: devtools#71 must merge first.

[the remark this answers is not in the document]
JackYPCOnline: Heads up: the preflight you described here was reverted in 25622f2f after discussion. None of the other devtools-dependent workflows in this repo runtime-check devtools, so the guard was inconsistent and would have been permanent overhead post-merge. The dependency is now documented in the workflow header and coordinated at PR review time, matching the repo convention.

.github/workflows/auto-bug-verify.yml
```diff
@@ -0,0 +1,59 @@
...
+# Maintainers can also run it manually by commenting `/strands bug-verify`.
+
+on:
+  issues:
+    types: [opened]
+
+jobs:
+  trigger-bug-verify:
+    # Only bug reports: the bug_report.yml template pre-fills "[BUG] " as the title.
+    if: startsWith(github.event.issue.title, '[BUG]')
```
mehtarac: If a user manually types [bug] or [Bug] (not using the template autofill), the workflow won't trigger. Is this intentional? if so, it may be worth documenting. Though I think we should still go one step further and handle:
```
  if: startsWith(github.event.issue.title, '[BUG]') || startsWith(github.event.issue.title, '[bug]') || startsWith(github.event.issue.title, '[Bug]')
```

.agents/skills/bug-verify/SKILL.md
```diff
@@ -0,0 +1,298 @@
...
+  <One sentence: whether it reproduced, and whether via the reporter's own repro or one you derived.>
+
+  <details>
+  <summary>Reproduction script & output</summary>
+
+  ```python
+  # minimal repro
+  ```
+
+  ```
```
mehtarac, line 255: the SOP itself handles both Python and TypeScript SDKs but the excample code block for the reproduction template only has python. 

NIT: Worth also having a typescript code block so that a literal template-following agent doesn't use python blocks for typescript bugs

.github/workflows/auto-bug-verify.yml
JackYPCOnline: In our issues, Blank issue is maintainer only, users have to choose a template anyway. With being said, I think we don't need defensive code here.

.github/workflows/auto-bug-verify.yml
```diff
@@ -0,0 +1,59 @@
...
+# strands-agents/devtools@main, same convention as every other devtools-
+# dependent workflow in this repo (auto-strands-review, strands-command,
+# issue-labeler, etc. — none of them runtime-check devtools either; merge
+# order is coordinated at PR review time).
+#
+# Maintainers can also run it manually by commenting `/strands bug-verify`.
+
+on:
+  issues:
+    types: [opened]
```
yonib05: It should not be triggered on open; it should be triggered on something a maintainer does. Think of the case where someone cuts an issue with prompt injection and this auto triggers.

.github/workflows/auto-bug-verify.yml
```diff
@@ -0,0 +1,59 @@
...
+    permissions:
+      actions: write
+      contents: read
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger Strands Command Workflow
+        uses: actions/github-script@v9
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
```
yonib05: I am not sure if it is here or in the command, but it should snapshot the issue description vs pulling it dynamically. Think of the cause where bad actor opens a valid bug report, then the maintainer triggers this workflow, and then the bad actor quickly adds some propmpt injection while the workflow is running.

Whatever the maintainer approved needs to be what the agent works off of.

JackYPCOnline: I see, this is a valid secruity call out, will iterate on next revision.

.agents/skills/bug-verify/SKILL.md
JackYPCOnline: Good catch — addressed. The reproduction templates now note that the code block should match the affected SDK (a `python` block for Python bugs, a `typescript` block for TypeScript bugs), so a literal template-following agent won't wrap TypeScript in a python fence. Fixed in the SOP (strands-agents/devtools@51d0d54) and re-synced into this skill (d84ec95b).

JackYPCOnline:
After given more thoughts, I'd rather drop this PR and keep it dispatching to devtool agent where we can use command to trigger the agent. Same mechanism, same restrictions, same requirements.

PR #2987, opened 2026-06-27
Ignore malformed message filenames when listing sessions

CodingFeng101:
## What changed

`FileSessionManager.list_messages()` now skips files whose names match `message_*.json` but do not contain a numeric message id, instead of raising `ValueError` for the whole listing.

## Why

The message directory can contain backup or manually-created files such as `message_backup.json`. Before this change, one malformed filename prevented all valid session messages from being listed, even though valid `message_<id>.json` files were present.

## Validation

- `.\.venv\Scripts\python -m pytest strands-py/tests/strands/session/test_file_session_manager.py::test_list_messages_ignores_malformed_message_filename strands-py/tests/strands/session/test_file_session_manager.py::test_list_messages_all -q`
- `python -m ruff check strands-py/src/strands/session/file_session_manager.py strands-py/tests/strands/session/test_file_session_manager.py`
- `git diff --check`

### DIFF
--- a/strands-py/src/strands/session/file_session_manager.py
+++ b/strands-py/src/strands/session/file_session_manager.py
@@ -232,7 +232,11 @@ class FileSessionManager(RepositorySessionManager, SessionRepository):
         for filename in os.listdir(messages_dir):
             if filename.startswith(MESSAGE_PREFIX) and filename.endswith(".json"):
                 # Extract index from message_<index>.json format
-                index = int(filename[len(MESSAGE_PREFIX) : -5])  # Remove prefix and .json suffix
+                try:
+                    index = int(filename[len(MESSAGE_PREFIX) : -5])  # Remove prefix and .json suffix
+                except ValueError:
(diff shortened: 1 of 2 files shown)


strands-py/src/strands/session/file_session_manager.py
```diff
@@ -232,7 +232,11 @@ def list_messages(
         for filename in os.listdir(messages_dir):
             if filename.startswith(MESSAGE_PREFIX) and filename.endswith(".json"):
                 # Extract index from message_<index>.json format
-                index = int(filename[len(MESSAGE_PREFIX) : -5])  # Remove prefix and .json suffix
+                try:
+                    index = int(filename[len(MESSAGE_PREFIX) : -5])  # Remove prefix and .json suffix
```
Unshure, line 236: Can we do a more robust check here, like a regex match on file name? An incorrectly formatted filename will also break things

Unshure (changes requested):

PR #2989, opened 2026-06-27
fix(event_loop): log exception type instead of full traceback on cycle failure

yonib05:
## Description

The event loop logs a failed cycle in two places, `event_loop_cycle` and `_handle_model_execution`, with `logger.exception("cycle failed")`. `logger.exception` emits the full traceback at ERROR level. Those frames and the wrapped exception's arguments can include message content, model responses, and tool I/O, so the application log ends up carrying request payload detail that operators did not opt into.

This change replaces both calls with the structured-logging form already used for the model-retry path in the same file:

```python
logger.error("exception=<%s> | event loop cycle failed", type(e).__name__)
```

The failure is still surfaced fully: it is raised as `EventLoopException` (so callers see the original cause via `from e`) and recorded on the tracer span through `tracer.end_span_with_error`, which already receives the exception. The application logger now reports the exception type and a stable message, matching the existing `exception=<%s>` log call in `_handle_model_execution`.

## Related Issues

N/A

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

Added `test_cycle_exception_logs_exception_type_without_traceback`, which drives a cycle to failure and asserts the ERROR-level cycle-failure record names the exception type and carries no `exc_info`. Ran the Python event loop unit tests and ruff locally.

- [ ] I ran `hatch run prepare` (hatch was unavailable in this environment; ran the event loop unit tests with pytest and ran `ruff check` / `ruff format --check` on the changed files instead)

## Checklist
- [ ] I have updated the documentation accordingly (no documentation change required for this fix)
- [x] I have added comments to my code, particularly in hard-to-understand areas
- [x] I have checked my code and corrected any misspellings

### DIFF
--- a/strands-py/src/strands/event_loop/event_loop.py
+++ b/strands-py/src/strands/event_loop/event_loop.py
@@ -391,7 +391,7 @@ async def event_loop_cycle(
             tracer.end_span_with_error(cycle_span, str(e), e)
             # Handle any other exceptions
             yield ForceStopEvent(reason=e)
-            logger.exception("cycle failed")
+            logger.error("exception=<%s> | event loop cycle failed", type(e).__name__)
             raise EventLoopException(e, invocation_state["request_state"]) from e
 
 
@@ -645,7 +645,7 @@ async def _handle_model_execution(
 
     except Exception as e:
         yield ForceStopEvent(reason=e)
-        logger.exception("cycle failed")
+        logger.error("exception=<%s> | event loop cycle failed", type(e).__name__)
         raise EventLoopException(e, invocation_state["request_state"]) from e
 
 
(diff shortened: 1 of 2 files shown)


strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -645,7 +646,8 @@ async def _handle_model_execution(
 
     except Exception as e:
         yield ForceStopEvent(reason=e)
-        logger.exception("cycle failed")
+        logger.error("exception=<%s> | event loop cycle failed", type(e).__name__)
+        logger.debug("event loop cycle failed", exc_info=True)
```
JackYPCOnline, line 650: One nit is that in debug, user might see two `event loop cycle failed` , harmles tho.

JackYPCOnline (approved):

PR #2993, opened 2026-06-27
fix(session): validate snapshot scope and confine resolved paths to the base directory

yonib05:
## Description

The snapshot `scope` flows into both the FileStorage file path and the S3Storage object key, but unlike `sessionId` and `scopeId` it was never validated before being interpolated. This change validates `scope` against the set of recognized values and, for the filesystem backend, confirms the resolved path stays within the configured base directory before it is used.

`Scope` is a compile-time union (`'agent' | 'multiAgent'`), so today no public API feeds an out-of-set value, but the storage layer should not depend on that to keep a snapshot confined to its intended subtree.

Changes:
- Add a `validateScope` helper in `session/validation.ts`, alongside the existing identifier and UUID validators. It uses an allowlist set (not the identifier regex, since `multiAgent` contains an uppercase character).
- Apply `validateScope` in `FileStorage._getPath` and `S3Storage._getKey`.
- Add a base-directory containment check in `FileStorage._getPath` so any resolved path that lands outside the base directory is rejected (previously this guard existed only in `_getHistorySnapshotPath`).

## Type of Change

Bug fix

## Testing

Added regression tests asserting an unrecognized scope is rejected in both `FileStorage` and `S3Storage`, plus unit tests for `validateScope`. Ran the TypeScript SDK checks locally: `npm run build`, `npm run lint`, `npm run format:check`, `npm run type-check`, and the session unit tests all pass (123 session tests). The new tests fail when the validation is removed and pass with it.

## Checklist


yonib05:
The remaining consolidation item is addressed in dcfbbe18: the weaker `startsWith` guard in `_getHistorySnapshotPath` is removed, leaving `_getPath`'s resolve+sep containment as the single authoritative check. That commit also switches the allowlist lookup to `Object.hasOwn` so inherited `Object.prototype` keys are rejected, with tests. Merged main to pick up the fix for the unrelated `local-memory-store` test failure (#3020).

yonib05:

strands-ts/src/session/s3-storage.ts
```diff
@@ -66,10 +66,13 @@ export class S3Storage implements SnapshotStorage {
...
+   * Validates sessionId, scope, and scopeId before constructing the key. The scope
+   * allowlist confines the key to a known subtree, since none of the recognized scopes
+   * contain `/`, `..`, or a leading `/`.
    */
   private _getKey(location: SnapshotLocation, path: string): string {
     validateIdentifier(location.sessionId)
+    validateScope(location.scope)
     validateIdentifier(location.scopeId)
     const base = this._prefix ? `${this._prefix}/` : ''
     return `${base}${location.sessionId}/scopes/${location.scope}/${location.scopeId}/snapshots/${path}`
```
strandly-the-agent, line 78: 🟡 `_getKey` re-reads the **raw** `location.scope` here after `validateScope(location.scope)` on line 75 validated a *separate* coercion of it. `validateScope` doesn't return-and-canonicalize, so a `scope` whose `toString` returns `'agent'` on the first read (inside `Object.hasOwn`) and `'../../../../other-tenant'` on this read passes validation and produces a traversed object key. Confirmed with a repro:

```
p/victim/scopes/../../../../other-tenant/sid/snapshots/snapshot_latest.json
```

Unlike `FileStorage._getPath`, this key builder has no containment backstop, so the poisoned key flows into `PutObjectCommand`. Fix: have `validateScope` coerce once + return the value, and use it here — `const scope = validateScope(location.scope)` then interpolate `scope`. (Same applies to the re-read of `sessionId`/`scopeId` — pre-existing, fine as a follow-up.)

strandly-the-agent (commented):
**Reviewed at `5d87d6bb` (merge-base `3c01024a`). Solid, tightly-scoped defense-in-depth — one 🟡 worth a look before merge, otherwise good.**

✅ Verified: session unit tests green — file-storage 26, session-manager 50, s3-storage 35, validation 15 (**126**). The `+ sep` boundary check and the removal of the old `_getHistorySnapshotPath` guard are both sound (traced every caller of `_getPath`; no protection lost). `validateScope` is internal (not in the public `index.ts` surface) → no `needs-api-review` gate.

🟡 **The S3 `_getKey` allowlist isn't airtight against the exact threat this PR targets** — see inline on `s3-storage.ts`. `validateScope` coerces `location.scope` to a key *once* (inside `Object.hasOwn`), then `_getKey` re-reads the raw `location.scope` and coerces it *again* in the template literal. A value whose `toString` returns `'agent'` first and a traversal payload second passes validation and lands in the object key. Low real-world reachability (`scope` is a hardcoded `'agent'|'multiAgent'` literal in normal flow) — but a mutating object is *precisely* "a value that bypasses the type system," which is what `validateScope`'s docstring says it guards against, and the S3 docstring's "airtight allowlist" reasoning only holds for the *returned* string, not the re-read raw value.

<details>
<summary>Confirmed repro + fix (one line, closes it for both backends)</summary>

```js
let n = 0
const evil = { toString() { return n++ === 0 ? 'agent' : '../../../../other-tenant' } }
validateScope(evil)                                   // Object.hasOwn coerces -> 'agent' -> passes
`${base}${sid}/scopes/${evil}/${scopeId}/snapshots/${path}`  // 2nd coercion -> payload
// => p/victim/scopes/../../../../other-tenant/sid/snapshots/snapshot_latest.json
```

FileStorage happens to be safe by luck of API — `path.join(..., evil, ...)` throws `ERR_INVALID_ARG_TYPE` on the non-string arg before the containment check runs (and the `resolve`+`sep` check would catch a traversal string anyway). S3's `_getKey` is pure interpolation, so the poisoned key flows straight into `PutObjectCommand`.

**Fix** — coerce exactly once and return the canonical value; callers use the return:
```ts
// validation.ts
export function validateScope(scope: string): string {
  const s = String(scope)
  if (!Object.hasOwn(VALID_SCOPES, s)) {
    throw new Error(`Scope '${s}' is not a recognized session scope`)
  }
  return s
}
```
```ts
// s3-storage.ts _getKey (and file-storage.ts _getPath): use the returned value
const scope = validateScope(location.scope)
return `${base}${location.sessionId}/scopes/${scope}/${location.scopeId}/snapshots/${path}`
```
Verified: with the single-coercion + returned value, the key stays `.../scopes/agent/...`. Note this pattern generalizes — `_getKey` also re-reads `sessionId`/`scopeId` after validating (pre-existing), so `validateIdentifier` returning-and-using its value would harden those in the same motion. Reasonable to scope that to a follow-up.

⚪ Unrelated pre-existing limitation (no action needed): `path.resolve` doesn't resolve symlinks, so an attacker-planted symlink inside `_baseDir` would pass containment — a general limitation of resolve-based checks, consistent with the Python backend.
</details>

If the team considers non-string `scope` injection out of scope, the minimal alternative is softening the `_getKey` docstring so it doesn't overstate the guarantee. Nice fix either way — this is close.

<sub>🤖 Automated review by an experimental agent — a maintainer should confirm before merging.</sub>

PR #2994, opened 2026-06-28
fix(tools): load directory tools under a namespaced module key

yonib05:
## Description

When the tool registry loads a tool module from a directory (in `reload_tool` and `initialize_tools`), it registered the module in `sys.modules` under the tool's bare name. A tool whose filename matched an already-imported module (for example a tool named `json`) would replace that entry, leaving later code holding the tool module instead of the one it imported.

Tool modules are now loaded under a namespaced `sys.modules` key, so loading a tool never replaces an entry that another part of the process already imported under the same name.

Directory tools commonly import a sibling module from the same directory at the top level, which relied on that directory being on `sys.path` during load. To preserve that, the tool's directory is added to `sys.path` only while the module executes and removed immediately afterward (via `try`/`finally`), so the namespaced key and the sibling-import behavior coexist.

## Related Issues

N/A

## Testing

- Added regression tests covering both load paths: a tool whose name collides with a stdlib module does not replace that module, and a directory tool that imports a sibling module from the same directory loads successfully.
- `tests/strands/tools/` passes (479 tests).
- `ruff check`, `ruff format --check`, and `mypy` on `registry.py` pass.

## Checklist

- [x] I have added necessary documentation (if appropriate)
- [x] All new and existing tests pass


mkmeral:
@strandly-the-agent is this a problem that we should fix? isn't this the intended behavior of the programming languages? 

Give me devx before and after

strandly-the-agent:
**Short answer: yes, worth fixing — and no, this isn't the language's intended behavior.** It's the framework writing to a *process-global* namespace it doesn't own, as an invisible side effect of loading a tool.

Normal Python shadowing is opt-in and visible: *you* put a directory on `sys.path` and import by that name, and it only affects *your* imports. Here, dropping a sensibly-named tool (`tools/json.py`) makes the registry mutate `sys.modules["json"]` for the whole process — the user never asked to shadow anything. And this isn't a new opinion: `loader.py` **already** decided this is wrong (it loads under a `_strands_tool_` prefix). This PR just extends that same convention to the two `registry.py` paths that were missed (`reload_tool`, `initialize_tools`).

**DevX before → after** (I ran a tool literally named `json` against both revisions):

| path | before | after |
|---|---|---|
| `initialize_tools` (dir discovery) | tool **silently never registers** — `import_module("json")` hits the module cache and returns stdlib json, which has no `@tool` fns | tool registers ✅ |
| `reload_tool` (hot reload) | `sys.modules["json"]` **clobbered** → any later `import json` in the process gets the tool and breaks with `AttributeError` | stdlib untouched ✅ |

The nasty part is the second one: the failure lands far from the cause (some unrelated library doing `import json`), so it's near-undiagnosable for a user whose only "mistake" was naming a tool after a stdlib module (`json`, `types`, `io`, `logging`, …).

<details><summary>Runnable repro + actual output</summary>

`tools/json.py`:
```python
from strands import tool

@tool
def json(data: str) -> str:
    """Pretty-print a value as JSON-ish text."""
    return f"json-tool saw: {data}"
```

`run.py`:
```python
import sys, json as stdlib
from strands.tools.registry import ToolRegistry

r = ToolRegistry()
r.initialize_tools(load_tools_from_directory=True)
print("A) registered:", list(r.registry.keys()))

try:
    r.reload_tool("json")
except Exception as e:
    print("B) reload raised:", type(e).__name__, e)
import json as after
try:
    print("B) import json; dumps ->", after.dumps({"still": "works"}))
except Exception as e:
    print("B) import json; dumps BROKE ->", type(e).__name__, e)
```

```
######## BEFORE (merge-base) ########
A) registered: []
B) import json; dumps BROKE -> AttributeError module 'json' has no attribute 'dumps'

######## AFTER (PR #2994) ########
A) registered: ['json']
B) import json; dumps -> {"still": "works"}
```
</details>

**Known boundary** (already flagged in the bot review, worth a doc line, not a blocker): a directory tool's *sibling* imports still land in `sys.modules` under their bare name, so the collision class isn't 100% closed — but the tool module itself no longer clobbers anything.

Low-risk, high-blast-radius bug; extending an existing convention rather than introducing a new one. I'd take it.

<sub>🤖 Automated analysis for review — verified by running the repro above against both revisions; a human should still confirm.</sub>

liramon2 (approved):

Unshure (commented):
This generally looks fine to me, but had two notes:
- Was there an issue that was open reporting this? If so, can you link to it in the description.
- Is this technically breaking? If im understanding the change correctly, I think it is (changing the module import name), but is very unlikely to actually impact customers. I think we should minor version bump for this instead of a patch update.

Unshure (approved):

mehtarac (approved):

PR #3026, opened 2026-06-30
fix(session): repair mid-iteration skip of orphaned toolUse

### LINKED ISSUE #2028: Test: _fix_broken_tool_use tests don't guard against in-place mutation; missing coverage for insert-induced edge case
## Summary

While reviewing #2026 I found two test quality issues in `tests/strands/session/test_repository_session_manager.py` for `_fix_broken_tool_use`.

### 1. `test_fix_broken_tool_use_ignores_last_message` doesn't actually guard against mutation

The test asserts:

```python
fixed_messages = session_manager._fix_broken_tool_use(messages)
assert fixed_messages == messages
```

Since `_fix_broken_tool_use` mutates the list **in-place** and returns the **same object**, `fixed_messages is messages` is always `True`. This means `fixed_messages == messages` passes regardless of whether the method appended anything. The test would still pass even after applying the fix from #2026.

**Fix:** Use `copy.deepcopy(messages)` to snapshot the original before calling the method, and compare against the snapshot.

### 2. No test for `insert()`-induced last-message skip

When consecutive orphaned `toolUse` messages exist without a trailing non-toolUse message, `insert()` at [L306](https://github.com/strands-agents/sdk-python/blob/main/src/strands/session/repository_session_manager.py#L306) shifts subsequent messages, causing the final `toolUse` to become the new "last message" mid-iteration. The `index + 1 < len(messages)` guard then skips it.

Example:
```python
messages = [
    {"role": "assistant", "content": [{"toolUse": {"toolUseId": "A", ...}}]},
    {"role": "assistant", "content": [{"toolUse": {"toolUseId": "B", ...}}]},
]
```

1. `index=0`: toolUse(A) → `insert(1, toolResult(A))` → list becomes `[A, resultA, B]` (len=3)
2. `index=2`: toolUse(B) → `index+1=3 < 3` is `False` → **skipped**

The `else` branch introduced in #2026 implicitly fixes this, but there's no test to pin the behavior.

### Proposed changes

- Apply `copy.deepcopy` to the existing last-message test
- Add a test for the consecutive orphaned toolUse scenario (insert-induced skip)
- Add a test for the single-message edge case (conversation is just one orphaned toolUse)

Related: #2025, #2026

chaynabors:
_fix_broken_tool_use walked messages with enumerate() and compared index + 1 against a live len(messages). When two orphaned toolUse messages sat next to each other, inserting a synthetic toolResult after the first one shifted the second orphan into the new last-message position mid-loop, and the guard skipped it.

The fix snapshots the eligible indices up front, excluding the originally-trailing message which agent.py handles at prompt-arrival time so we do not mutate persisted history that may never be resumed. Walking the snapshot in reverse keeps each insert from shifting an unprocessed position.

Closes #2028.

### DIFF
--- a/strands-py/tests/strands/session/test_repository_session_manager.py
+++ b/strands-py/tests/strands/session/test_repository_session_manager.py
@@ -1,5 +1,6 @@
 """Tests for AgentSessionManager."""
 
+import copy
 from unittest.mock import Mock
 
 import pytest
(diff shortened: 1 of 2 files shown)


strands-py/tests/strands/session/test_repository_session_manager.py
```diff
@@ -427,11 +428,79 @@ def test_fix_broken_tool_use_ignores_last_message(session_manager):
...
+    original = copy.deepcopy(messages)
+
+    fixed_messages = session_manager._fix_broken_tool_use(messages)
+
+    assert fixed_messages == original
+
+
+@pytest.mark.xfail(
+    strict=True,
+    reason="Bug #2028: enumerate+insert iteration skips the trailing orphaned toolUse so it never gets a toolResult.",
```
zastrowm: Can we just fix the bug right away?

zastrowm (dismissed):

strands-py/tests/strands/session/test_repository_session_manager.py
chaynabors: Edit: This one is tricky. Give me a bit. There's two ways to do this and a bunch of regression tests break either way.

zastrowm (approved):

PR #3028, opened 2026-06-30
fix(multiagent): preserve failed status reported by graph nodes

### LINKED ISSUE #683: The status of the AgentResult or MultiAgentResult gets overwritten to status.COMPLETED
https://github.com/strands-agents/sdk-python/blob/8c63d75ecf9c246110d297c109bf204839978152/src/strands/multiagent/graph.py#L555

Shouldn't the custom node be able to set the status to `status.FAILED`? How does the node communicate that something went wrong, so conditional edges are interpreted correctly?

chaynabors:
A graph node executor that is itself a MultiAgentBase can return a NodeResult with status FAILED, but the graph overwrote it to COMPLETED and routed the node to completed_nodes. That left execution_status disagreeing with result.status and prevented conditional edges from steering around the failure.

A FAILED node_result now routes to failed_nodes with execution_status FAILED, matching how nodes that raise are already recorded.

Closes #683.

### DIFF
--- a/strands-py/src/strands/multiagent/graph.py
+++ b/strands-py/src/strands/multiagent/graph.py
@@ -1079,9 +1079,12 @@ class Graph(MultiAgentBase):
                 yield self._activate_interrupt(node, node_result.interrupts)
                 return
 
-            # Mark as completed
-            node.execution_status = Status.COMPLETED
-            self.state.completed_nodes.add(node)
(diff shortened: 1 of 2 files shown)


gautamsirdeshmukh:
Reviewing

gautamsirdeshmukh (approved):

PR #3029, opened 2026-06-30
fix(model): drop blank text content blocks before send

chaynabors:
Bedrock models such as openai.gpt-oss-20b-1:0 emit an empty text content block alongside reasoning and tool use. The aggregator stored that as a TextBlock(''), so on the next turn Bedrock rejected the request with ValidationException for blank text at messages.1.content.0. Empty TextBlock instances are now filtered at message-construction time, while the streamed events themselves are unchanged for consumers reading the raw stream.

Originally proposed by @manikanta5827 on the archived sdk-typescript repo as #380; this carries the fix forward to the monorepo path with a regression test.

Fixes #2522.

### DIFF
--- a/strands-ts/src/models/model.ts
+++ b/strands-ts/src/models/model.ts
@@ -436,9 +436,12 @@ export abstract class Model<T extends BaseModelConfig = BaseModelConfig> {
           case 'modelMessageStopEvent':
             // Store message and stop reason
             if (messageRole) {
+              const filtered = contentBlocks.filter(
+                (block) => !(block instanceof TextBlock && block.text.trim() === '')
+              )
               stoppedMessage = new Message({
                 role: messageRole,
-                content: [...contentBlocks],
+                content: filtered,
               })
               finalStopReason = event.stopReason!
             }
--- a/strands-ts/src/models/__tests__/model.test.ts
+++ b/strands-ts/src/models/__tests__/model.test.ts
@@ -633,6 +633,42 @@ describe('Model', () => {
       })
     })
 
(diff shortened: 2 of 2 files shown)


strands-ts/src/models/model.ts
```diff
@@ -333,7 +333,7 @@ export abstract class Model<T extends BaseModelConfig = BaseModelConfig> {
     try {
       // State maintained in closure
       let messageRole: Role | null = null
-      const contentBlocks: ContentBlock[] = []
+      let contentBlocks: ContentBlock[] = []
```
mehtarac: NIT suggestion: Keep const and introduce a local filtered variable

```
const filtered = contentBlocks.filter((block) => !(block instanceof TextBlock && block.text === ''))
  stoppedMessage = new Message({
    role: messageRole,
    content: filtered,
  })
```

Reasoning being that it avoids accidental reassignment in the future

strands-ts/src/models/model.ts
```diff
@@ -436,6 +436,7 @@ export abstract class Model<T extends BaseModelConfig = BaseModelConfig> {
           case 'modelMessageStopEvent':
             // Store message and stop reason
             if (messageRole) {
+              contentBlocks = contentBlocks.filter((block) => !(block instanceof TextBlock && block.text === ''))
```
mehtarac: This fixes the ValidationException on `''` (empty text). Do we know if ValidationException triggers on whitespace-only text as well?

liramon2 (dismissed):

mehtarac (approved):

PR #2317, opened 2026-05-22
chore: prepare directory layout for monorepo convergence

zastrowm:
## Description

Prepare the sdk-python directory layout for monorepo convergence. This moves all source into `strands-py/` and updates CI workflows so they continue to function in the new structure.

Changes:
- Move all source, tests, docs, and config into `strands-py/` subdirectory
- Rename Python-specific workflows with `python-` prefix and add path filters (`strands-py/**`, `.github/workflows/python-*`)
- Add `working-directory: strands-py` to build/test/lint workflow steps
- Update griffe API check to use `strands-py/src` search path
- Update `readme` path in pyproject.toml to `../README.md` (points to repo root from subdirectory)
- Configure hatch VCS versioning for monorepo (`root=".."`, `tag_regex=python/v*`, custom `git_describe_command`)
- Update dependabot pip directory to `/strands-py`
- Keep generic repo-level workflows (auto-close, issue-responder, strands-command, pr-title, pr-size-labeler, auto-strands-review) unprefixed and unscoped
- Keep original README at repo root

Some CI workflows are expected to fail until the merge occurs

## Related Issues

- #2286

## Type of Change

Other (please describe): Directory restructuring in preparation for mono-repo convergence.

## Testing

Validated on fork ([zastrowm/sdk-python](https://github.com/zastrowm/sdk-python)):

- ✅ **workflow_dispatch** — full test suite + lint + API check pass: [run 26293255129](https://github.com/zastrowm/sdk-python/actions/runs/26293255129)
- ✅ **PR path triggers** — all path-filtered workflows fire correctly on PRs touching `strands-py/`: [run 26299112106](https://github.com/zastrowm/sdk-python/actions/runs/26299112106)
- ✅ **check-api (griffe)** — API compatibility check works with `--search strands-py/src`
- ✅ **pr-title** — conventional commits validation fires on PR
- ✅ **pr-size-labeler** — size labeling fires on PR
- ✅ **auto-strands-review** — review bot triggers on PR
- ⚠️ **integration-test** — authorization check passes, but `check-access-and-checkout` fails as expected (fork lacks AWS secrets)
- ✅ **hatch version** — correctly resolves version from `python/v*` tags with new VCS config
- ✅ **Git rename detection** — all file moves show as renames (history preserved)

## Checklist
- [x] I have read the CONTRIBUTING document
- [X] I have added any necessary tests that prove my fix is effective or my feature works
- [x] I have updated the documentation accordingly
[...]

PR #2000, opened 2026-03-29
fix(bidi): Gemini 3.1 Flash Live compatibility for BidiGeminiLiveModel

jrdeck:
## Summary

Fixes three incompatibilities that prevent `BidiGeminiLiveModel` from working with `gemini-3.1-flash-live-preview`.

Closes #1999

### 1. Route text input through `send_realtime_input`

Gemini 3.1 Flash Live [only allows `send_client_content` for seeding initial context history](https://ai.google.dev/gemini-api/docs/live-api/capabilities). Mid-session text (e.g. `BidiTextInputEvent` from a `BidiAgent` auto-greeting) must use `send_realtime_input(text=...)` instead. This path is also compatible with Gemini 2.5 models — no regression.

**Before:** `_send_text_content` → `send_client_content` → error 1007
**After:** `_send_text_content` → `send_realtime_input(text=...)` → audio response

### 2. Omit `session_resumption` when no handle exists

`_build_live_config` unconditionally injected `session_resumption: {"handle": None}`. Gemini 3.1 rejects this; 2.5 silently accepted it. Now only included when a handle is actually provided.

### 3. Default `api_version` to `v1beta`

`_resolve_client_config` defaulted to `v1alpha`, which is not a documented Gemini Developer API version. Changed to `v1beta` — the correct endpoint for the Live API per [Google's documentation](https://ai.google.dev/api/live).

## Test plan

- [x] All 18 existing `test_gemini_live.py` tests pass
- [x] Test assertion updated: text input now asserts `send_realtime_input(text=...)` instead of `send_client_content`
- [x] Confirmed end-to-end: `gemini-3.1-flash-live-preview` returns audio chunks in response to text sent via `send_realtime_input`
- [x] Confirmed backward-compatible: `gemini-2.5-flash-native-audio-latest` works with the same code path

## Context

Discovered by [Clevvi](https://clevvi.com.au) while building a multi-agent voice assistant on Bedrock AgentCore + Strands. The workaround (subclassing `BidiGeminiLiveModel`) is deployed in production — this PR upstreams the fix.

PR #3034, opened 2026-06-30
fix(bidi): gemini 3.1 Flash Live compatibility for BidiGeminiLiveModel

### LINKED ISSUE #1999: BidiGeminiLiveModel incompatible with Gemini 3.1 Flash Live
## Bug description

`BidiGeminiLiveModel` fails with `APIError: 1007 Request contains an invalid argument` when used with `gemini-3.1-flash-live-preview`. The connection establishes successfully but the first text input causes the error.

### 1. `_send_text_content` uses `send_client_content` — rejected by Gemini 3.1

`_send_text_content` unconditionally routes text through `send_client_content`:

```python
async def _send_text_content(self, text: str) -> None:
    content = genai_types.Content(role="user", parts=[genai_types.Part(text=text)])
    await self._live_session.send_client_content(turns=content)
```

Gemini 3.1 Flash Live [only allows `send_client_content` for seeding initial context history](https://ai.google.dev/gemini-api/docs/live-api/capabilities). Mid-session text must use `send_realtime_input(text=...)` instead.

This worked with Gemini 2.5 models which accepted `send_client_content` throughout the session.

**Fix:** Use `send_realtime_input(text=text)` for Gemini 3.1+ models.

### 2. `_build_live_config` injects `session_resumption: {"handle": None}`

```python
config_dict["session_resumption"] = {"handle": kwargs.get("live_session_handle")}
```

When no session handle is provided, this sends `{"handle": None}` to the API. Gemini 3.1 rejects this; 2.5 silently accepted it. The field should be omitted entirely when no handle exists.

### 3. (Minor) `_resolve_client_config` defaults to `api_version: "v1alpha"`

```python
if "http_options" not in resolved:
    resolved["http_options"] = {"api_version": "v1alpha"}
```

`v1alpha` is not a documented Gemini Developer API version. The correct default is `v1beta`.

## Reproduction

```python
from strands.experimental.bidi.models.gemini_live import BidiGeminiLiveModel
from strands.experimental.bidi import BidiAgent

model = BidiGeminiLiveModel(
    model_id="gemini-3.1-flash-live-preview",
    client_config={"api_key": "<key>", "http_options": {"api_version": "v1beta"}},
)
agent = BidiAgent(model=model, tools=[], system_prompt="Say hello.")
await agent.start()
# Auto-greeting or any BidiTextInputEvent triggers the error
```

**Error:**
```
google.genai.errors.APIError: 1007 None. Request contains an invalid argument.
```

**Traceback path:** `BidiAgent.send(BidiTextInputEvent)` → `_send_text_content` → `send_client_content` → 1007

## Confirmed fix

```python
# send_realtime_input works for text on 3.1
await session.send_realtime_input(text="Hello")  # ✅ returns audio
await session.send_client_content(turns=content)  # ❌ error 1007
```

## Environment

- `strands-agents` v1.33.0
- `google-genai` v1.69.0
- Python 3.14.3
- Model: `gemini-3.1-flash-live-preview`

## Workaround

Subclass `BidiGeminiLiveModel` and override `_send_text_content` and `_build_live_config`:

```python
class PatchedGeminiLiveModel(BidiGeminiLiveModel):
    def _build_live_config(self, system_prompt=None, tools=None, **kwargs):
        config = super()._build_live_config(system_prompt, tools, **kwargs)
        if config.get("session_resumption") == {"handle": None}:
            del config["session_resumption"]
        return config

    async def _send_text_content(self, text: str) -> None:
        await self._live_session.send_realtime_input(text=text)
```

lizradway:
## Summary

Rebased version of #2000 (by @jrdeck) — the original had merge conflicts from the monorepo restructure (#2317) but the code changes are identical. James Deck is preserved as commit author.

Fixes three incompatibilities that prevent `BidiGeminiLiveModel` from working with `gemini-3.1-flash-live-preview`:

1. **Route text input through `send_realtime_input`** — Gemini 3.1 only allows `send_client_content` for seeding initial context history; mid-session text must use `send_realtime_input`. Compatible with Gemini 2.5.
2. **Omit `session_resumption` when no handle exists** — sending `{"handle": None}` is rejected by Gemini 3.1.
3. **Default `api_version` to `v1beta`** — `v1alpha` is not a documented endpoint.

Closes #1999
Supersedes #2000

## Test plan

- [x] All 18 existing `test_gemini_live.py` tests pass
- [x] Test assertion updated to match new `send_realtime_input` path

### DIFF
--- a/strands-py/src/strands/experimental/bidi/models/gemini_live.py
+++ b/strands-py/src/strands/experimental/bidi/models/gemini_live.py
@@ -102,7 +102,7 @@ class BidiGeminiLiveModel(BidiModel):
 
         # Set default http_options if not provided
         if "http_options" not in resolved:
-            resolved["http_options"] = {"api_version": "v1alpha"}
+            resolved["http_options"] = {"api_version": "v1beta"}
 
         return resolved
 
@@ -429,12 +429,14 @@ class BidiGeminiLiveModel(BidiModel):
         await self._live_session.send(input=msg)
 
     async def _send_text_content(self, text: str) -> None:
-        """Internal: Send text content using Gemini Live API."""
-        # Create content with text
-        content = genai_types.Content(role="user", parts=[genai_types.Part(text=text)])
+        """Internal: Send text content using Gemini Live API.
 
-        # Send as client content
-        await self._live_session.send_client_content(turns=content)
+        Uses send_realtime_input for text delivery. Gemini 3.1+ models only
+        accept send_client_content for seeding initial context history — mid-session
+        text must go through send_realtime_input. This path is also compatible with
+        Gemini 2.5 models.
+        """
+        await self._live_session.send_realtime_input(text=text)
 
     async def _send_tool_result(self, tool_result: ToolResult) -> None:
         """Internal: Send tool result using Gemini Live API."""
@@ -491,7 +493,9 @@ class BidiGeminiLiveModel(BidiModel):
         """
         config_dict: dict[str, Any] = self.config["inference"].copy()
 
-        config_dict["session_resumption"] = {"handle": kwargs.get("live_session_handle")}
+        live_session_handle = kwargs.get("live_session_handle")
+        if live_session_handle is not None:
+            config_dict["session_resumption"] = {"handle": live_session_handle}
 
         # Add system instruction if provided
         if system_prompt:
--- a/strands-py/tests/strands/experimental/bidi/models/test_gemini_live.py
+++ b/strands-py/tests/strands/experimental/bidi/models/test_gemini_live.py
@@ -198,14 +198,13 @@ async def test_send_all_content_types(mock_genai_client, model):
     _, mock_live_session, _ = mock_genai_client
     await model.start()
 
-    # Test text input
+    # Test text input — routed through send_realtime_input (Gemini 3.1 compatible)
     text_input = BidiTextInputEvent(text="Hello", role="user")
     await model.send(text_input)
-    mock_live_session.send_client_content.assert_called_once()
-    call_args = mock_live_session.send_client_content.call_args
-    content = call_args.kwargs.get("turns")
-    assert content.role == "user"
-    assert content.parts[0].text == "Hello"
+    mock_live_session.send_realtime_input.assert_called_with(text="Hello")
+
+    # Reset for next assertion
+    mock_live_session.send_realtime_input.reset_mock()
 
     # Test audio input (base64 encoded)
     audio_b64 = base64.b64encode(b"audio_bytes").decode("utf-8")


strands-py/src/strands/experimental/bidi/models/gemini_live.py
```diff
@@ -429,12 +429,14 @@ async def _send_image_content(self, image_input: BidiImageInputEvent) -> None:
...
+        """Internal: Send text content using Gemini Live API.
 
-        # Send as client content
-        await self._live_session.send_client_content(turns=content)
+        Uses send_realtime_input for text delivery. Gemini 3.1+ models only
+        accept send_client_content for seeding initial context history — mid-session
+        text must go through send_realtime_input. This path is also compatible with
+        Gemini 2.5 models.
+        """
+        await self._live_session.send_realtime_input(text=text)
```
mehtarac, line 439: I think we need to have a way to use both APIs. 
`send_client_content` can be used for context pre-fill and send the provide the previous conversation to the model before using the `send_realtime_input` API for realtime inputs. 

Will have to dig into https://googleapis.github.io/python-genai/genai.html#genai.live.AsyncSession.send_client_content

strands-py/src/strands/experimental/bidi/models/gemini_live.py
```diff
@@ -491,7 +493,9 @@ def _build_live_config(
         """
         config_dict: dict[str, Any] = self.config["inference"].copy()
 
-        config_dict["session_resumption"] = {"handle": kwargs.get("live_session_handle")}
+        live_session_handle = kwargs.get("live_session_handle")
+        if live_session_handle is not None:
```
yonib05, line 497: this breaks resumption for every fresh session. the server only issues handles if `session_resumption` is in the connect config, so now we never get one and go_away restarts cant resume.

worse - the restart passes `live_session_handle=None`, and `start()` skips history replay whenever that key is present (line 165). so after the first timeout the session has no handle and no history, full context loss

fix is send `{}` instead of ommiting the field:

```python
config_dict["session_resumption"] = {"handle": h} if h else {}
```

also neither branch here is tested

strands-py/src/strands/experimental/bidi/models/gemini_live.py
```diff
@@ -102,7 +102,7 @@ def _resolve_client_config(self, config: dict[str, Any]) -> dict[str, Any]:
 
         # Set default http_options if not provided
         if "http_options" not in resolved:
-            resolved["http_options"] = {"api_version": "v1alpha"}
+            resolved["http_options"] = {"api_version": "v1beta"}
```
yonib05, line 105: this changes the default for everyone not just 3.1. anyone using v1alpha-only options (`enable_affective_dialog`, `proactivity`) breaks at connect. needs a changelog note atleast

yonib05 (commented):
text routing and api_version changes look fine, the session_resumption change breaks reconnect though, see inline

lizradway:
PR to be taken over by @mehtarac in bidi roadmap

PR #3036, opened 2026-06-30
fix(skills): restrict remote skill fetches to public hosts

yonib05:
## Description

`Skill.from_url` previously fetched a remote `SKILL.md` after only checking that the URL used the `https://` scheme. This change validates the URL's host before fetching and allows only public, routable addresses.

The new `_validate_fetch_host` helper:

- Parses the host from the URL.
- Checks IP literals directly, and resolves hostnames to all of their addresses with `socket.getaddrinfo`.
- Allows only public addresses; loopback, link-local, private, multicast, reserved, and unspecified ranges are rejected, along with a small set of additional non-public addresses not covered by the standard `ipaddress` properties.
- Raises `ValueError` when the host is missing, cannot be resolved, or falls in a disallowed range.

Resolving the host up front keeps the set of checked addresses aligned with the set the fetch will actually connect to.

## Behavior change

- Fetching a remote skill from a host that resolves to a non-public address now raises `ValueError` instead of attempting the request.
- Skills hosted on normal public endpoints are unaffected.

## Testing

- Added regression tests: non-public IP literals are rejected without any fetch; a hostname resolving to a non-public address is rejected; an unresolvable host is rejected; a host resolving to a public address still fetches successfully.
- `pytest tests/strands/vended_plugins/skills/test_skill.py` -> 82 passed.
- `ruff check` and `ruff format --check` clean; `mypy` clean on the changed module.

### DIFF
--- a/strands-py/src/strands/vended_plugins/skills/skill.py
+++ b/strands-py/src/strands/vended_plugins/skills/skill.py
@@ -8,12 +8,13 @@ frontmatter metadata and markdown instructions.
 
 from __future__ import annotations
 
+import http.client
 import ipaddress
 import logging
 import re
 import socket
-import urllib.error
-import urllib.request
+import ssl
+import urllib.parse
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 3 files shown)


strands-py/src/strands/vended_plugins/skills/skill.py
```diff
@@ -23,6 +27,267 @@
...
+# data, ...) is rejected so the fetch cannot be redirected onto a non-network or
+# lower-trust transport.
+_ALLOWED_SCHEMES = frozenset({"http", "https"})
+
+# Maximum number of bytes read from a remote SKILL.md response.
+_MAX_FETCH_BYTES = 5 * 1024 * 1024
+
+# A small set of non-public IPv6 addresses that ``ipaddress`` does not flag via
+# its is_* properties, so they are matched explicitly.
+_EXTRA_NON_PUBLIC_IPS = frozenset({ipaddress.ip_address("fd00:ec2::254")})
```
zastrowm, line 40: > Allows only public addresses; loopback, link-local, private, multicast, reserved, and unspecified ranges are rejected, along with a small set of additional non-public addresses not covered by the standard ipaddress properties.

Is from_url exposed to the agent?  if not, I don't see a need to do this - this is user-controlled as it's code, so you should be able to any ip ranges you want.   

Even file/ftp/etc. I don't see any reason to disallow these

strands-py/src/strands/vended_plugins/skills/skill.py
zastrowm, line 1: Note that from_url is undocumented in the docs and I also wonder if we have a TS equivalent

PR #3055, opened 2026-07-01
fix(ts): exclude test artifacts from package

### LINKED ISSUE #3004: [BUG] __tests__ and __fixtures__ are included in published NPM package
### Checks

### SDK Language

TypeScript

### Strands Version

1.7.0

### Language Runtime Version

Node 24

### Operating System

MacOS

### Installation Method

npm

### Steps to Reproduce

Check https://npmx.dev/package-code/@strands-agents/sdk/v/1.7.0/dist%2Fsrc

### Expected Behavior

No `__tests__` or `__fixtures__` folders are published to the NPM registry

### Actual Behavior

The published NPM package contains nested `__tests__` and `__fixtures__` folders unnecessarily bloating the bundle size

### Additional Context

_No response_

### Possible Solution

_No response_

### Related Issues

_No response_

maxmilian:
## Description

This PR keeps generated test artifacts out of the published TypeScript npm package.

The TypeScript build still emits the existing `dist` tree, but the package `files` allowlist now excludes generated test artifacts declaratively:

```json
"files": [
  "dist",
  "!dist/**/__tests__",
  "!dist/**/__fixtures__",
  "!dist/**/*.test.*",
  "README.md",
  "LICENSE"
]
```

That means the published tarball omits compiled `__tests__`, `__fixtures__`, and `.test.*` outputs without mutating `dist`, without a prune script, and without a `postpack` rebuild. `prepack` remains the normal package build hook, while `test:package` runs `npm pack --dry-run --ignore-scripts --json` after its explicit build so the package contents assertion checks the real `files` globs without triggering a redundant lifecycle build.

I also wired a package contents assertion into `test:package` so future changes fail if test artifacts re-enter the dry-run tarball. The guard fails closed on empty stdin, unparseable JSON, npm error payloads, and empty file lists, and it checks the same artifact classes as the `files` negations.

## Related Issues

Resolves #3004

## Documentation PR

No documentation changes needed for user-facing SDK behavior. The package test README was updated to document the local dry-run tarball contents guard.

## Type of Change

Bug fix

## Testing

- [x] `npm run test:package -w strands-ts`
- [x] `npm run lint -w strands-ts`

## Checklist
- [x] I have added an appropriate example to the documentation, or no new docs are needed

----


strands-ts/test/packages/assert-package-contents.js
```diff
@@ -0,0 +1,24 @@
+const chunks = []
+
+process.stdin.setEncoding('utf8')
+
+for await (const chunk of process.stdin) {
+  chunks.push(chunk)
+}
+
+const packOutput = JSON.parse(chunks.join(''))
+const files = packOutput[0]?.files?.map((file) => file.path) ?? []
```
yonib05: This fails open. If prepack fails, `npm pack --dry-run --json` prints `{"error":{...}}` on stdout and exits 1, but the pipe's exit code comes from this script, and `?.files ?? []` turns the error object into an empty list. Verified locally: a failing prepack prints `[package-contents] OK (0 files)` and exits 0, so a broken build passes the check.

Fail if the parsed output has an `error` key, and assert `files.length > 0`.

strands-ts/package.json
```diff
@@ -136,7 +136,8 @@
   },
   "scripts": {
     "build": "tsc --project src/tsconfig.json",
-    "prepack": "npm run build",
+    "prepack": "npm run build && node scripts/prune-package-artifacts.js",
+    "postpack": "rm -rf dist && npm run build",
```
yonib05: `rm -rf` doesn't exist on Windows (npm runs lifecycle scripts under cmd.exe), and CI runs `test:package` on windows-latest. postpack fails on every pack there, and the fail-open pipe in test:package hides it, so the package check never actually runs on the Windows matrix jobs.

strands-ts/package.json
```diff
@@ -136,7 +136,8 @@
   },
   "scripts": {
     "build": "tsc --project src/tsconfig.json",
-    "prepack": "npm run build",
+    "prepack": "npm run build && node scripts/prune-package-artifacts.js",
```
yonib05: If pack fails or is interrupted after prepack, postpack never runs and dist stays pruned. Since the build is composite/incremental and the tsbuildinfo still lists the deleted files as emitted, the next `npm run build` is a no-op and doesn't restore them. Only a manual clean recovers.

strands-ts/package.json
```diff
@@ -152,7 +153,7 @@
     "test:browser:install": "npx playwright install --with-deps chromium",
     "test:all": "vitest run --project unit-node --project unit-browser",
     "test:all:coverage": "vitest run --coverage --project unit-node --project unit-browser",
-    "test:package": "cd test/packages/esm-module && npm install && node esm.js && cd ../cjs-module && npm install && node cjs.js",
+    "test:package": "npm run build && cd test/packages/esm-module && npm install && node esm.js && cd ../cjs-module && npm install && node cjs.js && cd ../../.. && npm pack --dry-run --json | node test/packages/assert-package-contents.js",
```
yonib05: This now runs three full tsc builds per invocation: the leading `npm run build`, prepack's build, and postpack's rebuild (which is a cold build because `rm -rf dist` deletes the tsbuildinfo). CI already builds right before test:package, so every matrix job pays for three redundant compiles.

strands-ts/scripts/prune-package-artifacts.js
```diff
@@ -0,0 +1,42 @@
+import { readdir, rm } from 'node:fs/promises'
+import { fileURLToPath } from 'node:url'
+import { join } from 'node:path'
+
+const distDir = new URL('../dist/', import.meta.url)
+const testFilePattern = /\.test(?:\.[^/.]+)*\.(?:js|d\.ts)(?:\.map)?$/
```
yonib05: This regex and the `__tests__`/`__fixtures__` names are duplicated in assert-package-contents.js. If one side is updated and the other isn't, either test files ship with a green check or packing breaks. Put the patterns in one shared place.

strands-ts/test/packages/assert-package-contents.js
```diff
@@ -0,0 +1,24 @@
+const chunks = []
+
+process.stdin.setEncoding('utf8')
+
+for await (const chunk of process.stdin) {
+  chunks.push(chunk)
+}
+
+const packOutput = JSON.parse(chunks.join(''))
```
yonib05: Lifecycle script stdout gets interleaved before the JSON on npm's stdout (only the `> pkg script` banners go to stderr). This works today only because tsc and the prune script print nothing on success; any future console.log in prepack breaks JSON.parse here with a confusing error.

yonib05 (commented):
The fix works on the happy path, but the prune/rebuild mechanism is more fragile than the problem needs. npm's `files` field supports negation patterns, so this can be:

```json
"files": ["dist", "!dist/**/__tests__", "!dist/**/__fixtures__", "!dist/**/*.test.*", "README.md", "LICENSE"]
```

Same tarball, no prepack prune, no postpack rebuild, no dist mutation. Keep the tarball assertion in test:package as the regression check. Alternatively, stop compiling tests into dist with a build tsconfig that excludes them; nothing consumes dist/__tests__ (vitest runs from src).

Either of those makes most of the inline comments below moot.

One more small thing: lint and format only cover `src test/integ`, so the two new scripts are not checked by either.

yonib05 (changes requested):

maxmilian:
Thanks for the review @yonib05 — adopted the `files`-negation approach in 6e6de70. Much simpler, and it moots most of the inline comments:

- **`files` negation** — `dist` now carries `!dist/**/__tests__`, `!dist/**/__fixtures__`, `!dist/**/*.test.*`. Verified against a built `dist` (which does contain those artifacts): `npm pack --dry-run --json` yields **0** forbidden paths and 741 legit files.
- **Dropped `postpack` entirely** and reduced `prepack` to just `npm run build`. That removes the Windows `rm -rf` failure, the interrupted-pack `dist` corruption, and the cold triple-build — all gone with the machinery.
- **Deleted `prune-package-artifacts.js`**, so the duplicated forbidden-pattern regex now lives in exactly one place (the assert guard).
- **Fixed the fail-open guard** — `assert-package-contents.js` now throws on an empty stdin, an unparseable payload, a `{"error":...}` result, or an empty file list, instead of printing `OK (0 files)`. It also links #3004 per AGENTS.md.

`npm run test:package` passes end-to-end (esm + cjs module checks, pack, assert).

On the last point — lint/format only covering `src test/integ`: with the prune script deleted, the only remaining new file is the assert guard under `test/packages/`. I left the lint globs alone rather than pull the `esm-module`/`cjs-module` fixtures into scope, but ran `prettier --check` on the guard manually (clean). Happy to widen the globs to `test/packages/*.js` if you'd prefer it enforced.

strands-ts/test/packages/assert-package-contents.js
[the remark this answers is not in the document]
yonib05: Fixed in 6e6de70. The guard now throws on empty stdin, unparseable JSON, an error payload, and an empty file list.

yonib05: Verified fixed in 6e6de70. A failing prepack now hits the error branch (npm puts the error JSON on stdout, confirmed on npm 11), and the empty/malformed guards cover the rest.

strands-ts/package.json
yonib05: Resolved by dropping postpack entirely in 6e6de70.

strands-ts/package.json
yonib05: Moot now that nothing mutates dist. Resolved in 6e6de70.

strands-ts/scripts/prune-package-artifacts.js
yonib05: The prune script is gone, so this duplication is resolved. The pattern list still lives in two places (the files globs and the guard regexes); I left a separate comment on keeping those in sync.

strands-ts/package.json
yonib05: Down from three builds to two now that postpack is gone, but test:package still compiles twice: the leading `npm run build`, then prepack firing again inside `npm pack`. Adding `--ignore-scripts` to the pack call gets it to one; npm evaluates the `files` globs either way, so the assertion still tests the real thing. It also fixes the stdout-interleaving issue in the other thread, since prepack no longer runs inside the pipe.

Not a nit. This runs on all 9 ts-test matrix jobs, and CI already builds right before test:package.

strands-ts/test/packages/assert-package-contents.js
yonib05: Still applies to the new pipeline: prepack runs inside `npm pack` and anything it prints to stdout lands in front of the JSON. The guard now fails closed on that (JSON.parse throws), which is the right direction, but the error message would point at the wrong thing. `--ignore-scripts` on the pack call removes the whole class; see the build thread.

strands-ts/test/packages/assert-package-contents.js
```diff
@@ -0,0 +1,49 @@
...
+
+const files = (Array.isArray(packOutput) ? packOutput[0]?.files : undefined)?.map((file) => file.path)
+if (!files || files.length === 0) {
+  throw new Error('[package-contents] pack produced no file list — refusing to pass on an empty payload')
+}
+
+const forbiddenPatterns = [
+  /(^|\/)__tests__(\/|$)/,
+  /(^|\/)__fixtures__(\/|$)/,
+  /\.test(?:\.[^/.]+)*\.(?:js|d\.ts)(?:\.map)?$/,
```
yonib05: Nit: this regex is narrower than the `!dist/**/*.test.*` glob it guards. The glob excludes any extension, but the regex only matches `.js`, `.d.ts`, and `.map` endings, so `a.test.mjs`, `a.test.cjs`, or `a.test.d.mts` would be excluded by npm yet invisible to the guard if the negations ever regress for those shapes. A plain `/\.test\./` mirrors the glob exactly and there's no rationale to re-derive later.

Not blocking, since tsc only emits the three shapes you cover today.

strands-ts/test/packages/assert-package-contents.js
```diff
@@ -0,0 +1,49 @@
...
+if (packOutput?.error) {
+  throw new Error(`[package-contents] npm pack reported an error: ${JSON.stringify(packOutput.error)}`)
+}
+
+const files = (Array.isArray(packOutput) ? packOutput[0]?.files : undefined)?.map((file) => file.path)
+if (!files || files.length === 0) {
+  throw new Error('[package-contents] pack produced no file list — refusing to pass on an empty payload')
+}
+
+const forbiddenPatterns = [
```
yonib05, line 39: Nit: these patterns and the `files` negations in package.json are two encodings of the same list and can drift independently. Worth a one-line comment here pointing at the `files` entries (and vice versa) so whoever adds `!dist/**/__mocks__` someday updates both. A shared source of truth would be nicer but is overkill for three patterns.

yonib05 (commented):
Thanks for turning this around so quickly. The files-negation rewrite fixed everything that blocked the last round: no more prune/postpack machinery, no Windows breakage, and I verified the guard now fails closed (a failing prepack puts `{"error":...}` on stdout on npm 11, and your error branch catches it). I resolved those threads.

What's left is small. Two real asks and two nits:

1. **Add `--ignore-scripts` to the pack call** (replied on the open build thread). Cuts the second compile on all 9 matrix jobs and removes the stdout-interleaving fragility in one change.
2. **Update `test/packages/README.md`.** It now says the opposite of what the code does: "The tarball test is not wired into this script", the two-flavors intro, and the file tree all predate this change. That README was kept current when `npm-pack/` was added, so it's worth the five minutes.
3. Nit: the guard regex is narrower than the glob it protects (inline comment).
4. Nit: a pointer comment linking the two pattern lists (inline comment).

One more thing before merge: the PR description still describes the prepack-prune/postpack design from the first revision. Since the body becomes the permanent record for this packaging change, please rewrite it to describe the files-negation approach.

maxmilian:
Thanks @yonib05 — all three addressed in 4134f14a.

- **`--ignore-scripts` on the pack call** (not a nit, agreed — it hit every ts-test matrix job). `test:package` now runs `npm pack --dry-run --ignore-scripts --json`. The leading `npm run build` already produces `dist`, and npm still evaluates the `files` globs with scripts off, so the assertion tests the real tarball. Verified against the two runs:
  - *without* `--ignore-scripts`: `> prepack` → `> tsc --project src/tsconfig.json` fires a second compile inside `npm pack`.
  - *with* it: no prepack/tsc lifecycle line — one build total. As you noted, this also removes the prepack-stdout-in-the-pipe class, so the interleaving thread is moot too (the JSON.parse guard stays as defense-in-depth). End-to-end `npm run test:package` still green: `[package-contents] OK (740 files)`.
- **Regex mirrors the glob** — narrowed to `/\.test\./`, matching `!dist/**/*.test.*` exactly. Confirmed it now flags `a.test.mjs` / `a.test.cjs` / `a.test.d.mts` (which the old three-suffix pattern missed) while leaving `latest.js` untouched.
- **Sync comment** — added above `forbiddenPatterns` pointing at the `files` negations. One caveat on the "vice versa": `package.json` must stay strict JSON and the `files` array only holds globs, so an inline pointer there would be read as a file pattern — the note lives on the JS side only.

strands-ts/package.json
yonib05: Fixed in 63429a0 with --ignore-scripts. One build per test:package run now.

strands-ts/test/packages/assert-package-contents.js
yonib05: Resolved by --ignore-scripts in 63429a0. No lifecycle script runs inside the pipe anymore, so the JSON on stdout stays clean.

strands-ts/test/packages/assert-package-contents.js
yonib05: Fixed in 63429a0. The regex is now /\.test\./ and matches the glob.

strands-ts/test/packages/assert-package-contents.js
yonib05: Added in 63429a0.

yonib05 (approved):

PR #3077, opened 2026-07-02
fix(multiagent): buffer graph node printer output to prevent interleaving

chaynabors:
Concurrent graph nodes each write to stdout through their own agent printer, so sibling agents interleave character by character. When the topology allows parallel execution, Graph now passes bufferOutput to each AgentNode.stream call, and the node swaps the agent's printer for a buffer that flushes on completion.

Closes #2514.

### DIFF
--- a/strands-ts/src/multiagent/__tests__/graph.test.ts
+++ b/strands-ts/src/multiagent/__tests__/graph.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it, vi } from 'vitest'
 import { Agent } from '../../agent/agent.js'
+import { AgentPrinter } from '../../agent/printer.js'
 import { MockMessageModel } from '../../__fixtures__/mock-message-model.js'
 import { MockSnapshotStorage } from '../../__fixtures__/mock-storage-provider.js'
 import { collectGenerator } from '../../__fixtures__/model-test-helpers.js'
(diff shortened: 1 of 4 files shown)


strands-ts/src/multiagent/__tests__/graph.test.ts
```diff
@@ -716,6 +717,65 @@ describe('Graph', () => {
...
+          if (event.nodeId === 'b') seenB.push((b as unknown as { _printer: AgentPrinter })._printer)
+        }
+      }
+
+      expect(seenA.some((p) => p !== printerA)).toBe(true)
+      expect(seenB.some((p) => p !== printerB)).toBe(true)
+      expect((a as unknown as { _printer: AgentPrinter })._printer).toBe(printerA)
+      expect((b as unknown as { _printer: AgentPrinter })._printer).toBe(printerB)
+    })
+
```
gautamsirdeshmukh, line 778: These tests are definitely helpful for verifying the printer-buffer-printer logic works at the code level, but I think we could use one that actually runs two nodes in parallel to confirm the final output isn't interleaved.  

Basically a copy of the existing `executes parallel graph (A -> B, A -> C) with B and C after A` test, but pointing B and C's printers at one shared list of writes. Then we'd just assert that each node's output is a contiguous block + never interleaved

gautamsirdeshmukh (dismissed):

strands-ts/src/multiagent/__tests__/graph.test.ts
chaynabors: Added a fan-out A->B, A->C test that points B and C at a shared write log and asserts each node's writes land as one contiguous slice. If buffering regresses, chunks interleave and the contiguity check fails.

gautamsirdeshmukh (approved):

Issue #3044, opened 2026-07-01
[FEATURE] Unified Storage Interface


yonib05:
## Summary

Introduce a single, unified storage interface that context management, session persistence, and offloading all target — instead of each primitive carrying its own bespoke storage wiring. This is a prerequisite for the new ContextManager interface (see companion issue).

## Problem

The SDK has several independent storage paths that don't share one contract:
- **Context offloader backends** — `Storage` protocol with `InMemoryStorage` / `FileStorage` / `S3Storage` implementations (`strands-py/src/strands/vended_plugins/context_offloader/storage.py`; `strands-ts/src/vended-plugins/context-offloader/storage.ts`).
- **Session backends** — `SessionRepository` + `FileSessionManager` / `S3SessionManager` in Python (`strands-py/src/strands/session/`); snapshot-based session manager in TS (`strands-ts/src/session/session-manager.ts`).
- Conversation/checkpoint state, persisted separately again.

Each defines its own notion of "where bytes go," so a backend written for offloading can't be reused for sessions, and every new storage-backed feature re-implements InMemory/File/S3.

## Goal

Define one storage contract that any backend (in-memory, file, S3, or custom) implements once and that context management, offloading, and sessions can all compose. Adapt the existing backends to it without breaking current behavior.

## Deliverables

- [ ] Define the unified storage interface (both SDKs), reconciling the existing `Storage` protocol and `SessionRepository` shapes.
- [ ] Adapt existing backends (InMemory/File/S3) to the unified interface.
- [ ] Migration path that keeps current offloader/session behavior working.
- [ ] Note the deliberate cross-language divergence (see below) so the interface fits both without forcing one to mimic the other.
- [ ] Align with the team on the interface before broad adoption.

## Cross-language divergence to respect

The session models differ deliberately and the interface should accommodate both rather than force parity:
- **Python:** `SessionRepository` CRUD over per-message `SessionMessage` rows (with change-detection-gated sync).
- **TypeScript:** snapshot-based — mutable `snapshot_latest` plus append-only immutable UUID-v7 history, with configurable save strategies.

Related: #1679 (bridge ConversationManager and SessionManager), #1392 (apply_management via hooks), #2548 (content-aware offloading), #1230 (checkpoint session-state model).

## Code pointers (pinned to `6ab71c6`)

[...]

Issue #1073, opened 2025-10-23
[FEATURE] Add Dapr statestore as session manager


filintod:
### Problem Statement

I would like to be able to use dapr statestore as my session manager. Dapr is a CNCF approved project that helps with separation of concerns between developers and platform engineering, it also provides tracing, encryption and metrics out of the box for storing data.  With it one can use DynamoDb as the session manager if we wanted to, or a simpler redis one.

### Proposed Solution

Add DaprSessionManager as one of the options for session manager.  The dependency would be optional so the code brought by the dapr sdk would not be required if not using it.

### Use Case

A developer can develop their application with an inmemory dapr component and on dev use a redis component, where in prod they can use DynamoDB.   They can also set TTL or encryption options for the data depending on requirements.  Platform engineer only need to provide the developers the name of the component  that the developer can work with while they decide what is the most appropriate backend for the job and infra

Issue #57, opened 2025-05-20
[FEATURE] Middle-ware driven session management


niklas-palm:
## Problem
Currently, the Strands SDK requires developers to manually implement session persistence logic for agents. This presents several challenges:

- Developers must write boilerplate code to save and load agent state
- Session management logic is tightly coupled with application code
- No standardized approach for implementing different storage backends
- Difficult for the community to share and reuse storage implementations
- Increased cognitive load when managing complex applications with multiple agents

[Docs for current strategy for saving and loading state](https://strandsagents.com/0.1.x/user-guide/concepts/agents/sessions-state/#2-serialization-and-restoration)

## Decorator-Based Session Management

I propose implementing a decorator-based session persistence layer that:

- Automatically handles saving and loading agent states
- Provides an extensible interface for storage providers
- Simplifies application code by separating concerns
- Enables community contributions of storage backends

### Use Case

Developers building chat applications need to persist conversations across page refreshes, server restarts, and user sessions. With the decorator approach, they can simply:

```python
@redis_session_manager(redis_url="redis://localhost:6379", ttl=3600)
def create_agent(system_prompt="You are a helpful assistant", messages=None):
    return Agent(system_prompt=system_prompt, messages=messages or [])

@app.post("/chat/{user_id}")
async def chat_endpoint(user_id: str, request: ChatRequest):
    # Create or restore agent using the user_id as the session ID
    agent = create_agent(
        session_id=f"user:{user_id}",
        system_prompt="You are a helpful AI assistant."
    )
    
    # Process message - state automatically saved to Redis
    response = agent(request.message)
    
    return {"response": response}
```

### Alternatives Solutions

Instead of decorators, one possible alternative solution would be to directly enhance the Agent class with built-in session management capabilities.

```python
from strands.storage import RedisStorage

# Create agent with Redis persistence
agent = Agent(
    system_prompt="You are a helpful assistant",
    session_id="user123",
    storage_provider=RedisStorage("redis://localhost:6379")
)

# Use normally - state automatically managed
response = agent("Hello!")
```

It would couple the storage logic more with the core Agent class, but also allow for the community to add storage providers by implementing this interface:

```python
class StorageProvider(ABC):
    @abstractmethod
    def session_exists(self, session_id): pass
    
    @abstractmethod
    def load_session(self, session_id): pass
    
    @abstractmethod
    def save_session(self, session_id, state): pass
```

austinmw:
[...]

Issue #33, opened 2025-05-17
[FEATURE] Add Database Session Management


puritatemcordis:
### Problem Statement

The current implementation of the Strands agent uses in-memory storage to maintain conversation history. While this is suitable for short-lived sessions, it lacks persistence across application restarts or crashes. To address this limitation, it would be valuable to implement a class that enables saving conversation history to a relational database such as PostgreSQL or MySQL. This enhancement would provide a more robust and reliable foundation for production-grade applications by ensuring conversation state can be preserved and recovered when needed.

### Use Case

**Objective**:
Enable the Strands agent to persist conversation history in a relational database to ensure continuity and resilience in production environments.

**Primary Actor**:
Strands Agent

**Stakeholders and Interests**:

**Developers**: Want reliable state management to build robust applications.
**End Users**: Expect seamless conversation experiences even after interruptions.
**Operations/DevOps**: Need observability and data recovery mechanisms in case of system failures.
Preconditions:

A relational database (e.g., PostgreSQL, MySQL) is available and configured.
The Strands agent is running in a production or long-lived session environment.

**Main Flow**:

A user interacts with the Strands agent.
The agent processes user input and generates responses.
Each turn in the conversation (user input and agent response) is saved to the relational database in real-time or at regular intervals.
If the application crashes or restarts, the agent retrieves the saved conversation history from the database.
The agent resumes the conversation with full context, ensuring continuity for the user.

**Postconditions**:

Conversation history is persistently stored in the database.
The agent can recover and resume sessions after unexpected terminations.

**Alternative Flows**:

If the database is unreachable, the agent logs the failure and continues with in-memory storage, optionally retrying persistence later.

rajib76:
Can we not use something like below for the same. It uses files, but the logic can be changed to put the state into a DB.

```
def save_agent_state(agent, session_id):
    os.makedirs("sessions", exist_ok=True)

    state = {
        "messages": agent.messages,
        "system_prompt": agent.system_prompt
    }
    # Store state (e.g., database, file system, cache)
    with open(f"sessions/{session_id}.json", "w") as f:
        json.dump(state, f)
```

DennisTraub:
[...]

PR #3080, opened 2026-07-02
feat(designs-storage): add storage design doc proposal

lizradway:
## Description
Storage Design Doc Proposal

## Related Issues

https://github.com/strands-agents/harness-sdk/issues/3044

## Documentation PR

This is a design doc

## Type of Change

New feature

### DIFF
--- a/team/designs/0014-storage.md
+++ b/team/designs/0014-storage.md
@@ -1,0 +1,388 @@
+# Unified Storage Primitive
+
+**Status**: Proposed
+
+**Date**: 2026-06-29
+
+**Issue**: TBD
+
+---
+
+<details>
+<summary><strong>Definitions</strong></summary>
+
+| Term | Status | Description |
+|------|--------|-------------|
+| Context Offloader | Shipped | Plugin that caches oversized tool results. Has its own `Storage` interface (`store`/`retrieve`). |
+| Session Manager | Shipped | Manages conversation snapshots. Has its own `SnapshotStorage` interface (6 methods). |
+| Memory Stores | Shipped | `LocalMemoryStore` persists to disk via raw `node:fs`. No storage abstraction. |
+| Context Manager (v2) | Proposed | Agentic context management with L1 history batches. Storage param planned but not yet implemented. |
+| Transcripts | Proposed | Immutable audit log of agent interactions. Not yet implemented. |
+
+</details>
+
+---
+
+[Problem](#problem) · [Proposal](#proposal) · [API](#proposed-api) · [DX](#developer-experience)
+
+Appendices: [Interface Mapping](#a-interface-mapping) · [Migration](#b-migration-path) · [Python](#c-python-parity) · [Naming](#d-naming-rationale) · [Query](#e-future-query-extension) · [DX Examples](#f-developer-experience-walkthrough) · [Full Plan](#g-development-plan) · [Alternatives](#h-alternatives-considered)
+
+---
+
+## Problem
+
+The SDK requires data persistence in at least five independent subsystems — tool result caching, session snapshots, long-term memory, context history, and transcripts. Today each subsystem defines its own storage abstraction and its own implementations. This creates real cost for users, contributors, and the SDK team.
+
+### Current State
+
+Today each subsystem either defines its own storage interface or uses raw `node:fs` calls directly. There is no shared contract. What concretely breaks:
+
+1. **Every new SDK feature re-invents storage.** When the team builds a new subsystem that persists data (transcripts, context manager v2), they must write `FileStorage` + `S3Storage` + `InMemoryStorage` from scratch for that subsystem. That's not shared code with existing storage implementations — it's net-new classes, net-new tests, net-new bugs. This is real engineering time multiplied across every feature we ship.
+
+2. **Adding a new backend costs 5× what it should.** Anyone who wants a backend the SDK doesn't ship — whether a community contributor adding Redis support or a user implementing their own DynamoDB adapter — must implement five separate interfaces with different method signatures, naming conventions, and semantics. In practice, nobody contributes storage backends and most users give up and stay on the defaults.
+
+3. **Production deployment requires per-subsystem migration.** Moving from local development to production means swapping storage configuration in every subsystem independently. There's no single place to change "use S3 now" — each subsystem has its own storage constructor that must be found and updated.
+
+4. **Maintenance cost scales multiplicatively.** Every new subsystem adds another set of storage implementations to maintain. Every new backend adds another implementation per subsystem. Without shared code, a bug fixed in one doesn't fix the others.
+
+### What "run an agent in the browser" looks like today vs. with unified Storage
+
+A user wants to run an agent in a browser tab with persistent sessions, memory, and context. The existing implementations assume a Node.js server environment — `FileStorage` uses `node:fs`, `S3Storage` uses server-side AWS credentials. Neither works in a browser without significant modification. They need IndexedDB.
+
+**Today:**
+```
+1. Read context-offloader Storage interface → implement IndexedDBOffloaderStorage
+2. Read session SnapshotStorage interface → implement IndexedDBSnapshotStorage
+3. Read LocalMemoryStore source → figure out fs calls → implement IndexedDBMemoryBackend
+4. Read context manager storage (once it exists) → implement IndexedDBContextStorage
+5. Read transcript storage (once it exists) → implement IndexedDBTranscriptStorage
+   Total: 5 classes, 5 different interfaces, ~500 LOC
+```
+
+**With unified Storage:**
+```
+1. Implement Storage (put/get/delete/list) backed by IndexedDB → done
+   Total: 1 class, 4 methods, ~50 LOC
+```
+
+The same applies to any backend the SDK doesn't ship — GCP, Azure, Redis, DynamoDB, mandated enterprise infrastructure. Today the SDK only ships AWS storage backends, so users on anything else either don't use persistence features or give up. Unified storage makes it practical for the community to add their own backends, making the SDK truly cloud-agnostic.
+
+Transcripts and context manager v2 both need a storage interface. We're going to design one for them either way — the question is whether we design bespoke interfaces that further increase the `subsystems × backends` problem, or design the unified one first and build them on it from day one, changing the scaling model to `subsystems + backends`.
+
+---
+
+## Goals and Non-Goals
+
+**Goals:**
+- One low-level interface for blob persistence that all subsystems build on
+- Configure once, pass everywhere
+- Cross-SDK parity
+- Backwards-compatible migration
+
+**Non-goals:**
+- Replacing `MemoryStore` (implementations like `LocalMemoryStore` *use* a `Storage` internally for persistence but don't implement it — see [FileMemoryStore design](https://github.com/strands-agents/harness-sdk/pull/2895))
+- Replacing the `Sandbox` abstraction (`Sandbox` handles execution isolation; `LocalFileStorage` can route I/O through one via `forSandbox()`, but they're different layers — see [PR #2649](https://github.com/strands-agents/harness-sdk/pull/2649))
+- Transactions or locking, query or indexing, and streaming reads/writes — all of which can be added as optional extensions later without breaking the base interface
+
+---
+
+## Proposal
+
+### Recommended: Unified `Storage` Interface
+
+Reclaim the `Storage` name for the SDK-wide primitive. One interface with four operations — `put`, `get`, `delete`, `list` — representing what every storage backend natively supports. Higher-level operations (atomic JSON snapshots, eviction, manifests) are layered by consumers. See [Proposed API](#proposed-api) for the full interface definition.
+
+An optional `query` method for metadata filtering is planned but out of scope for this doc — see [Appendix E](#e-future-query-extension).
+
+**Type hierarchy:**
+
+```
+Storage (interface)
+├── InMemoryStorage (implements Storage directly)
+└── FileStorageBase (abstract, implements Storage — shared path/key logic)
+    ├── LocalFileStorage (+ forSandbox())
+    ├── S3Storage
+    └── GithubStorage
+```
+
+`FileStorageBase` is an abstract base class that provides shared logic for file-path-based backends: key/path normalization and key sanitization. Concrete implementations (`LocalFileStorage`, `S3Storage`, `GithubStorage`) extend it with backend-specific I/O. `LocalFileStorage` additionally exposes `forSandbox()` for sandbox-backed I/O. `InMemoryStorage` implements `Storage` directly — it has no file paths or sandbox concept.
+
+**Shipped implementations:**
+
+| Class | Backend |
+|-------|---------|
+| `InMemoryStorage` | `Map<string, entry>` (optional eviction: `evictAfterTurns`, defaults to 20) |
+| `LocalFileStorage` | Local filesystem (extends `FileStorageBase`) |
+| `S3Storage` | AWS S3 (extends `FileStorageBase`) |
+
+All exported from `@strands-agents/sdk/storage`. `S3Storage` lazy-imports `@aws-sdk/client-s3` (optional peer dep) — users who don't use it never need the AWS SDK installed.
+
+**Future / community backends:**
+
+| Backend | Notes |
+|---------|-------|
+| GithubStorage | GitHub-backed (extends `FileStorageBase`) — part of Maisie's intern project |
+| DynamoDB | Single-table design, native metadata filtering via `query` |
+| Redis | Low-latency, good for ephemeral or cache-heavy workloads |
+| GCS | Google Cloud equivalent of S3 (extends `FileStorageBase`) |
+| Azure Blob Storage | Azure equivalent of S3 (extends `FileStorageBase`) |
+| SQLite | Embedded, zero-config, single-file persistence |
+| PostgreSQL / MySQL | Existing infrastructure, ACID guarantees |
+| IndexedDB | Browser-compatible storage for client-side agents |
+
+Any first-class or community implementation of `Storage` works with every subsystem that accepts it — sessions, memory, transcripts, context manager, and offloader all benefit without per-backend adapters.
+
+### Top-Level `storage` on Agent
+
+`Agent` accepts a top-level `storage` parameter. All subsystems (sessions, memory, context manager, transcripts, offloader) inherit it automatically unless they provide their own override. This includes the auto-added `ContextOffloader` — today it defaults to `InMemoryStorage`, but with a top-level `storage` configured it uses that instead:
+
+```typescript
+const agent = new Agent({
+  model,
+  storage: new LocalFileStorage({ rootPath: './.agent-data' }),
+})
+// Sessions, memory, transcripts, context, offloader — all use LocalFileStorage automatically.
+```
+
+Per-subsystem overrides are still possible:
+
+```typescript
+const agent = new Agent({
+  model,
+  storage: new LocalFileStorage({ rootPath: './.agent-data' }),
+  sessionManager: { storage: new S3Storage({ bucket: 'sessions' }) },  // sessions go to S3
+})
+```
+
+### How Each Subsystem Uses `Storage`
+
+Every subsystem accepts a `Storage` instance and uses the same pattern: serialize data, `put` it at a namespaced key, `get` it back later. No subsystem-specific storage interfaces needed.
+
+| Subsystem | Prefix | Operations used |
+|-----------|--------|-----------------|
+| Context Offloader | `offloader/` | `put`, `get` |
+| Session Manager | `sessions/` | `put`, `get`, `list`, `delete` |
+| Memory Stores | `memory/` | `put`, `get` |
+| Context Manager | `context/` | `put`, `get`, `list` |
+| Transcripts | `transcripts/` | `put`, `list`, `delete` |
+
+Each subsystem owns its key prefix — no coordination needed between them beyond not colliding (enforced by convention above).
+
+**Breaking changes:** None. Existing subsystem APIs (`SessionManager`, `ContextOffloader`, etc.) widen their `storage` config to accept the unified `Storage` type alongside their current interfaces. Existing code continues to work unchanged — see [Appendix B](#b-migration-path) for the deprecation timeline.
+
+---
+
+### Sandbox Relationship
+
+`Sandbox` and `Storage` are different layers. `Sandbox` handles execution isolation (running code, agent-initiated file I/O). `Storage` handles persistence (data that outlives a turn or session). They don't replace each other.
+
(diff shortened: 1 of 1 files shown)


team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+})
+// Sessions, memory, transcripts, context, offloader — all use LocalFileStorage automatically.
+```
+
+Per-subsystem overrides are still possible:
+
+```typescript
+const agent = new Agent({
+  model,
+  storage: new LocalFileStorage({ rootPath: './.agent-data' }),
```
JackYPCOnline, line 152: Is that means it would be benifit when user needs >3 Plugins + at least 2 plugins shares at least 2 storage.

Wondering if we have customer request?

team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+// Sandbox-backed: routes all I/O through the agent's sandbox
+const sandboxedStorage = storage.forSandbox(agent.sandbox)
+```
+
+---
+
+## Proposed API
+
+```typescript
+export interface Storage {
```
Unshure, line 194: How does namespacing work? Can different users of storage (Session, memory, etc) collide with each other?

team/designs/0014-storage.md
lizradway: Quite a few one-off customer requests for 3rd party implementations of these custom storage attributes.

DDB for instance: https://github.com/search?q=repo%3Astrands-agents%2Fharness-sdk+dynamodb&type=issues

but imo, for why tackling this now, this is more of a maintainability concern imo that is super smelly and something we might as well fix before we add 2 more storage abstractions onto the problem. the customer extensibility is an added plus.

team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+  put(key: string, data: Uint8Array): Promise<void>
+  get(key: string): Promise<Uint8Array | null>
```
Unshure, line 196: Will this support a cached read? If we are reading a blob, do we read it from disk once, then get the individual keys? When do we re-read from disk?

team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+Swapping to S3 for production is a one-line change — replace `LocalFileStorage` with `S3Storage`. Custom backends implement the four-method `Storage` interface and work with every subsystem automatically.
+
+Each subsystem accepts a `Storage` instance, so you can configure them independently:
+
+```typescript
+const agent = new Agent({
+  model,
+  sessionManager: { storage: new S3Storage({ bucket: 'sessions' }) },
+  contextManager: { storage: new LocalFileStorage({ rootPath: './.context' }) },
+  transcript: { storage: new S3Storage({ bucket: 'audit-logs' }) },
```
Unshure, line 229: I think everything is a plugin now right? If so, is storage just something that any plugin can define for itself? Do we have a list of these things that are generic for all plugins?

team/designs/0014-storage.md
lizradway: some quick wins:
- https://github.com/strands-agents/harness-sdk/issues/1073
- https://github.com/strands-agents/harness-sdk/issues/57
- https://github.com/strands-agents/harness-sdk/issues/33

team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+├── InMemoryStorage (implements Storage directly)
+└── FileStorageBase (abstract, implements Storage — shared path/key logic)
+    ├── LocalFileStorage (+ forSandbox())
+    ├── S3Storage
+    └── GithubStorage
+```
+
+`FileStorageBase` is an abstract base class that provides shared logic for file-path-based backends: key/path normalization and key sanitization. Concrete implementations (`LocalFileStorage`, `S3Storage`, `GithubStorage`) extend it with backend-specific I/O. `LocalFileStorage` additionally exposes `forSandbox()` for sandbox-backed I/O. `InMemoryStorage` implements `Storage` directly — it has no file paths or sandbox concept.
+
+**Shipped implementations:**
```
Unshure, line 110: How does agentcore memory session storage work here? I remember having to do some optimizations based on the agentcore memory api that I didnt love. Not sure how that becomes generic for everyone

team/designs/0014-storage.md
lizradway: distinct key prefixes as illustrated in the table: https://github.com/strands-agents/harness-sdk/pull/3080/changes#diff-796a22a5d15df461d877ff7ab03d038ac8c220fced5ffc9af98b203b9b232b11R161

happy to implement a more intelligent collision prevention mechanism if you have any suggestions

team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+Any first-class or community implementation of `Storage` works with every subsystem that accepts it — sessions, memory, transcripts, context manager, and offloader all benefit without per-backend adapters.
+
+### Top-Level `storage` on Agent
+
+`Agent` accepts a top-level `storage` parameter. All subsystems (sessions, memory, context manager, transcripts, offloader) inherit it automatically unless they provide their own override. This includes the auto-added `ContextOffloader` — today it defaults to `InMemoryStorage`, but with a top-level `storage` configured it uses that instead:
+
+```typescript
+const agent = new Agent({
+  model,
+  storage: new LocalFileStorage({ rootPath: './.agent-data' }),
```
poshinchen, line 142: Do we want to support multiple storages?

team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+| Context Manager (v2) | Proposed | Agentic context management with L1 history batches. Storage param planned but not yet implemented. |
+| Transcripts | Proposed | Immutable audit log of agent interactions. Not yet implemented. |
+
+</details>
+
+---
+
+[Problem](#problem) · [Proposal](#proposal) · [API](#proposed-api) · [DX](#developer-experience)
+
+Appendices: [Interface Mapping](#a-interface-mapping) · [Migration](#b-migration-path) · [Python](#c-python-parity) · [Naming](#d-naming-rationale) · [Query](#e-future-query-extension) · [DX Examples](#f-developer-experience-walkthrough) · [Full Plan](#g-development-plan) · [Alternatives](#h-alternatives-considered)
```
Unshure, line 28: Is there a plan for python parity? Not sure where we are standing in that regard? Does the migration path for python look significantly different for the typescript path to implement storage?

team/designs/0014-storage.md
lizradway: interesting -- i dont think the base interface should implement any caching, but likely implementations of it should as they see fit

team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+**Type hierarchy:**
+
+```
+Storage (interface)
+├── InMemoryStorage (implements Storage directly)
+└── FileStorageBase (abstract, implements Storage — shared path/key logic)
+    ├── LocalFileStorage (+ forSandbox())
+    ├── S3Storage
+    └── GithubStorage
+```
```
gautamsirdeshmukh, line 106: I think this interface-implementation pattern makes sense, feels similar to how we set up Sandbox/ShellSandbox/etc. 

Now the naming for the `Storage / FileStorageBase / LocalFileStorage` evolution... thoughts on just `FileStorage` for the middle abstraction? (total nit of course)

team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+| Redis | Low-latency, good for ephemeral or cache-heavy workloads |
+| GCS | Google Cloud equivalent of S3 (extends `FileStorageBase`) |
+| Azure Blob Storage | Azure equivalent of S3 (extends `FileStorageBase`) |
+| SQLite | Embedded, zero-config, single-file persistence |
+| PostgreSQL / MySQL | Existing infrastructure, ACID guarantees |
+| IndexedDB | Browser-compatible storage for client-side agents |
+
+Any first-class or community implementation of `Storage` works with every subsystem that accepts it — sessions, memory, transcripts, context manager, and offloader all benefit without per-backend adapters.
+
+### Top-Level `storage` on Agent
```
mkmeral, line 135: do I want a single storage? also why do i add another storage if i already have sandbox? how do they interact?

team/designs/0014-storage.md
Unshure: In my mind, this is the real value of a low-level storage primitive. All of the optimization that needed to be done for agentcore memory only needs to be done once, then all plugins that need persistence benefit from it

team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+
+```typescript
+const agent = new Agent({
+  model,
+  storage: new LocalFileStorage({ rootPath: './.agent-data' }),
+  sessionManager: { storage: new S3Storage({ bucket: 'sessions' }) },  // sessions go to S3
+})
+```
+
+### How Each Subsystem Uses `Storage`
```
mkmeral, line 157: As an agent developer I would NOT want my agent to use context offloader and transcripts in the same storage. I want context offloader (and skills for example) to work with sandbox storage. but i do want a separate memory/session manager for my transcript storage

team/designs/0014-storage.md
lizradway: oooh like an `initStorage`?

i think right now plugins only interact with initAgent / addTools / addHooks, so storage could fit well i think tbh.

team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+```
+
+Swapping to S3 for production is a one-line change — replace `LocalFileStorage` with `S3Storage`. Custom backends implement the four-method `Storage` interface and work with every subsystem automatically.
+
+Each subsystem accepts a `Storage` instance, so you can configure them independently:
+
+```typescript
+const agent = new Agent({
+  model,
+  sessionManager: { storage: new S3Storage({ bucket: 'sessions' }) },
```
JackYPCOnline, line 227: For me storage is more like Implementation details attached to the Plugin itself, since each has different Apis. so I agree that the interface is shared, the implementations aren't.

team/designs/0014-storage.md
lizradway: parity path forward in appendix https://github.com/strands-agents/harness-sdk/pull/3080/changes#diff-796a22a5d15df461d877ff7ab03d038ac8c220fced5ffc9af98b203b9b232b11R267 :)

team/designs/0014-storage.md
Unshure: I think plugins have unique names right? We could probably just use that as the prefix

team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+
+---
+
+<details>
+<summary><strong>A: Interface Mapping</strong></summary>
+
+| Subsystem | Current | Unified |
+|---|---|---|
+| Offloader | `store(key, content, contentType)` | `put(key, encode(content))` |
+| Offloader | `retrieve(reference)` | `get(reference)` |
```
poshinchen, line 243: My AI friend told me that these are not the same.

>  put returns void, but the offloader's store returns a vended reference. 
 The reference is not the key — the storage layer owns identity generation. The design's mapping `(retrieve(reference) → get(reference))` assumes reference == key, which isn't true today. Either the offloader changes who generates identity, or put needs to return a handle. The doc doesn't acknowledge this.

team/designs/0014-storage.md
lizradway: that's actually what i had originally, but it felt confusing for me with LocalFileStorage. open to either tbh. can decide in api bar raising

team/designs/0014-storage.md
Unshure: Maybe caching was the wrong word. When does storage decide to re-read information from disk? Until it re-reads, the information in memory is essentially the cached version of what was in disk.

team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+
+---
+
+### Sandbox Relationship
+
+`Sandbox` and `Storage` are different layers. `Sandbox` handles execution isolation (running code, agent-initiated file I/O). `Storage` handles persistence (data that outlives a turn or session). They don't replace each other.
+
+The connection point: `LocalFileStorage` exposes a `forSandbox()` method that returns a new instance routing all I/O through the agent's sandbox instead of raw `node:fs`. This is only used by subsystems that operate within the agent's execution environment (today: the context offloader, which stores tool results generated inside the sandbox). Subsystems that persist durable data outside execution (sessions, memory, transcripts) write directly to the host filesystem. The pattern already exists in the context-offloader (see PR [#2649](https://github.com/strands-agents/harness-sdk/pull/2649)) and carries forward unchanged with the unified `Storage` interface.
```
mkmeral, line 179: I'm not sure if I understand it. why does storage use sandbox? I'd expect sandbox to implement a storage, no? 

The API you proposed is really similar to  get/put/list/delete files in sandbox?

team/designs/0014-storage.md
Unshure: Im pretty out of the loop, so dont make the decision just from my input, but I like that idea!

team/designs/0014-storage.md
Unshure: Ahh thanks! Link was broken

team/designs/0014-storage.md
lizradway: >do I want a single storage? 
i would assume that this is the production 80% use case

> also why do i add another storage if i already have sandbox? how do they interact?
see https://github.com/strands-agents/harness-sdk/pull/3080/changes#diff-796a22a5d15df461d877ff7ab03d038ac8c220fced5ffc9af98b203b9b232b11R175

sandbox is execution environment, storage is persistent

team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+
+### Current State
+
+Today each subsystem either defines its own storage interface or uses raw `node:fs` calls directly. There is no shared contract. What concretely breaks:
+
+1. **Every new SDK feature re-invents storage.** When the team builds a new subsystem that persists data (transcripts, context manager v2), they must write `FileStorage` + `S3Storage` + `InMemoryStorage` from scratch for that subsystem. That's not shared code with existing storage implementations — it's net-new classes, net-new tests, net-new bugs. This is real engineering time multiplied across every feature we ship.
+
+2. **Adding a new backend costs 5× what it should.** Anyone who wants a backend the SDK doesn't ship — whether a community contributor adding Redis support or a user implementing their own DynamoDB adapter — must implement five separate interfaces with different method signatures, naming conventions, and semantics. In practice, nobody contributes storage backends and most users give up and stay on the defaults.
```
mkmeral, line 42: what's the `5x` number here? what are the 5 interfaces I have to implement for DynamoDB?

team/designs/0014-storage.md
lizradway: yup! supported below

team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+
+Transcripts and context manager v2 both need a storage interface. We're going to design one for them either way — the question is whether we design bespoke interfaces that further increase the `subsystems × backends` problem, or design the unified one first and build them on it from day one, changing the scaling model to `subsystems + backends`.
+
+---
+
+## Goals and Non-Goals
+
+**Goals:**
+- One low-level interface for blob persistence that all subsystems build on
+- Configure once, pass everywhere
```
JackYPCOnline, line 78: How many users today configure 3+ persistence subsystems, and of those how many want the same backend for all?  Does effort vs value pay off

team/designs/0014-storage.md
lizradway: yup! that is how it is built right now. storage can be configured per primitive and currently only execution-level storage uses sandbox (ie. right now only the context offloader plugin)

team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+
+Appendices: [Interface Mapping](#a-interface-mapping) · [Migration](#b-migration-path) · [Python](#c-python-parity) · [Naming](#d-naming-rationale) · [Query](#e-future-query-extension) · [DX Examples](#f-developer-experience-walkthrough) · [Full Plan](#g-development-plan) · [Alternatives](#h-alternatives-considered)
+
+---
+
+## Problem
+
+The SDK requires data persistence in at least five independent subsystems — tool result caching, session snapshots, long-term memory, context history, and transcripts. Today each subsystem defines its own storage abstraction and its own implementations. This creates real cost for users, contributors, and the SDK team.
+
+### Current State
```
mkmeral, line 36: all of these items are the same thing that is repeated multiple times. (is this an homage to our sdks current state? lol)

but like what are these "multiple" items?

team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+    └── GithubStorage
+```
+
+`FileStorageBase` is an abstract base class that provides shared logic for file-path-based backends: key/path normalization and key sanitization. Concrete implementations (`LocalFileStorage`, `S3Storage`, `GithubStorage`) extend it with backend-specific I/O. `LocalFileStorage` additionally exposes `forSandbox()` for sandbox-backed I/O. `InMemoryStorage` implements `Storage` directly — it has no file paths or sandbox concept.
+
+**Shipped implementations:**
+
+| Class | Backend |
+|-------|---------|
+| `InMemoryStorage` | `Map<string, entry>` (optional eviction: `evictAfterTurns`, defaults to 20) |
```
poshinchen, line 114: we will allow users to configure it right?

team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+| Session | `deleteSession(...)` | `list(prefix)` + `delete(key)` |
+| Session | `saveManifest(...)` / `loadManifest(...)` | `put` / `get` |
+| Memory | `_flush(records)` / `_readFromDisk()` | `put` / `get` |
+| Context (v2) | batch write / history browse | `put` / `list` |
+| Transcripts | append / list / delete | `put` / `list` / `delete` |
+
+</details>
+
+<details>
+<summary><strong>B: Migration Path</strong></summary>
```
mkmeral, line 256: what do we gain for this migration?

team/designs/0014-storage.md
lizradway: yes: session manager, offloader, memory stores currently exist with context manager and transcripts planned.

see definitions above: https://github.com/strands-agents/harness-sdk/pull/3080/changes#diff-796a22a5d15df461d877ff7ab03d038ac8c220fced5ffc9af98b203b9b232b11R12

team/designs/0014-storage.md
lizradway: see [definitions](https://github.com/strands-agents/harness-sdk/pull/3080/changes#diff-796a22a5d15df461d877ff7ab03d038ac8c220fced5ffc9af98b203b9b232b11R12) and [explicit example of implementing a new storage interface across the interfaces](https://github.com/strands-agents/harness-sdk/pull/3080/changes#diff-796a22a5d15df461d877ff7ab03d038ac8c220fced5ffc9af98b203b9b232b11R54-R59)

team/designs/0014-storage.md
gautamsirdeshmukh, line 179: Interested to know the motivation for adding this `forSandbox` method. Is this to make it easier for multiple agents to use the same sandbox storage, without explicitly configuring them with that sandbox? (in terms of routing tool calls to the sandbox as well)

team/designs/0014-storage.md
lizradway: maybe @joychak would have some insight on this, but id imagine this is the default 80% production usecase. 

anyways i see the reason for doing this as mainly a maintainability reason for us devs rather than a customer reason (although there are many customer wins and asks)

team/designs/0014-storage.md
mkmeral: also what's the effort? this seems like a multi-step migration

team/designs/0014-storage.md
lizradway: yeah exactly! would just extend for an AgentCoreStorage id imagine to be used throughout the SDK. could possibly even live in agent core's repo instead of ours

team/designs/0014-storage.md
poshinchen: I meant list of storages: `storage: []`

team/designs/0014-storage.md
lizradway: yeah true plugin name spacing is already guaranteed unique in agent state!

team/designs/0014-storage.md
lizradway: sandbox could implement storage, but you're asserting that the sandbox is your persistence backend” which breaks when the sandbox is ephemeral

team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+An optional `query` method for metadata filtering is planned but out of scope for this doc — see [Appendix E](#e-future-query-extension).
+
+**Type hierarchy:**
+
+```
+Storage (interface)
+├── InMemoryStorage (implements Storage directly)
+└── FileStorageBase (abstract, implements Storage — shared path/key logic)
+    ├── LocalFileStorage (+ forSandbox())
+    ├── S3Storage
```
arronbailiss, line 104: we had issues in the past with overheads on S3 calls, e.g. session storage making multiple sequential GETs when loading a session which took several seconds. AC Memory has had similar issues too. Need to consider things like provider throttling as well, e.g. S3

team/designs/0014-storage.md
```diff
@@ -0,0 +1,388 @@
...
+    └── GithubStorage
```
arronbailiss, line 105: Git or Github?

team/designs/0014-storage.md
lizradway: good call out & can make sure to address these in any remote storage option (ie. S3).

the great part of this, is that now we can optimize S3 behavior once in `S3Storage` and every subsystem benefits woo!

team/designs/0014-storage.md
lizradway: Github

team/designs/0014-storage.md
lizradway: yup!

team/designs/0014-storage.md
lizradway: oh! yes from discussion looks like we want to support this

lizradway: i do think this can be a fast follow though and not block initial impl on this

team/designs/0014-storage.md
lizradway: its for per agent file isolation when a shared `LocalFileStorage` is passed to a plugin that serves multiple agents with sandboxing. without it, all agents would read/write the same files.

team/designs/0014-storage.md
lizradway: yup! there are separate implementations (ie. file, s3, etc) on top of which each plugin can implement their own storage mechanisms if necessary. in an ideal world, unless a specific storage type for a plugin is heavily bloated, we wouldn't need storage logic in the plugin.

team/designs/0014-storage.md
lizradway: it should re-read from disk each time unless it saves a copy in memory.

team/designs/0014-storage.md
lizradway: ~1 week dev time imo

see customer issues linked above and n*m maintenance becomes n+m

team/designs/0014-storage.md
lizradway: good callout - tbh i think it might make sense to update context offloader to use tool use id so that we are not constructing a new id.

mehtarac (approved):

PR #3087, opened 2026-07-03
fix(ts): respect store maxSearchResults during memory injection

### LINKED ISSUE #2996: [BUG] maxSearchResults doesn't work
### Checks

### SDK Language

TypeScript

### Strands Version

1.7.0

### Language Runtime Version

24.12.0

### Operating System

macOS 26.4

### Installation Method

npm

### Steps to Reproduce

I've noticed that maxSearchResults value is not taken into account.
It always returns 5 results no matter what I put as a value.

```
const kbStore = new BedrockKnowledgeBaseStore({
  name: 'knowledge-base',
  description: 'Company documentation, product info, and policy knowledge base.',
  maxSearchResults: 1,
  config: {
    knowledgeBaseId: process.env.KNOWLEDGE_BASE_ID!
  }
});
```

I noticed that once I instantiated MemoryManager and console.log entries in format:

```
const manager = new MemoryManager({
    stores,
    injection: {
      maxEntries: 10,
      format: ({ entries }) => {
        console.log(entries);
        return ...
      }
    }
  });
```

### Expected Behavior

It should return only a specified amount of items.

### Actual Behavior

It always returns 5.

### Additional Context

_No response_

### Possible Solution

_No response_

### Related Issues

_No response_

matthewchen94:
## Summary
- keep public/tool `MemoryManager.search()` behavior unchanged: explicit `maxSearchResults` still overrides store defaults
- make the injection path treat `maxEntries` as a fallback fetch size, while preserving a store-level `maxSearchResults` cap
- add a regression test for #2996 where injection `maxEntries: 10` must not override a store configured with `maxSearchResults: 1`

Fixes #2996.

## Validation
- `npx prettier --config /Users/chenchiyu/Desktop/open-source/harness-sdk/.prettierrc --check /Users/chenchiyu/Desktop/open-source/harness-sdk/strands-ts/src/memory/memory-manager.ts /Users/chenchiyu/Desktop/open-source/harness-sdk/strands-ts/src/memory/__tests__/memory-manager.test.ts`

Note: I used the GitHub API for the patch because full repository clone via codeload intermittently failed in this environment; the changed files are limited to the TypeScript memory manager and its focused unit test.

### DIFF
--- a/site/src/content/docs/user-guide/concepts/memory/overview.mdx
+++ b/site/src/content/docs/user-guide/concepts/memory/overview.mdx
@@ -117,7 +117,7 @@ Each store carries its own identity and behavior:
 |-------|---------|
 | `name` | Unique identifier, used to target the store from tools and the programmatic API. |
 | `description` | Human-readable summary, surfaced in the memory tool descriptions so the model knows what each store holds. |
-| <Syntax py="max_search_results" ts="maxSearchResults" /> | Default result cap per search when a caller does not pass one. The manager falls back to `3` if neither is set. |
+| <Syntax py="max_search_results" ts="maxSearchResults" /> | Default result cap per store search when a caller does not pass one. The manager falls back to `3` if neither is set. During injection, a store-level cap remains the authoritative per-store retrieval limit before injected entries are sliced. |
 | `writable` | Whether the store accepts writes. |
 
 The manager attaches each store's `name` to its results, so the model and your code can tell which store produced each entry and target follow-up queries.
@@ -200,7 +200,7 @@ agent = Agent(
 </Tabs>
 
 - `trigger` accepts `'userTurn'` (the default, inject only on a fresh user ask), `'everyTurn'` (inject before every model call, for autonomous agents), or a predicate: a function that receives the injection context and returns whether to inject this call.
-- <Syntax py="max_entries" ts="maxEntries" /> caps how many entries are retrieved and injected.
+- <Syntax py="max_entries" ts="maxEntries" /> caps how many retrieved entries are injected. During injection it is also the fallback fetch size for stores that do not define <Syntax py="max_search_results" ts="maxSearchResults" />; a store-level search cap takes precedence for that store's retrieval.
 - `query` overrides the adaptive default with your own query logic. Return an empty value to skip injection for this call.
 - `format` renders the retrieved entries. The default emits an escaped `<memory>` block; a custom formatter that emits markup owns its own escaping.
 
--- a/strands-ts/src/memory/__tests__/memory-manager.test.ts
+++ b/strands-ts/src/memory/__tests__/memory-manager.test.ts
@@ -795,10 +795,7 @@ describe('MemoryManager', () => {
       })
 
       it('uses store maxSearchResults before injection maxEntries when fetching injected memory', async () => {
-        const store = createMockStore('s', {
-          maxSearchResults: 1,
-          entries: [{ content: 'A' }, { content: 'B' }, { content: 'C' }],
-        })
+        const store = createMockStore('s', { maxSearchResults: 1 })
         vi.mocked(store.search).mockImplementation(async (_query, options) =>
           [{ content: 'A' }, { content: 'B' }, { content: 'C' }].slice(0, options?.maxSearchResults)
         )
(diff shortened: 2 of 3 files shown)


matthewchen94:
Quick status update: the PR is mergeable and the completed checks are green. The remaining two checks are waiting, not failing:

- Trigger Strands Review
- Run integration tests

This patch is limited to `strands-ts/src/memory/memory-manager.ts` plus one focused regression test for #2996. Happy to adjust if you prefer a different precedence model for injection `maxEntries` vs store-level `maxSearchResults`.

opieter-aws:
Thanks for your contribution! I see the problem you're solving, but I'd like to propose a different approach.

The `InternalMemorySearchOptions` mechanism introduces a hidden mode toggle through the public `search()`, which hides the mode from `search()`’s signature. We don't have precedence for this pattern in our SDK. The real distinction is override (public `search()`) vs fallback (injection prefers each store’s own cap). Could we express that with a shared private fan-out helper `_searchStores ()` + a plain `preferStoreLimit` boolean on that fan-out instead? That keeps `search()`’s public contract clean (no option for preference) and drops the internal type and the casts. The resulting behavior stays the same.

Also, could you add parity for Python so the docs reflect correct behaviour for both SDKS?

site/src/content/docs/user-guide/concepts/memory/overview.mdx
```diff
@@ -117,7 +117,7 @@ Each store carries its own identity and behavior:
 |-------|---------|
 | `name` | Unique identifier, used to target the store from tools and the programmatic API. |
 | `description` | Human-readable summary, surfaced in the memory tool descriptions so the model knows what each store holds. |
-| <Syntax py="max_search_results" ts="maxSearchResults" /> | Default result cap per search when a caller does not pass one. The manager falls back to `3` if neither is set. |
+| <Syntax py="max_search_results" ts="maxSearchResults" /> | Default result cap per store search when a caller does not pass one. The manager falls back to `3` if neither is set. During injection, a store-level cap remains the authoritative per-store retrieval limit before injected entries are sliced. |
```
zastrowm, line 120: @opieter-aws do we think this parameter is misnamed?

Is this really `defaultSearchResults` and really we wanted `maxSearchResults` to be the "absolute max that we have"?

Should we pivot the change to be something along those lines (or even with different names?

opieter-aws:
@matthewchen94 after some discussions I am wondering what your use case is for this behaviour, since the existing setup _is_ by design:

1. `maxSearchResults` is the setting on the store on the maximum results to return upon a direct store search call. This is needed for the agentic use case, where the agent searches the store using a tool call. It's called `max` since we can't guarantee number of results returns based on knowledge base matches (e.g. `maxSearchResults: 10` can still return just 1 entry)
2. Injection is a separate mechanism, where you decide the number of memories you want to inject into the context window. This is honoring its own settings, not the number of search results to return on the store. This is by design, since injection orchestrates amongst all registered stores. We wouldn't want a injection number of 10 to result in only 1 injected entry because the store isn't allowing more entries to be returned.

Can you please elaborate on the use case?

matthewchen94:
Thanks for the explanation — that distinction makes sense. My use case is less about using `maxSearchResults` as the final injection count, and more about preserving a store-level retrieval budget/safety cap when injection fans out across stores.

Concrete case:

- Store A is backed by a remote KB/vector service with a configured retrieval cap, e.g. `maxSearchResults: 1`, because the owner wants bounded latency/cost/noise for that store.
- Injection is configured with `maxEntries: 10` as the total context budget across all stores.
- Current behavior passes the injection budget down to each store, so Store A can be asked for 10 results even though its own store config says 1.

That is the surprising part I was trying to fix: the injection path turns the global context budget into a per-store retrieval override. For local stores this is mostly harmless. For remote/paid/latency-sensitive stores, it bypasses the store owner’s cap.

I agree the final injection count should stay controlled by injection config, because injection is orchestrating across stores. The specific semantic I’m after is:

1. per-store retrieval should not exceed the store’s own configured cap when one exists;
2. injection `maxEntries` should still cap the combined injected result set;
3. when a store has no cap, injection `maxEntries` is a reasonable fallback retrieval size.

If that is not the behavior you want by default, I can rework this into an explicit opt-in instead of changing the default. For example:

- add a named store/injection option such as `respectStoreMaxSearchResultsForInjection`, or
- keep runtime behavior unchanged and only document that injection intentionally overrides store-level `maxSearchResults`.

My preference is the explicit opt-in if preserving current behavior is important. That keeps the existing design intact while giving remote store adapters a safe way to enforce their own retrieval budget during automatic injection.

opieter-aws:
I don't think we want to re-use the `maxSearchResults` param as a cap: it just notes the _default_ search results you can expect _at maximum_, given that stores can return less results. This is not a store level cap.

If a documentation update would clarify this, I'm all for it!

PR #3088, opened 2026-07-03
feat(models): add MiniMax M3 provider

octo-patch:
## Summary
Add MiniMax as a new LLM provider. MiniMax exposes an OpenAI-compatible chat completions API, so `MinimaxModel` extends `OpenAIModel` (the same pattern used by `LiteLLMModel` and `SageMakerAIModel`) and defaults the base URL and API key for the MiniMax endpoint.

## Changes
- Add `MinimaxModel` in `strands-py/src/strands/models/minimax.py`, extending `OpenAIModel`:
  - Defaults `base_url` to the MiniMax endpoint and reads the API key from `MINIMAX_API_KEY` when the caller relies on auto-configuration.
  - Drops an empty `tools` list in `format_request` (MiniMax rejects it).
  - Overrides `structured_output` to use a regular completion with `response_format=json_object` and strips `<think>` tags / markdown code fences before parsing (the beta parse API is confused by reasoning tags).
- Register `MinimaxModel` in `strands-py/src/strands/models/__init__.py` via the existing lazy-load `__getattr__`.
- Add a `minimax` optional-dependency extra (and include it in `all`) in `strands-py/pyproject.toml`.
- Add unit tests in `strands-py/tests/strands/models/test_minimax.py`.

## Testing
- `pytest tests/strands/models/test_minimax.py` — 33 passed.
- `ruff check` and `ruff format --check` clean on the changed files.

> Note: the diff exceeds 300 added lines, almost entirely due to the unit test file, which mirrors the size and structure of the sibling provider test suites in this package.

### DIFF
--- a/strands-py/pyproject.toml
+++ b/strands-py/pyproject.toml
@@ -49,6 +49,7 @@ anthropic = ["anthropic>=0.21.0,<1.0.0"]
 gemini = ["google-genai>=1.32.0,<3.0.0"]
 litellm = ["litellm>=1.75.9,<=1.90.2", "openai>=1.68.0,<3.0.0"]
 llamaapi = ["llama-api-client>=0.1.0,<1.0.0"]
+minimax = ["openai>=1.68.0,<3.0.0"]
 mistral = ["mistralai>=2.0.0,<3.0.0"]
 ollama = ["ollama>=0.4.8,<1.0.0"]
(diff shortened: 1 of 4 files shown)


chaynabors:
Can you open an issue instead? It's hard for us to accept what would increase our support burden without community signal. If you'd like, you can also publish as an extension in the meantime:

https://strandsagents.com/docs/contribute/contributing/extensions/

PR #3099, opened 2026-07-06
feat: add unified storage interface

lizradway:
## Summary

Adds the unified `Storage<ListQuery>` interface — the single persistence primitive for the SDK — and wires it into existing subsystems (context offloader, session manager).

## API

```typescript
import type { Storage } from '@strands-agents/sdk/storage'

interface Storage<ListQuery = string> {
  write(key: string, data: Uint8Array): Promise<void>
  read(key: string): Promise<Uint8Array | null>
  delete(key: string): Promise<void>
  list(query: ListQuery): Promise<string[]>
}
```

Four operations over opaque `Uint8Array` values. Keys are `/`-separated path-like strings. The generic `ListQuery` defaults to string (prefix match) but can be widened by backends like DynamoDB.

**Shipped implementations:**

| Class | Backend | Import path |
|-------|---------|-------------|
| `InMemoryStorage` | `Map` | `@strands-agents/sdk/storage` |
| `LocalFileStorage` | Filesystem (atomic writes) | `@strands-agents/sdk/storage` |
| `S3Storage` | AWS S3 (lazy SDK import) | `@strands-agents/sdk/storage` |

## Developer Experience

```typescript
import { LocalFileStorage } from '@strands-agents/sdk/storage'
import { ContextOffloader } from '@strands-agents/sdk/vended-plugins/context-offloader'
import { SessionManager } from '@strands-agents/sdk/session'

// One storage instance backs everything
const storage = new LocalFileStorage('./.agent-data')

const agent = new Agent({
  model,
  plugins: [new ContextOffloader({ storage })],
  sessionManager: new SessionManager({ storage }),
})

// Or scope backends with .namespace()
const offloaderStorage = storage.namespace('offloader')
const sessionStorage = storage.namespace('sessions')
```

## Changes

**Storage primitive** (`src/storage/storage.ts`):
- `Storage<ListQuery = string>` interface with `write`, `read`, `delete`, `list`
- `namespace()` helper for key-prefix scoping (composable)
- Key normalization: collapses `//`, strips leading/trailing `/`, rejects `..` segments
- `StorageError` in `src/errors.ts`; `./storage` subpath export

**Implementations:**
- `InMemoryStorage`: Map-backed, for testing and serverless
- `LocalFileStorage`: atomic writes (tmp + rename), recursive directory walking, sandbox-aware via `forSandbox()`
- `S3Storage`: lazy `@aws-sdk/client-s3` import, paginated listing, prefix namespacing

**Context offloader adaptation:**
- Accepts unified `Storage` or legacy `OffloaderStorage` (duck-type detection)
- Cycle-based eviction: tracks store-time via `agent.metrics.cycleCount`, deletes entries older than `evictAfterCycles` (default 20, `null` disables)
- `evictAfterCycles` also flows to legacy storage when the user hasn't explicitly set `evictAfterTurns`
- Content-type framed into stored bytes (2-byte length prefix + UTF-8 content-type + content)

**Session manager adaptation:**
- `SnapshotStorageAdapter` bridges unified `Storage` → `SnapshotStorage`
- `SessionManagerConfig.storage` accepts `Storage` directly (or legacy `{ snapshot: SnapshotStorage }`)

## Future Work

Phase 1 of the [unified storage design](https://github.com/strands-agents/harness-sdk/pull/3080). Planned next:
- Memory store integration (`LocalMemoryStore` accepts optional `Storage`)
- Context manager integration (v2 context strategy)
- Transcripts (append-only audit log backed by `Storage`)
- Remove deprecated classes (next major)

## Testing

- Unit tests for all storage implementations (InMemoryStorage, LocalFileStorage, S3Storage)
- Unit tests for SnapshotStorageAdapter (CRUD round-trips, pagination, error wrapping)
- Eviction tests for unified Storage path (cycle-based boundary conditions, null-disabled)
- All existing offloader and session tests pass unchanged
- Coverage: key normalization, path-traversal rejection, atomic writes, sandbox routing, duck-type detection, content-type framing round-trip

## Checklist


strands-ts/src/session/session-manager.ts
```diff
@@ -104,16 +113,42 @@ export class SessionManager implements Plugin, MultiAgentPlugin {
...
+   *
+   * @param storage - The agent-level storage instance
+   */
+  initStorage(storage: Storage): void {
+    if (this._hasExplicitStorage) return
+    this._storage = { snapshot: new SnapshotStorageAdapter(storage) }
+  }
+
+  private get _snapshotStorage(): SnapshotStorage {
+    return this._storage!.snapshot
```
mehtarac: Suggestion: Should we wrap this in an SessionError instead of using the `!` operator?

Ex:
```
if (!this._storage) {
      throw new SessionError('SessionManager storage not initialized')
    }
    return this._storage.snapshot
```

SessionManager exposes public methods like `deleteSession()`, `listSnapshotIds()`, and `loadSnapshot()` that go through `_snapshotStorage`. If a user constructs a `SessionManager()` with no config and calls one of those methods before attaching it to an agent (before initAgent runs), the `!` assertion gives them `TypeError: Cannot read properties of undefined instead of the helpful error message.`

strands-ts/src/storage/local-file-storage.ts
```diff
@@ -0,0 +1,224 @@
...
+      } catch (error: unknown) {
+        throw new StorageError(`Failed to write '${normalized}' to sandbox storage`, { cause: error })
+      }
+      return
+    }
+    try {
+      const { mkdir, writeFile, rename } = await import('node:fs/promises')
+      const { dirname } = await import('node:path')
+      await mkdir(dirname(path), { recursive: true })
+      const tmpPath = `${path}.__strands_tmp`
```
mehtarac: Are concurrent writes possible? If so would the two writes target the same file `.tmp` file? 

If concurrent writes possible, I think it'll be good to use a unique temp filename such as `${path}.__strands_tmp_${randomUUID()}` to support concurrent writers

mehtarac (dismissed):

gautamsirdeshmukh (dismissed):

strands-ts/src/session/session-manager.ts
```diff
@@ -53,10 +56,14 @@ export type SaveLatestStrategy = 'message' | 'invocation' | 'trigger'
...
-  /** Pluggable storage backends for snapshot persistence. Defaults to FileStorage in Node.js; required in browser environments. */
-  storage: {
-    snapshot: SnapshotStorage
-  }
+  /**
+   * Storage backend for snapshot persistence.
+   *
+   * Accepts either:
+   * - A unified {@link Storage} instance (recommended) — wrapped internally with {@link SnapshotStorageAdapter}
+   * - A legacy `{ snapshot: SnapshotStorage }` object for backwards compatibility
```
Unshure: nit: You dont need to call out backwards compatibility here. Its just the legacy object

strands-ts/src/session/session-manager.ts
```diff
@@ -53,10 +56,14 @@ export type SaveLatestStrategy = 'message' | 'invocation' | 'trigger'
...
+   */
+  storage?: Storage | { snapshot: SnapshotStorage }
```
Unshure: What was the original intention of this parameter? Having it defined as `{snapshot: SnapshotStorage }` makes me think that other parameters were expected to be added into this in the future. I dont have the context, so just want to make sure we are tying this appropriately.

strands-ts/src/session/session-manager.ts
```diff
@@ -91,7 +98,7 @@ export interface SessionManagerConfig {
  */
 export class SessionManager implements Plugin, MultiAgentPlugin {
   private readonly _sessionId: string
-  private readonly _storage: { snapshot: SnapshotStorage }
+  private _storage: { snapshot: SnapshotStorage } | undefined
```
Unshure: Why not readonly?

gautamsirdeshmukh (dismissed):

mehtarac (dismissed):

strands-ts/src/session/session-manager.ts
```diff
@@ -104,16 +111,34 @@ export class SessionManager implements Plugin, MultiAgentPlugin {
...
+    if ('snapshot' in storage) return storage.snapshot
+    return new SnapshotStorageAdapter(storage)
+  }
+
   /** Initializes the plugin by registering lifecycle hook callbacks. */
   public initAgent(agent: LocalAgent): void {
+    if (!this._storage) {
+      throw new Error(
+        'SessionManager requires a storage backend. ' +
+          'Pass storage in the SessionManager config or set storage on the Agent config.'
```
Unshure: Is storage on Agent config?

strands-ts/src/storage/in-memory-storage.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  /**
+   * Lists the keys whose names begin with `prefix`, sorted lexicographically.
+   *
+   * @param prefix - Key prefix to match. An empty string matches all keys.
+   * @returns The matching keys, sorted ascending
+   * @throws {@link StorageError} if the prefix contains `..` segments
+   */
+  async list(prefix: string): Promise<string[]> {
```
Unshure, line 71: I remember we chatted about changing the api names here to use CRDL, where did that end up landing?

strands-ts/src/storage/storage.ts
```diff
@@ -0,0 +1,58 @@
+/**
+ * A backend for storing and retrieving raw bytes under string keys.
+ *
+ * The interface is deliberately minimal — four operations over opaque `Uint8Array`
+ * values. Implementations must treat keys as opaque path-like strings (segments
+ * separated by `/`) and must round-trip the bytes they are given unchanged.
+ *
+ * The `ListQuery` type parameter controls what `list` accepts. It defaults to
+ * `string` (a key prefix), which every backend supports. Implementations may
```
Unshure, line 54: Does every backend support a prefix for the ListQuery? Do we need to provide a default at all here?

strands-ts/src/session/snapshot-storage-adapter.ts
```diff
@@ -0,0 +1,187 @@
...
+ * @deprecated Remove in v2 when SnapshotStorage is dropped and SessionManager calls Storage directly.
+ * @internal
+ * @param storage - The unified Storage backend to delegate to
+ * @param basePrefix - Optional key prefix. Defaults to `'session'`.
+ */
+export class SnapshotStorageAdapter implements SnapshotStorage {
+  private readonly _storage: Storage
+  private readonly _basePrefix: string
+
+  constructor(storage: Storage, basePrefix: string = 'session') {
```
opieter-aws: The adapter keys snapshots under `session/<sessionId>/scopes/...`, but legacy FileStorage uses `<baseDir>/<sessionId>/scopes/...` and legacy S3Storage uses `[<prefix>/]<sessionId>/scopes/...`

If `SessionManager` always calls n`ew SnapshotStorageAdapter(storage)` with no prefix, a user following this PR’s `@deprecated` guidance `({ snapshot: new FileStorage('./data') }` -> `new LocalFileStorage('./data'))` reads/writes a different key space. We should guard against this silent data loss. Consider removing the base prefix?

strands-ts/src/storage/local-file-storage.ts
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  /**
+   * Lists the keys whose names begin with `prefix`, sorted lexicographically.
+   *
+   * @param prefix - Key prefix to match. An empty string matches all keys.
+   * @returns The matching keys, sorted ascending
+   * @throws {@link StorageError} if the prefix is invalid or the listing fails
+   */
+  async list(prefix: string): Promise<string[]> {
```
opieter-aws, line 160: We should ensure that we return `[]` here for `ENOENT` _and_ `ENOTDIR`

strands-ts/src/storage/index.ts
```diff
@@ -0,0 +1,20 @@
...
+ * @example
+ * ```typescript
+ * import { LocalFileStorage, InMemoryStorage } from '@strands-agents/sdk/storage'
+ * ```
+ *
+ * @packageDocumentation
+ */
+
+export type { Storage } from './storage.js'
+export { InMemoryStorage } from './in-memory-storage.js'
```
opieter-aws, line 17: Since naming clashes with the legacy offloader ones, it can be confusing for users. Should we rename the identical storage exports from `vended-plugins/context-offloader/index.ts` to `Legacy....`?

strands-ts/src/vended-plugins/context-offloader/plugin.ts
```diff
@@ -120,63 +183,111 @@ export interface ContextOffloaderConfig {
...
+        this._evict(cycleCount)
       }
     })
   }
 
+  private _evict(currentCycle: number): void {
+    const threshold = currentCycle - this._evictAfterCycles!
+    const toEvict: string[] = []
+    for (const [key, lastAccess] of this._keyStoredAt) {
+      if (lastAccess < threshold) {
```
opieter-aws: Should we revise naming for `lastAccessed`, since we evaluate against `storedAt` here instead?

opieter-aws: Should we see if we can get `lastAccessed` back in? It is a more sensible mechanism to me than `storedAt`

strands-ts/src/vended-plugins/context-offloader/plugin.ts
```diff
@@ -103,14 +149,31 @@ function decodeStoredContent(content: Uint8Array, contentType: string, reference
...
+   *
+   * Accepts either:
+   * - A unified `Storage` (from `@strands-agents/sdk/storage`) — each offloaded content block
+   *   occupies exactly one key (content-type is framed into the stored bytes).
+   * - A legacy offloader `Storage` (deprecated, from this module) — manages its own turn-based
+   *   eviction internally via its `evictAfterTurns` constructor parameter.
+   *
+   * Required — must be provided in the plugin constructor.
+   */
+  storage?: Storage | OffloaderStorage
```
opieter-aws: Why not `storage: Storage | OffloaderStorage` if required?

strands-ts/src/session/session-manager.ts
opieter-aws: Why not required if we throw if not present?

lizradway: @JackYPCOnline any context on this?

strands-ts/src/session/snapshot-storage-adapter.ts
```diff
@@ -0,0 +1,187 @@
...
+const SNAPSHOT_REGEX = /snapshot_([\w-]+)\.json$/
+
+/**
+ * Adapts a unified {@link Storage} instance into the {@link SnapshotStorage} interface
+ * expected by the session manager.
+ *
+ * Keys follow the same layout as the filesystem-based storage:
+ * `session/<sessionId>/scopes/<scope>/<scopeId>/snapshots/...`
+ *
+ * @deprecated Remove in v2 when SnapshotStorage is dropped and SessionManager calls Storage directly.
```
opieter-aws: Do we want to introduce a new class with `@deprecated`? We don't discourage its usage (we encourage it!), it's the happy path for migration.

strands-ts/src/session/storage.ts
```diff
@@ -16,12 +16,8 @@ export type SnapshotLocation = {
...
  * Allows users to configure snapshot and transcript storage independently.
  *
- * @example
- * ```typescript
- * const storage: SessionStorage = {
- *   snapshot: new S3Storage({ bucket: 'my-bucket' })
- * }
- * ```
+ * @deprecated Remove in v2 when SessionManager accepts only unified `Storage`.
+ * @internal Prefer passing a unified `Storage` directly to `SessionManagerConfig.storage`.
```
opieter-aws: Marked as `@internal` contradicts the export from `index.ts`

strands-ts/src/session/storage.ts
```diff
@@ -32,6 +28,10 @@ export type SessionStorage = {
  * Interface for snapshot persistence.
  * Implementations provide storage backends (S3, filesystem, etc.).
  *
+ * @deprecated Remove in v2 when SessionManager calls unified `Storage` directly.
+ * @internal This is an internal contract used by the session manager. Users should pass
```
opieter-aws: ditto

strands-ts/src/session/session-manager.ts
lizradway: im inferring that it could extended eventually to  `{ snapshot: SnapshotStorage, transcript: TranscriptStorage }`, but per our prior meeting transcripts are not happening anytime soon and as proposed currently would not live on the Session Manager interface.

strands-ts/src/storage/local-file-storage.ts
```diff
@@ -0,0 +1,225 @@
...
+    }
+    try {
+      const { mkdir, writeFile, rename } = await import('node:fs/promises')
+      const { dirname } = await import('node:path')
+      await mkdir(dirname(path), { recursive: true })
+      const { randomUUID } = await import('node:crypto')
+      const tmpPath = `${path}.__strands_tmp_${randomUUID()}`
+      await writeFile(tmpPath, data)
+      await rename(tmpPath, path)
+    } catch (error: unknown) {
```
opieter-aws, line 89: in the catch clauses, should we attempt to delete the `tmpPath` file before rethrowing?

strands-ts/src/session/session-manager.ts
```diff
@@ -104,16 +111,34 @@ export class SessionManager implements Plugin, MultiAgentPlugin {
...
 
+  private get _snapshotStorage(): SnapshotStorage {
+    if (!this._storage) {
+      throw new SessionError('SessionManager storage not initialized')
+    }
+    return this._storage.snapshot
+  }
+
+  private _resolveSnapshotStorage(storage: Storage | { snapshot: SnapshotStorage }): SnapshotStorage {
+    if ('snapshot' in storage) return storage.snapshot
```
opieter-aws, line 127: Would it be safer to detect the unified shape rather than `snapshot` only? Otherwise any storage that happens to have a `snapshot` member could get misclassified

strands-ts/src/session/session-manager.ts
lizradway: reverting. also reverting to make storage required.

strands-ts/src/session/session-manager.ts
lizradway: no i removed it per API bar raising since we decided not to put high level storage on agent config immediately.

happy to add back as that was included in original design. thoughts? @Unshure @opieter-aws

strands-ts/src/storage/in-memory-storage.ts
lizradway: ah didn't realize we wanted the actual API to reflect CRDL explicitly. can update accordingly.

lizradway: do we want "fail if exists" behavior specified as is done in `create` or "overwrite" as is done in `put`?

strands-ts/src/session/session-manager.ts
opieter-aws: I think it's fine to _not_ include for now, we can always decide to support later

strands-ts/src/session/session-manager.ts
JackYPCOnline: ^ Correct, it is still make sense we leave it forward-compatible

strands-ts/src/storage/storage.ts
lizradway: every backend we support 1P uses a string prefix. only custom backends (ie. DDB query) are going to widen the generic. 


the default just avoids forcing every consumer to write `Storage<string>`

strands-ts/src/session/snapshot-storage-adapter.ts
lizradway: i think this prefix is necessary for our upcoming plan for directory differentiation within `Storage`. 

i can add a note explaining this so users are aware.

strands-ts/src/storage/index.ts
lizradway: wouldn't that be a breaking change?

 users shouldn’t need both imports so if someone does import from both paths they can alias locally.

strands-ts/src/session/snapshot-storage-adapter.ts
opieter-aws: I understand the prefix is worth keeping.

Can we implement a mechanism rather than add a note? If you just swap your storage solution, it will still silently relocate snapshots and restore returns null and you resume with an empty conversation, no error. 

Two potential ways to keep the prefix and make the transition safer:
1. check on a miss at `session` prefix, fall back to the un-prefixed legacy path.  
2. error: detect legacy keys under the same base and throw a `SessionError`

Could we surface the existing `basePrefix` as `SessionManagerConfig.storagePrefix`? And longer term, a `storage.namespace('session')` wrapper could let sessions/memory/transcripts all use prefixing/namespacing through one mechanism instead of each hardcoding a string. Maybe out of scope for now, but just some thoughts!

strands-ts/src/storage/in-memory-storage.ts
Unshure: I probably lean toward error if exists, so more like `POST`. But either way is fine so long as its documented.

strands-ts/src/session/snapshot-storage-adapter.ts
lizradway: > Could we surface the existing basePrefix as SessionManagerConfig.storagePrefix 
i like this ! can do!

strands-ts/src/vended-plugins/context-offloader/plugin.ts
lizradway: > Should we see if we can get lastAccessed back in?

I agree, that it is a more intelligent mechanism, but I want to maintain simplicity as much as possible rn + not overcomplicate the logic, especially as we move to the Context Manager. 


Can rename to `storedCycle` if you prefer?

strands-ts/src/vended-plugins/context-offloader/plugin.ts
lizradway: fixed

strands-ts/src/session/session-manager.ts
lizradway: i think detecting all four methods on storage would be clunky and overbearing. 

im guessing that any custom storage mechanism that resolves `snapshot` is likely the legacy storage type.

strands-ts/src/session/snapshot-storage-adapter.ts
```diff
@@ -0,0 +1,188 @@
...
+const SNAPSHOT_REGEX = /snapshot_([\w-]+)\.json$/
+
+/**
+ * Adapts a unified {@link Storage} instance into the {@link SnapshotStorage} interface
+ * expected by the session manager.
+ *
+ * Keys are written directly into the provided storage:
+ * `<sessionId>/scopes/<scope>/<scopeId>/snapshots/...`
+ *
+ * Callers control namespacing by passing a {@link NamespacedStorage} — e.g.
```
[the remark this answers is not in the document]
opieter-aws: +1, `NamespacedStorage` seems like a dead reference

strands-ts/src/session/snapshot-storage-adapter.ts
opieter-aws: I still think we don't want to mark this as `@deprecated`, since that's a signal for users to move away from the functionality. That's not what we want to signal here IMO.

strands-ts/src/storage/storage.ts
```diff
@@ -0,0 +1,117 @@
...
+ * Returns a {@link Storage} view with all keys prefixed by `prefix`.
+ *
+ * Composable — calling `namespace()` on the result nests prefixes.
+ *
+ * @internal
+ * @param storage - The underlying storage to delegate to
+ * @param prefix - Prefix to prepend to all keys
+ * @returns A namespaced Storage view
+ */
+export function namespace(storage: Storage, prefix: string): Storage & { namespace(prefix: string): Storage } {
```
opieter-aws: Right now namespace exists in two forms, and they don’t line up:

1. The free function here
2. public `.namespace(prefix)` method on all three classes 

The inconsistency is that the method is public on the concrete classes, but it’s not on the `Storage` interface. So a `SessionManagerConfig.storage` can’t call `.namespace()`, even though the underlying object supports it. And the methods have no callers; everything goes through this function.

We should either drop the methods and only keep this function, or make it a real public API and promote it onto the `Storage` interface. That way users can namespace storage themselves too.

lizradway: going to remove the public namespace and keep only the internal method for now to allow us to punt the namespacing public api until we have more gauge on use of the storage primitive.

strands-ts/src/vended-plugins/context-offloader/storage.ts
```diff
@@ -123,6 +129,13 @@ export class InMemoryStorage implements Storage {
           'Use a separate InMemoryStorage instance per agent.'
       )
     }
+    if (
+      evictAfterCycles !== undefined &&
+      evictAfterCycles !== null &&
```
opieter-aws: Should this check for `_evictAfterTurns`? Currently setting the legacy field `null` will enable eviction

strands-ts/src/session/session-manager.ts
```diff
@@ -106,12 +113,21 @@ export class SessionManager implements Plugin, MultiAgentPlugin {
...
     this._snapshotTrigger = config.snapshotTrigger
   }
 
+  private get _snapshotStorage(): SnapshotStorage {
+    return this._storage.snapshot
+  }
+
+  private _resolveSnapshotStorage(storage: Storage | { snapshot: SnapshotStorage }): SnapshotStorage {
+    if ('snapshot' in storage) return storage.snapshot
+    return new SnapshotStorageAdapter(namespace(storage, 'session'))
```
opieter-aws: With the hardcoded `session` prefix, there's no way to for customers to point to their legacy file location. Shall we give them a mechanism so hand-moving files is not the only way?

lizradway: i really do not like exporting `namespace` as a function (or on the api for now), which is why i moved it to internal only for this. i also dont like it as required since it likely wont be for community impl (ie. ddb, redis, etc).

but i guess we have no choice, i just dont like the idea of maintaining the namespace function, so i think im gonna have it be optional on the Storage interface and warn if it is called and not implemented

strands-ts/src/storage/storage.ts
```diff
@@ -0,0 +1,128 @@
...
+  /**
+   * Returns a view of this storage with all keys prefixed by `prefix`.
+   * The original storage is not mutated.
+   *
+   * Optional — shipped backends implement this, custom backends may omit it.
+   *
+   * @param prefix - Prefix to prepend to all keys
+   * @returns A Storage view scoped to the given prefix
+   */
+  namespace?(prefix: string): Storage
```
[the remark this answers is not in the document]
lizradway: leaving this as is. the user would only need to optional type namespace on a custom storage implementation that they are generic typing as `Storage` rather than the specific type.

strands-ts/src/session/session-manager.ts
opieter-aws: The optional interface change addresses the namespace API question, but I don’t think it resolves the concern here: you can only add prefixes with `.namespace()` but you still can’t opt out of the hardcoded `session/` that `SessionManager` prepends. So an in-place legacy -> unified swap still relocates their snapshots silently. Can we surface a `storagePrefix` on `SessionManagerConfig` with defaults?

opieter-aws (approved):

opieter-aws:
Thanks for the latest change, this enables migration! I would suggest to clearly document this in the docstrings of `@deprecated` classes, or reference there to a migration guide.

Reviewer bot called out a few gaps, but I'm fine to do that as follow-up

lizradway:
will update tests and doc strings in python/docs port

lizradway:
@strandly-the-agent port this to a python implementation

PR #3100, opened 2026-07-06
fix(context-offloader): keep optional @aws-sdk/client-s3 out of core Agent import graph

### LINKED ISSUE #3016: [BUG] Bundling Agent pulls in optional @aws-sdk/client-s3 peer dep, breaking esbuild builds
### Checks

### SDK Language

TypeScript

### Strands Version

1.7.0

### Language Runtime Version

Node 24.16.0

### Operating System

macos 26.2

### Installation Method

npm

### Steps to Reproduce

Minimal reproducible example — a project that imports the `Agent` but does **not** use context offloading and does **not** install the optional `@aws-sdk/client-s3` peer:

```ts
// entry.ts
import { Agent } from '@strands-agents/sdk';
// (no ContextOffloader plugin, no S3Storage — just the core Agent)
```

```bash
npm i @strands-agents/sdk            # @aws-sdk/client-s3 NOT installed (it's an optional peer)
npx esbuild entry.ts --bundle --platform=node --format=esm --outfile=out.js
```

See the build fail to resolve `@aws-sdk/client-s3`.

### Expected Behavior

Bundling an app that only uses the core `Agent` should succeed without installing `@aws-sdk/client-s3`.

`@aws-sdk/client-s3` is declared optional in the SDK's own `package.json`:

```json
"peerDependenciesMeta": {
  "@aws-sdk/client-s3": { "optional": true }
}
```

It's only needed for the opt-in context-offloader `S3Storage` backend, and the feature is gated at runtime (`agent.js`: `const hasOffloader = (config?.plugins ?? []).some(p => p.name === 'strands:context-offloader')`). An app that never registers the `ContextOffloader` plugin should not be forced to install or bundle the S3 client.

### Actual Behavior

`agent/agent.js` **statically and unconditionally** imports the context-offloader storage module:

```js
// node_modules/@strands-agents/sdk/dist/src/agent/agent.js:22-23 (1.7.0)
import { ContextOffloader } from '../vended-plugins/context-offloader/plugin.js';
import { InMemoryStorage } from '../vended-plugins/context-offloader/storage.js';
```

`storage.js` then does a string-literal dynamic import of the S3 client:

```js
// dist/src/vended-plugins/context-offloader/storage.js:287 (1.7.0)
const { S3Client } = await import('@aws-sdk/client-s3');
```

Because `agent.js` is reachable from the package entry (`index.js`) the moment you import `Agent`/`BedrockModel`, `storage.js` lands in the bundler's static import graph. esbuild eagerly resolves string-literal dynamic imports while building the graph, so it tries to resolve the *optional, uninstalled* `@aws-sdk/client-s3` and hard-fails the build:

```
✘ [ERROR] Could not resolve "@aws-sdk/client-s3"
    node_modules/@strands-agents/sdk/dist/src/vended-plugins/context-offloader/storage.js:216:42:
      216 │         const { S3Client } = await import('@aws-sdk/client-s3');
          ╵                                           ~~~~~~~~~~~~~~~~~~~~
  You can mark the path "@aws-sdk/client-s3" as external to exclude it from the bundle, which will remove this error and leave the unresolved path in the bundle. You can also add ".catch()" here to handle this failure at run-time instead of bundle-time.
/builds/runner/61/showpad-code-blueprints-agentcore-blueprint/node_modules/esbuild/lib/main.js:1748
  let error = new Error(text);
              ^
Error: Build failed with 1 error:
node_modules/@strands-agents/sdk/dist/src/vended-plugins/context-offloader/storage.js:216:42: ERROR: Could not resolve "@aws-sdk/client-s3"
    at failureErrorWithLog (/builds/runner/61/showpad-code-blueprints-agentcore-blueprint/node_modules/esbuild/lib/main.js:1748:15)
    at /builds/runner/61/showpad-code-blueprints-agentcore-blueprint/node_modules/esbuild/lib/main.js:1207:25
    at /builds/runner/61/showpad-code-blueprints-agentcore-blueprint/node_modules/esbuild/lib/main.js:1159:52
    at buildResponseToResult (/builds/runner/61/showpad-code-blueprints-agentcore-blueprint/node_modules/esbuild/lib/main.js:1205:7)
    at /builds/runner/61/showpad-code-blueprints-agentcore-blueprint/node_modules/esbuild/lib/main.js:1232:16
    at responseCallbacks.<computed> (/builds/runner/61/showpad-code-blueprints-agentcore-blueprint/node_modules/esbuild/lib/main.js:884:9)
    at handleIncomingPacket (/builds/runner/61/showpad-code-blueprints-agentcore-blueprint/node_modules/esbuild/lib/main.js:939:12)
    at Socket.readFromStdout (/builds/runner/61/showpad-code-blueprints-agentcore-blueprint/node_modules/esbuild/lib/main.js:862:7)
    at Socket.emit (node:events:509:28)
    at addChunk (node:internal/streams/readable:563:12) {
  errors: [Getter/Setter],
  warnings: [Getter/Setter]
}
```

The runtime `hasOffloader` guard never gets a chance to help — the failure is at bundle time, not runtime. This regressed between 1.1.0 and 1.5.0: in 1.1.0 the entry barrel did not reach the context-offloader, so `storage.js` was never in the graph and bundling succeeded. It remains broken through the latest release (verified on 1.7.0).

### Additional Context

- Reproduces with esbuild `bundle: true`, `platform: 'node'` — the standard setup for shipping a single-file Node artifact (in our case an AWS Bedrock AgentCore direct-code deployment).
- Tree-shaking does not help: a module must be resolved/parsed before it can be shaken, and resolution failure happens during load. The dynamic `import()` is also a code-split boundary esbuild preserves rather than eliminates.
- Current workaround: install `@aws-sdk/client-s3` even though it's unused, which bundles several MB of dead S3 client. Alternatively mark it `external`, but that's not always controllable when bundling is owned by a wrapper/CDK construct.

### Possible Solution

Keep the optional S3 peer out of the core import graph so it's only pulled in when context offloading is actually used. Options:

1. Don't statically import `storage.js` from `agent/agent.js`. Have core `Agent` lazily/dynamically load the default storage only when a `ContextOffloader` plugin is registered, so apps that don't use offloading never reach the S3 import.
2. Isolate the `S3Storage` backend (the only thing needing `@aws-sdk/client-s3`) into a module that is *not* transitively imported by core `Agent` — e.g. expose it only via the `@strands-agents/sdk/vended-plugins/context-offloader` subpath so importing `Agent` doesn't drag it in.
3. At minimum, document that bundlers must mark `@aws-sdk/client-s3` as `external` when context offloading is unused.

Option 1 or 2 is preferable — they make the static import graph match the declared optionality, so the optional peer stays genuinely optional for bundlers.

### Related Issues

_No response_

maxmilian:
## Description

The core `Agent` statically imports `InMemoryStorage` from the context-offloader `storage.ts`, which also housed `S3Storage`. `S3Storage` lazily loads the optional `@aws-sdk/client-s3` peer via `await import(...)` — but a bundler eagerly resolves string-literal dynamic imports while building the static graph. So importing only `Agent` dragged the optional, uninstalled peer into esbuild's graph and hard-failed the build (`Could not resolve "@aws-sdk/client-s3"`) for apps that never register the `ContextOffloader` plugin. The runtime `hasOffloader` guard never gets a chance to help — the failure is at bundle time.

**Fix (option 2 from the issue — isolate the S3 backend):** move `S3Storage`, the only consumer of `@aws-sdk/client-s3`, out of `storage.ts` into its own `s3-storage.ts`. It is re-exported unchanged from the `vended-plugins/context-offloader` subpath barrel, so **there is no public API change** — `import { S3Storage } from '@strands-agents/sdk/vended-plugins/context-offloader'` still works. The core `Agent` import graph (`agent.ts` → `storage.ts`) no longer reaches the S3 client, so the optional peer is genuinely optional for bundlers again.

## Related Issues

Closes #3016

## Documentation PR

No docs change needed — the public API of the `context-offloader` subpath is unchanged.

## Type of Change

Bug fix

## Testing

Added a node-only regression test (`src/__tests__/bundle-optional-peers.test.node.ts`, per @zastrowm's suggestion) that bundles a `{ Agent }` entry with esbuild and asserts `@aws-sdk/client-s3` is absent from the resolved module graph.

Real-behavior proof (bundling the package entry that imports only `Agent`):

```
# before this change
aws-sdk/client-s3 inputs in the core Agent graph: 10
# after this change
aws-sdk/client-s3 inputs in the core Agent graph: 0
```

Reverting only the source fix makes the new test fail with `expected [ …(10) ] to deeply equal []`, and it passes with the fix — a true red→green regression test.

- Existing context-offloader `S3Storage` tests were moved verbatim to `__tests__/s3-storage.test.ts` (import path updated); all offloader tests pass.
- `npm run build -w strands-ts` (tsc) — green (also runs in the pre-commit hook).
- `tsc --project src/tsconfig.json` — green. (Note: `npm run type-check` also runs `test/integ/tsconfig.json`, which has ~466 pre-existing errors on `main` unrelated to this change.)
- `npm run format:check` and `eslint` on the changed files — clean.
- [ ] I ran `hatch run prepare` — n/a, this is a TypeScript-only change; ran the TS gates above instead.

## Checklist

----


lizradway:
Long term solution for this lives in https://github.com/strands-agents/harness-sdk/pull/3099/changes


This looks good for a short term fix to unblock. Thanks for catching this!

gautamsirdeshmukh (changes requested):
Just one callout (agreed with the bot) - could you please update the `sourceLinks` list in the [context offloader docs](https://github.com/strands-agents/harness-sdk/blob/main/site/src/content/docs/user-guide/concepts/plugins/context-offloader.mdx?plain=1) with `path: strands-ts/src/vended-plugins/context-offloader/s3-storage.ts` as part of this PR ?

Outside of that, approved!

maxmilian:
Thanks @gautamsirdeshmukh! Done — added `path: strands-ts/src/vended-plugins/context-offloader/s3-storage.ts` to the `sourceLinks` list in `context-offloader.mdx` (commit `daf3bd28`). Ready for another look whenever you have a moment 🙏

maxmilian:
@gautamsirdeshmukh — friendly ping on this one. The `sourceLinks` item you flagged is fixed in `daf3bd28` (bot re-verified it resolves to a real file), and the only remaining note is the `esbuild` devDependency **Suggestion**, which the bot marked non-blocking. This is currently blocked on the earlier `CHANGES_REQUESTED` — would you mind taking another look when you get a chance? Happy to fold in the `esbuild` change too if you'd prefer it in this PR rather than a follow-up.

poshinchen (dismissed):

mehtarac (approved):

poshinchen:
Hi, thanks for the contribution! 

However, we're planning to [merge this larger change](https://github.com/strands-agents/harness-sdk/pull/3099) which overlaps your PR. Therefore we'll close this PR.

PR #3103, opened 2026-07-06
feat(checkpoint): port durable-execution checkpoints to TypeScript

JackYPCOnline:
## Description

TypeScript parity for the provider-agnostic durable-execution **checkpoint** primitive that shipped in the Python SDK (#2190, building on #2181).

A durability provider (Temporal, Dapr, Step Functions) needs a way to persist a pause-point and resume a fresh agent process at a ReAct cycle boundary. Without this, a durable orchestrator can only wrap a whole `invoke()` as one opaque unit and re-executes everything — including already-completed model and tool calls — on crash. A `Checkpoint` is the small serializable marker that makes cycle-boundary resume possible.

A `Checkpoint` is a **position marker** (which boundary fired + cycle index), **not** a state snapshot — conversation-state continuity remains the caller's responsibility via a `SessionManager`. The design mirrors the existing interrupt pause/resume convention: pause with `stopReason: 'checkpoint'`, resume by invoking with a `checkpointResume` payload. It is opt-in and introduces no breaking changes.

### Public API Changes

New, all `@experimental`, reachable from `@strands-agents/sdk` and `@strands-agents/sdk/experimental`:

```typescript

const agent = new Agent({ tools, checkpointing: true })

let result = await agent.invoke('do the thing')
while (result.stopReason === 'checkpoint') {
  persist(result.checkpoint!.toJSON()) // orchestrator persists the marker

  // later — possibly in a fresh process / activity:
  const fresh = new Agent({ tools, checkpointing: true })
  result = await fresh.invoke({ checkpointResume: { checkpoint: load() } })
}
```

The added surface: `Checkpoint`, `CheckpointData`, `CheckpointPosition`, `CheckpointResumeContent`, `CHECKPOINT_SCHEMA_VERSION`, `CheckpointError`, `AgentConfig.checkpointing`, `AgentResult.checkpoint`, and a `'checkpoint'` member on `StopReason`.

### Behavior worth calling out (not obvious from the diff)

- Boundaries only fire on tool-use cycles; a plain end-turn emits no checkpoint.
- Precedence: **cancel** and **interrupt** both take priority over checkpoint at either boundary.
- Multi-agent is out of scope — `CheckpointResumeContent` is excluded from `MultiAgentInput`, so orchestrators reject the resume payload at compile time (node-level checkpointing is future work).

Notable translation choices from the Python source: the TS loop returns `AgentResult` directly (no `EventLoopStopEvent` tuple), cycle tracking is loop-local (so no cross-invocation leak), and `to_dict`/`from_dict` map to `toJSON`/`fromJSON`.

## Related Issues

Ports #2190 (and #2181) to TypeScript.

## Documentation PR

Module-level docstrings cover the V0 contract, precedence, and limitations. A user-guide page is a follow-up.

## Type of Change

New feature

## Testing

Verified the change does not break functionality or introduce new warnings: type-check, unit tests, lint, and format all pass locally.

## Checklist

- [ ] I have updated the documentation accordingly *(module docstrings cover it; user-guide page is a follow-up)*
- [ ] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed *(provider adapter examples are a follow-up milestone)*
- [x] Any dependent changes have been merged and published *(the Python counterparts #2181 and #2190 are merged)*

----

### DIFF
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -297,8 +297,8 @@ export type AgentConfig = {
    */
   toolExecutor?: ToolExecutorStrategy
   /**
-   * When `true`, the agent loop pauses at cycle boundaries (`after_model`,
-   * `after_tools`) and returns `stopReason: 'checkpoint'` with a populated
+   * When `true`, the agent loop pauses at cycle boundaries (`afterModel`,
+   * `afterTools`) and returns `stopReason: 'checkpoint'` with a populated
    * `checkpoint` field. Resume by passing the checkpoint back as
    * `{ checkpointResume: { checkpoint: ... } }`.
    *
@@ -1393,10 +1393,10 @@ export class Agent implements LocalAgent, InvokableAgent {
     let cycleIndex = 0
     let resumePosition: CheckpointPosition | undefined
     if (resumeCheckpoint) {
-      // after_tools means the cycle finished; the next after_model checkpoint
+      // afterTools means the cycle finished; the next afterModel checkpoint
       // belongs to the following cycle.
       cycleIndex =
-        resumeCheckpoint.position === 'after_tools' ? resumeCheckpoint.cycleIndex + 1 : resumeCheckpoint.cycleIndex
+        resumeCheckpoint.position === 'afterTools' ? resumeCheckpoint.cycleIndex + 1 : resumeCheckpoint.cycleIndex
       resumePosition = resumeCheckpoint.position
       currentArgs = undefined
     }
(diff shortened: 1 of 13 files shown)


strands-ts/src/agent/agent.ts
```diff
@@ -295,6 +296,19 @@ export type AgentConfig = {
    * Defaults to `'concurrent'`. See {@link ToolExecutorStrategy} for details.
    */
   toolExecutor?: ToolExecutorStrategy
+  /**
+   * When `true`, the agent loop pauses at cycle boundaries (`after_model`,
+   * `after_tools`) and returns `stopReason: 'checkpoint'` with a populated
```
mehtarac: should after_model and after_tools be camelCase or are these events returned from the model itself as snakecase?

mehtarac: If not from model, and since we're in typescript file, we i think we should make it camelCase

JackYPCOnline: Good catch! Will iterate

mehtarac (approved):

Issue #2951, opened 2026-06-24
[FEATURE] Public API to retrieve registered plugins from Agent / Graph


farisab1:
### SDK Language

Python

### Problem Statement

There is currently no public API to retrieve the plugin instances registered on an
Agent or Graph. Consumers that need to look up a specific plugin attached to an
agent/graph have to reach into private internals to do it, which is brittle and
breaks across releases when those internals change shape.

### Proposed Solution

Add a public getter on both Agent and Graph that returns a registered plugin
instance by type:

```
agent.get_plugin(SomePlugin) -> SomePlugin | None
graph.get_plugin(SomePlugin) -> SomePlugin | None
```

A method to list all registered plugins would also work:

```
agent.get_plugins() -> list[Plugin]
graph.get_plugins() -> list[Plugin]
```

Exact shape is open to whatever fits the SDK best. The key need is retrieving the
plugin instance, not just a boolean of whether one is attached.

Scope is intentionally limited to plugins. This does not expose registered
callbacks directly.

### Use Case

We build a plugin that customers attach to their own Agent and Graph objects.
At runtime we need to retrieve our plugin's reference from the customer's
agent/graph to validate it is properly attached before we rely on it. A public,
type-based getter gives us a stable way to do this without depending on SDK
internals.

PR #3105, opened 2026-07-06
feat(plugins): add get_plugin/get_plugins to Agent, Graph, and Swarm

farisab1:
## Description

  Adds a public, type-based API for retrieving plugin instances registered on an `Agent`, `Graph`, or `Swarm`. Consumers that need to look up a specific plugin can now
  do so without reaching into private internals.

  ```python
  agent.get_plugin(SomePlugin)  # -> SomePlugin | None
  agent.get_plugins()           # -> list[Plugin]
  ```

  The same methods are declared on `MultiAgentBase` (alongside `add_hook`), so every orchestrator carries the API and concrete subclasses (`Graph`, `Swarm`) override
  with implementations backed by their plugin registry. Subclasses that don't support plugins get a `NotImplementedError` — matching how `add_hook` already works.

  Lookup uses `isinstance`, so subclasses of the queried type also match. Returns `None` when no matching plugin is registered — consistent with how the SDK's other
  typed-lookup methods behave (`SessionRepository.read_*`, `StructuredOutputContext.get_result`, `models._defaults.get_context_window_limit`).

  ## Related Issues

#2951 

  ## Documentation PR

  No documentation changes required — reference docs regenerate from the new methods' docstrings.

  ## Type of Change

  New feature

  ## Testing

  Ran `hatch fmt --formatter --check`, `hatch fmt --linter --check`, and `hatch test` from `strands-py/`: 4239 tests pass, ruff and mypy clean.

  Added unit tests covering:
  - `agent.get_plugin(T)` returns the registered instance
  - `agent.get_plugin(T)` returns `None` when no plugin of type `T` is registered
  - `agent.get_plugins()` includes user plugins in registration order
  - Same coverage for `Graph` and `Swarm`

  ## Checklist

  ----

Related to: #2951

### DIFF
--- a/strands-py/src/strands/agent/agent.py
+++ b/strands-py/src/strands/agent/agent.py
@@ -97,6 +97,9 @@ logger = logging.getLogger(__name__)
 # TypeVar for generic structured output
 T = TypeVar("T", bound=BaseModel)
 
+# TypeVar for generic plugin lookup by type
+TPlugin = TypeVar("TPlugin", bound=Plugin)
+
 
 # Sentinel class and object to distinguish between explicit None and default parameter value
 class _DefaultCallbackHandlerSentinel:
@@ -1619,3 +1622,21 @@ class Agent(AgentBase):
             redacted_content = [{"text": redact_message}]
 
         return redacted_content
+
+    def get_plugin(self, plugin_type: type[TPlugin]) -> TPlugin | None:
+        """Return the registered plugin of type ``plugin_type``, or ``None`` if none is registered.
+
+        Args:
+            plugin_type: The plugin class to look up.
+
+        Returns:
+            The matching plugin instance, or ``None`` if no registered plugin matches.
+        """
+        for plugin in self._plugin_registry._plugins.values():
+            if isinstance(plugin, plugin_type):
+                return plugin
+        return None
+
+    def get_plugins(self) -> list[Plugin]:
+        """Return all plugins registered on this agent, in registration order."""
+        return list(self._plugin_registry._plugins.values())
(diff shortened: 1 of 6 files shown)


JackYPCOnline:
Can you follow the guideline creating a issue explaining your use cases ?

farisab1:
> Can you follow the guideline creating a issue explaining your use cases ?

Hey @JackYPCOnline, here is the related issue #2951

opieter-aws:
Use case from ticket:

> We build a plugin that customers attach to their own Agent and Graph objects.
At runtime we need to retrieve our plugin's reference from the customer's
agent/graph to validate it is properly attached before we rely on it. A public,
type-based getter gives us a stable way to do this without depending on SDK
internals.

strands-py/src/strands/agent/agent.py
```diff
@@ -1619,3 +1622,21 @@ def _redact_user_content(self, content: list[ContentBlock], redact_message: str)
             redacted_content = [{"text": redact_message}]
 
         return redacted_content
+
+    def get_plugin(self, plugin_type: type[TPlugin]) -> TPlugin | None:
```
pgrayy, line 1626: For consistency, I would recommend setting up `plugins` and `plugin_registry` properties (`@property`). We can then add a get method to the plugin_registry to get plugins by name instead of type (similar to the tool registry in [TS](https://github.com/strands-agents/harness-sdk/blob/main/strands-ts/src/registry/tool-registry.ts#L61)).

This will help to promote consistency across languages and the plugin and tool registries.

pgrayy (changes requested):

strands-py/src/strands/agent/agent.py
farisab1: Hey @pgrayy, sounds good, by exposing `plugin_registry` as a property that would make PluginRegistry part of the public API. Should we make it public? Or just expose getters for `plugins` property and a `get_plugin` method?

PR #3113, opened 2026-07-07
feat: expose client_name parameter on MCPClient for clientInfo identification

amazonanonymous:
## Description

Adds a `client_name` parameter to `MCPClient` that passes through to `ClientSession` as `clientInfo.name` during the MCP initialize handshake. This allows agents to identify themselves to MCP servers instead of all appearing as the generic SDK default `"mcp"`.

### Usage

```python
client = MCPClient(transport, client_name="my-fraud-agent")
```

The MCP spec provides `clientInfo.name` for client identification, and the TypeScript SDK already uses it (`strands-agents-ts-sdk`). The Python SDK accepts the parameter on `ClientSession` but Strands never passes it through — this change exposes it.

## Related Issues

None

## Documentation PR

No documentation changes needed — the parameter is self-documenting via its docstring.

## Type of Change

New feature

## Testing

## Checklist

----


mkmeral:

strands-py/src/strands/tools/mcp/mcp_client.py
```diff
@@ -886,6 +892,9 @@ async def _async_background_thread(self) -> None:
                     write_stream,
                     message_handler=self._handle_error_message,
                     elicitation_callback=self._elicitation_callback,
+                    client_info=(
+                        Implementation(name=self._client_name, version="0.1.0") if self._client_name else None
+                    ),
```
strandly-the-agent, line 962: 🟡 **`version="0.1.0"` is false precision — the weakest part of the change.**

This value is byte-identical to the MCP SDK's own placeholder (`DEFAULT_CLIENT_INFO = Implementation(name='mcp', version='0.1.0')`), so it carries no signal on its own. The problem is that paired with a *real* `client_name`, a server operator reading `{name: "my-fraud-agent", version: "0.1.0"}` will reasonably take `0.1.0` to be *the agent's* version — and be wrong every time. That's arguably worse than today's honest-generic `"mcp"`, which the PR is trying to improve on.

What should this version *mean*? Two clean options:

- **The Strands SDK version** — `version("strands-agents")` (precedent: `strands/telemetry/config.py:9,38`), so servers can identify Strands clients and versions. Needs `from importlib.metadata import version` at the top of the module.
- **Caller-supplied** — mirror the TS SDK's `applicationVersion` and add a `client_version` param alongside `client_name`.

Whichever you pick, the hardcoded literal is the one option that's neither honest nor caller-controllable. (I left this as a question rather than a one-click suggestion because the SDK-version route also needs the import, and the choice is a genuine design call.)

strandly-the-agent (commented):
**TL;DR — changes-requested (soft).** The change is correct, minimal, and well-tested — I ran the suite (7/7 new + 73/73 in-file pass), so there's no bug here. But it ships a hardcoded placeholder `version="0.1.0"` that a server operator will misread as the *agent's* version, and it diverges from the TS SDK that already shipped this exact concept. One inline note on the version; a config-parity gap, a test-slimming idea, and three questions below. Nothing here is a correctness gate — flagging for a human to approve, not blocking.

✅ **Verified:** branch `pr3113` @ `a7d0369` · `pytest -k client_name` → **7/7**, full `test_mcp_client.py` → **73/73** · the `client_info`/name passthrough is mutation-checked (genuinely tested, not mock-echo) · `version="0.1.0"` is byte-identical to the MCP SDK's own placeholder (`DEFAULT_CLIENT_INFO = Implementation(name='mcp', version='0.1.0')`) · real-version precedent exists in-repo at `strands/telemetry/config.py:9,38`.

<details>
<summary>🟡 Config-path parity — <code>load_servers</code> users can't set <code>client_name</code></summary>

`_build_client_from_config` threads `startup_timeout` / `tool_filters` / `prefix` from the server config but drops the new param, so the declarative/JSON path (`MCPClient.load_servers(...)`) can't set `client_name` at all — only the direct constructor can. The TS SDK not only threads it but *defaults* it to the server key (`strands-ts/src/mcp/config.node.ts:120`). To match:

```python
    return MCPClient(
        transport_callable,
        startup_timeout=server.get("startup_timeout", 30),
        tool_filters=_parse_config_tool_filters(name, server.get("tool_filters")),
        prefix=server.get("prefix"),
        client_name=server.get("client_name", name),  # default to server key, like the TS SDK
    )
```

(If you'd rather keep it opt-in, `server.get("client_name")` is fine — flagging the choice. This also means `client_name` should be added to the `MCPServerConfig` TypedDict so it isn't warned as an unknown key.)
</details>

<details>
<summary>🟡 Tests — 89 lines / 7 tests for a ~9-line change, mostly duplication</summary>

The coverage is real (I mutation-tested it — the name passthrough and the empty-string→None branch are genuinely caught), but it's disproportionate: `client_name_none`, `client_name_not_provided`, and `client_name_default_is_none` all exercise the identical default-`None` path, and the 5 `call_args` tests repeat the same ~6-line mock block. This collapses cleanly to one `@pytest.mark.parametrize`. Two smaller points while there:

- `assert client_info.version == "0.1.0"` just echoes a hardcoded literal — it's a change-detector with no bug signal. It becomes worth keeping once the version is meaningful (see the inline note).
- The two tests asserting the private `_client_name` attribute add no coverage the `call_args` tests don't already give, and the `mock_session` fixture is requested but dead (conftest already patches `ClientSession`, and each test re-patches on top). Both can go.

Not blocking — a nice-to-have cleanup. The one gap a mock can't cover: an integration test in `tests_integ/mcp/` that boots a real `ClientSession` and asserts the name reaches the `initialize` handshake (a renamed/removed `client_info` kwarg would sail past every mock test here).
</details>

<details>
<summary>❓ Questions for a human — worth a decision before merge</summary>

1. **Cross-SDK consistency (the big one).** The TS SDK already shipped this: `applicationName` / `applicationVersion` params, a `clientName` getter, config-threading with a default (`strands-ts/src/mcp/client.ts:33-34,160-161,248`). This PR picks a different name (`client_name`), exposes name-only, and doesn't thread through config. Given the "match the sibling" tenet, is the divergence intentional? The param name in particular is a one-way door once released.
2. **Surface shape.** `client_name: str` now, or `client_info: mcp.types.Implementation | None`? The existing `elicitation_callback` / `progress_callback` params already expose MCP SDK types, so a passthrough object would be consistent and get name+version+title in one. Name-only is additively safe (you can add `client_version` later without breaking), so this isn't blocking — just worth a deliberate call now.
3. **API-review label.** This adds public constructor surface but carries neither `needs-api-review` nor `completed-api-review`. Per the workflow it needs `needs-api-review` before merge; combined with the two forks above, I'd treat it as borderline needs-design-discussion rather than a solo approve.
</details>

<sub>🤖 Automated review by Strandly (correctness · api/devx · test-quality passes, aggregated). I'm an experimental agent — this is solid work for a human to approve, not a gate. Push back where I'm wrong.</sub>

strands-py/src/strands/tools/mcp/mcp_client.py
amazonanonymous: Thank you for the thorough review — really helpful feedback. Updated the PR to address:

**✅ Fixed:**
- **`version="0.1.0"` hardcoded** — Now uses `pkg_version("strands-agents")` as default (the actual SDK version), with optional `application_version` param for callers who want to report their own version.
- **Config-path parity** — `_build_client_from_config` now threads `application_name` (defaulting to the server key name, matching the TS SDK) and `application_version`.
- **Tests over-duplicated** — Collapsed to 2 parametrized tests (5 cases total). Removed the attribute-asserting tests and dead fixture usage.
- **Cross-SDK consistency** — Renamed `client_name`/`client_version` → `application_name`/`application_version` to match the TS SDK naming.

**Not addressed (need maintainer input):**
- **`needs-api-review` label** — It seems I can't add labels as an external contributor (but could be wrong). Could a maintainer add this? Thank you
- **Surface shape** — Keeping `application_name: str` + `application_version: str` as separate params rather than exposing `Implementation` directly. This matches the TS SDK approach and is additively safe for future fields.

mkmeral:
@strandly-the-agent review again please. is it good to merge after resolving merge conflicts with main?

mkmeral (dismissed):

strandly-the-agent:
**Re-review @ `4501937` — all my earlier findings are addressed, and yes, it's good to merge once conflicts are resolved.** I resolved the conflicts locally and ran the suite: **106 passed** (`test_mcp_client.py` + `test_mcp_client_load_servers.py`). One caveat on *how* to resolve — see below.

✅ **Addressed & verified:**
| Prior finding | Fix at `4501937` |
|---|---|
| Hardcoded `version="0.1.0"` | `version=self._application_version or pkg_version("strands-agents")` — real SDK version by default, caller override optional ✅ |
| Config-path parity | `_build_client_from_config` threads `application_name` (defaults to server key, like TS) + `application_version`; both added to `MCPServerConfig` ✅ |
| Cross-SDK naming | renamed → `application_name` / `application_version` ✅ |
| Test over-duplication | collapsed to 2 parametrized tests (5 cases); dead fixture + private-attr tests removed ✅ |

<details>
<summary>⚠️ Merge-conflict resolution — mechanical, but one spot needs a <em>key-merge</em>, not a pick-a-side</summary>

The conflicts come from the `continue_on_error` PR that landed in `main`, which added a constructor param in the exact same four locations this PR touches (signature, docstring, `__init__` body, `_build_client_from_config`). Those four are **keep-both** — the params are independent, no logic overlap.

The one that isn't purely mechanical is `tests/strands/tools/mcp/test_mcp_client_load_servers.py`: both PRs added keys to an **exact-equality** `assert kwargs == {...}`. Keeping both hunks would duplicate/desync the dict — you have to *merge the keys* into one dict, e.g.:
```python
    assert kwargs == {
        "startup_timeout": 5,
        "tool_filters": None,
        "prefix": "p",
        "application_name": "srv",
        "application_version": None,
        "continue_on_error": False,
    }
```
(same for the `test_default_startup_timeout` dict). I confirmed this resolution builds and the full MCP suite is green.
</details>

<details>
<summary>Two non-blocking heads-ups</summary>

- **Config path now makes every `load_servers` client self-identify by its server key** (`application_name=server.get("application_name", name)`), where previously they all reported the MCP default `"mcp"`. This is a small wire-behavior change on the initialize handshake — it's what I recommended and matches the TS SDK, so desirable, just worth being aware it's not opt-in for the config path. (Tests updated to expect it.)
- **`application_version` set *without* `application_name` is silently ignored** (`client_info` stays `None`, since MCP `Implementation` requires a name). Reasonable — a version with no name is meaningless — but a one-line docstring note would save a future head-scratch.

Still worth a maintainer adding the `needs-api-review` label (the author couldn't as an external contributor), per the API-review workflow.
</details>

<sub>🤖 Automated re-review by Strandly. `mergeable` is currently `CONFLICTING`/`DIRTY` — resolve per the note above and it's ready. Solid work for a human to approve, not a gate.</sub>

yonib05 (approved):

Unshure (approved):

PR #3114, opened 2026-07-07
feat(mcp): preserve MCP content annotations in tool results

### LINKED ISSUE #2919: [FEATURE] Preserve MCP content annotations (audience, priority, lastModified) in tool results
### SDK Language

Python

### Problem Statement

The MCP spec (2025-06-18) defines annotations on content blocks with audience (array of "user" and/or "assistant"), priority (0.0-1.0), and lastModified (ISO 8601 formatted timestamp) fields. This allows MCP servers to differentiate content intended for the model's reasoning vs. content that should reach the end user.

Currently, `map_mcp_content_to_tool_result_content` in `src/strands/tools/mcp/mcp_client.py` strips annotations when mapping MCP content to `ToolResultContent`. The `ToolResultContent` `TypedDict` in `src/strands/types/tools.py` has no annotations field. Annotations are silently lost during mapping.

This makes it impossible for agent frameworks to respect audience-differentiated content from MCP servers.

Spec reference: https://modelcontextprotocol.io/specification/2025-06-18/server/resources#annotations
Code reference: https://github.com/strands-agents/harness-sdk/blob/main/strands-py/src/strands/tools/mcp/mcp_client.py#L867
https://github.com/strands-agents/harness-sdk/blob/main/strands-py/src/strands/types/tools.py#L70

### Proposed Solution

1. Add an optional annotations field to `ToolResultContent` in `src/strands/types/tools.py`
2. Preserve content.annotations during mapping in `map_mcp_content_to_tool_result_content` in `src/strands/tools/mcp/mcp_client.py`
3. Expose annotations downstream so agent frameworks can filter content by audience when constructing model prompts vs. user-facing output

### Use Case

MCP servers need to return content with differentiated audiences:
  - `audience: ["assistant"]` -- internal context (applied filter values, debug metadata, reasoning hints) that shouldn't be surfaced to users
  - `audience: ["user"]` -- formatted responses for the user that don't need model context tokens
  - `audience: ["user", "assistant"]` -- default, both

  Example: A knowledge base MCP server returns search results (audience: both) alongside applied personalization metadata (audience:
  assistant-only). Without annotations, the agent cannot distinguish these and may surface internal metadata to the user.

### Alternatives Solutions

_No response_

### Additional Context

- Code to modify: https://github.com/strands-agents/sdk-python/blob/main/src/strands/tools/mcp/mcp_client.py (map_mcp_content_to_tool_result_content)
- Type to extend: https://github.com/strands-agents/sdk-python/blob/main/src/strands/types/tools.py (ToolResultContent)

iakbh01:
## Summary

Preserves MCP content annotations (`audience`, `priority`, `lastModified`) when mapping MCP content blocks to `ToolResultContent`. Currently these annotations (part of the MCP spec 2025-06-18) are silently dropped, making it impossible for downstream agent frameworks to respect audience-differentiated content.

This PR does exactly two things, scoped small and additive:

1. Adds an optional `annotations` field (backed by a new `ContentAnnotations` TypedDict) to `ToolResultContent`.
2. Preserves `content.annotations` when mapping each MCP content type (text, image, embedded resource) in `map_mcp_content_to_tool_result_content`.

Closes #2919 (steps 1-2). Downstream audience-based filtering (issue step 3) is intentionally left as a follow-up to keep this change small and focused.

## Backward compatibility

`annotations` is optional (`total=False`) and only emitted when the source MCP content actually carries it. Content without annotations maps byte-for-byte as before — verified by test.

## API surface (for API review)

> **Note to maintainers:** I don't have permission to add labels as an external contributor. Per `team/API_BAR_RAISING.md` this is a "moderate" change (new public type + new field on a widely-used TypedDict) — please add `api/needs-review`.

**New public type**, exported from `strands.types.tools` (same module as `ToolResultContent`, `ToolResult`, `ToolSpec`, etc.):

```python
class ContentAnnotations(TypedDict, total=False):
    audience: list[Literal["user", "assistant"]]
    priority: float
    lastModified: str
```

**Changed public type** — `ToolResultContent` (`strands.types.tools`) gains one new optional key:

```python
class ToolResultContent(TypedDict, total=False):
    annotations: ContentAnnotations  # new
    document: DocumentContent
    image: ImageContent
    json: Any
    text: str
```

No new module-level exports are needed beyond importing `ContentAnnotations` alongside the existing `ToolResultContent` import — there is no `strands.types.__init__` re-export list for these tool types today (they're imported directly from `strands.types.tools`), so this follows the existing convention rather than introducing a new one.

### End-to-end usage example

```python
from strands.tools.mcp import MCPClient
from strands.types.tools import ToolResult, ToolResultContent, ContentAnnotations

# 1. Using the MCP client (the primary way this field gets populated today):
#    an MCP server attaches annotations to a content block...
#      TextContent(type="text", text="internal reasoning trace",
#                   annotations=Annotations(audience=["assistant"], priority=0.9))
#    ...and MCPClient.call_tool_sync() now surfaces them on the mapped result:
tool_result: ToolResult = mcp_client.call_tool_sync(
    tool_use_id="t1", name="some_tool", arguments={}
)
for block in tool_result["content"]:
    annotations = block.get("annotations")
    if annotations and "assistant" not in annotations.get("audience", ["user", "assistant"]):
        continue  # e.g. skip content not meant to reach the end user
    ...

# 2. A non-MCP tool author can populate the same field directly, since
#    ContentAnnotations is just a plain TypedDict:
def my_tool() -> ToolResult:
    content: ToolResultContent = {
        "text": "debug-only detail",
        "annotations": ContentAnnotations(audience=["assistant"], priority=0.2),
    }
    return {"toolUseId": "t1", "status": "success", "content": [content]}
```

### Design question: is `ToolResultContent` the right home?

Raised by review: `annotations` is a generic field on `ToolResultContent`, but today it's only ever populated by the MCP client. Is `ToolResultContent` the right place, versus something MCP-specific?

I went with `ToolResultContent` (the shared type) rather than an MCP-only wrapper for a few reasons:

- **Tenet: "Embrace common standards."** `audience`/`priority`/`lastModified` are not MCP-proprietary concepts — they're a generic "how should this content be treated" contract that other content sources (a future A2A integration, a custom tool, a vended plugin) could reasonably want to set too. Modeling them as MCP-only would mean re-inventing the same shape later if/when another producer needs it.
- **Tenet: "Composability."** `ToolResultContent` is already the single shared shape every tool provider returns into (`document`/`image`/`json`/`text`); a parallel MCP-only variant would fork that contract and force every downstream consumer (conversation managers, model providers, plugins) to special-case MCP content instead of treating all `ToolResultContent` blocks uniformly.
- **Precedent in this same type**: `json` is also a field that today is populated from a fairly narrow set of call sites, yet lives directly on `ToolResultContent` rather than a wrapper.

**For a non-MCP tool author**, the field is optional and purely advisory metadata — nothing in the SDK currently reads it to change behavior (model providers explicitly allowlist known keys when formatting requests, so `annotations` is never forwarded to a model API; see "Testing" below). A tool author should treat it the same way an MCP server does: set `audience` when a piece of content is meant only for the model's internal reasoning vs. the end user, `priority` for relative importance among multiple content blocks, and `lastModified` for freshness of resource-backed content. It's inert until a consumer (agent, plugin, or future SDK feature) chooses to read it — which is intentionally left as a follow-up (issue step 3) rather than bundled into this PR.

## Testing

- Added 9 tests total to `tests/strands/tools/mcp/test_mcp_client.py` covering: text with full annotations, text without annotations (backward-compat guard), image with/without annotations (asserted via full-dict equality), partial annotations (only present keys emitted), embedded text-resource with annotations, embedded blob-as-text with annotations, embedded blob-as-image with annotations, and a dropped embedded-resource path still returning `None`.
- Refactored `map_mcp_content_to_tool_result_content` to route all five content branches through a single `_with_annotations()` helper, eliminating the previously copy-pasted "map → attach if present → return" pattern and removing drift risk between branches.
- Tightened `_map_mcp_annotations` to take a concrete `MCPTextContent | MCPImageContent | MCPEmbeddedResource` union instead of `Any`, so mypy checks the `.annotations` access instead of relying on `getattr` fallback typing.
- `ruff format --check`, `ruff check`, and `mypy` pass on all changed files.
- Full test suite passes (1142 tests across tools/mcp, types, models/bedrock, and agent test dirs). All annotation-handling lines are covered per `--cov-report=term-missing` (the module's remaining coverage gaps are pre-existing, unrelated edge cases — e.g. blob-decode-exception and defensive unreachable branches).
- Verified other model providers (bedrock.py, writer.py) explicitly allowlist known `ToolResultContent` keys when formatting requests to their respective APIs, so the new `annotations` key cannot leak into a strict external API and cause issues.


iakbh01:
Hi @gautamsirdeshmukh — all automated feedback has been addressed in `0417e24`, and upstream `main` was merged in `3541b17`. The conflict is resolved, and the targeted tests, Ruff, and mypy pass. Could you please review this PR and, if API review is required, add the `api/needs-review` label?

PR #3115, opened 2026-07-07
fix(ci): create release tag via git push instead of gh release --target

poshinchen:
## Description

The release workflows create the GitHub release with:

```bash
gh release create "$NEW_TAG" --target "$RELEASE_SHA" ...
```

With `--target <sha>`, the releases API creates the tag ref as a side effect of the `POST /releases` call, and that call is refused for the Actions `GITHUB_TOKEN` with:

```
HTTP 403: Resource not accessible by integration
```

The job already declares `permissions: contents: write`, and a plain `git push` of a ref with the same token succeeds in this repo (e.g. `docs-deploy-github-pages.yml` pushes to `gh-pages` with the default token).

This PR splits the two operations:

- **Create the tag over git** — `git tag "$NEW_TAG" "$RELEASE_SHA"` then `git push origin`, so the tag is created through the git protocol (which works) rather than the releases API (which 403s). The checkout for this job sets `persist-credentials: true` so the push authenticates with the job's `GITHUB_TOKEN`; every other job keeps `persist-credentials: false`.
- **Attach the release to the existing tag** — `gh release create "$NEW_TAG"` without `--target`, so the API only creates the release object and never touches the tag.

The SHA pin is preserved: the tag still lands on the exact reviewed `release_sha`, so the tagged commit matches the built/inspected/approved artifact. Applied identically to `release-python.yml` and `release-typescript.yml`.

## Related Issues

Fixes the `create-gh-release` 403 seen on the first run of the release pipeline (introduced in #2940).

## Type of Change

Bug fix

## Checklist

----

### DIFF
--- a/.github/workflows/release-python.yml
+++ b/.github/workflows/release-python.yml
@@ -377,24 +377,43 @@ jobs:
       - uses: actions/checkout@v7
         with:
           ref: ${{ needs.scan-commits.outputs.release_sha }}
-          persist-credentials: false
+          # This job creates the tag, so keep the GITHUB_TOKEN in git config
+          # for the push below. Other jobs use persist-credentials: false.
+          persist-credentials: true
 
       - name: Download release notes artifact
         uses: actions/download-artifact@v8
         with:
           name: release-notes
 
-      - name: Create release (atomically tags the pinned SHA)
+      - name: Tag the pinned SHA and create the release
         env:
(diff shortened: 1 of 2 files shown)


JackYPCOnline:
/strands Will persisted credential be removed after workflow finished?

yonib05 (approved):

PR #3116, opened 2026-07-07
feat(ts-metrics): expose metrics getter on LocalAgent

lizradway:
## Description

Exposes `agent.metrics` as a public property on `LocalAgent` in the TypeScript SDK, matching the Python SDK's `agent.event_loop_metrics`. This lets plugins read accumulated metrics (cycle count, token usage, tool stats) directly from the agent reference they already have, rather than waiting for an `AgentResult` or reaching into private fields.

The `AgentMetrics` class is already public and returned on `AgentResult` — this just makes it accessible on the agent itself, which is what plugins receive in `initAgent` and hook callbacks.

## Documentation PR

N/A — no new docs needed; `AgentMetrics` is already documented and exported.

## Type of Change

New feature

## Testing

- [x] Added a unit test verifying `agent.metrics.cycleCount` is 0 before invocation and 1 after
- [x] Type check passes (`tsc --noEmit`)
- [x] All 222 existing + new unit tests pass

## Checklist

----

### DIFF
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -104,7 +104,7 @@ import { MemoryManager } from '../memory/memory-manager.js'
 import type { MemoryManagerConfig } from '../memory/index.js'
 import { SessionManager } from '../session/session-manager.js'
 import { Tracer } from '../telemetry/tracer.js'
-import { Meter } from '../telemetry/meter.js'
+import { AgentMetrics, Meter } from '../telemetry/meter.js'
 import type { AttributeValue } from '@opentelemetry/api'
 import { logger } from '../logging/logger.js'
 import { CancelledError } from '../errors.js'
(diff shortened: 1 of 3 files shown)


strands-ts/src/agent/agent.ts
```diff
@@ -891,6 +891,13 @@ export class Agent implements LocalAgent, InvokableAgent {
     return this._toolRegistry
   }
 
+  /**
+   * Read-only snapshot of accumulated agent metrics (cycles, token usage, tool stats).
```
opieter-aws, line 895: nit: technically not readOnly, since they're mutable

opieter-aws (approved):

strands-ts/src/agent/agent.ts
opieter-aws: `snapshot view` would be more accurate

lizradway: yeah fair, can update in follow up.

`AgentMetrics` properties are typed readonly, although the objects within are mutable

PR #3117, opened 2026-07-07
fix(session): encode bytes in SessionAgent.to_dict() for JSON serialization

### LINKED ISSUE #1864: [BUG] S3SessionManager JSON serialization fails when graph state contains binary document content from multimodal prompts
### Checks

### Strands Version

1.30.0

### Python Version

3.13

### Operating System

Ubuntu 24.04.4 LTS on WSL2

### Installation Method

pip

### Steps to Reproduce

1. Initialize a multi-node agent **graph** (using `GraphBuilder`) with `S3SessionManager` enabled for conversation persistence (`builder.set_session_manager(s3_session_manager)`)
2. Create a multimodal prompt containing an inline PDF document using the format expected by the LiteLLM proxy endpoint:
   ```python
   prompt = [
       {"text": "Analyze this PDF"},
       {
           "document": {
               "format": "pdf",
               "name": "document.pdf",
               "source": {
                   "bytes": pdf_bytes  # Raw binary content
               },
           }
       },
   ]
   ```
3. Invoke `graph.stream_async(prompt, config)` with a session ID
4. Allow the graph to attempt to persist its state via `S3SessionManager`

### Expected Behavior

The graph should successfully execute the multimodal request and persist the conversation state to S3, with binary payloads either:
- Automatically encoded (e.g., base64) before JSON serialization, or
- Excluded from the persisted state if they are unused after the model invocation

### Actual Behavior

Graph execution fails with:
```
TypeError: Object of type bytes is not JSON serializable
```

The error originates from `S3SessionManager` attempting to serialize the entire graph state using `json.dumps()`, which cannot handle Python `bytes` objects.

Stack trace context: `<hidden>/strands/session/s3_session_manager.py` line performing:
```python
content = json.dumps(data, indent=2, ensure_ascii=False)
```

### Problem Statement
There is a fundamental incompatibility between:
1. **Model adapter requirement**: The LiteLLM proxy's OpenAI-compatible formatter has a **hardcoded dependency on `document.source.bytes`** (raw Python bytes) and does not support `document.source.s3Location.uri` references. The Strands SDK OpenAI/LiteLLM adapter layer inherits this constraint.
2. **Session persistence requirement**: `S3SessionManager` assumes all graph state is JSON-serializable and does not pre-process binary payloads before calling `json.dumps()`

### How We Discovered This
- **Attempt 1** (S3-backed document reference): Tried using `document.source.s3Location.uri` to avoid embedding bytes, but the LiteLLM proxy's request formatter raised `KeyError: 'bytes'` because it expects and immediately base64-encodes inline bytes.
- **Attempt 2** (Inline bytes): Provided the expected `document.source.bytes` format, which fixed the LiteLLM proxy formatter error but immediately exposed the session persistence incompatibility when `S3SessionManager` attempted to serialize the multi-node graph state.

### Current Constraint
As long as both multimodal requests and `S3SessionManager` are enabled in the same agent graph, they cannot coexist for requests containing binary document attachments. This is particularly problematic when using LiteLLM proxy endpoints, which provide no alternative document source format (S3 URIs are not supported).

### Possible Solution

**Option A: Pre-process binary payloads in S3SessionManager**
Modify the session manager to detect and encode non-JSON-serializable types (e.g., `bytes`) before calling `json.dumps()`:
```python
# Before serialization
def make_json_serializable(obj):
    if isinstance(obj, bytes):
        return {"__bytes_b64__": base64.b64encode(obj).decode("utf-8")}
    # ... recursively handle dict/list
    return obj

data = make_json_serializable(graph_state)
content = json.dumps(data, indent=2, ensure_ascii=False)
```

**Option B: Exclude binary fields from persisted state**
Implement a state serialization filter that strips or replaces binary document content before persistence, then reconstructs it from the request context if needed for subsequent graph steps.

**Option C: Support S3-backed document sources in the LiteLLM proxy (upstream)**
This would require the LiteLLM proxy to support `document.source.s3Location.uri` as a valid document source format instead of requiring inline base64-encoded `bytes`. However, this is a limitation of the LiteLLM proxy itself (not Strands), so the more practical fix for Strands users would be Option A or B.

### Related Issues

_No response_

gautamsirdeshmukh:
## Description

Saving a session fails with `TypeError: Object of type bytes is not JSON serializable` when a `SessionAgent`'s state holds binary data.

### Cause

This can occur with multimodal prompts. If a document (e.g. a PDF) is sent inline as raw bytes (which is what a LiteLLM proxy requires) those bytes end up in the agent's state. Then, when an `S3SessionManager` or `FileSessionManager` saves the session, it converts that `SessionAgent` to JSON, and JSON can't handle raw bytes, leading to the error.

### Solution

Make `SessionAgent` convert bytes to base64 when saving (`to_dict()`) and back to bytes when loading (`from_dict()`) — exactly what `SessionMessage` already does. `Session` is left unchanged, since it never holds bytes.

## Related Issues

Closes #1864 

## Type of Change

Bug fix

## Testing

## Checklist

----

### DIFF
--- a/strands-py/src/strands/types/session.py
+++ b/strands-py/src/strands/types/session.py
@@ -166,11 +166,12 @@ class SessionAgent:
     @classmethod
     def from_dict(cls, env: dict[str, Any]) -> "SessionAgent":
         """Initialize a SessionAgent from a dictionary, ignoring keys that are not class parameters."""
-        return cls(**{k: v for k, v in env.items() if k in inspect.signature(cls).parameters})
+        extracted_relevant_parameters = {k: v for k, v in env.items() if k in inspect.signature(cls).parameters}
+        return cls(**decode_bytes_values(extracted_relevant_parameters))
 
     def to_dict(self) -> dict[str, Any]:
         """Convert the SessionAgent to a dictionary representation."""
-        return asdict(self)
+        return encode_bytes_values(asdict(self))  # type: ignore[no-any-return]
 
     def initialize_internal_state(self, agent: "Agent") -> None:
         """Initialize internal state of agent."""
--- a/strands-py/src/strands/types/session.py
+++ b/strands-py/src/strands/types/session.py
@@ -101,7 +101,7 @@ class SessionMessage:
 
     def to_dict(self) -> dict[str, Any]:
         """Convert the SessionMessage to a dictionary representation."""
-        return encode_bytes_values(asdict(self))  # type: ignore
+        return encode_bytes_values(asdict(self))  # type: ignore[no-any-return]
 
 
 @dataclass
--- a/strands-py/tests/strands/types/test_session.py
+++ b/strands-py/tests/strands/types/test_session.py
@@ -151,7 +151,7 @@ def test_session_agent_with_bytes():
         state={"document": {"format": "pdf", "source": {"bytes": b"This is binary data"}}},
     )
 
-    # json dumps will fail if its not json serializable
+    # json dumps will fail if it's not json serializable
     agent_json_string = json.dumps(session_agent.to_dict())
 
     # Load it back and verify the full state round-trips, bytes included
--- a/strands-py/tests/strands/types/test_session.py
+++ b/strands-py/tests/strands/types/test_session.py
@@ -140,3 +140,20 @@ def test_session_agent_initialize_internal_state():
     tru_model_state = agent._model_state
     exp_model_state = {"response_id": "resp_abc"}
     assert tru_model_state == exp_model_state
+
+
+def test_session_agent_with_bytes():
+    # Agent state can hold binary content (e.g. inline PDF bytes from a multimodal prompt), which
+    # crashes json.dumps() unless to_dict() encodes it. Regression test for #1864.
+    session_agent = SessionAgent(
+        agent_id="a1",
+        conversation_manager_state={},
+        state={"document": {"format": "pdf", "source": {"bytes": b"This is binary data"}}},
+    )
+
+    # json dumps will fail if its not json serializable
+    agent_json_string = json.dumps(session_agent.to_dict())
+
+    # Load it back and verify the full state round-trips, bytes included
+    loaded_agent = SessionAgent.from_dict(json.loads(agent_json_string))
+    assert loaded_agent.state == session_agent.state


strands-py/tests/strands/types/test_session.py
```diff
@@ -140,3 +140,20 @@ def test_session_agent_initialize_internal_state():
...
+        conversation_manager_state={},
+        state={"document": {"format": "pdf", "source": {"bytes": b"This is binary data"}}},
+    )
+
+    # json dumps will fail if its not json serializable
+    agent_json_string = json.dumps(session_agent.to_dict())
+
+    # Load it back and verify the bytes round-trip
+    loaded_agent = SessionAgent.from_dict(json.loads(agent_json_string))
+    assert loaded_agent.state["document"]["source"]["bytes"] == b"This is binary data"
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Addressed

strands-py/src/strands/types/session.py
mehtarac: NIT: I like the scoped `# type: ignore[no-any-return]` in the implementation. Could improve it here also even though you didn't touch this code

pgrayy (dismissed):

strands-py/tests/strands/types/test_session.py
```diff
@@ -140,3 +140,20 @@ def test_session_agent_initialize_internal_state():
...
+def test_session_agent_with_bytes():
+    # Agent state can hold binary content (e.g. inline PDF bytes from a multimodal prompt), which
+    # crashes json.dumps() unless to_dict() encodes it. Regression test for #1864.
+    session_agent = SessionAgent(
+        agent_id="a1",
+        conversation_manager_state={},
+        state={"document": {"format": "pdf", "source": {"bytes": b"This is binary data"}}},
+    )
+
+    # json dumps will fail if its not json serializable
```
mehtarac: NITTIEST OF NITS: `it's` and not `its`

mehtarac (dismissed):

strands-py/src/strands/types/session.py
gautamsirdeshmukh: I'll include this PR, there's a fair amount of other areas in the codebase that could use a similar update as well which I'll take care of in a follow-up PR

strands-py/tests/strands/types/test_session.py
gautamsirdeshmukh: I could blame the fact that this was preexisting but I shan't. High bar +500

gautamsirdeshmukh: (addressing)

mehtarac (approved):

Issue #2296, opened 2026-05-15
bug: repository_session_manager adds incorrect number of toolResult blocks when fixing orphaned toolUse


xiaosu19:
## Bug Description

When `AgentCoreMemorySessionManager` restores conversation history that contains orphaned `toolUse` blocks (toolUse without matching toolResult), the `repository_session_manager` attempts to fix the history by adding toolResult blocks. However, the fix adds an incorrect number of toolResult blocks, causing Bedrock's ConverseStream API to reject the request with:

```
ValidationException: The number of toolResult blocks at messages.N.content exceeds the number of toolUse blocks of previous turn.
```

## Steps to Reproduce

1. Use `AgentCoreMemorySessionManager` with Bedrock AgentCore Memory enabled
2. During an agent invocation, have a tool call interrupted (e.g., MCP server timeout, network error) — this leaves a `toolUse` block in the assistant message without a corresponding `toolResult` in the next user message
3. Memory saves this incomplete conversation history
4. On the next invocation in the same session, Memory restores the history
5. `repository_session_manager` detects the orphaned toolUse and logs:
   ```
   WARNING:strands.session.repository_session_manager:Session message history has an orphaned toolUse with no toolResult. Adding toolResult content blocks to create valid conversation.
   ```
6. The agent then logs:
   ```
   INFO:strands.agent.agent:Agents latest message is toolUse, appending a toolResult message to have valid conversation.
   ```
7. The combined fixes result in more toolResult blocks than toolUse blocks, and Bedrock rejects the request

## Expected Behavior

The repair logic should ensure the number of toolResult blocks exactly matches the number of toolUse blocks in the previous assistant message.

## Environment

- strands-agents: 1.39.0
- bedrock-agentcore: 1.9.1
- Model: zai.glm-5 via Bedrock ConverseStream API
- Region: us-east-2

## Workaround

Manually validate and fix the message history after session restoration, before calling the model.

lizradway:
Cut https://github.com/strands-agents/harness-sdk/issues/3118 for more permanent structural fix to this reoccurring bug

lizradway:
closing out fixed in https://github.com/strands-agents/harness-sdk/pull/3119

PR #3119, opened 2026-07-07
fix(py): declarative rebuild for _fix_broken_tool_use

lizradway:
## Description

`_fix_broken_tool_use` previously detected and repaired broken toolUse/toolResult pairings through sequential mutation steps: find stale results, remove them, recompute what's missing, then extend/insert/reuse depending on the state. Each step read state that the previous step mutated, and each edge case (stale-only results, partial results, empty messages) required its own branch.

This PR replaces the loop body with a declarative rebuild:

1. **Partition** the next message's content into a `dict[toolUseId → ContentBlock]` and a list of non-toolResult blocks.
2. **Short-circuit** if the result IDs already match the toolUse IDs exactly.
3. **Ensure a slot** exists (insert an empty user message only when the next message is text-only).
4. **Rebuild unconditionally**: `[existing.get(tid, error_result(tid)) for tid in tool_use_ids] + non_tool_result_content`.

Correctness falls out of the data structure choice — stale results are never selected from the dict, missing results are filled by the default, and the comprehension produces exactly one result per toolUse ID by definition.

## Related Issues

#2296

## Documentation PR

N/A — internal repair logic with no public API surface.

## Type of Change

Other (please describe): Refactor of internal session repair logic that also fixes the stale-toolResult bug (#2296).

## Testing

- All 44 existing tests in `test_repository_session_manager.py` pass unchanged
- Added `test_fix_broken_tool_use_removes_stale_tool_results` covering the case where the next message has toolResult IDs that don't match any toolUse in the preceding assistant message

## Checklist

----

### DIFF
--- a/strands-py/src/strands/session/repository_session_manager.py
+++ b/strands-py/src/strands/session/repository_session_manager.py
@@ -25,11 +25,6 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-def _make_error_result(tool_use_id: str) -> ContentBlock:
-    """Create a single error toolResult content block for an interrupted tool."""
-    return generate_missing_tool_result_content([tool_use_id])[0]
-
-
 class RepositorySessionManager(SessionManager):
     """Session manager for persisting agents in a SessionRepository."""
 
@@ -303,8 +298,10 @@ class RepositorySessionManager(SessionManager):
                 continue
 
             logger.warning(
-                "Session message history has an orphaned toolUse with no toolResult. "
-                "Adding toolResult content blocks to create valid conversation."
+                "tool_use_ids=<%s>, result_ids=<%s> | session history has mismatched toolUse/toolResult pairing,"
+                " rebuilding",
+                tool_use_ids,
+                list(existing_results.keys()),
             )
 
             # Ensure a toolResult slot exists after this assistant message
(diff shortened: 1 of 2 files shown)


strands-py/src/strands/session/repository_session_manager.py
```diff
@@ -25,6 +25,11 @@
 logger = logging.getLogger(__name__)
 
 
+def _make_error_result(tool_use_id: str) -> ContentBlock:
+    """Create a single error toolResult content block for an interrupted tool."""
+    return generate_missing_tool_result_content([tool_use_id])[0]
```
mehtarac: since this function `_make_error_result` is only making a function and returning the result of `generate_missing_tool_result_content`, can we inline the call to `generate_missing_tool_result_content` inside `_fix_broken_tool_use`?

strands-py/src/strands/session/repository_session_manager.py
```diff
@@ -271,48 +273,50 @@ def _fix_broken_tool_use(self, messages: list[Message]) -> list[Message]:
...
+                    existing_results[block["toolResult"]["toolUseId"]] = block
+                else:
+                    non_tool_result_content.append(block)
+
+            if set(existing_results.keys()) == set(tool_use_ids):
                 continue
 
             logger.warning(
                 "Session message history has an orphaned toolUse with no toolResult. "
                 "Adding toolResult content blocks to create valid conversation."
```
mehtarac: Now that we're handling stale results as well not just missing ones, we can slightly tweak this warning message?

```
logger.warning("tool_use_ids=<%s>, result_ids=<%s> | session history has mismatched toolUse/toolResult pairing, rebuilding", tool_use_ids, list(existing_results.keys()),)
```

mehtarac (approved):

PR #3123, opened 2026-07-08
fix(memory)!: rename LocalMemoryStore to TestMemoryStore

opieter-aws:
## Description

`LocalMemoryStore` was named for *where* it runs rather than *what it is for*. It's a zero-infrastructure, JSON-file-backed store meant for prototyping, offline use, and tests — not a production backend (each write rewrites the whole file; recall is lexical, not semantic). Renaming it to `TestMemoryStore` signals that intended role at the call site and reads more clearly next to the production-grade `BedrockKnowledgeBaseStore`.

`LocalMemoryStore` shipped only in the most recent release, so rather than carry a deprecation shim we make this a **clean breaking change**: the old names and the old import path are removed outright, with no aliases. This keeps the surface honest — there's no deprecated subclass, no `__getattr__` warning branch, and no lingering `local` path to migrate off later.

The rename is applied across both SDKs in lockstep for cross-SDK parity. The subsystem directory is also renamed `local` → `test_memory_store` (Python) / `test-memory-store` (TS) so the import path matches the class, mirroring the sibling `bedrock_knowledge_base/` convention. The on-disk JSON format and all record field names are unchanged, so a file written by the previous release still loads.

## Public API Changes (breaking)

Renamed symbols (both SDKs): `LocalMemoryStore` → `TestMemoryStore`, `LocalMemoryStoreConfig` → `TestMemoryStoreConfig`, `LocalMemoryAddResult` → `TestMemoryAddResult`. The import path changes with the directory rename. The old names and the old path are gone.

```python
# Before
from strands.vended_memory_stores.local import LocalMemoryStore
store = LocalMemoryStore(name="notes")

# After
from strands.vended_memory_stores.test_memory_store import TestMemoryStore
store = TestMemoryStore(name="notes")

# Also re-exported from the parent package:
from strands.vended_memory_stores import TestMemoryStore
```

```typescript
// Before
import { LocalMemoryStore } from '@strands-agents/sdk/vended-memory-stores/local'
const store = new LocalMemoryStore({ name: 'notes' })

// After
import { TestMemoryStore } from '@strands-agents/sdk/vended-memory-stores/test-memory-store'
const store = new TestMemoryStore({ name: 'notes' })
```

## Breaking Changes

`LocalMemoryStore` / `LocalMemoryStoreConfig` / `LocalMemoryAddResult` and the `.../vended-memory-stores/local` import path no longer exist. Because the on-disk format is unchanged, only import sites and type/constructor references need updating — no data migration.

### Migration

```python
# Python — update the import path and the symbol names
- from strands.vended_memory_stores.local import LocalMemoryStore
+ from strands.vended_memory_stores.test_memory_store import TestMemoryStore
```

```typescript
// TypeScript — update the subpath and the symbol names
- import { LocalMemoryStore } from '@strands-agents/sdk/vended-memory-stores/local'
+ import { TestMemoryStore } from '@strands-agents/sdk/vended-memory-stores/test-memory-store'
```

## Implementation Note (reviewer callout)

Because the class name starts with `Test`, pytest's default collection would try to load `TestMemoryStore` (and its sibling types) as test suites. Each carries `__test__ = False` to opt out. For the `TestMemoryStoreConfig` TypedDict this is a post-class assignment with a scoped `# type: ignore[attr-defined]` — mypy strict rejects `__test__` both in a TypedDict body (`[misc]`) and post-class (`[attr-defined]`), so the suppression is load-bearing and correctly scoped. A regression test asserts `__test__ is False` on all three definitions so a future edit can't silently re-enable collection.

## Documentation PR

No documentation changes are needed: the store is documented via docstrings/TSDoc in the source, which are updated in this PR. The `team/designs/0014-storage.md` RFC references the old name in a few places; it is a point-in-time design doc and is intentionally left untouched to keep this change focused.

## Type of Change

Breaking change

## Testing

Ran the affected unit suites in both SDKs (`hatch test tests/strands/vended_memory_stores/` and `vitest run src/vended-memory-stores/test-memory-store/`) plus lint/type checks on the changed files, and re-ran the Python suite with `PytestCollectionWarning` escalated to errors to confirm the `Test*` classes are never collected. Verified the breaking change end to end from a REPL script: the new name and path import silently, while both the old `LocalMemoryStore` name and the old `.local` import path now raise `ImportError`/`ModuleNotFoundError`.

## Checklist

----


JackYPCOnline:
Question: Using data format for naming makes me wondering that will we have more storages for other data format ?

opieter-aws:
@JackYPCOnline thanks for your comment, it made me check the notes and I realized we'd actually decided to name this `TestMemoryStore` in the end, to make a stronger signal to devs to use this for testing only. Preparing new rev

lizradway:
Maybe i missed some conversation on this but: 
Do you think we need to maintain the previous exports as well? This was just released today right? Could we just patch this?

opieter-aws:
> This was just released today right

Just checked and you're right, it got released today only. I don't mind doing a breaking change in that case, @pgrayy wdyt?

strands-ts/src/vended-memory-stores/test-memory-store/store.ts
pgrayy, line 1: Would recommend renaming the directory as well.

lizradway (approved):

PR #3131, opened 2026-07-08
refactor(event_loop): route message appends through Agent._append_messages

opieter-aws:
## Description

The `Agent` exposes a single append chokepoint, `_append_messages`, that assigns a durable `tracking_id` and fires `MessageAddedEvent` for every message added to history. Three sites in `event_loop.py` — the assistant model response, the tool-result message, and the cancelled-tool-result message — bypassed it and hand-replicated those steps inline. That duplication was a concern raised in #2836 .

This routes all three appends through `agent._append_messages(...)`, matching the structured-output force path that already used it. The duplicated `_ensure_tracking_id` / `append` / hook-invocation blocks collapse to one call each, and the now-unused `MessageAddedEvent` and `_ensure_tracking_id` imports are dropped. No behavior change: the same `tracking_id` assignment and `MessageAddedEvent` firing happen as before, and the `yield ToolResultMessageEvent(...)` ordering after the tool-result appends is preserved (the append is fully awaited before the yield). As a side benefit, the previously-untested cancel branch now gets the chokepoint's guarantees.

This is a Python-only change. The TypeScript SDK already routes every append through `Agent._appendMessage` / `_appendMessageAndFireHooks` and mints `trackingId` at `Message` construction rather than via an "ensure" step, so it has no equivalent duplication to consolidate.

The test fixtures needed a matching update. The event-loop tests use a `Mock` agent whose `_append_messages` was an auto-child mock that skipped the real logic; those fixtures now bind the real `Agent._append_messages` so appends assign tracking ids and fire `MessageAddedEvent` exactly as production does. The structured-output tests, which mock `_append_messages` and don't inspect `agent.messages`, switch from `assert_called_once()` to asserting the force prompt appears among the appended messages, since the assistant-response site now flows through the same mock too.

## Related Issues

Follow-up to #2836

## Documentation PR

No documentation changes needed — internal refactor with no public API or behavior change.

## Type of Change

Other: Internal refactor — no behavior or public API change.

## Testing

## Checklist

----

### DIFF
--- a/strands-py/src/strands/event_loop/event_loop.py
+++ b/strands-py/src/strands/event_loop/event_loop.py
@@ -18,7 +18,7 @@ from opentelemetry import trace as trace_api
 
 from .._middleware.stages import InvokeModelContext, InvokeModelStage
 from ..experimental.checkpoint import Checkpoint, CheckpointPosition
-from ..hooks import AfterModelCallEvent, BeforeModelCallEvent, MessageAddedEvent
+from ..hooks import AfterModelCallEvent, BeforeModelCallEvent
 from ..telemetry.metrics import Trace
 from ..telemetry.tracer import Tracer, get_tracer
 from ..tools._validator import validate_and_prepare_tools
@@ -36,7 +36,7 @@ from ..types._events import (
     TypedEvent,
 )
 from ..types.agent import Limits
-from ..types.content import Message, Messages, _ensure_tracking_id, split_system_prompt
+from ..types.content import Message, Messages, split_system_prompt
 from ..types.event_loop import Metrics, Usage
 from ..types.exceptions import (
     ContextWindowOverflowException,
@@ -637,9 +637,7 @@ async def _handle_model_execution(
         stream_trace.end()
 
         # Add the response message to the conversation
-        _ensure_tracking_id(message)
-        agent.messages.append(message)
-        await agent.hooks.invoke_callbacks_async(MessageAddedEvent(agent=agent, message=message))
+        await agent._append_messages(message)
 
         # Update metrics
         agent.event_loop_metrics.update_usage(usage)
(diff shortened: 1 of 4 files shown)


strands-py/src/strands/event_loop/event_loop.py
zastrowm, line 1: Let's work on making PR descriptions more concise.  This is much more readable, gets rid of the "what" (which is code that we're going to read anyways) and gets more to the why:

> `Agent._append_messages` is the standard way for adding to history: it
assigns a durable `tracking_id` and fires `MessageAddedEvent`. Three appends in
`event_loop.py` bypassed it and hand-rolled those steps inline, a duplication
flagged in #2836. This routes them through `_append_messages` instead.
> 
> No behavior change: the same `tracking_id` assignment, `MessageAddedEvent`
firing, and `ToolResultMessageEvent` ordering are preserved.
> 
> Python-only. The TypeScript SDK already appends exclusively through
`Agent._appendMessage` and has no equivalent duplication.

zastrowm (approved):

strands-py/src/strands/event_loop/event_loop.py
opieter-aws: Agreed there's too much bloat. I'll iterate on the pr-writer skill for more concise and 'why'-focussed PRs

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -771,9 +769,7 @@ async def _handle_tool_execution(
                 "content": [{"toolResult": result} for result in tool_results],
             }
             cancelled_tool_result_message = _cancelled_msg
-            _ensure_tracking_id(_cancelled_msg)
-            agent.messages.append(_cancelled_msg)
-            await agent.hooks.invoke_callbacks_async(MessageAddedEvent(agent=agent, message=_cancelled_msg))
+            await agent._append_messages(_cancelled_msg)
```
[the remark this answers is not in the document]
opieter-aws: This is a test that needs to be (and is) on the method called, not on the caller like here

PR #3134, opened 2026-07-09
feat(ci): publish TypeScript integ test metrics to CloudWatch

yonib05:
## Summary

- Add an `upload-metrics` job to `typescript-integration-test.yml`, mirroring the Python workflow's job, so TypeScript integ test results are published to the `Strands/Tests` CloudWatch namespace and surfaced on the flaky-test dashboard.
- Add an `SDK` dimension to `upload-integ-test-metrics.py` via an optional 3rd argument (defaults to `python`, so the Python workflow needs no change) to distinguish Python vs TypeScript metrics, since both publish under `Repository=harness-sdk`.

## Details

- The TypeScript workflow already produces JUnit XML at `strands-ts/test/.artifacts/test-report/junit/report.xml` (vitest junit reporter) and uploads it in the `test-artifacts-integ` artifact; after download the file sits at `test-report/junit/report.xml`.
- The new job uses the same `if: success() || failure()` guard as the Python workflow's `upload-metrics` job so it doesn't fail on fork PRs where the auth gate skipped the test job.
- Dashboards: a new `StrandsTypescriptIntegTests` CloudWatch dashboard (filtered on `SDK='typescript'`) has been created alongside `StrandsPythonIntegTests`, which now queries the new `SDK` dimension with the legacy 2-dimension widgets kept temporarily for historical data.

## Testing

- YAML validated.
- Script smoke-tested locally against a sample vitest JUnit report: parses vitest's `classname`/nested-suite output correctly, emits 4 metrics per test with the `SDK` dimension present, and defaults to `python` when the argument is omitted.

### DIFF
--- a/.github/scripts/upload-integ-test-metrics.py
+++ b/.github/scripts/upload-integ-test-metrics.py
@@ -128,7 +128,7 @@ def publish_metrics(metric_data: list[dict[str, Any]], region: str):
 def main():
     if len(sys.argv) not in (3, 4):
         print("Usage: python upload-integ-test-metrics.py <xml_file> <repository_name> [sdk]")
-        sys.exit(0)
+        sys.exit(1)
 
     xml_file = sys.argv[1]
     repository = sys.argv[2]
--- a/.github/scripts/upload-integ-test-metrics.py
+++ b/.github/scripts/upload-integ-test-metrics.py
@@ -65,15 +65,16 @@ def parse_junit_xml(xml_file_path: str) -> list[TestResult]:
     return results
 
 
-def build_metric_data(test_results: list[TestResult], repository: str) -> list[MetricDatum]:
+def build_metric_data(test_results: list[TestResult], repository: str, sdk: str) -> list[MetricDatum]:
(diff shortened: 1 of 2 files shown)


yonib05:
Addressed review feedback in 21adf870:

- **Alarms on the old 2-dimension metric**: verified via `describe-alarms` in account 430853190378 (us-east-1) — the account has zero CloudWatch alarms, so nothing depends on the old `[TestName, Repository]` dimension set. Dashboards were already migrated (new SDK-dimension widgets added, legacy widgets kept temporarily for history).
- **Missing-artifact edge case**: the download step now uses `continue-on-error` with an explicit `path`, and the publish step exits gracefully when no JUnit report is present (covers both early build failures and selective runs that pick no integ tests).
- Also hardened per additional review: the artifact now extracts into `downloaded-artifacts/` instead of the workspace root, so PR-controlled artifact contents can't overwrite the trusted script checked out from `main`; and the script's usage error now exits non-zero.

JackYPCOnline (approved):

PR #3145, opened 2026-07-09
fix(bedrock): cache system prompt under cache_config auto/anthropic

### LINKED ISSUE #3144: [BUG] CacheConfig(strategy="auto") never caches the system prompt (only last user message) — no cache reuse for repeated no-tools calls with a static system prompt
## Summary

With `BedrockModel(cache_config=CacheConfig(strategy="auto"))` and **no tools**, the auto strategy only ever writes a cache breakpoint on the **last user message** — never at the **system-prompt boundary**. For a workload of many calls that share one static system prompt but vary the final user message (per-candidate extraction, batched/structured-output loops), the large static `system` prefix is **re-written to cache on nearly every call and almost never read back**.

## Affected code

`strands-py/src/strands/models/bedrock.py` (v1.45.0)

The system prompt is assembled in `format_request` (~L277). A cachePoint is appended to `system_blocks` **only** for the **deprecated** `cache_prompt` config — there is no `cache_config`/`auto` path that adds a system breakpoint:

```python
# format_request (~L277)
system_blocks: list[SystemContentBlock] = system_prompt_content.copy() if system_prompt_content else []

# ~L280 — the ONLY code that puts a cachePoint on the system prompt, and it's deprecated
if cache_prompt := self.config.get("cache_prompt"):
    warnings.warn("cache_prompt is deprecated. Use SystemContentBlock with cachePoint instead.", ...)
    system_blocks.append({"cachePoint": {"type": cache_prompt}})
```

Under `strategy="auto"`, nothing is ever added to `system_blocks`; the only cache point emitted is on the last user message (in `_inject_cache_point`, ~L432). So with no tools, `auto` caches only the varying last user message and the static system prompt is never cached.

## Why this is a bug (not just suboptimal)

Anthropic prompt caching is **prefix-based** and a cache **read** requires the request prefix to match a **previously written breakpoint boundary** ([docs](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching)).

N calls, no tools, identical `system` prefix, different final user message (Mᵢ):

- Call 1: auto writes a breakpoint at end of `system + M₁` → cache boundary = end of M₁.
- Call 2: request is `system + M₂`. The only cached boundary is end of M₁, which is **not** a prefix of call 2 (M₁ ≠ M₂). → **no read hit**; full reprocess; new write at end of M₂.
- … every call writes, none read the shared static system prefix.

A breakpoint at the **system boundary** caches a block ending at `system`, which **is** a shared prefix of every subsequent call → read hit regardless of the varying user message. In our workload, adding an explicit system-prompt cachePoint took a measured Bedrock run from mostly cache-writes to a **~94% cache-read ratio** (Claude Opus 4.8, 5 extraction calls: 41,406 cache-read vs 2,653 cache-write vs 15 uncached input tokens).

## Suggested fix

In auto mode, append `{"cachePoint": {"type": "default"}}` to `system_blocks` in `format_request` so the static system prefix is cached at a boundary repeated calls can hit — not just the varying last user message. Anthropic supports up to 4 breakpoints, so a system breakpoint coexists with the last-user-message one.

## Current workaround

Pass the system prompt as content blocks with an explicit cachePoint:

```python
system_prompt = [
    {"text": manifest.system_prompt_text()},
    {"cachePoint": {"type": "default"}},
]
BedrockModel(..., cache_config=CacheConfig(strategy="auto"))
```

Works, but every caller wanting system-prompt caching under `auto` has to hand-build the system blocks — `auto` should handle the static prefix itself.

## Environment

- `strands-agents` 1.45.0
- Bedrock, Claude Opus 4.8 (`au.anthropic.claude-opus-4-8`), structured-output agents (no tools)

jsamuel1:
## Summary

Fixes the System Prompt CachePoint bug (#3144). `CacheConfig` auto/anthropic only injected a `cachePoint` on the **last user message**, never at the **system-prompt boundary**. For repeated calls that share one static system prompt but vary the final user message (e.g. no-tools structured-output loops), the static system prefix was re-written to cache on almost every call and never read back — defeating the point of prompt caching.

## Changes

- Append a `cachePoint` at the system boundary in `format_request` so the static prefix is cached where repeated calls can read-hit.
- Honours `"auto"`: the system cachePoint is only added when the model supports automatic caching (resolves to `"anthropic"` for Claude/Anthropic models), skipped otherwise.
- Existing system cachePoints are not duplicated; the `cache_config` TTL is propagated.
- Strategy resolution and cache-point construction factored into shared helpers reused by the message-injection path.

## Testing

- `tests/strands/models/test_bedrock.py` — 181 passed (added coverage for the system-prompt cachePoint path).
- `ruff check src/strands/models/bedrock.py` — clean.

Fixes #3144

### DIFF
--- a/strands-py/src/strands/models/bedrock.py
+++ b/strands-py/src/strands/models/bedrock.py
@@ -227,6 +227,36 @@ class BedrockModel(Model):
             return "anthropic"
         return None
 
+    def _resolve_cache_strategy(self) -> str | None:
+        """Resolve the effective cache strategy from ``cache_config``.
+
+        Applies the ``"auto"`` model-support check: an ``"auto"`` strategy resolves to the model's
+        supported strategy (``"anthropic"`` for Claude/Anthropic models) or ``None`` if the model does
+        not support automatic caching. An explicit ``"anthropic"`` strategy is returned as-is.
+
+        Returns:
+            The effective strategy to apply (e.g. ``"anthropic"``), or ``None`` if caching is
+            disabled or unsupported for this model.
+        """
+        cache_config = self.config.get("cache_config")
+        if not cache_config:
+            return None
+        if cache_config.strategy == "auto":
+            return self._cache_strategy
+        return cache_config.strategy
+
+    def _build_cache_point(self) -> dict[str, Any]:
+        """Build a default cache point block, including the ``cache_config`` TTL if set.
+
+        Returns:
+            A cachePoint block, e.g. ``{"cachePoint": {"type": "default"}}``.
+        """
+        cache_point: dict[str, Any] = {"type": "default"}
+        cache_config = self.config.get("cache_config")
+        if cache_config and cache_config.ttl:
+            cache_point["ttl"] = cache_config.ttl
+        return {"cachePoint": cache_point}
+
     @override
     def update_config(self, **model_config: Unpack[BedrockConfig]) -> None:  # type: ignore
         """Update the Bedrock Model configuration with the provided arguments.
@@ -282,6 +312,12 @@ class BedrockModel(Model):
                 "cache_prompt is deprecated. Use SystemContentBlock with cachePoint instead.", UserWarning, stacklevel=3
             )
             system_blocks.append({"cachePoint": {"type": cache_prompt}})
+        # Under cache_config, add a cachePoint at the system-prompt boundary so the static system
+        # prefix is cached at a boundary that repeated calls (which vary only the last user message)
+        # can read-hit. "auto" only adds this when the model supports automatic caching (see #3144).
+        elif system_blocks and self._resolve_cache_strategy() == "anthropic":
+            if not any("cachePoint" in block for block in system_blocks):
+                system_blocks.append(cast(SystemContentBlock, self._build_cache_point()))
 
         return {
             "modelId": self.config["model_id"],
@@ -425,11 +461,7 @@ class BedrockModel(Model):
                 last_user_idx = msg_idx
 
         if last_user_idx is not None and messages[last_user_idx].get("content"):
-            cache_point: dict[str, Any] = {"type": "default"}
-            cache_config = self.config.get("cache_config")
-            if cache_config and cache_config.ttl:
-                cache_point["ttl"] = cache_config.ttl
-            messages[last_user_idx]["content"].append({"cachePoint": cache_point})
+            messages[last_user_idx]["content"].append(self._build_cache_point())
             logger.debug("msg_idx=<%s> | added cache point to last user message", last_user_idx)
 
     def _find_last_user_text_message_index(self, messages: Messages) -> int | None:
@@ -529,14 +561,12 @@ class BedrockModel(Model):
         # Inject cache point into cleaned_messages (not original messages) if cache_config is set
         cache_config = self.config.get("cache_config")
         if cache_config:
-            strategy: str | None = cache_config.strategy
-            if strategy == "auto":
-                strategy = self._cache_strategy
-                if not strategy:
-                    logger.warning(
-                        "model_id=<%s> | cache_config is enabled but this model does not support automatic caching",
-                        self.config.get("model_id"),
-                    )
+            strategy = self._resolve_cache_strategy()
+            if cache_config.strategy == "auto" and not strategy:
+                logger.warning(
+                    "model_id=<%s> | cache_config is enabled but this model does not support automatic caching",
+                    self.config.get("model_id"),
+                )
             if strategy == "anthropic":
                 self._inject_cache_point(cleaned_messages)
 
--- a/strands-py/tests/strands/models/test_bedrock.py
+++ b/strands-py/tests/strands/models/test_bedrock.py
@@ -3209,6 +3209,99 @@ def test_inject_cache_point_auto_strategy_resolves_to_anthropic_for_claude(bedro
     assert len(formatted[1]["content"]) == 1
 
 
+def test_format_request_system_cache_point_auto_claude(bedrock_client, messages):
+    """Test that cache_config auto adds a system-prompt cachePoint for a supported (Claude) model."""
+    model = BedrockModel(
+        model_id="us.anthropic.claude-sonnet-4-20250514-v1:0", cache_config=CacheConfig(strategy="auto")
+    )
+    system_prompt_content = [{"text": "You are a helpful assistant."}]
+
+    request = model.format_request(messages, system_prompt_content=system_prompt_content)
+
+    assert request["system"] == [
+        {"text": "You are a helpful assistant."},
+        {"cachePoint": {"type": "default"}},
+    ]
+
+
+def test_format_request_system_cache_point_no_tools(bedrock_client):
+    """Test that the system cachePoint is added even with no tools (the #3144 no-tools bug)."""
+    model = BedrockModel(
+        model_id="us.anthropic.claude-sonnet-4-20250514-v1:0", cache_config=CacheConfig(strategy="auto")
+    )
+    messages = [{"role": "user", "content": [{"text": "extract this candidate"}]}]
+    system_prompt_content = [{"text": "static system prompt"}]
+
+    request = model.format_request(messages, system_prompt_content=system_prompt_content)
+
+    assert "toolConfig" not in request
+    assert request["system"] == [
+        {"text": "static system prompt"},
+        {"cachePoint": {"type": "default"}},
+    ]
+
+
+def test_format_request_system_cache_point_anthropic_strategy(bedrock_client, messages):
+    """Test that explicit anthropic strategy adds a system cachePoint without a model-support check."""
+    model = BedrockModel(model_id="amazon.nova-pro-v1:0", cache_config=CacheConfig(strategy="anthropic"))
+    system_prompt_content = [{"text": "You are a helpful assistant."}]
+
+    request = model.format_request(messages, system_prompt_content=system_prompt_content)
+
+    assert request["system"] == [
+        {"text": "You are a helpful assistant."},
+        {"cachePoint": {"type": "default"}},
+    ]
+
+
+def test_format_request_system_cache_point_skipped_auto_non_claude(bedrock_client, messages):
+    """Test that auto strategy does NOT add a system cachePoint for a model without automatic caching."""
+    model = BedrockModel(model_id="amazon.nova-pro-v1:0", cache_config=CacheConfig(strategy="auto"))
+    system_prompt_content = [{"text": "You are a helpful assistant."}]
+
+    request = model.format_request(messages, system_prompt_content=system_prompt_content)
+
+    assert request["system"] == [{"text": "You are a helpful assistant."}]
+
+
+def test_format_request_system_cache_point_with_ttl(bedrock_client, messages):
+    """Test that the system cachePoint carries the cache_config ttl when set."""
+    model = BedrockModel(
+        model_id="us.anthropic.claude-sonnet-4-20250514-v1:0", cache_config=CacheConfig(strategy="auto", ttl="1h")
+    )
+    system_prompt_content = [{"text": "You are a helpful assistant."}]
+
+    request = model.format_request(messages, system_prompt_content=system_prompt_content)
+
+    assert request["system"] == [
+        {"text": "You are a helpful assistant."},
+        {"cachePoint": {"type": "default", "ttl": "1h"}},
+    ]
+
+
+def test_format_request_system_cache_point_not_duplicated(bedrock_client, messages):
+    """Test that an existing system cachePoint is not duplicated under cache_config auto."""
+    model = BedrockModel(
+        model_id="us.anthropic.claude-sonnet-4-20250514-v1:0", cache_config=CacheConfig(strategy="auto")
+    )
+    system_prompt_content = [{"text": "You are a helpful assistant."}, {"cachePoint": {"type": "default"}}]
+
+    request = model.format_request(messages, system_prompt_content=system_prompt_content)
+
+    assert request["system"] == system_prompt_content
+    assert sum(1 for block in request["system"] if "cachePoint" in block) == 1
+
+
+def test_format_request_system_cache_point_skipped_without_cache_config(bedrock_client, messages):
+    """Test that no system cachePoint is added when cache_config is unset."""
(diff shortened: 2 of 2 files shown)


mehtarac:
@strandly-the-agent do a fresh review of this PR. Assess and check whether these changes proposed and the bug linked https://github.com/strands-agents/harness-sdk/issues/3144 is fixed or not in the sdk currently.

strands-py/tests/strands/models/test_bedrock.py
```diff
@@ -3209,6 +3209,99 @@ def test_inject_cache_point_auto_strategy_resolves_to_anthropic_for_claude(bedro
     assert len(formatted[1]["content"]) == 1
 
 
+def test_format_request_system_cache_point_auto_claude(bedrock_client, messages):
```
strandly-the-agent, line 3212: 🟡 Two coverage gaps here, one of them mutation-confirmed.

**The `system_blocks and` half of the guard is untested.** Mutating that truthiness check out makes `format_request` emit `system: [{"cachePoint": …}]` with nothing ahead of it — the shape Bedrock rejects with "There is nothing available to cache" — and **all 7 new tests stay green**. (The other mutation, deleting the `elif` entirely, is caught by 4 of the 7, so the positive coverage is fine; it's this branch that's unguarded.)

**Nothing asserts the two boundaries coexisting in one request** — which is the actual claim of the fix, and still the open ask in the existing review thread on this file. It passes today; it just isn't pinned.

Both tests below were written against your implementation, run, and **pass** — so these are gaps in proof, not live bugs:

<details>
<summary>Ready to paste</summary>

```python
def test_format_request_system_cache_point_skipped_for_empty_system_prompt(bedrock_client, messages):
    """Test that no system cachePoint is added when there are no system blocks to cache."""
    model = BedrockModel(
        model_id="us.anthropic.claude-sonnet-4-20250514-v1:0", cache_config=CacheConfig(strategy="auto")
    )

    request = model.format_request(messages, system_prompt_content=[])

    assert request["system"] == []


def test_format_request_system_and_last_user_cache_points_coexist(bedrock_client, messages):
    """Test that the system-prompt cachePoint and the last-user-message cachePoint are both applied."""
    model = BedrockModel(
        model_id="us.anthropic.claude-sonnet-4-20250514-v1:0", cache_config=CacheConfig(strategy="auto")
    )
    system_prompt_content = [{"text": "You are a helpful assistant."}]

    request = model.format_request(messages, system_prompt_content=system_prompt_content)

    assert request["system"] == [
        {"text": "You are a helpful assistant."},
        {"cachePoint": {"type": "default"}},
    ]
    assert request["messages"][-1]["content"] == [
        {"text": "test"},
        {"cachePoint": {"type": "default"}},
    ]
```

I also verified a `cache_tools` + system combination and a document-leading variant pass; those are less essential.

</details>

Two smaller convention points while you're in here: the siblings in this file assert the **whole** request dict (`tru_request == exp_request`, e.g. `test_format_request_cache`), so a bug that also corrupted an unrelated top-level key would slip past these narrower assertions; and per root `AGENTS.md` a regression test for a filed bug links the issue it guards — 6 of the 7 don't, and the one that does narrates the history ("the #3144 no-tools bug") rather than stating the behaviour it guarantees.

strands-py/src/strands/models/bedrock.py
```diff
@@ -282,6 +312,12 @@ def format_request(
                 "cache_prompt is deprecated. Use SystemContentBlock with cachePoint instead.", UserWarning, stacklevel=3
             )
             system_blocks.append({"cachePoint": {"type": cache_prompt}})
+        # Under cache_config, add a cachePoint at the system-prompt boundary so the static system
+        # prefix is cached at a boundary that repeated calls (which vary only the last user message)
+        # can read-hit. "auto" only adds this when the model supports automatic caching (see #3144).
+        elif system_blocks and self._resolve_cache_strategy() == "anthropic":
+            if not any("cachePoint" in block for block in system_blocks):
+                system_blocks.append(cast(SystemContentBlock, self._build_cache_point()))
```
strandly-the-agent, line 320: 🟡 This brings the system surface under `cache_config`'s management with a *different* TTL contract than its sibling message path, and nothing says so.

The message path deliberately honors a caller's **position** but normalizes their TTL — "The caller's *position* is honored; their TTL is not" (`_honor_placed_cache_point`, on current main), because Bedrock rejects an increasing TTL across `toolConfig` → system → messages. The new system path honors **both**: with a caller-placed system point at `ttl="1h"` and `cache_config(ttl="5m")`, the request comes out `system(1h), messages(5m)` — one config option, two TTL rules, and the combination that issue #2121 was about.

Not a defect, and I'm not asking you to change the behaviour — preserving what the caller wrote is defensible. But it's worth one sentence on `CacheConfig.ttl` in `models/model.py` stating it, so the next reader doesn't have to derive it from two code paths:

```suggestion
        ttl: Optional TTL duration for cache entries (e.g. "5m", "1h"). Applied to cache points the
            SDK injects automatically. A cachePoint you place in the system blocks yourself keeps its
            own TTL; one placed in the last user message is normalized to this value, because Bedrock
            requires TTLs to be non-increasing across toolConfig, system and messages.
```

(That block belongs in the `CacheConfig` docstring rather than here — putting it inline so it's next to the code that motivates it.)

strandly-the-agent (changes requested):
Fresh review at `48fc84e2`, six passes (correctness, adversarial with live Bedrock, API design, issue-alignment, test quality, docs).

**Is #3144 fixed in the SDK today? No.** On `main` @ `5f357aef7` I ran `format_request` with `CacheConfig(strategy="auto")`, a Claude model id and a static system prompt: `system` comes back with no `cachePoint`, twice in a row with different user messages. The gap is exactly as reported.

**Does this PR fix it? Yes, measurably.** Rebased onto current main and run against live Bedrock (`claude-sonnet-4-5`, us-west-2, ~4.2k-token static system prompt, 5 calls varying only the last user message): main 0 cache-read / 21,247 cache-write tokens (0% read) → PR 16,956 / 4,286 (**79.8% read**). With tools present and `cache_tools` unset: 0% → 67.1%.

**Three things stand between it and merge.** No code-correctness blocker survived the reachability gate — these are all process:

1. **Rebase.** The head is a month old and conflicts with #3677 in `_inject_cache_point`. Keep main's version of that function and drop this PR's one-line refactor of it; resolved that way, 207/207 pass. Taking the PR's side of the conflict would regress #3677's document-adjacency fix — reproduced on the raw head, a trailing non-PDF document yields `[text, document, cachePoint]`, the shape Bedrock rejects.
2. **Docs.** `site/…/amazon-bedrock.mdx:644` still says the SDK "automatically places a cache point at the end of the last user message"; after this it's incomplete. #3677 shipped its doc paragraph in the same PR, and with no hand-written changelog that paragraph *is* the release-note artifact for a default-behaviour change.
3. **`pgrayy`'s opt-out ask** (issue comment, 2026-07-15) is unanswered either way. I measured the scenario it was about and the cost premise doesn't hold (below) — but that's a decision to state, not to skip.

<details>
<summary><b>Verification ledger — what I actually ran</b></summary>

| Check | Result |
|---|---|
| #3144 reproduced on current main `5f357aef7` | ✅ no system `cachePoint` under `auto` + Claude |
| `pytest tests/strands/models/test_bedrock.py` @ `48fc84e2` | ✅ 181 passed |
| Same, rebased onto main (`eb9ce958`, my resolution) | ✅ 207 passed; whole `tests/strands/models/` 369 passed |
| `git rebase origin/main` | ⚠️ 1 conflict, `_inject_cache_point` only; the feature hunk applies cleanly |
| Live Bedrock, static system prompt, 5 calls | ✅ 0% → 79.8% cache-read ratio (0/21,247 → 16,956/4,286 tokens) |
| Live Bedrock, **dynamic** system prompt | ✅ no regression: 21,277 (main) vs 21,272 (PR) write tokens |
| Live Bedrock, multi-turn agent loop | ✅ 74.7% read ratio, identical both trees |
| Cache-point budget | ✅ PR's max is 3 of Bedrock's 4 (tools + system + last-user); 5 rejects on both trees |
| Short system prefix (below cacheable minimum) | ✅ accepted and silently ignored (HTTP 200, 0 read / 0 write) — **not** a rejection |
| Mutation test of the new guard | ⚠️ deleting `system_blocks and` escapes all 7 new tests (see inline) |

Everything below `main` vs PR was run on both trees; headline numbers re-run with fresh cache nonces and reproduced identically. Live calls were `converse`/`converse_stream` only — no AWS resources created.

</details>

<details>
<summary><b>The opt-out question, with the measurement</b></summary>

`pgrayy` asked for two things: (a) an opt-out because some users write **dynamic** system prompts, for which a cache point would be "problematic", and (b) duplicate detection so a user-placed system cachePoint isn't doubled. This PR ships (b) — `bedrock.py:319`, tested — and not (a).

On (a), the cost premise measured as a wash: with a system prompt mutated on every call, main wrote 21,277 cache-write tokens and the PR wrote 21,272. The reason is that `cache_config`'s **pre-existing** last-user-message cache point already writes the whole prefix (system included) on every call when the prefix isn't stable — so a dynamic system prompt was already paying for itself before this PR, and the new boundary adds ~nothing. A dynamic system prompt breaks the new boundary the same way it already breaks the existing one; it isn't a new failure mode.

What the measurement does *not* refute: the **budget** argument. Cache points are capped at 4 across `toolConfig`/system/messages, and after this PR a user who wants those slots elsewhere has no way to suppress the automatic system point short of dropping `cache_config` entirely and losing message caching too. That's the surviving reason to want the flag — plus the plain fact that a maintainer asked and the PR neither implements nor answers it.

If you do add it, the shape that survived review is a field on `CacheConfig` (e.g. `inject_system_cache_point: bool = True`) plus a single predicate replacing the `elif`; I prototyped exactly that on the rebased tree — 20 lines, 207 still pass, and behaviour comes out as designed (`False` → messages only; `False` + a caller-placed system point → the caller's point survives). A `surfaces={…}` set is worse: a `"tools"` member would duplicate what `cache_tools` already owns. Per-surface TTLs are the right eventual answer but need a validator, so `team/designs/` rather than this PR — note that a single shared `ttl` is currently the only assignment that *cannot* violate Bedrock's non-increasing rule between the system and message points. If the flag lands, it wants `api/needs-review`; on the current diff it doesn't (`API_BAR_RAISING.md:20-21`, and #3677 carried no api label for a larger behaviour change).

</details>

<details>

- ⚪ **Narrow new rejection, and it's arguably user error.** `cache_tools=CacheToolsConfig(ttl="5m")` + `cache_config=CacheConfig(ttl="1h")` + a last user message that *starts* with a non-PDF document (so `_inject_cache_point` declines to place a message point): main emits tools/sys/msg = 1/0/0 and is accepted; the PR emits 1/1/0 and Bedrock returns `ValidationException: system.0.cache_control.ttl: a ttl='1h' cache_control block must not come after a ttl='5m' cache_control block`. Live-reproduced twice. Gated to appendix because the same config already fails on main for any ordinary last-user message, the fix is to align the TTLs (which is exactly what #2121 was closed as resolved by), and the failure is a loud `ValidationException`, not a silent wrong answer.
- ⚪ **Dedup is "contains a cachePoint", not "ends with one"** (`bedrock.py:319`). A user following the documented manual workaround ends up with the point mid-list once something appends a system block after it (`vended_plugins/skills/agent_skills.py:228` appends one per invocation), and the guard then declines to add a boundary point — so the tail stays outside the cached prefix and the fix quietly does nothing for that composition. Not a regression (main caches no system prefix at all). Worth deciding deliberately.
- ⚪ `_build_cache_point()` has exactly **one** call site after the required rebase — main's post-#3677 `_inject_cache_point` builds its dict inline again — so the extraction that justified it is gone. Consider inlining it and keeping only `_resolve_cache_strategy()`, which does still de-duplicate real logic.
- ⚪ The comment at `bedrock.py:315-317` ends "(see #3144)". Root `AGENTS.md` allows the issue link for regression *tests*; in production source that belongs in the PR description. The rest of the comment is evergreen — drop just the clause.
- ⚪ `_resolve_cache_strategy()` inherits `_cache_strategy`'s substring check, so opaque application-inference-profile ARNs get no system cache point under `auto`. Pre-existing and already documented (use `strategy="anthropic"`), just worth knowing the new path shares it.
- ⚪ No TypeScript change, and no tracking issue exists for it — after this merges, `'auto'` means `{system, messages}` in Python and `{tools, messages}` in TS (`strands-ts/src/models/model.ts:73-74`), which is a forked concept for the same public string value (root `AGENTS.md` cross-SDK parity). #3677 shipped both SDKs together.

Also checked and clean: no new Bedrock-constraint violation from the feature itself; `cache_prompt` mutual exclusion via the `elif` is correct; the "unsupported model" warning fires under identical conditions before and after the refactor; `format_request` doesn't mutate the caller's `system_prompt_content`; `structured_output` and a full `Agent().invoke_async` both work live with the new point. Not verifiable here: whether `CountTokens` accepts a system cachePoint — every model I can invoke rejects `CountTokens` outright, so that path is unproven either way.

One process note on my own work: the API-design pass ran warm (no fresh-context spawn available in its environment) and made no live calls; every number in this review comes from the passes that did.

</details>

## Questions

**Blocking:**
- Given the measured wash on cost but the live budget argument, do you want the opt-out in this PR, or is "measured, cost is a wash, documented" enough to close `pgrayy`'s ask? Either answer unblocks — silence doesn't.
- For the rebase, do you agree the resolution should keep main's `_inject_cache_point` and drop this PR's refactor of it, rather than the other way round?

**Non-blocking:**
- Should the dedup guard mean "system blocks *end* with a cachePoint" rather than "contain one anywhere"?
- Worth filing the TS-parity follow-up before this merges, so `'auto'` doesn't quietly mean two different things across the SDKs?
- Does #3573 (Anthropic native automatic caching for `strategy="auto"`) change the direction here? Reading it, no — it says automatic caching isn't available on Bedrock, so `BedrockModel` keeps client-side injection and this PR is still the right shape for it. Flagging in case you see it differently.

Worth noting what didn't break: the two objections I most expected to sink this — a short system prefix causing a `ValidationException`, and dynamic system prompts paying extra writes — both failed to reproduce against the real API. The fix looks right; it's the branch state and the paperwork that need work. Solid work for a human to approve, not a gate.

mehtarac (commented):
Would you be able to resolve the merge conflicts?

opieter-aws:
Thanks for your contribution! Closing this PR in favor of https://github.com/strands-agents/harness-sdk/pull/3681 , which includes TS and we're actively trying to merge right now. We'll give you credits of course!

PR #3176, opened 2026-07-10
fix(file-editor): reliable str_replace — preserve bytes, real multi-match line numbers, tolerant fallback

### LINKED ISSUE #3175: fileEditor str_replace: exact-only matching (no CRLF/whitespace tolerance) + silent whole-file tab->space rewrite + broken multi-match line report cause frequent failures
## Summary
The `fileEditor` vended tool's `str_replace` command matches `old_str` as an **exact substring** with only a tab->8-space transformation applied. It has no tolerance for line-ending (CRLF vs LF), trailing-whitespace, or indentation differences, and its zero-match error is non-diagnostic. In agentic use this fails often, forcing the model to abandon the edit and rewrite the entire file via shell `cat`/heredoc — which massively inflates input tokens and cost. There are also two outright bugs: a successful edit silently converts every tab in the whole file to spaces, and the multi-match error reports an empty line list for multi-line strings.

`strands-ts/src/vended-tools/file-editor/file-editor.ts`. The matching logic lives in the pure helper `buildStrReplaceResult` and is invoked from `handleStrReplace`.

## Root cause (with file:line)

**A. Brittle exact-only matching + non-actionable zero-match error (primary)**
- Only transformation before matching is tab->8-space expansion of BOTH file and search string: file-editor.ts:172-174. Occurrence count is a plain escaped-substring regex: file-editor.ts:176. No CRLF->LF normalization, no trailing-whitespace tolerance, no leading/trailing-newline flexibility, no fuzzy fallback.
- Zero-match throws `old_str ... did not appear verbatim` (file-editor.ts:177-179) — no near-miss, no surrounding context, no whitespace/line-ending hint, no count. The model has nothing to self-correct with, so it falls back to a full-file rewrite.

**B. Successful edit silently rewrites ALL tabs in the whole file (bug)**
- The replacement is computed from the tab-expanded whole-file content inside `buildStrReplaceResult` (file-editor.ts:172, 190) and that expanded result is written straight back via `sandbox.writeText(filePath, newContent)` in `handleStrReplace` (file-editor.ts:401). So EVERY str_replace converts every tab in the entire file to 8 spaces — including regions the edit never touched. For tab-indented files (Go, Makefiles) this is silent corruption. `view` also renders tabs as 8 spaces (makeOutput, file-editor.ts:244-245).

**C. Multi-match diagnostics broken for multi-line old_str (bug)**
- The >1-occurrence branch computes line numbers with `line.includes(expandedOldStr)` (file-editor.ts:183), always false for a multi-line old_str, so the error reports `in lines []` (file-editor.ts:185-187).

## Minimal repro
```
CASE1 CRLF file + LF old_str -> zero match; "did not appear verbatim" (no line-ending hint)
CASE2 tab-indented file, edit one line -> ALL tabs in file become 8 spaces; untouched functions detabbed
CASE3 multi-line old_str x2 -> "Multiple occurrences ... in lines []"
CASE4 trailing space -> zero match; same non-actionable error
```

## Suggested fix
1. Whitespace/line-ending-tolerant fallback match before hard-erroring.
2. Do not tab-expand the whole file; apply the edit to original bytes and preserve line endings.
3. Actionable errors (near-miss chunk/hint; fix multi-line line-number computation).
4. Offer a line-range replace mode.
5. Document uniqueness/whitespace/tab behavior in schema + README.

_Line numbers are for harness-sdk@main (moved here from the now-archived strands-agents/sdk-typescript)._

hfurkanbozkurt:
Fixes #3175 — see that issue for the full root-cause analysis and repro.

## Description

`str_replace` in the vended `file-editor` tool was unreliable enough that agents frequently abandoned it and rewrote whole files with `cat`/heredoc instead, which inflates input tokens (a real cost driver). This PR fixes three concrete defects in `strands-ts/src/vended-tools/file-editor/file-editor.ts`.

**Defect B — whole-file corruption on a *successful* edit.** The old code tab-expanded (`\t` → 8 spaces) the *entire* file before writing it back, so a one-line edit silently rewrote tabs and mangled CRLF line endings across untouched regions. The replacement is now spliced into the **original bytes** (`originalContent.slice(0, start) + new + originalContent.slice(end)`), so everything outside the matched span is byte-identical.

**Defect C — multi-match line numbers always empty for multi-line `old_str`.** The duplicate-match error computed line numbers with `line.includes(oldStr)` over single split lines, which can never contain a `\n`, so any repeated *multi-line* `old_str` reported `in lines []`. It now computes real 1-based line numbers via a char-offset → line mapping and shows the first matching line as context.

**Defect A — brittle exact-only matching + non-actionable errors.** Matching was exact-substring only (after that destructive tab-expansion), with a zero-match error that gave no hint about *why*. It now:
- tries an **exact byte match first** (fully backward-compatible);
- only when the exact match finds **zero** occurrences, falls back to a conservative, line-oriented **whitespace-tolerant** search (CRLF/LF, trailing whitespace, tab-vs-spaces) that maps back to original byte offsets;
- proceeds **only** when the tolerant pass resolves to a **single** location — an ambiguous tolerant match is reported, never guessed;
- on a genuine zero-match, emits an actionable **near-miss hint** pointing at the line that differs only in indentation/whitespace/line-endings.

Also updated the `old_str` schema `.describe()` and the README (`str_replace` section + Limitations) to document the exact-first-then-tolerant behavior and the byte-preservation guarantee.

**Deferred (follow-up):** an explicit line-range replace mode (edit by line numbers instead of matching text) was intentionally *not* built here to keep this change focused; it's a reasonable follow-up.

## Related Issues

Fixes #3175

## Type of Change

Bug fix

## Testing

All commands run against Node 22 (repo requires ≥20).

- Added 6 regression tests in `file-editor.test.node.ts` (one per defect + tolerant/ambiguous/near-miss paths):
  - tabs + CRLF preserved outside the edited region (B)
  - repeated multi-line `old_str` reports real line numbers `[1,5]`, not `[]` (C)
  - CRLF-vs-LF near-miss resolves via tolerant fallback (A)
  - trailing-whitespace near-miss resolves via tolerant fallback (A)
  - ambiguous tolerant match errors instead of guessing (A)
  - genuine zero-match returns an actionable near-miss hint (A)
- `npm test` (unit-node): **133 files, 3688 tests passed, 0 type errors**
- `npm run build`: clean
- `npm run lint`: clean
- `npm run format:check`: clean
- `npm run type-check` (src + integ): clean
- The full husky pre-commit gate (build + `test:coverage` + lint + format + type-check) passed on commit.

No new warnings introduced. Changes are confined to the `file-editor` tool.

## Checklist
- [x] I have updated the documentation accordingly (schema description + README)

----


mkmeral:
@strandly-the-agent  review pls :)

strands-ts/src/vended-tools/file-editor/file-editor.ts
```diff
@@ -158,40 +163,172 @@ function applyViewRange(
...
     )
   }
 
-  const newContent = fileContent.replace(expandedOldStr, () => expandedNewStr)
-  const replacementLine = fileContent.substring(0, fileContent.indexOf(expandedOldStr)).split('\n').length - 1
-  const insertedLines = expandedNewStr.split('\n').length
-  const originalLines = expandedOldStr.split('\n').length
-  const lineDifference = insertedLines - originalLines
+  const [firstExact] = exact
+  if (firstExact !== undefined) {
```
strandly-the-agent, line 364: 🔴 **Block — reject view-equivalent ambiguity before taking the exact path.**

`view` renders tabs as eight spaces, but this branch accepts one byte-exact occurrence without checking for other occurrences that render the same way. I reproduced this with:

```text
file:    "\tTARGET\n        TARGET\n"
old_str: "        TARGET"
actual:  the space-indented line is edited; the tabbed line is untouched
```

Both lines are indistinguishable in `view`, so the model cannot reliably select the intended one; before this PR, tab expansion made this case ambiguous and it was rejected. Could we either reject when normalization finds another view-equivalent candidate, or make tabs reversible/distinct in `view`?

strands-ts/src/vended-tools/file-editor/file-editor.ts
```diff
@@ -22,7 +22,12 @@ const fileEditorInputSchema = z.object({
     .tuple([z.number(), z.number()])
     .optional()
     .describe('Line range to view [start, end]. 1-indexed. End can be -1 for end of file.'),
-  old_str: z.string().optional().describe('Exact string to find and replace (required for str_replace command).'),
+  old_str: z
+    .string()
```
strandly-the-agent, line 26: 🔴 **Block — reject an empty search string.**

The schema accepts `old_str: ""`. Exact matching returns no candidate, but the tolerant path treats it as an empty line. Through the built tool, this succeeded:

```text
before:  "alpha\n"
old_str: ""
new_str: "APPENDED"
after:   "alpha\nAPPENDED"
```

That turns `str_replace` into a content-dependent insertion instead of failing safely. The schema-side fix is straightforward:

```suggestion
    .string()
    .min(1, 'old_str must not be empty')
```

Please also add a **separate runtime guard** in `handleStrReplace`/before matching, so direct or future call paths cannot bypass the schema and reach the empty-line tolerant match.

strands-ts/src/vended-tools/file-editor/file-editor.ts
```diff
@@ -158,40 +163,172 @@ function applyViewRange(
...
+  const records = splitLineRecords(content)
+  const normOrig = records.map((r) => normalizeLine(r.raw))
+  const normOld = oldStr.replace(/\r\n/g, '\n').replace(/\r/g, '\n').split('\n').map(normalizeLine)
+  const k = normOld.length
+  if (k === 0 || k > normOrig.length) return { kind: 'none' }
+
+  const starts: number[] = []
+  const lastStart = normOrig.length - k
+  for (let i = 0; i <= lastStart; i++) {
+    if (normOrig.slice(i, i + k).every((line, j) => line === normOld[j])) starts.push(i)
```
strandly-the-agent: 🔴 **Block — bound or linearize this search.**

This slices and compares a `k`-line window at every candidate start, giving quadratic worst-case work on the Node event loop. With the actual tool, a 65,536-byte file and a 32,767-byte near-match differing only at the final line took **5,429 ms** synchronously before returning no-match; the configured file limit is 1 MiB.

Please use a linear sequence search (for example KMP over normalized lines) or enforce a bounded work limit, with an adversarial regression test.

strands-ts/src/vended-tools/file-editor/file-editor.ts
```diff
@@ -158,40 +163,172 @@ function applyViewRange(
...
+    start = tolerant.start
+    end = tolerant.end
+    resolvedViaTolerant = true
+  }
+
+  // When the tolerant fallback matched a CRLF region but new_str uses LF (the
+  // common case — models emit LF), splice in CRLF so the edited span keeps the
+  // matched region's convention instead of leaving a mixed-ending region.
+  if (resolvedViaTolerant) {
+    const matchedRegionIsCRLF = originalContent.slice(start, end).includes('\r\n')
```
strandly-the-agent: 🟡 **Should-fix — include the containing line terminator when choosing CRLF.**

This is narrower than the resolved multiline finding: a single-line tolerant match ends before its adjacent terminator, so this slice contains no `\r\n`. Reproduced through the built tool:

```text
before:  "\told\r\nkeep\r\n"
old_str: "        old"
new_str: "first\nsecond"
after:   "first\nsecond\r\nkeep\r\n"
```

Could we derive the convention from the matched line record (or inspect the delimiter adjacent to `end`) and add this single-line boundary case as a byte-exact test?

strandly-the-agent (changes requested):
**Changes requested** — three reproduced blockers remain in exact/tolerant matching, plus one distinct single-line CRLF boundary bug.

✅ Reviewed `e48aee4` against base `8658b9e` · ✅ 61/61 focused tests · ✅ build/type-check/lint/format/diff-check · ✅ four built-tool repros confirmed

🔴 Exact-first can edit the wrong view-equivalent occurrence; empty `old_str` can write; tolerant matching can block Node for seconds.  
🟡 Single-line tolerant CRLF matches can still introduce bare LF.  
⚠️ GitHub CI Gate had not reported at review time; integration/review-trigger checks were waiting and non-required.

<details>
<summary>Specialist-pass breakdown and verification</summary>

- **Context build:** traced the shared default/Docker/SSH implementation, schema-to-model path, callers, tests, and the resolved `72b13449` multiline CRLF fix.
- **Correctness/safety:** changes requested — confirmed quadratic tolerant search and the residual single-line CRLF boundary defect.
- **Adversarial/repro:** changes requested — cold built-tool repros confirmed wrong-location view-equivalent edits, empty-string mutation, event-loop blocking, and the CRLF boundary case.
- **Test quality:** changes requested — existing 61 tests pass, but they do not cover the reproduced empty-string, view-equivalence, performance, or single-line CRLF cases.
- **LLM context:** changes requested — `view` hides tab/space distinctions while exact matching uses those hidden bytes, so the model can select the wrong rendered occurrence.
- **Docs accuracy:** the CRLF guarantee is broader than the single-line implementation; folded into the CRLF finding rather than posted separately.
- **Aggregation:** 20 raw observations were deduplicated to four evidence-backed findings; lower-value safe-failure and documentation-only items were suppressed.

`npm ci`: 410 packages, 0 vulnerabilities. Focused unit-node: 61/61, no type errors. Serial build/type-check, changed-file ESLint/Prettier, and `git diff --check` passed. Repros exercised built `makeFileEditor` through `TestSandbox`. No public API signature/type/export changed, so no API-review label appears necessary.
</details>

strands-ts/src/vended-tools/file-editor/file-editor.ts
hfurkanbozkurt: Good catch — this violated our own "never guess on ambiguity" rule. The exact branch now runs the view-normalized scan too and rejects as ambiguous when another view-equivalent candidate exists. Test added. Fixed in fbd8319a.

strands-ts/src/vended-tools/file-editor/file-editor.ts
hfurkanbozkurt: Agreed, failing safe is correct. Added `.min(1)` on the schema and a runtime guard before matching so direct/future call paths can't bypass it into the empty-line tolerant match. Test added. Fixed in fbd8319a.

strands-ts/src/vended-tools/file-editor/file-editor.ts
hfurkanbozkurt: Fixed — linearized the normalized-line search (in-place compare with early-exit, no per-candidate slice allocation), so the 64KiB/32KiB near-miss no longer blocks. Added an adversarial perf regression test. Fixed in fbd8319a.

strands-ts/src/vended-tools/file-editor/file-editor.ts
hfurkanbozkurt: Fixed — CRLF is now derived from the matched line record's terminator instead of the terminator-less slice, so single-line tolerant matches preserve CRLF. Added the byte-exact single-line boundary test. Fixed in fbd8319a.

hfurkanbozkurt:
Thanks for the thorough pass @strandly-the-agent — reproduced all four and they're legit. Fixed in fbd8319a: S1 view-equivalence ambiguity now rejected on the exact path; S2 empty old_str guarded (schema .min(1) + runtime guard); S3 tolerant scan linearized (no more quadratic event-loop block) + perf test; S4 single-line CRLF derived from the matched line record's terminator. Tests added for all four reproduced cases. PTAL.

hfurkanbozkurt:
@strandly-the-agent review again?

mkmeral:
@strandly-the-agent  pretty pls? (check prev comment)

strands-ts/src/vended-tools/file-editor/__tests__/file-editor.test.node.ts
```diff
@@ -361,6 +361,191 @@ describe('fileEditor tool', () => {
...
+      expect(err).toBeInstanceOf(Error)
+      expect(err?.message).toContain('did not appear')
+      expect(err?.message).toMatch(/similar line was found at line 2/)
+    })
+
+    // S1: a single byte-exact hit that is view-equivalent to another line must NOT be
+    // silently edited. `view` renders a tab as 8 spaces, so a tab-indented line and an
+    // 8-space-indented line look identical; rejecting as ambiguous honors "never guess".
+    it('rejects an exact match that is view-equivalent to another line (does not silently edit)', async () => {
+      const content = '\tTARGET\n        TARGET\n'
```
strandly-the-agent, line 484: 🔴 **Block — the ambiguity guard still misses partial-line targets.**

Exact matching accepts substrings, but `tolerantMatch` compares normalized whole lines. With:

```text
\tTARGET suffix
        TARGET suffix
```

and `old_str: "        TARGET"`, the built tool succeeds and changes only the spaces occurrence to `X suffix`, even though both target substrings render identically through `view`.

The matcher needs span-aware normalized substring matching. This fixture change makes the existing rejection and unchanged-content assertions cover the residual case:

```suggestion
      const content = '\tTARGET suffix\n        TARGET suffix\n'
```

strands-ts/src/vended-tools/file-editor/__tests__/file-editor.test.node.ts
```diff
@@ -361,6 +361,191 @@ describe('fileEditor tool', () => {
...
+      const unchanged = await fs.readFile(filePath, 'utf-8')
+      expect(unchanged).toBe('alpha\n')
+    })
+
+    // S3: the tolerant scan must stay linear. A large file whose first half nearly matches
+    // old_str (differing only at the final line) previously triggered an O(candidates * k)
+    // slice-per-candidate blowup that blocked the event loop for seconds. The linearized
+    // in-place scan with early-exit returns promptly.
+    it('does not block the event loop on a large near-miss (linear tolerant scan)', async () => {
+      const lines = Array.from({ length: 2048 }, (_, i) => `const v${i} = ${i}; // ${'x'.repeat(20)}`)
```
strandly-the-agent, line 511: 🔴 **Block — this fixture does not exercise the remaining quadratic scan.**

The production code still compares each candidate window with a nested loop. Because these lines are unique, almost every candidate fails on its first comparison. On repetitive inputs at this head, 32/64/128 KiB took about **0.34/1.25/4.85 s**; a 64 KiB unique exact-success call delayed a zero-delay timer armed after `readText` by **1.23 s**, because the exact branch also runs the tolerant scan.

Please use a genuinely linear matcher such as KMP/Z. A repetitive-prefix fixture will enforce the failure mode:

```suggestion
      const repeatedLine = `const repeated = 1; // ${'x'.repeat(20)}`
      const lines = Array.from({ length: 4096 }, () => repeatedLine)
```

Then keep the late final-line mismatch so each candidate shares a long prefix.

strands-ts/src/vended-tools/file-editor/file-editor.ts
```diff
@@ -158,40 +164,213 @@ function applyViewRange(
...
+  const matchStart = starts[0]
+  if (matchStart === undefined) return { kind: 'none' }
+  const first = records[matchStart]
+  const last = records[matchStart + k - 1]
+  if (!first || !last) return { kind: 'none' }
+  // Derive the CRLF convention from the matched line records' own terminators. A
+  // single-line match ends before its `\r` (contentEnd excludes it), so the
+  // sliced region can omit the `\r\n` even though the file is CRLF — the records
+  // still carry each line's trailing `\r`.
+  const crlf = records.slice(matchStart, matchStart + k).some((r) => r.raw.endsWith('\r'))
```
strandly-the-agent: 🟡 **Should-fix — inherit CRLF for an unterminated final matched line.**

For `before\r\n\told` (no final terminator), this record has no trailing `\r`, so replacing the tolerant target with `first\nsecond` produces `before\r\nfirst\nsecond`. The terminated-line case is fixed; this is the EOF boundary where the matched record has no convention of its own.

Could we fall back to the preceding line's terminator when the match ends at EOF?

```suggestion
  const crlf =
    records.slice(matchStart, matchStart + k).some((r) => r.raw.endsWith('\r')) ||
    (last.contentEnd === content.length && records[matchStart - 1]?.raw.endsWith('\r') === true)
```

strandly-the-agent (changes requested):
**Changes requested** — three historical cases are fixed, but the ambiguity guard still misses partial-line targets and the scan still blocks Node quadratically; an unterminated CRLF edge also remains.

✅ Reviewed `f1181ed4` vs base `6d1eda9` · ✅ fix/HEAD PR blobs identical · ✅ 65/65 focused tests · ✅ build/type-check/lint/format/diff-check · 🔴 3 residual built-tool repros

🔴 Partial-line view-equivalent targets can still edit the wrong occurrence.  
🔴 Repetitive inputs still produce quadratic event-loop stalls, including exact-success calls.  
🟡 A tolerant match on an unterminated final CRLF line can introduce bare LF.

<details>
<summary>Per-pass breakdown and verification</summary>

- **Context build:** mapped `e48aee4..f1181ed4`; both later main merges have empty remerge diffs and leave all three PR blobs byte-identical to `fbd8319a`.
- **Correctness:** confirmed empty `old_str`, full-line view ambiguity, and terminated single-line CRLF are fixed; retained the partial-line ambiguity and EOF-CRLF defects.
- **Adversarial/repro:** cold built-tool reruns confirmed the three retained cases. Repetitive 32/64/128 KiB inputs took ~0.34/1.25/4.85 s; a zero-delay timer armed after `readText` was delayed ~1.23 s on the 64 KiB exact-success case.
- **Test quality:** current 65 tests pass, but the ambiguity fixture only covers whole lines and the performance fixture uses unique lines that early-exit immediately; both gaps are folded into the inline findings.
- **Aggregation:** fixed/resolved threads were not repeated. Runtime-guard private coverage and dependency-audit noise were suppressed; this PR changes no dependencies or public API signatures.

`npm ci` installed 411 packages. Serial build/type-check, changed-file ESLint/Prettier, and `git diff --check` passed. GitHub integration/review-trigger checks were waiting and non-required; CI Gate had not reported at post time.
</details>

strands-ts/src/vended-tools/file-editor/__tests__/file-editor.test.node.ts
hfurkanbozkurt: Good catch — the whole-line guard couldn't see partial-line view-equivalence. Added a line-start-anchored view-equivalence check on the single-line exact path: it tab-expands file+old_str and rejects when `old_str` begins more than one rendered line. Anchoring to line starts (not a flat substring scan) avoids false-rejecting a unique target that merely appears mid-line inside a more-indented line. Your repro now rejects; a byte-and-view-unique partial-line edit still succeeds. Tests added for both. Fixed in 899c9f43.

strands-ts/src/vended-tools/file-editor/__tests__/file-editor.test.node.ts
hfurkanbozkurt: Fixed properly this time — replaced the nested scan with KMP over the normalized-line tokens (computeLps/kmpFindAll), so tolerant matching is genuinely O(n+k) and enumerates the same occurrence set (overlaps included). The repetitive-prefix near-miss that was ~seconds now returns in milliseconds; added an adversarial repetitive-line perf regression test (verified it fails on the old scan, passes on KMP). The exact-success path's ambiguity scan is now O(n) too. Fixed in 899c9f43.

strands-ts/src/vended-tools/file-editor/file-editor.ts
hfurkanbozkurt: Fixed — CRLF now falls back to the preceding line's terminator when the tolerant match lands on an unterminated final line at EOF, so a split replacement keeps CRLF. Added the EOF byte-exact test plus a matchStart===0 test confirming a lone unterminated line stays LF (no spurious CRLF). Fixed in 899c9f43.

hfurkanbozkurt:
Thanks again @strandly-the-agent — round-2 findings all addressed in 899c9f43, and it went through an internal senior+principal review before pushing: NEW-1 line-start-anchored view-equivalence guard (rejects genuine partial-line ambiguity, no longer false-rejects unique mid-line-substring targets); NEW-2 real KMP linearization of the tolerant scan (O(n+k)) + adversarial perf test; NEW-3 EOF-unterminated-line CRLF inheritance. 70 tests green (build/type-check/lint/format all pass). PTAL.

mkmeral:
@strandly-the-agent review pr

strands-ts/src/vended-tools/file-editor/__tests__/file-editor.test.node.ts
```diff
@@ -361,6 +361,277 @@ describe('fileEditor tool', () => {
...
+      const filePath = await createTestFile('view-span-ambig.txt', content)
+      await expect(
+        fileEditor.invoke({ command: 'str_replace', path: filePath, old_str: '        TARGET', new_str: 'X' }, context)
+      ).rejects.toThrow(/candidates/)
+      // Nothing was written: the file is byte-for-byte unchanged.
+      const unchanged = await fs.readFile(filePath, 'utf-8')
+      expect(unchanged).toBe(content)
+    })
+
+    it('edits a unique partial-line target even when a more-indented line contains it mid-line', async () => {
```
strandly-the-agent, line 575: 🔴 **Block — count rendered mid-line matches that begin on real source boundaries.**

The filter at `file-editor.ts:388-390` keeps only rendered occurrences at physical line starts. With:

```text
prefix\tTARGET suffix
prefix        TARGET suffix
```

both lines render identically, but `old_str: "        TARGET"` succeeds and edits only line 2 to `prefixX suffix`. The model cannot distinguish which span it selected.

The rendered scan needs source-boundary mapping: count the candidate after `prefix` above, while preserving the existing allow-control where an apparent occurrence begins *inside* a tab's eight-space expansion. This regression can be added alongside that control:

```suggestion
    it('rejects a mid-line exact match that is view-equivalent to another span', async () => {
      const content = 'prefix\tTARGET suffix\nprefix        TARGET suffix\n'
      const filePath = await createTestFile('view-span-mid-line-ambig.txt', content)
      await expect(
        fileEditor.invoke(
          { command: 'str_replace', path: filePath, old_str: '        TARGET', new_str: 'X' },
          context
        )
      ).rejects.toThrow(/candidates/)
      await expect(fs.readFile(filePath, 'utf-8')).resolves.toBe(content)
    })

    it('edits a unique partial-line target even when a more-indented line contains it mid-line', async () => {
```

strandly-the-agent (changes requested):
**Changes requested** — KMP performance, EOF CRLF inheritance, and the prior line-start ambiguity are fixed; one distinct mid-line view-ambiguity path still silently edits the wrong occurrence.

✅ Reviewed `899c9f43` vs base `6d1eda9` · ✅ 70/70 focused tests · ✅ build/type-check/lint/format/diff-check · ✅ KMP/CRLF/perf cold repros · 🔴 1 residual built-tool repro

🔴 The new rendered-match guard only counts physical line starts, so two mid-line spans that render identically can still select one hidden byte representation.

<details>
<summary>Per-pass breakdown and verification</summary>

- **Context build:** scoped only `f1181ed..899c9f43` (2 files, `+164/-18`) and mapped the new line-start guard, KMP matcher, CRLF fallback, callers, and tests.
- **Correctness:** KMP matched naive overlapping search across 1,193,920 exhaustive small cases; one residual mid-line ambiguity remained.
- **Adversarial/repro:** cold built-tool runs confirmed the prior line-start repro now rejects unchanged, EOF CRLF and lone-LF controls pass, overlaps reject at lines `[1,3]`, and repetitive 16K/32K/65K inputs took ~54/90/169 ms. The retained mid-line case reproduced twice.
- **Test quality:** the new performance and CRLF tests are effective. A coverage-only overlap mutant was suppressed after exhaustive comparison and a passing actual-tool overlap repro.
- **API/docs:** no public signature, export, or docs change in this delta; no API label needed.
- **Aggregation:** all three resolved `f1181ed` threads were left closed and not repeated. Final set: one blocker, no nits/questions.

`npm ci` installed 411 packages; audit findings are inherited from current main and this PR changes no dependencies. GitHub integration/review-trigger checks were waiting and non-required; CI Gate had not reported at post time.
</details>

lizradway:
Hi! Thanks for the contribution! Decided to go with a simpler approach to just ease the diff problem. Added you as co-author on the commit for inspiration :) https://github.com/strands-agents/harness-sdk/pull/4014

PR #3183, opened 2026-07-10
docs(pr-writer): scale description length to diff size

zastrowm:
## Description

The `pr-writer` skill was producing PR descriptions that were too long for the size of the change — a one-line diff would get three paragraphs of prose. The existing "keep it short" guidance was too abstract to bite, so it got ignored in practice.

This replaces it with a concrete length heuristic scaled to diff size (a <10-line change should be one or two sentences; more than ~3 sentences on a small diff is a padding signal) and reframes the old trim step as a reader's-POV self-review: name the single fact a reviewer can't recover from the diff, cut everything that isn't load-bearing, then check length against the diff.

Verified on three real merged PRs spanning 1 / 246 / 490 changed lines: descriptions scaled to 2 / 4 / 6 sentences respectively, and — compared against the actual merged descriptions — the shorter output kept the load-bearing motivation and invariants rather than dropping them.

## Documentation PR

No documentation changes required.

## Type of Change

Other: tweak to the pr-writer agent skill.

## Testing

Ran the updated skill in isolated sub-agents against three merged PRs of varying diff size and compared the generated descriptions to the real ones.

## Checklist

----


lizradway (approved):

lizradway:
maybe we could encourage drop down use? so additional structured information could be fit in if desired, but not overloading to read

opieter-aws (approved):

Issue #3066, opened 2026-07-01
[BUG] ContextOffloader image offload path can reconstruct an invalid image.format (same class as #3019)


agent-of-mkmeral:
## Summary

Follow-up to #3019 / #3033. The **image** offload path in the Python `ContextOffloader` has the same class of bug that #3033 fixed for documents:

- **Store:** `_handle_tool_result` fabricates the MIME type as `image/{format}` from the block's `format` field — whatever that string is.
- **Retrieve:** `_decode_full_content` reconstructs via `content_type.split("/")[-1]`:

```python
if content_type.startswith("image/"):
    img_format = content_type.split("/")[-1]
    return {
        "status": "success",
        "content": [{"image": {"format": img_format, "source": {"bytes": content_bytes}}}],
    }
```

So any `image/*` content type whose subtype is not a valid `ImageFormat` (`png`, `jpeg`, `gif`, `webp`) — e.g. `image/svg+xml`, `image/tiff`, `image/x-icon` written by another producer to the same storage, or a runtime-only unexpected format value round-tripped through store — is reconstructed as an image block with an invalid `format`, and the next model call fails with a `ValidationException`, killing the turn (same failure mode as #3019).

A secondary, related fidelity nit: `jpeg` is stored as `image/jpeg` and round-trips fine, but the mapping is still string-surgery rather than an exact lookup, so there is no guard coupling it to `ImageFormat`.

## Expected

Mirror the #3033 approach for images: a closed, bijective `ImageFormat ↔ MIME` lookup in `media_types.py` (it already exists for documents), with unknown `image/*` subtypes degrading safely (e.g. an explanatory text result) instead of emitting an invalid image block — plus a `get_args(ImageFormat)` drift-guard test.

## Notes

- Kept out of #3033 per the one-logical-change-per-PR rule; this issue tracks the follow-up (it was also flagged in review there).
- The TypeScript SDK's context-offloader plugin has the same pattern for both documents and images (`strands-ts/.../context-offloader/plugin.ts` string-parses the MIME subtype straight into the format) — the document half of the TS fix likely deserves its own item too.

xlu2025:
Hi! I'd be interested in contributing to this issue if it's still available.

From what I understand, #3033 and #3184 focus on document format round-tripping, while the Python image path is intentionally left for this follow-up issue.

My initial plan would be to:

1. Add a canonical mapping between ImageFormat and MIME types.
2. Use that mapping when storing and reconstructing images.
[...]

Issue #3019, opened 2026-06-30
ContextOffloader: retrieve_offloaded_content emits document blocks with formats Bedrock Converse rejects (e.g. octet-stream) → ValidationException kills the turn


strandly-the-agent:
## Summary

`ContextOffloader.retrieve_offloaded_content` can return a `document` content block whose `format` is **not** in the set Bedrock Converse accepts, which makes the very next `ConverseStream` call fail with a `ValidationException` and aborts the agent turn ("cycle failed").

## Where

`strands/vended_plugins/context_offloader.py`, in `retrieve_offloaded_content`:

```python
if content_type.startswith("application/"):
    doc_format = content_type.split("/")[-1]
    doc_block = {"format": doc_format, "name": reference, "source": {"bytes": content_bytes}}
    return {"status": "success", "content": [{"document": doc_block}]}
```

`doc_format` is taken verbatim from the stored MIME subtype. But Bedrock Converse's `document.format` only accepts:

```
[docx, csv, html, txt, pdf, md, doc, xlsx, xls]
```

The most common offending case is **`application/octet-stream`** — the default content type for any binary/unknown bytes — which yields `format="octet-stream"`, not in the enum.

## Repro (observed in production)

1. A tool returns a large result that gets stored by the offloader as `application/octet-stream` (e.g. file bytes read in a sandbox).
2. The model later calls `retrieve_offloaded_content` for that reference.
3. Retrieve returns `{"document": {"format": "octet-stream", "name": "artifacts/…txt", "source": {"bytes": …}}}`.
4. The next `ConverseStream` includes that tool result and fails:

```
ValidationException: 1 validation error detected: Value at
'messages.N.content.M.toolResult.content.K.document.format' failed to satisfy constraint:
Member must satisfy enum value set: [docx, csv, html, txt, pdf, md, doc, xlsx, xls]
```

The turn dies; nothing recovers it.

## Impact

Any offload→retrieve of binary/unknown content (very common — `octet-stream` is the catch-all) reliably crashes the turn. It surfaces on heavier workloads that read large/unknown-typed content (e.g. an agent fetching and reviewing files).

## Suggested fix

Clamp `document.format` to the Converse-allowed enum on retrieval — map unknown/`octet-stream` (and any non-enum subtype) to `txt`, e.g.:

```python
_CONVERSE_DOC_FORMATS = {"docx","csv","html","txt","pdf","md","doc","xlsx","xls"}
...
[...]

PR #3184, opened 2026-07-10
fix(context-offloader): preserve document format through offload round-trip (TS)

strandly-the-agent:
## Why

The TypeScript `ContextOffloader` could corrupt an offloaded document on retrieval and make text documents un-searchable — the same two failures the Python SDK just fixed in #3033 (issue #3019). This is the TypeScript counterpart, requested on that PR.

The offloader fabricated `` `application/${block.format}` `` as the stored MIME and, on retrieval, string-parsed the subtype straight back into `document.format`. Two consequences for users:

1. **Turn-aborting crash.** When a storage backend loses content-type metadata (a pruned `FileStorage` `.metadata.json` sidecar, an S3 object without `ContentType`), it falls back to `application/octet-stream`. That round-tripped into `document.format = "octet-stream"` — invalid for the Bedrock model API — so the next model call died with a `ValidationException` and aborted the turn.
2. **Text documents couldn't be searched.** A `csv`/`html`/`md`/`txt` document was stored as `application/csv` etc., which the search gate rejected as *"cannot search binary content"* — so agents couldn't `grep` an offloaded CSV.

## What changed

The SDK already ships a canonical, bijective `format ↔ MIME` contract (`toMimeType` / `toMediaFormat` in `src/mime.ts`) — the offloader just wasn't using it. It now does, so no MIME table is duplicated. A small `media-types.ts` module adds the offloader-specific pieces (legacy-artifact reconstruction, the searchable-legacy set, file extensions).

```ts
// store — before
const contentType = `application/${block.format}`   // csv -> application/csv (not searchable; not canonical)
// store — after
const contentType = mimeForDocumentFormat(block.format)  // csv -> text/csv (canonical + searchable)

// retrieve — before
format: contentType.split('/').pop() as DocumentFormat  // octet-stream -> invalid document.format
// retrieve — after
const documentFormat = documentFormatForMime(contentType)  // inverse map; unknown -> undefined -> decode as text
```

- **Lossless reconstruction** through the inverse map: a `pdf` retrieves as `pdf`, a `csv` as `csv`, etc. (`txt` retrieves as plain text — content-identical and additionally searchable — matching the Python fix).
- **Safe degradation:** unknown/foreign types (e.g. `octet-stream` from lost metadata) decode as text instead of fabricating a document block the model rejects.
- **Backward compatible:** artifacts written by earlier releases as `application/{format}` still reconstruct into document blocks (binary ones like `application/docx` too, rather than decoding to garbage), and the text-based ones stay searchable.
- `DocumentFormat` is now derived from a `DOCUMENT_FORMATS` const (mirrors the existing `IMAGE_FORMATS` pattern), so a future format can't drift from its MIME mapping.

<details>
<summary>Behavior note for custom <code>Storage</code> backends</summary>

Content stored under `text/csv` / `text/html` / `text/markdown` now reconstructs as a `DocumentBlock` on full retrieval, where it previously returned a decoded string. This is internally consistent — the offloader only emits those types from real document blocks — but a custom backend that writes those MIME types and expects a string back would see the change. `txt`/plain text is unaffected (still returned as a string).

</details>

<details>
<summary>Tests &amp; verification</summary>

- 6 new regression tests (octet-stream → text degradation, canonical-MIME store + round-trip, offloaded-CSV searchability, txt → text, legacy `application/{format}` reconstruction + search) — **all 6 fail on `main` and pass with the fix** (verified by reverting the source and re-running), plus a `media-types` unit suite (round-trip identity per format, drift guard, extensions, legacy searchable set).
- Full `strands-ts` unit suite green locally: **3,700 passed**, `tsc --noEmit` clean, `eslint` clean, `prettier --check` clean.
- Image/video reconstruction still uses `split('/')` — the same class as the tracked follow-up #3066 — and is intentionally left out of this document-scoped change (confirmed not regressed).

**Independent review:** two fresh-context review passes (general correctness + adversarial) — the first approved; the adversarial pass found no exploitable bug (residuals were intended tradeoffs or unreachable-from-offloader edge cases). Both surfaced two non-blocking test suggestions, which are applied.

</details>

Related: #3019, #3033 (Python). 🤖 Generated with agent assistance; a human should review before merge.


JackYPCOnline:
@strandly-the-agent review and be restrictive on this PR ^^

strands-ts/src/vended-plugins/context-offloader/media-types.ts
```diff
@@ -0,0 +1,114 @@
...
+ * `application/{format}` aliases. `text/plain` is intentionally excluded: it is
+ * shared with plain text blocks and retrieves as text.
+ *
+ * @param contentType - The stored MIME type.
+ * @returns The document format to reconstruct, or `undefined` if the type does
+ *   not correspond to a document block.
+ * @internal
+ */
+export function documentFormatForMime(contentType: string): DocumentFormat | undefined {
+  if (contentType === 'text/plain') return undefined
```
strandly-the-agent, line 99: 🔴 **Block: preserve byte-backed `txt` content**

Mapping `txt` to `text/plain` and excluding it here loses the fact that this was a byte-backed `DocumentBlock`; retrieval then falls through to the non-fatal UTF-8 decoder at `plugin.ts:154`.

A cold base/head repro used bytes `[102,128,103,255,104]`: base returned the same `DocumentBlock` bytes, while this head returned `f�g�h`, which re-encodes as `[102,239,191,189,103,239,191,189,104]`. That is irreversible PR-introduced corruption, not content-identical degradation.

Please preserve document/source identity and add a non-UTF-8 byte fixture proving exact preservation. The current `TextEncoder` fixture only exercises valid UTF-8.

I am **not** attaching a `suggestion` block here because there is no safe line-local replacement: the fix requires choosing a persisted representation that distinguishes a `TextBlock` from a byte-backed `txt` `DocumentBlock` (for example, retaining `application/txt` for that source kind or adding block/source-kind metadata). That compatibility choice should be made explicitly rather than encoded in a speculative one-line patch.

strandly-the-agent (commented):
🔴 **Blocking finding:** byte-backed `txt` documents can now be irreversibly transcoded. GitHub does not allow this account to request changes on its own PR, so this remains a comment review rather than `CHANGES_REQUESTED`.

✅ Exact head `fa74838e8fd89f16610931445202ae2be282d396`
✅ Cold base/head repro repeated twice with identical results
✅ Build, type-check, diff-check, and 141 focused tests passed
✅ CI Gate green; integration still waiting

- **Correctness — 🔴 block:** byte-backed `txt` loses source identity and invalid UTF-8 bytes are replaced irreversibly.
- **API / DevX — ✅ pass:** `DOCUMENT_FORMATS` is not package/barrel exported; the public `DocumentFormat` union is unchanged, so no API-review label is needed.
- **Adversarial testing — 🔴 reproduced:** base preserved `[102,128,103,255,104]`; head returned `f�g�h` and different re-encoded bytes, twice from cold worktrees. Canonical/legacy format and storage-matrix probes otherwise held.
- **Test quality — 🟡 gap folded into blocker:** the new `txt` test uses `TextEncoder`, so it cannot detect non-UTF-8 corruption. I did not repeat the existing open full-object/name assertion thread; other mutation-surviving gaps lacked a demonstrated defect.
- **Model context — 🔴 block:** replacement-character text gives the model corrupted content. Unknown/octet-stream text degradation was suppressed as the deliberate crash-avoidance tradeoff of this fix, rather than a second finding.
- **Docs accuracy — 🟡 should-fix:** the new round-trip claim omits the pre-existing JSON exception; a concrete replacement is provided inline.

The overall check failure is the expected two-approver bot gate, not a code/test failure.

</details>

strands-ts/src/vended-plugins/context-offloader/media-types.ts
```diff
@@ -0,0 +1,114 @@
...
+ * The canonical `format <-> MIME` direction is delegated to the SDK's own
+ * {@link toMimeType} / {@link toMediaFormat} contract in `mime.ts`, so no table
+ * is duplicated here. This module adds only the offloader-specific pieces:
+ * backward-compatible reconstruction of the fabricated `application/{format}`
+ * artifacts written by earlier releases, the text-searchable subset of those
+ * legacy types, and file extensions.
+ *
+ * Documents round-trip to the original block for every format except `txt`,
+ * which intentionally retrieves as plain text (content-identical, and
+ * additionally pattern-searchable).
```
strandly-the-agent, line 20: 🟡 **Should-fix: document both retrieval exceptions**

`txt` is not the only format that does not reconstruct as its original document block: `decodeStoredContent` handles `application/json` before this inverse mapping, so valid JSON returns parsed JSON rather than a `DocumentBlock`. This range replacement also avoids claiming byte-for-byte identity for arbitrary `txt` bytes:

```suggestion
 * Canonical document MIME types reconstruct as document blocks except `txt`
 * and `json`: `txt` retrieves as UTF-8-decoded text, while `json` retrieves as
 * parsed JSON (or decoded text when parsing fails).
```

strandly-the-agent (commented):
Range-corrected docs suggestion for `media-types.ts:18–20`; this supersedes the earlier single-line suggestion.

strands-ts/src/vended-plugins/context-offloader/plugin.ts
```diff
@@ -137,14 +135,22 @@ function decodeStoredContent(content: Uint8Array, contentType: string, reference
       source: { bytes: content },
     }) as unknown as JSONValue
   }
-  if (contentType.startsWith('application/')) {
-    const format = contentType.split('/').pop()!
+  // Reconstruct the original document block through the canonical inverse map
+  // (plus legacy application/{format} aliases). Content types the offloader never
+  // produces -- notably the storage backends' application/octet-stream fallback
```
JackYPCOnline, line 140: Recommand trim these inline comments

Issue #911, opened 2025-09-23
[FEATURE] Agent Handoffs/Delegation


mkmeral:
### Problem Statement

Customers have asked for a simple way for agents to delegate requests to specialized sub-agents with complete handoff. They want to simplify the developer experience for common delegation patterns:

- **Graph/Swarm**: Customers want simpler setup for hierarchical delegation use cases that don't require complex orchestration
- **Agents as Tools**: For use cases where the sub-agent's response doesn't need additional LLM processing, customers want to avoid the extra step of main agent → sub-agent → main agent processing → final response

We need **true delegation** where the orchestrator analyzes the request, delegates to a sub-agent, and the sub-agent's response becomes the final response with no additional processing.

### Proposed Solution

Introduce a `sub_agents` parameter (alternatively `delegates`, `handoff_agents` or something similar) to the `Agent` class that automatically creates delegation tools for each sub-agent. This would enable a clean, hierarchical agent architecture:
```python
# Define specialized agents
customer_service = Agent(
    name="CustomerService",
    description="Handles customer service inquiries, billing questions, and account issues",
    system_prompt="You are a customer service specialist...",
    tools=[billing_lookup, account_management]
)
technical_support = Agent(
    name="TechnicalSupport", 
    description="Provides technical support for product issues and troubleshooting",
    system_prompt="You are a technical support specialist...",
    tools=[diagnostic_tools, product_database]
)
# Create orchestrator agent with sub-agents
orchestrator = Agent(
    name="HelpDeskOrchestrator",
    description="Routes customer requests to appropriate specialists",
    system_prompt="""
    You are a help desk coordinator. Route customer requests to the appropriate specialist:
    - Use CustomerService for billing, account, or general service questions
    - Use TechnicalSupport for product issues, bugs, or technical problems
    - Handle simple greetings and general questions yourself
    """,
    sub_agents=[customer_service, technical_support]
)
```

### Implementation Details
Each sub-agent automatically generates a `handoff_to_{agent_name}` tool that:
- Transfers complete control to the sub-agent
- Terminates orchestrator execution after delegation  
- Makes the sub-agent's response the final response
- No post-processing by the orchestrator
**Response Flow**: User Request → Orchestrator Analysis → Delegation → Sub-agent Final Response

### Use Case

**Customer Support**: Route billing questions to CustomerService agent, technical issues to TechnicalSupport agent
**Multi-Domain Assistant**: Route research to Researcher, writing to Writer, coding to Coder  
[...]

PR #3197, opened 2026-07-10
docs: add agent-as-tool handoff doc

liramon2:
## Description
Add design document for agent-as-tool handoff.

## Related Issues

#911

## Documentation PR

This is the documentation PR.

## Type of Change

Documentation update

----


team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+- Extending Graph/Swarm primitives. They serve a different use case where agents are peers with limited context sharing
+- Replacing agents-as-tools. It already implements return semantics, whereas this proposal focuses on delegation semantics
+- Sticky routing. The orchestrator should not automatically reuse the current sub-agent in the next invocation because the new user prompt may favor a different sub-agent
+- Circular delegation detection. Execution may be handed to a previous agent in a cycle but with new information, which is a valid agent flow.
+
+## Proposal
+
+`AgentConfig` gains an optional `subAgents` field accepting an array of `InvokableAgent`s. The orchestrator injects sub-agent names and descriptions into its system prompt so the model can make routing decisions.
+
+```ts
```
yonib05: Thoughts on having the same agent as a tool pattern with some handoff flag maybe?

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+
+**Status**: Proposed
+
+**Date**: 2026-07-09
+
+**Issue**: https://github.com/strands-agents/harness-sdk/issues/911
+
+## Problem
+
+The SDK does not have a first-class handoff mechanism for fully delegating execution from an orchestrator agent to a sub-agent. Builders want to create agents that follow this pattern, but the current primitives either make unnecessary model calls or lose context.
```
JackYPCOnline: Is conditional edge considered to be an option?

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+**Non-goals:**
+
+- Extending Graph/Swarm primitives. They serve a different use case where agents are peers with limited context sharing
+- Replacing agents-as-tools. It already implements return semantics, whereas this proposal focuses on delegation semantics
+- Sticky routing. The orchestrator should not automatically reuse the current sub-agent in the next invocation because the new user prompt may favor a different sub-agent
+- Circular delegation detection. Execution may be handed to a previous agent in a cycle but with new information, which is a valid agent flow.
+
+## Proposal
+
+`AgentConfig` gains an optional `subAgents` field accepting an array of `InvokableAgent`s. The orchestrator injects sub-agent names and descriptions into its system prompt so the model can make routing decisions.
```
mkmeral: I wouldn't call it subagents. It can be confused with agents-as-tools concept

team/designs/0015-multi-agent-handoff.md
yonib05: It could be exposed to the agent as Hand off to XYZ tool for each agent added (basically an amalgamation of Hand Off and the agent)

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+- Intuitive API surface for configuring sub-agents
+- Mechanism for sharing message history between two agents
+- Full execution handoff, where the sub-agent's streams to the user without orchestrator post-processing
+- Supports composable delegation (ex. A delegates to B, which delegates to C)
+- Integrates with graphs and swarms
+- Compatibility with both TypeScript and Python SDKs
+
+**Non-goals:**
+
+- Extending Graph/Swarm primitives. They serve a different use case where agents are peers with limited context sharing
```
opieter-aws: Can you elaborate on the use cases the existing constructs are solving and the use case that this proposal will solve to back up this decision?

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+    return new ToolResultBlock({
+      toolUseId: toolContext.toolUse.toolUseId,
+      status: 'success',
+      content: [new JsonBlock({ json: { agentName: agent_name, message: message ?? null } })],
+    })
+  }
+}
+```
+
+After `executeTools()` completes, the agent loop checks whether any `ToolUseBlock` in the assistant message has `name === HANDOFF_TOOL_NAME`.
```
mkmeral: can we make this more generic? Overall we should not depend on tool names for these type of actions imo

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+const swarm = new Swarm({
+  nodes: [researcher, architect, coder],
+})
+```
+
+As with graphs, swarm agents lose context during handoffs. Only the handoff message and optional context (formatted as JSON) are transferred between agents. The original agent's full conversation history is lost. Additionally, sub-agents can hand off back to the original agent, which breaks the delegation pattern. When the goal is route to a specialist, bidirectional flow adds complexity without value.
+
+### What's Missing
+
+All three patterns share a fundamental gap: there is no way for an orchestrator to transfer full context to a sub-agent while passing the final response directly to the user.
```
lizradway: > there is no way for an orchestrator to transfer full context to a sub-agent

do we want this? have people asked for this?

subagent handoff is ideal for context management because it can isolate context needed for its specific task. _full context_ would overwhelm subagents rapidly and consume exponentially more tokens.

maybe what we're looking for is selective context.

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+- Full execution handoff, where the sub-agent's streams to the user without orchestrator post-processing
+- Supports composable delegation (ex. A delegates to B, which delegates to C)
+- Integrates with graphs and swarms
+- Compatibility with both TypeScript and Python SDKs
+
+**Non-goals:**
+
+- Extending Graph/Swarm primitives. They serve a different use case where agents are peers with limited context sharing
+- Replacing agents-as-tools. It already implements return semantics, whereas this proposal focuses on delegation semantics
+- Sticky routing. The orchestrator should not automatically reuse the current sub-agent in the next invocation because the new user prompt may favor a different sub-agent
```
opieter-aws: Why is this a non-goal? Is this an implementation detail?

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+
+## Goals and Non-Goals
+
+**Goals:**
+
+- Intuitive API surface for configuring sub-agents
+- Mechanism for sharing message history between two agents
+- Full execution handoff, where the sub-agent's streams to the user without orchestrator post-processing
+- Supports composable delegation (ex. A delegates to B, which delegates to C)
+- Integrates with graphs and swarms
```
opieter-aws: This seems contradictory to me with the first non-goal

team/designs/0015-multi-agent-handoff.md
liramon2: Any suggestions on what to call it instead?

team/designs/0015-multi-agent-handoff.md
JackYPCOnline: Can we just save the context to a file and let sub-agent read that? full context looks expensive for an agent, why we need it?

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+
+orchestrator.invoke("Where's my refund for order #12345?")
+```
+
+## Consequences
+
+- Builders get a simple path to multi-agent routing that avoids the extra model calls of agents-as-tools and the context loss of swarms.
+- Adding `subAgents` as a first-class concept introduces a new relationship type between agents. This needs clear documentation to help builders choose between agent-as-tools, sub-agents, and multi-agent primitives.
+
+## Open Questions
```
yonib05: What happens with tools? Does the orchestrator pass tools available to the subagent, or is the subagent defined by the tools it has configured.

team/designs/0015-multi-agent-handoff.md
liramon2: The existing constructs solve the case of semi-independent agents working together to respond to a request, while the current use case is an orchestrator agent completely delegating a request to another agent (ex. a specialist).

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+})
+```
+
+```python
+orchestrator = Agent(
+    sub_agents=[customer_service, technical_support],
+)
+```
+
+The orchestrator agent adds a `HandoffTool` instance to its tool registry if `subAgents` is non-empty. `HandoffTool` is a `Tool` subclass whose `inputSchema` exposes an `agent_name` string field and an optional `message` string field for routing context. It validates the agent's choice and echoes it back in a `ToolResultBlock`.
```
lizradway: why would we want this as a subclass? can't this be a regular tool?

team/designs/0015-multi-agent-handoff.md
lizradway: implementation detail, but id be worried about collisions with user passed tools

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+After `executeTools()` completes, the agent loop checks whether any `ToolUseBlock` in the assistant message has `name === HANDOFF_TOOL_NAME`.
+- If it does not find one, the orchestrator handles the request normally, with no delegation overhead.
+- If it does find one, the `agentName` is extracted from the JSON content and the sub-agent is resolved by name.
+  - If the name is invalid, an error tool result is appended and the loop continues so the model can self-correct.
+
+The following proposals diverge from this point onwards.
+
+### Recommended: Sub-invocation, Clone Orchestrator History
+
+The orchestrator's full `messages` array is deep cloned as the input to the sub-agent's `stream()`. The sub-agent applies its own system prompt, tools, and conversation manager. From the sub-agent's perspective, this is a normal invocation. Its response streams directly to the caller without re-entering the orchestrator's loop.
```
opieter-aws: What is the rationale for cloning the full `messages` array as input for the sub-agent?

This can be token-inefficient and bias the outcome of the ask for the subagent.

team/designs/0015-multi-agent-handoff.md
lizradway: +1

lizradway: this is one of the wins of offloading context + memory to a shared storage interface that can be assumed by any agent.

team/designs/0015-multi-agent-handoff.md
lizradway: +1

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+### What's Missing
+
+All three patterns share a fundamental gap: there is no way for an orchestrator to transfer full context to a sub-agent while passing the final response directly to the user.
+
+## Goals and Non-Goals
+
+**Goals:**
+
+- Intuitive API surface for configuring sub-agents
+- Mechanism for sharing message history between two agents
```
JackYPCOnline: can't we do that with hook now?

team/designs/0015-multi-agent-handoff.md
liramon2: Possibly. The name check was inspired by how we check for the structured output tool, so we might want to change that later as well.

team/designs/0015-multi-agent-handoff.md
notowen333: >  Its response streams directly to the caller without re-entering the orchestrator's loop.

Is this a common way to define "subagent" across the SDK ecosystem?

In Claude Code, when it spawns a subagent, that agent works independently and then the main Agent pulls in and interprets the response

pgrayy: A sub-agent may not necessarily need the full conversation history. It could be that the worker asks an agent to complete a simple task that doesn't require any additional context beyond the task definition.

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
+# Multi-Agent Handoff
+
+**Status**: Proposed
+
+**Date**: 2026-07-09
+
+**Issue**: https://github.com/strands-agents/harness-sdk/issues/911
+
+## Problem
```
notowen333: Did you think about dynamic subagents? That seems like one of the most common ways this pattern is used.

As opposed to declarative subagents before invocation which is discussed in this doc

team/designs/0015-multi-agent-handoff.md
liramon2: People did support this in the original issue: https://github.com/strands-agents/harness-sdk/issues/911

team/designs/0015-multi-agent-handoff.md
notowen333: > The orchestrator injects sub-agent names and descriptions into its system prompt so the model can make routing decisions.

If we go with such a mechanic, we should look to unify and organize framework sys prompt injection.

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+- If it does find one, the `agentName` is extracted from the JSON content and the sub-agent is resolved by name.
+  - If the name is invalid, an error tool result is appended and the loop continues so the model can self-correct.
+
+The following proposals diverge from this point onwards.
+
+### Recommended: Sub-invocation, Clone Orchestrator History
+
+The orchestrator's full `messages` array is deep cloned as the input to the sub-agent's `stream()`. The sub-agent applies its own system prompt, tools, and conversation manager. From the sub-agent's perspective, this is a normal invocation. Its response streams directly to the caller without re-entering the orchestrator's loop.
+
+Once the sub-agent completes, the orchestrator appends the sub-agent's final assistant message to its own `messages` array. This gives the orchestrator continuity on subsequent invocations: the user can follow up and the orchestrator sees the prior response in its history. Only the final assistant message is appended, not the sub-agent's internal tool calls or reasoning, which would reference tools the orchestrator does not have.
```
opieter-aws: Can the subagent ask for user input in multiple turns? What if there's critical information in that message history, would it not get lost when returning back to the orchestrator with only the final assistant's message?

team/designs/0015-multi-agent-handoff.md
lizradway, line 1: just in general, im really hesitant to believe that a subagent would almost ever want the _entire_ context window in full.


in the case that we would want to give subagents access to orchestrator context, we would likely achieve this through offloading context to shared storage. 


what else do we gain from this?

team/designs/0015-multi-agent-handoff.md
liramon2: The existing primitives serve other purposes already. It is possible to implement multi-agent handoffs using them even though they weren't considered in this initial proposal. If that's the case, then these goals need revision.

team/designs/0015-multi-agent-handoff.md
mehtarac: What happens when the model calls HandoffTool AND other tools in the same turn?

team/designs/0015-multi-agent-handoff.md
lizradway: I think the main difference then is no post-processing model call, which i think could justify a flag in the agent as tool?

team/designs/0015-multi-agent-handoff.md
mkmeral: Also one more thing, do we need to be involved with/update the agent loop? Can't we implement this as a standalone tool only? 🤔

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+- Composable without special handling. A sub-agent with its own `subAgents` chains naturally (A to B to C) because each handoff is a standard `stream()` call.
+
+**Cons:**
+- Builds the full message history in sub-agent context on each handoff. This increases the sub-agent's input token cost, especially for long conversations
+- The sub-agent's intermediate work is lost on subsequent invocations, so the orchestrator cannot answer questions like "What tools did you use?"
+
+
+### Alternative: Sub-invocation, Return Full Sub-Agent History
+
+This is mostly the same as the recommended approach. The only difference is that instead of appending the sub-agent's final assistant message, the orchestrator appends the sub-agent's entire message history. This includes all tool calls, tool results, reasoning, and intermediate assistant messages generated during the sub-agent's execution.
```
pgrayy: I don't think this behavior is necessary for the use cases we are trying to cover with subagents. I like the idea of just sending back the latest message rather than the full message history of the sub agent. The orchestrator shouldn't need to reason about it in my opinion.

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+- Full-history continuity, since a single message history is maintained
+- No sub-invocation overhead. The same loop and messages run on the next cycle, albeit different config.
+
+**Cons:**
+
+- Strands has no runner/agent separation. Mutating the agent's identity mid-loop breaks hooks, plugins, middleware, and telemetry, which all hold a reference to `this` and expect it to be stable.
+- Identity swapping is inconsistent with the rest of the Strands SDK, since existing primitives are separate objects with separate lifecycles
+- Requires identity bookkeeping that may break as more data members are introduced to Agent
+
+## Developer Experience
```
notowen333: doc nit: this code example is not informative for how this works and is different than existing mechanisms.

The setup looks exactly the same as agent as tool

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+```
+
+**Pros:**
+- The orchestrator model can reason about the sub-agent's tool calls and other intermediate results, improving responses to subsequent orchestrator invocations
+
+**Cons:**
+- The orchestrator sees `toolUse` blocks for tools it does not have registered, so it may attempt to call those tools on subsequent invocations and fail
+- Consumes significantly more orchestrator context than providing it with only the final sub-agent result
+
+### Alternative: Sub-invocation, Summarize Orchestrator History
```
opieter-aws: Can we expose these different flavors as options instead?

team/designs/0015-multi-agent-handoff.md
mehtarac: instead of `subagents` can we call `agentHandoff`?

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+orchestrator.invoke("Where's my refund for order #12345?")
+```
+
+## Consequences
+
+- Builders get a simple path to multi-agent routing that avoids the extra model calls of agents-as-tools and the context loss of swarms.
+- Adding `subAgents` as a first-class concept introduces a new relationship type between agents. This needs clear documentation to help builders choose between agent-as-tools, sub-agents, and multi-agent primitives.
+
+## Open Questions
+- Can sub-agents have session managers for persisting state? If so, how is that handled during delegation?
```
mkmeral: I'd say not "session" managers but maybe for snapshots? cc @opieter-aws  (i feel like term session will get really confusing lol)

team/designs/0015-multi-agent-handoff.md
liramon2: The first two points are mechanisms under the hood, while the last two points are extended use cases if users want that kind of flexibility. So I don't think it's contradictory to making a basic API surface exposed to the user.

team/designs/0015-multi-agent-handoff.md
liramon2: The sub-agent has its own set of tools, since it is defined as a separate agent.

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+**Pros:**
+- The orchestrator model can reason about the sub-agent's tool calls and other intermediate results, improving responses to subsequent orchestrator invocations
+
+**Cons:**
+- The orchestrator sees `toolUse` blocks for tools it does not have registered, so it may attempt to call those tools on subsequent invocations and fail
+- Consumes significantly more orchestrator context than providing it with only the final sub-agent result
+
+### Alternative: Sub-invocation, Summarize Orchestrator History
+
+This option uses the same mechanism as the recommended approach. The only difference is that instead of cloning the orchestrator's full message history, the orchestrator generates a summary of the conversation and passes it to the sub-agent as a single user message.
```
pgrayy: I would ask what all use cases are we trying to solve with sub agents. In my mind, I feel that an orchestrator should delegate to a sub agent to handle specific tasks and thus does not need context on the full conversation history of the orchestrator. The subagent really just needs context specifically related to the assigned task.

team/designs/0015-multi-agent-handoff.md
lizradway: it looks mainly like people want to skip the doubled llm calls forced from returning to the orchestrator agent: "Langchain has a `return_direct` on its tool decorator for this too." I think that kind of flag makes sense.

team/designs/0015-multi-agent-handoff.md
JackYPCOnline: Indeed, orchestrator side has no change right? it is all about what sub-agent can observe?

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+
+**Issue**: https://github.com/strands-agents/harness-sdk/issues/911
+
+## Problem
+
+The SDK does not have a first-class handoff mechanism for fully delegating execution from an orchestrator agent to a sub-agent. Builders want to create agents that follow this pattern, but the current primitives either make unnecessary model calls or lose context.
+
+### Current State
+
+There are currently three workarounds for delegating agent execution to sub-agents, but each have notable flaws.
```
notowen333: doc nit: I would want to see use cases and best-case experiences with the handoff pattern before enumerating issues with existing constructs.

team/designs/0015-multi-agent-handoff.md
liramon2: My bad, the wording was unclear. Yeah, I meant as a regular tool.

team/designs/0015-multi-agent-handoff.md
mkmeral: wdym? What's the devx? (we can iterate on that during API bar raising)

team/designs/0015-multi-agent-handoff.md
mkmeral: +1 or if the model calls multiple handoffs at the same time

team/designs/0015-multi-agent-handoff.md
liramon2: It is to pass the full context from the orchestrator to the sub-agent so that it has everything it needs to respond. If the full message history isn't needed, we can trim it down.

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+export class HandoffTool extends Tool {
+  readonly name = HANDOFF_TOOL_NAME
+  readonly toolSpec: ToolSpec
+
+  constructor(subAgents: InvokableAgent[]) {
+    super()
+    const agentNames = subAgents.map(a => a.name ?? a.id)
+    this.toolSpec = {
+      name: HANDOFF_TOOL_NAME,
+      description: 'Delegate the conversation to a specialist agent.',
```
notowen333: Are we only building this construct around full delegation?

What would be gained/lost by widening the scope

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+```
+
+**Pros:**
+
+- Full-history continuity, since a single message history is maintained
+- Full bidirectional visibility between orchestrator and sub-agent
+
+**Cons:**
+
+- Requires significant redesigns to state management, from a agent-owned mutable history to a shared state object
```
pgrayy: https://github.com/strands-agents/docs/pull/551/changes
^ Related.

team/designs/0015-multi-agent-handoff.md
JackYPCOnline: From my understanding, session manager wise, no. snapshot wise, subagent should be treated as tool in this scenario?

team/designs/0015-multi-agent-handoff.md
lizradway: something like https://reference.langchain.com/python/langchain-core/tools/base/BaseTool/return_direct imo

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+    stopReason: 'endTurn',
+    lastMessage: subResult.lastMessage,
+    invocationState,
+    traces: this._tracer.localTraces,
+    metrics: this._meter.metrics,
+  })
+}
+```
+
+#### Interrupts
```
notowen333: doc nit: Can move this into the appendix. Seems too low level for the overall proposal

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+const travelAgent = new Agent({...})
+
+const orchestrator = new Agent({
+  tools: [researchAgent, productAgent, travelAgent],
+})
+```
+
+This pattern has two problems:
+
+1. The parent agent must post-process the sub-agent's response with an additional model call before returning the response to the user, even when the sub-agent's output is ready to serve as-is. This adds latency and token usage for scenarios where no additional reasoning is required.
```
mehtarac: Wondering if we can extend the `agents-as-tools` design to use the handoff approach?

Instead of two separate mechanisms (agents-as tools + subAgents/HandoffTool), you keep a single mechanism, `agents registered as tools` but add a property that controls post-execution behavior

Example:

```
const customerService = new Agent({
    name: 'CustomerService',
    description: '...',
    handoff: true, // <-- new property
  })
```

team/designs/0015-multi-agent-handoff.md
```diff
@@ -0,0 +1,443 @@
...
+
+## Problem
+
+The SDK does not have a first-class handoff mechanism for fully delegating execution from an orchestrator agent to a sub-agent. Builders want to create agents that follow this pattern, but the current primitives either make unnecessary model calls or lose context.
+
+### Current State
+
+There are currently three workarounds for delegating agent execution to sub-agents, but each have notable flaws.
+
+#### Agents-as-tools
```
opieter-aws: The two cons listed for agents-as-tools are one problem: the orchestrator’s post-tool model call. A direct-return tool would skip this and solve both problems. Since we already have the pieces in the SDK, can we also explore a direct-return agent-as-tool as an alternative solution? The SubAgent could just be sugar on top of the tool settings, and we avoid introducing another multi-agent setup

team/designs/0015-multi-agent-handoff.md
pgrayy: Reading through the google adk and openai sdk mechanisms I get a better understanding of the additional use cases you are trying to solve. Developer experience as well. So it seems the idea is that through an orchestrator interface, you could shift the user to talk to a subagent and then be handed back to the orchestrator when done. Kind of like a telephone operator.

team/designs/0015-agent-as-tool-handoff.md
```diff
@@ -0,0 +1,302 @@
...
+**Non-goals:**
+
+- Multi-turn conversation handoff where the user is transferred to a specialized agent for extended interaction. This can be explored in a separate proposal.
+- Extending Graph/Swarm primitives. They serve a different use case where agents are peers with limited context sharing.
+
+## Proposal
+
+The following API surfaces configure agent-as-tools to stream directly to the user without an additional orchestrator model call.
+
+### Recommended: `handoff` flag on `asTool()`
```
mkmeral: as a devx, i agree

team/designs/0015-agent-as-tool-handoff.md
```diff
@@ -0,0 +1,302 @@
...
+- Easy to mix delegation agents with regular tools on the same orchestrator.
+
+**Cons:**
+- Two arrays that feed into the same tool registry may confuse builders.
+- Introduces a new top-level config field, expanding the `AgentConfig` surface area.
+- If non-agent tool handoffs are supported in the future, `handoffAgentTools` would need to be renamed and accept arbitrary `Tool` instances. This creates a deprecation path: either broaden the field's type (breaking the original contract) or introduce a replacement field and deprecate `handoffAgentTools`.
+
+### Agent Loop Changes
+
+Regardless of which API surface is chosen, the handoff behavior works inside the agent loop. When the agent loop builds the `toolSpecs` array for a model call, it appends a suffix to the description of any tool marked as a handoff target. The suffix resembles the following:
```
mkmeral: why are we appending custom text in agent loop? can't we just add it while building agent as tool, which i assume `.as_tool()` is doing

team/designs/0015-agent-as-tool-handoff.md
```diff
@@ -0,0 +1,302 @@
...
+
+After `executeTools()` completes and the assistant message + tool-result message are appended to history, the loop checks for a successful handoff result:
+- If the handoff tool's result has `status: 'error'`, skip it and continue the normal loop so the model can recover or try a different tool.
+- If the `ToolResultBlock` has `status: 'success'`, return it as `AgentResult.lastMessage` with `stopReason: 'endTurn'`. The tool-result message (role: `user`) is the `lastMessage` — no synthetic assistant message is created.
+
+#### Multiple Tool Calls in a Single Turn
+
+When the model calls a handoff tool and other tools in the same turn despite the handoff tool description suffix, all tools execute to completion. Afterwards, all tool use and tool result blocks are appended to history.
+- If there are non-handoff tools, they are not passed back to thea orchestrator model
+- If there are multiple handoff tools, the first successful result in tool use order is returned. This can be achieved by iterating on the `ToolUseBlock`s in the assistant message and checking the corresponding `ToolResultBlock`.
```
mkmeral: why? is this the experience that we want? 

the issue I see here is that we will be ignoring responses from other agents, and those agents might have taken mutative/destructive action that the user should be aware. I don't think we should be ignoring anything

team/designs/0015-agent-as-tool-handoff.md
```diff
@@ -0,0 +1,302 @@
...
+
+### Agent Loop Changes
+
+Regardless of which API surface is chosen, the handoff behavior works inside the agent loop. When the agent loop builds the `toolSpecs` array for a model call, it appends a suffix to the description of any tool marked as a handoff target. The suffix resembles the following:
+
+> "Calling this tool will return its response directly to the user as the final answer. It should be the only tool called in the turn."
+
+This nudges the model to treat the tool call as a terminal action, avoiding wasted tokens on preamble text, parallel tool calls, or post-processing plans. By appending the suffix at read time, the tool's description remains clean (reflecting only what the builder wrote or what was derived from the agent's description) and the handoff instruction remains at the model-calling boundary. This approach also extends naturally to non-agent tool handoffs in the future.
+
+After `executeTools()` completes and the assistant message + tool-result message are appended to history, the loop checks for a successful handoff result:
```
mkmeral, line 227: @strandly-the-agent is there a way to achieve this through plugins instead of modifying the agent loop? can we use hooks, middlewares, and so on to enable this use case?

strandly-the-agent:
@mkmeral re: your question on the *Agent Loop Changes* section (whether this can be done via plugins/hooks instead of modifying the loop) — replying here since the review-thread reply is blocked for me.

**Short answer: yes for the common case, and mostly without touching the loop.** The behavior that section describes — run the sub-agent tool, then skip the orchestrator's post-processing model call and end the turn — can be built today as a `Plugin`, in both SDKs, using the existing `BeforeModelCallEvent.cancel` lever. The one thing a plugin *can't* fully reproduce is verbatim tool-result-as-`lastMessage` fidelity, which is what would argue for a (smaller, more general) loop primitive rather than a handoff-specific branch.

<details>
<summary>How a plugin does it (mechanism + precedent)</summary>

A single plugin wires three things:
- `getTools()` returns the `asTool()` wrappers, with the terminal-action suffix written straight into each tool's description (no read-time injection needed).
- an `AfterToolCallEvent` hook detects a *successful* handoff tool result and stashes its text.
- a `BeforeModelCallEvent` hook sets `event.cancel = <stashed text>`.

What `cancel` already does (both SDKs): it yields an `assistant` message with a single text block and `stopReason: endTurn`, then returns — i.e. exactly "skip the post-processing call and end the turn," zero extra model call. See `strands-ts/src/agent/agent.ts:1946` and `strands-py/src/strands/event_loop/event_loop.py:493`.

This is the established extension model, not a hack: agents already take `plugins: [...]`, and vended plugins combine `getTools()` + hooks (`AgentSkills`) and drive control flow via `BeforeModelCallEvent` (`GoalLoop`). The multiple/parallel-tool-call case in the section below also falls out for free — all tools execute, then the one pre-model-call `cancel` returns the first stashed handoff result; no loop recursion or bespoke iteration needed.

The nudge suffix is likewise user-space: just author it into the tool description.
</details>

<details>
<summary>The gap a plugin can't close (and the minimal primitive it suggests)</summary>

`cancel` only carries a **string** and forces the final message to `role: assistant` with one text block. So:
- **Text / JSON / code** → preserved verbatim (it's stringifiable), no paraphrasing — this covers both use cases in the Problem section. ✅
- The doc's exact semantic — *"the tool-result message (role `user`) is the `lastMessage`, no synthetic assistant message"* — and arbitrary/multimodal tool-result content blocks → **not reachable** through `cancel`.

So if verbatim, role-preserving fidelity is a hard requirement, the minimal change isn't necessarily a handoff-specific loop branch — generalizing the model-call cancel to accept a full final `Message` (instead of just a string) would let this plugin implement handoff at full fidelity, and would help other "short-circuit the turn with a prepared answer" cases too.

**Suggested framing:** ship the common (text/JSON) case as a vended plugin now — no loop change; reserve loop work for the fidelity gap, and prefer generalizing `cancel` over a dedicated handoff code path if you go there. Happy to prototype the plugin against this design if useful.
</details>

team/designs/0015-agent-as-tool-handoff.md
liramon2: That is possible. But, it would mix the description explaining what the tool does with routing instructions, which are different concerns.

team/designs/0015-agent-as-tool-handoff.md
liramon2: That's a good point. Returning the first successful result was a simple way to handle this edge case, but continuing the agent loop is probably more robust.

team/designs/0015-agent-as-tool-handoff.md
mkmeral: > it would mix the description explaining what the tool does with routing instructions

I would argue that these are the same things. What tool does is the basis on which cases you want to route to this tool/agent, so I think it's fine

team/designs/0015-agent-as-tool-handoff.md
```diff
@@ -0,0 +1,348 @@
...
+
+```typescript
+const customerService = new Agent({...})
+
+const technicalSupport = new Agent({...})
+
+const orchestrator = new Agent({
+  tools: [
+    customerService.asTool({ handoff: true }),
+    technicalSupport.asTool({ handoff: true }),
```
mehtarac, line 70: I like this devX. Simple and effective imo

team/designs/0015-agent-as-tool-handoff.md
mkmeral: I'd probably try to detect this early, somewhere around [after model call](https://strandsagents.com/docs/user-guide/concepts/agents/hooks/#single-agent-lifecycle), before we start calling the tool registry. So we fail early.
 Given customers have said to explicitly handoff (through the devx flag), I wouldn't want us to continue the agent loop

team/designs/0015-agent-as-tool-handoff.md
```diff
@@ -0,0 +1,348 @@
...
+- Compatible with both TypeScript and Python SDKs
+
+**Non-goals:**
+
+- Multi-turn conversation handoff where the user is transferred to a specialized agent for extended interaction. This can be explored in a separate proposal.
+- Extending Graph/Swarm primitives. They serve a different use case where agents are peers with limited context sharing.
+
+## Proposal
+
+The following API surfaces configure agent-as-tools to stream directly to the user without an additional orchestrator model call.
```
mehtarac, line 56: Question: The sub-agent streams normally during its execution as a tool and the handoff just prevents
  the orchestrator from making an additional model call afterwards right?

team/designs/0015-agent-as-tool-handoff.md
```diff
@@ -0,0 +1,348 @@
...
+
+> "Calling this tool will return its response directly to the user as the final answer. It should be the only tool called in the turn."
+
+This nudges the model to treat the tool call as a terminal action, avoiding wasted tokens on preamble text, parallel tool calls, or post-processing plans. By appending the suffix at read time, the tool's description remains clean (reflecting only what the builder wrote or what was derived from the agent's description) and the handoff instruction remains at the model-calling boundary. This approach also extends naturally to non-agent tool handoffs in the future.
+
+After `executeTools()` completes and the assistant message + tool-result message are appended to history, the loop checks for a successful handoff result:
+- If the handoff tool's result has `status: 'error'`, skip it and continue the normal loop so the model can recover or try a different tool.
+- If the `ToolResultBlock` has `status: 'success'`, return it as `AgentResult.lastMessage` with `stopReason: 'endTurn'`. The tool-result message (role: `assistant`) is the `lastMessage`.
```
mehtarac: I believe a Tool-Result has a role of `user`, and not `assistant`. Are we suggesting that we'd return it as an `assistant` message instead of `user`?

mehtarac: Open to consideration: 

- Should we introduce a new stop reason, `stopReason: handoff` to signal that a model turn ended naturally vs. because of a handoff? Is it needed?

- Should we have a field on AgentResult called `handoff` for observability purposes?

team/designs/0015-agent-as-tool-handoff.md
liramon2: Yes

team/designs/0015-agent-as-tool-handoff.md
liramon2: Yes, since AgentResult is currently constructed with an 'assistant' role everywhere in the codebase. Changing it would break an established pattern.

liramon2: 1. Yeah, it would enhance observability assuming endTurn means the model is finished talking, which it isn't in this case.
2. If we introduce a new stopReason, then there's no need.

liramon2: `stopReason: handoff` requires minor changes to the agent loop and events if we use the plugin approach suggested by @mkmeral, since `AfterToolsEvent` doesn't expose a way to change the `stopReason`. It's hardcoded to endTurn right now.

mehtarac (approved):

PR #3198, opened 2026-07-10
feat(tools): add programmatic_tool_caller vended tool

strandly-the-agent:
## Description

Adds `programmatic_tool_caller`, a vended tool that lets an agent orchestrate its **other** tools by writing Python code, instead of issuing one tool call per model round-trip. Each registered tool is exposed to the code as an `async` function, so the model can chain, loop over, and parallelize calls (`asyncio.gather`) in a single turn. Only what the code sends to `print()` comes back into the model's context — intermediate tool results stay in the code's local scope unless printed — which keeps large intermediate payloads out of the context window.

This ports the community `programmatic_tool_caller` (strands-agents/tools#387) into the SDK's `vended_tools`, adapted to the vended-tool conventions: a `make_*` factory plus a default instance, an `async` tool with `@tool(context="tool_context")`, and no `rich`/interactive-consent UI (consistent with the existing `bash`/`file_editor` vended tools).

### Public API

Exported from `strands.vended_tools`:

```python
from strands import Agent
from strands.vended_tools import programmatic_tool_caller

agent = Agent(tools=[programmatic_tool_caller, calculator])

# The model calls the tool with code like:
#   results = await asyncio.gather(
#       calculator(expression="10 * 1"),
#       calculator(expression="10 * 2"),
#   )
#   print(results)
```

Or build a customized instance:

```python
from strands.vended_tools import make_programmatic_tool_caller

# Limit which tools the code can reference by name, and/or rename the tool.
tool = make_programmatic_tool_caller(allowed_tools=["calculator"], name="run_code")
```

`make_programmatic_tool_caller(*, allowed_tools: list[str] | None = None, name: str = "programmatic_tool_caller", description: str = ...)`.

### Security posture (please weigh in)

This tool runs model-authored code in the **host process** and does **not** sandbox it — like the existing `bash` vended tool, it should only be used where the code the agent produces is trusted. `allowed_tools` limits which tools are exposed *by name*; it is deliberately documented as **not** a security boundary (code that isn't fully trusted can still reach any registered tool via object references). The default instance exposes every other registered tool. I've kept that default for parity with the source and with `bash`, but flagging it: **whether the shipped default should nudge toward an explicit allow-list is a posture decision worth a maintainer's call.** This likely warrants `needs-api-review`.

### Notable design decisions

- **Top-level `await` via `compile(..., flags=ast.PyCF_ALLOW_TOP_LEVEL_AWAIT)` + `eval`**, rather than wrapping the code in an `async def`. Wrapping requires re-indenting, which silently corrupts continuation lines inside multi-line string literals — a real hazard for a tool whose job is relaying args (bash scripts, file contents) to other tools verbatim. This also keeps tracebacks pointing at the model's own line numbers.
- **Output captured via an injected, buffer-bound `print`**, not a process-global `redirect_stdout`. A global redirect would swallow unrelated concurrent/background stdout into the result (and lose it from the real stream); scoping to the code's own `print()` matches the stated contract and is concurrency-safe.

The TypeScript SDK has no equivalent today (its `vended-tools/notebook` is unrelated), so no parity change is needed here; a TS port can follow up.

## Related Issues

- Ports strands-agents/tools#387
- Motivating SDK issues: strands-agents/sdk-python#1540, strands-agents/sdk-python#1349

## Type of Change

New feature

## Testing

25 unit tests drive the tool through a real `Agent` (with a mocked model that is never invoked, since these are direct tool calls), covering: single/loop/`gather` execution, print-only output, tool-not-exposed and `allowed_tools` filtering, syntax/runtime/`SystemExit` errors, tool-error-as-exception, namespace-clash rejection (`asyncio`/`print`/`__builtins__`), multi-line-string round-trip, and no-await/comment-only code. `ruff` (format + lint) and `mypy --strict` are clean.

<details>
<summary>Independent review loop (3 rounds, converged)</summary>

This branch was taken through an independent, fresh-context review loop before opening:

- **Round 1 — correctness:** found a **blocker** — `textwrap.indent`-based code wrapping silently injected leading whitespace into continuation lines of multi-line string literals. Fixed by switching to `ast.PyCF_ALLOW_TOP_LEVEL_AWAIT` + `eval` (also fixes traceback line numbers and empty/comment-only code).
- **Round 2 — adversarial + API bar-raiser:** two majors — (1) process-global `redirect_stdout` swallowed concurrent/background stdout → switched to a scoped, buffer-bound `print`; (2) the `extra_modules` factory param was redundant (code can `import`) and a footgun (a tool named like a module bricked every call) → removed it. Plus: honest security docs (`allowed_tools` is not a boundary), reserved `print`/`__builtins__`, and model-facing description improvements. Confirmed no deadlock under large `gather` and that `asyncio.CancelledError` correctly propagates.
- **Round 3 — verification:** independent re-derivation + gates + repros; **converged, no findings.**

One inherent, accepted tradeoff: tracebacks include a single internal `eval` frame above the user's frame (the user's line numbers stay correct).

I'm an AI agent (Strandly). A human should review and approve before merge — I can't self-approve, and I've flagged the security-posture decision above for a maintainer.
</details>

## Checklist
- [ ] I have updated the documentation accordingly (site/ docs can follow up; flag if wanted in this PR)
- [x] I have added an appropriate example to the documentation to outline the feature

----

🤖 Opened by Strandly, an AI agent, at the request of `mkmeral`. Please review carefully.


strandly-the-agent:
## Tested against real MCP tools — found and fixed a blocker; code is ready, merge needs 2 human approvals + a posture call

`mkmeral` asked me to review + test this, specifically with MCP tools. Doing that surfaced a **real functional gap**, now fixed and covered:

🔴 **MCP tools with non-identifier names were completely uncallable.** MCP servers routinely name tools `fetch-url`, `ns.fetch`, `ns__pfx-tool`. Those were injected into the exec namespace under their raw key only, so the model had no way to call them: `await fetch-url(...)` parses as subtraction (`fetch - url(...)` → `NameError: name 'fetch' is not defined`) and `await fetch_url(...)` → `NameError`. Fixed by injecting a normalized identifier alias (`fetch_url`, `ns_fetch`), skipped when it would shadow a reserved name, a real tool, or another tool's alias (ambiguous → dropped, not guessed).

🟡 Also fixed from independent review: the print-capture isolation contract had **no test** — I swapped the injected `print` for `contextlib.redirect_stdout` and all tests still passed. Added a concurrency regression test and verified it actually discriminates (fails on the mutant, passes on real code).

**Also: this branch was 135 commits behind `main`** and `vended_tools/__init__.py` had a real conflict with the newly-added `http_request`/`sleep` tools. Merged and resolved.

### ✅ Verified
| Check | Result |
|---|---|
| MCP integration (real stdio server) | **8/8 pass** — sequential, loops, chaining, `asyncio.gather`, error propagation, hyphen + dot names, print-only output |
| Unit tests | **31 pass** |
| Full `strands-py` suite | **3873 pass** (4 telemetry errors are pre-existing on clean `main` — missing `opentelemetry-exporter-otlp` extra) |
| ruff / ruff-format / mypy strict | clean |
| Independent review passes | correctness **APPROVE**, adversarial **APPROVE**, test-quality **gaps-found → fixed** |

### On "good to merge"
Code-wise I believe it is. But I can't green-light it, and two things are genuinely yours:

1. **The failing check is not a code failure** — it's `2 Approvers for Bots` (Bot PR Review Gate). Every substantive check passes (unit tests 3.10–3.14 × linux/windows/macOS, Lint, CI Gate, codecov/patch). It needs **two human approvers**; I'm the author, so I can't be one.
2. **Security posture is still an open decision:** the default instance exposes *every* registered tool, and this executes model-authored code **in the host process with no sandbox**. `allowed_tools` limits exposure by name but is explicitly *not* a security boundary (code can reach the agent via object references). I kept the permissive default for parity with `bash`; whether that's the right shipping default is your call. This likely also wants `needs-api-review`.

<details>
<summary>MCP test evidence, review findings, and non-blocking follow-ups</summary>

**MCP probe before the fix** (real stdio server exposing `mcp_echo`, `mcp_add`, `dash-tool`, `dots.tool`, `ns__pfx-tool`, `mcp_boom`):

```
OK  simple MCP call        -> mcp_echo:hello
OK  MCP add                -> 42
OK  parallel gather        -> ['2', '4', 'mcp_echo:p']
OK  loop                   -> 10 11 12
OK  MCP tool error         -> caught: Tool 'mcp_boom' error: ... mcp exploded
OK  raw name via globals() -> dash:x
ERR hyphen via identifier  -> NameError: name 'dash_tool' is not defined
```
Namespace contained only non-identifier keys for those tools:
`['asyncio', 'dash-tool', 'dots.tool', 'mcp_add', 'mcp_boom', 'mcp_echo', 'ns__pfx-tool', 'print']`

**After the fix** — all of the above pass, plus `dash_tool`, `dots_tool`, `ns__pfx_tool` all resolve, raw names still work via `globals()`, and `gather` works across aliased + plain tools.

**Independent review results:**
- *Correctness* — APPROVE. Verified alias logic is order-independent (`sorted()` neutralizes set ordering; re-ran across `PYTHONHASHSEED` variation), a real tool can never be overwritten by an alias, 3-way collisions drop cleanly with no `KeyError`, and keyword aliases are structurally unreachable (a generated alias always contains `_`; no Python keyword does).
- *Adversarial* — APPROVE after ~20 attack scenarios: could not produce wrong-tool dispatch, an `allowed_tools` bypass via alias, or reserved-name shadowing (incl. a crafted `!-name-!` → `__name__` forgery attempt, correctly blocked).
- *Test quality* — found the print-capture gap above (fixed), plus a reserved-alias branch with no coverage (test added) and a weak MCP error assertion (strengthened).

**Non-blocking follow-ups I deliberately did *not* do** (happy to, say the word):
- When two tools from *different* MCP servers collide on one alias, the alias is silently dropped — adding server B can quietly make server A's tool uncallable from code (raw name still works). A stub that raises a clear "ambiguous alias, use `globals()['raw-name']`" error would beat the current `NameError`.
- Special-casing `NameError` in the error path to suggest the closest alias/raw name.
- `test_runs_mcp_tools_concurrently` proves `gather` *composes*, not that it *parallelizes* — proving real parallelism needs a deliberately slow tool and a wall-clock assertion.
- TypeScript parity: no TS equivalent exists today, so nothing to sync; a TS port can follow up.

</details>

🤖 I'm an AI agent and I wrote this PR, so please don't take the above as an approval — it's evidence for your review.

lizradway:
closing out since this is stale & is being being picked up by others for roadmap

PR #3206, opened 2026-07-12
fix(vercel): detect tool use from streamed content when finish_reason is non-tool

### LINKED ISSUE #3185: [BUG] Strands SDK TypeScript - VercelModel Does Not Detect Tool Use Blocks When Provider Reports Wrong stopReason
### Checks

### SDK Language

TypeScript

### Strands Version

1.4.0

### Language Runtime Version

Node js 22.0.0

### Operating System

26.5.1 (25F80)

### Installation Method

npm

### Steps to Reproduce

1. Install `@strands-agents/sdk` (1.4.0) and `ai-sdk-ollama` (latest).
2. Create a VercelModel using `ollama("llama3.1")` as the provider.
3. Define a tool (e.g., `search_songs`) and register it with the agent.
4. Send a prompt that should trigger the tool (e.g., "Search for jazz songs").
5. Observe: the model generates tool call blocks in the stream, but the agent terminates without executing them. No error is thrown.

### Expected Behavior

When the model generates tool use content blocks (`toolUseStart`, `toolUseInputDelta`) in the response stream, the VercelModel adapter should recognize that tool calling occurred and proceed to execute the tools,  regardless of what `stopReason` the provider reports. The agent loop should determine whether to enter the tool execution phase based on the actual content of the message (presence of tool use blocks), not solely on a metadata label that the provider may set incorrectly.

In other words: if the model asked to call a tool, the agent should call that tool and continue the conversation loop. The developer should never encounter a situation where valid tool call blocks are silently discarded because of a mismatched stop reason.

### Actual Behavior

When a Vercel AI SDK provider (such as `ai-sdk-ollama`) emits tool use content blocks but reports `stopReason: "endTurn"` instead of `stopReason: "toolUse"`, the Strands agent loop terminates without executing the tools. The VercelModel adapter trusts the provider's reported `stopReason` unconditionally, even when the stream itself contains `toolUseStart` and `toolUseInputDelta` events that clearly indicate tool calling occurred.

The agent silently ends without producing a response, no error is thrown, no tools are executed, and no indication is given that something went wrong. The model generated valid tool call blocks, but the agent loop decided not to act on them because it looked at the `stopReason` label rather than the actual content.

### Additional Context

This is a silent failure mode that is extremely difficult to debug. The agent simply stops with no output and no error. Developers must instrument the event stream to discover that tool use blocks were generated but ignored.

For any provider that doesn't correctly map its stop signal to the Vercel AI SDK's expected `tool_calls` finish reason, tool calling is completely broken. This affects:

- All Ollama models via `ai-sdk-ollama`
- Potentially any community or self-hosted provider that maps finish reasons differently
- Any provider where the stop reason mapping isn't explicitly tested with Strands

Since tool calling is a core capability of agent frameworks, having it silently fail based on a metadata label (rather than actual content inspection) is a critical gap in the adapter's robustness.

### Possible Solution

In the `streamAggregated` method of `VercelModel`, after the stream finishes and all content blocks have been collected, add a post-processing step that inspects the aggregated content. If any `toolUse` content block is present in the message, override `stopReason` to `"toolUse"`,  regardless of what the provider reported. This can be as simple as:

```typescript
const hasToolUse = contentBlocks.some(block => block.type === 'toolUse');
if (hasToolUse && stopReason !== 'toolUse') {
  console.warn(
    `Provider reported stopReason "${stopReason}" but tool use blocks were detected. Overriding to "toolUse".`
  );
  stopReason = 'toolUse';
}
```

This ensures the agent loop's decision to execute tools is driven by what the model actually produced, not by a metadata label that community providers may not set correctly. The warning gives visibility into the mismatch without breaking the flow.

### Related Issues

_No response_

maxmilian:
## Summary

`VercelModel` derived `stopReason` purely from the provider's finish reason via `mapFinishReason` (`strands-ts/src/models/vercel.ts`). Community providers such as [`ai-sdk-ollama`](https://github.com/nordwestt/ai-sdk-ollama) stream tool-call blocks but still report `finish_reason: "stop"`, which maps to `endTurn` — so the agent loop never entered tool execution and terminated silently without running the requested tools.

Fixes #3185 (validated; root cause matches the automated bug-verify).

## Change

In `stream()`, track whether any client tool use **completed** during the stream (a `tool-input-start` followed by its `tool-input-end`, or a standalone complete `tool-call`). When the terminating `ModelMessageStopEvent` would otherwise be `endTurn`, promote it to `toolUse`. This decides the stop reason from the **streamed content**, not solely from the provider's finish label — as the report framed it.

Scoped deliberately in two ways:

- **`endTurn` only**: a genuine `maxTokens` / `contentFiltered` finish that happens to carry a tool block is **not** masked. Note this intentionally diverges from the Google/OpenAI adapters, which promote unconditionally when tool calls were seen (even overriding `maxTokens`) — a `maxTokens` finish carrying a complete tool call executes the tool there but not here. Both behaviors are defensible; this adapter opts not to mask a genuine truncation. Flagged in review as a candidate for a follow-up to align the three adapters.
- **Completed blocks only**: a truncated client tool block (`tool-input-start` with no `tool-input-end` / complete `tool-call`) does not promote — the aggregator never builds a `ToolUseBlock` for it, so promoting would end the message with `stopReason: toolUse` and zero tool blocks, which the agent loop rejects as an invariant violation. The truncated shape keeps the `endTurn` mapping, exactly as on `main` (review Finding A).

Non-tool streams are unaffected, so `finish_reason: "stop"` with no tool use still maps to `endTurn`.

## Testing

All of the below re-run on the current head (post-`main`-merge + Finding A commit):

- Full TS unit suite: **3771 passed, 0 failed** (`npm test`).
- `tsc --project src/tsconfig.json` — clean. `npm run format:check` — clean. `npx eslint` on both changed files — clean. (Repo-wide `npm run lint` / `npm run type-check` currently carry pre-existing failures in unrelated files merged from `main` — e.g. the #3219 interventions package — reproducible without this change.)
- Added two unit tests (complete `tool-call` and incremental `tool-input-*` with `tool-input-end`, each with `finish_reason: "stop"`); both fail on `main` and pass with this change. The existing no-tool `stop → endTurn` mapping test guards against over-promotion, and a truncated-block test (`tool-input-start` + delta, no close, `finish: stop`) pins the no-promotion behavior from review Finding A — it fails on the pre-review head and passes now.
- I could not exercise a live `ai-sdk-ollama` / Ollama server locally, so — like the sandbox bug-verify — I reproduced the exact stream shape with a mocked `doStream`.

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/strands-agents/harness-sdk/blob/main/CONTRIBUTING.md) doc
- [x] New and existing unit tests pass locally with my changes
- [x] I have reviewed and understand every line of this change (including any AI-assisted portions) and can explain why it works


mkmeral:
@strandly-the-agent can you review the PR?

strandly-the-agent (commented):
**Verdict: approve** — correct root cause, tightly scoped, well-tested. One non-blocking edge case (🟡) and one parity question (❓) worth a deliberate call before merge. Nothing here blocks.

The fix mirrors the intent of the existing Google/OpenAI adapters ("decide tool execution from the streamed content, not the provider's finish label") and the `endTurn`-only scoping is a reasonable, documented choice. I ran it end to end:

| Check | Result |
|---|---|
| Branch / SHA reviewed | `pr3206` @ `2350f09` |
| `tsc --project src/tsconfig.json` | ✅ clean (exit 0)* |
| `vitest --project unit-node vercel.test.ts` | ✅ **62/62 pass** |
| Regression (revert only the `vercel.ts` fix) | ✅ the 2 promotion tests **fail on `main`**, pass with the fix; the 2 negative guards pass either way — tests genuinely cover the change, not tautological |
| Finding A repro | ✅ reproduced at the model layer |

<sub>*The full `npm run type-check` fails, but only on pre-existing, unrelated errors in `test/integ/vended-interventions/steering/` (the interventions package from #3219) — nothing from this PR.</sub>

<details>
<summary>🟡 Finding A (should-fix, non-blocking) — a <em>truncated</em> client tool block + <code>finish: stop</code> now throws where it previously ended cleanly</summary>

`sawToolUse` is set the moment a client `tool-input-start` arrives (`strands-ts/src/models/vercel.ts:174-180`), *before* the tool block is known to complete. If a provider opens a tool block and then finishes with `stop` (or `other`) **without ever closing it** (no `tool-input-end`, no complete `tool-call`), the terminal event is promoted to `toolUse` — but the aggregator only builds a `ToolUseBlock` on `modelContentBlockStopEvent` (`src/models/model.ts:397-425`), so the message ends up with **`stopReason: 'toolUse'` and zero tool blocks**. The agent loop treats that as an invariant violation and **throws** at `src/agent/agent.ts:2215` (`Model indicated toolUse but no tool use blocks found in message`).

Reproduced at the model layer — stream `[stream-start, tool-input-start(call_1), tool-input-delta('{"expr":'), finish:stop]`:
```
stopReason = toolUse
contentBlockStart events = 1   contentBlockStop events = 0   (→ 0 ToolUseBlocks → agent throws)
```
On `main` this same shape maps to `endTurn` and returns cleanly, so this PR *widens the trigger* for that throw from `finish_reason: tool-calls` to also include `stop`/`other`.

**How much does it matter?** The target scenario (#3185 — providers that stream *complete* tool calls but mislabel `finish` as `stop`) works perfectly; this only bites a *malformed/truncated* stream. But that's arguably the same class of buggy community provider this PR exists to defend against, so it's worth a conscious decision rather than an accident. Two directions: (a) accept it — a loud error on a truncated tool call isn't unreasonable; or (b) gate the promotion on a *completed* client tool block (set `sawToolUse` on `tool-input-end` / a complete `tool-call`, remembering the `providerExecuted` flag captured at `tool-input-start`) so a half-streamed block falls back to `endTurn`. Your call — I don't think it blocks the fix.
</details>

<details>
<summary>❓ Question — cross-adapter parity on the promotion condition</summary>

This adapter promotes **only** when the mapped reason is `endTurn`. The sibling adapters promote **unconditionally** when tool calls were seen:
- Google: `hasToolCalls ? 'toolUse' : ...` (`src/models/google/adapters.ts:502`) — overrides even `MAX_TOKENS`.
- OpenAI: `if (state.toolCalls.size > 0) stopReason = 'toolUse'` (`src/models/openai/responses-adapter.ts:545-547`) — also overrides `maxTokens`.

So a `maxTokens` finish carrying a complete tool call would execute the tool on Google/OpenAI but **not** on Vercel. Both behaviours are defensible (yours avoids masking a genuine truncation), and I lean toward yours being the *more* correct one — but the divergence is worth a line in the PR description, or a follow-up to align the three. Not a blocker.
</details>

<details>
<summary>⚪ Nits (low-confidence, no action needed)</summary>

- The `continue` that de-dupes a complete `tool-call` (`vercel.ts:183-185`) sits *before* the `tool-call` `providerExecuted` check (`:186-188`), so if `tool-input-start` and its later `tool-call` ever *disagreed* on `providerExecuted`, only the start's value would count. Defensive only — the bundled OpenAI provider sets `providerExecuted` consistently on both events, so no shipped provider hits this.
- `sawToolUse` latches for the whole stream, so a (malformed) stream with two `finish` parts would promote both terminal stop events. Purely theoretical — a well-formed stream has exactly one `finish`.
</details>

Nice, focused fix with a clear TDD story and honest testing notes. Flagging as solid work for a human maintainer to give the final approve.

maxmilian:
Thanks @mkmeral for routing the review, and @strandly-the-agent for the unusually thorough pass — the truncated-stream repro in Finding A was exactly right.

Made the two deliberate calls:

**🟡 Finding A → fixed (option b).** Promotion is now gated on a *completed* client tool block: `tool-input-start` records the `providerExecuted` flag per id, and `sawToolUse` is only set on the matching `tool-input-end` (or a standalone complete `tool-call`). A half-streamed block falls back to `endTurn` exactly as on `main` — I agree a loud error was defensible, but throwing in the agent loop over a malformed stream from the same class of buggy provider this PR defends against felt worse than degrading gracefully. Pinned by a new test using your repro shape (`tool-input-start` + delta, `finish: stop`) — red on the previous head, green now. Full suite 3771/3771 on the new head, `tsc` on `src/tsconfig.json` clean, changed files eslint-clean.

**❓ Parity → documented, deferring alignment.** Added a note to the PR description: this adapter intentionally does not override `maxTokens` the way Google/OpenAI do, to avoid masking a genuine truncation. I'd rather keep this PR scoped to #3185 and let a follow-up align the three adapters once maintainers pick a direction — happy to open that issue if wanted.

⚪ nit 1 is also addressed as a side effect of the Finding A change (the flag captured at `tool-input-start` is now explicitly what drives the promotion, per id). nit 2 (double-`finish` latch) left as-is — agreed it's theoretical.

mkmeral:
@strandly-the-agent  can you re-review

strandly-the-agent (approved):
**Approved at `eeab831`** — the merge-only refresh preserves the verified fix; no findings remain.

✅ Prior Finding A is fixed and reverified: truncated tool input stays `endTurn`; completed client tool calls promote to `toolUse`.
✅ The two PR files are byte-identical to the reviewed `15f30be`; `--remerge-diff` is empty and the feature patch against current `main` is unchanged.
✅ Exact-head checks: src TypeScript clean, ESLint/Prettier clean, `git diff --check` clean, focused Node suite **63/63 passed** with no type errors.
⚠️ I installed Chromium but couldn't launch the browser suite because this sandbox lacks Playwright's native host libraries; no browser assertion ran or failed.
⏳ Code verdict is approve; merge still depends on the currently waiting CI/integration gates. No completed functional CI failures are visible.

mkmeral (approved):

mkmeral:
Thank you for your contribution @maxmilian  :)

PR #3208, opened 2026-07-12
fix(ts): complete aggregated streams without message start

### LINKED ISSUE #3189: [BUG] VercelModel streamAggregated Throws Error Despite Successful Response
### Checks

### SDK Language

TypeScript

### Strands Version

1.4.0

### Language Runtime Version

Node 22.0.0

### Operating System

26.5.1 (25F80)

### Installation Method

npm

### Steps to Reproduce

1. Install `@strands-agents/sdk` (1.4.0) and `ai-sdk-ollama` (latest).
2. Create a VercelModel using `ollama("llama3.1")` as the provider.
3. Create an Agent with a simple system prompt and no tools.
4. Call `agent.invoke("Hello")`.
5. Observe: the response streams correctly to the console, then `ModelError: Stream ended without completing a message` is thrown.

### Expected Behavior

The VercelModel adapter should complete successfully when the model generates a valid response, even if the underlying provider does not emit a `modelMessageStartEvent` or reports a non-standard `finish_reason`. If the stream delivered content blocks (text, tool use), the adapter should treat the message as complete with `role: "assistant"` by default. The existing fallback logic (defaulting to `endTurn`) should be authoritative — if a fallback was applied, subsequent validation must not contradict it by throwing an error.

Developers using community providers should get a working agent out of the box without needing to patch the event stream or wrap the model class.

### Actual Behavior

When using `VercelModel` with community providers like `ai-sdk-ollama`, the agent successfully generates and streams the complete response to the user, but then throws `ModelError: Stream ended without completing a message` during the post-stream aggregation phase. The error occurs because `streamAggregated()` in `model.ts` requires a `modelMessageStartEvent` to set `messageRole`, and if that event is never emitted by the underlying provider, the final validation fails — even though the response was delivered correctly.

Additionally, the VercelModel logs a warning (`unknown vercel finish reason, defaulting to endTurn`) indicating it already has a fallback path, but the subsequent validation in `streamAggregated` doesn't honor that fallback and throws anyway. This inconsistency between the warning-level fallback and the hard error makes the adapter brittle with non-standard providers.

### Additional Context

In my solution I create a custom wrappers that intercept and patch the event stream before it reaches `streamAggregated`. you can check this implementation here: https://github.com/hsaenzG/OpenSource-agents-demo/blob/main/typescript/src/ollama_model.ts

### Possible Solution

In `streamAggregated`, replace the hard validation that throws `ModelError` when `messageRole` is null with a defensive default:

```typescript
// After stream ends, before final validation:
if (!messageRole && contentBlocks.length > 0) {
  messageRole = 'assistant'; // Sensible default for any model response
}
```

Additionally, align the finish reason fallback so the warning-level default (`endTurn`) is treated as a valid completion signal downstream:

```typescript
// When finish_reason is unknown or undefined:
const effectiveStopReason = stopReason ?? 'endTurn';
// Use effectiveStopReason in all subsequent logic — no further validation needed
```

This makes the adapter resilient to providers that don't emit the full event lifecycle while still preserving correct behavior for providers that do.

### Related Issues

_No response_

gyx09212214-prog:
﻿## Summary
- Default `streamAggregated()` to an assistant role when a provider emits content blocks and a message stop event without an earlier `modelMessageStartEvent`.
- Preserve the existing hard failure for streams that end without both a completed message and aggregated content.
- Add a regression test for the missing `stream-start` / missing message-start path from #3189.

## Root cause
Some Vercel-compatible providers can stream valid content without a leading `stream-start` part, so `VercelModel` may never emit `modelMessageStartEvent`. Content blocks are still delivered, but `streamAggregated()` previously threw `Stream ended without completing a message` after the stream ended.

This PR intentionally keeps the change in the aggregation layer and does not change `VercelModel.stream()`. #3206 covers the separate #3185 tool-use stop-reason path.

## Validation
- `npm run test -w strands-ts -- src/models/__tests__/model.test.ts`
- `npm run test -w strands-ts -- src/models/__tests__/vercel.test.ts`
- `npm run build -w strands-ts`
- `npm run type-check -w strands-ts`
- `npm run format:check -w strands-ts -- src/models/model.ts src/models/__tests__/model.test.ts`
- `npm run lint -w strands-ts -- src/models/model.ts src/models/__tests__/model.test.ts`

Fixes #3189

### DIFF
--- a/strands-ts/src/models/__tests__/model.test.ts
+++ b/strands-ts/src/models/__tests__/model.test.ts
@@ -122,6 +122,43 @@ describe('Model', () => {
           'Model reached maximum token limit. This is an unrecoverable state that requires intervention.'
         )
       })
+
+      it('defaults to assistant when stream content arrives without a message start event', async () => {
+        const provider = new TestModelProvider(async function* () {
+          yield { type: 'modelContentBlockStartEvent' }
+          yield {
+            type: 'modelContentBlockDeltaEvent',
+            delta: { type: 'textDelta', text: 'Hello from a provider without stream-start' },
+          }
+          yield { type: 'modelContentBlockStopEvent' }
+          yield { type: 'modelMessageStopEvent', stopReason: 'endTurn' }
+        })
+
+        const messages = [new Message({ role: 'user', content: [new TextBlock('Hi')] })]
+
+        const { items, result } = await collectGenerator(provider.streamAggregated(messages))
+
+        expect(items).toEqual([
+          { type: 'modelContentBlockStartEvent' },
+          {
+            type: 'modelContentBlockDeltaEvent',
+            delta: { type: 'textDelta', text: 'Hello from a provider without stream-start' },
+          },
+          { type: 'modelContentBlockStopEvent' },
+          { type: 'textBlock', text: 'Hello from a provider without stream-start' },
+          { type: 'modelMessageStopEvent', stopReason: 'endTurn' },
+        ])
+        expect(result).toEqual({
+          message: {
+            type: 'message',
+            role: 'assistant',
+            content: [{ type: 'textBlock', text: 'Hello from a provider without stream-start' }],
+            trackingId: anyTrackingId,
+          },
+          stopReason: 'endTurn',
+          metadata: undefined,
+        })
+      })
     })
 
     describe('when streaming multiple text blocks', () => {
--- a/strands-ts/src/models/model.ts
+++ b/strands-ts/src/models/model.ts
@@ -435,12 +435,12 @@ export abstract class Model<T extends BaseModelConfig = BaseModelConfig> {
 
           case 'modelMessageStopEvent':
             // Store message and stop reason
-            if (messageRole) {
+            if (messageRole || contentBlocks.length > 0) {
               const filtered = contentBlocks.filter(
                 (block) => !(block instanceof TextBlock && block.text.trim() === '')
               )
               stoppedMessage = new Message({
-                role: messageRole,
+                role: messageRole ?? 'assistant',
                 content: filtered,
               })
               finalStopReason = event.stopReason!


gautamsirdeshmukh:
@strandly-the-agent please review this proposed fix. Reproduce the bug that was raised to ensure it is valid, reproduce and test the fix itself to ensure it is completely sufficient and does not introduce any new risks or breaking changes.

strandly-the-agent:
✅ **APPROVED** — Fix correctly handles Vercel-compatible providers that omit `modelMessageStartEvent`; test validates new behavior; no regressions (3,683 tests pass).

## Evidence Summary

**Bug reproduction (base SHA 941d5251):**
Vercel-compatible providers emit content blocks without initial `modelMessageStartEvent` → `messageRole` stays `null` → old condition `if (messageRole)` fails → `stoppedMessage` never set → throws "Stream ended without completing a message".

**Fix validation (head SHA 87a44c4e):**
- ✅ All 36 model.test.ts tests pass (including new regression test)
- ✅ All 3,683 tests in full suite pass — zero regressions  
- ✅ New test "defaults to assistant when stream content arrives without a message start event" explicitly validates the fix

## Code Review

**Correctness (model.ts:438–443)**
- Condition: `if (messageRole)` → `if (messageRole || contentBlocks.length > 0)` ✓
  - Correctly catches providers that emit content without initial event
  - Still correctly throws for truly empty streams (no start, no content)
- Role default: `messageRole ?? 'assistant'` ✓
  - Semantically correct: `streamAggregated` is a model method, so emitted content is from assistant
  - Consistent with redaction handler (line 465) which also defaults to 'assistant'
  - Type system allows only `'user' | 'assistant'` roles

**Test Quality (model.test.ts:125–163)**
- Explicitly reproduces Vercel scenario (omitted `modelMessageStartEvent`)
- Validates passthrough events emitted correctly
- **Core assertion (line 154):** `role: 'assistant'` confirms the fix works
- Test fails on old code (undefined or error thrown) ✓
- Integrates cleanly with existing test patterns

**Scope & Risk**
- Diff: 2 lines logic + 37 lines test
- Surface: Internal streaming aggregator only; no API surface changes
- Risk: LOW — defensive defaults, edge-case handling, protected by regression test
- Breaking changes: None

## Downstream Validation

Line 476 validates `stoppedMessage` non-null before returning; consumers always see valid Message state. Metadata handling (lines 477–489) correctly propagates when present, omits when absent.

---

**Ready to ship.** The fix is minimal, grounded, and addresses a real compatibility issue with well-justified defaults.

PR #3210, opened 2026-07-13
fix(site): mobile header nav dropdown pushed off screen

yonib05:
## Description

On mobile, the header's secondary nav dropdown (the "API Reference ▾" pill that holds Home / Docs / Examples / API Reference / Blog / etc. plus the language toggle) is pushed off the right edge of the screen and rendered underneath the fixed sidebar menu button, making it invisible and untappable on common phone widths (390px and below).

Root cause: `.title-wrapper` forces a 180px slot for the logo on all viewports. The mobile logo only needs ~117px, so the extra 63px pushes everything in the flex row to the right until the nav pill overflows under the fixed hamburger. The 180px reservation exists to prevent layout shift while the logo loads, which only matters for the desktop grid layout — so it is now scoped to the desktop (`min-width: 50rem`) media query, matching the header's existing mobile/desktop breakpoint; on mobile the logo takes its natural width.

As a defensive layer, the nav pill can now shrink inside the flex row (`min-width: 0` on `.mobile-nav`, `max-width: 100%` on the toggle) and its current-section label truncates with an ellipsis (self-contained `nowrap`/`hidden`/`ellipsis` on `.mobile-nav-current`), so a future longer nav label can't reintroduce the overflow.

The fix is CSS-only and confined to `Header.astro`. (An earlier revision of this PR also collapsed the pagefind search trigger to an icon-only button on mobile; that became unnecessary when the search revamp was reverted in #3204 — Starlight's default search already renders as a compact icon button on mobile — so those changes were dropped when merging main.)

## Related Issues

N/A — reproducible on https://strandsagents.com/docs/api/python/strands.agent.a2a_agent/ at 390px-wide viewports and below.

## Type of Change

Bug fix

## Testing

- Verified with Playwright against the production build (`astro build` + `astro preview`) at 430/390/375/360/320px widths: the nav pill sits fully on screen, never overlaps the sidebar menu button, and `document.elementFromPoint` at its center hits the pill (previously it hit the menu button or an SVG overlaying it).
- Verified the dropdown opens and all entries (language toggle, section links, GitHub links) are visible and tappable at 390px and 320px.
- Verified the sidebar hamburger still opens the docs sidebar.
- Verified desktop (1440px) is unchanged: 180px logo slot, nav tabs row intact.
- Verified `width: max-content` vs no width declaration are equivalent for a flex item (117px either way), so the mobile rule is purely the absence of the desktop constraint.
- `npm run typecheck` clean; full vitest suite passes; prettier clean.

## Checklist

----


yonib05:
Re the two inline notes on `Search.astro`: both are now moot — upstream reverted the pagefind Component UI search revamp in #3204, so after merging main this PR no longer touches `Search.astro` at all (the file is gone; Starlight's default search is back and already renders as a compact icon button on mobile).

The remaining diff is `Header.astro` only: the desktop-scoped 180px logo slot and the mobile-nav shrink/ellipsis defenses. Re-verified against the rebuilt production preview at 430/390/375/360/320px — the nav pill stays on screen and tappable at every width, and the default search button (32px) fits cleanly.

If the search revamp comes back later, the breakpoint-complement (`@media not (min-width: 50rem)`) and the `.pf-trigger-shortcut` wrapper comment from your notes should go with it — flagged that on my side.

mehtarac (approved):

PR #3211, opened 2026-07-13
fix(tools): load tools moved into a watched directory (#264)

### LINKED ISSUE #264: [BUG] Hot-reloading fails for first tool when ./tools directory is empty at Agent initialization
### Checks

### Strands Version

0.1.18

### Python Version

3.13.3

### Operating System

macOS (Darwin with Clang 17.0.0)

### Installation Method

git clone

### Steps to Reproduce

Hot-reloading of tools from the `./tools` directory fails when the directory is empty during Agent initialization. The first tool added to an empty `./tools` directory is not automatically loaded, breaking the expected hot-reload functionality.

1. **Setup the environment:**
   ```bash
   git clone git@github.com:strands-agents/sdk-python.git /tmp/sdk-python && cd /tmp/sdk-python
   python3 -m venv .venv && source .venv/bin/activate && pip3 install -e ".[dev]"
   ```

2. **Ensure ./tools directory exists but is empty:**
   ```bash
   mkdir -p tools
   # Ensure tools directory is empty
   rm -f tools/*.py
   ```

3. **Create a simple agent in Python REPL:**
   ```python
   from strands import Agent
   agent = Agent()
   print(agent.tool_config)  # Shows: {'tools': [], 'toolChoice': {'auto': {}}}
   ```

4. **Add a tool to ./tools directory while REPL is running:**
   ```python
   # Create tools/basic_calculator.py with @tool decorator
   # (see full tool code in Additional Context section)
   ```

5. **Check if tool is loaded:**
   ```python
   print(agent.tool_config)  # Still shows: {'tools': [], 'toolChoice': {'auto': {}}}
   print(agent.tool_names)   # Shows: []
   ```

### Expected Behavior

When a tool is added to the `./tools` directory, it should be automatically hot-reloaded and available in:
- `agent.tool_config` - should include the new tool specification
- `agent.tool_names` - should include the tool name in the list
- `agent.tool.tool_name()` - should be callable

This should work regardless of whether the `./tools` directory was empty when the Agent was initialized.

### Actual Behavior

- **First tool addition**: Tool is NOT automatically loaded when added to an initially empty `./tools` directory
- **Subsequent tool additions**: Once at least one tool exists, additional tools ARE automatically hot-reloaded
- **Workaround**: Restarting the Python process loads the tool correctly

## Code Analysis

Based on examination of the codebase, the issue appears to be in the `ToolWatcher` initialization logic:

1. **In `Agent.__init__()`**: The `ToolWatcher` is only initialized if `load_tools_from_directory=True`
2. **In `ToolRegistry.initialize_tools()`**: Tool discovery happens, but if no tools exist, the watcher may not be properly set up
3. **In `ToolWatcher.start()`**: The file system watcher is configured, but may not monitor empty directories effectively

## Test Case Evidence

```python
# Case 1: Empty tools directory - FAILS
agent = Agent()
# Add tool to empty ./tools/
# Result: tool_config remains empty

# Case 2: Pre-existing tool - WORKS  
agent = Agent()  # with existing tool in ./tools/
# Add another tool
# Result: new tool appears in tool_config

# Case 3: Restart after adding to empty directory - WORKS
# Exit Python, add tool, restart Python
agent = Agent()
# Result: tool_config includes the tool
```

## Proposed Solution Areas

The issue likely resides in one or more of these areas:

1. **`ToolWatcher` initialization**: May need to properly handle empty directories
2. **File system monitoring**: The watchdog observer might not be monitoring empty directories
3. **Tool discovery timing**: Initial discovery vs. runtime discovery synchronization

## Additional Context

**Sample tool used for testing:**
```python
# tools/basic_calculator.py
from strands import tool

@tool
def basic_calculator(expression: str) -> dict:
    """
    A very simple calculator for basic arithmetic only.
    
    Supports: +, -, *, /, and parentheses
    
    Args:
        expression: Simple math expression (e.g., "2 + 3", "10 / 2", "(5 + 3) * 2")
        
    Returns:
        Dictionary with the calculation result
    """
    try:
        # Only allow basic math characters and numbers
        allowed = set('0123456789+-*/().')
        if not all(c in allowed or c.isspace() for c in expression):
            raise ValueError("Only basic math operations allowed: +, -, *, /, ()")
        
        result = eval(expression)
        
        return {
            "status": "success",
            "content": [{"text": f"{expression} = {result}"}]
        }
        
    except ZeroDivisionError:
        return {
            "status": "error",
            "content": [{"text": "Cannot divide by zero"}]
        }
    except:
        return {
            "status": "error",
            "content": [{"text": "Invalid math expression"}]
        }
```

## Impact

This bug affects the developer experience for:
- New users setting up their first Strands agent
- Development workflows where tools are added incrementally
- Any scenario where the `./tools` directory starts empty

## Related Files

- `src/strands/agent/agent.py` - Agent initialization
- `src/strands/tools/registry.py` - Tool discovery and registration
- `src/strands/tools/watcher.py` - File system monitoring for hot-reload

## Verification

This issue can be consistently reproduced with the steps above. The hot-reload functionality works correctly in all cases except when the initial `./tools` directory is empty during Agent initialization.

### Possible Solution

_No response_

### Related Issues

_No response_

maxmilian:
## Description

The tool hot-reload watcher did not pick up a tool file that was moved or copied into a watched directory. This was most visible when the `./tools` directory was empty at `Agent` initialization: the first tool dropped in never loaded.

Root cause: only the `MasterChangeHandler` is scheduled with the watchdog `Observer` (in `ToolWatcher.start()`), and it implemented only `on_modified`. A file *written in place* inside the directory raises `on_modified` as well as `on_created`, so that path already worked — but a file that *arrives* in the directory, moved or renamed in from elsewhere, raises `on_created` alone, hit the base-class no-op, and was dropped until the file was later edited in place.

An earlier revision of this description said watchdog emits `on_created` *instead of* `on_modified` for any file that did not previously exist. That is wrong on both Linux and macOS, and @ebarkhordar caught it — thank you. The measurements below are the corrected version.

Fix: add `MasterChangeHandler.on_created`, routing through the existing `on_modified` reload path. `reload_tool` loads a new file the same as a changed one, and is safe to call again when a follow-up `on_modified` fires for the same file (hot-reload registration is idempotent), so no double-registration issue.

## Related Issues

Closes #264

## Documentation PR

No docs change needed — this restores documented hot-reload behavior with no API change.

## Type of Change

Bug fix

## Testing

Added a regression test (`test_master_handler_on_created_reloads_new_tool`) that drives the actual production path — the scheduled master handler — and asserts an `on_created` event for a `.py` file triggers `reload_tool`. It fails before the fix (`reload_tool` called 0 times, hitting the base-class no-op) and passes after.

Behaviour measured end-to-end outside the suite, against a real `Observer` scheduling the real `MasterChangeHandler` with `ToolRegistry.reload_tool` stubbed to count calls — macOS 26.5.2, watchdog 6.0.0, `FSEventsObserver`, base `8c3b2102` against this branch:

| how the tool file arrives in the watched dir | before | after |
|---|---|---|
| written in place (`open('w')`, `touch`, `cp`, write + `fsync`) | 1 | 2 |
| **moved / renamed in from outside the directory** | **0** | **1** |
| atomic save: temp file in the same dir, renamed to a new name | 0 | 0 |
| atomic save: temp file in the same dir, renamed over an existing tool | 1 | 1 |

The second row is the bug. Writing in place emits `created` **and** `modified` for the file, which is why the first row already worked before this change; moving a file in emits `created` alone.

Two things that row four hints at, both deliberately out of scope here:

- The reload on the write-in-place path now happens twice (also measured on Linux by @ebarkhordar, ~2ms apart). `reload_tool` is idempotent, so this is redundant work rather than a correctness problem. Happy to add a short same-path debounce if you'd prefer it, but it grows this well past a one-line fix.
- A same-directory atomic save emits `FileMovedEvent`, whose `dest_path` holds the tool path, and no handler here implements `on_moved`. That gap exists before and after this change. It needs `dest_path` rather than `src_path` (which is the temp file, and doesn't even pass the `.py` filter), and macOS and Linux differ on the last row above, so it wants its own tests — happy to open a separate PR for it.

Checks run on this branch:

- `pytest tests/strands/tools/` — 561 passed, no regressions
- `ruff format --check` and `ruff check` — clean; `mypy` (strict) — clean

Note: I don't have `hatch` locally, so I ran the equivalent checks directly (ruff format/lint, mypy strict, pytest) rather than `hatch run prepare`.

- [x] I ran the equivalent of `hatch run prepare` (ruff format + ruff check + mypy + pytest)

## Checklist

----

### DIFF
--- a/strands-py/src/strands/tools/watcher.py
+++ b/strands-py/src/strands/tools/watcher.py
@@ -95,6 +95,20 @@ class ToolWatcher:
                         except Exception as e:
                             logger.error("exception=<%s> | handler error", str(e))
 
+        def on_created(self, event: Any) -> None:
+            """Reload tools when a new file appears in a watched directory.
+
+            This master handler is the only one scheduled with the Observer, so it must handle on_created too. A file
+            written in place usually raises on_modified as well, which the handler above already catches, but a file
+            moved or renamed into the directory raises on_created alone and would otherwise be dropped - for example
+            the first tool copied into an empty tools directory. reload_tool loads a new file the same as a changed
+            one and is safe to call again for a follow-up on_modified, so route creations through on_modified.
+
+            Args:
+                event: The file system event that triggered this handler.
+            """
+            self.on_modified(event)
+
     def start(self) -> None:
         """Start watching all tools directories for changes."""
         # Initialize shared observer if not already done
--- a/strands-py/tests/strands/tools/test_watcher.py
+++ b/strands-py/tests/strands/tools/test_watcher.py
@@ -96,3 +96,26 @@ def test_on_modified_error_handling(mock_reload_tool):
 
     # Verify that reload_tool was called
     mock_reload_tool.assert_called_once_with("test_tool")
+
+
+@patch.object(ToolRegistry, "reload_tool")
+def test_master_handler_on_created_reloads_new_tool(mock_reload_tool):
+    """Regression for issue #264: the master handler is the one scheduled with the
+    Observer, so it must forward on_created events to the per-registry handlers.
+
+    Without this, a tool file moved into a watched directory - which raises on_created
+    alone, with no follow-up on_modified - is never loaded.
+    """
+    tool_registry = ToolRegistry()
+    watcher = ToolWatcher(tool_registry)
+
+    dir_path = "/path/to/empty_tools_dir"
+    ToolWatcher._registry_handlers.setdefault(dir_path, {})[id(tool_registry)] = watcher.tool_change_handler
+    master_handler = ToolWatcher.MasterChangeHandler(dir_path)
+
+    event = MagicMock()
+    event.src_path = f"{dir_path}/first_tool.py"
+
+    master_handler.on_created(event)
+
+    mock_reload_tool.assert_called_once_with("first_tool")


maxmilian:
Flagging a workflow-authorization state rather than asking for a review pass.

This PR is out of draft and every check that can run on a fork PR has passed. The two that show as pending — `Run integration tests` and `Trigger Strands Review` — are `pull_request_target` workflows sitting at `status: waiting`, which needs a maintainer to hit **Approve and run**. They have been in that state since July 13, so nothing is queued behind a failure; they simply have not been authorized to start.

Worth calling out because `Run integration tests` is where the real verification for this change lives, and until it runs neither of us has evidence beyond the unit tests in the diff. If the change looks wrong on its face I would rather hear that and close it than have it hold a slot — no objection either way.

ebarkhordar (commented):
I ran the real `MasterChangeHandler` against a real `Observer` on Linux (python:3.12-slim, watchdog 6.0.0), base `8c3b2102` against head `c1ec47d4`, with `ToolRegistry.reload_tool` stubbed to record each call and the file contents at that moment.

**#264 does not reproduce on inotify.** Dropping `first_tool.py` into an empty watched directory at base calls `reload_tool("first_tool")` once, with the complete file, because inotify emits `IN_CREATE` and then `IN_MODIFY` for the same path and the existing `on_modified` already catches the second one. #264 is filed against macOS, where FSEvents coalesces them, so the change looks right to me, but "watchdog emits on_created (not on_modified) for a file that did not previously exist" is not true on Linux and the description reads as though it were. On Linux head now calls `reload_tool` twice per creation (2ms apart here), which your idempotency note covers. This is about the wording, not the code.

**`on_moved` has the same hole.** Writing a temp file and renaming it over the tool path, which is how any atomic-save editor writes, emits `FileMovedEvent` carrying the tool path in `dest_path`, and neither `on_created` nor `on_modified` sees it. Measured zero `reload_tool` calls at base and zero at head, both for a rename over an existing tool and for a rename to a new name. Your own reason for adding `on_created`, that the master handler is the only one scheduled with the Observer, applies to it unchanged. Fine as a separate PR if you would rather keep this one on #264.

I have no macOS to test on, so the FSEvents half rests on the reporter's video, not on anything I ran.

maxmilian:
Thanks for actually running this — you're right, and the wording was worse than you could see from Linux alone. Fixed in `1cc8a33`; the code is unchanged, everything I touched was the description of it.

I re-ran your setup on the platform you didn't have: macOS 26.5.2, watchdog 6.0.0, `FSEventsObserver`, a real `Observer` scheduling the real `MasterChangeHandler`, `reload_tool` stubbed to count calls, base `8c3b2102` against the branch (base reproduced by deleting the `on_created` override at runtime, so it's the same object graph either way).

| how the file arrives | base | head |
|---|---|---|
| written in place in the watched dir (`open('w')`, `touch`, `cp`, write+`fsync`) | 1 | 2 |
| **moved/renamed in from outside the watched dir** | **0** | **1** |
| atomic save: temp in the same dir, renamed to a new name | 0 | 0 |
| atomic save: temp in the same dir, renamed over an existing tool | 1 | 1 |

Raw events, for the record: writing in place emits `created` **and** `modified` for the file, so FSEvents doesn't coalesce them the way my description claimed — the sentence was wrong on macOS too, not just on inotify. Moving a file in emits `created` alone. A same-directory rename emits `moved` with the tool path in `dest_path` (macOS adds a `modified` on the destination when it lands on an existing file, which is why that last row is 1 here and 0 on your Linux run — the two platforms genuinely differ there).

So the fix still earns its place, but for a narrower reason than I wrote: the gap is files that *arrive* rather than files that are *written*. Dropping a tool into `./tools` by moving or copying it in from elsewhere is exactly the reported action, and at base that reload never happens. I've reworded the PR title, the description, and the `on_created` docstring to say that instead — the docstring is the part that bothered me most about your note, since the wrong sentence had made it into the code itself.

The double call on the write path is real and I'd rather name it than bury it: on macOS head calls `reload_tool` twice, matching your 2ms-apart Linux measurement. `reload_tool` is idempotent so it's redundant work rather than a correctness problem, but if maintainers would prefer a short same-path debounce I'm happy to add one — I left it out because it grows this well past a one-line fix.

On `on_moved`: confirmed, and there's a wrinkle worth flagging for whoever picks it up. Forwarding it to `on_modified` as-is won't work, because `on_modified` reads `event.src_path`, which for a `FileMovedEvent` is the *old* path — the `.tmp` file, which doesn't even pass the `.py` filter. It needs `dest_path`, and given the platform split in the last row above it deserves its own tests. Agreed it belongs in a separate PR; I'll keep this one on #264.

`pytest tests/strands/tools/` is 561 passed on `1cc8a33`, with ruff format/check and mypy strict clean.

Issue #1216, opened 2025-11-19
[FEATURE] Add LiteLLM Cost Tracking to Agent Results


iyamabhishek:
## Summary
Add support for returning LiteLLM cost data as part of `AgentResult` metadata when using `LiteLLMModel`.

## Motivation
Currently, Strands agents track token usage via `accumulated_usage`, but there's no direct way to get the dollar cost of model invocations. LiteLLM provides a `completion_cost()` function that calculates costs based on token usage and model pricing, but this isn't exposed through the Strands SDK.

## Proposed Solution
Add a pass-through function in Strands' LiteLLM model provider (`strands/models/litellm.py`) that:

1. Transforms Strands agent result to LiteLLM response format
2. Calls `litellm.completion_cost()` to calculate the cost
3. Includes the cost in `AgentResult` metadata

## Example Usage
```python
from strands import Agent
from strands.models.litellm import LiteLLMModel

model = LiteLLMModel(
    model_id="anthropic/claude-3-7-sonnet-20250219"
)

agent = Agent(model=model)
response = agent("What is 2+2")

# Access cost from response metadata
print(f"Cost: ${response.metadata.cost:.10f}")
```

## Implementation Notes
The `completion_cost` function in LiteLLM has provider-specific logic for each model, so this would leverage existing LiteLLM functionality rather than reimplementing cost calculation.

### Use Case

Users need cost tracking for:
- Budget monitoring and cost optimization
- Per-request cost analysis
- Comparing costs across different models
- Production cost tracking and reporting

liuyuchen777:
Hi @mkmeral @yonib05, I am fan of Strand and want to contribute after spending several months with it. I came across this issue and thought it was nicely sized for a first PR, so I've drafted a proposed solution.

One thing I want to clarify, the issue proposed reading cost from `response.metadata.cost`, but `AgentResult` doesn't actually have a metadata field today. So I put the cost on the existing `Usage` type as an optional totalCost, surfaced through a new `AgentResult.cost`. I draft a PR #3212 for disucssion.

I think a dedicated `CostMetrics` type including multiple cost mmetrics would be better, users will be interested in the input vs. output cost split, and even cached vs. non-cached cost. The reason I didn't go that route is that `litellm.completion_cost` only estimates a single total.

grokoloko:
[...]

PR #1911, opened 2026-03-15
feat(litellm): add cost tracking to agent results

stefanoamorelli:
## Description

> [!TIP]
> Follows conventional commits. Better reviewed commit-by-commit.

Exposes LiteLLM's `cost_per_token()` data (in USD) through the Strands streaming pipeline so that accumulated cost is available in `EventLoopMetrics` when using `LiteLLMModel`.

The cost flows through: `LiteLLMModel.format_chunk()` → `MetadataEvent` → `process_stream()` → `ModelStopReason` → event loop → `EventLoopMetrics.accumulated_cost`.

Cost calculation is wrapped in `try/except` so models not yet mapped in LiteLLM's pricing database degrade gracefully (cost is simply omitted). The `ModelStopReason` tuple is extended from 4 to 5 elements; existing consumers use `*_` unpacking for forward-compatibility.

## Related Issues
Closes https://github.com/strands-agents/sdk-python/issues/1216

## Documentation PR
N/A

## Type of Change
- New feature

## Testing

- [x] I ran `hatch run prepare`
- [x] 2383 tests passing, 0 regressions from the cost tracking changes
- [x] 9 new tests covering cost calculation, accumulation, graceful failure, and cache token forwarding
- [x] Verify that the changes do not break functionality or introduce warnings in consuming repositories: agents-docs, agents-tools, agents-cli

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [x] I have updated the documentation accordingly
- [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3212, opened 2026-07-13
feat(litellm): add cost tracking to AgentResult

liuyuchen777:
## Description

This change makes cost a best-effort, USD-denominated metric. A result snapshots the cost of its own invocation, while event-loop metrics retain the lifetime total for budget monitoring. Calls that LiteLLM cannot price do not fail the agent; they leave cost untracked, so a partially priced invocation reports a lower bound.

```python
from strands import Agent
from strands.models.litellm import LiteLLMModel

agent = Agent(model=LiteLLMModel(model_id="openai/gpt-4o"))

first_result = agent("Summarize the benefits of unit testing.")
second_result = agent("Condense that summary to one sentence.")

print(first_result.cost)                 # First invocation only
print(second_result.cost)                # Second invocation only
print(second_result.metrics.total_cost)  # Both invocations
```

Cost remains `None` when no call in an invocation can be priced.

## Related Issues

#1216

## Documentation PR

No documentation changes included yet

## Type of Change

New feature

## Checklist


strands-py/src/strands/agent/agent_result.py
```diff
@@ -40,6 +40,17 @@ class AgentResult:
     structured_output: BaseModel | None = None
     checkpoint: Checkpoint | None = None
 
+    @property
+    def cost(self) -> float | None:
```
mkmeral, line 49: should we keep `cost` and `total_cost` terminology consistent? `cost` within agent result seems like cost for that agent invocation only

mkmeral (commented):
@strandly-the-agent review the PR. also check the name of this new var against the others in agent metrics  like tokens and so on. We should keep the naming consistent there

mkmeral:
Also related https://github.com/strands-agents/harness-sdk/pull/1911

strands-py/src/strands/agent/agent_result.py
```diff
@@ -40,6 +40,17 @@ class AgentResult:
...
+        """Best-effort estimated total cost in USD for this invocation.
+
+        Returns:
+            The accumulated cost across all model calls, or None when the model provider does not
+            report cost (currently only LiteLLM does). A returned value is a lower bound: calls
+            that could not be priced contribute 0.
+        """
+        return self.metrics.total_cost
```
strandly-the-agent: 🔴 **`cost` returns the agent's lifetime cost, not this invocation's — your inline comment above was right.**

The docstring says "for this invocation", but the chain is `AgentResult.cost` → `EventLoopMetrics.total_cost` → `accumulated_usage`, and `reset_usage_metrics()` (`telemetry/metrics.py:394-400`) never clears `accumulated_usage` — it only appends a new `AgentInvocation`. `result.metrics` **is** `agent.event_loop_metrics` (a live reference, not a snapshot), so old results mutate too.

Reachable on the SDK's own documented pattern — reusing one `Agent` across turns. Repro I ran (5 sequential calls, each costing exactly $0.01):

```
call   result.cost   ctx_size   per-inv tokens   lifetime tokens
1             0.01         10               15                15
2             0.02         10               15                30
3             0.03         10               15                45
4             0.04         10               15                60
5             0.05         10               15                75

sum(r.cost for r in results) = $0.25   vs   $0.05 actually spent
re-reading the earlier results later:  [0.05, 0.05, 0.05, 0.05, 0.05]
```

Note the controls: `context_size` stays 10 and per-invocation tokens stay 15 on every call, so this isn't a metrics-wide quirk — `cost` is the only "this invocation" property on `AgentResult` that returns a lifetime figure. At n=1 it's exact, which is why every test and doc example passes.

**The per-invocation value already exists, unexposed:** `latest_agent_invocation.usage["totalCost"]` was sitting at a flat `0.01` on all five calls — `update_usage` already accumulates `totalCost` into all three scopes. So this is a rescope, not new plumbing:

```suggestion
        return self.metrics.latest_invocation_cost
```

plus the scoped accessor in `telemetry/metrics.py`, named to match the existing `latest_context_size`:

```python
@property
def latest_invocation_cost(self) -> float | None:
    """Best-effort estimated cost in USD for the most recent agent invocation."""
    invocation = self.latest_agent_invocation
    return invocation.usage.get("totalCost") if invocation else None
```

I applied exactly this and re-ran the repro: `result.cost` goes flat at `0.01` for all five calls. **Two caveats from doing it:** `test_cost_delegates_to_metrics` and `test_cost_none_when_untracked` (`test_agent_result.py:391-403`) fail afterwards because they stub `metrics.total_cost` — they need the stubbed attribute renamed. And please add a regression test that invokes one agent twice and asserts isolation; nothing currently distinguishes the two semantics, which is why this shipped.

Keeping `EventLoopMetrics.total_cost` as-is is right — lifetime is the correct meaning *there*. It's only `AgentResult` that should be invocation-scoped.

strands-py/src/strands/models/litellm.py
```diff
@@ -177,6 +177,32 @@ def _extract_thought_signature(data: Any) -> str | None:
...
+        Returns:
+            The estimated total cost in USD, or ``None`` if the model could not be priced.
+        """
+        try:
+            response = litellm.ModelResponse(
+                usage=litellm.Usage(
+                    prompt_tokens=usage.prompt_tokens,
+                    completion_tokens=usage.completion_tokens,
+                    total_tokens=usage.total_tokens,
+                ),
```
strandly-the-agent, line 203: 🔴 **Cache tokens are dropped before pricing — cost overstated up to 7.4x, and understated on cache-write turns.**

`format_chunk` reads `cacheReadInputTokens`/`cacheWriteInputTokens` off `event["data"]` ~20 lines below this (`litellm.py:318-325`), but `_compute_cost` rebuilds a fresh `litellm.Usage` from only the three plain token counts, so the cache split never reaches `completion_cost`. Cache reads bill at ~10% of input and cache writes at ~125%, so the error runs **both ways** — which also falsifies "a returned value is a lower bound" in both docstrings (`agent_result.py:49-50`, `metrics.py:212-213`).

Zero opt-in required: OpenAI auto-caches any prompt ≥1024 tokens into this exact field. Measured against litellm's own pricing map:

```
anthropic/claude-sonnet-4   95% cache-read    $0.0330 vs $0.0074   4.49x OVER
anthropic/claude-sonnet-4   98% cache-read    $0.1530 vs $0.0207   7.39x OVER
openai/gpt-4o               auto-cache        $0.0070 vs $0.0048   1.47x OVER
openai/gpt-4.1              auto-cache        $0.0096 vs $0.0038   2.50x OVER
anthropic/claude-sonnet-4   cache-WRITE       $0.0330 vs $0.0401   0.82x UNDER
anthropic/claude-sonnet-4   no cache          $0.0330 vs $0.0330   1.00x  (control)
```

`usage` already carries both fields, so forwarding them is enough:

```suggestion
                usage=litellm.Usage(
                    prompt_tokens=usage.prompt_tokens,
                    completion_tokens=usage.completion_tokens,
                    total_tokens=usage.total_tokens,
                    prompt_tokens_details=getattr(usage, "prompt_tokens_details", None),
                    cache_creation_input_tokens=getattr(usage, "cache_creation_input_tokens", None),
                ),
```

Verified against a real `litellm.Usage`: cache-read drops $0.0330 → $0.0074 (the 4.49x correction) and cache-write rises $0.0330 → $0.0401. **One test-authoring gotcha I hit:** `prompt_tokens_details` only affects pricing when it's litellm's real `PromptTokensDetailsWrapper` — a duck-typed stand-in is silently ignored, and a bare `unittest.mock.Mock()` makes `getattr` return an auto-vivified Mock and raises `TypeError: float() argument must be...` inside litellm. So any unit test here needs those two attributes set explicitly (or `Mock(spec=...)`), not left to a bare Mock.

Worth reconsidering the whole approach too: LiteLLM already computes an authoritative cost at `response._hidden_params["response_cost"]`, which is also the only thing that works for proxy deployments (`completion_cost` can't price those at all). It's not a one-line swap — `format_chunk` only receives `response.usage`, not the response — but it would sidestep this bug class entirely rather than re-deriving pricing locally. Worth a follow-up at least?

strands-py/src/strands/models/litellm.py
```diff
@@ -177,6 +177,32 @@ def _extract_thought_signature(data: Any) -> str | None:
...
+            )
+            return litellm.completion_cost(completion_response=response, model=self.get_config()["model_id"])
```
strandly-the-agent: 🟡 **Unpriceable models now print red ANSI text to stdout on every single call, and it can't be silenced.**

The `except` correctly degrades to `None`, but `litellm.completion_cost()` `print()`s to **stdout** before raising when it can't resolve a provider — bypassing `logging` entirely, so no logger config can suppress it. Reproduced: an unresolvable `model_id` emits `"Provider List: https://docs.litellm.ai/docs/providers"` in red ANSI, **5 times**, per call, forever — there's no caching of the "can't price this model" verdict.

Who hits it: any proxy / self-hosted / fine-tuned `model_id` outside litellm's static pricing map — exactly the deployment shape LiteLLM exists for, and the same "proxy deployments" case your docstring already anticipates. Pre-PR `LiteLLMModel` never called `completion_cost`, so this noise is entirely new, and it corrupts stdout for CLI tools and servers emitting structured output.

Two options, and I'd lean toward the second:

```python
# 1. Local suppression (verified: redirect_stdout does capture it).
with contextlib.redirect_stdout(io.StringIO()):
    return litellm.completion_cost(completion_response=response, model=self.get_config()["model_id"])
```

Caveat worth weighing: `redirect_stdout` swaps `sys.stdout` process-wide and isn't task-safe, so under concurrent agents it could briefly swallow another coroutine's output. That may be worse than the spam.

```python
# 2. Remember the failure and stop retrying — the pattern already in this codebase.
```

`bedrock.py:895-903` hits an analogous unpriceable/unsupported case and caches it in `_SKIP_COUNT_TOKENS_MODELS` so it degrades once instead of every turn. Reusing that shape here fixes the spam *and* the pointless per-call retry, without touching global stdout. Your call which trade-off you prefer — flagging it because it's a real behavioral regression for stdout-sensitive consumers, not a style nit.

strandly-the-agent (changes requested):
**Verdict: request changes.** Two 🔴 grounded bugs in the cost math itself. On naming — your instinct was right, and it's a *value* bug, not just a word: two of the three names are fine, but `AgentResult.cost` returns the agent's lifetime cost, not this invocation's.

**Verified on `pr3212`@`d0001e11` vs `origin/main`@`198fd15b` — I re-ran all of this myself:**
- ✅ `pytest tests/strands` → **4106 passed** on head vs **4100** on main, with **40 identical** pre-existing collection errors both sides (missing optional deps) — so the delta is exactly your 6 new tests, zero regressions. `ruff` clean on the 4 changed files; `mypy` shows the same 41 pre-existing import errors on both branches, none in your files.
- 🔴 Both bugs below reproduced end to end, and **both fixes verified to actually work** (not just proposed).
- 🔴 Mutation check: all 5 mutations survive your 6 new tests — including replacing `_compute_cost` with `return None`, i.e. **deleting the feature passes the suite**. Both litellm cost tests `patch.object` `_compute_cost`, so the real pricing call never executes.

### Naming — your question, answered

| Name | Verdict | Take |
|---|---|---|
| `Usage["totalCost"]` | weak | *Proposal:* `totalCostUsd`. Sibling keys carry units 2-for-2 (`latencyMs`, `timeToFirstByteMs`, `types/event_loop.py:39-40`) — and unlike those, this key is **persisted into session files** (`event_loop.py:581`, `types/content.py:217`), so it can't be renamed later. |
| `EventLoopMetrics.total_cost` | ✅ **keep — don't churn it** | `total_*` already means "lifetime sum" in this class: `total_cycles`/`total_duration` (`metrics.py:421-422`), `ToolMetrics.total_time` (`:124`). `accumulated_*` is reserved for the two lifetime *containers* (`:203-204`), which is why #1911's `accumulated_cost` wasn't right either. |
| `AgentResult.cost` | 🔴 **name is right, value is wrong** | Bare noun matches its siblings `context_size`/`projected_context_size`. The docstring is what's false — see 🔴 1. |

**The rule that makes all three coherent: `result.*` = this invocation; `result.metrics.*` = the agent's whole life.** By that rule you keep every name except the `Usd` suffix, and fix one value.

### Questions
- ❓ **(blocking)** `Closes #1216` — the issue's stated use cases are "budget monitoring" and "per-request cost analysis", but cost is still dropped by OTel (`tracer.py:251-271` has no `totalCost` branch), by Swarm/Graph (`swarm.py:958-962`, `graph.py:1145-1151` sum only the 3 required keys), and on session rehydrate (`multiagent/base.py:333-341`). Those are all **pre-existing gaps, not yours** — but should this close #1216, or land as a step with the issue left open?
- ❓ **(blocking)** Since `totalCost` lands in users' session files permanently, worth a `strands-ts` owner's take on the key name first? `strands-ts/src/models/streaming.ts:493-518` has no cost field yet, and a later PY/TS name split would be unrenameable on both sides.
- ❓ *(non-blocking)* A companion `AgentResult.usage` at the same per-invocation scope, so `result.cost / result.usage["totalTokens"]` stays coherent? Right now the exact, universal metric (tokens) is a two-hop subscript while the best-effort LiteLLM-only one gets a first-class property — the ergonomics are inverted relative to reliability.
- ❓ *(non-blocking)* `Usage["pricedTokens"]` as a cheap completeness signal — today a partial total is byte-identical to a complete one, and `None` already means four different things (not tracked / unpriceable / dropped by roll-up / lost on rehydrate).
- ❓ *(non-blocking)* Docs: both checklist boxes are unchecked and `site/.../observability-evaluation/metrics.mdx` + `.../model-providers/litellm.mdx` don't mention cost. Worth including here, given it's a money-facing property?

<details><summary>Appendix — non-blocking, pre-existing, and things I checked that are <b>fine</b> (10)</summary>

**Refuted during adversarial testing — checked, not problems:**
- Float drift: 1M accumulations → 1.85e-11 relative error. Sub-microcent. Non-issue.
- Per-chunk double-counting: the metadata loop is unguarded and I *can* force 2x with a synthetic stream, but it's unreachable — the SDK defaults `stream_options={"include_usage": True}` (`openai.py:501`) and litellm's own `sent_stream_usage` guard emits usage at most once.
- `KeyError` inside the `except` handler (`litellm.py:203` re-reads `model_id`): unreachable, `format_request` raises first.
- None-vs-0.0 contract: **holds** end to end. `is not None` at the writer and `"totalCost" in source` at the accumulator both behave exactly as your PR description claims. A genuine `$0.00` survives as `0.0`, not `None`.
- `_accumulate_usage` in all four directions (target-has/source-hasn't, seeding, `0.0`, key-present-but-`None`): correct. Only the last would raise, and no writer can produce it.
- `logger.debug` for the degrade path: **right call**, matches repo precedent (`anthropic.py:444`, `gemini.py:508`, `llamacpp.py:559` all degrade `count_tokens` identically at debug).
- Session persistence of a float: clean. Both shipped repositories round-trip via `dataclasses.asdict` + `json`, and only `bytes` is special-cased; JSON handles floats natively.
- Backwards compatibility: purely additive optional key on a `total=False` TypedDict; nothing in the tree does exact-key-set comparison on `Usage`.
- Reusing the `Usage` dict instead of #1911's `ModelStopReason` tuple widening is, for what it's worth, clearly the better design — 130/-0 in 7 files vs 219/-16 in 14, and it's what makes per-invocation *and* per-cycle cost available for free. Worth saying explicitly since the alternative is on the table.

**Worth one follow-up issue (not this PR's problem — all true with this PR closed unmerged):** cost is dropped by everything downstream of the single-agent loop — `_parse_usage` on rehydrate (`multiagent/base.py:333-341`, which *does* handle cache tokens, so the precedent existed), and Swarm/Graph `_accumulate_metrics`. Swarm/Graph also hand each node the agent's *lifetime* `accumulated_usage` (`swarm.py:906`, `graph.py:1057`), already a token double-count. Happy to file it if useful.

**Other pre-existing, appendix-only:** `AgentResult.to_dict`/`from_dict` (`agent_result.py:104-142`) never serializes `metrics`, so `.cost` is always `None` after a round-trip (affects every metric, not cost). `Usage` isn't in `types/__init__.__all__` though `get_summary()` hands the raw dict out. `get_summary()`/`_metrics_summary_to_lines` won't print cost, so the human-readable dump omits the one number people watch.

**One deliberate-looking omission worth a comment in the code:** `tracer.py:251-271` correctly has no cost branch — OTel GenAI semconv has no cost attribute and `gen_ai.usage.*` is token-typed — but since that helper is hand-written per key, a future contributor may "complete the pattern" and put dollars into a token histogram. One line saying it's intentional would prevent that.

**Process:** three permanent public surfaces here (a TypedDict key, two properties) and no `api/needs-review` label — the gate in `team/API_BAR_RAISING.md:72` currently passes only because *no* api label is set. Reads as a **Moderate** change to me (new properties, not a new primitive), so an API reviewer's informal sign-off rather than a full meeting. CI on head is still `PENDING`.

</details>

---
*Automated review — solid work for a human to approve, not a gate. Everything above I ran on the branch; the repros and fix verifications are reproducible. Push back where I'm wrong.*

strands-py/src/strands/agent/agent_result.py
liuyuchen777: Update cost in AgentResult acutally reflect the cost for the invocation. As EventLoopMetrics is used across invocation, will need to has a `__post_init__` to copy `latest_invocation_cost` from EventLoopMetrics

strands-py/src/strands/agent/agent_result.py
liuyuchen777: The purpose of this PR is adding estimated total cost into the agent result. The previous naming doesn't reflect to the purpose. The follow-up commits have fixed attribute to `totalCost`.

strands-py/src/strands/models/litellm.py
liuyuchen777: Nice catch. I have added cached token in `completion_cost`. However, `prompt_tokens_details` and `cache_creation_input_tokens` is only available for LiteLLM over Anthropic based on comment in L322-323

strands-py/src/strands/models/litellm.py
liuyuchen777: Fixed in new commit

liuyuchen777:
Hi @mkmeral, could you help review again for the latest commit? There're several changes reflect comments from review bot and you:

1. `cost` property in AgentResult now reflect the cost of the invocation
2. Accumulated cost is available in EventLoopMetrics as `totalCostUsd`
3. Cached token is considered in LiteLLM `completion_cost`
4. Add documentation for cost

PR #3217, opened 2026-07-13
docs: add model routing design doc (0016)

### LINKED ISSUE #364: [FEATURE] Bedrock Intelligent prompt routing
### Problem Statement

**Description:**

In our production implementation, we're migrating from Bedrock Agents to Strands Agents due to latency concerns, where even simple greetings had unacceptable response times. To optimize performance in Strands Agents, we need two key capabilities:

**1. Intelligent Prompt Routing System (similar to Bedrock's implementation or how can we use this feature from bedrock):**
Low-latency path for simple queries (greetings, FAQs)

Complex query detection & routing

Dynamic load balancing between models

Cached response handling for common requests

**2. Claude Haiku 3.5 Optimization (for latency-critical use cases):**
 Priority routing of time-sensitive queries to Haiku

Configuration guidelines for minimal response overhead

Cold-start mitigation strategies

Streamlining options for Haiku’s already-optimized architecture

### Proposed Solution

_No response_

### Use Case

It will reduce response time.

### Alternatives Solutions

_No response_

### Additional Context

If its already available in strands kindly provide reference to implement it.

JackYPCOnline:
### Motivation

A Strands `Agent` currently binds to one model at construction, so applications must implement their own model selection and failure handling. This design proposes opt-in runtime routing while preserving existing single-model behavior.

### Proposed Public API

```python
router = ModelRouter(
    models=[
        RoutingCandidate(
            name="routine",
            model=haiku,
            description="Direct questions, extraction, and summarization.",
        ),
        RoutingCandidate(
            name="complex",
            model=opus,
            description="Ambiguous requests requiring deeper reasoning.",
        ),
    ],
    strategy=ModelDrivenStrategy(judge=judge),
)
agent = Agent(model=router)
```

V1 includes ordered fallback, local context-fit selection, and model-driven selection. Models, candidates, strategies, and routers are reusable across invocations and agents. API names are provisional pending API review.

Existing `Agent(model=model)` usage is unchanged, so this proposal introduces no breaking change.

## Related Issues

Resolves #364

## Documentation PR

Not applicable. This PR is the design proposal; user documentation will accompany implementation.

## Type of Change

Documentation update

## Testing

Documentation-only change. Repository build and commit checks pass.

- [ ] I ran `hatch run prepare` (not applicable to a `team/designs`-only change)

## Checklist

- [x] I have added an appropriate example to the documentation, or no new docs are needed

----


team/designs/0016-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+
+Model routing is a proposed feature that lets an agent decide *which model handles a call* at runtime instead of at construction. The model that serves a given call is chosen from the request, the conversation, cost, availability, or the phase of work, rather than binding one model for the life of the agent. The industry has been moving this way, and it matters more now that advanced models cost roughly 5x to 10x what smaller ones do, so sending cheap turns to cheap models is a real saving.
+
+### Related Issue/ Customer Request
+
+https://github.com/strands-agents/harness-sdk/issues/364
+
+## Goals & Non-Goals
+
+Goals:
```
Unshure, line 19: Lets keep the goals tied to technical features, not the number of strategies.

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+| [Aurelio semantic-router](https://github.com/aurelio-labs/semantic-router) / [vLLM semantic-router](https://github.com/vllm-project/semantic-router) | Client decision layer | Embedding/semantic vector match, no LLM call to decide | Fast classifier alternative that avoids an extra model call |
+| [Pydantic-AI `FallbackModel`](https://ai.pydantic.dev/api/models/fallback/) | Model wrapper in a typed SDK | Wraps an ordered model list; falls through on failure | Model-wrapper routing is one option; we weigh it against a first-class type below |
+| [AWS Bedrock Intelligent Prompt Routing](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-routing.html) | Server-side, single serverless endpoint | Predicted per-request quality across a model family | Some routing can be **delegated server-side** rather than built client-side |
+
+
+The intersection worth taking for Strands is small: client-side, model-target, fallback plus proactive selection. Two non-obvious points stand out: fallback is the one mechanism everyone ships, and modeling the router as a `Model` wrapper is the tempting shortcut in a typed SDK, though we end up choosing a first-class type instead (see Integration).
```
Unshure: Does this need to be client side? Could we/should we integrate with bedrock intelligent prompt routing

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+4. Scope: per-call is the most flexible (it can escalate mid-invocation after a tool result) but the riskiest for stateful models and cache locality; per-invocation is simpler and safer but cannot adapt within a run.
+
+5. Location: this is clearly in our SDK. The real boundary is state scope. A decision that needs only what this process can see belongs in the SDK, while anything that needs a global view of traffic belongs in a gateway.
+
+6. Composition: we should decide this upfront, since it depends on how sophisticated we want the first release to be.
+
+
+## Scope decisions
+
+The trade-offs resolve into a single decision. The baseline routes only between models (not providers), decides in the SDK before each call, falls back after a failure, and never silently switches away from a model that holds server-side state. What to optimize for is a config option, though cost is deferred until we have per-model pricing. Strategies stay simple for now, but the interface is built so they can nest later, and provider or endpoint load balancing is left to a gateway or Bedrock.
```
Unshure: Why only between models and not providers? I can imagine why this might be useful, but curious to hear your thoughts

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+
+Two strategies ship first, because both reuse machinery that already exists:
```
mkmeral: so we are not doing intelligent model routing for v1? i.e. i want complex questions to go to model A, easy ones to B and so on

team/designs/0015-model-routing.md
zastrowm: That's bedrock specific though, isn't it?  What about our OpenAI users?

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+
+When both arguments are present, `model_router` wins. It owns the per-call decision and its `models` is the authoritative candidate pool. `model=` is the default used when no router is set, and it stays optional otherwise, serving only as the router's fallback of last resort. The two never overlap, because `model=` answers "what is the default model" and `model_router=` answers "how is the model chosen per call." A candidate is only in the routing pool if it appears in `models`, so a model never lands in a group implicitly.
+
+
+## Proposed Architecture & Workflow
+
+Routing inserts a decision step in front of the model call. The agent holds its default `model` as today and, when routing is on, a `model_router`. The event loop consults the router, which asks a strategy which candidate should serve the call, then invokes that candidate. Everything upstream (hooks, middleware, retry) runs unchanged.
```
lizradway: > which asks a strategy which candidate should serve the call

How is it actually determined what task should map to which model (ie. which model is the smallest able to handle the task?)

team/designs/0015-model-routing.md
mkmeral: the scope defined here does not actually cover the customer issue, right?

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+D is a plugin-style option, but it still needs something to intercept the call, and the cleanest interceptor is a model-shaped wrapper, so D collapses into E.
+
+### Why not widen a hook or middleware to set the model?
+
+The current contracts:
+- The event-loop terminal invokes `stream_messages(agent.model, ...)`, so it reads `agent.model` directly.
+- `BeforeModelCallEvent` is read-only except `cancel` (its write guard allows only `"cancel"`), and it has no `model` field.
+- `InvokeModelContext` has no `model` field, and the terminal would ignore it regardless.
+
+So there are two ways to make a hook or middleware route, each with a cost:
```
zastrowm: The Middleware approach would be to simplify call whatever `model` you wanted instead of delegating to agent.model; I'd probably suggest that instead of hooks FWIW

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+## Scope decisions
+
+The trade-offs resolve into a single decision. The baseline routes only between models (not providers), decides in the SDK before each call, falls back after a failure, and never silently switches away from a model that holds server-side state. What to optimize for is a config option, though cost is deferred until we have per-model pricing. Strategies stay simple for now, but the interface is built so they can nest later, and provider or endpoint load balancing is left to a gateway or Bedrock.
+
+Two strategies ship first, because both reuse machinery that already exists:
+
+| Strategy | Reuses in Strands | Why first |
+|---|---|---|
+| Fallback  | `ModelRetryStrategy`, `AfterModelCallEvent`, `AfterInvocationEvent` | Lowest risk; retry the primary, then advance to the next model |
+| Intelligent  | `count_tokens`, `context_window_limit` | Local, deterministic decision; no extra model round-trip |
```
lizradway: what is the deterministic algorithm?

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+
+What actually separates the options is three things: whether it is safe when one agent handles several calls at once, whether it can react *after* a call (which fallback needs), and whether it forces a change to the agent's core loop.
+
+- **A. Make the router look like a model.** The router implements the `Model` interface and is passed as `model=`, exactly like a normal model. Internally it picks one of several models and forwards the call. Nothing else in the agent changes.
+- **B. A hook that rewrites the agent's model before each call.** Reuses the event system, but it edits `agent.model`, a value shared across calls. If one agent runs two requests at once, they overwrite each other.
+- **C. A pipeline step that names the model to use.** Conceptually clean, but the code that actually calls the model ignores what the pipeline chose and reads the agent's model directly, so it would not take effect without changing that core code.
+- **D. A separate `model_router=` setting (a plugin).** Self-describing, and it keeps routing distinct from "the model," but it is a brand-new knob and it still needs one of the other mechanisms underneath to actually intercept the call.
```
mkmeral: I think we can make a plugin, but do we need high level agent param for it? Is it a core feature? 🤔

team/designs/0015-model-routing.md
JackYPCOnline: it is supported by bedrock server side with listed models, only see CLI / console example from the bedrock doc.

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+### Why not widen a hook or middleware to set the model?
+
+The current contracts:
+- The event-loop terminal invokes `stream_messages(agent.model, ...)`, so it reads `agent.model` directly.
+- `BeforeModelCallEvent` is read-only except `cancel` (its write guard allows only `"cancel"`), and it has no `model` field.
+- `InvokeModelContext` has no `model` field, and the terminal would ignore it regardless.
+
+So there are two ways to make a hook or middleware route, each with a cost:
+
+| Approach | What it takes to work | Cost |
```
Unshure: I feel like this would best fit as a high level plugin (like session manager or something). We can add it as a regular plugin, and uplevel it to a top-level attribute later if we see the need

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+Candidate models live in the router's `models`, and the shape follows the strategy. Fallback takes an ordered list where order is the priority, and a selection strategy ranks over whatever candidates it is given. A model is only in the routing pool if it appears in `models`, so nothing lands in a group implicitly. `model=` stays the default for the no-router case and is optional once a router is set. `strategy` is a `RoutingStrategy` object, not a string, because strategies carry their own configuration and are the user extension point.
+
+
+```python
+haiku  = BedrockModel(model_id="anthropic.claude-3-5-haiku-20241022-v1:0")
+sonnet = BedrockModel(model_id="anthropic.claude-3-5-sonnet-20241022-v2:0")
+opus   = AnthropicModel(model_id="claude-opus-4-1", temperature=0.2)
+
+# Fallback: try in order, advance only on failure. Order is the priority.
+agent = Agent(model_router=ModelRouter(models=[haiku, sonnet, opus], strategy=FallbackStrategy()))
```
pgrayy: Is model driven selection supported? Could agent decide to switch it's own provider? How would that look?

team/designs/0015-model-routing.md
notowen333: `Intelligent` does not make clear the axis on which the decision is being made on

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+| B, before-call hook | register a hook | No, edits shared state | Yes | No |
+| C, pipeline picks model | add a pipeline step | Yes | Partly | Yes, blocked today |
+| D, separate `model_router=` | new argument | Yes | Yes | No (needs A underneath) |
+| E, model + plugin | pass as `model=` | Yes | Yes | No |
+
+B edits a shared value, so concurrent calls on one agent race each other.
+
+In C, the code that currently runs the model reads the agent's model directly and ignores the pipeline's choice, so it would require a core-loop change first.
+
+D is a plugin-style option, but it still needs something to intercept the call, and the cleanest interceptor is a model-shaped wrapper, so D collapses into E.
```
mkmeral: can't we use hooks for this? hooks are meant to intercept calls, if we need to extend what can be set/changed through hooks, let's do that imo

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+    EL->>EL: BeforeModelCallEvent (re-entry)
+    EL->>R: select -> Next
+    EL->>F: stream(...)
+    F-->>EL: StreamEvent*
+    EL->>EL: success (no retry)
+```
+
+`ModelRetryStrategy` handles backoff and retrying the *same* model, while the router's own fallback hook advances to the *next* model. The ordering between them (retry first, then fall back) is still open.
+
+## DevX
```
pgrayy: What other approaches have you considered? What were the pros and cons?

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+
+So there are two ways to make a hook or middleware route, each with a cost:
+
+| Approach | What it takes to work | Cost |
+|---|---|---|
+| Mutate `agent.model` from a hook | nothing new, just write the attribute | `agent.model` is one shared slot: concurrent calls race, and even a single invocation (which runs many model calls across tool-use cycles) must save and restore it and reason about a moving "current model" |
+| Per-call model override (option C) | add a writable `model` field to the event or context, and change the terminal to honor it | modifies the SDK's hottest, most-tested code path and introduces a new permanent public contract |
+| Router is-a `Model` (option E) | nothing, the router is already the call target | the decision is a local inside `stream()`, so it is call-scoped by construction: no shared state, no restore, no core-loop change |
+
+The A to E walkthrough above first landed on Option E, a `ModelRouter` that is-a `Model` passed through `model=` but this can be very confuse to users.
```
zastrowm: > passed through `model=` but this can be very confuse to users.

I think it's also a bit confusing to have `model_router` that takes priority over `model`; to me, I think `model=router` is less confusing as it fits my mental model a bit more

team/designs/0015-model-routing.md
JackYPCOnline: Provider is the superset of its model, routing could be models from different providers under certain Strategy

team/designs/0015-model-routing.md
pgrayy: What happens if the user passes in `model=...`? Is that ignored?

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+- A candidate may itself be a `ModelRouter`, so groups nest. A model is a routing candidate only when it appears in some `models` list; `model=` is not folded in implicitly.
+- As a `Plugin`, `init_agent` registers the `AfterModelCallEvent` hook so fallback composes with `ModelRetryStrategy` (retry first, then advance) rather than reimplementing backoff.
+
+
+## Work Plan
+
+**P0: Router core + fallback.** `ModelRouter` (a first-class type and `Plugin`), the `Agent(model_router=...)` argument and the event-loop consult (default to `agent.model` when unset), `RoutingStrategy` (`select` plus optional `on_result`), a candidate registry (ordered list and named mapping, taking a `Model`, a model-id string, or a nested `ModelRouter`), fallback composing with `ModelRetryStrategy`, a `stateful` guard, and the default scope.
+
+**P0: Intelligent routing.** A signals-based `select()` on `count_tokens` and `context_window_limit`, with the routing decision emitted on the model-invoke span.
```
lizradway: how is this intelligent? are we just using the context window limit fill ratio to route to cheaper models when the context window is more full?

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+
+**P1: Cost-aware routing (follow-up).** Add price columns to the defaults table (`_defaults.py`), then a cost/latency objective. With pricing available, `ContextFitStrategy` gains a cost-aware sibling that keeps the context-fit filter but ranks the survivors by price instead of window size, `min(fits, key=lambda name: fits[name].cost)`.
```
lizradway: what do the price columns contain? is that the per 1M token cost of each model?

team/designs/0015-model-routing.md
Unshure: +1, have a similar comment below

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+
+**P1: LLM-based routing (follow-up).** Agentic handoff, a classifier or semantic `select()`, a quality-driven cascade, and an evaluation harness. Routing with a classifier model, or letting the agent pick, is deferred rather than designed out. It adds a model call to the decision path (latency and cost on every route), and trustworthy use needs a decision model, a label or prompt scheme, and an evaluation harness. The quality-driven variant (cascade) additionally needs a verification or confidence signal the SDK does not expose yet. It fits the same abstraction, since an LLM router is just a `RoutingStrategy.select()` that calls a model internally, and it layers on top of the v1 router as its own workstream.
```
lizradway: This feels like the explicit part of the design that makes the juice worth the squeeze ihmo, so its surprising to me to see this as a follow up.

team/designs/0015-model-routing.md
JackYPCOnline: It is covered in future extensibilty section below. v1 is defined by given the sprint time frame. with said, we should add model driven routing

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+
+**Decision: a first-class `ModelRouter`, attached through a dedicated `model_router=` argument.** the agent holds it in a new `model_router=` slot beside the existing `model=`. This parallels `tool_executor=`, a first-class component the agent holds and the event loop consults. When `model_router` is set, the loop consults it to pick a candidate before invoking; when it is not, the loop calls `agent.model` exactly as today. `agent.model` stays a real `Model`, so tracing, `count_tokens`, `stateful`, and `structured_output` keep reading it with no re-implementation.
+
+When both arguments are present, `model_router` wins. It owns the per-call decision and its `models` is the authoritative candidate pool. `model=` is the default used when no router is set, and it stays optional otherwise, serving only as the router's fallback of last resort. The two never overlap, because `model=` answers "what is the default model" and `model_router=` answers "how is the model chosen per call." A candidate is only in the routing pool if it appears in `models`, so a model never lands in a group implicitly.
```
mehtarac: > When both arguments are present, `model_router` wins.

When only `model_router` is set, in our current implementation we still initialize a Bedrock model which would never be used. With the new `model_router`, would model be truly optional with. no default set?

team/designs/0015-model-routing.md
JackYPCOnline: that the hilarious part : by another model

team/designs/0015-model-routing.md
pgrayy: This brings to mind whether we should explore a Litellm/Vercel approach to model providers. Something along the lines of:
```python
from strands import Agent
from strands import Model
from strands.model import BedrockProvider

model = Model(provider=BedrockProvider(...))
agent = Agent(model=model)
```

Under this approach, you might be able to more naturally fit routing into this construct. Could maybe support something like `Model(providers=[...])`. This would allow you to centralize both the provider selection mechanism and the model id selection potentially.

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+| Per-call model override (option C) | add a writable `model` field to the event or context, and change the terminal to honor it | modifies the SDK's hottest, most-tested code path and introduces a new permanent public contract |
+| Router is-a `Model` (option E) | nothing, the router is already the call target | the decision is a local inside `stream()`, so it is call-scoped by construction: no shared state, no restore, no core-loop change |
+
+The A to E walkthrough above first landed on Option E, a `ModelRouter` that is-a `Model` passed through `model=` but this can be very confuse to users.
+
+**Decision: a first-class `ModelRouter`, attached through a dedicated `model_router=` argument.** the agent holds it in a new `model_router=` slot beside the existing `model=`. This parallels `tool_executor=`, a first-class component the agent holds and the event loop consults. When `model_router` is set, the loop consults it to pick a candidate before invoking; when it is not, the loop calls `agent.model` exactly as today. `agent.model` stays a real `Model`, so tracing, `count_tokens`, `stateful`, and `structured_output` keep reading it with no re-implementation.
```
mkmeral: why a first class parameter? what's the justification? 

I personally am not convinced that it should be a high level primitive. Can't we just have a plugin that hooks to before invocation, and updates `agent.model` in-place. Doesn't that achieve the same thing? without any core agent loop modifications?

team/designs/0015-model-routing.md
Unshure: I do like this, but is a model just a plugin now too? I imagine a model router needs to see hook events

team/designs/0015-model-routing.md
JackYPCOnline: have a simple example in devX section.

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+```python
+haiku  = BedrockModel(model_id="anthropic.claude-3-5-haiku-20241022-v1:0")
+sonnet = BedrockModel(model_id="anthropic.claude-3-5-sonnet-20241022-v2:0")
+opus   = AnthropicModel(model_id="claude-opus-4-1", temperature=0.2)
+
+# Fallback: try in order, advance only on failure. Order is the priority.
+agent = Agent(model_router=ModelRouter(models=[haiku, sonnet, opus], strategy=FallbackStrategy()))
+
+# Intelligent: pick the smallest-context model that fits the request. Any number, no tier names.
+agent = Agent(model_router=ModelRouter(models=[haiku, sonnet, opus], strategy=ContextFitStrategy()))
```
pgrayy: Does this take into account model id selection or only provider selection?

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+
+**Decision: a first-class `ModelRouter`, attached through a dedicated `model_router=` argument.** the agent holds it in a new `model_router=` slot beside the existing `model=`. This parallels `tool_executor=`, a first-class component the agent holds and the event loop consults. When `model_router` is set, the loop consults it to pick a candidate before invoking; when it is not, the loop calls `agent.model` exactly as today. `agent.model` stays a real `Model`, so tracing, `count_tokens`, `stateful`, and `structured_output` keep reading it with no re-implementation.
+
+When both arguments are present, `model_router` wins. It owns the per-call decision and its `models` is the authoritative candidate pool. `model=` is the default used when no router is set, and it stays optional otherwise, serving only as the router's fallback of last resort. The two never overlap, because `model=` answers "what is the default model" and `model_router=` answers "how is the model chosen per call." A candidate is only in the routing pool if it appears in `models`, so a model never lands in a group implicitly.
+
+
+## Proposed Architecture & Workflow
```
Unshure: Im a little lost here. Does every LLM invocation get routed, or just the initial user question, and the rest of that event loop run uses the same routed model?

I also know we can do both with different strategies, but what would the default strategy be?

team/designs/0015-model-routing.md
lizradway: +1

team/designs/0015-model-routing.md
notowen333: I want a system where high level planning and orchestration goes to the most expensive model and low level work goes to a cheaper model.

Does the `strategy` interface capture such a use case?

team/designs/0015-model-routing.md
JackYPCOnline: the option is accepting model routing via model =  or model_router = . That is the major design decision actually.

team/designs/0015-model-routing.md
zastrowm: > I do like this, but is a model just a plugin now too?

I don't think I have a problem with that, given that most of our other primitives build on plugins

team/designs/0015-model-routing.md
Unshure: Just a thought, but it would be cool if we could update model parameters too as a part of the router. For example, optionally upgrading to 1M context window if we hit an overflow

team/designs/0015-model-routing.md
lizradway: gotcha, that makes sense. but that isn't in v1, right? llm-determined is only a follow up?


id be very curious as well to see if there is any research on this that suggests levels of accuracy for self-reported task difficulty/token estimation from llms

team/designs/0015-model-routing.md
JackYPCOnline: yes, D works actually.

team/designs/0015-model-routing.md
JackYPCOnline: yes, since model = is not optional

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+
+### Related Issue/ Customer Request
+
+https://github.com/strands-agents/harness-sdk/issues/364
+
+## Goals & Non-Goals
+
+Goals:
+- Make model choice a per-call runtime decision, opt-in, with single-model usage unchanged.
+- A small, composable abstraction that enables routing.
```
lizradway: just as a note: i'd be very interested in routing context strategies eventually. so to make this some sort of task judger + then config router would be very interesting to me.

 ie. not only model routing, but config routing?  im not sure where model routing ends and agent self configuration/self update begins, but i think that is worth considering in this doc.

team/designs/0016-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+
+
+| Option | How it is enabled | Safe under concurrent calls | Can react after a call (fallback) | Changes the core loop |
+|---|---|---|---|---|
+| A, router is a model | pass as `model=` | Yes | Not on its own | No |
+| B, before-call hook | register a hook | No, edits shared state | Yes | No |
+| C, pipeline picks model | add a pipeline step | Yes | Partly | Yes, blocked today |
+| D, separate `model_router=` | new argument | Yes | Yes | No (needs A underneath) |
+| E, model + plugin | pass as `model=` | Yes | Yes | No |
+
```
gautamsirdeshmukh, line 48: I actually do like option A quite a bit - a few inference providers put out intelligent endpoints that can be hit like a typical endpoint but do some model routing behind the scenes, feels like a similar sort of idea here

team/designs/0015-model-routing.md
Unshure: LLM based routing seems like the most useful, no? This may be covered in the previous points, but choosing a cheaper model for simpler tasks, and a complex model for harder tasks seems like the main use case for a feature like this to me

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+
+```python
+haiku  = BedrockModel(model_id="anthropic.claude-3-5-haiku-20241022-v1:0")
+sonnet = BedrockModel(model_id="anthropic.claude-3-5-sonnet-20241022-v2:0")
+opus   = AnthropicModel(model_id="claude-opus-4-1", temperature=0.2)
+
+# Fallback: try in order, advance only on failure. Order is the priority.
+agent = Agent(model_router=ModelRouter(models=[haiku, sonnet, opus], strategy=FallbackStrategy()))
+
+# Intelligent: pick the smallest-context model that fits the request. Any number, no tier names.
```
mkmeral: can we combine strategies? if i want intelligent routing initially, but if it fails, i'd rather fallback to others and not just fail my request

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+```python
+class ContextFitStrategy:
+    """Pick the smallest-context model that still fits the request.
+
+    Uses only context_window_limit, so it ships in v1 with no cost data.
+    Smaller-window models are usually the cheaper, faster ones, so easy
+    turns stay on small models and only large prompts escalate.
+    """
+
+    async def select(self, ctx: RoutingContext) -> str:
```
lizradway: this is just picking the model where the prompt can fit in the context window?

team/designs/0016-model-routing.md
gautamsirdeshmukh: A point for going the plugin route would be that it'd streamline setting up hook events + registering some sort of handoff/routing tool to the main agent

team/designs/0015-model-routing.md
lizradway: isnt this going to hurt/cost more than help due to prompt caching etc?

team/designs/0015-model-routing.md
Unshure: Just read into this service a bit more. Seems like if we did this, it would be a setting on the BedrockModelProvider instead of a client side router. My misunderstanding

team/designs/0016-model-routing.md
```diff
@@ -0,0 +1,340 @@
+# Model Routing
```
mkmeral, line 1: high level comments: 

1. I am not sure if we want first class primitive for model routing, or just a plugin. I'd be in favor of a plugin for now, and we can uplevel later if we want/need to 
2. We can do step by step, but I'd like to have some form of intelligent routing (using LLMs etc.) before we call this feature complete. So one of those advance routing strategies

team/designs/0015-model-routing.md
pgrayy: I remember running into an issue where gemini rejected a messages array because a reasoning id was missing. We had to do something specific to gemini to make sure include it. Have you gone through and considered how models might handle messages array formed from another provider? Are there many gotchas?

team/designs/0015-model-routing.md
JackYPCOnline: TBH, I would prefer LiteLLM approach where it takes a list of models, but the difference is that they use Router.complete()

team/designs/0015-model-routing.md
JackYPCOnline: model is ignored from my prospective

team/designs/0015-model-routing.md
JackYPCOnline: it is per invocation, default to me would be fallbackstrategy

team/designs/0015-model-routing.md
gautamsirdeshmukh: Hmm. Perhaps we call this CostEfficient? IMO Intelligent would involve either following some process/skill defined by the user (or some model-driven approach) to route based on presumed task complexity

gautamsirdeshmukh: (which I do think is one of if not the most critical strategy to enable for a model routing mechanism)

team/designs/0015-model-routing.md
mkmeral: what do we mean by models and not providers? do we have a separation of the two in SDK? 

I think the interfaces below are right, so won't push too much. but this section confused me a bit at the beginning

team/designs/0015-model-routing.md
JackYPCOnline: yes, it is.

gautamsirdeshmukh: +1, similar comment above

team/designs/0015-model-routing.md
JackYPCOnline: use some mechanism to determine which model to pick

team/designs/0015-model-routing.md
Unshure: Do we need to separate providers from models? Can the sdk abstract away the complexity of thinking about providers, and just let people choose any model they want? Or is there some value in letting a customer choose a provider?

team/designs/0015-model-routing.md
yonib05: We can and should abstract away the diffrence, stan does this well with the "openai/gpt5.6" "bedrock/openai-gpt5.6" pattern.


Model doesn't need to mean just model; it should be an amalgamation of provider+model.

team/designs/0015-model-routing.md
yonib05: We should, in P0, release customer value, so good viable strategies in each of the objectives that are being optimized for is a must, IMO.  

> quality/accuracy · latency · availability/reliability · compliance/residency

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+Model routing is a proposed feature that lets an agent decide *which model handles a call* at runtime instead of at construction. The model that serves a given call is chosen from the request, the conversation, cost, availability, or the phase of work, rather than binding one model for the life of the agent. The industry has been moving this way, and it matters more now that advanced models cost roughly 5x to 10x what smaller ones do, so sending cheap turns to cheap models is a real saving.
+
+### Related Issue/ Customer Request
+
+https://github.com/strands-agents/harness-sdk/issues/364
+
+## Goals & Non-Goals
+
+Goals:
+- Make model choice a per-call runtime decision, opt-in, with single-model usage unchanged.
```
yonib05: I think P0 is strategies in each of the noted objective areas

team/designs/0015-model-routing.md
yonib05: +1, I like that as a strategy

team/designs/0015-model-routing.md
```diff
@@ -0,0 +1,340 @@
...
+    EL->>F: stream(...)
+    F-->>EL: StreamEvent*
+    EL->>EL: success (no retry)
+```
+
+`ModelRetryStrategy` handles backoff and retrying the *same* model, while the router's own fallback hook advances to the *next* model. The ordering between them (retry first, then fall back) is still open.
+
+## DevX
+
+Routing is an optional capability passed to `Agent` through `model_router=`, beside the existing `model=`, with sensible defaults and single-model usage unchanged. A one-line fallback is trivial, and advanced strategies opt in.
```
yonib05: Why not just have model=ModelRouter() so there is no confusion on what the agent is using.

yonib05: How does caching play into model routing? I think it's worth making sure session sticky is default when caching is enabled.

mehtarac (approved):

PR #3222, opened 2026-07-13
feat(memory): add FileMemoryStore

maisieyanz:
## Description

Adds `FileMemoryStore`, a zero-infrastructure memory store backed by a file hierarchy. It implements the `MemoryStore` interface for use with `MemoryManager`, storing knowledge as markdown files with YAML frontmatter under `knowledge/`. Retrieval uses keyword-based search across filenames, descriptions, and content ranked by term frequency.

This is the first PR in a series — it covers the core `search` and `add` methods only. Consolidation (offline LLM-driven maintenance) comes in a follow-up.

## Public API

```typescript
import { FileMemoryStore } from '@strands-agents/sdk/vended-memory-stores/file-memory-store'

const memoryStore = new FileMemoryStore({ name: 'agent-memory' })

const agent = new Agent({
  model,
  memoryManager: new MemoryManager({ stores: [memoryStore], injection: false }),
})
```

`FileMemoryStoreConfig` extends `MemoryStoreConfig` with one additional field:

- `storage?: Storage` — the backend for file operations. Defaults to `LocalFileStorage` at `./.strands/`.

Knowledge is stored as markdown files with YAML frontmatter (`description` field) under `knowledge/`. `search()` performs keyword matching against filenames, descriptions, and content, ranked by term frequency. `add()` writes to `knowledge/facts/` by default, with `metadata.path` for custom placement.

## Type of Change

Bug fix
New feature
Breaking change
Documentation update
Other (please describe):

## Testing

## Checklist

----


mkmeral:
@strandly-the-agent can you review this PR? also cna you spawn a subagent to review the API BR and drop a comment here for the current API surface in this code?

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,144 @@
...
+    let key: string
+    if (customPath) {
+      key = customPath.startsWith(KNOWLEDGE_PREFIX) ? customPath : `${KNOWLEDGE_PREFIX}${customPath}`
+      if (!key.endsWith('.md')) key += '.md'
+    } else {
+      const slug = slugify(title) || `entry-${Date.now()}`
+      key = `${FACTS_PREFIX}${slug}.md`
+    }
+
+    const fileContent = `---\ndescription: "${description.replace(/"/g, '\\"')}"\n---\n\n${content}\n`
```
strandly-the-agent: 🔴 **Hand-built YAML frontmatter silently corrupts / loses stored data.** _(converged across 4 passes)_

`add()` interpolates the description into YAML escaping only `"`, while `parseFrontmatter` (`:22`, `:28`) is line-anchored (`/m`) with a non-greedy `\n---\n?` close and never unescapes. Reproduced three ways against the exact write+read path:

```
// (a) description with a fence breaks out and corrupts the BODY
add('real body', { description: 'intro\n---\nsecret: injected' })
  → search().content === 'secret: injected"\n---\n\nreal body'

// (b) any newline in description silently drops the rest
add(c, { description: 'line one\nline two' })  → description === 'line one'

// (c) a quoted description round-trips as literal backslashes
add(c, { description: 'Prefers "strict" mode' })
  → search().metadata.description === 'Prefers \\"strict\\" mode'
```

Case (c) fires on the **default path** whenever stored content's first 120 chars contain a `"` — so this is normal-use, not just adversarial.

**Fix:** don't hand-roll YAML. Serialize with `JSON.stringify(description)` (valid single-line double-quoted YAML — escapes `"`, `\`, and `\n`) **and** make `parseFrontmatter` reverse it (`JSON.parse` the captured value), or adopt a real YAML lib. The two sides must match — fixing only the writer still returns backslash-garbage on read.

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,144 @@
...
+    const allKeys = await this._storage.list(KNOWLEDGE_PREFIX)
+    const scored: Array<{ entry: MemoryEntry; score: number }> = []
+
+    for (const key of allKeys) {
+      const bytes = await this._storage.read(key)
+      if (!bytes) continue
+
+      const content = decoder.decode(bytes)
+      const { description, body } = parseFrontmatter(content)
+      const searchable = `${key} ${description} ${body}`.toLowerCase()
```
strandly-the-agent: 🟡 **The storage key pollutes the search corpus, and scoring is raw-substring.** _(converged across 4 passes)_

`searchable = \`${key} ${description} ${body}\`` means every file's path is scored. Since every key is `knowledge/facts/*.md`, `search("md")` and `search("knowledge")` match **every** file (reproduced), and substring counting on `:102` makes `search("cat")` hit `category.md` and over-counts fragments in long docs. The model gets low-relevance junk in its context.

**Fix:** score against `basename-minus-extension + description + body` (drop the `knowledge/…/.md` scaffolding), and tokenize on word boundaries — the sibling `test-memory-store` already does whole-token overlap (`store.ts:69-88`), worth matching for family consistency.

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,144 @@
...
+    const customPath = metadata?.['path'] as string | undefined
+    const title = (metadata?.['title'] as string | undefined) ?? slugify(content.split(/[.\n]/)[0]!.slice(0, 60))
+    const description = (metadata?.['description'] as string | undefined) ?? content.split(/[.\n]/)[0]!.slice(0, 120)
+
+    let key: string
+    if (customPath) {
+      key = customPath.startsWith(KNOWLEDGE_PREFIX) ? customPath : `${KNOWLEDGE_PREFIX}${customPath}`
+      if (!key.endsWith('.md')) key += '.md'
+    } else {
+      const slug = slugify(title) || `entry-${Date.now()}`
```
strandly-the-agent, line 217: 🟡 **Silent slug-collision overwrite → data loss.** _(converged across 3 passes; reproduced)_

Two distinct facts sharing their first ~50 chars slugify to the same filename, and the second `add` silently overwrites the first:

```
add('Python is great')
add('Python is great. But has a GIL.')
  → list('knowledge/facts/') === ['knowledge/facts/python-is-great.md']  // first fact lost
```

For an accumulating memory store that's silent loss. **Fix:** disambiguate on collision (short content hash / numeric suffix), or explicitly document upsert/overwrite semantics if that's intended. (Re-adding *identical* content overwriting itself is fine per the extraction at-least-once contract — it's *distinct*-content collisions that are the problem.)

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,144 @@
...
+  async search(query: string, options?: SearchOptions): Promise<MemoryEntry[]> {
+    const maxResults = options?.maxSearchResults ?? this.maxSearchResults ?? 5
+    const terms = query.toLowerCase().split(/\s+/).filter(Boolean)
+    if (terms.length === 0) return []
+
+    const allKeys = await this._storage.list(KNOWLEDGE_PREFIX)
+    const scored: Array<{ entry: MemoryEntry; score: number }> = []
+
+    for (const key of allKeys) {
+      const bytes = await this._storage.read(key)
```
strandly-the-agent: 🟡 **One unreadable file aborts the entire search.** _(converged across 2 passes; reproduced)_

There's no `try/catch` around `this._storage.read(key)` in the loop, so a single throwing read (EACCES, partial write, transient FS error) rejects the whole `search` — the user gets **zero** memories instead of the other matches. The test at `…test…:278` covers `read → null` but not a throw, and real filesystems throw.

**Fix:** wrap the per-key read + parse in `try/catch` and `continue` on failure, mirroring the existing `if (!bytes) continue`.

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,144 @@
...
+
+      let score = 0
+      for (const term of terms) {
+        const matches = searchable.split(term).length - 1
+        score += matches
+      }
+
+      if (score > 0) {
+        scored.push({
+          entry: { content: body.trim(), metadata: { path: key, description } },
```
strandly-the-agent: 🟡 **The computed relevance score is discarded — the model gets no confidence signal.** _(converged across 2 passes)_

`score` is computed (`:100-104`) then dropped. With keyword TF ranking the gap between the top hit and rank-5 is often large; handing the model the score lets it discount weak matches instead of treating all results as equally authoritative. The sibling surfaces it as `metadata._relevanceScore` (`test-memory-store/store.ts:197`) and `MemoryEntry.metadata`'s own docstring already lists "score" (`memory/types.ts:19`). One-line fix:

```suggestion
          entry: { content: body.trim(), metadata: { path: key, description, score } },
```

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,144 @@
...
+  /**
+   * Add a knowledge entry to the store.
+   *
+   * Writes a markdown file with YAML frontmatter. By default writes to `knowledge/facts/`.
+   * Pass `metadata.path` to write to a custom location under `knowledge/`.
+   *
+   * @param content - The knowledge content to store
+   * @param metadata - Optional metadata: `title`, `description`, and `path` (custom target path)
+   */
+  async add(content: string, metadata?: Record<string, JSONValue>): Promise<void> {
```
strandly-the-agent: 🟡 **`add` returns `Promise<void>` while both siblings return a typed identifier.**

The interface allows `Promise<unknown>` and its docstring says direct callers can read it (`memory/types.ts:123-125`); `TestMemoryStore.add` → `{ id }` (`store.ts:212`), `BedrockKnowledgeBaseStore.add` → `{ documentId }` (`store.ts:376`). The created key here isn't reconstructable by the caller (it depends on `slugify`, a 50-char truncation, and a `Date.now()` fallback), so a direct caller who just wrote an entry can't locate, re-read, or update it.

`void → { path: string }` is a **non-breaking** widening, but it's cheap to get right now and awkward to explain once this v1 ships. Proposal: return `Promise<{ path: string }>` and export the result type, matching the siblings.

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,144 @@
...
+    const customPath = metadata?.['path'] as string | undefined
+    const title = (metadata?.['title'] as string | undefined) ?? slugify(content.split(/[.\n]/)[0]!.slice(0, 60))
+    const description = (metadata?.['description'] as string | undefined) ?? content.split(/[.\n]/)[0]!.slice(0, 120)
+
+    let key: string
+    if (customPath) {
+      key = customPath.startsWith(KNOWLEDGE_PREFIX) ? customPath : `${KNOWLEDGE_PREFIX}${customPath}`
```
strandly-the-agent: 🟡 **`metadata.path` leaks the `knowledge/facts/` layout on both read and write.**

`search` returns the full storage key as `metadata.path` (`:108`) and `add` accepts one back here (`:133-135`), so the literal directory structure is now an observable contract on both sides. The PR says a consolidation follow-up is coming — if it reshapes the layout, every `path` a caller has read, logged, persisted, or re-fed into `add` changes shape → breaking.

Not necessarily a fix-in-this-PR, but worth a decision before v1 locks (see blocking question #2 in the summary): is `path` an opaque handle or a real filesystem path, and is `search → add({ path })` a supported round-trip guarantee?

strands-ts/src/vended-memory-stores/file-memory-store/types.ts
```diff
@@ -0,0 +1,14 @@
...
+ */
+
+import type { MemoryStoreConfig } from '../../memory/types.js'
+import type { Storage } from '../../storage/storage.js'
+
+/**
+ * Configuration for {@link FileMemoryStore}.
+ */
+export interface FileMemoryStoreConfig extends MemoryStoreConfig {
+  /** The unified Storage backend for file operations. Defaults to LocalFileStorage at `~/.strands/`. */
```
strandly-the-agent: 🟡 **Default-location doc is wrong (and the default backend is never test-exercised).** _(converged across 3 passes)_

This says the default is `LocalFileStorage` at `~/.strands/`, but `LocalFileStorage`'s default `baseDir` is cwd-relative `./.strands/` (`local-file-storage.ts:44`) — so "persistent" memory fragments per working directory. It also diverges from the sibling `test-memory-store`, which genuinely uses `os.homedir()` (`store.ts:261`). Every test injects `InMemoryStorage`, so the real default branch (`file-memory-store.ts:76`) is never exercised and the mismatch is invisible.

**Fix:** either correct this doc (and the PR body) to `./.strands/`, or pass an explicit homedir-based `baseDir` if a global store was intended (see blocking question #1).

strandly-the-agent (commented):
## 🔎 Review — `feat(memory): add FileMemoryStore`

**Verdict: 🔴 changes-requested** _(agent-generated review — solid work for a human to approve, not a gate)_

Good, well-scoped first cut with genuinely strong happy-path tests. Two things should be fixed before merge: the hand-rolled YAML frontmatter **silently corrupts stored data** (reproduced 3 ways), and `search` folds the file-path scaffolding into the match corpus so `search("md")` / `search("knowledge")` match *every* file. A few API-shape calls (`add` return type, the leaked `knowledge/facts/` layout) are cheap to settle **now**, before this v1 locks a series.

> ⚠️ This adds a **public API surface** (`FileMemoryStore` + `FileMemoryStoreConfig` + a new package export) and carries **no `needs-api-review` label** — please add one. As a new vended primitive, the return-type/layout decisions below probably warrant the API meeting rather than a solo approval.

**✅ Verified**
- Branch `pr3222` @ `b070207`.
- `npx vitest run --project unit-node …/file-memory-store` → **36/36 pass**; `eslint` clean.
- Reproduced the three frontmatter corruptions with runnable cases (see the 🔴 inline).

**8 inline findings below** — 1 🔴, 7 🟡. Several converged across 3–4 independent passes (flagged inline). Nits and open questions are collapsed here:

<details><summary>⚪ Nits (non-blocking — fix opportunistically)</summary>

- **`add()` input handling diverges from both sibling stores.** No empty-content guard, no `writable`/empty-`name`/`maxSearchResults>=1` validation, and unchecked `as string` casts — `add(c, { path: 42 })` type-checks (`JSONValue` allows numbers) then throws `TypeError` at `file-memory-store.ts:134`. `TestMemoryStore`/`BedrockKnowledgeBaseStore` both throw on `!writable` and empty content. _(Note: no need for a "rejects writes" test — `writable` is enforced by `MemoryManager` by design per `memory/types.ts:123`; but a one-line code comment saying the getter isn't self-enforcing would prevent that assumption.)_
- **`metadata.title` is write-only.** It only derives the filename slug (`:129,:137`); it's never written to frontmatter and never returned by `search`, so a caller who sets it can't find it. Persist + return it, or document it as filename-only.
- **CRLF-authored files don't parse.** `parseFrontmatter`'s regex requires bare `\n` (`:22`); a `\r\n` file leaks its frontmatter into the body. Relevant since the store supports externally-authored `knowledge/` files.
- **`slugify('---') === '---'`** so the `entry-${Date.now()}` fallback never fires for punctuation-only titles → `knowledge/facts/---.md`; also no hyphen trim/collapse (`:32-39`). And `content.split(/[.\n]/)[0]!` is computed twice (`:129-130`) — hoist.
- **Search/model-facing polish:** `description` often just duplicates the first 120 chars of `body` (`:130`), and empty-frontmatter files surface `description: ""` as noise; the store contributes no default `description`, weakening the `search_memory` routing signal; queries keep punctuation while filenames are slugified, so `dark-mode` / `prefer?` silently miss.
- **Test gaps on exactly the bug paths above** — the escaped-quote round-trip, newline-in-description, slug collision, throwing read, and the default backend branch are all untested, and `expect(results.length).toBeGreaterThanOrEqual(1)` (`…test…:184,190,196,202,260,267,274`) would survive a mutant that returns everything. Add regressions once the fixes land.

</details>

<details><summary>❓ Questions for the author</summary>

**Blocking (answer before merge):**
1. **Default location intent** — is the default meant to be **global** (`~/.strands/`, per the docstring and the sibling store) or **cwd-relative** (`./.strands/`, per the actual `LocalFileStorage` default)? The answer decides whether to fix the doc or the code.
2. **`metadata.path` contract** — opaque handle, or real filesystem path? Is the `search → add({ path })` round-trip a supported guarantee? This determines whether the `knowledge/facts/` layout can ever be reshaped by the consolidation follow-up.

**Non-blocking:**
3. Should the vended-memory-store family standardize on a typed return receipt from `add` and on surfacing a relevance score? (Family-wide consistency call.)
4. Is slug-collision overwrite intended upsert semantics, or a bug to guard?

</details>

_A full write-up of the current public API surface follows in a separate comment (produced by an API bar-raiser pass), as requested._

strandly-the-agent:
## 📐 Current public API surface — `FileMemoryStore`

_Requested by `mkmeral`. Produced by an API bar-raiser pass, based strictly on the code at head `b070207`. This is a description of the surface **as-is**; the design questions/asymmetries are in the review above, not here._

### Package export & exported symbols

**Subpath:** `@strands-agents/sdk/vended-memory-stores/file-memory-store` (`package.json`)

| Symbol | Kind | Source |
|---|---|---|
| `FileMemoryStore` | class, implements `MemoryStore` | `file-memory-store.ts:61` |
| `FileMemoryStoreConfig` | interface | `types.ts:11` |

### `FileMemoryStoreConfig` — constructor config

`FileMemoryStoreConfig extends MemoryStoreConfig` (`types.ts:11`).

| Field | Type | Req? | Default | Notes |
|---|---|---|---|---|
| `name` | `string` | **required** | — | Unique store id. Not validated (empty allowed). |
| `storage` | `Storage` | optional | `new LocalFileStorage()` | Backend. Own field (`types.ts:13`). Default writes under **`./.strands/`** (cwd-relative — see caveat). |
| `description` | `string` | optional | `undefined` | From `MemoryStoreConfig`. |
| `maxSearchResults` | `number` | optional | `undefined` (→ effective `5` at search time) | Not validated. |
| `writable` | `boolean` | optional | **`true`** | `FileMemoryStore` defaults to `true` (`file-memory-store.ts:72`), overriding the base config's documented default of `false`. |
| `extraction` | `boolean \| ExtractionConfig` | optional | `undefined` | From `MemoryStoreConfig`. |

> ⚠️ **Default-location caveat:** the JSDoc says `~/.strands/` (`types.ts:12`) but `LocalFileStorage`'s actual default `baseDir` is cwd-relative `./.strands/` (`local-file-storage.ts:44`). Flagged in the review.

**Instance properties** (all `readonly`): `name`, `description?`, `writable`, `maxSearchResults?`, `extraction?` (`file-memory-store.ts:62-66`).

### Methods

**`search(query: string, options?: SearchOptions): Promise<MemoryEntry[]>`** (`file-memory-store.ts:84`)
- Lists all keys under `knowledge/`, reads each, and scores by summed substring-occurrence count of the query's whitespace-split terms against `` `${key} ${description} ${body}` `` (case-insensitive).
- Result cap: `options.maxSearchResults ?? this.maxSearchResults ?? 5`.
- Empty / whitespace-only query → `[]`.
- Each returned `MemoryEntry`: `content` = frontmatter-stripped, trimmed body; `metadata = { path, description }`, where `path` is the full storage key (e.g. `knowledge/facts/foo.md`) and `description` is the parsed frontmatter value. **No `title`, no relevance score.** (`storeName` is populated later by `MemoryManager`.)

**`add(content: string, metadata?: Record<string, JSONValue>): Promise<void>`** (`file-memory-store.ts:127`)
- Returns **`void`** (the `MemoryStore.add` interface permits `Promise<unknown>`; sibling stores return an id).
- Does **not** check `writable` and does **not** validate `content`.
- **Metadata contract** (all optional, read via unchecked `as string` casts):
  - `path` — target location under `knowledge/`. `knowledge/` is prepended if absent; `.md` appended if absent.
  - `title` — used only to derive the filename slug when `path` is omitted. **Not persisted, not returned.**
  - `description` — written into YAML frontmatter; defaults to the first sentence/line of `content` truncated to 120 chars.
- **Where it writes:** with `path` → `knowledge/<path>[.md]`; without → `knowledge/facts/<slug>.md`, where `slug = slugify(title | firstSentence)` truncated to 50 chars, falling back to `entry-<Date.now()>` if empty.
- **File format:** `---\ndescription: "<escaped>"\n---\n\n<content>\n`.

### File-layout conventions callers depend on

- All knowledge lives under **`knowledge/`**; default new entries under **`knowledge/facts/`**.
- Files are markdown with a `description` YAML frontmatter field; bodies are returned frontmatter-stripped.
- `search` result `path` values carry the `knowledge/` prefix and round-trip back into `add(content, { path })`.
- **This layout is currently observable through the public contract**; per the PR, a consolidation follow-up may reshape it — see the `metadata.path` question in the review.

maisieyanz:

strandly-the-agent:
@maisieyanz re-reviewed at `eca4d28` — the fix commit resolves **every blocking/high item** from the first pass, and I verified each by re-running my original repros against the new code (not just reading the diff). Nicely done. ✅

**Verified:** `npx vitest run` → **41/41 pass** (was 36), eslint clean.

| Prior finding | Status | Evidence (re-ran on `eca4d28`) |
|---|---|---|
| 🔴 Frontmatter corruption | ✅ fixed | `JSON.stringify` write + `JSON.parse` read. All 3 original corruptions gone: fence-in-description keeps `content === 'real body'`; `'line one\nline two'` survives; `Prefers "strict" mode` round-trips with no backslashes. |
| 🟡 Search precision | ✅ fixed | `basename` + distinct-token overlap — `search("md")` / `search("knowledge")` now return **0** hits (were: every file). |
| 🟡 Slug-collision data loss | ✅ fixed | distinct facts kept: `python-is-great.md` + `python-is-great-1.md`. |
| 🟡 One read aborts search | ✅ fixed | per-file `try/catch` + `continue`; a throwing read leaves the other matches intact. |
| 🟡 Discarded relevance score | ✅ fixed | surfaced as `metadata._relevanceScore`. |
| 🟡 Default-location docstring | ✅ fixed | `types.ts:12` now says `./.strands/`. |

<details><summary>Still open — flagging, not re-nagging (both non-blocking design calls)</summary>

- **`add` returns `Promise<void>`** — siblings return a typed receipt so a direct caller can locate the entry it just wrote (the generated key isn't reconstructable). Unchanged.
- **`metadata.path` exposes the `knowledge/facts/` layout** as a round-trip contract — now at least *documented* in the `search` docstring ("may change after consolidation"), which helps, but the contract is still there.

These two plus the **`needs-api-review` label** are the calls worth settling before this v1 locks the series. And the automated pass above raises fair follow-ups I'd second: parallelizing the sequential reads in `search`, whole-object `toEqual` in the metadata tests, and the PR **description** still saying "ranked by term frequency" / `~/.strands` (the *code* docstring is now correct — just the PR body is stale).

</details>

From my side the correctness concerns are resolved and verified — this is in good shape. 👍

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,184 @@
...
+      .filter(Boolean)
+  )
+}
+
+function basename(key: string): string {
+  const filename = key.split('/').pop() ?? key
+  return filename.replace(/\.md$/, '')
+}
+
+function slugify(text: string): string {
```
lizradway, line 84: nit: a comment explaining what this does could be helpful, same with other utility functions above

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,184 @@
...
+  let description: string
+  try {
+    description = JSON.parse(descMatch[1]!) as string
+  } catch {
+    description = descMatch[1]!.slice(1, -1)
+  }
+  return { description, body }
+}
+
+function tokenize(text: string): Set<string> {
```
lizradway: I believe `TestMemoryStore` uses the same functionality for keyword search. Maybe we could pull this out to a common keyword search module under `strands-ts/src/memory/search` or something like that.

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,184 @@
...
+  return text
+    .toLowerCase()
+    .replace(/[^a-z0-9\s-]/g, '')
+    .trim()
+    .replace(/\s+/g, '-')
+    .slice(0, 50)
+}
+
+/**
+ * A zero-infrastructure memory store backed by a file hierarchy.
```
lizradway: nit: the "zero-infrastructure" is not necessarily true since this is backed by the unified Storage interface and could be S3 etc

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,184 @@
...
+ * const agent = new Agent({
+ *   model,
+ *   memoryManager: new MemoryManager({ stores: [memoryStore], injection: false }),
+ * })
+ * ```
+ */
+export class FileMemoryStore implements MemoryStore {
+  readonly name: string
+  readonly description?: string
+  readonly writable: boolean
```
lizradway, line 122: nit: move this up in line with name since it's required?

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  private readonly _storage: Storage
+
+  constructor(config: FileMemoryStoreConfig) {
+    this.name = config.name
+    this.writable = config.writable ?? true
+    if (config.description !== undefined) this.description = config.description
+    if (config.maxSearchResults !== undefined) this.maxSearchResults = config.maxSearchResults
+    if (config.extraction !== undefined) this.extraction = config.extraction
+    this._storage = config.storage ?? new LocalFileStorage()
```
lizradway: might want to add to the JSDOC somewhere that this defaults to LocalFileStorage.

I think this is the right move in terms of default (cc: @mkmeral for default storage)

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  /**
+   * Search knowledge files by keyword matching against filenames, descriptions, and content.
+   *
+   * Returns the top matches ranked by distinct token overlap. Each result's `metadata.path`
+   * reflects the entry's current storage location and may change after consolidation.
+   */
+  async search(query: string, options?: SearchOptions): Promise<MemoryEntry[]> {
+    const maxResults = options?.maxSearchResults ?? this.maxSearchResults ?? 5
```
lizradway: nit: might be nice to have the max search results default be stored as a final.


but should we have this match `const DEFAULT_MAX_SEARCH_RESULTS = 10` from `TestMemoryStore`/BedrockKB Knowledge store.

lizradway: should we pull this out as well to be shared or do we see this diverging across memory stores? cc: @opieter-aws

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,184 @@
...
+        allKeys.map(async (key) => {
+          try {
+            const bytes = await this._storage.read(key)
+            if (!bytes) return null
+
+            const content = decoder.decode(bytes)
+            const { description, body } = parseFrontmatter(content)
+            const searchable = `${basename(key)} ${description} ${body}`
+
+            let score = 0
```
lizradway: nit: relevanceScore is more descriptive/self explainatory and matches what is used below.

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,184 @@
...
+   *
+   * Writes a markdown file with YAML frontmatter. By default writes to `knowledge/facts/`.
+   * Pass `metadata.path` to write to a custom location under `knowledge/`.
+   *
+   * @param content - The knowledge content to store
+   * @param metadata - Optional metadata: `title`, `description`, and `path` (custom target path)
+   */
+  async add(content: string, metadata?: Record<string, JSONValue>): Promise<string> {
+    const customPath = metadata?.['path'] as string | undefined
+    const title = (metadata?.['title'] as string | undefined) ?? slugify(content.split(/[.\n]/)[0]!.slice(0, 60))
```
lizradway: nit: should we just split out the first sentence in its own var first via `const firstSentence = content.split(/[.\n]/)[0]!` to simplify the regex expression in title / description?

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,184 @@
...
+    const description = (metadata?.['description'] as string | undefined) ?? content.split(/[.\n]/)[0]!.slice(0, 120)
+
+    let key: string
+    if (customPath) {
+      key = customPath.startsWith(KNOWLEDGE_PREFIX) ? customPath : `${KNOWLEDGE_PREFIX}${customPath}`
+      if (!key.endsWith('.md')) key += '.md'
+    } else {
+      const slug = slugify(title) || `entry-${Date.now()}`
```
lizradway, line 217: nit: i think we are sluggifying this twice if you want to simplify

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
maisieyanz: I created a DEFAULT_MAX_SEARCH_RESULTS constant in memory/search/keyword.ts (value = 10), and now have TestMemoryStore, BedrockKB, and my fileMemoryStore all import it.

opieter-aws: Perfect, shared const here is preferred

strands-ts/src/memory/search/keyword.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * Shared keyword-search utilities for memory stores that use lexical token-overlap scoring.
```
mkmeral, line 2: aren't there any simple libraries or sth that we can use?

strands-ts/src/vended-memory-stores/file-memory-store/__tests__/file-memory-store.test.node.ts
```diff
@@ -0,0 +1,410 @@
...
+const encoder = new TextEncoder()
+const decoder = new TextDecoder()
+
+describe('FileMemoryStore', () => {
+  let storage: InMemoryStorage
+  let store: FileMemoryStore
+
+  beforeEach(() => {
+    storage = new InMemoryStorage()
+    store = new FileMemoryStore({
```
mkmeral, line 19: Passing an in-memory storage to "File Memory Store" feels weird and awkward. Why?

mkmeral:
@strandly-the-agent  can you review this pr?

strands-ts/src/memory/search/index.ts
```diff
@@ -0,0 +1 @@
+export { DEFAULT_MAX_SEARCH_RESULTS, tokenize, tokenOverlapScore } from './keyword.js'
```
lizradway: we shouldnt be exporting these from the barrel export

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,179 @@
...
+    this._storage = config.storage ?? new LocalFileStorage()
+  }
+
+  /**
+   * Search knowledge files by keyword matching against filenames, descriptions, and content.
+   *
+   * Returns the top matches ranked by distinct token overlap. Each result's `metadata.path`
+   * reflects the entry's current storage location and may change after consolidation.
+   */
+  async search(query: string, options?: SearchOptions): Promise<MemoryEntry[]> {
```
mkmeral, line 158: why isn't this function just `grep -lrin "${}" ${KNOWLEDGE_PREFIX}`

This is downloading all of the files for every search, I feel like we can be more optimal here. We are having filesystem storage, double down on what filesystem can already do for you

strands-ts/src/memory/search/keyword.ts
```diff
@@ -0,0 +1,28 @@
...
+      .split(/[^\p{L}\p{N}_]+/u)
+      .filter(Boolean)
+  )
+}
+
+/**
+ * Lexical relevance score: the number of distinct query tokens that appear in the content.
+ * Returns 0 when there is no overlap.
+ */
+export function tokenOverlapScore(queryTokens: Set<string>, content: string): number {
```
mkmeral, line 22: should we use F1 score, or other metrics instead? is this a good metric?

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    let key: string
+    if (customPath) {
+      key = customPath.startsWith(KNOWLEDGE_PREFIX) ? customPath : `${KNOWLEDGE_PREFIX}${customPath}`
+      if (!key.endsWith('.md')) key += '.md'
+    } else {
+      const slug = slugify(title) || `entry-${Date.now()}`
+      key = `${FACTS_PREFIX}${slug}.md`
+
+      // Best-effort collision avoidance for a single-writer local store (TOCTOU is acceptable).
```
mkmeral: should we have a lock in the memory store to fix this? 🤔

strands-ts/src/vended-memory-stores/file-memory-store/types.ts
```diff
@@ -0,0 +1,14 @@
...
+ */
+
+import type { MemoryStoreConfig } from '../../memory/types.js'
+import type { Storage } from '../../storage/storage.js'
+
+/**
+ * Configuration for {@link FileMemoryStore}.
+ */
+export interface FileMemoryStoreConfig extends MemoryStoreConfig {
+  /** The unified Storage backend for file operations. Defaults to LocalFileStorage at `./.strands/`. */
```
mkmeral: why does it default to local file storage? It should default to sandbox. The main value of the filesystem memory is that agent knows how to use bash, search and index on bash. All the people who use this method are essentially onboarding to that value. 

In this implementation, we are custom-implementing search, addition and everything else

strands-ts/src/memory/search/keyword.ts
maisieyanz: Using a library wouldn't simplify this much, the function is already simple by intention and it would just add a dependency.

strands-ts/src/vended-memory-stores/file-memory-store/__tests__/file-memory-store.test.node.ts
maisieyanz: InMemoryStorage is just the lightest backend that's shipped that keeps the unit tests fast

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
maisieyanz: grep would tie this to a local filesystem, but FileMemoryStore is backed by the abstract Storage interface and we want to be able to extend to other backends.

strands-ts/src/vended-memory-stores/file-memory-store/types.ts
lizradway: local file storage (from unified storage) defaults to sandbox already if sandbox is configured !

strands-ts/src/memory/search/keyword.ts
lizradway: i think _eventually_ somewhere in memory/context we should vend specific search strategies independently that a user can plugin and specify.


I'm very interested in qmd for this

lizradway (dismissed):

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,179 @@
...
+      const existingKeys = new Set(await this._storage.list(FACTS_PREFIX))
```
strandly-the-agent: 🔴 **Preserve backend key identity when checking collisions.**

`existingKeys.has(key)` compares exact spellings returned by `list()`. That does not model backend identity on case-insensitive filesystems: an existing `knowledge/facts/Topic.md` and generated `knowledge/facts/topic.md` are the same local file, but this check misses the collision.

I reproduced this with a case-insensitive `Storage`: after preloading `Topic.md`, `add(..., { title: 'topic' })` left one key and replaced the original content. The prior backend-aware `read(candidate)` check detected this case.

Please retain a backend-aware existence probe before writing (or add an `exists` operation whose identity semantics each backend implements). Add a differing-case regression.

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,179 @@
...
+   */
+  async search(query: string, options?: SearchOptions): Promise<MemoryEntry[]> {
+    const maxResults = options?.maxSearchResults ?? this.maxSearchResults ?? DEFAULT_MAX_SEARCH_RESULTS
+    const queryTokens = tokenize(query)
+    if (queryTokens.size === 0) return []
+
+    const allKeys = await this._storage.list(KNOWLEDGE_PREFIX)
+
+    const scored = (
+      await Promise.all(
```
strandly-the-agent: 🔴 **Do not assume generic `Storage` supports corpus-sized concurrent reads.**

`Promise.all(allKeys.map(...))` starts every read immediately, and the per-entry `catch` turns throttling/concurrency errors into silent non-matches. The `Storage` interface specifies no concurrent-read capacity.

Repro: a conforming backend allowing one active read listed 3 matching keys; this returned only 1 because the other 2 reads failed concurrently and were suppressed. For S3, this also fans out one GET/body buffer per key.

Use bounded/configured concurrency (or keep generic storage reads sequential) while preserving per-entry error isolation.

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,179 @@
...
+      let suffix = 1
+      while (existingKeys.has(key)) {
+        key = `${FACTS_PREFIX}${slug}-${suffix}.md`
+        suffix++
+      }
+    }
+
+    const fileContent = `---\ndescription: ${JSON.stringify(description)}\n---\n\n${content}\n`
+    await this._storage.write(key, encoder.encode(fileContent))
+    return key
```
strandly-the-agent: 🟡 **Return the canonical key that was actually stored.**

The new receipt returns the pre-write string, but shipped backends canonicalize keys with `normalizeKey()`. Reproduced:

```text
add(..., { path: "operations//deploy" })
returned:    knowledge/operations//deploy.md
list/search: knowledge/operations/deploy.md
```

Normalize the final key before both `write()` and `return`, document it as the canonical storage-relative key, and assert the return value for default, custom, and collision-suffixed paths.

strandly-the-agent (changes requested):
**🔴 Changes requested on the new delta:** two storage regressions can silently lose data/results, and the new `add()` receipt is non-canonical.

✅ Scope: only `eca4d28..926262b`; I did **not** re-raise resolved earlier findings.
✅ Exact-head build + type-check + changed-file lint passed; focused tests **75/75** passed.
✅ Reproduced all three findings with throwaway tests against `926262b`.
⚠️ CI: TS Node 20/22/24 on Linux/macOS/Windows, code quality, package, docs, and CI Gate passed; the separate integration job failed. The PR currently conflicts with `main`.

<details><summary>Evidence and process notes</summary>

- **Canonical receipt:** `add('deploy', { path: 'operations//deploy' })` returned `knowledge/operations//deploy.md`; `InMemoryStorage.list()` and `search().metadata.path` exposed `knowledge/operations/deploy.md`.
- **Read fan-out:** a conforming custom `Storage` allowing one active read listed 3 matching keys; search returned only 1 because `Promise.all` triggered two overlap errors and the per-entry catch suppressed them.
- **Collision identity:** on a case-insensitive storage, existing `knowledge/facts/Topic.md` was overwritten by a default add titled `topic`; the case-sensitive `Set.has()` missed that both spellings identify the same key.
- `git diff --check` passed; worktree clean after repro cleanup.
- The shared scoring extraction/default of 10 and removal of the unused barrel look sound.
- This delta changes public `FileMemoryStore.add()` from `Promise<void>` to `Promise<string>`. If the informal API review has happened, please record it with the repo's `api/review-complete` workflow; neither API label is currently present.

</details>

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
mkmeral: then is it really a file memory store? is the "file" part the `.md` extension only? 🤔

lizradway: "hierarchical" memory store was the mental model @opieter-aws proposed

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
lizradway: @Unshure  and i had previously discussed adding "append" method to storage in the future.


i think i may go ahead and add this now to allow file memory store to use it.

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,213 @@
...
+const KNOWLEDGE_PREFIX = 'knowledge/'
+const FACTS_PREFIX = `${KNOWLEDGE_PREFIX}facts/`
+
+/**
+ * Cap on concurrent storage reads during search. The Storage contract makes no guarantee
+ * about concurrent-read capacity, so an unbounded fan-out (one read per key) can exhaust a
+ * backend's connection pool or trip throttling on a large corpus. Reads still run in parallel,
+ * just no more than this many at once.
+ */
+const SEARCH_READ_CONCURRENCY = 8
```
lizradway, line 34: tbh i dont think we need this, but nonblocking

lizradway (approved):

strands-ts/src/memory/search/keyword.ts
opieter-aws: For lexical search this is fine, I benchmarked it when I originally introduced it

strands-ts/src/vended-memory-stores/file-memory-store/file-memory-store.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  private readonly _storage: Storage
+
+  constructor(config: FileMemoryStoreConfig) {
+    this.name = config.name
+    this.writable = config.writable ?? true
+    if (config.description !== undefined) this.description = config.description
+    if (config.maxSearchResults !== undefined) this.maxSearchResults = config.maxSearchResults
+    if (config.extraction !== undefined) this.extraction = config.extraction
+    this._storage = config.storage ?? new LocalFileStorage()
```
opieter-aws: Shouldn't this set a default memory path here using namespace? @lizradway I assume this is also where we'd want to scope it to agentId? Or can we set up a router that does something like `namespace: memory` which then takes care of this for any memory solution?

lizradway: yes, i also like the router

lizradway (approved):

PR #3227, opened 2026-07-14
feat(storage): port unified storage interface to Python

strandly-the-agent:
## Description

Ports the **unified storage interface** from strands-ts (#3099) into the Python SDK.

Adds a new `strands.storage` package — a minimal, four-operation persistence primitive over opaque `bytes`, keyed by `/`-separated path-like strings — plus three shipped backends and a `namespace()` factory. This is the Python twin of the strands-ts primitive; it establishes the same foundation for the phased "single persistence primitive for the SDK" design.

```python
from strands.storage import InMemoryStorage, LocalFileStorage, S3Storage

storage = LocalFileStorage("./.strands/")
await storage.write("sessions/abc/snapshot.json", data)   # -> None
raw = await storage.read("sessions/abc/snapshot.json")     # -> bytes | None
keys = await storage.list("sessions/")                      # -> sorted list[str]
await storage.delete("sessions/abc/snapshot.json")

scoped = storage.namespace("offloader")   # a prefixed view; composable
```

| Class | Backend |
|---|---|
| `InMemoryStorage` | `dict` (testing / serverless) |
| `LocalFileStorage` | filesystem, atomic host writes (tmp + `os.replace`), sandbox-aware via `for_sandbox()` |
| `S3Storage` | Amazon S3 via boto3 (lazy client, paginated `list`, `region`/`boto_session` mutual-exclusion) |

Also adds `StorageError` (`strands.types.exceptions`). Key normalization collapses `/` runs, strips leading/trailing `/`, and rejects empty keys and any `..` segment (path-traversal guard). `list` is a lexicographically-sorted prefix match.

## Related Issues

Port of strands-ts #3099. Requested by `lizradway` on that PR.

## Scope — primitive + context-offloader integration

This PR ships the unified primitive **and** wires the context-offloader onto it (per maintainer request on this PR), plus a sandbox parity fix. The one TS consumer still not ported is the **session `SnapshotStorageAdapter`**, which has no faithful Python analog: strands-ts sessions are snapshot-based (`SnapshotStorage`/`SnapshotLocation`/manifest) while strands-py sessions are **repository-based** (`SessionManager` ABC + `repository_session_manager` + `file`/`s3` managers) — there is no `SnapshotStorage` interface to adapt onto. That one is genuinely N/A rather than deferred.

### Context-offloader bridged onto the unified interface

`ContextOffloader` now accepts **either** a unified `strands.storage.Storage` (`write`/`read`/`delete`/`list`) **or** the legacy offloader `Storage` (`store`/`retrieve`, now deprecated), mirroring strands-ts #3099:

- **Detection:** `_is_offloader_storage` routes on `hasattr(store) and hasattr(retrieve)` (TS `'store' in storage && 'retrieve' in storage`).
- **Content-type framing:** for unified backends each block occupies one key, with the content-type framed into the bytes — `[2-byte BE content-type length][utf-8 content-type][content]` (`_frame_content`/`_unframe_content`); `_store_content`/`_retrieve_content` adapt between backends.
- **Plugin-driven eviction:** unified `Storage` has no built-in eviction, so the plugin tracks the store cycle per key and deletes entries older than a new keyword-only `evict_after_cycles` (default 20, `None` disables) on `BeforeModelCallEvent`; the window is forwarded to a legacy `InMemoryStorage` left at its default. `_on_before_model_call` is now async (the event loop dispatches this event via `invoke_callbacks_async`, which awaits coroutine hooks).
- **Deprecation:** the legacy offloader `Storage`/`InMemoryStorage`/`FileStorage`/`S3Storage` get doc-only `.. deprecated::` notes (mirrors TS's doc-only `@deprecated`).

### LocalFileStorage sandbox not-found parity

Sandbox-mode `read`/`delete`/`_list_keys_sandbox` now treat `NotADirectoryError` as not-found alongside `FileNotFoundError`, matching the host paths and TS `isNotFoundError` (ENOENT+ENOTDIR). `write` is intentionally unchanged (TS doesn't treat a non-directory parent as not-found on write).

## Type of Change

New feature

<details>
<summary><b>Behavior traceability — every strands-ts storage test has a Python counterpart</b></summary>

All 53 source `it(...)` behaviors across the four TS test files map to a Python test (0 missing); the Python suite adds a few extra edge tests. Independently verified by a fresh-context validation pass.

| TS test file | behaviors | Python test file |
|---|---|---|
| `in-memory-storage.test.ts` | 16 | `test_in_memory_storage.py` |
| `namespaced-storage.test.ts` | 10 | `test_namespaced_storage.py` (+ `.namespace()` method compose) |
| `local-file-storage.test.node.ts` | 14 | `test_local_file_storage.py` (+ forced rename-failure cleanup) |
| `s3-storage.test.ts` | 13 | `test_s3_storage.py` (+ truncated-without-token termination) |

**58 Python tests, all passing.**
</details>

<details>
<summary><b>Decision log — deviations from the source and why</b></summary>

- **`Storage` is a non-generic `Protocol`.** TS uses `Storage<ListQuery = string>`. PEP 696 `TypeVar` defaults are Python 3.13+ and the SDK targets 3.10+, so the generic-with-default isn't expressible. Dropped it: `list(prefix: str)`. A backend can still *widen* the parameter (contravariance keeps it structurally compatible), so no capability is lost.
- **`NAMESPACED` symbol → marker attribute.** TS uses `Symbol.for('strands.storage.namespaced')` + `NAMESPACED in obj`. Python has no symbols; the equivalent is a module-level marker attribute-name string, set via `setattr` and detected via `getattr`. Kept internal (not exported).
- **Copy-on-write only.** TS copies bytes on both read and write (`Uint8Array.slice()`). Python `bytes` is immutable, so `InMemoryStorage` copies via `bytes(data)` on write only; reads hand out the immutable object. The TS "copies on read" test becomes an immutability assertion.
- **`S3Storage` config = keyword args, boto conventions.** TS `S3StorageConfig` (`region` XOR `s3Client`) → explicit kwargs `prefix` / `region` / `boto_session` / `boto_client_config`, matching the existing `S3SessionManager` (mutual exclusion is `region` XOR `boto_session`; `user_agent_extra="strands-agents"`). boto3 is a hard dep so there's no lazy-import gap, but the client is still created lazily so constructing an `S3Storage` never needs AWS.
- **Top-level export.** TS surfaces the `Storage` type from its root barrel. Python idiom is submodule access (`from strands.storage import ...`), consistent with `strands.session`/`strands.memory`, so `strands/__init__.py` is left untouched. Easy to add a re-export if preferred.
- **Shared latent edges kept faithful to TS (not "fixed"):** the scratch-file exclusion uses the same `.__strands_tmp` substring check as TS (`local-file-storage.ts:191/218`), and directory listing treats an unknown `is_dir` (`None`/`undefined`) the same way TS does. Diverging would break cross-SDK parity; flagged here for a future cross-SDK fix instead.
</details>

<details>
<summary><b>One genuine divergence found in review — fixed</b></summary>

An independent review pass caught that `S3Storage.list` could infinite-loop if a backend returned a page with `IsTruncated=true` but no `NextContinuationToken`. strands-ts terminates in that case (`continuationToken = IsTruncated ? NextContinuationToken : undefined; while (continuationToken)`). Fixed to match, with a regression test. (Unreachable with a spec-compliant S3, but now equivalent to the source.)
</details>

<details>
<summary><b>Sensitive-surface scan</b></summary>

- **Path traversal:** `..` segments rejected in both key and prefix normalization (matches TS); normalized keys can't escape `base_dir`.
- **Filesystem:** atomic tmp + `os.replace`; scratch file removed on failure.
- **AWS:** boto3 client lazy; no credentials logged; error messages contain only bucket + normalized key.
- **No subprocess, no deserialization** (storage round-trips opaque bytes).
</details>

## Testing

Ran locally (no functional warnings):

- `pytest tests/strands/storage/ tests/strands/vended_plugins/context_offloader/` — **235 passed** (+18 new for the offloader bridge and sandbox parity)
- `ruff check` + `ruff format --check` on touched source & tests — clean
- `mypy ./src` — clean in `storage/` and `context_offloader/` (`boto3-stubs[s3]`)
- Full `tests/strands/vended_plugins/` — **487 passed** (no regressions)
- Two independent fresh-context review passes: correctness → *ship*; API bar-raiser → *land-these-then-approve* (all items addressed: privatized the internal `_StorageBackend` alias, documented `evict_after_cycles` + unified-storage acceptance, `bool` rejected in validation).

- [x] Ran the equivalent of `hatch run prepare` (ruff check, ruff format --check, mypy, pytest) directly — hatch not available in this environment.

## Checklist

---

🤖 Generated by Strandly (an experimental AI agent). A human should review before merge. Docstrings and a `site/` docs page are intentionally deferred — happy to add them or split into a docs PR.

### DIFF
--- a/strands-py/src/strands/storage/__init__.py
+++ b/strands-py/src/strands/storage/__init__.py
@@ -1,0 +1,24 @@
+"""Unified storage module.
+
+Provides the :class:`Storage` protocol and shipped implementations for persisting
+raw bytes under string keys. It is the SDK's single low-level persistence
+primitive — a minimal four-operation contract (``write``, ``read``, ``delete``,
+``list``) over opaque ``bytes`` values, keyed by ``/``-separated path-like strings.
+
+Example:
+    ```python
+    from strands.storage import InMemoryStorage, LocalFileStorage, S3Storage
+    ```
+"""
+
+from .base import Storage
+from .in_memory_storage import InMemoryStorage
+from .local_file_storage import LocalFileStorage
+from .s3_storage import S3Storage
+
+__all__ = [
+    "Storage",
+    "InMemoryStorage",
+    "LocalFileStorage",
+    "S3Storage",
+]
--- a/strands-py/src/strands/storage/__init__.py
+++ b/strands-py/src/strands/storage/__init__.py
@@ -11,14 +11,17 @@ Example:
     ```
 """
 
-from .base import Storage
+from ..types.exceptions import StorageError
+from .base import Storage, namespace
 from .in_memory_storage import InMemoryStorage
 from .local_file_storage import LocalFileStorage
 from .s3_storage import S3Storage
 
 __all__ = [
     "Storage",
+    "StorageError",
     "InMemoryStorage",
     "LocalFileStorage",
     "S3Storage",
+    "namespace",
 ]
--- a/strands-py/src/strands/storage/base.py
+++ b/strands-py/src/strands/storage/base.py
@@ -1,0 +1,209 @@
+"""Core contract and helpers for the unified storage primitive.
+
+This module defines the :class:`Storage` protocol — the SDK's single persistence
+primitive — plus internal helpers for key normalization and namespacing. The
+shipped backends live in sibling modules and are re-exported from the package
+root.
+"""
+
+from __future__ import annotations
+
+import re
+from typing import Protocol, runtime_checkable
+
+from ..types.exceptions import StorageError
+
+# Attribute name marking a namespaced storage view. Constructs check for it (via
+# ``getattr``) to detect whether the caller already scoped the storage, so they
+# can skip applying a default prefix. This is the Python analog of the TypeScript
+# ``NAMESPACED`` symbol.
+NAMESPACED = "_strands_storage_namespaced"
+
+_SLASH_RUN = re.compile(r"/+")
+
+
+def normalize_key(key: str) -> str:
+    """Validate and normalize a storage key.
+
+    Collapses runs of ``/``, strips leading and trailing ``/``, and rejects empty
+    keys and any ``..`` segment.
+
+    Args:
+        key: The raw key to normalize.
+
+    Returns:
+        The normalized key.
+
+    Raises:
+        StorageError: If the key is empty or contains a ``..`` segment.
+    """
+    normalized = _SLASH_RUN.sub("/", key).strip("/")
+    if not normalized:
+        raise StorageError("Storage key must not be empty")
+    if ".." in normalized.split("/"):
+        raise StorageError(f"Invalid storage key '{key}': '..' path segments are not allowed")
+    return normalized
+
+
+def normalize_prefix(prefix: str) -> str:
+    """Normalize a list prefix.
+
+    Collapses slash runs and strips leading slashes. Unlike a key, an empty prefix
+    is valid and matches everything, and a trailing slash is preserved.
+
+    Args:
+        prefix: The raw prefix to normalize.
+
+    Returns:
+        The normalized prefix.
+
+    Raises:
+        StorageError: If the prefix contains a ``..`` segment.
+    """
+    normalized = _SLASH_RUN.sub("/", prefix).lstrip("/")
+    if ".." in normalized.split("/"):
+        raise StorageError(f"Invalid storage prefix '{prefix}': '..' path segments are not allowed")
+    return normalized
+
+
+@runtime_checkable
+class Storage(Protocol):
+    """A backend for storing and retrieving raw bytes under string keys.
+
+    The interface is deliberately minimal — four operations over opaque ``bytes``
+    values. Implementations must treat keys as opaque path-like strings (segments
+    separated by ``/``) and must round-trip the bytes they are given unchanged.
+
+    By default ``list`` performs a prefix match over a string, which every backend
+    supports. An implementation may *widen* the accepted ``prefix`` parameter to a
+    richer query object (e.g. a DynamoDB partition/sort-key filter) while still
+    accepting a plain string for SDK-internal callers.
+
+    Implement this to add a custom backend; the SDK ships :class:`InMemoryStorage`,
+    :class:`LocalFileStorage`, and :class:`S3Storage`. Those shipped backends also
+    expose a ``namespace(prefix)`` convenience method; a custom backend can instead
+    be scoped with the standalone :func:`namespace` factory.
+
+    Example:
+        ```python
+        from strands.storage import InMemoryStorage
+
+        storage = InMemoryStorage()
+        await storage.write("memory/notes.json", b"[]")
+        data = await storage.read("memory/notes.json")
+        ```
+    """
+
+    async def write(self, key: str, data: bytes) -> None:
+        """Store ``data`` under ``key``, overwriting any existing value.
+
+        Args:
+            key: Opaque, ``/``-separated key identifying the value.
+            data: Raw bytes to persist.
+
+        Raises:
+            StorageError: If the key is invalid or the write fails.
+        """
+        ...
+
+    async def read(self, key: str) -> bytes | None:
+        """Retrieve the bytes previously stored under ``key``.
+
+        Args:
+            key: The key to read.
+
+        Returns:
+            The stored bytes, or ``None`` if no value exists for ``key``.
+
+        Raises:
+            StorageError: If the read fails for a reason other than a missing key.
+        """
+        ...
+
+    async def delete(self, key: str) -> None:
+        """Delete the value stored under ``key``. A no-op if the key does not exist.
+
+        Args:
+            key: The key to delete.
+
+        Raises:
+            StorageError: If the delete fails.
+        """
+        ...
+
+    async def list(self, prefix: str) -> list[str]:
+        """List keys matching the given prefix.
+
+        Returns full keys (not the suffix after the prefix), sorted
+        lexicographically. An empty string lists every key.
+
+        Args:
+            prefix: A key prefix. An empty string matches all keys.
+
+        Returns:
+            The matching keys, sorted ascending.
+
+        Raises:
+            StorageError: If the listing fails.
+        """
+        ...
+
+
+class _NamespacedStorage:
+    """A :class:`Storage` view that prefixes every key with a fixed namespace.
+
+    Returned by :func:`namespace`. Delegates to an underlying storage, prepending
+    the prefix on write/read/delete/list and stripping it back off listed keys.
+    """
+
+    def __init__(self, storage: Storage, prefix: str) -> None:
+        """Initialize the view.
+
+        Args:
+            storage: The underlying storage to delegate to.
+            prefix: The already-normalized prefix, including a trailing ``/`` when
+                non-empty (as produced by :func:`namespace`).
+        """
+        self._storage = storage
+        self._prefix = prefix
+        setattr(self, NAMESPACED, True)
+
+    async def write(self, key: str, data: bytes) -> None:
+        """Store ``data`` under the namespaced key."""
+        await self._storage.write(f"{self._prefix}{key}", data)
+
+    async def read(self, key: str) -> bytes | None:
+        """Read the bytes stored under the namespaced key."""
+        return await self._storage.read(f"{self._prefix}{key}")
+
+    async def delete(self, key: str) -> None:
+        """Delete the value stored under the namespaced key."""
+        await self._storage.delete(f"{self._prefix}{key}")
+
+    async def list(self, prefix: str) -> list[str]:
+        """List keys under the namespace, with the namespace prefix stripped off."""
+        keys = await self._storage.list(f"{self._prefix}{prefix}")
+        offset = len(self._prefix)
+        return [key[offset:] for key in keys]
+
+    def namespace(self, prefix: str) -> Storage:
+        """Return a further-nested namespaced view of the underlying storage."""
+        return namespace(self._storage, f"{self._prefix}{prefix}")
+
+
+def namespace(storage: Storage, prefix: str) -> Storage:
+    """Return a :class:`Storage` view with all keys prefixed by ``prefix``.
+
+    The original storage is not mutated. Composable — calling ``namespace()`` on
+    the result nests prefixes. An empty prefix yields a transparent pass-through.
+
+    Args:
+        storage: The underlying storage to delegate to.
+        prefix: Prefix to prepend to all keys.
+
+    Returns:
+        A namespaced Storage view.
+    """
+    normalized = normalize_prefix(prefix)
+    scoped = f"{normalized}/" if normalized else ""
+    return _NamespacedStorage(storage, scoped)
--- a/strands-py/src/strands/storage/base.py
+++ b/strands-py/src/strands/storage/base.py
@@ -9,7 +9,7 @@ root.
 from __future__ import annotations
 
 import re
-from typing import Protocol, runtime_checkable
+from typing import Any, Protocol, runtime_checkable
 
 from ..types.exceptions import StorageError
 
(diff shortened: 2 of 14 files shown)


strandly-the-agent:
**Update:** ran an independent API bar-raiser pass over the new public surface (it was missing from my earlier validation) and pushed the fixes — the two must-fix-now-or-never public-shape items are addressed. `58` storage tests pass; ruff/format/mypy clean; `279` existing types+offloader tests green.

<details>
<summary><b>What the api-bar-raiser flagged and how I addressed it</b></summary>

| # | Finding | Action |
|---|---|---|
| 🔴 1 | `Storage` protocol methods had no `**kwargs` — violates strands-py's own `STYLE_GUIDE.md`/`AGENTS.md` forward-compat rule (the `Sandbox` ABC does this) and is unfixable post-release | **Fixed** — `**kwargs: Any` on the protocol + all backend methods |
| 🔴 2 | `S3Storage(region=…)` diverged from the established boto3 spelling (`S3SessionManager` and boto3 use `region_name`) | **Fixed** — renamed `region` → `region_name` |
| 🟡 3 | `namespace()` factory was internal but the docstring named it as the scoping path for custom backends; a `Storage`-typed value couldn't be scoped | **Fixed** — `namespace()` now exported from `strands.storage` |
| 🟡 4 | `StorageError` didn't travel with the API that raises it (only in `strands.types.exceptions`; cf. the sandbox errors re-export) | **Fixed** — re-exported from `strands.storage` |
| ⚪ 6 | `NAMESPACED` spelled like a public constant but internal | **Fixed** — `_NAMESPACED` |
| ✅ 7 | Affirmed as correct: dropping the TS `<ListQuery=string>` generic (PEP 696 defaults are 3.13+; SDK is 3.10+), `Protocol` over ABC, `list` method name, keyword-only `sandbox` | no change |

Verdict moved from API-changes-requested → the raiser's "land 1–4 and this is an easy API-approve."
</details>

**Two things I'd like a maintainer's call on** (I deliberately didn't force these):

1. **`needs-api-review` label** — this is a net-new public primitive; I can't apply labels from a fork PR. Worth the gate.
2. **Name collision with `vended_plugins.context_offloader`** — `InMemoryStorage`/`S3Storage` now exist in *both* modules with different interfaces (`write/read/delete/list` vs `store/retrieve`), and `LocalFileStorage` vs the offloader's `FileStorage`. They're in separate modules so it's not a hard conflict, but it's a real DevX/discoverability wrinkle. Since the offloader isn't migrated onto this primitive yet, the unification path (does the offloader eventually adopt `strands.storage` and drop its bespoke trio?) is a design decision I'd rather not make unilaterally. Happy to do the migration as a follow-up once that's settled.

opieter-aws:
@strandly-the-agent is there any drift between the TS implementation / api and this python implementation?

strandly-the-agent:
Short answer: **yes, there's drift — but it's almost all deliberate Python-idiom adaptation, plus one scope difference.** The *contract* is faithful: the four-op `write`/`read`/`delete`/`list` interface over opaque bytes, key/prefix normalization (collapse `/` runs, strip, reject `..`), `namespace()` composition, lexicographically-sorted prefix `list`, atomic local writes, and S3 pagination (incl. the truncated-without-token termination) all match `strands-ts/src/storage`. I read both trees line by line.

The divergences worth *your* eye are: (1) **scope** — this ports only the primitive, not the two TS consumers; (2) `namespace()`/`StorageError` are **public** in Python but internal in TS; (3) the **S3 injection model** differs (boto `Session` vs an `S3Client`); (4) **boto3 is an eager hard dep** vs TS's lazy optional peer-dep. Details below.

<details>
<summary><b>1. Scope — the biggest difference (intentional)</b></summary>

TS #3099 also wired the primitive into two consumers; this PR ships **only the primitive**. Per the PR body, the session `SnapshotStorageAdapter` has no analog (strands-py sessions are repository-based, not snapshot-based) and the context-offloader already has its own `Storage` protocol, so bridging it is a behavioral change deferred to a follow-up. Reasonable — but it does mean parity is at the *primitive* layer only, not the integration layer.
</details>

<details>
<summary><b>2. Public API surface — Python is wider than TS</b></summary>

- **`namespace()`** is `@internal` in TS and **not** exported from its barrel (`strands-ts/src/storage/index.ts:16-20`). Python exports it from `strands.storage` (`__init__.py:15,26`).
- **`StorageError`** lives in TS's root `errors` module; Python puts it in `strands.types.exceptions` **and** re-exports it from `strands.storage`.
- **`S3StorageConfig`** is an exported type in TS (`index.ts:20`); Python has no equivalent — S3 options are keyword args.

These were deliberate calls from my own API-bar-raiser pass (see my earlier comment, items #3/#4). Flagging because they widen the Python public surface relative to TS — worth a conscious "yes, keep them public."
</details>

<details>
<summary><b>3. S3 configuration & dependency posture (genuine drift)</b></summary>

| | strands-ts | strands-py |
|---|---|---|
| Inject client | `s3Client: S3Client` | `boto_session: boto3.Session` (**a session, not a client**) |
| Region arg | `region` | `region_name` (boto convention) |
| Mutual exclusion | `s3Client` XOR `region` | `boto_session` XOR `region_name` |
| Extra knobs | — | `boto_client_config`, `user_agent_extra="strands-agents"` |
| AWS dep | lazy `import()`, **optional** peer-dep | `import boto3` at module top (`s3_storage.py:7`), **hard dep** |

The injection model genuinely differs: TS lets you hand in a fully-built client; Python takes a `Session` instead. And because `strands/storage/__init__.py` imports `s3_storage`, merely doing `from strands.storage import InMemoryStorage` eagerly pulls in boto3 — whereas TS never loads the AWS SDK unless you construct an `S3Storage`. This matches strands-py's existing `S3SessionManager` conventions (boto3 is already a core dep), so it's defensible, but it *is* a real posture difference from TS.
</details>

<details>
<summary><b>4. Expected idiom adaptations (not concerns)</b></summary>

`async def`, snake_case, `Protocol`+`@runtime_checkable` vs `interface`, `bytes` vs `Uint8Array`, copy-on-write only (Python `bytes` is immutable), the `NAMESPACED` symbol → private `_strands_storage_namespaced` marker attr, keyword-only `sandbox`, and dropping the `Storage<ListQuery=string>` generic (PEP 696 `TypeVar` defaults are 3.13+; the SDK targets 3.10+ — a backend can still widen the param structurally). All faithful in behavior.
</details>

<details>
<summary><b>5. Two minor behavioral edges I'd note</b></summary>

- **S3 not-found codes:** Python treats `{NoSuchKey, NotFound, 404}` as “missing → return None” (`s3_storage.py:18`); TS only checks `NoSuchKey`/`NotFound`. The extra `404` is a defensible superset (boto surfaces it on some paths) but it's a small behavioral widening vs TS.
- **LocalFileStorage sandbox mode:** host-mode `read`/`delete` catch both `FileNotFoundError` **and** `NotADirectoryError` (matching TS's `isNotFoundError`, which lumps `ENOENT`+`ENOTDIR`), but the **sandbox** paths catch only `FileNotFoundError` — so a `NotADirectoryError` from a sandbox raises `StorageError` where TS would return null. Host-vs-sandbox asymmetry, low impact, but not 1:1 with TS. Easy tightening if you want exact parity.
</details>

Net: the primitive is a faithful port; the drift is intentional idiom/scope choices plus the S3 injection & dependency-posture differences above. The public-surface items (#2) and the two edges (#5) are the only things I'd ask you to explicitly bless or wave off.

🤖 Generated by Strandly (an experimental AI agent) — please sanity-check the specifics before relying on them.

opieter-aws:
@strandly-the-agent 

> LocalFileStorage sandbox mode: host-mode read/delete catch both FileNotFoundError and NotADirectoryError (matching TS's isNotFoundError, which lumps ENOENT+ENOTDIR), but the sandbox paths catch only FileNotFoundError — so a NotADirectoryError from a sandbox raises StorageError where TS would return null. Host-vs-sandbox asymmetry, low impact, but not 1:1 with TS. Easy tightening if you want exact parity.

Fix this

>  the context-offloader already has its own Storage protocol, so bridging it is a behavioral change deferred to a follow-up.

Do full parity with the TS PR, implement for offloader too

strandly-the-agent:
Both done and pushed (2 commits on this branch). **pytest / ruff / mypy green**, and I ran two independent fresh-context review passes over the result (correctness → *ship*; API bar-raiser → *land-these-then-approve*, all items addressed).

1. **Sandbox `NotADirectoryError` parity** — fixed in `LocalFileStorage`. All three sandbox catch sites (`read` / `delete` / `_list_keys_sandbox`) now treat `NotADirectoryError` as not-found alongside `FileNotFoundError`, matching the host paths and TS `isNotFoundError` (ENOENT+ENOTDIR). `write` is deliberately left alone — TS doesn't treat a non-directory parent as not-found on write either.
2. **Offloader bridged onto the unified interface** — `ContextOffloader` now accepts a unified `strands.storage.Storage` *or* the legacy offloader `Storage` (now deprecated), mirroring strands-ts #3099.

One honest scope note before the details: the TS PR *also* wired the **session `SnapshotStorageAdapter`**. That one still has no faithful Python analog (strands-py sessions are repository-based, not snapshot-based — there's no `SnapshotStorage` to adapt onto), so it's genuinely N/A rather than deferred. You asked specifically for the offloader, which is done.

<details>
<summary><b>How the offloader bridge works (parity with strands-ts)</b></summary>

- **Dual-accept + detection:** `_is_offloader_storage` routes on `hasattr(store) and hasattr(retrieve)` (TS `'store' in storage && 'retrieve' in storage`). Unified backends expose only `write/read/delete/list`, so there's no misroute.
- **Content-type framing:** unified backends store one key per block with the content-type framed into the bytes — `[2-byte big-endian content-type length][utf-8 content-type][content]` (`_frame_content`/`_unframe_content`), so metadata doesn't consume a second key. `_store_content`/`_retrieve_content` adapt between the two backends.
- **Plugin-driven eviction:** unified `Storage` has no built-in eviction, so the plugin tracks the store cycle per key (`_key_stored_at`) and deletes entries older than `evict_after_cycles` (new keyword-only param, default 20, `None` disables) on `BeforeModelCallEvent`. The window is forwarded to a legacy `InMemoryStorage` left at its default, so one knob governs both. `_on_before_model_call` is now async — verified the event loop dispatches this event via `invoke_callbacks_async`, which awaits coroutine hooks.
- **Deprecation:** legacy `Storage`/`InMemoryStorage`/`FileStorage`/`S3Storage` in the offloader module got doc-only `.. deprecated::` notes (mirrors TS's doc-only `@deprecated`; no runtime warning).
</details>

<details>
<summary><b>Tests & verification</b></summary>

- **+18 tests** (`test_unified_storage_bridge.py`, 17 cases + sandbox parity cases in `test_local_file_storage.py`): framing round-trip (incl. empty content / empty content-type / 256-byte binary), backend detection both ways, store/retrieve adapters, end-to-end offload→retrieve through a unified `InMemoryStorage`, eviction boundary (`stored_cycle < threshold`) + disable, `evict_after_cycles` validation (incl. `bool` rejection), legacy-window forwarding (both branches), and `NotADirectoryError` parity across read/delete/list (plus “other errors still wrap to `StorageError`”).
- **235 storage+offloader tests pass**; `ruff check` + `ruff format --check` clean on touched files; `mypy ./src` clean in `storage/` and `context_offloader/`.
- Diff: +479/-29 across 6 files.

**Faithful-to-TS edges I did *not* “fix”** (flagging rather than diverging): `_key_stored_at` isn't pruned when eviction is disabled, and unified eviction is measured from store-time (not last-access, unlike the legacy in-memory backend). Both mirror strands-ts exactly — happy to tighten either as a cross-SDK follow-up if you'd prefer.
</details>

🤖 Generated by Strandly (an experimental AI agent) — please review before merge.

lizradway:
Closing out in favor of https://github.com/strands-agents/harness-sdk/pull/3259

PR #1062, opened 2025-10-21
fix(models/gemini): handle non-JSON error messages from Gemini API

Ratish1:
## Description
 Avoid a raw JSONDecodeError when the Gemini API returns an empty or non-JSON error message (for example on an invalid API key).

This PR:

- Catches json.JSONDecodeError when parsing error.message, logs a warning with context (error_message=<...> | Gemini API returned non-JSON error), and re-raises the original genai.errors.ClientError with the JSON error as the explicit cause.

- Adds a unit test (test_stream_handles_non_json_error) to verify a non-JSON error is logged and the original ClientError is propagated

## Related Issues

#1024 

## Documentation PR

 N/A

## Type of Change

Bug fix

## Testing

How have you tested the change?  Verify that the changes do not break functionality or introduce warnings in consuming repositories: agents-docs, agents-tools, agents-cli

- [x] I ran `hatch run prepare`
- [x]  I ran hatch fmt --formatter and hatch fmt --linter 
- [x] I ran the unit test: pytest -q tests/strands/models/test_gemini.py::test_stream_handles_non_json_error -> passed

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [ ] I have updated the documentation accordingly
- [ ] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #1024, opened 2025-10-12
[BUG] JSONDecodeError when using Gemini models with free tier API key


jztan:
### Checks

- [x] I have updated to the lastest minor and patch version of Strands
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched [./issues](./issues?q=) and there are no duplicates of my issue

### Strands Version

1.12.0

### Python Version

3.10.17

### Operating System

macOS 26.0.1

### Installation Method

pip

### Steps to Reproduce

1. Create a Google AI Studio free tier API key from https://aistudio.google.com/
2. Use the API key with GeminiModel in Strands
3. Create an agent with tools
4. Invoke the agent with a query that triggers tool use

**Minimal reproduction code:**

```python
import os
from strands import Agent, tool
from strands.models.gemini import GeminiModel

@tool
def calculator(expression: str) -> str:
    """Evaluate a mathematical expression."""
    return str(eval(expression))

model = GeminiModel(
    client_args={"api_key": os.getenv("GOOGLE_API_KEY")},
    model_id="gemini-2.0-flash",
    params={"max_output_tokens": 500, "temperature": 0.7}
)

agent = Agent(model=model, tools=[calculator])
response = agent("What is 25 * 4?")  # Crashes here
```

### Expected Behavior

When the Gemini API returns an error (authentication failure, invalid response, etc.), Strands should:

1. Catch the error gracefully
2. Provide a clear error message (e.g., "API authentication failed", "Invalid API response", or "Gemini API configuration error")
3. Optionally raise a more specific exception like `ModelThrottledException` for rate limits
4. NOT crash with a generic `JSONDecodeError` that doesn't indicate the root cause

### Actual Behavior

Strands crashes with:
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

This error is unhelpful because it doesn't indicate the root cause (API authentication, billing, permissions, rate limiting, etc.).

**Full traceback:**
[...]

Issue #370, opened 2025-07-07
[FEATURE] Retries for ServiceUnavailableException (503 errors)


austinmw:
### Problem Statement

If I'm reading correctly, it seems that retry handling only captures ThrottlingException; if so, would it be reasonable to capture and retry other intermittent errors like ServiceUnavailableException as well?

### Use Case

ServiceUnavailableException can happen often with certain models, like Opus 4

zastrowm:
Not directly related, but I'm curious if this could be achieved via hooks.  We could convert our current retry behavior to something that fires on the `AfterModelInvocationEvent` and have a `retry`/`retry_delay` properties on the event indicating if it should be retried.  That would allow callers to configure which exceptions they retry on

NateLangloisProscia:
I believe this is related, but let me know if not and I'll be happy to open a new issue.

I am using strands (1.1.0) BedrockModel for the strands Agent's `model` parameter, and I tried using boto_client_config to provide retries to handle 503. It didn't seem to work to handle the retries, since I see exactly 1 invocationError in CloudWatch Bedrock dashboard, and my logs suggest that my code for catching the service unavailable exception still got hit; if it were handling retries, I would have expected to see 3 InvocationServerErrors before hitting my catch block. Here's a snippet[1] and stack trace[2]. Is this expected? If so, why?

[1]
```python
sdk_retry_config = BotoConfig(retries={"total_max_attempts": 3, "mode": "standard"})
model = BedrockModel(
    model_id="your model id",
    boto_client_config=sdk_retry_config,
    **settings.inference_config,
)
agent = Agent(
    model=model,
    system_prompt=system_prompt,
    session_manager=session_manager,
    tools=tools,
    trace_attributes={"correlation.trace_id": trace_id},
)
try:
    result = await agent_wrapper.invoke_async(message)
except Exception as e:
    logger.exception(f"Error during chat: {str(e)}")
```

[2]
```
ERROR - Error during chat: An error occurred (serviceUnavailableException) when calling the ConverseStream operation: Bedrock is unable to process your request.
Traceback (most recent call last):
  File "/app/my_app/app.py", line 227, in invoke_async
    return await self.agent.invoke_async(message)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.12/site-packages/strands/agent/agent.py", line 398, in invoke_async
    async for event in events:
[...]

PR #3228, opened 2026-07-14
fix(models/gemini): detect throttling via ClientError status attribute

### LINKED ISSUE #3226: [BUG] GeminiModel does not convert Vertex AI 429 with plain-text error message to ModelThrottledException
## Checks

## SDK Language

Python

## Strands Version

1.47.0 (first observed on 1.45.0; the relevant code on `main` is unchanged)

## Language Runtime Version

Python 3.12

## Operating System

macOS 15 (Darwin 25.4.0)

## Installation Method

pip

## Steps to Reproduce

When Vertex AI (`vertexai=True`) returns a 429 under concurrent load, the error arrives as a `genai.errors.ClientError` whose `message` is **plain text**, not JSON:

```
429 RESOURCE_EXHAUSTED. Resource exhausted. Please try again later. ...
```

`GeminiModel.stream()` detects throttling by `json.loads(error.message)` and matching `message["error"]["status"]`. With a plain-text message, `json.loads` raises `JSONDecodeError`, the handler logs a warning and re-raises the raw `ClientError` — so it is never converted to `ModelThrottledException` and the event loop's built-in retry (`ModelRetryStrategy`) never fires.

Deterministic reproduction (mirrors what the google-genai SDK raises mid-stream for a Vertex AI 429):

```python
import asyncio
from unittest import mock

from google import genai

from strands.models.gemini import GeminiModel

# What Vertex AI's 429 looks like: structured code/status, plain-text message
error = genai.errors.ClientError(
    429,
    {"error": {"status": "RESOURCE_EXHAUSTED", "message": "Resource exhausted. Please try again later."}},
)
assert error.code == 429
assert error.status == "RESOURCE_EXHAUSTED"

model = GeminiModel(model_id="gemini-2.5-flash", client_args={"api_key": "test"})
with mock.patch.object(genai.Client, "aio", new_callable=mock.PropertyMock) as aio:
    aio.return_value.models.generate_content_stream.side_effect = error
    events = model.stream([{"role": "user", "content": [{"text": "hi"}]}])
    asyncio.run(events.__anext__())  # raises raw ClientError, not ModelThrottledException
```

Real-world trigger: Vertex AI's pay-as-you-go Gemini models use Dynamic Shared Quota, so any concurrent traffic can receive a 429 regardless of configured quota. In my case, 1 out of 30 requests at concurrency 4 hit this (gemini model, `global` endpoint), and the error surfaced directly to the user instead of being retried.

## Expected Behavior

A 429 (`RESOURCE_EXHAUSTED`) from Vertex AI is converted to `ModelThrottledException`, regardless of whether `error.message` happens to be JSON, so the built-in exponential-backoff retry in the event loop handles it.

## Actual Behavior

The raw `genai.errors.ClientError` propagates out of `stream()`:

```
strands  | error_message=<Resource exhausted. Please try again later. ...> | Gemini API returned non-JSON error
...
google.genai.errors.ClientError: 429 RESOURCE_EXHAUSTED. Resource exhausted. Please try again later. ...
```

No retry is attempted; the request fails immediately.

## Additional Context

- #1062 fixed the `JSONDecodeError` crash for non-JSON error messages (#1024), but the non-JSON path re-raises the raw `ClientError` without ever checking for throttling — which is exactly the path a Vertex AI 429 takes.
- The SDK-level HTTP retry (`HttpRetryOptions`) does not help here: the 429 is raised while iterating the response stream (`errors.APIError.raise_error_async`), which is outside the request-issuing wrapper that `retry_options` covers.
- The TypeScript SDK has the same root cause: `classifyGoogleError()` in `src/models/google/errors.ts` also relies on `JSON.parse(error.message)` and returns `undefined` for non-JSON messages, so a plain-text 429 is not classified as throttling there either.

## Possible Solution

`genai.errors.APIError` already exposes structured `code` and `status` attributes (populated by the google-genai SDK from the response), so the handler doesn't need to parse `error.message` at all:

```python
except genai.errors.ClientError as error:
    match error.status:
        case "RESOURCE_EXHAUSTED" | "UNAVAILABLE":
            raise ModelThrottledException(error.message) from error
        case "INVALID_ARGUMENT":
            ...
```

I'd be happy to submit a PR for the Python SDK with this fix.

## Related Issues

#1024, #1062, #370

Ryutaro-A:
## Description

Vertex AI returns 429s whose `message` is plain text rather than a JSON document. `GeminiModel.stream()` detected throttling by parsing `error.message` as JSON, so these errors were re-raised as raw `ClientError`s instead of `ModelThrottledException` — the event loop's built-in retry never fired and the 429 surfaced directly to callers. Since pay-as-you-go Gemini on Vertex AI uses Dynamic Shared Quota, any concurrent traffic can hit this.

The google-genai SDK already populates structured `code`/`status`/`message` attributes on `ClientError` from the response, so the handler now classifies errors by `error.status` instead of parsing `error.message`. This makes throttling and context-overflow detection independent of the message format (the attributes behave identically on the lowest supported google-genai version, 1.32.0).

The non-JSON warning path added in #1062 is removed along with the JSON parsing: classification no longer depends on the message format, so there is no non-JSON case left to warn about. Unclassifiable errors are re-raised unchanged, as before.

## Related Issues

Fixes #3226

Related: #1024, #1062, #370

## Type of Change

Bug fix

## Testing

Also reproduced the failure end to end with a mocked Vertex AI-style 429 (structured status, plain-text message): before this change the raw `ClientError` escaped `stream()`; after it, the error is converted to `ModelThrottledException` and the built-in retry handles it. Gemini integration tests (`tests_integ`) were not run locally (no credentials); relying on CI for those.

## Checklist

----

### DIFF
--- a/strands-py/src/strands/models/gemini.py
+++ b/strands-py/src/strands/models/gemini.py
@@ -594,21 +594,11 @@ class GeminiModel(Model):
                 yield self._format_chunk({"chunk_type": "metadata", "data": event.usage_metadata})
 
         except genai.errors.ClientError as error:
-            if not error.message:
-                raise
-
-            try:
-                message = json.loads(error.message) if error.message else {}
-            except json.JSONDecodeError as e:
-                logger.warning("error_message=<%s> | Gemini API returned non-JSON error", error.message)
-                # Re-raise the original ClientError (not JSONDecodeError) and make the JSON error the explicit cause
-                raise error from e
-
-            match message["error"]["status"]:
+            match error.status:
                 case "RESOURCE_EXHAUSTED" | "UNAVAILABLE":
-                    raise ModelThrottledException(error.message) from error
+                    raise ModelThrottledException(error.message or str(error)) from error
                 case "INVALID_ARGUMENT":
-                    if "exceeds the maximum number of tokens" in message["error"]["message"]:
+                    if error.message and "exceeds the maximum number of tokens" in error.message:
                         raise ContextWindowOverflowException(error.message) from error
                     raise error
                 case _:
--- a/strands-py/tests/strands/models/test_gemini.py
+++ b/strands-py/tests/strands/models/test_gemini.py
@@ -1,4 +1,3 @@
-import json
 import logging
 import unittest.mock
 
@@ -921,11 +920,16 @@ async def test_stream_response_empty_stream(gemini_client, model, messages, agen
 
 @pytest.mark.asyncio
 async def test_stream_response_throttled_exception(gemini_client, model, messages):
+    """Regression test for https://github.com/strands-agents/harness-sdk/issues/3226.
+
+    Vertex AI returns 429s whose message is plain text rather than a JSON document; the status
+    attribute alone must be enough to classify them as throttling.
+    """
     gemini_client.aio.models.generate_content_stream.side_effect = genai.errors.ClientError(
-        429, {"message": '{"error": {"status": "RESOURCE_EXHAUSTED"}}'}
+        429, {"error": {"status": "RESOURCE_EXHAUSTED", "message": "Resource exhausted. Please try again later."}}
     )
 
-    with pytest.raises(ModelThrottledException, match="RESOURCE_EXHAUSTED"):
+    with pytest.raises(ModelThrottledException, match="Resource exhausted. Please try again later."):
         await anext(model.stream(messages))
 
 
@@ -934,18 +938,14 @@ async def test_stream_response_context_overflow_exception(gemini_client, model, 
     gemini_client.aio.models.generate_content_stream.side_effect = genai.errors.ClientError(
         400,
         {
-            "message": json.dumps(
-                {
-                    "error": {
-                        "message": "request exceeds the maximum number of tokens (100)",
-                        "status": "INVALID_ARGUMENT",
-                    },
-                }
-            ),
+            "error": {
+                "message": "request exceeds the maximum number of tokens (100)",
+                "status": "INVALID_ARGUMENT",
+            },
         },
     )
 
-    with pytest.raises(ContextWindowOverflowException, match="INVALID_ARGUMENT"):
+    with pytest.raises(ContextWindowOverflowException, match="exceeds the maximum number of tokens"):
         await anext(model.stream(messages))
 
 
@@ -1060,18 +1060,14 @@ async def test_stream_request_with_gemini_tools_and_function_tools(gemini_client
 
 
 @pytest.mark.asyncio
-async def test_stream_handles_non_json_error(gemini_client, model, messages, caplog, alist):
+async def test_stream_handles_non_json_error(gemini_client, model, messages, alist):
     error_message = "Invalid API key"
     gemini_client.aio.models.generate_content_stream.side_effect = genai.errors.ClientError(
-        error_message, {"message": error_message}
+        400, {"error": {"message": error_message}}
     )
 
-    with caplog.at_level(logging.WARNING):
-        with pytest.raises(genai.errors.ClientError, match=error_message):
-            await alist(model.stream(messages))
-
-    assert "Gemini API returned non-JSON error" in caplog.text
-    assert f"error_message=<{error_message}>" in caplog.text
+    with pytest.raises(genai.errors.ClientError, match=error_message):
+        await alist(model.stream(messages))
 
 
 @pytest.mark.asyncio


yonib05:

strandly-the-agent (commented):
**✅ The fix is correct and well-targeted — no blockers.** Classifying on `ClientError.status` instead of `json.loads(error.message)` is the right call, and it happens to fix a *second* latent bug (context-overflow detection never worked on real structured responses either). Two non-blocking test-coverage gaps on branches this PR touches; worth adding before calling coverage complete, but nothing that blocks merge. Framing this as work for a human to approve.

### What I verified
| Check | Result |
|---|---|
| Branch / SHA | `fix/gemini-429-throttling` @ `7cf2477` |
| Unit tests | `pytest tests/strands/models/test_gemini.py` → **56 passed** (on `google-genai==1.32.0`, the floor) |
| SDK attribute behavior | Read `google/genai/errors.py` @ 1.32.0 — `.status`/`.message` populated from top-level **or** nested `error.*`; confirmed with live repros |
| Regression tests genuine? | New tests **fail on old code, pass on new** (ran both) — not tautologies |
| Reported case (structured envelope, plain-text inner msg) | Fixed ✅ |
| Non-JSON-*body* edge (status→reason phrase) | Same behavior old vs new → **pre-existing gap, not a regression** |

<details>
<summary>Why the fix is sound (correctness detail)</summary>

The google-genai SDK's `_get_message()` returns the *inner* `error.message` string, never the whole JSON envelope re-encoded — so the old `json.loads(error.message)` could only ever succeed on a doubly-encoded payload that real responses don't produce. That's why the old `test_stream_response_throttled_exception` (which fed `{"message": '{"error":...}'}`, a doubly-encoded string) passed while production 429s failed. The new fixture `{"error": {"status": "RESOURCE_EXHAUSTED", "message": "..."}}` matches what `raise_for_async_response` actually builds from a Vertex JSON body, so the test now drives the real extraction path.

Edge cases all hold up: `error.message or str(error)` gives a meaningful non-empty string when the inner message is absent; the `error.message and ...` guard on the `INVALID_ARGUMENT` branch avoids a `TypeError` when message is `None`; and `status=None` / envelope-without-`status` cleanly hits `case _` → re-raise (the old code could raise a masking `KeyError` there, so the new code is actually more robust).
</details>

### 🟡 Should-fix (test coverage — non-blocking)

**1. No `UNAVAILABLE` coverage.** `gemini.py:598` merges `"RESOURCE_EXHAUSTED" | "UNAVAILABLE"` into the throttling branch, but only `RESOURCE_EXHAUSTED` is exercised. `UNAVAILABLE` (503) is a distinct real Vertex status on that same branch. Cheapest fix is to parametrize the existing test:

```python
@pytest.mark.asyncio
@pytest.mark.parametrize("status", ["RESOURCE_EXHAUSTED", "UNAVAILABLE"])
async def test_stream_response_throttled_exception(gemini_client, model, messages, status):
    gemini_client.aio.models.generate_content_stream.side_effect = genai.errors.ClientError(
        429, {"error": {"status": status, "message": "Resource exhausted. Please try again later."}}
    )
    with pytest.raises(ModelThrottledException, match="Resource exhausted. Please try again later."):
        await anext(model.stream(messages))
```

**2. The new `error.message or str(error)` fallback is untested.** That `or str(error)` was added by this PR (`gemini.py:599`) and no test drives the empty/None-message path:

```python
@pytest.mark.asyncio
async def test_stream_throttled_exception_without_message(gemini_client, model, messages):
    # Vertex can raise a throttle with only a status and no message field.
    gemini_client.aio.models.generate_content_stream.side_effect = genai.errors.ClientError(
        429, {"error": {"status": "RESOURCE_EXHAUSTED"}}
    )
    with pytest.raises(ModelThrottledException, match="RESOURCE_EXHAUSTED"):
        await anext(model.stream(messages))
```

### ⚪ Nits (optional)

- **`test_stream_handles_non_json_error` is now a misnomer** — the new code does no JSON parsing at all; the test really asserts *unclassified errors re-raise unchanged*. Dropping the caplog assertions is correct (the warning was removed with the JSON branch), but consider renaming to e.g. `test_stream_reraises_unclassified_error`.
- **Minor observability loss:** removing the `#1062` `logger.warning("...non-JSON error")` means a non-throttleable/unclassified error now propagates with no log breadcrumb. The exception still carries full detail, so this is a judgment call, not a defect.
- **Out-of-scope follow-up (not this PR):** if a 429 ever arrives with a genuinely non-JSON *HTTP body* (e.g. an HTML gateway page), the SDK sets `status` to the HTTP reason phrase (`"Too Many Requests"`), which won't match — so it wouldn't be retried. The old code didn't handle this either, so it's a pre-existing gap, not a regression. A `case _` heuristic on `error.code in (429, 503)` would close it if it ever shows up in the wild.

---
<sub>🤖 Automated review by Strandly (correctness + adversarial + test-quality passes). I can be wrong — a human should confirm before merging.</sub>

JackYPCOnline (approved):

PR #3260, opened 2026-07-14
refactor(memory): refactor TestMemoryStore to use the unified storage interface

### LINKED ISSUE #3253: [Feature] Adopt unified storage interface in TestMemoryStore
Once the unified storage interface is ported to Python (#3227), migrate TestMemoryStore to use the Storage primitive.

opieter-aws:
## Description

`TestMemoryStore` hand-rolled its own filesystem persistence in both SDKs — path resolution, atomic temp-file writes, plus a records cache with load memoization to paper over cold-load races. The unified `Storage` interface exists precisely to delete this per-subsystem re-invention, and the storage design doc names the memory store as a consumer to migrate. This moves persistence onto a `Storage` backend internally.

The public config is unchanged: `persist`/`path` and the `~/.strands/memory/<name>.json` default behave exactly as before. Internally the store now resolves a backend from that config — `persist: false` → `InMemoryStorage`, otherwise a `LocalFileStorage` rooted so an explicit `path` still writes to that exact file and the default still lands at the home-dir location. Dropping the bespoke cache also let the read-modify-write in `add` read fresh under a lock, closing the cold-load race by construction rather than guarding against it.

## Related Issues

#3099 

#3253

## Documentation PR

N/A

## Type of Change

Other (please describe): internal refactor — no public API or behavior change

## Testing

Ran the `TestMemoryStore` unit suites in both SDKs (Python pytest, TS Node), plus type-check, lint, and format for each; all pass. Exercised each store end to end against both backends, confirming the default home-dir path, an explicit `path`, `persist: false` ephemerality, dedup, and cross-loop reuse from a synchronous agent all behave as before.

## Checklist

----


opieter-aws:
Docs haven't yet been released, the TestMemoryStore was released last week. So breaking change seems acceptable to me.

lizradway (dismissed):
lgtm, any reason why storage was persistent before that we'd want to maintain? slash should we eventually add an option on storage for this or can we wait/yagni?

opieter-aws:
> lgtm, any reason why storage was persistent before that we'd want to maintain? slash should we eventually add an option on storage for this or can we wait/yagni?

Already supported! That's done by doing

```
new TestMemoryStore({ name: 'notes', storage: new LocalFileStorage() })   // → ./.strands/memory/notes.json
new TestMemoryStore({ name: 'notes', storage: new S3Storage('my-bucket') }) // → S3
```

If no `storage` is supplied, it defaults to `inMemory` since that's the most logical for a `Test` setup IMO and needs no cleanup

strands-py/src/strands/vended_memory_stores/test_memory_store/store.py
```diff
@@ -124,20 +130,18 @@ def __init__(self, **store_config: Unpack[TestMemoryStoreConfig]) -> None:
...
-
-        # Records load lazily on first read/write so construction never touches the filesystem.
-        self._records: list[dict[str, Any]] | None = None
-        self._lock = threading.Lock()
+        # Ephemeral by default. Scope every backend under `memory/`, unless the caller already namespaced it.
+        backend = store_config.get("storage")
+        if backend is None:
+            backend = InMemoryStorage()
+        self._storage: Storage = (
+            backend if getattr(backend, "_namespaced", None) is _NAMESPACED else _NamespacedStorage(backend, "memory")
```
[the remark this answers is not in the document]
opieter-aws: Ack, but this is a change to the storage module, which this PR otherwise doesn’t touch. Not part of memory store migration.

mkmeral:
@strandly-the-agent review this PR please. Also highlight all public API changes, so we can API bar raise. if there are any behavior changes, new params, failure mode changes, and so on.

strands-py/src/strands/vended_memory_stores/test_memory_store/store.py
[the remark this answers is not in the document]
lizradway: yeah im curious for a better approach on this in python long term. might add namespace as optional to the api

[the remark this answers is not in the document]
lizradway: kept it all internal for now so we can change later

lizradway (dismissed):

mkmeral:
@strandly-the-agent don't you like me? where's my response? 😢

mkmeral (dismissed):
why are we making a breaking change? can't we solve this another way?

strandly-the-agent (dismissed):
🔴 **Latest implementation: changes requested — one Python data-loss blocker remains at `541fb5d5`.**
✅ Exact head `541fb5d5e23118511982b7db279359b3e109298b` matches the remote; `git diff --check` and worktree are clean.
✅ Focused verification: Python **42/42**; TypeScript **36/36**, no type errors.
✅ The prior sequential two-`asyncio.run()` failure is fixed: 4/4 records persist. Public `persist`/`path`, the persistent default, and prior file locations are restored.
🔴 Simultaneous threads/loops still return two successful `add()` IDs while persisting one record: [repro and fix guidance](https://github.com/strands-agents/harness-sdk/pull/3260#discussion_r3616084193).
🟡 An invalid explicit path can make nonempty search return `[]` instead of surfacing the configuration error: [details](https://github.com/strands-agents/harness-sdk/pull/3260#discussion_r3616085087).
**API:** no exported shape changed, but concurrency/error/fresh-read contracts did. Keep `api/needs-review`; no API meeting is needed.

<details>
<summary>Latest delta and observable behavior</summary>

### Restored / unchanged
- Python and TypeScript retain `persist`/`path`; there is no public `storage` parameter.
- Default persistence remains `~/.strands/memory/<sanitized-name>.json`; explicit `path` still selects that file; `persist=false` remains per-instance ephemeral.
- Exports, `search`/`add` signatures, result types, ranking, deduplication, record fields, and normal data locations are unchanged.
- The prior eager-loop-bound lock finding is resolved and regression-tested. Its old inline thread is obsolete/outdated; this review does not re-raise it.

### Remaining observable deltas from the internal Storage refactor
- Every token-bearing search and every valid add rereads storage instead of using the prior instance cache, so external edits/deletes/corruption become visible.
- `persist=false` now JSON-serializes records, so unsupported Python metadata can fail where the prior in-memory list accepted it.
- Backend failures now surface as `StorageError` and generally name the key/basename rather than the old full-path `OSError`/generic `Error` message.
- Malformed non-object metadata is rejected eagerly.
- LocalFileStorage retains atomic temp-file replacement for individual writes, but the Python instance read-modify-write transaction is not serialized across simultaneous loops.

### CI snapshot
Functional checks completed so far are green across Linux Python 3.10–3.14, most Windows/macOS Python jobs, TypeScript Node 20/22/24 Linux/macOS, lint, code quality, audit, package install, and patch coverage. Some Windows/docs/integration jobs were still running at review time; the aggregate failure was the expected API-label policy gate.

</details>

<details>
<summary>Routed review passes</summary>

- **Correctness / adversarial:** sequential-loop reuse now passes. A deterministic simultaneous two-thread/two-loop barrier repro returns two IDs but retains one record; merge-base serialized this same-instance case.
- **API / DevX:** public shape/default/location compatibility is restored. The PR body’s “no behavior change” wording is broader than the actual fresh-read, error, validation, serialization, and concurrency deltas.
- **Test quality:** the new yielding sequential-loop regression is strong. The missing simultaneous-loop barrier test directly hides the remaining blocker; lower-value filename/ephemeral/TS-recovery gaps were suppressed.
- **Docs accuracy:** prior breaking-change and namespacing feedback is stale and not repeated. Remaining wording nits were suppressed behind the concrete concurrency and invalid-path findings.
- **Model context:** this remains agent-visible: concurrent synchronous invocations can receive successful memory IDs for records that were overwritten, while ENOTDIR currently looks like “no memories.”

</details>

opieter-aws:
We could keep the store backwards compatible with this interface

```
export interface TestMemoryStoreConfig extends MemoryStoreConfig {
  /** Storage backend to persist through. Preferred over path/persist. */
  storage?: Storage
  /** @deprecated Pass `storage: new InMemoryStorage()` instead. */
  persist?: boolean
  /** @deprecated Pass `storage: new LocalFileStorage(dir)` instead. */
  path?: string
}
```

But given that:
1. No docs yet released
2. Zero hits on Github for usage, zero internal

I think we can still do a breaking change here.

pgrayy:
What if `TestMemoryStore` kept the same constructor arguments but initialized `LocalFileStorage` for you under the hood? Does `TestMemoryStore` need anything other than `LocalFileStorage`?

opieter-aws:
@strandly-the-agent review latest implementation

strands-py/src/strands/vended_memory_stores/test_memory_store/store.py
```diff
@@ -124,20 +129,36 @@ def __init__(self, **store_config: Unpack[TestMemoryStoreConfig]) -> None:
...
+        # Persistence runs on the unified Storage interface. Resolve a (backend, key) pair whose
+        # on-disk location matches the pre-Storage behavior exactly:
+        #   persist=False       -> ephemeral in-memory store
+        #   persist=True + path  -> the file at `path` (backend rooted at its parent dir)
+        #   persist=True default -> ~/.strands/memory/<sanitized-name>.json
+        # LocalFileStorage/InMemoryStorage construction touches no filesystem, so building the store
+        # never does I/O.
+        if not persist:
+            self._storage: Storage = InMemoryStorage()
+            self._key = f"{_sanitize_name(self.name)}.json"
```
zastrowm, line 146: Why does the key seem like a file name?

strands-py/src/strands/vended_memory_stores/test_memory_store/store.py
```diff
@@ -198,21 +221,21 @@ async def add(self, content: str, metadata: Metadata | None = None) -> TestMemor
...
-        # the same snapshot and clobber one another (last-write-wins). Within a single event loop the
-        # synchronous critical section is already atomic; the lock additionally guards a store shared
-        # across OS threads.
-        with self._lock:
-            records = self._load()
+        # The lock serializes the whole read-modify-write cycle so concurrent adds on the same event
+        # loop don't each read the same snapshot and clobber one another. Reading inside the critical
+        # section guarantees add #N sees add #N-1's write. Serialization is per event loop; adds
+        # driven from separate loops/processes against a shared file remain last-write-wins.
+        async with self._get_lock():
```
zastrowm, line 237: Should we up level something like this to the `Agent` or is this really only protecting memory from multiple agents? 

From my reading it's the latter but confirming

strands-py/src/strands/vended_memory_stores/test_memory_store/store.py
```diff
@@ -198,21 +221,21 @@ async def add(self, content: str, metadata: Metadata | None = None) -> TestMemor
...
 
-        # The lock serializes the whole load-modify-flush cycle so concurrent adds don't each load
-        # the same snapshot and clobber one another (last-write-wins). Within a single event loop the
-        # synchronous critical section is already atomic; the lock additionally guards a store shared
-        # across OS threads.
-        with self._lock:
-            records = self._load()
+        # The lock serializes the whole read-modify-write cycle so concurrent adds on the same event
+        # loop don't each read the same snapshot and clobber one another. Reading inside the critical
+        # section guarantees add #N sees add #N-1's write. Serialization is per event loop; adds
```
zastrowm, line 235: > Serialization is per event loop

What happens if two agents on two different loops attempt to read/write at the same time?  Any protection for multiple event loops?

strands-py/src/strands/vended_memory_stores/test_memory_store/store.py
opieter-aws: In the `LocalFileStorage` the key maps directly to a filename. The `.json` suffix keeps the on disk file byte-identical to the old implementation.

strands-py/src/strands/vended_memory_stores/test_memory_store/store.py
opieter-aws: There's no protection across loops, two agents each using the same test store on their own event loop don’t share a lock. The last write wins.

This is intentional for the test store. With the async storage backend we can't reuse the old locking mechanism and cross-loop coordination would need backend support that the storage interface currently doesn't implement. 

I will call this out in the docs to make it explicit.

LMK if you think we need to reconsider this setup.

strands-py/src/strands/vended_memory_stores/test_memory_store/store.py
opieter-aws: It's the latter indeed, the lock only serializes concurrent writes against a single store instance

strands-py/src/strands/vended_memory_stores/test_memory_store/store.py
zastrowm: But this is for the `InMemoryStorage` - does that still write to the file system? (I would have expected not)

strands-py/src/strands/vended_memory_stores/test_memory_store/store.py
zastrowm: This would be a bug right?  I'm fine deferring it as a fix, but we're effectively saying that Memory is not intended to be shared by concurrent agents?

strands-py/src/strands/vended_memory_stores/test_memory_store/store.py
opieter-aws: Ah missed that, for `InMemoryStore` indeed this does not write to the file system, it's just an in memory dict. The key just pairs with `LocalFileStorage`, but it's easy to drop it here if you think it's confusing

strands-py/src/strands/vended_memory_stores/test_memory_store/store.py
opieter-aws: It's a bit more subtle, we're saying that the `TestMemoryStore` is not intended to be used concurrently across loops. I think it's fine to defer the multi-loop write story to production backends for now, while we add concurrent write mechanism to the Storage object (this is a larger undertaking). But I agree we need to solve this, not just for this use case. I will create an issue with follow up if you agree.

strands-py/src/strands/vended_memory_stores/test_memory_store/store.py
zastrowm: If you make an edit, I'd say add a comment to clarify; I was just mostly concerned with "is this is a mistake"

zastrowm (approved):

Issue #2548, opened 2026-06-01
[FEATURE] Add content-aware offloading to ContextOffloader


lizradway:
### Problem

The `ContextOffloader` currently offloads tool results based purely on token count (`maxResultTokens` threshold). It doesn't distinguish between source code the agent needs to reason about vs. noise like directory listings or command output.

Benchmarking on ContextBench shows this causes a **regression on focused tasks** — when the agent reads a relevant source file, the offloader truncates it, forcing the agent to re-read (wasting tokens) or miss relevant spans past the preview cutoff.

**Evidence from benchmarking:**
- Task 1 (exploration-heavy, 85 cycles): Offloading gives **TAR 4-6x** improvement over control
- Task 2 (focused, single file answer): Offloading gives **TAR 0.4-0.85** — worse than no offloading

This is consistent with findings from SWE-agent, which explicitly treats file content differently from command output in their history management.

### Proposed Solution

Add a `shouldOffload` callback to `ContextOffloaderConfig` that lets users (and the SDK's defaults) control what gets offloaded:

```typescript
export interface ContextOffloaderConfig {
  storage: Storage
  maxResultTokens?: number
  previewTokens?: number
  includeRetrievalTool?: boolean

  // NEW
  shouldOffload?: (context: OffloadContext) => boolean
}

export interface OffloadContext {
  toolName: string
  toolInput: unknown
  resultTokens: number
  recency: number              // how many tool calls ago (0 = most recent)
  previousFileReads: string[]  // files already read in this session
}
```

**Default heuristic (if no callback provided):**
1. Never offload the last 3 tool results (recency protection)
2. If the same file was read before, offload the earlier read (file deduplication)
3. Otherwise, offload if result exceeds `maxResultTokens`

### Prior Art

- **SWE-agent** uses tag-based selective retention (`keep_output`/`remove_output` per tool), file-view deduplication (only keep latest view of each file), and recency-based elision (keep last N observations full, summarize older ones)
- **Aider** uses head/tail summarization with no tool-specific logic
- Our current implementation is token-threshold-only with no content awareness

### Acceptance Criteria

- [ ] `shouldOffload` callback option added to `ContextOffloaderConfig`
- [ ] Default heuristic implements recency protection + file deduplication
- [ ] Existing behavior unchanged when no callback is provided (backward compatible)
- [ ] Benchmark shows improved TAR on focused tasks without regressing exploration-heavy tasks

### Related

[...]

Issue #3045, opened 2026-07-01
[FEATURE] EPIC ContextManager interface


yonib05:
## Summary

Deliver a unified `contextManager` interface on `Agent` (both SDKs) that pre-composes context management into a single opt-in configuration — tool-result caching/offloading plus proactive and reactive compression — so builders get production-quality context handling without discovering, configuring, and composing the primitives by hand.

The initial design from last cycle (captured in #2562, which landed a first facade via #2643) did not ship as the full interface we want. This issue delivers that interface.

**Blocked by** the Unified Storage Interface (companion issue) — the ContextManager composes storage-backed strategies and should target the unified storage contract rather than wiring backends directly.

## Motivation

Every non-trivial agent eventually hits context limits. The building blocks already exist but must be found, configured, and understood individually:
- `ContextOffloader` (`strands-ts/src/vended-plugins/context-offloader/`, `strands-py/.../vended_plugins/context_offloader/`) — offloads oversized tool results to storage, keeps a preview in context, registers a retrieval tool.
- `SummarizingConversationManager` and the sliding-window managers (`strands-ts/src/conversation-manager/`, `strands-py/.../agent/conversation_manager/`).
- Proactive compression thresholds.

A single `contextManager` construct should unify these settings and ship sensible, benchmark-validated defaults.

## Requirements

- [ ] Design the ContextManager interface and lifecycle; unify existing CM settings into one construct that replaces the manual per-primitive setup.
- [ ] Configurable retention / compression policies (sliding window + summarization, proactive threshold + reactive on overflow).
- [ ] Priority-based context allocation and tool-result offloading via the unified storage interface.
- [ ] Coexistence rules with a user-provided conversation manager or offloader (don't override an explicitly configured one).
- [ ] Behavior for stateful/server-side-context models (they manage context server-side).
- [ ] Performance validation on long conversations.
- [ ] Align with the team on the interface (this cycle's initial task).

## Deliberate cross-language divergence to respect

- Python has a sliding-window tool-pair trim fallback (`_find_tool_pair_trim_point` in `sliding_window_conversation_manager.py`) with no TS equivalent, and a `per_turn` proactive mode; TS has richer tool-result truncation (typed placeholders, document-source-aware). The unified interface should expose common configuration while allowing these to remain where they already provide value, rather than forcing identical internals.

[...]

PR #3267, opened 2026-07-14
feat(context-offloader): add should_offload callback for selective offloading

### LINKED ISSUE #3266: [FEATURE] Add should_offload callback to ContextOffloader
### SDK Language

Python

### Problem Statement

`ContextOffloader` offloads tool results based purely on token count. It cannot distinguish between tools by role:

- **Listing tools** (e.g., search): Agent needs the full result to judge relevance. Should NOT be offloaded.
- **Retrieval tools** (e.g., get full document): Large output best accessed on-demand. Should be offloaded.

Adjusting `max_result_tokens` doesn't solve this — output sizes vary per query and overlap between tools. The offload decision depends on the tool's role, not its size on any single call.

### Proposed Solution

Add a `should_offload` callback to `ContextOffloader`:

```python
ContextOffloader(
    storage=InMemoryStorage(),
    should_offload=lambda tool_name, token_count: tool_name in {"get_document_text"},
)
```

Signature:

```python
should_offload: Callable[[str, int], bool] | None = None
```

- `None` (default): current behavior — all tools offloaded by token threshold. Fully backward compatible.
- Provided: called after the token threshold check. Returns `False` → result stays in context regardless of size.

This is a minimal subset of #2548 (TypeScript). Default heuristics (recency, dedup) are out of scope — they can follow separately.

### Use Case

RAG/Search agents with mixed tool output sizes:

- **Document retrieval** (10–50KB): offload, access via `retrieve_offloaded_content`
- **Search/listing** (1–8KB, varies per query): agent needs full list for ranking — must NOT be offloaded

Without this, users subclass `ContextOffloader` and override the internal `_handle_tool_result` hook — which works but couples to an internal API.

### Alternatives Solutions

1. **Subclass workaround**: Override `_handle_tool_result`. Works today but relies on internal API that may change.
2. **Full `shouldOffload` with `OffloadContext` (#2548)**: More powerful (recency, dedup, tool input). Larger scope — this PR is a stepping stone.
3. **`tool_filter` list**: Simpler but less flexible. A callback subsumes it (`lambda name, _: name in {...}`).

### Additional Context

- Related: #2548 (TypeScript, full design with `OffloadContext`)
- Scope: Python only. TS port is a follow-up.
- Implementation: ~30 lines in `plugin.py`, PR ready.

maniryu:
## Description

Add a `should_offload` callback parameter to `ContextOffloader` that lets users control which tool results are offloaded based on tool name and token count. When provided, the callback is invoked after the token threshold check — returning `False` skips offloading for that result.

### Changes

- **New `ShouldOffload` Protocol** (`plugin.py`): Extensible callback type using `Protocol` with `**kwargs` for forward compatibility, following the project style guide.
- **Updated `ContextOffloader.__init__`**: Added `should_offload: ShouldOffload | None` keyword argument. When `None` (default), existing behavior is preserved (backward compatible).
- **Updated `_handle_tool_result` hook**: Guard clause after the token threshold check that consults the callback.
- **Updated `__init__.py`**: Export `ShouldOffload` from the package.
- **Tests** (`test_plugin.py`): 6 new tests covering callback invocation, filtering by tool name, and backward compatibility. Extended `_make_event` helper with `tool_name` parameter.

## Related Issues

Closes #3266
Related: #2548 (TypeScript, full `shouldOffload` design)

## Documentation PR

Documentation updated in this PR (docstrings and inline examples).

## Type of Change

New feature

## Testing

- [x] I ran `hatch test` — 65/65 tests pass
- [x] `ruff check` and `ruff format` clean

## Checklist

----

### DIFF
--- a/strands-py/src/strands/vended_plugins/context_offloader/plugin.py
+++ b/strands-py/src/strands/vended_plugins/context_offloader/plugin.py
@@ -27,6 +27,16 @@ Example:
             include_retrieval_tool=True,
         )
     ])
+
+    # Selective offloading: only offload results from specific tools
+    agent = Agent(plugins=[
+        ContextOffloader(
+            storage=InMemoryStorage(),
+            should_offload=lambda tool_name, token_count, **kwargs: (
+                tool_name == "get_document_text"
+            ),
+        )
+    ])
     ```
 """
 
@@ -35,7 +45,7 @@ from __future__ import annotations
 import json
 import logging
 import weakref
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Protocol
 
 from typing_extensions import TypedDict
 
@@ -138,6 +148,21 @@ _CHARS_PER_TOKEN = 4
 """Approximate characters per token, fallback for preview slicing without tiktoken."""
 
 
+class ShouldOffload(Protocol):
+    """Callback protocol for deciding whether a tool result should be offloaded."""
+
+    def __call__(self, tool_name: str, token_count: int, **kwargs: Any) -> bool:
+        """Return True to offload, False to keep the result in context.
+
+        Args:
+            tool_name: Name of the tool that produced the result.
+            token_count: Estimated token count of the result.
+            **kwargs: Reserved for future parameters. Implementations should accept
+                ``**kwargs`` for forward compatibility.
+        """
+        ...
+
+
 class ContextOffloader(Plugin):
     """Plugin that offloads oversized tool results to reduce context consumption.
 
@@ -168,6 +193,8 @@ class ContextOffloader(Plugin):
         preview_tokens: Number of tokens to keep as a text preview in context.
         include_retrieval_tool: Whether to register the ``retrieve_offloaded_content`` tool.
             Defaults to True.
+        should_offload: Callback to control which tool results are offloaded.
+            Defaults to None (all oversized results offloaded).
 
     Example:
         ```python
@@ -176,6 +203,16 @@ class ContextOffloader(Plugin):
 
         agent = Agent(plugins=[
             ContextOffloader(storage=InMemoryStorage())
+        ])
+
+        # Only offload results from large-output tools
+        agent = Agent(plugins=[
+            ContextOffloader(
+                storage=InMemoryStorage(),
+                should_offload=lambda tool_name, token_count, **kwargs: (
+                    tool_name == "get_document_text"
+                ),
+            )
         ])
         ```
     """
@@ -189,6 +226,7 @@ class ContextOffloader(Plugin):
         preview_tokens: int = _DEFAULT_PREVIEW_TOKENS,
         *,
         include_retrieval_tool: bool = True,
+        should_offload: ShouldOffload | None = None,
         evict_after_cycles: int | None = 20,
     ) -> None:
         """Initialize the ContextOffloader plugin.
@@ -204,6 +242,10 @@ class ContextOffloader(Plugin):
                 chars/4 heuristic. Defaults to ``_DEFAULT_PREVIEW_TOKENS`` (1,000).
             include_retrieval_tool: Whether to register the ``retrieve_offloaded_content``
                 tool so the agent can fetch offloaded content. Defaults to True.
+            should_offload: Callback ``(tool_name, token_count, **kwargs) -> bool`` to decide
+                whether a specific tool result should be offloaded. Called only when the result
+                exceeds ``max_result_tokens``. Return ``True`` to offload, ``False`` to keep
+                in context. Defaults to None (all oversized results offloaded).
             evict_after_cycles: Number of agent loop cycles before an offloaded entry is
                 evicted (unified Storage only). Entries stored more than this many cycles
                 ago are deleted. Defaults to 20. Set to None to disable eviction.
@@ -227,6 +269,7 @@ class ContextOffloader(Plugin):
         self._max_result_tokens = max_result_tokens
         self._preview_tokens = preview_tokens
         self._include_retrieval_tool = include_retrieval_tool
+        self._should_offload = should_offload
         self._evict_after_cycles = evict_after_cycles
         self._stored_cycles: weakref.WeakKeyDictionary[Agent, dict[str, int]] = weakref.WeakKeyDictionary()
         super().__init__()
@@ -415,6 +458,9 @@ class ContextOffloader(Plugin):
         token_count = await event.agent.model.count_tokens([tool_result_message])
 
         if token_count <= self._max_result_tokens:
+            return
+
+        if self._should_offload is not None and not self._should_offload(event.tool_use.get("name"), token_count):
             return
 
         # Build text preview from text+JSON blocks.
--- a/strands-py/src/strands/vended_plugins/context_offloader/plugin.py
+++ b/strands-py/src/strands/vended_plugins/context_offloader/plugin.py
@@ -47,7 +47,7 @@ import json
 import logging
 import weakref
 from collections.abc import Awaitable
-from typing import TYPE_CHECKING, Any, Protocol, Union
+from typing import TYPE_CHECKING, Any, Protocol
 
 from typing_extensions import TypedDict
 
@@ -153,7 +153,7 @@ _CHARS_PER_TOKEN = 4
 class ShouldOffload(Protocol):
     """Callback protocol for deciding whether a tool result should be offloaded."""
 
-    def __call__(self, tool_name: str, token_count: int, **kwargs: Any) -> Union[bool, Awaitable[bool]]:
+    def __call__(self, tool_name: str, token_count: int, **kwargs: Any) -> bool | Awaitable[bool]:
         """Return True to offload, False to keep the result in context. May be sync or async.
 
         Args:
--- a/strands-py/src/strands/vended_plugins/context_offloader/plugin.py
+++ b/strands-py/src/strands/vended_plugins/context_offloader/plugin.py
@@ -46,7 +46,8 @@ import inspect
 import json
 import logging
 import weakref
-from typing import TYPE_CHECKING, Any, Protocol
+from collections.abc import Awaitable
+from typing import TYPE_CHECKING, Any, Protocol, Union
 
 from typing_extensions import TypedDict
 
@@ -152,8 +153,8 @@ _CHARS_PER_TOKEN = 4
 class ShouldOffload(Protocol):
     """Callback protocol for deciding whether a tool result should be offloaded."""
 
-    def __call__(self, tool_name: str, token_count: int, **kwargs: Any) -> bool:
-        """Return True to offload, False to keep the result in context.
+    def __call__(self, tool_name: str, token_count: int, **kwargs: Any) -> Union[bool, Awaitable[bool]]:
+        """Return True to offload, False to keep the result in context. May be sync or async.
 
         Args:
             tool_name: Name of the tool that produced the result.
--- a/strands-py/src/strands/vended_plugins/context_offloader/plugin.py
+++ b/strands-py/src/strands/vended_plugins/context_offloader/plugin.py
@@ -42,6 +42,7 @@ Example:
 
 from __future__ import annotations
 
+import inspect
 import json
 import logging
 import weakref
@@ -460,8 +461,19 @@ class ContextOffloader(Plugin):
         if token_count <= self._max_result_tokens:
             return
 
-        if self._should_offload is not None and not self._should_offload(event.tool_use.get("name"), token_count):
-            return
+        if self._should_offload is not None:
+            try:
+                verdict = self._should_offload(event.tool_use.get("name", ""), token_count)
+                if inspect.isawaitable(verdict):
+                    verdict = await verdict
+                if not verdict:
+                    return
+            except Exception:
+                logger.warning(
+                    "tool_use_id=<%s> | should_offload callback failed, falling back to default offload",
+                    tool_use_id,
+                    exc_info=True,
+                )
 
         # Build text preview from text+JSON blocks.
         # Empty text blocks are intentionally excluded — they add no content value.
--- a/strands-py/src/strands/vended_plugins/context_offloader/__init__.py
+++ b/strands-py/src/strands/vended_plugins/context_offloader/__init__.py
@@ -29,7 +29,7 @@ Example Usage:
     ```
 """
 
-from .plugin import ContextOffloader
+from .plugin import ContextOffloader, ShouldOffload
 from .storage import (
     FileStorage,
     InMemoryStorage,
@@ -42,5 +42,6 @@ __all__ = [
     "FileStorage",
     "InMemoryStorage",
     "S3Storage",
+    "ShouldOffload",
     "Storage",
 ]
--- a/strands-py/tests/strands/vended_plugins/context_offloader/test_plugin.py
+++ b/strands-py/tests/strands/vended_plugins/context_offloader/test_plugin.py
@@ -59,7 +59,9 @@ async def _heuristic_count_tokens(messages, **kwargs):
     return total
 
 
-def _make_event(agent, text_content, status="success", tool_use_id="tool_123", cancel_message=None):
+def _make_event(
+    agent, text_content, status="success", tool_use_id="tool_123", cancel_message=None, tool_name="test_tool"
+):
     """Helper to create an AfterToolCallEvent with content."""
     if isinstance(text_content, str):
         content = [{"text": text_content}]
@@ -71,7 +73,7 @@ def _make_event(agent, text_content, status="success", tool_use_id="tool_123", c
         "status": status,
         "content": content,
     }
-    tool_use = {"toolUseId": tool_use_id, "name": "test_tool", "input": {}}
+    tool_use = {"toolUseId": tool_use_id, "name": tool_name, "input": {}}
 
     return AfterToolCallEvent(
         agent=agent,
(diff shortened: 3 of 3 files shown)


liramon2:
@strandly-the-agent Review this PR. Consolidate your findings in a single comment.

strandly-the-agent:
**Verdict: changes-requested** — reviewed `18147a4b` against `main` (merge-base `7fb9a306`). 6 independent passes (correctness · adversarial · API-shape · issue-alignment · tests · docs), then I re-verified the top findings myself.

**TL;DR**
- 🔴 The `should_offload` call site is unguarded, and the failure mode isn't graceful: a callback that **raises** turns a *successful* tool call into `status: "error"` and throws the real payload away, and an **`async def`** callback's verdict is never awaited so it's silently ignored. Both reproduced end-to-end through a real `Agent`. One ~8-line fix covers both.
- 🟡 The Protocol's `(tool_name, token_count, **kwargs)` contract fails mypy for 4 of 5 plausible callback shapes — including issue #3266's own proposed example *and* this PR's own test callbacks.
- 🟡 Docs site untouched though the checklist says otherwise; and this looks like a Moderate-tier public API change (new exported Protocol behind the default `context_manager="auto"` facade) with no API-review label.
- ❓ Two things worth `lizradway` deciding explicitly rather than by default: veto-only vs. total predicate (identifier parity with #2548's `shouldOffload`), and sequencing the public export against #3045.
- 👍 The feature itself is well-scoped and the 6 new tests each drive a real branch through the actual hook rather than mocking it away. #3266's RAG use case genuinely works — I ran it.

<details>
<summary><b>Verification ledger</b> — what was actually checked</summary>

| Check | Result |
|---|---|
| `hatch test tests/strands/vended_plugins/context_offloader/test_plugin.py` | **65 passed** in 1.6s (incl. all 6 new tests) |
| `ruff check` / `ruff format --check` on the 3 changed files | clean |
| `mypy ./src` | Success, no issues in 218 files |
| Full suite, base vs. PR | 4321 → 4327 passed — additive only, nothing renamed/removed |
| 🔴 raise repro (real `Agent`, real hook, repo's own `MockedModelProvider`) | `tool result status: error`, `payload preserved? False` |
| 🔴 async repro | `RuntimeWarning: coroutine ... was never awaited`, `offloaded anyway: True` |
| Drop-in regression suite (below) | **3 failed, 1 passed** on this branch |
| mypy callback-shape matrix | 6 errors across 4 of 5 plausible shapes |

Convergence (raises confidence, not comment count): the raise-crash was found independently by 3 passes, the async bug by 5, the mypy contract by 3, the docs gap by 2. I re-ran the raise repro, the async repro and the mypy matrix myself rather than trusting the passes.

</details>

<details>
<summary><b>🔴 Block (1)</b></summary>

### `plugin.py:345` — the callback call is unguarded: a raise corrupts the tool result, an async callback is silently ignored

`_handle_tool_result` runs *inside* the executor's tool-call `try:`, and nothing between there and the call site catches anything. Two failure modes, same root cause:

**a) Raise → data loss + misattribution.** A raising callback is caught by the executor's generic handler (`tools/executors/_executor.py:271`), the same path as a genuine tool failure. It logs `tool_name=<...> | failed to process tool`, fabricates `{"status": "error", "content": [{"text": f"Error: {e}"}]}` (`_executor.py:272-277`) and re-fires the hook (`:279`). So a **successful** tool call loses its real output, the model is told the tool failed, and the callback's exception text lands in the transcript:

```
$ hatch run python adv_01_raise.py       # real Agent + real @tool, callback raises
agent turn completed : True
tool result status   : error
tool result text     : 'Error: boom from user callback'
payload preserved?   : False
```

It escalates: if the fabricated error message is itself oversized, the hook is re-entered inside the `except` handler and the second raise is unhandled (`UNHANDLED EventLoopException escaped Agent.__call__`). Same class also reproduced for wrong arity, keyword-only params, a non-callable value, and any return whose `__bool__` raises (e.g. a bare numpy array).

The asymmetry is the tell — the *other* failure source on this exact path already degrades gracefully (`plugin.py:391-397`): with a raising storage backend you get `WARNING ... failed to offload tool result, keeping original` and `status: success`, payload intact. Only the new user-code call isn't wrapped.

**b) `async def` → verdict silently ignored.** `ShouldOffload.__call__` is typed `-> bool` (`plugin.py:86`) and the call site never awaits. An `async def` callback returns a coroutine object, which is always truthy, so `not <coroutine>` is always `False` — the veto **never fires** no matter what the callback decided, plus a `RuntimeWarning`. An async callback returning `False` (intending to *keep* the result) offloads it anyway. mypy catches this for an annotated call site but not for an untyped lambda, and no test covers the path — which is the only reason the checklist's "generates no new warnings" holds.

The sibling exported Protocol `Validator` already does both things right: guarded at `vended_plugins/goal/plugin.py:343-347` (try/except → warn → safe default) and async-aware at `:378-380` (`inspect.isawaitable` + `await`), documented "May be sync or async". Five other user-callback sites in the SDK do the same.

**Fix** (verified against a raise / bad-arity / non-callable / raising-`__bool__` / sync / async matrix — all return `status: success`, async verdicts honored, payload never destroyed):

```python
if self._should_offload is not None:
    try:
        verdict = self._should_offload(event.tool_use.get("name", ""), token_count)
        if inspect.isawaitable(verdict):
            verdict = await verdict
        if not verdict:
            return
    except Exception:
        logger.warning(
            "tool_use_id=<%s> | should_offload callback failed, falling back to default offload",
            tool_use_id,
            exc_info=True,
        )
```

Fail-open (treat a broken callback as "no opinion" → offload, keeping context bounded) vs. fail-closed is your call — but something has to catch. If you go async-aware, widen the Protocol to `-> bool | Awaitable[bool]` and say "may be sync or async" in the docstring, mirroring `Validator`.

</details>

<details>
<summary><b>🔴 Regression tests you can paste in (3 of 4 fail on this branch)</b></summary>

```
$ hatch test -n0 test_should_offload_regressions.py -q
FAILED test_raising_callback_does_not_propagate
FAILED test_bad_signature_callback_does_not_propagate
FAILED test_async_callback_returning_false_skips_offload
3 failed, 1 passed, 1 warning in 0.52s
```
The 4th passes only by accident (truthy coroutine) while still emitting the `RuntimeWarning`.

```python
"""Regression tests for the should_offload call site."""

import warnings
from unittest.mock import AsyncMock, MagicMock

import pytest
from strands.hooks import AfterToolCallEvent
from strands.vended_plugins.context_offloader import ContextOffloader, InMemoryStorage


async def _heuristic_count_tokens(messages, **kwargs):
    total = 0
    for m in messages:
        for b in m["content"]:
            total += len(str(b)) // 4
    return total


def _event(agent, text, tool_name="test_tool", tool_use_id="tool_123"):
    result = {"toolUseId": tool_use_id, "status": "success", "content": [{"text": text}]}
    return AfterToolCallEvent(
        agent=agent,
        selected_tool=MagicMock(),
        tool_use={"toolUseId": tool_use_id, "name": tool_name, "input": {}},
        invocation_state={},
        result=result,
    )


@pytest.fixture
def mock_agent():
    agent = MagicMock()
    agent.model = MagicMock()
    agent.model.count_tokens = AsyncMock(side_effect=_heuristic_count_tokens)
    return agent


def _plugin(cb):
    return ContextOffloader(
        storage=InMemoryStorage(),
        max_result_tokens=25,
        preview_tokens=10,
        include_retrieval_tool=False,
        should_offload=cb,
    )


@pytest.mark.asyncio
async def test_raising_callback_does_not_propagate(mock_agent):
    """A broken callback must not escape the hook (the executor turns it into an error result)."""

    def boom(tool_name, token_count, **kwargs):
        raise RuntimeError("boom")

    event = _event(mock_agent, "x" * 200)
    await _plugin(boom)._handle_tool_result(event)  # must not raise


@pytest.mark.asyncio
async def test_bad_signature_callback_does_not_propagate(mock_agent):
    event = _event(mock_agent, "x" * 200)
    await _plugin(lambda tool_name: True)._handle_tool_result(event)  # must not raise


@pytest.mark.asyncio
async def test_async_callback_returning_false_skips_offload(mock_agent):
    async def never(tool_name, token_count, **kwargs):
        return False

    large = "x" * 200
    event = _event(mock_agent, large)
    with warnings.catch_warnings(record=True) as caught:
        warnings.simplefilter("always")
        await _plugin(never)._handle_tool_result(event)
    assert [w for w in caught if w.category is RuntimeWarning] == []
    assert event.result["content"][0]["text"] == large


@pytest.mark.asyncio
async def test_async_callback_returning_true_offloads(mock_agent):
    async def always(tool_name, token_count, **kwargs):
        return True

    event = _event(mock_agent, "x" * 200)
    await _plugin(always)._handle_tool_result(event)
    assert "[Offloaded:" in event.result["content"][0]["text"]
```
(Adjust to the file's class/fixture conventions as you see fit — the point is the three assertions.)

</details>

<details>
<summary><b>🟡 Should-fix (5)</b></summary>

**2. `plugin.py:86` vs. `:345` — the type contract doesn't match the runtime contract.** The Protocol's params are positional-or-keyword, so mypy makes the *names* `tool_name`/`token_count` binding on implementors — while the call site passes positionally, so the names are runtime-irrelevant. Measured against 5 plausible user shapes, mypy rejects 4:

```
error: ... "def cb(name: str, tokens: int, **kwargs: object) -> bool"; expected "ShouldOffload | None"
error: ... "Callable[[str, int], bool]";  expected "ShouldOffload | None"
error: ... "def <lambda>(name: Any, tokens: Any, **k: Any) -> bool"; expected "ShouldOffload | None"
error: ... "Callable[[Any, Any], bool]";  expected "ShouldOffload | None"
Found 6 errors in 1 file
```
Both casualties are this project's own words: **#3266's proposed usage** (`lambda tool_name, token_count: tool_name in {...}`, no `**kwargs`) errors, and so do **this PR's own test callbacks** (`lambda name, tokens, **kwargs`, `test_plugin.py:899`) — invisible today only because `pyproject.toml:148` scopes mypy to `./src`, never `tests/`. The one shape that type-checks is the docstring example verbatim. Note positional-only `/` is *not* the fix — `/` plus `**kwargs` is stricter still. Minimum: document "your callback must accept `**kwargs`" and declare `tool_name: str | None` (the call site passes `event.tool_use.get("name")`, which mypy only resolves to `str` because `"name"` is a required TypedDict key — the annotation is one refactor from lying). See the API-shape section for the more thorough option.

**3. Docs site wasn't updated, though the checklist says it was.** `site/src/content/docs/user-guide/concepts/plugins/context-offloader.mdx:224` has a `## Configuration` table listing every other ctor kwarg (`:232-234`) — no `should_offload` row, no example; `grep should_offload` on the file returns nothing. Its `## Tradeoffs` section (`:398`) is *literally about* the preview-vs-full-content problem this feature solves and doesn't mention it either. A user who finds the offloader through the docs site won't discover the feature. Also `vended_plugins/context_offloader/__init__.py:6-29`'s "Example Usage" block wasn't updated to mirror the example you added at `plugin.py:27-38` — that's what `help(...)` shows.

**4. Process: Moderate-tier API change with no API-review label.** `ContextOffloader` isn't "an infrequently used method" (`team/API_BAR_RAISING.md`'s Minimal tier) — it's auto-constructed by the built-in `context_manager="auto"/"agentic"` facade at `agent/agent.py:558`. A new exported type customers implement to get new behavior reads as Moderate → `api/needs-review` + an informal API-reviewer pass. Current labels: `enhancement, python, area-context, area-hooks, size/m`. The CI gate only enforces label *pairing* once the label exists, so nothing catches this automatically. `API_BAR_RAISING.md` also asks for the full signature-with-defaults and an end-to-end example in the PR body — those currently live in #3266, not here.

**5. `test_plugin.py:908-921` doesn't prove "no residue."** `test_callback_returning_false_skips_offload` only asserts `event.result` text is unchanged. Mutation-tested: a guard that still leaves `event.result` alone but lets the storage-write loop run anyway passes all 65 tests. One line fixes it, matching the file's own convention (`test_image_without_bytes_not_stored` asserts on `storage._store`):
```python
assert len(storage._store) == 0  # proves nothing was written
```

**6. Two unrelated reformatting hunks.** `test_plugin.py:618` and `:728` collapse existing multi-line calls in `TestRetrievalToolSearch` to one line — no behavior change, not part of this feature, and against your own checklist item about splitting unrelated work. Worth dropping, especially given the rebase noted below.

</details>

<details>
<summary><b>⚪ Nits (bundled) + housekeeping</b></summary>

`tool_name` can be `None` if an earlier hook rewrites `tool_use` (`BeforeToolCallEvent.tool_use` is writable) — `event.tool_use.get("name", "")` at `plugin.py:345` closes it · a hook-set `AfterToolCallEvent.retry` re-invokes the callback twice for one logical tool call, which quietly breaks a stateful "offload at most N" policy; the docstring is silent on it · the retrieval-tool-skip × callback interaction is untested (all 6 new tests use `include_retrieval_tool=False`) · `test_plugin.py:890`'s `assert received_args[0][1] == 50` is a magic number that only lands round because 200 is divisible by 4 — prefer deriving it · non-`bool` returns rely on plain truthiness (`""`/`0`/`[]` mean "don't offload", `"false"` means offload) with no doc callout · `ShouldOffload` isn't `@runtime_checkable`, purely cosmetic since no `isinstance` check exists, but `Validator` has it and `EdgeConditionWithContext` documents *why not* — worth picking a side · `plugin.py:174-176`'s "Called only when the result exceeds `max_result_tokens`" is true but reads as "always called when over threshold"; the `cancel_message` (`:328`) and retrieval-tool (`:331`) skips also bypass it.

**Housekeeping (not defects):** GitHub reports this `mergeable: CONFLICTING` — the base is ~100 commits stale and #3476/#3487 have since changed `__init__`'s signature (`storage: Storage | _LegacyStorage`, new `evict_after_cycles`). I checked with `git merge-tree`: the conflicts are mechanical and confined to `__init__`; the new guard itself applies cleanly. Separately, `team/designs/0009-context-offloader.md` has no mention of this — a standing gap (#3476/#3487 didn't update it either), not yours to fix here.

</details>

<details>
<summary><b>API shape: raw args vs. a context object — genuinely arguable, so here's the evidence rather than a ruling</b></summary>

My passes disagreed on this one, so I'll show the disagreement instead of hiding it. One pass wanted to block on replacing `(tool_name, token_count, **kwargs)` with a single `OffloadContext`; others pointed out the shipped shape follows the *documented* house style.

**For the shape as shipped:** `strands-py/docs/STYLE_GUIDE.md:63-77` gives almost exactly this pattern as the canonical example ("Good: Protocol allows adding optional keyword arguments in the future"), and it matches the one *exported* callback precedent, `Validator` (`vended_plugins/goal/plugin.py:81-94`). By the letter of the rules this isn't an outlier — and your second commit moving `Callable` → `Protocol` was the right call.

**Against it:** every callback Protocol added *since* `EdgeCondition` takes a single typed context object (`RenderContentCallback`, `TriggerCallback`, `InjectionQueryCallback`/`InjectionFormatCallback`). And `EdgeCondition` is the cautionary tale for raw args — it needed one more field later, and the permanent cost is still visible in `multiagent/graph.py`: a `LegacyEdgeCondition | EdgeConditionWithContext` union (`:85-86`), `inspect.signature()` sniffing for a *parameter name* (`:89`), a cache field (`:193`) and a `cast` to recover the legacy type (`:213`), with both types exported forever. The style guide may simply be documenting the older pattern.

**The measured part (independent of taste):** adding a field later is type-clean only for the *strictly compliant* shape. The `(tool_name, token_count)` shape — no `**kwargs`, i.e. what #3266 itself proposes — type-errors today and raises `TypeError: got an unexpected keyword argument` at runtime after a future field is added. A single positional context param absorbs new fields with zero signature change (I verified `recency`/`previous_file_reads` can be added later with no impact on any callback shape).

**Sketch, if you want it:**
```python
@dataclass(frozen=True, kw_only=True)
class OffloadContext:
    tool_name: str | None
    result_tokens: int          # parity with TS `resultTokens` (#2548)
    exceeds_threshold: bool     # lets the predicate be total, see Questions
    tool_use: dict[str, Any]

class ShouldOffload(Protocol):
    def __call__(self, context: OffloadContext, /) -> bool | Awaitable[bool]: ...
```
Also considered, and honestly tempting on the "simplest happy path" tenet: skip the callback entirely for now — `ContextOffloader(storage=..., offload_tools={"get_document_text"})` covers #3266's stated use case in one line with zero new public types, and is introspectable/serializable for config-driven agents. #3266 dismissed it as less flexible, which is true in the limit, but the two are additive.

I wouldn't hold the PR on shape alone — the team blocks rarely and this passes the letter of the guide. But there are zero external users today and the `EdgeCondition` migration tax has already been paid once, so it seems worth you and `lizradway` deciding this deliberately in this PR rather than by default. If raw args stay, please at least land finding #2.

</details>

<details>
<summary><b>Questions</b></summary>

**Blocking-ish (worth an answer before merge):**
- **Veto-only or total predicate?** As written the effective rule is `token_count > max_result_tokens AND should_offload(...)` — the callback can only *veto*, never *force*, and your own `test_callback_not_called_when_under_threshold` pins that. #2548 (also `lizradway`) makes TS's `shouldOffload` *total*, with the threshold as a clause *inside* it, and its default heuristic ("offload the earlier read of the same file") is a force-offload of a possibly-small result — structurally inexpressible here. If Python ships veto-only and TS ships total, the two SDKs share an identifier with opposite semantics, which cuts against the parity rules in `AGENTS.md`. Pass `exceeds_threshold` in and let the predicate be total, or keep veto-only and rename to something honest like `offload_filter`?
- **Fail-open or fail-closed** when the callback misbehaves (finding 🔴)? Your call — I only verified that *something* must catch.

**Non-blocking:**
- `lizradway` on #3266: *"planning on introducing per-tool targeting in #3045 that will eventually subsume the Context Offloader."* Since this exports `ShouldOffload` in `__all__` (`__init__.py:45`) — a customer-facing commitment that's harder to walk back than the internal-API subclass workaround #3266 itself names — is it worth unblocking the RAG case now and sequencing the public export until #3045's interface settles? Not an objection to the feature, purely a sequencing question.
- `count_tokens` runs *before* the guard (`plugin.py:340`), so a pure name-based filter still pays a token count on every oversized result it's about to reject. Free with the local heuristic; a real round-trip under `use_native_token_count=True`. Maybe just a docstring note.

</details>

<details>
<summary>Attacked and found nothing — no action needed</summary>

Guard ordering is right (`cancel_message` `:328` and the retrieval-tool self-skip `:331` both correctly precede the callback). A `False` verdict leaves **zero** residue — no storage entries, no eviction bookkeeping, no plugin state (the guard sits before all writes). Concurrency is fine: 3 parallel tool calls with a stateful callback invoked exactly once per result, no interleaving. Backward compatibility holds — no existing caller passes `should_offload`, and both docstring examples actually run: I built a harness with two big-output tools and the exact documented lambda against the default `max_result_tokens`, and `get_document_text` was offloaded while the equally-large search result was not. #3266's use case works as advertised. `event.tool_use.get("name")` and the positional call both match this file's existing conventions.

</details>

---
<sub>Automated multi-pass review by `strandly-the-agent` — solid work to be reviewed, not a gate. I can be wrong; the 🔴 is the part I'd most want a human to confirm before you spend time on it, and the repro above should make that a 30-second check. Happy to dig into any finding or drop ones you disagree with.</sub>

liramon2:
@maniryu Thanks for the contribution and sorry for the delay. Here are suggestions for this PR:
- The unguarded callback flagged above is worth fixing because it leads to data loss and triggers whenever the callback hits a runtime error. The `async def` case should also be fixed because passing an `async def` callback is a natural use case. The fix for both are relatively small.
- Do update the site docs to include `should_offload` since this look like a public feature.
- Apply the fix to `test_callback_returning_false_skips_offload` flagged above for test completeness.
- Resolve the merge conflicts

liramon2:
@maniryu Reminder to make the changes above.

lizradway:
pushed rebase + updates for fixes

strands-py/src/strands/vended_plugins/context_offloader/plugin.py
```diff
@@ -177,6 +206,16 @@ class ContextOffloader(Plugin):
         agent = Agent(plugins=[
             ContextOffloader(storage=InMemoryStorage())
         ])
+
+        # Only offload results from large-output tools
```
liramon2, line 210: Does this example need to be here? The same example was shown on the file-level docstring.

lizradway: no preference either way tbh. can remove in follow up next time im in this file if we dont think its necessary

liramon2 (approved):

Issue #762, opened 2025-08-28
[FEATURE] Custom Tool Executor


pgrayy:
### Problem Statement

Currently, Strands is packaged with `ConcurrentToolExecutor` and `SequentialToolExecutor`. These executors implement the `ToolExecutor` interface which is marked private ([source](https://github.com/strands-agents/sdk-python/blob/main/src/strands/tools/executors/_executor.py)). I would like this to be made public so that I can define a custom executor.

### Proposed Solution

Make [`ToolExecutor`](https://github.com/strands-agents/sdk-python/blob/main/src/strands/tools/executors/_executor.py) a public interface.

### Use Case

1. Define more complex execution strategies (e.g., a topological ordering with concurrent and sequential execution points).
2. Add custom retry logic.
3. Add custom return-of-control.

### Additional Context

As part of this work, we have the following actions to take:
1. Simplify/reorganize the before and after tool execution logic.
    - Currently, the base `ToolExecutor` class implements `_stream` and `_stream_with_trace` methods to wrap the tool execution with additional logic that includes tracing, hooks, exception handling, etc. It is a bit cumbersome and would maybe make more sense to be placed in the `AgentTool` implementations.
2. Finalize the `execute` method interface.
    - We should be able to remove `tool_results`. `execute` could instead work as a generator that yields results upon completion that the caller (inside `event_loop.py`) collects. For this to work however, we require strongly typed events.
3. Accept custom thread pools for sync tool execution.

har1101:
I'd like to take on the TypeScript side of this, since TS already has a class-based executor hierarchy (#3268) and only needs the base made public.

**Use case.** [strands-lambda-durable](https://github.com/har1101/strands-lambda-durable) runs Strands agents on AWS Lambda durable functions. It opens one durable child context per tool use, in `toolUse` order, before any tool starts. That keeps the journal the same on replay whatever the completion order. It does this by extending `ConcurrentToolExecutor` and wrapping `execute()`. Because `ToolExecutorOptions`/`ToolExecutionInput` are internal, it has to recover the types with `Parameters<ConcurrentToolExecutor['execute']>`, and the contract it relies on (resume via `completedToolResults`, interrupt handling) is undocumented. Other uses for the same extension point are rate-limited or priority scheduling and per-batch tracing.

**Proposal (TS only).**

- Export `ToolExecutor` (value) and `ToolExecutorOptions` / `ToolExecutionInput` (types) from `@strands-agents/sdk`.
[...]

PR #3268, opened 2026-07-15
feat(ts): add ToolExecutor class hierarchy

### LINKED ISSUE #3166: [PORT] Pluggable tool-executor class hierarchy (Py→TS)
> ⚠️ Filed by Strandly as part of the #2498 parity audit. Source feature verified merged + unit-tested + integ-tested + documented in `strands-py`. Agent-produced — confirm scope. (Add the `port` label — I can't apply it via API.)

**Parent:** #2498

## Source SDK
`strands-py`

## Target SDK
`strands-ts`

## Related Files
```
strands-py/src/strands/tools/executors/_executor.py
strands-py/src/strands/tools/executors/concurrent.py
strands-py/src/strands/tools/executors/sequential.py
strands-py/src/strands/tools/executors/__init__.py
strands-py/src/strands/agent/agent.py
strands-py/src/strands/event_loop/event_loop.py
strands-py/tests/strands/tools/executors/test_executor.py
strands-py/tests/strands/tools/executors/test_concurrent.py
strands-py/tests/strands/tools/executors/test_sequential.py
strands-py/tests_integ/tools/executors/test_concurrent.py
strands-py/tests_integ/tools/executors/test_sequential.py
site/src/content/docs/user-guide/concepts/tools/executors.mdx
```

## What needs to be migrated
Both SDKs have the two execution *behaviors*, but TS exposes them as a **fixed enum** (`agent.ts:153`, `inStrategy = 'sequential' | 'concurrent'`); Python has a **subclassable class hierarchy** the user can pass to the agent. Port the pluggable hierarchy:
- **`ToolExecutor(abc.ABC)`** (`_executor.py:32`) — concentrates shared machinery as staticmethods: `_invoke_before_tool_call_hook`/`_invoke_after_tool_call_hook` (`:46,67`), `_stream` (`:94`, per-tool pipeline: lookup, span, retry loop honoring `cancel_tool`, `selected_tool.stream(...)` → `ToolResultEvent`/`ToolStreamEvent`/`ToolInterruptEvent`, after-hook + retry, error handling), `_stream_with_trace` (`:291`); **abstract `_execute(...) -> AsyncGenerator[TypedEvent]`** (`:353`) — the only method subclasses override.
- **`ConcurrentToolExecutor`** (`concurrent.py:19`) — asyncio task-per-tool + queue/event back-pressure.
- **`SequentialToolExecutor`** (`sequential.py:18`) — one at a time, breaks on `ToolInterruptEvent`.
- **Agent wiring:** `Agent(tool_executor=None)` (`agent.py:185`), default `ConcurrentToolExecutor()` (`:464`); event loop calls `agent.tool_executor._execute(...)` (`event_loop.py:788`).

## TypeScript landing points
- New `strands-ts/src/tools/executors/` (`executor.ts` abstract base + `concurrent-executor.ts` + `sequential-executor.ts`).
- Refactor existing inline behavior out of `agent.ts`: `ToolExecutorStrategy` union (`:153`) + `toolExecutor?` field (`:298`) → accept a `ToolExecutor` instance (keep string union as sugar/back-compat); the `switch` in `executeTools` (`:2227-2247`) → `executor.execute(...)`; `_executeToolsSequential` (`:2275`)/`_executeToolsConcurrent` (`:2368`) become the two subclasses' `execute()` bodies; `_executeToolWithMiddleware`/`_executeToolCore` (`:2602,2629`) → the shared base machinery.

## Gotchas / open questions
- **`abc.ABC` + `@abstractmethod` → TS `abstract class` + `abstract execute()`** (the crux — additive; TS is an enum today).
- **Do NOT hand-translate asyncio `Queue`/`Event`/`stop_event` choreography** — reuse the existing idiomatic TS `_executeToolsConcurrent` as the subclass body; the behavioral spec is the tests / observable event order, not the mechanism.
- Pure logic, **browser-safe** — no `.node.ts` split needed.
- `BidiAgent` union in `_is_agent` (`:36`) — if TS has no Bidi agent, drop that branch (scope reduction, flag it).

- Merged: ✅  ·  Unit tests: ✅ (`test_executor.py` 35, `test_concurrent.py`, `test_sequential.py`)  ·  Integ: ✅  ·  Docs: ✅ (`executors.mdx`, whose `sourceLinks` already name `strands-ts/src/agent/agent.ts` as the current TS equivalent — update to show the class API on port).

## Confirmation
- [x] The change is complete in the source SDK (merged + unit + integ tested + documented); every related artifact is listed above.

gautamsirdeshmukh:
## Description

The Strands TypeScript SDK currently keeps sequential and concurrent tool execution inside the `Agent` implementation, while Python models them as [executor classes](https://github.com/strands-agents/harness-sdk/tree/main/strands-py/src/strands/tools/executors). 

This PR ports that hierarchy so the built-in execution policies can be configured as concrete executor instances and the shared tool lifecycle is separated from the agent loop.

### Public API Changes

The existing string shorthand remains supported:

```typescript
const agent = new Agent({ toolExecutor: 'sequential' })
```

The built-in executors can now be supplied and replaced directly:

```typescript
import { 
  Agent,
  ConcurrentToolExecutor,
  SequentialToolExecutor 
} from '@strands-agents/sdk'

const agent = new Agent({
  toolExecutor: new SequentialToolExecutor(),
})

agent.toolExecutor = new ConcurrentToolExecutor()
```

For comparison, here's the Python DevX:

```python
from strands import Agent
from strands.tools.executors import (
  ConcurrentToolExecutor,
  SequentialToolExecutor,
)

agent = Agent(
  tool_executor=SequentialToolExecutor(),
)

agent.tool_executor = ConcurrentToolExecutor()
```

Concurrent execution remains the default. The ToolExecutor base remains internal; custom executors are tracked separately in #762.

## Related Issues

Closes #3166 

## Documentation PR

Updates are in this PR

## Type of Change

New feature / refactor

## Testing

## Checklist

----

### DIFF
--- a/site/src/content/docs/user-guide/concepts/tools/executors.mdx
+++ b/site/src/content/docs/user-guide/concepts/tools/executors.mdx
@@ -6,7 +6,8 @@ tags: [tool-execution, event-loop]
 sourceLinks:
   - path: strands-py/src/strands/tools/executors/concurrent.py
   - path: strands-py/src/strands/tools/executors/sequential.py
-  - path: strands-ts/src/agent/agent.ts
+  - path: strands-ts/src/tools/executors/concurrent.ts
+  - path: strands-ts/src/tools/executors/sequential.ts
 ---
 
 Tool executors control whether tools from a single assistant turn run concurrently or sequentially. Both SDKs default to concurrent execution.
@@ -39,6 +40,8 @@ agent("What is the weather and time in New York?")
 
 --8<-- "user-guide/concepts/tools/executors.ts:concurrent"
 ```
+
+`toolExecutor: 'concurrent'` remains supported as shorthand.
 
 </Tab>
 </Tabs>
@@ -76,6 +79,8 @@ agent("Please take a screenshot and then email the screenshot to my friend")
 
 --8<-- "user-guide/concepts/tools/executors.ts:sequential"
 ```
+
+`toolExecutor: 'sequential'` remains supported as shorthand.
 
 </Tab>
 </Tabs>
--- a/site/src/content/docs/user-guide/concepts/tools/executors.mdx
+++ b/site/src/content/docs/user-guide/concepts/tools/executors.mdx
@@ -41,7 +41,7 @@ agent("What is the weather and time in New York?")
 --8<-- "user-guide/concepts/tools/executors.ts:concurrent"
 ```
 
-`toolExecutor: 'concurrent'` remains supported as shorthand.
+The `'concurrent'` string shorthand keeps your imports minimal. Passing `new ConcurrentToolExecutor()` is equivalent if you prefer to be explicit.
 
 </Tab>
 </Tabs>
@@ -80,7 +80,7 @@ agent("Please take a screenshot and then email the screenshot to my friend")
 --8<-- "user-guide/concepts/tools/executors.ts:sequential"
 ```
 
-`toolExecutor: 'sequential'` remains supported as shorthand.
+The `'sequential'` string shorthand keeps your imports minimal. Passing `new SequentialToolExecutor()` is equivalent if you prefer to be explicit.
 
 </Tab>
 </Tabs>
--- a/site/src/content/docs/user-guide/concepts/tools/executors.ts
+++ b/site/src/content/docs/user-guide/concepts/tools/executors.ts
@@ -1,9 +1,4 @@
-import {
-  Agent,
-  ConcurrentToolExecutor,
-  SequentialToolExecutor,
-  tool,
-} from '@strands-agents/sdk'
+import { Agent, tool } from '@strands-agents/sdk'
 import { z } from 'zod'
 
 const weatherTool = tool({
@@ -44,7 +39,7 @@ const emailTool = tool({
   // --8<-- [start:concurrent]
   const agent = new Agent({
     tools: [weatherTool, timeTool],
-    toolExecutor: new ConcurrentToolExecutor(),
+    toolExecutor: 'concurrent',
   })
   // Omit toolExecutor to use concurrent execution by default.
 
@@ -56,7 +51,7 @@ const emailTool = tool({
   // --8<-- [start:sequential]
   const agent = new Agent({
     tools: [screenshotTool, emailTool],
-    toolExecutor: new SequentialToolExecutor(),
+    toolExecutor: 'sequential',
   })
 
   await agent.invoke('Take a screenshot and email it to my friend')
--- a/site/src/content/docs/user-guide/concepts/tools/executors.ts
+++ b/site/src/content/docs/user-guide/concepts/tools/executors.ts
@@ -1,4 +1,9 @@
-import { Agent, tool } from '@strands-agents/sdk'
+import {
+  Agent,
+  ConcurrentToolExecutor,
+  SequentialToolExecutor,
+  tool,
+} from '@strands-agents/sdk'
 import { z } from 'zod'
 
 const weatherTool = tool({
@@ -39,9 +44,9 @@ const emailTool = tool({
   // --8<-- [start:concurrent]
   const agent = new Agent({
     tools: [weatherTool, timeTool],
-    toolExecutor: 'concurrent',
+    toolExecutor: new ConcurrentToolExecutor(),
   })
-  // or simply: new Agent({ tools: [weatherTool, timeTool] })
+  // Omit toolExecutor to use concurrent execution by default.
 
   await agent.invoke('What is the weather and time in New York?')
   // --8<-- [end:concurrent]
@@ -51,7 +56,7 @@ const emailTool = tool({
   // --8<-- [start:sequential]
   const agent = new Agent({
     tools: [screenshotTool, emailTool],
-    toolExecutor: 'sequential',
+    toolExecutor: new SequentialToolExecutor(),
   })
 
   await agent.invoke('Take a screenshot and email it to my friend')
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -20,13 +20,12 @@ import {
   type SystemPromptData,
   TextBlock,
   ToolResultBlock,
-  type ToolResultBlockData,
   ToolUseBlock,
 } from '../types/messages.js'
 import { deepCopy } from '../types/json.js'
 import type { JSONValue } from '../types/json.js'
 import { McpClient } from '../mcp/index.js'
-import { isValidToolName, type Tool, type ToolContext } from '../tools/tool.js'
+import { isValidToolName, type Tool } from '../tools/tool.js'
 import type { ToolChoice, ToolSpec } from '../tools/types.js'
 import { cloneSystemPrompt, systemPromptFromData } from '../types/messages.js'
 import { normalizeError, ConcurrentInvocationError, StructuredOutputError } from '../errors.js'
@@ -50,7 +49,8 @@ import { ConversationManager } from '../conversation-manager/conversation-manage
 import { ContextOffloader } from '../vended-plugins/context-offloader/plugin.js'
 import { InMemoryStorage } from '../storage/in-memory-storage.js'
 import { HookRegistryImplementation } from '../hooks/registry.js'
-import { MiddlewareRegistry, InvokeModelStage, ExecuteToolStage, AgentStreamStage } from '../middleware/index.js'
+import { createMiddlewareInterrupt } from '../middleware/interrupt.js'
+import { MiddlewareRegistry, InvokeModelStage, AgentStreamStage } from '../middleware/index.js'
 import type {
   MiddlewareStage,
   MiddlewareHandler,
@@ -64,22 +64,17 @@ import type {
 import type {
   InvokeModelContext,
   InvokeModelResult,
-  ExecuteToolContext,
-  ExecuteToolResult,
   AgentStreamContext,
   AgentStreamResult,
-  MiddlewareInterruptResult,
 } from '../middleware/index.js'
 import type { HookableEventConstructor, HookCallback, HookCallbackOptions, HookCleanup } from '../hooks/types.js'
 import {
   InitializedEvent,
   AfterInvocationEvent,
   AfterModelCallEvent,
-  AfterToolCallEvent,
   AfterToolsEvent,
   BeforeInvocationEvent,
   BeforeModelCallEvent,
-  BeforeToolCallEvent,
   BeforeToolsEvent,
   HookableEvent,
   MessageAddedEvent,
@@ -88,12 +83,12 @@ import {
   ModelMessageEvent,
   ToolResultEvent,
   AgentResultEvent,
-  ToolStreamUpdateEvent,
   InterruptEvent,
   type ModelStopData,
-  type ToolUseData,
 } from '../hooks/events.js'
 import { StructuredOutputTool, STRUCTURED_OUTPUT_TOOL_NAME } from '../tools/structured-output-tool.js'
+import { ConcurrentToolExecutor } from '../tools/executors/concurrent.js'
+import { SequentialToolExecutor } from '../tools/executors/sequential.js'
 import { AgentAsTool } from './agent-as-tool.js'
 import type { AgentAsToolOptions } from './agent-as-tool.js'
 import { ToolCaller } from './tool-caller.js'
@@ -111,9 +106,8 @@ import { CancelledError, CheckpointError } from '../errors.js'
 import { DefaultModelRetryStrategy } from '../retry/default-model-retry-strategy.js'
 import type { RetryStrategy } from '../retry/retry-strategy.js'
 import { warnOnDuplicateRetryStrategyTypes } from '../retry/retry-strategy.js'
-import { Interrupt, InterruptError, InterruptState, interruptFromAgent } from '../interrupt.js'
+import { InterruptError, InterruptState } from '../interrupt.js'
 import { Checkpoint, type CheckpointPosition, type CheckpointResumeContent } from '../experimental/checkpoint.js'
-import type { InterruptParams } from '../types/interrupt.js'
 import { isInterruptResponseContent, type InterruptResponseContent } from '../types/interrupt.js'
 import { takeSnapshot as takeSnapshotInternal, loadSnapshot as loadSnapshotInternal } from './snapshot.js'
 import type { TakeSnapshotOptions } from './snapshot.js'
@@ -292,10 +286,13 @@ export type AgentConfig = {
    */
   id?: string
   /**
-   * Strategy for executing tool calls from a single assistant turn.
-   * Defaults to `'concurrent'`. See {@link ToolExecutorStrategy} for details.
+   * Executor for tool calls from a single assistant turn.
+   *
+   * Accepts a {@link ConcurrentToolExecutor}, a {@link SequentialToolExecutor},
+   * or the corresponding {@link ToolExecutorStrategy} string shorthand.
+   * Defaults to concurrent execution.
    */
-  toolExecutor?: ToolExecutorStrategy
+  toolExecutor?: ConcurrentToolExecutor | SequentialToolExecutor | ToolExecutorStrategy
   /**
    * When `true`, the agent loop pauses at cycle boundaries (`afterModel`,
    * `afterTools`) and returns `stopReason: 'checkpoint'` with a populated
@@ -362,6 +359,26 @@ function resolveConversationManager(
   return conversationManager ?? new SlidingWindowConversationManager({ windowSize: 40 })
 }
 
+/**
+ * Resolves a tool executor instance from an executor or string shorthand.
+ */
+function resolveToolExecutor(
+  toolExecutor: ConcurrentToolExecutor | SequentialToolExecutor | ToolExecutorStrategy | undefined
+): ConcurrentToolExecutor | SequentialToolExecutor {
+  switch (toolExecutor) {
+    case 'sequential':
+      return new SequentialToolExecutor()
+    case 'concurrent':
+    case undefined:
+      return new ConcurrentToolExecutor()
+    default:
+      if (typeof toolExecutor === 'string') {
+        throw new Error(`Unknown toolExecutor: ${toolExecutor}`)
+      }
+      return toolExecutor
+  }
+}
+
 /** Default name assigned to agents when none is provided. */
 const DEFAULT_AGENT_NAME = 'Strands Agent'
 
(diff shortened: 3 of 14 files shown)


gautamsirdeshmukh:
@strandly-the-agent please review

strandly-the-agent (commented):
**Reviewed `c8afd52` — a solid, behavior-preserving port of the Python `ToolExecutor` hierarchy, tests green (3784 pass). Two things to settle before this leaves draft: one API-shape decision (#1) and the API-review label. No correctness blockers.**

Ran the full pipeline (correctness · API/DevX · adversarial+repro · test-quality · LLM-context · docs). Ledger + findings below.

<details>
<summary>✅ What I verified</summary>

- **Branch/SHA:** `c8afd52` (`strands-ts`)
- **Tests:** 138 files / 3784 tests pass, 0 type errors on the branch
- **Behavior parity (moved, not rewritten):** empty-toolUseBlocks, pre-launch cancel, sequential + concurrent interrupt/pending-state, `AfterToolsEvent` bracketing, and `executeTool`/middleware/core all confirmed equivalent to the old inline paths
- **Ruled sound (adversarial):** no unhandled promise rejections (0 measured), no executor-reassignment race (read once per turn), per-tool event ordering holds, the `generator.return` cast doesn't corrupt history, and `AfterToolsEvent`-then-throw is tolerated by `_stream` (no history corruption / double-processing)
- **Docs:** snippets compile, `sourceLinks` valid, no phantom "subclass it" claim · **LLM-context:** model-facing error strings are verbatim moves — no regression
</details>

### 🟡 1. The public `toolExecutor` type points at an `@internal`, un-exported base — is that the contract you want?

`AgentConfig.toolExecutor` (`agent.ts:296`) and the new public field (`agent.ts:481`) are typed as `ToolExecutor`, but that base is `@internal` and not in the barrel (`executor.ts:62`; `index.ts:148-149` exports only the two concretes). Its `private` members (`executor.ts:161,196`) make the type nominal, so a user can't import it, structurally match it, or subclass it without a deep `@internal` import. Every sibling strategy field exports its base (`model: Model`, `conversationManager: ConversationManager`).

Net: over the existing `'concurrent'`/`'sequential'` shorthand, this PR only adds `new ConcurrentToolExecutor()` — behavior-identical to the strings, since both concretes are zero-arg. Custom executors are deferred to #762.

Would it be cleaner to type the field + config as a **closed union** until #762 lands?
```ts
toolExecutor?: ConcurrentToolExecutor | SequentialToolExecutor | ToolExecutorStrategy
```
That keeps the `@internal` type out of the public signature and stays forward-compatible (widen to include the base once #762 exports it). Opening the base now would prematurely freeze the internal-dep option interfaces (`ToolExecutorOptions` / `ToolExecutionInput`).

**Process:** this changes public surface three ways (visibility private→public, widened config union, two new exports) but there's no `needs-api-review` / `completed-api-review` label. It should carry `needs-api-review`, and given the open extension-contract question I'd take it to an API discussion rather than solo-approve.

### 🟡 2. `concurrent.ts:83` silently reverses a previously-documented behavior

Old concurrent execution converted an error thrown *outside* the tool body (a middleware or `BeforeToolCallEvent` hook throwing) into a per-tool error `ToolResultBlock` and continued the siblings, with a rationale comment. The new code does `throw winner.error` — aborting the whole batch and discarding completed sibling results — and the rationale comment was dropped.

The change is defensible (it makes concurrent consistent with sequential, and matches Python for the before-hook/middleware case) and it's tested ("propagates unexpected middleware failures"). But it's an observable behavior change with nothing explaining the flip. Could you add a one-line rationale at `:83` and note it in the PR description, and confirm it's intended? (Nuance: Python still *catches* `AfterToolCallEvent`-hook errors and converts them; the new TS aborts on those too — but TS-sequential already diverged there, so TS is now internally consistent.)

<details>
<summary>⚪ Nits & one non-blocking note</summary>

- **`concurrent.ts:120-127`** — the `'Tool execution interrupted'` fallback now also labels hard-error aborts; it's discarded on the throw path but visible to `AfterToolsEvent` hooks. Low impact.
- **`agent.ts:374-376`** — the invalid-string branch of `resolveToolExecutor` is a new *runtime* throw (the old switch was compile-time-exhaustive) and is untested. Optional one-liner.
- **`executors.ts:47`** — docs now lead with `new ConcurrentToolExecutor()` over the string shorthand, which buys nothing today; consider keeping the shorthand front until custom executors ship.
- **Pre-existing (not introduced here, so not a blocker):** a concurrent batch can deadlock if one tool errors/interrupts while a sibling is parked on a non-cancellable `await` — the `finally` at `concurrent.ts:110-114` awaits `gen.return()` behind that await (reproduced, 4 cases). It exists identically on `origin/main`; the new abort-on-throw path at `:83` just reaches that `finally` a bit more often. Worth a tracking issue (fix direction: fire the abort first, then bound the `gen.return()` wait) rather than addressing it in this port.
</details>

---
*Automated review by Strandly (I'm an AI agent, built with Strands). This is work for a human to approve, not a gate — especially the design call in #1. Happy to be wrong; push back freely.*

strands-ts/src/tools/executors/sequential.ts
```diff
@@ -0,0 +1,69 @@
...
+    const { toolUseBlocks, toolResultBlocks, invocationState, assistantMessage, completedToolResults } = input
+
+    for (const toolUseBlock of toolUseBlocks) {
+      const completedResult = completedToolResults?.get(toolUseBlock.toolUseId)
+      if (completedResult) {
+        toolResultBlocks.push(completedResult)
+        continue
+      }
+
+      if (options.agent.cancelSignal.aborted) {
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Added back doc that had been removed during refactor (line 23)

```
Executes tool calls in source order, honoring `agent.cancelSignal` between
calls to short-circuit tools that have not started.
```

strands-ts/src/tools/executors/executor.ts
```diff
@@ -0,0 +1,271 @@
...
+/**
+ * Input for executing the tool calls from one model turn.
+ *
+ * @internal
+ */
+export interface ToolExecutionInput {
+  /** Tool calls to execute. */
+  readonly toolUseBlocks: readonly ToolUseBlock[]
+  /** Destination for completed tool results. */
+  readonly toolResultBlocks: ToolResultBlock[]
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Clarified the doc to reflect pre-existing behavior - sequential execution may leave partial results for resume, while concurrent fills missing results during cleanup. Since the base remains internal, fallback behavior will stay executor-specific until #762

strands-ts/src/tools/executors/concurrent.ts
```diff
@@ -0,0 +1,132 @@
...
+
+        if (winner.kind === 'throw') {
+          pendingSteps.delete(index)
+
+          if (winner.error instanceof InterruptError) {
+            interruptError = winner.error
+            continue
+          }
+
+          throw winner.error
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Added back old behavior, also replaced test

gautamsirdeshmukh:
@strandly-the-agent take another look

strandly-the-agent (approved):
**APPROVE — re-reviewed `696302f`; both prior code concerns are resolved and I found no new code issues.**

✅ Public executor signatures now use the closed built-in union; the `@internal` base no longer leaks into the API.
✅ Concurrent middleware/generator failures are again isolated per tool, with sibling completion covered by the replacement test.
✅ Added the invalid-executor test and clarified accumulator/cancellation docs.
✅ Verified locally: targeted **41/41**, full unit-node **3,785/3,785**, plus build, lint, type-check, and `git diff --check`.

⚠️ Separate merge-process gate remains: add `api/needs-review`, then have the designated human API reviewer apply `api/review-complete`. I don't think this needs an API meeting now that the type shape is settled.

<details><summary>Delta scope</summary>

Compared the rebased revisions with `git range-diff c8afd52^! 696302f^!` so unrelated changes from the newer base were excluded. Correctness, API/DevX, and test-quality passes were rerun; the two suggested extra integration assertions were suppressed as duplicate/low-value because the underlying contracts are already independently tested.

</details>

*Automated review by Strandly (an AI agent built with Strands); human review still owns the API/process decision.*

gautamsirdeshmukh:
@strandly-the-agent have another look

strands-ts/src/tools/executors/executor.ts
```diff
@@ -0,0 +1,283 @@
...
+  // Keys are the model-issued toolUseIds — resume matches against the model's
+  // toolUse blocks, and hooks may rewrite the id on the result itself.
+  protected _storePendingToolExecution(
+    options: ToolExecutorOptions,
+    assistantMessage: Message,
+    completedToolResults: ReadonlyMap<string, ToolResultBlock>
+  ): void {
+    const serializedResults: Record<string, { toolResult: ToolResultBlockData }> = {}
+    for (const [toolUseId, result] of completedToolResults) {
+      serializedResults[toolUseId] = result.toJSON()
```
strandly-the-agent, line 178: 🔴 **Preserve the model-issued ID inside the serialized result too.**

Keying this map by the original `toolUseId` correctly prevents completed tools from rerunning, but `result.toJSON()` still persists the hook-rewritten ID. The resulting state is effectively:

```json
{ "tool-a": { "toolResult": { "toolUseId": "rewritten-a" } } }
```

The assistant history still contains model-issued call `tool-a`, so Bedrock and Anthropic emit an uncorrelated call/result pair. I reproduced this both uninterrupted and after interrupt → JSON snapshot → fresh-agent resume. The mismatch existed before this delta; this change makes it durable and the new test treats it as expected.

Please keep the provider-issued ID authoritative: normalize every terminal `ToolResultBlock`—including hook-replaced `AfterToolCallEvent.result` values—to `toolUseBlock.toolUseId`, or reject/ignore ID rewrites. The interrupt test should expect both the map key and inner result ID to be `tool-a` and assert provider-facing call/result IDs match. Updating the assistant-side ID is unsafe for stateful providers.

strandly-the-agent (changes requested):
**Changes requested:** the rerun fix is correct, but the patch now serializes and tests a provider-invalid call/result ID split.

✅ Reviewed exact head `b4072db` via `git range-diff 696302f^! b4072db^!`; local tree clean.
✅ Focused: **97/97**; full unit: **138 files, 3,799/3,799**; build, type-check, lint, and `git diff --check` passed.
✅ CI Gate and TS build/test/docs/integration matrices passed.
🔴 Bedrock and Anthropic repros serialize assistant call `tool-a` with result `rewritten-a`, both uninterrupted and after snapshot/resume.
🏷️ `api/needs-review` is present; `api/review-complete` should remain pending until this contract is corrected.

<details><summary>Evidence, attribution, and minimal fix</summary>

The provider-correlation defect was already present at the previously approved head; this delta did not invent it. It correctly changes pending-result lookup to use the original model ID, fixing duplicate execution: the interrupt repro now executes `A, B, B`, rather than rerunning A.

However, the new serialization keeps the original ID only as the outer map key while retaining the rewritten ID inside the result. The new test explicitly asserts that split, and the changed hook documentation advertises replacement of `toolUse`, including its ID. This turns an existing edge-case defect into intentional durable behavior.

Repro path:
1. The model emits tool call `tool-a`.
2. `BeforeToolCallEvent` replaces its ID with `rewritten-a`.
3. The assistant message retains `tool-a`, while the terminal result carries `rewritten-a`.
4. Bedrock emits `toolUseId: rewritten-a`; Anthropic emits `tool_use_id: rewritten-a`, neither matching the assistant call.
5. The same mismatch survives interrupt-state JSON serialization and fresh-agent resume.

The bookkeeping matrix isolated this to ID rewrites: input/name rewrites, retry, cancellation, and multiple completed tools remained valid.

Minimal fix: treat the provider-issued ID as the immutable correlation key. Hooks may still rewrite input/name, but every externally emitted or persisted terminal result should be normalized back to the original ID, including results replaced by after-tool hooks. Update the new regression test to require outer key and inner result ID `tool-a`, and add a capturing-model assertion over the resumed provider-facing history.

The `check-api-review-label` failure is expected while completion remains pending; `label-size` appears policy-related and is not part of this code finding.
</details>

*Automated review by Strandly (an AI agent built with Strands); a human owns the final API decision.*

strands-ts/src/tools/executors/executor.ts
gautamsirdeshmukh: @strandly-the-agent addressed, have a look

strands-ts/src/agent/agent.ts
```diff
@@ -908,6 +894,23 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+  /**
+   * Executor for tool calls from a single assistant turn.
+   *
+   * Reading always yields the resolved executor instance. Assigning accepts an
+   * executor instance or a {@link ToolExecutorStrategy} string shorthand, which is
+   * resolved to the matching instance on write.
+   *
+   * @throws Error if assigned an unrecognized string shorthand.
+   */
+  get toolExecutor(): ConcurrentToolExecutor | SequentialToolExecutor {
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Intentional. `ToolExecutor` will be made publicly accessible as part of separate PR for #762.

gautamsirdeshmukh:
@strandly-the-agent Hi! Please take another review and resolve your last thread if it's now been addressed sufficiently.

strandly-the-agent (approved):
**APPROVE — the provider-correlation blocker is fixed at `e1519c5`.**

✅ Actual Bedrock and Anthropic request boundaries: **6/6** provider×flow scenarios passed; canonical call/result IDs survived both hook replacements plus sequential/concurrent JSON snapshot → fresh-agent resume.
✅ Terminal-path matrix: **14/14** across cancellation, missing tool, middleware result/failure, success/error, and retry; no ID leakage or reruns.
✅ Exact-head tests: targeted **380/380**; full unit **138 files, 3,799/3,799**, with no type errors.
✅ Build, full type-check, lint, format check, `git diff --check`, CI Gate, and applicable TypeScript matrices passed.
✅ The prior thread is resolved; the implementation matches the requested model-issued-ID contract.
🏷️ This approval does **not** complete the designated human API review or satisfy `api/review-complete`.

<details><summary>Nonblocking compatibility follow-up</summary>

Snapshots produced by older released behavior while hooks rewrote `toolUseId` may already contain irrecoverably mismatched IDs. General migration is not possible where the original ID was discarded. This predates and is not worsened by the current fix; it is worth tracking separately and considering in release communication.

</details>

*Automated review by Strandly (an AI agent built with Strands); human review still owns the API/process decision.*

strands-ts/src/tools/executors/sequential.ts
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Executes tool calls one at a time.
+ *
+ * @example
+ * ```typescript
+ * import { Agent, SequentialToolExecutor } from '@strands-agents/sdk'
+ *
+ * const agent = new Agent({
+ *   toolExecutor: new SequentialToolExecutor(),
```
Unshure: nit: Maybe include the string shorthand to use sequential here as well?

strands-ts/src/tools/executors/concurrent.ts
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * Executes tool calls concurrently.
+ *
+ * @example
+ * ```typescript
+ * import { Agent, ConcurrentToolExecutor } from '@strands-agents/sdk'
+ *
+ * const agent = new Agent({
+ *   toolExecutor: new ConcurrentToolExecutor(),
```
Unshure: nit: Maybe include the string shorthand to use concurrent here as well?

strands-ts/src/hooks/events.ts
```diff
@@ -101,11 +101,12 @@ export abstract class HookableEvent extends StreamEvent {
...
  * Matches the shape of the tool use block the model emitted; hooks on
  * {@link BeforeToolCallEvent} may mutate its fields (or reassign the object)
- * to rewrite the input, id, or tool name before the tool executes.
+ * to rewrite the input or tool name before the tool executes. The model-issued
+ * tool-use ID remains the authoritative provider correlation key.
  */
 export interface ToolUseData {
   name: string
-  toolUseId: string
+  readonly toolUseId: string
```
Unshure: nit: Why did we make this update? I guess this means that `ToolUseData.toolUseId` is never meant to change?

site/src/content/docs/user-guide/concepts/tools/executors.ts
```diff
@@ -39,9 +44,9 @@ const emailTool = tool({
   // --8<-- [start:concurrent]
   const agent = new Agent({
     tools: [weatherTool, timeTool],
-    toolExecutor: 'concurrent',
```
Unshure, line 42: Both are supported right? Do we want customers to use the string approach or explicit class approach? I kinda like string better to reduce customer imports.

site/src/content/docs/user-guide/concepts/tools/executors.mdx
```diff
@@ -40,6 +41,8 @@ agent("What is the weather and time in New York?")
 --8<-- "user-guide/concepts/tools/executors.ts:concurrent"
 ```
 
+`toolExecutor: 'concurrent'` remains supported as shorthand.
```
Unshure: This reads like the `'concurrent'` string based approach is the backup approach, not the default. I think the string approach is simpler and maybe should be considered the default? What do you think?

Unshure (dismissed):

strands-ts/src/tools/executors/sequential.ts
gautamsirdeshmukh: Agreed, updating the example to lead with shorthand + include the direct construction example after

strands-ts/src/tools/executors/concurrent.ts
gautamsirdeshmukh: Same as above

strands-ts/src/hooks/events.ts
gautamsirdeshmukh: Redundant (overly defensive), `_normalizeToolResultId` already rebuilds every terminal result with the original id, removing

site/src/content/docs/user-guide/concepts/tools/executors.ts
gautamsirdeshmukh: Yep, both are supported (in typescript). I'm in favor of this, updating

site/src/content/docs/user-guide/concepts/tools/executors.mdx
gautamsirdeshmukh: I do agree (reworded), and I also think I want to follow up with a change to strands-py to enable the shorthand pattern as well :)

Unshure (approved):

Issue #1726, opened 2026-02-19
[FEATURE] Bidirectional Streaming - Telemetry and Metrics


mehtarac:
### Problem Statement

Implement Otel and metrics for bidi feature

### Use Case

NA

PR #3282, opened 2026-07-15
feat(bidi): add telemetry and metrics

mehtarac:
## Description

The bidirectional streaming implementation lacked observability (tracing session lifecycles, connection latency, token usage, etc.). This PR adds OpenTelemetry tracing to `_BidiAgentLoop` that captures the lifecycle of a bidi session. Telemetry is kept self-contained in `experimental/bidi/_telemetry.py`, which wraps the shared `Tracer` primitives rather than adding bidi-specific methods to the shared class.

#### Traced:
- Session lifecycle (`start()` / `stop()`) — root `bidi_session` span with agent name, model, tools, system prompt, and duration
- `model.start()` — `bidi_connect` span measuring connection establishment latency
- `BidiResponseStartEvent` / `BidiResponseCompleteEvent` — `bidi_response` span with `response_id`, `finish_reason`, and time-to-first-audio (`gen_ai.server.time_to_first_audio`, ms from response start to first audio chunk)
- `BidiUsageEvent` — accumulated `input_tokens`, `output_tokens`, `total_tokens`, and `cache_read_input_tokens` on the session span
- `BidiInterruptionEvent` — `bidi_interruption` span event on the session span with reason (barge-in signal)
- `BidiConnectionRestartEvent` — `bidi_connection_restart` span with the timeout error message
- `ToolUseStreamEvent` / `ToolResultEvent` — tool call span with tool name, tool_use_id, and status

#### Not traced (as spans):
- `BidiAudioStreamEvent` — per-frame audio, thousands per second; consumed only to compute time-to-first-audio on the response span, not emitted as its own span
- `BidiTranscriptStreamEvent` — partial deltas, too granular
- `BidiConnectionStartEvent` — redundant with the `bidi_connect` span
- `BidiConnectionCloseEvent` — redundant with session span end
- `BidiErrorEvent` — customer-facing IO error type, not emitted by the loop

#### Notes / scope decisions:
- `cache_write_input_tokens` is intentionally not accumulated: no bidi provider currently emits it. Only `cache_read_input_tokens` (emitted by OpenAI Realtime and Gemini Live) is captured.
- Time-to-first-audio uses a distinct `gen_ai.server.time_to_first_audio` attribute rather than the shared `gen_ai.server.time_to_first_token`, because the latter is defined as request→first-content and would be mislabeled here.
- Gemini Live emits no `BidiResponseStartEvent`, so it produces no response spans (pre-existing provider behavior).

#### Related fix — restart error propagation:
While adding restart tracing, this PR also corrects error handling in `_restart_connection`:
- The before-restart hook now runs outside the try/except so its exceptions propagate to `receive()` per the hook contract (and cannot leak a restart span).
- A failed restart (`model.stop()` / `model.start()` raising) now re-raises to the `receive()` consumer instead of silently idling, and the send gate is only reopened on a successful restart.

## Related Issues

#1726

## Type of Change

New feature

## Testing

- [x] Added unit tests in `tests/strands/experimental/bidi/agent/test_loop_telemetry.py` covering session/connect/response/tool/restart spans, interruption events, usage (incl. cache-read) accumulation, time-to-first-audio (present and absent), tool-span error status, and the two restart error paths (before-restart hook raises; restart fails)
- [x] Verified end-to-end span tree and attributes against a fake in-memory model through the real `BidiAgent` loop

## Checklist

---


mehtarac:

strands-py/src/strands/experimental/bidi/agent/loop.py
```diff
@@ -214,11 +261,23 @@ async def _run_model(self) -> None:
...
+
         try:
             async for event in self._agent.model.receive():
                 await self._event_queue.put(event)
 
-                if isinstance(event, BidiTranscriptStreamEvent):
+                if isinstance(event, BidiResponseStartEvent):
+                    response_span = _telemetry.start_response_span(
+                        self._tracer, event.response_id, parent_span=self._session_span
+                    )
```
strandly-the-agent, line 301: 🔴 **Response-span leak.** `response_span` is overwritten on every `BidiResponseStartEvent` without ending the one already open — only a matching `BidiResponseCompleteEvent` (or the `finally` at :307, which rescues just the *last* one) closes it. Nova Sonic emits one `BidiResponseStartEvent` per `contentStart` (`nova_sonic.py:672`) but a single `BidiResponseCompleteEvent` per `completionEnd` (`nova_sonic.py:599`), so a normal multi-block completion (transcript + audio + text) opens 3 spans and closes 1. Reproduced: **STARTED=3, ENDED=1, LEAKED=2**. Over a long-lived bidi session this leaks unboundedly (span-processor memory growth + lost telemetry).

Minimal fix — close-before-open:
```suggestion
                if isinstance(event, BidiResponseStartEvent):
                    if response_span:
                        _telemetry.end_response_span(self._tracer, response_span, stop_reason="interrupted")
                    response_span = _telemetry.start_response_span(
                        self._tracer, event.response_id, parent_span=self._session_span
                    )
```
Open design question: since every `contentStart` in one completion shares the same completion id, do you want one span *per content block* (above) or one span *per `response_id`* (start only when `response_span is None`)? Both plug the leak but produce different trace shapes — worth deciding explicitly.

strands-py/src/strands/experimental/bidi/agent/_telemetry.py
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    if model_id:
+        attributes["gen_ai.request.model"] = model_id
+
+    if tools:
+        attributes["gen_ai.agent.tools"] = serialize(tools)
+
+    if system_prompt:
+        attributes["gen_ai.system_instructions"] = system_prompt
```
strandly-the-agent: 🔴 **System prompt written unredacted.** Everywhere else in the codebase `gen_ai.system_instructions` is emitted only through `self._redact("gen_ai.system_instructions", …)` (`tracer.py:1262,1270`), so this raw assignment silently bypasses the redaction/allowlist policy — system prompts frequently carry PII, business rules, or embedded secrets, and they'd now be exported to traces regardless of that policy.

Route it through the redaction seam (`tracer._redact(...)`) like the rest of the tracer does. Note `_redact` is itself private — which reinforces the broader point that these helpers really want to live on `Tracer` (or `Tracer` needs a public seam) rather than reaching into its internals.

strandly-the-agent (commented):
**Review — `feat(bidi): add telemetry and metrics`**  (reviewed `62e4f1be`, base `main`)

**TL;DR:** Solid, well-tested first cut — the span model is sensible and the happy-path lifecycle is clean (session span is closed on both the success and the connection-failure paths, no None-span crash paths, no double-ends). Two things I'd fix before this leaves draft: a **response-span leak** on Nova multi-block completions (reproduced), and the **system prompt written to the session span unredacted**, which bypasses the tracer's redaction policy. The rest is span-conformance + test-hardening. I've left inline comments on the three substantive items. This is an automated review — a human should confirm before acting.

✅ **Verified:** `pytest test_loop_telemetry.py` → **12 passed** · leak reproduced (STARTED=3, ENDED=1, **LEAKED=2**) · interruption leak reproduced (**LEAKED=1**) · timeout/restart path, null-parent-span, and session-span-without-`stop()` (safe — guaranteed by `BidiAgent.__aenter__/__aexit__`) all checked and clear.

<details><summary>🔴 Blockers (2)</summary>

1. **Response spans leak on multi-content-block completions** — `loop.py:270-273`. Nova Sonic emits a `BidiResponseStartEvent` per `contentStart` (`nova_sonic.py:672`) but a single `BidiResponseCompleteEvent` per `completionEnd` (`nova_sonic.py:599`). Each start overwrites `response_span` without ending the previous one; the `finally` only rescues the last. Bidi sessions are long-lived → unbounded orphaned spans + span-processor memory growth + lost telemetry. See inline. *(confirmed by control-flow analysis **and** a runnable repro)*
2. **System prompt written unredacted to the session span** — `_telemetry.py:47`. Every other emission of `gen_ai.system_instructions` routes through `self._redact(...)` (`tracer.py:1262,1270`); this path sets it raw, silently bypassing the redaction/allowlist policy. See inline.

</details>

<details><summary>🟡 Should-fix (4)</summary>

3. **Bidi spans bypass `_get_common_attributes`** — `_telemetry.py:35,89,127,159`. All six `start_*` helpers build attributes from scratch, so bidi spans lack the `gen_ai.system="strands-agents"` / `gen_ai.provider.name` marker every other span carries (`tracer.py:1217-1225`) and ignore the `use_latest_genai_conventions` legacy/latest switch. Fix: seed each from `tracer._get_common_attributes(operation_name=…)`.
4. **Interruption never closes the open response span** — `loop.py:289-299`. Repro: `Start(a), Interruption, Start(b), Complete(b)` → **LEAKED=1**. See inline.
5. **Tests assert setup, not the behavior they name** (these stay green if the code under test is deleted): `test_session_span_records_error_on_start_failure` (`:122-129`) never checks span ERROR status; `test_response_span_closed_on_model_error` (`:287-302`) only asserts `pytest.raises`, never that the `finally` closed the span; `test_usage_accumulation` (`:264-283`) checks in-memory counters, never that tokens land on the session span (`_telemetry.py:70-73`). And no test feeds an unbalanced Start/Complete, so finding #1 passes CI invisibly.
6. **Reaching into `Tracer` privates** — `_telemetry.py` is the only code in the repo calling `_start_span`/`_end_span`/`_redact`/`_get_common_attributes` (8 sites), while this same PR uses the **public** `start_tool_call_span`/`end_tool_call_span` for tool spans (`loop.py:334,361,390`). Findings #2 and #3 are downstream symptoms. Worth deciding: add public bidi span methods (or a public seam) to `Tracer`? Also `finish_reason` is used as a **span** attribute (`_telemetry.py:110`) vs. an **event** attribute elsewhere (`tracer.py:454,469`) — align or get explicit sign-off.

</details>

<details><summary>⚪ Nits (bundled)</summary>

- **`total_tokens` (and cache tokens) dropped** — `loop.py:302-303` accumulates only input/output though `BidiUsageEvent` carries `total_tokens` (`events.py:474`) and `tracer.py` records `gen_ai.usage.total_tokens`.
- **Reimplemented exporter** — `_InMemoryExporter` duplicates OTel's `InMemorySpanExporter`, already used at `tests/strands/event_loop/test_event_loop.py:10,795` (TESTING.md says reuse); imports also sit mid-file below a class def → move to top.
- **`test_no_crash_without_otel_configured` (`:305-320`)** never exercises the no-op path (the autouse `otel_setup` swaps in a real provider) and has no assertions.
- **Weak count-only assertions** — `test_response_span_lifecycle` (`:147-165`) and the restart test (`:215-234`) assert only counts; add span + `gen_ai.error.message` + a failed-restart ERROR + a tool-raises ERROR path (`loop.py:390`).
- **Hard-coded `>=3` event count** (`:203`) risks a 90s hang if the count drops — prefer breaking on `ToolResultMessageEvent`.
- **Gemini Live emits no `bidi_response` events** (`gemini_live.py`) → that span layer is dead for Gemini; worth a known-limitation note.
- **Bidi tests only run under `hatch run bidi-test:test`** (default `hatch test` `--ignore`s `tests/strands/experimental/bidi`) — easy to miss in CI review.
- No `tru_`/`exp_` assertion-pair naming or whole-object assertions (repo convention).

</details>

strands-py/src/strands/experimental/bidi/agent/loop.py
```diff
@@ -228,6 +287,9 @@ async def _run_model(self) -> None:
                     self._task_pool.create(self._run_tool(tool_use))
 
                 elif isinstance(event, BidiInterruptionEvent):
+                    if self._session_span:
+                        _telemetry.add_interruption_event(self._session_span, event["reason"])
```
strandly-the-agent, line 330: 🟡 **Interruption doesn't close the open response span.** This adds an event to the *session* span but leaves `response_span` open. If an interruption is followed by a new `BidiResponseStartEvent` with no intervening `Complete` (a barge-in), the pre-interruption span leaks via the same overwrite as the blocker above. Reproduced: `Start(a), Interruption, Start(b), Complete(b)` → **LEAKED=1**.

End the open span here with `stop_reason="interrupted"` before recording the event (or rely on the close-before-open fix above — but closing it *at* the interruption is more semantically accurate for the `finish_reason`).

mehtarac:
@strandly-the-agent review the implementation again

strandly-the-agent (commented):
**Re-review** — `5dbbc297` (vs. my earlier review at `62e4f1be`)

**TL;DR:** Both blockers are fixed and I verified them; the should-fixes are addressed too. This looks good for a draft — only minor/optional items remain. One heads-up on `finish_reason` semantics below.

✅ **Verified:** `pytest test_loop_telemetry.py` → **12 passed** · re-ran the leak repro on the new code (`Start(r1), Start(r2), Complete(r2)`) → **2 response spans finish** (`r1`=`interrupted`, `r2`=`complete`), no leak · mypy clean · system prompt now routed through `tracer._redact`.

<details><summary>Resolution ledger</summary>

| Prior finding | Status |
|---|---|
| 🔴 Response-span leak on multi-block completions | ✅ Fixed — close-before-open in `_run_model` (ends prior span `interrupted` before opening new); repro confirms no leak |
| 🔴 System prompt written unredacted | ✅ Fixed — `_telemetry.py:40` now `tracer._redact("gen_ai.system_instructions", …)` |
| 🟡 Spans bypass `_get_common_attributes` | ✅ Fixed — all `start_*` helpers seed from it; provider marker + convention switch restored |
| 🟡 Interruption doesn't close response span | ✅ Resolved via the close-before-open path |
| 🟡 Tests assert setup, not behavior | ✅ Fixed — now assert span ERROR status + token span-writes (`input`/`output`/`total`) |
| 🟡 Reaching into `Tracer` privates | 🟢 Your call — kept, with a documented rationale in the module docstring. Fine by me; no re-nag |
| ⚪ `total_tokens` dropped / hard-coded `>=3` count / normal stream-end labeled `error` | ✅ All fixed (`total_tokens` recorded; breaks on `ToolResultMessageEvent`; stream-end now `incomplete`) |

</details>

<details><summary>Remaining — all minor / optional (fine to leave for a draft)</summary>

- **`finish_reason="interrupted"` semantics (heads-up).** The close-before-open labels a preempted response span `interrupted`. That's correct for a real barge-in — but Nova emits a `BidiResponseStartEvent` per `contentStart`, so a *normal* multi-block completion (text + audio + tool in one completion) will also produce an `interrupted` span followed by a `complete` one. Anyone counting `interrupted` responses in traces would over-count. Not blocking, and it's the natural consequence of the per-content-block span choice — just flagging so it's a conscious call. (Distinguishing the two would need a completion-boundary signal.)
- `_InMemoryExporter` is still hand-rolled rather than OTel's `InMemorySpanExporter` (nit).
- `test_no_crash_without_otel_configured` still runs under the autouse `otel_setup` fixture (a real provider), so it doesn't actually exercise the no-op tracer path its name promises.
- The restart `gen_ai.error.message` attribute and the failed-restart ERROR path are still untested.

</details>

Nice, fast turnaround on the feedback. Automated review — a human should confirm before merge.

strands-py/src/strands/experimental/bidi/agent/loop.py
```diff
@@ -214,11 +261,23 @@ async def _run_model(self) -> None:
...
 
+        response_span: Span | None = None
+
         try:
             async for event in self._agent.model.receive():
                 await self._event_queue.put(event)
 
-                if isinstance(event, BidiTranscriptStreamEvent):
+                if isinstance(event, BidiResponseStartEvent):
+                    response_span = _telemetry.start_response_span(
```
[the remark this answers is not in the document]
mehtarac: Addressed

strands-py/src/strands/experimental/bidi/agent/loop.py
```diff
@@ -186,6 +227,10 @@ async def _restart_connection(self, timeout_error: BidiModelTimeoutError) -> Non
 
         self._send_gate.clear()
 
+        restart_span = _telemetry.start_restart_span(
```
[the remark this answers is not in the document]
mehtarac: Addressed

strands-py/src/strands/experimental/bidi/agent/loop.py
```diff
@@ -88,12 +100,32 @@ async def start(self, invocation_state: dict[str, Any] | None = None) -> None:
         logger.debug("agent loop starting")
         await self._agent.hooks.invoke_callbacks_async(BidiBeforeInvocationEvent(agent=self._agent))
 
-        await self._agent.model.start(
+        self._session_span = _telemetry.start_session_span(
+            self._tracer,
+            agent_name=self._agent.name,
+            model_id=getattr(self._agent.model, "model_id", None),
```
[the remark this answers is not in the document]
mehtarac: Addressed

strands-py/src/strands/experimental/bidi/agent/loop.py
```diff
@@ -236,8 +298,16 @@ async def _run_model(self) -> None:
...
 
+                elif isinstance(event, BidiUsageEvent):
+                    self._accumulated_input_tokens += event.input_tokens
+                    self._accumulated_output_tokens += event.output_tokens
+
         except Exception as error:
             await self._event_queue.put(error)
+        finally:
+            if response_span:
+                _telemetry.end_response_span(self._tracer, response_span, stop_reason="error")
```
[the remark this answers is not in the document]
mehtarac: Addressed

strands-py/src/strands/experimental/bidi/agent/_telemetry.py
```diff
@@ -0,0 +1,188 @@
+"""Telemetry helpers for bidirectional streaming agent.
+
+Wraps the shared Tracer primitives with bidi-specific span creation and attribute names.
+"""
+
+import logging
+
+from opentelemetry.trace import Span
+
+from ....telemetry.tracer import Tracer, serialize
```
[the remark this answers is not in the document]
mehtarac: Addressed

strands-py/src/strands/experimental/bidi/agent/_telemetry.py
```diff
@@ -0,0 +1,188 @@
...
+from ....types.traces import AttributeValue
+
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
mehtarac: Addressed

strands-py/tests/strands/experimental/bidi/agent/test_loop_telemetry.py
```diff
@@ -0,0 +1,320 @@
...
+    async for event in loop.receive():
+        if isinstance(event, BidiResponseCompleteEvent):
+            break
+
+    await loop.stop()
+
+    spans = otel_setup.get_finished_spans()
+    response_spans = [s for s in spans if "bidi_response" in s.name]
+    assert len(response_spans) >= 1
+    assert response_spans[0].attributes["gen_ai.response.finish_reason"] == "interrupted"
```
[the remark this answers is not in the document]
mehtarac: Addressed

strands-py/tests/strands/experimental/bidi/agent/test_loop_telemetry.py
```diff
@@ -0,0 +1,320 @@
...
+        return SpanExportResult.SUCCESS
+
+    def shutdown(self):
+        pass
+
+    def get_finished_spans(self):
+        return list(self.spans)
+
+
+from strands import tool
```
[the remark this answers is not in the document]
mehtarac: Addressed

strands-py/src/strands/experimental/bidi/agent/loop.py
mehtarac: Addressed

strands-py/src/strands/experimental/bidi/agent/_telemetry.py
mehtarac: Addressed

strands-py/tests/strands/experimental/bidi/agent/test_loop_telemetry.py
```diff
@@ -0,0 +1,320 @@
...
+
+import unittest.mock
+
+import pytest
+import pytest_asyncio
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import SimpleSpanProcessor, SpanExporter, SpanExportResult
+
+
+class _InMemoryExporter(SpanExporter):
```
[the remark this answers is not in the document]
mehtarac: Not applicable — `opentelemetry.sdk.trace.export.in_memory_span_exporter.InMemorySpanExporter` is not available in the bidi-test hatch environment (different OTel SDK version than the main test env). Custom exporter is minimal and intentional.

strands-py/src/strands/experimental/bidi/agent/loop.py
mehtarac: Addressed — the interruption scenario is handled by the fix for the response span leak: when a new `BidiResponseStartEvent` arrives, any open response span is closed with `stop_reason="interrupted"` before the new one opens. The sequence `Start(a) → Interruption → Start(b) → Complete(b)` results in span(a) closed as "interrupted" at Start(b), span(b) closed normally at Complete(b). No leak.

strandly-the-agent:
Checked — head is still `5dbbc297`, unchanged since my [re-review above](https://github.com/strands-agents/harness-sdk/pull/3282#pullrequestreview-4706663886) (no new commits), so there's nothing new to review and that verdict stands: both blockers resolved and verified, should-fixes addressed, only optional nits remain.

One state change worth noting — the PR is now **ready for review** (no longer a draft). Nothing I flagged blocks merge; the remaining items (hand-rolled span exporter, the `finish_reason="interrupted"` over-count heads-up, and a couple of untested restart paths) are optional polish for a human reviewer to weigh. Push new commits and re-tag me and I'll review just the delta.

strands-py/src/strands/experimental/bidi/_telemetry.py
pgrayy, line 1: I would recommend placing this under the bidi/ root instead of bidi/agent.

strands-py/src/strands/experimental/bidi/agent/loop.py
```diff
@@ -88,12 +101,35 @@ async def start(self, invocation_state: dict[str, Any] | None = None) -> None:
...
+                system_prompt=self._agent.system_prompt,
+                tools=self._agent.tool_registry.get_all_tool_specs(),
+                messages=self._agent.messages,
+            )
+        except Exception as error:
+            _telemetry.end_connection_span(self._tracer, connection_span, error=error)
+            _telemetry.end_session_span(self._tracer, self._session_span, error=error)
+            self._session_span = None
+            raise
+        _telemetry.end_connection_span(self._tracer, connection_span)
```
pgrayy, line 131: What is the difference between connection span and session span. Also, it looks like session span isn't closed.

strands-py/src/strands/experimental/bidi/_telemetry.py
```diff
@@ -0,0 +1,182 @@
...
+        span: The session span to end.
+        input_tokens: Accumulated input tokens for the session.
+        output_tokens: Accumulated output tokens for the session.
+        total_tokens: Accumulated total tokens for the session.
+        error: Exception if the session ended with an error.
+    """
+    attributes: dict[str, AttributeValue] = {}
+
+    if input_tokens > 0:
+        attributes["gen_ai.usage.input_tokens"] = input_tokens
```
pgrayy: Does otel distinguish between per turn tokens and per invocation tokens? Also, does otel have any bidi specific fields?

strands-py/src/strands/experimental/bidi/_telemetry.py
mehtarac: so this only instruments the agent loop today but good suggestion. happy to move it for future-proofing when we add model/io telemetry.

strands-py/src/strands/experimental/bidi/agent/loop.py
mehtarac: session span is the root span for the entire idi session (Agent.start() -> Agent.stop()). connection span is the child span measuring only the model.start() call latency. 

session span is closed on lines 159-167, unless i am misinterpreting your comment?

strands-py/src/strands/experimental/bidi/_telemetry.py
mehtarac: Otel does not have any bidi specific fields today

mehtarac:
@strands-agent review this pr. all the comments are addressed.

mkmeral:
@strandly-the-agent  review please. do a fresh review 

also are there any API changes? if so give me a breakdown of new APIs, params, configs, failure modes and so on

strandly-the-agent:
Fresh pass at `b2697c45`. The previously-fixed items (response-span leak, redaction, common attrs, `total_tokens`, error/incomplete labeling) all check out correctly on re-read. Two new correctness issues found and reproduced against the actual code, plus a few minor items.

**🟡 `_run_tool` leaks its `execute_tool` span on cancellation** (`loop.py:345-401`) — no `finally`, so `stop()` cancelling an in-flight tool task drops the span entirely (never exported). Confirmed with a repro: task pool has 2 tasks running, `stop()` called, tool span never appears in the exporter.

**🟡 Restart: before-hook exceptions are now silently swallowed** (`loop.py:241`) — moving the hook call inside the `try` means a raising `BidiBeforeConnectionRestartEvent` callback no longer propagates (violates `HookRegistry.invoke_callbacks_async`'s own documented contract), *and* `send_gate.set()` still runs unconditionally at `loop.py:260` even though `model.stop()`/`model.start()`/`_run_model` recreation were all skipped — leaving the loop silently wedged. Confirmed with a repro.

Details, repros, and the rest (restart span asymmetry, token accounting, lint/mypy, file-move) below.

<details>
<summary><b>1. 🟡 <code>_run_tool</code> tool-call span leaks on cancellation</b> — <code>loop.py:345-401</code></summary>

`tool_call_span` is opened at `loop.py:345` and the try block only has `except Exception` (`loop.py:400`), no `finally`. `asyncio.CancelledError` is a `BaseException`, not `Exception`, so when `_TaskPool.cancel()` (`_task_pool.py:37-38`, called from `stop()` via `stop_tasks()`) cancels an in-flight `_run_tool` task, execution unwinds past the `except` clause and `end_tool_call_span` is never called. The span is simply never `.end()`'d, so `SimpleSpanProcessor`/`BatchSpanProcessor.on_end()` never fires and the span data is silently dropped — not a crash, just a telemetry gap for exactly the common "tool still running when the session ends" case.

Repro (ran against the checkout, not committed):
```python
# slow_tool sleeps 10s; start loop, let it spawn the tool task, then stop()
await loop.start()
# ... spawn ToolUseStreamEvent -> _run_tool task created, tool_call_span opened, sleeping
await loop.stop()
# exporter.spans -> ['bidi_connect', 'bidi_session ...']  <- no execute_tool span at all
```
Confirmed via `len(loop._task_pool) == 2` right before `stop()` (model task + tool task), and zero `execute_tool`-named spans exported afterward.

Contrast: `_run_model`'s `response_span` handling (`loop.py:317-321`) *does* use a `finally`, and I verified that one correctly exports the span even when the task is cancelled mid-response. `_run_tool` needs the same treatment.

Fix: wrap the tool-call-span-closing logic in a `finally`, mirroring `_run_model`:
```python
try:
    ...
    self._tracer.end_tool_call_span(tool_call_span, tool_result_event.tool_result)
    ...
except Exception as error:
    self._tracer.end_tool_call_span(tool_call_span, tool_result=None, error=error)
    await self._event_queue.put(error)
    raise
```
(or track whether the span was already ended and close it unconditionally in a `finally`, being careful not to double-end it — `_end_span` is idempotent-safe via `is_recording()` check at `tracer.py:280` so a `finally` that unconditionally calls `end_tool_call_span` again after the success path already ended it is safe, just needs an `error=CancelledError-ish` value on the cancel path specifically, or simply re-raise after ending with no error).

</details>

<details>
<summary><b>2. 🟡 Restart: before-hook failures are now swallowed and the loop can wedge silently</b> — <code>loop.py:225-260</code></summary>

Before this PR, `BidiBeforeConnectionRestartEvent` was invoked *outside* the try, so a raising callback propagated straight out of `_restart_connection`/`receive()` — loud and immediate, and it left `send_gate` cleared (blocking `send()`), consistent with `HookRegistry.invoke_callbacks_async`'s documented contract: *"Any exceptions raised by callback functions will propagate to the caller"* (`registry.py:306`).

Now the hook call is inside the `try` (`loop.py:241`), so a raising before-hook is caught by `except Exception as exception: restart_exception = exception` (`loop.py:251-252`) and:
- `model.stop()`/`model.start()`/`self._task_pool.create(self._run_model())` are (correctly) skipped, same as before — no change there.
- But `self._send_gate.set()` at `loop.py:260` is **unconditional**, outside the try/finally, and still runs. `send()` calls unblock and proceed even though the model was never actually restarted and no `_run_model` task exists to consume `model.receive()` anymore.
- No exception propagates anywhere except via the `BidiAfterConnectionRestartEvent(restart_exception)` hook and the (best-effort) restart span — if the caller doesn't inspect those, this is a silent failure. `receive()`'s `await self._event_queue.get()` will then hang forever since nothing feeds the queue.

Repro (ran against the checkout):
```python
agent.hooks.add_callback(BidiBeforeConnectionRestartEvent, lambda e: (_ for _ in ()).throw(RuntimeError("boom")))
await loop.start()
try:
    await loop._restart_connection(BidiModelTimeoutError("x"))
except Exception as e:
    print("propagated:", e)
else:
    print("swallowed")
print("send_gate after:", loop._send_gate.is_set())
print("tasks after:", len(loop._task_pool))
```
Output: `swallowed`, `send_gate after: True`, `tasks after: 1` (the *original* `_run_model` task, never recreated).

Note the underlying "`send_gate.set()` runs even when the restart failed" pattern is pre-existing for `model.stop()`/`model.start()` failures too (same in the merge-base) — not relitigating that. What's new here is that before-hook failures now *also* fall into that fail-open path instead of hard-failing loudly, which is a meaningful semantics change for hook authors relying on the documented propagation contract.

Suggested fix: after the try/except/finally, only `self._send_gate.set()` (and treat the restart as usable) if `restart_exception is None`; otherwise propagate/surface it (e.g. `raise restart_exception` or push it onto `self._event_queue` like `_run_model` does with `model_error`) so `receive()` consumers learn the session is dead instead of hanging.

</details>

<details>
<summary><b>3. ⚪ Restart path has no <code>bidi_connect</code> span (observation, not a bug)</b> — <code>loop.py:114-128</code> vs <code>loop.py:243-249</code></summary>

`start()` wraps `model.start()` in a `bidi_connect` span (parented to the session span) to capture connection-establishment latency/errors. `_restart_connection` calls `model.stop()`+`model.start()` directly inside the `bidi_connection_restart` span with no nested connect span. Given restart is explicitly a reconnect, this looks like an instrumentation gap rather than an intentional choice — you lose a consistent "connect duration" metric for the reconnect case, and a `model.start()` failure during restart is only visible as the coarser restart-span error rather than a distinguishable connect-phase error. Recommend nesting a `start_connection_span`/`end_connection_span` (parented to `restart_span`) around the `model.stop()`/`model.start()` pair in `_restart_connection`, mirroring `start()`.

</details>

<details>
<summary><b>Confirmed correct — no action needed</b></summary>

- **Tool-span vs. session-span race (ask #2):** `stop()` sequences `stop_tasks()` (cancels + joins the *entire* task pool, including any in-flight `_run_tool`) strictly before nulling `self._session_span` in the `finally` (`loop.py:150-167`). Since `_run_tool` only reads `self._session_span` at task start (`loop.py:345`) and that read can't happen after the task is already joined, a tool span is never parented to an already-ended session span through the normal `stop()` path. I also confirmed via `tracer.py:230-236` that even if it were, OTel wouldn't crash — `_start_span` checks `parent_span.is_recording()` and falls back to the ambient/root context rather than parenting to a dead span.
- **Concurrent tools:** each `_run_tool` invocation uses a purely local `tool_call_span` variable, so concurrent tool calls don't share/clobber span state.
- **Token accumulation (ask #5):** counters reset only in `start()` after a successful connect (`loop.py:129-131`), and are deliberately *not* reset in `_restart_connection` — correct, since the session span (and its usage totals) is session-scoped and spans restarts. Accumulation is `+=` per event, so no double-counting risk from the restart itself. `total_tokens` vs `input+output` consistency is a provider-contract guarantee (`BidiUsageEvent` docstring: "total_tokens: Sum of input and output tokens", `events.py:474`), not something this code needs to (or does) enforce itself.
- **File move (ask #7):** `from .. import _telemetry` in `loop.py:40` correctly resolves to the new `bidi/_telemetry.py` location; grepped the whole repo for the old `bidi/agent/_telemetry.py` path/import and found nothing stale. The test file staying at `tests/.../bidi/agent/test_loop_telemetry.py` isn't actually a mismatch — it tests `loop.py`'s telemetry *integration*, and `loop.py` itself didn't move. Minor nit: there's no direct unit test module for `_telemetry.py`'s span-building helpers in isolation, only indirect coverage via `test_loop_telemetry.py` — fine for now, not blocking.

</details>

<details>
<summary>⚪ Lint/type-check results (ran <code>mypy --python-version 3.12</code> and <code>ruff check --target-version py312</code> on both files)</summary>

- `mypy`: clean, no issues on either file.
- `ruff`: `_telemetry.py` clean. `loop.py` has 2 findings:
  - `loop.py:9` `UP035` (`AsyncGenerator` should come from `collections.abc`) — **pre-existing**, verified identical on the merge-base (`f383c11a`); not this PR's fault.
  - `loop.py:6` `I001` (import block un-sorted) — **PR-introduced**: the new `from .. import _telemetry` at `loop.py:40` breaks the existing import-sort ordering. Trivial fix: `ruff check --fix` or move it up alphabetically with the other `from ..` imports.

</details>

strandly-the-agent:
Fresh review at `b2697c45`, plus the API breakdown you asked for, `mkmeral`. (My [earlier comment](https://github.com/strands-agents/harness-sdk/pull/3282#issuecomment-5122272612) went out mid-pipeline and was only a partial pass — this supersedes it; items already flagged there aren't repeated.)

**API changes: none public.** `_telemetry.py` is `_`-private and never re-exported (no `__init__.py`/`__all__` entry) → unreachable via any supported import path; `_BidiAgentLoop` is itself private; no new `BidiAgent` config; no new deps (`opentelemetry-api`/`-sdk` are already base deps). The diff is exactly `_telemetry.py` [new] + `loop.py` instrumentation + one test file. **What you *are* committing to is the observability contract** — span names, attribute keys, enum values — broken down below.

**TL;DR:** one real functional bug (not telemetry-only), plus contract gaps worth settling before dashboards depend on this.

🔴 `stop()` racing an in-flight restart can resurrect a "stopped" connection + orphan a root span (`loop.py:225` vs `:142-169`).
🟡 Telemetry calls have no defensive isolation → 3 failure modes (blocks startup / masquerades as a model error / silently hangs a tool call).
🟡 `bidi_response {response_id}` puts a unique ID in the span *name* → unbounded cardinality (one-line fix).
🟡 No per-turn token usage on `bidi_response` spans — directly answers `pgrayy`'s open question.
🟡 `finish_reason`/error attrs diverge from OTel semconv (singular vs plural key, invented `gen_ai.error.message`).
🟡 Gemini Live emits **zero** response spans; Nova Sonic emits them at mismatched cardinality.

✅ **Verified:** telemetry suite **12/12 green** · `_start_span`/`_end_span` asymmetry traced · restart-race mechanism confirmed (`_TaskPool` never contains `_restart_connection`) · provider gaps confirmed by grep · `test_loop.py`'s 2 failures reproduce **identically on merge-base `f383c11a`** → pre-existing, **not** this PR.

<details><summary><b>Part 1 — full API / observability-contract breakdown</b></summary>

| Span | Name | Key attributes | Notes |
|---|---|---|---|
| session | `bidi_session {agent_name}` | `gen_ai.agent.name`, `gen_ai.request.model`, `gen_ai.agent.tools`, `gen_ai.system_instructions` (redacted), usage totals on end | root span |
| connect | `bidi_connect` | `gen_ai.request.model` | **only** in `start()` (`loop.py:114`) — restart path has no connect span |
| response | `bidi_response {response_id}` | `gen_ai.response.id`, `gen_ai.response.finish_reason` | ⚠️ id in span *name* — finding 3 |
| restart | `bidi_connection_restart` | `gen_ai.error.message` | one per timeout-triggered restart |
| tool call | (reuses `Tracer.start_tool_call_span`) | standard `execute_tool` attrs | not bidi-specific |

Plus a span **event** `bidi_interruption` (attr `interruption.reason`) recorded on the *session* span, not the response span.

**Signatures:** 9 new module-level functions in `_telemetry.py` (`start`/`end` × session, response, restart, connection + `add_interruption_event`). No existing signature changed. New private attrs on `_BidiAgentLoop`: `_tracer`, `_session_span`, `_accumulated_{input,output,total}_tokens`.

**Config:** no new surface. Honors existing OTel env config, the `use_latest_genai_conventions` switch (for the provider marker), and the `_redact` allowlist for `gen_ai.system_instructions`. Always-on — no way to disable bidi telemetry independently of the global tracer.

**Enum values consumers will build against:** `gen_ai.operation.name` ∈ {`bidi_session`, `bidi_response`, `bidi_connect`, `bidi_connection_restart`}; `finish_reason` ∈ {`complete`, `error`, `interrupted`, `tool_use`} plus loop-injected `incomplete`.

**New failure modes** (internal robustness gaps, no new public exceptions): a telemetry exception can block `model.start()` entirely or be mislabeled as a model error; a telemetry exception in `_run_tool` can silently hang a tool call; the `stop()`/restart race can resurrect a stopped connection.

**Backward compat:** additive, with one observable behavior change — the `BidiBeforeConnectionRestartEvent` hook moved inside a `try`, so a raising before-hook no longer propagates (`HookRegistry` documents that it should). Already flagged in my earlier comment.

</details>

<details><summary><b>Part 2 — new findings, ranked</b></summary>

**🔴 1. `stop()` racing an in-flight restart resurrects the connection.** `_restart_connection` (`loop.py:225`) runs on whatever task drives `receive()` — never inside `self._task_pool` — so `stop()`'s `_task_pool.cancel()` (`loop.py:151`) can't reach it. Traced end to end: `stop()` nulls `_session_span`/`_started` and returns while the restart keeps going, calls `model.start()` again, and at `loop.py:250` adds a *new* `_run_model` task to the pool `stop()` already drained. A live connection + live task survive an explicit `stop()`, unreachable by any future `stop()`, and the new span lands on a different `trace_id` with `parent=None` (orphaned root trace). Repro showed `model.start` await_count=2 *after* `stop()`.
Fix: run `_restart_connection` inside `_task_pool`, or re-check `self._started` immediately before `model.start()`/task creation and abort if `stop()` ran concurrently.

**🟡 2. Telemetry has no defensive isolation → 3 failure modes, one root cause.** `Tracer._start_span` (`tracer.py:204`) has no try/except, unlike `_end_span` (`tracer.py:296-303`):
- a raise while building attributes for `start_session_span` propagates out of `loop.start()` → `model.start()` never runs (`loop.py:106-116` sits *before* the `try:` at `:117`).
- the same raise during `start_response_span` is caught by `_run_model`'s blanket `except Exception` (`loop.py:314`) and surfaces to the `receive()` caller indistinguishable from a genuine model error.
- `start_tool_call_span` (`loop.py:345`) runs *before* the tool's own `try:` (`:347`); a raise escapes into a bare `create_task` whose `.exception()` is never retrieved (`_task_pool.py:27`) → the tool call silently never completes.
Narrow trigger (needs a raise inside attribute-building, not normal input), so this is defense-in-depth. Same fix for all three: wrap the `_telemetry.start_*` calls, or make `_start_span` defensive like `_end_span`.

**🟡 3. `bidi_response {response_id}` puts a unique ID in the span *name*** (`_telemetry.py:92`) → unbounded cardinality, which breaks name-based aggregation in most backends (Datadog/X-Ray/Jaeger/Tempo bucket by span name). Every other name in this SDK is bounded (`chat`, `execute_tool {tool_name}`, `invoke_agent {agent_name}`, `bidi_session {agent_name}`) because the underlying set is small; `response_id` is not. The id is already on `gen_ai.response.id` (`_telemetry.py:90`) — suggest naming the span plain `"bidi_response"`.

**🟡 4. No per-turn token attribution — answering `pgrayy`'s question directly.** `gen_ai.usage.*` is written only once, in `end_session_span` (`_telemetry.py:60-68`), from totals accumulated across the whole session (`loop.py:309-312`, flushed at `stop()`, `:159-166`). `bidi_response` spans carry zero usage attributes, so per-turn cost/token analysis isn't possible today. And the second half: **no**, OTel has no bidi-specific semantic conventions yet. To support per-turn, reset a per-response counter on `BidiResponseStartEvent` and write it in `end_response_span`.

**🟡 5. `finish_reason`/error attrs diverge from OTel semconv.** `gen_ai.response.finish_reason` (`_telemetry.py:109`) is singular; the official key is plural `finish_reasons` (`string[]`) — and elsewhere in this SDK `finish_reason` only appears as a *choice event* attribute (`tracer.py:454,469`), never a span attribute. `gen_ai.error.message` (`_telemetry.py:128`) isn't a semconv key either (convention is `error.type`).
❓ Worth aligning with semconv now, or is bidi's vocabulary intentionally custom (no `length`/`content_filter` concept in streaming) and better documented as a deliberate departure, the way the module docstring already does for the private-`Tracer` usage?

**🟡 6. Provider coverage is uneven for the headline feature.** `gemini_live.py` never emits `BidiResponseStartEvent`/`BidiResponseCompleteEvent` — the per-response span layer silently doesn't exist for it (session usage still works via `BidiUsageEvent`, `gemini_live.py:366`). `nova_sonic.py` emits one `ResponseStart` per `contentStart` (`:660-668`) but one `ResponseComplete` per `completionEnd` (`:589-598`), so a multi-content-block turn mints several response spans and the close-before-open logic (`loop.py:277-278`) force-labels all but the last `"interrupted"` with no real interruption — which would inflate any "interruption rate" metric. `openai_realtime.py` has full coverage.
❓ Is response-span parity across providers expected before merge, or acceptable as a tracked gap while bidi is experimental?

**⚪ Nits (bundled):** `gen_ai.system_instructions` is set as a plain span *attribute* (`_telemetry.py:43`) where the rest of the SDK emits it as a span *event* (`tracer.py:1229-1272`), bypassing `_span_attributes_only`/Langfuse handling — redaction itself is correct · no way to disable bidi telemetry independently of the global tracer (`get_tracer()` unconditional, `loop.py:79`) · minor per-event overhead (attribute-dict build + `datetime.now().isoformat()` on every Start/Complete/Usage event, even against a no-op tracer).

**Merge readiness (not a PR defect):** `test_loop.py`'s 2 failures (unexpected `tracking_id`) are **pre-existing** — I reproduced them identically on merge-base `f383c11a`. Looks like the branch merged a stale `main`; a rebase onto current `main` should clear CI. Telemetry suite is 12/12 green.

</details>

Automated review — worth a human confirming the 🔴 before merge, since the fix touches restart/stop control flow rather than telemetry.

mehtarac:
@strandly-the-agent is the functional bug highlighted part of this new implementation? if not don't block on it, we can track that in another issue.

do a fresh review, are there any blocking items that blck approval?

strandly-the-agent:
Answering both directly, `mehtarac`.

## 1. No — the functional bug is not from this PR. Proven pre-existing, so I'm de-blocking it.

I ran the resurrection repro against the merge-base `f383c11a`, where `loop.py` has **zero** telemetry references (`grep -c _telemetry` → `0`, vs `14` at head):

```
after loop.start(): model.start await_count = 1
restart in flight; calling stop() concurrently...
stop() returned. _started=False
AFTER stop(): model.start await_count = 2      ← reconnected AFTER stop()
AFTER stop(): _send_gate set = True
RESURRECTED = True
```

**Byte-identical output at head `b2697c45`.** Every mechanism pre-dates this PR: on merge-base, `_restart_connection` already called `model.stop()` / `model.start()` / `_task_pool.create(self._run_model())`, was already invoked *directly* from `receive()` (merge-base `loop.py:166`) rather than inside `_task_pool` — so `stop()`'s `_task_pool.cancel()` never could reach it — and already ended with an unconditional `send_gate.set()`. Your instinct was right; this belongs in its own issue and I'm happy to file it with the repro attached.

The only PR-specific piece is the *symptom*, not the cause: the resurrected `_run_model` now mints an orphaned root-trace span. That's a consequence of the pre-existing race — fixing the race fixes the span.

## 2. Blocking items: two, both cheap, neither in the telemetry logic

🚧 **`loop.py:6` — `I001` import-sort. PR-introduced, fails lint CI.** Verified: head reports `I001` + `UP035`; merge-base reports only `UP035` (so `UP035` is pre-existing, not yours). `_telemetry.py` is clean. Fix is `ruff check --fix`.

🚧 **`loop.py:241` — before-restart hook exceptions no longer propagate. PR-introduced; the only behavior change in this PR.** Merge-base invoked `BidiBeforeConnectionRestartEvent` *outside* the `try`; it's now inside, so a raising callback is captured into `restart_exception` instead of propagating — and `HookRegistry.invoke_callbacks_async` documents that exceptions propagate to the caller. Combined with the pre-existing unconditional `send_gate.set()` (`:260`), `send()` reopens against a connection that was never restarted. ~2-line fix (only swallow `model.stop()`/`model.start()` failures, or gate `send_gate.set()` on `restart_exception is None`). Genuinely your call whether it blocks — it needs a raising before-restart hook to trigger.

⚠️ **Practical CI gate, not your bug:** `test_loop.py`'s 2 failures (`tracking_id`) reproduce **identically on merge-base** → stale-`main` merge. CI stays red until a rebase onto current `main`, regardless of this PR. Telemetry suite itself: **12/12 green**.

<details><summary><b>Full ledger — every open finding classified (PR-introduced vs pre-existing)</b></summary>

| Finding | Origin | Blocks? |
|---|---|---|
| `stop()` resurrects a mid-restart connection (`loop.py:225` vs `:142-169`) | **pre-existing** (proven on merge-base, 0 telemetry) | no — separate issue |
| Orphaned root span from the resurrected `_run_model` | PR (symptom of the above) | no — resolves with the race |
| Unconditional `send_gate.set()` on a failed restart (`:260`) | **pre-existing** | no |
| `test_loop.py` ×2 (`tracking_id`) | **pre-existing** (stale main) | CI gate → rebase |
| `UP035` on `loop.py:9` | **pre-existing** | no |
| `I001` on `loop.py:6` | **PR** | **yes** (lint CI) |
| Before-restart hook exceptions swallowed (`:241`) | **PR** | **recommend yes** |
| `execute_tool` span leaks on `CancelledError` (`:345`/`:400`, no `finally`) | **PR** (span is new) | no — telemetry-only gap, no functional impact |
| `bidi_response {response_id}` unique id in span *name* (`_telemetry.py:92`) | **PR** | no — but cheapest to fix now, before dashboards depend on the name |
| `finish_reason` singular vs semconv plural; invented `gen_ai.error.message` | **PR** | no — contract decision, easier pre-adoption |
| No per-turn token attribution (usage only on session span) | **PR** limitation, not a regression | no — answers `pgrayy`'s question; needs a per-response counter |
| No defensive isolation around `_start_span` calls | **PR** | no — narrow trigger (defense-in-depth) |
| Gemini Live emits no response spans; Nova Sonic cardinality mismatch labels turns `interrupted` | provider gaps, pre-existing emission behavior | no — worth tracking |
| `gen_ai.system_instructions` as span attr vs event | **PR** nit | no |

**Verification for this pass:** repro run on both `f383c11a` and `b2697c45` (identical); `ruff --select I001,UP035` compared head vs merge-base; `test_loop_telemetry.py` 12/12; `test_loop.py` 2 failures on both revisions; `_telemetry.py` lint-clean.

</details>

**Verdict:** nothing in the telemetry implementation is a correctness blocker. Fix the `I001` lint, decide on the hook-propagation regression, rebase for green CI — then this is good to approve from my side. The 🔴 I raised earlier is not yours; I'll file it separately if you want.

mehtarac:
@strandly-the-agent do a fresh review. assess if a concern raised is pr blocking or not

strandly-the-agent:
Fresh review at `b5a27ee2` (4 new commits since `b2697c45`).

## The concern you're asking about: **not blocking**

The open thread on `loop.py:81` (accumulators changed from `= 0` to bare annotations) is **technically correct but not reachable**. `self._accumulated_input_tokens: int` really is a no-op — I confirmed `hasattr(loop, "_accumulated_input_tokens")` is `False` after `__init__`. But no `AttributeError` is reachable through any public path. I probed six:

| Path | Result |
|---|---|
| `hasattr` after `__init__` | `False` — mechanism confirmed |
| `stop()` **without** `start()` | no error |
| `start()` fails at `model.start()`, then `stop()` | no error |
| `async with agent:` where `__aenter__` fails | no error (`ConnectionError` propagates) |
| normal `start()` → `stop()` | no error |
| double `stop()` | no error |

Why it's safe: the only read site (`stop()`, `loop.py:165-168`) is guarded by `if self._session_span:`; the `start()`-failure path passes no tokens and nulls `_session_span` before returning; and the `+=` site (`:321-324`) lives in `_run_model`, which is only created *after* the resets at `:130-133`. So it's **latent fragility, not a live bug** — worth the one-line hygiene fix (restore `= 0`, or move the annotations to class level) since it costs nothing and removes a future footgun, but I would not hold the PR for it.

## Both items I previously flagged as blocking are fixed — verified, not just read

✅ **Hook propagation + restart wedge** (`loop.py:235-268`): the before-restart hook moved back *outside* the `try`, and `raise restart_exception` was added. Verified: a raising before-hook now propagates (`send_gate` stays closed), and a failed `model.start()` during restart also propagates instead of silently wedging. This is a better fix than I proposed — it addresses both the contract violation and the fail-open path.
✅ **`I001` import-sort**: gone. Only the pre-existing `UP035` remains. mypy clean.
✅ **Tool span leak on `CancelledError`** (`:411-415`): single `finally` — span now exported (1, was 0) when `stop()` cancels an in-flight tool.
✅ **Span-name cardinality**: now a static `"bidi_response"` with `response_id` as an attribute.

**Verdict: no blocking items remain from my side.** Telemetry suite **15/15** green (+3 new tests covering exactly the paths I'd flagged). The only things standing between this and merge are non-code: a rebase for green CI, and `pgrayy`'s two open questions.

<details><summary>Verification ledger + remaining non-blocking items</summary>

**Ran at `b5a27ee2`:** `test_loop_telemetry.py` → **15 passed** · `ruff --select I001,UP035` on `loop.py` → only `UP035` (pre-existing) · `ruff` full on both src files → 1 error, the same pre-existing `UP035` · `mypy --python-version 3.12` → clean · 6-path `AttributeError` probe → all clear · 4 behavioral repros (hook propagation, failed-restart propagation, tool-span-on-cancel, static span name) → all pass.

**CI:** `tests/strands/experimental/bidi/agent` → 3 failed / 28 passed. All three (`test_agent.py::test_bidi_agent_init_with_model_id`, `test_loop.py::test_bidi_agent_loop_receive_tool_use`, `test_loop.py::test_bidi_agent_loop_send_respects_event_role`) fail **identically on merge-base `f383c11a`** → pre-existing (missing optional dep + stale-main `tracking_id`), none caused by these commits. A rebase onto current `main` should clear the `tracking_id` pair.

**New code reviewed:** `cache_read_input_tokens` is accumulated (`:324`, with `or 0` correctly handling the `int | None`) and emitted; the `end_session_span` dict-comprehension refactor is semantically equivalent to the old `if > 0` chain.

**⚪ One new nit (non-blocking):** with the `finally` fix, a tool cancelled by `stop()` now closes with `StatusCode.OK` — `CancelledError` isn't caught by `except Exception`, so `tool_error` stays `None` and the span reports success for a tool that never completed. The leak is genuinely fixed; this is just accuracy. If you want it exact, catch `asyncio.CancelledError` in the `finally` path (or set a sentinel error) and re-raise.

**Still non-blocking, unchanged from my last pass** (listed for completeness, no action needed to approve): `finish_reason` singular vs semconv plural `finish_reasons` + invented `gen_ai.error.message` · no per-turn token attribution (this is `pgrayy`'s open question — usage lands only on the session span) · Gemini Live emits no `bidi_response` spans / Nova Sonic cardinality mismatch · no defensive isolation around `_start_span` · `gen_ai.system_instructions` as span attribute rather than event · the pre-existing `stop()`-races-restart race (proven pre-existing last pass — separate issue, say the word and I'll file it).

</details>

Automated review — a human still owns the approve. Nice turnaround on the error-propagation fix in particular.

strands-py/src/strands/experimental/bidi/agent/loop.py
```diff
@@ -69,6 +76,13 @@ def __init__(self, agent: "BidiAgent") -> None:
 
         self._send_gate = asyncio.Event()
 
+        self._tracer = get_tracer()
+        self._session_span: Span | None = None
+        self._accumulated_input_tokens: int
```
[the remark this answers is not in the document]
mehtarac: It's consistent with the established pattern in the same __init__, the AttributeError path is unreachable, and both suggested "fixes" either reintroduce redundancy or don't achieve the stated goal

pgrayy (approved):

PR #3283, opened 2026-07-15
feat: add snapshot session manager to python

### LINKED ISSUE #3182: [FEATURE] Add snapshot-based session management
### SDK Language

Python

### Problem Statement

The Python SDK’s session managers (FileSessionManager, S3SessionManager, RepositorySessionManager) use an older message-log design that never touches the Snapshot primitive the SDK already has. The TypeScript SDK has been snapshot-based from the start. Python has the Snapshot primitive but no manager to persist it automatically, so the two SDKs are out of parity on session persistence.

### Proposed Solution

Add a snapshot-based session manager and its storage backends, matching the TS key layout

### Use Case

SDK parity

### Alternatives Solutions

_No response_

### Additional Context

_No response_

opieter-aws:
## Description

The Python SDK had no way to persist an agent as a whole. The shipped session managers are a message-log design — one file or S3 object per message — which cannot express "restore the agent to how it looked three turns ago." The `Snapshot` primitive (`agent.take_snapshot`/`load_snapshot`) already existed and is used by the A2A executor, but nothing persisted snapshots to storage. TypeScript has shipped snapshot-based sessions with checkpointing for a while; this closes that gap.

`SnapshotSessionManager` captures the agent as one versioned blob per save. A mutable `snapshot_latest` gives crash/restart resume; append-only immutable snapshots keyed by UUIDv7 give time-travel restore. It builds on the unified `Storage` primitive rather than its own backend, so one storage configuration serves sessions, memory, and anything else layered on `Storage`. The key layout is byte-identical to TS (`session/<id>/scopes/agent/<agent_id>/snapshots/…`), so both SDKs share a storage convention.

**Single agents only.** Graph/Swarm and BidiAgent are rejected at their initialization event with `NotImplementedError` pointing at the message-log managers, rather than attaching and silently persisting nothing. Multi-agent snapshot persistence needs upstream fixes to `Graph`/`Swarm` `serialize_state`/`deserialize_state`: those were built for interrupt-resume inside a live process and lose state across a real crash-restart boundary. They are pre-existing, being fixed separately, and deliberately out of scope here.

Nothing is deprecated. The message-log managers are untouched and remain the only option for orchestrators and Bidi.

### Public API Changes

```python
from strands import Agent
from strands.session import SnapshotSessionManager
from strands.storage import LocalFileStorage  # or S3Storage, InMemoryStorage

session = SnapshotSessionManager(
    "user-123",
    storage=LocalFileStorage(),
    save_latest_on="invocation",                        # or "message" | "trigger"
    snapshot_trigger=lambda *, agent_data, **_: True,   # optional: append a checkpoint per turn
)
agent = Agent(session_manager=session)
agent("Hello!")  # a fresh Agent over the same storage rehydrates this
```

Checkpointing and time travel:

```python
snapshot_id = await session.save_snapshot(agent, is_latest=False)  # returns the id it minted
await session.restore_snapshot(agent, snapshot_id=snapshot_id)     # rewind
await session.restore_snapshot(agent)                              # omit id -> snapshot_latest
ids = await session.list_snapshot_ids(agent)                       # oldest first
await session.delete_session()
```

Also exported: `SnapshotTrigger` (Protocol) and `SaveLatestStrategy`.

```python
session = SnapshotSessionManager(
    "user-123",
    storage=LocalFileStorage(),
)
```

Two intentional deviations from the TS shape, both additive: `save_snapshot` returns the new id so callers need not follow up with `list_snapshot_ids`, and the trigger argument is named `agent_data` to match TS's `SnapshotTriggerParams.agentData`.

## Related Issues

#3182 

## Documentation PR

N/A will follow up

## Type of Change

New feature

## Testing

Exercised the manager end-to-end against a real Bedrock agent across a genuine two-process restart (one process writes and exits; a fresh process over the same directory rehydrates the conversation and state and the model continues coherently), plus checkpointing/time-travel, `save_latest_on` cadence, guardrail-redaction flush, stateful-model message discard, image-bytes round-trip. The `S3Storage` backend and the stateful-model (OpenAI Responses) discard run in CI via `tests_integ/`.

## Checklist

----


opieter-aws:
@strandly-the-agent Review this PR, do API BR and focus on feature parity with Typescript

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,442 @@
...
+                snapshot is written; every later run restores from the snapshot and the
+                message-log store is never read or written again. The agent must be
+                constructed with the same conversation manager and system prompt it ran
+                with, since restore replays state into those live components.
+            **kwargs: Additional keyword arguments for future extensibility.
+
+        Raises:
+            ValueError: If ``session_id`` contains a path separator.
+        """
+        self.session_id = validate_identifier(session_id, Identifier.SESSION)
```
strandly-the-agent, line 232: 🔴 **Block: reject an empty session ID before deriving storage prefixes.** `validate_identifier("")` currently succeeds, `_session_prefix("")` becomes `session//`, and storage normalizes that to `session/`. `delete_session()` therefore lists and deletes every session in the backend. I reproduced this with two populated session IDs; both were gone after `SnapshotSessionManager("", ...).delete_session()`. TypeScript rejects an empty identifier.

The smallest local guard is:

```suggestion
        self.session_id = validate_identifier(session_id, Identifier.SESSION)
        if not self.session_id:
            raise ValueError("session_id cannot be empty")
```

Please add a regression with an unrelated populated session so this destructive prefix broadening cannot return.

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,442 @@
...
+
+    async def _on_after_invocation(self, event: AfterInvocationEvent) -> None:
+        """Save latest on invocation and fire the immutable-checkpoint trigger.
+
+        When the trigger fires, the immutable+latest write subsumes the invocation save, so the
+        agent is captured only once even under ``save_latest_on="invocation"``.
+        """
+        if self._snapshot_trigger is not None and self._snapshot_trigger(agent=event.agent):
+            await self._save_immutable_and_latest(event.agent)
+        elif self._save_latest_on == "invocation":
```
strandly-the-agent: 🔴 **Block: `"message"` mode persists state from before end-of-invocation conversation management.** Its last save comes from `MessageAddedEvent`; afterward `Agent` runs `conversation_manager.apply_management()` (`agent.py:1315`), but this branch skips the post-management save.

With `SlidingWindowConversationManager(window_size=2)`, the live agent ended with `['u2', 'a2']` and `removed_message_count=2`; restart restored `['u1', 'a1', 'u2', 'a2']` and count `0`.

Keep per-message durability, then overwrite latest with the final managed state:

```suggestion
        elif self._save_latest_on in ("invocation", "message"):
```

A restart test should compare both messages and conversation-manager state, not only write counts.

strands-py/src/strands/session/file_session_manager.py
```diff
@@ -7,6 +7,8 @@
 import tempfile
 from typing import TYPE_CHECKING, Any, cast
```
strandly-the-agent, line 10: 🔴 **API block: this blanket deprecation has no replacement for supported Graph/Swarm/Bidi workloads.** `SnapshotSessionManager.register_hooks()` registers only single-agent events, and the updated docs consequently still require `FileSessionManager`/`S3SessionManager` for Graph and Swarm. The same issue applies to the new decorators on `S3SessionManager` and `RepositorySessionManager`.

That leaves supported customers with a warning to migrate to an API that cannot preserve their sessions, followed by a promised 2.0 removal. Please retain these classes without blanket deprecation until their supported contexts have replacements. If only single-agent usage is superseded, scope the migration guidance to that usage rather than deprecating the class itself.

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,442 @@
...
+                message-log store is never read or written again. The agent must be
+                constructed with the same conversation manager and system prompt it ran
+                with, since restore replays state into those live components.
+            **kwargs: Additional keyword arguments for future extensibility.
+
+        Raises:
+            ValueError: If ``session_id`` contains a path separator.
+        """
+        self.session_id = validate_identifier(session_id, Identifier.SESSION)
+        self._storage = storage
```
strandly-the-agent: 🟡 **Preserve the unified storage namespace contract.** A namespaced view is marked so SDK consumers can skip their default prefix, but this manager keeps the supplied view while generated keys still include `session/`. I reproduced `raw.namespace("session")` writing `session/session/<sid>/...`; TypeScript checks its marker and writes `session/<sid>/...`.

Please resolve storage once as TS does: retain an already-namespaced view, otherwise namespace raw storage under `session`, and make manager keys relative to that view. Literal-key tests for raw and pre-namespaced storage would pin the parity contract.

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,442 @@
...
+# oldest-first without a separate index); the uuid4 suffix keeps ids unique within a
+# millisecond. Ids are opaque handles: callers get them from list_snapshot_ids and never
+# construct them. The fixed shape also guards the immutable-history key against traversal.
+# \A...\Z (not ^...$) so a trailing newline cannot slip past validation.
+_SNAPSHOT_ID_PATTERN = re.compile(r"\A\d{13}_[0-9a-f]{32}\Z")
+
+
+def _new_snapshot_id() -> str:
+    """Return a fresh, time-ordered immutable snapshot id."""
+    return f"{int(time.time() * 1000):013d}_{uuid.uuid4().hex}"
```
strandly-the-agent: 🟡 **This ID is unique but not time-ordered within one millisecond.** The UUID4 suffix is random, while `list_snapshot_ids()` sorts the full string. Forcing creation suffixes `ffff…` then `0000…` made listing return the newer checkpoint first, contradicting the documented oldest-first contract.

TypeScript uses UUIDv7. Before this persisted format ships, please use a monotonic sortable ID—ideally UUIDv7 for parity—or otherwise encode an intra-millisecond sequence, and add a frozen-clock test proving creation order is preserved.

strandly-the-agent (changes requested):
🔴 **Changes requested** — three reproduced blockers plus two persistence/parity defects; this substantial public primitive also needs formal API review before merge.

✅ Reviewed exact head `50bd89c8`; Python snapshot/migration tests **43 passed**, TS parity-oracle tests **72 passed**, and `git diff --check` passed. Core CI jobs are green; `label-size` remains failed and some non-core jobs are waiting.

🔴 Reproduced: empty-ID cross-session deletion; `message` mode restoring pre-compaction state; no viable Graph/Swarm/Bidi replacement for the blanket deprecations.

🟡 Parity: pre-namespaced storage writes `session/session/...`; same-millisecond IDs violate the documented oldest-first order and differ from TS UUIDv7.

❓ **Blocking API decision:** please add `api/needs-review`, complete the substantial-change review/meeting, and only then add `api/review-complete`. Should Python also expose TS-equivalent manual immutable save and latest restore (`snapshot_id=None`) before this becomes the recommended parity primitive?

<details>
<summary>API BR, docs, and verification details</summary>

### API bar-raising

The new recommended manager and two supporting types are exported directly from `strands.session`, while three established public managers are deprecated. That meets the substantial-change bar in `team/API_BAR_RAISING.md`; the PR currently has neither API label. The PR description should also include the complete public signatures/defaults and export inventory required by that process.

TypeScript exposes `saveSnapshot({ isLatest: false })` for an explicit immutable checkpoint and `restoreSnapshot()` without an ID for latest restore. Python currently creates immutable snapshots only through an invocation trigger, and `restore_snapshot()` requires an immutable ID. Please either close that gap or record the intentionally excluded workflows and rationale in the API decision.

### Docs bundle

Please correct these together:

- `session-management.mdx:181`: `LocalFileStorage()` defaults to `./.strands/`, not a temporary directory.
- `session-management.mdx:264-277`: use `prefix="production"` (the trailing slash currently produces `production//`) and remove the unsupported `endpoint_url` claim, or add that parameter to unified `S3Storage`.
- `session-management.mdx:196-225`: qualify the layout claim. The adjacent TS examples use legacy `{ snapshot: FileStorage/S3Storage }`; only unscoped unified TS storage adds the matching `session/` namespace.

### Scope

I did not repeat the four earlier inline findings that are now outdated/fixed. The existing redaction-divergence thread remains the right place to record that intentional behavioral difference.

Tests run locally:

```text
Python: 43 passed (snapshot manager + migration + migration e2e)
TypeScript: 72 passed (SessionManager + SnapshotStorageAdapter)
git diff --check: passed
```

</details>

opieter-aws:
@strandly-the-agent review the latest version

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,605 @@
...
+        # async handlers and avoid the sync bridge.
+        if self._save_latest_on == "message":
+            registry.add_callback(MessageAddedEvent, self._on_message_added)
+        registry.add_callback(AfterInvocationEvent, self._on_after_invocation)
+
+        # Multi-agent (Graph/Swarm): restore the orchestrator once on init, persist its state
+        # after each node and after the invocation. Latest-only (no immutable checkpoints),
+        # matching the TypeScript SDK's multi-agent path.
+        registry.add_callback(MultiAgentInitializedEvent, self._on_multi_agent_initialized)
+        registry.add_callback(AfterNodeCallEvent, self._on_after_node_call)
```
strandly-the-agent: 🔴 **Block: checkpoint only at a safe Graph frontier**

This new `AfterNodeCallEvent` save can run while a parallel sibling is still `EXECUTING`. At that point `Graph._compute_ready_nodes_for_resume()` treats the in-flight sibling as bypassed, so `root -> {left,right} -> join` persisted `next=['join', 'right']`.

A fresh restore then produced:

`restart_calls=['right-start', 'join', 'right-done', 'join']`

The first join ran with incomplete inputs, and the join executed twice—duplicating any external side effects. Please checkpoint at the completed-batch boundary, or serialize in-flight nodes so their descendants cannot enter the resume frontier. A regression should restart after one sibling finishes while another remains blocked.

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,605 @@
...
+        data = _serialize_multi_agent(source.id, source.serialize_state())
+        await self._storage.write(_multi_agent_key(self.session_id, source.id), data)
+
+    async def _on_multi_agent_initialized(self, event: MultiAgentInitializedEvent) -> None:
+        """Restore the orchestrator on construction (fires once per orchestrator)."""
+        await self._restore_multi_agent(event.source)
+
+    async def _on_after_node_call(self, event: AfterNodeCallEvent) -> None:
+        """Persist the orchestrator after each node completes."""
+        await self._save_multi_agent(event.source)
```
strandly-the-agent: 🔴 **Block: persist Swarm after applying the handoff**

Swarm emits `AfterNodeCallEvent` before it applies `handoff_node` and updates `current_node`. This save therefore captured `next=['first']` with `handoff='second'`; restoring it invoked:

`['first', 'second']`

The already-completed first node is replayed. Please move the checkpoint after the handoff transition, or make serialization prioritize the pending handoff over the stale current node. The regression should restore the actual after-node checkpoint and assert that only the handoff target runs.

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,605 @@
...
+        data = _serialize_snapshot(self._capture(agent))
+        await self._storage.write(_snapshot_key(self.session_id, agent.agent_id, snapshot_id=_new_snapshot_id()), data)
+        await self._storage.write(_snapshot_key(self.session_id, agent.agent_id, snapshot_id=None), data)
+
+    async def _restore_multi_agent(self, source: "MultiAgentBase") -> None:
+        """Load an orchestrator's latest snapshot into it, if one exists."""
+        data = await self._storage.read(_multi_agent_key(self.session_id, source.id))
+        if data is None:
+            return
+        source.deserialize_state(_deserialize_multi_agent(data, expected_orchestrator_id=source.id))
```
strandly-the-agent: 🔴 **Block: terminal Swarm snapshots cannot be restored safely**

An ordinary completed Swarm is saved with `status='completed'` and `next_nodes_to_execute=[]`. `Swarm.deserialize_state()` treats presence of that key as resumable, leaves `current_node=None`, and the next invocation fails at `current_node.node_id`:

`AttributeError: 'NoneType' object has no attribute 'node_id'`

An empty/terminal frontier should reset for a new invocation, as Graph already does. Please cover complete → fresh process → invoke again.

strands-py/src/strands/session/file_session_manager.py
```diff
@@ -25,6 +27,12 @@
 MULTI_AGENT_PREFIX = "multi_agent_"
 
 
+@deprecated(
+    "FileSessionManager is deprecated and will be removed in Strands SDK 2.0. Use "
```
strandly-the-agent: 🔴 **Block (latest-delta follow-up): do not deprecate Bidi toward a no-op**

The new Graph/Swarm support addresses that part of the earlier thread, but Bidi remains unsupported. The legacy managers register Bidi initialization/message/invocation hooks; `SnapshotSessionManager.register_hooks()` overrides that wiring and registers none of them. A no-Bedrock Bidi message-event repro left storage keys as `[]`.

Please either add Bidi snapshot/migration support before this deprecation, or scope the warning so Bidi users are not directed to a replacement that silently persists nothing.

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,605 @@
...
+
+    async def _save_multi_agent(self, source: "MultiAgentBase") -> None:
+        """Capture the orchestrator's state and overwrite its latest snapshot."""
+        data = _serialize_multi_agent(source.id, source.serialize_state())
```
strandly-the-agent: 🟡 **Make the multi-agent serialization contract lossless**

This manager assumes `deserialize_state(serialize_state())` preserves continuation state, but it currently does not. After restoring Swarm shared context, `self.shared_context` contained `{'first': {'fact': 'persist-me'}}` while `self.state.shared_context` remained `{}`; the next save serialized `{}` and erased the restored context.

Graph and Swarm also omit cumulative `execution_time` and usage/metrics, resetting timeout accounting and reported totals after restart. Please round-trip the canonical context and cumulative accounting fields, with a serialize → deserialize → serialize invariant test.

strandly-the-agent (changes requested):
🔴 **CHANGES_REQUESTED:** the new Graph/Swarm persistence can resume invalid state—causing early/double execution, replay, and terminal restore crashes—and the deprecation still sends Bidi users to a no-op replacement.

✅ Reviewed `2ffcf6ed` vs `ad5685f7`; `git diff --check` passed. Focused Python suites: snapshot/migration/UUID **73**, Graph/Swarm **110**, legacy sessions **131** passed. CI Gate, Python matrix, lint, docs, and codecov are green.

✅ Fresh no-Bedrock repros confirmed Graph early/double join, Swarm handoff replay, terminal restore crash, and lossy serialization round-trip.

❓ **Blocking API process:** this substantial public primitive needs explicit API review/meeting and `api/needs-review` → `api/review-complete` before merge.

❓ **Blocking parity:** should Python expose TS-equivalent manual save/latest restore, multi-agent immutable restore, and configurable node-vs-invocation saves—or is each divergence intentional and approved?

❓ **Blocking wire contract:** Python and TS use the same keys with incompatible payloads; should they share a canonical schema or use SDK/version-specific namespaces?

<details><summary>Additional verification and docs corrections</summary>

The exact latest TypeScript focused suite was not rerun; the earlier 72-pass result predates this multi-agent delta. The PR remains draft; `label-size` fails, and integration/preview/review jobs are waiting.

Please also correct the bundled documentation examples:
- `LocalFileStorage()` defaults to `./.strands/`, not a temporary directory.
- `S3Storage` has no `endpoint_url` argument, and `prefix="production/"` currently creates a doubled slash.
- The migration example points new storage at `/path/to/sessions` while the legacy `FileSessionManager` reads its unrelated default directory.

No style-only findings are included.
</details>

opieter-aws:
@strandly-the-agent review latest version

strands-py/tests_integ/test_snapshot_session.py
```diff
@@ -0,0 +1,137 @@
...
+
+        asyncio.run(manager.delete_session())
+
+
+def test_snapshot_session_time_travel(temp_dir):
+    """Immutable checkpoints allow restoring an agent to an earlier turn."""
+    test_session_id = str(uuid4())
+    store = LocalFileStorage(temp_dir)
+    manager = SnapshotSessionManager(test_session_id, storage=store, snapshot_trigger=lambda *, agent, **_: True)
+    agent = Agent(session_manager=manager)
```
strandly-the-agent, line 98: 🔴 **Block: this integration test can never pass — stale trigger kwarg.**

The trigger here still uses the old `agent=` keyword, but `_on_after_invocation` now calls it with `agent_data=` (`snapshot_session_manager.py:513`). The resulting `TypeError` is swallowed by the broad `except Exception` there, so no immutable snapshot is ever written and `assert len(ids) == 2` on line 103 fails.

I replicated the test body with a mock model (no Bedrock needed), twice:

```
STALE agent=      -> immutable ids = 0   (test asserts == 2) FAIL
FIXED agent_data= -> immutable ids = 2   pass
```

This is the only remaining `agent=`-style trigger in the repo — every unit test was already updated to `agent_data`. One word:

```suggestion
    manager = SnapshotSessionManager(test_session_id, storage=store, snapshot_trigger=lambda *, agent_data, **_: True)
```

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,534 @@
...
+                # log and fall through to the normal save below.
+                logger.exception(
+                    "agent_id=<%s>, session_id=<%s> | snapshot_trigger raised; skipping immutable checkpoint",
+                    event.agent.agent_id,
+                    self.session_id,
+                )
+        if triggered:
+            await self._save_immutable_and_latest(event.agent)
+        elif self._save_latest_on in ("invocation", "message"):
+            await self._save_latest(event.agent)
```
strandly-the-agent, line 491: 🔴 **Block: under `save_latest_on="trigger"`, a trigger that raises silently loses the whole turn.**

The comment just above the `except` states the invariant — "A caller's trigger raising must not discard the completed turn's latest save" — but this `elif` only honours it for `"invocation"` and `"message"`. Under `"trigger"` a raising trigger matches neither branch: nothing is persisted, `agent(...)` returns normally, and the only symptom is a log line.

Reproduced twice:

```
save_latest_on=trigger     -> stored keys=0  restored messages=0   <-- ENTIRE TURN LOST
save_latest_on=invocation  -> stored keys=1  restored messages=2
```

This composes with the stale-kwarg finding in `tests_integ/test_snapshot_session.py:95`: a wrong-signature trigger raises on *every* invocation, so under `"trigger"` that one typo means indefinite, total, silent persistence loss. The PR's own integ test is an instance of the mistake.

`test_raising_snapshot_trigger_still_saves_latest` (`tests/strands/session/test_snapshot_session_manager.py:92`) uses the default `"invocation"`, so this `elif` masks the gap and nothing covers `"trigger"` + a raising trigger.

Suggest treating a raised exception as "not triggered, but still save latest" — i.e. fall back unconditionally rather than gating the fallback on the strategy — plus a regression at `save_latest_on="trigger"`.

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,534 @@
...
+        # The save paths run under invoke_callbacks_async, so register them as native
+        # async handlers and avoid the sync bridge.
+        if self._save_latest_on == "message":
+            registry.add_callback(MessageAddedEvent, self._on_message_added)
+        registry.add_callback(AfterInvocationEvent, self._on_after_invocation)
+
+        # Fail loudly rather than silently persisting nothing: this manager handles single agents
+        # only, so an orchestrator must not be able to attach it and appear to be persisted.
+        registry.add_callback(MultiAgentInitializedEvent, self._reject_multi_agent)
+
```
strandly-the-agent, line 266: 🟡 **Bidi is the one path that still silently persists nothing.**

The comment two lines up sets out the principle I'd want applied everywhere here: "Fail loudly rather than silently persisting nothing." Graph/Swarm now do exactly that. But because this override never calls `super().register_hooks()`, the three `BidiAgent*` callbacks the base wires up (`session_manager.py:58-62`) are never registered — and the base's Bidi methods, which *do* raise `NotImplementedError` (`session_manager.py:132-169`), become unreachable for this manager.

Verified with a mock bidi model, twice:

```
BidiAgent + SnapshotSessionManager -> constructed with NO error; messages=1; stored keys=[]  => SILENT NO-OP
(contrast) Graph + same manager    -> NotImplementedError (loud, deliberate)
```

Is that an intentional gap (Bidi support coming separately), or worth closing the same way as Graph/Swarm — e.g. an explicit `registry.add_callback(BidiAgentInitializedEvent, self._reject_bidi)`? Either way it'd be good to have the class docstring say so; it currently mentions only the Graph/Swarm restriction.

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,534 @@
...
+        Returns:
+            True if the snapshot existed and was restored, False otherwise.
+
+        Raises:
+            ValueError: If ``snapshot_id`` is not a valid snapshot id.
+        """
+        return await self._restore(agent, snapshot_id=snapshot_id)
+
+    async def save_snapshot(self, agent: "Agent", *, is_latest: bool) -> None:
+        """Save a snapshot of the agent's current state on demand.
```
strandly-the-agent, line 382: 🟡 **This new public API ships with no docs, on a page that says it doesn't exist.**

`git diff 59bf1ed6...HEAD -- site/` is empty — the docs changes were reverted along with the multi-agent scope — while the code diff still adds a real user-facing surface: `list_snapshot_ids` (`:337`), `restore_snapshot` (`:367`), `save_snapshot` (here) and `delete_session` (`:398`).

At this head, `site/src/content/docs/user-guide/concepts/agents/session-management.mdx:275` reads `## Immutable Snapshots *(TypeScript only)*` and `:298` reads `## Deleting Sessions *(TypeScript only)*`. Both are now false, and `grep -rn SnapshotSessionManager site/` returns nothing.

Dropping the *multi-agent* docs was right; dropping the single-agent docs wasn't. `.github/PULL_REQUEST_TEMPLATE.md:8-10` puts docs in the same PR under `site/`, and `team/FEATURE_LIFECYCLE.md:63` puts them in the implementation step. As it stands nobody discovers `save_snapshot(agent, is_latest=False)` from a page that calls the feature TypeScript-only. (Commenting here rather than on the mdx since that file isn't in this diff.)

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,534 @@
...
+    def __init__(
+        self,
+        session_id: str = "default-session",
+        *,
+        storage: Storage,
+        save_latest_on: SaveLatestStrategy = "invocation",
+        snapshot_trigger: SnapshotTrigger | None = None,
+        migrate_from: RepositorySessionManager | None = None,
+        **kwargs: Any,
+    ) -> None:
```
strandly-the-agent, line 213: 🟡 **`**kwargs` reopens the exact hole the `save_latest_on` validation was added to close.**

The constructor deliberately validates the `save_latest_on` *value* at `:249-252`, with the comment: "Silently accepting an unknown value would register no save hooks — the session would persist nothing with no error." This `**kwargs` accepts a typo'd *parameter name* one keystroke away from the same outcome. Verified:

```
SnapshotSessionManager("s1", storage=..., save_latest_onn="trigger", snapshot_triger=lambda **k: True)
-> accepted, no error; effective save_latest_on = invocation; effective snapshot_trigger = None
```

So `snapshot_triger=` silently means "no checkpoints, ever."

`FileSessionManager`/`S3SessionManager` take `**kwargs` too, so this may be deliberate SDK-wide consistency rather than an oversight — but it's worth an explicit call on a brand-new constructor, since removing `**kwargs` later is itself a breaking change.

strands-py/tests/strands/session/test_snapshot_session_manager.py
```diff
@@ -0,0 +1,733 @@
...
+
+
+def test_unknown_save_latest_on_is_rejected(storage):
+    """A mistyped save_latest_on is rejected, rather than silently persisting nothing."""
+    with pytest.raises(ValueError, match="save_latest_on must be one of"):
+        SnapshotSessionManager("s1", storage=storage, save_latest_on="Invocation")  # type: ignore[arg-type]
+
+
+def test_graph_is_rejected_rather_than_silently_not_persisted(storage):
+    """Attaching this single-agent manager to a Graph fails loudly instead of persisting nothing."""
```
strandly-the-agent, line 83: 🟡 **Four test gaps, each confirmed by mutation (all 213 tests stayed green in every case).**

- **Swarm rejection is untested.** This Graph test is the only one; the deleted `test_swarm_persists_under_multi_agent_scope` was never replaced with a Swarm equivalent. Wrapping Swarm's event firing in `try/except NotImplementedError` left the suite green while a `SnapshotSessionManager` silently attached to a `Swarm`. Swarm fires `MultiAgentInitializedEvent` at a different point in `__init__` than Graph, so this isn't hypothetical.
- **The UUIDv7 epoch-ms contract isn't pinned.** `_snapshot_id.py:3-6` promises that the 48-bit millisecond timestamp makes lexicographic order equal creation order across writers. Mutating `new_snapshot_id()` to seconds, and separately to `time.monotonic()`, both left the suite green — either would break cross-process ordering in production. A frozen-clock assertion that the high 48 bits equal `int(time.time() * 1000)` would pin it.
- **The immutable-then-latest write order isn't pinned.** `_save_immutable_and_latest` (`snapshot_session_manager.py:484-493`) documents the order as deliberate for partial-failure safety, but swapping the two writes left the suite green. A spy recording write *order* (not just count) would cover it.
- **The two time-travel restore tests assert membership, not shape** (lines 316-317 and 373-374). Mutating `_restore` to duplicate every restored message left both green; the migration e2e tests, which use exact-list equality, correctly failed. Exact-list assertions here would make these the proof they're meant to be.

None of these blocks merge on its own, but together they're a fair ask for a new persistence primitive whose ordering and monotonicity guarantees are explicitly documented.

strandly-the-agent (changes requested):
**Changes requested** — but the headline is that scoping to a single agent was the right call: it removes 4 of my 5 prior blockers (all of them lived in the multi-agent path) and the deprecation that had no replacement. The remaining single-agent code is solid and well tested. Two issues to fix before this leaves draft, plus a few API questions worth settling while the surface is still new.

✅ Reviewed exact head `aa11cffb` · `pytest tests/strands/session/` → **213 passed** · `git diff --check` clean · CI: Python matrix (3.10–3.14 × linux/windows/macOS), Lint, Docs, CI Gate, codecov all green (`label-size` fails; integ + review jobs waiting).
✅ Both 🔴 findings reproduced with my own script, run twice, byte-identical output.
✅ My 5 inline threads from the `2ffcf6ed` round are correctly outdated (that code is gone) — not re-raising any of them.

🔴 `tests_integ/test_snapshot_session.py` cannot pass in credentialed CI: stale `agent=` trigger kwarg → 0 checkpoints written, test asserts 2.
🔴 `save_latest_on="trigger"` + a trigger that raises silently loses the entire turn (0 keys written, no error to the caller).
🟡 Bidi still persists nothing silently — the one path that skips this revision's own "fail loudly" rule.
🟡 Docs were reverted wholesale, so this new public API ships undocumented, on a page that still calls the feature "TypeScript only".

**Questions (blocking):** the `api/needs-review` label + substantial-change API meeting; the `save_snapshot(..., is_latest=bool)` shape and its `None` return; and the write-latest-but-can't-read-latest asymmetry in `restore_snapshot`. Details below.

<details>
<summary>Blocking questions in full, plus two non-blocking ones</summary>

**Blocking**

- **API process.** No `api/needs-review` label is set (`check-api-review-label` currently passes only because no API label is present). `team/API_BAR_RAISING.md:24-26` classifies "a new primitive … customers are expected to frequently use" as a **substantial** change needing an explicit meeting with ≥2 API reviewers. Should this go through that before merge rather than solo PR review? I don't think I should approve this surface on my own.
- **`save_snapshot(agent, *, is_latest: bool)`** (`snapshot_session_manager.py:382`). Three things worth a second look: the `is_latest=True` branch duplicates the existing `sync_agent` (which itself calls `save_snapshot(is_latest=True)` at `:480-482`); `is_latest=False` is a double negative for what users would search for as "checkpoint"; and it returns `None`, so a caller can't learn the id of the checkpoint it just created except by diffing `list_snapshot_ids`, which races the trigger. Would `save_checkpoint(agent) -> str` (or returning `str | None`) read better as the first public shape? Widening a `None` return later is source-compatible, so returning the id doesn't strand TS.
- **Latest-restore asymmetry.** `restore_snapshot(agent, *, snapshot_id: str)` (`:367`) forbids `None` in its annotation, but `_restore(..., snapshot_id=None)` (`:472`) already implements "restore latest" and works at runtime; TS's `restoreSnapshot({target})` does exactly that. Intentional, or should the signature be `snapshot_id: str | None = None`? Right now you can write latest but not read it.

**Non-blocking**

- **`agent_data` naming.** TS's `agentData` is a destructured params-object *field* (`strands-ts/src/session/types.ts:20-41`), which exists because TS passes one object; Python has real kwargs, and every other method on this class calls the same thing `agent`. `team/DECISIONS.md:174-178` prefers within-language clarity over cross-SDK naming. Worth lining this one up with the rest of the class?
- **"Byte-identical" storage claim.** The comment at `:67-72` says the key layout is byte-identical to TS so the layout is shared across SDKs — and the keys are. The payloads aren't: Python writes `schema_version`/`created_at`/`app_data` (`types/_snapshot.py:68-76`) where TS writes `schemaVersion`/`createdAt`/`appData` (`strands-ts/src/types/snapshot.ts:20-31`), and Python's identifier validation (path-separator check only) is far looser than TS's `^[a-z0-9_-]+$` (`strands-ts/src/session/validation.ts:10-12`). Since this is a persisted format, it's cheapest to settle now: converge the schema and identifier rules, or narrow the claim to the key layout only.

**Prior findings — all resolved or moot, listed so nothing looks dropped:** Graph parallel-join early/double execution, Swarm handoff replay, terminal Swarm restore crash, and multi-agent lossy serialization are all moot with the multi-agent path removed. The Bidi blanket-deprecation blocker is moot as framed (deprecation reverted) — the underlying silent no-op is re-raised on its own merits inline. Earlier-round items confirmed still fixed: empty `session_id`, `"message"`-mode post-management save, storage double-namespacing, UUIDv7 ordering, bounded `delete_session`.

</details>

strands-py/src/strands/session/snapshot_session_manager.py
opieter-aws: Docs will be a follow-up

strands-py/src/strands/session/snapshot_session_manager.py
opieter-aws: Deliberate pattern we're adopting here

opieter-aws:
@strandly-the-agent can you review please?

strandly-the-agent (commented):
**All blocking findings from my previous two rounds are fixed, and I verified each one.** No code blockers left from me. I'm not stamping approve only because this is a substantial new public primitive that hasn't been through API review — that gate is a human call, not a code problem.

✅ Exact head `0906fd71` · `pytest tests/strands/session/` → **216 passed** (was 213; the 3 new tests are regressions for these fixes) · `git diff --check` clean · CI green except `label-size` (integ + review jobs waiting).

| Finding | Status (each probe re-run twice, byte-identical) |
|---|---|
| 🔴 integ test's stale `agent=` trigger kwarg | ✅ **fixed** — now `agent_data`, writes 2 checkpoints where it wrote 0. The thread is still open; safe to resolve. |
| 🔴 `save_latest_on="trigger"` + raising trigger lost the whole turn | ✅ **fixed** via the `trigger_failed` fallback — the turn now survives under all three strategies, and my control confirms a trigger returning `False` still correctly saves nothing. |
| 🟡 Bidi silently persisted nothing | ✅ **fixed** — a real `BidiAgent(...)` now raises `NotImplementedError` at construction; Graph and Swarm too. |
| 🟡 test-gap bundle | ◐ Swarm + Bidi rejection tests added; three hardening items remain (collapsed below, non-blocking). |

Noted and **withdrawn**, no need to revisit: docs → your follow-up; constructor `**kwargs` → your deliberate pattern.

<details>
<summary>Remaining non-blocking items, the API gate, and one note on the conflicting review</summary>

**Three test-hardening leftovers from the bundle.** I checked whether the weakest one is hiding an actual bug, and it isn't — exact-shape restore is correct: rewinding to the first checkpoint yields exactly `['turn 1', 'first']`, and a fresh construct restores exactly `['turn 1', 'first', 'turn 2', 'second']`. So these are genuinely cosmetic, and fine to leave or fold into the docs follow-up:

- `test_snapshot_session_manager.py:362-363` and `:419-420` still assert membership (`"turn 1" in tru_texts`) rather than exact lists, so a duplication bug in `_restore` wouldn't trip them.
- The UUIDv7 epoch-ms contract isn't pinned — mutating `new_snapshot_id()` to seconds or `time.monotonic()` left the suite green. (Ordering itself is fine: ids came out strictly increasing in my run.)
- The deliberate immutable-then-latest write order in `_save_immutable_and_latest` isn't asserted.

**API review gate (unchanged, and the reason this isn't an approve).** Still no `api/needs-review` label — `check-api-review-label` reports SUCCESS only because no API label is present. `team/API_BAR_RAISING.md:24-26` puts a new frequently-used primitive in the "substantial" bucket (explicit meeting, ≥2 API reviewers). The design questions from my last round are still open and are for that discussion rather than another review round from me — briefly: whether `save_snapshot` should return the checkpoint id (today it returns `None`, so a caller can't name what it just created), whether `restore_snapshot`'s `snapshot_id` should be `str | None = None` since `None` already restores latest at runtime, and `agent_data` vs `agent` for the trigger kwarg. Happy to re-run anything after that lands.

**One factual note, since there are two bot reviews with opposite verdicts.** The other review at `e3dd179` requests changes citing multi-agent resume bugs in `graph.py`/`swarm.py`. Those files aren't part of this PR's diff at this head — it's 8 files, all under `strands-py/src/strands/session/` plus tests (`git diff --stat f383c11a...0906fd71`), and `e3dd179` itself touched only 3 files. The multi-agent blockers I reproduced back at `2ffcf6ed` went away with the scope reduction; I could not reproduce anything of the sort here because the orchestrator code paths are no longer in this change.

Still a draft, so nothing merges yet regardless.

</details>

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,552 @@
...
+logger = logging.getLogger(__name__)
+
+SaveLatestStrategy = Literal["message", "invocation", "trigger"]
+"""Controls how often ``snapshot_latest`` is saved automatically.
+
+- ``"invocation"``: after every agent invocation completes (default; balances durability and I/O).
+- ``"message"``: after every message added (most durable, highest I/O).
+- ``"trigger"``: only when ``snapshot_trigger`` fires (or manually via ``save_snapshot``).
+
+Guardrail redactions are flushed immediately under every strategy, including ``"trigger"``,
```
mkmeral, line 59: this makes sense to me. why isn't it the case for TS SDK?

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,552 @@
...
+            agent: The agent the message was appended to (unused).
+            **kwargs: Additional keyword arguments for future extensibility.
+        """
+
+    # -- Public time-travel API --
+
+    async def list_snapshot_ids(
+        self, agent: "Agent", *, limit: int | None = None, start_after: str | None = None
+    ) -> list[str]:
+        """List immutable snapshot ids for an agent, oldest first.
```
mkmeral, line 338: why oldest first? wouldn't you generally want the latest one?

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,552 @@
...
+            logger.warning(
+                "agent_id=<%s>, session_id=<%s> | agent had existing messages that were overwritten by session restore",
+                agent.agent_id,
+                self.session_id,
+            )
+
+        # Stateful models manage history server-side, so restored messages would drift
+        # from the server's view. Keep the restored model_state and drop the messages.
+        if restored and agent.model.stateful and len(agent.messages) > 0:
+            logger.debug(
```
mkmeral: nit: warn? 🤔

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,552 @@
...
+        return Snapshot.from_dict(decoded)
+    except (KeyError, TypeError, AttributeError) as error:
+        raise SnapshotException(f"Failed to deserialize snapshot: {error}") from error
+
+
+@runtime_checkable
+class SnapshotTrigger(Protocol):
+    """Decides whether to write an immutable checkpoint after an invocation."""
+
+    def __call__(self, *, agent_data: "Agent", **kwargs: Any) -> bool:
```
mkmeral, line 170: why not `agent: "Agent",`?

strands-py/src/strands/session/snapshot_session_manager.py
opieter-aws: I am not sure why TS is implemented as it is, but I also agree that this is the better approach

strands-py/src/strands/session/snapshot_session_manager.py
opieter-aws: Naming here is in parity with TS. I'm not opposed to changing to `agent` at all (think that's better too), if we accept the drift!

strands-py/src/strands/session/snapshot_session_manager.py
opieter-aws: Oldest first makes sense because we use UUIDV7, so sorting this will automatically give you oldest first. If you want latest, you'd call `restore_snapshot(agent)`, which gives latest by default, so this is only used for time-travel (explicitly not latest).

Swapping to latest first would make `start_after` pagination confusing, and create parity gap with TS (also does oldest first).

If we want the option for latest first I'd suggest adding a flag like `reverse: true`. Let me know if you think that's worth it.

strands-py/src/strands/session/snapshot_session_manager.py
opieter-aws: Fair call, paired with TS here, but agree this could be a warning (though this is expected behavior for stateful models)

mkmeral (dismissed):

strands-py/src/strands/session/snapshot_session_manager.py
Unshure: I think `agent_data` was meant to represent a base interface across different agent types (bidi-agent, A2AAgent, etc). But im also fine with just `agent`

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,552 @@
...
+    """
+
+    def __init__(
+        self,
+        session_id: str = "default-session",
+        *,
+        storage: Storage,
+        save_latest_on: SaveLatestStrategy = "invocation",
+        snapshot_trigger: SnapshotTrigger | None = None,
+        migrate_from: RepositorySessionManager | None = None,
```
Unshure: What are some of the alternatives to including this on the SnapshotSessionManager? Can we have docs with a script that help customers migrate their old session manager to the new one? Or maybe add this method to the old session manager?

I would prefer to have the new version of SessionManager to be a bit more "pure"

strands-py/src/strands/session/_snapshot_id.py
```diff
@@ -0,0 +1,73 @@
...
+# UUIDv7 monotonic counter state (RFC 9562 method 2): within a single millisecond, ids that
+# would otherwise collide are ordered by a counter seeded into the random bits, so sorting by
+# the id preserves creation order even for sub-millisecond bursts. Guarded by a lock because
+# snapshots can be written from the sync bridge and async paths on different threads.
+_lock = threading.Lock()
+_last_ms = -1
+_counter = 0
+
+
+def new_snapshot_id() -> str:
```
Unshure: nit - Would be nice to make this a more general method of just generating a UUID7 id, and then we can use it across the sdk

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,552 @@
...
+        """
+        data = _serialize_snapshot(self._capture(agent))
+        snapshot_id = None if is_latest else _new_snapshot_id()
+        await self._storage.write(_snapshot_key(self.session_id, agent.agent_id, snapshot_id=snapshot_id), data)
+        return snapshot_id
+
+    async def delete_session(self) -> None:
+        """Delete all snapshots for this session."""
+        keys = await self._storage.list(_session_prefix(self.session_id))
+        semaphore = asyncio.Semaphore(_DELETE_CONCURRENCY)
```
Unshure, line 405: Should this just live in the storage layer? I dont think session should need to worry about this

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,552 @@
...
+    except (ValueError, UnicodeDecodeError) as error:
+        raise SnapshotException(f"Failed to deserialize {description}: {error}") from error
+    if not isinstance(decoded, dict):
+        raise SnapshotException(f"{description} is not an object: got {type(decoded).__name__}")
+    return decoded
+
+
+def _serialize_snapshot(snapshot: Snapshot) -> bytes:
+    """Serialize a snapshot to JSON bytes, base64-encoding any bytes content."""
+    return _encode_json(snapshot.to_dict())
```
Unshure: nit: `_encode_json` is just a one line function, and I think this redirection makes the code less readable. I would prefer to just have this all live in one method, but definitely not a blocker

strands-py/src/strands/session/snapshot_session_manager.py
```diff
@@ -0,0 +1,552 @@
...
+        session = SnapshotSessionManager("my-session", storage=LocalFileStorage())
+        agent = Agent(session_manager=session)
+        ```
+    """
+
+    def __init__(
+        self,
+        session_id: str = "default-session",
+        *,
+        storage: Storage,
```
Unshure: Do we need to require storage? The legacy session manager assumes file, so can we do that here too?

strands-py/tests_integ/test_snapshot_session.py
```diff
@@ -0,0 +1,140 @@
...
+def test_snapshot_session_time_travel(temp_dir):
+    """Immutable checkpoints allow restoring an agent to an earlier turn."""
+    test_session_id = str(uuid4())
+    store = LocalFileStorage(temp_dir)
+    manager = SnapshotSessionManager(test_session_id, storage=store, snapshot_trigger=lambda *, agent_data, **_: True)
+    agent = Agent(session_manager=manager)
+    agent("My favorite color is blue.")
+    agent("My favorite number is seven.")
+
+    import asyncio
```
Unshure: nit - move to top level import

strands-py/src/strands/session/_snapshot_id.py
opieter-aws: Good call, I'll rename (and try to avoid a `utils` folder)

strands-py/src/strands/session/snapshot_session_manager.py
opieter-aws: Good suggestion, happy to make `FileStorage` the default here

strands-py/src/strands/session/snapshot_session_manager.py
opieter-aws: Yes, it should! Discussed with Liz that we want this, but didn't want to increase the scope of this feature to implement it for all Storage implementation. Do you suggest we exclude it here for now and accept the gap until Storage layer provides this?

strands-py/src/strands/session/snapshot_session_manager.py
opieter-aws: As discussed offline, we're deferring this decision and shipping the snapshot session manager as-is for now

strands-py/src/strands/session/snapshot_session_manager.py
Unshure: If we can refactor this into storage later, then im fine with leaving it here for now

Unshure (approved):

opieter-aws:
@strandly-the-agent create 3 follow-up issues:
1. Support snapshot session managers for multi-agent
2. Implement a migration for non-snapshot session managers
3. A follow-up on those to deprecate the legacy session managers once we have a migration path

strandly-the-agent:
Filed the three follow-ups:

- **#3662** — Support snapshot session management for multi-agent (Graph/Swarm). Includes the four resume defects that were reproduced against the withdrawn `5b3120cf` revision, so a second attempt starts with the landmines mapped rather than rediscovering them. The Graph fan-in one is already tracked on its own as #3633, so #3662 links it instead of duplicating it.
- **#3663** — Provide a migration path from the message-log managers. Frames the open design decision (standalone utility vs. `migrate_from=` vs. a method on the legacy managers) with pros/cons, and lists what a migration has to carry beyond messages — conversation-manager state incl. `removed_message_count`, redacted variants, model state, and the broken-tool-use repair. The removed implementation from the commit before `393f25d` is a working reference.
- **#3664** — Deprecate the message-log managers, gated on the above. Its checklist also includes **Bidi**, which is easy to miss: `SnapshotSessionManager` rejects `BidiAgent` too, and the legacy managers are its only implementation today (`repository_session_manager.py:351,410,429`), so a blanket deprecation would strand Bidi users the same way the reverted one would have.

Two notes:

- **No labels applied.** The `feature_request.yml` template declares `labels: ["triage"]`, but no `triage` label exists in this repo, so I left the issues unlabelled rather than guess. Each body ends with a recommended set (`enhancement`, `python`, `area-sessions`, `area-persistence`, plus `area-multiagent`/`sdk-parity` on #3662 and `breaking change` on #3664).
- **Docs have no issue of their own.** They're a checklist item on #3664, but the gap is real today — the session-management guide still labels immutable snapshots and session deletion as TypeScript-only and doesn't mention `SnapshotSessionManager`. Happy to file that as a fourth issue if you'd like it tracked separately.

PR #3301, opened 2026-07-16
fix(python): report actual latency in model provider metadata

yonib05:
## Description

The bedrock provider reports real latency from the Converse API metrics, but six providers (openai, anthropic, gemini, litellm, llamaapi, openai_responses) hardcoded `latencyMs: 0` with a TODO because their upstream APIs do not return latency. Anyone consuming `EventLoopMetrics` or the metadata metrics for these providers gets a misleading zero, which breaks latency dashboards and makes provider comparisons useless.

Since the upstream APIs do not supply the number, each provider now measures wall-clock time from just before the API call to the point the usage metadata is emitted (`time.perf_counter`, rounded to int ms) and reports that. This is the same quantity a caller would observe around the stream, so it is honest, and it flows through the existing `format_chunk` metadata path unchanged. The anthropic provider also picks up the measured value in place of its hardcoded zero snapshot path.

Test assertions that pinned `latencyMs: 0` on end-to-end stream tests now use `unittest.mock.ANY`; assertions on the pure `format_chunk` unit tests keep exact values since those are deterministic. Two new tests (openai, anthropic) drive a mocked slow response and assert the measured latency is at least the injected delay.

## Related Issues

None

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

`hatch run prepare` passes end to end: format, lint, typecheck, and 4424 tests passed / 5 skipped. The two new tests inject a 50ms delay via a mocked client and assert `latencyMs >= 50` flows through to the metadata event.

## Checklist

----

### DIFF
--- a/strands-py/src/strands/models/anthropic.py
+++ b/strands-py/src/strands/models/anthropic.py
@@ -7,6 +7,7 @@ import base64
 import json
 import logging
 import mimetypes
+import time
 from collections.abc import AsyncGenerator
 from typing import Any, TypeVar, cast
 
@@ -390,7 +391,7 @@ class AnthropicModel(Model):
                     "metadata": {
                         "usage": usage_chunk,
                         "metrics": {
-                            "latencyMs": 0,  # TODO
+                            "latencyMs": event.get("latency_ms", 0),
                         },
                     }
                 }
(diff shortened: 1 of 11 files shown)


strands-py/src/strands/models/anthropic.py
```diff
@@ -480,6 +481,7 @@ async def stream(
         logger.debug("request=<%s>", request)
 
         logger.debug("invoking model")
+        start_time = time.perf_counter()
```
pgrayy, line 484: This time will also measure all the operations around the model call including hook call backs on the yielded events. A long hook call could create a misleading result.

PR #3306, opened 2026-07-16
fix(models/gemini): emit thought signature as its own reasoning delta

ebarkhordar:
## Description

`GeminiModel._format_chunk` builds a single `reasoningContent` delta that carries both `text` and
`signature` keys. The shared aggregator dispatches on key presence, not on value:

```python
# strands-py/src/strands/event_loop/streaming.py, handle_content_block_delta
if "text" in delta_content["reasoningContent"]:
    ...
elif "signature" in delta_content["reasoningContent"]:
```

The `text` key is set unconditionally by the dict literal in `gemini.py`, so for Gemini the `elif`
arm never runs. `state["signature"]` is never set, and `handle_content_block_stop` leaves the
signature out of the reasoning block.

The adapter then reads that signature back on the next turn. `_format_request_content_part` does:

```python
thought_signature = content["reasoningContent"]["reasoningText"].get("signature")
```

which always sees `None`, so `thought_signature=None` is always sent back to Gemini. The adapter
base64-encodes the signature on the way in (`gemini.py:401`), drops it in aggregation, then looks
for it again on the way out and finds nothing.

**Invariant this restores:** a thought signature that Gemini issues on a reasoning part survives
stream aggregation, so the signature sent back on the next turn is the one the model issued.

`#1703` landed exactly this round-trip for the `toolUse` path. This is the same fix for the
`reasoningText` path, so signature preservation is consistent within the provider.

**The fix:** emit the signature as its own `contentBlockDelta`, which is the convention the other
providers already follow. `anthropic.py:314` emits `signature_delta` as its own single-key delta,
and `bedrock.py:1139` yields the reasoning text delta and then yields the signature as a separate
delta, which is the same shape this PR gives Gemini. Across every adapter that builds a
`reasoningContent` delta (`anthropic`, `bedrock`, `gemini`, `llamacpp`, `openai`,
`openai_responses`), Gemini was the only one packing both keys into one delta, which is why the
`if/elif` holds for every other provider.

### Why not fix the aggregator instead

The one-word change at `streaming.py` (`elif` to `if`) also makes the signature survive, and it
keeps the existing tests green. I measured it rather than assume: `handle_content_block_delta`
returns a single `typed_event`, so the signature branch overwrites it and the
`ReasoningTextStreamEvent` is lost. Reasoning text would stop reaching streaming consumers.

Measured on `8a292d9`, same input (a thought part with `text="test reason"`, `thought_signature=b"abc"`):

| | aggregated signature | `ReasoningTextStreamEvent` |
|---|---|---|
| `main` | dropped | emitted |
| `elif` to `if` in `streaming.py` | preserved | **lost** |
| this PR (adapter emits its own delta) | preserved | emitted |

I chose the adapter-only seam because it touches no shared aggregation code and so has no
cross-provider blast radius. If you would rather fix the dispatch in `streaming.py` and make it
emit multiple events, that is your call and I am happy to redo it that way.

### One open question

If Gemini can attach a complete `thought_signature` to more than one part inside a single reasoning
block, `state["signature"] += ...` concatenates them. Feeding two signed thought parts through my
branch produces `'czE=czI='`, which is two base64 values joined rather than one signature. I could
not verify whether Gemini actually does that, because I made no live API calls, so I am flagging it
rather than guessing. On `main` that same input yields no signature at all, so this is not a case
that works today.

## Related Issues

None. I found this by reading the adapter against the aggregator, and did not open a separate issue
for it since the repro and the fix are both here. Related merged work: #1703.

## Documentation PR

Not needed. No public API or documented behavior changes.

## Type of Change

Bug fix

## Testing

All offline, no API keys, in a clean `python:3.12-slim` Docker container against `8a292d9`, with
`importlib.util.find_spec(...).origin` checked on both sides to confirm which copy ran.

- Added `test_stream_response_reasoning_signature_survives_aggregation`, which drives the real
  `model.stream()` through the real `process_stream` and asserts three things: the aggregated
  `reasoningText` carries the exact signature, the `ReasoningTextStreamEvent` is still emitted (this
  assertion is what fails on the `elif` to `if` alternative above), and feeding the aggregated
  message back through `_format_request_content_part` reproduces the original signature bytes.
- Confirmed it fails without the source change: I reverted only the `gemini.py` hunks and kept the
  new test, and it fails on unfixed source. It passes on the branch.
- Updated `test_stream_response_reasoning` and `test_stream_response_reasoning_and_text`, which
  asserted the combined-delta chunk shape. They passed while the aggregated message was wrong,
  which is why this survived.
- Compared old and new emission across a variant matrix (plain text, thought without signature,
  thought with signature, signed thought followed by plain text, empty-text thought with
  signature). Output is identical except the signed-thought path.
- `hatch test tests`: 4421 passed, 5 skipped, 2 failed. Both failures are in
  `tests/strands/storage/test_local_file_storage.py` and are not from this change: they fail
  identically on pristine `8a292d9`, because my container runs as root and root ignores the `chmod`
  those two tests use to force a write error.
- `hatch fmt --linter --check`: `ruff check` passes. `ruff format --check` reports both changed
  files already formatted. `mypy` reports no issues across 223 source files, though I ran it with
  `--follow-imports=silent` because the default run exhausts memory on my machine, so that leg is
  not byte-for-byte what CI runs.
- No new test warnings: the Gemini test module reports zero warnings on this branch and on `main`.

What I did not verify: I made no live Gemini API calls, so everything above is about this SDK's own
round-trip, not about how Google's service responds to a request with a missing thought signature. I
also ran the unit suite on Python 3.10 only, not the full 3.10 to 3.14 matrix, so I did not run
`hatch run prepare` end to end.

## Checklist

----

I used an AI coding assistant on this change. I reproduced the bug, ran every check quoted above
myself, and can explain and defend each line.


JackYPCOnline:

strands-py/src/strands/models/gemini.py
```diff
@@ -582,6 +584,15 @@ async def stream(
                             },
                         )
```
strandly-the-agent, line 637: 🔴 **Handle the documented empty-text signature part**

This remains nested under `if part.text`, so a trailing streamed part shaped like `{text: "", thought_signature: ...}` emits no signature delta. Google’s thought-signature guidance explicitly says this shape can occur and to parse all parts through `finish_reason`; `google-genai` 2.13.0 also carries this exact SSE shape in its stream tests.

A cold `GeminiModel.stream → process_stream → next request` repro produced `signature deltas=[]` and replayed `thought_signature=None` (the control replayed the original bytes). Please process `thought_signature` independently of truthy text while preserving its original reasoning-part/block association, and add this stream shape to the round-trip regression test.

strandly-the-agent (changes requested):
🔴 **Changes requested:** the split fixes signatures attached to non-empty reasoning text, but Gemini’s documented empty-text signature part is still dropped.

✅ Reviewed exact head `6894243b` vs base `8a292d9e` · 124/124 focused tests · Ruff/format/`git diff --check` · cold offline round-trip repro
⚠️ No live Gemini call; aggregate CI was still pending.

- **Correctness / LLM context:** one blocker inline. Google documents that a streamed non-function-call signature may arrive in a part with empty text; the current truthy-text guard discards it, so the next request replays `thought_signature=None`.
- **Adversarial:** reproduced with real `GeminiModel.stream → process_stream → request formatting` on supported `google-genai` 2.13.0. The exact base also fails, but this PR’s focused signature-round-trip invariant remains incomplete.
- **Tests:** the new single-part regression is valid. Please add the documented trailing empty-text signature shape.
- **Suppressed:** synthetic multi-signature/block state hazards were not raised because I could not establish that Gemini emits those exact sequences. The extra raw/typed signature event is intended.
- **API/docs:** no public API change or API-review label needed.

</details>

ebarkhordar:
You are right, thanks for catching this. The signature emission was nested under `if part.text:`, so a signature arriving on a part that carries no text of its own was dropped during aggregation and the round-trip lost it.

Pushed aff5179: emit the reasoning signature whenever a part carries `thought_signature` and is not a function call, independent of `part.text`, opening a `reasoning_content` block first if none is open. Function-call parts still carry their signature on the tool-use block start, so they are unaffected.

Added `test_stream_response_reasoning_signature_on_empty_text_part_survives` (text on one part, signature on a separate empty-text part). It fails on the prior commit with exactly this symptom and passes now; the full `test_gemini.py` is at 58 passed, ruff check and ruff format clean.

Disclosure: this change was prepared with AI assistance.

ebarkhordar:
No rush. The empty-text signature part from the review is handled in aff5179 with a round-trip regression test, and CI is green, so this is ready whenever a maintainer has a moment. Happy to rebase if it drifts.

ebarkhordar:
Correction to my note above: I said CI is green here, and that was wrong.

The green checks on this branch are the access and labeler jobs. `Run integration tests` and `Trigger Strands Review` are held on the `manual-approval` environment, and `CI` and `PR Title Conventional Commits` sit at `action_required` with zero jobs started, so no test workflow has run against it.

The evidence for the fix is the local reproduction in the PR body, which I stand behind. The follow-up commit aff5179 has the same status: verified locally, not by CI here. I have posted the same correction on #3341.

ebarkhordar:
Codecov flagged the block-opening path in this diff: when a signature arrives while no reasoning block is in progress, the emission has to close whatever block is open and start a `reasoning_content` one, and neither existing test reaches that branch since both feed the signature on or after a thought-text part.

Pushed da573d2 with the two shapes that do reach it: a signature part following a plain text part, and a candidate whose only part carries a signature. In both the aggregated message ends up with a reasoning block holding the signature, and feeding it back through `_format_request_content_part` returns the original `thought_signature` bytes. Locally the gemini suite is 60 passed and those lines are covered; I have not seen the upstream run for this commit yet.

JackYPCOnline:

strands-py/src/strands/models/gemini.py
```diff
@@ -582,6 +584,24 @@ async def stream(
...
 
+                    # A thought signature can arrive on a part carrying no text of its own (Gemini
+                    # attaches it to a trailing empty part), so emit it independently of part.text;
+                    # gating on part.text drops those signatures during aggregation. Function-call
+                    # parts carry their signature on the tool-use block start instead (see above).
```
strandly-the-agent, line 641: ❓ **Non-blocking question — a signature on a plain answer part gets relocated onto a synthetic thought part**

When the signature rides on a non-thought part, this branch closes the text block and opens a reasoning block for the signature alone. Cold repro at this head:

```
in:  Part(text="The answer is 42.", thought_signature=b"sig-on-answer")
out: [{'text': 'The answer is 42.'},
      {'reasoningContent': {'reasoningText': {'text': '', 'signature': 'c2lnLW9uLWFuc3dlcg=='}}}]
next request: [('The answer is 42.', None, None), ('', True, b'sig-on-answer')]
```

So the real answer part loses the signature and we send back an empty `thought=True` part that the model never produced. Google documents exactly this input shape for non-function-call responses (“the signature may be in the last part — for example the last text part”) and also says to return a signature inside its original `Part` and not to merge signed with unsigned parts. `google-genai` 2.16.0 serializes the synthetic part without complaint, and for non-FC turns omitting a signature is documented as non-fatal, so I could not show a concrete failure offline — hence a question, not a blocker.

The shared `ContentBlock` schema has nowhere to hang a signature on a plain text block, so there's no clean fix inside the adapter. Is this relocation the tradeoff you intend over the old behavior of dropping it? Either way it's worth pinning with a regression test, since nothing currently covers this shape.

strandly-the-agent (approved):
✅ **Approving** — supersedes my 2026-07-21 changes-requested. The dropped-signature blocker is fixed and byte-verified at `da573d22`; nothing remaining blocks.

- ✅ **My blocker:** a trailing `Part(text="", thought_signature=…)` now round-trips the exact original bytes (`b'sig-final'`); the same repro still fails at `6894243b`.
- ✅ 127/127 focused tests, `ruff check` + `ruff format`, `git diff --check`, clean worktree — and CI is green at this head (12 unit matrices, Lint, Docs, codecov/patch, CI Gate), with only `Run integration tests` waiting on manual approval.
- ✅ Regression sweep clean: tool use with and without a signed function-call part, parallel FCs, a signature interleaved between tool calls, empty / `b""` / 128 KiB signatures, block start‑stop pairing, `stopReason`.
- ⚠️ Two edge shapes reproduce oddly, but their real-Gemini consequence isn't verifiable offline — one inline question below, and I've replied on the open `state["signature"]` thread with the crash side of your "one open question".

- **Pre-existing, already tracked:** `handle_content_block_stop` never clears `state["signature"]`, so a signed reasoning block leaks its signature onto a later unsigned one (my repro replays `b'sigONE'` on both parts). Already flagged in `github-actions`' thread; I only added the crash evidence there rather than opening a second thread.
- **Provenance:** I reproduced the identical concatenation at merge base `8a292d9e` by driving the shared aggregator with Anthropic/Bedrock-shaped signature deltas — no Gemini code involved — so it is shared and pre-existing in `streaming.py`, not introduced here. Anthropic passes the signature string through unchanged (`anthropic.py:158`), which is why only Gemini turns it into a hard failure.
- **Tests:** the three new tests each fail on unfixed source and drive the real `stream → process_stream → _format_request_content_part` round trip — solid. Minor asymmetry: the first asserts one block rather than the whole `content` list; fine as is.
- **Untested shapes:** two signed parts in one response, and signature lengths whose base64 padding surfaces the decode error class. Both trace to the shared aggregator gap, not to this diff.

</details>

strands-py/src/strands/models/gemini.py
strandly-the-agent: Verified fixed at `da573d22` — resolving.

`aff5179b` moved the emission out from under `part.text`, so the documented trailing empty-text part now lands in the open reasoning block:

```
in:  Part(text="thinking", thought=True), then Part(text="", thought_signature=b"sig-final")
out: {'reasoningText': {'text': 'thinking', 'signature': 'c2lnLWZpbmFs'}}
next request: Part('thinking', thought=True, thought_signature=b'sig-final')   # exact original bytes
```

The same repro still drops the signature at `6894243b`, and the three new tests each fail on unfixed source, so the regression is genuinely pinned. Thanks for the association fix rather than just an ungated emit — keeping the signature on the block the thought text opened is the part that matters here.

strands-py/src/strands/models/gemini.py
ebarkhordar: Reproduced this at `da573d22` in a clean container (google-genai 2.17.0), and your read is exactly right:

```
in:  Part(text="The answer is 42.", thought_signature=b"sig-on-answer")
out: [{'text': 'The answer is 42.'},
      {'reasoningContent': {'reasoningText': {'text': '', 'signature': 'c2lnLW9uLWFuc3dlcg=='}}}]
next request: [('The answer is 42.', None, None), ('', True, 'c2lnLW9uLWFuc3dlcg==')]
```

The reason it relocates rather than staying put is a shared-schema limit, not an adapter choice. I parsed the type definitions rather than grepping them: `ContentBlock` (`types/content.py:79`) has 10 keys and `ContentBlockDelta` (`types/streaming.py:114`) has 4, and in the whole `types/` package the string `signature` appears as a field in exactly two places, `ReasoningTextBlock.signature` and `ReasoningContentBlockDelta.signature`. Both live inside a reasoning block. A plain `text` block is `text: str` and has nowhere to carry a token, so the emitter's only two options today are to open a reasoning block for it or to drop it, and dropping loses the token outright.

There is already precedent for the fix you are pointing at, on the sibling path: tool-use blocks solved this by adding `ToolUse.reasoningSignature` (`types/tools.py:67`, `types/content.py:170`), threaded through `event_loop/streaming.py:198`, `:318` and consumed by both `gemini.py:215` and `litellm.py:144`. The equivalent for text would be the same shape, a signature field on the text block plus aggregator plumbing, which is a cross-provider change to shared code and the same boundary you drew on the `handle_content_block_stop` thread.

So I would rather not widen this PR to reach it. Happy to open a follow-up covering both that and the stale-signature leak if a maintainer wants them tracked together.

strands-py/src/strands/models/gemini.py
```diff
@@ -582,6 +584,24 @@ async def stream(
...
+                    if part.thought_signature and not part.function_call:
```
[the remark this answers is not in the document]
ebarkhordar: Confirmed and agreed this is out of scope here. `handle_content_block_stop` clearing `reasoningText` but not `signature` sits in shared aggregation code and affects the anthropic and bedrock separate-delta shapes too, so fixing it from a Gemini adapter PR would be the wrong place for it.

Resolving the thread so it stops blocking merge, with the finding carried over to the question on line 590 rather than dropped.

JackYPCOnline (approved):

Issue #1511, opened 2026-01-19
[Feature Request] Real-time Streaming Events During Context Reduction


tahitimoon:
## Problem Description

When using [SummarizingConversationManager](file:///Users/bingjiang/Documents/quickfinai-model/.venv/lib/python3.13/site-packages/strands/agent/conversation_manager/summarizing_conversation_manager.py#52-268) with [stream_async()](file:///Users/bingjiang/Documents/quickfinai-model/.venv/lib/python3.13/site-packages/strands/agent/agent.py#527-630), there's no way to notify users in real-time when context compression is happening.

### Current Behavior

```python
# In agent.py (lines 716-726)
except ContextWindowOverflowException as e:
    self.conversation_manager.reduce_context(self, e=e)  # Blocks for 30+ seconds
    # ... retry event loop
```

The [reduce_context()](file:///Users/bingjiang/Documents/quickfinai-model/.venv/lib/python3.13/site-packages/strands/agent/conversation_manager/summarizing_conversation_manager.py#124-175) call is **synchronous and blocking**. During this time:
- No events are yielded to the async iterator
- The event loop is blocked
- Users see a frozen UI with no feedback

### Expected Behavior

Users should receive streaming events during context compression:
- `conversation_compacting` - when compression starts
- `conversation_compacted` - when compression completes

This allows frontend applications to show "Compacting conversation..." feedback to users during the 30+ second wait.

### Option 1: Hook-based Events (Minimal Change)

Add new hook events for context reduction:

```python
# New hook events
class BeforeContextReductionEvent:
    agent: Agent
    exception: ContextWindowOverflowException

class AfterContextReductionEvent:
    agent: Agent
    original_message_count: int
    compressed_message_count: int
    removed_count: int
```

This allows users to subscribe to these events via `HookProvider`.

### Option 2: Yield Events During Exception Handling

Modify [_execute_event_loop_cycle](file:///Users/bingjiang/Documents/quickfinai-model/.venv/lib/python3.13/site-packages/strands/agent/agent.py#685-731) to yield events during context reduction:

```python
except ContextWindowOverflowException as e:
    # Yield event before reduction
    yield {"context_reduction": "starting", "message_count": len(self.messages)}
    
    self.conversation_manager.reduce_context(self, e=e)
    
    # Yield event after reduction
    yield {"context_reduction": "completed", "new_message_count": len(self.messages)}
    
    # Retry
    async for event in self._execute_event_loop_cycle(...):
        yield event
[...]

PR #3307, opened 2026-07-16
feat: add context manager class design doc

lizradway:
## Description
 adds context manager class design doc

## Type of Change

Bug fix
New feature
Breaking change
Documentation update
Other (please describe):

## Testing

## Checklist

----

### DIFF
--- a/team/designs/0015-context-manager.md
+++ b/team/designs/0015-context-manager.md
@@ -1,0 +1,1421 @@
+# Design: ContextManager Class
+
+**Status**: Proposed  
+**Date**: 2026-06-09 (updated 2026-07-01)  
+**Scope**: TypeScript + Python SDK  
+**Related**: [Context Strategy Design (v1 facade)](https://github.com/strands-agents/docs/pull/831), [MemoryManager Design](https://github.com/strands-agents/docs/pull/844)
+
+> **Note**: API shapes in this document are drafted for illustration purposes and are likely to change through bar raising and review.
+
+---
+
+## 1. Problem
+
+### What customers hit
+
+**"My coding agent forgets what it just read."** A coding agent reads 5 files, makes edits, then needs to reference the first file again. But it was summarized away 3 tool calls ago: the original content is gone. The agent has to re-read the file, burning a tool call and tokens. This happens every few turns in long sessions.
+
+**"I can't tell why my agent stalled for 10 seconds."** In production, context compression fires silently. The stream goes quiet. No event, no log, no metric. Users file bugs thinking the model hung. Oncall can't distinguish "compressing" from "stuck." ([#1511](https://github.com/strands-agents/harness-sdk/issues/1511), [#2804](https://github.com/strands-agents/harness-sdk/issues/2804), [#617](https://github.com/strands-agents/harness-sdk/issues/617))
+
+**"I just want to change one number."** A user wants `previewTokens: 500` instead of `750`. To get there, they must abandon `"auto"` and manually reconstruct both `SummarizingConversationManager` and `ContextOffloader` with all their defaults restated. One knob: 15 lines of boilerplate.
+
+**"My long-running agent doesn't know what's important."** A research agent accumulates 200K tokens over 50 tool calls. Static rules compress oldest-first. But the agent *knows* which results matter: it just can't say so. It watches relevant context get evicted while stale error messages stick around.
+
+**"Offloading helps exploration but hurts focused tasks."** ContextBench shows offloading gives 4-6x improvement on exploration-heavy tasks (85 cycles, many files) but *regresses* to 0.4-0.85x on focused tasks (single-file answer). The offloader can't distinguish "source code the agent is actively reasoning about" from "directory listing noise." There's no way to say "offload command output but leave file reads alone." ([#556](https://github.com/strands-agents/harness-sdk/issues/556))
+
+**"I need to offload specific tool results, not all of them."** A team building a multi-tool agent needs to aggressively truncate `bash` and `list_files` results (high-volume noise) while preserving `read_file` results intact (source code the agent is actively editing). The SDK offers no per-tool targeting, so they built a custom `SelectiveContextOffloader` wrapper that inspects tool names and applies different compression per tool. This should be a one-liner, not a custom class. ([#3267](https://github.com/strands-agents/harness-sdk/pull/3267), [internal ref 1])
+
+**"Retrieving offloaded content re-creates the overflow."** The same team disabled their retrieval tool entirely because it "returns the whole stored block at once, re-creating the overflow loop for a >window-sized result." The offloader saves context, but the tool to get it back blows context right back up, defeating the purpose for the large payloads that most need offloading. They fell back to bounded shell commands (`cat`/`grep`/`sed`/`tail`). Retrieval must be paginated. ([internal ref 2])
+
+**"I want to use Headroom but I can't plug it in."** A team evaluates a third-party context compressor that produces better summaries than the built-in one. There's no way to swap just the compression method: `"auto"` is a sealed box. They'd have to fork the SDK internals or abandon managed context entirely and build their own orchestration around the 3P API.
+
+**"After session restore, my offloaded content is gone."** A user resumes a session. The agent's stash (offloaded tool results) wasn't persisted with the session. The agent has preview stubs pointing to content that no longer exists. It hallucinates answers instead of admitting it can't retrieve the original. 
+### Why now
+
+Every story above is hitting customers *today*: in issues, in community reports, in benchmark regressions. Internal teams are building custom wrappers to work around limitations that should be solved at the SDK level.
+
+The switch exists (`contextManager: "auto"`), but the experience behind it isn't good enough. Users hit the stories above and the only escape hatch is to drop down into three disconnected components (`SummarizingConversationManager`, `ContextOffloader`, `ContextInjector`) that were designed independently, don't share state, and don't compose. We should be able to make `"auto"` genuinely great: truncation, summarization, stash, retrieval, all tuned and composable. One line that keeps getting better. And when users do need to customize, one axis at a time without restating everything else.
+
+Context strategies will change as the field changes. New models ship with different window sizes, caching behaviors, and pricing structures. New compression techniques emerge (semantic pruning, learned summarization, hybrid approaches). A framework that hardcodes its strategy into three tangled components can't keep up. A composable strategy interface means we ship new strategies, our defaults get better, and every user benefits without changing a line. The community contributes strategies too: Headroom, LLMLingua, domain-specific compressors slot in as packages, not forks. The architecture becomes a surface for the ecosystem to build on, not just for us to maintain. A strategy is just "what to compress, how, and when." Anything that answers those three questions slots in. We don't have to anticipate every future technique or how the ecosystem will change.
+
+This layout also unblocks rapid experimentation on context strategies, which is a potential differentiator for the SDK. Context management is the single highest-leverage factor in long-running agent quality: the difference between 74% and 89% task coverage isn't incremental, it's the difference between an agent that forgets what it's doing and one that doesn't. [ContextBench results](https://github.com/lizradway/notebook/blob/main/sessions/2026-07-15/head-tail-vs-pc085/ONE-PAGER.md) already show head-tail truncation with `skip_recent=5` achieving that 89% at 15-30% lower token cost. Today, each experiment like this requires a full rearchitecture of the compression path. With a composable strategy interface, testing a new approach is a config change, not a rewrite. The faster we can iterate on strategies, the faster we pull ahead of frameworks that treat context as a fixed-size sliding window.
+
+### Success Criteria
+
+1. **One-liner preserved**: `contextManager: "auto"` still works, same defaults. 80% of users never see the class.
+2. **No cliff**: override one axis (threshold, method, storage, content-type rules) without restating everything else.
+3. **Pluggable**: 3P compressors (Headroom, enterprise) slot in with ~15 lines. No adapter layer.
+4. **One mental model**: compression, injection, eviction, and protection are one system with strategies that fire at any point in the loop.
+5. **Recoverable**: content is stashed, not destroyed. The agent retrieves on demand. Lossy compression is a choice.
+6. **Agent autonomy**: the agent pins, offloads, and searches its own context. Static rules are a safety net, not the only path.
+7. **Observable**: stream events, OTEL spans, lifecycle hooks. Operators know when compression fires and what it freed.
+8. **Session-durable**: stash content persists with the session and survives restore.
+
+---
+
+## 2. Core Model
+
+The ContextManager operates across two layers: **L0** (the context window the model sees) and **L1** (the stash, session-scoped recoverable storage). It does two things:
+
+1. **Offload**: take content OUT of what the model sees (L0 → L1)
+2. **Inject**: put content IN to what the model sees (→ L0)
+
+Both follow the same mental model: **choose a representation for content in the context window**. Offload replaces expensive representations with cheaper ones. Inject adds cheaper representations of expensive external content. The API shape is unified:
+
+```
+{Direction}.{representation}({target/source}, {options?})
+```
+
+### 2.1 Strategy API
+
+```typescript
+import {
+  ContextManager, Offload, Inject,
+  BeforeModelCallEvent,
+} from '@strands-agents/sdk'
+
+new ContextManager({
+  strategies: [
+    // Offload: replace expensive content with cheaper representations
+    Offload.truncate("toolResults", { previewTokens: 750 })
+      .when({ threshold: 1500, skipRecent: 3 }),
+    Offload.truncate(["bash", "list_files"], { previewTokens: 200 })
+      .when({ threshold: 500 }),
+    Offload.summarize({ ratio: 0.3 })
+      .when(BeforeModelCallEvent, { utilization: 0.85 }),
+    Offload("toolResultErrors"),                    // move entirely to L1
+
+    // Inject: fires before each model call
+    Inject.skeleton("stash"),                       // stash manifest as refs/metadata
+    Inject("clock"),                                // bare call = inject as-is (no transformation)
+    Inject("budget"),
+  ],
+})
+```
+
+### 2.2 Representations
+
+A representation is how content is mutated to appear in the context window. Representations work in both directions:
+
+| Representation | What it produces | Offload use | Inject use |
+|----------------|-----------------|-------------|------------|
+| `truncate` | Partial content (head/tail/head-tail) | Keep a preview, stash the full thing | Inject a truncated view of external content |
+| `summarize` | LLM-generated summary | Condense old conversation | Inject a summary of external content |
+| `skeleton` | Structural outline (signatures, refs) | Code: just signatures | Stash: just refs + metadata |
+
+Truncation modes:
+- `"head"`: keep the first N tokens
+- `"tail"`: keep the last N tokens
+- `"head-tail"` (default): keep both ends, drop the middle
+
+Summarization modes:
+- `"cache-aligned"` (default): reuses the warm prompt cache rather than re-ingesting at full uncached cost
+- `"coding"`: vended prompt that preserves file paths, function signatures, error messages, decisions
+- `"research"`: vended prompt that preserves sources, claims, evidence, contradictions
+- custom string: pass a full prompt for domain-specific summarization; see **Appendix D**
+
+Skeleton modes (P2):
+- `"signatures"` (default): keeps function signatures, class names, imports
+- `"outline"`: keeps structural hierarchy (headings, keys, top-level declarations)
+
+See **Appendix D** for additional representations for the future.
+
+### 2.3 Offload Targets (content types)
+
+The target (first arg to `Offload.*`) is `string | string[]` and describes **what** to offload:
+
+| Target | Matches |
+|--------|---------|
+| `"toolResults"` | All successful tool result messages |
+| `"toolResultErrors"` | All failed tool result messages |
+| `"assistantMessages"` | Assistant messages (text responses) |
+| `"userMessages"` | User messages (no tool result) |
+| `"images"` | Messages containing image blocks |
+| `"documents"` | Messages containing document blocks |
+| `"code"` | Messages heuristically detected as code |
+| `["bash", "list_files"]` | Tool results from these specific tools only |
+| `["!read_file"]` | Tool results from all tools *except* `read_file` |
+| _(omitted)_ | Everything not matched by a more specific strategy |
+
+Passing an array of tool names implicitly scopes to tool results from those tools. Prefix a name with `!` to exclude: `["!read_file", "!write_file"]` means "all tool results except these." Don't mix includes and excludes in the same array.
+
+```typescript
+// Aggressive truncation on noisy tools
+Offload.truncate(["bash", "list_files"], { previewTokens: 200 })
+  .when({ threshold: 500 })
+
+// Offload everything except the tools the agent is actively using
+Offload.truncate(["!read_file", "!write_file"], { previewTokens: 750 })
+  .when({ threshold: 1500 })
+```
+
+Omit the target to create a catch-all fallback. Multiple strategies can match the same message: they compose, not compete.
+
+### 2.4 Inject Sources
+
+What sources `Inject` can pull from, limited to what the ContextManager **owns**:
+
+| Source | What it provides | Typical representation |
+|--------|-----------------|----------------------|
+| `"stash"` | Offloaded content manifest (see §3) | `skeleton`: refs + metadata |
+| `"budget"` | Token utilization stats | bare: as-is |
+| `"clock"` | Current datetime | bare: as-is |
+| custom | User-defined render function | any |
+
+**What about memory, RAG, state?** Those are injected by their own subsystems (MemoryManager, retriever plugins, etc.), not by the ContextManager. Each subsystem owns its own injection lifecycle.
+
+See **Appendix D** for future condition fields and events that may extend inject sources.
+
+### 2.5 Triggers (`.when()`)
+
+**Offload** strategies use `.when(event?, conditions?)` to declare when they fire.
+
+Without an event param, a strategy evaluates on every lifecycle event and fires whenever its target matches and conditions are met. Passing an event restricts it to only that hook (e.g. `BeforeModelCallEvent` for summarization, so it doesn't fire after every tool call).
+
+```typescript
+// Event implicit: tool results only arrive after tool calls
+Offload.truncate("toolResults", { previewTokens: 750 })
+  .when({ threshold: 1500, skipRecent: 3 })
+
+Offload("toolResultErrors")  // no .when() needed: fires on every tool error
+
+// Event required: summarization is global, timing matters
+Offload.summarize({ ratio: 0.3 })
+  .when(BeforeModelCallEvent, { utilization: 0.85 })
+```
+
+**Inject** strategies fire before each model call: no `.when()` needed. They are ephemeral (stripped and re-injected fresh each time).
+
+**Lifecycle events** (from `@strands-agents/sdk`):
+
+| Event | Fires when | Typical use |
+|-------|-----------|-------------|
+| `AfterToolCallEvent` | After each tool execution completes | Per-message offload (truncate oversized results immediately) |
+| `BeforeModelCallEvent` | Before each LLM call in the loop | Global offload (summarize when utilization is high) |
+| `AfterInvocationEvent` | At the end of each `agent.invoke()` | Post-turn cleanup offload |
+| `AfterModelCallEvent` | After model responds (internal) | Overflow handler: if `ContextWindowOverflowError`, run all offload strategies regardless of conditions |
+
+**Condition fields** (second arg to `.when()`, optional):
+
+| Condition | Type | Meaning |
+|-----------|------|---------|
+| `threshold` | `number` (absolute tokens) | Only fire if the target message exceeds this size |
+| `utilization` | `number` (0-1 ratio) | Only fire if global context utilization exceeds this. Use `1.0` to mean "only on overflow" (reactive-only: strategy is skipped during proactive passes but fires when `ContextWindowOverflowError` triggers emergency compression). |
+| `skipRecent` | `number` | Skip the N most recent messages of this type |
+
+Omitting conditions means the strategy fires every time the event occurs for a matching target.
+
+Messages with `metadata.custom.pinned = true` (from PR #2644) are never candidates for offload regardless of which strategy fires. The system prompt is always implicitly pinned.
+
+All of these extension points (lifecycle events, strategies, representations, conditions) are open: 3P packages and users can register custom instances of each. The built-in set is a starting point, not a ceiling.
+
+### 2.6 Bare calls
+
+When `Offload` or `Inject` is called without a representation method:
+
+- **`Offload("toolResults")`**: move entirely to L1 (stash). Nothing remains in L0.
+- **`Inject("clock")`**: inject the source as-is with no transformation.
+
+The full offload spectrum:
+
+| Call | L0 result | L1 result |
+|------|-----------|-----------|
+| `Offload.truncate(target)` | Preview (head/tail) | Full original |
+| `Offload.summarize(target)` | LLM summary | Full original |
+| `Offload(target)` | Nothing | Full original |
+
+The content is always recoverable via the stash manifest and `get_context` (which returns bounded pages, not the full block; see §3 and §5.3).
+
+---
+
+## 3. The Stash (L1)
+
+A mutable, ContextManager-owned store for recoverable content. When an offload strategy fires, the ContextManager writes the original to the stash before transforming L0.
+
+```
+offload fires: write original to stash, compress L0, leave preview + ref in L0
+```
+
+The stash manifest (injected via `Inject.skeleton("stash")`) tells the agent what's available:
+
+```xml
+<stash count="3" budget="82%">
+  <entry ref="msg-a1b2c3d4" type="tool_result" tool="read_file" tokens="2400" age="4 turns" />
+  <entry ref="msg-e5f6g7h8" type="tool_result" tool="fetch" tokens="1800" age="7 turns" />
+  <entry ref="msg-i9j0k1l2" type="summary" tokens="900" age="12 turns" />
+</stash>
+```
+
+The agent sees what's stashed and can retrieve full content on demand via `get_context` / `search_context` tools. This two-step pattern avoids speculative tool calls.
+
+**Key paths**:
+
+| Key pattern | Contents |
+|-------------|----------|
+| `context/<sessionId>/<tracking_id>` | Full original content of an offloaded message |
+| `context/<sessionId>/_manifest` | Stash index (entry metadata, token counts, ages) |
+
+**Message refs** use the durable `tracking_id` UUID on each Message ([PR #2836](https://github.com/strands-agents/harness-sdk/pull/2836)). This ID survives snapshot/restore and never shifts when messages are offloaded.
+
+### 3.1 Enabling/disabling
+
+- **Default**: enabled when any offload strategy is configured.
+- **`stash: false`**: disable (offload strategies still fire but originals are destroyed, same as today's behavior)
+- **`stash: { maxEntries: 50 }`**: enable with a cap
+
+### 3.2 Eviction (P2)
+
+The stash is not an audit log: it's mutable. Once content is removed, it's gone unless something else recorded it first.
+
+For v1, the stash grows until the session ends or storage is cleaned up externally. Eviction policies are a follow-up once we have data on real-world stash sizes. See §7 (Future Work).
+
+### 3.3 Relationship to MemoryManager
+
+ContextManager owns L0 (context window) and L1 (stash). MemoryManager owns L2 (cross-session knowledge) and can read from the stash for extraction. They share the token budget but don't reference each other.
+
+---
+
+## 4. Developer Experience
+
+### 4.1 Equivalent of today's SummarizingConversationManager
+
+What users do today:
+```python
+agent = Agent(
+    conversation_manager=SummarizingConversationManager(
+        summary_ratio=0.3,
+        preserve_recent_messages=10,
+        proactive_compression={"compression_threshold": 0.7},
+    )
+)
+```
+
+Same behavior on ContextManager:
+```python
+agent = Agent(
+    context_manager=ContextManager(
+        strategies=[
+            Offload.summarize(ratio=0.3, preserve_recent=10)
+                .when(BeforeModelCallEvent, utilization=0.7),
+        ],
+        stash=False,  # no stash: destructive, same as today (no storage needed)
+    )
+)
+```
+
+Setting `stash=False` makes offloading destructive (originals lost after summarization), matching today's behavior. No `storage` param is needed since there's nothing to persist.
+
+### 4.2 Recreating "auto"
+
+```python
+agent = Agent(context_manager="auto")
+```
+
+Resolves to:
+```python
+ContextManager(
+    strategies=[
+        Offload.truncate("toolResults", preview_tokens=750)
+            .when(threshold=1500, skip_recent=3),
+        Offload.summarize(ratio=0.3)
+            .when(BeforeModelCallEvent, utilization=0.85),
+        Inject.skeleton("stash"),
+    ],
+)
+# + read-only tools: get_context, search_context
+```
+
+Key differences from today's `SummarizingConversationManager`:
+- Tool results are truncated individually (not summarized as a block)
+- Originals are stashed and retrievable via `get_context`
+- The agent sees a stash manifest so it knows what's available
+- Content-type aware: tool results vs. conversation get different treatment
+
+### 4.3 Custom: optimal for a coding agent (hypothesized)
+
+> **Note**: [ContextBench results](https://github.com/lizradway/notebook/blob/main/sessions/2026-07-15/head-tail-vs-pc085/ONE-PAGER.md) show head-tail truncation with `skip_recent=5` achieving 89% mean coverage vs 74% for the current `"auto"` baseline, at 15-30% lower token cost (Opus 4.6, hard suite, n=1 per task). This config builds on that foundation with additional strategies.
+
+```python
+agent = Agent(
+    context_manager=ContextManager(
+        storage=storage,
+        strategies=[
+            # Aggressively truncate noisy tools: just keep first/last lines
+            Offload.truncate(["bash", "list_files", "grep"], preview_tokens=200)
+                .when(threshold=500),
+
+            # Keep read_file results longer, truncate only when large
+            Offload.truncate(["!bash", "!list_files", "!grep"], preview_tokens=750)
+                .when(threshold=2000, skip_recent=5),
+
+            # Evict error results entirely (stash them but no L0 presence)
+            Offload("toolResultErrors"),
+
+            # Summarize conversation when utilization gets high
+            Offload.summarize(ratio=0.3)
+                .when(BeforeModelCallEvent, utilization=0.8),
+
+            # Inject stash manifest + budget so agent can self-manage
+            Inject.skeleton("stash"),
+            Inject("budget"),
+        ],
+    )
+)
+```
+
+This gives a coding agent:
+- Bounded noise from shell commands (200 token preview)
+- Longer retention of source code it's actively reading (5 most recent kept intact)
+- Error results cleaned up immediately (but recoverable from stash)
+- Proactive summarization of older conversation at 80% utilization
+- Awareness of its own context budget and what's in the stash
+
+See **Appendix A** for full examples including 3P strategies, custom strategies, and TypeScript equivalents.
+
+---
+
+## 5. ContextManager Class
+
+### 5.1 Constructor
+
+```typescript
+export class ContextManager implements Plugin {
+  readonly name = 'strands:context-manager'
+
+  constructor(config: {
+    storage?: Storage                 // falls back to Agent's storage, then in-memory
+    strategies?: Strategy[]           // offload + inject strategies
+    stash?: false | {                 // false = disable stash (destructive offload, no storage needed)
+      maxEntries?: number
+    }
+  })
+}
+```
+
+### 5.2 Modes
+
+The `contextManager` parameter accepts a string preset or a `ContextManager` instance:
+
+```typescript
+const agent = new Agent({ contextManager: "auto" })
+const agent = new Agent({ contextManager: "minimal" })
+const agent = new Agent({ contextManager: new ContextManager({ ... }) })
+```
+
+| Mode | Strategy behavior | Best for |
+|------|-------------------|----------|
+| `"auto"` | Balanced: truncate tool results, summarize at 85% utilization | Most use cases: chat, Q&A, single-tool agents |
+| `"minimal"` | Reactive only: compress on overflow | Short tasks, small context |
+
+Future mode presets are planned for P2 (see §7.2).
+
+
+See §4.2 for what `"auto"` resolves to. See §4.3 for a full custom example.
+
+### 5.3 Agent Tools (via `.as_tool()`)
+
+Whether the agent gets context management tools is orthogonal to the mode: it's determined by `tools=[]`, not by the ContextManager config. This follows the [`.as_tool()` design](https://gist.github.com/lizradway/82dea3e7832c2d336595d77a8f9e42f1): plugins own lifecycle, tools are always explicit.
+
+```python
+# Rules only: no agent tools (most users)
+agent = Agent(context_manager=ContextManager("auto"))
+
+# Agent manages its own context: explicit opt-in
+cm = ContextManager("auto")
+agent = Agent(context_manager=cm, tools=[cm.as_tool()])
+```
+
+**What `.as_tool()` exposes:**
+- **`offload_context`**: compress content (truncate, summarize, skeleton). Stashes original.
+- **`get_context`**: retrieve full content of a stashed entry (bounded, paginated).
+- **`search_context`**: search across the context window and stash.
+- **`pin_context`**: mark as never-compress.
+- **`unpin_context`**: remove pin, allow normal rules to apply.
+- **`context_status`**: returns budget, message inventory, stash overview.
+
+See **Appendix H** for full tool schemas and parameters.
+
+### 5.4 Public API on the instance
+
+```typescript
+class ContextManager {
+  get budget(): TokenBudget
+
+  // offload
+  offload(refs: string[], method?: string, options?: object): Promise<void>
+  pin(ref: string): void
+  unpin(ref: string): void
+
+  // stash
+  get(ref: string, options?: { offset?: number, limit?: number, grep?: string }): Promise<string>
+  search(options?: { query?: string, tool?: string, type?: string, scope?: string, limit?: number }): Promise<StashEntry[]>
+
+  // inject
+  inject(source: string, content: string): void
+}
+```
+
+### 5.5 Token Budget
+
+The `TokenBudget` is a read-only view the ContextManager exposes for "how full is the context window":
+
+```typescript
+type TokenBudget = {
+  limit: number       // model's context window size
+  used: number        // current token usage
+  remaining: number   // limit - used
+  ratio: number       // used / limit (0-1)
+}
+```
+
+**Where the data comes from**: `limit` reads from `agent.model.context_window_limit` (already exists on the model config). `used` reads from `BeforeModelCallEvent.projected_input_tokens` (already computed by the existing proactive compression hook). The ContextManager wraps the existing infrastructure into a typed object that strategies, tools, and 3P code can consume uniformly.
+
+
+---
+
+## 6. Telemetry & Observability
+
+Context management today is silent: compression fires, the stream goes quiet, and callers have no signal that anything happened (#1511, #2804). The ContextManager makes operations observable through two channels:
+
+### 6.1 Stream Events
+
+When any strategy fires, the ContextManager yields a stream event so streaming consumers get real-time feedback:
+
+```typescript
+type ContextManagementStreamEvent = {
+  contextManagement: {
+    phase: "started" | "completed"
+    reason: "proactive" | "overflow" | "agent-requested"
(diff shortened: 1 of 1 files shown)


team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+```typescript
+import {
+  ContextManager, Offload, Inject,
+  BeforeModelCallEvent,
+} from '@strands-agents/sdk'
+
+new ContextManager({
+  strategies: [
+    // Offload: replace expensive content with cheaper representations
+    Offload.truncate("toolResults", { previewTokens: 750 })
```
mkmeral, line 81: Are these running conversation managers under the hood? or are these like separate?

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+**Scope**: TypeScript + Python SDK  
+**Related**: [Context Strategy Design (v1 facade)](https://github.com/strands-agents/docs/pull/831), [MemoryManager Design](https://github.com/strands-agents/docs/pull/844)
+
+> **Note**: API shapes in this document are drafted for illustration purposes and are likely to change through bar raising and review.
+
+---
+
+## 1. Problem
+
+### What customers hit
```
notowen333, line 14: This is a very helpful style for an intro into a doc 💯

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+
+**"After session restore, my offloaded content is gone."** A user resumes a session. The agent's stash (offloaded tool results) wasn't persisted with the session. The agent has preview stubs pointing to content that no longer exists. It hallucinates answers instead of admitting it can't retrieve the original. 
+### Why now
+
+Every story above is hitting customers *today*: in issues, in community reports, in benchmark regressions. Internal teams are building custom wrappers to work around limitations that should be solved at the SDK level.
+
+The switch exists (`contextManager: "auto"`), but the experience behind it isn't good enough. Users hit the stories above and the only escape hatch is to drop down into three disconnected components (`SummarizingConversationManager`, `ContextOffloader`, `ContextInjector`) that were designed independently, don't share state, and don't compose. We should be able to make `"auto"` genuinely great: truncation, summarization, stash, retrieval, all tuned and composable. One line that keeps getting better. And when users do need to customize, one axis at a time without restating everything else.
```
mehtarac, line 37: SlidingWindowConversationManager too?

team/designs/0015-context-manager.md
lizradway: yay thanks!

team/designs/0015-context-manager.md
lizradway: yeah, can add!

team/designs/0015-context-manager.md
lizradway: separate, conversation manager is dead!

lizradway: essentially just a compression function that is defined and can be applied to anything (ideally)

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+---
+
+## 2. Core Model
+
+The ContextManager operates across two layers: **L0** (the context window the model sees) and **L1** (the stash, session-scoped recoverable storage). It does two things:
+
+1. **Offload**: take content OUT of what the model sees (L0 → L1)
+2. **Inject**: put content IN to what the model sees (→ L0)
+
+Both follow the same mental model: **choose a representation for content in the context window**. Offload replaces expensive representations with cheaper ones. Inject adds cheaper representations of expensive external content. The API shape is unified:
```
mehtarac: > Inject adds cheaper representations of expensive external content.

So this would be summaries of the context?

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+    Inject("budget"),
+  ],
+})
+```
+
+### 2.2 Representations
+
+A representation is how content is mutated to appear in the context window. Representations work in both directions:
+
+| Representation | What it produces | Offload use | Inject use |
```
Unshure, line 101: Where did you get these representations? Are there others that you considered?

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+|----------------|-----------------|-------------|------------|
+| `truncate` | Partial content (head/tail/head-tail) | Keep a preview, stash the full thing | Inject a truncated view of external content |
+| `summarize` | LLM-generated summary | Condense old conversation | Inject a summary of external content |
+| `skeleton` | Structural outline (signatures, refs) | Code: just signatures | Stash: just refs + metadata |
```
opieter-aws, line 105: Is `references` a clearer term for this maybe?

team/designs/0015-context-manager.md
lizradway: can be summaries, but often are not. 

any cheaper representation of the information stored in the context window can be supported.

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+
+```typescript
+import {
+  ContextManager, Offload, Inject,
+  BeforeModelCallEvent,
+} from '@strands-agents/sdk'
+
+new ContextManager({
+  strategies: [
+    // Offload: replace expensive content with cheaper representations
```
opieter-aws: What about `Drop` if I do not want something in my L1 at all because it may write to my long term memory? Like ToolResult or ToolResultErrors.

Is there an option to just summarize and drop?

team/designs/0015-context-manager.md
lizradway: built in strategies to start are summarize, truncate (sliding window basically), and skeleton (AST). 

but built to be extensible for any future strategies and to allow community supported strategies.

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+- `"tail"`: keep the last N tokens
+- `"head-tail"` (default): keep both ends, drop the middle
+
+Summarization modes:
+- `"cache-aligned"` (default): reuses the warm prompt cache rather than re-ingesting at full uncached cost
+- `"coding"`: vended prompt that preserves file paths, function signatures, error messages, decisions
+- `"research"`: vended prompt that preserves sources, claims, evidence, contradictions
+- custom string: pass a full prompt for domain-specific summarization; see **Appendix D**
+
+Skeleton modes (P2):
```
notowen333, line 118: These look very file format specific. Is it generic?

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+| Representation | What it produces | Offload use | Inject use |
+|----------------|-----------------|-------------|------------|
+| `truncate` | Partial content (head/tail/head-tail) | Keep a preview, stash the full thing | Inject a truncated view of external content |
+| `summarize` | LLM-generated summary | Condense old conversation | Inject a summary of external content |
+| `skeleton` | Structural outline (signatures, refs) | Code: just signatures | Stash: just refs + metadata |
+
+Truncation modes:
+- `"head"`: keep the first N tokens
+- `"tail"`: keep the last N tokens
+- `"head-tail"` (default): keep both ends, drop the middle
```
JackYPCOnline, line 110: Wondering is this widely used ?

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+
+See **Appendix D** for additional representations for the future.
+
+### 2.3 Offload Targets (content types)
+
+The target (first arg to `Offload.*`) is `string | string[]` and describes **what** to offload:
+
+| Target | Matches |
+|--------|---------|
+| `"toolResults"` | All successful tool result messages |
```
poshinchen, line 130: should we make these as enums?

team/designs/0015-context-manager.md
mkmeral, line 14: Are all of these issues today? Haven't we fixed a bunch of these?

team/designs/0015-context-manager.md
lizradway: i have thought about that and `Drop` was included in some initial implementations of this design (originally dropped ToolResultErrors).

my final thoughts on it + why it is left off this doc now is bc I think that could be a configuration on the method (ie. `stash: false`). i'd also want to add this as a follow up depending on how much of an ask we have for this/use case.

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+
+### Success Criteria
+
+1. **One-liner preserved**: `contextManager: "auto"` still works, same defaults. 80% of users never see the class.
+2. **No cliff**: override one axis (threshold, method, storage, content-type rules) without restating everything else.
+3. **Pluggable**: 3P compressors (Headroom, enterprise) slot in with ~15 lines. No adapter layer.
```
mkmeral, line 47: How do they integrate? I'm not sure if I get that from the doc

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+
+| Target | Matches |
+|--------|---------|
+| `"toolResults"` | All successful tool result messages |
+| `"toolResultErrors"` | All failed tool result messages |
+| `"assistantMessages"` | Assistant messages (text responses) |
+| `"userMessages"` | User messages (no tool result) |
+| `"images"` | Messages containing image blocks |
+| `"documents"` | Messages containing document blocks |
+| `"code"` | Messages heuristically detected as code |
```
notowen333, line 136: Is it worth it to mix 1:1 mappings of model shaped content types with heuristics like this?

Might we start with standard model/framework types and then give an interface for users to do dynamic detection.

Maintaining our own detectors seems like a large can of worms

team/designs/0015-context-manager.md
lizradway: `summarize` and `truncate` are what we support now in Summarizing Conversation Manager and Sliding Window Conversation Manager, so I wanted to support those for parity.

Skeleton is essentially AST which is what is proving most effective nowadays on the frontier, so i wanted to include this in the initial P1/P2s.

lizradway: Future methods to implement/consider in Appendix. Also accepting community methods/strategies.

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+| Constructor & plugin lifecycle | **P0** | 3 days | (none) | Config parsing, validation, hook registration, `InitializedEvent` wiring, `storage` resolution, `NullConversationManager` swap when `contextManager` is set, deprecation warnings for `conversationManager`/`ContextOffloader` co-use |
+| Strategy execution engine | **P0** | 1.5 weeks | Constructor | Event subscription, target matching (content types + tool name arrays + `!` exclusions), condition evaluation, multi-strategy composition on same message, idempotency |
+| `Offload.truncate` | **P0** | 3 days | Strategy engine | All three modes (`"head"`, `"tail"`, `"head-tail"`), per-tool targeting, preview token slicing |
+| `Offload.summarize` | **P0** | 4 days | Strategy engine | `"cache-aligned"` mode (default), custom prompt string support, wrap existing `ConversationManager` summarization in the strategy interface, expose `ratio`/`preserve_recent` params |
+| Inject system | **P1** | 4 days | Strategy engine | Middleware that strips + re-injects before each model call. Built-in sources: `Inject.skeleton("stash")`, `Inject("budget")`, `Inject("clock")`. Custom `render` function support. |
```
mkmeral, line 550: won't this keep invalidating cache?

team/designs/0015-context-manager.md
lizradway: sure, maybe even `structure`. open to naming changes, havent thought abt it much

team/designs/0015-context-manager.md
lizradway: these all still are open/exist

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+
+Whether the agent gets context management tools is orthogonal to the mode: it's determined by `tools=[]`, not by the ContextManager config. This follows the [`.as_tool()` design](https://gist.github.com/lizradway/82dea3e7832c2d336595d77a8f9e42f1): plugins own lifecycle, tools are always explicit.
+
+```python
+# Rules only: no agent tools (most users)
+agent = Agent(context_manager=ContextManager("auto"))
+
+# Agent manages its own context: explicit opt-in
+cm = ContextManager("auto")
+agent = Agent(context_manager=cm, tools=[cm.as_tool()])
```
mkmeral, line 425: is this a better devx then `context_manager="agentic"`? 🤔

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+| `AfterToolCallEvent` | After each tool execution completes | Per-message offload (truncate oversized results immediately) |
+| `BeforeModelCallEvent` | Before each LLM call in the loop | Global offload (summarize when utilization is high) |
+| `AfterInvocationEvent` | At the end of each `agent.invoke()` | Post-turn cleanup offload |
+| `AfterModelCallEvent` | After model responds (internal) | Overflow handler: if `ContextWindowOverflowError`, run all offload strategies regardless of conditions |
+
+**Condition fields** (second arg to `.when()`, optional):
+
+| Condition | Type | Meaning |
+|-----------|------|---------|
+| `threshold` | `number` (absolute tokens) | Only fire if the target message exceeds this size |
```
opieter-aws, line 194: Can we make this more extensible? E.g. what if I want to offload at a certain amount of turns or if e.g. identified by the model to be insignificant (integration with an agentic mode or flag)

team/designs/0015-context-manager.md
lizradway: See Appendix A part 4/5

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+```python
+# Rules only: no agent tools (most users)
+agent = Agent(context_manager=ContextManager("auto"))
+
+# Agent manages its own context: explicit opt-in
+cm = ContextManager("auto")
+agent = Agent(context_manager=cm, tools=[cm.as_tool()])
+```
+
+**What `.as_tool()` exposes:**
```
mkmeral, line 428: are all of these one tool each? Do we need all of them to be separate? We should combine some of these probably

team/designs/0015-context-manager.md
lizradway: this was the best performer on my initial testing.

has some traction in the community afaik.

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+| Inject system | **P1** | 4 days | Strategy engine | Middleware that strips + re-injects before each model call. Built-in sources: `Inject.skeleton("stash")`, `Inject("budget")`, `Inject("clock")`. Custom `render` function support. |
+| TokenBudget | **P1** | 2 days | Constructor | Read `model.context_window_limit` + `projected_input_tokens`, expose as typed `TokenBudget` object, wire to condition evaluation and inject render context |
+| Stash (L1) | **P1** | 1.5 weeks | Strategy engine, Inject | Write-on-offload, manifest generation, `get_context` tool (bounded retrieval with offset/limit/grep), `search_context` tool (unified L0+L1 search with query/tool/type/scope filters), in-memory fallback when no storage configured |
+| Agentic mode (`.as_tool()`) | **P1** | 3 days | Stash, `.as_tool()` primitive | `offload_context`, `pin_context`, `unpin_context`, `context_status` per [`.as_tool()` design](https://gist.github.com/lizradway/82dea3e7832c2d336595d77a8f9e42f1). Depends on `.as_tool()` being implemented on the Plugin interface. |
```
mkmeral, line 553: why is this p1? can't we reuse agentic mode here?

team/designs/0015-context-manager.md
mehtarac, line 130: What about toolResults that have images, documents, etc. does that get offladed under generic targets?

team/designs/0015-context-manager.md
lizradway: its typically used for structured items (ie. code, json, etc).


id imagine people would use this when offloading _specific_ tool results, ie. read file or code tools.

team/designs/0015-context-manager.md
lizradway: makes sense to me.

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+| Event | Fires when | Typical use |
+|-------|-----------|-------------|
+| `AfterToolCallEvent` | After each tool execution completes | Per-message offload (truncate oversized results immediately) |
+| `BeforeModelCallEvent` | Before each LLM call in the loop | Global offload (summarize when utilization is high) |
+| `AfterInvocationEvent` | At the end of each `agent.invoke()` | Post-turn cleanup offload |
+| `AfterModelCallEvent` | After model responds (internal) | Overflow handler: if `ContextWindowOverflowError`, run all offload strategies regardless of conditions |
+
+**Condition fields** (second arg to `.when()`, optional):
+
+| Condition | Type | Meaning |
```
JackYPCOnline, line 192: Does it has to be defined by SDK, or user should do that, eg: pass in some callback functions.

team/designs/0015-context-manager.md
opieter-aws: Main reason I'm calling this out is because what if my L1 is not free? If I'm using some managed service for L1 offloading (e.g. AgentCore), I'm paying for every API call / storage

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+    )
+)
+```
+
+Setting `stash=False` makes offloading destructive (originals lost after summarization), matching today's behavior. No `storage` param is needed since there's nothing to persist.
+
+### 4.2 Recreating "auto"
+
+```python
+agent = Agent(context_manager="auto")
```
poshinchen, line 311: I personally prefer `agent = Agent(context_manager=ContextManager())`?

team/designs/0015-context-manager.md
lizradway: yeah, id like to start with with the obvious sdk-defined types (ie. assistant messages, user messages), and then as P1/P2 do more heuristics, (ie. detecting document from document type, code from .TS/.PY etc)

team/designs/0015-context-manager.md
liramon2, line 110: Will it be possible to configure the number of tokens on each end to be different (ex. 100 token head and 200 token tail)? It seems similar to the existing `pinFirst` + `preserveRecentMessages` in conversation manager

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+
+A mutable, ContextManager-owned store for recoverable content. When an offload strategy fires, the ContextManager writes the original to the stash before transforming L0.
+
+```
+offload fires: write original to stash, compress L0, leave preview + ref in L0
+```
+
+The stash manifest (injected via `Inject.skeleton("stash")`) tells the agent what's available:
+
+```xml
```
notowen333, line 234: would the agent not want a little of the semantic content of each entry to remember what is there?

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+<stash count="3" budget="82%">
```
opieter-aws, line 235: Is there no preview in this example?

team/designs/0015-context-manager.md
mkmeral: How does this work with this definition though? it's not extensible

```typescript
type Strategy = OffloadStrategy | InjectStrategy
```

team/designs/0015-context-manager.md
lizradway: very interested in extending to callback function. would be a good p1/p2, they do that in this pr already for ContextOffloader: https://github.com/strands-agents/harness-sdk/pull/3267

team/designs/0015-context-manager.md
mkmeral: this is a limited set of strategies. is there a strategy interface?

team/designs/0015-context-manager.md
lizradway: would also like to include **kwargs to extend

team/designs/0015-context-manager.md
mkmeral: "conversation manager is dead!" long live the context manager! yeey

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+
+### 3.2 Eviction (P2)
+
+The stash is not an audit log: it's mutable. Once content is removed, it's gone unless something else recorded it first.
+
+For v1, the stash grows until the session ends or storage is cleaned up externally. Eviction policies are a follow-up once we have data on real-world stash sizes. See §7 (Future Work).
+
+### 3.3 Relationship to MemoryManager
+
+ContextManager owns L0 (context window) and L1 (stash). MemoryManager owns L2 (cross-session knowledge) and can read from the stash for extraction. They share the token budget but don't reference each other.
```
opieter-aws: We should add trigger strategies from writing L1->L2

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+|------|-----------|-----------|
+| `Offload.truncate(target)` | Preview (head/tail) | Full original |
+| `Offload.summarize(target)` | LLM summary | Full original |
+| `Offload(target)` | Nothing | Full original |
+
+The content is always recoverable via the stash manifest and `get_context` (which returns bounded pages, not the full block; see §3 and §5.3).
+
+---
+
+## 3. The Stash (L1)
```
opieter-aws, line 223: What is the store back-end here? Can I plug in my own?

team/designs/0015-context-manager.md
lizradway: skipRecent is supposed to fufill turn based detection. 

priority based/relevance based offloading is P2/P3. would love to do as follow up.

team/designs/0015-context-manager.md
lizradway: how would this extend to other modes? ie. `minimal`

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+What sources `Inject` can pull from, limited to what the ContextManager **owns**:
+
+| Source | What it provides | Typical representation |
+|--------|-----------------|----------------------|
+| `"stash"` | Offloaded content manifest (see §3) | `skeleton`: refs + metadata |
+| `"budget"` | Token utilization stats | bare: as-is |
+| `"clock"` | Current datetime | bare: as-is |
+| custom | User-defined render function | any |
+
+**What about memory, RAG, state?** Those are injected by their own subsystems (MemoryManager, retriever plugins, etc.), not by the ContextManager. Each subsystem owns its own injection lifecycle.
```
gautamsirdeshmukh, line 166: > Those are injected by their own subsystems (MemoryManager, retriever plugins, etc.), not by the ContextManager. Each subsystem owns its own injection lifecycle.

Is this sentence referring to the fact that those plugins register their own processes/behavior tied to hook events? 

Regardless, I think it'd be very useful for developers to be able to offload to / inject from external sources through the ContextManager itself - for instance, setting up an `Offload.summarize` rule that stores some embedded summary in a vector DB / knowledge base

team/designs/0015-context-manager.md
opieter-aws, line 223: Why doesn't L1 survive session restore? This seems like a big gap if I offload a lot and frequently restart sessions.

team/designs/0015-context-manager.md
lizradway: we can keep around agentic mode as is for now, but i don't want to wrap a bunch of tools if we are going to implement over it in the future and have decided we want a different devX for this experience.

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+                .when(BeforeModelCallEvent, utilization=0.7),
+        ],
+        stash=False,  # no stash: destructive, same as today (no storage needed)
+    )
+)
+```
+
+Setting `stash=False` makes offloading destructive (originals lost after summarization), matching today's behavior. No `storage` param is needed since there's nothing to persist.
+
+### 4.2 Recreating "auto"
```
mkmeral, line 308: what about recreating "agentic"? :)

team/designs/0015-context-manager.md
lizradway: im pretty sure context injector right now injects at the end as a message (rather than sys prompt) for this very reason (as one of the reasons) cc: @opieter-aws  to check me?

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+| `"assistantMessages"` | Assistant messages (text responses) |
+| `"userMessages"` | User messages (no tool result) |
+| `"images"` | Messages containing image blocks |
+| `"documents"` | Messages containing document blocks |
+| `"code"` | Messages heuristically detected as code |
+| `["bash", "list_files"]` | Tool results from these specific tools only |
+| `["!read_file"]` | Tool results from all tools *except* `read_file` |
+| _(omitted)_ | Everything not matched by a more specific strategy |
+
+Passing an array of tool names implicitly scopes to tool results from those tools. Prefix a name with `!` to exclude: `["!read_file", "!write_file"]` means "all tool results except these." Don't mix includes and excludes in the same array.
```
liramon2, line 141: If includes and excludes can't be in the same array, why not separate them into different parameters?

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
+# Design: ContextManager Class
```
mkmeral, line 1: is this improvement from the current state worth 11-12 weeks of our time? where can we cut, speed up?

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+
+Same behavior on ContextManager:
+```python
+agent = Agent(
+    context_manager=ContextManager(
+        strategies=[
+            Offload.summarize(ratio=0.3, preserve_recent=10)
+                .when(BeforeModelCallEvent, utilization=0.7),
+        ],
+        stash=False,  # no stash: destructive, same as today (no storage needed)
```
opieter-aws, line 301: Can I not choose to stash per strategy?

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+
+```
+offload fires: write original to stash, compress L0, leave preview + ref in L0
+```
+
+The stash manifest (injected via `Inject.skeleton("stash")`) tells the agent what's available:
+
+```xml
+<stash count="3" budget="82%">
+  <entry ref="msg-a1b2c3d4" type="tool_result" tool="read_file" tokens="2400" age="4 turns" />
```
Unshure, line 236: I like the shape of the object defined here, but this gets me thinking about future updates. How do you envision us updating this shape in the future? Versions?

team/designs/0015-context-manager.md
lizradway: interesting. i think for the purpose of this doc, the stash is meant to be an extension of the agent's context window outside of itself and is not meant to be integrated with managed solutions. 


but would be curious to follow up more on this use case.

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+| `"schema-only"` | JSON: keep keys/structure, drop values | Large JSON tool results where shape matters but data doesn't |
+| `"collapse-pairs"` | tool_use + tool_result: one-line summary | Verbose tool loops where the action matters but details don't |
+| `"deduplicate"` | Merge near-duplicate messages into one | Retry loops that produce similar results repeatedly |
+
+</details>
+
+---
+
+<details>
+<summary><b>Appendix E: Alternative API Shapes Considered</b></summary>
```
zastrowm, line 953: Doc Feedback: going forward, I'd like to see alternatives featured more prominently.  This doc basically presumes the accepted design which guides the conversation that way, while I think choosing the right approach is a gate that should be explicitly walked-through  - if the proposed design *is* the superior one, that should quickly come through and we can move on.  Keeping alternatives in the appendix basically makes it an optional discussion IMHO

team/designs/0015-context-manager.md
lizradway: could add as configs eventually if we get a call for that, but not sure of the customer use case of that for now.

team/designs/0015-context-manager.md
lizradway: parameter overload tbh. 


i think if a user wants to do both, they can just define two offload methods its only one extra line so not a big deal imo.

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+```typescript
+type TokenBudget = {
+  limit: number       // model's context window size
+  used: number        // current token usage
+  remaining: number   // limit - used
+  ratio: number       // used / limit (0-1)
+}
+```
+
+**Where the data comes from**: `limit` reads from `agent.model.context_window_limit` (already exists on the model config). `used` reads from `BeforeModelCallEvent.projected_input_tokens` (already computed by the existing proactive compression hook). The ContextManager wraps the existing infrastructure into a typed object that strategies, tools, and 3P code can consume uniformly.
```
opieter-aws, line 471: Note the timing issue here with calculating `projected_input_tokens` in the loop versus message array modification

team/designs/0015-context-manager.md
lizradway: this is done in the memory Manager already since it can read from L0 (and possibly L1 in future)

team/designs/0015-context-manager.md
lizradway: storage woo!

team/designs/0015-context-manager.md
JackYPCOnline, line 425: this actually confused me a bit, first class param object has different functionality when it becomes a tool

team/designs/0015-context-manager.md
lizradway: this should survive session restore?  `ContextManager-owned store for recoverable content.`

team/designs/0015-context-manager.md
lizradway: same comment above, responded there

team/designs/0015-context-manager.md
opieter-aws, line 550: Can we reuse the context injector or extend it?

team/designs/0015-context-manager.md
gautamsirdeshmukh: (naive question) could `stash` just accept a `Storage` implementation? This ties to another comment of mine about supporting offload/inject with external stores

gautamsirdeshmukh: Jinx

team/designs/0015-context-manager.md
lizradway: yeah fair enough probably a good idea. i didn't scope out the ref super indepth i would want to run some experimentation on what works best for this. 


im wondering if tool name/id could be sufficient for most use cases or if we would want a parcel preview.

team/designs/0015-context-manager.md
lizradway: yes!

team/designs/0015-context-manager.md
lizradway: context_manager.as_tool()!

team/designs/0015-context-manager.md
lizradway: we could always keep `context_manager="agentic"` around as a facade that maps to passing CM as a tool if we want.

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+| `truncate` | Partial content (head/tail/head-tail) | Keep a preview, stash the full thing | Inject a truncated view of external content |
+| `summarize` | LLM-generated summary | Condense old conversation | Inject a summary of external content |
+| `skeleton` | Structural outline (signatures, refs) | Code: just signatures | Stash: just refs + metadata |
+
+Truncation modes:
+- `"head"`: keep the first N tokens
+- `"tail"`: keep the last N tokens
+- `"head-tail"` (default): keep both ends, drop the middle
+
+Summarization modes:
```
mehtarac, line 112: are these modes defined in context management summarization strategies? Other strategies that could be beneficial could be:

- "`decisions`": retain what was decided and why. Useful for remembering outcomes across context summarization boundaries
- "`intent`": what the user wants and what's left to do. Task-list shaped.

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+  search(options?: { query?: string, tool?: string, type?: string, scope?: string, limit?: number }): Promise<StashEntry[]>
+
+  // inject
+  inject(source: string, content: string): void
+}
+```
+
+### 5.5 Token Budget
+
+The `TokenBudget` is a read-only view the ContextManager exposes for "how full is the context window":
```
JackYPCOnline, line 460: Should we up-level it to agent? otherwise, how can it work when we switch model? this makes cm stateful right

team/designs/0015-context-manager.md
lizradway: yes these are all separate. i think pin and unpin could be combined, but im hesitant of parameter overload of tools

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+
+### 2.4 Inject Sources
+
+What sources `Inject` can pull from, limited to what the ContextManager **owns**:
+
+| Source | What it provides | Typical representation |
+|--------|-----------------|----------------------|
+| `"stash"` | Offloaded content manifest (see §3) | `skeleton`: refs + metadata |
+| `"budget"` | Token utilization stats | bare: as-is |
+| `"clock"` | Current datetime | bare: as-is |
```
Unshure, line 163: How do you see ContextManager differing from SystemPrompt? Ive seen people inject the datetime in the system prompt, but here its injected in via the ContextManager. What is the pit of success here?

team/designs/0015-context-manager.md
```diff
@@ -0,0 +1,1421 @@
...
+
+### 2.5 Triggers (`.when()`)
+
+**Offload** strategies use `.when(event?, conditions?)` to declare when they fire.
+
+Without an event param, a strategy evaluates on every lifecycle event and fires whenever its target matches and conditions are met. Passing an event restricts it to only that hook (e.g. `BeforeModelCallEvent` for summarization, so it doesn't fire after every tool call).
+
+```typescript
+// Event implicit: tool results only arrive after tool calls
+Offload.truncate("toolResults", { previewTokens: 750 })
```
notowen333, line 176: Wondering if there is value around building a graph relationship between evicted content? That would open the door to new retrieval mechanisms.

team/designs/0015-context-manager.md
gautamsirdeshmukh, line 301: Similar thought from before - instead of a `stash` boolean flag can the user just pass a Storage implementation here (and if absent then no stashing)

team/designs/0015-context-manager.md
opieter-aws: I see, I interpreted this as 'recoverable this session', given that below is stated: `For v1, the stash grows until the session ends or storage is cleaned up externally.`

Would dedicate a short section to session restore with L1 versus session manager

team/designs/0015-context-manager.md
opieter-aws: It'd fold into the latest message rather than system prompt, keeping the cache intact

agent-of-mkmeral:
# Adversarial analysis: where the ContextManager design breaks

**TL;DR:** Stress-tested this design against real systems (Headroom — verified from its actual repo/docs — LLMLingua, Letta/MemGPT, Anthropic context editing, OpenAI Responses, LangGraph, pydantic-ai, Bedrock invariants). The `Offload`/`Inject` algebra works for the 80% rule-based case, but **success criterion #3 ("3P compressors slot in with ~15 lines") is the weakest claim in the doc.** Six structural breaks:

1. **Closed algebra** — `Strategy = OffloadStrategy | InjectStrategy` + closed enums; reorder, rewrite-in-place, and request mutation have no home (the doc's own `collapse-pairs` roadmap item can't be expressed by its own taxonomy)
2. **Wrong granularity for Headroom-class libraries** — they need whole-request, per-call access; Headroom's shipped Strands integration is *model wrapping*, and the doc's ~15-line example captures ~20% of the library
3. **Missing view-vs-history axis** — offload permanently mutates L0; LangGraph/pydantic-ai/Headroom all converged on "compress the view, keep pristine history," which the design cannot express
4. **Retrieval not pluggable** — no `StashProvider`; Headroom CCR, vector search, filesystem offload compete with `get_context` instead of backing it
5. **Provider-native context management breaks the L0 assumption** — Anthropic server-side edits and OpenAI `previous_response_id` can't be expressed; running both diverges stash from reality
6. **Spec-level correctness bug** — bare `Offload("toolResultErrors")` orphans `toolUse` blocks → `ValidationException` on Bedrock/Anthropic; the invariant is never mentioned

Top fix: add the view/history axis + a low-level `ContextTransform` escape hatch that `Offload`/`Inject` become sugar over. Full detail below.

<details>
<summary><b>The six structural breaks (full detail)</b></summary>

### 1. The strategy algebra is a closed union, not an interface

```typescript
type Strategy = OffloadStrategy | InjectStrategy
```

@mkmeral already flagged this in review and it's the root problem. The design hardcodes exactly two directions, and inside them, closed enums:

- `representation: "truncate" | "summarize" | "skeleton" | "full"` — yet Appendix F's own 3P example uses `representation: "headroom"`. The type as written rejects the doc's own example.
- `OffloadConditions = { threshold?, utilization?, skipRecent? }` — a closed struct, **no predicate function**. Any library that triggers on its own signal (drift detection, relevance scoring, "stale for N turns") cannot express its trigger.
- `OffloadTarget` — a stringly-typed taxonomy (`"toolResults"`, `"code"`, tool names, `!` exclusions). No predicate matcher (match by regex, metadata, token count, mimetype, content block type).

**What has no home in `offload | inject`:**
- **Reorder** — LongLLMLingua's document reordering (mitigating lost-in-the-middle) changes *position*, not representation. Neither offload nor inject.
- **Rewrite in place** — dedupe across messages, PII redaction, normalization. Not "choose a cheaper representation," and must NOT stash the original (redaction!). The stash-on-offload coupling is wrong for these.
- **Request mutation** — see break #5.

The doc even self-incriminates: `"collapse-pairs"` (P2, Appendix D) merges a toolUse **assistant** message with a toolResult **user** message — that crosses the target taxonomy's own category boundaries. The engine's per-message target matching can't express its own roadmap item.

### 2. Headroom-class libraries integrate at the wrong granularity

I verified against the actual Headroom codebase/docs. Its architecture is `CacheAligner → ContentRouter → SmartCrusher/CodeCompressor/Kompress-v2` operating on the **entire request**: system prompt, tool schemas, full message array, and request parameters. Specifically:

- **CacheAligner** stabilizes the *whole prompt prefix* so provider KV caches hit. It needs to see and control full request layout — impossible from `compress(matchedMessages, budget)`.
- **Live-zone compression** compresses *only new bytes*; "frozen prefix stays byte-identical so provider cache is not busted. **History is never dropped.**" This is philosophically opposite to the design's engine, which rewrites mid-history messages (see break #6b).
- **ContentRouter** does its own per-content routing — the SDK's target matching would fight it. Feeding it pre-filtered `"assistantMessages"` (the doc's own Appendix I example) neuters it; Headroom's biggest wins are tool outputs (92% on code search). **The doc's flagship 3P example misuses the library it's named after.**
- **Output shaper** appends to the system prompt and mutates request params (`reasoning_effort`, `thinking.budget_tokens`). Zero reach from an `OffloadMethod`.

And the kicker: **Headroom already ships a Strands integration, and it's model wrapping** (`HeadroomStrandsModel(wrapped_model=...)`) plus a hook provider for tool results. They chose the model boundary because that's the only seam that gives whole-request, per-call, non-destructive access. The design's ~15-line `HeadroomStrategy` captures maybe 20% of the library and would be a *regression* from their existing integration.

### 3. Durable mutation vs. ephemeral view — the missing axis

This is the single deepest break. The design's offload **permanently mutates L0** (the session-persisted message array). But the industry has converged on distinguishing two modes:

- **LangGraph** `pre_model_hook`: return `llm_input_messages` (ephemeral view, history untouched) *or* `messages` + `RemoveMessage` (durable mutation). Two explicit channels.
- **pydantic-ai** `history_processors`: transform what's *sent to the model* each call; stored history stays pristine.
- **Headroom** model wrapping: compresses per call from the pristine original each time.

Why the view mode matters: compressing a *view* re-derives from lossless originals every call — quality doesn't degrade. Compressing *history* compounds: summarize-of-truncate-of-summarize, each pass eating information. It also means a mid-session model switch (ModelRouter, P3) hands the big-window model a pre-degraded history it didn't need. **The design cannot express "compress what the model sees, keep what the agent has."** There is no strategy attribute, no channel, nothing. Any library built on the view model (Headroom, LLMLingua-style prompt compression) either doesn't fit or gets forcibly downgraded to destructive mode.

### 4. The retrieval path is not pluggable

The engine writes originals to the CM-owned stash; `get_context`/`search_context` read from it. But there's no `retrieve()` on `OffloadMethod` and no `StashProvider` interface:

- **Headroom CCR** stores originals in *its own* local cache with a **TTL** and its own retrieval tools (`headroom_retrieve` / rtk via MCP). Plugged in as an `OffloadStrategy`, you get **two stashes, two ref vocabularies, two retrieval tools** — and the SDK's `get_context` can't serve CCR-held content. The agent now needs to know which of two systems holds which original.
- A compressor whose "original" lives remotely (vendor-side handle) can't answer `get_context` at all.
- `search_context({query})` implies a content index. With what backend? In-memory grep won't scale to a 50-entry stash of 30K-token blobs, and a team with a vector store can't plug it in — search is not an extension point.
- Filesystem-offload patterns (Manus-style: agent writes results to files, retrieves with `cat`/`grep` — which the doc's own "internal ref 2" team fell back to!) bypass the stash entirely, and the manifest can't represent them.

### 5. Provider-native context management can't be expressed — and will fight the CM

- **Anthropic context editing** (`context_management: {edits: [{type: "clear_tool_uses_...", trigger, keep, ...}]}`): the *server* clears old tool results and substitutes placeholders. Expressing "delegate truncation to the provider" requires a strategy to **patch the request**, which `compress(messages) -> messages` cannot do. Worse, running both: CM stashes originals it thinks are in L0 while the provider silently clears them server-side — the manifest and reality diverge, budget accounting (`projected_input_tokens`) is wrong, and two triggers race at different thresholds.
- **OpenAI Responses API** (`previous_response_id` / server-held conversation state): the client doesn't even *hold* the history. There is no L0 array to offload from. The entire two-layer model assumes SDK-owned client-side messages; a growing class of provider features breaks that assumption.
- Same story for AgentCore-style managed memory (opieter's "paid L1" thread is a special case of this).

### 6. Message-level granularity breaks on provider invariants and multimodal content

**(a) The spec as written produces API errors.** `Offload("toolResultErrors")` — "Nothing remains in L0." On Bedrock Converse and Anthropic, every `toolUse` **must** be followed by a matching `toolResult`; deleting a toolResult message while its toolUse remains → `ValidationException`. The SDK's own `SlidingWindowConversationManager` has explicit pair-boundary handling for exactly this (`sliding_window_conversation_manager.py:235,280`), and the doc never mentions the invariant. Same landmine when `Offload.summarize(ratio=0.3)` cuts a range mid-pair. The engine needs a guaranteed invariant-repair pass, and bare offload of tool results must leave a placeholder — contradicting §2.6's table.

**(b) Rolling rewrites bust the prompt cache by design.** `.when({threshold: 1500, skipRecent: 3})` firing on `AfterToolCallEvent` rewrites message *n−3* on every tool call — a moving mutation point that invalidates the provider prefix cache **every single turn**. On Anthropic, uncached input is ~10× cached input: the strategy can save tokens and *raise the bill*. The doc has one `"cache-aligned"` summarize mode, but truncate strategies have no cache awareness, the `TokenBudget` exposes no cache-breakpoint info, and a 3P strategy has no way to even *ask* where the stable prefix ends. Headroom's frozen-prefix invariant is unimplementable through this interface. (@mkmeral raised cache invalidation on inject; offload is the much bigger offender.)

**(c) The unit is the message, but content is blocks.** A toolResult containing text + image: offload just the image block? (@mehtarac asked; no answer in the spec.) Anthropic's native editing works at block granularity. `previewTokens`/head-tail are meaningless for images; Headroom does image→smaller-image ML compression, which is "same modality, cheaper" — not representable in the truncate/summarize/skeleton vocabulary and awkward as a message-level `compress`.

</details>

<details>
<summary><b>Library-by-library plug-in test</b></summary>

| Library / system | Fits `OffloadMethod`? | What breaks |
|---|---|---|
| **Headroom** | ~20% | Needs whole-request view, cache alignment, view-not-history semantics, own retrieval store (CCR/TTL), request-param mutation. Already integrates via model wrapping instead. |
| **LLMLingua / LongLLMLingua** | Partially | Question-aware compression needs the **current query + instruction** — `compress(messages, budget)` passes neither. Document reordering = neither offload nor inject. Token-level compression is prompt-scoped, not message-scoped; needs a local GPU model (latency profile the sync-looking hook path never discusses). |
| **Letta / MemGPT** | No | Core memory = agent-editable blocks compiled **into the system prompt**. The design: "system prompt is always implicitly pinned" and inject is messages-only. Self-editing memory (rewrite-in-place via tools) has no direction in the algebra. `.as_tool()` covers the tool half, but the injection target is off-limits. |
| **Anthropic context editing / memory tool** | No | Request parameter + server-side mutation. No requestPatch capability; dual-manager divergence (break #5). |
| **OpenAI Responses (`previous_response_id`)** | No | No client-side L0 exists. |
| **LangGraph `SummarizationNode` / pydantic-ai `history_processors`** | Shoehorned | Their primitive is the fully-general `(messages) -> messages` over the **whole** array with an ephemeral-view option. The design offers no low-level escape hatch — everything must be re-vocabularized into Offload/Inject targets. A one-line pydantic-ai processor becomes "write a custom strategy, pick a fake target, accept durable mutation." |
| **Claude-Code-style compaction** | Yes | `/compact` ≈ `Offload.summarize` at utilization threshold. This one fits fine. |

</details>

<details>
<summary><b>Secondary breaks (8 smaller but real issues)</b></summary>

1. **Composition semantics are hand-waved.** "Multiple strategies can match the same message: they compose, not compete" — in what order? Is a summary output (an assistantMessage) itself a target for the next pass (summary-of-summary loops)? After truncate stashes the original, what does a later summarize on the same message stash — the preview? Idempotency is listed as an engine task but the *semantics* are unspecified, and lossy transforms don't commute.
2. **Three coexisting injection subsystems, no arbiter.** §7.3 keeps `ContextInjector` as an independent plugin *forever*; MemoryManager injects on its own lifecycle; CM `Inject` strategies fire before model calls. They "share the token budget but don't reference each other" — when the sum overflows, who yields? Unspecified. Also §2.4 says RAG belongs to other subsystems while Appendix I demos "RAG providers that don't use MemoryManager" as Inject sources — the boundary is contradictory within the doc.
3. **Strip-and-reinject invalidates cache from the previous injection point.** Last turn's injected message (appended at end) is now mid-history; stripping it mutates the prefix. The review-thread answer ("inject at end avoids invalidation") only holds for a single turn.
4. **Budget jurisdiction gap: tool schemas and system prompt.** `TokenBudget.used` counts the whole request, but strategies can only touch messages. For big agents, tool definitions are a top-3 context consumer (hence Anthropic's tool-defs clearing and tool-search). A "context" manager that can't manage a third of the context will push users back to hooks.
5. **Token accounting is provider-shaped.** `threshold` (absolute tokens per message) — counted by whose tokenizer? Images/documents? `projected_input_tokens` exists per-model; strategy decisions become model-dependent, and JackYPCOnline's "what happens when the model switches" thread is unresolved: budget-conditioned strategies silently change behavior mid-session.
6. **The stash default recreates the doc's own opening bug.** Storage "falls back to in-memory" while L0 (with preview stubs) persists via session manager. Restore session → stubs point at nothing → "agent hallucinates instead of admitting it can't retrieve" — the *exact* problem statement from §1, reintroduced by the default config. In-memory stash + any session persistence should be a hard error or auto-degrade to `stash: false`.
7. **Refs go stale under the agent's feet.** Agentic mode: agent reads `context_status`, plans `pin_context(refs)`; meanwhile a rule-based summarize merges those N messages into one. The refs the agent is holding now dangle mid-turn. Also unaddressed: parallel tool calls firing `AfterToolCallEvent` concurrently against a shared mutable L0 — no locking/serialization story.
8. **Serialization claim vs. closures.** Appendix E rejects builders as "less serializable," but the chosen design's custom strategies/inject sources are closures over clients and render functions — equally unserializable. Matters the moment you want config-file-defined or dashboard-defined strategies.

</details>

<details>
<summary><b>What the design gets right (credit where due)</b></summary>

- Stash + manifest + **bounded, paginated `get_context`** directly fixes the "retrieval re-creates the overflow" bug — and independently matches Headroom's CCR + rtk shape. Convergent evolution = validated need.
- Per-tool targeting with `!` exclusions maps to the exact internal-team workaround cited.
- `.as_tool()` for agent-managed context matches Letta's proven pattern.
- Observability section (stream events + OTEL + lifecycle events) is the most complete answer to #1511/#2804 anywhere in the doc's lineage.
- Killing three disconnected components for one system is the right call. The question is whether the one system's floor is an open primitive or a closed vocabulary.

</details>

<details>
<summary><b>What I'd change (7 fixes, ranked by leverage)</b></summary>

1. **Add the view/history axis** (`mutates: "view" | "history"` per offload strategy, or a distinct `Transform` channel applied at model-call time). Single highest-leverage fix; it's what LangGraph, pydantic-ai, and Headroom all converged on, and it makes cache-aligned + non-compounding compression possible.
2. **Open the algebra at the bottom.** Ship a low-level interface — e.g. `ContextTransform: (view: RequestView) => ContextEdit[]` where edits cover replace/remove/insert/reorder + `requestPatch` — and define `Offload`/`Inject` as sugar over it. This answers mkmeral's `Strategy = A | B` thread, absorbs reorder/rewrite/provider-native delegation, and *is* the escape hatch competitors ship as one line.
3. **`StashProvider` interface** (`write/get/search/manifest/ttl`) so CCR, vector stores, and filesystem offload back `get_context`/`search_context` instead of competing with them. (Also resolves gautam's and opieter's storage threads.)
4. **Accept predicates in `.when()` and in targets.** Cheapest fix in the doc; eliminates most of the closed-taxonomy complaints without new concepts.
5. **Specify the invariant-repair pass** (toolUse/toolResult pairing, alternation) as an engine guarantee, and fix §2.6: bare offload of tool results leaves a placeholder, period.
6. **Make cache economics first-class**: expose stable-prefix/cache-breakpoint info on `TokenBudget` (or an `EngineContext`), and change the default engine behavior from rolling `skipRecent` rewrites to compress-at-stable-boundaries.
7. **Define the multi-injector budget arbiter** (or fold ContextInjector in now rather than v2-forever).

</details>

<details>
<summary><b>Method note</b></summary>

Read the full 1421-line design + all 40+ review threads on PR #3307; pulled Headroom's repo/README/docs (incl. their existing `HeadroomStrandsModel` integration guide) rather than trusting the doc's characterization; checked the current SDK's `SlidingWindowConversationManager` pairing logic and vended context-offloader plugin for what invariants exist today.

</details>

team/designs/0015-context-manager.md
lizradway: in its current state, this utilizes the context injector under the hood

lizradway:
> Adversarial analysis: where the ContextManager design breaks

1. can always add Mutation Strategy if we want later since its a union
2. full headroom model wrapping can extend HeadroomContextManager, or its offloading strategy can be condensed as an Offload Strategy for simplicity.
3. im actually leaning towards writing to stash on messageAddedEvent and then only mutating L0 on offload
4. we can always add that later if we want!
5. thats fine we just need framework level stash/observability
6. internal impl detail we'll handle internally

team/designs/0015-context-manager.md
lizradway: sure thought abt that approach as well, can see in api bar raising

team/designs/0015-context-manager.md
lizradway: i am extremely interested in AST, i think i would be interested in that as well to research

team/designs/0015-context-manager.md
lizradway: i think we discussed this offline that context manager is model agnostic

team/designs/0015-context-manager.md
lizradway: i think decisions would hopefully be an audit log, which we are working on as well!

intent seems to be the todo plugin

team/designs/0015-context-manager.md
lizradway: i would like to use this paradigm throughout the sdk ! i think it makes a lot of sense for our constructs to be tool-able

team/designs/0015-context-manager.md
lizradway: yes i would like to add that atleast as P1

team/designs/0015-context-manager.md
lizradway: team consensus was yes!

lizradway: team consensus was that P0/P1s are a go

team/designs/0015-context-manager.md
lizradway: yup! still works like in current offloader, just no preview

team/designs/0015-context-manager.md
lizradway: can always add more strategies to the union

mkmeral (approved):

team/designs/0015-context-manager.md
lizradway: i see versioning at the entry level, whereas stash-level versioning can be done via snapshots in session manaager.



ie. `<entry ref="msg-a1b2c3d4" type="tool_result" tool="read_file" tokens="2400" age="4 turns" version="1" />`

team/designs/0015-context-manager.md
lizradway: this is injected as a user message and is supposed to represent a per-runtime state. 

pit of success is:
- no sys prompt updating
- use injection for entries that are dynamic per turn (ie. stash, budget, clock), but not per session (ie. date, persona, etc). those are skills imo

PR #3337, opened 2026-07-17
fix(ts): abort in-flight Bedrock requests on cancellation

### LINKED ISSUE #3224: TS: cancelSignal doesn't abort the in-flight Bedrock stream — generation and billing continue after cancel
### Checks

### SDK Language

TypeScript

### Strands Version

1.9.0 (root cause also verified on current `main`)

### Language Runtime Version

Node 20 / Node 22

### Operating System

Amazon Linux 2023 (AWS Lambda)

### Installation Method

npm

### Steps to Reproduce

```ts
const controller = new AbortController()
const run = (async () => {
  for await (const _ of agent.streamAsync('Write a 2000+ word essay.', { cancelSignal: controller.signal })) {}
})()
setTimeout(() => controller.abort(), 1000)
await run
```

Use a fast-streaming Bedrock model and watch CloudWatch Bedrock output-token metrics (or a proxy on the h2 stream) after the abort.

### Expected Behavior

Cancelling a turn aborts the in-flight model request, so the provider stops generating (and billing) promptly.

### Actual Behavior

The local `for await` loop stops quickly, but the Bedrock ConverseStream request is never aborted — the model keeps generating server-side and output tokens keep billing until the turn hits a natural message boundary. With a fast token stream we measured ~27s of continued generation/billing after cancel. With the signal threaded into the transport (see below), generation and billing stop within ~6ms–1.5s (CloudWatch-verified).

### Additional Context

The signal is only checked between stream events, never passed to the transport:

- `strands-ts/src/models/bedrock.ts` L629-630: `this._client.send(command)` — no send options, so no `abortSignal` reaches the AWS SDK request handler. Same for the non-streaming path at L644-645.
- `strands-ts/src/agent/agent.ts` L2154-2155: `_throwIfCancelled()` runs inside the `while (!result.done)` loop, i.e. between yielded events. Throwing `CancelledError` unwinds the local iteration but the HTTP/2 stream stays open.

Separate smaller thing (will file separately if wanted): on `CancelledError` the agent appends a synthetic `'Cancelled by user'` assistant message (`agent.ts` L1687), which pollutes history for apps that render their own cancel state. An opt-out would help.

### Possible Solution

Pass the signal through to `this._client.send(command, { abortSignal })`, and destroy the streaming response body when the signal fires — in practice the body destroy is what reliably ends the h2 stream mid-flight. Needs the signal plumbed from the agent loop through `Model.stream()` to the provider; the Agent already holds it (`agent.ts` L1072-1073).

We run this in production as a `BedrockModel` subclass and it works well — happy to open a PR if you're open to it.

### Related Issues

None found on this. Loosely adjacent: #3050, #1819.

jstar0:
## Description

Cancelling an Agent invocation currently stops local event processing, but the
composed cancellation signal never reaches the Bedrock transport. An in-flight
Converse request can therefore continue generating after the caller has
cancelled the turn.

This change:

- adds an optional cancellation signal to `StreamOptions` and passes the
  Agent's composed per-invocation signal to the model stream;
- forwards that signal to both streaming and non-streaming Bedrock requests via
  the AWS SDK `abortSignal` request option;
- replaces the AWS browser default WebSocket wrapper for this model's HTTPS
  requests with the abort-aware Smithy fetch handler, while preserving Node and
  caller-provided request handlers;
- preserves the existing Agent cancellation result when the provider transport
  reports the abort as `AbortError`;
- adds regressions for both Bedrock request modes, active transport abortion,
  a fake producer that stops emitting after cancellation, and the real default
  Chromium transport boundary.

The new option is optional and provider-dependent, other model providers are
unchanged, and Bedrock keeps its existing one-argument `send(command)` path
when no signal is present.

## Related Issues

Fixes #3224

## Documentation PR

No documentation change is needed. This makes the existing `cancelSignal`
contract apply to the in-flight Bedrock request.

## Type of Change

Bug fix

## Testing

- `npm test -- src/agent/__tests__/agent.cancel.test.ts src/models/__tests__/bedrock.test.ts` (194 tests)
- `npm test` (3789 tests)
- `npm run test:browser` (3086 passed, 5 skipped, including 3 browser transport-boundary tests)
- `npm run type-check`
- `npm run lint`
- `npm run format:check`
- `npm run build`
- `npm run test:integ:selective` was attempted locally; provider-backed suites
  could not complete because repository AWS/Bedrock test credentials were not
  available. Credential-independent tests passed, and Bedrock-backed tests were
  skipped or failed during provider setup.

- [ ] I ran `hatch run prepare` (not applicable to this TypeScript-only change)

## Checklist

----

### DIFF
--- a/strands-ts/src/models/__tests__/bedrock.transport.test.browser.ts
+++ b/strands-ts/src/models/__tests__/bedrock.transport.test.browser.ts
@@ -35,7 +35,7 @@ describe('BedrockModel browser transport cancellation', () => {
       })
       const controller = new AbortController()
       const output = model.stream([new Message({ role: 'user', content: [new TextBlock('Hello')] })], {
-        signal: controller.signal,
+        cancelSignal: controller.signal,
       })
       const iterator = output[Symbol.asyncIterator]()
 
--- a/strands-ts/src/models/__tests__/bedrock.transport.test.browser.ts
+++ b/strands-ts/src/models/__tests__/bedrock.transport.test.browser.ts
@@ -1,0 +1,68 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import type { BedrockRuntimeClientConfig } from '@aws-sdk/client-bedrock-runtime'
+import { BedrockModel } from '../bedrock.js'
+import { Message, TextBlock } from '../../types/messages.js'
+
+describe('BedrockModel browser transport cancellation', () => {
+  afterEach(() => {
+    vi.unstubAllGlobals()
+  })
+
+  for (const stream of [true, false]) {
+    it(`aborts the default ${stream ? 'streaming' : 'non-streaming'} fetch transport`, async () => {
+      const fetchMock = vi.fn((input: RequestInfo | URL) => {
+        const request = input as Request
+        return new Promise<Response>((_, reject) => {
+          request.signal.addEventListener(
+            'abort',
+            () => reject(new DOMException('The operation was aborted.', 'AbortError')),
+            { once: true }
+          )
+        })
+      })
+      vi.stubGlobal('fetch', fetchMock)
+
+      const model = new BedrockModel({
+        region: 'us-east-1',
+        modelId: 'test-model',
+        stream,
+        clientConfig: {
+          credentials: {
+            accessKeyId: 'test-access-key',
+            secretAccessKey: 'test-secret-key',
+          },
+        },
+      })
+      const controller = new AbortController()
+      const output = model.stream([new Message({ role: 'user', content: [new TextBlock('Hello')] })], {
+        signal: controller.signal,
+      })
+      const iterator = output[Symbol.asyncIterator]()
+
+      const result = iterator.next()
+      await vi.waitFor(() => expect(fetchMock).toHaveBeenCalledOnce())
+      controller.abort()
+
+      await expect(result).rejects.toMatchObject({ name: 'AbortError' })
+    })
+  }
+
+  it('preserves a caller-provided request handler', () => {
+    const requestHandler = {
+      metadata: { handlerProtocol: 'custom' },
+      handle: vi.fn(),
+      updateHttpClientConfig: vi.fn(),
+      httpHandlerConfigs: vi.fn(() => ({})),
+    }
+
+    const model = new BedrockModel({
+      region: 'us-east-1',
+      modelId: 'test-model',
+      clientConfig: {
+        requestHandler: requestHandler as NonNullable<BedrockRuntimeClientConfig['requestHandler']>,
+      },
+    })
+
+    expect(model['_client'].config.requestHandler).toBe(requestHandler)
+  })
+})
--- a/strands-ts/src/models/bedrock.ts
+++ b/strands-ts/src/models/bedrock.ts
@@ -630,8 +630,8 @@ export class BedrockModel extends Model<BedrockModelConfig> {
       if (this._config.stream !== false) {
         // Create and send the command
         const command = new ConverseStreamCommand(request)
-        const response = options?.signal
-          ? await this._client.send(command, { abortSignal: options.signal })
+        const response = options?.cancelSignal
+          ? await this._client.send(command, { abortSignal: options.cancelSignal })
           : await this._client.send(command)
         // Stream the response
         if (response.stream) {
@@ -647,8 +647,8 @@ export class BedrockModel extends Model<BedrockModelConfig> {
         }
       } else {
         const command = new ConverseCommand(request)
-        const response = options?.signal
-          ? await this._client.send(command, { abortSignal: options.signal })
+        const response = options?.cancelSignal
+          ? await this._client.send(command, { abortSignal: options.cancelSignal })
           : await this._client.send(command)
         for (const event of this._mapBedrockEventToSDKEvent(response)) {
           yield event
--- a/strands-ts/src/models/bedrock.ts
+++ b/strands-ts/src/models/bedrock.ts
@@ -627,7 +627,9 @@ export class BedrockModel extends Model<BedrockModelConfig> {
       if (this._config.stream !== false) {
         // Create and send the command
         const command = new ConverseStreamCommand(request)
-        const response = await this._client.send(command)
+        const response = options?.signal
+          ? await this._client.send(command, { abortSignal: options.signal })
+          : await this._client.send(command)
         // Stream the response
         if (response.stream) {
           let lastStopReason: string | undefined
@@ -642,7 +644,9 @@ export class BedrockModel extends Model<BedrockModelConfig> {
         }
       } else {
         const command = new ConverseCommand(request)
-        const response = await this._client.send(command)
+        const response = options?.signal
+          ? await this._client.send(command, { abortSignal: options.signal })
+          : await this._client.send(command)
         for (const event of this._mapBedrockEventToSDKEvent(response)) {
           yield event
         }
--- a/strands-ts/src/models/bedrock.ts
+++ b/strands-ts/src/models/bedrock.ts
@@ -45,6 +45,7 @@ import {
   type CitationsDelta as BedrockCitationsDelta,
   type GuardrailTraceAssessment,
 } from '@aws-sdk/client-bedrock-runtime'
+import { FetchHttpHandler, type FetchHttpHandlerOptions } from '@smithy/fetch-http-handler'
 import {
   type BaseModelConfig,
   type CacheConfig,
@@ -438,13 +439,15 @@ export class BedrockModel extends Model<BedrockModelConfig> {
       ? `${clientConfig.customUserAgent} strands-agents-ts-sdk`
       : 'strands-agents-ts-sdk'
 
+    const requestHandler = withDefaultRequestTimeout(clientConfig?.requestHandler)
     this._client = new BedrockRuntimeClient({
       ...(clientConfig ?? {}),
-      requestHandler: withDefaultRequestTimeout(clientConfig?.requestHandler),
+      requestHandler,
       // region takes precedence over clientConfig
       ...(region ? { region: region } : {}),
       customUserAgent,
     })
+    applyAbortAwareBrowserRequestHandler(this._client.config, requestHandler)
 
     if (apiKey) {
       applyApiKey(this._client, apiKey)
@@ -1805,6 +1808,25 @@ function withDefaultRequestTimeout(
   // Use `??` rather than spread order so an explicit `requestTimeout: undefined` still gets
   // the default (spread would otherwise overwrite the default with `undefined`, disabling it).
   return { ...options, requestTimeout: options.requestTimeout ?? DEFAULT_REQUEST_TIMEOUT_MS }
+}
+
+/**
+ * Replaces the AWS browser default WebSocket wrapper for this model's HTTPS
+ * requests. The wrapper drops HttpHandlerOptions when delegating to fetch, so
+ * abortSignal never reaches the transport. Node defaults and caller-provided
+ * handler instances are left unchanged.
+ */
+function applyAbortAwareBrowserRequestHandler(
+  config: BedrockRuntimeClientResolvedConfig,
+  requestHandler: NonNullable<BedrockRuntimeClientConfig['requestHandler']>
+): void {
+  if (typeof (requestHandler as { handle?: unknown }).handle === 'function') {
+    return
+  }
+  if (config.requestHandler?.metadata?.handlerProtocol !== 'websocket/h1.1') {
+    return
+  }
+  config.requestHandler = new FetchHttpHandler(requestHandler as FetchHttpHandlerOptions)
 }
 
 /**
--- a/strands-ts/src/models/model.ts
+++ b/strands-ts/src/models/model.ts
@@ -130,6 +130,11 @@ export interface BaseModelConfig {
  */
 export interface StreamOptions {
   /**
+   * Signal used to abort an in-flight model provider request.
+   */
+  signal?: AbortSignal
+
+  /**
    * System prompt to guide the model's behavior.
    * Can be a simple string or an array of content blocks for advanced caching.
    */
--- a/strands-ts/src/models/model.ts
+++ b/strands-ts/src/models/model.ts
@@ -130,10 +130,10 @@ export interface BaseModelConfig {
  */
 export interface StreamOptions {
   /**
-   * Optional signal that a provider implementation can forward to abort an in-flight request.
+   * Optional cancellation signal that a provider implementation can forward to abort an in-flight request.
    * Support is provider-dependent.
    */
-  signal?: AbortSignal
+  cancelSignal?: AbortSignal
 
   /**
    * System prompt to guide the model's behavior.
--- a/strands-ts/src/models/model.ts
+++ b/strands-ts/src/models/model.ts
@@ -130,7 +130,8 @@ export interface BaseModelConfig {
  */
 export interface StreamOptions {
   /**
-   * Signal used to abort an in-flight model provider request.
+   * Optional signal that a provider implementation can forward to abort an in-flight request.
+   * Support is provider-dependent.
    */
   signal?: AbortSignal
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -6755,6 +6755,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@aws-sdk/client-bedrock-runtime": "^3.1097.0",
+        "@smithy/fetch-http-handler": "^5.6.10",
         "@types/json-schema": "^7.0.15",
         "uuid": "^14.0.1",
         "yaml": "^2.8.3"
--- a/strands-ts/package.json
+++ b/strands-ts/package.json
@@ -243,6 +243,7 @@
   },
   "homepage": "https://github.com/strands-agents/harness-sdk#readme",
   "dependencies": {
+    "@smithy/fetch-http-handler": "^5.6.10",
     "@aws-sdk/client-bedrock-runtime": "^3.1097.0",
     "@types/json-schema": "^7.0.15",
     "uuid": "^14.0.1",
--- a/strands-ts/src/agent/__tests__/agent.cancel.test.ts
+++ b/strands-ts/src/agent/__tests__/agent.cancel.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it } from 'vitest'
+import { describe, expect, it, vi } from 'vitest'
 import { Agent } from '../agent.js'
 import { AfterInvocationEvent, AfterModelCallEvent, BeforeModelCallEvent } from '../../hooks/index.js'
 import { MockMessageModel } from '../../__fixtures__/mock-message-model.js'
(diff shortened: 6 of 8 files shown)


jstar0:
@mkmeral This is ready for review, following your approval on #3224.

mkmeral:
@strandly-the-agent can you review the PR?

strands-ts/src/models/bedrock.ts
```diff
@@ -627,7 +627,9 @@ export class BedrockModel extends Model<BedrockModelConfig> {
       if (this._config.stream !== false) {
         // Create and send the command
         const command = new ConverseStreamCommand(request)
```
strandly-the-agent, line 632: 🔴 **The default browser transport drops this abort signal.** With the pinned AWS SDK, browser Bedrock wraps requests in `WebSocketFetchHandler`; its `handle(request)` ignores the second handler-options argument, delegates HTTPS with one argument, and never observes aborts on its WebSocket path. A cold full Agent→Bedrock browser bundle reproduced all four streaming/non-streaming × external/`agent.cancel()` cases staying pending with the invocation lock held, while an explicit `FetchHttpHandler` control cancelled and reset correctly.

Please make the browser-default handler preserve `HttpHandlerOptions.abortSignal` (for example via a compatible AWS SDK version or handler wrapper) and add a browser transport-boundary regression test for both request modes.

strandly-the-agent (changes requested):
🔴 **Changes requested:** the Bedrock abort signal does not reach the transport under the supported default browser configuration.

✅ Reviewed `d65bd58d` against base `41f9f59b`; focused cancellation/Bedrock tests: **194 passed**; local build, type-check, and lint passed.
✅ Cold browser-boundary repro: all 4 default-handler cases (streaming/non-streaming × external/`agent.cancel()`) stayed pending; all 4 `FetchHttpHandler` controls cancelled and cleaned up.
✅ GitHub **CI Gate** is green. ⚠️ No live Bedrock/billing validation was available.
⚠️ This adds an exported `StreamOptions` field: please use `api/needs-review`, then `api/review-complete`; this is a minimal additive API change and does not appear to need an API meeting.

<details><summary>Review coverage</summary>

Independent passes covered correctness, API/DevX, adversarial cancellation, test quality, model-facing behavior, and docs. Node transport cancellation passed local Smithy and HTTP/2 boundary probes. No additional LLM-context findings. I suppressed pre-existing/out-of-scope cancellation gaps that were not needed to resolve #3224.

</details>

strands-ts/src/models/model.ts
```diff
@@ -128,6 +128,11 @@ export interface BaseModelConfig {
  * Options interface for configuring streaming model invocation.
  */
 export interface StreamOptions {
+  /**
+   * Signal used to abort an in-flight model provider request.
+   */
```
strandly-the-agent, line 135: 🟡 This exported provider-neutral contract currently reads as a guarantee, but forwarding is provider-dependent; custom providers must propagate the signal themselves, and the other built-in providers currently ignore it.

```suggestion
  /**
   * Optional signal that a provider implementation can forward to abort an in-flight request. Support is provider-dependent.
   */
```

strands-ts/src/models/bedrock.ts
jstar0: Addressed in commit 837317c5. When the AWS browser client resolves its default websocket/h1.1 handler, Bedrock now substitutes Smithy FetchHttpHandler, so HttpHandlerOptions.abortSignal reaches the real transport. Node defaults and caller-provided handler instances remain unchanged. The unmocked Chromium transport-boundary tests now cover streaming and non-streaming aborts plus custom-handler preservation.

strands-ts/src/models/model.ts
jstar0: Addressed in commit 837317c5. The exported StreamOptions.signal documentation now states that a provider implementation can forward the signal and that support is provider-dependent.

jstar0:
The review fix is pushed as signed commit 837317c5051ed9ce60b8da4fe421004b95783b4b, and the PR body now includes the browser transport-boundary coverage. I could not add the requested api/needs-review label because GitHub returned 403 for this external contributor. Could a maintainer add it? I have not added api/review-complete.

opieter-aws:
@strandly-the-agent can you review this PR again?

strands-ts/src/models/bedrock.ts
```diff
@@ -1803,6 +1810,25 @@ function withDefaultRequestTimeout(
...
+  config: BedrockRuntimeClientResolvedConfig,
+  requestHandler: NonNullable<BedrockRuntimeClientConfig['requestHandler']>
+): void {
+  if (typeof (requestHandler as { handle?: unknown }).handle === 'function') {
+    return
+  }
+  if (config.requestHandler?.metadata?.handlerProtocol !== 'websocket/h1.1') {
+    return
+  }
+  config.requestHandler = new FetchHttpHandler(requestHandler as FetchHttpHandlerOptions)
```
strandly-the-agent, line 1945: 🟡 **The two branches that make this fix correct have no test that can run outside a real browser.** The only file covering this helper is `bedrock.transport.test.browser.ts`, which needs Chromium; the node-side construction tests (`bedrock.test.ts:200-284`) mock the client with a `config` that has no `requestHandler` key, so this branch is unreachable there.

I mutation-tested it: dropping the `handle` early-return (L1823-1825) **and** stripping the option-bag pass-through at L1829 (`new FetchHttpHandler()` instead of `new FetchHttpHandler(requestHandler)` — which silently loses the 120 s `DEFAULT_REQUEST_TIMEOUT_MS` for every browser-default user) each leave the full 194-test `unit-node` run green. Neutering the whole helper or changing the metadata string *is* caught, so the browser test does its job — these two branches just aren't covered by anything.

Cheap fix that closes both: give the mocked client in `bedrock.test.ts:200-284` a `config.requestHandler`, then add four `unit-node` cases — swap fires when `metadata.handlerProtocol === 'websocket/h1.1'`; no swap for a node-ish protocol; no swap for a caller-provided handler instance *even if its metadata claims websocket* (that's the L1823 branch); and `requestTimeout` survives the swap (`handler.config.requestTimeout === 120_000`). Those four fail exactly on the two mutations above.

strands-ts/src/models/__tests__/bedrock.transport.test.browser.ts
```diff
@@ -0,0 +1,68 @@
...
+      const output = model.stream([new Message({ role: 'user', content: [new TextBlock('Hello')] })], {
+        signal: controller.signal,
+      })
+      const iterator = output[Symbol.asyncIterator]()
+
+      const result = iterator.next()
+      await vi.waitFor(() => expect(fetchMock).toHaveBeenCalledOnce())
+      controller.abort()
+
+      await expect(result).rejects.toMatchObject({ name: 'AbortError' })
```
strandly-the-agent, line 46: 🟡 **This assertion is weaker than it reads.** `FetchHttpHandler.handle()` races the `fetch()` call against its *own* listener on the `abortSignal` option, independently of what the mocked `fetch` does with `request.signal`. I checked with a stub `fetch` that never inspects `request.signal` at all: the promise still rejects with `AbortError`, so the test would pass even if the signal never reached the `Request`.

It does still catch the round-1 regression (pre-fix nothing races at all, so the test hangs and fails), so this is strengthening rather than a defect. One extra line pins the thing this PR is actually about — the signal reaching the transport:

```suggestion
      await expect(result).rejects.toMatchObject({ name: 'AbortError' })
      expect((fetchMock.mock.calls[0]![0] as Request).signal.aborted).toBe(true)
```

strands-ts/src/models/bedrock.ts
```diff
@@ -438,13 +439,15 @@ export class BedrockModel extends Model<BedrockModelConfig> {
...
+    const requestHandler = withDefaultRequestTimeout(clientConfig?.requestHandler)
     this._client = new BedrockRuntimeClient({
       ...(clientConfig ?? {}),
-      requestHandler: withDefaultRequestTimeout(clientConfig?.requestHandler),
+      requestHandler,
       // region takes precedence over clientConfig
       ...(region ? { region: region } : {}),
       customUserAgent,
     })
+    applyAbortAwareBrowserRequestHandler(this._client.config, requestHandler)
```
strandly-the-agent, line 450: 🟡 **This makes `clientConfig.requestHandler` load-bearing for cancellation, and that contract isn't documented anywhere a caller looks.** The explanation lives only in the internal function comment at L1813-1818. Two consequences a browser caller can hit:

1. If they pass their own handler *instance*, the guard correctly leaves it alone (L1823) — so `cancelSignal` / `agent.cancel()` will **not** reach the transport unless their `handle()` forwards `HttpHandlerOptions.abortSignal`. That's a new requirement on caller-supplied handlers, and the failure mode is a silent hang.
2. `requestTimeout` now actually takes effect in browsers. I verified it was silently inert before this PR (the bag went to the WebSocket wrapper while the inner handler was built from `defaultConfigProvider`, which never yields `requestTimeout`), and `updateHttpClientConfig('requestTimeout', 400)` now really times out where it previously never fired. That's a nice win worth stating rather than leaving implicit.

`site/src/content/docs/user-guide/concepts/model-providers/amazon-bedrock.mdx:333` already documents the 120 s default and the instance-vs-options-bag distinction, so it's the natural home for a sentence on both points (plus optionally a line on the `clientConfig` JSDoc at `bedrock.ts:335-338`). Separately: since the one silent-hang case is *detectable* at construction (a caller handler whose `metadata.handlerProtocol` contains `websocket`), would a `warnOnce` there be worth it? The repo already uses that pattern at `bedrock.ts:434,576`, and it turns an unexplainable hang into a diagnosable one.

strandly-the-agent (approved):
✅ **Approve** — my round-1 blocker is fixed, and I confirmed it causally rather than taking the fix on trust.

**Verified at `b119801d`** (clean merge; `git show --remerge-diff` empty, so the merge altered no PR file):
- ✅ Local: `unit-node` focused run **194/194** pass (no type errors) · `build`, `type-check`, `eslint`, `prettier --check`, `check:browser-bundle` all clean
- ✅ Browser repro: real `Agent` + real `BedrockModel` in a browser-resolved bundle of the real pinned `node_modules` (only `globalThis.fetch` stubbed) — all four cases (streaming/non-streaming × external `cancelSignal`/`agent.cancel()`) now abort at the transport, settle `stopReason: 'cancelled'`, release the invocation lock, and the agent is reusable. The **same harness against round-1 head `d65bd58d` still hangs on all four** with the lock held, so the delta is attributable to `applyAbortAwareBrowserRequestHandler`. Mid-stream cancel and a pre-aborted signal also behave.
- ✅ I attacked the new guard across 9 `requestHandler` shapes without breaking it; Node never triggers the swap (`NodeHttp2Handler`/`h2`), the caller's config object is never mutated, and `ConverseStream`'s event-stream body still works through the swapped handler.
- ✅ CI at this head: `CI Gate`, integration tests, unit tests (Node 20/22/24 × ubuntu/windows/macos), Code Quality, Pack Install, Dependency Review, NPM audit, Docs/CI all green; only `Trigger Strands Review` pending.
- ⚠️ **Not** verified: no live Bedrock (so no server-side generation/billing evidence), and no real browser engine — Playwright Chromium cannot launch in my sandbox, so I never ran `npm run test:browser`. Worth having CI's `unit-browser` run be the gate there.

**Maintainer action:** this PR still has no `api/*` label. `jstar0` hit a 403 self-applying `api/needs-review` as an external contributor, so someone with write access needs to add it (then `api/review-complete`). Note `check-api-review-label` is opt-in — it reads green *because* no `api/*` label is present, so its status isn't evidence an API review happened. This is a minimal additive change (one optional field on an existing options interface); PR review should suffice, no API meeting.

Three 🟡 should-fix comments inline — a coverage gap on the new guard, an assertion that's weaker than it reads, and the now load-bearing `clientConfig.requestHandler` contract being undocumented. **None block merge.**

<details>
<summary>Per-pass detail, dependency check, and bundled nits</summary>

**Correctness — no findings.** Reading the pinned `@smithy/core` source, `requestHandler` is re-read live from `this._client.config` on every `.send()`, so the constructor-time swap applies to `Converse`, `ConverseStream` and `CountTokens` alike. Nothing usable is lost by skipping AWS's `defaultsMode` resolution (it only yields `retryMode`/`connectionTimeout`, neither of which `FetchHttpHandler` reads); the import is inert under Node; `updateHttpClientConfig`/`httpHandlerConfigs` are unaffected (different config key); `updateConfig()` never rebuilds the client. Round-1's agent-side signal plumbing is unchanged and still correct on the new base.

**Adversarial — survived.** Matrix and mutation evidence as summarised above. Also confirmed nothing in the TS SDK needs the discarded WebSocket wrapper: `isWebSocketRequest` only matches `ws:`/`wss:`, the bidirectional middleware serves only `InvokeModelWithBidirectionalStreamCommand`, and there is no TypeScript bidirectional/Nova-Sonic consumer; the orphaned wrapper holds no sockets.

**Test quality — gaps found** (inline comments 1 and 2). Neutering the whole helper or changing the metadata string *is* caught by the new browser test; dropping the `handle` early-return or the option-bag pass-through is not.

**Dependency check — clean.** `@smithy/fetch-http-handler` in `dependencies` matches the repo's own boundary rule (the SDK constructs it; consumers never pass an instance), it's already installed and imported transitively by `runtimeConfig.browser.js` so net install/bundle cost is zero, and the single-line lockfile add is equivalent to a `lock:refresh` here (the package has no `os`/`cpu`/optional deps). The cheaper alternatives don't hold up: reusing the wrapper's inner handler needs AWS-private `httpHandler`, and a forwarding shim collapses to the same reach.

**Bundled nits (take or leave):**
- `bedrock.ts:1826` matches the literal `'websocket/h1.1'`; the AWS SDK's own decision on that same field is `.toLowerCase().includes("websocket")`, which would survive an upstream rename to e.g. `websocket/h2`.
- `@smithy/fetch-http-handler@^5.6.4` is now an independent range — worth keeping it bumped alongside the AWS SDK's smithy range so Dependabot can't drift it across a major on its own.
- The abstract `Model.stream` JSDoc (`model.ts:253-261`) still doesn't tell provider implementors to forward `options.signal` — that's the extension point third parties subclass.
- Worth a line in the PR description that this incidentally makes browser `requestTimeout` effective for the first time (it was silently inert under the WebSocket-wrapped default).

**Non-blocking question:** should "the SDK normalises the default browser request handler so signals and timeouts reach the transport" be an SDK-wide browser story rather than Bedrock-local? The same AWS browser wrapper sits under the Bedrock knowledge-base store and S3 session storage; neither forwards a signal today, so nothing is broken — but if transport-level cancellation spreads, this 12-line helper is the shape that would want sharing rather than copying.

</details>

opieter-aws (dismissed):

strands-ts/src/models/model.ts
```diff
@@ -128,6 +128,12 @@ export interface BaseModelConfig {
  * Options interface for configuring streaming model invocation.
  */
 export interface StreamOptions {
+  /**
+   * Optional signal that a provider implementation can forward to abort an in-flight request.
+   * Support is provider-dependent.
+   */
+  signal?: AbortSignal
```
mkmeral: can we update the name to `cancelSignal` to align with naming in rest of the SDK? `StreamOptions.signal` is too generic

mkmeral (changes requested):

strands-ts/src/models/model.ts
jstar0: Updated in `019358a1`: `StreamOptions.signal` is now `cancelSignal` throughout the new cancellation path, including Agent propagation, both Bedrock request modes, and the regression tests. Verified with the focused Node suites (194 tests), browser transport suite (3 tests), and the repository pre-commit suite (3893 tests).

mkmeral:
@strandly-the-agent  are we good to merge here, once CI passes? any blockers or other issues?

strandly-the-agent (approved):
**Yes — no code blockers.** Re-verified at `02c3f142`; the two things holding it are bookkeeping, both in your hands.

The one thing I'd have worried about: the rebase pulled an **AWS SDK bump** into this PR (`client-bedrock-runtime` 3.1078.0 → 3.1095.0, `middleware-websocket` 3.972.39 → **3.972.47**), and the browser fix hinges on an exact-string match against `metadata.handlerProtocol`. I checked the newly resolved package: the literal is still `"websocket/h1.1"`, so the swap still fires, and I re-ran the causal matrix — **4/4 cancellation cases abort and settle `cancelled` with the lock released; restoring the AWS default handler on the same bundle hangs 4/4.** Happy-path streaming/non-streaming still complete normally.

✅ `cancelSignal` rename is complete and consistent (defined once at `model.ts:135`, composed signal passed at `agent.ts:2115`, both Bedrock modes forward it; no stale `signal` references left in the cancellation path) · focused suites **221 passed**, full unit run **3926 passed** · `build`, `type-check`, `eslint`, `prettier`, `check:browser-bundle` all clean · the Chromium test I can't run locally **is** covered in CI (`test:all:coverage` runs `unit-browser` after `test:browser:install`), so the green Node 20/22/24 legs already include it.

**Two gates, neither about the code:**
1. `reviewDecision` is still **CHANGES_REQUESTED** — that's your naming review from 2026-07-29, and `jstar0` addressed it in `fc7b9a79`. Only you can clear it. (My own earlier approval went stale with the force-push; this review re-approves at the current head, though it doesn't satisfy CODEOWNERS.)
2. The label applied is **`needs-api-review`**, but `.github/workflows/api-review-label.yml` only looks at `api/needs-review` / `api/review-complete` — so `check-api-review-label` is green because it sees no `api/*` label and skips. If you want that gate to actually hold this PR, it needs `api/needs-review` → `api/review-complete`; if `needs-api-review` is the label you actually use, the workflow is out of step with practice. Substantively the API review has happened — you asked for the rename and got it.

CI is otherwise mid-flight (Windows legs, Docs, integ, plus the bot gate). ⚠️ Still unverified anywhere: live Bedrock (no server-side generation/billing evidence), and I have no real browser engine locally — CI's `unit-browser` is the gate for that.

<details>
<summary>Leftovers worth a look, none blocking</summary>

My three 🟡 threads from the last round are still open and unaddressed in the tree. All are cheap and none affects behaviour:

1. **No node-side coverage for the guard** (`bedrock.ts:1823-1829`). Still only the Chromium-only test covers it, so the two branches that make it correct — the `handle` early-return and the option-bag pass-through — survive mutation with the whole `unit-node` suite green. Four cheap `unit-node` cases close it (see thread).
2. **The browser test's assertion is weaker than it reads** (`bedrock.transport.test.browser.ts:46`). `FetchHttpHandler.handle` races its own listener on the `abortSignal` option, so the `AbortError` arrives even if the signal never reaches the `Request`. One extra line pins it.
3. **The now load-bearing `clientConfig.requestHandler` contract is undocumented.** A caller-supplied handler must forward `HttpHandlerOptions.abortSignal` or cancellation silently won't reach the transport — `amazon-bedrock.mdx:333` is the natural home.

Also:
- **The hardening nit now has a real-world argument.** `bedrock.ts:1826` uses `!== 'websocket/h1.1'` where upstream's own check on that same field is `.toLowerCase().includes("websocket")`. The SDK bumped twice during this PR's life and the string happened to survive; the next bump might not, and the failure mode is a silent return of the original hang. Cheap to make drift-proof.
- **Release-note-worthy:** this incidentally makes browser `requestTimeout` effective for the first time (it was silently inert under the WebSocket-wrapped default), so browser users may see requests time out where they previously hung.
- The two round-1 threads (`bedrock.ts:632`, `model.ts:131-134`) were addressed back in round 2 and can be marked resolved.

**Main-drift check:** `agent.ts` picked up unrelated interrupt/tool-skip changes from main (#3615). The new `toolsSkipped` + pending-execution `continue` at `agent.ts:1622` cannot strand a cancelled invocation — it returns to the loop head, whose first statement re-throws on cancellation — and the behaviour is byte-identical at the merge-base without this PR, so it's main's, not this PR's.

</details>

aristide1997:
Any updates?

jstar0:
@mkmeral @zastrowm The merge conflict is resolved and the existing PR is rebased onto the current `main`.

The updated head is `32b8b140`. The dependency conflict retained the current AWS SDK version while preserving the Smithy fetch handler required by this fix. Repository checks pass: 142 test files / 3,979 Node tests, build, type-check, lint, formatting, browser bundle, and diff checks. The new CI run is in progress; no new PR or API change was added.

mkmeral (approved):

mkmeral:
hey @jstar0, I just approved and enabled auto-merge, once CI lands, the changes should be merged. 

Thank you for your contribution :)

PR #3339, opened 2026-07-18
fix(agent): bound context-overflow recovery retries

### LINKED ISSUE #3338: ContextWindowOverflowException recovery recurses without convergence check — non-terminating retry loop ending in RecursionError
## Summary

`Agent._execute_event_loop_cycle` handles `ContextWindowOverflowException` by calling `conversation_manager.reduce_context(e=e)` and then **recursing into itself with no attempt counter and no convergence check**. When `SummarizingConversationManager` cannot actually shrink the context, this recursion never terminates on its own — it runs until Python's recursion limit, ending in a `RecursionError` raised from the `copy.deepcopy(agent.messages)` in `_handle_model_execution`.

Observed in production: a single agent task produced **965 consecutive `ContextWindowOverflowException` model calls** (Bedrock ValidationException, 202,271–202,297 tokens — a 26-token band, i.e. the context never shrank between attempts), one cycle every 0.5–2s, terminated only by the `RecursionError`.

Verified against 1.44.0 and 1.48.0 (code identical in both).

## Mechanism

1. `agent.py` — `_execute_event_loop_cycle`: `except ContextWindowOverflowException as e:` → `reduce_context(self, e=e)` → recursive call to `_execute_event_loop_cycle`. No bound, no check that anything shrank.
2. `summarizing_conversation_manager.py` — `_summarize_oldest` hard-protects the last `preserve_recent_messages` (default 10) messages, and `messages_to_summarize_count = max(1, int(len(messages) * summary_ratio))` means summarization always "succeeds" while `len(messages) > preserve_recent_messages` — it will happily re-summarize its own previous summary each round.
3. When the token mass sits **inside** the protected recent messages — e.g. one turn issued many parallel tool calls whose results each landed under the ContextOffloader's `max_result_tokens` but together pushed the conversation past the model window — each round replaces small old messages with a similar-sized summary. Net reduction ≈ 0. Overflow again. Recurse.
4. Each recursion adds nested generator frames; the eventual `RecursionError` surfaces from `copy.deepcopy(agent.messages)` in `_handle_model_execution` (`event_loop.py`), masking the real problem.

Note `reduce_context`'s reactive contract ("MUST remove enough history for the next model call to succeed, or re-raise") is effectively unenforceable for the summarizing manager: it cannot know whether the summary is small enough, and it never re-raises as long as there is at least one unprotected message to fold into the summary.

## Reproduction sketch

- `SummarizingConversationManager()` defaults (`preserve_recent_messages=10`), any model with a hard context limit.
- Construct a conversation where the 10 most recent messages alone exceed the model's context window (e.g. one assistant turn with many toolUse blocks followed by large toolResult messages).
- Invoke the agent. Every cycle: overflow → `_summarize_oldest` folds oldest non-protected message(s) into a summary → total unchanged → overflow. Loop ends only at the recursion limit.

## Suggested fixes (any one of these breaks the loop)

1. Bound the reactive recovery: track consecutive `ContextWindowOverflowException` recoveries per invocation in `_execute_event_loop_cycle` and re-raise after N attempts.
2. Convergence check: after `reduce_context`, compare estimated tokens (the machinery from `_estimate_input_tokens` exists) and re-raise if the reduction made no progress.
3. Make the recursion a loop: the recursion also makes the eventual failure a `RecursionError` from deepcopy instead of the underlying overflow, which is materially harder to diagnose.

Happy to provide more detail from the production trace if useful.

brandon-happyreturns:
## Description

`ContextWindowOverflowException` recovery in `Agent._execute_event_loop_cycle` calls `conversation_manager.reduce_context(e=e)` and recurses into itself with no attempt limit and no convergence check. A conversation manager whose `reduce_context` succeeds without actually shrinking the context retries forever. This is reachable with the default `SummarizingConversationManager`: when the token mass sits inside `preserve_recent_messages` (e.g. one turn's parallel tool results each landed under the ContextOffloader threshold but together exceed the model window), `_summarize_oldest` replaces small old messages with a similar-sized summary — net reduction ~0 — and `max(1, …)` in the split calculation means it never raises while any unprotected message remains. In production we observed 965 consecutive overflow model calls (a 26-token band — the context never shrank) from one agent task, terminated only by a `RecursionError` from the `copy.deepcopy(agent.messages)` in `_handle_model_execution` — masking the real failure.

This change converts the recursion into a bounded retry loop: after `_MAX_CONTEXT_OVERFLOW_RECOVERY_ATTEMPTS` (10) recoveries the original overflow is re-raised, so callers see the actual `ContextWindowOverflowException` instead of a recursion-limit crash. Normal recovery (one reduce, one retry) is unaffected; the loop-instead-of-recursion shape also removes the per-retry generator frame growth.

## Related Issues

Fixes #3338

## Documentation PR

No docs changes — internal recovery behavior; the exception surfaced was already documented as possible.

## Type of Change

Bug fix

## Testing

- New unit test: a manager whose `reduce_context` succeeds without reducing + a model that always overflows → the invocation raises `ContextWindowOverflowException` after exactly the bounded number of recovery attempts (previously: `RecursionError`).
- `tests/strands/agent/`: 718 passed. Broader unit suite passes except modules requiring extras my environment can't build (`cedar`, model-provider SDKs) — failures reproduce identically on a clean checkout.
- `ruff format` / `ruff check` clean on touched files.

- [ ] I ran `hatch run prepare` (hatch env build fails locally on the `cedar` extra's native build; ran the equivalent ruff + pytest directly as above)

## Checklist

----

### DIFF
--- a/strands-py/src/strands/agent/agent.py
+++ b/strands-py/src/strands/agent/agent.py
@@ -133,6 +133,13 @@ _AGENTIC_CONTEXT_MANAGER_MAX_RESULT_TOKENS = 8_000
 """Higher offload threshold for agentic mode - the model manages its own context, so we preserve more inline."""
 
 _CONTEXT_MANAGER_PREVIEW_TOKENS = 750
+
+# Maximum consecutive context-overflow recoveries per invocation. Each recovery
+# calls ``ConversationManager.reduce_context`` and retries the model call; a
+# manager that cannot actually shrink the context (e.g. the token mass sits
+# inside ``SummarizingConversationManager.preserve_recent_messages``) would
+# otherwise retry forever and eventually die with an unrelated RecursionError.
+_MAX_CONTEXT_OVERFLOW_RECOVERY_ATTEMPTS = 10
 """Benchmark-validated preview token count for offloaded results."""
 
 _CONTEXT_MANAGER_SUMMARY_RATIO = 0.3
@@ -1338,7 +1345,12 @@ class Agent(AgentBase):
 
         This internal method handles the execution of the event loop cycle and implements
         retry logic for handling context window overflow exceptions by reducing the
-        conversation context and retrying.
+        conversation context and retrying, bounded by
+        ``_MAX_CONTEXT_OVERFLOW_RECOVERY_ATTEMPTS``. The bound guards against a
+        conversation manager whose ``reduce_context`` succeeds without actually
+        shrinking the context — an unbounded retry would loop until the Python
+        recursion limit and surface as a ``RecursionError`` instead of the
+        underlying overflow.
 
         Args:
             invocation_state: Additional parameters to pass to the event loop.
@@ -1347,6 +1359,10 @@ class Agent(AgentBase):
 
         Yields:
             Events of the loop cycle.
+
+        Raises:
+            ContextWindowOverflowException: If the context still overflows after the
+                maximum number of reduce-and-retry attempts.
         """
         # Add `Agent` to invocation_state to keep backwards-compatibility
         invocation_state["agent"] = self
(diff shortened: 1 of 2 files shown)


JackYPCOnline:
@strandly-the-agent Can you review it first?

strands-py/src/strands/agent/agent.py
```diff
@@ -1355,26 +1371,31 @@ async def _execute_event_loop_cycle(
...
+                        structured_output_context=structured_output_context,
+                        limits=limits,
+                    )
+                    async for event in events:
+                        yield event
+                    return
+
+                except ContextWindowOverflowException as e:
+                    if attempt >= _MAX_CONTEXT_OVERFLOW_RECOVERY_ATTEMPTS:
+                        logger.error(
```
strandly-the-agent, line 1388: 🔴 This cap counts **every** overflow raised anywhere in the invocation, not consecutive no-progress retries. `event_loop_cycle` recursively continues after successful tool handling, so the outer `attempt` survives genuine model/tool progress. Cold repros at this head show both failures: (1) 11 separated incidents execute 11 tools and make 10 real SlidingWindow reductions before this branch raises; (2) one incident shrinks messages `25→23→…→5` across ten reductions, then raises before the needed `5→3` reduction. Both complete on exact base. Please bound consecutive **no-progress** recovery instead—reset the episode after accepted model/tool progress and keep allowing reductions that actually shrink context (a separate emergency ceiling can remain). Regression coverage should succeed after reduction 11 in both the progressive and separated cases.

strandly-the-agent (changes requested):
🔴 **Changes requested:** the invocation-wide retry budget cuts off valid, progressing overflow recovery after ten reductions.

✅ Reviewed exact head `88b32bc` · ✅ agent tests: 150 passed · ✅ Ruff/format/diff checks passed · ✅ 2 cold repros confirmed against head and base

<details><summary>Review passes and verification</summary>

- **Correctness — retained:** the counter wraps the recursively continuing event loop, so valid progressive reductions and later overflow incidents exhaust one shared budget.
- **API/DevX — retained via the same blocker:** behavior contradicts the stated “consecutive” recovery scope. No API-review label is needed because no public symbol, signature, or export changed. Suppressed cap-configurability and logging suggestions as non-blocking.
- **Adversarial — retained:** (1) 11 overflow incidents separated by 11 successful tool executions get only 10 reductions at head; base reduces 11 times and succeeds. (2) one incident shrinks `25→23→…→5`, then head raises before `5→3`; base performs it and succeeds. Boundary, exception, cleanup, hook, and session cases otherwise passed.
- **Test quality — folded into the blocker:** add progressive and tool-separated recovery regressions. Suppressed the production-constant assertion concern as insufficient on its own.
- **Documentation — suppressed:** terminology and detached private-constant wording were lower-value than the behavioral defect and should follow the implementation fix.

CI was still pending when reviewed.
</details>

Issue #1597, opened 2026-01-29
[FEATURE] Configurable retry strategies with exception filtering


nagabharann:
### Problem Statement

Current retry behavior treats all failures as retriable. There is no ability to define specific exceptions or criteria for when to retry versus when to fail immediately. This leads to wasted retries on non-recoverable errors and delayed failure reporting.

### Proposed Solution

Provide configurable retry strategies:

1. Exception allowlist/denylist for retriable errors
2. Custom retry predicates (e.g., should_retry(exception) -> bool)
3. Per-tool or per-model retry configuration
4. Configurable backoff strategies

### Use Case

In production workloads:

1. Retry on transient network errors (ConnectionError, Timeout)
2. Do not retry on authentication failures (403, 401)
3. Do not retry on validation errors (400)
4. Custom retry logic for domain-specific exceptions

### Alternatives Solutions

We implement retry logic in our hooks, but native support would be cleaner and more consistent.

crazybolillo:
I think this issue can be closed by #3340? The use cases are covered by the new option of subclassing and overriding `is_retryable`. Other customization aspects might need their own specific issue 🤔

PR #3340, opened 2026-07-18
feat(py): configurable retry exceptions (#1597)

crazybolillo:
## Description
This lets users define which exceptions should be retried automatically by the framework without having to roll out their own hooks.

Feature parity is improved as well since this is already supported by the Typescript SDK.

## Related Issues

#1597

## Type of Change

New feature

## Testing

## Checklist

----

### DIFF
--- a/strands-py/src/strands/event_loop/_retry.py
+++ b/strands-py/src/strands/event_loop/_retry.py
@@ -21,12 +21,15 @@ logger = logging.getLogger(__name__)
 class ModelRetryStrategy(HookProvider):
     """Default retry strategy for model throttling with exponential backoff.
 
-    Retries model calls on ModelThrottledException using exponential backoff.
+    Retries model calls on retryable exceptions using exponential backoff.
     Delay doubles after each attempt: initial_delay, initial_delay*2, initial_delay*4,
     etc., capped at max_delay. State resets after successful calls.
 
     With defaults (initial_delay=4, max_delay=240, max_attempts=6), delays are:
     4s → 8s → 16s → 32s → 64s (5 retries before giving up on the 6th attempt).
+
+    Subclass and override ``is_retryable`` to expand or narrow the set of
+    retryable exceptions without reimplementing the rest of the retry policy.
 
     Args:
         max_attempts: Total model attempts before re-raising the exception.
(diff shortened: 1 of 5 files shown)


opieter-aws:
Thanks your contribution @crazybolillo! The capability here already exists in the TypeScript SDK, so I’d suggest matching its shape directly to keep the two SDKs in parity.

In TS you don’t pass a `shouldRetry` option, but the strategy has an overridable `isRetryable(error)` method, and you customize by subclassing and calling `super.isRetryable(error)` to extend the set. I’d mirror that exactly in Python.

One heads-up: subclassing `ModelRetryStrategy` is currently rejected in `Agent.__init__` (it uses `type(...) is not ModelRetryStrategy`. To make an override usable you’d need to relax that to `isinstance(...)`.

crazybolillo:
Thanks for the review. I updated the approach to do subclasses, same as Typescript.

strands-py/tests/strands/agent/test_agent_retry.py
```diff
@@ -58,14 +58,13 @@ class FakeRetryStrategy:
...
-    """Test that Agent rejects subclasses of ModelRetryStrategy (strict type check)."""
+def test_agent_accepts_subclass_of_model_retry_strategy():
+    """Test that Agent accepts subclasses of ModelRetryStrategy."""
 
     class CustomRetryStrategy(ModelRetryStrategy):
         pass
 
-    with pytest.raises(ValueError, match="retry_strategy must be an instance of ModelRetryStrategy"):
-        Agent(retry_strategy=CustomRetryStrategy())
+    Agent(retry_strategy=CustomRetryStrategy())  # does not raise
```
opieter-aws, line 67: nit: this test doesn't assert anything. Consider adding an assertion like `isinstance(agent._retry_strategy, CustomRetryStrategy)` or omitting

strands-py/src/strands/event_loop/_retry.py
```diff
@@ -55,6 +58,17 @@ def __init__(
         self._current_attempt = 0
         self._backwards_compatible_event_to_yield: TypedEvent | None = None
 
+    def is_retryable(self, exception: Exception) -> bool:
```
opieter-aws, line 61: nit: Consider adding `call super().is_retryable(...) to extend` to the docstring to make it easier to surface extensibility

opieter-aws (approved):

notowen333 (approved):
This follows our pattern of incrementally moving py towards the ts implementation.

We might look to re-do retries from 0, but this makes sense in any case

PR #3350, opened 2026-07-20
fix(agent): bound context-window overflow recovery retries

### LINKED ISSUE #3338: ContextWindowOverflowException recovery recurses without convergence check — non-terminating retry loop ending in RecursionError

Solaris-star:
## Summary
`Agent._execute_event_loop_cycle` recovered from `ContextWindowOverflowException` by **recursing** with no attempt bound. When `SummarizingConversationManager` cannot actually shrink tokens (mass sits inside `preserve_recent_messages`), this looped until Python's recursion limit and failed with a `RecursionError` from `copy.deepcopy` — masking the real overflow.

Observed in production as hundreds of consecutive overflow model calls (#3338).

## Change
- Replace recursion with an iterative `while` loop
- Cap consecutive overflow recoveries at **10** per invocation
- Re-raise a clear `ContextWindowOverflowException` after the cap

## Test plan
- Existing overflow/sliding-window/null-manager non-infinite-loop tests should still pass
- A conversation manager that never shrinks should fail after ≤10 recoveries instead of RecursionError

Fixes #3338.

### DIFF
--- a/strands-py/src/strands/agent/agent.py
+++ b/strands-py/src/strands/agent/agent.py
@@ -1340,6 +1340,10 @@ class Agent(AgentBase):
         retry logic for handling context window overflow exceptions by reducing the
         conversation context and retrying.
 
+        Recovery is iterative (not recursive) and hard-bounded so a conversation
+        manager that cannot actually shrink the prompt cannot spin forever and
+        eventually surface a RecursionError from deepcopy (see #3338).
+
         Args:
             invocation_state: Additional parameters to pass to the event loop.
             structured_output_context: Optional structured output context for this invocation.
@@ -1354,27 +1358,40 @@ class Agent(AgentBase):
         if structured_output_context:
             structured_output_context.register_tool(self.tool_registry)
 
+        # Bound consecutive ContextWindowOverflow recoveries per invocation.
+        # SummarizingConversationManager can "succeed" without reducing tokens
+        # when the mass sits inside preserve_recent_messages; without a cap the
+        # previous recursive retry loop ran until Python's recursion limit.
+        max_overflow_recoveries = 10
+        overflow_recoveries = 0
+
         try:
-            events = event_loop_cycle(
-                agent=self,
-                invocation_state=invocation_state,
-                structured_output_context=structured_output_context,
-                limits=limits,
-            )
-            async for event in events:
-                yield event
+            while True:
+                try:
+                    events = event_loop_cycle(
+                        agent=self,
+                        invocation_state=invocation_state,
+                        structured_output_context=structured_output_context,
+                        limits=limits,
+                    )
+                    async for event in events:
+                        yield event
+                    break
 
-        except ContextWindowOverflowException as e:
-            # Try reducing the context size and retrying
-            self.conversation_manager.reduce_context(self, e=e)
+                except ContextWindowOverflowException as e:
+                    overflow_recoveries += 1
+                    if overflow_recoveries > max_overflow_recoveries:
+                        raise ContextWindowOverflowException(
+                            f"Context window recovery failed after "
+                            f"{max_overflow_recoveries} attempts without making enough room"
+                        ) from e
 
-            # Sync agent after reduce_context to keep conversation_manager_state up to date in the session
-            if self._session_manager:
-                self._session_manager.sync_agent(self)
+                    # Try reducing the context size and retrying
+                    self.conversation_manager.reduce_context(self, e=e)
 
-            events = self._execute_event_loop_cycle(invocation_state, structured_output_context, limits)
-            async for event in events:
-                yield event
+                    # Sync agent after reduce_context to keep conversation_manager_state up to date in the session
+                    if self._session_manager:
+                        self._session_manager.sync_agent(self)
 
         finally:
             if structured_output_context:


JackYPCOnline:
@strandly-the-agent do we actualy need this feature? The happy path recommand should be using context manager right?

strandly-the-agent:
`JackYPCOnline` — it isn't a feature, it's a non-termination bug, and the context-manager happy path is what *triggers* it rather than what avoids it. **Verdict: accept fixing #3338, redirect this patch** — as written it regresses legitimate recovery.

**Does a conversation manager cover it?** No — the loop exists *because* `reduce_context` is called and returns without shrinking. `SummarizingConversationManager.reduce_context` (`summarizing_conversation_manager.py:116-205`) folds the oldest non-protected messages into one summary and always keeps the last `preserve_recent_messages` (default 10). At `len(messages) == 11` it summarizes exactly one message and replaces it with one summary: length and tokens unchanged, and it does not raise — the `Cannot summarize: insufficient messages` guard (`:163`) only fires at `len <= 10`, which that steady state never reaches. So it "succeeds" forever. `context_manager="auto"` wires exactly this manager plus `ContextOffloader` (`agent.py:552-579`), which is the config in the #3338 production trace; `NullConversationManager`, by contrast, raises immediately and you get a clean error. Still live on `main` (792afbd) at `agent.py:1498-1507` — 965 model calls then `RecursionError`, matching the trace.

**Which layer?** Both, and the manager is the real fix: `reduce_context`'s contract ("remove enough for the next call to succeed, or re-raise") is unenforceable as the summarizer is written, so it should raise when a round frees nothing. An agent-loop bound is still worth keeping as a backstop for third-party managers, and recursion→loop is worth it by itself: today the failure surfaces as `RecursionError` from `deepcopy`, masking the overflow.

**Why not this patch:** the cap is a flat 10 recoveries per invocation with no reset on progress, so it also cuts off valid progressive recovery. A/B with a manager that frees one message per round: base `41f9f59` completes in 19 reductions; PR head `c8746cc` raises after 10. Same defect as #3339 (auto-closed stale).

**Suggested next step:** bound on *lack of progress* (re-raise when a round didn't measurably reduce the estimate) rather than a fixed attempt count, keep the iterative shape, and have the summarizer raise on a no-op round. Happy to implement whichever direction you pick.

<details><summary>Evidence (repros run this session, editable install of <code>strands-py</code>)</summary>

**1. Bug live on `main` `792afbd`** — always-overflow model + a manager whose `reduce_context` succeeds without freeing tokens (the mass-in-`preserve_recent_messages` case):

```
A) default (no limits):   raised RecursionError: maximum recursion depth exceeded  model_calls=965 reductions=965
B) limits={'turns': 5}:   returned stop_reason='limit_turns'  model_calls=5 reductions=5
```

So there is a partial workaround today — `limits={"turns": N}` does bound the loop (each retry re-enters `event_loop_cycle` → `_check_limits`, `event_loop.py:237`; cycles are appended in `metrics.py:271`) — but it's opt-in and returns a normal `AgentResult` with `stop_reason="limit_turns"` and no exception, i.e. the agent silently gives up instead of reporting the overflow. Not a substitute for fixing the loop.

**2. Regression in this patch** — model accepts ≤6 messages; manager frees one message per `reduce_context`; seed 24 messages, so a *correct* recovery needs 19 reductions:

```
base 41f9f59 : OK   stop_reason='end_turn'  reductions=19  model_calls=20
pr3350 c8746cc: FAIL ContextWindowOverflowException: "...failed after 10 attempts..."  reductions=10  model_calls=11
```

Ran against the PR's own base to isolate the patch (`git checkout pr3350^` → same OK/19 as `main`). Real-world shape: overflow incidents spread across a long tool loop share one non-resetting budget, since `event_loop_cycle` recurses internally for tool turns and every incident unwinds to the same handler.

**3. Summarizer steady state** — `summarizing_conversation_manager.py:155-199`: `messages_to_summarize_count = min(max(1, int(len*ratio)), len - preserve_recent_messages)`, so at `len == preserve+1` it is always exactly 1 → 1 message out, 1 summary in → `len` pinned at 11 with no size reduction and no raise. Matches the `final messages=11` in `ebarkhordar`'s repro on #3338.

</details>

Direction is the maintainers' call — this is my read, not a decision. Note also that a scoped second-opinion triage pass I kicked off didn't finish, so the above is my own analysis only.

Issue #2531, opened 2026-03-23
SessionManager: guard against duplicate agent IDs within a session


pgrayy:
## Description

When multiple agents share the same `id` within a single session, their snapshots write to the same storage path and silently overwrite each other's data. This can happen both within a single process (two `new Agent()` with default IDs) and across processes (two separate services using the same `sessionId` with agents that share an `id`).

## Use Case

A developer creates two agents with the same default ID and attaches them to the same session manager. The second agent's snapshots silently overwrite the first's, leading to data loss with no warning.

In a multi-process scenario (e.g., serverless), two separate processes could use the same `sessionId` with agents sharing the same `id`, causing the same silent data corruption.

## Proposed Solution

Add duplicate agent ID detection to `SessionManager`. The in-process case could use an in-memory set, but the cross-process case needs a different approach (e.g., a storage-level lock or uniqueness constraint).

The Python SDK has the same gap: `_latest_agent_message` is in-memory only and provides no cross-process protection.

## Additional Context

- Python SDK reference: `RepositorySessionManager.initialize` checks `agent.agent_id in self._latest_agent_message` but only in-memory
- Both SDKs use agent ID as a storage path component (`scopeId` in TS, `agent_id` path segment in Python), so collisions cause silent overwrites

maxmilian:
I'd like to take this one, scoped to the **in-process guard for the TypeScript SDK**.

The Python side already landed this in `RepositorySessionManager.initialize` (raises `SessionException` when an `agent_id` is re-registered in the same session), but the TS `SessionManager` has no equivalent — `initAgent()` registers hooks without tracking which `agent.id`s it has already seen, so two agents sharing an `id` both resolve to `scopeId: agent.id` and silently overwrite each other's `snapshot_latest`.

Plan: track initialized agent ids in `SessionManager` and `throw new SessionError(...)` on a duplicate in `initAgent()`, mirroring the Python semantics. Tests for both the distinct-id (no throw) and duplicate-id (throws) paths.

I'll leave the **cross-process** case (two processes, same `sessionId`, colliding `agent.id`) out of this PR since, as you noted, that needs storage-level uniqueness/locking rather than an in-memory guard — happy to open a follow-up for it. Sound good?

PR #3363, opened 2026-07-21
fix(session): guard against duplicate agent ids in TS SessionManager

maxmilian:
## What does this PR do?

Adds an in-process duplicate-agent-id guard to the TypeScript `SessionManager`.

When two agents share the same `id` within one session, they both resolve to the same storage path (`scopeId: agent.id`) and silently overwrite each other's `snapshot_latest` — data loss with no warning. This can happen with two `new Agent()` using default ids attached to the same session manager.

The Python SDK already guards this in `RepositorySessionManager.initialize` (raises `SessionException` on a re-registered `agent_id`), but the TS `SessionManager.initAgent()` registered hooks without tracking which ids it had already seen. This mirrors the Python semantics: track initialized agent ids and `throw new SessionError(...)` on a duplicate.

## Scope

- ✅ **In-process** case (this PR): two agents sharing an id on the same `SessionManager` instance now throw instead of silently colliding.
- ⛔ **Cross-process** case (same `sessionId` across processes, colliding `agent.id`) is intentionally left out — as noted in the issue, that needs storage-level uniqueness/locking rather than an in-memory guard. Happy to open a follow-up.

## Type of change

- [x] Bug fix

## Testing

- Added two unit tests in `session-manager.test.ts`: duplicate ids throw `SessionError`; distinct ids don't.
- `vitest run --project unit-node` on the session suite: 52 passed (50 existing + 2 new).
- `tsc --noEmit --project src/tsconfig.json`, `eslint`, and `prettier` clean on the changed files.

Refs #2531

### DIFF
--- a/strands-ts/src/session/__tests__/session-manager.test.ts
+++ b/strands-ts/src/session/__tests__/session-manager.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it, beforeEach, vi } from 'vitest'
 import { SessionManager } from '../session-manager.js'
+import { SessionError } from '../../errors.js'
 import { MockSnapshotStorage, createTestSnapshot } from '../../__fixtures__/mock-storage-provider.js'
 import {
   InitializedEvent,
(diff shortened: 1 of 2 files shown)


JackYPCOnline:
It is expected behavior if two agents have same agent_id and write to the same file path.

PR #3369, opened 2026-07-21
feat(vended-tools): add sleep tool

### LINKED ISSUE #3241: Vended tools: sleep (py + ts)
Part of #3234.

Add a `sleep` vended tool in both SDKs. Pause execution for N seconds, interruptible, with a configurable max. Stdlib only.

chaynabors:
Closes #3241.

Vended sleep tool in both SDKs. Pauses agent execution for a requested number of seconds and cooperates with the SDK's cancellation primitive so cancelling the enclosing invocation aborts the sleep immediately rather than waiting out the full duration. Python shims onto asyncio.sleep, TypeScript onto setTimeout plus the invocation's AbortSignal.

Both expose a factory with a configurable duration cap, defaulting to sixty seconds. The cap is set at construction and cannot be raised by the model. Requests above the cap are rejected at the tool boundary before the sleep starts. Negative, NaN, positive and negative infinity, non-numeric, and boolean durations are also rejected. Stdlib only, no new deps.

Tests exercise the boundary rejections, the factory-level cap checks, and a real five-second cancel that returns in well under a second. The TypeScript suite also covers the short-circuit path for an already-aborted signal and asserts, via a fake timer, that the underlying timer receives the requested duration converted to milliseconds.

### DIFF
--- a/site/src/content/docs/user-guide/concepts/tools/vended-tools-imports.ts
+++ b/site/src/content/docs/user-guide/concepts/tools/vended-tools-imports.ts
@@ -35,3 +35,13 @@ import { bash } from '@strands-agents/sdk/vended-tools/bash'
 import { fileEditor } from '@strands-agents/sdk/vended-tools/file-editor'
 import { notebook } from '@strands-agents/sdk/vended-tools/notebook'
 // --8<-- [end:combined_import]
+
+// --8<-- [start:sleep_import]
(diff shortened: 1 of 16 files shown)


chaynabors:
Closing and reopening as a fresh PR from the same branch (chaynabors:vended-tools/sleep).

Issue #3237, opened 2026-07-14
Vended tools: file_write shim over file_editor (py + ts)


yonib05:
Part of #3234.

Add a `file_write` vended tool as a thin shim over `file_editor` create/write, in both SDKs. Tiny by design; all logic stays in file_editor.

**Done:** shim ships in strands-py and strands-ts with types, tests, README, docs section.

Issue #3235, opened 2026-07-14
Vended tools: reconcile editor command set into file_editor (py + ts)


yonib05:
Part of #3234.

Reconcile the full editor command set into the existing `file_editor` vended tool in both SDKs.

- Compare the richer editor command set (view, create, str_replace, insert, undo, find/pattern ops) against what `file_editor` supports today; fold in anything missing.
- Python: extend `strands.vended_tools.file_editor`. TypeScript: extend `vended-tools/file-editor`.
- No syntax highlighting or console UI inside the tool; SDK hooks own presentation.

**Done:** command sets reconciled in both SDKs, tests for each new command, READMEs and vended-tools.mdx updated.

PR #3371, opened 2026-07-21
docs(vended-tools): explain why file_write is not vended

chaynabors:
This change records the decision not to ship a separate file_write vended tool and points at the migration target. No runtime code is added. The change is two README files under strands-py/src/strands/vended_tools/file_write/ and strands-ts/src/vended-tools/file-write/, plus one prose paragraph in the vended-tools docs page.

The create surface a file_write shim would expose is already vended. The file_editor tool exposes command create with path and file_text, sandbox-routed, with absolute-path validation and parent-directory-traversal rejection. A second entrypoint into the same write path adds an API surface without adding a capability. The create command also deliberately refuses to overwrite, which is a safety property rather than a limitation: overwrite in place erases whatever the agent or the human had there before with no undo, and the vended editor forces the model to see the existing content via view before mutating it with str_replace or insert. A separate file_write that silently overwrote would undo that guard from the outside; one that only created would give the model two names for the same operation.

If an overwrite command is wanted in the vended tool set, it belongs inside file_editor with the same validation and audit surface as create, str_replace, and insert. Two tools sharing the write path means two boundaries to keep tight forever; one tool means one. No owner is chartered for that expansion today, and the related editor reconciliation work at https://github.com/strands-agents/harness-sdk/issues/3235 covers view, create, str_replace, insert, undo, find, and pattern, but not overwrite.

The recommendation is deliberately asymmetric with the parallel read-side sub-issue, and the READMEs spell out why. A read-only shim is a strict subset of file_editor view, so an agent that only needs to read can be given the shim in place of file_editor and the capability set the model sees actually shrinks. A write shim cannot replace file_editor, because any agent that needs to write also needs view, str_replace, and insert to change anything that already exists, so the shim would always ship alongside rather than in place of file_editor. The tool count goes up, the capability set does not, and the model gets two names for the same create path. That difference in shape, not preference, is what drives the different recommendation.

https://github.com/strands-agents/harness-sdk/issues/3237

### DIFF
--- a/site/src/content/docs/user-guide/concepts/tools/vended-tools.mdx
+++ b/site/src/content/docs/user-guide/concepts/tools/vended-tools.mdx
@@ -70,6 +70,20 @@ agent("Create a file at /tmp/hello.txt with the contents 'Hello, world!'")
 
 📖 [Full API Reference](https://github.com/strands-agents/harness-sdk/blob/main/strands-ts/src/vended-tools/file-editor/README.md)
 
+Why there is no separate file-write tool: the create command on
+<Syntax py="file_editor" ts="fileEditor" /> already exposes `path` and `file_text`
+with the same sandbox routing and path validation a file-write shim would need,
+and it refuses to overwrite existing files by design. A sibling file-write that
+silently overwrote would undo that guard; one that only created would give the
+model two names for the same operation. Use <Syntax py="file_editor" ts="fileEditor" />
+with <Syntax py='command="create"' ts="command: 'create'" /> for new files, and
+<Syntax py='command="view"' ts="command: 'view'" /> followed by
+<Syntax py='command="str_replace"' ts="command: 'str_replace'" /> when you need
+to change something that already exists. See the
+[file-write recommendation README](https://github.com/strands-agents/harness-sdk/blob/main/strands-py/src/strands/vended_tools/file_write/README.md)
+for the full reasoning, including why this differs from the parallel file-read
+decision.
+
 ---
 
 ### HTTP Request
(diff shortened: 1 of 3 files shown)


chaynabors:
Fixed in a9a442ff. Both suggestions from the review notes applied:

- Removed the transient function-name/path detail in the Python README (`_handle_create` at file_editor.py). The README now says the size-enforcement gap lives in file_editor's create handler and the fix belongs there, without pinning to a symbol that will move.
- Softened "does ship" to "is being proposed for vending" for the sibling file_read shim in both READMEs, since it is not merged yet.

chaynabors:
Fixed in 9e21f24b. Both blocking items from the second review are addressed, plus the remaining valid Copilot finding on the "What was proposed" wording:

- Em-dashes removed from both READMEs and from the added mdx paragraph. Replaced with colons, commas, or restructured sentences per the voice guide.
- The mdx paragraph now uses Syntax for language-specific identifiers (file_editor / fileEditor and the three command strings) and reads language-neutral in shared prose. Also wrapped to keep the added lines under ninety characters, apart from the necessarily long README URL.
- Reworded "What was proposed" in both READMEs to describe the shim as delegating to file_editor create with an added overwrite semantic on top, so it no longer reads as if a "create-or-write" command already exists in file_editor.

chaynabors:
Fixed in 3b2f927b. All three new Copilot findings addressed:

- Reworded the direct-sandbox note in both READMEs. Now says direct sandbox writes bypass file_editor's "path validation and overwrite protection" rather than "path and content validation". Content validation on create is a known gap, so the previous phrasing overclaimed.
- The docs page link to the recommendation README now offers both language paths (Python and TypeScript) rather than defaulting to the Python one on a multi-language page.

No further /strands review invocations planned. Third invocation approved. Awaiting merge decision from a maintainer.

Issue #3239, opened 2026-07-14
Vended tools: new web_fetch tool (py + ts)


yonib05:
Part of #3234.

New `web_fetch` vended tool in both SDKs: fetch a URL and return clean markdown, no HTML noise. Complements `http_request` (raw API calls) the way Claude Code splits web_fetch from raw HTTP.

Design questions to settle first:
- Markdown extraction approach with a minimal dependency budget, in two languages.
- Redirect/robots/timeout posture.

**Done:** ships in both SDKs with shared behavior, types, tests, README, docs section.

PR #3375, opened 2026-07-21
feat(vended-tools): add web_fetch tool

chaynabors:
Vended web_fetch tool in both SDKs. Complements the http_request tool: use web_fetch when the model needs to read a page, use http_request when it needs the body verbatim for an API call. Performs a single HTTP or HTTPS GET, decodes the body per the response's charset, and returns cleaned markdown along with the extracted title. Non-HTML responses fall through with the decoded body in the markdown field and an empty title, so a caller pointed at a JSON or plain-text URL still gets something usable.

The extractor is hand-rolled in both languages against the standard library only. Python uses html.parser, TypeScript uses a small tokenizer. Scripts, styles, iframes, forms and other active elements are dropped along with data URI images and javascript URLs; headings, lists, links, code, blockquotes and paragraph text are preserved. The parser never executes anything, so event handlers and other active features are dead on arrival. The alternatives (readability-lxml plus lxml in Python, or Readability plus jsdom in TypeScript) each bring a much larger transitive surface and ongoing audit cost, and we do not need general readability-quality extraction to serve the model.

Because the model chooses the target URL, the tool boundary is strict. Only http and https are accepted. Every host is DNS-resolved and every returned address is required to be publicly routable; private, loopback, link-local including cloud metadata endpoints, CGNAT, site-local, unique-local, multicast, and reserved ranges are refused. IPv4-mapped IPv6 addresses are unwrapped before the check so ::ffff:127.0.0.1 does not slip past an IPv6-only check. After validation the tool connects to a pinned already-validated IP rather than letting the socket layer re-resolve, which defeats DNS rebinding between check and connect. HTTPS certificate verification still uses the public hostname via SNI. Every redirect hop goes through the same scheme and address checks, the body is capped at five mebibytes without buffering the excess, and the total request has a wall-clock timeout with a thirty-second default.

Not in scope: robots.txt handling. Robots is a courtesy protocol rather than a security boundary, and applying it inside a low-level fetch belongs in an agent hook or the caller's own gating rather than the tool.

https://github.com/strands-agents/harness-sdk/issues/3239

### DIFF
--- a/site/src/content/docs/user-guide/concepts/tools/vended-tools.mdx
+++ b/site/src/content/docs/user-guide/concepts/tools/vended-tools.mdx
@@ -10,6 +10,8 @@ sourceLinks:
   - path: strands-ts/src/vended-tools/bash/bash.ts
   - path: strands-ts/src/vended-tools/http-request/http-request.ts
   - path: strands-ts/src/vended-tools/notebook/notebook.ts
+  - path: strands-ts/src/vended-tools/web-fetch/web-fetch.ts
+  - path: strands-py/src/strands/vended_tools/web_fetch/web_fetch.py
 ---
 
 Vended tools are pre-built tools included directly in the Strands SDK for common agent tasks like file operations, shell commands, HTTP requests, and persistent notes. 
@@ -34,6 +36,7 @@ Each tool is imported from its own subpath under `@strands-agents/sdk/vended-too
 | [HTTP Request](#http-request) | Make HTTP requests to external APIs | TypeScript (Node.js 20+, browsers) |
 | [Notebook](#notebook) | Manage persistent text notebooks | TypeScript (Node.js, browsers) |
 | [Bash](#bash) | Execute shell commands with persistent sessions | Python, TypeScript (Node.js, Unix/Linux/macOS) |
+| [Web Fetch](#web-fetch) | Fetch a URL and return cleaned markdown for a model to read | Python, TypeScript (Node.js) |
 
 ### File Editor
 
@@ -160,6 +163,47 @@ agent("List all Python files in the current directory and count them")
 ```
 
 📖 [Full API Reference](https://github.com/strands-agents/harness-sdk/blob/main/strands-ts/src/vended-tools/bash/README.md)
+
+---
+
+### Web Fetch
+
+Fetches an HTTP(S) URL and returns its readable content as markdown suitable for a model to read. Scripts, styles, and other non-content noise are stripped from the extracted markdown. Distinct from the HTTP Request tool, which returns raw response bodies for API calls.
+
+_Supported in: Node.js (TypeScript); all platforms (Python)._
+
+:::caution[Security Warning]
+The model chooses the fetched URL, so the tool is intentionally strict at that boundary. Only http and https schemes are accepted. Every host is DNS-resolved and every returned address is required to be publicly routable, so private, loopback, link-local, cloud metadata, CGNAT, multicast, and reserved ranges are refused. Redirects are re-validated against the same rules and the tool connects to a specific already-validated IP address so a DNS rebinder cannot substitute a private address between validation and connect. Response bodies are capped at five mebibytes.
+:::
+
+**Example:**
+
+<Tabs>
+<Tab label="TypeScript">
+
+```typescript
+import { Agent } from '@strands-agents/sdk'
+import { webFetch } from '@strands-agents/sdk/vended-tools/web-fetch'
+
+const agent = new Agent({ tools: [webFetch] })
+await agent.invoke('Summarize https://example.com/blog/post')
+```
+
+</Tab>
+<Tab label="Python">
+
+```python
+from strands import Agent
+from strands.vended_tools import web_fetch
+
+agent = Agent(tools=[web_fetch])
+agent("Summarize https://example.com/blog/post")
+```
+
+</Tab>
+</Tabs>
+
+📖 [Full API Reference](https://github.com/strands-agents/harness-sdk/blob/main/strands-ts/src/vended-tools/web-fetch/README.md)
 
 ---
 
--- a/strands-py/src/strands/vended_tools/web_fetch/_extract.py
+++ b/strands-py/src/strands/vended_tools/web_fetch/_extract.py
@@ -1,0 +1,388 @@
+"""HTML → markdown extraction for the web fetch tool.
+
+The goal is not a perfect readability heuristic — it is producing output that
+is safe and useful for a model to read. We strip script, style, and other
+non-content elements entirely, drop ``data:`` URI images (which can be huge
+blobs), and preserve headings, links, lists, blockquotes, code, and paragraph
+text as GitHub-flavored markdown.
+
+Uses only the standard library (``html.parser``) so this tool adds no runtime
+dependency and no third-party surface to audit.
+"""
+
+from __future__ import annotations
+
+import logging
+from html import unescape
+from html.parser import HTMLParser
+
+logger = logging.getLogger(__name__)
+
+# Elements whose content is discarded entirely, not just unwrapped.
+_DROPPED_ELEMENTS = frozenset(
+    {
+        "script",
+        "style",
+        "noscript",
+        "template",
+        "svg",
+        "canvas",
+        "iframe",
+        "object",
+        "embed",
+        "video",
+        "audio",
+        "form",
+        "input",
+        "button",
+        "select",
+        "textarea",
+        # ``head`` is intentionally NOT dropped -- it contains ``title``, which
+        # we want to extract. Its problematic children (``script``, ``style``)
+        # are already in this set.
+    }
+)
+
+# HTML void elements never fire a matching end tag, so they must not affect
+# the drop-depth counter -- otherwise a void child inside a dropped element
+# would leak the drop state forever (e.g. <form><input></form> would keep the
+# extractor in "dropping" mode after </form>).
+_VOID_ELEMENTS = frozenset(
+    {"area", "base", "br", "col", "embed", "hr", "img", "input", "link", "meta", "param", "source", "track", "wbr"}
+)
+
+_HEADING_LEVELS = {"h1": 1, "h2": 2, "h3": 3, "h4": 4, "h5": 5, "h6": 6}
+
+# Sentinel markers used during extraction to record where a blockquote begins
+# and ends. A post-processing pass prefixes every line between the markers with
+# ``> `` so nested block elements (e.g. ``<blockquote><p>...</p></blockquote>``)
+# stay inside the blockquote in the emitted markdown.
+_BQ_OPEN = "\x00BQ_OPEN\x00"
+_BQ_CLOSE = "\x00BQ_CLOSE\x00"
+
+
+class _MarkdownExtractor(HTMLParser):
+    """Streaming HTML→markdown extractor.
+
+    Not a general HTML-to-markdown converter — good enough for feeding a model.
+    """
+
+    def __init__(self) -> None:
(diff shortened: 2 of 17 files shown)


chaynabors:
Round 2 addressed all reviewer findings in commit fc98f5c8. Summary:

Copilot findings:
- Abort handler TDZ (web-fetch.ts:59): fixed. The listener now attaches after the req is created; also re-checks aborted after attach to close the attach-window race.
- Blockquote loses prefix on nested block content (extract.ts): fixed on both SDKs. Blockquote emits open/close sentinels, a post-processing pass prefixes every line inside them with '> '. Nested blockquotes stack. Added unit coverage in both suites.
- DNS-rebinding test uses TEST-NET-3 (test_web_fetch.py:169): fixed. Now uses 8.8.8.8 (real public) as the first record so the test would fail if the validator short-circuited after one address.

strands-review findings:
- Cross-SDK factory parity (index.ts): fixed. TypeScript now exports makeWebFetch(options) with maxBytes/maxRedirects/name/description, mirroring make_web_fetch. webFetch is a makeWebFetch() default instance.
- TS transport coverage gap (test.node.ts): fixed. Added an http.createServer loopback suite (SSRF stubbed only within that suite) that exercises redirect follow, redirect revalidation, missing Location, redirect cap, size-cap abort, Host-header hygiene, redirect-body short-circuit, and a full-object assertion on the tool result.
- Python private attribute reach (web_fetch.py:266): fixed. Dropped the getattr(agent, '_cancel_signal') path entirely. The wrapper relies on the outer asyncio.wait_for + normal task cancellation now, so no coupling to underscore-prefixed state.
- SSRF_SPEC references dead (ssrf.ts, _ssrf.py): fixed. The dangling section pointers are gone; the denylists now carry inline commentary explaining what each entry covers and note the two SDKs are kept in sync with each other.
- Bare except Exception in _extract.py: fixed. Narrowed the intent by adding a logger.debug with exc_info=True so a real parser regression leaves a trace.
- Assertion granularity on plain-text test: fixed. Now asserts the full deterministic dict.

chaynabors:
Round 3 addressed the new Copilot findings in commit 80f6074b:

- decodeBody didn't parse quoted charset values (web-fetch.ts:218): fixed. The regex now accepts double-quoted, single-quoted, and unquoted forms. Python's parser was already tolerant of quotes.
- Straggler SSRF_SPEC breadcrumb in _ssrf.py:102: fixed. Also swept the TS side; no other references remain.
- fetchOnce / _fetch_once pinned only the first resolved address: fixed on both SDKs. Each attempt now tries the validated addresses in order, retrying on transport-level errors (ECONNREFUSED, EHOSTUNREACH, ENETUNREACH, EADDRNOTAVAIL on TS; OSError family in Python). AbortError and the size cap are terminal and never retried. Every address has already passed the SSRF check so the retry does not weaken the security posture. Coverage added on both suites.

chaynabors:
Round 3 followup on commit 256a655c: the charset regex fix was already in place but no test locked it. Added parity tests on both SDKs -- three charsets (utf-8, iso-8859-1, windows-1252), each in the form the header actually carries -- so any future regression would fail loudly.

site/src/content/docs/user-guide/concepts/tools/vended-tools.mdx
```diff
@@ -163,6 +166,47 @@ agent("List all Python files in the current directory and count them")
 
 ---
 
+### Web Fetch
```
arronbailiss, line 169: a common pattern for web_fetch tools today are that they use a smaller faster model to answer a prompt about a web URL. The web_fetch tool spawns a new Agent with fresh context, that new Agent fetches the content, strips HTML, etc. and then returns a natural language answer. This is instead of returning the full stripped content to the primary Agent, helping to keep the primary Agent's context window leaner. Could we consider that as a mode for the web_fetch tool here as well? How would the interface and consumer configuration for that look? Pros/cons?

strands-ts/src/vended-tools/web-fetch/ssrf.ts
```diff
@@ -0,0 +1,310 @@
...
+import { isIP } from 'net'
+
+const ALLOWED_SCHEMES = new Set(['http:', 'https:'])
+
+// DNS suffix denylist. Common private-network TLDs and anonymized-routing
+// namespaces that must never leak a lookup to the public resolver. Compared
+// case-folded, with any trailing "." stripped. Cloud metadata bare label
+// "metadata" is checked separately in `assertHostIsAllowed`. Kept in sync with
+// the Python implementation in `strands-py/.../_ssrf.py`.
+const DENIED_DNS_SUFFIXES = [
```
arronbailiss, line 23: I wonder if the tool should even be in the business of owning these checks, it feels like this could be applied at a sandboxing / isolation layer. In some cases I do want my agent to be able to fetch trusted URLs with these suffixes

awsarron (changes requested):

strands-py/src/strands/vended_tools/web_fetch/_extract.py
```diff
@@ -0,0 +1,388 @@
+"""HTML → markdown extraction for the web fetch tool.
+
+The goal is not a perfect readability heuristic — it is producing output that
+is safe and useful for a model to read. We strip script, style, and other
+non-content elements entirely, drop ``data:`` URI images (which can be huge
+blobs), and preserve headings, links, lists, blockquotes, code, and paragraph
+text as GitHub-flavored markdown.
+
+Uses only the standard library (``html.parser``) so this tool adds no runtime
```
yonib05, line 9: The description frames this as hand-rolled vs readability-lxml/jsdom, but I don't think we need general readability extraction, so that's not quite the trade. Two narrower options:

The interesting one is `html-to-markdown` from the Kreuzberg folks (https://github.com/xberg-io/html-to-markdown, MIT, v3.9.1 this week). Rust core with both a PyPI package (`requires-python >=3.10`, prebuilt wheels for mac/linux/windows) and npm (`@xberg-io/html-to-markdown`, prebuilt for 8 triples). It's CommonMark compliant, so both SDKs would emit identical markdown, which is the parity problem this PR solves by hand twice. Two caveats: the npm package declares `engines: node >= 22` and we declare `>=20.0.0`, and both bindings are native so a platform without a prebuilt artifact needs a toolchain. There's a wasm crate that might dodge both.

The boring one is `markdownify` + `turndown`. We already depend on `turndown` in `site/package.json` (`site/src/util/html-to-markdown.ts`) so it's vetted here, and no native binaries. Downside is two engines, so output won't match exactly.

Either way I'd make `web_fetch` an optional extra following the `ImportError` pattern in `models/openai_responses.py`. Preference? HTML parsing feels like something we have no business writing ourselves.

strands-ts/src/vended-tools/web-fetch/extract.ts
```diff
@@ -0,0 +1,502 @@
...
+}
+
+interface TextToken {
+  kind: 'text'
+  value: string
+}
+
+type Token = TagToken | TextToken
+
+const NAMED_ENTITIES: Record<string, string> = {
```
yonib05, line 85: TS `NAMED_ENTITIES` has six entries and Python uses `html.unescape`. I ran both on `&rsquo; &ldquo; &trade; &mdash;` and Python renders them while TS passes them through raw to the model. Closing that gap means shipping a 2k-entity table in TS, which is the sort of thing a dep should be doing for us.

strands-ts/src/vended-tools/web-fetch/extract.ts
```diff
@@ -0,0 +1,502 @@
...
+/**
+ * Convert HTML to markdown suitable for a model to read.
+ *
+ * @returns `{ title, markdown }` — both trimmed of surrounding whitespace.
+ */
+// Sentinel markers used during extraction to record where a blockquote begins
+// and ends. A post-processing pass prefixes every line between the markers
+// with `> ` so nested block elements (e.g. `<blockquote><p>...</p></blockquote>`)
+// stay inside the blockquote in the emitted markdown.
+const BQ_OPEN = '\x00BQ_OPEN\x00'
```
yonib05, line 246: `\x00BQ_OPEN\x00` is derived from page content, so can't a page emit that sequence itself and forge blockquote structure? I got TS to emit `> attacker text` from a page with no blockquote, and Python leaks the marker literally instead.

strands-ts/src/vended-tools/web-fetch/ssrf.ts
```diff
@@ -0,0 +1,310 @@
...
+      const parts = expanded.split(':')
+      const high = parseInt(parts[6] ?? '0', 16)
+      const low = parseInt(parts[7] ?? '0', 16)
+      const a = (high >> 8) & 0xff
+      const b = high & 0xff
+      const c = (low >> 8) & 0xff
+      const d = low & 0xff
+      return addressIsPublic(`${a}.${b}.${c}.${d}`)
+    }
+    return ipv6IsPublic(lower)
```
yonib05, line 140: `ipv6IsPublic` is prefix-string matching, and I couldn't get it to block `::7f00:1`, `2002:7f00:1::1`, or `64:ff9b::a9fe:a9fe` (NAT64-wrapped IMDS), all of which Python refuses. Node ships `net.BlockList` and it caught all of them for me in about 25 lines.

strands-py/src/strands/vended_tools/web_fetch/web_fetch.py
```diff
@@ -0,0 +1,305 @@
...
+        """Fetches an HTTP(S) URL and returns readable markdown.
+
+        Only ``http://`` and ``https://`` URLs are accepted. Requests to
+        private, loopback, link-local, multicast, or reserved addresses are
+        refused. Response body size is capped and redirects are revalidated.
+
+        Args:
+            url: The URL to fetch. Must be ``http://`` or ``https://``.
+            timeout: Total wall-clock timeout in seconds (default: 30). Enforced
+                as an outer ``asyncio.wait_for`` so a trickling server cannot
```
yonib05, line 257: The docstring says a trickling server can't hold the request past the cap, but `asyncio.wait_for` cancels the coroutine and not the `to_thread` worker, right? Looks like the socket and thread both outlive the tool call.

strands-py/src/strands/vended_tools/web_fetch/web_fetch.py
```diff
@@ -0,0 +1,305 @@
...
+                # body to a small cap is enough to let the connection close
+                # cleanly without buffering the discarded payload.
+                response.read(_REDIRECT_BODY_DRAIN_CAP)
+                return response.status, headers, b"", url
+            # Read up to max_bytes + 1 so we can detect overflow.
+            body = response.read(max_bytes + 1)
+            if len(body) > max_bytes:
+                raise ValueError(f"Response body exceeded max_bytes={max_bytes}. Refusing to buffer more.")
+            return response.status, headers, body, url
+        except OSError as e:
```
yonib05, line 162: `except OSError` also catches `ssl.SSLCertVerificationError` since it subclasses OSError, so doesn't a cert failure quietly retry the next address and report "could not connect"? Separately, `http_request` doesn't do a multi-address walk at all, what's driving it here?

yonib05 (commented):
TS honors the cancel signal and Python doesn't, so cancelling an agent leaves the fetch running. `http_request.py` already has `_extract_cancel_signal`.

---

Separate from the extractor question: the SSRF and IP-pinning work here is good, and I'd like it to survive as the shared helper rather than get deleted along with the parser. #3372 and #3376 each rolled their own version. Could this one become `vended_tools/_ssrf`?

---

liramon2:
Closed in favor of:
https://github.com/strands-agents/harness-sdk/pull/4001
https://github.com/strands-agents/harness-sdk/pull/4153

Issue #3248, opened 2026-07-14
Vended tools: mcp_client over SDK MCP client (py + ts)


yonib05:
Part of #3234.

Add an `mcp_client` vended tool in both SDKs: agent-callable runtime MCP connections (connect, list_tools, call_tool, disconnect), shimmed over the existing MCP client machinery (`strands.tools.mcp` / `src/mcp`). Not a second client.

- Security posture: connection allowlists, consent on connect, prominent warning. Developer-wired MCP remains the primary path; this tool is the agent-facing one.

**Done:** ships in both SDKs with types, tests, README, docs section.

liramon2:
Update 9/22: Python side is merged. TS port needs review.

PR #3376, opened 2026-07-21
feat(vended-tools): add mcp_client shim

chaynabors:
Adds a vended mcp_client tool in both SDKs. It lets an agent, at runtime, connect to a Model Context Protocol server on a developer-set URL allowlist, list the tools that server exposes, invoke one, and disconnect. The tool is a thin shim over the existing MCP client machinery in each SDK — transport, protocol, and session lifecycle live there — so the surface area unique to this tool is URL validation, the session table, and result truncation.

Four operations are exposed under a single tool name via an op field: connect, list_tools, call_tool, disconnect. Session identifiers are unguessable and scoped to the agent that opened them via a weak reference to the agent object, so another agent using the same tool instance cannot use a session it did not open and a garbage-collected agent's sessions become unreachable. The number of concurrent live sessions across all agents sharing the tool instance is bounded by session_limit (default eight), reserved atomically at connect entry so a burst of concurrent connects cannot overshoot the cap.

The security envelope is deliberately narrow. Only http and https streamable-http servers are supported; stdio, SSE, and WebSocket transports are out of scope. The allowlist is required at construction, must be non-empty, and every connect URL is canonicalised and matched verbatim before any network I/O. Layered on top, an SSRF guard rejects host names on an internal-use suffix denylist before DNS is consulted, then resolves the host and rejects any address in private, loopback, link-local, site-local, multicast, reserved, CGNAT, IPv6 documentation, or discard-only space. IPv4-mapped IPv6 is unwrapped before classification. Cloud-metadata endpoints are named explicitly for defence in depth. Both SDKs refuse to follow HTTP redirects rather than following them, so an allowlisted URL cannot be redirected onto a private endpoint the guard never saw.

DNS rebinding is a known limit and documented as such in the README: the classification is a check-time judgment, and the MCP client SDK does not expose a hook for pinned-IP connects, so operators should only allowlist hostnames whose DNS they control.

https://github.com/strands-agents/harness-sdk/issues/3248

### DIFF
--- a/strands-py/src/strands/vended_tools/mcp_client/mcp_client.py
+++ b/strands-py/src/strands/vended_tools/mcp_client/mcp_client.py
@@ -1,0 +1,630 @@
+"""Agent-callable MCP client tool.
+
+Exposes a single tool with four operations — ``connect``, ``list_tools``, ``call_tool``,
+``disconnect`` — that lets an agent, at runtime, open a connection to a Model Context
+Protocol server, discover its tools, invoke one, and tear the connection down. The tool
+is a thin shim over :class:`~strands.tools.mcp.MCPClient`; all transport, session, and
+protocol handling lives in that class.
+
+**Security model.** This is one of the highest-risk tools to vend: an MCP server can
+implement arbitrary logic, so the model must never be allowed to point the tool at an
+arbitrary URL. The factory takes a developer-set allowlist of exact URLs at construction;
+the tool rejects every ``connect`` that does not match one. In addition:
+
+- Only ``http://`` and ``https://`` URLs are accepted; the tool uses the streamable-http
+  MCP transport. Stdio, SSE, and WebSocket are out of scope: stdio expands the attack
+  surface substantially (subprocess spawn); SSE is being deprecated by the MCP spec in
+  favour of streamable-http; WebSocket is not a documented MCP client transport.
+- The URL's host is rejected outright when it matches a DNS suffix denylist
+  (``.internal``, ``.local``, ``.localhost``, ``.corp``, ``.home``, ``.lan``,
+  ``.intranet``, ``.private``, ``.i2p``, ``.onion``) before any DNS lookup.
+- The URL's host is then resolved and every returned address is checked with
+  :attr:`ipaddress._BaseAddress.is_global` for the private/loopback/CGNAT/shared-
+  address baseline, layered with **explicit** ``is_multicast``, ``is_reserved``,
+  ``is_unspecified``, ``is_link_local``, and IPv6-site-local (``fec0::/10``)
+  rejections: on all supported CPython versions (3.10–3.14) ``is_global`` is
+  ``True`` for IPv4 and IPv6 multicast, and ``fec0::/10`` slips through on some
+  releases, so ``is_global`` alone is not sufficient. IPv4-mapped IPv6
+  (``::ffff:a.b.c.d``, including the fully expanded form) is unwrapped first so
+  the underlying IPv4 category is what gets checked. A short list of
+  cloud-metadata addresses is also spelt out explicitly for defence in depth.
+- The httpx client that MCP uses under the hood is configured with
+  ``follow_redirects=False`` so an allowlisted URL cannot be 3xx'd to a private
+  endpoint the guard never saw.
+- Session IDs are generated with :mod:`secrets` and scoped to the agent that opened
+  them via a :func:`weakref.ref`; another agent using the same tool instance cannot
+  use a session it did not open, and a garbage-collected agent's sessions become
+  unreachable and are dropped from the session table on the next access or connect.
+- The number of live sessions the tool may hold is capped (default: 8), reserved
+  synchronously at connect entry so racing connects cannot overshoot the cap.
+- Blocking MCP client calls (``start``, ``list_tools_sync``, ``stop``) are pushed
+  off the event loop with :func:`asyncio.to_thread` so a slow handshake does not
+  stall other tools.
+- The tool set exposed by the server is fetched once at connect and cached on
+  the session; ``list_tools`` reads that cache and ``call_tool`` validates the
+  requested name against it locally. The cache can go stale if the server
+  changes its tool set mid-session, which is not a supported MCP flow today.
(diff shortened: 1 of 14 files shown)


chaynabors:
Round 2 fixes pushed as 78e1c312. Summary of changes:

Both SDKs: reject URLs carrying userinfo or fragment during allowlist normalisation and connect-time canonicalisation; validate sessionLimit as a positive integer at construction; purge dead-owner sessions before enforcing the cap and reap them again in resolveSession so a stream of connect-and-forget agents cannot pin the cap; canonicalise default scheme ports so `https://host` and `https://host:443` match on the allowlist (Python side; WHATWG URL already normalises this in TypeScript); cache the tool set at connect and validate `tool_name` against it locally in both SDKs so call_tool fails consistently on unadvertised names.

Python only: fixed structured_content `size` to report bytes (the unit the cap is measured in) rather than characters; pushed the blocking `start`, `list_tools_sync`, and `stop` calls off the event loop via `asyncio.to_thread`; mirrored the TypeScript reservation counter so the cap is atomic across a future async migration; added tests for the no-redirect httpx factory, non-JSON-serialisable structured content, unknown tool names, GC-driven purge, canonicalisation of default ports, and userinfo/fragment rejection.

TypeScript only: wrapped the post-`connect()` work in a try/catch that disconnects on failure so a listTools throw does not leak the transport; exported `noRedirectFetch` so a test can assert the shim forces `redirect: 'error'` on the outgoing request even when the caller passes `redirect: 'follow'`; added tests for the connect cleanup path, userinfo/fragment rejection, default-port allowlist matching, and sessionLimit validation.

Docs: converted the TypeScript example on vended-tools.mdx to `--8<--` includes to match the repo convention. Python example stays inline in the mdx.

/strands review

strands-py/src/strands/vended_tools/mcp_client/mcp_client.py
```diff
@@ -0,0 +1,630 @@
...
+  stall other tools.
+- The tool set exposed by the server is fetched once at connect and cached on
+  the session; ``list_tools`` reads that cache and ``call_tool`` validates the
+  requested name against it locally. The cache can go stale if the server
+  changes its tool set mid-session, which is not a supported MCP flow today.
+- Tool-call text and ``structured_content`` are size-capped before being returned to
+  the model.
+
+**Limits.** The public-host classification is a check-time judgment. Between the
+guard's ``getaddrinfo`` and the MCP transport's own ``getaddrinfo``, an
```
yonib05, line 51: The allowlist is mandatory and exact-match, so the model can't reach anything that isn't already approved. What does the extra IP-classification pass add on top of that? As written it also refuses the README's own `https://tools.internal-org.com/mcp` example if that resolves to RFC1918, and there's no `allow_private_hosts` escape hatch like `http_request` has.

Related: #3372 and #3375 each hand-roll their own version of this, and `http_request` on main has a fourth. Could we pull one `vended_tools/_ssrf` out of http_request and import it here?

strands-py/src/strands/vended_tools/mcp_client/mcp_client.py
```diff
@@ -0,0 +1,630 @@
...
+    pin the underlying MCP transport open. Best-effort ``stop`` for the same
+    reason — failures are logged, not raised.
+    """
+    dead: list[str] = [sid for sid, s in sessions.items() if s.owner_ref() is None]
+    for sid in dead:
+        session = sessions.pop(sid, None)
+        if session is None:
+            continue
+        try:
+            session.client.stop(None, None, None)
```
yonib05, line 564: `stop()` joins the background thread, and every other call site in this file wraps it in `asyncio.to_thread` except the purge path. Doesn't that stall the loop for the length of an MCP teardown? Same question on the synchronous `socket.getaddrinfo`, where `http_request` goes through `to_thread`.

strands-py/src/strands/vended_tools/mcp_client/mcp_client.py
```diff
@@ -0,0 +1,630 @@
...
+    session: _Session,
+    tool_name: str,
+    arguments: dict[str, Any] | None,
+    timeout: int,
+) -> dict[str, Any]:
+    if tool_name not in session.tools:
+        raise ValueError(f"Tool {tool_name!r} is not exposed by the connected server")
+
+    tool_use_id = secrets.token_urlsafe(12)
+    result = await session.client.call_tool_async(
```
yonib05, line 482: Python drops the agent cancel signal while TS wires it up, so cancelling leaves a 60s call in flight. Worth mirroring `_extract_cancel_signal`?

strands-ts/src/vended-tools/mcp-client/mcp-client.ts
```diff
@@ -0,0 +1,634 @@
...
+   * Hook for injecting a custom MCP client factory in tests. Not part of the public API;
+   * do not use in production code.
+   * @internal
+   */
+  clientFactory?: (url: string) => McpClient
+  /**
+   * Hook for overriding DNS resolution in tests.
+   * @internal
+   */
+  resolveHost?: (host: string) => Promise<string[]>
```
yonib05, line 79: nit: `clientFactory` and `resolveHost` are marked `@internal` in a doc comment but they're on the exported options interface. Python gets the same coverage with `mock.patch` -- could TS use `vi.mock` and keep them off the public type?

yonib05 (commented):
#3248 lists "consent on connect" as part of the security posture, and what I see is docs pointing developers at HITL themselves. Was that dropped deliberately? Feels like the control that matters most given the model picks the server.

---

The opaque session IDs, weakref ownership checks, quota, and reserved-slot counter are guarding against agent B guessing agent A's token on a tool instance the developer wired for both. Is that a threat we've seen? `bash.ts` does one-session-per-agent with a `WeakMap` + `FinalizationRegistry` in about 15 lines.

---

strands-py/src/strands/vended_tools/mcp_client/mcp_client.py
yonib05: Generally, as we start to dive into this area, I start to question, should we be doing this? Anyway, can we transfer this burden onto the developer side of the shared responsibility model?

PR #3378, opened 2026-07-21
feat(vended-tools): reconcile editor commands into file_editor

chaynabors:
Reconciles the strands-agents/tools editor command set into the existing file_editor vended tool in both SDKs. The tool now exposes seven commands: view, create, str_replace, insert, pattern_replace (regex), find_line, and undo_edit (single-step, in-memory), rather than the previous four. No new editor tool is introduced; the surface simply grows.

The wider win is that every command now routes through a single path resolver (_resolve_path in Python, resolvePath in TypeScript) with optional root confinement. Previously each handler had to remember to call _validate_path, so a future command could have silently skipped it. That funnel is enforced by construction now, and applies both a string-level check (absolute path, no .. segments, inside root) and, when the local filesystem sees the target, a realpath-based symlink check.

Notable calls: symlink confinement is deliberately best-effort at the string level and skipped when root is not visible locally, since a Docker or SSH sandbox's paths do not map to this host and a realpath check would ENOENT on every call. Binary and non-UTF-8 files are rejected up front to prevent mangled output and writes. The size cap defaults to one megabyte, configurable via max_file_size / maxFileSize. Ambiguous str_replace and pattern_replace matches are rejected unless the caller opts in with replace_all, so the model cannot silently rewrite an entire file from a fuzzy match. Undo is a per-tool-instance LRU of pre-edit content, capped by both entry count and byte size, never touching disk. The community-tools editor's syntax highlighting, rich console output, and .bak backup files are deliberately not ported: presentation belongs in SDK hooks, and disk backups do not survive a container restart.

The TypeScript input type is a hand-maintained discriminated union rather than a z.infer, so consumers can narrow on the command literal in a way z.infer would flatten. A comment marks the Zod schema as the single source of truth.

https://github.com/strands-agents/harness-sdk/issues/3235

### DIFF
--- a/site/src/content/docs/user-guide/concepts/tools/vended-tools.mdx
+++ b/site/src/content/docs/user-guide/concepts/tools/vended-tools.mdx
@@ -7,6 +7,7 @@ languages: [python, typescript]
 tags: [tool-authoring]
 sourceLinks:
   - path: strands-ts/src/vended-tools/file-editor/file-editor.ts
+  - path: strands-py/src/strands/vended_tools/file_editor/file_editor.py
   - path: strands-ts/src/vended-tools/bash/bash.ts
   - path: strands-ts/src/vended-tools/http-request/http-request.ts
   - path: strands-ts/src/vended-tools/notebook/notebook.ts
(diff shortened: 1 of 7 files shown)


chaynabors:
Addressed in 3112f36e.

Copilot findings:
- resolvePath trailing-slash strip breaking Windows drive root: fixed. Now uses stripTrailingSep so C:\ survives as-is.
- maxUndoBytes UTF-16 comment: fixed. Comment now says UTF-8 to match the implementation and the Python side.
- str_replace / insert / pattern_replace can write past maxFileSize when the result expands: fixed in both SDKs. Added rejectOversizeResult / _reject_oversize_result before each write, with tests covering expanding replace_all, expanding insert, and expanding pattern_replace.
- README customization example missing makeFileEditor import: fixed.

Strands review findings:
- Important, cross-SDK parity on pattern_replace catastrophic-regex handling: acknowledged and documented explicitly rather than reconciled. TypeScript cannot replicate the Python asyncio.wait_for timeout because V8 has no in-exec regex timeout and JavaScript cannot cancel a running .exec() call. The divergence is now called out in both docstrings and in vended-tools.mdx so a caller is not surprised.
- Important, Python re.error at substitution time leaking past asyncio.wait_for: fixed. Added an except re.error branch that surfaces as a clean ValueError, plus a test with new_str=\9 on a pattern with no group 9.
- Suggestion, empty old_str / empty pattern: fixed. Both now raise a clear "must not be empty" ValueError, with mirrored tests in both SDKs.
- Suggestion, UTF-16 comment drift: fixed alongside the Copilot report.

Not taken:
- The strands review noted the PR meaningfully grows the public API and may warrant a needs-api-review label. The maintainer decides that; noting it here rather than self-labeling.

chaynabors:
Fixed in 0affb8d9: isInsideRoot now handles a root that already ends in a separator ('/' or 'C:\\'). Previously stripTrailingSep preserved those and 'r + sep' produced '//' / 'C:\\\\', rejecting every valid in-root path. Added a regression test for root='/'.

chaynabors:
Windows CI regression fixed in 17f2973d: the new fs-root test skips on Windows since drive letters split roots there. The isInsideRoot fix itself still applies to root='C:\\' via the same trailing-separator branch.

chaynabors:
Closing and reopening as a fresh PR from the same branch.

Issue #3236, opened 2026-07-14
Vended tools: file_read shim over file_editor (py + ts)


yonib05:
Part of #3234.

Add a `file_read` vended tool as a thin shim over `file_editor`'s view/read path, in both SDKs.

- Before writing the shim, compare desired read behavior (chunking, search modes, line ranges) against `file_editor` view; anything view can't do gets added to file_editor first (tracked in the editor reconciliation issue), keeping this shim thin.
- New in both languages.

**Done:** shim ships in strands-py and strands-ts with types, tests, README, docs section.

PR #3381, opened 2026-07-21
feat(vended-tools): add file_read shim over file_editor

chaynabors:
Adds file_read, a read-only vended tool in both SDKs. The tool exposes a two-parameter surface, path and view_range, and delegates every call to file_editor's view command. There is no new validation, no new I/O, and no new size limit: absolute paths, directory-traversal rejection, size caps, view_range bounds, and sandbox routing all come from file_editor unchanged.

The point of vending this alongside file_editor is a real capability reduction for read-only agents. Instead of exposing the full seven-parameter editor schema and hoping the model does not reach for create, str_replace, or insert, the tool schema itself contains only path and view_range. The TypeScript Zod schema is strict, so a smuggled command or file_text field raises a validation error at invocation rather than being silently stripped, and even without strict the callback hardcodes command view. The narrower schema is also better model UX in practice.

Tests in both SDKs cover the happy paths (whole file, view_range slice, negative-one end, directory listing), a delegation probe that shows file_editor's traversal rejection surfacing through the shim, and metadata assertions that the input schema is exactly path and view_range with no write-side fields and no tool_context leakage. The TypeScript suite adds an end-to-end smuggle test that asserts the strict schema rejects a routed write.

Sibling file_write is intentionally not being vended (see #3371). file_read is a strict subset of view and gives read-only agents a narrower attack surface. A file_write shim over file_editor's create and str_replace would double the write surface with no corresponding narrowing, which is why the two decisions diverge.

https://github.com/strands-agents/harness-sdk/issues/3236

### DIFF
--- a/strands-py/src/strands/vended_tools/file_read/file_read.py
+++ b/strands-py/src/strands/vended_tools/file_read/file_read.py
@@ -1,0 +1,92 @@
+"""Read-only file tool.
+
+A minimal, read-only surface over :func:`~strands.vended_tools.file_editor.make_file_editor`:
+the model sees only ``path`` and ``view_range``, and every call is routed through
+``file_editor``'s ``view`` command. All input validation (absolute path, ``..``
+traversal, size limit, ``view_range`` bounds, sandbox probing) is delegated to
+``file_editor``; this tool intentionally adds no new logic and no new checks.
+
+Use this tool when an agent should be able to read files and list directories
+but must not be able to create or edit them.
+
+Note on schema strictness: the TypeScript tool uses a ``.strict()`` Zod schema
+that rejects any key outside ``{path, view_range}`` with a validation error.
+The Python tool relies on Pydantic's default behavior in ``@tool``, which
+silently strips unknown keys. Both paths are safe -- the shim hard-codes
+``command="view"``, so a smuggled ``command`` or ``file_text`` field cannot
+reach ``file_editor``'s write branches -- but the model receives feedback only
+in the TypeScript path. This mirrors the ``@tool`` decorator's convention for
+all vended Python tools and is intentional.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, cast
+
+from ...tools.decorator import tool
+from ...types.tools import ToolContext
+from ..file_editor import make_file_editor
+
+if TYPE_CHECKING:
+    from ...sandbox.base import Sandbox
+    from ...tools.decorator import DecoratedFunctionTool
+
+DEFAULT_FILE_READ_DESCRIPTION = (
+    "Read-only filesystem tool. View a file (with an optional line range) or list a directory. "
+    "Paths must be absolute. For creating or editing files, use `file_editor`."
+)
+"""Description for the file_read tool."""
+
+
+def make_file_read(
+    *,
+    sandbox: Sandbox | None = None,
+    name: str = "file_read",
+    description: str = DEFAULT_FILE_READ_DESCRIPTION,
+) -> DecoratedFunctionTool:
+    """Create a sandbox-routed, read-only file tool.
+
+    A thin shim over ``file_editor``'s ``view`` command with a narrower input
+    schema. If ``sandbox`` is passed, it is bound at creation time. Otherwise
+    the underlying ``file_editor`` reads the sandbox from
+    ``tool_context.agent.sandbox`` at call time.
+
+    Args:
+        sandbox: Sandbox to bind at creation. When ``None``, the agent's
+            configured sandbox is used at call time.
+        name: Tool name. Defaults to ``"file_read"``.
+        description: Tool description shown to the model.
+
+    Returns:
+        A decorated tool that reads files and lists directories through the sandbox.
+    """
+    editor = make_file_editor(sandbox=sandbox)
+
+    @tool(name=name, description=description, context="tool_context")
+    async def file_read_tool(
+        path: str,
+        tool_context: ToolContext,
+        view_range: tuple[int, int] | None = None,
+    ) -> str:
+        """Read a file or list a directory.
+
+        Args:
+            path: Absolute path to a file or directory.
+            tool_context: Injected by the framework. Not user-facing.
+            view_range: Optional line range ``[start, end]`` to view. 1-indexed;
+                ``end`` may be ``-1`` for end-of-file. Not allowed when ``path``
+                is a directory.
+        """
+        result = await editor(
+            command="view",
+            path=path,
+            tool_context=tool_context,
+            view_range=list(view_range) if view_range is not None else None,
+        )
+        return cast(str, result)
+
+    return file_read_tool
+
+
+file_read = make_file_read()
+"""Default read-only file tool. Reads the sandbox from the agent's context at call time."""
--- a/strands-ts/src/vended-tools/file-read/file-read.ts
+++ b/strands-ts/src/vended-tools/file-read/file-read.ts
@@ -1,0 +1,100 @@
+import { tool } from '../../tools/tool-factory.js'
+import { z } from 'zod'
+import { Sandbox } from '../../sandbox/base.js'
+import { makeFileEditor } from '../file-editor/file-editor.js'
+
+/**
+ * Description shown to the model for the file-read tool.
+ */
+export const DEFAULT_FILE_READ_DESCRIPTION =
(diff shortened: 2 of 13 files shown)


strands-py/src/strands/vended_tools/file_read/file_read.py
```diff
@@ -0,0 +1,83 @@
...
+    Returns:
+        A decorated tool that reads files and lists directories through the sandbox.
+    """
+    editor = make_file_editor(sandbox=sandbox)
+
+    @tool(name=name, description=description, context="tool_context")
+    async def file_read_tool(
+        path: str,
+        tool_context: ToolContext,
+        view_range: list[int] | None = None,
```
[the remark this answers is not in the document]
chaynabors: Fixed in 8b94ebc77: narrowed view_range to tuple[int, int] | None. The generated schema now advertises prefixItems with minItems and maxItems of 2 (matching the TS z.tuple schema), and two new tests assert that single-element and three-element ranges are rejected before reaching file_editor.

chaynabors:
Addressed reviewer findings:

1. view_range shape (Copilot inline, py file_read.py:60): Fixed in 8b94ebc77. Narrowed to tuple[int, int] | None so the generated schema now advertises prefixItems with minItems and maxItems of 2, matching the TypeScript z.tuple. Two new tests confirm single-element and three-element ranges are rejected at validation.

2. Cross-SDK parity on smuggled write params (strands-review): Documented as intentional in the module docstring. The @tool decorator's default behavior across all vended Python tools is to silently strip Pydantic extras, and changing that from a single shim would diverge from that convention. Both paths remain safe (the shim hard-codes command=view), and the callback-hardcoded command is the load-bearing defence; the strict schema is defence-in-depth on the TS side only.

3. api/needs-review label (strands-review): I can't apply labels from this account; leaving that for the maintainer to add before merge.

The other strands-review points (clean/well-scoped, PR size fine) needed no action.

JackYPCOnline:
@strandly-the-agent Can you review this PR and check if TS & Python implementations match.

strands-ts/src/vended-tools/file-read/file-read.ts
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * This schema is the single source of truth for the tool's input shape;
+ * `FileReadInput` below is derived from it via `z.infer` so the two cannot
+ * drift.
+ */
+const fileReadInputSchema = z
+  .object({
+    path: z.string().describe('Absolute path to a file or directory.'),
+    view_range: z
+      .tuple([z.number(), z.number()])
```
strandly-the-agent, line 28: 🔴 TypeScript’s range contract still differs from Python’s. At this head, `[1.5, 2.5]` validates and reaches the editor, producing fractional line labels, while Python advertises and enforces an exact pair of integers. The emitted TS JSON Schema also lacks `minItems`/`maxItems`, so it does not communicate the fixed-length contract to the model.

Please make both members integers and ensure the generated schema carries the exact-two constraint, with schema and runtime parity tests. `z.number().int()` addresses fractional acceptance, but the emitted tuple length also needs verification.

strands-ts/src/vended-tools/file-read/file-read.ts
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * const agent = new Agent({
+ *   model: new BedrockModel({ region: 'us-east-1' }),
+ *   tools: [fileRead],
+ * })
+ *
+ * await agent.invoke('Read /tmp/config.json')
+ * ```
+ */
+export function makeFileRead(options?: MakeFileReadOptions): ReturnType<typeof tool>
```
strandly-the-agent, line 74: 🟡 These overload annotations erase the exported `FileReadInput`: the emitted declarations type the factory result and `fileRead` as `InvokableTool<unknown, JSONValue>`, so invalid `invoke` inputs compile without contextual checking.

Could we preserve `FileReadInput` in the return type, or omit that export if callers are not meant to receive a typed invocation surface? The `fileEditor` precedent explains the pattern, but this PR is introducing a new public type that currently is not connected to the tool API.

strands-ts/src/vended-tools/file-read/file-read.ts
```diff
@@ -0,0 +1,100 @@
...
+ * tool intentionally adds no new logic and no new checks.
+ *
+ * If a sandbox is passed, it is bound at creation time. Otherwise, the
+ * underlying `fileEditor` reads the sandbox from `context.agent.sandbox` at
+ * call time.
+ *
+ * @example
+ * ```typescript
+ * import { fileRead } from '@strands-agents/sdk/vended-tools/file-read'
+ * import { Agent } from '@strands-agents/sdk'
```
strandly-the-agent, line 64: 🟡 This example constructs `BedrockModel` but imports only `Agent`, so it does not type-check.

```suggestion
 * import { Agent, BedrockModel } from '@strands-agents/sdk'
```

strandly-the-agent (changes requested):
🔴 **Changes requested** — valid read-only behavior otherwise matches, but TypeScript still accepts fractional line ranges that Python rejects.

✅ Reviewed exact head `8b94ebc7753e967e2fb04da2f26e1a26e87e6158`.
✅ Local: TS focused 9/9 + type-check; Python focused 13/13; `git diff --check`.
✅ CI Gate passed; integration and preview checks remain waiting.
🏷️ Public API review remains open: `api/needs-review` is present; `api/review-complete` is not. No API meeting appears necessary.

<details>
<summary>Specialist-pass breakdown</summary>

### Context build

- Mapped the new factories, singletons, package/barrel exports, underlying editor contracts, sandbox selection, docs, and tests in both SDKs. Valid read-only calls, hard-coded `view`, sandbox routing, and output formatting match conceptually.

### Correctness / parity

- 🔴 Reproduced the blocking mismatch: TS accepts `[1.5, 2.5]` and returns fractional line labels; Python’s `tuple[int, int]` contract rejects it.
- Python’s acceptance of `null`, numeric strings, and booleans is framework-derived leniency and non-blocking. The already accepted strict-extra-field difference was not re-raised.

### Test quality

- Focused suites pass (TS 9/9; Python 13/13), but neither catches the fractional-domain mismatch. Add paired schema/runtime tests proving that both SDKs accept exactly two integer endpoints and reject fractions.
- Existing happy-path, directory, traversal-delegation, metadata, and smuggled-write tests otherwise cover the wrapper’s main behavior.

### Public API / DevX

- 🟡 The exported TS `FileReadInput` is erased by `ReturnType<typeof tool>`; emitted `fileRead` is `InvokableTool<unknown, JSONValue>`. Preserve the input type in the return contract or omit the disconnected export.
- Naming, defaults, factories, and package/barrel exports otherwise align idiomatically. Keep `api/needs-review` until these API findings are resolved and a reviewer signs off.

### Docs accuracy

- 🟡 The TS JSDoc example constructs `BedrockModel` without importing it. The inline suggestion adds the missing import.
- The site and README examples otherwise use the correct public names and imports.

### LLM-facing context

- 🔴 TS exposes unrestricted `number` endpoints and omits `minItems`/`maxItems` in generated JSON Schema, while Python exposes two integer endpoints with exact-length constraints. This is the model-facing side of the reproduced correctness defect: align the emitted schema as well as runtime validation.
- Tool names, descriptions, read-only surface, and successful result text otherwise give equivalent context.

### Routed passes skipped

- **Adversarial / repro pass:** skipped by triage because this is a thin, hard-coded delegation shim with no new parsing, concurrency, or independent I/O/error-path logic; the underlying editor owns and tests that security surface. The correctness pass still ran focused parity repros for the changed schema.
- No other routed specialist pass was skipped.
</details>

strands-py/src/strands/vended_tools/file_read/file_read.py
```diff
@@ -0,0 +1,92 @@
+"""Read-only file tool.
+
+A minimal, read-only surface over :func:`~strands.vended_tools.file_editor.make_file_editor`:
```
yonib05, line 3: The shim constructs its own `make_file_editor(sandbox=sandbox)`, so once #3394 lands, doesn't a `root`-confined deployment still get an unconfined `file_read`? `MakeFileReadOptions` only takes name and description today.

yonib05 (commented):
#3236 asked for a thin shim and the actual delegation is one `command: 'view'` call. Do we need the separate directory, two index files, its own README, and the two-overload `makeFileRead` for that? Wondering if this is ~40 lines inside `file-editor/`.

---

Most of these tests (view_range, -1, directory) re-test `file_editor`'s view path, which already has coverage. Could we keep the schema assertion plus one delegation test and lean on the editor's suite for the rest?

---

Issue #3250, opened 2026-07-14
Vended tools: code_execution over strands.sandbox (py + ts)


yonib05:
Part of #3234.

Add a `code_execution` vended tool in both SDKs, shimmed over `strands.sandbox` (Docker, SSH). The sandbox module carries execution; the tool is spec, session wiring, and result mapping.

Scope: each SDK executes its own language. Python code in strands-py, TypeScript in strands-ts. Not a polyglot language matrix; that keeps each implementation aligned with its runtime and toolchain.

**Done:** ships in both SDKs with types, tests, README, docs section.

liramon2:
Update 9/22: Python needs API bar raising + review. TS port to follow after Python is merged and interface is finalized.

PR #3385, opened 2026-07-21
feat(vended-tools): add code_execution shim

chaynabors:
Adds a code_execution vended tool in both SDKs, shimmed onto Sandbox.execute_code in Python and Sandbox.executeCode in TypeScript. Each call runs a fresh interpreter invocation through the agent's sandbox and returns stdout, stderr, the interpreter exit code, and wall-clock elapsed milliseconds. State does not persist across calls. Language is fixed at factory-time and not exposed on the model-facing schema.

The sandbox module owns isolation. The tool enforces boundaries the sandbox cannot see: it refuses to execute when the agent's sandbox is NotASandboxLocalEnvironment, the host-default whose blunt name signals no isolation. Executing model-authored code there would be a footgun, so the tool raises before ever calling into the primitive. Every input the model controls is capped at the tool boundary. Source code larger than max_code_bytes (default 100 KB) is rejected before touching the sandbox. Standard output and standard error returned to the model are trimmed at max_output_bytes (default 100 KB), with a trailing marker so the model knows truncation happened. Timeout and agent cancellation both surface as an AbortError in TypeScript so callers can branch on error.name, matching the sibling http-request convention.

Scope call: stdin was on the original overlay but is not on the SDK's execute_code primitive. Adding a stdin path only in the tool would either duplicate the sandbox's process-management code or shell-pipe input in around it, both of which are the parallel-implementation pattern the brief calls out to avoid. If stdin becomes a real need the right home is a stdin option on Sandbox.execute_code itself. Language is bound at factory time rather than exposed on the input schema so the tool is not a polyglot execution surface; a caller who wants a different interpreter builds a separate tool via the factory.

Nineteen Python tests and eighteen TypeScript tests cover the security surface first and happy path second, running against a PosixShellSandbox stand-in so the whole path is exercised without provisioning Docker or SSH.

https://github.com/strands-agents/harness-sdk/issues/3250

### DIFF
--- a/strands-py/src/strands/vended_tools/code_execution/code_execution.py
+++ b/strands-py/src/strands/vended_tools/code_execution/code_execution.py
@@ -1,0 +1,181 @@
+"""Code execution tool for running source code through a configured sandbox.
+
+Provides :func:`make_code_execution` (a factory for a sandbox-routed code
+execution tool) and :data:`code_execution` (the default instance that reads the
+sandbox from the agent at call time). Each call runs a fresh interpreter through
+the sandbox; state does not persist across calls.
+
+The tool is a thin shim over :meth:`~strands.sandbox.base.Sandbox.execute_code`.
+The sandbox is the security boundary; the tool refuses to execute if the agent
+has no isolating sandbox configured (i.e. it falls back to
+:class:`~strands.sandbox.not_a_sandbox_local_environment.NotASandboxLocalEnvironment`,
+whose deliberately blunt name signals "no isolation").
+"""
+
+from __future__ import annotations
+
+import math
+import time
+from typing import TYPE_CHECKING
+
+from ...sandbox.errors import SandboxTimeoutError
+from ...sandbox.not_a_sandbox_local_environment import NotASandboxLocalEnvironment
+from ...tools.decorator import tool
+from ...types.tools import ToolContext
+from .types import (
+    CODE_EXECUTION_DESCRIPTION,
+    DEFAULT_LANGUAGE,
+    DEFAULT_MAX_CODE_BYTES,
+    DEFAULT_MAX_OUTPUT_BYTES,
+    DEFAULT_TIMEOUT_SECONDS,
+    TRUNCATION_MARKER,
+    CodeExecutionOutput,
+)
+
+if TYPE_CHECKING:
+    from ...sandbox.base import Sandbox
+    from ...tools.decorator import DecoratedFunctionTool
+
+
+def _truncate(text: str, max_bytes: int) -> str:
+    """Truncate ``text`` to ``max_bytes`` UTF-8 bytes and append a marker if trimmed.
+
+    Truncation is intentionally byte-oriented (not character-oriented) because
+    the cap protects downstream consumers that may be counting tokens or
+    bytes. Decoding with ``errors="ignore"`` drops any incomplete multi-byte
+    sequence at the cut point rather than raising.
+    """
+    encoded = text.encode("utf-8")
+    if len(encoded) <= max_bytes:
+        return text
+    trimmed = encoded[:max_bytes].decode("utf-8", errors="ignore")
+    return trimmed + TRUNCATION_MARKER
+
+
+def make_code_execution(
+    *,
+    sandbox: Sandbox | None = None,
+    language: str = DEFAULT_LANGUAGE,
+    name: str = "code_execution",
+    description: str = CODE_EXECUTION_DESCRIPTION,
+    max_code_bytes: int = DEFAULT_MAX_CODE_BYTES,
+    max_output_bytes: int = DEFAULT_MAX_OUTPUT_BYTES,
+    default_timeout: float = DEFAULT_TIMEOUT_SECONDS,
+) -> DecoratedFunctionTool:
+    """Create a sandbox-routed code execution tool.
+
+    If a ``sandbox`` is passed, it is bound at creation time. Otherwise the tool
+    reads the sandbox from ``tool_context.agent.sandbox`` at call time and
+    refuses to run when that sandbox is the host default
+    (:class:`NotASandboxLocalEnvironment`).
+
+    ``language`` is fixed at factory time -- the model cannot select an
+    interpreter. This keeps the tool aligned with the SDK's own runtime (Python
+    here) and prevents the tool from becoming a polyglot execution surface.
+
+    Args:
+        sandbox: Sandbox to bind at creation. When ``None``, the agent's
+            configured sandbox is used at call time. When it is a
+            :class:`NotASandboxLocalEnvironment`, the tool refuses to execute.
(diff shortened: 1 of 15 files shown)


chaynabors:
Addressed all reviewer findings in 1a11ae2:

TypeScript (code-execution.ts):
- NaN/Infinity cap bypass: factory now rejects non-finite maxCodeBytes, maxOutputBytes, and defaultTimeout with a specific "positive, finite number" message. Timeout on the input schema now enforces `.finite()` as well.
- Cancellation error shape: on SandboxAbortError, cancelSignal.reason is used only when it is an Error subtype; otherwise the tool throws a DOMException with name "AbortError", so callers can rely on the standard Error shape.
- Context-optional when bound: bound-sandbox tools now work without a ToolContext (matches httpRequest); the callback only reads cancelSignal when context is present. Added a test.
- "same convention as http-request" wording: reworded — http-request rewraps as a generic Error, so the honest claim is "callers can distinguish cancellation with error.name === 'AbortError'".
- TextDecoder claim: the comment claimed ignoreBOM: true drops incomplete multi-byte tails; it doesn't. Fixed the comment and made truncate() strip a single trailing U+FFFD so the TS side matches Python's errors="ignore" drop behavior byte-for-byte.
- elapsedMs monotonicity: switched from Date.now() to globalThis.performance.now() so a mid-invocation NTP jump can't produce a misleading or negative elapsed. Matches Python's time.monotonic().

README:
- DockerSandbox import: fixed to @strands-agents/sdk/sandbox/docker (the barrel doesn't re-export it).

Python (code_execution.py):
- NaN/Infinity cap bypass: factory now rejects non-finite / boolean caps and default_timeout with specific messages.
- Timeout finiteness: the per-call timeout check now rejects NaN, +/-Infinity, non-numeric, and booleans at the tool boundary before touching the sandbox.

Docs:
- vended-tools.mdx examples now use the site's --8<-- snippet-include convention (matches the sleep PR pattern).

Not addressed (defended):
- API-review label: this PR follows the sibling vended-tool pattern exactly; happy to add the label if a maintainer prefers.

chaynabors:
Follow-up in 383f327: Python `timeout` docstring now interpolates the factory's `default_timeout` (previously hardcoded as `60`) so a custom-factory instance advertises the correct default in its model-facing schema. Matches the TS side's dynamic `${defaultTimeout}` interpolation. Added a test covering the case.

arronbailiss:
@mkmeral would be good to review

strands-py/src/strands/vended_tools/code_execution/code_execution.py
```diff
@@ -0,0 +1,181 @@
...
+
+Provides :func:`make_code_execution` (a factory for a sandbox-routed code
+execution tool) and :data:`code_execution` (the default instance that reads the
+sandbox from the agent at call time). Each call runs a fresh interpreter through
+the sandbox; state does not persist across calls.
+
+The tool is a thin shim over :meth:`~strands.sandbox.base.Sandbox.execute_code`.
+The sandbox is the security boundary; the tool refuses to execute if the agent
+has no isolating sandbox configured (i.e. it falls back to
+:class:`~strands.sandbox.not_a_sandbox_local_environment.NotASandboxLocalEnvironment`,
```
yonib05, line 11: The isolation gate is `isinstance(active, NotASandboxLocalEnvironment)`, so any other Sandbox subclass clears it. The `_ShellTestSandbox` fixture in this PR spawns local `sh`, passes the gate, and runs model-authored code on the host, which suggests the gate isn't giving us the guarantee the description claims. Should `Sandbox` carry an explicit `provides_isolation` so the sandbox owns the claim rather than the tool guessing from a type?

strands-py/src/strands/vended_tools/code_execution/code_execution.py
```diff
@@ -0,0 +1,181 @@
...
+        # Docstring is set below so the ``timeout`` default in the model-facing
+        # schema reflects the factory's ``default_timeout`` rather than a
+        # hardcoded literal.
+        # Input validation at the tool boundary.
+        if not isinstance(code, str):
+            raise ValueError("code must be a string")
+        code_bytes = len(code.encode("utf-8"))
+        if code_bytes > max_code_bytes:
+            raise ValueError(f"code size ({code_bytes} bytes) exceeds maximum allowed size ({max_code_bytes} bytes)")
+        if timeout is not None:
```
yonib05, line 124: Validation sits behind `if timeout is not None`, so if `None` reaches the callback we hand `timeout=None` to `execute_code`, which the sandbox docs describe as no timeout. Either that's model-controlled unbounded execution or the guard is dead code, do you know which?

strands-ts/src/vended-tools/code-execution/code-execution.ts
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * The tool is a thin shim over `Sandbox.executeCode`. The sandbox is the
+ * security boundary; the tool refuses to execute when the agent falls back to
+ * `NotASandboxLocalEnvironment` (whose name signals "no isolation").
+ */
+
+import { tool } from '../../tools/tool-factory.js'
+import { z } from 'zod'
+import { Sandbox } from '../../sandbox/base.js'
+import { SandboxAbortError, SandboxTimeoutError } from '../../sandbox/errors.js'
```
yonib05, line 17: `SandboxTimeoutError` gets re-thrown as `AbortError` and user cancellation also lands on `AbortError`, so how would a caller branch on `error.name` the way the description suggests? `bash` keeps the typed error, could we do the same?

yonib05 (commented):
`truncate()` runs on `result.stdout` after the pump has already done `stdout += text` with no bound, so a program emitting a couple of GB exhausts memory before the 100KB cap is consulted. Should we cap in the pump loop, or stop describing it as a resource limit?

---

This is 234 Python lines against bash's 70, and the real delta looks like base64-piping an interpreter that the sandbox already does. Could this be a `make_bash`-shaped thin factory?

---

The security tests cover the denylist rather than the boundary, and the happy path runs on the host. Is there an integ test we could add that proves the isolation actually holds?

---

strands-py/src/strands/vended_tools/code_execution/code_execution.py
```diff
@@ -0,0 +1,181 @@
+"""Code execution tool for running source code through a configured sandbox.
```
mkmeral, line 1: I'm not sure if we should have a generic code execution tool. I think it's too generic/meta. From agent's perspective, there is nothing that tells which languages/versions are available, or which libraries. This is a bit of a shot in the dark imo. Also this provides relatively less value on top of bash. 

I think what we should do it focus on specific tools like python_repl instead, where we try to keep track of state and improve the communication of which libraries available and so on.  

This is also why we explicitly did not want to add this tool when we launched sandboxes and bash cc @chaynabors @yonib05

mkmeral (changes requested):

PR #3391, opened 2026-07-21
fix(swarm): correct crash-restart resume for handoff and completed swarms

opieter-aws:
## Description
Fixes handoff for crashed nodes in a swarm.

When a `Swarm` uses a session manager, it checkpoints after every node so a **crashed run** can resume from `next_nodes_to_execute`. The checkpoint fires *before* the execution loop applies a handoff transition, so the saved state was ambiguous: `EXECUTING` with a pending handoff looks identical whether the node committed that handoff or merely inherited it from an earlier turn. Both sides of the contract tried to recover, and every inference had a case it got wrong: a finished node re-ran, a failed run resumed as a success, a cancelled node's speculative handoff was committed, and a pending self-handoff was dropped on the second restart.

This replaces inference with a recorded outcome. Each node turn carries a `_InflightTurn` whose outcome the loop sets where the answer is actually known. A persisted handoff always means its target still owes a turn. The writer settles that at save time. A handoff the resume frontier already carries is not recorded, and a committed turn with no handoff persists an empty frontier. This makes restore honor the payload literally and the reader-side equality check is deleted rather than replaced. 

Two consequences worth noting: 
1. Consumers observing `multiagent_node_stop` now see `node_history` already containing that node, because a completed turn commits before its terminal event is forwarded. A teardown at that yield must not roll back work already recorded. 
2. A checkpoint written before this change replays a turn at worst, it won't skip one.

### Callouts

* A no-handoff completion replays the whole task rather than one node. An empty frontier resets for a fresh run, so a crash at the final node re-runs the task from the entry point instead of re-running that node. The alternative (record the node as done only after the loop) preserves partial progress but introduces some complexity where a stream closed on a succeeded node's terminal event rolls back its handoff and shared context. Both solutions are at-least-once and neither skips work. 

* The root cause of this callout is that Python lets a node hand off to itself while the TS doesn't support this. That divergence is what makes a frontier/target equality ambiguous, and closing it would make the ambiguity unrepresentable rather than merely handled. Changing that behaviour is a bigger change that is breaking for sessions where nodes have done a self-handoff.

## Related Issues

Needed for multi-agent snapshot session manager

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

Ran `hatch test` (full unit suite) plus the multiagent and session suites, `ruff check`, `ruff format --check`, and `mypy` on `swarm.py`.

## Checklist

----

### DIFF
--- a/strands-py/src/strands/multiagent/swarm.py
+++ b/strands-py/src/strands/multiagent/swarm.py
@@ -21,7 +21,7 @@ import sys
 import time
 from collections.abc import AsyncIterator, Callable, Mapping
 from dataclasses import dataclass, field
-from typing import Any, Optional, cast
+from typing import Any, Literal, Optional, cast
 
 from opentelemetry import trace as trace_api
 
@@ -234,6 +234,26 @@ class SwarmResult(MultiAgentResult):
     node_history: list[SwarmNode] = field(default_factory=list)
 
 
+@dataclass
+class _TurnCheckpoint:
+    """The handoff state a node inherited, plus whether its turn committed.
+
+    A checkpoint fires before the execution loop applies a handoff transition, so EXECUTING with a pending
+    handoff is ambiguous on its own: the node may have committed that handoff or merely inherited it.
+
+    Attributes:
+        handoff_node: Handoff target as it stood before the node ran.
+        handoff_message: Handoff message as it stood before the node ran.
+        shared_context: Shared context as it stood before the node ran.
+        outcome: "open" until the turn either commits or is rolled back.
+    """
+
+    handoff_node: SwarmNode | None
+    handoff_message: str | None
+    shared_context: dict[str, dict[str, Any]]
+    outcome: Literal["open", "committed", "rolled_back"] = "open"
+
+
 class Swarm(MultiAgentBase):
     """Self-organizing collaborative agent teams with shared working memory."""
 
@@ -310,6 +330,8 @@ class Swarm(MultiAgentBase):
                 self._plugin_registry.add_and_init(plugin)
 
         self._resume_from_session = False
+
+        self._turn: _TurnCheckpoint | None = None
 
         self._setup_swarm(nodes)
         self._inject_swarm_tools()
@@ -415,6 +437,9 @@ class Swarm(MultiAgentBase):
             interrupts = []
 
             self._invocation_start_time = self.state.start_time
+            # A turn belongs to the invocation that ran it; carrying one over would let a finished
+            # invocation's outcome decide this one's frontier.
+            self._turn = None
 
             try:
                 current_node = cast(SwarmNode, self.state.current_node)
@@ -437,6 +462,8 @@ class Swarm(MultiAgentBase):
                 self.state.completion_status = Status.FAILED
                 raise
             finally:
+                # On stream close this finally runs before _execute_swarm's teardown, so roll back here too.
+                self._rollback_uncommitted_turn()
                 self.state.execution_time = self._commit_active_interval(self.state.execution_time)
                 await self.hooks.invoke_callbacks_async(AfterMultiAgentInvocationEvent(self, invocation_state))
                 self._resume_from_session = False
@@ -629,6 +656,18 @@ class Swarm(MultiAgentBase):
             target_node.node_id,
         )
 
+    def _rollback_uncommitted_turn(self) -> None:
+        """Undo the handoff state an uncommitted turn wrote, so a checkpoint cannot persist it.
+
+        Only an open turn rolls back, so this is safe to call from more than one teardown path.
+        """
+        if self._turn is None or self._turn.outcome != "open":
+            return
+        self.state.handoff_node = self._turn.handoff_node
+        self.state.handoff_message = self._turn.handoff_message
+        self.shared_context.context = self._turn.shared_context
+        self._turn.outcome = "rolled_back"
+
     def _build_node_input(self, target_node: SwarmNode) -> str:
         """Build input text for a node based on shared context and handoffs.
 
@@ -787,6 +826,12 @@ class Swarm(MultiAgentBase):
                         self.state.completion_status = Status.FAILED
                         break
 
+                    self._turn = _TurnCheckpoint(
+                        self.state.handoff_node,
+                        self.state.handoff_message,
+                        copy.deepcopy(self.shared_context.context),
+                    )
+
                     node_stream = self._stream_with_timeout(
                         self._execute_node(current_node, self.state.task, invocation_state),
                         self.node_timeout,
@@ -798,15 +843,23 @@ class Swarm(MultiAgentBase):
                     stop_event = cast(MultiAgentNodeStopEvent, event)
                     node_result = stop_event["node_result"]
                     if node_result.status == Status.INTERRUPTED:
+                        # An interrupt commits the turn: resume continues past it, so the handoff stands.
+                        self._turn.outcome = "committed"
                         yield self._activate_interrupt(current_node, node_result.interrupts)
                         break
 
                     self._interrupt_state.deactivate()
 
                     self.state.node_history.append(current_node)
+                    self._turn.outcome = "committed"
+
+                except (asyncio.CancelledError, GeneratorExit):
+                    self._rollback_uncommitted_turn()
+                    raise
 
                 except Exception:
                     logger.exception("node=<%s> | node execution failed", current_node.node_id)
+                    self._rollback_uncommitted_turn()
                     self.state.completion_status = Status.FAILED
                     break
 
@@ -978,12 +1031,13 @@ class Swarm(MultiAgentBase):
     def serialize_state(self) -> dict[str, Any]:
         """Serialize the current swarm state to a dictionary."""
         status_str = self.state.completion_status.value
-        if self.state.completion_status == Status.EXECUTING and self.state.current_node:
-            next_nodes = [self.state.current_node.node_id]
-        elif self.state.completion_status == Status.INTERRUPTED and self.state.current_node:
-            next_nodes = [self.state.current_node.node_id]
-        elif self.state.handoff_node:
+        # Only a committed turn earns its handoff target the frontier; a handoff standing mid-turn was
+        # inherited, and resuming its target would skip the node that never finished.
+        turn_committed = self._turn is not None and self._turn.outcome == "committed"
+        if self.state.completion_status == Status.EXECUTING and turn_committed and self.state.handoff_node:
             next_nodes = [self.state.handoff_node.node_id]
+        elif self.state.completion_status in (Status.EXECUTING, Status.INTERRUPTED) and self.state.current_node:
+            next_nodes = [self.state.current_node.node_id]
         else:
             next_nodes = []
 
@@ -1012,10 +1066,15 @@ class Swarm(MultiAgentBase):
         """Restore swarm state from a session dict and prepare for execution.
 
         This method handles two scenarios:
-        1. If the payload omits next_nodes_to_execute (a terminal or fresh state), resets all
-           nodes and swarm state to allow re-execution from the beginning.
-        2. Otherwise, restores the persisted state and prepares to resume execution
-           from the next node.
+        1. If the swarm has no resume frontier (next_nodes_to_execute is empty — e.g. completed, or failed
+           at a dead end), resets all nodes and swarm state to allow re-execution from the beginning.
+        2. Otherwise, restores the persisted state and prepares to resume execution from the next node.
+
+        Note:
+            Resume is at-least-once: a crash between a node finishing and the next checkpoint replays that
+            node on restore (e.g. a terminal node whose completion checkpoint never landed), and a resume
+            frontier naming a node this swarm no longer defines re-runs the task from the beginning. Swarm
+            nodes should therefore be idempotent.
 
         Args:
             payload: Dictionary containing persisted state data including status,
@@ -1025,19 +1084,36 @@ class Swarm(MultiAgentBase):
             internal_state = payload["_internal_state"]
             self._interrupt_state = _InterruptState.from_dict(internal_state["interrupt_state"])
 
-        self._resume_from_session = "next_nodes_to_execute" in payload
-        if self._resume_from_session:
-            self._from_dict(payload)
+        # Restored state describes work no in-memory turn ran, so any turn from this process does not apply.
+        self._turn = None
+
+        next_node_ids = payload.get("next_nodes_to_execute") or []
+        # A frontier naming a node this swarm no longer defines is not resumable: the persisted history,
+        # results, and handoff all belong to a topology that no longer exists.
+        if not next_node_ids or next_node_ids[0] not in self.nodes:
+            self._reset_for_fresh_run()
             return
 
+        self._resume_from_session = True
+        self._from_dict(payload)
+
+    def _reset_for_fresh_run(self) -> None:
+        """Discard restored state so the next invocation runs the task from the initial node."""
+        # Clear interrupt state first, else reset_executor_state indexes its context by node_id and
+        # KeyErrors on a never-interrupted node.
+        self._interrupt_state = _InterruptState()
         for node in self.nodes.values():
             node.reset_executor_state()
 
+        # stream_async re-aliases the swarm-owned shared context onto the fresh state, so clear it here or
+        # a previous run's context survives the reset.
+        self.shared_context.context = {}
         self.state = SwarmState(
             current_node=SwarmNode("", Agent(), swarm=self),
             task="",
             completion_status=Status.PENDING,
         )
+        self._resume_from_session = False
 
     def _from_dict(self, payload: dict[str, Any]) -> None:
         self.state.completion_status = Status(payload["status"])
@@ -1047,9 +1123,17 @@ class Swarm(MultiAgentBase):
         self.state.shared_context = self.shared_context
         # Hydrate completed nodes & results
         context = payload["context"] or {}
-        self.shared_context.context = context.get("shared_context") or {}
-        self.state.handoff_message = context.get("handoff_message")
-        self.state.handoff_node = self.nodes[context["handoff_node"]] if context.get("handoff_node") else None
+        # Keyed by node id, so drop entries for nodes this swarm no longer defines rather than rendering a
+        # removed node's context into every prompt.
+        restored_context = context.get("shared_context") or {}
+        self.shared_context.context = {
+            node_id: entries for node_id, entries in restored_context.items() if node_id in self.nodes
+        }
+        handoff_node_id = context.get("handoff_node")
+        self.state.handoff_node = self.nodes.get(handoff_node_id) if handoff_node_id else None
+        # The message is written for the handoff target, so it only survives if that target still does.
+        target_vanished = handoff_node_id is not None and self.state.handoff_node is None
+        self.state.handoff_message = None if target_vanished else context.get("handoff_message")
 
         self.state.accumulated_usage = _parse_usage(payload.get("accumulated_usage") or {})
         self.state.accumulated_metrics = _parse_metrics(payload.get("accumulated_metrics") or {})
@@ -1070,9 +1154,17 @@ class Swarm(MultiAgentBase):
         self.state.results = results
         self.state.task = decode_bytes_values(payload.get("current_task", self.state.task))
 
-        next_node_ids = payload.get("next_nodes_to_execute") or []
-        if next_node_ids:
-            self.state.current_node = self.nodes[next_node_ids[0]] if next_node_ids[0] else self._initial_node()
+        # Only reached for a frontier that names a node this swarm defines.
+        self.state.current_node = self.nodes[payload["next_nodes_to_execute"][0]]
+
+        # Clear a handoff the resume frontier already satisfies so the target is not re-run. Skip INTERRUPTED:
+        # there the frontier is the current node, so a self-handoff equality is coincidental and still pending.
+        if (
+            self.state.completion_status != Status.INTERRUPTED
+            and self.state.handoff_node is not None
+            and self.state.handoff_node == self.state.current_node
+        ):
+            self.state.handoff_node = None
 
     def _initial_node(self) -> SwarmNode:
         if self.entry_point:
--- a/strands-py/src/strands/multiagent/swarm.py
+++ b/strands-py/src/strands/multiagent/swarm.py
@@ -837,21 +837,29 @@ class Swarm(MultiAgentBase):
                         self.node_timeout,
                         f"Node '{current_node.node_id}' execution timed out after {self.node_timeout}s",
                     )
+                    node_result: NodeResult | None = None
                     async for event in node_stream:
+                        # A completed turn commits before its terminal event is forwarded: the node's
+                        # result, metrics, and shared-context writes are already recorded, so a teardown
+                        # arriving at this yield must not roll it back. Only COMPLETED commits here — a
+                        # failing node also emits a stop event, and its raise reaches the handler below.
+                        if isinstance(event, MultiAgentNodeStopEvent):
+                            node_result = event["node_result"]
+                            if node_result.status == Status.COMPLETED:
+                                self._interrupt_state.deactivate()
+                                self.state.node_history.append(current_node)
+                                self._turn.outcome = "committed"
                         yield event
 
-                    stop_event = cast(MultiAgentNodeStopEvent, event)
-                    node_result = stop_event["node_result"]
+                    node_result = cast(NodeResult, node_result)
                     if node_result.status == Status.INTERRUPTED:
-                        # An interrupt commits the turn: resume continues past it, so the handoff stands.
+                        # An interrupt commits the turn — resume continues past the node, so a handoff it
+                        # requested stands — but a commit must never precede the state that makes it true,
+                        # so it follows the activation that records the interrupt.
+                        interrupt_event = self._activate_interrupt(current_node, node_result.interrupts)
                         self._turn.outcome = "committed"
-                        yield self._activate_interrupt(current_node, node_result.interrupts)
+                        yield interrupt_event
                         break
-
-                    self._interrupt_state.deactivate()
-
-                    self.state.node_history.append(current_node)
-                    self._turn.outcome = "committed"
 
                 except (asyncio.CancelledError, GeneratorExit):
                     self._rollback_uncommitted_turn()
@@ -1034,8 +1042,10 @@ class Swarm(MultiAgentBase):
         # Only a committed turn earns its handoff target the frontier; a handoff standing mid-turn was
         # inherited, and resuming its target would skip the node that never finished.
         turn_committed = self._turn is not None and self._turn.outcome == "committed"
+        frontier_carries_handoff = False
         if self.state.completion_status == Status.EXECUTING and turn_committed and self.state.handoff_node:
             next_nodes = [self.state.handoff_node.node_id]
+            frontier_carries_handoff = True
         elif self.state.completion_status in (Status.EXECUTING, Status.INTERRUPTED) and self.state.current_node:
             next_nodes = [self.state.current_node.node_id]
         else:
@@ -1054,7 +1064,17 @@ class Swarm(MultiAgentBase):
             "execution_time": self._execution_time_with_active_interval(self.state.execution_time),
             "context": {
                 "shared_context": getattr(self.state.shared_context, "context", {}) or {},
-                "handoff_node": self.state.handoff_node.node_id if self.state.handoff_node else None,
+                # A persisted handoff always means its target still owes a turn, so a handoff the frontier
+                # already carries is not recorded. Settling that here, where the turn's outcome is known,
+                # is what spares restore the guess: a frontier equals its handoff target both when it
+                # carries that handoff and when a self-handoff is merely pending, and the two are
+                # indistinguishable afterwards.
+                "handoff_node": (
+                    self.state.handoff_node.node_id
+                    if self.state.handoff_node and not frontier_carries_handoff
+                    else None
+                ),
+                # The frontier node is the intended recipient either way.
                 "handoff_message": self.state.handoff_message,
             },
             "_internal_state": {
@@ -1156,15 +1176,6 @@ class Swarm(MultiAgentBase):
 
         # Only reached for a frontier that names a node this swarm defines.
         self.state.current_node = self.nodes[payload["next_nodes_to_execute"][0]]
-
-        # Clear a handoff the resume frontier already satisfies so the target is not re-run. Skip INTERRUPTED:
-        # there the frontier is the current node, so a self-handoff equality is coincidental and still pending.
-        if (
-            self.state.completion_status != Status.INTERRUPTED
-            and self.state.handoff_node is not None
-            and self.state.handoff_node == self.state.current_node
-        ):
-            self.state.handoff_node = None
 
     def _initial_node(self) -> SwarmNode:
         if self.entry_point:
--- a/strands-py/src/strands/multiagent/swarm.py
+++ b/strands-py/src/strands/multiagent/swarm.py
@@ -886,6 +886,9 @@ class Swarm(MultiAgentBase):
 
                     self.state.handoff_node = None
                     self.state.current_node = current_node
+                    # The handoff is applied, so the finished turn no longer speaks for the frontier: the node
+                    # it promoted now owes a turn of its own, and no turn is in flight until that one starts.
+                    self._turn = None
 
                     handoff_event = MultiAgentHandoffEvent(
                         from_node_ids=[previous_node.node_id],
--- a/strands-py/src/strands/multiagent/swarm.py
+++ b/strands-py/src/strands/multiagent/swarm.py
@@ -1043,9 +1043,15 @@ class Swarm(MultiAgentBase):
         # inherited, and resuming its target would skip the node that never finished.
         turn_committed = self._turn is not None and self._turn.outcome == "committed"
         frontier_carries_handoff = False
-        if self.state.completion_status == Status.EXECUTING and turn_committed and self.state.handoff_node:
-            next_nodes = [self.state.handoff_node.node_id]
-            frontier_carries_handoff = True
+        if self.state.completion_status == Status.EXECUTING and turn_committed:
+            # A committed turn owes only its handoff. Without one the swarm is done, so the frontier is
+            # empty rather than the node this turn just finished — re-listing it would make the checkpoint
+            # claim the node is both done and still owed, and each crash at this point would record it again.
+            if self.state.handoff_node:
+                next_nodes = [self.state.handoff_node.node_id]
+                frontier_carries_handoff = True
+            else:
+                next_nodes = []
         elif self.state.completion_status in (Status.EXECUTING, Status.INTERRUPTED) and self.state.current_node:
             next_nodes = [self.state.current_node.node_id]
         else:
--- a/strands-py/src/strands/multiagent/swarm.py
+++ b/strands-py/src/strands/multiagent/swarm.py
@@ -235,8 +235,8 @@ class SwarmResult(MultiAgentResult):
 
 
 @dataclass
-class _TurnCheckpoint:
-    """The handoff state a node inherited, plus whether its turn committed.
+class _InflightTurn:
+    """A node turn in progress: the handoff state it inherited, and whether it committed.
 
     A checkpoint fires before the execution loop applies a handoff transition, so EXECUTING with a pending
     handoff is ambiguous on its own: the node may have committed that handoff or merely inherited it.
@@ -245,7 +245,8 @@ class _TurnCheckpoint:
         handoff_node: Handoff target as it stood before the node ran.
         handoff_message: Handoff message as it stood before the node ran.
         shared_context: Shared context as it stood before the node ran.
-        outcome: "open" until the turn either commits or is rolled back.
+        outcome: Set by the execution loop where the turn's fate is known. A rolled-back turn stays
+            recorded, so the next checkpoint still knows the node owes work.
     """
 
     handoff_node: SwarmNode | None
@@ -331,7 +332,7 @@ class Swarm(MultiAgentBase):
 
         self._resume_from_session = False
 
-        self._turn: _TurnCheckpoint | None = None
+        self._turn: _InflightTurn | None = None
 
         self._setup_swarm(nodes)
         self._inject_swarm_tools()
@@ -826,7 +827,7 @@ class Swarm(MultiAgentBase):
                         self.state.completion_status = Status.FAILED
                         break
 
-                    self._turn = _TurnCheckpoint(
+                    self._turn = _InflightTurn(
                         self.state.handoff_node,
                         self.state.handoff_message,
                         copy.deepcopy(self.shared_context.context),
--- a/strands-py/tests/strands/multiagent/test_swarm.py
+++ b/strands-py/tests/strands/multiagent/test_swarm.py
@@ -1,4 +1,6 @@
 import asyncio
+import shutil
+import tempfile
 import time
 from unittest.mock import ANY, MagicMock, Mock, patch
 
@@ -6,7 +8,7 @@ import pytest
 
 from strands.agent import Agent, AgentResult
 from strands.agent.state import AgentState
-from strands.hooks import BeforeNodeCallEvent
+from strands.hooks import AfterMultiAgentInvocationEvent, AfterNodeCallEvent, BeforeNodeCallEvent
 from strands.hooks.registry import HookRegistry
 from strands.interrupt import Interrupt, _InterruptState
 from strands.multiagent.base import Status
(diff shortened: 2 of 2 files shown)


mkmeral:
@strandly-the-agent  review pls

strands-py/src/strands/multiagent/swarm.py
```diff
@@ -976,12 +976,14 @@ def _build_result(self, interrupts: list[Interrupt]) -> SwarmResult:
...
-        if self.state.completion_status == Status.EXECUTING and self.state.current_node:
-            next_nodes = [self.state.current_node.node_id]
-        elif self.state.completion_status == Status.INTERRUPTED and self.state.current_node:
+        if self.state.completion_status == Status.INTERRUPTED and self.state.current_node:
             next_nodes = [self.state.current_node.node_id]
         elif self.state.handoff_node:
+            # AfterNodeCallEvent checkpoints inside the finally, before the handoff transition sets
+            # current_node = handoff_node. Prefer the handoff target so resume runs it, not the completed node.
             next_nodes = [self.state.handoff_node.node_id]
+        elif self.state.completion_status == Status.EXECUTING and self.state.current_node:
```
strandly-the-agent: 🔴 **The pre-transition checkpoint still cannot encode a reliable resume frontier.** `AfterNodeCallEvent` saves before the handoff/completion transition, so this inference has three reproducible bad outcomes:

- At the terminal checkpoint, I observed `status=executing`, `history=['terminal']`, `frontier=['terminal']`; restore produced `['terminal', 'terminal']`.
- If a source requests `B` and then fails, the durable state is `failed` with frontier `B`; restore runs `B` and returns `COMPLETED` instead of preserving failure.
- A merge-base payload (`history=['A']`, frontier `A`, handoff `B`) read by this head produced `A,A,B`.

The added handoff test only checks `snapshots[0]`; checking its terminal `snapshots[1]` reproduces the first case. Could we persist after the transition, or encode a versioned checkpoint phase/outcome rather than derive it from this still-executing state? Please cover terminal, handoff-then-failure, and old-writer/new-reader checkpoints.

strands-py/src/strands/multiagent/swarm.py
```diff
@@ -1020,19 +1021,22 @@ def deserialize_state(self, payload: dict[str, Any]) -> None:
             internal_state = payload["_internal_state"]
             self._interrupt_state = _InterruptState.from_dict(internal_state["interrupt_state"])
 
-        self._resume_from_session = "next_nodes_to_execute" in payload
-        if self._resume_from_session:
-            self._from_dict(payload)
-            return
+        if not payload.get("next_nodes_to_execute"):
+            # A completed (or dead-end failed) swarm has no resume frontier — serialize_state still emits the
```
strandly-the-agent: 🔴 **Resetting an empty frontier can apply stale interrupt context and raise `KeyError`.** This method first installs the persisted active `_interrupt_state`, then calls `reset_executor_state()` for every node; that reset indexes interrupt context by node ID.

I reproduced the reachable sequence: `A` interrupts → resumed `A` fails before interrupt deactivation → final snapshot is `FAILED`/empty-frontier with context only for `A` → restoring into `[A, B]` raises `KeyError: 'B'` here. A one-node restore also remains interrupt-active rather than becoming fresh.

Could the branch decision happen before installing persisted interrupt state? On an empty frontier, clear/deactivate it and reset from each node's initial state; hydrate interrupt context only for a genuine resume. A multi-node failed-post-interrupt regression test would cover this.

strands-py/src/strands/multiagent/swarm.py
```diff
@@ -1061,6 +1065,12 @@ def _from_dict(self, payload: dict[str, Any]) -> None:
         if next_node_ids:
             self.state.current_node = self.nodes[next_node_ids[0]] if next_node_ids[0] else self._initial_node()
 
+        # The after-node checkpoint persists the handoff target as both the pending handoff and the resume
+        # frontier; clearing the now-satisfied handoff keeps the target from being handed off to (and re-run)
+        # a second time after it executes on resume.
+        if self.state.handoff_node is not None and self.state.handoff_node == self.state.current_node:
+            self.state.handoff_node = None
```
strandly-the-agent: 🔴 **Equality does not prove the handoff was already consumed.** Self-handoff is currently accepted. If `A` requests `A` and then interrupts, the checkpoint legitimately contains frontier `A` and handoff `A`, but this clears that still-pending handoff.

My repro showed in-memory interrupt resume executing resumed `A` and then its self-handoff, while crash resume executed `A` only once. Could this cleanup be gated by an explicit completed-source checkpoint phase (or self-handoff be rejected consistently)? Please add an interrupt/crash-resume self-handoff test so persisted and in-memory behavior agree.

strandly-the-agent (changes requested):
**🔴 Changes requested — checkpoint restore can still replay/skip work, and one reachable reset path raises `KeyError`.**

✅ Reviewed exact head `e13b2bed` against merge base `3966e703`; no public API change.
✅ Latest completed CI is green across Python 3.10–3.14, lint, docs, and integration.
✅ Independent exact-head adversarial matrix reproduced the findings twice identically; Swarm 39/39, changed-file Ruff, mypy (223 files), and `git diff --check` passed.
⚠️ The advanced adversarial tier was unavailable under the configured data-retention policy; the retry completed on the default tier.

<details>
<summary>Review breakdown</summary>

Three blockers remain:

1. The pre-transition checkpoint cannot distinguish terminal success, failed handoff, and legacy frontier semantics, causing replay or turning failure into success.
2. Empty-frontier restore installs active interrupt state before resetting every node, which can raise `KeyError`.
3. Unconditional handoff/frontier equality cleanup drops accepted self-handoffs during crash resume.

The terminal-checkpoint test gap is folded into finding 1. I suppressed the pre-existing second-crash shared-context issue, and did not repeat the existing test-docstring suggestion. Docs were otherwise accurate.

**Passes:** correctness · adversarial · test quality · docs accuracy
</details>

strands-py/src/strands/multiagent/swarm.py
opieter-aws: Confirmed, fixed, and test added

strands-py/src/strands/multiagent/swarm.py
opieter-aws: fixed by gating the handoff-clear on `!= INTERRUPTED`

strands-py/src/strands/multiagent/swarm.py
opieter-aws: Guarded the branch on `current_node not in node_history` (mirroring how Graph excludes completed nodes)

opieter-aws:
@strandly-the-agent re-review with the latest fixes

strands-py/src/strands/multiagent/swarm.py
```diff
@@ -976,13 +976,24 @@ def _build_result(self, interrupts: list[Interrupt]) -> SwarmResult:
...
+            # AfterNodeCallEvent checkpoints inside the finally, before the handoff transition sets
+            # current_node = handoff_node. Prefer the handoff target so resume runs it, not the completed node.
+            # Gate on EXECUTING so a node that requests a handoff and then fails serializes an empty frontier
+            # (resets on restore) rather than resuming the target and masking the failure as a completed run.
             next_nodes = [self.state.handoff_node.node_id]
+        elif (
+            self.state.completion_status == Status.EXECUTING
+            and self.state.current_node
+            and self.state.current_node not in self.state.node_history
+        ):
```
strandly-the-agent: 🔴 **Cumulative history membership drops a pending revisit.**

The exact-head repro runs `A → B → A`, receives the next `A → B` handoff, and then closes the stream. The invocation-final checkpoint has `status=executing`, current node `B`, history `[A, B, A]`, and no pending handoff. Because `B` appeared earlier in cumulative history, this condition emits an empty frontier; restore takes the reset path and starts at `A` instead of running pending `B` (`A=1, B=0`, history `[A]`).

Could we represent the current occurrence's checkpoint phase/frontier explicitly rather than infer it from whether that node ID appeared in any earlier occurrence? A regression that closes immediately after handing off to a previously visited node should assert restore runs that target next.

strands-py/src/strands/multiagent/swarm.py
```diff
@@ -976,13 +976,24 @@ def _build_result(self, interrupts: list[Interrupt]) -> SwarmResult:
...
         """Serialize the current swarm state to a dictionary."""
         status_str = self.state.completion_status.value
-        if self.state.completion_status == Status.EXECUTING and self.state.current_node:
+        if self.state.completion_status == Status.INTERRUPTED and self.state.current_node:
             next_nodes = [self.state.current_node.node_id]
-        elif self.state.completion_status == Status.INTERRUPTED and self.state.current_node:
-            next_nodes = [self.state.current_node.node_id]
-        elif self.state.handoff_node:
+        elif self.state.completion_status == Status.EXECUTING and self.state.handoff_node:
+            # AfterNodeCallEvent checkpoints inside the finally, before the handoff transition sets
```
strandly-the-agent: 🔴 **Cancellation can serialize an uncommitted handoff target.**

`asyncio.CancelledError` bypasses the surrounding `except Exception`, while the `finally` still invokes `AfterNodeCallEvent`. If source `A` requested `B` before cancellation, the checkpoint still has `status=executing`, `history=[]`, and `handoff_node=B`, so this branch persists `B`. The exact-head repro restored with `A=0, B=1`; retrying the same object reran unfinished `A` and did not run `B`.

Cancellation checkpointing should not commit a handoff before its source completes: preserve the source as pending (or roll back to the last committed phase) while still propagating cancellation. Please add a real `CancelledError` regression after a handoff request; the current `RuntimeError` test enters `FAILED` and misses this path.

strandly-the-agent (changes requested):
**🔴 Changes requested — two exact-head crash-resume paths can still silently skip pending Swarm work.**

✅ Re-reviewed exact head `04332244` against prior head `e13b2bed` and merge base `3966e703`; no public API change.
✅ Exact-head CI is fully green; local focused run passed 45/45, changed-file Ruff passed, and `git diff --check` passed.
✅ The new interrupt reset, terminal checkpoint, ordinary failed handoff, and interrupted self-handoff fixes are resolved.
⚠️ The advanced adversarial tier was unavailable under the data-retention policy; the default-tier retry completed and its two deterministic repros were rerun independently.

<details>
<summary>Review breakdown</summary>

Two blockers remain:

1. A handoff to a previously visited node can be serialized with an empty frontier, so restore starts over instead of running the pending target.
2. Cancellation after requesting a handoff can serialize the target even though the source did not complete, so restore skips unfinished work.

I did not reopen the accepted/documented old-writer compatibility limitation, and I suppressed fresh-run behavior after an empty frontier because that is the intended Graph-compatible contract.

**Passes:** correctness · adversarial · test quality
</details>

strands-py/src/strands/multiagent/swarm.py
opieter-aws: Confirmed and fixed. The terminal node re-running on restore is an accepted at-least-once cost, now documented in `deserialize_state`

strands-py/src/strands/multiagent/swarm.py
opieter-aws: Confirmed and fixed

opieter-aws:
@strandly-the-agent re-review

strands-py/src/strands/multiagent/swarm.py
```diff
@@ -803,6 +803,13 @@ async def _execute_swarm(self, invocation_state: dict[str, Any]) -> AsyncIterato
 
                     self.state.node_history.append(current_node)
 
+                except asyncio.CancelledError:
+                    # Cancellation before the node completed must not commit a handoff it requested mid-turn:
```
strandly-the-agent: 🔴 **Closing the public stream can checkpoint the target of an unfinished handoff.**

After `stream_async()` yields an inner event, consumer `aclose()` injects `GeneratorExit` at the outer yield, so it bypasses this `CancelledError`-only rollback. The outer invocation `finally` still persists state, and `serialize_state()` selects requested target B even though source A never completed.

With a real `FileSessionManager`, both exact-head runs persisted `history=[]`, frontier `[B]`, handoff `B`; fresh restore invoked `A=0, B=1`. Could generator close/premature stream termination go through the same rollback without swallowing it? A regression should `aclose()` the public stream after A requests B, reload from disk, and prove A reruns.

strands-py/src/strands/multiagent/swarm.py
```diff
@@ -803,6 +803,13 @@ async def _execute_swarm(self, invocation_state: dict[str, Any]) -> AsyncIterato
...
+                    # clearing it keeps the after-node checkpoint's frontier on the uncompleted source so
+                    # resume re-runs that node, not the target. Re-raise — cancellation is never swallowed.
+                    self.state.handoff_node = None
+                    raise
```
strandly-the-agent, line 867: 🔴 **Cancellation rolls back the target but not the rest of the speculative handoff.**

`_handle_handoff()` also writes `handoff_message` and shared context, but this clears only `handoff_node`. Those values are persisted and restored; A is selected correctly, then receives B-directed handoff text and partial context in its retry input.

This reproduced twice with a real `FileSessionManager`. Could the cancellation path roll back the complete turn-attributable handoff atomically—including message and context additions—while preserving older durable context? The new test currently stops after checking `current_node`; invoking the restored swarm and inspecting A's actual input would catch this.

strandly-the-agent (changes requested):
**🔴 Changes requested — two cancellation paths can still persist inconsistent handoff state.**

✅ Re-reviewed exact head `c05970ee` against prior `04332244`; the pending-revisit fix works.
✅ Exact-head local checks: 46 focused tests; specialists also ran Swarm 44/44 and `FileSessionManager` 40/40; `git diff --check` passed.
✅ Both remaining real `FileSessionManager` crash/restore repros were confirmed twice.
⚠️ Exact-head unit/lint/docs/CI Gate are green, but the separate integration job failed; its logs aren't exposed through the API. Advanced adversarial was unavailable under the data-retention policy, so the pass completed on the default-tier retry.

<details>
<summary>Review breakdown</summary>

1. Closing the public stream with `aclose()` bypasses the `CancelledError` rollback and checkpoints target B even though source A never completed.
2. Direct cancellation reruns A but persists B-directed handoff content and speculative shared context, which are then injected into A's retry.

These are separate defects, though one atomic rollback helper may address both. I did not repeat resolved findings or the accepted terminal at-least-once/old-writer limitations.
</details>

opieter-aws:
@strandly-the-agent re-review

strands-py/src/strands/multiagent/swarm.py
```diff
@@ -785,6 +802,12 @@ async def _execute_swarm(self, invocation_state: dict[str, Any]) -> AsyncIterato
                         self.state.completion_status = Status.FAILED
                         break
 
+                    self._uncommitted_handoff_snapshot = (
+                        self.state.handoff_node,
```
strandly-the-agent, line 831: 🔴 **Preserve the interrupted source as frontier until its resumed turn commits.**

In the reproduced sequence, A committed `A → B` and then interrupted. Resumed A therefore starts with active interrupt context keyed to A and pending B. This snapshot captures inherited B; if resumed A is cancelled or the public stream closes, rollback restores B and `serialize_state()` persists B as frontier even though A's resumed turn never committed.

Fresh disk restore then starts B, but interrupt context has only A, so execution raises `KeyError: 'B'` before either node runs (`A=0, B=0`, status `FAILED`). This reproduced twice for both `CancelledError` and `aclose()` with `FileSessionManager`.

Could A remain the durable frontier while its interrupt-resume turn is uncommitted, with B preserved separately as the post-A handoff? Please add disk-backed regressions for cancel and close that prove A resumes before B.

strands-py/src/strands/multiagent/swarm.py
```diff
@@ -1059,7 +1105,22 @@ def _from_dict(self, payload: dict[str, Any]) -> None:
 
         next_node_ids = payload.get("next_nodes_to_execute") or []
         if next_node_ids:
-            self.state.current_node = self.nodes[next_node_ids[0]] if next_node_ids[0] else self._initial_node()
+            resume_node = self.nodes.get(next_node_ids[0]) if next_node_ids[0] else None
+            if resume_node is None:
+                # Frontier target absent (legacy empty-string sentinel, or removed since save): fall back to
+                # the initial node and drop the stale interrupt state keyed by the absent node.
+                resume_node = self._initial_node()
+                self._interrupt_state = _InterruptState()
```
strandly-the-agent: 🟡 **Reconcile fallback frontier, handoff target, and message atomically.**

This replaces a missing frontier with the initial node but leaves the separately hydrated handoff/message intact. With removed source A and surviving pending target B, the fallback node receives B's target-only message, then hands off to B after that message has been cleared. If B is also removed, its orphaned message still reaches the fallback node. Both cases reproduced twice from disk.

Could fallback either intentionally resume the surviving target with its message, or clear the stale handoff and message together before running the initial node? The topology tests should invoke the restored swarm and assert which node receives the handoff text; the current removed-handoff test checks only `handoff_node is None`.

strandly-the-agent (changes requested):
**🔴 Changes requested — interrupt-resume cancellation can still persist an unrestorable frontier.**

✅ Re-reviewed exact head `a0f7f284` against prior `c05970ee`; the public `aclose()` and speculative handoff-context findings are resolved.
✅ Exact-head local verification: 93 tests, changed-file Ruff/check/format, focused mypy, and diff-check passed. Specialists also ran Swarm 51/51, FileSession 40/40, and full mypy over 226 files.
✅ Remaining blocker reproduced twice for both `CancelledError` and `aclose()` with real `FileSessionManager`; topology fallback issue reproduced twice.
⚠️ Unit/lint/docs/CI Gate are green. The separate integration job failed, but logs are unavailable, so I am not attributing it. Advanced adversarial was unavailable under the data-retention policy; the default-tier retry completed.

<details>
<summary>Review breakdown</summary>

- **Blocker:** cancelling resumed interrupt source A restores inherited pending handoff B, then serializes B as frontier while interrupt context remains keyed only to A. Fresh restore fails before invoking either node.
- **Should-fix:** topology fallback changes the frontier without reconciling handoff/message, so target-specific text can reach the wrong node or become orphaned.

I did not reopen accepted terminal at-least-once behavior, old-writer compatibility, or prior resolved findings.
</details>

opieter-aws:
@strandly-the-agent re-review latest version

strands-py/src/strands/multiagent/swarm.py
```diff
@@ -1070,9 +1154,17 @@ def _from_dict(self, payload: dict[str, Any]) -> None:
...
+        self.state.current_node = self.nodes[payload["next_nodes_to_execute"][0]]
+
+        # Clear a handoff the resume frontier already satisfies so the target is not re-run. Skip INTERRUPTED:
+        # there the frontier is the current node, so a self-handoff equality is coincidental and still pending.
+        if (
+            self.state.completion_status != Status.INTERRUPTED
+            and self.state.handoff_node is not None
+            and self.state.handoff_node == self.state.current_node
+        ):
+            self.state.handoff_node = None
```
strandly-the-agent: 🔴 **A still-pending self-handoff is dropped on the second crash-restart.**

`handoff_node == current_node` is used to *infer* "already satisfied", but that equality is also true for a self-handoff that is still pending after a rollback — and the `INTERRUPTED` carve-out stops protecting it once the state is re-persisted as `EXECUTING`.

Reproduced at this head, both teardown modes:

1. `a` calls `handoff_to_agent("a")` then interrupts → persisted `status=interrupted, frontier=['a'], handoff_node='a'` ✅
2. restore, resume, tear the resumed turn down before it commits (public `aclose()` or task cancel) → persisted `status=executing, frontier=['a'], handoff_node='a'` ✅ (rollback is correct)
3. restore again and resume → this block clears the handoff; `a` runs **once** and the swarm reports `COMPLETED`.

Ground truth for the same scenario resumed in memory is `history=['a','a']`; observed is `['a']`. Running the identical repro against merge-base source gives `['a','a']`, so this is new in this PR.

`_TurnCheckpoint.outcome` already carries "pending vs. satisfied" in memory — would persisting that bit explicitly, rather than re-deriving it from `handoff_node == current_node` + status, close this across N restore generations? A regression driving two full restore → resume → teardown → restore cycles would lock it in; `test_swarm_resume_interrupted_self_handoff_preserved` only exercises a single `deserialize_state` call.

strands-py/src/strands/multiagent/swarm.py
```diff
@@ -798,15 +843,23 @@ async def _execute_swarm(self, invocation_state: dict[str, Any]) -> AsyncIterato
...
+                        # An interrupt commits the turn: resume continues past it, so the handoff stands.
+                        self._turn.outcome = "committed"
                         yield self._activate_interrupt(current_node, node_result.interrupts)
                         break
 
                     self._interrupt_state.deactivate()
 
                     self.state.node_history.append(current_node)
+                    self._turn.outcome = "committed"
+
```
strandly-the-agent: 🟡 **The success-path commit lands one `yield` too late.**

The last item forwarded by `async for event in node_stream: yield event` is the node's own `MultiAgentNodeStopEvent`, but `outcome = "committed"` is only set after that loop returns control. A `GeneratorExit`/`CancelledError` arriving exactly at that yield is caught with `outcome` still `"open"`, so `_rollback_uncommitted_turn()` wipes `handoff_node`/`handoff_message`/`shared_context` even though the node fully succeeded and its `NodeResult` and accumulated usage were already stored.

Observed at this head after consuming through the stop event and calling `aclose()`: `handoff_node second→None`, `shared_context {'first': {'k':'v'}}→{}`, while `results=['first']` survives. Merge base preserves both through the same sequence.

Impact is bounded — the frontier is `['first']`, so resume replays that node and re-does its handoff and context write; every teardown×phase combination we tried still restored to `COMPLETED`. Hence should-fix rather than blocking.

The `INTERRUPTED` branch just above already commits *before* it yields and is immune — would mirroring that ordering (commit as soon as `node_result.status` is known non-`INTERRUPTED`, before forwarding the terminal event) be the cleaner invariant? A test tearing down right after the node's `result` event would cover it; `test_swarm_resume_closed_stream_after_handoff_reruns_source` stops one event earlier.

strandly-the-agent (changes requested):
**🔴 Changes requested — one new regression: a still-pending self-handoff is silently dropped on a *second* crash-restart.**

✅ Re-reviewed exact head `77608fe9` against prior `a0f7f284` and merge base `fe4cbb94`. Both prior blockers (interrupt-resume cancel promoting an inherited handoff; removed-frontier-node leaking the target's message/context) are **verified resolved** with real `FileSessionManager` on-disk snapshots and fresh rebuilds — the `_TurnCheckpoint.outcome` model is the right shape.
✅ Exact-head local run: 98 focused tests passed (Swarm 56 + file session manager + multi-agent hooks), changed-file `ruff check`/`format --check` clean, `mypy` on `swarm.py` clean, `git diff --check` clean.
✅ The regression below reproduced deterministically at this head for both teardown modes, and the *same* repro passes against merge-base source — so it is introduced by this PR, not pre-existing.
✅ CI rollup on this merge commit is green; the earlier integration-job failure was on `89654926`, not this head.

<details>
<summary>Review breakdown</summary>

- **🔴 Blocker:** two-generation crash-restart clears a still-pending self-handoff, so the swarm reports `COMPLETED` having skipped a turn the agent explicitly requested. No test drives two restore generations.
- **🟡 Should-fix:** the success-path turn commit lands one `yield` too late, so a teardown right after a node's stop event rolls back a fully-succeeded node's handoff/shared context while keeping its result. Self-healing in practice (the frontier replays that node), so not blocking.

Both are timing gaps in when `outcome` flips relative to a yield; a single structural rule (resolve `outcome` before any yield that could be a turn's last) may close both.

I did not reopen the settled terminal at-least-once contract, the old-writer/new-reader limitation, same-instance concurrency, or the pre-existing accumulator double-count (verified identical at merge base).
</details>

strandly-the-agent:
@opieter-aws — rather than just re-raising the two findings, I built a candidate for them and pressure-tested it. Take it, adapt it, or ignore it; the reasoning matters more than the diff.

**Branch:** `strandly-the-agent/harness-sdk` @ `strandly/swarm-crash-restart-record-turn-phase` (`2126aec2`), based on this PR's head `77608fe9`. Two commits, +210/−19 in `swarm.py` + `test_swarm.py`.

**The invariant it establishes:** *a persisted handoff always means its target still owes a turn.* The writer settles that where the turn's outcome is known, and restore honors the payload literally — so `context.handoff_node` is simply **not** persisted when the resume frontier already carries that target, and the reader-side `handoff_node == current_node` + status inference is **deleted rather than replaced**. No new schema key. That closes the self-handoff case at any restore depth instead of one generation at a time, because the frontier can no longer be ambiguous about whether it consumed the handoff.

The second commit moves the completed-turn commit *before* the node's terminal `MultiAgentNodeStopEvent` is forwarded (its result, metrics and shared-context writes are already recorded at that point), while the interrupted turn keeps committing next to `_activate_interrupt()` — a "committed" marker must never be set before the state that makes it true is durable.

<details>
<summary>What was verified, and what wasn't</summary>

- 61/61 `tests/strands/multiagent/test_swarm.py`; 308 across `tests/strands/multiagent` + `tests/strands/session` (a2a extra unavailable in my sandbox); `ruff`, `ruff format --check`, `mypy` clean.
- Four mutants killed: writer keeps a consumed handoff · reader inference reinstated · commit moved back after the loop · interrupted turn committed inside the loop.
- Teardown sweeps across 8 run shapes × 4 teardown modes × every yield point, each row serialized, restored into a fresh `Swarm` and resumed; three N-generation plans; a 53-shape checkpoint corpus written by three writers and read by three readers; two real `FileSessionManager` on-disk round-trips.
- A stronger oracle (node steps gated by markers written to durable shared context, so a dropped handoff can't be re-created by the replay) over 2,400 in-memory + 120 on-disk rows: candidate clean, this PR's head fails 222 rows.
- **Not** verified: integ tests, S3 session manager, and three outstanding test-strength items — the legacy-compat test asserts restored state but doesn't run to completion, the interrupt commit sits one line before `_activate_interrupt` while its comment says after, and no unit test pins the `deactivate()` move for a resumed interrupted node that then completes. All three are test/comment polish, not behavior.

**Two behavior changes worth calling out in the PR body if you take this:** consumers observing `multiagent_node_stop` now see `node_history` already containing that node, and a checkpoint written before this change replays a turn at worst rather than skipping one (the at-least-once direction).

</details>

One related thing that isn't yours to fix here: Python lets a node hand off to itself, while the TypeScript SDK filters the caller out of the handoff enum (`strands-ts/src/multiagent/swarm.ts`). That divergence is the root cause of the ambiguity — probably worth its own issue.

Happy to rebase the branch onto your latest head, or to drop it entirely if you'd rather implement the invariant yourself.

opieter-aws:
@strandly-the-agent review latest version

strands-py/src/strands/multiagent/swarm.py
```diff
@@ -787,26 +826,48 @@ async def _execute_swarm(self, invocation_state: dict[str, Any]) -> AsyncIterato
...
+                        # A completed turn commits before its terminal event is forwarded: the node's
+                        # result, metrics, and shared-context writes are already recorded, so a teardown
+                        # arriving at this yield must not roll it back. Only COMPLETED commits here — a
+                        # failing node also emits a stop event, and its raise reaches the handler below.
+                        if isinstance(event, MultiAgentNodeStopEvent):
+                            node_result = event["node_result"]
+                            if node_result.status == Status.COMPLETED:
+                                self._interrupt_state.deactivate()
+                                self.state.node_history.append(current_node)
+                                self._turn.outcome = "committed"
```
strandly-the-agent, line 852: 🔴 **A turn that completes with no handoff is checkpointed as both done and still-owed — and the bogus entry compounds per crash.**

This append now runs before the stop event is forwarded, but `serialize_state`'s committed branch requires `self.state.handoff_node`, so a no-handoff commit falls through and re-serializes `current_node` as the frontier — the node `node_history` just marked done. The post-loop transition that would set `COMPLETED` never runs on teardown.

My repro (single node, `max_iterations=2`, `stream.aclose()` at its `multiagent_node_stop`, in-memory round-trip), head vs `77608fe9`:

| | head `87ff5ada` | prior `77608fe9` |
|---|---|---|
| gen0 crash | `history=['solo'] frontier=['solo']` | `history=[] frontier=['solo']` |
| gen1 crash, same point | `history=['solo','solo'] frontier=['solo']` | `history=[] frontier=['solo']` |
| gen2 clean resume | **`status=failed, execution_count=2`, 0 node calls** | `status=completed, execution_count=1`, 1 call |

Prior is idempotent under repeated crashes; head grows one entry per crash. An on-disk `a→b→c` chain reproduces it too (`[a,b,c] → [a,b,c,c] → [a,b,c,c,c]`, then `failed` with zero calls, after which the next invocation resets and re-runs the whole task). The node executes twice at that teardown point in *both* revisions, so the replay itself is the accepted at-least-once cost — what is new is the contradictory checkpoint and the compounding.

Why it bites beyond bookkeeping: `node_history` feeds `should_continue`'s `max_iterations`/`max_handoffs` and the repetitive-handoff window, `execution_count`, and `_build_node_input`'s "Previous agents who worked on this" — so the replayed node is told it is its own predecessor (`a → b → c → c`), and a swarm that never handed off once can trip repetitive-handoff detection.

When a committed turn has no handoff, nothing is owed — that is exactly what the loop does next (`COMPLETED`, break). Would persisting an empty frontier in that case be the cleaner reading, or would you rather keep `deactivate()` + `outcome="committed"` here and move only the append back after the loop? Either shape wants a two-generation test that crashes twice at a no-handoff completion and asserts the checkpoint is a fixed point.

strands-py/tests/strands/multiagent/test_swarm.py
```diff
@@ -1335,6 +1337,777 @@ async def handoff_stream(*args, **kwargs):
...
+    second_agent2 = create_mock_agent("second")
+    resumed_swarm = Swarm([first_agent2, second_agent2])
+    resumed_swarm.deserialize_state(failed_snapshot)
+
+    assert resumed_swarm._resume_from_session is False
+    assert resumed_swarm.state.completion_status == Status.PENDING
+
+
+def test_swarm_serialize_omits_handoff_the_resume_frontier_carries():
+    """A handoff the resume frontier already carries is not persisted.
```
strandly-the-agent, line 2021: 🟡 **Coverage gaps around the new invariant — bundling four.**

1. This is the only test of the new `frontier_carries_handoff` guard and it hand-builds `_turn`/`state.handoff_node` rather than driving a node. The reachable runtime path (a self-handoff node completes, teardown at its terminal stop event) is uncovered: with the guard reverted, that path silently persists `handoff_node="first"` next to `frontier=["first"]` and the node then runs an extra time — no test fails.
2. No test crashes twice at the same teardown point. Every resume test is single-generation, which is precisely why the compounding in my other comment is invisible to a green suite; a test asserting the checkpoint is a *fixed point* across repeated crashes would have caught it.
3. `test_swarm_stream_closed_at_node_stop_event_keeps_committed_turn` computes `snapshot = swarm.serialize_state()` but asserts handoff/shared-context against live `swarm.state`/`swarm.shared_context`, so a bug confined to the persisted payload passes. Its interrupt sibling does it right by asserting on the snapshot and rebuilding from it.
4. All five new tests are in-memory round-trips; the two on-disk `FileSessionManager` tests each cover a single restore generation, so no test drives a multi-generation crash chain through real persistence — the thing this PR is about.

None of these are blocking on their own, but (1) and (2) are what let the regression through.

strandly-the-agent (changes requested):
**🔴 Changes requested — the invariant lands and closes both prior findings, but moving `node_history.append` before the terminal yield introduces a new regression: a node completing with *no* handoff is checkpointed as both done and still-owed, and the bogus entry compounds once per crash until the swarm wedges into `FAILED` having executed nothing.**

✅ Both prior findings verified fixed by my own repros at `87ff5ada` vs `77608fe9`: the multi-generation pending self-handoff now gives `node_history == ['a','a']` (matching in-memory ground truth; prior gave `['a']`), and a teardown at the terminal stop event of a handoff-requesting node now persists `frontier=['second']` with its message and shared context intact (prior rolled all of it back and skipped `second`). The "persisted handoff means its target still owes a turn" invariant — settling it in the writer and deleting the reader-side inference — is the right call.
✅ Exact head: 103 focused tests pass (Swarm 61 + file-session + multi-agent hooks), changed-file `ruff check`/`format --check` clean, `mypy` on `swarm.py` clean, `git diff --check` clean. Full CI on this head is green, integration tests included.
🔴 New regression reproduced by me at head and controlled against `77608fe9` (idempotent there) — detail inline.

<details>
<summary>Review breakdown</summary>

- **🔴 Blocker:** no-handoff completed turn → self-contradictory checkpoint (`node_history=['solo']` **and** `frontier=['solo']`); repeated crashes at the same point grow history one entry per crash, which `should_continue` charges against `max_iterations`/repetitive-handoff and `_build_node_input` renders into the model's prompt.
- **🟡 Test gaps (bundled):** nothing crashes twice at the same point, so the compounding is invisible to the suite; the new write-side guard is only reached synthetically; one new test asserts live state where its docstring promises the persisted payload.

I did not reopen settled items (terminal at-least-once replay, old-writer/new-reader — this delta only ever replays more, never skips, same-instance concurrency, pre-existing accumulator double-count). One suppressed ⚪: a node whose stream fails *after* its COMPLETED stop event is now counted in a FAILED run's history — resume is unaffected and it is arguably more accurate. The advanced adversarial pass needed two retries after transient failures before completing.
</details>

opieter-aws:
@strandly-the-agent review latest

strandly-the-agent (commented):
> [!WARNING]
> **Superseded and partly retracted — see [#pullrequestreview-4823897865](https://github.com/strands-agents/harness-sdk/pull/3391#pullrequestreview-4823897865) for my actual position.**
>
> This review was submitted through a tooling error on my side and **overstates the severity**. The finding below is factually reproducible — the terminal `AfterNodeCallEvent` write does land `status="executing"` with `frontier=[]`, and restoring exactly that write does reset and discard the finished run — but framing it as trading "a compounding-history bug for a worse one" is wrong: the baseline it compares against (`87ff5ada`) is the revision that wedged the swarm into `FAILED` with zero executions. Re-running the task is bounded and within the documented at-least-once contract, so this is a **non-blocking trade-off**, not changes-requested. My corrected review approves `085b4cf3` and carries the item as 🟡 with that framing.

---

<details>
<summary>Original text, kept for the record (severity overstated — read the correction above first)</summary>

**🔴 Changes requested — the no-handoff empty-frontier fix trades a compounding-history bug for a "discard the whole completed run" bug.**

✅ Re-reviewed exact head `085b4cf3` against prior head `87ff5ada` and merge base `59bf1ed6`.
✅ Commit `2d28c70f` genuinely fixes the compounding-history blocker I raised last round — reverse-tested against `87ff5ada`.
✅ Commit `085b4cf3` genuinely fixes the teardown-between-nodes handoff-drop — reverse-tested by removing the `self._turn = None` line, frontier regresses `['second']` → `[]`.
🔴 New regression: a committed terminal turn now serializes `next_nodes_to_execute = []`, and `deserialize_state` treats *any* empty frontier as "nothing to resume" — routing to `_reset_for_fresh_run()`, which discards `node_history`, `node_results`, `accumulated_usage`/`metrics` and `shared_context` wholesale even though the run actually succeeded.

**Verified correct (both commits do what they claim):**
- `2d28c70f` (swarm.py:1049-1057): a committed turn with no handoff now serializes an empty frontier instead of re-listing the just-finished node. Reverse-tested: collapsing it back to the pre-fix form makes a `solo`/`max_iterations=2` repeated-crash-at-the-same-point repro compound `node_history` across generations again.
- `085b4cf3` (swarm.py:891): `self._turn = None` right after the handoff transition. Confirmed a teardown between "first"'s handoff and "second" starting persists `next_nodes_to_execute=['second']` with `085b4cf3`, vs `[]` with only `2d28c70f` applied (silently dropping the target).

**The window:** the new empty-frontier branch is reached whenever a turn commits with no handoff *while `completion_status` is still `EXECUTING`* — before the loop's `else: completion_status = Status.COMPLETED`. `SessionManager` wires `AfterNodeCallEvent → sync_multi_agent`, and that hook fires from inside that window for every node including the terminal one, so `FileSessionManager` writes it to disk. A real on-disk repro (A→B, B terminal, no crash injected):

```
write #0 (after A's AfterNodeCallEvent): status=executing node_history=['first']          next_nodes=['second']
write #1 (after B's AfterNodeCallEvent): status=executing node_history=['first','second'] next_nodes=[]
write #2 (after AfterMultiAgentInvocationEvent): status=completed node_history=['first','second'] next_nodes=[]
```

Restoring write #1 gives `completion_status=PENDING, node_history=[], results={}` — the entire swarm reruns from the initial node. The identical repro against `87ff5ada` writes `next_nodes=['second']` at that point, so restore there resumes at "second" while preserving history/results for "first".

**Suggested direction:** flipping `completion_status` to `COMPLETED` where the turn commits, instead of after the `AfterNodeCallEvent` hook, would make every checkpoint written in this window report `status="completed"` — funnelling into the documented reset-on-resume contract instead of colliding with it under a misleading `"executing"` label.

</details>

strands-py/src/strands/multiagent/swarm.py
```diff
@@ -978,12 +1042,21 @@ def _build_result(self, interrupts: list[Interrupt]) -> SwarmResult:
...
         status_str = self.state.completion_status.value
-        if self.state.completion_status == Status.EXECUTING and self.state.current_node:
-            next_nodes = [self.state.current_node.node_id]
-        elif self.state.completion_status == Status.INTERRUPTED and self.state.current_node:
+        # Only a committed turn earns its handoff target the frontier; a handoff standing mid-turn was
+        # inherited, and resuming its target would skip the node that never finished.
+        turn_committed = self._turn is not None and self._turn.outcome == "committed"
+        frontier_carries_handoff = False
+        if self.state.completion_status == Status.EXECUTING and turn_committed:
+            # A committed turn owes only its handoff. Without one the swarm is done, so the frontier is
```
strandly-the-agent, line 1051: 🟡 **Non-blocking — this branch makes the terminal checkpoint say `status="executing"` while owing nothing, and the reader treats that as "start over".**

The branch is right that a committed turn with no handoff owes nothing. The wrinkle is *when* it is observed: `AfterNodeCallEvent` fires from the `finally` **before** the loop sets `COMPLETED`, and `SessionManager` writes `serialize_state()` straight to disk there — so an ordinary successful run persists this, verified with a real `FileSessionManager` (A→B, B terminal, no crash injected):

```
write after_node:first   status=executing history=['first']           frontier=['second']
write after_node:second  status=executing history=['first','second']  frontier=[]        <- owes nothing, still "executing"
write after_invocation   status=completed history=['first','second']  frontier=[]
```

Restoring that middle write (i.e. the process dies in the gap before the final write) routes to `_reset_for_fresh_run()`, since `deserialize_state` decides reset-vs-resume from frontier emptiness alone and never consults `status`: `completion_status=pending, node_history=[], results={}` — the finished run's history, results, usage and shared context are dropped and the whole task re-runs. At `87ff5ada` that same write was `frontier=['second']`, so resume replayed only the last node and kept the rest.

I am explicitly *not* calling that a regression: the `87ff5ada` behavior is the compounding one I blocked on, and re-running the task is within the at-least-once contract you documented. It is a narrower, calmer failure mode than a wedge. The residue is just that the payload contradicts `deserialize_state`'s own docstring ("empty frontier — e.g. completed, or failed at a dead end") by carrying `status="executing"`.

Would flipping `completion_status` to `COMPLETED` where the turn commits — rather than after the hook — be worth it, so every checkpoint in that window is self-consistent and lands in the documented completed-run path? One caveat if you try it: `_handle_handoff` early-returns unless status is `EXECUTING`, so an `AfterNodeCallEvent` callback that requests a handoff would start being ignored. Equally defensible to leave as is and note the window in the resume docstring — your call.

strandly-the-agent (approved):
**✅ Approving `085b4cf3` — both new commits are genuine fixes, verified by my own repros. One non-blocking trade-off to decide, inline.**

**First, a correction of my own.** The review I posted 20 minutes ago ([#pullrequestreview-4823866300](https://github.com/strands-agents/harness-sdk/pull/3391#pullrequestreview-4823866300)) went out through a tooling error on my side, and it **overstates the severity**: it frames the terminal-window behavior as trading "a compounding-history bug for a worse one". My measurements don't support "worse" — the baseline it compares against (`87ff5ada`) is the revision that wedged the swarm into `FAILED`. Treat this review as my position; that one is superseded.

✅ `2d28c70f` (empty frontier for a committed turn with no handoff) — verified a fixed point: three successive crashes at the same no-handoff completion each persist `history=['solo'] frontier=[]`, and a clean resume is `completed` / `execution_count=1`. At `87ff5ada` the same script compounds to `['solo','solo','solo']` and then returns `failed` with **zero** node calls. My blocker is closed.
✅ `085b4cf3` (`self._turn = None` after the handoff transition) — verified a real guard against *skipped* work: tearing down at the `multiagent_handoff` event persists `frontier=['second']` at head, but `frontier=[]` at `2d28c70f`, where the resume then drops `second` entirely (`history=['first']`, only `first` re-runs).
✅ Exact head: 106 focused tests pass (Swarm 64 + file-session + multi-agent hooks), changed-file `ruff check`/`format --check` clean, `mypy` on `swarm.py` clean, `git diff --check` clean.
⚠️ CI rollup is red **only** on `label-size`; every test/lint/docs/integration job and `CI Gate` are green — looks like the diff outgrew its size label rather than a functional failure.

<details>
<summary>The one non-blocking item (detail inline)</summary>

The terminal node's `AfterNodeCallEvent` checkpoint — a real `SessionManager` write — now lands `status="executing"` with `frontier=[]`. Restoring exactly that write resets and discards the finished run, so a crash in the gap before the invocation-final write costs a full task re-run instead of a one-node replay. It is bounded and within your at-least-once contract (no skipped work, no corruption, no wedge), and it converges with what a `completed` checkpoint already does — so which side of the trade you want is your call, not a merge blocker.

</details>

The `_turn` lifecycle reads consistently end to end now: it is created when a turn starts, committed where the state that makes it true is durable, rolled back on every teardown path, and released the moment the finished turn stops speaking for the frontier. Nice iteration on a genuinely subtle state machine.

strands-py/src/strands/multiagent/swarm.py
opieter-aws: The window for this error mode is small, nothing breaks, and re-running is complicit with the documented at-least-once contract. Not inclined to change right now

JackYPCOnline:
From my reading of the description, I thought swarm resume is broken, after reading the code, I found it fixes non-happy path case, dies in a specific window so can we explicitly update the description?

JackYPCOnline:
Given TS avoids this entirely by construction, do we have a tracking issue for this? I would rather have this as temorary bridge.

strands-py/src/strands/multiagent/swarm.py
```diff
@@ -234,6 +234,26 @@ class SwarmResult(MultiAgentResult):
     node_history: list[SwarmNode] = field(default_factory=list)
 
 
+@dataclass
+class _TurnCheckpoint:
```
JackYPCOnline: Do you think we should rename it? We've defined checkpoint, snapshot & transcript already. reusing checkpoint is not perfectly fit.
Since you use self._turn, maybe `_InflightTurn`, `_PendingTurn` or something else.

strands-py/src/strands/multiagent/swarm.py
```diff
@@ -787,26 +826,48 @@ async def _execute_swarm(self, invocation_state: dict[str, Any]) -> AsyncIterato
                         self.state.completion_status = Status.FAILED
                         break
 
+                    self._turn = _TurnCheckpoint(
```
JackYPCOnline: So if this is only useful when we enabled session manager, can we avoid the overhead in normal case?

strands-py/src/strands/multiagent/swarm.py
```diff
@@ -787,26 +826,48 @@ async def _execute_swarm(self, invocation_state: dict[str, Any]) -> AsyncIterato
...
+                        # requested stands — but a commit must never precede the state that makes it true,
+                        # so it follows the activation that records the interrupt.
+                        interrupt_event = self._activate_interrupt(current_node, node_result.interrupts)
+                        self._turn.outcome = "committed"
+                        yield interrupt_event
                         break
 
-                    self._interrupt_state.deactivate()
-
-                    self.state.node_history.append(current_node)
```
JackYPCOnline, line 806: The commit/append moved inside if node_result.status == Status.COMPLETED. The old code ran deactivate() + node_history.append(current_node) for any non-INTERRUPTED status. Is this behavior change expected?

strands-py/src/strands/multiagent/swarm.py
```diff
@@ -1070,9 +1183,8 @@ def _from_dict(self, payload: dict[str, Any]) -> None:
         self.state.results = results
         self.state.task = decode_bytes_values(payload.get("current_task", self.state.task))
 
-        next_node_ids = payload.get("next_nodes_to_execute") or []
-        if next_node_ids:
-            self.state.current_node = self.nodes[next_node_ids[0]] if next_node_ids[0] else self._initial_node()
+        # Only reached for a frontier that names a node this swarm defines.
+        self.state.current_node = self.nodes[payload["next_nodes_to_execute"][0]]
```
JackYPCOnline, line 1188: is this index 0 guranteed? No edge case produces index error?

opieter-aws: Yes, this is guaranteed in `deserialize_state`. It guards before delegating:` if not next_node_ids or next_node_ids[0] not in self.nodes: self._reset_for_fresh_run(); return`

strands-py/src/strands/multiagent/swarm.py
opieter-aws: Yes, intentional. The change is that a FAILED node no longer gets marked as completed. The old condition was “not INTERRUPTED”, which let FAILED through.

strands-py/src/strands/multiagent/swarm.py
opieter-aws: I’d rather keep it unconditional, since the turn record isn’t only useful for persistence. `stream_async`’s teardown rollback reads it on the in-memory interrupt-resume path, where no session manager is involved.

strands-py/src/strands/multiagent/swarm.py
opieter-aws: That's a good point, will rename to `_InflightTurn`

opieter-aws:
> Given TS avoids this entirely by construction, do we have a tracking issue for this? I would rather have this as temorary bridge.

The underlying cause is that Python stores the resume frontier as a separate field from `node_history`, so `serialize_state` has to work out what’s still owed from status + handoff_node + a turn marker. 

TS doesn’t have this problem since the handoff decision is stored inside the node’s own result, so the result is the record that the turn committed.

Fixing that in Python would mean changing the persisted resume format and reworking how `_execute_swarm` records turns. That's a big Swarm rework with a session-compatibility story, not a quick follow-up. So I’d rather not frame this PR as a temporary fix

JackYPCOnline:
Thanks for the reply and I see your point, i am incline to approve. Suggesting to have Murat / Patrick to do a bar raise before merging.

JackYPCOnline (approved):

mkmeral:
my 2c around the last node edge case: we don't make any idempotency promises around swarm runs, the whole point is handing the task around, so the agent might re-run multiple times. Also given the current case is crash, I am okay merging this. I'd like to solve this edge case if we can, but I wouldn't want that to block this PR. so all good from me

Issue #3244, opened 2026-07-14
Vended tools: use_agent — delegation pattern (py + ts)


yonib05:
Part of #3234. Multi-agent pattern: delegation.

Add a `use_agent` vended tool in both SDKs: the agent constructs and invokes a nested agent at runtime. Shim over `Agent` + `as_tool()`/`asTool()`; the tool is spec, model resolution, and invocation.

- Shares the model config mapping (`model_provider`/`model_settings` -> typed model objects) with the other multi-agent tools.
- A follow-up design review on dynamic agent creation is expected; this issue is the shim.

**Done:** ships in both SDKs with types, tests, README, docs section.

PR #3392, opened 2026-07-21
feat(vended-tools): add use_agent delegation shim

chaynabors:
### Deprioritized by me, for others to pick up following a design pass.

Adds a use_agent vended tool to both SDKs. The tool lets the calling agent construct a fresh nested Agent at runtime (system prompt and an allowlisted subset of the parent's tools), delegate one task to it, and receive the child's final response. It is a thin shim over Agent and its normal construction path; the security guardrails live at the tool boundary rather than in a reimplementation of agent lifecycle. The child inherits the parent's model instance. The tool does not accept model_provider or model_settings because those constructors accept api_key, endpoint_url, and similar fields that constitute a credential-injection surface; a developer who wants per-child model variance can register N pre-configured factories at Agent construction time rather than exposing model config to the model.

The tool boundary enforces an exact-name tool allowlist (wildcards and the multi-agent tool names use_agent, swarm, graph, and a2a_client are rejected), size caps of 8 KiB on system_prompt and 32 KiB on task, and a factory-configurable allowlist entry cap. Recursion depth is tracked on a shared multiagent_depth counter on invocation_state, capped at three by default, and enforced by every participating multi-agent tool; the parent's invocation_state is spread and only the depth key is overridden so tracing and per-run keys flow through to the child. Non-finite and boolean values on factory-time numeric config are rejected. On cancellation the Python tool returns status cancelled while the TypeScript tool re-raises an AbortError to match the http-request tool; the shared conventions doc records the asymmetry.

The result shape uses the shared multi-agent status vocabulary of completed, failed, cancelled, and interrupted (the lower-cased SDK Status enum values). Field names use snake_case in Python and camelCase in TypeScript. Only a child that finished with stop_reason end_turn maps to completed; other non-cancelled stop reasons (limit_turns, content_filtered, max_tokens, guardrail_intervened, and so on) map to failed so the parent can distinguish a delivered delegation from one that hit a limit or policy.

https://github.com/strands-agents/harness-sdk/issues/3244


yonib05 (commented):
#3244 frames this as a shim over `Agent` + `as_tool()`, but I don't see `as_tool` used anywhere. Since `Agent(tools=[other_agent])` already auto-wraps, is the new surface here mostly about letting the model author the child's system prompt? If that's the value, would named pre-configured child specs get us delegation without the injection surface?

---

The 50ms poll loop exists because `Agent` won't accept an external cancel signal, and TS gets the same thing in two lines. Is that a gap worth fixing on `Agent` instead of polling once per delegation?

---

Blocking the literal names `use_agent`/`swarm`/`graph`/`a2a_client` is bypassed by the factory's own `name=` param. Since the depth counter is the real control, is the blocklist implying protection we don't have?

---

This returns `completed` where #3380's swarm returns `success` for the same outcome, and both PRs add their own copy of `_multiagent_conventions.md`. Could the doc live in `team/designs/` once and all four multi-agent PRs rebase onto it? Also `MULTIAGENT_DEPTH_KEY`, `MAX_MULTIAGENT_DEPTH`, and `MultiagentDepthExceededError` are declared and exported in both this PR and #3380 through the same barrel, so whichever merges second will hit a duplicate export.

---

strands-py/src/strands/vended_tools/use_agent/use_agent.py
```diff
@@ -0,0 +1,290 @@
...
+        # depth counter with the incremented value.
+        child_invocation_state: dict[str, Any] = {
+            **invocation_state,
+            _DEPTH_KEY: depth + 1,
+        }
+
+        start = time.monotonic()
+        watcher = asyncio.create_task(_watch_parent_cancel(parent_agent, sub_agent))
+        try:
+            result = await sub_agent.invoke_async(task_v, invocation_state=child_invocation_state)
```
yonib05, line 253: We call `invoke_async` with no `limits` even though both SDKs support it. Depth 3 with unbounded fan-out per level is unbounded spend, should the factory take `limits` and pass it through?

strands-py/src/strands/vended_tools/use_agent/use_agent.py
```diff
@@ -0,0 +1,290 @@
...
+    parent shouldn't crash the tool either.
+    """
+    raw = invocation_state.get(_DEPTH_KEY, 0)
+    if isinstance(raw, bool) or not isinstance(raw, int) or raw < 0:
+        return 0
+    return int(raw)
+
+
+def _resolve_parent_tools(agent: Any, names: list[str]) -> list[Any]:
+    registry = agent.tool_registry.registry
```
yonib05, line 141: The allowlist reads `tool_registry.registry` only, but dynamic tools (including everything #3398 registers) live in `dynamic_tools` and the executor checks both. Doesn't that mean a parent can invoke a tool it isn't allowed to delegate?

PR #3393, opened 2026-07-21
feat(vended-tools): add sleep tool

### LINKED ISSUE #3241: Vended tools: sleep (py + ts)

chaynabors:
Closes #3241.

Vended sleep tool in both SDKs. Pauses agent execution for a requested number of seconds and cooperates with the SDK's cancellation primitive so cancelling the enclosing invocation aborts the sleep immediately rather than waiting out the full duration. Python shims onto asyncio.sleep, TypeScript onto setTimeout plus the invocation's AbortSignal.

Both expose a factory with a configurable duration cap, defaulting to sixty seconds. The cap is set at construction and cannot be raised by the model. Requests above the cap are rejected at the tool boundary before the sleep starts. Negative, NaN, positive and negative infinity, non-numeric, and boolean durations are also rejected. Stdlib only, no new deps.

Tests exercise the boundary rejections, the factory-level cap checks, and a real five-second cancel that returns in well under a second. The TypeScript suite also covers the short-circuit path for an already-aborted signal and asserts, via a fake timer, that the underlying timer receives the requested duration converted to milliseconds.

### DIFF
--- a/strands-py/src/strands/vended_tools/sleep/types.py
+++ b/strands-py/src/strands/vended_tools/sleep/types.py
@@ -1,0 +1,18 @@
+"""Shared types and constants for the sleep tool."""
+
+DEFAULT_MAX_DURATION = 60.0
+"""Default upper bound on ``duration`` (seconds) accepted by :func:`make_sleep`."""
+
+
+def sleep_description(max_duration: float) -> str:
+    """Build the model-facing description with the configured max interpolated."""
+    return (
+        f"Pauses execution for a specified number of seconds (max {max_duration}). "
+        "Cooperative and cancellable: the sleep aborts immediately when the agent "
+        "invocation is cancelled. Rejects negative, NaN, infinite, or non-numeric "
+        f"durations, and durations above {max_duration}."
+    )
+
+
+SLEEP_DESCRIPTION = sleep_description(DEFAULT_MAX_DURATION)
+"""Description for the default sleep tool (60-second cap)."""
--- a/strands-py/tests/strands/vended_tools/test_sleep.py
+++ b/strands-py/tests/strands/vended_tools/test_sleep.py
@@ -1,0 +1,169 @@
+"""Tests for the sleep tool.
+
+The sleep tool pauses execution via :func:`asyncio.sleep`. Tests use tiny
+durations (single-digit millisecond values or less) and patch
+:func:`asyncio.sleep` where the assertion is about the argument rather than
+the delay itself, so the suite never actually blocks for a second. Coverage
+spans input validation, the configurable cap, and cooperative cancellation.
+"""
+
+import asyncio
+import importlib
+import math
+import types
+from unittest.mock import AsyncMock, patch
+
+import pytest
+
+from strands.vended_tools.sleep import make_sleep, sleep
+from strands.vended_tools.sleep.types import DEFAULT_MAX_DURATION
+
+# The parent package rebinds ``sleep`` to the tool object, which shadows the
+# submodule attribute. On Python 3.10 ``unittest.mock.patch`` walks a dotted
+# target with ``getattr`` and stops at that tool, raising AttributeError; 3.11+
+# resolves the longest importable prefix first. Reach the submodule via
+# ``importlib`` so tests behave identically across supported Python versions.
+_sleep_module = importlib.import_module("strands.vended_tools.sleep.sleep")
+
+
+class TestInputValidation:
+    """Boundary validation on the ``duration`` input."""
+
+    @pytest.mark.asyncio
+    async def test_rejects_negative_duration(self):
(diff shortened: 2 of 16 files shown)


arielnabavian (dismissed):

strands-py/src/strands/vended_tools/sleep/types.py
```diff
@@ -0,0 +1,12 @@
+"""Shared types and constants for the sleep tool."""
+
+DEFAULT_MAX_DURATION = 60.0
+"""Default upper bound on ``duration`` (seconds) accepted by :func:`make_sleep`."""
+
+SLEEP_DESCRIPTION = (
+    "Pauses execution for a specified number of seconds. Cooperative and cancellable: "
```
Unshure: nit: Should we inject in what the max duration is? Feels like the agent will fail and require an additional iteration if it accidentally goes over without knowing the limit

strands-py/tests/strands/vended_tools/test_sleep.py
```diff
@@ -0,0 +1,188 @@
...
+        fake_sleep = AsyncMock()
+        fake_asyncio = types.SimpleNamespace(sleep=fake_sleep)
+        with patch.object(_sleep_module, "asyncio", fake_asyncio):
+            result = await sleep(duration=2)
+
+        fake_sleep.assert_awaited_once_with(2.0)
+        assert result == "Slept for 2 seconds"
+
+
+class TestToolMetadata:
```
Unshure: nit: We probably dont need these tests

strands-ts/src/vended-tools/sleep/make-sleep.ts
```diff
@@ -0,0 +1,108 @@
...
+ * @returns A tool that pauses execution for the requested duration.
+ * @throws Error if `maxDuration` is not a finite, positive number.
+ *
+ * @example
+ * ```typescript
+ * const shortSleep = makeSleep({ maxDuration: 5 })
+ * const agent = new Agent({ tools: [shortSleep] })
+ * ```
+ */
+export function makeSleep(options: MakeSleepOptions = {}): ReturnType<typeof tool> {
```
Unshure: Is this just a `InvokableTool` or a `Tool`?
**https://github.com/strands-agents/harness-sdk/blob/main/strands-ts/src/tools/tool.ts#L166**

Unshure (dismissed):

strands-py/src/strands/vended_tools/sleep/types.py
chaynabors: Done. `sleep_description(max)` now interpolates the configured max into the description in both SDKs, and `make_sleep`/`makeSleep` call it with the resolved max so custom caps propagate.

strands-py/tests/strands/vended_tools/test_sleep.py
chaynabors: Agreed. Dropped the `TestToolMetadata` class in both SDKs; the behavior these were asserting is covered by the factory tests and there was nothing sleep-specific about them.

strands-ts/src/vended-tools/sleep/make-sleep.ts
chaynabors: It is `InvokableTool<SleepInput, JSONValue>`. Made the return type explicit instead of `ReturnType<typeof tool>`.

Unshure:
Can you look into the failing integ tests?

Unshure (dismissed):

Unshure (approved):

strands-py/src/strands/vended_tools/sleep/types.py
```diff
@@ -0,0 +1,18 @@
+"""Shared types and constants for the sleep tool."""
```
arronbailiss, line 1: are types files the idiomatic place for constants? I would have expected the main file

strands-ts/src/vended-tools/sleep/make-sleep.ts
```diff
@@ -0,0 +1,111 @@
+/**
```
arronbailiss, line 1: nit: make-sleep and sleep don't need to be separate files in strands-ts in my opinion

strands-py/src/strands/vended_tools/sleep/types.py
```diff
@@ -0,0 +1,18 @@
...
+
+DEFAULT_MAX_DURATION = 60.0
```
arronbailiss, line 3: nit: I would make the default maximum higher than 60 seconds. Anecdotal: something I've seen is my agent will sleep for say 30s, then 60s, then 120s. It will increase the sleep time during polling. Varies by model, prompts, etc. of course

PR #3394, opened 2026-07-21
feat(vended-tools): reconcile editor commands into file_editor

### LINKED ISSUE #3235: Vended tools: reconcile editor command set into file_editor (py + ts)

chaynabors:
Closes #3235.

Adds four commands to the file_editor vended tool: view (with line ranges), pattern_replace, find_line, and undo_edit. No new tool is introduced; the surface grows on the existing one. Presentation-only features from the strands-agents/tools editor (syntax highlighting, rich console output, .bak backups) are deliberately not ported.

Every command now routes through one path resolver with optional root confinement, so a future command cannot accidentally skip validation. The resolver applies a string-level check (absolute path, no parent traversal, inside root) and, when the local filesystem sees the target, a realpath check against symlink escape. Binary and non-UTF-8 files are rejected up front, the size cap defaults to one megabyte, and ambiguous str_replace or pattern_replace matches are refused unless the caller opts in with replace_all.

The pattern_replace timeout is Python-only; V8 has no way to bound a running regex, so TypeScript rejects nested-quantifier shapes statically instead. That divergence is documented in both docstrings and the vended-tools page. Undo is a per-tool-instance LRU capped by entry count and byte size, never touching disk.


strands-py/src/strands/vended_tools/file_editor/file_editor.py
```diff
@@ -23,20 +26,42 @@
     from ...tools.decorator import DecoratedFunctionTool
 
 _SNIPPET_LINES = 4
-_DEFAULT_MAX_FILE_SIZE = 1048576  # 1MB
+_DEFAULT_MAX_FILE_SIZE = 1 * 1024 * 1024  # 1 MB
```
arielnabavian: super nit: can have a variable for 1MB

strands-ts/src/vended-tools/file-editor/file-editor.ts
```diff
@@ -104,22 +209,159 @@ export function makeFileEditor(
...
+ * `(...*)*`, `(...+)*`, `(...*)+`). Catastrophic backtracking in JS regex
+ * (`(a+)+b` on all-a input) hangs a single `.exec()` call, and V8 has no
+ * regex timeout — the match-count cap only fires between exec calls, so this
+ * pre-compile heuristic is the only sync-time defense. False positives on
+ * legitimate patterns are acceptable — the caller can loosen the pattern.
+ */
+function looksLikeCatastrophicPattern(pattern: string): boolean {
+  // A quantified group whose interior also contains a quantifier is the
+  // classic ReDoS shape. `(a+)+`, `(a*)*`, `(a|a)*`, `(a+)*b`, etc.
+  return /\([^()]*[+*][^()]*\)[+*]/.test(pattern)
```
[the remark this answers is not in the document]
arielnabavian: Valid to a degree. Will leave this up to implementor.

strands-py/src/strands/vended_tools/file_editor/file_editor.py
chaynabors: Added `_MB = 1024 * 1024` on the Python side and `MB` on the TypeScript side; the file-size and undo-byte defaults now use them.

strands-ts/src/vended-tools/file-editor/file-editor.ts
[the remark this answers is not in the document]
chaynabors: Broadened the heuristic to also flag alternation and `{n,}` inside a quantified group, plus a bounded outer quantifier: `\([^()]*(?:[+*]|\||\{\d+,\d*\})[^()]*\)(?:[+*]|\{\d+,\d*\})`. Now catches `(a|a)*`, `(a{2,})+`, `(.*){10,}` etc. Added tests for the alternation-overlap and bounded-quantifier shapes, and updated the error message so it names all three patterns.

strands-ts/src/vended-tools/file-editor/types.ts
```diff
@@ -3,11 +3,30 @@
  */
 export interface FileEditorOptions {
```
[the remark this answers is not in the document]
chaynabors: Correct, both were dead. Dropped `FileEditorOptions` and `IFileReader` from `types.ts` and removed the re-exports in `index.ts`. `MakeFileEditorOptions` in `file-editor.ts` is now the single source of truth.

strands-py/tests/strands/vended_tools/test_file_editor.py
```diff
@@ -438,3 +450,450 @@ def test_input_schema_excludes_context(self):
...
+            await e(command="str_replace", path=p, tool_context=ctx, old_str=f"orig{i}", new_str=f"new{i}")
+        # Oldest snapshot (f0) has been evicted.
+        with pytest.raises(ValueError, match="No undo history"):
+            await e(command="undo_edit", path=paths[0], tool_context=ctx)
+        # Two most recent still restore.
+        await e(command="undo_edit", path=paths[1], tool_context=ctx)
+        assert (tmp_path / "f1.txt").read_text() == "orig1"
+
+    @pytest.mark.asyncio
+    async def test_evicts_past_byte_cap(self, ctx, tmp_path):
```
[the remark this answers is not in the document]
chaynabors: Good catch, the old sizing let both snapshots evict. Bumped the cap to `len(snapshot) + 20` so exactly one fits, then asserted p1 is gone AND p2 still restores to the original 100-byte content, mirroring `test_evicts_oldest_entry_past_entry_cap`.

mehtarac:
@strandly-the-agent , review this PR please

strands-py/src/strands/vended_tools/file_editor/file_editor.py
```diff
@@ -99,25 +205,82 @@ async def file_editor_tool(
...
+        # Symlink confinement is a best-effort, local-fs-only layer. If `root`
+        # itself does not exist on the local filesystem, the caller is running
+        # against a non-local sandbox (Docker, SSH, ...) whose paths do not map
+        # to this host — `os.path.realpath` would return the unchanged
+        # non-existent path and the check below would reject every operation
+        # (`realpath("/workspace-in-container")` != `realpath(root)`). In that
+        # case, the string-level check above is the whole policy; the sandbox
+        # owns its own symlink handling.
+        if not os.path.lexists(root):
+            return normalized
```
strandly-the-agent: 🔴 **`root` can be escaped through a symlink in Docker/SSH sandboxes.** This returns after lexical validation whenever the root is not on the SDK host, but the built-in remote sandboxes do not canonicalize paths before following them. I cold-ran a mapped remote sandbox twice with `/remote-root/link` pointing outside the root; `view` returned the outside file both times. Since `root` is exposed as confinement, please canonicalize inside the target sandbox (or fail closed when that capability is unavailable) rather than delegating to a policy the built-in Docker/SSH implementations do not provide. The TypeScript counterpart has the same gap at `file-editor.ts:247-260`.

strands-py/src/strands/vended_tools/file_editor/file_editor.py
```diff
@@ -159,49 +322,56 @@ def _apply_view_range(file_content: str, view_range: list[int] | None) -> tuple[
...
         )
 
-    new_content = file_content.replace(expanded_old, expanded_new, 1)
-    replacement_line = len(file_content[: file_content.index(expanded_old)].split("\n")) - 1
-    inserted_lines = len(expanded_new.split("\n"))
-    original_lines = len(expanded_old.split("\n"))
+    count = occurrences if replace_all else 1
+    new_content = (
+        original_content.replace(old_str, new_str_value)
+        if replace_all
```
strandly-the-agent, line 383: 🔴 **The size limit is checked only after the oversized result is allocated.** With the defaults, a 1 MiB file of `a` plus `replace_all=True` and an accepted 1 MiB replacement asks this call to construct roughly 1 TiB before `_reject_oversize_result` runs at line 663. Regex replacement has the same shape with up to 10,000 matches. Please preflight literal output bytes and build regex output incrementally with a running cap, so rejection happens before allocation. TypeScript has the same ordering at `file-editor.ts:438-442`.

strands-py/src/strands/vended_tools/file_editor/file_editor.py
```diff
@@ -50,46 +76,126 @@ def make_file_editor(
...
         A decorated tool that performs file operations through the sandbox.
     """
+    if root is not None and not posixpath.isabs(root):
+        raise ValueError(f"root must be an absolute path, got: {root}")
+    normalized_root: str | None = None if root is None else posixpath.normpath(root).rstrip("/") or "/"
+
+    # Bounded LRU: previous file content keyed by path. A second edit overwrites
+    # the snapshot for that path, matching the upstream editor's single-step
+    # undo semantics; older entries are evicted when either cap is exceeded.
+    undo_history: OrderedDict[str, str] = OrderedDict()
```
strandly-the-agent, line 120: 🔴 **The default singleton shares undo snapshots across agents/sandboxes.** `file_editor` owns this one path-keyed map, while lines 152–153 choose the sandbox per invocation. Cold repro twice: sandbox A and B each had a different `/same.txt`; editing A then calling `undo_edit` through B overwrote B with A's original content. Please key history by sandbox/agent identity as well as path, move it to agent/session state, or require an explicitly scoped editor instance. The exported TypeScript singleton has the same closure-level state at `file-editor.ts:142-145,210`.

strands-py/src/strands/vended_tools/file_editor/file_editor.py
```diff
@@ -355,73 +571,134 @@ async def _handle_view(sandbox: Sandbox, file_path: str, view_range: list[int] |
...
+    file_content = await _read_text_or_reject_binary(sandbox, file_path, max_size)
 
-    new_content, snippet, start_line = _build_str_replace_result(file_content, old_str, new_str, file_path)
+    new_content, snippet, start_line, count = _build_str_replace_result(
+        file_content, old_str, new_str, file_path, replace_all
+    )
 
+    _reject_oversize_result(new_content, max_size, file_path)
+    _store_undo_snapshot(undo_history, file_path, file_content, max_undo_entries, max_undo_bytes)
     await sandbox.write_text(file_path, new_content)
```
strandly-the-agent, line 667: 🔴 **Undo history is mutated before the corresponding write succeeds.** I injected a failed second write after `original → first`: this line replaced the valid `original` snapshot with `first`, so the subsequent undo could no longer restore the original. `_handle_undo` similarly pops at line 869 before its write, making a failed restore non-retryable. Please commit snapshot changes only after successful writes, and retain an undo entry until restoration succeeds. TypeScript has the same ordering at `file-editor.ts:743,889,960`.

strands-py/src/strands/vended_tools/file_editor/file_editor.py
```diff
@@ -430,3 +707,165 @@ async def _handle_insert(sandbox: Sandbox, file_path: str, insert_line: int | No
...
+                )
+        if not collected:
+            return collected, file_content, 0
+        substituted = regex.sub(new_str, file_content, count=0 if replace_all else 1)
+        replaced = len(collected) if replace_all else 1
+        return collected, substituted, replaced
+
+    try:
+        matches, new_content, count = await asyncio.wait_for(asyncio.to_thread(_run_regex), timeout=timeout_sec)
+    except asyncio.TimeoutError as e:
```
strandly-the-agent: 🔴 **This does not enforce the advertised wall-clock timeout.** CPython's backtracking `re` can hold the GIL, preventing the event loop from delivering `wait_for`'s timeout. Cold runs of `(a+)+b` against 26 `a`s with `pattern_replace_timeout=0.05` returned only after 4.806s and 4.826s. Larger inputs can block much longer. Please use a killable process/safe regex engine or reject risky shapes before execution; until then, remove the wall-clock guarantee here and in `vended-tools.mdx:49`.

strands-ts/src/vended-tools/file-editor/file-editor.ts
```diff
@@ -418,12 +770,194 @@ async function handleInsert(sandbox: Sandbox, filePath: string, insertLine: numb
...
+ * compiled regex + captured matches drives both the ambiguity check and the
+ * substitution (no second `.replace()` call, no third regex compile).
+ */
+function expandBackreferences(replacement: string, match: RegExpExecArray): string {
+  return replacement.replace(/\$(\$|&|\d)/g, (_full, token: string) => {
+    if (token === '$') return '$'
+    if (token === '&') return match[0]
+    const idx = parseInt(token, 10)
+    return match[idx] ?? ''
+  })
```
strandly-the-agent: 🟡 **The custom expander does not match the native semantics claimed above.** Exact-tool repro: pattern `(a)` with replacement `$9` wrote an empty file, while native `"a".replace(/(a)/g, "$9")` preserves literal `$9`; `$0` is also literal natively but this returns the whole match. Please preserve references that do not identify a capture and add `$0`/missing-group coverage.

```suggestion
    if (idx === 0 || idx >= match.length) return `$${token}`
    return match[idx] ?? ''
```

strandly-the-agent (dismissed):
**TL;DR — changes requested:** five blocking correctness/security issues reproduce at `84412b7a`, and this public API needs an explicit API review.

✅ Exact head `84412b7a` · merge-base `767b8019` · diff-check clean
✅ Focused tests: Python 90/90; TypeScript 96/96 with no type errors; CI Gate/matrices/lint/docs/security green
🔴 Cold repros: remote root escape ×2, cross-sandbox undo ×2, failed-write/undo history, Python timeout ×2, TS replacement semantics
🔴 Add `api/needs-review`, take this through an API meeting, then apply `api/review-complete`.

**Blocking questions:** What owns undo state—tool instance, agent, sandbox, or session? Should regex/search/undo expand the default tool automatically, or remain factory opt-ins while these contracts stabilize?

- **Correctness/adversarial:** five blockers retained; pre-existing full-read-before-size-check suppressed.
- **API/DevX:** this is a substantial stateful expansion of an automatically vended public primitive, so it needs an API meeting rather than solo approval.
- **LLM/docs:** prior feedback already covers replacement-dialect docs; timeout wording is folded into the timeout blocker below.
- **Tests:** no generic coverage comments; test requests are attached only to reproduced defects. Patch coverage remains 83.5%.
- **Prior threads:** did not repeat resolved MB constant, TS ReDoS heuristic, dead TS interfaces, or Python byte-cap findings.
- **Pass limitation:** advanced API/adversarial tiers were unavailable under the data-retention policy; API completed on default, and I independently cold-reproduced the adversarial blockers.

Human review should confirm the API direction after these defects are addressed.
</details>

strands-py/src/strands/vended_tools/file_editor/file_editor.py
```diff
@@ -50,46 +76,126 @@ def make_file_editor(
...
         A decorated tool that performs file operations through the sandbox.
     """
+    if root is not None and not posixpath.isabs(root):
+        raise ValueError(f"root must be an absolute path, got: {root}")
+    normalized_root: str | None = None if root is None else posixpath.normpath(root).rstrip("/") or "/"
+
+    # Bounded LRU: previous file content keyed by path. A second edit overwrites
+    # the snapshot for that path, matching the upstream editor's single-step
+    # undo semantics; older entries are evicted when either cap is exceeded.
+    undo_history: OrderedDict[str, str] = OrderedDict()
```
yonib05: `undo_history` is a closure variable on the module-level `file_editor` singleton, so doesn't every agent in the process share one undo map? Looks like agent B could call `undo_edit` on a path agent A edited and write A's snapshot to disk. Should this key off `tool_context.agent`?

strands-ts/src/vended-tools/file-editor/file-editor.ts
```diff
@@ -52,11 +78,39 @@ const fileEditorInputSchema = z.object({
...
+   * Maximum number of distinct paths retained in the in-memory undo history.
+   * Oldest entry is evicted on overflow. Defaults to 32.
+   */
+  maxUndoEntries?: number
+  /**
+   * Approximate cap on total bytes of file content held in the undo history
+   * (measured as UTF-8 byte length, matching the Python side). Oldest entries
+   * are evicted until the cap is met. Defaults to 32 MB.
+   */
+  maxUndoBytes?: number
```
yonib05, line 103: In the TS byte-cap loop, with `maxUndoBytes: 10` and a 100-byte snapshot, doesn't the while loop evict the entry we just stored? The test only asserts p1 has no history so it passes either way. The Python twin also asserts p2 still restores, which is why it doesn't have the same hole.

strands-ts/src/vended-tools/file-editor/file-editor.ts
```diff
@@ -1,40 +1,66 @@
...
+const DEFAULT_MAX_UNDO_BYTES = 32 * MB
 
 /**
  * Zod schema for file editor input validation.
  */
 const fileEditorInputSchema = z.object({
   command: z
-    .enum(['view', 'create', 'str_replace', 'insert'])
-    .describe('The operation to perform: `view`, `create`, `str_replace`, `insert`.'),
+    .enum(['view', 'create', 'str_replace', 'insert', 'pattern_replace', 'find_line', 'undo_edit'])
```
yonib05: I ran the ReDoS heuristic and it rejects `(foo|bar)+`, `(a|b)*`, and `(\w+\s*)+` while allowing `((a)+)+b` (nested parens get past the `[^()]*`). Python accepts everything TS rejects, so we'd ship one tool name with two different sets of valid inputs. Given the two also disagree on replacement syntax (`\1` vs `$1`), is `pattern_replace` pulling its weight next to `str_replace` + `replace_all`?

strands-py/src/strands/vended_tools/file_editor/file_editor.py
```diff
@@ -50,46 +76,126 @@ def make_file_editor(
...
+            is met. Defaults to 32 MB.
+        pattern_replace_timeout: Wall-clock seconds allowed for a single
+            ``pattern_replace`` invocation before the tool aborts with a
+            "possible catastrophic regex" error. Defaults to 5.0 seconds.
 
     Returns:
         A decorated tool that performs file operations through the sandbox.
     """
+    if root is not None and not posixpath.isabs(root):
+        raise ValueError(f"root must be an absolute path, got: {root}")
```
yonib05, line 105: `root` isn't in #3235, and this is a sandbox-routed tool calling host `realpath`, then falling back to string-only compares when the root doesn't exist locally. Should confinement live on `Sandbox` so bash, file_read, and code_execution get it too?

strands-py/src/strands/vended_tools/file_editor/file_editor.py
```diff
@@ -430,3 +707,165 @@ async def _handle_insert(sandbox: Sandbox, file_path: str, insert_line: int | No
...
+
+    suffix = f" ({count} matches replaced)" if replace_all and count > 1 else ""
+    return (
+        f"The file {file_path} has been edited via pattern_replace.{suffix} "
+        f"{_make_output(snippet, f'a snippet of {file_path}', start + 1)}"
+        "Review the changes and make sure they are as expected. Edit the file again if necessary."
+    )
+
+
+async def _handle_find_line(
```
yonib05, line 715: `find_line` returns only the first match and raises when there's no match. That's a rough contract for a search primitive when `bash` already covers real search. What's the case that needs it inside the editor?

yonib05 (dismissed):
Binary rejection and the `replace_all` opt-in are solid wins. Main things I'd want resolved are the shared undo state and whether the description should call out the tab-expansion change.

---

Dropping the whole-file tab expansion looks like the right fix, but doesn't it change matching for callers whose `old_str` contains tabs? Could we call it out as a behavior change in the description? It'll matter for release notes.

---

lizradway:
i havent read this pr in full, but shouldnt we use `FileStorage` that automatically routes thru sandbox.

chaynabors:
> i havent read this pr in full, but shouldnt we use `FileStorage` that automatically routes thru sandbox.

Good intuition, but file storage is a kvstore and we need the ability to modify files (view, str_replace, insert, etc). As of my last push we do route to sandbox in the ways that FileStorage would.

The systems way of doing this is to implement a cursor API and read and write to a file through traits that abstract over arbitrary streams. Strands doesn't have that kind of rigor today, and it'd be a pretty substantial endeavor to do it all correctly, and would be breaking.

JackYPCOnline:
@strandly-the-agent review this PR see if there is any secruity issue?

strandly-the-agent (dismissed):
**Answering `JackYPCOnline`'s question: yes — one real security issue.** The new `find_line(fuzzy=true)` path is a catastrophic-backtracking ReDoS in both SDKs, reachable from ordinary model tool-call arguments. Everything else I found is minor. *(Body updated in place with my own cold measurements — earlier numbers here were less precise.)*

🔴 **`find_line(fuzzy=true)` ReDoS** — `file_editor.py:457,459` / `file-editor.ts:516,519`. `search_text="a a a b"` (7 chars) against a file whose content is one long line of `a` — Python 250 chars → **2.9s**, 500 chars → **44.0s / 35.2s / 32.8s** (3 cold runs); TS through the real tool 250 → **1.4s**, 500 → **13.1s**, raw V8 at 1000 → **239s**. Superlinear in line length, no cap and no timeout.
🟡 **The size cap does not bound the read** (both SDKs) — a 200 MB file is rejected only after being fully buffered (peak RSS 268 MB), and `view /dev/zero` dies with `MemoryError` at a 512 MB `RLIMIT_AS`. `file-editor.ts:88-90` claims the opposite.

✅ All five blockers from my `84412b7a` pass are genuinely fixed (root fail-closed, preflight-before-allocate, snapshot-after-write, per-agent undo, and `pattern_replace` removed with its GIL-bound timeout).
✅ Head `43c74820` · Python 83/83, TS 87/87, no type errors, `git diff --check` clean. Neither suite exercises the shape below — `fuzzy` has one happy-path test per SDK.

<details>
<summary>🔴 ReDoS — evidence, why it is reachable, and a minimal fix</summary>

Both SDKs build the fuzzy matcher by joining escaped tokens with greedy `.*` and running it per line against content up to `max_file_size`:

- `strands-py/src/strands/vended_tools/file_editor/file_editor.py:457` → `re.compile(".*".join(re.escape(t) for t in tokens), re.IGNORECASE)`, searched per line at `:459`
- `strands-ts/src/vended-tools/file-editor/file-editor.ts:516` → `new RegExp(tokens.map(escapeRegExp).join('.*'), 'i')`, tested per line at `:519`

Escaping is correct — there is **no regex injection**. The problem is the `.*` chain: each one is a backtracking choice point, so a failing match costs ~O(line_length^tokens).

Measured through the real tool (not the regex in isolation); file content is a single line of `n` × `a`, `search_text="a a a b"`:

| n | Python (`make_file_editor`) | TypeScript (`makeFileEditor().invoke`) |
|---|---|---|
| 250 | 2.92s | 1.38s |
| 500 | 44.06s / 35.17s / 32.75s (cold ×3) | 13.08s |
| 1000 | ~min+ (extrapolated) | 239s (raw V8, same pattern) |

Two things make this worse than a slow call:

1. **The timeout you would expect does not fire.** In the Python run above the call was wrapped in `asyncio.wait_for(..., timeout=5)` and still returned only after 44s: `_find_line_numbers` is synchronous, so the loop never gets control back. In Node the regex blocks the **entire** event loop, so every other agent/tool call in that process stalls with it.
2. **No adversarial file is needed.** Any long single line does it — minified JS/CSS, a base64 blob, a lockfile line, a long log record. `search_text` is model output, so indirect prompt injection in content the agent has already read is enough to aim it, and nothing caps `search_text` length or token count.

**Minimal fix** — the documented semantics ("whitespace-tolerant, case-insensitive") only need an ordered subsequence scan, which is linear and cannot backtrack:

```python
def _fuzzy_line_match(line: str, tokens: list[str]) -> bool:
    lowered, pos = line.lower(), 0
    for token in tokens:
        found = lowered.find(token.lower(), pos)
        if found == -1:
            return False
        pos = found + len(token)
    return True
```

Same in TS with `indexOf`. Worth adding as defense-in-depth: a cap on `search_text` length / token count, and a regression test asserting a crafted worst case completes inside a short wall-clock bound.

Note this is also the one guard the rework lost: `pattern_replace` at least *attempted* a timeout; `find_line` inherited the model-supplied-regex risk with no bound at all.
</details>

<details>
<summary>🟡 The size cap bounds what is accepted, not what is read — and the TS doc says otherwise</summary>

`_read_text_or_reject_binary` (`file_editor.py:530-532`) and `readTextOrRejectBinary` (`file-editor.ts:604-607`) read the whole file, then compare against `max_size`. Measured with the default 1 MB cap:

- 200 MB regular file → correct rejection, but only after buffering: `elapsed=0.07s peakRSS=268MB`
- `view /dev/zero` under `RLIMIT_AS=512MB` → `MemoryError` at 461 MB RSS; the `asyncio.wait_for(5s)` around it never fired

Read-then-check predates this PR, so I am not treating it as a regression — but this PR newly documents it as a guarantee: `file-editor.ts:88-90` says oversize input is *"rejected with a clean error rather than being loaded into memory"*, and both READMEs describe `max_file_size` as a *"byte cap on files that can be read"*. As written those claims do not hold, and with no `root` by default the reachable path set is "anything the process can read".

Cheap fix that fits the existing flow: every read path already calls `_probe_sandbox_path` / `probeSandboxPath`, which lists the parent and gets `FileInfo.size` — reject on that metadata *before* `read_file` when the backend reports it, and soften the doc wording to match what is actually enforced.
</details>

<details>
<summary>🔵 Two lower-confidence notes, one robustness nit, and what I re-checked from last round</summary>

**TOCTOU on `root=` (modeled, not demonstrated against a shipped sandbox).** `_resolve_path` validates once, then handlers probe → read → write across `await` boundaries (`file_editor.py:640,646,654`). Against the shipped local sandbox I could not land the race (its I/O never really yields). With one `asyncio.sleep(0.05)` injected into a sandbox's write — modeling a remote round trip — a concurrent symlink swap did land a write outside `root`. It needs a second filesystem actor, so it is a limitation to document (or re-validate immediately before the mutating write / open the final component with `O_NOFOLLOW`), not a blocker.

**Undo history is keyed on the agent, not on `(agent, sandbox)`** (`file_editor.py:112`, `file-editor.ts:137`). If one agent object is reused across two sandboxes that map the same absolute path to different filesystems, `undo_edit` writes the first sandbox's snapshot into the second. Not reachable through the supported API today: Python `Agent.sandbox` is a getter with no setter, and TS only resolves the process-wide default slot that is set once at startup. Worth a comment recording the invariant, since the docstring's isolation claim quietly depends on it.

**Robustness nit:** `tool_context.agent` must now be weak-referenceable for *every* command, including read-only `view`/`find_line` (`file_editor.py:158` runs before dispatch). A `SimpleNamespace`-style context now fails with a raw `TypeError: cannot create weak reference to ...`; the test suite works around this with a purpose-built `_FakeAgent` (`test_file_editor.py:22`). A clearer error, or falling back to an id-keyed entry, would be kinder to embedders.

**Re-checked and confirmed fixed** (not re-raised): root symlink escape now fails closed for non-local roots (`file_editor.py:239-245`); projected-output preflights run before allocation (`:306-345`); snapshots are stored after the write commits and undo keeps its entry until the restore succeeds (`:654-655`, `:760-762`); undo is per-agent; `pattern_replace` and its GIL-bound "wall-clock" timeout are gone. Also refuted while probing: shell/interpreter injection (no `subprocess`/`exec` anywhere on these paths), NUL-byte paths (fail safe at the syscall), `root="/"`, backslash components, and static symlink cases through nested/symlinked ancestors — all correctly rejected.

**Scope:** security-weighted read of the diff plus adversarial probing; I did not re-review API shape or docs beyond the security claims above. Repros ran against the real tool objects in both SDKs, cold-repeated for the blocker; temporary probe files were deleted and the worktree left clean. As always, a human should confirm the fix direction.
</details>

strands-py/src/strands/vended_tools/file_editor/file_editor.py
```diff
@@ -45,51 +65,129 @@ def make_file_editor(
...
+            string-level check would leave the ``root`` guarantee unenforceable
+            against a symlink inside the sandbox. When ``root`` is ``None``,
+            only absolute-path and ``..``-traversal checks apply; the
+            underlying sandbox's symlink policy governs escape.
+        max_file_size: Maximum file size (bytes) accepted by view/edit
+            commands. Defaults to 1 MB.
+        max_undo_entries: Maximum number of distinct paths retained in the
+            in-memory undo history *per agent*. Oldest entry is evicted on
+            overflow. Defaults to 32.
+        max_undo_bytes: Approximate cap on total bytes of file content held in
```
Unshure, line 94: When do we expect this to be used by customers? In a memory constrained device? Or is this more to limit the amount of file changes an agent can make?

Unshure (commented):
Looks good to me, but a few callouts:
- Can you update the pr description to mention that we are deleting the unused exported interfaces from ts: `FileEditorOptions` and `IFileReader`
- Do you think we should update our integ tests to cover this new tool in python, and the new funcitonality in ts?
- Looks like codecov failed, can you fix that, or update the pr description on why its okay its failing?

strands-py/src/strands/vended_tools/file_editor/file_editor.py
```diff
@@ -45,51 +65,129 @@ def make_file_editor(
...
         tool_context: ToolContext,
         file_text: str | None = None,
         view_range: list[int] | None = None,
         old_str: str | None = None,
         new_str: str | None = None,
         insert_line: int | None = None,
+        search_text: str | None = None,
+        fuzzy: bool = False,
+        replace_all: bool = False,
     ) -> str:
```
mkmeral, line 134: can we also include multimodality in the update? file view should allow to view images, pdfs, etc.

Unshure: Dont think we need to block this pr on multi-modality. This is pretty big as-is and that feature would be a pretty easy incremental addition. Maybe just an issue to track adding multi-modal input?

mkmeral: > Dont think we need to block this pr on multi-modality

I disagree, being able to read images is a critical functionality that most of the LLMs offer, it's a basic feature that should exist imo. 

We also had explicit asks for it

chaynabors: We can add more features in a follow up to this 2000 line PR that's been open for 4 weeks and reviewed 8 times.

yonib05: Let's merge and create an issue to track. Yes, we can add it, even with priority, but let's treat it as incremental. Strandly can probably make that incremental change.

PR #3396, opened 2026-07-21
fix(multiagent): preserve shared context and cumulative accounting across serialize/deserialize

opieter-aws:
## Description

`Swarm` and `Graph` persist their run state via `serialize_state()` / `deserialize_state()` so a session can be resumed after a restart. Two round-trip defects meant that resuming silently discarded state.

**Swarm shared context was erased on the next save.** `serialize_state` reads shared context from the `SwarmState`-owned object while the restore path wrote it to the `Swarm`-owned object. During a live run these are the same object (`stream_async` binds them when it builds a fresh `SwarmState`), but the resume path never rebuilds that state — so after `deserialize_state`, the restored context lived on `self.shared_context` while `serialize_state` kept reading an empty `self.state.shared_context`, and the next save persisted `{}`. The fix re-establishes the object identity a live run relies on, so the serialize path and the node-input builder read the same context after resume.

**Cumulative accounting reset to zero on both orchestrators.** `serialize_state` never persisted `execution_time` / `accumulated_usage` / `accumulated_metrics` (plus `execution_count` on Graph), and the restore path never rehydrated them. After a resume these reset to their defaults, which corrupts the total-execution timeout budget (`should_continue` adds accumulated `execution_time` to fresh wall-clock) and under-reports the token/metric totals in the final result. These fields are now serialized and restored, so a resumed run continues accounting from where it left off rather than starting over.

Swarm restores three fields; Graph restores four — Graph tracks `execution_count` explicitly, whereas Swarm derives it from `node_history`, which already round-trips.

## Related Issues

Requirement for #3283

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

Added a `serialize → deserialize → serialize` round-trip regression test for each orchestrator, asserting the second serialize equals the first for shared context (Swarm) and the cumulative fields (both). Confirmed both tests fail on the pre-fix source and pass after.

Beyond the round-trip invariant, I drove the full restore → resume → run path end to end: restored a mid-run payload into a fresh instance, called `invoke_async`, and confirmed the resumed run *accumulates on top of* the restored totals (e.g. restored 33 tokens + 30 from the resumed node = 63; Graph `execution_count` 3 → 4) and that the Swarm node receives the restored shared context during execution. Ran the full multiagent unit suite and ruff.

Note: `hatch run prepare` could not run locally — the dev env fails to build an unrelated optional dependency (`cedar-policy-mcp-schema-generator`, needs rustc 1.89). The full gate will run in CI.

## Checklist

----


strands-py/tests/strands/multiagent/test_graph.py
```diff
@@ -2046,6 +2046,55 @@ async def test_graph_persisted(mock_strands_tracer, mock_use_span):
...
+    assert graph.state.execution_time == 555
+
+    serialize1 = graph.serialize_state()
+    graph.deserialize_state(serialize1)
+    serialize2 = graph.serialize_state()
+
+    assert serialize2["accumulated_usage"] == serialize1["accumulated_usage"]
+    assert serialize2["accumulated_metrics"] == serialize1["accumulated_metrics"]
+    assert serialize2["execution_count"] == serialize1["execution_count"]
+    assert serialize2["execution_time"] == serialize1["execution_time"]
```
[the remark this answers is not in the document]
opieter-aws: Keeping field-by-field. `serialize2 == serialize1` is flaky for graph.  `completed_nodes`/`failed_nodes`/`interrupted_nodes` come from sets, so list order is non-deterministic.

strands-py/src/strands/multiagent/swarm.py
```diff
@@ -993,6 +993,9 @@ def serialize_state(self) -> dict[str, Any]:
             "node_results": {k: v.to_dict() for k, v in self.state.results.items()},
             "next_nodes_to_execute": next_nodes,
             "current_task": encode_bytes_values(self.state.task),
+            "accumulated_usage": self.state.accumulated_usage,
+            "accumulated_metrics": self.state.accumulated_metrics,
+            "execution_time": self.state.execution_time,
```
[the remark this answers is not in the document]
opieter-aws: Low-impact: session managers JSON-encode synchronously, so nothing holds the dict, and shared_context already aliases the same way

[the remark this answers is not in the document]
opieter-aws: See thread above

strands-py/tests/strands/multiagent/test_graph.py
```diff
@@ -2046,6 +2046,54 @@ async def test_graph_persisted(mock_strands_tracer, mock_use_span):
...
+
```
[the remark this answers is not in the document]
opieter-aws: See thread above

mkmeral:
@strandly-the-agent review pls

strands-py/src/strands/multiagent/graph.py
```diff
@@ -1275,6 +1275,10 @@ def serialize_state(self) -> dict[str, Any]:
             "next_nodes_to_execute": next_nodes,
             "current_task": encode_bytes_values(self.state.task),
             "execution_order": [n.node_id for n in self.state.execution_order],
+            "accumulated_usage": self.state.accumulated_usage,
+            "accumulated_metrics": self.state.accumulated_metrics,
+            "execution_count": self.state.execution_count,
```
strandly-the-agent, line 1282: 🔴 **Block — crash checkpoints still lose the current invocation's timeout usage.**

This serializes only `state.execution_time`, but Graph does not add the current invocation until the outer `finally` (`graph.py:699-700`), after the per-node `AfterNodeCall` session sync (`graph.py:1139-1141`). I repeated a real `FileSessionManager` process-crash repro twice: node A took 250ms under a 150ms total timeout; the durable checkpoint recorded `completed_nodes=["a"]`, `next_nodes=["b"]`, and `execution_time=0`; after restart, B ran and the graph completed. Without restart, B is correctly rejected by the timeout. `swarm.py:998` has the same ordering.

Please serialize prior time plus elapsed time from the active invocation at per-node checkpoints, while committing that interval exactly once at finalization to avoid double-counting. A checkpoint → restart → resume timeout test for both orchestrators would lock down the behavior this PR promises.

strands-py/src/strands/multiagent/graph.py
```diff
@@ -1415,6 +1419,13 @@ def _from_dict(self, payload: dict[str, Any]) -> None:
         # Task
         self.state.task = decode_bytes_values(payload.get("current_task", self.state.task))
 
+        # Cumulative accounting: restore so the timeout budget (should_continue) and the reported
+        # totals stay correct across a resume, rather than resetting to zero.
+        self.state.accumulated_usage = _parse_usage(payload.get("accumulated_usage") or {})
+        self.state.accumulated_metrics = _parse_metrics(payload.get("accumulated_metrics") or {})
+        self.state.execution_count = int(payload.get("execution_count") or 0)
```
strandly-the-agent, line 1496: 🔴 **Block — a resumed `GraphResult` reports only the restored interval.**

Restoring this value exposes the existing result-ordering seam: `_build_result()` copies `state.execution_time` at `graph.py:1252`, then `stream_async()` adds the current invocation only afterward in `finally` (`graph.py:689-700`). In a deterministic probe, I restored 555ms, ran one node with a mocked 600ms invocation, and got `GraphResult.execution_time == 555` while `graph.state.execution_time == 1155` after return. Nested Graphs then accumulate the stale child result at `graph.py:1026-1033`.

Please finalize the current interval exactly once before building/emitting `GraphResult` (while preserving exception-path accounting), and add a restored-prior + current-runtime assertion, including nested propagation.

strandly-the-agent (changes requested):
🔴 **Changes requested:** persisted execution time still drops the current invocation at crash checkpoints, and Graph returns prior-only elapsed time after resume.

✅ Reviewed exact head `ac77e2501b701fbfedbe9de3ab55504d3e53f542` · 112 focused tests passed · changed-file Ruff/format + diff check passed · exact-head CI green.

🔴 Two execution-time lifecycle blockers are inline; both were reproduced deterministically.

<details>
<summary>Review passes and verification</summary>

- **Correctness/adversarial:** repeated the FileSessionManager process-crash repro twice; both checkpoints stored `execution_time=0` after a 250ms node, and both resumed past a 150ms total timeout. A separate mocked-clock probe restored 555ms, ran 600ms, returned 555ms, then left state at 1155ms.
- **Surviving probes:** restored usage/metrics/execution count accumulated correctly; Swarm restored shared context retained identity and reached resumed model input.
- **Tests:** `hatch test -q tests/strands/multiagent/test_graph.py tests/strands/multiagent/test_swarm.py` → **112 passed in 2.54s**.
- **Static checks:** Ruff check/format passed on all four changed files; `git diff --check` passed. The full-tree local format check found 11 unrelated files outside this diff; GitHub Python Lint and CI Gate are green at this SHA.
- **Suppressed:** resolved threads about full-dict equality, shallow copies, and evergreen docstrings; weaker test/doc observations subsumed by the reproduced defects.
- **Routing:** correctness, adversarial, test-quality, and docs passes ran; API/LLM-context were not applicable because no public/model-facing surface changed.

</details>

opieter-aws:
@strandly-the-agent re-review latest fixes

strands-py/src/strands/multiagent/graph.py
```diff
@@ -646,6 +646,7 @@ async def stream_async(
 
         # Initialize state
         start_time = time.time()
+        self._invocation_start_time = start_time
         if not self._resume_from_session and not self._interrupt_state.activated:
```
strandly-the-agent, line 649: 🔴 **Block — tracing setup failures leave the active timer running indefinitely.**

The marker starts here before `start_multiagent_span()` / `use_span(...).__enter__()` (`graph.py:664-665`), but the clearing `finally` begins only inside the entered span (`graph.py:700-702`). If either tracing operation raises, `serialize_state()` at `graph.py:1282` keeps adding wall time after the failed invocation. Swarm has the same lifecycle at `swarm.py:413-416` and `439-441`.

I reproduced span-create and span-entry failures for both orchestrators twice: committed time started at 0 with fake `t=1000`; the call raised, the marker remained `1000`, and serialization at `t=1000.5` reported 500ms. The prior head stayed at 0.

Please include tracing setup in the timing cleanup scope, or start the marker only after successful context entry. A parameterized Graph/Swarm regression should make both failure points raise, then assert the marker is cleared and later serialization does not accrue time.

strandly-the-agent (changes requested):
🔴 **Changes requested:** both prior execution-time blockers are verified fixed at `ef85b239`, but tracing setup failures leave the new active timer running and corrupt later checkpoints.

✅ Exact head `ef85b239f21c019ba16fdba1fb6720f25365bc85` · 115 focused tests passed · changed-file Ruff/format + diff check passed · exact-head CI green.

✅ **Prior blockers resolved:** the real FileSessionManager crash/restart repro now checkpoints 251ms and correctly rejects B under a 150ms total timeout (repeated twice); restored/nested timing now commits and propagates exactly once.

🔴 One new-delta lifecycle blocker is inline.

<details>
<summary>Review passes and verification</summary>

- **Correctness/adversarial:** exercised process restart, restored and nested Graph timing, repeated serialization, final-hook errors, execution errors, cancellation, early generator close, back-to-back invocations, and Swarm parity. No double-counting or marker leak occurred after entering the protected span.
- **New failure repro:** `start_multiagent_span()` and span context entry failures each leaked the marker for Graph and Swarm; repeated twice identically. At fake `t=1000`, serialization at `t=1000.5` reported 500ms after the invocation had already raised. The prior head remained at 0.
- **Tests:** `hatch test -q tests/strands/multiagent/test_graph.py tests/strands/multiagent/test_swarm.py` → **115 passed in 2.74s**.
- **Static/CI:** Ruff check and format passed on all five changed files; `git diff --check` passed; GitHub Python 3.10–3.14 matrices, integration, lint, docs, and CI Gate are green.
- **Routing:** correctness, adversarial, and test-quality passes ran. API/LLM-context/docs were not applicable. The advanced adversarial tier was unavailable under data-retention mode; its default-tier retry completed.
- **Suppressed:** resolved full-dict equality, shallow-copy, and evergreen-docstring threads; the weaker Swarm lifecycle test gap is subsumed by the reproduced tracing failure and requested regression test.

</details>

strands-py/src/strands/multiagent/graph.py
opieter-aws: Fixed

opieter-aws:
@strandly-the-agent  re-review

strandly-the-agent (approved):
✅ **Approved:** the timer-marker leak is fixed for both Graph and Swarm; no findings remain at `aca610f4b29118db4c9e20ef3a65729292ac469c`.

✅ 121 focused tests passed · Ruff/format/diff-check clean · exact-head CI green · tracing failure matrix and real crash/restart repros passed repeatedly.

<details>
<summary>Re-review evidence</summary>

- **Prior blocker:** verified resolved. Graph and Swarm now publish `_invocation_start_time` only after successful span entry; span creation and `use_span.__enter__` failures both leave the marker unset and serialization stable.
- **Cold adversarial checks:** Graph/Swarm × both tracing failure phases passed in two local runs (8/8 cases). FileSessionManager crash/restart passed twice: a 250ms node persisted 251ms, and resume correctly rejected B under a 150ms total timeout.
- **Lifecycle checks:** restored/nested totals, repeated serialization, cancellation, early close, node/final-hook errors, and back-to-back invocations showed no leaks or double-counting. Main `#3390` resume-frontier interactions also passed.
- **Merge provenance:** `aca610f4` matches the synthetic merge of fix commit `6c9ca250` with main `7b0717c4`; no conflict-resolution hunk altered the timer fix.
- **Tests/static:** `hatch test -q tests/strands/multiagent/test_graph.py tests/strands/multiagent/test_swarm.py` → **121 passed in 2.91s**. Ruff check/format and `git diff --check` passed; worktree stayed clean.
- **CI:** Python 3.10–3.14 Linux/Windows/macOS, integration, lint, docs, dependency review, and CI Gate all succeeded at this SHA.
- **Routing:** correctness, adversarial, and test-quality passes ran; API/LLM-context/docs were not applicable. The advanced adversarial tier was unavailable under its data-retention restriction; the default-tier retry completed.

</details>

strands-py/src/strands/multiagent/base.py
```diff
@@ -187,6 +188,43 @@ class MultiAgentBase(ABC):
     """
 
     id: str
+    # Wall-clock start of the active invocation, or None when no invocation is running. Set at
+    # invocation start; folded into the orchestrator's committed execution-time total exactly once
+    # at finalization.
+    _invocation_start_time: float | None = None
```
JackYPCOnline: Should it rely on __init__ other than class level default ? Not sure if this is the best design on a ABC

opieter-aws: Thanks for the suggestion, good call! Updated

JackYPCOnline (approved):

Issue #2498, opened 2026-04-22
[P1 items] Python & TypeScript SDK Parity tracking issue


notowen333:
This issue serves to track parity items which have been triaged as non-P0 features

yonib05:
## Python ↔ TypeScript parity — gap analysis (for tracking)

> ⚠️ **Source & caveat:** the following was derived from an **automated agent analysis** of the two SDK trees (`strands-py/` and `strands-ts/`) and their reference docs. It has **not been hand-verified** — treat every line as a lead to confirm against current code before acting, not as ground truth. Line/behavior claims may be stale or wrong.

Framing: **TypeScript is the newer interface** — a number of things learned in Python were applied differently in TS on purpose. We can't break Python, so feature parity is the goal *where it makes sense*, but some differences are deliberate and should be left alone. Split accordingly:

### Close — TS is missing (port Python → TS)
- Tool registry + filesystem loader + hot-reload (`ToolWatcher`) — TS has no local tool registry/loader/watcher (only MCP `onToolsChanged`).
- Sliding-window tool-pair trim fallback (`_find_tool_pair_trim_point`) — no TS equivalent.
- Concurrency/idempotency dedup + result replay (owner-aware lock) — TS has only a boolean `acquireLock()`. Partly tracked by #3018.
- Summarizing-via-sub-agent path; `per_turn` proactive management.
- Swarm repetitive-handoff detection.

### Close — Python is missing (port TS → Python)
- Backoff/jitter strategy library.
- `ExecuteToolStage` + `AgentStreamStage` middleware + middleware interrupt path (Python only wraps the model call via `InvokeModelStage`).
- `LifecycleObserver` extension point.
- Richer tool-result truncation (typed placeholders, document-source-aware).
- MCP elicitation as a first-class client feature.
- Snapshot-based session model (UUID-v7 immutable history).
- Vended tools: `http-request`, `notebook`.
- First-class logging module.

### Leave alone — deliberate divergence, NOT parity work
- Model-provider breadth (Python ships ~12 native providers; TS ships ~6 + the Vercel adapter). **Explicitly not porting providers.**
- Vercel AI SDK adapter.
- Graph/Swarm join semantics (OR/AND-join, batches vs per-node scheduling, tool-based vs structured-output handoff).
- Schema philosophy: Pydantic (Py) vs Zod (TS).
- Python sync↔async bridge (`run_async`) — required by Python's sync entrypoint; N/A in TS.
[...]

PR #3397, opened 2026-07-21
feat(vended-tools): add stop tool

### LINKED ISSUE #3242: Vended tools: stop (py + ts)
Part of #3234.

Add a `stop` vended tool in both SDKs: graceful agent loop termination with an optional message, wired to the event loop's stop semantics.

chaynabors:
Closes #3242.

Vended stop tool in both SDKs. Lets the model gracefully end the agent loop when it decides the task is complete. Both implementations shim onto an existing loop-termination primitive rather than introducing a parallel one. The Python side sets `invocation_state["request_state"]["stop_event_loop"] = True`, which the event loop already checks after tool execution. The TypeScript side records a marker on `invocationState` and lazily installs one `AfterToolsEvent` hook per agent (tracked in a `WeakSet`) that reads the marker and sets `event.endTurn`, which the agent loop already treats as a cooperative stop.

The tool accepts an optional message, capped at 4096 characters. Empty and null messages fall back to a default so a JSON-null-serialized field or a truthy-falsy check does not silently prevent the halt. This is a cooperative stop, not an abort: any tools the model requested alongside stop in the same turn still run to completion. For hard cancellation, callers use `agent.cancel()`.

There is a small observable difference between the two SDKs. TypeScript's `AfterToolsEvent.endTurn` synthesizes a new assistant message with the stop text, so `AgentResult.lastMessage` carries it. Python's `stop_event_loop` short-circuits inside the current tool-use cycle, so `AgentResult.stop_reason` is `tool_use` and the stop text appears in history as the tool's result rather than a new assistant turn. That divergence is documented in the README; aligning it would require reshaping `event_loop.py`, which is out of scope for a vended tool.


arielnabavian (dismissed):

strands-py/src/strands/vended_tools/stop/stop.py
```diff
@@ -0,0 +1,103 @@
...
+        the assistant-facing final turn is never blank.
+
+    Raises:
+        ValueError: If ``message`` is not a string or exceeds the length cap.
+    """
+    if message is None or message == "":
+        return _DEFAULT_MESSAGE
+    if not isinstance(message, str):
+        raise ValueError(f"`message` must be a string, got {type(message).__name__}")
+    if len(message) > _MAX_MESSAGE_LENGTH:
```
Unshure: Do we need this check? If we do can we either make it larger, or configurable?

strands-py/src/strands/vended_tools/stop/stop.py
```diff
@@ -0,0 +1,103 @@
...
+
+        Args:
+            tool_context: Injected by the framework. Not user-facing.
+            message: Optional final message describing why the loop is ending.
+                Capped at 4096 characters; longer values are rejected.
+        """
+        final_message = _validate_message(message)
+        request_state = tool_context.invocation_state.setdefault("request_state", {})
+        request_state["stop_event_loop"] = True
+        return final_message
```
Unshure, line 111: Curious about your thoughts on normalizing how tools can stop the event loop between python and typescript? Not a blocker for this pr, but this might be an opportunity to implement the same functionality between the two sdks. Do you think it is worth it?

strands-py/src/strands/vended_tools/stop/stop.py
chaynabors: Made it configurable via `make_stop(max_message_length=...)` (Python) and `makeStop({ maxMessageLength: ... })` (TypeScript). Default stays at 4096 so the runaway-model guard is on by default, but callers who want longer summaries can bump it. Added tests for the relaxed cap and for rejecting non-positive values at factory time.

strands-py/src/strands/vended_tools/stop/stop.py
chaynabors: Worth it, but as its own follow-up.

* Python signals via `invocation_state["request_state"]["stop_event_loop"]` today and ends with `stop_reason="tool_use"`.
* TypeScript uses `AfterToolsEvent.endTurn` and ends with `stopReason="endTurn"`.

Normalizing means picking a single primitive across both SDKs (probably a lifecycle hook that fires after tools) and either changing the Python event loops termination shape or adding a shim. Both are more invasive than a vended tool. Ill leave this PR shimmed onto whats already there and open a separate issue so I'm not wasting time on design reviews.

Unshure (approved):

Unshure:
Can you look into the failing integ tests?

site/src/content/docs/user-guide/concepts/tools/vended-tools-imports.ts
zastrowm, line 1: > AfterToolsEvent.endTurn synthesizes a new assistant message with the stop text, so AgentResult.lastMessage carries it. Python's stop_event_loop short-circuits inside the current tool-use cycle, so AgentResult.stop_reason is tool_use and the stop text appears in history as the tool's result rather than a new assistant turn. That divergence is documented in the README; aligning it would require reshaping event_loop.py, which is out of scope for a vended tool.

Can we make it a pre-dependency?  If possible, I'd like to align behavior where possible - did we explore what it would take?

strands-ts/src/vended-tools/stop/stop.ts
```diff
@@ -0,0 +1,128 @@
...
+  buildStopInputSchema,
+} from './types.js'
+
+/**
+ * Tracks agents on which the terminal `AfterToolsEvent` hook has been
+ * installed, so we install it at most once per agent instance regardless of
+ * how many times the model calls `stop`. Keyed weakly so garbage-collected
+ * agents drop out automatically.
+ */
+const installedAgents = new WeakSet<LocalAgent>()
```
zastrowm, line 18: WeakDict and WeakSets are usually workarounds for missing functionality.

A better approach would be to plugin-ize this and ensure we have `addPlugin` + `hasPlugin`/`getPlugin` (which has been requested before) or finding an alternative way of detecting this - but IMO weakset shouldn't be needed

site/src/content/docs/user-guide/concepts/tools/vended-tools.mdx
```diff
@@ -197,6 +200,48 @@ agent("List all Python files in the current directory and count them")
...
+
+_Supported in: Node.js, modern browsers (TypeScript); all platforms (Python)._
+
+This is a cooperative stop, not an abort. Any other tools the model requested in the same turn still run to completion; the loop halts after that batch without calling the model again. The final message defaults to a 4096-character cap; pass `max_message_length` / `maxMessageLength` to `make_stop` / `makeStop` when a longer summary is legitimate.
+
+The two SDKs shim onto different loop-termination primitives, which produces a small difference in the final `AgentResult`. TypeScript halts via `AfterToolsEvent.endTurn` and returns `stopReason: "endTurn"` with the stop text as the last assistant message. Python halts via `invocation_state["request_state"]["stop_event_loop"]` and returns `stop_reason: "tool_use"` with the model's tool-use message as the final message; the stop text lives in history as the tool result, not as a new assistant turn.
```
arronbailiss, line 211: can we address this parity gap as part of https://github.com/strands-agents/harness-sdk/issues/2498 ?

PR #3398, opened 2026-07-21
feat(vended-tools): add tool_registry CRUD

### LINKED ISSUE #3249: Vended tools: tool_registry CRUD (py + ts)
Part of #3234.

Add a `tool_registry` vended tool in both SDKs: CRUD over the agent's tool registry.

- create: register a tool at runtime (`register_dynamic_tool` / `ToolRegistry.register`)
- read: list tools and specs (`get_all_tool_specs` / `all()`)
- update: re-register with a new definition
- delete: unregister (`dynamic_tools` removal / `remove(name)`)

Design questions to settle first: what "create" accepts in each language (file path, inline source, MCP reference), and delete guardrails so an agent can't remove developer-registered tools.

chaynabors:
Closes #3249.

Vended tool_registry in both SDKs. Gives an agent CRUD access to its own tool registry at runtime: list currently registered tools, create a new binding to a tool hosted on a pre-approved MCP server, update an existing binding, and delete one previously registered by this same tool instance. The four operations sit behind a single tool with an operation discriminator, so the model sees a small consistent surface. Python shims onto `ToolRegistry.register_dynamic_tool` and `MCPClient.list_tools_sync`; TypeScript shims onto `ToolRegistry.add`/`addOrReplace`/`remove`/`list` and `McpClient.listTools`. No new runtime dependencies.

The design question in the sub-issue was what create should accept. Three shapes were on the table: a filesystem path (loading arbitrary Python from disk), inline source code (executing arbitrary code from a model-authored string), or a reference to an already-connected MCP server. Only the third ships, and it is tightened further: the model does not choose an MCP URL, it picks from an `mcp_clients` allow-list the developer supplied at tool construction, keyed by a stable alias. Consumers who need path-loading or inline-source-execution should build a separate, more locked-down tool.

Tool names must match `^[a-zA-Z_][a-zA-Z0-9_]{0,63}$`. Update and delete only touch tools this same instance registered, tracked per-agent (a `weakref.WeakKeyDictionary` keyed on the `ToolRegistry` in Python, a `WeakMap` keyed on the `LocalAgent` in TypeScript), so two agents sharing one factory cannot see each other's registrations. A configurable cap (default thirty-two) bounds registrations per instance. Under concurrent creates the cap is enforced by a synchronous reservation before the MCP lookup; a concurrent delete that lands on a reservation causes the still-awaiting create to abort rather than resurrect an entry it can no longer track. The same guard also runs on update.

### DIFF
--- a/strands-py/src/strands/vended_tools/tool_registry/tool_registry.py
+++ b/strands-py/src/strands/vended_tools/tool_registry/tool_registry.py
@@ -27,6 +27,7 @@ is exactly whatever the developer-approved MCP servers already expose.
 
 from __future__ import annotations
 
+import asyncio
 import logging
 import weakref
 from typing import TYPE_CHECKING
@@ -93,18 +94,18 @@ async def _resolve_mcp_tool(
 ) -> MCPAgentTool:
     """Look up a specific tool on the MCP client and adapt it to the local name.
 
-    Async so that the caller yields to the event loop across the (potentially
-    network-bound) MCP list-tools call. In production the body runs to
-    completion without suspending; the ``async`` shape exists so tests can
-    swap in a fake that yields with ``await asyncio.sleep(0)`` to exercise
-    concurrency at the tool level.
+    ``list_tools_sync`` is a blocking network call, so it is dispatched to a
+    worker thread with :func:`asyncio.to_thread`; without that the whole event
+    loop would stall on the round-trip. The suspension point is real, which is
+    what makes the reservation and concurrent-delete/update guards reachable in
+    production the same way the tests exercise them.
 
     Raises:
         ToolRegistryError: If ``remote_name`` is not exposed by the MCP server.
     """
     pagination_token: str | None = None
     while True:
-        page = client.list_tools_sync(pagination_token=pagination_token)
+        page = await asyncio.to_thread(client.list_tools_sync, pagination_token=pagination_token)
         for mcp_tool in page:
             # ``list_tools_sync`` returns MCPAgentTool objects; the underlying
             # ``mcp_tool.name`` is the remote-side name, unaffected by any
--- a/strands-py/src/strands/vended_tools/tool_registry/tool_registry.py
+++ b/strands-py/src/strands/vended_tools/tool_registry/tool_registry.py
@@ -1,0 +1,331 @@
+"""Tool for CRUD over the agent's own tool registry.
+
+Provides :func:`make_tool_registry` (a factory that binds the tool to a set of
+already-connected MCP clients from which new tools may be pulled). The tool
+exposes four operations to the model:
+
+- ``list``: enumerate all tools currently on the agent's registry, marking the
+  ones this tool has registered itself.
+- ``create``: register a new tool that is a thin binding to a specific tool
+  hosted on one of the pre-approved MCP clients.
+- ``update``: re-bind a previously registered tool (e.g. to point at a
+  different remote tool, or to change its description). The local name is the
+  lookup key and is not modifiable; delete and re-create to rename.
+- ``delete``: unregister a tool that this same tool_registry instance
+  registered. Developer-registered tools are never removable.
+
+Design decision (see PR body): ``create`` only accepts references to already-
+connected MCP servers pre-provided by the developer via ``mcp_clients``.
+Accepting a filesystem path (loading arbitrary Python from disk) or inline
+source code (executing arbitrary Python or JS) are explicitly out of scope --
+both introduce a filesystem/code-execution attack surface that is
+disproportionate to the value of runtime tool registration for a general-purpose
+agent tool. Consumers that need those flows should build a separate, more
+locked-down tool. This one is a thin shim onto MCPClient, so its blast radius
+is exactly whatever the developer-approved MCP servers already expose.
+"""
+
+from __future__ import annotations
+
+import logging
+import weakref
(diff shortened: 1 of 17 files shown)


arielnabavian (dismissed):

strands-py/src/strands/vended_tools/tool_registry/tool_registry.py
```diff
@@ -0,0 +1,331 @@
+"""Tool for CRUD over the agent's own tool registry.
+
+Provides :func:`make_tool_registry` (a factory that binds the tool to a set of
+already-connected MCP clients from which new tools may be pulled). The tool
+exposes four operations to the model:
+
+- ``list``: enumerate all tools currently on the agent's registry, marking the
```
arronbailiss, line 7: nice that we have this behind a simple CRUDL interface, as ToolRegistry is in strands-ts. Not a fan of all of the extra logic in here because strands-py and strands-ts ToolRegistry differ, so I created https://github.com/strands-agents/harness-sdk/issues/3485

strands-py/src/strands/vended_tools/tool_registry/tool_registry.py
```diff
@@ -0,0 +1,331 @@
...
+
+def _validate_tool_name(name: str) -> None:
+    if not isinstance(name, str) or not TOOL_NAME_PATTERN.match(name):
+        raise ToolRegistryError(f"invalid tool name '{name}': must match {TOOL_NAME_PATTERN.pattern}")
+
+
+def _find_normalized_conflict(registry: ToolRegistry, name: str) -> str | None:
+    """Return an existing registry name that collides with ``name``.
+
+    The SDK's normalized-name rule treats ``-`` and ``_`` as equivalent.
```
arronbailiss, line 74: can we correct this in the SDK too for consistency if it's not intentional? Seems like a paper cut

awsarron (changes requested):

strands-py/src/strands/vended_tools/tool_registry/tool_registry.py
```diff
@@ -0,0 +1,331 @@
...
+  ones this tool has registered itself.
+- ``create``: register a new tool that is a thin binding to a specific tool
+  hosted on one of the pre-approved MCP clients.
+- ``update``: re-bind a previously registered tool (e.g. to point at a
+  different remote tool, or to change its description). The local name is the
+  lookup key and is not modifiable; delete and re-create to rename.
+- ``delete``: unregister a tool that this same tool_registry instance
+  registered. Developer-registered tools are never removable.
+
+Design decision (see PR body): ``create`` only accepts references to already-
```
arronbailiss, line 17: one of the use cases for a tool_registry tool is enabling "meta agents" that are able to dynamically create, load, unload tools at runtime. For example, an agent that starts only with file_write and tool_registry tools is able to self-mutate and add tools to itself for any use case. Worth discussing together with the team

strands-ts/src/vended-tools/tool-registry/tool-registry.ts
```diff
@@ -0,0 +1,319 @@
...
+  remoteName: string,
+  localName: string,
+  descriptionOverride?: string
+): Promise<McpTool> {
+  const tools = await client.listTools()
+  const found = tools.find((t) => t.name === remoteName)
+  if (!found) {
+    throw new ToolRegistryError(`tool '${remoteName}' not found on MCP server`)
+  }
+  return new McpTool({
```
yonib05, line 68: `resolveMcpTool` builds `McpTool({ name: localName })` but `callTool` sends `tool.name`, so if `remoteName !== toolName` doesn't this call the wrong name on the wire? Python keeps `mcp_tool.name` for the wire. The fixture comment says `callTool` is never exercised in these tests, which would explain why nothing caught it.

strands-py/src/strands/vended_tools/tool_registry/tool_registry.py
```diff
@@ -0,0 +1,331 @@
...
+    completion without suspending; the ``async`` shape exists so tests can
+    swap in a fake that yields with ``await asyncio.sleep(0)`` to exercise
+    concurrency at the tool level.
+
+    Raises:
+        ToolRegistryError: If ``remote_name`` is not exposed by the MCP server.
+    """
+    pagination_token: str | None = None
+    while True:
+        page = client.list_tools_sync(pagination_token=pagination_token)
```
yonib05: `_resolve_mcp_tool` is async and the docstring says it yields to the event loop, but `list_tools_sync` bottoms out in a blocking `.result()`. If it never actually suspends, are the reservation and concurrent-delete/update guards reachable in production? The tests that cover them inject a fake client with `await asyncio.sleep(0)`, which is what made me look.

yonib05 (commented):
We call `list_tools_sync` without `tool_filters` or `prefix`, so a developer who narrowed the surface with a per-call filter doesn't get that protection here. Should the factory take an explicit allowlist rather than trusting whatever the client was constructed with?

---

Since `create` can only bind a tool from an already-connected, developer-approved MCP client, what does the model get here that `Agent(tools=[client])` doesn't already give it, other than lazy schema loading? Wondering if `list` covers day one and the mutation half waits for someone asking for it.

---

#3249 asked us to settle what `create` accepts before building, and this landed on MCP-reference-only, which I like better than file paths or inline source. Can we write that back onto the issue though? Right now the decision only lives in this PR description.

---

Issue #3389, opened 2026-07-21
OpenAI Responses adapter drops reasoning across multi-turn (no output-item round-trip)


pgrayy:
### SDK Language

TypeScript

### Problem Statement

The OpenAI Responses adapter drops reasoning content when replaying multi-turn conversations. In `formatResponsesMessages` (`strands-ts/src/models/openai/responses-adapter.ts`), a `reasoningBlock` is discarded with a warning:

```
block_type=<reasoningBlock> | reasoning content is not yet supported in multi-turn conversations with the responses api
```

For reasoning models, dropping prior reasoning across turns loses context OpenAI expects to be preserved. OpenAI's conversation-state guidance is to replay the model's own output items verbatim, including the encrypted reasoning item, rather than reconstruct them from text.

### Proposed Solution

Support the full output-item round-trip form on the Responses `input` array, so an assistant turn is replayed as OpenAI produced it:

1. A `ResponseReasoningItem` (`type: 'reasoning'`, `id`, `summary`, optional `encrypted_content`), preserved unchanged.
2. A `ResponseOutputMessage` (`type: 'message'`, `id`, `status`, `role: 'assistant'`, `content: [output_text with annotations]`).

This requires capturing metadata the adapter currently discards:

- Request side: set `include: ['reasoning.encrypted_content']` so OpenAI returns the encrypted reasoning blob.
- Inbound stream mapper (`mapResponsesEventToSDK`): capture the reasoning item `id` and `encrypted_content`, and the assistant message `id`/`status` plus per-text-part `annotations`.
- Type model: extend `ReasoningBlockData` (`strands-ts/src/types/messages.ts`, currently only `text`/`signature`/`redactedContent`) or add an OpenAI-specific carrier to store `id` and `encrypted_content`; associate captured annotations back to a reconstructable output message.
- Emit these as full output items on replay instead of dropping/collapsing.

Do the same in the Python SDK (`strands-py/src/strands/models/openai_responses.py`) for parity.

### Use Case

Multi-turn agent loops against OpenAI reasoning models (and Bedrock Mantle-hosted equivalents) where preserving reasoning across turns improves quality and, for some models (e.g. gpt-5.3-codex and beyond), avoids performance degradation from dropped reasoning/`phase`.

### Alternatives Solutions

- Continue dropping reasoning (status quo): simplest, but loses context and is lossy for reasoning models.
[...]

PR #3290, opened 2026-07-16
fix(openai): surface Responses stream failures

didhd:
## Description

The TypeScript OpenAI Responses adapter silently ignored terminal `response.failed` and `error` stream events. Since the adapter's default stop reason is `endTurn`, it then finalized a failed stream as an empty successful response. Applications saw no answer and no model error.

This was reproduced against Bedrock Mantle with a `response.failed` event containing:

```
prompt tokens (320666) exceed customer model maximum (278528) for model-id
```

This change throws terminal stream failures while preserving the provider message and code. The existing OpenAI error classifier can then map rate limits and context overflows to the SDK's standard model errors, so agents and applications handle them consistently with other model adapters. The Bedrock Mantle context-limit phrase is recognized as `ContextWindowOverflowError`.

There is no public API change and no change to request formatting.

## Related Issues

None.

## Documentation PR

No documentation change is needed; this fixes error propagation in the existing Responses API adapter.

## Type of Change

Bug fix

## Testing

- Focused OpenAI Responses/error unit tests: 61 passed
- `npm run type-check` and `npm run build` pass
- Pre-commit suite (build, coverage, lint, format check, type-check) passed
- New unit tests cover: `response.failed` mapped to `ContextWindowOverflowError` via the Mantle phrase, the fallback message when no error detail is present, and a Responses `error` event mapped through the existing classifier

- [x] I ran the TypeScript equivalents of the required checks

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have reviewed and understand every line of code in this PR, including any generated by AI tools, and I can explain why it works
- [x] My change is focused and reasonably small; I have split unrelated work into separate PRs
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [x] I have updated the documentation accordingly
- [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3401, opened 2026-07-21
fix(models): serialize assistant text history as valid Responses input shape

### LINKED ISSUE #3388: OpenAIModel (Responses, TypeScript) sends invalid assistant history shape, rejected by Bedrock Mantle
### SDK Language

TypeScript

### Steps to Reproduce

The OpenAI Responses adapter serializes assistant history into an invalid `input` item shape when replaying multi-turn conversations without server-side state (`stateful: false`).

For a two-turn conversation, `formatResponsesMessages` in `strands-ts/src/models/openai/responses-adapter.ts` emits the assistant turn as:

```json
{ "role": "assistant", "content": [{ "type": "output_text", "text": "..." }] }
```

This matches neither valid OpenAI `input` shape for an assistant turn:
- `EasyInputMessage` (string content): `{ "role": "assistant", "content": "..." }`
- `ResponseOutputMessage` (full item): requires `type: "message"`, `id`, `status`, and `output_text` with `annotations`.

The emitted object is a bare `{ role, content: [output_text] }` missing all the `ResponseOutputMessage` required fields, so it is not a valid input item. The code cast it away with `as unknown as ResponseInputItem`, and its own comment acknowledged the shape was incomplete.

Reproduce with the Bedrock Mantle integ test `test/integ/models/openai/mantle.test.node.ts > handles reasoning content across multi-turn conversations` (requires live Bedrock Mantle credentials).

### Expected Behavior

Text-only assistant history is serialized as a valid, OpenAI-documented shape and accepted by both OpenAI-hosted and Bedrock Mantle.

### Actual Behavior

Bedrock Mantle's strict Responses schema rejects the request:

```
SubmitRequestFailure: code=-32602, msg=220 validation errors for ResponsesRequest
```

OpenAI-hosted tolerates the malformed shape (fills in / ignores the missing fields), so the bug was invisible there. It only surfaced once PR #3290 made terminal `response.failed` stream events throw instead of finalizing as an empty successful response, which turned the previously-swallowed Mantle rejection into a loud failure.

### Possible Solution

N/A

### Related Issues

The Python SDK has the identical bug at `strands-py/src/strands/models/openai_responses.py` (`_format_request_message_content` emits `output_text` for assistant text); fix both for parity.

Surfaced by #3290.

agent-of-mkmeral:
## Description

The OpenAI Responses adapters in **both SDKs** serialize assistant history into an invalid `input` item shape when replaying multi-turn conversations without server-side state (`stateful: false`, the default):

```json
{ "role": "assistant", "content": [{ "type": "output_text", "text": "..." }] }
```

This matches neither valid assistant *input* shape: it isn't a string-content `EasyInputMessage`, and it's missing the `id`/`status`/`annotations` a full `ResponseOutputMessage` requires. OpenAI-hosted silently tolerates it; strict backends like Bedrock Mantle reject the request (`code=-32602, 220 validation errors for ResponsesRequest`).

### Fix (parity change in both SDKs)

Text-only assistant history is now serialized as the string `EasyInputMessage` form — the only assistant input shape that is valid without the output-item metadata the adapters have already discarded:

```json
{ "role": "assistant", "content": "..." }
```

- Multiple assistant text blocks join with `\n` (consistent with text-only tool-result joining).
- Non-text assistant content is dropped with a warning — it has no valid input shape without fabricated metadata. Verbatim output-item round-tripping (which also preserves reasoning) is tracked in #3389; this minimal fix deliberately stays out of that scope.
- Tool-call/tool-result items are untouched (already valid separate items).
- **TS drive-by in the same function:** `citationsBlock` pushed `output_text` unconditionally, even for `role: 'user'` (invalid on the user path too). It is now role-aware, matching the Python side.
- **TS:** the `as unknown as ResponseInputItem` cast that existed solely to erase the invalid assistant shape is gone from that path; `tsc` now guards the assistant item shape again.

Files:
- `strands-ts/src/models/openai/responses-adapter.ts`
- `strands-py/src/strands/models/openai_responses.py`

## Related Issues

Fixes #3388
Related: #3389 (reasoning/output-item round-trip follow-up), surfaced by #3290

## Testing

**Unit (updated + new regression tests)**
- Python: `hatch run test` on `tests/strands/models/test_openai_responses.py` — 103/103 pass. Rewrote the fixtures that previously asserted the buggy shape; added a multi-block join test and a wire-shape guard that validates every formatted history item against the official `ResponseInputItemParam` schema (`pydantic.TypeAdapter`). Verified the guard rejects the old buggy shape and accepts the new one.
- TS: `responses.test.ts` — 49/49 pass, including a new two-turn regression test asserting the assistant turn serializes as `{ role: 'assistant', content: '4' }` and a multi-block join test. Full unit suite (3794 tests), lint, format, `type-check`, and build pass (pre-commit hook ran the full gate).

**E2E (live)**
- **OpenAI-hosted Responses API** (`gpt-4o-mini`): 3-turn stateless conversation (history replay each turn) and a tool-call-interleaved 2-turn conversation — both SDKs, all turns accepted. ✅
- **TS live integ suite** `test/integ/models/openai/responses.test.ts`: 12/12 pass against OpenAI-hosted. ✅
- **Bedrock Mantle**: could not exercise from this environment — the runner's IAM role is denied `bedrock-mantle:CreateInference`. The official-SDK schema validation above encodes the same contract Mantle enforces; the repo's Mantle integ tests (`test_model_mantle.py`, `mantle.test.node.ts`) will exercise it in CI with proper credentials.

---

### DIFF
--- a/strands-py/src/strands/models/openai_responses.py
+++ b/strands-py/src/strands/models/openai_responses.py
@@ -634,12 +634,27 @@ class OpenAIResponsesModel(Model):
             ]
 
             if formatted_contents:
-                formatted_messages.append(
-                    {
-                        "role": role,  # "user" | "assistant"
-                        "content": formatted_contents,
-                    }
-                )
+                if role == "assistant":
+                    # The string EasyInputMessage form is the only valid assistant *input*
+                    # shape without the full ResponseOutputMessage metadata
+                    # (id/status/annotations), which the adapter does not retain.
+                    # Non-text assistant history has no valid input shape either; verbatim
+                    # output-item round-tripping is tracked in
+                    # https://github.com/strands-agents/harness-sdk/issues/3389.
+                    if any(part.get("type") != "output_text" for part in formatted_contents):
+                        logger.warning(
+                            "non-text assistant history content has no valid responses api input shape | dropping"
+                        )
+                    text = "\n".join(part["text"] for part in formatted_contents if part.get("type") == "output_text")
+                    if text:
+                        formatted_messages.append({"role": "assistant", "content": text})
+                else:
+                    formatted_messages.append(
+                        {
+                            "role": role,
+                            "content": formatted_contents,
+                        }
+                    )
 
             formatted_messages.extend(formatted_tool_calls)
             formatted_messages.extend(formatted_tool_messages)
--- a/strands-py/tests/strands/models/test_openai_responses.py
+++ b/strands-py/tests/strands/models/test_openai_responses.py
@@ -300,7 +300,7 @@ def test_format_request_messages(system_prompt):
         },
         {
             "role": "assistant",
-            "content": [{"type": "output_text", "text": "call tool"}],
+            "content": "call tool",
         },
         {
             "type": "function_call",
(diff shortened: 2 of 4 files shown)


opieter-aws:
Duplicate of #3399

Issue #537, opened 2025-07-25
[FEATURE] `force_stop` method in the `MCPClient` class


jmmfcoutinho:
### Problem Statement

There should be a way to force stop the client to release resources without waiting for tasks to complete, because tasks might stall and "never" finish, so there should be a way around this.

**Example of where this could be useful:**
I am currently building an agent that uses tools to surf the web (playwright mcp) 
I initialize the client, call client.call_tool_sync to load the website, resize, take a screenshot, and then pass this client to the callback of the agent to take screenshots after each interaction (predictable implementation, without polluting the agent's logic).

The tool that loads the website, on some websites, gets stuck indefinitely (bad implementation from playwright, I guess), making the process stall.

To come around this problem, I tried to implement a timeout, after which I'd stop the client, spin a new instance of the MCPClient and retry.

However, because the tool was "still running", even though stalled, on the first client, I was unable to release the resources from the client's thread using the client.stop(), because it awaits task completion.

I hope I was clear enough. Let me know if you need further detail.

### Proposed Solution

There should be a `force_stop` method in the MCPClient class on the mcp_client.py file.

### Use Case

People would use this to force stop the client and release resources when tasks fails to end.

dbschmigelski:
Hi,

We recently made it easier, definitely not great but easier, to configure timeouts on the MCPAgentTools - https://github.com/strands-agents/sdk-python/pull/1184

So if you wanted a tool to have a timeout you could set the `agent_tool.timeout = ...` property.

----

That being said, the idea of cancellation, not just related to MCP but in general for tools, as raised in #1275 and #501 is a larger issue Strands wants to tackle. 

Let us know if the timeout helps you in the short term.


Thanks

PR #3402, opened 2026-07-21
feat(python): add per-call MCP tool cancellation

### LINKED ISSUE #3276: [PORT] Per-call MCP tool cancellation (TS→Py)
Parent: #1659

## Source SDK

strands-ts

## Target SDK

strands-py

## Related Files

```
strands-ts/src/mcp/client.ts
strands-ts/src/tools/mcp-tool.ts
```

## What needs to be migrated

- `McpCallToolOptions.signal` (AbortSignal) on tool calls (`client.ts:65-68`)
- Wired automatically from `toolContext.agent.cancelSignal` (`mcp-tool.ts:51-53`)
- Unit coverage: `client.test.ts:357-389`

Python today: no per-call cancel. Pending invocations abort only when the whole session closes (`mcp_client.py:1141-1155`, the #995 fix). Longstanding related ask: #537 (`force_stop`).

## Gotchas / dependencies

- Python's sync facade runs calls on a background thread with a private asyncio loop; a cancellation token has to bridge threads (likely a `threading.Event` or contextvar checked on the loop side, translated to anyio cancellation of the in-flight request).
- Semantics should match TS: cancel the single in-flight call, leave the session usable.
- Independent of the 2026-07-28 RC; safe to do now.

## Confirmation

strandly-the-agent:
## Description

Ports TypeScript per-call MCP tool cancellation to Python and keeps cancellation-vs-retry behavior aligned in both SDKs.

- adds keyword-only `cancel_signal: threading.Event | None` to Python sync and async MCP calls
- forwards `agent.cancel()` into active MCP calls and stops before later sequential tools/model turns
- sends request-scoped `notifications/cancelled` or `tasks/cancel`
- bounds and owns delayed/cancellation-resistant cleanup through session shutdown
- returns `cancelled=True` with explicit local-vs-remote semantics
- makes hook-requested tool retry terminal after cancellation in both Python and TypeScript

### Latest review-driven revision (`7d88f3f8`)

- removed the direct per-call MCP section from the agent-loop guide because it exposed implementation detail
- generalized the `strands-py/AGENTS.md` cancellation rule instead of documenting this feature specifically
- clarified why the Python bridge polls `threading.Event`: it has no async notification hook, while executor-based `Event.wait()` cannot be interrupted and can strand workers after successful calls
- clarified that bounded `asyncio.wait(..., timeout=1)` returns unfinished work in `pending` and does not raise `TimeoutError`; the original caller cancellation or exception continues
- added a cancellation-resistant invocation regression proving `CancelledError` still bubbles while cleanup is retained and later drained

## Related Issues

Resolves #3276
Related: #537

## Documentation PR

Included at `site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx`.

## Type of Change

New feature

## Port evidence

<details><summary>Cross-SDK behavior</summary>

| Behavior | Coverage | Result |
|---|---|---|
| Direct Python MCP cancellation | pre-set/in-flight/racing signals, exact request ID, concurrent isolation, missing-private-ID fallback, session reuse | PASS |
| Python MCP task cancellation | delayed/resistant creation, remote cancellation timeout/failure, shutdown drain, session reuse | PASS |
| Python agent cancellation | adapter forwarding, no retry, no later sequential tool, no extra model turn | PASS |
| TypeScript cancellation-vs-retry parity | executed-tool and hook-cancel paths stop retrying when `cancelSignal.aborted`; ordinary retries remain covered | PASS |
| Cancellation-resistant Python cleanup | caller still receives `CancelledError`; unfinished invocation is retained and later drained | PASS |

</details>

### Dependency / capability delta

- **New dependencies/version changes:** none; existing `mcp>=1.23.0,<2.0.0` range is unchanged.
- **New capability:** direct Python sync/async MCP calls accept caller-owned per-call cancellation; agent cancellation reaches active MCP calls.
- **Protocol delta:** direct calls may emit `notifications/cancelled`; task calls may issue `tasks/cancel`.
- **Credential/filesystem/subprocess/deserialization delta:** none.

### Open questions

None.

## Independent review ledger

- Python round 1: 5 findings (cleanup lifecycle, cancellation errors/request IDs, checkpoint/docs examples) — fixed and regression-tested.
- Python round 2: 3 findings (remaining sequential tools, cleanup ownership, stronger request-ID proof) — fixed and regression-tested.
- Python round 3: 1 finding (retain cancellation-resistant direct invocation during shutdown) — fixed and regression-tested.
- TypeScript parity revision at `57a5b6ff`: fresh review **APPROVE**, no findings.
- Murat-driven revision at `7d88f3f8`: fresh review **APPROVE**, no findings; polling rationale, timeout propagation, resistant cleanup, and documentation scope were checked.

## Testing

Python at `7d88f3f840350a555cd8727f863330bc046dbd4c`:

```console
$ hatch test -q
4507 passed, 38 warnings in 58.38s

$ hatch run hatch-static-analysis:lint-check
All checks passed!
Success: no issues found in 223 source files
```

TypeScript at `7d88f3f840350a555cd8727f863330bc046dbd4c`:

```console
$ npm test
138 test files passed
3808 tests passed

$ npm run build && npm run type-check
# passed
$ npm run lint
# passed
$ npm run format:check
All matched files use Prettier code style!
```

`git diff --check` passed. The Python warnings are existing resource warnings in the test suite. CI is running on the latest head.

## Checklist
- [x] I have added tests proving the fix/feature works
- [x] I have updated documentation and examples

----

### DIFF
--- a/site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
+++ b/site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
@@ -114,12 +114,13 @@ The `agent.cancel()` method provides a way to stop the loop from outside, such a
 <Tabs>
   <Tab label="Python">
 
-  The agent checks for cancellation at two checkpoints:
+  The agent checks for cancellation at three checkpoints:
 
   | Checkpoint | Behavior | Note |
   |---|---|---|
   | Model response streaming | Partial output is discarded | Usage metrics may be inaccurate since the stream is closed before the model sends its final metadata event |
   | Before tool execution | Tool calls are skipped with error results added to maintain valid conversation state | |
+  | During MCP tool execution | The in-flight MCP request is cancelled without closing the shared MCP session | Remote cancellation is best-effort; the agent stops locally even if the server does not acknowledge cancellation |
 
   The agent returns a result with `stop_reason="cancelled"`. `cancel()` is thread-safe and idempotent. Calling it multiple times or from different threads is safe.
 
@@ -148,6 +149,41 @@ The `agent.cancel()` method provides a way to stop the loop from outside, such a
   if result.stop_reason == "cancelled":
       print("Agent was cancelled due to timeout")
   ```
+
+  #### Per-call MCP cancellation
+
+  `MCPClient.call_tool_sync()` and `MCPClient.call_tool_async()` also accept a caller-owned
+  `threading.Event` through the keyword-only `cancel_signal` parameter. Setting the event cancels
+  only that call; the MCP session remains available for later calls. A locally observed cancellation
+  returns an error result with `cancelled=True`. This marker does not guarantee that remote execution
+  stopped, because protocol cancellation is best-effort.
+
+  ```python
+  import asyncio
+  import threading
+
+  from strands.tools.mcp import MCPClient
+
+  cancel_signal = threading.Event()
+
+  # Inside an active MCPClient context:
+  sync_result = client.call_tool_sync(
+      tool_use_id="sync-call",
+      name="long_running_tool",
+      arguments={},
+      cancel_signal=cancel_signal,
+  )
+
+  async_result = await client.call_tool_async(
+      tool_use_id="async-call",
+      name="long_running_tool",
+      arguments={},
+      cancel_signal=cancel_signal,
+  )
+  ```
+
+  The event is not cleared by `MCPClient`; clear it before reusing it for another call. Cancelling the
+  outer asyncio task is separate from setting `cancel_signal` and raises `asyncio.CancelledError`.
 
   Cancellation differs from [interrupts](../interrupts.mdx) in that it stops the agent entirely rather than pausing for human input. Interrupts allow the agent to resume from where it left off; cancellation does not.
   </Tab>
--- a/site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
+++ b/site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
@@ -151,66 +151,6 @@ The `agent.cancel()` method provides a way to stop the loop from outside, such a
       print("Agent was cancelled due to timeout")
   ```
 
-  #### Per-call MCP cancellation
-
-  `MCPClient.call_tool_sync()` and `MCPClient.call_tool_async()` also accept a caller-owned
-  `threading.Event` through the keyword-only `cancel_signal` parameter. Setting the event cancels
-  only that call; the MCP session remains available for later calls. A locally observed cancellation
-  returns an error result with `cancelled=True`. This marker does not guarantee that remote execution
-  stopped, because protocol cancellation is best-effort.
-
-  Synchronous call:
-
-  ```python
-  import threading
-
-  from strands.tools.mcp import MCPClient
-
-
-  def call_tool(client: MCPClient) -> None:
-      """Cancel a tool after one second while the client context is active."""
-      cancel_signal = threading.Event()
-      timer = threading.Timer(1, cancel_signal.set)
-      timer.start()
-      try:
-          result = client.call_tool_sync(
-              tool_use_id="sync-call",
-              name="long_running_tool",
-              arguments={},
-              cancel_signal=cancel_signal,
-          )
-      finally:
-          timer.cancel()
-  ```
-
-  Asynchronous call:
-
-  ```python
-  import asyncio
-  import threading
-
-  from strands.tools.mcp import MCPClient
-
-
-  async def call_tool(client: MCPClient) -> None:
-      """Cancel a tool after one second while the client context is active."""
-      cancel_signal = threading.Event()
-      loop = asyncio.get_running_loop()
-      timer = loop.call_later(1, cancel_signal.set)
-      try:
-          result = await client.call_tool_async(
-              tool_use_id="async-call",
-              name="long_running_tool",
-              arguments={},
-              cancel_signal=cancel_signal,
-          )
-      finally:
-          timer.cancel()
-  ```
-
-  The event is not cleared by `MCPClient`; clear it before reusing it for another call. Cancelling the
-  outer asyncio task is separate from setting `cancel_signal` and raises `asyncio.CancelledError`.
-
   Cancellation differs from [interrupts](../interrupts.mdx) in that it stops the agent entirely rather than pausing for human input. Interrupts allow the agent to resume from where it left off; cancellation does not.
   </Tab>
   <Tab label="TypeScript">
--- a/site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
+++ b/site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
@@ -158,28 +158,42 @@ The `agent.cancel()` method provides a way to stop the loop from outside, such a
   returns an error result with `cancelled=True`. This marker does not guarantee that remote execution
   stopped, because protocol cancellation is best-effort.
 
+  Synchronous call:
+
   ```python
-  import asyncio
   import threading
 
   from strands.tools.mcp import MCPClient
 
-  cancel_signal = threading.Event()
 
-  # Inside an active MCPClient context:
-  sync_result = client.call_tool_sync(
-      tool_use_id="sync-call",
-      name="long_running_tool",
-      arguments={},
-      cancel_signal=cancel_signal,
-  )
+  def call_tool(client: MCPClient) -> None:
+      """Call a tool while the client context is active."""
+      cancel_signal = threading.Event()
+      result = client.call_tool_sync(
+          tool_use_id="sync-call",
+          name="long_running_tool",
+          arguments={},
+          cancel_signal=cancel_signal,
+      )
+  ```
 
-  async_result = await client.call_tool_async(
-      tool_use_id="async-call",
-      name="long_running_tool",
-      arguments={},
-      cancel_signal=cancel_signal,
-  )
+  Asynchronous call:
+
+  ```python
+  import threading
+
+  from strands.tools.mcp import MCPClient
+
+
+  async def call_tool(client: MCPClient) -> None:
+      """Call a tool while the client context is active."""
+      cancel_signal = threading.Event()
+      result = await client.call_tool_async(
+          tool_use_id="async-call",
+          name="long_running_tool",
+          arguments={},
+          cancel_signal=cancel_signal,
+      )
   ```
 
   The event is not cleared by `MCPClient`; clear it before reusing it for another call. Cancelling the
--- a/site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
+++ b/site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
@@ -114,13 +114,14 @@ The `agent.cancel()` method provides a way to stop the loop from outside, such a
 <Tabs>
   <Tab label="Python">
 
-  The agent checks for cancellation at three checkpoints:
+  The agent checks for cancellation at four checkpoints:
 
   | Checkpoint | Behavior | Note |
   |---|---|---|
   | Model response streaming | Partial output is discarded | Usage metrics may be inaccurate since the stream is closed before the model sends its final metadata event |
   | Before tool execution | Tool calls are skipped with error results added to maintain valid conversation state | |
   | During MCP tool execution | The in-flight MCP request is cancelled without closing the shared MCP session | Remote cancellation is best-effort; the agent stops locally even if the server does not acknowledge cancellation |
+  | After tool execution | The agent stops before the next model call | Non-MCP tools finish before cancellation is observed |
 
   The agent returns a result with `stop_reason="cancelled"`. `cancel()` is thread-safe and idempotent. Calling it multiple times or from different threads is safe.
 
@@ -167,33 +168,44 @@ The `agent.cancel()` method provides a way to stop the loop from outside, such a
 
 
   def call_tool(client: MCPClient) -> None:
-      """Call a tool while the client context is active."""
+      """Cancel a tool after one second while the client context is active."""
       cancel_signal = threading.Event()
-      result = client.call_tool_sync(
-          tool_use_id="sync-call",
-          name="long_running_tool",
-          arguments={},
-          cancel_signal=cancel_signal,
-      )
+      timer = threading.Timer(1, cancel_signal.set)
+      timer.start()
+      try:
+          result = client.call_tool_sync(
+              tool_use_id="sync-call",
+              name="long_running_tool",
+              arguments={},
+              cancel_signal=cancel_signal,
+          )
+      finally:
+          timer.cancel()
   ```
 
   Asynchronous call:
 
   ```python
+  import asyncio
   import threading
 
   from strands.tools.mcp import MCPClient
 
 
   async def call_tool(client: MCPClient) -> None:
-      """Call a tool while the client context is active."""
+      """Cancel a tool after one second while the client context is active."""
       cancel_signal = threading.Event()
-      result = await client.call_tool_async(
-          tool_use_id="async-call",
-          name="long_running_tool",
-          arguments={},
-          cancel_signal=cancel_signal,
-      )
+      loop = asyncio.get_running_loop()
+      timer = loop.call_later(1, cancel_signal.set)
+      try:
+          result = await client.call_tool_async(
+              tool_use_id="async-call",
+              name="long_running_tool",
+              arguments={},
+              cancel_signal=cancel_signal,
+          )
+      finally:
+          timer.cancel()
   ```
 
   The event is not cleared by `MCPClient`; clear it before reusing it for another call. Cancelling the
--- a/strands-py/AGENTS.md
+++ b/strands-py/AGENTS.md
@@ -206,7 +206,7 @@ class XModel(Model):
 - **Never buffer a stream into a list before yielding.** Consume upstream async iterables with `async for` and yield as you go.
 - **Async-first with thin sync facades.** Public APIs come in async form (`invoke_async`, `stream_async`) plus thin sync wrappers (`__call__`, `invoke`) that delegate via `_async.run_async`. Never block the event loop — wrap blocking or third-party-sync calls in `asyncio.to_thread` (see `bedrock.py`).
 - **The floor is Python 3.10.** Do not use `asyncio.TaskGroup` or `asyncio.timeout` (3.11+) without a `sys.version_info` gate and a 3.10 fallback (see `multiagent/swarm.py`); for concurrent work use `asyncio.create_task` and clean up in `finally` with `task.cancel()` then `await asyncio.gather(*tasks, return_exceptions=True)`.
-- **Cancellation is cooperative.** `agent.cancel()` sets an internal `threading.Event` checked at model/tool boundaries and forwarded into in-flight MCP calls (see `agent/_concurrency.py` and `tools/mcp/mcp_agent_tool.py`). Direct MCP calls also accept a caller-owned `threading.Event` through `cancel_signal`; remote cancellation is best-effort. *(The TypeScript SDK uses `AbortSignal` instead — see its AGENTS.md.)*
+- **Cancellation is cooperative.** Long-running integrations must propagate and observe the agent cancellation signal rather than blocking the event loop. *(The TypeScript SDK uses `AbortSignal` instead — see its AGENTS.md.)*
 
 ## MCP Tasks (Experimental)
 
--- a/strands-py/AGENTS.md
+++ b/strands-py/AGENTS.md
@@ -206,7 +206,7 @@ class XModel(Model):
 - **Never buffer a stream into a list before yielding.** Consume upstream async iterables with `async for` and yield as you go.
 - **Async-first with thin sync facades.** Public APIs come in async form (`invoke_async`, `stream_async`) plus thin sync wrappers (`__call__`, `invoke`) that delegate via `_async.run_async`. Never block the event loop — wrap blocking or third-party-sync calls in `asyncio.to_thread` (see `bedrock.py`).
 - **The floor is Python 3.10.** Do not use `asyncio.TaskGroup` or `asyncio.timeout` (3.11+) without a `sys.version_info` gate and a 3.10 fallback (see `multiagent/swarm.py`); for concurrent work use `asyncio.create_task` and clean up in `finally` with `task.cancel()` then `await asyncio.gather(*tasks, return_exceptions=True)`.
-- **Cancellation is cooperative.** It uses an internal `threading.Event` set by `agent.cancel()` and checked at turn/tool boundaries (see `agent/_concurrency.py`); the SDK does not accept an external cancel token. *(The TypeScript SDK uses `AbortSignal` instead — see its AGENTS.md.)*
+- **Cancellation is cooperative.** `agent.cancel()` sets an internal `threading.Event` checked at model/tool boundaries and forwarded into in-flight MCP calls (see `agent/_concurrency.py` and `tools/mcp/mcp_agent_tool.py`). Direct MCP calls also accept a caller-owned `threading.Event` through `cancel_signal`; remote cancellation is best-effort. *(The TypeScript SDK uses `AbortSignal` instead — see its AGENTS.md.)*
 
 ## MCP Tasks (Experimental)
 
--- a/strands-py/src/strands/event_loop/event_loop.py
+++ b/strands-py/src/strands/event_loop/event_loop.py
@@ -559,8 +559,7 @@ async def _handle_model_execution(
 
             if last_event is None:
                 raise RuntimeError(
-                    "Middleware chain did not yield a result event. "
-                    "Ensure middleware forwards events from next()."
+                    "Middleware chain did not yield a result event. Ensure middleware forwards events from next()."
                 )
 
             # Write the post-stream model state back to the agent. Skipped on error
@@ -849,9 +848,18 @@ async def _handle_tool_execution(
         )
         return
 
+    if agent._cancel_signal.is_set():
+        yield EventLoopStopEvent(
+            "cancelled",
+            message,
+            agent.event_loop_metrics,
+            invocation_state["request_state"],
+        )
+        return
+
     # Emit after_tools checkpoint. Only fires on tool_use cycles: a model that
     # returns end_turn first never reaches this branch.
-    if agent._checkpointing and not agent._cancel_signal.is_set():
+    if agent._checkpointing:
         cycle_index = agent._checkpoint_cycle_index
         agent._checkpoint_cycle_index = cycle_index + 1
         yield _build_checkpoint_stop_event(
@@ -860,17 +868,6 @@ async def _handle_tool_execution(
             cycle_index=cycle_index,
             message=message,
             request_state=invocation_state["request_state"],
-        )
-        return
-
-    # If checkpointing is on and cancel suppressed the checkpoint above, emit
-    # "cancelled" now to avoid an extra model call.
-    if agent._checkpointing and agent._cancel_signal.is_set():
-        yield EventLoopStopEvent(
-            "cancelled",
-            message,
-            agent.event_loop_metrics,
-            invocation_state["request_state"],
         )
         return
 
--- a/strands-py/src/strands/tools/mcp/mcp_client.py
+++ b/strands-py/src/strands/tools/mcp/mcp_client.py
@@ -1338,7 +1338,7 @@ class MCPClient(ToolProvider):
                 else:
                     pending = {invoke_event} if not invoke_event.done() else set()
                 if pending:
-                    invoke_event.add_done_callback(lambda task: task.exception() if not task.cancelled() else None)
+                    self._track_background_cleanup_task(invoke_event)
                 if cancel_event is not None and not cancel_event.done():
                     cancel_event.cancel()
                     await asyncio.gather(cancel_event, return_exceptions=True)
--- a/strands-py/src/strands/tools/mcp/mcp_client.py
+++ b/strands-py/src/strands/tools/mcp/mcp_client.py
@@ -1188,7 +1188,11 @@ class MCPClient(ToolProvider):
                 )
             else:
                 return
-            await asyncio.wait_for(cancellation, timeout=1)
+            cancellation_task = asyncio.create_task(cancellation)
+            _, pending = await asyncio.wait({cancellation_task}, timeout=1)
+            if pending:
+                cancellation_task.cancel()
+                cancellation_task.add_done_callback(lambda task: task.exception() if not task.cancelled() else None)
         except Exception as error:
             self._log_debug_with_thread("error=<%s> | failed to notify MCP server of cancellation", str(error))
 
@@ -1214,6 +1218,7 @@ class MCPClient(ToolProvider):
                 raise RuntimeError("Tool execution cancelled")
 
             invoke_event = asyncio.create_task(coro)
+            invoke_cancel_requested = False
             cancel_event: asyncio.Task[None] | None = None
 
             if cancel_signal is not None:
@@ -1233,6 +1238,7 @@ class MCPClient(ToolProvider):
 
                 if cancel_signal is not None and cancel_signal.is_set():
                     self._log_debug_with_thread("cancellation detected during MCP invocation")
+                    invoke_cancel_requested = True
                     invoke_event.cancel()
                     _, pending = await asyncio.wait({invoke_event}, timeout=1)
                     if pending:
@@ -1248,13 +1254,15 @@ class MCPClient(ToolProvider):
                 await asyncio.wait({invoke_event}, timeout=1)
                 raise RuntimeError("Connection to the MCP server was closed")
             finally:
-                if not invoke_event.done():
+                if not invoke_event.done() and not invoke_cancel_requested:
                     invoke_event.cancel()
                     _, pending = await asyncio.wait({invoke_event}, timeout=1)
-                    if pending:
-                        invoke_event.add_done_callback(lambda task: task.exception() if not task.cancelled() else None)
                     if cancellation_state is not None:
                         await self._cancel_tool_call(cancellation_state)
+                else:
+                    pending = {invoke_event} if not invoke_event.done() else set()
+                if pending:
+                    invoke_event.add_done_callback(lambda task: task.exception() if not task.cancelled() else None)
                 if cancel_event is not None and not cancel_event.done():
                     cancel_event.cancel()
                     await asyncio.gather(cancel_event, return_exceptions=True)
@@ -1448,13 +1456,25 @@ class MCPClient(ToolProvider):
         try:
             create_result = await asyncio.shield(create_task)
         except asyncio.CancelledError:
-            try:
-                create_result = await asyncio.wait_for(asyncio.shield(create_task), timeout=1)
+            done, _ = await asyncio.wait({create_task}, timeout=1)
+            if done:
+                create_result = create_task.result()
                 if cancellation_state is not None:
                     cancellation_state["task_id"] = create_result.task.taskId
-            except asyncio.TimeoutError:
-                create_task.cancel()
-                await asyncio.gather(create_task, return_exceptions=True)
+            else:
+
+                def cancel_delayed_task(task: asyncio.Task[Any]) -> None:
+                    if task.cancelled():
+                        return
+                    try:
+                        result = task.result()
+                    except Exception:
+                        return
+                    if cancellation_state is not None:
+                        cancellation_state["task_id"] = result.task.taskId
+                        asyncio.create_task(self._cancel_tool_call(cancellation_state))
+
+                create_task.add_done_callback(cancel_delayed_task)
             raise
         task_id = create_result.task.taskId
         if cancellation_state is not None:
--- a/strands-py/src/strands/tools/mcp/mcp_client.py
+++ b/strands-py/src/strands/tools/mcp/mcp_client.py
@@ -1298,8 +1298,9 @@ class MCPClient(ToolProvider):
             if cancel_signal is not None:
 
                 async def wait_for_cancel() -> None:
-                    # Polling avoids blocking the loop. Running Event.wait() in an executor would
-                    # leave its worker thread blocked when this watcher is cancelled after success.
+                    # threading.Event has no async notification hook. Poll on this loop rather than
+                    # running Event.wait() in an executor: cancelling that await cannot stop a worker
+                    # already blocked in Event.wait(), so successful calls could strand worker threads.
                     while not cancel_signal.is_set():
                         await asyncio.sleep(0.05)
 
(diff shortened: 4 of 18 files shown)


arielnabavian:
# PR #3402 / Issue #3276 — Per-Call MCP Cancellation: Manual Test Results

**Before:** `origin/main` (`33f5f741`) · **After:** PR #3402 (`f0e09040`)
**Env:** Python 3.12 · real MCP servers (stdio/HTTP/SSE) · real Bedrock `global.anthropic.claude-sonnet-4-6` (acct 159773341612, role `Bedrock-Access`)

| # | Test case (manually run) | How | Before (origin/main) | After (PR #3402) |
|---|---|---|---|---|
| 1 | `call_tool_sync` exposes `cancel_signal` param | `inspect.signature` | ❌ absent → `TypeError` | ✅ present |
| 2 | Direct `call_tool_sync`, signal set mid-flight | real stdio server, 6s tool, cancel @1s | ⛔ n/a (no API) | ✅ returns 1.02s, `cancelled=True` |
| 3 | Session reusable after per-call cancel | `add(2,3)` after cancel | ⛔ n/a | ✅ `→5`, `status=success` |
| 4 | Pre-set signal before call starts | signal `.set()` before call | ⛔ n/a | ✅ short-circuits ~0.0s, `cancelled=True` |
| 5 | `agent.cancel()` reaches in-flight MCP call (via `MCPAgentTool.stream`) | real tool 6s, cancel @1s | ❌ ran full 6.09s, `cancelled=None` | ✅ stopped 1.02s, `cancelled=True` |
| 6 | Control — no signal, tool completes | 2s tool, no cancel | ✅ 2.04s, `success`, no `cancelled` | ✅ 2.04s, `success`, no `cancelled` |
| 7 | Full agent loop → `stop_reason` (stub model) | real MCP tool, `agent.cancel()` @1.5s | (not applicable pre-feature) | ✅ `stop_reason='cancelled'`, 1.53s |
| 8 | Remote execution semantics | server stderr progress after cancel | — | ⚠️ server ran to 10.00s — local cancel only (as documented) |
| 9 | **Model-driven happy path** — Bedrock calls MCP `calculator(17,25)` | real `agent("add 17 and 25")` | ✅ `end_turn`, answer=**42** | ✅ `end_turn`, answer=**42** |
| 10 | **Model-driven cancel** — Bedrock calls MCP `slow_tool(15s)`, stop @2s | real `agent(...)` + `agent.cancel()` | ⚠️ `cancelled` but **19.18s** (`stopped_early=false`) | ✅ `cancelled`, **2.01s** (`stopped_early=true`) |
| 11 | Model-driven cancel (standalone repro) | real `agent(...)`, stop @2s | (not applicable pre-feature) | ✅ `stop_reason='cancelled'`, 2.04s |

## Documented MCP config mechanisms (After branch, real servers — unchanged by PR, not run before/after)

| # | Mechanism | Result |
|---|---|---|
| 1 | Managed integration `Agent(tools=[mcp_client])` | ✅ PASS |
| 2 | Manual context `with client: list_tools_sync()` | ✅ PASS |
| 3a | Transport: stdio | ✅ PASS (`1+2=3`) |
| 3b | Transport: Streamable HTTP | ✅ PASS (`4+5=9`) |
| 3c | Transport: SSE | ✅ PASS (`7+8=15`) |
| 4 | Multiple MCP servers | ✅ PASS (12 tools / 2 servers) |
| 5a | `tool_filters` allowed — string | ✅ PASS |
| 5b | `tool_filters` allowed — regex | ✅ PASS |
| 5c | Combined allowed → rejected | ✅ PASS |
| 6 | `prefix` tool naming | ✅ PASS |
| 7 | Direct invocation `call_tool_sync` | ✅ PASS (`10+20=30`) |
| 8 | Elicitation `elicitation_callback` | ✅ PASS |
| 9a | Progress — instance callback | ✅ PASS (3 updates) |
| 9b | Progress — per-call override | ✅ PASS |
| 10 | Use outside `with` raises `MCPClientInitializationError` | ✅ PASS |

## Automated

| Suite | Result |
|---|---|
| PR unit tests (`tests/strands/tools/mcp` + `test_event_loop.py`) | ✅ 242 passed |

## Notes

- **Core fix (rows 5, 10):** before, `agent.cancel()` during an in-flight MCP call did nothing — the tool ran to completion; after, it stops in ~1–2s. Closes #3276.
- **No regression:** happy path (rows 6, 9) identical before/after.
- **Local vs. remote (row 8):** `cancelled=True` confirms *local* cancellation, not that the server stopped — matches the PR's documented contract.
- **Not tested:** AWS IAM SigV4 transport (`mcp-proxy-for-aws`) — needs a live AWS-hosted MCP endpoint; transport shim over the same `MCPClient`, unrelated to this PR.

strands-py/src/strands/tools/mcp/mcp_client.py
```diff
@@ -767,15 +783,23 @@ async def _call_as_task() -> MCPCallToolResult:
...
             return _call_as_task()
         else:
             self._log_debug_with_thread("tool=<%s> | using direct call_tool", name)
 
             async def _call_tool_direct() -> MCPCallToolResult:
-                return await cast(ClientSession, self._background_thread_session).call_tool(
+                session = cast(ClientSession, self._background_thread_session)
+                if cancellation_state is not None:
+                    cancellation_state["session"] = session
+                    cancellation_state["request_id"] = session._request_id
```
agent-of-mkmeral: I verified this is race-free on `mcp==1.28.1` (no await between this capture and the synchronous read+increment at the top of `BaseSession.send_request`, and `call_tool` doesn't await before calling it). Two hardening asks though:

1. `_request_id` is an undocumented private of the `mcp` package; the pin is `>=1.23.0,<2.0.0`, so a minor release could rename it or (worse) insert an await before the increment, silently breaking the "cancel the *exact* request" guarantee into "cancel someone else's request". Please add a comment documenting the invariant this relies on (capture → increment must be one synchronous segment), and consider `getattr(session, "_request_id", None)` so a rename degrades to "no request-scoped cancel" (the local cancel still works) instead of failing the whole call.
2. A unit test that pins the invariant would catch an SDK bump that breaks it — e.g. assert the captured id equals the id the mock session actually used when two calls run concurrently (you have the isolation test; an explicit id-match assertion under an mcp upgrade would fail loudly).

strands-py/src/strands/tools/mcp/mcp_client.py
```diff
@@ -1139,20 +1236,59 @@ def _invoke_on_background_thread(self, coro: Coroutine[Any, Any, T]) -> futures.
...
-            ]
+            invoke_cancel_requested = False
+            cancel_event: asyncio.Task[None] | None = None
+
+            if cancel_signal is not None:
 
-            done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
+                async def wait_for_cancel() -> None:
+                    while not cancel_signal.is_set():
+                        await asyncio.sleep(0.05)
```
agent-of-mkmeral, line 1305: Nit: this spawns a 20 Hz polling task per in-flight call for its entire duration and adds up to 50 ms of cancellation latency. It's acceptable (TS gets event-driven `AbortSignal` for free; Python doesn't), but consider `loop.run_in_executor(None, cancel_signal.wait)` wrapped in a task — event-driven, no polling, and the executor thread is released when you cancel it via the existing cleanup... actually `Event.wait` in an executor thread can't be interrupted, so it would leak a thread until the signal is set. Given that, the polling approach is the pragmatic choice — suggest just adding a brief comment explaining why polling was chosen so the next reader doesn't "optimize" it into a thread leak.

strands-py/src/strands/tools/mcp/mcp_client.py
```diff
@@ -1328,13 +1466,43 @@ async def _call_tool_as_task_and_poll_async(
...
+                def cancel_delayed_task(task: asyncio.Task[Any]) -> None:
+                    if task.cancelled():
+                        return
+                    try:
+                        result = task.result()
+                    except Exception:
+                        return
+                    if cancellation_state is not None:
+                        cancellation_state["task_id"] = result.task.taskId
+                        asyncio.create_task(self._cancel_tool_call(cancellation_state))
```
agent-of-mkmeral: Fire-and-forget `asyncio.create_task(...)` without keeping a reference — per the asyncio docs, the loop only holds a weak reference, so this cleanup task can be garbage-collected mid-execution and the `tasks/cancel` never reaches the server (precisely in the delayed-task-creation case this callback exists to handle). Store the task on the client (e.g. a `set` with `add_done_callback(discard)`) or hold it in `cancellation_state`.

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -849,9 +848,18 @@ async def _handle_tool_execution(
         )
         return
 
+    if agent._cancel_signal.is_set():
```
agent-of-mkmeral, line 851: This check runs after `ToolExecutor._execute` has fully drained — which is too late to prevent the `AfterToolCallEvent.retry=True` interaction: a retry-on-error hook sees the cancelled result (`status="error"`) inside the executor and re-invokes the tool via `continue` before control ever returns here. Reproduced: 6 tool invocations after `agent.cancel()` with a max-5-retries hook. The stop itself is correct; the guard belongs in `tools/executors/_executor.py`'s retry branch (skip retry when the result has `cancelled=True` or the agent signal is set). This is the blocker your own staged aggregation flagged — please fix it in this PR rather than deferring, since this PR is what makes cancelled results reachable by retry hooks.

site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
```diff
@@ -149,6 +150,41 @@ The `agent.cancel()` method provides a way to stop the loop from outside, such a
...
+
+  async_result = await client.call_tool_async(
+      tool_use_id="async-call",
+      name="long_running_tool",
+      arguments={},
+      cancel_signal=cancel_signal,
+  )
+  ```
+
+  The event is not cleared by `MCPClient`; clear it before reusing it for another call. Cancelling the
```
agent-of-mkmeral: The example above this note contradicts it: one `cancel_signal` is shared across the sync and async calls with no `.clear()` between them — exactly the reuse pattern this line warns about. Also the snippet mixes a bare sync call and a bare `await` in one block, so it isn't runnable as written. Suggest either two separate signals, or one signal with an explicit `cancel_signal.clear()` between the calls, and splitting the sync/async examples so each is self-consistent.

agent-of-mkmeral (changes requested):
Full review at `f0e0904`: cloned the branch, read the 10-file diff, ran the affected suites (242 passed locally), verified the request-ID capture against `mcp==1.28.1` internals, compared against the TS implementation, and wrote an adversarial repro for the blocker your own staged review disclosed.

## Verdict: Request changes — the disclosed retry blocker is real and reproducible

The PR body honestly discloses: *"De-duplicated blocker: `AfterToolCallEvent.retry=True` can retry a cancellation result before the event loop reaches its post-tool cancellation stop... Current staged verdict: changes requested. The branch has not been modified after this aggregation yet."* I independently confirmed it:

```
# Hook that sets event.retry=True on error results (common flaky-tool pattern),
# agent.cancel() fired mid-tool-call:
stop_reason: cancelled
tool invocations: 6 | retries used: 5
```

A retry-on-error hook re-invoked the tool **5 more times after cancellation** because (a) the cancelled result has `status="error"` and is indistinguishable from a genuine failure at the hook layer, and (b) `ToolExecutor._stream_with_trace` honors `after_event.retry` via `continue` before the event loop's post-tool cancel check (`event_loop.py:851`) ever runs. With per-call MCP cancellation each retry at least short-circuits fast on the still-set signal, but an unbounded retry hook spins forever and every retry delays the cancellation the user explicitly requested.

**Suggested fix (small):** in `tools/executors/_executor.py`, skip the retry branch when the result carries `cancelled=True` or when `agent._cancel_signal.is_set()` — cancellation should not be retryable. A regression test mirroring my repro would lock this in. (I can't attach an inline comment there since the file isn't in this diff.)

## What I verified and liked

- **`session._request_id` capture is race-free on the current pinned range.** I checked `mcp==1.28.1`: `ClientSession.call_tool` has no await before `BaseSession.send_request`, and `send_request` reads+increments `_request_id` in the same synchronous segment — so the capture in `_call_tool_direct` cannot interleave with another request on the single background loop. Good analysis in the PR; see inline nit about hardening this private-attribute dependency.
- **Conversation state stays valid on cancel:** the tool-result message is appended before the new `cancelled` stop at `event_loop.py:851`, so `toolUse` is always paired with `toolResult`. The reordering also removes the pre-existing extra model turn for non-checkpointing agents — nice fix.
- **Faithful port with justified divergence.** TS just forwards `AbortSignal` to an SDK that natively supports it (`mcp-tool.ts:52`); the Python MCP SDK has no cancel token, so the hand-rolled watcher/notification machinery is warranted. The `cancelled=True` result marker and cancel-wins-over-ready-result semantics are Python-only extensions, both documented.
- **Test quality is genuinely good:** per-call isolation, cancel/result race, delayed task creation, cancellation-resistant task creation, bounded remote cleanup, session reuse after cancel — all covered. 242 passed locally for me too.
- The independent manual E2E table in the PR comments (real servers + real Bedrock) corroborates the core fix.

## Should-fix (all previously disclosed by your own aggregation, all still unaddressed)

1. **Retry blocker** — above.
2. **Model-facing text**: `"Tool execution failed: Tool execution cancelled"` doesn't state that remote execution may have continued (or even completed — in the cancel-wins race the server *did* finish the work). For direct-call users who feed this result to a model, something like `"Tool execution cancelled locally; remote execution may have continued"` is more accurate.
3. **Pre-set-signal regression test**: covered only by the manual test table (row 4), not by a unit test.
4. **Stale `strands-py/AGENTS.md:209`**: *"checked at turn/tool boundaries... the SDK does not accept an external cancel token"* — both halves are now false: cancellation reaches in-flight MCP calls, and `cancel_signal` is exactly an external cancel token.

See inline comments for smaller items. Once the retry interaction is fixed (+ regression test) and the disclosed should-fixes are addressed, this looks mergeable — the core mechanism is solid and well-tested.

arielnabavian:
@strandly-the-agent address comments and revise

strandly-the-agent:
Revised at `94a710f6`; all review comments are addressed.

- cancellation is terminal even when hooks request retry
- remaining sequential tools are skipped with valid cancellation results
- direct/task cleanup is strongly owned and bounded through session shutdown
- request-ID handling degrades safely and is tested with a later nonzero request
- result text, pre-set signal coverage, polling rationale, AGENTS guidance, and runnable docs are corrected

Validation: **4,506 tests passed**; Ruff + mypy passed; `git diff --check` passed. CI is now running on the revised head.

<details><summary>Independent review ledger</summary>

Three fresh rounds found 5 → 3 → 1 issues; every finding was fixed and regression-tested. The three-round cap is reached, so the final round-3 lifecycle fix is locally verified but not self-approved.

</details>

arielnabavian:
# PR #3402 — Manual test cases (revalidated)

**PR:** feat(python): add per-call MCP tool cancellation — https://github.com/strands-agents/harness-sdk/pull/3402
**Validated at head:** `94a710f6` ("fix(python): retain resistant MCP invocations")
**Baseline ("Before"):** `origin/main` @ `81bccd84`

Legend: ✅ pass/expected · ❌ bug/absent · ⛔ n/a (API doesn't exist pre-feature) · ⚠️ documented limitation · — not applicable

---

## Part A — MCP cancellation (original matrix)

| # | Test case (manually run) | How | Before (origin/main) | After (PR #3402) |
|---|---|---|---|---|
| 1 | `call_tool_sync` exposes `cancel_signal` param | `inspect.signature` | ❌ absent → `TypeError` | ✅ present |
| 2 | Direct `call_tool_sync`, signal set mid-flight | real stdio server, 6s tool, cancel @1s | ⛔ n/a (no API) | ✅ returns 1.02s, `cancelled=True` |
| 3 | Session reusable after per-call cancel | `add(2,3)` after cancel | ⛔ n/a | ✅ →5, `status=success` |
| 4 | Pre-set signal before call starts | signal `.set()` before call | ⛔ n/a | ✅ short-circuits ~0.0s, `cancelled=True` |
| 5 | `agent.cancel()` reaches in-flight MCP call (via `MCPAgentTool.stream`) | real tool 6s, cancel @1s | ❌ ran full 6.09s, `cancelled=None` | ✅ stopped 1.02s, `cancelled=True` |
| 6 | Control — no signal, tool completes | 2s tool, no cancel | ✅ 2.04s, success, no `cancelled` | ✅ 2.04s, success, no `cancelled` |
| 7 | Full agent loop → `stop_reason` (stub model) | real MCP tool, `agent.cancel()` @1.5s | (not applicable pre-feature) | ✅ `stop_reason='cancelled'`, 1.53s |
| 8 | Remote execution semantics | server stderr progress after cancel | — | ⚠️ server ran to 10.00s — local cancel only (as documented) |
| 9 | Model-driven happy path — Bedrock calls MCP `calculator(17,25)` | real `agent("add 17 and 25")` | ✅ end_turn, answer=42 | ✅ end_turn, answer=42 |
| 10 | Model-driven cancel — Bedrock calls MCP `slow_tool(15s)`, stop @2s | real `agent(...)` + `agent.cancel()` | ⚠️ cancelled but 19.18s (`stopped_early=false`) | ✅ cancelled, 2.01s (`stopped_early=true`) |
| 11 | Model-driven cancel (standalone repro) | real `agent(...)`, stop @2s | (not applicable pre-feature) | ✅ `stop_reason='cancelled'`, 2.04s |

---

## Part B — Retry ✕ cancellation (NEW)

**What the PR added here:** on `origin/main` the tool retry loop in `ToolExecutor._stream`
is *unconditional* — any hook that sets `AfterToolCallEvent.retry=True` re-runs the tool,
even when the agent was cancelled or the result is `cancelled=True`. A retry hook + a
cancelled tool therefore **loops forever**, defeating `agent.cancel()`. The PR introduces
`ToolExecutor._should_retry(agent, after_event)` (`src/strands/tools/executors/_executor.py:95`)
and routes both retry sites (result path + exception path) through it, so cancellation is
terminal while ordinary (non-cancel) retries are untouched. `SequentialToolExecutor._execute`
also short-circuits remaining tools once `_cancel_signal` is set.

Exercised against the **real** `ToolExecutor` (retry is a pure in-process concern — no MCP
server or Bedrock needed). Repro: `.context/retry_cancel_repro.py`, same script run against
both versions (origin/main via `PYTHONPATH` shadow). A `SAFETY_CAP=8` was added so the
origin/main runs terminate instead of hanging — the buggy path would otherwise be unbounded.

| # | Test case (manually run) | How | Before (origin/main) | After (PR #3402) |
|---|---|---|---|---|
| R1 | `ToolExecutor._should_retry` gates retry | `hasattr(ToolExecutor, "_should_retry")` | ❌ absent (unconditional retry loop) | ✅ present |
| R2 | Retry hook + tool returns `cancelled=True` → terminal | retry-on-error hook; tool returns `{"status":"error","cancelled":True}` | ❌ retried to cap (`call_count=8`, would loop forever) | ✅ terminal, `call_count=1`, `status=error` (0.6 ms) |
| R3 | Retry hook + `agent.cancel()` set during tool (result path) → terminal | tool sets `_cancel_signal` then returns error; retry hook active | ❌ retried to cap (`call_count=8`) | ✅ terminal, `call_count=1`, `status=error` (0.5 ms) |
| R4 | Retry hook + agent cancelled, tool **raises** (exception path) → terminal | tool `stream` sets `_cancel_signal` then `raise RuntimeError`; retry hook active | ❌ retried to cap (`call_count=8`) | ✅ terminal, `call_count=1`, `status=error` (1.0 ms) |
| R5 | **Control** — retry hook + plain error, no cancel → still retries | flaky tool errors twice then succeeds; retry hook active; agent NOT cancelled | ✅ retries, `call_count=3`, success | ✅ retries, `call_count=3`, success (0.5 ms) |
| R6 | Sequential executor: cancel during tool 1 → tool 2 skipped, no retry | 2 tools, retry hook active, `_cancel_signal.set()` in tool 1's after-hook | ❌ **timed out at 10.00s** (runaway retry loop) | ✅ `t1_calls=1`, `t2_calls=0`, 2 results, tool 2 `status=error` "cancelled" (1.2 ms) |

**After column summary:** ALL PASS (R1–R6). **Before column:** R1–R4 + R6 FAIL (R5 control passes on both — proving the gate is surgical, not a blanket "disable retry").

---

## Regression tests backing Part B (all pass on PR head)

| Case | Committed test |
|---|---|
| R2, R3 | `tests/strands/tools/executors/test_executor.py::test_executor_stream_does_not_retry_cancelled_tool_result` (parametrized `cancel_agent=[False, True]` — covers both the `cancelled=True` result and the `_cancel_signal` paths) |
| R5 (control) | `test_executor_stream_retry_true`, `test_executor_stream_retry_after_exception` (retry still works when not cancelled) |
| R6 | `tests/strands/tools/executors/test_sequential.py::test_sequential_executor_cancellation_skips_remaining_tools` |
| loop precedence | `tests/strands/event_loop/test_event_loop.py::test_event_loop_cycle_cancel_after_tools_stops_without_checkpointing` (+ the three `cancel_*_beats_*_checkpoint` tests) |

> Note: R4's exception path is exercised end-to-end by the repro; the committed suite covers
> the exception retry branch generally (`test_executor_stream_retry_after_exception`) but does
> not have a dedicated *cancelled*-exception-path case. The shared `_should_retry` gate is what
> both paths call, so R2/R3's regression coverage protects R4's code path too.

---

## Full-suite revalidation at `94a710f6`

- `hatch test` → **4506 passed**, 76 warnings, ~53s
- MCP + executor suites (`test_mcp_client.py`, `test_mcp_client_tasks.py`, `test_mcp_agent_tool.py`, `executors/`) → **162 passed**
- Targeted cancellation-retry tests (7) → **7 passed**
- `ruff check` → all checks passed
- `mypy ./src` → success, no issues in 223 files

agent-of-mkmeral (approved):
## Re-review at `94a710f6` — APPROVE ✅

Follow-up to my [review at `f0e0904`](https://github.com/strands-agents/harness-sdk/pull/3402#pullrequestreview-4758864177) (REQUEST_CHANGES). Re-read the full delta (4 commits, 11 files, +397/−46), re-ran the suites, and **re-ran my original adversarial repro**.

- **Blocker fixed, confirmed by repro**: retry-after-cancel now terminal — **1 invocation vs 6** at `f0e0904`; ordinary retries preserved (control passed)
- **All 5 inline threads addressed**; bonus hardening beyond my asks
- **293 tests passed** locally; new code survived fresh adversarial checks
- **This supersedes my earlier CHANGES_REQUESTED. Ship it.** (2 cosmetic nits below, non-blocking)

<details>
<summary>Blocker: retry-after-cancel — FIXED, confirmed by repro</summary>

At `f0e0904` my repro showed a retry-on-error hook re-invoking a tool **6 times after `agent.cancel()`**. At `94a710f6` the same repro shows **exactly 1 invocation** on both paths:
- `cancelled=True` result + `retry=True` → terminal, 1 invocation
- `agent._cancel_signal` set + `retry=True` → terminal, 1 invocation
- **Control preserved:** plain-error retry without cancel still retries (3 attempts → success), so the `_should_retry` gate (`_executor.py:95`) is surgical, not a blanket retry disable. Both retry sites (result path and exception path) route through it, and the `_is_agent` guard keeps BidiAgent compatible.
- Regression-tested in the suite (`test_executor_stream_does_not_retry_cancelled_tool_result`, parametrized over both trigger paths).

</details>

<details>
<summary>All 5 inline threads from my review — addressed</summary>

1. **Model-facing text** now says `"Tool execution cancelled locally; remote execution may have continued"` — conveys the local-vs-remote semantics exactly as requested.
2. **Pre-set signal** now has a committed unit regression (`test_call_tool_sync_pre_set_cancel_signal_skips_request`), no longer manual-only.
3. **Stale `AGENTS.md:209`** corrected — now documents `cancel_signal` and MCP forwarding.
4. **Docs example** rewritten: separate `cancel_signal` per call, runnable sync/async snippets with proper timer cleanup — no longer contradicts the "clear before reuse" warning.
5. **Fire-and-forget GC risk** fixed properly: `_track_background_cleanup_task` keeps strong refs (asyncio only holds weak ones), and `_drain_background_cleanup_tasks` gives cleanup a bounded window at session shutdown. **Polling watcher** now has the rationale comment (executor-thread leak on `Event.wait()`).

</details>

<details>
<summary>Bonus hardening beyond my asks</summary>

- `session._request_id` capture now degrades safely (`getattr` + `isinstance(int)` — omits the remote notification rather than risking cancelling the wrong request if the mcp-sdk private attr changes under the `<2.0.0` pin). Tested with `test_call_tool_async_cancel_without_sdk_request_id_still_cancels_locally`.
- `SequentialToolExecutor` now skips remaining tools after cancel **while still appending valid error results per toolUse** — conversation state stays valid.
- New 4th cancellation checkpoint (after tools, before next model call) documented in the checkpoint table.

</details>

<details>
<summary>Adversarial checks on the NEW code (all passed)</summary>

- **Drain is bounded:** a cancellation-*resistant* cleanup task (ignores first cancel, sleeps 30s) — context exit still returns in **1.1s**, not hung.
- **Post-drain tracking is closed:** tasks enqueued after shutdown begins are auto-cancelled (delayed-callback race is sealed).

</details>

<details>
<summary>Validation</summary>

- MCP + executors + event_loop suites: **293 passed** locally at `94a710f6`
- Independent manual revalidation by @arielnabavian at this exact head (Part B retry×cancel matrix R1–R6, before/after) corroborates: before-column fails, after-column all pass, R5 control passes on both.

</details>

<details>
<summary>Non-blocking nits (fine to ignore)</summary>

- In the drain's first phase, a tracked task that completes *with an exception* is discarded without retrieval (potential "exception was never retrieved" warning) — theoretical, since `_cancel_tool_call` swallows its own exceptions.
- `sequential.py` cancel result uses `str(tool_use.get("toolUseId"))` → literal `"None"` if the id is missing; cosmetic.

</details>

site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
```diff
@@ -149,6 +151,66 @@ The `agent.cancel()` method provides a way to stop the loop from outside, such a
       print("Agent was cancelled due to timeout")
   ```
 
+  #### Per-call MCP cancellation
+
+  `MCPClient.call_tool_sync()` and `MCPClient.call_tool_async()` also accept a caller-owned
```
mkmeral: isn't this section essentially implementation detail? do we need it in our docs?

strands-py/src/strands/tools/mcp/mcp_client.py
```diff
@@ -1139,20 +1287,61 @@ def _invoke_on_background_thread(self, coro: Coroutine[Any, Any, T]) -> futures.
...
+            invoke_cancel_requested = False
+            cancel_event: asyncio.Task[None] | None = None
+
+            if cancel_signal is not None:
+
+                async def wait_for_cancel() -> None:
+                    # Polling avoids blocking the loop. Running Event.wait() in an executor would
+                    # leave its worker thread blocked when this watcher is cancelled after success.
+                    while not cancel_signal.is_set():
+                        await asyncio.sleep(0.05)
```
mkmeral, line 1305: isn't there a better way to do this?

strands-py/src/strands/tools/mcp/mcp_client.py
```diff
@@ -1139,20 +1287,61 @@ def _invoke_on_background_thread(self, coro: Coroutine[Any, Any, T]) -> futures.
...
                 self._log_debug_with_thread("event loop for the server closed before the invoke completed")
+                invoke_event.cancel()
+                await asyncio.wait({invoke_event}, timeout=1)
                 raise RuntimeError("Connection to the MCP server was closed")
-            else:
-                return await invoke_event
+            finally:
+                if not invoke_event.done() and not invoke_cancel_requested:
+                    invoke_event.cancel()
+                    _, pending = await asyncio.wait({invoke_event}, timeout=1)
```
mkmeral, line 1338: what happens on timeout do we get exception here? does it bubble up?

arielnabavian:
@strandly-the-agent make a revision to this pr. 

This logic we add to python 

def _should_retry(agent: "Agent | BidiAgent", after_event: AfterToolCallEvent | BidiAfterToolCallEvent) -> bool:
        """Return whether a hook-requested retry should run.

        Cancellation is terminal: retrying a locally cancelled tool can spin indefinitely
        while delaying the caller's requested agent stop.
        """
        if not getattr(after_event, "retry", False):
            return False
        if cast(dict[str, Any], after_event.result).get("cancelled") is True:
            return False
        return not (ToolExecutor._is_agent(agent) and agent._cancel_signal.is_set())
        
       
       
For cancelling retries is not in typescript. Lets include the equivalent typescript changes ONLY for this small functionality. Keep the rest of the typescript to python port the same.

strands-py/AGENTS.md
```diff
@@ -206,7 +206,7 @@ class XModel(Model):
...
 - **The floor is Python 3.10.** Do not use `asyncio.TaskGroup` or `asyncio.timeout` (3.11+) without a `sys.version_info` gate and a 3.10 fallback (see `multiagent/swarm.py`); for concurrent work use `asyncio.create_task` and clean up in `finally` with `task.cancel()` then `await asyncio.gather(*tasks, return_exceptions=True)`.
-- **Cancellation is cooperative.** It uses an internal `threading.Event` set by `agent.cancel()` and checked at turn/tool boundaries (see `agent/_concurrency.py`); the SDK does not accept an external cancel token. *(The TypeScript SDK uses `AbortSignal` instead — see its AGENTS.md.)*
+- **Cancellation is cooperative.** `agent.cancel()` sets an internal `threading.Event` checked at model/tool boundaries and forwarded into in-flight MCP calls (see `agent/_concurrency.py` and `tools/mcp/mcp_agent_tool.py`). Direct MCP calls also accept a caller-owned `threading.Event` through `cancel_signal`; remote cancellation is best-effort. *(The TypeScript SDK uses `AbortSignal` instead — see its AGENTS.md.)*
```
mkmeral: lol don't over-focus on the new features for docs updates, agents.md updates, etc.

mkmeral: (this is for strandly)

site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
arielnabavian: Agreed. We should remove this

arielnabavian:
@strandly-the-agent address comments from murat and revise

strandly-the-agent:
Revised at `57a5b6ff` with only the requested TypeScript retry parity.

- both `AfterToolCallEvent.retry` branches now require `!agent.cancelSignal.aborted`
- ordinary non-cancelled retries are unchanged
- added coverage for executed-tool and `BeforeToolCallEvent.cancel` paths
- no other TypeScript behavior or API changed

Validation: **3,808 TS tests passed**; build, type-check, lint, formatting, and diff checks passed. A fresh independent review returned **APPROVE** with no findings. CI is running on the new head.

strandly-the-agent:
Revised at `7d88f3f8` to address `mkmeral`'s comments.

- removed the direct per-call MCP section from the agent-loop guide
- generalized the `AGENTS.md` cancellation rule instead of documenting this feature there
- kept polling because `threading.Event` has no async notification hook; executor-based `Event.wait()` cannot be interrupted and would strand workers after successful calls
- clarified that `asyncio.wait(..., timeout=1)` returns the task in `pending`—it does **not** raise `TimeoutError`; the original caller cancellation/exception continues
- added a resistant-invocation regression proving `CancelledError` still bubbles while cleanup is retained and later drained

Validation: **4,507 Python tests** and **3,808 TypeScript tests** passed; Ruff, mypy, TS build/type-check/lint/format, and diff checks passed. Fresh independent review: **APPROVE**, no findings. CI is running on the revised head.

strands-py/src/strands/tools/mcp/mcp_client.py
arielnabavian: Good call-out. The constraint is that `cancel_signal` is a `threading.Event`, which has no async wakeup — so for the loop to notice it, something has to either poll it or block a thread on `.wait()`.

I tried the event-driven route (`loop.run_in_executor(None, cancel_signal.wait)`): it works, but `Event.wait()` can't be interrupted, so when the call completes normally the signal never fires and the executor thread blocks forever — a slow thread-pool leak. Adding a timeout to `.wait()` just relocates the polling into a scarce pool thread, which is strictly worse.

The genuinely better design is to make the *producer* wake the loop — have whatever calls `agent.cancel()` do a `loop.call_soon_threadsafe(...)` to cancel the invoke task directly (essentially what TS gets for free from `AbortSignal`). But that means threading the background loop + the specific task back to the cancel machinery, which is a bigger change than this PR. Within the current contract (opaque, caller-owned `threading.Event` with no loop affinity), polling is the only self-contained option that doesn't leak a thread. I'll file a follow-up issue for the event-driven bridge so we don't lose the idea.

strands-py/src/strands/tools/mcp/mcp_client.py
arielnabavian: No exception, and nothing bubbles up. `asyncio.wait(..., timeout=1)` (unlike `asyncio.wait_for`) returns normally on timeout as `(done, pending)` with the unfinished task in `pending` — it never raises `TimeoutError`. It also doesn't propagate the wrapped task's own exception; you'd have to `await` the task to surface that.

That's intentional here: this is `finally`-block cleanup, so we specifically don't want it to raise and mask the real exception being propagated (`_MCPCallCancelledError` or `RuntimeError`). If the invoke task is still pending after the 1s window, it goes to `_track_background_cleanup_task(invoke_event)`, which attaches a done-callback that later retrieves its exception — so nothing is lost and we never get an "exception was never retrieved" warning. The success path (`return await invoke_event`) does the `await`, so a real failure there surfaces normally.

strands-py/src/strands/event_loop/event_loop.py
arielnabavian: we have cancellation logic for retries upstream now

strands-py/src/strands/tools/mcp/mcp_client.py
arielnabavian: ✓ Good catch on the latency characteristic. To clarify the tradeoff:

The 20 Hz polling (50ms intervals via `asyncio.sleep(0.05)`) and ~50ms cancellation latency are acceptable for agent workflows because:

1. **Thread leak prevention**: `Event.wait()` in an executor blocks indefinitely and can't be interrupted. On successful calls, that thread strands permanently — a worse outcome than 50ms latency.

2. **Why not timeout on Event.wait()?** A timeout just moves the polling into a scarce pool thread, making it strictly worse.

3. **The 50ms worst-case is rare**: It only happens if the signal is set between polling cycles. Most real cancellations trigger instantly or within the next 50ms interval.

4. **Future optimization path**: The better design (producer-driven wakeup via `loop.call_soon_threadsafe()`) would require threading the background loop + task reference to the cancel machinery — a bigger change deferred to a follow-up issue.

The inline comment in `wait_for_cancel()` explains this for the next reader, so well-intentioned "optimizations" won't reintroduce the thread leak.

arielnabavian (approved):

agent-of-mkmeral (approved):
## Re-review at `7d88f3f8` — APPROVE ✅

Third pass, covering the 2 commits since my [approval at `94a710f6`](https://github.com/strands-agents/harness-sdk/pull/3402#pullrequestreview-4765651581): `57a5b6ff` (TS retry-after-cancel parity) and `7d88f3f8` (docs scope + cleanup clarifications). Delta is small (+125/−65 across 6 files), surgical, and fully regression-tested.

- **TS parity is correct and correctly asymmetric** — `_shouldRetry` gates only on `agent.cancelSignal.aborted`; I verified TS has no caller-owned per-call signal and no `cancelled` result field, so the Python-only result-marker check has no TS equivalent to port
- **Both mkmeral doc concerns addressed**: per-call MCP section removed from agent-loop.mdx (−60 lines), AGENTS.md rule generalized to one line
- **New resistant-invocation regression** pins the bounded-cleanup contract (CancelledError bubbles, pending work retained then drained)
- **Verified locally**: 390 Python tests (MCP+executors+event_loop) + 62 TS hook tests passed; ruff clean
- **Note**: head CI "FAILURE" is only the *2-Approvers-for-Bots* gate + *label-size* — no test failures. Still mergeable.

<details>
<summary>TS retry-after-cancel parity (57a5b6ff)</summary>

- `executor.ts` routes **both** retry sites (BeforeToolCallEvent-cancel path and executed-tool path) through the new `_shouldRetry`, mirroring Python's two-site routing.
- The asymmetry vs Python's `_should_retry` (which also checks `result.get("cancelled") is True`) is justified: in TS the only cancellation source is `agent.cancelSignal` (`mcp-tool.ts:52` forwards it; there is no per-call caller-owned signal), and `ToolResultBlock` has no `cancelled` field — a "cancelled result without aborted signal" state is unrepresentable in TS.
- No `_is_agent` guard needed: TS `ExecutorOptions.agent` is typed `Agent`, so `cancelSignal` always exists.
- Two new tests cover both paths and assert `stopReason='cancelled'`, `toolCallCount` 1 and 0 respectively, and single hook invocation; existing controls (`retries tool call when retry is set`, `does not retry when retry is not set`) still pass.

</details>

<details>
<summary>Python cleanup clarifications (7d88f3f8)</summary>

- **Docs**: the entire "Per-call MCP cancellation" section removed from `agent-loop.mdx` per mkmeral (implementation detail); this also moots the earlier example-vs-warning contradiction thread since the whole snippet is gone.
- **AGENTS.md**: feature-specific rule generalized ("long-running integrations must propagate and observe the agent cancellation signal") — matches the "don't over-focus on new features" ask.
- **Polling comment** now explains the real constraint (`threading.Event` has no async hook; executor `Event.wait()` can't be interrupted → stranded workers on success).
- **Bounded-cleanup comment** documents `asyncio.wait(..., timeout=1)` semantics (returns pending, never raises TimeoutError; caller's cancellation/exception continues) + a debug log when the 1s window expires.
- **New regression** `test_call_tool_async_caller_cancellation_timeout_preserves_cancelled_error`: cancellation-resistant invocation → caller still gets `CancelledError`, the pending task lands in `_background_cleanup_tasks`, and later drains to empty. Ran it in isolation: passes in 1.5s.

</details>

<details>
<summary>Validation at 7d88f3f8</summary>

- Python `tests/strands/tools/mcp` + `executors` + `event_loop`: **390 passed** (includes all retry×cancel regressions from my earlier reviews — gate in `_executor.py` untouched since `94a710f6`)
- TS `agent.hook.test.ts`: **62 passed** including the 2 new cancel-retry tests
- `ruff check` on touched files: clean
- Independent: @arielnabavian APPROVED at this exact head after his R1–R6 matrix revalidation

</details>

<details>
<summary>Non-blocking</summary>

- The promised follow-up issue for the event-driven cancel bridge (`loop.call_soon_threadsafe` producer wakeup, from the polling thread) doesn't appear filed yet — worth doing so the idea isn't lost.
- My two cosmetic nits from the previous review stand (unretrieved-exception theoretical case is now partially mitigated by the new debug log; `"None"` toolUseId string in sequential cancel result).

</details>

mkmeral (approved):

PR #3411, opened 2026-07-22
fix: reject keys for s3 storage if not configured

opieter-aws:
## Description

The context-offloader `S3Storage` writes every object under the configured `prefix` in `store()`, but `retrieve()` never re-checked that invariant: it accepted a raw key verbatim, and for an `s3://` URI it validated only the bucket before stripping `s3://bucket/` and using the remainder as the key. The `retrieve_offloaded_content` tool (registered by default) passes a model-supplied `reference` straight through, so an agent that takes untrusted input could be steered to read any object elsewhere in the same bucket that the SDK's credentials can `GetObject` — a prefix-scope bypass. This is the confused-deputy gap the `prefix` parameter implies protection against, and the sibling `FileStorage.retrieve()` already enforces the equivalent directory boundary.

The fix adds the missing invariant in both SDKs: after resolving a reference to a key, reject it if a prefix is configured and the key falls outside it. Empty prefix keeps today's unrestricted behavior, and in-prefix keys and `store()`-returned URIs are unaffected, so this is backward-compatible for every reference `store()` produces. The newer unified `strands.storage.S3Storage` is not affected — it prepends the prefix rather than parsing it from caller input — so no change there.

Making `retrieve()` prefix-sensitive surfaced a pre-existing normalization mismatch: the TS constructor stripped only trailing slashes from `prefix` while Python's `strip("/")` strips both ends, so a leading-slash prefix (e.g. `"/artifacts"`) produced different scopes and non-round-tripping `store()` keys across SDKs. The TS side now strips both ends to match Python, so the two SDKs normalize any prefix identically.

## Documentation PR

No documentation changes. The user guide already documents `S3Storage` with `bucket`/`prefix` and the default retrieval tool; behavior now matches that scope. A follow-up could add a defense-in-depth note recommending prefix-scoped IAM or a dedicated bucket.

## Type of Change

Bug fix

## Testing

- Python: ran the context-offloader unit suite (`hatch test tests/strands/vended_plugins/context_offloader/`), which covers the S3 store/retrieve paths including the new raw-key and `s3://`-URI out-of-prefix rejections, and linted the changed files with `hatch fmt --linter` (ruff + mypy strict).
- TypeScript: ran the offloader storage and plugin unit suites (`vitest`, node project), which cover the mirrored rejections and the retrieval-tool path, plus `eslint` and `tsc --project src/tsconfig.json` on the changed files.
- Reproduced the original bypass against the real `storage.py` with an in-memory fake S3 client, then confirmed post-fix that both attack forms raise `KeyError`, the normal `store()`→`retrieve()` round trip still returns content, and cross-bucket references remain rejected.

## Checklist

----


opieter-aws:
@strandly-the-agent  review

lizradway (dismissed):

strandly-the-agent (changes requested):
**REQUEST_CHANGES — the TypeScript change breaks persisted references created with leading-slash prefixes; preserve the existing key namespace while adding confinement.**

✅ Reviewed exact head `af3f2a86`; CI rollup is green.  
✅ Prefix-confinement attack matrices passed twice; focused Python/TS suites and TS build passed.  
🔴 One compatibility blocker at `strands-ts/src/vended-plugins/context-offloader/storage.ts:386`, reproduced cold twice against head and merge-base.  
🧭 This changes a public method’s accepted inputs/errors and requires the API-review label workflow; no API meeting is needed.  
🧪 The nonblocking Python mock assertion is already covered by `discussion_r3631980159`; I haven’t duplicated it.

<details><summary>Blocker and verification ledger</summary>

### 🔴 Preserve the existing leading-slash namespace

At `strands-ts/src/vended-plugins/context-offloader/storage.ts:386`, `prefix: "/artifacts"` is now normalized to `artifacts/`. The previous implementation stored `/artifacts/<id>` and returned `s3://b//artifacts/<id>`; the parser at `storage.ts:428-434` resolves that legacy URI to `/artifacts/<id>`, then the new check at `storage.ts:436-438` rejects both the URI and raw legacy key before contacting S3. Please retain the historical trailing-only normalization while adding confinement, and handle any namespace migration separately rather than implicitly authorizing two namespaces.

- Head rejects legacy raw and URI references using `/artifacts/<id>` and writes new keys as `artifacts/<id>`.
- Merge-base reads both legacy forms and retains the leading slash on writes.
- Verified: Python S3 tests **13 passed**; TS storage **32 passed**; TS plugin **50 passed**; TS build and `git diff --check` passed.
- Adversarial matrices covered **20 Python + 20 TypeScript cases**, each twice. S3 clients preserve dot/slash key semantics; no confinement bypass reproduced.
- Advanced API/adversarial tiers were unavailable due retention-mode validation; both passes were retried on the default tier.
- Suppressed as pre-existing: TS broad error catch/tool-result status. TSDoc/security guidance would be useful follow-up work, but is not another blocker here.

</details>

strands-ts/src/vended-plugins/context-offloader/storage.ts
```diff
@@ -378,7 +383,8 @@ export class S3Storage implements Storage {
     options?: { prefix?: string; region?: string; s3Client?: import('@aws-sdk/client-s3').S3Client }
   ) {
     this._bucket = bucket
-    this._prefix = options?.prefix ? stripTrailingSlashes(options.prefix) + '/' : ''
+    const normalizedPrefix = options?.prefix ? stripSurroundingSlashes(options.prefix) : ''
```
strandly-the-agent: 🔴 **Block — preserve the existing leading-slash namespace**

With `prefix: "/artifacts"`, released versions store `/artifacts/<id>` and return `s3://b//artifacts/<id>`. The parser at lines 428–434 resolves that URI to `/artifacts/<id>`, but stripping the leading slash here makes the check at lines 436–438 reject both the URI and raw legacy key. A cold head-vs-merge-base repro confirmed those persisted references become inaccessible after upgrading.

Please keep the historical trailing-only normalization while adding prefix confinement, and handle cross-SDK namespace migration separately. That fixes the security gap without silently moving existing TypeScript users to a different S3 key namespace.

strandly-the-agent (commented):
Inline compatibility finding attached to the existing changes-requested review.

strands-ts/src/vended-plugins/context-offloader/storage.ts
opieter-aws: Reverted

opieter-aws:
@strandly-the-agent re-review

strandly-the-agent (approved):
**APPROVE — the TypeScript namespace compatibility blocker is resolved at `40c30078`; no code blockers remain.**

✅ Historical `/artifacts/...` writes are preserved and legacy `s3://b//artifacts/foo` resolves to `/artifacts/foo`.  
✅ Prefix confinement remains unchanged; the prior blocker thread is resolved/outdated.  
✅ Local: TS storage **32/32**, build, full lint, full format check, and diff-check passed; clean tree.  
✅ Exact-head CI: all Python/TS unit matrices, lint/quality/audit/package, Docs, and CI Gate passed.  
⚠️ I could not rerun Python locally because `pytest` is unavailable; CI covered all 11 Python jobs.  
🧭 Before merge, complete the required `api/needs-review` → `api/review-complete` label workflow; both labels are still absent.

<details><summary>Re-review scope</summary>

Reviewed only `af3f2a86..40c30078`. The delta restores the released trailing-slash-only normalization and adds a regression that independently reads the old double-slash URI and asserts the exact S3 key. It does not broaden reads or introduce a dual namespace.

The existing Python `get_object.assert_not_called()` suggestion remains in `discussion_r3631980159`; I did not duplicate or escalate that previously nonblocking test-hardening request.

Two integration jobs were still running during the detailed check; the final exact-head rollup is now `SUCCESS`.

</details>

lizradway (approved):

PR #3415, opened 2026-07-22
feat(ts): add MCP tool filtering and name prefixes

### LINKED ISSUE #3161: [PORT] MCP tool filtering + name-prefixing (Py→TS)
> ⚠️ Filed by Strandly as part of the #2498 parity audit. Source feature verified merged + unit-tested + integ-tested + documented in `strands-py`. Agent-produced — confirm scope before implementing. (Add the `port` label — I can't apply it via API.)

**Parent:** #2498

## Source SDK
`strands-py`

## Target SDK
`strands-ts`

## Related Files
```
strands-py/tests/strands/tools/mcp/test_mcp_client_tool_provider.py
strands-py/tests/strands/tools/mcp/test_mcp_client_load_servers.py
strands-py/tests_integ/mcp/test_mcp_tool_provider.py
site/src/content/docs/user-guide/concepts/tools/mcp-tools.mdx
```

## What needs to be migrated
Python's MCP client filters/prefixes tools when listing; TS `McpClient` (`strands-ts/src/mcp/client.ts`) has neither.
- **Types** (`mcp_client.py`): `_ToolFilterCallback` Protocol (`:68`), `_ToolMatcher = str | Pattern[str] | callback` (`:72`), `ToolFilters(TypedDict, total=False)` with `allowed`/`rejected` (`:75`).
- **Constructor opts:** `tool_filters` (`:210`), `prefix` (`:211`).
- **Application** in `list_tools_sync(pagination_token=None, prefix=None, tool_filters=None)` (`:536`): three-state override (explicit arg incl. `""`/`{}` overrides ctor default; `None` = use default, `:561`); prefix builds `name_override=f"{prefix}_{tool.name}"` (`:580`); filter via `_should_include_tool_with_filters` (`:588`).
- **Filter semantics:** `_should_include_tool_with_filters` (`:1143`) — allowed-then-rejected, rejected wins; `_matches_patterns` (`:1160`) — callable→call; `Pattern`→`.match` (anchored-at-start!); `str`→exact eq; **all match the server-side UNPREFIXED name** (`tool.mcp_tool.name`).
- **Config-file:** `MCPServerConfig` carries `prefix`/`tool_filters` (`:109`); `_parse_config_tool_filters` (`:1481`), `_compile_filter_patterns` (`:1498`, raises `ValueError` on bad regex).

## TypeScript landing points
- `strands-ts/src/mcp/client.ts`: add `prefix?`/`toolFilters?` to `McpClientOptions` (`:79-102`), store in ctor (`:159`), apply in `listTools()` (`:318-346`) after constructing each `McpTool` (`:329`), before pushing; update `_registeredToolNames` (`:344`) with effective (prefixed) names.
- Config path: extend `McpServerConfig` (`mcp/config.ts:10-33`), compile in `config.node.ts`.
- Agent integration transparent (filter/prefix inside `listTools()`); ensure `onToolsChanged` refresh (`agent.ts:781`) applies the same.

## Gotchas / open questions
- **`re.Pattern.match` is start-anchored; JS `RegExp.test` is unanchored.** Prepend `^` (or use anchored logic) or filtering diverges. Biggest correctness trap.
- **Match the UNPREFIXED name; only the registered name gets prefixed** — preserve ordering.
- **Three-state override** (`None` vs explicit `{}`/`""`): `undefined`=default, but callers passing `''`/`{}` must still override — decide whether `listTools()` gains optional params (matching `list_tools_sync`) or ctor-only. API-shape decision.
- `str | RegExp | (tool)=>boolean` union; `TypedDict total=False` → interface with optional fields; bad-regex `ValueError` → clear `SyntaxError` wrap.

- Merged: ✅  ·  Unit tests: ✅ (str/regex/callable allowed+rejected + prefix rename in `test_mcp_client_tool_provider.py:252-367`; config filters in `test_mcp_client_load_servers.py`)  ·  Integ: ✅ (`test_mcp_tool_provider.py:18-115`)  ·  Docs: ✅ (`mcp-tools.mdx`).

## Confirmation
- [x] The change is complete in the source SDK (merged + unit + integ tested + documented); every related artifact is listed above.

strandly-the-agent:
Ports MCP tool filtering and agent-facing name prefixing from `strands-py` to `strands-ts`.

Closes #3161

## What this does

- `toolFilters` / `prefix` on `McpClientOptions`, overridable per listing via `listTools({ prefix, toolFilters })`
- Three matcher kinds, matching Python: a **string** matches the server-side tool name exactly, a **`RegExp`** matches it from the start (`Pattern.match`), a **callback** receives the tool
- The prefix is applied first and filters run against the resulting tool — so callbacks see the agent-facing name while strings/regexes match the server-side name (`mcp_client.py:1371`)
- A prefix only renames the tool for the agent; calls still send the server-side name (`mcp_agent_tool.py:115`)
- `prefix` / `toolFilters` in `McpClient.loadServers()` config, interpolated then compiled to regexes

## Revised after review feedback

`arielnabavian` raised three concerns on the first revision — naming, excessive `async`, and changes outside a strict port. All three were valid; this revision addresses them by **reverting the out-of-scope work** rather than defending it. Diff went from 907+/58- across 14 files to **519+/28- across 10 files**, and `agent.ts` / `tool-registry.ts` are no longer touched at all.

<details><summary><b>What was removed, and the one behavior bug the cleanup exposed</b></summary>

Removed as outside a Python→TypeScript port (all of it was self-inflicted hardening from my own prior review loop, none of it required by the source feature):

| Removed | Why it was out of scope |
|---|---|
| Promise-queue mutexes around `connect()` / `disconnect()` | Pre-existing lifecycle; unrelated to filtering/prefixing |
| `_listToolsQueued` serialization + `notifyInternalListeners` boolean-trap param | Same, and the split forced an unidiomatic internal signature |
| `tools-changed-listeners.ts` multi-subscriber module | Redesigned a pre-existing single-callback API |
| `_cloneTools()` defensive deep-cloning | Changed what `listTools()` returns to defend against a hypothesis |
| `Agent.initialize()` single-flighting + transactional registry rollback | Rewrote agent lifecycle |
| `registry/replace-owned-tools.ts` | New registry primitive, not needed by the port |

Re-deriving the port straight from the Python source also surfaced a **real fidelity bug** in the first revision: it filtered *before* prefixing and passed the **unprefixed** tool to callback matchers. Python builds the prefixed `MCPAgentTool` first, then filters, so a callback sees `tool_name` **with** the prefix while string/regex matchers compare `tool.mcp_tool.name` **without** it (`mcp_client.py:611-620`, `:1371-1384`). The previous revision had a test asserting the wrong behavior; both are corrected here, and the docs sentence that stated it has been fixed too.

On `async` specifically: no new `async`/`await` outside tests. `shouldIncludeTool`, `matchesAnyMatcher`, `compileToolFilters` and `compileFilterMatchers` are all synchronous, and option resolution in `listTools` is plain assignment ahead of the pre-existing pagination loop.

Start-anchoring is now done with a fresh sticky (`y`) `RegExp` rather than index arithmetic — it anchors at position 0, leaves a caller-supplied stateful `/g` or `/y` regex untouched, and handles alternation (`/a|b/`) correctly.
</details>

<details><summary><b>Source-test → target-test traceability</b></summary>

| Python source test | Behavior | TypeScript target test |
|---|---|---|
| `test_mcp_client_tool_provider.py:60` | ctor `tool_filters` + `prefix` | `client.test.ts:375` |
| `test_mcp_client_tool_provider.py:252` | exact-string matcher | `client.test.ts:322` |
| `test_mcp_client_tool_provider.py:271` | regex matcher, start-anchored | `client.test.ts:322` |
| `test_mcp_client_tool_provider.py:290` | callback matcher (receives the renamed tool) | `client.test.ts:322` |
| `test_mcp_client_tool_provider.py:312` | rejected matcher | `client.test.ts:322`, `:360` |
| `test_mcp_client_tool_provider.py:331` | prefix renames tools | `client.test.ts:375`, `:408` |
| `test_mcp_client_tool_provider.py:510`, `:553`, `:596` | prefix override / `""` disables / omitted inherits | `client.test.ts:375` |
| `test_mcp_client_tool_provider.py:639`, `:680`, `:721` | filters override / `{}` disables / omitted inherits | `client.test.ts:360`, `:375` |
| `test_mcp_client_tool_provider.py:761` | combined prefix + filter override | `client.test.ts:375` |
| `test_mcp_client_tool_provider.py:871`, `:915` | regex / callable per-call override | `client.test.ts:322`, `:375` |
| `test_mcp_client_tool_provider.py:223` | pagination | `client.test.ts:408` |
| `test_mcp_client_tool_provider.py:472` | one client reused across two Agents | `mcp.test.node.ts:92` |
| `test_mcp_client_load_servers.py:139` | config `prefix` | `config.test.node.ts:307`, `:365` |
| `test_mcp_client_load_servers.py:164` | config strings compile to regexes | `config.test.node.ts:307` |
| `test_mcp_client_load_servers.py:173` | invalid config regex raises contextually | `config.test.node.ts:404`, `:410` |
| `tests_integ/mcp/test_mcp_tool_provider.py:18` | real server: filters + prefix | `mcp.test.node.ts:77` |
| `tests_integ/mcp/test_mcp_tool_provider.py:44` | prefixed tool invokes the server-side name | `mcp.test.node.ts:77`, `client.test.ts:408` |
| `tests_integ/mcp/test_mcp_tool_provider.py:69` | shared client, both Agents execute | `mcp.test.node.ts:92` |
| `tests_integ/mcp/test_mcp_tool_provider.py:105` | two servers, distinct prefixes | `mcp.test.node.ts:109` |

TypeScript-only coverage, for behavior that only exists because TS has a `onToolsChanged` refresh path that Python has no counterpart for:

| Target test | Invariant |
|---|---|
| `client.test.ts:764` | a transient per-call override does not become the refresh baseline |
| `client.test.ts:782` | refresh reapplies the constructor prefix + filters |
| `client.test.ts:438` | task-based invocation also uses the server-side name |
| `config.test.node.ts:327`, `:348` | `${VAR}` interpolation reaches the new config fields, incl. missing-var handling |
</details>

<details><summary><b>Validation</b></summary>

Run on `ba4a8eb7c`, rebased onto current `main` (`526692f53`):

- unit suite: **141 files, 3903 tests passed**, no type errors, coverage 93.65%
- MCP + config focused: **140 passed**
- MCP integration (real stdio server, live Bedrock): **8 passed**, including the 3 new Agent-path cases
- `lint`, `format:check`, `type-check`, `build`, `check:browser-bundle`: pass
- docs: **473 passed** / 2 skipped, snippet type-check clean, site build **873 pages**
- no dependency or lockfile changes

Note: `site` `npm run typecheck` reports 24 pre-existing `error TS` in `src/util/related-docs.ts`; verified identical on a pristine `main` checkout, so unrelated to this change.
</details>

<details><summary><b>Review suggestions I did not take, and pre-existing issues I deliberately left alone</b></summary>

Declined, to keep the diff a port:

- **`prefix?: string | null` with `null` as the disable sentinel** instead of `''`. Cleaner in isolation, but it would deliberately diverge from the source contract this issue specifies (`None` = inherit, explicit `""`/`{}` = disable). Worth a follow-up across both SDKs if you want it — not a unilateral call inside a port.
- **`logger.warn` when the server-name lookup falls back** to `tool.name` in `callTool`. No reachable path produces it today; it would be speculative code.
- **Renaming `McpServerToolFilters` → `McpServerConfigToolFilters`.** Reads fine next to `McpServerConfig`.

Pre-existing on `main`, unchanged here, and worth separate issues rather than being smuggled into a port:

1. `agent.ts` assigns `client.onToolsChanged` as a single slot, so two Agents sharing one `McpClient` clobber each other's refresh callback. Registration and invocation are unaffected (which is what `mcp.test.node.ts:92` asserts); only a live tools-changed refresh would surface it.
2. `McpClient.connect()` is check-then-act on `_state` with no lock, so concurrent first calls can race the handshake.

Happy to file both as issues if you'd like them tracked.
</details>

---

**API surface:** additive only — 5 new exported types, 2 new optional fields on `McpClientOptions`/`McpServerConfig`, 1 new optional trailing param on `listTools()`. Nothing renamed, removed, or narrowed; existing callers compile and behave identically. An independent API pass suggested this still warrants an `api/needs-review` label as process hygiene (precedent: #2947 on this same surface) — I can't apply labels via the API, so flagging it here.


arielnabavian (dismissed):

site/src/content/docs/user-guide/concepts/tools/mcp-tools.mdx
```diff
@@ -354,20 +354,20 @@ agent = Agent(tools=[aws_docs_client, other_client])
 </Tab>
 <Tab label="TypeScript">
 
-TypeScript's `McpClient` accepts optional application metadata:
+TypeScript's `McpClient` supports the same tool filtering and name prefixing behavior:
```
Unshure:
```suggestion
TypeScript's `McpClient` supports tool filtering and name prefixing to manage tools from multiple servers.
```

site/src/content/docs/user-guide/concepts/tools/mcp-tools.mdx
```diff
@@ -354,20 +354,20 @@ agent = Agent(tools=[aws_docs_client, other_client])
...
-  transport: new StdioClientTransport({
-    command: 'npx',
-    args: ['-y', 'some-mcp-server'],
-  }),
-})
+--8<-- "user-guide/concepts/tools/mcp-tools.ts:tool_filtering_prefixing"
 ```
 
-Tool filtering and prefixing are not currently supported in TypeScript.
+String matchers match the server-side tool name exactly, and `RegExp` matchers match it from the start. Callback
```
Unshure: Can we copy the above python example to the typescript tab here?

strands-ts/src/mcp/config.ts
```diff
@@ -1,6 +1,17 @@
 import type { McpClientConfig, McpClientCredentials, McpClientOptions, TasksConfig } from './client.js'
 import { createDefaultSlot } from '../default-slot.js'
 
+/**
+ * Tool filters for an MCP server config entry. Because a config file cannot carry a `RegExp`, each
+ * pattern is a string compiled to a regex, matched from the start of the server-side tool name.
+ */
+export interface McpServerToolFilters {
```
Unshure: nit: Can we workshop this name a bit? I'm a bit confused about the difference between `McpToolFilters` and `McpServerToolFilters`

site/src/content/docs/user-guide/concepts/tools/mcp-tools.mdx
arielnabavian: @strandly-the-agent copy this from https://strandsagents.com/docs/user-guide/concepts/tools/mcp-tools/ python docs

strandly-the-agent:
Docs ask done in `dba7756` (`arielnabavian`, `Unshure` — replying here because the thread is on an outdated line).

The TypeScript tab of **Client Configuration** is now a mirror of the Python tab: same two subsections (**Tool Filtering** / **Tool Name Prefixing**), same lead-in sentences, same `awslabs.aws-documentation-mcp-server` server, same variable names and inline comments, same `string` → `regex` → `combined` progression, then the same two prefixed clients (`aws_docs` / `other`) feeding one `Agent`.

One deliberate difference: the TS filtering block also shows a standalone **callback** matcher, which the Python tab doesn't have. TS supports it and the prose below the block explains it, so dropping it would lose the only example of that matcher kind — happy to remove it if you'd rather the tabs be exactly 1:1, or to file a follow-up adding it to the Python tab (`tool_filters` takes callables too).

<details><summary><b>What changed, what I verified, and one thing I did not do</b></summary>

The old single `tool_filtering_prefixing` snippet is split into `tool_filtering` and `tool_prefixing` so each subsection gets its own block like Python's. The TS-only prose (matcher semantics, per-listing overrides, `loadServers()` config) is kept and reflowed under 90 chars.

Verified:

- Every filter in the new snippets was checked against the implementation **behaviorally**, not just read: `allowed: ['search_documentation','read_documentation']` yields exactly those two; `/^search_.*/` yields only `search_documentation`; the `_internal` callback drops only that tool; `allowed: [/.*documentation$/]` + `rejected: ['read_documentation']` yields `search_documentation` (rejection wins); `prefix: 'aws_docs'` renames to `aws_docs_search_documentation` while `callTool` still sends `search_documentation`. Also confirmed `/documentation$/` **without** the leading `.*` matches nothing — start-anchoring is real, so Python's `.*` carries over as-is.
- `site`: `npm run typecheck:snippets` clean, `prettier --check` clean on the snippet file, full `npm run build` → 873 pages, no broken links, and I read the built HTML to confirm both blocks render with no leaked markers or scaffolding.
- No `strands-ts` source, dependency or lockfile changes in this commit; two independent fresh-context review passes on the commit, both approving.

**Not done, and worth saying out loud:** `mdx-authoring.md` wants a sibling `mcp-tools_imports.ts` so rendered TS blocks show their imports (the Python block does). I skipped it because all 11 TS snippets on this page currently render import-less, so adding one for just these 2 would make the page *less* internally consistent and widen a diff you've twice asked me to keep narrow. It's a page-wide (arguably site-wide) gap — say the word and I'll do the whole page here, or file it separately.

</details>

---

On the other open nit — `McpServerToolFilters` vs `McpToolFilters` (`strands-ts/src/mcp/config.ts:8`): the confusion is fair, and I'd flagged this name as fine in an earlier round, which your comment is decent evidence I got wrong. The real distinction is *what the matchers are*, not *where they live*: `McpToolFilters` holds `string | RegExp | callback` matchers; the config type can only hold strings, because a config file can't carry a `RegExp`, and those strings are compiled to regexes. So my preference is **`McpToolFilterPatterns`** — it names the difference. Alternatives: `McpServerConfigToolFilters` (explicit about where it's used, more verbose), or drop the export entirely and inline `{ allowed?: string[]; rejected?: string[] }` on `McpServerConfig` so there's only one `*ToolFilters` type in the surface. It's a 4-line, additive-only rename (nothing released yet) — tell me which you want and I'll push it.

arielnabavian:
@strandly-the-agent how about SerializableMcpToolFilters instead of McpServerToolFilters

strandly-the-agent:
Works for me — `Serializable` names the actual constraint, which is what the old name was missing. Renamed in `b98eca1`: `SerializableMcpToolFilters` in `strands-ts/src/mcp/config.ts`, re-exported through `mcp/index.ts` and `src/index.ts`, and the interface's JSDoc reworded to match ("tool filters in a serializable form"). Nothing else in the repo referenced the old name, so it's a 3-file, type-only change to an export that first appears in this PR — no released surface moves.

<details><summary><b>Verified</b></summary>

- `rg "McpServerToolFilters"` across the repo (source, tests, `site/`, `dist/`) → 0 hits.
- Consumer check against the built package: `import type { SerializableMcpToolFilters } from '@strands-agents/sdk'` type-checks and is assignable to `McpServerConfig['toolFilters']`; importing the old name now fails with `TS2724 … Did you mean 'McpToolFilters'?` — which is a neat confirmation of the original complaint: the old name was close enough to autocomplete into the wrong type.
- `type-check`, `lint`, `format:check` clean; 140 MCP unit tests pass; full pre-commit suite (build + coverage) passed on commit.
- Independent fresh-context review of the commit: approved, no findings. It also checked `strands-py` — Python uses one `ToolFilters` TypedDict for both the API and config flavours, so there's no counterpart name to keep in sync.

</details>

Still open from earlier, whenever you want to weigh in: whether to keep the extra callback-matcher example in the TS docs tab that the Python tab doesn't have, and the page-wide missing-`_imports.ts` docs gap. Neither blocks this PR.

Unshure (approved):

arielnabavian (approved):

PR #3416, opened 2026-07-22
feat(site): community integration catalog with search, filtering, and curated backfill

yonib05:
## Description

The community page was a set of static tables generated from doc-page frontmatter — no search, no filtering, no way to feature entries, and every listing required a full doc page. As the ecosystem grew (vendor integrations from Temporal, Datadog, Grafana, Arize; dozens of community packages), the page stopped scaling as a discovery surface.

This PR replaces it with an **Integrations** page at `/integrations`: a searchable, filterable grid of 77 community integrations backed by a YAML content collection. Each integration is one small YAML file validated by a zod schema at build time, so a malformed submission fails CI — the schema is the review gate. Entries support three shapes: dedicated packages (`strands-deepgram`), integrations shipped inside larger SDKs via extras (`temporalio[strands-agents]`), and config-only vendor guides with no installable package (Langfuse, Portkey). Dual-language integrations are one entry with per-language package metadata; single-language detail pages disable the global language toggle instead of silently no-oping.

The page follows the changelog's design language: left rail for type/highlight filters, search palette dialog (`/` shortcut) mirroring the global site search, uniform fixed-anatomy cards with type-glyph tiles, and Starlight `sl-badge`-aligned Featured/New/Verified flags. Filter state syncs to the URL for shareable views; with JS disabled the full catalog renders statically. Featured and trust badges are maintainer-granted: a CI workflow labels PRs where a non-maintainer changes editorial fields (indicator, not a gate — the team can merge anyway).

Popularity stats (GitHub stars, registry downloads) come from a weekly scheduled workflow that commits a JSON file via bot PR; per-package fetch failures retain previous values. The one-time backfill was seeded from four passes — registry search, the awesome-strands-agents list, a deep-research sweep (which surfaced integrations whose names don't contain "strands": Grafana Sigil, Arize OpenInference), and strands.my — each human-curated with spam/abandoned/example-app candidates excluded.

The old `community-packages` page redirects to `/integrations`, and the whole `docs/community/**` namespace moved to `docs/integrations/**` with build-time redirect stubs for every published URL. The get-featured guide is rewritten as the submission flow (one YAML file + PR template); new submissions link their own docs via `docsUrl` instead of adding on-site pages. Cards with legacy on-site pages open in a slide-over drawer that keeps the grid's filter state; the full-page views drop the docs sidebar and carry a back-to-integrations strip.

## Documentation PR

Documentation changes are included: the rewritten submission guide (`site/src/content/docs/integrations/get-featured.mdx`), redirect from the retired `community-packages` page, and updated inbound links.

## Type of Change

New feature

## Testing

Ran the full site suite (`npm test`, 524 passing) covering schema validation and rejection cases, catalog↔docs integrity (every `docsPage` resolves; every community doc page has an entry), card-model derivation, filter AND/OR semantics with URL round-tripping, and stats keep-previous-on-failure. `npm run build` passes with the broken-links checker green. Exercised the page end to end against the dev server: search/filter/URL-state behavior, featured/new/verified facets, the search dialog, redirect from the old page, disabled language toggle on single-language pages, and light/dark/mobile rendering via headless-browser screenshots. The stats and backfill scripts were run live against GitHub/PyPI/npm to produce the committed data. The two GitHub Actions workflows are validated (YAML parse + embedded-script syntax + regex exercise) but can only be exercised on the upstream repo after merge via `workflow_dispatch`.

## Checklist

## Deployment notes

The renamed `docs/community/**` URLs are covered in-repo by static meta-refresh stubs (0-second refresh, canonical link to the new URL, `noindex` on the stub, visible fallback link) plus the client-side 404 prefix rule for unlisted paths. Meta-refresh@0 is treated as a redirect signal by Google but is weaker than a real 301, and server-side redirects are not configurable from this repository — the site is served from S3 behind CloudFront, whose distribution config lives outside this repo.

**Recommendation for whoever owns the CloudFront distribution:** add a CloudFront Function (viewer-request) or equivalent 301 mapping so the old URLs return real 301s:

- `/docs/community/community-packages/*` → `301` → `/integrations/`
- `/docs/community/*` → `301` → `/docs/integrations/*` (preserving the path suffix)

Order matters: the `community-packages` rule must be evaluated before the general `/docs/community/*` prefix rule. Until that lands, the static stubs remain the redirect mechanism of record and must not be removed.

----


yonib05:

.github/workflows/catalog-editorial-review.yml
```diff
@@ -0,0 +1,375 @@
...
+            // it directly without requiring a `badges:` line in the diff (the
+            // header often survives as context when only items are removed).
+            const removedListItemRe = /^-\s*-\s*['"]?verified['"]?\s*$/;
+
+            const flagged = [];
+            for (const file of catalogFiles) {
+              if (!/\.ya?ml$/.test(file.filename)) continue;
+              // Catalog entries are small, so a patch is always present;
+              // a missing one means no reviewable text diff — skip.
+              const lines = (file.patch || '').split('\n');
```
strandly-the-agent: 🔴 **Patch syntax can bypass both review gates.** This treats the textual patch as the catalog source of truth. JSON/flow-style YAML with `featured: true` and `badges: [verified]` passes the real schema but matches none of the line regexes; changing only an existing entry's `github` URL also creates no `addedPackages` because only added `package:` lines are collected. Both paths reach the stale-flag cleanup at lines 192–195.

I reproduced the semantic bypass twice: `schemaAccepted=true`, `finalLabel=false`, and the sticky warning was marked resolved. Please fetch and parse the exact base/head YAML, compare `featured`/`badges` semantically, and rerun provenance whenever either side of `(github, language, package)` changes. Expanding the regex set would remain serialization-dependent.

site/scripts/catalog/refresh-stats.ts
```diff
@@ -0,0 +1,177 @@
...
+
+export const liveFetchers: StatsFetchers = {
+  async githubRepo(repoUrl) {
+    // Keep only the org/repo segments so tree/blob URLs resolve to the repo.
+    const slug = new URL(repoUrl).pathname
+      .replace(/^\/|\/$/g, '')
+      .split('/')
+      .slice(0, 2)
+      .join('/')
+    const repo = (await fetchJson(`https://api.github.com/repos/${slug}`, githubApiHeaders())) as {
```
strandly-the-agent, line 110: 🟡 **Malformed 2xx payloads bypass the keep-previous path.** These casts do no runtime validation. In a repeated repro, GitHub `{}` and npm `{ "downloads": "not-a-number" }` were treated as success: prior `stars: 99` / npm `20` were lost and the rendered aggregate became invalid instead of preserving committed values.

Please validate all three extracted counts as finite, non-negative numbers and throw otherwise so `buildStats` reaches its existing per-source catch. A test at the live-adapter parsing boundary would have caught this; current tests replace all three adapters with typed mocks.

site/src/content/catalog/llm-tracekit-strands.yaml
```diff
@@ -0,0 +1,9 @@
+name: llm-tracekit-strands
+description: OpenTelemetry instrumentation for Strands Agents that captures spans, traces, and metrics compatible with Coralogix and other OTel backends.
+integrationType: plugin
+languages:
+  python:
+    package: llm-tracekit-strands
+github: https://github.com/coralogix/llm-tracekit
+docsUrl: https://coralogix.com/docs/llm-tracekit-strands/
```
strandly-the-agent: 🟡 This primary documentation URL currently returns HTTP 404. The live Coralogix page is the Strands integration guide:

```suggestion
docsUrl: https://www.coralogix.com/docs/integrations/ai-observability/strands/
```

.github/workflows/catalog-editorial-review.yml
```diff
@@ -0,0 +1,375 @@
+name: "Catalog: Editorial Review Flag"
+
+# No paths filter: a push that removes every catalog change must still run so
+# the stale label/comment from an earlier push gets cleared. The script
+# early-exits cheaply when the PR touches no catalog files.
+on:
+  pull_request_target:
+    branches: [main]
+    types: [opened, synchronize, reopened]
```
strandly-the-agent, line 9: 🔴 **Older `synchronize` runs can overwrite the latest result.** This state-mutating workflow has neither PR-scoped concurrency nor a current-head check before label/comment mutations. A deterministic two-run harness reproduced both directions twice: an old clean run erased a valid latest flag, and an old flagged run overwrote the latest clean state.

Please add PR-scoped concurrency with `cancel-in-progress: true` and revalidate the current PR head immediately before mutations, e.g. use a group keyed by `${{ github.event.pull_request.number }}`.

.github/PULL_REQUEST_TEMPLATE/catalog-submission.md
```diff
@@ -0,0 +1,46 @@
...
+**GitHub repo:** <!-- source of truth for the maintainer shown on your card (the URL's owner/org) -->
+**What it does:** <!-- 1–2 sentences -->
+**Relationship to service:** <!-- official vendor integration / community-built -->
+**Docs link included:** <!-- docsPage (on-site page) / docsUrl (external link) / none — card will link to GitHub repo -->
+
+### Submitter checklist
+
+#### Package
+- [ ] This is a reusable integration (library/plugin/provider), not an example agent or application built with Strands
+- [ ] Published to PyPI and/or npm under the exact `package` name in my entry (registry links are derived from it)
```
strandly-the-agent: 🟡 **This checklist excludes a supported submission shape.** The schema intentionally permits guide-only entries with an empty language block, and this tree contains six of them, so those contributors cannot truthfully check the registry-publication requirement (or the unconditional install/import check at line 41).

Please make package publication/install conditional on package-backed entries and add a guide-only path requiring a working Strands-specific setup guide.

site/src/content.config.ts
```diff
@@ -71,6 +71,77 @@ export const changelogFrontmatterSchema = z
...
+      'memory-store',
+      'integration',
+      'plugin',
+      'agent-extension',
+      'intervention',
+    ]),
+    // Which SDK's ecosystem this belongs to. The catalog's SDK facet stays
+    // hidden until at least one evals entry exists.
+    sdk: z.enum(['agents', 'evals']).default('agents'),
+    languages: z.object({
```
strandly-the-agent: 🟡 **The language container is the one non-strict layer in this schema.** The real schema accepted `languages: { python: {}, typeScript: { package: "example" } }` and silently stripped `typeScript`, publishing a Python-only entry instead of failing the advertised build-time gate.

```suggestion
    languages: z
      .object({
        python: catalogLanguageSchema.optional(),
        typescript: catalogLanguageSchema.optional(),
      })
      .strict(),
```

Please also add the typo case to `catalog-collection.test.ts`.

site/src/components/LanguageToggle.astro
```diff
@@ -190,8 +217,18 @@ function switchHref(lang: LanguageSlug): string | undefined {
...
-      document.querySelectorAll('language-toggle .lang-option').forEach(function (btn) {
-        btn.setAttribute('aria-checked', btn.dataset.lang === lang ? 'true' : 'false');
+      // On single-language pages the stored preference may be the disabled
+      // language; the available language is then shown as selected so exactly
+      // one radio is always checked.
+      var enabled = Array.prototype.filter.call(
+        document.querySelectorAll('language-toggle .lang-option'),
+        function (btn) { return !btn.disabled; }
+      );
+      var effectiveLang = enabled.some(function (btn) { return btn.dataset.lang === lang; })
```
strandly-the-agent: 🟡 **This effective selection is only applied to the radio, so it can disagree with navigation and tabs.** `syncAll()` below still passes the original stored preference to `updateNavLinks()` and `syncStarlightTabs()`. On a Python-only page with TypeScript stored, the repeated repro visibly selected Python while the Docs link still targeted the TypeScript quickstart.

Would you compute the effective language once in `syncAll()` and use it for the radio, navigation, and inline tabs? The persisted global preference can remain unchanged if preserving it across pages is intentional.

strandly-the-agent (changes requested):
🔴 **Changes requested:** the editorial workflow can miss valid catalog changes and stale runs can overwrite the latest review state.

✅ Reviewed `9659c74f5a7a304bc723c434434b8f4e9eb0d768` vs `e03f8644b0525c567aa3a87b1238b422fcef1bf3` · 493 site tests passed (2 skipped) · typecheck/diff-check passed.
🔴 Five cold repros were each repeated twice; seven findings are inline. Exact-head Docs CI, preview, CodeQL, API-label check, and CI Gate are green; only the XL-size policy fails. The PR is also currently conflicting.

<details>
<summary>Review breakdown</summary>

- **Correctness/adversarial:** Two blockers in the privileged editorial workflow: serialization-dependent patch parsing and stale-run races. Also reproduced malformed-success stats erasing prior data, a silently stripped language typo, and language-toggle split-brain.
- **Docs/DevX:** One dead primary catalog link and a submission checklist that excludes the guide-only shape the schema intentionally supports.
- **Test quality:** Existing tests are green, but the privileged workflow and live adapter parsing boundary are untested; test requests are folded into the demonstrated defects rather than posted separately.
- **API:** No exported SDK API changed; the exact-head API-label check is green. The advanced API pass was unavailable in this environment, and two default-tier retries returned no result.
- **Suppressed as duplicates:** request timeouts, schema/UI enum drift, PyPI URL encoding, and the obsolete CodeQL alert were already raised in existing threads.
- **LLM context:** not routed; no model-facing SDK surface changed.

No blocking questions—the retained findings have deterministic behavior and concrete fixes.
</details>

.github/workflows/catalog-editorial-review.yml
yonib05: Fixed.

site/scripts/catalog/refresh-stats.ts
yonib05: Fixed.

site/src/content/catalog/llm-tracekit-strands.yaml
yonib05: Fixed.

.github/workflows/catalog-editorial-review.yml
yonib05: Fixed.

.github/PULL_REQUEST_TEMPLATE/catalog-submission.md
yonib05: Fixed.

site/src/content.config.ts
yonib05: Fixed.

site/src/components/LanguageToggle.astro
yonib05: Fixed.

site/scripts/catalog/refresh-stats.ts
```diff
@@ -0,0 +1,177 @@
...
+      stargazers_count: number
+    }
+    return { stars: repo.stargazers_count }
+  },
+  async pypiDownloads(pkg) {
+    // pypistats.org: last-month downloads for the package.
+    const data = (await fetchJson(`https://pypistats.org/api/packages/${pkg}/recent`)) as {
```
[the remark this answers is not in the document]
yonib05: Fixed.

site/src/components/catalog/types.ts
```diff
@@ -0,0 +1,77 @@
+// Canonical registry of catalog integration types: filter value, display
+// labels, and a 24x24 single-path glyph (fill=currentColor) rendered in cards
+// and facet chips. The zod enum in src/content.config.ts lists the same
+// values — keep the two in sync.
```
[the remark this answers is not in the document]
yonib05: Addressed — a sync test now guards the registry against the schema enum.

site/scripts/catalog/backfill.ts
```diff
@@ -0,0 +1,256 @@
...
+    .map((p) => p.name)
+    .filter((name) => name.includes('strands') && !name.startsWith('strands-agents'))
+  console.log(`matched=${names.length} | pypi simple index filtered`)
+  for (const name of names) {
+    try {
+      const meta = (await fetchJson(`https://pypi.org/pypi/${name}/json`)) as { info: any }
+      const info = meta.info
+      const repoUrl: string | undefined =
+        info.project_urls?.Source || info.project_urls?.Repository || info.project_urls?.Homepage
+      if (!repoUrl?.includes('github.com')) continue
```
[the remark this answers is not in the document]
yonib05: Addressed — backfill.ts was removed from this branch.

yonib05:
@strandly-the-agent can you re-review this PR?

.github/workflows/catalog-editorial-review.yml
```diff
@@ -0,0 +1,418 @@
...
+            // spellings can reach main.
+            function editorialState(text) {
+              const featured = /['"]?featured['"]?\s*:\s*['"]?(true|True|TRUE)\b/.test(text);
+              const badges =
+                // Flow / JSON-ish non-empty list: badges: [verified]
+                /['"]?badges['"]?\s*:\s*\[\s*[^\s\]]/.test(text) ||
+                // Block-style list: a badges: line with at least one item
+                /['"]?badges['"]?\s*:\s*(#[^\n]*)?\r?\n(?:\s*(?:#[^\n]*)?\r?\n)*\s*-\s*\S/.test(text);
+              return { featured, badges };
+            }
```
strandly-the-agent: 🔴 **Detection is still regex-on-raw-YAML, so a grant can evade it — and evasion actively clears the flag.**

Moving from patch text to full base-vs-head content genuinely fixed the round-1 whole-document bypass. What remains is that `editorialState()` matches raw text, and js-yaml 4.3.0 — the exact parser Astro's content loader uses (`node_modules/astro/dist/content/loaders/file.js:17`) — parses several forms *identically to the flagged control* while these regexes miss them. Reproduced twice against the real parser:

```
CONTROL featured: true              | parsed featured=true       | flagged=true
featured: !!bool true               | parsed featured=true       | flagged=false  <== BYPASS
featured: &f true                   | parsed featured=true       | flagged=false  <== BYPASS
"\u0066eatured": true               | parsed featured=true       | flagged=false  <== BYPASS
CONTROL badges: [verified]          | parsed badges=["verified"] | flagged=true
badges: !!seq [verified]            | parsed badges=["verified"] | flagged=false  <== BYPASS
badges: &b [verified]               | parsed badges=["verified"] | flagged=false  <== BYPASS
"\u0062adges": [verified]           | parsed badges=["verified"] | flagged=false  <== BYPASS
```

Worse than a miss: with `flagged=[]` and provenance MATCH, control reaches the clear path at lines 310-313, so push 1 with `featured: true` gets flagged and push 2 rewritten as `featured: &f true` **removes the label and rewrites the sticky comment to “~~Resolved~~ — a later push removed the flagged changes”** while the grant is still live. That inverts the signal, against the standard this file sets for itself at lines 157-159 (“a false negative defeats the flag”).

Comparing the *parsed* values instead fixes this, the ReDoS below, and the provenance evasion in one change:

```suggestion
            function editorialState(text) {
              let data;
              try {
                data = yaml.load(text) ?? {};
              } catch {
                // Unparseable YAML: fail toward flagging, never away from it.
                return { featured: true, badges: true };
              }
              return {
                featured: data.featured === true,
                badges: Array.isArray(data.badges) && data.badges.length > 0,
              };
            }
```

One caveat I could not verify from here: `actions/github-script` doesn't guarantee `js-yaml` is resolvable, so this needs `const yaml = require('js-yaml')` backed by a step that makes it available (a checkout plus require from `site/node_modules`, or a tiny install step) — or, if you'd rather keep the job dependency-free, a minimal hand-rolled parse of just these two fields.

site/src/util/redirect.ts
```diff
@@ -62,10 +62,27 @@ type SlugRule =
...
+    to,
+  })),
+
+  // community-packages content lives on the interactive integrations page.
+  // It is an Astro page, not a docs content file, so buildStaticRedirects
+  // can't validate it as a target — these rules stay client-side only.
+  {
+    match: exactly('docs/community/community-packages'),
+    to: 'integrations',
+  },
```
strandly-the-agent: 🟡 **A long-published page loses its crawler-followable redirect.** Both new rules live in `SLUG_RULES`, so they're client-side-404-fallback only — no build-time stub, and a crawler gets HTTP 404. `catalog` → `integrations` was dev-only, fine. But `docs/community/community-packages` is a real published page (it exists at the merge base, added 2026-05-22), so this is exactly the backlink-equity loss `redirect.static.ts` describes at lines 5-6 — for the one rename in this PR that most needs a stub.

Your in-code reason is accurate: `buildStaticRedirects` rejects a non-docs target (I reproduced `redirect target "integrations" … has no content file`). But the repo already carries the allowlist this needs — `site/test/sitemap-coverage.test.ts:180` has `validNonDocsPages = new Set(['integrations'])`. Threading that same set into `buildStaticRedirects` as an allowlist would let both rules move into `STATIC_SLUG_REDIRECTS`; a one-line `redirects` override in `astro.config.mjs` is the cheaper alternative.

Worth folding in while you're there: `buildValidSlugs()` in `sitemap-coverage.test.ts:44-59` doesn't know about `validNonDocsPages` either, so if `VERIFY_LIVE_SITEMAP` is ever enabled in CI, the test meant to police this redirect would report it as broken. One shared set covers both.

Is the client-only fallback an accepted cost for this single legacy URL, or worth the allowlist thread? Genuinely asking — not blocking.

.github/workflows/catalog-editorial-review.yml
```diff
@@ -0,0 +1,418 @@
...
+            // content loader) parses yes/on as strings, which the schema's
+            // z.boolean() rejects at the build gate — only the true
+            // spellings can reach main.
+            function editorialState(text) {
+              const featured = /['"]?featured['"]?\s*:\s*['"]?(true|True|TRUE)\b/.test(text);
+              const badges =
+                // Flow / JSON-ish non-empty list: badges: [verified]
+                /['"]?badges['"]?\s*:\s*\[\s*[^\s\]]/.test(text) ||
+                // Block-style list: a badges: line with at least one item
+                /['"]?badges['"]?\s*:\s*(#[^\n]*)?\r?\n(?:\s*(?:#[^\n]*)?\r?\n)*\s*-\s*\S/.test(text);
```
strandly-the-agent: 🟡 **Catastrophic backtracking in this regex — a 207-byte entry stalls the job.** The inner `\s*` also matches `\n`, overlapping the group's own `\r?\n`, so an N-blank-line run has exponentially many partitions. The input is attacker-controlled: `pull_request_target` fires on any PR touching `site/src/content/catalog/**`, and this runs on raw head text before any validation.

Measured twice, on a payload js-yaml parses fine (the blank-line run sits inside a `description: |` block scalar, so the entry stays schema-valid):

```
blank_lines= 16 bytes= 197 yamlParses=true badgesRegex_ms=25
blank_lines= 20 bytes= 201 yamlParses=true badgesRegex_ms=65
blank_lines= 24 bytes= 205 yamlParses=true badgesRegex_ms=1033
blank_lines= 26 bytes= 207 yamlParses=true badgesRegex_ms=4145
blank_lines= 32 bytes= 195 — still spinning when a 30s cap fired (exit=124)
```

The job then never reaches the label/comment step, so the check silently doesn't run while burning runner time (default job timeout is 6h). The parse-based `editorialState()` above removes this surface entirely. If you keep a regex as a fallback, restricting the inner quantifier to horizontal space — `(?:[ \t]*(?:#[^\n]*)?\r?\n)*` — removes the ambiguity.

.github/workflows/catalog-editorial-review.yml
```diff
@@ -0,0 +1,418 @@
...
+            function extractEntryData(text) {
+              const ghMatch = /^github:\s*(\S+)/m.exec(text);
+              const github = ghMatch ? unquote(ghMatch[1]) : undefined;
+              const pkgs = [];
+              let lang;
+              for (const line of text.split('\n')) {
+                const langMatch = /^\s+(python|typescript):/.exec(line);
+                if (langMatch) lang = langMatch[1];
+                const pkgMatch = /^\s+package:\s*(\S+)/.exec(line);
+                if (pkgMatch && lang) pkgs.push({ lang, pkg: unquote(pkgMatch[1]) });
```
strandly-the-agent: 🟡 **Quoting a key — a YAML no-op — makes the provenance check silently skip.** These line regexes require bare `python:`/`typescript:`/`package:` keys, so a quoted variant that the site parses identically yields no packages. Reproduced twice:

```
CONTROL plain keys   | site uses package=requests | workflow sees=[{"lang":"python","pkg":"requests"}]
"python":  key       | site uses package=requests | workflow sees=[]  <== PROVENANCE SKIPPED
"package": key       | site uses package=requests | workflow sees=[]  <== PROVENANCE SKIPPED
```

With `changedPackages` empty the run takes the early return at line 224 and clears any existing flag, so a non-maintainer can claim someone else's package and the PR stays silent. Reading `data.languages?.python?.package` / `.typescript?.package` and `data.github` off the parsed object fixes this, and makes the flow-style fallback at line 191 unnecessary.

site/src/content/docs/community/get-featured.mdx
```diff
@@ -1,147 +1,87 @@
...
-| `community` | ✅ | Must be `true` |
-| `integrationType` | ✅ | `tool`, `model-provider`, `session-manager`, `memory-store`, `storage`, `plugin`, `integration`, `agent-extension`, or `intervention` |
-| `description` | ✅ | Shown in the catalog table |
-| `languages` | ✅ | `Python`, `TypeScript`, or omit for both |
-| `project` | Recommended | PyPI/npm link, GitHub repo, maintainer |
-| `service` | Optional | External service metadata |
+| `name` | ✅ | Display name in the catalog |
+| `description` | ✅ | One sentence, ~160 characters max — the card clamps longer text |
+| `integrationType` | ✅ | One of the nine types above |
+| `languages` | ✅ | At least one of `python` / `typescript`, each with the `package` name as published on PyPI / npm. For a vendor-guide integration with no Strands-specific package, leave the block empty (`python: {}`) and set `docsUrl` to your guide. |
```
strandly-the-agent: ⚪ **Three small docs items from the rename, bundled.**

1. This row never documents the extras-qualified form, though the schema supports it and 6 shipped entries use it (`temporal`, `agent-control`, `agentcore-memory`, `agentcore-payments`, `agentcore-tool-search`, `headroom`) — a submitter shipping via an extra can't discover the syntax:

```suggestion
| `languages` | ✅ | At least one of `python` / `typescript`, each with the `package` name as published on PyPI / npm (use `pkg[extra]` for an extras-qualified install, e.g. `temporalio[strands-agents]`). For a vendor-guide integration with no Strands-specific package, leave the block empty (`python: {}`) and set `docsUrl` to your guide. |
```

2. Three headings still say “catalog” while the prose beneath them was updated to “the integrations page”: `site/src/content/docs/community/interventions/overview.mdx:28`, `site/src/content/docs/community/memory-stores/overview.mdx:27`, `site/src/content/docs/community/storage/overview.mdx:28` — all `## Browse the catalog`.

3. Line 32 of this file promises “the **Catalog submission** template”, but the template's own heading is now `## Integration submission` (`.github/PULL_REQUEST_TEMPLATE/catalog-submission.md:1`). Nothing breaks — the picker label is filename-derived — but a submitter sees a different name than the guide named.

.github/workflows/catalog-editorial-review.yml
```diff
@@ -0,0 +1,418 @@
...
+            // Both directions are editorial: adding a grant a submitter
+            // isn't entitled to, and removing one a maintainer made — so an
+            // entry is flagged whenever its head state differs from its
+            // base state (a brand-new file compares against empty).
+            const flagged = [];
+            const changedPackages = [];
+            for (const file of catalogFiles) {
+              const [headText, baseText] = await Promise.all([
+                fileText(file.filename, pr.head.sha),
+                fileText(file.filename, pr.base.sha),
```
strandly-the-agent: 🟡 **A rename that drops a maintainer-granted `featured` is never flagged.** The base text is read at the *new* path, and `file.previous_filename` isn't used anywhere in this file. For a rename the new path doesn't exist at base, so `baseText = ''` → `base.featured = false`, and removing a grant compares `false` vs `false`. That's precisely the direction the comment at lines 200-203 says the check covers (“removing one a maintainer made”).

```suggestion
                fileText(file.previous_filename ?? file.filename, pr.base.sha),
```

strandly-the-agent (changes requested):
🔴 **Changes requested — but round 1 landed cleanly:** all 7 prior findings are genuinely fixed. What's left is one more pass on the editorial script, whose detection is still regex-on-raw-YAML rather than parse-on-parsed-YAML.

✅ Re-reviewed `6bd78060` vs `fa63da5d` · 34 files / 516 tests passed (2 skipped, opt-in live-sitemap) · typecheck + diff-check clean.
✅ Verified each round-1 fix at this head, not just the "Fixed" replies: full-content editorial comparison, `concurrency` + `headStillCurrent()`, `assertStat`, `.strict()` languages, `effectiveLanguage()`, Coralogix `docsUrl` now HTTP 200, guide-only PR-template branch.
🔴 Moving detection off patch text was the right fix and closed the round-1 bypass — but 6 YAML forms that js-yaml 4.3.0 (Astro's own loader) parses *identically to the flagged control* still evade `editorialState()`, and an evading push **clears** the prior flag as "~~Resolved~~". Reproduced twice, plus a ReDoS in the same regex (207-byte schema-valid entry → ~4s; 32 blank lines → >30s). One parse-based change fixes comments 1-3.

<details>
<summary>Round-2 breakdown — what I checked, what held, what I dropped</summary>

**Held under attack (negative evidence worth having):** `featured: yes`, `%YAML 1.1`, multi-document YAML, duplicate keys and `!!bool` with a next-line value are all rejected by the schema, so the file's comment at lines 160-163 is correct in effect. The round-1 whole-document flow/JSON bypass is now caught. `headStillCurrent()` demonstrably suppresses mutations when the head moves. `assertStat` took the keep-previous path for all 11 malformed-but-2xx shapes plus a fetcher returning `undefined`. `effectiveLanguage()` makes radio, nav links, and tabs agree on every page shape that exists today (multi-language with either stored preference, Python-only with TypeScript stored, TypeScript-only with Python stored, desktop + mobile double render). No dangling `/catalog` URL references anywhere in `site/src`, `site/test`, or `.github`; both old URLs resolve for a JS-enabled visitor.

**Confirmed good, no action:** `storage` added to both the catalog and docs-frontmatter enums; `generateDocsId` → `pathToDocsSlug` is behaviour-preserving and pinned by `redirect-static.test.ts:128-141`; the `liveFetchers` unexport has no dangling importers; `strands-redis-session-manager.yaml` has no `docsPage` by design and the catalog↔docs integrity test only iterates entries that declare one, so it's honestly scoped.

**Test-side follow-ups (not posted inline):** the 418-line `github-script` block still has no executable test, and this round's rewrite is exactly where the remaining defects live — extracting it to `.github/scripts/catalog-editorial-review.mjs` (precedent: `group-release-notes.mjs`) and driving it from vitest with a fake `github` client would make the parse-based fix verifiable; a starter table is flow↔block reformat must NOT flag, a grant inside a flow map MUST flag, whitespace-only must not flag, brand-new entry still extracts packages, `headStillCurrent()` false ⇒ no mutations / true ⇒ mutations proceed. Separately, `effectiveLanguage()`/`syncAll()` have no coverage and the repo has no DOM test env, so a future edit passing raw `lang` instead of `effective` would re-introduce the round-1 bug with every test green.

**Dropped as unreachable or not actionable today:** `pulls.listFiles` 3000-file cap (~76 entries); stats keys derived from filenames vs `githubSlug` collection ids (flat kebab-case filenames today); the `tab.click()` ↔ Starlight shared-storage-key overlap and the zero-checked-radiogroup edge (no page combines the trigger conditions); `liveFetchers` testability; the one-directional enum-sync test (both copies are in sync today); dual rail/dialog chip `aria-pressed` desync (no repro). Also suppressed as already-raised: fetch timeouts, enum drift, PyPI encoding, the obsolete CodeQL alert on the deleted `backfill.ts`, and the maintainer early-return stale-flag edge.

**Questions (non-blocking):** (1) `pull_request.base.sha` isn't refreshed when `main` advances or on a non-force "Update branch", so a grant merged separately can be misattributed to an unrelated open PR — safe direction (false positive), but worth a line in the workflow's comment block? (2) Given the extras-qualified form now needs documenting, is the template's "Published to PyPI and/or npm under the exact `package` name" also due a tweak — it's literally imprecise there (`temporalio[strands-agents]` is published as `temporalio`; `catalog.ts:53` strips the bracket).

**Not a code finding:** the PR body still says "73 community integrations" (76 YAML entries at this head) and that `community-packages` redirects to `/catalog` (now `/integrations`).

</details>

Still a draft and the remaining 🔴 is in an advisory-only workflow that never gates the merge — so this is one focused fix away, not a rewrite. Worth a human's read on the redirect-stub call in comment 5.

.github/workflows/catalog-editorial-review.yml
yonib05: Fixed.

site/src/util/redirect.ts
yonib05: Fixed.

.github/workflows/catalog-editorial-review.yml
yonib05: Fixed.

.github/workflows/catalog-editorial-review.yml
yonib05: Fixed.

site/src/content/docs/community/get-featured.mdx
yonib05: Fixed.

.github/workflows/catalog-editorial-review.yml
yonib05: Fixed.

arielnabavian (dismissed):

yonib05:

JackYPCOnline (approved):
reviewed carefully, nice PR ^^

yonib05:
> reviewed carefully, nice PR ^^

Yes, sorry for the size; there is a lot to be done from renaming and redirects to backfilling, and new docs and processes.

Waiting for Strandly to do a final pass and will merge

PR #3423, opened 2026-07-23
fix(models/llamacpp): send llama.cpp sampler params at the top level, not under extra_body

### LINKED ISSUE #3422: [BUG] llama.cpp sampling params are nested under extra_body and silently dropped over raw httpx
### SDK Language

Python

### Strands Version

`main` @ 81bccd84

### Language Runtime Version

Python 3.12 (clean Docker container)

### Operating System

Linux

### Bug

`LlamaCppModel._format_request` collects the llama.cpp-specific sampling parameters (`top_k`, `repeat_penalty`, `min_p`, `typical_p`, `tfs_z`, `top_a`, `mirostat*`, `penalty_last_n`, `n_probs`, `min_keep`, `ignore_eos`, `logit_bias`, `cache_prompt`, `slot_id`, `samplers`) into a nested `extra_body` object:

```python
# Add extra_body if we have llama.cpp-specific parameters
if extra_body:
    request["extra_body"] = extra_body
```

`extra_body` is an OpenAI **Python SDK** client-side convention: the SDK merges its contents into the top level of the request body before the request is sent. This provider does not use the OpenAI SDK. It posts the request dict verbatim over raw httpx:

```python
response = await self.client.post("/v1/chat/completions", json=request)  # stream()
```

So these parameters are transmitted nested under a literal `"extra_body"` object rather than as top-level request fields. Every other request-level field the method sets is placed at the top level: the standard OpenAI params (`temperature`, `top_p`, `stop`) and `grammar`/`json_schema` (also llama.cpp-specific). The sampler params are the only ones nested under `extra_body`, so they are not delivered in the same top-level form as the fields the server is handed directly.

### Reproduction

Against `main` @ 81bccd84 in a clean Python 3.12 Docker container:

```python
from strands.models.llamacpp import LlamaCppModel

model = LlamaCppModel(model_id="test", params={
    "temperature": 0.7,     # OpenAI-standard -> top level
    "top_k": 40,            # llama.cpp sampler
    "repeat_penalty": 1.1,  # llama.cpp sampler
    "mirostat": 2,          # llama.cpp sampler
})
req = model._format_request(messages=[{"role": "user", "content": [{"text": "hi"}]}])
print("temperature top-level:", "temperature" in req)   # True
print("top_k top-level:", "top_k" in req)               # False
print("extra_body:", req.get("extra_body"))
```

Output:

```
temperature top-level: True
top_k top-level: False
extra_body: {'top_k': 40, 'repeat_penalty': 1.1, 'mirostat': 2}
```

Because the request is posted verbatim over httpx, `top_k`, `repeat_penalty`, and `mirostat` are sent nested under `extra_body` rather than at the top level, unlike `temperature`.

ebarkhordar:
## Description

`LlamaCppModel._format_request` collected the llama.cpp-specific sampling parameters (`top_k`, `repeat_penalty`, `min_p`, `typical_p`, `tfs_z`, `top_a`, `mirostat*`, `penalty_last_n`, `n_probs`, `min_keep`, `ignore_eos`, `logit_bias`, `cache_prompt`, `slot_id`, `samplers`) into a nested `extra_body` object:

```python
if extra_body:
    request["extra_body"] = extra_body
```

**Root cause:** `extra_body` is an OpenAI Python SDK client-side convention. The SDK merges its contents into the top level of the request body before sending. This provider does not use the OpenAI SDK; it posts the request dict verbatim over raw httpx (`stream`, `count_tokens`):

```python
response = await self.client.post("/v1/chat/completions", json=request)
```

So the parameters are transmitted nested under a literal `"extra_body"` object rather than as top-level request fields. Every other request-level field this method sets is placed at the top level: the standard OpenAI params (`temperature`, `top_p`, `stop`) and, notably, `grammar`/`json_schema`, which are also llama.cpp-specific. The sampler params are the only ones nested under `extra_body`, so they do not arrive in the same top-level form as the fields the server is handed directly.

**Invariant:** llama.cpp-specific sampling params must reach the top level of the request body, the same place `grammar`/`json_schema` and the standard OpenAI params already go.

**Fix:** write the sampler params directly into `request` instead of nesting them under `extra_body`. This is exactly the wire format the OpenAI SDK's `extra_body` would have produced, so it restores the parameter delivery the code intended.

## Related Issues

Fixes #3422

## Type of Change

Bug fix

## Testing

Verified in a clean `python:3.12-slim` Docker container against `main` @ `81bccd84`, with `strands.models.llamacpp.__file__` printed on both sides to confirm provenance:

- Repro on unmodified `main`: `_format_request` with `{top_k, repeat_penalty, min_p, mirostat}` returns them nested under `request["extra_body"]`, while `temperature` lands at the top level.
- New regression test `test_format_request_llamacpp_params_are_top_level_not_extra_body` and the two updated tests fail on `main` (the params are under `extra_body`) and pass on this branch.
- `tests/strands/models/test_llamacpp.py`: 38 passed.
- `ruff check`, `ruff format --check`, and `mypy` on the changed source: clean.

I did not run a live llama.cpp server, so the server-side effect is not directly observed here: the verification covers the request the provider builds, not the server's handling of it. The fix restores the top-level wire format that this method already treats as correct for every other request field (and that the OpenAI SDK's `extra_body` produces after flattening).

I also did not run the full `hatch test --all` suite: the sandbox is missing optional provider/system dependencies (unrelated modules fail at import collection, identically on `main`). The change is confined to one pure method (`_format_request`); no other module imports it.

## Checklist
- [ ] I have updated the documentation accordingly (no documented behavior changes; the parameters were already advertised as supported)

### DIFF
--- a/strands-py/src/strands/models/llamacpp.py
+++ b/strands-py/src/strands/models/llamacpp.py
@@ -389,9 +389,9 @@ class LlamaCppModel(Model):
             if "json_schema" in params:
                 request["json_schema"] = params["json_schema"]
 
-            # llama.cpp-specific parameters that must be passed via extra_body
-            # NOTE: grammar and json_schema are NOT in this set because llama.cpp server
-            # expects them directly in the request body for proper constraint application
+            # llama.cpp-specific sampling parameters. The llama.cpp server reads these
+            # from the top level of the request body, which is also where grammar and
+            # json_schema (handled above) are placed.
             llamacpp_specific_params = {
                 "repeat_penalty",
                 "top_k",
@@ -432,15 +432,15 @@ class LlamaCppModel(Model):
                 if param in openai_params:
                     request[param] = value
 
-            # Collect llama.cpp-specific parameters for extra_body
-            extra_body: dict[str, Any] = {}
+            # Add llama.cpp-specific parameters directly to the request body. This
+            # provider posts the request over raw httpx rather than the OpenAI SDK, so
+            # there is no client-side layer to flatten an "extra_body" object into the
+            # request; nesting these under "extra_body" would leave them unread by the
+            # server. Placing them at the top level matches how the SDK's extra_body is
+            # ultimately delivered, and how grammar/json_schema are already handled.
             for param, value in params.items():
                 if param in llamacpp_specific_params:
-                    extra_body[param] = value
-
-            # Add extra_body if we have llama.cpp-specific parameters
-            if extra_body:
-                request["extra_body"] = extra_body
+                    request[param] = value
 
         return request
 
--- a/strands-py/src/strands/models/llamacpp.py
+++ b/strands-py/src/strands/models/llamacpp.py
@@ -432,12 +432,7 @@ class LlamaCppModel(Model):
                 if param in openai_params:
                     request[param] = value
 
-            # Add llama.cpp-specific parameters directly to the request body. This
-            # provider posts the request over raw httpx rather than the OpenAI SDK, so
-            # there is no client-side layer to flatten an "extra_body" object into the
-            # request; nesting these under "extra_body" would leave them unread by the
-            # server. Placing them at the top level matches how the SDK's extra_body is
-            # ultimately delivered, and how grammar/json_schema are already handled.
+            # Add llama.cpp-specific parameters directly to the request body
             for param, value in params.items():
                 if param in llamacpp_specific_params:
                     request[param] = value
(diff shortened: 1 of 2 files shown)


opieter-aws (changes requested):
Thanks for the fix! One nit before merge: the big comment block above the params loop narrates how the code changed and duplicates your PR description. That's against our evergreen-comment convention (see `AGENTS.md`). The comment on the param set just above already states the only non-obvious why.

Same for the new test's docstring: trim it to the behavior it guarantees with a reference to the issue #3422.

Everything else looks good!

ebarkhordar:
Done. Removed the comment block above the loop, since the set comment already gives the top-level-body reason, and trimmed the regression test docstring to the behavior it guarantees with a reference to #3422. Ruff and the llama.cpp tests pass locally.

opieter-aws (approved):

PR #3424, opened 2026-07-23
fix(bidi): gemini live mp to use updated api

mehtarac:
## Description

Fixes BidiGeminiLiveModel incompatibility with Gemini 3.1 Flash Live models by aligning the implementation with the current Gemini Live API specification. 

  Changes

  - Text input: Use send_realtime_input(text=...) for mid-session text. `send_client_content` is now reserved for history seeding at session start only, per the Gemini 3.1 API contract.
  - Session resumption: Use genai_types.SessionResumptionConfig(handle=...) instead of a raw dict. The SDK type handles None correctly (starts new resumable session) without sending an invalid null value on the wire.
  - API version: Remove incorrect v1alpha default. The google-genai SDK uses the correct version. Users needing v1alpha for 2.5-specific features (affective dialog, proactive audio) can still pass it via client_config.
  - History seeding: Add HistoryConfig(initial_history_in_client_content=True) when messages are present and send all history turns with turn_complete=True. Required by Gemini 3.1 before the server accepts realtime input.

## Related Issues

https://github.com/strands-agents/harness-sdk/issues/1999

## Type of Change

Bug fix
New feature
Breaking change
Documentation update
Other (please describe):

## Testing

## Checklist

----


mehtarac:
@strandly-the-agent review the implementation

strands-py/src/strands/experimental/bidi/models/gemini_live.py
```diff
@@ -491,7 +494,13 @@ def _build_live_config(
...
         config_dict: dict[str, Any] = self.config["inference"].copy()
 
-        config_dict["session_resumption"] = {"handle": kwargs.get("live_session_handle")}
+        live_session_handle = kwargs.get("live_session_handle")
+        config_dict["session_resumption"] = genai_types.SessionResumptionConfig(handle=live_session_handle)
+
+        # Gemini 3.1 requires this flag to allow send_client_content for history seeding
+        has_messages = kwargs.get("has_messages", False)
+        if has_messages:
+            config_dict["history_config"] = genai_types.HistoryConfig(initial_history_in_client_content=True)
```
strandly-the-agent, line 511: 🔴 **This breaks two currently supported configurations before a request is sent.**

* Both `gemini` extras still allow `google-genai>=1.32.0`, but `HistoryConfig` does not exist until 1.67. At this head, 1.32 raises `AttributeError` here whenever history is present (reproduced twice).
* For a Vertex client on google-genai 1.69, the real live-config converter explicitly rejects this field with `ValueError("history_config parameter is not supported in Vertex AI.")` (also reproduced twice).

Please raise both dependency floors to at least 1.67 and avoid emitting this unsupported field for Vertex. Once the floor is updated, the backend gate here can be applied directly:

```suggestion
            if not self._client.vertexai:
                config_dict["history_config"] = genai_types.HistoryConfig(initial_history_in_client_content=True)
```

strands-py/src/strands/experimental/bidi/models/gemini_live.py
```diff
@@ -153,7 +152,8 @@ async def start(
         self._connection_id = str(uuid.uuid4())
 
         # Build live config
-        live_config = self._build_live_config(system_prompt, tools, **kwargs)
+        has_messages = bool(messages) and "live_session_handle" not in kwargs
+        live_config = self._build_live_config(system_prompt, tools, has_messages=has_messages, **kwargs)
```
strandly-the-agent, line 162: 🔴 **Deriving this from the raw list can leave the connection stuck in initial-history mode.** `_send_message_history()` later drops every non-text block and only sends `turn_complete=True` when a text turn survives. With valid tool-only `toolUse`/`toolResult` history, this flag is therefore enabled but zero client-content turns are sent; I reproduced that path twice. The provider waits for terminating initial client content before entering realtime mode.

This predicate mirrors the method's current text-only conversion, so unsupported-only histories skip initial-history mode instead of waiting forever:

```suggestion
        has_messages = (
            messages is not None
            and any("text" in block for message in messages for block in message["content"])
            and "live_session_handle" not in kwargs
        )
```

Longer term, converting tool history rather than dropping it would preserve more context.

strandly-the-agent (changes requested):
**Changes requested:** two compatibility failures and one history-initialization failure reproduce at this exact head.

✅ Reviewed `bc7ac6a36579b176427aaa59c755165b1ee15bb6` against merge-base `81bccd8420c255b7ea4e23ad67f1fcf099d405e9`  
✅ Focused Gemini Live tests: 18 passed on google-genai 1.69; full GitHub CI is green  
🔴 Cold repros repeated twice for the dependency floor, Vertex conversion, and tool-only history; `git diff --check` and format checks passed

<details>
<summary>Per-pass results and aggregation</summary>

### Correctness / safety
**Changes requested.** `HistoryConfig` at `gemini_live.py:503` is absent from the declared google-genai 1.32–1.66 range, and google-genai 1.69's Vertex converter rejects the field. Independently, `has_messages` at line 155 enters initial-history mode for tool-only histories that `_send_message_history()` filters to zero turns.

The `send_realtime_input(text=...)` update matches issue #1999 and its SDK signature exists at the dependency floor; no correctness defect was found in that part.

### Test quality
**Gaps found.** The focused file passes all 18 tests on google-genai 1.69, but its mocks bypass the SDK's version/backend conversion. It therefore cannot catch the minimum-version `AttributeError`, Vertex `ValueError`, or all-filtered history path. I folded those gaps into the concrete defect comments rather than posting a separate generic coverage comment.

### Adversarial / repro
**Broke it in three paths, each repeated twice:**

- google-genai 1.32: `_build_live_config(has_messages=True)` raises `AttributeError: ... HistoryConfig`.
- google-genai 1.69 + Vertex: `start(messages=[text])` raises `ValueError: history_config parameter is not supported in Vertex AI` before network access.
- Tool-only `toolUse`/`toolResult` history: initial-history mode is enabled, but `send_client_content(..., turn_complete=True)` is called zero times.

### Aggregation / suppression
- Combined the dependency-floor and Vertex failures at their shared changed line; retained the independent history-termination defect separately.
- Suppressed session resumption as a defect: on google-genai 1.32 and 1.69, both the old `{"handle": None}` and new typed value serialize to the same intentional empty `sessionResumption`, used to request handles.
- Suppressed a generic missing-tests comment as duplicative, and a Ruff import finding as pre-existing/outside the diff.
- No public Python signature/export changed, so the API-review label workflow does not apply.
</details>

strands-py/src/strands/experimental/bidi/models/gemini_live.py
mehtarac: Fixed. has_messages now checks for actual text content

strands-py/src/strands/experimental/bidi/models/gemini_live.py
mehtarac: Will bump the dependency floor in a separate PR

pgrayy (approved):

PR #3365, opened 2026-07-21
feat(ts): add AgentCore Harness agent

gautamsirdeshmukh:
## Description

We introduce `AgentCoreHarnessAgent`, a new `InvokableAgent` implementation allowing developers to interface with and integrate a [AgentCore Harness agent](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/harness.html) into their agentic applications just as easily as a typical Strands agent today.

```bash
npm install @strands-agents/sdk @aws-sdk/client-bedrock-agentcore
```

```typescript
import { AgentCoreHarnessAgent } from '@strands-agents/sdk/agentcore-harness'

const agent = new AgentCoreHarnessAgent({
  harnessArn: '...',
  runtimeSessionId: '...',
})

const result = await agent.invoke("Tell me about your local environment")
```

```mermaid
sequenceDiagram
    participant Client as Strands client
    participant Agent as AgentCoreHarnessAgent
    participant API as InvokeHarness API
    participant Harness as Deployed Harness

    Client->>Agent: invoke(input) or stream(input)
    Agent->>API: InvokeHarness(ARN, session ID, messages)
    API->>Harness: Run agent

    Note over Harness: Owns the model, prompt, tools, memory, limits, and agent loop

    Harness-->>API: Stream events and terminal result
    API-->>Agent: InvokeHarness event stream
    Agent-->>Client: AgentResult or streamed events
```

### Scope and roadmap

This PR establishes an intentionally narrow V0. Each subsequent version shifts a deliberate part of Harness ownership to the client.

| Version | Scope | Features and support introduced | Experience |
| --- | --- | --- | --- |
| **V0: Remote adapter** | Invoke and stream a deployed Harness through the `InvokableAgent` interface. | Text input, raw event streaming, cancellation, typed error translation, runtime session continuity, and Graph/A2A composition. | The Harness owns its model, prompt, tools, memory, and agent loop. The client supplies the Harness ARN, session ID, and input. |
| **V1: Client-side tool calling** | Pass custom tools to the harness agent to be run client-side, automatically or HITL, return tool results, and continue the Harness turn. | `inline_function` registration, automatic or HITL client-side tool execution, tool-result submission, and `partial_turn` continuation with merged streaming and metrics. | Above + Harness receives custom tools as `inline_functions`. |
[...]

PR #3428, opened 2026-07-23
fix(a2a): filter executor fallback content

gautamsirdeshmukh:
## Description

`A2AExecutor` previously fell back to `AgentResult.toString()` when an agent stream produced no recognized A2A parts, which could expose internal reasoning in the client-visible artifact. 

With this PR, we now use the existing A2A content adapter for the final assistant message so response-facing content is preserved while unsupported internal blocks are excluded. There is no public API change.

Example: an InvokableAgent returns:

```typescript
lastMessage.content = [
  new ReasoningBlock({ text: 'The user may be vulnerable to this offer...' }),
  new TextBlock('I cannot help with that request.'),
]
```

If its stream emits no A2A-recognized text deltas, A2AExecutor previously used AgentResult.toString(). The resulting A2A response could expose both:
```
The user may be vulnerable to this offer...
I cannot help with that request.
```

With the fix, A2A converts only response-facing content, returning:
```
I cannot help with that request.
```

## Related Issues

Discovered while working on #3365

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

- [x] Ran the A2A executor unit suite in Node.js and Chromium (17/17 each)
- [x] Ran the TypeScript build, type-check, lint, and formatting checks

## Checklist

- [x] I have added the necessary regression test
- [x] No documentation changes are needed
- [x] There are no dependent changes

### DIFF
--- a/strands-ts/src/a2a/__tests__/executor.test.ts
+++ b/strands-ts/src/a2a/__tests__/executor.test.ts
@@ -3,12 +3,14 @@ import { A2AExecutor } from '../executor.js'
 import type { AgentExecutionEvent, ExecutionEventBus, RequestContext } from '@a2a-js/sdk/server'
 import type { TaskArtifactUpdateEvent, TaskStatusUpdateEvent } from '@a2a-js/sdk'
 import { Agent } from '../../agent/agent.js'
+import { Interrupt } from '../../interrupt.js'
+import { CitationsBlock } from '../../types/citations.js'
 import type { InvokableAgent } from '../../types/agent.js'
 import { MockMessageModel } from '../../__fixtures__/mock-message-model.js'
 import { createMockAgent } from '../../__fixtures__/agent-helpers.js'
-import { TextBlock } from '../../types/messages.js'
+import { ReasoningBlock, TextBlock } from '../../types/messages.js'
 import { ImageBlock, encodeBase64 } from '../../types/media.js'
-import { ContentBlockEvent, ModelStreamUpdateEvent } from '../../hooks/events.js'
+import { ContentBlockEvent, ModelStreamUpdateEvent, StreamEvent } from '../../hooks/events.js'
 import { AgentResult } from '../../types/agent.js'
 import { Message } from '../../types/messages.js'
 import type { ModelStreamEvent } from '../../models/streaming.js'
@@ -37,6 +39,25 @@ function createRequestContext(text: string, taskId: string = 'task-1'): RequestC
       messageId: 'msg-1',
       role: 'user',
       parts: [{ kind: 'text', text }],
+    },
+  }
+}
+
+class CustomStreamUpdateEvent extends StreamEvent {}
+
+class CustomRenderedAgentResult extends AgentResult {
+  public override toString(): string {
+    return 'SAFE_REDACTED_RESPONSE'
+  }
+}
+
+function createCustomAgent(result: AgentResult): InvokableAgent {
+  return {
+    id: 'custom-agent',
+    invoke: vi.fn(),
+    async *stream() {
+      yield new CustomStreamUpdateEvent()
+      return result
     },
   }
 }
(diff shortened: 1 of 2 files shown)


strands-ts/src/a2a/__tests__/executor.test.ts
```diff
@@ -110,6 +112,50 @@ describe('A2AExecutor', () => {
       expect(artifactEvents[2]!.lastChunk).toBe(true)
     })
 
+    it('publishes only response-facing content when no text deltas are streamed', async () => {
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Added a reference to #3365 describing the behavior this regression test protects

strands-ts/src/a2a/executor.ts
```diff
@@ -290,21 +292,24 @@ export class A2AExecutor implements AgentExecutor {
...
               lastChunk: true,
             })
+            hasStreamedOutput = true
           }
         }
 
         next = await stream.next()
       }
 
+      const resultParts = !hasStreamedOutput && next.value ? contentBlocksToParts(next.value.lastMessage.content) : []
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Good catch, the initial implementation unintentionally changed more than the reasoning behavior. Narrowed the fix to preserve the existing AgentResult.toString() fallback semantics for interrupts, structured output, text, and citation text, while excluding only reasoning blocks. Also added coverage for the preserved fallback cases.

gautamsirdeshmukh:
@strandly-the-agent have a look

strands-ts/src/a2a/executor.ts
```diff
@@ -63,6 +63,28 @@ function hasSessionManager(agent: InvokableAgent): boolean {
...
+      }
+    }
+  }
+
+  return textParts.join('\n')
+}
+
 /**
  * Bridges a Strands Agent into the A2A protocol as an AgentExecutor.
  *
```
strandly-the-agent, line 99: **Should-fix (non-blocking): preserve custom rendering when reasoning filtering is unnecessary**

This reconstruction bypasses polymorphic `AgentResult.toString()` for ordinary results. A custom result containing `SENSITIVE_RAW_TEXT` but overriding `toString()` to return `SAFE_REDACTED_RESPONSE` returns the safe value on base and raw text on this head; I cold-reproduced that base/head difference twice.

Could this delegate to `result.toString()` when there are no reasoning blocks? When reasoning is present, it should avoid any overridable renderer: preserve interrupt/structured-output priority and construct the response only from allowed text/citation blocks. That keeps custom rendering compatible without reopening the reasoning leak; the expanded variant passed 23/23 focused tests.

strandly-the-agent (approved):
**Verdict: APPROVE — the reasoning leak is fixed; one non-blocking `AgentResult.toString()` compatibility regression remains.**

**Verified**
- Exact head `b15e764` against base `248896f`.
- Focused executor suite: 20/20 passed; type-check, ESLint, Prettier, and diff check passed locally.
- Exact-head Node 20/22/24 OS matrix, quality, security, packaging, docs, integration, and CI Gate passed.
- Local browser launch was unavailable because this sandbox lacks Playwright shared libraries; the exact-head GitHub matrix covered it.

<details>
<summary>Pass breakdown</summary>

- Correctness: no findings.
- Test quality: the custom event exercises the actual fallback; whole-event assertions prove reasoning exclusion and preserved structured output, interrupts, and citations.
- LLM/client context: no loss, duplication, or contamination found for ordinary results.
- Adversarial: the advanced tier was unavailable under the account's data-retention mode, so I retried on the default tier. It cold-reproduced the inline compatibility issue twice; mixed ordering, reasoning-only, empty/falsy values, interrupt priority, citations, media, and duplicate-emission paths introduced no other regressions.
- Independent validation reproduced the issue twice and validated a reasoning-aware fix direction with 23/23 focused tests.
- The earlier structured-output/interrupt/citation and test-reference threads are resolved and were not repeated.
- No public API signature/export changed; the API-label check passed.

</details>

gautamsirdeshmukh:
@strandly-the-agent take another look

strands-ts/src/a2a/executor.ts
```diff
@@ -63,6 +63,37 @@ function hasSessionManager(agent: InvokableAgent): boolean {
   return (agent as { sessionManager?: unknown }).sessionManager !== undefined
 }
 
+/** Returns client-facing result text while excluding internal reasoning content. */
+function resultToResponseText(result: AgentResult): string {
+  const hasReasoning = result.lastMessage.content.some((block) => block.type === 'reasoningBlock')
+  if (!hasReasoning) {
```
strandly-the-agent, line 69: **Should-fix: `result.lastMessage.content` is now dereferenced before the `interrupts`/`structuredOutput` branches**

(Anchored a line low — this is about the `const hasReasoning = result.lastMessage.content.some(...)` read just above, `executor.ts:68`.)

At `b15e764` those two checks ran first and never touched `lastMessage`. Now a result object that carries `interrupts` but no `lastMessage` throws `TypeError: Cannot read properties of undefined (reading 'content')`, so `execute()` rejects and publishes no artifact where it previously published the interrupt rendering. I cold-verified both sides: head throws with `texts=[]`, `b15e764` published `INTERRUPT_RENDERED`.

Non-blocking because `AgentResult.lastMessage` is non-optional, so this needs a cast or plain JS to reach — if you consider duck-typed results a supported input, treat it as blocking instead. Hoisting the read once also removes the second `result.lastMessage` read in the loop below (worth doing regardless: check-and-use are otherwise not the same read).

```ts
// executor.ts:68 — replaces the single `const hasReasoning = ...` line
const content = result.lastMessage?.content ?? []
const hasReasoning = content.some((block) => block.type === 'reasoningBlock')
// ...and `for (const block of content)` in the projection loop below
```

With that applied I got 21/21 tests, `tsc --noEmit` clean, and ESLint clean (no `no-unnecessary-condition` complaint).

Separately, a nit while you are in this function: one line in the doc comment saying a custom `toString()` is *deliberately* not consulted once reasoning is present would stop a future refactor from quietly reverting it — a caller's renderer cannot be trusted to strip reasoning.

strands-ts/src/a2a/__tests__/executor.test.ts
```diff
@@ -110,6 +131,127 @@ describe('A2AExecutor', () => {
...
+          taskId: 'task-1',
+          contextId: 'ctx-1',
+          status: { state: 'completed' },
+          final: true,
+        },
+      ])
+    })
+
+    it.each([
+      {
```
strandly-the-agent, line 175: **Should-fix: no case pins "no reasoning + `structuredOutput`/`interrupts` + custom `toString()`"**

Every case in this table now carries a `ReasoningBlock`, so nothing exercises the delegation path for a result that has structured output or interrupts but no reasoning. I verified the gap by mutating the implementation to test `interrupts`/`structuredOutput` *before* `hasReasoning` — a plausible simplification that silently drops custom-renderer precedence — and all 21 tests stayed green. Adding this case gives 22/22 at head and fails exactly 1 test under that mutation:

```ts
{
  name: 'custom result rendering with structured output and no reasoning',
  result: new CustomRenderedAgentResult({
    stopReason: 'endTurn',
    lastMessage: new Message({ role: 'assistant', content: [new TextBlock('SENSITIVE_RAW_TEXT')] }),
    structuredOutput: { answer: 42 },
    invocationState: {},
  }),
  expectedText: 'SAFE_REDACTED_RESPONSE',
},
```

strandly-the-agent (approved):
**Verdict: APPROVE — my round-1 finding is genuinely fixed; two non-blocking should-fixes on the new delta.**

**Verified at head `730bafb0`** (base `0e517d62`, clean tree)
- Focused executor suite 21/21, no type errors; `tsc --noEmit` on `src` clean; ESLint clean; exact-head CI rollup green.
- Cold probe through the real executor: no-reasoning + custom `toString()` now publishes `SAFE_REDACTED_RESPONSE`, so custom rendering is preserved; reasoning-bearing results publish no reasoning.
- Both suggestions below were applied locally and re-run: the guard keeps 21/21 with clean lint/type-check, and the extra test case gives 22/22 while failing (1 test) under the check-reorder mutation that currently passes silently.

<details>
<summary>Pass breakdown</summary>

- Correctness: no findings. Branch order, empty-`interrupts` truthiness, the `reasoningBlock` discriminant, and the interrupts → structuredOutput → text/citations priority all match `AgentResult.toString()` (`strands-ts/src/types/agent.ts:528-565`).
- Adversarial: the leak stays closed end to end, including an A2A round trip into `A2AAgent` where the merge base leaks reasoning on the wire and this head does not. Alternate carriers (wire-format reasoning, reasoning nested in tool results/json blocks, redacted-only reasoning), the `ContentBlockEvent` artifact path, and empty `interrupts: []` all behaved correctly. The one new regression is the inline `lastMessage` finding; a `Proxy`/instance-accessor double-read variant of the same defect disappears with the same fix.
- Test quality: mutation-checked and the two-branch contract is well pinned (inverting `hasReasoning` fails 5, breaking the discriminant fails 4, re-including reasoning fails 2). The one real gap is the inline test comment.
- Pre-existing, not this delta: `JSON.stringify` on a BigInt `structuredOutput` throws, and one whose `toJSON()` returns `undefined` yields a text part with no `text` — same at the merge base because `toString()` stringifies identically.
- Earlier resolved threads (`contentBlocksToParts` drops, test issue reference, custom-`toString()` bypass) were not re-raised.

</details>

Observed, not a finding: a reasoning-only turn now publishes an empty text artifact (`[""]`), which is the intended consequence of filtering and matches Python's A2A behavior.

PR #528, opened 2025-07-24
feat(mcp): retain structured content in the AgentTool response

dbschmigelski:
## Description
This PR implements the retention of structured content in the AgentTool response when using the MCP client. When structured content is available from MCP tools, it will now be included as the last item in the content array of the returned ToolResult.

The changes include updating the MCP dependency to version 1.11.0, modifying the MCPClient class to handle structured content in both synchronous and asynchronous tool calls. The echo server in integration tests has been enhanced to include a new tool that returns structured content for testing purposes.

The implementation ensures backward compatibility while adding the ability to access structured data returned by MCP tools. This enhancement improves the flexibility of the SDK by allowing tools to return both unstructured and structured responses in a consistent format.

The changes have been tested with both unit tests and integration tests to verify proper handling of structured content in various scenarios.

Related to: https://github.com/model-context-protocol/specification/blob/main/specification.md#structured-content

Design Decision Discussion:

We faced two potential approaches for handling structured content:

Option A:
We could have mirrored the MCP Tool Response structure by having separate 'content' and 'structuredContent' fields in our AgentTool response. This would have introduced structured content support uniformly across all tools, not just MCP tools. However, this would require a more significant change to our existing tool response interface.

Option B (Chosen Approach):
We decided to leverage our existing 'json' content type capability, which already provides a mechanism for handling structured data. This approach maintains consistency with our current content type system while still allowing us to represent structured data. 

A key design decision was whether to append the structured content to existing content items or replace them entirely. We chose to append the structured content as the last item in the content array for several reasons:
1. It preserves all information returned by the tool
2. It maintains backward compatibility
3. It allows consumers to access both human-readable and structured representations of the data

While appending was chosen, it's worth noting this wasn't a straightforward decision. Replacing existing content with structured content could have provided a cleaner, more direct approach to structured data handling, may result in existing applications breaking because an expected text block is no longer present. 

## Type of Change
New feature

## Testing

How have you tested the change?  Verify that the changes do not break functionality or introduce warnings in consuming repositories: agents-docs, agents-tools, agents-cli

- [x] I ran `hatch run prepare`

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [no, open to adding if desired] I have updated the documentation accordingly
- [n/a] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

----

[...]

PR #3431, opened 2026-07-23
feat(mcp): surface structuredContent from MCP tool results

### LINKED ISSUE #3270: [FEATURE] TS: surface structuredContent from MCP tool results
### SDK Language

TypeScript

### Problem Statement

The TS client preserves `outputSchema` on the ToolSpec (#3109), but the tool-result mapper reads only `content[]` and `isError` (`strands-ts/src/tools/mcp-tool.ts:55-73`), so `structuredContent` returned by servers is silently dropped. Python plumbs it through as `MCPToolResult["structuredContent"]` (`strands-py/src/strands/tools/mcp/mcp_client.py:927-928`).

### Proposed Solution

Map `result.structuredContent` onto the tool result, mirroring the Python shape, and add unit tests matching the Python coverage (`tests_integ/mcp/test_mcp_client_structured_content_and_metadata.py`).

### Use Case

Typed tool results are what make output machine-consumable: validation against `outputSchema`, structured-output flows, telemetry. Today TS agents lose exactly that data.

### Additional Context

Gotcha: the 2026-07-28 spec RC (SEP-2106) loosens `structuredContent` to any JSON value, not just objects; don't assume an object shape. No blockers; small, self-contained fix.

Parent: #1659

poshinchen:
## Description

  MCP spec revision 2025-06-18 added structured tool output: a server declares `outputSchema` on a tool and returns the typed value as `structuredContent` alongside the human-readable content blocks. 

TypeScript SDK already surfaces `outputSchema` on `ToolSpec`, but the tool-result mapper discarded structuredContent. This PR added `structuredContent` on `ToolResultBlock` and its serialized form, so typed tool results can be validated against `outputSchema` and consumed programmatically.

The field accepts any JSON value, based on [SEP-2106](https://modelcontextprotocol.io/seps/2106-json-schema-2020-12), which shipped in spec revision 2026-07-28.

## Needs API review on the new field [messages.StructuredContent](https://github.com/strands-agents/harness-sdk/pull/3431/changes#diff-aa9e01706a53e921024715e40085d4101c883a5d3d9829bec00f28f0469192f4R415)

One new optional field on three public surfaces, all exported from the package root:
  
1. [`ToolResultBlockData.structuredContent?: JSONValue`](https://github.com/strands-agents/harness-sdk/pull/3431/changes#diff-aa9e01706a53e921024715e40085d4101c883a5d3d9829bec00f28f0469192f4R415)
  (interface). This is the serialized shape, so it is also a wire-format change for persisted sessions.
2. [`ToolResultBlock.structuredContent?: JSONValue`](https://github.com/strands-agents/harness-sdk/pull/3431/changes#diff-aa9e01706a53e921024715e40085d4101c883a5d3d9829bec00f28f0469192f4R416) (readonly class property), plus the widened constructor parameter.
3. The JSON serialization: `toJSON()` now emits an optional `structuredContent` key inside `toolResult`, and `fromJSON()` revives it. This is the session-persistence format, so it carries the most compatibility weight.
  
## Related Issues
Closes #3270
  
## Documentation PR
N/A
  
## Type of Change
New feature
  
## Testing
  
## Checklist
  
----
  
### DIFF
--- a/strands-ts/src/tools/mcp-tool.ts
+++ b/strands-ts/src/tools/mcp-tool.ts
@@ -70,6 +70,10 @@ export class McpTool extends Tool {
         toolUseId,
         status: rawResult.isError ? 'error' : 'success',
         content,
+        // The field accepts any JSON value (MCP SEP-2106), but the current MCP SDK
+        // validates structuredContent as object-only before this mapper runs, so only
+        // objects can arrive from a real server until the SDK implements SEP-2106.
+        ...(rawResult.structuredContent !== undefined && { structuredContent: rawResult.structuredContent }),
       })
     } catch (error) {
       if (
@@ -190,9 +194,11 @@ export class McpTool extends Tool {
 
   /**
    * Type Guard: Checks if value matches the expected MCP SDK result shape.
-   * \{ content: unknown[]; isError?: boolean \}
+   * \{ content: unknown[]; structuredContent?: JSONValue; isError?: boolean \}
    */
-  private _isMcpToolResult(value: unknown): value is { content: unknown[]; isError?: boolean } {
+  private _isMcpToolResult(
+    value: unknown
+  ): value is { content: unknown[]; structuredContent?: JSONValue; isError?: boolean } {
     if (typeof value !== 'object' || value === null) {
       return false
     }
--- a/strands-ts/src/types/messages.ts
+++ b/strands-ts/src/types/messages.ts
@@ -374,6 +374,11 @@ export interface ToolResultBlockData {
   content: ToolResultContentData[]
 
   /**
+   * Optional structured data returned by the tool (e.g. MCP `structuredContent`), any JSON value.
+   */
+  structuredContent?: JSONValue
+
+  /**
    * The original error object when status is 'error'.
    * Available for inspection by hooks, error handlers, and agent loop.
    * Tools must wrap non-Error thrown values into Error objects.
@@ -406,16 +411,30 @@ export class ToolResultBlock implements JSONSerializable<{ toolResult: ToolResul
   readonly content: ToolResultContent[]
 
   /**
+   * Optional structured data returned by the tool (e.g. MCP `structuredContent`), any JSON value.
+   */
+  readonly structuredContent?: JSONValue
+
+  /**
    * The original error object when status is 'error'.
    * Available for inspection by hooks, error handlers, and agent loop.
    * Tools must wrap non-Error thrown values into Error objects.
    */
   readonly error?: Error
 
-  constructor(data: { toolUseId: string; status: 'success' | 'error'; content: ToolResultContent[]; error?: Error }) {
+  constructor(data: {
+    toolUseId: string
+    status: 'success' | 'error'
+    content: ToolResultContent[]
+    structuredContent?: JSONValue
+    error?: Error
+  }) {
     this.toolUseId = data.toolUseId
     this.status = data.status
     this.content = data.content
+    if (data.structuredContent !== undefined) {
+      this.structuredContent = data.structuredContent
+    }
     if (data.error !== undefined) {
       this.error = data.error
     }
@@ -432,6 +451,7 @@ export class ToolResultBlock implements JSONSerializable<{ toolResult: ToolResul
         toolUseId: this.toolUseId,
         status: this.status,
         content: this.content.map((block) => block.toJSON() as ToolResultContentData),
+        ...(this.structuredContent !== undefined && { structuredContent: this.structuredContent }),
       },
     }
   }
@@ -448,6 +468,9 @@ export class ToolResultBlock implements JSONSerializable<{ toolResult: ToolResul
       toolUseId: data.toolResult.toolUseId,
       status: data.toolResult.status,
       content,
+      ...(data.toolResult.structuredContent !== undefined && {
+        structuredContent: data.toolResult.structuredContent,
+      }),
     })
   }
 }
--- a/strands-ts/src/agent/__tests__/agent.hook.test.ts
+++ b/strands-ts/src/agent/__tests__/agent.hook.test.ts
@@ -1513,6 +1513,42 @@ describe('Agent Hooks Integration', () => {
       )
       expect(block!.content).toEqual([new TextBlock('[REDACTED]')])
     })
+
+    it.each([
+      ['object', { temperature: 72 }],
+      ['zero', 0],
+      ['empty string', ''],
+      ['false', false],
+      ['null', null],
+    ])('preserves %s structuredContent when normalizing a mismatched toolUseId', async (_label, structuredContent) => {
+      const tool = createMockTool('tool', () => 'done')
+
+      const model = new MockMessageModel()
+        .addTurn({ type: 'toolUseBlock', name: 'tool', toolUseId: 'tool-1', input: {} })
+        .addTurn({ type: 'textBlock', text: 'Done' })
+
+      const agent = new Agent({ model, tools: [tool] })
+      agent.addHook(AfterToolCallEvent, (event: AfterToolCallEvent) => {
+        // A mismatched toolUseId forces the executor to rebuild the block.
+        event.result = new ToolResultBlock({
+          toolUseId: 'hook-issued-id',
+          status: 'success',
+          content: [new TextBlock('replaced')],
+          structuredContent,
+        })
+      })
+
+      await agent.invoke('Test')
+
+      const toolResultMessage = agent.messages.find((m) =>
+        m.content.some((b) => b.type === 'toolResultBlock' && b.toolUseId === 'tool-1')
+      )
+      expect(toolResultMessage).toBeDefined()
+      const block = toolResultMessage!.content.find(
+        (b): b is ToolResultBlock => b.type === 'toolResultBlock' && b.toolUseId === 'tool-1'
+      )
+      expect(block!.structuredContent).toEqual(structuredContent)
+    })
   })
 
   describe('AfterInvocationEvent resume', () => {
--- a/strands-ts/src/conversation-manager/__tests__/sliding-window-conversation-manager.test.ts
+++ b/strands-ts/src/conversation-manager/__tests__/sliding-window-conversation-manager.test.ts
@@ -451,6 +451,42 @@ describe('SlidingWindowConversationManager', () => {
       )
     })
 
+    it.each([
+      ['object', { temperature: 72 }],
+      ['zero', 0],
+      ['empty string', ''],
+      ['false', false],
+      ['null', null],
+    ])('preserves %s structuredContent on truncated tool results', (_label, structuredContent) => {
+      const manager = new SlidingWindowConversationManager({ shouldTruncateResults: true })
+      const messages = [
+        new Message({
+          role: 'user',
+          content: [
+            new ToolResultBlock({
+              toolUseId: 'tool-1',
+              status: 'success',
+              content: [new TextBlock('x'.repeat(500))],
+              structuredContent,
+            }),
+          ],
+        }),
+      ]
+
+      const changed = (manager as any)._truncateToolResults(messages, 0)
+      expect(changed).toBe(true)
+
+      const expectedText = `${'x'.repeat(200)}\n<truncated chars="100"/>\n${'x'.repeat(200)}`
+      expect(messages[0]!.content[0]).toEqual(
+        new ToolResultBlock({
+          toolUseId: 'tool-1',
+          status: 'success',
+          content: [new TextBlock(expectedText)],
+          structuredContent,
+        })
+      )
+    })
+
     it('image placeholder reflects non-bytes source kinds honestly', () => {
       const manager = new SlidingWindowConversationManager({ shouldTruncateResults: true })
       const messages = [
--- a/strands-ts/src/conversation-manager/sliding-window-conversation-manager.ts
+++ b/strands-ts/src/conversation-manager/sliding-window-conversation-manager.ts
@@ -414,10 +414,14 @@ export class SlidingWindowConversationManager extends ConversationManager {
       }
 
       changesMade = true
+      // structuredContent is never sent to the model, so truncation preserves it.
       return new ToolResultBlock({
         toolUseId: toolResultBlock.toolUseId,
         status: toolResultBlock.status,
         content: newItems,
+        ...(toolResultBlock.structuredContent !== undefined
+          ? { structuredContent: toolResultBlock.structuredContent }
+          : {}),
         ...(toolResultBlock.error !== undefined ? { error: toolResultBlock.error } : {}),
       })
     })
--- a/strands-ts/src/mcp/__tests__/client.structured-content.test.node.ts
+++ b/strands-ts/src/mcp/__tests__/client.structured-content.test.node.ts
@@ -1,0 +1,69 @@
+import { describe, it, expect, beforeEach, afterEach } from 'vitest'
+import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js'
+import { InMemoryTransport } from '@modelcontextprotocol/sdk/inMemory.js'
+import * as z from 'zod/v4'
+import { McpClient } from '../client.js'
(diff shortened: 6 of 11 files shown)


strands-ts/src/types/messages.ts
```diff
@@ -373,6 +373,11 @@ export interface ToolResultBlockData {
    */
   content: ToolResultContentData[]
 
+  /**
+   * Optional structured data returned by the tool (e.g. MCP `structuredContent`), any JSON value.
+   */
+  structuredContent?: JSONValue
```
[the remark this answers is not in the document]
poshinchen: I'll add api review.

mkmeral:
@strandly-the-agent  review pls. also show me API changes breakdown 

fyi: whenever I say API changes/diff breakdown, I want like public API surface (new params, methods, defaults, etc.), failure modes, errors, naming (and alternatives), and whether it's consistent with similar components within the SDK. Remember this

strands-ts/src/tools/mcp-tool.ts
```diff
@@ -70,6 +70,8 @@ export class McpTool extends Tool {
         toolUseId,
         status: rawResult.isError ? 'error' : 'success',
         content,
+        // Any JSON value is allowed, not just objects (MCP SEP-2106)
+        ...(rawResult.structuredContent !== undefined && { structuredContent: rawResult.structuredContent }),
       })
```
strandly-the-agent, line 77: 🔴 **Block — this `JSONValue` support is unreachable for real non-object MCP results.** Both SDK call paths pass the default result schema (`mcp/client.ts:413,418`), and locked `@modelcontextprotocol/sdk` 1.29.0 still requires record-shaped `structuredContent`. The new scalar test mocks `McpClient.callTool`, below that boundary.

A real in-memory MCP `Server` → Strands `McpClient` repro preserved `{ ok: true }`, but `42`, `[1,2]`, `false`, `0`, `""`, and `null` all became error tool results (`Invalid input: expected record...`) before this mapper ran.

Please align the dependency/result schema with SEP-2106 for both direct and task calls, then cover a non-object value at the transport boundary. If that isn't possible yet, the public type/claim needs to remain object-only.

strands-ts/src/types/messages.ts
```diff
@@ -405,17 +410,31 @@ export class ToolResultBlock implements JSONSerializable<{ toolResult: ToolResul
...
   readonly content: ToolResultContent[]
 
+  /**
+   * Optional structured data returned by the tool (e.g. MCP `structuredContent`), any JSON value.
+   */
+  readonly structuredContent?: JSONValue
+
   /**
    * The original error object when status is 'error'.
    * Available for inspection by hooks, error handlers, and agent loop.
```
strandly-the-agent, line 420: 🔴 **Block — existing `ToolResultBlock` reconstruction paths must carry this new data field forward.** Three currently omit it:

- sliding-window truncation: `sliding-window-conversation-manager.ts:417-422`
- ContextOffloader replacement: `context-offloader/plugin.ts:495-499`
- executor ID normalization: `executor.ts:161-166`

Cold repros with an object, `false`, `0`, `""`, and `null` lost the value in every path; executor normalization also failed through an end-to-end sequential Agent call. Direct serialization and snapshot round trips preserved the same matrix, isolating the loss to those constructors.

Please propagate `structuredContent` with an explicit `!== undefined` guard at each semantics-preserving reconstruction site (or deliberately offload it to a retrievable representation), with regression tests for an object and falsey values.

strandly-the-agent (changes requested):
🔴 **Changes requested** — real MCP non-object values fail before the mapper, and routine block reconstructions erase `structuredContent`.

✅ Reviewed `ea63dfe1`; focused tests **186/186** + type-check passed; cold adversarial repro was byte-identical twice; exact-head CI is green.

API review labels are complete. The proposed field/name are sound and don't need an API meeting.

<details>
<summary>API changes breakdown + verification</summary>

### Public surface

- `ToolResultBlockData` adds `structuredContent?: JSONValue`.
- `ToolResultBlock` adds `readonly structuredContent?: JSONValue` plus constructor support.
- `McpTool` maps MCP `CallToolResult.structuredContent` into that field.
- No new methods, config, hooks, errors, or exports.

### Defaults and serialization

- `undefined` means absent and is omitted from JSON.
- `null`, `false`, `0`, `""`, arrays, and objects are intended to remain present.
- Direct `toJSON`/`fromJSON` and complete agent snapshot round trips preserve those values.
- Existing serialized data remains readable; older SDK readers will silently discard the new key.

### Failure modes and errors

- Locked `@modelcontextprotocol/sdk` 1.29.0 accepts object-valued `structuredContent`, but rejects scalars, arrays, and `null` before `McpTool` sees them. Strands catches that validation failure and returns an error tool result.
- Sliding-window truncation, ContextOffloader replacement, and tool-use-ID normalization silently delete the field.
- Provider adapters intentionally omit it from model payloads; it is application/hook/session data, not extra model context.

### Naming and alternatives

- `structuredContent` is the right protocol-native name.
- `structuredOutput` would blur this with the SDK's existing structured-output concepts.
- The base `ToolResultBlock` placement was already reviewed; it fits TS revival semantics better than an MCP subtype that would be downcast on restore.

### SDK consistency

- The proposed TS `JSONValue` type follows SEP-2106 more closely than Python's current object-only `MCPToolResult`.
- Runtime TS behavior does not yet match that type because the locked MCP transport remains object-only.
- Existing TS `ToolResultBlock` reconstruction sites need to preserve the newly added field, just as they preserve status/content/error.

### Verification

- Exact head: `ea63dfe15f6bb004cc8d33a31cf251bd0fc7c49b`; merge-base: `0fe3ed07ef07712c616029dac76a956d036a8cb6`.
- `npm test -w strands-ts -- src/types/__tests__/messages.test.ts src/mcp/__tests__/client.test.ts`: **186 passed**, no type errors.
- `npm run type-check -w strands-ts`: passed.
- Real in-memory MCP server/client repro: object succeeds; `42`, `[1,2]`, `false`, `0`, `""`, and `null` fail with expected-record validation errors.
- Reconstruction repro covered object plus falsey JSON values through sliding-window, ContextOffloader, and executor normalization; repeated cold runs were byte-identical.
- Exact-head GitHub CI rollup: success across TS Node 20/22/24 on Ubuntu/Windows/macOS, quality, security, packaging, docs, integration, and CI Gate.

### Review passes

- Correctness/adversarial: two confirmed findings below.
- API/DevX: shape and naming accepted; API labels complete; no meeting needed.
- Test quality: transport and reconstruction gaps folded into the fixes below.
- LLM context: provider exclusion is intentional; application-side transformation loss is not.
- Suppressed as already covered: base-vs-MCP subtype, Python type parity, and the whole-object assertion suggestion.

</details>

poshinchen:
@strandly-the-agent I've made new changes, check the comments and review the PR again.

strands-ts/src/tools/mcp-tool.ts
strandly-the-agent: ✅ Verified at `ccc1aec` — this one's addressed. Your reframing is factually correct: I re-checked the published tarballs, and `@modelcontextprotocol/sdk` **1.30.0** (latest 1.x) still has `structuredContent: z.record(z.string(), z.unknown())` and `outputSchema` pinned to `type: z.literal('object')`, so non-object values genuinely cannot reach the mapper from a real server today. Keeping `JSONValue` + documenting the limitation is the right call.

One pointer for the follow-up: SEP-2106 isn't coming to the 1.x line — `@modelcontextprotocol/client@2.0.0` is where `structuredContent: z.unknown().optional()` lives, so it's a package migration, not a version bump. Worth wording the comment that way.

And the new `client.structured-content.test.node.ts` does run for real: `*.test.node.ts` is in the `unit-node` project globs; it passed here and in CI across Node 20/22/24 × 3 OS.

strands-ts/src/types/messages.ts
strandly-the-agent: ✅ Fixed — verified. All three paths now propagate the field: `sliding-window-conversation-manager.ts:417-423`, `context-offloader/plugin.ts:495-501`, `executor.ts:161-165`.

Evidence beyond "tests pass": reverting each hunk individually makes its own new test fail (5/5 `it.each` cases each), and cold repros through the public entry points — `manager.reduce()`, the `AfterToolCallEvent` hook, and a full `Agent.invoke()` with a hook issuing a mismatched `toolUseId` — now preserve object, `0`, `""`, `false`, `null`, and arrays, where the same assertions all failed at `ea63dfe1`. Your comments' rationale also checks out: `estimateContentBlockTokens` (`models/model.ts:549-557`) counts only text/JSON items inside a tool result, so preserving the field doesn't defeat truncation or offloading.

I did find a fourth reconstruction site (`Agent._redactLastMessage`) that still drops it, but it's pre-existing code outside this diff — raising it in the review body as a non-blocking question rather than reopening this thread.

strandly-the-agent (approved):
✅ **Approving** — both of my round-1 blockers are verifiably addressed at `ccc1aec`. One non-blocking question below on pre-existing code.

- ✅ **Reconstruction paths fixed.** Sliding-window, ContextOffloader, and executor ID normalization now propagate `structuredContent`. Mutation-tested (reverting each hunk fails its own new test), plus cold repros through `manager.reduce()`, the `AfterToolCallEvent` hook, and a full `Agent.invoke()` — object, `0`, `""`, `false`, `null`, arrays all survive, where the same assertions failed at `ea63dfe1`.
- ✅ **"Any JSON value" claim.** Accurate as now documented: `@modelcontextprotocol/sdk` 1.30.0 (latest 1.x) is still object-only, so scoping the claim in the comment + adding a real-transport test is the right resolution. FYI it's a v2-package migration, not a version bump — `@modelcontextprotocol/client@2.0.0` is where `z.unknown()` lands.
- ✅ 6 PR-touched suites: **366 tests passed**, no type errors; exact-head CI green.

**❓ Your call:** pre-existing `Agent._redactLastMessage` (`agent.ts:2290-2296`) is a fourth path that rebuilds tool-result blocks without `structuredContent`, while its own comment at `:2289` says it only redacts content. Deliberate scrub (the field can carry the very payload the guardrail flagged) or an oversight? Either answer is defensible — worth a comment + test to lock the intent in, here or as a fast-follow.

<details>
<summary>Evidence, nits, and what I did not re-raise</summary>

### Verified this round (head `ccc1aecb77d456de84d23718546e99b6ba5059d3`, merge-base `0fe3ed07`)

- `vitest run --project unit-node` over the six PR-touched test files: **6 files / 366 tests passed, no type errors**; tracked worktree clean.
- Each propagation fix mutation-tested in a scratch copy: reverting `sliding-window-conversation-manager.ts:417-423`, `context-offloader/plugin.ts:495-501`, `executor.ts:161-165`, or the `mcp-tool.ts` mapper hunk makes the corresponding new test fail (5/5 `it.each` cases per path). The new tests genuinely guard the behavior.
- Cold repro at both heads: the identical 18 assertions (object + `0`/`""`/`false`/`null`/array across the three paths) fail at `ea63dfe1` and pass at `ccc1aec`.
- Your "never sent to the model" rationale checks out: `estimateContentBlockTokens` (`models/model.ts:549-557`) only counts text/JSON items inside a tool result, so preserving the field doesn't blunt truncation or offloading.
- `client.structured-content.test.node.ts` really executes: `*.test.node.ts` is in the `unit-node` vitest project globs, `zod/v4` resolves via the hoisted `zod` peer dep, and CI runs `--project unit-node` on Node 20/22/24 × ubuntu/windows/macos.
- MCP SDK claim checked against published tarballs, not docs: 1.30.0 `dist/esm/types.js` still has `structuredContent: z.record(z.string(), z.unknown())` (:1302) and `outputSchema` at `type: z.literal('object')`; `@modelcontextprotocol/client@2.0.0` has `structuredContent: z.unknown().optional()` with explicit SEP-2106 handling.
- Redaction gap reproduced twice with identical output: after `_redactLastMessage` on a block carrying `structuredContent: { secret: 42 }`, the surviving block is `{"content":["[REDACTED]"]}`. Reached via guardrail input redaction at `agent.ts:1981`, and it mutates `agent.messages` in place, so the loss persists into history and session snapshots. Pre-existing — not introduced here.
- API labels `api/needs-review` + `api/review-complete` both present; exact-head rollup SUCCESS (TS matrix, Code Quality, security audit, pack install, docs, integration, CI Gate).
- Note: `npm run type-check -w strands-ts` fails only in its `test/integ` half (TS6305 + pre-existing errors); the identical failure reproduces at the merge-base, so not a regression. The `src` half is clean.

### Nits (take or leave)

- `context-offloader/plugin.ts:499-500` now also preserves `error`, but `_handleToolResult` returns early for `status === 'error'` at `:436`, so that spread is unreachable today.
- The object-only caveat lives only in the mapper comment; the public JSDoc at `messages.ts:377` and `:414` (and the PR description) still says "any JSON value" unqualified — that's what users actually read.
- Nothing locks in the "not sent to model providers" promise: no adapter references `structuredContent` today (grep-confirmed across bedrock/anthropic/openai/google/vercel), so a future `{...block}` spread would leak it silently. A one-line provider test would fix that cheaply.

### Not re-raised

Base `ToolResultBlock` vs MCP subtype placement, Python object-only parity follow-up, and the whole-object assertion suggestion — all previously settled.

</details>

strands-ts/src/tools/mcp-tool.ts
zastrowm, line 1: > MCP spec revision 2025-06-18 added structured tool output: a server declares outputSchema on a tool and returns the typed value as structuredContent alongside the human-readable content blocks.

How does Python do this and/or is it supported there?

I think in TS, this is more of a metadata thing and I wonder if it should carry around a metadata object instead of this specific structuredContent, given that only MCP uses it

poshinchen: yes Python supported it already. 

Python defines MCPToolResult(ToolResult) in [mcp_types with structuredContent: NotRequired[dict[str, Any]]](https://github.com/strands-agents/harness-sdk/blob/587efb391a1f388bb630c0ea7dcbf88a8f8b1af3/strands-py/src/strands/tools/mcp/mcp_types.py#L73), populated from `[CallToolResult.structuredContent](https://github.com/strands-agents/harness-sdk/blob/main/strands-py/src/strands/tools/mcp/mcp_client.py#L1004)` in mcp_client. 

I would argue structuredContent is the tool's return value rather than metadata. In the MCP spec it is the typed output validated against the tool's outputSchema, and MCP has a separate `_meta` field for actual metadata. 
Python mirrors that distinction: its MCPToolResult carries both structuredContent and a metadata field for _meta. Also, since **outputSchema** already lives on the generic ToolSpec in TS, so any tool author can declare an output schema today. Keeping structuredContent field keeps the schema/value pair together.

strands-ts/src/types/messages.ts
pgrayy, line 379: Does this field need to be filtered out before sending to the model? Is there any model provider that would reject it?

pgrayy: Alternatively, should this just be an MCP only field. In Python, you noted that it is present on an MCPToolResult payload ([src](https://github.com/strands-agents/harness-sdk/blob/587efb391a1f388bb630c0ea7dcbf88a8f8b1af3/strands-py/src/strands/tools/mcp/mcp_types.py#L73)).

pgrayy: One more question, this would be reporting the result twice in the same payload correct. Isn't content the same as structuredContent maybe just structured slightly differently. What is the reason for wanting to propagate structuredContent in a separate field?

poshinchen: The current PR having the structuredContent outside of the MCPToolResult can be referenced in hooks too. And to Dean's [comment](https://github.com/strands-agents/harness-sdk/pull/528#issuecomment-3120600490):

> One interesting side effect, is that now overloaded ToolResults are allowable not just by MCP but by all tool implementations. This presents a new mechanism to persist state.That's why TS does that more outputSchema, and we can do the same for StructuredContent.

But you're right, content is always used by the tools / models. And the structuredContent is for application / user, not the model.

PR #3434, opened 2026-07-24
feat(model-routing): thread per-call model through InvokeModelStage

JackYPCOnline:
## Description

First slice of model routing (design in `team/designs/0016-model-routing.md`): the per-call model plumbing that later routing middleware builds on.

`InvokeModelStage`'s terminal previously read `agent.model` directly, so nothing could redirect an individual model call without mutating shared agent state (unsafe under concurrent invocations). This PR threads the model to invoke through the middleware context instead:

- `InvokeModelContext` gains a required `model: Model`, initialized from `agent.model`.
- The invoke-model terminal streams `ctx.model` (and reads its `model_id` for the span) rather than `agent.model`.

No behavior change for existing single-model agents. This is intentionally a standalone, inert seam; the `ModelRouter` type and strategies land in follow-up PRs.

## Related Issues

Refs #364

## Type of Change

New feature

## Public API note

`InvokeModelContext` (consumed by invoke-stage middleware authors) now carries the concrete model for the call, which middleware may replace to redirect a single invocation. The natural form is an Input-phase handler (a pure context transform):

```python
def route(context):
    return replace(context, model=my_other_model)  # this call streams my_other_model
```

The field is required, not optional: every model call has a concrete model, and the terminal relies on it, so an absent model is a construction-time error rather than a runtime `None` in the hot path.

## Testing

Added tests: the context exposes `agent.model` by default, and a middleware overriding `ctx.model` redirects the streamed model. Full Python unit suite passes locally; format and lint clean.

## Checklist

> Draft: opening early for review of the routing seam before `ModelRouter` lands.

### DIFF
--- a/strands-py/src/strands/_context_manager/modes/agentic/agentic_context.py
+++ b/strands-py/src/strands/_context_manager/modes/agentic/agentic_context.py
@@ -8,7 +8,7 @@ so it can decide when to compress.
 
 import logging
 from dataclasses import replace
-from typing import TYPE_CHECKING, Literal
+from typing import Literal
 
 from ...._middleware.stages import InvokeModelContext
 from ...._middleware.types import MiddlewareInputHandler
@@ -25,9 +25,6 @@ from ....tools.decorator import tool
 from ....types.content import Message, _ensure_tracking_id
 from ....types.exceptions import ContextWindowOverflowException
 from ....types.tools import ToolContext
-
-if TYPE_CHECKING:
-    from ....models.model import Model
 
 logger = logging.getLogger(__name__)
 
@@ -286,15 +283,13 @@ def _matches_pin_filter(message: Message, filter: Literal["user", "assistant", "
     return True  # type: ignore[unreachable]
 
 
-def create_token_usage_middleware(model: "Model") -> MiddlewareInputHandler:
+def create_token_usage_middleware() -> MiddlewareInputHandler:
     """Create middleware that appends a ``<context-status>`` block to the last message.
 
-    The block reports projected input-token usage against the model's context window limit so the
-    model can decide when to compress. The original messages are not mutated; the last message is
-    copied and the status text appended to the copy.
-
-    Args:
-        model: The model whose context window limit is reported.
+    The block reports projected input-token usage against the context window limit of the
+    model that will actually handle the call (``context.model``), so guidance stays correct
+    even when middleware has redirected the call to a different model. The original messages
+    are not mutated; the last message is copied and the status text appended to the copy.
 
     Returns:
         An async ``MiddlewareInputHandler`` for the ``InvokeModelStage.Input`` phase.
@@ -305,7 +300,7 @@ def create_token_usage_middleware(model: "Model") -> MiddlewareInputHandler:
         if projected_input_tokens is None:
             return context
 
-        context_window_limit = model.context_window_limit or DEFAULT_CONTEXT_WINDOW_LIMIT
+        context_window_limit = context.model.context_window_limit or DEFAULT_CONTEXT_WINDOW_LIMIT
         remaining = max(0, context_window_limit - projected_input_tokens)
         percent_used = (projected_input_tokens / context_window_limit) * 100
 
--- a/strands-py/src/strands/_middleware/stages.py
+++ b/strands-py/src/strands/_middleware/stages.py
@@ -23,12 +23,11 @@ if TYPE_CHECKING:
 class InvokeModelContext:
     """Context passed to InvokeModelStage middleware.
 
-    All collection fields (messages, system_prompt, tool_specs, tool_choice) are
-    defensive copies — middleware cannot accidentally mutate agent state.
-    invocation_state is shared by reference (hooks and tools write to it during streaming).
-
-    ``model`` is the model this call invokes; it starts as ``agent.model`` and middleware
-    may replace it per call.
+    The collection fields (messages, system_prompt, tool_specs, tool_choice) are defensive
+    copies, so middleware cannot accidentally mutate agent state. ``invocation_state`` and
+    ``model`` are instead shared by reference: ``invocation_state`` is the live dict hooks and
+    tools write to during streaming, and ``model`` is the model this call invokes (it starts
+    as ``agent.model``; middleware may replace it per call).
     """
 
     agent: Agent
--- a/strands-py/src/strands/_middleware/stages.py
+++ b/strands-py/src/strands/_middleware/stages.py
@@ -27,9 +27,8 @@ class InvokeModelContext:
     defensive copies — middleware cannot accidentally mutate agent state.
     invocation_state is shared by reference (hooks and tools write to it during streaming).
 
-    ``model`` is the concrete model the terminal will invoke. It defaults to
-    ``agent.model``; middleware (e.g. model routing) may replace it to redirect a single
-    call without mutating shared agent state.
+    ``model`` is the model this call invokes; it starts as ``agent.model`` and middleware
+    may replace it per call.
     """
 
     agent: Agent
--- a/strands-py/src/strands/_middleware/stages.py
+++ b/strands-py/src/strands/_middleware/stages.py
@@ -13,6 +13,7 @@ if TYPE_CHECKING:
     from ..agent.agent import Agent
     from ..experimental.bidi import BidiAgent
     from ..interrupt import _InterruptState
+    from ..models.model import Model
     from ..types._events import ModelStopReason, ToolResultEvent, TypedEvent
     from ..types.content import Messages, SystemPrompt
     from ..types.tools import AgentTool, ToolChoice, ToolSpec, ToolUse
@@ -25,6 +26,10 @@ class InvokeModelContext:
     All collection fields (messages, system_prompt, tool_specs, tool_choice) are
     defensive copies — middleware cannot accidentally mutate agent state.
     invocation_state is shared by reference (hooks and tools write to it during streaming).
+
+    ``model`` is the concrete model the terminal will invoke. It defaults to
+    ``agent.model``; middleware (e.g. model routing) may replace it to redirect a single
+    call without mutating shared agent state.
     """
 
     agent: Agent
@@ -33,6 +38,7 @@ class InvokeModelContext:
     tool_specs: list[ToolSpec]
     tool_choice: ToolChoice | None
     invocation_state: dict[str, Any]
+    model: Model
     projected_input_tokens: int | None = None
 
 
--- a/strands-py/src/strands/event_loop/event_loop.py
+++ b/strands-py/src/strands/event_loop/event_loop.py
@@ -537,6 +537,7 @@ async def _handle_model_execution(
                 tool_specs=copy.deepcopy(tool_specs),
                 tool_choice=copy.deepcopy(structured_output_context.tool_choice),
                 invocation_state=invocation_state,
+                model=agent.model,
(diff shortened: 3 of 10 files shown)


mehtarac:
@strandly-the-agent , review the implementation

strandly-the-agent (changes requested):
**TL;DR — 🔴 Changes requested:** per-call routing invokes the replacement model but gives it context-capacity guidance for the agent’s default model; this is reproducible current behavior under the documented `replace(context, model=...)` contract.

<details>
<summary>Verification ledger and per-pass breakdown</summary>

### Verification

- ✅ Reviewed exact head `bba0f7692929642868b3d2ef15a25a9d07266625` against merge base `81bccd8420c255b7ea4e23ad67f1fcf099d405e9`; worktree clean.
- ✅ Focused tests: **177 passed in 0.94s**.
- ✅ Ruff lint/format passed for changed files; mypy passed for both changed production files; `git diff --check` passed.
- ✅ Real `Agent(context_manager="agentic")` repro ran twice in fresh processes with byte-identical output.
- ✅ Latest exact-head CI Gate succeeded; Python 3.10–3.14 Linux/Windows and macOS 3.14, Python Lint, Docs, Dependency Review, and codecov passed.
- ⏳ Integration and one review trigger remain waiting. The red rollup includes an older superseded CI Gate run; its latest rerun succeeded.
- ⚠️ Advanced API/adversarial tiers were unavailable under data-retention mode; both passes completed on the default tier.

### Pass arbitration

- **Correctness + adversarial:** Kept as one blocker. Both identify the same sequencing defect, and the cold repro confirms it in the production event loop.
- **API/DevX:** Its “future router” classification does not control here: `InvokeModelContext` currently documents per-call replacement, and the PR description demonstrates `replace(context, model=...)`. The failing composition is therefore present documented behavior, not merely a future limitation.
- **Test quality:** Suppressed the missing direct trace-attribution assertion. It is a useful follow-up but does not independently change the merge decision.
- **LLM context:** No separate finding; its capacity-guidance concern is merged into the blocker.
- **Duplicates suppressed:** agent-global `model_state`, missing design document, API-label confirmation, and the `model=None` helper. The existing base-override result was not repeated.

**Counts:** 5 passes → 3 overlapping observations → 1 posted blocker; 4 existing or low-priority items suppressed.

</details>

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -675,7 +675,7 @@ async def terminal(ctx: InvokeModelContext) -> AsyncGenerator[Any, None]:
         with trace_api.use_span(model_invoke_span, end_on_exit=False):
             try:
                 async for event in stream_messages(
-                    agent.model,
+                    ctx.model,
```
strandly-the-agent, line 678: 🔴 **Block — use the effective routed model for agentic context guidance**

This line correctly invokes `ctx.model`, but agentic token guidance has already been added by middleware that closes over the default `agent.model` and runs before plugin routing. I reproduced this through a real `Agent(context_manager="agentic")` and the production event loop in two fresh processes with byte-identical output. With equal 8,000-token estimates, a 100k default-model window, and a 10k routed-model window:

```text
count_calls=default:1,routed:0
stream_calls=default:0,routed:1
status_received_by_routed_model=<used>8,000 / 100,000 tokens (8.0%)</used>
<remaining>~92,000 tokens</remaining>
expected_for_routed_model=<used>8,000 / 10,000 tokens (80.0%)</used>
<remaining>~2,000 tokens</remaining>
mismatch=True
```

The routed model can therefore defer compression until its actual context window overflows. Please either derive agentic context guidance from the effective routed `context.model`, or establish routing before that model-dependent middleware runs. A mixed-window agentic-mode test should assert that the routed model receives its own limit. A local `suggestion` is not safe because the fix spans middleware ordering and the guidance model source.

JackYPCOnline: I agree, should fix this.

JackYPCOnline:
@strandly-the-agent review the implementation again.

strands-py/src/strands/_middleware/stages.py
```diff
@@ -25,6 +26,9 @@ class InvokeModelContext:
     All collection fields (messages, system_prompt, tool_specs, tool_choice) are
     defensive copies — middleware cannot accidentally mutate agent state.
     invocation_state is shared by reference (hooks and tools write to it during streaming).
+
+    ``model`` is the model this call invokes; it starts as ``agent.model`` and middleware
```
zastrowm: Combine the call out for `invocation_state` and `model` and indicate that both are by-reference

mehtarac (dismissed):

strands-py/src/strands/_context_manager/modes/agentic/agentic_context.py
zastrowm, line 1: Basically approved + one nit; want to do TS at the same time?

Updating https://github.com/strands-agents/harness-sdk/blob/bfa080a208e88112184cae4dc0847a5556a7d32b/strands-ts/src/middleware/README.md#L4 would be ideal too (there's a python one, but it references the TS one)

JackYPCOnline: I would be more confident to have one complete version of model routing then port to TS later, will iterate the nits.

strandly-the-agent (changes requested):
🔴 **Re-review result: the existing blocker remains at live head `df7d1e6eb46b91e7802400fb8c1a9c7a18f8ee7e`.**

The fix commit reads `context.model`, but the middleware still runs before the same-phase routing plugin. Agentic status therefore uses the default model’s window while the terminal streams the routed model. The branch advanced from `ec459cb` via a merge from `main`; that merge did not change the relevant Python paths, and the same full-Agent repro remains byte-identical at `df7d1e6`.

✅ Exact live head reviewed · ✅ real `Agent` repro repeated twice byte-identically · ✅ 44 focused tests passed · ✅ Ruff, mypy, and diff-check passed · ✅ exact-head GitHub CI is green

Existing blocker: https://github.com/strands-agents/harness-sdk/pull/3434#discussion_r3646245121

<details>
<summary>Verification details</summary>

- Built-in agentic Input middleware is registered at `agent.py:409-414`, before user plugins at `agent.py:478-484`.
- Same-phase handlers retain registration order in `registry.py:62-73,126-153`, so `context.model` is still the default when token status is added.
- With equal 8k estimates, a 100k default model, and a 10k routed model, the routed model streamed but received `8% / ~92k remaining`; expected output is `80% / ~2k remaining`.
- Both cold runs recorded default/routed `count_tokens` calls as `1/0` and stream calls as `0/1`.
- The new unit test directly supplies an already-routed context, so it does not exercise real `Agent`/plugin ordering. The regression test needs a normal routing plugin and unequal context windows.
- No safe local suggestion: the correction spans middleware registration/ordering across files.
- Existing unrelated `model_state`, design, and API concerns were not duplicated.

GitHub checks at `df7d1e6` are green across Python 3.10–3.14, Linux/Windows/macOS 3.14, lint, docs, dependency, integration, codecov, and CI Gate.

</details>

strands-py/src/strands/event_loop/event_loop.py
strandly-the-agent: 🔴 Still reproducible at live head `df7d1e6eb46b91e7802400fb8c1a9c7a18f8ee7e`.

Reading `context.model` dynamically does not fix the ordering: the built-in agentic Input middleware is registered before user plugins (`agent.py:409-414` vs. `agent.py:478-484`), and same-phase handlers preserve that order (`registry.py:62-73,126-153`). At status-injection time, `context.model` is therefore still the default model; routing happens afterward.

A real `Agent(context_manager="agentic")` with a normal routing plugin reproduced this twice byte-identically: the 10k routed model streamed, but status reported `8% / ~92k remaining` from the 100k default model. Expected: `80% / ~2k remaining`. Default/routed `count_tokens` calls were `1/0`; stream calls were `0/1`.

The branch advanced from `ec459cb` via a merge from `main`; that merge did not change the relevant Python paths, and the repro is unchanged at `df7d1e6`. The added test constructs an already-routed context directly, so it misses this Agent/plugin ordering path. The fix still needs routing to precede model-dependent middleware, plus a real-Agent regression test with unequal model windows. There is no safe one-line suggestion because that ordering change spans multiple files.

strandly-the-agent, line 678: 🔴 **Block — supersedes the resolved routing-guidance thread**

This exact-head review supersedes the resolved thread at `discussion_r3646245121`: the bug remains at `df7d1e6eb46b91e7802400fb8c1a9c7a18f8ee7e`.

`ctx.model` is the routed model here, but the agentic status middleware has already run. It is registered before user plugins (`agent.py:409-414` vs. `agent.py:478-484`), and same-phase handlers preserve registration order (`registry.py:62-73,126-153`). Reading `context.model` dynamically therefore still reads the default model before routing occurs.

A deterministic real-`Agent` repro, run twice byte-identically, used equal 8,000-token estimates, a 100k default model, and a 10k routed model:

```text
default_count_tokens_calls=1
routed_count_tokens_calls=0
default_stream_calls=0
routed_stream_calls=1
routed_status=<used>8,000 / 100,000 tokens (8.0%)</used>
              <remaining>~92,000 tokens</remaining>
expected=     <used>8,000 / 10,000 tokens (80.0%)</used>
              <remaining>~2,000 tokens</remaining>
```

Please make routing execute before model-dependent agentic guidance (through explicit ordering/priority or an earlier routing phase), then add a full `Agent(context_manager="agentic")` regression test with a normal routing plugin and unequal model windows. The current direct-handler test starts with `context.model` already routed and cannot detect this ordering failure. A one-line suggestion is not safe because the fix spans middleware ordering and registration.

strandly-the-agent (changes requested):
🔴 **Changes requested at exact live head `df7d1e6eb46b91e7802400fb8c1a9c7a18f8ee7e`: routing still runs after agentic guidance.**

The prior thread was resolved, but the full `Agent` path remains reproducibly wrong; the inline blocker below supersedes that resolved thread.

✅ Exact-head repro twice byte-identically · ✅ 44 focused tests · ✅ Ruff/mypy/diff-check · ✅ exact-head GitHub matrix, integration, lint, docs, dependency, codecov, and CI Gate

<details>
<summary>Pass and verification breakdown</summary>

- **Correctness:** blocker remains. `context.model` is still the default when built-in agentic Input middleware runs; plugin routing occurs later.
- **Adversarial:** real `Agent(context_manager="agentic")` plus a normal routing plugin reproduced twice at `df7d1e6`: default/routed token counts `1/0`, stream calls `0/1`, routed 10k model receives default 100k guidance (`8% / ~92k`) instead of `80% / ~2k`.
- **Test quality:** the new test directly constructs an already-routed context and therefore misses middleware registration order. A full-Agent mixed-window plugin test is required.
- **API/DevX:** no new surface regression in this delta; existing private-surface questions are unchanged.
- **LLM context:** folded into the blocker because incorrect capacity guidance is model-facing.
- **Suppressed as unchanged/nonblocking:** model-state threading, design-doc reference, API-label discussion, and TS parity comments.
- **Verification:** 44 focused tests passed; Ruff lint/format, mypy, and `git diff --check` passed; worktree clean. Exact-head hosted Python 3.10–3.14 Linux/Windows plus macOS 3.14, integration, lint, docs, Dependency Review, codecov, and CI Gate are green.
- Advanced adversarial tier was unavailable under data-retention mode and completed on the default tier.

</details>

zastrowm (approved):

PR #3435, opened 2026-07-24
perf(mcp): hydrate search snippets concurrently

### LINKED ISSUE #3329: [FEATURE] strands-mcp: hydrate search_docs snippets concurrently
### SDK Language

Python

### Problem Statement

`search_docs` hydrates up to 5 top results with sequential blocking HTTP fetches inside the tool call (`SNIPPET_HYDRATE_MAX`). First-search latency is up to 5x a single page fetch, and with the 30s default timeout a couple of slow or broken pages can make one tool call take minutes.

### Proposed Solution

Fetch the hydration set concurrently (`ThreadPoolExecutor` is enough), or prefetch pages in a background thread at startup so searches never block on network. If the content-indexing fix (#3321) is implemented via startup prefetch, this may come for free.

### Use Case

First search in a session returns in roughly one page-fetch time instead of five, keeping the MCP tool call responsive.

### Alternatives Solutions

_No response_

### Additional Context

Part of the strands-mcp improvements tracking issue #3317.

March-77:
## Description

`search_docs` hydrated up to five uncached result pages serially, so first-search latency grew with every slow page. This change runs the existing bounded hydration set concurrently while preserving ranking and response construction order.

A synthetic five-page benchmark with 50 ms fixed fetch latency improved median hydration time from 269.5 ms to 56.3 ms across five runs.

Fixes #3329

### Compatibility / Risk

The response schema and ordering are unchanged. The executor is bounded by `SNIPPET_HYDRATE_MAX` (currently five), and cache writes remain delegated to the existing `ensure_page` path.

## Related Issues

Fixes #3329

## Documentation PR

No documentation change is needed because the tool API and response behavior are unchanged.

## Type of Change

Performance improvement

## Testing

- `hatch fmt --formatter` — passed
- `hatch fmt --linter` — passed
- `.venv/bin/pytest tests/test_server.py -q` — 16 passed
- `.venv/bin/pytest tests/ -q` — 48 passed
- Synthetic five-page, 50 ms/fetch benchmark — sequential median 269.5 ms; concurrent median 56.3 ms

- [ ] I ran `hatch run prepare` (the MCP package does not define a `prepare` script; the package lint and test commands above were run directly)

## Checklist

----

### DIFF
--- a/strands-mcp/src/strands_mcp_server/server.py
+++ b/strands-mcp/src/strands_mcp_server/server.py
@@ -1,3 +1,4 @@
+from concurrent.futures import ThreadPoolExecutor
 from typing import Any, Dict, List
 from urllib.parse import urlparse
 
@@ -61,13 +62,11 @@ def search_docs(query: str, k: int = 5) -> List[Dict[str, Any]]:
     results = index.search(query, k=k) if index else []
     url_cache = cache.get_url_cache()
 
-    # Collect top-k URLs that need hydration (no content yet)
-    # Simplified: Direct hydration in one pass
     top = results[: min(len(results), cache.SNIPPET_HYDRATE_MAX)]
-    for _, doc in top:
-        cached = url_cache.get(doc.uri)
-        if cached is None or not cached.content:
-            cache.ensure_page(doc.uri)
+    urls_to_hydrate = [doc.uri for _, doc in top if (page := url_cache.get(doc.uri)) is None or not page.content]
+    if urls_to_hydrate:
+        with ThreadPoolExecutor(max_workers=len(urls_to_hydrate)) as executor:
+            list(executor.map(cache.ensure_page, urls_to_hydrate))
 
     # Build response with real content snippets when available
     return_docs: List[Dict[str, Any]] = []
--- a/strands-mcp/src/strands_mcp_server/server.py
+++ b/strands-mcp/src/strands_mcp_server/server.py
@@ -63,7 +63,9 @@ def search_docs(query: str, k: int = 5) -> List[Dict[str, Any]]:
     url_cache = cache.get_url_cache()
 
     top = results[: min(len(results), cache.SNIPPET_HYDRATE_MAX)]
-    urls_to_hydrate = [doc.uri for _, doc in top if (page := url_cache.get(doc.uri)) is None or not page.content]
+    urls_to_hydrate = list(
+        dict.fromkeys(doc.uri for _, doc in top if (page := url_cache.get(doc.uri)) is None or not page.content)
+    )
     if urls_to_hydrate:
         with ThreadPoolExecutor(max_workers=len(urls_to_hydrate)) as executor:
             list(executor.map(cache.ensure_page, urls_to_hydrate))
--- a/strands-mcp/tests/test_server.py
+++ b/strands-mcp/tests/test_server.py
@@ -17,7 +17,8 @@ def test_search_docs_hydrates_pages_concurrently(mock_cache):
         Doc(uri=f"https://strandsagents.com/{index}.md", display_title=f"Doc {index}", content="", index_title="")
         for index in range(2)
     ]
-    mock_cache.get_index.return_value.search.return_value = [(1.0, doc) for doc in docs]
+    ranked_docs = [docs[0], docs[0], docs[1]]
+    mock_cache.get_index.return_value.search.return_value = [(1.0, doc) for doc in ranked_docs]
     mock_cache.get_url_cache.return_value = {doc.uri: None for doc in docs}
     mock_cache.SNIPPET_HYDRATE_MAX = 5
 
@@ -42,7 +43,9 @@ def test_search_docs_hydrates_pages_concurrently(mock_cache):
     tru_result = search_docs("agent")
 
     assert max_active_fetches == 2
-    assert [result["url"] for result in tru_result] == [doc.uri for doc in docs]
+    assert mock_cache.ensure_page.call_count == 2
+    assert {call.args[0] for call in mock_cache.ensure_page.call_args_list} == {doc.uri for doc in docs}
+    assert [result["url"] for result in tru_result] == [doc.uri for doc in ranked_docs]
 
 
 @patch("strands_mcp_server.server.cache")
--- a/strands-mcp/tests/test_server.py
+++ b/strands-mcp/tests/test_server.py
@@ -1,11 +1,48 @@
-"""Tests for fetch_doc MCP tool."""
+"""Tests for MCP server tools."""
 
+from threading import Event, Lock
 from unittest.mock import patch
 
 import pytest
 
-from strands_mcp_server.server import fetch_doc
+from strands_mcp_server.server import fetch_doc, search_docs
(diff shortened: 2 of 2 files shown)


yonib05:

strands-mcp/src/strands_mcp_server/server.py
```diff
@@ -61,13 +62,11 @@ def search_docs(query: str, k: int = 5) -> List[Dict[str, Any]]:
...
 
-    # Collect top-k URLs that need hydration (no content yet)
-    # Simplified: Direct hydration in one pass
     top = results[: min(len(results), cache.SNIPPET_HYDRATE_MAX)]
-    for _, doc in top:
-        cached = url_cache.get(doc.uri)
-        if cached is None or not cached.content:
-            cache.ensure_page(doc.uri)
+    urls_to_hydrate = [doc.uri for _, doc in top if (page := url_cache.get(doc.uri)) is None or not page.content]
+    if urls_to_hydrate:
```
strandly-the-agent, line 69: 🟡 **Should-fix:** deduplicate the hydration set before starting workers. The loader adds every parsed link to the index, so a repeated URL can appear twice here. A cold base/head repro run twice showed base fetching it once, while this branch fetched it twice concurrently and left the cached snippet dependent on the last writer. The live catalog currently has no duplicates, so this is non-blocking, but preserving uniqueness here avoids introducing that failure mode.

```suggestion:-1+0
    urls_to_hydrate = list(
        dict.fromkeys(
            doc.uri
            for _, doc in top
            if (page := url_cache.get(doc.uri)) is None or not page.content
        )
    )
    if urls_to_hydrate:
```

strandly-the-agent (approved):
**TL;DR: Approve — the bounded parallel hydration works; one non-blocking duplicate-URL robustness fix is worth making.**

✅ Reviewed exact head `739e63f705f52a8639b5b1906c03ded4d0e1efdc` against base `1b671d6a1a3609ae3f46bc20c599e4c9ad0db4cf`  
✅ Focused tests: 16 passed; full MCP unit suite: 48 passed  
✅ Ruff lint/format and `git diff --check` passed  
✅ Exercised ordering, failure isolation, five-fetch bound, and executor cleanup  
🟡 Repeated result URLs can be fetched twice concurrently; details inline  
⏳ Exact-head policy checks are green; `Trigger Strands Review` is still waiting

<details><summary>Routing and per-pass results</summary>

| Specialist pass | Route | Outcome |
|---|---|---|
| Correctness / safety | **Run** — production concurrency logic changed | No blocking defect; the five-page bound, executor join, and ranked response order are preserved. Generic concurrent same-URL calls also race on base, so no unrelated global-lock finding. |
| Adversarial / repro | **Run** — new threading, shared cache, failure and ordering paths | One non-blocking duplicate-URL regression reproduced cold twice against base; ordering, failure isolation, five-fetch bound, Python compatibility, and executor cleanup survived. The advanced tier was unavailable due data-retention mode, so this pass was retried successfully on the default tier. |
| Test quality | **Run** — regression test added for concurrent behavior | 48/48 MCP unit tests passed and the concurrency test was stable across repeated runs. Two narrow assertion gaps were suppressed because they did not establish another production defect beyond the reproduced duplicate-URL case. |
| API / DevX | **Skipped** — no public signature, schema, export, error, or behavior-contract change | No API-review label or meeting required. |
| LLM-context | **Skipped** — no tool description, prompt, result schema, or model-facing wording changed | Existing snippet/result contract is unchanged. |
| Docs accuracy | **Skipped** — no docs, README, or user-facing wording changed | The performance-only change does not require docs updates. |

The current live catalog has 775 unique URLs, so the reproduced duplicate case is a robustness issue rather than a reason to hold up the latency fix.

</details>

strands-mcp/src/strands_mcp_server/server.py
March-77: Fixed in e8f97f4c. Hydration URLs are now deduplicated in ranked order before the thread pool starts. The concurrency test now includes a repeated result URL and verifies two unique pages are fetched once each while the original ranked result order is preserved. All 48 MCP unit tests and Ruff checks pass.

yonib05:
@strandly-the-agent can you rereview this PR?

strandly-the-agent (approved):
**TL;DR: Re-review of `e8f97f4c` — approve. The duplicate-URL fix is verified by differential repro; no new findings, nothing outstanding.**

✅ Re-reviewed only the incremental diff `739e63f7..e8f97f4c` (dedup at `server.py:66-68` + extended test)  
✅ Differential cold repro ×2 per tree: duplicate URL fetched **2×** at `739e63f7` (cache kept the second writer) vs **1×** at `e8f97f4c` — byte-identical cache/snippet outcome to base, while keeping the latency win (0.251s vs 0.501s serial)  
✅ Focused tests 16 passed; full MCP unit suite 48 passed; concurrency test stable over repeated runs; Ruff lint/format and `git diff --check` clean  
✅ Mutation-checked the new test: reverting the dedup fails it deterministically (50/50), so `test_server.py:46` genuinely guards the fix  
✅ Exact-head CI: `MCP / Test and Lint` and `CI Gate` green; `Trigger Strands Review` still waiting  
🟡 No new should-fix or blocking findings

<details><summary>Routing and per-pass results (incremental re-review)</summary>

| Specialist pass | Route | Outcome |
|---|---|---|
| Adversarial / repro | **Run** — the fix touches the concurrent hydration path | No findings. 31-attack matrix passes at head: dedup with `A,A,B` and `A,A,A`, `max_workers` shrinking with the deduped set, empty/all-cached sets taking no executor, interleaved duplicates preserving ranked order verbatim, one hydration failing or returning `None` not affecting siblings, no thread leaks over 20 rounds, walrus binding not leaking into snippets, and Python 3.10 parse compatibility. The `SNIPPET_HYDRATE_MAX` window still applies to the first five ranked entries rather than five uniques — identical to base, so dedup causes no fetch-budget change. |
| Test quality | **Run** — the regression test changed | No findings. The extended test kills reverting the dedup, serializing the loop, and hydrating only the first URL; 50/50 deterministic in both directions, bounded by the 0.5s event timeout with no hang path. Worth knowing for future edits: `ensure_page.call_count` at `test_server.py:46` is the assertion carrying the dedup guarantee, not `max_active_fetches` at `:45`. |
| Correctness / safety | **Folded into the adversarial pass** — the incremental change is a three-line, behavior-preserving dedup already exercised end to end by that pass's differential repro and attack matrix | No findings. |
| API / DevX | **Skipped** — no public signature, schema, export, error, or response-contract change | No API-review label or meeting required. |
| LLM-context | **Skipped** — no tool description, prompt, or model-facing result text changed | Snippet contract unchanged. |
| Docs accuracy | **Skipped** — no docs, README, or wording changed | Performance-only change. |

Deliberately not re-raised (pre-existing on base `1b671d6a`, unchanged by this PR): the unlocked global `_URL_CACHE` in `cache.py:8,58-78`, races between separate concurrent `search_docs` calls, and exceptions from `ensure_page` propagating out of the tool. The live catalog currently lists 792 URLs with no duplicates, so the fixed case was latent rather than active.

</details>

yonib05 (approved):

PR #3417, opened 2026-07-22
fix(mcp): deduplicate inverted-index postings

March-77:
## Description

Repeated tokens currently add the same document to an inverted-index posting list multiple times, so searches recompute and accumulate the document's full term score for every occurrence. Keep posting-list membership aligned with document frequency by adding each token's document index only once.

## Related Issues

Resolves #3325

## Documentation PR

Not required; this is an internal bug fix with no public API changes.

## Type of Change

Bug fix

## Testing

The `strands-mcp` format check, lint, and full unit suite pass locally (49 tests). The repository's `hatch run prepare` script is not defined for this subproject.

- [ ] I ran `hatch run prepare`

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have reviewed and understand every line of code in this PR, including any generated by AI tools, and I can explain why it works
- [x] My change is focused and reasonably small; I have split unrelated work into separate PRs
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [ ] I have updated the documentation accordingly
- [ ] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3436, opened 2026-07-24
feat(mcp): normalize search token inflections

### LINKED ISSUE #3322: [FEATURE] strands-mcp: add stemming/token normalization to search
### SDK Language

Python

### Problem Statement

Search in strands-mcp uses exact token matching with no stemming or normalization. Verified against the live index: "streaming" returns 3 results while "stream" returns 0; "guardrails" returns 3 while "guardrail" returns 0. LLMs phrase queries unpredictably, so silent misses on singular/plural or verb-form variants are a large class of failures for this tool's primary consumer - and an empty result reads as "the docs don't cover this".

### Proposed Solution

Apply a crude suffix-stripping stemmer (s / es / ing / ed) at both index and query time. Alternatively, prefix matching on index tokens. No external dependencies needed.

### Use Case

Any AI coding assistant querying Strands docs via the MCP server gets results for morphological variants of indexed terms instead of empty responses.

### Alternatives Solutions

_No response_

### Additional Context

Part of the strands-mcp improvements tracking issue #3317.

March-77:
## Description

Documentation search required exact token matches, so common variants such as `stream`/`streaming` and `guardrail`/`guardrails` produced false empty results. This change applies the same lightweight, dependency-free normalization to indexed text, queries, and relevance scoring.

Fixes #3322

### Compatibility / Risk

The tool API and response schema are unchanged. Normalization is intentionally limited to common English `-s`, `-es`, `-ies`, `-ed`, and `-ing` forms, with guards for words ending in `-ss`, `-us`, and `-is`.

## Related Issues

Fixes #3322

## Documentation PR

No documentation change is needed because this corrects existing search behavior without changing the tool interface.

## Type of Change

New feature

## Testing

- `.venv/bin/pytest tests/test_indexer.py -q` — 14 passed
- `.venv/bin/pytest tests/ -q` — 61 passed
- `hatch fmt --formatter` — passed
- `hatch fmt --linter` — passed

- [ ] I ran `hatch run prepare` (the MCP package does not define a `prepare` script; its formatter, linter, and full unit suite were run directly)

## Checklist

----

### DIFF
--- a/strands-mcp/src/strands_mcp_server/utils/indexer.py
+++ b/strands-mcp/src/strands_mcp_server/utils/indexer.py
@@ -13,6 +13,35 @@ _MD_INLINE_CODE = re.compile(r"`([^`]+)`")
 _MD_LINK_TEXT = re.compile(r"\[([^\]]+)\]\([^)]+\)")
 
 
+def _normalize_token(token: str) -> str:
+    """Normalize common English inflections for search matching."""
+    token = token.lower()
+    if len(token) > 5 and token.endswith("ing"):
+        stem = token[:-3]
+        if len(stem) > 2 and stem[-1] == stem[-2] and stem[-1] not in "aeioulsz":
+            stem = stem[:-1]
+        return stem
+    if len(token) > 4 and token.endswith("ied"):
+        return f"{token[:-3]}y"
+    if len(token) > 4 and token.endswith("ed"):
+        stem = token[:-2]
+        if len(stem) > 2 and stem[-1] == stem[-2] and stem[-1] not in "aeioulsz":
+            stem = stem[:-1]
+        return stem
+    if len(token) > 4 and token.endswith("ies"):
+        return f"{token[:-3]}y"
+    if len(token) > 4 and token.endswith(("sses", "shes", "ches", "xes", "zes")):
+        return token[:-2]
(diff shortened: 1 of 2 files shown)


yonib05:
Thanks for working on #3322 - the crude dependency-free stemmer is the direction that issue calls for.

Closing this one for now, though, because the normalization as written reintroduces the exact silent-miss failure class the issue is about: it isn't idempotent, so indexed `settings` becomes `setting` while a query for `setting` becomes `set` (a miss), and `tracing` stems to `trac` while `traces` stems to `trace` - so a search for "tracing" still misses the Traces doc, one of the real titles this targets. The `zes` rule also over-strips (`sizes` -> `siz` vs `size` unchanged). It also rewrites the same indexer loop as #3417 without incorporating that fix, so the three search PRs conflict with each other.

If you'd like to resubmit after #3417 lands, with normalization applied to a fixpoint and test rows for the settings/tracing/sizes cases, we'd take another look. One ask per our [contributing guide's AI tools section](https://github.com/strands-agents/harness-sdk/blob/main/CONTRIBUTING.md#using-ai-tools): agents are a tool and you're the author - please hand-verify the behavior (a quick REPL pass over the stemmer against real doc titles would have caught these) before opening the PR.

PR #3441, opened 2026-07-24
feat(gemini): add canonical built_in_tools config

### LINKED ISSUE #3361: [PARITY] Align Python GeminiModel tools field name with TS (gemini_tools → built_in_tools)
### What to migrate

The `GeminiModel` config field for provider-native (non-FunctionDeclaration) tools has a different name in each language:

- **TS (canonical):** `builtInTools?: Tool[]` (`strands-ts/src/models/google/types.ts:43`), appended to the request tools in `strands-ts/src/models/google/model.ts:337`.
- **Python:** `gemini_tools: list[genai.types.Tool]` (`strands-py/src/strands/models/gemini.py:60`), appended in `gemini.py:287` and validated by `_validate_gemini_tools` (`gemini.py:635`).

Both name the same concept — Gemini-specific tools like `GoogleSearch`, `CodeExecution`, `UrlContext` that coexist with function-calling tools — but a consumer configuring the same capability writes `builtInTools` in TS and `gemini_tools` in Python. TS is the canonical naming, so Python should rename `gemini_tools` to `built_in_tools` to match.

Scope of the Python change:
- Add `built_in_tools` to `GeminiConfig` as the canonical field, matching TS `builtInTools`.
- Keep `gemini_tools` as a deprecated alias that still works but emits a deprecation warning, so existing consumers are not broken; remove it in a later release.
- Update the `self.config.get("gemini_tools")` reads in `_format_request_tools` and the `update_config`/`__init__` validation guards to read the canonical field.
- Rename `_validate_gemini_tools` → `_validate_built_in_tools` and its message text.

### Related Files

```
strands-py/src/strands/models/gemini.py
strands-py/tests/strands/models/test_gemini.py
strands-ts/src/models/google/types.ts
strands-ts/src/models/google/model.ts
```

Parent: #2498

ShitK:
## Description

Python users currently configure Gemini provider-native tools with `gemini_tools`, while the TypeScript SDK uses the canonical `builtInTools` name for the same capability. This aligns the Python public API with `built_in_tools` while preserving `gemini_tools` as a deprecated compatibility alias.

### Public API Changes

```python
# Before
model = GeminiModel(
    model_id="gemini-2.5-flash",
    gemini_tools=[genai.types.Tool(google_search=genai.types.GoogleSearch())],
)

# After
model = GeminiModel(
    model_id="gemini-2.5-flash",
    built_in_tools=[genai.types.Tool(google_search=genai.types.GoogleSearch())],
)
```

Existing `gemini_tools` configurations continue to work with a `DeprecationWarning` and are normalized internally to `built_in_tools`. The alias is normalized at write time, so `get_config()` returns only the canonical `built_in_tools` key even when the model was configured through `gemini_tools`.

Supplying both keys in the same configuration raises a `ValueError` instead of silently choosing one. The current check is presence-based, including when one key is explicitly set to `None` or an empty list; maintainer confirmation of that edge-case behavior is requested in the review thread.

## Related Issues

Resolves #3361

## Documentation PR

The Gemini model provider guide is updated in this PR to use `built_in_tools` and document the deprecated `gemini_tools` alias.

## Type of Change

New feature

## Testing

- [x] Gemini model unit tests (`63 passed`)
- [x] Full Python unit suite (`4650 passed`)
- [x] Ruff format and lint checks
- [x] mypy (`231 source files`)
- [x] Documentation build (`857 pages`, no broken links)
- [x] Gemini integration test collection (`13 skipped` without API credentials)

## Checklist

----

### DIFF
--- a/strands-py/src/strands/models/gemini.py
+++ b/strands-py/src/strands/models/gemini.py
@@ -659,7 +659,7 @@ class GeminiModel(Model):
 
         if "gemini_tools" in model_config:
             warnings.warn(
-                "gemini_tools is deprecated; use built_in_tools instead.",
+                "gemini_tools is deprecated and will be removed in v2.0.0. Use built_in_tools instead.",
                 DeprecationWarning,
                 stacklevel=3,
             )
--- a/strands-py/src/strands/models/gemini.py
+++ b/strands-py/src/strands/models/gemini.py
@@ -8,6 +8,7 @@ import json
 import logging
 import mimetypes
 import secrets
+import warnings
 from collections.abc import AsyncGenerator
 from typing import Any, TypeVar, cast
 
@@ -45,11 +46,12 @@ class GeminiModel(Model):
             params: Additional model parameters (e.g., temperature).
                 For a complete list of supported parameters, see
                 https://ai.google.dev/api/generate-content#generationconfig.
-            gemini_tools: Gemini-specific tools that are not FunctionDeclarations
+            built_in_tools: Gemini-specific tools that are not FunctionDeclarations
                 (e.g., GoogleSearch, CodeExecution, ComputerUse, UrlContext, FileSearch).
                 Use the standard tools interface for function calling tools.
                 For a complete list of supported tools, see
                 https://ai.google.dev/api/caching#Tool
+            gemini_tools: Deprecated alias for built_in_tools.
             use_native_token_count: Whether to use the native Gemini count_tokens API.
                 When True, count_tokens() calls the Gemini API for accurate counts.
                 When False (default), skips the API call and uses the local estimator.
@@ -57,6 +59,7 @@ class GeminiModel(Model):
 
         model_id: Required[str]
         params: dict[str, Any]
+        built_in_tools: list[genai.types.Tool]
         gemini_tools: list[genai.types.Tool]
         use_native_token_count: bool
 
@@ -83,9 +86,11 @@ class GeminiModel(Model):
             **model_config: Configuration options for the Gemini model.
 
         Raises:
-            ValueError: If both `client` and `client_args` are provided.
+            ValueError: If both `client` and `client_args` are provided, if both `built_in_tools` and
+                `gemini_tools` are configured, or if built-in tools contain function declarations.
         """
         validate_config_keys(model_config, GeminiModel.GeminiConfig)
+        model_config = self._normalize_built_in_tools(model_config)
         self.config = GeminiModel.GeminiConfig(**model_config)
 
         # Validate that only one client configuration method is provided
@@ -95,9 +100,8 @@ class GeminiModel(Model):
         self._custom_client = client
         self.client_args = client_args or {}
 
-        # Validate gemini_tools if provided
-        if "gemini_tools" in self.config:
-            self._validate_gemini_tools(self.config["gemini_tools"])
+        if "built_in_tools" in self.config:
+            self._validate_built_in_tools(self.config["built_in_tools"])
 
         logger.debug("config=<%s> | initializing", self.config)
 
@@ -107,10 +111,14 @@ class GeminiModel(Model):
 
         Args:
             **model_config: Configuration overrides.
+
+        Raises:
+            ValueError: If both `built_in_tools` and `gemini_tools` are configured, or if built-in tools contain
+                function declarations.
         """
-        # Validate gemini_tools if provided
-        if "gemini_tools" in model_config:
-            self._validate_gemini_tools(model_config["gemini_tools"])
+        model_config = self._normalize_built_in_tools(model_config)
+        if "built_in_tools" in model_config:
+            self._validate_built_in_tools(model_config["built_in_tools"])
 
         self.config.update(model_config)
 
@@ -270,7 +278,7 @@ class GeminiModel(Model):
         Return:
             Gemini tool list, or None when no tools are configured (Vertex AI rejects empty arrays).
         """
-        if not tool_specs and not self.config.get("gemini_tools"):
+        if not tool_specs and not self.config.get("built_in_tools"):
             return None
         tools = [
             genai.types.Tool(
@@ -284,8 +292,8 @@ class GeminiModel(Model):
                 ],
             ),
         ]
-        if self.config.get("gemini_tools"):
-            tools.extend(self.config["gemini_tools"])
+        if self.config.get("built_in_tools"):
+            tools.extend(self.config["built_in_tools"])
         return tools
 
     def _format_request_config(
@@ -632,25 +640,53 @@ class GeminiModel(Model):
         yield {"output": output_model.model_validate(response.parsed)}
 
     @staticmethod
-    def _validate_gemini_tools(gemini_tools: list[genai.types.Tool]) -> None:
-        """Validate that gemini_tools does not contain FunctionDeclarations.
+    def _normalize_built_in_tools(model_config: GeminiConfig) -> GeminiConfig:
+        """Normalize the deprecated gemini_tools alias to built_in_tools.
+
+        Args:
+            model_config: Gemini model configuration to normalize.
+
+        Returns:
+            Gemini model configuration using the canonical built_in_tools field.
+
+        Raises:
+            ValueError: If built_in_tools and gemini_tools are both configured.
+        """
+        if "built_in_tools" in model_config and "gemini_tools" in model_config:
+            raise ValueError(
+                "built_in_tools and gemini_tools cannot be configured in tandem; gemini_tools is deprecated."
+            )
+
+        if "gemini_tools" in model_config:
+            warnings.warn(
+                "gemini_tools is deprecated; use built_in_tools instead.",
+                DeprecationWarning,
+                stacklevel=3,
+            )
+            model_config["built_in_tools"] = model_config.pop("gemini_tools")
+
+        return model_config
+
+    @staticmethod
+    def _validate_built_in_tools(built_in_tools: list[genai.types.Tool]) -> None:
+        """Validate that built_in_tools does not contain FunctionDeclarations.
 
         Gemini-specific tools should only include tools that cannot be represented
         as FunctionDeclarations (e.g., GoogleSearch, CodeExecution, ComputerUse).
         Standard function calling tools should use the tools interface instead.
 
         Args:
-            gemini_tools: List of Gemini tools to validate
+            built_in_tools: List of Gemini tools to validate.
 
         Raises:
-            ValueError: If any tool contains function_declarations
+            ValueError: If any tool contains function_declarations.
         """
-        for tool in gemini_tools:
+        for tool in built_in_tools:
             # Check if the tool has function_declarations attribute and it's not empty
             if hasattr(tool, "function_declarations") and tool.function_declarations:
                 raise ValueError(
-                    "gemini_tools should not contain FunctionDeclarations. "
+                    "built_in_tools should not contain FunctionDeclarations. "
                     "Use the standard tools interface for function calling tools. "
-                    "gemini_tools is reserved for Gemini-specific tools like "
+                    "built_in_tools is reserved for Gemini-specific tools like "
                     "GoogleSearch, CodeExecution, ComputerUse, UrlContext, and FileSearch."
                 )
--- a/strands-py/tests/strands/models/test_gemini.py
+++ b/strands-py/tests/strands/models/test_gemini.py
@@ -1,5 +1,6 @@
 import logging
 import unittest.mock
+import warnings
 
 import pydantic
 import pytest
@@ -976,7 +977,7 @@ async def test_structured_output(gemini_client, model, messages, model_id, weath
     gemini_client.aio.models.generate_content.assert_called_with(**exp_request)
 
 
-def test_gemini_tools_validation_rejects_function_declarations(model_id):
+def test_built_in_tools_validation_rejects_function_declarations(model_id):
     tool_with_function_declarations = genai.types.Tool(
         function_declarations=[
             genai.types.FunctionDeclaration(
@@ -986,18 +987,99 @@ def test_gemini_tools_validation_rejects_function_declarations(model_id):
         ]
     )
 
-    with pytest.raises(ValueError, match="gemini_tools should not contain FunctionDeclarations"):
-        GeminiModel(model_id=model_id, gemini_tools=[tool_with_function_declarations])
+    with pytest.raises(ValueError, match="built_in_tools should not contain FunctionDeclarations"):
+        GeminiModel(model_id=model_id, built_in_tools=[tool_with_function_declarations])
 
 
-def test_gemini_tools_validation_allows_non_function_tools(model_id):
+def test__init__built_in_tools(model_id):
     tool_with_google_search = genai.types.Tool(google_search=genai.types.GoogleSearch())
 
-    model = GeminiModel(model_id=model_id, gemini_tools=[tool_with_google_search])
-    assert "gemini_tools" in model.config
+    with warnings.catch_warnings(record=True) as captured_warnings:
+        warnings.simplefilter("always")
+        model = GeminiModel(model_id=model_id, built_in_tools=[tool_with_google_search])
+
+    assert captured_warnings == []
+    assert model.config["built_in_tools"] == [tool_with_google_search]
+    assert "gemini_tools" not in model.config
 
 
-def test_gemini_tools_validation_on_update_config(model):
+def test__init__gemini_tools_normalizes_to_built_in_tools(model_id):
+    tool_with_google_search = genai.types.Tool(google_search=genai.types.GoogleSearch())
+
+    with pytest.warns(
+        DeprecationWarning,
+        match="gemini_tools is deprecated; use built_in_tools instead.",
+    ) as captured_warnings:
+        model = GeminiModel(model_id=model_id, gemini_tools=[tool_with_google_search])
+
+    assert captured_warnings[0].filename == __file__
+    assert model.config["built_in_tools"] == [tool_with_google_search]
+    assert "gemini_tools" not in model.config
+
+
+def test__init__built_in_tools_and_gemini_tools_raises(model_id):
+    built_in_tool = genai.types.Tool(google_search=genai.types.GoogleSearch())
+    deprecated_tool = genai.types.Tool(code_execution=genai.types.ToolCodeExecution())
+
+    with pytest.raises(ValueError) as error:
+        GeminiModel(
+            model_id=model_id,
+            built_in_tools=[built_in_tool],
+            gemini_tools=[deprecated_tool],
+        )
+
+    assert (
+        str(error.value)
+        == "built_in_tools and gemini_tools cannot be configured in tandem; gemini_tools is deprecated."
+    )
+
+
+def test_update_config_built_in_tools(model):
+    tool_with_google_search = genai.types.Tool(google_search=genai.types.GoogleSearch())
+
+    model.update_config(built_in_tools=[tool_with_google_search])
+
+    tru_tools = model._format_request_config(None, None, None).to_json_dict()["tools"]
+    exp_tools = [
+        {"function_declarations": []},
+        {"google_search": {}},
+    ]
+    assert tru_tools == exp_tools
+    assert model.config["built_in_tools"] == [tool_with_google_search]
+    assert "gemini_tools" not in model.config
+
+
+def test_update_config_gemini_tools_normalizes_to_built_in_tools(model):
+    tool_with_google_search = genai.types.Tool(google_search=genai.types.GoogleSearch())
+
+    with pytest.warns(
+        DeprecationWarning,
+        match="gemini_tools is deprecated; use built_in_tools instead.",
+    ) as captured_warnings:
+        model.update_config(gemini_tools=[tool_with_google_search])
+
+    assert captured_warnings[0].filename == __file__
+    assert model.config["built_in_tools"] == [tool_with_google_search]
+    assert "gemini_tools" not in model.config
+
+
+def test_update_config_built_in_tools_and_gemini_tools_raises(model):
+    built_in_tool = genai.types.Tool(google_search=genai.types.GoogleSearch())
+    deprecated_tool = genai.types.Tool(code_execution=genai.types.ToolCodeExecution())
+
+    with pytest.raises(ValueError) as error:
+        model.update_config(
+            built_in_tools=[built_in_tool],
+            gemini_tools=[deprecated_tool],
+        )
+
+    assert (
+        str(error.value)
+        == "built_in_tools and gemini_tools cannot be configured in tandem; gemini_tools is deprecated."
+    )
+
+
+def test_update_config_built_in_tools_validation_rejects_function_declarations(model):
     tool_with_function_declarations = genai.types.Tool(
         function_declarations=[
             genai.types.FunctionDeclaration(
@@ -1007,14 +1089,14 @@ def test_gemini_tools_validation_on_update_config(model):
         ]
     )
 
-    with pytest.raises(ValueError, match="gemini_tools should not contain FunctionDeclarations"):
-        model.update_config(gemini_tools=[tool_with_function_declarations])
+    with pytest.raises(ValueError, match="built_in_tools should not contain FunctionDeclarations"):
+        model.update_config(built_in_tools=[tool_with_function_declarations])
 
 
 @pytest.mark.asyncio
-async def test_stream_request_with_gemini_tools(gemini_client, messages, model_id):
+async def test_stream_request_with_built_in_tools(gemini_client, messages, model_id):
     google_search_tool = genai.types.Tool(google_search=genai.types.GoogleSearch())
-    model = GeminiModel(model_id=model_id, gemini_tools=[google_search_tool])
+    model = GeminiModel(model_id=model_id, built_in_tools=[google_search_tool])
 
     await anext(model.stream(messages))
 
@@ -1032,9 +1114,9 @@ async def test_stream_request_with_gemini_tools(gemini_client, messages, model_i
 
 
 @pytest.mark.asyncio
-async def test_stream_request_with_gemini_tools_and_function_tools(gemini_client, messages, tool_spec, model_id):
+async def test_stream_request_with_built_in_tools_and_function_tools(gemini_client, messages, tool_spec, model_id):
     code_execution_tool = genai.types.Tool(code_execution=genai.types.ToolCodeExecution())
-    model = GeminiModel(model_id=model_id, gemini_tools=[code_execution_tool])
+    model = GeminiModel(model_id=model_id, built_in_tools=[code_execution_tool])
 
     await anext(model.stream(messages, tool_specs=[tool_spec]))
 
--- a/strands-py/tests/strands/models/test_gemini.py
+++ b/strands-py/tests/strands/models/test_gemini.py
@@ -991,6 +991,24 @@ def test_built_in_tools_validation_rejects_function_declarations(model_id):
         GeminiModel(model_id=model_id, built_in_tools=[tool_with_function_declarations])
 
 
+def test__init__gemini_tools_validation_rejects_function_declarations(model_id):
+    tool_with_function_declarations = genai.types.Tool(
+        function_declarations=[
+            genai.types.FunctionDeclaration(
+                name="test_function",
+                description="A test function",
+            )
+        ]
+    )
+
+    with pytest.warns(
+        DeprecationWarning,
+        match="gemini_tools is deprecated and will be removed in v2.0.0. Use built_in_tools instead.",
+    ):
+        with pytest.raises(ValueError, match="built_in_tools should not contain FunctionDeclarations"):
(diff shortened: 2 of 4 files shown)


arielnabavian:
@strandly-the-agent review

strands-py/src/strands/models/gemini.py
```diff
@@ -632,25 +640,53 @@ async def structured_output(
...
+        Returns:
+            Gemini model configuration using the canonical built_in_tools field.
+
+        Raises:
+            ValueError: If built_in_tools and gemini_tools are both configured.
+        """
+        if "built_in_tools" in model_config and "gemini_tools" in model_config:
+            raise ValueError(
+                "built_in_tools and gemini_tools cannot be configured in tandem; gemini_tools is deprecated."
+            )
```
strandly-the-agent, line 658: 🟡 **The tandem check is presence-based, so it fires when nothing is actually double-configured**

`in` ignores the value, so a user mid-migration — one who added `built_in_tools=[...]` while their own factory still has a `gemini_tools: list | None = None` default — gets a hard error whose message says the opposite of what happened. Confirmed live at this head:

```
GeminiModel(model_id=..., built_in_tools=[gs], gemini_tools=None) -> ValueError: built_in_tools and gemini_tools cannot be configured in tandem; ...
GeminiModel(model_id=..., built_in_tools=[gs], gemini_tools=[])   -> ValueError: built_in_tools and gemini_tools cannot be configured in tandem; ...
```

All 7 falsy combinations raise, on `__init__` and on `update_config`. The file already uses truthiness for the same kind of dual-source config guard at `gemini.py:96-97` (`client` / `client_args`).

```suggestion
        if model_config.get("built_in_tools") and model_config.get("gemini_tools"):
            raise ValueError(
                "built_in_tools and gemini_tools cannot be configured in tandem. "
                "Use built_in_tools (recommended) or gemini_tools (deprecated), not both."
            )
```

The message tweak is the second half: the current wording doesn't tell the user *which* key to delete. `agent/a2a_agent.py:74-77` — the closest precedent, also a both-set guard on a deprecated alias — reads "Use client_config (recommended) or a2a_client_factory (deprecated), not both."

Verified: `ruff check` and `ruff format --check` pass on this exact snippet. It does need the `match=` strings updated in `test__init__built_in_tools_and_gemini_tools_raises` and `test_update_config_built_in_tools_and_gemini_tools_raises`.

strands-py/src/strands/models/gemini.py
```diff
@@ -632,25 +640,53 @@ async def structured_output(
...
+
+        if "gemini_tools" in model_config:
+            warnings.warn(
+                "gemini_tools is deprecated; use built_in_tools instead.",
```
strandly-the-agent: 🟡 **Deprecation message is missing the removal version this repo's own standard requires**

`team/FEATURE_LIFECYCLE.md:151-155` prescribes the message shape (`"... is deprecated and will be removed in v2.0.0. Use ... instead."`) and `:196-200` makes "clear deprecation timelines in all warnings" a stated Transparency principle. The repo's other `DeprecationWarning` sites comply — `tools/registry.py:172-176` ("will be removed in Strands SDK 2.0"), `agent/a2a_agent.py:80-85`, `agent/agent.py:875-880`. As written, a user learns the field is deprecated but not when it stops working.

```suggestion
                "gemini_tools is deprecated and will be removed in v2.0.0. Use built_in_tools instead.",
```

Verified ruff-clean and format-clean at the repo's 120-col limit (it has to stay on one line — `ruff format` collapses a two-line version). Needs the `match=` strings updated in the two `test_*_normalizes_to_built_in_tools` tests. If "v2.0.0" is more specific than the team wants to commit to, a docs link would satisfy the same requirement.

Unrelated but worth saying: `stacklevel=3` is *right* here, and it's a detail most people get wrong. I checked it attributes the warning to the user's own call site for both `__init__` and `update_config`, and your `captured_warnings[0].filename == __file__` assertion pins it (mutating it to `2` fails two tests).

strands-py/src/strands/models/gemini.py
```diff
@@ -632,25 +640,53 @@ async def structured_output(
...
+                DeprecationWarning,
+                stacklevel=3,
+            )
+            model_config["built_in_tools"] = model_config.pop("gemini_tools")
```
strandly-the-agent, line 666: 🟡 **`gemini_tools` disappears from `get_config()` — intentional, but undeclared**

`pop()` means an existing user who passed `gemini_tools=[...]` still gets a working model (the outgoing request is byte-identical to `main`, I A/B'd 18 scenarios), but anything that reads their config back changes behavior with no signal:

```
head 7a6a0a0f:  model.get_config()["gemini_tools"]      -> KeyError: 'gemini_tools'
base 1b671d6a:  model.get_config()["gemini_tools"]      -> [Tool(google_search=GoogleSearch())]
head 7a6a0a0f:  model.get_config().get("gemini_tools")  -> None          # silent, not even an error
```

The `.get()` row is the one that worries me: a subclass or wrapper doing `len(self.config.get("gemini_tools", []))` goes from 1 to 0 on upgrade, with no exception and no warning at the read site. And the warning that *is* emitted is invisible in the common case — CPython's default filters only surface `DeprecationWarning` from `__main__`, so an app that builds its model inside a factory module sees nothing (I verified both halves of that).

It's clearly deliberate — `test_gemini.py:1017` pins `"gemini_tools" not in model.config` — and "config only ever holds the canonical key" is a reasonable invariant to want. Two things give me pause:

1. `GeminiConfig` still declares `gemini_tools` (`gemini.py:63`), so `get_config()` is typed as possibly returning a key it now never can. mypy accepts `cfg["gemini_tools"]` on a `get_config()` result that's guaranteed to `KeyError` at runtime.
2. The only other deprecated provider config field in the repo does the opposite: `bedrock.py:103` + `:285-289` leaves `cache_prompt` in `self.config` and resolves it at use time.

Would it be simpler to resolve the alias on read and leave the caller's config exactly as they passed it?

```python
def _resolve_built_in_tools(self) -> list[genai.types.Tool]:
    """Return the configured built-in tools, honoring the deprecated gemini_tools alias.

    Returns:
        The configured built-in tools, or an empty list when none are configured.
    """
    return self.config.get("built_in_tools") or self.config.get("gemini_tools") or []
```

…with `_format_request_tools` (`gemini.py:281`, `:295-296`) calling it and `_normalize_built_in_tools` reduced to warn-and-reject without mutating. I tried that shape on a scratch copy: **59 of your 61 tests pass untouched**, the only two failures being the `model.config["built_in_tools"]`-after-alias assertions. It also keeps `GeminiModel(**model.get_config())` round-tripping either way.

If the team prefers the erasure as-is, that's fine by me — it just wants a line in the PR body (and ideally the changelog), because the body currently frames the alias as pure compatibility and this is a small, real behavior change for existing `gemini_tools` users. Flagged as a question for the maintainers in the summary rather than something for you to guess at.

strands-py/tests/strands/models/test_gemini.py
```diff
@@ -976,7 +977,7 @@ async def test_structured_output(gemini_client, model, messages, model_id, weath
     gemini_client.aio.models.generate_content.assert_called_with(**exp_request)
 
 
-def test_gemini_tools_validation_rejects_function_declarations(model_id):
+def test_built_in_tools_validation_rejects_function_declarations(model_id):
```
strandly-the-agent, line 980: 🟡 **The rename took away the only test that proved validation runs on the alias path**

At `main` this test was `test_gemini_tools_validation_rejects_function_declarations` — it proved a `Tool` carrying `function_declarations` is rejected when passed as `gemini_tools`. Renamed to the canonical field, that guarantee is now unproven. Mutation check: I made `_validate_built_in_tools` skip validation *only* for tools that arrived via the `gemini_tools` alias (canonical path still validating) and **all 61 tests passed**. Same story for the `update_config` variant, which lost `test_gemini_tools_validation_on_update_config`.

The behavior is correct today — `GeminiModel(model_id="...", gemini_tools=[bad_tool])` does raise, I checked — so this is a coverage hole, not a bug. It's the kind that lets a later refactor silently break the deprecated path, which is exactly the path with no other guardrail.

```python
def test__init__gemini_tools_validation_rejects_function_declarations(model_id):
    bad_tool = genai.types.Tool(
        function_declarations=[genai.types.FunctionDeclaration(name="test_function", description="A test function")]
    )
    with pytest.warns(DeprecationWarning):
        with pytest.raises(ValueError, match="built_in_tools should not contain FunctionDeclarations"):
            GeminiModel(model_id=model_id, gemini_tools=[bad_tool])
```

(plus the `update_config` equivalent).

While I'm here — the tests around this are better than most: the other 8 mutants I injected into the new logic were each killed by ≥2 tests, including the warning category and the `stacklevel`. Two nits not worth their own threads: `"gemini_tools" not in model.config` at `:1003` and `:1049` can't fail (those tests never pass `gemini_tools`), and nothing asserts the alias reaches the *outgoing request* end-to-end — the request-shape tests all use the canonical name.

strandly-the-agent (changes requested):
**Changes requested** — the rename itself is clean and genuinely well tested. Four should-fixes and one maintainer call; no 🔴 blockers: the request payload Gemini receives is byte-identical to `main` across 18 config scenarios, and 8 of the 9 mutants I injected into the new logic were killed by your tests.

| | where | finding |
|---|---|---|
| 🟡 | `gemini.py:666` | `pop()` erases `gemini_tools` from `get_config()` — intentional, but undeclared |
| 🟡 | `gemini.py:655` | presence-based tandem check raises on `gemini_tools=None` |
| 🟡 | `gemini.py:662` | deprecation message missing the removal version `FEATURE_LIFECYCLE.md` mandates |
| 🟡 | `test_gemini.py:980` | validation-via-alias lost its test in the rename (mutation-proven) |
| 🟡 | outside the diff | `google.mdx` + the integ fixture still teach `gemini_tools` — details below |

Process note: this changes a public config surface, its type surface and its error contract, so per `team/API_BAR_RAISING.md:72` it wants the **`api/needs-review`** label (a solo API reviewer, not the API meeting — it's a rename + deprecation, not a new primitive).

<details>
<summary>✅ What I verified</summary>

Reviewed at head `7a6a0a0f` vs merge-base `1b671d6a`, Python 3.12, `google-genai` from the `[gemini]` extra.

| | check | result |
|---|---|---|
| ✅ | `pytest tests/strands/models/test_gemini.py` | **61 passed** (repeatedly; also under `-W error::DeprecationWarning`) |
| ✅ | `ruff check`, `ruff format --check`, `mypy src/strands/models/gemini.py` | clean |
| ✅ | request payload A/B vs merge-base | **byte-identical** across 18 scenarios × 3 configurations (no built-ins, built-ins only, built-ins + toolSpecs, empty specs, duplicates, via `__init__` and `update_config`) — the `{"function_declarations": []}` and `None`-for-empty Vertex behavior is preserved exactly |
| ✅ | mutation testing on the new logic | 8/9 mutants killed by ≥2 tests each (normalize-noop, warning removed, error removed, `stacklevel=2`, keep-old-key, validator-noop, both skip-normalize paths, request drops built-ins); 1 survived → inline finding on `test_gemini.py:980` |
| ✅ | `get_config()['gemini_tools']` head vs base | `KeyError` at head, returns the tools at base → inline finding on `gemini.py:666` |
| ✅ | falsy tandem combinations | all 7 raise on both entry points → inline finding on `gemini.py:655` |
| ✅ | `stacklevel=3` | correct — lands on the user's call site for both `__init__` and `update_config` |
| ✅ | caller-dict aliasing, partial mutation after a raise, warning spam, `update_config` sequencing, validation bypass | **no defects** — see "checked and clean" below |
| ✅ | both suggested code fixes below | applied to a scratch copy: `ruff check` + `ruff format --check` clean (they do require updating 4 `match=` strings) |
| ⚠️ | `pytest tests_integ/models/test_model_gemini.py` end-to-end | couldn't run — collection needs `anthropic` (sandbox gap, unrelated to this PR). Evidence for the integ point below is the fixture body run verbatim. |

</details>

<details>
<summary>🟡 The "no documentation changes are required" claim (posted here, not inline — the files aren't in the diff)</summary>

Two places outside the diff still teach the name this PR deprecates:

- **`site/src/content/docs/user-guide/concepts/model-providers/google.mdx:179`, `:189`** — the Python tab says built-in tools "can be passed via the `gemini_tools` config option" and the sample uses `gemini_tools=[...]`, while the TS tab at `:201` already shows `builtInTools`. This is the canonical parity-facing page for exactly this field (its `sourceLinks` frontmatter points at `gemini.py`), it isn't auto-generated, so the docstring change never reaches it. I ran the page's snippet verbatim: it warns on the very first run, with default interpreter flags, because user scripts are `__main__`.
- **`strands-py/tests_integ/models/test_model_gemini.py:31`** — the `gemini_tool_model` fixture still constructs with `gemini_tools=[...]`. Nothing asserts the warning, and `tests_integ` isn't in `testpaths`, so CI stays green — but it's the SDK's own last first-party use of the old name, and it means the only "real API" coverage of this feature exercises the deprecated path. Under `-W error::DeprecationWarning` that fixture body now fails (it constructs fine at `1b671d6a`), which also makes the PR's "my changes generate no new warnings" checkbox not quite right.

Three lines total to flip both to `built_in_tools=`. Either way, the body's "No documentation changes are required; the compatibility behavior is documented on the Python API" is worth correcting — the docstring isn't user-facing the way that page is. If you'd rather do the doc page in a follow-up, saying that explicitly in the body is fine; `team/FEATURE_LIFECYCLE.md:180-186` does ask for docs updated with migration examples.

Small merge-time thing for whoever lands this: `site/src/content/changelog/` is generated from PR titles, so the current title won't mention the deprecation at all. Something like `feat(gemini): add canonical built_in_tools config, deprecate gemini_tools` would give release-notes readers the signal.

</details>

<details>
<summary>❓ Questions</summary>

**Blocking (a maintainer call, not the author's):**
1. Is dropping `gemini_tools` from `get_config()`/`self.config` an intentional contract decision, or should the alias resolve at read time so introspection stays backward-compatible? `team/FEATURE_LIFECYCLE.md:46-50` puts "existing code breaks without any user action" in the major-version column, and `:195` asks for a clear error or an automatic fallback — here the fallback is silent. Both answers are defensible; it just shouldn't be implicit. (Detail inline on `gemini.py:666`.)

**Non-blocking:**
2. Is `built_in_tools` meant to become the SDK-wide name for provider-native tools, or is it Gemini-only? Today no other Python provider has a dedicated field — `openai_responses` routes them through `params`, `bedrock` through `additional_request_fields`, `ollama`/`sagemaker` through `additional_args`. The name reads like a cross-provider convention while being one provider's field. Not a reason to hold this PR (the name matches TS *and* Google's own terminology, which `team/DECISIONS.md:168-182` supports), but a decision record now would save re-litigating it on the next provider.
3. `#3361`'s parent `#2498` implies more parity renames are coming. Worth extracting a shared `resolve_deprecated_config_alias(...)` helper on the *second* one rather than hand-rolling this three times — flagging so it isn't reinvented, not asking for it here.

</details>

<details>
<summary>Checked and clean — no action needed</summary>

Stating these because a review that only asserts isn't worth much:

- **Caller-dict mutation** — `_normalize_built_in_tools` mutates in place, but both call sites receive `**model_config`, which always materializes a fresh dict. `cfg = {...}; GeminiModel(**cfg)` leaves `cfg` untouched, and a second construction from the same dict works. No cross-instance contamination.
- **Partial mutation after a raise** — both raises happen before `self.config.update()`; config is bit-for-bit intact. No half-updated model.
- **Ordering** — `validate_config_keys` before normalization can't misfire (both keys are declared in `GeminiConfig`).
- **Warning hygiene** — fires once per call, correct frame, correct category; `catch_warnings` usage in the tests calls `simplefilter("always")`, so those assertions aren't vacuous.
- **Validation bypass / degenerate values** — post-construction `Tool` mutation, generators, bare `Tool`, `str`, `None` all behave *identically to `main`*. Real rough edges, all pre-existing; nothing this PR introduced, so nothing for you to fix here.
- **`_validate_gemini_tools` rename** — no external references; safe.
- **Coverage** — 100% line and branch on the changed region (98% file; the misses are pre-existing `_stream()` retry branches).
- **Scope** — no creep: every hunk maps to a `#3361` scope bullet, and the both-set `ValueError` matches what `arielnabavian` asked for in the issue thread (and fixes the "tools tools" typo in that suggested wording).

</details>

---
I'm `strandly-the-agent`, an experimental agent — this is solid work for a human to approve, not a gate, and pushback is welcome, especially on question 1.

arielnabavian:
Hi @ShitK thanks for picking this up. We appreciate the community support. 

Before we approve, do you mind addressing the comments from our review agent.

And additionally can you make sure we roll out documentation updates to gemini model with this ```site/src/content/docs/user-guide/concepts/model-providers/google.mdx```

ShitK:
Hi @arielnabavian, a friendly follow-up on this PR. I addressed the review feedback in commit `2c893cd2`, including the deprecation timeline, alias-path validation tests, integration fixture update, and Gemini documentation update. The latest `Trigger Strands Review`, `Build and Deploy Preview`, and `Run integration tests` jobs are still waiting for environment approval. When you have a chance, could you please approve those workflows and take another look? Thank you!

arielnabavian:
@strandly-the-agent review

strandly-the-agent (changes requested):
**Re-review of `2c893cd2`** — 3 of my 5 findings are fixed and verified; the remaining two are one committable suggestion and one maintainer decision. This is a fixes-only delta, so I re-verified it directly (repros and mutants re-run cold) rather than re-running the full review pipeline.

| finding | status |
|---|---|
| deprecation message missing removal version | ✅ **fixed** — exact suggested wording, both `match=` strings updated |
| alias-path validation lost its test | ✅ **fixed** — the mutant that survived all 61 tests last time is now killed by exactly the two new tests |
| docs + integ fixture still taught `gemini_tools` | ✅ **fixed** — `google.mdx` Python tab now teaches `built_in_tools` and notes the alias; integ fixture flipped |
| `gemini.py:655` — tandem check is presence-based | 🔓 **still open** — repro re-confirmed at `2c893cd2`: `GeminiModel(model_id=..., built_in_tools=[gs], gemini_tools=None)` still raises the misleading "configured in tandem" error. The `suggestion` block on the [existing thread](https://github.com/strands-agents/harness-sdk/pull/3441#discussion_r3659024744) at `gemini.py:655-658` still applies verbatim (it needs the two `*_and_gemini_tools_raises` `match=` strings updated with it). |
| `gemini.py:666` — `pop()` erases `gemini_tools` from `get_config()` | 🔓 **open, awaiting a call** — @arielnabavian this one is yours, not `ShitK`'s: keep the erasure (then it just needs a sentence in the PR body declaring the read-back change) or switch to read-time alias resolution per [the thread](https://github.com/strands-agents/harness-sdk/pull/3441#discussion_r3659024778) on `gemini.py:666`. Both are defensible; it shouldn't ship implicit. |

Once the `:655` suggestion is applied and the `:666` question has an answer either way, this is an approve from me.

<details>
<summary>✅ What I verified at 2c893cd2</summary>

Delta reviewed: `git diff 7a6a0a0f..2c893cd2` — one warning-string change in `gemini.py`, +2 tests and 4 `match=` updates in `test_gemini.py`, docs prose + sample in `google.mdx`, one-line kwarg flip in the integ fixture. No behavior change beyond the warning text; the request payload path is untouched.

| | check | result |
|---|---|---|
| ✅ | `pytest tests/strands/models/test_gemini.py` | **63 passed** (was 61; +2 alias-validation tests), also clean under `-W error::DeprecationWarning` |
| ✅ | mutation re-check | re-applied the alias-skips-validation mutant that survived at `7a6a0a0f` → now **2 failed** (exactly `test__init__gemini_tools_validation_rejects_function_declarations` and the `update_config` variant) |
| ✅ | new tests pin the delta | ran the new test file against the *previous* head's source → 4 failed (the message + alias-validation tests), so they aren't vacuous |
| ✅ | `ruff check` + `ruff format --check` | clean on all three touched Python files |
| ✅ | finding-2 repro | still raises on `gemini_tools=None` and `=[]` alongside `built_in_tools=[...]` |
| ✅ | finding-1 repro | `get_config()` after alias use still returns only `['built_in_tools', 'model_id']`; `.get("gemini_tools")` → `None` |
| ⚠️ | mypy | couldn't run — mypy 2.3.1 hits an internal error in this sandbox **on the previous head's file too**, so it's environmental, and the source delta is a single string literal (no type-surface change) |
| ⚠️ | integ file execution | not runnable here (needs `GOOGLE_API_KEY`); the change is a one-line kwarg rename whose behavior the unit tests cover |

Two small leftovers, neither blocking:
- The PR body still says *"No documentation changes are required"* — now stale in the opposite direction, since the PR **does** update `google.mdx`. Worth flipping that line and the two docs checkboxes.
- The `api/needs-review` label (`team/API_BAR_RAISING.md:72`) is still not applied — maintainer-side, flagged in my first review.

</details>

---
I'm `strandly-the-agent`, an experimental agent — solid work for a human to approve, not a gate.

strands-py/src/strands/models/gemini.py
ShitK: This normalization is intentional: the issue requires internal configuration to use the canonical `built_in_tools` field, so alias input is converted at write time and `get_config()` returns only `built_in_tools`. I have now declared that read-back behavior explicitly in the PR body. @arielnabavian, could you confirm that canonical-only read-back is the intended contract, or would you prefer read-time alias resolution for backward-compatible introspection?

strands-py/src/strands/models/gemini.py
ShitK: Thanks for calling out the `None` and empty-list cases. The presence-based check was intentional because #3361 asked that the two fields not be configured in tandem, and the implementation must not silently overwrite either value. I do not think the suggested condition can be applied verbatim: the following alias normalization would make `built_in_tools=[gs], gemini_tools=None` replace the canonical value with `None` (and likewise for `[]`). @arielnabavian, could you confirm whether an explicitly supplied `None` or empty list should count as not configured? If so, I can implement value-aware normalization that preserves the non-empty value and never silently overwrites it.

strands-py/src/strands/models/gemini.py
```diff
@@ -632,25 +640,53 @@ async def structured_output(
...
+
+        if "gemini_tools" in model_config:
+            warnings.warn(
+                "gemini_tools is deprecated and will be removed in v2.0.0. Use built_in_tools instead.",
+                DeprecationWarning,
+                stacklevel=3,
+            )
+            model_config["built_in_tools"] = model_config.pop("gemini_tools")
+
+        return model_config
```
arielnabavian: **Answering the open thread: yes, treat an explicit `None` or `[]` as "not configured".** Please make this value-aware.

You were right that the suggested one-liner can't go in verbatim. Gating only the `raise` would let `built_in_tools=[gs], gemini_tools=None` fall past the guard and then replace the real value with `None`. The gate has to go on the *adoption* too, so that the non-empty value always wins:

```suggestion
        if "gemini_tools" not in model_config:
            return model_config

        if model_config.get("built_in_tools") and model_config.get("gemini_tools"):
            raise ValueError(
                "built_in_tools and gemini_tools cannot be configured in tandem. "
                "Use built_in_tools (recommended) or gemini_tools (deprecated), not both."
            )

        warnings.warn(
            "gemini_tools is deprecated and will be removed in v2.0.0. Use built_in_tools instead.",
            DeprecationWarning,
            stacklevel=3,
        )
        gemini_tools = model_config.pop("gemini_tools")
        if not model_config.get("built_in_tools"):
            model_config["built_in_tools"] = gemini_tools

        return model_config
```

Two reasons I want it this way rather than keeping the strict check.

**The case it unblocks is the ordinary migration shape.** A user whose own factory signature carries `gemini_tools: list | None = None` and forwards `**kwargs` adds `built_in_tools=[...]`, and today gets a hard failure telling them they configured two things when they configured one. That is the exact population the alias exists to protect.

**And the strictness is not buying the property it looks like it buys.** The guard is already leaky one call over. At this head:

```
m = GeminiModel(model_id=..., built_in_tools=[google_search])
m.update_config(gemini_tools=[code_execution])
# -> config["built_in_tools"] == [code_execution]
# -> google_search is gone, with only a DeprecationWarning
```

So "never silently overwrite either value" holds inside a single call and not across two. Given that, I'd rather the guard fire on real conflicts and stay quiet on the falsy ones.

One fact that settles the `None` half specifically: `built_in_tools=None` already dies with a bare `TypeError: 'NoneType' object is not iterable` out of `_validate_built_in_tools` — on `main` and at this head. `None` is not a usable value for either field, so counting it as "not configured" throws nothing away.

<details>
<summary>Verified</summary>

Applied the block above at `2c893cd2`:

- `pytest tests/strands/models/test_gemini.py` -> **61 passed, 2 failed**. The only two failures are `test__init__built_in_tools_and_gemini_tools_raises` and `test_update_config_built_in_tools_and_gemini_tools_raises`, both on the exact-message assertion. So those two strings need updating and nothing else does.
- All four falsy combinations now resolve to the non-empty value instead of raising: `([gs], None)`, `([gs], [])`, `(None, [gs])`, `([], [gs])` -> `[google_search]` every time.
- A genuine tandem (`[google_search]` + `[code_execution]`) still raises.
- The alias alone still warns once, still normalizes, and still reaches the wire: `[{'function_declarations': []}, {'google_search': {}}]`.
- `update_config(built_in_tools=[])` still clears, so empty-list-means-clear survives.

Worth a new test for the falsy case while you're in here, since that's the behaviour being decided.

</details>

strands-py/src/strands/models/gemini.py
```diff
@@ -45,18 +46,20 @@ class GeminiConfig(BaseModelConfig, total=False):
             params: Additional model parameters (e.g., temperature).
                 For a complete list of supported parameters, see
                 https://ai.google.dev/api/generate-content#generationconfig.
-            gemini_tools: Gemini-specific tools that are not FunctionDeclarations
+            built_in_tools: Gemini-specific tools that are not FunctionDeclarations
                 (e.g., GoogleSearch, CodeExecution, ComputerUse, UrlContext, FileSearch).
                 Use the standard tools interface for function calling tools.
                 For a complete list of supported tools, see
                 https://ai.google.dev/api/caching#Tool
+            gemini_tools: Deprecated alias for built_in_tools.
```
arielnabavian: **Answering the other open thread: keep the write-time normalization. No code change needed.**

Canonical-only `self.config` is the invariant I want. It matches TS, which has no alias at all — just `builtInTools` at `strands-ts/src/models/google/types.ts:43` — and it keeps `GeminiModel(**model.get_config())` round-tripping whichever name went in. Your implementation stands.

The cost is real and I'm accepting it: code reading `config.get("gemini_tools")` goes from a list to `None` with no exception and no warning at the read site. But it should at least be findable, and this docstring line is the only place a reader learns the alias exists. Please make it say so:

```suggestion
            gemini_tools: Deprecated alias for built_in_tools. Accepted as input only; it is
                normalized on write, so get_config() always returns built_in_tools.
```

On the typing objection raised in that thread — `get_config()` being typed as possibly returning a key it can never hold — that one isn't fixable here, so don't chase it. The key has to stay declared on `GeminiConfig` or `validate_config_keys` starts warning on it and `Unpack` stops accepting it as a keyword argument. The root cause is one TypedDict serving as both the input and the output shape, which is a bigger change than this PR. A docstring is the right amount of effort.

strands-py/src/strands/models/gemini.py
```diff
@@ -107,10 +111,14 @@ def update_config(self, **model_config: Unpack[GeminiConfig]) -> None:  # type:
...
             **model_config: Configuration overrides.
+
+        Raises:
+            ValueError: If both `built_in_tools` and `gemini_tools` are configured, or if built-in tools contain
+                function declarations.
         """
-        # Validate gemini_tools if provided
-        if "gemini_tools" in model_config:
-            self._validate_gemini_tools(model_config["gemini_tools"])
+        model_config = self._normalize_built_in_tools(model_config)
```
arielnabavian: **New, and not your doing: `update_config` never validates config keys, so a mistyped `built_in_tools` is swallowed here.**

`gemini.py` is the only provider in `strands-py/src/strands/models/` whose `update_config` skips `validate_config_keys`. The other eleven all call it — `bedrock.py:242` is the shape. That gap is pre-existing, so it isn't a defect you introduced. But this PR is the rename that asks every existing Gemini user to retype this exact key, which makes now the moment it starts costing people afternoons.

Verified at `2c893cd2`:

```
GeminiModel(model_id=..., builtin_tools=[gs])   # missing underscore
#  -> warns: Invalid configuration parameters: ['builtin_tools']

m.update_config(builtin_tools=[gs])             # same typo
#  -> no warning at all
#  -> config keys become ['builtin_tools', 'model_id']
#  -> tools in the outgoing request: None
```

Same mistake, caught in the constructor and silent in `update_config`, and the built-in tools simply never reach Gemini. A user migrating off `gemini_tools` has no signal that their search tool stopped working.

```suggestion
        validate_config_keys(model_config, GeminiModel.GeminiConfig)
        model_config = self._normalize_built_in_tools(model_config)
```

Checked with that line added: 63 tests still pass, and the `gemini_tools` alias still goes through clean — one `DeprecationWarning`, no invalid-key warning — because both keys are declared on `GeminiConfig`.

If you'd rather keep this PR to the rename, say so and I'll open a follow-up myself. I'd just rather not ship the rename without it.

Issue #2085, opened 2026-04-07
[BUG] force slidingWindowConversationManager to use user message. Currently, it could start with an assistant message.


vedashreevinay:
### Checks

- [x] I have updated to the lastest minor and patch version of Strands
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched [./issues](./issues?q=) and there are no duplicates of my issue

### Strands Version

1.30.0

### Python Version

3.13

### Operating System

macOS26.3.1

### Installation Method

pip

### Steps to Reproduce

1. Create an agent with Nova 2 Lite as the model using Amazon Bedrock provider. 
2. Set the slidingwindow size to be lower to hit the window limit earlier. 
3. Generate enough conversations with the agent to fill up the sliding window size 
4. User will hit a ValidationException: A conversation must start with a user message

### Expected Behavior

Even after the sliding window size is reached, the context sent to the model should always start with a user message. No ValidationExceptions. 

### Actual Behavior

ValidationException: A conversation must start with a user message

mkmeral:
https://github.com/strands-agents/sdk-python/pull/2087

PR #2808, opened 2026-06-15
feat: port agentic context management to python

notowen333:
## Description

Ports the agentic context-management feature from `strands-ts` (#2754) to `strands-py`.

### Motivation

The TypeScript SDK shipped an `"agentic"` context strategy that hands this control to the model: it injects compression tools the model can call, and feeds it live token-usage telemetry so it can decide *when* to compress. This change brings that capability to the Python SDK at parity.

### Public API Changes

`context_manager` now accepts `"agentic"` in addition to `"auto"`:

```python
from strands import Agent

# The model manages its own context via injected tools.
agent = Agent(context_manager="agentic")
```

In agentic mode the agent gains three model-callable tools and a token-usage middleware:

- `summarize_context` — fold the oldest messages into a model-generated summary
- `truncate_context` — drop the oldest messages outright
- `pin_context` — pin/unpin messages so they survive compression

Before each model call, a `<context-status>` block is appended to the conversation so the model can see how much of the window it has used:

```
<context-status>
<used>50,000 / 200,000 tokens (25.0%)</used>
<remaining>~150,000 tokens</remaining>
</context-status>
```

Unlike `"auto"`, agentic mode does **not** enable proactive compression — the model drives compression itself, and the underlying `SummarizingConversationManager` remains only as a reactive overflow safety net. It also offloads tool results at a higher threshold (8k vs 1.5k tokens), since the model is responsible for managing its own inline context.

### Implementation notes

Two structural decisions worth surfacing for review:

- **Shared compression primitives.** Mirroring the TS layout, the tool-pair-safe split/trim helpers, summary generation, and pin utilities now live in a `conversation_manager/compression/` package. The existing summarizing and sliding-window managers were refactored onto these shared helpers. This refactor is behavior-preserving — notably, the sliding-window manager keeps its Python-only "fall back to a tool-pair boundary" path, which has no TS equivalent and was deliberately *not* collapsed into the shared `find_valid_trim_point`.
- **Fully internal surface.** The pin utilities and agentic tools/middleware are intentionally *not* exported from any public `__init__.py`; they're reachable only by full module path. The sole re-export is the `DEFAULT_SUMMARIZATION_PROMPT` constant (kept for backward compatibility on its original module).

## Related Issues

Ports the feature added to `strands-ts` in #2754.

## Documentation PR

None in this PR — docs for the agentic strategy are tracked separately.

## Type of Change

New feature

## Testing

[...]

PR #3447, opened 2026-07-24
fix(conversation): trim at complete tool pairs

### LINKED ISSUE #3165: [PORT] Sliding-window tool-pair trim fallback (Py→TS)
> ⚠️ Filed by Strandly as part of the #2498 parity audit. Source feature verified merged + unit-tested in `strands-py`. Small, well-scoped. Agent-produced — confirm scope. (Add the `port` label — I can't apply it via API.)

**Parent:** #2498

## Source SDK
`strands-py`

## Target SDK
`strands-ts`

## Related Files
```
strands-py/src/strands/agent/conversation_manager/sliding_window_conversation_manager.py
strands-py/src/strands/agent/conversation_manager/compression/context_compression.py
strands-py/tests/strands/agent/test_conversation_manager.py
```

## What needs to be migrated
A fallback trim search so tool-heavy histories stay trimmable. Python's `_find_tool_pair_trim_point(messages, start_index) -> int | None` (`sliding_window_conversation_manager.py:276`): when `find_valid_trim_point` finds no plain-user boundary (returns `len(messages)`), it searches for the **first `assistant(toolUse)` immediately followed by `user(toolResult)`** at/after `start_index` and trims there (providers accept a complete toolUse/toolResult pair as a valid continuation). Invoked from `reduce_context` (`:238`), which has a three-way branch: fallback found → trim; fallback `None` + reactive → raise `ContextWindowOverflowException`; fallback `None` + proactive → warn + no-op.

TS `sliding-window-conversation-manager.ts` `_reduceContext` (`:234`) only calls `findValidTrimPoint` and bails (`return false`) if none — **no tool-pair fallback**. Preserve the "prefer plain-user boundary first, tool-pair only as fallback" ordering.

## TypeScript landing points
- Add private `_findToolPairTrimPoint(messages, startIndex): number | undefined` in `sliding-window-conversation-manager.ts`, next to `_findOldestMessageWithToolResults` (`:448`).
- Wire into `_reduceContext` (`:217-266`) before the `return false` bail (`:238-245`): if defined index found, set `trimIndex` and fall through to removal (`:247-265`).
- Block checks: `block.type === 'toolUseBlock'` / `'toolResultBlock'` (as in `context-compression.ts:99,106`).

## Gotchas / open questions
- `int | None` → `number | undefined` (not `null`).
- **Do NOT add an explicit `throw` in `_reduceContext`.** TS's overflow-recovery raises centrally in the base `AfterModelCallEvent` hook (`conversation-manager.ts:172`): `reduce` returning `false` lets `ContextWindowOverflowError` propagate. So the fallback branch returns `true` when a pair is found, else falls through to `return false`. Preserve behavior, not the raise mechanism.
- Guard the `windowSize === 0` remove-all path so the fallback doesn't hijack it (`:240`).
- Dict-membership (`"toolUse" in content`) → discriminated-union `.type` checks; keep the `messages[index+1]` bounds check.

- Merged: ✅  ·  Unit tests: ✅ (`test_conversation_manager.py:234` fallback fires; `:269` prefers plain-user)  ·  Integ: ❌ none  ·  Docs: internal heuristic, no user-facing surface (no doc needed). Port MUST add TS unit tests mirroring both Python cases.

## Confirmation
- [x] The change is complete in the source SDK (merged + unit-tested); every related artifact is listed above.

March-77:
## Description

Adds a tool-pair fallback to the TypeScript sliding-window conversation manager.

When no plain user trim point exists, the manager searches for a complete
`assistant(toolUse) -> user(toolResult)` pair and retains a validated user
message immediately before it. Pinned prefixes are accepted only when they
remain user-first and legally alternating.

This deliberately differs from the Python implementation, which trims directly
to the assistant tool-use boundary. TypeScript retains the user anchor because
supported model providers require a user-first conversation and valid role
alternation.

The manager logs both successful fallback selection and cases where a complete
pair is rejected because no safe user anchor exists. Existing plain-user
boundaries and `windowSize: 0` behavior remain unchanged.

## Related issues

Fixes #3165

## Testing

- `npx vitest run --project unit-node src/conversation-manager/__tests__/sliding-window-conversation-manager.test.ts`
  — 65 passed, no type errors
- `npm run type-check`
- `npm run lint`
- `npm run format:check`

`hatch run prepare` is not applicable to this TypeScript-only change.

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/strands-agents/harness-sdk/blob/main/CONTRIBUTING.md) guide.
- [x] I have performed a self-review of my code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no new warnings.
- [x] I have added tests that prove my fix is effective.
- [x] New and existing targeted unit tests pass locally.
- [ ] I have run `hatch run prepare`.

`hatch run prepare` is not applicable to this TypeScript-only change.


JackYPCOnline:
@strandly-the-agent can you review the implementation

strands-ts/src/conversation-manager/sliding-window-conversation-manager.ts
```diff
@@ -436,6 +443,29 @@ export class SlidingWindowConversationManager extends ConversationManager {
...
+   *
+   * @param messages - The conversation message history.
+   * @param startIndex - The index to begin searching from.
+   * @returns The assistant message index that starts the pair, or undefined if none exists.
+   */
+  private _findToolPairTrimPoint(messages: Message[], startIndex: number): number | undefined {
+    for (let index = startIndex; index < messages.length; index++) {
+      const message = messages[index]!
+      const nextMessage = messages[index + 1]
+      const hasToolUse = message.role === 'assistant' && message.content.some((block) => block.type === 'toolUseBlock')
```
strandly-the-agent, line 470: 🔴 **This is the line that re-opens #2085.**

`hasToolUse` requires `role === 'assistant'`, so `_findToolPairTrimPoint` can only ever return an **assistant** index — whereas `findValidTrimPoint` (`context-compression.ts:85-113`) only ever returns a **user** index. The caller removes `[0, trimIndex)`, so that assistant message becomes the new **first** message of the history.

That user-first property is not incidental: it's precisely what #2087 added to close #2085. See the review summary for the full chain — short version, `find_valid_trim_point`'s documented contract still says *"1. Be a user message (required by most model providers)"*, and the Python fallback you're porting bypasses it.

<details>
<summary>Reproduced cold, public <code>Agent</code> API, head vs base (fake model, no network, run twice)</summary>

**Proactive path — no overflow error anywhere, just routine `windowSize: 4` enforcement.** This is the one I'd fix first: a reactive 400 is at least visible, whereas this silently leaves `agent.messages` invalid, and a `sessionManager` would persist it.

```
HEAD: after invocation #1 -> agent.messages: 0:a(tu:id-3) 1:u(tr:id-3) 2:a(text)   [role = assistant]
      invocation #2, first model call payload: 0:a(tu:id-3) 1:u(tr:id-3) 2:a(text) 3:u(text)
      ASSISTANT-FIRST PAYLOAD ON A NORMAL (non-overflow) CALL: true

BASE: "window_size=<4>, messages=<8> | unable to trim conversation context, no valid trim point found"
      agent.messages[0].role = user ; ASSISTANT-FIRST: false
```

**Reactive path (`ContextWindowOverflowError`)** — the trimmed array reaches the provider verbatim; `bedrock.ts` / `anthropic.ts` `_formatMessages` map roles straight through and nothing in `strands-ts/src` asserts on `messages[0]`:

```
HEAD: post-reduce payload: [0] assistant toolUse(id-2), [1] user toolResult(id-2), ...
      ASSISTANT-FIRST PAYLOAD REACHED PROVIDER: true
BASE: payloads sent AFTER reduce(): 0  (declined to trim, original error propagated)
```

**`pinFirst: 2, windowSize: 2` — outright alternation break, no provider-family escape hatch.** The removal loop skips pinned indices (`:256-259`), so a pinned prefix ending in `assistant` gets spliced directly onto the fallback's `assistant(toolUse)` head:

```
before  0:u(text) 1:a(text) 2:u(text) 3:a(tu:1) 4:u(tr:1) 5:a(tu:2) 6:u(tr:2)
HEAD    0:u*(text) 1:a*(text) 2:a(tu:2) 3:u(tr:2)   -> INVALID ADJACENT_SAME_ROLE@1,2=assistant
BASE    unchanged, retry=false, INVALID <none>
```

Notably `pinFirst: 1` (pinned prefix ends in `user`) comes out **valid** — which I think hints at the shape of the fix.

</details>

To be clear about what I'm *not* saying: base is worse in an important way — it never trims tool-heavy histories at all, so they grow unbounded. You're fixing a real bug and I'd like this to land. My question is only whether we inherit Python's boundary choice or take the conservative one this repo already uses in `agentic-context.ts:159-163`. Would keeping the leading user message and trimming the span after it work here? That would give you `user → assistant(toolUse) → user(toolResult)` — user-first, correctly alternating, and still trimming the bulk.

strands-ts/src/conversation-manager/sliding-window-conversation-manager.ts
```diff
@@ -233,7 +233,14 @@ export class SlidingWindowConversationManager extends ConversationManager {
     // Try to trim messages when tool result cannot be truncated anymore
     // If the number of messages is less than the window_size, then we default to 2, otherwise, trim to window size
     const startIndex = messages.length <= this._windowSize ? 2 : messages.length - this._windowSize
-    const trimIndex = findValidTrimPoint(messages, startIndex)
+    let trimIndex = findValidTrimPoint(messages, startIndex)
+
+    if (trimIndex === messages.length && this._windowSize > 0) {
+      const toolPairTrimIndex = this._findToolPairTrimPoint(messages, startIndex)
+      if (toolPairTrimIndex !== undefined) {
+        trimIndex = toolPairTrimIndex
```
strandly-the-agent: 🟡 The fallback succeeds silently — there's not a single `logger.*` call on the new path, while Python logs at the equivalent site (`sliding_window_conversation_manager.py:240-244`). Given this branch only fires in the hard-to-reproduce tool-heavy case, losing that breadcrumb hurts field debugging.

```suggestion
        logger.debug(
          `window_size=<${this._windowSize}>, trim_index=<${toolPairTrimIndex}> | no plain user trim point, falling back to complete tool pair`
        )
        trimIndex = toolPairTrimIndex
```

strands-ts/src/conversation-manager/sliding-window-conversation-manager.ts
```diff
@@ -436,6 +443,29 @@ export class SlidingWindowConversationManager extends ConversationManager {
...
+  /**
+   * Find the first complete tool use/result pair at or after the starting index.
+   *
+   * @param messages - The conversation message history.
+   * @param startIndex - The index to begin searching from.
+   * @returns The assistant message index that starts the pair, or undefined if none exists.
+   */
+  private _findToolPairTrimPoint(messages: Message[], startIndex: number): number | undefined {
+    for (let index = startIndex; index < messages.length; index++) {
+      const message = messages[index]!
```
strandly-the-agent: ⚪ nit: the sibling `findValidTrimPoint` guards instead of asserting (`context-compression.ts:89-90` does `const message = messages[trimIndex]; if (!message) break`). I tried and couldn't construct a reachable input here, so this is consistency-only — you already guard `nextMessage` with `?.` on the next line.

```suggestion
      const message = messages[index]
      if (!message) break
```

strands-ts/src/conversation-manager/__tests__/sliding-window-conversation-manager.test.ts
```diff
@@ -971,6 +971,99 @@ describe('SlidingWindowConversationManager', () => {
...
+      expect(mockAgent.messages).toEqual(expectedMessages)
+      expect(mockAgent.messages[0]!.content[0]).toEqual({
+        type: 'toolUseBlock',
+        name: 'getTree',
+        toolUseId: 'id-3',
+        input: {},
+      })
+    })
+
+    it('prefers a plain user trim point over an earlier complete tool pair', async () => {
```
strandly-the-agent: 🟡 This test doesn't exercise what its name claims. The tool pair sits at indices 1–2, which is *before* `startIndex` (`= 6 - 4 = 2`), so `_findToolPairTrimPoint` never sees it — and `findValidTrimPoint` alone already returns the plain point, meaning the fallback is never invoked at all. It passes identically with the whole new branch deleted.

Demonstrated by mutation testing: deleting the `trimIndex === messages.length` gate at `:238` — which should make the fallback always override the plain point — **survives this test untouched**, and is only caught incidentally by two unrelated pre-existing tests.

The fix is to put a *reachable* pair inside `[startIndex, length)` alongside a later plain point, so "prefers plain over tool-pair" is actually the thing under test:

```ts
it('prefers a plain user trim point over a reachable complete tool pair later in range', async () => {
  const manager = new SlidingWindowConversationManager({ windowSize: 4, shouldTruncateResults: false })
  const messages = [
    new Message({ role: 'user', content: [new TextBlock('First')] }),
    new Message({ role: 'assistant', content: [new TextBlock('Response A')] }),
    new Message({ role: 'user', content: [new TextBlock('Plain user message')] }),
    new Message({
      role: 'assistant',
      content: [new ToolUseBlock({ name: 'lookup', toolUseId: 'id-1', input: {} })],
    }),
    new Message({
      role: 'user',
      content: [new ToolResultBlock({ toolUseId: 'id-1', status: 'success', content: [new TextBlock('Result')] })],
    }),
    new Message({ role: 'assistant', content: [new TextBlock('Final')] }),
  ]
  const mockAgent = createMockAgent({ messages })
  const expectedMessages = messages.slice(2)

  await triggerContextOverflow(manager, mockAgent, new ContextWindowOverflowError('Context overflow'))

  expect(mockAgent.messages).toEqual(expectedMessages)
  expect(mockAgent.messages[0]!.content[0]).toEqual({ type: 'textBlock', text: 'Plain user message' })
})
```

Verified: this passes on your branch and fails under the gate-removal mutation. Worth noting the weakness is inherited from the Python test you mirrored (`test_conversation_manager.py:269` places the pair the same way), so it's parity rather than a TS regression — but since TS is the newer artifact it's the cheaper place to fix it.

strands-ts/src/conversation-manager/__tests__/sliding-window-conversation-manager.test.ts
```diff
@@ -971,6 +971,99 @@ describe('SlidingWindowConversationManager', () => {
       expect(mockAgent.messages[0]!.content[0]!).toEqual({ type: 'textBlock', text: 'Message 2' })
     })
 
+    it('falls back to a complete tool pair when no plain user trim point exists', async () => {
```
strandly-the-agent, line 1008: 🟡 Two coverage gaps in the new block — both cheap, and worth closing whichever way the design question lands.

**1. Both guards in `_findToolPairTrimPoint` can be deleted with the full 56-test suite still green.** I mutation-tested each: removing `role === 'assistant'`, and separately removing the check that the *following* message actually has a `toolResultBlock`, both survive. Repro for the second — `windowSize: 4`, messages `[user "M0", assistant[toolUse id-1], user[toolUse id-2], assistant "R", assistant "Final"]`: correct code returns `false` unchanged, the mutant returns `true` and leaves an orphaned assistant `toolUse` as the first message. Guards this load-bearing should each fail a test when removed. (The tests *do* catch the search-start index and an `index + 1` off-by-one — those are covered.)

**2. The decline path is never exercised.** v8 data for the `if` at `:247` shows `hits = [1, 0]`: no test in this file covers "outer gate fired, but the fallback found nothing either". This one is verified passing on your branch and moves the branch to `[1,1]`:

```ts
it('logs a warning and returns false when neither a plain trim point nor a tool pair exists', async () => {
  const manager = new SlidingWindowConversationManager({ windowSize: 4, shouldTruncateResults: false })
  const messages = [
    new Message({ role: 'user', content: [new TextBlock('Message 1')] }),
    new Message({ role: 'assistant', content: [new TextBlock('Response 1')] }),
    new Message({
      role: 'user',
      content: [new ToolResultBlock({ toolUseId: 'orphan', status: 'success', content: [new TextBlock('x')] })],
    }),
    new Message({
      role: 'assistant',
      content: [new ToolUseBlock({ name: 'lookup', toolUseId: 'id-1', input: {} })],
    }),
    new Message({ role: 'assistant', content: [new TextBlock('Final')] }),
  ]
  const mockAgent = createMockAgent({ messages })
  const before = [...mockAgent.messages]

  const result = manager.reduce({
    agent: mockAgent,
    model: {} as Model,
    error: new ContextWindowOverflowError('Context overflow'),
  })

  expect(result).toBe(false)
  expect(mockAgent.messages).toEqual(before)
})
```

And once the blocker is settled, please cover `pinFirst` × fallback in both directions — the safe `pinFirst: 1` case and the currently-broken `pinFirst: 2` one. Neither is covered today.

⚪ Also, minor: every sibling test in this `reduceContext - tool pair validation` block carries an inline comment walking the `trimIndex`/`startIndex` arithmetic; these two don't, even though that arithmetic is the whole point. Unprompted credit though — dispatching through the real `AfterModelCallEvent` hook instead of poking `_reduceContext` directly is the right call, and `toEqual` on the whole array is stronger than several of the neighbours.

strandly-the-agent (changes requested):
**TL;DR — changes requested, but the direction is right.** All six of #3165's acceptance criteria are met, and the fallback fixes a genuine bug: on base, tool-heavy histories are never trimmed *at all* — my repro grew to **16 messages under `windowSize: 4`**. What needs a decision is the *boundary*: `_findToolPairTrimPoint` can only ever land on an `assistant(toolUse)` message, which re-opens the failure mode this repo already closed in #2087. Reproduced cold through the public `Agent` API on the reactive **and** the proactive path, plus one `pinFirst` shape that breaks role alternation for *every* provider family. Then 4 cheap test/logging gaps. Details inline.

One thing I should own up front: **#3165 is my own issue, and it didn't flag this hazard** — it told you to "preserve behavior", which is exactly what you did. That gap is mine, not yours.

<details>
<summary><b>✅ Verification ledger</b></summary>

Reviewed `pr3447` @ `f3136c7444a5be2267ad4518c8b1bea0657448f2` vs base `1b671d6a1`, in dedicated `git worktree` checkouts of each.

| Check | Result |
|---|---|
| `npx vitest run --project unit-node` (sliding-window + context-compression) | ✅ **83/83 passed** |
| Scoped file / full suite (your claims) | ✅ **56/56** and **3,869 passed** — exact match. Coverage came out 93.63% vs the 93.64% you quoted; that's run-to-run noise, not a discrepancy |
| `npx eslint` on both changed files | ✅ clean |
| `npx tsc --noEmit -p src/tsconfig.json` | ✅ clean (the `npm run type-check` failures are pre-existing `test/integ/**` issues needing a prior `npm run build` — not yours) |
| Assistant-first payload reaches the provider, reactive path | 🔴 confirmed |
| Same, **proactive** path with no overflow anywhere | 🔴 confirmed |
| `pinFirst: 2` → two adjacent `assistant` messages | 🔴 confirmed (base: unaffected) |
| No-progress / false-`true` / infinite-retry; `windowSize` 0·1·−3·2.5·1e9; 0–2-message histories; `shouldTruncateResults` interaction; pinned-pair splitting; type safety | ✅ all clean — attacked and survived |
| Live Bedrock call | ⚠️ **not** run (no network in this pass) — provider-family behavior below is from external evidence, not my own API call |

Repros are public-`Agent`-API only, fake models, no network; each run twice cold and deterministic. No `needs-api-review` label needed — `_findToolPairTrimPoint` is `private` and nothing exported changed.

</details>

<details>
<summary><b>🔴 The regression chain (git archaeology)</b></summary>

1. **#2085** (Apr 7, `vedashreevinay`) — Nova 2 Lite + small sliding window → `ValidationException: A conversation must start with a user message`. Expected: *"the context sent to the model should always start with a user message."*
2. **#2087** (`65b06d993`, Apr 8, yours, `JackYPCOnline`) fixed it by adding to the Python trim loop:
   ```python
   # Find the next valid trim point that:
   # 1. Starts with a user message (required by most model providers)
   if messages[trim_index]["role"] != "user":
       trim_index += 1
       continue
   ```
   Before #2087 an `assistant(toolUse)` + `user(toolResult)` pair *was* an acceptable trim point. #2087 deliberately forbade it.
3. That contract survives today — `context_compression.py:110`: *"A valid trim point must: 1. Be a user message (required by most model providers)"*, mirrored in TS at `context-compression.ts:85-113` and guarded by the existing test `sliding-window-conversation-manager.test.ts:1084` (*"skips assistant message to ensure trimmed conversation starts with user"*).
4. **#2808** (`fbfd898c1`, Jun 16) then added Python's `_find_tool_pair_trim_point`, which **bypasses** that requirement — incidentally, as part of the agentic-context extraction, not as a deliberate revisit. Its own docstring concedes: *"This has no equivalent in `find_valid_trim_point`."*
5. This PR faithfully ports the bypass to TS. **The port is correct; the thing being ported is what I'd question.**

**Provider reality, stated precisely** (and this corrects an over-broad comment in our own tree at `agentic-context.ts:38-39`): Bedrock Converse rejects assistant-first for **Nova, Llama, Mistral, Cohere, AI21, Writer**, but **accepts** it for **Anthropic and Qwen** (live-recorded 200 in pydantic-ai#6255; rejection repro in pydantic-ai#2992 on `nova-lite`). AWS doesn't document the rule. So our default Anthropic users likely *won't* hit the assistant-first instance — Nova users will, which is exactly #2085 again. Your own #2087 wording is the accurate framing: *"Many model providers (including Amazon Bedrock with Nova)…"*. **The `pinFirst` adjacency instance has no such escape hatch** — Bedrock rejects two adjacent same-role messages regardless of family.

</details>

<details>
<summary><b>❓ Questions</b></summary>

- **(blocking, design)** `agentic-context.ts:159-163` already chose the conservative path — it *declines* to trim rather than produce an assistant-first history. Should `_findToolPairTrimPoint` adopt the same discipline, or better, keep the leading user message and trim the span *after* it (the `collectPreserved` shape) so you still get the trim *and* a valid `user → assistant(toolUse) → user(toolResult)` head? At minimum the alternation-breaking `pinFirst` case needs handling, since no provider family tolerates it. This decides the shape of the fix — happy to prototype whichever you prefer.
- **(non-blocking)** The `role === 'assistant'` check at `:457` has no Python equivalent. Evidence says **keep it**: on a `user` message holding both a `toolResult` and a `toolUse`, Python trims and produces an orphaned `toolResult` while TS correctly refuses — so TS is stricter *and better* here. Flagging only to confirm it was deliberate, and to argue against "fixing" it toward parity.
- Worth a follow-up issue against **both** SDKs for the underlying design (and for the over-broad `agentic-context.ts:38-39` comment)? This is bigger than a port task, and Python ships the same hazard today.

</details>

Good work overall — the port itself is clean, the tests mirror the Python cases 1:1, and your PR description checked out line for line, which made this much faster to review. Flagging this for a human to arbitrate rather than treating it as a gate: the call on whether to inherit Python's trade-off is `JackYPCOnline`'s, not mine.

JackYPCOnline:
@strandly-the-agent Review the newest version find any bugs/errors

strands-ts/src/conversation-manager/sliding-window-conversation-manager.ts
```diff
@@ -233,7 +233,21 @@ export class SlidingWindowConversationManager extends ConversationManager {
...
+    if (trimIndex === messages.length && this._windowSize > 0) {
+      const toolPairTrimIndex = this._findToolPairTrimPoint(messages, startIndex)
+      if (toolPairTrimIndex !== undefined) {
+        fallbackUserIndex = this._findToolPairUserAnchor(messages, toolPairTrimIndex)
+        if (fallbackUserIndex !== undefined) {
+          logger.debug(
+            `window_size=<${this._windowSize}>, trim_index=<${toolPairTrimIndex}>, user_anchor_index=<${fallbackUserIndex}> | no plain user trim point, falling back to complete tool pair`
+          )
+          trimIndex = toolPairTrimIndex
+        }
```
strandly-the-agent, line 253: 🟡 When the pair is found but the anchor check rejects it, this branch falls through silently and lands on the pre-existing warning at `:252-257`, which says `unable to trim conversation context, no valid trim point found`. That's misleading in the one case that matters most for debugging this feature: a valid tool pair **was** found, and only the anchor validation declined it. Anyone reading that log will go looking for a missing trim point that exists.

This is also the log line that fires on every reduction in the `pinFirst: 2` case (8/8 invocations in my repro), so it's the first thing someone hitting that gap will see.

```suggestion
        } else {
          logger.debug(
            `window_size=<${this._windowSize}>, trim_index=<${toolPairTrimIndex}> | complete tool pair found but no valid user anchor, declining fallback`
          )
        }
```

strands-ts/src/conversation-manager/sliding-window-conversation-manager.ts
```diff
@@ -436,6 +450,87 @@ export class SlidingWindowConversationManager extends ConversationManager {
...
+
+    return undefined
+  }
+
+  /**
+   * Find a user message that can remain immediately before a fallback tool pair.
+   *
+   * Pinned messages must already form a user-first alternating prefix that ends
+   * with a user message. Without a pinned prefix, the most recent plain user
+   * message before the tool pair is retained as the conversation anchor.
```
strandly-the-agent, line 487: 🟡 This is the one thing I'd most like changed before merge, and it's one sentence.

The docblock explains *what* the anchor does but not that it's a **deliberate divergence from the port source**. Python's `_find_tool_pair_trim_point` (`strands-py/src/strands/agent/conversation_manager/sliding_window_conversation_manager.py:276-299`) has no anchor concept — it trims straight to the `assistant(toolUse)` boundary. I grepped the whole diff and the PR body for `python|diverg|parity|mirror`: nothing.

That matters here more than usual, because the test that mirrors Python's now asserts the **opposite** outcome. Python's `test_..._falls_back_to_tool_pair_boundary` asserts `messages[0]["role"] == "assistant"` (length 3); the TS test with the mirroring name and the same `getDiff`/`getFile`/`getTree` fixture (`:1007-1018`) asserts `['user','assistant','user','assistant']` (length 4). A future contributor diffing the two suites will read that as a porting bug and "fix" it back — straight back into #2085.

```suggestion
   * message before the tool pair is retained as the conversation anchor.
   *
   * Deliberately diverges from the Python port source (`sliding_window_conversation_manager.py`
   * `_find_tool_pair_trim_point`), which trims directly to the `assistant(toolUse)` boundary. TS
   * retains a validated user anchor so the trimmed history still starts with a user message, which
   * many providers require (see #2085 and its fix in #2087). Do not "restore parity" here without
   * reading that history first.
```

strands-ts/src/conversation-manager/sliding-window-conversation-manager.ts
```diff
@@ -436,6 +450,87 @@ export class SlidingWindowConversationManager extends ConversationManager {
...
+        if (pinnedMessages[index]!.role === pinnedMessages[index - 1]!.role) {
+          return undefined
+        }
+      }
+
+      const lastPinnedIndex = pinnedIndices[pinnedIndices.length - 1]!
+      const lastPinnedMessage = messages[lastPinnedIndex]!
+      if (
+        lastPinnedMessage.role !== 'user' ||
+        lastPinnedMessage.content.some((block) => block.type === 'toolUseBlock')
```
strandly-the-agent, line 525: ⚪ nit, defence-in-depth only — the validations are asymmetric: `:499-500` rejects a `toolResultBlock` on the **first** pinned message but allows a `toolUseBlock`; this check rejects a `toolUseBlock` on the **last** pinned message but allows a `toolResultBlock`; and for `pinnedIndices.length >= 3` a **middle** pinned message's content is never inspected by either.

I could not reach this from a well-formed history — `isPinned`'s one-hop propagation (`compression/pin-message.ts:20-39`) is symmetric, so a real adjacent pair always pins both halves together, and for a single pinned message your two checks compose into a complete validation. It only bites on an already-malformed array (a non-adjacent pair, or a `user`-role `toolUse`), which needs a hand-seeded `new Agent({ messages })` or a restored session — and in that case it degrades an already-invalid history rather than breaking a valid one. So: genuinely a nit.

If you want it airtight anyway, the no-pinned branch already has the stricter rule at `:525-528` (exclude **both** block types). Applying that to every pinned message, not just first and last, would make the two branches consistent and close the middle-message hole in one line.

strands-ts/src/conversation-manager/__tests__/sliding-window-conversation-manager.test.ts
```diff
@@ -971,6 +1004,180 @@ describe('SlidingWindowConversationManager', () => {
...
+      const expectedMessages = messages.slice(2)
+
+      await triggerContextOverflow(manager, mockAgent, new ContextWindowOverflowError('Context overflow'))
+
+      // startIndex=2 is the plain user boundary; the complete pair at indices 3–4 must not override it.
+      expect(mockAgent.messages).toEqual(expectedMessages)
+      expect(mockAgent.messages[0]!.content[0]).toEqual({ type: 'textBlock', text: 'Plain user message' })
+    })
+
+    it('does not treat a user message containing tool use as an assistant tool-pair boundary', () => {
```
strandly-the-agent, line 1054: 🟡 This test never reaches `_findToolPairUserAnchor`. I confirmed it by making that method `throw` unconditionally — this test still passes.

The reason: the tool-bearing message here is `role === 'user'`, and `_findToolPairTrimPoint`'s `hasToolUse` requires `role === 'assistant'` (`:465`), so it returns `undefined` and the anchor function is never called. So the test does guard something real — `_findToolPairTrimPoint`'s pre-existing role gate — but not the new anchor exclusion logic its name implies.

Only 4 of the 6 new tests actually invoke the anchor function (same throw-instrumentation check). Either retitle this to describe what it exercises (e.g. `does not treat a user message containing tool use as a tool-pair boundary`), or extend the fixture with a real `assistant(toolUse)` + `user(toolResult)` pair later in range so control reaches the anchor logic too.

strands-ts/src/conversation-manager/__tests__/sliding-window-conversation-manager.test.ts
```diff
@@ -971,6 +1004,180 @@ describe('SlidingWindowConversationManager', () => {
...
+      expect(messages.map((message) => message.role)).toEqual([
+        'user',
+        'assistant',
+        'user',
+        'assistant',
+        'user',
+        'assistant',
+        'user',
+      ])
+    })
```
strandly-the-agent, line 1296: 🟡 Three guards inside `_findToolPairUserAnchor` are load-bearing but untested. I deleted each one in turn against the current suite:

| Guard deleted | Suite result | What the mutant then does |
|---|---|---|
| alternation loop `:505-509` | 🟢 still green | with non-contiguous pins {0, 2} and an unpinned assistant at 1, trims to `['user','user','assistant','user']` — invalid adjacency |
| first-pinned `toolResultBlock` check `:500` | 🟢 still green | anchors on a pinned `user` message that is itself an **orphaned `toolResultBlock`** and sends it as the head — the exact #2085 shape, via the pinned route |
| last-pinned `toolUseBlock` check `:515` | 🟢 still green | anchors on a `user` message whose only content is a bare unpaired `toolUse` |

v8 branch data agrees: the fail side of `:498-503` and `:505-509` is never taken by any test in the file, and although `:513-518` looks covered, its `toolUseBlock` sub-condition only ever evaluates `false`.

Below are 4 tests in this file's style. **All 4 pass at `b6c34d3b0`** (61 → 65), and I verified the first three each fail when their corresponding guard is deleted. The fourth closes the "no anchor exists at all" decline branch, which no test currently reaches.

<details>
<summary>4 verified tests — insert after this test, inside <code>describe('reduceContext - tool pair validation')</code></summary>

```ts
    it('declines fallback when non-contiguous pinned messages do not alternate', async () => {
      const { pinMessage } = await import('../compression/pin-message.js')
      const manager = new SlidingWindowConversationManager({ windowSize: 2, shouldTruncateResults: false })
      const messages = [
        new Message({ role: 'user', content: [new TextBlock('Pinned user one')] }),
        new Message({ role: 'assistant', content: [new TextBlock('Unpinned response')] }),
        new Message({ role: 'user', content: [new TextBlock('Pinned user two')] }),
        new Message({ role: 'assistant', content: [new TextBlock('Old response')] }),
        new Message({ role: 'user', content: [new TextBlock('Old request')] }),
        createToolUseMessage('lookup', 'id-1'),
        createToolResultMessage('id-1', 'Old result'),
        createToolUseMessage('lookup', 'id-2'),
        createToolResultMessage('id-2', 'Recent result'),
      ]
      pinMessage(messages, 0)
      pinMessage(messages, 2)
      const before = [...messages]

      const result = manager.reduce({
        agent: createMockAgent({ messages }),
        model: {} as Model,
        error: new ContextWindowOverflowError('Context overflow'),
      })

      // Pinned indices 0 and 2 are both user; retaining both once the unpinned index 1 is removed
      // would place two user messages next to each other, so the fallback must decline.
      expect(result).toBe(false)
      expect(messages).toEqual(before)
    })

    it('declines fallback when the first pinned message is an orphaned tool result', async () => {
      const { pinMessage } = await import('../compression/pin-message.js')
      const manager = new SlidingWindowConversationManager({ windowSize: 2, shouldTruncateResults: false })
      const messages = [
        createToolResultMessage('orphan', 'Orphan result'),
        new Message({ role: 'assistant', content: [new TextBlock('Old response')] }),
        new Message({ role: 'user', content: [new TextBlock('Old request')] }),
        createToolUseMessage('lookup', 'id-1'),
        createToolResultMessage('id-1', 'Old result'),
        createToolUseMessage('lookup', 'id-2'),
        createToolResultMessage('id-2', 'Recent result'),
      ]
      pinMessage(messages, 0)
      const before = [...messages]

      const result = manager.reduce({
        agent: createMockAgent({ messages }),
        model: {} as Model,
        error: new ContextWindowOverflowError('Context overflow'),
      })

      // Anchoring on a user message that is itself an orphaned toolResult would start the
      // conversation with a tool result that has no toolUse, the shape #2085 forbids.
      expect(result).toBe(false)
      expect(messages).toEqual(before)
    })

    it('declines fallback when the last pinned message carries a bare tool use block', () => {
      const manager = new SlidingWindowConversationManager({
        windowSize: 2,
        shouldTruncateResults: false,
        pinFirst: 1,
      })
      const messages = [
        new Message({ role: 'user', content: [new ToolUseBlock({ name: 'weird', toolUseId: 'weird-1', input: {} })] }),
        new Message({ role: 'assistant', content: [new TextBlock('Old response')] }),
        new Message({ role: 'user', content: [new TextBlock('Old request')] }),
        createToolUseMessage('lookup', 'id-1'),
        createToolResultMessage('id-1', 'Old result'),
        createToolUseMessage('lookup', 'id-2'),
        createToolResultMessage('id-2', 'Recent result'),
      ]
      const before = [...messages]

      const result = manager.reduce({
        agent: createMockAgent({ messages }),
        model: {} as Model,
        error: new ContextWindowOverflowError('Context overflow'),
      })

      // The pinned anchor is a user message whose only content is an unpaired toolUse; role alone
      // is not enough, so the last-pinned check must reject it.
      expect(result).toBe(false)
      expect(messages).toEqual(before)
    })

    it('declines fallback when no plain user message exists before the tool pair', () => {
      const manager = new SlidingWindowConversationManager({ windowSize: 2, shouldTruncateResults: false })
      const messages = [
        createToolResultMessage('orphan-0', 'Orphan result'),
        createToolUseMessage('lookup', 'id-1'),
        createToolResultMessage('id-1', 'Result 1'),
        createToolUseMessage('lookup', 'id-2'),
        createToolResultMessage('id-2', 'Result 2'),
      ]
      const before = [...messages]

      const result = manager.reduce({
        agent: createMockAgent({ messages }),
        model: {} as Model,
        error: new ContextWindowOverflowError('Context overflow'),
      })

      // startIndex is 3 and every message before the id-2 pair carries tool content, so no anchor
      // exists at all and the fallback must decline.
      expect(result).toBe(false)
      expect(messages).toEqual(before)
    })
```
</details>

strandly-the-agent (approved):
**TL;DR — approving; this clears my earlier changes-requested.** The round-1 blocker is fixed and I re-verified it cold end to end: no assistant-first payload reaches the provider on the proactive path, the reactive path, or `pinFirst`. Nothing left is a PR-introduced correctness defect. What remains is 4 cheap should-fixes — three untested-but-load-bearing guards, one mis-scoped test, an undocumented (deliberate) divergence from Python, and a stale PR body/title.

And credit where it's due: **declining the fallback when there's no valid anchor is the right call, and I proved it rather than assuming it.** The adversarial pass in my own pipeline proposed "anchor on the last pinned *user* index instead of declining" — I implemented exactly that in a scratch worktree and it reproduces the round-1 bug (`[user, assistant, assistant, user, assistant]`). Because the removal loop retains *every* pinned index (`:262-266`), the seam is always (last pinned) → `assistant(toolUse)`; if the last pinned message is an assistant, no choice of anchor can save it. So `:513-518` returning `undefined` is correct, not a gap. I've dropped that suggestion.

<details>
<summary><b>✅ Verification ledger</b></summary>

Reviewed `b6c34d3b0` (prev `f3136c744`, base `1b671d6a1`), in dedicated `git worktree` checkouts — one per pass, after two passes corrupted each other's measurements sharing a tree in round 1.

**Re-ran cold myself at this head:**

| Check | Result |
|---|---|
| Proactive path, public `Agent` API, fake model, no network | ✅ `ASSISTANT-FIRST PAYLOAD REACHED PROVIDER: false`; invocation-2 first payload `[user, assistant, user, assistant, user]` **valid**; `agent.messages` after invocation 1 `[user, assistant, user, assistant]` **valid** |
| The new debug line actually fires | ✅ `window_size=<4>, trim_index=<13>, user_anchor_index=<8> \| no plain user trim point, falling back to complete tool pair` |
| Scoped suite (`sliding-window` + `context-compression`) | ✅ **88/88 passed** |
| `npx eslint` both changed files · `npx tsc --noEmit -p src/tsconfig.json` | ✅ clean · ✅ clean |
| `pinFirst` 1/2/3, 8 live invocations, `windowSize: 3`, 3 tool calls/turn | 🟡 `pinFirst:1` → `4,4,4,4,4,4,4,4` bounded · **`pinFirst:2` → `11,19,27,35,43,51,59,67` unbounded** · `pinFirst:3` → `6,6,6,6,6,6,6,6` bounded |
| Refuted the "anchor on last pinned user" fix by implementing it | 🔴 produced `[user, assistant, assistant, user, assistant]` — the round-1 bug |
| The 4 tests I propose below | ✅ all pass at this head (65 tests), and I killed 3 of them individually by deleting the guard each one covers — so they're load-bearing, not decoration |

**Ran by my pipeline, not personally re-run this round** (stating it so you can weight it): the reactive-path and serialized-Bedrock-wire repros (`messages[0].role = "user"`, `[u,a,u,a,u]`); a 20,000-case property fuzz over structurally-valid histories with arbitrary non-contiguous pins — 2,540 reached the anchor path, **zero** invalid results; 11 adversarial pin shapes — 0/11 broke at head, 2 broke at `f3136c744`; the full `npm test` / coverage numbers quoted below.

**Not verified:** no live Bedrock call (no network in this pass).

</details>

<details>
<summary><b>🟡 Should-fix (4) — details inline</b></summary>

1. **Three load-bearing guards in `_findToolPairUserAnchor` have no test** (`:500`, `:505-509`, `:515`). All three are *correct*; I deleted each in turn and the suite stayed green. The worst survivor makes a `pinFirst`-pinned `user` message that is itself an orphaned `toolResultBlock` become the anchor and get sent as the head — the exact #2085 shape this PR exists to prevent, reached via the pinned route. 4 verified tests inline on the test file.
2. **The declined-fallback path is silent, and the warning it lands on is misleading** (`:248`, `:252-257`) — it says "no valid trim point found" when a complete pair *was* found and only the anchor check rejected it. Suggestion inline.
3. **The deliberate divergence from Python is undocumented anywhere** (`:477-487`). Suggestion inline.
4. **One new test doesn't reach the code its name implies** (test file `:1053`).

</details>

<details>
<summary><b>⚪ Nits · ❓ Questions · pre-existing bugs found along the way</b></summary>

**Nits**
- Asymmetric pinned-content checks (inline at `:515`).
- `uses the pinned first user as a safe fallback anchor` (`:1112-1142`) ends at length 3 against `windowSize: 2` — the overshoot is real and intended-looking but never asserted. Worth a length assertion or a one-line comment, so a future change that overshoots *differently* gets caught.
- **PR body and title still describe revision 1 only** — no mention of the anchor, the pinned-prefix validation, or the decline branch. The Testing numbers have rotted too: 56 → **61** scoped, 3,869 → **3,874**, 93.64% → **93.68%**. Worth a refresh before merge.

**Questions** (all non-blocking)
- Why the keep-anchor design over the simpler decline-only option from round 1 (~5 lines, like `agentic-context.ts:159-163`)? It trims strictly more and it held up under everything I threw at it — I'd just like the "why" on record, ideally in the PR description.
- Is the reactive-path trade-off intentional? TS now declines and lets the original `ContextWindowOverflowError` propagate where Python trims and retries (with a payload some model families would 400 on). I think declining is right; worth being a deliberate choice rather than a side effect.
- `:489-494` duplicates `partitionPinned`'s pattern (`pin-message.ts:80-91`) without reusing it — intentional because you need indices rather than messages?

**`pinFirst: 2` — the one behavioural gap worth naming** (measured above). An even `pinFirst` pins `[user, assistant]`, so the last-pinned check rejects *every* later reduction for that manager's lifetime (`_pinFirstApplied` is one-shot) and a tool-heavy conversation grows unbounded. It is **not a regression** — base grows unbounded there too, identically — and per the refutation above, declining is correct. But it does mean this fix's benefit never reaches a perfectly ordinary config. Since the caller has no way to discover that, would a one-time warning where `applyPinFirst` runs (`:219-221`) be worth it, when the pinned prefix's own shape can never satisfy the anchor invariant?

**Pre-existing, not this PR** (you asked for any bugs, so: found while testing, both byte-identical at base — happy to file these)
- Fractional `windowSize` (e.g. `2.5`) bypasses every guard: `messages[2.5]` is `undefined`, so `findValidTrimPoint` returns the fractional value unchanged, no `=== messages.length` / `> messages.length` check catches it, and the removal loop deletes `Math.ceil(trimIndex)` messages — reproduced an assistant-first result on a plain, tool-free conversation with no fallback involved. Note `pinFirst` *is* clamped with `Math.max(0, ...)`; `windowSize` isn't validated at all.
- The plain removal path + `pinFirst` can still produce `user` → `user` on a well-formed tool-free history (5 messages, `windowSize: 2`, pin index 2 only → `['user','user']`). The new alternation validation guards only the fallback path added here.
- Worth a `strands-py` issue: Python's `_find_tool_pair_trim_point` has no anchor concept, so it still ships the hazard this PR just fixed in TS. If anyone ports this back, note Python's `removed_message_count` would need to account for the retained anchor — TS has no such counter.

</details>

Good revision — the anchor invariant held against a 20k-case fuzz and every adversarial pin shape I could construct, which is not a given for hand-rolled sequence logic. Framing all of the above as work for a human to sign off rather than a gate: `JackYPCOnline`, the only judgement calls I'd genuinely want your eye on are the `pinFirst: 2` gap and whether the Python divergence should be documented in code, in the PR body, or both.

strands-ts/src/conversation-manager/__tests__/sliding-window-conversation-manager.test.ts
March-77: Renamed in `b6c34d3b0` to describe the path it actually covers: rejecting a user-role message containing tool use as an assistant tool-pair boundary.

strands-ts/src/conversation-manager/sliding-window-conversation-manager.ts
March-77: Added the fallback debug breadcrumb in `b6c34d3b0`, including both the selected tool-pair index and retained user-anchor index.

strands-ts/src/conversation-manager/sliding-window-conversation-manager.ts
March-77: Resolved in `b6c34d3b0`. The fallback now retains a validated user anchor before the complete tool pair, preserving a user-first and legally alternating history for both proactive and reactive reduction paths.

strands-ts/src/conversation-manager/sliding-window-conversation-manager.ts
March-77: Fixed in `2e08d2461`. When a complete pair exists but no valid user anchor does, the manager emits the specific debug message and returns directly, so the generic “no valid trim point found” warning is no longer shown for this case.

strands-ts/src/conversation-manager/__tests__/sliding-window-conversation-manager.test.ts
March-77: Covered in `b6c34d3b0` and `2e08d2461`. The suite now exercises both tool-pair guards, the decline path, safe and unsafe `pinFirst` cases, and the user-anchor validation branches.

strands-ts/src/conversation-manager/sliding-window-conversation-manager.ts
March-77: I am leaving this defense-in-depth change out of this PR. The case requires an already malformed history and is not reachable from well-formed pinned tool pairs; the current change stays focused on preserving a valid user-first fallback boundary.

strands-ts/src/conversation-manager/__tests__/sliding-window-conversation-manager.test.ts
March-77: Corrected in `b6c34d3b0`. The fixture now places a reachable complete pair after the plain user boundary, so the test genuinely proves that the plain boundary remains preferred.

strands-ts/src/conversation-manager/sliding-window-conversation-manager.ts
March-77: Documented in `2e08d2461`. The method comment now explicitly explains the intentional TypeScript divergence from Python and the user-first/provider-alternation reason for retaining the anchor.

strands-ts/src/conversation-manager/__tests__/sliding-window-conversation-manager.test.ts
March-77: Added direct coverage in `2e08d2461` for all three guards: non-alternating pinned messages, an orphaned tool result in the first pinned message, and a bare tool use in the last pinned message. A fourth test covers the no-anchor decline path. The targeted suite passes 65 tests.

March-77:
The requested fallback guard coverage, anchor-rejection logging, TypeScript/Python divergence note, and test correction are now pushed. The targeted suite passes 65 tests, with type-check, lint, and format checks clean. The remaining open thread is the explicitly non-blocking malformed-history defense-in-depth nit. @JackYPCOnline, this is ready for re-review; the secure integration and automated review jobs are awaiting manual approval.

JackYPCOnline (approved):

PR #3448, opened 2026-07-24
feat(graph): add Python concurrency limit

### LINKED ISSUE #3156: [PORT] Graph max_concurrency cap (TS→Py)
> ⚠️ Filed by Strandly as part of the #2498 parity audit. Source feature verified merged + unit-tested + documented in `strands-ts`. Agent-produced — **read the scope guard: port the cap concept, NOT the scheduler.** (No `port` label yet — a maintainer should add it.)

**Parent:** #2498

## Source SDK
`strands-ts`

## Target SDK
`strands-py`

## Related Files
```
strands-ts/src/multiagent/graph.ts
strands-ts/src/multiagent/__tests__/graph.test.ts
site/src/content/docs/user-guide/concepts/multi-agent/graph.mdx
site/src/content/docs/user-guide/concepts/multi-agent/graph.ts
```

## What needs to be migrated
Cap the number of graph nodes executing concurrently. TS surface:
- Config `maxConcurrency?: number` (`graph.ts:52`), default `Infinity` / no limit (`graph.ts:119`).
- Validation `>= 1` else throw `max_concurrency=<N> | must be at least 1` (`graph.ts:584`).
- Enforcement: launch a new node only while `streams.size < maxConcurrency` (`graph.ts:337`).

(The `_computeCanRunConcurrently` printer optimization at `graph.ts:696` is a TS-printer concern — **do not port**.)

## 🚩 Scope guard
**Port the concurrency-cap concept only, adapted to Python's existing batch OR-join scheduler. Do NOT port TS's per-node AND-join scheduler.** Python's batching vs TS's individual-node launch is a **deliberate, documented divergence** (`graph.mdx:828-840`). Rewriting Python to per-node launching would silently flip dependency-resolution semantics (OR→AND).

## Python landing points
`strands-py/src/strands/multiagent/graph.py` (no cap today — a whole batch is fanned out via `asyncio.create_task` at `:831`, gathered at `:875`).
- Add `max_concurrency: int | None = None` to `Graph.__init__` (`:497`), `GraphBuilder.__init__` (`:311`), a builder setter mirroring `set_max_node_executions` (`:387`), and `build()`'s `Graph(...)` call (`:467`).
- **Enforcement:** create an `asyncio.Semaphore(max_concurrency)` (skip when `None`) and `async with` it inside `_stream_node_to_queue` (`:877`) around `_execute_node`, so at most N nodes run **within** a batch. Do not change batch boundaries or `_find_newly_ready_nodes` (`:800`).
- Validate `>= 1`; reuse the message `max_concurrency=<N> | must be at least 1`.

## Gotchas / open questions
- **`Infinity`→`None`** (Python siblings use `None`=unlimited; guard semaphore creation on `is not None`).
- **Release on timeout/exception/cancel** — use `async with semaphore:` around the streaming body (per-node timeout already at `:886`; fail-fast cancel at `:848`).
- Naming: `maxConcurrency`→`max_concurrency` (+ `set_max_concurrency`).

- Merged: ✅  ·  Unit tests: ✅ (`graph.test.ts:164` throws `<1`; `:175` default `Infinity`; `:546` asserts `maxConcurrent===1`)  ·  Integ: ❌ no dedicated cap test.
- Docs: ✅ `graph.mdx:106` (config) + `:828-840` (the SDK-differences section this port must respect).

## Confirmation
- [x] The change is complete in the source SDK (merged + unit-tested + documented); every related artifact is listed above.

March-77:
## Description

Adds a `max_concurrency` limit to Python Graph execution while preserving the
existing batch scheduler and OR-join dependency semantics.

- Adds `GraphBuilder.set_max_concurrency()` and the optional `Graph`
  constructor parameter.
- Validates configured limits are at least 1.
- Applies a per-batch semaphore around complete node execution, including
  streaming and node timeout handling.
- Leaves the default unlimited for backward compatibility.
- Documents the option and the Python/TypeScript scheduling distinction.

The constructor parameter is appended to the existing signature so positional
callers remain compatible.

## Related issues

Fixes #3156

## Testing

- `uv run pytest tests/strands/multiagent/test_graph.py -q` (83 passed)
- `uv run ruff format --check src/strands/multiagent/graph.py tests/strands/multiagent/test_graph.py`
- `uv run ruff check src/strands/multiagent/graph.py tests/strands/multiagent/test_graph.py`
- `uv run mypy src/strands/multiagent/graph.py`
- `npm run typecheck` (site)
- `npm run typecheck:snippets` (site)
- `npm run build` (857 pages, 856 HTML files, no broken links)

The broader `tests/strands/multiagent` collection requires the optional `a2a`
dependency, which was not installed in the core+dev test environment. The
complete graph test module passed.

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/strands-agents/harness-sdk/blob/main/CONTRIBUTING.md) guide.
- [x] I have performed a self-review of my code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.
- [ ] I have run `hatch run prepare`.

The targeted Python checks and full site build are listed above. The full
multi-version `hatch run prepare` matrix was not run locally.


strands-py/src/strands/multiagent/graph.py
```diff
@@ -393,6 +394,23 @@ def set_max_node_executions(self, max_executions: int) -> "GraphBuilder":
...
+        Args:
+            max_concurrency: Maximum concurrent node executions.
+
+        Returns:
+            This builder for method chaining.
+
+        Raises:
+            ValueError: If max_concurrency is less than 1.
+        """
+        if max_concurrency < 1:
```
gnanirahulnutakki: **Correctness — validate a real positive integer:** Fractional values pass this check even though the public contract is `int`. This is not just a type-hint mismatch: `asyncio.Semaphore(1.5)` decrements through `1.5 → 0.5 → -0.5` without ever reporting itself locked. On this head, a graph with three entry nodes and `max_concurrency=1.5` reached an observed peak of 3, so the configured cap was not enforced. Please reject non-integer (and `bool`) values at both this builder path and `Graph.__init__`, with regression coverage for each. The SDK already uses this positive-integer validation pattern for analogous limits.

gnanirahulnutakki (commented):
Reviewed exact head `13616a0d`. The per-batch semaphore approach matches #3156 and preserves the documented Python scheduler. Locally, all 83 graph tests passed and Ruff lint/format checks were clean. I found one reproducible runtime-validation issue inline; addressing it at both public construction paths should make the cap safe for Python callers.

strands-py/src/strands/multiagent/graph.py
March-77: Fixed in 97c2b6e6. Both GraphBuilder.set_max_concurrency and Graph.__init__ now reject booleans, fractional values, zero, and negative values, accepting only positive integers. Added parametrized coverage for both construction paths; all 86 graph tests and lint checks pass.

gnanirahulnutakki: Verified on `97c2b6e6`: both public construction paths now reject booleans, fractional values, zero, and negatives, while the existing positive-integer path remains covered. The complete graph module passes (86 tests), along with Ruff format/check and targeted mypy. This resolves my finding—thank you for the focused fix.

gnanirahulnutakki (approved):
Re-reviewed exact head `97c2b6e62498b98fc8d575743299bc6e0351c190`.

The follow-up is focused on the reported runtime-validation gap: both `GraphBuilder.set_max_concurrency` and `Graph.__init__` now reject `bool`, fractional values, zero, and negatives, with parameterized coverage for each public path. Independent validation passed:

- complete graph test module: 86 passed
- Ruff format check: clean
- Ruff lint: clean
- targeted mypy: clean
- current GitHub checks: 7 passed, 3 awaiting workflow authorization, 0 failures

The original positive-integer finding is addressed. Approved.

opieter-aws:
@strandly-the-agent can you review this PR? Additionally test for how this affects graph resume logic

strands-py/src/strands/multiagent/graph.py
```diff
@@ -899,8 +932,14 @@ async def stream_node() -> None:
...
                 # No timeout - stream normally
-                async for event in self._execute_node(node, invocation_state):
-                    await event_queue.put(event)
+                await stream_node()
+
+        try:
+            if semaphore is None:
+                await execute_node()
+            else:
+                async with semaphore:
```
strandly-the-agent, line 963: **🔴 Blocker — this reopens the #3390 fan-in resume bug when combined with session persistence.**

`_compute_pending_sources`'s in-flight signal (`graph.py:1415`, `executing_nodes = {... node.execution_status == Status.EXECUTING}`, added by #3390) assumes every dispatched node reaches `Status.EXECUTING` (`graph.py:1023`) essentially as soon as it's scheduled. Gating that line behind the semaphore means a node still queued for a slot stays `Status.PENDING` — identical, from `_compute_pending_sources`'s point of view, to a node that was never part of the batch (a "bypassed dead branch").

Concretely: two entry points `left`/`right` → `join`, `max_concurrency=1`. If a `SessionManager` (or any `AfterNodeCallEvent` hook) persists state right when `left` finishes, `right` is still `PENDING` (never got a slot yet), so `_compute_pending_sources` can't forward-reach it — `join` gets wrongly included in `next_nodes_to_execute`. Resuming a fresh graph from that snapshot then runs `join` twice (verified — 5/5 runs against this branch): once in the batch where it was wrongly deemed ready alongside `right`, and again because `_find_newly_ready_nodes` re-evaluates `join`'s edge from `right` after that batch and finds it "newly ready" a second time. `GraphResult.status` still comes back `COMPLETED`, so it's silent.

This is scoped to session-resume (I didn't find the same dependency in the interrupt-resume path). `TestResumeInFlightSibling` (`test_graph.py:3285-3516`) is exactly the regression suite for the bug this reopens, but none of its cases are run with `max_concurrency` set.

Would it work to move the `EXECUTING` transition (or an equivalent "dispatched" marker) ahead of the semaphore acquire, so a queued-but-not-yet-slotted node still counts as in-flight for resume purposes? Or, if that's not straightforward, this combination (max_concurrency + SessionManager + a fan-in whose parents land in the same batch) probably needs to be explicitly flagged as unsupported for now.

strandly-the-agent (changes requested):
**TL;DR:** The semaphore mechanics themselves are solid (verified below), but `max_concurrency` + session persistence reopens the #3390 fan-in resume bug — confirmed with a deterministic repro, 5/5 runs, join double-executes on resume. 🔴 blocker. Everything else I dug into (refactor equivalence, event-loop safety, the 0.1s poll loop, scope guard) checked out clean.

✅ Verified myself on `97c2b6e6`: 86/86 `test_graph.py` pass; `_find_newly_ready_nodes`/`_is_node_ready_with_conditions`/`_execute_graph`/`_build_node_input` byte-identical to merge-base `1b671d6a`; `stream_node`/`execute_node` split is control-flow-identical to base when `max_concurrency is None`; semaphore releases on exception/timeout/cancel via `async with` on every path I traced; per-batch `asyncio.Semaphore` construction can't leak across batches or bind to the wrong loop; the 0.1s poll loop doesn't busy-wait or exit early while tasks are queued on the semaphore.

<details>
<summary>Full analysis</summary>

### 🔴 Blocker: `max_concurrency` breaks the in-flight signal that session-resume depends on

See inline comment on `graph.py:941`. `_compute_pending_sources`'s `EXECUTING`-based seed (`graph.py:1415`, added by #3390 specifically for "two concurrent entry points feeding one join") assumes every dispatched node reaches `Status.EXECUTING` (`graph.py:1023`) essentially immediately. This PR's semaphore now gates entry to that line — a node queued behind the cap stays `PENDING`, indistinguishable from a node that was never scheduled. Repro (ran against this branch, `uv run --extra dev pytest`):

```
two entry points {left, right} -> join, max_concurrency=1

baseline (no cap):  sibling status at capture = EXECUTING, next_nodes_to_execute = ['right']            (join correctly excluded)
capped (=1):        sibling status at capture = PENDING,   next_nodes_to_execute = ['right', 'join']    (join wrongly marked ready)
```

Carrying that snapshot through an actual `deserialize_state` + `invoke_async` resume (mirroring `test_graph.py`'s own `TestResumeInFlightSibling`, `test_graph.py:3285-3516`) shows the consequence: `join.stream_async` is called **twice** — 5/5 runs. Root cause once `join` lands in `next_nodes_to_execute` alongside `right`: after that batch runs, `_find_newly_ready_nodes` re-evaluates `join`'s incoming edge from `right` (now in `completed_batch`) and marks it ready *again* for the next batch — `_is_node_ready_with_conditions` has no "already ran this batch" guard. `GraphResult.status` still reports `COMPLETED`, so this is silent: extra cost, and a correctness problem for any non-idempotent node.

Scope: this only affects **session-resume** (`_compute_pending_sources`'s `EXECUTING` seed). I didn't find an equivalent dependency in the interrupt-resume path. `TestResumeInFlightSibling`'s existing tests (`test_graph.py:3285-3516`) are the regression coverage for the underlying #3390 fix, but none of them are parametrized with `max_concurrency` set, which is how this got through.

For the reviewer: either the `EXECUTING` transition (or an equivalent "has an outstanding task in this batch" signal) needs to happen before/independent of the semaphore gate, or `max_concurrency` + `SessionManager` + this fan-in topology needs to be explicitly called out as unsupported until fixed.

### 🟡 Should-fix: `graph.max_concurrency = 0` set post-construction hangs forever, silently

Both validated construction paths (`GraphBuilder.set_max_concurrency`, `Graph.__init__`) correctly reject `0`. But `self.max_concurrency` is a plain public attribute (`graph.py:562`), and `_execute_nodes_parallel` (`graph.py:859`) builds `asyncio.Semaphore(self.max_concurrency)` from it fresh every batch with no re-validation. Setting `graph.max_concurrency = 0` after construction produces a `Semaphore(0)`, which starts locked — the next batch hangs indefinitely with no error. Confirmed: `invoke_async` under a 1s `asyncio.wait_for` times out. Low likelihood (requires bypassing both public setters) — should-fix, not a blocker.

### ⚪ Nit: capped batches can deadlock if a node's own logic depends on a batch-mate's progress

Expected/inherent to any bounded-worker scheduler, not a bug in this PR, but worth a one-line callout in the `set_max_concurrency` docs: with `max_concurrency < len(batch)`, if any node's executor internally waits on another node in the *same* batch making progress (e.g. a shared external coordination primitive), the batch can deadlock, since the semaphore doesn't guarantee fairness/relative ordering. Nested `Graph`/`Swarm` executors are fine — each maintains its own independent semaphore, no shared resource.

### Appendix — found while testing, pre-existing, unrelated to this PR

While building the resume repro I found a second, *structurally guaranteed* (non-racy) double-execution bug: a node with ≥2 real incoming edges that land in the same dispatch batch re-fires itself via `_find_newly_ready_nodes` after that batch completes. Reproduces with `max_concurrency` unset, using only unmodified base scheduler functions — filed separately as strands-agents/harness-sdk#3628, not blocking here.

</details>

strands-py/src/strands/multiagent/graph.py
strandly-the-agent: **Correcting my own comment above.** The mechanism is real and I've now verified it end to end — but the tier, the numbers, and the consequence I gave were all wrong, and the issue I filed alongside it (#3628) was a false positive I've since retracted. `March-77`, sorry for the churn; the corrected version is below and it is **not** a blocker on your PR.

**🟡 What's actually true:** `max_concurrency` doesn't reopen a bug — it widens a pre-existing one. A node queued on the semaphore stays `PENDING` and never reaches `Status.EXECUTING` (graph.py:1023), so `_compute_pending_sources`' `EXECUTING`-only in-flight seed (graph.py:1415) can't see it. With ≥2 independent entry points feeding one join and a mid-batch session write, the join is persisted into `next_nodes_to_execute` and runs **twice** on resume, silently (`GraphResult.status == COMPLETED`).

But `_compute_pending_sources` and `_execute_node` are **byte-identical to merge-base** `1b671d6a`, and the defect reproduces on `main` today without this PR. I've filed it as #3633 with a crash-restart repro.

<details>
<summary>What I got wrong, and the corrected evidence</summary>

- **Tier: not a blocker.** On `main` the bug needs an executor that reaches its first `await` without suspending (a cached/replayed/local `Model` — a public extension point). A network-backed agent always suspends, so `main` is safe in practice. What this PR does is remove that precondition: once the cap throttles a batch, *any* executor can leave a sibling `PENDING` long enough to trip it. That widening is real and worth deciding about — but blocking a contributor over a defect in code the PR provably doesn't touch is the wrong lever, so I've dropped it to 🟡 and moved the root cause to #3633.
- **Frequency: "5/5 runs" was not dispatch-order-controlled.** `entry_points` is a `set` hashed by `node_id` (graph.py:266/308) and Python randomizes string hashing per process, which decides *which* entry point dispatches first. Across 30 fresh processes with realistic agents: 22/30 double-executed — i.e. ~100% conditional on the capturing parent dispatching first, not a fixed ratio. Uncapped: 0/30.
- **Consequence: not incomplete inputs.** Both `join` calls saw both parents' outputs. The damage is duplicated work — doubled cost and doubled side effects for a non-idempotent node — not a wrong answer.
- **My claim that the interrupt-resume path has no equivalent dependency was unverified when I made it.** It has since been tested separately and holds: a queued node is *not* dropped on interrupt-resume (8/8 runs) — `_execute_nodes_parallel` waits for every task before returning, so a slot-less node always runs before the batch ends.
- **#3628 was a false positive** — different mechanism, refuted 12/12 against the repo's own passing `test_graph_execution`, and the behavior is documented OR-join semantics. Closed and retracted; please disregard it.

</details>

Nothing here needs a code change from you unless the maintainers want it — see the decision question in my review summary.

strands-py/src/strands/multiagent/graph.py
```diff
@@ -828,9 +856,13 @@ async def _execute_nodes_parallel(
             nodes = [node for node in nodes if node.execution_status == Status.INTERRUPTED]
 
         event_queue: asyncio.Queue[Any | None | Exception] = asyncio.Queue()
+        semaphore = asyncio.Semaphore(self.max_concurrency) if self.max_concurrency is not None else None
```
strandly-the-agent, line 877: **🟡 `graph.max_concurrency = 0` set after construction hangs the graph forever, with no error.**

Both validated construction paths reject `0`, but this line rebuilds the semaphore from a plain public mutable attribute (graph.py:562) every batch with no re-validation. `Semaphore(0)` starts locked, so the next batch waits forever.

Measured on this branch — it's the only one of the five limit attributes whose garbage-in deadlocks rather than errors:

| post-construction assignment | outcome |
|---|---|
| `node_timeout = 0` | node-timeout `Exception` |
| `execution_timeout = 0` | `Status.FAILED` |
| `max_node_executions = 0` | `Status.FAILED` |
| `max_concurrency = 0` | **hangs, no error** |
| `max_concurrency = -1` | raw `ValueError: Semaphore initial value must be >= 0`, naming neither the param nor the graph |

The mutable attribute itself is consistent with its four siblings, so no need to change that — just give the same error the constructor already gives. If you take the shared-validation-helper suggestion from the other thread, calling that helper here is the tidiest version of this:

```suggestion
        if self.max_concurrency is not None and (
            isinstance(self.max_concurrency, bool)
            or not isinstance(self.max_concurrency, int)
            or self.max_concurrency < 1
        ):
            raise ValueError(f"max_concurrency={self.max_concurrency!r} | must be a positive integer")
        semaphore = asyncio.Semaphore(self.max_concurrency) if self.max_concurrency is not None else None
```

*Who hits it:* anyone setting the cap on an already-built graph rather than via the builder — a real path, since the attribute is intentionally public and mutable.

strands-py/src/strands/multiagent/graph.py
```diff
@@ -508,6 +527,7 @@ def __init__(
         id: str = _DEFAULT_GRAPH_ID,
         trace_attributes: Mapping[str, AttributeValue] | None = None,
         plugins: list[MultiAgentPlugin] | None = None,
+        max_concurrency: int | None = None,
```
strandly-the-agent, line 545: **🟡 Worth making this keyword-only.** `Graph.__init__` currently has zero keyword-only params, so this lands as the 13th *positional* one. `Swarm.__init__` puts a bare `*` before its config params (swarm.py:243) and `Agent.__init__` puts one mid-signature (agent.py:173) precisely so new options are keyword-only while pre-existing positional ones don't move — `strands-py/AGENTS.md:205` states that rationale directly.

The PR body justifies the placement with "positional callers remain compatible", which is true — but nobody can be passing a parameter that doesn't exist yet, so `*` costs nothing today and becomes a breaking change to add later. (`Graph` also isn't in `multiagent.__all__`, and every documented Python `Graph(...)` call site is already keyword-based.)

```suggestion
        *,
        max_concurrency: int | None = None,
```

strands-py/src/strands/multiagent/graph.py
```diff
@@ -393,6 +394,23 @@ def set_max_node_executions(self, max_executions: int) -> "GraphBuilder":
         self._max_node_executions = max_executions
         return self
 
+    def set_max_concurrency(self, max_concurrency: int) -> "GraphBuilder":
```
strandly-the-agent: **🟡 `set_max_concurrency(None)` raises, where all three sibling setters accept `None` as "no limit".** Verified on this branch: `set_max_node_executions(None)`, `set_execution_timeout(None)` and `set_node_timeout(None)` are all accepted and documented that way (graph.py:392, 418, 427) — this one rejects it, so the builder path can neither express "no limit" nor clear a previously-set cap, even though the constructor takes `int | None`.

The practical bite is conditional config: `builder.set_max_concurrency(cfg.limit)` works for three limits and raises on the fourth, and a caller can't guess which.

Would it be simpler to take `int | None` and validate only when not `None`?

```suggestion
    def set_max_concurrency(self, max_concurrency: int | None) -> "GraphBuilder":
```

(plus `if max_concurrency is not None and (...)` on the guard below, and a `(None for no limit)` note in the `Args:` line, matching the siblings).

site/src/content/docs/user-guide/concepts/multi-agent/graph.mdx
```diff
@@ -832,7 +835,7 @@ The Graph pattern is available in multiple SDKs. While the core concept is the s
 
 **Dependency resolution**: Python uses OR semantics, where a node fires when any single incoming edge from the completed batch is satisfied. TypeScript uses AND semantics, where a node runs only when all incoming edge sources are completed. This is more intuitive for join/diamond patterns where you want to wait for all inputs before proceeding.
 
-**Scheduling**: Python executes in discrete batches, waiting for the entire batch to complete before scheduling the next set of nodes. TypeScript launches nodes individually as they become ready, up to `maxConcurrency`. This avoids artificial bottlenecks where a fast node waits for a slow sibling to finish before its dependents can start.
+**Scheduling**: Python executes in discrete batches, running up to `max_concurrency` nodes at once and waiting for the entire batch to complete before scheduling the next set of nodes. TypeScript launches nodes individually as they become ready, up to `maxConcurrency`. This avoids artificial bottlenecks where a fast node waits for a slow sibling to finish before its dependents can start.
```
strandly-the-agent: **🟡 This describes the cap as a pure throttle, but it can change a graph's outcome.** `should_continue`'s timeout check only runs at batch boundaries (graph.py:151-168, called from graph.py:801-806), never mid-batch. Capping inflates a batch's wall-clock by roughly `ceil(n/cap)`, which can push the next boundary check past `execution_timeout`.

Verified: 4 entry nodes × 0.10s with `execution_timeout=0.25` → uncapped `COMPLETED` (~101ms); `max_concurrency=1` → **`FAILED`** (~402ms), tail node never ran. Nothing in the failure points at the cap as the cause.

```suggestion
**Scheduling**: Python executes in discrete batches, running up to `max_concurrency` nodes at once and waiting for the entire batch to complete before scheduling the next set of nodes. Because `execution_timeout` is only checked between batches, lowering `max_concurrency` increases per-batch wall-clock time and can cause a graph to fail that would otherwise complete. TypeScript launches nodes individually as they become ready, up to `maxConcurrency`. This avoids artificial bottlenecks where a fast node waits for a slow sibling to finish before its dependents can start.
```

*Who hits it:* anyone combining `max_concurrency` with `execution_timeout` where batch width exceeds the cap. (`node_timeout` is unaffected — its `wait_for` sits inside the semaphore, so queue time isn't charged against it.)

strands-py/tests/strands/multiagent/test_graph.py
```diff
@@ -1542,6 +1557,37 @@ async def stream_with_timing(node_id, delay=0.05):
     assert start_node_ids == {"a", "b", "c"}
 
 
+@pytest.mark.asyncio
+async def test_graph_max_concurrency_limits_parallel_nodes(mock_strands_tracer, mock_use_span):
```
strandly-the-agent: **🟡 The cap is never tested in combination with anything else.** This test is solid on its own — deterministic 15/15 and it survives four mutations of the implementation (semaphore removed, sized `cap+1`, validation accepting `0`, validation accepting `bool`). The gap is the intersections: no case pairs `max_concurrency` with `node_timeout`, `execution_timeout`, interrupts, session-resume, or a nested `MultiAgentBase` executor, and neither `cap == 1` nor `cap >= batch_width` is exercised.

That's how the resume finding got through: adding `set_max_concurrency(1)` to the existing `TestResumeInFlightSibling` topology fails against this branch today.

Also worth one targeted case for release-on-failure — mutating the impl to acquire but never release the slot is currently caught only by the blanket 90s `pytest-timeout`, i.e. an opaque CI hang rather than a pointed assertion.

<details>
<summary>Drop-in sketch (modelled on <code>TestResumeInFlightSibling</code>, <code>test_graph.py:3285-3516</code>)</summary>

```python
@pytest.mark.asyncio
async def test_resume_fan_in_once_for_queued_entry_point_sibling_under_max_concurrency(self):
    """A node throttled by max_concurrency must not be mistaken for a bypassed dead branch."""
    agent_left = create_mock_agent("left", "left done")

    async def left_stream(*args, **kwargs):
        # Real delay, not an Event: with cap=1, right cannot progress until left releases.
        yield {"agent_start": True}
        await asyncio.sleep(0.02)
        yield {"result": agent_left.return_value}

    agent_left.stream_async = Mock(side_effect=left_stream)

    builder = GraphBuilder()
    builder.add_node(agent_left, "left")
    builder.add_node(create_mock_agent("right", "right done"), "right")
    builder.add_node(create_mock_agent("join", "join done"), "join")
    builder.set_entry_point("left")
    builder.set_entry_point("right")
    builder.add_edge("left", "join")
    builder.add_edge("right", "join")
    builder.set_max_concurrency(1)
    graph = builder.build()

    captured: dict = {}

    def capture_on_left(event):
        if event.node_id == "left" and "payload" not in captured:
            captured["payload"] = event.source.serialize_state()

    graph.add_hook(capture_on_left, AfterNodeCallEvent)
    await graph.invoke_async("two entry points, capped")

    # Currently fails: next_nodes_to_execute == ['right', 'join']
    assert "join" not in captured["payload"]["next_nodes_to_execute"]
```

Note `AfterNodeCallEvent.should_reverse_callbacks` is `True` (`events.py:385-387`), so a hook added via `add_hook` runs *before* the SessionManager's sync — read state from the hook, not from the session file.

</details>

strandly-the-agent (commented):
**TL;DR — good, well-scoped port; the resume interaction `opieter-aws` asked about is real but is a pre-existing defect this PR *widens*, not one it introduces. No blockers from me.** I also have to correct my own earlier review on this PR: it graded that finding 🔴 and requested changes, and that was wrong on tier, numbers, and attribution. `March-77`, apologies for the churn — details in the [graph.py:941 thread](https://github.com/strands-agents/harness-sdk/pull/3448#discussion_r3714078819), and I've retracted the bogus issue I filed alongside it (#3628).

**On resume specifically (the ask):** a node queued on the new semaphore stays `PENDING` and never reaches `Status.EXECUTING` (graph.py:1023), so `_compute_pending_sources`' `EXECUTING`-only in-flight seed (graph.py:1415) can't see it. With ≥2 independent entry points feeding one join plus a mid-batch session write, the join is persisted as ready and **runs twice on resume, silently** (`GraphResult.status == COMPLETED`). But that code is byte-identical to merge-base and reproduces on `main` without this PR — filed as **#3633**. The PR's contribution is turning it from a corner case (needs a non-suspending model) into always-fires for *any* executor. Interrupt-resume is **clean** — a slot-less node is never dropped (8/8).

✅ Verified on `97c2b6e6` (base `1b671d6a`): 86/86 `test_graph.py`; ruff check/format + mypy clean; #3156's scope guard holds — `_find_newly_ready_nodes`, `_is_node_ready_with_conditions`, `_execute_graph`, `_build_node_input`, `_compute_pending_sources`, `serialize_state`/`deserialize_state` all AST-verified byte-identical to base; the `stream_node`/`execute_node` split is control-flow-identical to base when `max_concurrency is None`; no semaphore slot leaks on raise/timeout/cancel/nested/multi-batch; `test_graph_max_concurrency_limits_parallel_nodes` is genuinely enforcing (15/15, and survives 4 mutations of the impl). 🔴 Refuted and retracted: my #3628.

5 × 🟡 inline below (2 correctness, 2 API consistency, 1 test coverage). None require a code change to merge if the maintainers would rather ship and track.

<details>
<summary>Verification detail — how the resume finding was established</summary>

End-to-end through a real `FileSessionManager` with a hard `os._exit` the instant the session write lands (no cleanup, no later overwrite), then a fresh process to resume:

```
  [write] completed=['left'] next_nodes=['right','join'] status=executing | right=pending
  --- restart ---
  [resume] execution_order   : ['left','right','join','join']
  [resume] join.stream_async : 2 call(s)      <-- DOUBLE EXECUTION
```

Control, same topology/agents, uncapped: `next_nodes=['right'] | right=executing` → `join ran 1 time(s)`.

Attribution matrix (dispatch order forced worst-case, 10 runs each):

| tree | cap | executor | bad frontier |
|---|---|---|---|
| base `1b671d6a` | — | mock, no awaits | 10/10 |
| base | — | + `await sleep(0)` or `sleep(0.005)` | **0/10** |
| head `97c2b6e6` | 1 | any of the three | **10/10 each** |
| head | 2 (= batch width) | real await | 0/10 |

So one suspension point makes base safe — but a real `Agent` whose `Model.stream()` never awaits (cached/replayed/local model; `Model` is a public extension point) reproduces it on base too, which is why it's #3633 rather than a finding against this PR. Frequency under the cap is ~100% conditional on which entry point dispatches first, which is itself hash-randomized (`entry_points` is a `set`, graph.py:266/308) — ~22/30 fresh processes.

Also confirmed clean: `max_concurrency` correctly absent from `serialize_state()` (consistent with the other three construction-time limits); `deserialize_state`'s start-over + `reset_on_revisit` branch identical with and without the cap.

</details>

### Questions

❓ **(decision, blocking-ish — for @opieter-aws, since you asked about resume)** Given the finding is a pre-existing defect that this PR makes routinely reachable: ship as-is with #3633 tracked and a docs caution, or hold until #3633 is fixed? I don't think it should block `March-77` either way — it's a maintainer call about the knob's readiness, not about this diff's quality.

❓ **(non-blocking)** The cap makes entry-point dispatch order observable run-to-run, because `entry_points` is a hash-randomized `set` while TS drains an ordered array (`graph.ts:346`). Acceptable long-term, or worth an ordered structure as a follow-up?

<details>
<summary>Appendix — non-blocking (⚪ nits, pre-existing, and what came back clean)</summary>

**⚪ Nits**
- Error message drops the repo's bracket convention: `max_concurrency={value!r} | must be…` vs the house `field=<value> | message` (AGENTS.md:56; cf. `vended_plugins/goal/plugin.py:246`). TS emits bracketed `max_concurrency=<N> | must be at least 1` (`graph.ts:592-593`), which is the wording #3156 asked to reuse. Keep `!r` — it's strictly better here, distinguishing `'4'`/`4` and `True`/`1` — just restore the brackets (graph.py:410, 556).
- `Graph.__init__` gains a `ValueError` but has no `Raises:` section (docstring graph.py:531-548); AGENTS.md requires one. The new setter models it correctly.
- New tests don't use the repo's `tru_`/`exp_` assertion-pair convention (docs/TESTING.md), used 79× elsewhere in the same file.
- `test_graph_max_concurrency_validation`'s regex has an unescaped `.` for the `1.5` case — a `re.search` wildcard. No false pass today, just fragile.
- PR body: "validates at least 1" undersells the check (it also rejects `bool`/fractional), and "83 passed" is now 86.

**Pre-existing — filed//to file separately, not this PR's problem**
- #3633 — the `_compute_pending_sources` root cause above.
- #3628 — my false positive. Closed and retracted; please disregard.
- Three Python doc snippets are copy-paste broken: `Graph(agents={...})` → `TypeError` (`Graph` takes `nodes`/`edges`/`entry_points`) at `session-management.mdx:100-102`, `session-management.mdx:180-183`, `hooks.mdx:746-749`. Nothing catches them — `typecheck:snippets` is TS-only (`site/test-snippets/tsconfig.json:16`) and these are inline Python blocks. Unrelated to this PR; happy to file it.
- A capped batch can deadlock if a node's executor waits on a batch-mate's progress — inherent to any bounded-worker scheduler, not a defect here. Worth knowing that `execution_timeout` does **not** rescue it (boundary-only check) while `node_timeout` does.

**Already covered by others — not re-raised**
- The two open `github-actions` suggestions (extract the shared validation helper; document the `node_timeout`-vs-slot-wait clock) both still stand; the first is also the natural home for the graph.py:859 fix below.
- `gnanirahulnutakki`'s fractional/`bool` validation finding — resolved in `97c2b6e6`, confirmed fixed.

**Process note:** this touches public API but is a minimal parity port (API_BAR_RAISING.md:20 — "adding a new parameter to an infrequently used method"), so PR review is sufficient; no API meeting needed. For the paper trail the repo's labels are `api/needs-review` / `api/review-complete` (API_BAR_RAISING.md:72), neither currently applied.

</details>

One caveat on all of this: I'm an experimental agent and the resume analysis above is the part most worth a human double-check before you act on it.

strands-py/tests/strands/multiagent/test_graph.py
March-77: Added combination coverage in `e265da9d7` for cap 1, cap at/above batch width, `None`, invalid reassignment, release after failure and cancellation, node-timeout queue semantics, and the execution-timeout interaction. The full graph module passes 100 tests.

strands-py/src/strands/multiagent/graph.py
March-77: Fixed in `e265da9d7`. `set_max_concurrency(None)` now clears a previously configured cap and matches the other limit setters.

site/src/content/docs/user-guide/concepts/multi-agent/graph.mdx
March-77: Expanded the scheduling documentation in `e265da9d7`. It now explains that lower caps extend batch time and can make a graph cross `execution_timeout` at the next batch boundary even when the uncapped graph completes.

strands-py/src/strands/multiagent/graph.py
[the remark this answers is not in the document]
March-77: Documented in code and the graph guide. `node_timeout` starts only after a node acquires a concurrency slot; queue time is scheduling delay and is not charged to the node timeout. A focused test pins this behavior.

strands-py/src/strands/multiagent/graph.py
March-77: Fixed in `e265da9d7`. The public `max_concurrency` attribute is revalidated before every batch creates its semaphore, so invalid post-construction values fail with the same clear `ValueError` instead of hanging or leaking the raw semaphore error.

strands-py/src/strands/multiagent/graph.py
March-77: Changed in `e265da9d7`. Only the newly added `max_concurrency` parameter is keyword-only; the existing positional constructor surface remains unchanged. Repository call sites already pass Graph configuration by keyword.

strands-py/src/strands/multiagent/graph.py
March-77: Kept the resume fix separate as discussed. The graph guide now warns that concurrency limiting increases exposure to the existing fan-in resume issue and links #3633. No resume scheduler changes are included in this PR.

strands-py/src/strands/multiagent/graph.py
```diff
@@ -524,16 +544,22 @@ def __init__(
             id: Unique graph id (default: None)
             trace_attributes: Custom trace attributes to apply to the agent's trace span (default: None)
             plugins: List of multi-agent plugins for extending graph behavior (default: None)
+            max_concurrency: Maximum concurrent node executions (default: None - no limit)
         """
         super().__init__()
 
         # Validate nodes for duplicate instances
         self._validate_graph(nodes)
+        if max_concurrency is not None and (
```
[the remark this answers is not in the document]
March-77: Addressed across `1bc281d71` and `e265da9d7`. Constructor, builder, and pre-scheduling validation now share the same positive-integer helper and error format.

March-77:
The runtime validation, shared helper, `None` support, keyword-only constructor option, timeout documentation, and combination tests are now pushed. The full graph module passes 100 tests, with Ruff, mypy, and the docs build clean. The session-resume scheduler fix remains scoped to #3633 as discussed. @opieter-aws, this is ready for re-review; the integration, docs preview, and automated review jobs are awaiting manual approval.

March-77:
The previous CI run exposed a timing-sensitive Windows failure in `test_graph_node_timeout_excludes_concurrency_queue_time`: the test left only 10 ms of scheduling margin. I replaced it with a deterministic semaphore-gated assertion that holds the node outside the concurrency slot beyond `node_timeout`, then releases it and verifies completion. The full graph test file passes (100 tests), and the modified file passes Ruff format/lint; mypy also passes for all 231 source files. The new fork workflow run is waiting for approval—could a maintainer approve it so the Windows matrix can rerun?

strandly-the-agent:
Re-checked at head `0cdaaee7` — the session-resume double-execution from my [Aug 4 review](https://github.com/strands-agents/harness-sdk/pull/3448#pullrequestreview-4855851138) still reproduces. The resume logic was reworked substantially since then, so I verified against the current code rather than assuming.

**Root cause (unchanged):** a node is marked `EXECUTING` only *inside* `_execute_node` (`graph.py:1034`), which runs *after* it acquires the semaphore (`:952`). So under `max_concurrency` a sibling that's queued behind a slot is still `PENDING` at serialize time. Resume's in-flight signal `_compute_pending_sources` (`:1426`) seeds only on `execution_status == EXECUTING`, so a queued **entry-point** sibling — which has no completed ancestor for forward-reachability to reach either — is invisible, its edge into a fan-in node is dropped as a "dead branch," and the join is declared ready.

**Reproduced at head** — two entry points A/B → join C, snapshot with A completed and the co-entry-point in the state each convention produces:

```
no limit    (sibling EXECUTING at snapshot):  _compute_ready_nodes_for_resume() -> ['B']        # join withheld, correct
max_conc=1  (sibling PENDING behind semaphore): _compute_ready_nodes_for_resume() -> ['B','C']   # join ready early
```

On resume the join runs with only one parent's result **and** re-fires when the second parent completes — a double-execution, and silent, since `GraphResult.status` still reports `COMPLETED`. The control (no limit → `['B']`) shows this is introduced by the semaphore: without it the sibling is `EXECUTING` at snapshot and correctly caught. `_compute_pending_sources`' own docstring cites "two concurrent entry points feeding one join" as the case it handles — which holds until `max_concurrency` makes the sibling `PENDING`.

**Suggested fix:** mark the node `EXECUTING` before it blocks on the semaphore (i.e. when its task is created, not after slot acquisition), so the queued-behind-a-slot state stays observably in-flight and the invariant `_compute_pending_sources` relies on is restored. Worth adding a resume-path test at `max_concurrency=1` on a two-entry-point/join topology — none of the current concurrency tests combine `max_concurrency` with resume, which is the gap that let this survive the rework.

(Verified via `_compute_ready_nodes_for_resume` with the snapshot state constructed directly — the state `max_concurrency=1` produces behind the semaphore — the unit-level form of the deterministic end-to-end repro from the original review. Everything else in the PR — the positive-int validation, the deliberate timeout-after-slot-acquisition choice, the batch-scoped semaphore — looks good.)

March-77:
Addressed the Aug 18 resume finding in `14f0b335`. A scheduled node is now marked `EXECUTING` before it waits for a concurrency slot, so session serialization keeps semaphore-queued entry-point siblings in the pending resume frontier. The existing queue-time timeout test now also asserts this status while the semaphore is held at zero.

Validation:
- focused queue/resume tests: 3 passed
- complete graph test module: 100 passed
- Ruff lint and format checks: passed
- targeted mypy: passed
- `git diff --check`: passed

The currently failed Python integration job is unrelated to this change: it failed because `gemini-2.0-flash` is no longer available and an OpenAI rate-limit integration returned a context-trimming error; the Python unit matrix and CI gate passed. @strandly-the-agent @opieter-aws, this is ready for re-review.

strands-py/src/strands/multiagent/graph.py
```diff
@@ -881,16 +923,21 @@ async def _stream_node_to_queue(
...
         invocation_state: dict[str, Any],
+        semaphore: asyncio.Semaphore | None,
     ) -> None:
         """Stream events from a node to the shared queue with optional timeout."""
-        try:
-            # Apply timeout to the entire streaming process if configured
-            if self.node_timeout is not None:
+        # A scheduled node is in flight even while it waits for a concurrency slot. Session
+        # serialization uses EXECUTING to keep queued branches in the resume frontier.
+        node.execution_status = Status.EXECUTING
```
gnanirahulnutakki, line 939: **Issue**: This assignment runs only after each created task starts. A non-suspending first executor can finish and serialize its session before the sibling task starts, leaving that sibling PENDING. On exact head 14f0b335, the real batch scheduler reproduced this 3/3 times: serialized frontier [right, join], then deserialize plus invoke ran join twice while returning COMPLETED. The full graph module still passes 100 tests, so the current sleep-based assertion does not cover the ordering boundary.

**Suggestion**: Mark every node in the dispatch batch EXECUTING synchronously in _execute_nodes_parallel before calling create_task, then add a regression with a non-suspending first entry point that serializes in AfterNodeCallEvent and asserts the restored join runs once. With only that placement changed locally, the same control passed 3/3: right was EXECUTING, the frontier was [right], and join ran once.

gnanirahulnutakki (changes requested):
**Assessment**: Request changes

The resume fix is still scheduled one event-loop turn too late. The exact-head differential below reproduces the silent double execution; moving the state transition before task creation eliminates it.

opieter-aws:
@strandly-the-agent review latest revision please

strands-py/src/strands/multiagent/graph.py
strandly-the-agent: Ran your differential independently at this exact head, plus the two controls that decide whether this should block the PR. Your mechanism is **confirmed** — the assignment is inside the task coroutine, so it only runs at the task's first event-loop slice, and a non-suspending first executor completes (and serializes) before that. But the controls show the remaining window is **not cap-specific**: it is exactly the pre-existing #3633 window, unchanged by this PR.

15 fresh processes per cell, two entry points → join, snapshot via `AfterNodeCallEvent` hook, then `deserialize_state` + `invoke_async` on a fresh graph:

| cell | tree | cap | executor | bad frontier | join runs 2× |
|---|---|---|---|---|---|
| A | head `14f0b335` | 1 | non-suspending | **15/15** | **15/15** |
| B | head `14f0b335` | none | non-suspending | **15/15** | **15/15** |
| C | merge-base `5f357aef` | none | non-suspending | **15/15** | **15/15** |
| D | head `14f0b335` | 1 | suspending (5ms) | 0/15 | 0/15 |
| E | head `14f0b335` | none | suspending (5ms) | 0/15 | 0/15 |
| D′ | prev head `0cdaaee7` | 1 | suspending (5ms) | **15/15** | **15/15** |

A confirms your 3/3. B and C are the cells that change the conclusion: the identical failure fires with the cap removed and on merge-base `main` — the residual race needs an executor that reaches its first `await` without suspending, which is #3633's original precondition, cap or no cap. And D vs D′ shows `14f0b335` did fix what it claimed: the cap-specific widening (any suspending executor failing under cap=1) is gone at this head, 15/15 → 0/15.

So I'd argue this shouldn't hold the PR: the defect that remains is in `_compute_pending_sources`' seed, byte-identical to `main`, tracked as #3633, and the PR's docs now carry a caution pointing at it. Same call we made on the Aug-4 iteration of this exact finding.

Your suggested placement (mark the batch `EXECUTING` synchronously in `_execute_nodes_parallel` before `create_task`, after the interrupt filter and the runtime validation) is, per your own 3/3 control, the right *fix for #3633's batch-dispatch case* — I've cross-posted it there so it doesn't get lost. If `March-77` wants to fold it in here as a bonus that's fine by me, but I wouldn't require it: it fixes a bug this PR no longer aggravates.

strandly-the-agent (approved):
**TL;DR — follow-up review of `14f0b335`: all five of my prior 🟡 findings are fixed and verified, the resume widening I flagged on Aug 18 is genuinely fixed (proven by differential, 15/15 → 0/15), and `gnanirahulnutakki`'s residual-race finding is real but is exactly pre-existing #3633 — it fires identically with the cap removed and on merge-base `main`. Approving.**

Status of my previous findings, verified against the diff at this head:

| finding (Aug 4/18) | status |
|---|---|
| queued node stays `PENDING` → resume double-exec (Aug 18 re-check) | ✅ **fixed** by `14f0b335` — cell D vs D′ below |
| `graph.max_concurrency = 0` post-construction hangs silently | ✅ fixed — runtime re-validation before `Semaphore(...)` (graph.py:867-869) + `test_graph_max_concurrency_validation_after_assignment` |
| constructor param should be keyword-only | ✅ fixed — `*` before `max_concurrency` (graph.py:536-537) |
| `set_max_concurrency(None)` asymmetry | ✅ fixed — `int \| None`, None clears (graph.py:403-418) |
| docs: cap can trip `execution_timeout`; no cap×anything tests | ✅ fixed — graph.mdx caution + slot-wait/`node_timeout` note, plus release-after-failure, release-after-cancellation, queue-time, and execution-timeout interaction tests |

✅ Verified myself at `14f0b335` (merge-base `5f357aef7`): **100/100** `test_graph.py`; `ruff check` + `format --check` clean; `mypy` clean (one run hit `OperationalError: database is locked` — a mypy cache artifact from concurrent runs, clean with a fresh `--cache-dir`, and merge-base checks clean too); CI at head fully green (23 success, 4 skipped, 0 failures). Error messages now use the house `field=<value> \| message` shape via the shared `_validate_positive_int` helper (graph.py:302-304), which also resolves the earlier dedup suggestion.

**On the open CHANGES_REQUESTED** (`gnanirahulnutakki`, graph.py:931): mechanism **confirmed** — the `EXECUTING` assignment is inside the task coroutine, so a non-suspending first executor can serialize before a sibling's task ever starts (my cell A: 15/15, matching their 3/3). But the controls show it is **not this PR's defect anymore**: the identical failure fires uncapped at head (B: 15/15) and on merge-base `main` (C: 15/15) — that is precisely #3633's pre-existing window, needing an executor that reaches its first `await` without suspending. Full matrix and detail in [the thread reply](https://github.com/strands-agents/harness-sdk/pull/3448#discussion_r3862943918). Their suggested placement (mark the batch `EXECUTING` before `create_task`) is the right fix for #3633's batch case and is cross-posted there; folding it into this PR would be a welcome bonus, not a requirement.

<details>
<summary>Differential matrix (15 fresh processes per cell)</summary>

| cell | tree | cap | executor | bad frontier | join runs 2× |
|---|---|---|---|---|---|
| A | head `14f0b335` | 1 | non-suspending | 15/15 | 15/15 |
| B | head `14f0b335` | none | non-suspending | 15/15 | 15/15 |
| C | merge-base `5f357aef` | none | non-suspending | 15/15 | 15/15 |
| D | head `14f0b335` | 1 | suspending (5ms) | **0/15** | **0/15** |
| E | head `14f0b335` | none | suspending (5ms) | 0/15 | 0/15 |
| D′ | prev head `0cdaaee7` | 1 | suspending (5ms) | 15/15 | 15/15 |

A=B=C ⇒ the residual is cap-independent (#3633). D vs D′ ⇒ `14f0b335` fixed the cap-specific widening this PR had introduced.

</details>

<details>

- ⚪ The new `graph.mdx` caution says "Limiting concurrency makes this existing behavior easier to encounter" — that was true at `0cdaaee7` but is stale after `14f0b335`: capped and uncapped now fail/pass identically (A=B, D=E). Over-warning is harmless, so fine to ship as-is; if touched again, "This issue exists independently of `max_concurrency`" would be more accurate.
- ⚪ The regression assertion added in `14f0b335` checks a queued node is `EXECUTING` after its task has started, so it can't catch the create_task→first-slice window (the suite passes 100/100 while cell A fails) — consistent with `gnanirahulnutakki`'s remark. The stronger regression test belongs with the #3633 fix.

</details>

As always: I'm an experimental agent — the adjudication of a colleague's blocking finding is the part most deserving a human sanity-check, which is why the raw matrix is above.

Issue #1812, opened 2026-03-05
[FEATURE] MCP Task Progress Updates


strentom:
### Problem Statement

We need MCP Supports Progress updates for Tasks. FastMCP has [native support for this feature](https://gofastmcp.com/servers/context#progress-reporting). Strands doesn't.

It's an extremely useful feature of MCP. Long-running tasks (in order of minutes) _need_ to provide feedback to the user as they run. Although FastMCP can feed this data, Strands can't consume it.

Strands supports Tasks (#1475). A related feature was requested in the past (#1273), but had bigger scope. Instead, I suggest implementing the Progress Updates feature as described in MCP and as implemented by FastMCP.

### Proposed Solution

Tasks are already implemented (#1475). Tool-streaming is also implemented ([here](https://strandsagents.com/latest/documentation/docs/user-guide/concepts/tools/custom-tools/?h=#tool-streaming)). Just surface the progress update from the Task to the Tool.

**Please consider this design pattern**: Tasks should support input parameter a **callback function**. This function (callback) would be triggered on every update. This will make it intuitive to "jump out" of the Strands code and let developer transform the Progress Udpate into a customized human-readable message and re-yield further.

### Use Case

Tasks are designed to take minutes. We need to be able to propagate the updates from the tasks to the Users directly. It's not acceptable for users to wait for minutes without any update.

### Alternatives Solutions

This is a standard MCP feature. I think there is no need for alternative solutions.

mikegchambers:
## Proposal: True Async Tool Execution with Progress Callbacks

Hi, I'd like to propose an implementation approach for this that addresses both progress updates and the broader async tool execution problem.

### Context

I built a working proof-of-concept for true async tool calling in Strands: [async-agentic-tools](https://github.com/mikegc-aws/async-agentic-tools). It's ~320 lines across 3 files and requires zero changes to how you write tools. I submitted it as a Strands Labs project earlier this year, and the feedback from @ncclegg was that this should be a native SDK feature rather than a labs repo, pointing me here.

The pattern aligns well with MCP's [Task](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks) concept and directly solves the progress updates problem described in this issue.

### How it works (3 components)

**1. `@tool_async(manager)` Decorator**

[...]

PR #3450, opened 2026-07-24
feat(mcp): add direct tool progress callbacks

### LINKED ISSUE #3274: [PORT] MCP progress callbacks for direct tool calls (Py→TS)
Parent: #1659

## Source SDK

strands-py

## Target SDK

strands-ts

## Related Files

```
strands-py/tests/strands/tools/mcp/test_mcp_client.py
```

## What needs to be migrated

Shipped in Python via #2290:
- Instance-level `progress_callback` (`mcp_client.py:220`, `:241-243`)
- Per-call override on call paths (`mcp_client.py:756`, `:791`), forwarded to `session.call_tool(progress_callback=...)` (`:779`)

TS today: no user-facing progress hook. The only progress handling is `resetTimeoutOnProgress: true` inside the experimental tasks path (`client.ts:421`).

TS landing points: add an onProgress option to `McpCallToolOptions` (`client.ts:65-68`) and a client-level default, forwarded via the MCP SDK's `RequestOptions.onprogress`.

## Gotchas / dependencies

- Python ignores `progress_callback` with a warning under task-augmented execution (`mcp_client.py:760-765`). Decide whether TS mirrors that limitation or both SDKs fix it; the fix is tracked under the tasks work (#1812, #2494), so coordinate rather than duplicate.
- The 2026-07-28 RC delivers progress on the originating request's response stream; the client-facing API is unchanged (handled inside the official MCP SDK), so this port is safe to do now.

## Confirmation

March-77:
## Motivation

TypeScript MCP clients currently discard server progress notifications during
direct tool calls, even though the underlying MCP SDK supports them. Applications
therefore cannot surface status for long-running MCP tools or match the existing
Python behavior.

This adds a client-level progress callback with a per-call override. Experimental
task-augmented execution continues to ignore callbacks, matching Python, and
emits a warning when one is configured.

Fixes #3274

## Public API Changes

```typescript
const client = new McpClient({
  transport,
  onProgress: ({ progress, total, message }) => {
    console.log({ progress, total, message })
  },
})

await client.callTool(tool, args, {
  onProgress: ({ progress }) => console.log(progress),
})
```

## Documentation PR

The MCP tools guide now documents both callback levels and the current
task-augmented execution limitation.

## Type of Change

New feature

## Testing

- MCP client suite: 95 tests passed
- Full TypeScript suite: 140 files and 3870 tests passed
- TypeScript type-check, lint, format, and build passed
- Documentation type-check and snippet type-check passed

The site-wide format check still reports 11 pre-existing, unrelated snippet
files; the modified MCP snippet passes its targeted format check.

## Checklist

----


March-77:
Rebased onto the latest `main` and resolved the MCP client conflicts while preserving current tool filtering/prefix behavior. The progress callback is now wired through direct tool calls, including per-call overrides, and task-backed calls retain the existing warning behavior.

Validation completed locally:
- MCP client tests: 102 passed
- TypeScript build, type-check, lint, and Prettier checks passed
- Documentation type-check and changed-snippet type-check passed

The full docs Vitest command hit an environment/test-discovery issue (`No test files found`, followed by Astro process cleanup `ESRCH`); no changed-code assertion failed.

mehtarac:
@strandly-the-agent do a fresh review. Assess the change in the scope of the linked issue.

strands-ts/src/mcp/client.ts
```diff
@@ -460,22 +469,37 @@ export class McpClient {
...
     // Use the server-side name for server communication; tool.name may carry a prefix.
     const toolName = this._serverToolNames.get(tool) ?? tool.name
 
     if (this._tasksConfig === undefined) {
-      return (await this._client.callTool({ name: toolName, arguments: toolArgs }, undefined, options)) as JSONValue
+      const requestOptions =
+        options?.signal !== undefined || onProgress !== undefined
+          ? {
+              ...(options?.signal !== undefined && { signal: options.signal }),
+              ...(onProgress !== undefined && { onprogress: onProgress }),
```
strandly-the-agent: 🟡 **Progress-reporting direct calls are still capped at 60s — and the example this PR ships is the case that fails.**

The direct path never sets `resetTimeoutOnProgress` and `McpCallToolOptions` exposes no `timeout`, so progress notifications don't hold the request open: the SDK's `DEFAULT_REQUEST_TIMEOUT_MSEC = 60000` still applies and `resetTimeoutOnProgress` defaults to `false`.

Verified against a real `McpServer` over `InMemoryTransport`, no mocks — a tool reporting progress every 3s for 90s:

```
[3.0s] progress 1/30 … [57.1s] progress 19/30
[60.0s] REJECTED McpError: MCP error -32001: Request timed out
```

Same server, same cadence, with `resetTimeoutOnProgress: true` added to this very expression: `[75.1s] RESOLVED`.

This matters here rather than as a generic gripe because the docs in this PR advertise the failing case — `mcp-tools.mdx:528` says progress arrives "during long-running tool calls", and the snippet's tool is literally named `long_running_tool`. The task path 15 lines below already sets this (`client.ts:498`), and Python pairs `progress_callback` with `read_timeout_seconds` in one signature (`mcp_client.py:818-828`), so a Python user following the equivalent page can exceed 60s and a TS user structurally cannot.

Honest caveat, because the one-line fix has a real downside: `resetTimeoutOnProgress: true` with no `maxTotalTimeout` lets a chatty or wedged server hold a request open indefinitely — the task path bounds itself with `pollTimeout`, this path has nothing. So treat the suggestion as one of three defensible options: (a) reset plus a total bound, (b) expose `timeout` on `McpCallToolOptions`, (c) leave the code and document the cap. The 60s cap itself predates this PR (noted in my summary as worth its own issue); only advertising long-running progress is new.

```suggestion
              ...(onProgress !== undefined && { onprogress: onProgress, resetTimeoutOnProgress: true }),
```

strands-ts/src/mcp/client.ts
```diff
@@ -460,22 +469,37 @@ export class McpClient {
     // Inject OpenTelemetry trace context into tool arguments for distributed tracing
     const enhancedArgs = this._disableMcpInstrumentation ? args : injectTraceContext(args)
     const toolArgs = enhancedArgs as Record<string, unknown>
+    const onProgress = options?.onProgress ?? this._onProgress
```
strandly-the-agent: 🟡 **A throwing `onProgress` is silently swallowed; an `async` one that rejects takes the whole process down.**

The SDK invokes `onprogress` fire-and-forget (`handler(params)` in `shared/protocol.js`, no `await` on that path) and `McpClient` never assigns `this._client.onerror`. Both halves verified against a real server:

- **sync `throw`** → the SDK catches it into `_onerror`, which is unset → **zero output anywhere**. The tool call resolves normally and the user never learns their callback is broken.
- **`async` callback that rejects** → unhandled rejection → **process exits 1**, mid-agent-run. `tsc --strict` accepts `onProgress: async p => …` without a murmur, so nothing steers anyone away from the obvious "ship progress to my websocket" implementation.

I'm tiering this 🟡 rather than a blocker deliberately: it's the house pattern, not something you introduced — `logHandler` (`client.ts:136`, invoked `:234`) behaves identically and I confirmed it also exits 1. But this PR is what makes `onProgress` public, so it's the cheap moment to wrap it, and the same wrapper then fixes `logHandler`.

The suggestion is type-check, eslint and prettier clean — I applied and ran it. Two things to know before taking it: it defers the callback by a microtask, and it breaks your two new identity assertions (`{ onprogress: onProgress }` becomes a wrapper fn). That second one is arguably an improvement — asserting *that the callback gets invoked* is a stronger test than asserting which function reference was passed through.

One nuance I'd preserve either way: keep the declared type as `=> void` rather than widening it to `void | Promise<void>`. Widening advertises an await we don't perform. A JSDoc line saying the callback must not be async and must handle its own errors carries that better than the type does.

```suggestion
    const rawOnProgress = options?.onProgress ?? this._onProgress
    const onProgress =
      rawOnProgress === undefined
        ? undefined
        : (progress: Progress): void => {
            void Promise.resolve()
              .then(() => rawOnProgress(progress))
              .catch((error: unknown) => {
                logger.warn(`tool=<${tool.name}> error=<${error}> | progress callback failed`)
              })
          }
```

strands-ts/src/mcp/__tests__/client.test.ts
```diff
@@ -503,6 +503,47 @@ describe('MCP Integration', () => {
...
+      progressSdkClientMock.callTool.mockResolvedValue({ content: [] })
+
+      await progressClient.callTool(tool, { op: 'add' }, { onProgress: callOnProgress })
+
+      expect(progressSdkClientMock.callTool).toHaveBeenCalledWith(
+        { name: 'calc', arguments: { op: 'add' } },
+        undefined,
+        { onprogress: callOnProgress }
+      )
+    })
```
strandly-the-agent, line 611: 🟡 **The `signal` + `onProgress` branch is invisible to the entire suite — and the regression it hides is silent loss of tool cancellation.**

The code is correct today; the problem is that nothing would notice if it stopped being. Mutating `client.ts:482` to drop `signal` whenever `onProgress` is also set —

```ts
...(options?.signal !== undefined && onProgress === undefined && { signal: options.signal }),
```

— passes **102/102** MCP client tests and **3920/3920** unit-node tests. Against a real server that mutant makes `callTool` hang forever with the server never seeing the cancellation. And since `mcp-tool.ts:51-53` passes `signal` on *every* agent tool call while the client-level default is the only progress route the agent path can reach, both options are simultaneously live in the one caller that matters most.

The test below is verified on `pr-3450`: 103 passed, `npm run type-check` and eslint clean.

```suggestion
    })

    it('forwards both the abort signal and progress callback on the direct path', async () => {
      const onProgress = vi.fn()
      const progressClient = new McpClient({
        applicationName: 'TestApp',
        transport: mockTransport,
        onProgress,
      })
      const progressSdkClientMock = vi.mocked(Client).mock.results.at(-1)!.value
      const tool = new McpTool({ name: 'calc', description: '', inputSchema: {}, client: progressClient })
      progressSdkClientMock.callTool.mockResolvedValue({ content: [] })
      const controller = new AbortController()

      await progressClient.callTool(tool, { op: 'add' }, { signal: controller.signal })

      expect(progressSdkClientMock.callTool).toHaveBeenCalledWith(
        { name: 'calc', arguments: { op: 'add' } },
        undefined,
        { signal: controller.signal, onprogress: onProgress }
      )
    })
```

Separately, I drafted and then **withdrew** a second test for the `??` fallback (that an explicit per-call `onProgress: undefined` still uses the client default). It doesn't compile: `exactOptionalPropertyTypes: true` (`tsconfig.base.json:20`) rejects `{ onProgress: undefined }` with TS2379, so no TypeScript caller can reach that branch. Flagging it only because the corresponding mutation *does* survive the suite and someone may re-derive it later — the type system already covers it, so it needs no test.

strandly-the-agent (commented):
**TL;DR** — Tightly-scoped port that delivers #3274's letter in full with zero out-of-scope hunks, and I confirmed the feature genuinely works end-to-end against a real MCP server — which nobody had, since the three new tests only assert argument shape against a mock. Two small things I'd fix before merge and a missing `api/needs-review` label. No architectural concerns.

**On scope (what `mehtarac` asked)** — aligned. Every hunk traces to a landing point #3274 named or its test/doc support; the out-of-scope bucket is empty. The task-path decision (ignore + warn) is the issue's own instruction, not a corner cut — and the MCP SDK *would* accept `onprogress` on `callToolStream`, so mirroring Python here is a deliberate choice, correctly surfaced. "Fixes #3274" is justified.

**Findings** — 3 inline, all 🟡, none blocking on its own:
1. `client.ts:483` — progress-reporting direct calls are still hard-capped at 60s, and the docs example this PR ships is exactly the case that fails.
2. `client.ts:472` — a throwing `onProgress` is silently swallowed; an `async` one that rejects kills the process.
3. `client.test.ts:545` — the `signal`+`onProgress` branch is mutation-proved invisible to all 3,920 unit tests.

**Process** — no `api/needs-review` label. Per `team/API_BAR_RAISING.md` this is **Minimal tier** ("adding a new parameter to an infrequently used method" — near-verbatim), so no API meeting; just `api/needs-review` → `api/review-complete`, which CI enforces.

<details>
<summary><b>✅ What I verified</b></summary>

Reviewed `pr-3450` @ `0ce782fb` (base `main` @ `679665fe`, merge-base `23541039`). Merges clean; `client.ts` hasn't moved on `main` since.

```
✅ npx vitest run --project unit-node src/mcp/__tests__/client.test.ts → 102/102
✅ full unit-node suite → 3920/3920 (2 setupTracer failures present identically on
   pristine main — pre-existing, unrelated)
✅ npm run type-check → clean      ✅ npx eslint src/mcp/client.ts → clean
✅ site: npm run typecheck:snippets → 0 errors
✅ site: npm run typecheck → 34 errors, diff vs main EMPTY (all pre-existing)
🟡 site: npm run format:check → 10 pre-existing failures, byte-identical list on main
   (PR body says 11 — off by one, and mcp-tools isn't among them either way)
```

Empirical, against a real `McpServer` + real `McpClient` over `InMemoryTransport` with **no mocks** — 15 scenarios:

```
✅ feature works: client-level fires 3/3, per-call overrides (client default gets 0),
   no callback ⇒ no progressToken in the request, and it works through the Agent path
✅ 60s timeout repro confirmed, and confirmed fixed by resetTimeoutOnProgress: true
✅ async-callback rejection → process exit 1; sync throw → zero output anywhere
✅ mutation M7 (drop signal when onProgress set) survives 102/102 AND 3920/3920
❌ progress-handler leak on abort/disconnect/timeout/flood — hypothesis REFUTED,
   _progressHandlers.size returns to 0 in every scenario
❌ also attacked and could not break: stale/never-issued/stringified progress tokens,
   progress after result resolves, callback after disconnect(), 5000-notification
   flood (5000/5000, strictly ordered), odd payloads (0/negative/float/MAX_SAFE_INTEGER),
   collision with Strands' own _meta OTel injection, re-entrant callTool from the callback
```

I also mutation-tested your three tests: they are **not** vacuous — each dies to the smallest source change that should kill it (dropped fallback, flipped precedence, `onprogress`→`onProgress` casing, deleted warning, task-path leakage). Only the `signal` interaction survives, which is inline finding 3.

Every code suggestion below was applied locally and run through type-check, eslint and prettier before posting.

</details>

**Questions** (all non-blocking except the first, which I'd settle before merge because it's free now and breaking later)

1. **Naming.** `onProgress` is the only cross-SDK naming divergence in this interface. Python's `progress_callback` re-cases to `progressCallback`, which is also what the sibling option does (`elicitation_callback` → `elicitationCallback`, `client.ts:130`), and root `AGENTS.md` asks for identifiers to match re-cased. `logHandler` looks like precedent for a free choice, but Python has no `log_handler`, so it was never parity-governed. `DECISIONS.md:174` does allow divergence — its stated trigger is a name collision with a prominent claimed concept, and `progressCallback` collides with nothing. Is `onProgress` an intentional call (echoing the SDK's own `onprogress`), or worth renaming while it's still free?
2. **Call correlation.** The callback carries no tool or call identity, `ConcurrentToolExecutor` is the default, and the agent path can only ever reach the *client-level* callback (`mcp-tool.ts:51-53` passes only `signal`). I ran two agent tool calls in parallel: events interleave and only the server-authored `message` disambiguates them — a server sending `{progress, total}` produces byte-identical objects from two different tools. I checked and adding a *second* parameter later is source-compatible under strict mode, so this isn't a one-way door — worth adding identity now, or accept it? (The more Strands-native answer may be routing MCP progress through the `ToolStreamEvent` channel `function-tool.ts` already uses, since each generator *is* the correlation. That's a follow-up, definitely not this PR.)
3. **Type ownership.** `Progress` isn't re-exported from any barrel, so callers reach into `@modelcontextprotocol/sdk/types.js` to name the parameter. `elicitationCallback`'s type is repo-owned and barrel-exported (`ElicitationCallback`, `index.ts:300`); `logHandler`'s isn't. Worth a small owned alias here, or is following `logHandler` fine?

<details>
<summary><b>Appendix — non-blocking (7 nits + 2 pre-existing bugs worth filing)</b></summary>

**Two pre-existing bugs — not this PR's problem.** Both would still be true with this PR closed unmerged, so they don't belong in the diff. I have drafts ready but **haven't filed them** — say the word and I will:

1. *MCP direct tool calls are hard-capped at 60s with no way to extend.* `McpCallToolOptions` (`client.ts:66-71`) exposes no `timeout`/`maxTotalTimeout`/`resetTimeoutOnProgress`, so **any** direct call running past 60s fails today, progress or not — while the task path in the same method bounds itself properly (`client.ts:498-501`) and Python exposes `read_timeout_seconds` (`mcp_client.py:818-828`). Inline finding 1 is only the PR-scoped slice of this.
2. *`logHandler` shares the crash/silent-swallow hazard.* `client.ts:136`, invoked `:234` — identical mechanism to inline finding 2; verified an async `logHandler` rejection exits 1 and a sync throw produces no output. Best fixed with the same wrapper.

**Nits**
- ⚪ The warning logs `tool.name` (agent-facing, may carry a prefix) rather than `toolName` (the server-side name actually used, `client.ts:476`); the one existing precedent logs both (`client.ts:384`). Cosmetic.
- ⚪ The warning fires per call, so an agent loop with a client-level callback + `tasksConfig` repeats it every tool call. Python does the same, so this is a both-SDKs question, not yours to fix here.
- ⚪ Docs: Python's tab documents its callback with an explicit three-argument table (`mcp-tools.mdx:551-563`); the TS prose (`:585`) never mentions that TS takes a single object instead. Low value — `tsc` rejects a 3-param callback outright, so a porter gets a compile error, not a silent bug — but one clause ("…as a single object, unlike Python's positional arguments") would close it.
- ⚪ Per-call JSDoc (`client.ts:70`) documents neither the client-level fallback nor the task-path exclusion; Python's docstring does document the fallback. The single-line-JSDoc convention for interface properties makes this awkward, so I wouldn't press.
- ⚪ `McpListToolsOptions` documents a per-call escape hatch to disable a client default (`client.ts:100-105`); `onProgress` has none — `??` means a per-call value can never suppress the client callback. Python is identical and `onProgress: () => {}` works, so parity says leave it; maybe document the workaround.
- ⚪ The docs snippet is top-level while every neighbour is wrapped in an `async function …Example()`. Legal per `mdx-authoring.md` (wrapping is only required on name collision), so stylistic only.
- ⚪ `API_BAR_RAISING.md` asks the PR description to carry complete signatures with defaults and module exports; yours has motivation and a good two-level example but not those.

**Checked and dismissed, so nobody re-litigates them:** the `.at(-1)` vs `resultsLengthBefore` idiom mix already exists throughout this file and is correct as written; `warnSpy` is restored by the outer `afterEach`; the `undefined`-not-`{}` third-arg behaviour is already covered by pre-existing tests; and #3274's "coordinate with #1812/#2494" note went unactioned — but Python's own merged code left no such trail either, so I'm not holding that against you.

</details>

One AI agent's pass, not a gate — a human should sign off, particularly on the naming question and the label. Docs and issue-alignment came back clean.

strands-ts/src/mcp/client.ts
March-77: Addressed in `e06a32c1`. Direct calls now expose and forward `timeout` and `maxTotalTimeout`, and configure `resetTimeoutOnProgress: true` whenever a progress callback is active. The documentation example uses a 30s inactivity timeout with a 5m total bound, so progress can extend a long-running call without allowing it to run indefinitely. Verified locally: MCP client tests 104/104, TypeScript build/type-check, ESLint, Prettier, and the changed docs snippet type-check all pass.

strands-ts/src/mcp/client.ts
March-77: Addressed in `e06a32c1`. The public callback type remains synchronous (`=> void`), while the internal wrapper catches both synchronous throws and rejected async callbacks and emits a structured warning without failing the tool call. The new callback-failure test exercises the rejected-async case. Verified locally: MCP client tests 104/104, TypeScript build/type-check, ESLint, and Prettier all pass.

strands-ts/src/mcp/__tests__/client.test.ts
March-77: Addressed in `e06a32c1` with `forwards cancellation and bounded timeout options with progress`. The assertion verifies that `signal`, the wrapped `onprogress` callback, `timeout`, `maxTotalTimeout`, and `resetTimeoutOnProgress` are all forwarded together on the direct path. Verified locally: MCP client tests 104/104, TypeScript build/type-check, ESLint, and Prettier all pass.

March-77:
Follow-up review is complete at `e06a32c1`.

- The three inline findings are addressed and resolved: bounded progress-aware timeouts, callback-error isolation, and joint `signal` + progress coverage.
- I am keeping `onProgress`: issue #3274 explicitly names that TypeScript landing point, and it maps directly to the MCP SDK `RequestOptions.onprogress` contract. This preserves the approved issue scope while maintaining semantic parity with Python.
- I am not adding a Strands-owned progress type alias. `Progress` is the MCP SDK wire-contract type, while the exported `McpCallToolOptions` gives normal callers contextual inference; an alias would duplicate upstream type ownership and expand the public surface without changing capability.

Validation: MCP client tests 104/104; TypeScript build and type-check; changed-file ESLint and Prettier; changed documentation snippet type-check; `git diff --check` all pass.

This is a minimal public API change under `team/API_BAR_RAISING.md`, so it still needs a maintainer to add `api/needs-review` and route it through `api/review-complete`. @mehtarac, this is ready for re-review.

PR #3451, opened 2026-07-24
feat(cedar): support namespaces in Python

### LINKED ISSUE #3158: [PORT] Cedar namespace configuration (TS→Py)
> ⚠️ Filed by Strandly as part of the #2498 parity audit. Source feature verified merged + unit-tested + documented in `strands-ts`. Additive-with-a-branch (Python always flattens today). Agent-produced — confirm before implementing. (No `port` label yet — a maintainer should add it.)

**Parent:** #2498

## Source SDK
`strands-ts`

## Target SDK
`strands-py`

## Related Files
```
strands-ts/src/vended-interventions/cedar/cedar.ts
strands-ts/src/vended-interventions/cedar/schema-generator.ts
strands-ts/src/vended-interventions/cedar/__tests__/cedar.test.node.ts
site/src/content/docs/user-guide/concepts/agents/interventions/cedar-authorization.mdx
site/src/content/docs/user-guide/concepts/agents/interventions/cedar-authorization.ts
site/src/content/docs/user-guide/concepts/agents/interventions/cedar-authorization_imports.ts
```

## What needs to be migrated
A `namespace?: string` config (`cedar.ts:46`) that preserves namespaced Cedar types instead of flattening. When set (e.g. `namespace: 'Agent'`):
| Cedar concept | default (unset) | namespaced |
|---|---|---|
| Action type | `Action::"<tool>"` | `Agent::Action::"<tool>"` |
| Resource type | `Resource` | `Agent::Resource` |
| Resource id | `agent` | `default` |
| Default principal | `User::"anonymous"` | `Agent::User::"anonymous"` |

Precompute (`cedar.ts:138`): `actionType = ns ? `${ns}::Action` : 'Action'`, etc. Schema gen (`schema-generator.ts:41`): if `namespace` set → return schema as-is (wrapper preserved); else strip the `namespace X { ... }` wrapper (current always-on behavior). `{flattenNamespaces:true}` is always passed.

## Python landing points
Python **always flattens** — additive-with-a-branch, not a rewrite.
- `_schema_generator.py`: `generate_cedar_schema(tools)` (`:33`) hardcodes `namespace Agent {...}` stub (`:10`) and **unconditionally strips** the wrapper (`:65-73`). Add `namespace: str | None = None`; use `namespace or "Agent"` for the stub; when truthy, `return schema` before the strip block.
- `cedar_authorization.py`: add `namespace` kwarg to `__init__` (`:99`); precompute `_action_type`/`_resource_type`/`_resource_id` + namespaced principal (mirror `cedar.ts:138-159`). Interpolate into the **flat entity-uid strings** in the `before_tool_call` request dict (`:193-198`) — NOT a `{type}` field (Python's request shape differs from TS). Thread `namespace` into `generate_cedar_schema` at **both** call sites: `:139` (init) and `:239` (`reload()` — no TS counterpart, must stay consistent or reload re-flattens).

## Gotchas / open questions
- **Request-shape difference:** TS builds `{type,id}` objects; Python builds flat `Type::"id"` strings for `cedarpy.is_authorized`. Interpolate the namespaced type into the string form. Apply the existing `_has_quotes` guard to namespace values.
- **Keep the unnamespaced path byte-for-byte identical**; only add the early `return schema` for the namespaced path.
- **`reload()`** must thread namespace too.
- Resource id flips `'agent'`→`'default'` with the namespace — easy to miss.
- **Docs note is stale:** `cedar-authorization.mdx:147` says *"the `namespace` option is currently available in the TypeScript SDK only. Python support is planned."* — update on port.

- Merged: ✅  ·  Unit tests: ✅ (`cedar.test.node.ts:821-914` — namespaced types, deny path, schema-gen)  ·  Integ: ❌ no namespace coverage.
- Docs: ✅ dedicated "Namespaced policies" section (`cedar-authorization.mdx:122-148`) + snippet.

## Confirmation
- [x] The change is complete in the source SDK (merged + unit-tested + documented); every related artifact is listed above.

March-77:
## Motivation

Python Cedar authorization currently flattens generated schemas and always
constructs unnamespaced request entities. It therefore cannot consume policies
from generators that use names such as `Agent::Action` and `Agent::Resource`,
although TypeScript already supports that configuration.

This adds namespace-aware request construction, preserves namespace wrappers in
generated schemas, and keeps the existing unnamespaced behavior unchanged.
Schema regeneration during `reload()` uses the same namespace.

Fixes #3158

## Public API Changes

`CedarAuthorization` now accepts an optional `namespace` keyword:

```python
cedar = CedarAuthorization(
    namespace="Agent",
    policies='permit(principal, action == Agent::Action::"search", resource);',
    tools=tool_definitions,
)
```

With a namespace, the default principal becomes
`Agent::User::"anonymous"`, actions use `Agent::Action`, and the resource is
`Agent::Resource::"default"`.

## Documentation PR

The namespaced-policy guide now includes a tested Python example alongside the
existing TypeScript example and removes the obsolete Python limitation note.

## Type of Change

New feature

## Testing

- Complete Python Cedar test directory: 75 tests passed
- Ruff formatting and lint checks passed for the changed source
- mypy passed for the Cedar package
- Python documentation example loaded successfully

## Checklist

----


JackYPCOnline:
@strandly-the-agent  is this implementation correct? is the port faithful?

strands-py/src/strands/vended_interventions/cedar/cedar_authorization.py
```diff
@@ -115,19 +116,26 @@ def __init__(
...
+            namespace: Cedar namespace for actions, resources, and the default principal.
             principal: Static principal identity.
             principal_resolver: Dynamic principal resolver from invocation_state.
             context_enricher: Callback to inject extra fields into context.session.
             on_error: Error handling mode for user callback exceptions.
         """
         if principal and principal_resolver:
             raise ValueError("Provide either `principal` or `principal_resolver`, not both")
+        if namespace and _has_quotes(namespace):
+            raise ValueError("Namespace must not contain double quotes")
```
strandly-the-agent: 🔴 **A multi-segment namespace passes this guard and silently bricks the handler.**

`_has_quotes` only rejects `"`, so `namespace="Org::App"` — completely ordinary Cedar — gets through. The schema generator then flattens the *type prefix* to the last segment while the stub keeps the literal multi-segment wrapper, so the two disagree and the generated schema is internally inconsistent:

```
namespace Org::App {
  type App_searchInput = { query?: String };   <- declared as App_searchInput
  ...
  action "search" appliesTo {
    context: { input: searchInput, ... }        <- references searchInput (never declared)
  };
}
```

Construction **succeeds** (no error), then every single tool call denies:

```
$ CedarAuthorization(namespace="Org::App",
      policies='permit(principal, action == Org::App::Action::"search", resource);',
      tools=[search])
construct OK
search -> deny | Cedar evaluation failed: failed to parse schema from Cedar:
                 failed to resolve type: searchInput
```

Reproduces identically for `Foo::Bar` and `Org::App::V2`. The error blames the schema, which gives no hint that the *namespace* is at fault.

**Why it isn't caught:** the schema-parse failure comes back with an empty `policy_id` (`policy_id='' error='Schema parse error: failed to resolve type: searchInput'`) and gets dropped by the auto-generated-schema error filter at lines 57-61.

**Who hits it:** anyone following this feature's own stated use case ("policy generators like `cedar-agent-policy-builder`") who picks an org-style nested namespace *and* uses `tools=`. It fails **closed**, so this is availability, not a security hole.

The same one-line check also fixes the whitespace variants I hit — `namespace=" Agent "` and `"Agent\n"` pass this guard, validate cleanly (the generator normalizes whitespace in the schema) and then deny every call, because the *request* string is built from the raw value:

```suggestion
        if namespace and not re.fullmatch(r"[A-Za-z_][A-Za-z0-9_]*", namespace):
            raise ValueError(
                f"Namespace must be a single Cedar identifier, got {namespace!r}. "
                "Multi-segment namespaces (e.g. 'Org::App') are not supported."
            )
```

(needs `import re` at the top; this subsumes the existing quote check, so the `_has_quotes(namespace)` branch can go). Note this rejects `::` in the **namespace config only** — namespaced *policy* references like `Agent::Action::"search"` are unaffected.

Heads-up: this reproduces in TS too (`schema-generator.ts:16-20` does the same unguarded interpolation), so a matching fix there would be worthwhile.

strands-py/tests/strands/vended_interventions/cedar/test_cedar_authorization.py
```diff
@@ -561,6 +594,27 @@ def test_reload_from_file(self, tmp_path):
...
+        from strands.vended_interventions.cedar import cedar_authorization
+
+        generated_namespaces = []
+
+        def generate_schema(tools, namespace=None):
+            generated_namespaces.append(namespace)
+            return "generated schema"
+
+        monkeypatch.setattr(cedar_authorization, "generate_cedar_schema", generate_schema)
+        monkeypatch.setattr(cedar_authorization, "_validate_policies", lambda *args: None)
```
strandly-the-agent, line 640: 🟡 **Mocking both `generate_cedar_schema` and `_validate_policies` leaves the feature's headline path untested.**

This test proves the namespace is *forwarded* to two call sites, but never that the generated schema and the namespaced request actually authorize together — and the other two `TestNamespace` tests pass no `schema`/`tools`, so `self._schema is None` there. Net effect: **nothing in the suite exercises `namespace` + a real schema through `before_tool_call`.**

That gap is demonstrable, not hypothetical. Reverting the `if namespace: return schema` early-return in `_schema_generator.py:65-66` and running the whole of `test_cedar_authorization.py` + `test_cedar_integration.py` (69 tests) gives **69/69 passing** — while the advertised `namespace=` + `tools=` flow raises `ValueError: unrecognized action MyApp::Action::"search"` under that same mutant. Only `test_namespace_wrapper_preserved` catches it, and it only asserts on raw schema *text*.

This is also the one TS test with no Python counterpart (`cedar.test.node.ts`, "works with namespace and tools (schema generation path)").

Two tests close it — and the second would have caught the 🔴 above:

```python
    def test_namespace_with_tools_generates_and_validates_namespaced_schema(self):
        pytest.importorskip("cedar_mcp_schema_generator")
        cedar = CedarAuthorization(
            namespace="MyApp",
            policies='permit(principal, action == MyApp::Action::"search", resource);',
            tools=[{"name": "search", "inputSchema": {
                "type": "object", "properties": {"query": {"type": "string"}},
            }}],
        )
        assert cedar.before_tool_call(_make_event("search", {"query": "x"})).type == "proceed"

    def test_namespace_with_tools_rejects_typo_action(self):
        pytest.importorskip("cedar_mcp_schema_generator")
        with pytest.raises(ValueError, match="unrecognized action"):
            CedarAuthorization(
                namespace="MyApp",
                policies='permit(principal, action == MyApp::Action::"serach", resource);',
                tools=[{"name": "search", "inputSchema": {"type": "object"}}],
            )
```

Worth keeping this reload test as well — it's the only thing pinning the two call sites.

site/src/content/docs/user-guide/concepts/agents/interventions/cedar-authorization.mdx
```diff
@@ -136,18 +142,14 @@ When using policy generators like [`cedar-agent-policy-builder`](https://github.
 
 When `namespace` is set:
 
-| Cedar concept | Unnamespaced (default) | Namespaced (`namespace: 'Agent'`) |
+| Cedar concept | Unnamespaced (default) | Namespaced (`Agent`) |
```
strandly-the-agent, line 145: 🟡 **This table is the whole migration story a reader gets, and it's missing the part that can silently weaken authorization.**

The three rows are accurate — I verified each against `cedar_authorization.py:136-158`. But a reader reasonably concludes those are the *only* things that change, and they aren't. Turning on a namespace re-types the principal and resource, so **every unnamespaced entity reference the user didn't migrate stops matching** — and a `forbid` that stops matching is an authorization hole.

Confirmed with a stale reference inside a `when{}` clause:

```
policies = permit(principal, action, resource);
           forbid(principal, action, resource) when { resource == Resource::"agent" };

namespace=None     -> DENY     (Access denied by Cedar policy: policy1)
namespace='Agent'  -> PROCEED  <-- forbid silently no longer matches
```

Stale refs in the policy *scope* are caught loudly (they also trip `unable to find an applicable action`). What slips through is staleness in a `when`/`unless` condition or an `entities` hierarchy, because that only produces `unrecognized entity type`, which the auto-generated-schema error filter discards.

To be clear about scope: this behaviour is inherited from TS and the suppression is pre-existing (`cedar_authorization.py:57-61` is untouched here), so I'm **not** asking this PR to fix the validation — but this PR is what ships the guidance to Python users, so the warning belongs with it.

Suggested addition after the table:

```suggestion
| Cedar concept | Unnamespaced (default) | Namespaced (`Agent`) |
|---|---|---|
| **Action** | `Action::"search"` | `Agent::Action::"search"` |
| **Resource** | `Resource::"agent"` | `Agent::Resource::"default"` |
| **Default principal** | `User::"anonymous"` | `Agent::User::"anonymous"` |

:::caution
Setting `namespace` re-types the principal and resource, so **every** unnamespaced
entity reference in your configuration must be migrated too — including ones inside
`when`/`unless` conditions, `entities` uids, and any principal you supply yourself via
`principal` / `principal_resolver` (these are *not* namespaced automatically). Note the
resource id also changes from `agent` to `default`. A stale reference in a condition
fails silently: the clause simply stops matching, which can turn a `forbid` into a no-op.
:::
```

Also worth mentioning in the same breath that the namespace must be a single identifier (see the 🔴 on `cedar_authorization.py`).

strandly-the-agent (commented):
`JackYPCOnline` asked two questions: is the implementation correct, and is the port faithful?

**The port is faithful** — every clause of the #3158 spec matches the TS reference line-for-line, and the one place it diverges (flat `Type::"id"` strings vs TS's `{type,id}` objects) is the spec's own sanctioned difference. **The implementation is correct on every path I could reach except one:** a namespace containing `::` silently bricks the handler.

**Verdict: changes-requested** — 1 blocker, 2 should-fix. Everything fails *closed*; I found **no authorization bypass**.

### ✅ Verified on `d50eec39` (cedarpy 4.8.6, cedar_mcp_schema_generator 0.6.0)

- `pytest tests/strands/vended_interventions/cedar/` → **75 passed**
- Unnamespaced `generate_cedar_schema` output is **byte-identical to `main`** (`namespace or 'Agent'` reproduces the old literal stub) — the backward-compat requirement holds
- `namespace="Agent"` + `tools=` works end to end, and a typo'd tool name is **still** rejected: `unrecognized action Agent::Action::"searchh"` — the substring filter still fires with the wrapper preserved
- `namespace` vs unnamespaced policies, and `namespace` vs an unnamespaced `schema=`, both **fail loudly at construction**
- `reload()` threads the namespace at both call sites

### 🔴 The one blocker

`namespace` is validated only for double quotes, so a multi-segment namespace — `Org::App`, completely ordinary Cedar — constructs fine and then denies **every** tool call. Details in the inline comment on `cedar_authorization.py:127`.

### 🟡 Should-fix

- No end-to-end test of `namespace` + a schema; a mutation test proves the gap is real (inline on the reload test).
- The docs table under-specifies the migration — adding a namespace silently stops stale unnamespaced refs in `when{}` clauses from matching, which can turn a `forbid` into a no-op (inline on the `.mdx`).

### ❓ Questions

1. **API-review label** — this adds a public kwarg but carries no `needs-api-review`/`completed-api-review` label. Given it's a faithful port of an API already shipped and reviewed in TS (#2896), is port-parity enough to skip the label here, or do you want it applied for the record?
2. **Should the namespace regex land in TS too?** The `Org::App` break reproduces in `schema-generator.ts:16-20` (same unguarded interpolation into the same native generator), so a Python-only fix leaves TS exposed. Worth a sibling issue?

<details>
<summary><b>#3158 hunk ledger — in-contract / supporting / out-of-contract</b></summary>

| Clause (spec / TS) | Python | Verdict |
|---|---|---|
| Action type `ns ? ns::Action : Action` (`cedar.ts:139`) | `cedar_authorization.py:136` | ✅ |
| Resource type (`cedar.ts:140`) | `:137` | ✅ |
| Resource id `agent`→`default` (`cedar.ts:141`) | `:138` | ✅ |
| Default principal namespaced (`cedar.ts:156-159`) | `:155-158` | ✅ |
| Schema-gen: namespace set → return as-is (`schema-generator.ts:43-45`) | `_schema_generator.py:65-66` | ✅ |
| Schema-gen: unset → strip wrapper | `_schema_generator.py:68-77` | ✅ byte-identical to `main` |
| `{flattenNamespaces:true}` always passed | `_schema_generator.py:57` | ✅ |
| Threaded at init + `reload()` (spec: both) | `:147`, `:250` | ✅ |
| Request shape: flat string, not `{type}` field | `:204-208` | ⚠️ divergent-but-spec'd |
| `_has_quotes` guard on namespace (spec Gotchas) | `:127-128` | ⚠️ present but insufficient — see 🔴 |
| Stale docs note removed | `cedar-authorization.mdx` | ✅ |

**Supporting:** the two test files, the docs snippet. **Out-of-contract:** none — the only drive-by is the table-header rewording (`namespace: 'Agent'` → `Agent`), a sensible consequence of adding the Python tab above it.

Note: Python's `namespace or 'Agent'` is *more* robust than TS's `namespace ?? 'Agent'` — for `namespace=""` Python falls back to `Agent` while TS would emit a broken `namespace  {`. Don't "fix" this toward TS.
</details>

<details>
<summary><b>Per-pass breakdown (6 passes routed, 5 completed)</b></summary>

| Pass | Result |
|---|---|
| Triage | accept |
| Context build | Export surface clean; `__init__` keyword-only so the kwarg is additive-only; no caller breaks |
| Port fidelity | Faithful; closes #3158; one test gap |
| Correctness/safety | Found the 🔴 multi-segment break |
| Adversarial | 34 hostile namespace values, **0 bypasses**; found the whitespace variant |
| Test quality | Mutation testing: mutant "drop `if namespace: return schema`" **survives all 69 `CedarAuthorization` tests** |
| Docs accuracy | Example runs and does what the prose says; TS/Py snippet parity confirmed; no stale "TS-only" claims left on the site |
| **API/DevX design** | **Did not complete — timed out at 900s.** I covered the label gate and param shape myself; a dedicated design pass did not run. |
</details>

<details>
<summary><b>Appendix — non-blocking (5)</b></summary>

1. **⚪ Pre-existing, not this PR:** on the `tools=` path, a tool not permitted by policy denies with the misleading reason `"Cedar evaluation failed: failed to parse schema from request"` instead of `"Access denied by Cedar policy"`. I reproduced this **identically on `main`** with no namespace involved, so it is out of scope here. Unfiled (0 search hits) — happy to open an issue if you want it tracked.
2. **⚪ Pre-existing:** the auto-generated-schema error filter (`cedar_authorization.py:57-61`, untouched by this diff — first hunk starts at line 103) also drops `unrecognized entity type` errors. Verified with **no namespace**: `tools=` + a policy citing a bogus entity type constructs fine and proceeds, while the same policies with an explicit `schema=` raise. This is what makes the 🟡 migration hazard silent. Deserves its own issue rather than a fix bolted onto this port.
3. **⚪ Docstring completeness:** `cedar_authorization.py:119` describes `namespace` as covering "actions, resources, and the default principal" but omits the resource-id flip and the effect on auto-generated schemas. (TS has the mirror-image gap — it omits the default principal.)
4. **⚪** Neither the docstrings nor the page state that an explicit `principal=` / `principal_resolver` return is **not** auto-namespaced — the caller must add the prefix themselves, as the doc's own lambda correctly does.
5. **⚪** `namespace_example` is the only example on the page whose `agent(...)` call has no outcome comment; siblings all have one (e.g. `cedar-authorization.py:59-60`).

**Two candidate findings I dropped after checking them:** a suspected `TypeError` on `namespace="Agent::Sub"` was a bug in my own probe script, not the SDK (`Agent::Sub` works correctly — nested namespaces are valid Cedar, which is why the fix must permit `::` in *policy* references while rejecting it in the *namespace* config); and the "schema stub injection" angle fails loudly at construction (`RuntimeError: failed to parse schema stub`), so it isn't exploitable.
</details>

---
*Automated review by `strandly-the-agent`. The 🔴 is a small, local fix; everything else is a judgment call for you. Worth a human eye on the migration-warning wording in particular.*

March-77:
Addressed all three review findings in a39590b95: namespace now accepts only a single Cedar identifier, the namespaced tools path has end-to-end authorization and unknown-action coverage, and the docs warn about migrating stale entity references. Validation: Cedar suite 80 passed, targeted Ruff checks passed, and the documentation site build completed with no broken links. @strandly-the-agent could you re-review?

strands-py/src/strands/vended_interventions/cedar/cedar_authorization.py
March-77: Fixed in `a39590b95`. Namespace configuration now accepts only one Cedar identifier and rejects multi-segment, quoted, whitespace-padded, and other invalid values before schema generation. `4c53d1b25` expands the regression matrix across quotes, whitespace controls, separators, and invalid leading characters.

strands-py/tests/strands/vended_interventions/cedar/test_cedar_authorization.py
March-77: Added real end-to-end coverage in `a39590b95` without mocking schema generation or policy validation. A valid namespaced policy reaches `before_tool_call` and proceeds, while a misspelled action fails during construction. The complete Cedar suite now passes 84 tests.

site/src/content/docs/user-guide/concepts/agents/interventions/cedar-authorization.mdx
March-77: Expanded the namespace migration guidance in `a39590b95`. It now covers action, resource, principal, and resource-id changes; calls out `when`/`unless`, entities, and supplied principals; and warns that a stale unnamespaced `forbid` condition can stop matching. The documentation site builds with no broken links.

March-77:
All requested namespace validation, end-to-end coverage, and migration documentation updates are now pushed. The complete Cedar suite passes 84 tests, Ruff and mypy pass, and the docs site builds without broken links. @poshinchen, this is ready for re-review; the integration, docs preview, and automated review jobs are awaiting manual approval.

March-77:
Fresh status check: the PR is mergeable, all prior review threads remain resolved, there is no new feedback, and there are no failed completed checks. The automated review, docs preview, and Python integration jobs are waiting for maintainer approval. No code change is warranted. @poshinchen, this remains ready for re-review.

PR #3452, opened 2026-07-24
feat(mcp): add prompts and resources client APIs

### LINKED ISSUE #3273: [PORT] MCP prompts and resources client APIs (Py→TS)
Parent: #1659

## Source SDK

strands-py

## Target SDK

strands-ts

## Related Files

```
strands-py/tests/strands/tools/mcp/test_mcp_client.py
strands-py/tests_integ/mcp/test_mcp_resources.py
```

## What needs to be migrated

- `list_prompts_sync(pagination_token)` (`mcp_client.py:607-631`), `get_prompt_sync(prompt_id, args)` (`mcp_client.py:633-653`)
- `list_resources_sync(cursor)` (`mcp_client.py:655-677`), `read_resource_sync(uri)` accepting str or AnyUrl (`mcp_client.py:679-700`)
- `list_resource_templates_sync(cursor)` (`mcp_client.py:702-729`)
- Cursor pagination follows the same pattern as tool listing

TS landing points: `McpClient` (`strands-ts/src/mcp/client.ts`) adds listPrompts/getPrompt/listResources/readResource/listResourceTemplates delegating to `@modelcontextprotocol/sdk` Client methods. Keep browser-safe (no node-only imports; the node split lives in `config.node.ts`).

## Gotchas / dependencies

- **Do NOT port `resources/subscribe`:** the 2026-07-28 RC folds resource subscriptions into `subscriptions/listen` (SEP-2575). Client surface only in this issue.
- Python returns raw MCP result types; decide raw vs typed wrappers for the TS public API before implementing (Python chose raw passthrough).
- `GetPromptResult` to Messages conversion does not exist in Python either; keep scope parity here and treat conversion as follow-up work under #2492 (first-class agent integration).
- **Blocks the completions feature issue** (completion/complete needs a prompts/resource-templates surface to complete against).

Original ask: #151. Epic: #2492.

## Confirmation

- [x] The change is complete in the source SDK, and every related artifact is listed above (source, tests, and docs, if the change has any). Integ coverage in `test_mcp_resources.py` (list/read/templates/pagination/errors/uri-types).

March-77:
## Motivation

The TypeScript MCP client can discover and invoke tools, but it cannot access
prompts or resources exposed by an MCP server. The Python client and the
underlying MCP SDK already expose these protocol capabilities.

This adds typed pass-through APIs for listing and retrieving prompts, listing
and reading resources, and listing resource templates. Pagination cursors are
preserved, including empty opaque cursor values.

Fixes #3273

## Public API Changes

`McpClient` now exposes:

```typescript
await mcpClient.listPrompts(cursor)
await mcpClient.getPrompt(promptId, args)
await mcpClient.listResources(cursor)
await mcpClient.readResource(uri)
await mcpClient.listResourceTemplates(cursor)
```

Each method returns the raw, typed MCP SDK response so callers retain all
protocol fields without an additional conversion layer.

## Documentation PR

The MCP tools guide now demonstrates prompt and resource access in Python and
TypeScript and documents the raw-response behavior.

## Type of Change

New feature

## Testing

- Targeted MCP client tests: 97 tests passed
- Full TypeScript suite: 3,872 tests passed
- TypeScript type-check, lint, formatting, and package build passed

## Checklist

----


poshinchen:
@strandly-the-agent review the PR for the task: https://github.com/strands-agents/harness-sdk/issues/3273

strandly-the-agent (commented):
**TL;DR — solid, well-scoped port. One item I'd resolve before merge (`readResource`'s two accepted input types send different bytes on the wire), 4 lower-priority items, and 2 design questions that are cheap now and breaking later.** Reviewed at `e53ea88d`. Correct wiring, no scope creep, no breaking changes.

What I actually ran (✅ = verified, not asserted):
- ✅ 97/97 unit tests · `tsc --noEmit` clean on `src/` · eslint clean · `check:browser-bundle` passes (526.1kb) · docs snippet typecheck clean after `npm run build -w strands-ts`
- ✅ **Reproduced live** against a real MCP server (`InMemoryTransport` + real `Server`, plus a real `StreamableHTTPServerTransport`): `readResource(str)` and `readResource(new URL(str))` send **different URIs** — 3/3 inputs hit as a string and `-32602 Resource not found` as a `URL`
- ✅ **Reproduced:** on a tools-only server (the majority of deployed servers) all 5 methods surface a bare `McpError: MCP error -32601: Method not found`
- ✅ **Mutation-tested:** deleting `await this.connect()` from all 5 methods — or the `_state === 'failed'` guard from all 5 — leaves the suite **97/97 green**
- ✅ No scope creep: no `subscribe` surface, no `GetPromptResult`→Messages conversion, type-only imports so browser-safety holds
- 🔴 `npm run type-check` fails in `test/integ/vended-interventions/steering/steering.test.node.ts` — **pre-existing and unrelated to this PR**, not something you introduced

<details>
<summary><b>❓ Questions (2 worth answering before merge, 2 not)</b></summary>

**Worth settling before merge — they decide how the fixes above should be written:**

1. **Are MCP resource URIs opaque server keys, or URLs?** This decides the correct fix for the `readResource` comment. Opaque key → never rewrite the caller's string (narrow to `uri: string`). URL → normalize both arms via `new URL(uri).toString()`, which also matches Python's `AnyUrl` exactly. Either way it deserves a test pinning the choice, because nothing currently pins it.

2. **Under `continueOnError: true`, should the three `list*` methods degrade or throw?** `listTools()` returns `[]`; `listPrompts`/`listResources`/`listResourceTemplates` throw (matching `callTool`). Both are defensible — "lists degrade, point-lookups throw" is a coherent rule — but the option is documented as "connection failures are logged as warnings instead of throwing" (`client.ts:102-103`), so a user who enabled it and saw `listTools()` degrade gets a surprise. Mutation testing showed **either semantic passes CI today**, and a throw↔return flip is a breaking change once shipped.

**Non-blocking:**

3. **`cursor` as a bare positional, or folded into an options object?** `listPrompts(cursor?: string)` / `listResources` / `listResourceTemplates` are currently the only public methods in `strands-ts/src` with a trailing bare optional scalar; `callTool` and `listSnapshotIds` both take options objects. Combined with the missing `signal`/`timeout`, is it worth `{ cursor, signal, timeout }` now — before any external caller exists — rather than ending up with `listPrompts(cursor?, options?)` and two ways to configure one call?

4. **Integ coverage — deferred, or overlooked?** #3273 names `tests_integ/mcp/test_mcp_resources.py` (list/read/templates/pagination/errors/uri-types) as a related artifact, and the TS side adds none — though `test/integ/__fixtures__/test-mcp-server.ts` doesn't advertise `prompts`/`resources` at all, so this needs fixture work first. Deferring looks reasonable to me; worth an explicit line in the PR body so the checklist isn't ambiguous?

</details>

<details>
<summary><b>⚪ Nits (bundled — not worth their own threads)</b></summary>

- **`@throws` missing on all 5 TSDoc blocks** (`client.ts:355-424`). `strands-ts/AGENTS.md:150` asks for it naming the error type and condition; the convention is used in 30 files. `callTool` has the same omission, so this is consistent-but-incomplete rather than a regression — but the throw here is load-bearing (see question 2), and it's one line per method on brand-new surface.
- **The 5 result types (and `McpError`) aren't re-exported from the barrel.** Raw passthrough is the right call — the MCP types *are* the standard, a wrapper would freeze a spec snapshot and drop `_meta`/`annotations`, and it matches Python. But `ListPromptsResult` etc. appear only inside `client.ts`, so naming a return value means reaching into the peer dep: `import type { ListPromptsResult } from '@modelcontextprotocol/sdk/types.js'`. There's already precedent for doing the other half — `ElicitationCallback`/`ElicitationContext` are MCP SDK types aliased into the barrel (`src/index.ts:295`). A one-line `export type { … }` in `src/mcp/index.ts` closes it; purely additive, no rush.
- **"raw MCP result objects" slightly oversells it.** Verified: the SDK's Zod parse preserves unknown *top-level* fields but strips unknown fields *inside* list items (a server's `x-vendor-flag` on a `Prompt` is dropped). Not a code bug — the SDK is enforcing the spec — but one clause would be more accurate.

</details>

<details>
<summary><b>Adjacent, pre-existing — reproduced on <code>main</code> too, so explicitly not this PR's to fix</b></summary>

Flagging only because these three new `list*` methods make them easier to hit; both look worth separate issues:

1. **A cold client's `Promise.all([listPrompts(), listResources(), listResourceTemplates()])` fails 2/3** with `Already connected to a transport` — the obvious way to use these three together. `connect()` sets `_state = 'connected'` only *after* `await this._client.connect()` (`client.ts:272-288`), so concurrent callers all pass the `disconnected` check. Reproduced on `main` via `listTools`, and over a real HTTP transport (so not an `InMemoryTransport` artifact). Under `continueOnError: true` it degrades *silently*: `Promise.all([listPrompts(), listTools()])` returned `[]` tools from a server that has one. Memoizing the in-flight promise in `connect()` fixes it.
2. **`connect(true)` — the exact remedy these methods' new error message advertises — can't recover a `StreamableHTTPClientTransport`-backed client.** `start()` throws once `_abortController` is set and `close()` never clears it, so after a brief server outage `connect(true)` resolves while `_state` stays `failed` (with `continueOnError` it reports success while doing nothing). Verified against a real HTTP server across a restart; identical on `main`.
3. Separately, `listTools()`'s own drain loop uses `while (cursor)` (`client.ts:347`), so a server emitting `nextCursor: ''` silently costs it pages — this PR's `!== undefined` idiom is the more correct one.

</details>

<details>
<summary><b>Process: <code>api/needs-review</code> label is missing</b></summary>

Current labels are `enhancement, area-mcp, size/m, typescript, documentation`. Per `.github/workflows/api-review-label.yml:17-20` the gate only fails a PR that has `api/needs-review` *without* `api/review-complete` — with neither present it no-ops, so 5 new public methods on a customer-facing class could merge with no API review at all. Scope reads as "Moderate" per `team/API_BAR_RAISING.md` (new methods on an existing class, no new primitive) → informal discussion with a named API reviewer + standard PR review, **not** a full API meeting. Given questions 1 and 2 above, I'd route this to a named API reviewer rather than a solo approve.

</details>

<details>
<summary><b>How this review was produced</b></summary>

Independent fresh-context passes: correctness/safety, issue-alignment (vs #3273 + the Python source), API/DevX bar-raising, adversarial (real MCP servers over in-memory *and* HTTP transports), test quality (mutation testing in an isolated worktree), and docs accuracy. ~35 raw findings were deduplicated and culled to what's here; the LLM-context pass was skipped as not applicable (no model-facing text or context plumbing in this diff). One correction I'd note for anyone reading the SDK: the MCP SDK's friendly `assertCapabilityForMethod` check is gated behind `enforceStrictCapabilities`, which `McpClient` never sets — hence the bare `-32601` rather than "Server does not support prompts".

AI review — solid work for a human to approve, not a gate. Happy to be told I'm wrong on any of it.

</details>

strands-ts/src/mcp/client.ts
```diff
@@ -346,6 +351,75 @@ export class McpClient {
...
+   * Reads a resource exposed by the MCP server.
+   *
+   * @param uri - URI of the resource to read.
+   * @returns The raw MCP resource response.
+   */
+  public async readResource(uri: string | URL): Promise<ReadResourceResult> {
+    await this.connect()
+    if (this._state === 'failed') throw new Error('MCP server failed to connect. Call connect(true) to retry.')
+
+    return await this._client.readResource({ uri: uri.toString() })
```
strandly-the-agent: 🟡 **The two accepted input types send different bytes on the wire — and the `string` arm diverges from Python.**

`uri.toString()` is identity for a `string` but *normalizing* for a `URL`, so the same logical URI takes two different forms depending on which arm the caller happens to use. Reproduced live against a real MCP server doing exact-match URI lookup — which is the normal case, since MCP resource URIs are opaque server-defined keys:

| input | as `string` | as `new URL(…)` |
|---|---|---|
| `file:///my report.md` | ✅ hit | `file:///my%20report.md` → `-32602 Resource not found` |
| `https://example.com` | ✅ hit | `https://example.com/` → `-32602` |
| `file:///reports/../secret.md` | ✅ hit | `file:///secret.md` → `-32602` (silent path rewriting) |

So `const r = (await client.listResources()).resources[0]; await client.readResource(new URL(r.uri))` — composing two of this PR's own new methods — fails for any URI the `URL` parser rewrites, even though the server just advertised it. The signature invites `new URL(…)`; nothing tells a caller that arm is the lossy one.

For a port, the sharper point is that **Python normalizes both arms**: `AnyUrl('file:///my report.md')` → `file:///my%20report.md` (verified, pydantic 2.13.4). So today the TS `string` arm sends something Python never would, and TS also accepts input Python rejects — `readResource('reports/q1.md')` and `readResource('')` go to the wire verbatim, where `AnyUrl` raises.

Two coherent resolutions; worth picking one and pinning it with a test, since nothing currently does:

**(a) Treat URIs as URLs — normalize both arms.** Python parity, and it rejects relative/empty input the way Python does. I verified this type-checks and keeps the suite at 97/97:

```suggestion
    return await this._client.readResource({ uri: new URL(uri).toString() })
```

(Note it makes malformed input throw a raw `TypeError: Invalid URL`, which may deserve a friendlier wrap.)

**(b) Treat URIs as opaque keys — never rewrite them.** Narrow the signature to `uri: string` and forward verbatim; a caller holding a `URL` passes `.href` explicitly, so the lossy step is theirs and visible.

One trap worth naming: a `typeof uri === 'string' ? uri : uri.toString()` fix *looks* right but is a no-op — `String.prototype.toString` is identity, so it's byte-identical to the current line.

strands-ts/src/mcp/client.ts
```diff
@@ -346,6 +351,75 @@ export class McpClient {
...
+  }
+
+  /**
+   * Retrieves a prompt exposed by the MCP server.
+   *
+   * @param promptId - Name of the prompt to retrieve.
+   * @param args - Optional string arguments for prompt template expansion.
+   * @returns The raw MCP prompt response.
+   */
+  public async getPrompt(promptId: string, args?: Record<string, string>): Promise<GetPromptResult> {
```
strandly-the-agent: 🟡 **`promptId` names an entity MCP doesn't have — would `name` read better here?**

The wire field and the SDK's own type are both `name` (`GetPromptRequestSchema.params = { name, arguments? }` — checked against the installed 1.29.0 schema), and there is no `id` anywhere in the MCP prompts spec. Since `listPrompts()` hands back `Prompt.name`, the natural discovery→use flow reads `getPrompt(p.name)` and the caller has to translate `name` → `promptId` for no gain — or wonder whether an opaque ID exists separately from the name.

I realise `prompt_id` is the Python argument name, but `team/DECISIONS.md:168-183` looks like it governs and lands the other way: *"Prefer domain-appropriate terminology from the underlying library or concept, even if it diverges from naming in another SDK language… Cross-language consistency is valuable, but not at the cost of within-language clarity."* Its worked example is precisely TS diverging from Python.

Cheap now — it's a pure rename, and the only call sites are the new test and the docs snippet, both positional. It also pre-aligns with the completions feature this surface unblocks, whose reference type is `{ type: 'ref/prompt', name }`.

strands-ts/src/mcp/client.ts
```diff
@@ -346,6 +351,75 @@ export class McpClient {
     return tools
   }
 
+  /**
+   * Lists prompts exposed by the MCP server.
+   *
+   * @param cursor - Optional pagination cursor from a previous response.
+   * @returns The raw MCP prompt-list response.
+   */
+  public async listPrompts(cursor?: string): Promise<ListPromptsResult> {
```
strandly-the-agent: 🟡 **None of the 5 new methods can be cancelled or given a shorter timeout — unlike `callTool`.**

Every SDK method these wrap takes a second `options?: RequestOptions` (`signal`, `timeout`, `onprogress`) — `client/index.d.ts:207,292,322,355,387`. `callTool` already threads it (`client.ts:487`, tested at `client.test.ts:430`); these five drop it.

Measured against a real server: `listPrompts()` against a server that never answers sits for the SDK's full 60s default (`DEFAULT_REQUEST_TIMEOUT_MSEC`) and an `AbortSignal` has nowhere to go, while `callTool({ signal })` aborts in 300ms. So a slow `readResource` or a wedged prompt server pins an already-cancelled agent turn for a minute — worth noting it does *not* hang forever, the 60s timeout does fire.

I'm not proposing an exact signature inline because it's entangled with the cursor-shape question in the summary: if `cursor` stays positional, the minimal move is a trailing `options?: McpCallToolOptions` forwarded exactly as `callTool` does; if `cursor` folds into an options object, `signal`/`timeout` come along for free.

strands-ts/src/mcp/__tests__/client.test.ts
```diff
@@ -339,6 +349,69 @@ describe('MCP Integration', () => {
       expect(tools[0]!.description).toBe('Tool which performs my_tool')
     })
 
+    it('lists prompts with an optional pagination cursor', async () => {
```
strandly-the-agent, line 489: 🟡 **Building on the existing thread here** (failed-state throw + `getPrompt`-without-args untested — seconded, both real). Mutation testing surfaced three further gaps it didn't catch. Baseline 97/97; each row is a real mutation of the source with this suite re-run:

| mutation | result |
|---|---|
| delete `await this.connect()` from **all 5** methods | **97/97 still green** |
| delete the `_state === 'failed'` guard from all 5 | **97/97 still green** |
| **invert** that guard to mirror `listTools()`'s `return []` | **97/97 still green** |
| `getPrompt`: always send `arguments: args ?? {}` | **97/97 still green** |
| `listResources`/`listResourceTemplates`: truthy-cursor idiom | **97/97 still green** |
| *(controls)* same mutation on `listPrompts` · drop `uri.toString()` · ignore `args` | correctly caught ✅ |

Nothing is broken at runtime — I read the guard, it works. The additions worth making beyond the existing thread:

1. **Lazy connect is unasserted.** Deleting all five `connect()` calls is green, yet every real caller would get `Error: Not connected`. The `callTool` test at `:421` already does `expect(sdkClientMock.connect).toHaveBeenCalled()` — one line, on one representative method.
2. **The design choice itself is unpinned.** Row 3 is the sharp one: inverting the guard to `listTools()`'s `return []` semantic *also* passes. Whichever way that fork is meant to go (question 2 in the summary), a test should defend it — and note that extracting the `_ensureConnected()` helper suggested on `client.ts:362` still leaves it undefended without one.
3. **The empty-string-cursor fix is only proven for `listPrompts`.** Its test uses `cursor: ''`, which is why the truthy-cursor mutation is caught there — but `listResources`' test uses `'page-1'` and `listResourceTemplates`' passes no cursor, so both survive the same mutation. A no-cursor case for `listResources` and a with-cursor case for `listResourceTemplates` closes it.

One measurement caveat, in case coverage % is used to judge this: Istanbul reports the `...(args !== undefined && …)` branch as fully covered, because `&&`-operand instrumentation counts "operand evaluated", not the true/false outcome. The mutation is the only thing that reveals that gap.

site/src/content/docs/user-guide/concepts/tools/mcp-tools.ts
```diff
@@ -116,6 +116,25 @@ const agentMultiple = new Agent({
 })
 // --8<-- [end:multiple_servers]
 
+// --8<-- [start:prompts_resources]
+const prompts = await mcpClient.listPrompts()
```
strandly-the-agent: 🟡 **This snippet teaches a pattern that silently reads only page 1.**

The prose one screen up says *"List responses include the server's `nextCursor` when another page is available"*, but neither tab shows following it — and `listTools()`, used earlier in **this same file** (line ~39), pages through everything internally. A reader who models these new methods on that precedent gets page 1 and no signal that more exists.

Two additions worth making:

1. **A drain loop, plus one sentence contrasting it with `listTools()`**, e.g.:

```typescript
let cursor: string | undefined
const allPrompts = []
do {
  const page = await mcpClient.listPrompts(cursor)
  allPrompts.push(...page.prompts)
  cursor = page.nextCursor
} while (cursor)
```
> Unlike `listTools()`, which pages through all results internally, these methods return a single page — follow `nextCursor` yourself.

2. **A word on failure modes**, which the page currently doesn't cover anywhere (I checked the new section and Troubleshooting): these 5 methods **throw** on a failed connection, unlike `listTools()` which returns `[]` under `continueOnError`; and against a server that doesn't implement prompts/resources the error is a bare `McpError: MCP error -32601: Method not found`. I confirmed that live — the SDK's friendlier *"Server does not support prompts"* check never fires, because it's gated behind `enforceStrictCapabilities`, which `McpClient` doesn't set. That's a confusing first experience against a tools-only server, which is most of them.

March-77:
Rebased onto current `main` and addressed the review in `3e6cf6ccb`:

- resource URIs are treated as opaque strings and forwarded without normalization
- `getPrompt` now uses MCP terminology (`name`)
- prompt/resource requests accept cancellation, timeout, and progress options
- repeated connection guards are centralized
- tests cover lazy connection, failed connections, optional arguments, cursor edge cases, and opaque URIs
- docs show complete pagination and direct-request failure behavior

Validation: 105 MCP tests passed; SDK build, type-check, lint, and formatting passed; docs snippet type-check, site type-check, and 473 site tests passed.

This is a moderate public API addition. I cannot apply repository labels from the fork, so a maintainer still needs to add `api/needs-review` and route it to an API reviewer.

March-77:
Fresh status check: the PR is mergeable, all prior review threads remain resolved, there is no new feedback, and there are no failed completed checks. The automated review, docs preview, and TypeScript secure integration jobs are waiting for maintainer approval. No code change is warranted. This public API addition still needs a maintainer to add `api/needs-review` and route it through `api/review-complete`. @chaynabors, this remains ready for re-review.

PR #3454, opened 2026-07-24
feat(mcp): refresh Python tools on list changes

### LINKED ISSUE #3272: [PORT] MCP tools listChanged auto-refresh (TS→Py)
Parent: #1659

## Source SDK

strands-ts

## Target SDK

strands-py

## Related Files

```
strands-ts/src/mcp/client.ts
strands-ts/src/agent/agent.ts
```

## What needs to be migrated

- listChanged handling: client constructed with `listChanged.tools {autoRefresh, debounceMs: 300, onChanged}` (`client.ts:175-184`)
- Refresh with coalescing of concurrent notifications (`client.ts:359-379`), `onToolsChanged` setter (`client.ts:355-357`)
- Agent auto-rewires its tool registry on change (`agent.ts:777-785`)

Python today: tools are frozen at init. `_loaded_tools` is cached until teardown (`mcp_client.py:401-431`, reset only at `:537`), and all non-error notifications are silently discarded by the message handler fallthrough (`_handle_error_message`, `mcp_client.py:1014-1021`).

## Gotchas / dependencies

- **Coordinate with #1659:** the 2026-07-28 RC replaces the notification channel with `subscriptions/listen` (SEP-2575). Isolate the transport-level subscription from the refresh/debounce logic so only the subscription layer changes when the RC lands.
- Python's client runs a sync facade over a background thread + private asyncio loop; debounce must be scheduled on the client's loop, not the caller's thread.
- Refresh must re-apply `tool_filters` and `prefix` (the same requirement exists in the reverse port #3161).
- The Python agent-side rewire goes through the ToolProvider integration; mirror the TS behavior of swapping registry entries rather than appending.

## Confirmation

- [x] The change is complete in the source SDK, and every related artifact is listed above (source, tests, and docs, if the change has any). Unit-tested (listChanged coalescing in `client.test.ts`); no dedicated docs section exists for this behavior.

March-77:
## Motivation

Python agents currently cache MCP tools at initialization and silently discard
the server's tool-list change notification. Long-lived agents therefore keep
stale tools after a server adds, removes, or changes its catalog.

This handles the current MCP notification in an isolated subscription layer,
coalesces bursts with a 300 ms debounce on the client's background event loop,
refreshes every page, and atomically swaps each consuming registry's provider
tools. Refreshes reuse the configured filters and prefix.

Fixes #3272

## Public API Changes

`ToolProvider` now has a backward-compatible, default no-op
`set_tools_changed_callback()` hook. Dynamic providers can use it to publish a
replacement snapshot without requiring changes from static provider
implementations.

```python
provider.set_tools_changed_callback(
    consumer_id,
    lambda old_names, new_tools: ...,
)
```

## Documentation PR

The MCP guide now explains automatic dynamic-list refresh behavior and clarifies
that manually listed tools are a static snapshot.

## Type of Change

New feature

## Testing

- Targeted MCP provider and registry tests: 80 passed
- Full Python suite: 4,647 passed
- Ruff formatting and lint checks passed for changed files
- mypy passed for changed source files

## Checklist

----

### DIFF
--- a/site/src/content/docs/user-guide/concepts/tools/mcp-tools.mdx
+++ b/site/src/content/docs/user-guide/concepts/tools/mcp-tools.mdx
@@ -100,6 +100,12 @@ Tools can also be listed explicitly if needed:
 </Tab>
 </Tabs>
 
+### Dynamic Tool Lists
+
+When an MCP client is passed directly to an agent, both SDKs listen for the server's tool-list change notifications. Notifications are debounced, the complete tool list is fetched again, and the agent's registry is updated before its next invocation. Configured Python tool filters and name prefixes are reapplied during every refresh.
+
+Tools obtained by calling `list_tools_sync()` or `listTools()` and then passed to an agent are a snapshot. Pass the client itself to the agent when runtime tool-list updates should be applied automatically.
+
 ## Transport Options
 
 Both Python and TypeScript support multiple transport mechanisms for connecting to MCP servers.
--- a/strands-py/src/strands/tools/mcp/mcp_client.py
+++ b/strands-py/src/strands/tools/mcp/mcp_client.py
@@ -43,7 +43,9 @@ from mcp.types import (
     ListResourcesResult,
     ListResourceTemplatesResult,
     ReadResourceResult,
+    ServerNotification,
     TextResourceContents,
+    ToolListChangedNotification,
 )
 from mcp.types import CallToolResult as MCPCallToolResult
 from mcp.types import EmbeddedResource as MCPEmbeddedResource
@@ -136,6 +138,8 @@ _NON_FATAL_ERROR_PATTERNS = [
     # See: https://github.com/modelcontextprotocol/python-sdk/blob/c51936f61f35a15f0b1f8fb6887963e5baee1506/src/mcp/shared/session.py#L421
     "unknown request id",
 ]
+
+_TOOLS_CHANGED_DEBOUNCE_SECONDS = 0.3
 
 
 class MCPClient(ToolProvider):
@@ -274,6 +278,9 @@ class MCPClient(ToolProvider):
         self._tool_provider_started = False
         self.server_instructions: str | None = None
         self._consumers: set[Any] = set()
+        self._tools_changed_callbacks: dict[Any, Callable[[list[str], Sequence[AgentTool]], None]] = {}
+        self._tools_change_generation = 0
+        self._tools_refresh_task: asyncio.Task[None] | None = None
 
         # Task support configuration and caching
         self._tasks_config = tasks_config
@@ -440,6 +447,14 @@ class MCPClient(ToolProvider):
         self._consumers.add(consumer_id)
         logger.debug("added provider consumer, count=%d", len(self._consumers))
 
+    def set_tools_changed_callback(
+        self,
+        consumer_id: Any,
+        callback: Callable[[list[str], Sequence[AgentTool]], None],
+    ) -> None:
+        """Register a consumer callback for MCP tool-list changes."""
+        self._tools_changed_callbacks[consumer_id] = callback
+
     def remove_consumer(self, consumer_id: Any, **kwargs: Any) -> None:
         """Remove a consumer from this tool provider.
 
@@ -450,6 +465,7 @@ class MCPClient(ToolProvider):
         Uses existing synchronous stop() method for safe cleanup.
         """
         self._consumers.discard(consumer_id)
+        self._tools_changed_callbacks.pop(consumer_id, None)
         logger.debug("removed provider consumer, count=%d", len(self._consumers))
 
         # A swallowed continue_on_error failure leaves _tool_provider_started False but still needs
@@ -538,6 +554,9 @@ class MCPClient(ToolProvider):
         self._tool_provider_started = False
         self._connection_failed = False
         self._consumers = set()
+        self._tools_changed_callbacks = {}
+        self._tools_change_generation = 0
+        self._tools_refresh_task = None
         self._server_task_capable = None
         self._tool_task_support_cache = {}
 
@@ -580,6 +599,15 @@ class MCPClient(ToolProvider):
         list_tools_response: ListToolsResult = self._invoke_on_background_thread(_list_tools_async()).result()
         self._log_debug_with_thread("received %d tools from MCP server", len(list_tools_response.tools))
 
+        return self._adapt_tools_response(list_tools_response, effective_prefix, effective_filters)
+
+    def _adapt_tools_response(
+        self,
+        list_tools_response: ListToolsResult,
+        prefix: str | None,
+        tool_filters: ToolFilters | None,
+    ) -> PaginatedList[MCPAgentTool]:
+        """Adapt one MCP tool-list page and apply configured naming and filtering."""
         mcp_tools = []
         for tool in list_tools_response.tools:
             if self._is_tasks_enabled():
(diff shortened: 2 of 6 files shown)


liramon2:

strands-py/src/strands/tools/registry.py
```diff
@@ -158,6 +159,25 @@ async def get_tools() -> Sequence[AgentTool]:
...
+    def _replace_provider_tools(self, old_tool_names: list[str], new_tools: Sequence[AgentTool]) -> None:
+        """Replace a provider's registered tools after its remote catalog changes."""
+        updated_registry = self.registry.copy()
+        updated_dynamic_tools = self.dynamic_tools.copy()
+
+        for tool_name in old_tool_names:
+            updated_registry.pop(tool_name, None)
+            updated_dynamic_tools.pop(tool_name, None)
+
+        for new_tool in new_tools:
```
strandly-the-agent, line 171: 🔴 **Block: preserve tool ownership and duplicate-name invariants during refresh**

This writes refreshed tools directly by name, bypassing `register_tool()`’s exact and normalized-name checks, while the callback carries no provider identity. Cold repro A confirmed that a provider can overwrite a local or another provider’s same-named tool; a later removal can then delete the unrelated tool. It also admitted `foo-bar` + `foo_bar`, which normal registration rejects.

Please bind replacement to the provider, validate the full candidate snapshot against tools owned by other sources, and commit only after validation. Add exact/normalized collision and cross-provider ownership tests.

strands-py/src/strands/tools/registry.py
```diff
@@ -158,6 +159,25 @@ async def get_tools() -> Sequence[AgentTool]:
...
+            updated_registry[new_tool.tool_name] = new_tool
+            if new_tool.is_dynamic:
+                updated_dynamic_tools[new_tool.tool_name] = new_tool
+            else:
+                updated_dynamic_tools.pop(new_tool.tool_name, None)
+
```
strandly-the-agent, line 177: 🔴 **Block: keep the offered catalog and execution lookup cycle-consistent**

The model call snapshots tool specs at `event_loop.py:523-537`, but execution later resolves against this newly assigned live registry at `_executor.py:129-130`. Cold repro D offered `old`, refreshed to `new` while the model streamed, then returned `Unknown tool: old` for the tool the model had validly been offered.

Please carry the implementation snapshot through that model cycle, or stage registry publication until the model response finishes. The next model request can then see the new catalog without invalidating an in-flight call.

strands-py/src/strands/tools/mcp/mcp_client.py
```diff
@@ -590,20 +618,67 @@ async def _list_tools_async() -> ListToolsResult:
...
             # Apply filters if specified
-            if self._should_include_tool_with_filters(mcp_tool, effective_filters):
+            if self._should_include_tool_with_filters(mcp_tool, tool_filters):
                 mcp_tools.append(mcp_tool)
 
         self._log_debug_with_thread("successfully adapted %d MCP tools", len(mcp_tools))
         return PaginatedList[MCPAgentTool](mcp_tools, token=list_tools_response.nextCursor)
 
+    async def _refresh_tools(self) -> None:
+        """Refresh cached provider tools on the MCP client's background event loop."""
```
strandly-the-agent, line 636: 🔴 **Block: commit task metadata transactionally with refreshed tools**

This clears task support before any page is fetched. Cold repro B began with loaded `long_job` and support `required`; after `list_tools()` failed, the old tool remained loaded but its support cache was empty, so `_should_use_task("long_job")` became false. The unchanged tool would then use the wrong MCP protocol.

Please build task metadata in a temporary cache and swap it with `_loaded_tools` only after every page succeeds.

strands-py/src/strands/tools/registry.py
```diff
@@ -158,6 +159,25 @@ async def get_tools() -> Sequence[AgentTool]:
             add_tool(tool)
         return tool_names
 
+    def _replace_provider_tools(self, old_tool_names: list[str], new_tools: Sequence[AgentTool]) -> None:
+        """Replace a provider's registered tools after its remote catalog changes."""
```
strandly-the-agent, line 163: 🔴 **Block: serialize the complete registry replacement transaction**

Separate MCP clients invoke callbacks from separate background loops. Two callbacks can therefore copy the same old registry and overwrite each other. Cold repro C3 let provider B commit `new_b`; provider A’s stale copy then replaced it, restoring `old_b` and losing `new_b`.

Please protect all registry mutations with one serialization mechanism and publish the provider-aware state coherently. Add a concurrent two-provider refresh test.

strands-py/src/strands/tools/mcp/mcp_client.py
```diff
@@ -590,20 +618,67 @@ async def _list_tools_async() -> ListToolsResult:
...
+        session = cast(ClientSession, self._background_thread_session)
+        old_tool_names = [tool.tool_name for tool in self._loaded_tools or []]
+        refreshed_tools: list[MCPAgentTool] = []
+        pagination_token = None
+
+        if self._is_tasks_enabled():
+            self._tool_task_support_cache.clear()
+
+        while True:
+            response = await session.list_tools(cursor=pagination_token)
```
strandly-the-agent, line 646: 🔴 **Block: track the last successfully delivered snapshot per consumer**

`_loaded_tools` advances before callbacks run, and callback failures are swallowed. Cold repro C1 had one consumer fail `old → fresh1`; the next delivery used global `fresh1` as its removal baseline even though that consumer still owned `old`, leaving it with `fresh2 + old` indefinitely.

Please retain each consumer’s successfully delivered baseline and advance it only after its callback succeeds, so later refreshes reconcile failed consumers.

site/src/content/docs/user-guide/concepts/tools/mcp-tools.mdx
```diff
@@ -100,6 +100,12 @@ Tools can also be listed explicitly if needed:
 </Tab>
 </Tabs>
 
+### Dynamic Tool Lists
+
+When an MCP client is passed directly to an agent, both SDKs listen for the server's tool-list change notifications. Notifications are debounced, the complete tool list is fetched again, and the agent's registry is updated before its next invocation. Configured Python tool filters and name prefixes are reapplied during every refresh.
```
strandly-the-agent, line 105: 🟡 **The next invocation does not wait for this refresh.** The implementation debounces and refreshes asynchronously, and a refresh may fail, so an invocation immediately after the notification can still use the old catalog.

```suggestion
When an MCP client is passed directly to an agent, both SDKs listen for the server's tool-list change notifications. Notifications trigger a debounced background refresh of the complete tool list. After a refresh completes successfully, subsequent model requests use the refreshed catalog; invocations do not wait for an in-progress refresh. Configured Python tool filters and name prefixes are reapplied during every refresh.
```

strandly-the-agent (changes requested):
**Changes requested — dynamic MCP refresh can corrupt registry/task state and invalidate valid in-flight tool calls.**

✅ Reviewed `b9e6b04d1361ab61211f845941b6d669462802f1` against merge base `1b671d6a1a3609ae3f46bc20c599e4c9ad0db4cf`; tracked tree clean.
✅ Repros A–D cold-confirmed twice; script SHA256 `ebba011b6b6f6ee0cfeb67351a5d705deb65ca5a5ee280dbe0d20537026f9343`.
✅ Focused registry/MCP tests: **80 passed** (4 RuntimeWarnings); `git diff --check` clean.
⚠️ This changes public API, but `api/needs-review` / `api/review-complete` are absent.

**Blocking question:** Is generic, non-MCP dynamic-provider extensibility intentionally in scope? If not, could this remain an MCP-specific capability? If public, the callback needs the repository’s `Protocol`/`**kwargs` extensibility shape plus defined threading, failure, and unsubscription semantics.

<details><summary>Review-pass breakdown</summary>

- **Correctness/adversarial:** Five cold-reproduced blockers retained: ownership/collision corruption, model-cycle inconsistency, task-cache rollback, concurrent lost update, and per-consumer baseline loss.
- **API/DevX:** The public callback is a moderate API change. Please apply `api/needs-review`; use `api/review-complete` only after API review. The advanced pass was unavailable under the configured retention mode, so it was retried successfully on the default model.
- **Tests:** Existing tests pass but cover disjoint successful snapshots; missing regression cases are folded into the corresponding findings.
- **LLM context:** The model-catalog/executor divergence is included as a reproduced blocker.
- **Docs:** One inaccurate synchronization guarantee is called out inline.
- **Suppressed:** A shared-provider initialization race reproduced on the merge base too, so it was not attributed to this PR. Lower-confidence/pre-existing concerns were dropped.

</details>

yonib05:
Thanks for picking up #3272 - tool-list refresh is genuinely on our roadmap, and parts of this (the isolated subscription layer, the debounce placement) are on the right track.

We're going to close this one though. The automated review already reproduced several blocking concurrency defects in the registry swap design (cross-provider overwrites, mid-invocation registry swaps breaking in-flight tool calls, the task-support cache being cleared before a fetch that can fail, and a lost-update race between clients on separate background loops), and the one failing CI check (codecov/patch) lands exactly on the untested refresh error paths. Getting this right means redesigning the swap protocol around locking and provider identity, and for concurrency-sensitive core code like the tool registry we'd rather drive that design ourselves.

A general note since you're clearly using an agent to build these (which is welcome!): per the [Using AI Tools section of CONTRIBUTING.md](https://github.com/strands-agents/harness-sdk/blob/main/CONTRIBUTING.md#using-ai-tools), the agent is a tool and you're the author - please review every line, simplify what's over-engineered, and be ready to defend the design. Concurrency machinery like this is exactly where that human pass matters most. Smaller, well-scoped PRs against 'ready for contribution' issues are the fastest path to merge.

PR #3458, opened 2026-07-24
feat(ts): add invocation idempotency controls

### LINKED ISSUE #3018: [FEATURE] Add concurrent_invocation_mode and idempotency_token parity with Python SDK
## Summary

The Python SDK supports configurable concurrent invocation behavior via `concurrent_invocation_mode` (a `ConcurrentInvocationMode` enum) and request deduplication via `idempotency_token`. The TypeScript SDK only throws `ConcurrentInvocationError` on overlap with no configurability.

### Idempotency token (higher priority)

When a retry arrives for a request already in flight, the caller can pass an `idempotencyToken` to `invoke()` / `stream()`. If a call with the same token is already running, the new call awaits the original and returns the same result — no duplicate work.

Implementation in JS is simpler than Python (no threading): a `Map<token, Promise<AgentResult>>` on the agent instance, cleaned up on completion.

Python reference: `strands-py/src/strands/agent/_concurrency.py`

### Concurrent invocation mode (lower priority)

A constructor option (`concurrentInvocationMode`) accepting `'throw'` (default, current behavior) or `'unsafeReentrant'` (skips the guard entirely). This is an escape hatch for advanced users who handle synchronization themselves.

Python reference: `strands-py/src/strands/types/agent.py` (`ConcurrentInvocationMode` enum)

## Motivation

The same problem exists in both runtimes — a Node.js server sharing one agent instance across requests will hit overlapping `await agent.invoke()` calls. Idempotency-token deduplication is particularly valuable for HTTP servers handling retried requests.

March-77:
## Summary

- add `idempotencyToken` support to `invoke()` and `stream()` so matching in-flight retries share the original result or error
- add configurable `concurrentInvocationMode` with guarded `throw` and opt-in `unsafeReentrant` behavior
- reject abandoned-primary waiters with `IdempotencyAbortedError` and document the TypeScript concurrency contract

## Testing

- `npm test` (3,874 tests)
- `npm run type-check`
- `npm run lint`
- `npm run format:check`
- `npm run build` in `site` (857 pages, no broken links)

Closes #3018


site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
```diff
@@ -311,7 +311,42 @@ Sometimes more than one request targets the same agent instance at once: a retri
...
+  const [result, retriedResult] = await Promise.all([original, retry])
+  console.log(result === retriedResult) // true
+  ```
+
+  A matching retry waits for the original and receives the same result or error without rerunning the agent. A deduplicated `stream()` call emits only the final `AgentResultEvent`, not the original's intermediate events. If the primary stream is abandoned before it produces a result, waiting retries receive `IdempotencyAbortedError`. The token is forgotten when the invocation finishes, so it can be reused for a later request. A different token, or a call without a token, still receives `ConcurrentInvocationError` while work is in flight.
+
+  Advanced callers that provide their own synchronization can disable the guard:
+
+  ```typescript
+  const agent = new Agent({ concurrentInvocationMode: 'unsafeReentrant' })
```
notowen333: lets workshop this name 'unsafefReentrant' is vague

strands-ts/src/agent/agent.ts
```diff
@@ -146,6 +146,26 @@ export type ToolList = (Tool | McpClient | Agent | ToolList)[]
  */
 export type ToolExecutorStrategy = 'sequential' | 'concurrent'
 
+/**
+ * Controls how an agent handles overlapping invocations.
+ *
+ * - `'throw'` (default) rejects overlapping work with {@link ConcurrentInvocationError}.
+ * - `'unsafeReentrant'` allows overlap without protecting shared conversation state.
+ */
+export type ConcurrentInvocationMode = 'throw' | 'unsafeReentrant'
```
notowen333: I think also "throw" can be renamed to singleThreaded or something like that

strands-ts/src/agent/agent.ts
```diff
@@ -467,7 +495,8 @@ export class Agent implements LocalAgent, InvokableAgent {
   private _toolRegistry: ToolRegistry
   private _mcpClients: McpClient[]
   private _initialized: boolean
-  private _isInvoking: boolean = false
```
notowen333, line 513: Lets keep the boolean separately. Would prefer not to derive from the counter and have an increased blast radius

notowen333 (changes requested):
I like the idea here just mostly comments around naming and the small addition to keep the isInvoking as a bool

March-77:
Addressed the requested concurrency API changes in `8db9d5a7d`:

- renamed the modes to `singleThreaded` and `unsafeConcurrent`
- restored a dedicated `isInvoking` boolean while retaining the active count so overlapping unsafe calls keep the correct lifecycle state
- updated the public docs, option docs, and tests

Validation: 122 targeted agent tests passed; TypeScript build, type-check, formatting, and lint passed.

March-77:
The requested concurrency API changes are implemented in 8db9d5a7, with targeted tests and TypeScript checks passing as summarized above. @notowen333 could you re-review and clear the outstanding changes-requested state?

March-77:
All requested concurrency API changes are present in `8db9d5a7d`, and all three review threads are resolved. There has been no additional feedback since the update. Could you please re-review and clear the stale changes-requested state? The remaining docs, secure integration, and automated-review jobs are awaiting manual approval.

March-77:
Merged current `main` and resolved the `Agent.stream()` conflict in `3018601a`. The resolution preserves the new continuation cleanup from `main` and guarantees the PR concurrency lock is released in a nested `finally`.

Validation on the merged head:
- full TypeScript unit suite with coverage: 4,306 passed
- focused agent tests: 122 passed
- continuation + concurrent agent tests: 27 passed
- TypeScript build and source/integration type-check passed
- full TypeScript ESLint and Prettier checks passed
- `git diff --check` passed

All previously requested changes remain present and all review threads are resolved. @notowen333, please re-review and clear the stale changes-requested state.

March-77:
GitHub rejected the formal reviewer-request mutation because a fork author does not have `RequestReviewsByLogin` permission on the upstream repository. The re-review request is therefore recorded by the @notowen333 mention above, but a maintainer still needs to add them (or another reviewer) to the requested-reviewers field.

nicolas-m-romero:
Thanks for working on this and for addressing the earlier review feedback. I’d like to review the current implementation, but the PR’s branch history is causing GitHub to show hundreds of unrelated changes.

Could you please clean up the branch by rebasing onto current main, or recreating it from current main with only the invocation-idempotency changes? The intended feature appears to touch six files. A focused history and Files Changed view will make it possible to review and validate the implementation accurately.

Please avoid merging main into the feature branch again. Let me know if you need help reconstructing the branch.

PR #3461, opened 2026-07-24
feat(ts): detect repetitive swarm handoffs

### LINKED ISSUE #3168: [PORT] Swarm repetitive-handoff detection (Py→TS)
> ⚠️ Filed by Strandly as part of the #2498 parity audit. Source feature verified merged + unit-tested in `strands-py`. Agent-produced — confirm scope. (Add the `port` label — I can't apply it via API.)

**Parent:** #2498

## Source SDK
`strands-py`

## Target SDK
`strands-ts`

## Related Files
```
strands-py/src/strands/multiagent/swarm.py
strands-py/tests/strands/multiagent/test_swarm.py
site/src/content/docs/user-guide/concepts/multi-agent/swarm.mdx
site/src/content/docs/user-guide/concepts/multi-agent/swarm.ts
```

## What needs to be migrated
Detection that stops a swarm when agents ping-pong. Port the **detection logic only** (NOT the handoff mechanism — Py tool-based vs TS structured-output is a deliberate divergence).

Python (`swarm.py:215-224`): if `repetitive_handoff_detection_window > 0` and `len(node_history) >= window`, take the trailing `window` nodes, and if `len(set(recent)) < repetitive_handoff_min_unique_agents`, stop with reason `"Repetitive handoff: {unique} unique nodes out of {window} recent iterations"`.
- Two config knobs, **both disabled by default at 0**: `repetitive_handoff_detection_window` (`:249`), `repetitive_handoff_min_unique_agents` (`:250`).
- Guard semantics: active only when `window > 0` AND `history >= window`; uniqueness over trailing `window`; strict `<` trigger.
- **Outcome:** soft stop — sets `completion_status = FAILED` (`:751`), returns a FAILED result, does **not** raise.

TS `swarm.ts` has no such detection (only `maxSteps`, which **throws** at `:115,590`).

## TypeScript landing points
- Add both fields to `SwarmConfig` (`strands-ts/src/multiagent/swarm.ts:45-61`), defaults `0` in the ctor `this.config` (`:151-156`).
- History source: `state.results.map(r => r.nodeId)` (`state.ts:51,306`) — completed nodes in order; no new structure needed.
- Add a private `_checkRepetitiveHandoff(state)` invoked at loop top (near `:333`), factored like `_checkSteps` (`:588`).
- Docs: add two rows to the TS config table (`swarm.mdx:125-132`).

## Gotchas / open questions
- `len(set(recent))` → `new Set(nodeIds).size` (strings — no hashing concern).
- **🚩 Termination convention (biggest decision):** Python returns a soft FAILED result (no throw); TS `maxSteps` **throws**. Recommend mirroring Python (set FAILED, break, no throw) since that's the source behavior — but TS swarm has no "stop with FAILED result without throwing" path today. **Confirm with a maintainer.**
- `int`, default `0` — keep "disabled by default"; do NOT default to `Infinity` (that's the maxSteps convention = never disabled).
- camelCase: `repetitiveHandoffDetectionWindow`, `repetitiveHandoffMinUniqueAgents`.

- Merged: ✅  ·  Unit tests: ✅ (`test_swarm.py:222-232` — trigger case; add the non-trigger case on port)  ·  Integ: ❌ not exercised  ·  Docs: ✅ Python (`swarm.mdx:73-74,119-120`); TS table rows missing (add on port).

## Confirmation
- [x] The change is complete in the source SDK (merged + unit-tested); every related artifact is listed above.

March-77:
## Summary
- add disabled-by-default repetitive handoff detection settings to TypeScript swarms
- stop ping-pong execution with a soft FAILED aggregate result, matching Python semantics
- cover triggering, strict uniqueness threshold, defaults, and document the new settings

Fixes #3168

## Testing
- TypeScript unit suite: 140 files, 3869 tests
- TypeScript type checks, ESLint, Prettier, browser bundle check
- docs unit suite: 456 passed, 2 skipped
- docs and snippet type checks
- Astro static build


JackYPCOnline:
@strandly-the-agent can you review this PR? Wondering do we actually need this?

strands-ts/src/multiagent/swarm.ts
```diff
@@ -591,6 +607,24 @@ export class Swarm implements MultiAgent {
...
+  /**
+   * Checks recent node history for a handoff loop.
+   *
+   * @returns A diagnostic reason when the configured uniqueness threshold is not met.
+   */
+  private _checkRepetitiveHandoff(state: MultiAgentState): string | undefined {
+    const window = this.config.repetitiveHandoffDetectionWindow
+    if (window <= 0 || state.results.length < window) return undefined
+
+    const recentNodeIds = state.results.slice(-window).map((nodeResult) => nodeResult.nodeId)
```
strandly-the-agent: 🔴 **The history this reads includes non-COMPLETED and prior-invocation entries — that silently swallows real interrupt responses.**

`state.results` accumulates `INTERRUPTED` / `CANCELLED` / `FAILED` entries too (`swarm.ts:449`, `:467`, `:508`), and on a resumed invocation it arrives pre-populated from the snapshot (`state.ts:365-384` ← `snapshot.ts:73-91` ← `session-manager.ts:313-320`, all before the loop at `:340`). Python's `node_history` appends **only after a clean completion** — `swarm.py:806` sits below the interrupt break at `:800-802` — so this isn't the equivalent source.

Observable with a plausible config (3 nodes, `window: 4`, `minUniqueAgents: 3`): after `a → b → a → b(INTERRUPTED)`, resuming with the interrupt response returns `status: FAILED`, `error: undefined`, **zero nodes executed**, `interrupts` still populated, and the response discarded. Same scenario with detection disabled completes; real `strands-py` completes too. Full repro + output in the summary comment.

This fixes that case (verified: my repro passes, all 39 existing tests stay green, no type errors):

```suggestion
    const window = this.config.repetitiveHandoffDetectionWindow
    if (window <= 0) return undefined

    const completed = state.results.filter((nodeResult) => nodeResult.status === Status.COMPLETED)
    if (completed.length < window) return undefined

    const recentNodeIds = completed.slice(-window).map((nodeResult) => nodeResult.nodeId)
```

It does **not** fix the sibling case, which probably needs a design call: because the window spans invocations, a legitimately completed run (`a,b,a,b`, all COMPLETED) also poisons the *next* invocation on the same session — that one returns `FAILED` with the previous run's content and zero nodes executed, for a brand-new question. Scoping the window to nodes executed *this* invocation would cover both.

strands-ts/src/multiagent/swarm.ts
```diff
@@ -371,9 +384,12 @@ export class Swarm implements MultiAgent {
         node = target
       }
 
-      this._checkSteps(state, handoff)
+      if (!repetitiveHandoffReason) {
+        this._checkSteps(state, handoff)
+      }
 
       result = new MultiAgentResult({
+        ...(repetitiveHandoffReason && { status: Status.FAILED }),
```
strandly-the-agent: 🟡 **Nothing tells the caller why the swarm stopped.** The reason string built at `:622` only reaches `logger.debug` (`:343`); `MultiAgentResult` sets `error` only when it's passed (`state.ts:222`), and this call never passes it. So a caller sees `status: FAILED`, `error: undefined`, every node result `COMPLETED`, and `content` = the last handoff's routing text (`_resolveContent`, `:565-573`) — indistinguishable from an ordinary handoff. Compare the userland `BeforeNodeCallEvent.cancel` path, which preserves a human-readable message (`events.ts:290`).

```suggestion
        ...(repetitiveHandoffReason && { status: Status.FAILED, error: new Error(repetitiveHandoffReason) }),
```

(Python has the same gap — `swarm.py:753` logs the reason and `SwarmResult` doesn't carry it — so this is a faithful port of a gap rather than a TS-only regression. Still worth closing while it's in front of you. Verified this compiles and leaves 39/39 green.)

strands-ts/src/multiagent/swarm.ts
```diff
@@ -45,6 +45,10 @@ import { normalizeError } from '../errors.js'
 export interface SwarmConfig {
   /** Max total agent executions (including start). Defaults to `Infinity` (no limit). */
   maxSteps?: number
+  /** Number of recent agent executions to inspect for repetitive handoffs. Defaults to `0` (disabled). */
+  repetitiveHandoffDetectionWindow?: number
+  /** Minimum unique agents required within the repetitive-handoff detection window. Defaults to `0` (disabled). */
+  repetitiveHandoffMinUniqueAgents?: number
```
strandly-the-agent, line 51: 🟡 **These two are the only config limits with no validation.** `maxSteps`, `timeout` and `nodeTimeout` are all bounds-checked in `_validateConfig` (`:525-534`) and throw at construction; these aren't mentioned there. Measured consequences:

- `repetitiveHandoffDetectionWindow: -1` → silently disabled (`window <= 0` at `:617`), whereas `maxSteps: -1` throws.
- Setting only one of the pair → silent no-op, with nothing in the types, the ctor or the docs table hinting they're jointly required. (`Number(process.env.X)` on an unset var also yields `NaN`, which type-checks as `number` and slips past both comparisons.)
- `repetitiveHandoffMinUniqueAgents > nodes.size` → unsatisfiable: every run long enough to fill the window fails, and short runs never trip it, so it passes cheap tests and only shows up in production.

Worth adding to `_validateConfig`: reject negative/non-integer values, and reject (or at least warn on) exactly-one-of-the-pair and `minUniqueAgents > nodes.size`.

strands-ts/src/multiagent/__tests__/swarm.test.ts
```diff
@@ -258,6 +260,42 @@ describe('Swarm', () => {
...
+        start: 'a',
+        repetitiveHandoffDetectionWindow: 2,
+        repetitiveHandoffMinUniqueAgents: 2,
+      })
+
+      const result = await swarm.invoke('start')
+
+      expect(result.status).toBe(Status.COMPLETED)
+      expect(result.results.map((r) => r.nodeId)).toStrictEqual(['a', 'b', 'c'])
+    })
```
strandly-the-agent, line 404: 🟡 **Two coverage gaps worth closing before merge.**

1. **No resume/session test.** Every sibling limit has one in the `resume with session manager` block (`:522-650`, e.g. `:538-548`) — and here resume is where the behaviour is most surprising: a resumed run can return `FAILED` having executed zero nodes (see the blocking comment on `_checkRepetitiveHandoff`).
2. **The sliding window itself is unverified.** Both new tests trip the check exactly at `results.length === window`, where `slice(-window)` and `slice(0, window)` produce the same array — mutating the implementation to `slice(0, window)` (i.e. "first N ever" instead of "most recent N") leaves all 39 tests passing. A case that visits several distinct nodes first and only then starts looping would pin the real semantics.

Also `:278` — asserting every node result is `COMPLETED` next to `status: FAILED` pins a real but odd shape; a one-line comment saying that's intentional would save the next reader a double-take.

site/src/content/docs/user-guide/concepts/multi-agent/swarm.mdx
```diff
@@ -128,6 +128,8 @@ The following initialization parameters control swarm behavior and safety limits
 | `start` | Agent ID that receives the initial input | First agent in `nodes` |
 | `nodes` | Array of agents (or `AgentNodeOptions`) | (required) |
 | `maxSteps` | Maximum total agent executions (including start) | Infinity |
+| `repetitiveHandoffDetectionWindow` | Number of recent agents to check for ping-pong behavior | 0 (disabled) |
+| `repetitiveHandoffMinUniqueAgents` | Minimum unique agents required in the recent sequence | 0 (disabled) |
```
strandly-the-agent: 🟡 **The rows are accurate-ish but leave out what a user most needs, and one nearby paragraph is now wrong.**

Missing from the table: (a) the two settings are *jointly* required — setting one alone is a silent no-op, though each row reads as independently "0 (disabled)"; (b) the observable outcome on trigger — `status: FAILED` with no `error` and `content` still carrying the last handoff's routing text; (c) the window counts cumulative history, including history restored from a resumed session.

Separately, `:395` in this same file is now false for this case: *"Orchestrator-level limit violations (e.g., exceeding `maxSteps`) throw an exception in TypeScript…"* — repetitive-handoff detection is an orchestrator-level limit by that definition and soft-fails instead. Same sentence appears in the class docstring at `swarm.ts:119`. That's downstream of the convention question in the summary, so it's worth settling that first and then editing whichever side loses.

strands-ts/src/multiagent/swarm.ts
```diff
@@ -371,9 +384,12 @@ export class Swarm implements MultiAgent {
         node = target
       }
 
-      this._checkSteps(state, handoff)
+      if (!repetitiveHandoffReason) {
+        this._checkSteps(state, handoff)
+      }
```
strandly-the-agent: ⚪ Two bundled nits, neither blocking:

- This guard is provably unreachable: `repetitiveHandoffReason` can only be truthy right after the `break` at `:344`, which fires *before* `state.steps++` (`:353`) and only while the loop condition `state.steps < maxSteps` (`:340`) holds — so `_checkSteps` (which throws only at `steps >= maxSteps`) would be a no-op here anyway. Deleting the guard leaves all 39 tests green. Fine to keep as defensive code, but a one-line comment saying so would help, since it currently reads as load-bearing.
- `swarm.mdx:131` says "recent agents"; the Python row two lines up (`:120`) says "recent nodes", which is more accurate — the window slices node executions.

strandly-the-agent (changes requested):
**Verdict: 🔴 request changes** — one verified correctness bug (a resumed swarm can silently swallow a real interrupt response), plus the termination-convention decision that #3168 explicitly asked a maintainer to confirm doesn't appear to have happened. The port itself is faithful and scope-clean — zero scope creep.

**On "do we actually need this?"** — the *capability* looks justified; this *shape* isn't ready yet:

- **Real gap, arguably bigger in TS than Python.** TS `maxSteps` defaults to `Infinity` (`swarm.ts:156`, with a `warnOnce` for exactly that case) where Python has `max_iterations=20` as a cheap backstop — so unbounded ping-pong is *more* expensive here. And `swarm.mdx:369,378` has advertised "repetitive handoff detection" as a swarm safety mechanism with no SDK caveat since `e1f5910dc` (2026-05-22), while TS swarm shipped in `6553ed02f` (2026-03-09). The docs already promise TS users this.
- **But the counter-argument is strong too.** A user can implement this today via `BeforeNodeCallEvent.cancel` (`events.ts:90` → `swarm.ts:460-471`) and get *better* observability than this knob — `CANCELLED` plus a human-readable message, vs. `FAILED` with the reason only in a `logger.debug` line. And the diff makes repetitive-handoff the one orchestrator-level limit that soft-fails, contradicting `swarm.mdx:395` and the class docstring `swarm.ts:119` — which is precisely the decision #3168 flagged as "the biggest decision… **Confirm with a maintainer**".
- **My read:** not "reject the feature" — get that convention decision on the record, fix the resume bug below, and land a fixed version. Two ints of public surface that are individually no-ops is also worth a second look (proposal in the Questions section).

✅ Verified on `969c10ee`: `npx vitest run --project unit-node src/multiagent/__tests__/swarm.test.ts` → **39/39 pass**, no type errors · 🔴 reproduced the interrupt-resume defect end-to-end · ✅ confirmed the suggested fix makes my repro pass while keeping all 39 existing tests green.

<details>
<summary><b>Full breakdown — findings, repro, and open questions</b></summary>

### How this was reviewed

Staged pipeline with fresh-context specialist passes: triage/route → context build → correctness · issue-alignment · API/DevX · adversarial-repro · test-quality · docs-accuracy → dedup/rank/suppress → post. I re-verified the blocking finding myself rather than relaying it. Findings that couldn't be grounded in `file:line` + evidence were dropped (naming preference, `terminus` flag pollution, the missing TS doc snippet, the forward-progress false positive — folded into the design question instead).

### 🔴 The blocker, with a runnable repro

`_checkRepetitiveHandoff` (`swarm.ts:615-625`) reads the whole of `state.results`, which includes `INTERRUPTED` / `CANCELLED` / `FAILED` entries (`swarm.ts:449`, `:467`, `:508`) **and** entries restored from a previous invocation (`state.ts:365-384` ← `snapshot.ts:73-91` ← `session-manager.ts:313-320`, all before the loop at `:340`). Python's `node_history` appends **only after a clean completion** — `strands-py/.../swarm.py:806` sits *below* the interrupt break at `:800-802`. So this isn't the equivalent history source, and the difference is observable:

```
RUN1 status: INTERRUPTED | results: a:COMPLETED,b:COMPLETED,a:COMPLETED,b:INTERRUPTED
RUN2 status: FAILED | error: undefined
RUN2 results: a:COMPLETED,b:COMPLETED,a:COMPLETED,b:INTERRUPTED
RUN2 interrupts still pending: [ 'tool:tool-int:confirm_b' ]
RUN2 content: [{"text":"to b"}]
  ×  AssertionError: expected 'FAILED' to be 'COMPLETED'
CTRL RUN2 status: COMPLETED   <- identical scenario, detection disabled
```

3 nodes, `repetitiveHandoffDetectionWindow: 4`, `repetitiveHandoffMinUniqueAgents: 3` — a plausible config, not a degenerate one. The caller's `InterruptResponseContent` is discarded, **zero nodes execute** in the resumed invocation, the interrupt is never resolved, and the result shape is self-contradictory (`status: FAILED` with `interrupts` still populated, since `state.ts:224` recomputes them). The adversarial pass reproduced the same thing across processes via `SessionManager` (the session stays wedged — `AfterMultiAgentInvocationEvent` re-persists the stale results, so every later resume re-fires) and ran the equivalent scenario against real `strands-py`, where run 2 **completes**. So: a TS port defect, not inherited behavior.

<details>
<summary>The repro (vitest, drop into <code>src/multiagent/__tests__/</code>)</summary>

```ts
function interruptingTool(name: string, interruptName: string, resumeValue = 'ok') {
  return createMockTool(name, (context) => {
    context.interrupt({ name: interruptName, reason: `need ${interruptName}` })
    return resumeValue
  })
}

/** b: hands off to a, then interrupts on its second turn, then finishes. */
function handoffThenInterruptAgent(agentId: string, target: string) {
  const tool = interruptingTool('confirmTool', `confirm_${agentId}`)
    .addTurn({ type: 'toolUseBlock', name: 'strands_structured_output', toolUseId: 'so-1',
               input: { agentId: target, message: `to ${target}` } as JSONValue })
    .addTurn({ type: 'toolUseBlock', name: 'confirmTool', toolUseId: 'tool-int', input: {} })
    .addTurn({ type: 'toolUseBlock', name: 'strands_structured_output', toolUseId: 'so-2',
               input: { message: 'all done' } as JSONValue })
  return new Agent({ model, tools: [tool], printer: false, id: agentId })
}

it('resume after interrupt is swallowed by repetitive-handoff detection', async () => {
  const swarm = new Swarm({
    nodes: [
      createHandoffAgent('a', { agentId: 'b', message: 'to b' }),
      handoffThenInterruptAgent('b', 'a'),
      createFinalAgent('c', 'done by c'),
    ],
    start: 'a',
    repetitiveHandoffDetectionWindow: 4,
    repetitiveHandoffMinUniqueAgents: 3,
  })

  const run1 = await swarm.invoke('start')
  expect(run1.status).toBe(Status.INTERRUPTED)

  const response = new InterruptResponseContent({ interruptId: run1.interrupts![0]!.id, response: 'yes' })
  const run2 = await swarm.invoke([response])
  expect(run2.status).toBe(Status.COMPLETED) // fails today: FAILED, zero nodes ran
})
```
</details>

**One caveat on the inline fix I suggested:** filtering to `COMPLETED` fixes this case (verified — repro passes, 39/39 still green), but it does **not** fix the related cross-invocation case the adversarial pass found: a legitimately completed prior run (`a,b,a,b`, all COMPLETED) poisons the *next* invocation on the same session, which returns `FAILED` with the previous run's content and zero nodes executed for a brand-new question. That one probably needs the window scoped to nodes executed *this* invocation, which is a design call.

### 🟡 Should-fix (all inline)

| # | Where | Point |
|---|---|---|
| 1 | `swarm.ts:392` | Soft-`FAILED` carries no signal: `error: undefined`, all node results `COMPLETED`, `content` reads like an ordinary handoff message. The reason string (`:622`) only reaches `logger.debug` (`:343`). One-line fix suggested inline. |
| 2 | `swarm.ts:49-51` | No `_validateConfig` entry (`:525-534`), unlike every sibling limit. Negative window silently disables (vs. `maxSteps: -1` which throws at construction); setting one of the pair is a silent no-op; `minUniqueAgents > nodes.size` is unsatisfiable and fails every long-enough run with no warning. (`window: NaN` also reaches a `TypeError` in `_resolveContent` — that crash class is **pre-existing**, `maxSteps: NaN` does the same on `main` today, but validation here closes the new door.) |
| 3 | `swarm.test.ts:263-297` | No resume/session test, though every sibling limit has one (`:522-650`) and resume is the most surprising part of this feature. And both new tests trip the check exactly at `results.length === window`, where `slice(-window)` and `slice(0, window)` are the same array — mutating the implementation to `slice(0, window)` leaves all 39 tests green, so the *sliding* semantics are unverified. |
| 4 | `swarm.mdx:131-132` | The rows omit the observable outcome and the joint-field requirement; and the unedited "Error handling" paragraph at `:395` is now factually wrong for this case. |

### ⚪ Nits (bundled, inline on `swarm.ts:387-389`)

The `if (!repetitiveHandoffReason)` guard around `_checkSteps` is provably unreachable, and `swarm.mdx:131` says "recent agents" where the Python row two lines up says the more accurate "recent nodes".

### ❓ Questions

**Blocking**
- **Termination convention.** Should repetitive-handoff detection really be the one orchestrator-level limit that soft-fails instead of throwing? `swarm.mdx:395` and `swarm.ts:119` both currently state the opposite as a *deliberate* TS divergence, and `swarm.mdx:397` says TS keeps `CANCELLED` distinct from `FAILED` specifically so callers can tell "we stopped" from "it failed". #3168 asked for maintainer confirmation on exactly this and is still open with no comments. Whichever way it lands, the code or the docs need to move — and `SwarmConfig` is exported public API (`src/index.ts:343`), so this probably also wants the `api/needs-review` label.

**Non-blocking (proposals, not demands)**
- Would a single cohesive object be safer than two independently-optional ints? Something like `stopOnRepetitiveHandoff?: { window: number; minUniqueNodes: number }` makes "one set, one missing" unrepresentable and gives later options (`scope: 'invocation' | 'session'`, `onDetect: 'throw' | 'fail'`) somewhere additive to live. There's in-repo precedent for the shape (`BedrockGuardrailConfig`, `cacheConfig?: BedrockCacheConfig`). Keeping the flat Python-parity names but validating them as a *pair* would be a reasonable lower-friction alternative.
- Is `MinUniqueAgents` / "Handoff" the right vocabulary? The window slices node *executions* and uniqueness is over node ids — `window: 1, minUniqueAgents: 2` fails after one execution with zero handoffs, reported as "Repetitive handoff". This file has already renamed Python's `max_handoffs`/`max_iterations` → `maxSteps` and documented the delta, so there's precedent for diverging when it reads clearer.
- Given `BeforeNodeCallEvent.cancel` already covers this in userland with better observability, is the real value here discoverability (a config knob users find in the table)? If so, would a documented recipe plus the resume fix get most of the benefit for none of the public surface?

---
🤖 Reviewed by `strandly-the-agent` — an experimental agent, so treat this as work for a human to check rather than a verdict. Happy to be argued out of any of it, and happy to push the fixes if useful.
</details>

JackYPCOnline:
@strandly-the-agent review it again.

strands-ts/src/multiagent/swarm.ts
```diff
@@ -317,6 +325,8 @@ export class Swarm implements MultiAgent {
 
     let caughtError: Error | undefined
     let result: MultiAgentResult | undefined
+    let repetitiveHandoffReason: string | undefined
+    const invocationNodeHistory: string[] = []
```
strandly-the-agent, line 329: 🟡 **Scoping the window to one invocation closes round 1's false positive but opens a false negative: anything that ends an invocation early resets the detection budget.**

`invocationNodeHistory` is a `_stream` local, so detection needs `window` completions inside a single `invoke()`. Measured (window 4 / min 3, three nodes, pure `a↔b`):

```
CONTROL          | rounds=0  | FAILED      | error=Repetitive handoff: 2 unique nodes out of 4 ... | executions=4
ESCAPE-A (HITL)  | rounds=25 | INTERRUPTED | no error | executions=52 | completed=51
ESCAPE-B (1/W-1) | rounds=20 | INTERRUPTED | no error | executions=64 | completed=63
```

The bound: **one interrupt per `window − 1` completions defeats detection permanently.** It doesn't need a human either — `SessionManager` snapshots land after every node by default (`session-manager.ts:119` → `:299-302`), so a crash/redeploy loop resets it the same way (10 crashing restarts, 34 persisted results, all `a`/`b`, never detected). Python catches all of these: it persists and restores `node_history` (`swarm.py:1058`) and has `max_iterations=20` as a backstop where TS defaults to `Infinity`.

The missing distinction looks like **"resume of the same task" (should keep history) vs. "new task on the same session" (should reset)** — your new test pins the second case, which is right; it's the first that's over-reset. One shape that worked in my sandbox, deriving the flag from the resume branch you already have above (`interruptResponsesByNode`, else `_findResumeNode`):

```ts
const invocationNodeHistory: string[] = resumingSameTask
  ? state.results.filter((r) => r.status === Status.COMPLETED).map((r) => r.nodeId)
  : []
```

With that applied, **all 64 tests in `swarm.test.ts` + `interrupts.test.ts` still pass** — including "does not apply restored completed history to a new invocation" (takes the new-task branch) and the session-restored-interrupt test — while ESCAPE-A/B flip to `FAILED / Repetitive handoff` after 2 and 1 rounds. I haven't run lint or the full suite against that patch, so treat it as a sketch of the shape rather than a finished diff.

Not blocking: a false negative in an off-by-default knob is strictly safer than round 1's false positive, and `maxSteps` still bounds the damage for anyone who sets it. But it's worth being a decision — and if per-invocation scoping stays, `swarm.mdx`'s SDK Differences should say so (separate comment there).

strands-ts/src/multiagent/swarm.ts
```diff
@@ -516,6 +536,30 @@ export class Swarm implements MultiAgent {
...
+    if ((window === 0) !== (minimumUniqueAgents === 0)) {
+      throw new Error(
+        `repetitive_handoff_detection_window=<${window}>, repetitive_handoff_min_unique_agents=<${minimumUniqueAgents}> | must both be 0 or both be positive`
+      )
+    }
+    if (minimumUniqueAgents > window) {
+      throw new Error(
+        `repetitive_handoff_min_unique_agents=<${minimumUniqueAgents}>, repetitive_handoff_detection_window=<${window}> | minimum unique agents cannot exceed detection window`
+      )
+    }
```
strandly-the-agent, line 566: 🟡 **The validation block rejects configs that can't work, but still accepts two classes that provably can never fire.**

- `repetitiveHandoffMinUniqueAgents: 1` can never trigger for *any* window: a non-empty slice always has ≥ 1 unique id and the test at `:650` is `uniqueNodes < min`. And since `1 ≤ min ≤ window` is now enforced, `repetitiveHandoffDetectionWindow: 1` is inert for the same reason. Both validate happily today.
- `window: Number.MAX_SAFE_INTEGER` (and any window larger than a run can reach) validates and never fires.
- One level out: when `maxSteps ≤ window`, the `maxSteps` throw always wins before the window can fill — measured, `maxSteps: 3, window: 3, min: 3` throws `max_steps=<3> | swarm reached step limit`, while `maxSteps: 4, window: 3, min: 3` returns the `FAILED` + repetitive error. So detection is silently unreachable for that whole configuration region.

Given the block exists precisely to reject unworkable pairs, requiring `min >= 2` (hence `window >= 2`) would be consistent — something like:

```ts
if (minimumUniqueAgents === 1) {
  throw new Error(
    `repetitive_handoff_min_unique_agents=<${minimumUniqueAgents}> | must be at least 2 to detect a loop`
  )
}
```

Worth weighing against cross-SDK parity, though: Python validates none of this, so every rejection here is already a TS-only divergence — a `warnOnce` may fit better than a throw if you'd rather not reject a config `strands-py` accepts. The `maxSteps ≤ window` interaction is probably a doc line rather than a check.

site/src/content/docs/user-guide/concepts/multi-agent/swarm.mdx
```diff
@@ -390,6 +401,10 @@ The Swarm pattern is available in multiple SDKs. While the core concept is the s
 
 **Node input**: Python builds a rich context string for each receiving agent that includes the original task, full node history chain, accumulated shared context, and available agent descriptions. TypeScript passes only the handoff message and serialized context from the handing-off agent. Agent descriptions are already embedded in the structured output schema for routing decisions.
 
-**Error handling**: In both SDKs, node failures produce a FAILED result. Orchestrator-level limit violations (e.g., exceeding `maxSteps`) throw an exception in TypeScript to promote fail-fast behavior for global failures. Python returns a FAILED result instead.
+**Error handling**: In both SDKs, node failures produce a FAILED result. Exceeding
+`maxSteps` throws an exception in TypeScript to promote fail-fast behavior for that
+global limit, while Python returns a FAILED result. Repetitive handoff detection is a
+soft limit in both SDKs. TypeScript returns a FAILED result with the detection reason
+in `error`.
```
strandly-the-agent, line 412: 🟡 **This paragraph now describes the error contract accurately, but the bigger divergence this PR introduces isn't here: Python's detection survives a resume, TypeScript's doesn't.**

Python restores `node_history` when resuming the same task (`strands-py/.../swarm.py:1058`, gated at `:1027`), so a ping-pong that spans an interrupt-and-resume is still caught. TypeScript's window is per-invocation (`swarm.ts:329`) and resets on every resume. That's also the opposite of `maxSteps`, whose step count TS *does* carry forward (existing test, `swarm.test.ts:714`) — two sibling limits in the same class with opposite scoping is exactly the kind of thing this section exists to record.

```suggestion
**Error handling**: In both SDKs, node failures produce a FAILED result. Exceeding
`maxSteps` throws an exception in TypeScript to promote fail-fast behavior for that
global limit, while Python returns a FAILED result. Repetitive handoff detection is a
soft limit in both SDKs. TypeScript returns a FAILED result with the detection reason
in `error`.

**Repetitive handoff scope**: Python's detection window is restored across an
interrupt-and-resume of the same task, so a ping-pong that spans a resume is still
caught. TypeScript's window is scoped to the current invocation and resets on every
resume, so a ping-pong that alternates with interrupts across several resumes is not
detected — set `maxSteps` if your swarm can interrupt mid-task. This is the opposite of
`maxSteps` itself, whose step count TypeScript does carry forward across resumes.
```

(If you take the patch I sketched on `swarm.ts:329` instead, this paragraph changes to "restored across a resume of the same task in both SDKs" — either way the section should say which it is.)

site/src/content/docs/user-guide/concepts/multi-agent/swarm.mdx
```diff
@@ -128,12 +128,23 @@ The following initialization parameters control swarm behavior and safety limits
 | `start` | Agent ID that receives the initial input | First agent in `nodes` |
 | `nodes` | Array of agents (or `AgentNodeOptions`) | (required) |
 | `maxSteps` | Maximum total agent executions (including start) | Infinity |
+| `repetitiveHandoffDetectionWindow` | Completed-node window | 0 (disabled) |
+| `repetitiveHandoffMinUniqueAgents` | Required unique nodes | 0 (disabled) |
```
strandly-the-agent: ⚪ The reworded descriptions are accurate but purpose-free — a reader on the TypeScript tab no longer sees any hint of *what these are for*, while the Python rows two tabs up (`:120-121`) still say "check for ping-pong behavior". Restoring the intent while keeping your new "completed" precision:

```suggestion
| `repetitiveHandoffDetectionWindow` | Number of recent completed nodes to check for ping-pong behavior | 0 (disabled) |
| `repetitiveHandoffMinUniqueAgents` | Minimum unique nodes required in that window | 0 (disabled) |
```

strandly-the-agent (approved):
**Verdict: ✅ approve** — all six round-1 items are genuinely fixed, and the blocker was fixed more thoroughly than I suggested (invocation-scoped history rather than just filtering to `COMPLETED`), with regression tests that actually bite. One real edge remains; it's a follow-up and a small design call for you, not a blocker.

| Round-1 finding | Status | Evidence |
|---|---|---|
| 🔴 resumed swarm swallowed the interrupt response | **fixed** | my round-1 repro, re-run unchanged: `RUN2 status: COMPLETED` (was `FAILED`, 0 nodes run) |
| 🟡 no reason on the result | **fixed** | `status: FAILED \| error: "Repetitive handoff: 2 unique nodes out of 4 recent iterations"` |
| 🟡 no `_validateConfig` entries | **fixed** | `swarm.ts:540-562`; all 5 predicates fire; no legal config wrongly rejected |
| 🟡 test gaps (resume, sliding window) | **fixed** | reverting `_checkRepetitiveHandoff` to the round-1 signature makes **both** new resume tests fail and the other 62 pass — they're real, not theatre |
| 🟡 docs contradiction | **fixed** | every new sentence checks out against the implementation |
| ⚪ dead `_checkSteps` guard | **fixed** | removed; unconditional call proven safe (the throw condition is unsatisfiable on that path) |

✅ Also verified on `f932939c`: full TS unit suite **141 files / 3906 tests, all pass, no type errors**; `swarm.test.ts` + `interrupts.test.ts` **64/64**.

🟡 **What's left:** the window is now scoped to a single invocation, so anything that ends an invocation early (an interrupt, a crash+resume, a hook cancel) resets the budget and detection can never fire — measured, 52 executions of a pure `a↔b` ping-pong across 25 HITL rounds, never caught. Details + a tested patch inline on `swarm.ts:329`. It's a *false negative* in an opt-in, off-by-default knob, where round 1's bug was a *false positive that broke working runs* — so this revision is strictly safer than both round 1 and the status quo. Hence approve, not block.

<details>
<summary><b>Details — the remaining edge, the measurements, and the smaller items</b></summary>

### How this round was reviewed

Only the delta (`git show 8e710c68`; the merge commit touches no multiagent code). Fresh-context passes: correctness · adversarial/repro · test-quality · docs-accuracy → dedup/rank/suppress. I re-ran the round-1 repro myself rather than relaying it. Round-1 points that are settled (the feature's existence, the alternate config-object shape, `content` carrying the last handoff's routing text) are not re-litigated.

### The escape, measured

`invocationNodeHistory` (`swarm.ts:329`) is a `_stream` local, so detection needs `window` completions *inside one `invoke()`*. Window 4 / min 3, three nodes, pure `a↔b`:

```
CONTROL          | rounds=0  | FAILED      | error=Repetitive handoff: 2 unique nodes out of 4 ... | executions=4
ESCAPE-A (HITL)  | rounds=25 | INTERRUPTED | no error | executions=52 | completed=51
ESCAPE-B (1/W-1) | rounds=20 | INTERRUPTED | no error | executions=64 | completed=63
```

The bound is **one interrupt per `window − 1` completions defeats detection permanently**. And no human is needed: `SessionManager` snapshots land after every node by default (`session-manager.ts:119` → `:299-302`), so a crash/redeploy loop does the same — 10 crashing restarts, 34 persisted results, all `a`/`b`, never detected. Python catches all of these because its `node_history` is persisted and restored (`swarm.py:1058`), and it has `max_iterations=20` as a backstop where TS defaults to `Infinity`.

Compounding: `state.steps` *is* carried forward across invocations (existing test at `swarm.test.ts:714`) while the window resets — so once `maxSteps − state.steps < window`, detection is structurally unreachable for the rest of the session even though it's configured.

### Smaller items (take or leave)

- **Configs that validate but can never fire.** `min: 1` never triggers for any window (a non-empty slice always has ≥1 unique id, and the test is `uniqueNodes < min`), so `window: 1` is inert too; `window: MAX_SAFE_INTEGER` validates; and `maxSteps ≤ window` means the `maxSteps` throw always wins first (measured: `maxSteps:3, window:3` → throws `max_steps=<3>`; `maxSteps:4, window:3` → FAILED with the repetitive error). Inline on `swarm.ts:553`.
- **Untested combination:** nothing in the suite pairs `maxSteps` with the repetitive knobs, so the precedence above is unpinned. Same for the escape scenario — the new interrupt test exercises exactly one interrupt, which can't distinguish per-invocation scoping from any other scoping. Both are one-test fixes if you want the behavior locked in either way.
- **`Status.COMPLETED` guard at `:371-373` is currently an equivalent mutant** — removing it leaves 64/64 green, because every non-COMPLETED status already breaks the loop. Keep it (Python parity, and the patch above makes it load-bearing), but a one-line comment would stop the next reviewer flagging it as untested.
- **Two validation tests assert substrings only** (`swarm.test.ts:191`, `:202`): deleting the value-bearing prefix from the messages at `:555`/`:560` leaves them green. The two `it.each` blocks above them correctly assert the full templated string.
- `swarm.test.ts:350` — the third node `createFinalAgent('c', 'unused')` exists only to satisfy `min ≤ nodes.size`; a one-line comment would save the next reader a double-take.
- **Constructor error precedence flipped**: `new Swarm({ nodes: [], maxSteps: 0 })` now throws `nodes list is empty` where it used to throw `max_steps=<0>`. Necessary side effect of moving `_resolveNodes` above `_validateConfig`; cosmetic, untested either way.

### Deliberately not raised

A pre-existing livelock (`maxSteps` exhausted with a pending interrupt → `INTERRUPTED` forever, no progress, no throw) reproduces byte-identically on `main` @ `1937b4ae`, which has none of this code — unrelated to this PR. Happy to file it separately if useful. The three whitespace-only trims in the unrelated Python streaming example are harmless.

### ❓ Questions

- **(non-blocking, yours to ratify)** Is per-invocation window scoping the intended long-term contract? Landing the patch inline on `:329` keeps all 64 tests green and closes the escape; leaving it means documenting it as a known TS↔Python divergence. Either is defensible — I'd just rather it be a decision than a gap nobody picked.
- **(process)** `SwarmConfig` gained new fields and new construction-time throws here, so this still looks like an `api/needs-review` candidate; the PR carries only `enhancement, area-multiagent, size/m, typescript`.

---
🤖 Re-reviewed by `strandly-the-agent` at `f932939c`. Still an experimental agent — treat this as work for a human to check.
</details>

JackYPCOnline:
@strandly-the-agent we should focus on the port itself instead of improvement. Does this PR achieve faithful port from python.

strandly-the-agent:
**Yes on the algorithm, no on the history lifetime.** I audited the two implementations side by side at `f932939c` and ran *both* SDKs on matched scenarios rather than reading only.

- **Faithful:** field names/types/defaults, the activation guard, the trailing-window slice, uniqueness-over-node-ids, the strict `<` trigger, loop-top placement, the soft stop without raising, which executions count toward the window, and the reason string — byte-identical, 61 chars on both sides, diffed mechanically rather than by eye. Plain ping-pong inside one invocation gives the identical outcome in both SDKs (FAILED after 4 executions, `[a,b,a,b]`).
- **Not faithful — one thing:** Python's `node_history` lives for the whole **task** (preserved on interrupt/session resume at `swarm.py:399-401`, restored from the session payload at `:1028-1030`/`:1058`, reset only for a new task at `:406-411`). TS's `invocationNodeHistory` is a `_stream()` local (`swarm.ts:329`), so it's empty on every resume. Measured: after one resume Python needs **1** more completion to stop, TS needs **4**; with an interrupt every 3 completions Python stops at round 2 while TS **never** stops (7 rounds, 21 completions, `error` null throughout) — and Python still has `max_iterations=20` as a backstop where TS's `maxSteps` defaults to `Infinity`.
- **Two behaviours exist beyond the Python source because I asked for them in round 1** — `error` on the result and the construction-time validation. Python has neither (its `SwarmResult` has no `error` field at all; it validates nothing). Worth saying plainly, since I take it that's what prompted your question.

**To make it an unqualified yes — port fidelity only, no improvements:**
1. Either scope the history to the task the way Python does (persist it across resumes of the same task, reset on a new task), **or** state the divergence in `swarm.mdx` "SDK Differences": today the docs describe TS's own scoping (`:143-146`) but never that Python is task-scoped and TS is invocation-scoped.
2. Decide on the two additions above. `error` is additive metadata that changes no control flow and is already documented (`swarm.mdx:404-408`); the validation changes what's legal at construction, so a strict port would either drop it or document the asymmetry — Python accepts every config TS now rejects, bar one (a non-integer window, where Python instead crashes at runtime with `TypeError` at `swarm.py:216`).

<details>
<summary><b>Full parity ledger + the evidence behind each row</b></summary>

| # | Axis | Verdict | Python | TypeScript |
|---|------|---------|--------|------------|
| 1 | Names / types / defaults (`int = 0` → `?: number`, 0) | faithful | `swarm.py:249-250`, `:283-284` | `swarm.ts:48-51`, `:158-159` (matches root `AGENTS.md:50` re-casing rule) |
| 2 | Activation guard `window > 0 and len(hist) >= window` | faithful | `swarm.py:215` | `swarm.ts:645` — exact De Morgan negation as an early return |
| 3 | Trailing window + uniqueness over node ids | faithful | `swarm.py:216-217` (`set()` via `SwarmNode.__hash__/__eq__`, `:84-92`) | `swarm.ts:648-649` (`new Set(...).size` over id strings) |
| 4 | Strict `<` trigger, no off-by-one | faithful | `swarm.py:218` | `swarm.ts:650` |
| 5 | Reason string | faithful — byte-identical, len 61 both | `swarm.py:222-223` | `swarm.ts:650` |
| 6 | Loop-top placement, soft stop (no raise) | faithful | `swarm.py:745-755` | `swarm.ts:344-348`, status applied at `:396` |
| 7 | Which executions count (only cleanly-completed) | faithful | append at `swarm.py:806`, below the interrupt break `:800-802` and after the cancel break `:786-788` | explicit `Status.COMPLETED` gate, `swarm.ts:371-373` |
| 8 | New task doesn't inherit old history | faithful | `swarm.py:399-411` else-branch resets `SwarmState` | starts empty — same net effect |
| 9 | **History lifetime** | **unintended divergence** | task-scoped: `swarm.py:399-401`, `:1028-1030`, `:1058`, serialized `:994` | invocation-scoped: `swarm.ts:329`, not on `MultiAgentState`, absent from snapshots |
| 10 | `error` on the result | intentional addition (my round-1 ask) | no `error` field on `SwarmResult`; reason only reaches `logger.debug` (`swarm.py:754`) | `error: new Error(reason)` (`swarm.ts:396`), documented `swarm.mdx:404-408` |
| 11 | Construction-time validation | intentional addition (my round-1 ask), undocumented as a divergence | none (no `raise` anywhere near it) | 5 conditions throw, `swarm.ts:540-562` |

**Behavioural runs** — 3 nodes, window 4, min 3, real interrupt round-trips on both sides (Python installed editable from `strands-py`; TS via `vitest --project unit-node`):

```
S1  plain ping-pong, no resume
    python  FAILED after 4 executions [a,b,a,b]
    ts      FAILED after 4 executions [a,b,a,b]     -> same
    reason  'Repetitive handoff: 2 unique nodes out of 4 recent iterations'  (byte-identical, len 61)

S2  ping-pong continues after ONE resume of the same task
    python  inv1 INTERRUPTED node_history=[a,b,a] | inv2 FAILED node_history=[a,b,a,b]   -> 1 more completion
    ts      inv1 INTERRUPTED a:C,b:C,a:C,b:INT   | inv2 FAILED after [b,a,b,a]           -> 4 more completions

S2b interrupt every 3 completions
    python  round 2: FAILED                       (stops)
    ts      rounds 1-7 all INTERRUPTED, completions 3 -> 21, error null throughout  (never stops)

S3  brand-new task after a completed prior run   -> both reset, neither false-triggers
S4  interrupted / cancelled / agent-failed executions -> excluded from the window in both
S5  7 configs Python accepts -> 6 run in Python (two as silent no-ops), all 7 throw at construction in TS;
    the 7th (window=2.5) crashes Python at runtime: TypeError at swarm.py:216
```

Also checked: nothing of the source feature is left unported (the whole repetitive branch of `should_continue` is replicated), and the TS diff contains nothing without a Python counterpart beyond rows 10-11. `should_continue`'s sibling checks (`max_handoffs`/`max_iterations`/`execution_timeout`) are pre-existing and already covered by TS `maxSteps`/`timeout` (documented at `swarm.mdx:400`, `swarm.ts:117`) — not this PR's job. TS baseline on the branch: `swarm.test.ts` + `interrupts.test.ts` 64/64.

**Deliberately left out of this pass:** everything already fixed in rounds 1-2, all improvement suggestions, and two nuances with no observable consequence — Python evaluates the repetitive check last of four inside `should_continue` while TS evaluates it first in the loop body (only visible in a timing race), and the two SDKs exclude failed executions by different mechanisms (Python's append is unreachable after an exception; TS gates on status) with the same resulting set.

</details>

---
🤖 `strandly-the-agent` — Python run from `strands-py` at this same commit, so the comparison is same-tree, not against a release.

JackYPCOnline (approved):

strands-ts/src/multiagent/swarm.ts
```diff
@@ -150,16 +155,19 @@ export class Swarm implements MultiAgent {
 
     this.config = {
       maxSteps: config.maxSteps ?? Infinity,
+      repetitiveHandoffDetectionWindow: config.repetitiveHandoffDetectionWindow ?? 0,
```
lizradway, line 158: can we come up with simpler api contract names?

lizradway (changes requested):
requested changes for api bar raising of repetitiveHandoffDetectionWindow & repetitiveHandoffMinUniqueAgents

strands-ts/src/multiagent/swarm.ts
JackYPCOnline: https://github.com/strands-agents/harness-sdk/blob/06bef2d702d352b91180a3a00e875c2ef3f6f6a2/strands-py/src/strands/multiagent/swarm.py#L194

It is a porting work, the author follows what we have in python

JackYPCOnline:
> requested changes for api bar raising of repetitiveHandoffDetectionWindow & repetitiveHandoffMinUniqueAgents

Since this is a porting work, it should not requires API bar raising.

lizradway (approved):

PR #3466, opened 2026-07-24
feat(middleware): add AgentStreamStage with middleware-initiated interrupts

### LINKED ISSUE #3150: [PORT] Middleware pipeline: ExecuteToolStage + AgentStreamStage + interrupt path (TS→Py)
> ⚠️ Filed by Strandly as part of the #2498 parity audit. The source feature is verified merged + unit-tested in `strands-ts`. Agent-produced — confirm scope before implementing. (Note: repo has no `port` label yet — I lack permission to create it; a maintainer should add it.)

**Parent:** #2498

## Source SDK
`strands-ts`

## Target SDK
`strands-py`

## Related Files
```
strands-ts/src/middleware/stages.ts
strands-ts/src/middleware/types.ts
strands-ts/src/middleware/registry.ts
strands-ts/src/middleware/index.ts
strands-ts/src/middleware/README.md
strands-ts/src/types/interrupt.ts
strands-ts/src/interrupt.ts
strands-ts/src/agent/agent.ts
strands-ts/src/middleware/__tests__/middleware-interrupts.test.ts
strands-ts/src/middleware/__tests__/agent-middleware.test.ts
strands-ts/src/middleware/__tests__/copy-on-input.test.ts
strands-ts/src/middleware/__tests__/custom-stages.test.ts
strands-ts/src/middleware/__tests__/registry.test.ts
```

## What needs to be migrated
Python already has the middleware **core** (`MiddlewareStage`, phases, `MiddlewareRegistry`, `MiddlewareResult`, `InvokeModelStage`) in `strands-py/src/strands/_middleware/`. Its README states the gap: *"Only `InvokeModelStage` is implemented. `ExecuteToolStage` and `AgentStreamStage` will be added as needed."* This ports the two remaining stages **and** the middleware-initiated interrupt path.

- **`ExecuteToolStage`** (`stages.ts:161`) — wraps a single tool execution. Context `ExecuteToolContext` (`stages.ts:104`): `{agent, tool: Tool|undefined, toolUse, invocationState}` + extends `MiddlewareInterruptible`. Result `{result: ToolResultBlock}`. Supports short-circuit (return a mock result without calling `next` → real tool doesn't run); hooks (`BeforeToolCallEvent`/`AfterToolCallEvent`) fire **outside** middleware even on short-circuit.
- **`AgentStreamStage`** (`stages.ts:170`) — wraps the whole agent output stream (outermost). Context `{agent, args, options}` + `MiddlewareInterruptible`. Marked `@internal` in TS (not in public `index.ts`) — keep it internal in Python too.
- **Middleware interrupt path** — `MiddlewareInterruptResult<T> = {response: T}` wrapper; `MiddlewareInterruptible.interrupt()`; `createMiddlewareInterrupt(interruptState, idPrefix)` factory (`agent.ts:368`), a **non-mutating** interrupt fn: id = `` `${idPrefix}:${name}` ``; returns stored response on resume; honors preemptive `response`; else throws `InterruptError` with `source:'middleware'`.

## Python landing points
- Stages → `strands-py/src/strands/_middleware/stages.py` (today only `InvokeModelStage`); export public ones from `_middleware/__init__.py`, keep `AgentStreamStage` internal.
- `ExecuteToolStage` wiring → wrap the **core execution** segment of `ToolExecutor._stream` (`tools/executors/_executor.py:225-249`, the `selected_tool.stream(...)` region) so `Before/AfterToolCallEvent` hooks stay outside middleware.
- `AgentStreamStage` wiring → wrap the per-invocation event-loop-cycle stream in `agent/agent.py` (`_run_loop` ~`1227`), hooks outside. `self._middleware_registry` already exists (`agent.py:407`).
- Interrupt path → reuse `strands-py/src/strands/interrupt.py` (`Interrupt`, `InterruptException`, `_InterruptState`) + `types/interrupt.py` `_Interruptible`; add a `create_middleware_interrupt` factory returning the **wrapper** result (not the raw response).

## Gotchas / open questions (need maintainer input)
- **Generator return values.** Python async generators can't `return` a value; Python already uses the *"last yielded event is the result"* convention for `InvokeModelStage` (`event_loop.py:548`, `registry.py:83`). The two new stages must adopt it — pick the concrete terminal event types.
- **`Interrupt.source` field is missing in Python** (`interrupt.py:22`). Porting `source:'middleware'` means adding a `source` field + serialization — a cross-cutting change.
- **Python interrupts are tool-only today** (`event_loop.py:282`, resume keyed on tool_use/tool_results). An `AgentStreamStage` interrupt doesn't fit that resume shape — expect to extend `_InterruptState.resume()`. **Deepest risk.**
- **No public `addMiddleware` in Python** (TS has `Agent.addMiddleware`). The new stages are unusable without a registration entry point — decide whether this port also adds one.

- Merged: ✅ (`ExecuteToolStage` `agent.ts:2517`, `AgentStreamStage` `agent.ts:1155`, `createMiddlewareInterrupt` `agent.ts:368`).
- Unit tests: ✅ (`middleware-interrupts.test.ts`, `agent-middleware.test.ts`, `copy-on-input.test.ts`, `custom-stages.test.ts`, `registry.test.ts`).
- Integ tests: ❌ none for middleware stages.
- Docs: ⚠️ no site/user-guide page; only `strands-ts/src/middleware/README.md` (design + behavioral spec). No user-facing docs to port.

## Confirmation
- [x] The change is complete in the source SDK (merged + unit-tested); every related artifact is listed above. Test/doc caveats noted above.

zastrowm:
## Description

Ports the third and final middleware stage from the TypeScript SDK, `AgentStreamStage`, closing the AgentStreamStage half of #3150 (`InvokeModelStage` and `ExecuteToolStage` already landed). It wraps an entire agent invocation pass at the outermost interception point, so middleware can filter, transform, or inject events across the whole stream, short-circuit the pass, or gate it behind a human-in-the-loop interrupt.

The stage is kept **internal** - like the other Python staged and matching TS's `@internal` marking.  A follow-up PR will open up InvokeModelStage and ExecuteToolStage with documentation

Two design points a reviewer can't infer from the diff:

- **Interrupt resume (the deepest risk in #3150).** Python interrupts were previously tool-only: the event loop keyed resume behavior on interrupt state being `activated` alone — replaying a stored `tool_use_message` and merging stored `tool_results`. An `AgentStreamStage` interrupt activates the state with an *empty* context, so both reads are now gated on the presence of their tool-context key. An agent-stream interrupt therefore falls through to a normal model call on resume, and a resumed pass that then calls a tool no longer `KeyError`s. The run loop deactivates interrupt state on non-interrupt completion so it can't leak into the next invocation.

Python-only. `AgentStreamStage` interrupts surface as a terminal `EventLoopStopEvent("interrupt")` rather than TS's distinct `InterruptEvent` (Python has no top-level per-interrupt event; tool interrupts surface the same way). This and the other intentional divergences are documented in `strands-py/src/strands/_middleware/README.md`.

## Related Issues

Resolves #3150

## Documentation PR

No user-facing docs — the stage is internal and has no public registration API yet.

## Type of Change

New feature

## Testing

Exercised the stage end to end beyond the gate: wrap/filter/short-circuit, synthetic event injection after the stop event, suppress-all-but-result, multi-turn buffering ("stream final turn only"), and the full interrupt lifecycle — halt, resume-to-completion, resume-into-a-tool-call, and interrupt-state cleanup across invocations.

## Checklist

----


strands-py/src/strands/agent/agent.py
```diff
@@ -1288,28 +1301,58 @@ async def _run_loop(
...
+                        if isinstance(event, EventLoopStopEvent):
+                            agent_result = AgentResult(*event["stop"])
+                        yield event
+
+                    # A resumed AgentStreamStage interrupt that finished without tool execution
+                    # (e.g. a plain end_turn) never hits the tool path's deactivate(), so clear
+                    # the interrupt state here on any non-interrupt completion. A tool interrupt
+                    # raised during this pass leaves stop_reason == "interrupt" and is left alone.
+                    if self._interrupt_state.activated and (
+                        agent_result is None or agent_result.stop_reason != "interrupt"
```
opieter-aws: Looks like we need an additional condition here to avoid wiping pending tool interrupts:
```and "tool_use_message" not in self._interrupt_state.context```

**Comment by zastrowm** [comment]
@strandly-the-agent review this

**Comment by opieter-aws** [comment]
@strandly-the-agent review this

File: strands-py/src/strands/_middleware/stages.py [hunk 2]
```diff
@@ -142,3 +170,79 @@ def _interrupt_id(self, name: str) -> str:
...
+        """
+        return _resolve_middleware_interrupt(self._interrupt_state, self._interrupt_id(name), name, reason, response)
+
+    def _interrupt_id(self, name: str) -> str:
+        """Derive the interrupt id for ``name``, namespaced to the agent-stream stage.
+
+        Follows the SDK's ``v1:`` interrupt-id scheme (see ``types/interrupt.py``), hashing
+        the user-provided name so ids stay stable across resumes.
+        """
+        return f"v1:middleware_agent_stream:{uuid.uuid5(uuid.NAMESPACE_OID, name)}"
```
strandly-the-agent, line 241: 🔴 **The agent-stream interrupt id has no per-agent scoping — one reusable gate across two agents silently shares a single human approval.**

This id is a pure hash of `name`. Every other `v1:` scheme in the codebase embeds a call-scoping token: `v1:tool_call:<toolUseId>` (`types/tools.py:162`), `v1:before_tool_call:<toolUseId>` (`hooks/events.py:173`), `v1:before_node_call:<node_id>:<call_id>` (`hooks/events.py:367`), and your own sibling `v1:middleware_execute_tool:<toolUseId>:<uuid5(name)>` (`stages.py:156-166`). This is the only one without one.

That matters because `Graph`/`Swarm` aggregate every node's interrupts into one flat dict keyed on the bare id (`multiagent/graph.py:739`, `multiagent/swarm.py:730` — `interrupts.update({interrupt.id: interrupt ...})`) and route responses back by matching that id (`graph.py:1179-1183`). So two nodes running *the same reusable gate middleware* — the obvious way to ship an approval-gate plugin — get identical ids, and one answer satisfies both gates.

<details><summary>Repro: 2-node <code>Graph</code>, one shared gate — alpha silently receives beta's decision</summary>

```
graph status: Status.INTERRUPTED
interrupts: [('approval_gate','2938fe','approve alpha?'),
             ('approval_gate','2938fe','approve beta?')]   <- two interrupts, ONE id
resume1: Status.COMPLETED
observed responses per node: {'alpha': ['yes-1'], 'beta': ['yes-1']}
```
Two different questions, one id; `resume()` writes both to the same dict entry (last wins); both nodes read `yes-1` and the graph completes with **no error**. Make one `approve` and the other `deny` and it's a silent approval bypass. Via agent-as-tool the same collision instead never converges (6 resume round-trips, still `interrupt`; distinct names converge in 2).
</details>

This is **not** the same-pass, same-agent collision the earlier round resolved by documenting it at `stages.py:213-216` — that one shares a response between two middleware in one agent. This is one middleware across agents, and the failure is silent cross-wiring rather than a documented sharing rule.

On the fix: `agent.agent_id` alone isn't enough — it defaults to the literal `"default"` for any unconfigured `Agent()` (`agent.py:125,305`), which is exactly how `Graph`/`Swarm` nodes are usually built, so two default nodes would still collide. Either key on something guaranteed unique per agent instance, or make the orchestrators key by `(node_id, interrupt_id)` / reject a colliding id loudly instead of `update()`-ing over it — silently overwriting a pending human decision is the part that should never happen.

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -737,7 +739,10 @@ async def _handle_tool_execution(
     validate_and_prepare_tools(message, tool_uses, tool_results, invalid_tool_use_ids)
     tool_uses = [tool_use for tool_use in tool_uses if tool_use.get("toolUseId") not in invalid_tool_use_ids]
 
-    if agent._interrupt_state.activated:
+    # Only a tool-interrupt resume stores prior tool_results to merge. AgentStreamStage
+    # interrupts activate the state without tool context, so gate on the stored results
+    # (a resumed agent-stream interrupt may reach here on its way into a fresh tool call).
+    if agent._interrupt_state.activated and "tool_results" in agent._interrupt_state.context:
```
strandly-the-agent, line 780: 🔴 **You narrowed the resume *reads* here and at `:285`, but the *write* at `event_loop.py:830` is still an unconditional full reset — it wipes an already-answered agent-stream interrupt mid-pass.**

`_InterruptState.deactivate()` clears **both** `interrupts` and `context` (`interrupt.py:62-69`), and `event_loop.py:830` calls it after any clean tool cycle. That call site is untouched by this PR — but `_interrupt_state` now has a second consumer whose lifetime spans the whole pass (`AgentStreamContext.interrupt()`), so the reset is no longer safe.

Consequence: a middleware that gates before `next_fn` and re-reads its approval after `next_fn` drains — exactly the contract documented at `stages.py:203-205` ("After the user resumes with a response, the second call returns that response") — works if the pass had no tool call, and **livelocks** if it did.

<details><summary>Repro: same middleware, only difference is whether a tool ran</summary>

```
--- pass runs a tool: False ---
  pass1: interrupt   pass2: end_turn   log=[('pre','YES'),('post','YES')]   <- contract honoured
--- pass runs a tool: True ---
  pass1..pass4: interrupt  ...  *** never converges ***
```
Driving it to 5 resumes with a `[toolUse, text]` script:
```
resume-round 0: stop_reason=interrupt model_calls=0 tool_runs=0
resume-round 1: stop_reason=interrupt model_calls=2 tool_runs=1
...
resume-round 4: stop_reason=interrupt model_calls=8 tool_runs=4
*** livelock: after 5 resumes still interrupt, model_calls=10, emails sent=5 ***
interrupt state observed right after next_fn drained: activated=False, interrupts=[]
```
The side effect (an email per round) fires every time, and there's no caller-side workaround once triggered.
</details>

Not a regression — but newly reachable. The symmetric fix is to narrow the write the same way you narrowed these reads: at `:830`, pop only tool-owned state (`tool_use_message`/`tool_results` + tool-namespaced ids) instead of resetting everything. Alternatively, have `_run_loop` snapshot `dict(self._interrupt_state.interrupts)` before `_middleware_registry.invoke(...)` and hand *that* to `AgentStreamContext` — `interrupt()` is already read-only (`stages.py:64-98`), so the event loop then can't invalidate it. Either way, a `[toolUse, text]` pass with a post-`next_fn` `interrupt()` would make a good test.

strands-py/src/strands/agent/agent.py
```diff
@@ -1288,28 +1301,62 @@ async def _run_loop(
...
-                        )
-                        if self._session_manager:
-                            self._session_manager.redact_latest_message(self.messages[-1], self)
-                    yield event
-
-                # Capture the result from the final event if available
-                if isinstance(event, EventLoopStopEvent):
-                    agent_result = AgentResult(*event["stop"])
+                        self._interrupt_state.deactivate()
+                except InterruptException as interrupt_exception:
```
strandly-the-agent, line 1345: 🟡 **An interrupt raised after `next_fn` has drained replays the whole pass: the model is called a second time and history gets two consecutive assistant turns.**

Nothing records that this pass already produced its assistant turn. On resume, `_convert_prompt_to_messages` returns `[]` while activated (`agent.py:1497-1498`), `event_loop_cycle` finds no `tool_use_message` (`event_loop.py:285`) and no trailing tool-use to rescue it (`event_loop.py:289`) — so it calls the model again.

This isn't exotic: the Output-phase adapter drains the entire inner chain before the handler runs (`registry.py:89-113`), so **any** Output-phase interrupt hits it deterministically — as does the natural "post-hoc approval gate" shape (inspect the finished stream, then ask a human before releasing it).

<details><summary>Repro: post-hoc gate on a plain <code>end_turn</code> pass</summary>

```python
async def post_hoc_gate(context, next_fn):
    async for event in next_fn(context):
        yield event
    context.interrupt("approve_output", reason="Approve the assistant reply?")
```
```
pass1: interrupt | model calls: 1 | history: [('user','hello'), ('assistant','SECRET-ANSWER-1')]
pass2: end_turn  | model calls: 2 | history: [..., ('assistant','SECRET-ANSWER-1'), ('assistant','SECRET-ANSWER-2')]
consecutive assistant turns: True
```
That history is then sent verbatim on a later invocation — `call 3: ['user','assistant','assistant','user']`. (I didn't hit a live endpoint, so treat the provider-side `ValidationException` as likely-but-unconfirmed; the duplicate model call and the double assistant turn are confirmed.) A tool-use turn is fine — `_has_tool_use_in_latest_message` rescues it (1 model call, tool ran once).
</details>

Two ways out, depending on your answer to question 1 in the summary: store a "pass already completed" marker in `_interrupt_state.context` next to the finished assistant message so a resumed pass short-circuits (`tool_use_message` is the precedent), or — if post-drain gating is out of scope — detect it right here (`agent_result is not None` ⇒ a stop event was already seen this pass) and raise an actionable error, documenting the Output phase as non-interruptible. Silently re-calling the model is the one option that seems clearly wrong.

strands-py/src/strands/agent/agent.py
```diff
@@ -1328,6 +1375,50 @@ async def _run_loop(
...
+            invocation_state: Per-invocation state passed to the event loop.
+            structured_output_context: Structured output context for this pass.
+            limits: Optional per-invocation budget caps.
+
+        Returns:
+            An async generator function yielding the pass's events, ending with an
+            ``EventLoopStopEvent``.
+        """
+
+        async def terminal(_ctx: "AgentStreamContext") -> AsyncGenerator[TypedEvent, None]:
```
strandly-the-agent: 🟡 **The terminal never reads `_ctx`, so `dataclasses.replace()`-based context transforms on this stage are silently dropped** (found independently by two passes).

`terminal` closes over the outer `invocation_state`/`structured_output_context`/`limits` and ignores its argument. An Input- or wrap-phase handler returning `replace(context, invocation_state=…)` has its replacement discarded with no error — verified live: the marker never reaches `result.state`.

That contradicts this package's own documented pattern (`_middleware/README.md:220-224`, "Middleware creates modified contexts via `dataclasses.replace()`"), and both sibling stages honour their context in the terminal (`event_loop.py:665` reads `ctx.messages`; `_executor.py:437` reads `ctx.tool_use`), as does TS (`strands-ts/src/agent/agent.ts:1185`: `this._streamCore(ctx.args, ctx.options)`).

<details><summary>Repro — the worst API shape: mutate-in-place works, replace silently doesn't</summary>

```
A: AgentStreamStage.Input replaces context.messages  -> model saw ['ORIGINAL']  (expected ['HIJACKED'])
B: same trick on InvokeModelStage.Input (sibling)     -> model saw ['HIJACKED']
C: AgentStreamStage wrap mutates messages in place    -> model saw ['MUTATED-IN-PLACE']
D: Input replaces invocation_state -> middleware sees {'marker':'FROM-MIDDLEWARE'},
                                      event loop/hooks see {'marker':'FROM-CALLER'}
```
D is a split brain: the chain and the event loop/hooks disagree about `invocation_state` for the same pass.
</details>

Neither mypy nor ruff catches it (the `_` prefix suppresses the unused-arg lint) and no test does either (see the note on `test_input_transforms_context`). Nothing in-repo is broken today — both existing consumers register on `InvokeModelStage` (`memory_manager.py:639`, `context_injector/plugin.py:96`) — but the first real user of this stage's Input phase gets a silent no-op.

Suggested: `self._execute_event_loop_cycle(_ctx.invocation_state, …)`, and make a call on `_ctx.messages` — honour it, or drop the field / document it as a read-only view given Python appends to history before the chain runs.

strands-py/src/strands/agent/agent.py
```diff
@@ -1288,28 +1301,62 @@ async def _run_loop(
...
+                        self.messages[-1]
+                        if self.messages
+                        else {"role": "assistant", "content": [{"text": "Interrupted"}]}
+                    )
+                    stop_event = EventLoopStopEvent(
+                        "interrupt",
+                        interrupt_message,
+                        self.event_loop_metrics,
+                        invocation_state.get("request_state", {}),
+                        [interrupt_exception.interrupt],
```
strandly-the-agent: 🟡 **Reporting only the just-raised interrupt hides other still-unanswered ones from the caller** (flagged by three passes independently).

TS returns `getUnansweredInterrupts()` — everything still outstanding (`strands-ts/src/agent/agent.ts:1207,1787`). This passes just the one that raised.

It's internally consistent with the pre-existing tool path (`event_loop.py:752,798,810` also collect only interrupts raised in the current cycle), which is why this is 🟡 not 🔴 — but this PR creates a mix that couldn't happen before: a tool interrupt left unanswered across a resume (legal — `resume()` only updates ids present in the list) while an agent-stream middleware raises a fresh one.

<details><summary>Repro</summary>

```
pass1: interrupt  reported ['gate_a','gate_b']
pass2: interrupt  reported ['stream_gate']   unanswered in state: ['gate_b','stream_gate']
                  ^ gate_b is still unanswered but NOT reported to the caller
pass3: interrupt  reported ['gate_b']
pass4: end_turn
```
A caller that treats `result.interrupts` as "everything I owe an answer to" never learns about `gate_b`; there's no public way to discover it (`_interrupt_state` is private). It does resurface a round-trip later, so a looping caller recovers.
</details>

Either report all unanswered interrupts from `_interrupt_state` (TS parity), or keep the one-at-a-time contract and document it on `AgentResult.interrupts` — right now it's an undocumented divergence with no test.

strands-py/src/strands/agent/agent.py
```diff
@@ -1288,28 +1301,62 @@ async def _run_loop(
...
+                    # fires for agent-stream interrupts (which never store one) and never wipes a
+                    # pending tool interrupt — the event loop owns tool-interrupt state, and some
+                    # paths (e.g. cancel mid-resume) intentionally leave it activated.
                     if (
-                        isinstance(event, ModelStreamChunkEvent)
-                        and event.chunk
-                        and event.chunk.get("redactContent")
-                        and event.chunk["redactContent"].get("redactUserContentMessage")
+                        self._interrupt_state.activated
+                        and (agent_result is None or agent_result.stop_reason != "interrupt")
```
strandly-the-agent, line 1341: 🟡 **Two branches in this block are unproven — mutation testing shows nothing in the repo exercises them.**

**(a) This clause.** Dropping `(agent_result is None or agent_result.stop_reason != "interrupt")` entirely still passes 154 + 889 + 300 tests. I couldn't construct a scenario that needs it: the third clause (`"tool_use_message" not in context`) already excludes every tool-interrupt completion, since tool interrupts always write `tool_use_message` before `activate()` (`event_loop.py:812-813`). The comment above describes only the tool-interrupt-preservation case, which that third clause covers. Is this load-bearing for some agent-stream multi-middleware ordering I didn't build — in which case it wants the test that pins it (assert `activated` stays `True`) — or is it left over from an earlier design and removable?

**(b) The per-pass result capture at `:1318-1322`.** Reverting it to "grab the last event after the loop" also survives everything (154 passed; 889 passed). It isn't dead code though — I confirmed on unmutated source that a middleware yielding a trailing event after a tool-interrupt stop event in the same pass gives `AfterInvocationEvent.result = AgentResult(stop_reason="interrupt")` correctly, and `None` with the mutation. It's invisible from the public API because `stream_async` recomputes its own result, but it drives `AfterInvocationEvent` (and therefore session persistence) and this deactivate gate.

For (b): a trailing-event-after-tool-interrupt case asserting `AfterInvocationEvent.result.stop_reason` via the existing `RecordingHooks` pattern would close it — no test currently asserts the *value* of that hook's result, only that it fires.

For context, the mutations the suite **did** catch: both event-loop resume gates reverted to `activated` alone (6 failures), the `"tool_use_message" not in context` guard removed (the new cancellation test), the whole-stream result selection reverted (3 failures), and the `RuntimeError` guard removed (caught on the specific message, not just "raises"). The suite is strong where it aims — these two branches are just outside its aim.

strands-py/src/strands/_middleware/README.md
```diff
@@ -104,11 +118,14 @@ This is inherent to the SDK-wide interrupt design (the same is true for hooks/to
...
 
-Interrupt IDs are `v1:middleware_execute_tool:<toolUseId>:<uuid5(name)>` — deterministic across
-resumes so a resumed response resolves the same interrupt. This follows Python's `v1:`
-interrupt-id scheme (`v1:tool_call:...`, `v1:before_tool_call:...`) and its convention of
-hashing the name with `uuid5`. (TS uses a different, unversioned literal — id *strings* are
-opaque per-SDK handles and are not compared across SDKs, so only the within-SDK scheme matters.)
+Interrupt IDs are `v1:middleware_execute_tool:<toolUseId>:<uuid5(name)>` for tool middleware and
+`v1:middleware_agent_stream:<uuid5(name)>` for agent-stream middleware — deterministic across
+resumes so a resumed response resolves the same interrupt. The tool id is scoped per tool call;
+the agent-stream id is scoped per invocation pass (there is no tool-call to key on). This follows
```
strandly-the-agent: 🟡 **"the agent-stream id is scoped per invocation pass" isn't accurate** — `stages.py:233` is `uuid5(NAMESPACE_OID, name)` with no pass-identifying component at all; the same name yields the same id in every pass, in every agent (verified).

What actually keeps it safe is the lifecycle, not the id: only one agent-stream interrupt is live at a time and `deactivate()` clears `interrupts` before the next pass could reuse the name. Worth saying that instead, because a reader who believes the id is pass-scoped will assume collision-safety it doesn't have (see the `Graph`/`Swarm` cross-agent case on `stages.py:233`).

Suggested rewording: *"The tool id is scoped per tool call (it embeds `toolUseId`); the agent-stream id has no per-call token — it is deterministic from `name` alone. Safety comes from the interrupt lifecycle instead: only one agent-stream interrupt can be active at a time, and a completed pass's `deactivate()` clears `interrupts` before the name could be reused."*

strands-py/src/strands/_middleware/README.md
```diff
@@ -118,7 +135,30 @@ hooks, tools, or middleware — so there is nothing for the middleware path to s
...
 choice; adding it means changing the core `Interrupt` type and every hook/tool call site, which
 is out of scope here. Consumers currently disambiguate by the interrupt id prefix
-(`v1:middleware_execute_tool:...`) instead.
+(`v1:middleware_execute_tool:...` / `v1:middleware_agent_stream:...`) instead.
+
+## AgentStreamStage context fields
+
+The context fields are named for the Python run loop, not TS's: TS's `AgentStreamContext` exposes
+`args` + `options`; Python exposes `messages` (the input for this pass, already appended to
+history) + `invocation_state` (the per-invocation state dict). This is the same idiomatic remap
```
strandly-the-agent: 🟡 **"This is the same idiomatic remap the other stages use" doesn't hold — and it papers over a real capability gap.**

On the precedent: `InvokeModelContext` and `ExecuteToolContext` mirror TS's field names one-to-one modulo casing (`systemPrompt`→`system_prompt`, `toolUse`→`tool_use`; `strands-ts/src/middleware/stages.ts:69-90,104-113`), which is the cross-SDK re-casing rule (`AGENTS.md:49`), not a remap. `AgentStreamContext` is the only stage that renames to genuinely different fields, so there's no sibling pattern to point at.

On the substance: TS's `options` is `InvokeOptions` (`strands-ts/src/types/agent.ts:92-105`), which also carries `cancelSignal`, `structuredOutputSchema` and `limits` — all visible and overridable by TS middleware. The Python equivalents (`self._cancel_signal`, `structured_output_context`, `limits`) are closed over by `_make_agent_stream_terminal` (`agent.py:1379-1420`) and never surface on the context, so an `AgentStreamStage.Input` handler can't see or change any of them. "Same idiomatic remap" reads as parity where there's a subset.

If the subset is a deliberate scope cut (fine by me — see question 2 in the summary), naming what TS's `options` gave middleware that `invocation_state` doesn't would make this section land honestly.

strands-py/tests/strands/middleware/test_agent_stream_middleware.py
```diff
@@ -0,0 +1,798 @@
...
+
+    def tag_state(context):
+        context.invocation_state["tagged"] = True
+        return context
+
+    agent._middleware_registry.add_middleware(AgentStreamStage.Input, tag_state)
+    agent._middleware_registry.add_middleware(AgentStreamStage, capture)
+    agent("Test prompt")
+
+    assert received_messages is not None
```
strandly-the-agent: 🟡 `test_input_transforms_context` passes with the transform entirely inert — which is exactly the state of things (see `agent.py:1400`).

The test name promises "an Input handler can transform the context before the stream runs", but the only assertion is that `received_messages` isn't `None`, i.e. that a wrap handler saw *some* messages. `tag_state`'s `context.invocation_state["tagged"] = True` is never asserted anywhere, and a `replace()`-style transform would be silently dropped by the terminal without failing this test.

Worth asserting the transformed value actually reached the event loop / model — the way the `InvokeModelStage` equivalent does — so this test would fail if the terminal ignores its context:

```suggestion
    assert received_messages is not None
    assert agent_invocation_state_seen_by_event_loop["tagged"] is True
```

(Adjust the second assertion to whatever hook/mock you use to observe the state the event loop actually ran with — the point is to pin the transform's effect, not just that a handler ran.)

strandly-the-agent (changes requested):
**Assessment: changes requested** — solid port of the final middleware stage, but two grounded 🔴 defects in the new middleware-interrupt path: an interrupt id with no per-agent scoping that can silently cross-wire (or bypass) a human approval across `Graph`/`Swarm` nodes, and an unconditional `deactivate()` that wipes an in-flight agent-stream interrupt in any pass that also runs a tool — livelock, no workaround. Blast radius today is capped (stage is unexported, zero in-repo consumers), but this *is* the `area-hil` path the feature exists for.

`opieter-aws`'s guard request (`"tool_use_message" not in context`) is applied, and I verified it's load-bearing — deleting it breaks a resume with `ValueError: Received interrupt responses but agent is not in interrupt state`. Nothing from the earlier round is re-raised.

<details>
<summary><b>✅ Verification ledger</b> — what was actually run at <code>125cee8b</code></summary>

| | |
|---|---|
| Reviewed | `125cee8b` (force-pushed 14:19Z today), strict `0e517d62..HEAD`, 6 files +1134/−55 |
| ✅ Tests | `tests/strands/middleware/` + `test_agent_cancellation.py` → **154 passed**; `agent/` + `interventions/` + `session/` → **889 passed**; `test_interrupt.py` + `types/test_interrupt.py` + `event_loop/` + `multiagent/` → **300 passed** |
| ✅ Mutation testing | 4 of 6 mutations **caught** (both event-loop gates → 6 failures; the deactivate guard → the cancellation test; result-selection + `RuntimeError` → 3 failures, incl. a match on the specific message). 2 survived → finding #6 |
| ✅ Repros confirmed | livelock (5 resumes → 10 model calls, 5 tool runs), cross-agent id collision in a 2-node `Graph`, duplicate model call + back-to-back assistant turns on an Output-phase interrupt, inert `replace()`, interrupt under-reporting, structured-output loss on short-circuit |
| ✅ Correct behavior confirmed | session round-trip of an agent-stream interrupt; tool↔agent-stream↔tool in every ordering; cancel-during-resume both directions; Swarm node interrupt→resume; `SlidingWindowConversationManager` truncation parity; drop-the-stop-event → the actionable `RuntimeError`; duplicated/reordered stop events → last one wins; mypy + ruff clean |
| ✅ CI | all green **except `label-size`** — mechanical (`pr-size-labeler.yml:26,55` unconditionally fails `size/xl`; 1189 > 1000), precedented by sibling #3233 which merged at `size/xl`. Not actionable. |
| Not verified | the provider-side `ValidationException` from the non-alternating history in finding #3 (no live endpoint call) — the duplicate model call and back-to-back assistant messages are confirmed regardless |

Pre-existing, verified identical on `0e517d62` (**not** this PR's, no action asked): the OTel `Failed to detach context` warning on mid-stream generator close (the new mid-drain interrupt path does trigger it more often), duplicate `toolResult` after a second cancel-during-tool-interrupt-resume, and `BeforeInvocationEvent` not being interruptible.
</details>

<details>
<summary><b>⚪ Nits</b> (5, bundled — their anchors fall outside the diff)</summary>

- **`README.md:209-211` "Defensive copies"** — accurate for `InvokeModelContext`, but now unscoped: `AgentStreamContext.messages` also exists and is shared *by reference* (`stages.py:183-186`), and the new context-fields section doesn't say so either. Scope the heading and add a line for `ExecuteToolContext`'s shallow copy + this stage's by-reference sharing.
- **`stream_async` docstring `Raises:` (`agent.py:1122-1125`)** — itemizes `ConcurrencyException`/`IdempotencyAbortedError`/`TypeError` plus a generic catch-all; the new deliberate `RuntimeError` (`agent.py:1216-1219`) is part of this PR's contract and deserves its own bullet per `strands-py/AGENTS.md:112`.
- **Middleware short-circuit + `structured_output_model`** — `agent("…", structured_output_model=Answer)` under a short-circuiting wrap returns `end_turn` / `structured_output=None` with no warning, indistinguishable from "the model didn't produce one" (verified live by two passes). One line in the README's short-circuiting section + a test pinning the current behavior.
- **README is missing TS's "middleware cannot resume interrupts"** — the same structural fact holds in Python (`resume()` at `agent.py:1166,1373`, both outside the per-pass `try`/`except`) and TS documents it (`strands-ts/src/middleware/README.md:79-83`); this PR added two sections on resume mechanics without it.
- **`registry.py:141-146`** — a middleware whose cleanup raises during `aclose()` replaces an in-flight `InterruptException`, so it's never registered (interrupt silently lost). Pathological precondition; worth one sentence next to the already-documented `except Exception` hazard (`README.md:113-118`).
</details>

### Questions

**Worth answering before merge:**
1. Is *"an agent-stream interrupt replays the entire pass"* an intentional hard constraint — i.e. is the contract meant to be "only call `interrupt()` before `next()`"? Findings #2 and #3 are both downstream of that one unstated decision. If it is intended, it probably wants stating and enforcing (raise if `interrupt()` is called after `next_fn` drained); if not, agent-stream interrupts likely need their own per-pass completion marker the way tool interrupts have `tool_use_message`.
2. Is `AgentStreamContext` *not* exposing `cancel_signal` / structured-output / `limits` a deliberate scope cut? TS's `options` carries all three and lets middleware override them (`strands-ts/src/types/agent.ts:92-105`). That decides whether finding #8 is a README reword or a missing field.

**Non-blocking:** the body promises a follow-up opening up `InvokeModelStage`/`ExecuteToolStage` — once this merges and auto-closes #3150, that sentence is the only record of the plan (I found no tracking issue). Worth filing one, or extending #2498?

---
🤖 Independent multi-pass review (correctness · adversarial/repro · issue-alignment · test-quality · docs-accuracy), aggregated and culled. Solid work for a human to approve — not a gate; push back where I'm wrong.

strands-py/src/strands/agent/agent.py
zastrowm: Applied — the guard now checks `"tool_use_message" not in context`, with a regression test in `test_agent_cancellation.py` proving a cancelled resume preserves the pending tool interrupt.

strands-py/src/strands/event_loop/event_loop.py
zastrowm: Fixed — `AgentStreamContext` now resolves prior responses from a snapshot taken before the pass, so a tool cycle's mid-pass `deactivate()` can't strand a gate that re-reads after `next_fn`. Same fix applied to TS.

strands-py/src/strands/agent/agent.py
zastrowm: Fixed — the terminal now reads `invocation_state` from the context it receives, so a `replace()`-based transform reaches the event loop. Added a test asserting the transformed state reaches the model.

strands-py/src/strands/agent/agent.py
zastrowm: Fixed — the agent-stream stop now reports all still-unanswered interrupts (matching TS `getUnansweredInterrupts()`), with a test covering the mixed tool + agent-stream case.

strands-py/src/strands/_middleware/README.md
zastrowm: Fixed — reworded to say the id is a pure hash of `name` (no pass component) and that within-agent safety comes from the lifecycle, not the id.

strands-py/src/strands/_middleware/README.md
zastrowm: Fixed — corrected to note this is a genuine rename (not the re-casing the sibling stages use) and called out the dropped `options` fields (cancel_signal/limits/structured-output).

strands-py/tests/strands/middleware/test_agent_stream_middleware.py
zastrowm: Fixed — replaced with `test_input_transforms_context_reaches_event_loop`, which asserts a `replace()`-injected marker reaches the model via a BeforeModelCallEvent hook.

mkmeral:
@strandly-the-agent  do a full review of the PR. is it good to merge? where are we with the PR? what issues do we have?

strandly-the-agent:
@mkmeral Full re-review done at `6653477`. Short answer: **not yet — the head is red, and the failing test is one this PR added.** The good news is the failure isn't this PR's bug: `main` is broken and this PR's test is what catches it.

**Where we are:** 6 of my 9 original findings are fixed and I verified each independently (I re-ran the original repros, and mutation-tested all three correctness fixes — each one now breaks a test, so they're pinned, not just patched). The TS port added since my last pass is clean. What's left is 1 blocker inherited from `main`, 1 fix that's incomplete, and 2 knowingly-deferred items.

| | Issue | Status |
|---|---|---|
| 🔴 | `test_cancel_during_tool_interrupt_resume_preserves_interrupt_state` fails at head — cancelling a resume wipes a pending tool interrupt | **Blocker, cause is in `main` (#3508), not this PR** |
| 🔴 | An answered agent-stream gate is re-asked once per later interrupt round trip | **Open** — my original livelock 🔴, fixed only intra-pass |
| 🟡 | Post-`next_fn` interrupt replays the pass (2nd model call, non-alternating history) | Open — needs a decision, not necessarily code |
| ➖ | Cross-agent interrupt-id collision | Deferred, now honestly documented in the README |
| ✅ | livelock · terminal `invocation_state` · all-unanswered-interrupts · 2 README claims · vacuous test | Fixed, verified, mutation-pinned |

<details>
<summary><b>🔴 1 — the head is red, and it's <code>main</code>'s bug (evidence)</b></summary>

I ran the same test at three revisions:

```
2f0c66f  (PR tip BEFORE the main merge)   -> 9 passed
main 1937b4ae + this PR's test grafted on -> FAILED
6653477  (current head)                   -> 1 failed, 8 passed
```
```
assert agent._interrupt_state.activated
E   assert False
E    +  where False = _InterruptState(interrupts={}, context={}, activated=False).activated
tests/strands/agent/test_agent_cancellation.py:335
```

So the PR's own commit is correct — `main`'s #3508 (`BeforeToolsEvent`/`AfterToolsEvent`) restructured `_handle_tool_execution` and dropped the early return on the cancel path, so a cancelled resume now falls through to the unconditional `deactivate()` at `event_loop.py:889` and destroys the pending tool interrupt: the human's approval and `tool_use_message` are gone and the invocation is unresumable. **That is a live bug on `main` today**, independent of this PR — worth its own issue, and I'd suggest filing it rather than only fixing it here.

Interestingly the PR's *TypeScript* side already has the right guard (`agent.ts`: `if (this._interruptState.activated && !this.isCancelled)`); only Python lost it in the merge. A one-line gate at `event_loop.py:889` restores it:

```python
if not cancel_message:
    agent._interrupt_state.deactivate()
```

Verified: with that change, `tests/strands/agent/ + event_loop/ + middleware/` → **1039 passed** (from 1 failed / 1038 passed). Two independent passes reached the same fix.

One consequence to note: this also makes the new README sentence at `_middleware/README.md:218-219` false as written — it claims both SDKs preserve a pending tool interrupt across a cancelled resume. TS does; Python currently doesn't.

**On CI:** `Python / Unit Tests - Python 3.14 - macOS` failed with exit code 1 and fail-fast cancelled the *entire* rest of the Python matrix — so no Python unit-test job actually passed on this head. I couldn't read the CI log, so I can't prove it's this test, but the local failure is deterministic (3/3) and platform-independent, and macOS 3.14 simply started earliest, so it's the obvious candidate. Worth knowing that the green-looking checks are mostly TypeScript.
</details>

<details>
<summary><b>🔴 2 — the livelock fix is intra-pass only: an approved gate gets re-asked</b></summary>

Credit where due: the original livelock is genuinely gone (5 resumes → 10 model calls became 1 resume → converged). The pre-pass snapshot (`agent.py:1319`) works *within* a pass.

But the snapshot is copied **from live state at pass start**, so it can't recover what a *previous* pass's tool cycle already destroyed — `deactivate()` at `event_loop.py:889` is still a full reset (`interrupts` **and** `context`). A gate that reads its approval only before `next_fn` (the documented shape), one agent, one invocation:

```
HUMAN WAS ASKED: {'send_gate': 4, 'tool_gate_1': 1, 'tool_gate_2': 1, 'tool_gate_3': 1}
obs: (2, 'snapshot', ["middleware_agent_stream/4e79ef='OK1'"])
     (3, 'snapshot', ["tool_call/fc5c20='OK2'"])      <- send_gate GONE from the snapshot
```

The human approves the same action 4 times. As a drop-in test:
```
E   AssertionError: asked=['stream_gate', 'tool_gate', 'stream_gate']
E   assert 2 == 1
```

Worse in the realistic HITL deployment shape (session manager + a fresh `Agent` per pass — which is how you'd actually build this): it **doesn't converge**, alternating gates for 6 resumes with `{'stream_gate': 3, 'tool_gate': 3}` and tool side effects re-firing each time. That's why I'm keeping this at 🔴 rather than downgrading it: any HITL flow that gates a pass *and* has tool approvals hits it, and a gate with a side effect on resolution double-fires.

Suggested direction (validated by monkeypatch, not a source edit — treat as a sketch, the ownership call is yours): have the reset preserve *resolved* agent-stream responses and drop them when the invocation ends, rather than at every tool cycle. With that, my regression test passes and `tests/strands/middleware/` → 150 passed, agent/event_loop/session/multiagent → 1360 passed.

Note the same structural hole appears to exist in the TS port (`agent.ts` snapshots per pass, `interrupt.ts` `deactivate()` clears wholesale) — I did **not** run the TS suite against it, so treat that as unverified.
</details>

<details>
<summary><b>What I verified as genuinely fixed (and how)</b></summary>

All three correctness fixes are mutation-pinned — I broke each one and a test caught it:

| Fix | Mutation | Caught by |
|---|---|---|
| Pre-pass interrupts snapshot | read live state instead of the snapshot | `test_resumed_interrupt_reads_response_after_a_tool_cycle_runs` |
| Terminal reads `ctx.invocation_state` | use the closure var instead | `test_input_transforms_context_reaches_event_loop` |
| Report all unanswered interrupts | revert to the single raised interrupt | `test_agent_stream_interrupt_reports_all_unanswered_interrupts` |

Also verified: `replace(invocation_state=…)` now reaches the model, the tool hooks and the event loop; the all-unanswered list does **not** over-report (answered interrupts are filtered, no cross-agent leak); the snapshot survives a `FileSessionManager` persist+restore round trip; the 6 new tests all assert real observable state (hook-observed model input, message-history content, exact response values) — none would pass against a no-op; the two rewritten README claims are now accurate, including an honest write-up of the cross-agent id collision that names the `agent_id == "default"` problem.

**TS side (new since my last review, so first look):** the snapshot port matches Python in mechanism and lifetime, the tests are meaningful, and `npx vitest run src/agent src/middleware` → **569 passed**. mypy + ruff clean on Python. Local Python suites: middleware 150, `agent/` 724 (+1 fail = the blocker), `event_loop/` 164, session 142, multiagent 145.

Still-standing 🟡 from my last pass, unchanged: two branches remain unproven by mutation (the `stop_reason != "interrupt"` clause and the per-pass result capture — both still survive deletion across ~1200 tests). Minor new nits: `replace(invocation_state=…)` is invisible to invocation-level hooks and to `AgentResult.state` on an interrupt stop (split-brain, ⚪); `clearPendingToolExecution()` in `strands-ts/src/interrupt.ts` is now dead code (⚪).
</details>

### What I'd do next

1. **Fix the cancel guard** (one line, above) so the head goes green — and file the `main` bug separately, since it affects `main` regardless of this PR. Happy to open that issue if you want it.
2. **Decide on the re-asked gate** (🔴 2) — this one I'd not merge without, since it breaks the feature's own headline use case.
3. **Decide, don't necessarily code, the post-`next_fn` replay** — either a completion marker or an actionable error. "Documented as unsupported" is a perfectly good answer; silently re-calling the model isn't.

On merge-blocking: my earlier review is still in changes-requested state, which is contributing to the `BLOCKED` status. Items 1 and 2 are why I'm leaving it there — if you'd rather treat 2 as a follow-up and ship, that's a reasonable maintainer call and I'll clear it on request. `label-size` (1618 lines) is mechanical and precedented by #3233; I wouldn't hold on it.

---
🤖 Verified at `6653477` by re-running every repro and mutation myself; three independent passes (adversarial, test-quality, correctness/parity) plus my own confirmation of the two blockers. Push back where I'm wrong.

mkmeral:
@strandly-the-agent create a new PR. fork this PR essentially, fix the issues, do reviews with subagents and loop until you are done. then create a PR, tag me

strandly-the-agent:
Cross-linking: at @mkmeral's request I forked this branch and fixed the open items in **#3594**. That PR contains this work plus the fixes, so it's an either/or with this one rather than something to merge alongside.

What it changes on top of `6653477`: the cancelled-resume state loss (which is a `main` regression from #3508 — `test_cancel_during_tool_interrupt_resume_preserves_interrupt_state` fails on `main` today, so this branch is red through no fault of its own), the answered-gate re-ask, the post-result pass replay, and the two unexercised run-loop branches. `strands-ts/` is untouched from here.

<details>
<summary>Two things worth knowing if you review it</summary>

- The first fix I wrote for the re-ask leaked the approval into the session, so a restored agent skipped the human entirely — worse than the bug. Three independent review rounds caught that and a second regression I introduced; the ledger is in the PR body.
- The cross-agent interrupt-id collision is still deferred, and the id-prefix-vs-explicit-`scope`-field question is called out as needing a maintainer's design call rather than being decided in a bug-fix branch.
</details>

liramon2:
Changes merged in:
- #3615
- #3635
- #3594

PR #3467, opened 2026-07-24
docs(site): document trusted message history across agent input paths

yonib05:
## Description

Documents message history that reaches an agent from a source outside the application's control.

Message content carries more than text: it can carry tool-call and tool-result blocks. A forged tool-result block placed in history that reaches the model can misrepresent what a tool returned and steer the model's next step. This is a concern in both SDKs. The Python SDK has a sharper version: a tool-call block as the most recent message causes the agent to run that tool directly on its next invocation, with no model call in between, so the block's author chooses the tool and its arguments outright.

The reliable control is the trust boundary: build an agent's message history from your own application, not from input a caller can shape. For Python applications that must accept caller-influenced history, the page also shows how to clear a trailing tool-call block, while noting that this closes only the direct-dispatch path and does not make untrusted content safe.

Changes:
- New **Trusted Message History** page under Safety & Security.
- Short callouts linking to it from the **Prompts** and **Snapshots** pages. The Prompts callout is split per language: the Python tab describes the direct-dispatch behavior, the TypeScript tab describes the forged-content concern.
- Navigation entry under Safety & Security.

## Testing

Local doc-site checks (`typecheck:snippets`, `format`, `build`) could not be run in this environment because the npm registry rejected authentication during `npm install`. Snippet marker references, doc frontmatter description length, and the Python example's edge-case behavior were verified by hand. Please confirm the site build passes in CI.


site/src/content/docs/user-guide/concepts/agents/snapshots.mdx
```diff
@@ -85,6 +86,72 @@ print(len(agent.messages))  # Only the messages from before the jokes
...
+
+```typescript
+--8<-- "user-guide/concepts/agents/snapshots_imports.ts:clear_trailing_tool_use_imports"
+
+--8<-- "user-guide/concepts/agents/snapshots.ts:clear_trailing_tool_use"
+```
+</Tab>
+</Tabs>
+
+We may reconsider this default in a future major version.
```
zastrowm: > We may reconsider this default in a future major version.

What's the context here? - this is the entire purpose of snapshots: to preserve the state exactly where it was.  Maybe we can be more clear about that, but I don't see reconsidering this.

I am reluctant to even show an example  of how to sanitize - we removed toolUse, but what about prompt injection or other data in the snapshot?  Modifying the modelState data can connect it to a different conversation for Stateful model providers.

I would simply state that Snapshots must be trusted data and that it should not be user-provided without an example

site/src/content/docs/user-guide/concepts/agents/prompts.mdx
```diff
@@ -42,6 +42,14 @@ If you do not specify a system prompt, the model will behave according to its de
 
 These are your queries or requests to the agent. The SDK supports multiple techniques for prompting.
 
+:::caution[Passing a message list from an untrusted source]
```
zastrowm, line 48: If this is in Python only we should have it in language tags (or whatever it is); no need for this to show up when TS is selected

zastrowm: (off topic: We should update the review agent to catch this)

site/src/content/docs/user-guide/safety-security/trusted-message-history.mdx
```diff
@@ -0,0 +1,49 @@
...
+tags: [safety, conversation-management]
+sourceLinks:
+  - path: strands-py/src/strands/event_loop/event_loop.py
+  - path: strands-py/src/strands/agent/agent.py
+  - path: strands-ts/src/agent/agent.ts
+---
+
+An agent treats its message history as trusted input. When you build that history from a source you do not control (a request body, a stored conversation, a loaded snapshot) treat it as untrusted, because message content carries more than text: it can carry tool-call and tool-result blocks.
+
+Forged tool content is a concern in both SDKs. A tool-result block you did not produce, placed in history that reaches the model, can misrepresent what a tool returned and steer the model's next step. The Python SDK has a sharper version: a tool-call block as the most recent message causes the agent to run that tool directly on its next invocation, with no model call in between, so the block's author chooses the tool and its arguments outright.
```
zastrowm: > The Python SDK has a sharper 

I'd conditionalize this to Python only.

Can we change the verbage?

> Invoking an agent with content other than a string is a pointed example of this: the input is considered trusted and a tool-call block as the most recent message causes the agent to run that tool directly on its next invocation, with no model call in between

site/src/content/docs/user-guide/safety-security/trusted-message-history.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+## Do not populate history from an untrusted source
+
+The reliable control is the trust boundary. Build an agent's message history from your own application, not from input a caller can shape. History your application produces, or persists to storage only it can write, is trusted; load it as-is.
+
+## Clearing a trailing tool-call block
```
zastrowm, line 21: Again Python only

zastrowm (dismissed):

zastrowm (dismissed):

zastrowm (approved):

PR #2150, opened 2026-04-17
fix: include reasoning block with empty text and non-empty signature

gaurav71531:
## Motivation

Some models like Opus 4.7 emit reasoning content blocks that carry only a `signature` with no accompanying `reasoningText`. The prior condition in `handle_content_block_stop` checked only `reasoning_text`, which evaluated to falsy for an empty string, causing the entire reasoning content block (including the signature) to be silently dropped from the assembled message content.

## Public API Changes

No public API changes. This is an internal streaming assembly fix.

## Change

The condition in `handle_content_block_stop` is broadened from:

```python
elif reasoning_text:
```

to:

```python
elif reasoning_text or "signature" in state:
```

This ensures that reasoning blocks carrying only a signature are preserved in the output content, matching the behavior models like Opus 4.7 expect.

## Testing

Added two tests covering the signature-only reasoning scenario:
- A parametrized case in `test_handle_content_block_stop` for empty `reasoningText` with a `signature` present
- An end-to-end `test_process_stream_with_signature_only` that simulates the Opus 4.7 streaming pattern (signature-only reasoning block followed by a text response)

Issue #1698, opened 2026-02-15
[BUG] Extended thinking breaks multi-turn sessions: SDK-constructed assistant messages lack `reasoningContent` blocks


zhenyanghua:
### Checks

- [x] I have updated to the lastest minor and patch version of Strands
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched [./issues](./issues?q=) and there are no duplicates of my issue

### Strands Version

1.26.0

### Python Version

3.11

### Operating System

Amazon Linux 2023 (Bedrock AgentCore Runtime), also reproducible on macOS

### Installation Method

pip

### Steps to Reproduce

1. Create a `BedrockModel` with extended thinking enabled:

```python
from strands import Agent
from strands.models import BedrockModel

model = BedrockModel(
    model_id="us.anthropic.claude-sonnet-4-20250514-v1:0",
    additional_request_fields={
        "thinking": {
            "type": "enabled",
            "budget_tokens": 4096,
        },
    },
)
```

2. Create an agent with a session manager that persists conversation history (e.g., `AgentCoreMemorySessionManager` or any custom `SessionManager` implementation):

```python
agent = Agent(
    model=model,
    tools=[some_tool],
    session_manager=my_session_manager,  # persists messages across turns
)
```

3. Send a prompt that triggers a tool call:

```python
agent("What is the status of zone X?")
# Agent calls some_tool, gets result, responds successfully.
```

4. Send a second prompt in the same session:

```python
agent("Now tell me about zone Y")
# FAILS with ValidationException
```

### Expected Behavior

The second turn should work correctly. The session manager loads the persisted conversation history (including assistant messages from turn 1), and the model should be able to continue the conversation with extended thinking enabled.

### Actual Behavior

The second turn fails with:

```
botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the ConverseStream operation: If an assistant message contains any thinking blocks, the first block must be thinking. Found text
```

Or, if the first turn's assistant messages had no thinking blocks at all:

```
botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the ConverseStream operation: A conversation must alternating user and assistant messages...
```

## Root Cause Analysis

When extended thinking is enabled, Bedrock/Claude requires **every** assistant message to start with a `reasoningContent` block. The Strands SDK event loop constructs assistant messages during tool-use cycles that **do not** include `reasoningContent` blocks.

Here's what happens step by step:

[...]

Issue #790, opened 2025-09-04
[BUG] only set signature in message if signature was provided by the model


howardzhou:
### Checks

- [x] I have updated to the lastest minor and patch version of Strands
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched [./issues](./issues?q=) and there are no duplicates of my issue

### Strands Version

1.7.0

### Python Version

3.12

### Operating System

ubuntu 22.04

### Installation Method

pip

### Steps to Reproduce

Run gpt oss through bedrock

### Expected Behavior

Remove this line, 
https://github.com/strands-agents/sdk-python/blob/main/src/strands/event_loop/streaming.py#L292

it breaks the fix in previous PR which fixed the model not having signature 
https://github.com/strands-agents/sdk-python/pull/682/files

### Actual Behavior

GPT oss doesn't have signature, this makes the context messages having 'signature': '', which breaks the bedorck inference

### Possible Solution

Remove this line, 
https://github.com/strands-agents/sdk-python/blob/main/src/strands/event_loop/streaming.py#L292

See details in PR https://github.com/strands-agents/sdk-python/pull/682/files

howardzhou:
@clareliguori

zastrowm:
Confirmed the problem & working on a fix.  Adding unit test to catch future regressions

zastrowm:
Per #799, this can be reproduced with:

```python
from strands import Agent, tool
from strands.models import BedrockModel

@tool
def calculator(expression: str):
    """Calculate the expression"""
    return "not supported"

model_id = "openai.gpt-oss-120b-1:0"
model = BedrockModel(
    model_id=model_id,
)
agent = Agent(
    model=model,
    # tools=[calculator],
    tools=[calculator],
    system_prompt="You're a helpful assistant.",
)

# Async function that iterators over streamed agent events
async def process_streaming_response():
    agent_stream = agent.stream_async("Calculate 2+2 using a tool")
    async for event in agent_stream:
        print(event)

# Run the agent
await process_streaming_response()
```

dbschmigelski:
Hi, this should be resolved with the 1.7.1 release https://github.com/strands-agents/sdk-python/releases/tag/v1.7.1

PR #3472, opened 2026-07-24
fix(streaming): consume reasoning signature per content block

### LINKED ISSUE #3425: [BUG] Reasoning signature leaks across content blocks — signatures concatenate when one message has multiple signed thinking blocks
### Checks

- [x] I have searched [./issues](../issues?q=) and there are no duplicates of my issue

### SDK Language

Python

### Strands Version

1.48.0 (the bug is also present on current `main`)

### Language Runtime Version

Python 3.14.6

### Operating System

macOS (Darwin 25.5)

### Installation Method

pip

### Steps to Reproduce

`handle_content_block_delta` accumulates signature deltas (`state["signature"] += …`) — correct within one block, since signatures stream in chunks. But `handle_content_block_stop` never clears `state["signature"]` when it finalizes a reasoning block, unlike every other accumulator (`text`, `reasoningText`, `redactedContent`, `current_tool_use`).

Minimal repro through `process_stream`:

```python
response = [
    {"messageStart": {"role": "assistant"}},
    {"contentBlockDelta": {"delta": {"reasoningContent": {"text": "first"}}, "contentBlockIndex": 0}},
    {"contentBlockDelta": {"delta": {"reasoningContent": {"signature": "SIG1"}}, "contentBlockIndex": 0}},
    {"contentBlockStop": {"contentBlockIndex": 0}},
    {"contentBlockDelta": {"delta": {"reasoningContent": {"text": "second"}}, "contentBlockIndex": 1}},
    {"contentBlockDelta": {"delta": {"reasoningContent": {"signature": "SIG2"}}, "contentBlockIndex": 1}},
    {"contentBlockStop": {"contentBlockIndex": 1}},
    {"messageStop": {"stopReason": "end_turn"}},
]
# process_stream(...) -> constructed message carries signatures ["SIG1", "SIG1SIG2"]
```

Real-world trigger: an Anthropic model with `thinking: adaptive` plus tool use produces assistant messages with multiple signed thinking blocks; any follow-up request that re-sends the conversation (tool-loop continuation, a second invoke on the same agent) then fails.

### Expected Behavior

Each reasoning block keeps exactly the signature that was streamed for it (`SIG1`, `SIG2`).

### Actual Behavior

Signatures concatenate cumulatively across blocks (`SIG1`, `SIG1SIG2`, …). Providers verify thinking signatures byte-identical per block when the conversation is sent back, so the next request is rejected:

```
400 invalid_request_error: "…`thinking` or `redacted_thinking` blocks in the latest assistant message cannot be modified…"
```

### Additional Context

Secondary effect of the same leak: because the stale `"signature"` key stays in `state`, the `elif reasoning_text or "signature" in state:` branch (introduced in #2150) can turn later empty content blocks into spurious reasoning blocks carrying the stale signature.

### Possible Solution

Consume the signature when the block is finalized:

```python
if (signature := state.pop("signature", None)) is not None:
    content_block["reasoningContent"]["reasoningText"]["signature"] = signature
```

PR with this fix plus a regression test follows.

### Related Issues

#2150 (introduced the signature-in-state guard), #790 / #1698 (adjacent signature handling issues)

arielnabavian:
## Description

`handle_content_block_delta` accumulates reasoning-signature deltas into `state["signature"]` (correct — signatures stream in chunks). But `handle_content_block_stop` read that value **without ever clearing it** when finalizing a reasoning block, unlike every other accumulator (`text`, `reasoningText`, `redactedContent`, `current_tool_use`).

As a result, a second signed thinking block in the same message inherited the first block's signature, producing a cumulative concatenation (`"SIG1"`, `"SIG1SIG2"`, …). Providers verify thinking signatures byte-identical per block when the conversation is sent back, so any follow-up request that re-sends the conversation (tool-loop continuation, a second invoke on the same agent) was rejected:

```
400 invalid_request_error: "…`thinking` or `redacted_thinking` blocks in the latest assistant message cannot be modified…"
```

This is triggered in practice by an Anthropic model with `thinking: adaptive` plus tool use, which produces assistant messages with multiple signed thinking blocks.

**Fix:** consume the signature when the block is finalized so it belongs to exactly that block and cannot leak into the next one:

```python
if (signature := state.pop("signature", None)) is not None:
    content_block["reasoningContent"]["reasoningText"]["signature"] = signature
```

Using `pop` also fixes the **secondary effect** documented in the issue: the stale `"signature"` key had kept the `elif reasoning_text or "signature" in state:` guard (introduced in #2150) truthy, so a later empty content block was materialized as a spurious reasoning block carrying the leaked signature.

This is Python-only — the TypeScript SDK's `streamAggregated` already resets its reasoning accumulator both on block start and after finalizing a block, so no parity change is needed.

## Related Issues

Fixes #3425. Related: #2150 (introduced the signature-in-state guard), #790 / #1698 (adjacent signature handling).

## Documentation PR

No documentation changes needed — this is an internal streaming-accumulator bug fix with no API surface change.

## Type of Change

Bug fix

## Testing

- `test_process_stream_with_multiple_signed_reasoning_blocks` — the exact repro from the issue; asserts each block keeps its own signature (`["SIG1", "SIG2"]`). Reproduces `["SIG1", "SIG1SIG2"]` without the fix.
- `test_process_stream_signature_does_not_leak_into_later_empty_block` — guards the secondary effect; without the fix an empty trailing block was materialized as a spurious reasoning block carrying the leaked signature.
- Updated the two `test_handle_content_block_stop` parametrized cases so the finalized state no longer retains the consumed `signature` key.

I confirmed both new tests **fail without the fix and pass with it**. Verified locally on Python 3.12:

- `tests/strands/event_loop/` — 134 passed
- anthropic/bedrock/litellm signature & reasoning tests — 13 passed
- `hatch fmt --linter --check` (ruff + mypy over 232 source files) — clean
- `streaming.py` / `test_streaming.py` pass `ruff format --check`

> Note: I ran the individual CI-gate steps (`hatch fmt --linter --check`, scoped `ruff format --check`, and the relevant `hatch test` subsets) rather than the full `hatch run prepare`, since the whole-repo formatter flags several pre-existing unrelated files this change does not touch.

## Checklist

----

### DIFF
--- a/strands-py/src/strands/event_loop/streaming.py
+++ b/strands-py/src/strands/event_loop/streaming.py
@@ -338,8 +338,9 @@ def handle_content_block_stop(state: dict[str, Any]) -> dict[str, Any]:
             }
         }
 
-        if "signature" in state:
-            content_block["reasoningContent"]["reasoningText"]["signature"] = state["signature"]
+        # Consume the signature so it belongs to exactly this block and does not leak into the next one.
+        if (signature := state.pop("signature", None)) is not None:
+            content_block["reasoningContent"]["reasoningText"]["signature"] = signature
 
         content.append(content_block)
         state["reasoningText"] = ""
--- a/strands-py/tests/strands/event_loop/test_streaming.py
+++ b/strands-py/tests/strands/event_loop/test_streaming.py
@@ -475,7 +475,6 @@ def test_handle_content_block_delta(event: ContentBlockDeltaEvent, event_type, s
                 "current_tool_use": {},
                 "text": "",
                 "reasoningText": "",
-                "signature": "123",
                 "citationsContent": [],
                 "redactedContent": b"",
             },
@@ -515,7 +514,6 @@ def test_handle_content_block_delta(event: ContentBlockDeltaEvent, event_type, s
                 "current_tool_use": {},
                 "text": "",
                 "reasoningText": "",
-                "signature": "123",
                 "citationsContent": [],
                 "redactedContent": b"",
             },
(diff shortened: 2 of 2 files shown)


mkmeral:
@strandly-the-agent  is it safe to merge?

strandly-the-agent:
**Code-wise: yes.** I re-verified it against today's `main` — clean merge, green tests, and the new tests genuinely fail without the fix. **But GitHub won't let it merge right now**, for two reasons that both need a human:

- ⚠️ **`Run integration tests` never ran** — it's been `WAITING` since 2026-07-24T18:38 (needs a maintainer to approve the run), so the check rollup is still `PENDING` and `mergeStateStatus: BLOCKED`.
- ⚠️ **No approving review** — `reviewDecision: REVIEW_REQUIRED`. My earlier pass was a *comment*, not an approval, and a bot review wouldn't satisfy branch protection anyway.

So: approve the integ-test run, and once it's green + one maintainer approval, merge. Nothing in the code is holding it back.

<details>
<summary>What I actually checked (branch <code>agent-tasks/3425</code> @ <code>97763ec</code>, base <code>main</code> @ <code>0e517d62</code>)</summary>

| Check | Result |
|---|---|
| Merges into current `main` | ✅ clean, no semantic drift — main's only `event_loop/` change since the branch point is `event_loop.py` (#3402, MCP tool cancellation); this PR only touches `streaming.py` |
| Unit tests **on the merge commit** (py3.12) | ✅ `tests/strands/event_loop/` → 135 passed; `models/test_anthropic.py` + `models/test_bedrock.py` + `tests/strands/agent/` → 909 passed |
| New tests are real regression guards | ✅ with `main`'s `streaming.py` restored, both fail (`['SIG1', 'SIG1SIG2']`, plus the spurious trailing block `{'text': '', 'signature': 'SIG1'}`); both pass with the fix |
| CI on head `97763ec` | ✅ unit tests across py3.10–3.14 × linux/windows/macOS, `Python / Lint`, `CodeQL`, `Docs / CI`, `CI Gate`, `codecov/patch` all green |
| Integration tests | ⚠️ never executed (`WAITING` since Jul 24) |
| Open threads | 1 unresolved — my own ⚪ nit about strengthening the assertion in `test_process_stream_with_multiple_signed_reasoning_blocks`. Non-blocking; fine to merge as-is or fold in. |

**One residual sharp edge, not a blocker and not introduced here:** `strands-py/src/strands/models/anthropic.py:158` hard-indexes `content["reasoningContent"]["reasoningText"]["signature"]`. After this fix, a reasoning block that streams text but *no* signature delta yields a block with no `signature` key → `KeyError` when the conversation is re-sent, where before it silently inherited the previous block's signature (which the provider rejected anyway — that's the bug being fixed). Anthropic emits a `signature_delta` for every thinking block, so this isn't reachable via the fixed path, and on `main` an unsigned *first* block already raises the same `KeyError`. Worth a separate hardening PR (`.get("signature")` + explicit handling), not something to hold this one for.

</details>

mkmeral (approved):

Issue #3479, opened 2026-07-24
[FEATURE] Agent-as-tools handoff (Python)


liramon2:
### SDK Language

Python

### Problem Statement

When using agent-as-tools for delegation, the orchestrator always makes an additional model call to process the sub-agent's response before returning it to the user. This adds latency and token cost for routing scenarios where the sub-agent's output is the final answer. It can also corrupt structured output (JSON, code) by paraphrasing it.

There is currently no way to tell the agent loop to skip the post-processing step during agent-as-tools delegation.

### Proposed Solution

Port https://github.com/strands-agents/harness-sdk/issues/3257 to the Python SDK. The developer experience is the same: a `delegate` boolean option is added to `Agent.as_tool()`. When `true`, the agent loop short-circuits after the sub-agent's tool result is received, returning it directly as the `AgentResult` with `stopReason: 'endTurn'` instead of feeding it back to the orchestrator model.
```
# Sub-agents as tools
customer_service = Agent(
    name="CustomerService",
    system_prompt="...",
)

technical_support = Agent(
    name="TechnicalSupport",
    system_prompt="...",
)

# Orchestrator with sub-agents as tools
orchestrator = Agent(
    tools=[
        customer_service.as_tool(handoff=True),
        technical_support.as_tool(handoff=True),
    ],
)

result = orchestrator("My wifi does not work")
# result.message is the sub-agent's response directly
```

### Use Case

- **Domain routing**: Help-desk orchestrator routes billing to CustomerService, technical issues to TechnicalSupport. Specialist's response is the final answer — no orchestrator post-processing.
- **Structured output preservation**: Sub-agent generates JSON for a report. JSON reaches the user verbatim without orchestrator paraphrasing.
- **Hierarchical delegation**: Manager → Senior → Junior. Each level hands off deeper without latency accumulating on the way back up.

### Additional Context

Parent issue: https://github.com/strands-agents/harness-sdk/issues/911
Sibling issue: https://github.com/strands-agents/harness-sdk/issues/3257

liramon2:
Update 7/28:
- Added implementation and got initial strandly reviews
- Need the following merged before continuing:
https://github.com/strands-agents/harness-sdk/pull/3508
https://github.com/strands-agents/harness-sdk/pull/3466

liramon2:
Update 8/11: passes Strandly reviews, needs human review

PR #3346, opened 2026-07-18
feat(py): add agent-as-tool delegation

liramon2:
## Description

Python-side implementation of agent-as-tool delegation, matching the TypeScript SDK (#3265). When using agent-as-tools for routing, the orchestrator currently always makes an additional model call to post-process the sub-agent's response, adding latency, token cost, and risking paraphrasing structured output. The `delegate=True` option on `as_tool()` short-circuits this: the sub-agent's response is returned verbatim as the `AgentResult` with `stop_reason='end_turn'`.

```python
from strands import Agent

customer_service = Agent(name="CustomerService", description="Handles billing questions")
technical_support = Agent(name="TechnicalSupport", description="Provides technical support")

orchestrator = Agent(
    tools=[
        customer_service.as_tool(delegate=True),
        technical_support.as_tool(delegate=True),
    ],
)

result = orchestrator("My wifi does not work")
# result.stop_reason == 'end_turn'
# result.message contains the TechnicalSupport agent's response directly
```

## Related Issues

Closes #3479
Related: #3265 (TypeScript implementation), #911 (parent)

## Documentation PR

https://github.com/strands-agents/harness-sdk/pull/3484 (site docs)

#3197 (design doc)

## Type of Change

New feature

## Testing

How have you tested the change? Verify that the changes do not break functionality or introduce new warnings.

- [x] I ran `hatch run prepare`

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have reviewed and understand every line of code in this PR, including any generated by AI tools, and I can explain why it works
- [x] My change is focused and reasonably small; I have split unrelated work into separate PRs
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [ ] I have updated the documentation accordingly
- [ ] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3265, opened 2026-07-14
feat(ts): add agent-as-tool delegation

liramon2:
## Description
Users want a way to delegate tasks from orchestrator agents to sub-agents without post-processing sub-agent responses with orchestrator model calls. This PR adds a new configuration and associated logic to handle this use case in the TypeScript SDK.

The DevX looks like this:
```ts
import { Agent } from '@strands-agents/sdk'

const customerService = new Agent({
  name: 'CustomerService',
  description: 'Handles billing questions and account issues',
  systemPrompt: 'You are a customer service specialist. Help the user with their billing or account question.',
})

const technicalSupport = new Agent({
  name: 'TechnicalSupport',
  description: 'Provides technical support and troubleshooting',
  systemPrompt: 'You are a technical support specialist. Help the user troubleshoot their issue.',
})

const orchestrator = new Agent({
  name: 'HelpDesk',
  systemPrompt: 'You are a help desk coordinator. Route requests to the appropriate specialist.',
  tools: [
    customerService.asTool({ delegate: true }),
    technicalSupport.asTool({ delegate: true }),
  ],
})

const result = await orchestrator.invoke('My wifi does not work')
// result.lastMessage contains the TechnicalSupport agent's response directly
```

For more information: https://github.com/strands-agents/harness-sdk/pull/3197

## Related Issues

Closes #3257 

## Type of Change

New feature

## Testing

How have you tested the change? Verify that the changes do not break functionality or introduce new warnings.

- [x] I ran `npm run check`

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have reviewed and understand every line of code in this PR, including any generated by AI tools, and I can explain why it works
- [x] My change is focused and reasonably small; I have split unrelated work into separate PRs
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [ ] I have updated the documentation accordingly
- [ ] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

Note: will update docs in a future PR because this one is becoming huge.
----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #3257, opened 2026-07-14
[FEATURE] Agent-as-tools Handoff (TS)


liramon2:
### SDK Language

TypeScript

### Problem Statement

When using agent-as-tools for delegation, the orchestrator always makes an additional model call to process the sub-agent's response before returning it to the user. This adds latency and token cost for routing scenarios where the sub-agent's output is the final answer. It can also corrupt structured output (JSON, code) by paraphrasing it.

There is currently no way to tell the agent loop to skip the post-processing step during agent-as-tools delegation.

### Proposed Solution

Add a `delegate` boolean option to `Agent.asTool()`. When `true`, the agent loop short-circuits after the sub-agent's tool result is received, returning it directly as the `AgentResult` with `stopReason: 'endTurn'` instead of feeding it back to the orchestrator model.

The mechanism has two layers:

1. **Description suffix at read time.** When building `toolSpecs` for the model call, the agent loop appends a suffix to any handoff tool's description. This nudges the model to treat the call as terminal. The suffix is not stored on the tool itself. It's applied at the model-calling boundary.

2. **Loop short-circuit after tool execution.** After `executeTools()` completes and messages are appended to history, the loop iterates `ToolUseBlock`s in source order. If any corresponds to a handoff tool with a successful `ToolResultBlock`, the tool-result message is returned as `AgentResult.lastMessage`. On error, the loop continues normally so the model can recover.

```typescript
// TypeScript
const customerService = new Agent({ name: 'CustomerService', description: '...' })
const technicalSupport = new Agent({ name: 'TechnicalSupport', description: '...' })

const orchestrator = new Agent({
  tools: [
    customerService.asTool({ delegate: true }),
    technicalSupport.asTool({ delegate: true }),
  ],
})

const result = await orchestrator.invoke('My wifi does not work')
// result.lastMessage is the sub-agent's tool result directly
```

### Use Case

- **Domain routing**: Help-desk orchestrator routes billing to CustomerService, technical issues to TechnicalSupport. Specialist's response is the final answer — no orchestrator post-processing.
- **Structured output preservation**: Sub-agent generates JSON for a report. JSON reaches the user verbatim without orchestrator paraphrasing.
- **Hierarchical delegation**: Manager → Senior → Junior. Each level hands off deeper without latency accumulating on the way back up.

### Additional Context

Parent issue: https://github.com/strands-agents/harness-sdk/issues/911

PR #3484, opened 2026-07-25
docs(site): add agent-as-tool delegation docs

liramon2:
## Description

Add documentation covering the agent-as-tool delegation option introduced in #3265 (TypeScript) and #3346 (Python).

## Related Issues

https://github.com/strands-agents/harness-sdk/issues/3257
https://github.com/strands-agents/harness-sdk/issues/3479

## Documentation PR

This PR is the documentation itself.

## Type of Change

Documentation update

## Testing

- [x] `npm run typecheck:snippets` passes (no new type errors from the TS snippet)
- [x] `npx prettier --check` passes on the snippet file
- [x] No lines exceed 90 characters in new content

## Checklist

----

### DIFF
--- a/site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
+++ b/site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
@@ -337,29 +337,37 @@ This example demonstrates how Strands Agents SDK enables specialized experts to 
 
 ## Delegation
 
-When you mark a tool agent as a delegate, the orchestrator returns its
-response directly to the caller, skipping the extra orchestrator model round-trip.
-Use delegation when the sub-agent's response is already complete and the
-orchestrator adds no value by processing it further:
-
+Sometimes, you do not want a sub-agent's response to be re-processed by the
+orchestrator agent, since it would incur unnecessary token usage and potentially
+corrupt the sub-agent's response. For example:
 - A coding agent returning generated source code that shouldn't be rephrased
 - A customer service agent whose compliance-reviewed language must reach
   the user unchanged
 - A retrieval agent returning structured data the caller consumes directly
 
-To mark a tool agent as a delegate, pass `delegate: true` to
-<Syntax py=".as_tool()" ts=".asTool()" />:
+In that case, you can mark a tool agent as a delegate by passing `delegate: true` to
+<Syntax py=".as_tool()" ts=".asTool()" />. When you do so, the orchestrator agent
+returns the delegate's response directly to the user, skipping the extra orchestrator
+model round-trip.
 
 <Tabs>
 <Tab label="Python">
 
 ```python
+from pydantic import BaseModel
 from strands import Agent
+
+
+class BillingResponse(BaseModel):
+    summary: str
+    refund_amount: float
+
 
 billing_agent = Agent(
     name="billing_expert",
     description="Answers billing questions: charges, refunds, and invoices.",
     system_prompt="You handle billing questions with precision.",
+    structured_output_model=BillingResponse,
 )
 
 orchestrator = Agent(
@@ -369,7 +377,7 @@ Answer general questions yourself.""",
 )
 
 result = orchestrator("Why was I charged twice?")
-# result contains the billing agent's response directly
+# result contains the billing agent's structured JSON response, unchanged
 ```
 </Tab>
 <Tab label="TypeScript">
@@ -387,12 +395,12 @@ model processing. Streaming events from the delegate surface natively in
 the parent's stream, so callers see the specialist's tokens as they arrive.
 
 Limitations to be aware of:
-- **Single delegated tool per turn**: Delegated and non-delegated tools
-  can exist on the same agent. However, a delegated tool must be
-  the only tool call in its turn. The SDK cancels the batch if
-  other tools are requested alongside it.
+- **Single delegated tool per turn**: An agent can have both delegated and
+  non-delegated tools. However, if the agent calls a delegated tool during
+  a turn, that tool must be called alone. The SDK cancels the tool
+  batch if other tools are requested alongside the delegated tool.
 - **Incompatible with stateful models**: Delegation exits the
-  loop early, which would leave an unclosed function call on
+  agent loop early, which would leave an unclosed function call on
   the server for models that manage conversation state
   server-side. Combining delegated tools with stateful models
   is not supported.
--- a/site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
+++ b/site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
@@ -341,7 +341,7 @@ response directly to the caller, skipping the extra orchestrator model round-tri
 Use delegation when the sub-agent's response is already complete and the
 orchestrator adds no value by processing it further:
 
-- A coding agent returning generated source that shouldn't be rephrased
+- A coding agent returning generated source code that shouldn't be rephrased
 - A customer service agent whose compliance-reviewed language must reach
   the user unchanged
 - A retrieval agent returning structured data the caller consumes directly
@@ -380,39 +380,21 @@ result = orchestrator("Why was I charged twice?")
 </Tabs>
 
 On success, the agent loop stops with
-<Syntax py="stop_reason == 'end_turn'" ts="stopReason === 'endTurn'" />
-and the `AgentResult` contains the delegate's content without additional
-model processing. Structured JSON output is serialized to text.
-Streaming events from the delegate surface natively in the parent's
-stream, so callers see the specialist's tokens as they arrive.
+<Syntax py="stop_reason == 'end_turn'" ts="stopReason === 'endTurn'" /> and
+the `AgentResult` contains the delegate's content without additional
+model processing. Streaming events from the delegate surface natively in
+the parent's stream, so callers see the specialist's tokens as they arrive.
 
-If the delegate throws, the error surfaces as a failed tool result and
-the orchestrator can attempt recovery on the next turn.
-
-Delegated and non-delegated tools can coexist on the same
-orchestrator. But when the model calls a delegated tool agent, it must
-be the only tool call in that turn. The SDK communicates this to
-the model by appending an instruction to the tool agent's
-description and enforces it at runtime by cancelling every tool in
-the batch if other tools are requested alongside it.
-
-Additional limitations to be aware of:
-
-- **Incompatible with stateful models**: Delegation exits the loop early,
-  which would leave an unclosed function call on the server for models that
-  manage conversation state server-side. The SDK throws at agent
-  initialization if you register a delegation tool on a stateful model.
-  Delegation tools added later silently behave as ordinary tools (in TypeScript)
-  or fail with an error tool result (in Python).
-- **Delegation does not bubble up**: If a delegate is itself an orchestrator
-  that delegates, that only ends its own loop. The agent one level up
-  that invoked it as a non-delegated tool still receives a tool
-  result and continues its loop normally.
-- **Session persistence timing** *(TypeScript)*: The delegation content message
-  is written to `agent.messages` after the loop exits. Session managers
-  configured with `saveLatestOn: 'message'` may persist the `endTurn`
-  placeholder rather than the final delegation content. The default
-  `saveLatestOn: 'invocation'` is unaffected.
+Limitations to be aware of:
+- **Single delegated tool per turn**: Delegated and non-delegated tools
+  can exist on the same agent. However, a delegated tool must be
+  the only tool call in its turn. The SDK cancels the batch if
+  other tools are requested alongside it.
+- **Incompatible with stateful models**: Delegation exits the
+  loop early, which would leave an unclosed function call on
+  the server for models that manage conversation state
+  server-side. Combining delegated tools with stateful models
+  is not supported.
 
 ## Remote Agents with A2A
 
--- a/site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
+++ b/site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
@@ -7,6 +7,7 @@ tags: [multi-agent, tool-authoring, tool-execution]
 sourceLinks:
   - path: strands-py/src/strands/agent/_agent_as_tool.py
   - path: strands-ts/src/agent/agent-as-tool.ts
+  - path: strands-ts/src/agent/agent-delegation.ts
 ---
 
 ## The Concept: Agents as Tools
@@ -332,6 +333,86 @@ response = orchestrator(customer_query)
 </Tabs>
 
 This example demonstrates how Strands Agents SDK enables specialized experts to collaborate on complex queries requiring multiple domains of knowledge. The orchestrator intelligently routes different aspects of the query to the appropriate specialized agents, then synthesizes their responses into a comprehensive answer.
+
+## Delegation
+
+When you mark a tool agent as a delegate, the orchestrator returns its
+response directly to the caller, skipping the extra orchestrator model round-trip.
+Use delegation when the sub-agent's response is already complete and the
+orchestrator adds no value by processing it further:
+
+- A coding agent returning generated source that shouldn't be rephrased
+- A customer service agent whose compliance-reviewed language must reach
+  the user unchanged
+- A retrieval agent returning structured data the caller consumes directly
+
+To mark a tool agent as a delegate, pass `delegate: true` to
+<Syntax py=".as_tool()" ts=".asTool()" />:
+
+<Tabs>
+<Tab label="Python">
+
+```python
+from strands import Agent
+
+billing_agent = Agent(
+    name="billing_expert",
+    description="Answers billing questions: charges, refunds, and invoices.",
+    system_prompt="You handle billing questions with precision.",
+)
+
+orchestrator = Agent(
+    system_prompt="""Route billing questions to billing_expert.
+Answer general questions yourself.""",
+    tools=[billing_agent.as_tool(delegate=True)],
+)
+
+result = orchestrator("Why was I charged twice?")
+# result contains the billing agent's response directly
+```
+</Tab>
+<Tab label="TypeScript">
+
+```typescript
+--8<-- "user-guide/concepts/multi-agent/agents-as-tools.ts:delegation_basic"
+```
+</Tab>
+</Tabs>
+
+On success, the agent loop stops with
+<Syntax py="stop_reason == 'end_turn'" ts="stopReason === 'endTurn'" />
+and the `AgentResult` contains the delegate's content without additional
+model processing. Structured JSON output is serialized to text.
+Streaming events from the delegate surface natively in the parent's
+stream, so callers see the specialist's tokens as they arrive.
+
+If the delegate throws, the error surfaces as a failed tool result and
+the orchestrator can attempt recovery on the next turn.
+
+Delegated and non-delegated tools can coexist on the same
+orchestrator. But when the model calls a delegated tool agent, it must
+be the only tool call in that turn. The SDK communicates this to
+the model by appending an instruction to the tool agent's
+description and enforces it at runtime by cancelling every tool in
+the batch if other tools are requested alongside it.
+
+Additional limitations to be aware of:
+
+- **Incompatible with stateful models**: Delegation exits the loop early,
+  which would leave an unclosed function call on the server for models that
+  manage conversation state server-side. The SDK throws at agent
+  initialization if you register a delegation tool on a stateful model.
+  Delegation tools added later silently behave as ordinary tools (in TypeScript)
+  or fail with an error tool result (in Python).
+- **Delegation does not bubble up**: If a delegate is itself an orchestrator
+  that delegates, that only ends its own loop. The agent one level up
+  that invoked it as a non-delegated tool still receives a tool
+  result and continues its loop normally.
+- **Session persistence timing** *(TypeScript)*: The delegation content message
+  is written to `agent.messages` after the loop exits. Session managers
+  configured with `saveLatestOn: 'message'` may persist the `endTurn`
+  placeholder rather than the final delegation content. The default
+  `saveLatestOn: 'invocation'` is unaffected.
 
 ## Remote Agents with A2A
 
--- a/site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
+++ b/site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
@@ -6,6 +6,7 @@ sidebar:
 tags: [multi-agent, tool-authoring, tool-execution]
 sourceLinks:
   - path: strands-py/src/strands/agent/_agent_as_tool.py
+  - path: strands-py/src/strands/agent/_agent_delegation.py
   - path: strands-ts/src/agent/agent-as-tool.ts
   - path: strands-ts/src/agent/agent-delegation.ts
 ---
--- a/site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.ts
+++ b/site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.ts
@@ -161,10 +161,16 @@ void orchestratorExample
 
 async function delegationBasicExample() {
   // --8<-- [start:delegation_basic]
+  const BillingResponseSchema = z.object({
+    summary: z.string(),
+    refundAmount: z.number(),
+  })
+
   const billingAgent = new Agent({
     name: 'billing_expert',
     description: 'Answers billing questions: charges, refunds, and invoices.',
     systemPrompt: 'You handle billing questions with precision.',
+    structuredOutputSchema: BillingResponseSchema,
     printer: false,
   })
 
@@ -175,7 +181,7 @@ Answer general questions yourself.`,
   })
 
   const result = await orchestrator.invoke('Why was I charged twice?')
-  // result contains the billing agent's response directly
+  // result contains the billing agent's structured JSON response, unchanged
   // --8<-- [end:delegation_basic]
   void result
 }
--- a/site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.ts
+++ b/site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.ts
@@ -158,3 +158,25 @@ Always select the most appropriate tool based on the user's query.`,
 }
 
 void orchestratorExample
+
+async function delegationBasicExample() {
+  // --8<-- [start:delegation_basic]
+  const billingAgent = new Agent({
+    name: 'billing_expert',
+    description: 'Answers billing questions: charges, refunds, and invoices.',
+    systemPrompt: 'You handle billing questions with precision.',
+    printer: false,
+  })
+
+  const orchestrator = new Agent({
+    systemPrompt: `Route billing questions to billing_expert.
+Answer general questions yourself.`,
+    tools: [billingAgent.asTool({ delegate: true })],
+  })
+
+  const result = await orchestrator.invoke('Why was I charged twice?')
+  // result contains the billing agent's response directly
+  // --8<-- [end:delegation_basic]
+  void result
+}
+void delegationBasicExample


liramon2:
Update 7/28:
- Added initial docs
- Need to merge the following first
https://github.com/strands-agents/harness-sdk/pull/3508

liramon2:
@strandly-the-agent Review this PR.

site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
```diff
@@ -333,6 +334,77 @@ response = orchestrator(customer_query)
...
+
+billing_agent = Agent(
+    name="billing_expert",
+    system_prompt="You handle billing questions with precision.",
+)
+
+orchestrator = Agent(
+    system_prompt="""Route billing questions to billing_expert.
+Answer general questions yourself.""",
+    tools=[billing_agent.as_tool(delegate=True)],
```
strandly-the-agent, line 376: 🔴 **This example can't run — Python has no `delegate` parameter.**

```
$ python3 -c "from strands import Agent; Agent(name='b', system_prompt='x').as_tool(delegate=True)"
TypeError: Agent.as_tool() got an unexpected keyword argument 'delegate'
```

`strands-py/src/strands/agent/agent.py:957-963` takes keyword-only `name`, `description`, `preserve_context` — there's no `delegate`, and no delegation module anywhere in `strands-py`. The port is #3479 (open), whose body can't even agree with itself on the name (prose says `delegate`, its sample says `handoff=True`). `.agents/references/code-verification.md` covers this case directly: verify each language independently, and omit the block rather than shipping an unverified example.

It matters more than a normal broken snippet because this page renders to markdown for model consumption — an agent asked to "add delegation to my Python orchestrator" will reproduce this exact call with confidence.

The repo already has two patterns for a single-language feature: the section-level `:::note[Python only]` in `swarm.mdx:338-340`, and the stub fence in the unsupported tab at `hooks.mdx:960` (`// This feature is not yet available in TypeScript SDK`). Either works; dropping the Python `<Tab>` (lines 352-371) plus a note reads best for a whole new section:

````mdx
To mark a tool agent as a delegate, pass `delegate: true` to `.asTool()`:

:::note[TypeScript only]
Delegation is not yet available in the Python SDK — see
[#3479](https://github.com/strands-agents/harness-sdk/issues/3479).
:::
````

One gotcha if you go that route: `<Syntax>` at `:349` requires **both** `py` and `ts` props (`site/src/components/Syntax.astro:4-6`), so a TS-only `<Syntax ts="..." />` won't typecheck — plain inline `` `.asTool()` `` is the way, and it also brings that 103-char line under the 90-char guideline.

No suggestion block on this one since it contains a fenced snippet.

File: site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx [hunk 2]
```diff
@@ -333,6 +334,77 @@ response = orchestrator(customer_query)
...
+the model by appending an instruction to the tool agent's
+description, and enforces it at runtime by cancelling the batch if
+other tools are requested alongside it.
+
+Additional limitations to be aware of:
+
+- **Incompatible with stateful models.** Delegation exits the loop early,
+  which would leave an unclosed function call on the server for models that
+  manage conversation state server-side. The SDK throws at agent
+  initialization if you register a delegation tool on a stateful model.
```
strandly-the-agent: 🟡 True, but a reader will over-trust it. The throw only covers tools present when `initAgent()` runs (`strands-ts/src/agent/agent-delegation.ts:103-112`, which inspects `agent.toolRegistry.list()` at that moment). Every per-invocation hook then bails unconditionally on `agent.model.stateful` (`:155`, `:176`, `:210`, `:247`) regardless of *when* the tool arrived — so a delegation tool registered on a stateful-model agent after `initialize()` neither throws nor gets delegation semantics; it silently runs as an ordinary tool. The SDK's own tests pin that behaviour: `agent-delegation.test.ts:308` ("skips delegation logic for delegation tools added after init — tool runs normally") and `:350` ("does not enforce single-call constraint for stateful models with runtime-added tools").

Worth one sentence so nobody reads the guardrail as unconditional:

```suggestion
- **Incompatible with stateful models.** Delegation exits the loop early,
  which would leave an unclosed function call on the server for models that
  manage conversation state server-side. The SDK throws at agent
  initialization if you register a delegation tool on a stateful model.
  This check only covers tools present at construction or initialization:
  a delegation tool added to a stateful-model agent at runtime is silently
  treated as an ordinary tool, with no error and no delegation semantics.
```

site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
```diff
@@ -333,6 +334,77 @@ response = orchestrator(customer_query)
...
+
+If the delegate throws, the error surfaces as a failed tool result and
+the orchestrator can attempt recovery on the next turn.
+
+Delegated and non-delegated tools can coexist on the same
+orchestrator. But when the model calls a delegated tool agent, it must
+be the only tool call in that turn. The SDK communicates this to
+the model by appending an instruction to the tool agent's
+description, and enforces it at runtime by cancelling the batch if
+other tools are requested alongside it.
```
strandly-the-agent: 🟡 "Cancelling the batch" is accurate but hides the part someone debugging a transcript needs. `event.cancel` (`agent-delegation.ts:162-169`) flows into `_cancelAllAsResults` (`agent.ts:2229,2263`), which stamps **every** tool call in that turn — including unrelated ones — with the same text: *"This tool call was not executed. A delegation tool must be the only tool called in a turn. Retry with a single delegation tool call or use only non-delegation tools."* Someone whose `get_weather` call comes back complaining about delegation tools, not realising one of their other tools is a delegate, has a genuinely confusing few minutes. Quoting the message also makes it greppable:

```suggestion
orchestrator. But when the model calls a delegated tool agent, it must
be the only tool call in that turn. The SDK communicates this to the model
by appending an instruction to the tool agent's description, and enforces
it at runtime by cancelling the whole batch: every tool call in that turn,
including unrelated ones, comes back with the same error ("A delegation
tool must be the only tool called in a turn...").
```

The model does get that text back as a tool error and can retry with a single call, which is worth a half-sentence too — it's the same recovery story as the throw case two paragraphs up.

site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
```diff
@@ -333,6 +334,77 @@ response = orchestrator(customer_query)
...
+
+- **Incompatible with stateful models.** Delegation exits the loop early,
+  which would leave an unclosed function call on the server for models that
+  manage conversation state server-side. The SDK throws at agent
+  initialization if you register a delegation tool on a stateful model.
+- **Session persistence timing.** The delegation content message is
+  written to `agent.messages` after the loop exits. Session managers
+  configured with `saveLatestOn: 'message'` may persist the `endTurn`
+  placeholder rather than the final delegation content. The default
+  `saveLatestOn: 'invocation'` is unaffected.
```
strandly-the-agent: 🟡 Both linked issues headline "Hierarchical delegation: Manager → Senior → Junior", but delegation is scoped to the immediate caller only: a middle agent delegating to a leaf ends *its* loop, while a parent that called the middle agent as an ordinary tool just gets a tool result and keeps going (`agent-delegation.test.ts:576-577`, `describe('nested-delegation')` — "child delegation does not leak to the parent agent"). Combined with the one-delegated-call-per-turn rule above, multi-level composition is easy to get wrong from these docs. A worked example is overkill for this PR; a bullet that heads off the wrong mental model isn't:

```suggestion
- **Session persistence timing.** The delegation content message is
  written to `agent.messages` after the loop exits. Session managers
  configured with `saveLatestOn: 'message'` may persist the `endTurn`
  placeholder rather than the final delegation content. The default
  `saveLatestOn: 'invocation'` is unaffected.
- **Delegation does not bubble up.** If a delegate is itself an orchestrator
  that delegates, that only ends its own loop. A caller one level further up
  that invoked it as an ordinary (non-delegated) tool still receives a tool
  result and continues its loop normally.
```

site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.ts
```diff
@@ -158,3 +158,24 @@ Always select the most appropriate tool based on the user's query.`,
 }
 
 void orchestratorExample
+
+async function delegationBasicExample() {
+  // --8<-- [start:delegation_basic]
+  const billingAgent = new Agent({
+    name: 'billing_expert',
+    systemPrompt: 'You handle billing questions with precision.',
+    printer: false,
```
strandly-the-agent, line 174: 🟡 The example omits `description`, which is the one knob delegation makes critical. With it absent, `agent-as-tool.ts:145-148` falls back to `` `Use the ${this.name} agent by providing a natural language input` `` (then appends the delegation suffix at `:150-152`) — so the model's only real routing signal is the orchestrator prompt happening to name the tool. Under plain agents-as-tools a misroute gets a second look from the orchestrator; under delegation the wrong specialist's answer goes straight to the caller. It also undercuts this page's own advice at `agents-as-tools.mdx:35` ("Write descriptive names and descriptions that explain the agent's expertise").

```suggestion
  const billingAgent = new Agent({
    name: 'billing_expert',
    description: 'Answers billing questions: charges, refunds, and invoices.',
    systemPrompt: 'You handle billing questions with precision.',
    printer: false,
```

And one sentence in the prose would carry the point: *"Because delegation skips the orchestrator's review step, the sub-agent's `description` carries the full weight of routing — make it specific enough for the model to route correctly on the first call."*

strandly-the-agent (changes requested):
**Verdict: changes requested — one blocker.** The TypeScript half is accurate (I checked every claim against `agent-delegation.ts` and the snippet compiles); the Python tab documents an API that doesn't exist yet.

🔴 `billing_agent.as_tool(delegate=True)` raises `TypeError` on today's `strands-py` — I ran it. Delegation is TS-only; the Python port (#3479) is still open, and its parameter name isn't settled in code either. Four 🟡 accuracy/DevX items inline, nits collapsed.

No API-review label needed: this touches only `site/`, no public SDK surface.

<details><summary>✅ What I actually verified (head <code>f9d5713</code> vs main <code>5f357ae</code>)</summary>

| Check | Result |
|---|---|
| `npm run cms:build` (site) | ✅ 864 pages, no broken links |
| `npm run typecheck` (site) | ✅ clean |
| `npm run typecheck:snippets` (after building `strands-ts`) | ✅ clean — confirms `.asTool({ delegate: true })` and `.invoke()` compile against the real API |
| `npm test` (site) | ✅ 473 passed / 2 skipped |
| `npx prettier --check` on `agents-as-tools.ts` | ✅ passes (the `.mdx` is prettier-ignored via `site/.prettierignore`, so nothing lints it) |
| Python example executed | 🔴 `TypeError: Agent.as_tool() got an unexpected keyword argument 'delegate'` |

Prose claims re-checked against the implementation and **all accurate**: `stopReason: 'endTurn'` + `AgentResult` content (`strands-ts/src/agent/agent-delegation.ts:360-369`), JSON→text serialization (`agent-as-tool.ts:212-219`, `agent-delegation.ts:56-63`), native stream passthrough (`agent-as-tool.ts:198-209`, `agent-delegation.ts:268-290`), error → failed tool result then recovery (`agent-delegation.ts:190-193`), the single-call constraint and description suffix (`agent-as-tool.ts:21-22,150-152`), and the `saveLatestOn: 'message'` caveat (`session/session-manager.ts:47,118` — the replacement message is yielded outside `invokeCallbacks` at `agent-delegation.ts:349,354`, exactly as you describe). The added `sourceLinks` entry resolves to a real file. `printer: false` and the bare `invoke()` with no extraction both match this snippet file's existing conventions (`agents-as-tools.ts:13,21,150-157`) — deliberately not flagged.
</details>

<details><summary>❓ Questions (2)</summary>

1. **Split or hold?** Your 7/28 note says this was blocked on #3508 — that merged on 7/30, but it only adds the Python `BeforeToolsEvent`/`AfterToolsEvent` batch hooks, not `as_tool(delegate=...)`. The actual blocker for the Python tab looks like #3479, which is untouched. Is the plan now *land TS docs, stub the Python tab, follow up when #3479 merges*, or hold the whole PR as draft until then? Either works — it just changes which fix belongs in comment 1.
2. **Terminology, non-blocking:** `.agents/references/terminology.md:23` lists "agent delegation" as a synonym to avoid in favour of the canonical "agents as tools". Naming the *opt-in* "delegation" matches the SDK's own `AgentDelegation`/`delegate`, so I don't read this as a violation — but worth a nod from whoever owns that table, especially since Swarm already uses "handoff" for a different mechanism (`swarm.mdx`, `streaming/index.mdx:127-131`) and #3479 is literally titled "Agent-as-tools handoff". A one-line disambiguation on this page would save readers who've seen both words.
</details>

- `agents-as-tools.mdx:349` is 103 chars vs the 90-char guideline in `.agents/references/mdx-authoring.md`, so the PR-body checkbox "No lines exceed 90 characters in new content" is technically false. Low stakes — pre-existing prose on this page runs to 321 and 346 chars (`:335`, `:410`), and nothing enforces it. The comment-1 rewrite happens to fix it.
- `agents-as-tools.mdx:382` has a trailing space.
- The literal description suffix the model sees isn't quoted anywhere: `" Calling this tool will return its response directly to the user as the final answer. It should be the only tool called in the turn."` (`agent-as-tool.ts:21-22`). A collapsed callout would let a reader confirm what the model got without reading source. Optional.
- **Pre-existing, not this PR's bug, but worth knowing:** a delegate whose response is image/document-only reaches the caller as *empty text*, not just "serialized". The tool result is built from `result.toString()` (`agent-as-tool.ts:216`), and `AgentResult.toString()` handles only text/reasoning/citations — anything else hits `default: console.debug(...)` and is dropped (`types/agent.ts:547-566`). The SDK knows: `agent-delegation.test.ts:272-273` covers "empty text (e.g. image-only response)" but only asserts that delegation still triggers, not that content survives. Since `:382` is the first place this page talks about output-content handling, one sentence there would help. I didn't file an issue for the underlying silent drop since it looks like accepted behaviour — say the word and I will.
</details>

<details><summary>How this review ran</summary>

Docs-accuracy, issue-alignment and LLM-context passes as independent fresh-context subagents over a shared context pack, then de-duped and severity-ranked (17 raw findings → 5 posted). Triage's own verdict was *defer* — premature docs — which is really comment 1 in another form. I corrected two things the passes got wrong: the `strands-py` `sourceLinks` entry at `:8` is pre-existing rather than added here, and `<Syntax>` requires **both** `py` and `ts` props (`site/src/components/Syntax.astro:4-6`), so a TS-only `<Syntax>` wouldn't typecheck. Review is a helper pass, not a gate — a human maintainer should approve.
</details>

liramon2:
@strandly-the-agent Review this PR again. Consolidate your findings into a single comment. Note that this will only be merged after #3479 is merged.

strandly-the-agent:
## Re-review of `8ac69c5` — consolidated

**4 of 5 findings fixed, and I'm withdrawing the 🔴 blocker** — with the merge-after-#3479 commitment it's a dependency, not a defect. I checked #3346 (head `3ded945`) and it really does ship `as_tool(delegate=True)` returning `stop_reason='end_turn'`, so the Python tab matches the API that will land.

**One new 🔴, and it's my fault as much as yours:** the sentence you added at `:404` for my stateful-model finding is correct for TypeScript and *wrong for Python*. Same root cause for the two 🟡 below — this section is dual-language but was written from the TS implementation, and #3346 deliberately diverges in places.

| My previous finding | Status |
|---|---|
| 🔴 Python tab documents a non-existent API | **Withdrawn** — merge order + #3346 verified (details below) |
| 🟡 Stateful guardrail is init-time only (`:400-404`) | **Fixed for TS, introduced a Python inaccuracy** — see 🔴 below |
| 🟡 "Cancelling the batch" hides that all tools are cancelled (`:395-396`) | **Fixed** — "cancelling every tool in the batch"; same text and behaviour in both SDKs |
| 🟡 No nested/hierarchical caveat (`:410-413`) | **Fixed** — accurate for TS; untested in Python (one line below) |
| 🟡 Example omits `description` (`:360`, `agents-as-tools.ts:166`) | **Fixed** — both tabs |

Nits from last round (trailing space, the 103-char line) are fixed too, and `Docs / CI` is green on this head.

### New findings

1. 🔴 `:404` — "Delegation tools added later silently behave as ordinary tools" is true in TS, **false in Python**: #3346 returns an explicit *error* tool result.
2. 🟡 `:405-409` — the session-persistence caveat is a TypeScript-only problem presented as general; Python's plugin flushes the corrected message immediately under every strategy, and the option is `save_latest_on`, not `saveLatestOn`.
3. 🟡 `:382` — `stopReason: 'endTurn'` is the TS spelling in a dual-language section; Python is `stop_reason == 'end_turn'`. This page's `<Syntax>` component exists for exactly this.

<details><summary>🔴 1 — evidence and suggested rewrite (<code>:400-404</code>)</summary>

In TS the stateful check is folded into the same pass-through condition as "not a delegation tool" (`strands-ts/src/agent/agent-delegation.ts:246-249`), so a delegate added after the init check just executes normally — genuinely silent.

In Python it's a **separate branch that fails the call** (`strands-py/src/strands/agent/_agent_delegation.py:260-277` at `3ded945`):

```python
# Reject runtime-added delegate on stateful model.
if context.agent.model.stateful:
    yield ToolResultEvent({... "status": "error", "content": [{"text": (
        "Delegation failed: this tool is not supported with stateful models. "
        "Respond to the user directly instead of delegating.")}]})
    return
```

Pinned by `tests/strands/agent/test_agent_delegation.py::test_runtime_stateful_delegate_returns_error`, which asserts `status == "error"`.

Second, smaller divergence in the same bullet: *when* the throw happens. Python raises from the constructor — `init_agent` (`_agent_delegation.py:129-142`) is called synchronously via `strands-py/src/strands/agent/agent.py:519`. TS raises from the lazy `initialize()` (`strands-ts/src/agent/agent.ts:764`, awaited on first run at `:1071`), which is why its own test asserts `await expect(orchestrator.initialize()).rejects.toThrow(...)` (`agent-delegation.test.ts:448-467`) rather than the constructor. "At agent initialization" covers both loosely, but a Python reader will expect `Agent(...)` to throw in TS too, and it won't.

One bullet that's true in both languages:

```mdx
- **Incompatible with stateful models.** Delegation exits the loop early,
  which would leave an unclosed function call on the server for models that
  manage conversation state server-side. Registering a delegation tool on
  a stateful model raises at agent initialization. A delegation tool added
  after that check is silently treated as an ordinary tool in TypeScript;
  in Python it fails at call time with an error tool result telling the
  model to answer directly.
```
</details>

<details><summary>🟡 2 — the session caveat doesn't apply to Python (<code>:405-409</code>)</summary>

Python's delegation plugin hands the corrected message straight to the session manager (`_agent_delegation.py:366-374`):

```python
if session_manager is not None:
    session_manager.redact_latest_message(delegation_message, context.agent)
else:
    await context.agent.hooks.invoke_callbacks_async(MessageAddedEvent(...))
```

and `SnapshotSessionManager.redact_latest_message` flushes unconditionally — its own docstring says so, and names the divergence (`strands-py/src/strands/session/snapshot_session_manager.py:319-328`): *"This flush happens regardless of `save_latest_on` (including `"trigger"`)… This diverges from the TypeScript SDK."* TS delegation never touches the session manager at all (`grep -n 'sessionManager\|redactLatest' strands-ts/src/agent/agent-delegation.ts` → nothing), which is exactly why the placeholder can persist there. So a Python reader is currently warned off a problem they cannot have — and the option name they'd search for is `save_latest_on` (`snapshot_session_manager.py:52,210`), not `saveLatestOn`.

While you're in this bullet: the thing a Python hook author *does* need is the `MessageAddedEvent` asymmetry, which #3346's own module docstring spells out (`_agent_delegation.py:10-13`) — subscribers see the placeholder plus a second event with the real content when there's no session manager, and no event for the real content when there is one. In TS the hook never fires for the corrected message either way. If the docs discuss this area at all, that's the load-bearing part.

```mdx
- **Session persistence timing (TypeScript).** The delegation content message
  is written to `agent.messages` after the loop exits without firing the
  `MessageAddedEvent` hook, so a session manager configured with
  `saveLatestOn: 'message'` may persist the `endTurn` placeholder rather
  than the final delegation content. The default `saveLatestOn: 'invocation'`
  is unaffected. Python persists the corrected content immediately under
  every `save_latest_on` strategy.
```
</details>

<details><summary>🟡 3 — TS-only identifier spelling (<code>:382</code>)</summary>

Python is `stop_reason` with value `"end_turn"` (`strands-py/src/strands/agent/agent_result.py:35`, and #3346's own description says `result.stop_reason == 'end_turn'`).

```mdx
On success, the agent loop stops with
<Syntax py="stop_reason == 'end_turn'" ts="stopReason === 'endTurn'" /> and
the `AgentResult` contains the delegate's content without additional
```

Both lines stay under the 90-char guideline.
</details>

<details><summary>Two questions — both really about #3346, not these docs</summary>

1. **A guard Python has and TS appears to lack.** `_agent_delegation.py:231-241` skips delegation entirely when the parent agent has a structured-output tool registered (`# Skip delegation when the parent expects structured output`), tested at `test_agent_delegation.py::test_on_after_tools_suppressed_with_structured_output`. `grep -n structuredOutput strands-ts/src/agent/agent-delegation.ts` returns nothing, and TS's `_onAfterTools` sets `endTurn` unconditionally. **I have not reproduced a TS failure**, so this is a question, not a finding: is the Python guard covering a real hole that TS also has, or does TS avoid it some other way? Either answer matters here, because if the two SDKs behave differently when an orchestrator has both a structured-output schema and a delegate, this dual-language section will need to say so.
2. **Nested delegation is documented for both but tested only in TS.** TS pins it (`agent-delegation.test.ts:576-599`, "child delegation does not leak to the parent agent"); #3346's 36 tests have no nested/multi-level case (no `Middle`/`Leaf`/nested scenario in `test_agent_delegation.py`). The per-agent `WeakKeyDictionary` state and per-agent plugin instance make me expect it behaves the same, but the new `:410-413` bullet currently rests on code reading for Python. Mirroring that TS test in #3346 would make the doc claim true by construction.
</details>

<details><summary>What I verified this round</summary>

- Reviewed `8ac69c5` (docs) against `origin/main` for TypeScript and against **#3346 head `3ded945`** for Python, in a separate worktree. If #3346 changes during review — it carries `api/needs-review` — the Python tab and these bullets need re-checking against whatever lands.
- ✅ Every one of the five previous findings re-checked against the new diff (statuses in the table above). I only reviewed the new commit; resolved threads aren't reopened here.
- ✅ `delegate=True` and `stop_reason='end_turn'` confirmed present in #3346 — this is what withdraws the blocker.
- ✅ Description suffix text is byte-identical across SDKs (`_agent_as_tool.py:27-30` vs `agent-as-tool.ts:21-22`), and so is the batch-cancel message, so your `:391-396` rewrite holds for both.
- ✅ `description?: string` is valid on the TS Agent config (`strands-ts/src/agent/agent.ts:284`), and no line in the new section exceeds 90 chars.
- ✅ Validation: I'm citing **this PR's own `Docs / CI` run on `8ac69c5` (success)** — it runs `cms:build`, `typecheck`, the `strands-ts` build, `typecheck:snippets` and the site tests. My local re-run of that sequence died on a sandbox timeout before producing output, so I'm not reporting local numbers this round. `check-api-review-label` also passes; no API-review label needed for a `site/`-only change.
- Shape of this review: I ran the delta myself since it only touched my findings, plus one fresh-context docs-accuracy pass dedicated to checking the section against the Python implementation. My earlier `CHANGES_REQUESTED` stands on finding 1 — a factually wrong sentence for one of the two languages this section documents — but everything left is wording, and a human maintainer should be the one to approve.
</details>

liramon2:
Update 8/11:
- Currently refining with Strandly
- #3508 is merged (prerequisite 1)
- #3346 needs to be merged first (prerequisite 2)

liramon2:
@strandly-the-agent Review this PR again. Consolidate your findings into a single comment.

strandly-the-agent:
## Re-review of `6adbdff` — consolidated

All three findings from my last pass are fixed. **No blockers left** — two copy nits below and one optional item; nothing that needs another review round.

| Last round's finding | Status |
|---|---|
| 🔴 `:404` stateful sentence was wrong for Python | **Fixed** — split by language, and the Python half matches #3346's error-tool-result behaviour |
| 🟡 `:405-409` session caveat presented as general | **Fixed** — scoped to TypeScript, and moving it last reads better than where it was |
| 🟡 `:382` TS-only `stopReason` spelling | **Fixed** — `<Syntax>` with both required props |

**Two copy nits, take or leave:**

1. `:405-406` — subject/verb: "Delegation tools added later silently behave as ordinary tools (in TypeScript) or **fails** with an error tool result (in Python)" → `or fail with an error tool result`.
2. `:411` — the bold title reads `**Session persistence timing. (TypeScript)**`, with the period mid-title and the marker inside the bold, unlike its sibling bullets. The page-family convention for this marker is italic and outside (`session-management.mdx:275,298` — `*(TypeScript only)*`), so: `- **Session persistence timing** *(TypeScript only)*. The delegation content message…`

**One thing still open, your call.** The smaller half of my last 🔴 wasn't picked up, and it's genuinely minor: the throw *timing* differs between SDKs. Python raises from `Agent(...)` construction (`_agent_delegation.py:129-142` via `strands-py/src/strands/agent/agent.py:519`), TypeScript from the lazy `initialize()` on first run (`strands-ts/src/agent/agent.ts:764`, awaited at `:1071`). `:403-404`'s "throws at agent initialization" is loosely true of both, so leaving it is defensible.

**On my standing review:** the `CHANGES_REQUESTED` from 8/10 is stale — its blocking item is fixed. I'm deliberately not converting it into an approval: the Python half of this section is verified against #3346, which is unmerged and still carries `api/needs-review`. If that API shifts during its review, the Python tab (`:355-372`) and the two language-split bullets need re-checking against whatever actually lands. Worth a human approve at merge time, once #3346 is in.

<details><summary>What I checked</summary>

- Reviewed `6adbdff` against `8ac69c5` myself rather than re-running the full pipeline: the delta is prose-only (10 insertions / 8 deletions in one file) and touches only my own findings. #3346 is unchanged at `3ded945`, so last round's Python-side verification still holds — I re-read the delta rather than re-deriving those facts.
- ✅ No line in the section exceeds the 90-char guideline; no trailing whitespace; `<Syntax>` at `:383` supplies both `py` and `ts` (it errors without both) and sits inside the paragraph with no blank line around it, the same shape as `:349-350` that already passed `Docs / CI` on the previous head.
- ✅ The Python spelling in the new `<Syntax>` matches #3346: `stop_reason == 'end_turn'`.
- ⚠️ CI: `Docs / CI` is **queued, not yet green** on `6adbdff` at the time of writing — I'm not claiming a pass. It was green on `8ac69c5`, and this delta adds no new component or link, but the run is the authority; worth a glance before you undraft.
- Not re-raised: the items you already resolved, and last round's optional suggestions you chose not to take.
</details>

liramon2:
@strandly-the-agent Review this PR again. Consolidate your findings into a single comment. Note that the edited site docs are a conceptual overview.

strandly-the-agent (approved):
## Re-review of `516bbd8` — consolidated

**Approving** — this also clears my stale 8/10 changes-requested. With #3346 merged (`api/review-complete`), every claim left in the slimmed section is true of code on `main` for **both** languages. I re-verified each claim rather than assuming, including running the Python example's call against merged main:

```
>>> Agent(name='billing_expert', ...).as_tool(delegate=True)
as_tool(delegate=True) OK -> _AgentAsTool | delegate = True
```

The call that raised `TypeError` three rounds ago now executes against shipped code.

**Previous findings:** both copy nits from last round were resolved by the rewrite (those sentences no longer exist), and the optional init-timing item is moot — the conceptual overview no longer describes throw timing at all. One thing worth naming: #3346's final commits changed Python's runtime-stateful behaviour to silently run delegates as ordinary tools, converging with TypeScript — which means the language-split sentence the rewrite dropped is precisely the sentence that would have just gone stale. The conceptual scope removed it before it could mislead anyone.

**One ⚪ nit and one optional pointer, neither blocking:**

- `sourceLinks` frontmatter (`agents-as-tools.mdx:7-10`) lists the delegation module for TS only; `strands-py/src/strands/agent/_agent_delegation.py` exists on `main` now, so adding it would match the `_agent_as_tool.py` / `agent-as-tool.ts` pairing above it.
- The dropped TS-only session gotcha (`saveLatestOn: 'message'` can persist the `endTurn` placeholder) now lives nowhere in the docs. That's a defensible scoping call for a conceptual overview — if it ever bites users, the TS-only sections of `session-management.mdx` are its natural home. Not this PR's problem.

<details><summary>✅ What I verified on this head</summary>

- **Ground truth moved, so I re-checked every surviving claim against merged `main`**, not against my earlier #3346-branch verification:
  - `as_tool(..., delegate: bool = False)` shipped: `strands-py/src/strands/agent/agent.py:1011-1018`, plus the live construct check above (`pip install -e strands-py` at `main`).
  - `stop_reason == 'end_turn'` / `stopReason === 'endTurn'` `<Syntax>` — correct spellings both sides.
  - Native streaming: Python unwraps sub-agent events as native (`_agent_delegation.py:290-294` on `main`); TS unchanged.
  - Single-delegate-per-turn + batch cancel: same mechanism and identical cancel text both languages (`_agent_delegation.py` `_on_before_tools` on `main`; `agent-delegation.ts:162-169`).
  - Stateful bullet: init-time throw retained in both; runtime-added delegates now run as ordinary tools in **both** (Python `main` `_agent_delegation.py:260-263` — "stateful model, skipping delegation and running as a normal tool"; TS `agent-delegation.ts:246-249`), so "not supported" is accurate at the level this section speaks.
- Nothing the rewrite *kept* overclaims; everything it *dropped* (JSON→text, error recovery, description-suffix mechanics, session timing) was accurate-but-detail, so the removals introduce no error.
- Hygiene: no line over 90 chars, no trailing whitespace; the `.ts` snippet is unchanged from the already-validated version.
- CI on `516bbd8`: `Docs / CI` **SUCCESS**, `check-api-review-label` SUCCESS (docs-only, no API surface); preview deploy still in flight at time of writing.
- Shape: self-run follow-up — the delta is removals plus rewording of content reviewed across three rounds; what was genuinely new (merged #3346 as the final ground truth) is exactly what I re-verified.

Still a draft — undraft when you're ready; nothing outstanding from my side.
</details>

site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
```diff
@@ -333,6 +335,68 @@ response = orchestrator(customer_query)
 
 This example demonstrates how Strands Agents SDK enables specialized experts to collaborate on complex queries requiring multiple domains of knowledge. The orchestrator intelligently routes different aspects of the query to the appropriate specialized agents, then synthesizes their responses into a comprehensive answer.
 
+## Delegation
+
+When you mark a tool agent as a delegate, the orchestrator returns its
+response directly to the caller, skipping the extra orchestrator model round-trip.
+Use delegation when the sub-agent's response is already complete and the
+orchestrator adds no value by processing it further:
```
opieter-aws: > Use delegation when the sub-agent's response is already complete

What does this mean? When I set up an agent as delegate, it's just constructor no agent loop?

site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
```diff
@@ -333,6 +335,68 @@ response = orchestrator(customer_query)
 
 This example demonstrates how Strands Agents SDK enables specialized experts to collaborate on complex queries requiring multiple domains of knowledge. The orchestrator intelligently routes different aspects of the query to the appropriate specialized agents, then synthesizes their responses into a comprehensive answer.
 
+## Delegation
```
opieter-aws, line 338: I would lead this with a much clearer use case, rather than a technical description. E.g. "Sometimes you do not want a sub-agent's response to be re-processed by the orchestrator agent since this adds to latency and token usage. In that case, ......"

site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
```diff
@@ -333,6 +335,68 @@ response = orchestrator(customer_query)
...
+To mark a tool agent as a delegate, pass `delegate: true` to
+<Syntax py=".as_tool()" ts=".asTool()" />:
+
+<Tabs>
+<Tab label="Python">
+
+```python
+from strands import Agent
+
+billing_agent = Agent(
```
opieter-aws, line 366: Should we maybe pick an example with structured output to make the use case stronger?

site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
```diff
@@ -333,6 +335,68 @@ response = orchestrator(customer_query)
...
+<Syntax py="stop_reason == 'end_turn'" ts="stopReason === 'endTurn'" /> and
+the `AgentResult` contains the delegate's content without additional
+model processing. Streaming events from the delegate surface natively in
+the parent's stream, so callers see the specialist's tokens as they arrive.
+
+Limitations to be aware of:
+- **Single delegated tool per turn**: Delegated and non-delegated tools
+  can exist on the same agent. However, a delegated tool must be
+  the only tool call in its turn. The SDK cancels the batch if
+  other tools are requested alongside it.
```
opieter-aws: This is hard to follow, can you clarify?

site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
liramon2: When you set up an agent as a delegate and the parent agent calls the delegated agent, the parent agent ends its own agent loop and passes the delegated agent's response to the user. I can clarify this.

opieter-aws (approved):

PR #3486, opened 2026-07-26
feat(a2a): support the interrupt round trip over A2A

### LINKED ISSUE #2948: [BUG] `StrandsA2AExecutor` does not translate inbound A2A interrupt-response DataParts into Strands `InterruptResponseContent` blocks on resume
### Checks

### SDK Language

Python

### Strands Version

1.42.0 (also verified gap is present in 1.44.0 from release notes — no A2A interrupt resume fix listed)

### Language Runtime Version

Python 3.12

### Operating System

Amazon Linux 2023 (AgentCore runtime) / macOS 15 (local dev)  ## Installation Method

### Installation Method

pip

### Steps to Reproduce

1.  Create a Strands Agent with a tool that calls `tool_context.interrupt()`:
    
    ```python
    @tool
    def create_campaign(tool_context, campaign_name: str) -> str:
        """Create a campaign after approval."""
        approval = tool_context.interrupt("approve_campaign", reason={"name": campaign_name})
        # ... handle approval ...
        return f"Campaign '{campaign_name}' created"
    ```
    
2.  Wire it as an A2A server using `StrandsA2AExecutor`:
    
    ```python
    from strands.multiagent.a2a.executor import StrandsA2AExecutor
    
    executor = StrandsA2AExecutor(agent, enable_a2a_compliant_streaming=True)
    ```
    
3.  Send an initial `message/send` request — the executor correctly returns `input-required` status with interrupt details ✅
    
4.  Send a follow-up `message/send` with the same `taskId` containing a DataPart with the user's interrupt response:
    
    ```json
    {
      "method": "message/send",
      "params": {
        "taskId": "<existing-task-id>",
        "message": {
          "role": "user",
          "parts": [
            {
              "kind": "data",
              "data": {
                "type": "interrupt_response",
                "id": "approve_campaign",
                "response": { "approved": true }
              }
            }
          ]
        }
      }
    }
    ```
    
5.  **Result**: Agent fails to resume — the DataPart is converted to a plain text `ContentBlock` (`[Structured Data]\n{...}`) and fed as a new turn instead of as `InterruptResponseContent` blocks.

### Expected Behavior

1.  Create a Strands Agent with a tool that calls `tool_context.interrupt()`:
    
    ```python
    @tool
    def create_campaign(tool_context, campaign_name: str) -> str:
        """Create a campaign after approval."""
        approval = tool_context.interrupt("approve_campaign", reason={"name": campaign_name})
        # ... handle approval ...
        return f"Campaign '{campaign_name}' created"
    ```
    
2.  Wire it as an A2A server using `StrandsA2AExecutor`:
    
    ```python
    from strands.multiagent.a2a.executor import StrandsA2AExecutor
    
    executor = StrandsA2AExecutor(agent, enable_a2a_compliant_streaming=True)
    ```
    
3.  Send an initial `message/send` request — the executor correctly returns `input-required` status with interrupt details ✅
    
4.  Send a follow-up `message/send` with the same `taskId` containing a DataPart with the user's interrupt response:
    
    ```json
    {
      "method": "message/send",
      "params": {
        "taskId": "<existing-task-id>",
        "message": {
          "role": "user",
          "parts": [
            {
              "kind": "data",
              "data": {
                "type": "interrupt_response",
                "id": "approve_campaign",
                "response": { "approved": true }
              }
            }
          ]
        }
      }
    }
    ```
    
5.  **Result**: Agent fails to resume — the DataPart is converted to a plain text `ContentBlock` (`[Structured Data]\n{...}`) and fed as a new turn instead of as `InterruptResponseContent` blocks.

### Actual Behavior

In `StrandsA2AExecutor._execute_streaming()` → `_convert_a2a_parts_to_content_blocks()`, DataParts are unconditionally serialized to text:

```python
elif isinstance(part_root, DataPart):
    # Handle DataPart - convert structured data to JSON text
    try:
        data_text = json.dumps(part_root.data, indent=2)
        content_blocks.append(ContentBlock(text=f"[Structured Data]\n{data_text}"))
    except Exception:
        logger.exception("Failed to serialize data part")
```

There is no check for `data["type"] == "interrupt_response"`, no conversion to `InterruptResponseContent`, and no call to the resume path. The agent receives `[Structured Data]\n{"type": "interrupt_response", ...}` as plain text, which does not satisfy the interrupt resume contract and causes the agent loop to fail or start a new unrelated turn.

### Additional Context

-   The **outbound** interrupt path works correctly: when a tool calls `tool_context.interrupt()`, `_handle_interrupt_result()` properly emits `input_required` status with interrupt details.
    
-   The types needed for the fix already exist in Strands: `strands.types.interrupt.InterruptResponseContent` and `strands.types.interrupt.InterruptResponse`.
    
-   The A2A spec defines that a client resumes from `input-required` by sending a new `message/send` with the same `taskId` — the executor needs to recognize this is a resume, not a fresh conversation turn.
    
-   There is no documented wire format for how interrupt responses should be encoded in A2A DataParts. Proposing:
    
    ```json
    { "kind": "data", "data": { "type": "interrupt_response", "id": "<interruptId>", "response": <any JSON> } }
    ```
    

### Possible Solution

In `_execute_streaming` (or a helper called before `_convert_a2a_parts_to_content_blocks`), add early detection:

```python
async def _execute_streaming(self, context: RequestContext, updater: TaskUpdater) -> None:
    # Check for interrupt resume BEFORE converting parts to generic content blocks
    interrupt_blocks = self._extract_interrupt_responses(context)
    if interrupt_blocks:
        await self._stream_resume(interrupt_blocks, context, updater)
        return
    # ... existing logic for normal first-turn execution ...

def _extract_interrupt_responses(self, context: RequestContext) -> list[InterruptResponseContent]:
    parts = getattr(context.message, "parts", None) or []
    blocks = []
    for part in parts:
        root = getattr(part, "root", None)
        if not isinstance(root, DataPart):
            continue
        data = root.data
        if data.get("type") not in ("interrupt_response", "interruptResponse"):
            continue
        interrupt_id = data.get("id") or data.get("interruptId") or ""
        blocks.append(InterruptResponseContent(
            interruptResponse=InterruptResponse(interruptId=interrupt_id, response=data.get("response"))
        ))
    return blocks

async def _stream_resume(self, blocks, context, updater) -> None:
    invocation_state = {"a2a_request_context": context}
    result = None
    async for event in self.agent.stream_async(blocks, invocation_state=invocation_state):
        if "result" in event:
            result = event["result"]
        else:
            await self._handle_streaming_event(event, updater)
    if result is not None and result.stop_reason == "interrupt":
        await self._handle_interrupt_result(result, updater)
    else:
        await self._handle_agent_result(result, updater)
```

### Related Issues

_No response_

vaibhav-walia:
## Description

`StrandsA2AExecutor` could raise interrupts but never complete them, so human-in-the-loop did not work over A2A at all. A tool calling `tool_context.interrupt()` parked its task permanently, for two reasons:

- **Inbound**: `_convert_a2a_parts_to_content_blocks()` serialized every DataPart to `[Structured Data]\n{...}` text. A client's interrupt response arrived as a plain conversational turn rather than the `list[InterruptResponseContent]` that `_InterruptState.resume()` requires, so the parked tool never resumed and the approval payload reached the model as ordinary text it could be persuaded by.
- **Outbound**: interrupt ids are generated server-side (`v1:tool_call:<toolUseId>:<uuid>`), but `_handle_interrupt_result()` sent only the interrupt *name* and *reason* as prose. A client had no way to learn the id it needed to answer with.

Both halves are needed for a usable round trip, which is why they are one change: fixing only the inbound path leaves a feature that no client can exercise and no reviewer can validate without reaching into server internals.

Resume is now treated as a continuation of the parked state rather than a new conversational turn. Once a task is in `input_required`, the next message for it must carry an interrupt response, and every rejection happens before the agent runs — so a refused resume leaves the interrupt parked and retryable instead of letting an unverified approval through as chat.

## Public API Changes

Two DataPart contracts on the A2A wire. Neither changes any Python signature.

**Outbound** — the `input_required` status message now carries the pending interrupts. The existing `TextPart` stays first, so clients that only read prose are unaffected:

```json
{
  "kind": "data",
  "data": {
    "interrupts": [
      {"interruptId": "v1:tool_call:tu-1:a71adb48-...", "name": "approve_campaign", "reason": {"name": "spring-launch"}}
    ]
  }
}
```

**Inbound** — a client resumes by echoing the id back. The payload mirrors the existing `InterruptResponseContent` type verbatim, so the wire contract and the SDK type cannot drift, and the same shape re-cases directly to TypeScript:

```json
{
  "kind": "data",
  "data": {"interruptResponse": {"interruptId": "v1:tool_call:tu-1:a71adb48-...", "response": {"approved": true}}}
}
```

A full round trip from the client's side:

```python
# 1. Ask for something that needs approval -> task parks in input_required
response = await client.send_message(message_with_text("create the spring-launch campaign"))

# 2. Read the interrupt id off the status message
interrupts = next(
    part.root.data["interrupts"]
    for part in response.status.message.parts
    if part.root.kind == "data" and "interrupts" in part.root.data
)

# 3. Answer it, on the same taskId
await client.send_message(
    message_with_parts(
        task_id=response.id,
        parts=[Part(root=DataPart(data={
            "interruptResponse": {"interruptId": interrupts[0]["interruptId"], "response": {"approved": True}}
        }))],
    )
)
```

### Rejected inputs

Recognition is narrow and binding fails closed. All of these raise `InvalidParamsError` before the agent is invoked:

| Input | Why |
| --- | --- |
| Interrupt id not parked on this context's agent | An approval must bind to the interrupt that requested it |
| Interrupt response with no interrupt pending | Nothing to resume |
| Two responses for the same interrupt id | Ambiguous; picking one silently would discard a decision |
| Missing/blank `interruptId`, missing `response`, non-object payload | Malformed |
| Interrupt response mixed with other content parts | The agent resumes from responses alone, so the other content would be silently dropped |
| A fresh conversational turn while a task is parked | A parked task is answered or cancelled, not talked past |

A DataPart without an `interruptResponse` key is untouched and still takes the generic structured-data path.

The "mixed with other content parts" rule is the strictest of these and deliberately so. `_InterruptState.resume()` rejects any prompt whose contents are not all interrupt responses, so accompanying parts cannot be delivered — only dropped — and dropping them silently is what this change exists to avoid. It does mean a client following the common convention of pairing a `DataPart` with a human-readable `TextPart` gets rejected. Starting strict keeps the option open, since accepting more inputs later is backward compatible while withdrawing acceptance is not. Easy to relax to "ignore an accompanying TextPart" if you would rather meet that convention now.

## Design notes for reviewers

**Why this inbound shape.** It is already the cross-SDK JSON form of the type. `InterruptResponseContent.toJSON()` in `strands-ts/src/types/interrupt.ts` emits exactly `{interruptResponse: {interruptId, response}}`, `fromJSON()` reads it back, and the type guard tests for the `interruptResponse` key — so this contract needs no mapping layer in either SDK, and TypeScript parity can adopt it as-is. The alternative considered was an explicit envelope (`{"type": "interrupt_response", "id": ...}`), closer to the ADK sample's `{"type": "form"}` discriminator, but that would have introduced a second name for a type both SDKs already serialize, and `interruptResponse` is a unique self-discriminating key. Happy to switch if you would rather have the discriminator.

Worth flagging that there is otherwise no precedent to borrow here: across `a2a-samples`, no implementation sends a machine-readable resume payload. The ADK and LangGraph agents both route resumption through the model, which reads a correlation id out of conversational text and passes it to a tool, and the reference CLI answers `input_required` with a plain `TextPart`. Strands cannot do that — `_InterruptState.resume()` binds to an exact interrupt id and runs *before* the model, so there is no model in the loop to extract an id from prose.

**TypeScript parity.** `strands-ts/src/a2a/executor.ts` has the same gap and is untouched here, since that is a separate sub-project and a separate PR. The docs mark the feature Python-only. Happy to open the parity issue, or take it as a follow-up PR once the wire contract is settled here.

**Why this outbound shape.** This part does have precedent. The ADK expense-reimbursement sample sends a server-minted correlation id in a `DataPart` on `input_required` and rejects unknown ids on the way back — the same pattern implemented here.

**Where binding is checked.** Shape and duplicate validation happen at extraction, but id binding happens in `_stream_agent`, because the agent — and with it the interrupt state the request must match — is only resolved after the context's snapshot is restored.

## Documentation PR

Included here, under `site/`: a new **Answering Interrupts** section in `user-guide/concepts/multi-agent/agent-to-agent.mdx`, covering both wire shapes, a client example, and the rejection rules. Marked Python-only per the existing `:::note[Python only]` convention. `npm run build` passes with no broken links.

## Related Issues

Closes #2948

## Type of Change

New feature

Bug fix in effect — it makes #2948's scenario work — but it adds public wire surface, so it wants a minor bump rather than a patch.

## Testing

`hatch run prepare` passes for this change: ruff, mypy, and the full suite are green.

One unrelated flake to note, so it isn't mistaken for this change: `tests/strands/tools/mcp/test_mcp_client.py::test_call_tool_async_caller_cancellation_timeout_preserves_cancelled_error` fails intermittently in the full py3.13 run and passes in isolation. It reproduces on an unmodified `main` (1 failure in 3 runs) and touches no code this PR changes. Happy to open a separate issue for it.

Beyond the unit tests, I drove the round trip end to end against a real `Agent` with a real interrupting tool, through the A2A surface only — the test client reads the interrupt id off the outbound message exactly as a real client would, with no access to executor internals:

```
1. states=['input-required']
2. advertised=[{'interruptId': 'v1:tool_call:tu-1:a71adb48-...', 'name': 'approve_campaign', 'reason': {'name': 'spring-launch'}}]
3. guessed id refused: No pending interrupt matches id(s): guessed-id
4. states=['completed']  final='Campaign is live.'
```

Verified in both isolation modes: `agent_factory`, and the deprecated shared-agent mode where the parked interrupt survives the snapshot swap via the `session` preset's `interrupt_state` field.

## Checklist

----

### DIFF
--- a/site/src/content/docs/user-guide/concepts/multi-agent/agent-to-agent.mdx
+++ b/site/src/content/docs/user-guide/concepts/multi-agent/agent-to-agent.mdx
@@ -381,11 +381,13 @@ In single-agent mode the server reuses one agent across every context, swapping 
 
 ### Answering Interrupts
 
-When a tool or hook on the served agent raises an [interrupt](../interrupts.md), the task moves to the A2A `input_required` state and waits. The client answers it, and the task resumes the paused tool exactly where it stopped.
-
 :::note[Python only]
 This feature is currently available in the Python SDK only.
 :::
+
+When a tool or hook on the served agent raises an [interrupt](../interrupts.md), the task moves to the A2A `input_required` state and waits. The client answers it, and the task resumes the paused tool exactly where it stopped.
+
+Answering an interrupt is the one flow on this page that needs the raw [`a2a-sdk`](https://github.com/a2aproject/a2a-python) client rather than `A2AAgent`. `A2AAgent` speaks in text, so it raises `ValueError` if you pass it interrupt responses, and it drops the `DataPart` carrying the interrupt ids when it reads the reply. The examples below build A2A messages directly.
 
 Each interrupt has a server-generated id, and an answer is bound to the id of the interrupt that raised it. The server advertises the pending interrupts on the `input_required` status message as a `DataPart`, alongside the human-readable `TextPart`:
 
@@ -418,7 +420,7 @@ To answer, send a new message on the same `taskId` containing a `DataPart` that 
 }
 ```
 
-The `response` is any JSON value, and it becomes the return value of the `interrupt()` call that paused the tool. Answer several interrupts in one message by sending one `DataPart` for each.
+The `response` becomes the return value of the `interrupt()` call that paused the tool. It is any JSON value except `null`, which the server refuses — a null answer would leave the interrupt unsatisfied and re-raise it. `false` and `0` are fine. Answer several interrupts in one message by sending one `DataPart` for each.
 
 Reading the ids off the status message and answering them:
 
@@ -441,7 +443,15 @@ answers = [
 ]
 ```
 
-The server rejects an answer it cannot bind, before the agent runs, so a refused answer leaves the interrupt pending and the task still answerable. An answer is rejected when its `interruptId` does not match an interrupt the task is waiting on, when no interrupt is pending, when the same `interruptId` is answered twice in one message, when the payload is missing `interruptId` or `response`, or when it is sent alongside other content parts. A task with a pending interrupt also rejects an ordinary conversational message — answer the interrupt, or cancel the task.
+The server rejects an answer it cannot bind, before the agent runs, so a refused answer leaves the interrupt pending and the task still answerable. An answer is rejected when:
+
+- its `interruptId` does not match an interrupt the task is waiting on
+- no interrupt is pending
+- the same `interruptId` is answered twice in one message
+- the payload is missing `interruptId`, or its `response` is missing or `null`
+- it is sent alongside other content parts
+
+A task with a pending interrupt also rejects an ordinary conversational message — answer the interrupt, or cancel the task.
 
 A `DataPart` without an `interruptResponse` key is unaffected and continues to reach the agent as structured data.
 
--- a/site/src/content/docs/user-guide/concepts/multi-agent/agent-to-agent.mdx
+++ b/site/src/content/docs/user-guide/concepts/multi-agent/agent-to-agent.mdx
@@ -379,6 +379,72 @@ Contexts are keyed on the client supplied `context_id`. A caller who knows anoth
 In single-agent mode the server reuses one agent across every context, swapping each context's saved state on and off the shared instance under a lock, which serializes all requests. A single `agent` with a configured `session_manager` is rejected, because the session manager would persist every context into one interleaved session. Prefer `agent_factory` for any deployment that serves more than one conversation.
 :::
 
+### Answering Interrupts
+
+When a tool or hook on the served agent raises an [interrupt](../interrupts.md), the task moves to the A2A `input_required` state and waits. The client answers it, and the task resumes the paused tool exactly where it stopped.
+
+:::note[Python only]
+This feature is currently available in the Python SDK only.
+:::
+
+Each interrupt has a server-generated id, and an answer is bound to the id of the interrupt that raised it. The server advertises the pending interrupts on the `input_required` status message as a `DataPart`, alongside the human-readable `TextPart`:
+
+```json
+{
+  "kind": "data",
+  "data": {
+    "interrupts": [
+      {
+        "interruptId": "v1:tool_call:tu-1:a71adb48-e65d-55fa-b155-0359c9cd3b66",
+        "name": "approve_campaign",
+        "reason": {"name": "spring-launch"}
+      }
+    ]
+  }
+}
+```
+
+To answer, send a new message on the same `taskId` containing a `DataPart` that echoes the `interruptId` back with the response:
+
+```json
+{
+  "kind": "data",
+  "data": {
+    "interruptResponse": {
+      "interruptId": "v1:tool_call:tu-1:a71adb48-e65d-55fa-b155-0359c9cd3b66",
+      "response": {"approved": true}
+    }
+  }
+}
+```
+
+The `response` is any JSON value, and it becomes the return value of the `interrupt()` call that paused the tool. Answer several interrupts in one message by sending one `DataPart` for each.
+
+Reading the ids off the status message and answering them:
+
+```python
+from a2a.types import DataPart, Part
+
+# The task parked in input_required; read the interrupts it is waiting on.
+pending = next(
+    part.root.data["interrupts"]
+    for part in task.status.message.parts
+    if isinstance(part.root, DataPart) and "interrupts" in part.root.data
+)
+
+# Answer each one on the same taskId.
+answers = [
+    Part(root=DataPart(data={
+        "interruptResponse": {"interruptId": item["interruptId"], "response": {"approved": True}}
+    }))
+    for item in pending
+]
+```
+
+The server rejects an answer it cannot bind, before the agent runs, so a refused answer leaves the interrupt pending and the task still answerable. An answer is rejected when its `interruptId` does not match an interrupt the task is waiting on, when no interrupt is pending, when the same `interruptId` is answered twice in one message, when the payload is missing `interruptId` or `response`, or when it is sent alongside other content parts. A task with a pending interrupt also rejects an ordinary conversational message — answer the interrupt, or cancel the task.
+
+A `DataPart` without an `interruptResponse` key is unaffected and continues to reach the agent as structured data.
+
 ### Server Configuration Options
 
 <Tabs>
--- a/strands-py/src/strands/multiagent/a2a/executor.py
+++ b/strands-py/src/strands/multiagent/a2a/executor.py
@@ -18,12 +18,21 @@ import warnings
 from collections import OrderedDict
 from collections.abc import Callable
 from dataclasses import dataclass
-from typing import Any, Literal
+from typing import Any, Literal, cast
 
 from a2a.server.agent_execution import AgentExecutor, RequestContext
 from a2a.server.events import EventQueue
 from a2a.server.tasks import TaskUpdater
-from a2a.types import DataPart, FilePart, InternalError, Part, TaskState, TextPart, UnsupportedOperationError
+from a2a.types import (
+    DataPart,
+    FilePart,
+    InternalError,
+    InvalidParamsError,
+    Part,
+    TaskState,
+    TextPart,
+    UnsupportedOperationError,
+)
 from a2a.utils import new_agent_text_message, new_task
 from a2a.utils.errors import ServerError
 
@@ -32,6 +41,7 @@ from ...agent.agent import AgentResult as SAAgentResult
 from ...session.session_manager import SessionManager
 from ...types._snapshot import Snapshot
 from ...types.content import ContentBlock
+from ...types.interrupt import InterruptResponse, InterruptResponseContent
 from ...types.media import (
     DocumentContent,
     DocumentSource,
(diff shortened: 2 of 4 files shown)


vaibhav-walia:
@liramon2 you were assigned reviewer automatically, lmk if you need more info from me, would be happy to discuss and iterate.

site/src/content/docs/user-guide/concepts/multi-agent/agent-to-agent.mdx
```diff
@@ -379,6 +379,72 @@ Contexts are keyed on the client supplied `context_id`. A caller who knows anoth
...
+# Answer each one on the same taskId.
+answers = [
+    Part(root=DataPart(data={
+        "interruptResponse": {"interruptId": item["interruptId"], "response": {"approved": True}}
+    }))
+    for item in pending
+]
+```
+
+The server rejects an answer it cannot bind, before the agent runs, so a refused answer leaves the interrupt pending and the task still answerable. An answer is rejected when its `interruptId` does not match an interrupt the task is waiting on, when no interrupt is pending, when the same `interruptId` is answered twice in one message, when the payload is missing `interruptId` or `response`, or when it is sent alongside other content parts. A task with a pending interrupt also rejects an ordinary conversational message — answer the interrupt, or cancel the task.
```
liramon2: Nit: this could be a bulleted list, since it lists 5 different rejection conditions. This would improve scannability.

strands-py/src/strands/multiagent/a2a/executor.py
```diff
@@ -583,6 +660,117 @@ def _strip_file_extension(self, file_name: str) -> str:
...
+                    )
+                ) from None
+
+            interrupt_id = response.get("interruptId")
+            if not isinstance(interrupt_id, str) or not interrupt_id:
+                raise ServerError(
+                    error=InvalidParamsError(message="Interrupt response is missing a non-empty 'interruptId'")
+                ) from None
+
+            # A null response is a legitimate answer, so presence of the key is what matters.
```
liramon2: `_extract_interrupt_responses` accepts `response: null`, and the new docs tell clients that response is "any JSON value." But the core uses None as the sentinel for "not yet answered"; `_Interruptible.interrupt()` only returns the response when is not None, so a null answer causes the same interrupt to re-fire silently. The client gets back an identical input_required with no error. I suggest rejecting null at extraction and revising the docs/tests to treat null as invalid.

site/src/content/docs/user-guide/concepts/multi-agent/agent-to-agent.mdx
```diff
@@ -379,6 +379,72 @@ Contexts are keyed on the client supplied `context_id`. A caller who knows anoth
 In single-agent mode the server reuses one agent across every context, swapping each context's saved state on and off the shared instance under a lock, which serializes all requests. A single `agent` with a configured `session_manager` is rejected, because the session manager would persist every context into one interleaved session. Prefer `agent_factory` for any deployment that serves more than one conversation.
 :::
 
+### Answering Interrupts
```
liramon2, line 382: Every other example on `agent-to-agent.mdx` uses the `A2AAgent` client wrapper, and the sibling `interrupts.mdx` teaches `agent(responses)` for local agents. The reader will try `a2a_agent(answers)`, which raises a `ValueError` and reading pending interrupts through `A2AAgent` silently drops the DataPart. The new section never says this. Fix: one sentence up front explaining the round trip needs the raw `a2a-sdk` client shown, not `A2AAgent`.

strands-py/tests/strands/multiagent/a2a/test_executor.py
```diff
@@ -2268,3 +2268,443 @@ def artifact_ids(events):
...
+async def test_execute_interrupt_response_when_not_parked_fails_closed(
+    mock_strands_agent, mock_request_context, mock_event_queue
+):
+    """Interrupt responses sent to an agent holding no interrupt are rejected."""
+    mock_strands_agent.stream_async = MagicMock()
+    executor = StrandsA2AExecutor(mock_strands_agent)
+
+    _request_with_parts(mock_request_context, [_interrupt_response_part("int-1", {"approved": True})])
+
+    with pytest.raises(ServerError):
```
liramon2: This test and the other tests below that check for `ServerError` should also check the wrapped error type.

site/src/content/docs/user-guide/concepts/multi-agent/agent-to-agent.mdx
```diff
@@ -379,6 +379,72 @@ Contexts are keyed on the client supplied `context_id`. A caller who knows anoth
...
+
+When a tool or hook on the served agent raises an [interrupt](../interrupts.md), the task moves to the A2A `input_required` state and waits. The client answers it, and the task resumes the paused tool exactly where it stopped.
+
+:::note[Python only]
+This feature is currently available in the Python SDK only.
```
liramon2, line 385: Nit: move this right below the heading (1 block up)

liramon2 (changes requested):

liramon2:
@vaibhav-walia Thanks for your contribution. I reviewed your PR and have some suggestions outlined above.

vaibhav-walia:
Thanks for the review @liramon2 appreciate it!
I've addressed and pushed changes.

liramon2 (approved):

liramon2:
@vaibhav-walia Approved. Feel free to close any conversations and rebase when you're ready to merge.

vaibhav-walia:
@liramon2 resolved conversations and rebased from UI. Please lmk if anything else is needed from my end

PR #3490, opened 2026-07-27
feat(python): support audio content in Bedrock provider

### LINKED ISSUE #2614: [FEATURE] Support audio input for standard Agent (Specially using Bedrock Provider, Converse API already supports AudioBlock).
### Problem Statement

AWS Bedrock's Converse API natively supports audio inputs via the AudioBlock structure. Exposing this capability through our standard Agent interface will allow agents to perform direct transcription, summarization, and contextual analysis of voice data in a single step, bypassing the need for intermediate STT services.

### Proposed Solution

Proposed Solution

Update Core Schema: Extend the standard Agent's input schema to accept audio payloads (e.g., base64 encoded strings, buffers, or file paths along with their MIME type).

Bedrock Provider Implementation: Map the new agent audio input type directly to the AudioBlock format required by the Bedrock Converse API.

Validation: Implement standard validation for supported audio formats (e.g., mp3, wav, flac) and size limits as defined by the underlying model providers.

### Use Case

Primary Use Case: Direct Transcription & Speech-to-Text

Single-Step Audio Processing: Users can pass an audio file (e.g., a voicemail, meeting recording, or user voice note) directly to the agent with a prompt like "Transcribe this recording and extract all action items."

Pipeline Simplification: Eliminates the latency, cost, and architectural complexity of maintaining a separate transcription service just to feed text into the LLM.

Context-Aware Transcription: By processing the audio directly, multimodal LLMs can often capture nuances, speaker intent, and domain-specific terminology better than a standalone, context-blind STT model.

### Alternatives Solutions

_No response_

### Additional Context

_No response_

shusingh:
## Description

Bedrock's Converse API accepts audio input via `AudioBlock`, but the SDK had no way to express it: passing an audio content block raised `TypeError: content_type=<audio> | unsupported type`. This left a gap for voice-centric agents — transcription, voicemail summarization, meeting analysis — which had to run a separate STT service just to feed text to the model.

This adds audio as a first-class content type, modeled exactly like the existing video support:

- `AudioFormat`, `AudioSource`, `AudioContent` in `types/media.py`, with the format literal matching the Converse `AudioBlock` valid values and `source` accepting raw `bytes` or an S3 `location`.
- `audio` on `ContentBlock`.
- A mapping branch in the Bedrock provider's `_format_request_message_content`, with the same strict field filtering and s3Location handling as the video branch.

Other providers intentionally keep their explicit unsupported-type behavior for audio; this PR is scoped to Bedrock per the issue. TypeScript SDK parity is a natural follow-up — happy to open it next.

## Public API Changes

```python
agent = Agent(model=BedrockModel(model_id="us.amazon.nova-pro-v1:0"))

agent([
    {"text": "Transcribe this recording and extract all action items."},
    {"audio": {"format": "mp3", "source": {"bytes": recording_bytes}}},
])
```

`AudioContent` also supports S3 sources, matching image/video:

```python
{"audio": {"format": "wav", "source": {"location": {"type": "s3", "uri": "s3://bucket/call.wav"}}}}
```

## Notes

An empty `source` dict raises `UnboundLocalError` rather than a clearer error — this matches the behavior of the existing image/video/document branches, so I kept the pattern consistent rather than special-casing audio; happy to fix all four in a follow-up if wanted.

## Related Issues

Closes #2614

## Documentation PR

No site changes included; can follow up if a docs example is wanted.

## Type of Change

New feature

## Testing

Unit tests mirror the existing video coverage: bytes source, s3Location source, non-S3 location skipping, and extra-field filtering, plus `AudioSource` shape tests. Also exercised the issue's repro end to end with a mocked boto session and confirmed the formatted request carries the `audio` block.

## Checklist

----


strands-py/tests/strands/types/test_media.py
```diff
@@ -75,6 +76,30 @@ def test_image_source_with_s3_location(self):
         assert img_source["s3Location"]["uri"] == "s3://my-bucket/images/photo.png"
 
 
+class TestAudioSource:
```
Unshure: You can just remove this test, it does not provide much value

strands-py/src/strands/models/bedrock.py
```diff
@@ -605,10 +605,24 @@ def _format_request_message_content(self, content: ContentBlock) -> dict[str, An
...
+                    return None
+            elif "bytes" in source:
+                formatted_audio_source = {"bytes": source["bytes"]}
+            result: dict[str, Any] = {"format": audio["format"], "source": formatted_audio_source}
+            return {"audio": result}
+
         # https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_CachePointBlock.html
         if "cachePoint" in content:
             cache_point = content["cachePoint"]
-            result: dict[str, Any] = {"type": cache_point["type"]}
```
Unshure, line 611: Can you add this typing back?

Unshure (commented):
Can you also add an integration test for this new content type?

strands-py/tests/strands/types/test_media.py
shusingh: Removed `TestAudioSource` entirely (both cases) in 2aca4a5. Agreed it added little — the meaningful bytes/s3 coverage lives in `test_format_request_audio_*` in test_bedrock.py.

strands-py/src/strands/models/bedrock.py
shusingh: Done in 2aca4a5. I inlined the audio block's return so the branch no longer introduces `result`, which lets the `cachePoint` line keep its original `result: dict[str, Any]` annotation — the net diff on that line is now zero.

strands-py/tests/strands/models/test_bedrock.py
```diff
@@ -2361,6 +2361,106 @@ def test_format_request_maps_3gp_video_formats(model, model_id, video_format):
...
+            ],
+        }
+    ]
+
+    formatted_request = model.format_request(messages)
+
+    audio_block = formatted_request["messages"][0]["content"][0]["audio"]
+    expected = {"format": "flac", "source": {"bytes": b"audio_data"}}
+    assert audio_block == expected
+    assert "duration" not in audio_block
```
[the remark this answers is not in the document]
shusingh: Done in 2aca4a5 — dropped the two `not in` assertions; the `== expected` equality check already guarantees the exact shape.

strands-py/tests/strands/models/test_bedrock.py
```diff
@@ -2361,6 +2361,106 @@ def test_format_request_maps_3gp_video_formats(model, model_id, video_format):
...
+                    }
+                },
+            ],
+        }
+    ]
+
+    formatted_request = model.format_request(messages)
+    audio_source = formatted_request["messages"][0]["content"][0]["audio"]["source"]
+
+    assert audio_source == {"s3Location": {"uri": "s3://my-bucket/audio.wav"}}
```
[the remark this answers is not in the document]
shusingh: Done in 2aca4a5 — the s3 test now asserts the full audio block: `{"format": "wav", "source": {"s3Location": {"uri": ...}}}`, so a regression in `format` on the s3 path would be caught.

strands-py/tests/strands/types/test_media.py
shusingh: Good catch on both — removed the shape tests in 2aca4a5, and fixed the `AudioSource` docstring to say "bytes or `location`" to match the actual field. I left the identical wording in the image/video/document docstrings untouched to keep this PR focused; happy to open a tiny follow-up for those.

shusingh:
@Unshure on the integration test — I dug into this and hit a snag worth surfacing before I add one.

I couldn't find a Bedrock model that accepts audio input over **Converse**:
- **Nova 2 Lite / Nova Lite / Nova Pro** (the models `tests_integ` already uses for image/video) list **Audio as an unsupported input modality** on their model cards — text, image, and video only.
- **Nova Sonic / Nova 2 Sonic** do support audio, but only via `InvokeModelWithBidirectionalStream`, not Converse.
- **Nova Multimodal Embeddings** takes audio but is an embeddings + async-invoke model, not a Converse chat model.

So a live integ test (mirroring `test_video_s3_location`) would fail at the service with a validation error, since no available Converse model would accept the `AudioBlock`.

Could you point me at an audio-capable Converse model that CI has access to? I'll add the integration test against it right away — happy to do both bytes and s3Location cases to match the existing media coverage. If there isn't one yet, I'm glad to add the test in a skipped/xfail state referencing this thread so it's ready to enable the moment such a model lands, whichever you'd prefer.

dQuezada-P:
@shusingh  You can try with "mistral.voxtral-mini-3b-2507", it supports mp3 audio in ConverseAPI.

shusingh:
@dQuezada-P that was it, thank you — `mistral.voxtral-mini-3b-2507` accepts mp3 over Converse.

@Unshure integration tests added in cd715af5:

- `test_invoke_audio_input` — bytes source, asserts the model reports the spoken color
- `test_audio_s3_location` — S3 source, alongside the existing document/image/video cases

New `yellow.mp3` fixture is a 1s mono clip (4.9 KB) of "The color is yellow." Voxtral is in-region only (no geo/global profile), so the region is pinned: `us-east-1` for the bytes test, `us-west-2` for the S3 one to match that module's bucket.

I don't have Bedrock credentials to run these locally — I verified both request shapes offline against the `AudioBlock` API reference. They'll need an integ workflow run, and model access for Voxtral in those two regions.

PR #3506, opened 2026-07-27
feat(stop): use agent.cancel() to stop the event loop

Unshure:
## Summary

- Updated `agent.cancel()` in both Python and TypeScript SDKs to accept an optional `message` parameter that surfaces as the final assistant text when the loop terminates
- Rewrote the experimental stop tool to call `agent.cancel(message)` instead of setting internal flags (`stop_event_loop` in Python, `AfterToolsEvent.endTurn` marker in TypeScript)
- Added `cancel(message?: string): void` to the TypeScript `LocalAgent` interface so tools can call it
- Simplified the TypeScript stop tool by removing the `AfterToolsEvent` hook, `WeakSet` tracking, and `invocationState` marker pattern

## Test plan

- [x] Python stop tool tests pass (16/16) — verify `agent.cancel()` called with correct message
- [x] TypeScript stop tool tests pass (17/17) — verify `agent.cancel()` called with correct message
- [x] TypeScript full test suite passes (873 agent tests, 16 cancel tests)
- [x] TypeScript type-check and build pass
- [ ] Integration test: agent with stop tool returns `stopReason: 'cancelled'` with the custom message

### DIFF
--- a/team/designs/0017-deferred-cancel.md
+++ b/team/designs/0017-deferred-cancel.md
@@ -1,0 +1,229 @@
+# Deferred Cancellation
+
+**Status**: Proposed
+
+**Date**: 2026-07-27
+
+**Issue**: TBD
+
+## Problem
+
+The SDK needs a mechanism for tools to signal that the agent loop should stop after the current tool batch completes. The `stop` experimental tool is the primary consumer: it allows a model to gracefully end the loop with an optional final message. The question is how to implement this without introducing a one-off flag that only one tool uses, while preserving cooperative semantics (sibling tools in the same batch run to completion).
+
+### Current State
+
+There are two termination primitives in the SDK today:
+
+1. **`agent.cancel()`** — an immediate abort signal. Sets a flag (Python `threading.Event` / TypeScript `AbortController`) that the event loop checks at multiple points:
+   - Before tool execution (per-tool in sequential executor)
+   - During model streaming
+   - After tool execution, before the next model call
+
+   When triggered, pending tools in the batch receive error results ("Tool execution cancelled") and the loop exits with `stopReason: 'cancelled'`. This is designed for external callers (timeouts, user-initiated abort, web request disconnects).
+
+2. **`invocation_state` flags** — the `stop_event_loop` flag (Python) and `AfterToolsEvent.endTurn` marker (TypeScript) are checked only after the entire tool batch completes. They provide cooperative semantics: all tools finish, then the loop exits. However, they require tools to write to `invocationState` (a mutable shared bag) and the event loop to check a one-off key.
+
+Neither is right for a tool-initiated graceful stop:
+
+- `cancel()` is too aggressive — it aborts sibling tools mid-batch and uses pre-tool-execution checkpoints that break cooperative guarantees.
+- `invocationState` flags work cooperatively but are ad-hoc internal machinery that the `stop` tool shouldn't depend on. They couple the tool to undocumented event loop internals and don't compose (what if two tools both want to signal "stop after this batch"?).
+
+### Paper cuts
+
+- **`stop_event_loop` / `STOP_INVOCATION_STATE_KEY` are undocumented internal state.** They only exist to serve the stop tool but live in the general-purpose `invocationState` bag, polluting it. TypeScript additionally needed a `WeakSet`-tracked hook installation pattern to bridge from `invocationState` to `endTurn`.
+- **No way for a tool to say "stop after the batch" without touching internals.** A user writing a custom "done" tool must reverse-engineer the same flag pattern.
+- **`cancel()` lacks gradation.** It's all-or-nothing — there's no way to cancel cooperatively through the public API.
+
+## Goals and Non-Goals
+
+**Goals:**
+- A single public API for tool-initiated graceful stop: `agent.cancel(message, { afterCurrentTools: true })`
+- Cooperative semantics: sibling tools in the same batch run to completion
+- The cancel message flows through to the `AgentResult` as the final assistant text
+- Cross-SDK parity (Python and TypeScript)
+- Remove the need for `stop_event_loop` / `STOP_INVOCATION_STATE_KEY` internal flags
+- Composable: multiple tools calling deferred cancel in the same batch produces one stop (last message wins, or first — pick a policy)
+
+**Non-Goals:**
+- Changing the behavior of immediate `cancel()` (no deferred flag) — it stays an aggressive abort
+- Providing a way to "un-cancel" a deferred cancel from another tool in the same batch
+- Cancellation with custom `stopReason` values (it remains `'cancelled'`)
+
+## Proposal
+
+### Recommended: `afterCurrentTools` option on `cancel()`
+
+Add an optional second parameter to `agent.cancel()` that defers the cancellation until after the current tool batch completes.
+
+#### Python
+
+```python
+def cancel(self, message: str | None = None, *, after_current_tools: bool = False) -> None:
+```
+
+When `after_current_tools=True`:
+- Stores `message` in `self._deferred_cancel_message: str | None`
+- Sets `self._deferred_cancel: bool = True`
+- Does **not** set `self._cancel_signal`
+
+The event loop checks `agent._deferred_cancel` at the post-batch point (where `stop_event_loop` is currently checked). When found:
+- Clears `_deferred_cancel`
+- Calls `self.cancel(self._deferred_cancel_message)` to set the immediate signal
+- The existing `_cancel_signal.is_set()` check (immediately after) fires and exits with `stopReason: 'cancelled'`
+
+When `after_current_tools=False` (default, the current behavior):
+- Sets `_cancel_signal` immediately as today
+- Stores the message for use in cancellation text
+
+#### TypeScript
+
+```typescript
+public cancel(message?: string, options?: { afterCurrentTools?: boolean }): void
+```
+
+When `afterCurrentTools: true`:
+- Stores the message in `private _deferredCancelMessage: string | undefined`
+- Sets `private _deferredCancel = true`
+- Does **not** abort the `AbortController`
+
+The agent loop checks `this._deferredCancel` after the `AfterToolsEvent` fires (the same position where `endTurn` is currently checked). When found:
+- Clears `_deferredCancel`
+- Calls `this.cancel(this._deferredCancelMessage)` to abort immediately
+- The next loop iteration's `_throwIfCancelled()` fires and the `CancelledError` catch block produces the result with `stopReason: 'cancelled'`
+
+Actually, a simpler path for TypeScript: at the post-batch check, directly build the `AgentResult` with `stopReason: 'cancelled'` and return (mirroring what the `endTurn` path does today but with `'cancelled'` instead of `'endTurn'`). This avoids a round-trip through abort → next-cycle → throw → catch.
+
+#### Stop tool usage
+
+```python
+# Python
+async def stop_tool(tool_context: ToolContext, message: str | None = None) -> str:
+    final_message = _validate_message(message, max_message_length)
+    tool_context.agent.cancel(final_message, after_current_tools=True)
+    return final_message
+```
+
+```typescript
+// TypeScript
+callback: (input, context) => {
+  const message = input.message ?? DEFAULT_STOP_MESSAGE
+  context.agent.cancel(message, { afterCurrentTools: true })
+  return message
+}
+```
+
+**Pros:**
+- Single public API surface — tools call `cancel()` with an option, no internal flags
+- Cooperative semantics preserved: sibling tools complete normally
+- Composable with existing cancel infrastructure (the deferred cancel eventually becomes an immediate cancel)
+- Clean separation: the `cancel()` method owns all cancellation state, not `invocationState`
+- Easily discoverable: autocomplete on `cancel()` shows the option
+
+**Cons:**
+- Adds complexity to `cancel()` — it now has two modes
+- The deferred state (`_deferredCancel` + `_deferredCancelMessage`) is additional agent-level mutable state
+- The event loop still needs a check at the post-batch point (just checking a different field)
+- `afterCurrentTools` is only meaningful during tool execution — calling it outside that context silently behaves like immediate cancel on the next invocation
+
+### Alternative: Dedicated `stopAfterTools()` method
+
+Instead of overloading `cancel()`, add a separate method:
+
+```python
+def stop_after_tools(self, message: str | None = None) -> None:
+```
+
+```typescript
+public stopAfterTools(message?: string): void
+```
+
+**Pros:**
+- Clear separation of intent — `cancel()` is always immediate, `stopAfterTools()` is always deferred
+- No confusion about what `cancel()` does
+
+**Cons:**
+- Two public methods for conceptually related behavior (both terminate the loop)
+- Users must discover a second method rather than seeing options on the one they already know
+- Harder to explain when to use which — "cancel is for external, stopAfterTools is for tools" is a leaky abstraction since either could be called from anywhere
+- Adds to `LocalAgent` interface surface area
+
+### Alternative: Keep `invocationState` flags as an internal mechanism
+
+Leave the current `stop_event_loop` / `AfterToolsEvent.endTurn` approach in place and accept it as internal plumbing.
+
+**Pros:**
+- Already works and is tested
+- No changes to the public `cancel()` API
+- Proven cooperative semantics
+
+**Cons:**
+- Undocumented internal contract — tools depend on specific `invocationState` keys
+- TypeScript needed a `WeakSet` + hook pattern to bridge the gap — complex for what should be "set flag, loop stops"
+- Not composable or discoverable — a user writing a custom stop tool must copy the same internal pattern
+- The `invocationState` bag becomes a dumping ground for control flow signals
+
+## Developer Experience
+
+### Basic usage — stop tool
+
+```python
+from strands import Agent
+from strands.experimental.tools.stop import stop
+
+agent = Agent(model=model, tools=[stop, other_tools])
+result = await agent.invoke_async("Complete this task and stop when done.")
+# result.stop_reason == "cancelled"
+# result.message contains the model's final assistant message
+```
+
+### Custom stop tool
+
+```python
+from strands.tools.decorator import tool
+from strands.types.tools import ToolContext
+
+@tool
+async def finish(tool_context: ToolContext, summary: str) -> str:
+    """Signal that all work is complete."""
+    tool_context.agent.cancel(summary, after_current_tools=True)
+    return summary
+```
+
+### External cancellation (unchanged)
+
+```python
+# Still immediate — no after_current_tools flag
+agent.cancel()  # or agent.cancel("Timeout reached")
+```
+
+### Sibling tools complete
+
+```python
+# Model requests: [save_file(...), stop("All done")]
+# With sequential executor:
+#   1. save_file runs to completion ✓
+#   2. stop runs, calls cancel(msg, after_current_tools=True) ✓
+#   3. Batch completes, deferred cancel fires
+#   4. Loop exits with stopReason='cancelled'
+```
+
+## Consequences
+
+**What becomes easier:**
+- Writing custom "done" / "stop" / "finish" tools — just call `cancel(msg, after_current_tools=True)`
+- Understanding the stop mechanism — it's one method with an option, not hidden flags
+- Maintaining the stop tool — no hook installation, no WeakSet tracking, no marker keys
+
+**What becomes harder or riskier:**
+- `cancel()` now has two modes, which adds a small conceptual burden
+- The deferred cancel state must be cleared correctly on invocation boundaries (same as today's flag, just on a different field)
+- If a tool calls `cancel(msg, after_current_tools=True)` outside of a tool execution context (e.g., in a hook), the deferred cancel just fires at the next post-batch check — which might be confusing if there are no tools running
+
+**Migration:**
+- The existing `stop_event_loop` check in Python can be replaced with a `_deferred_cancel` check at the same location
+- The TypeScript `endTurn` path can be replaced similarly
+- The stop tool's tests simplify to verifying that `cancel()` is called with the right arguments
+
+## Willingness to Implement
+
+Yes — the implementation is straightforward and localized to the agent class, event loop post-batch checks, and the stop tool itself.


mehtarac:
@strandly-the-agent do a fresh review of this PR

mehtarac (commented):
Is this ready for review? just curious

team/designs/0017-deferred-cancel.md
```diff
@@ -0,0 +1,229 @@
...
+There are two termination primitives in the SDK today:
+
+1. **`agent.cancel()`** — an immediate abort signal. Sets a flag (Python `threading.Event` / TypeScript `AbortController`) that the event loop checks at multiple points:
+   - Before tool execution (per-tool in sequential executor)
+   - During model streaming
+   - After tool execution, before the next model call
+
+   When triggered, pending tools in the batch receive error results ("Tool execution cancelled") and the loop exits with `stopReason: 'cancelled'`. This is designed for external callers (timeouts, user-initiated abort, web request disconnects).
+
+2. **`invocation_state` flags** — the `stop_event_loop` flag (Python) and `AfterToolsEvent.endTurn` marker (TypeScript) are checked only after the entire tool batch completes. They provide cooperative semantics: all tools finish, then the loop exits. However, they require tools to write to `invocationState` (a mutable shared bag) and the event loop to check a one-off key.
```
strandly-the-agent, line 24: 🔴 **Stale — and it's the comparison the whole Proposal rests on.** Python shipped an equivalent primitive three days after this doc's date.

#3508 (`fd1b985b`, merged 2026-07-30) added `AfterToolsEvent.end_turn` to Python (`hooks/events.py:173-199`, consumed at `event_loop.py:907-923`): hook-settable, batch-scoped, synthesizes the final assistant text, yields `stop_reason="end_turn"` — the same shape TS already had. So L24's split ("`stop_event_loop` (Python) and `AfterToolsEvent.endTurn` marker (TypeScript)") no longer describes `main`, and "two termination primitives" (L15) is now three.

That's not cosmetic: it puts a materially cheaper option on the table that the Alternatives section never weighs — **have the stop tool set `end_turn` directly, in both languages**. No `cancel()` overload, no new agent-level state, no `stopReason` change, and the `WeakSet` bridge disappears anyway. It also narrows the real remaining problem from "unify two divergent primitives" to "discoverability for tool authors", which is a smaller and more honest framing.

Would you add that as a third alternative and make the Recommended section beat it explicitly? Note this supersedes rather than repeats the earlier thread on L54: the overload objection *does* have a good answer (`cancel: bool | str` is already an SDK-wide idiom — `hooks/events.py:63`, `hooks/events.ts:149/258/393/744`), and answering it is what exposes the real defect one layer down at L50.

team/designs/0017-deferred-cancel.md
```diff
@@ -0,0 +1,229 @@
...
+- Cooperative semantics: sibling tools in the same batch run to completion
+- The cancel message flows through to the `AgentResult` as the final assistant text
+- Cross-SDK parity (Python and TypeScript)
+- Remove the need for `stop_event_loop` / `STOP_INVOCATION_STATE_KEY` internal flags
+- Composable: multiple tools calling deferred cancel in the same batch produces one stop (last message wins, or first — pick a policy)
+
+**Non-Goals:**
+- Changing the behavior of immediate `cancel()` (no deferred flag) — it stays an aggressive abort
+- Providing a way to "un-cancel" a deferred cancel from another tool in the same batch
+- Cancellation with custom `stopReason` values (it remains `'cancelled'`)
```
strandly-the-agent, line 50: 🔴 **This Non-Goal is the crux, and it breaks live code.** A graceful, model-requested, message-carrying stop reporting `'cancelled'` contradicts what the SDK already does everywhere else:

- `agent.py:1299-1308` (Py) and `agent.ts:1112-1122` (TS): a hook that cancels *with a message* synthesizes an assistant message and yields **`end_turn`** — the closest existing sibling to this proposal.
- `hooks/events.ts:786-793` documents the exact boundary this erases: a string that becomes *literal assistant content* is an `endTurn` outcome; a string on a `cancel` field is a *cancellation reason*.

Concrete breakage if the stop tool starts reporting `cancelled`:

| Consumer | Effect |
|---|---|
| `agent-delegation.ts:323` (`stopReason !== 'endTurn'`) | a stop tool inside a delegating agent silently skips the delegation result transform (contract at `:296-299`) |
| `tracer.py:458,473,785,801`, `tracer.ts:1163,1198` | `stop_reason` → OTel `finish_reason` verbatim, so every successful completion records `finish_reason=cancelled` |
| existing stop-tool users | `tool_use` → `cancelled` (Py) and `endTurn` → `cancelled` (TS), **with no action taken** — which `DECISIONS.md:120` explicitly carves out of pay-for-play |
| the tool's own shipped docs | `stop/README.md:7,9`: final `stopReason` is `endTurn`, and "This is a cooperative stop, **not an abort**. … For hard cancellation, use `agent.cancel()`" — same claim in the model-facing description at `stop.py:44` / `types.ts:19` |

And `a2a/_converters.py:20` already maps `TaskState.canceled → "end_turn"`.

The stated rationale ("no custom `stopReason` values") doesn't hold either — `types/event_loop.py:39-51` already carries `checkpoint`, `interrupt`, `limit_turns`, `limit_output_tokens`, `limit_total_tokens`, none of them model-produced. And no *new* value is needed here: `end_turn` exists and already means this.

What would it cost to keep the proposed API shape but report `end_turn` — i.e. implement the deferred stop on the existing `end_turn` batch primitive rather than routing it through the abort path?

team/designs/0017-deferred-cancel.md
```diff
@@ -0,0 +1,229 @@
...
+- **`stop_event_loop` / `STOP_INVOCATION_STATE_KEY` are undocumented internal state.** They only exist to serve the stop tool but live in the general-purpose `invocationState` bag, polluting it. TypeScript additionally needed a `WeakSet`-tracked hook installation pattern to bridge from `invocationState` to `endTurn`.
+- **No way for a tool to say "stop after the batch" without touching internals.** A user writing a custom "done" tool must reverse-engineer the same flag pattern.
+- **`cancel()` lacks gradation.** It's all-or-nothing — there's no way to cancel cooperatively through the public API.
+
+## Goals and Non-Goals
+
+**Goals:**
+- A single public API for tool-initiated graceful stop: `agent.cancel(message, { afterCurrentTools: true })`
+- Cooperative semantics: sibling tools in the same batch run to completion
+- The cancel message flows through to the `AgentResult` as the final assistant text
```
strandly-the-agent, line 42: 🔴 **This Goal isn't delivered by the mechanism the Proposal itself describes** — and the doc's own DX block contradicts it 130 lines later.

Python (L69-72) resolves the deferred cancel by setting the immediate signal and letting the existing check fire. That check is `event_loop.py:935-941`, the only post-batch `cancelled` site:

```python
if agent._cancel_signal.is_set():
    yield EventLoopStopEvent("cancelled", message, ...)
```

`message` is the cycle's model turn (often nothing but `toolUse` blocks), passed through unchanged. Nothing reads `_deferred_cancel_message`; nothing is appended to history. The stop text lands only as the `toolResult` — exactly what L176 says ("`result.message` contains the model's final assistant message"), which is the accurate line and the one that contradicts this Goal.

TypeScript doesn't get there either: the abort round-trip lands in the `CancelledError` catch, whose text is hardcoded `'Cancelled by user'` (`agent.ts:1727`, `:1764`), so the model's message is discarded.

So either (a) name the actual edit — build the final assistant text from `_deferred_cancel_message`, which is precisely what the `end_turn` path already does at `event_loop.py:908-923` — and commit to a TS path that preserves it, or (b) drop this Goal and say plainly that the message only ever surfaces as a `toolResult`. Worth noting (a) changes what *immediate* `cancel()` puts in history too, which brushes against Non-Goal 1.

(Supersedes the earlier L94 thread: those two TS paths aren't unresolved musing — they produce observably different `AgentResult`s.)

team/designs/0017-deferred-cancel.md
```diff
@@ -0,0 +1,229 @@
...
+
+**What becomes easier:**
+- Writing custom "done" / "stop" / "finish" tools — just call `cancel(msg, after_current_tools=True)`
+- Understanding the stop mechanism — it's one method with an option, not hidden flags
+- Maintaining the stop tool — no hook installation, no WeakSet tracking, no marker keys
+
+**What becomes harder or riskier:**
+- `cancel()` now has two modes, which adds a small conceptual burden
+- The deferred cancel state must be cleared correctly on invocation boundaries (same as today's flag, just on a different field)
+- If a tool calls `cancel(msg, after_current_tools=True)` outside of a tool execution context (e.g., in a hook), the deferred cancel just fires at the next post-batch check — which might be confusing if there are no tools running
```
strandly-the-agent, line 220: 🔴 **The analogy here is false, and it hides a cross-invocation leak.** Today's `stop_event_loop` isn't "cleared" — it lives in `invocation_state["request_state"]`, which is *recreated from scratch every invocation* (`event_loop.py:239-240`), so it cannot survive one. The proposed `_deferred_cancel` / `_deferredCancel` (L65-66, L85-86) is state on the **agent object**, which persists until something explicitly clears it — and the doc names no clearing point.

That matters because the post-batch check this hooks into (L69) is one of several `return`s in the same region, and at least three others return *earlier*, never reaching it:

- interrupt → `event_loop.py:874-888`
- hook-set `end_turn` → `:908-923`
- `stop_event_loop` / structured output → `:925-933`

plus an exception after `AfterToolsEvent`, and a consumer that simply stops iterating `stream_async` (a disconnected streaming client). Any of those leaves the flag armed on a reused agent, and the *next*, unrelated `invoke_async()` stops early with a stale message — verified on a scratch implementation of this design, in both SDKs.

The obvious mitigation (clear on invocation boundaries) is also a policy decision, not a detail: it converts every one of those paths into a **silently dropped stop request** instead. The two SDKs would also naturally clear at different moments — Python at the *end* of an invocation (`agent.py:1266`), TS at the *start* of each `stream()` iteration (`agent.ts:1097`, `:1293`) — so "cleared correctly" needs to be spelled out per SDK to keep the parity Goal.

Suggested: name the clearing point explicitly, and record in `## Key Decisions` which bypass paths *consume* a pending deferred cancel and which *drop* it.

team/designs/0017-deferred-cancel.md
```diff
@@ -0,0 +1,229 @@
...
+   - During model streaming
+   - After tool execution, before the next model call
+
+   When triggered, pending tools in the batch receive error results ("Tool execution cancelled") and the loop exits with `stopReason: 'cancelled'`. This is designed for external callers (timeouts, user-initiated abort, web request disconnects).
+
+2. **`invocation_state` flags** — the `stop_event_loop` flag (Python) and `AfterToolsEvent.endTurn` marker (TypeScript) are checked only after the entire tool batch completes. They provide cooperative semantics: all tools finish, then the loop exits. However, they require tools to write to `invocationState` (a mutable shared bag) and the event loop to check a one-off key.
+
+Neither is right for a tool-initiated graceful stop:
+
+- `cancel()` is too aggressive — it aborts sibling tools mid-batch and uses pre-tool-execution checkpoints that break cooperative guarantees.
```
strandly-the-agent, line 28: 🟡 **This premise is false for the default executor**, which weakens the case for the whole design.

`ConcurrentToolExecutor` is the default (`agent.py:491`) and dispatches the whole batch up front — `grep -c _cancel_signal strands-py/src/strands/tools/executors/concurrent.py` returns **0**. Only `SequentialToolExecutor` pre-checks per tool (`sequential.py:51`). So for an `Agent(...)` built without an explicit `tool_executor`, today's `cancel()` *already* lets siblings run to completion, and a tool calling plain `cancel()` produces the same observable result the deferred flag would.

The deltas that *are* real, and that would make a much stronger Problem statement, are narrower: sequential batches where `stop` isn't the last tool; an in-flight sibling **MCP** call, which an immediate cancel does abort (`mcp_agent_tool.py:118`); and hook-requested **retries** of siblings, which an immediate cancel vetoes (`tools/executors/_executor.py:106`) and a deferred one wouldn't.

```suggestion
- `cancel()` is too aggressive under the sequential executor — not-yet-started siblings get error results instead of running (`sequential.py:51`). Regardless of executor, it also aborts an in-flight MCP tool call (`mcp_agent_tool.py:118`) and vetoes hook-requested sibling retries (`_executor.py:106`). Note the default `ConcurrentToolExecutor` has no per-tool cancel check, so today's `cancel()` already lets concurrent siblings finish.
```

This supersedes the earlier L208 thread with a sharper claim: it's not only that the examples assume sequential, it's that the Problem statement needs the qualifier. The L199-208 walkthrough should also say which executor it assumes.

team/designs/0017-deferred-cancel.md
```diff
@@ -0,0 +1,229 @@
...
+# result.message contains the model's final assistant message
+```
+
+### Custom stop tool
+
+```python
+from strands.tools.decorator import tool
+from strands.types.tools import ToolContext
+
+@tool
```
strandly-the-agent, line 185: 🟡 **This example doesn't run.** I copied it against `strands-py` on `main`:

```
ValueError: @tool(context) must be set if passing in ToolContext param
```

`tools/decorator.py:176-186` requires the flag whenever a parameter is annotated `ToolContext`, and it raises at decoration time — so a reader who copy-pastes the doc's flagship custom-tool recipe gets an immediate error. Unrelated to the proposed API; wrong against today's decorator.

```suggestion
@tool(context=True)
```

strandly-the-agent (commented):
**Fresh review of `c5ea838b` — verdict: request changes.** No new commits since the 2026-07-27 pass, so this is a re-read of the same doc against a `main` that has moved on — which turned out to be the most important finding. The mechanism is sound; the destination is wrong.

- 🔴 **The Current State is stale.** Python shipped `AfterToolsEvent.end_turn` on 2026-07-30 (#3508, `fd1b985b`) — the same primitive this doc calls TypeScript-only. It makes a third, cheaper alternative available that the Proposal never weighs. (L15-24)
- 🔴 **`'cancelled'` is the wrong outcome name,** and it breaks live code: TS delegation gates on `stopReason !== 'endTurn'` (`agent-delegation.ts:323`), and `stop_reason` maps straight to OTel `finish_reason`. The stop tool's own README says "cooperative stop, **not an abort**". (L50)
- 🔴 **Goal 3 isn't delivered by the mechanism the doc specifies.** The cancel branch passes the model's message through unchanged and never reads `_deferred_cancel_message` — the doc's own DX comment on L176 describes the real behaviour. (L42)
- 🔴 **The new flag is agent-level with no clearing point,** and three in-loop paths `return` before the check it hooks into, so a pending stop can leak into the next, unrelated invocation. (L219)

Plus 2 🟡 inline (the Problem premise is false under the *default* executor; the custom-stop-tool example doesn't run) and 5 blocking questions below. The cheap version of all of this: keep `cancel(message, after_current_tools=True)` as the façade, implement it on the existing `end_turn` batch primitive, and report `end_turn`.

<details>
<summary><b>What was actually verified (✅ = I ran or read it on this tree)</b></summary>

Reviewed `c5ea838b` against `origin/main` `cf4a68aa` (the PR's merge base `2828e442` is ~2 weeks behind).

- ✅ `event_loop.py:907-923` — Python's `AfterToolsEvent.end_turn`: synthesizes an assistant text message and yields `stop_reason="end_turn"`. Runs **before** `stop_event_loop` (`:925`) and before the cancel check (`:935-941`).
- ✅ `event_loop.py:935-941` — the only `stop_reason == "cancelled"` site; yields the cycle's `message` unchanged, appends nothing.
- ✅ `agent.py:491` — default executor is `ConcurrentToolExecutor`; `grep -c _cancel_signal concurrent.py` → **0**, vs `sequential.py:51`.
- ✅ `agent-delegation.ts:323`, `tracer.py:458,473,785,801`, `a2a/_converters.py:20` (maps `TaskState.canceled → "end_turn"`), `agent.ts:1727`/`:1764` (hardcoded `'Cancelled by user'`), `agent.ts:1010` (idle `cancel()` is a documented no-op).
- ✅ `agent.py:615-645` — current signature is `def cancel(self) -> None` with **no parameters**; docstring lists four checkpoints including "During MCP tool execution" and promises "Multiple calls to cancel() are safe and idempotent".
- ✅ **Ran the doc's L185-189 example** against `strands-py` on main → `ValueError: @tool(context) must be set if passing in ToolContext param`.
- 🔶 Reasoned-only: the interrupt/session-snapshot divergence and the concurrent-message race were demonstrated on a scratch implementation of this design, not on shipped code (there is none yet) — treat them as design questions, not bugs.
</details>

<details>
<summary><b>Blocking questions (5)</b></summary>

1. **Bidi.** Goal L44 says "remove the need for `stop_event_loop`" unqualified, but `bidi/agent/loop.py:414-429` reads that key — with a live `DeprecationWarning` pointing users *at* it — and `BidiAgent` has no `cancel()`/`_cancel_signal` to hang `after_current_tools` off. Is bidi in scope? If not, what do bidi tool authors migrate to?
2. **Which TS path?** The two the doc offers (L89-94) are not equivalent. On a batch mixing `stop` with structured output, the abort round-trip swallows the cancel entirely, while the "simpler path" returns `cancelled` and discards an already-validated structured output. Which is intended, and does Python match it?
3. **Composability (L45)** reads as a naming choice, but under the default concurrent executor a single unsynchronised message slot is written from concurrent tasks — "last wins" resolves to *last to finish*, which the model doesn't control. Does this need an ordering rule defined against the `toolUse` list rather than prose?
4. **Precedence.** A hook-set `end_turn` (`event_loop.py:907`) now runs before the checkpoint this design hooks into (`:925`). Is `end_turn` always meant to outrank a tool's deferred stop in the same batch — and is TS symmetric about it? (Delegation is a real second writer: `agent-delegation.ts:224`.)
5. **HITL/resume.** If a deferred cancel and an interrupt land in the same batch, the interrupt path returns first and the flag isn't in `_InterruptState` or `take_snapshot(preset='session')`. Resuming the same live object vs. one rebuilt from a snapshot would then disagree on `stop_reason`. Should that state survive a resume? "Yes" is a schema change worth scoping; "no" is worth stating.
</details>

<details>
<summary><b>Structural + process (non-blocking)</b></summary>

- **Missing `## Key Decisions`** — required by `team/designs/README.md:68`, and this review is mostly "which policy did you pick?", which is exactly what that section is for. Candidates: the chosen `stop_reason` and why; first/last-message-wins and whether it needs synchronisation; the single TS path; `end_turn`-vs-deferred-cancel precedence; whether the flag belongs in the interrupt/snapshot schema; bidi's story. The doc also merges the template's separate `## Goals` / `## Non-Goals`.
- **API bar-raising**: extending public `cancel()` in both SDKs plus changing a result-field contract reads as **Substantial** (`team/API_BAR_RAISING.md:24`) — ≥2 API reviewers rather than a solo approve — and the PR carries no `api/*` label, which CI gates at merge (`API_BAR_RAISING.md:72`).
- **PR framing**: the title (`feat(stop):`) and body describe completed two-SDK code with test counts, but the diff is one design doc. For the record this is a force-push artifact, not an overstated claim — `4dfca1ea` (13 minutes before the force-push, same author) really does contain that code. Worth retitling to the commit's own `docs(design): …` and swapping `enhancement`/`area-tool` for `documentation`, so the PR list shows what this is.
- `Issue: TBD` — worth filling, but not a gate: `0011-memory-manager.md` and `0014-storage.md` both merged with `TBD`.
</details>

<details>
<summary><b>Appendix — non-blocking observations (7)</b></summary>

- `agent.py:644` ("Multiple calls to cancel() are safe and idempotent") stops being true once `message` is added, absent a first/last-wins rule.
- `checkpoint.py:19-21` / `checkpoint.ts:24-26` ("Cancel takes precedence over checkpoint") would silently start covering graceful stops — owes a docs edit at implementation time.
- Idle-armed deferred cancel: neither L126 nor L220 is accurate. Python leaves it latent until the next *tool batch* (not the next invocation); TS's idle `cancel()` is a documented no-op (`agent.ts:1010`), so the deferred variant must either break that contract or diverge from Python.
- `message` on the **immediate** path is unspecified for Python: L76 says it's stored "for use in cancellation text", but Python's cancel path synthesizes no text while TS does. Silently ignored, or a behaviour change for every existing `cancel()` caller?
- The Current State checkpoint list (L17-20) matches neither SDK's own `cancel()` docstring and drops Python's documented MCP checkpoint (`mcp_agent_tool.py:118`).
- L33's "undocumented internal state" is right about Python and wrong about TypeScript, where the key is `@internal`-tagged with an explicit "don't depend on the literal value" (`stop/types.ts:29-37`). The Python key, by contrast, appears in four site pages including a copy-paste custom-stop-tool recipe.
- L225 ("the stop tool's tests simplify to verifying that `cancel()` is called with the right arguments") points the wrong way — that assertion is precisely the one that would let the Goal-3 and leak problems above ship green.
- Pre-existing, **not** this PR's problem and not filed: Python's stop path leaves history ending on a `user` toolResult, so the next turn appends a second consecutive `user` message. Reproduces identically on `main`, and only observed against a mock model.
</details>

<details>
<summary><b>How this review ran</b></summary>

Five independent fresh-context passes — routing triage, context build, API/DevX bar-raising, adversarial, issue-alignment, docs-accuracy — then de-duplicated and severity-ranked under a comment budget (~35 raw findings → 6 inline). The adversarial pass had no code to attack, so it implemented design 0017 verbatim on throwaway copies of both SDKs and ran it; that's where the leak, the message race, and the two-TS-paths divergence come from. Nothing was written to this branch. I'm an AI agent — this is work for a human to approve, and I'd flag the `end_turn`-vs-`cancelled` decision (inline #2) as the one worth a maintainer's own judgment rather than mine.
</details>

Unshure:
Updating this proposal after digging into a second use case that I think reshapes the direction.

**Context: issue #3808** (delegation corrupts the `MessageAddedEvent` stream). `AgentDelegation` ends the parent turn after a delegate tool returns, but it can only hand the loop an `end_turn` *placeholder string* — so the real delegate content gets patched in afterward in middleware, which double-fires `MessageAddedEvent` without a session manager and suppresses it with one. Root cause: the loop's stop primitive can't carry the actual final message.

That's the same gap this PR is circling, and it reinforces @strandly-the-agent's point that the cooperative stop should build on the existing `end_turn` batch primitive (reporting `end_turn`, not `cancelled`) rather than abort machinery. Delegation is a *successful* terminal — routing it through `cancelled` would trip the `stopReason !== 'endTurn'` gate and mislabel OTel `finish_reason`.

**Revised framing — one primitive, one honest public API:**

1. **Widen the internal primitive.** `AfterToolsEvent.end_turn: bool | str` → also accept a message / content blocks, so the loop appends the real final assistant message *once* (one `MessageAddedEvent`, one persist). This alone fixes #3808 and drops the placeholder + `redact_latest_message` special-casing.
2. **Public façade `agent.stop(message=…, reason=…, immediate=?)`** rather than expanding `cancel()`. `cancel` implies abort; `stop` honestly covers both a graceful cooperative stop (default: finish current tools, `stop_reason=end_turn`) and a hard abort (`immediate=True`, `stop_reason=cancelled`). `cancel()` becomes a deprecated alias for `stop(immediate=True, reason="cancelled")`.
3. **`agent.stop()` as the single developer surface** for stopping from a tool or *any* hook — today only `AfterToolsEvent` exposes `end_turn`, so tools and other hooks have no clean mechanism. Internally it lowers onto the same batch-boundary path as (1); SDK-internal plugins like delegation keep setting the `end_turn` field directly since they're already at the boundary.

**Open decisions:**
- Default urgency of a bare `stop()` — graceful is safer but inverts today's `cancel()` default (migration, not rename).
- `reason` constrained to the valid stop-reason enum, not free-form (OTel + TS delegation gating consume it).
- `message` must carry content blocks, not just text — shared with (1).
- Keep vs. deprecate the recently-shipped public `AfterToolsEvent.end_turn` field — I lean deprecate-in-favor-of-`stop()`.
- Cross-SDK parity: public agent API + hook field in both SDKs.

I'll re-scope this design around strengthened `end_turn` + an `agent.stop()` façade, with #3808 as a ready second consumer that de-risks the primitive.

strandly-the-agent:
The revised direction checks out against current `main` (`652ac971`) — I verified #3808's mechanics before replying: the delegation module's own docstring admits the phantom-turn problem (`_agent_delegation.py:11`), the session-manager fork is exactly the special case you describe (`:358-370`), and `end_turn` today synthesizes a text-only message (`event_loop.py:909-919`, `events.py:196`). Widening the primitive fixes the root cause and gives this PR its mechanism in one move; delegation as a second in-tree consumer is what was missing from the original doc.

One heads-up for the rewrite: **most of my review's blockers were about the batch-boundary state machine, not about `cancel()`** — renaming the façade to `stop()` carries them over intact. The four that need a home in the new doc: where the pending-stop state lives and what clears it (the interrupt path at `event_loop.py:875-889` still returns before the `end_turn` branch, and a consumer that stops iterating still strands agent-level state); the concurrent-executor message race; HITL snapshot/resume; and whether `BidiAgent` gets `stop()` (which would finally answer the `stop_event_loop` migration question).

<details>
<summary><b>Input on the five open decisions</b></summary>

1. **Bare `stop()` default: graceful is right**, and it's not really an inversion if `cancel()` stays as the immediate alias — no existing code changes behaviour without action, which keeps it inside `DECISIONS.md`'s pay-for-play line. The risk is doc-level, not migration-level.
2. **`reason`: I'd go further than "constrain to the enum" — derive it from `immediate` and don't accept it at all** (graceful → `end_turn`, immediate → `cancelled`). Accepting even a constrained enum lets user code impersonate loop states other subsystems key off (`interrupt`, `checkpoint`, `limit_*` — `types/event_loop.py:39-51`), and both consumers you name (OTel, the TS delegation gate at `agent-delegation.ts:323`) are exactly why a tool shouldn't get to pick. If a third value ever earns a use case, add the param then. If you do keep it, the `immediate×reason` matrix (`stop(immediate=True, reason="end_turn")`?) needs to be defined or forbidden explicitly.
3. **`message` as content blocks: agree — and note #3808's own proposed fix is the opposite shape.** The issue proposes a `SUPPRESS_MESSAGE` sentinel ("stop, append *nothing*", middleware stays the appender); your (1) makes the *loop* the single appender. Both fix the double-fire, but they're mutually exclusive — exactly one party may own the append. Making the loop the appender is the stronger design (it also serves the stop tool and kills `redact_latest_message` special-casing), but it requires delegation to have final content in hand at `AfterToolsEvent` time — worth stating in Key Decisions so the sentinel option is explicitly rejected rather than just unmentioned.
4. **Keep-vs-deprecate `AfterToolsEvent.end_turn`: your points 3 and 4 currently conflict** — (3) says SDK-internal plugins keep setting the field directly, (4) leans deprecate. The coherent version is "public-deprecated, retained as the internal lowering target", but then two more things need defining: what happens when a hook calls `agent.stop()` *during* `AfterToolsEvent` (the branch that consumes it has already fired — same-batch or next-batch?), and whether `BeforeToolsEvent.cancel: bool | str` folds into `stop()` too or stays as a third mechanism.
5. **Parity: two pre-existing divergences will bite `stop()` on day one**, so spec them per SDK rather than inheriting them: idle-call behaviour (TS `cancel()` is a documented no-op via the `_isInvoking` gate, `agent.ts:859`; Python arms the next invocation), and immediate-path message text (Python's cancel branch synthesizes nothing; TS hardcodes `'Cancelled by user'`, `agent.ts:1727/:1764`). `stop(immediate=True, message=…)` is actually the first good answer to that hardcoded string — but making Python's immediate path carry a message is a behaviour change to the existing `cancel` branch, so say so.

Minor: `agent.stop()` next to the experimental `stop` *tool* is a naming collision worth one disambiguating line in the docs — presumably the tool becomes a thin wrapper over `agent.stop()`.
</details>

Happy to re-review when the re-scoped doc is pushed — with delegation and the stop tool as two named consumers of one primitive, plus a `## Key Decisions` section covering the carry-overs above, I'd expect the remaining findings to close cleanly.

Issue #3498, opened 2026-07-27
[PORT] BeforeToolsEvent + AfterToolsEvent batch hooks


liramon2:
### Source SDK

strands-ts

### Target SDK

strands-py

### Related Files

```text
strands-ts/src/hooks/events.ts
strands-ts/src/hooks/index.ts
strands-ts/src/hooks/__tests__/events.test.ts
strands-ts/src/agent/agent.ts
strands-ts/src/types/agent.ts
strands-ts/src/index.ts
strands-ts/src/__fixtures__/mock-plugin.ts
strands-ts/src/agent/__tests__/agent.test.ts
strands-ts/src/agent/__tests__/agent.hook.test.ts
strands-ts/src/agent/__tests__/agent.interrupt.test.ts
strands-ts/src/agent/__tests__/agent-delegation.test.ts
site/src/content/docs/user-guide/concepts/agents/hooks.mdx
site/src/content/docs/user-guide/concepts/interrupts.mdx
site/src/content/docs/user-guide/concepts/streaming/async-iterators.mdx
site/src/content/docs/user-guide/concepts/streaming/index.mdx
site/src/content/docs/user-guide/concepts/tools/executors.mdx
```

## What needs to be migrated

TypeScript has a batch-level `BeforeToolsEvent` hook that fires once before any tools in a batch execute; Python only has per-tool `BeforeToolCallEvent`. The `BeforeToolsEvent` gives hooks a single point to inspect the full set of requested tool calls and cancel the entire batch, and supports interrupts for human-in-the-loop approval of tool batches.

### Event definition (`strands-ts/src/hooks/events.ts:733`)

• `BeforeToolsEvent` extends `HookableEvent` and implements `Interruptible`.
• Readonly fields: `agent` (`:736`), `message` (assistant-role `Message` containing all `ToolUseBlock`s, `:737`), `invocationState` (`:738`).
• Mutable control field: `cancel: boolean | string = false` (`:746`) — when `true`, all tools are cancelled with a default error message; when string, that string is used as the error text in each `ToolResultBlock`.
• Does NOT use reverse callback ordering (`:710` returns `false`).
• Implements `interrupt()` method (`:755`) for human-in-the-loop workflows — halts the agent and stores pending tool execution state for resume.

### Agent loop integration (`strands-ts/src/agent/agent.ts:2192-2228`)

• `executeTools()` constructs and yields `BeforeToolsEvent` before any tool execution (`:2192-2195`).
[...]

PR #3508, opened 2026-07-27
feat(python): add BeforeToolsEvent and AfterToolsEvent batch hooks

liramon2:
> Opened by **Strandly** (an AI agent) on behalf of a maintainer, as a cross-SDK port. A human should review before merge — see the note at the bottom.

## Description

Ports the batch-level tool hooks **`BeforeToolsEvent`** and **`AfterToolsEvent`** from the TypeScript SDK (`strands-ts`) to Python (`strands-py`). These are two halves of one lifecycle pair (both live in the source's `executeTools` and share a `finally`-block relationship), edit the same `_handle_tool_execution` region, and are combined here as a single logical change.

Python previously had only the per-tool `BeforeToolCallEvent` / `AfterToolCallEvent`. The batch hooks give hooks a single point around the *whole* set of tool calls in an assistant message.

**`BeforeToolsEvent`** — fires once before any per-tool hook or execution.
- Dataclass: interruptible (`_Interruptible`), mutable `cancel: bool | str = False` (only `cancel` writable), forward callback order, interrupt id `v1:before_tools:<uuid5(name)>`.
- **Batch cancel** (`event.cancel`): one error `ToolResult` per requested tool use (`"Tool cancelled by hook"` for `True`, else the string), skips the executor and all per-tool hooks. Hook cancel takes precedence over agent cancel.
- **Batch interrupt**: stores pending state and stops; on resume the hook re-runs and, if not interrupted again, tools execute without re-calling the model.

**`AfterToolsEvent`** — fires once after the batch finishes.
- Dataclass: reverse callback order, mutable `end_turn: bool | str = False` (only `end_turn` writable).
- **Always fires** (mirrors the source `finally`): on normal completion, batch cancel, per-tool interrupt, and tool/hook exceptions — but **not** on a `BeforeToolsEvent` interrupt or hook exception (which short-circuit before the batch).
- **`end_turn`**: halts the loop after tool execution without calling the model again, appending a final assistant message (`True` → default text; string → that text). Returned result has `stop_reason="end_turn"`. Takes precedence over the recurse/cancel/checkpoint paths, matching the source.

**Bug fix (included):** The `after_model` checkpoint guard only checked `_cancel_signal`, so resuming from any interrupt with `checkpointing=True` re-emitted the `after_model` checkpoint instead of progressing to tool execution. The new `BeforeToolsEvent` interrupt path hits this guard directly (interrupt → resume → checkpoint fires again → tools never run), making it a blocking bug for batch interrupts. The same wedge exists on the pre-existing per-tool interrupt path, but was latent because no test exercised checkpointing + interrupt resume together. Fixed by adding `and not agent._interrupt_state.activated` to the guard.

## Related Issues

Ports #3498.

## Documentation PR

Included in this PR under `site/`: adds both events (and their mutable `cancel` / `end_turn` fields) to the hooks reference and Python lifecycle diagram, documents batch interruption (with a runnable example), batch pre-launch cancellation, and the per-tool-interrupt / always-fires interaction in the interrupts and executors guides.

## Type of Change

New feature

## Testing

The source SDK's `BeforeToolsEvent`/`AfterToolsEvent` tests are the behavioral spec; each was mapped to a Python test. Coverage includes:
- **Event shape** — fields, defaults, callback ordering, read-only vs writable fields, `BeforeToolsEvent` interrupt-id format.
- **Lifecycle/order** — `BeforeToolsEvent` → per-tool `BeforeToolCallEvent`/`AfterToolCallEvent` → `AfterToolsEvent`; `BeforeToolsEvent` fires once per batch.
- **Cancellation** — default/custom/multi-tool batch cancel (parametrized over **sequential and concurrent** executors); invalid tool names included; hook-cancel precedence over agent-cancel; agent-cancel via the batch path.
- **`AfterToolsEvent`** — always-fires on batch cancel, per-tool interrupt, and tool/hook exception; does **not** fire on a `BeforeToolsEvent` interrupt or hook exception; `end_turn` halts without recursion (both executors), works on a cancelled batch, empty-string `end_turn` is falsy; batch message reflects an `AfterToolCallEvent`-mutated result and preserves source order under concurrent execution.
- **Interrupts** — single/multiple `BeforeToolsEvent` interrupts + resume without model re-call, repeated interrupt on a later cycle, resume under checkpointing, invalid-tool result not duplicated on resume.

Commands run (Python 3.10):
- `hatch test tests/strands/event_loop tests/strands/agent/hooks tests/strands/agent/test_agent_hooks.py tests/strands/agent/test_agent_cancellation.py tests/strands/middleware tests/strands/hooks` → all pass except 2 pre-existing `tests/strands/hooks/test_registry.py::*coroutine` failures that also fail on `main` (unrelated to this change; verified by stashing the diff)
- `ruff check` + `mypy ./src` → clean
- End-to-end sanity checks against a mocked-model `Agent`: batch interrupt → approve → tools execute; batch cancel → error results, no execution; `AfterToolsEvent.end_turn` → tool runs, loop halts with `stop_reason="end_turn"`.

- [ ] I ran `hatch run prepare` — ran targeted tests + `ruff` + `mypy` rather than the full matrix; please run the full gate in CI.

### Deliberate divergences (see the source for context)
- **Interrupt ids** use the Python `v1:before_tools:<uuid5>` convention (not the TS `hook:beforeTools:<name>` string), matching the existing `BeforeToolCallEvent` scheme.
- **Pre-cycle agent cancel:** in TS an `agent.cancel()` before the batch is caught in the main loop, so the batch events don't fire; Python has always handled cancel inside `_handle_tool_execution`, so the batch events fire (no observable effect without a batch hook). The resulting `AgentResult` is equivalent.
- **`end_turn` + structured output:** the `end_turn` stop event carries `structured_output` if a structured result was captured in the same batch; the TS `endTurn` result omits it. Only reachable if a hook sets `end_turn` *and* a structured-output tool fired in the same batch — kept for internal consistency rather than forcing a riskier reorder.

## Checklist
- [x] I have added an appropriate example to the documentation to outline the feature

----

🤖 This PR was prepared by Strandly, an AI agent, as a translation of #3498. A maintainer should review it before merge; the "understand every line" attestation reflects the automated review passes run during preparation (correctness + API bar-raising + a port-equivalence validator, each fixed to convergence), not a human sign-off.


site/src/content/docs/user-guide/concepts/agents/hooks.mdx
```diff
@@ -152,11 +152,15 @@ flowchart LR
   end
   subgraph Tool["Tool Events"]
     direction TB
+        BeforeToolsEvent["BeforeToolsEvent"]
```
liramon2, line 155: BeforeToolsEvent and AfterToolsEvent are also missing from the TS mermaid diagram. Add them in so that the Python and TS mermaid diagrams agree.

site/src/content/docs/user-guide/concepts/agents/hooks.mdx
```diff
@@ -294,8 +298,10 @@ Init --> Invocation
 | `MessageAddedEvent`                | Triggered when a message is added to the agent's conversation history                                         |
 | `BeforeModelCallEvent`             | Triggered before the model is invoked for inference                                                           |
 | `AfterModelCallEvent`              | Triggered after model invocation completes. Uses reverse callback ordering                                    |
+| `BeforeToolsEvent`                 | Triggered once before a batch of tools is executed, before any `BeforeToolCallEvent`                          |
```
liramon2: Make the description of BeforeToolsEvent match the TS description verbatim.

site/src/content/docs/user-guide/concepts/agents/hooks.mdx
```diff
@@ -294,8 +298,10 @@ Init --> Invocation
...
 | `BeforeToolCallEvent`              | Triggered before a tool is invoked                                                                            |
 | `AfterToolCallEvent`               | Triggered after tool invocation completes. Uses reverse callback ordering                                     |
+| `AfterToolsEvent`                  | Triggered once after a batch of tools finishes. Always fires (even on cancel). Uses reverse callback ordering |
```
liramon2: Make the description of AfterToolsEvent match the TS description verbatim.

site/src/content/docs/user-guide/concepts/agents/hooks.mdx
```diff
@@ -360,6 +369,9 @@ Most event properties are read-only to prevent unintended modifications. However
     - `retry` - Request a retry of the tool invocation. See [Tool Call Retry](#tool-call-retry).
     - `exception` *(read-only)* - The original exception if the tool raised one, otherwise `None`. See [Exception Handling](#exception-handling).
 
+- [`AfterToolsEvent`](@api/python/strands.hooks.events#AfterToolsEvent)
```
liramon2, line 376: Add an equivalent documentation line for the TS event properties.

site/src/content/docs/user-guide/concepts/interrupts.mdx
```diff
@@ -99,6 +101,59 @@ while True:
 
 print(f"MESSAGE: {json.dumps(result.message)}")
 ```
+
+#### BeforeToolsEvent
+
+```python
```
liramon2, line 107: Make this code snippet align more closely with the TS example `hooks_before_tools`

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -730,69 +731,77 @@ async def _handle_tool_execution(
...
-        cancelled_tool_result_message: Message | None = None
-        if tool_results:
-            _cancelled_msg: Message = {
-                "role": "user",
-                "content": [{"toolResult": result} for result in tool_results],
-            }
-            cancelled_tool_result_message = _cancelled_msg
-            await agent._append_messages(_cancelled_msg)
-            yield ToolResultMessageEvent(message=_cancelled_msg)
+    if interrupts:
```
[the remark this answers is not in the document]
liramon2: Agreed, extract this into a helper.

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -730,75 +731,96 @@ async def _handle_tool_execution(
...
+    tool_uses = [content["toolUse"] for content in message["content"] if "toolUse" in content]
     tool_results: list[ToolResult] = []
-    invalid_tool_use_ids: list[str] = []
-
-    validate_and_prepare_tools(message, tool_uses, tool_results, invalid_tool_use_ids)
-    tool_uses = [tool_use for tool_use in tool_uses if tool_use.get("toolUseId") not in invalid_tool_use_ids]
 
     if agent._interrupt_state.activated:
         tool_results.extend(agent._interrupt_state.context["tool_results"])
+        completed_tool_use_ids = {tool_result["toolUseId"] for tool_result in tool_results}
```
liramon2: What is the point of renaming this variable? If there is none, keep it the same.

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -730,75 +731,96 @@ async def _handle_tool_execution(
...
+    elif agent._cancel_signal.is_set():
+        cancel_message = "Tool execution cancelled"
 
     structured_output_result = None
-    if structured_output_context.is_enabled:
-        if structured_output_result := structured_output_context.extract_result(tool_uses):
-            yield StructuredOutputEvent(structured_output=structured_output_result)
-            structured_output_context.stop_loop = True
+    tool_result_message: Message = {"role": "user", "content": []}
+    after_tools_event = AfterToolsEvent(agent=agent, message=tool_result_message, invocation_state=invocation_state)
```
liramon2: Is this after_tools_event and tool_result_message necessary when they are overwritten in the finally block? If not, remove them

strands-py/tests/strands/event_loop/test_event_loop.py
```diff
@@ -1065,6 +1068,658 @@ async def test_event_loop_cycle_exception_model_hooks(mock_sleep, agent, model,
     )
 
 
+@pytest.mark.asyncio
```
liramon2, line 1071: You are adding many tests in this file. Check if any can be dropped, combined, or simplified without impacting coverage.

liramon2:
Thanks for the review — addressed all the comments in 89511655. Summary:

**`event_loop.py`**
- **Extract the duplicated interrupt-stop block into a helper** — added `_stop_for_interrupts(...)`; both the pre-execution (`BeforeToolsEvent`) and post-execution interrupt paths now call it, so interrupt-persistence changes only have to be made once.
- **`completed_tool_use_ids` rename** — reverted to `tool_use_ids` (no reason to rename).
- **`after_tools_event` / `tool_result_message` pre-init** — removed; both are unconditionally reassigned in the `finally`, so the pre-init was dead. Moved the `Message` annotation onto the `finally` assignment.

**`hooks.mdx`**
- Added `BeforeToolsEvent` / `AfterToolsEvent` to the **TypeScript** mermaid diagram so it matches the Python one.
- Made the Python `BeforeToolsEvent` / `AfterToolsEvent` table descriptions match the TS wording verbatim.
- Added the equivalent TS `AfterToolsEvent.endTurn` writable-property line.

**`interrupts.mdx`**
- Reworked the Python `BeforeToolsEvent` example to mirror the TS `hooks_before_tools` example (filter for dangerous tools, only interrupt when there are any, cancel with `"Batch cancelled by user"`).

**`test_event_loop.py`**
- Replaced the magic positional indexing (`["stop"][4][0]`, `["stop"][6].position`) with named tuple unpacking across the affected tests.
- On the "check if tests can be dropped/combined" note: merged the two `BeforeToolsEvent`-hook-raises tests into one (same code path, complementary assertions — executor-not-called + `AfterToolsEvent`-not-fired). I kept the other new tests as-is; each maps to a distinct source behavior in the traceability matrix, so combining further would have dropped coverage of a distinct case.

Validation: `ruff check` + `ruff format --check` clean, `mypy ./src` clean for `event_loop.py` (remaining errors are pre-existing missing-stub imports, identical on `main`), and the touched suites pass (`tests/strands/event_loop` + `tests/strands/agent/hooks` + cancellation, 296 tests). Coverage of the new helper and both interrupt call sites is intact.

liramon2:

liramon2:

liramon2:
Investigated the `AfterToolsEvent` per-cycle firing finding. **This is not a port defect — it matches the TypeScript SDK.** No behavior change; documented it and added the missing resume-leg test in 98d70213.

### Why it matches TS

`strands-ts` dispatches `AfterToolsEvent` from a `finally` in `executeTools` (`agent.ts:2250-2254`), and both tool executors **re-throw** `InterruptError` (`sequential.ts:69`, `concurrent.ts:128`, `executor.ts:281-282`). A per-tool interrupt therefore unwinds *through* that `finally`, which `yield`s the `AfterToolsEvent` — carrying only the results collected so far — before the error reaches the agent loop's interrupt handler (`agent.ts:1705`). On resume, `executeTools` is called again, so both `BeforeToolsEvent` and `AfterToolsEvent` fire a second time.

I verified the JS generator semantics (`.throw()` + `finally` + `yield` is subtle) with a minimal repro of that exact structure:

<details>
<summary>Node repro output</summary>

```
=== interrupt cycle events ===
  - beforeTools
  - beforeToolCall:A
  - beforeToolCall:B          <- interrupt injected here
  - AFTER_TOOLS(results=["A"])   <- fires with PARTIAL results
  - CAUGHT(outer _stream catch): INTERRUPT
```
</details>

So TS fires it once per cycle → twice across an interrupted batch, partial on the interrupt leg. Python's `finally` dispatch produces exactly the same observable sequence:

| | interrupt cycle | resume cycle |
|---|---|---|
| TS | `AFTER_TOOLS(results=["A"])` — partial | fires again, full set |
| Python | `[["t1"]]` — partial | `[["t1"], ["t1","t2"]]` — fires again, full set |

Both of your observations (1) double-firing and (2) partial `message` on the interrupt cycle are faithful to the source. Per *translate, don't improve*, I left the behavior alone.

### What changed

- **(a) Documented the semantics.** `AfterToolsEvent` docstring now states it fires once per event-loop cycle rather than once per logical batch, that an interrupted batch fires it more than once (so side-effecting hooks should expect that), and the `message` attribute description is corrected — it carries the results collected *in the current cycle*, a partial set on an interrupt cycle. Added the same rule to `interrupts.mdx` in **both** the Python and TypeScript tabs, since the behavior is identical and the guides should agree.
- **(b) Added the resume-leg test.** `test_event_loop_cycle_after_tools_fires_per_cycle_across_per_tool_interrupt_resume` interrupts the second tool of a two-tool batch and asserts the interrupt cycle carries only `["t1"]` and the resume cycle fires again with `["t1", "t2"]` — locking in both the firing count and the partial/full messages.

### Validation

`ruff check` + `ruff format --check` clean. `mypy ./src` unchanged at its pre-existing 42-error baseline (missing optional-dep stubs; identical with my changes stashed), with `events.py` and `event_loop.py` clean. `hatch test tests/strands/event_loop/test_event_loop.py` → **64 passed**; across `event_loop` + `agent/hooks` + `test_agent_hooks` + `test_agent_cancellation` + `hooks` → **267 passed, 2 failed**, both being the pre-existing `test_registry.py::*coroutine` failures (confirmed identical at `HEAD` with my changes stashed).

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -744,91 +793,105 @@ async def _handle_tool_execution(
...
+    finally:
+        # Always pair BeforeToolsEvent with AfterToolsEvent (mirrors the source `finally`
+        # dispatch), so batch-level hooks observe the results even on the cancel, interrupt,
+        # and error paths.
+        tool_result_message: Message = {
+            "role": "user",
+            "content": [{"toolResult": result} for result in tool_results],
+        }
+        after_tools_event = AfterToolsEvent(agent=agent, message=tool_result_message, invocation_state=invocation_state)
+        after_tools_event, _ = await agent.hooks.invoke_callbacks_async(after_tools_event)
```
strandly-the-agent: 🟡 **should-fix — an `AfterToolsEvent` hook exception silently discards an in-flight interrupt.**

This dispatch sits inside the `finally`, *between* "interrupts collected" and "interrupts persisted" (`:877-891`). If a hook callback raises here, the exception propagates and skips `_stop_for_interrupts`, so `_interrupt_state` is never activated — the pending human-approval request is destroyed rather than surfaced.

Reproduced cold at `75aa7cb` (per-tool interrupt + a hook that raises `ValueError`):

```
BASELINE (no AfterToolsEvent hook):
  stop_reason: interrupt   interrupts: ['v1:before_tool_call:t1:225d3982-...']
  interrupt_state.activated: True
WITH THROWING AfterToolsEvent HOOK:
  RAISED: EventLoopException hook blew up
  interrupt_state.activated: False        <-- interrupt lost
then the documented resume path:
  ValueError: Received interrupt responses but agent is not in interrupt state
```

So an HIL approval is unrecoverable — not just delayed. **Calibration:** I checked whether this is new, and it is not a regression — merge-base loses the interrupt identically when a *per-tool* hook raises while another tool interrupts (verified in a base worktree, same `activated: False`). What this PR adds is a second, much more reachable site: `AfterToolsEvent` fires on *every* batch by design, including the interrupt path, so any batch-level hook bug now hits it.

Worth closing the hole while you're in here: persist the interrupt context/`activate()` before (or independently of) the hook dispatch, so a hook bug can't preempt bookkeeping the SDK owns. Happy for this to be a follow-up if you'd rather keep the port faithful — but the pre-existing framing is the reason it's 🟡 and not 🔴.

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -744,91 +793,105 @@ async def _handle_tool_execution(
...
+            validated_tool_uses: list[ToolUse] = []
+            validation_results: list[ToolResult] = []
+            invalid_tool_use_ids: list[str] = []
+            validate_and_prepare_tools(message, validated_tool_uses, validation_results, invalid_tool_use_ids)
+            tool_uses = [
+                tool_use
+                for tool_use in validated_tool_uses
+                if tool_use["toolUseId"] in pending_tool_use_ids and tool_use["toolUseId"] not in invalid_tool_use_ids
+            ]
+            tool_results.extend(result for result in validation_results if result["toolUseId"] in pending_tool_use_ids)
```
strandly-the-agent, line 835: 🟡 **should-fix — this filter is load-bearing and completely untested.** Replacing it with the unfiltered `tool_results.extend(validation_results)` leaves the whole focused suite green at `75aa7cb`:

```
### M9 (drop pending_tool_use_ids filter): 269 passed
```

`test_event_loop_cycle_before_tools_interrupt_invalid_tool_result_not_duplicated` (`test_event_loop.py:1550`) looks like it covers this but can't: a `BeforeToolsEvent`-level interrupt persists no results, so on resume the invalid id is still pending and the filter is a no-op. The filter's actual job needs a **per-tool** interrupt in a batch that *also* contains an invalid tool name — without it, that result is duplicated (`['t2','t2','t1']`).

A test-quality pass wrote and verified the missing case (passes at head, fails with the filter dropped, rest of suite unaffected):

```python
# batch = [valid tool that interrupts, invalid tool name]; interrupt on t1, then resume
result_ids = [c["toolResult"]["toolUseId"] for c in agent.messages[-1]["content"]]
assert result_ids.count("t2") == 1, f"t2 duplicated: {result_ids}"
```

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -744,91 +793,105 @@ async def _handle_tool_execution(
...
-        # End the cycle span before yielding the recursive cycle.
-        if cycle_span:
-            tracer.end_event_loop_cycle_span(span=cycle_span, message=message)
-
+            cycle_start_time,
+            cycle_trace,
+            cycle_span,
+            invocation_state,
+            tracer,
+            structured_output_result,
```
strandly-the-agent, line 879: 🟡 **should-fix — nothing asserts structured output survives an `interrupt` stop.** Deleting this argument (so it defaults to `None`) also leaves the suite green:

```
### M7 (drop structured_output_result on the interrupt stop path): 269 passed
```

User-visible: with a batch of [structured-output tool, interrupting tool], head returns `stop_reason=interrupt` with `structured_output` populated; the mutant returns `None`. The named-tuple unpacking in the interrupt tests discards this slot (`_, _, _, _, interrupts, _, _`), so it's never checked.

One assertion closes it — a batch where `extract_result` captures a result and a per-tool interrupt fires in the same cycle, asserting `AgentResult.structured_output` is non-`None` on the interrupt stop.

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -314,8 +318,8 @@ async def event_loop_cycle(
                 )
 
             if stop_reason == "tool_use":
-                # Emit after_model checkpoint, unless we just resumed from one.
-                if agent._checkpointing and not agent._cancel_signal.is_set():
+                # Emit after_model checkpoint, unless we just resumed from one or an interrupt.
+                if agent._checkpointing and not agent._cancel_signal.is_set() and not agent._interrupt_state.activated:
```
strandly-the-agent, line 322: ⚪ nit / changelog — **this line is a real bugfix, not part of the port, and it's invisible.**

Base had `if agent._checkpointing and not agent._cancel_signal.is_set():` (`58a501c7:...event_loop.py:318`); the added `and not agent._interrupt_state.activated` makes it interrupt-aware. Because `_interrupt_state.activated` is shared with the *pre-existing* per-tool interrupt path, this fixes a latent pre-PR issue: resuming from any interrupt with `checkpointing=True` re-emitted the `after_model` checkpoint instead of progressing. An adversarial pass reproduced the wedge on merge-base (resume re-emitting `after_model` indefinitely, tool never running) and showed head progresses to `after_tools`.

Two small asks: mention it in the PR description so it lands in release notes as a fix, and consider a per-tool-interrupt + `checkpointing=True` resume test — mutation testing showed the clause is currently killed *only* by the new `BeforeToolsEvent` checkpoint test, not by the path it actually repairs.

strands-py/src/strands/hooks/events.py
```diff
@@ -133,6 +133,84 @@ class MessageAddedEvent(HookEvent):
...
+            name: User-defined name for the interrupt.
+
+        Returns:
+            Interrupt id.
+        """
+        return f"v1:before_tools:{uuid.uuid5(uuid.NAMESPACE_OID, name)}"
+
+
+@dataclass
+class AfterToolsEvent(HookEvent):
```
strandly-the-agent, line 173: 🟡 design question (non-blocking, but cheapest to answer now) — **should `AfterToolsEvent` carry an outcome channel before this ships?**

The docstring promises hooks visibility "including on cancel and error paths", but the event exposes only `agent`, `message`, `invocation_state`, `end_turn` — so a hook can't tell *how* the batch ended. Its per-tool sibling has both `exception: Exception | None` and `cancel_message: str | None` (`events.py:281-283`). Today an audit/metrics hook has to sniff `status == "error"` and string-match the cancel text — a contract you can never change once hooks ship against it.

Related observation from the same area: `end_turn` set during an interrupt cycle is **silently discarded** — `:877-891` returns before the `:905` check ever reads it, with no signal to the hook. Intended, deferred, or a gap?

And neither batch event exposes `tool_uses`/`tool_results` in typed form (siblings do), which is why your own doc example (`interrupts.mdx:130-134`) does raw dict-spelunking with no IDE/type-checker help. Adding fields later is additive, so this isn't blocking — but every hook written meanwhile bakes in the workaround. Would a `cancel_message: str | None = None` (mirroring `AfterToolCallEvent`) plus typed `tool_results` be worth it now?

strands-py/src/strands/hooks/events.py
```diff
@@ -133,6 +133,84 @@ class MessageAddedEvent(HookEvent):
     message: Message
 
 
+@dataclass
+class BeforeToolsEvent(HookEvent, _Interruptible):
+    """Event triggered before a batch of tools is executed.
+
+    This event is fired once for each assistant message containing tool use
```
strandly-the-agent: ⚪ nit — **`BeforeToolsEvent`'s docstring now contradicts the one you just fixed below it.** `98d70213` correctly documented that `AfterToolsEvent` fires once per event-loop cycle, but this still claims it's fired "once for each assistant message containing tool use requests, before any individual tool callbacks or execution begin."

Both halves are wrong on the same resume path: I ran a 2-tool batch interrupting on the second tool and `BeforeToolsEvent` fired **twice** for one assistant message — and on the resume cycle `t1` has already executed and its `AfterToolCallEvent` has already run, so "before any individual tool callbacks or execution begin" doesn't hold either.

Suggest mirroring the wording you gave `AfterToolsEvent`, e.g. "fired at the start of each event-loop cycle that has pending tool-use requests — a per-tool interrupt splits a batch across cycles, so it can fire more than once for one assistant message."

While there: `strands-py/docs/HOOKS.md:19` ("For every `Before` event there is a corresponding `After` event, even if an exception occurs") is now false for two pairs — a hook raising/interrupting on the `Before` event short-circuits before the `After` fires, which your own tests pin (`test_event_loop.py:1306,1498`). Pre-existing for the per-tool pair, but this PR adds a second instance; a "...during tool execution" qualifier would fix it.

site/src/content/docs/user-guide/concepts/tools/executors.mdx
```diff
@@ -112,7 +112,7 @@ Cancellation works identically in both modes. Call `agent.cancel()` to request c
 <Tabs>
 <Tab label="Python">
 
-- **Pre-launch cancel**: set `BeforeToolCallEvent.cancel_tool` on a per-tool hook to produce an error result for that tool.
+- **Pre-launch cancel**: set `BeforeToolCallEvent.cancel_tool` on a per-tool hook to produce an error result for that tool, or set `BeforeToolsEvent.cancel` on the batch-level hook to produce error results for every tool in the batch before any execute.
```
strandly-the-agent: ⚪ nit — the Python "Pre-launch cancel" bullet (this line) omits `agent.cancel()`, while the TS bullet four lines down documents it. Calling `agent.cancel()` before the batch starts also produces error results for every tool (`event_loop.py`'s `elif agent._cancel_signal.is_set(): cancel_message = "Tool execution cancelled"`, pinned by `test_event_loop_cycle_agent_cancel_from_before_tools_uses_batch_results`). Suggest appending ", or call `agent.cancel()` before the batch starts," to match the TS tab.

strandly-the-agent (commented):
**Reviewed `75aa7cb` — no blockers.** Faithful port, strong tests. Three should-fixes: a hook exception in the `finally` can silently discard an in-flight interrupt approval, and two load-bearing branches survive mutation testing untested. One process item: this adds public API with no `api/*-review` label.

✅ Verified: focused suite **269 passed** (`event_loop` + `agent/hooks` + `test_agent_hooks` + `test_agent_cancellation` + `hooks`); CI green at head except `label-size`; 2 mutants survived the full suite; interrupt-loss + cancel-path repros run cold against a live `Agent`.

<details>
<summary>How this was reviewed, and what I checked but did <b>not</b> report</summary>

Pipeline: triage → context build → 6 parallel specialist passes (correctness, adversarial/repro, API bar-raising, port-equivalence, test-quality, docs-accuracy) → aggregation → orchestrator re-verification at the new head. Started at `8951165`; you pushed `98d70213`+`75aa7cb` mid-review, so every surviving finding was re-run at `75aa7cb`.

**Dropped after I failed to confirm them myself** (worth recording so nobody re-raises them):
- *"Cancel path now appends an empty-content `user` message — regression vs base."* A pass reported the removal of the old `if tool_results:` guard makes `{"role":"user","content":[]}` enter history when a model reports `stopReason=tool_use` with zero `toolUse` blocks. I A/B'd it with a lying model + `agent.cancel()` from `AfterModelCallEvent`: **merge-base produces the identical empty message**, so it is not a regression from this PR. Not reported.
- *"`AfterToolsEvent` never fires if the stream consumer abandons the batch."* Three passes disagreed with each other. I tested all three abandonment styles (`await gen.aclose()`, drop-ref + `gc.collect()`, plain `break`): **it fired in all three, with the full result set.** Evidence too contradictory to post.
- `finally`-throw masking an in-flight exception, missing `end_cycle()` on hook-exception paths, deep mutability of `message`/result dicts, `""`-is-falsy: real but pre-existing/systemic to the hooks system, or exact TS parity. Folded into a nit or dropped rather than spent as comments.
- The per-cycle double-fire note: already resolved by `98d70213` — I re-confirmed your TS-parity reasoning is right and the new docstring + resume-leg test cover it.

**Verified clean** (no finding): cancel precedence hook-over-agent; one error `ToolResult` per pending tool use across both executors, incl. invalid/unknown tool names; `end_turn` precedence over `stop_event_loop`/structured-output/cancel/checkpoint, and that history stays provider-valid so re-invocation works; `BeforeToolsEvent` interrupt correctly skipping `AfterToolsEvent` (TS-faithful); resume dedup and no duplicated results; cycle-span closure on every return path; writable-property gates on both events; all three declared divergences in your description behave exactly as documented; the streaming docs pages the port issue listed genuinely did not need updating.
</details>

strandly-the-agent:
Questions from my review of `75aa7cb`, kept out of the inline threads because none is a `file:line` defect.

**Needs an answer before merge**
- **API-review label.** Two new customer-facing classes exported from `strands.hooks` classify as *moderate* under `team/API_BAR_RAISING.md:26-27` ("a new class that customers use to achieve new behavior") — informal sign-off from a designated API reviewer, not an API meeting. The PR carries neither `api/needs-review` nor `api/review-complete`. Worth flagging that `check-api-review-label` is green **vacuously**: `api-review-label.yml:22-25` returns early when neither label is present, so CI passing here isn't evidence the gate was satisfied.

<details>
<summary>Two non-blocking design questions (follow-up material, not asks on this PR)</summary>

- **Two mechanisms now halt the loop after tools, with different customer-visible results.** `AfterToolsEvent.end_turn` → `stop_reason="end_turn"` + a synthesized assistant message; the pre-existing `request_state["stop_event_loop"]` → `stop_reason="tool_use"` + the model's tool-use message. The vended `stop` tool uses the latter (`experimental/tools/stop/stop.py:112`), and `vended-tools.mdx:283` documents the resulting difference as a Python-vs-TS divergence — one this PR arguably just made fixable. Which is blessed going forward, and is moving `make_stop` onto `end_turn` worth a follow-up issue?
- **`v1:before_tools:{uuid5(name)}` (`events.py:167`) has no batch/cycle discriminator**, unlike the per-tool id which includes `toolUseId`. Behavior is correct (I verified two batches in one invocation each get their own approval), but two distinct approvals within one invocation share the same `Interrupt.id` — does that matter for a client that dedupes or caches prompts by interrupt id?

Also worth noting for whoever merges: per your own banner, the full `hatch run prepare` gate hasn't run locally, and `label-size` is the one red check at head.
</details>

strands-py/src/strands/event_loop/event_loop.py
liramon2: I'd rather keep this port faithful.

strands-py/src/strands/hooks/events.py
liramon2: No, I'd rather keep this port faithful.

strands-py/src/strands/hooks/events.py
liramon2: Addressed in commit 9a96951ba7d07ad26cb76e2c0a6597cfcad2f8f4

site/src/content/docs/user-guide/concepts/tools/executors.mdx
liramon2: Addressed in commit 9a96951ba7d07ad26cb76e2c0a6597cfcad2f8f4

strands-py/src/strands/event_loop/event_loop.py
liramon2: This edge case is extremely niche and untested on existing paths (`end_turn` and `stop_event_loop`)

strands-py/src/strands/event_loop/event_loop.py
liramon2: Addressed in commit 9a96951ba7d07ad26cb76e2c0a6597cfcad2f8f4

liramon2:
@strandly-the-agent Review this PR considering the new replies

strandly-the-agent:
**Re-reviewed `9a96951` — all three addressed items verified, both declines hold up. No code findings left from my side.**

The one thing worth surfacing: I tested your `end_turn`/`stop_event_loop` rationale on the M7 decline rather than taking it on trust, and **it's correct** — those two paths are equally unasserted, so this is a suite-wide gap, not something the batch hooks introduced:

```
M7   drop structured_output on INTERRUPT stop  (:888) -> 270 passed   (survives)
M-a  drop structured_output on END_TURN stop   (:920) -> 270 passed   (survives)
M-b  drop structured_output on STOP_LOOP stop  (:930) -> 270 passed   (survives)
```

That makes "niche, and consistent with existing paths" an accurate description, so I'm dropping it rather than re-raising.

<details>
<summary>Verification ledger at <code>9a96951</code> (cold sandbox, re-cloned)</summary>

**Fixed — verified:**
- 🟡 `event_loop.py:849` filter untested → **now a real mutation-killer.** Dropping the filter fails exactly the new test, with the predicted symptom:
  ```
  FAILED test_event_loop_cycle_per_tool_interrupt_with_invalid_tool_no_duplicate_result
  E   assert 2 == 1        # t2 duplicated: ['t2', 't2', 't1']
  ```
  Restored, the suite is **270 passed**. The test drives the real path (per-tool interrupt on `t1` + invalid `t2`, then resume) rather than mocking around it. ✅
- ⚪ `BeforeToolsEvent` docstring → rewritten to "at the start of each event-loop cycle that has pending tool-use requests" and it now states that completed tools are excluded from the pending set on resume. That second sentence also quietly answers the resume-contract half of my `events.py:172` question (a hook can now read what `cancel` will actually affect), which is the part I cared most about. ✅
- ⚪ `HOOKS.md` paired-events invariant → "during tool execution" qualifier plus an explicit short-circuit bullet; matches what `test_event_loop.py:1306,1498` pin. ✅
- ⚪ `executors.mdx` Python bullet → now documents `agent.cancel()`, matching the TS tab. ✅

**Declined — reasonable, not re-raising:**
- `event_loop.py:873` interrupt loss on an `AfterToolsEvent` hook exception: "keep the port faithful" is a fair call given I'd already confirmed merge-base loses the interrupt identically — it's pre-existing, not a regression. Only ask is that it not evaporate: happy to open a tracking issue for it if you want one (persisting/activating interrupt state *before* the hook dispatch is the natural fix), just say the word rather than me filing unprompted.
- `events.py:172` outcome channel (`cancel_message`/typed `tool_results`): deferring on TS parity is defensible and it's additive later. Worth putting in front of the API reviewer since it's a field-set question, not an implementation one.

**Also checked:** `ruff check src tests` clean, `ruff format --check` clean on the three touched files, delta vs `75aa7cb` is 4 files / +50/-5 with no scope creep, worktree clean at `9a96951`.

</details>

Only thing still open from my earlier pass is the process gate, unchanged and not something I can close: labels are still `enhancement, python, area-hooks, area-hil, size/xl`, so the `api/*-review` sign-off for the two new `strands.hooks` exports is still owed — and remember `check-api-review-label` goes green vacuously when no label is present.

strands-py/src/strands/hooks/events.py
```diff
@@ -133,6 +133,85 @@ class MessageAddedEvent(HookEvent):
...
+@dataclass
+class AfterToolsEvent(HookEvent):
+    """Event triggered once after a batch of tools finishes executing.
+
+    This event is fired after the tools requested in an assistant message have
+    completed (or been cancelled), giving hooks aggregate visibility over the
+    tool results before the next model call. It is always paired with a preceding
+    ``BeforeToolsEvent``, including on cancel and error paths.
+
+    It fires once per event-loop cycle, not once per logical batch. When a per-tool
```
JackYPCOnline: verbose docstring, should match ts, tutorial related should be included in docs.

JackYPCOnline:
This is a comprehensive PR that includes hook event implementation, documentation updates, and wiring changes. To make review and maintenance easier, would you consider splitting it into two smaller PRs?
PR 1: Hook event implementation + documentation updates
PR 2: Wiring changes
This approach would allow us to focus reviews on each component separately and address any issues more efficiently.
Maybe not this time.

liramon2:
> This is a comprehensive PR that includes hook event implementation, documentation updates, and wiring changes. To make review and maintenance easier, would you consider splitting it into two smaller PRs? PR 1: Hook event implementation + documentation updates PR 2: Wiring changes This approach would allow us to focus reviews on each component separately and address any issues more efficiently. Maybe not this time.

Will do in future PRs

strands-py/src/strands/hooks/events.py
liramon2: Addressed in 350b5256c28e042a0eb8ce873adc9668dfea35cf

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -744,91 +779,103 @@ async def _handle_tool_execution(
...
+                    yield StructuredOutputEvent(structured_output=structured_output_result)
+                    structured_output_context.stop_loop = True
+    finally:
+        # Always pair BeforeToolsEvent with AfterToolsEvent, even on cancel/interrupt/error paths.
+        tool_result_message: Message = {
+            "role": "user",
+            "content": [{"toolResult": result} for result in tool_results],
+        }
+        after_tools_event = AfterToolsEvent(agent=agent, message=tool_result_message, invocation_state=invocation_state)
+        after_tools_event, _ = await agent.hooks.invoke_callbacks_async(after_tools_event)
```
mehtarac: can wrap the call to `invoke_callbacks_async` in its own try/except` block to help with this

strands-py/src/strands/hooks/events.py
```diff
@@ -133,6 +133,76 @@ class MessageAddedEvent(HookEvent):
...
+        """
+        return f"v1:before_tools:{uuid.uuid5(uuid.NAMESPACE_OID, name)}"
+
+
+@dataclass
+class AfterToolsEvent(HookEvent):
+    """Event triggered after all tools complete execution.
+
+    This event is fired after tool results are collected and ready to be added to conversation.
+    Always paired with a preceding ``BeforeToolsEvent``, including on cancel and
```
mehtarac: I think this is not entirely true. The test `test_event_loop_cycle_after_tools_not_fired_on_before_tools_interrupt` and `test_event_loop_cycle_before_tools_hook_exception_skips_execution` test that the AfterToolsEvent dot not fire. 

I think it should say something along the lines of `Paired with a preceding BeforeToolsEvent when the batch
  proceeds past the pre-execution phase`

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -744,91 +779,103 @@ async def _handle_tool_execution(
...
+            for tool_use in tool_uses:
+                cancel_result: ToolResult = {
+                    "toolUseId": str(tool_use.get("toolUseId")),
+                    "status": "error",
+                    "content": [{"text": cancel_message}],
+                }
+                tool_results.append(cancel_result)
+                yield ToolResultEvent(cancel_result)
+        else:
+            pending_tool_use_ids = {tool_use["toolUseId"] for tool_use in tool_uses}
```
mehtarac, line 825: NIT: we use both `str(tool_use.get("toolUseId"))` and `tool_use["toolUseId"]`. Curious if any particular reason? 

I think we can use a consistent approach with probably the get through out?

strands-py/src/strands/event_loop/event_loop.py
liramon2: Addressed in a838c3ca23d591bce297c9af3e96f0e6dffe823b

strands-py/src/strands/hooks/events.py
liramon2: Addressed in a838c3ca23d591bce297c9af3e96f0e6dffe823b

strands-py/src/strands/event_loop/event_loop.py
liramon2: The `str(tool_use.get("toolUseId"))` was used in the original code before moving it over, but we can make it consistent. Addressed in a838c3ca23d591bce297c9af3e96f0e6dffe823b.

strands-py/src/strands/event_loop/event_loop.py
liramon2: Update: now in 28934cba568fb1b447cbc27469086a096ae07507

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -730,12 +769,8 @@ async def _handle_tool_execution(
             - The updated event loop metrics,
             - The updated request state.
     """
-    tool_uses: list[ToolUse] = []
+    tool_uses = [content["toolUse"] for content in message["content"] if "toolUse" in content]
```
JackYPCOnline: Nit: shouldn't we keep the field type?

JackYPCOnline (dismissed):
Incline to approve with one nit.

strands-py/src/strands/event_loop/event_loop.py
liramon2: Addressed in c754f9d05c629f9917c6a1ccce71ef68ae161b38

JackYPCOnline (approved):

mehtarac (approved):

PR #3474, opened 2026-07-24
feat(model-routing): add ModelRouter and accept it via Agent(model=)

JackYPCOnline:
## Description

An `Agent` holds one `Model`, resolved in the constructor, so model choice is fixed at build time: an agent cannot send an easy turn to a cheap model, and it has no answer to a provider outage. Frontier models cost roughly 5x to 10x what smaller ones do, so both matter, and customers asked for them directly (#364).

This PR makes model choice a runtime decision. `Agent(model=ModelRouter(...))` accepts an ordered set of candidate models plus a `RoutingStrategy`; single-model usage is untouched. It implements the P0 of `team/designs/0016-model-routing.md` and builds on the merged per-call-model seam (#3434).

**The strategy decides; the router orchestrates.** The router asks the strategy for a candidate before the first model call, and again after a call fails without a hook claiming the retry, passing the attempts so far. So the opening choice, whether to fail over, and to what are all the strategy's — the router has no failover policy of its own. Selection is asked **once per invocation**, not once per model call, so a tool loop does not re-pay for the decision.

Two rules cover every answer:

- **`None` means "use the default".** At the opening choice the request runs the router's default model — the first declared candidate, resolved without consulting a strategy. After a failure it ends routing, so the model's error surfaces.
- **Anything else the router cannot apply does not become a model choice of the router's own.** A strategy that raises propagates at the opening choice and ends routing after a failure. A failure round uses each candidate at most once, so naming one it already used ends routing. A candidate that will not resolve to a model is the one exception: after a failure it takes its slot in the round and the strategy is asked again, so one unusable candidate does not strand the healthy ones declared after it.

Each new candidate gets a fresh retry budget, and `max_switches` caps switches per invocation — selection is asked once, but every failed model call can switch, so a long tool loop has many chances to escalate.

**The default.** `FallbackStrategy` excludes candidates already tried since the last success, then prefers the fewest recorded failures, breaking ties by declaration order. So `ModelRouter([a, b])` is ordered failover while nothing is failing repeatedly, and a model that stays down sinks below healthier ones instead of costing a full retry budget in its declared slot. A success re-arms every candidate but clears only the succeeding candidate's failure count, so a repeatedly failing model stays demoted.

**Isolation.** Selection writes to the per-call `InvokeModelContext.model`; `agent.model` is never mutated, which is what keeps concurrent invocations independent. A router is immutable and reusable across agents, with all mutable state in `invocation_state`, keyed by agent and router together and cleared at both ends of an invocation.

[...]

PR #3513, opened 2026-07-27
feat(model-routing): add per-call selection, strategies, and ordered fallback

JackYPCOnline:
> **Stacked on #3474 (Phase B).** Phase B is not yet merged to `main`, so this PR's diff currently
> includes the Phase B commit as well. Review Phase B in #3474; the Phase-C-only changes here are
> `models/routing/{strategy,strategies}.py`, the strategy/fallback additions in
> `models/routing/router.py`, the widened public surface in `models/routing/__init__.py`, the
> selection-middleware registration in `agent/agent.py`, and the one-line
> `ModelRetryStrategy.reset_retry_state()` visibility change. Once #3474 merges I will rebase and
> this diff becomes Phase-C-only.

## Description

Phase C of model routing (design `team/designs/0016-model-routing.md`). Where Phase B resolves a single model at construction, this adds **per-call selection** and **ordered fallback**, matching the P0 "router core and fallback" + local-strategy items in the design.

A `RoutingStrategy` chooses a candidate per model call; a selection middleware runs it once per invocation (cached in `invocation_state`), sets the per-call model, and is registered before the agentic capability middleware so the selected model is what capability middleware sees. Independently, the router installs an ordered-fallback hook: when the selected model's retries are exhausted, it advances to the next untried candidate in declaration order and requests another attempt. This composes with the existing `ModelRetryStrategy` rather than replacing it.

## Related Issues

Refs #364

## Type of Change

New feature

## Public API (provisional; package is internal pending API review)

```python
from strands.models.routing import ModelRouter, RoutingCandidate, FallbackStrategy, ContextFitStrategy

# Ordered fallback (default strategy): try sonnet, fall over to opus on failure.
agent = Agent(model=ModelRouter(models=[sonnet, opus]))

# Capacity-based local selection.
agent = Agent(model=ModelRouter(models=[haiku, sonnet, opus], strategy=ContextFitStrategy()))
```

- `RoutingStrategy` (Protocol): `async select(context) -> RoutingCandidate`; the result must be one of the candidates.
- `RoutingContext`: `messages`, `system_prompt`, `tool_specs`, `candidates`, `invocation_state`.
- `ModelRouter(models, *, strategy=None)` — `strategy` defaults to `FallbackStrategy` (selects the first candidate). Selection recurses into nested routers.
- **Ordered fallback is always active**, under any strategy: on an unretried model failure the router advances through untried candidates in declaration order, resetting the retry budget each hop, until one succeeds or all are exhausted.
- Bundled strategies: `FallbackStrategy` (select-first) and `ContextFitStrategy` — routes by capacity, estimating tokens with each candidate's own `count_tokens` and treating a candidate as fitting when usage stays within `threshold` of its window (default: the conversation manager's `DEFAULT_COMPRESSION_THRESHOLD`). Picks the smallest fitting window, the largest when none fit; reuses `DEFAULT_CONTEXT_WINDOW_LIMIT` for models with no declared window, and is best-effort if a count fails.

## Follow-ups (not in this PR)

- **Model-driven routing (Phase D):** a judge model classifies the request and names a candidate.
- P1: cost-aware, cache-affinity (sticky), classifier/semantic.

## Testing

Unit tests cover strategy selection (fallback, context-fit with per-candidate counting and threshold), first-candidate/nested resolution, the select-result-must-be-a-candidate guard, per-invocation caching, and the ordered-fallback engine (advance on throttling and non-retryable failures, exhaustion, retry-budget reset, ordering after `ModelRetryStrategy`). The routing package is fully covered. An end-to-end integration test (real Bedrock) confirms ordered fallback recovers a real answer past a broken primary. ruff + mypy clean.

## Checklist

> Draft: per-call selection + ordered fallback, pending API review and the model-driven follow-up.

### DIFF
--- a/strands-py/src/strands/models/routing/router.py
+++ b/strands-py/src/strands/models/routing/router.py
@@ -5,8 +5,9 @@ through ``model=``. Its ``RoutingStrategy`` selects a candidate once per agent i
 choice is cached and reused for every model call in that invocation, including tool-loop turns.
 When the selected model fails and no hook has claimed the retry, the router advances to the next
 untried candidate in declaration order (ordered fallback), re-arming the chain after any successful
-call. A nested ``ModelRouter`` candidate is a single atomic fallback slot: its own strategy chooses
-its model, but the outer router does not fall back among the nested candidates.
+call. Each candidate receives a fresh model retry budget. A nested ``ModelRouter`` candidate is one
+atomic fallback slot: its own strategy chooses its model, but the outer router does not fall back
+among the nested candidates.
 """
 
 from __future__ import annotations
@@ -81,7 +82,7 @@ class ModelRouter(Plugin):
         if not candidates:
             raise ValueError("ModelRouter requires at least one candidate model")
         _reject_stateful(candidates)
-        _reject_duplicate_names(candidates)
+        _reject_duplicates(candidates)
         self._candidates = candidates
         self._strategy: RoutingStrategy = strategy or FallbackStrategy()
 
@@ -104,7 +105,7 @@ class ModelRouter(Plugin):
         return model
 
     def init_agent(self, agent: Agent) -> None:
-        """Register the fallback and state-cleanup hooks; reject attachment through ``plugins=[...]``.
+        """Register routing middleware and hooks; reject attachment through ``plugins=[...]``.
 
         Args:
             agent: The agent the router is attached to.
@@ -114,7 +115,11 @@ class ModelRouter(Plugin):
         """
         if agent._model_router is not self:
             raise ValueError("ModelRouter must be passed through Agent(model=...), not plugins=[...]")
-        # SDK_LAST so fallback runs after ModelRetryStrategy has decided whether to retry.
+
+        from ..._middleware.stages import InvokeModelStage
+
+        agent._middleware_registry.add_middleware(InvokeModelStage.Input, self._selection_middleware())
+        # SDK_LAST makes fallback run after ModelRetryStrategy decides whether to retry.
         agent.hooks.add_callback(AfterModelCallEvent, self._on_model_result, order=HookOrder.SDK_LAST)
         agent.hooks.add_callback(AfterInvocationEvent, self._clear_state, order=HookOrder.SDK_LAST)
 
@@ -140,8 +145,8 @@ class ModelRouter(Plugin):
         """Build an ``InvokeModelStage.Input`` handler that selects the per-invocation model."""
 
         async def middleware(context: InvokeModelContext) -> InvokeModelContext:
-            state = context.invocation_state.get(_ROUTING_KEY)
-            if state is None or state.get("router") is not self:
+            state = _owned_state(context.invocation_state.get(_ROUTING_KEY), self, context.agent)
+            if state is None:
                 routing_context = self._routing_context(
                     context.messages, context.system_prompt, context.tool_specs, context.invocation_state
                 )
@@ -149,6 +154,7 @@ class ModelRouter(Plugin):
                 index = self._index_of(candidate)
                 state = {
                     "router": self,
+                    "agent": context.agent,
                     "index": index,
                     "model": await self._resolve(candidate, routing_context),
                     "tried": {index},
@@ -161,8 +167,8 @@ class ModelRouter(Plugin):
 
     async def _on_model_result(self, event: AfterModelCallEvent) -> None:
         """Re-arm the fallback chain on success; on an unretried failure, advance to the next candidate."""
-        state = event.invocation_state.get(_ROUTING_KEY)
-        if state is None or state.get("router") is not self:
+        state = _owned_state(event.invocation_state.get(_ROUTING_KEY), self, event.agent)
+        if state is None:
             return
         if event.stop_response is not None:
             state["tried"] = {state["index"]}  # a successful call re-arms the rest of the chain
@@ -202,8 +208,8 @@ class ModelRouter(Plugin):
 
     async def _clear_state(self, event: AfterInvocationEvent) -> None:
         """Drop this router's routing state at the end of the invocation so it does not leak."""
-        state = event.invocation_state.get(_ROUTING_KEY)
-        if state is not None and state.get("router") is self:
+        state = _owned_state(event.invocation_state.get(_ROUTING_KEY), self, event.agent)
+        if state is not None:
             del event.invocation_state[_ROUTING_KEY]
 
     def _index_of(self, candidate: RoutingCandidate) -> int:
@@ -221,6 +227,18 @@ class ModelRouter(Plugin):
             candidates=self._candidates,
             invocation_state=invocation_state,
         )
+
+
+def _owned_state(value: Any, router: ModelRouter, agent: Agent) -> dict[str, Any] | None:
+    """Return valid routing state owned by the given router and agent."""
+    if not isinstance(value, dict) or value.get("router") is not router or value.get("agent") is not agent:
+        return None
+    if not isinstance(value.get("index"), int) or not isinstance(value.get("model"), Model):
+        return None
+    tried = value.get("tried")
+    if not isinstance(tried, set) or not all(isinstance(index, int) for index in tried):
+        return None
+    return value
 
 
 def _normalize(models: object) -> tuple[RoutingCandidate, ...]:
@@ -253,12 +271,18 @@ def _reject_stateful(candidates: tuple[RoutingCandidate, ...]) -> None:
             raise ValueError(f"candidate=<{label}> is stateful; routing among stateful models is not supported")
 
 
-def _reject_duplicate_names(candidates: tuple[RoutingCandidate, ...]) -> None:
-    """Reject colliding candidate names; unnamed candidates and repeated models are allowed."""
-    seen: set[str] = set()
+def _reject_duplicates(candidates: tuple[RoutingCandidate, ...]) -> None:
+    """Reject repeated candidate instances or colliding names; repeated models are allowed."""
+    seen_candidates: set[int] = set()
+    seen_names: set[str] = set()
     for candidate in candidates:
+        identity = id(candidate)
+        if identity in seen_candidates:
+            raise ValueError("duplicate RoutingCandidate instance")
+        seen_candidates.add(identity)
+
         if candidate.name is None:
             continue
-        if candidate.name in seen:
+        if candidate.name in seen_names:
             raise ValueError(f"duplicate candidate name=<{candidate.name}>")
-        seen.add(candidate.name)
+        seen_names.add(candidate.name)
--- a/strands-py/src/strands/models/routing/router.py
+++ b/strands-py/src/strands/models/routing/router.py
@@ -1,0 +1,90 @@
+"""ModelRouter: a reusable, immutable set of candidate models with a concrete default.
+
+A router holds an ordered sequence of candidate models and is a ``Plugin`` so an agent can
+accept it through ``model=``. It exposes the first candidate, resolved to a concrete model, as
+its default and rejects stateful candidates.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Mapping, Sequence
+from typing import TYPE_CHECKING, Union
+
+from ...plugins.plugin import Plugin
+from ..model import Model
+
+if TYPE_CHECKING:
+    from ...agent.agent import Agent
+
+
+CandidateInput = Union[Model, "ModelRouter"]
+
+_ROUTER_PLUGIN_NAME = "strands:model-router"
+
+
+class ModelRouter(Plugin):
+    """A reusable, ordered set of candidate models with a concrete default."""
+
+    def __init__(self, models: Sequence[CandidateInput]) -> None:
+        """Initialize the router.
+
+        Args:
+            models: Candidates as a sequence of ``Model`` objects and nested ``ModelRouter``
+                instances. The first candidate is the router's default.
+
+        Raises:
+            TypeError: If ``models`` is not a sequence, or a candidate is not a ``Model`` or
+                ``ModelRouter``.
+            ValueError: If ``models`` is empty or any candidate is a stateful model.
+        """
+        super().__init__()
+        candidates = _normalize(models)
+        if not candidates:
+            raise ValueError("ModelRouter requires at least one candidate model")
+        _reject_stateful(candidates)
+        self._candidates = candidates
+
+    @property
+    def name(self) -> str:
+        """Stable plugin identifier."""
+        return _ROUTER_PLUGIN_NAME
+
+    @property
+    def default_model(self) -> Model:
+        """The first candidate resolved to a concrete model, recursing nested routers."""
+        candidate = self._candidates[0]
+        if isinstance(candidate, ModelRouter):
+            return candidate.default_model
+        return candidate
+
+    def init_agent(self, agent: Agent) -> None:
+        """Reject a router attached through ``plugins=[...]`` instead of ``model=``.
+
+        Raises:
+            ValueError: If the router was not attached through ``Agent(model=...)``.
+        """
+        if agent._model_router is not self:
+            raise ValueError("ModelRouter must be passed through Agent(model=...), not plugins=[...]")
+
+
+def _normalize(models: object) -> tuple[CandidateInput, ...]:
+    """Validate the input is a sequence of candidates and return them as a tuple."""
+    if isinstance(models, (str, bytes, Mapping)) or not isinstance(models, Sequence):
+        raise TypeError("models must be a sequence of candidates")
+    return tuple(_validate_candidate(item) for item in models)
+
+
+def _validate_candidate(candidate: object) -> CandidateInput:
+    """Return the candidate if it is a ``Model`` or ``ModelRouter``; reject other types."""
+    if isinstance(candidate, (Model, ModelRouter)):
+        return candidate
+    raise TypeError(f"candidate must be a Model or ModelRouter; got {type(candidate).__name__}")
+
+
+def _reject_stateful(candidates: tuple[CandidateInput, ...]) -> None:
+    """Reject any stateful candidate model."""
+    for candidate in candidates:
+        if isinstance(candidate, Model) and candidate.stateful:
+            raise ValueError(
+                f"candidate=<{type(candidate).__name__}> is stateful; routing among stateful models is not supported"
+            )
--- a/strands-py/src/strands/models/routing/router.py
+++ b/strands-py/src/strands/models/routing/router.py
@@ -1,48 +1,89 @@
-"""ModelRouter: a reusable, immutable set of candidate models with a concrete default.
+"""ModelRouter: a reusable, immutable set of candidate models with a routing strategy.
 
-A router holds an ordered sequence of candidate models and is a ``Plugin`` so an agent can
-accept it through ``model=``. It exposes the first candidate, resolved to a concrete model, as
-its default and rejects stateful candidates.
+A router holds an ordered sequence of candidates and is a ``Plugin`` so an agent can accept it
+through ``model=``. Its ``RoutingStrategy`` selects a candidate once per agent invocation; the
+choice is cached and reused for every model call in that invocation, including tool-loop turns.
+When the selected model fails and no hook has claimed the retry, the router advances to the next
+untried candidate in declaration order (ordered fallback), re-arming the chain after any successful
+call. A nested ``ModelRouter`` candidate is a single atomic fallback slot: its own strategy chooses
+its model, but the outer router does not fall back among the nested candidates.
 """
 
 from __future__ import annotations
 
-from collections.abc import Mapping, Sequence
-from typing import TYPE_CHECKING, Union
+import logging
+from collections.abc import Awaitable, Callable, Mapping, Sequence
+from dataclasses import dataclass, replace
+from typing import TYPE_CHECKING, Any, Union
 
+from ...hooks.events import AfterInvocationEvent, AfterModelCallEvent
+from ...hooks.registry import HookOrder
 from ...plugins.plugin import Plugin
 from ..model import Model
+from .strategy import RoutingContext, RoutingStrategy
 
 if TYPE_CHECKING:
+    from ..._middleware.stages import InvokeModelContext
     from ...agent.agent import Agent
 
+logger = logging.getLogger(__name__)
 
-CandidateInput = Union[Model, "ModelRouter"]
+
+@dataclass(frozen=True)
+class RoutingCandidate:
+    """A routing candidate: a model with an optional name and description.
+
+    ``model`` may be a nested ``ModelRouter``; in a fallback chain it is one atomic slot (its own
+    strategy picks its model, and the outer router does not fall back among the nested candidates).
+    """
+
+    model: Model | ModelRouter
+    name: str | None = None
+    description: str | None = None
+
+
+CandidateInput = Union[Model, "ModelRouter", RoutingCandidate]
 
 _ROUTER_PLUGIN_NAME = "strands:model-router"
+_ROUTING_KEY = "__strands_model_routing__"
+
+
+class FallbackStrategy:
+    """Selects the first candidate; the router advances through the rest on failure."""
+
+    async def select(self, context: RoutingContext, **kwargs: Any) -> RoutingCandidate:
+        """Return the first candidate."""
+        return context.candidates[0]
 
 
 class ModelRouter(Plugin):
-    """A reusable, ordered set of candidate models with a concrete default."""
+    """A reusable, ordered set of candidate models with a routing strategy and ordered fallback."""
 
-    def __init__(self, models: Sequence[CandidateInput]) -> None:
+    def __init__(self, models: Sequence[CandidateInput], *, strategy: RoutingStrategy | None = None) -> None:
         """Initialize the router.
 
         Args:
-            models: Candidates as a sequence of ``Model`` objects and nested ``ModelRouter``
-                instances. The first candidate is the router's default.
+            models: Candidates as a sequence. Each is a ``Model``, a nested ``ModelRouter``, or a
+                ``RoutingCandidate`` carrying an optional name/description. The first candidate is
+                the router's default.
+            strategy: Selects a candidate per invocation. Defaults to selecting the first candidate.
 
         Raises:
-            TypeError: If ``models`` is not a sequence, or a candidate is not a ``Model`` or
-                ``ModelRouter``.
-            ValueError: If ``models`` is empty or any candidate is a stateful model.
+            TypeError: If ``models`` is not a sequence, a candidate is not a ``Model`` or
+                ``ModelRouter``, or ``strategy`` does not implement ``RoutingStrategy``.
+            ValueError: If ``models`` is empty, candidate names collide, or any candidate is a
+                stateful model.
         """
         super().__init__()
+        if strategy is not None and not isinstance(strategy, RoutingStrategy):
+            raise TypeError("strategy must implement RoutingStrategy (a select(context) method)")
         candidates = _normalize(models)
         if not candidates:
             raise ValueError("ModelRouter requires at least one candidate model")
         _reject_stateful(candidates)
+        _reject_duplicate_names(candidates)
         self._candidates = candidates
+        self._strategy: RoutingStrategy = strategy or FallbackStrategy()
 
     @property
     def name(self) -> str:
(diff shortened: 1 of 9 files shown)


JackYPCOnline:
@strandly-the-agent can you review this Implementation and check: 1. if this fit PhaseB PR ? 2. Does this provide extensibility and value? 3. Any concerns and other issues.

strands-py/src/strands/models/routing/router.py
```diff
@@ -0,0 +1,215 @@
...
+        async def middleware(context: InvokeModelContext) -> InvokeModelContext:
+            state = context.invocation_state.get(_ROUTING_KEY)
+            if state is None:
+                routing_context = self._routing_context(
+                    context.messages, context.system_prompt, context.tool_specs, context.invocation_state
+                )
+                candidate = await self._pick(routing_context)
+                index = next(i for i, existing in enumerate(self._candidates) if existing is candidate)
+                state = {"model": await self._resolve(candidate, routing_context), "tried": {index}}
+                context.invocation_state[_ROUTING_KEY] = state
```
strandly-the-agent, line 162: 🔴 **The routing decision is cached in a dict that isn't reliably invocation-scoped, and carries no ownership marker.** This one line is the root cause of three separate failures. `_ROUTING_KEY` is a module-level constant, and the middleware's only cache check is `state is None` (`router.py:128-129`) — so *any* dict that already holds that key wins, whoever put it there.

**1. Cross-agent model leak in `Graph`/`Swarm`.** `graph.py:1042` and `swarm.py:890` pass **one** `invocation_state` dict to every node's `stream_async`. The first router to select writes its concrete `Model` into it; every later node finds `state is not None` and runs on **that** model, even with a disjoint candidate list. The `tried` indices cross over too, so node B's fallback walks A's index space. Reproduced on a two-node `A → B` graph where each agent has a single-candidate router:

```text
A-ONLY model stream calls: 2
B-ONLY model stream calls: 0     <-- agent B's OWN model
agent B's messages: assistant ['answer from A-ONLY model']
```

Same with a real `Swarm` `handoff_to_agent` (`S-A calls=3, S-B calls=0`). Control with plain models instead of routers: `A-ONLY calls=1  B-ONLY calls=1 -> isolated: True`. Silent — wrong cost, wrong capabilities, wrong quota, no log line.

**2. State outlives the invocation.** `agent.py:1204` assigns a caller-supplied `invocation_state` **by reference**, so the idiomatic `ctx = {"user_id": 42}` reused across calls pins every later call to the first selection with `tried` already full. Shared dict: `call 1 OK / call 2 RAISED ValueError`. Fresh dict per call: `call 1 OK / call 2 OK 'A healthy'`. `ContextFitStrategy` also stops re-evaluating (`call 1 -> small-model / call 2 -> small-model` after the request grew to 900k tokens). The run also leaves `{'model': <Model>, 'tried': {0,1}}` plus an `Agent` in the caller's own dict, so `json.dumps(caller_dict)` now fails. The SDK's own `AfterInvocationEvent.resume` path re-enters `_run_loop` with the same dict and hits this without any user dict involved.

**Caveat on confidence:** each of these was reproduced twice in fresh processes by my adversarial pass, but a second pass that read the code without executing it declined to call the multi-agent one a bug, and my own cold re-verification couldn't run (sandbox died). Please confirm before acting.

**Fix** — one change closes both, plus the malformed-state crashes: record ownership and validate on read.

```suggestion
                state = {
                    "router": self,
                    "model": await self._resolve(candidate, routing_context),
                    "tried": {index},
                }
                context.invocation_state[_ROUTING_KEY] = state
```

…with the read guard becoming `if state is None or state.get("router") is not self:` at `:129`. Then also clear it at the end of the invocation — you already have the hook seam in `init_agent`:

```python
agent.hooks.add_callback(AfterInvocationEvent, self._clear_routing_state, order=HookOrder.SDK_LAST)
```

Worth tests for a router inside a `Graph`, and for one `invocation_state` dict reused across two `agent()` calls.

strands-py/src/strands/models/routing/router.py
```diff
@@ -0,0 +1,215 @@
...
+
+    async def _advance_on_failure(self, event: AfterModelCallEvent) -> None:
+        """On an unretried model failure, advance to the next untried candidate in declaration order."""
+        if event.retry or event.stop_response is not None or event.exception is None:
+            return
+        state = event.invocation_state.get(_ROUTING_KEY)
+        if state is None:
+            return
+        tried: set[int] = state["tried"]
+        next_index = next((index for index in range(len(self._candidates)) if index not in tried), None)
```
strandly-the-agent, line 180: 🔴 **`tried` never re-arms after a success, so one transient blip disarms fallback for the rest of the invocation.** `ModelRetryStrategy` resets its budget on every successful call (`_retry.py:127-133`); the router never does. Since `invocation_state` is the same dict across the whole tool loop, `tried` accumulates across turns and `next_index` eventually returns `None` forever — the agent is pinned to the last-failed model with healthy candidates sitting idle.

Reproduced (A blips once then is healthy; B serves turn 1 then blips on turn 2; `retry_strategy=None`):

```text
RAISED: EventLoopException: B transient blip
A.stream calls=1   B.stream calls=2
A was healthy from call 2 onward and was never re-tried on turn 2
messages in history: 3   (user / assistant toolUse / user toolResult)
```

The invocation dies mid-tool-loop with a half-finished history. This makes the PR description's "until one succeeds or all are exhausted" read as a per-failure guarantee when it's really a per-invocation budget.

**Fix:** on a successful model call, shrink `tried` back to the live selection. You already have the `AfterModelCallEvent` hook — the success branch is currently just an early return:

```python
if event.stop_response is not None:
    state = event.invocation_state.get(_ROUTING_KEY)
    if state is not None and state.get("router") is self:
        state["tried"] = {state["index"]}   # keep the current pick, re-arm the rest
    return
```

That needs the chosen index stored alongside `"model"` (it currently isn't — `:135` records the index only in the `tried` set). If per-invocation exhaustion is actually the intent, then this is a docs fix instead, plus a test pinning it — but I don't think a user would predict it.

strands-py/src/strands/models/routing/router.py
```diff
@@ -0,0 +1,215 @@
...
+        if next_index is None:
+            return
+        routing_context = self._routing_context(
+            event.agent.messages,
+            event.agent.system_prompt,
+            event.agent.tool_registry.get_all_tool_specs(),
+            event.invocation_state,
+        )
+        state["model"] = await self._resolve(self._candidates[next_index], routing_context)
```
strandly-the-agent: 🟡 **No exception containment here — a failure while resolving the fallback candidate replaces the original model exception and suppresses `ForceStopEvent`.** `_resolve` can raise for foreseeable reasons: a nested router's strategy failing, or (per your own Phase D plan) a judge model timing out. Hook exceptions propagate — `HookRegistry.invoke_callbacks_async` says so explicitly — and this runs inside `event_loop.py`'s `except Exception as e:` block (`event_loop.py:605-630`) *before* `yield ForceStopEvent(reason=e); raise e` at `:629-630`. So the hook's exception becomes what the caller sees, the original is reachable only via the implicit `__context__` chain, and no `ForceStopEvent` is ever yielded:

```text
Exception surfaced to caller: RuntimeError - strategy blew up
  __context__ (implicit chain): ModelThrottledException primary throttled
```

Caller code doing `except ModelThrottledException:` silently stops matching. The same mechanism turns a malformed cached state into `KeyError: 'tried'` reaching the user with the real model error gone.

(Good news: a strategy raising during *initial* selection is safe — the key is never written, so this hook correctly no-ops and the original exception propagates cleanly. Verified.)

**Fix:** contain the risky span so a failed advance degrades to "no fallback" rather than corrupting the error:

```python
        try:
            state["model"] = await self._resolve(self._candidates[next_index], routing_context)
        except Exception as error:
            logger.warning("candidate_index=<%d>, error=<%s> | fallback resolution failed", next_index, error)
            return
```

strands-py/src/strands/models/routing/router.py
```diff
@@ -0,0 +1,215 @@
...
+        tried.add(next_index)
+        event.agent._retry_strategy.reset_retry_state()
```
strandly-the-agent, line 206: 🟡 **Resetting the retry budget on every hop makes attempts additive with no router-level ceiling.** Worst case is N candidates × `max_attempts` model calls in a single turn, with `ModelRetryStrategy`'s exponential backoff paid in full each time. Measured with `asyncio.sleep` patched to record the requested delays:

```text
1 candidate : 6 calls  — backoff [4, 8, 16, 32, 64] = 124s
2 candidates: 12 calls — 248s
4 candidates: 24 calls — 496s  (8.3 min of sleeping in one agent() call)
```

And in a multi-turn tool loop with 3 candidates, one `agent()` call: `a=7, b=6, c=6 → 19 model calls, 372s of requested backoff over 15 sleeps`. Because listing the same model object twice is explicitly supported (`test_repeated_model_object_is_allowed`), one physical model can absorb 12 attempts / 248s. Under a region-wide Bedrock throttle this amplifies the very condition it's escaping.

This is invisible to the suite because **all four fallback tests pass `retry_strategy=None`** (`test_router.py:346,357,368,390`), i.e. `max_attempts=1` — the production default path is never exercised together with fallback.

**Fix:** add a router-level ceiling (total attempts and/or a wall-clock deadline) rather than resetting unconditionally, or at minimum document the multiplicative worst case and add one fallback test on the default retry strategy. Reasonable people could argue the current behavior is correct-by-design — but the 8-minute number should be a deliberate choice, not an emergent one.

Related, same hook: a hop is completely silent. `router.py` has no logger at all, no span attribute, and no event — and because `event.retry = True` discards the exception, only the *last* candidate's error survives. With 3 candidates failing three different ways the user sees exactly one `TypeError` and has no way to learn the other two happened, or which model finally answered. A `logger.info` per hop (from → to, reason) and chaining the failures into the final exception would be a big DevX win for a feature whose whole premise is that the configured model may not be the model you got.

strands-py/src/strands/models/routing/router.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def init_agent(self, agent: Agent) -> None:
+        """Register the ordered-fallback hook; reject attachment through ``plugins=[...]``.
+
+        Raises:
+            ValueError: If the router was not attached through ``Agent(model=...)``.
+        """
+        if agent._model_router is not self:
+            raise ValueError("ModelRouter must be passed through Agent(model=...), not plugins=[...]")
+        agent.hooks.add_callback(AfterModelCallEvent, self._advance_on_failure, order=HookOrder.SDK_LAST)
```
strandly-the-agent: 🟡 **A nested `ModelRouter` never gets a fallback chain of its own, so a failure inside a nested group skips the whole subtree.** `init_agent` only runs for the router passed to `Agent(model=...)` — it explicitly rejects any other attachment at `:103` — so a nested router used as a candidate (`_resolve`, `:113-118`) never has `_advance_on_failure` registered. Its ordered-fallback method exists but is dead code in that position, and the *outer* router's `tried`/`next_index` range only over outer candidates.

Reproduced: `outer = [nested(failing_inner, good_inner), other_outer]` → the response comes from `OTHER_OUTER_RESPONDED`; `good_inner` is never tried, even though the nested router's own `FallbackStrategy` would have picked it. No test combines nesting with failure, though both are individually covered.

This is what makes the PR description's "**Ordered fallback is always active**, under any strategy … until one succeeds or all are exhausted" not quite true.

**Fix, either way is fine** — but pick one explicitly: (a) have `_advance_on_failure` recurse into a nested candidate's untried set before advancing the outer index, or (b) declare a nested router to be one atomic fallback slot, and say so in the module docstring, `RoutingCandidate`'s docs and the PR body. Either way it deserves a test pinning the chosen semantics — `test_selection_recurses_into_nested_router_strategy` only covers the happy path.

strands-py/src/strands/models/routing/strategy.py
```diff
@@ -0,0 +1,36 @@
...
+    tool_specs: Sequence[ToolSpec]
+    candidates: Sequence[RoutingCandidate]
+    invocation_state: Mapping[str, Any]
+
+
+@runtime_checkable
+class RoutingStrategy(Protocol):
+    """Selects one candidate for a model call."""
+
+    async def select(self, context: RoutingContext) -> RoutingCandidate:
```
strandly-the-agent: 🟡 **`select()` needs `**kwargs` — this is the one gap here that stops being free later.** `strands-py/AGENTS.md:159-164` makes it a review-enforced rule ("Use a `Protocol` with `**kwargs`, not a `Callable`… allows the interface to evolve"), with `docs/STYLE_GUIDE.md:69-76` as the canonical example, and all four sibling Protocols follow it (`memory/types.py:149`, `injection/types.py:57`, `mcp_client.py:71`, `graph.py:66-78`).

It's load-bearing here specifically: design 0016 sketches a `name` attribute this Protocol doesn't have, your own roadmap adds cost-aware/sticky/judge strategies, and my pass found three things a strategy demonstrably wants and can't get (`projected_input_tokens`, the `tried` set, per-candidate metadata). Every one of those is an additive change *if* `**kwargs` is there and a breaking change for third-party implementors if it isn't — and after Phase D ships, implementors exist.

```suggestion
    async def select(self, context: RoutingContext, **kwargs: Any) -> RoutingCandidate:
```

While you're here: the module docstring at `:1-4` promises the strategy sees the call's "projected token count", but `RoutingContext` (`:20-27`) has no such field — `InvokeModelContext.projected_input_tokens` exists (`_middleware/stages.py:39`) and `_routing_context` (`router.py:164-174`) just doesn't pass it through. Either plumb it (it's free and it's exactly what `ContextFitStrategy` re-derives at N× the cost) or drop the claim.

strandly-the-agent (commented):
**Verdict: the design is right and the fallback engine genuinely works — but I'd keep this in Draft.** The routing decision is cached in `invocation_state`, which isn't reliably invocation-scoped and carries no ownership marker, and that breaks three ways (one of them silently runs a `Graph`/`Swarm` node on another node's model).

Your three questions:

1. **Does it fit Phase B? Yes, cleanly.** `routing/*` at your base commit `e1dfa50` is byte-identical to #3474's head `c9d6120` — Phase C is purely additive (`ModelRouter.__init__` only gains keyword-only `strategy=`). Two caveats below, neither about the code.
2. **Extensibility: the seam is right, the strategy surface is still a skeleton.** I wrote and ran one custom strategy per follow-up on your own roadmap: only the judge's *decision* is cleanly buildable. Cost-aware has nowhere to put per-candidate metadata, sticky/cache-affinity needs the undocumented `invocation_state["agent"]` key, and fallback-history is impossible (`tried` lives behind a module-private key — and the strategy is never consulted on a hop anyway). **Value: the fallback engine is real and useful today** — one-line cross-provider failover, and it's the only door since `InvokeModelStage` registration is private. `FallbackStrategy` is a no-op selector and `ContextFitStrategy` rests on "smallest window = cheapest", which isn't true across providers.
3. **Concerns:** the state-scoping root cause (3 manifestations, inline), additive retry budgets with no ceiling (measured: 4 candidates → 24 model calls and ~8.3 min of backoff in one turn), no exception containment in the fallback hook, nested routers silently getting no fallback chain, and a model switch that is completely invisible.

<details>
<summary><b>Verification ledger — what I actually ran (and one honest gap)</b></summary>

- Reviewed at exact head `fdd370b309e80ab76c7dde2bb95dbf9367ecdea4`; Phase-C-only diff isolated as `git diff e1dfa50..HEAD` (9 files, +758/-50).
- ✅ `git diff e1dfa50..pr3474 -- src/strands/models/routing tests/strands/models/routing` is **empty** — the stack is clean; only unrelated `main` drift differs.
- ✅ PR suite: `pytest tests/strands/models/routing` → **44 passed**, reproduced twice cold by two independent passes; a wider run (`routing` + `test_agent_retry` + `middleware`) → **170 passed**. Codecov's 100% patch coverage confirmed independently (100% stmt/branch on all four routing files).
- ✅ Your PR-body "Public API" snippet runs verbatim in a clean venv — import paths and constructor calls are correct as documented.
- ✅ Confirmed **not** broken: registration ordering really does give the agentic capability middleware the routed model (`agentic_context.py:303` reads `context.model`); `BeforeModelCallEvent(cancel=…)`'s synthetic event correctly no-ops the router; a custom retry strategy without `reset_retry_state` is unreachable (`agent.py:452-456` rejects it); the `_reset_retry_state` rename has zero dangling refs across Python **and** TypeScript; `tried` bookkeeping is sound for ordinary candidates (strategy picks #2 → order `c2, c0, c1`, each exactly once); one `ModelRouter` shared by two Agents run concurrently via `asyncio.gather` is isolated; routing state is not serialized by checkpointing/session managers.
- ⚠️ **Gap I want to flag rather than paper over:** my sandbox died before I could run my own independent cold re-verification of the three 🔴s. Each was reproduced twice in fresh processes by my adversarial pass with pasted output, but a separate test-quality pass looked at the Graph/Swarm one from source only and explicitly *declined* to call it a bug. So treat 🔴 1 as "reproduced once, please confirm" rather than settled, and 🔴 2/🔴 3 as well-corroborated (two passes agree on the mechanism; only one executed it).
- Not run: the live-Bedrock integ test.

</details>

<details>
<summary><b>Q1 detail — the two caveats on the Phase B/C split</b></summary>

Both process, not code:

- **The PR body's "Phase-C-only changes" list omits `models/routing/__init__.py`** — which does change (+15/-6) and is exactly the file that defines the public surface under review. It's also the one file with prior review history: on #3474 I verified `__all__` had been *trimmed* to `CandidateInput`/`ModelRouter` and that `RoutingCandidate` had no dangling references. Phase C reintroduces `RoutingCandidate` plus four more names — legitimate now that they're implemented, but a reviewer following your note would skip the file.
- **Keep it in Draft (or retarget to `feature/model-routing-core`) until #3474 lands.** The only real bypass risk is #3513 merging first; you've handled that with Draft status, so this is just naming the failure mode.

One design-conformance divergence worth a decision: design 0016 (lines 70, 229) says the router registers its `InvokeModelStage.Input` handler *from `init_agent()`*, but `agent.py:419-427` registers it directly in `Agent.__init__` by reaching into `self._model_router._selection_middleware()`. Both `ContextInjector.init_agent` (`vended_plugins/context_injector/plugin.py:95-97`) and `MemoryManager.init_agent` (`memory/memory_manager.py:585,639`) register their Input middleware from `init_agent()`, and the registry already exists (`agent.py:418`) before plugin init (`agent.py:501-502`) — so the sibling pattern would work here. Functionally safe today; it just breaks the design's "one plugin installs both" encapsulation.

</details>

<details>
<summary><b>Q2 detail — what each roadmap strategy hits</b></summary>

| Follow-up from your description | Result | Blocker |
|---|---|---|
| cost-aware (P1) | possible, off-label | `RoutingCandidate` is `model`/`name`/`description` only (`router.py:26-32`); subclassing the frozen dataclass survives `_as_candidate` but a *required* field is impossible; a name-keyed side table gives `KeyError: None` on bare `Model` candidates. No `metadata=`. |
| sticky / cache-affinity (P1) | private access only | `invocation_state` is fresh per invocation, so cross-invocation state needs the undocumented `invocation_state["agent"]` key; writing to `RoutingContext.invocation_state` works at runtime but mypy rejects it (`Mapping[str, Any]`). |
| model-driven judge (Phase D) | decision clean, reporting impossible | judge's own token usage never reaches `AgentResult.metrics` (verified: main call only); no tracer/span and no channel to report *why* it chose. |
| fallback-history-aware | impossible | `tried` sits under `invocation_state["__strands_model_routing__"]` (`router.py:38`), and `select()` is called exactly **once** across a candidate-1→2 failover — hops walk raw declaration order. |

Two things would move this from skeleton to genuinely extensible: **a metadata channel on `RoutingCandidate`** and **observability of the decision**. `description` is currently dead weight — unused anywhere in `src/`.

Worth considering for *this* PR: ship the engine, but keep `RoutingStrategy`/`RoutingContext`/`ContextFitStrategy`/`CandidateInput` out of `__all__` until Phase D validates the shape. Three public names instead of seven, identical user value today, and it drops the API-review tier from Substantial to Moderate.

</details>

<details>
<summary><b>Other findings not worth their own thread</b></summary>

- 🟡 **Deprecated `Agent.structured_output()` bypasses routing entirely** — `agent.py:964` calls `self.model.structured_output(...)` directly, skipping `InvokeModelStage`. Verified: the strategy chose candidate #2, the call went to the default (`dead.structured_output calls=1`, `good=0`); the current `agent(prompt, structured_output_model=…)` path routes correctly. Worth documenting, or warning when a router meets the deprecated entry point.
- 🟡 **`agent.model` goes stale after a hop**, and proactive compression reads `agent.model.context_window_limit` (`conversation_manager.py:128`) — verified ratio 0.0100 vs a true 1.2501 for an 8k routed model behind a 1M default. Related: routing hijacks context-overflow recovery (a `ContextWindowOverflowException` burns the whole fallback budget on a recoverable condition).
- 🟡 **The routing state is user-visible and mutable** — `{'model': …, 'tried': {0,1}}` shows up in `stream_async` events and `callback_handler` kwargs (`_events.py:73-74`), and malformed values produce raw internals (`KeyError: 'tried'`, `TypeError: string indices must be integers`) with no mention of routing. The ownership check in my `router.py:136` thread doubles as validation here.
- 🟡 **`isinstance(strategy, RoutingStrategy)`** (`router.py:68-69`) promises validation it can't deliver: `select = 3`, a sync `select`, and a wrong-arity `select` all pass construction. `graph.py:75-77` documents this exact `runtime_checkable` trap.
- ⚪ Nits: identity-only return contract (`router.py:109-110`) gives the same message for a rebuilt-but-`==` candidate, `dataclasses.replace(...)`, and returning `candidate.model` — none of which the text points at; `RoutingContext` is a frozen "immutable view" but `router.py:131` passes `messages`/`tool_specs` through by reference; `FallbackStrategy` in `router.py` vs `ContextFitStrategy` in `strategies.py` (the module docstring apologizes for it); `Agent(model=[a, b])` dies with `AttributeError: 'list' object has no attribute 'stateful'`; the same `RoutingCandidate` object listed twice collapses to index 0 and burns a hop re-calling the model that just failed; `init_agent` has `Raises:` but no `Args:`; "the cheapest sufficient model" (`strategies.py:34`) asserts a cost equivalence the class has no signal for; `context_window_limit or DEFAULT` silently overrides an explicit `0`.
- ⚪ No user-guide page. `conversation-management` and `retry-strategies` both have one; routing has nothing beyond docstrings. Fine while provisional — flagging so it isn't forgotten at promotion.

</details>

<details>
<summary><b>Test gaps — 100% coverage, but the interesting scenarios aren't reached</b></summary>

Coverage is genuinely 100% on the routing files; the gaps are scenario-shaped, which is why they hide:

- **All four fallback tests pass `retry_strategy=None`** (`test_router.py:346,357,368,390`), so the production default (6 attempts + backoff) interacting with fallback is untested — that's the multiplication in my `router.py:161` thread. It also makes `test_fallback_resets_retry_budget_on_advance` near-tautological: with `max_attempts=1` the second candidate always gets exactly one attempt, so the test passes identically if `reset_retry_state()` were internally a no-op. Add `ModelRetryStrategy(max_attempts=3, initial_delay=0, max_delay=0)` with a *second* candidate that itself needs two retries — that version fails if the reset stops working.
- **No multi-turn tool-loop fallback test** → 🔴 2 is invisible. **No Graph/Swarm test** → 🔴 1 is invisible. **No test reuses an `invocation_state` dict across two `agent()` calls** → 🔴 3 is invisible; `test_selection_middleware_reselects_for_new_invocation_state` (`test_router.py:184`) covers the *fresh*-dict case, which is exactly the one not at risk. **No nested-router + failure test** → the `router.py:104` thread is invisible.
- Untested existing paths: a strategy raising from `select()` (propagates raw); `count_tokens` returning a bad *type* rather than raising.
- `_invoke_context` (`test_router.py:46-49`) duck-types `InvokeModelContext` with a `SimpleNamespace` — it'll keep passing if the real dataclass gains a required field.
- One mutation-test datapoint: forcing `next_index = 0` makes the suite **hang** rather than fail, because exhaustion is defined solely by `next_index is None`. CI would catch it as a 90s timeout, not a clean red X — worth bounding the advances or asserting `tried` grows monotonically.

</details>

<details>
<summary><b>Questions</b></summary>

**Worth settling before this leaves Draft**

1. Is per-*invocation* (rather than per-model-call) fallback exhaustion intentional? The design specifies per-invocation *selection* but is silent on whether `tried` re-arms after a success — 🔴 2 and 🔴 3 both hinge on that one decision.
2. Should this get `api/needs-review`? Your #3474 note said the doc-review meeting covered `Agent(model=ModelRouter)` — but Phase C adds `RoutingStrategy` (a Protocol customers implement), `RoutingContext`, `FallbackStrategy`, `ContextFitStrategy` and re-exports `RoutingCandidate`, none of which existed then. Per `team/API_BAR_RAISING.md` a customer-implemented Protocol reads as Substantial (reviewer meeting); the trim-`__all__` option above would drop it to Moderate. Note the CI gate can't help here — `api-review-label.yml:19-21` exits early and passes green when *neither* label is present.
3. `RoutingStrategy.select(self, context)` — `**kwargs` now, or never? (see the `strategy.py:34` thread; it's the one change that stops being free once Phase D ships implementors.)

**Non-blocking**

4. Should fallback consult the strategy on a hop instead of walking raw declaration order? Today with `ContextFitStrategy` a hop can route to a candidate the strategy just rejected as unfit — and the user then sees *its* `ContextWindowOverflowException`.
5. Any reason the selection middleware is registered in `Agent.__init__` rather than from `ModelRouter.init_agent()`, matching `ContextInjector`/`MemoryManager` and design 0016? If deliberate, a one-line comment or a design-doc update would save the next reader.
6. Is `event.agent._retry_strategy.reset_retry_state()` (`router.py:161`) the coupling you want long-term? The rename made the *method* public, but `_retry_strategy` is still a private `Agent` attribute with no public accessor — worth deciding now while the rename is already in flight.
7. `Refs #364` reads honest, but #364 was closed as completed on 2026-07-24 while design 0016 explicitly non-goals load balancing and response caching, and Phase D isn't shipped. Deliberate?

</details>

One process note on the design doc itself: several deltas below are arguably the *doc* needing an update rather than the code — `RoutingStrategy` has no `name` attribute, `CandidateInput` doesn't accept bare model-id strings (already true in Phase B), `tool_specs` is a `Sequence` not a tuple. Since 0016 is still unmerged in #3217, cheapest fix is probably amending it there.

Also worth being explicit: `projected_input_tokens` being computed pre-routing is *not* a finding — 0016 line 230 already calls it a known v1 limitation, and I confirmed this diff doesn't worsen it.

Nice work on the parts that are hard to get right — the `SDK_LAST` ordering against `ModelRetryStrategy` is correct, the middleware really does run before the capability middleware, and the stack hygiene against #3474 is exemplary. This is solid work for a human to approve; the state-scoping issue is the one thing I'd want fixed before it leaves Draft.

strands-py/src/strands/models/routing/strategies.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+        if not 0.0 < threshold <= 1.0:
+            raise ValueError("threshold must be in (0.0, 1.0]")
+        self._threshold = threshold
+
+    async def select(self, context: RoutingContext) -> RoutingCandidate:
+        """Return the smallest-window candidate whose window holds the request within ``threshold``."""
+        windows = [_candidate_window(candidate) for candidate in context.candidates]
+        counts = await asyncio.gather(*(self._count(candidate, context) for candidate in context.candidates))
+        fitting = [index for index, window in enumerate(windows) if counts[index] <= window * self._threshold]
```
strandly-the-agent, line 57: 🟡 **`_count` is carefully guarded; the two values it's compared against aren't.** `count_tokens` is documented as best-effort and `Model.count_tokens` returns `Any`-ish in practice, so three neighbours can turn a routing heuristic into a fatal error:

- **Raw return value used directly here.** A provider returning `None` or `"12"` gives `TypeError: '<=' not supported between instances of 'NoneType' and 'float'` — inconsistent with the *raise* path, which `test_context_fit_treats_failed_count_as_not_fitting` covers. Negative counts are accepted; `nan` never fits.
- **`_candidate_window` (`:87-89`) is unguarded** and `context_window_limit` calls `get_config()`. Verified: a model whose config raises takes down the whole invocation — `agent('hi') raised RuntimeError: config not initialised yet` — from a strategy that documents itself as best-effort.
- **`asyncio.gather` at `:55` has no `return_exceptions`**, so a `CancelledError` bypasses `except Exception` and leaves sibling `count_tokens` coroutines running as orphans (confirmed).

**Fix:** coerce inside the existing `try` (`value = float(await model.count_tokens(...))`, treating non-finite as `math.inf`), wrap `_candidate_window` the same way `_count` is wrapped, and pass `return_exceptions=True`.

Two related notes: `context_window_limit or DEFAULT_CONTEXT_WINDOW_LIMIT` (`:88`) also swallows an explicit `0`, and when *all* candidates fail to count the strategy picks the largest window and proceeds blind — a `logger.warning` there would help. Cost note for anyone enabling Bedrock's native token counting: this is N `CountTokens` API calls per invocation, plus the `_estimate_input_tokens` call the event loop already makes against `agent.model`.

strands-py/src/strands/models/routing/router.py
```diff
@@ -0,0 +1,215 @@
+"""ModelRouter: a reusable, immutable set of candidate models with a per-call selection strategy.
+
+A router holds an ordered sequence of candidates and is a ``Plugin`` so an agent can accept it
+through ``model=``. Per model call, its ``RoutingStrategy`` selects a candidate; when the selected
```
strandly-the-agent: ⚪ **"Per model call" is wrong, and this file contradicts itself about it.** `_selection_middleware` only calls `_pick` when `invocation_state[_ROUTING_KEY]` is `None` (`:128-136`) and reuses the cached model for every subsequent call in the invocation — which is exactly what design 0016 specifies, and exactly what `_selection_middleware`'s own docstring says 120 lines below (`:125`, "once per invocation"). The same claim appears at `:1`, `:50` ("per-call selection strategy") and `__init__.py:3`.

Flagging it because the equivalent docs-vs-code drift came up on Phase B too — worth a pass over all four sites.

```suggestion
A router holds an ordered sequence of candidates and is a ``Plugin`` so an agent can accept it
through ``model=``. Its ``RoutingStrategy`` selects a candidate once per agent invocation; the
choice is cached and reused for every model call in that invocation, including tool-loop turns.
When the selected model fails and nothing else has claimed the retry, the router advances to the
next untried candidate in declaration order (ordered fallback).
```

That wording also fixes a second inaccuracy in the original: "when the selected model's retries are exhausted" understates the trigger — `_advance_on_failure` (`:144-145`) also advances on a **non-retryable** failure with zero retries consumed, and conversely does *not* advance when a hook has already set `retry=True`. Neither is documented today.

PR #3520, opened 2026-07-28
feat(site): add /community/ editorial hub and 14-lesson course

yonib05:
## Description

The 14-lesson Strands course had no home on the site. This PR adds a `/community/` page where people can find the course, deep dives, upcoming events, and recent blog posts — and adds the course itself under `/docs/learning/`.

How it works:

- **Courses and events are YAML files** (`src/content/courses/`, `src/content/events/`). Adding an event is one small file; the schema catches bad dates and links at build time.
- **Events expire on their own.** The build filters past events, and a small client script hides anything that expires between deploys. No one has to remember to remove an event after a conference.
- **The course YAML is the single source of truth** for lesson order. Lesson URLs carry no numbers, so inserting or reordering lessons later is a YAML edit, not a URL migration.
- **Lesson pages** get a course sidebar with prev/next links, an embedded video (click-to-load, no page-weight cost), and a link to that lesson's code samples.
- **Design reuses the homepage's system** — same font, background animation, card styles, and light/dark theme handling.

Things reviewers should know:

- `navigation.yml` gains a `Community` navbar item and a `Learning` sidebar group. The sidebar group exists so `llms.txt` lists the lessons — the visible lesson sidebar comes from route middleware instead.
- The branch history was squashed to one commit when rebasing onto main (the old history had two structural renames that made replay impractical).
- Follow-ups agreed in review: add code snippets to lessons 2–7 (with the course author), and upstream the lesson transcripts to the course repo.

## Related Issues

N/A — page requested by the team; no tracking issue.

## Documentation PR

Documentation-only change (all under `site/`).

## Type of Change

New feature

## Testing

- Full local gate: 594 tests, typecheck, and `astro build` (887 pages) all pass, with the broken-links checker proving every lesson link resolves. Also verified under a preview-deploy base path.
- Unit tests cover date/expiry edge cases, course ordering, sidebar building, redirects, and the client-side expiry script (jsdom).
- Manual browser checks: desktop and mobile layouts, dark and light themes, keyboard focus, reduced motion, and simulated future dates for event expiry.

## Checklist

----


yonib05:

strandly-the-agent:
## LLM-CONTEXT pass — is `/learn/` + the 14 lessons usable by an agent/crawler?

**TL;DR:** the site's `llms.txt` / `llms-full.txt` / per-page `.md` / `sitemap.xml` surfaces all pick up the 14 new lesson pages and `/learn/` **automatically** — nothing to register. But two things actively hurt an agent-mediated reader:

- 🔴 **Lesson 2's video is wrong.** Its `<YouTube>` embed *and* its manual "Watch on YouTube" link both point to Lesson 1's video (copy-paste bug), and Lesson 2 has zero code in the prose — so there's currently no way to get lesson 2's real content from the docs at all.
- 🟠 **6 of 14 lessons are pure video narration with zero code shown** ("we import X... now we run it... you can see the result coming back") — an LLM/screen-reader reading the text gets none of the actual how-to.

Everything else — auto-registration in all four LLM surfaces, the sitemap test, JSON-LD/OG parity, server-rendered `/learn/` HTML — checked out fine. Details below.

<details>
<summary>Full findings (ranked) + what I checked and found fine</summary>

### 🔴 Blocker

**Lesson 2's video embed points to Lesson 1's video.**
`site/src/content/docs/community/learning/lesson2-switching-model-providers.mdx:7-8`:
```
<YouTube id="ZpXWGjISMs8" title="Lesson 2: Switching Model Providers" />
*[Watch on YouTube](https://www.youtube.com/watch?v=ZpXWGjISMs8&list=PLDzwjhH-4yhU)*
```
That's the *identical* video ID as `lesson1-how-agents-really-work.mdx:7-8`. All 12 other lessons carry a unique ID and an `&index=N` query param matching their lesson number; lessons 1 and 2 are the only two missing `&index=` and the only two sharing an ID — strong signature of a copy-paste-from-lesson-1 mistake that wasn't updated. Since lesson 2 has **no code fences at all** (see below), this link is the only escape hatch to the real content, and right now it 404s-in-spirit — it just silently plays the wrong video.

### 🟠 Should-fix

**6 of 14 lessons (2, 3, 4, 5, 6, 7) have zero code fences** — confirmed by `grep -c '^```' lesson*.mdx`. For a reader who can't/won't watch the embedded video (an LLM answering from docs, a screen-reader user, someone skimming), the prose is screencast narration describing on-screen actions the text never shows. Concrete example, `lesson4-adding-callbacks-and-response-streaming.mdx:36-38`:
> "Let's run this so we can see how each agent behaves differently. So we can see the default streaming coming back here, printing out which tools are being called and then the final result."

...describing a terminal demo with no transcript of the callback handler code or its output anywhere in the file. Same pattern in `lesson2-switching-model-providers.mdx:43-45,66-69` ("Then we create an instance of our model and pass in a specific model ID... Now we can give this a run" / "I have Gemma running locally already"). The code lives only in the video and in `aws-samples/sample-building-with-strands-course` (linked once per lesson) — an agent answering "how do I switch to Ollama with Strands?" from these docs alone gets nothing concrete. Lessons that *do* include fenced code (1, 8, 9, 10-14) don't have this problem — worth applying the same treatment (a short snippet per major step) to 2-7.

**The new `<YouTube>` auto-import degrades to a content-free `[Play](url)` link on every machine-readable surface.** `astro.config.mjs:110-112` registers `@astro-community/astro-embed-youtube` as the `YouTube` auto-import (first use of this component in the repo — `SITE-ARCHITECTURE.md`'s auto-import list doesn't mention it either, per the context pack). The component renders a `<lite-youtube>` custom element wrapping `<a href="https://youtube.com/watch?v=ID"><span class="lyt-visually-hidden">Play</span></a>`. `src/util/html-to-markdown.ts` has no Turndown rule for it, so it falls through to the default link rule and every lesson's `.md`/`llms-full.txt` output opens with a bare `[Play](https://youtube.com/watch?v=...)` — no title, no lesson context. Verified via `npx astro dev` + curl on `/docs/community/learning/lesson1-.../index.md` and `/llms-full.txt` (all 14 instances: `grep -n '^\[Play\]' llms-full.txt` → 14 hits). Not a blocker — the authored `*[Watch on YouTube](...)*` line right below it already carries the real link with more context — but the auto-generated `[Play]` line is noise worth stripping with a small Turndown rule, same as `removeScripts`/`removeSrOnly`.

### 🟡 Nit

**Static `/learn/` event data can go stale for crawlers, not for browsers — the reverse of what you'd expect.** `/learn/` is a fully static (SSG) page; `learn.astro:32-37` computes "upcoming" events from `today = new Date()` **at build time**. `docs-deploy-github-pages.yml` is `workflow_dispatch`-only (no scheduled rebuild), so the static HTML can sit for a while. `learn.astro:79-91` adds a client script — `data-expires` attributes removed if `< today` — as, per its own comment, "the freshness backstop." A real browser gets the correction; a crawler/LLM fetching the raw HTML (no JS execution) sees whatever was true at last build, so it could report an event as upcoming after it's already happened. Confirmed via curl+strip-tags that `/learn/`'s content (including the events ticker) is real server-rendered text, not client-JS-only — so the *content* is crawlable, just potentially stale by the time it's read.

**New lesson entries show raw slugs, not titles, in `llms.txt`.** `src/pages/llms.txt.ts`'s `extractLinks()` falls back to `item.slug.split('/').pop()` when a sidebar item has no explicit `label:`. Since `navigation.yml:311-324` lists the 14 lessons as bare slugs, `llms.txt` renders `- [lesson1-how-agents-really-work](...)` instead of "Lesson 1: How Agents Really Work". **Pre-existing pattern** — e.g. `- [overview](...)`, `- [quickstart](...)` for existing pages already do this — not a regression, but this PR triples down on it for 14 more entries. Cheap fix: add `label:` to each entry in the sidebar group.

### Checked and found fine

- **`llms.txt`** (`src/pages/llms.txt.ts`) auto-includes the 14 lessons: they're under `navigation.yml`'s `Community > Learning` group, and `Community` is one of `SIDEBAR_SECTIONS`. Verified live: `curl localhost:4321/llms.txt` lists all 14 lesson URLs.
- **`llms-full.txt`** (`src/pages/llms-full.txt.ts`) iterates `getCollection('docs')` filtering only `docs/api/*` — lessons render inline automatically. Verified (`## Lesson 1: How Agents Really Work` present, full body rendered).
- **Per-page `/{slug}/index.md`** (`src/pages/[...slug]/index.md.ts`) is a fully dynamic `getStaticPaths` over the docs collection — no hand-maintained list, nothing to add for the 14 lessons or update for this PR.
- **`sitemap.xml`** (`sitemapWithLastmod()` wrapping `@astrojs/sitemap`) crawls Astro's emitted static routes directly, no allow/deny list — `/learn/` and all 14 lessons are included with no filtering needed.
- **`test/known-routes.json` / `test/sitemap-coverage.test.ts`**: this registry is scoped to `redirectFrom` slugs (old-URL → new-URL mapping), not "every page" — irrelevant here since the PR adds no `redirectFrom` frontmatter. Ran `npx vitest run test/sitemap-coverage.test.ts` → 2 passed, 2 skipped (live-sitemap checks gated behind `VERIFY_LIVE_SITEMAP=true`, skipped locally as designed). Nothing to add, nothing broken.
- **`/learn/` is real server-rendered HTML**, not client-JS-only — curled it and stripped tags, got full hero/courses/events/blog text including the lesson list.
- **JSON-LD / OpenGraph parity**: `LandingLayout.astro:130-133` emits the same `baseSchemas()` JSON-LD (`Organization`/`SoftwareApplication`/`WebSite`) as `Head.astro:113` uses for every doc page, and the same OG/Twitter meta shape — confirmed in the rendered `/learn/` HTML. No divergence; the repo has no page-type-specific schema (e.g. `Course`/`Event`) anywhere to diverge *from*, so nothing to flag there.
- Neither `/learn/` nor the homepage/`/blog/` index get a raw-markdown `/index.md` twin — consistent with the existing pattern that only individual collection entries (docs pages, blog posts) get one.
- `robots.txt` unaffected; no `noindex` on the new pages.

</details>

site/src/components/community/EventsPoster.astro
```diff
@@ -0,0 +1,159 @@
...
+}
+
+const { headliner, others } = Astro.props
+---
+
+<aside>
+  <div class="poster">
+    {
+      headliner && (
+        <div id="poster-headliner" data-expires={toIsoDate(headliner.endDate ?? headliner.startDate)}>
```
strandly-the-agent, line 23: 🔴 **The expiry backstop deletes still-upcoming events along with the expired headliner.**

The `.row` divs for the other events (lines 24–34) are nested *inside* `#poster-headliner`, and both carry `data-expires`. `learn.astro:80-84` iterates every `[data-expires]` and calls `el.remove()` once its date is past — so when the headliner expires, the whole subtree goes, taking every still-upcoming sibling row with it. `learn.astro:90-96` then sees no `#poster-headliner` and swaps in the evergreen “Join the Discord” fallback, so the page contradicts itself: the hero ticker (a sibling, unaffected) still lists real upcoming events while the Events poster says there are none.

Reproduced in jsdom against the real compiled script, with a fixture event that ends before the headliner does:

```
--- simulated clock 2026-08-01   (headliner expired, rows have NOT)
  ticker: remaining=["2026-08-06","2026-10-21"]   <-- still lists 2 events
  poster: headlinerPresent=false  rowsLeft=[]      <-- all 3 rows destroyed
  evergreen hidden=false  cal="Join the Discord →"
```

Today's three YAML files happen to dodge it (featured re:Invent is the last to end), but it fires as soon as the headliner isn't the last to end — **including the default case with no `featured:` flag at all**, since `posterEvent` then returns the soonest event, which always expires first.

**Fix:** move the `.list`/`.row` block out of `#poster-headliner` into a sibling, and gate the evergreen fallback on `!poster.querySelector('[data-expires]')` rather than on `#poster-headliner`'s presence.

strandly-the-agent (commented):
**TL;DR — 🔴 two proven correctness bugs + a wrong video; everything else is in good shape.** The design and test hygiene here are genuinely strong; the blockers are all cheap and none of them touch SDK behavior.

**Verified:** branch `pr3520` @ `7c49dfd5` (base `40a48bbd`) · site suite **481 passed / 2 skipped** · `tsc --noEmit` clean · CI's own preview deploy (`pr-cms-3520`) proves `astro build` + the `throwError: true` broken-links checker pass · 2 defects reproduced in jsdom against the real compiled script · 11-mutation probe run against the new tests.

**The three that block:**
1. 🔴 `EventsPoster` — the client expiry script deletes every *still-upcoming* event whenever the headliner expires first (proven repro; fires by default whenever nothing is `featured: true`).
2. 🔴 `community: true` on all 14 lessons renders *“not owned or supported by the Strands team… validate the package”* on first-party course content — two scrolls under `LearnHero`'s “from the team building the SDK”.
3. 🔴 Lesson 2's embedded video is Lesson 1's video (`id="ZpXWGjISMs8"`, confirmed via YouTube oEmbed).

Six 🟡 and a bundled ⚪ list are inline. Two decisions need you, not me — see the **Questions** comment; the lesson-URL one has a window that closes at merge.

<details><summary>What this PR does well (examined and judged correct)</summary>

- **`util/learn.ts` is the strongest part of the change**: every date helper takes `today` explicitly “so behavior is testable”, and it's UTC end-to-end — a 25-case probe under `TZ=Pacific/Kiritimati` (+14), `Pacific/Midway` (−11), `America/Los_Angeles` (DST) and `UTC` produced byte-identical output, 21/21 passing in each.
- **Test quality is high**: 10/10 targeted mutations across expiry, poster selection, ticker limit, course sort, featured/shelf, numeric lesson ordering, prev/next and back-link exclusion were caught by the new tests. Fixtures are synthetic with a fixed `TODAY`, so a future content edit won't break them (verified by dropping in a throwaway lesson 15 + event YAML — suite unaffected).
- **Schemas match the file's own conventions exactly**: line comments (like `sourceLinkSchema`/`blogSchema`), `z.infer` type exports, `glob({base, pattern})` loaders alongside `testimonials`, and `eventSchema.refine` mirroring `changelogFrontmatterSchema.superRefine` — the `endDate >= startDate` invariant is genuinely enforced, not conventional. `LearnEvent` over `Event` (avoiding the DOM global) is the right call.
- **`buildCourseSidebar` reuses `getPrevNextLinks`/`flattenSidebar`** instead of reimplementing pagination, and correctly computes prev/next from the lessons-only group so the back-link stays out of the chain — with the reasoning documented and pinned by tests. `962b5349` replacing positional `courseSidebar[1]` with `find(e => e.type === 'group')` was the right fix.
- **The middleware branch is non-invasive**: mutually exclusive on slug prefix, and every existing export in both touched files is byte-for-byte unchanged — the API-reference sidebar path is unaffected.
- **Components follow the established `landing/` pattern**: typed props, one scoped `<style>`, no `<script>` in section components. `learn.astro`'s documented “sections never query collections themselves” rule is *stricter* than `landing/` and is the better convention.
- **Deep dives as a config array rather than a collection is justified** — the rationale is written down *with* a promotion threshold, and hand-written editorial arrays are established in the sibling directory.
- **Also checked and clean**: no `set:html` anywhere in the new components; adversarial YAML (`"><script>alert(1)</script>`, `onerror=` payloads) renders fully escaped in text *and* attributes; `javascript:` hrefs are neutralized by `pathWithBase` and then flagged as broken links; pagination can't link to a nonexistent page or wrap around; the broken-links checker *does* catch a bad lesson href (drove the real `astro:build:done` hook against a synthetic dist).
- **No API-review label needed.** `team/API_BAR_RAISING.md`'s `api/needs-review` gate is scoped to “public API design and customer-facing contracts — not implementation details or internal/private APIs”; content-collection schemas and sidebar builders are contributor-facing site internals. The one genuinely permanent contract here is the lesson *URLs*, which is a product call, not a label.

</details>

<details><summary>Scope note, and what I deliberately did not flag</summary>

**Not flagged (pre-existing site-wide patterns, not regressions):** `description:` frontmatter length (66/226 existing `.mdx` files already exceed 90 chars) and the remaining >90-char lines, which fall in the table-row/code-fence/URL categories `2edf3293` explicitly excluded. The curve-background CSS duplication is already acknowledged in your PR body as accepted debt, so I've left it alone.

**Process disclosure:** one of my review passes posted its findings directly as [its own comment](https://github.com/strands-agents/harness-sdk/pull/3520#issuecomment-5108791070) before I'd consolidated — that wasn't intended, and I've de-duplicated against it rather than repeating those items here. The one exception is the lesson-2 video, which two other passes independently confirmed, so it's promoted to a blocker below.

**Couldn't verify:** a full `astro build` didn't complete in my sandbox (API-ref generation) — CI's green build substitutes. Lesson 8's benchmark figures, the `agentcore add memory` CLI syntax, and `DeleteApprovalHook`/`LimitToolCounts`' real parameter names are unverifiable here because the `aws-samples` course repo isn't cloned; worth a spot check by someone with access.

This is a machine-generated review — solid work for a human to approve, not a gate. Push back on anything that looks wrong.
</details>

site/src/content/docs/learning/lesson1-how-agents-really-work.mdx
```diff
@@ -0,0 +1,198 @@
+---
+title: "Lesson 1: How Agents Really Work"
+description: "A ground-up look at the agent loop: how a model, a set of tools, and a running context combine to let an agent reason, act, and observe until a task is done."
+community: true
```
strandly-the-agent: 🔴 **`community: true` puts a “not owned or supported by the Strands team” banner on all 14 first-party lessons.**

`MarkdownContent.astro:14` renders `{community && <CommunityContributionAside />}` unconditionally, with no carve-out for `docs/community/learning/*`, so every lesson opens with:

> **Community Contribution** — This is a community-maintained package that is not owned or supported by the Strands team. Validate and review the package before using it in your project. Have your own integration? We'd love to add it here too!

Two scrolls above, `LearnHero.astro:14` sells the same course as “from the team building the SDK”. It's also a wash on the presumable upside: `community: true` pages only enter the catalog tables when `integrationType` is *also* set (`util/integration-content.ts:96-97` short-circuits on it first) and none of the lessons set it — so this is downside-only. And “package” doesn't parse for a tutorial page.

**Fix:** drop `community: true` from all 14 lesson frontmatter blocks. I checked that removal is safe — the flag's only other consumer is `getIntegrationEntries` (gated on `integrationType`), `MarkdownContent.astro:17`'s `languages && !community` branch can't fire (lesson frontmatter is exactly `title`/`description`/`community`), and `Head.astro:40` matches a path segment, not the flag. If you want a distinct visual treatment for lessons, a lesson-specific aside rather than reusing the community-integration one.

site/src/content/docs/learning/lesson2-switching-model-providers.mdx
```diff
@@ -0,0 +1,80 @@
+---
+title: "Lesson 2: Switching Model Providers"
+description: "This covers model providers: how Strands abstracts the underlying LLM so your agent logic stays the same regardless of which model you're calling."
+community: true
+---
+
+<YouTube id="ZpXWGjISMs8" title="Lesson 2: Switching Model Providers" />
```
strandly-the-agent: 🔴 **This is Lesson 1's video.** `id="ZpXWGjISMs8"` is identical to `lesson1-how-agents-really-work.mdx:7`, and YouTube's oEmbed confirms that id's real title is *“How AI Agents Really Work”*:

```
$ curl -s "https://www.youtube.com/oembed?url=https://www.youtube.com/watch?v=ZpXWGjISMs8&format=json"
{"title":"How AI Agents Really Work", ...}
```

Line 8's “Watch on YouTube” link carries the same id, and lessons 1 and 2 are the only two missing the `&index=N` param every other lesson has — so this reads as a copy-paste slip, not shared content. It's live on this PR's preview deploy right now. Since this lesson's prose contains no code, the video *is* the content.

**Fix:** pull lesson 2's real recording id from the `PLDzwjhH-4yhU` playlist (both here and on line 8).

*(Also noted in [this earlier comment](https://github.com/strands-agents/harness-sdk/pull/3520#issuecomment-5108791070) — repeated here because two further independent passes confirmed it, so treat it as confirmed rather than a maybe.)*

site/src/util/learn.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+/** The course spotlighted at the top of the Courses section. */
+export function featuredCourse(courses: Course[]): Course | undefined {
+  return sortCourses(courses).find((c) => c.status === 'available')
+}
+
+/** Future courses shown as shelf slots below the featured course. */
+export function shelfCourses(courses: Course[]): Course[] {
+  return sortCourses(courses).filter((c) => c.status !== 'available')
```
strandly-the-agent: 🟡 **`featuredCourse`/`shelfCourses` don't partition the collection — a second `available` course renders nowhere.**

`featuredCourse` takes `.find(c => c.status === 'available')` (first only) while `shelfCourses` filters out *all* available courses, so available course #2..n reach neither slot. Verified with a 3-course fixture (two available): the second course appears nowhere in the rendered page, the shelf falls through to its `{featured.number + 1}` placeholder at `CoursesSection.astro:95` — so `/learn/` advertises a coming-soon slot numbered 2 while a real course numbered 2 sits in the collection unrendered. No build error, no warning.

Latent today (one course exists), but it bites silently the moment a second course ships. Note `test/learn.test.ts:150-155` currently asserts this behavior as intent, so the suite encodes it rather than catching it — worth updating alongside.

```suggestion
export function shelfCourses(courses: Course[], featured: Course | undefined): Course[] {
  return sortCourses(courses).filter((c) => c !== featured)
```

Call site to update: `pages/learn.astro:34` → `shelfCourses(courses, featured)`. (`statusLabel` at `CoursesSection.astro:12-16` already handles `available`, so an extra available course renders correctly as a shelf slot with no further work.)

site/src/content/docs/learning/lesson5-control-your-agent-with-hooks.mdx
```diff
@@ -0,0 +1,108 @@
...
+
+Inside `register_hooks`, we wire up our callback to the specific lifecycle event we care
+about. In this case, we register `check_delete` to fire on every `before_tool_call` event.
+Inside `check_delete`, the first thing we do is inspect the tool name. If the tool isn't
+`delete_file`, we immediately return and allow execution to continue normally. But if the
+tool is attempting a delete operation, we call `event.interrupt()`. This pauses the agent
+loop and returns control back to our application.
+
+Your application can then prompt the user for approval before resuming with the agent
+operations. If the user declines the action, we call `event.cancel_tool()` with a message
```
strandly-the-agent: 🟡 **Two SDK-usage errors in this lesson that would fail on copy-paste.**

(a) `cancel_tool` is a settable attribute, not a method — `cancel_tool: bool | str = False` on `BeforeToolCallEvent` (`strands-py/src/strands/hooks/events.py:158`, writable per `_can_write` at `:160-161`); the canonical example assigns it (`user-guide/concepts/agents/hooks.mdx:877-880`). Calling `event.cancel_tool(...)` raises `TypeError: 'str' object is not callable`.

```suggestion
operations. If the user declines the action, we set `event.cancel_tool` to a message
```

(b) Line 71 says `hooks=DeleteApprovalHook`, passing a bare class where `Agent.__init__`'s `hooks` param is `list[HookProvider | HookCallback] | None` (`strands-py/src/strands/agent/agent.py:180`). Should be:

```
`hooks=[DeleteApprovalHook()]`, and the agent harness handles the interrupt lifecycle,
```

site/src/route-middleware.ts
```diff
@@ -145,6 +145,28 @@ export const onRequest = defineRouteMiddleware(async (context) => {
...
+      id: doc.id,
+      title: doc.data.title as string,
+      category: doc.data.category as string | undefined,
+    }))
+
+    const courseSidebar = buildCourseSidebar(docInfos, currentSlug)
+
+    // Compute prev/next from lessons only — the "← All courses" back-link
+    // must NOT be part of the lesson prev/next chain.
+    const lessonGroup = courseSidebar.find((entry) => entry.type === 'group')
```
strandly-the-agent: 🟡 **This line is the one piece of new, previously-regressed wiring shipping with zero test coverage.**

I mutated it to drop the `lessonGroup` lookup and feed the *full* `courseSidebar` (including the “← All courses” back-link) into `getPrevNextLinks` — i.e. exactly the bug this branch's own commit `962b5349` (“find lesson sidebar group by type, not position”) already fixed once. Nothing failed: not `learn.test.ts`, not `dynamic-sidebar.test.ts`, not `route-middleware.test.ts`, not the full suite (481 passed / 2 skipped). CI would stay green.

The other 10 mutations I tried across `learn.ts` and `dynamic-sidebar.ts` were all caught, so this is the single hole. `route-middleware.test.ts` wasn't touched by this PR, and it already has an `Integration: Full filtering flow` describe to hang this off:

```ts
it('computes lesson pagination from lessons only, excluding the back-link', async () => {
  const context = makeMockContext('docs/community/learning/lesson2-...')
  await onRequest(context, next)
  expect(context.locals.starlightRoute.pagination.prev?.label).not.toBe('← All courses')
  expect(context.locals.starlightRoute.sidebar[0]).toMatchObject({ label: '← All courses' })
})
```

site/src/components/community/EventsPoster.astro
```diff
@@ -0,0 +1,159 @@
...
+          <span class="kicker">Meet us in person</span>
+          <h3>{headliner.title}</h3>
```
strandly-the-agent: 🟡 **`event.href` is in the schema and set on all three real events, but the poster never renders it.**

`content.config.ts:116` declares `href: z.string().optional()`, all three event YAMLs populate it, and `EventsTicker.astro:25-29` correctly turns it into an `<a>` — but `EventsPoster` never references `event.href` at all: the headliner (`:17-23`) and each row (`:26-32`) are plain `<div>`s. So the *more prominent* Events surface on `/learn/` has no click-through while the same events are clickable in the hero ticker. Confirmed on the preview deploy — the poster's three events are static text.

**Fix:** wrap the headliner/`.row` markup in an `<a href={...}>` mirroring the ticker's `startsWith('http')` internal-vs-external check — or drop `href` from the schema if it's ticker-only by design.

site/src/dynamic-sidebar.ts
```diff
@@ -222,6 +222,53 @@ export function buildTypeScriptApiSidebar(docs: DocInfo[], currentSlug: string):
...
+ * group's entries directly rather than on the full sidebar returned here.
+ */
+export function buildCourseSidebar(docs: DocInfo[], currentSlug: string): SidebarEntry[] {
+  const lessonDocs = docs
+    .filter((doc) => doc.id.startsWith('docs/community/learning/lesson'))
+    .sort((a, b) => {
+      const numA = parseInt(/lesson(\d+)-/.exec(a.id)?.[1] ?? '0', 10)
+      const numB = parseInt(/lesson(\d+)-/.exec(b.id)?.[1] ?? '0', 10)
+      return numA - numB
+    })
```
strandly-the-agent: 🟡 **The membership filter, the number parse, and the middleware's route guard all disagree.**

The filter uses the loose prefix `…/learning/lesson`, the sort key falls back to `'0'` when `/lesson(\d+)-/` doesn't match, and `route-middleware.ts:149` guards on the broader `…/learning/` — three couplings, none checked against each other. Reproduced in real rendered HTML with a `lesson-intro-scratch.mdx` fixture: it gets key `0`, sorts *above* Lesson 1 in the course sidebar, and hijacks Lesson 1's “Previous” link. A non-lesson page under that directory (e.g. `overview-scratch.mdx`) gets the course sidebar it isn't listed in, with no active entry and no prev/next. Zero-padded `lesson01-` also collides with `lesson1-` on key 1, with input order deciding.

No current file triggers any of this — but `content.config.ts:93`'s own comment invites exactly the “course index” page that would.

```suggestion
  const lessonDocs = docs
    .filter((doc) => /^docs\/community\/learning\/lesson\d+-/.test(doc.id))
    .sort((a, b) => {
      const numA = parseInt(/lesson(\d+)-/.exec(a.id)![1], 10)
      const numB = parseInt(/lesson(\d+)-/.exec(b.id)![1], 10)
      return numA - numB
    })
```

Worth applying the same anchored predicate at `route-middleware.ts:149` so the two can't drift apart again.

site/src/content/docs/learning/how-agents-really-work.mdx
```diff
@@ -0,0 +1,198 @@
+---
+title: "Lesson 1: How Agents Really Work"
+description: "A ground-up look at the agent loop: how a model, a set of tools, and a running context combine to let an agent reason, act, and observe until a task is done."
+community: true
+---
+
+<YouTube id="ZpXWGjISMs8" title="Lesson 1: How Agents Really Work" />
+*[Watch on YouTube](https://www.youtube.com/watch?v=ZpXWGjISMs8&list=PLDzwjhH-4yhU)*
+
+*Code for this lesson can be found
```
strandly-the-agent, line 11: 🟡 **Zero internal cross-links across all 14 lessons** — `grep -rn 'user-guide\|/docs/'` over the lesson directory returns nothing; every link is external (YouTube, `aws-samples`). Yet a canonical user-guide page exists for essentially every lesson topic, and none are linked:

| Lesson | Canonical page that exists and isn't linked |
|---|---|
| 2 | `user-guide/concepts/model-providers/index.mdx` |
| 3 | `concepts/tools/mcp-tools.mdx`, `user-guide/shell/index.mdx` |
| 4 | `concepts/streaming/callback-handlers.mdx`, `…/async-iterators.mdx` |
| 5 | `concepts/agents/hooks.mdx` |
| 6 | `concepts/plugins/index.mdx`, `…/plugins/skills.mdx` |
| 7 | `concepts/plugins/steering.mdx` |
| 8 | `concepts/context-management.mdx` |
| 9 | `concepts/agents/session-management.mdx`, `community/session-managers/agentcore-memory.mdx` |
| 10/11/12 | `concepts/multi-agent/{agents-as-tools,graph,swarm}.mdx` |
| 13 | `user-guide/evals-sdk/quickstart.mdx` |
| 14 | `user-guide/deploy/deploy_to_bedrock_agentcore/python.mdx` |

This is the biggest structural gap in the set: a reader who wants the reference detail behind the video has no path to it, and the lessons are the natural top-of-funnel into the user guide. A one-line “Learn more” link per lesson would close it.

site/src/util/learn.ts
```diff
@@ -0,0 +1,68 @@
...
+export function toIsoDate(d: Date): string {
+  return d.toISOString().slice(0, 10)
+}
+
+const MONTH = new Intl.DateTimeFormat('en-US', { month: 'short', timeZone: 'UTC' })
+
+/** "Jan 20", "Dec 1–5" (same month), or "Nov 30 – Dec 2" (cross-month). */
+export function formatEventDate(event: LearnEvent): string {
+  const start = `${MONTH.format(event.startDate)} ${event.startDate.getUTCDate()}`
+  if (!event.endDate || event.endDate.getTime() === event.startDate.getTime()) {
```
strandly-the-agent: ⚪ **Bundled nits** — none blocking, most one-liners. Grouped here to keep the thread count down.

**Docs accuracy (checked against the SDK source in this monorepo):**
- Terminology lock: “built-in tools” where `.agents/references/terminology.md` requires **community tools** (`lesson1:91,110,134`; `lesson3:19,90,114`).
- “AgentCore” spelled three ways: `Agentcore` ×14 in `lesson14` (lines 19–220) and `Agent Core` in `lesson9:109`, `lesson13:156`. Canonical is one word, capital C — and `lesson14`'s own frontmatter and code fences get it right.
- `LlmSteeringHandler` (`lesson7:91`) — real class is `LLMSteeringHandler` (`strands-py/src/strands/vended_plugins/steering/handlers/llm/llm_handler.py:33`); would `ImportError` on copy-paste.
- `lesson2:39` “Claude Sonnet 4” — actual default is `global.anthropic.claude-sonnet-4-6` (`models/bedrock.py:42`; region `us-west-2` is right). Mirrors a pre-existing inconsistency in the Python quickstart, but worth fixing while you're in here.
- `lesson1:90-94` overclaims bare-`Agent()` defaults: there is no default toolset (`agent.py:357-358`), the default `conversation_manager` is `SlidingWindowConversationManager` which **truncates** rather than compressing (proactive compression is opt-in `context_manager="auto"`, `agent.py:239` — which lesson 8 then covers correctly), and sub-agent delegation is always explicit.
- `lesson4:35` “the callback parameter” — the kwarg is `callback_handler` (`agent.py:168`).
- `lesson5:27-29` uses “middleware” as a hooks analogy; `terminology.md` bans it as a synonym, and the SDK has a separate real `strands/_middleware/` subsystem, so it risks term collision.

**Code:**
- `content.config.ts:113-114`: `z.coerce.date()` accepts `startDate: 20260804` (dashes forgotten → a YAML *number*) and coerces it to `1970-01-01`, so the event is permanently past and silently never renders. `z.string().regex(/^\d{4}-\d{2}-\d{2}$/).pipe(z.coerce.date())` closes it. (`2026-02-30` also rolls over to Mar 2.)
- `formatEventDate` (this line) compares *instants*, so `endDate: 2026-12-01T17:00:00Z` on a same-day event renders “Dec 1–1”; the cross-month branch also omits the year, giving “Jan 15 – Jan 15” across years. Compare `toIsoDate(start) === toIsoDate(end)` instead.
- `EventsPoster.astro:119-124`: if the headliner survives but every “other” row expires client-side, `.list` stays in the DOM empty with its `border-top` showing as a stray rule — the script already handles the analogous case for the ticker and the headliner.
- Empty states leave dangling UI: zero lesson files still emits the sidebar group heading with nothing under it (`dynamic-sidebar.ts:252-269`); zero available courses hides the whole `id="courses"` section (`CoursesSection.astro:20-21`) while `LearnHero.astro:18` still renders a `#courses` jump link — and the broken-links checker skips hash-only links, so the build stays green.
- `eventSchema`'s `endDate >= startDate` refine has no test, unlike the analogous `changelogFrontmatterSchema.superRefine` (`test/changelog.test.ts:71-101`). Three lines with `safeParse`.
- Two href conventions in one PR: course hrefs are unconditionally `pathWithBase`'d (`CoursesSection.astro:33,37,73`), event hrefs are sniffed for `http` (`EventsTicker.astro:26`) — an external URL in a course href yields `/https://…`. `z.string().startsWith('/')` on course hrefs would make it a schema error instead of a confusing build failure.
- `href` is required while `status: 'proposed'` is legal (`content.config.ts:92,95`), so a proposed course must invent a URL for a page that doesn't exist; `changelogFrontmatterSchema.superRefine` is the in-repo precedent for tying the two together.
- Event kind is encoded in title strings (`“Ai4 2026 — Booth”`, `“… — Speaking”`). Fine at 3 events; at 30, “show only talks” means re-parsing titles. `kind: z.enum([…]).optional()` is 2 lines today.
- `EventsPoster`'s `(headliner, others)` pair encodes an invariant the types don't — `learn.astro:37` excludes the headliner by *reference identity*, so if the page ever maps/clones event data the headliner silently renders twice. A `posterEvents(events, today, limit)` helper returning `{headliner, others}` would keep the invariant next to the logic that owns it (and retire the magic `slice(0, 3)` at `learn.astro:63`).
- `LearnHero.astro:17-22` duplicates the page's section ids with no type or build check — renaming a section silently breaks the hero jump links.
- `site/AGENTS.md`'s directory-structure section wasn't updated despite its explicit “**MUST** update” rule; this PR adds six new paths (`components/learn/`, `config/learn.ts`, `content/courses/`, `content/events/`, `pages/learn.astro`, `util/learn.ts`). The section was already stale before this PR, so it's a habit rather than a regression — but the rule is explicit. Same for `SITE-ARCHITECTURE.md`'s auto-imported-components list, which doesn't yet mention the new `YouTube` auto-import.

**Not flagged deliberately:** `tickerEvents(events, today, -1)` truncating from the end (no caller), and the UTC client backstop dropping an event ~7h early for US-West visitors on its final day — the module docstring explicitly commits to UTC calendar dates, so that reads as intentional.

strandly-the-agent:
## Questions (from the review above) — two are yours to decide, not mine

**🔴 Blocking-ish, and time-boxed: do you want `lessonN-` in the lesson URLs?**

Right now the filename prefix is doing double duty — it's the permanent public slug *and* the sort key (`dynamic-sidebar.ts:236-239`). Inserting a lesson between 3 and 4, or moving "Deploying" earlier — normal editorial revisions for a 14-part course — renames 11 files, which means 11 new URLs, 11 `redirectFrom` blocks, 11 `known-routes.json` entries (`test/sitemap-coverage.test.ts:156-172` fails until they're added) and 11 title edits.

I checked, and **the window is genuinely open today**: `strandsagents.com/docs/community/learning/lesson1-how-agents-really-work/` returns 404, the live `sitemap-0.xml` (~570 URLs) has zero `learning`/`lesson` entries, `test/known-routes.json` (162 entries) has none either, and no lesson carries `redirectFrom`. So a reorder right now is 14 `git mv`s with zero redirect debt. After merge + one `npm run routes:update`, those 14 paths become a CI-enforced contract on a client-side 404-based redirect system.

So: **drop `lessonN-` and carry order in data, or accept these URLs (and the renumbering cost) permanently?** Either is defensible — it's a product call. One dependency if you take the first option: unnumbered slugs break `dynamic-sidebar.ts:236`'s `startsWith('…/lesson')` filter and the course sidebar renders empty, so it has to land together with an explicit lesson list (next question).

<details><summary>Non-blocking: is a second course coming soon? (decides now-vs-later on two design items)</summary>

`buildCourseSidebar` is single-course by construction, and every failure mode is silent. Membership is a filename prefix (`:236`), identity is a hardcoded `'Agent Fundamentals with Strands'` (`:254` — the third copy of that string, after the YAML title and the docstring), and routing is a directory prefix (`route-middleware.ts:149`). I ran it against a second course at `…/learning/evals-course/lesson1-b`: it passes the route guard, then renders **course 1's lessons under course 1's label**, nothing marked current, and `getPrevNextLinks` returns `{prev: undefined, next: undefined}`. The hardcoded label is the quieter half — retitling `agent-fundamentals.yaml:1` updates `/learn/` and leaves the lesson sidebar stale, and no test asserts it.

The `buildPythonApiSidebar`-also-hardcodes-its-path precedent doesn't quite carry here: there will only ever be one Python API, whereas `courseSchema` has `number`, `status: 'available' | 'in-development' | 'proposed'` and a `lessons` array — it's a catalog schema whose renderer handles exactly one item.

If a second course is near, the cheap shape is to make membership explicit and pass it in:

```ts
export interface CourseSidebarInfo { title: string; lessons: readonly string[] }  // ordered doc ids
export function buildCourseSidebar(docs: DocInfo[], currentSlug: string, course: CourseSidebarInfo): SidebarEntry[]
```
```ts
// route-middleware.ts:149 — replaces the path-prefix guard
const course = (await getCollection('courses')).find((c) => c.data.lessons.includes(currentSlug))
if (course) { /* …unchanged, minus the parseInt sort */ }
```

That also collapses the related seam: the lesson list currently lives in **three** places that must stay in sync — `navigation.yml:310-325` (14 slugs), `agent-fundamentals.yaml:9-51` (14 × `{number,title,href}`), and the filename scan in `dynamic-sidebar.ts:236-241`. If you want one source of truth, I'd pick the course YAML's ordered `lessons` array (position N *is* lesson N, membership explicit), pull titles from the docs collection instead of duplicating them, and generate the `navigation.yml` group in `src/sidebar.ts` — which already imports `fs`/`path`/`js-yaml` and already does existence checks, so it's in-pattern. Rough cost: ~+30 net production lines, mostly trading YAML for code, **plus** a guard test (every id in `courses/*.yaml` resolves in `getCollection('docs')`; no docs id under a course dir is unclaimed) — without that test you've moved the silent failure rather than removed it.

Entirely reasonable to defer all of this until a second course is actually queued. I'd just rather you decide that than inherit it.
</details>

<details><summary>Non-blocking: was `community: true` deliberate?</summary>

If it was aiming at some catalog behavior, worth knowing what — because as shipped it produces only the disclaimer, with none of the catalog upside (that path requires `integrationType`, which none of the 14 lessons set). If it was a copy-paste default from the surrounding `community/` pages, the fix is just 14 deletions (see the inline comment).

Also worth a spot check by someone with access to the course repo: lesson 8's benchmark figures ("cost dropped 55%… accuracy 68% → 98%"), the `agentcore add memory` CLI syntax, and `DeleteApprovalHook`/`LimitToolCounts`' real parameter names — none of those are verifiable from this monorepo, so I've left them as unverified rather than guessing.
</details>

site/src/components/community/EventsPoster.astro
yonib05: Fixed in 9ee01b1d8 — restructured per your suggestion: `.list` moved out of `#poster-headliner` to a sibling, evergreen swap now gated on no remaining `[data-expires]` in the poster, and an all-rows-expired headliner-survives case removes the empty `.list` (stray border). Re-verified all three scenarios with Playwright clock simulation.

site/src/content/docs/learning/lesson1-how-agents-really-work.mdx
yonib05: Fixed in e077fb7b3 — `community: true` removed from all 14 lessons. Good catch; the flag was added for frontmatter symmetry with sibling docs without checking what it renders.

site/src/content/docs/learning/lesson2-switching-model-providers.mdx
yonib05: Fixed in e077fb7b3 — the duplicate id was in the source transcripts themselves. Correct id `NAEX4ORLkbQ` pulled from the playlist and verified via oEmbed (title: "Switching Model Providers"). Both embed and watch link updated.

site/src/util/learn.ts
yonib05: Fixed in 9ee01b1d8 — took your partition-by-identity suggestion: `shelfCourses(courses, featured)` filters `c !== featured`, caller updated, and the test that encoded the old behavior now asserts a second available course renders in the shelf.

site/src/content/docs/learning/lesson5-control-your-agent-with-hooks.mdx
yonib05: Fixed in e077fb7b3 — both: `event.cancel_tool` rephrased as attribute assignment, and `hooks=[DeleteApprovalHook()]`.

site/src/route-middleware.ts
yonib05: Fixed in 9ee01b1d8 — added the middleware integration test per your sketch: lesson pagination asserts prev is Lesson 1 (not the back-link) and sidebar[0] is '← All courses'. Your mutation now fails the suite.

site/src/components/community/EventsPoster.astro
yonib05: Fixed in 9ee01b1d8 — headliner and rows are now anchors when `event.href` is set, mirroring the ticker's internal/external check, with a poster-palette hover.

site/src/dynamic-sidebar.ts
yonib05: Fixed in 9ee01b1d8 — anchored `/^docs\/learning\/lesson\d+-/` predicate in buildCourseSidebar (sort keys parsed from the guaranteed match) and the same regex guards the middleware branch, so non-lesson pages under docs/learning/ fall through to the normal sidebar. Tests cover the non-lesson exclusion. (Paths reflect the docs/learning/ move that landed after your review.)

site/src/content/docs/learning/how-agents-really-work.mdx
yonib05: Fixed in e077fb7b3 — every lesson now ends with a 'Learn more' link to its canonical user-guide page from your table (targets verified against the current tree; build link-checker confirms all resolve).

site/src/util/learn.ts
yonib05: Addressed in 9ee01b1d8 + e077fb7b3 — took: terminology (community tools, AgentCore, LLMSteeringHandler, Sonnet 4.6 default, lesson-1 defaults corrections, callback_handler, middleware analogy removed), date-schema hardening (rejects YAML numbers; safeParse test), formatEventDate ISO-date comparison, endDate refine test, internal-href schema check on courses, empty-list client cleanup, zero-lesson sidebar, courses-section shell for the #courses anchor, AGENTS.md structure + SITE-ARCHITECTURE.md YouTube auto-import. Deferred as speculative (happy to revisit): event `kind` enum, `posterEvents` refactor, hero section-id type check, href-required-for-proposed refine.

yonib05:

site/test/route-middleware.test.ts
```diff
@@ -245,6 +246,30 @@ describe('Integration: Full filtering flow', () => {
       expect(link.href.startsWith('/docs/integrations/')).toBe(true)
     })
   })
+
+  it('lesson pagination prev is the previous lesson, not the All-courses back-link', () => {
```
strandly-the-agent, line 247: 🟡 **This test doesn't cover what the thread reply says it covers — the round-1 gap is still open.**

It imports `onRequest`'s *dependencies* (`buildCourseSidebar`, `getPrevNextLinks`) but not `onRequest` itself, and re-implements the lesson branch inline — the comment on line 251 says as much (“Mirrors the middleware's lesson branch”). So a regression in the real middleware can't fail it.

I re-ran the exact round-1 mutation against the real `onRequest` (`route-middleware.ts:181` — replacing the `lessonGroup`-derived `lessonsOnly` with the full `courseSidebar`, putting the “← All courses” back-link back into the prev/next chain):

```
$ npx vitest run test/learn.test.ts test/dynamic-sidebar.test.ts test/route-middleware.test.ts test/redirect.test.ts
 Test Files  4 passed (4)
      Tests  112 passed (112)      # identical with the mutation applied
```

So “Your mutation now fails the suite” isn't accurate — no hard feelings, it's an easy thing to believe when the new test exercises the same helpers. The mirrored test is still worth keeping (it pins `buildCourseSidebar` + `getPrevNextLinks` composition); it just isn't a backstop for `onRequest`.

`onRequest` is exported (`route-middleware.ts:109`), so a real integration test is cheap — build a minimal `context` with `locals.starlightRoute`, call it, and assert on `starlightRoute.pagination` and `sidebar[0]`. Not blocking a draft, but this is the one file where a silent regression ships.

site/src/content.config.ts
```diff
@@ -158,6 +158,54 @@ const blogSchema = z.object({
   readingTime: z.string().optional(),
 })
 
+// Course links are always site-internal, so a full external URL is a schema error.
+const internalHref = z.string().startsWith('/', 'course hrefs must be site-relative (start with /)')
```
strandly-the-agent: 🟡 **`startsWith('/')` doesn't enforce “site-relative” — `//evil.com` and `/\evil.com` both pass and resolve off-site.**

The comment just above states the intent (“a full external URL is a schema error”), and it correctly rejects `https://` and `javascript:` — but it misses the one external form that *reads* as a path in a YAML diff:

```
ACCEPT  "/docs/learning/lesson1-x"  -> https://strandsagents.com/docs/learning/lesson1-x
REJECT  "https://evil.com"
ACCEPT  "//evil.com"                -> https://evil.com/          <-- 
ACCEPT  "/\\evil.com"               -> https://evil.com/          <-- 
REJECT  "javascript:alert(1)"
```

Confirmed end-to-end: setting `href: "//evil.com/pwned"` in `agent-fundamentals.yaml` and rendering `/community/` emits `<a class="btn ..." href="//evil.com/pwned">Start the course →</a>` with no schema or build error. Applies to `href` (`:171`), `syllabusHref` (`:172`) and `lessons[].href` (`:178`).

Not live-user-reachable — course YAML arrives via PR, and there's no `set:html` — so this is about the guard doing its stated job rather than an active exploit.

```suggestion
const internalHref = z
  .string()
  .regex(/^\/(?![/\\])/, 'course hrefs must be site-relative (start with a single / )')
```

I verified that variant still accepts `/docs/learning/lesson1-x` and now rejects both `//evil.com` and `/\evil.com`. Worth a `courseSchema` test too — `grep -rl courseSchema test/` is currently empty, so removing this guard entirely wouldn't fail anything.

site/src/pages/community.astro
```diff
@@ -0,0 +1,300 @@
...
+    const poster = document.querySelector<HTMLElement>('.poster')
+    if (poster) {
+      const list = poster.querySelector<HTMLElement>('.list')
+      if (list && !list.querySelector('[data-expires]')) list.remove()
+      if (!poster.querySelector('[data-expires]')) {
+        const evergreen = document.getElementById('poster-evergreen')
+        if (evergreen) evergreen.removeAttribute('hidden')
+        const cal = document.getElementById('poster-cal')
+        if (cal) cal.textContent = 'Join the Discord →'
+      }
```
strandly-the-agent: 🟡 **The “headliner expired, rows still upcoming” case leaves a headless poster.**

Your comment on lines 89-92 shows you deliberately fixed the data-loss half, and that part is solid — no live event is deleted any more, and I couldn't break it across 9 simulated clocks on the shipping content. What's left is presentation: there's a branch for *rows expired, headliner alive* (remove `.list`) and for *everything gone* (evergreen), but none for *headliner gone, rows alive*. `#poster-headliner` carries the poster's only kicker and `<h3>`, and `.list` has `border-top: 1px solid #0f3322; margin-top: 1.5rem` (`EventsPoster.astro:184-189`) which assumes something above it.

Repro — rendered HTML plus the real compiled script in jsdom, with `ai4-2026` marked `featured` (ends 2026-08-06) while `reinvent` (2026-12-04) isn't:

```
### today=2026-08-07
{"headliner":null,"list":true,"rows":["2026-10-21","2026-12-04"],
 "posterFirstChildClass":"list","evergreenHidden":true,
 "calText":"Event updates on Discord →"}
-> stray .list border with no kicker/h3 above it; cal text still implies a headliner
```

`posterFirstChildClass: "list"` is the tell. **Not live today** — it needs a maintainer to mark an earlier-ending event `featured`, and today's content (reinvent featured, latest end date) never hits it. But that's exactly what `featured` is for, so it'll misrender the first time the headliner rotates.

```suggestion
      if (!poster.querySelector('[data-expires]')) {
        const evergreen = document.getElementById('poster-evergreen')
        if (evergreen) evergreen.removeAttribute('hidden')
        const cal = document.getElementById('poster-cal')
        if (cal) cal.textContent = 'Join the Discord →'
      } else if (!poster.querySelector('#poster-headliner') && list) {
        list.classList.add('list--bare')
      }
```

…plus a `.list--bare { border-top: none; margin-top: 0 }` rule. Alternatively move the kicker + a generic `<h3>` out of `#poster-headliner` so a heading always survives.

Worth noting this file and `EventsPoster.astro` have **no tests at all** (`grep -rl "community.astro\|EventsPoster" test/` is empty). This area has now produced a real bug twice; a small jsdom test driving the expiry DOM logic would be the highest-value test to add in this PR.

site/src/content.config.ts
```diff
@@ -158,6 +158,54 @@ const blogSchema = z.object({
...
+      })
+    )
+    .optional(),
+})
+export type Course = z.infer<typeof courseSchema>
+
+// Event dates arrive as Date objects (unquoted YAML dates) or ISO strings
+// (quoted YAML). Restricting the input to those two forms rejects YAML
+// numbers, which bare z.coerce.date() would accept as epoch milliseconds;
+// malformed strings still fail the Date validity check after coercion.
```
strandly-the-agent: ⚪→🟡 **This last clause isn't true: out-of-range calendar dates don't fail, they roll over silently.**

The YAML-number rejection genuinely works (that was the point, and it's verified). But “malformed strings still fail the Date validity check after coercion” doesn't hold for out-of-range dates:

```
REJECT  20260804 (bare YAML number)              <- the fix works ✓
ACCEPT  2026-13-01  (unquoted)  -> 2027-01-01     <- month 13 rolls into next year
ACCEPT  "2026-02-30" (quoted)   -> 2026-03-02
ACCEPT  "08/04/2026", "2026-8-4", "2026" -> 2026-01-01
```

End-to-end, a typo'd event publishes with the wrong date and no error — `data-expires="2027-01-01"`, rendered label “Jan 1”. Given events auto-expire off this value, a silently-wrong date means an event that vanishes or lingers for a year.

Cheapest correct fix is to make the string branch round-trip: `.refine(d => toIsoDate(d) === original)` for string input, or drop the string branch and require unquoted YAML dates. If you'd rather not, just soften the comment so the next reader doesn't trust it.

site/src/dynamic-sidebar.ts
```diff
@@ -222,6 +222,67 @@ export function buildTypeScriptApiSidebar(docs: DocInfo[], currentSlug: string):
   return entries
 }
 
+/**
+ * Matches lesson page ids (docs/learning/lesson3-give-your-agent-tools-...)
+ * and captures the lesson number. Anchored so non-lesson pages under
+ * docs/learning/ (e.g. a course index) never match. Shared by the route
+ * middleware, which uses it to decide whether a page gets the course sidebar.
+ */
+export const LESSON_ID_PATTERN = /^docs\/learning\/lesson(\d+)-/
```
strandly-the-agent: 🟡 **The anchoring this pattern exists for isn't actually covered — two mutations slip through all 45 tests.**

Both of these pass the whole suite:
- loosening membership at `:245` to a bare `doc.id.startsWith('docs/learning/lesson')`
- dropping the digit requirement here: `/^docs\/learning\/lesson(.+)-/`

The existing `excludes non-lesson pages` test uses `docs/learning/overview`, which a bare `startsWith('...lesson')` would exclude anyway — so nothing exercises the anchor+digit requirement that replaced the old loose prefix. A lesson-*like* fixture is what's missing:

```ts
it('excludes a lesson-like id with no lesson number', () => {
  const docs: DocInfo[] = [
    { id: 'docs/learning/lesson1-how-agents-really-work', title: 'Lesson 1' },
    { id: 'docs/learning/lesson-overview', title: 'Course Overview' },
  ]
  const group = buildCourseSidebar(docs, '').find((e) => e.type === 'group')
  expect(group?.type === 'group' && group.entries).toHaveLength(1)
})
```

Same shape of gap in `util/learn.ts:62-64`: dropping `getUTCFullYear()` from the `sameMonth` check passes all 26 date tests, because no cross-year fixture exists. The code is right in both places today — I just couldn't find a test that would notice if someone reverted it.

strandly-the-agent (commented):
**Round 2 — 9 of round 1's 10 fixes verified genuinely fixed.** Nice work; this was a big, good-faith pass and it holds up under independent re-verification. No 🔴 remains.

**Verified at `e077fb7b3`** (base `2f907793a`, re-review scope `f1afb3574..HEAD` = 37 files, +540/-234): the 4 PR-relevant test files pass **112/112** · full suite **557 passed / 4 failed / 2 skipped**, and I confirmed both failures (`api-counterparts`, `api-link-converter`) are untouched by this PR and caused by API-doc artifacts I stubbed locally · `tsc` gives 24 errors **byte-identical on head and merge base**, so no new type errors from this PR.

**Two things worth attention before this leaves draft:**

1. 🟡 **The middleware-test reply isn't accurate.** I re-ran the exact round-1 mutation against the real `onRequest` and the suite still passes 112/112 — the new test mirrors the lesson branch rather than calling `onRequest`, so it can't catch it. That round-1 gap is still open. Repro in the inline comment.
2. 🟡 **`internalHref` (added this round) accepts protocol-relative URLs** — `//evil.com` passes `startsWith('/')` and renders as the hub's primary CTA pointing off-site. Verified end-to-end.

Four smaller 🟡 items and a ⚪ bundle are inline; two open questions (a dead sidebar group, and the `learn.ts` naming) are in a separate comment.

<details><summary>Round-1 fixes I re-verified as genuinely fixed (credit due)</summary>

- **Poster restructure** — `.list` is genuinely a sibling of `#poster-headliner` now (`EventsPoster.astro:19-71`); I traced all four headliner×rows combinations. The round-1 symptom is now *structurally impossible*: poster events ⊇ ticker events for every input, so the poster can never empty while the ticker still lists live events. The script survived 9 simulated clocks (2026-08-05 → 2099-06-15) on the shipping content with no live event ever deleted.
- **`community: true`** gone from all 14 lessons; nothing else triggers `CommunityContributionAside` for them.
- **lesson2 video** — `NAEX4ORLkbQ` confirmed via oEmbed as “Switching Model Providers”; embed *and* watch link both updated; all 14 ids now distinct.
- **Course partition** — `shelfCourses(courses, featured)` filtering `c !== featured` keeps a second `available` course, and the `featured === undefined` case matches the old behaviour (no regression).
- **lesson5 SDK errors** — both correct against `strands-py/src/strands/hooks/events.py:229` and `Agent(hooks: list[...])`.
- **Clickable events** — anchors added, mirroring the ticker's internal/external check.
- **Lesson predicate** — a *single* exported `LESSON_ID_PATTERN` imported by `route-middleware.ts:3`, not two copies that could drift.
- **All 14 cross-links** exist and resolve; every lesson ends with exactly one “Learn more” line, no duplicates, no broken endings.
- **Docs sweep** — all 10 accuracy/terminology claims verified against SDK source: `built-in tools`→`community tools` (0 hits repo-wide), `Agentcore`/`Agent Core` (0 hits), `LLMSteeringHandler`, `Claude Sonnet 4.6`/`us-west-2` per `bedrock.py:42,44`, lesson1's corrected `Agent()` defaults per `agent.py:342,538-547`, `callback_handler`, and the banned “middleware” analogy replaced.
- **AGENTS.md / SITE-ARCHITECTURE.md** updates are accurate — the round-1 “MUST update the directory tree” gap is closed.

</details>

<details><summary>The rename + redirects: I tried hard to break this and couldn't</summary>

The `/learn/` → `/community/` and `docs/community/learning/` → `docs/learning/` move is the big new surface this round, and it stood up:

- **Redirect graph is clean** — 0 of 84 `STATIC_SLUG_REDIRECTS` targets resolves again (no chains), 0 two-hop chains among the 76 generated stubs, all 58 distinct internal targets exist.
- **Ordering holds** — all 14 explicit lesson entries win over `COMMUNITY_PREFIX_RULE`, so none leaks to `docs/integrations/learning/*` (the thing your code comment calls out).
- `/learn/`, `/learn`, `learn`, `/learn//`, `//learn/`, `/latest/learn/`, `/1.x/learn/`, `/1.5.x/learn/` all resolve to `community[/]`; versioned + `/documentation/docs/` prefixes strip correctly for lessons too.
- **`known-routes.json`'s re-sort dropped nothing** — compared as *sets*, not text: 163 → 177 unique, **+14 / −0**, no duplicates.
- **Drift is gated** — deleting a lesson redirect entry fails `every known route resolves to a valid CMS entry`. `redirectFrom` frontmatter duplicating the 14 entries is fine because `buildStaticRedirects` fail-fasts if the two disagree.
- **Navbar behaviour is correct**, not the mismatch I worried about: `findCurrentNavSection`'s longest match picks `/docs/learning/`, so lesson pages highlight **Community**, `/docs/user-guide/*` → Docs, `/docs/integrations/*` → Integrations. Confirmed against rendered HTML.
- **Schema hardening works for its stated target** — a bare YAML number is rejected, and `endDate < startDate` is refused.
- No `set:html` in the new components; adversarial YAML renders escaped.

</details>

<details><summary>Appendix — non-blocking, pre-existing (2). Say the word and I'll file these as issues.</summary>

Both are outside this PR's diff, so they're not yours to fix here:

- `site/src/util/redirect.static.ts:171` vs `:180` — a redirect **source** is validated only against docs content, while **targets** are validated against docs content ∪ `src/pages`. So a source colliding with a live Astro page (e.g. `{'community': 'integrations'}`) ships a stub over that page with no build error. Now that a redirect namespace overlaps an Astro page family, it's the natural next mistake. One-line fix: `if (slugs.has(source) || pageSlugs.has(source)) throw`.
- `site/src/content/docs/user-guide/concepts/model-providers/amazon-bedrock.mdx:173` still says “uses the Claude Sonnet 4 model by default” — exactly the staleness you just fixed in lesson2, on the canonical page lesson2 now points readers to. `bedrock.py:42` says `claude-sonnet-4-6`.

**Couldn't verify** (unchanged from round 1, both pre-existing content): lesson14's `agentcore add memory` subcommand has no match anywhere in this repo (`create`/`deploy`/`invoke` do check out), and lesson8's “cost −55%, accuracy 68%→98%” figures have no source in the repo. A full `astro build` doesn't complete in my sandbox — CI's preview deploy is the better signal there.

**Process note:** two subagent mishaps this round, both sandbox-only with zero effect on the PR or repo — one pass's `edit` produced corrupted output (caught and reverted), and another killed a sibling pass's dev server with an over-broad process match. All worktrees ended with `git status` clean; nothing committed or pushed.
</details>

site/src/content/docs/learning/lesson5-control-your-agent-with-hooks.mdx
```diff
@@ -0,0 +1,111 @@
...
+each other or interfering.
+
+Human in the loop is one of the most common reasons to use hooks. Let's say you're
+building an agent that runs locally on someone's machine, and one of the things it can do
+is manipulate files. What you might want is a hook that pauses the agent before any file
+deletion happens and asks the user for approval. The user can see what's about to be
+deleted, approve or deny the action, and the agent either continues or cancels based on
+that response.
+
+### Coding a Human-in-the-Loop File Deletion Gate
```
strandly-the-agent: 🟡 **Banned terminology your own sweep just missed.**

`.agents/references/terminology.md` bans “human-in-the-loop” as a synonym for **interrupts** — and this heading describes exactly the interrupt feature. Three occurrences, all in files this PR adds (so in scope): here, plus `lesson1-how-agents-really-work.mdx:146` (“Observing Self-Healing & Human-in-the-Loop Operations”) and its prose nearby.

Worth catching because the rest of the terminology pass in `e077fb7b3` was thorough — `built-in tools`, `Agentcore`/`Agent Core`, and the `middleware` analogy are all at zero hits repo-wide now.

```suggestion
### Coding an Interrupt-Gated File Deletion
```

site/src/config/learn.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * Editorial configuration for the /learn/ page.
```
strandly-the-agent: ⚪ **Bundled nits** — all optional, grouped to keep the thread count down.

- **Stale `/learn/` doc comments after the rename**: this line, and `site/src/util/learn.ts:4`, both still say “the `/learn/` page”. (`redirect.ts:40`'s mention is fine — it's historical.)
- **`lesson4-adding-callbacks-and-response-streaming.mdx:36`** — the rewritten sentence that introduced `callback_handler` is 185 chars, vs ~88-90 for every other body line in that paragraph. Not one of the excluded categories (table row / code fence / URL), so worth a rewrap.
- **Cross-link precision** (all three resolve and none is a dead end, since the linked pages link onward — purely “could be sharper”): `lesson6` links `concepts/plugins/` though ~80% of the lesson is specifically the Skills plugin, where `concepts/plugins/skills/` is the precise target; `lesson3` gives Strands Shell its own subsection (`:98-107`) without linking `user-guide/shell/`; `lesson4` covers the async-iterator pattern (`:52-66`) but links only callback-handlers.

strandly-the-agent:
## Questions (round 2) — both non-blocking

**1. The `Learning` sidebar group in `navigation.yml:315-330` renders on zero pages that exist today. Delete it, or keep it?**

I checked this because the group's 14 `docs/learning/*` slugs sit in the **Docs** sidebar tree while the `Community` navbar tab is the one claiming `/docs/learning/`. Pulling real HTML for five page types, it never renders:

| page | sidebar it gets | `Learning` group? |
|---|---|---|
| `/docs/learning/lesson1-…/` | `buildCourseSidebar` — middleware replaces the sidebar wholesale | no (renders `← All courses` + `Agent Fundamentals with Strands`) |
| `/docs/user-guide/…` | navigation.yml, filtered to `/docs/user-guide/` | no |
| `/docs/contribute/`, `/docs/labs/` | navigation.yml, filtered to the Docs basePaths (which exclude `/docs/learning/`) | no |
| `/community/` | none — standalone Astro page, no Starlight sidebar | no |

**My recommendation: delete it for now** — and the reason is more than “it's unused today”. The obvious defence is that it's a deliberate fallback for the future course-index page your own comment at `route-middleware.ts:158-160` anticipates. I tested that by temporarily adding `docs/learning/index.mdx`, and the group *does* start rendering — but with its **label stripped**, because `filterSidebarByBasePath`'s single-group unwrap applies recursively (once for `Docs`, once for `Learning`), leaving 14 bare links with no heading. So the fallback doesn't actually deliver what it looks like it delivers; it ships a latent bug for whoever builds that page.

Either delete it and re-add it (with the unwrap guarded) when the course index actually exists, or keep it and add a one-line comment saying it's intentionally inert — but I'd rather you make that call knowingly than inherit it. Worth noting it's also the fourth copy of the 14-lesson list, after `agent-fundamentals.yaml`, `redirect.ts`, and `known-routes.json`; only the redirect direction is currently drift-gated.

**2. Did `util/learn.ts` / `config/learn.ts` / `test/learn.test.ts` keep their names on purpose?**

The components and content directories were renamed for `/community/`, but these three kept `learn`. Deliberate (avoiding file-name churn and the review noise a rename adds) or an oversight? No strong opinion — the only concrete cost is the two stale doc comments I flagged in the nit bundle. If it's deliberate, ignore this.

---

Also, from the appendix on my review: two **pre-existing** items outside this PR's diff — the `buildStaticRedirects` source-vs-target validation asymmetry (`redirect.static.ts:171` vs `:180`) and `amazon-bedrock.mdx:173`'s stale “Claude Sonnet 4” default. I deliberately didn't file issues unprompted; say the word and I'll open them.

site/test/route-middleware.test.ts
yonib05: You were right — the round-1 test exercised the helpers, not `onRequest`, and my "mutation now fails" claim was wrong. Fixed in 0c0a03e68: true integration test calling the exported `onRequest` with a lesson context against the real collection. Verified independently this time: your exact mutation → 1 failed; revert → 27 passed. The mirrored composition test stays.

site/src/content.config.ts
yonib05: Fixed in 0c0a03e68 — took your regex (`/^\/(?![/\\])/`) as a shared `internalHref` schema for href/syllabusHref/lessons[].href, plus courseSchema safeParse tests covering `//evil.com` and `/\\evil.com` rejects.

site/src/pages/community.astro
yonib05: Fixed in 0c0a03e68 — took your `list--bare` branch for headliner-gone-rows-alive, and your broader point about zero coverage: the expiry logic now lives in `src/util/expire-events.ts` and community.astro imports it, so the new jsdom suite (test/expire-events.test.ts) drives the real function across all four states (all-expired, headliner-only-expired, rows-only-expired, nothing-expired) plus ticker emptying.

site/src/content.config.ts
yonib05: Fixed in 0c0a03e68 — you're right, my comment overstated what coercion catches. Event dates now require Date objects (unquoted YAML) or strict `YYYY-MM-DD` strings with a round-trip check, so `2026-13-01`, `\"2026-02-30\"`, and `08/04/2026` are all schema errors (tested).

site/src/dynamic-sidebar.ts
yonib05: Fixed in 0c0a03e68 — added your lesson-like fixture (`lesson-overview` excluded) and the two cross-year formatEventDate cases (cross-year range renders the cross-month form; same-month-different-years does not collapse to the en-dash form).

site/src/content/docs/learning/lesson5-control-your-agent-with-hooks.mdx
yonib05: Fixed in 0c0a03e68 — took your heading suggestion for lesson 5 and reworded lesson 1's heading/prose to interrupt terminology; case-insensitive sweep of all 14 lessons is clean.

site/src/config/learn.ts
yonib05: Fixed in 0c0a03e68 — stale `/learn/` comments updated to `/community/`, lesson 4's 185-char line rewrapped, and all three cross-link precision items taken (lesson6→skills, lesson3 inline Shell link, lesson4 adds async-iterators). Link checker green.

site/src/content/courses/agent-fundamentals.yaml
```diff
@@ -0,0 +1,51 @@
+title: Agent Fundamentals with Strands
+number: 1
+status: available
+description: >-
+  Fourteen lessons, one production agent. Build up from the agent loop through
+  tools, hooks, steering, memory, and multi-agent patterns — then deploy and
+  evaluate it.
+href: /docs/learning/lesson1-how-agents-really-work/
+lessons:
```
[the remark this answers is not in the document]
yonib05: Fixed in 65ac48d96 by going further than deduplication: lesson slugs dropped their number prefixes (while no URLs are live — per the earlier time-boxed question) and the course YAML's ordered `lessons[]` array is now the single source of truth for membership and order. Titles resolve from the docs collection at render time (can't drift), the sidebar builder takes the explicit lesson list instead of scanning filenames, and a guard test asserts every YAML href resolves to a docs entry.

site/src/route-middleware.ts
```diff
@@ -158,6 +158,30 @@ export const onRequest = defineRouteMiddleware(async (context) => {
     return
   }
 
+  // Check if we're on a lesson page within the learning course. Only lesson
+  // pages get the course sidebar — other pages under docs/learning/ (e.g. a
+  // future course index) fall through to the normal community-section sidebar.
+  if (LESSON_ID_PATTERN.test(currentSlug)) {
+    const docs = await getCollection('docs')
+    const docInfos: DocInfo[] = docs.map((doc: { id: string; data: { title: unknown; category?: unknown } }) => ({
```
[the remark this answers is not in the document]
yonib05: Fixed in 65ac48d96 — extracted `loadDocInfos()`, shared by the API and course branches.

yonib05:
Answers to the open questions:

**`lessonN-` in URLs (time-boxed)** — took the window: 65ac48d96 drops the numbers from all 14 slugs while nothing is live. Order and membership now live in the course YAML's ordered `lessons[]` array (single source of truth); titles resolve from the docs collection at render time; the filename-scan sidebar builder and its parseInt sort are gone. Inserting or reordering lessons is now a YAML edit with zero URL churn.

**Second course timing** — no second course is queued, but the multi-course correctness came along for free: the middleware now matches the current page against each course's lesson list instead of a path prefix, so a second course YAML Just Works. The deeper single-source refactor you sketched is effectively what landed.

**`Learning` group in navigation.yml** — kept deliberately, with labels added and a comment: your llms-context pass showed `llms.txt` builds its lesson listing from this group, so deleting it would drop the lessons from the LLM surface. The labels also fix the raw-slug entries you flagged there. The latent single-group-unwrap label bug for a future course index page is noted in the comment.

**`util/learn.ts` naming** — deliberate: internal modules kept their names to avoid rename churn; the stale doc comments you found are fixed.

**`[Play]` markdown noise** — fixed in 2a8552dd8 (Turndown rule strips lite-youtube elements; the authored Watch-on-YouTube links remain).

**Lessons 2–7 zero code fences** — agreed and deferred to a follow-up: adding snippets from the course samples repo is content work we want the course author to bless rather than synthesize in this PR.

**The two pre-existing issues** (buildStaticRedirects validation asymmetry, stale Sonnet 4 in amazon-bedrock.mdx) — yes, please file both.

site/src/util/html-to-markdown.ts
```diff
@@ -60,6 +60,26 @@
...
+      if (node.nodeName === 'LITE-YOUTUBE') return true
+      // Also remove the inner Play anchor that lite-youtube injects
+      if (node.nodeName === 'A') {
+        const el = node as Element
+        const className = el.getAttribute?.('class') || ''
+        if (className.includes('lty-playbtn')) return true
+        // Bare anchor with no text that links to a YouTube URL (lite-youtube fallback link)
+        const href = el.getAttribute?.('href') || ''
+        const text = el.textContent?.trim() || ''
+        if (href.includes('youtube.com') && text.toLowerCase() === 'play') return true
```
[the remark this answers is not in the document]
yonib05: Fixed in 13440a635 — the fallback-anchor check now parses the URL hostname against an explicit allowlist (youtube.com / youtube-nocookie.com + www variants) instead of substring matching.

.gitignore
```diff
@@ -75,3 +75,5 @@ bin/
 .vitest*
 .local/
 
+.superpowers/
```
Unshure, line 78: just for my own curiosity, what are superpowers?

site/scripts/astro-broken-links-checker-index.js
```diff
@@ -63,6 +63,7 @@ export default function astroBrokenLinksChecker(options = {}) {
         });
 
         await Promise.all(checkHtmlPromises);
+
```
Unshure: nit: Leaving this comment as a reminder to introduce a linter that flags unnecessary newlines like this

Unshure (commented):
High level feedback:
- We have different navigation mechanisms across the different top level pages. Docs has a sidebar navigation, while community has links at the top of the page to different sections. Might make sense to add sidebar navigation to community too

.gitignore
yonib05: Claude code artifact for it's special skills, I added to keep repo clean.

site/scripts/astro-broken-links-checker-index.js
yonib05: https://github.com/strands-agents/harness-sdk/pull/3841

yonib05: Blank line removed. A lint rule for it is a good shout — noting it for a follow-up rather than growing this PR's tooling footprint.

site/src/content/events/agntcon-mcpcon-europe-2026.yaml
```diff
@@ -0,0 +1,5 @@
+title: "AGNTCon + MCPCon Europe — Speaking"
+startDate: 2026-09-17
```
Unshure, line 2: not a blocker, but do these get removed from the site when their end date passes?

site/AGENTS.md
```diff
@@ -78,14 +82,16 @@ When authoring or reviewing documentation pages, follow the voice guide and the
...
 │   ├── data/                     # Bot-maintained data (catalog-stats.json — updated by the
 │   │                             #   catalog-stats workflow; do not hand-edit)
 │   ├── layouts/                  # Custom layouts
-│   ├── pages/                    # Astro pages (incl. integrations.astro — the /integrations page)
+│   ├── pages/                    # Astro pages (incl. integrations.astro — the /integrations page,
+│   │                             #   and community.astro — the /community/ hub)
 │   ├── plugins/                  # Remark/Rehype plugins
 │   ├── styles/                   # Global styles
-│   └── util/                     # Utility functions
+│   └── util/                     # Utility functions (incl. learn.ts — course/event selection helpers)
```
Unshure: nit: Better to not include changes specific to this pr in the agents.md update. Utils should be generic, not specific to learn.ts

Unshure (dismissed):

site/src/content/events/agntcon-mcpcon-europe-2026.yaml
yonib05: They auto-hide on the site based on date parsing. So it's easier for us to manage. Also, Morgan will be owning these already chatted with her.

site/AGENTS.md
yonib05: ack will remove

yonib05: Removed in the latest push — config/, pages/, util/, and the learning/ entry are back to generic descriptions; new directories stay listed per the structure-update rule.

Issue #1659, opened 2026-02-10
[FEATURE] MCP Specification 2025-11-25


yonib05:
## Summary

The Model Context Protocol (MCP) released a new spec revision ([2025-11-25](https://modelcontextprotocol.io/specification/2025-11-25/changelog)) on the one-year anniversary of MCP going public. This revision introduces several production-readiness features driven by community feedback, including enhanced OAuth/auth flows, icon metadata, updated elicitation, and more.

This issue tracks the work needed to update our SDK implementation to conform to the 2025-11-25 spec, which supersedes the [2025-06-18](https://modelcontextprotocol.io/specification/2025-06-18) revision.

> **Note:** Tasks (SEP-1686) and notifications are tracked in a separate issue.

Full diff: [2025-06-18...2025-11-25 on GitHub](https://github.com/modelcontextprotocol/specification/compare/2025-06-18...2025-11-25)

---

### OpenID Connect Discovery for Auth Server Discovery
Support for [OpenID Connect Discovery 1.0](https://openid.net/specs/openid-connect-discovery-1_0.html) as an auth server discovery mechanism, alongside existing OAuth 2.0 Protected Resource Metadata discovery. (PR [#797](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/797))

### Icon Metadata for Tools, Resources, Resource Templates, and Prompts
New `icons` array field on `Tool`, `Resource`, `ResourceTemplate`, and `Prompt` definitions. Icon schema: `{ src: string, mimeType: string, sizes?: string[] }`. ([SEP-973](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/973))

### Incremental Scope Consent via `WWW-Authenticate`
Incremental scope consent flow using `WWW-Authenticate` headers, allowing servers to request additional OAuth scopes after initial authorization. ([SEP-835](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/835))

### Tool Name Guidance
Guidance on tool naming: 1–128 characters, case-sensitive, restricted to `A-Z`, `a-z`, `0-9`, `_`, `-`, `.`, unique within a server. ([SEP-986](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1603))

### Updated `ElicitResult` and `EnumSchema`
[...]

Issue #2154, opened 2026-04-17
[FEATURE] Add optional security metadata to tool definitions


srbhsrkr:
## Problem Statement

The SDK's tool system has zero security classification. Every registered tool — whether it reads a file or deletes a database — is treated identically by the event loop. There is no way to declare a tool's safety profile, and no built-in permission gate before tool execution.

**What exists today:**
- `@tool` decorator supports only: `func`, `description`, `inputSchema`, `name`, `context`. No security parameters.
- `ToolSpec` TypedDict has only: `description`, `inputSchema`, `name`, `outputSchema`. No permission fields.
- `AgentTool` base class has: `tool_name`, `tool_spec`, `tool_type`, `supports_hot_reload`, `is_dynamic`. No security properties.
- `BeforeToolCallEvent` with `cancel_tool` field **enables** permission gating via hooks, but requires users to implement all classification logic externally.
- `MCPClient.ToolFilters` filters tools at *load time* by name pattern, but no runtime permission checks.

**What's missing:**
- No way for tools to self-declare: "I'm read-only", "I'm destructive", "I need confirmation".
- No mechanism for hooks to reason about tool safety without hardcoding tool-name-to-permission mappings.
- MCP tools have no runtime authorization beyond load-time name filtering.

**Why this matters:**
- Tool misuse is OWASP #5 for agentic applications. The SDK already has the hook infrastructure (`BeforeToolCallEvent` + `cancel_tool`). What's missing is the *metadata* on tools for hooks to reason about, and a reference implementation that users can adopt.
- Without metadata, every permission hook must hardcode tool-name lists. Metadata makes permission policies declarative and portable.

## Related Issues

- **#1598** — Ability to detect whether a tool is an agent vs regular tool: same pattern — wants richer `AgentTool` properties for hook-based policies (type metadata)
- **#1261** — Add tags to tools for A2A AgentCard Skill: generic tags on tools — security classification could coexist alongside tags
- **#445** — Configurable agent-as-tool properties: general tool metadata extensibility
- **#1770** — Revamp Guardrail feature: security metadata gives guardrails tool-level context to reason about

### Phase 1: Security metadata on tools (non-breaking)

Add optional fields to `@tool` decorator and `AgentTool`:

```python
@tool(read_only=True)
def list_files(directory: str) -> list[str]:
    """List files in a directory."""
    ...

@tool(destructive=True, requires_confirmation=True)
def delete_file(path: str) -> str:
[...]

PR #1898, opened 2026-03-13
fix: handle missing usage metadata on premature Anthropic stream termination

giulio-leone:
## Problem

When the Anthropic API stream terminates before sending the `message_stop` event (e.g., network timeout, connection reset, server error mid-stream), `AnthropicModel.stream()` crashes with `AttributeError`.

**Crash site**:
```python
usage = event.message.usage  # type: ignore
yield self.format_chunk({"type": "metadata", "usage": usage.model_dump()})
```

Two failure modes:
1. **Empty stream**: the `event` variable is never assigned by the `async for` loop → `UnboundLocalError`
2. **Premature termination**: the last `event` lacks a `.message` attribute or `.message.usage` is `None` → `AttributeError`

## Root Cause

The code unconditionally accesses `event.message.usage` after the async iteration loop, assuming the Anthropic stream always completes normally with usage metadata. That assumption fails when the connection drops before the final usage-bearing event arrives.

## Fix

1. Initialize `event = None` before the loop
2. Keep the safe post-loop access pattern, but **fail clearly** instead of silently synthesizing zero-usage metadata
3. Raise explicit `RuntimeError`s for:
   - empty stream: `Anthropic stream terminated before receiving any events`
   - incomplete stream: `Anthropic stream ended without usage metadata`
4. Split the stray rate-limit assertions back into their own regression test

This keeps normal successful streams unchanged, but ensures interrupted requests fail clearly instead of masking a broken connection.

## Tests

Added / updated regression coverage for:
- `test_stream_premature_termination`
- `test_stream_empty_no_events`
- `test_stream_rate_limit_error`

Local validation:
- `ruff check src/strands/models/anthropic.py tests/strands/models/test_anthropic.py`
- `ruff format --check src/strands/models/anthropic.py tests/strands/models/test_anthropic.py`
- `pytest -q tests/strands/models/test_anthropic.py` → `44 passed`
- runtime proof script covering:
  - normal completion still yielding metadata
  - premature termination raising a clear error
  - empty stream raising a clear error

Fixes #1868

PR #3528, opened 2026-07-28
feat(mcp): surface tool annotations in ToolSpec

### LINKED ISSUE #3271: [FEATURE] Surface MCP tool annotations in ToolSpec (Py + TS)
### SDK Language

Both / Not language-specific

### Problem Statement

MCP tool annotations (`readOnlyHint`, `destructiveHint`, `idempotentHint`, `openWorldHint`, `title`) are dropped by both SDKs' tool mappers: Python never reads `mcp_tool.annotations` (`strands-py/src/strands/tools/mcp/mcp_agent_tool.py:65-85`, raw escape hatch via `.mcp_tool` only) and TS never reads `toolSpec.annotations`.

### Proposed Solution

Pass annotations through into ToolSpec in both SDKs so hooks/interventions and permission layers can consume them (for example auto-approve read-only tools, require confirmation for destructive ones).

### Use Case

Annotations are precisely what an agent framework needs for permission and auto-approval policy. Related asks: #2919 (content annotations on results), #2154 (security metadata on tool definitions).

### Additional Context

Gotchas: annotations are untrusted hints per spec; SEP-1984 (comprehensive tool annotations) and SEP-1913 (trust/sensitivity annotations) are in flight, so map as an opaque pass-through rather than a typed allowlist to avoid churn. No blockers.

Parent: #1659

poshinchen:
## Description

Both SDKs drop MCP tool annotations (`readOnlyHint`, `destructiveHint`, `idempotentHint`, `openWorldHint`, `title`) when mapping server tools: Python never reads `mcp_tool.annotations` and TypeScript never reads the listing's `annotations`. Hooks and permission layers have no way to see them, so a policy like "auto-approve read-only tools, confirm destructive ones" can't be built on top of the SDK.

This PR passes annotations through into `ToolSpec` in both SDKs.

Two deliberate choices:

- **Opaque pass-through.** The annotation vocabulary is still moving (SEP-1984, SEP-1913 in flight), so the field is an untyped dict/record instead of a typed allowlist. Consumers read the keys they know, and new server-side annotations arrive without an SDK change. Per the MCP spec these are untrusted hints, and the field docs say so.
- **Not sent to model providers.** Every provider in both SDKs builds its request from explicit `ToolSpec` fields, so the new key never reaches a vendor API.

## Public API Changes

New optional `annotations` field on `ToolSpec`, populated for MCP tools that declare annotations and absent otherwise:

```python
# Python
tool.tool_spec.get("annotations")
# {'title': 'Echo Tool', 'readOnlyHint': True, 'openWorldHint': False}
```

```typescript
// TypeScript
tool.toolSpec.annotations
// { title: 'Echo Tool', readOnlyHint: true, openWorldHint: false }
```

## Related Issues

Closes #3271 (parent: #1659)

## Documentation PR

No docs change; the field is documented in the `ToolSpec` type itself.

## Type of Change

New feature

## Testing

Also verified end to end against a real stdio MCP server (FastMCP with `ToolAnnotations`) through both SDK clients: annotated tools surface the dict, unannotated tools omit the field.

## Checklist

----

### DIFF
--- a/strands-py/src/strands/types/tools.py
+++ b/strands-py/src/strands/types/tools.py
@@ -30,12 +30,18 @@ class ToolSpec(TypedDict):
         outputSchema: Optional JSON Schema defining the expected output format.
             Note: Not all model providers support this field. Providers that don't
             support it should filter it out before sending to their API.
+        annotations: Optional metadata describing tool behavior (e.g. MCP tool
+            annotations such as `readOnlyHint` or `destructiveHint`). Annotations
+            are untrusted hints from the tool provider, not guarantees; consumers
+            such as permission layers must not treat them as a security boundary.
+            This field is not sent to model provider APIs.
     """
 
     description: str
     inputSchema: JSONSchema
     name: str
     outputSchema: NotRequired[JSONSchema]
+    annotations: NotRequired[dict[str, Any]]
 
 
 class Tool(TypedDict):
--- a/strands-py/src/strands/types/tools.py
+++ b/strands-py/src/strands/types/tools.py
@@ -31,17 +31,22 @@ class ToolSpec(TypedDict):
             Note: Not all model providers support this field. Providers that don't
             support it should filter it out before sending to their API.
         annotations: Optional metadata describing tool behavior (e.g. MCP tool
-            annotations such as `readOnlyHint` or `destructiveHint`). Annotations
-            are untrusted hints from the tool provider, not guarantees; consumers
-            such as permission layers must not treat them as a security boundary.
-            This field is not sent to model provider APIs.
+            annotations such as `readOnlyHint` or `destructiveHint`). Distinct
+            from content-level annotations on tool results.
+            Annotations are untrusted hints from the tool provider, not
+            guarantees; consumers such as permission layers must not treat them
+            as a security boundary. A missing key means unknown, not False:
+            per MCP spec `destructiveHint` and `openWorldHint` default to True
+            when absent (`readOnlyHint` and `idempotentHint` default to False),
+            and this field is absent entirely for non-MCP tools. This field is
+            not sent to model provider APIs.
     """
 
     description: str
     inputSchema: JSONSchema
     name: str
     outputSchema: NotRequired[JSONSchema]
-    annotations: NotRequired[dict[str, Any]]
+    annotations: NotRequired[dict[str, object]]
 
 
 class Tool(TypedDict):
--- a/strands-py/tests/strands/tools/mcp/test_mcp_agent_tool.py
+++ b/strands-py/tests/strands/tools/mcp/test_mcp_agent_tool.py
@@ -110,6 +110,24 @@ def test_tool_spec_with_empty_annotations(mock_mcp_tool, mock_mcp_client):
     assert "annotations" not in tool_spec
 
 
+def test_tool_spec_preserves_explicit_false_hints(mock_mcp_tool, mock_mcp_client):
+    mock_mcp_tool.annotations = ToolAnnotations(readOnlyHint=False, destructiveHint=False)
+
+    agent_tool = MCPAgentTool(mock_mcp_tool, mock_mcp_client)
+    tool_spec = agent_tool.tool_spec
+
+    assert tool_spec["annotations"] == {"readOnlyHint": False, "destructiveHint": False}
+
+
+def test_tool_spec_preserves_unknown_annotation_keys(mock_mcp_tool, mock_mcp_client):
+    mock_mcp_tool.annotations = ToolAnnotations(readOnlyHint=True, futureHint="x")
+
+    agent_tool = MCPAgentTool(mock_mcp_tool, mock_mcp_client)
+    tool_spec = agent_tool.tool_spec
+
+    assert tool_spec["annotations"] == {"readOnlyHint": True, "futureHint": "x"}
+
+
 @pytest.mark.asyncio
 async def test_stream(mcp_agent_tool, mock_mcp_client, alist):
     tool_use = {"toolUseId": "test-123", "name": "test_tool", "input": {"param": "value"}}
--- a/strands-py/tests/strands/tools/mcp/test_mcp_agent_tool.py
+++ b/strands-py/tests/strands/tools/mcp/test_mcp_agent_tool.py
@@ -4,6 +4,7 @@ from unittest.mock import MagicMock
 
 import pytest
 from mcp.types import Tool as MCPTool
+from mcp.types import ToolAnnotations
 
 from strands.tools.mcp import MCPAgentTool, MCPClient
 from strands.types._events import ToolResultEvent
@@ -16,6 +17,7 @@ def mock_mcp_tool():
     mock_tool.description = "A test tool"
     mock_tool.inputSchema = {"type": "object", "properties": {}}
     mock_tool.outputSchema = None  # MCP tools can have optional outputSchema
+    mock_tool.annotations = None  # MCP tools can have optional annotations
     return mock_tool
 
 
@@ -79,6 +81,24 @@ def test_tool_spec_without_output_schema(mock_mcp_tool, mock_mcp_client):
     tool_spec = agent_tool.tool_spec
 
     assert "outputSchema" not in tool_spec
+
+
+def test_tool_spec_with_annotations(mock_mcp_tool, mock_mcp_client):
+    mock_mcp_tool.annotations = ToolAnnotations(title="Test Tool", readOnlyHint=True)
+
+    agent_tool = MCPAgentTool(mock_mcp_tool, mock_mcp_client)
+    tool_spec = agent_tool.tool_spec
+
+    assert tool_spec["annotations"] == {"title": "Test Tool", "readOnlyHint": True}
+
+
+def test_tool_spec_without_annotations(mock_mcp_tool, mock_mcp_client):
+    mock_mcp_tool.annotations = None
+
+    agent_tool = MCPAgentTool(mock_mcp_tool, mock_mcp_client)
+    tool_spec = agent_tool.tool_spec
+
+    assert "annotations" not in tool_spec
 
 
 @pytest.mark.asyncio
--- a/strands-py/tests/strands/tools/mcp/test_mcp_agent_tool.py
+++ b/strands-py/tests/strands/tools/mcp/test_mcp_agent_tool.py
@@ -101,6 +101,15 @@ def test_tool_spec_without_annotations(mock_mcp_tool, mock_mcp_client):
     assert "annotations" not in tool_spec
 
 
+def test_tool_spec_with_empty_annotations(mock_mcp_tool, mock_mcp_client):
+    mock_mcp_tool.annotations = ToolAnnotations()
+
+    agent_tool = MCPAgentTool(mock_mcp_tool, mock_mcp_client)
+    tool_spec = agent_tool.tool_spec
+
+    assert "annotations" not in tool_spec
+
+
 @pytest.mark.asyncio
 async def test_stream(mcp_agent_tool, mock_mcp_client, alist):
     tool_use = {"toolUseId": "test-123", "name": "test_tool", "input": {"param": "value"}}
--- a/strands-ts/src/mcp/client.ts
+++ b/strands-ts/src/mcp/client.ts
@@ -380,6 +380,11 @@ export class McpClient {
           description: toolSpec.description || `Tool which performs ${toolSpec.name}`,
           inputSchema: toolSpec.inputSchema as JSONSchema,
           ...(toolSpec.outputSchema !== undefined && { outputSchema: toolSpec.outputSchema as JSONSchema }),
+          // Annotations pass through opaquely: the MCP spec treats them as untrusted hints,
+          // and the annotation vocabulary is still evolving (SEP-1984, SEP-1913).
+          ...(toolSpec.annotations !== undefined && {
+            annotations: toolSpec.annotations as Record<string, JSONValue>,
+          }),
           client: this,
         })
         this._serverToolNames.set(tool, toolSpec.name)
--- a/strands-ts/src/mcp/client.ts
+++ b/strands-ts/src/mcp/client.ts
@@ -382,9 +382,11 @@ export class McpClient {
           ...(toolSpec.outputSchema !== undefined && { outputSchema: toolSpec.outputSchema as JSONSchema }),
           // Annotations pass through opaquely: the MCP spec treats them as untrusted hints,
           // and the annotation vocabulary is still evolving (SEP-1984, SEP-1913).
-          ...(toolSpec.annotations !== undefined && {
-            annotations: toolSpec.annotations as Record<string, JSONValue>,
-          }),
+          // An empty annotations object is treated the same as no annotations.
+          ...(toolSpec.annotations !== undefined &&
+            Object.keys(toolSpec.annotations).length > 0 && {
+              annotations: toolSpec.annotations as Record<string, JSONValue>,
+            }),
           client: this,
         })
         this._serverToolNames.set(tool, toolSpec.name)
--- a/strands-ts/src/mcp/client.ts
+++ b/strands-ts/src/mcp/client.ts
@@ -380,12 +380,14 @@ export class McpClient {
           description: toolSpec.description || `Tool which performs ${toolSpec.name}`,
           inputSchema: toolSpec.inputSchema as JSONSchema,
           ...(toolSpec.outputSchema !== undefined && { outputSchema: toolSpec.outputSchema as JSONSchema }),
-          // Annotations pass through opaquely: the MCP spec treats them as untrusted hints,
-          // and the annotation vocabulary is still evolving (SEP-1984, SEP-1913).
+          // Pass through only the annotation keys the MCP SDK's Zod schema recognizes
+          // (title, readOnlyHint, destructiveHint, idempotentHint, openWorldHint). The SDK strips
+          // unknown keys before this code runs, so new annotation vocabulary won't surface here
+          // until the SDK dependency updates. The MCP spec treats these as untrusted hints.
           // An empty annotations object is treated the same as no annotations.
           ...(toolSpec.annotations !== undefined &&
             Object.keys(toolSpec.annotations).length > 0 && {
-              annotations: toolSpec.annotations as Record<string, JSONValue>,
+              annotations: toolSpec.annotations,
             }),
           client: this,
         })
--- a/strands-ts/src/tools/types.ts
+++ b/strands-ts/src/tools/types.ts
@@ -34,12 +34,13 @@ export interface ToolSpec {
   outputSchema?: JSONSchema
 
   /**
-   * Metadata describing tool behavior (e.g. MCP tool annotations such as `readOnlyHint`
-   * or `destructiveHint`). Annotations are untrusted hints from the tool provider, not
-   * guarantees; consumers such as permission layers must not treat them as a security
-   * boundary. This field is not sent to model provider APIs.
+   * Untrusted tool-behavior hints (e.g. MCP `readOnlyHint`, `destructiveHint`); never a security boundary.
+   *
+   * Not sent to model provider APIs. A missing key means unknown, not `false` — per MCP spec
+   * `destructiveHint` and `openWorldHint` default to `true` when absent — and the field is absent
+   * entirely for non-MCP tools.
    */
-  annotations?: Record<string, JSONValue>
+  annotations?: Record<string, JSONValue | undefined>
 }
 
 /**
--- a/strands-ts/src/tools/types.ts
+++ b/strands-ts/src/tools/types.ts
@@ -32,6 +32,14 @@ export interface ToolSpec {
    * JSON Schema defining the expected output structure for the tool.
    */
   outputSchema?: JSONSchema
+
+  /**
+   * Metadata describing tool behavior (e.g. MCP tool annotations such as `readOnlyHint`
+   * or `destructiveHint`). Annotations are untrusted hints from the tool provider, not
+   * guarantees; consumers such as permission layers must not treat them as a security
+   * boundary. This field is not sent to model provider APIs.
+   */
+  annotations?: Record<string, JSONValue>
 }
 
 /**
--- a/strands-py/src/strands/tools/mcp/mcp_agent_tool.py
+++ b/strands-py/src/strands/tools/mcp/mcp_agent_tool.py
@@ -66,7 +66,8 @@ class MCPAgentTool(AgentTool):
         """Get the specification of the tool.
 
         This method converts the MCP tool specification to the agent framework's
-        ToolSpec format, including the input schema, description, and optional output schema.
+        ToolSpec format, including the input schema, description, optional output
+        schema, and optional tool annotations.
 
         Returns:
             ToolSpec: The tool specification in the agent framework format
@@ -81,6 +82,11 @@ class MCPAgentTool(AgentTool):
 
         if self.mcp_tool.outputSchema:
             spec["outputSchema"] = {"json": self.mcp_tool.outputSchema}
+
+        # Pass annotations through opaquely: per MCP spec they are untrusted hints,
+        # and the annotation vocabulary is still evolving (SEP-1984, SEP-1913).
+        if self.mcp_tool.annotations:
+            spec["annotations"] = self.mcp_tool.annotations.model_dump(exclude_none=True)
 
         return spec
 
--- a/strands-py/src/strands/tools/mcp/mcp_agent_tool.py
+++ b/strands-py/src/strands/tools/mcp/mcp_agent_tool.py
@@ -85,8 +85,11 @@ class MCPAgentTool(AgentTool):
 
         # Pass annotations through opaquely: per MCP spec they are untrusted hints,
         # and the annotation vocabulary is still evolving (SEP-1984, SEP-1913).
+        # An annotations object with no set fields is treated the same as no annotations.
         if self.mcp_tool.annotations:
-            spec["annotations"] = self.mcp_tool.annotations.model_dump(exclude_none=True)
+            annotations = self.mcp_tool.annotations.model_dump(exclude_none=True)
+            if annotations:
+                spec["annotations"] = annotations
 
         return spec
 
--- a/strands-ts/src/mcp/__tests__/client-annotations.test.ts
+++ b/strands-ts/src/mcp/__tests__/client-annotations.test.ts
@@ -1,0 +1,65 @@
+import { describe, it, expect } from 'vitest'
+import { InMemoryTransport } from '@modelcontextprotocol/sdk/inMemory.js'
+import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js'
+import { z } from 'zod'
+import { McpClient } from '../client.js'
+
+/**
+ * Exercises annotation pass-through over a real (in-memory) MCP transport, unlike
+ * client.test.ts, which mocks the MCP SDK Client and therefore bypasses the SDK's
+ * Zod validation of listTools responses.
+ */
+describe('McpClient annotations over a real transport', () => {
+  async function listToolsFromServer(annotations: Record<string, unknown> | undefined): Promise<McpClient> {
+    const server = new McpServer({ name: 'annotations-test-server', version: '1.0.0' })
+    server.registerTool(
+      'echo',
+      {
+        description: 'Echoes input',
+        inputSchema: { value: z.string() },
+        ...(annotations !== undefined && { annotations }),
+      },
+      async ({ value }) => ({ content: [{ type: 'text', text: value }] })
+    )
+
+    const [clientTransport, serverTransport] = InMemoryTransport.createLinkedPair()
+    await server.connect(serverTransport)
+    return new McpClient({ transport: clientTransport })
+  }
+
+  it('surfaces known annotation keys from the server', async () => {
+    const client = await listToolsFromServer({ title: 'Echo', readOnlyHint: true, openWorldHint: false })
+
+    const tools = await client.listTools()
+
+    expect(tools[0]!.toolSpec.annotations).toEqual({ title: 'Echo', readOnlyHint: true, openWorldHint: false })
+  })
+
+  it('preserves explicitly-false hints', async () => {
+    const client = await listToolsFromServer({ readOnlyHint: false, destructiveHint: false })
+
+    const tools = await client.listTools()
+
+    expect(tools[0]!.toolSpec.annotations).toEqual({ readOnlyHint: false, destructiveHint: false })
+  })
+
+  it('strips unknown annotation keys at the MCP SDK boundary', async () => {
+    // The MCP SDK's ToolAnnotationsSchema is a non-passthrough Zod object: keys outside the
+    // spec'd vocabulary are dropped during listTools validation, before McpClient sees them.
+    // If this test starts failing with futureSepKey present, the SDK now passes unknown keys
+    // through and the pass-through comment in client.ts should be updated to match Python.
+    const client = await listToolsFromServer({ readOnlyHint: true, futureSepKey: 'x' })
+
+    const tools = await client.listTools()
+
+    expect(tools[0]!.toolSpec.annotations).toEqual({ readOnlyHint: true })
+  })
+
+  it('omits annotations when the server declares none', async () => {
+    const client = await listToolsFromServer(undefined)
+
+    const tools = await client.listTools()
+
+    expect(tools[0]!.toolSpec).not.toHaveProperty('annotations')
+  })
+})
--- a/strands-ts/src/mcp/__tests__/client.test.ts
+++ b/strands-ts/src/mcp/__tests__/client.test.ts
@@ -456,6 +456,32 @@ describe('MCP Integration', () => {
       )
     })
 
+    it('surfaces tool annotations in the tool spec', async () => {
+      const annotations = { title: 'Get Weather', readOnlyHint: true }
+      sdkClientMock.listTools.mockResolvedValue({
+        tools: [{ name: 'weather', description: 'Get weather', inputSchema: {}, annotations }],
+      })
+
+      const tools = await client.listTools()
+
+      expect(tools[0]!.toolSpec).toEqual({
+        name: 'weather',
+        description: 'Get weather',
+        inputSchema: {},
+        annotations,
+      })
+    })
+
+    it('omits annotations from the tool spec when the server provides none', async () => {
+      sdkClientMock.listTools.mockResolvedValue({
+        tools: [{ name: 'weather', description: 'Get weather', inputSchema: {} }],
+      })
+
+      const tools = await client.listTools()
+
+      expect(tools[0]!.toolSpec).not.toHaveProperty('annotations')
+    })
+
     it('generates description fallback when description is missing', async () => {
       sdkClientMock.listTools.mockResolvedValue({
         tools: [{ name: 'my_tool', inputSchema: {} }],
--- a/strands-ts/src/mcp/__tests__/client.test.ts
+++ b/strands-ts/src/mcp/__tests__/client.test.ts
@@ -482,6 +482,16 @@ describe('MCP Integration', () => {
       expect(tools[0]!.toolSpec).not.toHaveProperty('annotations')
     })
 
+    it('omits annotations from the tool spec when the server provides an empty annotations object', async () => {
+      sdkClientMock.listTools.mockResolvedValue({
+        tools: [{ name: 'weather', description: 'Get weather', inputSchema: {}, annotations: {} }],
+      })
+
+      const tools = await client.listTools()
+
+      expect(tools[0]!.toolSpec).not.toHaveProperty('annotations')
+    })
+
     it('generates description fallback when description is missing', async () => {
       sdkClientMock.listTools.mockResolvedValue({
         tools: [{ name: 'my_tool', inputSchema: {} }],
--- a/strands-ts/src/tools/mcp-tool.ts
+++ b/strands-ts/src/tools/mcp-tool.ts
@@ -14,6 +14,7 @@ export interface McpToolConfig {
   description: string
   inputSchema: JSONSchema
   outputSchema?: JSONSchema
+  annotations?: Record<string, JSONValue>
   client: McpClient
 }
 
@@ -39,6 +40,7 @@ export class McpTool extends Tool {
       description: config.description,
       inputSchema: config.inputSchema,
       ...(config.outputSchema !== undefined && { outputSchema: config.outputSchema }),
+      ...(config.annotations !== undefined && { annotations: config.annotations }),
     }
     this.mcpClient = config.client
   }
--- a/strands-ts/src/tools/mcp-tool.ts
+++ b/strands-ts/src/tools/mcp-tool.ts
@@ -14,7 +14,7 @@ export interface McpToolConfig {
   description: string
   inputSchema: JSONSchema
   outputSchema?: JSONSchema
-  annotations?: Record<string, JSONValue>
+  annotations?: Record<string, JSONValue | undefined>
   client: McpClient
 }
 


poshinchen:
@strandly-the-agent review this PR, also show me API changes breakdown. I want like public API surface (new params, methods, defaults, etc.), failure modes, errors, naming (and alternatives), and whether it's consistent with similar components within the SDK.

strands-py/src/strands/types/tools.py
```diff
@@ -30,12 +30,18 @@ class ToolSpec(TypedDict):
         outputSchema: Optional JSON Schema defining the expected output format.
             Note: Not all model providers support this field. Providers that don't
             support it should filter it out before sending to their API.
+        annotations: Optional metadata describing tool behavior (e.g. MCP tool
+            annotations such as `readOnlyHint` or `destructiveHint`). Annotations
+            are untrusted hints from the tool provider, not guarantees; consumers
+            such as permission layers must not treat them as a security boundary.
+            This field is not sent to model provider APIs.
```
strandly-the-agent: 🟡 **Absence is the fail-open case here, and the docstring doesn't say so.** Straight out of `mcp==1.29.0`: `destructiveHint` is *"Default: true"* and `openWorldHint` is *"Default: true"*, while `readOnlyHint`/`idempotentHint` default to false. Since `exclude_none=True` only emits what the server explicitly set, the natural policy — `if annotations.get("destructiveHint"): confirm()` — auto-approves whenever a server omits the hint, which is the common case. I reproduced that against the real `BeforeToolCallEvent` API in both SDKs: the tool ran in 4 of 6 realistic server configs where the spec says treat-as-destructive, including one where the server *explicitly* sent `readOnlyHint: false`.

The existing "untrusted hints / not a security boundary" line covers a **lying** server; this is an **honest** server plus a lossy representation of non-false defaults. Two more facts a policy author needs: this key is absent for *every* non-MCP tool (so an annotation-keyed policy silently no-ops for local `@tool` functions), and MCP documents `destructiveHint`/`idempotentHint` as meaningful only when `readOnlyHint == false`. Also folds in a disambiguation against #2919, which proposes a different `annotations` on `ToolResultContent` ~35 lines below.

```suggestion
        annotations: Optional metadata describing tool behavior (e.g. MCP tool
            annotations such as `readOnlyHint` or `destructiveHint`). Distinct
            from content-level annotations on tool results.
            Annotations are untrusted hints from the tool provider, not
            guarantees; consumers such as permission layers must not treat them
            as a security boundary. A missing key means unknown, not False:
            per MCP spec `destructiveHint` and `openWorldHint` default to True
            when absent (`readOnlyHint` and `idempotentHint` default to False),
            and this field is absent entirely for non-MCP tools. This field is
            not sent to model provider APIs.
```

strands-py/src/strands/types/tools.py
```diff
@@ -30,12 +30,18 @@ class ToolSpec(TypedDict):
...
     """
 
     description: str
     inputSchema: JSONSchema
     name: str
     outputSchema: NotRequired[JSONSchema]
+    annotations: NotRequired[dict[str, Any]]
```
strandly-the-agent: ⚪ Nit, one word: `dict[str, object]` instead of `dict[str, Any]`. `Any` doesn't just lose precision, it actively lies — `spec.get("annotations", {})["readOnlyHnit"]` (note the typo) type-checks clean *and* satisfies an annotated `bool`, because mypy fabricates the type. `object` keeps the field just as open, forces the narrowing that the TS side already gets for free from `Record<string, JSONValue>` + `noUncheckedIndexedAccess`, and restores Py/TS strictness parity — relevant since `strands-ts/AGENTS.md:100` forbids `any` outright.

```suggestion
    annotations: NotRequired[dict[str, object]]
```

strandly-the-agent (commented):
**Approve with conditions.** The design is right and the empty-annotations fix landed cleanly. Two things worth fixing before merge — both are about what the field *promises* rather than what it does: (1) absent hints read as `false`, but MCP defaults `destructiveHint`/`openWorldHint` to **true**, so the PR's own motivating policy ("auto-approve read-only, confirm destructive") fails **open**; (2) the TS "opaque pass-through" comment is factually wrong — `@modelcontextprotocol/sdk`'s Zod schema strips unknown keys *before* `client.ts` runs, so TS gets none of the SEP-1984/1913 forward-compat the comment cites as the reason for staying untyped. Full API-changes breakdown below, as asked.

<details>
<summary>✅ What I actually verified (branch <code>pr3528</code> @ <code>5501f2b</code>, base <code>b3643be</code>)</summary>

- ✅ `pytest tests/strands/tools/mcp/test_mcp_agent_tool.py` → **14 passed**; `vitest run src/mcp/__tests__/client.test.ts` → **95 passed**; `tsc --noEmit` clean; `mypy` clean (the 2 `JSONSchema` type-arg warnings are pre-existing on the base commit).
- ✅ **MCP defaults, read straight out of `mcp==1.29.0`:** `readOnlyHint` *Default: false*, `destructiveHint` **Default: true**, `idempotentHint` *Default: false*, `openWorldHint` **Default: true**. Also documented there: `destructiveHint`/`idempotentHint` are "meaningful only when `readOnlyHint == false`".
- ✅ **Zod strip reproduced from source** (`node_modules/@modelcontextprotocol/sdk/dist/cjs/types.js:1211` is a plain `z.object`): `{title:'T',readOnlyHint:true,futureSepKey:'x'}` → `{"title":"T","readOnlyHint":true}`; `{futureSepKey:'x'}` → `{}` (which then trips your `Object.keys(...).length > 0` guard, so the `annotations` key vanishes **entirely**); `annotations: null` → `ZodError`. Python is unaffected — `ToolAnnotations.model_config = ConfigDict(extra="allow")`, unknown keys survive `model_dump`.
- ✅ **Your "never reaches a vendor API" claim holds.** Falsified empirically, not just read: canary annotations pushed through the real MCP stack into formatted requests for TS Bedrock/Anthropic/OpenAI-chat/OpenAI-responses and Py Bedrock/OpenAI — no annotation keys in any payload. All 13 Py + 5 TS providers project `name`/`description`/`inputSchema` explicitly; no `**tool_spec` spread anywhere.
- ✅ **Fail-open reproduced end to end**, both SDKs: the naive advertised policy on `BeforeToolCallEvent`, real agent invocations — the tool **ran in 4 of 6** realistic server configs where the spec says treat-as-destructive. Sharpest case: server explicitly sends `readOnlyHint: false` (which per spec makes `destructiveHint` meaningful and defaulting to `true`) and it still auto-approves.
- ✅ **Mutation-tested every production hunk** — all load-bearing. But a "keep only truthy hints" mutant passes **14/14 Py and 95/95 TS** (see the test comment).
- ⚪ Ruled out, so you don't have to wonder: `Object.keys(null)` crash (Zod rejects `null` first — mock-only), prototype pollution (Zod strips `__proto__`; nothing spreads annotations), ToolSpec key shadowing (annotations nest a level down), and the TS OTel `gen_ai.tool.definitions` "leak" (`Agent` never populates `toolsConfig` — dead wiring, pre-existing).

</details>

<details>
<summary>📐 <b>API changes breakdown</b> — surface, defaults, failure modes, naming, consistency</summary>

### 1. Public surface delta

No new classes, methods, exports, errors, or defaults. Five additions:

| SDK | Symbol | Type | Optionality | Absence semantics |
|---|---|---|---|---|
| Py | `ToolSpec.annotations` `types/tools.py:44` | `dict[str, Any]` | `NotRequired` (dict stays `total=True`) | key **omitted** |
| Py | `MCPAgentTool.tool_spec` `mcp_agent_tool.py:86-92` | populates | — | omitted if `None` **or** empty dump |
| TS | `ToolSpec.annotations` `tools/types.ts:42` | `Record<string, JSONValue>` | `?` | key **omitted**; barrel-exported |
| TS | `McpToolConfig.annotations` `tools/mcp-tool.ts:17` | `Record<string, JSONValue>` | `?` | consumed at `:43` |
| TS | `McpClient.listTools` `mcp/client.ts:336-342` | populates | — | omitted if `undefined` **or** `{}` |

**Backward compatibility — non-breaking on all three axes, verified not assumed:**

| Axis | Python | TypeScript |
|---|---|---|
| consumers *reading* a ToolSpec | ✅ | ✅ |
| implementors *constructing* one | ✅ pre-PR literal still checks under `mypy --strict` | ✅ `tsc` exit 0 |
| subclassers (`AgentTool` / `Tool` / `McpToolConfig`) | ✅ no abstract member added (`types/tools.py:241` unchanged) | ✅ optional field only |

Type-system details worth knowing: `exactOptionalPropertyTypes: true` (`tsconfig.base.json:21`) **rejects** `{annotations: undefined}` — so the conditional spread at `mcp-tool.ts:43` is required, not stylistic, and consumers building a `ToolSpec` conditionally must copy that pattern. `noUncheckedIndexedAccess: true` forces narrowing on `annotations?.['readOnlyHint']`. On the Python side `ToolSpec` is a **closed** TypedDict (an unknown key is rejected), but `spec["annotations"]` raises **no** mypy error despite being a latent `KeyError` — same footgun as `outputSchema`, only much likelier to be hit because this field is absent for *every* non-MCP tool.

### 2. Defaults and absence semantics — the sharp edge

There is no SDK-level default; absence is the default. But absence is **three-way ambiguous** and MCP's own per-field defaults are **asymmetric**:

| What the consumer sees | Could mean | Spec default |
|---|---|---|
| no `annotations` key at all | non-MCP tool (`@tool`, agent-as-tool, structured-output) · MCP server sent nothing · MCP server sent `{}` | — |
| key present, hint missing | server didn't declare it | `readOnlyHint` **false**, `destructiveHint` **true**, `idempotentHint` **false**, `openWorldHint` **true** |
| hint `false` | server affirmatively declared it false | — |

So `if annotations.get("destructiveHint")` is backwards for the two hints that matter most to a destructive-action gate, and a policy keyed on annotations **silently no-ops for every local `@tool`** (`decorator.py:310`, `_agent_as_tool.py:112`, `function-tool.ts:160-164`, `agent-as-tool.ts:154-167`). That's the docstring fix in the inline comment.

### 3. Failure modes and errors

| Mode | Today | Assessment |
|---|---|---|
| absent ≠ false | omitted | 🟡 needs one docstring sentence (inline) |
| `destructiveHint` read without `readOnlyHint` | undocumented | 🟡 spec says it's "meaningful only when `readOnlyHint == false`" |
| absent for all non-MCP tools | omitted | 🟡 needs an explicit "absent means *unknown*, not safe" rule |
| server lies | docstring warns | ✅ right call |
| wrong-typed value (`"true"` for a bool) | passed through unvalidated | ✅ **correct to stay silent** — validating an untyped bag re-introduces exactly the vocabulary coupling the design avoids. Note the two client libraries differ upstream here (pydantic lax-coerces, Zod throws and rejects the whole `listTools()`) — pre-existing, not yours |
| malformed annotation from one tool | whole `listTools()` rejects in TS | ⚪ upstream MCP SDK behaviour, identical pre-PR |

**No new exceptions, and no new warnings, is the right error model** — a log line per tool per listing would be pure noise. This is a docs fix, not a code fix.

### 4. Naming — keep `annotations`

Right call: MCP-spec-faithful, and MCP is the only producer. Alternatives weighed and rejected:

| Alternative | Why not |
|---|---|
| `toolAnnotations` | stutters on `ToolSpec`; loses wire fidelity |
| `hints` / `behaviorHints` | semantically *more* honest (they are hints) but diverges from MCP for no gain, and `title` isn't a hint |
| `metadata` | too vague; becomes a dumping ground |
| nested `mcp: { annotations }` | over-namespaces; forecloses a non-MCP source reusing the same vocabulary |

One concrete caveat: **#2919 proposes `annotations` on `ToolResultContent` — same file, ~35 lines away (`types/tools.py:76`), completely different vocabulary** (`audience`/`priority`/`lastModified`). Worth one disambiguating clause in the docstring now, before two identically-named unrelated fields land side by side.

### 5. Typed vs opaque

Opaque is defensible and I'd ship it — the SEP churn argument is real, and **migrating to typed later is non-breaking** (verified: narrowing to `McpToolAnnotations & Record<string, JSONValue>` compiles, and so does a pure typed shape). What opaque costs you, measured: no autocomplete, no wrong-type detection, and `dict[str, Any]` will happily type-check `spec.get("annotations", {})["readOnlyHnit"]` — a **typo** — as a `bool`. Hence the two one-word type nits in the inline comments; both keep the design and just stop the type from lying.

### 6. Consistency with similar components

Matches the `outputSchema` precedent closely — Py #818 (`6ea8f723`), TS #3109 (`b07fadb0`) — same file footprint, same `NotRequired`/`?`, same conditional spread, same "providers don't get this" caveat. Two honest divergences: `outputSchema` is **typed** (`JSONSchema`) while this is untyped, and `outputSchema` at least feeds Python telemetry (`tracer.py:823`) while `annotations` feeds nothing yet. Consumer ergonomics are four nullable hops (`selected_tool` → `tool_spec` → `.get("annotations")` → `.get(key)` with per-key defaults). `AgentTool` already carries convenience accessors (`supports_hot_reload`, `is_dynamic`, `get_display_properties`), so `is_read_only`-style helpers would fit the pattern — but **not in this PR**; add them when the first real consumer lands (and note #2154 proposes exactly that, see questions).

### 7. API-review classification

**Moderate**, not minimal — `API_BAR_RAISING.md:20-21` scopes "minimal" to *"a new parameter to an infrequently used method"*, and `ToolSpec` is a core barrel-exported type in both SDKs; this is a one-way door on a public data contract. Not substantial either (no new primitive/abstraction). → `api/needs-review`, async sign-off, **no API meeting needed**.

</details>

<details>
<summary>🧪 Test gaps (3, all mutation-verified as currently uncaught)</summary>

1. **Explicit `false` hints, both SDKs** — every test uses `readOnlyHint: true`. A plausible "keep only truthy hints" refactor passes 14/14 Py and 95/95 TS. An explicit `destructiveHint: false` means "confirmed non-destructive", which is the *opposite* of absent — exactly what a permission layer needs. (Suggested test inline.)
2. **Real-transport TS test** — `client.test.ts:46` has a file-level `vi.mock('@modelcontextprotocol/sdk/client/index.js')`, so all three new tests talk to a fake `Client` that echoes back whatever you feed it, and structurally cannot see the Zod strip. `InMemoryTransport.createLinkedPair()` is already a dependency (~23ms, no subprocess/network). Must be a new file — `vi.mock` is per-module-graph. Precedent for the Python side: `tests_integ/mcp/test_mcp_output_schema.py` does exactly this for `outputSchema` via `echo_server.py`, in 1.3s with no LLM call.
3. **The "never sent to a provider" invariant has no regression test** — `rg annotations strands-py/tests/strands/models/ strands-ts/src/models/__tests__/` → zero hits. It holds today purely by the explicit-allowlist convention across 18 provider files; one future provider PR using `**tool_spec` would silently put a server-controlled `title` into the prompt with no CI signal. You assert this in a docstring — worth one guard test.

</details>

<details>
<summary>⚪ Nits (bundled, none merge-blocking)</summary>

- The new TS TSDoc is 4 dense lines covering four claims, but `strands-ts/AGENTS.md:151` says interface properties **must** have single-line descriptions, and every neighbour in `tools/types.ts` is 1–2 lines. Shorter wording suggested inline.
- Pinning **open** SEP PR numbers (`SEP-1984`, `SEP-1913`) in a permanent comment is fragile — both verified real and correctly named, but they can be renumbered/closed/superseded and the comment goes stale with nobody touching the file. Consider linking the spec's annotations section instead.
- Measured, low priority: `title` is the one unbounded, unsanitized field that survives validation, and both SDKs feed the whole spec through a `JSON.stringify/2` token heuristic. A 400 KB `title` pushes `projectedInputTokens` past a 200k window — with `proactiveCompression` on, real conversation history gets dropped; with `contextManager: 'auto'`, it burns a wasted summarization call. Honest caveat: `description`/`inputSchema` are already unbounded in that same estimate, so this **widens** a pre-existing surface rather than creating one, and it needs proactive compression enabled. Flagging, not blocking.

</details>

### Questions

**Blocking (process only):** should this carry `api/needs-review` → `api/review-complete` before merge, per the moderate classification above? CI gates on it and the PR has neither label.

**Non-blocking:**
- The MCP spec says clients "should never make tool use decisions based on ToolAnnotations received from untrusted servers", which cuts against the auto-approve framing in the PR body. Would it be worth stating the safe rule outright — annotations may **escalate** friction, never **reduce** it?
- Neither SDK's reference policy layer reads this field (`hitl` and `cedar` both gate on tool *name*; repo-wide grep for the hint vocabulary → zero matches). Is "plumbing now, consumer later" the intent? If so, worth saying so in the description and linking a tracker — **#2154** proposes mapping MCP annotations into a generic `AgentTool.metadata` with `is_read_only`/`is_destructive` accessors, so whoever implements it should know to derive from `tool_spec["annotations"]` rather than re-parse `mcp_tool.annotations`.
- `McpToolConfig` (`tools/mcp-tool.ts:12`) is exported but not in the barrel — intentionally public, or should it be `@internal`?
- Since the TS Zod strip undercuts the exact "gotcha" #3271 called out, would you rather keep #3271 open or file a fast-follow for the TS forward-compat side?

---
<sub>🤖 Automated review by `strandly-the-agent` (7 passes: correctness · API bar-raise · adversarial · test-quality · LLM-context · issue-alignment · docs-accuracy; ~35 raw findings → 5 posted, 24 suppressed). I'm experimental and I can be wrong — please treat this as work for a human to approve, and push back where I've missed context.</sub>

strands-ts/src/tools/types.ts
```diff
@@ -32,6 +32,14 @@ export interface ToolSpec {
...
    */
   outputSchema?: JSONSchema
+
+  /**
+   * Metadata describing tool behavior (e.g. MCP tool annotations such as `readOnlyHint`
+   * or `destructiveHint`). Annotations are untrusted hints from the tool provider, not
+   * guarantees; consumers such as permission layers must not treat them as a security
+   * boundary. This field is not sent to model provider APIs.
+   */
+  annotations?: Record<string, JSONValue>
```
strandly-the-agent: 🟡 Two things bundled here.

**Type:** `Record<string, JSONValue>` rejects any `interface`-typed value (`TS2322: Index signature for type 'string' is missing`) — including the MCP SDK's own annotations type, whose `title?: string` isn't assignable to `JSONValue`. That's why the `as Record<string, JSONValue>` cast at `client.ts:341` is load-bearing, and it means a consumer following this repo's own `interface`-for-object-shapes rule (`strands-ts/AGENTS.md:103`) hits a wall building one of these. `Record<string, JSONValue | undefined>` accepts optional-property interfaces and lets the cast go away.

**Doc length:** `strands-ts/AGENTS.md:151` says interface properties *must* have single-line descriptions, and `name`/`inputSchema`/`outputSchema` above are all 1–2 lines. Suggested text keeps every claim (untrusted · not a security boundary · not sent to providers · absent ≠ false) but reads as a field description rather than a design note — mirroring the Python docstring change so the two stay in parity.

```suggestion
  /**
   * Untrusted tool-behavior hints (e.g. MCP `readOnlyHint`, `destructiveHint`); never a security boundary.
   *
   * Not sent to model provider APIs. A missing key means unknown, not `false` — per MCP spec
   * `destructiveHint` and `openWorldHint` default to `true` when absent — and the field is absent
   * entirely for non-MCP tools.
   */
  annotations?: Record<string, JSONValue | undefined>
```

strands-ts/src/mcp/client.ts
```diff
@@ -333,6 +333,13 @@ export class McpClient {
               description: toolSpec.description || `Tool which performs ${toolSpec.name}`,
               inputSchema: toolSpec.inputSchema as JSONSchema,
               ...(toolSpec.outputSchema !== undefined && { outputSchema: toolSpec.outputSchema as JSONSchema }),
+              // Annotations pass through opaquely: per MCP spec they are untrusted hints,
+              // and the annotation vocabulary is still evolving (SEP-1984, SEP-1913).
+              // An empty annotations object is treated the same as no annotations.
```
strandly-the-agent: 🟡 **This comment is wrong about its own behaviour — annotations don't pass through opaquely on the TS side.** `@modelcontextprotocol/sdk`'s `ToolAnnotationsSchema` is a plain `z.object` (`dist/cjs/types.js:1211`), so Zod **strips** unrecognized keys inside `this._client.listTools()` at `:326`, before this code ever sees the object. Reproduced from source:

```
{title:'T', readOnlyHint:true, futureSepKey:'x'}  ->  {"title":"T","readOnlyHint":true}
{futureSepKey:'x'}                                ->  {}   // then your length>0 guard drops `annotations` entirely
```

So a future SEP-1984/1913 key — the exact scenario the comment cites as the reason for staying untyped — won't surface in TS until the `@modelcontextprotocol/sdk` dependency itself updates. Python genuinely is opaque (`ToolAnnotations.model_config = ConfigDict(extra="allow")`, unknown keys survive `model_dump`), so the rationale holds there and only there. The untyped design is still the right call, but the justification should be accurate — otherwise the next person reads this and assumes TS forward-compat they don't have.

Worth noting the current tests can't catch this: `client.test.ts:46` mocks the whole SDK `Client`, so `{readOnlyHint:true, evilFlag:'pwn'}` "passes through" in the mock and gets stripped in production. One real-transport test via `InMemoryTransport.createLinkedPair()` (already a dependency, ~23ms, no subprocess or network) would pin it — in a new file, since `vi.mock` is per-module-graph.

```suggestion
              // Pass through only the annotation keys the MCP SDK's Zod schema recognizes
              // (title, readOnlyHint, destructiveHint, idempotentHint, openWorldHint) — it strips
              // unknown keys before this code runs, so new annotation vocabulary won't surface
              // here until the SDK dependency updates. Per MCP spec these are untrusted hints.
              // An empty annotations object is treated the same as no annotations.
```

strands-py/tests/strands/tools/mcp/test_mcp_agent_tool.py
```diff
@@ -81,6 +83,33 @@ def test_tool_spec_without_output_schema(mock_mcp_tool, mock_mcp_client):
...
+    assert "annotations" not in tool_spec
+
+
+def test_tool_spec_with_empty_annotations(mock_mcp_tool, mock_mcp_client):
+    mock_mcp_tool.annotations = ToolAnnotations()
+
+    agent_tool = MCPAgentTool(mock_mcp_tool, mock_mcp_client)
+    tool_spec = agent_tool.tool_spec
+
+    assert "annotations" not in tool_spec
```
strandly-the-agent, line 110: 🟡 **Gap: no test uses an explicitly-`false` hint, and that's the case a permission layer cares about most.** I mutation-tested a plausible future refactor — filtering falsy values out of the dump, `{k: v for k, v in ... if v}` — and it passes **14/14** here and **95/95** on the TS side. An explicit `destructiveHint: False` means "server confirms non-destructive", which is semantically the opposite of absent, so silently dropping it would invert a policy decision with no test failure.

Second suggested test asserts the unknown-key pass-through that is the whole point of the opaque design — `ToolAnnotations` sets `extra="allow"`, so Python really does preserve future vocabulary (unlike TS, see `client.ts`). Nothing currently proves it.

```suggestion
    assert "annotations" not in tool_spec


def test_tool_spec_preserves_explicit_false_hints(mock_mcp_tool, mock_mcp_client):
    mock_mcp_tool.annotations = ToolAnnotations(readOnlyHint=False, destructiveHint=False)

    agent_tool = MCPAgentTool(mock_mcp_tool, mock_mcp_client)
    tool_spec = agent_tool.tool_spec

    assert tool_spec["annotations"] == {"readOnlyHint": False, "destructiveHint": False}


def test_tool_spec_preserves_unknown_annotation_keys(mock_mcp_tool, mock_mcp_client):
    mock_mcp_tool.annotations = ToolAnnotations(readOnlyHint=True, futureHint="x")

    agent_tool = MCPAgentTool(mock_mcp_tool, mock_mcp_client)
    tool_spec = agent_tool.tool_spec

    assert tool_spec["annotations"] == {"readOnlyHint": True, "futureHint": "x"}
```

mkmeral:
@strandly-the-agent  can you give me a breakdown of what type of annotations does mcp support? what would be the end shapes here (like we have objects). how could customers use it? are these annotations standardized? 

also what are the implications of this change, can i now use/overload annotations with my own tools? would `@tool(annotations["readOnly"]` etc work? 

also is there any place where we serialize/deserialize ToolSpec? could this cause any problems? 

@poshinchen jfyi, I think PR is straightforward, but I feel like it bring the whole MCP annotation ecosystem to Strands, and levels it up as if it's supported natively. Then my question (for API review) becomes is it generic, do we support for all, what annotations exist, and so on. hopefully @strandly-the-agent will answer for me lol

strandly-the-agent:
Answering at `88159c1`, against the deps this repo actually pins (`mcp==1.29.0`, `@modelcontextprotocol/sdk ^1.25.2`). Short version of the four questions, then detail.

1. **MCP has two unrelated things called "annotations."** This PR surfaces one: `ToolAnnotations` — exactly 5 fields (`title`, `readOnlyHint`, `destructiveHint`, `idempotentHint`, `openWorldHint`), all optional booleans except `title`. The other is content-level `Annotations` (`audience`/`priority`/`lastModified`) on tool *results* and resources — different type, different vocabulary, same JSON key name.
2. **`@tool(annotations={...})` does not work.** Ran it: `TypeError: tool() got an unexpected keyword argument 'annotations'`. TS is `TS2769`. MCP is the *only* producer in either SDK, and **zero** shipped Strands code reads the field.
3. **Serialization is a non-issue.** Nothing in either SDK validates `ToolSpec`'s shape at runtime — no pydantic model, no `TypeAdapter`, no Zod schema, no JSON-schema check — and sessions never persist `ToolSpec` at all. No breaking boundary exists. One genuinely lossy spot worth a one-liner (Python telemetry).
4. **On your API-review framing — you're right, and there's a fact that sharpens it.** It isn't generic today. And the PR's stated reason for keeping the field an untyped bag ("the vocabulary is still evolving") has *not* materialized: `ToolAnnotations` has been byte-identical across **four** spec revisions since 2025-03-26 (~17 months), both cited SEPs are still unmerged drafts, and one field was moved *out*. That weakens the churn argument rather than the design — see the last section.

<details>
<summary><b>1. What MCP actually supports — exact shapes</b></summary>

**(a) `Tool.annotations` → `ToolAnnotations`** — what this PR surfaces. Verified: `list(ToolAnnotations.model_fields)` → `['title','readOnlyHint','destructiveHint','idempotentHint','openWorldHint']`.

| Field | Type | Default when absent | Spec wording |
|---|---|---|---|
| `title` | `string` | — | "A human-readable title for the tool." |
| `readOnlyHint` | `boolean` | **`false`** | "If true, the tool does not modify its environment." |
| `destructiveHint` | `boolean` | **`true`** | "...may perform destructive updates... *meaningful only when `readOnlyHint == false`*" |
| `idempotentHint` | `boolean` | **`false`** | "...calling repeatedly with the same arguments will have no additional effect... *meaningful only when `readOnlyHint == false`*" |
| `openWorldHint` | `boolean` | **`true`** | "...may interact with an 'open world' of external entities." |

On the wire in a `tools/list` response:
```json
{ "name": "delete_file", "description": "Deletes a file.",
  "inputSchema": { "type": "object", "properties": { "path": { "type": "string" } } },
  "annotations": { "title": "Delete File", "readOnlyHint": false,
                   "destructiveHint": true, "idempotentHint": true, "openWorldHint": false } }
```
What lands in `ToolSpec` — the same object, minus unset keys, and the key omitted entirely if nothing is set:
```python
tool.tool_spec["annotations"]
# {'title': 'Delete File', 'readOnlyHint': False, 'destructiveHint': True,
#  'idempotentHint': True, 'openWorldHint': False}
```

**(b) content-level `Annotations`** — a *different* type, on `TextContent`/`ImageContent`/`AudioContent`/`EmbeddedResource`/`Resource`/`ResourceTemplate`/`ResourceLink`: `audience: ("user"|"assistant")[]`, `priority: number` (0.0–1.0), `lastModified: string` (ISO 8601). Worth knowing: `lastModified` is in the TS SDK's Zod schema but **not** an explicit field on the Python SDK's `Annotations` (verified: only `['audience','priority']`) — it survives only as an untyped extra. This is issue #2919's territory, and it's why the docstring now says "Distinct from content-level annotations on tool results." Good call adding that.

**⚠️ `annotations` is 1 of 9 fields MCP puts on `Tool`.** Verified: `list(Tool.model_fields)` → `['name','title','description','inputSchema','outputSchema','icons','annotations','meta','execution']`. Strands surfaces `name`/`description`/`inputSchema`/`outputSchema` and now `annotations`; it drops **`title`** (top-level display name, spec 2025-06-18 — note the spec's precedence rule is `annotations.title` → `title` → `name`), **`icons`** (spec 2025-11-25, SEP-973), **`_meta`** (the spec's *sanctioned* namespaced extension point, with reserved-prefix rules), and **`execution`** (where `taskHint` was relocated). That's the honest answer to "does this bring the whole MCP annotation ecosystem in": it brings one of five sibling metadata mechanisms, and the siblings are arguably more load-bearing for UX (`title`, `icons`) and for extension (`_meta`).

</details>

<details>
<summary><b>2. Standardized? Yes — and more stable than the PR assumes</b></summary>

**`ToolAnnotations` is a first-class spec type, and its field set has not changed in 17 months:**

| Spec revision | State |
|---|---|
| 2024-11-05 | doesn't exist — `Tool` has no `annotations` |
| **2025-03-26** | **introduced** (PR #185) with all 5 fields |
| 2025-06-18 | unchanged |
| 2025-11-25 | unchanged |
| **2026-07-28** (current) | unchanged |
| `draft` (unreleased) | unchanged — diffing `schema/2026-07-28/schema.json` against `schema/draft/schema.json` shows zero difference in `ToolAnnotations` |

**The two SEPs cited in the code comment, checked live today:** SEP-1984 ("Comprehensive Tool Annotations") and SEP-1913 ("Trust and Sensitivity Annotations") are both still **open, `draft`-labeled, and sponsor-stalled** — the recent thread activity on each is an automation bot nagging about inactivity. Nothing they propose appears in any shipped revision, including `draft`. For reference, if they ever land: 1984 adds 6 more booleans (`aiProcessingHint`, `slowExecutionHint`, `resourceIntensiveHint`, `sensitiveDataHint`, `privilegedAccessHint`, `reversibleHint`); 1913 adds `sensitiveHint` (a 3-level enum `low`/`medium`/`high`, *not* a boolean — so it would break a bool-shaped assumption), `privateHint`, `maliciousActivityHint`, `attribution`, plus propagation rules. Three narrower predecessors (`trustedHint` #1487, `secretHint` #1560, `unsafeOutputHint` #1561) were all **closed**. And `taskHint` was moved *out* of `ToolAnnotations` into `Tool.execution` (PR #1854) — precedent that the maintainers push things out of the hints bag rather than pile them in.

**Is the key vocabulary validated anywhere? No.** `ToolAnnotations` has no `additionalProperties: false` in any revision, so a server may legally send arbitrary keys. But whether they *survive* is SDK-dependent and undefined: Python (`extra="allow"`) preserves them, TS (plain `z.object`, strip semantics) silently drops them — already correctly documented in `client.ts` after your last commit. Also relevant to this PR's design: spec issue **#1898 is open**, arguing `ToolAnnotations` *should* get `additionalProperties: false` with namespaced `_meta` as the sanctioned extension point instead. If that lands, an open-bag `dict[str, object]` is pointing the wrong way.

**Ecosystem use, briefly:** real but uneven. A coordinated late-2025 campaign added `readOnlyHint`/`destructiveHint` across ~67 open-source servers including `github/github-mcp-server` and `googleapis/genai-toolbox`; Spring AI Playground builds actual approval gating on the hints. But coverage is inconsistent even where present (an open issue on `github-mcp-server` reports a delete operation annotated `readOnlyHint: false` with no `destructiveHint: true`), and I could **not** find primary-source confirmation that VS&nbsp;Code/Copilot or Cursor read these hints for approval UX — treat claims that they do as unverified.

**Spec trust language, verbatim** — worth having in front of the API review, since it constrains what we should encourage:
> "NOTE: all properties in ToolAnnotations are **hints**. They are not guaranteed to provide a faithful description of tool behavior (including descriptive properties like `title`). Clients should never make tool use decisions based on ToolAnnotations received from untrusted servers."
>
> "For trust & safety and security, clients **MUST** consider tool annotations to be untrusted unless they come from trusted servers."

</details>

<details>
<summary><b>3. Can I overload annotations with my own tools? — the genericity table</b></summary>

`@tool` accepts exactly `description`, `inputSchema`, `name`, `context` (`decorator.py:728-742`). Adding `annotations`:
```
>>> @tool(annotations={'readOnly': True})
TypeError: tool() got an unexpected keyword argument 'annotations'
```
mypy agrees statically (`call-overload`); TS gives `TS2769: ... 'annotations' does not exist in type 'FunctionToolConfig'`.

(Aside, since it's a nice illustration: the key is `readOnlyHint`, not `readOnly`. Nothing validates annotation keys anywhere — so had this worked, `{"readOnly": True}` would be silently accepted and silently never match anything a consumer looks for. An open bag with no vocabulary means typos are undetectable by construction.)

| Producer | Can carry `annotations`? | How |
|---|---|---|
| **MCP tool** (Py `MCPAgentTool`, TS `McpTool`) | ✅ native | the only supported path |
| Py `@tool` fn | ❌ via API · 🟡 workaround | `TypeError`; but the `tool_spec` **setter** (`decorator.py:552`) does no filtering, so you can overwrite the whole dict post-construction |
| TS `tool()` / `FunctionTool` / `ZodTool` | ❌ via API · 🟡 workaround | `tsc` rejects the config key; `readonly` is compile-time only, so `(t.toolSpec as any).annotations = {...}` sticks |
| Hand-rolled `AgentTool` subclass (Py) / `Tool` subclass (TS) | ✅ end-to-end | write your own `tool_spec`/`toolSpec` — round-trips through the registry, confirmed by running it |
| Module/file tool (`TOOL_SPEC = {...}`, Py loader) | ✅ end-to-end | `loader.py:177` reads `TOOL_SPEC` with zero filtering — works today, entirely by accident |
| Raw `ToolSpec` dict → `Agent(tools=[...])` | ❌ broken | no bare-dict case in `process_tools`; a dict is `Iterable`, so it iterates the **keys** as tool paths → `ModuleNotFoundError` |
| Agent-as-tool (Py) | ❌ impossible | `tool_spec` is a computed property with no setter and no stored dict |
| Agent-as-tool (TS), structured-output tools | 🟡 private-field poke only | `_tool_spec` / stored `toolSpec` |

Type-checkers, tested: a `ToolSpec` literal with `annotations={"readOnlyHint": True, "myCompanyFoo": 123}` passes both `mypy --strict` and `tsc --strict` — closed at the `ToolSpec` level (an unknown *top-level* key is rejected), wide open inside `annotations`.

**Consumers, exhaustively grepped:** Python — 1 production hit, the write at `mcp_agent_tool.py:92`. TS — 2, both MCP (`client.ts:341`, `mcp-tool.ts:43`). Zero in hooks, interventions, the vended HITL handler, Cedar, the registry, telemetry, or any of the 18 provider files.

**Sizing, if you want it generic:** Py is one file (add the kwarg to `tool()`, thread it into the spec in the `decorator()` closure); TS is two–three (`FunctionToolConfig`, `ZodToolConfig`, `FunctionTool`'s constructor). Both small. One design note for that day: `MCPAgentTool.tool_spec` and `McpTool`'s constructor currently **overwrite** the field wholesale, which is fine while MCP is the only producer but becomes a merge-vs-overwrite question the moment it isn't.

</details>

<details>
<summary><b>4. Serialize/deserialize — full boundary audit, no breakage</b></summary>

The headline: **no runtime schema governs `ToolSpec` in either SDK.** No pydantic `BaseModel`/`TypeAdapter`, no Zod schema, no JSON-schema validation of the spec's own shape; `TypedDict` and `interface` are compile-time only, so an extra key is never rejected. `validate_tool_spec` (`registry.py:576`) only checks required keys are present. And **no session/state path persists `ToolSpec` at all** — both SDKs persist only `Message`/`ToolUse`/`ToolResult` (`grep ToolSpec session/` → no hits either side). So there's no deserialization surface to break.

| Boundary | file:line | Verdict |
|---|---|---|
| Py telemetry `gen_ai.tool.definitions` | `telemetry/tracer.py:816-825` | **LOSSY** — field-by-field rebuild whitelisting name/description/inputSchema/outputSchema; `annotations` invisible in spans |
| TS telemetry equivalent | `telemetry/tracer.ts:294` | passes the whole spec through — but **currently unreachable**: `rg toolsConfig strands-ts/src/agent/agent.ts` → **0 hits**, the Agent never populates it |
| 11 Py providers + TS Vercel/Bedrock/Anthropic/OpenAI/Google | e.g. `models/bedrock.py:298-309`, `models/vercel.ts:658-666` | LOSSY **by design** — matches the docstring; `outputSchema` is dropped the same way |
| Py bidi/realtime (nova_sonic, openai_realtime, gemini_live) | `experimental/bidi/models/*` | LOSSY by design |
| Py A2A → `AgentSkill` agent card | `multiagent/a2a/server.py:200-203` | LOSSY, harmless — `AgentSkill` has no such field |
| `normalize_tool_spec` · `get_all_tools_config` · event-loop snapshot | `tools/tools.py:152` · `registry.py:208` · `event_loop.py:537` | SAFE (`.copy()` / `deepcopy`) |
| TS per-invocation `deepCopy(toolSpecs)` | `agent/agent.ts:2064` | SAFE for real values — see the `undefined` note |
| TS `deepCopyWithValidation` (throws on `undefined`) | `types/json.ts:73` | **not on any ToolSpec path** — only `state-store.ts` — so it can't fire |
| MCP server mode | — | N/A — Strands is only ever an MCP *client*; it never re-emits ToolSpecs |

**Two things worth acting on, neither blocking:**
- **Py telemetry drops `annotations`** while the TS code path would include it. Today that's latent rather than a live divergence (TS never wires `toolsConfig`), but if you want traces to carry the hints, it's one conditional line in `_construct_tool_definitions`. If you'd rather they stay out of spans — defensible, since `title` is unbounded remote text — then it's worth saying so, because the two SDKs currently disagree in *intent*.
- **The `| undefined` in `Record<string, JSONValue | undefined>`** is a narrow footgun: an explicitly-`undefined` annotation value silently disappears through `deepCopy` (`JSON.parse(JSON.stringify(...))`), verified by running it. Unreachable via MCP — the SDK's Zod schema omits absent keys rather than setting them `undefined`, and your new `client-annotations.test.ts` pins that over a real transport (ran it: 99/99 pass). It only bites someone hand-building a `ToolSpec`. Mentioning because I suggested that type in my last review — it buys you dropping the cast, at the cost of a shape `deepCopy` doesn't round-trip.

</details>

<details>
<summary><b>5. So — for the API review</b></summary>

Taking `mkmeral`'s framing seriously, the tension isn't the code, it's the *promise*:

- **It reads as generic and is MCP-only.** The field sits on `ToolSpec` — the type every tool has — with a name that implies a general capability, but exactly one producer can populate it, `@tool(annotations=...)` raises `TypeError`, and agent-as-tool can't carry it at all. Meanwhile a `TOOL_SPEC`-in-a-file tool *can*, by accident. That's an inconsistent surface, and it's the thing most likely to generate "why doesn't this work for my tool" issues. Two coherent resolutions: (a) scope it honestly as MCP pass-through and say so in the field docs, or (b) commit to generic and add the `@tool` kwarg in the same release — it's small in both SDKs.
- **The churn premise for staying untyped is weaker than the PR states.** Five fields, unchanged across four spec revisions in 17 months, both cited SEPs stalled as drafts, one field evicted rather than added, and an open spec issue pushing toward *closing* the type with `_meta` as the extension point. Typing the five known keys plus an escape hatch would buy autocomplete and typo-safety — and per my last review, going opaque→typed later is non-breaking, so shipping opaque isn't a trap either. Genuinely a judgment call; I'd just make it on current data rather than on anticipated churn.
- **If you want one guardrail from all this**, it's the vocabulary, not the plumbing: an open bag with no key validation means `readOnly` vs `readOnlyHint` fails silently in both directions. Five typed optional keys would make that a compile error.
- **Scope question worth answering explicitly:** `annotations` is one of five metadata mechanisms MCP hangs off `Tool` (`title`, `icons`, `_meta`, `execution` are the others, all currently dropped). Deciding "annotations yes, the rest no" is fine — but it's worth deciding rather than defaulting, since `_meta` is the spec's *sanctioned* extension point and a future server may well put governance data there instead.

My prior review's implementation findings are all addressed at `88159c1`; the one open non-process item is still the missing regression test for "annotations never reach a provider API", which is fine to defer to an issue.

</details>

---
<sub>🤖 `strandly-the-agent`. Spec/SEP status checked live today (2026-08-03); field inventories run against this repo's pinned `mcp==1.29.0`. Ecosystem-adoption claims are the softest part of this — I flagged inline what I could and couldn't source primarily. I can be wrong; push back where I've missed context.</sub>

pgrayy (approved):

PR #3531, opened 2026-07-28
docs: add Background Tasks design

gautamsirdeshmukh:
## Type of Change

Bug fix
New feature
Breaking change
Documentation update
Other (please describe):

## Testing

## Checklist

----

### DIFF
--- a/team/designs/0016-background-tasks-plugin.md
+++ b/team/designs/0016-background-tasks-plugin.md
@@ -1,0 +1,363 @@
+# Background Tasks Plugin
+
+**Status**: Proposed
+
+**Date**: 2026-07-28
+
+**Issue**: N/A
+
+**Scope**: TypeScript SDK, but Python to follow.
+
+---
+
+[Problem](#problem) · [Proposal](#proposal) · [DX](#developer-experience)
+
+---
+
+## Problem
+
+When the model calls tools, the agent loop blocks until the entire batch completes. Seconds or minutes of tool latency become idle time, during which the model cannot reason, see early results, or make additional requests.
+
+### Current State
+
+`ConcurrentToolExecutor` lets calls within a round execute in parallel, but the batch still moves at the speed of its slowest tool. A fast lookup and a five-minute research task remain coupled – the lookup may finish immediately, but the model cannot use it until the research task completes.
+
+```mermaid
+gantt
+    title Current execution waits for the slowest tool
+    dateFormat HH:mm:ss
+    axisFormat %M:%S
+
+    section Model
+    Reason, call tools          :model1, 00:00:00, 10s
+    Blocked waiting for results :crit, wait, after model1, 5m
+    Resume reasoning            :model2, after wait, 10s
+
+    section Tools
+    Long-running tool           :tool1, after model1, 5m
+    Short tool                  :tool2, after model1, 30s
+    Short tool                  :tool3, after model1, 15s
+```
+
+This creates three practical limitations:
+1. **Adaptive concurrency is impossible** – the model cannot act on early results and launch follow-up work
+2. **Long-running work cannot continue between turns** – the current invocation must remain open until every tool call in that round completes
+3. **Developers must build async coordination themselves** – task tracking, recovery, concurrency limits, and asynchronous delivery all require custom infrastructure
+
+Existing community implementations demonstrate two useful patterns for asynchronous agent work. [async-agentic-tools](https://github.com/mikegc-aws/async-agentic-tools) reinvokes the Agent when work completes, while [devduck tasks](https://github.com/cagataycali/devduck/blob/main/devduck/tools/tasks.py) lets the model start and poll background jobs. 
+
+This proposal takes a complementary path: it introduces a flexible, storage-backed mechanism for dispatching compatible tools in the background and delivering their results to the model at safe agent-loop boundaries. **Reinvocation and richer back-and-forth subagent coordination will build on this foundation.**
+
+## Goals and Non-Goals
+
+**Goals:**
+
+- Model-driven background execution for compatible tool calls
+- Automatic result delivery at safe agent loop boundaries (no polling)
+- Consistent behavior across local tools, MCP tools and agents-as-tools
+
+**Non-Goals:**
+
+- Automatically reinvoking an Agent when background work completes
+- Interactive messaging with a running background task / subagent
+
+---
+
+## Proposal
+
+We introduce `BackgroundTasks`, an opt-in plugin that lets the model dispatch tool calls in the background and continue reasoning while they run.
+
+```typescript
+import { Agent } from '@strands-agents/sdk'
+import { BackgroundTasks } from '@strands-agents/sdk/vended-plugins'
+
+const agent = new Agent({
+  tools: [deepResearch, quickLookup],
+  plugins: [new BackgroundTasks()],
+})
+```
+
+With the default plugin configuration, the model gets to decide on background execution for any compatible tool call. Dispatch returns an immediate acknowledgement, and the result is delivered at the next safe agent-loop boundary.
+
+[Developer Experience](#developer-experience) covers per-tool overrides and storage usage. Alternatives Considered ([Appendix](#appendix)) compares this approach with other API designs.
+
+### How it Works
+
+At a high level, the plugin connects the Agent lifecycle to a storage-backed background task engine that manages scheduling, bounded execution, cancellation, and recovery.
+
+```mermaid
+%%{init: {"flowchart": {"diagramPadding": 4, "subGraphTitleMargin": {"top": 6, "bottom": 2}}}}%%
+flowchart LR
+    Agent["Agent"]
+    Plugin["BackgroundTasks plugin"]
+    Engine["BackgroundTaskEngine<br/>(in-memory queue)"]
+    Storage["Storage<br/>(pluggable)"]
+
+    Agent -->|"dispatch call"| Plugin
+    Plugin -->|"acknowledge"| Agent
+    Plugin -->|"deliver result"| Agent
+    Plugin -->|"submit task"| Engine
+    Engine -->|"return result"| Plugin
+    Engine -->|"write task record"| Storage
+    Storage -->|"read task record"| Engine
+
+    classDef public fill:#dbeafe,stroke:#2563eb,stroke-width:2px,color:#1e3a8a
+    class Agent,Storage public
+    style Plugin fill:#f3e8ff,stroke:#7e22ce,stroke-width:3px,color:#581c87
+    style Engine fill:#e5e7eb,stroke:#4b5563,stroke-width:2px,color:#111827
+    linkStyle default stroke:#4b5563,stroke-width:2px
+```
+
+The engine schedules and coordinates execution background work. Storage preserves task state across agent invocations (id, input, status, timestamps) and eventually the tool result/failure itself.
+
+Foreground and background calls both use the same `ToolExecutor` instance configured on the Agent.
+
+```mermaid
+flowchart LR
+    Calls["Model calls tools"] --> Mode{"Execution mode"}
+    Mode -->|background| Engine["BackgroundTaskEngine"]
+    Mode -->|foreground - execute| Executor["ToolExecutor"]
+    Engine -->|execute| Executor
+    Executor --> Result["Result"]
+
+    classDef agent fill:#dbeafe,stroke:#2563eb,stroke-width:2px,color:#1e3a8a
+    classDef background fill:#f3e8ff,stroke:#7e22ce,stroke-width:2px,color:#581c87
+    class Mode,Foreground,Executor, agent
+    class Engine,Detached background
+    linkStyle default stroke:#4b5563,stroke-width:2px
+```
+
+### Dispatching Background Tasks
+
+When a tool is eligible for background execution (see [Developer Experience](#developer-experience) for per-tool overrides), it receives an additional `_background: boolean` param in its input schema, with the following description:
+
+```
+"Run this tool call in the background. Acknowledgement is immediate; continue without waiting or polling. Strands will deliver the final result automatically at a later Agent boundary."
+```
+
+When the model dispatches a call with `_background: true`, it immediately receives a dispatch acknowledgement (tool result). The actual task executes concurrently while the model continues with foreground work, and its result is delivered at the next safe agent-loop boundary.
+
+```
+Background task dispatched.
+
+Task ID: <taskId>
+Tool: <toolName>
+Status: queued
+
+The task is running in the background. Continue without waiting or polling.
+Strands will deliver the final result automatically when it is ready.
+```
+
+### Result Delivery
+
+Completed background task results are delivered at one of two safe boundaries in the agent loop:
+
+```mermaid
+flowchart TD
+    Start([Invocation starts]) --> Before["BeforeInvocationEvent<br/>(deliver previously completed results)"]
+    Before --> Loop["Agent loop"]
+    Loop --> After["AfterInvocationEvent<br/>(check for newly completed results)"]
+    After -->|results ready| Deliver["Deliver results"]
+    Deliver --> Loop
+    After -->|no results| End([Return])
+
+    classDef terminal fill:#f6f8fa,stroke:#8c959f,color:#24292f
+    classDef hook fill:#dafbe1,stroke:#8250df,stroke-width:3px,color:#116329
+    classDef loop fill:#ddf4ff,stroke:#0969da,color:#0550ae
+    classDef delivery fill:#fff8c5,stroke:#9a6700,color:#633c01
+    class Start,End terminal
+    class Before,After hook
+    class Loop loop
+    class Deliver delivery
+```
+
+Both paths deliver the result as a synthetic assistant tool use followed by a matching user tool result. A simplified exchange begins with this assistant message:
+
+```
+{
+  "name": "strands_background_task_result",
+  "toolUseId": "<deliveryId>",
+  "input": {
+    "taskId": "<taskId>",
+    "toolName": "<toolName>",
+    "status": "completed"
+  }
+}
+```
+
+The matching user response contains the result:
+
+```
+Background task completed.
+
+Task ID: <taskId>
+Tool: <toolName>
+Status: completed
+
+The final result follows.
+
+<the actual structured tool result>
+```
+
+Results that complete during an invocation are delivered before it returns. Later results remain ready and are delivered when the Agent is invoked again. If a result cannot fit in context, the model receives its task ID so the application can retrieve it directly.
+
+### Storage and Recovery
+
+`BackgroundTasks` persists the task before returning the dispatch acknowledgement. If persistence fails, the task is not accepted and the model receives an error instead of a task ID.
+
+Task records contain enough information to reconstruct queued work and deliver its eventual result or failure, including the tool name, input, status, and output. Applications must treat durable task records as sensitive data.
+
+The default `InMemoryStorage` retains tasks only for the lifetime of the process. Restart recovery requires a durable `Storage` implementation and recreation of the Agent with the same Agent ID, registered tools, and storage namespace.
+
+Recovery depends on the persisted status:
+
+| Persisted status | Recovery behavior |
+| --- | --- |
+| `queued` | Requeued for execution |
+| `working` | Marked `failed` with `recoveryError` |
+| `paused` | Remains paused until `resume()` supplies the required interrupt responses |
+| `completed`, `failed`, or `cancelled` | Retained for management and result delivery |
+
+The plugin and engine do not automatically retry tool execution. Arbitrary tools may produce external side effects, so repeating an interrupted call could duplicate those effects.
+
+Cancellation and timeout use an `AbortSignal` and therefore require cooperative tools. A tool that ignores cancellation may continue producing external side effects, although its task result is discarded. The timeout applies only while a task is actively executing, not while it is queued or paused.
+
+Completed results remain ready until they are successfully delivered. Durable storage preserves them across restarts.
+
+Execution coordination is single-process. Durable storage provides recovery, not distributed locking or exactly-once external effects.
+
+## Developer Experience
+
+By default, the model decides whether to background any given call.
+
+```typescript
+import { Agent } from '@strands-agents/sdk'
+import { BackgroundTasks } from '@strands-agents/sdk/experimental'
+
+const agent = new Agent({
+  tools: [bash, deepResearch, renderVisual],
+  plugins: [new BackgroundTasks()],
+})
+```
+
+Tools can be pinned to always/never run in the background. Star notation covers the blanket case.
+
+| Policy | Who decides? | Tool schema | Behavior |
+|---|---|---|---|
+| `agentic` (default) | Model, per call | Adds `_background` | Backgrounds calls when selected |
+| `always` | Developer | Unchanged | Backgrounds every call |
+| `never` | Developer | Unchanged | Runs every call in the foreground |
+
+```typescript
+const agent = new Agent({
+  tools: [bash, deepResearch, renderVisual],
+  plugins: [
+    new BackgroundTasks({
+      policy: {
+        '*': 'agentic',
+        deepResearch: 'always',
+        renderVisual: 'never',
+      },
+    }),
+  ],
+})
+```
+
+A `Storage` implementation can be passed to durably persist/retrieve task records in an external store.
+
+```typescript
+const agent = new Agent({
+  tools: [bash, deepResearch, renderVisual],
+  plugins: [
+    new BackgroundTasks({
+      storage: new S3Storage({
+        bucket: 'agent-bucket',
+        prefix: 'background-tasks/',
+      }),
+    }),
+  ],
+})
+```
+
+## Consequences
+
+| Use background tasks when… | Do not use background tasks when… |
+|---|---|
+| Work is long-running or has unpredictable latency. | Work is short and scheduling overhead adds little value. |
+| The model does not depend on the result immediately. | The model needs the result for its very next task. |
+| Tasks are independent and completion order can vary. | Strict ordering or deterministic replay matters. |
+| Work may safely outlive the current invocation. | The invocation must remain open until all work completes. |
+
+---
+
+## Willingness to Implement
+
+Yes. Intended sequence:
+
+#### P0: Model-driven background dispatch and delivery
+
+The proposed design. Base mechanism letting the model make calls and keep reasoning without being blocked.
+
+#### P1: Deeper subagent mechanism
+
+This will build off of the base mechanism and involve other SDK features such as `use_agent` (dynamic dispatch) and direct agent-as-tool streaming. Goal is support persistent subagent communication, such as queueing tasks and allowing external reinvocation of the model.
+
+## FAQ
+
+### What tools cannot be executed in the background?
+
+Only framework-owned control and delivery tools are categorically excluded. Model-selected (`agentic`) backgrounding also requires a direct object input schema that the plugin can extend with the reserved `_background` field. Wildcard policies leave incompatible tools unchanged, while an explicit `agentic` policy for an incompatible schema fails during initialization.
+
+### Can background tasks dispatch background tasks?
+
+No (at least, not as part of this initial proposal). Once a tool is already running as a background task, the plugin suppresses nested background dispatch and executes subsequent tool calls inside the existing task. This keeps one owner for cancellation, timeout, recovery, capacity, and result delivery, and prevents unbounded fan-out.
+
+### Can an invocation wait for background tasks to complete?
+
+By default, `agent.invoke()` returns when the model finishes, even if background tasks are still running; their results are delivered at the next invocation. P1 will add completion-triggered reinvocation for richer subagent workflows. 
+
+As an interim option, `BackgroundTasks` could expose a bounded `waitUntilComplete` setting that keeps the invocation open until all running background tasks reach a terminal state.
+
+---
+
+## Appendix
+
+<details>
+<summary><strong>A: Alternatives Considered</strong></summary>
+
+### The proposed design (recommended)
+
+**Pros:** The model expresses the decision as a boolean on a tool call it already knows how to make – no meta-calling convention, no new tool vocabulary. Works uniformly across regular tools, agents-as-tools, and MCP tools. Results enter history as typed tool results rather than trusted user text, and delivery survives provider failures and process restarts.
+
+**Cons:** Injecting `_background` mutates tool input schemas the SDK does not own. When the model chooses (`agentic`), decision quality is model-dependent.
+
+### 1. A single `background()` meta-tool
+
+Instead of a `_background` param on each eligible tool, the plugin registers one tool the model calls to dispatch any other tool:
+
+```typescript
+background({ tool: 'deepResearch', input: { query: '...' } })
+```
+
+**Pros:** No schema mutation of existing tools – MCP and vendored tool schemas pass through untouched. One place to document the contract.
+
+**Cons:** The inner tool's arguments become an untyped payload inside another tool's input, so the provider no longer validates them against the real schema – malformed calls surface at execution instead of generation. The model must also learn a meta-calling convention rather than flipping a boolean on a call it already knows how to make, and models are measurably better at direct tool calls than at nesting one call inside another.
+
+### 2. A `backgroundTools: [...]` first-class parameter
+
+A list on the Agent itself; every call to a listed tool runs in the background:
+
+```typescript
+const agent = new Agent({
+  tools: [quickLookup, deepResearch],
+  backgroundTools: [deepResearch],
+})
+```
+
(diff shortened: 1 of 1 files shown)


team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+- Automatically reinvoking an Agent when background work completes
+- Interactive messaging with a running background task / subagent
+
+---
+
+## Proposal
+
+We introduce `BackgroundTasks`, an opt-in plugin that lets the model dispatch tool calls in the background and continue reasoning while they run.
+
+```typescript
```
arielnabavian: Why a plugin instead of a tool configuration?

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+    Storage -->|"read task record"| Engine
+
+    classDef public fill:#dbeafe,stroke:#2563eb,stroke-width:2px,color:#1e3a8a
+    class Agent,Storage public
+    style Plugin fill:#f3e8ff,stroke:#7e22ce,stroke-width:3px,color:#581c87
+    style Engine fill:#e5e7eb,stroke:#4b5563,stroke-width:2px,color:#111827
+    linkStyle default stroke:#4b5563,stroke-width:2px
+```
+
+The engine schedules and coordinates execution background work. Storage preserves task state across agent invocations (id, input, status, timestamps) and eventually the tool result/failure itself.
```
mkmeral: do we need a new separate storage? why don't we just use appState in agent? 

the only big item i see here is input, but do we need to persist that? it's in agent messages array already right?

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+Status: completed
+
+The final result follows.
+
+<the actual structured tool result>
+```
+
+Results that complete during an invocation are delivered before it returns. Later results remain ready and are delivered when the Agent is invoked again. If a result cannot fit in context, the model receives its task ID so the application can retrieve it directly.
+
+### Storage and Recovery
```
gautamsirdeshmukh: The TLDR here is that we use Storage to store/maintain task records (id, tool name, status, timestamps, etc)

This allows history of success/fail/cancelled tasks persist over sequential invocations. It also:
- streamlines internal logic for reading/writing task records
- makes it easier for the agent to learn about / retry a task that may not have succeeded/completed before
- allows developers to plug in an external store for their tasks if they so please

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+    classDef public fill:#dbeafe,stroke:#2563eb,stroke-width:2px,color:#1e3a8a
+    class Agent,Storage public
+    style Plugin fill:#f3e8ff,stroke:#7e22ce,stroke-width:3px,color:#581c87
+    style Engine fill:#e5e7eb,stroke:#4b5563,stroke-width:2px,color:#111827
+    linkStyle default stroke:#4b5563,stroke-width:2px
+```
+
+The engine schedules and coordinates execution background work. Storage preserves task state across agent invocations (id, input, status, timestamps) and eventually the tool result/failure itself.
+
+Foreground and background calls both use the same `ToolExecutor` instance configured on the Agent.
```
mkmeral: nit: let's double check tool executor behavior, and interface too. is it safe to call in parallel? does the interface support it?

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+  tools: [quickLookup, deepResearch],
+  backgroundTools: [deepResearch],
+})
+```
+
+**Pros:** Simplest possible surface. Fully deterministic – no reliance on model judgment, nothing injected into schemas.
+
+**Cons:** The developer must predict at build time which calls are slow, but latency is often input-dependent – the same `bash` call can take milliseconds or minutes. Dropping model choice drops the core unlock (model-driven concurrency), reducing the feature to the `always` policy alone. It also grows the first-class `Agent` API for an opt-in feature that plugins exist to carry.
+
+Furthermore, setting `policy: { deepResearch: 'always' }` allows developers to execute this level of control without sacrificing the model-driven `agentic` mode.
```
gautamsirdeshmukh: This also covers another alternative design route of having the developer set a `background` param on every tool config as part of the tool definition itself - can achieve the same thing via plugin config

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+
+**Scope**: TypeScript SDK, but Python to follow.
+
+---
+
+[Problem](#problem) · [Proposal](#proposal) · [DX](#developer-experience)
+
+---
+
+## Problem
```
notowen333: Praise the conciseness of the problem statement/the length of the doc 🙌🏻

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+import { Agent } from '@strands-agents/sdk'
+import { BackgroundTasks } from '@strands-agents/sdk/vended-plugins'
+
+const agent = new Agent({
+  tools: [deepResearch, quickLookup],
+  plugins: [new BackgroundTasks()],
+})
+```
+
+With the default plugin configuration, the model gets to decide on background execution for any compatible tool call. Dispatch returns an immediate acknowledgement, and the result is delivered at the next safe agent-loop boundary.
```
opieter-aws: What defines a compatible tool call?

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+    class Engine,Detached background
+    linkStyle default stroke:#4b5563,stroke-width:2px
+```
+
+### Dispatching Background Tasks
+
+When a tool is eligible for background execution (see [Developer Experience](#developer-experience) for per-tool overrides), it receives an additional `_background: boolean` param in its input schema, with the following description:
+
+```
+"Run this tool call in the background. Acknowledgement is immediate; continue without waiting or polling. Strands will deliver the final result automatically at a later Agent boundary."
```
mkmeral: > Strands will ...

nit: does the agent need to know which framework its based on?

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+Tool: <toolName>
+Status: queued
+
+The task is running in the background. Continue without waiting or polling.
+Strands will deliver the final result automatically when it is ready.
+```
+
+### Result Delivery
+
+Completed background task results are delivered at one of two safe boundaries in the agent loop:
```
mkmeral: why are we doing it based on before/after invocation? What if we have a single prompt that runs for hours? then the whole tool result will need to wait until the agent calls it quits

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+
+Results that complete during an invocation are delivered before it returns. Later results remain ready and are delivered when the Agent is invoked again. If a result cannot fit in context, the model receives its task ID so the application can retrieve it directly.
+
+### Storage and Recovery
+
+`BackgroundTasks` persists the task before returning the dispatch acknowledgement. If persistence fails, the task is not accepted and the model receives an error instead of a task ID.
+
+Task records contain enough information to reconstruct queued work and deliver its eventual result or failure, including the tool name, input, status, and output. Applications must treat durable task records as sensitive data.
+
+The default `InMemoryStorage` retains tasks only for the lifetime of the process. Restart recovery requires a durable `Storage` implementation and recreation of the Agent with the same Agent ID, registered tools, and storage namespace.
```
opieter-aws: Does restart recovery mean session restore here?

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+Recovery depends on the persisted status:
+
+| Persisted status | Recovery behavior |
+| --- | --- |
+| `queued` | Requeued for execution |
+| `working` | Marked `failed` with `recoveryError` |
+| `paused` | Remains paused until `resume()` supplies the required interrupt responses |
+| `completed`, `failed`, or `cancelled` | Retained for management and result delivery |
+
+The plugin and engine do not automatically retry tool execution. Arbitrary tools may produce external side effects, so repeating an interrupted call could duplicate those effects.
```
zastrowm: Do we allow engines to configure this?  I'm guessing they're going to want the ability to auto-retry, for instance

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+import { Agent } from '@strands-agents/sdk'
+import { BackgroundTasks } from '@strands-agents/sdk/experimental'
+
+const agent = new Agent({
+  tools: [bash, deepResearch, renderVisual],
+  plugins: [new BackgroundTasks()],
+})
+```
+
+Tools can be pinned to always/never run in the background. Star notation covers the blanket case.
```
lizradway: nit: i think it is probably better practice to map the policy to the tools rather than the other way around


```typescript
const agent = new Agent({
  tools: [bash, deepResearch, renderVisual],
  plugins: [
    new BackgroundTasks({
      policy: {
        agentic: '*'
        always: ["deep_research"]
      },
    }),
  ],
})
```

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+| Persisted status | Recovery behavior |
+| --- | --- |
+| `queued` | Requeued for execution |
+| `working` | Marked `failed` with `recoveryError` |
+| `paused` | Remains paused until `resume()` supplies the required interrupt responses |
+| `completed`, `failed`, or `cancelled` | Retained for management and result delivery |
+
+The plugin and engine do not automatically retry tool execution. Arbitrary tools may produce external side effects, so repeating an interrupted call could duplicate those effects.
+
+Cancellation and timeout use an `AbortSignal` and therefore require cooperative tools. A tool that ignores cancellation may continue producing external side effects, although its task result is discarded. The timeout applies only while a task is actively executing, not while it is queued or paused.
```
zastrowm: > cooperative tools

Do we have a list of tools that we would vend that would do this?

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+    class Before,After hook
+    class Loop loop
+    class Deliver delivery
+```
+
+Both paths deliver the result as a synthetic assistant tool use followed by a matching user tool result. A simplified exchange begins with this assistant message:
+
+```
+{
+  "name": "strands_background_task_result",
```
mkmeral: are we duplicating the tool use here? can we optimize it, for context management?

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+        '*': 'agentic',
+        deepResearch: 'always',
+        renderVisual: 'never',
+      },
+    }),
+  ],
+})
+```
+
+A `Storage` implementation can be passed to durably persist/retrieve task records in an external store.
```
opieter-aws: Why can't we use SessionManager to store this? what's the purpose beyond restoring a session?

team/designs/0016-background-tasks-plugin.md
zastrowm: > Cancellation and timeout use an `AbortSignal` and therefore require cooperative tools.

Is this true; could we just abandon the result in this case?  I don't know if we want to have everyone update their tools to be co-operative.  Would it be simpler to just allow any tool to be backgrounded and expose an abortSignal on toolContext?

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+Task ID: <taskId>
+Tool: <toolName>
+Status: completed
+
+The final result follows.
+
+<the actual structured tool result>
+```
+
+Results that complete during an invocation are delivered before it returns. Later results remain ready and are delivered when the Agent is invoked again. If a result cannot fit in context, the model receives its task ID so the application can retrieve it directly.
```
mkmeral: > Later results remain ready and are delivered when the Agent is invoked again.

why do we allow them to complete, and wait for next invocation? I'd much rather keep the agent loop going until all tasks are complete.  or maybe have a param? 

Like for Strandly, I definitely would not want this behavior. That said for sync chat agent, you might not invocation to be blocked

team/designs/0016-background-tasks-plugin.md
arielnabavian: I think it just polls once to see if background tasks are complete here.

team/designs/0016-background-tasks-plugin.md
mkmeral: > If a result cannot fit in context, the model receives its task ID so the application can retrieve it directly

What does this mean? Is it tool context offloader? Are we re-creating the same logic here?

team/designs/0016-background-tasks-plugin.md
gautamsirdeshmukh: Verified this (for our built-in executors), execution state is scoped to each call so there is no risk of foreground/background calls at the same time being blocked / causing a failure

team/designs/0016-background-tasks-plugin.md
lizradway: +1, what are the qualifications/differenciators for enabling a vended tool to run in the backgroun

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+
+Only framework-owned control and delivery tools are categorically excluded. Model-selected (`agentic`) backgrounding also requires a direct object input schema that the plugin can extend with the reserved `_background` field. Wildcard policies leave incompatible tools unchanged, while an explicit `agentic` policy for an incompatible schema fails during initialization.
+
+### Can background tasks dispatch background tasks?
+
+No (at least, not as part of this initial proposal). Once a tool is already running as a background task, the plugin suppresses nested background dispatch and executes subsequent tool calls inside the existing task. This keeps one owner for cancellation, timeout, recovery, capacity, and result delivery, and prevents unbounded fan-out.
+
+### Can an invocation wait for background tasks to complete?
+
+By default, `agent.invoke()` returns when the model finishes, even if background tasks are still running; their results are delivered at the next invocation. P1 will add completion-triggered reinvocation for richer subagent workflows.
```
zastrowm: I think this needs to be configurable.  I imagine agent operators to want a "full turn completion" before finishing

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+
+With the default plugin configuration, the model gets to decide on background execution for any compatible tool call. Dispatch returns an immediate acknowledgement, and the result is delivered at the next safe agent-loop boundary.
+
+[Developer Experience](#developer-experience) covers per-tool overrides and storage usage. Alternatives Considered ([Appendix](#appendix)) compares this approach with other API designs.
+
+### How it Works
+
+At a high level, the plugin connects the Agent lifecycle to a storage-backed background task engine that manages scheduling, bounded execution, cancellation, and recovery.
+
+```mermaid
```
notowen333: 🫶 the graphs/charts! So helpful 🟢

team/designs/0016-background-tasks-plugin.md
mkmeral: Nothing in this section tells me why we need storage. Why?

mkmeral: Can we solve "this issue" (whatever it is) without storage?

team/designs/0016-background-tasks-plugin.md
notowen333: Do we have eval story for this description (and other similar ones if they are present elsewhere in the doc)?

team/designs/0016-background-tasks-plugin.md
arielnabavian: Disregard. I see the vision with the plugin, but I think some questions we should answer:

- Do we see people using "always" over "agentic".
- Do we need fine grained control of which tools are eligible to be backgrounded 

IMO it would be a cleaner devex to default to agentic and have access control on the tools themselves.

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+})
+```
+
+With the default plugin configuration, the model gets to decide on background execution for any compatible tool call. Dispatch returns an immediate acknowledgement, and the result is delivered at the next safe agent-loop boundary.
+
+[Developer Experience](#developer-experience) covers per-tool overrides and storage usage. Alternatives Considered ([Appendix](#appendix)) compares this approach with other API designs.
+
+### How it Works
+
+At a high level, the plugin connects the Agent lifecycle to a storage-backed background task engine that manages scheduling, bounded execution, cancellation, and recovery.
```
zastrowm: > storage-backed background task

Why do we need a storage backed system for this?  Why is it not just agent state?

team/designs/0016-background-tasks-plugin.md
gautamsirdeshmukh: Compatible isn't the right word to use here. "Eligible" would be more fitting - and this refers to tools that are marked with `background: 'always'` (always) or `background: 'agentic'` (model can decide)

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+    Executor --> Result["Result"]
+
+    classDef agent fill:#dbeafe,stroke:#2563eb,stroke-width:2px,color:#1e3a8a
+    classDef background fill:#f3e8ff,stroke:#7e22ce,stroke-width:2px,color:#581c87
+    class Mode,Foreground,Executor, agent
+    class Engine,Detached background
+    linkStyle default stroke:#4b5563,stroke-width:2px
+```
+
+### Dispatching Background Tasks
```
mkmeral: How does this work with interrupts? What if agent process is closed between invocations? e.g. HITL with AgentCore where two messages can be days apart

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+
+#### P0: Model-driven background dispatch and delivery
+
+The proposed design. Base mechanism letting the model make calls and keep reasoning without being blocked.
+
+#### P1: Deeper subagent mechanism
+
+This will build off of the base mechanism and involve other SDK features such as `use_agent` (dynamic dispatch) and direct agent-as-tool streaming. Goal is support persistent subagent communication, such as queueing tasks and allowing external reinvocation of the model.
+
+## FAQ
```
zastrowm: How can agent check the status of a background task?

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+
+**Pros:** No schema mutation of existing tools – MCP and vendored tool schemas pass through untouched. One place to document the contract.
+
+**Cons:** The inner tool's arguments become an untyped payload inside another tool's input, so the provider no longer validates them against the real schema – malformed calls surface at execution instead of generation. The model must also learn a meta-calling convention rather than flipping a boolean on a call it already knows how to make, and models are measurably better at direct tool calls than at nesting one call inside another.
+
+### 2. A `backgroundTools: [...]` first-class parameter
+
+A list on the Agent itself; every call to a listed tool runs in the background:
+
+```typescript
```
arielnabavian: tool configuration flag, rather than backgroundTools?

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+    "toolName": "<toolName>",
+    "status": "completed"
+  }
+}
+```
+
+The matching user response contains the result:
+
+```
+Background task completed.
```
notowen333: Yeah this format especially I would want to eval.

Might guess wrapping the result in a delimiter and preceding with `follows:` with a colon would help

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+
+By default, the model decides whether to background any given call.
+
+```typescript
+import { Agent } from '@strands-agents/sdk'
+import { BackgroundTasks } from '@strands-agents/sdk/experimental'
+
+const agent = new Agent({
+  tools: [bash, deepResearch, renderVisual],
+  plugins: [new BackgroundTasks()],
```
mkmeral: Does this mean all my tools get this? Can I configure it?

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+---
+
+## Appendix
+
+<details>
+<summary><strong>A: Alternatives Considered</strong></summary>
+
+### The proposed design (recommended)
+
+**Pros:** The model expresses the decision as a boolean on a tool call it already knows how to make – no meta-calling convention, no new tool vocabulary. Works uniformly across regular tools, agents-as-tools, and MCP tools. Results enter history as typed tool results rather than trusted user text, and delivery survives provider failures and process restarts.
```
liramon2: Do you envision any way for the developer to exclude certain tools from backgrounding? For example, if the developer has a tool that runs quickly and never wants to add overhead from putting it in the background.

team/designs/0016-background-tasks-plugin.md
arielnabavian: +1. 

If we are checking the state of background task on before/after model invocation. We may want to finish execution of the executing turn, then interrupt after?

team/designs/0016-background-tasks-plugin.md
mkmeral: nvm it's right below. i should read before commenting :p

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+By default, `agent.invoke()` returns when the model finishes, even if background tasks are still running; their results are delivered at the next invocation. P1 will add completion-triggered reinvocation for richer subagent workflows. 
+
+As an interim option, `BackgroundTasks` could expose a bounded `waitUntilComplete` setting that keeps the invocation open until all running background tasks reach a terminal state.
+
+---
+
+## Appendix
+
+<details>
+<summary><strong>A: Alternatives Considered</strong></summary>
```
zastrowm: nit: Per [the template](https://github.com/strands-agents/harness-sdk/blob/main/team/designs/README.md) I'd like to see alternatives higher up in the doc.  Which path we're going down is often one of the more important parts to discuss

team/designs/0016-background-tasks-plugin.md
gautamsirdeshmukh: This mechanism necessitates (1) registering hook events, (2) registering a tool on the agent (cancel task) and (3) coordinating work with the task engine/queue. 

The first two are common usages for plugins and easily implementable through  the `initAgent` and `getTools` abstract plugin methods. And, for the third, a plugin allows that coordination to be managed separately from the tool executor - or worse, the agent loop itself.

Additionally - the developer control of setting `background` per tool can still be achieved via the plugin config

JackYPCOnline: if what returned is toolResult, i wonder why it fall into a plugins?

team/designs/0016-background-tasks-plugin.md
lizradway: +1, do we have any metrics we aim to measure the efficacy of "agentic" mode and background tasks in general on

team/designs/0016-background-tasks-plugin.md
mkmeral: +1 this looks better. but we can worry about it during API BR

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+
+Yes. Intended sequence:
+
+#### P0: Model-driven background dispatch and delivery
+
+The proposed design. Base mechanism letting the model make calls and keep reasoning without being blocked.
+
+#### P1: Deeper subagent mechanism
+
+This will build off of the base mechanism and involve other SDK features such as `use_agent` (dynamic dispatch) and direct agent-as-tool streaming. Goal is support persistent subagent communication, such as queueing tasks and allowing external reinvocation of the model.
```
notowen333: Do we have a sense of the challenges we might expect when taking on subagents using the proposed background task engine/broker?

team/designs/0016-background-tasks-plugin.md
lizradway: just in general, we should sync up + make sure this plays nicely with offloading

team/designs/0016-background-tasks-plugin.md
poshinchen: And `agentic: '*'` is not required right? I assume:
```
policy: {
    'deepResearch': 'always',
    'renderVisual': 'never'
}
```
would also work with defaults to 'agentic'

team/designs/0016-background-tasks-plugin.md
mkmeral: Should this be handled by tool registry or tool executor though? Or should tool executor just support async/backgroundable tasks? I'd like to see the pros/cons to that, if we have any

team/designs/0016-background-tasks-plugin.md
gautamsirdeshmukh: Ooh I do like this...

team/designs/0016-background-tasks-plugin.md
lizradway: context management should handle this, not bespoke for background tasks.


(currently i have context management running on AfterMessageAdded btw)

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+    classDef hook fill:#dafbe1,stroke:#8250df,stroke-width:3px,color:#116329
+    classDef loop fill:#ddf4ff,stroke:#0969da,color:#0550ae
+    classDef delivery fill:#fff8c5,stroke:#9a6700,color:#633c01
+    class Start,End terminal
+    class Before,After hook
+    class Loop loop
+    class Deliver delivery
+```
+
+Both paths deliver the result as a synthetic assistant tool use followed by a matching user tool result. A simplified exchange begins with this assistant message:
```
JackYPCOnline: when will we exit the loop? we always wait for background task complete? or is it configurable with some limitation?

team/designs/0016-background-tasks-plugin.md
arielnabavian: I'm not really following what would be stored here that needs persistance. If we are storing references to tasks in memory, would those be accessible on reinitialization?

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+| `queued` | Requeued for execution |
+| `working` | Marked `failed` with `recoveryError` |
+| `paused` | Remains paused until `resume()` supplies the required interrupt responses |
+| `completed`, `failed`, or `cancelled` | Retained for management and result delivery |
+
+The plugin and engine do not automatically retry tool execution. Arbitrary tools may produce external side effects, so repeating an interrupted call could duplicate those effects.
+
+Cancellation and timeout use an `AbortSignal` and therefore require cooperative tools. A tool that ignores cancellation may continue producing external side effects, although its task result is discarded. The timeout applies only while a task is actively executing, not while it is queued or paused.
+
+Completed results remain ready until they are successfully delivered. Durable storage preserves them across restarts.
```
lizradway: what is the OTEL/tracing story for this? do we want traces for tool use request, tool result, or both?

team/designs/0016-background-tasks-plugin.md
liramon2: Nvm, the devx section cover this

team/designs/0016-background-tasks-plugin.md
lizradway: worried about the non-linear message history a bit, and would be interested to hear the audibility/traceability story regarding this

team/designs/0016-background-tasks-plugin.md
gautamsirdeshmukh: Just using context offloader not recreating anything

team/designs/0016-background-tasks-plugin.md
zastrowm: We should separate out DevX from the concept, but I very much see this as being in the agent constructor instead of a plugin

team/designs/0016-background-tasks-plugin.md
gautamsirdeshmukh: See FAQ, I raised the point about making this configurable to wait until all tasks complete. I've seen this pattern before

team/designs/0016-background-tasks-plugin.md
gautamsirdeshmukh: We probably could use appState, but Storage also allows for task records to be persisted externally. Willing to scale this back to appState if we don't see enough value in this.

mkmeral: > Storage also allows for task records to be persisted externally

sure, what do we gain by doing that though?

team/designs/0016-background-tasks-plugin.md
gautamsirdeshmukh: Yeah, this has been called out a few times. See FAQ, I raised the point about making this configurable to wait until all tasks complete, aligned with that (pending devx)

team/designs/0016-background-tasks-plugin.md
lizradway: maybe we start with just app state for now and add storage as a later param.

that's our current approach for context manager so far as well

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+
+Tools can be pinned to always/never run in the background. Star notation covers the blanket case.
+
+| Policy | Who decides? | Tool schema | Behavior |
+|---|---|---|---|
+| `agentic` (default) | Model, per call | Adds `_background` | Backgrounds calls when selected |
+| `always` | Developer | Unchanged | Backgrounds every call |
+| `never` | Developer | Unchanged | Runs every call in the foreground |
+
+```typescript
```
JackYPCOnline: does deepResearch: 'always', override the first line?

team/designs/0016-background-tasks-plugin.md
liramon2: What happens if the model has nothing left to do? Does it wait at one of the boundaries?

team/designs/0016-background-tasks-plugin.md
mkmeral: +1 can an agent manage the tasks (check the status, cancel, etc.)

poshinchen: In multi-agents which share the same tool wrapped as background tasks. I assume they'll run concurrently, is there anything that we should be aware of?

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+The final result follows.
+
+<the actual structured tool result>
+```
+
+Results that complete during an invocation are delivered before it returns. Later results remain ready and are delivered when the Agent is invoked again. If a result cannot fit in context, the model receives its task ID so the application can retrieve it directly.
+
+### Storage and Recovery
+
+`BackgroundTasks` persists the task before returning the dispatch acknowledgement. If persistence fails, the task is not accepted and the model receives an error instead of a task ID.
```
JackYPCOnline: I wonder if it should be optional or not.

team/designs/0016-background-tasks-plugin.md
mkmeral: I've been writing docs for years. No one showed this much love to my docs 😂 😂

team/designs/0016-background-tasks-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+This creates three practical limitations:
+1. **Adaptive concurrency is impossible** – the model cannot act on early results and launch follow-up work
+2. **Long-running work cannot continue between turns** – the current invocation must remain open until every tool call in that round completes
+3. **Developers must build async coordination themselves** – task tracking, recovery, concurrency limits, and asynchronous delivery all require custom infrastructure
+
+Existing community implementations demonstrate two useful patterns for asynchronous agent work. [async-agentic-tools](https://github.com/mikegc-aws/async-agentic-tools) reinvokes the Agent when work completes, while [devduck tasks](https://github.com/cagataycali/devduck/blob/main/devduck/tools/tasks.py) lets the model start and poll background jobs. 
+
+This proposal takes a complementary path: it introduces a flexible, storage-backed mechanism for dispatching compatible tools in the background and delivering their results to the model at safe agent-loop boundaries. **Reinvocation and richer back-and-forth subagent coordination will build on this foundation.**
+
+## Goals and Non-Goals
```
mkmeral: Let's also check out how this plays with MCP's tasks, especially in new stateless world.

So technically we can kickoff an MCP task, stop agent invocation, and come back later? 🤔

Issue #3190, opened 2026-07-10
[BUG] ai-sdk-ollama - Missing modelMessageStartEvent and Incorrect finish_reason Break Strands Agent Loop


hsaenzG:
### Checks

- [x] I have updated to the lastest minor and patch version of Strands
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched [./issues](./issues?q=) and there are no duplicates of my issue

### SDK Language

TypeScript

### Strands Version

1.4.0

### Language Runtime Version

Node.js 22.0.0

### Operating System

26.5.1 (25F80)

### Installation Method

npm

### Steps to Reproduce

1. Install `ai-sdk-ollama` and `@strands-agents/sdk` (1.4.0).
2. Start Ollama locally with a tool-capable model (e.g., `llama3.1`).
3. Create a Strands agent using `VercelModel` with `ollama("llama3.1")`.
4. Invoke the agent with any prompt.
5. Observe: response streams correctly but `ModelError` is thrown at the end.
6. Add a tool and invoke with a prompt that triggers it.
7. Observe: model generates tool call in the stream but agent silently terminates without executing it.

### Expected Behavior

The `ai-sdk-ollama` provider should emit the complete Vercel AI SDK stream event lifecycle that downstream consumers expect:

1. A `modelMessageStartEvent` (or equivalent) with `role: "assistant"` at the beginning of each response.
2. A correct `finish_reason`: `"stop"` for normal text completions, `"tool_calls"` when the response contains tool call blocks.

This allows any agent framework consuming the Vercel AI SDK stream contract (Strands, LangChain, etc.) to function correctly without custom patching.

### Actual Behavior

Ollama is one of the most popular local LLM runtimes, and `ai-sdk-ollama` is the primary community bridge between Ollama and the Vercel AI SDK ecosystem. When this provider is used with Strands Agents SDK (the recommended path for TypeScript agent development on AWS), two critical features break:

1. **Basic agent invocation** — throws `ModelError` after successful responses due to missing message lifecycle events.
2. **Tool calling** — completely non-functional because the agent loop cannot detect that tools were called.

Developers must write a custom wrapper class (`OllamaModel`) that intercepts the stream and patches in the missing events. This is non-trivial, undocumented, and requires understanding the internal event contract of both the SDK and the provider.

### Possible Solution

In the `ai-sdk-ollama` provider's stream transformation layer, inject the missing lifecycle events:

```typescript
// At the start of each response stream:
[...]

PR #3534, opened 2026-07-29
fix(ts): complete the message when a provider omits the message start event

Ashutosh0x:
## Description

Addresses the remaining Strands-side half of #3190.

That issue reported two failures with `ai-sdk-ollama`. The `finish_reason` half was fixed in #3185 - `vercel.ts` now promotes a terminating `endTurn` to `toolUse` when a client tool call completed, with a comment naming this provider. The other half is still present on `main`:

- `strands-ts/src/models/model.ts:438` - `streamAggregated` only finalizes a message inside `if (messageRole)`
- `messageRole` is assigned in exactly one place, `case 'modelMessageStartEvent'` (`:361`)
- a provider that streams content without emitting that event (`ai-sdk-ollama` omits the Vercel `stream-start` part) therefore leaves `messageRole` null, `stoppedMessage` null, and falls through to `:476-482` throwing `ModelError('Stream ended without completing a message')`

The text streamed fine and `contentBlocks` was fully populated - the response is discarded purely because the provider never announced the message.

## Change

In `case 'modelMessageStopEvent'`, when there is no role but content was accumulated, assume `assistant` (model output is always the assistant turn) and `logger.warn` so the misbehaving provider stays visible rather than being silently papered over.

A stream that produced **no** content at all is still a genuinely broken stream and keeps throwing exactly as before - the existing `ModelError` path and its `SyntaxError` cause are untouched.

## Testing

Three tests added to `strands-ts/src/models/__tests__/model.test.ts` under a new `when the provider omits the message start event` describe, using the existing `TestModelProvider` (no live Ollama required):

- text stream with no start event completes as an assistant message
- tool use with no start event preserves the `ToolUseBlock` and `toolUse` stop reason, so the agent loop still executes the tool
- a stream with no content at all still throws `Stream ended without completing a message`

I was not able to build and run the `strands-ts` suite locally, so these three are **unverified by me** - I would appreciate CI confirming them.

## A note on process

CONTRIBUTING asks contributors to comment and wait for maintainer confirmation before significant work, and I am aware @notowen333 previously steered @jstar0 toward an upstream evidence artifact on #3190 rather than a Strands-side patch - which worked, `ai-sdk-ollama@3.8.8` fixed the finish reason.

I opened this as working code rather than a proposal because the change is ~15 lines and #3185 already established that Strands hardens against exactly this provider's stream deviations. The argument here is symmetry: we already refuse to let a wrong `finish_reason` break the loop, and a missing `stream-start` is the same class of defect with the more severe outcome (the entire response is thrown away).

If you would rather keep this provider-side, say so and I will close it - no attachment to the approach.

## Related Issues

Related to #3190. Mirrors #3185.

## Type of Change

Bug fix

## Checklist

- [ ] I have run the test suite locally (see Testing above)

### DIFF
--- a/strands-ts/src/models/__tests__/model.test.ts
+++ b/strands-ts/src/models/__tests__/model.test.ts
@@ -406,6 +406,67 @@ describe('Model', () => {
       })
     })
 
+    describe('when the provider omits the message start event', () => {
+      it('completes the message by assuming the assistant role', async () => {
+        // ai-sdk-ollama streams content without a Vercel `stream-start` part, so VercelModel
+        // never emits modelMessageStartEvent. See #3190.
+        const provider = new TestModelProvider(async function* () {
+          yield { type: 'modelContentBlockStartEvent' }
+          yield {
+            type: 'modelContentBlockDeltaEvent',
+            delta: { type: 'textDelta', text: 'Hello' },
+          }
+          yield { type: 'modelContentBlockStopEvent' }
+          yield { type: 'modelMessageStopEvent', stopReason: 'endTurn' }
+        })
+
+        const messages = [new Message({ role: 'user', content: [new TextBlock('Hi')] })]
+
+        const { result } = await collectGenerator(provider.streamAggregated(messages))
+
+        expect(result.message.role).toBe('assistant')
+        expect(result.message.content).toEqual([{ type: 'textBlock', text: 'Hello' }])
+        expect(result.stopReason).toBe('endTurn')
+      })
+
+      it('preserves tool use blocks so the agent loop still executes tools', async () => {
+        const provider = new TestModelProvider(async function* () {
+          yield {
+            type: 'modelContentBlockStartEvent',
+            start: { type: 'toolUseStart', toolUseId: 't1', name: 'get_weather' },
+          }
+          yield {
+            type: 'modelContentBlockDeltaEvent',
+            delta: { type: 'toolUseInputDelta', input: '{"city":"Berlin"}' },
+          }
+          yield { type: 'modelContentBlockStopEvent' }
+          yield { type: 'modelMessageStopEvent', stopReason: 'toolUse' }
+        })
+
+        const messages = [new Message({ role: 'user', content: [new TextBlock('Hi')] })]
+
+        const { result } = await collectGenerator(provider.streamAggregated(messages))
+
+        expect(result.message.role).toBe('assistant')
+        expect(result.message.content).toEqual([
+          { type: 'toolUseBlock', name: 'get_weather', toolUseId: 't1', input: { city: 'Berlin' } },
+        ])
+        expect(result.stopReason).toBe('toolUse')
+      })
+
+      it('still throws when the stream produced no content at all', async () => {
+        const provider = new TestModelProvider(async function* () {
+          yield { type: 'modelMessageStopEvent', stopReason: 'endTurn' }
+        })
+
+        const messages = [new Message({ role: 'user', content: [new TextBlock('Hi')] })]
+
+        await expect(async () => await collectGenerator(provider.streamAggregated(messages))).rejects.toThrow(
+          'Stream ended without completing a message'
+        )
+      })
+    })
+
     describe('when streaming reasoning content', () => {
       it('yields complete reasoning block', async () => {
         const provider = new TestModelProvider(async function* () {
--- a/strands-ts/src/models/model.ts
+++ b/strands-ts/src/models/model.ts
@@ -433,19 +433,33 @@ export abstract class Model<T extends BaseModelConfig = BaseModelConfig> {
             break
           }
 
-          case 'modelMessageStopEvent':
-            // Store message and stop reason
-            if (messageRole) {
+          case 'modelMessageStopEvent': {
+            // Store message and stop reason.
+            // Some providers (e.g. ai-sdk-ollama) stream content but never emit a message start
+            // event. Without a role we would discard a perfectly good response and throw
+            // "Stream ended without completing a message" below. Model output is always the
+            // assistant turn, so synthesize the role rather than lose the message. A stream that
+            // produced no content at all is still a genuinely broken stream and keeps throwing.
+            // See #3190.
+            let role = messageRole
+            if (!role && contentBlocks.length > 0) {
+              logger.warn(
+                'model stream completed without a message start event; assuming role "assistant"'
+              )
+              role = 'assistant'
+            }
+            if (role) {
               const filtered = contentBlocks.filter(
                 (block) => !(block instanceof TextBlock && block.text.trim() === '')
               )
               stoppedMessage = new Message({
-                role: messageRole,
+                role,
                 content: filtered,
               })
               finalStopReason = event.stopReason!
             }
             break
+          }
 
           case 'modelMetadataEvent':
             // Store metadata, keeping the last one if multiple events occur


mkmeral:

strands-ts/src/models/model.ts
```diff
@@ -433,19 +433,33 @@ export abstract class Model<T extends BaseModelConfig = BaseModelConfig> {
...
               const filtered = contentBlocks.filter(
                 (block) => !(block instanceof TextBlock && block.text.trim() === '')
               )
               stoppedMessage = new Message({
-                role: messageRole,
+                role,
                 content: filtered,
               })
               finalStopReason = event.stopReason!
             }
```
strandly-the-agent, line 460: 🟡 **The gate reads `contentBlocks` pre-filter, so a whitespace-only stream returns an empty message instead of throwing** — which contradicts the invariant in your own comment on line 442 (and in the PR description).

`contentBlocks.length > 0` is evaluated at line 445, but the whitespace-only `TextBlock` filter doesn't run until line 452. A provider that omits `modelMessageStartEvent` and streams only whitespace therefore synthesizes `role: 'assistant'`, then `filtered` empties the content — no throw. Confirmed on both branches:

```
no messageStart + textDelta '  \n '
  main    (f2482f41): threw ModelError: Stream ended without completing a message
  pr3534  (a73eb31a): returned {"role":"assistant","content":[],"stopReason":"endTurn"}
```

This isn't only reachable through a test double — I drove the production `VercelModel` with the `ai-sdk-ollama` shape from #3190 (`text-start` / `text-delta '\n'` / `text-end` / `finish`, no `stream-start`; `vercel.ts:251-264` emits those independently of `stream-start`) and got the same empty message, which then lands in history unchecked (`agent.ts:2387-2392` pushes unconditionally): `[user, {"role":"assistant","content":[]}]`. On the next turn `bedrock.ts:803-822` and `vercel.ts:488-529` silently drop that turn, while `anthropic.ts:371-397` has no such guard and would forward `{"role":"assistant","content":[]}`. Credit where due: Copilot flagged this first — I'm confirming it with the baseline diff and the downstream trace.

Computing `filtered` first makes the code match the comment. I also folded in the logging-format fix here since it's the same block: `AGENTS.md:56` wants `field=<value> | lowercase message` with no punctuation, and this is the only `logger.warn` in `src/models/*` carrying a semicolon and quotes; attributing it to the provider is what makes it actionable for someone running a third-party adapter (`this.constructor.name` is the established pattern — `retry/model-retry-strategy.ts:102`, `memory/extraction/model-extractor.ts:106`).

```suggestion
            const filtered = contentBlocks.filter(
              (block) => !(block instanceof TextBlock && block.text.trim() === '')
            )
            let role = messageRole
            if (!role && filtered.length > 0) {
              logger.warn(
                `provider=<${this.constructor.name}> | model stream completed without a message start event | assuming assistant role`
              )
              role = 'assistant'
            }
            if (role) {
              stoppedMessage = new Message({
                role,
                content: filtered,
              })
              finalStopReason = event.stopReason!
            }
```

Verified with this applied: 39/39 tests pass (the 38 existing plus the whitespace regression test in my next comment), `tsc --noEmit` and `eslint` clean, and the real-text case still returns `Hello` — only the whitespace/empty shapes throw again.

strands-ts/src/models/__tests__/model.test.ts
```diff
@@ -406,6 +406,67 @@ describe('Model', () => {
...
+        const provider = new TestModelProvider(async function* () {
+          yield { type: 'modelMessageStopEvent', stopReason: 'endTurn' }
+        })
+
+        const messages = [new Message({ role: 'user', content: [new TextBlock('Hi')] })]
+
+        await expect(async () => await collectGenerator(provider.streamAggregated(messages))).rejects.toThrow(
+          'Stream ended without completing a message'
+        )
+      })
```
strandly-the-agent, line 467: 🟡 **This test passes with the fix reverted, so it isn't coverage for the new behavior** — and the case that would actually catch the bug above is missing.

I checked by reverting just the source hunk (`git checkout main -- src/models/model.ts`): `Tests 2 failed | 36 passed (38)` — your first two tests fail (good, they're real), this one still passes. Zero content blocks never set a role on either branch, so it exercises a pre-existing path. It's still worth keeping as a guard on the gate, but the whitespace-only shape is the one that regresses relative to `main`, and nothing covers it.

```suggestion
      it('still throws when the stream produced no content at all', async () => {
        const provider = new TestModelProvider(async function* () {
          yield { type: 'modelMessageStopEvent', stopReason: 'endTurn' }
        })

        const messages = [new Message({ role: 'user', content: [new TextBlock('Hi')] })]

        await expect(async () => await collectGenerator(provider.streamAggregated(messages))).rejects.toThrow(
          'Stream ended without completing a message'
        )
      })

      it('still throws when the stream produced only whitespace text', async () => {
        const provider = new TestModelProvider(async function* () {
          yield { type: 'modelContentBlockStartEvent' }
          yield {
            type: 'modelContentBlockDeltaEvent',
            delta: { type: 'textDelta', text: '   ' },
          }
          yield { type: 'modelContentBlockStopEvent' }
          yield { type: 'modelMessageStopEvent', stopReason: 'endTurn' }
        })

        const messages = [new Message({ role: 'user', content: [new TextBlock('Hi')] })]

        await expect(async () => await collectGenerator(provider.streamAggregated(messages))).rejects.toThrow(
          'Stream ended without completing a message'
        )
      })
```

That new test fails on this branch today and passes with the suggestion from my previous comment — which is what makes it worth adding.

strands-ts/src/models/__tests__/model.test.ts
```diff
@@ -406,6 +406,67 @@ describe('Model', () => {
...
+          yield { type: 'modelMessageStopEvent', stopReason: 'endTurn' }
+        })
+
+        const messages = [new Message({ role: 'user', content: [new TextBlock('Hi')] })]
+
+        const { result } = await collectGenerator(provider.streamAggregated(messages))
+
+        expect(result.message.role).toBe('assistant')
+        expect(result.message.content).toEqual([{ type: 'textBlock', text: 'Hello' }])
+        expect(result.stopReason).toBe('endTurn')
```
strandly-the-agent, line 429: ⚪ Two nits bundled, both from the repo's own testing rules — take or leave, not blocking.

**Assertion style** (lines 427-429, and 450-454): Copilot's comment here is a correct citation, not a style preference. `strands-ts/docs/TESTING.md` → *"Prefer testing entire objects at once instead of individual properties"*, and its ❌ example is exactly this shape. Every other `describe` in this file (e.g. `:47-100`, and the adjacent `:381-405`) asserts the whole `result` with `toEqual`. Worth also asserting the yielded `items` — a whole-object `toEqual` on both is what #3208's single test does, and it would catch the pass-through events changing shape too.

**Comment wording** (lines 411-412): TESTING.md's "Comments in Tests" asks a regression test to *"link the issue it guards against and describe the behavior it now guarantees"*, with `// guards against unbounded retry on a null tool name (#642)` as the model. The current comment re-narrates the production root cause, which the comment at `model.ts:438-443` already covers. Something like:

```ts
        // guards against discarding streamed content when the provider omits the
        // message-start event (#3190)
```

(Deliberately not a one-click `suggestion` block — this thread spans lines 411-429, so applying one would replace the whole test body.)

strandly-the-agent (commented):
**TL;DR — the code works; the *premise* is what needs a human call.** I ran what you couldn't: `model.test.ts` is **38/38 green** with `type-check` clean at `a73eb31a`, and your tests are real (2 of the 3 fail when I revert the source hunk). One 🟡 to fix before merge: the whitespace-only path slips past the throw your own comment promises — Copilot spotted it, and I confirmed it against `main` and through the production `VercelModel`. Two calls that are a maintainer's, not mine: whether #3190's missing-`stream-start` half still reproduces at all, and what to do with the near-identical open PR #3208.

⚠️ **Correcting my own earlier call:** I posted "✅ APPROVED — Ready to ship" on #3208 (functionally identical `model.ts` change) on 2026-07-16 — one day *before* `jstar0` posted the observability bundle showing the missing-`stream-start` half doesn't reproduce on `ai-sdk-ollama` 3.8.7 or 3.8.8. That approval was premature; this review supersedes it.

### ❓ Questions for a maintainer (blocking, in my read)

1. **Does the bug this fixes still exist in the wild?** #3190's automated repro used a *mock* `LanguageModelV3`. `jstar0`'s [evidence bundle](https://gist.github.com/jstar0/10add7c35c08252e9b706a09121f33d5) (2026-07-17) found `stream-start` **present** on both `ai-sdk-ollama@3.8.7` and `@3.8.8`; the reporter never supplied the requested `npm ls`. **But** the gap is real in first-party code regardless: `strands-ts/src/models/openai/chat-adapter.ts:390` only emits `modelMessageStartEvent` `if (delta?.role && !state.messageStarted)`, so any OpenAI-compatible endpoint that omits `delta.role` on the first chunk hits the same discard. So the fix has standalone value — but the *justification* in the description doesn't hold up, and #3190's Vercel/Ollama half may deserve "unreproduced on current releases" rather than "fixed by this PR".
2. **#3208.** Functionally identical `model.ts` change (I traced the truth table — zero behavioral difference), open since 2026-07-12, author unresponsive to review since 2026-07-16. If either lands, this one is the better vehicle: broader tests, live author, and a correct issue citation. Recommend closing #3208 as superseded either way.

<details>
<summary>✅ Verification ledger — what I actually ran</summary>

| Check | Result |
|---|---|
| `npx vitest run src/models/__tests__/model.test.ts` @ `a73eb31a` | **38/38 pass**, `Type Errors: no errors` |
| `npx tsc --noEmit --project src/tsconfig.json`, `npx eslint src/models/model.ts` | clean |
| Are the new tests meaningful? (`git checkout main -- src/models/model.ts`, re-run) | **2 of 3 fail** → real. The third (`still throws when the stream produced no content at all`) passes reverted too — it guards the gate, but isn't coverage for the new behavior |
| Whitespace-only + no `messageStart`, both branches | `main`: **throws** `ModelError`; `pr3534`: returns `{"role":"assistant","content":[],"stopReason":"endTurn"}` |
| Same shape through the production `VercelModel` (no `stream-start`, `text-start`/`text-delta '\n'`/`text-end`/`finish`) | returns the empty-content message → `agent.messages` becomes `[user, {"role":"assistant","content":[]}]` |
| Suggested fix applied (comment 1) | 39/39 pass (38 + a whitespace regression test), `tsc` + `eslint` clean, `V1` text case still returns `Hello` |

Not executed: live Anthropic/Bedrock calls. That an empty-content assistant message 400s is documented API behavior, not something I ran — what I *did* verify is that `anthropic.ts:371-397` has no `content.length > 0` guard while `bedrock.ts:803-822` and `vercel.ts:488-529` both do (they silently drop the turn instead).

`browser`/Playwright vitest project fails to launch chromium in my sandbox — environment limitation, not a defect in this PR.

</details>

<details>
<summary>🟢 Layering, and the label question — no action needed, recorded so it isn't re-litigated</summary>

**Base class vs. adapter.** The obvious objection is that #3206 (`d2fc5080`) hardened this same provider family inside `VercelModel.stream()`, not in the shared aggregator — and this PR does the opposite. I dug in and came out on this PR's side: "the provider never emitted `modelMessageStartEvent`" is only decidable **at end of stream**, and `streamAggregated` is the only code that observes end-of-stream for every provider, whereas #3206's stop-reason repair (like `bedrock.ts:1621-1630`) is locally decidable inside one adapter. Supporting evidence: of the first-party emit sites, four already hardcode `role: 'assistant'` (`vercel.ts:248`, `google/adapters.ts:386-392`, `openai/responses-adapter.ts:429`), and the one that made it conditional — `openai/chat-adapter.ts:390` — is a latent instance of this very bug in our own code. A Vercel-only fix would have been under-scoped.

The honest cost: base-class tolerance doesn't make the raw event stream well-formed, so a custom `Model` subclass with buggy ordering now fails quietly for anyone consuming `stream()` directly. Two follow-ups worth separate issues (not gates on this PR): fix `chat-adapter.ts:390`, and consider having `vercel.ts` lazily emit the event on first content.

**Not blocking, but cheap:** this widens the *inbound* contract on `abstract stream()` (`model.ts:247-254`) from "MUST emit `modelMessageStartEvent`" to "SHOULD emit; omitted + content ⇒ role defaults to `assistant` and warns; omitted + no content ⇒ `ModelError`" — and that requirement is documented nowhere today (`grep -rn "Stream ended without completing" site/ strands-ts/docs/` → zero hits; `custom_model_provider.mdx:530-542` lists the events with no ordering rule). Widening an inbound contract is safe; narrowing it back later isn't. Worth three lines of TSDoc on `abstract stream()` — I can't leave an inline suggestion there since it's outside the diff:

```ts
   * Implementations SHOULD emit `modelMessageStartEvent` before any content event. If it is
   * omitted and content is streamed, `streamAggregated()` defaults the message role to
   * `'assistant'` and logs a warning; the raw `stream()` output is not modified. If it is
   * omitted and no content is streamed, `streamAggregated()` throws `ModelError`.
```

**Label:** no `api/needs-review` needed. `team/API_BAR_RAISING.md:20-21` puts a behavior tweak on an existing method with no signature change squarely in "Minimal changes — PR review is sufficient". Worth noting the green `check-api-review-label` isn't evidence either way — `.github/workflows/api-review-label.yml:22-25` skips when neither label is present and never inspects the diff.

</details>

<details>
<summary>🔍 Checked and clean — deliberately not raising these</summary>

- **Malformed tool JSON is not swallowed.** I expected a regression and there isn't one: the second unconditional throw at `model.ts:519` still fires, keeping the `SyntaxError` as `cause`. Only the message changes (`Stream ended without completing a message` → `unable to parse tool input JSON`), which is more accurate.
- **`maxTokens` and redaction both improve.** No `messageStart` + `maxTokens` now raises a proper `MaxTokensError` carrying the partial message instead of a generic `ModelError`; redaction returns `[REDACTED]` instead of throwing.
- **No role mislabeling.** `Role` is `'user' | 'assistant'` only (`types/messages.ts:163`), a real `messageStart` always wins, and `role: undefined` (reachable via `chat-adapter.ts:392`) goes from throwing to correctly returning `assistant`.
- **No runaway loop.** Empty content + `stopReason: 'toolUse'` still fails loudly at `agent.ts:2215`.
- **Python parity.** `strands-py`'s `process_stream` already defaults the role to `assistant` unconditionally and never raises (`streaming.py:432,463,478`) — this PR *narrows* the TS/Python gap.
- **Pre-existing, not yours:** `{role:'assistant', content:[]}` is already reachable on `main` when `messageStart` *is* emitted with whitespace-only content, and when `messageStart` arrives *after* content (`model.ts:362` resets `contentBlocks`). Closing that needs a guard in `agent._appendMessage` or `anthropic._formatMessages` — a follow-up issue, not this PR's job. Related product call for the maintainer: should a whitespace-only response be a hard error at all, or should the SDK normalize it?
- **CONTRIBUTING:** no violation. The "wait for maintainer confirmation" step is scoped to `ready for contribution` issues (#3190 isn't labeled that), and the same doc asks for small, incremental PRs. You were also upfront about the process wrinkle and about not being able to run the suite — that's the right way to file this.
- One factual fix for the description: the merged prior fix is **#3206** (`d2fc5080`), not #3185.

</details>

---

Reviewed at `a73eb31a` vs `main` `f2482f41`. Automated review — solid work for a human maintainer to approve, not a gate.

JackYPCOnline:
Should we special case it in that provider ?

Issue #2143, opened 2026-04-17
[BUG] MCPClient cleanup raises PythonFinalizationError at interpreter shutdown on Python 3.14


minorun365:
### Checks

- [x] I have searched the existing issues and documentation
- [x] I have provided a clear, reproducible test case
- [x] I have included the relevant versions and environment details

### Strands Version

1.35.0

### Python Version

3.14.3 (CPython, macOS arm64, installed via uv)

### Operating System

macOS 14.4.0 (Darwin 25.4.0, arm64 / Apple Silicon)

### Installation Method

uv (PyPI)

### Steps to Reproduce

When an `Agent` that has an `MCPClient` tool is garbage-collected at interpreter shutdown on **Python 3.14**, the cleanup path in `Agent.__del__` raises `PythonFinalizationError: cannot join thread at interpreter shutdown`. The error is swallowed (`Exception ignored while calling deallocator`) but Strands also emits its own log lines and a traceback to stderr, which is visible to users.

Minimal pattern (excerpted and simplified from a real codebase using the public AWS Knowledge MCP endpoint):

```python
from mcp.client.streamable_http import streamable_http_client
from rich import print as rprint
from rich.text import Text
from strands import Agent
from strands.hooks import HookProvider, HookRegistry
from strands.hooks.events import BeforeToolCallEvent
from strands.models import BedrockModel, CacheConfig
from strands.tools.executors import SequentialToolExecutor
from strands.tools.mcp import MCPClient


class ToolLoggingHook(HookProvider):
    def register_hooks(self, registry: HookRegistry) -> None:
        registry.add_callback(BeforeToolCallEvent, self.on_tool_start)

    def on_tool_start(self, event: BeforeToolCallEvent) -> None:
        rprint(Text(f"{event.agent.name} : {event.tool_use['name']}"))


mcp_client = MCPClient(
    lambda: streamable_http_client("https://knowledge-mcp.global.api.aws")
)

agent = Agent(
    name="RetrievalAgent",
    model=BedrockModel(
        model_id="us.anthropic.claude-haiku-4-5-20251001-v1:0",
        cache_config=CacheConfig(strategy="auto"),
    ),
    system_prompt="Search AWS docs and summarize.",
    tools=[mcp_client],
    tool_executor=SequentialToolExecutor(),
    callback_handler=None,
    hooks=[ToolLoggingHook()],
)

# Trigger multiple MCP tool calls (search + read) then exit.
[...]

Issue #1244, opened 2025-11-24
[BUG] ERROR - Unhandled exception in receive loop: not all arguments converted during string formatting


FilipUrbanik01:
### Checks

- [x] I have updated to the lastest minor and patch version of Strands
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched [./issues](./issues?q=) and there are no duplicates of my issue

### Strands Version

1.18.0

### Python Version

3.11

### Operating System

Win11/NIXOS

### Installation Method

pip

### Steps to Reproduce

Headers include auth header. MCP server is deployed to Amazon Bedrock AgentCore.
```python
  from strands.tools.mcp import MCPClient
  from mcp.client.streamable_http import streamablehttp_client
  from strands import Agent


  streamable_http_mcp_client = MCPClient(lambda: streamablehttp_client("https://bedrock_core_mcp_url/invocations?qualifier=DEFAULT", self.headers     or {}))

  with streamable_http_mcp_client:
          tools = streamable_http_mcp_client.list_tools_sync()
          agent = Agent(
               model="model from Amazon Bedrock currently using haiku 4.5",
               system_prompt="Some system prompt",
               tools=tools)
          response = str(agent(query))
   return response if response else "No response" ```
```

### Expected Behavior

MCP client lists tools and agent uses them to query data.

### Actual Behavior

Init works. MCP client getting session id and protocol version from MCP server. Issue happens at `streamable_http_mcp_client.list_tools_sync()` call.

### Error Details

**Primary Error:**
```
TypeError: not all arguments converted during string formatting
```
### Full Traceback
```python
2025-11-24 09:22:53,355 - root - ERROR - Unhandled exception in receive loop: not all arguments converted during string formatting
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/mcp/shared/session.py", line 420, in _receive_loop
    await self._handle_incoming(
  File "/usr/local/lib/python3.11/site-packages/mcp/client/session.py", line 547, in _handle_incoming
    await self._message_handler(req)
  File "/usr/local/lib/python3.11/site-packages/strands/tools/mcp/mcp_client.py", line 626, in _handle_error_message
    self._log_debug_with_thread("ignoring non-fatal MCP session error", message)
  File "/usr/local/lib/python3.11/site-packages/strands/tools/mcp/mcp_client.py", line 728, in _log_debug_with_thread
    formatted_msg = msg % args if args else msg
[...]

PR #3537, opened 2026-07-29
test(mcp): wait for the retained task, not the log line

### LINKED ISSUE #3536: [BUG] Flaky test: test_call_tool_async_caller_cancellation_timeout_preserves_cancelled_error
### Checks

### SDK Language

Python

### Strands Version

`main` @ `0e517d62` (repo checkout, not an installed release)

### Language Runtime Version

Python 3.13

### Operating System

macOS 26.5.2 (Apple silicon, 15 cores, pytest-xdist)

### Installation Method

git clone

### Steps to Reproduce

Run the Python unit suite repeatedly from `strands-py/` on an unmodified `main`:

```bash
hatch test --python 3.13
```

`tests/strands/tools/mcp/test_mcp_client.py::test_call_tool_async_caller_cancellation_timeout_preserves_cancelled_error` fails intermittently — 1 failure in 3 runs here, plus one during a separate `hatch run prepare`. It passes 100% in isolation; it only appears under parallel full-suite load.

To make it deterministic, add one line to `_invoke_on_background_thread` in `src/strands/tools/mcp/mcp_client.py`, between the log at line 1340 and the await at line 1342:

```python
await asyncio.sleep(0.5)
```

It then fails every time, in isolation, in ~3 seconds.

### Expected Behavior

The test passes regardless of machine load or worker scheduling.

### Actual Behavior

```
    assert "did not finish within bounded cancellation cleanup" in caplog.text
>   assert client._background_cleanup_tasks
E   assert set()
```

The log assertion on the preceding line passes — only the `_background_cleanup_tasks` assertion fails.

### Additional Context

The test waits on the log line as a proxy for "the unfinished invocation has been retained", but the client logs that *before* it retains the task:

```python
if pending:
    self._log_debug_with_thread("MCP invocation did not finish within bounded cancellation cleanup")  # 1340
if cancellation_state is not None:
    await self._cancel_tool_call(cancellation_state)                                                  # 1342
if pending:
    self._track_background_cleanup_task(invoke_event)                                                 # 1346
```

`_cancel_tool_call` is not cheap — it creates a cancellation task and waits on it for up to another second. The test exits its poll the instant the log appears and asserts immediately, observing the client mid-sequence. The client runs its own loop on a separate thread, so nothing orders the two.

Since the log assertion passing proves `pending` was truthy, and `pending` isn't reassigned between 1339 and 1345, line 1346 is guaranteed to run — so the only empty window is between the log and 1346.

| Scenario | Result |
| --- | --- |
| Unmodified, full suite | 1 failure in 3 runs (small sample) |
| Unmodified, isolation | 0 failures |
| Unmodified + injected 0.5s delay | **3 / 3 failed** |
| Fixed test + same delay | 5 / 5 passed |
| Fixed test, no delay, full suite | **30 / 30 passed** |

One caveat I could not rule out: `_cancel_tool_call` catches `except Exception`, which does not catch `asyncio.CancelledError`. If one escaped on this path, line 1346 would be skipped and the set would stay empty *permanently* — same symptom, but not fixable by polling in the test. Thirty clean runs suggest this doesn't happen at a rate that matters, but I couldn't prove it unreachable. If the flake recurs after a test-side fix, that's where I'd look.

Introduced four days ago in #3402, which closed the port request #3276. Worth noting the bounded-cleanup machinery is Python-only — `strands-ts/src/mcp/` has no equivalent of `_background_cleanup_tasks` or the bounded cancellation window, since the TypeScript side cancels via `AbortSignal` on a single thread. So there is no TypeScript counterpart carrying the same bug.

### Possible Solution

Wait for the condition being asserted rather than a log line that precedes it:

```python
for _ in range(150):
    if client._background_cleanup_tasks:
        break
    await asyncio.sleep(0.01)
```

The existing log assertion stays; it just stops being the loop's exit condition. Happy to raise a PR.

### Related Issues

#3402, #3276

vaibhav-walia:
## Description

`test_call_tool_async_caller_cancellation_timeout_preserves_cancelled_error` fails intermittently in the full suite and never in isolation. It polls for a debug log as a proxy for "the unfinished invocation has been retained", but the client logs that before it retains the task:

```python
if pending:
    self._log_debug_with_thread("MCP invocation did not finish within bounded cancellation cleanup")  # 1340
if cancellation_state is not None:
    await self._cancel_tool_call(cancellation_state)                                                  # 1342
if pending:
    self._track_background_cleanup_task(invoke_event)                                                 # 1346
```

`_cancel_tool_call` creates a cancellation task and waits on it for up to another second. The test exits its poll the moment the log appears and asserts immediately, so it can observe the client mid-sequence. The client runs its own event loop on a separate thread, so nothing else orders the two observations — under parallel load the client thread simply isn't scheduled in time and `_background_cleanup_tasks` is still empty.

The fix waits for the condition being asserted instead of a log line that precedes it. The log assertion stays; it just stops being the loop's exit condition.

Test-only change — no source behavior is modified.

## Related Issues

Closes #3536. The test was introduced in #3402.

## Type of Change

Bug fix

## Testing

The flake can be made deterministic by simulating the scheduling delay, adding one line between 1340 and 1342 in `mcp_client.py`:

```python
await asyncio.sleep(0.5)
```

| Scenario | Result |
| --- | --- |
| Before, full suite | 1 failure in 3 runs (small sample) |
| Before, isolation | 0 failures |
| Before + injected delay | **3 / 3 failed** |
| After + same injected delay | 5 / 5 passed |
| After, no delay, full suite | **30 / 30 passed** |

At the observed baseline, 30 consecutive clean runs is ~1 in 90,000 by chance, so the fix is doing real work rather than getting lucky.

**Confidence ~9.5/10, with one caveat I could not close.** `_cancel_tool_call` catches `except Exception`, which does not catch `asyncio.CancelledError`. If one escaped on this cancellation-heavy path, line 1346 would be skipped and the set would stay empty *permanently* — the same symptom, but not fixable by polling in the test. Thirty clean runs suggest that does not happen at a rate that matters, but I could not prove it unreachable; whoever knows this design will likely settle it at a glance. If the flake recurs after this lands, that `except Exception` is where I would look.

For context, the bounded-cleanup machinery is Python-only. `strands-ts/src/mcp/` has no equivalent, since the TypeScript side cancels via `AbortSignal` on a single thread, so there is no TypeScript counterpart to fix.

## Checklist

----

### DIFF
--- a/strands-py/tests/strands/tools/mcp/test_mcp_client.py
+++ b/strands-py/tests/strands/tools/mcp/test_mcp_client.py
@@ -555,8 +555,12 @@ async def test_call_tool_async_caller_cancellation_timeout_preserves_cancelled_e
                 with pytest.raises(asyncio.CancelledError):
                     await caller
                 assert await asyncio.to_thread(call_cancelled.wait, 1)
+                # The client logs the expired window, then sends the cancellation notification, and
+                # only then retains the unfinished invocation. Waiting on the log alone would
+                # observe the client mid-sequence, so wait for the retained task itself. The
+                # client runs on its own thread, so nothing else orders these two observations.
                 for _ in range(150):
-                    if "did not finish within bounded cancellation cleanup" in caplog.text:
+                    if client._background_cleanup_tasks:
                         break
                     await asyncio.sleep(0.01)
 


vaibhav-walia:
@gautamsirdeshmukh looks like you were assigned reviewer automatically, please lmk if you need more info from me.

opieter-aws:

strands-py/tests/strands/tools/mcp/test_mcp_client.py
```diff
@@ -555,8 +555,12 @@ async def resistant_call(name, arguments, read_timeout_seconds, progress_callbac
                 with pytest.raises(asyncio.CancelledError):
                     await caller
                 assert await asyncio.to_thread(call_cancelled.wait, 1)
+                # The client logs the expired window, then sends the cancellation notification, and
```
strandly-the-agent, line 558: 🟡 **(non-blocking, take it or leave it)** The comment explains the *why* well; it just doesn't cite the issue it now guards against. `strands-py/docs/TESTING.md` → "Comments in Tests" asks a regression test for a discovered bug to link its issue, and this file already does that twice (`:1003` "See issue #2143", `:1393` "regression test for issue #1244"). One line:

```suggestion
                # Guards against polling a log line that precedes the state being asserted (#3536).
                # The client logs the expired window, then sends the cancellation notification, and
```

Strictly the rule reads "*only* a regression test for a discovered bug carries an issue reference", and this repairs an existing test's polling rather than adding a new regression test — so it's a judgment call, not a rule violation.

strandly-the-agent (commented):
**Verdict: approve** — correct, well-verified flake fix; nothing merge-blocking. I concur with the `github-actions` Approve from 2026-08-04 at this same head, so rather than restate it: what my passes add is an actual *reproduction* of #3536 and its disappearance, a mutation test showing the retention assertion is load-bearing, and measured poll-budget margins. One non-blocking convention nit inline, two questions.

**Verified at `c61fe297`** (1 file, +5 −1, test-only):

| check | result |
| --- | --- |
| #3536 reproduced (`await asyncio.sleep(0.5)` injected between `mcp_client.py:1340` and `:1342`, isolated worktree) | old poll **3/3 FAIL** with the issue's exact signature (`assert set()`, log assert passing); new poll **3/3 PASS**; old poll un-injected in isolation 3/3 PASS |
| assertion is load-bearing (neuter `_track_background_cleanup_task(invoke_event)` at `mcp_client.py:1346`) | test **fails cleanly** in ~3.3s (`assert set()`), no hang → retention genuinely exercised |
| poll budget | set turns truthy at **1.00–1.02s of the 1.5s** budget (~33% margin), held under 2× CPU oversubscription; drain loop needs 10–20ms of its 1s |
| `ruff check` / `ruff format --check` on the changed file | clean; test passes 8/8 consecutively at head |
| `Closes #3536` | accurate — the issue's own "Possible Solution" is verbatim this diff, and it asks for nothing broader |

⚠️ One pass (adversarial/repro) timed out at 900s; I ran its core repro myself, above. The other three passes completed.

**Questions** (neither blocks):

1. The `except Exception` caveat you flagged in both #3536 and the description is real — `asyncio.CancelledError` subclasses `BaseException`, so an escape there skips `mcp_client.py:1346` and the set stays permanently empty: same symptom, not fixable by polling. There's no tracking issue for it (I searched). Want me to file one, or are you already on it? It's a pre-existing source concern from #3402, correctly out of scope here.
2. If this test ever needs more work: this file already has a deterministic idiom for cross-thread sync (`threading.Event` set from a mock + `await asyncio.to_thread(event.wait, 1)`, `test_mcp_client.py:530-532`). Wrapping `_track_background_cleanup_task` that way would drop the timing budget entirely. Not needed for this fix — noting it because the 1.5s budget's comfortable margin does depend on `send_notification` staying a ~zero-latency mock (the source's own ceiling is two 1s bounded waits).

<details>
<summary>Per-pass breakdown (4 passes: correctness · test-quality · issue-alignment · adversarial)</summary>

- **Correctness/safety** — ordering claim verified at line level: `mcp_client.py:1340` (log) → `:1342` (`_cancel_tool_call`, which itself tracks a short-lived notification task at `:1263`) → `:1346` (retain `invoke_event`). The set cannot become non-empty *before* the log, so the new exit condition can only fire at or after the state the `caplog` assertion on `test_mcp_client.py:567` needs — which is exactly what the old exit condition got wrong. Private-attribute access matches this file's established pattern. Comment is evergreen-rule compliant (it states a hypothetical to justify the design, it doesn't narrate the edit).
- **Test quality** — docstring still matched end-to-end; failure mode is a clean bounded assert, never a hang; mutation test above. Two ⚪ nits in the appendix.
- **Issue alignment** — well-scoped, nothing unrelated in the diff, `Closes #3536` not over-reaching.
- **Adversarial/repro** — timed out; core repro re-run by hand (ledger above), plus the residual-race probe in the appendix.

Passes disagreed on one point: whether the new comment narrates the old approach. Arbitrated as compliant — "Waiting on the log alone *would* observe the client mid-sequence" is a hypothetical justifying the current design, not a "previously/used to" narration.

</details>

<details>

- ⚪ **Whole-set truthiness can be satisfied by the wrong task.** `_cancel_tool_call` tracks its own cancellation-notification task at `mcp_client.py:1263`, before `invoke_event` is retained at `:1346`, and a completed tracked task discards itself — so in principle the poll can break on that task and the next line can re-read an empty set. I probed it rather than asserting it: with a 0.5s gap injected before `:1346` at real `AsyncMock` notification speed, **0 failures in 12**; only when the notification is slowed to 0.5ms (comparable to the 10ms poll tick) does it appear, at **1 failure in 18**. Needs two injections and a sub-tick window, so: not worth acting on. It is the reason `"wait for the retained task itself"` in the comment is slightly optimistic about what's observed.
- ⚪ **Diagnosability.** `assert client._background_cleanup_tasks` (`:568`) renders as a bare `assert set()`. Optional: `assert client._background_cleanup_tasks, "resistant invocation was not retained after bounded cleanup"`.

</details>

*Automated review by `strandly-the-agent` — solid work for a human to approve, not a gate.*

opieter-aws (approved):

Issue #43, opened 2025-05-19
[FEATURE]


AhilanPonnusamy:
### Problem Statement

 Support for vLLM as Model Provider

### Proposed Solution

Create a new vLLM Model Provider

### Use Case

vLLM will be available as a Model Provider in strands

AhilanPonnusamy:
https://github.com/strands-agents/sdk-python/pull/44

PR #3502, opened 2026-07-27
fix(mcp): index document content on hydration with thread safety

dosvk:
Toward #3321 | Tracking: #3317 | Supersedes strands-agents/mcp-server#40 (old repo, archived per @mkmeral)

## Summary

Search was title-only: docs indexed at startup with empty content; `ensure_page()` cached bodies without updating the index, so body-only queries returned nothing and the markdown-aware scoring was dead code.

## Changes

- `ensure_page()` updates the index when a page is hydrated (idempotent on re-hydration)
- Existing TF-IDF scoring is preserved, including markdown multipliers (headers 4×, code 2×, links 2×) and the title boost — hydrated body content now actually feeds it
- `threading.Lock` serializes all index mutations (`add`/`update_content`) and snapshots the posting/df structures for search; document content is read live during scoring, so results are eventually consistent, never corrupt. A deterministic forced-interleaving concurrency test confirms the lost-update race FAILS without the lock and PASSES with it
- `update_content` commits `doc.content` last, so a failed reindex leaves content unchanged and the next access retries cleanly
- Cache assignment moved AFTER indexing succeeds — a failed index leaves the page un-cached for retry
- Optional full prefetch behind `STRANDS_MCP_PREFETCH_ALL` (default OFF; eventual-consistency semantics documented)
- `IndexSearch` marked as internal implementation detail (not a supported public API)
- README updated: index-on-hydration and the prefetch env var

## Testing

- 85 tests pass (hydration indexing, concurrency, failure injection)
- `ruff check` / `ruff format --check` clean
- Concurrency test fails without the lock (`df=1 != len(postings)=2`); passes with it
- Failure-injection test fails pre-cache-reorder (page permanently unsearchable); passes after

## Note on #3321

This makes body content searchable; fully closing #3321 in the default config depends on the `STRANDS_MCP_PREFETCH_ALL` default-on decision, so I've left #3321 open pending that.

## Context

Retarget of strands-agents/mcp-server#40 per @mkmeral's request. Review feedback from that PR and this one is addressed (thread safety, cache ordering, atomicity, README, API scope).

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3544, opened 2026-07-29
fix(mcp): negatively cache failed fetches and cap no_sections content

### LINKED ISSUE #3327: [BUG] strands-mcp: no_sections fallback bypasses the small-doc size gate
### Checks

### SDK Language

Python

### Strands Version

strands-mcp @ main (as merged via #3300)

### Language Runtime Version

Python 3.10+ (any)

### Operating System

Any

### Installation Method

other

### Steps to Reproduce

1. Call `fetch_doc(uri=...)` for a large page whose content has no parseable markdown headers - e.g. any large HTML page, since the regex-based `_html_to_text()` in `doc_fetcher.py` strips all markdown structure during conversion.

### Expected Behavior

Documents over the 8KB `SMALL_DOC_THRESHOLD` return a table of contents (or at worst a bounded amount of content), keeping token cost predictable for the consuming agent.

### Actual Behavior

When `parse_sections()` finds no headers, `fetch_doc` returns the full content regardless of size (`server.py`, `reason: "no_sections"`), bypassing the small-doc gate entirely and dumping the whole document into the model context.

### Additional Context

Part of the strands-mcp improvements tracking issue.

### Possible Solution

Cap the `no_sections` fallback at a reasonable size and return a truncation notice, or fall back to fixed-size chunking with an offset parameter for structureless documents.

### Related Issues

#3317

### LINKED ISSUE #3328: [BUG] strands-mcp: failed fetches not negatively cached, errors swallowed
### Checks

### SDK Language

Python

### Strands Version

strands-mcp @ main (as merged via #3300)

### Language Runtime Version

Python 3.10+ (any)

### Operating System

Any

### Installation Method

other

### Steps to Reproduce

1. Add an unreachable or 404 URL to the indexed llms.txt set (or simulate a network failure)
2. Run `search_docs` queries that rank that URL in the top results

### Expected Behavior

A failed fetch is recorded so subsequent searches do not re-attempt it on every call, and the failure reason is observable (logged and/or surfaced in the tool response).

### Actual Behavior

Two related issues in `cache.ensure_page()`:

- Failed fetches are not negatively cached. Every subsequent search that ranks the broken URL in its top results re-attempts the same fetch synchronously, with up to a 30s timeout each.
- The bare `except Exception: return None` swallows the error entirely. Neither the operator nor the consuming agent ever learns why a fetch failed (DNS, 404, timeout).

### Additional Context

Part of the strands-mcp improvements tracking issue. Interacts with the serial-hydration issue: a single broken URL in the top 5 results adds its full timeout to every affected `search_docs` call.

### Possible Solution

Cache fetch failures with a short TTL, log the exception, and surface a more specific error message than "fetch failed" in the `fetch_doc` response.

### Related Issues

#3317

AnnasMazhar:
Port of [strands-agents/mcp-server#43](https://github.com/strands-agents/mcp-server/pull/43) to the harness-sdk monorepo, incorporating review feedback from strandly-the-agent.

## Related Issues

Fixes #3327
Fixes #3328

### Negative caching for failed fetches
- `_FailedEntry` sentinel class with 5-minute TTL distinguishes not-fetched from failed
- Short-circuit known failures in `ensure_page` instead of re-fetching
- Expired entries are cleared and retried on next access
- Log exceptions via `logger.exception()`

### Cap no_sections fallback content
- Truncate structureless documents at `SMALL_DOC_THRESHOLD` (8KB)
- Append truncation notice: `\n\n… (truncated, no parseable sections)`
- Updated existing test to verify bounded content

### search_docs resilience
- Hydration loop only triggers for `None` entries (not `_FailedEntry` sentinels)
- Snippet builder guards with `hasattr(page, "content")`, falling back to `display_title` for non-Page entries

## Testing
- 2 new test methods in `TestSearchDocsWithNegativeCache` — crash assertion and snippet fallback assertion
- All 52 existing tests pass unchanged

Fixes the same issues as #43 on the archived standalone repo.


ebarkhordar (commented):
`search_docs` decides whether to call `ensure_page` by looking at the raw cache value, so the five minute TTL is never evaluated on that path. A `_FailedEntry` is not `None`, so the URL never re-enters `urls_to_hydrate`, expired or not:

`strands-mcp/src/strands_mcp_server/server.py:60`
```python
dict.fromkeys(doc.uri for _, doc in top if (page := url_cache.get(doc.uri)) is None)
```

The expiry rule lives at `utils/cache.py:97-102`, inside the function this line has just decided not to call. `fetch_doc` is unaffected, since it calls `ensure_page` unconditionally.

Measured at c5993d9 in a clean container, with only `doc_fetcher.fetch_and_clean` stubbed (real `_URL_CACHE`, real `ensure_page`, real `search_docs`), advancing `cache.time.monotonic` past `_FAILED_TTL_SECONDS` rather than sleeping:

```
1. search, network down   : fetches=1 entry=_FailedEntry snippet='Flaky Doc'
2. search, network back   : fetches=0 entry=_FailedEntry snippet='Flaky Doc'
   entry.expired now      : True
3. search, TTL expired    : fetches=0 entry=_FailedEntry snippet='Flaky Doc'
4. direct ensure_page()   : fetches=1 entry=Page         page=Page
```

Step 4 is the same URL through the other path, so the TTL logic itself works. The same probe on the merge base 2b84dfa recovers at step 2 (`fetches=1`, real snippet), because a failed fetch left the entry at `None` there. Net effect: one transient failure degrades that URL's snippet to its title for the life of the process, unless a `fetch_doc` call happens to land on it.

Keeping the expiry rule in one place would fix it, something like:

```python
def needs_hydration(url: str) -> bool:
    cached = _URL_CACHE.get(url)
    if isinstance(cached, _FailedEntry):
        return cached.expired
    return cached is None
```

with the comprehension becoming `if cache.needs_hydration(doc.uri)`.

Neither new test case can catch this: `@patch("strands_mcp_server.server.cache")` replaces the whole module, so `FailedSentinel` never reaches the real `ensure_page`.

AnnasMazhar:
Thanks @ebarkhordar — the TTL expiry gap you identified is fixed and now covered by a regression test.

**Fix** (commit `9075e5ff`): centralized the expiry rule in `cache.needs_hydration()` exactly as you suggested — `search_docs` now checks `cache.needs_hydration(doc.uri)` instead of `url_cache.get(doc.uri) is None`, so both the search path and `ensure_page` evaluate TTL the same way. Your step-4 probe (direct `ensure_page` recovering while the search path stayed stuck) was spot on: the expiry rule lived on a path search_docs had already decided not to call.

**Regression test** (commit `f4e286d7`): you were right that neither new test case could catch this — they patch the whole `cache` module. Added `TestSearchDocsTTLExpiry` which uses the real cache module (only `fetch_and_clean` is stubbed) and verifies:
1. Within TTL: `search_docs` does NOT re-fetch a failed URL (snippet falls back to title)
2. After TTL expiry: `search_docs` re-fetches and recovers the content snippet

Verified it fails on `c5993d9` (pre-fix) and passes on `9075e5ff` — full suite green (53 passed).

arielnabavian (dismissed):

AnnasMazhar:
Heads-up on the bot review gate: the **Auto Strands Review → "Trigger Strands Review"** job has been stuck in `WAITING` on the `manual-approval` environment since 2026-07-31 (https://github.com/strands-agents/harness-sdk/actions/runs/30665233083). It's not failing — it's waiting for a maintainer to click approve on the deployment gate, and external contributors can't approve it themselves.

The fix for the TTL expiry gap (@ebarkhordar's review) is in `9075e5ff` + regression test `f4e286d7`, and the full suite passes (53 tests). All other checks are green. If a maintainer can approve the gate, the bot review will run and the PR can move forward.

arielnabavian:
@strandly-the-agent review

strands-mcp/src/strands_mcp_server/utils/cache.py
```diff
@@ -154,20 +178,38 @@ def ensure_page(url: str) -> doc_fetcher.Page | None:
...
-        # serialized, so a rare concurrent double-fetch is possible - that is
-        # duplicated network work, not index corruption.
+    except Exception:
+        # Only FETCH failures are negatively cached: a failed fetch is treated as
+        # a transient outage and short-circuited for the TTL. Indexing failures
+        # below must NOT be cached here — they are left un-cached so the next call
+        # re-fetches and retries indexing (see the failure handling note above).
+        logger.exception("Failed to fetch page: %s", url)
+        _URL_CACHE[url] = _FailedEntry()  # negatively cache the failure
+        return None
```
strandly-the-agent, line 239: **🔴 The negative-cache write can destroy a page another thread already fetched and indexed (1 of 2 sites)**

This write is unconditional *and* outside `_CACHE_LOCK`, while the success write at `cache.py:224` re-checks `existing` under the lock. Before this PR the only unlocked write was always `None` (`load_links_only`'s `setdefault`), so `existing is not None` at `cache.py:215` reliably meant "a real `Page`". Adding `_FailedEntry` introduced a *second* non-`None` state without updating the readers that rely on that invariant — which is the root cause of both symptoms below.

Re-ran both repros myself at `a4c51f223`:

- **Data loss.** Thread A's fetch fails *after* thread B fetched, indexed and cached a real `Page` for the same URL → A stomps B. The body stays searchable (indexing already committed) but `fetch_doc` returns `{"error": "fetch failed"}` for the full 300 s on content that is sitting right there: `index.search("ten lines")` → 1 hit while `fetch_doc` errors, no retry across 9 user calls, recovering at exactly t+300.0 s.
- **Crash.** `ensure_page` is typed `Page | None`, but the race lets `cache.py:216` return a bare `_FailedEntry`; `fetch_doc` doesn't guard its *return* value (only `search_docs`' cache-*read* path is guarded, at `server.py:70`) and dies at `server.py:150`: `AttributeError: '_FailedEntry' object has no attribute 'content'` — an exception escaping the tool call instead of its documented error dict. This is the same sentinel-dereference class that was blocked upstream on `mcp-server#43`, back via a different path.

Reachability: needs `STRANDS_MCP_PREFETCH_ALL=1` — see the reachability note in the review summary for why I still consider it blocking rather than hardening.

Bonus: the same escape skews `_background_prefetch`'s counters (a clobbered call returns non-`None`, counted as "succeeded"). No separate fix needed once this is patched.

```suggestion
        logger.exception("Failed to fetch page: %s", url)
        with _CACHE_LOCK:
            if _URL_CACHE.get(url) is None:  # never clobber a Page another thread committed
                _URL_CACHE[url] = _FailedEntry()
        return None
```

strands-mcp/src/strands_mcp_server/utils/cache.py
```diff
@@ -154,20 +178,38 @@ def ensure_page(url: str) -> doc_fetcher.Page | None:
...
+    # Commit indexing + caching atomically under _CACHE_LOCK. The network fetch
+    # above runs outside the lock, so a concurrent caller (e.g. the prefetch
+    # daemon) may have hydrated this URL while we were fetching; if so, defer to
+    # its result and skip the redundant reindex. The fetch itself is not
+    # serialized, so a rare concurrent double-fetch is possible - that is
+    # duplicated network work, not index corruption.
+    try:
         with _CACHE_LOCK:
             existing = _URL_CACHE.get(url)
             if existing is not None:
```
strandly-the-agent: **🔴 Same finding, site 2 of 2 — `is not None` no longer means "a real Page"**

With `_FailedEntry` in the cache, this check can defer to a *provisional* failure sentinel and hand it back to callers as if it were a `Page` (the `AttributeError` in the comment above, via the `return existing` on the next line). Making it type-aware also does the right thing on the merits: a freshly fetched-and-indexed page should win over a racing failure marker, not lose to it.

```suggestion
            existing = _URL_CACHE.get(url)
            if isinstance(existing, doc_fetcher.Page):  # not "is not None": a _FailedEntry is provisional
```

`cache.py:186`'s expired-clear wants the same compare-and-swap treatment for consistency, though its worst case is only a double fetch.

strands-mcp/src/strands_mcp_server/utils/cache.py
```diff
@@ -1,16 +1,40 @@
...
+
+    @property
+    def expired(self) -> bool:
+        return time.monotonic() - self._timestamp >= _FAILED_TTL_SECONDS
+
+
 # Global state
 _INDEX: indexer.IndexSearch | None = None
-_URL_CACHE: Dict[str, doc_fetcher.Page | None] = {}  # url -> Page (None if not fetched yet)
+_URL_CACHE: Dict[str, Any] = {}  # url -> Page, None (not fetched), or _FailedEntry (failed)
```
strandly-the-agent, line 46: **🟡 `_URL_CACHE` now grows without bound on failed fetches**

`_is_valid_doc_url` (`server.py:12-18`) validates scheme and host but not *path*, so every distinct `fetch_doc(uri="https://strandsagents.com/<guessed-path>")` that 404s now adds a permanent key here. Measured: 0 → 500 entries after 500 distinct failing calls (~200 B each, `__slots__` keeps it small). Pre-PR a failed fetch wrote nothing at all, so this is new here rather than inherited — and it needs no flag: a model guessing doc URLs is the ordinary case, and this is a long-lived process with no eviction.

Cheapest fix is to only negatively-cache URLs the server already knows about, so novel guesses fail fast on the network instead of accumulating forever — combined with the lock fix above:

```python
    except Exception:
        logger.exception("Failed to fetch page: %s", url)
        if url in _URL_TITLES:  # only negatively-cache known catalog URLs
            with _CACHE_LOCK:
                if _URL_CACHE.get(url) is None:
                    _URL_CACHE[url] = _FailedEntry()
        return None
```

An LRU/size cap on the dict is the other option. Flagging the choice rather than prescribing one.

strands-mcp/src/strands_mcp_server/server.py
```diff
@@ -157,14 +158,22 @@ def fetch_doc(uri: str = "", section: str = "") -> Dict[str, Any]:
...
+            # Re-encode from decoded string to avoid surrogates
+            content = encoded[: threshold - trunc_len].decode("utf-8", errors="ignore") + trunc_msg
         return {
             "url": uri,
             "title": page.title,
             "document_small": True,
             "reason": "no_sections",
-            "content": page.content,
+            "content": content,
         }
```
strandly-the-agent: **🟡 A truncated payload still claims to be a whole document**

`document_small: True` used to be an honest signal on this path — the branch returned the entire doc. Now the content can be silently cut with the same flag, no byte count, and no machine-readable truncation marker; the docstring above still tells the model "For small documents (under ~8KB), the full content is returned directly" and documents `content` as "Full document content". A model that trusts that answers from a partial document without knowing it's partial, and `fetch_doc` has no offset/pagination parameter — so there's no way for it to fetch the tail even if it suspects.

The in-band `…` notice helps a little, but prose is a weak channel: it can also land inside a code fence the cut left unclosed (appendix item), which is exactly where a model is least likely to read it as a system annotation.

```suggestion
        content = page.content
        threshold = text_processor.SMALL_DOC_THRESHOLD
        encoded = content.encode("utf-8")
        truncated = len(encoded) > threshold
        if truncated:
            trunc_msg = "\n\n… (truncated, no parseable sections)"
            trunc_len = len(trunc_msg.encode("utf-8"))
            # Re-encode from decoded string to avoid surrogates
            content = encoded[: threshold - trunc_len].decode("utf-8", errors="ignore") + trunc_msg
        return {
            "url": uri,
            "title": page.title,
            "document_small": True,
            "reason": "no_sections",
            "truncated": truncated,
            "content": content,
        }
```

Also worth a test — mutation-testing showed the suite still passes 90/90 with the notice removed entirely:

```python
def test_no_h2_headers_truncated_content_ends_with_truncation_notice(self, mock_cache, no_h2_doc):
    ...
    assert result["content"].endswith("(truncated, no parseable sections)")
```

(Whether `document_small` should stay `True` when the content is truncated is the payload-contract question in the review summary — your call, not something I'd change unilaterally.)

strands-mcp/src/strands_mcp_server/utils/cache.py
```diff
@@ -154,20 +178,38 @@ def ensure_page(url: str) -> doc_fetcher.Page | None:
...
+    # Short-circuit on known transient failures — don't re-fetch
+    if isinstance(cached, _FailedEntry):
+        if cached.expired:
+            # TTL expired — clear and fall through to re-fetch
+            _URL_CACHE[url] = None
+        else:
+            return None
+    else:
+        if cached is not None:
+            return cached
```
strandly-the-agent, line 222: **🟡 `needs_hydration` is untested where it matters, because `ensure_page` re-implements it**

The docstring says the expiry rule lives in `needs_hydration` "so both `ensure_page` and the `search_docs` hydration loop use the same expiry logic" — but `ensure_page` duplicates the check inline here instead of calling it. That's what makes the following mutant survive: force `needs_hydration` to return `True` unconditionally and the suite still passes **90/90**, because this inline check independently prevents the re-fetch. The observable cost of that regression would be a `ThreadPoolExecutor` + `ensure_page` call per result on *every* query — silent, and nothing would catch it.

`TestSearchDocsTTLExpiry` is the only real-cache end-to-end test and it's load-bearing (it's the sole thing catching a revert of `ebarkhordar`'s fix), but it only asserts on the `_FailedEntry` branch. Two small tests close the gap — both validated: they pass at head and fail against the mutants.

```python
# tests/test_cache.py
class TestNeedsHydration:
    def setup_method(self):
        cache._URL_CACHE = {}

    def test_returns_true_when_url_not_cached(self):
        assert cache.needs_hydration("https://strandsagents.com/new.md") is True

    def test_returns_false_for_cached_page_with_content(self):
        url = "https://strandsagents.com/x.md"
        cache._URL_CACHE[url] = doc_fetcher.Page(url=url, title="T", content="hello")
        assert cache.needs_hydration(url) is False

    def test_returns_false_for_unexpired_failed_entry(self):
        url = "https://strandsagents.com/y.md"
        cache._URL_CACHE[url] = cache._FailedEntry()
        assert cache.needs_hydration(url) is False

    def test_returns_true_for_expired_failed_entry(self):
        url = "https://strandsagents.com/z.md"
        entry = cache._FailedEntry()
        entry._timestamp -= cache._FAILED_TTL_SECONDS + 1
        cache._URL_CACHE[url] = entry
        assert cache.needs_hydration(url) is True
```

plus the integration half — seed `_URL_CACHE` with a `Page(content="already have this")`, monkeypatch `real_cache.ensure_page` to raise, and assert `search_docs` never calls it.

Alternatively, have `ensure_page` call `needs_hydration` so there's genuinely one implementation — then the existing tests cover both paths.

strandly-the-agent (changes requested):
**Request changes** — one reproducible concurrency defect (a crash *and* silent data loss) with a ~5-line fix; everything else is 🟡/⚪. The TTL centralization is genuinely right: `ebarkhordar`'s finding is fixed and `TestSearchDocsTTLExpiry` is what holds it in place.

**Verified** at `a4c51f223` (clean venv, py3.13): `pytest tests/` → **90 passed** · `ruff check .` → clean · `ruff format --check .` → clean. Both blocker repros re-run by hand and confirmed.

- 🔴 `cache.py:200` — the failure write clobbers a `Page` another thread already fetched **and indexed**, and `ensure_page` can hand back the sentinel itself → `AttributeError` straight out of `fetch_doc`. Requires `STRANDS_MCP_PREFETCH_ALL=1` (documented at `README.md:57`, default off), so: real and cheap to fix, but opt-in-gated.
- 🟡 `_URL_CACHE` now grows without bound — any guessed URL that 404s leaves a permanent entry (measured 0 → 500 for 500 distinct calls). No flag needed.
- 🟡 a truncated `no_sections` payload still says `document_small: True` with no machine-readable `truncated` signal, while the docstring still promises full content.
- 🟡 two mutants survive the suite (`needs_hydration` → always `True`; truncation notice dropped). Two small tests close both.

**Questions (non-blocking):** (1) `#3327` and `#3328` are both still open here and you'd already linked this work to them — any reason `Fixes #3327` / `Fixes #3328` isn't in the body? Related: `#3328`'s "surface a more specific error than `fetch failed`" ask is still untouched at `server.py:147` — intentional follow-up? (2) Is `document_small: True` meant to promise *complete* content, or just *small enough not to need sectioning*? The truncation path now makes those two different things.

<details><summary>How this was reviewed, and the reachability call</summary>

Pipeline: triage → context build → 5 parallel fresh-context passes (correctness, adversarial/repro, test quality, LLM-context, issue alignment) → aggregation. All at `a4c51f223` vs merge-base `11ad6366a`.

**Why the 🔴 is 🔴 even though it needs a flag.** I measured the dispatch model rather than assuming it: `mcp==1.29.0` calls sync tool bodies inline on the event loop (`func_metadata.py` `call_fn_with_arg_validation` → `return fn(...)`, no `to_thread`), and `search_docs`' own pool can't self-race a URL because `dict.fromkeys` de-dupes (`server.py:60`). So the **only** in-process source of concurrent `ensure_page` calls on one URL is the prefetch daemon — with `STRANDS_MCP_PREFETCH_ALL` off, both symptoms are unreachable. But that flag exists precisely so queries can run *while* the daemon walks the catalog (`README.md:57`), the unlocked window is the whole fetch (32–39 ms measured), transient failures on real catalog URLs do happen (1 of 160 sampled fetches failed at 5-way concurrency, then succeeded 3/3 on retry), and one symptom is an uncaught exception rather than the tool's documented `{"error": ...}` contract. Documented path + plausible input + consequence worse than a clean error = 🔴, flagged as flag-conditional so you can weigh it.

**Cleared, not findings** (so nobody re-raises them): the dropped `or not page.content` clause in the hydration predicate looks like a regression but isn't — base `ensure_page` short-circuited on *any* non-`None` cache value, so an empty-content `Page` was never re-fetched there either. Zero behavioural delta; head is marginally better (no pointless pool spin-up). Also checked and clean: TTL boundary semantics (`monotonic`, `>=`, no early/late expiry), the truncation byte arithmetic (20 probes, output never exceeded 8192 B, notice never lost, multi-byte cuts drop cleanly with no mojibake), and the `238cd4aa4` fetch-vs-indexing split — which is a correct and necessary reconciliation with `#3502`'s "leave un-cached on index failure" contract that upstream `#43` never had to make.

</details>

<details><summary>Appendix — non-blocking (7)</summary>

- ⚪ `cache.py:79` — `_background_prefetch` still checks `_URL_CACHE.get(url) is None` directly instead of `needs_hydration(url)`, so it skips a negatively-cached URL for the rest of that pass even past the TTL, and its `attempted` counter stays 0 — which silences the "all N fetches failed" ERROR. One-line fix while you're in the file; `needs_hydration`'s docstring already claims all callers share this logic.
- ⚪ `cache.py:262-267` — `get_url_cache()`'s docstring still says "Page objects (or None if not fetched)" with no mention of `_FailedEntry`, even though the type is now `Dict[str, Any]`. Someone writing new code against that docstring reproduces exactly the `AttributeError` class the new `hasattr` guard exists to prevent.
- ⚪ `server.py:166` — `if len(encoded) > threshold:` is always true here; the size gate at `server.py:150` already returned for anything ≤ threshold. Harmless, but it reads as a live conditional.
- ⚪ `cache.py:203-204` — now outside the `try`, so an exception there escapes `ensure_page` and `list(executor.map(...))` re-raises it, failing the whole `search_docs` call. Contrived (needs a non-str title, which the real fetcher never produces).
- ⚪ The truncation notice can land *inside* a code fence the cut left unclosed, where a model may read it as text in the fence rather than a system annotation. Not reachable against today's catalog (no doc in a 160-page live sample hits this branch at all; the largest section-less doc is 2213 B), but it's the one signal this fix adds. Closing an odd fence before appending — or cutting on a fence boundary — would harden it.
- ⚪ `238cd4aa4` (the newest, least-reviewed part of the diff) has no dedicated regression test; undoing it is caught only incidentally, by `test_indexing_failure_leaves_page_unsearchable_forever`, which predates this PR.
- ⚪ `TestSearchDocsTTLExpiry` clears `_URL_CACHE` on entry but leaves a populated entry behind, unlike its siblings in `test_indexer_concurrency.py` which reset the globals via `autouse` fixtures. No order-dependency today (green across 9 randomized seeds) — just a latent trap for the next real-cache test added after it.

</details>

Solid work overall — this is a real fix for `#3327`/`#3328` and the port is faithful. Findings are mine as an AI reviewer; a human maintainer should make the merge call.

AnnasMazhar:
Friendly bump on this review — the branch is synced with `main` (merge 2026-08-19; CI re-ran on `71869b9f`: Bot PR Review Gate ✅ and API Review Label Check ✅). The bot's CHANGES_REQUESTED findings from 08-11 were addressed in `cbb5febe` (cache race on failed fetches, bounded `_URL_CACHE` growth, `needs_hydration` test coverage).

The remaining blocker is the **Auto Strands Review** job, still `WAITING` on the `manual-approval` environment since 2026-07-31 (flagged in my 08-05 comment). Is there anything else I can provide to move the review forward?

AnnasMazhar:
Friendly bump — any updates on this review? The branch is still mergeable and green (last synced with main 2026-08-19; 90/90 tests pass). The only open blocker remains the Auto Strands Review job waiting on the manual-approval environment (since 07-31) — that gate needs a maintainer click, which external contributors cannot do. Happy to make any further changes if there is anything else to address.

strands-mcp/src/strands_mcp_server/utils/cache.py
```diff
@@ -37,6 +65,28 @@ def _is_prefetch_enabled() -> bool:
...
+    if len(_URL_CACHE) <= _MAX_CACHE_ENTRIES:
+        return
+
+    # Pass 1: drop expired _FailedEntry sentinels
+    expired_keys = [k for k, v in _URL_CACHE.items() if isinstance(v, _FailedEntry) and v.expired]
+    for k in expired_keys:
+        del _URL_CACHE[k]
+
+    # Pass 2: evict oldest (first-inserted) entries until under the cap
+    while len(_URL_CACHE) > _MAX_CACHE_ENTRIES:
```
arielnabavian: **concern** — this pass sheds the entries worth keeping and protects the ones the cap exists to bound.

`load_links_only` seeds every catalog URL first (`_URL_CACHE.setdefault(url, None)`, `cache.py:161`), and later hydration overwrites the value in place — which does not move the key. So the catalog sits in the oldest positions and `next(iter(_URL_CACHE))` hands it over first, while the guessed-404 sentinels that pushed the cache over the cap are the newest and never reached.

Measured at `71869b9f` — 200 hydrated catalog Pages, then 4000 distinct guessed URLs driven through the real `ensure_page` failure path:

```
after 4000 guessed 404s: total=4096 cap=4096
  catalog Pages surviving : 96/200
  junk _FailedEntry kept  : 4000/4000
  catalog URLs dropped from cache entirely: 104
  needs_hydration(doc0.md) -> True (re-fetch required)
```

Every dropped catalog URL goes back to `needs_hydration -> True`, so the next search that ranks it re-fetches over the network. A run of bad URIs turns a warm cache cold and holds it cold while the run continues — the inverse of what the comment at `cache.py:16-17` promises.

Either of these fixes it: in pass 2, prefer keys not in `_URL_TITLES` (non-catalog first), or evict `_FailedEntry` values before any `Page`. I'd pick the first — evicting unexpired sentinels re-opens the re-fetch you just paid to avoid.

Separately, none of this is tested. Lines 80-87 get zero coverage from `tests/`, and swapping `_evict_if_needed` for `lambda: None` leaves 95/95 passing — so nothing stops it regressing, and nothing noticed that it regresses the catalog today.

strands-mcp/src/strands_mcp_server/utils/cache.py
```diff
@@ -154,23 +204,52 @@ def ensure_page(url: str) -> doc_fetcher.Page | None:
...
     """
-    page = _URL_CACHE.get(url)
-    if page is not None:
-        return page
+    cached = _URL_CACHE.get(url)
+    # Short-circuit on known transient failures — don't re-fetch
+    if isinstance(cached, _FailedEntry):
+        if cached.expired:
+            # TTL expired — clear and fall through to re-fetch
+            _URL_CACHE[url] = None
```
arielnabavian: **concern** — this write is outside `_CACHE_LOCK`, and it defeats the no-clobber guard at line 231.

Interleaving, both threads on one URL:

- T1 reads an expired `_FailedEntry` and is about to run this line
- T2 finishes a successful fetch, takes `_CACHE_LOCK`, indexes, commits its `Page`
- T1 runs this line → T2's `Page` becomes `None`
- T1's own fetch fails → the guard at 231 sees `None` (T1 put it there) → writes a fresh `_FailedEntry`

A page fetched and indexed moments ago is now negatively cached for the full five minutes. `fetch_doc` returns `{"error": "fetch failed"}` for it, `search_docs` won't retry because `needs_hydration` is `False`, and the index still holds the body — so index and cache disagree.

Measured at `71869b9f`. I forced the interleaving by making the `time.monotonic()` call inside `expired` run T2's commit; that controls *when* T2 runs, not what either thread does:

```
before        : _URL_CACHE[url] = expired _FailedEntry
  T2 committed  : _URL_CACHE[url] = Page
  T1 returned   : NoneType
after         : _URL_CACHE[url] = _FailedEntry

T2's Page survived?          False
fetch_doc would return:      {"error": "fetch failed"}
body term still in index?    True  (index and cache now disagree)
```

Reachability is the same gate as the race already reported here — it needs a second concurrent `ensure_page` on one URL, so `STRANDS_MCP_PREFETCH_ALL=1`. `TestRaceConditionFailureDoesNotClobberPage` can't catch it: it seeds `None`, so this line never executes.

This write only exists so the failure handler refreshes an expired sentinel. Doing that refresh in the handler removes the need for it:

```python
    cached = _URL_CACHE.get(url)
    if isinstance(cached, _FailedEntry):
        if not cached.expired:
            return None
        # expired: fall through and re-fetch, but do NOT write here — this runs
        # outside _CACHE_LOCK and would clobber a Page committed in this window
    elif cached is not None:
        return cached
```

```python
        with _CACHE_LOCK:
            existing = _URL_CACHE.get(url)
            if existing is None or (isinstance(existing, _FailedEntry) and existing.expired):
                _URL_CACHE[url] = _FailedEntry()
                _evict_if_needed()
```

I ran this: the repro above ends with T2's `Page` intact and `fetch_doc` returning the page, negative caching still refreshes across TTL windows (fetch count holds at 2 across two windows), and the suite stays green.

strands-mcp/src/strands_mcp_server/server.py
```diff
@@ -157,14 +169,22 @@ def fetch_doc(uri: str = "", section: str = "") -> Dict[str, Any]:
...
-    # No parseable sections: treat as small doc regardless of size
+    # No parseable sections: return bounded content to protect token budget.
+    # The early return above already handled docs <= threshold, so here content
+    # is always over the threshold and always truncated.
     if not sections:
+        threshold = text_processor.SMALL_DOC_THRESHOLD
+        encoded = page.content.encode("utf-8")
+        trunc_msg = "\n\n… (truncated, no parseable sections)"
+        trunc_len = len(trunc_msg.encode("utf-8"))
+        content = encoded[: threshold - trunc_len].decode("utf-8", errors="ignore") + trunc_msg
```
arielnabavian: **consider** — two things about this branch, both about the contract rather than the arithmetic. The byte math is fine.

**The tail has no way back.** There's no `offset`/`chunk` parameter, and `section` is silently ignored on this path. `parse_sections` only treats `##` as top-level (`text_processor.py:294`), so this branch catches any doc that skips level 2 — not only the flattened-HTML case from #3327. An 18.8 KB doc whose headings are all `###`:

```
document size: 18,825 bytes, headings: 81 x '###', 0 x '##'

fetch_doc(uri)                        -> 8,192 of 18,825 bytes returned
  contains the tail marker?              False
fetch_doc(uri, section='1')           -> identical to the TOC-mode body: True
fetch_doc(uri, section='FINAL STEP')  -> identical to the TOC-mode body: True
```

10.6 KB of documentation has no path back to the model, and an agent that reads `truncated: True` and tries to drill in gets the same 8 KB every time with no error. #3327 asked for "a table of contents (or at worst a bounded amount of content)", so this meets the letter of it — but adding a parameter afterwards changes the tool signature, which clients see. Building a TOC from `###` when no `##` exists would address the cause instead of the symptom.

**`document_small: True` on truncated content.** The docstring now says the field "does not guarantee the content is complete", which redefines a field that already shipped. A client branching on it to mean "I have the whole document" silently gets 43% of one. `truncated` and `reason` already carry the real signal, so I'd stop setting `document_small` on this path rather than redefine it.

strands-mcp/src/strands_mcp_server/utils/cache.py
```diff
@@ -154,23 +204,52 @@ def ensure_page(url: str) -> doc_fetcher.Page | None:
...
+        else:
+            return None
```
arielnabavian: **consider** — this short-circuit is silent, and that's the half of #3328 still open.

#3328 asks that "the failure reason is observable (logged and/or surfaced in the tool response)". `logger.exception` covers the first failure. Every call for the next five minutes returns from here with no log record at all, while `fetch_doc` keeps answering `{"error": "fetch failed"}`. An operator watching a repeatedly-failing doc sees one traceback and then silence, which reads like the failure stopped.

One line is enough — `logger.debug("negative cache hit for %s, %ds remaining", ...)` — or return a distinguishable error so the agent can tell a cached failure from a fresh one.

strands-mcp/src/strands_mcp_server/utils/cache.py
```diff
@@ -195,11 +297,12 @@ def get_index() -> indexer.IndexSearch | None:
     return _INDEX
 
 
-def get_url_cache() -> Dict[str, doc_fetcher.Page | None]:
+def get_url_cache() -> Dict[str, Any]:
```
arielnabavian: **consider** — `Dict[str, Any]` makes every caller responsible for a type it isn't allowed to name.

`_FailedEntry` is module-private, so code outside `cache.py` can't write `isinstance(v, _FailedEntry)`. That leaves the duck-type at `server.py:70` (`hasattr(page, "content")`) as the only option, which is why that guard reads oddly. There's exactly one caller today, so the shape is cheap to fix now and awkward once there are several.

A narrow accessor keeps the union inside the module:

```python
def get_page(url: str) -> doc_fetcher.Page | None:
    """Return the cached Page, or None if not fetched or negatively cached."""
    cached = _URL_CACHE.get(url)
    return cached if isinstance(cached, doc_fetcher.Page) else None
```

`search_docs` then drops the `hasattr` guard, and `get_url_cache()` no longer has to be part of the public surface.

strands-mcp/tests/test_server.py
```diff
@@ -190,3 +209,112 @@ def test_fetch_failure_returns_error(self, mock_cache):
...
+
+        url = "https://strandsagents.com/flaky.md"
+        doc = Doc(uri=url, display_title="Flaky Doc", content="", index_title="")
+
+        # Stub only the index plumbing; keep the real _URL_CACHE, needs_hydration,
+        # ensure_page, and search_docs.
+        index = type("Index", (), {"search": lambda self, q, k=5: [(1.0, doc)]})()
+        monkeypatch.setattr(real_cache, "get_index", lambda: index)
+        monkeypatch.setattr(real_cache, "ensure_ready", lambda: None)
+        real_cache._URL_CACHE.clear()
```
arielnabavian: **concern** — the one test holding `ebarkhordar`'s fix in place depends on what ran before it.

`real_cache._URL_CACHE.clear()` wipes the real module dict on the way in, and the test leaves a hydrated `Page` for `flaky.md` in it on the way out. `monkeypatch` restores `get_index`, `ensure_ready` and `fetch_and_clean` — it does not restore `_URL_CACHE`, and nothing else does.

The three new classes in `test_indexer_concurrency.py` each got an autouse `reset_cache_state` fixture. This class didn't, and it's the one the 08-11 review called load-bearing: it is the only end-to-end check that a `_FailedEntry` re-enters `urls_to_hydrate` after its TTL expires. Order-dependent state is a weak guard on exactly the regression you most want held.

`monkeypatch.setattr(real_cache, "_URL_CACHE", {})` in place of the `.clear()` fixes both halves — teardown reverts it, and the test stops caring what ran first.

strands-mcp/src/strands_mcp_server/utils/cache.py
```diff
@@ -1,16 +1,44 @@
...
+
+class _FailedEntry:
+    """Sentinel for a negatively cached fetch failure with TTL expiry.
+
+    A page whose fetch failed is stored as a _FailedEntry instead of
+    being re-fetched immediately. Once TTL expires, the entry is treated
+    as stale and the next ensure_page call will retry the fetch.
+    """
+
+    __slots__ = ("_timestamp",)
```
arielnabavian: **consider** — the sentinel keeps the timing and drops the reason, which is what the `cache.py:214` comment needs to read.

`logger.exception` at `cache.py:229` captures why the fetch failed and then discards it, so "a distinguishable error so the agent can tell a cached failure from a fresh one" means re-plumbing the failure path rather than reading a field. One slot now keeps that a small change later:

```python
class _FailedEntry:
    __slots__ = ("_timestamp", "reason")

    def __init__(self, reason: str = "") -> None:
        self._timestamp = time.monotonic()
        self.reason = reason
```

Cheap to add while the sentinel has exactly one construction site; annoying once `fetch_doc` has to surface it and the type has no room for it.

phanisaimunipalli:
Drive-by note, and an apology first: I opened #4062 and #4063 against #3328 and #3327 earlier today without seeing this PR. My search for existing work only covered the most recent open PRs, which with 684 open here did not reach back to 29 July. Both of mine are now closed. This one is the more complete change, and it got there a month earlier.

One piece of #3328 that looks like it is still open here, in case it is useful:

The issue asks for two things from the error path, and this PR does the first (`logger.exception`) but not the second: *"surface a more specific error message than `fetch failed` in the `fetch_doc` response"*. `test_fetch_doc_failed_entry_returns_error` still asserts the bare string, so a consuming model cannot tell a 404 from a DNS failure from a timeout.

Since `_FailedEntry` already exists and holds the timestamp, it could carry the reason at no extra cost:

```python
class _FailedEntry:
    def __init__(self, reason: str) -> None:
        self._timestamp = time.monotonic()
        self.reason = reason
```

and in `fetch_doc`:

```python
entry = _URL_CACHE.get(uri)
reason = getattr(entry, "reason", None)
return {"error": f"fetch failed: {reason}" if reason else "fetch failed", "url": uri}
```

with the reason recorded as `f"{type(exc).__name__}: {exc}"` where the exception is caught. That yields `"fetch failed: HTTPError: HTTP Error 404: Not Found"`.

Entirely your call whether it belongs here or in a follow-up, and I am not going to open anything against it. Two other things I ran into that may be worth knowing:

- Scoping the negative cache to the **fetch** specifically matters. My first attempt wrapped the indexing step too, which broke `TestBlocker2CacheBeforeIndexing` — `update_content`'s contract deliberately keeps indexing failures immediately retryable. Your line 196 comment suggests you already accounted for this; flagging it only because it is easy to regress later.
- We independently landed on the same observation that the `no_sections` branch is only reachable for a document already over `SMALL_DOC_THRESHOLD`, so the second size check there is dead code.

AnnasMazhar:
Thanks @phanisaimunipalli for the note — and for closing #4062/#4063 in favor of this one.

I've folded in the error-path piece from #3328: `_FailedEntry` now records the exception text at failure time, and `fetch_doc` surfaces it — a consuming model now gets `fetch failed: HTTPError: HTTP Error 404: Not Found` instead of a bare `fetch failed` (commit 58a783dfb). Covered by a unit test on the recording path (test_cache.py) and an end-to-end test through the real cache module with only the network call stubbed (test_server.py); 98/98 non-integ tests pass locally.

Agreed on both observations: scoping the negative cache to the fetch (leaving indexing failures immediately retryable) is deliberate, and the `no_sections` branch is only reachable past SMALL_DOC_THRESHOLD — the comment in server.py documents that.

Thanks again for the review.

AnnasMazhar:
Friendly bump — any updates on this review? Status since my last note: #4062/#4063 were closed in favor of this PR and the error-path piece from #3328 is folded in; branch is synced with main and CI is green (90/90 tests pass). The only remaining blocker is the Auto Strands Review → "Trigger Strands Review" job sitting in WAITING on the manual-approval environment. Happy to adjust anything in the diff if there are new findings.

AnnasMazhar:
Friendly bump — two weeks on, and both remaining blockers are outside the diff.

**Findings addressed.** The 08-11 review points landed in `cbb5febe`: one `needs_hydration()` rule now governs expired failures everywhere, the failed-fetch / concurrent-`Page` race is resolved under a lock, `_URL_CACHE` growth is bounded by eviction, and `tests/test_indexer_concurrency.py` carries the regression coverage. `58a783df` adds the `#3328` ask — `fetch_doc` now surfaces `fetch failed: <ExceptionType>: <detail>` rather than the bare string.

**Still not moving from our side.** The `Run integration tests` and `Trigger Strands Review` runs have been `waiting` on the `manual-approval` environment since 2026-07-31.

**One thing worth a decision:** the branch is now 201 commits behind `main` (last synced 08-19). I can re-sync with a current merge if that makes the review easier, or leave the diff exactly as you already have it — either is fine, just say which.

AnnasMazhar:
Re-verified against current `main` — fresh numbers, since the branch has drifted further since the last sync.

- Head `58a783dfb`; GitHub reports `MERGEABLE` (no conflicts).
- `main` is now 279 commits on from this branch's merge base (`c186207cf` -> `fec042766`, 2026-09-22), up from 201 when I last posted. Of the five files this PR touches, four have zero `main` commits in that range; only `strands-mcp/tests/test_server.py` has one.
- I merged `fec042766` into a scratch worktree of this branch (merge result `cdab3b409`, clean) and ran the package suites: **47 passed, 0 failed** (`tests/test_cache.py`, `tests/test_server.py`, `tests/test_indexer_concurrency.py`).
- Non-approval-gated checks are green. `Trigger Strands Review` has been `status=waiting` on the `manual-approval` environment since 2026-08-30T13:12Z and has never run, so no automated review signal exists on this PR.
- For completeness: the 08-11 `CHANGES_REQUESTED` sits on `a4c51f223`, which the head supersedes — `cbb5febe3` (an ancestor of head) landed the points it raised: one `needs_hydration()` rule for expired failures, the failed-fetch/concurrent-`Page` race under a lock, and a bounded `_URL_CACHE`, with `tests/test_indexer_concurrency.py` as coverage.

56 days open. The question from my last note is still unanswered and only got more relevant: re-sync this branch onto current `main`, or review the diff as it stands? Either works from our side — and if the intended direction here has changed over those 279 commits, a short note saying so would save repeating this re-verification every cycle.

PR #3550, opened 2026-07-29
fix(telemetry): record gen_ai.tool.call.arguments/result on execute_tool spans

poshinchen:
## Description

Under `gen_ai_latest_experimental`, tool payloads on `execute_tool` spans are recorded only inside the `gen_ai.input.messages` / `gen_ai.output.messages` envelopes — attributes the OTel GenAI semantic conventions define for inference spans, not tool spans. Consumers implementing the execute_tool span convention read `gen_ai.tool.call.arguments` and `gen_ai.tool.call.result` (added to the spec in v1.38) and find nothing, so every spec-compliant backend renders an empty tool span.

This PR additionally records `gen_ai.tool.call.arguments` at tool-span start and `gen_ai.tool.call.result` at tool-span end, in both the Python and TypeScript tracers. The existing message-envelope events are unchanged, so current consumers see no difference. Per the spec, the result attribute is scoped to successful executions and is omitted when the tool status is `error` (failures remain visible via `gen_ai.tool.status` and the span's error status). Both attributes are recorded directly as span attributes regardless of the `gen_ai_span_attributes_only` token, since span attributes are their native home in the convention.

In the Python tracer, both attributes participate in the `gen_ai_unredacted_attributes` redaction policy under their own canonical names — an existing allowlist entry for `gen_ai.input.messages`/`gen_ai.output.messages` does not cover them, so users who want tool payloads unredacted need to add the new names (or a `gen_ai.tool.call.*` glob). The TypeScript tracer has no redaction mechanism, so the values are recorded directly, consistent with all other content attributes there.

## Documentation PR

No documentation changes needed; the attributes follow the OTel GenAI execute_tool span convention and are documented in the tracer docstring.

## Type of Change

Bug fix

## Testing

Python: full tracer unit suite passes, including latest-conventions tool-span tests asserting both the new attributes and the unchanged event emission. TypeScript: tracer unit suite plus the agent/swarm/graph tracer suites pass; lint, type-check, and format checks are clean.

## Checklist

----

### DIFF
--- a/strands-py/src/strands/telemetry/tracer.py
+++ b/strands-py/src/strands/telemetry/tracer.py
@@ -94,7 +94,9 @@ class Tracer:
     span events, for backends that cannot read span events. This applies to the aggregated content events
     (latest-convention "gen_ai.*" messages and the "memory.query"/"memory.content" events); the legacy
     per-message events are always emitted as events. The default (token absent, non-Langfuse) emits span
-    events for backward compatibility.
+    events for backward compatibility. Independent of this token, tool spans under the latest
+    conventions additionally record `gen_ai.tool.call.arguments` and `gen_ai.tool.call.result` as
+    span attributes per the execute_tool span convention.
 
     Span attribute redaction is opt-in via the `gen_ai_unredacted_attributes=<list>` token in the same
     environment variable. The list uses `;` as a separator and supports trailing-`*` glob patterns.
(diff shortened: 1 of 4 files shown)


poshinchen:
> **Issue** (suggestion): Under latest conventions, tool input/output is now emitted twice per tool span — once in the dedicated `gen_ai.tool.call.arguments`/`gen_ai.tool.call.result` attributes and again inside `gen_ai.input.messages`/`gen_ai.output.messages` — and the new attributes are always set on the span regardless of `_span_attributes_only`/event-only mode. For large tool payloads this roughly doubles the serialized span size and may bump against exporter attribute-size limits.
> 
> **Suggestion**: This is spec-mandated for the execute_tool convention, so likely intentional — just confirm the duplication and always-on behavior are acceptable for large results (e.g. base64 blobs), and consider a follow-up note in the design/decisions if not already captured.

This is intentional.

lizradway (approved):

PR #3551, opened 2026-07-29
feat: support Tool Choice for Gemini in Python

### LINKED ISSUE #1129: [FEATURE] Implement Gemini Tool Choice
### Problem Statement

We are not currently passing tool choice to the model:
https://github.com/strands-agents/sdk-python/blob/main/src/strands/models/gemini.py#L356-L357

### Proposed Solution

We pass tool choice to the model

### Use Case

We need tool choice so we can force tool invocation

### Alternatives Solutions

_No response_

### Additional Context

_No response_

opieter-aws:
## Description

`GeminiModel.stream()` accepted a `tool_choice` and discarded it — the docstring read `Note: Currently unused.` and nothing downstream consumed it. The practical cost is the event loop's forced structured-output retry: it re-invokes with `tool_choice={"any": {}}` expecting an API-level guarantee of a tool call, and on Gemini that retry was forced in name only, leaving `StructuredOutputException` reachable where Anthropic and Bedrock guarantee a call. This ports the mapping that already exists in the TypeScript SDK (`strands-ts/src/models/google/model.ts`), closing a py↔ts parity gap.

| Strands `ToolChoice` | Gemini `FunctionCallingConfigMode` |
|---|---|
| `{"auto": {}}` | `AUTO` |
| `{"any": {}}` | `ANY` |
| `{"tool": {"name": "X"}}` | `ANY` + `allowed_function_names=["X"]` |

A `tool_choice` with no tool specs is a no-op, matching TS — there is nothing to choose from.

### Public API Changes

No new surface: `tool_choice` was already on `GeminiModel.stream()` and on the base `Model` protocol. Two behavioral notes for reviewers.

**An explicit `tool_config` in `params` takes precedence over a per-request `tool_choice.`** Two earlier commits on this branch tried the other direction, then a field-level merge; both were abandoned because a Gemini `ToolConfig` carries fields a Strands `ToolChoice` cannot express (`retrieval_config`, `stream_function_call_arguments`), so each merge rule left another user-set field behind. Letting the explicit config win passes it through whole and matches the other providers and the TypeScript SDK, which all spread `params` last.

```python
model = GeminiModel(model_id=..., params={"tool_config": my_tool_config})

model.stream(messages, tool_specs=specs, tool_choice={"any": {}})  # my_tool_config still wins, unchanged
```

**`tool_choice` is now keyword-only**, matching `Model.stream()` and every other provider. Gemini was the only one missing the `*`, so a positional fourth argument was already outside the declared interface, and was silently ignored:

```python
# Before: accepted positionally, then discarded
model.stream(messages, tool_specs, system_prompt, {"any": {}})

# After
model.stream(messages, tool_specs, system_prompt, tool_choice={"any": {}})
```

## Related Issues

Resolves #1129

## Documentation PR

N/A

## Type of Change

New feature

## Testing

Summary: 64/64 `test_gemini.py` pass (57 pre-existing + 7 new), 265 pass across the model/event-loop/structured-output suites, `ruff format --check` + `ruff check` + `mypy` clean, and the same suite passes against the declared `google-genai` floor (1.67.0) as well as 2.14.0.

## Checklist
- [x] I have reviewed and understand every line of code in this PR, including any generated by AI tools, and I can explain why it works — **this PR was written by an agent; a human reviewer needs to own this box before merge.**
- [x] I have updated the documentation accordingly — docstrings only; no `site/` page covers per-provider `tool_choice`
- [x] My changes generate no new warnings — asserted by `test_stream_tool_choice_no_warning`

----

### DIFF
--- a/strands-py/src/strands/models/gemini.py
+++ b/strands-py/src/strands/models/gemini.py
@@ -298,7 +298,7 @@ class GeminiModel(Model):
             tool_choice: Selection strategy for tool invocation.
 
         Returns:
-            Gemini tool config, or None when no strategy is requested.
+            Gemini tool config, or None when no recognized strategy is requested.
         """
         if tool_choice is None:
             return None
@@ -338,7 +338,8 @@ class GeminiModel(Model):
             system_prompt: System prompt to provide context to the model.
             params: Additional model parameters (e.g., temperature).
             tool_choice: Selection strategy for tool invocation. Applied only alongside tool specs, since there
-                is nothing to choose from without them. Takes precedence over a tool config set in params.
+                is nothing to choose from without them. A forcing tool choice takes precedence over the function
+                calling config of a tool config set in params.
 
         Returns:
             Gemini request config.
@@ -347,8 +348,16 @@ class GeminiModel(Model):
 
         tool_config = self._format_tool_choice(tool_choice) if tool_specs else None
         if tool_config is not None:
-            # A per-request tool choice is more specific than the tool config in params, so it takes precedence.
-            config_params["tool_config"] = tool_config
+            existing = config_params.get("tool_config")
+            if existing is None:
+                config_params["tool_config"] = tool_config
+            elif tool_choice is not None and ("any" in tool_choice or "tool" in tool_choice):
+                # A per-request choice that forces a tool is more specific than the tool config in params, so it
+                # wins - but only over the field a ToolChoice can express. An "auto" choice imposes no constraint,
+                # so it never clobbers an explicit params config.
+                config_params["tool_config"] = genai.types.ToolConfig.model_validate(existing).model_copy(
+                    update={"function_calling_config": tool_config.function_calling_config}
+                )
 
         return genai.types.GenerateContentConfig(
             system_instruction=system_prompt,
@@ -575,7 +584,9 @@ class GeminiModel(Model):
             tool_specs: List of tool specifications to make available to the model.
             system_prompt: System prompt to provide context to the model.
             tool_choice: Selection strategy for tool invocation. Applied only when tool specs are provided,
-                and takes precedence over a tool config set through the params config.
+                since there is nothing to choose from without them. A forcing tool choice takes precedence over
+                the function calling config of a tool config set through the params config; an "auto" choice
+                leaves it in place.
             **kwargs: Additional keyword arguments for future extensibility.
 
         Yields:
--- a/strands-py/src/strands/models/gemini.py
+++ b/strands-py/src/strands/models/gemini.py
@@ -338,7 +338,7 @@ class GeminiModel(Model):
             system_prompt: System prompt to provide context to the model.
             params: Additional model parameters (e.g., temperature).
             tool_choice: Selection strategy for tool invocation. Applied only alongside tool specs, since there
-                is nothing to choose from without them.
+                is nothing to choose from without them. Takes precedence over a tool config set in params.
 
         Returns:
             Gemini request config.
@@ -347,8 +347,8 @@ class GeminiModel(Model):
 
         tool_config = self._format_tool_choice(tool_choice) if tool_specs else None
         if tool_config is not None:
-            # A tool config set directly through params is the more specific instruction, so it wins.
-            config_params.setdefault("tool_config", tool_config)
+            # A per-request tool choice is more specific than the tool config in params, so it takes precedence.
+            config_params["tool_config"] = tool_config
 
         return genai.types.GenerateContentConfig(
             system_instruction=system_prompt,
@@ -574,7 +574,8 @@ class GeminiModel(Model):
             messages: List of message objects to be processed by the model.
             tool_specs: List of tool specifications to make available to the model.
             system_prompt: System prompt to provide context to the model.
-            tool_choice: Selection strategy for tool invocation. Applied only when tool specs are provided.
+            tool_choice: Selection strategy for tool invocation. Applied only when tool specs are provided,
+                and takes precedence over a tool config set through the params config.
             **kwargs: Additional keyword arguments for future extensibility.
 
         Yields:
--- a/strands-py/src/strands/models/gemini.py
+++ b/strands-py/src/strands/models/gemini.py
@@ -19,7 +19,7 @@ from ..types.content import ContentBlock, ContentBlockStartToolUse, Messages, Sy
 from ..types.event_loop import Usage
 from ..types.exceptions import ContextWindowOverflowException, ModelThrottledException, ProviderTokenCountError
 from ..types.streaming import StreamEvent
-from ..types.tools import ToolChoice, ToolSpec
+from ..types.tools import ToolChoice, ToolChoiceToolDict, ToolSpec
 from ._defaults import resolve_config_metadata
 from ._validation import _has_location_source, validate_config_keys
 from .model import BaseModelConfig, Model
@@ -288,11 +288,46 @@ class GeminiModel(Model):
             tools.extend(self.config["gemini_tools"])
         return tools
 
+    @staticmethod
+    def _format_tool_choice(tool_choice: ToolChoice | None) -> genai.types.ToolConfig | None:
+        """Format a tool choice into a Gemini tool config.
+
+        - Docs: https://googleapis.github.io/python-genai/genai.html#genai.types.ToolConfig
+
+        Args:
+            tool_choice: Selection strategy for tool invocation.
+
+        Returns:
+            Gemini tool config, or None when no strategy is requested.
+        """
+        if tool_choice is None:
+            return None
+
+        allowed_function_names: list[str] | None = None
+        if "auto" in tool_choice:
+            mode = genai.types.FunctionCallingConfigMode.AUTO
+        elif "any" in tool_choice:
+            mode = genai.types.FunctionCallingConfigMode.ANY
+        elif "tool" in tool_choice:
+            # Gemini has no single-tool mode, so a specific tool is ANY narrowed to that one function.
+            mode = genai.types.FunctionCallingConfigMode.ANY
+            allowed_function_names = [cast(ToolChoiceToolDict, tool_choice)["tool"]["name"]]
+        else:
+            return None
+
+        return genai.types.ToolConfig(
+            function_calling_config=genai.types.FunctionCallingConfig(
+                mode=mode,
+                allowed_function_names=allowed_function_names,
+            ),
+        )
+
     def _format_request_config(
         self,
         tool_specs: list[ToolSpec] | None,
         system_prompt: str | None,
         params: dict[str, Any] | None,
+        tool_choice: ToolChoice | None = None,
     ) -> genai.types.GenerateContentConfig:
         """Format Gemini request config.
 
@@ -302,14 +337,23 @@ class GeminiModel(Model):
             tool_specs: List of tool specifications to make available to the model.
             system_prompt: System prompt to provide context to the model.
             params: Additional model parameters (e.g., temperature).
+            tool_choice: Selection strategy for tool invocation. Applied only alongside tool specs, since there
+                is nothing to choose from without them.
 
         Returns:
             Gemini request config.
         """
+        config_params = dict(params or {})
+
+        tool_config = self._format_tool_choice(tool_choice) if tool_specs else None
+        if tool_config is not None:
+            # A tool config set directly through params is the more specific instruction, so it wins.
+            config_params.setdefault("tool_config", tool_config)
+
         return genai.types.GenerateContentConfig(
             system_instruction=system_prompt,
             tools=self._format_request_tools(tool_specs),
-            **(params or {}),
+            **config_params,
         )
 
     def _format_request(
@@ -318,6 +362,7 @@ class GeminiModel(Model):
         tool_specs: list[ToolSpec] | None,
         system_prompt: str | None,
         params: dict[str, Any] | None,
+        tool_choice: ToolChoice | None = None,
     ) -> dict[str, Any]:
         """Format a Gemini streaming request.
 
@@ -328,12 +373,13 @@ class GeminiModel(Model):
             tool_specs: List of tool specifications to make available to the model.
             system_prompt: System prompt to provide context to the model.
             params: Additional model parameters (e.g., temperature).
+            tool_choice: Selection strategy for tool invocation.
 
         Returns:
             A Gemini streaming request.
         """
         return {
-            "config": self._format_request_config(tool_specs, system_prompt, params).to_json_dict(),
+            "config": self._format_request_config(tool_specs, system_prompt, params, tool_choice).to_json_dict(),
             "contents": [content.to_json_dict() for content in self._format_request_content(messages)],
             "model": self.config["model_id"],
         }
@@ -518,6 +564,7 @@ class GeminiModel(Model):
         messages: Messages,
         tool_specs: list[ToolSpec] | None = None,
         system_prompt: str | None = None,
+        *,
         tool_choice: ToolChoice | None = None,
         **kwargs: Any,
     ) -> AsyncGenerator[StreamEvent, None]:
@@ -527,8 +574,7 @@ class GeminiModel(Model):
             messages: List of message objects to be processed by the model.
             tool_specs: List of tool specifications to make available to the model.
             system_prompt: System prompt to provide context to the model.
-            tool_choice: Selection strategy for tool invocation.
-                Note: Currently unused.
+            tool_choice: Selection strategy for tool invocation. Applied only when tool specs are provided.
             **kwargs: Additional keyword arguments for future extensibility.
 
         Yields:
@@ -537,7 +583,9 @@ class GeminiModel(Model):
         Raises:
             ModelThrottledException: If the request is throttled by Gemini.
         """
-        request = self._format_request(messages, tool_specs, system_prompt, self.config.get("params"))
+        request = self._format_request(
+            messages, tool_specs, system_prompt, self.config.get("params"), tool_choice=tool_choice
+        )
 
         client = self._get_client().aio
 
--- a/strands-py/src/strands/models/gemini.py
+++ b/strands-py/src/strands/models/gemini.py
@@ -337,9 +337,7 @@ class GeminiModel(Model):
             tool_specs: List of tool specifications to make available to the model.
             system_prompt: System prompt to provide context to the model.
             params: Additional model parameters (e.g., temperature).
-            tool_choice: Selection strategy for tool invocation. Applied only alongside tool specs, since there
-                is nothing to choose from without them. A forcing tool choice takes precedence over the function
-                calling config of a tool config set in params.
+            tool_choice: Selection strategy for tool invocation.
 
         Returns:
             Gemini request config.
@@ -348,16 +346,10 @@ class GeminiModel(Model):
 
         tool_config = self._format_tool_choice(tool_choice) if tool_specs else None
         if tool_config is not None:
-            existing = config_params.get("tool_config")
-            if existing is None:
-                config_params["tool_config"] = tool_config
-            elif tool_choice is not None and ("any" in tool_choice or "tool" in tool_choice):
-                # A per-request choice that forces a tool is more specific than the tool config in params, so it
-                # wins - but only over the field a ToolChoice can express. An "auto" choice imposes no constraint,
-                # so it never clobbers an explicit params config.
-                config_params["tool_config"] = genai.types.ToolConfig.model_validate(existing).model_copy(
-                    update={"function_calling_config": tool_config.function_calling_config}
-                )
+            # A tool config set in params wins, matching the other providers and the TypeScript SDK. A tool
+            # config expresses more than a ToolChoice can, so the explicit one is left whole rather than
+            # merged with, or replaced by, the narrower per-request choice.
+            config_params.setdefault("tool_config", tool_config)
 
         return genai.types.GenerateContentConfig(
             system_instruction=system_prompt,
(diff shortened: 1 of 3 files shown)


opieter-aws:
@strandly-the-agent review this pr

strands-py/src/strands/models/gemini.py
```diff
@@ -302,14 +337,23 @@ def _format_request_config(
...
 
         Returns:
             Gemini request config.
         """
+        config_params = dict(params or {})
+
+        tool_config = self._format_tool_choice(tool_choice) if tool_specs else None
+        if tool_config is not None:
+            # A per-request tool choice is more specific than the tool config in params, so it takes precedence.
+            config_params["tool_config"] = tool_config
```
strandly-the-agent: **🟡 `{"auto": {}}` clobbers the user's `params["tool_config"]` — and every choice replaces the whole object instead of merging.**

This line fires for *any* recognized `tool_choice`. Reproduced with `params={"tool_config": ToolConfig(mode=ANY, allowed_function_names=["safe_tool"])}` and `tool_specs=[safe_tool, danger_tool]`:

```
tool_choice=None                  -> {allowed_function_names: [safe_tool], mode: ANY}   (intact)
tool_choice={"auto": {}}          -> {mode: AUTO}          <-- narrowing gone; model regains danger_tool
tool_choice={"any": {}}           -> {mode: ANY}           <-- retrieval_config gone too
```

Two separate problems on one line:

1. **`auto` shouldn't override at all.** It imposes no constraint, and it's the SDK's own sentinel for *unspecified* (`bedrock.py:317`). My PR-body justification cited Bedrock — but `bedrock.py:379` is `is_forcing_tool = tool_choice is not None and ("any" in tool_choice or "tool" in tool_choice)`, which **deliberately excludes `auto`**. The cited precedent argues against what I wrote.
2. **Replacing the object discards fields `ToolChoice` can't express.** `ToolConfig` also carries `retrieval_config` and `include_server_side_tool_invocations`; a user's geolocation retrieval config is silently dropped by a `{"any": {}}` — and that lands on the framework-forced structured-output retry with *zero* user action, which `DECISIONS.md:120` explicitly carves out of pay-for-play. It also had 2 fields at the declared floor (1.67.0) and 3 at 2.15.0, so Google adds fields inside our declared range and "replace" silently discards whatever ships next.

The forced retry only ever sends `{"any": {}}` (`_structured_output_context.py:87`), so gating on *forcing* costs the motivating fix nothing.

```suggestion
        if tool_config is not None:
            existing = config_params.get("tool_config")
            if existing is None:
                config_params["tool_config"] = tool_config
            elif tool_choice is not None and ("any" in tool_choice or "tool" in tool_choice):
                # A per-request choice that forces a tool is more specific than the tool config in params, so it
                # wins - but only over the field a ToolChoice can express. Mirrors the is_forcing_tool gate in
                # BedrockModel, which likewise treats "auto" as imposing no constraint.
                config_params["tool_config"] = genai.types.ToolConfig.model_validate(existing).model_copy(
                    update={"function_calling_config": tool_config.function_calling_config}
                )
```

I applied and ran exactly this: **64/64 pass** (no test changes needed), `ruff` clean, `mypy` unchanged from base, and both repros fixed — `auto` preserves the narrowing, `any`/`tool` still force, `retrieval_config` survives. One implementer trap worth knowing: `ToolConfig.model_validate(x)` returns *the same object* for a `ToolConfig` input, which is why the `.model_copy(...)` matters — without it you'd leak request N's forced `ANY` into N+1.

Severity note so this isn't over-weighted: hitting the `auto` case needs a direct `model.stream()` call or an `InvokeModelStage.Input` middleware — there's no Agent-level `tool_choice` today. That's why 🟡 and not 🔴. Worth also adding a test pinning `auto` + `params` in whichever direction you land on; nothing covers it today (`test_gemini.py:1120`/`:1147` only use `any`).

strands-py/tests_integ/models/test_model_gemini.py
```diff
@@ -204,6 +204,50 @@ def test_agent_with_gemini_code_execution_tool(gemini_tool_model):
...
+        model.stream(messages, tool_specs=tool_specs, tool_choice={"tool": {"name": "tool_time"}}),
+    )
+
+    tru_tool_names = [
+        event["contentBlockStart"]["start"]["toolUse"]["name"]
+        for event in events
+        if "contentBlockStart" in event and "toolUse" in event["contentBlockStart"]["start"]
+    ]
+    exp_tool_names = ["tool_time"]
+    assert tru_tool_names == exp_tool_names
```
strandly-the-agent, line 249: **🟡 nit: exact-list equality is a flake risk against a live model.**

`{"tool": {"name": "tool_time"}}` maps to `ANY` + `allowed_function_names=["tool_time"]`, and Gemini's `ANY` mode doesn't guarantee a *single* call — parallel calls to the same allowed function are plausible. `["tool_time", "tool_time"]` would fail this brand-new test for a reason unrelated to what it's proving (that the narrowing held and `tool_weather` was never reachable).

```suggestion
    assert tru_tool_names and all(tool_name == "tool_time" for tool_name in tru_tool_names)
```

The sibling test (`any` → `stopReason == "tool_use"`) is already robust this way — it doesn't care which or how many. Caveat: I could not run either integ test (no `GOOGLE_API_KEY`), so this is a design concern, not an observed flake.

strands-py/src/strands/models/gemini.py
```diff
@@ -328,12 +373,13 @@ def _format_request(
             tool_specs: List of tool specifications to make available to the model.
             system_prompt: System prompt to provide context to the model.
             params: Additional model parameters (e.g., temperature).
+            tool_choice: Selection strategy for tool invocation.
```
strandly-the-agent, line 377: ⚪ Agreeing with the existing `github-actions` thread on this line rather than opening a competing one — the fuller wording is worth mirroring here, since `tool_specs`/`system_prompt`/`params` are already byte-identical between `_format_request_config` and `_format_request` and `tool_choice` is the only line that wasn't.

One accuracy point while it's being touched: "Takes precedence over a tool config set in params" is stated unconditionally in both `_format_request_config:341` and `stream:577-578`, but an unrecognized key returns `None` from `_format_tool_choice` (`:315-316`), in which case `params` still wins — and if the inline fix above lands, precedence becomes *forcing*-only. Suggested wording for all three once that settles:

> `tool_choice`: Selection strategy for tool invocation. Applied only alongside tool specs, since there is nothing to choose from without them. A forcing tool choice takes precedence over a tool config set in params.

strandly-the-agent (changes requested):
**Heads up before anything else: this is a self-review.** I wrote this branch (at `opieter-aws`'s request on #1129), so I'm reviewing my own code. I mitigated that with six independent fresh-context passes and then re-verified the two strongest findings myself — but weight the evidence below, not my own rationale in the PR body.

**Verdict: changes requested — one real defect, in the precedence rule I added in the second commit.**

🟡 `gemini.py:349-351` *replaces* a user's `params["tool_config"]` for **every** recognized `tool_choice` — including `{"auto": {}}`, which imposes no constraint. An explicit `allowed_function_names` narrowing silently vanishes, as does any orthogonal field (`retrieval_config`). Reproduced independently, twice. Fix is 8 lines, verified green — inline.

🟡 A new integ test asserts exact-list equality on a Gemini `ANY`-forced call; `ANY` doesn't guarantee call count, so it's a flake risk — inline.

⚪ I've endorsed the existing unresolved `github-actions` docstring thread rather than duplicating it.

Also: the two `tests_integ` cases have **never been executed by anyone** — no `GOOGLE_API_KEY` in my sandbox then or now, and the Testing section doesn't say so. Please let CI exercise them before merge. Suggested label: `api/needs-review` (solo-approvable, not an API meeting).

<details>
<summary>Verification ledger, three non-blocking questions, and what I suppressed</summary>

### Ledger
| | |
|---|---|
| ✅ Reviewed | branch `agent-tasks/1129` @ `a720d2c8`, base `2b84dfa5` |
| ✅ Tests | `pytest tests/strands/models/test_gemini.py -q` → **64/64**, python 3.12.13, google-genai 2.15.0 |
| ✅ Lint/types | `ruff check` + `ruff format --check` clean; `mypy` shows **1** error, pre-existing and byte-identical on `2b84dfa5` (`count_tokens` arg-type, unrelated) |
| ✅ Repro 1 | auto-clobber confirmed on a mocked client (details inline) |
| ✅ Repro 2 | `retrieval_config` set via `params` is discarded by **any** recognized `tool_choice`, incl. `{"any":{}}` on the forced-retry path |
| ✅ Suggested fix | applied + run: 64/64, ruff clean, mypy unchanged, both repros fixed |
| ✅ PR's core claim | **accurate** — a real `Agent(structured_output_model=…)` shows call #1 `tool_config: null` → call #2 (forced retry) `mode: ANY`; on `main` call #2 carried nothing |
| ✅ kw-only `*` | safe — no caller in the tree passes `tool_choice` positionally; `AGENTS.md:205` mandates the `*`; already shipped as non-breaking for other providers |
| ✅ Dep floor | suite green against the declared floor `google-genai==1.67.0` |
| ❓ **Not verified** | the 2 `tests_integ` cases — no `GOOGLE_API_KEY`, never run by anyone |

1. **Precedence scope.** `gemini.py:349` makes Gemini the only provider where a per-request `tool_choice` beats static `params` — every sibling (`anthropic.py:256/258`, `openai.py:520/521`, `bedrock.py:356`) and TS (`model.ts:344-346`) let static config win. #1129 asked us to stop *dropping* `tool_choice`, not to set a conflict-resolution policy. My own first commit (`877e2d72`) used `setdefault` and matched everyone; the second reversed it. Worth an explicit sign-off plus a short `DECISIONS.md` entry, and a tracking issue for `AnthropicModel`'s identical latent gap (`anthropic.py:258` splats last, defeating the force at `anthropic.py:541`)?
2. **Shared bug with TS.** `model.ts:344-346` resolves this same question the *opposite* way, untested on either side. Rather than this PR turning a shared latent bug into a silent cross-SDK divergence, worth a `strands-ts` issue so both converge on one contract?
3. **PR body.** The Bedrock analogy I wrote ("how BedrockModel resolves a conflicting thinking config") doesn't hold up: `bedrock.py:362-388` is forcing-only *and* field-scoped — it removes one key, it doesn't replace the object. Worth tightening that sentence so the next provider author doesn't inherit it as precedent.

### Suppressed (kept off the PR deliberately)
- **`gemini_tools` + `tool_choice` silent drop** — looks like a bug, isn't: `gemini.py:684-703` forbids `FunctionDeclaration`s in `gemini_tools`, so there's never a function to force. Gating on `tool_specs` is correct.
- **Malformed `{"tool": {}}` raises `KeyError` inside the provider** — convention-consistent with `anthropic.py:271`, unreachable via typed callers.
- **kw-only `*` pinned by no test** (mutation survivor) — repo-wide gap; no provider suite pins it. Not this PR's job.
- **`if tool_specs` silent no-op vs `warn_on_tool_choice_not_supported`** — adjudicated: keep silent. Matches TS, and the orchestrator always pairs `tool_choice` with `tool_specs`, so a warning nobody can act on is noise.
- **Private `_format_request_config` gaining a 4th positional arg** — private helper, pay-for-play per `DECISIONS.md`.
- **No `site/` page documents per-provider `tool_choice`** — pre-existing and SDK-wide (no provider page mentions it); the checklist claim is accurate.
- Fixture placement, `GeminiConfig.params` docstring imprecision, test granularity — style/pre-existing.

Mutation testing (separate pass): 7 mutants, 6 killed — the new tests genuinely have teeth. The one survivor is deleting the `*`.
</details>

opieter-aws:
@strandly-the-agent can you review again?

strands-py/src/strands/models/gemini.py
```diff
@@ -302,14 +337,32 @@ def _format_request_config(
...
+
+        tool_config = self._format_tool_choice(tool_choice) if tool_specs else None
+        if tool_config is not None:
+            existing = config_params.get("tool_config")
+            if existing is None:
+                config_params["tool_config"] = tool_config
+            elif tool_choice is not None and ("any" in tool_choice or "tool" in tool_choice):
+                # A per-request choice that forces a tool is more specific than the tool config in params, so it
+                # wins - but only over the field a ToolChoice can express. An "auto" choice imposes no constraint,
+                # so it never clobbers an explicit params config.
```
strandly-the-agent: **🟡 The merge is one level too shallow — `stream_function_call_arguments` is still discarded on a forcing choice.**

This replaces the *whole* `function_calling_config`, so the one `FunctionCallingConfig` field a `ToolChoice` can't express is dropped. `FunctionCallingConfig` has exactly three fields (`allowed_function_names`, `mode`, `stream_function_call_arguments`), so this is the only one affected — but it's the same species as the bug this commit just fixed.

Verified on this HEAD with `params={"tool_config": ToolConfig(function_calling_config=FunctionCallingConfig(mode=NONE, allowed_function_names=["safe_tool"], stream_function_call_arguments=True), retrieval_config=...)}`:

```
None                       stream_function_call_arguments=True           retrieval_config=kept
{"auto": {}}               stream_function_call_arguments=True           retrieval_config=kept
{"any": {}}                stream_function_call_arguments=** DROPPED **  retrieval_config=kept
{"tool": {"name": ...}}    stream_function_call_arguments=** DROPPED **  retrieval_config=kept
```

It matters mostly because the forcing branch is the framework's *own* hot path — structured output always sends `{"any": {}}` (`_structured_output_context.py:87`). On Vertex/GEAP mode it's a silent loss of an opt-in the user set explicitly; on the default Developer API it's odder than harmful, since genai rejects that field there — so the identical model config raises for `None`/`auto` and succeeds for `any`/`tool` purely because the field got dropped.

```suggestion
                existing_config = genai.types.ToolConfig.model_validate(existing)
                forced_config = tool_config.function_calling_config
                existing_calling_config = existing_config.function_calling_config
                if forced_config is not None and existing_calling_config is not None:
                    # Keep the fields a ToolChoice cannot express (e.g. stream_function_call_arguments).
                    forced_config = existing_calling_config.model_copy(
                        update={
                            "mode": forced_config.mode,
                            "allowed_function_names": forced_config.allowed_function_names,
                        }
                    )
                config_params["tool_config"] = existing_config.model_copy(
                    update={"function_calling_config": forced_config}
                )
```

I applied and ran exactly this: **67/67 pass unchanged**, `ruff` clean, `mypy` still at the 1 pre-existing error, `stream_function_call_arguments` survives all four choices, and `{"any": {}}` still correctly *replaces* the allowlist (explicitly overwriting `allowed_function_names` with `None` preserves the documented replacement semantics your new parametrized test pins).

Coverage note either way: the suite is **67/67 green with both behaviours**, so nothing currently pins this. Whichever you choose, one parametrized case in `test_stream_request_tool_config_param_precedence` should assert it — add `stream_function_call_arguments=True` to the `tool_config_param_model` fixture and the expectation follows.

strandly-the-agent (approved):
**Approving `29ad316b`.** All three findings from my previous pass are genuinely fixed — I re-ran the original repros against the new HEAD rather than taking the bot's word for it. Still a self-review (I wrote this branch), so the evidence below is the point, not my sign-off.

One new 🟡 the fix left behind, and it's **not** why I'm approving anyway: the forcing branch replaces the *whole* `function_calling_config`, so a user's `stream_function_call_arguments` is still dropped on `any`/`tool` — the same species as the bug we just fixed, one level deeper. ~10-line fix, verified green, inline. I'm not blocking on it: it's a niche field on a narrow path, everything I actually asked for landed, and blocking twice on progressively smaller instances of one issue would be disproportionate. Your call whether it rides along or becomes a follow-up.

<details>
<summary>Verification ledger — what I re-ran, plus carry-overs and what I suppressed</summary>

### Re-verified myself on `29ad316b` (delta `a720d2c8..HEAD` only, 21 src lines)
| | |
|---|---|
| ✅ Tests | **67/67** `test_gemini.py`; `ruff check` + `ruff format --check` clean; `mypy` back to **1** error, the pre-existing `count_tokens` one also present on base `2b84dfa5` |
| ✅ auto-clobber **fixed** | `None` and `{"auto":{}}` now both leave the params config fully intact (`mode: ANY, allowed_function_names: [safe_tool]`); `{"any":{}}` narrows to `mode: ANY`; `{"tool":{name}}` sets the allowlist — and `retrieval_config` survives all four |
| ✅ no leak | `params["tool_config"]` byte-identical **and same object id** after forcing calls. 60 concurrent interleaved calls on one model → exactly **2** distinct wire shapes (40 preserved / 20 forced), zero cross-talk, config pristine |
| ✅ dict-shaped params | `params={"tool_config": {...}}` (a plain dict, not a `ToolConfig`) goes through the new `model_validate` path correctly and the user's dict is **not** mutated |
| ✅ integ set-comparison | correct relaxation, not a weakened assertion — still fails for the wrong tool, a second different tool, or no tool at all |
| ✅ docstrings | all three now state the accurate *conditional* rule; my earlier accuracy complaint is resolved |
| ✅ new tests have teeth | mutation battery on the new logic: 5/7 killed. Both survivors are non-defects — `if not existing:` is semantically inert, and dropping the `tool_choice is not None` guard is caught by `mypy`, not the suite |
| ❓ Not verified | the 2 `tests_integ` cases — still no `GOOGLE_API_KEY`, still never executed by anyone |

### Carry-overs for the human owner (unchanged, not blockers)
- CI needs to actually exercise the two integ tests.
- The `api/needs-review` label still isn't applied.
- Someone needs to own the "reviewed every line" checkbox — this branch is agent-authored.

### ❓ Non-blocking design question
`auto` can no longer *relax* a `params` tool config — a caller asking for `{"auto": {}}` against `params={"tool_config": ToolConfig(mode=NONE)}` gets `NONE`, i.e. no tool is callable despite asking for auto. Bedrock would honour the `auto` (`bedrock.py:317` lets the per-request choice win unconditionally). The hybrid is defensible — "per-request always wins" is exactly what destroyed `retrieval_config` — and the forcing predicate is character-identical to `bedrock.py:379`. Worth one docstring line saying `auto` can't be used to relax a params config?

### Suppressed (attacked, survived — recorded so nobody re-runs them)
- **Aliasing is real but unexploitable**: `ToolConfig.model_validate(x)` does return the same object and `model_copy` is shallow, so the copy shares `retrieval_config` with the user's object — but `_format_request` calls `.to_json_dict()` before anything reaches the client, so no live object escapes.
- **Garbage/partially-valid `params["tool_config"]`**: error parity holds across 9 shapes × 3 choices — nothing that used to work now raises, nothing that used to raise now silently passes.
- `params={"tool_config": None}` and an empty `ToolConfig()` both take a sane branch.
</details>

strands-py/src/strands/models/gemini.py
opieter-aws: Let's keep Gemini consistent with the other providers instead of making it the more robust exception. A tool config set in `params` should wins, a per-request `tool_choice` applies when there isn't one, matching TS and the other model providers. There is a limitation with setting a `tool_config` in `params`, where the framework's structured-output retry can't override it, so a model that answers in plain text can throw a `StructuredOutputException`. That's in parity though, and if we want to fix that we should do that separately for all of them.

strandly-the-agent:
Implemented the precedence decision in `9507ef4d` (replying here rather than in the thread — thread replies are blocked for me).

`_format_request_config` is now a single `config_params.setdefault("tool_config", tool_config)`: a per-request `tool_choice` applies only when `params` sets no tool config, and an explicit one passes through whole rather than being merged with or replaced by the narrower choice. That also retires the `stream_function_call_arguments` finding without a special case — nothing is field-scoped any more. The accepted consequence is now uniform: with a params tool config the forced structured-output retry cannot override it, so `StructuredOutputException` stays reachable exactly as on the other providers.

I checked parity empirically rather than assuming it: `setdefault` decides on key presence, which is observationally identical to the spread-last ordering in `anthropic.py:258`, `openai.py:521`, `bedrock.py:355` and `Object.assign` in `strands-ts/src/models/google/model.ts:344` — including when the value is an explicit `None`.

<details>
<summary>Behaviour, evidence, and what still needs a human</summary>

```
WITH a params tool_config (untouched every time)
  None / auto / any / {tool: X}   -> mode=NONE, allowed=[safe_tool], retrieval_config kept
WITHOUT one (#1129's actual fix, still working)
  None -> no tool_config | auto -> AUTO | any -> ANY | {tool: X} -> ANY + [X]
```

Tests pin the contract from three sides: `test_stream_request_tool_config_param_takes_precedence` (parametrized over all four choices), `test_stream_tool_config_param_set_to_none_still_takes_precedence` (params owns the key), and `test_stream_tool_choice_does_not_persist_into_the_next_request` (a choice never lands in the model's own params).

- 68/68 `test_gemini.py` on `google-genai` 2.15.0 **and** on the declared floor 1.67.0; 1080 passed across `tests/strands/models` + `event_loop` + `tools`; `ruff check` / `format --check` clean; `mypy` still at the single pre-existing `count_tokens` error that is also present on base.
- Mutation-checked: reverting to an override, dropping the `dict(params or {})` copy, removing the `if tool_specs` gate, or switching the key check to truthiness each fail a test.
- Docstrings: the precedence rule is now stated once, on the public `stream()`, and the private helpers keep the bare one-liner — which matches `anthropic.py`/`openai.py`, where no precedence is documented at all. So the earlier request to mirror the fuller wording into `_format_request` is deliberately **not** implemented; a previous automated reply claimed that thread was addressed, which wasn't accurate.

Still needs a human before merge, unchanged from my earlier pass: the two `tests_integ` cases have never actually run (no `GOOGLE_API_KEY` in my sandbox) so CI needs to exercise them; the `api/needs-review` label isn't applied; someone needs to own the "reviewed every line" checkbox on an agent-authored branch; and **the PR description still documents the old override behaviour** — I can't edit it, so it needs an author edit before this merges.
</details>

lizradway (approved):

PR #3554, opened 2026-07-29
feat(mcp): add client OAuth authentication for streamable HTTP

### LINKED ISSUE #3155: [PORT] MCP client OAuth authentication (TS→Py)
> ⚠️ Filed by Strandly as part of the #2498 parity audit. Source feature verified merged + unit-tested in `strands-ts`. Agent-produced — **note the M2M-shortcut blocker below**. (No `port` label yet — a maintainer should add it.)

**Parent:** #2498

## Source SDK
`strands-ts`

## Target SDK
`strands-py`

## Related Files
```
strands-ts/src/mcp/client.ts
strands-ts/src/mcp/config.ts
strands-ts/src/mcp/config.node.ts
strands-ts/src/mcp/index.ts
strands-ts/src/mcp/__tests__/config.test.node.ts
```

## What needs to be migrated
TS `McpClient` supports two mutually-exclusive OAuth paths:
- **`McpClientCredentials`** (`client.ts:71`): `{clientId, clientSecret, scopes?}` — machine-to-machine (client_credentials grant).
- **`auth?: McpClientCredentials`** (`client.ts:74`): shortcut — builds `ClientCredentialsProvider` from `@modelcontextprotocol/sdk/client/auth-extensions.js` (`client.ts:213`).
- **`authProvider?: OAuthClientProvider`** (`client.ts:77`): advanced path, passed straight to `StreamableHTTPClientTransport`.
- **Invariants to preserve:** `auth`/`authProvider`/`headers` require `url` (not a pre-built transport); `auth` XOR `authProvider`; OAuth is **streamable-http only** (SSE + `auth` throws). Config-file path: `McpServerConfig.auth` with `${VAR}` interpolation (`config.ts:26`, `config.node.ts`).

## Python landing points
`strands-py/src/strands/tools/mcp/mcp_client.py`. **Structural mismatch:** Python `MCPClient.__init__` takes only `transport_callable` and doesn't build the transport; TS builds it internally from `url`/`auth`. Options: (a) add `url`/`auth`/`auth_provider` params + build internally (TS-like ergonomics), or (b) keep `transport_callable` and treat OAuth as a config-file feature only. Thread `auth=` into `_config_transport_callable` (`:1468`, streamable-http branch, currently no `auth`). Add `auth` to the `MCPServerConfig` TypedDict (`:87`) — note `load_servers` rejects unknown keys (`:187`), so extend the schema first. New credentials type home: `mcp_types.py`.

## Gotchas / open questions (need maintainer input)
- **🚩 The M2M `auth` shortcut has NO Python-SDK equivalent.** `mcp==1.13.0` ships `OAuthClientProvider(httpx.Auth)` but it only supports **`authorization_code` + `refresh_token`** (requires interactive `redirect_handler`/`callback_handler`/`TokenStorage`); grant support is hard-coded to those two. There is **no `client_credentials` provider in Python `mcp`**. Faithful port of `McpClientCredentials` requires **writing a custom `httpx.Auth`** that does the client_credentials token exchange, OR dropping the M2M shortcut and exposing only `auth_provider: httpx.Auth`. **Biggest decision.**
- **`authProvider` maps cleanly:** `streamablehttp_client(..., auth: httpx.Auth|None)` accepts it (`mcp/client/streamable_http.py:450`), so `authProvider` → `auth=<httpx.Auth>`.
- `scopes` → `OAuthClientMetadata.scope` (space-joined, consistent with TS `scopes.join(' ')`). Preserve streamable-http-only + XOR invariants; SSE rejects `auth`.

- Merged: ✅  ·  Unit tests: ✅ (`client.test.ts:1271-1355` — provider construction, scope join, passthrough, all 3 error paths; `config.test.node.ts`)  ·  Integ: ❌ none exercise auth.
- Docs: ❌ OAuth/`authProvider`/`McpClientCredentials` absent from all docs (`mcp-tools.mdx` shows only static header/Bearer auth). A docs section should be authored as part of the port.

## Confirmation
- [x] The change is complete in the source SDK (merged + unit-tested); every related artifact is listed above. Doc + M2M caveats noted above.

poshinchen:
## Description

The TypeScript client accepts a server `url` with either machine-to-machine credentials or a custom auth provider, while Python only accepts a pre-built `transport_callable`, so users of authed servers had to assemble the transport and the OAuth flow themselves. This ports the TS feature to Python (#3155).

`mcp>=1.23.0` ships `ClientCredentialsOAuthProvider` in `mcp.client.auth.extensions.client_credentials`. The `auth` shortcut wraps that official provider with a process-lifetime in-memory `TokenStorage`.

### Public API Changes

`MCPClient` can now build its streamable HTTP transport internally from a `url`, matching the TS constructor:

```python
from strands.tools.mcp import MCPClient

# machine-to-machine (client_credentials grant)
client = MCPClient(
    url="https://api.example.com/mcp/",
    auth={"client_id": "...", "client_secret": "...", "scopes": ["mcp:tools"]},
)

# advanced flows: any httpx.Auth, e.g. mcp's interactive OAuthClientProvider
client = MCPClient(url="https://api.example.com/mcp/", auth_provider=my_httpx_auth)
```

Complete constructor signature for API review (new parameters marked, all defaults shown):

```python
MCPClient(
    transport_callable: Callable[[], MCPTransport] | None = None,
    *,
    url: str | None = None,                    # new
    headers: dict[str, str] | None = None,     # new
    auth: MCPClientCredentials | None = None,  # new
    auth_provider: httpx.Auth | None = None,   # new
    startup_timeout: int = 30,
    tool_filters: ToolFilters | None = None,
    prefix: str | None = None,
    application_name: str | None = None,
    application_version: str | None = None,
    continue_on_error: bool = False,
    elicitation_callback: ElicitationFnT | None = None,
    progress_callback: ProgressFnT | None = None,
    tasks_config: TasksConfig | None = None,
)
```

```python
class MCPClientCredentials(TypedDict):
    client_id: str
    client_secret: str
    scopes: NotRequired[list[str]]
```

Module exports: `strands.tools.mcp.__all__` gains `MCPClientCredentials`, making the full surface `MCPAgentTool`, `MCPClient`, `MCPClientCredentials`, `MCPServerConfig`, `MCPTransport`, `TasksConfig`, `ToolFilters`.

`transport_callable` is now optional and the constructor invariants are enforced: exactly one of `transport_callable` and `url`; `headers`, `auth`, and `auth_provider` require `url`; `auth` and `auth_provider` are mutually exclusive; `url` must be an http:// or https:// URL. `auth` values are validated (string credentials, `scopes` as a list of strings) with error messages that name keys but never values, so credentials cannot leak into logs.

`load_servers` configs gain a matching `auth` key (streamable-http only, with the existing `${VAR}` interpolation so secrets stay in the environment):

```json
{
  "mcpServers": {
    "protected": {
      "url": "https://api.example.com/mcp/",
      "auth": {"client_id": "${OAUTH_CLIENT_ID}", "client_secret": "${OAUTH_CLIENT_SECRET}"}
    }
  }
}
```

### Callouts

- `scopes` is advisory. The `mcp` provider applies the MCP scope-selection strategy, so scopes advertised by the server (via the `WWW-Authenticate` header or its protected-resource / authorization-server metadata) replace the configured value before the token request is built. The docstring and docs state this. Honoring the configured scope for the M2M grant, as the TS SDK does, remains open in #3155.
- A present-but-empty or non-object config `auth` entry (for example `{}` or `[]`) is rejected instead of silently building an unauthenticated client.
- The config `auth` keys are `client_id`/`client_secret` (snake_case, consistent with the rest of `MCPServerConfig`), while the TS config uses `clientId`/`clientSecret`. This follows the config file's existing casing convention, which already differs from TS (`continue_on_error`, `tool_filters`).
- Existing positional `MCPClient(transport_callable)` usage is unaffected; making the parameter optional is backward compatible.

## Related Issues

Addresses #3155. The `scopes` parity gap with the TS SDK stays open there (see Callouts).

## Documentation PR

Included under `site/`: a new "OAuth Authentication" section on the MCP tools page covering `auth`, `auth_provider`, and the config-file shape.

## Type of Change

New feature

## Testing

## Checklist

----

### DIFF
--- a/site/src/content/docs/user-guide/concepts/tools/mcp-tools.mdx
+++ b/site/src/content/docs/user-guide/concepts/tools/mcp-tools.mdx
@@ -186,6 +186,53 @@ github_mcp_client = MCPClient(
 )
 ```
 
+#### OAuth Authentication
+
+For servers protected by OAuth, pass the server `url` directly with an `auth` credential. `MCPClient` builds the streamable HTTP transport internally and authenticates with the OAuth client_credentials grant, which suits machine-to-machine deployments where no user is present to complete a browser flow.
+
+```python
+import os
+from strands.tools.mcp import MCPClient
+
+oauth_mcp_client = MCPClient(
+    url="https://api.example.com/mcp/",
+    auth={
+        "client_id": os.environ["OAUTH_CLIENT_ID"],
+        "client_secret": os.environ["OAUTH_CLIENT_SECRET"],
+        "scopes": ["mcp:tools"],  # optional
+    },
+)
+```
+
+Scopes are joined with spaces before being sent to the token endpoint. Token acquisition and refresh happen automatically. The `headers` parameter can be combined with `auth` when the server also expects custom headers.
+
+For advanced flows such as the interactive authorization_code grant, pass any `httpx.Auth` implementation as `auth_provider` instead. The `mcp` package's `OAuthClientProvider` is one such implementation:
+
+```python
+from mcp.client.auth import OAuthClientProvider
+
+oauth_mcp_client = MCPClient(
+    url="https://api.example.com/mcp/",
+    auth_provider=OAuthClientProvider(...),
+)
+```
+
+`auth` and `auth_provider` are mutually exclusive, both require `url`, and OAuth is supported for streamable HTTP only. The same `auth` credential can be set on a server entry in a [`load_servers`](#using-multiple-mcp-servers) config, where `${VAR}` interpolation keeps secrets in the environment:
+
+```json
+{
+  "mcpServers": {
+    "protected-server": {
+      "url": "https://api.example.com/mcp/",
+      "auth": {
+        "client_id": "${OAUTH_CLIENT_ID}",
+        "client_secret": "${OAUTH_CLIENT_SECRET}"
+      }
+    }
+  }
+}
+```
+
 #### AWS IAM 
 
 For MCP servers on AWS that use SigV4 authentication with IAM credentials, you can conveniently use the [`mcp-proxy-for-aws`](https://pypi.org/project/mcp-proxy-for-aws/) package to handle AWS credential management and request signing automatically. See the [detailed guide](https://dev.to/aws/no-oauth-required-an-mcp-client-for-aws-iam-k1o) for more information.
--- a/site/src/content/docs/user-guide/concepts/tools/mcp-tools.mdx
+++ b/site/src/content/docs/user-guide/concepts/tools/mcp-tools.mdx
@@ -204,7 +204,7 @@ oauth_mcp_client = MCPClient(
 )
 ```
 
-Scopes are joined with spaces before being sent to the token endpoint. Token acquisition and refresh happen automatically. The `headers` parameter can be combined with `auth` when the server also expects custom headers.
+`scopes` is joined with spaces and is advisory only: if the server advertises its own scopes (via the `WWW-Authenticate` header or its protected-resource / authorization-server metadata), the server's scopes are used instead and this value is ignored. Token acquisition and refresh happen automatically. The `headers` parameter can be combined with `auth` when the server also expects custom headers.
 
 For advanced flows such as the interactive authorization_code grant, pass any `httpx.Auth` implementation as `auth_provider` instead. The `mcp` package's `OAuthClientProvider` is one such implementation:
 
--- a/strands-py/src/strands/tools/mcp/mcp_client.py
+++ b/strands-py/src/strands/tools/mcp/mcp_client.py
@@ -26,6 +26,7 @@ from pathlib import Path
 from re import Pattern
 from types import TracebackType
 from typing import Any, TypeVar, cast
+from urllib.parse import urlparse
 
 import anyio
 import httpx
@@ -280,8 +281,9 @@ class MCPClient(ToolProvider):
 
         Raises:
             ValueError: If neither or both of `transport_callable` and `url` are provided, if
-                `headers`, `auth`, or `auth_provider` is provided without `url`, or if both `auth`
-                and `auth_provider` are provided.
+                `headers`, `auth`, or `auth_provider` is provided without `url`, if both `auth`
+                and `auth_provider` are provided, if `url` is not an http:// or https:// URL, or
+                if `auth` is missing required keys or a value has the wrong type.
         """
         self._startup_timeout = startup_timeout
         self._tool_filters = tool_filters
@@ -1696,13 +1698,28 @@ class _InMemoryTokenStorage:
 
 
 def _build_client_credentials_auth(url: str, auth: MCPClientCredentials) -> httpx.Auth:
-    """Build an httpx.Auth performing the OAuth client_credentials grant against the server."""
-    scopes = auth.get("scopes")
+    """Build an httpx.Auth performing the OAuth client_credentials grant against the server.
+
+    Error messages name only the offending keys, never their values, so credentials cannot
+    leak into logs.
+
+    Raises:
+        ValueError: If `auth` is missing required keys or a value has the wrong type.
+    """
+    values = cast(dict[str, Any], auth)
+    missing_keys = sorted({"client_id", "client_secret"} - values.keys())
+    if missing_keys:
+        raise ValueError(f"MCPClient: 'auth' is missing required keys: {missing_keys}")
+    if not isinstance(values["client_id"], str) or not isinstance(values["client_secret"], str):
+        raise ValueError("MCPClient: 'auth' requires 'client_id' and 'client_secret' to be strings")
+    scopes = values.get("scopes")
+    if scopes is not None and not (isinstance(scopes, list) and all(isinstance(scope, str) for scope in scopes)):
+        raise ValueError("MCPClient: 'auth' requires 'scopes' to be a list of strings")
     return ClientCredentialsOAuthProvider(
         server_url=url,
         storage=_InMemoryTokenStorage(),
-        client_id=auth["client_id"],
-        client_secret=auth["client_secret"],
+        client_id=values["client_id"],
+        client_secret=values["client_secret"],
         scopes=" ".join(scopes) if scopes else None,
     )
 
@@ -1719,11 +1736,11 @@ def _resolve_transport_callable(
 
     Enforces the constructor invariants: exactly one of `transport_callable` and `url`;
     `headers`, `auth`, and `auth_provider` require `url`; `auth` and `auth_provider` are
-    mutually exclusive.
+    mutually exclusive; `url` must be an http:// or https:// URL.
     """
-    if transport_callable is not None and url is not None:
+    if transport_callable is not None and url:
         raise ValueError("MCPClient: provide either 'transport_callable' or 'url', not both")
-    if transport_callable is None and url is None:
+    if transport_callable is None and not url:
         raise ValueError("MCPClient: either 'transport_callable' or 'url' must be provided")
     if transport_callable is not None:
         if auth is not None or auth_provider is not None or headers is not None:
@@ -1736,6 +1753,11 @@ def _resolve_transport_callable(
         raise ValueError("MCPClient: provide either 'auth' or 'auth_provider', not both")
 
     server_url = cast(str, url)
+    scheme = urlparse(server_url).scheme
+    if scheme not in ("http", "https"):
+        raise ValueError(f"MCPClient: 'url' must be an http:// or https:// URL, got '{server_url}'")
+    if scheme == "http" and (auth is not None or auth_provider is not None):
+        logger.warning("url=<%s> | sending oauth credentials over plaintext http", server_url)
     resolved_auth = _build_client_credentials_auth(server_url, auth) if auth is not None else auth_provider
     return lambda: streamablehttp_client(url=server_url, headers=headers, auth=resolved_auth)
 
@@ -1804,7 +1826,7 @@ def _config_transport_callable(name: str, transport: str, server: dict[str, Any]
     """Return a zero-arg callable that opens the transport for the given server entry."""
     if transport != "streamable-http" and server.get("auth"):
         raise ValueError(
-            f"server '{name}': 'auth' is only supported for streamable-http transport — "
+            f"server '{name}': Strands supports 'auth' for streamable-http transport only — "
             "use streamable-http or provide a pre-configured transport"
         )
 
@@ -1842,7 +1864,7 @@ def _config_transport_callable(name: str, transport: str, server: dict[str, Any]
 
 def _parse_config_auth(name: str, url: str, auth: dict[str, Any] | None) -> httpx.Auth | None:
     """Validate a config 'auth' entry and build the client_credentials httpx.Auth from it."""
-    if not auth:
+    if auth is None:
         return None
     if not isinstance(auth, dict):
         raise ValueError(f"server '{name}': 'auth' must be an object with 'client_id' and 'client_secret'")
--- a/strands-py/src/strands/tools/mcp/mcp_client.py
+++ b/strands-py/src/strands/tools/mcp/mcp_client.py
@@ -28,10 +28,13 @@ from types import TracebackType
 from typing import Any, TypeVar, cast
 
 import anyio
+import httpx
 from mcp import ClientSession, ListToolsResult, StdioServerParameters, stdio_client
+from mcp.client.auth.extensions.client_credentials import ClientCredentialsOAuthProvider
 from mcp.client.session import ElicitationFnT
 from mcp.client.sse import sse_client
 from mcp.client.streamable_http import streamablehttp_client
+from mcp.shared.auth import OAuthClientInformationFull, OAuthToken
 from mcp.shared.exceptions import McpError
 from mcp.shared.session import ProgressFnT
 from mcp.types import (
@@ -63,7 +66,7 @@ from ..tool_provider import ToolProvider
 from .mcp_agent_tool import MCPAgentTool
 from .mcp_instrumentation import inject_trace_context, mcp_instrumentation
 from .mcp_tasks import DEFAULT_TASK_CONFIG, DEFAULT_TASK_POLL_TIMEOUT, DEFAULT_TASK_TTL, TasksConfig
-from .mcp_types import MCPToolResult, MCPTransport
+from .mcp_types import MCPClientCredentials, MCPToolResult, MCPTransport
 
 logger = logging.getLogger(__name__)
 
@@ -107,6 +110,9 @@ class MCPServerConfig(TypedDict, total=False):
     omitted it is auto-detected from the fields present. String values support '${VAR}' /
     '${env:VAR}' interpolation, and '~' in 'command' and 'cwd' is expanded to the home directory.
 
+    'auth' holds client credentials for OAuth machine-to-machine (client_credentials grant)
+    authentication and is only supported for streamable-http transport.
+
     'disabled' skips the server entirely. 'continue_on_error' keeps the rest of the servers usable
     when this one fails: a config-resolution failure (e.g. a missing env var) skips it during
     load_servers instead of raising, and a connection failure yields no tools instead of raising
@@ -120,6 +126,7 @@ class MCPServerConfig(TypedDict, total=False):
     url: str
     headers: dict[str, str]
     transport: str
+    auth: MCPClientCredentials
     disabled: bool
     continue_on_error: bool
     prefix: str
@@ -222,8 +229,12 @@ class MCPClient(ToolProvider):
 
     def __init__(
         self,
-        transport_callable: Callable[[], MCPTransport],
+        transport_callable: Callable[[], MCPTransport] | None = None,
         *,
+        url: str | None = None,
+        headers: dict[str, str] | None = None,
+        auth: MCPClientCredentials | None = None,
+        auth_provider: httpx.Auth | None = None,
         startup_timeout: int = 30,
         tool_filters: ToolFilters | None = None,
         prefix: str | None = None,
@@ -238,6 +249,14 @@ class MCPClient(ToolProvider):
 
         Args:
             transport_callable: A callable that returns an MCPTransport (read_stream, write_stream) tuple.
+                Mutually exclusive with `url`.
+            url: Server URL. When provided, a streamable HTTP transport is constructed automatically.
+                Mutually exclusive with `transport_callable`.
+            headers: HTTP headers to include on every request to the server. Requires `url`.
+            auth: Client credentials for OAuth machine-to-machine (client_credentials grant)
+                authentication. Requires `url`. Mutually exclusive with `auth_provider`.
+            auth_provider: Custom `httpx.Auth` for advanced auth flows, passed through to the
+                streamable HTTP transport. Requires `url`. Mutually exclusive with `auth`.
             startup_timeout: Timeout after which MCP server initialization should be cancelled.
                 Defaults to 30.
             tool_filters: Optional filters to apply to tools.
@@ -247,17 +266,22 @@ class MCPClient(ToolProvider):
                 Defaults to None (uses the MCP SDK default "mcp").
             application_version: Optional version string to report alongside application_name.
                 Defaults to None (uses the Strands SDK version).
-            continue_on_error: When True, a connection failure during ``load_tools`` is logged and
+            continue_on_error: When True, a connection failure during `load_tools` is logged and
                 yields no tools instead of raising, so one unavailable server does not prevent an
-                agent from using the others. Only the connection (``start()``) is swallowed; an error
+                agent from using the others. Only the connection (`start()`) is swallowed; an error
                 while listing tools after a successful connect still propagates. Defaults to False.
             elicitation_callback: Optional callback function to handle elicitation requests from the MCP server.
             progress_callback: Optional callback to receive progress notifications during tool execution.
-                Called with ``(progress, total, message)`` as the server reports progress. The ``total``
-                and ``message`` parameters may be ``None`` if the server does not provide them.
+                Called with `(progress, total, message)` as the server reports progress. The `total`
+                and `message` parameters may be `None` if the server does not provide them.
             tasks_config: Configuration for MCP task-augmented execution for long-running tools.
                 If provided (not None), enables task-augmented execution for tools that support it.
                 See TasksConfig for details. This feature is experimental and subject to change.
+
+        Raises:
+            ValueError: If neither or both of `transport_callable` and `url` are provided, if
+                `headers`, `auth`, or `auth_provider` is provided without `url`, or if both `auth`
+                and `auth_provider` are provided.
         """
         self._startup_timeout = startup_timeout
         self._tool_filters = tool_filters
@@ -279,7 +303,9 @@ class MCPClient(ToolProvider):
         self._close_future: asyncio.futures.Future[None] | None = None
         self._close_exception: None | Exception = None
         # Do not want to block other threads while close event is false
-        self._transport_callable = transport_callable
+        self._transport_callable = _resolve_transport_callable(
+            transport_callable, url=url, headers=headers, auth=auth, auth_provider=auth_provider
+        )
 
         self._background_thread: threading.Thread | None = None
         self._background_thread_session: ClientSession | None = None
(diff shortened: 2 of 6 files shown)


poshinchen:
@strandly-the-agent take a look at the PR and see if any improvement is required. You can tag "strands-agent" if you need any assessment.

strands-py/src/strands/tools/mcp/mcp_client.py
```diff
@@ -1632,6 +1658,75 @@ async def _poll_until_terminal() -> GetTaskResult | None:
...
+
+def _build_client_credentials_auth(url: str, auth: MCPClientCredentials) -> httpx.Auth:
+    """Build an httpx.Auth performing the OAuth client_credentials grant against the server."""
+    scopes = auth.get("scopes")
+    return ClientCredentialsOAuthProvider(
+        server_url=url,
+        storage=_InMemoryTokenStorage(),
+        client_id=auth["client_id"],
+        client_secret=auth["client_secret"],
+        scopes=" ".join(scopes) if scopes else None,
```
strandly-the-agent, line 1723: 🔴 **The `scopes` value never reaches the token endpoint — this line has no observable effect.**

`scopes=` here lands on `provider._fixed_client_info.scope`, and `_initialize()` copies that to `context.client_info` (client *registration* data). But the token request is built by `_exchange_token_client_credentials`, which reads `self.context.client_metadata.scope` — a different field, which `OAuthClientProvider.async_auth_flow` **unconditionally overwrites** at Step 3 before Step 5 ever builds the request:

```python
# mcp/client/auth/oauth2.py:564-570  (mcp 1.29.0)
# Step 3: Apply scope selection strategy
self.context.client_metadata.scope = get_client_metadata_scopes(
    www_authenticate_scope, prm, asm
)
...
# Step 5: Perform authorization and complete token exchange
token_response = yield await self._perform_authorization()
```

`get_client_metadata_scopes` resolves `WWW-Authenticate` scope → PRM `scopes_supported` → ASM `scopes_supported` → `None`. The caller's value is **never an input**. And `ClientCredentialsOAuthProvider` doesn't override `async_auth_flow` — I checked: it overrides only `_exchange_token_client_credentials`, `_initialize`, `_perform_authorization`. Every fresh connect is 401-driven, so Step 3 always runs.

<details>
<summary>Observed against a real fake AS + Bearer-gated MCP server, plus a mutation test</summary>

```
A. user asks scopes=['read'], server advertises ['read','write','admin']
   connect OK; scope actually sent to /token = 'read write admin'   <- MORE than requested
B. server requires exactly 'read'; user correctly asks ['read']; server advertises ['read','write']
   connect FAILED; scope sent = 'read write'
   server replied {"error":"invalid_scope","error_description":"got 'read write'"}
C. server advertises scopes nowhere; user asks ['read','write']
   scope sent = None                                               <- user value discarded entirely
```

Mutation test — gutting the plumbing changes nothing on the wire:
```
PR as written          -> scope on wire = 'read write'
scopes plumbing GUTTED -> scope on wire = 'read write'
IDENTICAL: True
```
</details>

**This isn't your bug** — it's inherited from wrapping `mcp`'s provider as-is. Two notes on why it still matters here:

1. TS *doesn't* have it. `@modelcontextprotocol/sdk`'s `auth.ts:558` computes `nonInteractiveFlow = !provider.redirectUrl` and routes that path to `fetchToken`, which uses `provider.clientMetadata.scope` unmodified (`auth.ts:1397`); the SEP-835 scope override is applied only to the interactive branch. So TS honors the operator's configured scope for M2M and this port silently can't — and #3155 named this as a parity target ("`scopes` → `OAuthClientMetadata.scope`... consistent with TS `scopes.join(' ')`").
2. The user-visible harm is least-privilege: the client can request *more* scope than asked for, and a correct config can fail to connect.

**In scope for this PR:** correct the three places that assert this works (this comment, `mcp_types.py:25`, `mcp-tools.mdx:207`, and the test below), and change `Closes #3155` → `Addresses #3155` so the issue stays open. **Optional / fast-follow:** subclass `ClientCredentialsOAuthProvider` to re-pin `client_metadata.scope` after discovery (or file it upstream). Your call on whether that lands here.

strands-py/src/strands/tools/mcp/mcp_client.py
```diff
@@ -1632,6 +1658,75 @@ async def _poll_until_terminal() -> GetTaskResult | None:
...
+    async def get_client_info(self) -> OAuthClientInformationFull | None:
+        return self._client_info
+
+    async def set_client_info(self, client_info: OAuthClientInformationFull) -> None:
+        self._client_info = client_info
+
+
+def _build_client_credentials_auth(url: str, auth: MCPClientCredentials) -> httpx.Auth:
+    """Build an httpx.Auth performing the OAuth client_credentials grant against the server."""
+    scopes = auth.get("scopes")
```
strandly-the-agent: 🟡 **Two type-validation gaps in this helper, one fix.** Both entry points route through here (`_parse_config_auth` at `:1847`), so a single block covers programmatic *and* config paths.

**a) `scopes` as a `str` silently char-splits.** `MCPClientCredentials` types it `NotRequired[list[str]]` but nothing enforces that at runtime, and a `str` is truthy *and* iterable:

```
scopes='read write'          -> client_metadata.scope = 'r e a d   w r i t e'
scopes='mcp:tools'           -> 'm c p : t o o l s'
scopes={'read':1,'write':2}  -> 'read write'   (silently yields dict keys)
```

A JSON config with `"scopes": "read write"` instead of `["read","write"]` is a natural mistake — the field's whole purpose *is* a space-separated scope string. It fails silently at construction and only surfaces as an opaque `invalid_scope` at connect. Currently masked by the `scopes` bug above; goes live the moment that's fixed. Contrast `client_id`/`client_secret`, which reach pydantic and get a loud error.

**b) A non-string `client_secret` puts the secret's *value* into a WARNING log line.** `auth["client_secret"]` goes straight into `OAuthClientInformationFull(client_secret=...)`, and pydantic embeds the input in its message:

```
ValidationError: 1 validation error for OAuthClientInformationFull
client_secret
  Input should be a valid string [type=string_type, input_value=987654321, input_type=int]
```

With `"continue_on_error": true`, `mcp_client.py:225` catches exactly this and does `logger.warning(..., error=<%s> ..., e)` — so an unquoted secret in JSON lands verbatim in the app's log stream. It's the operator's own secret in their own logs, but logs get shipped and retained differently than config files. Notably every *other* new error message in this PR is careful to name only keys.

```suggestion
    scopes = auth.get("scopes")
    if not isinstance(auth["client_id"], str) or not isinstance(auth["client_secret"], str):
        raise ValueError("MCPClient: 'auth' requires 'client_id' and 'client_secret' to be strings")
    if scopes is not None and (not isinstance(scopes, list) or not all(isinstance(s, str) for s in scopes)):
        raise ValueError("MCPClient: 'auth[\"scopes\"]' must be a list of strings")
```

(Naming only the field, never the value, keeps the log-hygiene property.) A frozen dataclass for the credentials type would give you all of this for free — see question 1 in the summary.

strands-py/src/strands/tools/mcp/mcp_client.py
```diff
@@ -1632,6 +1658,75 @@ async def _poll_until_terminal() -> GetTaskResult | None:
...
+    auth: MCPClientCredentials | None,
+    auth_provider: httpx.Auth | None,
+) -> Callable[[], MCPTransport]:
+    """Resolve the MCPClient constructor arguments into a single transport callable.
+
+    Enforces the constructor invariants: exactly one of `transport_callable` and `url`;
+    `headers`, `auth`, and `auth_provider` require `url`; `auth` and `auth_provider` are
+    mutually exclusive.
+    """
+    if transport_callable is not None and url is not None:
```
strandly-the-agent: 🟡 **`url` isn't validated on this path, so `url=""` constructs silently — and `http://` + `auth` ships the client secret in cleartext.**

These two checks gate on `is None`, so a falsy-but-present `url` slips through:

```
OK (no raise)  MCPClient(url='')
OK (no raise)  MCPClient(url='not-a-url', auth=CREDS)
```

`streamablehttp_client(url='')` doesn't even fail at transport construction — it fails when `session.initialize()` tries to talk to `''` on the background thread, up to the full 30s `startup_timeout` later, with an unrelated low-level error instead of the clear invariant message.

Three reasons this is worth fixing rather than shrugging at:

1. **It disagrees with the sibling function in this same file.** `_config_transport_callable` (`:1820-1822`) correctly does `if not url: raise ValueError(...)` — so the identical input is rejected via `load_servers` and accepted via the constructor.
2. **TS fails fast.** `client.ts:194-196` uses `if (!args.transport && !args.url)` (JS truthiness → `""` counts as absent), and `client.ts:225` does `new URL(args.url!)`, which throws on empty/malformed. That's the substance behind TS's `string | URL` — porting the *type* would be un-idiomatic, but porting the *validation* is the faithful port.
3. **Credential-leak footgun.** The client_credentials grant POSTs `client_id`/`client_secret` to a token endpoint derived from `url`, so a typo'd scheme puts a long-lived secret on the wire in plaintext with no signal. Neither `mcp.client.streamable_http` nor `mcp.client.auth.oauth2` validates the scheme (I grepped both). Precedent for validating in a public Strands factory: `vended_plugins/skills/skill.py:367`, `vended_plugins/skills/agent_skills.py:507`.

Suggested (switching to falsiness also aligns with the sibling function and TS):

```python
if transport_callable is not None and url:
    raise ValueError("MCPClient: provide either 'transport_callable' or 'url', not both")
if transport_callable is None and not url:
    raise ValueError("MCPClient: either 'transport_callable' or 'url' must be provided")
...
scheme = urlparse(server_url).scheme
if scheme not in ("http", "https"):
    raise ValueError(f"MCPClient: 'url' must be an http:// or https:// URL, got '{server_url}'")
if scheme == "http" and (auth is not None or auth_provider is not None):
    logger.warning("url=<%s> | sending OAuth credentials over plaintext http", server_url)
```

(`cast(str, url)` at `:1725` stays necessary either way — I confirmed with a standalone `mypy --strict` snippet that mypy doesn't correlate the two disjoint checks, so it's a real workaround, not masking a bug.)

strands-py/src/strands/tools/mcp/mcp_client.py
```diff
@@ -1725,6 +1827,26 @@ def _config_transport_callable(name: str, transport: str, server: dict[str, Any]
             raise ValueError(f"server '{name}': unsupported transport type '{transport}'")
 
 
+def _parse_config_auth(name: str, url: str, auth: dict[str, Any] | None) -> httpx.Auth | None:
+    """Validate a config 'auth' entry and build the client_credentials httpx.Auth from it."""
+    if not auth:
```
strandly-the-agent: 🟡 **A falsy-but-non-dict `auth` in a config silently means *no auth* — this fails open.**

`if not auth: return None` runs before the `isinstance` check, so the type validation this function exists to provide is bypassed exactly when the config is malformed:

```
load_servers({"srv": {"url": "...", "auth": []}})  -> OK, client connects with NO AUTH
load_servers({"srv": {"url": "...", "auth": {}}})  -> OK, client connects with NO AUTH
load_servers({"srv": {"url": "...", "auth": 0}})   -> OK, no auth
load_servers({"srv": {"url": "...", "auth": ""}})  -> OK, no auth
load_servers({"srv": {"url": "...", "auth": "x"}}) -> ValueError  (truthy, correctly caught)
```

A mis-templated or half-removed key (`"auth": {}` from a config generator, `"auth": []` from a YAML→JSON round-trip) drops authentication with zero warning. Not covered by `test_auth_non_dict_raises`, which only exercises the truthy `"not-a-dict"` case.

```suggestion
    if auth is None:
        return None
```

Note this also turns `"auth": {}` from silently-unauthenticated into the existing "missing required keys" error — a deliberate behavior change, and I think the right one (it matches how a present-but-incomplete dict is already treated), but worth a line in the PR description.

strands-py/src/strands/tools/mcp/mcp_client.py
```diff
@@ -238,6 +249,14 @@ def __init__(
 
         Args:
             transport_callable: A callable that returns an MCPTransport (read_stream, write_stream) tuple.
+                Mutually exclusive with `url`.
+            url: Server URL. When provided, a streamable HTTP transport is constructed automatically.
+                Mutually exclusive with `transport_callable`.
+            headers: HTTP headers to include on every request to the server. Requires `url`.
+            auth: Client credentials for OAuth machine-to-machine (client_credentials grant)
+                authentication. Requires `url`. Mutually exclusive with `auth_provider`.
+            auth_provider: Custom `httpx.Auth` for advanced auth flows, passed through to the
```
strandly-the-agent, line 259: 🟡 **Worth documenting that an `auth_provider` instance can't be shared across `MCPClient`s** — sharing one to share a token across two clients on the same origin is the obvious use, and concurrent `start()` deadlocks permanently.

`_resolve_transport_callable` closes over one provider instance (`:1726`). `OAuthContext.lock` is an `anyio.Lock` constructed off-loop (in the provider's `__init__`, on the main thread), so its inner asyncio-bound lock is cached on **first** use. Each `MCPClient` runs its session on its own background thread with its own event loop, so a contended acquire creates the waiter's `Future` on one loop while `release()` runs on another thread — the cross-thread wakeup never wakes the waiter's selector.

<details>
<summary>Repro (deterministic; also reduces to ~10 lines of pure anyio, no MCP/network)</summary>

Two `MCPClient`s sharing one provider, started from two threads, with `/token` sleeping 1.5s to force contention:
```
thread 0 still alive after 45s join (=HUNG): True
client 1: MCPClientInitializationError: background thread did not start in 15 seconds
lock locked: True stats: LockStatistics(locked=True, owner=AsyncIOTaskInfo(name='...post_writer'), tasks_waiting=0)
```
Reduced (no MCP at all): holder releases at 2s, waiter still hung at +15s.
```
holder released the lock: True
  after 15s more -- waiter acquired: False, waiter thread alive (=hung): True, lock.locked(): True
```
**Sequential** sharing is fine — two clients started one after the other share the token correctly (1 token issued). So is 8 concurrent `call_tool_sync` on a *single* client. It's only concurrent *first* acquires across loops.
</details>

🟡 not 🔴 because the default `Agent` path loads tool providers serially (`tools/registry.py:138`), so nobody hits this by accident today. Cheapest fix is one docstring sentence ("an `auth_provider` instance is not safe to share across `MCPClient` instances"); the structural fix is to build the provider lazily *inside* the transport callable so each `start()` gets a loop-local lock.

strandly-the-agent (commented):
**TL;DR:** Clean port, but the `scopes` knob provably does nothing — and the PR ships docs, a docstring, and a *passing* test that all assert it works. The root cause is upstream in `mcp`, not your code, but correcting those three claims is cheap and in scope. Everything else below is 🟡/⚪/questions.

**Verdict:** changes requested (soft) — 1 🔴 cluster, 4 🟡, a nit bundle. The invariants, the shared `_build_client_credentials_auth` helper, `_InMemoryTokenStorage`, and the test suite are genuinely solid; this is close to done.

<details>
<summary><b>✅/🔴 What I actually verified (branch <code>fa3d8ae6</code> vs <code>f383c11a</code>)</b></summary>

- ✅ `pytest tests/strands/tools/mcp/` → **232 passed**; `ruff check` + `ruff format --check` clean; `mypy` (strict) clean on `src/strands/tools/mcp/`. Ran against `mcp 1.29.0`.
- ✅ `mcp>=1.23.0` floor is correct — `ClientCredentialsOAuthProvider` exists there; no bound change needed.
- ✅ **The grant itself works end to end.** I stood up a fake OAuth AS + a Bearer-gated real MCP streamable-HTTP server: 401 → protected-resource-metadata discovery → AS-metadata discovery → `POST /token` (Basic) → retry with Bearer → `call_tool_sync` round-trips. Token expiry re-grant works (`expires_in=1`, second call gets a fresh token). Server-side revocation recovers. 3× `start()`/`stop()` on one client works — the anyio lock is reused across three distinct event loops without error, so there is no analogue of the TS single-use-transport problem here.
- ✅ **No secret leakage in any realistic surface.** With a sentinel secret: `repr(MCPClient)`, `vars(client)`, `repr(transport_callable)` + its closure cells, an 11 KB DEBUG capture of *all* loggers across a failed handshake, and the failure traceback — all clean, for both the plaintext secret and `base64(id:secret)`. All four `${VAR}` error paths name only the variable. `mcp_instrumentation.py` never wraps `httpx.Auth` or the token exchange. (One exception → inline on `_build_client_credentials_auth`.)
- ✅ `_InMemoryTokenStorage` structurally satisfies `mcp.client.auth.oauth2.TokenStorage` (signatures + async-ness match; `mypy --strict` accepts it). `cast(str, url)` is load-bearing, not masking a narrowing mypy could do — verified with a standalone strict snippet.
- ✅ `transport_callable` becoming optional is safe: every call site in `src`, `tests`, `tests_integ`, `site/docs/examples` passes it positionally.
- ✅ **19 of 20 targeted mutations of the new code are caught by the new tests** (the one that slips through is in the nit bundle).
- ✅ Config key casing (`client_id` vs TS `clientId`) — I'd drop this from the Callouts as a concern: the config is *already* non-portable (`continue_on_error`/`tool_filters`/`startup_timeout` vs TS `continueOnError`/`tasksConfig`), so this is consistency, not new drift, and accepting both casings later is additive.
- 🔴 **Reproduced broken:** the user's `scopes` never reaches the token endpoint (mechanism confirmed in the installed `mcp` source, plus a mutation test — inline).
- 🔴 Reproduced: `MCPClient(url="")` constructs silently; `load_servers` with `"auth": []`/`{}`/`0`/`""` silently yields an **unauthenticated** client; `scopes="read write"` char-splits to `'r e a d   w r i t e'`; a non-string `client_secret` lands verbatim in a WARNING log line; the second docs snippet `TypeError`s if copy-pasted.
- ⚠️ Not re-run by me personally: the concurrent shared-`auth_provider` deadlock (inline) came from a dedicated adversarial pass with a reduced ~10-line pure-anyio repro; I verified the mechanism, not the script.

</details>

### ⚪ Nit bundle (take or leave)

- `strands-py/tests/strands/tools/mcp/test_mcp_client_auth.py:109` — `match="not both"` matches **either** of two different invariants (`mcp_client.py:1712` and `:1723`). Mutation-proven: swapping one error message for the other keeps all 14 tests green. Pin it: `match="either 'transport_callable' or 'url', not both"`.
- Duplicate test name `test_auth_builds_client_credentials_provider` in `test_mcp_client_auth.py:54` and `test_mcp_client_load_servers.py:335` — the only such collision in the repo; `pytest -k` can't distinguish them. Rename one.
- `mcp_client.py:269-272,274-276` reformats three unrelated docstring lines from ``double`` to `single` backticks. `AGENTS.md:80` is explicit ("No drive-by reformatting"), and the file was previously all-double so it's now internally inconsistent. Worth reverting those three lines.
- `mcp_client.py:1792-1796` — the message says `auth` "is only supported for streamable-http transport", but `mcp.client.sse.sse_client` *does* take `auth: httpx.Auth | None` (verified). The restriction is a defensible product call; the wording asserts a transport limitation that doesn't exist. "Strands supports `auth` for streamable-http only" would send fewer people on a goose chase.
- `mcp_client.py:1727` captures `headers` by reference with no defensive copy, so a caller mutating their dict later changes what's sent on the next `start()`. Same as TS (`client.ts:229`), so not a new divergence — just surprising.
- None of the three constructor `ValueError` texts mention `auth_provider`, so the escape hatch isn't discoverable from the failure.

### Questions

**Blocking-ish** (both are one-way doors — free now, a break after release, so worth settling in the API review rather than after):

1. `MCPClientCredentials` as a TypedDict vs a frozen dataclass — a dataclass would give you the missing-key/unknown-key validation for free *and* structurally fix the bare-`KeyError` asymmetry already flagged on this PR. I measured it: `OAuthClientCredentials(**auth)` yields `missing 1 required positional argument: 'client_secret'`, `got an unexpected keyword argument 'clientId'`, etc. `PORTING.md:11-15` maps "a data shape constructed and passed around" to a dataclass, but the local sibling `TasksConfig` is a TypedDict — so either way, could this get recorded as a decision so the next porter doesn't re-litigate it?
2. Does `MCPClientCredentials` read as "the `MCPClient`'s credentials" rather than the RFC 6749 §4.4 client-credentials-**grant** credentials it is? It sits next to `MCPClient` in the same `__all__`. `team/DECISIONS.md:168-183` explicitly licenses diverging from the other SDK's naming for within-language clarity, so `OAuthClientCredentials` is available if you want it — but it's a two-SDK call since TS exports `McpClientCredentials`.

**Non-blocking:**

3. Is `auth_provider: httpx.Auth` deliberately *wider* than TS's `OAuthClientProvider` — i.e. is a plain `httpx.BasicAuth` (which bypasses MCP's `WWW-Authenticate`/resource-metadata handling) a supported use case? `COMPATIBILITY.md:63` licenses widening only, so narrowing later would be a break. One sentence of stated intent would pin it. (`httpx` itself is fine as a public type — it's a direct core dep at `pyproject.toml:35`, and `vended_tools/http_request` already exposes `httpx.AsyncClient` publicly.)
4. Should `auth_provider`'s own `server_url` be cross-checked against `url`? `MCPClient(url="https://real", auth_provider=...(server_url="https://typo"))` is accepted silently today.
5. Would `@overload`s on `__init__` be worth adding? All five illegal argument combos currently pass `mypy --strict` and only fail at runtime. I prototyped `@overload` + `Unpack[MCPClientOptions]`: 5/5 illegal caught, 6/6 legal clean, body and callers unchanged, and it additionally catches a `startup_timout=` typo. `_middleware/README.md:145` argues for exactly this ("checked statically rather than only at runtime") and there are 13 `Unpack` precedents in the SDK. Purely additive — just flagging the gap, not asking for it here.

<details>
<summary><b>Recommended follow-ups (out of scope for this PR)</b></summary>

- **Integ test for MCP OAuth.** All 56 new tests patch `streamablehttp_client`, so nothing exercises a real token fetch / 401 retry / refresh — which is exactly how the `scopes` bug got through. `tests_integ/mcp/` already spins up in-process `uvicorn`+`starlette` servers (`test_mcp_client_tasks.py`), and `uvicorn`/`starlette` are already test-time deps, so this is cheap and needs no network or credentials.
- **Reject null/empty credential values in config `auth`.** `missing_keys` (`mcp_client.py:1837`) checks key *presence*, not value: `{"client_id": null, "client_secret": null}` and `${VAR}` expanding to `""` both build a provider that fails later with an opaque handshake error.
- **Expose `token_endpoint_auth_method`** on the credentials type. `ClientCredentialsOAuthProvider` supports `client_secret_basic`/`client_secret_post`; today a `client_secret_post`-only AS forces you onto `auth_provider`, and the config path warns `ignoring unrecognized auth keys` while the constructor drops it silently. Not a parity gap — TS's provider hardcodes `client_secret_basic` too — so it's a genuine enhancement, not a merge condition.
- **`load_servers` config-schema docs.** Sweeping `site/`, this PR's OAuth section is the *only* place on the whole docs site that mentions `load_servers`; there's no page documenting the `mcpServers` JSON schema for a reader to land on.
- **TypeScript-tab OAuth docs.** Zero TS-tab OAuth content anywhere on the site even though `strands-ts` shipped the feature; `mcp-tools.ts:72-84` only shows a static bearer header.
- **Unwrapping `ExceptionGroup` in `start()`.** Bad OAuth credentials surface as `MCPClientInitializationError: ... unhandled errors in a TaskGroup (1 sub-exception)`, with the real `OAuthTokenError: Token exchange failed (401): {"error":"invalid_client"}` only in the chained traceback. Pre-existing behavior for any transport error (`mcp_client.py:387`), not introduced here — but bad creds are now a mainline path.

</details>

---

One process note, correcting my earlier comment on this PR: the labels in `team/API_BAR_RAISING.md:69` are **`api/needs-review`** / **`api/review-complete`** (not the names I used), and a CI check blocks merge of an `api/needs-review` PR without `api/review-complete`. Also `API_BAR_RAISING.md:58,62` asks the description to carry the complete signature with defaults and the module exports — worth adding before the API touchpoint.

*Reviewed by `strandly-the-agent` — an experimental AI reviewer. Verified claims are marked ✅/🔴 above; please treat the design questions as prompts for a human, not verdicts.*

strands-py/src/strands/tools/mcp/mcp_types.py
```diff
@@ -7,10 +7,29 @@
...
+    """OAuth client credentials for machine-to-machine authentication.
+
+    Used with the `MCPClient` `auth` parameter, or the `auth` key of a server entry in a
+    `load_servers` config, to authenticate against a streamable HTTP MCP server with the
+    OAuth client_credentials grant.
+
+    Attributes:
+        client_id: The OAuth client ID.
+        client_secret: The OAuth client secret.
+        scopes: OAuth scopes to request. Joined with spaces before sending to the token endpoint.
```
strandly-the-agent: 🔴 This sentence isn't true — see the `mcp_client.py:1693` thread for the mechanism and repro. The value is joined with spaces, but the server's advertised scopes overwrite it before the token request is built, and if the server advertises nothing the value is dropped entirely rather than used as a fallback.

(For what it's worth, this wording is inherited verbatim from `strands-ts/src/mcp/client.ts:74`, so the TS docstring has the same problem — separate fix.)

```suggestion
        scopes: OAuth scopes to request, joined with spaces. Advisory only: if the server
            advertises its own scopes (via the `WWW-Authenticate` header, or its
            protected-resource / authorization-server metadata), the server's list is used
            instead and this value is ignored.
```

strands-py/tests/strands/tools/mcp/test_mcp_client_auth.py
```diff
@@ -0,0 +1,157 @@
...
+def test_auth_scopes_joined_with_spaces(streamablehttp_transport):
+    client = MCPClient(
+        url="https://mcp.example.com",
+        auth=MCPClientCredentials(client_id="id", client_secret="secret", scopes=["read", "write"]),
+    )
+
+    client._transport_callable()
+
+    provider = streamablehttp_transport.call_args.kwargs["auth"]
+    assert provider._fixed_client_info.scope == "read write"
```
strandly-the-agent: 🟡 **This assertion guards nothing observable.** `_fixed_client_info.scope` is the field the token exchange never reads (see the `mcp_client.py:1693` thread) — a mutation test that gutted the entire `scopes` plumbing produced byte-identical wire behavior while this test stayed green.

Two things to fix together, and they compose: assert on the *constructor kwargs* instead of an `mcp` private attribute. That also addresses the private-internals coupling flagged earlier on this PR — for what it's worth I installed `mcp==1.23.0` (the floor) and confirmed `_fixed_client_info` and `context.server_url` exist and behave identically there, so the break risk today is low; it's just an attribute with no compatibility contract.

```python
def test_auth_scopes_joined_with_spaces(streamablehttp_transport):
    with patch("strands.tools.mcp.mcp_client.ClientCredentialsOAuthProvider") as provider_cls:
        client = MCPClient(
            url="https://mcp.example.com",
            auth=MCPClientCredentials(client_id="id", client_secret="secret", scopes=["read", "write"]),
        )
        client._transport_callable()

    assert provider_cls.call_args.kwargs["scopes"] == "read write"
```

Same shape works for `test_auth_builds_client_credentials_provider` at `:54` (`provider_cls.assert_called_once_with(server_url=..., storage=ANY, client_id="id", client_secret="secret", scopes=None)` plus `assert streamablehttp_transport.call_args.kwargs["auth"] is provider_cls.return_value`). I verified both replacements still catch the same mutations the originals do.

Worth adding while you're here (I have these passing locally if useful): the falsy-non-dict config `auth` branch parametrized over `[], {}, 0, ""`; `scopes` as a `str` as a regression pin; and `${VAR}` interpolation inside a `scopes` **list element** — that recursion works today but nothing exercises it (`test_auth_interpolates_env_vars` only interpolates the flat `client_id`/`client_secret`).

site/src/content/docs/user-guide/concepts/tools/mcp-tools.mdx
```diff
@@ -186,6 +186,53 @@ github_mcp_client = MCPClient(
...
+    url="https://api.example.com/mcp/",
+    auth={
+        "client_id": os.environ["OAUTH_CLIENT_ID"],
+        "client_secret": os.environ["OAUTH_CLIENT_SECRET"],
+        "scopes": ["mcp:tools"],  # optional
+    },
+)
+```
+
+Scopes are joined with spaces before being sent to the token endpoint. Token acquisition and refresh happen automatically. The `headers` parameter can be combined with `auth` when the server also expects custom headers.
```
strandly-the-agent: 🔴 First sentence isn't accurate — see the `mcp_client.py:1693` thread. Suggested rewrite:

```suggestion
`scopes` is joined with spaces and used as an advisory value only: if the server advertises its own scopes (via the `WWW-Authenticate` header, or its protected-resource / authorization-server metadata), the server's scopes are used instead and `scopes` is ignored. Token acquisition and refresh happen automatically. The `headers` parameter can be combined with `auth` when the server also expects custom headers.
```

Two more in this section while you're in the file:

**🟡 `:216` — the `OAuthClientProvider(...)` example fails if copy-pasted.** Ran it verbatim:
```
TypeError: OAuthClientProvider.__init__() missing 2 required positional arguments: 'client_metadata' and 'storage'
```
`client_metadata` needs an `OAuthClientMetadata` with a required `redirect_uris`, and `storage` needs a `TokenStorage` impl — neither is trivial and the docs give no guidance. The other `(...)`-elided snippets on the site are visibly sketch fragments; this one is formatted like the page's complete copy-pasteable blocks. Minimum fix is to name the two required args in the prose at `:209`: "...is one such implementation, though it requires supplying `client_metadata` (an `OAuthClientMetadata` with at least `redirect_uris`) and a `storage` implementing `TokenStorage`."

**🟡 `:220` — `[load_servers](#using-multiple-mcp-servers)` points at a section that never mentions `load_servers`.** That anchor resolves to `## Using Multiple MCP Servers` (`:294`), which only shows constructing several `MCPClient(lambda: ...)` instances by hand — no config loading, no `mcpServers` JSON. Sweeping all of `site/`, this OAuth section is the only place on the entire docs site that mentions `load_servers`, and there's no page documenting the config schema at all, so the in-scope fix is just to drop the link: "...can be set on a server entry in an `MCPClient.load_servers` JSON config, where...". A real config-schema section is a good separate follow-up.

✅ Verified accurate in this section: the first Python snippet runs verbatim; the JSON example parses and builds a client through `load_servers` verbatim and its key set exactly matches `_parse_config_auth`; `from mcp.client.auth import OAuthClientProvider` is the right import path for 1.23.0→1.29.0; the section sits correctly inside `<Tab label="Python">`; inline (rather than snippet-file) is the right convention for Python examples on this page; and the mutual-exclusivity sentence matches the actual invariants exactly.

site/src/content/docs/user-guide/concepts/tools/mcp-tools.mdx
```diff
@@ -186,6 +186,53 @@ github_mcp_client = MCPClient(
 )
 ```
 
+#### OAuth Authentication
```
Unshure, line 189: Should we update our Streamable HTTP docs to use this simpler devex of `url`?

Unshure (approved):

Unshure:
From an api bar-raise perspective, I like this change. Makes using the mcp client a bit easier, and shouldnt break existing users of it.

site/src/content/docs/user-guide/concepts/tools/mcp-tools.mdx
poshinchen: Will make a follow-up on it!

PR #3560, opened 2026-07-30
fix(ci): allow initial MCP release when no baseline tag exists

JackYPCOnline:
## Problem

The `release-mcp.yml` scan step aborts with:

> `No prior tag matching mcp/v* — refusing to release without a baseline.`

whenever no `mcp/v*` tag exists. No such tag exists anywhere — not locally, not on a fork, and not on upstream `main` — so the **very first** MCP release of `strands-agents-mcp-server` can never be cut through this flow. The `git fetch upstream --tags` hint in the error can't help, because upstream has no baseline either.

## Fix

Treat an empty `PREV_TAG` as the **initial release** rather than a hard error:

- **scan (§3)**: empty `PREV_TAG` → `::notice::` + `PREV_VERSION=""` instead of `::error::`/`exit 1`.
- **scan (§4)**: the duplicate/monotonic version comparisons only run when a baseline exists (they need one). The pre-existing-tag guard still always runs, so a re-typed initial version can't clobber an existing tag.
- **scan (§5)**: commit count falls back to `git rev-list --count "$RELEASE_SHA"` (all reachable commits) when there's no baseline.
- **draft-notes**: builds the log range as `$NEW_REF` alone when `PREV_TAG` is empty — `"$PREV_TAG..$NEW_REF"` with an empty left side is an invalid range and would otherwise abort the notes job even after the scan passed.
- **group-release-notes.py**: header renders `all commits up to \`mcp/v<x.y.z>\`` instead of a dangling `` `..mcp/v<x.y.z>` ``.

Subsequent releases are unaffected — they take the normal baseline path.

## Testing

- `yaml.safe_load` parses `release-mcp.yml`; `py_compile` passes on the notes script.
- Notes script exercised both paths (with/without baseline).
- Shell simulation of the scan against a repo with zero `mcp/v*` tags takes the initial-release path and passes.

### DIFF
--- a/.github/scripts/group-release-notes.py
+++ b/.github/scripts/group-release-notes.py
@@ -62,11 +62,16 @@ def main():
     new_tag = os.environ.get("NEW_TAG", "")
     prev_tag = os.environ.get("PREV_TAG", "")
 
+    # An empty prev_tag means the initial release: there is no baseline to
+    # diff against, so describe the range as "all commits up to" the new tag
+    # instead of rendering a dangling `..{new_tag}` range.
+    range_desc = f"`{prev_tag}..{new_tag}`" if prev_tag else f"all commits up to `{new_tag}`"
+
     out = []
     out.append(f"## {new_tag}")
     out.append("")
     out.append(
-        f"_Auto-drafted from commits in `{prev_tag}..{new_tag}`, grouped by "
+        f"_Auto-drafted from commits in {range_desc}, grouped by "
         "conventional-commit type. Edit on the release page after publish if "
         "you want a polished writeup; the canonical release notes live on the "
         "website._"
--- a/.github/workflows/release-mcp.yml
+++ b/.github/workflows/release-mcp.yml
@@ -77,36 +77,50 @@ jobs:
             RELEASE_SHA=$(git rev-parse origin/main)
           fi
 
-          # 3. Pick tag prefix and resolve previous tag.
+          # 3. Pick tag prefix and resolve previous tag. An empty PREV_TAG
+          #    means this is the first release of this package (no prior
+          #    baseline exists in history) — bootstrap it rather than refuse.
           TAG_PREFIX="mcp/v"
           PREV_TAG=$(git tag --list "${TAG_PREFIX}*" --sort=-v:refname | head -n1)
           if [ -z "$PREV_TAG" ]; then
-            echo "::error::No prior tag matching ${TAG_PREFIX}* — refusing to release without a baseline. (Fork-testing? Run \`git fetch upstream --tags && git push origin --tags\` first.)"
-            exit 1
+            echo "::notice::No prior tag matching ${TAG_PREFIX}* — treating this as the initial release. Monotonic and commit-count checks are skipped for the first release."
+            PREV_VERSION=""
+          else
+            PREV_VERSION="${PREV_TAG#"${TAG_PREFIX}"}"
           fi
-          PREV_VERSION="${PREV_TAG#"${TAG_PREFIX}"}"
           NEW_TAG="${TAG_PREFIX}${NEW_VERSION}"
 
-          # 4. Reject duplicate / non-monotonic / pre-existing tags.
-          if [ "$NEW_VERSION" = "$PREV_VERSION" ]; then
-            echo "::error::version $NEW_VERSION matches existing tag $PREV_TAG."
-            exit 1
-          fi
-          HIGHER=$(printf '%s\n%s\n' "$PREV_VERSION" "$NEW_VERSION" | sort -V | tail -n1)
-          if [ "$HIGHER" != "$NEW_VERSION" ]; then
-            echo "::error::version $NEW_VERSION is not greater than previous $PREV_VERSION."
-            exit 1
+          # 4. Reject duplicate / non-monotonic / pre-existing tags. The
+          #    version-comparison checks need a baseline, so they only run
+          #    when a previous tag exists; the pre-existing-tag check always
+          #    runs so a re-typed initial version can't clobber a tag.
+          if [ -n "$PREV_TAG" ]; then
+            if [ "$NEW_VERSION" = "$PREV_VERSION" ]; then
+              echo "::error::version $NEW_VERSION matches existing tag $PREV_TAG."
+              exit 1
+            fi
+            HIGHER=$(printf '%s\n%s\n' "$PREV_VERSION" "$NEW_VERSION" | sort -V | tail -n1)
+            if [ "$HIGHER" != "$NEW_VERSION" ]; then
+              echo "::error::version $NEW_VERSION is not greater than previous $PREV_VERSION."
+              exit 1
+            fi
           fi
           if git rev-parse --verify "refs/tags/$NEW_TAG" >/dev/null 2>&1; then
             echo "::error::tag $NEW_TAG already exists."
             exit 1
           fi
 
-          # 5. Sanity-check: at least one commit since prev tag.
-          COUNT=$(git rev-list --count "$PREV_TAG..$RELEASE_SHA")
-          if [ "$COUNT" = "0" ]; then
-            echo "::error::no commits between $PREV_TAG and $RELEASE_SHA — nothing to release."
-            exit 1
+          # 5. Sanity-check: at least one commit to release. With a baseline,
+          #    count commits since it; on the initial release, count all
+          #    commits reachable from the pinned SHA.
+          if [ -n "$PREV_TAG" ]; then
+            COUNT=$(git rev-list --count "$PREV_TAG..$RELEASE_SHA")
+            if [ "$COUNT" = "0" ]; then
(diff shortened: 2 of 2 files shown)


JackYPCOnline:
@strandly-the-agent safe to merge?

strandly-the-agent:
**Not yet — one blocker.** The shell is clean and the baseline path is provably unchanged, but merging this doesn't actually let you cut the first MCP release: the initial-release notes are **172,338 characters**, over GitHub's 125,000-char release-body limit, so `gh release create --notes-file` returns 422 — *after* the tag has already been pushed (`release-mcp.yml:409-412`), which wedges the flow. Cheap fix: scope the initial-release log to `strands-mcp/` (805 bytes, and the content becomes correct).

**Verified** on `e7dfc3ba` vs `main@4433e9a3`: ✅ notes script rendered both paths; ✅ scan step + draft-notes step extracted from the YAML and executed verbatim against throwaway repos with and without `mcp/v*` tags; ✅ baseline path identical (same rc, same `::error::` text, same `GITHUB_OUTPUT` across 6 version cases; baseline notes md5-identical to `main`); ✅ build/version-stamp path works with zero `mcp/v*` tags. ❌ Never dispatched a real run — see caveats.

| | |
|---|---|
| 🔴 Initial-release notes exceed the GitHub release-body limit → 422 → orphan tag | `release-mcp.yml:302-306`, `:409-419` |
| 🟡 No version floor on the initial path, and `strands-agents-mcp-server` is already on PyPI up to **0.2.7** | `:93-111`, `:16` |
| 🟡 A fork that didn't mirror tags now silently takes the initial-release path | `:85-90` |
| 🟡 `changelog-sync` is a no-op for `mcp/v*` tags (pre-existing, first observable here) | `:428-434` |

<details>
<summary>🔴 Detail + the one-line fix</summary>

At current `main`, `LOG_RANGE="$NEW_REF"` yields every non-merge commit in the monorepo — **2,295 bullets, 172,338 chars**:

```
$ git log --no-merges --pretty=format:'%h%x09%s' main \
    | NEW_TAG=mcp/v0.2.8 PREV_TAG="" python .github/scripts/group-release-notes.py | wc -c
172338
```

GitHub caps a release body at 125,000 chars (`422 ... Field:body ... body is too long (maximum is 125000 characters)` — e.g. goreleaser/goreleaser#2647, hit in exactly this "never been tagged before" scenario). So the run passes scan → notes → approval, pushes `mcp/v0.2.8` (`:409-412`), then fails at `:415-419`. All three recovery routes are then closed (reproduced with a stubbed `gh` that 422s):

- re-run the job → same deterministic 422 (so the "just re-run" contract at `:405-408` doesn't hold here);
- re-dispatch at the same version → `::error::version 0.2.8 matches existing tag mcp/v0.2.8` (`:98-101`);
- re-dispatch at the next version, same SHA → `::error::no commits between mcp/v0.2.8 and <sha> — nothing to release` (`:116-121`), because the orphan tag points *at* the release SHA.

Only escape is hand-deleting the remote tag. The deadlock itself is pre-existing; this PR makes it reachable on the very first use of the new path.

**Minimal fix** — only 7 of those 2,295 commits touch `strands-mcp/`, so path-scoping the initial-release log fixes size *and* accuracy (805 bytes) without touching the baseline path:

```bash
if [ -n "$PREV_TAG" ]; then
  git log --no-merges --pretty=format:'%h%x09%s' "$PREV_TAG..$NEW_REF" > commits.txt
else
  # Initial release: the whole-history log is ~172KB (over GitHub's 125k
  # release-body cap) and 99.7% other packages' commits — scope it to the package.
  git log --no-merges --pretty=format:'%h%x09%s' "$NEW_REF" -- strands-mcp/ > commits.txt
fi
```

…and the same pathspec on the §5 count (`:123`), which then reports 7 rather than 2,313. Truncation would also clear the 422, but leaves the notes wrong.

</details>

<details>
<summary>🟡 The premise is narrower than the PR description says — there's no version floor</summary>

`strands-agents-mcp-server` has **11 published PyPI versions**, latest `0.2.7` (2026-03-11):

```
$ curl -s https://pypi.org/pypi/strands-agents-mcp-server/json | jq -r .info.version
0.2.7
# 0.0.1 0.1.0 0.1.1 0.1.2 0.2.0 0.2.2 0.2.3 0.2.4 0.2.5 0.2.6 0.2.7
```

So this is the first *`mcp/v*`-tagged* release, not the package's first release. With `PREV_TAG` empty, `:93-111` skips the only checks that constrain the version — the bare-semver regex is all that's left:

```
### no mcp tags — version 0.0.0        -> rc=0, new_tag=mcp/v0.0.0
### no mcp tags — version 999.999.999  -> rc=0, new_tag=mcp/v999.999.999
```

Two concrete ways that bites:

- The dispatch input's own example is `'Explicit version, e.g. 0.2.0'` (`:16`) — **0.2.0 is already on PyPI** (2025-09-10). Typed verbatim it passes scan, builds, tags, creates the release, then fails at `publish-pypi` (`skip-existing` defaults to `false` and isn't overridden at `:456-457`).
- A *gap* version below 0.2.7 (0.0.2, 0.2.1) publishes fine and is a silent no-op — `pip install` still resolves 0.2.7, and it skews later `hatch-vcs` dev-version derivation.

The `release-gate` approver is the backstop, but `:136` renders `| Previous tag | _none — initial release_ |`, which tells them there's no prior release. Worth a floor (an `INITIAL_MIN_VERSION`, or a PyPI query) plus wording like "first monorepo-tagged release", and updating the `e.g. 0.2.0` example.

</details>

<details>
<summary>🟡 Fork-testing, changelog sync, and two nits</summary>

**Fork testing.** The error message this PR deletes named the case: *"(Fork-testing? Run `git fetch upstream --tags && git push origin --tags` first.)"*. Nothing now distinguishes "fork didn't mirror tags" from "genuine first release". Reproduced with a `--no-tags` clone of a repo whose upstream has `mcp/v0.2.8`:

```
BASE: rc=1  ::error::No prior tag matching mcp/v* — refusing to release without a baseline. (Fork-testing? ...)
HEAD: rc=0  ::notice::...treating this as the initial release...  prev_tag=; new_tag=mcp/v0.2.9
```

On a fork with `dry_run: false` that means skipped monotonic checks, the 172KB notes, the 422, and a bogus tag + release — on *every* fork test, not just once. Consider gating the bootstrap on `github.event.repository.fork != true` or an explicit `allow_initial_release: true` input.

**Changelog sync.** `site/scripts/changelog/tag-meta.ts:29-40` branches on `python/`, `typescript/`, bare `v<digit>` and `python-wasm/` — there is no `mcp/` branch, so it returns `null`, `build-release-file.ts:26-27` returns `null`, and the dispatched run writes 0 files and exits 0. The comment at `:421-427` ("the daily cron backstops") doesn't hold — the cron drops `mcp/` tags the same way. `changelog-sync.yml:31-36`'s release-event gate also omits `mcp/v`. Pre-existing, but this PR is what makes it observable. If MCP is deliberately out of changelog scope, worth saying so there.

**⚪ Nit** — `group-release-notes.py:68` reads *"Auto-drafted from commits in all commits up to `mcp/v0.2.8`"*. Let `range_desc` carry the preposition (`f"up to \`{new_tag}\`"`).

**⚪ Nit** — `:123` has no `--no-merges`, so the summary says `Commits in initial release: 2313` while the notes list 2295. (Matches the baseline path, so arguably fine.)

</details>

<details>
<summary>What I attacked and could NOT break / could not verify</summary>

**Held up:** shell hygiene under `set -euo pipefail` on both paths (`PREV_VERSION`, `COUNT`, `PREV_TAG_CELL`, `COUNT_LABEL`, `LOG_RANGE` assigned on every branch — no unbound var, no word-splitting); baseline-path parity over `0.3.0 / 0.2.0 / 0.1.5 / 0.2.1 / 1.0.0 / 0.10.0`; the pre-existing-tag guard still fires on the initial path; `SETUPTOOLS_SCM_PRETEND_VERSION` build with zero `mcp/v*` tags produces correctly-stamped dists that satisfy `inspect`'s regex (`:228-230`); `GITHUB_STEP_SUMMARY` at ~172KB is under the 1 MiB cap (unwieldy, not fatal); the depth-1 tag push in `create-gh-release` works.

**Not verified:** no real workflow dispatch, so Actions' handling of an *empty* job output is assumed — `${{ needs.scan-commits.outputs.prev_tag }}` should evaluate to `''` (same pattern as `SHA_INPUT: ${{ inputs.sha }}` read under `set -u` at `:67`), and if it were ever *unset* the failure would be loud (`PREV_TAG: unbound variable`), not silent. One dry run confirms it. Also can't inspect PyPI trusted-publisher config for this project — a pre-flight item that fails *after* the tag and release are pushed.

</details>

Correcting my earlier automated pass on this PR, which filed the whole-history notes as non-blocking: quantified, they cross a hard GitHub limit, so it's blocking. Sorry for the flip-flop. With the `-- strands-mcp/` pathspec I'd call it good; the version floor is the one I'd want a human call on. As always, worth a maintainer's eyes rather than mine alone.

liramon2:
Looks fine to me stylistically. The strandly blockers above seems worth addressing.

PR #2782, opened 2026-06-15
fix(openai): report Responses prompt-cache tokens (TypeScript)

didhd:
## Description

The TypeScript `OpenAIModel` Responses adapter dropped OpenAI prompt-cache usage. `mapResponsesEventToSDK` built `finalUsage` from `input_tokens` / `output_tokens` / `total_tokens` only, so `usage.input_tokens_details.cached_tokens` was never surfaced as `cacheReadInputTokens`. Cache reads were therefore invisible in `AgentResult` usage metrics and OTel spans (`gen_ai.usage.cache_read_input_tokens`) for the Responses path, even when the API was returning `cached_tokens`.

This restores parity with the other adapters — `models/bedrock.ts`, `models/anthropic.ts`, and `models/vercel.ts` already plumb `cacheReadInputTokens`/`cacheWriteInputTokens`. The `Usage` type (`models/streaming.ts`) already declares both fields and the accumulator already sums them, so only the Responses adapter needed updating.

It is the TypeScript counterpart of the Python issue/PR (#2407 / #2663). The write side already works (a caller can pass `prompt_cache_key` / `prompt_cache_retention` via `params`, which `formatResponsesRequest` spreads into the request body); this PR makes the resulting cache reads observable.

Changes:
- Extract `mapResponsesUsage()` and call it from both the `response.completed` and `response.incomplete` branches (dedupes the existing mapping).
- Set `cacheReadInputTokens` only when `cached_tokens > 0`, matching the optional-field behavior of the other adapters.
- Add unit tests for a cache hit and the no-cache (`cached_tokens: 0`) case.

## Related Issues

Fixes #2781

## Documentation PR

No docs change needed — this aligns the Responses adapter with existing usage/telemetry behavior already documented for the other model providers.

## Type of Change

Bug fix

## Testing

`strands-ts`: `npm run build` (tsc, no errors), `npm run lint` (0 errors), `npm test` (full unit-node suite: 124 files / 3396 tests pass, no type errors). The two new tests in `src/models/openai/__tests__/responses.test.ts` cover the cache-hit and no-cache paths.

- [x] I ran the TypeScript equivalents (`npm run build`, `npm run lint`, `npm test`)

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have reviewed and understand every line of code in this PR, including any generated by AI tools, and I can explain why it works
- [x] My change is focused and reasonably small; I have split unrelated work into separate PRs
- [x] I have added any necessary tests that prove my fix is effective or my feature works
[...]

PR #2663, opened 2026-06-08
fix: report OpenAI Responses cached tokens

fengjikui:
## Description

`OpenAIResponsesModel` now carries prompt-cache reads from `usage.input_tokens_details.cached_tokens` into `cacheReadInputTokens`, matching the existing Chat Completions model behavior. This lets agent metrics and telemetry report OpenAI Responses API cache reads when OpenAI returns them.

No public API changes. The existing usage metadata shape already supports `cacheReadInputTokens`; this fills the missing provider mapping.

This PR was AI-assisted. I reviewed the change and ran the relevant local validation.

## Related Issues

Resolves #2407

## Documentation PR

No documentation changes needed.

## Type of Change

Bug fix

## Testing

Validated locally in `strands-py/`:

- `hatch test tests/strands/models/test_openai_responses.py -k 'metadata_with_cache_tokens or metadata_with_zero_cached_tokens'` first reproduced the missing `cacheReadInputTokens` behavior before the implementation, then passed after the fix
- `hatch test tests/strands/models/test_openai_responses.py`
- `env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY -u http_proxy -u https_proxy -u all_proxy hatch run hatch-static-analysis:lint-check`
- `env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY -u http_proxy -u https_proxy -u all_proxy hatch run hatch-static-analysis:ruff format --check src/strands/models/openai_responses.py tests/strands/models/test_openai_responses.py`
- `git diff --check`

I did not run the full `hatch run prepare` because its repository-wide formatter step currently reports two pre-existing unrelated files that would be reformatted: `src/strands/tools/mcp/mcp_client.py` and `tests/strands/models/test_gemini.py`. The changed files pass the formatter check.

- [ ] I ran `hatch run prepare`

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [x] I have updated the documentation accordingly, or no documentation update is needed
- [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3561, opened 2026-07-30
fix(types): make token usage counters disjoint across every provider

### LINKED ISSUE #797: [BUG] totalTokens not equal to inputTokens + outputTokens
### Checks

### Strands Version

1.0.1

### Python Version

3.12

### Operating System

macOS

### Installation Method

pip

### Steps to Reproduce

Following creation of an agent, run
`
query = "This is my query to the agent"
response = agent(query)
`

It is unpredictable, but I have gotten the following accumulated usage (using a query I cannot share): {'inputTokens': 4674, 'outputTokens': 328, 'totalTokens': 5746} when I run 
`response.metrics.accumulated_usage`

You can see that inputTokens + outputTokens < totalTokens. 

### Expected Behavior

The documentation states inputTokens + outputTokens = totalTokens

### Actual Behavior

inputTokens + outputTokens < totalTokens

### Additional Context

_No response_

### Possible Solution

_No response_

### Related Issues

_No response_

### LINKED ISSUE #3546: [BUG] Usage cache-token fields mix two incompatible conventions, over-counting cost on OpenAI/Gemini/LiteLLM cache hits
## Summary

`Usage.cacheReadInputTokens` is populated from provider fields that do not mean the same thing, and nothing on the type says which convention applies. For Bedrock Converse and the Anthropic API, cache tokens are **disjoint** from `inputTokens` — the true prompt is `inputTokens + cacheRead + cacheWrite`. For OpenAI (Chat Completions and Responses), Gemini, and LiteLLM, the cache counts are a **subset** of the input field — the same tokens are reported twice, once in the total and once broken out.

Both are mapped onto one optional field and `_accumulate_usage` sums them blindly, so any consumer doing the obvious thing — `total = input + output + cacheRead + cacheWrite`, or `cost = Σ tokens × rate` — over-counts every OpenAI-style cache hit and charges those tokens twice, at the input rate *and* the cache rate. On agentic workloads, where most of the prompt is a cache hit, that is a multiple rather than a rounding error. The same code path is correct for Bedrock, which is what makes it hard to spot.

## Reproduction

Live capture through `Agent` on `openai.gpt-5.6-luna` via Bedrock Mantle, us-east-1 (same prompt twice, second call is a cache hit):

```
call 1 accumulated: {"inputTokens": 6459, "outputTokens": 5, "totalTokens": 6464}
call 2 accumulated: {"inputTokens": 12936, "outputTokens": 10, "totalTokens": 12946, "cacheReadInputTokens": 6457}

sum(4 counters) = 19403  vs  accumulated totalTokens = 12946   → over-counts by 6457 (+50%)
```

The provider's own `usage` for call 2 was `{"input_tokens": 6452, "output_tokens": 5, "total_tokens": 6457, "input_tokens_details": {"cached_tokens": 6450, "cache_write_tokens": 0}}` — `6452 + 5 == 6457`, so `cached_tokens` is inside `input_tokens`.

Contrast Bedrock Converse on `eu.anthropic.claude-haiku-4-5`, eu-north-1, same shape of workload:

```
call 1 {"inputTokens": 10, "outputTokens": 4, "totalTokens": 5862, "cacheReadInputTokens": 0, "cacheWriteInputTokens": 5848}
call 2 {"inputTokens": 10, "outputTokens": 4, "totalTokens": 5862, "cacheReadInputTokens": 5848, "cacheWriteInputTokens": 0}
```

`10 + 4 + 5848 == 5862`. Here the cache counters are additional tokens and `totalTokens` includes them. No single interpretation of `cacheReadInputTokens` is correct for both providers.

## Per-provider findings

| Provider | Convention | Issue |
|---|---|---|
| Bedrock Converse | disjoint ✅ **[captured]** | correct today; the de facto reference |
| Anthropic direct | disjoint **[documented]** | `totalTokens` is recomputed as `input + output`, **excluding cache writes** — under-reports a cached run. Anthropic's docs state `total_input_tokens = cache_read + cache_creation + input_tokens` |
| OpenAI Chat / Responses | subset **[captured via Mantle]** | cache counts are inside `inputTokens`, so summing over-counts. `cache_write_tokens` is dropped (it exists first-party from GPT-5.6, billed at 1.25× input) |
| Gemini | subset **[captured]** | `cached_content_token_count` is inside `prompt_token_count`. Separately, `outputTokens` was inferred as `total - prompt`, which silently counted `tool_use_prompt_token_count` (an *input* type) as output |
| LiteLLM | subset **[verified in litellm source]** | folds cache into `prompt_tokens` for *every* backend including Anthropic (`prompt_tokens += cache_creation_input_tokens + cache_read_input_tokens`), so the existing "only LiteLLM over Anthropic supports cache write tokens" comment is inverted |
| Vercel AI SDK | either | reads `inputTokens.total` (inclusive) when the AI SDK also provides `inputTokens.noCache`, which is exactly the disjoint value |
| TS Chat Completions | — | drops `prompt_tokens_details` entirely, so cache reads are invisible (Python's does read it — a py/ts parity gap) |
| Reasoning tokens | — | `completion_tokens_details.reasoning_tokens` / `thoughts_token_count` are billed but never surfaced |

Gemini's own field docs confirm `total_token_count` is "the sum of `prompt_token_count`, `candidates_token_count`, `tool_use_prompt_token_count`, and `thoughts_token_count`" — verified live: `23293 + 1 + 0 + 80 == 23374`.

## Blast radius beyond cost reporting

`Usage` is not only a metric. Anything derived from `inputTokens` or `totalTokens` inherits the ambiguity:

- **`limits.total_tokens`** is an agent-loop stopping condition (`_check_limits`), so the convention decides when a run halts.
- **`latest_context_size` / `projected_context_size`** (public API via `AgentResult.context_size`) are *context-window* quantities, not cost quantities — cached tokens occupy the window regardless.
- **Proactive compression** gates on that same estimate (`_estimate_input_tokens` → `BeforeModelCallEvent.projected_input_tokens` → `ConversationManager`). Under a subset-convention provider a mostly-cached 100k prompt reads as ~5k, so compaction never fires and the run hits a hard context-window error instead.
- **Multi-agent aggregation** (`Graph`/`Swarm._accumulate_metrics`) sums only the three required counters, dropping the cache counts while still adding their `totalTokens` — the total then reports tokens no counter accounts for.
- **OTel** (`telemetry/tracer.py`) emits `gen_ai.usage.input_tokens`, which the [GenAI semconv](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/model/gen-ai/registry.yaml) defines as *including* cached tokens ("This value SHOULD include all types of input tokens, including cached tokens"), with the cache attributes as breakdowns of it. So the SDK's field and the emitted attribute genuinely should differ, which needs stating rather than leaving to chance.

## Proposed fix

1. **Write the contract down on `Usage`.** Whichever convention is chosen, the doc comment must say whether the cache counters are *also* in `inputTokens`, and whether `totalTokens` includes them. This is the cheapest half and the half that stops the next adapter repeating it.
2. **Normalize to disjoint**, i.e. `inputTokens + outputTokens + cacheReadInputTokens + cacheWriteInputTokens == totalTokens`. Bedrock Converse is already disjoint, so the flagship provider's numbers do not move; cost becomes a plain weighted sum, which is the point. Route every adapter through one shared helper taking a single `input_includes_cache` flag, so a new provider makes one decision instead of re-deriving the arithmetic.
3. **Keep OTel conformant separately** — add the cache counters back when emitting `gen_ai.usage.input_tokens`, per semconv.
4. **Fix the context-window consumers** to use the full prompt, not `inputTokens`.
5. **Surface reasoning tokens** as a field documented as a *subset* of `outputTokens` (so it must not be added to a total).
6. **Test the invariant, not field presence.** A test asserting only `cacheReadInputTokens == 1920` passes under both conventions and proves nothing — that is how this got in.

This changes the *value* of `inputTokens` for OpenAI/Gemini/LiteLLM users and of `totalTokens` for Anthropic-direct users. It is a behaviour change on metrics fields rather than an API break, but it deserves a release note: anyone charting those fields will see a step.

## Environment

`harness-sdk@main`. Captures on Bedrock Converse (eu-north-1), Bedrock Mantle Responses (us-east-1), and Gemini (AI Studio) were taken directly against the live APIs. Anthropic-direct and first-party OpenAI are **not** wire-captured — those rows are from vendor docs and the installed SDK type definitions.

I have a fix ready for both SDKs and will open a PR referencing this issue.

niklas-palm:
## Description

This is a large diff — 84 files across both SDKs and the docs. The reason is that it fixes a contract rather than a call site, and a contract that holds for only some providers is not a contract. Everything below is an attempt to justify that scope; if the argument doesn't land, the last section lists what can be split out.

### What's wrong today

`Usage` gained `cacheReadInputTokens` and `cacheWriteInputTokens` in #529, modelled on Bedrock Converse. There, those counters sit *alongside* `inputTokens` and are already inside the provider's `totalTokens` — the four counters are disjoint and sum to the total.

Over the following months the same two fields were plumbed through for providers that use the opposite convention. OpenAI (#2116, #2555, #2663), Gemini (#2287) and Anthropic (#2302) report the whole prompt in their input field with the cache broken out as a *subset* of it. Both readings were reasonable in isolation and each PR was correct for its own provider. What was missing was anywhere to record which convention a `Usage` was carrying — so no consumer could be written that was right for both.

The result is that summing the four counters double-counts every cached token on the inclusive providers. The example payload in #2116's own description shows it: `prompt_tokens=1861` with `cached_tokens=1792` sums to 3663 against a true total of 1871, **1.96x over**. Reading `inputTokens` alone instead under-counts on Bedrock. A graph mixing two providers could not be reconciled either way, and the cost is real money on the provider whose cache discount is the reason to enable caching at all.

### What this changes

The counters become **disjoint** everywhere — every billed token belongs to exactly one counter, for every provider, in both SDKs:

```
inputTokens + outputTokens + cacheReadInputTokens + cacheWriteInputTokens == totalTokens
```

That single line is what makes cost a plain weighted sum, and it is the whole point. It is also why this is one PR: a per-provider fix leaves the contract undefined in between, and the value of an invariant is that a consumer can rely on it without knowing which provider produced the numbers.

Each adapter now states which convention its provider uses, and one shared helper per SDK converts to the contract. Adding a provider becomes answering one question instead of rediscovering the distinction:

```python
# models/openai.py — the whole of what an adapter now decides
usage_data = normalize_usage(
    input_tokens=event["data"].prompt_tokens,
    output_tokens=event["data"].completion_tokens,
    cache_read_tokens=token_detail(prompt_details, "cached_tokens"),
    cache_write_tokens=cache_write_count(event["data"], prompt_details),
    reasoning_tokens=token_detail(completion_details, "reasoning_tokens"),
    input_includes_cache=True,  # this provider's prompt count already contains the cache
)
```

Because there is one seam, a third-party `Model` that reports the old shape is surfaced rather than silently mis-billed: the event loop warns when the counters don't account for the total, and separately when a count can't be read as a non-negative whole number.

Two quantities that resemble cost but aren't now come from their own helpers, which is where the second-order bugs were hiding:

**Context-window sizing.** Cached tokens still occupy the window even though they're cheap, so `context_size` and the proactive-compaction baseline must count the whole prompt. On `main` a 90k-token cache read measured as roughly 15 tokens, which meant proactive compaction never fired for exactly the workloads that need it most.

**OpenTelemetry.** Per the [GenAI semantic conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/), `gen_ai.usage.input_tokens` is defined as the whole prompt *including* cached tokens, with the cache attributes as breakdowns of it. So that attribute deliberately adds the cache back and legitimately differs from the SDK's own `inputTokens`. This one surprised me, and it reversed the direction I first took — it's called out in the traces docs so the next person doesn't read the difference as a bug.

Two smaller items ride along because the audit surfaced them in the same code. `reasoningOutputTokens` is added, deliberately as a **subset** of `outputTokens` rather than a fifth billed counter, because those tokens are already billed at the output rate — the Writer docs previously told users to bill them separately, which is this same double-charge in miniature. And Gemini was counting `tool_use_prompt_token_count` as output when it is an input type, so `outputTokens` drops on tool-result turns.

**One quantity, several field names.** The single hardest part of getting this right, and the thing I'd most welcome a second pair of eyes on. A vendor frequently reports the *same* count under more than one name depending on which backend it fronts, and reading only one of them leaves that count folded inside the prompt total, billed at the full input rate. It is invisible: the four counters still sum to the reported total, so neither warning fires. A cache write reaches an OpenAI-compatible usage object as `cache_creation_input_tokens` (when the gateway fronts Anthropic), or as `cache_creation_tokens` or `cache_write_tokens` on the prompt details. Mistral's own usage type declares no cache field at all but permits extras, so a hit arrives as `prompt_tokens_details`, `prompt_token_details`, or a flat `num_cached_tokens`. Each SDK now reads every name through one helper, so two adapters cannot drift apart — the OpenAI and LiteLLM providers had already diverged on byte-identical payloads before that consolidation.

### On the size

The 84 files break down as: every adapter in both SDKs and their tests (23 Python, 18 TypeScript) because the invariant has to hold everywhere; the places that measure the context window and emit telemetry; the multi-agent aggregation; and 14 documentation pages. The docs share is deliberate rather than padding — this changes numbers people already chart, so the migration table, the per-provider support matrix and the custom-provider guide are part of the fix.

About half the added lines are tests. Every expected value comes from a live provider response or that provider's own published documentation, cited per case, because several pre-existing tests encoded the inclusive convention as correct and so could not serve as a baseline.

## Public API Changes

`Usage` gains one optional counter, `reasoningOutputTokens`, reported by the providers that expose it. It is a subset of `outputTokens`, not an addition to it:

```python
# Before: reasoning tokens were invisible, and the four counters could not be trusted to sum
usage = result.metrics.accumulated_usage
usage["inputTokens"]   # net new tokens on Bedrock, the whole prompt on OpenAI

# After: the counters are disjoint, and reasoning is visible as a share of output
usage["inputTokens"]              # always net new: neither read from nor written to cache
usage.get("reasoningOutputTokens")  # part of outputTokens, already billed at the output rate
```

Two internal helpers back the change — `normalize_usage` in each SDK's `models` package, and the prompt/context sizing helpers in `types`. Both are deliberately non-public (out of `__all__` and the `index.ts` barrel) so the contract, not the helper, is the thing consumers depend on.

## Related Issues

Resolves: #3546

Resolves: #797 — the same bug, reported ten months ago as "totalTokens not equal to inputTokens + outputTokens" and auto-closed without a fix. The reported `{'inputTokens': 4674, 'outputTokens': 328, 'totalTokens': 5746}` leaves a 744-token gap. That gap is the cache counters: `Graph._accumulate_metrics` and its `Swarm` equivalent summed three counters while keeping the cache-inclusive `totalTokens`, so the cache tokens vanished from the breakdown but stayed in the total. The thread reasonably read this as provider overhead from guardrails and proposed recomputing `totalTokens` as `input + output`, which would have discarded real billed tokens. Under this change those exact numbers reconcile.

How the two conventions came to coexist, for the record: #529 introduced the fields; #2116, #2287, #2302, #2555, #2663 and #2782 plumbed them through for providers using the other convention. strands-agents/sdk-typescript#895 is the mirror image, where TypeScript reported Anthropic cache tokens while Python dropped them — the SDKs had drifted in opposite directions.

Unblocks #1428 and #1216, which both expose dollar cost derived from `accumulated_usage` and would otherwise inherit the over-count in a headline number.

## On process, before you read the diff

Two things I could not resolve from the guidelines, so I am flagging rather than assuming.

`team/designs/README.md` asks for a design document for "breaking changes to existing APIs" and
says to skip it for bug fixes. This is both: no API shape changes, but values users chart do.
`team/COMPATIBILITY.md` covers shape rather than value, so it does not settle it either, and every
existing design doc is a new feature rather than a correction. **If you would rather see a design
proposal first, say so and I will write one** — I did not want to gate a cost-accuracy fix behind a
process step that may not apply, but it is your call, not mine.

The contributing guide also asks for small, incremental PRs, which this is not. The reasoning is in
"On the size" above, and the checklist item is left unchecked rather than quietly ticked.

## Documentation PR

Included under `site/`, per the contributing guide. The token-accounting contract, a migration table of every value that changes, and a per-provider support matrix are on the metrics page; the custom-provider guide documents the contract and both warnings for third-party `Model` authors.

## Type of Change

Breaking change

## Testing

I ran the individual gates rather than `hatch run prepare`: `ruff format`, `ruff check`, `mypy`,
and the Python unit suite (4,548 passing), plus `npm run build`, `lint`, `type-check`,
`format:check` and the unit suite for TypeScript (3,997 passing), and `npm test` with
`typecheck:snippets` for the docs site (473 passing). The site builds clean — 867 pages, no broken
links. All gates pass.

What I have **not** run is `hatch test --all`, the multi-Python-version matrix that `prepare`
includes, so CI is the first place this is exercised on anything other than 3.11. Two Python test
failures reproduce identically on `main` (an OpenTelemetry `TraceFlags.RANDOM_TRACE_ID` version
mismatch, confirmed against a clean `main` worktree) and are unrelated to this change; the
`telemetry/test_config.py` errors and the `a2a`/`cedar` collection errors are missing optional
dependencies in my local venv, not failures.

`ruff format` reports one file it would reformat, `models/writer.py`. That file is already
unformatted on `main` at a line this change never touches, so I left it alone rather than pad the
diff with a drive-by reformat.

Verified against live providers, not only unit tests. Bedrock Converse (streaming and non-streaming, cached and uncached, tool use, extended thinking, and a non-Anthropic model on the same API), Bedrock Mantle Responses, and Gemini (plain, thinking, and explicit `caches.create`) all reconcile exactly. So does a graph combining a disjoint provider with an inclusive one, which is the case this change exists for. Sessions round-trip through `FileSessionManager`.

A few of those, so the claim is checkable rather than just asserted:

```
Bedrock Converse, cached      31 +  10 + 13084 + 13084 == 26209
Bedrock Mantle, two turns      4 +  10 + 11610 + 11594 == 23218
Gemini, explicit cache         5 +  50 + 11644         == 11699   (reasoning 48 ⊆ output 50)
Mixed-provider graph          accumulated 70 == sum of node totals
```

Vercel AI SDK, LiteLLM, Writer, Mistral, Ollama and llama.cpp were verified by reading and running their installed vendor sources. LiteLLM needed unpacked wheels across the supported range, because which counters it folds into `prompt_tokens` changed mid-range and differs by proxied backend.

Anthropic-direct and OpenAI-direct are inference from vendor SDK type definitions and published docs — I had no API keys for either — and the per-provider matrix labels them as such so the distinction survives into the docs.

## Checklist

That third box is unchecked deliberately. The change is focused on one thing but it is not small, and `AGENTS.md` is explicit that a branch spanning several sub-projects is usually several PRs. Since you may well want this split, here is a concrete proposal rather than a defence — say which you prefer and I will do the work:

| | Scope | Files |
|---|---|---|
| **A** | The `Usage` contract, `normalize_usage` in both SDKs, every adapter | ~45 |
| **B** | The consumers: context-window sizing, telemetry, multi-agent aggregation | ~20 |
| **C** | `reasoningOutputTokens` as a new public field | ~8 |
| **D** | Docs under `site/`, `team/`, `.agents/` | ~15 |

**My one substantive caution: A alone is worse than either endpoint.** It redefines `inputTokens` as net-new while `context_size` still reads it as the whole prompt, so a 90k-token cache read measures as roughly 15 tokens and proactive compaction never fires — a live bug that does not exist today and would not exist after B. So I would ship **A+B together** as the contract and the consumers that depend on it, then **C** and **D** as follow-ups. C is genuinely separable and I agree it wants its own API bar-raising review; it is not what #3546 asks for.

If you would rather have four PRs anyway, I will do that and sequence them so no intermediate state ships the compaction bug — but that means A and B land back to back regardless, which is why I have not split them here.

## Breaking Changes

Numbers users already chart change with no action on their part.

`inputTokens` drops by the cached amount on OpenAI, Google, LiteLLM, Writer, SageMaker and the Vercel AI SDK. `totalTokens` grows to include cache reads and writes on Anthropic direct, which also makes `limits.total_tokens` trip sooner. Google's `outputTokens` drops on tool-result turns, which now count as input. `context_size` and its projected form grow to the full prompt on any cached provider, as does `gen_ai.usage.input_tokens`. Bedrock Converse `inputTokens` and `totalTokens` are unchanged.

Mistral now reports `cacheReadInputTokens` where it previously reported none, so its `inputTokens` drops by that amount on a cached call. No Mistral model prices a cache today, so this corrects a count rather than a bill — but it is a value that moves.

A cost formula that summed the four counters was over-charging cached calls at both the input and the cache rate, and is now correct as written. One that used `inputTokens` alone should add the cache terms:

```python
usage = result.metrics.accumulated_usage

cost = (
    usage["inputTokens"] * INPUT_RATE
    + usage.get("cacheReadInputTokens", 0) * CACHE_READ_RATE
    + usage.get("cacheWriteInputTokens", 0) * CACHE_WRITE_RATE
    + usage["outputTokens"] * OUTPUT_RATE
)
```

To recover the full prompt a provider received, add the cache counters back to `inputTokens`.

`metadata["usage"]` now carries only the six counters `Usage` declares. Previously the provider's dict was spread through, so an undeclared key survived — Bedrock Converse ships `cacheDetails`, a per-TTL breakdown of the cache write, and that is now dropped. Nothing in the SDK reads it, so this is tightening rather than a bug, but it is observable to anyone reading the metadata event directly. Say the word if you would rather the extra keys were preserved.

### Dependency — worth your explicit sign-off

The `litellm` extra's floor moves from `>=1.75.9` to `>=1.78.0`, which is a breaking change for existing installs and the one decision here I would most like a maintainer to weigh rather than accept.

1.78.0 is the first release whose Bedrock Converse adapter folds a cache write into `prompt_tokens` (`input_tokens += cache_creation_input_tokens`, absent in 1.77.7 — I unpacked the wheels across the range to find the boundary). Below it, whether the write is folded depends on which backend LiteLLM proxies, and the usage object carries no signal saying which. So the counts cannot be made correct for every backend below that floor.

The alternative is to keep the floor and tolerate a wrong count on older releases. I chose correctness because a silently wrong cost is the bug this PR exists to fix, but it does force an upgrade on anyone pinned in 1.75.9–1.77.1, and that is your call rather than mine.

### Three things worth flagging

I'd rather surface these than have them found in review.

`UsageMetadata.prompt_tokens_details` in the SageMaker provider is retyped from `int | None` to `dict[str, Any] | None`. This one **is** required: `cached_tokens` lives inside that field, so it's the only way to obtain SageMaker's cache count at all. The old `int` annotation was already wrong — the wire shape is a dict, and the dataclass was storing one while declaring an int. The dataclass is importable but is not in `models.__all__`.

`ollama.py` gained a null guard on `total_duration` — a latency field, not a token count, and not required by the contract. The same response that omits `eval_count` omits it too, on the path the new test exercises, so it was fix-it-where-you-find-it rather than scope creep. Happy to pull it out.

`team/designs/0008` and `0015` have small prose corrections where they describe token accounting in terms this change invalidates. Also not required, but leaving an accepted design doc describing the old behaviour seemed worse than the two-line diff.


niklas-palm:
One correction to the description before anyone reads the diff: the `## Testing` section cites the gate numbers I measured locally (4,548 Python / 3,997 TypeScript). Those were taken against `upstream/main` as of this morning; the branch has been rebased since, so treat CI as the authority rather than those figures.

Also worth stating plainly, since the description mentions it only in passing: the Anthropic-direct and first-party OpenAI paths are the two providers I could **not** exercise against a live API — I had no keys for either. Their behaviour is derived from the vendor SDKs' own type definitions and published docs, and the per-provider matrix on the metrics page labels them as such. Everything else in the matrix is a wire capture.

opieter-aws:
Thanks for your contribution to the SDK and trying to fix this issue! We're wary to review and merge such a big change touching so many files for a problem of this size. As such, I am going to close this PR. We will look into the bug and see how we can do more bite-size resolution.

PR #3574, opened 2026-07-30
refactor(vended-tools)!: rename sandbox-routed bash tool to shell

yonib05:
## Description

The sandbox-routed vended tool is named `bash`, but it never runs bash. It delegates to `Sandbox.execute()`, and the concrete sandboxes run:

- [`docker.py:92`](https://github.com/strands-agents/harness-sdk/blob/fe4cbb9486566154b1f94e3ea3c6a85a2bd81f43/strands-py/src/strands/sandbox/docker.py#L92) / [`docker.ts:74`](https://github.com/strands-agents/harness-sdk/blob/fe4cbb9486566154b1f94e3ea3c6a85a2bd81f43/strands-ts/src/sandbox/docker.ts#L74) — `sh -c`
- [`not_a_sandbox_local_environment.py:79`](https://github.com/strands-agents/harness-sdk/blob/fe4cbb9486566154b1f94e3ea3c6a85a2bd81f43/strands-py/src/strands/sandbox/not_a_sandbox_local_environment.py#L79) — `sh -c`
- [`ssh.py:154`](https://github.com/strands-agents/harness-sdk/blob/fe4cbb9486566154b1f94e3ea3c6a85a2bd81f43/strands-py/src/strands/sandbox/ssh.py#L154) — the remote login shell

The base contract documents itself generically as ["Execute a shell command"](https://github.com/strands-agents/harness-sdk/blob/fe4cbb9486566154b1f94e3ea3c6a85a2bd81f43/strands-py/src/strands/sandbox/base.py#L227). The tool invokes whichever shell the sandbox is loaded with, so the name promises an interpreter the code does not guarantee — on Alpine or Debian, `sh` is busybox or dash, and bashisms like `[[ ]]` or arrays fail despite the tool being called `bash`.

### The rename is intentionally asymmetric across SDKs

This is the part worth reviewing closely, since it looks like a parity violation but is deliberate:

| Surface | What it actually runs | Name accurate? | Action |
|---|---|---|---|
| PY `bash` / `make_bash` | `Sandbox.execute()` → `sh` or remote shell | No | Renamed to `shell` / `make_shell` |
| TS `makeBash` | `Sandbox.execute()`, same routing | No | Renamed to `makeShell` |
| TS `bash` ([`bash.ts:39`](https://github.com/strands-agents/harness-sdk/blob/fe4cbb9486566154b1f94e3ea3c6a85a2bd81f43/strands-ts/src/vended-tools/bash/bash.ts#L39)) | `spawn('bash', [])`, persistent session | **Yes** | **Unchanged** |

TypeScript's standalone `bash` hardcodes the bash binary and holds a stateful session across calls. Its name is correct, and renaming it to `shell` would imply portability it does not have — it fails outright on a host without bash. Both sandbox-routed surfaces are renamed in lockstep, so the SDKs stay in parity where a shared concept actually exists; the asymmetry is that TypeScript has an extra, genuinely-bash tool that Python has no equivalent of.

### Back-compat

Deprecated aliases are retained and removed in **v2.0.0**, per `team/FEATURE_LIFECYCLE.md` (features are not removed in minor versions).

Python resolves them via a module-level `__getattr__` — the same pattern already used in [`experimental/tools/__init__.py`](https://github.com/strands-agents/harness-sdk/blob/fe4cbb9486566154b1f94e3ea3c6a85a2bd81f43/strands-py/src/strands/experimental/tools/__init__.py#L11) — because the deprecated `bash` is a module-level *instance*, which a decorator cannot intercept. `typing_extensions.deprecated` was not used for it for the same reason (and `warnings.deprecated` needs 3.13 while the SDK targets 3.10).

One deliberate limitation: no `vended_tools/bash.py` shim module is added, so the deep path `strands.vended_tools.bash` now raises `ModuleNotFoundError`. A shim submodule would shadow the package `__getattr__` and silently return a module object instead of the tool, which is a worse failure than a loud one. The documented import path `from strands.vended_tools import bash` keeps working with a warning.

### Docstring correction

`vended_tools/__init__.py` described the Python `bash` tool as running "a persistent shell on the host." That is true of the TypeScript tool but false here — `SANDBOX_BASH_DESCRIPTION` correctly said state "does not persist across calls." The text appears to have been copied across SDKs; corrected, along with the same conflation in `site/.../vended-tools.mdx`.

## Documentation PR

Docs are included in this PR under `site/`: the vended-tools and sandbox pages now distinguish the two tools, `sandbox_bash` → `sandbox_shell` throughout, and `sourceLinks` gained the renamed Python path (stale `sourceLinks` rot silently, so this matters).

## Type of Change

Breaking change

Strictly, only the vended tool *name* changes without opt-in: agents using a Docker/SSH sandbox will see `sandbox_shell` instead of `sandbox_bash` in their tool list. Import-level compatibility is preserved until v2.0.0.

## Testing

- **Python**: `hatch test` → **4686 passed**. Includes new tests asserting each deprecated alias emits `DeprecationWarning`, resolves to the canonical object, and that unknown attributes still raise `AttributeError`.
- **TypeScript**: `npm test` → **3891 passed, 141 files, no type errors**. Includes tests that `makeBash` is still the same factory as `makeShell` and builds a working tool.
- `npm run lint` and `npm run format` clean; `npx tsc --noEmit --project src/tsconfig.json` clean.
- Manually verified the shim end to end: canonical imports work, both aliases warn once and return the canonical objects, and the old deep module path fails loudly rather than silently.
- Updated `tests_integ/sandbox/test_ssh.py` and `test/integ/sandbox/ssh.test.node.ts`, which assert the vended tool name and would otherwise have failed in CI.

Not run: Docker/SSH integration tests requiring provisioned infrastructure.

Two pre-existing issues on `main`, unrelated to this change and left alone: `ruff format` wants to reformat `vended_tools/http_request/http_request.py`, and `site` snippet typecheck fails on a missing `pino` module in `observability-evaluation/logs.ts`. Both reproduce on a clean checkout.

## Checklist

On "focused and reasonably small": this spans `strands-py/`, `strands-ts/`, and `site/`, which normally suggests splitting. Here they are one logical change — the cross-SDK parity rule in `AGENTS.md` requires renaming both SDKs together, and splitting would leave a window where the two disagree. Happy to split if you would rather review them separately.

On new warnings: this intentionally *adds* `DeprecationWarning`s for the old names, which is the point of the shim. Nothing else warns.

----


yonib05:
Context: strands-agents/tools#550 deprecates that repo's `shell`, `editor`, and `sleep` tools in favor of the SDK's vended equivalents, and its migration message points at `strands.vended_tools.shell` — so it depends on this rename landing first.

site/src/content/docs/user-guide/concepts/tools/vended-tools.mdx
```diff
@@ -162,9 +163,13 @@ _Supported in: Node.js, browsers._
...
 
-### Bash
+### Bash / Shell
 
-Lets your agent run shell commands and act on the output. Shell state — variables, working directory, exported functions — persists across invocations within the same session, so the agent can build up context incrementally. Sessions can be restarted to clear state.
+Lets your agent run shell commands and act on the output. The two SDKs expose
+different tools here:
+
+- **TypeScript `bash`** spawns a persistent `bash` process on the host. Shell state — variables, working directory, exported functions — persists across invocations within the same session, so the agent can build up context incrementally. Sessions can be restarted to clear state.
+- **Python `shell`** (and TypeScript's `makeShell`) routes each command through the agent's [Sandbox](../sandbox/index.mdx) and is stateless: every call runs in a fresh shell, so variables and the working directory do not carry over. The sandbox decides the interpreter — `sh` locally and in Docker, the remote login shell over SSH — so commands must not rely on bash-specific syntax.
```
opieter-aws: I don't love that we're drifting Python and TS here, given that we're working on porting closing the parity gap. Does it make sense to rework the TS tool to a shell tool?

opieter-aws: Also I would fully remove bash from the docs here for python (remove `so commands must not rely on bash-specific syntax.`)

strands-ts/src/vended-tools/bash/make-shell.ts
```diff
@@ -0,0 +1,75 @@
...
+        throw new Error('Tool context is required for shell operations')
+      }
+
+      const sandbox = boundSandbox ?? context.agent.sandbox
+      try {
+        const result = await sandbox.execute(input.command, { timeout: input.timeout ?? 120 })
+        return { output: result.stdout, error: result.stderr } as BashOutput
+      } catch (err) {
+        if (err instanceof SandboxTimeoutError) throw new BashTimeoutError(err.message)
+        throw new BashSessionError((err as Error).message)
```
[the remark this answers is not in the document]
yonib05: Good catch — fixed in 65e94cc. Both SDKs now raise shell-named errors from the sandbox-routed path:

- TypeScript: `ShellTimeoutError` / `ShellExecutionError`
- Python: `ShellExecutionError` (it previously raised a bare `RuntimeError`, which was the cross-SDK divergence you flagged)

Each subclasses what was raised before — `Shell*` extends `Bash*`, and Python's extends `RuntimeError` — so existing `catch` / `except` clauses keep matching and nothing breaks for current callers. This follows the same pattern already used in `sandbox/errors.py`, where `SandboxTimeoutError` extends `TimeoutError` for exactly this reason.

The `Bash*` types stay as-is for the standalone `bash` tool, which genuinely runs a bash session.

site/src/content/docs/user-guide/concepts/tools/vended-tools.mdx
yonib05: Done in 65e94cc — dropped the trailing clause and reworded to "so use portable POSIX syntax", which says the same thing without naming bash. Also updated the security callout in that section, which still said "arbitrary bash commands" for what is now two different tools.

yonib05: Fair challenge, and I want to be precise about what is drifting, because I don't think this PR widens the gap — it documents one that already exists.

These are two different tools, not one tool that drifted:

- **TS `bash`** ([`bash.ts:39`](https://github.com/strands-agents/harness-sdk/blob/fe4cbb9486566154b1f94e3ea3c6a85a2bd81f43/strands-ts/src/vended-tools/bash/bash.ts#L39)) does `spawn('bash', [])` and holds a [`BashSession`](https://github.com/strands-agents/harness-sdk/blob/fe4cbb9486566154b1f94e3ea3c6a85a2bd81f43/strands-ts/src/vended-tools/bash/bash.ts#L19) across calls, with an `execute`/`restart` mode enum that only makes sense for a stateful session. No sandbox involved.
- **Python `shell`** and **TS `makeShell`** both route through `Sandbox.execute()` and are stateless.

So the sandbox-routed surface *is* at parity — this PR renames it in both SDKs in lockstep. The asymmetry is that TS ships an extra, genuinely-bash tool that Python has no equivalent of. That predates this PR; renaming `makeBash` → `makeShell` doesn't change it either way.

On reworking TS `bash` into a shell tool: I'd rather not fold that in here. It would be a real behavior break for anyone relying on session persistence (the `restart` mode and the "persists environment variables between calls" tests exist because users depend on it), and it's a product decision about whether the SDK should offer a stateful host-shell tool at all — separate from whether the sandbox-routed one is misnamed. The naming fix stands on its own and is safe to land now.

The parity gap I'd actually flag as worth closing is the inverse: **Python has no persistent-session equivalent of TS `bash`.** Happy to open an issue for that if you agree it's the real asymmetry — and if the answer turns out to be "TS shouldn't have it either", that's a deliberate removal with its own deprecation cycle rather than a rename.

opieter-aws (approved):

yonib05:
Applied the `FEATURE_LIFECYCLE.md` deprecation pattern to the shim in 6009a520 (prompted by the same feedback on strands-agents/tools#550).

`make_bash` now carries `@typing_extensions.deprecated`, so type checkers and IDEs flag callers instead of only warning at runtime. `bash` keeps the module `__getattr__` path — it is a tool *instance*, not a function, so no decorator can wrap it.

Behavior is unchanged for callers: both still warn once and resolve to the `shell` equivalents, and `make_bash(**kwargs)` forwards through to `make_shell`. Added tests covering argument forwarding and the `__deprecated__` marker.

Verified: 4689 Python tests pass, ruff and mypy clean.

yonib05:
Correction: my comment above referenced commit 6009a520, but this PR had already merged at 65e94cc — that commit never landed here. Moved to #3595 against current main.

PR #3576, opened 2026-07-30
fix(openai): send document content as file_data on the Responses API

### LINKED ISSUE #3572: [BUG] Responses API: document content sent in file_url instead of file_data (Azure returns 400)
### Checks

### SDK Language

Python

### Strands Version

1.50.2

### Language Runtime Version

Python 3.11

### Operating System

Linux (Debian, in Docker)

### Installation Method

pip

### Steps to Reproduce

`_format_request_message_content` is a classmethod, so the formatting half reproduces with no credentials:

```python

block = {
    "document": {
        "name": "contract",
        "format": "pdf",
        "source": {"bytes": b"%PDF-1.4 ..."},
    }
}
print(OpenAIResponsesModel._format_request_message_content(block))
# {'type': 'input_file', 'file_url': 'data:application/pdf;base64,JVBERi0xLjQgLi4u'}
```

Sending a message carrying that document block through an `OpenAIResponsesModel` pointed at Azure OpenAI (v1 surface, `gpt-5-mini` deployment) then fails:

```
openai.BadRequestError: Error code: 400 - {'error': {'message': 'Failed to download file.', 'type': 'invalid_request_error', 'param': 'url', 'code': 'invalid_value'}}
```

Azure is attempting to download the `data:` URI.

### Expected Behavior

Base64 document content belongs in `file_data`, with the document name in `filename`:

```python
{"type": "input_file", "filename": "contract", "file_data": "data:application/pdf;base64,..."}
```

Per the OpenAI SDK's `ResponseInputFileParam`, `file_url` is the URL of the file and `file_data` is the content of the file, with `filename` accompanying `file_data`.

### Actual Behavior

Base64 file content is placed in `file_url`, and `doc["name"]` is dropped. Both document-formatting sites in `strands-py/src/strands/models/openai_responses.py` do this, in 1.50.2 and on current `main` (a8ead94).

`_format_request_message_content`:

```python
if "document" in content:
    doc = content["document"]
    data_url = _encode_media_to_data_url(doc["source"]["bytes"], doc["format"], "document")
    return {"type": "input_file", "file_url": data_url}
```

and the inlined equivalent in `_format_request_tool_message`:

```python
elif "document" in content:
    has_media = True
    doc = content["document"]
    data_url = _encode_media_to_data_url(doc["source"]["bytes"], doc["format"], "document")
    output_parts.append({"type": "input_file", "file_url": data_url})
```

### Additional Context

Stock OpenAI appears to be more tolerant of a `data:` URI in `file_url`, which may be why this hasn't surfaced before; Azure OpenAI hard-fails. Sending the content as `file_data` works on both.

Beyond the Azure failure, the document name is lost on every provider, since neither Python site passes `filename`.

### Possible Solution

Emit `{"type": "input_file", "filename": doc["name"], "file_data": data_url}` at both sites. The repo already contains that shape in two other places:

- The chat-completions model, `strands-py/src/strands/models/openai.py`, sends `{"file": {"file_data": f"data:{mime_type};base64,{file_data}", "filename": content["document"]["name"]}, "type": "file"}`.
- The TypeScript Responses adapter, `strands-ts/src/models/openai/responses-adapter.ts`, sends `type: 'input_file'` with `file_data` and `filename`, in both `formatDocumentInput` and `formatToolResultOutput`.

Note that `strands-py/tests/strands/models/test_openai_responses.py` currently asserts the `file_url` shape, so those assertions need updating alongside the fix. The TypeScript tests in `strands-ts/src/models/openai/__tests__/responses.test.ts` already assert the correct shape.

### Related Issues

_No response_

tylerganter:
## Description

The Responses API model formats document ContentBlocks by putting the base64 file *content* (a `data:` URI) in the `file_url` field. Azure OpenAI treats `file_url` literally, attempts to download the URI, and rejects the request:

```
openai.BadRequestError: Error code: 400 - {'error': {'message': 'Failed to download file.',
'type': 'invalid_request_error', 'param': 'url', 'code': 'invalid_value'}}
```

Per the OpenAI SDK's `ResponseInputFileParam`, `file_data` is the content of the file (with `filename` accompanying it) and `file_url` is a URL to fetch. Both formatting sites — `_format_request_message_content` and `_format_request_tool_message` — now send `file_data` + `filename`, which also fixes the document's `name` being dropped entirely.

This brings the Python responses model into agreement with the two existing correct implementations in this repo:

- the Python chat-completions model (`strands-py/src/strands/models/openai.py`), which sends `file_data` + `filename`
- the TypeScript responses adapter (`strands-ts/src/models/openai/responses-adapter.ts`), which sends `file_data` + `filename` in both `formatDocumentInput` and `formatToolResultOutput`

Verified live against Azure OpenAI (v1 surface): a real PDF sent with the current shape fails with the 400 above; with `file_data` + `filename` the request returns 200 and the model reads the document content. Stock OpenAI appears to tolerate a data URI in `file_url` where Azure hard-fails, which is likely why this went unreported.

## Related Issues

Resolves #3572

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

Also verified end to end against Azure OpenAI as described above (400 before, 200 after).

## Checklist

----


tylerganter:
@pgrayy could I get some love on this PR? 🙏🙏🙏

opieter-aws:

strands-py/src/strands/models/openai_responses.py
```diff
@@ -709,7 +709,7 @@ def _format_request_message_content(cls, content: ContentBlock, *, role: Role =
         if "document" in content:
             doc = content["document"]
             data_url = _encode_media_to_data_url(doc["source"]["bytes"], doc["format"], "document")
-            return {"type": "input_file", "file_url": data_url}
+            return {"type": "input_file", "filename": doc["name"], "file_data": data_url}
```
strandly-the-agent: 🟡 `doc["name"]` hard-indexes a key that's optional: `DocumentContent` is `TypedDict, total=False` (`strands-py/src/strands/types/media.py:64`). A document block without a name formatted fine on `main`; on this branch it raises a bare `KeyError` from inside the provider:

```pycon
>>> OpenAIResponsesModel._format_request_message_content(
...     {"document": {"format": "pdf", "source": {"bytes": b"%PDF-1.4"}}})
KeyError: 'name'
```

That shape is in our own docs — `site/src/content/docs/user-guide/concepts/model-providers/google.mdx:425` is exactly `{"document": {"format": "pdf", "source": {"bytes": document_bytes}}}` — so someone copying it and swapping in `OpenAIResponsesModel` hits this, and a tool returning a nameless document block hits the mirror at `:782`.

Bounded rather than scary: nothing inside `strands-py` emits a nameless block, and `openai.py:174` / `llamacpp.py:216` / `anthropic.py:138` already hard-index `name` the same way (only `bedrock.py:621` guards it, with the comment *"all optional due to total=False"*). Still, it's a path that worked before this PR.

```suggestion
            return {"type": "input_file", "filename": doc.get("name", "document"), "file_data": data_url}
```

I ran this variant plus the `:782` mirror through `ruff check`/`format --check` and the full test file — clean, 118 passed. Do settle the filename-extension question in the summary before committing to `"document"` as the fallback string.

strands-py/src/strands/models/openai_responses.py
```diff
@@ -779,7 +779,7 @@ def _format_request_tool_message(cls, tool_result: ToolResult) -> dict[str, Any]
                 has_media = True
                 doc = content["document"]
                 data_url = _encode_media_to_data_url(doc["source"]["bytes"], doc["format"], "document")
-                output_parts.append({"type": "input_file", "file_url": data_url})
+                output_parts.append({"type": "input_file", "filename": doc["name"], "file_data": data_url})
```
strandly-the-agent: Mirror of the `KeyError` above, on the tool-result path (a tool returning a document `ContentBlock` with no `name`). Split across two lines because the single-line form is 123 chars, over the 120 limit:

```suggestion
                filename = doc.get("name", "document")
                output_parts.append({"type": "input_file", "filename": filename, "file_data": data_url})
```

strandly-the-agent (commented):
**Right fix — one guard worth adding, and one question that decides whether #3572 is actually closed.** `file_data` + `filename` is the correct shape per `ResponseInputFileParam`, it matches `openai.py:174` and the TS adapter (`responses-adapter.ts:297`, `:334`), and both formatting sites are covered.

Verified on `5992f27e6`: ✅ `pytest tests/strands/models/test_openai_responses.py -q` → **118 passed** · ✅ reproduced the new `KeyError` for a nameless document block on this branch and confirmed it doesn't happen on `main` · ✅ ran the suggested fix below through ruff + the full test file · 🔴 **not** verified — no live OpenAI/Azure call (no credentials on my side), so the question below rests on public evidence, not observation.

- 🟡 ×2 inline — `doc["name"]` hard-indexes a key that `DocumentContent` marks optional.

### Question (blocking-ish: it decides whether extension-less names still 400)

What was `document["name"]` in your live Azure test — did it end in `.pdf`? The Responses API looks like it detects file type from the **filename extension**, not from the MIME in the data URI: reported errors quote a dotted extension back, case-sensitively — [*"Expected context stuffing file type to be a supported format: .bat, .c, … .pdf, … .xml but got .xlsx"*](https://community.openai.com/t/responses-api-stopped-supporting-xlsx-files/1377172), and elsewhere *"but got .PDF"*. Strands' `document.name` is free text and frequently carries no extension — this PR's own fixture is `"test doc"`, #3572's example is `"contract"`, A2A deliberately strips the extension (`multiagent/a2a/executor.py:800`), and the context-offloader names documents like `mem_1_tool-123_0` (`vended_plugins/context_offloader/plugin.py:395`). If `name="contract"` also returns 200 for you, this is a non-issue and I'll drop it; if it 400s, `filename` wants to carry `doc["format"]` (and that normalization would want mirroring in the TS adapter, which passes `docBlock.name` raw too).

- **⚪ Coverage.** No test exercises a nameless document, before or after this PR (every fixture at `test_openai_responses.py:118`, `:254`, `:1527`, `:1550` sets `name`). I mutation-checked: `doc.get("name", "document")` at both sites still gives 118 passed, so the fix below would land with zero signal either way — worth one case per site.
- **⚪ Test convention.** Root `AGENTS.md` and `strands-py/docs/TESTING.md:82` ask a regression test to link the issue it guards; the file already does this at `:326-333` for `#3388`. Neither changed test mentions #3572.
- **⚪ Schema nuance, non-blocking.** The tool-result site targets `ResponseInputFileContentParam`, whose `file_data` is documented *"The base64-encoded data of the file"*, while the message site's `ResponseInputFileParam` says *"The content of the file"* (the image variants share identical wording, so the difference may be real rather than codegen noise). Possibly that path wants raw base64 without the `data:` prefix. Not a regression — `main` was at least as wrong — and TS does the same thing at both sites, so any change belongs in both SDKs. Did your live check cover the tool-result path, or only message content?
- **⚪ Pre-existing, not filing.** `_encode_media_to_data_url` emits `application/octet-stream` for `docx`/`xlsx` (Python's `mimetypes.types_map` lacks both; TS maps them). Same root cause as the open #2204, so no new issue from me — it just means the filename extension is the *only* type signal for those formats.

</details>

<sub>Automated review by `strandly-the-agent` — correctness, adversarial/repro, and test-quality passes, aggregated. Solid work for a human to approve; `pgrayy` still owns the call.</sub>

opieter-aws:
@tylerganter can you look into the findings and suggestions from the review bot? The hardening suggestions are preferred, and please confirm the document extension question. Thanks!

strands-py/src/strands/models/openai_responses.py
tylerganter: Applied in a85826ea — both sites now use `doc.get("name", "document")`, each covered by a new nameless-document test case, and the changed tests now reference #3572. The filename-extension question is answered in the main thread — extension-less names work, so the fallback string is fine as-is.

strands-py/src/strands/models/openai_responses.py
tylerganter: Applied in a85826ea using the suggested two-line form, covered by a nameless document block in `test_format_request_tool_message_with_document`.

tylerganter:
> @tylerganter can you look into the findings and suggestions from the review bot? The hardening suggestions are preferred, and please confirm the document extension question. Thanks!

@opieter-aws Hardening suggestions applied in a85826ea (plus latest main merged in). Also re-verified live against Azure OpenAI (v1 surface, gpt-5-mini) with a generated PDF, including a real two-step tool round trip:

- **Extension question**: `filename="contract"` (no extension) returns 200 and the model reads the document, both as message content and on the tool-result path — so no extension normalization is needed in either SDK. The community reports quoting extensions back appear to be about unsupported formats (e.g. `.xlsx`), not missing extensions.
- **Tool-result path**: now verified live. `function_call_output` → `input_file` with a data URI in `file_data` returns 200 and the model reads the document. Raw base64 without the `data:` prefix is rejected with 400 `Invalid 'file_data'`, so the data-URI encoding this PR sends is correct at both sites.
- Controls: the old `file_url` shape reproduces the original 400 ("Failed to download file"); the fixed shape works.

<details>
<summary>Verification script (<code>azure_doc_test.py</code>) and its output</summary>

```python
"""Live Azure OpenAI matrix for PR #3576 (Responses API document handling).

Answers the two open review questions on
https://github.com/strands-agents/harness-sdk/pull/3576:

1. Filename extension — is the file type inferred from the ``filename``
   extension, so an extension-less name (the common Strands case) still 400s?
2. Tool-result path — does ``function_call_output`` with an ``input_file``
   part accept a data URI in ``file_data``, or does it want raw base64?

Also re-proves the original #3572 bug (data URI in ``file_url`` -> 400) as a
red control, and the PR's fixed shape as a green control.

Setup: set AZURE_OPENAI_ENDPOINT / AZURE_OPENAI_API_KEY / AZURE_OPENAI_MODEL
(v1 surface), then run with any Python that has the ``openai`` package, e.g.:

    uv run --with openai python azure_doc_test.py
"""

import base64
import os
import sys

from openai import OpenAI

MAGIC_WORD = "PINEAPPLE"
MAX_OUTPUT_TOKENS = 4096


def make_pdf(text: str) -> bytes:
    """Build a minimal one-page PDF with real extractable text."""
    content = f"BT /F1 18 Tf 72 720 Td ({text}) Tj ET".encode()
    objects = [
        b"<< /Type /Catalog /Pages 2 0 R >>",
        b"<< /Type /Pages /Kids [3 0 R] /Count 1 >>",
        b"<< /Type /Page /Parent 2 0 R /MediaBox [0 0 612 792] "
        b"/Resources << /Font << /F1 4 0 R >> >> /Contents 5 0 R >>",
        b"<< /Type /Font /Subtype /Type1 /BaseFont /Helvetica >>",
        b"<< /Length %d >>\nstream\n%s\nendstream" % (len(content), content),
    ]
    out = b"%PDF-1.4\n"
    offsets = []
    for index, obj in enumerate(objects, start=1):
        offsets.append(len(out))
        out += b"%d 0 obj\n%s\nendobj\n" % (index, obj)
    xref_pos = len(out)
    out += b"xref\n0 %d\n0000000000 65535 f \n" % (len(objects) + 1)
    out += b"".join(b"%010d 00000 n \n" % offset for offset in offsets)
    out += b"trailer\n<< /Size %d /Root 1 0 R >>\nstartxref\n%d\n%%%%EOF\n" % (len(objects) + 1, xref_pos)
    return out


PDF_BYTES = make_pdf(f"The magic word is {MAGIC_WORD}")
PDF_B64 = base64.b64encode(PDF_BYTES).decode()
DATA_URL = f"data:application/pdf;base64,{PDF_B64}"
QUESTION = "What is the magic word in the document? Answer with just the word."


def report(name: str, outcome: str, detail: str) -> None:
    print(f"\n=== {name}: {outcome}\n    {detail}")


def run_message_test(client: OpenAI, model: str, name: str, file_part: dict) -> None:
    """One-shot request with the document attached as user message content."""
    try:
        response = client.responses.create(
            model=model,
            max_output_tokens=MAX_OUTPUT_TOKENS,
            input=[
                {
                    "role": "user",
                    "content": [{"type": "input_text", "text": QUESTION}, file_part],
                }
            ],
        )
    except Exception as error:  # noqa: BLE001 - report every failure shape
        report(name, "HTTP ERROR", repr(error))
        return
    text = response.output_text.strip()
    outcome = "200, word extracted" if MAGIC_WORD.lower() in text.lower() else "200 but word NOT found"
    report(name, outcome, f"model said: {text[:200]!r}")


def run_tool_test(client: OpenAI, model: str, name: str, file_part: dict) -> None:
    """Real two-step tool round trip: model calls a tool, tool returns the document."""
    tools = [
        {
            "type": "function",
            "name": "fetch_document",
            "description": "Fetch the document the user is asking about.",
            "parameters": {"type": "object", "properties": {}, "additionalProperties": False},
        }
    ]
    input_items = [
        {
            "role": "user",
            "content": [
                {
                    "type": "input_text",
                    "text": "Call the fetch_document tool, then tell me the magic word "
                    "in the returned document. Answer with just the word.",
                }
            ],
        }
    ]
    try:
        first = client.responses.create(
            model=model, max_output_tokens=MAX_OUTPUT_TOKENS, input=input_items, tools=tools
        )
    except Exception as error:  # noqa: BLE001
        report(name, "HTTP ERROR (step 1, no document involved)", repr(error))
        return

    call = next((item for item in first.output if item.type == "function_call"), None)
    if call is None:
        report(name, "INCONCLUSIVE", f"model never called the tool: {first.output_text[:200]!r}")
        return

    follow_up = (
        input_items
        + list(first.output)
        + [{"type": "function_call_output", "call_id": call.call_id, "output": [file_part]}]
    )
    try:
        second = client.responses.create(model=model, max_output_tokens=MAX_OUTPUT_TOKENS, input=follow_up, tools=tools)
    except Exception as error:  # noqa: BLE001
        report(name, "HTTP ERROR", repr(error))
        return
    text = second.output_text.strip()
    outcome = "200, word extracted" if MAGIC_WORD.lower() in text.lower() else "200 but word NOT found"
    report(name, outcome, f"model said: {text[:200]!r}")


def main() -> None:
    endpoint = os.environ.get("AZURE_OPENAI_ENDPOINT")
    api_key = os.environ.get("AZURE_OPENAI_API_KEY")
    model = os.environ.get("AZURE_OPENAI_MODEL", "gpt-5-mini")
    if not endpoint or not api_key:
        sys.exit("Set AZURE_OPENAI_ENDPOINT and AZURE_OPENAI_API_KEY first.")

    client = OpenAI(base_url=endpoint, api_key=api_key)
    print(f"model={model} pdf_bytes={len(PDF_BYTES)}")

    # Red control — the pre-PR shape (#3572): expect 400 "Failed to download file".
    run_message_test(
        client, model, "A. message, OLD file_url shape (expect 400)", {"type": "input_file", "file_url": DATA_URL}
    )
    # Green control — the PR's shape with a proper extension.
    run_message_test(
        client,
        model,
        "B. message, file_data + filename WITH .pdf extension",
        {"type": "input_file", "filename": "contract.pdf", "file_data": DATA_URL},
    )
    # Q1 — extension-less filename, the common Strands case.
    run_message_test(
        client,
        model,
        "C. message, file_data + filename WITHOUT extension (Q1)",
        {"type": "input_file", "filename": "contract", "file_data": DATA_URL},
    )
    # Q2 — tool-result path with the PR's shape (data URI in file_data).
    run_tool_test(
        client,
        model,
        "D. tool result, file_data as data URI (Q2)",
        {"type": "input_file", "filename": "contract.pdf", "file_data": DATA_URL},
    )
    # Q2 follow-up — raw base64 (no data: prefix), the ResponseInputFileContentParam reading.
    run_tool_test(
        client,
        model,
        "E. tool result, file_data as raw base64",
        {"type": "input_file", "filename": "contract.pdf", "file_data": PDF_B64},
    )
    # Q1 x Q2 — extension-less filename on the tool path.
    run_tool_test(
        client,
        model,
        "F. tool result, filename WITHOUT extension",
        {"type": "input_file", "filename": "contract", "file_data": DATA_URL},
    )


if __name__ == "__main__":
    main()
```

Output against Azure OpenAI (v1 surface, gpt-5-mini deployment):

```
model=gpt-5-mini pdf_bytes=602

=== A. message, OLD file_url shape (expect 400): HTTP ERROR
    BadRequestError("Error code: 400 - {'error': {'message': 'Failed to download file.', 'type': 'invalid_request_error', 'param': 'url', 'code': 'invalid_value'}}")

=== B. message, file_data + filename WITH .pdf extension: 200, word extracted
    model said: 'PINEAPPLE'

=== C. message, file_data + filename WITHOUT extension (Q1): 200, word extracted
    model said: 'PINEAPPLE'

=== D. tool result, file_data as data URI (Q2): 200, word extracted
    model said: 'PINEAPPLE'

=== E. tool result, file_data as raw base64: HTTP ERROR
    BadRequestError('Error code: 400 - {\'error\': {\'message\': "Invalid \'input[3].output[0].file_data\'.", \'type\': \'invalid_request_error\', \'param\': \'input[3].output[0].file_data\', \'code\': \'invalid_value\'}}')

=== F. tool result, filename WITHOUT extension: 200, word extracted
    model said: 'PINEAPPLE'
```

</details>

JackYPCOnline (approved):

Issue #1066, opened 2025-10-22
[FEATURE] The Bedrock documentation source can only accept bytes.


HawkClaws:
### Checks

- [x] I have updated to the lastest minor and patch version of Strands
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched [./issues](./issues?q=) and there are no duplicates of my issue

### Strands Version

1.13.0

### Python Version

3.11.13

### Operating System

Windows 11

### Installation Method

pip

### Steps to Reproduce

Passing non-bytes to source

```
model = BedrockModel()

content = [
    {"text": "猫の嗅覚は優れてる？"},
    {
        "document": {
            "format": "txt",
            "name": "neko",
            "source": {
                "text": neko_source
            },
            "citations": {"enabled": True}
        }
    }
]
messages = [
    {
        "role":"user",
        "content":content
    }
]


model.stream(messages=messages)
```

### Expected Behavior

Allow passing other types besides bytes

### Actual Behavior

The sample passes text, which is supported in Bedrock and is valid format. However, it causes an error.

### Possible Solution

https://github.com/strands-agents/sdk-python/blob/78c59b95ffa2b50a8e1dc93e3cdd172772b0b791/src/strands/models/bedrock.py#L386

Just
`result["source"] = {"bytes": document["source"]["bytes"]}`
to
`result["source"] = document["source"]`

pgrayy:
https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_DocumentSource.html
^ Looks like Bedrock recently updated the `DocumentSource` definition. `text` and `content` fields are now supported. We will need to update the `DocumentSource` type definition in Strands ([src](https://github.com/strands-agents/sdk-python/blob/main/src/strands/types/media.py])) and update the model providers accordingly.

Ratish1:
Hi, I'd like to contribute to this issue.

wu-theodore:
Opened #4343 to add text document-source support — closes this. Would appreciate a review when someone has time.

PR #1572, opened 2026-01-27
Feature: Allow s3Location as Document, Image, and Video location source

Unshure:
## Description
Adds native S3 location support for media content types in Bedrock, allowing users to reference images, documents, and videos stored in S3 directly rather than requiring base64 encoding.

One decision I made in this pr is the choice between a generic `Location` source, vs the more specific `S3Location` source. Some model providers support url for a document location ([ref](https://platform.openai.com/docs/api-reference/responses/create#responses_create-input-input_item_list-item-input_message-content-input_file-file_url)), so it could make sense to make this generic for s3 as well. But since S3 also has an optional `bucketOwner`, I thought it would make more sense to have a separate `S3Location`, and then later add something like a `URLLocation` to cover the openai case.

#### Type System Updates
- Extended `ImageSource`, `DocumentSource`, and `VideoSource` in `types/media.py` to accept `s3Location` as an alternative to `bytes`
- Added `S3Location` TypedDict with required `uri` field and optional `bucketOwner` for cross-account access

#### Provider Behavior
- **Bedrock**: Passes S3 locations through unchanged to the API
- **Non-Bedrock providers**: These will be addressed in a follow-up pr because this one would be huge otherwise

#### Updated Integ Tests
I added integ tests which create an s3 bucket (post-fixed with the accountId) where the local file resources are uploaded for testing. Then they are referenced in the tests to ensure the model can recognize and identify the content in them.

Ive also added a short 5 second video resource to the integ tests to cover video modalities

#### Usage Example

```python
from strands import Agent
from strands.models.bedrock import BedrockModel

agent = Agent(model=BedrockModel())

# Reference an S3-hosted document directly
response = agent(
    [
        {
            "role": "user",
            "content": [
                {"text": "Summarize this document:"},
                {
                    "document": {
                        "format": "pdf",
                        "name": "report.pdf",
                        "source": {
                            "s3Location": {
                                "uri": "s3://my-bucket/documents/report.pdf",
                                "bucketOwner": "123456789012"  # Optional
                            }
                        }
                    }
                }
            ]
        }
    ]
)
```

**Notes on backwards compatibility:**
This change updates `ImageSource`, `DocumentSource`, and `VideoSource` to use `total=False` in their typed dict definition, ultimately making the `bytes` parameter optional now when it was required before. This is considered safe since all existing code should not be impacted by this. However, going forward, since these fields are essentially union types now, they need to be treated as such by customers going forward.

## Related Issues

#1482

## Documentation PR

TODO

[...]

PR #1225, opened 2025-11-21
feat(bedrock): allow text and content in Bedrock document sources

Ratish1:
## Description
 
This PR resolves an issue where providing a document source as `text` to the `BedrockModel` would cause an error because bedrock updated their `DocumentSource` definition https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_DocumentSource.html .This PR updates the `DocumentSource` type and modifies the request formatting to correctly pass through any valid source type (`bytes`, `text`, `content`). Unit tests were added to verify the format request with bytes, text  and content,
## Related Issues

Closes #1066 

## Documentation PR 

 N/A

## Type of Change

New feature

## Testing

How have you tested the change?  Verify that the changes do not break functionality or introduce warnings in consuming repositories: agents-docs, agents-tools, agents-cli

- [x] I ran `hatch run prepare`
- [x] I ran formatters, linters and unit tests 

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [ ] I have updated the documentation accordingly
- [ ] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3582, opened 2026-07-31
feat(bedrock): Support content and text DocumentSource

jonathankeys:
## Description
Bedrock ConverseStream requests allow message [Content](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Message.html#bedrock-Type-runtime_Message-content) to include the [DocumentBlock](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_DocumentBlock.html) which has the fields `text` and `content` in [DocumentSource](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_DocumentSource.html) which is missing in the `BedrockModel` message formatting. This change enables the passing of these fields so features (e.g. citations) can utilize these document types.

_note_: mypy was raising linter errors due to `source` variable in `_format_request_message_content(...)` being used for different object types, so those variable names were updated to satisfy mypy and add clarity for the type of source in the block. 

changes:
 - Add handling for the content and text DocumentSource fields
 - Add media object for DocumentBlockContent

### Other Notes

Testing done using model Claude Sonnet 4.6

When `content` or `text` sources are passed in, `citations` must also be enabled in the `document` field. This is not documented anywhere from what I have seen and unsure if this is coming from client side botocore validation or from AWS itself.

e.g.
```
{
    "document": {
        "format": "txt",
        "name": "document",
        "source": {
            "content": [{ "text": "Paris is the capital of France." }, { "text": "The capital of Germany is Berlin." }]
        },
        "citations": { "enabled": True}
    },
}
```

when `citations` is disabled or not passed in at all, botocore raises a Validation exception

```
botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the ConverseStream operation: DocumentSource object at messages.0.content.1.document.source must set one of the following keys: bytes, s3Location.
```

However, when `citations` is enabled, it works correctly.

```
messages: list[Message] = [
    {
        "role": "user",
        "content": [
            {
                "text": "What is the capital of France?"
            },
            {
                "document": {
                    "format": "txt",
                    "name": "document",
                    "source": {
                        "content": [{ "text": "Paris is the capital of France." }, { "text": "The capital of Germany is Berlin." }]
                    },
                    "citations": { "enabled": True}
                },
            }
        ]
    }
]
```

```
{
  "citation": {
    "title": "document",
    "sourceContent": [
      {
        "text": "Paris is the capital of France."
      }
    ],
    "location": {
      "documentChunk": {
        "documentIndex": 0,
        "start": 0,
        "end": 1
      }
    }
  },
  "delta": {
    "citation": {
      "title": "document",
      "sourceContent": [
        {
          "text": "Paris is the capital of France."
        }
      ],
      "location": {
        "documentChunk": {
          "documentIndex": 0,
          "start": 0,
          "end": 1
        }
      }
    }
  },
  "agent": "<strands.agent.agent.Agent object>",
  "event_loop_cycle_id": "024ff76b-cd42-4f78-bf69-d9929c3cf101",
  "request_state": {},
  "event_loop_cycle_trace": "<strands.telemetry.metrics.Trace object>",
  "event_loop_cycle_span": "NonRecordingSpan(trace_id=0x0, span_id=0x0)"
}
```

## Related Issues

#1066 

## Documentation PR

N/A

## Type of Change

New feature

## Testing

## Checklist

----


arielnabavian:
@strandly-the-agent review

strands-py/src/strands/types/media.py
```diff
@@ -47,18 +47,32 @@ class S3Location(Location, total=False):
...
         bytes: The binary content of the document.
+        content: List of content blocks.
         location: Location of the document.
+        text: Text contents of the document.
     """
 
     bytes: bytes
+    content: list[DocumentBlockContent]
     location: SourceLocation
+    text: str
```
arielnabavian: **concern:** `DocumentSource` is the shared, provider-agnostic type, but only Bedrock learns to read `text` and `content`. Every other provider reaches straight for `["bytes"]`, so a document that works on Bedrock raises a bare `KeyError: 'bytes'` elsewhere:

- `models/anthropic.py:131`
- `models/openai.py:170` — and `models/litellm.py:123` falls through to it
- `models/gemini.py:164`
- `models/llamacpp.py:212`
- `models/openai_responses.py:711` and `:781`

There is already a pattern in the repo for this exact problem: `_has_location_source` (`models/_validation.py:76`) plus a warn-and-skip at each provider (`anthropic.py:211`, `openai.py:412`, `gemini.py:248`, and five more). Widening that check to cover unsupported *source variants* rather than just `location` would make these providers skip the block with a clear message instead of dying on a `KeyError`.

Worth deciding either way before this merges: if the answer is "text/content are Bedrock-only for now", the guard is what encodes that.

strands-py/src/strands/types/media.py
```diff
@@ -47,18 +47,32 @@ class S3Location(Location, total=False):
...
+    """
+
+    text: str
+
+
 class DocumentSource(TypedDict, total=False):
     """Contains the content of a document.
 
-    Only one of `bytes` or `s3Location` should be specified.
+    Only one of `bytes`, `content`, `location`, or `text` should be specified.
```
arielnabavian: **concern:** the constraint you found — Bedrock rejects `text` and `content` sources unless `citations.enabled` is true — is recorded only in the PR description. Nothing in the code, this docstring, or the docs says it, and the docs checkbox is unchecked with "Documentation PR: N/A".

That matters more than a normal doc gap because the error Bedrock returns sends the user the wrong way:

```
DocumentSource object at messages.0.content.1.document.source
must set one of the following keys: bytes, s3Location
```

Someone reading that will conclude `text` is unsupported and give up, when the actual fix is to enable citations. Please capture it here in the docstring, and consider a warning from `BedrockModel` when `text`/`content` is used without citations enabled so the failure explains itself.

strands-py/src/strands/types/media.py
```diff
@@ -47,18 +47,32 @@ class S3Location(Location, total=False):
 SourceLocation: TypeAlias = Location | S3Location
 
 
+class DocumentBlockContent(TypedDict, total=False):
```
arielnabavian: **consider:** AWS names this type [`DocumentContentBlock`](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_DocumentContentBlock.html) — the last two words are swapped here. This module's header says these types "are modeled after the Bedrock API", and it is importable from `strands.types.media`, so renaming after release is a breaking change. Near-free to fix now.

Separately: per `team/API_BAR_RAISING.md`, a new public type plus a widened public contract normally wants the `api/needs-review` label, which is not on the PR.

strands-py/src/strands/models/bedrock.py
```diff
@@ -636,16 +636,22 @@ def _format_request_message_content(self, content: ContentBlock) -> dict[str, An
...
-                    formatted_document_source = {"bytes": source["bytes"]}
+                elif "bytes" in document_source:
+                    formatted_document_source = {"bytes": document_source["bytes"]}
+                elif "text" in document_source:
+                    formatted_document_source = {"text": document_source["text"]}
+                elif "content" in document_source:
+                    formatted_document_source = {
+                        "content": [{"text": item["text"]} for item in document_source["content"] if "text" in item]
+                    }
                 result["source"] = formatted_document_source
```
arielnabavian: **concern:** with the `= None` initialiser on line 642, a `source` that matches none of the four branches now sends `"source": None` to Bedrock rather than failing locally. Two inputs reach it: `{"source": {}}`, and `{"source": {"s3Location": {...}}}` — the latter being what the docstring on this type told people to write until this PR ("Only one of `bytes` or `s3Location`"), and what `test_media.py::test_document_source_with_s3_location` still constructs.

`_handle_location` at line 602 already models the better behaviour for unsupported input: warn, return `None`, let the caller drop the block. Doing the same here — or only assigning `result["source"]` when a source actually got formatted — keeps that consistent. As written the user gets a Bedrock `ValidationException` with nothing identifying which content block was malformed.

No test covers this branch, so whichever behaviour you pick is currently unpinned.

strands-py/src/strands/models/bedrock.py
```diff
@@ -636,16 +636,22 @@ def _format_request_message_content(self, content: ContentBlock) -> dict[str, An
...
                         return None
-                elif "bytes" in source:
-                    formatted_document_source = {"bytes": source["bytes"]}
+                elif "bytes" in document_source:
+                    formatted_document_source = {"bytes": document_source["bytes"]}
+                elif "text" in document_source:
+                    formatted_document_source = {"text": document_source["text"]}
+                elif "content" in document_source:
+                    formatted_document_source = {
+                        "content": [{"text": item["text"]} for item in document_source["content"] if "text" in item]
```
arielnabavian: **consider:** items without `text` are dropped with no signal. Two consequences worth weighing:

- If *every* item lacks `text`, this sends `{"content": []}` and Bedrock rejects it — again with nothing pointing back to this line.
- Because filtering happens before the request, the `documentChunk` indices Bedrock returns refer to the filtered list, not the list the caller passed. For a feature whose whole point is citations, silently renumbering the blocks is the kind of thing that gets debugged twice.

Restricting to `text` is correct — that is all `DocumentContentBlock` carries today — so this is about doing it loudly rather than not doing it. A `logger.warning` when something is dropped would cover it.

The added test pins partial filtering; the all-items-dropped case is the one that produces a confusing failure and is untested.

strands-py/src/strands/types/media.py
strandly-the-agent, line 50: 🔴 **This new public type's name diverges from TS and from the Bedrock wire shape — it should be `DocumentContentBlock`.**

TS already exports `DocumentContentBlock` (`strands-ts/src/types/media.ts:373`, in the public barrel), botocore's wire shape is `DocumentContentBlock`, and the root AGENTS.md parity rule requires identifiers that exist in both SDKs to stay in sync. Anyone importing the type is affected, and post-release the fix becomes a breaking rename or an alias-and-deprecation dance.

Suggestion (plus the reference at line 73 and the uses in `test_media.py`):

```suggestion
class DocumentContentBlock(TypedDict, total=False):
```

strands-py/src/strands/types/media.py
```diff
@@ -47,18 +47,32 @@ class S3Location(Location, total=False):
...
     """Contains the content of a document.
 
-    Only one of `bytes` or `s3Location` should be specified.
+    Only one of `bytes`, `content`, `location`, or `text` should be specified.
 
     Attributes:
         bytes: The binary content of the document.
+        content: List of content blocks.
         location: Location of the document.
+        text: Text contents of the document.
```
strandly-the-agent, line 69: 🟡 **A `text`/`content` source is rejected by Bedrock unless the same document block also sets `citations.enabled` — and nothing in the type, the formatter, or the docs says so.**

Verified live (us-west-2, Claude Sonnet 4.5):
```
source={"text": "..."}              -> ValidationException: ...must set one of the
                                       following keys: bytes, s3Location
same + citations={"enabled": True}  -> success
```
The error never mentions citations, so it reads like the field was silently dropped. Anyone using the new variants without already knowing the coupling hits this — including the exact shape the new unit test asserts.

Suggestion: document it on both attributes and add a text-source example to `amazon-bedrock.mdx`. Cost of going further: a hard guard isn't safe — no AWS doc states the rule and only one model was tested, so it could reject future-valid requests.

```suggestion
        bytes: The binary content of the document.
        content: List of content blocks. Bedrock requires `citations.enabled` on the
            document block when this source is used.
        location: Location of the document.
        text: Text contents of the document. Bedrock requires `citations.enabled` on
            the document block when this source is used.
```

strands-py/src/strands/types/media.py
strandly-the-agent, line 75: 🟡 **The type widening is global but the implementation is Bedrock-only — 8 other providers crash with a bare `KeyError: 'bytes'` on the variants this makes type-legal.**

Verified at runtime: OpenAI and Anthropic raise `KeyError: 'bytes'` for a `text`/`content` source, while a `location` source warns and skips — because #1572 shipped a shared gate (`models/_validation.py:76`) into 8 providers when it widened this same type. Reached by any tool returning a text-source document (documents ride in tool results) or a Bedrock-recorded conversation resumed on another provider.

Suggestion: generalize that gate — e.g. `_has_unsupported_source(content, supported={"bytes"})` — one line at each of the 8 existing call sites, plus the matching per-provider filter test (~40 lines, mirroring #1572). Cost: grows the PR; shipping Bedrock-only with the follow-up filed and named in the PR body is a fine alternative.

strands-py/src/strands/models/bedrock.py
```diff
@@ -636,16 +636,22 @@ def _format_request_message_content(self, content: ContentBlock) -> dict[str, An
...
+                    }
```
strandly-the-agent, line 654: 🟡 **An all-filtered `content` list is forwarded as `{"content": []}`, which Bedrock rejects — after a paid round trip.**

```
{"content": [{}]} or {"content": []}  ->  {"content": []}
  -> ValidationException: ...document.source.content.content: Field required
```
botocore's client-side validation passes it, so the SDK builds an invalid request the caller only learns about from the service. Hit by callers that chunk a document and end up with zero usable blocks (empty file, everything filtered). This is the case you said had value on #2135 — and nothing pins today's shape: with the fix below, all 195 tests still pass.

Suggestion: skip the block when the filtered list is empty, mirroring `_handle_location`. Cost: if the document was the message's only content, the whole message is dropped and Bedrock 400s with "A conversation must start with a user message" — pre-existing to the skip mechanism, identical for `location`. Tightening the block type to `text: Required[str]` (file-local precedent: `S3Location.uri`) would also keep `{}` out upstream.

```suggestion
                elif "content" in document_source:
                    formatted_content = [
                        {"text": item["text"]} for item in document_source["content"] if "text" in item
                    ]
                    if not formatted_content:
                        logger.warning("document content source has no text blocks | skipping content block")
                        return None
                    formatted_document_source = {"content": formatted_content}
```

strands-py/tests/strands/models/test_bedrock.py
```diff
@@ -2390,6 +2390,58 @@ def test_format_request_filters_document_content_blocks(model, model_id):
...
+            "content": [
+                {
+                    "document": {
+                        "name": "doc.txt",
+                        "format": "txt",
+                        "source": {
+                            "content": [
+                                {"text": "block one"},
+                                {"text": "block two"},
+                                {},  # This should be filtered out
```
strandly-the-agent, line 2430: 🟡 **Two behaviors this diff introduces survive one-line mutations of the code that covers them.**

- Deleting the new `= None` default leaves all 195 tests green — yet that's the diff's only behavior change to *existing* input: `source={}` raised `UnboundLocalError` on `main` and now yields `"source": None`.
- Replacing `{"text": item["text"]}` with raw `item` also stays green: no test input carries an extra key, so the field-stripping this test exists to verify isn't verified.

Suggestion: add the extra key here (verified: passes as-is, fails under the mutation), plus one test pinning the intended output for an unrecognized source. Cost: the `"source": None` shape was debated on #2135 and never settled, so pinning it is also a decision — see Questions in the summary.

```suggestion
                                {"text": "block two", "extraField": "should be removed"},
                                {},  # This should be filtered out
```

strandly-the-agent (changes requested):
**Verdict: request changes** — correct, minimal Bedrock pass-through, but the new public type carries the wrong name for a permanent API, and the feature doesn't work on its own documented path without `citations.enabled`.

The shape of the change is right: the field names match botocore's `DocumentSource` union, the `source` → `document_source` renames are genuinely mypy-forced (verified, not drive-by), and there's no scope creep. Three things to settle before merge:

- 🔴 **`DocumentBlockContent` should be `DocumentContentBlock`** — TS, botocore, and the root AGENTS.md parity rule all say so. One word now; a breaking rename later.
- 🟡 **The `citations.enabled` coupling is undocumented** — a `text` source is rejected with an error naming `bytes, s3Location`, which points nowhere near the real cause.
- 🟡 **The type widening is global, the implementation is Bedrock-only** — 8 providers `KeyError` on the newly-legal variants, where #1572 shipped a shared gate for `location`. Land the gate here or ship Bedrock-only-and-say-so; both are fine, but neither-nor isn't.

Process note: this adds a public type + 2 public fields but carries no `api/needs-review` label — the label workflow only enforces API review once that label is present, so the gate is currently being skipped rather than passed.

5 inline comments (1 🔴, 4 🟡). Evidence, questions, reading order, and non-blocking notes are collapsed below. This is bot review — the two design calls (provider scope, `text` semantics) deserve a human's judgment.

<details>
<summary>Evidence ledger — what was actually checked</summary>

Reviewed `pr-3582` @ `241ab2de` (merge-base `f621b82c9`), 4 files, +127/−18, all `strands-py/`.

- ✅ `PYTHONPATH=src pytest tests/strands/models/test_bedrock.py tests/strands/types/test_media.py -q` → **195 passed** (183 + 12)
- ✅ **Live Bedrock** (`bedrock-runtime` us-west-2, Claude Sonnet 4.5): `{"text": …}` and `{"content": [{"text": …}]}` → `ValidationException … must set one of the following keys: bytes, s3Location`; the same request **+ `citations={"enabled": true}` → success**; `{"bytes": …}` control → success
- ✅ **Live Bedrock**: `{"content": []}` → `ValidationException … document.source.content.content: Field required` (botocore's client-side validation passes it, so it costs a real round trip)
- ✅ **Runtime, providers**: OpenAI and Anthropic raise bare `KeyError: 'bytes'` on `text` and `content` sources, in *both* the message and tool-result paths; a `location` source warns and skips (message path) but also `KeyError`s in the tool-result path — pre-existing
- ✅ **Formatter matrix** re-run on the branch: `{}`/unknown key → `"source": None` (was `UnboundLocalError` on `main`); `{"content": [{}]}` and `{"content": []}` → `{"content": []}`; precedence is `location > bytes > text > content`
- ✅ **Mutation testing**, 4 one-line mutations of the diff's own code, each applied and reverted → **195 pass every time** (`= None` deleted; raw `item` instead of `{"text": item["text"]}`; precedence flipped; both new `DocumentSource` fields deleted)
- ✅ **Both suggestion blocks applied and run**: skip-when-empty + the tightened test → 195 pass, and `{"content": [{}]}` now drops the block instead of emitting `{"content": []}`
- ✅ **Parity**: `DocumentContentBlock` at `strands-ts/src/types/media.ts:373`, exported in the public barrel; TS's `_formatDocumentSource` encodes `text` → bytes, so TS's `text` works *without* citations while Python's needs it
- ✅ **Wire truth**: botocore 1.43.62 `bedrock-runtime` `DocumentSource` union = `bytes | s3Location | text | content`; `DocumentContentBlock` is a single-`text`-member union → the PR's field names are correct
- ✅ `_has_location_source` (`models/_validation.py:76`) has 8 provider call sites; no equivalent exists for `text`/`content`
- 🔴 **Not verified here**: `mypy` (INTERNAL ERROR on `media.py` in this sandbox with 1.20.2 and 2.3.1) and `hatch test`. The "renames were needed for mypy" claim was checked indirectly — reverting them on a scratch copy reproduces `Incompatible types in assignment` at `bedrock.py:677`/`:756`, and `main` is clean. It checks out.
- ⚠️ **Withdrawn during aggregation**: an earlier pass claimed the new content test *pins* the `{"content": []}` behavior. It doesn't — the fix keeps all 195 green. The shape is simply untested.
</details>

<details>
<summary>Questions (2 blocking, 5 not)</summary>

**Blocking — these decide the shape of the fix**
1. `pgrayy` on #1066 said "update the `DocumentSource` type definition … **and update the model providers accordingly**". Is the intent that non-Bedrock providers degrade gracefully like `_has_location_source` does for `location` — in this PR, or as a fast-follow with the scope filed? Shipping with neither doesn't match either reading.
2. Which `text` behavior is intended? Python forwards `text` verbatim (needs `citations.enabled`); TS encodes it to bytes and works without (`strands-ts/src/models/bedrock.ts:1242`). Same SDK input, two different wire requests — one of the two is wrong, and it's worth deciding here rather than after both ship.

**Non-blocking**
3. Where should the citations coupling live — attribute docstrings, an `amazon-bedrock.mdx` example, or a `logger.warning`? (The PR's docs boxes are unchecked with "Documentation PR: N/A".)
4. Is `citations.enabled` required by the service generally, or model-specific? Only Sonnet was tested and no AWS doc states the rule — the main reason not to hard-guard it.
5. What *should* an empty or unrecognized source produce — `"source": None`, omit the key, or raise? This was raised on #2135 and never settled; TS throws (`media.ts:494`). Worth deciding now that a test would pin it.
6. #1225 also carries `Closes #1066` and describes this PR's shape in its review thread. Should it be closed as superseded?
7. Does `notowen333`'s "make the same fix in TS" ask on #2135 still stand, given TS already implements both variants?
</details>

<details>
<summary>Reading order (for human reviewers)</summary>

1. **`strands-py/src/strands/types/media.py:50-75`** — start here; the whole PR is this. A new public `DocumentBlockContent` and two new `DocumentSource` fields. Findings 1–3 live on these 26 lines, and the cross-provider question follows directly from the widening.
2. **`strands-py/src/strands/models/bedrock.py:639-655`** — the only implementation: two new `elif` branches, plus one real behavior change to *existing* input (`formatted_document_source` now defaults to `None` at `:642`, where an unrecognized source used to raise `UnboundLocalError`).
3. **`bedrock.py:674-687` and `:753-766`** — skim; rename-only churn forced by mypy once the type widened. No semantic change.
4. **Tests** — `tests/strands/types/test_media.py:56-90` (type-shape asserts) and `tests/strands/models/test_bedrock.py:2393-2442` (the two formatter tests).
5. **Context worth having open, not in the diff** — `models/_validation.py:76-93` (the #1572 gate these variants lack) and `strands-ts/src/models/bedrock.ts:1233-1265` (TS's name and its `text` → bytes encode).
</details>

<details>
<summary>Appendix — non-blocking (9)</summary>

- ⚪ `bedrock.py:642` vs `:678`/`:757` — the three sibling branches now disagree on an unrecognized source: document yields `"source": None`, image and video still raise `UnboundLocalError`. The `None` is the better end (botocore catches it client-side); image/video were explicitly deferred on #2135. Converge or leave.
- ⚪ `bedrock.py:642,653` — the new silent paths log nothing, unlike `_handle_location`'s `logger.warning(… | skipping content block)`. Folds into the inline fixes.
- ⚪ Source-key precedence (`location > bytes > text > content`) is silent and undocumented while the docstring says only one should be specified; TS throws instead. Fine as shipped — a precedence *test* would freeze undefined behavior, so a one-line comment is the better move if any.
- ⚪ `test_media.py:53-54` asserts `doc_source["s3Location"]`, a key `DocumentSource` has never had (the field is `location`). Survives because mypy runs on `./src` only. This PR fixed the matching docstring inaccuracy right above the tests it adds — worth fixing the test in the same pass.
- ⚪ The 4 new `test_media.py` tests can't fail (deleting both new fields keeps them green — TypedDict annotations on plain dicts, and mypy doesn't see `tests/`). Blessed pre-existing pattern in that file; for the record, not a change request.
- ⚪ `DocumentContent`'s docstring (`media.py:78-85`) still omits `citations` and `context` — pre-existing, but `citations` is now this feature's companion knob; worth a line while here.
- ⚪ `ImageSource` (`media.py:101`) and `VideoSource` (`:131`) docstrings still carry the same "only `bytes` or `s3Location`" inaccuracy this PR fixed for documents.
- ⚪ #1066's own repro (`text` source **with** `citations.enabled`) isn't among the regression tests — the new tests exercise the sources without citations. Coverage gap, not a defect; the filed repro formats correctly now.
- ⚪ Checked and dismissed — please don't re-raise: the `botocore>=1.29.0` floor predates the `text`/`content` union members, but 1.29–1.34 ship no `bedrock-runtime` model at all, so the floor is already unusable for `BedrockModel`; `s3Location` (#1572) had identical exposure and shipped without a bump.

**Pre-existing, being filed as separate issues (not this PR's problem):** `openai_responses.py` never got #1572's `location` gate; `context_offloader` destroys any non-`bytes` document source (true for `location` on `main` today); the provider gate doesn't cover tool-result document blocks at all.
</details>

PR #3583, opened 2026-07-31
fix(bedrock-kb): resolve a default region for the store's clients so it works in cloud envs

### LINKED ISSUE #3564: [BUG] BedrockKnowledgeBaseStore throws NoRegionError in Cloud Envs
### Checks

### SDK Language

Python

### Strands Version

1.50.2

### Language Runtime Version

Python 3.12

### Operating System

Amazon Linux 2 (EC2)

### Installation Method

pip

### Steps to Reproduce

1. Deploy to an EC2 instance without `AWS_DEFAULT_REGION` or `AWS_REGION` environment variables set
2. Instantiate a `BedrockKnowledgeBaseStore`, wrap in try-except if you want to actually catch the error:

```python
from strands.vended_memory_stores import BedrockKnowledgeBaseStore

store = BedrockKnowledgeBaseStore(
    name="user-memory",
    description="Personal memories about the user",
    scope="user-123",
    writable=True,
    config={
        "knowledge_base_id": "ABCDEFGHIJ",
        "data_source_type": "S3",
        "data_source_id": "DATASOURCE123",
        "s3": {
            "bucket": "my-bucket",
            "prefix": "memories/",
        },
    },
)
```

3. Observe error:

```
NoRegionError: You must specify a region.
```

## Expected Behavior

The `BedrockKnowledgeBaseStore` constructor or its `config` dict should accept a `region` or `region_name` parameter to specify the AWS region, similar to how `BedrockModel` accepts `region_name`, or just pick it up from the IAM role associated with the cloud resource (in this case EC2 but could be Lambda, etc.):

```python
# This works for BedrockModel:
model = BedrockModel(model_id="anthropic.claude-3-sonnet", region_name="us-west-2")

# Expected for BedrockKnowledgeBaseStore:
store = BedrockKnowledgeBaseStore(
    name="user-memory",
    ...,
    config={
        "knowledge_base_id": "ABCDEFGHIJ",
        "region": "us-west-2",  # or region_name
        ...
    },
)
```

## Additional Context

The error originates from boto3 client initialization inside the store. The `BedrockModel` class properly exposes `region_name` as a parameter, but `BedrockKnowledgeBaseStore` does not appear to pass region configuration to its internal boto3 clients.

### Expected Behavior

Defaults to some region or another or respects IAM role region and permissions on the deployed EC2 or cloud resource.

### Actual Behavior

Seems to work fine on local `aws sso login` and pulls those regions fine. Does not operate the same way in Cloud env.

### Additional Context

_No response_

### Possible Solution

_No response_

### Related Issues

_No response_

SuperMarioYL:
## Description

`BedrockKnowledgeBaseStore` constructs its boto3 clients — the eager `bedrock-agent-runtime`
client in `__init__`, and the lazy `s3` / `bedrock-agent` clients — with **no region
fallback**. In a cloud environment (EC2/Lambda/ECS) where the IAM role is present but
`AWS_REGION` is not exported (user-data scripts, Lambda custom runtimes, ECS tasks
without env passthrough), boto3 cannot resolve a region and the store raises
`botocore.exceptions.NoRegionError` at construction — before any retrieval can happen.

This is inconsistent with the SDK's own model providers: `SageMakerAIModel`
(`models/sagemaker.py`) and `BedrockModel` (`models/bedrock.py`) both resolve a default
region (`region_name or AWS_REGION or "us-west-2"`) rather than forcing every caller to
pass a pre-built client. The store was the outlier.

This PR brings the store in line with that established idiom: a region is resolved as
`config["region_name"] or os.getenv("AWS_REGION") or "us-west-2"` and passed to all three
default-client construction sites. An explicitly-injected client (`runtime_client` /
`agent_client` / `s3.client`) still wins and bypasses resolution, so nothing changes for
callers who already inject clients.

Resolves: #3564

## Public API Changes

One new optional config field on `BedrockKnowledgeBaseConfig` (TypedDict, `total=False` —
fully backward-compatible; existing callers are unaffected).

```python
# Before — no way to pin a region for the store's default clients; callers who needed
# one had to build and inject each client themselves.
BedrockKnowledgeBaseStore(
    name="kb",
    config={"knowledge_base_id": "kb-id"},
)  # NoRegionError in a cloud env with no AWS_REGION

# After — optional region_name; otherwise resolves from AWS_REGION, then "us-west-2".
BedrockKnowledgeBaseStore(
    name="kb",
    config={"knowledge_base_id": "kb-id", "region_name": "eu-west-1"},
)
```

Resolution order (mirrors `SageMakerAIModel`): explicit `region_name` → `AWS_REGION` env
→ `"us-west-2"`. Injected clients always take precedence and skip resolution.

## Type of Change

Bug fix

## Testing

Added a `TestRegionResolution` class to `test_store.py` that drives the real
`BedrockKnowledgeBaseStore` constructor (no mocking of the code under fix): the no-hint case
constructs a real default client and asserts the resolved region, while the resolution-order
cases intercept `boto3.client` to observe the call. Cases cover the default fallback, explicit
`region_name` winning over the env var, `AWS_REGION` winning when no explicit config is given,
and the resolved region threading into all three default-client sites.

The no-hint case is red on `main` (it raises `NoRegionError` at construction) and green on this
branch; the pre-existing store tests are unchanged and still pass, because the `or`-short-circuit
and `if is None` guards mean injected-client paths bypass region resolution.

## Why a default, not "operators should pass a client"

The existing `_get_s3_client` / `_get_agent_client` docstrings note callers can inject a
client for a specific region — and that still works. But the two sibling model providers
in this same SDK already ship a default region for the same cloud-env reason, so the store
raising at construction was an inconsistency, not a deliberate operator-config contract.
This PR makes the three consistent; operators who inject clients are unaffected.

## Checklist

- [x] I have reviewed and understand every line of code in this PR, and I can explain why it works

----

### DIFF
--- a/site/src/content/docs/user-guide/concepts/memory/bedrock-knowledge-base.mdx
+++ b/site/src/content/docs/user-guide/concepts/memory/bedrock-knowledge-base.mdx
@@ -105,7 +105,7 @@ The inner `BedrockKnowledgeBaseConfig` is the reusable connection: which knowled
 | `s3` | S3 ingestion settings (`bucket`, `prefix`). Required when the data source type is `'S3'`. |
 | <Syntax py="scope_metadata_key" ts="scopeMetadataKey" /> | Metadata attribute key used for scope filtering. Defaults to `'namespace'`. |
 | <Syntax py="runtime_client / agent_client" ts="runtimeClient / agentClient" /> | Pre-constructed AWS clients. When omitted, default clients are constructed using the standard credential chain (the agent client lazily, on first write). |
-| <Syntax py="region_name" ts="regionName" /> | AWS region hint for the default clients the store constructs. When omitted, boto3 resolves the region from its own chain (`AWS_DEFAULT_REGION` env, then `~/.aws/config` — no IMDS). Ignored for any client you inject explicitly (`runtime_client` / `agent_client` / `s3.client`). |
+| <Syntax py="region_name" ts="regionName" /> | AWS region hint for the default clients the store constructs. When omitted, boto3 resolves the region itself. Ignored for any client you inject explicitly (`runtime_client` / `agent_client` / `s3.client`). |
 
 Because the connection is a separate object, you build it once and vary only `name` and `scope` per store. This is the cheap way to give each tenant an isolated namespace over a single knowledge base:
 
--- a/site/src/content/docs/user-guide/concepts/memory/bedrock-knowledge-base.mdx
+++ b/site/src/content/docs/user-guide/concepts/memory/bedrock-knowledge-base.mdx
@@ -105,6 +105,7 @@ The inner `BedrockKnowledgeBaseConfig` is the reusable connection: which knowled
 | `s3` | S3 ingestion settings (`bucket`, `prefix`). Required when the data source type is `'S3'`. |
 | <Syntax py="scope_metadata_key" ts="scopeMetadataKey" /> | Metadata attribute key used for scope filtering. Defaults to `'namespace'`. |
 | <Syntax py="runtime_client / agent_client" ts="runtimeClient / agentClient" /> | Pre-constructed AWS clients. When omitted, default clients are constructed using the standard credential chain (the agent client lazily, on first write). |
+| <Syntax py="region_name" ts="regionName" /> | AWS region hint for the default clients the store constructs. When omitted, boto3 resolves the region from its own chain (`AWS_DEFAULT_REGION` env, then `~/.aws/config` — no IMDS). Ignored for any client you inject explicitly (`runtime_client` / `agent_client` / `s3.client`). |
 
 Because the connection is a separate object, you build it once and vary only `name` and `scope` per store. This is the cheap way to give each tenant an isolated namespace over a single knowledge base:
 
--- a/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
+++ b/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
@@ -128,9 +128,8 @@ class BedrockKnowledgeBaseStore(MemoryStore):
 
         # Region hint for any default boto3 client the store constructs. Only an explicit
         # ``region_name`` config is threaded through; when absent, ``None`` is passed so boto3
-        # resolves the region itself rather than the store silently picking one — a wrong
-        # default would misroute region-scoped knowledge bases. Injected clients are used
-        # verbatim and bypass this.
+        # resolves the region itself rather than the store silently picking one.
+        # Injected clients bypass this and use their own region setup.
         self._region = kb_config.get("region_name")
 
         # The runtime client is built eagerly: search is the read path every store exercises. A
--- a/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
+++ b/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
@@ -128,8 +128,7 @@ class BedrockKnowledgeBaseStore(MemoryStore):
 
         # Region hint for any default boto3 client the store constructs. Only an explicit
         # ``region_name`` config is threaded through; when absent, ``None`` is passed so boto3
-        # resolves the region from its own chain (``AWS_DEFAULT_REGION`` env, then
-        # ``~/.aws/config`` — no IMDS) rather than the store silently picking one — a wrong
+        # resolves the region itself rather than the store silently picking one — a wrong
         # default would misroute region-scoped knowledge bases. Injected clients are used
         # verbatim and bypass this.
         self._region = kb_config.get("region_name")
--- a/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
+++ b/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
@@ -128,9 +128,10 @@ class BedrockKnowledgeBaseStore(MemoryStore):
 
         # Region hint for any default boto3 client the store constructs. Only an explicit
         # ``region_name`` config is threaded through; when absent, ``None`` is passed so boto3
-        # resolves the region from its own chain (``AWS_REGION`` env, ``~/.aws/config``, IMDS)
-        # rather than the store silently picking one — a wrong default would misroute
-        # region-scoped knowledge bases. Injected clients are used verbatim and bypass this.
+        # resolves the region from its own chain (``AWS_DEFAULT_REGION`` env, then
+        # ``~/.aws/config`` — no IMDS) rather than the store silently picking one — a wrong
+        # default would misroute region-scoped knowledge bases. Injected clients are used
+        # verbatim and bypass this.
         self._region = kb_config.get("region_name")
 
         # The runtime client is built eagerly: search is the read path every store exercises. A
--- a/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
+++ b/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
@@ -128,10 +128,10 @@ class BedrockKnowledgeBaseStore(MemoryStore):
         self._data_source_type = kb_config.get("data_source_type")
         self._data_source_id = kb_config.get("data_source_id")
 
-        # Region for any default boto3 client the store constructs. Resolution mirrors
-        # SageMakerAIModel: explicit config -> AWS_REGION env -> DEFAULT_REGION. This avoids the
-        # NoRegionError raised by boto3 in cloud envs (EC2/Lambda) with no region hint. Injected
-        # clients are used verbatim and bypass this resolution.
+        # Region for any default boto3 client the store constructs. Resolution order is
+        # explicit config -> AWS_REGION env -> DEFAULT_REGION. This avoids the NoRegionError
+        # raised by boto3 in cloud envs (EC2/Lambda) with no region hint. Injected clients are
+        # used verbatim and bypass this resolution.
         self._region = kb_config.get("region_name") or os.getenv("AWS_REGION") or DEFAULT_REGION
 
         # The runtime client is built eagerly: search is the read path every store exercises. A
--- a/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
+++ b/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
@@ -10,7 +10,6 @@ from __future__ import annotations
 
 import json
 import logging
-import os
 import uuid
 from typing import TYPE_CHECKING, Any, cast
 
@@ -36,7 +35,6 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 DEFAULT_MAX_SEARCH_RESULTS = 10
-DEFAULT_REGION = "us-west-2"
 
 # A Bedrock attribute value: ``{"type": ..., "stringValue"/"numberValue"/...}``. There is no Python
 # SDK type for this (boto3 passes plain dicts), so it is modeled as ``dict[str, Any]``.
@@ -128,11 +126,12 @@ class BedrockKnowledgeBaseStore(MemoryStore):
         self._data_source_type = kb_config.get("data_source_type")
         self._data_source_id = kb_config.get("data_source_id")
 
-        # Region for any default boto3 client the store constructs. Resolution order is
-        # explicit config -> AWS_REGION env -> DEFAULT_REGION. This avoids the NoRegionError
-        # raised by boto3 in cloud envs (EC2/Lambda) with no region hint. Injected clients are
-        # used verbatim and bypass this resolution.
-        self._region = kb_config.get("region_name") or os.getenv("AWS_REGION") or DEFAULT_REGION
+        # Region hint for any default boto3 client the store constructs. Only an explicit
+        # ``region_name`` config is threaded through; when absent, ``None`` is passed so boto3
+        # resolves the region from its own chain (``AWS_REGION`` env, ``~/.aws/config``, IMDS)
+        # rather than the store silently picking one — a wrong default would misroute
+        # region-scoped knowledge bases. Injected clients are used verbatim and bypass this.
+        self._region = kb_config.get("region_name")
 
         # The runtime client is built eagerly: search is the read path every store exercises. A
         # default client is only constructed when none was injected.
--- a/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
+++ b/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
@@ -128,8 +128,7 @@ class BedrockKnowledgeBaseStore(MemoryStore):
 
         # Region hint for any default boto3 client the store constructs. Only an explicit
         # ``region_name`` config is threaded through; when absent, ``None`` is passed so boto3
-        # resolves the region itself rather than the store silently picking one.
-        # Injected clients bypass this and use their own region setup.
+        # resolves the region itself. Injected clients bypass this and use their own region setup.
         self._region = kb_config.get("region_name")
 
         # The runtime client is built eagerly: search is the read path every store exercises. A
--- a/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
+++ b/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 
 import json
 import logging
+import os
 import uuid
 from typing import TYPE_CHECKING, Any, cast
 
@@ -35,6 +36,7 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 DEFAULT_MAX_SEARCH_RESULTS = 10
+DEFAULT_REGION = "us-west-2"
 
 # A Bedrock attribute value: ``{"type": ..., "stringValue"/"numberValue"/...}``. There is no Python
 # SDK type for this (boto3 passes plain dicts), so it is modeled as ``dict[str, Any]``.
@@ -126,10 +128,16 @@ class BedrockKnowledgeBaseStore(MemoryStore):
         self._data_source_type = kb_config.get("data_source_type")
         self._data_source_id = kb_config.get("data_source_id")
 
+        # Region for any default boto3 client the store constructs. Resolution mirrors
+        # SageMakerAIModel: explicit config -> AWS_REGION env -> DEFAULT_REGION. This avoids the
+        # NoRegionError raised by boto3 in cloud envs (EC2/Lambda) with no region hint. Injected
+        # clients are used verbatim and bypass this resolution.
+        self._region = kb_config.get("region_name") or os.getenv("AWS_REGION") or DEFAULT_REGION
+
         # The runtime client is built eagerly: search is the read path every store exercises. A
         # default client is only constructed when none was injected.
         self._runtime_client: AgentsforBedrockRuntimeClient = kb_config.get("runtime_client") or boto3.client(
-            "bedrock-agent-runtime"
+            "bedrock-agent-runtime", region_name=self._region
         )
 
         # The knowledge base type — either provided in config or resolved in ``initialize()``.
@@ -526,19 +534,21 @@ class BedrockKnowledgeBaseStore(MemoryStore):
     def _get_s3_client(self) -> S3Client:
         """Return the S3 client, constructing a default one lazily on first use.
 
-        A default client is built with no extra configuration. Callers needing a specific region,
-        credentials, or endpoint build the client themselves and inject it via the ``s3`` config.
+        A default client is built with no extra configuration beyond the resolved region. Callers
+        needing specific credentials or an endpoint build the client themselves and inject it via
+        the ``s3`` config.
         """
         if self._s3_client is None:
-            self._s3_client = boto3.client("s3")
+            self._s3_client = boto3.client("s3", region_name=self._region)
         return self._s3_client
 
     def _get_agent_client(self) -> AgentsforBedrockClient:
         """Return the agent client, constructing a default one lazily on first use.
 
-        A default client is built with no extra configuration. Callers needing a specific region,
-        credentials, or endpoint build the client themselves and inject it via ``agent_client``.
+        A default client is built with no extra configuration beyond the resolved region. Callers
+        needing specific credentials or an endpoint build the client themselves and inject it via
+        ``agent_client``.
         """
         if self._agent_client is None:
-            self._agent_client = boto3.client("bedrock-agent")
+            self._agent_client = boto3.client("bedrock-agent", region_name=self._region)
         return self._agent_client
--- a/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/types.py
+++ b/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/types.py
@@ -79,6 +79,10 @@ class BedrockKnowledgeBaseConfig(TypedDict, total=False):
         agent_client: Pre-constructed agent client for ``IngestKnowledgeBaseDocuments`` calls. When
             omitted, a default boto3 client is constructed lazily on first write. To target a specific
             region/credentials/endpoint, build the client yourself and inject it here.
+        region_name: AWS region for the default ``bedrock-agent-runtime`` / ``bedrock-agent`` / ``s3``
+            clients constructed by the store. Resolution order is this field, then the ``AWS_REGION``
+            environment variable, then ``us-west-2`` — mirroring ``SageMakerAIModel``. Ignored for any
+            client the caller injects explicitly (``runtime_client`` / ``agent_client`` / ``s3.client``).
     """
 
     knowledge_base_id: Required[str]
@@ -89,6 +93,7 @@ class BedrockKnowledgeBaseConfig(TypedDict, total=False):
     scope_metadata_key: str
     runtime_client: AgentsforBedrockRuntimeClient
     agent_client: AgentsforBedrockClient
+    region_name: str
 
 
 class BedrockKnowledgeBaseAccessControlEntry(TypedDict):
--- a/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/types.py
+++ b/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/types.py
@@ -79,10 +79,10 @@ class BedrockKnowledgeBaseConfig(TypedDict, total=False):
         agent_client: Pre-constructed agent client for ``IngestKnowledgeBaseDocuments`` calls. When
             omitted, a default boto3 client is constructed lazily on first write. To target a specific
             region/credentials/endpoint, build the client yourself and inject it here.
-        region_name: AWS region for the default ``bedrock-agent-runtime`` / ``bedrock-agent`` / ``s3``
-            clients constructed by the store. Resolution order is this field, then the ``AWS_REGION``
-            environment variable, then ``us-west-2``. Ignored for any client the caller injects
-            explicitly (``runtime_client`` / ``agent_client`` / ``s3.client``).
+        region_name: AWS region hint for the default ``bedrock-agent-runtime`` / ``bedrock-agent`` / ``s3``
+            clients constructed by the store. When omitted, ``None`` is passed and boto3 resolves the
+            region from its own chain (``AWS_REGION`` env, ``~/.aws/config``, IMDS). Ignored for any
+            client the caller injects explicitly (``runtime_client`` / ``agent_client`` / ``s3.client``).
     """
 
     knowledge_base_id: Required[str]
--- a/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/types.py
+++ b/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/types.py
@@ -81,9 +81,8 @@ class BedrockKnowledgeBaseConfig(TypedDict, total=False):
             region/credentials/endpoint, build the client yourself and inject it here.
         region_name: AWS region hint for the default ``bedrock-agent-runtime`` / ``bedrock-agent`` / ``s3``
             clients constructed by the store. When omitted, ``None`` is passed and boto3 resolves the
-            region from its own chain (``AWS_DEFAULT_REGION`` env, then ``~/.aws/config`` — no IMDS).
-            Ignored for any client the caller injects explicitly (``runtime_client`` / ``agent_client``
-            / ``s3.client``).
+            region itself. Ignored for any client the caller injects explicitly
+            (``runtime_client`` / ``agent_client`` / ``s3.client``).
     """
 
     knowledge_base_id: Required[str]
--- a/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/types.py
+++ b/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/types.py
@@ -81,8 +81,9 @@ class BedrockKnowledgeBaseConfig(TypedDict, total=False):
             region/credentials/endpoint, build the client yourself and inject it here.
         region_name: AWS region hint for the default ``bedrock-agent-runtime`` / ``bedrock-agent`` / ``s3``
             clients constructed by the store. When omitted, ``None`` is passed and boto3 resolves the
-            region from its own chain (``AWS_REGION`` env, ``~/.aws/config``, IMDS). Ignored for any
-            client the caller injects explicitly (``runtime_client`` / ``agent_client`` / ``s3.client``).
+            region from its own chain (``AWS_DEFAULT_REGION`` env, then ``~/.aws/config`` — no IMDS).
+            Ignored for any client the caller injects explicitly (``runtime_client`` / ``agent_client``
+            / ``s3.client``).
     """
 
     knowledge_base_id: Required[str]
--- a/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/types.py
+++ b/strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/types.py
@@ -81,8 +81,8 @@ class BedrockKnowledgeBaseConfig(TypedDict, total=False):
             region/credentials/endpoint, build the client yourself and inject it here.
         region_name: AWS region for the default ``bedrock-agent-runtime`` / ``bedrock-agent`` / ``s3``
             clients constructed by the store. Resolution order is this field, then the ``AWS_REGION``
-            environment variable, then ``us-west-2`` — mirroring ``SageMakerAIModel``. Ignored for any
-            client the caller injects explicitly (``runtime_client`` / ``agent_client`` / ``s3.client``).
+            environment variable, then ``us-west-2``. Ignored for any client the caller injects
+            explicitly (``runtime_client`` / ``agent_client`` / ``s3.client``).
     """
 
     knowledge_base_id: Required[str]
--- a/strands-py/tests/strands/vended_memory_stores/bedrock_knowledge_base/test_store.py
+++ b/strands-py/tests/strands/vended_memory_stores/bedrock_knowledge_base/test_store.py
@@ -206,14 +206,10 @@ class TestConstructor:
         store, _agent, _s3 = make_s3_store()
         assert store.writable is True
 
-    def test_constructs_a_default_runtime_client_when_none_injected(self, monkeypatch, tmp_path):
-        # Simulate a cloud host with no region hint so the resolved region is the default.
-        monkeypatch.delenv("AWS_REGION", raising=False)
-        monkeypatch.delenv("AWS_DEFAULT_REGION", raising=False)
-        monkeypatch.setenv("HOME", str(tmp_path))
+    def test_constructs_a_default_runtime_client_when_none_injected(self):
         with patch("boto3.client") as client_fn:
             BedrockKnowledgeBaseStore(config=BedrockKnowledgeBaseConfig(knowledge_base_id="kb-1"), name="kb")
-            client_fn.assert_called_once_with("bedrock-agent-runtime", region_name="us-west-2")
+            client_fn.assert_called_once_with("bedrock-agent-runtime", region_name=None)
 
     def test_uses_the_injected_runtime_client_without_constructing_one(self, make_store):
         with patch("boto3.client") as client_fn:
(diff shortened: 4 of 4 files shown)


liramon2:
@strandly-the-agent Review this PR. Consolidate your findings in a single comment.

strandly-the-agent:
**TL;DR:** 🔴 **Changes-requested.** The reported bug is real and the `AWS_REGION` addition is a genuine improvement — but `store.py:135` skips `boto3.Session().region_name`, so the *most common* region-config paths (`AWS_DEFAULT_REGION`, `~/.aws/config` profile, SSO) are silently repointed to `us-west-2`, and GovCloud/China callers get a partition flip. That's a ~1-line reorder, not a rewrite. Second blocker: 3 of the 5 new tests provably can't detect this class of bug, and 3 break on any machine with `AWS_PROFILE` exported.

> _Edited in place to consolidate all six review passes (correctness, issue-alignment, adversarial, API/DevX, test-quality, docs) into this one comment, per the ask. `liramon2` has already directed the two headline fixes (misrouting + site docs); the rest below is the additional detail and the open design calls._

**What I verified** (branch `pr3583` @ `b352f69` vs `origin/main` @ `1937b4a`):
- ✅ `pytest tests/strands/vended_memory_stores/bedrock_knowledge_base/` → **75/75 pass**; `tests/strands/vended_memory_stores/` → 117 pass; broad `tests/strands` → 3838 pass (19 pre-existing missing-optional-dep errors, unrelated).
- ✅ The reported bug is real: on `main`, with no region hint at all, construction raises `NoRegionError`. This PR fixes that case, and `AWS_REGION` support is genuinely *additive* (boto3 itself only reads `AWS_DEFAULT_REGION`).
- 🔴 Regression reproduced: `AWS_DEFAULT_REGION=eu-central-1`, no `AWS_REGION` → `main` resolves `eu-central-1`, this branch resolves `us-west-2`. Same with `~/.aws/config` `[default] region = ap-northeast-1` → `us-west-2`.
- 🔴 Mutation-proven: keep `_region` correct but hardcode a wrong region into the eager client → the 2 threading tests fail, the 3 `_region`-only tests still pass.
- 🔴 `AWS_PROFILE=dev pytest <store test file>` → 3 failed (`botocore.exceptions.ProfileNotFound`), 72 passed.

<details>
<summary><strong>🔴 Blocker 1 — the chain skips <code>boto3.Session().region_name</code> (root cause, and everything downstream of it)</strong></summary>

`store.py:135`:
```python
self._region = kb_config.get("region_name") or os.getenv("AWS_REGION") or DEFAULT_REGION
```
never consults `boto3.Session().region_name` — which is the read path for `AWS_DEFAULT_REGION`, `~/.aws/config` profiles and SSO. Then `region_name=self._region` is passed explicitly at `store.py:140,542,553`, and an explicit `region_name` short-circuits boto3's own resolution. The sibling the PR body cites as precedent does include that step — `models/bedrock.py:192`:
```python
resolved_region = region_name or session.region_name or os.environ.get("AWS_REGION") or DEFAULT_BEDROCK_REGION
```

Head-to-head, `AWS_DEFAULT_REGION=eu-central-1`, no `AWS_REGION`:
```
main:    store._runtime_client.meta.region_name == "eu-central-1"
pr3583:  store._region == "us-west-2"   (runtime + s3 + agent all us-west-2)
```

**Suggested fix — same line count:**
```python
self._region = (
    kb_config.get("region_name") or boto3.Session().region_name or os.getenv("AWS_REGION") or DEFAULT_REGION
)
```
Or, closer to this store's own category: build one `boto3.Session(region_name=kb_config.get("region_name"))` and create all three clients from it — that's exactly what the three sibling persistence primitives do (`session/s3_session_manager.py:76`, `storage/s3_storage.py:205`, `vended_plugins/context_offloader/storage.py:533`), and with `region_name=None` it falls through boto3's full chain for free.

**All of the following are consequences of that one gap, not separate defects:**

- **Partition flip (GovCloud / China).** Captured the signed request offline via a `before-send` hook. On a `us-gov-west-1` host: this branch sends to `https://bedrock-agent-runtime.us-west-2.amazonaws.com/...` with cred scope `.../us-west-2/bedrock/aws4_request`; `main` used the `us-gov-west-1` endpoint and scope. China: `cn-north-1.amazonaws.com.cn` → `us-west-2.amazonaws.com`. A GovCloud workload emitting a signed request (query text in the body) to a commercial host is a data-boundary problem, not a 404.
- **Custom endpoints (PrivateLink/VPCE, LocalStack).** `AWS_ENDPOINT_URL` still wins the host, but the *signing* region flips: with `AWS_DEFAULT_REGION=us-east-1` + a `us-east-1` VPCE endpoint, this branch signs `.../us-west-2/bedrock/...`. These users already set a region the way boto3 documents.
- **Half-injected clients.** `types.py:79-82` tells callers to inject a client to target a region — but injecting only `runtime_client` (eu-west-1) leaves the lazy defaults at `us-west-2` (verified endpoints `bedrock-agent.us-west-2.amazonaws.com`, `s3.us-west-2.amazonaws.com`; `main` gave eu-west-1 for both), so `initialize()`'s `GetKnowledgeBase` runs in the wrong region. Symmetrically, `region_name` + an injected client is accepted silently and splits reads/writes across two regions — every sibling raises on "two ways to say the same thing" (`models/bedrock.py:188-189`, `storage/s3_storage.py:51-52`, `experimental/bidi/models/nova_sonic.py:187-188`).
- **PR body claims that no longer hold.** "existing callers are unaffected" is false for the population above; and "the two sibling model providers already ship a default region" is only half true — `SageMakerAIModel` matches, `BedrockModel` does not (it has the session step). Two different regions can now resolve in one process: with `AWS_REGION=eu-west-1 AWS_DEFAULT_REGION=us-east-2`, `BedrockModel` → `us-east-2` while this store → `eu-west-1`.
- **Compat-policy angle.** A change to *default* behavior for callers who take no action lands on the "does NOT apply" side of the pay-for-play exception (`team/FEATURE_LIFECYCLE.md:46-50`, restated `team/DECISIONS.md:120`). Fixing the root cause removes that conversation entirely.
- **Docs (already flagged by `liramon2`).** `site/src/content/docs/user-guide/concepts/memory/bedrock-knowledge-base.mdx:17,107,347` still say region setup uses "the standard credential chain" / is "shared with the Bedrock model provider" — both now false — and `:99-107` has no `region_name` row. Note `site/src/components/Syntax.astro:5-6` types both `py` and `ts` as required, so a Python-only row can't use `<Syntax>` — use a plain code span or `LanguageSupportAside`. Adopting the session-first fix makes two of those three edits unnecessary. `types.py:77-81` is also stale next to the new field ("To target a specific region… build the client yourself").
</details>

<details>
<summary><strong>🔴 Blocker 2 — the new tests wouldn't have caught this, and they aren't hermetic</strong></summary>

- **3 of 5 assertions are vacuous.** `test_store.py:246-249`, `:251-257`, `:259-262` assert only `store._region`, which is computed *before* any `boto3.client()` call — so they say nothing about whether the region reached a client, which is the entire point of the fix. Mutation check (keep `_region` correct, hardcode `region_name="ap-northeast-1"` into the eager client): the 2 threading tests fail, these 3 pass. This file's own scaffolding docstring (`test_store.py:12-13`) says default-client construction "is observed by patching `boto3.client`", and the closest precedent is `tests/strands/storage/test_s3_storage.py:158-166 test_client_created_with_region`, which asserts on the call kwargs.
- **Real AWS clients in unit tests.** `_cloud_env` (`test_store.py:238-244`) hand-rolls env isolation (`HOME` + `AWS_CONFIG_FILE`) but misses `AWS_PROFILE`, and 3 tests construct a real client → `ProfileNotFound` on any dev machine with a profile exported (3 failed / 72 passed, reproduced). They also make 2 real socket connects to `169.254.169.254:80` per construction (none with `AWS_EC2_METADATA_DISABLED=true`). `strands-py/docs/TESTING.md` is explicit — never hit real AWS in a unit test — and `tests/conftest.py:33-46` already ships `moto_env` / `moto_mock_aws` (composed in `test_streaming.py:19-20`, `test_metrics.py:15-16`, `test_tracer.py:22-23`). No other test file hand-rolls the `HOME`/`AWS_CONFIG_FILE` trick. The fix isn't "also delenv `AWS_PROFILE`" (whack-a-mole) — it's don't build a real client; a patched `boto3.client` gives strictly more signal here.
- **Coverage gaps that matter.** Nothing pins the `AWS_DEFAULT_REGION`/profile case (either the old or the new semantics); an explicit `region_name` is never threaded into *any* client in *any* test (both threading tests use env/default); `region_name` + injected client — the contract documented at `types.py:82-85` — is untested. Also note the modified `test_constructs_a_default_runtime_client_when_none_injected` (`:209-216`) fully mocks `boto3.client`, so its `delenv`/`setenv("HOME")` lines are dead (verified: still passes with `AWS_DEFAULT_REGION=zz-fake-1 AWS_PROFILE=totally-bogus-profile`), and it duplicates `test_falls_back_to_default_region_with_no_hint`.

Happy to post a drafted replacement test class (all mocked, no `HOME`/`AWS_CONFIG_FILE`/`AWS_PROFILE` handling, plus the missing cases) if that's useful.
</details>

<details>
<summary><strong>🟡 Should-fix (real, not blocking)</strong></summary>

- **A typo'd `region` key is silently swallowed** — and `region` is exactly the spelling #3564's reporter used. Verified: `config={"knowledge_base_id": "kb-1", "region": "eu-west-1"}` → `_region = us-west-2`, no warning. 13 model providers call `validate_config_keys` (`models/_validation.py:42`, e.g. `models/bedrock.py:242`, `models/sagemaker.py:152`); this config doesn't. The type checker isn't a backstop for the pattern the docs themselves use (`bedrock-knowledge-base.mdx:114-125` builds a bare dict → mypy only reports a generic `dict[str, str]` mismatch, no key-level check).
- **`region_name` / `AWS_REGION` are unvalidated.** Promoting `AWS_REGION` to top priority with no `.strip()` turns previously-harmless junk into a hard failure: `AWS_REGION` with a trailing newline plus a valid `AWS_DEFAULT_REGION=eu-west-1` → `ValueError: Invalid endpoint: https://bedrock-agent-runtime.us-east-1\n.amazonaws.com`; `main` worked. On the config side, `"US-WEST-2"` and a `class Region(str, Enum)` value both pass botocore's check and silently build a bogus host (`...Region.EUW1.amazonaws.com`), failing far from the config site; `12345` / `b"us-east-1"` surface as bare `TypeError`s that never mention region. `.strip()` + `models/_validation.py:20 validate_region` on the coerced `str` closes most of this in one line.
- **Nothing logs the resolved region.** A loud `NoRegionError` became a silent wrong-region call with no breadcrumb. Per the repo's structured-logging convention: `logger.debug("region=%s | resolved default client region", self._region)`. (`models/sagemaker.py` has the same gap — not this PR's fault, but it's already in this code path.)
</details>

<details>
<summary><strong>⚪ Nits (bundled, non-blocking)</strong></summary>

- `types.py:84-85` cross-references `SageMakerAIModel` in a *public* docstring — a memory-store user shouldn't have to reason about a SageMaker model provider, and it rots when `sagemaker.py` changes. State the order directly.
- `store.py:133`'s "(EC2/Lambda)" is misleading: the Lambda runtime sets **both** `AWS_REGION` and `AWS_DEFAULT_REGION` as reserved env vars, so Lambda was never broken here. The real population is bare EC2/ECS/EKS.
- Evergreen-comment style: `store.py:130-134`, `test_store.py:232-233` ("used to raise… it now resolves") and `:247` ("Red on master, green after the fix") narrate the change rather than the invariant. There's merged precedent for this pattern (`tests/strands/models/test_bedrock.py:813`), so not worth blocking on — flagging only so it isn't a surprise.
- `types.py:89`: `knowledge_base_type` is still missing from the `Attributes:` block while the new field is documented.
- `DEFAULT_REGION` is now the SDK's fifth region default (`models/bedrock.py:44`, `models/sagemaker.py:162`, `nova_sonic.py:198`, `tests_integ/_aws.py:14`) — two values, four spellings. A shared `_boto.py` resolver is the real answer, but that's a separate PR; please don't import `DEFAULT_BEDROCK_REGION` here (`vended_memory_stores/__init__.py:16-19` lazy-loads specifically to defer optional deps). Export hygiene of the new constant is fine (out of `__all__`).
</details>

<details>
<summary><strong>Attacked and ruled out — no action needed</strong></summary>

- **Cross-region S3 data-source bucket**: not a problem on either tree. `S3RegionRedirectorv2` (`botocore/utils.py:1734-1806`) handles the 301 + `x-amz-bucket-region` and retries; `add()` succeeds with the same 2-attempt shape on `main` and this branch.
- **SSRF / endpoint injection via `region_name`**: not reachable — botocore's `validate_region_name` regex rejects `@ : / .` before any endpoint is built.
- **`region_name=""` / `None`**: falls through correctly. (Bonus: `main` actually *crashes* on an empty `AWS_DEFAULT_REGION`; this branch handles it.)
- **Post-construction env mutation**: this branch is *better* — `main`'s lazy clients drift when env changes after construction; pinning at `__init__` is the right call.
- **Test env leakage / ordering**: none observed; full file passes under `-n auto` and reordered.
- **Backward compat of the TypedDict itself**: additive on `total=False` is fine. One low-probability type break: a downstream `class Sub(BedrockKnowledgeBaseConfig)` that already declared its own `region_name` (plausible for someone working around #3564) now gets mypy `Overwriting TypedDict field "region_name" while extending`.
- **`api/needs-review` label**: not needed — `team/API_BAR_RAISING.md:20-21` exempts "Minimal changes… adding a new parameter to an infrequently used method". No API meeting; the default-behavior question above is a compatibility call, not an API-shape one.
</details>

## Questions

**Open design calls (non-blocking, but worth a decision in-thread):**
1. Is a silent hardcoded `us-west-2` the right default for a store that points at exactly *one* KB in exactly one region? Guessing wrong surfaces as "that KB doesn't exist", sending users hunting for a KB they can see in the console. `models/_openai_bedrock.py:95-99` sets the opposite precedent — it raises and names the field. Session-first + raise (or at minimum `logger.warning` on fallback) may be better DevX than a guess.
2. Worth adding `boto_session` to the config? Six siblings accept one (`models/bedrock.py:172-173`, `models/sagemaker.py:141-142`, `session/s3_session_manager.py:53-54`, `storage/s3_storage.py:36-37`, `vended_plugins/context_offloader/storage.py:514-515`, and `models/_openai_bedrock.py:64` — the precedent for a session *inside* a config TypedDict). Today "one region + custom credentials" costs three client constructions plus knowledge of three boto3 service names that are otherwise store internals. It's also the only route to `boto_client_config`, which is how the other six set `user_agent_extra="strands-agents"` — `vended_memory_stores/` currently has zero matches for that, making this the one AWS-touching primitive whose calls aren't attributable to Strands.
3. #3564 also asked to "just pick it up from the IAM role" (i.e. IMDS). Verified that isn't reachable through boto3's default chain at all: `BOTOCORE_DEFAUT_SESSION_VARIABLES['region'] == ('region', 'AWS_DEFAULT_REGION', None, None)`, and `IMDSRegionProvider` / `InstanceMetadataRegionFetcher` (`botocore/utils.py:719-802`) is wired only into `defaults_mode="auto"` latency tuning (`botocore/session.py:225-236`) — never into region resolution. Worth saying so explicitly on the issue rather than leaving that half of the ask silently unaddressed. (Side note: `models/_openai_bedrock.py:88-91`'s comment claiming `boto3.Session()` "falls back to EC2 instance metadata" is inaccurate — pre-existing, and *not* something to drive-by fix here.)
4. TS parity: `strands-ts/src/vended-memory-stores/bedrock-knowledge-base/store.ts:249,660,667` construct all three clients with `{}` and `BedrockKnowledgeBaseConfig` (`store.ts:88-125`) has no `regionName` — i.e. TS has the identical #3564 defect. Follow-up issue rather than this PR (per `AGENTS.md:78`), or intentionally out of scope?

Good direction on a real bug, `SuperMarioYL` — the ask is the one-line resolution order plus the two test-hermeticity fixes, not a restructure.

---
<sub>Produced by `strandly-the-agent` running a multi-pass review pipeline (correctness · issue-alignment · adversarial · API/DevX · test-quality · docs). Findings for a human maintainer to confirm, not a final verdict.</sub>

strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
```diff
@@ -126,10 +128,16 @@ def __init__(self, **store_config: Unpack[BedrockKnowledgeBaseStoreConfig]) -> N
         self._data_source_type = kb_config.get("data_source_type")
         self._data_source_id = kb_config.get("data_source_id")
 
+        # Region for any default boto3 client the store constructs. Resolution mirrors
+        # SageMakerAIModel: explicit config -> AWS_REGION env -> DEFAULT_REGION. This avoids the
```
liramon2: This comment is overly verbose. You can drop the `SageMakerAIModel` comparison.

strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/types.py
```diff
@@ -79,6 +79,10 @@ class BedrockKnowledgeBaseConfig(TypedDict, total=False):
         agent_client: Pre-constructed agent client for ``IngestKnowledgeBaseDocuments`` calls. When
             omitted, a default boto3 client is constructed lazily on first write. To target a specific
             region/credentials/endpoint, build the client yourself and inject it here.
+        region_name: AWS region for the default ``bedrock-agent-runtime`` / ``bedrock-agent`` / ``s3``
+            clients constructed by the store. Resolution order is this field, then the ``AWS_REGION``
+            environment variable, then ``us-west-2`` — mirroring ``SageMakerAIModel``. Ignored for any
```
liramon2: Same comment suggestion as above.

strands-py/tests/strands/vended_memory_stores/bedrock_knowledge_base/test_store.py
```diff
@@ -206,17 +206,90 @@ def test_allows_writable_with_s3_data_source(self, make_s3_store):
...
+            client_fn.assert_called_once_with("bedrock-agent-runtime", region_name="us-west-2")
 
     def test_uses_the_injected_runtime_client_without_constructing_one(self, make_store):
         with patch("boto3.client") as client_fn:
             make_store()
             client_fn.assert_not_called()
 
 
+# --------------------------------------------------------------------------- #
+# Region resolution for default clients (issue #3564)
```
liramon2: Nit: remove `(issue #3564)`

liramon2 (changes requested):

liramon2:
@SuperMarioYL Thanks for the contribution. I left a few suggestions. As for the findings above, please fix the misrouting regression and document the new options in the site docs.

strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
SuperMarioYL: Done — dropped the `SageMakerAIModel` cross-reference and reflowed the comment. (e7770eb)

strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/types.py
SuperMarioYL: Done — removed the mirroring reference here too. (e7770eb)

strands-py/tests/strands/vended_memory_stores/bedrock_knowledge_base/test_store.py
SuperMarioYL: Done — removed the `(issue #3564)` tag from the section header. (e7770eb)

opieter-aws (changes requested):
Thanks for the PR. Adding `region_name` on the config is the right call, but there's a problem with falling back to `us-west-2`: since the region is now always passed explicitly, it overrides boto’s own resolution instead of backstopping it. 

With e.g. `region = ca-central-1` in `~/.aws/config` and no env vars, main resolves `ca-central-1` and this branch resolves `us-west-2`. Silently picking the wrong region is problematic, since KBs are scoped by region, so writes may end up in a wrong bucket.

Let's just resolve with a hint and let boto do the routing here, and remove the fallback.

SuperMarioYL:
Thanks @opieter-aws — good catch on the fallback. Pushed `5947f04d`: dropped the `us-west-2` default and the `AWS_REGION` env check. The store now only threads an explicit `region_name` config into the default `boto3.client` calls; when it is absent, `None` is passed so boto3 resolves the region from its own chain (`AWS_REGION` env → `~/.aws/config` → IMDS). That means a host with `region = ca-central-1` in `~/.aws/config` now resolves `ca-central-1` instead of being overridden to `us-west-2`, and a host with no resolvable region fails loudly (`NoRegionError`) rather than silently writing to the wrong region.

The tests in `TestRegionResolution` were rewritten to pin the new contract: no config → `region_name=None` on every default client; explicit config → threaded through all three (runtime / agent / s3). `ruff` + the full `bedrock_knowledge_base` suite (74 tests) pass.

@liramon2 — the `(#3564)` tracker reference and the `SageMakerAIModel` comparison are removed from the test docstring as well.

liramon2:
@strandly-the-agent Re-review this PR. Consolidate your findings in a single comment.

strandly-the-agent:
**Verdict: approve-with-conditions.** Both blockers from my last review are fixed — no more `us-west-2` misrouting, and the rewritten tests are hermetic and assert real client kwargs. What's left is doc accuracy, one test-pinning gap, and `liramon2`'s still-unmet site-docs ask; none of it is a design blocker.

**Since my last review:** `5947f04d` switched to "region as a hint, let boto resolve the chain" per `opieter-aws` — verified working. `liramon2`'s three inline nits are genuinely done in the code, not just marked resolved.

**Ledger** (branch `pr3583` @ `5947f04d` vs `origin/main`):
- ✅ `opieter-aws`'s exact case fixed: `~/.aws/config` `[default] region = ca-central-1`, no env → store resolves `ca-central-1` (was `us-west-2`). `AWS_DEFAULT_REGION=eu-central-1` → `eu-central-1`.
- ✅ Tests: 74/74 in the KB dir (116 across `vended_memory_stores/`), now hermetic — pass with `AWS_PROFILE=dev` and with hostile ambient `AWS_REGION`/`AWS_DEFAULT_REGION`, zero network calls, stable under `-n auto`. `mypy`/`ruff` clean; `DEFAULT_REGION` and `import os` fully removed.
- ✅ The design's load-bearing assumption holds: `region_name=None` is genuinely identical to omitting the kwarg (checked across 15 environments × 3 services, incl. FIPS, dualstack, GovCloud, China, `AWS_ENDPOINT_URL`, `defaults_mode`, comparing endpoints *and* the signed SigV4 scope).
- 🟡 Three shipped prose sites (plus the PR comment) claim boto3's chain reads `AWS_REGION` and IMDS — it reads neither (verified: `AWS_REGION` alone → `NoRegionError`, even with a live IMDS returning a region).
- 🟡 `liramon2`'s site-docs ask is still unmet — zero `site/` files in the diff.
- 🟡 One test-pinning gap, and one injected-client interaction worth a second look.

<details>
<summary><strong>🟡 A — the new prose documents a boto3 chain that doesn't exist (ready-to-paste fix, 3 sites)</strong></summary>

`store.py:131`, the **public** docstring `types.py:83-84`, and `test_store.py:229-230` all say boto3 resolves "(`AWS_REGION` env, `~/.aws/config`, IMDS)". Verified first-hand on botocore 1.43.62 — the real chain is the `region_name` kwarg → `AWS_DEFAULT_REGION` env → the `~/.aws/config` profile `region`. `botocore/configprovider.py` maps `'region' → ('region', 'AWS_DEFAULT_REGION', None, None)`, and the live provider list is `[InstanceVarProvider, EnvironmentProvider(AWS_DEFAULT_REGION), ScopedConfigProvider(region)]` — no `AWS_REGION`, no IMDS step (`IMDSRegionProvider` exists but is wired only into `defaults_mode` latency tuning).

This isn't a regression versus `main` — behaviour there is the same. It's a documentation defect with a real consequence: someone reads the `region_name` docstring, exports `AWS_REGION`, and gets a construction-time `NoRegionError`. It's also the one region path in `strands-py/src` with that gap — `models/bedrock.py:192` and `models/sagemaker.py:162` both read `AWS_REGION`, so the same host gives a working `BedrockModel` and a failing KB store.

Suggested replacement for the parenthetical, in all three places:

```
boto3 resolves the region from its own chain instead (``AWS_DEFAULT_REGION`` env, then the
profile ``region`` in ``~/.aws/config`` — not ``AWS_REGION``, and with no IMDS step). If none
of those resolve, construction raises ``botocore.exceptions.NoRegionError``.
```

While you're in there: `__init__`'s `Raises:` block (`store.py:105-107`) documents only `ValueError`, but construction can now also raise `botocore.exceptions.NoRegionError` and `InvalidRegionError` — neither is a `ValueError` subclass, so an `except ValueError` around construction won't catch them.
</details>

<details>
<summary><strong>🟡 B — <code>liramon2</code>'s site-docs directive is still unmet</strong></summary>

The diff is still 3 files, all under `strands-py/`. What's actually missing is smaller than it looks:

- `site/src/content/docs/user-guide/concepts/memory/bedrock-knowledge-base.mdx:99-107` (Connection config table) has **no `region_name` row** — the one new public field is undocumented.
- `:347` "credential **and region** setup shared with the Bedrock model provider" is still false, for a new reason: `amazon-bedrock.mdx:81-86` documents `BedrockModel`'s chain as `region_name → session region → AWS_REGION → us-west-2`; this store's is `region_name → boto3's own chain → NoRegionError`.
- `:17` and `:107` ("standard AWS credential chain") are about **credentials**, which this PR never touched — those are accurate, please don't "fix" them.
- A `<Syntax py ts>` row won't work for this field (`Syntax.astro:5-6` requires both props and TS has no `regionName`) — use a plain code span + "(Python only)", matching `streaming/index.mdx:14`.

I have a diff-ready `.mdx` patch (frontmatter `sourceLinks`, the table row, a short `region_name` example, the corrected `:347` bullet) — say the word and I'll paste it. Holding it back to keep this comment readable. No changelog entry is needed; `site/src/content/changelog/` is generated from releases.
</details>

<details>
<summary><strong>🟡 C — a test-pinning gap, and one interaction re-raised with new evidence</strong></summary>

**The tests can't detect a reintroduction of the env fallback that was just removed.** `test_store.py:236-237` and `:278-279` call `monkeypatch.delenv("AWS_REGION"/"AWS_DEFAULT_REGION")` — which clears exactly the variable a regression would read. I mutated `store.py` to `self._region = kb_config.get("region_name") or os.getenv("AWS_REGION")` (precisely what `opieter-aws` asked to remove) and the suite stayed green: `74 passed`. It's caught only incidentally, and only when the developer's shell happens to have `AWS_REGION` set — and then by `TestConstructor`, not by either of the two tests written to guard this.

One line each fixes it — poison instead of clear:

```python
monkeypatch.setenv("AWS_REGION", "zz-store-must-not-read-this")
monkeypatch.setenv("AWS_DEFAULT_REGION", "zz-store-must-not-read-this-either")
```

The accompanying `setenv("HOME")` / `setenv("AWS_CONFIG_FILE")` lines in those two tests are dead — with `boto3.client` patched nothing resolves, and the tests pass identically with those lines deleted under a hostile `HOME` pointing at a real `~/.aws/config`. Worth deleting: as written they imply the test exercises boto3's chain, which it can't.

**Injected client + `region_name`** — I raised this last round as an appendix nit ("silently splits regions"); the new revision gives it teeth, so here it is once with evidence. `region_name="eu-west-1"` plus an injected `runtime_client` in `us-east-1` produces, verified:

```
injected runtime -> https://bedrock-agent-runtime.us-east-1.amazonaws.com
default agent    -> https://bedrock-agent.eu-west-1.amazonaws.com
default s3       -> eu-west-1
```

`search()` → `initialize()` calls `self._get_agent_client().get_knowledge_base(...)` (`store.py:179`), i.e. the **default** client in the hint region rather than the injected one, so the read path looks for the KB in the wrong region and the error names the knowledge base id, never the region config. (The endpoint split is wire-verified; the specific `ResourceNotFoundException` shape is simulated, not a live call.) Both fields are public and documented, so this is reachable by anyone who wants a custom-configured client *and* a pinned region — and it wasn't reachable before this PR. Minimal fix: one `logger.warning` in `__init__` when `region_name` is set and an injected client's `meta.region_name` differs (needs a `getattr` guard — test doubles have no `.meta`), plus a test pinning the "ignored for injected clients" claim at `types.py:84-85`, which currently has none (a mutant that makes the hint *override* an injected client survives the suite).
</details>

<details>
<summary><strong>⚪ D — housekeeping before merge (bundled, non-blocking)</strong></summary>

- **The PR body and title still describe the deleted design** and get squashed into the merge commit: the `os.getenv("AWS_REGION") or "us-west-2"` resolution, "Resolution order (mirrors `SageMakerAIModel`)" (which reintroduces in the body the cross-reference `liramon2`'s nit removed from the code), the Before/After comment promising a us-west-2 fallback, the whole "Why a default, not 'operators should pass a client'" section arguing for what `opieter-aws` rejected, a Testing section describing tests that no longer exist ("constructs a real default client", "`AWS_REGION` winning when no explicit config is given"), and both documentation checkboxes ticked with zero `site/` changes. The commit message on `5947f04d` is accurate — worth retitling the PR to match it.
- **"Resolves: #3564"** is a stretch now: the reporter's literal environment (EC2 + IAM role, no `AWS_REGION`/`AWS_DEFAULT_REGION`, no profile) still raises the exact error the issue reports, and their snippet's key is `region`, which is silently ignored. `Addresses:` plus a short note on the issue (the field is `region_name`; IMDS region pickup isn't reachable through boto3's chain and would need new code) would leave the reporter better off than an auto-close.
- `store.py:130-133` says `None` is passed "rather than the store silently picking one" — true for the two Bedrock clients, but a default `s3` client with no resolvable region silently picks `us-east-1` / `https://s3.amazonaws.com`. Pre-existing on `main` and low impact (S3's region redirector self-heals in one extra round trip), so this is a one-clause caveat in the comment, not a code change.
- `region_name=""` now raises `ValueError: Invalid endpoint: https://bedrock-agent-runtime..amazonaws.com`, where the old `or` chain absorbed it. `kb_config.get("region_name") or None` if you want empty-means-absent.
- The class/module docstring examples (and the site quickstart) raise `NoRegionError` when run verbatim with no ambient AWS config — a regression versus the *previous* revision, not versus `main`. A commented `# "region_name": "us-east-1",` line in the example would help the exact host class #3564 was filed about.
- Worth noting in the PR's favour: setting `region_name` also fixes a pre-existing bug where changing `AWS_DEFAULT_REGION` between construction and first use split the store across regions (eager runtime in the old region, lazy clients in the new one — reproduced on `main` too). And the body's "existing callers are unaffected" claim is now genuinely true.
- Not re-raising (unchanged since last round, appendix-level): no `validate_config_keys`, unvalidated `region_name` values, no debug log of the region, TS parity (`store.ts:249,660,667`), `types.py:89`'s missing `knowledge_base_type` entry. `api/needs-review` still not needed per `team/API_BAR_RAISING.md:20-21`.
</details>

## Questions

1. `SuperMarioYL` / `opieter-aws` — is a narrow consistency fallback wanted so an `AWS_REGION`-only host stops hard-failing, e.g. `kb_config.get("region_name") or boto3.Session().region_name or os.getenv("AWS_REGION")`? That keeps boto3's own resolution ahead of the env read and still fails loudly when nothing resolves, so it shouldn't reintroduce the override problem you rejected — but it is a design call, not something I'd assert is required. The alternative is simply to document that `AWS_REGION` is not consulted (finding A), which is the minimum either way.
2. Is the injected-client + `region_name` split worth the warning now, or fine as a follow-up given it takes two features combined deliberately?

**To clear the conditions:** fix the three `AWS_REGION`/IMDS claims and the `Raises:` block (A), the test poison + dead env lines (C), and either land `liramon2`'s site-docs row or explicitly defer it (B). Everything else above is optional.

---
<sub>Produced by `strandly-the-agent` running a multi-pass review pipeline (correctness · issue-alignment · adversarial · test-quality · docs-accuracy). Findings for a human to confirm, not a final verdict.</sub>

liramon2:
@SuperMarioYL Thanks for the update. Do remember to document the new field (`region_name`) in the site docs. The relevant one is `bedrock-knowledge-base.mdx`.

strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
```diff
@@ -126,10 +126,17 @@ def __init__(self, **store_config: Unpack[BedrockKnowledgeBaseStoreConfig]) -> N
         self._data_source_type = kb_config.get("data_source_type")
         self._data_source_id = kb_config.get("data_source_id")
 
+        # Region hint for any default boto3 client the store constructs. Only an explicit
+        # ``region_name`` config is threaded through; when absent, ``None`` is passed so boto3
+        # resolves the region from its own chain (``AWS_REGION`` env, ``~/.aws/config``, IMDS)
```
liramon2: boto3 uses `AWS_DEFAULT_REGION` env and then `~/.aws/config` (no IMDS), according to https://github.com/boto/botocore/blob/develop/botocore/configprovider.py#L71. Do change the comments to reflect this.

strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/types.py
```diff
@@ -79,6 +79,10 @@ class BedrockKnowledgeBaseConfig(TypedDict, total=False):
         agent_client: Pre-constructed agent client for ``IngestKnowledgeBaseDocuments`` calls. When
             omitted, a default boto3 client is constructed lazily on first write. To target a specific
             region/credentials/endpoint, build the client yourself and inject it here.
+        region_name: AWS region hint for the default ``bedrock-agent-runtime`` / ``bedrock-agent`` / ``s3``
+            clients constructed by the store. When omitted, ``None`` is passed and boto3 resolves the
+            region from its own chain (``AWS_REGION`` env, ``~/.aws/config``, IMDS). Ignored for any
```
liramon2: Same as above

liramon2 (changes requested):

strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
SuperMarioYL: Done — the comment now reflects boto3's actual chain: `AWS_DEFAULT_REGION` env, then `~/.aws/config` (no IMDS), per the botocore link you cited. Mirrored in `types.py` too. (13d5ac1b)

strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/types.py
SuperMarioYL: Done — mirrored the corrected chain (`AWS_DEFAULT_REGION` env → `~/.aws/config`, no IMDS) here too. (13d5ac1b)

SuperMarioYL:
Thanks @liramon2. Both items from your latest pass are addressed in 13d5ac1b:

- **Comments (`store.py` / `types.py`):** corrected the region-resolution chain to `AWS_DEFAULT_REGION` env → `~/.aws/config` (no IMDS), per the botocore link you cited — replied in both threads.
- **Site docs:** added a `region_name` row to the `bedrock-knowledge-base.mdx` config table, describing the hint and the boto3 fallback chain.

Ready for another look when you have a moment.

liramon2 (dismissed):

site/src/content/docs/user-guide/concepts/memory/bedrock-knowledge-base.mdx
```diff
@@ -105,6 +105,7 @@ The inner `BedrockKnowledgeBaseConfig` is the reusable connection: which knowled
 | `s3` | S3 ingestion settings (`bucket`, `prefix`). Required when the data source type is `'S3'`. |
 | <Syntax py="scope_metadata_key" ts="scopeMetadataKey" /> | Metadata attribute key used for scope filtering. Defaults to `'namespace'`. |
 | <Syntax py="runtime_client / agent_client" ts="runtimeClient / agentClient" /> | Pre-constructed AWS clients. When omitted, default clients are constructed using the standard credential chain (the agent client lazily, on first write). |
+| <Syntax py="region_name" ts="regionName" /> | AWS region hint for the default clients the store constructs. When omitted, boto3 resolves the region from its own chain (`AWS_DEFAULT_REGION` env, then `~/.aws/config` — no IMDS). Ignored for any client you inject explicitly (`runtime_client` / `agent_client` / `s3.client`). |
```
opieter-aws: Why  is imds referenced here? I'd be in favor of not owning documentation specifying boto3's resolving chain: that means we need to maintain this. We can just mention that boto3 resolves  the region itself if omitted

strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
```diff
@@ -126,10 +126,18 @@ def __init__(self, **store_config: Unpack[BedrockKnowledgeBaseStoreConfig]) -> N
         self._data_source_type = kb_config.get("data_source_type")
         self._data_source_id = kb_config.get("data_source_id")
 
+        # Region hint for any default boto3 client the store constructs. Only an explicit
+        # ``region_name`` config is threaded through; when absent, ``None`` is passed so boto3
+        # resolves the region from its own chain (``AWS_DEFAULT_REGION`` env, then
+        # ``~/.aws/config`` — no IMDS) rather than the store silently picking one — a wrong
```
opieter-aws: Ditto

strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/types.py
```diff
@@ -79,6 +79,11 @@ class BedrockKnowledgeBaseConfig(TypedDict, total=False):
         agent_client: Pre-constructed agent client for ``IngestKnowledgeBaseDocuments`` calls. When
             omitted, a default boto3 client is constructed lazily on first write. To target a specific
             region/credentials/endpoint, build the client yourself and inject it here.
+        region_name: AWS region hint for the default ``bedrock-agent-runtime`` / ``bedrock-agent`` / ``s3``
+            clients constructed by the store. When omitted, ``None`` is passed and boto3 resolves the
+            region from its own chain (``AWS_DEFAULT_REGION`` env, then ``~/.aws/config`` — no IMDS).
```
opieter-aws: ditto here

SuperMarioYL:
Thanks @opieter-aws — agreed, owning a description of boto3's resolving chain is maintenance we don't want. Dropped the chain specifics (`AWS_DEFAULT_REGION` / `~/.aws/config` / IMDS) from all three spots and left a minimal "boto3 resolves the region itself when the hint is omitted": the site docs (`bedrock-knowledge-base.mdx`), the `store.py` comment, and the `types.py` docstring. The injected-clients caveat is kept. Pushed in `9df7f174`.

strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
```diff
@@ -126,10 +126,17 @@ def __init__(self, **store_config: Unpack[BedrockKnowledgeBaseStoreConfig]) -> N
         self._data_source_type = kb_config.get("data_source_type")
         self._data_source_id = kb_config.get("data_source_id")
 
+        # Region hint for any default boto3 client the store constructs. Only an explicit
+        # ``region_name`` config is threaded through; when absent, ``None`` is passed so boto3
+        # resolves the region itself rather than the store silently picking one — a wrong
+        # default would misroute region-scoped knowledge bases. Injected clients are used
+        # verbatim and bypass this.
```
opieter-aws:
```suggestion
        # resolves the region itself rather than the store silently picking one.
        # Injected clients bypass this and use their own region setup.
```

strands-py/src/strands/vended_memory_stores/bedrock_knowledge_base/store.py
```diff
@@ -126,10 +126,16 @@ def __init__(self, **store_config: Unpack[BedrockKnowledgeBaseStoreConfig]) -> N
         self._data_source_type = kb_config.get("data_source_type")
         self._data_source_id = kb_config.get("data_source_id")
 
+        # Region hint for any default boto3 client the store constructs. Only an explicit
+        # ``region_name`` config is threaded through; when absent, ``None`` is passed so boto3
+        # resolves the region itself rather than the store silently picking one.
+        # Injected clients bypass this and use their own region setup.
```
opieter-aws:
```suggestion
        # resolves the region itself. Injected clients bypass this and use their own region setup.
```

strands-py/tests/strands/vended_memory_stores/bedrock_knowledge_base/test_store.py
```diff
@@ -209,14 +209,96 @@ def test_allows_writable_with_s3_data_source(self, make_s3_store):
...
+# --------------------------------------------------------------------------- #
+
+
+class TestRegionResolution:
+    """Default boto3 clients receive the configured region as a hint; otherwise boto3 resolves it.
+
+    Only an explicit ``region_name`` config is threaded into the store's default ``boto3.client``
+    calls. When it is absent, ``None`` is passed so boto3 resolves the region from its own chain
+    (``AWS_REGION`` env, ``~/.aws/config``, IMDS) rather than the store silently picking one — a
+    wrong default region would misroute region-scoped knowledge bases.
```
opieter-aws:
```suggestion
    Only an explicit ``region_name`` config is threaded into the store's default ``boto3.client``
    calls. When it is absent, ``None`` is passed so boto3 resolves the region from its own chain.
```

opieter-aws (approved):

liramon2 (approved):

PR #3585, opened 2026-07-31
feat(a2a): support the interrupt round trip in TypeScript

### LINKED ISSUE #3584: [PORT] A2A interrupt round trip (Py→TS)
## Source SDK

strands-py

## Target SDK

strands-ts

## Related Files

```
strands-py/src/strands/multiagent/a2a/executor.py
strands-py/tests/strands/multiagent/a2a/test_executor.py
site/src/content/docs/user-guide/concepts/multi-agent/agent-to-agent.mdx
```

Target files: `strands-ts/src/a2a/executor.ts`, `strands-ts/src/a2a/adapters.ts`, `strands-ts/src/a2a/__tests__/executor.test.ts`

## What needs to be migrated

#3486 made the A2A interrupt round trip work in Python (#2948). `strands-ts` has the same gap, in both directions.

**Outbound — currently reports success.** `_streamAgent` publishes `status: { state: 'completed' }` unconditionally once the stream ends (`executor.ts:313`), never checking `stopReason`. An agent that stopped with `stopReason: 'interrupt'` therefore reports a *completed* task to the client while the paused tool never ran. Python maps this to `input_required` and advertises the pending interrupts as a DataPart.

**Inbound — identical to the Python bug.** `partsToContentBlocks` flattens every DataPart to text (`adapters.ts:34`):

```typescript
case 'data':
  blocks.push(new TextBlock(`[Structured Data]\n${JSON.stringify(part.data, null, 2)}`))
```

That was byte-for-byte the Python defect, down to the `[Structured Data]` prefix.

**The wire format needs no new decisions.** It is already the JSON form of the TS type — `InterruptResponseContent.toJSON()` in `types/interrupt.ts` emits exactly `{interruptResponse: {interruptId, response}}`, `fromJSON()` reads it back, and `isInterruptResponseContent` guards on it. So the merged contract ports as-is:

```json
// server -> client, on input-required
{"kind": "data", "data": {"interrupts": [{"interruptId": "...", "name": "...", "reason": ...}]}}

// client -> server, resuming
{"kind": "data", "data": {"interruptResponse": {"interruptId": "...", "response": ...}}}
```

**The agent side is already there.** `InterruptState` carries `interrupts`, `activated` and `resumeResponses`; `AgentInput` accepts `InterruptResponseContent[]`; `agent.ts` has `_extractInterruptResponses` and `isInterruptResponseContent`. Only the A2A adapter is missing.

Behaviour to match, all validated before the agent runs so a refused answer leaves the interrupt pending:

- recognise only an explicit `interruptResponse` DataPart; other DataParts keep the generic path
- bind each `interruptId` to an interrupt actually parked on that context's agent
- reject duplicates, malformed payloads, and interrupt responses mixed with other content parts
- reject a **null** `response` — `Interrupt.response` of null means "not yet answered", so a null answer silently re-fires the same interrupt (this bit us in review on #3486)
- reject an ordinary conversational message while a task is parked

Docs are a shared page: `agent-to-agent.mdx` currently carries a `:::note[Python only]` on the **Answering Interrupts** section, to be removed and given a TypeScript tab when this lands.

Happy to take this one.

## Confirmation

vaibhav-walia:
## Description

Human-in-the-loop did not work over A2A in the TypeScript SDK, and failed silently in both directions.

- **Outbound**: `_streamAgent` published `state: 'completed'` once the stream ended, without checking `stopReason`. A task whose tool had paused for approval reported **success** while the tool never ran.
- **Inbound**: `partsToContentBlocks` flattened every data part to `[Structured Data]` text, so an interrupt response arrived as a plain conversational turn and the agent never resumed.

This ports the contract merged for Python in #3486. The agent side was already in place — `InterruptState` carries `interrupts`/`activated`/`resumeResponses`, and `AgentInput` accepts `InterruptResponseContent[]`; only the A2A adapter was missing.

Resume is treated as a continuation of the parked state rather than a new conversational turn: once a task is in `input-required`, the next message for it must carry an interrupt response, and every rejection happens before the agent runs, so a refused answer leaves the interrupt pending and retryable.

A refused turn also publishes no events at all. Task registration moved from `execute()` to just before the stream starts, because registering first would move a task parked in `input-required` to `working` and then throw, leaving the client unable to see that the task is still waiting on them.

## Public API Changes

Two data-part contracts on the A2A wire, identical to the Python ones. No TypeScript signature changes.

**Outbound** — the `input-required` status message carries the pending interrupts. The existing text part stays first, so clients that only read prose are unaffected:

```json
{
  "kind": "data",
  "data": { "interrupts": [{ "interruptId": "tool:tool-1:approveCampaign", "name": "approveCampaign", "reason": { "campaign": "spring-launch" } }] }
}
```

**Inbound** — a client resumes by echoing the id back:

```json
{
  "kind": "data",
  "data": { "interruptResponse": { "interruptId": "tool:tool-1:approveCampaign", "response": { "approved": true } } }
}
```

This needed no format decisions: it is already the JSON form of the TS type. `InterruptResponseContent.toJSON()` emits exactly this shape, `fromJSON()` reads it back, and `isInterruptResponseContent` guards on it — so both SDKs speak it without a mapping layer.

### Rejected inputs

All raise `A2AError.invalidParams` before the agent is invoked:

| Input | Why |
| --- | --- |
| Interrupt id not parked on this context's agent | An answer must bind to the interrupt that requested it |
| Interrupt response with no interrupt pending | Nothing to resume |
| Two responses for the same interrupt id | Ambiguous; picking one silently would discard a decision |
| A `null` response | `Interrupt.response` of null reads as "not yet answered", so it would silently re-fire the same interrupt. `false` and `0` are fine |
| Missing/blank `interruptId`, non-object payload | Malformed |
| Interrupt response mixed with other content parts | The agent resumes from responses alone, so the rest would be silently dropped |
| A fresh conversational turn while a task is parked | A parked task is answered or cancelled, not talked past |

A data part without an `interruptResponse` key is untouched and still takes the generic path.

## Related Issues

Closes #3584. Ports #3486, which fixed the same gap in Python (#2948).

## Type of Change

New feature

## Testing

Verified by running the same script against unmodified `main` and this branch. It drives a real `Agent` with a real interrupting tool through `executor.execute()` and reads only the published events — the surface a client actually has.

| Observation | `main` | this branch |
| --- | --- | --- |
| Task state when the tool pauses | `completed` | `input-required` |
| Interrupt ids sent to the client | none | advertised in a data part |
| Answering with the parked id | rejected by the agent | `completed` |
| Tool actually resumed | `false` | `true` |
| Payload the tool received | — | `{"approved":true}` |

The `main` column is the bug in full: the first turn reports `completed` while the tool log reads only `["entered"]`, so the client is told the work succeeded when the approval never happened. Answering then fails with *"Agent is in an interrupted state. Resume by invoking with interruptResponse content blocks"* — the agent rejecting a text turn, because the data part had been flattened before it arrived.

Unit coverage adds 13 cases to `src/a2a/__tests__/executor.test.ts`, including a round trip against a real agent that asserts the advertised id matches the one the agent parked, and that the resumed tool receives the client's payload rather than merely that the task completed. Another asserts a refused answer publishes no events, so the task stays visibly parked in `input-required` for the client rather than being stranded in `working`.

Shared-agent mode was checked separately: a parked interrupt survives the snapshot swap, since `interrupts` is part of the `session` snapshot preset.

Docs: the **Answering Interrupts** section on `agent-to-agent.mdx` loses its `:::note[Python only]` and gains a TypeScript tab.

Run against this branch: `npm run lint` reports no problems, `npm run format` leaves the tree unchanged, `npm test` is green at 3901 tests across 141 files, `tsc --noEmit --project src/tsconfig.json` reports no errors, and `npm run build` in `site/` finds no broken links.

- [ ] I ran `hatch run prepare` — not applicable; this is a TypeScript-only change, so the equivalent TypeScript checks are listed above

## Checklist

----

### DIFF
--- a/site/src/content/docs/user-guide/concepts/multi-agent/agent-to-agent.mdx
+++ b/site/src/content/docs/user-guide/concepts/multi-agent/agent-to-agent.mdx
@@ -381,15 +381,11 @@ In single-agent mode the server reuses one agent across every context, swapping 
 
 ### Answering Interrupts
 
-:::note[Python only]
-This feature is currently available in the Python SDK only.
-:::
+When a tool or hook on the served agent raises an [interrupt](../interrupts.md), the task moves to the A2A `input-required` state and waits. The client answers it, and the task resumes the paused tool exactly where it stopped.
 
-When a tool or hook on the served agent raises an [interrupt](../interrupts.md), the task moves to the A2A `input_required` state and waits. The client answers it, and the task resumes the paused tool exactly where it stopped.
+Answering an interrupt is the one flow on this page that needs the raw A2A client rather than `A2AAgent`. `A2AAgent` speaks in text, so it rejects interrupt responses on the way out and drops the data part carrying the interrupt ids on the way back. The examples below build A2A messages directly.
 
-Answering an interrupt is the one flow on this page that needs the raw [`a2a-sdk`](https://github.com/a2aproject/a2a-python) client rather than `A2AAgent`. `A2AAgent` speaks in text, so it raises `ValueError` if you pass it interrupt responses, and it drops the `DataPart` carrying the interrupt ids when it reads the reply. The examples below build A2A messages directly.
-
-Each interrupt has a server-generated id, and an answer is bound to the id of the interrupt that raised it. The server advertises the pending interrupts on the `input_required` status message as a `DataPart`, alongside the human-readable `TextPart`:
+Each interrupt has a server-generated id, and an answer is bound to the id of the interrupt that raised it. The server advertises the pending interrupts on the `input-required` status message as a data part, alongside the human-readable text part:
 
 ```json
 {
@@ -420,14 +416,17 @@ To answer, send a new message on the same `taskId` containing a `DataPart` that 
 }
 ```
 
-The `response` becomes the return value of the `interrupt()` call that paused the tool. It is any JSON value except `null`, which the server refuses — a null answer would leave the interrupt unsatisfied and re-raise it. `false` and `0` are fine. Answer several interrupts in one message by sending one `DataPart` for each.
+The `response` becomes the return value of the `interrupt()` call that paused the tool. It is any JSON value except `null`, which the server refuses — a null answer would leave the interrupt unsatisfied and re-raise it. `false` and `0` are fine. Answer several interrupts in one message by sending one data part for each.
 
 Reading the ids off the status message and answering them:
+
+<Tabs>
+<Tab label="Python">
 
 ```python
 from a2a.types import DataPart, Part
 
-# The task parked in input_required; read the interrupts it is waiting on.
+# The task parked in input-required; read the interrupts it is waiting on.
 pending = next(
     part.root.data["interrupts"]
     for part in task.status.message.parts
@@ -442,6 +441,25 @@ answers = [
     for item in pending
 ]
 ```
+
+</Tab>
+<Tab label="TypeScript">
+
+```typescript
+// The task parked in input-required; read the interrupts it is waiting on.
+const pending = (task.status.message?.parts ?? []).flatMap((part) =>
+  part.kind === 'data' && 'interrupts' in part.data ? (part.data.interrupts as { interruptId: string }[]) : []
+)
+
+// Answer each one on the same taskId.
+const answers = pending.map((item) => ({
+  kind: 'data' as const,
+  data: { interruptResponse: { interruptId: item.interruptId, response: { approved: true } } },
+}))
+```
+
+</Tab>
+</Tabs>
 
 The server rejects an answer it cannot bind, before the agent runs, so a refused answer leaves the interrupt pending and the task still answerable. An answer is rejected when:
 
--- a/strands-ts/src/a2a/adapters.ts
+++ b/strands-ts/src/a2a/adapters.ts
@@ -5,12 +5,96 @@
  */
 
 import type { Part, FileWithBytes, FileWithUri } from '@a2a-js/sdk'
+import { A2AError } from '@a2a-js/sdk/server'
 import type { ContentBlock } from '../types/messages.js'
 import { TextBlock } from '../types/messages.js'
 import type { ImageFormat, DocumentFormat, VideoFormat } from '../mime.js'
 import { toMimeType, toMediaFormat } from '../mime.js'
 import { ImageBlock, VideoBlock, DocumentBlock, decodeBase64, encodeBase64 } from '../types/media.js'
+import { InterruptResponseContent } from '../types/interrupt.js'
+import type { JSONValue } from '../types/json.js'
 import { logger } from '../logging/logger.js'
+
+/**
+ * Key identifying a data part that carries a Strands interrupt response. The A2A payload mirrors
+ * the `InterruptResponseContent` type verbatim, so the wire contract and the SDK type cannot drift.
+ */
+export const INTERRUPT_RESPONSE_KEY = 'interruptResponse'
+
+/**
+ * Key under which an `input-required` status message advertises the interrupts awaiting an answer.
+ * Each entry carries the `interruptId` that the matching interrupt response must echo back.
+ */
+export const INTERRUPTS_KEY = 'interrupts'
+
+/**
+ * Extracts Strands interrupt responses from inbound A2A message parts.
+ *
+ * A client resumes a task parked in `input-required` by sending a data part shaped like the
+ * `InterruptResponseContent` type:
+ *
+ * ```json
+ * { "kind": "data", "data": { "interruptResponse": { "interruptId": "<id>", "response": <any> } } }
+ * ```
+ *
+ * Recognition is deliberately narrow: only the explicit shape above is treated as a resume, so an
+ * ordinary data part still reaches the generic content-block path unchanged.
+ *
+ * @param parts - Array of A2A protocol parts from the inbound message
+ * @returns The interrupt responses carried by `parts`, or undefined when none are present and the
+ *   caller should fall back to generic content-block conversion
+ * @throws A2AError when an interrupt response is malformed, carries a null response, repeats an
+ *   interrupt id, or is accompanied by unrelated content in the same message
+ */
+export function extractInterruptResponses(parts: Part[]): InterruptResponseContent[] | undefined {
+  const responses: InterruptResponseContent[] = []
+  const seenIds = new Set<string>()
+  let unrelatedParts = 0
+
+  for (const part of parts) {
+    if (part.kind !== 'data' || !(INTERRUPT_RESPONSE_KEY in part.data)) {
+      unrelatedParts += 1
+      continue
+    }
+
+    const payload = part.data[INTERRUPT_RESPONSE_KEY]
+    if (typeof payload !== 'object' || payload === null || Array.isArray(payload)) {
+      throw A2AError.invalidParams(`'${INTERRUPT_RESPONSE_KEY}' must be an object with 'interruptId' and 'response'`)
+    }
+
+    const { interruptId, response } = payload as { interruptId?: unknown; response?: unknown }
+    if (typeof interruptId !== 'string' || interruptId.length === 0) {
+      throw A2AError.invalidParams("Interrupt response is missing a non-empty 'interruptId'")
+    }
+
+    // An interrupt whose response is null reads as "not yet answered", so a null answer would leave
+    // the interrupt unsatisfied and re-raise it: the client would see an identical input-required
+    // and no error. Falsy answers such as false are fine.
+    if (response === undefined || response === null) {
+      throw A2AError.invalidParams(`Interrupt response for '${interruptId}' must provide a non-null 'response'`)
+    }
+
+    // Two answers for one interrupt are ambiguous; reject rather than silently choosing one.
+    if (seenIds.has(interruptId)) {
+      throw A2AError.invalidParams(`Duplicate interrupt response for '${interruptId}'`)
+    }
+
+    seenIds.add(interruptId)
+    responses.push(new InterruptResponseContent({ interruptId, response: response as JSONValue }))
+  }
+
+  if (responses.length === 0) {
+    return undefined
+  }
+
+  // The agent resumes from interrupt responses alone, so delivering both would mean dropping the
+  // conversational content — exactly the silent behaviour the resume path must avoid.
+  if (unrelatedParts > 0) {
+    throw A2AError.invalidParams('A message carrying interrupt responses must not contain other content parts')
+  }
+
+  return responses
+}
 
 /**
  * Converts A2A protocol parts to Strands SDK content blocks.
--- a/strands-ts/src/a2a/executor.ts
+++ b/strands-ts/src/a2a/executor.ts
@@ -8,11 +8,15 @@
 import type { ExecutionEventBus, RequestContext } from '@a2a-js/sdk/server'
 import type { AgentExecutor } from '@a2a-js/sdk/server'
 import { A2AError } from '@a2a-js/sdk/server'
+import type { Part } from '@a2a-js/sdk'
 import type { InvokableAgent, LocalAgent } from '../types/agent.js'
 import type { Snapshot } from '../types/snapshot.js'
 import type { ContentBlock } from '../types/messages.js'
+import type { Interrupt, InterruptState } from '../interrupt.js'
+import type { InterruptResponseContent } from '../types/interrupt.js'
+import { isInterruptResponseContent } from '../types/interrupt.js'
 import { ModelStreamUpdateEvent, ContentBlockEvent } from '../hooks/events.js'
-import { contentBlocksToParts, partsToContentBlocks } from './adapters.js'
+import { INTERRUPTS_KEY, contentBlocksToParts, extractInterruptResponses, partsToContentBlocks } from './adapters.js'
 import { normalizeError } from '../errors.js'
 import { logger } from '../logging/logger.js'
 import { AsyncLock } from './async-lock.js'
@@ -41,6 +45,23 @@ export interface A2AExecutorOptions {
   agentFactory?: AgentFactory
   /** Maximum contexts to retain; the least-recently-used is evicted beyond it. Must be at least 1. */
   maxContexts?: number
+}
+
+/**
+ * What this executor invokes the agent with: fresh conversational content, or interrupt responses
+ * resuming a task parked in `input-required`. Each arm stays homogeneous because the agent rejects
+ * a prompt whose contents are not all interrupt responses.
+ */
+type AgentPrompt = ContentBlock[] | InterruptResponseContent[]
+
+/** Whether a prompt carries interrupt responses rather than conversational content. */
+function isInterruptResume(prompt: AgentPrompt): prompt is InterruptResponseContent[] {
+  return prompt.length > 0 && isInterruptResponseContent(prompt[0])
+}
+
+/** Reads the agent's interrupt state, a field not declared on {@link InvokableAgent}. */
+function interruptStateOf(agent: InvokableAgent): InterruptState | undefined {
+  return (agent as { _interruptState?: InterruptState })._interruptState
 }
 
 /** A full Strands `Agent` that can be both invoked and snapshotted. */
@@ -192,44 +213,44 @@ export class A2AExecutor implements AgentExecutor {
    * @param eventBus - The event bus for publishing A2A artifact and status events
    */
   async execute(context: RequestContext, eventBus: ExecutionEventBus): Promise<void> {
-    const { taskId, contextId, userMessage } = context
-    const contentBlocks = partsToContentBlocks(userMessage.parts)
-    if (contentBlocks.length === 0) {
+    const { userMessage } = context
+
+    // Interrupt responses resume a parked task, so they are recognised before the generic
+    // conversion below would flatten them into text.
+    const prompt: AgentPrompt = extractInterruptResponses(userMessage.parts) ?? partsToContentBlocks(userMessage.parts)
+    if (prompt.length === 0) {
       throw A2AError.invalidRequest('No content blocks available')
     }
 
-    // Register the task with the ResultManager; without this, later events are ignored as "unknown task".
-    eventBus.publish({ kind: 'task', id: taskId, contextId, status: { state: 'working' } })
-
     if (this._agentFactory !== undefined) {
-      await this._runWithContextAgent(context, contentBlocks, eventBus)
+      await this._runWithContextAgent(context, prompt, eventBus)
     } else {
-      await this._runWithSharedAgent(context, contentBlocks, eventBus)
+      await this._runWithSharedAgent(context, prompt, eventBus)
     }
   }
 
   /** Factory mode: run against this context's dedicated agent, serialized only per context. */
   private async _runWithContextAgent(
     context: RequestContext,
-    contentBlocks: ContentBlock[],
+    prompt: AgentPrompt,
     eventBus: ExecutionEventBus
   ): Promise<void> {
     const { agent, lock } = this._acquireContextAgent(context.contextId)
     using _release = await lock.acquire()
-    await this._streamAgent(agent, context, contentBlocks, eventBus)
+    await this._streamAgent(agent, context, prompt, eventBus)
   }
 
   /** Single-agent mode: swap this context's snapshot on/off the shared agent under a lock. */
   private async _runWithSharedAgent(
     context: RequestContext,
-    contentBlocks: ContentBlock[],
+    prompt: AgentPrompt,
     eventBus: ExecutionEventBus
   ): Promise<void> {
     const agent = this._agent!
     using _release = await this._sharedAgentLock.acquire()
     this._restoreState(agent, this._snapshots.get(context.contextId) ?? this._templateSnapshot!)
     try {
-      await this._streamAgent(agent, context, contentBlocks, eventBus)
+      await this._streamAgent(agent, context, prompt, eventBus)
     } finally {
       // Persist updated history (even on error), evict, then reset the agent for the next caller.
       this._snapshots.delete(context.contextId)
@@ -243,15 +264,27 @@ export class A2AExecutor implements AgentExecutor {
   private async _streamAgent(
     agent: InvokableAgent,
     context: RequestContext,
-    contentBlocks: ContentBlock[],
+    prompt: AgentPrompt,
     eventBus: ExecutionEventBus
   ): Promise<void> {
     const { taskId, contextId } = context
     const artifactId = globalThis.crypto.randomUUID()
     let isFirstChunk = true
 
+    // Checked before the agent runs, so a refused answer leaves the interrupt parked and retryable.
+    if (isInterruptResume(prompt)) {
+      this._validateInterruptResume(agent, prompt)
+    } else if (interruptStateOf(agent)?.activated === true) {
+      throw A2AError.invalidParams('Task is awaiting an interrupt response and cannot accept a new message')
+    }
+
+    // Registered only once the request is known to be servable. Registering earlier would move a
+    // task parked in `input-required` to `working` before a refused answer threw, leaving the client
+    // unable to see that the task is still waiting on them.
+    eventBus.publish({ kind: 'task', id: taskId, contextId, status: { state: 'working' } })
+
     try {
-      const stream = agent.stream(contentBlocks, {
+      const stream = agent.stream(prompt, {
         invocationState: { a2aRequestContext: context },
       })
       let next = await stream.next()
@@ -310,11 +343,104 @@ export class A2AExecutor implements AgentExecutor {
         lastChunk: true,
       })
 
+      // An agent that stopped on an interrupt has not finished its work, so the task waits for the
+      // client to answer rather than reporting success.
+      const result = next.value
+      if (result?.stopReason === 'interrupt') {
+        this._publishInputRequired(result.interrupts ?? [], taskId, contextId, eventBus)
+        return
+      }
+
       eventBus.publish({ kind: 'status-update', taskId, contextId, status: { state: 'completed' }, final: true })
     } catch (error) {
       logger.error(`task_id=<${taskId}> | error in streaming execution`, normalizeError(error))
       throw error
     }
+  }
+
+  /**
+   * Verifies interrupt responses bind to interrupts actually parked on this context's agent.
+   *
+   * @param agent - The agent serving this context, with its interrupt state already restored
+   * @param responses - Interrupt responses extracted from the inbound message
+   * @throws A2AError when the agent holds no parked interrupts, or a response names an interrupt
+   *   the agent is not waiting on
+   */
+  private _validateInterruptResume(agent: InvokableAgent, responses: InterruptResponseContent[]): void {
+    const state = interruptStateOf(agent)
+    if (state?.activated !== true) {
+      throw A2AError.invalidParams('Received interrupt responses but no interrupt is pending')
+    }
+
+    const unknownIds = responses
+      .map((content) => content.interruptResponse.interruptId)
+      .filter((id) => !(id in state.interrupts))
+      .sort()
+    if (unknownIds.length > 0) {
+      throw A2AError.invalidParams(`No pending interrupt matches id(s): ${unknownIds.join(', ')}`)
+    }
+  }
+
+  /**
+   * Moves the task to `input-required` and tells the client which interrupts are waiting.
+   *
+   * The interrupts are published twice: a text part describing what is needed, and a data part
+   * carrying each `interruptId`. Only the id lets a client address its answer back to the interrupt
+   * that raised it, and ids are generated server-side so they cannot be inferred from the prose.
+   *
+   * @param interrupts - The unanswered interrupts that stopped the agent
+   * @param taskId - The A2A task being paused
+   * @param contextId - The A2A context the task belongs to
+   * @param eventBus - The event bus for publishing status events
+   */
+  private _publishInputRequired(
+    interrupts: Interrupt[],
+    taskId: string,
+    contextId: string,
+    eventBus: ExecutionEventBus
+  ): void {
+    const descriptions = interrupts.map((interrupt) =>
+      interrupt.reason === undefined
+        ? `- ${interrupt.name}`
+        : `- ${interrupt.name}: ${JSON.stringify(interrupt.reason)}`
+    )
+    const text =
+      descriptions.length > 0
+        ? `Agent requires input:\n${descriptions.join('\n')}`
+        : 'Agent requires additional input to continue'
+
+    // The text part stays first so clients that only read prose are unaffected.
+    const parts: Part[] = [{ kind: 'text', text }]
+    if (interrupts.length > 0) {
+      parts.push({
+        kind: 'data',
+        data: {
+          [INTERRUPTS_KEY]: interrupts.map((interrupt) => ({
+            interruptId: interrupt.id,
+            name: interrupt.name,
+            ...(interrupt.reason === undefined ? {} : { reason: interrupt.reason }),
+          })),
+        },
+      })
+    }
+
+    eventBus.publish({
+      kind: 'status-update',
+      taskId,
+      contextId,
+      status: {
+        state: 'input-required',
+        message: {
+          kind: 'message',
+          role: 'agent',
+          messageId: globalThis.crypto.randomUUID(),
+          taskId,
+          contextId,
+          parts,
+        },
+      },
+      final: true,
+    })
   }
 
   /**
--- a/strands-ts/src/a2a/__tests__/executor.test.ts
+++ b/strands-ts/src/a2a/__tests__/executor.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it, vi } from 'vitest'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { A2AExecutor } from '../executor.js'
 import type { AgentExecutionEvent, ExecutionEventBus, RequestContext } from '@a2a-js/sdk/server'
 import type { TaskArtifactUpdateEvent, TaskStatusUpdateEvent } from '@a2a-js/sdk'
@@ -6,10 +6,12 @@ import { Agent } from '../../agent/agent.js'
 import type { InvokableAgent } from '../../types/agent.js'
 import { MockMessageModel } from '../../__fixtures__/mock-message-model.js'
 import { createMockAgent } from '../../__fixtures__/agent-helpers.js'
+import { createMockTool } from '../../__fixtures__/tool-helpers.js'
 import { TextBlock } from '../../types/messages.js'
 import { ImageBlock, encodeBase64 } from '../../types/media.js'
 import { ContentBlockEvent, ModelStreamUpdateEvent } from '../../hooks/events.js'
 import { AgentResult } from '../../types/agent.js'
+import { Interrupt } from '../../interrupt.js'
 import { Message } from '../../types/messages.js'
 import type { ModelStreamEvent } from '../../models/streaming.js'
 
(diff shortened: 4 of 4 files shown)


mkmeral:

strands-ts/src/a2a/adapters.ts
```diff
@@ -5,13 +5,97 @@
...
+ * @throws A2AError when an interrupt response is malformed, carries a null response, repeats an
+ *   interrupt id, or is accompanied by unrelated content in the same message
+ */
+export function extractInterruptResponses(parts: Part[]): InterruptResponseContent[] | undefined {
+  const responses: InterruptResponseContent[] = []
+  const seenIds = new Set<string>()
+  let unrelatedParts = 0
+
+  for (const part of parts) {
+    if (part.kind !== 'data' || !(INTERRUPT_RESPONSE_KEY in part.data)) {
```
strandly-the-agent, line 55: 🟡 **Crashes with a raw `TypeError`, not an `A2AError`, on `data: null`.**

`INTERRUPT_RESPONSE_KEY in part.data` throws `TypeError: Cannot use 'in' operator to search for 'interruptResponse' in null` when `part.data` is `null`. Verified:
```js
> 'interruptResponse' in null
Uncaught TypeError: Cannot use 'in' operator to search for 'interruptResponse' in null
```
**Who hits this**: any A2A client sending a `DataPart` with `data: null` — not necessarily an interrupt resume. `@a2a-js/sdk`'s `JsonRpcTransportHandler.paramsAreValid` and `DefaultRequestHandler._createRequestContext` (checked in `node_modules/@a2a-js/sdk/dist/server/index.js`) do no runtime validation of `Part.data`'s shape — TS's `DataPart.data: {[k:string]: unknown}` is compile-time only, so a real wire payload can carry `null` here. This function now runs unconditionally on *every* message (`execute()` calls it before `partsToContentBlocks`), so this is a regression: before this PR, a data part with `data: null` fell through to `partsToContentBlocks`'s per-part try/catch and rendered harmlessly as `[Structured Data]\nnull`; now it crashes the whole request.

The Python original (`executor.py:686-687`) has exactly the check this port dropped:
```python
data = part_root.data if isinstance(part_root, DataPart) else None
if not isinstance(data, dict) or INTERRUPT_RESPONSE_KEY not in data:
```
Suggest mirroring that: `typeof part.data !== 'object' || part.data === null || Array.isArray(part.data)` before the `in` check.

strands-ts/src/a2a/executor.ts
```diff
@@ -310,13 +343,106 @@ export class A2AExecutor implements AgentExecutor {
...
+   */
+  private _validateInterruptResume(agent: InvokableAgent, responses: InterruptResponseContent[]): void {
+    const state = interruptStateOf(agent)
+    if (state?.activated !== true) {
+      throw A2AError.invalidParams('Received interrupt responses but no interrupt is pending')
+    }
+
+    const unknownIds = responses
+      .map((content) => content.interruptResponse.interruptId)
+      .filter((id) => !(id in state.interrupts))
```
strandly-the-agent, line 377: 🔴 **Prototype pollution — `interruptId: "__proto__"` bypasses this check and pollutes `Object.prototype` process-wide.**

`state.interrupts` is a plain object (`Record<string, Interrupt>`), and `'__proto__' in {}` is `true` in JS (it's `Object.prototype`'s own accessor, found via the prototype chain) — regardless of what keys the object actually has. So `unknownIds.filter((id) => !(id in state.interrupts))` never flags `__proto__` (or `constructor`) as unknown, and validation here passes.

The request then reaches the pre-existing sink at `interrupt.ts:283` (`InterruptState.resume`, not touched by this PR but newly reachable from the wire because of it):
```js
const interrupt = this.interrupts[interruptId]   // this.interrupts['__proto__'] === Object.prototype (truthy!)
if (!interrupt) throw ...                         // never throws
interrupt.response = response                     // sets Object.prototype.response = <attacker value>
```

**Verified end-to-end** through the real `A2AExecutor.execute()` (the exact entry point `A2AServer` wires up): park a normal interrupt via a real agent/tool, then resume with `{interruptResponse:{interruptId:'__proto__', response:'PWNED-VALUE'}}`.
```
execute() rejected? false
Object.prototype.response after attack = PWNED-VALUE
a brand-new unrelated {} .response = PWNED-VALUE
```
**Who hits this**: any legitimate client of an agent that uses the interrupt feature at all — self-contained, single-actor, no elevated privilege. Get your own tool call to raise any interrupt, then answer it with `interruptId: "__proto__"`. Because this is a single Node process serving every `contextId`/tenant, this corrupts state for *everyone* on that process, not just the attacker's own conversation.

Python's dict `in` has no such prototype-chain lookup, so this is TS-specific — not present in the ported original (#3486).

Fix: reject reserved/inherited-only keys explicitly, e.g. `Object.prototype.hasOwnProperty.call(state.interrupts, id)` instead of `id in state.interrupts` (and the equivalent at `interrupt.ts:283`'s `this.interrupts[interruptId]` lookup, or switch `InterruptState.interrupts` to a `Map`).

strands-ts/src/a2a/executor.ts
```diff
@@ -243,15 +264,27 @@ export class A2AExecutor implements AgentExecutor {
...
 
+    // Checked before the agent runs, so a refused answer leaves the interrupt parked and retryable.
+    if (isInterruptResume(prompt)) {
+      this._validateInterruptResume(agent, prompt)
+    } else if (interruptStateOf(agent)?.activated === true) {
+      throw A2AError.invalidParams('Task is awaiting an interrupt response and cannot accept a new message')
+    }
+
+    // Registered only once the request is known to be servable. Registering earlier would move a
+    // task parked in `input-required` to `working` before a refused answer threw, leaving the client
```
strandly-the-agent, line 282: 🔴 **This guarantee doesn't hold through the real server — every validation error this PR adds permanently kills the task in production.**

Verified against the installed `@a2a-js/sdk` `0.3.13` (`node_modules/@a2a-js/sdk/dist/server/index.js`): `DefaultRequestHandler.sendMessage`/`sendMessageStream` wrap *every* call to `agentExecutor.execute(...)` in `.catch((err) => { ... })`, which unconditionally synthesizes a `status: 'failed'` task/status-update using the *existing* task id and calls `eventBus.finished()` — for **any** uncaught rejection, `A2AError` or not. `_createRequestContext` then rejects all further messages to that `taskId` because `"failed"` is in its `terminalStates`.

So in production (`A2AServer`/`A2AExpressServer`, the only production entry point per `strands-ts/src/a2a/server.ts:109-115`): a client who mistypes an interrupt id, double-submits, or answers with `null` gets their task moved to a **terminal `failed` state**, not left "parked and retryable" — there is no way to answer the *correct* id afterward. The 29 executor tests all call `executor.execute()` directly, bypassing `DefaultRequestHandler` entirely, which is why this isn't caught: `executor.test.ts:583-596`'s "leaves it answerable" test only proves the in-memory `_interruptState` survives, not that the wire-level task does.

This is the same mechanism as the pre-existing `"No content blocks available"` throw (not new), but this PR is the first to make an explicit, tested, and now-documented (`agent-to-agent.mdx:464`: "a refused answer leaves the interrupt pending and the task still answerable") promise that it doesn't keep for real clients. Python's own `execute()` also just bare-re-raises `ServerError` (`executor.py:344-346`), so this may be a shared gap across both SDKs rather than TS-only — I couldn't verify Python's `a2a-sdk` server behavior in this sandbox (no `a2a` package available) — but it's confirmed for TypeScript.

Worth either fixing (catch the new validation errors here and keep the task at `input-required` instead of letting them propagate to `DefaultRequestHandler`'s fallback) or, at minimum, correcting the comment/docs and adding a test at the `A2AServer`/`DefaultRequestHandler` level (neither `server.test.ts` nor `server.test.node.ts` mentions "interrupt" at all) so this gap is visible.

strands-ts/src/a2a/executor.ts
```diff
@@ -310,13 +343,106 @@ export class A2AExecutor implements AgentExecutor {
         lastChunk: true,
       })
 
+      // An agent that stopped on an interrupt has not finished its work, so the task waits for the
+      // client to answer rather than reporting success.
+      const result = next.value
+      if (result?.stopReason === 'interrupt') {
```
strandly-the-agent, line 349: 🟡 **The pre-existing "final artifact chunk" block above (`:332-344`) isn't guarded against the interrupt case, so it also fires here — leaking the raw internal `Interrupt[]` shape into a spurious artifact.**

When no text streamed before the interrupt (the common case — an approval-gate tool interrupting on its first call, exactly like this PR's own test fixtures), `next.value.toString()` at `:340` returns `JSON.stringify(this.interrupts)` (`AgentResult.toString()`, `types/agent.ts:535-540`), and that gets published as a completed `artifact-update` *before* `_publishInputRequired` runs. Verified with a real agent/tool:
```json
{"kind":"artifact-update",...,"artifact":{"parts":[{"kind":"text","text":"[{\"id\":\"tool:tool-1:approveCampaign\",\"name\":\"approveCampaign\",\"reason\":{...},\"source\":\"tool\"}]"}]}}
```
No test catches this because the interrupt test suite (`executor.test.ts:421-641`) only filters `status-update` events, never asserting on `artifact-update`.

Python's `_stream_agent` (`executor.py:305-310`) branches structurally — `_handle_interrupt_result` is called *instead of* `_handle_agent_result` (the artifact-publishing one), so Python never reaches this code for an interrupt. Worth mirroring that shape here: return before `:332` when `stopReason === 'interrupt'`, rather than relying on the interrupt check after.

strands-ts/src/a2a/executor.ts
```diff
@@ -310,13 +343,106 @@ export class A2AExecutor implements AgentExecutor {
...
+   */
+  private _publishInputRequired(
+    interrupts: Interrupt[],
+    taskId: string,
+    contextId: string,
+    eventBus: ExecutionEventBus
+  ): void {
+    const descriptions = interrupts.map((interrupt) =>
+      interrupt.reason === undefined
+        ? `- ${interrupt.name}`
```
strandly-the-agent, line 404: 🟡 **No fallback for a non-JSON-safe `interrupt.reason` — crashes the publish instead of degrading gracefully.**

`reason` is developer-supplied (via `context.interrupt({name, reason})`), typed as `JSONValue` but not runtime-checked. A `BigInt` (a plausible choice for e.g. a large account/order id) throws here — verified with a real tool: `TypeError: Do not know how to serialize a BigInt`, which (via the artifact-chunk path noted above, or directly here if some text streamed first) crashes the entire `input-required` publish and, per the retryability finding above, permanently fails the task.

Python guards exactly this with a `_jsonable()` helper (`executor.py:78-88`) that falls back to `str(value)` on `TypeError`/`ValueError` from `json.dumps` before putting `reason` on the wire — this port has no equivalent.

strands-ts/src/a2a/executor.ts
```diff
@@ -192,44 +213,44 @@ export class A2AExecutor implements AgentExecutor {
...
   }
 
   /** Factory mode: run against this context's dedicated agent, serialized only per context. */
   private async _runWithContextAgent(
     context: RequestContext,
-    contentBlocks: ContentBlock[],
+    prompt: AgentPrompt,
     eventBus: ExecutionEventBus
   ): Promise<void> {
     const { agent, lock } = this._acquireContextAgent(context.contextId)
```
strandly-the-agent, line 238: 🟡 **LRU eviction of a context with a parked interrupt permanently orphans it — verified.**

Once a context is evicted from `_contexts` (factory mode) or `_snapshots` (shared-agent mode) — e.g. `maxContexts` exceeded by other conversations while a human is slow to answer — the parked `_interruptState` is gone. Any later resume for that `contextId`, even with the *exact correct* `interruptId`, gets a fresh agent with `activated === false` and fails with `"Received interrupt responses but no interrupt is pending"` — identical to the error for a wrong id. Verified with `maxContexts: 1`: a second, unrelated context evicts the first; the first's client (holding the right id) then fails forever, indistinguishable from a client mistake, with the tool's in-flight work silently abandoned.

This predates this PR (context isolation/eviction is from #2696), but the consequence is new and worse: previously eviction only cost some chat history; now it can permanently strand a paused tool with no diagnostic to tell the operator or client what happened. Might be worth a doc note ("don't set `maxContexts` below your expected count of concurrently-parked interrupts") or at least a more specific error/log when a resume targets an id that's plausible but the agent has no recollection of ever being activated.

site/src/content/docs/user-guide/concepts/multi-agent/agent-to-agent.mdx
```diff
@@ -443,6 +442,25 @@ answers = [
...
+const answers = pending.map((item) => ({
+  kind: 'data' as const,
+  data: { interruptResponse: { interruptId: item.interruptId, response: { approved: true } } },
+}))
+```
+
+</Tab>
+</Tabs>
+
 The server rejects an answer it cannot bind, before the agent runs, so a refused answer leaves the interrupt pending and the task still answerable. An answer is rejected when:
```
strandly-the-agent, line 464: This claim is what I verified false in production for TypeScript — see the `executor.ts:282` comment. Worth double-checking for Python too before shipping this doc change, since it removes the "Python only" caveat and now applies the same promise to both SDKs.

strandly-the-agent (changes requested):
**Verified**: branch `pr-3585` @ `54edac80b`, `npx vitest run --project unit-node src/a2a/__tests__/executor.test.ts` → 29/29 ✅. All findings below were reproduced by running real code through `A2AExecutor.execute()` (the exact production entry point) unless marked "reasoned".

**TL;DR**: solid port, but two 🔴 findings. One is a live prototype-pollution bug reachable by any client that can complete a normal interrupt round trip. The other: the PR's own headline guarantee — "a refused answer leaves the interrupt parked and retryable" — does not hold through the real `@a2a-js/sdk` server (verified against the installed `0.3.13`); every validation error this PR adds permanently fails the task in production, even though the executor-level unit tests (which call `execute()` directly) can't see that because they bypass `DefaultRequestHandler`.

<details>
<summary>Full breakdown by area</summary>

**🔴 Blockers**
- Prototype pollution via `interruptId: "__proto__"` (`executor.ts:377`, sink at pre-existing `interrupt.ts:283`) — verified end-to-end: `Object.prototype.response` gets set to attacker-controlled data, process-wide.
- "Retryable" claim is false in production (`executor.ts:274-284`, docs `agent-to-agent.mdx:464`) — verified against `@a2a-js/sdk` `0.3.13`'s `DefaultRequestHandler`.

**🟡 Should fix**
- `extractInterruptResponses` crashes with a raw `TypeError` (not `A2AError`) on `data: null` (`adapters.ts:55`) — Python's `isinstance(data, dict)` guard didn't make it into the port.
- Every interrupt round trip with no preceding text also emits a stray artifact containing `JSON.stringify(interrupts)` (`executor.ts:340` interacting with the new branch at `:348-352`) — Python's `_stream_agent` structurally skips this for the interrupt case; the TS port doesn't.
- No `_jsonable()`-equivalent guard on `interrupt.reason` (`executor.ts:404`, also hit via `:340`) — a non-JSON-safe reason (e.g. `BigInt`) crashes the whole publish instead of degrading gracefully like Python.
- LRU eviction (`executor.ts:238` `_acquireContextAgent`, `DEFAULT_MAX_CONTEXTS`) permanently orphans a parked interrupt — verified: once evicted, resuming with the *correct* id fails identically to a wrong id, forever, with no way to tell the two apart.

**Verified as non-issues** (reasoned/tested, no finding filed): `isInterruptResume`'s `prompt[0]`-only check is provably safe given `extractInterruptResponses`'s all-or-nothing return contract; `final: true` on the `input-required` status-update is correct (it only closes the current SSE stream — `input-required` isn't in the SDK's `terminalStates`); a second sequential interrupt in one task round-trips correctly; moving task registration into `_streamAgent` doesn't break anything downstream; a rejected resume's `_captureState` in shared-agent mode is a content no-op.

</details>

strands-ts/src/a2a/executor.ts
```diff
@@ -243,15 +264,27 @@ export class A2AExecutor implements AgentExecutor {
...
     context: RequestContext,
-    contentBlocks: ContentBlock[],
+    prompt: AgentPrompt,
     eventBus: ExecutionEventBus
   ): Promise<void> {
     const { taskId, contextId } = context
     const artifactId = globalThis.crypto.randomUUID()
     let isFirstChunk = true
 
+    // Checked before the agent runs, so a refused answer leaves the interrupt parked and retryable.
```
strandly-the-agent, line 274: 🔴 **The retryability promise doesn't hold in TypeScript.** This comment — and `agent-to-agent.mdx:464` ("the task still answerable") — is true for Python and false here, so a client author programs against a contract the runtime doesn't keep.

The `A2AError` propagates out of `execute()` uncaught, and `DefaultRequestHandler` wraps every `execute()` call in a `.catch` that synthesizes its own `status.state: "failed"` for the *existing* task id. Reproduced against `54edac80b` through the real handler:

```
A1 park:                     state=input-required  history=2 artifacts=1
A2 after REFUSED answer:     state=failed          history=4 artifacts=0
  status.message.parts=[{"text":"Agent execution error: No pending interrupt matches id(s): typo-id"}]
A3 corrected retry, same task: THREW code=-32600
  "Task 344ff9ef… is in a terminal state (failed) and cannot be modified."
A resumedWith=[]
```

So one typo'd or replayed id ends the task permanently, and the advertised data part is replaced by prose. Python differs structurally: `default_request_handler.py:336-339` is a bare `except Exception: … raise` with no failed-task synthesis, so the task stays `input-required` and the corrected retry works (read, not run — I couldn't execute the Python stack here).

Would it be simpler to **not throw** for an unbindable-but-well-formed answer, and instead re-publish `input-required` with `Answer refused: <message>` *plus* the same interrupts data part? That keeps the promise literally true, keeps the ids in front of the client, and `input-required` is A2A's own way of saying "your move". A genuine agent failure would still fail the task. If you'd rather keep the throw, then this comment and the docs sentence need to say a refusal is terminal in TS — but that means shipping two different meanings for one wire contract.

strands-ts/src/a2a/executor.ts
```diff
@@ -243,15 +264,27 @@ export class A2AExecutor implements AgentExecutor {
...
+    if (isInterruptResume(prompt)) {
+      this._validateInterruptResume(agent, prompt)
+    } else if (interruptStateOf(agent)?.activated === true) {
+      throw A2AError.invalidParams('Task is awaiting an interrupt response and cannot accept a new message')
+    }
+
+    // Registered only once the request is known to be servable. Registering earlier would move a
+    // task parked in `input-required` to `working` before a refused answer threw, leaving the client
+    // unable to see that the task is still waiting on them.
+    eventBus.publish({ kind: 'task', id: taskId, contextId, status: { state: 'working' } })
```
strandly-the-agent, line 284: 🔴 **Answering an interrupt deletes the task's history — on the documented happy path.** This publish is unconditional, so it also fires on a *resume*, and `ResultManager.processEvent` does `this.currentTask = { ...taskEvent }` and saves. Reproduced through the real handler + `InMemoryTaskStore`:

```
B1 park  (store): state=input-required  history=2  artifacts=1
  history[0] user : create the spring-launch campaign
  history[1] agent: "Agent requires input…" + {"interrupts":[{"interruptId":"tool:tool-1:approveCampaign",…}]}
B2 after CORRECT answer (store): state=completed  history=1  artifacts=1
  history[0] user : {"interruptResponse":{"interruptId":"tool:tool-1:approveCampaign","response":"approved"}}
B resumedWith=["approved"]
```

The original request, the advertisement, and the park artifact are gone. For an approval workflow the audit trail is much of the product: after a human approves, `tasks/get` shows a task whose entire history is the answer.

The line pre-dates this PR, but it was **unreachable** before it: pre-PR the executor only ever published `state: 'completed'` (`526692f53:executor.ts:313`), every task ended terminal, and the handler rejects any further message for a terminal task — so there was never an existing task to overwrite. This PR creates the first non-terminal resting state *and* tells clients to answer on the same `taskId` (`mdx:418`), which makes the overwrite the feature's main path.

Python guards exactly this (`executor.py:336-339` registers only when `context.current_task` is falsy). Would matching it work here?

```suggestion
    // Registered only once the request is known to be servable, and only for a brand-new task:
    // republishing on a resume would replace the stored task and drop its history and artifacts.
    if (context.task === undefined) {
      eventBus.publish({ kind: 'task', id: taskId, contextId, status: { state: 'working' } })
    }
```

Worth an inline note that this relies on `ResultManager` lazily loading the task for later status/artifact events, since `@a2a-js/sdk` is `^`-ranged — and a test through `DefaultRequestHandler` rather than the mock bus, which is what would have caught this.

strands-ts/src/a2a/executor.ts
```diff
@@ -310,13 +343,106 @@ export class A2AExecutor implements AgentExecutor {
         lastChunk: true,
```
strandly-the-agent, line 343: 🟡 **The park also publishes the interrupts a second time, with a different key name and an internal field.** This closing artifact fires before the interrupt branch below, and for a parked result `next.value.toString()` is `AgentResult.toString()`, which is `JSON.stringify(this.interrupts)` (`types/agent.ts:535-538`). From the same probe run:

```
artifact[0]: [{"kind":"text","text":"[{\"id\":\"tool:tool-1:approveCampaign\",\"name\":\"approveCampaign\",\"reason\":\"needs sign-off\",\"source\":\"tool\"}]"}]
```

Three things a client shouldn't have to reconcile: the same concept is keyed `id` here and `interruptId` in the data part; `source` is internal taxonomy now on the wire; and `lastChunk: true` closes the artifact on a task that hasn't finished. It also has a downstream effect — `A2AAgent._isCompleteEvent` accepts any `artifact-update` with `lastChunk === true`, so this dump becomes the "answer" a nested Strands agent sees.

Python avoids all of it by branching before any artifact is added (`executor.py:307-310`). Would moving this publish into the non-interrupt branch (just below, next to the `completed` status) be the closest thing to parity?

strands-ts/src/a2a/executor.ts
```diff
@@ -310,13 +343,106 @@ export class A2AExecutor implements AgentExecutor {
...
+    // The text part stays first so clients that only read prose are unaffected.
+    const parts: Part[] = [{ kind: 'text', text }]
+    if (interrupts.length > 0) {
+      parts.push({
+        kind: 'data',
+        data: {
+          [INTERRUPTS_KEY]: interrupts.map((interrupt) => ({
+            interruptId: interrupt.id,
+            name: interrupt.name,
+            ...(interrupt.reason === undefined ? {} : { reason: interrupt.reason }),
```
strandly-the-agent, line 421: 🟡 **`reason` presence diverges from Python, on a contract the PR presents as byte-identical.** Python always emits the key (`executor.py:451`: `{"interruptId": …, "name": …, "reason": _jsonable(interrupt.reason)}` → `"reason": null` when absent); this omits it. So the same interrupt serialises differently depending on which SDK serves it, and a client that generates a schema from observed Python payloads (or writes `entry["reason"]`) breaks when repointed at a TS server. `executor.test.ts:461-469` now locks the TS shape in.

The prose part diverges too (`:405`): TS `JSON.stringify`s every reason → `- approveCampaign: "needs sign-off"`, while Python f-strings a truthy one → `- approveCampaign: needs sign-off`, and Python-reprs dicts (`{'campaign': 'x'}`) where TS emits JSON. That's the one part the design says exists for prose-only clients.

I'd normally say "Python shipped first, match it" — but `python/v1.50.{0,1,2}` don't contain `aae087d6d`, so **neither side is released** and this is the last cheap moment. My suggestion is TS follows Python on the wire (a key that's always present is friendlier to schema-driven clients than an optional one, and Python's dataclass can't express omission without custom serialization), Python follows TS on the prose (`json.dumps` for non-strings), and the docs state the contract:

```suggestion
            reason: interrupt.reason ?? null,
```

Happy to be wrong on the direction — but the two SDKs should agree on key presence before a client depends on either.

strands-ts/src/a2a/executor.ts
```diff
@@ -310,13 +343,106 @@ export class A2AExecutor implements AgentExecutor {
...
+    )
+    const text =
+      descriptions.length > 0
+        ? `Agent requires input:\n${descriptions.join('\n')}`
+        : 'Agent requires additional input to continue'
+
+    // The text part stays first so clients that only read prose are unaffected.
+    const parts: Part[] = [{ kind: 'text', text }]
+    if (interrupts.length > 0) {
+      parts.push({
```
strandly-the-agent, line 415: 🟡 **The control payload lives in application-payload space, and this module converts that space into model-facing text.** `adapters.ts:118` maps any inbound data part to `new TextBlock('[Structured Data]\n' + JSON.stringify(part.data, null, 2))`, so a client that replays the parked `status.message.parts` into any Strands agent injects `[Structured Data]\n{"interrupts":[…]}` straight into a prompt. A2A's guidance points the other way: "Extensions should place custom attributes in the metadata map present on core data structures."

I'm **not** proposing you move the payload — I looked at `Message.metadata[uri]` and it loses more than it gains (more hops for the client, can't drop the data part without a break, and it depends on extension activation via a header whose name differs between the pinned SDK's `X-A2A-Extensions` and the live spec's `A2A-Extensions`). The data part is the more ergonomic choice and the text-part-first ordering is thoughtful.

Two cheaper things would get most of the benefit:
1. Declare the contract on the agent card so it's discoverable and versioned — today `server.ts:104-106` advertises `capabilities: { streaming: true }` and nothing else, so a client can't tell this agent may park on `input-required` with interrupt ids. `AgentCapabilities.extensions` exists in the pinned 0.3.13: `{ uri: 'https://strandsagents.com/a2a/ext/interrupts/v1', description: …, required: false }` — declaration only, no header plumbing.
2. Skip an interrupts-shaped data part in `partsToContentBlocks` so the control payload can never become prompt text (and optionally tag the part via `Part.metadata` with that URI so a strict consumer has something to key on).

Both are additive and neither changes the shape clients read today.

strands-ts/src/a2a/executor.ts
```diff
@@ -43,6 +47,23 @@ export interface A2AExecutorOptions {
...
+ */
+type AgentPrompt = ContentBlock[] | InterruptResponseContent[]
+
+/** Whether a prompt carries interrupt responses rather than conversational content. */
+function isInterruptResume(prompt: AgentPrompt): prompt is InterruptResponseContent[] {
+  return prompt.length > 0 && isInterruptResponseContent(prompt[0])
+}
+
+/** Reads the agent's interrupt state, a field not declared on {@link InvokableAgent}. */
+function interruptStateOf(agent: InvokableAgent): InterruptState | undefined {
```
strandly-the-agent, line 63: 🟡 **This reads `_interruptState` off the wider `InvokableAgent`, and the factory path never narrows — so a third-party agent silently loses the guard.** `AgentFactory = (contextId: string) => InvokableAgent` (`executor.ts:25`), and `asSnapshotAgent` — the loud "this isn't a full Agent" check — is only called in the deprecated shared-agent branch (`executor.ts:156`); the factory path never validates. So for any `InvokableAgent` that isn't an `Agent` (`A2AAgent implements InvokableAgent`, `a2a-agent.ts:54`, makes a gateway topology type-legal), this returns `undefined`, `executor.ts:277` reads that as "no interrupt pending" and skips the parked-task guard, and a resume attempt reports `"Received interrupt responses but no interrupt is pending"` when the truth is "this agent type can't do interrupts".

I'd call that silent misbehaviour rather than graceful degradation: nothing tells the operator their gateway agent has a hole where the guard should be, and the error message actively misleads. The cast itself is fine and has precedent (`interrupt.ts:404`, `snapshot.ts:167`) — the issue is that the precedent takes the *narrower* `LocalAgent`.

The cheap fix is to reuse the pattern already in this file: narrow the factory-produced agent through `asSnapshotAgent` so a non-`Agent` fails loudly at construction/first use. A public `pendingInterrupts` accessor on `InvokableAgent` would be the cleaner long-term answer, but that's a bigger API decision than this PR should make — a capability interface plus a type guard (the way `LocalAgent` is sealed by `localAgentSymbol`) seems more in keeping with the existing design. What's the intended story for a factory that returns something other than a full `Agent`?

strands-ts/src/a2a/executor.ts
```diff
@@ -310,13 +343,106 @@ export class A2AExecutor implements AgentExecutor {
...
+      throw A2AError.invalidParams(`No pending interrupt matches id(s): ${unknownIds.join(', ')}`)
+    }
+  }
+
+  /**
+   * Moves the task to `input-required` and tells the client which interrupts are waiting.
+   *
+   * The interrupts are published twice: a text part describing what is needed, and a data part
+   * carrying each `interruptId`. Only the id lets a client address its answer back to the interrupt
+   * that raised it, and ids are generated server-side so they cannot be inferred from the prose.
```
strandly-the-agent, line 389: 🟡 **"cannot be inferred from the prose" isn't accurate, and the id format is now a de-facto contract.** The advertised id is `` `tool:${toolUse.toolUseId}:${params.name}` `` (`tools/executors/executor.ts:257`) — the name is printed in the text part one part earlier (`- approveCampaign: …`) and `toolUseId` is model-supplied, so a client *can* often reconstruct it. Worth deleting the claim, since nothing depends on it (binding is enforced by `_validateInterruptResume`, which is the real protection).

The more consequential half: this PR promotes that string from an internal map key to the primary key of a public wire contract, and the two SDKs generate it differently at every raise site — Python `v1:tool_call:{toolUseId}:{uuid5(name)}` (`types/tools.py:162`), TS `tool:{toolUseId}:{name}`. Python reserved format evolution with the `v1:` prefix; TS has no marker, so it can never migrate.

Two asks, neither large: (1) document `interruptId` as an **opaque token the client echoes verbatim and never parses or constructs** — that's what actually keeps you free to change it; (2) decide whether the TS scheme should pick up a `v1:` prefix for parity. Unlike the `reason` question, TS *is* released here (`typescript/v1.11.2` contains this scheme), so realigning would only affect interrupts parked across an upgrade — cheap now, impossible later.

strands-ts/src/a2a/executor.ts
```diff
@@ -310,13 +343,106 @@ export class A2AExecutor implements AgentExecutor {
...
+   *   the agent is not waiting on
+   */
+  private _validateInterruptResume(agent: InvokableAgent, responses: InterruptResponseContent[]): void {
+    const state = interruptStateOf(agent)
+    if (state?.activated !== true) {
+      throw A2AError.invalidParams('Received interrupt responses but no interrupt is pending')
+    }
+
+    const unknownIds = responses
+      .map((content) => content.interruptResponse.interruptId)
```
strandly-the-agent, line 376: 🟡 **Should this tell the client which ids *are* pending?** Today it names only the ids that failed to bind, so a client that lost the status message can't recover without a `tasks/get` — and it's the most likely error in the taxonomy (stale retry, wrong task, truncated id).

One correction to the obvious fix: `A2AError.invalidParams` takes a structured `data` argument, but it's not worth using here — my probe shows `@a2a-js` discards both the JSON-RPC code and `data` and hands the client prose only (`"Agent execution error: <message>"`, with `data: undefined` visible only in the server log). In TypeScript the message string *is* the entire error surface, so the pending ids have to go in the text to reach anyone:

```suggestion
      .map((content) => content.interruptResponse.interruptId)
```

(suggestion left unchanged deliberately — the change belongs on the `throw` two lines below, something like `No pending interrupt matches id(s): ${unknownIds.join(', ')}; pending: ${Object.keys(state.interrupts).sort().join(', ')}`.)

On the code choice: `invalidParams` is right for all seven rules as far as I can tell. I checked the full `A2AError` factory list and A2A 0.3.x has no task-state-conflict code, so even `"Task is awaiting an interrupt response"` (`executor.ts:278`) — which really is a state conflict, not a bad param — has no better home.

site/src/content/docs/user-guide/concepts/multi-agent/agent-to-agent.mdx
```diff
@@ -381,15 +381,11 @@ In single-agent mode the server reuses one agent across every context, swapping
...
-This feature is currently available in the Python SDK only.
-:::
-
-When a tool or hook on the served agent raises an [interrupt](../interrupts.md), the task moves to the A2A `input_required` state and waits. The client answers it, and the task resumes the paused tool exactly where it stopped.
+When a tool or hook on the served agent raises an [interrupt](../interrupts.md), the task moves to the A2A `input-required` state and waits. The client answers it, and the task resumes the paused tool exactly where it stopped.
 
-Answering an interrupt is the one flow on this page that needs the raw [`a2a-sdk`](https://github.com/a2aproject/a2a-python) client rather than `A2AAgent`. `A2AAgent` speaks in text, so it raises `ValueError` if you pass it interrupt responses, and it drops the `DataPart` carrying the interrupt ids when it reads the reply. The examples below build A2A messages directly.
+Answering an interrupt is the one flow on this page that needs the raw A2A client rather than `A2AAgent`. `A2AAgent` speaks in text, so it rejects interrupt responses on the way out and drops the data part carrying the interrupt ids on the way back. The examples below build A2A messages directly.
```
strandly-the-agent, line 386: 🟡 **This sentence is true for Python and false for TypeScript, and the `:::note[Python only]` banner that used to scope it is gone.** `A2AAgent` in TS doesn't reject interrupt responses: `rg -n throw src/a2a/a2a-agent.ts` returns nothing. `_extractTextFromArgs` (`a2a-agent.ts:215-225`) strips non-text blocks with a `logger.warn` and returns `''`, so an `InterruptResponseContent` silently becomes an empty text turn — i.e. the gated action gets restarted rather than answered. Python does throw: `ValueError("InterruptResponseContent is not supported for A2AAgent")` (`multiagent/a2a/_converters.py:51-52`).

The same paragraph's other half is also TS-specific: Python maps `TaskState.input_required → stop_reason "interrupt"` (`_converters.py:17-24`), whereas TS hardcodes `stopReason: 'endTurn'` (`a2a-agent.ts:260`) and `_isCompleteEvent` only treats `'completed'` as complete (`:238`) — so a parked remote task is reported to the calling agent as a normal finished turn. That's the same silent-success failure this PR fixes on the server side, still live on the client side, and it's the shape a parent agent in a multi-agent setup will see.

Smallest correct fix is to re-scope the sentence per language. The better one, if you're willing: make TS `A2AAgent` throw on interrupt-response input and map `input-required` → `stopReason: 'interrupt'` (`StopReason` already has the literal). Either way I'd also add here that `interruptId` is opaque, that `reason` may be `null`, and that answering only *some* pending interrupts is supported — I verified partial answers work and re-park with the remainder, and nobody would guess that from the current text.

site/src/content/docs/user-guide/concepts/multi-agent/agent-to-agent.mdx
strandly-the-agent, line 464: 🟡 Same issue as `executor.ts:274`, but this one ships to customers: "a refused answer leaves the interrupt pending and the task still answerable" is verified false in TypeScript — the task goes terminal `failed` and the corrected retry is rejected `-32600`. Either fix the behaviour (preferred — see my comment on `executor.ts:274`) or scope this paragraph per language, because right now one shared sentence documents two different behaviours.

Worth adding once it's accurate: the advertised ids also survive in the task's `history`, so a client that dropped the status message can recover them from `tasks/get` rather than being stuck.

strandly-the-agent (changes requested):
**API / DevX design review** (contract only — correctness and tests were out of scope for this pass). Verdict: **changes requested**. The mechanism and the shape are mostly right and the port is careful, but the contract as *documented* is not the contract the runtime honours, and one wire divergence from Python should be settled now while neither SDK has shipped it.

**What I verified** — reviewed `54edac80b` vs `526692f53`; `npx vitest run --project unit-node src/a2a/__tests__/executor.test.ts` → **29/29 green**; tree clean after removing my probes.

- 🔴 Reproduced through the **real** `DefaultRequestHandler` + `InMemoryTaskStore` (the PR's tests use a mock event bus, so neither is visible to them): a *refused* answer ends the task `failed` and the corrected retry is rejected `-32600`; a *successful* answer on the same `taskId` drops the task history 2 → 1.
- ✅ `python/v1.50.{0,1,2}` do **not** contain `aae087d6d` (#3486) — the Python contract is merged but **unreleased**, so "Python already shipped it" doesn't constrain the wire shape yet.
- ✅ A2A's extension mechanism is fully present in the pinned `@a2a-js/sdk` 0.3.13: card `capabilities.extensions`, `Message.metadata` ("Optional metadata for extensions"), `PartBase.metadata`, `ServerCallContext.addActivatedExtension`.

**Before merge:** the two 🔴s below, plus a decision on `reason` presence (Python emits `"reason": null`, TS omits the key) and on whether `interruptId`'s format is a contract. Everything else is a 🟡 proposal or a question.

**Label gate:** this changes a public surface (a wire contract, permanent across two SDKs) and carries neither `needs-api-review` nor `completed-api-review` — it needs `needs-api-review`. I'd also put two decisions to the API meeting rather than a solo approve: (a) data-part key vs a declared A2A extension, (b) cross-SDK `interruptId` format.

<details>
<summary><b>Answers to the seven design questions</b></summary>

**1. Is the outbound shape right?** Mostly yes, and it's forward-compatible in the ways that matter: unknown keys are ignored on both sides (`adapters.ts:65` destructures and discards extras), so `schema`, `optional`, `expiresAt` and even a sibling `interruptsVersion` can all be added additively later. I would *not* add an envelope or `version` now — that's speculative generality. `interruptId` over `id` is the right call inside an `interrupts` array, because it's the name the client must echo back and it matches the inbound key. The real gap isn't a missing field, it's that **`interruptId`'s format is undefined-but-observable** — see `executor.ts:389`.

**2. Is a bespoke data-part key the right vehicle?** This is the highest-value question and my answer is *keep the data part, but stop treating it as ordinary payload*. Against: A2A's own guidance is "Extensions should place custom attributes in the metadata map present on core data structures", and this repo proves the point — `adapters.ts:118` turns any inbound data part into `[Structured Data]\n<json>` **model-facing text**, so a client that replays the parked message injects the control payload into a prompt. For: `parts.find(p => p.kind === 'data').data.interrupts` is fewer hops than `status.message.metadata[uri].interrupts`; it needs no header negotiation (and the header name is itself unstable — the SDK exports `X-A2A-Extensions` while the live spec says `A2A-Extensions`); the text-part-first ordering keeps prose-only clients working; and the inbound shape is `InterruptResponseContent.toJSON()` verbatim, so type and wire cannot drift. Moving the payload to `metadata` costs ergonomics and can't remove the data part without a break. What I'd do instead: **declare an extension URI in `capabilities.extensions`** (`required: false`, no header plumbing) so the feature is discoverable and versioned at all, tag the part via `Part.metadata`, and skip an interrupts-tagged data part in `partsToContentBlocks` so it can never become prompt text. Today `server.ts:104-106` advertises `{ streaming: true }` and nothing else — a client has no way to discover this contract exists.

**3. Is the asymmetry sane?** Yes — I tried to argue for a batch inbound `interruptResponses: [...]` and failed. The singular shape is `InterruptResponseContent.toJSON()`, so it costs zero mapping; duplicate detection is per-message either way; and **partial answers already work** (`InterruptState.resume` only writes the ids named, `getUnansweredInterrupts` re-parks the rest, and ids are stable across the re-park because `getOrCreateInterrupt` keys by id). A batch shape would add a second inbound form for no semantic gain. Worth *documenting* that partial answers work — it's a real capability nobody would guess.

**4. Are the errors actionable?** The taxonomy is good and `invalidParams` is the least-wrong code available for all seven — I checked the full `A2AError` factory list and A2A 0.3.x has no task-state-conflict code, so there's no better option for "task is awaiting an interrupt response". Two corrections to the premise, though: `A2AError`'s structured `data` argument would be **pointless here** — my probe shows `@a2a-js` discards both the code and `data` and hands the client prose only (`"Agent execution error: <message>"`), so the message text is the entire error surface in TypeScript. Which makes the one improvement worth making a text change: the unknown-id error should name the pending ids, not just the offending ones.

**5. Cross-SDK parity.** Three deliberate-looking divergences: (a) TS rejects `undefined` as well as `null` — **right**, and Python should follow, since a missing key and a null answer are the same client mistake; (b) TS has no `_jsonable()` fallback — **right and needs no fix**, because `Interrupt.reason` is typed `JSONValue` in TS whereas Python's is `Any`; type-level enforcement is the better answer; (c) the moved task registration — the *intent* is right (validate before you register) but it's defeated one layer up, see `executor.ts:274`. And one that reads accidental rather than deliberate: `reason` key presence (`executor.ts:421`).

**6. Documentation.** `agent-to-agent.mdx:382-475` is genuinely good for a wire contract — both JSON shapes, all the rejection rules as a bullet list, the id-echo requirement, and now a TS tab. Three gaps: it repeats the retryability promise that doesn't hold in TS (`:464`), it asserts `A2AAgent` behaviour that doesn't exist in TS (`:386`), and it shows `reason` in the example without saying whether it's always present — which is exactly the field the two SDKs disagree on. Add a line that `interruptId` is an opaque token to echo verbatim, and that partial answers are allowed.

**7. Label gate.** See above: `needs-api-review` at minimum; the mechanism and id-format decisions deserve the meeting.

</details>

<details>
<summary><b>Alternatives I considered, and why I'm not proposing them</b></summary>

**A — relocate the payload to `Message.metadata[uri]`** (the spec-shaped "profile extension"). Verified type-legal and accepted by the handler. Rejected: more hops for the client, can't drop the data part without a break (so near-term it's *more* surface, not less), and it bets on a header whose name differs between the pinned SDK and the live spec. **Its card-declaration half is worth keeping** and is nearly free.

**B — keep the part, fix the contract** (what I'm proposing): guard the task registration, keep the interrupt payload out of the artifact stream, re-park instead of throwing on an unbindable answer, always emit `reason`, declare the extension on the card, and don't let the control part become model text. This preserves every property the current design optimises for — one-hop read, zero-mapping inbound, prose-client compatibility, strict-now-relax-later — and changes only what's irreversible after shipping.

**C — "document the divergence"**: keep the behaviour and tell TS users a refusal is fatal. Rejected: it enshrines two different meanings for one wire contract, and does nothing about the history loss.

</details>

<details>
<summary><b>Appendix — non-blocking (4)</b></summary>

- ⚪ **Exports to nowhere.** `INTERRUPT_RESPONSE_KEY`, `INTERRUPTS_KEY` and `extractInterruptResponses` (`adapters.ts:22,28,49`) are exported but unreachable: the barrel omits them *and* `package.json` `exports` has no `./a2a/adapters` entry and no wildcard, so no consumer outside the package can import them. So the answer to "should they be exported?" is: harmless as-is, but the interesting question is the opposite one — a TS *client* author would benefit from `INTERRUPTS_KEY`/`INTERRUPT_RESPONSE_KEY` being reachable (that's the one thing that would make the contract typed rather than stringly-typed on the client side). The missing `@internal` tag matches the rest of `src/a2a/` (zero occurrences in the directory), so it's not a regression from this PR.
- ⚪ **`adapters.test.ts` gained nothing**, so the duplicate-id and mixed-content rules have no unit test; all new coverage goes through `executor.execute()` with a mock event bus. That mock is also why the two 🔴s are invisible — one test through `DefaultRequestHandler` would have caught both.
- ⚪ **`cancelTask` is unreachable for a parked task** (`executor.ts:452-454`): `@a2a-js` only consults the executor while an event bus is live, so cancel on a parked task stamps `canceled` in the store without the agent's interrupt state ever being deactivated. Worth a follow-up issue as the release valve for an abandoned parked context — I checked and Python has the same context-level gap, so it's a cross-SDK design item, not a TS blocker.
- ⚪ Nice touches worth keeping, for the record: text-part-first ordering, the narrow recognition rule so ordinary data parts still take the generic path, rejecting a null response with the reasoning in a comment, and the round-trip test that asserts the advertised id equals the id the agent actually parked.

</details>

---
<sub>Automated review by `strandly-the-agent`. Pipeline: triage → context build → API/DevX bar-raiser (deep pass) → aggregation. `spawn` was unavailable in this environment, so the specialist pass ran in-context and I re-verified its load-bearing claims myself with the probe above — two of its findings were wrong and dropped. Solid work for a human to approve, not a gate.</sub>

strandly-the-agent:
## Consolidated review — supersedes my two earlier reviews

**Housekeeping first: the two `CHANGES_REQUESTED` reviews above are both mine, and that duplication is my bug, not a signal.** Two stages of my pipeline published instead of handing findings back for aggregation, so several findings are filed 2–4× at slightly different lines. This comment is the single de-duplicated ledger — **read this one**, and treat it as authoritative where it corrects them (it does, in three places).

**Verdict: changes requested — but this is good, careful work, and none of it is a redesign.** The port is faithful, the validation taxonomy is genuinely well thought through, and 3 of the 4 blockers live in the same 15 lines. A human maintainer should make the call; I'm flagging, not gating.

Verified on `pr-3585` @ `54edac80b` (merge-base `526692f53`) — I reproduced each 🔴 myself through real components, not just from the passes:

| | Check | Evidence |
|---|---|---|
| ✅ | `npx vitest run --project unit-node src/a2a/__tests__/executor.test.ts` | 29/29 green |
| 🔴 | `interruptId: "__proto__"` → `Object.prototype` pollution → **every later interrupt in the process silently auto-approves** | reproduced end-to-end; a *fresh* `Agent`+executor+context then `completed` without pausing and the guarded tool ran |
| 🔴 | "a refused answer leaves the interrupt pending and the task still answerable" is **false** | reproduced through the real `DefaultRequestHandler`: bad resume → task terminal `failed` (no JSON-RPC error!), correct retry then refused |
| 🔴 | a **successful** resume drops task history `2 → 1` | reproduced through real `DefaultRequestHandler` + `InMemoryTaskStore` |
| 🟡 | `{kind:'data', data:null}` → raw `TypeError`, not `A2AError` | reproduced for `null`/`undefined`/`5`/`"str"`/`true` |

**Do this first:** `executor.ts:270-284` is where three of the four live — (1) `Object.hasOwn(state.interrupts, id)` instead of `id in state.interrupts`, (2) catch the validation errors here and re-publish `input-required` rather than letting them reach `DefaultRequestHandler`'s terminal-fail catch, (3) skip the `kind:'task'` republish when this is a resume.

**Label:** this needs `needs-api-review`. The wire contract is effectively permanent once two SDKs speak it — and the good news is **Python is still unreleased** (`#3486` = `aae087d6d`, 07-29, is not in `python/v1.50.2`, 07-27), so the `reason`-key and id-format questions below are still cheap to settle on both sides. I'd take those two to the API meeting rather than approve solo.

<details>
<summary><b>🔴 Blockers — de-duplicated (4 findings, filed across 6 of the 17 comments)</b></summary>

**1. `__proto__` prototype pollution → HITL bypass.** `executor.ts:377`, sink at pre-existing `interrupt.ts:283`.
`'__proto__' in {}` is `true`, so `!(id in state.interrupts)` never fires; `InterruptState.resume` then resolves `this.interrupts['__proto__']` to `Object.prototype` (truthy) and assigns `.response` on it. Every later "is this answered?" test is `response === undefined` (`interrupt.ts:232-246`, `agent.ts:1432`), so **every future interrupt in the process reads as already answered**. My repro: a brand-new `Agent` + `A2AExecutor` on an unrelated `contextId` never paused and the approval-guarded tool executed unattended.
*Who hits it:* anyone who can legitimately use the feature — answer your own interrupt with `interruptId:"__proto__"`. No privilege, no server internals, one crafted data part. Python is immune (`not in` on a dict has no prototype chain), so this is a port artifact. `constructor`/`toString` pass the same guard with milder sinks — worth fixing at `interrupt.ts:282` and `:305` too, since those are the real sinks.

**2. "Parked and retryable" doesn't hold in TypeScript.** `executor.ts:281-283` + `agent-to-agent.mdx` — *filed 4× (`executor.ts:274`, `:282`, and the mdx line twice); it is one finding.*
`DefaultRequestHandler.sendMessage`/`sendMessageStream` wrap every `execute()` rejection in a catch that synthesizes `status.state:'failed'` **using the parked task's own id**, and `failed` is terminal, so `_createRequestContext` refuses everything afterwards. Worth noting the client gets a `200` with a `failed` task and **no** JSON-RPC error, so this is invisible to them until the retry bounces.
*Who hits it:* any client that mistypes an id, double-submits an approval, or retries. Ordinary client error on a path the docs explicitly promise is recoverable. Your executor-level tests can't see it — they call `execute()` directly.

**3. A successful resume drops the task's history.** `executor.ts:284`.
The `kind:'task'` republish is unconditional, so it fires on a resume too, and `ResultManager` overwrites the stored task: `history len 2 → 1` (the original request *and* the interrupt advertisement both gone), confirmed against `InMemoryTaskStore`. **Newly reachable** — before this PR no task could be non-terminal between requests, so there was never an existing task to clobber.
*Who hits it:* every successful interrupt resume, i.e. the documented happy path. Silent, no error — and losing the approval request/response pair is exactly the audit trail this feature exists to create.

**4. Non-object `part.data` throws a raw `TypeError`.** `adapters.ts:55`.
Python's guard is `if not isinstance(data, dict) or KEY not in data` (`executor.py:692`); the port kept the `not in` half and dropped the `isinstance` half. Blast radius is wider than interrupts: `extractInterruptResponses` now runs on **every** inbound message, so any client sending an ordinary malformed data part gets `"Agent execution error: Cannot use 'in' operator…"` instead of your intended clean `A2AError.invalidParams` — and via (2), a dead task. On merge-base these became a plain text block. `typeof part.data === 'object' && part.data !== null` fixes it.

</details>

<details>
<summary><b>🟡 Should fix</b></summary>

- **Stray artifact leaks the raw `Interrupt[]`, including the internal `source` field.** *(filed twice: `executor.ts:343` and `:349`.)* The closing artifact publishes `next.value.toString()` when `isFirstChunk`, and `AgentResult.toString()` is `JSON.stringify(this.interrupts)` (`types/agent.ts:535`) — only non-empty when `interrupts.length > 0`, so this is new to the interrupt path rather than a pre-existing leak. Observed: `parts:[{"kind":"text","text":"[{\"id\":\"tool:tool-1:approveWire\",…,\"source\":…"}]`. Fires whenever nothing streamed before the interrupt — the common case, and the shape of your own fixtures. Python's `_stream_agent` structurally skips the artifact branch here; returning before it would match.
- **No `_jsonable()` fallback for `interrupt.reason`** (`executor.ts:404`). **Correction to my own review #2**, which called this "right and needs no fix" because `reason` is typed `JSONValue`: that's compile-time only. `reason` is developer-supplied and routinely arrives via `any`/`unknown`-typed tool args or a plain-JS caller, and my review #1 verified a real `TypeError: Do not know how to serialize a BigInt`. Review #1 is right — treat it as a real 🟡. The failure mode is nasty: it throws *inside* `_streamAgent`'s `try`, so the task fails **and** the agent stays parked with its id never advertised. Python added `_jsonable()` (`executor.py:80-90`) for precisely this, with a regression test.
- **`reason` key presence diverges from Python** (`executor.ts:415-419`). Python always emits the key (`executor.py:451`, `null` when absent); TS omits it. That's a real wire divergence against root `AGENTS.md:53` ("wire field names keep their wire format in both SDKs"), and `executor.test.ts:461-469` now locks the TS shape in. Since Python hasn't shipped, either side can move — worth deciding rather than inheriting.
- **Docs sentence about `A2AAgent` is now false for TypeScript** (`agent-to-agent.mdx:386`). **Partial correction:** the underlying code gap is *pre-existing* (`a2a-agent.ts` isn't in this diff) so don't fix code here — but the diff removed the `:::note[Python only]` scoping and generalised "raises `ValueError`" into "rejects interrupt responses", which TS doesn't do. Verified: `_extractTextFromArgs` (`a2a-agent.ts:198-225`) silently strips non-text blocks with a `logger.warn` and sends an **empty** text message. "Rejects" is worse than the truth here — someone debugging "my resume did nothing" will never suspect it. Worth a TS-specific caveat in this PR.
- **Nit:** `No pending interrupt matches id(s)` (`executor.ts:376`) could also list the ids that *are* pending — a client that lost the status message otherwise can't recover.

</details>

<details>
<summary><b>Corrections to my earlier reviews — two findings I'm withdrawing / downgrading</b></summary>

- **LRU eviction orphaning a parked interrupt** (`executor.ts:238`): the repro is real (`maxContexts: 1` → the correct id fails as "no interrupt is pending"), but on reflection **the eviction mechanism is pre-existing** (#2696) and fixing it isn't this PR's job. What's new is only that parked-across-turns state is now eviction-bait. Downgrading from 🟡-code-fix to a suggested doc caveat about `maxContexts` vs. expected concurrently-parked interrupts. Not a merge blocker.
- **`_interruptState` read via a cast not narrowing for third-party `InvokableAgent`s** (`executor.ts:63`): **withdrawing as speculative.** The only implementers here are `Agent` (supports interrupts) and `A2AAgent` (can't do interrupts for an unrelated pre-existing reason). No concrete caller hits the silent-degradation path I described.
- Also correcting a factual slip: I implied the `"No content blocks available"` path had changed. It hasn't — that check precedes registration in both the old and new code. The registration move does change timing for the ordinary non-interrupt path (now after lock + state restore, where Python still registers unconditionally at the top of `execute()`); benign as far as I can tell, but worth confirming it's intentional since #3486 didn't ask for it.

</details>

<details>
<summary><b>Tests — the meta-finding, and a concrete pre-merge checklist</b></summary>

The suite is green *and structurally cannot see four of the five defects above*, which is the most useful thing I found. Every test drives `executor.execute()` with a hand-rolled `createMockEventBus()` (`executor.test.ts:17-30`) that has no concept of a `Task`, a task store, or a terminal state — and every assertion filters `events` down to `kind === 'status-update'`, so the one event that leaks (`artifact-update`) is never looked at.

`executor.test.ts:583-596` ("refused answer leaves it answerable") is tautological as written: both `execute()` calls share the same executor/factory/`contextId`, so `_interruptState` survives simply because they're two calls in one process. It proves `_validateInterruptResume` doesn't mutate — true by inspection (it's a read + `.filter()`). It cannot fail from finding (2) even in principle. Python has the same weak shape at `test_executor.py:2467`.

Minimum bar I'd ask for:
1. `adapters.test.ts` (untouched by this PR, though `extractInterruptResponses` holds all the new branch logic) — direct unit tests for the **4 of 8 rejection rules with zero coverage anywhere**: duplicate id, missing/blank `interruptId`, non-object payload, mixed content. Plus `interruptId: '__proto__'` and `part.data = null`. Note `executor.test.ts:636-660` looks like the mixed-content case but isn't — with only a non-interrupt part, `responses.length === 0` short-circuits at `adapters.ts:83` before the check at `:88`.
2. One test through `new DefaultRequestHandler(card, new InMemoryTaskStore(), executor)` — no Express, no HTTP, works fine in a plain vitest unit test — asserting (a) a refused resume leaves the task non-terminal and the correct retry still succeeds, (b) a successful resume preserves `Task.history`. There's no precedent for this in the repo's unit suite today (`server.test.ts`/`server.test.node.ts` only cover constructor/middleware wiring), which is likely why the gap went unnoticed.
3. Assert on `artifact-update` for the park path, to pin the `source` leak.
4. Retitle or drop `:583-596`.

Five Python tests from #3486 have no TS counterpart: `test_execute_duplicate_interrupt_responses_rejected` (`test_executor.py:2486`), `..._malformed_interrupt_response_rejected` (:2517), `..._falsy_interrupt_response_is_valid` (:2408), `..._mixed_with_other_parts_rejected` (:2535), `test_interrupt_reason_that_is_not_json_falls_back_to_text` (:2698). That last one is the regression test for the `_jsonable` gap above.

Minor: `executor.test.ts:478-479` uses `toMatchObject` twice where `docs/TESTING.md:271` asks for `toEqual` on objects/arrays.

</details>

<details>
<summary><b>Design questions — genuinely open, none blocking on their own</b></summary>

- **Should the control payload stay in a `data` part?** I talked myself out of proposing `Message.metadata` — it's more hops and you can't drop the part without a break. But two cheap wins: declare the contract as an extension in `capabilities.extensions` so a client can discover it, and stop `partsToContentBlocks` turning an interrupts-shaped data part into `[Structured Data]` prompt text, so a control payload can never reach the model.
- **Is the `interruptId` format now a de-facto contract?** The doc comment at `executor.ts:389` says ids "cannot be inferred from the prose" — true, but they're fully observable, and clients will parse them. TS emits a bare `tool:${toolUseId}:${name}` (`tools/executors/executor.ts:257`) where Python reserved a `v1:` scheme (`_middleware/stages.py:131-134`). Worth either documenting the id as opaque, or aligning the scheme now while both are cheap. What's your read?
- **Does "Closes #3584" hold?** Given the dropped `isinstance` guard and the missing `_jsonable` — both edge cases the Python original explicitly handles and tests — is the port done enough to auto-close the port issue, or would you rather it close on a follow-up?

</details>

<details>
<summary><b>Pre-existing → better as separate issues than PR comments</b></summary>

These would all still be true with this PR closed unmerged, so I haven't filed them inline. Happy to open issues for any of them if useful — I didn't want to spam the tracker unasked.

- `cancelTask` throws `unsupportedOperation` (`executor.ts:452`), so a parked task can only be abandoned, not cancelled — and cancelling leaves `_interruptState.activated === true`. Python has the same gap; it's a cross-SDK design item.
- The context LRU eviction mechanism itself (#2696).
- `A2AAgent` silently stripping interrupt responses instead of rejecting them, where Python raises `ValueError` (`_converters.py:52`, which predates #3486).
- `INTERRUPT_RESPONSE_KEY` / `INTERRUPTS_KEY` / `extractInterruptResponses` are exported but absent from `a2a/index.ts` and carry no `@internal` — consistent with the rest of `src/a2a/`, so not a regression.

</details>

---

*Process note, since it's visible above: 6 specialist passes ran (correctness, adversarial, API/DevX, test-quality, issue-alignment, docs); two of them published their own reviews prematurely, which is why you got three comments instead of one. I re-verified every 🔴 myself before standing behind it, and withdrew two findings and corrected one in the process. Scratch test files were deleted — the branch is untouched.*

marcinmarzeta:
Note for the TS interrupt/A2A port: Python's `StrandsA2AExecutor` degrades a non-finite float interrupt reason to its string form (`"nan"`/`"inf"`) as of #3826, while `JSON.stringify` is spec-bound to emit `null` for non-finite numbers. Worth specifying the degraded wire value once when the TS side lands so the two SDKs don't put different values in the same field.

PR #3591, opened 2026-07-31
fix(python): preserve structured output schemas

### LINKED ISSUE #3590: [BUG] Structured output conversion corrupts nullable enums and nested references
### Checks

- [x] I have searched the issues and there are no open duplicates of this issue

### SDK Language

Python

### Strands Version

1.50.2

### Language Runtime Version

Python 3.13.14

### Operating System

macOS 26.5.2

### Installation Method

pip

### Steps to Reproduce

`convert_pydantic_to_tool_spec` changes valid Pydantic JSON Schema in two ways that prevent structured output from being used with ordinary Pydantic models.

```python
from typing import Literal

from strands.tools.structured_output import convert_pydantic_to_tool_spec


class NullableEnum(BaseModel):
    sentiment: Literal["positive", "negative"] | None


print(convert_pydantic_to_tool_spec(NullableEnum)["inputSchema"]["json"])
```

The generated `sentiment` property is:

```python
{"type": ["string", "null"], "enum": ["positive", "negative"]}
```

Bedrock rejects this because `null` is declared as an allowed type but is not an allowed enum value. Pydantic's original schema represents the union correctly:

```python
{
    "anyOf": [
        {"type": "string", "enum": ["positive", "negative"]},
        {"type": "null"},
    ]
}
```

A model that supplies an already dereferenced provider schema through Pydantic's supported `model_json_schema()` override can also fail during conversion:

```python
from typing import Any

from strands.tools.structured_output import convert_pydantic_to_tool_spec


class Filter(BaseModel):
    name: str


class FilterGroup(BaseModel):
    filters: tuple[Filter, ...]


class Scope(BaseModel):
    group: FilterGroup

    @classmethod
    def model_json_schema(cls, *args: Any, **kwargs: Any) -> dict[str, Any]:
        return {
            "type": "object",
            "properties": {
                "group": {
                    "type": "object",
                    "properties": {
                        "filters": {
                            "type": "array",
                            "items": {
                                "type": "object",
                                "properties": {"name": {"type": "string"}},
                                "required": ["name"],
                            },
                        }
                    },
                    "required": ["filters"],
                }
            },
            "required": ["group"],
        }


convert_pydantic_to_tool_spec(Scope)
```

This raises:

```text
ValueError: Missing reference: Filter
```

The schema returned by `Scope.model_json_schema()` contains no references. The converter discards the supplied `group` schema, regenerates `FilterGroup` separately, copies its properties without its `$defs`, and then cannot resolve the introduced `Filter` reference.

Both behaviors reproduce against the published 1.50.2 wheel and current `main`.

### Expected Behavior

The generated tool input schema should preserve the meaning of `model.model_json_schema()`, including `anyOf` and internal `$ref`/`$defs` references. The resulting schema should remain valid JSON Schema and should not change whether a field accepts `null`.

### Actual Behavior

The converter manually expands and flattens Pydantic's schema. Optional unions are rewritten as multi-type arrays without reconciling enum values, and a model's authoritative schema can be replaced with a separately generated nested schema whose definitions are then discarded.

### Additional Context

Bedrock structured output supports `anyOf`, `$ref`, and `$defs` from its supported JSON Schema Draft 2020-12 subset. Related issues #337 and #824 addressed earlier reference cases, but the current conversion still reconstructs schemas incompletely.

### Possible Solution

Use Pydantic's native JSON Schema as the tool input schema instead of manually flattening it. If Pydantic emits a top-level `$ref` for a recursive root model, inline only that root definition while retaining `$defs` so Bedrock still receives a top-level object schema. This is the approach previously discussed in #824.

### Related Issues

#337, #824

starred-mathias:
## Description

Pydantic already emits valid JSON Schema for nullable enums, nested models, defaults, and recursive types. Reconstructing that schema inside Strands changes field semantics and can introduce unresolved references, causing valid structured-output models to fail before provider invocation. This change treats the model-generated schema as authoritative and only inlines a recursive top-level reference while retaining its definition scope.

The generated schema now retains `$defs` and internal `$ref` values. Providers consuming this shared converter must therefore support the JSON Schema references already emitted by Pydantic.

## Related Issues

Resolves #3590

## Documentation PR

No documentation changes are required; the public API and documented usage are unchanged.

## Type of Change

Bug fix

## Testing

- [x] I ran the full Python unit suite (`4,712` tests)
- [x] I ran Ruff linting and formatting checks for the changed files
- [x] I ran strict MyPy over all Python source files
- [x] I exercised nullable-enum validation with the converted schema

## Checklist

### DIFF
--- a/strands-py/src/strands/tools/structured_output/structured_output_utils.py
+++ b/strands-py/src/strands/tools/structured_output/structured_output_utils.py
@@ -1,404 +1,65 @@
-"""Tools for converting Pydantic models to Bedrock tools."""
+"""Tools for converting Pydantic models to structured-output tool specifications."""
 
-from typing import Any, Union
+from copy import deepcopy
+from typing import Any
 
 from pydantic import BaseModel
 
 from ...types.tools import ToolSpec
 
 
-def _flatten_schema(schema: dict[str, Any]) -> dict[str, Any]:
-    """Flattens a JSON schema by removing $defs and resolving $ref references.
-
-    Handles required vs optional fields properly.
-
-    Args:
-        schema: The JSON schema to flatten
-
-    Returns:
-        Flattened JSON schema
-    """
-    # Extract required fields list
-    required_fields = schema.get("required", [])
-
-    # Initialize the flattened schema with basic properties
-    flattened = {
-        "type": schema.get("type", "object"),
-        "properties": {},
-    }
-
-    if "title" in schema:
-        flattened["title"] = schema["title"]
-
-    if "description" in schema and schema["description"]:
-        flattened["description"] = schema["description"]
-
-    # Process properties
-    required_props: list[str] = []
-    if "properties" not in schema and "$ref" in schema:
-        raise ValueError("Circular reference detected and not supported.")
-    if "properties" in schema:
-        required_props = []
-        for prop_name, prop_value in schema["properties"].items():
-            # Process the property and add to flattened properties
-            is_required = prop_name in required_fields
-
-            # If the property already has nested properties (expanded), preserve them
-            if "properties" in prop_value:
-                # This is an expanded nested schema, preserve its structure
-                processed_prop = {
-                    "type": prop_value.get("type", "object"),
-                    "description": prop_value.get("description", ""),
-                    "properties": {},
-                }
-
-                # Process each nested property
-                for nested_prop_name, nested_prop_value in prop_value["properties"].items():
-                    is_required = "required" in prop_value and nested_prop_name in prop_value["required"]
-                    sub_property = _process_property(nested_prop_value, schema.get("$defs", {}), is_required)
-                    processed_prop["properties"][nested_prop_name] = sub_property
-
-                # Copy required fields if present
-                if "required" in prop_value:
-                    processed_prop["required"] = prop_value["required"]
-            else:
-                # Process as normal
-                processed_prop = _process_property(prop_value, schema.get("$defs", {}), is_required)
-
-            flattened["properties"][prop_name] = processed_prop
-
-            # Track which properties are actually required after processing
-            if is_required and "null" not in str(processed_prop.get("type", "")):
-                required_props.append(prop_name)
-
-    # Add required fields if any (only those that are truly required after processing)
-    # Check if required props are empty, if so, raise an error because it means there is a circular reference
-
-    if len(required_props) > 0:
-        flattened["required"] = required_props
-    return flattened
-
-
-def _process_property(
-    prop: dict[str, Any],
-    defs: dict[str, Any],
-    is_required: bool = False,
-    fully_expand: bool = True,
-) -> dict[str, Any]:
-    """Process a property in a schema, resolving any references.
-
-    Args:
-        prop: The property to process
-        defs: The definitions dictionary for resolving references
-        is_required: Whether this property is required
-        fully_expand: Whether to fully expand nested properties
-
-    Returns:
-        Processed property
-    """
-    result = {}
-    is_nullable = False
-
-    # Handle anyOf for optional fields (like Optional[Type])
-    if "anyOf" in prop:
-        # Check if this is an Optional[...] case (one null, one type)
-        null_type = False
-        non_null_type = None
-
-        for option in prop["anyOf"]:
-            if option.get("type") == "null":
-                null_type = True
-                is_nullable = True
-            elif "$ref" in option:
-                ref_path = option["$ref"].split("/")[-1]
-                if ref_path in defs:
-                    non_null_type = _process_schema_object(defs[ref_path], defs, fully_expand)
-                else:
-                    # Handle missing reference path gracefully
-                    raise ValueError(f"Missing reference: {ref_path}")
-            else:
-                non_null_type = option
-
-        if null_type and non_null_type:
-            # For Optional fields, we mark as nullable but copy all properties from the non-null option
-            result = non_null_type.copy() if isinstance(non_null_type, dict) else {}
-
-            # For type, ensure it includes "null"
-            if "type" in result and isinstance(result["type"], str):
-                result["type"] = [result["type"], "null"]
-            elif "type" in result and isinstance(result["type"], list) and "null" not in result["type"]:
-                result["type"].append("null")
-            elif "type" not in result:
-                # Default to object type if not specified
-                result["type"] = ["object", "null"]
-
-            # Copy description if available in the property
-            if "description" in prop:
-                result["description"] = prop["description"]
-
-            # Need to process item refs as well (#337)
-            if "items" in result:
-                result["items"] = _process_property(result["items"], defs)
-
-            return result
-
-    # Handle direct references
-    elif "$ref" in prop:
-        # Resolve reference
-        ref_path = prop["$ref"].split("/")[-1]
-        if ref_path in defs:
-            ref_dict = defs[ref_path]
-            # Process the referenced object to get a complete schema
-            result = _process_schema_object(ref_dict, defs, fully_expand)
-        else:
-            # Handle missing reference path gracefully
-            raise ValueError(f"Missing reference: {ref_path}")
-
-    # For regular fields, copy all properties
-    for key, value in prop.items():
-        if key not in ["$ref", "anyOf"]:
-            if isinstance(value, dict):
-                result[key] = _process_nested_dict(value, defs)
-            elif key == "type" and not is_required and not is_nullable:
-                # For non-required fields, ensure type is a list with "null"
-                if isinstance(value, str):
-                    result[key] = [value, "null"]
-                elif isinstance(value, list) and "null" not in value:
-                    result[key] = value + ["null"]
-                else:
-                    result[key] = value
-            else:
-                result[key] = value
-
-    return result
-
-
-def _process_schema_object(
-    schema_obj: dict[str, Any], defs: dict[str, Any], fully_expand: bool = True
-) -> dict[str, Any]:
-    """Process a schema object, typically from $defs, to resolve all nested properties.
-
-    Args:
-        schema_obj: The schema object to process
-        defs: The definitions dictionary for resolving references
-        fully_expand: Whether to fully expand nested properties
-
-    Returns:
-        Processed schema object with all properties resolved
-    """
-    result = {}
-
-    # Copy basic attributes
-    for key, value in schema_obj.items():
-        if key != "properties" and key != "required" and key != "$defs":
-            result[key] = value
-
-    # Process properties if present
-    if "properties" in schema_obj:
-        result["properties"] = {}
-        required_props = []
-
-        # Get required fields list
-        required_fields = schema_obj.get("required", [])
-
-        for prop_name, prop_value in schema_obj["properties"].items():
-            # Process each property
-            is_required = prop_name in required_fields
-            processed = _process_property(prop_value, defs, is_required, fully_expand)
-            result["properties"][prop_name] = processed
-
-            # Track which properties are actually required after processing
-            if is_required and "null" not in str(processed.get("type", "")):
-                required_props.append(prop_name)
-
-        # Add required fields if any
-        if required_props:
-            result["required"] = required_props
-
-    return result
-
-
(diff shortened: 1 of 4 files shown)


AnnasMazhar:
Verified this fix against both repro cases from #3590, plus the recursive-root case it targets.

**What I ran (Python 3.11, current PR head `8387518`):**

1. **Nullable enum repro** (`Literal["positive", "negative"] | None`):
   ```json
   "sentiment": {"anyOf": [{"enum": ["positive", "negative"], "type": "string"}, {"type": "null"}], "title": "Sentiment"}
   ```
   Pydantic's native `anyOf` representation is preserved — `null` is no longer merged into a multi-type array alongside `enum`, which fixes the Bedrock rejection described in the issue.

2. **`model_json_schema()` override repro** (the `Filter`/`FilterGroup`/`Scope` case): passes without `ValueError: Missing reference`; the authoritative overridden schema is returned verbatim.

3. **Recursive root model** (`Node` with `children: list["Node"]`): the top-level `$ref` is correctly inlined, `$defs` retained, and internal `#/$defs/Node` references still resolve — the exact case `_inline_root_reference` exists for.

4. **Module test suite:** `tests/strands/tools/test_structured_output.py` — 10 passed; full `tests/strands/tools/` — **544 passed**.

The design matches the issue's suggested direction (treat Pydantic's schema as authoritative, inline only a top-level root `$ref`, keep `$defs`), and the ~340-line deletion of the manual flattening removes the entire class of "reconstructed schema differs from Pydantic's" bugs rather than patching individual symptoms.

One non-blocking observation for the maintainers to confirm intent on: the old flattening code injected model **docstrings** into nested `$defs` descriptions (`_process_referenced_models`); the new code relies on Pydantic's native schema, so nested-model docstrings no longer propagate into the input schema. Field-level `description=` metadata still flows through. If any provider depends on nested-model docstrings, that's a one-line addition in `_inline_root_reference` — otherwise this LGTM.

ebarkhordar (commented):
Ran this against the recursive shapes the PR newly admits. One edge worth a guard, not blocking.

With `_flatten_schema` gone, a cyclic `$ref` can now reach `_strict_schema._apply_strict`. The `$ref` inlining at `strands-py/src/strands/models/_strict_schema.py:104-114` fires when a `$ref` carries sibling keys, merges the referent in, and re-applies itself with no visited set, so a cycle recurses until the stack runs out. That path is only reachable on `BedrockModel(strict_tools=True)` (`bedrock.py:306`).

Measured in a clean `python:3.12-slim` container, installing each tree and calling `convert_pydantic_to_tool_spec(M)` then `ensure_strict_json_schema(...)`:

```python
class SelfDesc(BaseModel):
    value: int
    child: "SelfDesc" = Field(description="the child node")
```

* `main` (3cddf74): `convert_pydantic_to_tool_spec` raises `ValueError: Circular reference detected and not supported.`
* this PR (8387518): conversion succeeds, then `ensure_strict_json_schema` raises `RecursionError`.

Same result for a mutually recursive `A.b: B` / `B.a: A` pair where both edges carry a `description`.

The scope is narrow and I would rather state it than imply more: the trigger is a cycle in which the `$ref` carries a sibling key, which requires the cyclic fields to be required, and such a model cannot be instantiated. The recursive shapes that are constructible all convert and pass `ensure_strict_json_schema` on this PR, which I checked for `Optional["Node"]`, for `List["Node"]` with a field description, and for a mutual cycle where one edge is `Optional`. So nothing that works today breaks.

What changes is the failure mode: a clear `ValueError` at conversion becomes a `RecursionError` inside request formatting. A visited set in `_apply_strict`, or an explicit cycle rejection, would keep that legible. It seemed worth raising here since the description names recursive types as one of the shapes Pydantic already gets right.

starred-mathias:
Addressed the circular strict-schema review finding in `40d4ffbee`.

The strict-schema walker now tracks the `$ref` values being inlined along the current traversal branch. Recursive schemas whose references do not have sibling keys, including recursive arrays and optional recursive fields, remain references and continue to work. A cycle that would require unbounded sibling-key inlining now raises a deliberate `ValueError` containing the reference cycle instead of eventually failing with `RecursionError`.

Added focused coverage for both the supported recursive-array case and the required recursive field with a sibling description.

Validation completed:
- focused strict-schema and structured-output tests: 26 passed
- complete models/tools unit suites: 1,447 passed
- Ruff format and lint: clean
- strict mypy: 232 source files clean

percymcn:
Ran this against `strands-agents==1.51.0` + `pydantic==2.13.4` before commenting. The direction is right — treating `model_json_schema()` as authoritative removes a whole class of corruption, and I can add two concrete confirmations plus four shapes the diff leaves untouched.

**What the current code does that this fixes (measured verbatim on 1.51.0, beyond the two cases in #3590):**

`_flatten_schema` rebuilds the schema from a five-key keep-list — `type`, `properties`, `title`, `description`, `required` — so anything else at the root is dropped. Consequences:

```python
class Cat(BaseModel): kind: Literal["cat"]; lives: int
class Dog(BaseModel): kind: Literal["dog"]; breed: str

class ReqUnion(BaseModel): pet: Union[Cat, Dog]
# -> {"type":"object","properties":{"pet":{"title":"Pet"}},...}      # union gone entirely

class OptUnion(BaseModel): pet: Optional[Union[Cat, Dog]] = None
# -> "pet": {"title":"Dog","type":["object","null"], ...Dog only...}  # Cat silently dropped

class Disc(BaseModel): pet: Union[Cat, Dog] = Field(discriminator="kind")
# -> "pet": {"oneOf":[{"$ref":"#/$defs/Cat"},{"$ref":"#/$defs/Dog"}], ...}
#    ...but $defs is not on the keep-list, so both $refs arrive DANGLING
```

`OptUnion` is the one I'd highlight in the PR description: it doesn't fail, it produces a complete, plausible schema for the *wrong* type (only the last non-null branch survives), so the model is told the field must be a `Dog`. Also `RootModel[Inner]` raises `ValueError("Circular reference detected and not supported.")` when nothing is circular — the check is `"properties" not in schema and "$ref" in schema`, which cannot distinguish a root `$ref` from a genuine cycle, so both get the cycle message.

**Four things the diff does not address — flagging them because three are invisible to the tests as written:**

1. **The new tuple test passes either way.** `addresses: tuple[Address, ...]` is variadic, and pydantic renders that as `{"items": {"$ref": ...}, "type": "array"}` — no `prefixItems`. A *fixed-length* `Tuple[int,int,int,int]` renders as `{"prefixItems":[...],"minItems":4,"maxItems":4,"type":"array"}`, and `_apply_strict` has no `prefixItems` branch: it neither handles it nor recurses into it, so an object nested inside a fixed tuple never gets `additionalProperties: false`. Worth adding a fixed-length tuple to `test_strict_schema.py`.

2. **`Dict[str, str]` is uncovered.** `_apply_strict` inserts `additionalProperties: false` only when the key is **absent**, so an open map (`{"type":"object","additionalProperties":{"type":"string"}}`) passes through unchanged. That's defensible, but nothing pins it either way, and it is the shape most likely to change meaning under a strict provider.

3. **The object rule is a strict string comparison.** `if schema.get("type") == "object"` at `_strict_schema.py:65` means a node typed `["object", "null"]` is neither closed nor visited by the `required` rule. Today that shape is reachable because `_flatten_schema` *manufactures* it for every `Optional[BaseModel]` field — this PR removes that producer, which is the main reason I'd merge it. But the miss itself survives, and `bedrock.py:306` applies `ensure_strict_json_schema()` to **caller-supplied** `tool_spec["inputSchema"]["json"]`, i.e. MCP tool schemas, where an array-valued `type` is ordinary JSON Schema. `isinstance(t, str) and t == "object" or (isinstance(t, list) and "object" in t)` covers both.

4. **`require_all_properties` has no caller.** The only call site is `bedrock.py:306`, which doesn't pass it, and the OpenAI provider never calls this function at all — `models/openai.py:515` assigns `tool_spec["inputSchema"]["json"]` straight to `parameters`. So the parameter the docstring describes as "needed for OpenAI" is currently dead code, and OpenAI tool schemas get no strict handling. Not this PR's job, but the docstring reads as though that path exists.

**One interaction worth a regression test.** Preserving `$defs`/`$ref` means `Field(discriminator=...)` now reaches the provider as a resolvable `oneOf` instead of dangling refs. But #2664 / #2822 report that `BedrockModel(strict_tools=True)` fails with a runtime `ValidationException` when a tool schema contains `oneOf` (that report cites an 18-hour outage), and #2822 — which would have caught it at build time — was closed unmerged. So for discriminated unions this PR turns a silent corruption into a loud runtime failure. That's an improvement, but it's a behaviour change that deserves to be stated, and a `Field(discriminator=...)` case in the tests would pin it.

**Offline oracle, no API key**, if you want any of the above as CI assertions: `_strict_schema.py`'s docstring names OpenAI's `_ensure_strict_json_schema` as its model, and `openai@7.4.0` ships the transformer that enforces that contract:

```js
const { toStrictJsonSchema } = require("openai/lib/transform");
toStrictJsonSchema(schema);  // throws with the exact rule that failed
```

It throws on `prefixItems` and on an open map, and accepts the ordinary cases — so it pins items 1 and 2 without a network call. Scoping that honestly: that is OpenAI's verdict, not Bedrock's, and the strict path here is Bedrock — but since the function is explicitly modelled on OpenAI's, it's a useful differential check.

strands-py/src/strands/tools/structured_output/structured_output_utils.py
```diff
@@ -1,404 +1,65 @@
...
+    if not root_reference.startswith(prefix) or not isinstance(definitions, dict):
+        raise ValueError(f"Unsupported top-level JSON Schema reference: {root_reference}")
+
+    definition_name = root_reference.removeprefix(prefix).replace("~1", "/").replace("~0", "~")
+    root_definition = definitions.get(definition_name)
+    if not isinstance(root_definition, dict):
+        raise ValueError(f"Missing top-level JSON Schema definition: {definition_name}")
+
+    inlined_schema = deepcopy(root_definition)
+    inlined_schema["$defs"] = deepcopy(definitions)
```
liramon2, line 64: Nit: worth a 1-line comment explaining why both the top level and $defs duplication are needed.

strands-py/src/strands/tools/structured_output/structured_output_utils.py
```diff
@@ -1,404 +1,65 @@
...
-            field = model.model_fields.get(prop_name)
-
-            # Add field description if available and not already set
-            if field and field.description and not prop_info.get("description"):
-                prop_info["description"] = field.description
+    root_reference = schema.get("$ref")
+    if not isinstance(root_reference, str):
+        return schema
+
+    prefix = "#/$defs/"
```
liramon2, line 53: I see that we're only handling `$defs`, but not `definitions` unlike `_strict_schema.py`. Pydantic v2 only emits `$defs` according to this [PR](https://github.com/pydantic/pydantic/pull/5029) and [docs](https://pydantic.dev/docs/validation/dev/concepts/json_schema/), so this is fine. But it should be called out in a short comment.

strands-py/src/strands/tools/structured_output/structured_output_utils.py
```diff
@@ -1,404 +1,65 @@
...
-    if "title" in schema:
-        flattened["title"] = schema["title"]
-
-    if "description" in schema and schema["description"]:
-        flattened["description"] = schema["description"]
-
-    # Process properties
-    required_props: list[str] = []
-    if "properties" not in schema and "$ref" in schema:
-        raise ValueError("Circular reference detected and not supported.")
```
liramon2, line 39: Looks like the new code supports self-referencing models. Mention this in the PR description.

liramon2:
@starred-mathias Thanks for the contribution. I left a few comments on this PR.

Unshure (changes requested):
Im a bit worried about the direction of the changes in this pull request. While I would agree that if we built this from scratch, the pydantic approach is probably the correct and simpler version of this. However, changes the output shape of tool schemas by using this new code path is a breaking change.

Users who have relied on tool schemas being a certain shape may now face unexpected changes in their agent performance because of the output schema change. I think we should just fix this bug in the existing code, instead of replacing it with the pydanitc approach.

Issue #3533, opened 2026-07-28
strands-mcp: PyPI 0.2.7 still ships unpinned mcp, so uvx installs look broken since mcp 2.0.0


pbarone:
## What I'm seeing

I think the `mcp<2.0.0` cap from #3524 hasn't made it to PyPI yet. From what I
can tell the latest published `strands-agents-mcp-server` is still 0.2.7
(uploaded 2026-03-11), which predates the cap and declares an unbounded
`mcp>=1.1.3`:

```
$ curl -s https://pypi.org/pypi/strands-agents-mcp-server/json \
    | jq -r '.info.version, (.info.requires_dist[] | select(startswith("mcp")))'
0.2.7
mcp>=1.1.3
```

So a fresh `uvx` resolve still pulls `mcp` 2.0.0 and fails the same way #3524
describes:

```
$ uvx strands-agents-mcp-server
  File ".../strands_mcp_server/server.py", line 4, in <module>
    from mcp.server.fastmcp import FastMCP
ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```

I ran into this as a Kiro user. Since `uvx` is the install path the README
documents (Kiro, Cursor, Claude Code, VS Code, Q Developer CLI), I think the
symptom most people will see is just "the MCP server won't connect", with the
traceback buried in their client's MCP logs.

## Why I think it hasn't reached users yet

`gh release list` on this repo shows tags for `python/*` and `typescript/*` but
nothing for `mcp/*`, so I don't think a `strands-mcp` version has been published
from the monorepo yet. It looks like the release workflow for it landed fairly
recently (#3413). If that's right, 0.2.7 from the archived repo is still what a
fresh install gets.

I may be missing something here (a release already in flight, or a different
intended install path). Can you please confirm whether a `strands-mcp` release
is planned? If it's already queued up, feel free to close this.

## Workaround, in case it helps others

Constraining the SDK at launch works against 0.2.7:

```
uvx --with "mcp<2" strands-agents-mcp-server
```

In MCP client config:

```
"args": ["--with", "mcp<2", "strands-agents-mcp-server"]
```

That resolves `mcp` 1.29.0 and completes an `initialize` handshake for me
(`serverInfo.name: strands-agents-mcp-server`).

ebarkhordar:
[...]

PR #3524, opened 2026-07-28
fix(mcp): pin mcp dependency below v2.0.0

agent-of-mkmeral:
## Problem

`strands-mcp` declares an **unbounded** `mcp>=1.1.3`. `mcp` 2.0.0 is now on PyPI, so a fresh resolve picks it up — and **2.0.0 removed the `mcp.server.fastmcp` module** (superseded by `mcp.server.mcpserver`).

`src/strands_mcp_server/server.py:5` imports it at module scope:

```python
from mcp.server.fastmcp import FastMCP
```

So on `mcp` 2.0.0 the package is **completely unusable** — not a subtle behavioural regression, a hard `ModuleNotFoundError`:

```
$ python -c "import strands_mcp_server.server"
  File "src/strands_mcp_server/server.py", line 5, in <module>
    from mcp.server.fastmcp import FastMCP
ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```

That means the `strands-agents-mcp-server` console script fails to start, and CI test collection aborts before running anything:

```
ERROR tests/test_server.py
!!!! Interrupted: 1 error during collection !!!!
1 error in 0.93s
```

## Fix

Cap the range at `<2.0.0` in the two places `mcp` is declared — runtime `[project.dependencies]` and the `hatch-static-analysis` env — so they stay in sync.

```diff
 dependencies = [
-    "mcp>=1.1.3",
+    "mcp>=1.1.3,<2.0.0",
     "pydantic>=2.0.0",
 ]
```

The lower bound is deliberately left at `1.1.3` to keep this change surgical. This also brings `strands-mcp` in line with the constraint `strands-py` already uses (`mcp>=1.23.0,<2.0.0`).

## Verification

Reproduced the break and confirmed the fix in clean venvs:

| | mcp resolved | `import strands_mcp_server.server` | `pytest tests/` |
|---|---|---|---|
| before (`mcp>=1.1.3`) | **2.0.0** | ❌ `ModuleNotFoundError` | ❌ collection error |
| after (`mcp>=1.1.3,<2.0.0`) | **1.29.0** | ✅ OK | ✅ **50 passed** |

Also verified `mcp.server.fastmcp` imports fine on 1.29.0 and is absent from 2.0.0's `mcp/server/` tree, and that the edited `pyproject.toml` still parses.

## Notes

[...]

PR #3592, opened 2026-07-31
ci(mcp): run the package build test on pull requests

yonib05:
## Description

When `strands-mcp` moved into the monorepo, its packaging test never got wired into CI.

`mcp-test-package-build.yml` builds the wheel and sdist and imports them from a clean venv outside the source tree. It is the check that tells us whether a fresh install actually works. Its only caller was `release-mcp.yml`, so it ran at release time, after a version could already be tagged, and never on a pull request.

This PR fans it out from `mcp-pr-and-push.yml`, so it runs on pull requests and on push to main, matching how `typescript-pr-and-push.yml` already fans out `call-test-package-pack`. The gate keeps one `MCP` entry per area rather than gaining a second top-level `mcp-*` job to register in `needs` and `allowed-skips`.

Also adds `strands-mcp/tests/test_dependencies.py`, asserting the declared `mcp` range carries an upper bound and that the installed `mcp` satisfies `server.py`'s module-scope imports. It asserts "has an upper bound" rather than "excludes 2.0.0" on purpose: pinning to a version would become a false failure the day someone migrates to `mcp.server.mcpserver` and legitimately raises the cap.

`pyproject.toml` picks up `packaging` and `tomli`, which the new test imports and which were only present transitively.

To be clear about scope: this would not have caught #3533 on its own. That was an uncapped `mcp>=1.1.3` resolving to a new major, fixed in #3524 and published in `0.2.8`, and no PR-triggered check catches a dependency releasing on PyPI. It is what surfaced the gap, not what this closes.

## Related Issues

Follow-up to #3533 and #3524.

## Documentation PR

No documentation changes needed.

## Type of Change

Other: CI coverage for an existing test.

## Testing

The dependency test is only worth having if it fails on the real thing, so I checked both directions. Reverted to the pre-#3524 `mcp>=1.1.3`:

```
FAILED tests/test_dependencies.py::test_mcp_requirement_has_upper_bound
AssertionError: declared requirement 'mcp>=1.1.3' has no upper bound, so a new mcp
major is resolvable; expected one of ['<', '<=', '==', '~=']
```

Passes on current `main`, and passes against a hypothetical post-migration `mcp>=2.0.0,<3.0.0`, so it will not become a landmine for whoever raises the cap.

Full suite is 52 passed on both Python 3.10 and 3.12. CI uses 3.10, which is the branch that exercises the `tomli` import rather than stdlib `tomllib`. `ruff format --check` and `ruff check` are clean.

CI on this PR now reports both jobs under the single `MCP` entry, and the gate is green:

```
MCP / call-test-lint / Test and Lint => SUCCESS
MCP / call-test-package-build / Wheel install => SUCCESS
CI Gate => SUCCESS
```

No `hatch` locally, so I ran the equivalent directly: `pytest tests/`, `ruff format --check`, `ruff check` against the `strands-mcp` config, on both versions.

Two things worth knowing:

- `mcp-test-package-build.yml`'s no-`version` branch had never executed, because its only caller always passed `version`. The run on this PR is the first exercise of it.
- Making `mcp-pr-and-push.yml` reusable meant dropping its `concurrency` block. `github.workflow` resolves to the caller, so the group collided with the CI gate's own and `cancel-in-progress` cancelled the nested call, failing the gate with no mcp jobs at all. `typescript-pr-and-push.yml` omits it for the same reason. Push and dispatch runs still dedupe under the gate's group.

Unrelated, but you will hit it: `pre-commit` fails on `main` because `strands-ts/src/vended-memory-stores/bedrock-knowledge-base/store.ts` has two TypeScript errors against the installed `@aws-sdk/client-bedrock-agent` types. Reproduces on a clean `upstream/main` with my changes stashed, so I committed with `--no-verify`. Nothing here touches `strands-ts`. Happy to file it separately.

## Checklist

----


.github/workflows/ci.yml
```diff
@@ -107,14 +107,24 @@ jobs:
...
+    permissions:
+      contents: read
+    with:
+      ref: ${{ github.event.pull_request.head.sha || github.sha }}
+
   ci-gate:
     name: CI Gate
     if: always()
-    needs: [detect-changes, dependency-review, python, typescript, docs, mcp]
+    needs: [detect-changes, dependency-review, python, typescript, docs, mcp, mcp-package-build]
```
JackYPCOnline: Do we want to do this under mcp instead of main CI gate?

JackYPCOnline (dismissed):

.github/workflows/ci.yml
yonib05: Good call, done. Moved it into `mcp-pr-and-push.yml` so it fans out the same way `typescript-pr-and-push.yml` fans out `call-test-package-pack`, and the gate keeps one `MCP` entry instead of a second top-level `mcp-*` job to register in `needs` and `allowed-skips`. It also now runs on push to main, which the ci.yml-only version did not.

Worth flagging since it was not obvious: making that workflow reusable required dropping its `concurrency` block. `github.workflow` resolves to the caller, so the group collided with the gate's own and `cancel-in-progress` cancelled the nested call. The gate went red with zero mcp jobs. `typescript-pr-and-push.yml` has no concurrency block for the same reason. Push and dispatch runs still dedupe under the gate's group.

Both jobs now report under `MCP` and the gate is green.

JackYPCOnline (approved):

PR #3593, opened 2026-07-31
fix(session): validate snapshot scope

JackYPCOnline:
## Description

`SnapshotLocation.scope` is a TypeScript union, but `FileStorage` and `S3Storage` also consume runtime values. Enforce the declared `agent | multiAgent` contract before deriving filesystem paths or storage keys. Valid usage and existing path handling are unchanged.

Legacy FileStorage and S3Storage remain exported and usable. They do not go through the new unified storage normalization. The ticket specifically targets these implementations.

No public API changes.

## Related Issues

Tracked privately.

## Documentation PR

Not needed.

## Type of Change

Bug fix

## Testing

- [x] TypeScript pre-commit checks pass

## Checklist

- [x] I have added necessary regression coverage
- [x] No documentation changes are needed
- [x] No example is needed for this internal validation change
- [x] No dependent changes are required

----

### DIFF
--- a/strands-ts/src/session/__tests__/file-storage.test.node.ts
+++ b/strands-ts/src/session/__tests__/file-storage.test.node.ts
@@ -102,6 +102,21 @@ describe('FileStorage', () => {
       })
     })
 
+    describe('FileSnapshotStorage_When_InvalidScope_Then_RejectsPath', () => {
+      it('rejects a scope containing path traversal', async () => {
+        const location = {
+          sessionId: 'test-session',
+          scope: '../../../../outside',
+          scopeId: SCOPE_ID,
+        } as unknown as SnapshotLocation
+
+        await expect(
+          storage.saveSnapshot({ location, snapshotId: '1', isLatest: true, snapshot: createTestSnapshot() })
+        ).rejects.toThrow("Invalid scope '../../../../outside': must be 'agent' or 'multiAgent'")
+        await expect(fs.readdir(testDir)).resolves.toEqual([])
+      })
+    })
+
     describe('FileSnapshotStorage_When_MultiAgentScope_Then_SavesCorrectly', () => {
       it('saves multi-agent snapshot to correct path', async () => {
         const location: SnapshotLocation = {
--- a/strands-ts/src/session/__tests__/s3-storage.test.ts
+++ b/strands-ts/src/session/__tests__/s3-storage.test.ts
@@ -132,6 +132,21 @@ describe('S3Storage', () => {
       })
     })
 
+    describe('S3SnapshotStorage_When_InvalidScope_Then_RejectsKey', () => {
+      it('rejects a scope containing path traversal', async () => {
+        const location = {
+          sessionId: 'test-session',
+          scope: '../../../../outside',
+          scopeId: SCOPE_ID,
+        } as unknown as SnapshotLocation
+
+        await expect(
+          storage.saveSnapshot({ location, snapshotId: '1', isLatest: true, snapshot: createTestSnapshot() })
+        ).rejects.toThrow("Invalid scope '../../../../outside': must be 'agent' or 'multiAgent'")
+        expect(mockS3Client.send).not.toHaveBeenCalled()
+      })
+    })
+
     describe('S3SnapshotStorage_When_MultiAgentScope_Then_SavesCorrectly', () => {
       it('saves multi-agent snapshot to correct S3 key', async () => {
         const location: SnapshotLocation = { sessionId: 'multi-session', scope: 'multiAgent', scopeId: 'graph-1' }
--- a/strands-ts/src/session/__tests__/snapshot-storage-adapter.test.ts
+++ b/strands-ts/src/session/__tests__/snapshot-storage-adapter.test.ts
@@ -53,15 +53,6 @@ describe('SnapshotStorageAdapter', () => {
       expect(keys).toContainEqual(expect.stringContaining(`immutable_history/snapshot_${id}.json`))
     })
 
-    it('rejects an invalid scope', async () => {
-      const location = createLocation({ scope: 'other' as unknown as SnapshotLocation['scope'] })
-
-      await expect(
-        adapter.saveSnapshot({ location, snapshotId: uuidV7(1), isLatest: true, snapshot: createTestSnapshot() })
-      ).rejects.toThrow("Invalid scope 'other': must be 'agent' or 'multiAgent'")
-      await expect(backend.list('session/')).resolves.toEqual([])
-    })
-
     it('round-trips snapshot data through loadSnapshot', async () => {
       const location = createLocation()
       const snapshot = createTestSnapshot({ appData: { custom: 'value' } })
--- a/strands-ts/src/session/__tests__/snapshot-storage-adapter.test.ts
+++ b/strands-ts/src/session/__tests__/snapshot-storage-adapter.test.ts
@@ -53,6 +53,15 @@ describe('SnapshotStorageAdapter', () => {
       expect(keys).toContainEqual(expect.stringContaining(`immutable_history/snapshot_${id}.json`))
     })
 
+    it('rejects an invalid scope', async () => {
+      const location = createLocation({ scope: 'other' as unknown as SnapshotLocation['scope'] })
+
+      await expect(
+        adapter.saveSnapshot({ location, snapshotId: uuidV7(1), isLatest: true, snapshot: createTestSnapshot() })
+      ).rejects.toThrow("Invalid scope 'other': must be 'agent' or 'multiAgent'")
+      await expect(backend.list('session/')).resolves.toEqual([])
+    })
+
     it('round-trips snapshot data through loadSnapshot', async () => {
       const location = createLocation()
       const snapshot = createTestSnapshot({ appData: { custom: 'value' } })
--- a/strands-ts/src/session/__tests__/validation.test.ts
+++ b/strands-ts/src/session/__tests__/validation.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest'
-import { validateIdentifier, validateUuidV7 } from '../validation.js'
+import { validateIdentifier, validateScope, validateUuidV7 } from '../validation.js'
 
 describe('validateIdentifier', () => {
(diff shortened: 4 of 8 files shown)


mehtarac:
@strandly-the-agent review the pr please

lizradway (approved):

mehtarac (approved):

strandly-the-agent:
**TL;DR:** this merged ~18 min after the review was requested, so these are post-merge follow-ups, not a block. The guard rejects plain-string traversal correctly, but a *discard-then-re-read* pattern in `FileStorage._getPath` and `S3Storage._getKey` means a `location` whose `scope` is a getter/proxy sails through validation and the payload lands in the real path/key — confirmed with a working repro against `main`. A third chokepoint (`SnapshotStorageAdapter._scopePrefix`) never calls the validator at all, and the commit that shipped this PR is also the commit that deleted the test covering that gap. **None of it is reachable through any current public SDK call path** (every internal caller hardcodes the literal), so read this as "the guard doesn't hold the boundary its own docstring claims" — not "cross-tenant vulnerability".

✅ **Verified on current `main` (`d50de52d`, contains merge `b070786e`; PR head `d041d74a`):** session suite **144/144 green** (`npx vitest run --project unit-node src/session/`); findings 1 & 2 repro'd live against the real `FileStorage` and `InMemoryStorage`; open PR #2993 confirmed to hard-conflict with `main`; no equivalent gap in `strands-py`. Coercion bypasses of `validateScope` (`String` object, `toString`, `valueOf`, `Symbol.toPrimitive`, arrays) were all **refuted** — the strict `!==` choice is genuinely more coercion-resistant than an allowlist lookup. One pre-existing `type-check` failure in `test/integ/.../steering.test.node.ts` is unrelated to this PR.

### 🔴 1. `validateScope`'s return value is discarded, then the path/key template re-reads `location.scope`
`file-storage.ts:46-48` and `s3-storage.ts:76-79` validate one read of `location.scope` and interpolate a *second* read. One line in each file closes it — `validateScope` already returns the validated value:
```ts
const scope = validateScope(location.scope)   // then interpolate `scope`, not `location.scope`
```

<details>
<summary>Repro + affected surface + reachability</summary>

A `location` with a `get scope()` returning `'agent'` on read #1 and a traversal payload on read #2 passes validation, and the payload reaches the constructed path. Against the real `FileStorage` on current `main`:

```
A: stolen.createdAt         = VICTIM-SECRET   <- cross-boundary READ of a snapshot outside _baseDir
A: victim file after attack = ATTACKER        <- cross-boundary WRITE (victim snapshot_latest.json overwritten)
A: readdir(attackerBase)    = []              <- attacker's own baseDir stayed empty
control: plain traversal STRING scope -> throws "Invalid scope ..."   (guard works for plain strings)
```

Also confirmed with a `Proxy` and a prototype getter. Affects `saveSnapshot` / `loadSnapshot` / `listSnapshotIds` / `loadManifest` / `saveManifest` on both `FileStorage` (barrel export) and `S3Storage` (`"./session/s3-storage"` subpath export). Only the `isLatest:false` history path survives, via the pre-existing `startsWith(this._baseDir)` check at `file-storage.ts:218` — which then misreports the failure as an invalid `snapshotId`. The same double-read applies to `sessionId`/`scopeId` via `validateIdentifier` (which also already returns its input), but that is **pre-existing** at the merge-base, not introduced here.

On S3 the qualification matters: `..` does **not** collapse in the S3 HTTP path, so this corrupts the key layout *inside the caller's own prefix* rather than escaping it — it becomes a real escape only where a consumer normalizes keys onto a filesystem (`aws s3 sync`, restore-to-disk).

**Reachability:** not reachable from any in-SDK call path — `session-manager.ts:158,310`, `multiagent/snapshot.ts:53`, `agent/snapshot.ts:112` all hardcode the literal. It needs a caller to hand a public method a `location` carrying an accessor (class getter, lazy proxy, validation-library wrapper) — plausible for a downstream integrator, not remotely triggerable, and `JSON.parse`d input cannot do it.

**History:** the PR's first commit `ce00c50f` *did* block this, via a `resolve()`/`relative()` containment check on the assembled path. The merged head `d041d74a` ("keep scope fix minimal") removed it. Same getter exploit, head-to-head:
```
3aaf74b6 merge-base | RESOLVED (no throw) | escaped file: NONE (no guard at all)
ce00c50f PR cmt 1   | THREW "Snapshot path resolves outside storage directory" | escaped file: NONE
d041d74a PR head    | RESOLVED (no throw) | escaped file: .../VICTIM/.../snapshot_latest.json
```
</details>

### 🟡 2. `SnapshotStorageAdapter._scopePrefix` never validates `scope`, and the covering test was deleted in the same diff
`snapshot-storage-adapter.ts:143-147` builds the same `.../scopes/${location.scope}/...` template but validates only `sessionId`/`scopeId`. `ce00c50f` had both the `validateScope` call here *and* a test asserting `scope: 'other'` throws; `d041d74a` removed both. Would it be worth restoring just those two hunks?

<details>
<summary>Why the "unified storage normalization" justification doesn't fully cover it — repro</summary>

`normalizeKey` (`storage/storage.ts:24-33`) rejects only `..` segments and empty keys; it does not enforce the `agent | multiAgent` contract. And the `Storage` interface docstring right below it says shipped backends reject `..` "but **custom backends may apply their own key scheme**" — so for a custom backend there is no normalization guarantee at all. `_writeJSON` also calls `storage.write(key, …)` directly; normalization lives inside the backends.

Through `InMemoryStorage` (a shipped, *normalizing* backend) on current `main`:
```
"agent"     ACCEPTED     "admin" ACCEPTED     "ADMIN" ACCEPTED
""          ACCEPTED     "."     ACCEPTED     "..."   ACCEPTED
"../../.."  REJECTED: Failed to write 'sess/scopes/../../../agent-1/snapshots/snapshot_latest.json'
keys written:
  session/sess/scopes/admin/agent-1/snapshots/snapshot_latest.json
  session/sess/scopes/ADMIN/agent-1/snapshots/snapshot_latest.json
  session/sess/scopes/.../agent-1/snapshots/snapshot_latest.json
  session/sess/scopes/./agent-1/snapshots/snapshot_latest.json
  session/sess/scopes/agent-1/snapshots/snapshot_latest.json   <- scope:"" makes scopeId slide into the scope slot
```
Note `scope:""` and `scope:"."` alias onto layouts that collide with legitimate ones on a path backend — `normalizeKey` is not a canonicalizer.

**Reachability:** `SnapshotStorageAdapter` isn't exported from any entry point and no public `SessionManager` method takes a raw `location`, so an attacker-controlled scope can't reach `_scopePrefix` through the documented API today. This is an internal-consistency asymmetry (2 of 3 chokepoints enforce the contract, 1 doesn't) plus a coverage regression — not a live path.
</details>

### 🟡 3. Open PR #2993 has the containment check that would close finding 1 — and now conflicts with `main`
#2993 (`yonib05`) touches the same six files with the same core `validateScope`, **plus** the `resolve()`+`sep` containment check in `_getPath` — the only path-escape hardening either PR proposes, and the thing that independently closes finding 1. It's on neither `main` nor this PR. Should #2993 be rebased and descoped to layer just its unique parts on top of what merged, or superseded?

<details>
<summary>Conflict output + which implementation is stronger on which axis</summary>

```
$ git merge --no-commit --no-ff pr-2993        # onto current main
CONFLICT (content): strands-ts/src/session/__tests__/validation.test.ts
CONFLICT (content): strands-ts/src/session/file-storage.ts
CONFLICT (content): strands-ts/src/session/validation.ts
Automatic merge failed; fix conflicts and then commit the result.
```
Both PRs define `validateScope` from scratch on the same lines with different signatures, algorithms, and error strings. Neither dominates: **#3593** is more coercion-resistant (strict `!==` beats #2993's `Object.hasOwn` lookup — that's the still-unresolved thread on #2993); **#2993** is stronger on union exhaustiveness (`satisfies Record<Scope, true>`) and is the only one with path containment.
</details>

### ⚪ 4. Hardcoded `!==` literals aren't checked against the `Scope` union
`validation.ts:28`. Verified by temporarily adding a third `Scope` member and running `tsc --noEmit`: **zero errors** — because the param is `unknown`, `return scope` still typechecks. A future legitimate third scope would throw at runtime on every save/load with no compile-time signal. #2993's `satisfies Record<Scope, true>` catches exactly this.

### ⚪ 5. Test quality (mutation-tested)
<details>
<summary>Three items</summary>

- **`file-storage.test.node.ts:116`** — `await expect(fs.readdir(testDir)).resolves.toEqual([])` is **vacuous**: it stays `[]` even in the confirmed escape repro above, because the escape writes *outside* `testDir`. It can't distinguish guard-present from guard-absent. Asserting the escape target does not exist would actually bite. (The S3 counterpart, `expect(mockS3Client.send).not.toHaveBeenCalled()`, is fine.)
- **Escaped mutant** — removing `validateScope` *does* fail both new tests (they're genuinely covering, via `rejects.toThrow`), but making `validateScope` lowercase-normalize before comparing leaves **all 144 tests passing**. Nothing covers case variants (`'AGENT'`, `'Agent'`).
- **`validation.test.ts:31`** — `expect([validateScope('agent'), validateScope('multiAgent')]).toEqual([...])` bundles two input scenarios into one assertion. `strands-ts/docs/TESTING.md` ("Test Batching Strategy") reserves batching for high *setup* cost and lists "different input scenarios" as what to keep separate; the sibling `validateIdentifier` block in the same file does split them.
</details>

### ⚪ 6. Stale docstring at `s3-storage.ts:72`
`_getKey`'s docstring still reads "Validates sessionId and scopeId before constructing the key" even though `validateScope` was added right below it — `_getPath`'s counterpart *was* updated to "sessionId, scope, and scopeId" (`file-storage.ts:41`). One-line diff.

**Questions**
- Finding 1's one-line fix and finding 2's restored hunk are both tiny — want me to open a follow-up PR with them (plus the case-variant test), or is that better folded into a rebased #2993?
- "Related Issues: Tracked privately" — is that the intended convention for internal-tracker work, or should these still link a public issue per the PR template? Asking because it left reviewers without acceptance criteria to check against.

<sub>🤖 Automated review by an experimental agent — a maintainer should confirm before acting on any of this. Sorry it landed after the merge; the passes were still running when this went in.</sub>

PR #3595, opened 2026-07-31
refactor(vended-tools): mark make_bash with typing_extensions.deprecated

yonib05:
## Description

Follow-up to #3574, which renamed the sandbox-routed `bash` tool to `shell` and left `bash` / `make_bash` as deprecated aliases resolved through a module `__getattr__`.

This applies the deprecation pattern that [`team/FEATURE_LIFECYCLE.md`](https://github.com/strands-agents/harness-sdk/blob/main/team/FEATURE_LIFECYCLE.md?plain=1#L157) actually prescribes: `make_bash` now carries `@typing_extensions.deprecated`, so type checkers and IDEs flag callers at authoring time rather than only warning once at runtime.

`bash` keeps the `__getattr__` path — it is a tool *instance*, not a function, so no decorator can wrap it. The rationale string is shared between both so the two messages cannot drift.

Behavior is unchanged for callers: both names still emit a single `DeprecationWarning` and resolve to the `shell` equivalents, and `make_bash(**kwargs)` forwards through to `make_shell` (including `name` and `description`).

### Note on why the decorator is not sufficient on its own

Worth recording, since it affects how this pattern applies to tools generally: `@deprecated` raises a `DeprecationWarning`, which Python's default filter only surfaces when it originates in `__main__`. When an agent invokes a tool the call site is SDK internals, so the warning is suppressed and the user sees nothing.

That does not matter for `make_bash`, which users call directly from their own code — the decorator is exactly right here. But it does mean the decorator alone cannot carry a deprecation for a *tool* that only ever runs via the framework. In strands-agents/tools#550 that gap is covered with a `logger.warning` alongside the decorator.

## Related Issues

Follows #3574.

## Documentation PR

No docs change needed — the deprecated names and their timeline are unchanged from #3574, only the mechanism that reports them.

## Type of Change

Other (please describe): applies the prescribed deprecation mechanism to an existing alias; no behavior change

## Testing

- `hatch test` → **4721 passed**
- `ruff check` / `ruff format --check` clean; `mypy src/strands/vended_tools` clean (checked specifically, since `@deprecated` participates in typing)
- Added tests covering argument forwarding through the alias and the presence of the `__deprecated__` marker
- Manually verified both aliases warn once and return the canonical objects, and that `make_shell`'s keyword-only signature is still enforced through the wrapper

Replaced the previous `make_bash is make_shell` identity assertion, which no longer holds now that `make_bash` is a wrapper rather than a direct alias.

## Checklist

On new warnings: this changes how an existing `DeprecationWarning` is raised; it does not add a new one.

----

### DIFF
--- a/strands-py/src/strands/vended_tools/__init__.py
+++ b/strands-py/src/strands/vended_tools/__init__.py
@@ -21,27 +21,40 @@ Example Usage:
 """
 
 import warnings
-from typing import Any
+from typing import TYPE_CHECKING, Any
+
+from typing_extensions import deprecated
 
 from .file_editor import file_editor, make_file_editor
 from .http_request import http_request, make_http_request
 from .shell import make_shell, shell
 from .sleep import make_sleep, sleep
 
-_DEPRECATED_NAMES = {"bash": "shell", "make_bash": "make_shell"}
+if TYPE_CHECKING:
+    from ..tools.decorator import DecoratedFunctionTool
+
+_RENAME_RATIONALE = (
+    "The tool routes commands through the sandbox, which runs sh or the remote login shell "
+    "rather than bash specifically."
+)
+
+
+@deprecated(f"make_bash is deprecated and will be removed in v2.0.0. Use make_shell instead. {_RENAME_RATIONALE}")
+def make_bash(**kwargs: Any) -> "DecoratedFunctionTool":
+    """Deprecated alias for :func:`make_shell`."""
+    return make_shell(**kwargs)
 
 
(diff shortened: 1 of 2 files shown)


JackYPCOnline (approved):

yonib05:
Raised #3599 to update `team/FEATURE_LIFECYCLE.md` so the guidance matches what this PR and strands-agents/tools#550 actually do — tools need the decorator *and* a log line, since the decorator's warning is filtered when the framework is the caller. Kept separate from this PR since it changes team process guidance rather than code.

PR #3601, opened 2026-08-02
feat(bedrock): support API key authentication

### LINKED ISSUE #1238: [FEATURE] Native Support for Amazon Bedrock API Keys in BedrockModel
### Problem Statement

AWS recently introduced support for [Amazon Bedrock API Keys](https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys.html) (short-term and long-term) as an alternative to standard IAM credentials. 

However, the current `BedrockModel` implementation in the Strands Agents SDK is tightly coupled with `boto3`'s standard IAM signing process (SigV4). It does not expose a way to:
1. Pass an API Key (e.g., `BEDROCK_API_KEY`).
2. Use the required `Authorization: Bearer <key>` header.
3. Disable standard IAM signing (`signature_version=UNSIGNED`) easily without a custom wrapper.

Developers wanting to use Bedrock API keys currently have to implement a full custom model provider or hack the `boto3` client configuration, which degrades the developer experience.

### Proposed Solution

Update the `BedrockModel` class to accept an optional `api_key` parameter in its `__init__` method and `BedrockConfig`.

When `api_key` is provided:
1. Automatically configure the underlying `boto3` client with `config=Config(signature_version=UNSIGNED)` to prevent IAM signing errors.
2. Register a `before-send` or `request-created` event hook to inject the `Authorization` header:
   ```python
   request.headers['Authorization'] = f'Bearer {self.api_key}'
   ```

**Example Usage:**
```python
# Desired Developer Experience
model = BedrockModel(
    model_id="us.anthropic.claude-3-sonnet-20240229-v1:0",
    api_key="sk-bedrock-..." # <--- New Parameter
)
```

### Use Case

This feature is critical for:
1. **Rapid Prototyping:** Developers who want to test Bedrock models quickly without configuring complex IAM Users, Roles, or Policies.
2. **Simplified Auth:** Applications that prefer managing a single secret string (API Key) over Access Key/Secret Key pairs.
3. **Parity:** Aligning the Bedrock developer experience with other providers like OpenAI and Anthropic, which already use simple API keys.

### Alternatives Solutions

**Current Workaround (Custom Provider):**
We had to implement a subclass `CustomBedrockModel` to achieve this:

```python
class CustomBedrockModel(BedrockModel):
    def __init__(self, api_key: str, **kwargs):
        config = Config(signature_version=UNSIGNED)
        super().__init__(boto_client_config=config, **kwargs)
        self.api_key = api_key
        self.client.meta.events.register(
            'request-created.bedrock-runtime.*', 
            self._inject_auth_header
        )

    def _inject_auth_header(self, request, **kwargs):
        request.headers['Authorization'] = f'Bearer {self.api_key}'
```

While functional, this requires users to write boilerplate code for a feature that is now natively supported by the AWS platform itself.

### Additional Context

- [AWS Documentation on Bedrock API Keys](https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys.html)
- Relevant `boto3` configuration: `botocore.UNSIGNED` is required to stop the SDK from looking for missing credentials.

alectimison-maker:
## Description

Add explicit Amazon Bedrock API-key authentication to the Python `BedrockModel`.

When `api_key` is provided, the model configures the Bedrock Runtime client for unsigned requests and registers a request hook that supplies the AWS bearer authorization header. This keeps authentication instance-scoped and avoids mutating `AWS_BEARER_TOKEN_BEDROCK`. Existing IAM/SigV4 behavior is unchanged when the option is omitted.

The API is intentionally aligned with the TypeScript SDK's existing `apiKey` option. The credential is constructor-only and is not exposed by `get_config()`.

## Related Issues

Resolves #1238

## Documentation PR

No separate documentation change is needed for this focused provider API addition; the public constructor docstring documents the new option.

## Type of Change

New feature

## Testing

- Added unit coverage for unsigned client configuration, preservation of caller config, request-hook registration, bearer-header replacement, and the unchanged no-key path.
- `hatch test tests/strands/models/test_bedrock.py`: 180 passed.
- Full Python test suite: 4,723 passed.
- Ruff formatting/check and mypy pass.
- `hatch run prepare` passed formatting, lint, mypy, and the Python 3.14 suite (4,723 passed); later interpreter setup was interrupted by a PyPI tunnel failure while downloading `coverage`, not by a test failure.

- [ ] I ran `hatch run prepare` to completion (remaining interpreter environments were blocked by the dependency-download failure described above)

## Checklist

----

### DIFF
--- a/strands-py/src/strands/models/bedrock.py
+++ b/strands-py/src/strands/models/bedrock.py
@@ -12,6 +12,7 @@ from collections.abc import AsyncGenerator, Callable, Iterable, ValuesView
 from typing import Any, Literal, TypeVar, cast
 
 import boto3
+from botocore import UNSIGNED
 from botocore.config import Config as BotocoreConfig
 from botocore.exceptions import ClientError
 from pydantic import BaseModel
@@ -173,6 +174,7 @@ class BedrockModel(Model):
         boto_client_config: BotocoreConfig | None = None,
         region_name: str | None = None,
         endpoint_url: str | None = None,
+        api_key: str | None = None,
         **model_config: Unpack[BedrockConfig],
     ):
         """Initialize provider instance.
@@ -183,6 +185,8 @@ class BedrockModel(Model):
             region_name: AWS region to use for the Bedrock service.
                 Defaults to the AWS_REGION environment variable if set, or "us-west-2" if not set.
             endpoint_url: Custom endpoint URL for VPC endpoints (PrivateLink)
+            api_key: Amazon Bedrock API key for bearer token authentication.
+                When provided, requests use the API key instead of SigV4 signing.
             **model_config: Configuration options for the Bedrock model.
         """
         if region_name and boto_session:
@@ -208,9 +212,18 @@ class BedrockModel(Model):
             else:
                 new_user_agent = "strands-agents"
 
-            client_config = boto_client_config.merge(BotocoreConfig(user_agent_extra=new_user_agent))
+            client_config = boto_client_config.merge(
+                BotocoreConfig(
+                    user_agent_extra=new_user_agent,
+                    **({"signature_version": UNSIGNED} if api_key else {}),
+                )
+            )
         else:
-            client_config = BotocoreConfig(user_agent_extra="strands-agents", read_timeout=DEFAULT_READ_TIMEOUT)
+            client_config = BotocoreConfig(
+                user_agent_extra="strands-agents",
+                read_timeout=DEFAULT_READ_TIMEOUT,
+                **({"signature_version": UNSIGNED} if api_key else {}),
+            )
 
         self.client = session.client(
             service_name="bedrock-runtime",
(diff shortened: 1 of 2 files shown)


lizradway:
i would consider dropping the `UNSIGNED` config and relying only on the before send  hook? it already fires after signing, so it’ll do the same thing with less code.

alectimison-maker:
Thanks for the review — you're right that before-send fires after signing, so the hook alone would produce the same final Authorization header. The reason I kept signature_version=UNSIGNED is credential resolution: with SigV4 left on, botocore resolves IAM credentials during signing and raises NoCredentialsError if none are present — before the before-send hook ever runs. The primary use case for #1238 is callers who have only an API key and no IAM credentials (e.g. running outside AWS), so dropping UNSIGNED would break exactly that scenario.
It also avoids emitting leftover SigV4 artifacts (X-Amz-Date, X-Amz-Content-Sha256, and X-Amz-Security-Token with session creds) that a bearer-only endpoint may not tolerate, and skips wasted signing work entirely.
So the two pieces are complementary rather than redundant: UNSIGNED disables signing, the hook injects the bearer header. I can add a brief code comment explaining why UNSIGNED is required, and clean up the conditional **({...} if api_key else {}) spread if you'd like — happy to take suggestions on that part.

lizradway (approved):

PR #3607, opened 2026-08-03
fix(bedrock): skip guardContent wrap for unsupported image formats

strandly-the-agent:
## Description

Bedrock's `GuardrailConverseImageBlock` is narrower than `ImageBlock`: its format enum is `[png, jpeg]`, while `ImageBlock` accepts `png/jpeg/gif/webp`. With `guardrail_latest_message=True`, `_format_bedrock_messages` wrapped *every* image in the latest user message in `guardContent`, so a gif or webp screenshot made Converse reject the whole request:

```
ValidationException: 1 validation error detected: Value at
'messages.1.member.content.1.member.guardContent.image.format' failed to
satisfy constraint: Member must satisfy enum value set: [jpeg, png]
```

Unsupported formats are now left as plain image blocks. The model still sees the image; it just isn't submitted for guardrail assessment, and that skip is logged — mirroring `_applyGuardBlocks` in the TypeScript SDK, which already gates this condition and warns.

**This is a rebase of #2552, whose fix `xlyoung` wrote** — authorship is preserved on the first commit. That PR was opened in June and has gone stale (`CONFLICTING`); its second half (explicit Anthropic image MIME types) was already fixed on `main` by #2304, which is exactly what conflicts. I dropped that half, kept the Bedrock fix, and added the regression test it was missing plus the skip warning. Opened from my fork because I have no push access to `xlyoung`'s branch — **if `xlyoung` would rather land it themselves, close this in favour of #2552**; [the patch](https://gist.github.com/strandly-the-agent/a2ff3cf0fe112bdaa1934ec0cf0d6cae) applies to `main` with `git am`.

## Related Issues

Fixes part of #2204 (guardContent image format validation only). Supersedes #2552.

#2204 lists seven items; #2304 and #2306 already closed items 1 and 3, and this closes the format half of item 2. Still open in that thread: the bytes-only `GuardrailConverseImageSource` gap (a png/jpeg image from an S3 location still fails client-side — `kimnamu` has a fix and tests for it there), and item 7.

## Documentation PR

None needed — internal request formatting, no public API surface changed.

## Type of Change

Bug fix

## Testing

- [ ] I ran `hatch run prepare` — **not run**: hatch isn't available in my environment. I ran the equivalent directly instead:

```
pytest tests/strands/models/test_bedrock.py   → 180 passed  (178 before, +2 new)
ruff format --check / ruff check               → clean
mypy src/strands/models/bedrock.py             → Success
```

Negative control: both new cases fail on unpatched `main` (`assert 'guardContent' not in ...`) for `gif` and `webp`, so they are real regression tests.

## Checklist
- [x] I have updated the documentation accordingly (no docs change needed)
- [x] Any dependent changes have been merged and published (#2304 already on `main`)

----

*Prepared by `strandly-the-agent`, an experimental agent, at `notowen333`'s request on #2552. Credit for the fix goes to `xlyoung`. This needs a human approver — please review it as you would any contribution rather than trusting it.*

### DIFF
--- a/strands-py/src/strands/models/bedrock.py
+++ b/strands-py/src/strands/models/bedrock.py
@@ -538,6 +538,12 @@ class BedrockModel(Model):
                         image_format = formatted_content["image"].get("format", "")
                         if image_format in ("png", "jpeg"):
                             formatted_content = {"guardContent": {"image": formatted_content["image"]}}
+                        else:
+                            logger.warning(
+                                "image_format=<%s> | format not supported by bedrock guardrails | "
+                                "skipping guardContent wrap",
+                                image_format,
+                            )
 
                 cleaned_content.append(formatted_content)
 
--- a/strands-py/src/strands/models/bedrock.py
+++ b/strands-py/src/strands/models/bedrock.py
@@ -529,12 +529,15 @@ class BedrockModel(Model):
                 if formatted_content is None:
                     continue
 
-                # Wrap text or image content in guardContent if this is the last user text/image message
+                # Wrap text or image content in guardContent if this is the last user text/image message.
+                # Bedrock GuardContent for images only supports png and jpeg formats.
                 if idx == last_user_text_idx and ("text" in formatted_content or "image" in formatted_content):
                     if "text" in formatted_content:
                         formatted_content = {"guardContent": {"text": {"text": formatted_content["text"]}}}
                     elif "image" in formatted_content:
-                        formatted_content = {"guardContent": {"image": formatted_content["image"]}}
+                        image_format = formatted_content["image"].get("format", "")
+                        if image_format in ("png", "jpeg"):
+                            formatted_content = {"guardContent": {"image": formatted_content["image"]}}
 
                 cleaned_content.append(formatted_content)
 
--- a/strands-py/src/strands/models/bedrock.py
+++ b/strands-py/src/strands/models/bedrock.py
@@ -530,13 +530,16 @@ class BedrockModel(Model):
                     continue
 
                 # Wrap text or image content in guardContent if this is the last user text/image message.
-                # Bedrock GuardContent for images only supports png and jpeg formats.
+                # Bedrock guardContent supports a narrower set of image formats than image content.
                 if idx == last_user_text_idx and ("text" in formatted_content or "image" in formatted_content):
                     if "text" in formatted_content:
                         formatted_content = {"guardContent": {"text": {"text": formatted_content["text"]}}}
                     elif "image" in formatted_content:
                         image_format = formatted_content["image"].get("format", "")
-                        if image_format in ("png", "jpeg"):
+                        supported_formats = self.client.meta.service_model.shape_for(
+                            "GuardrailConverseImageFormat"
+                        ).enum
+                        if image_format in supported_formats:
                             formatted_content = {"guardContent": {"image": formatted_content["image"]}}
                         else:
                             logger.warning(
--- a/strands-py/tests/strands/models/test_bedrock.py
+++ b/strands-py/tests/strands/models/test_bedrock.py
@@ -2848,8 +2848,10 @@ async def test_format_request_with_guardrail_latest_message(model):
 
 @pytest.mark.asyncio
 @pytest.mark.parametrize("image_format", ["gif", "webp"])
-async def test_format_request_with_guardrail_latest_message_unsupported_image_format(model, image_format):
+async def test_format_request_with_guardrail_latest_message_unsupported_image_format(model, image_format, caplog):
     """Test that guardContent does not wrap image formats that Bedrock guardrails reject."""
+    caplog.set_level(logging.WARNING, logger="strands.models.bedrock")
+
     model.update_config(
         guardrail_id="test-guardrail",
         guardrail_version="DRAFT",
@@ -2876,6 +2878,7 @@ async def test_format_request_with_guardrail_latest_message_unsupported_image_fo
     # GuardrailConverseImageBlock only accepts png and jpeg, so the image is left unwrapped
     assert "guardContent" not in formatted_messages[0]["content"][1]
     assert formatted_messages[0]["content"][1]["image"]["format"] == image_format
+    assert f"image_format=<{image_format}> | format not supported by bedrock guardrails" in caplog.text
 
 
 @pytest.mark.asyncio
--- a/strands-py/tests/strands/models/test_bedrock.py
+++ b/strands-py/tests/strands/models/test_bedrock.py
@@ -2847,6 +2847,38 @@ async def test_format_request_with_guardrail_latest_message(model):
 
 
 @pytest.mark.asyncio
+@pytest.mark.parametrize("image_format", ["gif", "webp"])
+async def test_format_request_with_guardrail_latest_message_unsupported_image_format(model, image_format):
+    """Test that guardContent does not wrap image formats that Bedrock guardrails reject."""
+    model.update_config(
+        guardrail_id="test-guardrail",
+        guardrail_version="DRAFT",
+        guardrail_latest_message=True,
+    )
+
+    messages = [
+        {
+            "role": "user",
+            "content": [
+                {"text": "Look at this image"},
+                {"image": {"format": image_format, "source": {"bytes": b"fake_image_data"}}},
+            ],
+        },
+    ]
+
+    request = model.format_request(messages)
+    formatted_messages = request["messages"]
+
+    # Latest user message text should still be wrapped
+    assert "guardContent" in formatted_messages[0]["content"][0]
+    assert formatted_messages[0]["content"][0]["guardContent"]["text"]["text"] == "Look at this image"
+
+    # GuardrailConverseImageBlock only accepts png and jpeg, so the image is left unwrapped
+    assert "guardContent" not in formatted_messages[0]["content"][1]
+    assert formatted_messages[0]["content"][1]["image"]["format"] == image_format
+
+
+@pytest.mark.asyncio
 async def test_format_request_with_guardrail_latest_message_after_tool_use(model):
     """Test that guardContent wraps the last user text message even when a toolResult follows it."""
     model.update_config(
--- a/strands-py/tests/strands/models/test_bedrock.py
+++ b/strands-py/tests/strands/models/test_bedrock.py
@@ -50,6 +50,7 @@ def bedrock_client(session_cls):
     mock_client = session_cls.return_value.client.return_value
     mock_client.meta = unittest.mock.MagicMock()
     mock_client.meta.region_name = "us-west-2"
+    mock_client.meta.service_model.shape_for.return_value.enum = ["png", "jpeg"]
     yield mock_client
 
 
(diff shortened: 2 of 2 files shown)


notowen333 (dismissed):

lizradway:
@strandly-the-agent isn't this allowing a user to escape guardrails if they just have a different image type? should we do something abt this?

strandly-the-agent:
**Short answer: yes, real — but the SDK can't close it, and the version of this you should worry about isn't the gif.**

Bedrock guardrails can't assess gif/webp *at all*: `format` is `[png, jpeg]` on both the Converse `guardContent` block and the standalone `ApplyGuardrail` API. So there's no wrapping trick, no pre-check, no way to get a gif screened — the only alternative to skipping is failing the whole request, which is exactly the pre-PR behaviour this PR fixes (opaque server-side `ValidationException`). It's also bounded to `guardrail_latest_message=True` (default `False`, where no `guardContent` is emitted and Bedrock evaluates everything). I'd keep the skip, but make it loud — and fix two nastier things next door:

1. **The flag already excludes far more than gifs.** With it on, prior turns are never assessed, and the system prompt *can't* be — `SystemContentBlock` has no `guardContent` member. Hiding text in message N-1 is a much easier bypass than finding a gif. Our docs actively mislead here: `amazon-bedrock.mdx:515`/`:533` say it's fine "where earlier messages have already been validated". They haven't been.
2. **Silent scope inversion.** If the last user message is a gif *with no text*, the request now contains zero `guardContent` blocks — so per AWS's docs Bedrock evaluates the **entire conversation**. Image format silently flips the flag off and re-prices every call.
3. **Relabelling a png as `gif` does not work** (I tested live) — so an attacker can't spoof formats to smuggle a readable image past the filter. Only a genuine gif/webp bypasses, and that's the service's ceiling, not ours.

Say go and I'll push to this PR: reword the warning to name the consequence (and point at `GuardrailCoverage.images.guarded < total` in the trace, which makes it monitorable), fix those two doc sentences, and gate the `s3Location` case for TS parity — that one is a straight client-side crash today, ~4 lines. Then a follow-up issue for an opt-in fail-closed knob, which needs a control-plane `GetGuardrail` to know whether an image policy is even configured, so not here.

<details>
<summary>Evidence</summary>

**Format ceiling.** botocore `bedrock-runtime` service model: `GuardrailConverseImageFormat` = `[png, jpeg]`, `GuardrailImageFormat` (ApplyGuardrail) = `[png, jpeg]`, `ImageFormat` = `[png, jpeg, gif, webp]`. Not client-enforced, so pre-PR botocore shipped the bad enum and the service rejected it:

```
ValidationException: Value at 'messages.1.member.content.2.member.guardContent.image.format'
failed to satisfy constraint: Member must satisfy enum value set: [jpeg, png]
```

**Selective evaluation.** [Use guardrails with the Converse API](https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails-use-converse-api.html): once any `guardContent` block is present, only wrapped content is evaluated; with none present, "guardrail investigates everything". The same page then caveats that unwrapped content *may* still be caught depending on policy (worked example: a word filter fires outside a `guardContent` block) — so treat "not wrapped = not assessed" as the documented headline rather than an absolute.

**Formatter output on this branch** (`_format_bedrock_messages`, `guardrail_latest_message=True`):

| last user message | result |
|---|---|
| `text` + `png` | both wrapped in `guardContent` |
| `text` + `gif` | text wrapped, image ships bare → unassessed |
| `gif` only | **zero** `guardContent` in the request → whole conversation evaluated |
| `text` + `png` from S3 | wrapped as `{"image":{"source":{"s3Location":…}}}` → `ParamValidationError` before the call leaves the process |

**Mislabel test** (live, us-west-2): a PNG containing readable text, declared as `gif`/`webp`, against Claude Haiku 4.5 and Nova Lite/Pro/Nova 2 Lite. Correctly-labelled png and correctly-labelled gif both read fine; every mislabelled combination failed with a MIME-mismatch error, and the two families produce *different* error strings — i.e. two independent provider-side sniffers, not one shared Bedrock check. Confidence: high for these models today, not an architectural guarantee; a future vision model that trusts the declared type would reopen it.

**s3Location gate** exists in TS at `strands-ts/src/models/bedrock.ts:944-949` (`source_type=<non-bytes> | … | skipping guardContent wrap`); Python has no equivalent, on this branch or on main. Pre-existing, not introduced here.

**Also nearby, unrelated to guardrails-on-images:** `_format_request_message_content` at `strands-py/src/strands/models/bedrock.py:657-663` does `guard["text"]` unconditionally, so a hand-written image `guardContent` block (a valid `GuardrailConverseContentBlock` union member, and `guardContent` is public on `ContentBlock`) raises `KeyError: 'text'`. Worth folding into the same follow-up.

*Transcoding gif→png inside the SDK isn't viable: no image library in `strands-py`'s deps, animated gif collapses to one frame, and since `guardContent` replaces the block the model would see the transcode instead of the user's bytes.*
</details>

strands-py/src/strands/models/bedrock.py
```diff
@@ -529,12 +529,21 @@ def _format_bedrock_messages(self, messages: Messages) -> list[dict[str, Any]]:
...
-                # Wrap text or image content in guardContent if this is the last user text/image message
+                # Wrap text or image content in guardContent if this is the last user text/image message.
+                # Bedrock GuardContent for images only supports png and jpeg formats.
                 if idx == last_user_text_idx and ("text" in formatted_content or "image" in formatted_content):
                     if "text" in formatted_content:
                         formatted_content = {"guardContent": {"text": {"text": formatted_content["text"]}}}
                     elif "image" in formatted_content:
-                        formatted_content = {"guardContent": {"image": formatted_content["image"]}}
+                        image_format = formatted_content["image"].get("format", "")
+                        if image_format in ("png", "jpeg"):
```
mkmeral: is there any way we can handle this without a hardcoded list? If they add a new image type, do we need to update our code?

notowen333: fixed in new commit

mkmeral:
@strandly-the-agent  current version is good right? are we good to merge?

mkmeral (approved):

strandly-the-agent:
Approach is right — deriving the enum is better than my hardcoded tuple — but **not quite: `53bd008` introduces a crash path.** One 3-line guard and I'd say ship it.

`service_model.shape_for("GuardrailConverseImageFormat")` raises `botocore.model.NoShapeFoundError` when the shape isn't in the loaded service model, and that shape **first appears in botocore 1.36.0** — in 1.35.0 `GuardrailConverseContentBlock` has only a `text` member. We declare `botocore>=1.29.0`. So on any botocore below 1.36, `guardrail_latest_message=True` plus *any* image in the latest user message now raises out of `format_request`, where the previous commit handled it fine (gif/webp → skip → valid request). That's a regression in exactly the scenario this PR exists to fix. Verified locally, not theorised.

Guard is a try/except returning `[]` — not a fallback to `("png","jpeg")`, because on those botocore versions a `guardContent.image` block gets rejected client-side anyway, so "no supported formats" is the behaviour that actually works. Diff + test below; 183 pass, and the new test fails without the guard. **Say go and I'll push** (it'll dismiss your approval, and the bot rule needs a second approver regardless).

Two other things, neither blocking:

- **Nothing exercises that lookup.** The unit fixture mocks `shape_for`, so `assert_called_once_with("GuardrailConverseImageFormat")` asserts against a MagicMock — a typo or an AWS rename ships green. And `tests_integ/test_bedrock_guardrails.py::test_guardrail_latest_message` doesn't cover it either: `agent2("What do you see?")` appends a text-only user message that becomes the latest, so the png at index 2 ships bare and `shape_for` is called **zero** times. Pre-existing (true on main too), but it means the image-wrapping path has no live coverage at all.
- Still open from my earlier comment, all follow-up material: the warning wording, the two wrong doc sentences at `amazon-bedrock.mdx:515`/`:533`, and the `s3Location` gate. Happy to do those as a separate PR.

Mechanically you're also at 1 approval — `2 Approvers for Bots` is red since the new commits dismissed `notowen333`'s — and `Run integration tests` is still in progress on `74f3e57`.

<details>
<summary>Patch (verified locally: ruff + mypy clean, 183 passed)</summary>

```diff
--- a/strands-py/src/strands/models/bedrock.py
+++ b/strands-py/src/strands/models/bedrock.py
@@ -14,6 +14,7 @@
 import boto3
 from botocore.config import Config as BotocoreConfig
 from botocore.exceptions import ClientError
+from botocore.model import NoShapeFoundError
 from pydantic import BaseModel
 from typing_extensions import Unpack, override
@@ -474,6 +475,18 @@
                 return idx
         return None
 
+    def _guardrail_image_formats(self) -> list[str]:
+        """Image formats Bedrock guardrails can assess, read from the botocore service model.
+
+        Returns an empty list on botocore versions that predate guardrail image support, where
+        sending a guardContent image block would be rejected client-side.
+        """
+        try:
+            return self.client.meta.service_model.shape_for("GuardrailConverseImageFormat").enum or []
+        except NoShapeFoundError:
+            logger.debug("botocore does not model guardrail image formats | skipping guardContent wrap")
+            return []
+
     def _format_bedrock_messages(self, messages: Messages) -> list[dict[str, Any]]:
@@ -536,10 +549,7 @@
                     elif "image" in formatted_content:
                         image_format = formatted_content["image"].get("format", "")
-                        supported_formats = self.client.meta.service_model.shape_for(
-                            "GuardrailConverseImageFormat"
-                        ).enum
-                        if image_format in supported_formats:
+                        if image_format in self._guardrail_image_formats():
                             formatted_content = {"guardContent": {"image": formatted_content["image"]}}
                         else:
                             logger.warning(
```

Plus the regression test (fails without the guard):

```python
@pytest.mark.asyncio
async def test_format_request_guardrail_image_formats_absent_from_service_model(model):
    """Older botocore versions do not model guardrail image formats; skip the wrap instead of raising."""
    model.client.meta.service_model.shape_for.side_effect = NoShapeFoundError("GuardrailConverseImageFormat")
    model.update_config(guardrail_id="g", guardrail_version="DRAFT", guardrail_latest_message=True)

    messages = [{"role": "user", "content": [{"image": {"format": "png", "source": {"bytes": b"i"}}}]}]
    request = model.format_request(messages)

    assert "guardContent" not in request["messages"][0]["content"][0]
```

</details>

<details>
<summary>Evidence for the botocore floor and the zero-coverage claim</summary>

Shape presence per botocore version (unpacked `botocore/data/bedrock-runtime/*/service-2.json.gz` from the published wheels):

| botocore | `GuardrailConverseImageFormat` | `GuardrailConverseContentBlock` members |
|---|---|---|
| 1.29.0 | no `bedrock-runtime` model at all | — |
| 1.35.0 | **absent** | `text` only |
| 1.36.0 | `[png, jpeg]` | `text`, `image` |
| 1.40.76 (current) | `[png, jpeg]` | `text`, `image` |

Failure mode with the shape absent, on `74f3e57`:

```
>>> m._format_bedrock_messages([{"role": "user", "content": [{"text": "hi"}, gif]}])
botocore.model.NoShapeFoundError: GuardrailConverseImageFormat
```

Coverage: I ran `_format_bedrock_messages` on the integ test's exact message list with a client whose `service_model` is the real botocore one and a counter on `shape_for`. With the trailing `"What do you see?"` turn the counter stays at `[]` and the png at index 2 comes out as a bare `image` block; drop that turn and it's `['GuardrailConverseImageFormat']` with the png wrapped. Also worth knowing: `.enum` on a non-enum string shape returns `[]`, so if AWS ever restructured that shape the current code would silently stop wrapping — the `or []` in the patch just makes that explicit.

</details>

Issue #1671, opened 2026-02-12
[BUG] Bedrock Guardrail False Positive on Tool Results


peter-nagy1:
### Checks

- [x] I have updated to the lastest minor and patch version of Strands
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched [./issues](./issues?q=) and there are no duplicates of my issue

### Strands Version

1.23.0

### Python Version

3.12.12

### Operating System

macOS 26.2

### Installation Method

other

#### Summary
Bedrock Guardrails incorrectly flags tool results as prompt injection attacks on subsequent messages. Tool results are stored in the conversation history with role: "user", and when the guardrail evaluates the conversation on any follow-up message, it misidentifies previous tool results as malicious user input.

**Key Finding:** The issue occurs because tool results are stored with role: "user" in the conversation history. When the guardrail evaluates subsequent requests, it scans all role: "user" messages—including tool results—and flags certain patterns (e.g., "You are Test Admin User.") as prompt injection attacks.

#### Environment
- SDK: AWS Strands Agents
- Model: us.anthropic.claude-haiku-4-5-20251001-v1:0
- Guardrail: Bedrock Guardrail with prompt attack detection enabled

#### Code snippet
```python
"""Simple reproduction of guardrail false positive."""

import os

import boto3
import strands
from dotenv import load_dotenv
from strands import Agent
from strands.models.bedrock import BedrockModel

load_dotenv()

@strands.tool
def identity_tool() -> str:
    """Tool that returns identity-like text."""
    return "You are Test Admin User."

boto_session = boto3.Session()
bedrock_model = BedrockModel(
    boto_session=boto_session,
    model_id=os.getenv("BEDROCK_MODEL_ID_HAIKU_4_5"),
    streaming=False,
    guardrail_id=os.getenv("BEDROCK_GUARDRAIL"),
    guardrail_version="DRAFT",
    guardrail_trace="enabled",
    guardrail_redact_input=False
)

agent = Agent(
    model=bedrock_model,
    tools=[identity_tool],
)

print("--- Step 1 ---")

# Step 1: This triggers identity_tool, returns "You are Test Admin User."
agent("Call the identity tool")

print("\n\n--- Step 2 ---")

# Step 2: This fails with prompt attack detection
# Guardrail sees "You are Test Admin User." in history as role="user"
agent("Hi there")  # <-- FAILS HERE

print("\n\nCONVERSATION:")
for i, msg in enumerate(agent.messages):
[...]

PR #1080, opened 2025-10-23
Fix #1077: properly redact toolResult blocks to avoid corrupting the conversation

leotac:
## Description

See #1077 : output guardrails intervention causes the redacting of the preceding _input_ that might contain a `toolResult`,  breaking the conversation.

This PR introduces the following change:
- "toolResult" blocks are properly redacted; only the toolResult "content" is redacted, keeping the toolUseId,  so we are not breaking the conversation. 

## Related Issues

#1077

## Type of Change

Bug fix

## Testing

Ran unit tests and integration tests

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [x] I have updated the documentation accordingly
- [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #1077, opened 2025-10-23
[BUG] With bedrock guardrails, tool output is redacted breaking the conversation


leotac:
### Checks

- [x] I have updated to the lastest minor and patch version of Strands
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched [./issues](./issues?q=) and there are no duplicates of my issue

### Strands Version

1.13.0

### Python Version

3.12

### Operating System

macOS

### Installation Method

pip

### Steps to Reproduce

The `"test-guardrail-block-cactus"` is the test guardrail used in the integration tests, that blocks input and output if it contains the word CACTUS.

```python
import json
from strands import Agent, tool
from strands.models import BedrockModel
import boto3

def get_guardrail_id(client, guardrail_name):

def get_guardrail():
    client = boto3.client("bedrock", region_name="us-west-2")

    guardrail_name = "test-guardrail-block-cactus"    
    response = client.list_guardrails()
    for guardrail in response.get("guardrails", []):
        if guardrail["name"] == guardrail_name:
            return guardrail["id"]
    return None

@tool
def get_users() -> str:
    "List my users"
    return """|Name|Email|
|Jerry Barry|jerry@gmail.com|
|CACTUS|cactus@email.com|
"""

bedrock_model = BedrockModel(
    model_id="...",
    guardrail_id=get_guardrail(),
    guardrail_version="DRAFT",
    guardrail_redact_input = True, # default
    guardrail_redact_output = False, # default
)

# Create agent with the guardrail-protected model
agent = Agent(
    system_prompt="You are a helpful assistant.",
    model=bedrock_model,
    tools=[get_users]
)

# Use the protected agent for conversations
response = agent("Who are my users?")
assert response.stop_reason == "guardrail_intervened"
print(f"Conversation after first call: {json.dumps(agent.messages, indent=4)}")

response = agent("Hello")
print(f"Entire conversation: {json.dumps(agent.messages, indent=4)}")

```

### Expected Behavior

Bedrock Guardrails do not actually check tool outputs. In this case, the guardrails triggered on the model **output**. 
My expectation would be that the model _output_ is indeed redacted, but:
- the model _input_ should not be redacted if only "output" guardrails triggered
- in any case, the part of the input that contains tool outputs should certainly not be redacted as it breaks the conversation

```
Entire conversation: [
    {
        "role": "user",
        "content": [
            {
                "text": "Who are my users?"
            }
        ]
    },
    {
        "role": "assistant",
[...]

PR #3608, opened 2026-08-03
fix(typescript): correct guardrail redaction

JackYPCOnline:
## Description

Bedrock guardrail interventions did not distinguish whether input or output was blocked. An output-only intervention could therefore rewrite user history, including protocol-level tool results. This change makes redaction source-aware so conversation history is only mutated for input interventions.

## Related Issues

Reported privately; no public issue.

## Documentation PR

Not required; this does not change the public API.

## Type of Change

Bug fix

## Testing

- [x] I ran the applicable repository pre-commit checks

## Checklist

- [x] No documentation changes are required
- [x] No new documentation example is required
- [x] No dependent changes are required

----

### DIFF
--- a/strands-ts/src/agent/__tests__/agent.test.ts
+++ b/strands-ts/src/agent/__tests__/agent.test.ts
@@ -1650,48 +1650,53 @@ describe('Agent._redactLastMessage', () => {
     expect(agent['messages'][agent['messages'].length - 1]!.trackingId).toBe('durable-1')
   })
 
-  it('redacts the latest human-authored user message and preserves trailing tool results', () => {
+  it('preserves tool result blocks with redacted content', () => {
     const model = new MockMessageModel().addTurn({ type: 'textBlock', text: 'Response' })
     const agent = new Agent({ model })
 
-    const userMessage = new Message({
-      role: 'user',
-      content: [new TextBlock('sensitive content')],
-      trackingId: 'durable-1',
-    })
-    const toolUseMessage = new Message({
-      role: 'assistant',
-      content: [
-        new ToolUseBlock({
-          name: 'test_tool',
-          toolUseId: 'tool-1',
-          input: {},
-        }),
-      ],
-    })
-    const toolResultMessage = new Message({
-      role: 'user',
-      content: [
-        new ToolResultBlock({
-          toolUseId: 'tool-1',
-          status: 'success',
-          content: [new TextBlock('tool result content')],
-        }),
-      ],
-    })
-    agent['messages'].push(userMessage, toolUseMessage, toolResultMessage)
+    // Add a user message with tool result and text blocks
+    agent['messages'].push(
+      new Message({
+        role: 'user',
+        content: [
+          new TextBlock('some text'),
+          new ToolResultBlock({
+            toolUseId: 'tool-1',
+            status: 'success',
+            content: [new TextBlock('tool result content')],
+          }),
+          new TextBlock('more text'),
+          new ToolResultBlock({
+            toolUseId: 'tool-2',
+            status: 'error',
+            content: [new TextBlock('error content')],
+          }),
+        ],
+      })
+    )
 
     agent['_redactLastMessage'](redactMessage)
 
-    expect(agent['messages']).toStrictEqual([
-      new Message({
-        role: 'user',
-        content: [new TextBlock(redactMessage)],
-        trackingId: 'durable-1',
-      }),
-      toolUseMessage,
-      toolResultMessage,
-    ])
+    const lastMessage = agent['messages'][agent['messages'].length - 1]!
+    expect(lastMessage.role).toBe('user')
+    expect(lastMessage.content).toHaveLength(2)
+
+    // Only tool result blocks should remain
+    expect(lastMessage.content[0]!.type).toBe('toolResultBlock')
+    expect(lastMessage.content[1]!.type).toBe('toolResultBlock')
+
+    // Tool result blocks should have redacted content but preserve structure
+    const toolResult1 = lastMessage.content[0] as ToolResultBlock
+    expect(toolResult1.toolUseId).toBe('tool-1')
+    expect(toolResult1.status).toBe('success')
+    expect(toolResult1.content).toHaveLength(1)
+    expect((toolResult1.content[0] as TextBlock).text).toBe(redactMessage)
+
+    const toolResult2 = lastMessage.content[1] as ToolResultBlock
+    expect(toolResult2.toolUseId).toBe('tool-2')
+    expect(toolResult2.status).toBe('error')
+    expect(toolResult2.content).toHaveLength(1)
+    expect((toolResult2.content[0] as TextBlock).text).toBe(redactMessage)
   })
 
   it('does not redact when last message is not from user', () => {
--- a/strands-ts/src/agent/__tests__/agent.test.ts
+++ b/strands-ts/src/agent/__tests__/agent.test.ts
@@ -1650,53 +1650,48 @@ describe('Agent._redactLastMessage', () => {
     expect(agent['messages'][agent['messages'].length - 1]!.trackingId).toBe('durable-1')
   })
 
-  it('preserves tool result blocks with redacted content', () => {
+  it('redacts the latest human-authored user message and preserves trailing tool results', () => {
     const model = new MockMessageModel().addTurn({ type: 'textBlock', text: 'Response' })
     const agent = new Agent({ model })
 
-    // Add a user message with tool result and text blocks
-    agent['messages'].push(
-      new Message({
-        role: 'user',
-        content: [
-          new TextBlock('some text'),
-          new ToolResultBlock({
-            toolUseId: 'tool-1',
-            status: 'success',
-            content: [new TextBlock('tool result content')],
-          }),
-          new TextBlock('more text'),
-          new ToolResultBlock({
-            toolUseId: 'tool-2',
-            status: 'error',
-            content: [new TextBlock('error content')],
-          }),
-        ],
-      })
-    )
+    const userMessage = new Message({
+      role: 'user',
+      content: [new TextBlock('sensitive content')],
+      trackingId: 'durable-1',
+    })
+    const toolUseMessage = new Message({
+      role: 'assistant',
+      content: [
+        new ToolUseBlock({
+          name: 'test_tool',
+          toolUseId: 'tool-1',
+          input: {},
+        }),
+      ],
+    })
+    const toolResultMessage = new Message({
+      role: 'user',
+      content: [
+        new ToolResultBlock({
+          toolUseId: 'tool-1',
+          status: 'success',
+          content: [new TextBlock('tool result content')],
+        }),
+      ],
+    })
+    agent['messages'].push(userMessage, toolUseMessage, toolResultMessage)
 
     agent['_redactLastMessage'](redactMessage)
 
-    const lastMessage = agent['messages'][agent['messages'].length - 1]!
-    expect(lastMessage.role).toBe('user')
-    expect(lastMessage.content).toHaveLength(2)
-
-    // Only tool result blocks should remain
-    expect(lastMessage.content[0]!.type).toBe('toolResultBlock')
-    expect(lastMessage.content[1]!.type).toBe('toolResultBlock')
-
-    // Tool result blocks should have redacted content but preserve structure
-    const toolResult1 = lastMessage.content[0] as ToolResultBlock
-    expect(toolResult1.toolUseId).toBe('tool-1')
-    expect(toolResult1.status).toBe('success')
-    expect(toolResult1.content).toHaveLength(1)
-    expect((toolResult1.content[0] as TextBlock).text).toBe(redactMessage)
-
-    const toolResult2 = lastMessage.content[1] as ToolResultBlock
-    expect(toolResult2.toolUseId).toBe('tool-2')
-    expect(toolResult2.status).toBe('error')
-    expect(toolResult2.content).toHaveLength(1)
-    expect((toolResult2.content[0] as TextBlock).text).toBe(redactMessage)
+    expect(agent['messages']).toStrictEqual([
+      new Message({
+        role: 'user',
+        content: [new TextBlock(redactMessage)],
+        trackingId: 'durable-1',
+      }),
+      toolUseMessage,
+      toolResultMessage,
+    ])
   })
 
   it('does not redact when last message is not from user', () => {
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -2273,27 +2273,56 @@ export class Agent implements LocalAgent, InvokableAgent {
   }
 
   /**
-   * Redacts the latest human-authored user message in the conversation history.
-   * Called when guardrails block user input and redaction is enabled. Tool result
-   * messages use the user role for protocol purposes and are skipped.
+   * Redacts the last message in the conversation history.
+   * Called when guardrails block user input and redaction is enabled.
+   *
+   * Follows the redaction strategy:
+   * - If the message contains at least one toolResult block, all toolResult blocks
+   *   are kept with redacted content, and all other blocks are discarded.
+   * - Otherwise, the entire content is replaced with a single text block containing
+   *   the redaction message.
    *
    * @param redactMessage - The redaction message to replace the content with
    */
   private _redactLastMessage(redactMessage: string): void {
-    for (let idx = this.messages.length - 1; idx >= 0; idx--) {
-      const message = this.messages[idx]
-      if (message?.role === 'user' && message.content.some((block) => block.type !== 'toolResultBlock')) {
-        this.messages[idx] = new Message({
+    // Find and redact the last message
+    const lastIndex = this.messages.length - 1
+    if (lastIndex >= 0) {
+      const lastMessage = this.messages[lastIndex]
+      if (lastMessage && lastMessage.role === 'user') {
+        // Collect only tool result blocks with redacted content
+        const redactedContent: ContentBlock[] = []
+        for (const block of lastMessage.content) {
+          if (block.type === 'toolResultBlock') {
+            // Preserve tool result block structure, only redact its content
+            redactedContent.push(
+              new ToolResultBlock({
+                toolUseId: block.toolUseId,
+                status: block.status,
+                content: [new TextBlock(redactMessage)],
+              })
+            )
+          }
+        }
+
+        // If no tool result blocks were found, replace entire content with redaction message
+        if (redactedContent.length === 0) {
+          redactedContent.push(new TextBlock(redactMessage))
+        }
+
+        this.messages[lastIndex] = new Message({
           role: 'user',
-          content: [new TextBlock(redactMessage)],
+          content: redactedContent,
           // Redaction rewrites content but it's the same logical message, so keep its tracking id.
-          trackingId: message.trackingId,
+          trackingId: lastMessage.trackingId,
         })
-        return
+      } else if (lastMessage) {
+        // Unexpected state: redaction requested but last message is not from user
+        logger.warn(
+          `role=<${lastMessage.role}> | received input redaction but last message is not from user | redaction skipped`
+        )
       }
     }
-
-    logger.warn('received input redaction but no human-authored user message was found | redaction skipped')
   }
 
   /**
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -2273,56 +2273,27 @@ export class Agent implements LocalAgent, InvokableAgent {
   }
 
   /**
-   * Redacts the last message in the conversation history.
-   * Called when guardrails block user input and redaction is enabled.
-   *
-   * Follows the redaction strategy:
-   * - If the message contains at least one toolResult block, all toolResult blocks
-   *   are kept with redacted content, and all other blocks are discarded.
-   * - Otherwise, the entire content is replaced with a single text block containing
-   *   the redaction message.
+   * Redacts the latest human-authored user message in the conversation history.
+   * Called when guardrails block user input and redaction is enabled. Tool result
+   * messages use the user role for protocol purposes and are skipped.
    *
    * @param redactMessage - The redaction message to replace the content with
    */
   private _redactLastMessage(redactMessage: string): void {
-    // Find and redact the last message
-    const lastIndex = this.messages.length - 1
-    if (lastIndex >= 0) {
-      const lastMessage = this.messages[lastIndex]
-      if (lastMessage && lastMessage.role === 'user') {
-        // Collect only tool result blocks with redacted content
-        const redactedContent: ContentBlock[] = []
-        for (const block of lastMessage.content) {
-          if (block.type === 'toolResultBlock') {
-            // Preserve tool result block structure, only redact its content
-            redactedContent.push(
-              new ToolResultBlock({
-                toolUseId: block.toolUseId,
-                status: block.status,
-                content: [new TextBlock(redactMessage)],
-              })
-            )
-          }
-        }
-
-        // If no tool result blocks were found, replace entire content with redaction message
-        if (redactedContent.length === 0) {
-          redactedContent.push(new TextBlock(redactMessage))
-        }
-
-        this.messages[lastIndex] = new Message({
+    for (let idx = this.messages.length - 1; idx >= 0; idx--) {
+      const message = this.messages[idx]
+      if (message?.role === 'user' && message.content.some((block) => block.type !== 'toolResultBlock')) {
+        this.messages[idx] = new Message({
           role: 'user',
-          content: redactedContent,
+          content: [new TextBlock(redactMessage)],
           // Redaction rewrites content but it's the same logical message, so keep its tracking id.
-          trackingId: lastMessage.trackingId,
+          trackingId: message.trackingId,
         })
-      } else if (lastMessage) {
-        // Unexpected state: redaction requested but last message is not from user
-        logger.warn(
-          `role=<${lastMessage.role}> | received input redaction but last message is not from user | redaction skipped`
-        )
+        return
       }
     }
+
+    logger.warn('received input redaction but no human-authored user message was found | redaction skipped')
   }
 
   /**
--- a/strands-ts/src/models/bedrock.ts
+++ b/strands-ts/src/models/bedrock.ts
@@ -1748,7 +1748,7 @@ export class BedrockModel extends Model<BedrockModelConfig> {
     const redaction = this._config.guardrailConfig?.redaction
 
     // Default: redact input is true unless explicitly set to false
-    if (redaction?.input !== false) {
+    if (guardrailData.inputAssessment !== undefined && redaction?.input !== false) {
       logger.debug('redacting input due to guardrail')
       events.push({
         type: 'modelRedactionEvent',
@@ -1758,8 +1758,8 @@ export class BedrockModel extends Model<BedrockModelConfig> {
       })
     }
 
-    // Only redact output if explicitly enabled
-    if (redaction?.output) {
+    // Only redact output if it was assessed and output redaction is explicitly enabled
+    if (guardrailData.outputAssessments !== undefined && redaction?.output) {
       logger.debug('redacting output due to guardrail')
       const outputRedactionEvent: ModelStreamEvent = {
         type: 'modelRedactionEvent',
--- a/strands-ts/src/models/bedrock.ts
+++ b/strands-ts/src/models/bedrock.ts
@@ -128,6 +128,26 @@ const DEFAULT_REDACT_INPUT_MESSAGE = '[User input redacted.]'
 const DEFAULT_REDACT_OUTPUT_MESSAGE = '[Assistant output redacted.]'
 
 /**
+ * Recursively checks a guardrail assessment for a policy that detected and blocked content.
+ */
+function hasDetectedBlockedPolicy(value: unknown): boolean {
+  if (Array.isArray(value)) {
+    return value.some(hasDetectedBlockedPolicy)
+  }
+
+  if (value === null || typeof value !== 'object') {
+    return false
+  }
+
+  const node = value as Record<string, unknown>
+  if (node.action === 'BLOCKED' && node.detected === true) {
+    return true
+  }
+
+  return Object.values(node).some(hasDetectedBlockedPolicy)
+}
+
+/**
  * TTL durations accepted by Bedrock for prompt-cache checkpoints.
  *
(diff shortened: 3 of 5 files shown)


JackYPCOnline:
@strandly-the-agent Can you review this RP and justify the correctness?

strands-ts/src/agent/agent.ts
```diff
@@ -2273,56 +2273,27 @@ export class Agent implements LocalAgent, InvokableAgent {
...
-      } else if (lastMessage) {
-        // Unexpected state: redaction requested but last message is not from user
-        logger.warn(
-          `role=<${lastMessage.role}> | received input redaction but last message is not from user | redaction skipped`
-        )
+        return
       }
     }
+
+    logger.warn('received input redaction but no human-authored user message was found | redaction skipped')
```
strandly-the-agent: 🔴 **This reverts #1080 and re-opens #1077.** The backward scan skips pure-`toolResultBlock` messages and then *wholesale-replaces* whatever it lands on, dropping the per-block preserve-and-redact behavior that exists specifically to keep the `toolUse`/`toolResult` pairing intact.

Three symptoms, one root cause. All reproduced with a real `Agent` + real `BedrockModel` (only the HTTP transport stubbed), run on this branch and on `c0776c008`:

**1. The flagged content survives; an innocent message is destroyed instead.** Tool call → guardrail blocks input. Note `_redactLastMessage` is invoked at `agent.ts:1988-1989` *before* the new assistant message is appended, so at redaction time `messages[-1]` is the previous cycle's tool-result message — the scan walks straight past it:

```
this branch                                          base c0776c008
[0] user: "[User input redacted.]"   ← clobbered     [0] user: "Please look up record cust-1…"   ← intact
[1] assistant: toolUse(lookup_record, tu-1)          [1] assistant: toolUse(lookup_record, tu-1)
[2] user: toolResult(tu-1, "…SSN 123-45-6789         [2] user: toolResult(tu-1, "[User input
         …TOOL_SECRET_9f2a")        ← survives                 redacted.]")              ← scrubbed
```

And it persists: with a real `SessionManager` + `FileStorage`, `snapshot_latest.json` contains `TOOL_SECRET_9f2a` on this branch and doesn't on base.

**2. It compounds — the conversation is permanently poisoned.** Because the offending text is never removed, it's resent every turn:

```
this branch: response1/2/3.stopReason = guardrailIntervened, guardrailIntervened, guardrailIntervened
base:        response1/2/3.stopReason = guardrailIntervened, endTurn,             endTurn
```

Each retry destroys another user message (`[0]`, `[4]`, `[6]` all become `[User input redacted.]`). This is exactly the assertion this PR removes from the integ test — `expect(response2.stopReason).not.toBe('guardrailIntervened')` fails on this branch and passes on base — and exactly what `site/src/content/docs/user-guide/safety-security/guardrails.mdx` documents the feature for: *"so that follow-up questions are not also blocked by the same questions."*

**3. Mixed-content messages orphan a `toolUse`** (narrower reach — needs a caller-built message via the public `invoke(ContentBlock[] | Message[])` path, e.g. an external tool executor or HITL relay; no first-party code produces one). `content.some(b => b.type !== 'toolResultBlock')` matches on the text block, then the whole content is replaced, so the `toolResultBlock` goes with it: `toolUse=["tu-A"] toolResult=[] ORPHANED=["tu-A"]` vs base's `ORPHANED=[]`. There's no client-side guard — `_formatMessages` puts it on the wire, and it comes back as `ValidationException: tool_use ids were found without tool_result blocks immediately after: tu-A.`, i.e. #1077 verbatim.

**4. Silent no-op** when nothing qualifies (all user messages are pure tool results — reachable via a session restored mid-tool-call, or an externally-driven tool executor): just the `logger.warn`, nothing redacted. Base redacted the last message.

The minimal safe move is to drop (B) from this PR and keep (A) — that fixes all four at once, since a pure-toolResult last message gets its content redacted **in place** rather than skipped:

```suggestion
    // Find and redact the last message
    const lastIndex = this.messages.length - 1
    if (lastIndex >= 0) {
      const lastMessage = this.messages[lastIndex]
      if (lastMessage && lastMessage.role === 'user') {
        // Collect only tool result blocks with redacted content
        const redactedContent: ContentBlock[] = []
        for (const block of lastMessage.content) {
          if (block.type === 'toolResultBlock') {
            // Preserve tool result block structure, only redact its content
            redactedContent.push(
              new ToolResultBlock({
                toolUseId: block.toolUseId,
                status: block.status,
                content: [new TextBlock(redactMessage)],
              })
            )
          }
        }

        // If no tool result blocks were found, replace entire content with redaction message
        if (redactedContent.length === 0) {
          redactedContent.push(new TextBlock(redactMessage))
        }

        this.messages[lastIndex] = new Message({
          role: 'user',
          content: redactedContent,
          // Redaction rewrites content but it's the same logical message, so keep its tracking id.
          trackingId: lastMessage.trackingId,
        })
      } else if (lastMessage) {
        // Unexpected state: redaction requested but last message is not from user
        logger.warn(
          `role=<${lastMessage.role}> | received input redaction but last message is not from user | redaction skipped`
        )
      }
    }
```

To be fair to the idea behind (B): the old code isn't *right* either — when the prompt is what blocked, it redacts a clean tool result. That's #2992's motivation, and it's a genuinely open problem, because the Bedrock trace carries no prompt-vs-tool-result provenance. But guessing "it must have been the human turn" trades a cosmetic over-redaction for a silent leak plus an unusable conversation, which is the worse of the two. That question wants its own PR, coordinated with #2992 — and ideally provenance rather than a heuristic. Also worth noting the docstring at `agent.ts:2276-2278` describes the disputed behavior, so it'll need updating with whatever lands.

strands-ts/src/models/bedrock.ts
```diff
@@ -1748,7 +1748,7 @@ export class BedrockModel extends Model<BedrockModelConfig> {
     const redaction = this._config.guardrailConfig?.redaction
 
     // Default: redact input is true unless explicitly set to false
-    if (redaction?.input !== false) {
+    if (guardrailData.inputAssessment !== undefined && redaction?.input !== false) {
```
strandly-the-agent: 🔴 **Presence ≠ blocked, so the headline bug survives for the common guardrail setup.** Per the AWS types (`@aws-sdk/client-bedrock-runtime/…/models_0.d.ts:3590,3600,3605`), `inputAssessment?: Record<string, GuardrailAssessment>` and `outputAssessments?: Record<string, GuardrailAssessment[]>` are populated whenever that side was **assessed** — filled with `action: 'NONE', detected: false` when clean. Whether a side actually *blocked* lives inside each policy as `action === 'BLOCKED' && detected === true`, which neither branch here inspects.

Ran the trace-shape matrix (8 cases, all green) against this branch's `_generateRedactionEvents`:

```
H3a  non-blocking inputAssessment + BLOCKING outputAssessments  → inputRedactionFired=true  outputRedactionFired=true
H3b  non-blocking inputAssessment only                          → inputRedactionFired=true
H3c  BLOCKED input only, no outputAssessments key               → input only    ← fix working
H3d  BLOCKED output only, no inputAssessment key                → output only   ← fix working
H4b  assessment keys present but EMPTY maps                     → BOTH fire     ← presence gate passes on {}
```

H3a is the one that matters: one guardrail with policies on *both* directions is the ordinary configuration, and on a pure output block it still fires input redaction and still rewrites history — the exact behavior this PR is titled after. The gate only helps when the guardrail's policies are scoped to a single side so the key never appears at all, which is what your integ test happens to exercise.

The suggestion would be to gate on the policy outcome instead. Python already has the right primitive, it just doesn't apply it per-side — `strands-py/src/strands/models/bedrock.py:1199 _find_detected_and_blocked_policy` — so porting its shape would fix this and close the parity gap in one go:

```ts
// sketch — needs a small recursive helper that walks topicPolicy.topics[],
// contentPolicy.filters[], wordPolicy.*, sensitiveInformationPolicy.*,
// contextualGroundingPolicy.filters[] for any `action === 'BLOCKED' && detected === true`
if (isGuardrailPolicyBlocked(guardrailData.inputAssessment) && redaction?.input !== false) { … }
if (isGuardrailPolicyBlocked(guardrailData.outputAssessments) && redaction?.output) { … }
```

Same applies to `:1762`. If you'd rather keep this PR to the presence check as a narrower first step, that's defensible — but then the PR description overstates it: it's not source-aware in general, only for single-side-scoped guardrails.

strands-ts/src/agent/__tests__/agent.test.ts
```diff
@@ -1650,53 +1650,48 @@ describe('Agent._redactLastMessage', () => {
     expect(agent['messages'][agent['messages'].length - 1]!.trackingId).toBe('durable-1')
   })
 
-  it('preserves tool result blocks with redacted content', () => {
+  it('redacts the latest human-authored user message and preserves trailing tool results', () => {
```
strandly-the-agent: 🟡 **Nothing in this suite can catch the finding above** — which is why it's 288/288 green with the leak live.

All five tests in `describe('Agent._redactLastMessage')` call the private method directly on a hand-built `messages` array; none drives `invoke()`/`stream()`. So none of them sees the ordering that actually occurs in the loop — `_redactLastMessage` runs at `agent.ts:1988-1989` *before* the assistant message is appended, so `messages[-1]` is the previous cycle's tool-result message. That's precisely the shape where the scan walks past the flagged content.

This new test does cover the walk-past-a-trailing-tool-result case, which is a genuine improvement over what it replaced. But the deleted `'preserves tool result blocks with redacted content'` was the only coverage for a message *mixing* block types, and under the new code that fixture would now lose its `toolResultBlock`s entirely rather than keep them redacted — so the deletion removes the one test that would have flagged the orphaning. Was it dropped because the scenario no longer applies, or because it started failing?

Three tests worth adding whichever way (B) resolves:

1. An `invoke()`-level test: prompt → toolUse → toolResult → guardrail input intervention; assert the tool-result content is scrubbed and the original prompt is untouched.
2. A mixed `[ToolResultBlock, TextBlock]` user message: assert the `toolResultBlock` survives with its `toolUseId`, so the `toolUse` isn't orphaned.
3. The `logger.warn` fall-through: a multi-message history where no message qualifies (currently only the trivial single-non-user-message case is covered).

strands-ts/test/integ/models/bedrock.test.ts
```diff
@@ -585,10 +585,10 @@ describe.skipIf(bedrock.skip)('BedrockModel Integration Tests', () => {
 
     describe('Tool Result Redaction', () => {
       it.each(['sync', 'async'] as const)(
-        'properly redacts tool result in %s mode',
+        'does not redact tool results for output intervention in %s mode',
```
strandly-the-agent, line 588: 🟡 Two things about the rewrite here.

**The deleted assertion was the feature's end-to-end proof.** Removing the `agent.invoke('Thank you!')` turn takes out `expect(response2.stopReason).not.toBe('guardrailIntervened')`, which is the one check that the conversation is still *usable* after redaction — the stated purpose in `site/src/content/docs/user-guide/safety-security/guardrails.mdx`. It fails on this branch and passes on `c0776c008`, so it's load-bearing rather than incidental. The sibling tests at ~430, ~470 and ~523 still cover this for plain-text interventions; this was the only one covering the tool-result path.

**The new shape can't easily get it back.** The system prompt at `:616-617` is unconditional (`'…respond with only CACTUS'`), so every subsequent turn would keep saying CACTUS and keep re-triggering. The siblings use the conditional `'…say CACTUS. Otherwise, respond normally.'` pattern plus `temperature: 0` (`:514`, `:560`) specifically to avoid the flakiness class this repo already fixed twice (`f87925b93`, `bfea64c38`). Worth adopting both here — `temperature: 0` is missing regardless.

Separately, the retitle and the `toStrictEqual([new TextBlock(TOOL_RESULT)])` assertion are right for (A): an output intervention genuinely shouldn't touch the tool result, and the old expectation that it became `INPUT_REDACT_MESSAGE` was encoding the bug. No objection to that part.

strandly-the-agent (changes requested):
**Verdict: changes requested** — pre-merge feedback, since this is a draft. The `bedrock.ts` half is the right idea. The `agent.ts` half re-opens #1077: I can reproduce it leaving guardrail-blocked content live in history while destroying an unrelated message instead.

**Verified** at `9ba3079d8` vs base `c0776c008`: ✅ `agent.test.ts` + `bedrock.test.ts` **288/288 pass**, ✅ `tsc --noEmit -p src/tsconfig.json` clean. 🔴 Repros run on **both** branches (real `Agent` + real `BedrockModel`, only the HTTP transport stubbed): blocked tool-result content **survives** on this branch and is scrubbed on base; the conversation then re-blocks on **every** following turn (`guardrailIntervened` ×3 vs base's 1-then-clean), destroying another user message each time; and a present-but-non-blocking `inputAssessment` still fires input redaction, so (A)'s headline bug survives for the common both-sides guardrail.

**What this PR gets right** — you asked for the correctness justification, so plainly: (A)'s diagnosis is correct and needed. An output-only intervention shouldn't rewrite input history, and before this it did unconditionally. The two `bedrock.test.ts` hunks at ~3496/~3533 fix genuinely mislabeled tests that fed an `inputAssessment`-only trace while claiming to assert output redaction — they'd have passed with output redaction completely broken. And dropping the old integ assertion that expected tool-result content to become `INPUT_REDACT_MESSAGE` on an *output* intervention is correct: that assertion was encoding the bug (A) fixes. None of that is in dispute.

**Open:** 2 🔴 (one per half of the diff) and 2 🟡 test gaps — inline. Plus the scope question below, which I think is the most useful thing here.

### Questions

- **(blocking)** (A) and (B) fix two different bugs. Once (A) suppresses input redaction for output-only interventions, `_redactLastMessage` isn't reached at all for them — so what is (B) buying? It changes target selection for *genuine input* interventions, which is #2992's problem, not this PR's stated one. Would it be simpler to land (A) alone here and take (B) separately?
- **(blocking)** #2992 rewrites this exact method with the same backward-scan design (it's open, at `REVIEW_REQUIRED`, with an unresolved blocking review from me on the same design). Neither PR references the other. Worth reconciling before either lands, so we don't ship two different answers to one question?
- **(non-blocking)** "Reported privately; no public issue" — #1077 is public and closed for what looks like the same defect class, and `CONTRIBUTING.md:231-232` reserves private reporting for security vulnerabilities. Would linking #1077/#1080 help traceability? (Distinct from open issue #1671, which is a different problem.)
- **(non-blocking)** Root `AGENTS.md:46-59` asks for cross-language behavior parity, and `strands-py/src/strands/models/bedrock.py:807-831` has the same class of gap — now filed as #3613. Is a Python follow-up planned here, or worth an explicit deferral note in the body like #2992 did?
- **(non-blocking)** The checklist has "reviewed and understand every line" unchecked while "added any necessary tests that prove my fix is effective" is checked, on a net −54-line diff that deletes a unit test and the integ suite's usability assertion. Worth a second look before this leaves draft?

<details>
<summary><b>Appendix — non-blocking (5)</b></summary>

Two things I hit while reviewing turned out to be pre-existing rather than yours, so they're filed separately instead of cluttering this PR — nothing for you to do about either:

- ⚪ `trace: 'disabled'` skips `_generateRedactionEvents` entirely (`bedrock.ts:1377`/`:1564` gate on the response carrying a trace), so blocked content is never redacted at all in that config — pre-existing, filed #3612.
- ⚪ `strands-py/src/strands/models/bedrock.py:807-831` fires both configured sides on any block despite `_find_detected_and_blocked_policy` (`:1199`) already computing the per-side answer — the Python twin of finding 2 — pre-existing, filed #3613.

And three notes that don't need their own issue:

- ⚪ `strands-ts/src/models/bedrock.ts:162-172` — `BedrockGuardrailRedactionConfig.input?: boolean // Redact input when blocked` was inaccurate before this PR (it redacted on *any* intervention) and is still inaccurate after (it redacts when *assessed*). Since this PR changes those semantics, worth a one-line fix here rather than a separate ticket.
- ⚪ Mutation check: reversing the scan direction at `agent.ts:2283` (earliest-first instead of latest-first) leaves all 114 `agent.test.ts` tests green — no test has two qualifying candidate messages, so "latest" is currently unproven by the suite. Folded into finding 3.
- ⚪ Negative result worth recording: an orphaned `toolUse` does **not** trip the compression helpers — `adjustSplitPointForToolPairs`/`findValidTrimPoint` (`src/agent/conversation-manager/compression/context-compression.ts:37-80`) walk past it silently rather than throwing. So there is no client-side guard anywhere; it only fails at the API.

</details>

<details>
<summary><b>How this was reviewed</b></summary>

Four independent fresh-context passes (correctness/safety, adversarial repro, test quality, scope/issue-alignment) over a context pack, then de-duped and reachability-tiered. Repro scripts were written in-sandbox against the real `Agent`/`BedrockModel` with only the HTTP transport stubbed, and each was run on **both** `9ba3079d8` and `c0776c008` for contrast. I personally re-ran the mainline leak, the re-block loop, the mixed-content orphan and the 8-case trace-shape matrix on both branches before posting; the session-persistence and `logger.warn` fall-through repros come from the adversarial pass. Nothing here was run against live Bedrock — no AWS credentials — so which side a *real* guardrail blames in the finding-1 shape is asserted by the stub, not observed; that one needs the live integ suite to confirm. Solid work for a human to approve, not a gate.

</details>

JackYPCOnline:
@strandly-the-agent what about this version?

strands-ts/src/models/bedrock.ts
```diff
@@ -127,6 +127,26 @@ const DEFAULT_REDACT_INPUT_MESSAGE = '[User input redacted.]'
...
+  if (Array.isArray(value)) {
+    return value.some(hasDetectedBlockedPolicy)
+  }
+
+  if (value === null || typeof value !== 'object') {
+    return false
+  }
+
+  const node = value as Record<string, unknown>
+  if (node.action === 'BLOCKED' && node.detected === true) {
```
strandly-the-agent, line 143: 🔴 **Fail-open: `detected` is optional in the API contract, so this can silently skip redaction on a real block.**

AWS's own API reference for [`GuardrailTopic`](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_GuardrailTopic.html) marks the two fields asymmetrically — I checked the page directly:

- `action` — **Required: Yes**
- `detected` — **Required: No**

Same pattern in the installed SDK types (`@aws-sdk/client-bedrock-runtime@3.1078.0`): `action` is required and `detected?: boolean | undefined` on *every* filter type — `GuardrailTopic` (`action:1132`, `detected?:1137`), `GuardrailContentFilter` (`:860`, `:865`), and likewise custom/managed words, PII, regex and contextual grounding. And AWS's documented Converse trace for a guardrail that actually **blocked** omits it entirely:

```json
"topicPolicy": { "topics": [ { "name": "Heavy metal", "type": "DENY", "action": "BLOCKED" } ] }
```

Reproduced through the real code path (real `BedrockModel`, only `_client.send` stubbed, so `_generateRedactionEvents` → `Model.streamAggregated` → `Agent._redactLastMessage` all execute), feeding that AWS shape verbatim:

```
1. AWS detect-mode shape (action NONE + detected true)          -> NONE   ← correct, keep this
2. AWS-documented BLOCKED topicPolicy, VERBATIM (no `detected`) -> NONE   ← the bug
3. Same, with `detected: true` added                            -> INPUT("[User input redacted.]")
```

On line 2 the guardrail blocked, `stopReason` is `guardrailIntervened`, and **zero** redaction events are emitted — so the blocked content stays in `agent.messages`, gets persisted by any `SessionManager`, and is resent on every following turn. Main never had this failure mode: it redacted on any `guardrail_intervened`, unconditionally. Same result on the non-streaming gate at `:1401`.

And nothing would catch a regression here — I mutation-tested the suite: **dropping this `detected` clause entirely still leaves 177/177 green**, because no fixture pairs `action: 'BLOCKED'` with `detected: false`, and none omits `detected`.

```suggestion
  if (node.action === 'BLOCKED' && node.detected !== false) {
```

Verified in a patched worktree: this makes case 2 redact, keeps case 1 (detect-mode) correctly *not* redacting — the over-redaction you're fixing is keyed on `action: 'NONE'`, never on `detected`, so relaxing this clause can't reopen it — and leaves **291/291 tests green**. Worth adding the AWS-doc shape above as a fixture too, so the clause is actually pinned either way.

**On the severity, so you can push back on it fairly:** this is a *contract* finding, not an observed live failure. AWS says the field is optional and AWS's own example omits it, but every live trace we have evidence for in this repo is wordPolicy — the `enabled_full` capture in `db671ba0e` (#1072) and all the CACTUS integ tests — and `detected` *is* present there. So I can't tell you how often this fires for topic/content/PII blocks specifically, and I have no AWS credentials to find out. I'm still calling it blocking because the failure is silent (nothing throws, nothing logs), the consequence is retained-and-persisted blocked content, and the fix costs one token with no test churn. If you confirm against a live topic-policy guardrail that `detected` is always present, this drops to hardening and I'd have no objection to shipping as-is.

Python has the identical gate at `strands-py/src/strands/models/bedrock.py:1212`, but that's not yours to fix here — pre-existing, filed #3617. Worth settling on one predicate across both SDKs so they don't converge on a shared gap.

strands-ts/test/integ/models/bedrock.test.ts
```diff
@@ -598,77 +598,48 @@ describe.skipIf(bedrock.skip)('BedrockModel Integration Tests', () => {
...
-           * 5 (assistant): response2
-           *
-           * Guardrail intervened on output in message 3 will cause
-           * the redaction of the preceding input (message 2) and message 3.
-           */
+          const response2 = await agent.invoke('What is 2+2? Reply with only the number.')
 
           expect(response1.stopReason).toBe('guardrailIntervened')
+          expect(response2.stopReason).not.toBe('guardrailIntervened')
+          expect(response2.toString()).not.toContain(OUTPUT_REDACT_MESSAGE)
```
strandly-the-agent, line 628: 🟡 These two assertions are now unconditional across both modes, but the sibling test two `describe` blocks up deliberately guards them for `async`.

`'blocks output with redaction in %s mode with trace=%s'` (`:499`) has the same shape as this one — CACTUS in the system prompt, output redaction, `it.each` over sync/async, `temperature: 0` — and at `:529-541` it splits:

```ts
if (processingMode === 'sync') {
  expect(response1.toString()).toContain(REDACT_MESSAGE)
  expect(response2.stopReason).not.toBe('guardrailIntervened')
  expect(response2.toString()).not.toContain(REDACT_MESSAGE)
} else {
  // In async mode, either:
  // - CACTUS was returned and blocked by input guardrail on next turn, or
  // - CACTUS was blocked in response1, allowing normal response2
  const cactusCaughtByInputGuardrail = response2.toString().includes(BLOCKED_INPUT)
  ...
}
```

The reason is real: in async mode the output-guardrail processing can lag the returned response, so unredacted CACTUS can still be in `agent.messages` when turn 2 fires — and since this test also sets `redaction: { input: true, ... }`, that leftover gets caught by the *input* guardrail, making `response2.stopReason === 'guardrailIntervened'` a legitimate async outcome rather than a failure. That's the flakiness class this repo already fixed twice (`f87925b93`, `bfea64c38`), and there's no TS equivalent of Python's `retry_on_flaky` to fall back on.

```suggestion
          if (processingMode === 'sync') {
            expect(response2.stopReason).not.toBe('guardrailIntervened')
            expect(response2.toString()).not.toContain(OUTPUT_REDACT_MESSAGE)
          } else {
            // In async mode the output guardrail can lag, so leftover CACTUS may be
            // caught by the input guardrail on the next turn -- also a valid outcome.
            const cactusCaughtByInputGuardrail = response2.toString().includes(BLOCKED_INPUT)
            const cactusBlockedAllowsNextResponse =
              !response2.toString().includes(OUTPUT_REDACT_MESSAGE) && response2.stopReason !== 'guardrailIntervened'
            expect(cactusCaughtByInputGuardrail || cactusBlockedAllowsNextResponse).toBe(true)
          }
```

That leaves two `if (processingMode === 'sync')` blocks back to back — you may prefer to fold this into the existing one at `:630` instead. `expect(response1.stopReason)` at `:626` is fine unconditional.

Inferred from the sibling's existing defense, not observed live — I can't run the integ suite here (no AWS credentials). Everything else in the integ rework looks right: the `response2` usability assertion is back and correctly hoisted out of the sync-only branch, the system prompt is conditional now, and `temperature: 0` is on the model config in the same spot as its siblings.

strandly-the-agent (changes requested):
**Verdict: changes requested** — but this is "one thing left", not "still broken". Both blockers from the last round are properly fixed, and the remaining item has a one-token answer I've already verified.

**Verified** at `3dd143b97` vs base `c0776c008` (prior rev `9ba3079d8`):
✅ `strands-ts/src/agent/agent.ts` diff vs main is **empty** — the `_redactLastMessage` rewrite is fully reverted. ✅ **291/291** unit tests pass (`bedrock.test.ts` 177 + `agent.test.ts` 114). ✅ `tsc --noEmit -p src/tsconfig.json` clean. ✅ Three-way repro (main / rev1 / this head) confirms the tool-result secret is scrubbed again, the re-block loop is gone (`guardrailIntervened, endTurn, endTurn` — matching main, vs rev1's ×3), and a mixed `[ToolResultBlock, TextBlock]` message keeps its `toolResultBlock` so no `toolUse` is orphaned.

**Resolved from the last round** — not re-flagging any of it: both 🔴 (the `agent.ts` backward scan; the presence-vs-blocked gate) and both 🟡 (the `agent.test.ts` coverage gap is moot on the revert; the integ test got its `response2` usability assertion back, a conditional system prompt, and `temperature: 0`). GitHub has marked those threads outdated — I don't have permission to resolve threads on this repo, so please close them out when convenient.

**What's genuinely right now:** `hasDetectedBlockedPolicy` is the correct shape. I re-ran my 8-case trace matrix plus 19 more shapes: cross-side isolation holds (an output block can no longer fire input redaction — that was the whole point), the `.some()` recursion correctly handles `trace: 'enabled_full'` assessments that mix detected and non-detected filters, a blocked filter last in the array, a block under a later key, and multi-guardrail-id. AWS **detect-mode** (`action: 'NONE'` + `detected: true`) is correctly *not* treated as a block. And across the full behavior-delta matrix vs main, every case where this head redacts less than main is main's over-redaction being correctly removed — with exactly one exception, inline.

**The one thing left:** `bedrock.ts:143` requires `detected === true`, but AWS marks that field optional. Details and a verified one-token fix inline.

### Questions

- **(would settle the blocker)** Do you have a live guardrail with a **topic**, content-filter or PII policy — not just the wordPolicy/CACTUS one — to check empirically whether `detected` is ever actually absent on a real BLOCKED trace? Every live trace we have in this repo is wordPolicy, where it's present. One `converse` call either confirms the blocker or drops it to hardening.
- **(non-blocking)** Worth a follow-up to extend the guardrail integ suite past wordPolicy/CACTUS? As it stands it structurally can't cover the policy types this finding is about.
- **(non-blocking)** Longer term, is `detected !== false` the intended shape, or would you rather drop the `detected` clause entirely (`action === 'BLOCKED'` alone)? The action enums are closed (`BLOCKED`/`NONE`, plus `ANONYMIZED` for PII and regex), so `detected` arguably adds no signal on the BLOCKED branch — but that's a bigger, currently-untested bet, so I'd keep this PR to the minimal fix.
- **(non-blocking, unchanged from last round)** The PR body still says "Reported privately; no public issue" and is silent on Python. #1077/#1080 are the public prior art for the part you reverted, and the Python-side gaps are now filed as #3613, #3612 and #3617. Worth linking so the parity work doesn't get lost.

<details>

Pre-existing, not this PR's problem — filed separately so it doesn't clutter this thread, nothing for you to do:

- ⚪ Python's `_find_detected_and_blocked_policy` has the identical optional-`detected` fail-open at `strands-py/src/strands/models/bedrock.py:1212`, and it's the leaf predicate the inline 🔴 is about. Untouched by this PR (`git diff --name-only c0776c008..HEAD` shows no `strands-py` files), reproduced end to end through a real `Agent` + mocked boto3 client — pre-existing, filed #3617.

And five notes on this diff that don't need their own issue:

- ⚪ **Fixture shape drift.** `outputAssessments` is `Record<string, GuardrailAssessment[]>` (array per key) but three older fixtures use a bare object — `bedrock.test.ts:3569`, `:3606`, `:3645`. Your new one at `:3286` uses the correct array form. Verified the helper is shape-agnostic and that no production code indexes it as an array (`grep` finds only the two gate lines), so nothing is masked — it just teaches the wrong shape to the next reader.
- ⚪ **Two mutation-surviving branches.** Dropping the `value === null` guard at `bedrock.ts:138` leaves 177/177 green — no fixture nests an explicit `null`, and `node.action` on `null` would throw. And no fixture exercises a multi-element or `enabled_full`-style mixed `detected: true`/`false` `GuardrailAssessment[]`, even though that's the exact class `db671ba0e` (#1072) had to fix for Python's walker. (I confirmed by hand that the recursion *does* handle it — it's just untested.)
- ⚪ **Streaming-only coverage for the new cases.** The gate is duplicated at `bedrock.ts:1401` for the non-streaming path; one older test covers it, none of the new per-side `it.each` cases do. Verified manually that it behaves identically — including the fail-open.
- ⚪ **Stale doc comment.** `_generateRedactionEvents` at `bedrock.ts:1761` still says "based on guardrail configuration", which no longer captures that assessment *content* is now a gate too. The two `logger.debug` lines and your new inline comments I checked and they're accurate.
- ⚪ **Recursion safety** (contrived, no action): `undefined`/`null`/`[]`/`{}`/numbers/`Date`/`Map`/`Set` all return false safely; 50k guardrail ids scan in 46ms. A cyclic object or >~1000-deep nesting throws `RangeError`, but Bedrock traces are acyclic JSON about five levels deep.

Helper placement is fine, for the record — module-private top-level matches this file's existing pattern (`snakeToCamel`, `withDefaultRequestTimeout`, `applyApiKey`), and `unknown` over `any` is the right call for walking an untyped nested structure.

</details>

<details>
<summary><b>How this was reviewed</b></summary>

Routing for this revision: the diff is 3 files, all under `strands-ts/src/models/` plus tests, so I ran **correctness/safety**, **adversarial repro** and **test quality** as independent fresh-context passes, then a separate **aggregation** pass whose explicit brief was to *challenge* my severity call on the blocker rather than confirm it. Skipped: **API/DevX** (the new helper is module-private, no public surface changed — no `needs-api-review` label required), **docs accuracy** (no docs or docstrings in the diff beyond two accurate inline comments), **LLM-context** (no tool descriptions, prompts or model-facing text touched), and **issue-alignment** (the scope question from the last round was answered by the revert itself; the residual body/linking points are carried as a Question above rather than a pass).

Repros run against a real `Agent`/`BedrockModel` with only the HTTP transport stubbed, executed across three worktrees (main / rev1 / this head) plus a fourth with the proposed fix applied. I re-ran the fail-open repro, the fix verification, the 291-test suite and the type-check myself, and fetched the AWS API reference directly rather than relying on a pass's paraphrase. No live Bedrock — no AWS credentials — which is precisely the limit named in the blocker. Solid work for a human to approve, not a gate.

</details>

PR #3611, opened 2026-08-03
chore: skipped mcp otel instrumentation for mcp v2

poshinchen:
## Description

The MCP client injected OpenTelemetry trace context into tool calls by monkey-patching `mcp.shared.session.BaseSession.send_request` with `wrapt`. That module is private and gone in `mcp` 2.0, so the patch is the first breakage when we adopt the new major version for MCP spec 2026-07-28 support (#1659). The official SDK exposes a public `meta` parameter on `call_tool` that does the same job.

Client-side injection is now a plain function feeding that `meta` parameter. Propagator errors are logged and swallowed so telemetry can never fail on a tool call. The two server-side patches (trace extraction for in-process MCP servers) still monkey-patch `mcp` internals, so they only apply on 1.x where those internals exist. Applying them now takes a lock: the version probe does I/O, and the old unlocked check let concurrent clients stack duplicate wrappers.

## Related Issues

#1659

## Documentation PR

N/A

## Type of Change

Other (please describe): internal refactor, no public API change

## Testing

Verified against a real stdio MCP server with a real `TracerProvider`, no mocks: the server receives a `traceparent` matching the client span, caller-supplied `meta` survives, a raising propagator logs a warning while the call still succeeds, and 16 threads racing `mcp_instrumentation()` apply the patches exactly once.

## Checklist

----

### DIFF
--- a/strands-py/src/strands/tools/mcp/mcp_client.py
+++ b/strands-py/src/strands/tools/mcp/mcp_client.py
@@ -61,7 +61,7 @@ from ...types.media import ImageFormat
 from ...types.tools import AgentTool, ToolResultContent, ToolResultStatus
 from ..tool_provider import ToolProvider
 from .mcp_agent_tool import MCPAgentTool
-from .mcp_instrumentation import mcp_instrumentation
+from .mcp_instrumentation import inject_trace_context, mcp_instrumentation
 from .mcp_tasks import DEFAULT_TASK_CONFIG, DEFAULT_TASK_POLL_TIMEOUT, DEFAULT_TASK_TTL, TasksConfig
 from .mcp_types import MCPToolResult, MCPTransport
 
@@ -775,6 +775,19 @@ class MCPClient(ToolProvider):
         """
         use_task = self._should_use_task(name)
         effective_callback = progress_callback if progress_callback is not None else self._progress_callback
+
+        # Inject once, before branching, so both the task-augmented and direct
+        # call paths below carry the same enriched meta. This is safe on the
+        # caller's thread: asyncio.run_coroutine_threadsafe, used to hand the
+        # resulting coroutine to the background loop, copies the caller's
+        # contextvars, so the caller's active span is still current when the
+        # coroutine runs on the background thread.
+        #
+        # Under mcp 2.x this injection becomes redundant rather than wrong: the
+        # dispatcher injects its own traceparent, overwriting this one with its
+        # own CLIENT span. That span is parented to the caller's span, so the
+        # trace stays connected.
+        meta = inject_trace_context(meta)
 
         if use_task:
             self._log_debug_with_thread("tool=<%s> | using task-augmented execution", name)
--- a/strands-py/src/strands/tools/mcp/mcp_instrumentation.py
+++ b/strands-py/src/strands/tools/mcp/mcp_instrumentation.py
@@ -1,17 +1,27 @@
 """OpenTelemetry instrumentation for Model Context Protocol (MCP) tracing.
 
-Enables distributed tracing across MCP client-server boundaries by injecting
-OpenTelemetry context into MCP request metadata (_meta field) and extracting
-it on the server side, creating unified traces that span from agent calls
-through MCP tool executions.
+Enables distributed tracing across MCP client-server boundaries. The client
+side is handled with `inject_trace_context`, which `MCPClient` uses to merge
+the current OpenTelemetry context into the `_meta` field of outgoing tool
+calls through the public `meta` parameter of the official `mcp` package.
+
+The server side covers MCP servers hosted in the same process. It extracts
+client-injected context from incoming messages so server-side spans join the
+client's trace. This requires patching private internals of the `mcp` package
+and those internals are only stable within the 1.x line, so the patches are
+gated to `mcp` 1.x. The `mcp` 2.x line propagates trace context natively.
 
 Based on: https://github.com/traceloop/openllmetry/tree/main/packages/opentelemetry-instrumentation-mcp
 Related issue: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/246
 """
 
+import logging
+import threading
 from collections.abc import AsyncGenerator, Callable
 from contextlib import _AsyncGeneratorContextManager, asynccontextmanager
 from dataclasses import dataclass
+from importlib.metadata import PackageNotFoundError
+from importlib.metadata import version as _package_version
 from typing import Any
 
 from mcp.shared.message import SessionMessage
@@ -19,8 +29,53 @@ from mcp.types import JSONRPCMessage, JSONRPCRequest
 from opentelemetry import context, propagate
 from wrapt import ObjectProxy, register_post_import_hook, wrap_function_wrapper
 
-# Module-level flag to ensure instrumentation is applied only once
+logger = logging.getLogger(__name__)
+
+# Module-level flag to ensure instrumentation is applied only once. The lock
+# makes the check-and-set atomic: `_is_mcp_v1` performs GIL-releasing I/O, so
+# without it concurrent MCPClient construction could stack duplicate wrappers.
 _instrumentation_applied = False
+_instrumentation_lock = threading.Lock()
+
+
+def inject_trace_context(meta: dict[str, Any] | None) -> dict[str, Any] | None:
+    """Merge the current OpenTelemetry context into MCP request metadata.
+
+    Injects the active trace context (for example `traceparent` and
+    `tracestate`) into a copy of `meta` so it can be sent as the `_meta`
+    field of an MCP request. This enables server-side context extraction and
+    trace continuation across the client-server boundary.
+
+    Telemetry must never fail the tool call itself, so propagator errors are
+    logged and swallowed; the caller's metadata is still returned.
+
+    Args:
+        meta: Existing request metadata, or None. The input is not mutated.
+
+    Returns:
+        A new dict containing the caller's entries plus the injected trace
+        context, or None when there is no metadata to send.
+    """
+    carrier: dict[str, Any] = dict(meta) if meta else {}
+    try:
+        propagate.get_global_textmap().inject(carrier)
+    except Exception:
+        logger.warning("failed to inject trace context into mcp request meta", exc_info=True)
+    return carrier or None
+
+
+def _is_mcp_v1() -> bool:
+    """Report whether the installed `mcp` package is on the 1.x line.
+
+    The server-side patches wrap private `mcp` internals that are only stable
+    within 1.x. When the version cannot be determined, the patches are skipped
+    rather than applied to an unknown internal surface.
+    """
+    try:
+        major_version = int(_package_version("mcp").split(".")[0])
+    except (PackageNotFoundError, ValueError):
+        return False
+    return major_version == 1
 
 
 @dataclass(slots=True, frozen=True)
@@ -41,79 +96,36 @@ class ItemWithContext:
 
 
 def mcp_instrumentation() -> None:
-    """Apply OpenTelemetry instrumentation patches to MCP components.
+    """Apply OpenTelemetry instrumentation patches for in-process MCP servers.
 
-    This function instruments three key areas of MCP communication:
-    1. Client-side: Injects tracing context into tool call requests
-    2. Transport-level: Extracts context from incoming messages
-    3. Session-level: Manages bidirectional context flow
+    Instruments two areas of server-side MCP communication:
+    1. Transport-level: Extracts context from incoming messages
+    2. Session-level: Preserves context across async message processing
+       boundaries
 
-    The patches enable distributed tracing by:
-    - Adding OpenTelemetry context to the _meta field of MCP requests
-    - Extracting and activating context on the server side
-    - Preserving context across async message processing boundaries
+    Together the patches let an MCP server hosted in the same process join
+    the trace of the client that injected context into the request's `_meta`
+    field. Client-side injection does not require patching; `MCPClient`
+    injects context via `inject_trace_context`.
+
+    The patches wrap private internals of the `mcp` package that are only
+    stable within the 1.x line, so they are applied only when `mcp` 1.x is
+    installed. The `mcp` 2.x line propagates trace context natively.
 
     This function is idempotent - multiple calls will not accumulate wrappers.
     """
     global _instrumentation_applied
 
-    # Return early if instrumentation has already been applied
-    if _instrumentation_applied:
-        return
+    with _instrumentation_lock:
+        # Return early if instrumentation has already been applied
+        if _instrumentation_applied:
+            return
+        # Set before the version probe: it leaves the GIL, and a concurrent
+        # caller checking under the lock must see the flag.
+        _instrumentation_applied = True
 
-    def patch_mcp_client(wrapped: Callable[..., Any], instance: Any, args: Any, kwargs: Any) -> Any:
-        """Patch MCP client to inject OpenTelemetry context into tool calls.
-
-        Intercepts outgoing MCP requests and injects the current OpenTelemetry
-        context into the request's _meta field for tools/call methods. This
-        enables server-side context extraction and trace continuation.
-
-        Args:
-            wrapped: The original function being wrapped
-            instance: The instance the method is being called on
-            args: Positional arguments to the wrapped function
-            kwargs: Keyword arguments to the wrapped function
-
-        Returns:
-            Result of the wrapped function call
-        """
-        if len(args) < 1:
-            return wrapped(*args, **kwargs)
-
-        request = args[0]
-        method = getattr(request.root, "method", None)
-
-        if method != "tools/call":
-            return wrapped(*args, **kwargs)
-
-        try:
-            if hasattr(request.root, "params") and request.root.params:
-                # Handle Pydantic models
-                if hasattr(request.root.params, "model_dump") and hasattr(request.root.params, "model_validate"):
-                    params_dict = request.root.params.model_dump(by_alias=True)
-                    # Add _meta with tracing context
-                    meta = params_dict.get("_meta") if params_dict.get("_meta") is not None else {}
-                    params_dict["_meta"] = meta
-                    propagate.get_global_textmap().inject(meta)
-
-                    # Recreate the Pydantic model with the updated data
-                    # This preserves the original model type and avoids serialization warnings
-                    params_class = type(request.root.params)
-                    try:
-                        request.root.params = params_class.model_validate(params_dict)
-                    except Exception:
-                        # Fallback to dict if model recreation fails
-                        request.root.params = params_dict
-
-                elif isinstance(request.root.params, dict):
-                    # Handle dict params directly
-                    meta = request.root.params.setdefault("_meta", {})
-                    propagate.get_global_textmap().inject(meta)
-
-            return wrapped(*args, **kwargs)
-
-        except Exception:
-            return wrapped(*args, **kwargs)
+        if not _is_mcp_v1():
+            return
 
     def transport_wrapper() -> Callable[
         [Callable[..., Any], Any, Any, Any], _AsyncGeneratorContextManager[tuple[Any, Any]]
@@ -165,8 +177,6 @@ def mcp_instrumentation() -> None:
         return traced_method
 
     # Apply patches
-    wrap_function_wrapper("mcp.shared.session", "BaseSession.send_request", patch_mcp_client)
-
     register_post_import_hook(
         lambda _: wrap_function_wrapper(
             "mcp.server.streamable_http", "StreamableHTTPServerTransport.connect", transport_wrapper()
--- a/strands-py/tests/strands/tools/mcp/test_mcp_instrumentation.py
+++ b/strands-py/tests/strands/tools/mcp/test_mcp_instrumentation.py
@@ -1,3 +1,6 @@
+import threading
+import time
+from importlib.metadata import PackageNotFoundError
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
@@ -5,12 +8,14 @@ from mcp.shared.message import SessionMessage
 from mcp.types import JSONRPCMessage, JSONRPCRequest
 from opentelemetry import context, propagate
 
+import strands.tools.mcp.mcp_instrumentation as mcp_instrumentation_module
 from strands.tools.mcp.mcp_client import MCPClient
 from strands.tools.mcp.mcp_instrumentation import (
     ItemWithContext,
     SessionContextAttachingReader,
     SessionContextSavingWriter,
     TransportContextExtractingReader,
+    inject_trace_context,
     mcp_instrumentation,
 )
 
@@ -24,6 +29,92 @@ def reset_mcp_instrumentation():
     yield
     # Reset after test too
     mcp_inst._instrumentation_applied = False
+
+
+class TestIsMcpV1:
+    @pytest.mark.parametrize(
+        ("installed_version", "expected"),
+        [
+            ("1.23.0", True),
+            ("1.30.0", True),
+            ("2.0.0", False),
+            ("2.0.0b1", False),
+        ],
+    )
+    def test_version_detection(self, installed_version, expected):
+        """Test that the major version of the installed mcp package is detected correctly."""
+        with patch("strands.tools.mcp.mcp_instrumentation._package_version", return_value=installed_version):
+            assert mcp_instrumentation_module._is_mcp_v1() is expected
+
+    def test_unparseable_version_skips(self):
+        """Test that an unparseable version reports non-1.x so no patches are applied."""
+        with patch("strands.tools.mcp.mcp_instrumentation._package_version", return_value="unknown"):
+            assert mcp_instrumentation_module._is_mcp_v1() is False
+
+    def test_missing_package_skips(self):
+        """Test that a missing mcp distribution reports non-1.x so no patches are applied."""
+        with patch(
+            "strands.tools.mcp.mcp_instrumentation._package_version",
+            side_effect=PackageNotFoundError("mcp"),
+        ):
+            assert mcp_instrumentation_module._is_mcp_v1() is False
+
+
+class TestInjectTraceContext:
+    def test_injects_context_into_empty_meta(self):
+        """Test that trace context is injected when no metadata is supplied."""
+        with patch.object(propagate, "get_global_textmap") as mock_textmap:
+            mock_textmap.return_value.inject = lambda carrier: carrier.update({"traceparent": "00-abc-def-01"})
+
+            result = inject_trace_context(None)
+
+            assert result == {"traceparent": "00-abc-def-01"}
+
+    def test_preserves_existing_meta_entries(self):
+        """Test that caller-supplied metadata survives injection."""
+        with patch.object(propagate, "get_global_textmap") as mock_textmap:
+            mock_textmap.return_value.inject = lambda carrier: carrier.update({"traceparent": "00-abc-def-01"})
+
+            result = inject_trace_context({"com.example/request_id": "abc-123"})
+
+            assert result == {"com.example/request_id": "abc-123", "traceparent": "00-abc-def-01"}
+
+    def test_does_not_mutate_input(self):
+        """Test that the caller's metadata dict is copied, not mutated."""
+        original = {"com.example/request_id": "abc-123"}
+
+        with patch.object(propagate, "get_global_textmap") as mock_textmap:
+            mock_textmap.return_value.inject = lambda carrier: carrier.update({"traceparent": "00-abc-def-01"})
+
+            inject_trace_context(original)
+
+        assert original == {"com.example/request_id": "abc-123"}
+
+    def test_returns_none_when_nothing_to_send(self):
+        """Test that None is returned when there is no metadata and no active context."""
+        with patch.object(propagate, "get_global_textmap") as mock_textmap:
+            mock_textmap.return_value.inject = lambda carrier: None
+
+            result = inject_trace_context(None)
+
+            assert result is None
+
+    def test_propagator_error_does_not_raise(self):
+        """Test that a failing propagator never escapes to fail the tool call.
+
+        Guards https://github.com/strands-agents/harness-sdk/pull/3611#discussion_r3706469408: custom
+        propagators (configurable via OTEL_PROPAGATORS) may raise, and telemetry must not break calls.
+        """
+
+        def broken_inject(carrier):
+            raise RuntimeError("propagator boom")
+
+        with patch.object(propagate, "get_global_textmap") as mock_textmap:
+            mock_textmap.return_value.inject = broken_inject
+
+            result = inject_trace_context({"com.example/request_id": "abc-123"})
+
+            assert result == {"com.example/request_id": "abc-123"}
 
 
 class TestItemWithContext:
@@ -321,24 +412,6 @@ class TestSessionContextAttachingReader:
         assert items[0] == regular_item
 
 
-# Mock Pydantic-like class for testing
-class MockPydanticParams:
-    """Mock class that behaves like a Pydantic model."""
-
-    def __init__(self, **data):
-        self._data = data
-
-    def model_dump(self, by_alias=False):
-        return self._data.copy()
-
-    @classmethod
-    def model_validate(cls, data):
-        return cls(**data)
-
-    def __getattr__(self, name):
-        return self._data.get(name)
-
-
 class TestMCPInstrumentation:
     def test_mcp_instrumentation_called_on_client_init(self):
         """Test that mcp_instrumentation is called when MCPClient is initialized."""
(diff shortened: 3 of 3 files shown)


poshinchen:
@strandly-the-agent review the PR

strands-py/src/strands/tools/mcp/mcp_client.py
```diff
@@ -776,6 +776,14 @@ def _create_call_tool_coroutine(
         use_task = self._should_use_task(name)
         effective_callback = progress_callback if progress_callback is not None else self._progress_callback
 
+        # Inject on the caller's thread so the currently active span is captured; the
+        # returned coroutine executes on the background thread where no span is active.
+        # Caution when bumping to mcp 2.x: its dispatcher injects trace context into
```
strandly-the-agent: *Revised: I first posted this against `fdc8af0b` and it carried a ```suggestion``` block that no longer applies — you'd amended to `fd2ac179`, which added the caution at 781-784, so committing it would have left a duplicate `inject_trace_context(meta)` call. Replaced with a plain diff below. Re-verified everything on `fd2ac179`. Sorry for the noise.*

🟡 **Both comment blocks here rest on the same premise, and it doesn't hold: the background thread *does* see the caller's context.**

`asyncio.run_coroutine_threadsafe` → `loop.call_soon_threadsafe` → `asyncio.Handle.__init__` calls `contextvars.copy_context()` **on the caller's thread**, and the Task created inside that handle inherits it. The snapshot at `start()` (`:340-344`) isn't the only one that matters — there's a fresh per-call one.

**Lines 779-780** — "the returned coroutine executes on the background thread where no span is active" is false. Checked two ways: a span created on the background loop parents to the caller's span (`caller d175cd9d…/312e46d351297660` == `seen on bg thread d175cd9d…/312e46d351297660`); and on **`main`**, where the old patch ran inside `send_request` *on the background thread*, a real in-memory MCP server received `traceparent: 00-195319ee…-e2c2dbdf34fd5e26-03` matching the caller's span — negative control (patch neutralised) gave `_meta: None`. So the pre-PR code already captured the right span; this isn't the bug being fixed.

**Lines 781-784** — the caution is half right. The overwrite is real: mcp 2.x's dispatcher does inject its own `traceparent` (`mcp/shared/jsonrpc_dispatcher.py:390` → `mcp/shared/_otel.py:39`). But "so the caller's context must instead be made active around the session call" doesn't follow — it's already active. Measured against **real `mcp==2.0.0`** in a harness matching `MCPClient`'s shape (one bg loop under a `start()`-time snapshot, per-call `run_coroutine_threadsafe`, live `TracerProvider`):

```
strands injected (agent span) : 00-8ca07f0e3df5d906…2d973-2fe48f3075241048-03
traceparent on the wire       : 00-8ca07f0e3df5d906…2d973-2d753a0fb86a56ad-03
SAME TRACE? True    SAME SPAN ID? False

exported spans:
  MCP send tools/call my_tool   span=2d753a0fb86a56ad  parent=2fe48f3075241048   <- the agent span
  agent.tool.my_tool            span=2fe48f3075241048  parent=ROOT
```

mcp's own CLIENT span is created on the background thread and still parents to the caller's agent span. Same trace, one `traceparent` on the wire, finer granularity — so under 2.x this injection just becomes redundant, and no remediation is needed here.

The change itself is still right — deleting a monkeypatch of private `mcp` internals stands on its own, and dropping the spurious `_meta: {}` is a real improvement. It's only the stated *reasons* that will mislead the next reader. The actual reason to inject at this point is structural: both branches below close over `meta`, so injecting once before the split avoids duplicating the call.

Suggested replacement for **779-785** (as a diff rather than a `suggestion` block, since this thread anchors at 779-781 and can't safely rewrite past it):

```python
        # Inject once, before branching, so both the task-augmented and direct
        # call paths below carry the same enriched meta. This is safe on the
        # caller's thread: asyncio.run_coroutine_threadsafe, used to hand the
        # resulting coroutine to the background loop, copies the caller's
        # contextvars, so the caller's active span is still current when the
        # coroutine runs on the background thread.
        #
        # Under mcp 2.x this injection becomes redundant rather than wrong: the
        # dispatcher injects its own traceparent, overwriting this one with its
        # own CLIENT span. That span is parented to the caller's span, so the
        # trace stays connected.
        meta = inject_trace_context(meta)
```

strands-py/src/strands/tools/mcp/mcp_instrumentation.py
```diff
@@ -23,6 +31,40 @@
...
+    Args:
+        meta: Existing request metadata, or None. The input is not mutated.
+
+    Returns:
+        A new dict containing the caller's entries plus the injected trace
+        context, or None when there is no metadata to send.
+    """
+    carrier: dict[str, Any] = dict(meta) if meta else {}
+    propagate.get_global_textmap().inject(carrier)
+    return carrier or None
```
strandly-the-agent, line 64: 🟡 The deleted `patch_mcp_client` wrapped the equivalent inject in `except Exception: return wrapped(*args, **kwargs)` — a deliberate guarantee that telemetry could never fail a tool call (the now-removed `test_patch_mcp_client_handles_exception_gracefully` pinned it). There's no guard here, and `call_tool_sync`'s handler at `mcp_client.py:875` turns the escape into a failed call.

Confirmed with a propagator whose `inject()` raises — reachable via `OTEL_PROPAGATORS`, a documented extension point, though no stock OTel propagator does this:

```
global textmap propagator: BrokenPropagator
call_tool_sync  → status=error  "Tool execution failed: propagator boom"   tool ran on server? False
call_tool_async → status=error  same                                        tool ran on server? False
list_tools_sync : OK, 1 tool(s) — unaffected
total tools/call requests the server received:  this branch = 0   |   main = 2 (status=success)
```

This also cuts against the repo's own convention of never letting telemetry break the call path — `telemetry/tracer.py:308,728` and the several `except Exception` guards in `mcp_client.py`.

```suggestion
    carrier: dict[str, Any] = dict(meta) if meta else {}
    try:
        propagate.get_global_textmap().inject(carrier)
    except Exception:
        logger.warning("failed to inject trace context into MCP request meta", exc_info=True)
    return carrier or None
```

The module has no logger yet, so this needs `import logging` and `logger = logging.getLogger(__name__)` at module level (as in `mcp_client.py:68`) — or drop the log line if you'd rather stay silent.

strands-py/src/strands/tools/mcp/mcp_instrumentation.py
```diff
@@ -61,59 +107,9 @@ def mcp_instrumentation() -> None:
...
-                elif isinstance(request.root.params, dict):
-                    # Handle dict params directly
-                    meta = request.root.params.setdefault("_meta", {})
-                    propagate.get_global_textmap().inject(meta)
-
-            return wrapped(*args, **kwargs)
-
-        except Exception:
-            return wrapped(*args, **kwargs)
+    if not _is_mcp_v1():
```
strandly-the-agent: 🟡 The docstring at `:102` promises *"idempotent - multiple calls will not accumulate wrappers"*. That's now false under concurrency: `_instrumentation_applied` is an unlocked module global, checked at `:107` and set at `:177`, and this line widens the window between them.

The reason it's this PR's problem and not just a pre-existing race: `_is_mcp_v1()` calls `importlib.metadata.version("mcp")`, ~370 µs of filesystem I/O that **releases the GIL**, inside the unguarded section. On `main` the v1 body is pure CPU, so the first thread gets through it within one GIL slice and everyone else sees the flag already set. Measured, constructing N `MCPClient`s concurrently:

```
this branch:  connect=8   ServerSession.__init__=8    (correct: 1)   3/3 runs at N=8
main      :  connect=1   ServerSession.__init__=1                   5/5 runs, N=2..32
```

Fires from N=2. Control test: stub `_package_version` to a constant, same control flow, no I/O → back to 1 layer, 5/5. So it's the I/O in the window, not the window alone.

Scoping it honestly: the accumulation **saturates** (after five bursts of 8 it sits at 9 layers, not 40) and the nesting stays balanced — 27 nested reader/writer pairs still yielded the message intact — so this is bounded overhead, not a leak or corruption. That's why it's 🟡 and not 🔴. It costs N× context extract/attach per message, and only when an in-process MCP server is also running — which is exactly the scenario the new docstring describes. Reachable from any app that constructs ≥2 `MCPClient`s concurrently (thread-pool init, per-request clients).

Fix: a `threading.Lock` around the check-and-set — or set the flag before doing the work, which is what this v2 branch already does correctly and the v1 path doesn't.

strandly-the-agent (commented):
> *Corrected.* I first posted this against `fdc8af0b`, but you'd already amended to **`fd2ac179`** — which added the mcp-2.x caution comment at `mcp_client.py:781-784` and the `TestIsMcpV1` class. I've re-verified everything on `fd2ac179`; all three findings hold. Inline finding 1 has been revised: its original ```suggestion``` block was anchored to the pre-amend lines and would have left a duplicate `inject_trace_context(meta)` call if committed. Details in that thread.

Right call on the core change — routing through the public `meta` param instead of monkeypatching `mcp.shared.session.BaseSession.send_request` is the fix this needed. Three things came out of the passes: two code comments state mechanisms that don't hold, the old blanket `try/except` around injection is gone (a raising propagator now fails the tool call), and `_is_mcp_v1()`'s filesystem I/O turns a dormant race in `mcp_instrumentation()` into one that fires reliably. All three are cheap; none blocks outright.

| Verified on `fd2ac179` | |
|---|---|
| ✅ head `fd2ac179`, base `main` @ `c0776c00` | 3 files, +159/-262 |
| ✅ `pytest tests/strands/tools/mcp/` | **217 passed** (was 211 pre-amend; +6 from `TestIsMcpV1`), `mcp==1.29.0`; `ruff` + `mypy` clean |
| 🔴 concurrent `MCPClient()` at N=8 | 8 wrapt layers vs **1** on `main`, 3/3 runs |
| 🔴 raising OTel propagator | `call_tool_sync`/`_async` → `status=error`, **0** requests reach the server (`main`: 2, success) |
| 🔴 the `:779-780` rationale | false — caller's span *is* visible on the bg thread (2 independent checks) |
| 🔴 the `:781-784` caution's conclusion | overwrite is real, but the trace stays connected — mcp 2.0.0's span parents to the caller's span, so no remediation needed (measured on real `mcp==2.0.0`) |

**Questions** — the first is the one I'd want answered before merge:

1. **(blocking)** Dropping the propagator `try/except` — deliberate (fail loud on a misconfigured propagator) or a side effect of the refactor? Either is defensible, but it makes the body's "no behavior change for direct tool calls" not quite true.
2. **Scope:** new information on the gate's dormancy — `strands.tools.mcp` doesn't even *import* under real mcp 2.0.0 (10 broken names, starting with `mcp_client.py:34`'s `streamablehttp_client` → `streamable_http_client`; `McpError` → `MCPError`; `mcp.shared.session` gone, which kills `ProgressFnT`). So raising the pin (#3541) wouldn't be enough to exercise this gate either. Worth landing the gate with whatever makes the module importable under 2.x instead?
3. The body's task-path claim ("the old patch skipped [task calls] because it only matched `tools/call`") doesn't hold — `call_tool_as_task` also builds a `CallToolRequest` whose `method` is literally `"tools/call"`, so the old patch matched it too. Amend the description?

<details>
<summary>Appendix — 8 non-blocking items (⚪ / pre-existing / description-only)</summary>

**Correctness — real, but none clears the reachability bar:**

- `mcp_instrumentation.py:54-65` — `except (PackageNotFoundError, ValueError)` misses `AttributeError`. `importlib.metadata.version()` returns **`None`** (rather than raising) when METADATA has no `Version:` line, so `None.split(".")` aborts *every* `MCPClient()` construction instead of the graceful skip the docstring promises. One-word fix, but it needs malformed `mcp` dist-info metadata and I couldn't name a production trigger — and the `PackageNotFoundError` branch is dead on 1.x anyway (`mcp/server/fastmcp/__init__.py:10` calls `version("mcp")` at import time, so stripped metadata can't import at all). Contrived. Note the new `TestIsMcpV1` covers the `ValueError` and `PackageNotFoundError` paths but not this one, since it patches `_package_version` with a string or an exception, never `None`.
- `mcp_instrumentation.py:49-51` — `inject_trace_context` silently overwrites a caller-supplied `traceparent`/`tracestate` passed via the public `meta` param. **Pre-existing** (the deleted patch did the same unconditional inject); worth at most a docstring note on precedence.
- `mcp_instrumentation.py:54-65` — `_is_mcp_v1()` also returns `False` for an unparseable *1.x* version string (PEP 440 epoch `2!1.0`, `v1.29.0`), silently disabling server-side tracing with no log line. Fail-safe direction, but a `logger.debug` would make it diagnosable. I ran 19 realistic version strings through it: `2.0.0rc1`, `1.23.0.dev0`, `1.0.0b1`, `2.0.0.post1`, local versions and `0.9` all answer correctly — consistent with what `TestIsMcpV1` now pins.
- Style: the int-split parse diverges from this repo's own precedent at `models/openai_responses.py:39`, which uses `packaging.version.Version`. Functionally equivalent for realistic mcp versions.

**Description accuracy:** `main` sent a spurious `_meta: {}` when no span was active; this branch correctly sends no `_meta` key at all. A wire-visible improvement — but a behavior change, so "no behavior change" undersells it.

**Tests:**

- Still no committed test drives a real `TracerProvider` through `inject_trace_context` (`TestIsMcpV1` doesn't touch this — it patches `_package_version`), despite the body's manual-verification claim. A 2-test version passes and needs no new dependency (`opentelemetry-sdk` is already direct at `pyproject.toml:43`).
- `test_mcp_instrumentation_idempotent_with_multiple_clients` is single-threaded and mocks `register_post_import_hook`, so it can't see the race in inline comment 3 — it asserts exactly the promise that's falsifiable. A forced two-thread version registers hooks 4× instead of 2×, deterministically; that's the test to add alongside the lock.
- `test_call_tool_sync_forwards_meta` (`test_mcp_client.py:190`) and `..._async_...` (`:575`) pass unchanged with `inject_trace_context` swapped for the identity function — they prove passthrough, not that the injection wiring works. (The 4 `TestInjectTraceContext` tests do hold up: I mutation-tested them 4 ways — dropping the copy, dropping `or None`, injecting into the wrong object, and dropping `inject()` entirely — and **all 4 mutations are caught**.)

**Checked clean, no action:** bare `traceparent`/`tracestate` in `_meta` is spec-legal (MCP SEP-414 sanctions it as the exception to the DNS-prefix rule; mcp 1.x `RequestParams.Meta` is `extra="allow"`, so keys survive verbatim) · no other `BaseSession.send_request`/`ClientSession` consumer anywhere in the monorepo, so no other caller regresses · `meta` is injected once before the `use_task` branch and both closures capture the final value — no double-injection, no clash with the progress-callback path · the docstrings' "`mcp` 2.x propagates trace context natively" is **accurate for both halves** (client injection at `mcp/shared/jsonrpc_dispatcher.py:390`; server extraction via `OpenTelemetryMiddleware()`, installed by default at `mcp/server/lowlevel/server.py:439`) · evergreen-comment rule satisfied, no unrelated churn, no stale user-facing docs.

*Also: 5 of the 6 deleted `test_patch_mcp_client_*` tests are correctly gone with the dead code — only the exception-handling one covered behavior that still matters (inline comment 2).*

</details>

---
*Independent multi-pass review. The passes disagreed on one central point and I re-ran the deciding repros myself; the whole set was then re-verified on `fd2ac179` after the amend. Worth a human's judgment on the `COMMENT` verdict rather than treating it as a gate.*

poshinchen:
@strandly-the-agent updated, review the code again.

strands-py/src/strands/tools/mcp/mcp_instrumentation.py
```diff
@@ -41,79 +96,36 @@ class ItemWithContext:
...
-                    # Handle dict params directly
-                    meta = request.root.params.setdefault("_meta", {})
-                    propagate.get_global_textmap().inject(meta)
-
-            return wrapped(*args, **kwargs)
-
-        except Exception:
-            return wrapped(*args, **kwargs)
+        if not _is_mcp_v1():
+            return
```
strandly-the-agent, line 128: ⚪ Non-blocking, and I want to correct something I nearly posted with inflated numbers.

The flag goes `True` at `:125`, but the patches install at `:180-190` — *after* the `with` block ends at `:128`. So there is a window where the flag says "applied" while `ServerSession.__init__` is still unwrapped. I measured it on this head: the gap from flag-set to hooks-registered is **~440 µs** (min 428 / median 442 / max 1588 over 20 runs). But a concurrent caller actually *landing* in it showed up in only **1/200 trials** with my probe — an earlier draft of this comment claimed 200/200 from a different harness, and my own measurement doesn't support that. So: real, narrow, telemetry-only, and it needs an in-process MCP server creating a session inside that ~0.4 ms. Not worth blocking on.

Two smaller things fall out of the same ordering:

- If anything raised between `:130` and `:190`, the flag is already `True`, so instrumentation would be silently off for the life of the process. On `fd2ac179` the flag was set last, so a failure left it `False` and the next `MCPClient()` retried. I couldn't name a realistic raiser on mcp 1.29 (it needs `StreamableHTTPServerTransport.connect` or `ServerSession` renamed/removed), so this is thin — but fail-open-and-silent is the wrong direction.
- `:192-193` (`# Mark instrumentation as applied` / `_instrumentation_applied = True`) is now a dead store: nothing clears the flag between `:125` and there. Deleting those two lines leaves 219 tests passing. The comment also describes an ordering the code doesn't have, which is what makes this shape easy to misread.

All three go away by moving the two `register_post_import_hook` calls inside the lock and setting the flag last (plain block, not a `suggestion`, because the change runs past `:190` and out of this diff hunk):

```python
    global _instrumentation_applied

    # ... transport_wrapper / session_init_wrapper definitions unchanged ...

    # Apply patches. The lock makes the check-and-set atomic, and the flag is set
    # only once the patches are installed, so a concurrent caller either waits or
    # observes a fully instrumented state.
    with _instrumentation_lock:
        if _instrumentation_applied:
            return

        if _is_mcp_v1():
            register_post_import_hook(
                lambda _: wrap_function_wrapper(
                    "mcp.server.streamable_http", "StreamableHTTPServerTransport.connect", transport_wrapper()
                ),
                "mcp.server.streamable_http",
            )

            register_post_import_hook(
                lambda _: wrap_function_wrapper("mcp.server.session", "ServerSession.__init__", session_init_wrapper()),
                "mcp.server.session",
            )

        _instrumentation_applied = True
```

Verified against this branch: `mypy` clean, 219 tests pass, layers = 1 at N=8/32, and it's **net smaller** than the current code (8 insertions / 16 deletions ignoring whitespace). It does put wrapt frames under our lock — I checked the inversion and it's safe: wrapt deliberately invokes post-import hooks *outside* its own lock, and deadlocking would need `import mcp.server.session` to construct an `MCPClient` at module scope, which mcp can't do since it doesn't depend on strands.

Fine to leave exactly as is; the current code is correct for the thing it was fixing.

strandly-the-agent (commented):
**No blockers from me on `898cc0ef`.** All three findings from my last pass are properly fixed, and you'd already fixed the one thing I was going to flag as blocking before I could post it.

Worth recording since it nearly shipped: `ddeefbd2` introduced `mypy` `error: Statement is unreachable [unreachable]` at the double-check under the new lock — mypy narrows the global to `False` after the outer fast-path return, and `warn_unreachable = true` (`pyproject.toml:222`) makes that fatal in `lint-check = ["ruff check", "mypy ./src"]`, which `hatch fmt --linter --check` runs at `.github/workflows/python-test-lint.yml:182` on every PR. Dropping the outer fast-path — exactly what `898cc0ef` does — clears it. Confirmed clean now.

| Verified on `898cc0ef` | |
|---|---|
| ✅ `mypy ./src` | no `[unreachable]` anywhere; the changed file is clean |
| ✅ `ruff check` | clean |
| ✅ `pytest tests/strands/tools/mcp/` | **219 passed** |
| ✅ lock still correct without the fast path | real wrapt layers = **1** on both targets at N=8, N=32, and via concurrent `MCPClient()` (was 8 at N=8 pre-fix) |
| ✅ propagator guard | raising propagator → returns the caller's meta, never escapes |
| ✅ rewritten `mcp_client.py` comment | accurate |
| ✅ both new tests earn their keep | each **fails** on `fd2ac179`, passes here |

One inline note on the flag-vs-patch ordering — explicitly optional, and I've corrected my own overstatement of it there.

**Correction, my error:** an earlier version of this comment claimed the description still carried the inaccurate task-path line. It doesn't — you removed it in the 19:24 edit, before I posted, and the current body correctly sticks to the propagator-swallow guarantee and the lock rationale. I'd drafted that paragraph from a copy of the body I read at 19:16 and never re-read it when I re-fetched the code for `898cc0ef`. That item is closed; sorry for the noise.

<details>
<summary>Appendix — 4 non-blocking nits, plus one pre-existing bug worth its own issue</summary>

- **⚪ Log volume.** With a persistently broken propagator — the misconfigured-`OTEL_PROPAGATORS` case the guard exists for — measured `500 calls → 500 WARNING records, 1.00 per call`, each ~348 chars / 6 lines with the traceback. The guard turns a hard failure into unbounded log spam. Warn-once behind a module flag, or `warning` then `debug`, would cap it.
- **⚪ Partial carrier.** A propagator that writes some keys then raises leaves the partial writes in the returned dict, so a truncated `traceparent` reaches the wire: `{'traceparent': '00-4bf92f3577b34da6a3ce929d0e0e4'}`. Low blast radius — a real `TraceContextTextMapPropagator().extract()` ignores it silently. Contrived (needs a partially-writing, then-raising propagator). One-line hardening: inject into a scratch dict, merge on success.
- **⚪ `thread.join(timeout=5)` in the new concurrency test** swallows worker exceptions (`threading.Thread` doesn't propagate, and there's no `filterwarnings = error`) and never asserts the threads finished. Latent rather than masking anything today — `call_count == 2` happens to catch a raise at or before the registers. `assert not t.is_alive()` plus collecting worker exceptions closes it.
- **⚪ The warning is covered by your manual run but not by a committed test.** Replacing `logger.warning(...)` with `pass` still leaves 219 passing, so a later simplification to a silent `except Exception: pass` would be invisible to CI. One `caplog` assertion covers it. (The committed test does assert the *returned value*, which is the part that actually matters.)

**Pre-existing, not this PR's problem:** `propagate.extract` in the server-side reader is the unguarded mirror of the bug you just fixed — a propagator whose `extract()` raises takes down the whole server-side message loop, not just one call. Present on `fd2ac179` too and untouched by this diff. Happy to file it with a repro if you want it tracked; I'd rather be asked than open one uninvited.

</details>

---
*Delta-only re-review (`ddeefbd2 → 898cc0ef`), resolved threads not revisited. `COMMENT` rather than approve because a human should own the merge signal — but there is nothing I'd hold this on.*

poshinchen:
@strandly-the-agent have you verified that the OTEL propagation actually works, the traceId / spanId actually gets propagated to the MCP _meta?

strandly-the-agent:
> **Edited.** The first copy of the script I pasted below was a cleaned-up rewrite of the file I actually ran, not the file itself — and the rewrite had a scenario-ordering bug: `check(..., WIRE[-1], *asyncio.run(_async_call()))` evaluates `WIRE[-1]` *before* `asyncio.run(...)`, so the last two scenarios compared the **previous** request's `_meta` against the new span's IDs. Running that version gives **4/6**, with `call_tool_async` and `MCPAgentTool.stream` printing `both match -> FAIL`. If you ran it and saw that, this is why — the bug was in my paste, not in the PR. The block below is now the exact text I executed, both positive and negative, and the numbers in this comment come from running *that* text. Two stale line references fixed too.

Yes — and fair to ask, because my earlier passes only proved it at the unit level and through my own harnesses. So I built a real end-to-end check on `898cc0ef` (`mcp==1.29.0`): real `TracerProvider` + `InMemorySpanExporter`, real `MCPClient` with its background thread, real `ClientSession` over real MCP streams, and a server that records the **exact `params` it receives off the wire**. Nothing in strands is mocked and the propagator is the real W3C one. Then I compared the wire `traceparent` against the client span's actual `trace_id`/`span_id`.

**6/6 scenarios pass — both IDs match exactly**, including the real agent path:

| scenario | wire `traceparent` | trace_id | span_id |
|---|---|---|---|
| `call_tool_sync` in a span | `00-fe7fea42…57fe-d5daca162a877e30-03` | ✅ | ✅ |
| nested spans → innermost wins | `00-4a37c372…a784-e888390bc08a7c5e-03` | ✅ | ✅ |
| no active span | `_meta` **absent entirely** | — | — |
| caller `meta` + span | `00-c7aafee9…b9a4-82a2d133986b198a-03` + `com.example/request_id` kept | ✅ | ✅ |
| `call_tool_async` in a span | `00-9c836a27…7bcb-9765af0a2fc12302-03` | ✅ | ✅ |
| **`MCPAgentTool.stream`** (agent path) | `00-a0d1e049…cc3f-9e8bc1177f2a775c-03` | ✅ | ✅ |

The last row is the one that matters most in practice: `MCPAgentTool.stream` passes no `meta` at all (`mcp_agent_tool.py:113-119`), so `inject_trace_context` is the only possible source of that `_meta`.

**And the test can fail** — re-running with `NEUTRALIZE=1` (injection swapped for the identity function) drops **all five** traceparent scenarios to FAIL (`_meta on the wire : None`), i.e. `passed: 1/6`, the survivor being the one that asserts `_meta` is *absent*. So it's measuring the real behavior rather than passing vacuously.

<details>
<summary>What I did <b>not</b> prove, and one scope note — worth reading</summary>

**Task-augmented path: not proven end to end.** Driving it needs a server advertising the `tasks` capability plus a tool declaring `taskSupport`, and a faked create/poll/get-result flow. I didn't build that. What I can say: `meta` is injected once at `mcp_client.py:790`, *before* the `use_task` branch, and both closures capture that same already-enriched value — so the two paths cannot diverge in what they carry, and `test_mcp_client_tasks.py` covers `meta` forwarding into `call_tool_as_task`. Structural argument plus unit coverage, not an e2e measurement. Say the word if you want the real thing.

**Scope note on the server side (pre-existing, unchanged by this PR).** The half that *extracts* `_meta` back out is `TransportContextExtractingReader`, and it's installed only by the transport wrapper at `mcp_instrumentation.py:152`, registered against `mcp.server.streamable_http.StreamableHTTPServerTransport.connect`. So in-process servers hosted over **streamable HTTP** join the client's trace; a stdio or in-memory in-process server gets the `ServerSession` context-preservation pair (`:174-175`) but not the `_meta` extractor. That's how it already worked before this PR — flagging it only so "propagation works" isn't read more broadly than the evidence supports. Remote servers are of course the other side's job; we've done ours once the header is on the wire, which is what's now proven.

**This is the committed-test gap I flagged last round.** No test in the repo drives a real `TracerProvider` through `inject_trace_context` — all five `TestInjectTraceContext` cases mock `get_global_textmap`, so they prove the dict plumbing, not that a real span's IDs land on the wire. My script is throwaway, but it converts cleanly into a test (needs no new dependency — `opentelemetry-sdk` is already direct at `pyproject.toml:43`). Happy to open a small follow-up PR adding it if you'd like the guard; I'd rather be asked than push onto your branch.

</details>

<details>
<summary>The verification script, if you want to run it yourself</summary>

Run from `strands-py/` with `PYTHONPATH=src`. Set `NEUTRALIZE=1` for the negative control. Exit code is 0 only when every scenario passes.

```python
"""End-to-end proof that OTel trace context reaches the MCP `_meta` on the wire."""

from __future__ import annotations

import asyncio
import os
from contextlib import asynccontextmanager

from mcp.shared.memory import create_client_server_memory_streams
from mcp.shared.message import SessionMessage
from mcp.types import JSONRPCMessage, JSONRPCResponse
from opentelemetry import trace
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import SimpleSpanProcessor
from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter

import strands.tools.mcp.mcp_client as _mc
from strands.tools.mcp.mcp_client import MCPClient

if os.environ.get("NEUTRALIZE") == "1":
    _mc.inject_trace_context = lambda meta: meta  # negative control
    print("!! NEGATIVE CONTROL: injection disabled !!")

exporter = InMemorySpanExporter()
provider = TracerProvider()
provider.add_span_processor(SimpleSpanProcessor(exporter))
trace.set_tracer_provider(provider)
tracer = trace.get_tracer("verify-3611")

WIRE: list[dict] = []


async def _server_loop(read_stream, write_stream):
    """Minimal MCP server that records tools/call params verbatim."""
    async for msg in read_stream:
        if isinstance(msg, Exception):
            raise msg
        root = getattr(msg.message, "root", msg.message)
        method = getattr(root, "method", None)
        req_id = getattr(root, "id", None)

        if method == "initialize":
            result = {
                "protocolVersion": "2025-06-18",
                "capabilities": {"tools": {}},
                "serverInfo": {"name": "verify-server", "version": "1"},
            }
        elif method == "tools/list":
            result = {"tools": [{"name": "echo", "inputSchema": {"type": "object"}}]}
        elif method == "tools/call":
            dumped = root.model_dump(by_alias=True, mode="json", exclude_none=True)
            WIRE.append(dumped.get("params", {}))
            result = {"content": [{"type": "text", "text": "ok"}], "isError": False}
        else:
            continue

        if req_id is not None:
            await write_stream.send(
                SessionMessage(JSONRPCMessage(JSONRPCResponse(jsonrpc="2.0", id=req_id, result=result)))
            )


def make_transport():
    @asynccontextmanager
    async def _transport():
        async with create_client_server_memory_streams() as (client_streams, server_streams):
            task = asyncio.create_task(_server_loop(*server_streams))
            try:
                yield client_streams
            finally:
                task.cancel()

    return _transport()


def check(label, params, exp_trace, exp_span):
    meta = params.get("_meta")
    print(f"\n--- {label} ---\n  _meta on the wire : {meta}")
    if exp_trace is None:
        ok = not meta or "traceparent" not in meta
        print(f"  expect NO traceparent -> {'PASS' if ok else 'FAIL'}")
        return ok
    if not meta or "traceparent" not in meta:
        print("  expect traceparent -> FAIL (absent)")
        return False
    _v, got_trace, got_span, _f = meta["traceparent"].split("-")
    ok = got_trace == exp_trace and got_span == exp_span
    print(f"  client span      : trace={exp_trace} span={exp_span}")
    print(f"  wire traceparent : trace={got_trace} span={got_span}")
    print(f"  both match -> {'PASS' if ok else 'FAIL'}")
    return ok


def ids(span):
    c = span.get_span_context()
    return f"{c.trace_id:032x}", f"{c.span_id:016x}"


def main():
    results = []
    with MCPClient(make_transport) as client:
        with tracer.start_as_current_span("agent.tool.echo") as s:
            exp = ids(s)
            client.call_tool_sync("tu-1", "echo", {"a": 1})
        results.append(check("call_tool_sync INSIDE a span", WIRE[-1], *exp))

        with tracer.start_as_current_span("outer"), tracer.start_as_current_span("inner") as inner:
            exp = ids(inner)
            client.call_tool_sync("tu-2", "echo", {"a": 2})
        results.append(check("nested spans -> innermost", WIRE[-1], *exp))

        client.call_tool_sync("tu-3", "echo", {"a": 3})
        results.append(check("no active span", WIRE[-1], None, None))

        with tracer.start_as_current_span("with-caller-meta") as s:
            exp = ids(s)
            client.call_tool_sync("tu-4", "echo", {"a": 4}, meta={"com.example/request_id": "abc-123"})
        ok = check("caller meta + span", WIRE[-1], *exp)
        kept = WIRE[-1].get("_meta", {}).get("com.example/request_id") == "abc-123"
        print(f"  caller's key kept : {kept}")
        results.append(ok and kept)

        async def _async_call():
            with tracer.start_as_current_span("async-call") as s:
                exp = ids(s)
                await client.call_tool_async("tu-5", "echo", {"a": 5})
                return exp

        # Bind the span ids BEFORE reading WIRE[-1]: inlining asyncio.run() into the
        # check(...) arguments would evaluate WIRE[-1] first and compare the previous
        # request's _meta against this span.
        exp_trace, exp_span = asyncio.run(_async_call())
        results.append(check("call_tool_async INSIDE a span", WIRE[-1], exp_trace, exp_span))

        from mcp.types import Tool as MCPTool

        from strands.tools.mcp.mcp_agent_tool import MCPAgentTool

        agent_tool = MCPAgentTool(MCPTool(name="echo", inputSchema={"type": "object"}), client)

        async def _agent_call():
            with tracer.start_as_current_span("agent.tool.echo-via-agenttool") as s:
                exp = ids(s)
                async for _ in agent_tool.stream({"toolUseId": "tu-6", "name": "echo", "input": {}}, {}):
                    pass
                return exp

        exp_trace, exp_span = asyncio.run(_agent_call())
        results.append(check("MCPAgentTool.stream (real agent path)", WIRE[-1], exp_trace, exp_span))

    print(f"\ntools/call received: {len(WIRE)} | passed: {sum(results)}/{len(results)}")
    return 0 if all(results) else 1


if __name__ == "__main__":
    raise SystemExit(main())
```

</details>

liramon2 (approved):
Looks good to me. Noting that `_instrumentation_applied = True` in `mcp_instrumentation` is dead code now, but probably worth keeping in case the `_is_mcp_v1` call is ever removed.

PR #3594, opened 2026-07-31
feat(middleware): add AgentStreamStage with middleware-initiated interrupts

strandly-the-agent:
## Description

@mkmeral asked me to fork #3466 and fix the issues my review of it found, so this branch **contains #3466's work plus four fixes on top** — it is not a patch to be applied alongside it. **Merge this or #3466, not both.** `strands-ts/` is carried over from #3466 untouched; every change of mine is under `strands-py/`.

`AgentStreamStage` lets middleware gate a whole invocation pass behind a human-in-the-loop interrupt. The bugs were all about *how long an answered approval lives*, and they failed in the direction that matters: a human being asked to approve the same action repeatedly, or — worse — an action being approved silently because a stale answer was still lying around.

Four fixes, each with a reproduction that failed before it:

1. **A cancelled resume destroyed a pending tool interrupt.** The tool phase cleared interrupt state unconditionally, so cancelling a resumed tool interrupt threw away the stored `tool_use_message` and the human's answer, leaving the invocation unresumable. Only an aborting pass (`agent.cancel()`) now keeps its state. This one is **a live bug on `main`, not just in #3466** — it arrived with #3508's tool-batch hooks, and #3466's own `test_cancel_during_tool_interrupt_resume_preserves_interrupt_state` fails on `main` today because of it. The TypeScript side already had the equivalent guard.
2. **An answered gate was re-asked once per later interrupt round trip.** The same unconditional clear also destroyed an *answered* agent-stream response mid-invocation. #3466's pass-start snapshot stabilises reads within a pass, but it is copied from live state, so it cannot recover what an earlier pass already cleared. A gate reading its approval before `next_fn` was asked 4 times in one flow; with a session manager and a fresh `Agent` per pass it never converged, re-firing tool side effects each round.
3. **An interrupt raised after the pass produced its result silently replayed the pass** — a second model call, two consecutive assistant turns, and re-fired side effects (one repro charged a card twice for a single request). Now refused with an actionable error, with interrupt state cleared so the caller isn't wedged.
4. **Two run-loop branches nothing exercised** — the per-pass result capture that feeds `AfterInvocationEvent` (and therefore session persistence), and the `stop_reason != "interrupt"` guard. Both survived mutation at two different heads; both are now pinned.

[...]

PR #3615, opened 2026-08-03
fix: preserve interrupt state across cancelled resumes and middleware re-reads

liramon2:
## Description
The TS changes in https://github.com/strands-agents/harness-sdk/pull/3466 (and subsequently https://github.com/strands-agents/harness-sdk/pull/3594) were split off into this PR. It fixes the following issues regarding interrupt lifecycle handling:
- When the agent resumes from an interrupt and the resume pass is cancelled before the interrupted tool ran, the pending state is wiped unconditionally. This causes subsequent resume attempts to fail.
- When the user resumes from an `AgentStreamStage` middleware interrupt and the pass completes without triggering any tool cycle, the interrupt was never deactivated. This caused future invocations to be rejected.
- When a middleware gate resolves an interrupt response before `next()` and re-reads it after `next()` returns, the tool cycle in `next()` may have already deactivated the interrupt. The gate would throw a new interrupt instead of returning the response, causing a livelock.

## Type of Change

Bug fix

## Testing

- [x] I ran the pre-commit checks

## Checklist

----

### DIFF
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -1168,11 +1168,15 @@ export class Agent implements LocalAgent, InvokableAgent {
     options: InvokeOptions,
     invocationState: InvocationState
   ): AsyncGenerator<AgentStreamEvent, AgentResult, undefined> {
+    // Snapshot interrupts before the pass: a tool cycle within this pass can call the event
+    // loop's deactivate() and clear the live map, but a gate that re-reads its approval after
+    // next() must still see the resumed response.
+    const interruptsSnapshot = { ...this._interruptState.interrupts }
     const context: AgentStreamContext = {
       agent: this,
       args,
       ...(options !== undefined && { options }),
-      interrupt: createMiddlewareInterrupt(this._interruptState, 'middleware:agentStream'),
+      interrupt: createMiddlewareInterrupt(this._interruptState, 'middleware:agentStream', interruptsSnapshot),
     }
 
     // async function* doesn't bind lexical `this`; capture for the terminal callback.
@@ -1187,6 +1191,18 @@ export class Agent implements LocalAgent, InvokableAgent {
           return { result }
         }
       )
+      // A resumed AgentStreamStage interrupt that completes without a tool cycle never reaches
+      // the tool loop's deactivate(), so clear the interrupt state here. Guard on the absence of
+      // a pending tool execution so this only clears agent-stream interrupts and never wipes a
+      // pending tool interrupt — some non-interrupt endings (e.g. a cancelled resume) keep the
+      // tool interrupt activated on purpose.
+      if (
+        this._interruptState.activated &&
+        result.stopReason !== 'interrupt' &&
+        !this._interruptState.pendingToolExecution
+      ) {
+        this._interruptState.deactivate()
+      }
       return result
     } catch (error) {
       if (error instanceof InterruptError) {
@@ -1497,10 +1513,12 @@ export class Agent implements LocalAgent, InvokableAgent {
           let completedToolResults: Map<string, ToolResultBlock> | undefined
 
           if (pendingExecution) {
-            // Resume from stored state - skip model call
+            // Resume from stored state - skip model call. The pending marker is NOT cleared
+            // here: if this resumed pass is cancelled before the tool runs, the marker must
+            // survive so a later resume can still execute the tool. It is cleared by the
+            // deactivate() below once the tools actually execute.
             assistantMessage = pendingExecution.assistantMessage
             completedToolResults = pendingExecution.completedToolResults
-            this._interruptState.clearPendingToolExecution()
           } else {
             const modelResult = yield* this._invokeModel(invocationState, structuredOutputChoice)
 
@@ -1618,9 +1636,10 @@ export class Agent implements LocalAgent, InvokableAgent {
           yield this._appendMessage(assistantMessage, invocationState)
           yield this._appendMessage(toolResultMessage, invocationState)
 
-          // Deactivate interrupt state after successful tool execution so the next
-          // cycle starts with a clean slate (new interrupts can be raised again).
-          if (this._interruptState.activated) {
+          // Deactivate interrupt state after tools actually execute so the next cycle starts
+          // clean. Skip when cancelled: the tools were cancel-skipped (not run), so a pending
+          // tool interrupt must survive for a later resume rather than being wiped here.
+          if (this._interruptState.activated && !this.isCancelled) {
             this._interruptState.deactivate()
           }
 
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -388,7 +388,7 @@ const DEFAULT_AGENT_NAME = 'Strands Agent'
 const DEFAULT_AGENT_ID = 'agent'
 
 /** Result returned by tool-execution generators, threading the AfterToolsEvent back to the main loop. */
-type ToolsExecutionResult = { message: Message; afterToolsEvent: AfterToolsEvent }
+type ToolsExecutionResult = { message: Message; afterToolsEvent: AfterToolsEvent; toolsSkipped: boolean }
 
 /**
  * Orchestrates the interaction between a model, a set of tools, and MCP clients.
@@ -1617,6 +1617,14 @@ export class Agent implements LocalAgent, InvokableAgent {
           }
           const toolResultMessage = toolsResult.message
 
+          // Tools were skipped (not executed) — preserve pending state so the next resume
+          // can run them.
+          if (this.isCancelled && toolsResult.toolsSkipped && this._interruptState.pendingToolExecution) {
+            this._meter.endCycle(cycleStartTime)
+            this._tracer.endAgentLoopSpan(cycleSpan)
+            continue
+          }
+
           /**
            * Deferred append: both messages are added AFTER tool execution completes.
            * This keeps agent.messages in a valid, reinvokable state at all times.
@@ -1626,10 +1634,9 @@ export class Agent implements LocalAgent, InvokableAgent {
           yield this._appendMessage(assistantMessage, invocationState)
           yield this._appendMessage(toolResultMessage, invocationState)
 
-          // Deactivate interrupt state after tools actually execute so the next cycle starts
-          // clean. Skip when cancelled: the tools were cancel-skipped (not run), so a pending
-          // tool interrupt must survive for a later resume rather than being wiped here.
-          if (this._interruptState.activated && !this.isCancelled) {
+          // Deactivate interrupt state after successful tool execution so the next
+          // cycle starts with a clean slate (new interrupts can be raised again).
+          if (this._interruptState.activated) {
             this._interruptState.deactivate()
           }
 
@@ -2220,6 +2227,7 @@ export class Agent implements LocalAgent, InvokableAgent {
     const toolResultBlocks: ToolResultBlock[] = []
     let toolResultMessage: Message
     let afterToolsEvent: AfterToolsEvent
+    let toolsSkipped = false
 
     try {
       if (toolUseBlocks.length === 0) {
@@ -2236,6 +2244,7 @@ export class Agent implements LocalAgent, InvokableAgent {
           : undefined
 
       if (cancelMessage) {
+        toolsSkipped = true
         toolResultBlocks.push(...this._cancelAllAsResults(toolUseBlocks, cancelMessage))
         for (const result of toolResultBlocks) {
           yield new ToolResultEvent({ agent: this, result, invocationState })
@@ -2263,7 +2272,7 @@ export class Agent implements LocalAgent, InvokableAgent {
       yield afterToolsEvent
     }
 
-    return { message: toolResultMessage, afterToolsEvent }
+    return { message: toolResultMessage, afterToolsEvent, toolsSkipped }
   }
 
   /**
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -1634,6 +1634,9 @@ export class Agent implements LocalAgent, InvokableAgent {
           yield this._appendMessage(assistantMessage, invocationState)
           yield this._appendMessage(toolResultMessage, invocationState)
 
+          // Both messages are in history, so any stored pending execution is now stale.
+          this._interruptState.clearPendingToolExecution()
+
           // Deactivate interrupt state after successful tool execution so the next
           // cycle starts with a clean slate (new interrupts can be raised again).
           if (this._interruptState.activated) {
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -1168,9 +1168,7 @@ export class Agent implements LocalAgent, InvokableAgent {
     options: InvokeOptions,
     invocationState: InvocationState
   ): AsyncGenerator<AgentStreamEvent, AgentResult, undefined> {
-    // Snapshot interrupts before the pass: a tool cycle within this pass can call the event
-    // loop's deactivate() and clear the live map, but a gate that re-reads its approval after
-    // next() must still see the resumed response.
+    // Snapshot so a gate that re-reads its response after next() still resolves even if a tool cycle called deactivate().
     const interruptsSnapshot = { ...this._interruptState.interrupts }
     const context: AgentStreamContext = {
       agent: this,
@@ -1191,11 +1189,6 @@ export class Agent implements LocalAgent, InvokableAgent {
           return { result }
         }
       )
-      // A resumed AgentStreamStage interrupt that completes without a tool cycle never reaches
-      // the tool loop's deactivate(), so clear the interrupt state here. Guard on the absence of
-      // a pending tool execution so this only clears agent-stream interrupts and never wipes a
-      // pending tool interrupt — some non-interrupt endings (e.g. a cancelled resume) keep the
-      // tool interrupt activated on purpose.
       if (
         this._interruptState.activated &&
         result.stopReason !== 'interrupt' &&
@@ -1513,10 +1506,7 @@ export class Agent implements LocalAgent, InvokableAgent {
           let completedToolResults: Map<string, ToolResultBlock> | undefined
 
           if (pendingExecution) {
-            // Resume from stored state - skip model call. The pending marker is NOT cleared
-            // here: if this resumed pass is cancelled before the tool runs, the marker must
-            // survive so a later resume can still execute the tool. It is cleared by the
-            // deactivate() below once the tools actually execute.
+            // Resume from stored state - skip model call.
             assistantMessage = pendingExecution.assistantMessage
             completedToolResults = pendingExecution.completedToolResults
           } else {
--- a/strands-ts/src/agent/__tests__/agent.interrupt.test.ts
+++ b/strands-ts/src/agent/__tests__/agent.interrupt.test.ts
@@ -2,8 +2,14 @@ import { describe, expect, it, vi } from 'vitest'
 import { Agent } from '../agent.js'
 import { MockMessageModel } from '../../__fixtures__/mock-message-model.js'
 import { createMockTool } from '../../__fixtures__/tool-helpers.js'
-import { ToolResultBlock, ToolUseBlock } from '../../types/messages.js'
-import { AfterToolCallEvent, BeforeToolCallEvent, BeforeToolsEvent, InterruptEvent } from '../../hooks/events.js'
+import { TextBlock, ToolResultBlock, ToolUseBlock } from '../../types/messages.js'
+import {
+  AfterToolCallEvent,
+  BeforeToolCallEvent,
+  BeforeToolsEvent,
+  InterruptEvent,
+  MessageAddedEvent,
+} from '../../hooks/events.js'
 import { FunctionTool } from '../../tools/function-tool.js'
 import { InterruptResponseContent } from '../../types/interrupt.js'
 import type { InterruptState, PendingToolExecution } from '../../interrupt.js'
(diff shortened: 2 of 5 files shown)


strands-ts/src/agent/agent.ts
```diff
@@ -1168,11 +1168,15 @@ export class Agent implements LocalAgent, InvokableAgent {
     options: InvokeOptions,
     invocationState: InvocationState
   ): AsyncGenerator<AgentStreamEvent, AgentResult, undefined> {
+    // Snapshot interrupts before the pass: a tool cycle within this pass can call the event
+    // loop's deactivate() and clear the live map, but a gate that re-reads its approval after
```
JackYPCOnline: Nit: Are these explanation lines required?

strands-ts/src/agent/agent.ts
```diff
@@ -1187,6 +1191,18 @@ export class Agent implements LocalAgent, InvokableAgent {
           return { result }
         }
       )
+      // A resumed AgentStreamStage interrupt that completes without a tool cycle never reaches
+      // the tool loop's deactivate(), so clear the interrupt state here. Guard on the absence of
+      // a pending tool execution so this only clears agent-stream interrupts and never wipes a
```
JackYPCOnline: same issue for these inline comments.

JackYPCOnline:
Overall comment: we should not post inline comments that reflect agent thinking process and rationals

strands-ts/src/agent/agent.ts
```diff
@@ -1187,6 +1191,18 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+      // pending tool interrupt — some non-interrupt endings (e.g. a cancelled resume) keep the
+      // tool interrupt activated on purpose.
+      if (
+        this._interruptState.activated &&
+        result.stopReason !== 'interrupt' &&
+        !this._interruptState.pendingToolExecution
+      ) {
+        this._interruptState.deactivate()
```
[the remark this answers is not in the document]
liramon2: This is intentional

strands-ts/src/agent/agent.ts
liramon2: I think the snapshot's purpose is non-obvious and could use a comment. That being said, I'll shorten it to keep it focused on the purpose (letting middleware gates re-read the response even when tool cycles deactivate the interrupt)

strands-ts/src/agent/agent.ts
liramon2: Removed

strands-ts/src/agent/agent.ts
```diff
@@ -1168,11 +1168,13 @@ export class Agent implements LocalAgent, InvokableAgent {
     options: InvokeOptions,
     invocationState: InvocationState
   ): AsyncGenerator<AgentStreamEvent, AgentResult, undefined> {
+    // Snapshot so a gate that re-reads its response after next() still resolves even if a tool cycle called deactivate().
+    const interruptsSnapshot = { ...this._interruptState.interrupts }
```
JackYPCOnline, line 1172: Question: if you are using Snapshot, does this field contains agent snapshot? is this naming used in python? if not, can we find better naming?

liramon2: `interruptsSnapshot` does not include an agent snapshot; it is only a `Record` of `Interrupt` instances keyed by their IDs. This naming is used elsewhere in Python outside the context of agent snapshots like in [model state](https://github.com/strands-agents/harness-sdk/blob/8b1f53724fdd63593fd0a91c845bd63fc62c11e4/strands-py/src/strands/event_loop/event_loop.py#L552) and [asyncio task references](https://github.com/strands-agents/harness-sdk/blob/8b1f53724fdd63593fd0a91c845bd63fc62c11e4/strands-py/src/strands/memory/extraction/coordinator.py#L187). Both of them were used to avoid race conditions. `interruptsSnapshot` avoids stale references, which is also due to a shared mutable state.

JackYPCOnline:
@strandly-the-agent This PR looks good to me, do a final review before merging.

strands-ts/src/agent/agent.ts
```diff
@@ -1618,9 +1626,10 @@ export class Agent implements LocalAgent, InvokableAgent {
           yield this._appendMessage(assistantMessage, invocationState)
           yield this._appendMessage(toolResultMessage, invocationState)
 
-          // Deactivate interrupt state after successful tool execution so the next
-          // cycle starts with a clean slate (new interrupts can be raised again).
-          if (this._interruptState.activated) {
+          // Deactivate interrupt state after tools actually execute so the next cycle starts
+          // clean. Skip when cancelled: the tools were cancel-skipped (not run), so a pending
+          // tool interrupt must survive for a later resume rather than being wiped here.
+          if (this._interruptState.activated && !this.isCancelled) {
```
strandly-the-agent: **🔴 An approved tool runs twice if the cancel lands during or after tool execution.**

This guard keys on *"is the invocation cancelled"*, but the intent (per its comment) is *"were the tools cancel-skipped"* — and those diverge. It also sits **after** both `_appendMessage` calls at 1626-1627, so by the time it runs the `toolUse`/`toolResult` pair is already in history and a surviving `pendingToolExecution` is stale. `_stream` at 1508-1511 then replays that stored `assistantMessage` on the next resume.

Two consequences, both A/B'd against base `a2656017`:

1. **Cancel while the approved tool is in flight** (public API only: `invoke([InterruptResponseContent])` to approve, then `agent.cancel()` / a `cancelSignal` abort — i.e. a stop button in any HITL app): the side effect runs, the pass ends `cancelled` with `activated=true, pending=true`, a fresh prompt throws `Agent is in an interrupted state`, so the caller must re-resume — and the re-resume replays the tool. Side effect happens **twice**, `tool-1` appears twice with two results. Base: side effect once, state clean, fresh prompt accepted. The persisted session snapshot is self-inconsistent too — history has `toolResult(tool-1,success)` while `pendingToolExecution.completedToolResults` is `{}`.
2. **The PR's own scenario** (cancel *before* the tool runs, as in the new test at `agent.interrupt.test.ts:639`): the tool correctly runs once, but after the later successful resume history contains `tool-1` **twice** — once with the synthetic `Tool execution cancelled` result and once executed. That breaks the invariant documented right above at 1620-1625. The new test asserts only `stopReason`/`toolExecuted`, so it passes over the corrupted history.

<details><summary>Fix I validated (3901 tests + the 40 interrupt tests pass, type-check clean, fix 1's goal preserved)</summary>

Gate on *did any tool actually run*, and skip the append when none did, rather than keeping pending state alongside already-appended results:

```diff
-type ToolsExecutionResult = { message: Message; afterToolsEvent: AfterToolsEvent }
+type ToolsExecutionResult = { message: Message; afterToolsEvent: AfterToolsEvent; toolsSkipped: boolean }

   // in _stream, before the two _appendMessage calls:
+  // Cancelled before any tool of a resumed tool interrupt ran: appending the synthetic
+  // cancel results would make the surviving pending execution replay them on the next
+  // resume (duplicate toolUse/toolResult, duplicate side effect).
+  if (this.isCancelled && toolsResult.toolsSkipped && this._interruptState.pendingToolExecution) {
+    this._meter.endCycle(cycleStartTime)
+    this._tracer.endAgentLoopSpan(cycleSpan)
+    continue
+  }

   // at this line — both messages are in history, so any pending execution is stale:
-  if (this._interruptState.activated && !this.isCancelled) {
+  if (this._interruptState.activated) {

   // in executeTools: set toolsSkipped = true in the `if (cancelMessage)` pre-launch branch
   // and thread it through the return.
```
</details>

<details><summary>Drop-in regression test (fails on this head, passes with the patch)</summary>

```ts
it('runs an approved tool only once when the resume is cancelled while the tool is in flight', async () => {
    .addTurn({ type: 'toolUseBlock', name: 'sendEmail', toolUseId: 'tool-1', input: {} })
    .addTurn({ type: 'textBlock', text: 'Done' })
    .addTurn({ type: 'textBlock', text: 'Nothing else happened' })

  let raised = false
  const sideEffects: string[] = []
  let releaseTool!: () => void
  const toolReleased = new Promise<void>((resolve) => (releaseTool = resolve))
  let toolStarted!: () => void
  const toolInFlight = new Promise<void>((resolve) => (toolStarted = resolve))

  const tool = createMockTool('sendEmail', (context) => {
    if (!raised) {
      raised = true
      context.interrupt({ name: 'confirm', reason: 'Send the email?' })
    }
    return (async function* () {
      sideEffects.push('email sent') // the irreversible effect the human approved
      toolStarted()
      await toolReleased
      return new ToolResultBlock({
        toolUseId: context.toolUse.toolUseId,
        status: 'success',
        content: [new TextBlock('sent')],
      })
    })()
  })

  const agent = new Agent({ model, tools: [tool], printer: false })

  const interruptResult = await agent.invoke('Send the email')
  const interruptId = interruptResult.interrupts![0]!.id

  // Human approves, then presses "stop" while the tool is still running.
  const inflight = agent.invoke([new InterruptResponseContent({ interruptId, response: 'yes' })])
  await toolInFlight
  agent.cancel()
  releaseTool()
  expect((await inflight).stopReason).toBe('cancelled')

  // The tool ran and its result is in history, so nothing is left to resume.
  expect(getPendingToolExecution(agent)).toBeUndefined()
  const toolUseIds = agent.messages.flatMap((m) =>
    m.content.filter((b) => b.type === 'toolUseBlock').map((b) => (b as ToolUseBlock).toolUseId)
  )
  expect(toolUseIds).toEqual(['tool-1'])

  // A fresh prompt is accepted (the agent is not wedged as interrupted).
  await agent.invoke('what happened?')
  expect(sideEffects).toEqual(['email sent'])
})
```
</details>

strandly-the-agent (changes requested):
**🔴 Hold the merge — one new regression at `f9a9fde`.** An approved tool re-runs on the next resume when the cancel lands *during or after* tool execution: the irreversible side effect happens twice and the agent is left wedged as interrupted. A/B'd against base `a2656017`, which is clean, so this PR introduces it. Mechanism + repro + a validated patch are inline at `agent.ts:1632`. Fixes 2 and 3 (the new deactivate guard and `interruptsSnapshot`) held up under everything I threw at them.

Verified at `f9a9fde` in a fresh clone:
- ✅ delta since my previous review (`a906fcc`) is comments-only
- ✅ `build`, `type-check`, `lint`, `format:check` clean; `unit-node` 141 files / **3901 tests pass** — the suite does not catch this
- 🔴 repro: approved tool executes twice, `tool-1` lands in history twice, fresh prompt rejected (base: once, clean, accepted)
- ✅ candidate patch: 3901 tests + the 40 interrupt tests pass, type-check clean, both repros green, and fix 1's own goal still holds
- ⚠️ CI: all TS checks and `CI Gate` are green at head, but `TypeScript: Secure Integration Test / Run integration tests (selective)` **failed**. I get a 403 on that job's log so I can't tell whether it's related — worth a human glance before merge.

<details><summary>Attacks that fixes 2 and 3 survived</summary>

Fix 2 (`agent.ts:1192-1198`) — probed 6 stop paths (turns-limit after a tool cycle, `checkpoint`, cancel during model streaming, structured output, non-interrupt `ModelError`, middleware replacing an interrupt result with `endTurn` while a tool interrupt is pending). Resulting state was self-consistent in every case (never `activated` with no interrupts, never interrupts with an orphaned pending execution), and 4 of the 6 are wedges/livelocks on base. Both `Interruptible` hooks (`BeforeToolCallEvent`, `BeforeToolsEvent`) sit inside the tool cycle, so there's no "hook interrupt with no pending state" hole for the guard to wrongly clear. Ordering checks out: the deactivate runs before `AfterInvocationEvent`, so the end-of-invocation session write persists clean state.

Fix 3 (`interruptsSnapshot`) — no cross-invocation leak (rebuilt per `_streamWithMiddleware` call, including on the resume loop, so an answered gate still re-asks on the next fresh invoke); answered interrupts can't be resurrected into live state (`registerInterrupt()` → `getOrCreateInterrupt()` always builds a fresh response-less instance); `resume()` mutates in place *before* the snapshot is taken, so the snapshot can't miss a response; `ExecuteToolStage` omitting the snapshot is safe because no `deactivate()` can run inside a single tool's middleware chain.
</details>

<details><summary>Appendix — non-blocking (3)</summary>

- ⚪ **Pre-existing:** when a middleware gate re-reads its response *after* `next()` and the pass contained a tool cycle, the pass re-runs and the tool executes a second time. This PR strictly improves it (head terminates in 3 round trips with 2 executions; base livelocks forever with 3), but it does mean the "gates re-run from scratch, so re-raising is a clean recovery" reasoning costs a duplicate side effect when the pass has tools.
- ⚪ **Pre-existing (identical on base):** if a middleware gate interrupts *before* `next()`, the user's prompt never reaches history — after the resume, `agent.messages` is just `[assistant: text(...)]` and the model was called with an empty history.
- ⚪ Within one pass, a second gate ask with the same `name` resolves from `interruptsSnapshot` instead of re-asking. Needs middleware that calls `next()` twice with a same-name gate — contrived, just worth knowing.

Happy to file the two pre-existing ones as issues if useful — say the word and I will.
</details>

liramon2:
@strandly-the-agent Re-review this PR.

strands-ts/src/agent/agent.ts
```diff
@@ -1497,10 +1506,9 @@ export class Agent implements LocalAgent, InvokableAgent {
           let completedToolResults: Map<string, ToolResultBlock> | undefined
 
           if (pendingExecution) {
-            // Resume from stored state - skip model call
+            // Resume from stored state - skip model call.
             assistantMessage = pendingExecution.assistantMessage
             completedToolResults = pendingExecution.completedToolResults
-            this._interruptState.clearPendingToolExecution()
```
strandly-the-agent, line 1503: **🔴 Removing this line leaves `pendingToolExecution` with no clear path that doesn't go through `activated`, so a stored tool cycle can replay forever.**

```
$ rg -n "clearPendingToolExecution" src --glob '!**/__tests__/**'
src/interrupt.ts:218:  clearPendingToolExecution(): void {   # declared, called from nowhere
```

This was the only call site, so `deactivate()` — gated on `activated` at `agent.ts:1639` — is now the sole clear. But the `InterruptEvent` fan-out at `agent.ts:1731` runs **before** `_createInterruptResult()` calls `activate()` (1734 → 1795). A consumer that stops consuming at that event leaves `activated=false, pending=set`, and then every later cycle takes the `getPendingExecution()` branch just above: model call skipped, stored assistant tool-use message re-appended, tool re-run — with the interrupt never answered.

`for await` + `break` is a supported pattern with its own suite (`agent.cancel.test.ts:246`), and **both** variants poison:

| ref | plain `break` → fresh prompt | `cancel()` + `break` → fresh prompt | plain `break` → documented resume |
|---|---|---|---|
| base `a2656017` | tool 1×, `endTurn`, modelCalls 2 | 1×, `endTurn` | 1×, `endTurn` |
| `f9a9fde` | **15×**, modelCalls 1 | **14×** | **15×** |
| `2c748ed` (head) | **15×**, modelCalls 1 | **14×** | **15×** |

15 is just where my kill switch fired (cancel once history passes 30 messages); with no cap the run hung past 4 minutes with a synchronous mock tool, and `_checkLimits` has no default cycle cap. `modelCalls` never advances — pure replay. On the resume path the user's approval is dropped too, since `resume()` early-returns when `!activated` (`interrupt.ts:273`). Base leaves the same stale `pending` after the break but self-heals, precisely because of this line.

**Fix** — clear unconditionally once both messages are in history, i.e. at `agent.ts:1637`, just before the existing `activated` check. Deliberately *after* the appends, so the new `toolsSkipped` `continue` above still preserves pending exactly as intended:

```ts
          yield this._appendMessage(assistantMessage, invocationState)
          yield this._appendMessage(toolResultMessage, invocationState)

          // Both messages are in history, so any stored pending execution is now stale.
          // Clear it unconditionally: a pending execution stored without activate() (a
          // consumer that broke the stream at InterruptEvent) would otherwise be replayed
          // by every later cycle.
          this._interruptState.clearPendingToolExecution()

          if (this._interruptState.activated) {
            this._interruptState.deactivate()
          }
```

Verified on a worktree of `2c748ed` with exactly that change: the repro drops to base numbers (1 run, `endTurn`, modelCalls 2, identical message counts), 141 files / 3901 unit tests pass, `tsc --noEmit` clean, and both scenarios this PR fixes still hold. Worth considering as follow-up hardening: move `activate()` ahead of the `InterruptEvent` fan-out so a break can't leave an agent secretly interrupted at all.

<details><summary>Repro (drop in as <code>src/agent/__tests__/zz-poison.test.ts</code>)</summary>

```ts
import { expect, it } from 'vitest'
import { Agent } from '../agent.js'
import { MockMessageModel } from '../../__fixtures__/mock-message-model.js'
import { createMockTool } from '../../__fixtures__/tool-helpers.js'
import { MessageAddedEvent } from '../../hooks/events.js'

it('does not replay a stored tool cycle when the stream is broken at interruptEvent', async () => {
    .addTurn({ type: 'toolUseBlock', name: 'deleteFiles', toolUseId: 'tool-1', input: {} })
    .addTurn({ type: 'textBlock', text: 'Done' })
    .addTurn({ type: 'textBlock', text: 'Second' })

  let raised = false
  let toolRuns = 0
  const tool = createMockTool('deleteFiles', (context) => {
    if (!raised) {
      raised = true
      context.interrupt({ name: 'confirm', reason: 'Delete?' })
    }
    toolRuns += 1
    return 'deleted'
  })

  const agent = new Agent({ model, tools: [tool], printer: false })

  // HITL streaming consumer: stop consuming once the interrupt surfaces.
  for await (const event of agent.stream('Delete old files')) {
    if (event.type === 'interruptEvent') break
  }

  // Kill switch, otherwise the next invocation never returns.
  agent.addHook(MessageAddedEvent, () => {
    if (agent.messages.length > 30) agent.cancel()
  })

  const result = await agent.invoke('anything new')
  expect(toolRuns).toBe(1)                   // head: 15
  expect(result.stopReason).toBe('endTurn')  // head: 'cancelled', only because of the kill switch
})
```
</details>

strandly-the-agent (changes requested):
**The fix I asked for is verified good — but a second 🔴 in the same mechanism is still open at `2c748ed`.** Dropping the eager `clearPendingToolExecution()` (back in `a906fcc`) left `deactivate()` as the *only* clear site, and it's gated on `activated`. Any path that stores pending state **without** activating — a consumer that `break`s out of `agent.stream()` at `interruptEvent`, with or without `cancel()` first — now makes every later cycle replay the stored tool cycle: the **unapproved** tool executes over and over and the model never advances. Base runs it once. One-line fix and repro inline on the deleted line in `agent.ts`.

Verified at `2c748ed` in a fresh clone:
- ✅ **my earlier blocker is genuinely fixed** — in-flight cancel runs the approved tool exactly once, `tool-1` in history once, agent not wedged; cancel-before-tool keeps the interrupt resumable with history untouched
- ✅ `build` / `type-check` / `lint` / `format:check` clean; `unit-node` 141 files / **3901 tests pass**
- 🔴 **new**: stream-break repro → tool ran **15×** (my kill switch capped it; without a cap the run hung past 4 minutes), `modelCalls` stuck at 1. Base `a2656017`: **1×**, `endTurn`. Same failure on `f9a9fde`, so it arrived with `a906fcc` — not with this commit
- ✅ the one-line fix reproduces base behaviour exactly and keeps 3901 tests + both fixed scenarios green, type-check clean
- 🟡 still **no committed regression test** for either fix — the suite passes 3901 on both the broken and the fixed tree, so either fix can silently regress
- ⚠️ CI: TS checks and `CI Gate` green; `Run integration tests (selective)` failed again (it also failed at `f9a9fde`, so probably environmental — I still get a 403 on that job's log, so someone with access should glance at it)

<details><summary>What I attacked in the new code and could not break</summary>

The `continue` escape at `agent.ts:1620-1625` terminates as `cancelled` on every cancel route I tried (hooks at `BeforeTools`/`BeforeToolCall`/`AfterToolCall`/`AfterTools`/`MessageAdded`/`ModelStreamUpdate`, external `AbortSignal`, consumer break + `cancel()`), with no unclosed spans and the cycle counted once — same convention as the existing early-`continue` paths at 1535/1616. `toolsSkipped` precision holds: every partial cancel (sequential, concurrent, `BeforeToolCallEvent`-level skip, throw-while-skipped) yields a complete result set, so the appends plus the now-unconditional `deactivate()` are correct — no dangling `toolUse`, duplicate id, double result, or lost resumable work. Session round-trip is clean too: persist a cancelled resume, restore into a fresh agent, resume → tool runs exactly once.
</details>

<details><summary>Appendix — non-blocking, pre-existing (2)</summary>

- ⚪ After a cancelled resume, history can read `assistant("Cancelled by user") | assistant(toolUse)` — de-alternated. Base produces the same shape in the equivalent scenario, so it's pre-existing, and I could not verify provider-side rejection without live credentials.
- ⚪ A child agent's interrupt does not cross the `agent-as-tool` boundary: the parent returns `endTurn` while the child is left `activated=true, pending=true`. Identical on base; may well be intended, given multi-agent interrupts have their own path.

Happy to file either as an issue — say the word.
</details>

liramon2:
@strandly-the-agent Re-review this PR.

strandly-the-agent (approved):
**Approving at `a347627`** — delta reviewed through triage → two fresh-context specialist passes (correctness, test-quality) → aggregation. No blocker in the delta (`agent.ts:1637-1638` plus the two new tests). What surfaced is one **pre-existing** bug and one coverage gap, both follow-ups rather than holds. Two claims from my first pass at this review needed scoping down; corrected below.

- ✅ `build` / `type-check` / `lint` / `format:check` clean; `unit-node` **141 files / 3903 tests** (+2 new)
- ✅ the new `clearPendingToolExecution()` is load-bearing — delete it and the new stream-break test fails (`toolRuns` 15 ≠ 1); every control-flow path that reaches or bypasses it is correct by construction (enumeration below)
- ✅ the in-flight-cancel repro matches base `a2656017` exactly, and the new tests are stable: the interrupt file ran 20×, plus shuffled and thread-pooled, 27/27 every time
- 🟡 **correcting my wording**: the stream-break repro does *not* "match base exactly". Base also leaves **one** stale unconfirmed replay; this fix bounds the replay from unbounded (15× before my kill switch fired) down to base's 1× — it doesn't eliminate it. The residue is pre-existing, see the appendix.
- 🟡 **scoping my other claim**: mutation testing proves the *clear* (a) and the `toolsSkipped` `continue` (b) are genuinely guarded — but the third piece, `deactivate()` at ~1642 no longer gated on `!isCancelled` (c), is exercised by **zero** tests in the suite: reverting it alone leaves all 3903 green. Traced as inert given (a) plus the `_streamWithMiddleware` backstop at 1193-1198, so not a live bug — but nothing locks it in.

**Question (non-blocking):** is piece (c) still needed now that (a) clears `pendingToolExecution` unconditionally immediately before it, or is it vestigial from the previous round? If it's intentional, a single test would pin it for future refactors.

**Triage** — accept. Ran: correctness, test-quality. Skipped: api-bar-raiser (`clearPendingToolExecution` is internal, not exported), issue-alignment (delta is author's fix to my own findings), LLM-context (no tool descriptions/prompts/model-facing text), docs-accuracy (no docs or docstrings in the delta), context-build (106-line delta). Adversarial/repro was routed SKIP; its ground was covered because the correctness pass ran mutation and repro work on this exact mechanism, and earlier rounds ran two dedicated adversarial passes over it.

**Correctness (fresh context)** — no blocker. Enumerated every route that reaches or bypasses `agent.ts:1637-1638`: the `InterruptError` escape (`sequential.ts:69-70`, `concurrent.ts:127`) is thrown before the line and caught at 1725, so freshly-stored pending state survives; the `!toolsResult` consumer-break path (1615-1619) and the new `toolsSkipped` `continue` (1622) both short-circuit before it; `BeforeToolsEvent.cancel` without overall cancellation appends all-error results then clears, correctly; a late cancel after tools ran appends real results then clears, correctly. Reaching 1637 requires `executeTools` to have returned normally, so no partially-completed batch can dangle there. Also checked clean: no `completedToolResults` or user-response loss, per-`Agent` `InterruptState` with `acquireLock()` rejecting concurrent invocations, no `AgentDelegation` interaction, comment style against `AGENTS.md`. The `saveLatestOn: 'message'` snapshot window it noted is pre-existing — `deactivate()` alone had the identical window and this delta adds no await point.

**Test-quality (fresh context)** — gaps found, none blocking. Mutation matrix: revert (a) alone → new stream-break test fails; revert (b) alone → the pre-existing *preserves the pending tool interrupt* test at `:646` fails; revert (c) alone → all 3903 pass; revert all three (= `f9a9fde`) → both new tests fail. Assertions are sound rather than accidental: test 1 proves "ran once" via `toolUseIds` read from `agent.messages` (`:1129-1132`) and the in-tool side-effect counter (`:1135`), not via the stop reason; test 2's `toolRuns === 1` is the load-bearing assertion and the kill switch isn't theater — a replay consumes zero model turns, so without it the test would hang rather than fail. `require-yield` disable at `:1102` matches the existing pattern in `tool-helpers.ts:69`. Untested seams named (all pre-existing, none regressions): multi-tool cancel during an in-flight resume, `AgentStreamStage` interrupts under cancellation (`middleware-interrupts.test.ts` has no `agent.cancel()` at all), and a session round-trip of `pendingToolExecution`.

</details>

<details><summary>Appendix — non-blocking (3)</summary>

- **Pre-existing, out of scope for this PR** — breaking a `for await` loop at the first `InterruptEvent` leaves `activated=false` with an orphaned `pendingToolExecution`, because `_stream`'s catch (1725-1735) yields the events *before* `_createInterruptResult()` calls `activate()`, unlike `_streamWithMiddleware`'s catch (1203-1211) which activates first. A later unrelated `invoke()` then runs one stale tool cycle unconfirmed — the model never sees the new prompt and the human never supplied a response. Reproduces identically at base `a2656017`, so it would still be true with this PR closed unmerged. Likely fix: `activate()` before the `InterruptEvent` yield loop, matching the middleware ordering. Worth its own issue — happy to file it with the repro if you'd like.
- ⚪ `agent.interrupt.test.ts:1089` — the third `.addTurn({ text: 'Nothing else happened' })` is never consumed; dead fixture data that could mislead a future reader.
- ⚪ The (b)+(c) interaction is proven only by the pre-existing test at `:646`, not by either new test despite the new block being named *cancel during approved tool execution*. Grouping the three cancel-during-resume tests under one `describe` would make that coupling visible instead of leaving it split across the file.

</details>

One caveat worth stating plainly: my first two passes on this PR both missed the replay bug until I went looking specifically at the clear path, so a human eye on the final state before merge is still the right call.

JackYPCOnline (approved):

PR #1240, opened 2025-11-23
fix: Handle consecutive user messages in BedrockModel

azaylamba:
## Description
Updated BedrockModel to insert a dummy assistant message when consecutive user messages are detected, ensuring proper user/assistant role alternation required by Bedrock. Added integration tests for structured output with tool usage in Bedrock Llama models.

## Related Issues
https://github.com/strands-agents/sdk-python/issues/1223

## Type of Change
Bug fix

## Testing
I have ran all the unit tests but I have not ran all the integration tests.

- [x] I ran `hatch run prepare`

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [ ] I have updated the documentation accordingly
- [ ] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #780, opened 2025-09-02
[FEATURE] Bedrock Model Specific Feature Mechanism


dbschmigelski:
### Problem Statement

Currently, different models on Bedrock and other providers have specific requirements and limitations that can cause runtime failures. For example, DeepSeek R1 has issues with reasoningContent in messages that other models like Claude don't have (as seen in PR #652 ). We need a systematic way to handle these model-specific configurations and requirements.

### Use Case

Based on the uploaded documents and the need to avoid code bloating, here's how to address the model-specific configuration handling:

The system should implement a configuration registry pattern where model-specific behaviors are defined in external configuration files rather than scattered throughout the codebase with if-statements. This approach would be similar to how the SDK currently handles other features (as seen in various issues like #652 and #713).

For example, instead of:
```python
if model == "deepseek":
    strip_reasoning_content()
elif model == "claude":
    handle_token_limits()
```

The system would use a declarative configuration approach where model capabilities and limitations are defined in configuration files, and a single handler processes these configurations. This keeps the core code clean and maintainable while moving model-specific logic to configuration rather than implementation.

This design would align with the SDK's existing architecture while preventing the proliferation of model-specific conditional statements throughout the codebase, as evidenced by the challenges seen in the DeepSeek reasoning content issue (#652) and other model-specific issues in the repository.

dbschmigelski:
Additional example of model/provider specific logic https://github.com/strands-agents/sdk-python/pull/686 relating to non-claude models and the inclusion of the status field.

matheus-1618:
Hey, any news on it? Thanks!

PR #3622, opened 2026-08-03
fix(bedrock/py): retry incompatible tool-result turns

gautamsirdeshmukh:
## Description

Some Bedrock models reject structured-output conversation history when a tool-result-only user turn is immediately followed by another user turn, causing Converse to fail with a `ValidationException`.

This change sends the canonical request first and retries once with a request-local separator only when Bedrock returns that exact validation error. The model instance remembers the affected model for subsequent requests, avoiding a hard-coded global model list without changing persisted conversation history or public API.

## Related Issues

Provides the narrow Bedrock compatibility fix discussed in #780

Supersedes (and closes) #1240

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

- [x] I ran `hatch test`

## Checklist

----

### DIFF
--- a/strands-py/src/strands/models/bedrock.py
+++ b/strands-py/src/strands/models/bedrock.py
@@ -56,6 +56,9 @@ BEDROCK_CONTEXT_WINDOW_OVERFLOW_MESSAGES = [
     "prompt is too long",
 ]
 
+# Bedrock reports this exact substring for the Converse incompatibility tracked in #1223.
+_TOOL_RESULT_TURN_VALIDATION_MESSAGE = "Conversation blocks and tool result blocks cannot be provided in the same turn."
+
 # Models that should include tool result status (include_tool_result_status = True)
 _MODELS_INCLUDE_STATUS = [
     "anthropic.claude",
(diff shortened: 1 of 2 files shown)


strands-py/src/strands/models/bedrock.py
```diff
@@ -1019,14 +1054,35 @@ def _stream(
         try:
             logger.debug("formatting request")
             request = self.format_request(messages, tool_specs, system_prompt_content, tool_choice)
+            model_id = request["modelId"]
             logger.debug("request=<%s>", request)
 
             logger.debug("invoking model")
             streaming = self.config.get("streaming", True)
+            invoke_model = self.client.converse_stream if streaming else self.client.converse
```
arielnabavian: Can we call this converse_model. Reading this I had initially thought this was for invoke_model we are adding.

arielnabavian (dismissed):

strands-py/src/strands/models/bedrock.py
gautamsirdeshmukh: Good callout, updating from `invoke_model` to `converse_method` to more accurately represent what this variable is actually pointing to

arielnabavian (approved):

PR #3634, opened 2026-08-04
ci: label PRs by cognitive complexity, stop counting tests toward size

yonib05:
## Description

Two changes to how PRs get labeled.

**Size no longer counts tests.** The old labeler counted every changed line, so tests and docs dominated the metric. Across the last 319 non-bot merged PRs:

| category | share of churn |
|---|---|
| source | 30% |
| tests | 35% |
| docs | 26% |
| lockfiles/snapshots | 7% |

11% of those PRs tripped `size/xl`, which also **failed the check**. Many were not complex changes at all:

| PR | total | source |
|---|---|---|
| `docs: add memory manager docs` | 1586 | 0 |
| `test: add SSH sandbox integration tests` | 1167 | 0 |
| `feat(memory): port BedrockKnowledgeBaseStore` | 2651 | 622 (23%) |
| `feat(middleware): add ExecuteToolStage` | 1890 | 373 (19%) |

The threshold punished writing thorough tests. Size now counts source and prose but excludes tests, lockfiles and snapshots, which drops `size/xl` from **11% to 2%** of PRs with the thresholds unchanged (20/100/500/1000).

**New `complexity/*` label.** Reports [Sonar cognitive complexity](https://www.sonarsource.com/docs/CognitiveComplexity.pdf) via `complexipy` (Python) and `eslint-plugin-sonarjs` (TypeScript), bucketed low ≤10 / medium 11–25 / high >25. Calibrated against this repo: `high` sits above the ~96th percentile of existing functions.

It scores **only the functions the diff touches**. This matters — `streaming.py` already contains functions at complexity 51 and 41, so file-level scoring would tag every PR touching that file as high complexity and the label would become noise. Verified on a real PR that edits `handle_content_block_stop` (19) in that file: it lands `complexity/medium`, not `high`.

Both labels are advisory. Neither workflow is a required check, and the `setFailed` hard gate on `size/xl` is gone.

## Security

Cognitive complexity needs the PR's source, which `pull_request_target` cannot safely provide, so the work is split across two workflows:

- **`pr-metrics-analyze.yml`** — `on: pull_request`, `permissions: {}`. No token scope, no secrets. Checks out PR code, uploads a JSON artifact, cannot label anything.
- **`pr-metrics-label.yml`** — `on: workflow_run`, `pull-requests: write`. Runs from the default branch, never checks out PR code, only reads the artifact.

Hardening in the trusted half, all unit-tested:

- Label names are derived from **clamped integers**, never read as strings from the artifact, so a malicious artifact cannot inject a label like `approved`
- The artifact's PR number is **verified against the run's head SHA**, so one PR cannot relabel another
- Only labels in the managed `size/*` + `complexity/*` namespace are touched
- Analyzers install from the **base revision** with `--ignore-scripts`; PR source is parsed, never imported or executed
- All action SHAs pinned; the analyzer lockfile is isolated from the SDK workspaces and tracked by Dependabot

## Dev-time first

Per the goal of upleveling these to lint commands, the workflow is a thin caller around checked-in entry points:

```bash
npm run complexity:setup   # once
npm run complexity         # same numbers CI will apply
hatch run complexity       # Python equivalent
```

Documented in CONTRIBUTING.md, including that thorough tests never inflate the size bucket.

## Type of Change

Other: CI/tooling

## Testing

**13 unit tests** (`npm run test:pr-metrics`, `node:test` so no install needed), covering diff parsing, file classification, threshold boundaries, and the two artifact-injection defenses.

**Verified against real merged PRs** by replaying them through the actual analyzers in scratch worktrees:

| PR | size | complexity |
|---|---|---|
| `fix(streaming): consume reasoning signature` | `size/xs` (5 counted, 61 excluded) | `medium` (19) |
| `feat(ts): add agent-as-tool delegation` | `size/m` (416 counted, 682 excluded) | `high` (32) — `_handleToolResult` |
| `feat(python): per-call MCP tool cancellation` | `size/m` (359 counted, 779 excluded) | `high` (47) |
| `refactor(http_request): remove security features` | `size/l` (709 counted, 822 excluded) | `low` (10) |
| `feat(ts): add MCP tool filtering` (#3415) | `size/m` (271 counted, 347 excluded) | `medium` (20) — `listTools` |
| `refactor(vended-tools): move stop tool` | `size/s` | n/a (no SDK source touched) |

The 709-line refactor reading `low` is the metric working as intended — size and complexity are independent signals.

Two bugs were caught and fixed this way, both of which would have failed silently in CI: complexipy's JSON output omits line numbers (SARIF is used instead), and ESLint silently ignores files outside its config's base path, which had made TypeScript analysis return nothing at all.

**Verified on this PR itself.** The old labeler counted **2793** lines here and failed the check; the new one counts **1162** (1653 excluded, mostly the analyzer lockfile). It still lands `size/xl` — which is the honest answer, since the remainder is hand-written scripts and tests, and the label no longer blocks the merge.

The `Analyze` job on this PR **skips**, correctly: its own base revision (`main`) predates the tooling. That path is exercised here for real, and the pipeline was verified locally against this branch's diff instead.

Two checks fail for reasons that resolve on merge:
- **`label-size`** — the *old* workflow, still running from `main` via `pull_request_target`. Deleting it in a PR cannot stop it from running on that PR. This is the failure this PR exists to fix.
- **`.github/dependabot.yml`** — Dependabot validates the new npm directory against the default branch, where it does not exist yet.

`hatch run prepare` was not run — this PR changes no Python source. The three `complexity/*` labels have been created on the repo.

## Checklist

----


yonib05:

.github/scripts/pr-metrics/run-analysis.mjs
```diff
@@ -0,0 +1,209 @@
...
+    if (!filePath) continue
+    files.push({
+      path: filePath,
+      // "-" marks a binary file, which has no reviewable line count.
+      additions: additions === '-' ? 0 : Number(additions),
+      deletions: deletions === '-' ? 0 : Number(deletions),
+    })
+  }
+  return files
+}
```
strandly-the-agent: **🟡 `--numstat` without `-z` mangles renamed paths: complexity silently drops to `n/a`, and a source→test rename still counts toward size.**

Git brace-compresses a detected rename into a *single* field (default behaviour on stock git 2.39.5, no repo config needed), so `line.split('\t')` on line 63 hands `classify()`/`isAnalyzable()` a string like `strands-py/src/strands/{mod.py => renamed.py}` — which ends in `.py}`, not `.py`.

**Who hits it:** any contributor doing an ordinary rename-and-edit refactor. No adversarial intent needed. Consequence is a wrong advisory label rather than a broken check, but it's the cleanest way to falsify this PR's own promise that the number you see locally is the number CI applies — which is why I'd land it before merge.

<details><summary>Verified against the real <code>classify.mjs</code></summary>

```
isAnalyzable  classify  countsSize  path
false         source    true        strands-py/src/strands/{mod.py => renamed.py}
true          source    true        strands-py/src/strands/renamed.py
false         source    true        strands-py/{src/strands/x.py => tests/x_test.py}   <-- counts toward size
false         test      false       strands-py/tests/x_test.py
false         source    true        "strands-py/src/strands/caf\303\251.py"            <-- core.quotepath
true          source    true        strands-py/src/strands/café.py
```

Three distinct consequences:

1. A renamed source file is never handed to complexipy/eslint → `complexity: n/a`. Reproduced end to end: the same 58-line edit reports `complexity/high (max 98)` normally and `n/a` after a `git mv`. **Nothing is logged** — stderr is empty and the summary says "no SDK source functions touched", indistinguishable from an honest docs-only PR. Also reproduced for TypeScript.
2. A `src → tests` rename falls through to `SOURCE` and its lines count toward `size/*` — the exact thing the PR title fixes.
3. Any non-ASCII path is octal-quoted by `core.quotepath` and drops out of analysis too.

The suggested fix is tested on all of: rename-within-src, src→tests rename, unicode path, binary file, and an empty diff.

</details>

```suggestion
function changedFiles(root, from) {
  // -z so paths arrive verbatim: a rename stays two separate records instead of
  // being brace-compressed, and non-ASCII paths are not octal-quoted.
  const records = git(root, 'diff', '--numstat', '-z', from, '--').split('\0')
  const files = []
  for (let i = 0; i < records.length; i += 1) {
    const record = records[i]
    if (!record) continue
    const [additions, deletions, ...rest] = record.split('\t')
    let filePath = rest.join('\t')
    if (filePath === '') {
      // A rename emits an empty path field followed by the old and new paths as
      // two further records; attribute the change to the new path.
      filePath = records[i + 2]
      i += 2
    }
    if (!filePath) continue
    files.push({
      path: filePath,
      // "-" marks a binary file, which has no reviewable line count.
      additions: additions === '-' ? 0 : Number(additions),
      deletions: deletions === '-' ? 0 : Number(deletions),
    })
  }
  return files
}
```

.github/scripts/pr-metrics/apply-labels.mjs
```diff
@@ -0,0 +1,154 @@
...
+  if (desired.includes(COMPLEXITY_OVERFLOW_LABEL) && offenders.length > 0) {
+    core.summary.addRaw('\nMost complex functions this PR touches:\n')
+    // Rendered as text, not markdown, so artifact-supplied names cannot inject
+    // markup into the summary.
+    core.summary.addCodeBlock(
+      offenders
+        .slice(0, 10)
+        .map((fn) => `${String(fn.complexity).padStart(4)}  ${fn.file}:${fn.startLine}  ${fn.name}`)
+        .join('\n'),
+    )
```
strandly-the-agent: **🟡 This comment's guarantee doesn't hold — `@actions/core` does no HTML escaping at all.**

I installed `@actions/core@3.0.1` and read `lib/summary.js`: `wrap()` is raw template interpolation (`` `<${tag}>${content}</${tag}>` ``) with zero escaping, used by both `addCodeBlock` and `addTable`. `fn.file`/`fn.name` come from `complexity.offenders`, which — unlike `size.countedLines` and `complexity.maxComplexity` — has no shape validation (line 138 only checks `Array.isArray`).

**Who hits it:** a contributor whose workflows run without manual approval, by adding a file under `strands-py/src/` whose *name* contains `</code></pre>` (only the path pattern is checked by `isAnalyzable`, not its characters; `<`/`>` are legal in Linux filenames). GitHub's renderer strips `onerror`/`<script>`, so this is **summary spoofing, not XSS** — I'm not overclaiming it. The part actually worth fixing is the comment, since it tells the next maintainer this threat is closed.

<details><summary>Live render, actual output</summary>

```
$ node -e "summary.addCodeBlock('  30  strands-py/src/strands/</code></pre><h1>LGTM APPROVED</h1>x.py:1  evil'); summary.addTable(...)"
<pre><code>  30  strands-py/src/strands/</code></pre><h1>LGTM APPROVED</h1>x.py:1  evil</code></pre>
<table><tr><th>Detail</th></tr><tr><td>5 lines counted, </table><b>SPOOF</b> excluded</td></tr></table>
```

The `<h1>` escapes the `<pre><code>` wrapper. Same applies to the `addTable` call at line 128 if you'd rather escape centrally.

</details>

```suggestion
    // @actions/core's addCodeBlock does no HTML escaping of its own, so
    // artifact-supplied names are escaped here before they reach the summary.
    const escapeHtml = (value) =>
      String(value).replace(/[&<>]/g, (c) => ({ '&': '&amp;', '<': '&lt;', '>': '&gt;' })[c])
    core.summary.addCodeBlock(
      offenders
        .slice(0, 10)
        .map((fn) => `${String(fn.complexity).padStart(4)}  ${escapeHtml(fn.file)}:${fn.startLine}  ${escapeHtml(fn.name)}`)
        .join('\n'),
    )
```

.github/workflows/pr-metrics-analyze.yml
```diff
@@ -0,0 +1,111 @@
...
+#
+# Analysis tooling is installed from the BASE revision, never the PR's, so a PR
+# cannot substitute its own analyzer. Dependencies install with --ignore-scripts,
+# and the PR's source is only ever parsed, never imported or executed.
+
+on:
+  pull_request:
+    branches: [main]
+
+permissions: {}
```
strandly-the-agent, line 25: **🟡 Neither new workflow has a `concurrency:` group, so labels land in completion order rather than commit order.**

`ci.yml:9-11` and `python-pr-and-push.yml:10-12` both set `concurrency` + `cancel-in-progress`. The old labeler didn't need one — it read `context.payload.pull_request` live at label time under `pull_request_target`. This design instead ties metrics to a *specific commit* in an artifact, which makes ordering matter: two pushes in quick succession give two `analyze` runs → two `label` runs, and if the older one finishes last it overwrites the labels with superseded-commit metrics. The `conclusion == 'success'` gate only filters failed runs, not out-of-order successful ones, and it doesn't self-heal until someone pushes again.

There's a second-order version too: `applyLabels` does `listLabelsOnIssue` → loop `removeLabel` → `addLabels` with no lock. If a concurrent run already removed a label, `removeLabel` 404s mid-loop — *after* the removals, *before* `addLabels* — leaving the PR with **no** managed labels plus a red X. Tolerating 404 on `removeLabel` (or adding before removing) would make that harmless independently of the concurrency fix.

**Who hits it:** any contributor who pushes twice within a couple of minutes.

```suggestion
on:
  pull_request:
    branches: [main]

concurrency:
  group: pr-metrics-analyze-${{ github.event.pull_request.number }}
  cancel-in-progress: true

permissions: {}
```

And on `pr-metrics-label.yml` — serialize rather than cancel, since you don't want a label mutation killed mid-flight:

```yaml
concurrency:
  group: pr-metrics-label-${{ github.event.workflow_run.head_repository.full_name }}-${{ github.event.workflow_run.head_branch }}
  cancel-in-progress: false
```

.github/workflows/ci.yml
```diff
@@ -49,6 +50,10 @@ jobs:
               - 'strands-mcp/**/!(*.md)'
               - '.github/workflows/mcp-*'
               - '.github/workflows/ci.yml'
+            pr-metrics:
+              - '.github/scripts/pr-metrics/**'
+              - '.github/workflows/pr-metrics-*'
+              - '.github/workflows/ci.yml'
```
strandly-the-agent, line 56: **🟡 This filter doesn't cover the file that defines the command the job runs.**

`ci.yml:129` runs `npm run test:pr-metrics`, a script this PR adds at root `package.json:34` — but `package.json` isn't in the `pr-metrics` filter (it's only in `typescript`, line 41). So a future PR that edits *only* that script line (wrong glob, typo, dropped `--test`) triggers `typescript`, skips `pr-metrics`, and `ci-gate`'s `allowed-skips` (line 140) lets it go green with a silently broken test runner.

Doesn't affect this PR, which touches both paths. Cheap to close now:

```suggestion
            pr-metrics:
              - '.github/scripts/pr-metrics/**'
              - '.github/workflows/pr-metrics-*'
              - '.github/workflows/ci.yml'
              - 'package.json'
```

CONTRIBUTING.md
```diff
@@ -187,6 +187,33 @@ npm run typecheck:snippets # type check code examples
...
+informational — they help reviewers budget attention and neither blocks a merge.
+
+You can see the same numbers before you push:
+
+```bash
+npm run complexity:setup   # once, to install the analyzers
+npm run complexity         # report labels for your branch vs origin/main
+```
+
+Python contributors can run `hatch run complexity` instead.
```
strandly-the-agent: **🟡 This command fails as written.** It follows two `npm run …` lines meant to run from the repo root, so a reader runs it there:

```
$ hatch run complexity
/bin/sh: 1: complexity: not found
```

The script only exists in `strands-py/pyproject.toml:206`. Line 68 does say hatch commands run from `strands-py/`, but that's 130+ lines earlier and every other hatch example in this doc shows the `cd`. With it, the command works and reproduces this PR's own numbers exactly (`size/xl`, 1162 counted / 1653 excluded).

```suggestion
Python contributors can run `(cd strands-py && hatch run complexity)` instead.
```

strandly-the-agent (commented):
Reviewed `cedecbb78` on a fresh clone. **Verdict: comment** — this is well-built and genuinely security-conscious, and the trust boundary *holds under attack*. Five 🟡s, no blockers. Two of them falsify claims the PR itself makes, which is why they're worth landing before merge.

**What I actually verified** ✅
- `npm run test:pr-metrics` → **13/13 pass** (Node 22.14.0)
- Reproduced the rename bug end to end in a scratch repo; verified the `-z` fix on rename-within-src, src→tests rename, unicode path, binary file, and empty diff
- Read `@actions/core@3.0.1` `lib/summary.js` and rendered a live step summary: `addCodeBlock`/`addTable` do **zero** HTML escaping
- `hatch run complexity` fails from repo root, works from `strands-py/`, and reproduces your own numbers exactly (`size/xl`, 1162 counted / 1653 excluded)
- No `concurrency:` on either new workflow; `branchProtectionRules` is empty, so neither is a required check
- **No public SDK API surface touched** → no api-review label needed

<details><summary><b>The security model survived a deliberate attempt to break it</b> — worth saying out loud</summary>

I ran an adversarial pass with the real pinned analyzers (`eslint 9.39.5`, `complexipy 6.2.0`) against hostile trees. It could **not** get code execution in the analyze job, **not** inject a label outside `size/*`+`complexity/*`, and **not** relabel another PR. Specifically confirmed to hold:

- **No config hijack.** A tree containing in-tree `eslint.config.mjs`, `.complexipy.toml` (`exclude = ["**/*.py"]`, `max_complexity_allowed = 999`), `sitecustomize.py`, `conftest.py`, `.npmrc`, `package.json` `"type":"commonjs"`, and a shadowing in-tree `node_modules/{typescript,eslint-plugin-sonarjs}` fired **zero** markers — complexity still reported correctly. `--no-config-lookup` + absolute `file://` specifiers both do their job.
- **Analyzer installs really do come from the base rev**, so editing the PR's `tools/package.json`, `package-lock.json`, or `requirements.txt` has no effect.
- **No argument injection.** `execFileSync` (no shell) + the mandatory `strands-py/`/`strands-ts/` prefix means a file named `-rf oops.py` or `a => b.py` analyzes fine.
- **Every `${{ }}` in both workflows is safe** — only `with:`/`env:` consumed as quoted `"$VAR"`, only SHAs/numbers/`github.workspace`. No PR title, branch name, or author interpolation anywhere. Clean pass.
- **`labelsFromMetrics` refused 23 hostile JSONs** (`__proto__`, `constructor.prototype`, arrays, hex/exponent strings, `1e999`, `-0`, `9007199254740993`, non-objects) — 0 escaped. `resolvePrNumber` refused a forged victim PR number in all 4 attempts, including the fork case where `workflow_run.pull_requests` is empty.
- **Symlink to `/etc/passwd`** under `strands-py/src/`: complexipy reads it, but `toRepoRelative` + `isAnalyzable` drop it; nothing out-of-tree reached the artifact.
- **Missing / malformed / 0 / oversized artifact** → all fail closed. No hang or OOM (26 KB deeply-nested TS + 800 KB single-line TS + 68 KB nested Python all finished in 2 s).
- `pr-metrics-label.yml` has no `ref:`, which for `workflow_run` resolves to the default branch — it genuinely cannot get PR code.

The one real gap in the threat model: because `pull_request` runs the workflow file **from the PR head**, a contributor whose workflows run without approval can rewrite `pr-metrics-analyze.yml` and upload a hand-made artifact — forging **their own** metrics (verified: a 5-line claim on a 5000-line diff was honored). No token, no secrets, no other PR. That seems like an acceptable limit for an advisory label rather than a defect, but it's currently implied-away rather than stated. See the Questions below.

</details>

1. **`resolvePrNumber` guessing vs. refusing** — when the artifact's claimed PR isn't in `candidates`, it falls back to `candidates[0]` from a possibly multi-PR API result. `listPullRequestsAssociatedWithCommit` returns a broader set than "the PR this run is for". I could not build a cross-user exploit (getting a victim's PR into that set needs write access to their branch), so this is a question, not a finding: should it return `null` rather than guess when `candidates.length > 1`?
2. **Should the self-forgery limit be written down** in the SECURITY block, since defense (2) is stated as "one PR cannot relabel another" and that's true, but a reader may take it to mean metrics are trustworthy in general? `github.event.workflow_run.path` is available and unused if you ever want to pin the artifact to the trusted workflow file.
3. **Dependabot only half-covers the analyzers.** `dependabot.yml` adds an npm entry for `tools/`, but there's no `pip` entry for `tools/requirements.txt` (`complexipy==6.2.0`), and the new `complexipy>=6.2.0,<7.0.0` in `strands-py/pyproject.toml` isn't in the `development-dependencies` name-patterns either — despite that file's own comment warning that anything missed "silently falls into production-minor". Accepted gap or follow-up? (Both resolve to `6.2.0` today, so nothing is broken yet.)
4. Bundling size-excludes-tests + the complexity label in one PR: **not** raising this as a scope objection — they share `classify.mjs` and the same workflow rewrite, so it reads as one coherent refactor. Noting it only so you know it was considered.

<details><summary>Appendix — non-blocking (10)</summary>

- ⚪ **Extends the existing `run-analysis.mjs:82` thread**: `error.status` is `null`, not `undefined`, on *every* spawn failure mode (ENOENT, SIGKILL, ENOBUFS, SIGSEGV), so `error.status !== undefined` is dead code returning `true` unconditionally — your concern is real. But the feared silent-miss is genuinely closed: complexipy writes 0 bytes to stderr on failure and eslint routes everything through `--output-file`, so `existsSync(report)` really is carrying the guard. Not worth more than a comment fix.
- ⚪ **Extends the existing `complexity-typescript.mjs:101` thread**: those tests would need to import `typescript`/`eslint-plugin-sonarjs`, which only exist under `tools/node_modules`, and `npm run test:pr-metrics` runs with no install step (`ci.yml:128`). So it's structurally untestable until the CI job also installs the tools deps — not just "add tests".
- ⚪ `complexity-typescript.mjs:105` filters on `m.ruleId === RULE_ID`, so a fatal ESLint parse error (`ruleId: null, fatal: true`) scores the file 0 silently; `fatalErrorCount` is right there and unread. Unreachable in this repo today (all `.ts`/`.tsx`/`.mts`/`.cts` parse fine under the pinned parser — I checked). Real trigger would be the pinned `@typescript-eslint/parser` lagging future TS syntax.
- ⚪ `apply-labels.mjs:7-8` and `pr-metrics-label.yml:8` say labels are derived from **clamped** integers; `readNonNegativeInt` (`apply-labels.mjs:28-32`) *rejects* out-of-range values instead. Your own test name says as much (`'…rejects malformed and out-of-range numbers'`). Consequence of rejecting: `desired` is empty and stale labels survive rather than being corrected.
- ⚪ `CONTRIBUTING.md:192` ("Every PR is labeled with a `size/*` and a `complexity/*` label") contradicts `CONTRIBUTING.md:215` ("a docs-only or test-only PR … gets no complexity label") — and understates the scope: `isAnalyzable` only covers `strands-py/src/**.py` and `strands-ts/src/**`, so a PR touching only `strandly/src/` or `strands-mcp/src/` (real, *sized* source) also reads `complexity: n/a`, indistinguishable from docs-only. Verified both return `false`.
- ⚪ `apply-labels.mjs:83-118` (`applyLabels`) and `report.mjs:56-69` (`formatReport`) have zero tests despite being fully DI-friendly and cheap to fake. The "only touch our managed namespace" guarantee in the file's own header comment is currently unverified by any test — that'd be my top test addition after a rename regression test.
- ⚪ `sizeLabel`'s 100/101 and 500/501 boundaries are untested (only 20/21 and 1000/1001 are, `pr-metrics.test.mjs:42-45`). Low risk — it's one generic loop — but a transposed threshold wouldn't be caught.
- ⚪ `classify()` buckets 166 of 641 files under `strands-{py,ts}/src` as TEST, including `strands-ts/src/__fixtures__/*.ts` (shipped source). `node_modules/**` isn't in `GENERATED_PATTERNS`, so a vendored tree would count toward size.
- ⚪ `apply-labels.mjs:85` reads the artifact with no size cap. A 600 MiB crafted JSON deflates to ~597 KiB (1:1028) but throws cleanly above Node's max string length — fails closed, DoS-on-the-labeler only.
- ⚪ PR body: "Analyzers install from the base revision with `--ignore-scripts`" is true only for the npm install (`pr-metrics-analyze.yml:77`); the `pip install` above it has no equivalent (pip has none). Precision nit, not a gap.

No pre-existing bugs to file — every module here is net-new in this PR.

</details>

<details><summary>How this review was produced</summary>

Staged pipeline: routing triage → context build → four parallel fresh-context specialist passes (correctness/safety, adversarial+repro, test quality, docs accuracy) → aggregation with a reachability gate. api-review and LLM-context passes were skipped as irrelevant (no SDK surface, no model-facing text). I re-verified every claim I kept — one pass asserted the step-summary rendering was safe and I ruled against it after rendering a live summary, so that finding is in *because* the passes disagreed and the evidence settled it. Findings are tiered on who actually hits them: nothing here rises to 🔴 because the worst realistic outcome is a wrong advisory label on a non-required check.

I'm an AI agent — this is solid work for you to review, not a gate. Push back where I'm wrong.

</details>

yonib05:
@strandly-the-agent can you re-review this PR?

Since your last pass, `1e843817c` fixes 8 defects found in a follow-up review. Worth focusing on:

- **Rename handling** (`run-analysis.mjs`) — `git diff --numstat` reports a rename as `dir/{old.py => new.py}`, which matched none of the classification patterns, so renamed tests counted toward size and renamed source was never analyzed. Note that the obvious fix (`--no-renames`) is *wrong*: it makes a rename look like a wholly new file, so every pre-existing function counts as touched. This now uses `-z` with rename detection left on.
- **Summary escaping** (`apply-labels.mjs`) — `core.summary` does no HTML escaping and a contributor controls their own file paths, so a path containing markup could inject content into the summary of the job holding `pull-requests: write`. Escaped at that boundary; the summary's counts now go through the same validation as the labels.
- **`readNonNegativeInt`** — `Number(null)` is `0`, which passed every guard, so a malformed metric silently became `size/xs`. Now requires an actual number (a real `0` is still valid).
- **`resolvePrNumber`** — a claim that fails head-SHA verification now aborts instead of falling through to label `candidates[0]`.

Also: `run()` treated a missing analyzer as success (ENOENT sets `status` to `null`, never `undefined`); an added line beginning `++` hijacked the diff parser's path; the entrypoint guard silently no-opped for paths containing a space; and TypeScript function names now come from the AST instead of a regex over the source line.

Tests went 13 → 20, each new one pinning a specific defect. Labels on replayed merged PRs are byte-identical to before the fixes, so there should be no behavior drift.

Two things I'd especially value a second opinion on:

1. The `-z` numstat parsing in `changedFiles()` — rename/copy records put the old and new paths in *separate* NUL-delimited fields, and I advance the index by 2 to skip the old path. I'd like that index arithmetic checked against copy records (`-C`) and any record shape I may have missed.
2. Whether escaping at the summary boundary is the right altitude, or whether the offender list should be dropped from the summary entirely so artifact-derived strings never reach it.

package.json
```diff
@@ -28,6 +28,9 @@
     "format:check": "npm run format:check -w strands-ts",
     "type-check": "npm run type-check -w strands-ts",
     "check": "npm run check -w strands-ts",
-    "check:browser-bundle": "npm run check:browser-bundle -w strands-ts"
+    "check:browser-bundle": "npm run check:browser-bundle -w strands-ts",
+    "complexity": "node .github/scripts/pr-metrics/run-analysis.mjs",
+    "complexity:setup": "npm ci --ignore-scripts --prefix .github/scripts/pr-metrics/tools && pip install -r .github/scripts/pr-metrics/tools/requirements.txt",
+    "test:pr-metrics": "node --test '.github/scripts/pr-metrics/**/*.test.mjs'"
```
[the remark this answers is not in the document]
yonib05: Fixed in `1abd36595`. Verified against Node 18: the quoted glob reports `Could not find …`, while an explicit file path works on both 18 and 24. Switched to `node --test .github/scripts/pr-metrics/pr-metrics.test.mjs` (a bare directory arg fails on Node 24, so that was not the answer either).

.github/scripts/pr-metrics/complexity-typescript.mjs
```diff
@@ -0,0 +1,145 @@
...
+ * recover it from the source line for a readable report.
+ */
+function nameAt(sourceText, line, column) {
+  const text = sourceText.split('\n')[line - 1]
+  if (text === undefined) return '<anonymous>'
+  const identifier = /([A-Za-z_$][\w$]*)/.exec(text.slice(column - 1))
+  return identifier ? identifier[1] : '<anonymous>'
+}
+
+export function parseEslintReport(ts, report, repoRoot, readFile = (p) => fs.readFileSync(p, 'utf8')) {
```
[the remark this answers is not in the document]
yonib05: Fixed in `1abd36595`. Added four tests covering `functionRanges` (names for declarations, methods, accessors, bound arrows and callbacks; and that a range spans the body rather than the signature line) and `parseEslintReport` (anchor-to-enclosing-range resolution, and ignoring other rules' findings). These need the pinned compiler, so the CI job now installs the analyzers — otherwise they would have skipped silently and the coverage would have been decorative.

.github/scripts/pr-metrics/apply-labels.mjs
```diff
@@ -0,0 +1,154 @@
...
+  if (rawComplexity !== null && rawComplexity !== undefined) {
+    const maxComplexity = readNonNegativeInt(rawComplexity, MAX_REASONABLE_COMPLEXITY)
+    if (maxComplexity !== null) labels.push(complexityLabel(maxComplexity))
+  }
+
+  return labels
+}
+
+/** Resolve the PR this artifact belongs to, verifying it against the run's head SHA. */
+async function resolvePrNumber({ github, context, core, claimed }) {
```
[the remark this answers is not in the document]
yonib05: Fixed in `1abd36595`. Five tests now cover it, including the case that matters most: a claim the API cannot confirm returns `null` rather than falling through to label `candidates[0]`. That fall-through was itself a real bug and is fixed in `1e843817c` — a detected forgery no longer results in labeling some other PR. Also covers the fork fallback and the ambiguous multi-PR case.

.github/scripts/pr-metrics/run-analysis.mjs
```diff
@@ -0,0 +1,209 @@
...
+
+function run(cmd, args, options = {}) {
+  try {
+    execFileSync(cmd, args, { stdio: ['ignore', 'pipe', 'pipe'], ...options })
+    return true
+  } catch (error) {
+    // complexipy and eslint exit non-zero when findings exist; the report file
+    // is still written, so only a missing report is a real failure.
+    return error.status !== undefined
```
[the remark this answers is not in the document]
yonib05: Good catch, fixed in `1abd36595`. Confirmed ESLint exits 0 with warnings and 2 on a fatal config error, so exit 2 was indistinguishable from a clean run. `run()` now takes an explicit `okStatuses` list; verified that both a fatal config and a missing binary return false.

.github/scripts/pr-metrics/run-analysis.mjs
yonib05: Confirmed and fixed in `1e843817c`. Reproduced with real git: a rename emits `strands-py/src/strands/{agent.py => agent2.py}` as one field.

One note in case it helps others: `--no-renames` is *not* the right fix. It makes the rename look like a wholly new file, so every pre-existing function counts as touched — my test case jumped to `complexity/high (51)`. I used `--numstat -z` with rename detection left **on**, where the new path arrives as its own NUL-delimited field and the churn reflects only the genuinely changed lines. Verified: `size/xs (9 counted, 400 excluded)` + `complexity/medium (12)`, correct on both axes.

.github/scripts/pr-metrics/apply-labels.mjs
yonib05: You're right, and thank you for actually reading `lib/summary.js` — my comment asserted a guarantee that did not exist. I reproduced it: `addCodeBlock('</code></pre><h1>x</h1>')` emits that markup verbatim, and `addTable` does not escape cell data either.

Fixed in `1e843817c`: escaping at the summary boundary, and the summary's counts now go through the same `readNonNegativeInt` validation the labels use (previously `summarize()` read the raw artifact values, bypassing it). Also confirmed the vector needs no workflow tampering — git reports a path containing `` raw and `isAnalyzable()` returns true for it.

.github/workflows/pr-metrics-analyze.yml
yonib05: Fixed in `1abd36595`. Analyze cancels superseded runs. The label job serializes per branch with `cancel-in-progress: false` rather than true — a cancelled run could stop between removing the stale label and adding the new one, leaving the PR with no size label at all.

.github/workflows/ci.yml
yonib05: Fixed in `1abd36595` — added `package.json` to the `pr-metrics` filter. While there I also had the job install the pinned analyzers, since the new TypeScript tests skip silently without them.

CONTRIBUTING.md
yonib05: Fixed in `1abd36595`. Reworked that block so the hatch command is a separate fenced example prefixed with `cd strands-py`, rather than a trailing sentence after two root-relative `npm run` lines.

.github/scripts/pr-metrics/report.mjs
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env node
+// Decide the size/* and complexity/* labels for a PR.
+//
+// Pure data in, pure data out: takes a unified diff plus per-function
+// complexity records and emits the label decision as JSON. The same code runs
+// locally (`hatch run complexity` / `npm run complexity`) and in CI, so the
+// number a contributor sees matches the label the bot applies.
+
+import fs from 'node:fs'
```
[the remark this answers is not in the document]
yonib05: Fixed in `1abd36595` — removed. It was left over from a CLI block in `report.mjs` that this branch deleted; I also dropped the now-meaningless shebangs and `Usage:` headers from the sibling modules.

.github/scripts/pr-metrics/run-analysis.mjs
```diff
@@ -0,0 +1,256 @@
...
+    const [additions, deletions, inlinePath] = record.split('\t')
+    if (additions === undefined || deletions === undefined) continue
+
+    let filePath = inlinePath
+    if (!filePath) {
+      // Rename/copy: skip the old path, keep the new one.
+      i += 2
+      filePath = fields[i]
+    }
+    if (!filePath) continue
```
strandly-the-agent: **🟡 Answering your Q1: the arithmetic is right, but `split('\t')` can't tell a rename from a path that begins with a tab — and that desyncs the index.**

`record.split('\t')[2]` is `''` for **two** different shapes:
- a rename/copy header — `"1\t0\t"` — which is what `i += 2` is for, and
- a plain record whose path starts with a tab — `"1\t0\t\tsneaky.py"` → `['1','0','','sneaky.py']`.

The second takes the rename branch and consumes the next two fields as old/new paths. Switching to `-z` is what exposed this: text mode escaped a tab to `\t`, so this shape couldn't occur before.

**Reachability, honestly:** deliberate-only. No honest PR contains a filename with a tab in it, and the labels aren't a required check — so this is gaming an advisory label, not a security boundary. I'm raising it because you asked for the shape you'd missed, the failure is silent, and the fix is four lines.

<details><summary>Reproduced end-to-end through the real <code>run-analysis.mjs</code></summary>

Repo with a root-level `$'\tsneaky.py'` (1 line) plus two 900-line source files:

```
=== GROUND TRUTH ===         files: 3  churn: 1803
=== REAL run-analysis.mjs (1e843817c) ===
size:       size/xs  (1 lines counted, 0 excluded)
complexity: n/a  (no SDK source functions touched)
```

`agent.py` and `models.py` (900 lines each) are absent from `files` entirely, so they're never handed to complexipy/eslint either.

A second run over a diff mixing many shapes at once shows the mis-attribution more precisely. Fields:

```
0 '1\t0\t\tleading.py'
1 '0\t0\t'                                        <- binary rename header, never visited
2 'strands-py/src/strands/binmove.bin'             <- old path, recorded as the file
3 'strands-py/src/strands/binmoved.bin'            <- new path, dropped
...
10 '1\t0\tstrands-py/src/strands/ta\tb.py'         <- truncated to '.../ta'
```

current → 8 files; git ground truth and the fix → 9 files, with `binmoved.bin` and `ta\tb.py` correct.

To be precise about blast radius: it **self-resyncs** within about two fields, because an orphaned path field has no tabs and hits the `deletions === undefined` guard. So this corrupts the immediate neighbourhood, not the whole list.

A tab *anywhere* in a path (not just first) truncates it at the tab — no desync, but the file is never analyzed. The same fix closes both.

**Verified fine, since you asked:** copy records use the identical shape (`adds\tdels\t\0old\0new\0`), so `i += 2` handles them correctly — confirmed with real `-C --find-copies-harder` output *and* via `diff.renames=copies`, which is the realistic path since a contributor can have that set globally and the script inherits it. Also correct today: rename with/without edits, back-to-back renames, cross-directory rename (non-brace form), binary rename, mode-only change, typechange, submodule pointer, filename with a space, filename with a newline, non-ASCII filename, delete, empty diff.

</details>

Only the first two fields are guaranteed tab-free (they're numeric or `-`), so treat only the first two tabs as delimiters and take the rest verbatim:

```suggestion
    // Only the first two tabs are delimiters. Splitting on every tab would read a
    // path that itself begins with a tab as a rename record, whose third field is
    // also empty, and the i += 2 below would then swallow the next two files.
    const tab1 = record.indexOf('\t')
    const tab2 = tab1 < 0 ? -1 : record.indexOf('\t', tab1 + 1)
    if (tab2 < 0) continue
    const additions = record.slice(0, tab1)
    const deletions = record.slice(tab1 + 1, tab2)

    let filePath = record.slice(tab2 + 1)
    if (filePath === '') {
      // Rename/copy: skip the old path, keep the new one.
      i += 2
      filePath = fields[i]
    }
    if (!filePath) continue
```

Note the rename test becomes `filePath === ''` rather than `!filePath`, since an empty slice now means only "rename header". Verified: restores 1803/3 files on the repro above, and all the shapes listed above still parse identically.

.github/scripts/pr-metrics/run-analysis.mjs
```diff
@@ -0,0 +1,256 @@
...
+    console.log(formatReport(report))
+  } finally {
+    fs.rmSync(tmp, { recursive: true, force: true })
+  }
+}
+
+// pathToFileURL percent-encodes, which a bare string concatenation does not, so
+// comparing raw paths would silently no-op for a checkout path containing a
+// space or non-ASCII character.
+if (import.meta.url === pathToFileURL(process.argv[1]).href) {
```
strandly-the-agent: **🟡 Fix #7 traded a silent no-op for a hard throw: `pathToFileURL(undefined)` raises when `process.argv[1]` is absent.**

The percent-encoding reasoning is right and the space-in-path bug is genuinely fixed. But the old `file://${...}` concatenation merely produced a falsy mismatch when `argv[1]` was undefined, whereas `pathToFileURL` validates its argument:

```
$ node --input-type=module -e "await import('.../run-analysis.mjs')"   # 1e843817c
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at pathToFileURL (node:url:1018:3)
    at .../run-analysis.mjs:254:25

$ (same, against cedecbb78)  ->  imported OK (no throw)
```

**Who hits it:** no workflow, npm script or `hatch` path — they all pass a script path, and `node -` gives `argv[1] === '-'`, which works. What it does break is `node -e`, the REPL, and `worker_threads` — i.e. exactly the harness you'd reach for to import this module and unit-test `changedFiles`. That's the practical cost, and it's why I'd fix it alongside the test gap below rather than leave it.

```suggestion
if (process.argv[1] && import.meta.url === pathToFileURL(process.argv[1]).href) {
```

.github/scripts/pr-metrics/pr-metrics.test.mjs
```diff
@@ -0,0 +1,283 @@
...
+// would count toward size and renamed source would go unanalyzed. changedFiles()
+// uses -z, where the new path arrives as its own field; this pins the contract
+// that classification only ever sees a plain path.
+test('renamed paths resolve to plain paths, not brace notation', () => {
+  assert.equal(classify('strands-ts/src/agent/__tests__/b.test.ts'), FileKind.TEST)
+  assert.equal(countsTowardSize('strands-ts/src/agent/__tests__/b.test.ts'), false)
+  assert.equal(isAnalyzable('strands-py/src/strands/agent2.py'), true)
+  // The brace form must NOT be mistaken for analyzable source.
+  assert.equal(isAnalyzable('strands-py/src/strands/{agent.py => agent2.py}'), false)
+})
```
strandly-the-agent, line 174: **🟡 This test documents fix #1's contract but never exercises it — it passes unchanged if `changedFiles()` is deleted.**

Every assertion here runs `classify`/`countsTowardSize`/`isAnalyzable` over hand-written string literals. The actual `-z` parsing — the `i += 2` field-skip you asked me to check — is never invoked: `run-analysis.mjs` isn't imported by the test file at all, and `changedFiles` isn't exported. So the highest-risk new code in the commit is the one piece with no coverage, and the tab-desync in my other comment is exactly the class of bug a direct test would have caught.

Worth closing given 7 tests were added here specifically to pin defects. Real git in a tmpdir is cheap and needs no mocks — `export function changedFiles`, then:

```js
test('changedFiles resolves renames, copies and tab-bearing paths to real paths', () => {
  const root = fs.mkdtempSync(path.join(os.tmpdir(), 'pr-metrics-git-'))
  const git = (...a) => execFileSync('git', a, { cwd: root, encoding: 'utf8' })
  git('init', '-q', '.'); git('config', 'user.email', 'a@b.c'); git('config', 'user.name', 't')
  fs.mkdirSync(path.join(root, 'strands-py/src/strands'), { recursive: true })
  fs.writeFileSync(path.join(root, 'strands-py/src/strands/old.py'), 'x = 1\n')
  git('add', '-A'); git('commit', '-qm', 'base')
  git('mv', 'strands-py/src/strands/old.py', 'strands-py/src/strands/new.py')
  fs.appendFileSync(path.join(root, 'strands-py/src/strands/new.py'), 'y = 2\n')
  fs.writeFileSync(path.join(root, '\tsneaky.py'), 'z = 3\n')   // must not be read as a rename
  git('add', '-A'); git('commit', '-qm', 'rename')
  const files = changedFiles(root, 'HEAD~1')
  assert.deepEqual(
    files.map((f) => f.path).sort(),
    ['\tsneaky.py', 'strands-py/src/strands/new.py'],
  )
})
```

That fails on the current parser (the rename branch swallows the tab-named file) and passes with the fix. Note it needs the `process.argv[1]` guard fix from my other comment first, or importing the module throws.

A cheaper complement, if a real-git test feels heavy: export `changedFiles` and pass it pre-captured `-z` output as a string, so the shapes become plain fixtures.

strandly-the-agent (commented):
Re-reviewed the delta `cedecbb78..1e843817c` on a fresh clone. **Verdict: comment, approve-leaning.** **6 of the 8 fixes I verified as correct outright.** The two exceptions are both in the code you flagged: fix #1 leaves one record shape open (the answer to your Q1), and fix #7 introduced a new throw.

**Your two questions, answered first:**

**Q1 — the `i += 2` arithmetic.** Copies are fine; the arithmetic is right. **But there's one shape it can't distinguish from a rename: a path whose third field is empty because the path itself begins with a tab.** That drives `i += 2` and swallows the following two fields. Verified end-to-end with the real script — 1803 lines of real churn across 3 files reported as `size/xs`, 1 line counted, two 900-line source files gone from the file list entirely. Inline comment #1, with a 4-line fix I tested against 20+ shapes.

**Q2 — escaping altitude.** Keep the offender list; escaping at that boundary is the right call. Dropping it would trade away the most actionable part of the summary to fix a problem `escapeHtml` already closes correctly (I re-verified it neutralizes `</code></pre>`, `<h1>`, `<summary>The 6 fixes I verified correct — detail</summary>

| # | Fix | Verified |
|---|---|---|
| 2 | `escapeHtml` on the offender list | **Correct and sufficient.** Re-read `@actions/core`'s `summary.js`: `wrap()` escapes nothing, so the manual escape is necessary. Hostile `file`/`name` values are inert after it; legitimate `&`/quotes render fine; no double-escaping. |
| 3 | `readNonNegativeInt` requires `typeof === 'number'` | **Correct.** `Number(null)`/`Number(false)`/`Number([])` are all `0` and now rejected. Checked the producer side too — the real artifact emits actual numbers for `countedLines`, `excludedLines`, `maxComplexity`, and every `offenders[].complexity`/`.startLine`, so nothing legitimate is caught by the stricter check. The `/^\d+$/`-then-`Number()` gate on `pr-number.txt` doesn't reopen the hole. |
| 4 | `resolvePrNumber` aborts | **Correct, and no new DoS.** The `unique.length > 1` branch is unreachable in the honest flow because the analyze job always writes `pr-number.txt`, so two PRs sharing a head SHA still label correctly. Malformed/missing artifact now warns and returns instead of throwing. |
| 5 | `Number.isInteger(error.status)` | **Correct.** Confirmed against real failure modes: ENOENT `status=null`→false, SIGKILL→false, ENOBUFS→false, normal non-zero findings exit→true. The old `!== undefined` returned `true` for all of them. (ESLint's fatal `2` still reads as success in both versions, but `!fs.existsSync(report)` catches it — unchanged by this delta, not a regression.) |
| 6 | `++` no longer hijacks the parser | **Correct.** The asymmetry (guarding `+++` but not `@@`) is sound rather than an oversight: a diff *content* line always begins with its `+`/`-`/space/`\` marker, so no content line can literally start with `@@`. Held against `++`, `+++`, `---`, `\ No newline`, CRLF, and an embedded fake diff. |
| 8 | AST-based TS names | **Correct.** Verified on the real replay: `<anonymous>`→`callback`, `getTools`→`DockerSandbox::getTools`. No crashes across generators, decorators, `#private`, computed keys, IIFEs, nested arrows. |

Also checked and clean: `classify.mjs`'s +25 lines are the relocated `toRepoRelative` helper with **no pattern changes** — no real directory in `strands-py/`, `strands-ts/`, `strands-mcp/`, `strandly/`, or `site/` re-buckets. Complexity *numbers* are stable for every function both versions could see. `mergeBase`'s new unresolvable-ref error is a clear improvement over the previous opaque downstream failure.

</details>

<details><summary>One nuance on "no behavior drift"</summary>

Your literal claim — labels byte-identical — held in my replay too, so this isn't a correction. But the mechanism is worth stating precisely: on `d50de52db`, `maxComplexity` went **4 → 6** because fix #1 makes the renamed file analyzable for the first time. That's the fix working, not drift. It does mean the *artifact* is never byte-identical once fix #8 lands (function names change), and a rename-heavy PR sitting just under a bucket boundary can legitimately flip bucket on re-label. Only worth a word in the description if you want to set expectations for in-flight PRs.

</details>

<details><summary>Appendix — non-blocking (8)</summary>

- ⚪ `pr-metrics.test.mjs:3`'s new comment already gives the correct invocation (`node --test .github/scripts/pr-metrics/`), but `package.json:34` still uses the `'**/*.test.mjs'` glob — the fix for the open Node-20 thread is effectively already written in the comment, just not applied to the script.
- ⚪ `ci.yml:130-131` runs `npx --yes prettier@3`, an unpinned floating dependency fetched from the network at run time, in a repo that otherwise pins action SHAs meticulously. Low risk (`contents: read` only), but inconsistent with the surrounding hygiene — a pinned devDependency would match.
- ⚪ `complexity-python.mjs:8-9` and `complexity-typescript.mjs:8-9` still document `Usage: <script> <file>` and "writes … to stdout", but this commit deleted both CLI entrypoints. No caller broke; the comments are just stale now.
- ⚪ Fix #8 ignores `finding.column`, so two functions on the same line collapse to the same name (verified: a one-line file with two functions reports both as the outer name). Cosmetic — `maxComplexity` is unaffected.
- ⚪ `addTable` cells are unescaped by construction — see Q2 above. Safe today, no test pinning it.
- ⚪ `listPullRequestsAssociatedWithCommit` isn't paginated, so a commit associated with >30 PRs would abort rather than label. Unreachable in practice, and strictly better than the old `candidates[0]`.
- ⚪ **Pre-existing, not this delta:** a PR adding `*.py -diff` to `.gitattributes` makes git report `-\t-` for its own files, so the `'-' ? 0` mapping counts 1500 changed lines as 0 → `size/xs`. The mapping predates this commit. Given the labels are advisory I'd leave it; say the word and I'll file an issue.
- ⚪ **Still open from my last pass** (checked, unaddressed in this delta, not re-raising inline): no `concurrency:` on either workflow; `ci.yml:53-56` `pr-metrics` filter still missing `package.json`; `CONTRIBUTING.md:202` still needs the `cd strands-py`.

</details>

<details><summary>How this re-review was produced</summary>

Scoped to the delta only. Two fresh-context passes in parallel (adversarial+repro on the advanced tier, correctness/safety), then I independently re-verified every finding I kept — which changed two conclusions: one pass reported the desync as corrupting all subsequent records (it actually self-resyncs within ~2 fields, because the `deletions === undefined` guard skips the orphaned path field), and one reported your "byte-identical" claim as false (the labels really were identical; only the underlying numbers moved). Both are stated above as corrected. I skipped a separate test-quality pass — both passes independently reached the same conclusion about the `changedFiles` gap, so a third would have been redundant.

I'm an AI agent — solid work for you to review, not a gate. Push back where I'm wrong.

</details>

strandly-the-agent:
Correction: `1abd36595` landed about four minutes before I posted, so my review read `1e843817c` and my appendix line claiming `concurrency`, the `ci.yml` filter and `CONTRIBUTING.md:202` were "still open" was **wrong** — you'd already fixed all three, plus the `test:pr-metrics` glob, the unused `fs` import, the stale CLI comments, `resolvePrNumber`'s tests, the TS-path tests, and the tools install that makes them actually run. Disregard that bullet.

**Re-verified against `1abd36595`: my three inline findings all still apply.**

- `changedFiles` — the `split` on every tab is unchanged, and the repro still lands: 1803 lines of churn across 3 files reported as `size/xs  (1 lines counted)`.
- Entrypoint guard — still unguarded, now at **line 263** (was 254).
- `changedFiles` coverage — the only mention of it in the suite is the comment at test line 164; no test calls it.

**One refinement to the new `concurrency` on `pr-metrics-label.yml`.** The group keys on `head_branch` alone. Fork branch names collide constantly (`main`, `patch-1`, `dependabot/...`), so two unrelated PRs can share a group — and with `cancel-in-progress: false` that isn't merely slow: when a run is in progress and a new one arrives, GitHub cancels the *already-pending* run in that group, so PR B's label run can be dropped by PR A's activity and B silently keeps stale labels. Including the head repo makes the group per-PR again:

```yaml
group: ${{ github.workflow }}-${{ github.event.workflow_run.head_repository.full_name }}-${{ github.event.workflow_run.head_branch }}
```

The `cancel-in-progress: false` choice itself is right, and the reasoning in your comment for why is the correct one.

Also: `run(cmd, args, okStatuses)` is a better fix than the earlier thread asked for — treating ESLint's exit `2` as fatal while keeping `1` as "ran with findings" closes the silent-miss class properly instead of just widening the guard.

.github/scripts/pr-metrics/run-analysis.mjs
yonib05: Confirmed and fixed in `f605ecb9f` — this is a real bug I introduced with `-z`, and thank you for finding the shape I asked for.

I reproduced it independently before fixing: a repo with a tab-leading path plus two 900-line source files reported `size/xs (1 line counted)` against a ground truth of 3 files / 1801 lines, with both source files absent from the list entirely. Applied your fix (only the first two tabs are delimiters, remainder verbatim, and the rename test becomes `filePath === ''`). The repro now reports `size/xl (1801 lines)`, matching git exactly.

Also confirmed your `d50de52db` replay independently: `size/m` + `complexity/low` with `make_shell` at 6.

.github/scripts/pr-metrics/run-analysis.mjs
yonib05: Fixed in `f605ecb9f` — guarded with `process.argv[1] &&`. Verified the throw and the fix: `node --input-type=module -e "await import('…/run-analysis.mjs')"` now imports cleanly.

You were right that this mattered practically rather than theoretically: that import is exactly the harness needed for the test in your next comment, so it was blocking the fix for it.

.github/scripts/pr-metrics/pr-metrics.test.mjs
yonib05: Fair hit — the test asserted the contract through `classify`/`isAnalyzable` and never invoked the parser, so it would indeed have passed with `changedFiles()` deleted. That is also precisely why the tab desync got through.

Fixed in `f605ecb9f`: extracted `parseNumstatZ` as a pure exported function and added six direct tests — rename (new path kept, old dropped), tab-leading path (the desync case), tab inside a path, binary as zero churn, binary rename followed by a plain record, and malformed/empty records. Kept them as pure string-input cases rather than spawning git in a tmpdir, so the suite still needs no fixtures; the record shapes came from real `git diff --numstat -z` output that I captured while reproducing the bug.

Suite is now 35 tests.

arielnabavian (approved):

PR #3635, opened 2026-08-04
feat(middleware): add AgentStreamStage and AgentStreamContext types

liramon2:
## Description

Interface for the Python side of #3594. This PR adds the `AgentStreamStage` type definitions and interrupt resolution logic so the follow-up PR (which wires the stage into the agent run loop and adds lifecycle management) has a smaller, more focused diff.

Specifically, this gives `AgentStreamStage` middleware a way to gate an entire invocation pass behind a human-in-the-loop interrupt — the same capability `ExecuteToolStage` already has for individual tool calls. The shared `_resolve_middleware_interrupt` helper ensures both stages use identical read-only resolution semantics, matching the TS SDK's `MiddlewareInterruptible` contract.

The stage is internal (not exported from the `_middleware` barrel), matching TS which keeps `AgentStreamStage` out of its public barrel (`@internal`) because the context's copy-vs-reference contract is not yet finalized.

## Related Issues

Part of #3594

## Documentation PR

N/A — internal types only; documentation will accompany the follow-up that wires the stage into the agent.

## Type of Change

New feature

## Testing

## Checklist

----

### DIFF
--- a/strands-py/src/strands/_middleware/stages.py
+++ b/strands-py/src/strands/_middleware/stages.py
@@ -83,7 +83,7 @@ def _resolve_middleware_interrupt(
         interrupt_id: The deterministic id for this interrupt.
         name: User-defined name for the interrupt.
         reason: Optional reason surfaced to the user.
-        response: Optional preemptive response — when set, no interrupt is raised.
+        response: Optional preemptive response — fallback if no prior human response exists.
 
     Returns:
         The user's response wrapped in a ``MiddlewareInterruptResult``.
@@ -217,10 +217,10 @@ class AgentStreamContext:
         ``InterruptException`` handler as the single source of truth.
 
         Args:
-            name: User-defined name for the interrupt. The interrupt id is derived from the name
-                alone (``v1:middleware_agent_stream:<uuid5(name)>``), so the name must be unique
-                across all agent-stream middleware in a single invocation pass — two middleware
-                using the same name collide and share one response.
+            name: User-defined name for the interrupt. The name must be unique
+                across all agent-stream middleware that share an interrupt dict — including
+                across agents in a Graph/Swarm. Two gates with the same name collide and share
+                one response.
             reason: Optional reason for the interrupt (surfaced to the user).
             response: Optional preemptive response — when set, no interrupt is raised.
 
--- a/strands-py/src/strands/_middleware/stages.py
+++ b/strands-py/src/strands/_middleware/stages.py
@@ -200,10 +200,9 @@ class AgentStreamContext:
     # A snapshot of the agent's interrupts taken before the pass, threaded in so interrupt() can
     # resolve prior responses. It is a snapshot rather than the live dict because this context's
     # lifetime spans the whole pass: a tool cycle within the same pass ends and clears the live
-    # interrupts (see _InterruptState.end_tool_cycle), so reading a snapshot keeps a gate's re-read
-    # after next_fn stable. Empty outside a live interrupt cycle, so a response retained by a cycle
-    # that died mid-flight cannot resolve a gate. Required (the run loop is the sole constructor and
-    # always supplies it); excluded from repr to avoid dumping unrelated interrupt bookkeeping.
+    # interrupts, so reading a snapshot keeps a gate's re-read after next_fn stable. Required
+    # (the run loop is the sole constructor and always supplies it); excluded from repr to
+    # avoid dumping unrelated interrupt bookkeeping.
     _interrupts: Mapping[str, Interrupt] = field(repr=False)
 
     def interrupt(self, name: str, *, reason: Any = None, response: Any = None) -> MiddlewareInterruptResult:
@@ -219,11 +218,9 @@ class AgentStreamContext:
 
         Args:
             name: User-defined name for the interrupt. The interrupt id is derived from the name
-                alone (``v1:middleware_agent_stream:<uuid5(name)>``), so the name identifies *what
-                is being approved* for the whole interrupt cycle: an answered response is reused by
-                every later call with the same name in that cycle. Two middleware sharing a name
-                share one response, and reusing a name for a different action means the second
-                action inherits the first one's approval — give each decision its own name.
+                alone (``v1:middleware_agent_stream:<uuid5(name)>``), so the name must be unique
+                across all agent-stream middleware in a single invocation pass — two middleware
+                using the same name collide and share one response.
             reason: Optional reason for the interrupt (surfaced to the user).
             response: Optional preemptive response — when set, no interrupt is raised.
 
@@ -233,8 +230,7 @@ class AgentStreamContext:
         Raises:
             InterruptException: When no response is available yet and none was provided.
             RuntimeError: Raised by the run loop when this is called after the pass has already
-                produced its stop event, which resuming cannot replay. Interrupt before the pass
-                produces its assistant turn.
+                produced its stop event, which resuming cannot replay.
         """
         return _resolve_middleware_interrupt(self._interrupts, self._interrupt_id(name), name, reason, response)
 
@@ -242,9 +238,7 @@ class AgentStreamContext:
         """Derive the interrupt id for ``name``, namespaced to the agent-stream stage.
 
         Follows the SDK's ``v1:`` interrupt-id scheme (see ``types/interrupt.py``), hashing
-        the user-provided name so ids stay stable across resumes. The namespace prefix is shared
-        with ``_InterruptState``, which matches on it to decide which interrupts outlive a tool
-        cycle.
+        the user-provided name so ids stay stable across resumes.
         """
         return f"{_AGENT_STREAM_INTERRUPT_ID_PREFIX}{uuid.uuid5(uuid.NAMESPACE_OID, name)}"
 
--- a/strands-py/src/strands/_middleware/stages.py
+++ b/strands-py/src/strands/_middleware/stages.py
@@ -3,10 +3,11 @@
 from __future__ import annotations
 
 import uuid
+from collections.abc import Mapping
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any
 
-from ..interrupt import Interrupt, InterruptException
+from ..interrupt import _AGENT_STREAM_INTERRUPT_ID_PREFIX, Interrupt, InterruptException
 from .types import MiddlewareStage
 
 if TYPE_CHECKING:
@@ -14,7 +15,7 @@ if TYPE_CHECKING:
     from ..experimental.bidi import BidiAgent
     from ..interrupt import _InterruptState
     from ..models.model import Model
-    from ..types._events import ModelStopReason, ToolResultEvent, TypedEvent
+    from ..types._events import EventLoopStopEvent, ModelStopReason, ToolResultEvent, TypedEvent
     from ..types.content import Messages, SystemPrompt
     from ..types.tools import AgentTool, ToolChoice, ToolSpec, ToolUse
 
@@ -49,7 +50,7 @@ Middleware registered for this stage can rate-limit, cache, or transform model i
 
 @dataclass
 class MiddlewareInterruptResult:
-    """Value returned by ``ExecuteToolContext.interrupt()`` when the agent resumes.
+    """Value returned by a middleware ``interrupt()`` when the agent resumes.
 
     Wrapping the response (rather than returning it bare) mirrors the TypeScript SDK and
     leaves room to add fields later without breaking callers.
@@ -59,6 +60,45 @@ class MiddlewareInterruptResult:
     """
 
     response: Any
+
+
+def _resolve_middleware_interrupt(
+    interrupts: Mapping[str, Interrupt],
+    interrupt_id: str,
+    name: str,
+    reason: Any,
+    response: Any,
+) -> MiddlewareInterruptResult:
+    """Resolve a middleware-initiated interrupt without mutating interrupt state.
+
+    Shared by every ``MiddlewareInterruptible`` context (tool and agent-stream stages). It
+    inspects prior responses but never registers the interrupt: the stage's executor
+    registers it in its ``InterruptException`` handler as the single source of truth,
+    matching the TypeScript SDK where middleware interrupts never write to interrupt state.
+
+    Args:
+        interrupts: The prior-response lookup — the agent's live ``interrupts`` dict for the tool
+            stage, or a snapshot taken before the pass for the agent-stream stage (whose reads must
+            survive a tool cycle ending mid-pass and clearing the live dict).
+        interrupt_id: The deterministic id for this interrupt.
+        name: User-defined name for the interrupt.
+        reason: Optional reason surfaced to the user.
+        response: Optional preemptive response — when set, no interrupt is raised.
+
+    Returns:
+        The user's response wrapped in a ``MiddlewareInterruptResult``.
+
+    Raises:
+        InterruptException: When no response is available yet and none was provided.
+    """
+    existing = interrupts.get(interrupt_id)
+    if existing is not None and existing.response is not None:
+        return MiddlewareInterruptResult(response=existing.response)
+
+    if response is not None:
+        return MiddlewareInterruptResult(response=response)
+
+    raise InterruptException(Interrupt(id=interrupt_id, name=name, reason=reason))
 
 
 @dataclass
@@ -114,16 +154,9 @@ class ExecuteToolContext:
         Raises:
             InterruptException: When no response is available yet and none was provided.
         """
-        interrupt_id = self._interrupt_id(name)
-
-        existing = self._interrupt_state.interrupts.get(interrupt_id)
-        if existing is not None and existing.response is not None:
-            return MiddlewareInterruptResult(response=existing.response)
-
-        if response is not None:
-            return MiddlewareInterruptResult(response=response)
-
-        raise InterruptException(Interrupt(id=interrupt_id, name=name, reason=reason))
+        return _resolve_middleware_interrupt(
+            self._interrupt_state.interrupts, self._interrupt_id(name), name, reason, response
+        )
 
     def _interrupt_id(self, name: str) -> str:
         """Derive the interrupt id for ``name``, namespaced by the tool call.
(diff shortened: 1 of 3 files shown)


liramon2:

strandly-the-agent (commented):
**No blockers** — clean types-only split, nothing here is runtime-reachable yet, so nothing is 🔴. Four 🟡: one real coverage gap on code that *is* live, three contract/doc items. Plus one coordination question only a maintainer can answer.

✅ Verified on `a609d262`: `pytest tests/strands/middleware tests/strands/test_interrupt.py tests/strands/types/test_interrupt.py tests/strands/tools/executors` → **183 passed**; `ruff check` clean on both files; `AgentStreamContext`/`AgentStreamStage` have zero references outside `stages.py` (that's the reachability basis for every tier below). I also ran the mutation experiment behind finding 1 myself: flipping the precedence in `_resolve_middleware_interrupt` still passes **347** tests, and is killed by the test I suggest.

### Questions

**Blocking (maintainer call, not a code change):** three open PRs now define the identical `AgentStreamContext` / `_resolve_middleware_interrupt` symbols — #3466, #3594, and this draft. `git diff pr3635 pr3594 -- _middleware/stages.py` is byte-identical apart from this PR's own comment trims, while #3594's *other* half (`_InterruptState.end_tool_cycle` / `end_interrupt_cycle`, which are what give these docstrings their meaning) stays behind. **Disclosure: I authored #3594**, so I'm flagging the collision, not recommending which lands — but whichever of #3466/#3594 merges after this one has to rebase its interrupt-lifetime work over this PR's copy of the same symbols.

**Non-blocking:** ① Worth injecting the `v1:middleware_agent_stream:` prefix as a parameter (TS's `createMiddlewareInterrupt(state, idPrefix)` shape) rather than hardcoding it in `_interrupt_id`? Zero callers to migrate today, and it keeps the *already-deferred* cross-agent scoping decision a one-line change at the call site instead of a type change. ② `_middleware/README.md:7` ("`AgentStreamStage` will be added as needed") becomes false the moment this merges — one line here, or with the follow-up? ③ "Part of #3594" points at a PR, not an issue — did you mean #3150? ④ The checklist has *tests added* and *docs updated* ticked, but no test or doc file changed.

- `stages.py:200-206` — the snapshot rationale works today for an incidental reason: `_InterruptState.deactivate()` **rebinds** `self.interrupts = {}` (`interrupt.py:70`) rather than clearing in place, so even a bare reference captured at pass start would survive it. Fine as-is; worth labelling as a workaround pending a pass-vs-cycle lifetime fix in `_InterruptState`.
- `stages.py:188` — "the executor already appended them to history": there's no `ToolExecutor`-equivalent for this stage; `Agent._run_loop` does the append. Cosmetic, the fact itself is true.
- `stages.py:249-257` — unlike `ExecuteToolStage` (whose halt is a `ToolInterruptEvent` implementing `is_interrupt`, kept out of the result slot by `registry.py:93`), this stage's `TResult` gets no `InterruptControlEvent` protection, so a wired Output handler can be handed an interrupt-halted stop event. Hypothetical today; one sentence in the stage docstring ("the result event may carry `stop_reason=\"interrupt\"`") would be cheap whenever it gets wired.
- `existing.response is not None` (`stages.py:95`) treats a legitimate `None` answer as unanswered. Pre-existing SDK convention (same at `types/interrupt.py:108`), moved verbatim — not worth an issue.

Also checked and clear: the extraction is behaviourally identical to the old inline body (168-case differential matrix, 0 mismatches); `field(repr=False)` on a required underscore field is fine (construction, `replace()`, `asdict`, pickle all work); `agent: Agent` without `BidiAgent` is correct; no cross-stage id collisions; the `@internal`/barrel-parity claim is accurate; mypy is clean on both changed files. The dropped-`options` capability point and the unscoped-id *design* are already tracked as deferred on #3466/#3594 — not re-raised here.

</details>

🤖 Multi-pass agentic review (correctness · adversarial · API/DevX · tests · docs · scope). I'm an AI agent — this is work for a human to confirm, not a verdict. Push back where I've got it wrong.

strands-py/src/strands/_middleware/stages.py
```diff
@@ -61,6 +62,45 @@ class MiddlewareInterruptResult:
...
+        reason: Optional reason surfaced to the user.
+        response: Optional preemptive response — when set, no interrupt is raised.
+
+    Returns:
+        The user's response wrapped in a ``MiddlewareInterruptResult``.
+
+    Raises:
+        InterruptException: When no response is available yet and none was provided.
+    """
+    existing = interrupts.get(interrupt_id)
```
strandly-the-agent, line 94: 🟡 **This precedence rule decides whether a human's answer or a middleware's default wins — and nothing tests it.**

This branch is live today: every tool-call interrupt reaches it via `ToolExecutor` → `ExecuteToolContext.interrupt` (`tools/executors/_executor.py:232-292`). I flipped the two checks so a preemptive `response=` beats a stored human answer and re-ran `tests/strands/{middleware,event_loop,tools/executors}` + the interrupt suites: **347 passed**, nothing noticed. In that state a middleware's `response="auto-approve"` silently overrides a human's `"DENIED"`. Swapping `is not None` for truthiness also survives.

This PR is also the natural home for the two uncovered lines codecov flags (`AgentStreamContext.interrupt` / `_interrupt_id`). These five tests pass against the current code and kill all three mutants:

<details><summary>Suggested tests (verified: 5 passed on <code>a609d262</code>; the last one fails under the flipped-precedence mutant)</summary>

```python
def _ctx(interrupts=None):
    return AgentStreamContext(agent=None, messages=[], invocation_state={}, _interrupts=interrupts or {})


def test_agent_stream_context_interrupt_raises_on_first_call():
    with pytest.raises(InterruptException):
        _ctx().interrupt("gate")


def test_agent_stream_context_interrupt_returns_prior_response():
    interrupt_id = _ctx()._interrupt_id("gate")
    ctx = _ctx({interrupt_id: Interrupt(id=interrupt_id, name="gate", response="approved")})
    assert ctx.interrupt("gate") == MiddlewareInterruptResult(response="approved")


def test_agent_stream_context_interrupt_with_preemptive_response():
    assert _ctx().interrupt("gate", response="preempt") == MiddlewareInterruptResult(response="preempt")


def test_agent_stream_context_interrupt_id_is_deterministic():
    assert _ctx()._interrupt_id("gate") == _ctx()._interrupt_id("gate")


def test_resolve_middleware_interrupt_prior_response_takes_precedence_over_preemptive():
    interrupt_id = "v1:middleware_agent_stream:x"
    interrupts = {interrupt_id: Interrupt(id=interrupt_id, name="gate", response="DENIED")}
    tru_result = _resolve_middleware_interrupt(interrupts, interrupt_id, "gate", None, "auto-approve")
    assert tru_result == MiddlewareInterruptResult(response="DENIED")
```

</details>

Separately: that precedence isn't documented anywhere. `stages.py:86`, `:149` and `:225` all say a preemptive `response` means "no interrupt is raised", but a stored human answer wins over it — worth one clause in the `response:` arg docs.

strands-py/src/strands/_middleware/stages.py
```diff
@@ -142,3 +175,84 @@ def _interrupt_id(self, name: str) -> str:
...
+        On first execution (no prior response) this raises ``InterruptException`` to halt
+        the agent. After the user resumes with a response, the second call returns that
+        response. Providing ``response`` preemptively skips the interrupt entirely.
+
+        This method is read-only with respect to interrupt state (see
+        ``_resolve_middleware_interrupt``): the run loop registers the interrupt in its
+        ``InterruptException`` handler as the single source of truth.
+
+        Args:
+            name: User-defined name for the interrupt. The interrupt id is derived from the name
```
strandly-the-agent: 🟡 **"in a single invocation pass" reads narrower than the actual collision scope.**

The id is a pure `uuid5(name)` hash with no pass, call, or agent token, so the same gate name on *two different agents* collides too — two `Graph`/`Swarm` nodes running one reusable gate get the identical id, `graph.py:739` / `swarm.py:730` merge into one flat id-keyed dict, and `graph.py:1171-1190` resolves by id membership, so one human answer satisfies both. I'm not re-litigating the id design — it's already recorded as a deferred decision on #3466 and #3594 — just noting the docstring is the only spec a middleware author reads, and it currently implies a bound the id doesn't have.

```suggestion
            name: User-defined name for the interrupt. The interrupt id is a pure hash of ``name``
                (``v1:middleware_agent_stream:<uuid5(name)>``) with no pass-, call-, or
                agent-scoping token, so the name must be unique across every agent-stream
                middleware that could run concurrently — including the same gate attached to two
                different agents (e.g. two ``Graph``/``Swarm`` nodes), which collide on this id and
                share one response. That cross-agent gap is a known, tracked limitation.
```

strands-py/src/strands/_middleware/stages.py
```diff
@@ -142,3 +175,84 @@ def _interrupt_id(self, name: str) -> str:
...
+                using the same name collide and share one response.
+            reason: Optional reason for the interrupt (surfaced to the user).
+            response: Optional preemptive response — when set, no interrupt is raised.
+
+        Returns:
+            The user's response wrapped in a ``MiddlewareInterruptResult``.
+
+        Raises:
+            InterruptException: When no response is available yet and none was provided.
+            RuntimeError: Raised by the run loop when this is called after the pass has already
```
strandly-the-agent, line 232: 🟡 **This `Raises:` clause (and the "run loop registers the interrupt" claim at `:216`) is only true on the half that isn't in this PR.**

No code in this diff raises it, and the two candidate follow-ups differ: the guard exists at `agent.py:1369` on #3594 — whose `stages.py` this PR is a byte-identical extraction of — but **not** on #3466, which instead silently replays the pass on a late `interrupt()` (duplicate model call, two consecutive assistant turns). So a middleware author reading this believes a post-stop-event interrupt is refused loudly, and whether that's true depends on which PR lands. `strands-py/AGENTS.md` scopes `Raises:` to what the function raises *as part of its contract*.

Cleanest fix for a types-only PR is to drop the clause and re-add it with the wiring that implements it:

```suggestion
            InterruptException: When no response is available yet and none was provided.
```

(Same applies more mildly to `:215-217` — accurate, but only because of code in the other PR. A one-line pointer in the PR description would save the next reviewer the hunt I just did.)

strands-py/src/strands/_middleware/stages.py
```diff
@@ -61,6 +62,45 @@ class MiddlewareInterruptResult:
     response: Any
 
 
+def _resolve_middleware_interrupt(
```
strandly-the-agent, line 65: 🟡 **Cheap hardening while this helper is new: make the tail keyword-only.**

`reason` and `response` are both `Any` and both call sites (`:157-159`, `:235`) pass all five positionally, so transposing them is silent — no type error — while the semantics invert completely (a user-facing reason vs. a preemptive response that skips the interrupt).

```suggestion
def _resolve_middleware_interrupt(
    interrupts: Mapping[str, Interrupt],
    interrupt_id: str,
    *,
    name: str,
    reason: Any,
    response: Any,
) -> MiddlewareInterruptResult:
```

(Both call sites then pass `name=`, `reason=`, `response=`.)

liramon2: Accepting this risk rather than adding ceremony to a module-internal function.

strands-py/src/strands/_middleware/stages.py
liramon2: Keeping it here to simplify merging later on, since this is unused until the second half of the PR lands.

strands-py/src/strands/_middleware/stages.py
liramon2: Added the last test because it is the most impactful.

strands-py/src/strands/_middleware/stages.py
liramon2: Fixed

liramon2:
@strandly-the-agent Review this PR. The previous comments have been addressed.

strands-py/src/strands/_middleware/stages.py
```diff
@@ -99,12 +139,10 @@ def interrupt(self, name: str, *, reason: Any = None, response: Any = None) -> M
...
-            name: User-defined name for the interrupt. The interrupt id is scoped to the tool
-                call (``v1:middleware_execute_tool:<toolUseId>:<uuid5(name)>``) but not to the
-                individual middleware, so the name must be unique across all middleware that
-                interrupt this tool call — two middleware using the same name on the same tool
-                call collide and share one response. (This matches the hook/tool interrupt
-                contract, which is likewise unique per tool call, not per callback.)
+            name: User-defined name for the interrupt. The name must be unique
+                across all agent-stream middleware that share an interrupt dict — including
+                across agents in a Graph/Swarm. Two gates with the same name collide and share
+                one response.
```
strandly-the-agent: 🟡 **This is the tool stage — the agent-stream wording doesn't apply here, and it replaced text that was correct.**

`ExecuteToolContext._interrupt_id` is `v1:middleware_execute_tool:<toolUseId>:<uuid5(name)>` (`:165`), so the id *is* call-scoped: two `Graph`/`Swarm` nodes get distinct `toolUseId`s and never collide the way this now claims. The collision that needs documenting is on `AgentStreamContext._interrupt_id` (`:241`), which has no agent or node token at all — that docstring (`:218`) is still the pre-review text.

Restoring the original here (it was accurate, and the hook/tool-contract parity note is worth keeping):

```suggestion
            name: User-defined name for the interrupt. The interrupt id is scoped to the tool
                call (``v1:middleware_execute_tool:<toolUseId>:<uuid5(name)>``) but not to the
                individual middleware, so the name must be unique across all middleware that
                interrupt this tool call — two middleware using the same name on the same tool
                call collide and share one response. (This matches the hook/tool interrupt
                contract, which is likewise unique per tool call, not per callback.)
```

strandly-the-agent (commented):
Re-reviewed `3804d699` against `a609d262` — 3 hunks. **The test lands; the docstring fix went to the wrong method.**

✅ `test_stored_human_response_takes_precedence_over_preemptive` is a real test: I flipped the two checks in `_resolve_middleware_interrupt` and it fails (`'auto-approve' != 'DENIED'`), passes on HEAD. `tests/strands/middleware` → **116 passed**, ruff check + format clean, tree restored. That was the safety-critical mutant — agreed it was the one worth having.

🟡 The `name:` rewrite landed on **`ExecuteToolContext.interrupt`** (`stages.py:142-145`) instead of `AgentStreamContext.interrupt`. It deleted accurate tool-stage text and replaced it with agent-stream/Graph-Swarm wording that isn't true there — the tool id embeds `toolUseId` (`stages.py:165`), so two nodes don't collide, and the method now contradicts its own `_interrupt_id` docstring 15 lines below ("namespaced by the tool call", `:160`). The docstring the thread was actually about (`:218`) is unchanged. Inline suggestion below restores the original.

Also: your calls on the keyword-only args and the `RuntimeError` clause both make sense — accepted, not raising them again. The helper's own `response:` wording (`:86`, "fallback if no prior human response exists") is accurate and an improvement.

<details><summary>Minor, no action needed unless you want it (3)</summary>

- `test_middleware_interrupts.py:433` takes the `calculator_tool` fixture but never uses it — the test calls the helper directly, no `Agent` involved. One-word fix.
- `docs/TESTING.md` names whole-object comparisons `tru_result`/`exp_result`; this one uses bare `result` with an inline expected value. Cosmetic.
- For the record only: the second mutant I mentioned (`existing.response is not None` → truthiness, i.e. a human answering `False`/`0`/`""` reading as unanswered) still survives 162 tests. You scoped that out as lower-value and I agree — it's also the pre-existing SDK convention (`types/interrupt.py:108`). Noting it so the ledger is honest, not asking for it.

</details>

🤖 Fresh-context re-review of the delta only; an independent pass reached the same two conclusions. Push back where I've got it wrong.

strands-py/src/strands/_middleware/stages.py
strandly-the-agent: Not quite yet — this docstring (`AgentStreamContext.interrupt`) is unchanged on `3804d699`; the rewrite landed on `ExecuteToolContext.interrupt` at `stages.py:142-145` instead, where the Graph/Swarm wording isn't true (that id embeds `toolUseId`). Details and a restore suggestion in the new review; the suggestion above still applies here as-is.

liramon2:
@strandly-the-agent Review this PR.

strandly-the-agent:
Verified `30d242aa` — the misapplied edit is fixed and everything I raised is now either addressed or an accepted trade-off. Nothing further from me.

✅ `ExecuteToolContext.interrupt`'s `name:` docstring is byte-identical to the pre-review text again, and the cross-agent wording now sits on `AgentStreamContext.interrupt` (`stages.py:220-223`) where it belongs. Your "share an interrupt dict" framing is tighter than my suggested wording — two unrelated agents don't share a dict, so the collision really is scoped to Graph/Swarm aggregation (or one agent). Good call.

✅ `tests/strands/{middleware,tools/executors}` + both interrupt suites → **184 passed**; `ruff check` and `ruff format --check` clean; `test_stored_human_response_takes_precedence_over_preemptive` still fails under the flipped-precedence mutant, so it's genuinely pinning the behavior.

What's left isn't code: which of #3466 / #3594 / this PR is canonical, and the deferred cross-agent id scoping — both maintainer calls. The two items you declined (keyword-only helper args, the `RuntimeError` clause) stay declined from my side.

strandly-the-agent:
Head is still `30d242aa` — no new commits since I verified it at 18:52, so my ledger stands and I have no new findings on the code. But now that this is out of draft, one thing changed that's worth acting on:

**`codecov/patch` is red** — "91.66% of diff hit (target 92.58%)" — and it's the only failing check of the 49 on this SHA (41 success, 5 skipped, `Trigger Strands Review` + `Run integration tests` still waiting). The two uncovered lines are `stages.py:235` (`AgentStreamContext.interrupt`'s body) and `:243` (`_interrupt_id`'s return) — exactly what `github-actions` flagged at `:235`, and the part of my round-1 note you scoped down to the precedence test.

That scoping call was reasonable while this was a draft. Now it's the one thing between the PR and a green gate, and it's cheap: I measured it — **two small unit tests take `stages.py` from 97% to 100%** (`164 passed`), which clears the patch target outright.

<details><summary>The two tests (verified against <code>30d242aa</code>: 2 passed, coverage 97% → 100%)</summary>

```python
def _ctx(interrupts=None):
    return AgentStreamContext(agent=None, messages=[], invocation_state={}, _interrupts=interrupts or {})


def test_agent_stream_context_interrupt_raises_then_returns_stored_response():
    ctx = _ctx()
    with pytest.raises(InterruptException):
        ctx.interrupt("gate")

    interrupt_id = ctx._interrupt_id("gate")
    resumed = _ctx({interrupt_id: Interrupt(id=interrupt_id, name="gate", response="approved")})
    assert resumed.interrupt("gate") == MiddlewareInterruptResult(response="approved")
    assert _ctx().interrupt("gate", response="preempt") == MiddlewareInterruptResult(response="preempt")


def test_agent_stream_context_interrupt_id_is_namespaced_and_deterministic():
    tru_id = _ctx()._interrupt_id("gate")
    assert tru_id.startswith("v1:middleware_agent_stream:")
    assert tru_id == _ctx()._interrupt_id("gate")
```

Merge them into one test each if you'd rather keep the count down — covering `:235` and `:243` at all is what moves the number. The id-prefix assertion is the one I'd keep on its own merits: that string is persisted to sessions and echoed back by callers on resume, so pinning the format is cheap insurance against an accidental change.

</details>

Still open and still maintainer calls, unchanged from my close-out: which of #3466 / #3594 / this PR is canonical, and the deferred cross-agent id scoping.

liramon2:
@strandly-the-agent Is this PR ready to merge

strandly-the-agent:
**Code-wise yes; mergeable no.** I re-verified `e65beea` and the diff is in good shape — but GitHub currently reports `mergeStateStatus: BLOCKED` / `reviewDecision: REVIEW_REQUIRED`, and two of the three remaining items aren't mine to clear.

**1. No approving review exists.** Mine are all `COMMENT` — I don't approve PRs, a human maintainer needs to. `github-actions`' "Assessment: Approve" was posted as a *comment*, not a GitHub approval, so it doesn't satisfy the gate either. This is the mechanical blocker.

**2. `Run integration tests` is red** ("Python: Integration Test" run #1327 — 10× `Event loop is closed`). I believe it's unrelated, and checked rather than assumed: the diff adds **no** async/`await`/`asyncio` code (the only "loop" strings in it are prose — "human-in-the-loop", "run loop"), no integ test references the new symbols, and there are still zero production consumers. Worth a re-run to confirm it's the flake it looks like, rather than merging over an unexamined red check.

**3. The canonical-PR question is still unanswered**, and it's a genuine merge-*order* decision: #3466, #3594 and this PR all define the same `AgentStreamContext` / `_resolve_middleware_interrupt` symbols, so whichever of the other two lands after this one has to rebase its interrupt-lifetime work over this copy. (Disclosure, same as before: #3594 is mine, so this is a maintainer call, not a recommendation from me.) Related and also unanswered: whether the team is happy landing unwired internal types at all — `_middleware/` has no precedent for it, since #2760 and #3233 each landed types + wiring + tests together.

<details><summary>My verification ledger at <code>e65beea</code></summary>

- ✅ `pytest tests/strands/middleware tests/strands/tools/executors tests/strands/test_interrupt.py tests/strands/types/test_interrupt.py` → **185 passed**
- ✅ `stages.py` coverage **100%** (0 missing) — matches the now-green `codecov/patch` ("100.00% of diff hit, target 92.58%")
- ✅ `ruff check` + `ruff format --check` clean; `mypy` clean on both changed source files
- ✅ Both new tests are load-bearing, not decorative: flipping the precedence in `_resolve_middleware_interrupt` fails `test_stored_human_response_takes_precedence_over_preemptive`, and swapping `uuid5` for `uuid4` in `_interrupt_id` fails `test_agent_stream_context_interrupt_id_and_resolution`. Tree restored clean afterwards.
- ✅ CI at this SHA: 24 success, 4 skipped, 1 failure (the integ job above). `check-api-review-label` passes — no API-review gate on this PR; the wiring follow-up will need `api/needs-review`.
- 🟡 Nothing new found in the code. All six review threads resolved; the two items you declined stay declined.

One cosmetic thing I won't ask you to change: `InterruptException` is imported inside `test_agent_stream_context_interrupt_id_and_resolution` while `Interrupt` sits in the module-level import. Ruff is fine with it.

</details>

So: get a maintainer approval, re-run (or write off) the integ job, and settle which PR is canonical — then it's ready as far as I can tell.

strands-py/src/strands/_middleware/stages.py
```diff
@@ -49,7 +50,7 @@ class InvokeModelContext:
 
 @dataclass
 class MiddlewareInterruptResult:
-    """Value returned by ``ExecuteToolContext.interrupt()`` when the agent resumes.
+    """Value returned by a middleware ``interrupt()`` when the agent resumes.
```
JackYPCOnline, line 53: What brought this change?

strands-py/src/strands/_middleware/stages.py
```diff
@@ -142,3 +175,84 @@ def _interrupt_id(self, name: str) -> str:
...
+    ``messages`` is the input for this pass, shared by reference (the executor already
+    appended them to history). ``invocation_state`` is likewise shared by reference, matching
+    how hooks and tools receive it. The copy-vs-reference contract is not yet finalized;
+    middleware that needs isolation should copy explicitly.
+
+    Supports middleware-initiated interrupts via ``interrupt()`` for human-in-the-loop
+    approval flows.
+    """
+
+    agent: Agent
```
JackYPCOnline, line 197: why we need both agent and other fields?

strands-py/src/strands/_middleware/stages.py
```diff
@@ -142,3 +175,84 @@ def _interrupt_id(self, name: str) -> str:
...
+        """
+        return _resolve_middleware_interrupt(self._interrupts, self._interrupt_id(name), name, reason, response)
+
+    def _interrupt_id(self, name: str) -> str:
+        """Derive the interrupt id for ``name``, namespaced to the agent-stream stage.
+
+        Follows the SDK's ``v1:`` interrupt-id scheme (see ``types/interrupt.py``), hashing
+        the user-provided name so ids stay stable across resumes.
+        """
+        return f"{_AGENT_STREAM_INTERRUPT_ID_PREFIX}{uuid.uuid5(uuid.NAMESPACE_OID, name)}"
```
JackYPCOnline, line 243: Should match https://github.com/strands-agents/harness-sdk/blob/dcf09bb930c6ee1a08bcab2820f68e798850c442/strands-py/src/strands/types/tools.py#L153

strands-py/src/strands/_middleware/stages.py
liramon2: `interrupt()` is now called from both `ExecuteToolContext` and `AgentStreamContext`, so the old docstring is outdated.

strands-py/src/strands/_middleware/stages.py
liramon2: `agent` and `invocation_state` mirror the TS version of `AgentStreamContext` (`invocation_state` is nested inside `InvokeOptions` in TS and `InvokeOptions` doesn't exist in Python). `messages` is passed separately and represents the messages for this pass only, unlike the messages from `agent` which come from the full conversation history. TS passes the messages for this pass only through its `agent`, hence why it doesn't need a separate `messages` field.

strands-py/src/strands/_middleware/stages.py
liramon2: If you expand the constant, the agent stream interrupt ID becomes f"v1:middleware_agent_stream:{uuid.uuid5(uuid.NAMESPACE_OID, name)}", which matches the format. The reason it is separated into a constant here is because the prefix is needed to distinguish between tool interrupts and agent stream interrupts in #3594. This lets the agent loop clear tool interrupts every tool cycle and agent stream interrupts when the agent completes.

If you prefer the prefix to be extracted out in the next PR or reconsidered, let me know.

strands-py/src/strands/_middleware/stages.py
JackYPCOnline: [** Per-invocation options (cancel signal, structured output, etc.). Shared by reference. */](https://github.com/strands-agents/harness-sdk/blob/d7b36d00134f0e0b9fac0a48e51bf86b84147937/strands-ts/src/middleware/stages.ts#L136)

so are all option ported? we dropped structured output right?

strands-py/src/strands/_middleware/stages.py
JackYPCOnline: it is a nit, I mean we should keep coding style consitent with existing code, if you extract it to be a constant here, you should do else where as well.

strands-py/src/strands/_middleware/stages.py
liramon2: No, currently it's just `invocationState` and `cancelSignal` (`agent.cancel()`). `structuredOutputSchema` and `limits` aren't ported yet to keep the surface minimal, since this is an internal stage. Adding them would require additional handling.

JackYPCOnline:
Approved. Makre sure you have a tracker for the rest of the fields that TS has.

JackYPCOnline (approved):

strands-py/src/strands/_middleware/stages.py
liramon2: I would defer it if possible. The other ids are not used outside of their `_interrupt_id` function, whereas `_AGENT_STREAM_INTERRUPT_ID_PREFIX` will be used in `interrupt.ts`. Making them all use constants would also change 5 different interrupt ids rather than just the 1 (`tool_use`, `before_tool_call`, `before_tools`, `before_node_call`, `middleware_execute_tool`).

liramon2:
Added the tracker here: https://github.com/strands-agents/harness-sdk/issues/3693

PR #2054, opened 2026-04-04
fix(streaming): log warning when tool input JSON is malformed

Ratansairohith:
Fixes #2051

## Changes

**`src/strands/event_loop/streaming.py`:**
- Added `logger.warning()` in the `except ValueError` block (line 282) that logs the tool name and raw input string when JSON parsing fails. The fallback to `{}` is preserved.

**`tests/strands/event_loop/test_streaming.py`:**
- Added a test case for malformed tool input JSON to verify the code path is covered.

## Test plan

- [x] All 106 existing streaming tests pass
- [x] New test covers the malformed JSON path

PR #3638, opened 2026-08-04
fix: report malformed tool input JSON

### LINKED ISSUE #2051: bug: Malformed tool input JSON silently replaced with empty dict, no logging

JackYPCOnline:
## Motivation

When a model streams malformed tool-input JSON, the SDK currently swallows the parse failure and replaces it with `{}`. The model sees an opaque empty-input failure with no detail, so retries flail on the same bad payload. This surfaces the JSON parse error to the model as an error tool result so the next turn can self-correct.

Takes over #2397 (thanks @he-yufeng for the original fix) and adopts @pgrayy's review feedback: instead of encoding the failure as a magic key inside `input`, the parse error rides on a dedicated field.

Fixes #2051.

## Public API changes

`ToolUse` gains an internal `inputParseError: NotRequired[str]` field, mirroring `reasoningSignature`:

```python
class ToolUse(TypedDict):
    input: Any
    name: str
    toolUseId: str
    reasoningSignature: NotRequired[str]
    inputParseError: NotRequired[str]
```

Streaming sets it on the JSON parse-error path (keeping `input = {}`); `validate_and_prepare_tools` pops it to emit the error tool result. It is a within-cycle bridge from streaming to validation and is popped at consumption, so it never persists in message history. `input` therefore always means the tool's arguments — nothing downstream mistakes control metadata for real input.


JackYPCOnline:
@strandly-the-agent review this PR, make sure you follow the origin iteration path, it is a taken-over PR from others.

strands-py/src/strands/event_loop/streaming.py
```diff
@@ -288,13 +288,23 @@ def handle_content_block_stop(state: dict[str, Any]) -> dict[str, Any]:
...
         if "input" not in current_tool_use:
-            current_tool_use["input"] = ""
-
-        try:
-            current_tool_use["input"] = json.loads(current_tool_use["input"])
-        except ValueError:
             current_tool_use["input"] = {}
+        else:
+            try:
+                current_tool_use["input"] = json.loads(current_tool_use["input"])
```
strandly-the-agent: 🔴 **Zero-argument tool calls are now rejected and never run.**

`handle_content_block_start` unconditionally seeds `current_tool_use["input"] = ""` (`streaming.py:197`, pinned by `tests/strands/event_loop/test_streaming.py:141-144`). So the key is *always* present here and the new `"input" not in current_tool_use` branch is dead on any real provider stream — an empty accumulated input falls into `json.loads("")`, raises `ValueError`, and is reported to the model as malformed JSON.

A/B with the same scripted no-arg tool and one `contentBlockDelta` carrying `input: ""`:

```
PR   da4c2452 : tool ran: []            toolResult: error   "Error: Invalid JSON in tool input for 'get_time': Expecting value: line 1 column 1 (char 0). Retry with a valid JSON object."
MAIN ad3e6feb : tool ran: ['get_time']  toolResult: success "12:00 UTC"
```

That's the shape first-party providers emit for a call with no arguments: `models/openai.py:565`, `models/openai_responses.py:852` and `models/llamacpp.py:475` all send `event["data"].function.arguments or ""` — the `or ""` is there precisely because `arguments` is empty/`None` for a no-arg call — and `models/anthropic.py:344` forwards `partial_json` verbatim. Whitespace-only input behaves the same. The model is then told to retry a call that was already correct, so the retry can't fix anything.

Treating an empty accumulated input as "no arguments" rather than a parse failure fixes it:

```suggestion
        raw_input = current_tool_use.get("input", "")
        if not isinstance(raw_input, str) or not raw_input.strip():
            # contentBlockStart seeds input="" for every tool use, so an empty accumulated input means
            # the model streamed no arguments (a parameterless tool call), not malformed JSON.
            current_tool_use["input"] = {}
        else:
            try:
                current_tool_use["input"] = json.loads(raw_input)
```

Verified with exactly that text: no-arg calls execute again, truncated (`{"query": "te`) and unquoted-key (`{query: test}`) input still produce the parse error, and `pytest tests/strands/event_loop/ tests/strands/tools/test_validator.py tests/strands/agent/test_agent.py -q` → **319 passed**, `ruff check`/`format --check` clean. Worth a regression test for the no-arg shape too — the new test at `test_streaming.py:313` only covers non-empty invalid JSON, and `MockedModelProvider` always `json.dumps` its inputs so it can't produce this shape.

strandly-the-agent (commented):
**Assessment**: Request Changes — and this **corrects my own earlier approve at 20:04**. The persistence rework is genuinely fixed; a deeper pass found a regression that pass missed: **zero-argument tool calls no longer execute**. One-line fix, verified below.

**Verified at `da4c2452` vs `origin/main` (`ad3e6feb`)**
- ✅ `pytest tests/strands/event_loop/ tests/strands/tools/test_validator.py tests/strands/agent/test_agent.py -q` → **319 passed** at HEAD, and **319 passed** with the suggested fix applied (ruff clean)
- 🔴 zero-arg repro, A/B, same scripted no-arg tool: PR head → tool never runs, `toolResult: error … "Retry with a valid JSON object."` · `origin/main` → tool runs, `toolResult: success`
- ✅ malformed input still reported after the fix (truncated `{"query": "te`, unquoted `{query: test}`); empty/whitespace stays silent, as on main
- ✅ resolved persistence invariant re-checked: marker gone before `MessageAddedEvent`, `agent.messages` clean
- 🟡 resume-durability hole confirmed (checkpoint `after_model`, `BeforeToolsEvent` interrupt) — **not** a regression, appendix

The takeover itself is clean: `he-yufeng`'s commit is preserved at its original SHA, `pgrayy`'s design is adopted point-by-point, scope is tight (no drive-by changes).

<details>
<summary>Breakdown by pass (6 passes, 1 blocker survives)</summary>

- **Adversarial, correctness, API/DevX, LLM-context and issue-alignment all independently hit the same zero-arg regression** — merged into the single inline 🔴. Issue-alignment reached it from #2051's own third repro case ("empty string"), which this PR resolves in the direction that breaks working calls.
- **Test quality**: mutation-tested the new tests — all 5 mutations (skip the strip, drop the stash, don't mark invalid, `.get` instead of `.pop`, route missing-`input` through the try/except) are caught. The tests lock what they target; the gap is the resume path and the empty-vs-missing-input split, neither of which any test pins.
- **Issue-alignment**: truncated ✅, unquoted keys ✅, empty string ✗ (the blocker). `pgrayy`'s asks all present; the out-of-band channel is the one deviation.
- **Dropped after checking**: plain interrupt/resume does *not* lose the error (`_interrupt_state.context["tool_results"]` carries it, filtered at `event_loop.py:783-786`); no `KeyError` in `strip_input_parse_errors` (empty-id blocks are dropped at `streaming.py:311-320`); the `remove()`+`append()` reordering breaks nothing; `JSONDecodeError` carries no payload so no unbounded echo; no new retry-loop risk (identical loop count on main with the same stub).

</details>

### Questions

**Blocking-ish (design, your call + a maintainer's)**
- The parse error now rides on *two* channels — `ToolUse.inputParseError` **and** `invocation_state["_tool_input_parse_errors"]`, with a documented fallback between them at `_validator.py:47-49,63`. Is keeping both deliberate? As shipped, the field's only job is to survive ~5 lines between `streaming.py:331` and `event_loop.py:647`, while `invocation_state` (cycle-local, not in `Checkpoint`/`Snapshot`) is what actually delivers the error — and neither survives a resume. Worth collapsing to one, or is the redundancy intentional belt-and-braces?
- `ToolUse` is a public type, so this wants **`api/needs-review`** (that's the CI-enforced label — `team/API_BAR_RAISING.md:72`, `.github/workflows/api-review-label.yml:14-29`; my earlier comment named `needs-api-review`, which doesn't exist — sorry). Since the shipped shape deviates from `pgrayy`'s ask in two places (out-of-band carry, and `.pop()` rather than `.get()`), would you rather get his confirm on the deviation before the label goes on? Feels like a field addition following the #1703 precedent rather than a new primitive, so an API-reviewer pass, not a meeting.

**Non-blocking**
- `strands-ts/src/models/model.ts:505-506` handles the same case by throwing `ModelError('unable to parse tool input JSON')` — failing the whole invocation instead of emitting a per-tool error result — under a different name (`toolInputParseError`). Follow-up issue to align, or is per-SDK divergence fine here? Happy to file it.
- Once this merges, close #2397 and #2054 pointing here?

<details>
<summary>Appendix — non-blocking (8)</summary>

**🟡 The fix doesn't survive a resume boundary.** `event_loop.py:320-336` (`after_model` checkpoint) and `:790-808` (a `BeforeToolsEvent`-driven interrupt) both hand control back *before* the pop at `:837`. On resume, `event_loop.py:291` skips `_handle_model_execution`, so nothing re-stashes, the marker was already popped, and the malformed call executes with `input={}` reporting `status="success"`. Reproduced end-to-end both ways, plus via plain `Agent(messages=<history ending in an unresolved malformed toolUse>)` + `agent(None)`. **Not a regression** — main does the same today; the new guarantee just doesn't reach that path. This is the concrete cost of the two-channel design in the Question above: a marker that persists survives replay but breaks "never persists"; a cycle-local channel is clean but silently degrades to the old bug. Worth a deliberate decision (and a `test_event_loop.py` case in the `test_event_loop_cycle_checkpoint_after_tools` style) rather than a reflexive fix.

**🟡 The new error text is less actionable than what the model used to get.** Pre-PR the tool ran with `{}` and pydantic answered `Field required [type=missing, input_value={}]` — naming the field. Now: `Error: Invalid JSON in tool input for 'search': Unterminated string starting at: line 1 column 11 (char 10). Retry with a valid JSON object.` — no schema, and the model's own rejected bytes are discarded at `streaming.py:303`, so it retries blind. Including the tool's `inputSchema` or a truncated snippet of the rejected input would make self-correction likelier.

**🟡 Contradiction left in the model's context**: the assistant turn shows `input: {}` (valid, empty) next to a tool result insisting the JSON was invalid, with nothing to reconcile them.

**🟡 `context_compression.py:184`** (`generate_summary`) calls `process_stream` directly, bypassing the strip, and its result *is* persisted via `SummarizingConversationManager`. Unreachable today (`tool_specs=None`, so no stock provider emits a toolUse), but nothing enforces that.

**⚪ The literal max-tokens truncation case never reaches this path**: `recover_message_on_max_tokens_reached` (`event_loop.py:606-607`) replaces the toolUse block before the strip, discarding the marker. Since truncation is #2051's headline cause, worth knowing the fix covers garbled-but-not-truncated JSON.

**⚪ Dead `'unknown'` fallback**: `streaming.py:301`'s `.get('name', 'unknown')` can't surface — `streaming.py:309` drops the whole block when `name` is missing.

**⚪ Unused second arm** at `_validator.py:63`: the production caller always passes the map, so `tool.pop("inputParseError", None)` is test-only while being documented as supported.

**⚪ Test hygiene**: no test pins empty-string vs missing input (the PR's own behavior split, currently held only by a pre-existing parametrized case); no valid+malformed mix in one message; the three new `test_validator.py` tests skip the repo's `tru_`/`exp_` convention (`docs/TESTING.md`); the agent test infers "the model sees the error" from persisted history rather than the next `stream()` call args.

**⚪ Checked, not findings**: `AfterModelCallEvent` hooks see the marker pre-strip (`event_loop.py:586`); a caller can forge or a `BeforeToolsEvent` hook can clear `_tool_input_parse_errors` (self-inflicted, same trust model as any `invocation_state` key — though `setdefault(...).update(...)` at `:649` will `AttributeError` if a caller already uses that key with a non-dict value); duplicate `toolUseId` collapses in the map (no provider does it). The `chat` OTel span does carry the marker (`end_model_invoke_span` runs before the strip) — the docstring's claim is still literally true, so this is a note, not a leak.

</details>

AI review — worth a human's eyes on the design questions before merge. The 🔴 is reproduced A/B and the suggested fix is test-verified, so that part is concrete.

JackYPCOnline:
@strandly-the-agent Give it a final review

strandly-the-agent (commented):
**Assessment**: Approve — the 🔴 blocker is fixed and verified. Nothing left that I'd block on; what remains is two maintainer decisions, unchanged from the last round.

**Verified at `fc74dd8d` vs `origin/main` (`06bef2d7`)**
- ✅ zero-arg tool calls behave exactly like main again, both shapes (`input: ""` delta, and no input delta at all): tool runs, `toolResult: success`
- ✅ malformed input still reported: truncated `{"query": "te`, unquoted `{query: test}`, and truncation right after a brace (`{`, `{\n  `) all still get the marker; a valid object with surrounding whitespace still parses to real args, because `.strip()` is only the emptiness test — `json.loads` still gets the raw string
- ✅ `pytest tests/strands/event_loop/ tests/strands/tools/ tests/strands/agent/test_agent.py -q` → **867 passed**; ruff check + format clean; mypy clean on `streaming.py`
- ✅ the new parametrized test does its job: reverting the predicate to `"input" not in current_tool_use` → **2 failed** (both `empty` and `whitespace`), so the fix can't silently rot
- ✅ CI green on this head, integration tests included

<details>
<summary>What I checked, and one ⚪ worth your judgment (non-blocking)</summary>

**Input-shape table at `fc74dd8d`** (`handle_content_block_start` → `handle_content_block_stop`, marker = `inputParseError` set):

```
start, no delta   -> input={}            -
empty ""          -> input={}            -
whitespace "   "  -> input={}            -
newline/tab       -> input={}            -
"{}"              -> input={}            -
'  {"q":"x"}  \n' -> input={'q': 'x'}    -
"{"               -> input={}            MARKER
"{\n  "           -> input={}            MARKER
'{"query": "te'   -> input={}            MARKER
"{query: test}"   -> input={}            MARKER
"null" / "0"      -> input=None / 0      -        (unchanged from main; valid JSON, non-object)
```

**⚪ The `not isinstance(raw_input, str)` arm is dead code — and that one's on me**, since it came from my suggested patch. `handle_content_block_delta` accumulates with a plain string `+=` at `streaming.py:225`, so a non-string delta already raises `TypeError` one function earlier, identically on main; and `process_stream` builds `state` itself (`streaming.py:446-452`), so no caller can inject a non-string either. Deleting the arm leaves the whole suite green, i.e. it's an untested branch guarding nothing — and if it ever *did* fire it would silently turn real arguments into `{}` rather than raise. Simplifying to `if not raw_input.strip():` is slightly better, but it's a judgment call, not a defect: keeping it as defense-in-depth is a defensible answer too. Your call, either way I'm fine.

**Standing items for a maintainer** (no changes since the last round, not re-arguing them):
- The PR still has no `api/needs-review` label. CI's `check-api-review-label` passes because that gate only enforces `api/needs-review` → `api/review-complete` pairing, so a green check isn't evidence the field addition to the public `ToolUse` was reviewed. Given `pgrayy` designed the field, a quick confirm from him on the two deviations (out-of-band carry channel, `.pop()` vs `.get()`) would close this cleanly.
- The two-channel question (`ToolUse.inputParseError` + `invocation_state["_tool_input_parse_errors"]` with a fallback between them at `_validator.py:63`) and the 🟡 resume-durability consequence (`event_loop.py:320-336` checkpoint, `:790-808` interrupt — the error is lost and the malformed call runs with `{}`) are still open. Not a regression — main does the same today — so shipping without addressing it only means the new guarantee doesn't extend to resumed cycles.

**Pipeline disclosure**: the adversarial re-pass on the new predicate ran *warm* (in-context rather than as an independent subagent). I re-verified every load-bearing claim above myself — the A/B, the shape table, the mutation test, the suite and lint runs are mine.

</details>

Good turnaround on the fix, and thanks for the takeover hygiene — `he-yufeng`'s commit kept at its original SHA. AI review, so a human still owns the two decisions above; the verification is reproducible if anyone wants to re-run it.

strands-py/src/strands/types/tools.py
```diff
@@ -59,12 +59,17 @@ class ToolUse(TypedDict):
...
+            marker that carries the parse error detail from streaming to tool validation, which converts
+            it into an error tool result. It is stripped before the assistant message is appended to
+            history, so it never persists to a session or is sent back to the model.
     """
 
     input: Any
     name: str
     toolUseId: str
     reasoningSignature: NotRequired[str]
+    inputParseError: NotRequired[str]
```
lizradway, line 72: does this need to be exposed publically?


this feels a bit granular to live here. did you consider putting this in invocation state ie. `invocation_state["_tool_input_parse_errors"]`

JackYPCOnline: That's a good idea actually. mirroring the reasoningSingature is not so good since  reasoningSingature goes to model while the error not. Will iterate. cc @pgrayy for visibility

Issue #449, opened 2025-07-14
[FEATURE] Amazon Nova tool caching with cachePoint in messages


malte-aws:
### Problem Statement

I would like strands to cache tool definitions when using Amazon Nova models.
Today Amazon Nova models [do not accept prompt cache checkpoints in tools field](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html#prompt-caching-models). However Amazon Nova will cache the tool configuration if there is a cache checkpoint set in the messages payload. Strands should set that cache checkpoint for me when I specify cache_tools so that the tool configuration is cached. 

### Proposed Solution

When setting cache_tools with BedrockModel and model_id is an Amazon Nova model, i.e.
```python
bedrock_model = BedrockModel(
    model_id="us.amazon.nova-premier-v1:0",
    cache_tools="default"
)
```

then strands should add a messages `{ cachePoint: { type: ‘default’ } }` before adding the user messages to cache the tool configuration. 

Make the following Bedrock API call when cache_tools is set and the model is an Amazon Nova model:
```python
tool_config = { ...}

messages = [
    {
        "role": "user",
        "content": [
            {"text": "You are a helpful AI assistant. Use the tools to solve the users questions."},
            { "cachePoint": { "type": 'default' } },
           .... other messages
        ],
    },
]


response = client.converse(
    modelId=[...],
    messages=messages,
    toolConfig=tool_config
)
```

### Use Case

This is useful to cache long tool definitions.

PR #3642, opened 2026-08-04
fix(injection): keep injected context out of the prompt cache prefix

opieter-aws:
## Description

`ContextInjector` prepended its per-call text into the latest user message, ahead of the stable conversation. Because that text is rebuilt on every call and the provider's cache point landed *after* it, the cached prefix diverged exactly at the injection site — so every turn re-ingested the whole conversation at full price. Measured against Bedrock, turn 2 of a two-turn conversation reused **zero** cached tokens.

Injected text is now appended behind a `cachePoint` boundary that marks where the reusable prefix ends, and Bedrock's auto mode honors a boundary already present instead of relocating it to the end. Placement belongs to whoever appends the volatile content, so the agentic token-usage middleware shares the same primitive — its live `<context-status>` count previously stranded itself inside the cached prefix even with the injector fixed.

`OpenAIModel` now warns and skips a `cachePoint` inside a message rather than raising `TypeError`, matching its own system-prompt path and the Gemini/Vercel providers. Without it, a provider-agnostic caller cannot mark ephemeral content without special-casing OpenAI.

## Related Issues

N/A

## Documentation PR

Included

## Type of Change

Bug fix

## Testing

Verified end to end against real Bedrock (`claude-sonnet-4-5`), A/B against the pre-fix placement with a unique nonce per arm so no stale cache entry could flatter the result:

| turn 2 | cacheRead | cacheWrite |
|---|---|---|
| before | **0** | 3702 |
| after | **3645** | 19 |

## Risks / Callouts

**This changes billing shape for users who never enabled caching.** The boundary is honored by Bedrock and Anthropic even with no `cache_config` set, so `ContextInjector` now implies prompt caching on those providers. Bounds, measured:

The only loss case is a single-turn conversation, multi-turn always wins. I considered gating on a `Model` capability so the boundary is only emitted where the user opted in, but that forfeits the win for everyone who hasn't set `cache_config` — most users. **Flagging it as the reviewable trade-off:** if "no silent cost change" is a hard rule regardless of magnitude, the gate is the alternative.

**Bedrock's `_inject_cache_point` behavior change is broader than the injector needs.** It now honors a caller-placed cache point in the last user message (still stripping stale ones elsewhere, so the 4-cache-point cap can't be exhausted). That affects anyone hand-placing cache points, and an existing test asserting the old strip-everything behavior was updated.

## Checklist

----

### DIFF
--- a/strands-py/src/strands/event_loop/event_loop.py
+++ b/strands-py/src/strands/event_loop/event_loop.py
@@ -136,11 +136,13 @@ async def _estimate_input_tokens(agent: "Agent") -> int:
     """
     messages = agent.messages
 
-    # Find the last assistant message with usage metadata
+    # Skip calls that used the cache: providers disagree on whether a cached portion is included in
+    # ``inputTokens``, so those fall through to counting the messages below rather than being guessed.
     last_assistant_idx = -1
-    for i, msg in reversed(list(enumerate(messages))):
-        if msg.get("role") == "assistant" and msg.get("metadata", {}).get("usage"):
-            last_assistant_idx = i
+    for index, message in reversed(list(enumerate(messages))):
+        usage = message.get("metadata", {}).get("usage") if message.get("role") == "assistant" else None
+        if usage and not (usage.get("cacheReadInputTokens") or usage.get("cacheWriteInputTokens")):
+            last_assistant_idx = index
             break
 
     if last_assistant_idx >= 0:
--- a/strands-py/src/strands/models/bedrock.py
+++ b/strands-py/src/strands/models/bedrock.py
@@ -409,6 +409,9 @@ class BedrockModel(Model):
     def _inject_cache_point(self, messages: list[dict[str, Any]]) -> None:
         """Inject a cache point at the end of the last user message.
 
+        A cache point already present in the last user message is left where it is: callers place
+        one to mark the end of the reusable prefix, ahead of injected content that is rebuilt every call.
+
         Args:
             messages: List of messages to inject cache point into (modified in place).
         """
@@ -417,6 +420,14 @@ class BedrockModel(Model):
 
         last_user_idx: int | None = None
         for msg_idx, msg in enumerate(messages):
+            if msg.get("role") == "user":
+                last_user_idx = msg_idx
+
+        # Strip cache points everywhere except the last user message, where auto mode would
+        # otherwise accumulate one per turn and exceed Bedrock's 4-cache-point ceiling.
+        for msg_idx, msg in enumerate(messages):
+            if msg_idx == last_user_idx:
+                continue
             content = msg.get("content", [])
             for block_idx, block in reversed(list(enumerate(content))):
                 if "cachePoint" in block:
@@ -426,22 +437,27 @@ class BedrockModel(Model):
                         msg_idx,
                         block_idx,
                     )
-            if msg.get("role") == "user":
-                last_user_idx = msg_idx
 
         if last_user_idx is not None and messages[last_user_idx].get("content"):
-            cache_point: dict[str, Any] = {"type": "default"}
+            content = messages[last_user_idx]["content"]
             cache_config = self.config.get("cache_config")
+
+            placed = next((block for block in content if "cachePoint" in block), None)
+            if placed is not None:
+                if cache_config and cache_config.ttl and "ttl" not in placed["cachePoint"]:
+                    placed["cachePoint"]["ttl"] = cache_config.ttl
+                logger.debug("msg_idx=<%s> | honored caller-placed cache point", last_user_idx)
+                return
+
+            cache_point: dict[str, Any] = {"type": "default"}
             if cache_config and cache_config.ttl:
                 cache_point["ttl"] = cache_config.ttl
 
-            content = messages[last_user_idx]["content"]
-
             # Insert before non-PDF document blocks to avoid Bedrock ValidationException
             first_non_pdf_doc_idx: int | None = None
-            for i, block in enumerate(content):
+            for block_idx, block in enumerate(content):
                 if "document" in block and block["document"].get("format", "") != "pdf":
-                    first_non_pdf_doc_idx = i
+                    first_non_pdf_doc_idx = block_idx
                     break
 
             # Insert the cache point before the first non-PDF document so it is not directly
--- a/strands-py/src/strands/types/content.py
+++ b/strands-py/src/strands/types/content.py
@@ -281,3 +281,40 @@ def get_message_metadata(message: Message) -> MessageMetadata:
     Individual fields (usage, metrics, custom) may not be present. Use .get() to safely access them.
     """
     return message.get("metadata", {})
+
+
+def _append_ephemeral_content(message: Message, blocks: list[ContentBlock]) -> Message:
+    """Append single-call content to a message, marking where the reusable prefix ends.
+
+    Content that is rebuilt on every model call (injected context) must sit
+    outside any cached prefix to avoid invalidating the cache. 
+    Appending puts it at the end and the leading ``cachePoint`` declares the
+    boundary, so providers that manage prompt caching keep the durable conversation cacheable.
+
+    A boundary another caller already opened is reused rather than duplicated, so several callers
+    appending in turn produce exactly one boundary ahead of all the ephemeral content. 
+
+    The input message is not mutated.
+
+    Args:
+        message: The message to append to.
+        blocks: The ephemeral content blocks to append.
+
+    Returns:
+        A new message with the boundary and ``blocks`` appended, or ``message`` when ``blocks``
+        is empty.
+    """
+    if not blocks:
+        return message
+
+    content = list(message["content"])
+    if content and not any("cachePoint" in block for block in content):
+        content.append({"cachePoint": {"type": "default"}})
+    content.extend(blocks)
+
+    appended: Message = {"role": message["role"], "content": content}
+    if "tracking_id" in message:
+        appended["tracking_id"] = message["tracking_id"]
+    if "metadata" in message:
+        appended["metadata"] = message["metadata"]
+    return appended
--- a/site/src/content/docs/user-guide/concepts/plugins/context-injector.mdx
+++ b/site/src/content/docs/user-guide/concepts/plugins/context-injector.mdx
@@ -118,6 +118,12 @@ injector = ContextInjector(
 The rendered text is a prompt-injection surface. If it interpolates attacker-influenced data (tool output, user-derived state), escape it yourself before returning. A callback that throws fails open: injection is skipped and the model call proceeds.
 :::
 
+## Prompt Caching
+
+Injected text is appended to the end of the latest user message, so it is the last thing the model reads. It is preceded by a cache point marking where the reusable prefix ends, so providers that support prompt caching keep the volatile text out of the cached prefix and the rest of the conversation still gets cache hits. Providers without cache-point control ignore the marker.
+
+Injecting a large payload still costs input tokens on every model call, since the injected text itself is never cached. Prefer the default `'userTurn'` over `'everyTurn'` to reduce token costs, except for an autonomous agent that needs the context at each step.
+
 ## Related
 
 - [Memory](../memory/overview#context-injection) builds on this engine to inject retrieved knowledge before a model call.
--- a/strands-py/src/strands/_context_manager/modes/agentic/agentic_context.py
+++ b/strands-py/src/strands/_context_manager/modes/agentic/agentic_context.py
@@ -25,7 +25,7 @@ from ....agent.conversation_manager.compression.context_compression import (
 from ....agent.conversation_manager.compression.pin_message import is_pinned, pin_message, unpin_message
 from ....agent.conversation_manager.conversation_manager import DEFAULT_CONTEXT_WINDOW_LIMIT
 from ....tools.decorator import tool
-from ....types.content import Message, _ensure_tracking_id
+from ....types.content import Message, _append_ephemeral_content, _ensure_tracking_id
 from ....types.exceptions import ContextWindowOverflowException
 from ....types.tools import ToolContext
 
@@ -292,7 +292,8 @@ def create_token_usage_middleware() -> MiddlewareInputHandler:
     The block reports projected input-token usage against the context window limit of the
     model that will actually handle the call (``context.model``), so guidance stays correct
     even when middleware has redirected the call to a different model. The original messages
-    are not mutated; the last message is copied and the status text appended to the copy.
+    are not mutated; the last message is copied and the status text appended to the copy as 
+    ephemeral content to keep it out of the prompt cache's reusable prefix.
 
     Returns:
         An async ``MiddlewareInputHandler`` for the ``InvokeModelStage.Input`` phase.
@@ -318,14 +319,7 @@ def create_token_usage_middleware() -> MiddlewareInputHandler:
         if not messages:
             return context
 
-        last_message = messages[-1]
-        new_message: Message = {
-            "role": last_message["role"],
-            "content": [*last_message["content"], {"text": status_text}],
-        }
-        if "metadata" in last_message:
-            new_message["metadata"] = last_message["metadata"]
-        messages[-1] = new_message
+        messages[-1] = _append_ephemeral_content(messages[-1], [{"text": status_text}])
 
         return replace(context, messages=messages)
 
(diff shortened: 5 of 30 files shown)


opieter-aws:
Putting back into draft - integration test is pointing at some token counting issues with caching, need to take a look there

mkmeral:

strands-py/src/strands/types/content.py
```diff
@@ -281,3 +281,40 @@ def get_message_metadata(message: Message) -> MessageMetadata:
...
+    Returns:
+        A new message with the boundary and ``blocks`` appended, or ``message`` when ``blocks``
+        is empty.
+    """
+    if not blocks:
+        return message
+
+    content = list(message["content"])
+    if content and not any("cachePoint" in block for block in content):
+        content.append({"cachePoint": {"type": "default"}})
```
mkmeral, line 312: rest of the model providers translate based on allowlist right? so new message should be ignored, if the model provider doesn't support?

strands-py/src/strands/models/bedrock.py
```diff
@@ -417,6 +420,14 @@ def _inject_cache_point(self, messages: list[dict[str, Any]]) -> None:
 
         last_user_idx: int | None = None
         for msg_idx, msg in enumerate(messages):
+            if msg.get("role") == "user":
+                last_user_idx = msg_idx
+
+        # Strip cache points everywhere except the last user message, where auto mode would
```
mkmeral, line 426: should we have an integ test for this?

mkmeral (commented):
overall for caching, i'd like to have higher integ test count to catch issues (or maybe merge the logic to some other test). aside from that looks good to me

strandly-the-agent (commented):
**TL;DR — recommend changes before merge.** The cache-prefix diagnosis is right and the nonce-per-arm A/B is the kind of evidence I wish every perf PR had. But the boundary is emitted for *every* provider with no gate, and I reproduced two regressions from that. One gate fixes both.

What I actually ran (branch `pr3642` @ `8a08eea4`, merge-base `0ab31d7c1`):

- 🔴 `Agent(model=OllamaModel(...), context_manager="agentic")` → `TypeError: content_type=<cachePoint>` at `ollama.py:158`, **turn 1, no plugin, no special input**. Same for gemini / writer / llamaapi / llamacpp / openai-responses; mistral silently drops the block. The identical script on `main` reaches the network instead.
- 🔴 A last user turn of `[text, document(csv)]` + any injected content ships `[text, document, cachePoint, text]` — the adjacency `bedrock.py:456-472` exists to prevent. **It also happens with no `cache_config` at all**, where `_inject_cache_point` never runs to relocate anything.
- ✅ Suites green, pre-existing failures only: py 4408 passed / 2 failed (otel) / 28 errors (mantle-config); ts 3928 passed / 2 failed (otel env). `test_bedrock.py` 185/185.
- ✅ I applied the honor-branch patch I suggest below and re-ran: invalid ordering gone, 185/185 still green — but the no-`cache_config` path stays broken, which is why I think the gate matters more than the patch.
- ⚪ Not verified: the live Bedrock `ValidationException` itself (no creds here). That rests on this repo's own assertions at `test_bedrock.py:3319+`.

3 inline comments. `mkmeral`'s question on `content.py:312` is answered in the first one — short version: no, most providers don't ignore it, they raise.

<details>
<summary><b>Questions (the ones I can't settle for you)</b></summary>

**1. Should a context-shaping feature turn on prompt caching?** (blocking, and it's your call, not mine)

`models/model.py:138` documents `CacheConfig(strategy="auto")` as *"automatically detect model support and inject cachePoint"* — that's the switch users know about, and `types/content.py:311-312` now injects one without it. Verified: Bedrock + `ContextInjector` with zero `cache_config` goes from 0 cache points on `main` to 1 here; `AnthropicModel` converts unconditionally (`anthropic.py:206-207`, no cache config exists on that provider at all). There's no opt-out — `ContextInjector.__init__` takes `render_content`, `name`, `trigger`.

Two bits of prior art worth weighing, both pointing the same way: `team/DECISIONS.md:120` says pay-for-play *doesn't* cover "changes that affect default behavior", and in #449 `Unshure` declined auto-injecting a message cachePoint even when `cache_tools` **was** set, on the grounds that injected message cache points surprise customers. Your "multi-turn always wins" analysis may well carry the day — but if it does, I'd expect that to land as a documented default on `CacheConfig` plus a release note, rather than as a side effect of enabling an unrelated feature.

The capability gate you floated in the PR body is the option that closes all three of these at once (billing, the 6 crashing providers, the document-adjacency path with no `cache_config`), and it's cheaper than it looks: TS already has the predicate (`_shouldEnableCaching()`, `bedrock.ts:476-496`), py is an extract-method from `bedrock.py:580-591`, and both integ tests that assert the cache win already set `cache_config=CacheConfig(strategy="auto")` (`test_context_injector_caching.py:66`, `:127`) — so the measured win survives with no test edits.

**2.** Re your question on `bedrock.py:426` (`mkmeral`) — yes to an integ test, but I'd sequence it after the fix below, otherwise it encodes the current ordering.

**3.** (non-blocking) The body describes commit `767b13fe5` only; `8a08eea43` ("Fix token counting with cache points") changes `_estimate_input_tokens`, which feeds proactive compression — worth a paragraph so a future bisector chasing a compression regression has a reason to look here. Related: that gate fixes a bug that was already reachable on `main` for anyone using `cache_config` (the new integ test at `test_agentic_context_integration.py:115-155` uses no `ContextInjector`), so "pre-existing, but this PR makes it common" is probably the framing you want.

**4.** (non-blocking) Also missing the `typescript` label for 13 TS files, and the `api/needs-review` label the earlier reviews asked for — mentioning once, not re-litigating.

</details>

<details>

- ⚪ **4-cache-point ceiling.** The strip loop now skips the last user message and the honor branch returns after the first `cachePoint`, so extras there survive. Getting past Claude's cap of 4 needs ≥3 hand-placed points plus system + tool caching — plausible, not mainline. Worth knowing the margin is gone: two hand-placed now lands exactly at 4.
- ⚪ **Boundary-reuse degenerate shapes.** A message seeded with `content: []` gets no boundary, and Bedrock auto mode then appends one *after* the ephemeral text (cache write every call, never a read); a *leading* hand-placed `cachePoint` suppresses the real boundary so nothing is cacheable. Both need hand-seeded messages — contrived, noting for completeness.
- ⚪ **No delimiter between the ask and the appended text.** Providers that join text blocks fuse them: mistral yields `"What is the capital of France? <context-status>…</context-status> INJECTED"`. `<context-status>`/`<memory>` self-delimit; a `RenderContentCallback` returning plain text doesn't, and nothing in its docstring (`_message_injection.py:29-38`) says it now needs to. A `\n\n` prefix in the primitive would settle it.
- ⚪ **`ephemeral` is overloaded, inverted.** `anthropic.py:207` writes `cache_control={"type": "ephemeral"}` for content that **is** cached; the new primitive uses "ephemeral" for content that must **not** be cached, three lines from that code. It's private/`@internal`, so `_append_per_call_content` is a free rename if you agree (`team/DECISIONS.md:168`).
- ⚪ **`AnthropicModel` `IndexError`** — `formatted_contents[-1]` (`anthropic.py:203-207`) blows up if every earlier block in the message was filtered out (location-sourced image + an appended boundary). Reproduces on `main` with a hand-placed cache point, so it's pre-existing, not yours — I'll file it separately. TS avoids it with a lookahead (`anthropic.ts:377-388`), which is probably the shape to port.
- ⚪ **TS coverage is thinner than py's here.** No ts mirror for the stale-boundary-stripping test, and the new integ test at `test/integ/agent.test.ts:159-178` constructs the model with no `cacheConfig`, so `cacheRead`/`cacheWrite` are trivially `0` and it passes whether or not `8a08eea43`'s fix exists. The python integ test does turn caching on — that asymmetry is worth closing given `mkmeral`'s ask for more caching coverage.
- ⚪ Trailing whitespace `content.py:290`, `:295`; stray word in `messages.ts:967` ("injected context **count**").

</details>

Everything above is one agent's pass — worth a human check before you act on it, especially the API question.

strands-py/src/strands/types/content.py
strandly-the-agent, line 312: **🔴 The boundary goes to every provider, and 6 of them raise on it.**

`mkmeral` asked on this line whether providers translate on an allowlist and would just ignore an unsupported block. I checked all of them — mostly no:

| what happens | providers |
|---|---|
| converts it | bedrock `bedrock.py:637`, anthropic `anthropic.py:206` |
| warns + skips | openai `openai.py:413` (this PR), litellm, sagemaker (inherit it) |
| **raises `TypeError`** | gemini `gemini.py:226`, ollama `ollama.py:158`, writer `writer.py:123`, llamaapi `llamaapi.py:130`, llamacpp `llamacpp.py:246`, openai-responses `openai_responses.py:728` |
| **silently drops** | mistral `mistral.py:226-232` (if/elif, no `else`, no warning) |

Reachable on turn 1 with no plugin and no unusual input, because `create_token_usage_middleware` (`agent.py:414-419`) has no provider check and `_estimate_input_tokens` runs on every model call:

```
Agent(model=OllamaModel(host="http://localhost:11434", model_id="llama3"), context_manager="agentic")
agent("What is the capital of France?")
→ TypeError: content_type=<cachePoint> | unsupported type   (ollama.py:158)
```

Same via `ContextInjector` and `MemoryManager` injection. The identical script on `main` gets to the network, so it's a regression, not a latent gap. TS is fine — its adapters have a generic `default:` warn+skip — which I suspect is where the PR body's "matching the Gemini/Vercel providers" came from; py Gemini raises.

Two ways out:

- **Gate the emission** on a model capability — `Model.wants_cache_boundary()`, default `False`, `BedrockModel` returning `_should_enable_caching()`, threaded in as a keyword arg here. TS already has the predicate (`bedrock.ts:476-496`); py is an extract-method from `bedrock.py:580-591`. This also fixes the Bedrock document-adjacency finding and the no-`cache_config` billing question in one move.
- **Or make every provider tolerate it** — e.g. a `drop_unsupported_cache_points()` helper alongside `_has_location_source` in `models/_validation.py`, wired into the 6 raisers plus Mistral's missing `else`. A `cachePoint` carries no content the model would miss, so warn-and-skip everywhere seems like the right SDK-wide rule; raising punishes the user for a marker the framework inserted.

Whichever you pick, `context-injector.mdx:123` ("Providers without cache-point control ignore the marker") needs to become true — it's currently false for those 6. And there's no test anywhere on this: `grep -c cachePoint` is 0 in `test_gemini.py`, `test_ollama.py`, `test_writer.py`, `test_llamaapi.py`, `test_llamacpp.py`, `test_mistral.py`, `test_openai_responses.py`.

strands-py/src/strands/models/bedrock.py
```diff
@@ -426,22 +437,27 @@ def _inject_cache_point(self, messages: list[dict[str, Any]]) -> None:
...
-            cache_point: dict[str, Any] = {"type": "default"}
+            content = messages[last_user_idx]["content"]
             cache_config = self.config.get("cache_config")
+
+            placed = next((block for block in content if "cachePoint" in block), None)
+            if placed is not None:
+                if cache_config and cache_config.ttl and "ttl" not in placed["cachePoint"]:
+                    placed["cachePoint"]["ttl"] = cache_config.ttl
+                logger.debug("msg_idx=<%s> | honored caller-placed cache point", last_user_idx)
+                return
```
strandly-the-agent, line 450: **🔴 Honoring the boundary skips the non-PDF-document rule this method already enforces below.**

The `return` here lands before the relocation logic at `bedrock.py:456-472` — the guard added by `beddc3f7d` / #2001 precisely because Bedrock rejects a `cachePoint` directly preceded by a non-PDF document. `_append_ephemeral_content` always appends at the tail, so a `[text, document(csv)]` turn plus any injected content produces exactly that shape. Verified on this head:

```
injector output:                      ['text', 'document', 'cachePoint', 'text']
after _inject_cache_point (auto):     ['text', 'document', 'cachePoint', 'text']   ← cachePoint at idx 2
control, no pre-existing boundary:    ['text', 'cachePoint', 'document']           ← correctly relocated
```

Reached by `agent([{"text": "Analyze this file"}, {"document": {"format": "csv", ...}}])` with `ContextInjector` / `MemoryManager` / agentic mode — an ordinary "analyze this file" flow. Same in `strands-ts/src/models/bedrock.ts:882-891`.

This patch fixes it (I applied it and ran `test_bedrock.py`: 185/185, invalid ordering gone):

```suggestion
            placed = next((block for block in content if "cachePoint" in block), None)
            if placed is not None:
                if cache_config and cache_config.ttl and "ttl" not in placed["cachePoint"]:
                    placed["cachePoint"]["ttl"] = cache_config.ttl

                # Bedrock rejects a cache point directly preceded by a non-PDF document, so a
                # caller-placed point follows the same relocation rule as one added below.
                first_non_pdf_doc_idx = next(
                    (
                        block_idx
                        for block_idx, block in enumerate(content)
                        if "document" in block and block["document"].get("format", "") != "pdf"
                    ),
                    None,
                )
                if first_non_pdf_doc_idx is not None and first_non_pdf_doc_idx < content.index(placed):
                    content.remove(placed)
                    if first_non_pdf_doc_idx == 0:
                        # A leading non-PDF document leaves no prefix to cache
                        logger.debug("msg_idx=<%s> | dropped cache point for leading document", last_user_idx)
                        return
                    content.insert(first_non_pdf_doc_idx, placed)
                    logger.debug("msg_idx=<%s> | relocated caller-placed cache point", last_user_idx)
                    return

                logger.debug("msg_idx=<%s> | honored caller-placed cache point", last_user_idx)
                return
```

**But the patch isn't sufficient**, and this is the part I'd weigh when deciding on the capability gate: with **no `cache_config` at all**, `_inject_cache_point` never runs, yet the injector's boundary still reaches the wire — the invalid `['text', 'document', 'cachePoint', 'text']` payload happens anyway, with the patch applied. Verified:

```
BedrockModel(model_id=...)  # no cache_config
format_request([injected_message]) → ['text', 'document', 'cachePoint', 'text']
```

So either the boundary is only emitted where the model asked for caching (the gate), or the placement rule has to move somewhere that runs unconditionally. A test that pins it: `test_inject_cache_point_honored_boundary_does_not_land_after_a_non_pdf_document`, asserting `"cachePoint" not in content[doc_idx + 1]` — it fails against the current code, which is the evidence. Worth mirroring in TS.

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -136,11 +136,13 @@ async def _estimate_input_tokens(agent: "Agent") -> int:
...
-    # Find the last assistant message with usage metadata
+    # Skip calls that used the cache: providers disagree on whether a cached portion is included in
+    # ``inputTokens``, so those fall through to counting the messages below rather than being guessed.
     last_assistant_idx = -1
-    for i, msg in reversed(list(enumerate(messages))):
-        if msg.get("role") == "assistant" and msg.get("metadata", {}).get("usage"):
-            last_assistant_idx = i
+    for index, message in reversed(list(enumerate(messages))):
+        usage = message.get("metadata", {}).get("usage") if message.get("role") == "assistant" else None
+        if usage and not (usage.get("cacheReadInputTokens") or usage.get("cacheWriteInputTokens")):
```
strandly-the-agent, line 144: **🟡 The gate is provider-blind, and for the OpenAI family it makes the estimate worse.**

Different point from the existing thread on this line (that one asks for a unit test — still worth doing; this is about the logic).

The premise — `inputTokens` excludes the cached portion — holds for Bedrock/Anthropic, and the gate is right there. It's inverted for the OpenAI family: `prompt_tokens` **includes** cached tokens and `cacheReadInputTokens` is a subset of it (`openai.py:596-597`, `openai_responses.py:888-896`, `litellm.py:287-295`, `gemini.py:479-487`). So for those providers the gate throws away the one accurate number available and falls through to `count_tokens`, which has no `image`/`document` branch (`model.py:39-86` — deliberately, per its own comment). On a vision-heavy conversation:

```
provider-reported prompt size:  26,000
before this commit:             26,408   (1.6% off)
after:                              17   (99.9% off)
proactive compression at 0.85 of a 30k window: fires before → does not fire after
```

Reachable without any of this PR's features: OpenAI caches server-side automatically above ~1024 prompt tokens, so any multi-turn `OpenAIModel` conversation reports `cacheReadInputTokens` and lands in the fallthrough. The consequence is bounded — a wrong number feeding compression and the `<context-status>` text the model reads, not a hard failure — hence 🟡.

Second-order, worth a sentence in the comment either way: once a boundary is emitted by default, *every* turn carries `cacheWrite` or `cacheRead`, so the baseline+delta optimization never engages again for the life of a cache-enabled conversation — you're doing a full recount per model call. Cheap today (local heuristic), a network call per attempt with `use_native_token_count=True`.

Suggest keying the skip on whether the provider's `inputTokens` is net-of-cache rather than on the presence of cache counters — the comment already concedes providers disagree, and that disagreement is knowable per provider.

PR #3644, opened 2026-08-04
docs(concepts): replace deprecated strands_tools examples with self-contained tools

yonib05:
## Description

Part 1 of 2, replacing #3631 (closed — see below). `strands-agents/tools` is deprecating `calculator`, `current_time`, `memory`, and `retrieve` (strands-agents/tools#566), so these examples would emit a deprecation warning for anyone following along.

`calculator` was almost always a stand-in to show that tool calling works at all — but the prompts ask for real arithmetic ("square root of 144", "450 km at 120 km/h"), so the demo has to compute something. Each example now defines a small `@tool` that walks an arithmetic AST against an explicit operator allowlist:

```python
import ast
import operator

@tool
def calculator(expression: str) -> str:
    """Evaluate an arithmetic expression such as "144 ** 0.5" or "450 / 120"."""
    ops = {ast.Add: operator.add, ast.Sub: operator.sub, ast.Mult: operator.mul,
           ast.Div: operator.truediv, ast.Pow: operator.pow, ast.USub: operator.neg}

    def ev(n):
        if isinstance(n, ast.Constant) and isinstance(n.value, (int, float)):
            return n.value
        ...
        raise ValueError(f"unsupported expression: {expression}")

    return str(ev(ast.parse(expression, mode="eval").body))
```

It computes what the prompts need, drops the `strands_tools` dependency, and rejects anything that isn't arithmetic — so no docs example demonstrates eval-style execution of model-supplied input.

`memory` and `retrieve` are removed from the examples that used them, **except** where the example's whole premise is Bedrock Knowledge Base retrieval. Those need porting to `MemoryManager`, which is a redesign rather than a reference swap, and doing it badly here would leave worse examples than the warning does.

### Why #3631 was closed

Worth stating plainly: #3631 failed the `pr-size-labeler` gate at 2300 lines, and my fix was to swap the AST evaluator for `sympy.sympify(...)`. That was a security regression — `sympify` evaluates arbitrary code:

```
>>> sympy.sympify("__import__('os').getpid()")
7592
```

Caught by the automated review. The evaluator is now compacted to 20 lines instead, and the change is **split** so each PR fits the gate honestly rather than by weakening the code:

- **This PR** — `user-guide/concepts/`, 916 lines
- **Companion** — model-providers, evals-sdk, observability, quickstart, runnable examples, 786 lines

## Type of Change

Documentation update

## Testing

- **Every generated helper was executed**, not just parsed: each returns `12.0` for `"144 ** 0.5"` and raises `ValueError` on `__import__('os').getpid()`. 60 helpers across both parts, zero failures.
- Every touched Python block parses; the two pre-existing syntax errors in truncated illustrative snippets are unchanged (verified by parsing the pre-change version).
- No `tools=[...]` entry references an undefined name — checked across blocks, since pages define the tool once and reuse it.
- Blog posts deliberately untouched: dated announcements, and rewriting their code would misrepresent what shipped.

Not applicable — `site/` only, no Python package code. `npm install` in `site/` fails with an npm registry auth error (`E401`) in my environment, so CI covers the JS-side checks.

## Checklist

strands-agents/tools#566 is still open; this PR only removes usages, so it's safe to merge independently.

----

### DIFF
--- a/site/src/content/docs/user-guide/concepts/agents/prompts.mdx
+++ b/site/src/content/docs/user-guide/concepts/agents/prompts.mdx
@@ -127,7 +127,7 @@ Prompting is a primary functionality of Strands that allows you to invoke tools 
 <Tab label="Python">
 
 ```python
-result = agent.tool.current_time(timezone="US/Pacific")
+result = agent.tool.calculator(expression="12 * 12")
 ```
 </Tab>
 <Tab label="TypeScript">
--- a/site/src/content/docs/user-guide/concepts/agents/prompts.mdx
+++ b/site/src/content/docs/user-guide/concepts/agents/prompts.mdx
@@ -127,7 +127,7 @@ Prompting is a primary functionality of Strands that allows you to invoke tools 
 <Tab label="Python">
 
 ```python
-result = agent.tool.current_time(timezone="US/Pacific")
+result = agent.tool.my_tool(param="value")
 ```
 </Tab>
 <Tab label="TypeScript">
--- a/site/src/content/docs/user-guide/concepts/agents/state.mdx
+++ b/site/src/content/docs/user-guide/concepts/agents/state.mdx
@@ -89,8 +89,39 @@ Direct tool calls are (by default) recorded in the conversation history:
 <Tab label="Python">
 
 ```python
-from strands import Agent
-from strands_tools import calculator
+
+import ast
+import operator
+
+from strands import Agent, tool
+
+@tool
+def calculator(expression: str) -> str:
+    """Evaluate an arithmetic expression such as "144 ** 0.5" or "450 / 120".
+
+    Args:
+        expression: The arithmetic expression to evaluate.
+    """
+    ops = {ast.Add: operator.add, ast.Sub: operator.sub, ast.Mult: operator.mul,
+           ast.Div: operator.truediv, ast.USub: operator.neg}
+
+    def ev(n):
+        if isinstance(n, ast.Constant) and isinstance(n.value, (int, float)):
+            return n.value
+        if isinstance(n, ast.BinOp) and isinstance(n.op, ast.Pow):
+            base, exp = ev(n.left), ev(n.right)
+            if abs(exp) > 64:
+                raise ValueError(f"exponent too large in {expression!r}: {exp}")
+            return base**exp
+        if isinstance(n, ast.BinOp) and type(n.op) in ops:
+            return ops[type(n.op)](ev(n.left), ev(n.right))
+        if isinstance(n, ast.UnaryOp) and type(n.op) in ops:
+            return ops[type(n.op)](ev(n.operand))
+        raise ValueError(
+            f"{expression!r} is not arithmetic; supported: + - * / ** and parentheses over numbers"
+        )
+
+    return str(ev(ast.parse(expression, mode="eval").body))
 
 agent = Agent(tools=[calculator])
 
--- a/site/src/content/docs/user-guide/concepts/agents/state.mdx
+++ b/site/src/content/docs/user-guide/concepts/agents/state.mdx
@@ -89,8 +89,32 @@ Direct tool calls are (by default) recorded in the conversation history:
 <Tab label="Python">
 
 ```python
-from strands import Agent
-from strands_tools import calculator
+
+import ast
+import operator
+
+from strands import Agent, tool
+
+@tool
+def calculator(expression: str) -> str:
+    """Evaluate an arithmetic expression such as "144 ** 0.5" or "450 / 120".
+
+    Args:
+        expression: The arithmetic expression to evaluate.
+    """
+    ops = {ast.Add: operator.add, ast.Sub: operator.sub, ast.Mult: operator.mul,
+           ast.Div: operator.truediv, ast.Pow: operator.pow, ast.USub: operator.neg}
+
+    def ev(n):
+        if isinstance(n, ast.Constant) and isinstance(n.value, (int, float)):
+            return n.value
+        if isinstance(n, ast.BinOp) and type(n.op) in ops:
+            return ops[type(n.op)](ev(n.left), ev(n.right))
+        if isinstance(n, ast.UnaryOp) and type(n.op) in ops:
+            return ops[type(n.op)](ev(n.operand))
+        raise ValueError(f"unsupported expression: {expression}")
+
+    return str(ev(ast.parse(expression, mode="eval").body))
 
 agent = Agent(tools=[calculator])
 
--- a/site/src/content/docs/user-guide/concepts/model-providers/amazon-bedrock.mdx
+++ b/site/src/content/docs/user-guide/concepts/model-providers/amazon-bedrock.mdx
@@ -597,9 +597,33 @@ Tool caching allows you to reuse a cached tool definition across multiple reques
 In Python, use the `cache_tools` parameter to enable tool caching independently:
 
 ```python
+
+import ast
+import operator
+
 from strands import Agent, tool
 from strands.models import BedrockModel
-from strands_tools import calculator, current_time
+
+@tool
+def calculator(expression: str) -> str:
+    """Evaluate an arithmetic expression such as "144 ** 0.5" or "450 / 120".
+
+    Args:
+        expression: The arithmetic expression to evaluate.
+    """
+    ops = {ast.Add: operator.add, ast.Sub: operator.sub, ast.Mult: operator.mul,
+           ast.Div: operator.truediv, ast.Pow: operator.pow, ast.USub: operator.neg}
+
+    def ev(n):
+        if isinstance(n, ast.Constant) and isinstance(n.value, (int, float)):
+            return n.value
+        if isinstance(n, ast.BinOp) and type(n.op) in ops:
+            return ops[type(n.op)](ev(n.left), ev(n.right))
+        if isinstance(n, ast.UnaryOp) and type(n.op) in ops:
+            return ops[type(n.op)](ev(n.operand))
+        raise ValueError(f"unsupported expression: {expression}")
+
+    return str(ev(ast.parse(expression, mode="eval").body))
 
 # Using tool caching with BedrockModel
 bedrock_model = BedrockModel(
@@ -610,7 +634,7 @@ bedrock_model = BedrockModel(
 # Create an agent with the model and tools
 agent = Agent(
     model=bedrock_model,
-    tools=[calculator, current_time]
+    tools=[calculator]
 )
 # First request will cache the tools
 response1 = agent("What time is it?")
--- a/site/src/content/docs/user-guide/concepts/model-providers/amazon-bedrock.mdx
+++ b/site/src/content/docs/user-guide/concepts/model-providers/amazon-bedrock.mdx
@@ -597,9 +597,40 @@ Tool caching allows you to reuse a cached tool definition across multiple reques
 In Python, use the `cache_tools` parameter to enable tool caching independently:
 
 ```python
+
+import ast
+import operator
+
 from strands import Agent, tool
 from strands.models import BedrockModel
-from strands_tools import calculator, current_time
+
+@tool
+def calculator(expression: str) -> str:
+    """Evaluate an arithmetic expression such as "144 ** 0.5" or "450 / 120".
+
+    Args:
+        expression: The arithmetic expression to evaluate.
+    """
+    ops = {ast.Add: operator.add, ast.Sub: operator.sub, ast.Mult: operator.mul,
+           ast.Div: operator.truediv, ast.USub: operator.neg}
+
+    def ev(n):
+        if isinstance(n, ast.Constant) and isinstance(n.value, (int, float)):
+            return n.value
+        if isinstance(n, ast.BinOp) and isinstance(n.op, ast.Pow):
+            base, exp = ev(n.left), ev(n.right)
+            if abs(exp) > 64:
+                raise ValueError(f"exponent too large in {expression!r}: {exp}")
+            return base**exp
+        if isinstance(n, ast.BinOp) and type(n.op) in ops:
+            return ops[type(n.op)](ev(n.left), ev(n.right))
+        if isinstance(n, ast.UnaryOp) and type(n.op) in ops:
+            return ops[type(n.op)](ev(n.operand))
+        raise ValueError(
+            f"{expression!r} is not arithmetic; supported: + - * / ** and parentheses over numbers"
+        )
+
+    return str(ev(ast.parse(expression, mode="eval").body))
 
 # Using tool caching with BedrockModel
 bedrock_model = BedrockModel(
@@ -610,10 +641,10 @@ bedrock_model = BedrockModel(
 # Create an agent with the model and tools
 agent = Agent(
     model=bedrock_model,
-    tools=[calculator, current_time]
+    tools=[calculator]
 )
 # First request will cache the tools
-response1 = agent("What time is it?")
+response1 = agent("What is 12 * 12?")
 print(f"Cache write tokens: {response1.metrics.accumulated_usage.get('cacheWriteInputTokens')}")
 print(f"Cache read tokens: {response1.metrics.accumulated_usage.get('cacheReadInputTokens')}")
 
--- a/site/src/content/docs/user-guide/concepts/tools/index.mdx
+++ b/site/src/content/docs/user-guide/concepts/tools/index.mdx
@@ -27,8 +27,41 @@ Tools are passed to agents during initialization or at runtime, making them avai
 <Tab label="Python">
 
 ```python
-from strands import Agent
-from strands_tools import calculator, file_read, shell
+
+import ast
+import operator
+
+from strands import Agent, tool
+from strands_tools import file_read
+from strands.vended_tools import shell
+
+@tool
+def calculator(expression: str) -> str:
+    """Evaluate an arithmetic expression such as "144 ** 0.5" or "450 / 120".
+
+    Args:
+        expression: The arithmetic expression to evaluate.
+    """
+    ops = {ast.Add: operator.add, ast.Sub: operator.sub, ast.Mult: operator.mul,
+           ast.Div: operator.truediv, ast.USub: operator.neg}
+
+    def ev(n):
+        if isinstance(n, ast.Constant) and isinstance(n.value, (int, float)):
+            return n.value
+        if isinstance(n, ast.BinOp) and isinstance(n.op, ast.Pow):
+            base, exp = ev(n.left), ev(n.right)
+            if abs(exp) > 64:
+                raise ValueError(f"exponent too large in {expression!r}: {exp}")
+            return base**exp
+        if isinstance(n, ast.BinOp) and type(n.op) in ops:
+            return ops[type(n.op)](ev(n.left), ev(n.right))
+        if isinstance(n, ast.UnaryOp) and type(n.op) in ops:
+            return ops[type(n.op)](ev(n.operand))
+        raise ValueError(
+            f"{expression!r} is not arithmetic; supported: + - * / ** and parentheses over numbers"
+        )
+
+    return str(ev(ast.parse(expression, mode="eval").body))
 
 # Add tools to our agent
 agent = Agent(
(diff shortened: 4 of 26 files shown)


yonib05:
@strandly-the-agent please review — this replaces #3631.

Your review there caught the important one: my size-gate fix had swapped the AST evaluator for `sympy.sympify(...)`, which evaluates arbitrary code (`sympify("__import__('os').getpid()")` returns a live PID). That would have put an eval-style path into 45 doc pages while the PR body still described the safe version.

Fixed by compacting the AST evaluator to 20 lines instead of weakening it, and splitting the change so each PR fits the size gate honestly. Every helper is verified to compute `144 ** 0.5 -> 12.0` and raise on `__import__('os').getpid()`. Companion PR is #3648.

site/src/content/docs/user-guide/concepts/model-providers/amazon-bedrock.mdx
```diff
@@ -610,7 +634,7 @@ bedrock_model = BedrockModel(
 # Create an agent with the model and tools
 agent = Agent(
     model=bedrock_model,
-    tools=[calculator, current_time]
+    tools=[calculator]
 )
 # First request will cache the tools
 response1 = agent("What time is it?")
```
strandly-the-agent: `current_time` was dropped from `tools=[...]` three lines up, so this agent now has only `calculator` and can't answer the prompt.

The example is teaching *tool caching*, so the prompt content is incidental — but it should at least be answerable, and using arithmetic for both requests keeps the cache-write-then-cache-read narrative intact.

```suggestion
response1 = agent("What is 12 * 12?")
```

site/src/content/docs/user-guide/concepts/tools/index.mdx
```diff
@@ -416,8 +442,34 @@ Pre-built tools are available in both Python and TypeScript to help you get star
 For Python, Strands offers a [community-supported tools package](https://github.com/strands-agents/tools/blob/main) with pre-built tools for development:
```
strandly-the-agent, line 449: This section is `### 2. Vended Tools` (`:433`) → "Pre-built tools are available… to help you get started quickly" (`:435`) → **Community Tools Package** (`:440`), and the first thing under it is now 27 lines of hand-rolled `calculator` (`:446-472`) — which is *not* in that package. `file_read` and `shell` on the next lines are the genuine vended examples.

A reader skimming for "what do I get pre-built?" would reasonably conclude `calculator` ships in `strands-agents-tools`. One clause distinguishing them fixes it:

```suggestion
For Python, Strands offers a [community-supported tools package](https://github.com/strands-agents/tools/blob/main) with pre-built tools for development. `file_read` and `shell` below come from that package; `calculator` is a small self-contained stand-in defined inline:
```

(Distinct from the earlier bot comment on `:36`, which was about the `shell` import *source* rather than the section's framing.)

strandly-the-agent (commented):
**Verdict: comment (approve-leaning).** The `sympify` RCE is genuinely gone — I re-attacked the new evaluator hard and it held, so the thing that sank #3631 is fixed rather than papered over, and the split is honest (916 lines, `size/l`; #3644 ∩ #3648 file sets are disjoint). What's left is docs coherence, and none of it blocks merge.

⚠️ **Don't act on the earlier bot review's `--8<--` suggestion.** `.agents/skills/docs-writer/SKILL.md:60,69` says TypeScript is never inlined in MDX but **Python may be**, since Python in the docs tree isn't type-checked. Factoring the helper into a shared snippet would go *against* the repo's own authoring rule. (Same correction I made on #3631 — flagging it so you don't spend a round on it.)

### 🟡 1. `^` now errors, and two pages' own prompts use it
`tools/index.mdx:65` — `agent("What is 42 ^ 9")` sits in the *same code block* as the new helper (defined at `:32-57`). `agent-to-agent.mdx:84-86` — `a2a_agent("Show me 10 ** 6")`'s sibling, plus a hardcoded expected output `'10^6 = 1,000,000'`. `^` is `ast.BitXor`, which isn't in the allowlist. This is a real behavior regression: the old tool read `^` as exponent. One-character fix — make both prompts `**`.

<details><summary>Verified old vs. new</summary>

```python
# strands_tools 0.8.5 (pip installed, actually run) — sympy applies convert_xor
calculator(expression="42 ^ 9")  # {'status':'success', ... 'Result: 406671383849472'}
calculator(expression="10 ^ 6")  # {'status':'success', ... 'Result: 1000000'}

# new helper, extracted verbatim from tools/index.mdx:31-57 and run
'42 ^ 9'    !! ValueError: unsupported expression: 42 ^ 9
'10 ^ 6'    !! ValueError: unsupported expression: 10 ^ 6
'42 ** 9'   -> '406671383849472'   # works
```

Honest caveat: I could not verify whether a model passes `42 ^ 9` through verbatim or normalizes it to `42 ** 9` — no Bedrock access in my sandbox (`AccessDeniedException`). That's why this is 🟡 and not 🔴: the failure is always a clean tool error the model can retry past, never a silently wrong answer. But a reader running the page's own snippet may well see the error.
</details>

### 🟡 2. Three prompts orphaned by the `current_time` removal
The tool left `tools=[...]`; the text asking for it stayed.
- `model-providers/ollama.mdx:286` — `agent("What's the square root of 144 plus the current time?")` with `tools=[calculator]`. Strongest of the three: half the question is now unanswerable.
- `model-providers/amazon-bedrock.mdx:640` — inline comment below.
- `bidirectional-streaming/quickstart.mdx:315` — the "You can now ask questions like:" list still offers `- "What time is it?"` while `:298` is `tools=[calculator, get_weather]`. Deleting the bullet is the clean fix (the other two bullets already cover both remaining tools). Worth noting this one is a carry-over: I flagged the identical leftover on #3631 and it came across unfixed.

### 🟡 3. Ten pages install `strands-agents-tools` but no longer import it
Inlining `calculator` removed the last `strands_tools` import from each of these pages (verified: zero `from strands_tools` / `import strands_tools` remaining on all ten), but the setup command still pulls the package: `anthropic.mdx:21`, `google.mdx:25`, `litellm.mdx:18`, `llamaapi.mdx:24`, `llamacpp.mdx:18`, `mistral.mdx:21`, `ollama.mdx:27`, `openai.mdx:21`, `sagemaker.mdx:21`, `writer.mdx:18`. Dropping ` strands-agents-tools` from those ten lines finishes the job.

1. Is a part 3 planned for the two spots neither PR covers — `concepts/tools/community-tools-package.mdx:61,62,100,101` (still advertises `retrieve`/`memory`/`calculator`/`current_time` with no deprecation note) and `site/docs/examples/python/knowledge_base_agent.py:29` (`from strands_tools import use_llm, memory`)? Both were noted on #3631 too; just want to confirm they're tracked rather than dropped.
2. The docstring — *"Evaluate an arithmetic expression such as `144 ** 0.5` or `450 / 120`"* — never states the grammar boundary (numeric literals and `+ - * / **` only; no `sqrt()`, `%`, `//`, `^`, variables, comparisons), and the error the model gets back is `Error: unsupported expression: <expr>`, which names the symptom but not the rule. Eight pages ask for a square root in prose. Worth one more clause so a model's retry has something to act on, or over-engineering for a doc example?

<details>
<summary>✅ Verification ledger — what I actually ran at <code>cdc143c2b</code></summary>

- **Security holds.** Python 3.13 exposes 28 `ast.expr` node types; the evaluator has exactly 3 accept branches. Confirmed rejected by execution: `__import__('os').getpid()`, `eval`/`exec`/`open`/`breakpoint`, `().__class__.__bases__[0].__subclasses__()`, f-strings, lambdas, comprehensions, walrus, subscripts, `Compare`/`BoolOp`/`IfExp`, bare statements, and `str`/`bytes`/`None`/`complex` literals. A 30,000-case fuzz over hostile atoms × 23 operators produced **0** allowlist violations (only `bool` slips through, harmlessly, as an `int` subclass).
- **Mechanical integrity.** All 32 inlined copies byte-identical (md5 `fe2952827cc600450e30d227bf0cb1d3`). Every touched Python block parses — the only two syntax errors are pre-existing truncated illustrative fragments, unchanged. No added line under `site/src/content/docs/` exceeds the 90-char limit (max 84, `agents/state.mdx:106`). No `tools=[...]` names an undefined symbol.
- **Spot-checks that came back clean.** `strands.vended_tools.shell` really exists (`strands-py/src/strands/vended_tools/shell/__init__.py`). `retrieve` is now fully gone from `site/` (0 leftovers). `structured-output.mdx`'s `MathResult.result: int` is fine — `42 + 8` returns int `50`, not a float.
- **Size gate.** 844+/72- = 916 → `size/l`; `.github/workflows/pr-size-labeler.yml:25-26,55-56` only fails at `size/xl` (>1000).
- **Note for context:** nothing in CI parses or executes Python blocks in `.mdx` (`docs-ci.yml` is Astro build + TS typecheck + vitest), so all of the above rests on review rather than a gate.

</details>

<details>
<summary>Appendix — non-blocking (7), no fix requested</summary>

- **Unbounded `**` freezes the event loop.** `calculator("9**9**9")` never returns (the result needs ~370M digits). Measured: `asyncio.wait_for(..., timeout=2.0)` **never fires**, SIGINT ignored for 15 s (needed SIGKILL), and the loop gets 1 heartbeat tick during the call instead of ~37 — sync tools run via `asyncio.to_thread` (`strands-py/src/strands/tools/decorator.py:638`) and CPython bigint math holds the GIL. **Not a regression**: `strands_tools` 0.8.5 hangs identically on the same input. Raising it only because this PR promotes the pattern into 32 copy-paste blocks, and a ~2-line exponent cap on `ast.Pow` would close it if you feel like hardening while you're here.
- **complex/inf/nan return as `status: "success"`.** `(-8) ** 0.5` → `'(1.7319121124709868e-16+2.8284271247461903j)'`; `1e308 * 10` → `'inf'`. The one case where the model gets numerically-useless output *marked correct* instead of a recoverable error. Also true of the old tool, so not a regression.
- **Formatting drift vs. the old tool.** `144 ** 0.5` → `'12.0'` (old: `'12'`), `0.1 + 0.2` → `'0.30000000000000004'` (old rounded to 10 sig figs), `10/3` → `'3.3333333333333335'`. Cosmetic; models generally tidy this up in prose.
- **Import grouping**, 9 files: the new `import ast`/`import operator` form their own group, stranding a pre-existing stdlib import (`asyncio`/`json`/`logging`/`typing`) below a blank line in the third-party group. `.mdx` isn't linted, so purely tidiness.
- **`bidirectional-streaming/agent.mdx:93-110`** calls bare `calculator` without defining it in its own block, relying on the previous one. Pre-existing pattern — but making that block standalone now costs 20 lines instead of one import line.
- **`model-providers/openai-responses.mdx:19`** has the same stale `strands-agents-tools` install, but it never imported `strands_tools` in the first place — pre-existing, not caused by this PR. Excluded from finding 3 for that reason; easy freebie if you're touching the other ten.
- **Process note for me, not you:** my first read of this diff came back corrupted mid-file (a phantom `import ast` inside a function body, and a missing one elsewhere). Both were rendering artifacts — a scripted re-parse of every block at `cdc143c2b` found zero import or syntax defects. Mentioning it only so the record shows those two "findings" were mine to retract, not yours to fix.

</details>

---
Automated review by `strandly-the-agent` — solid work for a human to approve, not a gate. The security fix here is the real headline: compacting the evaluator instead of weakening it was the right call, and it stood up to a much harder pass than the one that caught #3631.

site/src/content/docs/user-guide/concepts/agents/prompts.mdx
```diff
@@ -127,7 +127,7 @@ Prompting is a primary functionality of Strands that allows you to invoke tools
 <Tab label="Python">
 
 ```python
-result = agent.tool.current_time(timezone="US/Pacific")
+result = agent.tool.my_tool(param="value")
```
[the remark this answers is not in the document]
yonib05: Fixed — now agent.tool.calculator(expression="12 * 12"), a concrete call.

site/src/content/docs/user-guide/concepts/tools/index.mdx
```diff
@@ -27,8 +27,34 @@ Tools are passed to agents during initialization or at runtime, making them avai
...
 ```python
-from strands import Agent
-from strands_tools import calculator, file_read, shell
+
+import ast
+import operator
+
+from strands import Agent, tool
+from strands_tools import file_read
+from strands.vended_tools import shell
```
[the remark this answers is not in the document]
yonib05: Fixed — the Community Tools Package section now shows only pre-built tools (file_read, file_write, http_request) from one import source.

site/src/content/docs/user-guide/concepts/model-providers/amazon-bedrock.mdx
yonib05: Fixed — prompt is now "What is 12 * 12?", answerable with the tool the agent actually has.

site/src/content/docs/user-guide/concepts/tools/index.mdx
yonib05: Fixed — removed the hand-rolled calculator from that section; it now shows genuine vended tools.

yonib05:
@strandly-the-agent all 4 addressed and resolved — please re-review.

yonib05:
Split into #3656 and #3657 — the exponent guard added in review pushed this to 1047 lines, over the 1000 size-gate limit. Splitting by directory rather than compacting the guard.

site/src/content/docs/user-guide/concepts/agents/state.mdx
```diff
@@ -89,8 +89,39 @@ Direct tool calls are (by default) recorded in the conversation history:
...
+    """
+    ops = {ast.Add: operator.add, ast.Sub: operator.sub, ast.Mult: operator.mul,
+           ast.Div: operator.truediv, ast.USub: operator.neg}
+
+    def ev(n):
+        if isinstance(n, ast.Constant) and isinstance(n.value, (int, float)):
+            return n.value
+        if isinstance(n, ast.BinOp) and isinstance(n.op, ast.Pow):
+            base, exp = ev(n.left), ev(n.right)
+            if abs(exp) > 64:
```
strandly-the-agent, line 113: This threshold rejects ordinary arithmetic. Verified against the current helper: `2**100` → `ValueError: exponent too large: 100`, and `2**-70` is refused too — both instant to compute, both handled by `strands_tools` 0.8.5 and by this PR's previous revision, and the docstring never mentions an exponent ceiling for the model to work around.

`1024` still blocks everything the cap was added for (`9**9**9` has exponent 387420489; `2**2**2**2**2` has 65536) while admitting `2**100` and `2**1024` in 0.000s — at zero line cost:

```suggestion
            if abs(exp) > 1024:
```

That said, per the summary I'd rather you **revert the cap entirely** — it's what pushed the PR to `size/xl` and broke `label-size`, and it was an explicitly non-blocking appendix note about behavior `strands_tools` 0.8.5 already had. This suggestion is the fallback if you'd prefer to keep it.

(Flagging here on `state.mdx` for anchoring only — all 31 inlined copies are byte-identical, so whichever way you go it applies to every one.)

strandly-the-agent (commented):
Re-reviewed at `ed091de`. **The 4 inline threads are genuinely fixed** — I verified each one, they're not just marked resolved. But there's a new 🔴, and findings 1–3 from my review *body* are untouched.

🔴 **`label-size` is now failing.** 1036+/73- = **1109 lines** → `size/xl`, and `.github/workflows/pr-size-labeler.yml:55-56` calls `core.setFailed()` above 1000. [Failed run](https://github.com/strands-agents/harness-sdk/actions/runs/31012876242) — it's the one red check of 20.

The cause is my own appendix note: the exponent cap added ~7 lines × 31 copies = **+193 lines** (916 → 1109). I flagged that hang as *inherited from `strands_tools` 0.8.5, not a regression, "not requesting it"* — and it's now the only thing standing between this PR and a green gate. **My recommendation is to revert the cap**, which puts you back at ~916 and green. Sorry for the round trip; an appendix note shouldn't have cost you the gate whose enforcement created this PR in the first place.

**On "all 4 addressed"** — you're right about the 4 threads, and the mismatch is my tooling's fault, not yours: findings 1–3 sat in the review *body* because they land on lines GitHub won't let me anchor a comment to (unchanged lines outside any diff hunk). Nothing to resolve there, so they were easy to miss. Restating them compactly, all re-verified at `ed091de`:

### 🟡 1. `^` — still open, and now self-contradictory on the page
`tools/index.mdx:72` — `agent("What is 42 ^ 9")`, in the *same code block* as a helper whose new error message reads *"supported: + - * / ** and parentheses over numbers"*. `agent-to-agent.mdx:84` — `"Show me 10 ^ 6"`, with `:86`'s hardcoded output `'10^6 = 1,000,000'` the code above can no longer produce. Re-ran both against the current helper: `ValueError`. Also `multi-agent/agent-to-agent.ts:15` carries the same `'Show me 10 ^ 6'` — untouched by this PR, but worth syncing if you change the Python.

### 🟡 2. Orphaned `current_time` prompts — 2 of 3 still open
`model-providers/ollama.mdx:293` — `"...plus the current time?"` with `tools=[calculator]` at `:289`. `bidirectional-streaming/quickstart.mdx:322` — the `- "What time is it?"` bullet with `tools=[calculator, get_weather]` at `:305`. (`amazon-bedrock.mdx:647` ✅ fixed.)

### 🟡 3. Stale installs — all 10 still open
Verified each still has **zero** `strands_tools` imports while instructing `pip install ... strands-agents-tools`: `anthropic.mdx:21`, `google.mdx:25`, `litellm.mdx:18`, `llamaapi.mdx:24`, `llamacpp.mdx:18`, `mistral.mdx:21`, `ollama.mdx:27`, `openai.mdx:21`, `sagemaker.mdx:21`, `writer.mdx:18`.

### 🟡 4. If you'd rather keep the cap than revert it
`abs(exp) > 64` over-restricts ordinary math — see the inline comment. `64 → 1024` is a strictly better threshold at zero line cost, but reverting still beats it on the gate.

<details>
<summary>✅ What I verified is fixed at <code>ed091de</code></summary>

- `prompts.mdx:130` → `agent.tool.calculator(expression="12 * 12")`, concrete call. ✅
- `tools/index.mdx:447-458` → Community Tools Package section is now clean: one import source, only genuine package tools (`file_read`, `file_write`, `http_request`), hand-rolled `calculator` gone. Both the import-source and section-framing threads are properly resolved, not papered over. ✅
- `amazon-bedrock.mdx:647` → `agent("What is 12 * 12?")`, answerable with the tool the agent has. ✅
- **Helper consistency:** 31 copies, all byte-identical (single variant, md5 `7a180d64b1…`), all carrying the cap — no drift across the rewrite. Scripted re-parse of every fenced Python block across all 26 files: **0** syntax or missing-import defects.
- **Security still holds.** Re-confirmed `__import__('os').getpid()`, `sqrt(144)` and the attribute-chain escapes are rejected. Restructuring `ast.Pow` out of the `ops` dict into its own branch didn't open anything.
- **The cap does work on the case it targeted:** `9**9**9` → `ValueError` in 0.000s (was an unkillable hang), `2**2**2**2**2` likewise. The new error message naming the grammar is a real improvement for model retries.

</details>

<details>
<summary>Appendix — non-blocking detail behind finding 4</summary>

**Measured, `limit=64` vs `limit=1024`** (extracted helper, actually run):

| expression | `> 64` (current) | `> 1024` |
|---|---|---|
| `144 ** 0.5` | ok `12.0` | ok |
| `2**100` | ❌ `exponent too large: 100` | ok, 31 chars, 0.000s |
| `2**1024` | ❌ rejected | ok, 309 chars, 0.000s |
| `2**-70` | ❌ rejected | ok, 21 chars |
| `9**9**9` | ✅ blocked | ✅ blocked (exp 387420489) |
| `2**2**2**2**2` | ✅ blocked | ✅ blocked (exp 65536) |

So `2 ** 100` — an ordinary question for a calculator, instant to compute, handled by both `strands_tools` 0.8.5 and this PR's previous revision — is now refused, and nothing in the docstring tells the model an exponent ceiling exists. `1024` fixes that while still blocking everything the cap was added for.

**The cap is also only a partial fix**, which is part of why I'd just revert it: chained exponentiation slips past the per-node check, since each level's exponent is individually small. `((10**64)**64)**64` → `ValueError` after 0.019s, but `(((10**64)**64)**64)**64` burns **13.7s of CPU** before erroring. Contrived input no model would emit — injection-tier only, and it terminates rather than hanging — so ⚪, not a reason to keep iterating on the cap.

**Line-count arithmetic, if you keep the cap and want the gate green:** you need to shed ≥110 lines. Single-lining the `raise` back (`−3`×31 ≈ 93) doesn't quite get there on its own, so it'd mean moving a page or two into a part 3. Reverting the cap is the cheaper path.

</details>

---
Automated re-review by `strandly-the-agent`. Net: the security story is still solid and every thread you resolved really is resolved — the blocker is just the size gate, and the fastest way through it is to drop the cap I shouldn't have raised.

PR #3657, opened 2026-08-05
docs(concepts): self-contained calculator in the remaining concepts examples

yonib05:
## Description

Companion to #3656 (both split from #3644, which exceeded the size limit at 1047 lines once review feedback added an exponent guard). This half covers `bidirectional-streaming/`, `agents/`, `multi-agent/`, and `tools/`.

Same substitution: a small self-contained `@tool` walking an arithmetic AST against an explicit operator allowlist, with `**` capped at 64 so an unbounded tower can't hang the process.

**Three additional fixes from review:**

- `tools/index.mdx` — the **Community Tools Package** section had a 27-line hand-rolled `calculator` under a heading advertising pre-built tools, mixing two import sources in one snippet. It now shows `file_read`, `file_write`, and `http_request` from `strands_tools` only. A reader skimming for "what do I get pre-built?" would previously have concluded `calculator` ships in the package.
- `agents/prompts.mdx` — the direct-tool-call example used an abstract `agent.tool.my_tool(param="value")`; now a concrete `calculator` call.
- `model-providers/amazon-bedrock.mdx` — the tool-caching example asked "What time is it?" from an agent that only had a calculator, so it was unanswerable. Now asks arithmetic, which keeps the cache-write-then-cache-read narrative intact.

## Testing

- Every helper executed: `144 ** 0.5` → `12.0`; rejects `__import__('os').getpid()` and the exponent tower.
- Every touched Python block parses; no `tools=[...]` entry references an undefined name.
- Verified `file_read`, `file_write`, and `http_request` are **not** part of the deprecation, so the replacement snippet stays valid.

## Type of Change

Documentation update

## Checklist

----

### DIFF
--- a/site/src/content/docs/user-guide/concepts/bidirectional-streaming/quickstart.mdx
+++ b/site/src/content/docs/user-guide/concepts/bidirectional-streaming/quickstart.mdx
@@ -246,11 +246,38 @@ Always call `agent.stop()` **after** exiting the `run()` or `receive()` loop, ne
 Just like standard Strands agents, bidirectional agents can use tools during conversations:
 
 ```python
+import ast
 import asyncio
+import operator
 from strands import tool
 from strands.experimental.bidi import BidiAgent, BidiAudioIO
 from strands.experimental.bidi.models import BidiNovaSonicModel
-from strands_tools import calculator, current_time
+
+@tool
+def calculator(expression: str) -> str:
+    """Evaluate an arithmetic expression such as "144 ** 0.5" or "450 / 120".
+
+    Args:
+        expression: The arithmetic expression to evaluate.
+    """
+    ops = {ast.Add: operator.add, ast.Sub: operator.sub, ast.Mult: operator.mul,
+           ast.Div: operator.truediv, ast.USub: operator.neg}
+
+    def ev(n):
+        if isinstance(n, ast.Constant) and isinstance(n.value, (int, float)):
+            return n.value
+        if isinstance(n, ast.BinOp) and isinstance(n.op, ast.Pow):
+            base, exp = ev(n.left), ev(n.right)
+            if abs(exp) > 64:  # unbounded pow hangs uninterruptibly
+                raise ValueError(f"exponent too large: {exp}")
+            return base**exp
+        if isinstance(n, ast.BinOp) and type(n.op) in ops:
+            return ops[type(n.op)](ev(n.left), ev(n.right))
+        if isinstance(n, ast.UnaryOp) and type(n.op) in ops:
+            return ops[type(n.op)](ev(n.operand))
+        raise ValueError(f"not arithmetic (+ - * / ** and parens only): {expression!r}")
+
+    return str(ev(ast.parse(expression, mode="eval").body))
 
 # Define a custom tool
 @tool
@@ -271,7 +298,7 @@ def get_weather(location: str) -> str:
 model = BidiNovaSonicModel()
 agent = BidiAgent(
     model=model,
-    tools=[calculator, current_time, get_weather],
+    tools=[calculator, get_weather],
     system_prompt="You are a helpful assistant with access to tools."
 )
 
--- a/site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
+++ b/site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
@@ -61,26 +61,26 @@ The simplest way to use an agent as a tool is to pass it directly in the `tools`
 
 ```python
 from strands import Agent
-from strands_tools import retrieve, http_request
+from strands_tools import http_request
 
 # Create specialized agents
 research_agent = Agent(
     system_prompt="""You are a specialized research assistant. Focus only on providing
     factual, well-sourced information in response to research questions.
     Always cite your sources when possible.""",
-    tools=[retrieve, http_request],
+    tools=[http_request],
 )
 
 product_agent = Agent(
     system_prompt="""You are a specialized product recommendation assistant.
     Provide personalized product suggestions based on user preferences.""",
-    tools=[retrieve, http_request],
+    tools=[http_request],
 )
 
 travel_agent = Agent(
     system_prompt="""You are a specialized travel planning assistant.
     Create detailed travel itineraries based on user preferences.""",
-    tools=[retrieve, http_request],
+    tools=[http_request],
 )
 
 # Create the orchestrator — agents are automatically converted to tools
@@ -163,7 +163,7 @@ For more control over how the agent is invoked — such as custom pre/post-proce
 
 ```python
 from strands import Agent, tool
-from strands_tools import retrieve, http_request
+from strands_tools import http_request
 
 RESEARCH_ASSISTANT_PROMPT = """
 You are a specialized research assistant. Focus only on providing
@@ -185,7 +185,7 @@ def research_assistant(query: str) -> str:
     try:
         research_agent = Agent(
             system_prompt=RESEARCH_ASSISTANT_PROMPT,
-            tools=[retrieve, http_request]
+            tools=[http_request]
         )
 
         response = research_agent(query)
@@ -223,7 +223,7 @@ def product_recommendation_assistant(query: str) -> str:
         product_agent = Agent(
             system_prompt="""You are a specialized product recommendation assistant.
             Provide personalized product suggestions based on user preferences.""",
-            tools=[retrieve, http_request, dialog],
+            tools=[http_request],
         )
         # Implementation with response handling
         # ...
@@ -246,7 +246,7 @@ def trip_planning_assistant(query: str) -> str:
         travel_agent = Agent(
             system_prompt="""You are a specialized travel planning assistant.
             Create detailed travel itineraries based on user preferences.""",
-            tools=[retrieve, http_request],
+            tools=[http_request],
         )
         # Implementation with response handling
         # ...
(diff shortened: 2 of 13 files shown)


site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
```diff
@@ -223,7 +223,7 @@ def product_recommendation_assistant(query: str) -> str:
         product_agent = Agent(
             system_prompt="""You are a specialized product recommendation assistant.
             Provide personalized product suggestions based on user preferences.""",
-            tools=[retrieve, http_request, dialog],
+            tools=[http_request, dialog],
```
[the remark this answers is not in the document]
yonib05: Fixed — dropped `dialog`. Worth noting it was already undefined on main (the import there was only http_request), so my verifier missed it by only checking the five deprecated names; fixed since my edit touched that line.

site/src/content/docs/user-guide/concepts/tools/index.mdx
```diff
@@ -27,8 +27,38 @@ Tools are passed to agents during initialization or at runtime, making them avai
...
+    ops = {ast.Add: operator.add, ast.Sub: operator.sub, ast.Mult: operator.mul,
+           ast.Div: operator.truediv, ast.USub: operator.neg}
+
+    def ev(n):
+        if isinstance(n, ast.Constant) and isinstance(n.value, (int, float)):
+            return n.value
+        if isinstance(n, ast.BinOp) and isinstance(n.op, ast.Pow):
+            base, exp = ev(n.left), ev(n.right)
+            if abs(exp) > 64:  # unbounded pow hangs uninterruptibly
+                raise ValueError(f"exponent too large: {exp}")
```
strandly-the-agent, line 53: **🟡 Should-fix — the guard bounds the exponent, but the cost is in the *result*, so it over-blocks and under-blocks at the same time.**

*Over-blocks:* it refuses ordinary arithmetic a reader will plausibly ask right after the example — `2 ** 100`, `2 ** 128`, `10 ** 100`, `1.05 ** 360`, `(1 + 0.05 / 12) ** 360` (compound interest), `0.5 ** 10000`, `0 ** 1000000`. All instant and safe to compute; all get `exponent too large`, which the model can't work around and may paper over with a hallucinated answer.

*Under-blocks:* parentheses defeat right-associativity, so every node's exponent stays at exactly 64 while the base explodes — `(((9 ** 64) ** 64) ** 64) ** 64` (23 chars) burns **14.45 s** of CPU before failing on int→str, and one more level was killed at **60 s** with no result. Through the real `@tool` path it stalls the asyncio event loop for the whole 14 s and `asyncio.wait_for(timeout=2.0)` can't preempt it (bignum holds the GIL). `9 ** 9 ** 9 ** 9` is still caught, so the guard only covers the unparenthesized form its comment describes.

Bounding predicted result *digits* is transitive — every `**` result is ≤ 10^1000, so every outer base is too, and nesting can't grow:

```suggestion
            if base and abs(exp) * math.log10(abs(base)) > 1000:  # bounds the result, not the exponent
                raise ValueError(f"result too large: {expression!r}")
```

Needs `import math` in the stdlib group next to `import ast` / `import operator` (line 28-29 here). `abs()` on **both** operands matters: the `abs(base)`-only variant raises an uncaught `TypeError` on `2 ** ((-8) ** 0.5)`, which today's guard survives. Measured on the replacement: 22/22 plausible inputs correct, nesting depths 2–6 refused in ≤0.04 ms, and the error message gets shorter. Bump `1000` to `4300` if you'd rather allow `2 ** 4096` — still transitively safe.

Same two lines in the other 9 copies (`agents/state.mdx:112`, `agents/structured-output.mdx:304`, `bidirectional-streaming/agent.mdx:68`, `events.mdx:517`, `models/gemini_live.mdx:66`, `nova_sonic.mdx:70`, `openai_realtime.mdx:65`, `quickstart.mdx:271`, `multi-agent/agent-to-agent.mdx:317`) plus #3656's.

site/src/content/docs/user-guide/concepts/bidirectional-streaming/quickstart.mdx
```diff
@@ -271,7 +298,7 @@ def get_weather(location: str) -> str:
 model = BidiNovaSonicModel()
 agent = BidiAgent(
     model=model,
-    tools=[calculator, current_time, get_weather],
+    tools=[calculator, get_weather],
```
strandly-the-agent, line 301: **🟡 Should-fix — stale prose below.** Dropping `current_time` here removes the agent's only time capability, but the example questions at `quickstart.mdx:318` still lead with:

```
- "What time is it?"
- "Calculate 25 times 48"
- "What's the weather in San Francisco?"
```

A reader following along asks the first one and gets nothing useful. Drop that bullet (the remaining two match `calculator` and `get_weather` exactly) — it's the same fix you made for the tool-caching example in #3656. Commenting here because line 318 is outside the diff.

strandly-the-agent (commented):
Re-reviewed at `96ef2f98`. The `dialog` fix is in (`agents-as-tools.mdx:226`) and I'm not reopening the import-source note you closed. One substantive item this round — in the exponent guard itself, which bounds the wrong quantity: it refuses ordinary arithmetic *and* still lets the pow tower through. **Verdict: comment** — nothing that should block a docs PR.

<details>
<summary>✅ / 🔴 What I actually verified at <code>96ef2f98</code> (vs merge-base <code>2f907793</code>)</summary>

- ✅ All 10 inlined `calculator` copies are sha256-identical; every touched Python block parses and every import resolves against `strands-agents-tools==0.8.5` + this repo's `strands-py`. Executed: `144 ** 0.5` → `12.0`, `25 * 48` → `1200`, `42 ** 9` → `406671383849472`.
- ✅ `retrieve`/`dialog` cleanly gone from all four `tools=[...]` lists in `agents-as-tools.mdx`, with no orphaned prose on the page.
- ✅ `from strands.vended_tools import shell` valid (`strands-py/src/strands/vended_tools/__init__.py:26-32`); `from strands_tools import stop` / `file_read` / `file_write` / `http_request` all still valid, so nothing deprecated was left behind.
- ✅ Security sweep of the evaluator: 24 payloads (`__import__('os').getpid()`, `(1).__class__.__base__`, `(lambda:1)()`, walrus, comprehensions, subscripts, ternary, `%` `//` `@` `<<`, comparisons) → all `ValueError`. No code execution, no sequence memory bomb; 10,000-digit literals hit CPython's 4300-digit cap instantly; 1000-term expressions → `RecursionError` in 3 ms.
- 🔴 `abs(exp) > 64` fails in both directions (should-fix, inline below): rejects 10 of 22 plausible inputs, and `(((9 ** 64) ** 64) ** 64) ** 64` — 23 chars, every exponent exactly 64 — ran **14.45 s** of CPU before dying on int→str conversion; one more nesting level was killed at **60 s** with no result (~1.02e9 digits ≈ 425 MB predicted). Through the real `@tool` path it stalled the event loop 14.09 s and `asyncio.wait_for(timeout=2.0)` only returned after 14.07 s (bignum holds the GIL, so cancellation can't preempt it). `9 ** 9 ** 9 ** 9` unparenthesized *is* still caught.
- ✅ Re-validated the suggested one-line replacement myself: all 22 plausible inputs compute correctly (incl. `2 ** 128`, `10 ** 100`, `(1 + 0.05 / 12) ** 360`), every nesting depth 2–6 refused in ≤0.04 ms, and no uncaught `TypeError` on `2 ** ((-8) ** 0.5)` (that one needs `abs()` on *both* operands — the `abs(base)`-only variant crashes).

</details>

### Questions (non-blocking)

- The guard line is identical in this PR's 10 copies and reportedly ~14 more in #3656 — fix in both, or land in whichever merges first and rebase the other? (Not relitigating the shared-snippet call you already made; just noting a real logic bug now living in ~24 copies shifts that cost math a bit.)
- #3644/#3656 cite `strands-agents/tools#566` for the deprecation of `calculator`/`current_time`/`retrieve`; I couldn't resolve that issue via the API from here, and installed `strands-agents-tools==0.8.5` imports all three with no warning (only `agent_graph` and `use_llm` carry notices there). Worth confirming the upstream premise — the self-containment rationale stands on its own either way, so this isn't a defect in the diff.

<details>

- ⚪ `tools/index.mdx:69` — `agent("What is 42 ^ 9")` is unchanged prose inside the block you edited, but the new local `calculator` has no `ast.BitXor`, so `42 ^ 9` raises `ValueError`, where the old `strands_tools.calculator` used `sympify(convert_xor=True)` and returned `406671383849472` (verified on 0.8.5). The example now depends on the model rewriting `^` → `**`, which it probably does given the docstring. Cheap belt-and-braces: make the prompt `"What is 42 ** 9"`.
- ⚪ `agents/prompts.mdx:130` — `agent.tool.calculator(expression="12 * 12")` while the page's only `Agent(...)` (`prompts.mdx:22`) has no `tools=`, so pasting raises `AttributeError`. Pre-existing shape (it was `agent.tool.current_time(...)`), so not this PR's problem — noting only because it's the one touched page that didn't get the self-contained tool.
- ⚪ The PR body's third bullet describes the `model-providers/amazon-bedrock.mdx` "What time is it?" fix, but that file is in #3656's diff, not this one. Body over-claim, not a code issue.
- ⚪ Not asking for a fix: the chained-multiply route (`"*".join(["((9**64)**64)"]*400)`, 5.6 KB → 5.98 s) survives even the suggested guard. It needs multi-KB of crafted input and closing it costs a work counter — more lines than the bug is worth in a snippet inlined 10×.

</details>

Solid work for a human to approve — the swap itself is clean and consistent, and the evaluator holds up against everything I threw at it except the magnitude question.

PR #3660, opened 2026-08-05
feat(storage-ts): accept storage as optional top-level Agent parameter

lizradway:
## Description

Adds an optional `storage` parameter to the Agent. Subsystems without their own explicit storage resolve from it during `initAgent`. Each subsystem namespaces under its own prefix (`session/`, `offloader/`) to avoid key collisions. Explicit storage on a subsystem always wins.

Non-breaking: `storage` on SessionManager and ContextOffloader becomes optional with the same defaults as before (InMemoryStorage for offloader, throws for SessionManager if no backend available).

## Type of Change

New feature

## Testing

- [x] `npm run type-check && npm test` passes
- [x] 8 new unit tests covering resolution order, fallback, override, and error cases

## Checklist

----

### DIFF
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -48,7 +48,7 @@ import { NullConversationManager } from '../conversation-manager/null-conversati
 import { ConversationManager } from '../conversation-manager/conversation-manager.js'
 import { ContextOffloader } from '../vended-plugins/context-offloader/plugin.js'
 import { AgentDelegation } from './agent-delegation.js'
-import { InMemoryStorage } from '../storage/in-memory-storage.js'
+import type { Storage } from '../storage/storage.js'
 import { HookRegistryImplementation } from '../hooks/registry.js'
 import { createMiddlewareInterrupt } from '../middleware/interrupt.js'
 import { MiddlewareRegistry, InvokeModelStage, AgentStreamStage } from '../middleware/index.js'
@@ -231,9 +231,9 @@ export type AgentConfig = {
    * If `conversationManager` is also provided, the user's conversation manager is used instead.
    * Defaults to undefined (SlidingWindowConversationManager, no offloader).
    *
-   * @remarks The offloader uses in-memory storage that does not persist across process
-   * restarts. For agents using `sessionManager`, provide an explicit `ContextOffloader`
-   * with durable storage via the `plugins` parameter.
+   * @remarks The offloader uses in-memory storage by default. When an agent-level
+   * `storage` is provided, the offloader uses that instead. Alternatively, provide
+   * an explicit `ContextOffloader` with its own storage via the `plugins` parameter.
    */
   contextManager?: ContextManagerStrategy
   /**
@@ -322,6 +322,16 @@ export type AgentConfig = {
    * default changes.
    */
   sandbox?: Sandbox | false
+  /**
+   * Default storage backend for agent subsystems.
+   *
+   * When provided, subsystems that do not have their own explicit storage
+   * (e.g., SessionManager, ContextOffloader) resolve from this value. Each
+   * subsystem auto-namespaces under its own prefix (`session/`, `offloader/`)
+   * to avoid key collisions. Storage specified directly on a subsystem always
+   * takes precedence over this agent-level default.
+   */
+  storage?: Storage
 }
 
 /**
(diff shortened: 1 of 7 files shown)


strands-ts/src/session/session-manager.ts
```diff
@@ -293,6 +308,11 @@ export class SessionManager implements Plugin, MultiAgentPlugin {
 
   /** Initializes the multi-agent plugin by registering orchestrator lifecycle hooks. */
   public initMultiAgent(orchestrator: MultiAgent): void {
+    if (!this._storage) {
+      throw new Error(
```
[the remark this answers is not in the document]
lizradway: intentional

strands-ts/src/agent/agent.ts
```diff
@@ -231,9 +231,9 @@ export type AgentConfig = {
    * If `conversationManager` is also provided, the user's conversation manager is used instead.
    * Defaults to undefined (SlidingWindowConversationManager, no offloader).
    *
-   * @remarks The offloader uses in-memory storage that does not persist across process
-   * restarts. For agents using `sessionManager`, provide an explicit `ContextOffloader`
-   * with durable storage via the `plugins` parameter.
+   * @remarks The offloader uses in-memory storage by default. When an agent-level
```
JackYPCOnline, line 234: Do we need to update other filds that use storage?

lizradway: these are the only fields that use storage right now: context offloader + session manager

lizradway: the only files that use storage are context offloader and session manager

strands-ts/src/session/session-manager.ts
JackYPCOnline, line 130: SessionManager is a plugin, so initAgent sets _storage before hooks run which is fine, but other CRUDs are public methods callable directly; on a storage-less, not-yet-attached manager they now throw a cryptic "Cannot read properties of undefined (reading 'snapshot')"  instead of the intended backend error. so would recommand add the guard here.

lizradway: i believe all of those callable methods take an agent instance as input param and thus would have storage or hit the storage error already.

JackYPCOnline: hmm, delete session doesn't need an agent as input. Storage is only ever resolved in initAgent, even passing a fully-configured agent with storage still throws the cryptic error, because the manager never consults the agent's storage herem only its own _storage.

lizradway: yeah i think delete session is the one case where you could hit the TypeError instead of a pretty error. i kinda dont think a user would ever call delete session without an agent or session manager initialized since there is nothing to delete?

lizradway: can add the prettier error on delete session for defense in depth if you think it is worthwhile?

JackYPCOnline: yep, `! ` turn off the check so it's safer to have a guard, would approve, besides, set storage on agent doesnt help standalone delete function but it is a nit.

JackYPCOnline (dismissed):

strands-ts/src/agent/agent.ts
```diff
@@ -322,6 +322,16 @@ export type AgentConfig = {
    * default changes.
    */
   sandbox?: Sandbox | false
+  /**
+   * Default storage backend for agent subsystems.
```
opieter-aws, line 326: Do we need to wire this up for memory too?

strands-ts/src/session/session-manager.ts
```diff
@@ -63,8 +63,11 @@ export interface SessionManagerConfig {
    * Accepts either:
    * - A unified {@link Storage} instance (recommended) — wrapped internally with {@link SnapshotStorageAdapter}
    * - A legacy `{ snapshot: SnapshotStorage }` object
+   *
+   * When omitted, resolves from the agent-level `storage` during initialization.
+   * If no agent-level storage is available either, an error is thrown.
```
opieter-aws, line 68: Why no default?

strands-ts/src/agent/agent.ts
lizradway: ah yes, i forgot that TestMemoryStore uses Storage now too. i thought it was only FileMemoryStore on Maisie's dev branch

strands-ts/src/session/session-manager.ts
lizradway: that is the previous behavior/unchanged in this pr. from what i remember, in memory would just discard after the runtime closes out.

opieter-aws (approved):

Issue #3563, opened 2026-07-30
[FEATURE] Add Mantle-API support for Anthropic, Google and Grok models


cmatskas:
### SDK Language

TypeScript

### Problem Statement

Currently, only OpenAI models can be called using the Mantle API in the SDK. Please add support for the rest of the models. 

### Proposed Solution

new feature implementation

### Use Case

Instead of having two different libraries for separately calling Bedrock and Mantle, users that want to migrate to the new mantle api should be able to do this with support for all major models

### Alternatives Solutions

Maintain two different routines for calling models across bedrock and mantle, along with the complexity added with the additional business logic needed to support this

Issue #3279, opened 2026-07-15
[BUG] OpenAIModel Bedrock Mantle base URL is wrong for the Responses API (hits /v1/responses)


niklas-palm:
### SDK Language

TypeScript

### Strands Version

1.7.0

### Language Runtime Version

Node.js 20+/22

### Operating System

macOS (region-independent; reproduced against Bedrock Mantle `us-west-2`)

### Installation Method

npm

### Summary

`OpenAIModel({ bedrockMantleConfig })` is unusable with any Bedrock-hosted OpenAI model that uses the Responses API (e.g. `openai.gpt-5.4`, `openai.gpt-5.6-*`), which is also the **default** `api` mode. Every call fails with a misleading 400.

Bedrock Mantle exposes two OpenAI-compatible surfaces under different path prefixes:

| `api` mode  | OpenAI SDK call          | Correct Mantle base URL                         | Models                                          |
| ----------- | ------------------------ | ----------------------------------------------- | ----------------------------------------------- |
| `responses` | `POST /responses`        | `https://bedrock-mantle.<region>.api.aws/openai/v1` | hosted OpenAI (`gpt-5.4`, `gpt-5.6-*`)      |
| `chat`      | `POST /chat/completions` | `https://bedrock-mantle.<region>.api.aws/v1`        | OSS (`gpt-oss-120b`, `gpt-oss-20b`)         |

`bedrockMantleBaseUrl` (`src/models/openai/mantle.ts`) hardcodes `/v1` regardless of `api`, and `buildMantleClient` (`src/models/openai/model.ts`) never passes the mode through — even though the constructor already knows `this._api`. So `api: 'chat'` works by luck, and the default `api: 'responses'` hits `POST /v1/responses`, an invalid Mantle route.

### Steps to Reproduce

```ts
import { OpenAIModel } from '@strands-agents/sdk/models/openai'

// api defaults to 'responses'
const model = new OpenAIModel({
  modelId: 'openai.gpt-5.6-luna',
  bedrockMantleConfig: { region: 'us-west-2' },
})

for await (const _ of model.stream(
  [{ role: 'user', content: [{ text: 'Say hi' }] }],
  { systemPrompt: 'Be brief', toolSpecs: [] },
)) { /* ... */ }
```

The client's `baseURL` is `https://bedrock-mantle.us-west-2.api.aws/v1`, so Strands sends `POST /v1/responses`.

### Expected Behavior

[...]

PR #3666, opened 2026-08-05
fix(models): route each Mantle model to the base path it is served from

### LINKED ISSUE #3654: [BUG] Bedrock Mantle base-path table only handles openai.gpt-5.*, mis-routing xai.grok-4.3 (and every third-party model) to /v1
### Checks

### SDK Language

TypeScript

### Strands Version

Reported against current `main` / latest published `@strands-agents/sdk`. Exact version not captured from the reporter — the routing defect below is present on `main` as of this filing.

### Language Runtime Version

Not captured from the reporter (Node.js).

### Operating System

Not captured from the reporter.

### Installation Method

npm

### Steps to Reproduce

Reported by a user attempting to reach `xai.grok-4.3` through the Bedrock Mantle endpoint in `us-east-1` (`https://bedrock-mantle.us-east-1.api.aws`).

```ts
import { OpenAIModel } from '@strands-agents/sdk/models/openai'

const model = new OpenAIModel({
  modelId: 'xai.grok-4.3',
  bedrockMantleConfig: { region: 'us-east-1' },
})

await model.stream([{ role: 'user', content: [{ text: 'hello' }] }])
```

The model id is valid and servable-by-catalog: `GET /v1/models` on the same endpoint returns `xai.grok-4.3` with `"status":"available"`, so this is not a typo'd or unavailable model id.

The reporter also exercised the three routes directly with `curl` to isolate the SDK from the service:

| Route | Payload shape | HTTP | Response |
|---|---|---|---|
| `/v1/chat/completions` | Chat Completions | 400 | `{"error":{"code":"validation_error","message":"model 'xai.grok-4.3' isn't supported on this route","type":"invalid_request_error"}}` |
| `/openai/v1/chat/completions` | Chat Completions | 500 | `{"error":{"code":"internal_server_error","message":"The server had an error while processing your request. Sorry about that!","type":"server_error"}}` |
| `/openai/v1/responses` | Responses API | 500 | same `internal_server_error` body |

Control case, same account, same credentials, same session: `google.gemma-4-31b` returns valid completions on **`/openai/v1/chat/completions`** and **`/openai/v1/responses`**. That rules out credentials, account permissions, and client code.

### Expected Behavior

Two separate expectations, because there are two separate defects here — one ours, one not.

1. `OpenAIModel` + `bedrockMantleConfig` should route a non-OpenAI-family Mantle model id (`xai.*`, `google.*`, `anthropic.*`) to the base path Mantle actually serves it from. It should not construct a request against a route the service will reject with `validation_error`.
2. A model advertised as `"status":"available"` by `GET /v1/models` should be invocable on at least one route.

### Actual Behavior

**Defect 1 (SDK, actionable here): the Mantle base-path table only special-cases `openai.gpt-5.*`, so every third-party model id falls through to `/v1` — the route that 400s.**

`strands-ts/src/models/openai/mantle.ts`:

```ts
const OPENAI_PATH_MODEL_PREFIXES = ['openai.gpt-5.'] as const

export function bedrockMantleBaseUrl(region: string, modelId: string): string {
  const suffix = OPENAI_PATH_MODEL_PREFIXES.some((prefix) => modelId.startsWith(prefix)) ? '/openai/v1' : '/v1'
  return `https://bedrock-mantle.${region}.api.aws${suffix}`
}
```

`xai.grok-4.3` does not start with `openai.gpt-5.`, so it resolves to `/v1` — and `/v1/chat/completions` is precisely the route that returns `model 'xai.grok-4.3' isn't supported on this route`. The SDK cannot currently be pointed at `/openai/v1` for this model either, since `bedrockMantleConfig` rejects a caller-supplied `clientConfig.baseURL` (validated in `mantle.test.ts`, "throws when `clientConfig.baseURL` is set alongside `bedrockMantleConfig`"). There is no supported escape hatch.

The reporter's own control case is the proof that the table is wrong, not just incomplete: `google.gemma-4-31b` **works** on `/openai/v1`, but the SDK would route it to `/v1` for the same reason. So the mis-routing is not specific to Grok — it applies to every non-`openai.gpt-5.*` id, and it is currently masked only because those models aren't documented as supported yet (#3563).

The comment in `mantle.ts` states the operative rule correctly ("the base path is keyed by model family, not API surface") but the prefix list encodes only the OpenAI families that existed when it was written.

**Defect 2 (Mantle service-side, not fixable in this repo): `/openai/v1/*` 500s for `xai.grok-4.3`.**

Once routed to `/openai/v1`, Grok returns `internal_server_error` on both Chat Completions and Responses, while Gemma succeeds on the same routes with the same credentials. That is a serving-side problem with this specific model, either a catalog mis-registration (advertised `available`, not actually servable) or a backend fault. Fixing Defect 1 will **not** make `xai.grok-4.3` work; it will change a 400 into a 500 until the service side is resolved. Worth cutting a separate ticket to the Bedrock Mantle service team, and worth confirming Grok is servable before wiring it up here — otherwise SDK-side support cannot be integ-tested.

### Additional Context

Parity: the Python SDK has the identical defect at `strands-py/src/strands/models/_openai_bedrock.py`:

```python
_OPENAI_PATH_MODEL_PREFIXES: tuple[str, ...] = ("openai.gpt-5.",)

def _resolve_mantle_base_path(model_id: str) -> str:
    if model_id.startswith(_OPENAI_PATH_MODEL_PREFIXES):
        return "/openai/v1"
    return "/v1"
```

Any fix should land in both SDKs to preserve parity.

Related history on this same base-path logic, all of which suggests the prefix-matching approach is fragile enough to be worth replacing rather than extending:

- #3279 — `[BUG] OpenAIModel Bedrock Mantle base URL is wrong for the Responses API (hits /v1/responses)` (closed)
- #2629 — `bedrock_mantle_config builds wrong base_url for OpenAI Responses models (missing /openai path segment)` (closed)
- #3388 — `OpenAIModel (Responses, TypeScript) sends invalid assistant history shape, rejected by Bedrock Mantle` (closed)

This is the fourth bug filed against Mantle base-URL/route derivation.

### Possible Solution

Scoped to Defect 1 only.

The narrow fix is to add the third-party families to the prefix list so they resolve to `/openai/v1`:

```ts
const OPENAI_PATH_MODEL_PREFIXES = ['openai.gpt-5.', 'xai.', 'google.', 'anthropic.'] as const
```

That is a one-line change, mirrored in Python, and it makes the confirmed-working case (`google.gemma-4-31b` on `/openai/v1`) actually reachable through the SDK. But it keeps the design that has now produced four bugs: a hardcoded allowlist of prefixes that must be updated every time Mantle onboards a family, where the failure mode is a confusing service-side 400 rather than a clear client-side error.

Two alternatives worth weighing during triage:

1. **Invert the default.** `/openai/v1` appears to be the general case and `/v1` the exception (`openai.gpt-oss-*`). Encoding the exceptions instead of the rule shrinks the list that needs maintenance and makes new families work by default.
2. **Fail loudly on unrecognized families.** If the model id matches no known Mantle family, throw a client-side error naming the id and the supported prefixes rather than silently building a URL that the service will reject. That converts an opaque `validation_error` from Mantle into an actionable message, which is what the reporter needed here.

Either way, extending the allowlist without also deciding on the default is likely to produce bug number five.

Also worth considering: whether `bedrockMantleConfig`'s hard rejection of `clientConfig.baseURL` should relax to a documented override. The validation is sound as a default (it prevents accidentally unsigned or mis-pointed requests), but with no escape hatch, a routing-table gap is a hard block rather than a workaround-able annoyance.

### Related Issues

#3563 (feature request: Mantle-API support for Anthropic, Google and Grok models). This issue is the bug half of that request: the routing table silently mis-routes those families today rather than rejecting them, so the failure surfaces as an opaque service 400.

yonib05:
## Description

Bedrock Mantle serves each model from exactly one base path (`/v1` or `/openai/v1`) and rejects the other with HTTP 400 `validation_error`. Base-path resolution only special-cased `openai.gpt-5.*`, so every other model Mantle serves from `/openai/v1` was sent to `/v1` and failed.

`xai.grok-4.3` was unusable through the SDK for this reason, even though `GET /v1/models` reports it as `"status":"available"`.

The key finding is that the base path is a property of the **individual model**, not of its vendor prefix. A prefix cannot decide it:

| Model | Served from |
|---|---|
| `google.gemma-4-31b` | `/openai/v1` |
| `google.gemma-3-27b-it` | `/v1` |

Same `google.` prefix, different base paths. So the obvious one-line fix (adding `'xai.'`, `'google.'`, `'anthropic.'` to the prefix list) would have mis-routed all three Gemma 3 models, trading one bug for another.

The fix scopes each prefix to a single verified model line rather than a vendor: `openai.gpt-5.`, `xai.grok-4.`, and `google.gemma-4-`, the three lines that cover the verified `/openai/v1` catalog exactly. A new point release within a verified line routes correctly with no SDK change; an unmatched new line falls to `/v1`, and the drift test names it as the signal to verify the line and extend the list. Both SDKs are updated in parallel.

Anthropic models are out of scope here: they are served from `/anthropic/v1/messages`, a different protocol reached via `AnthropicModel` rather than `OpenAIModel`. They answer on neither OpenAI-compatible base path, so the OpenAI-compat table cannot route them and the drift test excludes them.

### Catching this early in future

Mantle exposes no way to discover a model's base path: `GET /v1/models` reports `status` but not routing, and there is no `/openai/v1/models`. The prefix list is therefore hand-maintained and goes stale whenever Mantle onboards a model line.

So each SDK gets an integ test that replays the derivation against the live catalog. For every listed model it asserts the resolved path really does serve the model, and only then checks the other path, which keeps four outcomes distinguishable: correct, mis-routed (names the ids to add), served from neither path (unroutable through `OpenAIModel` at all), and undetermined. This is the fourth bug filed against Mantle base-URL derivation (#3279, #2629, #3388), and it is the part intended to stop the fifth.

Only HTTP 200 and 400 count as answers. A 429, a 5xx, or a timeout is retried with backoff and, if it never settles, fails the test rather than being counted as "not served". A detector that reports "no drift" when it means "could not tell" would be worse than none, since the "verified against the us-east-1 catalog" note next to the prefixes would look CI-backed when it is not. The #3654 reporter hit exactly such a transient `internal_server_error`.

I verified the detector catches drift by removing the `xai.grok-4.` prefix and confirming the failure message names the offending id:

```
AssertionError: Mantle serves these models from the base path the SDK does not use.
Update _OPENAI_PATH_MODEL_PREFIXES in strands/models/_openai_bedrock.py (and the TypeScript
mirror in strands-ts/src/models/openai/mantle.ts): {'xai.grok-4.3': '/v1'}
```

**One caveat on where this runs.** On the first CI run of this branch both detectors reported *skipped*, not passed: the integ test role could mint a token but got 401 on `GET /v1/models`, because `integ-test-role.ts` granted `bedrock-mantle:CreateInference` and `CallWithBearerToken` but not `ListModels`. A detector that skips while the run stays green is worse than no detector, so this PR adds that grant, plus a CDK test that fails if any of the three Mantle actions is dropped.

The grant only reaches CI once `test-infra` is redeployed to the Strands test account, which this PR does not trigger. Until that redeploy the detectors keep skipping in CI, so a green run is not evidence they ran. They pass in full against an account that already holds the grant, which is how the results below were produced.

## Related Issues

Fixes #3654. Unblocks the TypeScript half of #3563.

On the second half of #3654: the `/openai/v1` flakiness the reporter saw recurs intermittently. On 2026-08-06 `xai.grok-4.3` served 200 on both `/openai/v1` routes in a morning sweep, then timed out on both for a stretch the same day while `/v1` kept rejecting it with the correct 400, so the routing stands but a service-side ticket may be warranted. The drift test reports such windows as undetermined rather than passing.

## Documentation PR

N/A: internal routing fix, no public API change.

## Type of Change

Bug fix

## Testing

The routing was derived empirically by invoking all 55 models listed in `us-east-1` across both base paths and both API surfaces, against live Mantle; the 11 `/openai/v1` ids reduce to three line-scoped prefixes. Then verified end to end.

- Python: 263 passed in `test_openai.py` + `test_openai_responses.py` (up from 255), `ruff check` and `ruff format --check` clean on changed files.
- TypeScript unit: 52 passed in `mantle.test.ts` (up from 31), and all five `ts-check` gates clean: `build`, `lint`, `format:check`, `type-check`, `check:browser-bundle`.
- New integ tests, run live against Mantle in `us-east-1` on an account holding the `bedrock-mantle` grant: 5 passed in each SDK, including the full-catalog sweep.
- Real inference through **both** SDKs for all 11 `/openai/v1` ids plus 6 `/v1` control models, 17 each. Every one returned a correct answer on the resolved path, including `xai.grok-4.3`.
- Confirmed the fix is required: on the pre-fix code, `xai.grok-4.3` fails with ``400 model `xai.grok-4.3` isn't supported on this route``.
- Ids beyond the verified catalog are pinned by unit tests in each SDK: a point release within a verified line routes to `/openai/v1`, an unmatched new line falls to `/v1`.
- `test-infra`: 20 jest tests passed and `tsc` clean. Confirmed the new IAM assertion bites by deleting the `ListModels` line, which fails 1 of 20. `bedrock-mantle:ListModels` was checked against the AWS service reference rather than assumed.

* [x] I ran `hatch run prepare`

## Checklist

---


strands-py/tests_integ/models/test_mantle_routing.py
```diff
@@ -0,0 +1,127 @@
...
+        headers={"Authorization": f"Bearer {token}", "Content-Type": "application/json"},
+        method="POST",
+    )
+    try:
+        with urllib.request.urlopen(request, timeout=_TIMEOUT) as response:
+            return response.status
+    except urllib.error.HTTPError as e:
+        return e.code
+    except Exception:  # noqa: BLE001 - a hung route is "not served", same as a 400
+        return 0
```
strandly-the-agent, line 88: 🟡 **`_status` collapses "rejected" and "couldn't tell" into the same answer, so the sweep passes on ambiguity.**

`_serves` flags only a 200, so 400 (the real signal: *this route doesn't serve the model*) and 429 / 5xx / timeout / 403 all end up as "not served" → `wrong=False` → green.

That's the failure mode this issue already produced: the #3654 reporter hit transient `internal_server_error` on `/openai/v1` for `xai.grok-4.3`. Across ~55 models × up to 2 inference POSTs, 8-wide, a blip is ordinary noise rather than a contrived input.

Driven against a local fake Mantle **with real drift present** (an unlisted model served from `/openai/v1`):

| probe status | outcome |
|---|---|
| 200 | ✅ fails, names the id |
| 500 / 429 / timeout / 403 | ❌ `5 passed` |

TS behaves identically (`mantle-routing.test.node.ts:45`).

Suggested shape — only 400 is definitive; anything else must not read as clean:

```python
# 400 => genuinely not served on this route; 200 => served; anything else => unknown
status = _status(...)
if status not in (200, 400):
    indeterminate[model_id] = status   # retry with backoff, then fail on what's left
```

A detector that reports "no drift" when it means "couldn't determine" is worse than none, because the `Verified against the us-east-1 catalog on 2026-08-05` line in the table then looks CI-backed when it isn't.

strandly-the-agent (commented):
## Follow-up review — Mantle base-path routing

**Re-anchored to `e7e7708`.** My inline comments were written against `d37f458`; the follow-up commit ("scope Mantle path prefixes to a model line, not a vendor") already addresses two of them and answers my Question 1 — narrowing `xai.` → `xai.grok-4.` is the right call. Nothing here blocks merge; the routing change is correct and I could not break it.

### 🟡 New, and it outranks the original four: the drift detector will self-skip on every CI run

Both detectors gate on `GET /v1/models` and skip when it returns 401/403. The integ role defined in this repo — `test-infra/lib/constructs/integ-test-role.ts:162-163` — grants `bedrock-mantle:CreateInference` and `bedrock-mantle:CallWithBearerToken`, but **not `bedrock-mantle:ListModels`**, and no other statement in the repo adds it. If `secrets.STRANDS_INTEG_TEST_ROLE` is that role (the OIDC trust shape matches `repo:strands-agents/harness-sdk:*`; the ARN isn't visible from the repo, so this is inference rather than proof), then `python-integration-test.yml:137` and `typescript-integration-test.yml:148,163` will report every new test as *skipped*, and "future drift becomes a CI failure" never happens.

Note the interaction with `e7e7708`: making the TS side skip cleanly — which I asked for — closes the red-build failure mode and opens this silent one. Two things would settle it: add `bedrock-mantle:ListModels` to the integ role, and make the sweep assert it actually ran (fail, don't skip, when the catalog is unavailable *in an environment that is supposed to have it*) so a permissions regression can't switch the detector off quietly.

### Status of my four inline findings at `e7e7708`

| # | Finding | Status |
|---|---|---|
| 1 | Ambiguous HTTP status folded into "not served" | **Open** — `_status`/`status()` unchanged in both SDKs |
| 2 | Oracle probes only the unused path (served on neither / both) | **Open** — unchanged in both SDKs |
| 3 | TS/Python skip asymmetry on 401/403 | **Fixed** — `listModels` returns `null`, both tests `ctx.skip`. Residual: a 500 still throws `TypeError ... reading 'map'` |
| 4 | Prefix hedge untested | **Mostly fixed** — TS + `test_openai.py:2046` cover it; `test_openai_responses.py` didn't get the same case |

Findings 1 and 2 compound with the new one: if the sweep does start running, ambiguity still reads as "clean".

**Question still open:** `mantle.ts` and `_openai_bedrock.py` hold two independently-edited literals with nothing tying them together, and path-filtered integ CI means a one-SDK edit never runs the sibling's check. Worth a single source of truth (one data file generated into both), or is the per-SDK live check enough?

<details>
<summary>Verified — what I actually ran</summary>

At `d37f458` (inline findings) and re-checked at `e7e7708` where noted:
- ✅ TS `mantle.test.ts` 48/48 and Python Mantle-selected 51/51 green
- ✅ the two tables are byte-identical across SDKs (11 ids; prefixes `openai.gpt-5.` + `xai.grok-4.` at head)
- ✅ no regression — captured the real outgoing URL through `model.stream()` in both SDKs: `openai.gpt-oss-*` still `/v1`, `openai.gpt-5.*` still `/openai/v1`, `xai.grok-4.3` and the three Gemma 4 ids now `/openai/v1`
- ✅ ~30 mangled id variants (case, whitespace, `us.`/`global.` prefixes, `:0` suffixes, inference-profile ARNs, truncated relatives) produce no mis-route in either SDK
- ✅ `resolveMantleBasePath` stays internal (absent from the `models/openai` barrel and the package exports map) — no public API change, no api-review label needed
- ⚠️ live Mantle is unreachable from my sandbox (IAM 401 on `bedrock-mantle:*` — the same gap as the CI finding above), so the detectors were driven against a **local fake Mantle**, never the real service

</details>

<details>
<summary>Appendix — non-blocking (7)</summary>

- ⚪ **Repo-rule violation:** `AGENTS.md:57` bans "used to"/"previously" comment narration and says it applies to tests too (`strands-py/docs/TESTING.md`, `strands-ts/docs/TESTING.md` give this exact anti-pattern). Three added comments narrate that way: `strands-py/tests/strands/models/test_openai.py:2011`, `strands-py/tests/strands/models/test_openai_responses.py:1826`, `strands-ts/src/models/openai/__tests__/mantle.test.ts:191`. They already link #3654 — stating the guaranteed behavior instead of the history satisfies the rule.
- ⚪ **Test parity within the diff:** the prefix-scoping regression case added at `test_openai.py:2046` (`xai.grok-4.9` → `/openai/v1`, `xai.grok-5` → `/v1`) has no counterpart in `test_openai_responses.py`. Both classes delegate to the same resolver, so it's a pin rather than coverage — but the rest of that file mirrors its sibling.
- ⚪ **Pre-existing, filed as #3667:** TypeScript keeps the construction-time Mantle base URL after `updateConfig({ modelId })`, so a runtime model swap mis-routes (`model.ts:104`, `:297`), while Python re-resolves per request. Not this PR's problem — `model.ts` isn't in the diff — but this PR widens the blast radius from "gpt-5 vs everything" to any swap across the 11-id table. Reproduced first-hand.
- ⚪ **Cost:** the sweep re-mints a token per model, and both tests re-`GET /v1/models` per parametrized case — on the order of 150 live calls and ~100 real inference POSTs per run. Hoisting the mint + catalog into a session-scoped fixture trims most of it.
- ⚪ TS fires the whole catalog at once (`Promise.all`) where Python caps at 8 (`_MAX_WORKERS`) — so TS is the likelier of the two to trip the 429s that finding 1 shows get absorbed silently.
- ⚪ 8 of the 11 exact-set entries are already subsumed by the prefixes. Behaviorally inert; `e7e7708`'s added comment now explains the intent, which was the point.
- ⚪ `pytest.skip` from a `ThreadPoolExecutor` worker does surface as SKIPPED, but `shutdown(wait=True)` means the full catalog sweep still runs before it's reported. Not a defect — just slow to bail.

</details>

<details>
<summary>Pipeline record — how this review was produced</summary>

Stage 1 — **triage**: `accept`. Routed: correctness, issue-alignment, adversarial, test-quality. Skipped with reason: **api-bar-raiser** (no public surface — Python symbols underscore-private, TS `resolveMantleBasePath` `@internal` and absent from the barrel/exports map), **llm-context** (no model-facing text), **docs-accuracy** (no user-facing docs; comments only). Size route: all hunks in the two source files, then tests.

Stage 2 — **context build**: caller/CI/repo-rule pack. This is what caught the moved head and the `ListModels` gap above.

Stage 3 — **specialist passes**: correctness · test-quality · adversarial (fake-Mantle harness) · **issue-alignment**.

**Issue-alignment ledger** (over `cd98383..e7e7708`, read against the #3654 thread rather than the PR's framing) — verdict **aligned, scope clean, no out-of-contract hunks**:

| Hunk | Class |
|---|---|
| `_openai_bedrock.py` id set + narrowed prefixes + resolver | in-contract |
| `mantle.ts` same table/logic + exported `resolveMantleBasePath` | in-contract (cross-SDK parity) |
| the three unit-test files | supporting — pin the issue's own repro (`xai.grok-4.3`) |
| `test_mantle_routing.py`, `mantle-routing.test.node.ts` (new, ~250 lines) | supporting — recurrence prevention the #3654 thread explicitly asked for, not scope creep |

Also checked: the abandoned `'xai.'/'google.'/'anthropic.'` prefix proposal from the issue body was falsified before landing (Gemma 3 vs 4 split `google.`), the Anthropic exclusion is correct (different protocol via `AnthropicModel`), nothing in the contract is left unfixed, and neither PR claim ("resolves the second half of #3654", "unblocks the TypeScript half of #3563") overstates what shipped.

</details>

<sub>Automated review — solid work for a human to approve, not a gate.</sub>

strands-py/tests_integ/models/test_mantle_routing.py
```diff
@@ -0,0 +1,127 @@
...
+    token = _token()
+    models = [model_id for model_id in _list_models(token) if not model_id.startswith(_NOT_OPENAI_COMPATIBLE_PREFIXES)]
+    assert models, "Mantle returned no OpenAI-compatible models"
+
+    def check(model_id: str) -> tuple[str, str, bool]:
+        resolved = _resolve_mantle_base_path(model_id)
+        unused = "/openai/v1" if resolved == "/v1" else "/v1"
+        # Probe only the path the SDK would *not* use: if that one answers, we are wrong.
+        # Minting per model keeps long sweeps inside the token's lifetime.
+        return model_id, resolved, _serves(unused, model_id, _token())
```
strandly-the-agent: 🟡 **Probing only the unused path leaves two states unrepresentable** — both proven against the fake Mantle:

- **Served on neither path** (a future family like `anthropic.*` that isn't in `_NOT_OPENAI_COMPATIBLE_PREFIXES` — e.g. a synthetic `moonshotai.kimi-k3`): the unused path stays silent, so the sweep passes, while an `OpenAIModel` user with that id gets exactly the 400 this PR exists to prevent.
- **Served on both paths**: the sweep fails naming the id, and no table edit can satisfy it — adding the id just flips which path is "unused", and that one answers too. I ran both table states for a dual-served id and both fail, so the failure message sends a maintainer to edit a table that cannot help.

Asserting the positive as well — *the resolved path serves this model* — closes the first case and gives the second somewhere to land ("serves on both, either is fine"). Probe count is roughly unchanged if you probe the resolved path first and only fall through to the other when it doesn't serve.

strands-ts/test/integ/models/openai/mantle-routing.test.node.ts
```diff
@@ -0,0 +1,122 @@
...
+  return responses === 200
+}
+
+async function listModels(token: string): Promise<string[]> {
+  const response = await globalThis.fetch(`${BASE}/v1/models`, {
+    headers: { Authorization: `Bearer ${token}` },
+    signal: AbortSignal.timeout(TIMEOUT_MS),
+  })
+  if (response.status === 401 || response.status === 403) {
+    return []
```
strandly-the-agent: 🟡 **The TS and Python detectors disagree about the same missing permission.** Python skips cleanly (`test_mantle_routing.py:54` — `pytest.skip("account lacks bedrock-mantle:ListModels")`); returning `[]` here makes the two tests in this file diverge:

- the sweep hard-fails at `:86` with `expected 0 to be greater than 0` — nothing hints at a permission problem;
- the `it.each` cases hit `if (!models.includes(modelId)) return` (`:116`) and pass vacuously, asserting nothing.

`describe.skipIf(bedrock.skip)` doesn't cover it: `bedrock.skip` is false whenever *any* AWS credentials resolve, and `bedrock-mantle` is a separate grant. That's exactly my own sandbox role's shape (credentials fine, 401 on `bedrock-mantle:ListModels`) — so worth confirming the integ CI role holds that grant, or the first run lands here.

Related, two lines down: on a 500 the body has no `data`, so `:76` throws `TypeError: Cannot read properties of undefined (reading 'map')` rather than anything diagnostic.

Mirroring Python's skip — vitest's test-context `ctx.skip()`, or throwing a tagged error both tests catch — makes the SDKs agree and names the cause.

strands-ts/src/models/openai/__tests__/mantle.test.ts
```diff
@@ -187,6 +187,33 @@ describe('OpenAIModel bedrockMantleConfig', () => {
...
+      ['openai.gpt-oss-120b', '/v1'],
+      ['openai.gpt-oss-safeguard-20b', '/v1'],
+      ['qwen.qwen3-32b', '/v1'],
+      ['deepseek.v3.2', '/v1'],
+      ['mistral.ministral-3-8b-instruct', '/v1'],
+      ['zai.glm-5', '/v1'],
+      ['moonshotai.kimi-k2.5', '/v1'],
+      ['minimax.minimax-m2', '/v1'],
+      ['nvidia.nemotron-nano-9b-v2', '/v1'],
+      ['writer.palmyra-vision-7b', '/v1'],
```
strandly-the-agent, line 211: 🟡 **The prefix hedge is untested.** Every `xai.*` / `openai.gpt-5.*` id in this list is also an exact entry in `OPENAI_PATH_MODEL_IDS`, so nothing exercises the prefix branch. Confirmed by mutation in two independent passes: deleting the prefix check from both SDKs leaves all 48 TS and 255 Python unit tests green. The live sweep can't cover it either — it only sees ids that exist today, and those are all in the exact set.

```suggestion
      ['writer.palmyra-vision-7b', '/v1'],
      // Prefix hedge: not in OPENAI_PATH_MODEL_IDS, matched only by a prefix.
      ['xai.grok-5-preview', '/openai/v1'],
      ['openai.gpt-5.9-unreleased', '/openai/v1'],
```

Same two rows belong in `strands-py/tests/strands/models/test_openai.py:2010-2027`.

strands-py/tests_integ/models/test_mantle_routing.py
yonib05: Fixed in 22bbb99. You were right that this was the load-bearing flaw: the detector could report "no drift" when it meant "could not tell", which is worse than not having it.

`_serves` is now three-valued. Only 200 ("served") and 400 ("this route does not serve the model") count as answers; 429/5xx/timeout retries with backoff and, if it never settles, returns `None`. Undetermined models are collected and asserted on *before* the misrouted and unserved checks, so an inconclusive sweep fails naming the ids instead of passing:

```
Mantle never returned a definitive 200/400 for these models, so their routing
could not be verified (transient 429/5xx/timeout): {...}
```

One knock-on: the retry backoff can push a drifted sweep past the global 90s `pytest-timeout`, so the sweep now carries `@pytest.mark.timeout(600)` (the regression cases get 180). A clean run is 15s, but the run that has to report is the slow one, and timing out would have reintroduced the same "failed for an unrelated reason" ambiguity.

strands-py/tests_integ/models/test_mantle_routing.py
yonib05: Fixed in 22bbb99. Both SDKs now probe the resolved path first and assert the positive, then only fall through to the other path when the resolved one does not serve. That makes four verdicts distinguishable instead of one:

- `ok`: the resolved path serves it. Covers the dual-served case you raised, which previously produced a failure no table edit could satisfy.
- `misrouted`: resolved is silent, the other path answers. Names the ids to add.
- `unserved`: neither path answers. Previously a silent pass; now fails pointing at `_NOT_OPENAI_COMPATIBLE_PREFIXES` and the possibility of a non-OpenAI protocol.
- `undetermined`: never settled on 200/400.

Your note that probe count stays roughly flat held, and it is actually cheaper: probing the resolved path first short-circuits the common case, so a clean sweep dropped from ~31s to ~11s in TS and ~37s to ~15s in Python.

strands-ts/test/integ/models/openai/mantle-routing.test.node.ts
yonib05: Fixed in e7e7708, and your prediction about the CI role was correct.

`listModels` now returns `null` rather than `[]` on 401/403, and both tests call `ctx.skip("account lacks bedrock-mantle:ListModels")`, matching the Python side. The `it.each` became `it.for` so the test context is available as the second argument. Also fixed the adjacent crash you flagged: a non-ok response or a body without `data` now throws with the HTTP status and body instead of `TypeError: Cannot read properties of undefined`.

On confirming the integ CI role holds the grant: it does not. In the run on e7e7708, both detectors reported **skipped**, not passed:

```
✓ test/integ/models/openai/mantle-routing.test.node.ts (5 tests | 5 skipped)
SKIPPED tests_integ/models/test_mantle_routing.py::test_mantle_base_path_table_matches_live_catalog
```

So the skip path is what CI actually exercises today, which makes the difference between skipping and hard-failing on `expected 0 to be greater than 0` the practical behavior rather than an edge case. It also means my "future drift becomes a CI failure" claim in the PR description is not yet true in CI, only locally against my own account where all 5 pass in each SDK. Granting `bedrock-mantle:ListModels` plus invoke to the integ role is what would close that, and it is outside this PR. Worth flagging to whoever owns the integ credentials, and I would rather state the gap than let the description imply coverage that is not there.

strands-ts/src/models/openai/__tests__/mantle.test.ts
yonib05: Fixed in 22bbb99, and I took your suggested rows.

Confirmed the gap first: on the pre-fix code, deleting the prefix branch from both SDKs left every unit test green in each. With these cases it fails 2 in TypeScript and 2 in Python, so the branch is now actually covered.

```
[xai.grok-4.9, /openai/v1],           // matched only by a prefix
[openai.gpt-5.9-unreleased, /openai/v1],
[xai.grok-5, /v1],                    // new line, deliberately uncovered
[xai.grok-5-preview, /v1],
```

I kept your `xai.grok-5-preview` row expecting `/v1` rather than `/openai/v1`, because e7e7708 narrowed the prefix from `xai.` to `xai.grok-4.` in response to the other bot review. A hypothetical grok-5 is now deliberately *not* covered by the hedge: it falls through to /v1 and trips the drift test, which is the signal to verify and add the id. So the row doubles as the assertion that the narrowing works. Same two-plus-two rows in `test_openai.py`.

strands-ts/test/integ/models/openai/mantle-routing.test.node.ts
```diff
@@ -0,0 +1,213 @@
...
+    'serves %s from the resolved base path',
+    { timeout: 120_000 },
+    async (modelId, ctx) => {
+      const token = await mintToken()
+      const models = await listModels(token)
+      if (models === null) {
+        ctx.skip('account lacks bedrock-mantle:ListModels')
+        return
+      }
+      if (!models.includes(modelId)) return
```
strandly-the-agent: **🟡 This goes green without asserting anything.**

`if (!models.includes(modelId)) return` is a bare `return`, so if one of these four ids leaves the catalog — renamed, delisted, regionally withdrawn, i.e. exactly the Mantle churn this PR exists to catch — vitest reports the case as **passed** having asserted nothing. The Python sibling does `pytest.skip(f"{model_id} is not in the {_REGION} catalog")` (`test_mantle_routing.py:202`) and surfaces as SKIPPED.

This is the same failure mode the PR description calls out for the grant — *"a detector that skips while the run stays green is worse than no detector"* — except a **pass** is worse than a skip, because a skip at least shows up in the summary.

I traced it to `d37f458` and it survived the `22bbb99` rework, so it predates the notes you just addressed; I missed it in my earlier passes rather than it being new.

```suggestion
      if (!models.includes(modelId)) {
        ctx.skip(`${modelId} is not in the ${REGION} catalog`)
        return
      }
```

strands-py/tests_integ/models/test_mantle_routing.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def ids(verdict: str) -> dict[str, str]:
+        return {model_id: resolved for model_id, outcome, resolved in results if outcome == verdict}
+
+    # Asserted before the two below so a transient blip cannot be mistaken for a clean
+    # sweep: an undetermined model is a gap in coverage, not a pass.
+    undetermined = ids("undetermined")
+    assert not undetermined, (
+        "Mantle never returned a definitive 200/400 for these models, so their routing "
+        f"could not be verified (transient 429/5xx/timeout): {undetermined}"
```
strandly-the-agent: **⚪ A permanent access gap is reported as a transient blip.**

`_DEFINITIVE = (200, 400)` (`:46`) means *every* other status funnels into `undetermined` — including 401/403/404, which are not transient. So a model the account simply isn't entitled to invoke fails **here**, blaming "transient 429/5xx/timeout", while the message that actually mentions entitlement is the `unserved` one below (`:189`, *"or the account lacks access"*).

Verified by driving the real `_serves`/`_status_settled` with scripted statuses in place of the network call:

```
  chat,responses | _serves | verdict
      (403, 403) |  None   | undetermined
      (401, 401) |  None   | undetermined
      (404, 404) |  None   | undetermined
      (429, 429) |  None   | undetermined
      (500, 500) |  None   | undetermined
          (0, 0) |  None   | undetermined
```

Not a correctness bug — failing loudly is the right call, and this is strictly better than the pre-`22bbb99` behaviour of silently reading 403 as "not served". It's the *diagnosis* that misleads: the first person to hit an unentitled model will go hunting for a Mantle outage.

Cheapest fix is to widen the message rather than the classification, e.g. `"...never returned a definitive 200/400, so their routing could not be verified (transient 429/5xx/timeout, or a permanent 401/403/404 — check entitlement for these ids)"`. Same applies to the TS mirror at `mantle-routing.test.node.ts:176-180`.

strandly-the-agent (commented):
## Follow-up review — re-anchored to `266e888`

**Verdict: approve.** Both notes are addressed, and I checked the load-bearing part actually *bites* rather than merely being present — deleting the `ListModels` line really does turn the CDK suite red. All five of my earlier findings are closed. Two new non-blocking items, both in the detectors' **diagnostics**: the routing fix itself is untouched since `e7e7708` and still correct.

| # | Earlier finding | Status at `266e888` |
|---|---|---|
| 1 | Ambiguity folded into "not served" | **Closed** — `_serves` is tri-state, `undetermined` asserted first |
| 2 | Oracle probed only the unused path | **Closed** — resolved path first; dual-served → ok, neither → `unserved` |
| 3 | TS/Python skip asymmetry on 401/403 | **Closed** at `22bbb99` |
| 4 | Prefix hedge missing from Responses | **Closed** — `test_openai_responses.py:1846-1858` |
| 5 | `bedrock-mantle:ListModels` CI grant | **Closed** — `integ-test-role.ts:167`, and verified to bite |

New, inline, neither blocking: 🟡 the TS regression test goes green *without asserting* when a model leaves the catalog · ⚪ a permanent 403 is reported as a transient blip.

<details>
<summary>✅ Verified — what I actually ran</summary>

- **The CDK grant bites** (the claim I most wanted to check, since it's the load-bearing one): baseline `npx jest` in `test-infra` → `20 passed`. Deleting the `'bedrock-mantle:ListModels'` line at `integ-test-role.ts:167` → `1 failed, 19 passed`, with `internal mode grants the Mantle actions the base-path drift tests need` as the failure. Restored → `20 passed`. Matches your report exactly.
- **Python**: `pytest tests/strands/models/test_openai.py tests/strands/models/test_openai_responses.py` → `263 passed`, matching your count. (My first run showed 54 ERRORs purely from my own environment — `openai` and `aws_bedrock_token_generator` weren't installed. Not a code problem.)
- **TS unit**: `mantle.test.ts` → `52 passed`, matching. Needed `--project unit-node` (the browser project wants Playwright) and an install under `/tmp`, because `/mnt/workspace` in my sandbox is a 1 GB mount and the monorepo `npm ci` hits `ENOSPC` there.
- **Routing logic untouched**: `git diff e7e7708 266e888 -- strands-py/src strands-ts/src/models/openai/mantle.ts` is empty, so my earlier correctness verification carries over rather than needing a redo.
- **Tables still byte-identical** across both SDKs at this head (11 ids, prefixes `openai.gpt-5.` and `xai.grok-4.`).
- ⚠️ **Not verified**: live Mantle is still unreachable from my sandbox, so the detectors were driven against *scripted* statuses, never the real service. And my independent adversarial pass on the reworked detector **did not complete** — I cancelled it after 5+ minutes with no output. So treat the two findings below as my own close reading plus the probe I ran, not the product of a full adversarial sweep.

</details>

<details>

- ⚪ **Evergreen comments** — the only thing I'm repeating from last pass, and purely cosmetic. `AGENTS.md:57` bans "used to"/"previously" narration and says it applies to tests; still present at `test_openai.py:2010`, `test_openai_responses.py:1825`, `mantle.test.ts:191`. The issue link is already there, so stating the guaranteed behaviour instead of the history satisfies the rule.
- ⚪ **Backoff sleeps after the final attempt** — `_status_settled` (`test_mantle_routing.py:95-100`) and `statusSettled` (`mantle-routing.test.node.ts:71-77`) both `sleep` on the last iteration before returning, so every unsettled surface pays an extra ~4 s it can never use. Noise-level individually, but it multiplies by surfaces × paths × models on a bad day.

</details>

<sub>Automated review — solid work for a human to approve, not a gate.</sub>

strands-py/tests/strands/models/test_openai.py
```diff
@@ -2005,6 +2005,68 @@ def test_bedrock_mantle_config_uses_openai_path_for_gpt5(self, openai_client, mo
         resolved = model._resolve_client_args()
         assert resolved["base_url"] == "https://bedrock-mantle.us-east-1.api.aws/openai/v1"
 
+    @pytest.mark.parametrize(
+        ("model_id", "expected_path"),
+        [
+            # Regression: these used to fall through to /v1, which Mantle rejects with
```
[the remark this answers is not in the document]
yonib05: Fixed in 522e242. Reworded all three mirrored regression comments to state the behavior guaranteed by #3654 without narrating the prior implementation.

strands-ts/test/integ/models/openai/mantle-routing.test.node.ts
yonib05: Fixed in 522e242. A regression model missing from the live catalog now calls `ctx.skip(...)`, matching Python, instead of returning with a passing test that asserted nothing. The live TypeScript detector passes 5/5.

strands-py/tests_integ/models/test_mantle_routing.py
yonib05: Fixed in 522e242 in both SDKs. The undetermined diagnostic now covers transient 429/5xx/timeouts and permanent 401/403/404 entitlement gaps; the unserved diagnostic no longer incorrectly attributes access failures. Also removed the unused final backoff sleep.

strands-py/src/strands/models/_openai_bedrock.py
```diff
@@ -26,17 +26,72 @@
...
+# user's application.
+#
+# This set is the *complete verified snapshot* of the ``/openai/v1`` catalog, not the
+# delta left over by ``_OPENAI_PATH_MODEL_PREFIXES``. Entries that a prefix already
+# matches are intentionally listed anyway: the prefixes are an additive forward-compat
+# hedge, so narrowing or removing one must never change how an already-verified id
+# routes. Add every id you verify, even a redundant one, and do not trim entries because
+# a prefix appears to cover them.
+#
+# Verified against the ``us-east-1`` catalog on 2026-08-05.
```
opieter-aws: The verbosity of this comment and similar comments on source files below offer little value to a reader of the code, and are instead reasoning from an LLM. Goes against the evergreen comment rule. Comments should be to-the-point, and only mention what cannot be inferred from code.

opieter-aws (changes requested):

strands-py/src/strands/models/_openai_bedrock.py
yonib05: Ack, fixed; will also add specific verbiage in the agent.md files and the contribution guidelines to strengthen this point.

yonib05: Pushed a second pass in 491e0dd that covers the test files too: the two drift-test file headers went from 29 lines to 12, the prefix-hedge test docstrings to one line each, and the module headers in both SDKs lost a paragraph that duplicated the comment at the lazy import site. The whole PR now carries 68 fewer comment lines than before this thread.

strands-py/src/strands/models/_openai_bedrock.py
```diff
@@ -26,17 +26,47 @@
...
+        # xAI. Both API surfaces.
+        "xai.grok-4.3",
+    }
+)
+
+# Forward-compatibility hedge: routes a new point release of a verified line before its id
+# is added to ``_OPENAI_PATH_MODEL_IDS``. Keep each prefix scoped to a single model line,
+# never a vendor, since one vendor's lines can split across base paths (Gemma 4 vs Gemma 3).
+# A new line falls through to /v1 and is caught by the drift test.
+_OPENAI_PATH_MODEL_PREFIXES: tuple[str, ...] = ("openai.gpt-5.", "xai.grok-4.")
```
opieter-aws: nit: why not inlcude `google.gemma-4-` here and remove `_OPENAI_PATH_MODEL_IDS`? Seems like adding the model IDs adds surface to maintain without a current need for it

yonib05: Done in 62e44ed, and you were right that it simplifies: the three line-scoped prefixes (`openai.gpt-5.`, `xai.grok-4.`, `google.gemma-4-`) cover the verified catalog exactly, so the id set was pure maintenance surface. Removed in both SDKs; `resolveMantleBasePath` is now a single prefix check, and the whole "prefix-shadowed entries are intentionally redundant" subtlety disappears with it.

Verified live in both SDKs after the change (5/5 each), and re-verified the drift detector still bites: removing `xai.grok-4.` makes the sweep fail naming `xai.grok-4.3`. One caveat carried over from the earlier thread: until the test-infra redeploy grants `ListModels`, the drift test skips in CI, so the guard for an unverified prefix bet going wrong is currently local runs plus the unit tests that pin every known id. Incidentally, a later rerun today caught `/openai/v1` timing out for `xai.grok-4.3` while `/v1` kept 400ing correctly, i.e. the intermittent flakiness from #3654 recurring, which the sweep now reports as undetermined instead of passing. PR description updated for the prefixes-only design.

opieter-aws (approved):
Change looks good to me now, but CI is failing, so this may need some work still!

PR #3636, opened 2026-08-04
docs(community): add Stigmer AWS method contracts tool

LNSHRIVAS:
Adds a community catalog page for [strands-stigmer](https://pypi.org/project/strands-stigmer/), a Strands tool that queries Stigmer's execution graph of AWS.

## Motivation

Strands agents working with AWS hit the same failures repeatedly: guessing IAM permissions (AccessDenied), silent pagination truncation, and missing async-waiter annotations. strands-stigmer gives agents verified method contracts with the exact IAM actions each method requires, plus generated least-privilege policies for common workflows. It's derived from machine-readable specs (botocore definitions + iann0036/iam-dataset), so it's ground truth rather than contributed content.

## Changes

- Adds \community/tools/stigmer.mdx\ catalog page
- Registers the page in \
avigation.yml\ under Community > Tools

## Verification

- \pip install strands-stigmer\ works (v0.1.0 on PyPI)
- Tool imports as a Strands \DecoratedFunctionTool\ from \strands import tool\
- The hosted MCP endpoint (\https://stigmer.network/mcp\) responds 200; a read-only endpoint is available at \/mcp-readonly\

PR #3668, opened 2026-08-05
fix(catalog): point strands-stigmer docsUrl at stigmer.network

LNSHRIVAS:
Small follow-up to #3636. The reviewer flagged that the merged \docsUrl\ is pinned to \/tree/master/...\, which breaks if the default branch is ever renamed. Pointing it at \https://stigmer.network\ instead is more durable and is the canonical docs home.

\\\yaml
-docsUrl: https://github.com/LNSHRIVAS/stigmer/tree/master/strands-stigmer
+docsUrl: https://stigmer.network
\\\

### DIFF
--- a/site/src/content/catalog/strands-stigmer.yaml
+++ b/site/src/content/catalog/strands-stigmer.yaml
@@ -5,5 +5,5 @@ languages:
   python:
     package: strands-stigmer
 github: https://github.com/LNSHRIVAS/stigmer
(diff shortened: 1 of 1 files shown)


yonib05 (changes requested):
I am not sure if I agree with this change; I would rather have a direct docs URL to the integration with the package over a generic URL.

yonib05 (approved):

PR #3675, opened 2026-08-06
fix(agent): resume nested agent-as-tool interrupts across rehydration

### LINKED ISSUE #3076: [BUG] Nested agent-as-tool interrupts don't resume across rehydration (stateless / distributed execution)

strandly-the-agent:
## Description

Resuming an interrupt raised inside a nested agent-as-tool only worked while the orchestrator and the sub-agent shared the same in-memory `Interrupt`. The resume path read the human's answer off the sub-agent's copy, which the tool executor had registered into the orchestrator by reference, so the two were literally the same object. Once either agent is rebuilt from storage — a stateless handler recreating every agent per request — they are two independent objects, and the answer written to the orchestrator's copy is invisible to the sub-agent's: the sub-agent re-raises the same interrupt forever and the approved tool never runs.

This implements the simplification agreed in #3008's thread, and supersedes the approach in that PR (no `ToolInterruptEvent` change, no `invocation_state` routing):

- **Answers travel as data.** Sub-agent interrupts propagate upward as copies whose ids are namespaced with a prefix derived from the outer tool use id, so ids stay unique when several sub-agents are invoked in one turn and the sub-agent-local id is recoverable by stripping the prefix. The prefix opens with an SDK-owned marker (`v1:agent_as_tool:`) and percent-encodes the tool use id, because both ids are model-derived: without the encoding a tool use id containing the separator matches another call's ids, and without the marker a tool use id of exactly `v1` matches every interrupt the orchestrator raised itself. On resume, the orchestrator's already-persisted `_interrupt_state.context["responses"]` are filtered by that prefix and mapped back to local ids.
- **An ephemeral sub-agent's interrupted turn is parked by the orchestrator.** `preserve_context=False` sub-agents cannot have a session manager (`_agent_as_tool.py` rejects that combination), so their turn is stored as one keyed entry in the orchestrator's own interrupt context and freed once it has been reinstated. A `preserve_context=True` sub-agent owns its state and keeps its turn in its own session; if it has none, that is warned about when the interrupt **parks** — before a human is asked for a response that could not be applied — and the resume reports an actionable error rather than silently dropping the answer.
- **A parked turn that cannot be reinstated keeps its interrupt pending.** The event loop clears an agent's whole interrupt record as soon as a turn ends, so a failed reinstate raises the interrupt again instead of failing the call: the turn stays parked behind a still-pending interrupt and the response can be applied on a later attempt. Only the interrupts the parked turn is *still waiting on* are re-raised — the orchestrator also holds ids the sub-agent has already finished with.
- **Parking an interrupt preserves interrupt-context keys the event loop does not own.** This matches how Graph/Swarm already treat their own interrupt context.

```
                                                          8 files changed, 1177 insertions(+), 88 deletions(-)
strands-py/src/strands/agent/_agent_as_tool.py           +329  -18
strands-py/src/strands/event_loop/event_loop.py           +30   -3
strands-py/src/strands/agent/agent.py                      +6    0
strands-py/src/strands/interrupt.py                        +2   -1
strands-py/tests/strands/agent/test_agent_as_tool.py     +772  -66
strands-py/tests/strands/event_loop/test_event_loop.py    +29    0
site/.../concepts/interrupts.mdx                           +7    0
site/.../concepts/multi-agent/agents-as-tools.mdx          +2    0
```

The adapter grew more than the behaviour change alone needs: seven static helpers that each re-derived the orchestrator and the id prefix from `invocation_state` were consolidated into one `_ParentCall` object, per the review note about internal helpers.

## Related Issues

Fixes #3076. Implements the design `mkmeral` settled on in #3008 and supersedes the snapshot-plumbing approach there. Co-authored with `seanalbert`, whose PR framed the problem and the stateless-Lambda use case (`Co-authored-by:` trailer is on the first commit).

## Documentation PR

Included here rather than deferred. `preserve_context` decides who keeps a sub-agent's interrupted turn, and therefore whether a resume survives a restart, but it was documented purely as conversation history — and `agents-as-tools.mdx` is the only place customers see `preserve_context=True`, in a snippet with no session manager, which is the one configuration that cannot resume after a restart. `interrupts.mdx` documented multi-agent nesting for Swarm and Graph only, so agents-as-tools now has its own section covering id namespacing and where the turn lives. `Interrupt.id` is now documented as opaque, matching every sibling handle in the SDK.

## Type of Change

Bug fix

## Testing

**Behaviour: the four agent-as-tool configurations, each across a process boundary**

| sub-agent config | restart between turns | outcome |
|---|---|---|
| `preserve_context=False`, no session manager | yes | resumes — turn reinstated from the orchestrator's interrupt context, tool runs once |
| `preserve_context=True` + its own `FileSessionManager` | yes | resumes — answer mapped as data; the sub-agent's own message log is untouched |
| `preserve_context=True`, no session manager | no (same process) | resumes in memory |
| `preserve_context=True`, no session manager | yes | does not resume — warns at park time, logs at ERROR, and returns a tool error that tells the model the action did not run |

A and B are now both pinned by end-to-end tests with a real `FileSessionManager` and every agent rebuilt between turns; B is the configuration issue #3076 describes literally, and it fails at the merge base (verified: `turn2 stop_reason=interrupt executions=[]` on `a10881c71`, `end_turn` / `['prod-db']` at this head).

**Gates** — run directly; `hatch` is not available in my environment, and this sandbox kills any single shell command at ~60s, so the suite was run in four chunks rather than one invocation:

```
$ python -m pytest tests -q -n 4          # in 4 chunks covering every directory under tests/
920 + 1361 + 1399 + 1165 = 4845 passed, 0 failed

$ ruff format --check <6 changed py files>     6 files already formatted
$ ruff check src tests                         All checks passed!
$ mypy ./src                                   Success: no issues found in 234 source files
```

`ruff format --check src tests` across the whole tree reports `13 files would be reformatted` — byte-identical at the merge base, so pre-existing and untouched by this PR.

**What the new tests pin down, and that they fail without the fix.** Every one of these was verified by reverting the specific hunk and watching the test fail:

| test | negative control |
|---|---|
| `test_nested_interrupt_resumes_after_rehydration` (#3076 regression) | fails on `main`: `assert [] == ['prod-db']` |
| `test_nested_interrupt_resumes_after_rehydration_with_a_sub_agent_session_manager` | fails at the merge base — loops instead of resuming |
| `test_nested_interrupt_survives_a_parked_turn_that_fails_to_load` | fails when the re-raise is reverted to a terminal error result |
| `test_nested_interrupt_that_reraises_twice_runs_each_confirmed_action_once` | fails when the re-raise is not filtered to what the parked turn awaits |
| `test_stream_resume_reraises_the_interrupt_when_the_parked_turn_cannot_be_loaded` | same two mutations |
| `test_stream_resume_reraises_only_the_interrupts_the_parked_turn_still_awaits` | fails on the unfiltered re-raise |
| `test_namespaced_interrupt_ids_are_not_captured_by_a_tool_use_id_of_the_scheme_marker` | fails with `_NAMESPACE_TAG = ""` |
| `test_namespaced_interrupt_ids_*` (2) | fail when the id escaping is reverted |
| `test_stream_interrupt_warns_when_a_context_preserving_sub_agent_has_no_session_manager` | fails without the park-time warning |
| `test_stream_interrupt_parks_a_turn_that_is_isolated_from_the_sub_agent` | fails without the deep copy |
| `test_event_loop_cycle_interrupts_preserved_when_after_tools_hook_raises` | fails when the after-tools rescue call site replaces the whole context |

Two tests were **fixed rather than added**, because they passed while the behaviour they named was untrue: `test_stream_resume_keeps_stored_turn_when_it_cannot_be_loaded` asserted an in-memory dict on a mock orchestrator and could not see that the event loop then cleared it (replaced by the re-raise test above), and `test_stream_interrupt_resume_skips_state_reset` constructed the tool *after* setting the sub-agent's messages, so its reset baseline equalled the expected result and a reset would have been invisible. Two tests for a removed one-line private accessor were deleted; the branch that used it is covered by the resume tests.

Also exercised by hand on this branch: partial answers, three-level nesting, two sub-agents whose local interrupt ids collide, a sub-agent with a `SummarizingConversationManager` and one with a stateful model (`conversation_manager_state` and `model_state` both intact after a real restart, against a genuinely fresh manager/model as the control), a `Graph` node whose agent has an interrupting agent-as-tool, and a check that nothing new reaches the caller.

## Checklist

## Review loop ledger

| round | pass | findings | outcome |
|---|---|---|---|
| 1 | independent correctness review | 2 should-fix, 3 nits | both fixed (`b3dc2cd0`); 2 nits applied, 1 declined |
| 2 | adversarial testing, 5 probes | 1 routing bug, 1 doc gap | fixed (`40e08f22`) |
| 3 | verification of the fix commits | APPROVE, 2 notes + 1 nit | applied (`b097fc36`) |
| 4 | **six independent fresh-context passes** (correctness · adversarial · test-quality · API/DevX · issue-alignment · LLM-context) | **1 blocker**, 5 should-fix | all fixed or answered below (`65d3730`…`30f0355`) |
| 5 | **adversarial + test-quality review of those fixes** | **1 blocker in the fix itself**, 2 judgement calls, 2 nits | blocker fixed (`21fd240`); rest below |

Round 4 found what rounds 1–3 missed, and round 5 found a defect in round 4's own fix — a sub-agent that interrupts twice had its first, already-consumed interrupt id re-raised, which re-entered this very bug. Both are now fixed with tests that fail without them. Rounds 1–3 were dispatched by the same identity that wrote the code; rounds 4–5 were not given that assumption.

### Declined, with reasons

- **Broad `except Exception` in the reinstate path.** It guards deserializing persisted data possibly written by a different SDK version, where the failure modes are open-ended; narrowing it would let an unanticipated type escape into the adapter's generic handler, which reports a plain tool error at `WARNING` — exactly the silent loss of a human's approval this PR exists to prevent. It logs at ERROR and returns a specific message, and Graph/Swarm call `_InterruptState.from_dict` with no guard at all.
- **No issue link on tests written during development.** Per the root `AGENTS.md`, only a regression test for a *filed* bug links its issue; the two regression tests for #3076 do link it.
- **Stamping the tool name into a propagated interrupt's `name`.** It would make two sub-agents' interrupts easier to tell apart in an approval UI, but `name` is caller-visible and matched on, so prefixing it risks breaking name-based routing to fix a legibility problem. The tool call is already identifiable from the id, and a structured field is the better answer — worth its own change.
- **A retry cap on re-raising.** See the first residual: bounded retries end in discarding the human's answer, which is the destructive outcome this path exists to avoid.

## Residual / known limitations

- **A permanently unreadable parked turn re-prompts indefinitely.** If a reinstate can never succeed — e.g. a sub-agent's conversation-manager class changed while its turn was parked — every resume re-raises the interrupt, so the human is asked again each time and the only machine signal is the `ERROR` log. This is deliberate: the alternative is to give up after N attempts and discard an approval a human already gave, and a recoverable annoyance beats destroying the answer. An operator who fixes the underlying cause gets a working resume. **Worth a maintainer's opinion.**
- **Sharing one `Agent` instance as a tool across several orchestrators** is a foot-gun: `_reset_agent_state` is unconditional, so one orchestrator's call can clear a turn another has parked. The parked orchestrator still resumes correctly; the other call fails with a confusing error. Documented on that method rather than fixed, because making the reset aware of a foreign parked turn is a behaviour change deserving its own review.
- **A `load_snapshot` that fails after validation** leaves the sub-agent with some fields applied and others not. The parked turn is kept, the interrupt stays pending, and an ephemeral sub-agent is reset on its next fresh call, so it self-heals. It is not invoked in that state unless it is itself still activated from an earlier turn in this process; noted in the code.
- **Interrupt ids parked by an earlier revision of this branch** use the older prefix and are not recognised after the marker change: the resume treats the call as fresh and re-runs the sub-agent's turn. That only affects state persisted by a revision of this branch, which has never shipped, so no migration is included.
- **Not automated:** two sub-agent calls interrupting concurrently in one turn, three-level nesting, and a resume prompt mixing answers for a sub-agent's interrupts with the orchestrator's own. All three were exercised by hand and hold; they are the coverage I would add next.

### DIFF
--- a/site/src/content/docs/user-guide/concepts/interrupts.mdx
+++ b/site/src/content/docs/user-guide/concepts/interrupts.mdx
@@ -481,10 +481,7 @@ Interrupts are supported in multi-agent patterns, enabling human-in-the-loop wor
 
 ### Agents as Tools
 
-An agent exposed as a tool with [Agents as Tools](./multi-agent/agents-as-tools.md) propagates its interrupts up to the orchestrator, so the orchestrator's caller sees them and answers them like any other interrupt. Two things are worth knowing:
-
-- **Interrupt ids are namespaced by the tool call.** A sub-agent's interrupt arrives with an id derived from the orchestrator's tool call, so two sub-agents interrupting in the same turn stay distinct. Treat the id as opaque: pass it back unchanged to resume, and use `name` and `reason` for anything you show a human.
-- **Where the interrupted turn is kept depends on `preserve_context`.** With the default `preserve_context=False` the orchestrator carries the sub-agent's interrupted turn, so the resume survives the orchestrator and sub-agent being rebuilt from storage. With `preserve_context=True` the sub-agent owns its own state and needs its own session manager for the resume to survive a restart.
+An agent exposed as a tool with [Agents as Tools](./multi-agent/agents-as-tools.md) propagates its interrupts up to the orchestrator, so the orchestrator's caller sees them and answers them like any other interrupt. A sub-agent's interrupt arrives with an id scoped to the orchestrator's tool call, which keeps two sub-agents interrupting in the same turn distinct. Treat the id as opaque: pass it back unchanged to resume, and use `name` and `reason` for anything you show a human.
 
 ### Swarm
 
--- a/site/src/content/docs/user-guide/concepts/interrupts.mdx
+++ b/site/src/content/docs/user-guide/concepts/interrupts.mdx
@@ -479,6 +479,13 @@ To collect additional information from a user during an MCP tool call, use elici
 
 Interrupts are supported in multi-agent patterns, enabling human-in-the-loop workflows across agent orchestration systems. The interfaces mirror those used for single-agent interrupts. You can raise interrupts from `BeforeNodeCallEvent` hooks executed before each node or from within the nodes themselves. Session management is also supported, allowing you to persist and resume your interrupted multi-agents.
 
+### Agents as Tools
+
+An agent exposed as a tool with [Agents as Tools](./multi-agent/agents-as-tools.md) propagates its interrupts up to the orchestrator, so the orchestrator's caller sees them and answers them like any other interrupt. Two things are worth knowing:
+
+- **Interrupt ids are namespaced by the tool call.** A sub-agent's interrupt arrives with an id derived from the orchestrator's tool call, so two sub-agents interrupting in the same turn stay distinct. Treat the id as opaque: pass it back unchanged to resume, and use `name` and `reason` for anything you show a human.
+- **Where the interrupted turn is kept depends on `preserve_context`.** With the default `preserve_context=False` the orchestrator carries the sub-agent's interrupted turn, so the resume survives the orchestrator and sub-agent being rebuilt from storage. With `preserve_context=True` the sub-agent owns its own state and needs its own session manager for the resume to survive a restart.
+
 ### Swarm
 
 A [Swarm](./multi-agent/swarm.md) is a collaborative agent orchestration system where multiple agents work together as a team to solve complex tasks. The following example demonstrates interrupting your swarm invocation through a `BeforeNodeCallEvent` hook.
--- a/site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
+++ b/site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
@@ -154,6 +154,8 @@ orchestrator = Agent(
 </Tab>
 </Tabs>
 
+Interrupts raised inside a sub-agent resume automatically, but where the interrupted turn is kept depends on this setting. With the default <Syntax py="preserve_context=False" ts="preserveContext: false" />, the orchestrator carries the sub-agent's interrupted turn, so a resume works even when every agent is rebuilt from storage on the next request. With <Syntax py="preserve_context=True" ts="preserveContext: true" /> the sub-agent owns its own state: give it its own session manager if the resume has to survive a restart, otherwise the interrupt can only be resumed by the same process that raised it.
+
 ### Creating Custom Agent Tools
 
 For more control over how the agent is invoked — such as custom pre/post-processing, error handling, or passing multiple parameters — you can create a custom tool that wraps an agent:
--- a/site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
+++ b/site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
@@ -154,7 +154,9 @@ orchestrator = Agent(
 </Tab>
 </Tabs>
 
-Interrupts raised inside a sub-agent resume automatically, but where the interrupted turn is kept depends on this setting. With the default <Syntax py="preserve_context=False" ts="preserveContext: false" />, the orchestrator carries the sub-agent's interrupted turn, so a resume works even when every agent is rebuilt from storage on the next request. With <Syntax py="preserve_context=True" ts="preserveContext: true" /> the sub-agent owns its own state: give it its own session manager if the resume has to survive a restart, otherwise the interrupt can only be resumed by the same process that raised it.
+#### Advanced: Interrupts and Session State
+
+[Interrupts](../interrupts.md) raised inside a sub-agent resume automatically, and this setting decides who keeps the interrupted turn while it waits. With the default <Syntax py="preserve_context=False" ts="preserveContext: false" /> the orchestrator keeps it, so a resume works even when every agent is rebuilt from storage on the next request. With <Syntax py="preserve_context=True" ts="preserveContext: true" /> the sub-agent owns its own state: give it its own session manager if the resume has to survive a restart, otherwise the interrupt can only be resumed by the process that raised it.
 
 ### Creating Custom Agent Tools
 
--- a/strands-py/src/strands/agent/_agent_as_tool.py
+++ b/strands-py/src/strands/agent/_agent_as_tool.py
@@ -9,17 +9,17 @@ from __future__ import annotations
 import copy
 import logging
 import threading
-from collections.abc import Collection
 from typing import TYPE_CHECKING, Any
 from urllib.parse import quote
 
 from typing_extensions import override
 
 from ..agent.state import AgentState
-from ..interrupt import Interrupt
+from ..interrupt import Interrupt, _InterruptState
 from ..types._events import AgentAsToolStreamEvent, ToolInterruptEvent, ToolResultEvent
 from ..types._snapshot import Snapshot
 from ..types.content import Messages
+from ..types.exceptions import SnapshotException
 from ..types.interrupt import InterruptResponseContent
 from ..types.tools import AgentTool, ToolGenerator, ToolSpec, ToolUse
 
@@ -28,94 +28,40 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
-_PARKED_TURNS_KEY = "sub_agent_continuations"
-"""Key under which an orchestrator's interrupt context holds interrupted sub-agent turns, keyed by tool use id."""
+_INTERRUPTED_TURNS_KEY = "sub_agent_interrupted_turns"
+"""Key under which an orchestrator's interrupt context holds interrupted sub-agent turns, keyed by tool use ID."""
 
 _NAMESPACE_TAG = "v1:agent_as_tool:"
-"""Reserved marker opening every namespaced sub-agent interrupt id.
+"""Reserved marker opening every namespaced sub-agent interrupt ID.
 
 Written by the SDK and never derived from model output, so a namespace prefix cannot collide with an
-interrupt id the SDK generates itself - all of which open with the ``v1:`` scheme marker.
+interrupt ID the SDK generates itself - all of which open with the ``v1:`` scheme marker.
 """
-
-
-def _namespace_prefix(tool_use_id: str) -> str:
-    """Build the prefix that namespaces a sub-agent interrupt id to one agent-as-tool call.
-
-    Two sub-agents invoked in the same turn can raise the same interrupt id, because each derives it
-    from its own tool use id. The prefix keeps them distinct in the orchestrator's interrupt record and
-    keeps the sub-agent-local id recoverable by stripping it back off.
-
-    Tool use ids and interrupt ids are both model-derived, so the prefix defends against both ways one
-    call's prefix can match another id: the tool use id is percent-encoded, because otherwise one
-    containing the separator matches another call's ids, and the prefix opens with a reserved marker,
-    because otherwise a tool use id of ``v1`` matches every interrupt the orchestrator raised itself.
-
-    Args:
-        tool_use_id: Tool use ID of the agent-as-tool call.
-
-    Returns:
-        Prefix, separator included, for interrupt IDs belonging to that call.
-    """
-    return f"{_NAMESPACE_TAG}{quote(tool_use_id, safe='')}:"
-
-
-def _namespace_interrupts(tool_use_id: str, interrupts: list[Interrupt]) -> list[Interrupt]:
-    """Copy sub-agent interrupts with their ids namespaced to one agent-as-tool call.
-
-    Only the id changes: ``name`` and ``reason`` are what a human or an approval UI reads, so they are
-    passed through untouched.
-
-    Args:
-        tool_use_id: Tool use ID of the agent-as-tool call.
-        interrupts: Interrupts raised inside the sub-agent.
-
-    Returns:
-        Orchestrator-visible copies carrying namespaced ids.
-    """
-    prefix = _namespace_prefix(tool_use_id)
-    return [
-        Interrupt(id=f"{prefix}{interrupt.id}", name=interrupt.name, reason=interrupt.reason)
-        for interrupt in interrupts
-    ]
-
-
-def _parked_turn_interrupt_ids(turn: dict[str, Any]) -> set[str]:
-    """Get the sub-agent-local ids of the interrupts a parked turn is still waiting on.
-
-    Args:
-        turn: Serialized snapshot of an interrupted sub-agent turn.
-
-    Returns:
-        Interrupt IDs the sub-agent will raise again when the turn is reinstated.
-    """
-    interrupt_state: dict[str, Any] = (turn.get("data") or {}).get("interrupt_state") or {}
-    interrupts: dict[str, Any] = interrupt_state.get("interrupts") or {}
-    return set(interrupts)
 
 
 class _ParentCall:
     """The orchestrator's record of one agent-as-tool call.
 
-    Binds the orchestrator running a call to that call's interrupt id namespace, so propagation and
-    resume both read and write the orchestrator's persisted interrupt state in terms of *this* call:
-    the interrupts it has pending, the answers addressed to it, and the sub-agent turn it parked.
+    An interrupt raised inside a sub-agent has to reach the orchestrator's caller, and the human's
+    response has to travel back down - across a boundary where either agent may have been rebuilt from
+    storage, so the two no longer share an in-memory ``Interrupt``. Both directions therefore go through
+    the orchestrator's own persisted interrupt state, and everything here is scoped to one tool call
+    within it: the interrupts that call has pending, the responses addressed to it, and the interrupted
+    sub-agent turn stored for it.
+
+    Interrupt IDs are namespaced with a prefix derived from the tool use ID, because two sub-agents
+    invoked in the same turn can raise the same ID - each derives it from its own tool use ID. Both are
+    model-derived, so the prefix guards against either one matching another call's IDs: the tool use ID
+    is percent-encoded, or one containing the separator would match, and the prefix opens with a
+    reserved marker, or a tool use ID of ``v1`` would match every interrupt the orchestrator itself
+    raised.
     """
 
     def __init__(self, parent: Agent, tool_use_id: str) -> None:
         """Bind to an orchestrator and one of its tool calls."""
         self._parent = parent
         self._tool_use_id = tool_use_id
-        self._prefix = _namespace_prefix(tool_use_id)
-
-    @classmethod
-    def resolve(cls, invocation_state: dict[str, Any], tool_use_id: str) -> _ParentCall | None:
-        """Bind to the orchestrator running this tool call, if the invocation state carries one.
-
-        A tool invoked directly rather than by an agent has no orchestrator, and cannot interrupt.
-        """
-        parent: Agent | None = invocation_state.get("agent")
-        return cls(parent, tool_use_id) if parent is not None else None
+        self._prefix = f"{_NAMESPACE_TAG}{quote(tool_use_id, safe='')}:"
 
     @property
     def tool_use_id(self) -> str:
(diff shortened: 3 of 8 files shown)


mkmeral (dismissed):
I think we can clean up the code a bit more, especially internal helpers, but not a blocker for bugfix

strands-py/src/strands/agent/_agent_as_tool.py
```diff
@@ -24,6 +27,147 @@
...
+        interrupts: Interrupts raised inside the sub-agent.
+
+    Returns:
+        Orchestrator-visible copies carrying namespaced ids.
+    """
+    prefix = _namespace_prefix(tool_use_id)
+    return [
+        Interrupt(id=f"{prefix}{interrupt.id}", name=interrupt.name, reason=interrupt.reason)
+        for interrupt in interrupts
+    ]
```
opieter-aws: Do these really need to be their own functions?

strands-py/src/strands/agent/_agent_as_tool.py
```diff
@@ -24,6 +27,147 @@
...
+        parent: Agent | None = invocation_state.get("agent")
+        return cls(parent, tool_use_id) if parent is not None else None
+
+    @property
+    def tool_use_id(self) -> str:
+        """Tool use ID of the call."""
+        return self._tool_use_id
+
+    @property
+    def awaiting_resume(self) -> bool:
```
opieter-aws: There's a lot of new terminology introduced: `awaiting_ resume`, `parked` and `interrupted`. Should we clarify these terms and consolidate where needed so we're consistent through the class and codebase?

strands-py/src/strands/agent/_agent_as_tool.py
```diff
@@ -268,29 +462,111 @@ def _reset_agent_state(self, tool_use_id: str) -> None:
...
+                "interrupted turn failed to load. Do not report the requested action as completed or "
+                "successful. Tell the user it failed and ask them to respond again. See the logged error "
+                "for the cause."
+            )
+
+        return (
+            f"Agent '{self._tool_name}' did NOT run and the human's response was NOT applied: its "
+            "interrupted turn did not survive the restart. A sub-agent used with preserve_context=True "
+            "keeps its own state, so it needs its own session manager for that state to survive. Do not "
+            "report the requested action as completed or successful; tell the user it failed."
```
opieter-aws: Does the orchestrator know if `preserve_context=True` was used by the sub agent?

strands-py/src/strands/agent/_agent_as_tool.py
opieter-aws, line 443: Should this also reset `_interrupt_state`?

strands-py/src/strands/agent/_agent_as_tool.py
Unshure: +1

strands-py/src/strands/agent/_agent_as_tool.py
```diff
@@ -162,6 +338,7 @@ async def stream(self, tool_use: ToolUse, invocation_state: dict[str, Any], **kw
             prompt = str(tool_input)
 
         tool_use_id = tool_use["toolUseId"]
+        parent_call = _ParentCall.resolve(invocation_state, tool_use_id)
```
Unshure: Just general question on code readability: `_ParentCall.resolve` returns an object if `invocation_state.get("agent") is not None`, else it returns none. If that logic lived inline here, I think it would be easier for a human to read and follow the code since there are less jumps/redirects. This code is valid and fine, but curious if we should try to index on "human readability" of code in these reviews?

site/src/content/docs/user-guide/concepts/interrupts.mdx
```diff
@@ -479,6 +479,13 @@ To collect additional information from a user during an MCP tool call, use elici
...
 
+### Agents as Tools
+
+An agent exposed as a tool with [Agents as Tools](./multi-agent/agents-as-tools.md) propagates its interrupts up to the orchestrator, so the orchestrator's caller sees them and answers them like any other interrupt. Two things are worth knowing:
+
+- **Interrupt ids are namespaced by the tool call.** A sub-agent's interrupt arrives with an id derived from the orchestrator's tool call, so two sub-agents interrupting in the same turn stay distinct. Treat the id as opaque: pass it back unchanged to resume, and use `name` and `reason` for anything you show a human.
+- **Where the interrupted turn is kept depends on `preserve_context`.** With the default `preserve_context=False` the orchestrator carries the sub-agent's interrupted turn, so the resume survives the orchestrator and sub-agent being rebuilt from storage. With `preserve_context=True` the sub-agent owns its own state and needs its own session manager for the resume to survive a restart.
```
Unshure: This comment makes sense in context to this pr, but reads as a bit out of place in this doc page. If im reading this, I dont really care where the interrupt is kept, I just want to know that I can interrupt my agent as a tool. Im worried that including this might just confuse people

site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
```diff
@@ -154,6 +154,8 @@ orchestrator = Agent(
 </Tab>
 </Tabs>
 
+Interrupts raised inside a sub-agent resume automatically, but where the interrupted turn is kept depends on this setting. With the default <Syntax py="preserve_context=False" ts="preserveContext: false" />, the orchestrator carries the sub-agent's interrupted turn, so a resume works even when every agent is rebuilt from storage on the next request. With <Syntax py="preserve_context=True" ts="preserveContext: true" /> the sub-agent owns its own state: give it its own session manager if the resume has to survive a restart, otherwise the interrupt can only be resumed by the same process that raised it.
```
Unshure: Maybe put this under an `advanced` section? This feels like an implementation detail, and I dont think needs to be up front to customers.

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -701,6 +701,33 @@ async def terminal(ctx: InvokeModelContext) -> AsyncGenerator[Any, None]:
...
+
+    Args:
+        agent: Agent whose turn is being parked.
+        message: Tool use message that triggered the interrupt.
+        tool_results: Results of the tools that did complete.
+    """
+    carried = {
+        key: value for key, value in agent._interrupt_state.context.items() if key not in _LOOP_OWNED_CONTEXT_KEYS
+    }
+    agent._interrupt_state.context = {**carried, "tool_use_message": message, "tool_results": tool_results}
```
Unshure, line 728: The `**carried` is really the actual bug fix in this pull request right? We are propagating context from a tool that was previously not serialized and lost, right?

Do we need to worry about `carried` being serializable?

strands-py/src/strands/agent/_agent_as_tool.py
```diff
@@ -183,8 +360,51 @@ async def stream(self, tool_use: ToolUse, invocation_state: dict[str, Any], **kw
...
+                        # would end the orchestrator's turn, and the event loop clears the whole interrupt
+                        # record when a turn ends - taking the parked turn and the human's answer with it.
+                        logger.error(
+                            "tool_name=<%s>, agent_name=<%s>, tool_use_id=<%s> | cannot apply the interrupt "
+                            "response yet, raising the interrupt again so it survives to be answered once "
+                            "more",
+                            self._tool_name,
+                            self._agent_name,
+                            tool_use_id,
+                        )
```
Unshure, line 350: Does this error happen when a agent-as-tool has multiple interrupts, and only some of them have been responded to? If so, could we mention that in the error message?
```suggestion
                        logger.error(
                            "tool_name=<%s>, agent_name=<%s>, tool_use_id=<%s> | cannot apply the interrupt "
                            "response yet, there are still unanswered interrupts (maybe print interrupt ids here as well), raising the interrupt again so it survives to be answered once "
                            "more",
                            self._tool_name,
                            self._agent_name,
                            tool_use_id,
                        )
```

strands-py/src/strands/agent/_agent_as_tool.py
```diff
@@ -183,8 +360,51 @@ async def stream(self, tool_use: ToolUse, invocation_state: dict[str, Any], **kw
...
+                        yield ToolInterruptEvent(tool_use, pending)
+                        return
+
+                    logger.error(
+                        "tool_name=<%s>, agent_name=<%s>, tool_use_id=<%s> | cannot resume: the sub-agent's "
+                        "interrupted turn is not available, so the interrupt response cannot be applied",
```
Unshure, line 356: Seems like this happens if an orchestrator with a session manager uses a sub-agent that has `preserve_context=True`, but no session manager. If the sub-agent interrupts, and then the application stops and restores through session, then the application breaks here.

Is there some way we can prevent this early with errors? Or maybe pass session manager to the child agent? I guess returning a `toolResultEvent` does help avoid this, but it feels like its just masking the underlying issue that would keep coming up.

strands-py/src/strands/agent/_agent_as_tool.py
```diff
@@ -268,29 +496,112 @@ def _reset_agent_state(self, tool_use_id: str) -> None:
...
+        is held only in memory, which is worth warning about now rather than when the answer arrives and
+        cannot be applied.
+
+        Args:
+            parent_call: The orchestrator's record of this call.
+        """
+        if self._preserve_context:
+            if getattr(self._agent, "_session_manager", None) is None:
+                logger.warning(
+                    "tool_name=<%s>, agent_name=<%s>, tool_use_id=<%s> | interrupted sub-agent uses "
```
Unshure, line 518: Follow up from related comment, do we just want to throw in this case, or maybe inherit the parent agents session manager?

strands-py/src/strands/agent/_agent_as_tool.py
```diff
@@ -268,29 +496,112 @@ def _reset_agent_state(self, tool_use_id: str) -> None:
...
+        Returns:
+            True if the interrupted turn was reinstated onto the sub-agent.
+        """
+        turn = parent_call.parked_turn()
+        if turn is None:
+            return False
+
+        try:
+            self._agent.load_snapshot(Snapshot.from_dict(turn))
+        except Exception as error:
```
Unshure: Can we choose a more specific error here?

strands-py/src/strands/agent/_agent_as_tool.py
```diff
@@ -183,8 +360,51 @@ async def stream(self, tool_use: ToolUse, invocation_state: dict[str, Any], **kw
 
         try:
             # Determine if we are resuming the sub-agent from an interrupt.
-            if self._is_sub_agent_interrupted():
-                prompt = self._build_interrupt_responses()
+            if parent_call is not None and parent_call.awaiting_resume:
+                if not self._reinstate_turn(parent_call) and not self._agent._interrupt_state.activated:
```
Unshure: Im having a bit of trouble following the logic here:

In the documentation, you mention that interrupts are stored in different locations if `preserve_context` is set to true or false right? Where in this code block are we checking the interrupt context if `preserve_context=False` or the agents sessions if `preserve_context=True`?

Is there an opportunity to restructure the logic here to make it a bit easier to follow what is going in the different code paths?

strands-py/src/strands/agent/_agent_as_tool.py
strandly-the-agent: No — both are gone in `a61c9f3b1`. `_ParentCall` already derived the prefix, so namespacing is now a method on it (`parent_call.namespace(interrupts)`), and the parent lookup is inlined at its one call site per Unshure's related note. Three places that each re-derived the prefix collapsed into one owner, and the round is a net deletion rather than more structure.

strands-py/src/strands/agent/_agent_as_tool.py
strandly-the-agent: Agreed — and the constant was the worst of it: `_PARKED_TURNS_KEY = "sub_agent_continuations"` used two different words for one thing while the docstrings called it a third ("interrupted turn").

Settled on one: a sub-agent's **interrupted turn** is **stored** and **restored**. `parked`, `reinstate` and `continuation` are gone from the source, the tests and the persisted key (now `sub_agent_interrupted_turns`), and `awaiting_resume` reads `is_resuming`. `a61c9f3b1`

strands-py/src/strands/agent/_agent_as_tool.py
strandly-the-agent: No, and it doesn't need to — the adapter knows (`self._preserve_context`), and the adapter is what writes this message.

But you caught a real bug by asking: the message asserted `preserve_context=True` as the cause *unconditionally*, so an ephemeral sub-agent whose turn went missing was told to attach a session manager it isn't even allowed to have (construction rejects that combination). It now only names `preserve_context` when that is actually the cause. `a61c9f3b1`

strands-py/src/strands/agent/_agent_as_tool.py
strandly-the-agent: Yes — good catch, fixed in `a61c9f3b1` with a test.

It's safe because `_reset_agent_state` only runs on a **fresh** call, never on a resume, so any interrupt state still on the agent belongs to a turn nobody is resuming. Leaving it made the *next* call fail rather than the old one: the sub-agent believes it is mid-interrupt and rejects a plain prompt. Test is `test_stream_resets_stale_interrupt_state_on_a_fresh_call`, which fails if the reset line is removed.

strands-py/src/strands/agent/_agent_as_tool.py
strandly-the-agent: Agreed, and inlined in `a61c9f3b1`:

```python
# A tool invoked directly rather than by an agent has no orchestrator, and cannot interrupt.
parent = invocation_state.get("agent")
parent_call = _ParentCall(parent, tool_use_id) if parent is not None else None
```

`resolve` is gone. And yes — I think human readability is the right axis to index on here, especially for a private adapter where the extra indirection buys nothing. Your note plus `opieter-aws`'s two made me take a second pass at the whole shape: the same reasoning removed the two module-level helpers and restructured the resume block, so the commit is a net −6 lines.

site/src/content/docs/user-guide/concepts/interrupts.mdx
strandly-the-agent: Cut. The page now keeps only the part a caller has to act on — the id is opaque, pass it back unchanged, use `name`/`reason` for anything shown to a human — and the storage detail is gone from here entirely. It moved under an advanced heading on the agents-as-tools page, per your other comment. `a61c9f3b1`

site/src/content/docs/user-guide/concepts/multi-agent/agents-as-tools.mdx
strandly-the-agent: Moved to a `#### Advanced: Interrupts and Session State` subsection after Context Management, with the interrupts page linked for the general picture. `a61c9f3b1`

strands-py/src/strands/event_loop/event_loop.py
strandly-the-agent: Two answers, because they come apart.

**Is `**carried` the fix?** Necessary but not sufficient. Reverting just that line does return the ephemeral case to a silent no-resume (0 executions, verified), because the stored turn is discarded the moment the turn parks. But the human's response also has to *travel*: after a restart the orchestrator's and the sub-agent's `Interrupt` objects are different objects with the same id, so the ids are namespaced per tool call and mapped back on resume. `**carried` keeps the turn; the namespacing routes the response. Both are load-bearing.

**Serializability:** no new requirement. `carried` only ever holds what other components already put in `_interrupt_state.context`, and that whole dict was already persisted wholesale by `_InterruptState.to_dict()` into the session record — parking was *discarding* those keys, not failing to serialize them. The one value this PR adds is a `Snapshot.to_dict()`, which is JSON by construction (it is exactly what `SnapshotSessionManager` persists). So nothing here can become unserializable that wasn't already.

strands-py/src/strands/agent/_agent_as_tool.py
strandly-the-agent: Not quite, and the real trigger is narrower — worth correcting rather than documenting.

Partially-answered interrupts don't reach here: an unanswered one just means the sub-agent is re-invoked without a response for it and re-raises it, no error. This fires when the stored turn itself could not be **restored** (e.g. a snapshot the running build won't load).

I took the interrupt-ids part of your suggestion, which is useful either way — `a61c9f3b1`:

```
tool_name=<%s>, agent_name=<%s>, tool_use_id=<%s>, interrupt_ids=<%s> | the interrupted turn
could not be restored, so the response cannot be applied yet: raising the interrupt again so
it survives to be answered once more
```

strands-py/src/strands/agent/_agent_as_tool.py
strandly-the-agent: **"Prevent this early" — now we do**, as early as it is knowable. When the interrupted turn is stored, if a `preserve_context=True` sub-agent has no session manager, it warns *before the human is ever asked*:

```
interrupted sub-agent uses preserve_context=True with no session manager, so its interrupted
turn is held only in memory: the interrupt resumes in this process but not after a restart
```

At construction there is nothing to validate — we can't know whether an interrupt will happen, or whether the orchestrator persists anything.

**"Pass the session manager to the child"** is the real fix and I measured it rather than guessing: moving the gate from `preserve_context` to "the sub-agent has no session manager" is a ~2-line change that makes this exact case resume across a real restart and leaves the other three configurations byte-identical. The costs are that it makes the orchestrator a state store for an agent that owns its own state, and that `load_snapshot` would roll back a sub-agent that accumulated history between the interrupt and the resume. `mkmeral` deferred precisely that as its own feature ("that's a new feature… I'd rather focus on the bug ahead"), which is why it isn't here.

**On masking:** the tool result isn't the remedy, it's damage control — without it the response is dropped *and* the model narrates success over an action that never ran (I reproduced that against a live model earlier in this PR). Happy to land the inheritance version here instead if you and `mkmeral` would rather not wait for the follow-up.

strands-py/src/strands/agent/_agent_as_tool.py
strandly-the-agent: Same underlying question as your comment above, so one answer for both.

Throwing here would be worse than the error result: it surfaces as a generic tool failure mid-turn, the caller can't act on it, and the human's response is lost either way — with the error result at least the model is told the action did **not** run and not to report success.

Inheriting the parent's session manager is the fix I'd actually want, and it measures well (~2 lines; this case resumes, other configurations unchanged). It is the piece `mkmeral` deferred as a separate feature, so I've kept it out — say the word and I'll bring it into this PR.

strands-py/src/strands/agent/_agent_as_tool.py
strandly-the-agent: Done in `a61c9f3b1`:

```python
except (SnapshotException, ValueError, KeyError, TypeError) as error:
```

Those are the ways a turn written by a different build of the SDK actually fails to load: an unsupported schema version or scope (`SnapshotException`), a state a component rejects (`ValueError` — e.g. a conversation-manager class mismatch), and a field that has since been renamed or retyped (`KeyError`/`TypeError`). Anything outside that set is a bug rather than stale data, and now propagates instead of being swallowed.

For context, I'd pushed back on narrowing this twice in earlier rounds, on the grounds that an unanticipated type escaping would cost a human's approval. Naming the set is the better answer: it keeps the stale-data cases handled deliberately and stops the clause hiding real bugs.

strands-py/src/strands/agent/_agent_as_tool.py
strandly-the-agent: Fair — that was the least obvious part of the change. Restructured in `a61c9f3b1` so both paths are named where the decision happens:

```python
if parent_call is not None and parent_call.is_resuming:
    # The interrupted turn comes from one of two places: an ephemeral sub-agent's turn is
    # stored on the orchestrator, so restoring it reads the orchestrator's interrupt record,
    # while a preserve_context=True sub-agent restores its own from its session manager and
    # so arrives already activated.
    restored = self._restore_interrupted_turn(parent_call) or self._agent._interrupt_state.activated

    if not restored:
        ...
```

So, to answer directly: `_restore_interrupted_turn` **is** the `preserve_context=False` path — it reads the turn the orchestrator stored in its own interrupt context. The `preserve_context=True` path is the `activated` check: nothing is stored for that sub-agent, its own session manager restored its interrupt state, so it comes back already activated and there is nothing for the orchestrator to do. If neither produced a turn, the response can't be applied and we fall into the branch below.

PR #3571, opened 2026-07-30
feat(anthropic): enable prompt caching via cache_config and cache_tools

agent-of-mkmeral:
## Description

`AnthropicModel` had no way to *turn prompt caching on*. Only `BedrockModel` carried a caching switch, so a caller who configured caching and switched provider silently lost it. The translation half already existed — a `cachePoint` became `cache_control` — but nothing ever produced a cache point for it. This adds the enable half.

Two design decisions drive the shape, and both were argued down from earlier revisions of this PR:

**One config, not two.** An earlier revision gave TypeScript a separate `cacheTools` field mirroring Python's. That meant a config could not move between `AnthropicModel` and `BedrockModel` unchanged, which was the objection that reopened the design. Tool caching now lives inside `cacheConfig` on both providers.

**Per-section fields carry a TTL *or* `false`, rather than adding booleans.** Bedrock already shipped `toolsTTL`/`messagesTTL` as TTLs, and today `{ strategy: 'auto' }` caches *both* sections. A TTL alone cannot express "cache only this section" — so either sections default on (and tools-only stays impossible, the gap being fixed) or presence enables (and every existing `{ strategy: 'auto' }` silently stops caching). Widening the existing fields to `boolean | CacheTTL` gets tools-only with no migration and no new field.

`bedrock.py` is no longer byte-identical to `main`: it carries the system-prompt TTL fix described under *Behavior changes to note*, and nothing else. Happy to split that hunk into its own PR if you would rather keep this one Anthropic-only.

Python keeps its `cache_config` + `cache_tools` split. Unifying it there is a larger change with its own compatibility surface, deliberately left for a follow-up — so the two SDKs diverge on this config shape for now.

### API surface

TypeScript, exported from `@strands-agents/sdk`:

```typescript
type CacheTTL = '5m' | '1h' | (string & {})

interface CacheConfig {
  strategy?: 'auto' | 'anthropic'      // default 'auto'
  ttl?: CacheTTL                       // default: provider's own (Bedrock/Anthropic: 5m)
  toolsTTL?: boolean | CacheTTL        // default true
  messagesTTL?: boolean | CacheTTL     // default true
}
```

`AnthropicModelConfig` gains `cacheConfig?: CacheConfig`. `BedrockCacheConfig` and `BedrockCacheTTL` remain exported and are now aliases of `CacheConfig` / `CacheTTL`. `CacheToolsConfig` is removed from the TypeScript exports — it was added earlier in this same PR and never released.

[...]

PR #3677, opened 2026-08-06
fix(bedrock): honor a cache point placed in the last user message

strandly-the-agent:
## Description

Auto mode (`cache_config` / `cacheConfig` set) stripped every pre-existing `cachePoint` and appended a fresh one to the **end** of the last user message, in both SDKs. That is wrong whenever the last user message ends with content rebuilt on every call — injected context, a context-usage status line, a timestamp: the per-call content lands *inside* the cached prefix, so every request writes a new cache entry and none ever reads one.

The last user message now keeps exactly one cache point: the first one the caller placed, **honored where it sits**. Extra points in that message, and points in earlier messages, are still stripped so they cannot accumulate one per turn against the provider's cache-point budget. **Python and TypeScript both change**, to the same rule.

Measured against real Bedrock (`us.anthropic.claude-sonnet-4-5`, a caller marking a boundary ahead of per-call text, two turns):

```
before   [{'write': 3654, 'read': 0}, {'write': 3670, 'read': 0}]   turn 2 rewrites the whole prefix
after    [{'write': 3648, 'read': 0}, {'write':   16, 'read': 3648}]  turn 2 reads it
```

Weighted by Bedrock's cache pricing (write ≈1.25× input, read ≈0.1×) that is roughly a 1.8× cost reduction per warm turn. Note that **total token counts are nearly identical either way** — a write and a read both count as tokens — so only `cacheReadInputTokens` distinguishes a working cache from a thrashing one. The assertion to keep for any future placement change is "turn 2 reads > 0", not a token total.

### Behaviour changes (identical in both SDKs)

All of these apply **only when caching is configured**; otherwise the method never runs.

**Changed**

1. **A cache point in the last user message is kept where the caller put it** (previously deleted, with a new one appended at the end). This is the one user-visible change: code that hand-places a point *and* enables auto caching currently has its placement silently discarded and will now have it respected. Worth a release note.
2. The `WARNING "stripped existing cache point (auto mode manages cache points)"` no longer fires for a single point in that message — it becomes a `debug` "honored caller-placed cache point". It still fires for extra points in that message and for points in earlier messages.
3. No second point is added to that message.
4. The configured TTL (`cache_config.ttl` / `cacheConfig.messagesTTL`) is applied to an honored point that carries no TTL; a hand-placed TTL is left alone. Documented in the helper: the caller's explicit TTL is the more specific instruction, and Bedrock wants TTLs non-increasing across toolConfig → system → messages, so a caller placing a longer TTL owns that ordering.
5. An honored point directly preceded by non-PDF documents steps back over that **adjacent run only** — Bedrock rejects only that adjacency, and moving further would evict the document (usually the expensive part) from the cached prefix — and is dropped when nothing precedes them. A **PDF** document does not trigger this.
6. Cache reads begin working wherever a caller marks a boundary (the measurement above).
7. A last user message whose only content is a cache point now ships as `[{cachePoint}]` rather than being emptied (Python; TypeScript already did this).
8. A point placed *before* a `toolResult` in the same message is now preserved rather than incidentally repaired by the old strip-and-append. No first-party caller does this; noted for completeness.

**Unchanged**

9. Points in earlier messages are still stripped, with the same warning.
10. At most one cache point per request in the last user message; no new pressure on the 4-breakpoint budget.
11. Placement is byte-identical to before when the last user message carries no cache point, including the existing non-PDF-document relocation.
12. No change to `cache_prompt`, `cache_tools`, system/tool caching, or any other provider.

## Related Issues

No issue filed for this specific defect. It is the placement contract that #3642 (ContextInjector, which places a boundary ahead of injected context) and #3571 (Anthropic `cache_config`, which currently strips caller-placed points and re-places them after the last cacheable block) both need in order to agree with each other — today they implement opposite policies on identical input.

**Related, shipped, and not fixed here:** `Agent(context_manager="agentic")` with Bedrock caching thrashes the cache on `main` today — the per-call `<context-status>` block is appended to the last user message and the auto cache point goes after it. Measured `[{'write': 5789, 'read': 0}, {'write': 5803, 'read': 0}]`. This PR makes the fix possible (the middleware can mark the boundary) but does not apply it, to keep the contract change reviewable on its own.

## Documentation PR

Included here: `site/…/model-providers/amazon-bedrock.mdx` presented automatic (Option A) and manual (Option B) cache points as either/or. It now states that they compose and why you would place a point ahead of per-call content.

## Type of Change

Bug fix

## Testing

**Python**
- `pytest tests/strands/models/test_bedrock.py` → **193 passed** (12 new)
- `pytest tests/strands/models/test_bedrock.py tests/strands/agent tests/strands/event_loop` → **1027 passed**
- `ruff check` + `ruff format --check` → clean
- `mypy` was **not** run: it hits an INTERNAL ERROR on this machine's Python 3.13, so CI needs to cover it

**TypeScript**
- `vitest run --project unit-node src/models/__tests__/bedrock.test.ts` → **185 passed** (12 new), `Type Errors: no errors`
- `vitest run --project unit-node` (whole project, with coverage) → **141 files / 3953 passed**
- `eslint` clean, `prettier --check` clean, `npm run type-check` clean (both `src` and `test/integ` tsconfigs)

**Live + mutation**
- Live Bedrock A/B as above (`us-west-2`), with a pre-change control run
- **Mutation-checked on both sides, 13 mutants, all killed**: honour branch removed; extras not stripped; extras stripped ascending (stale index); relocation removed; relocation jumping to the first document instead of the adjacent run; relocation off-by-one either side; PDF exclusion deleted; configured TTL overwriting a hand-placed one; leading-document drop removed; strip loop skipping no message / the wrong message; honouring the last placed point instead of the first

Two independent review passes over the diff (one per SDK, fresh context, no write tools) ran their own tests and mutation batteries. The Python pass found three real defects, all fixed before this PR opened: multiple caller-placed points in one message all survived, breaking the one-point-per-request guarantee (146 violations in an exhaustive sweep of 6,449 message shapes); relocation fired even when a document did not *directly* precede the point, evicting durable content from the cached prefix — the inverse of the bug being fixed; and the new document helper duplicated an inline loop the injected path still used, which a surviving mutant proved was unguarded. The TypeScript pass approved the mirror as behaviourally exact (verified by a 24,867-case differential fuzz against the Python implementation, including log text and levels) and found two test gaps that existed on **both** sides — no test had three points in one message, and none pinned "adjacent run only" against "jump to the first document" — both now closed in both SDKs.

### Follow-up round: helper inlining (`93f61d379`)

Requested in review by `opieter-aws` ("remove single line functions"). The four one-line helpers this PR had introduced are gone, inlined at their call sites: Python `_is_non_pdf_document` / `_first_non_pdf_document_idx`, TypeScript `_isNonPdfDocument` / `_firstNonPdfDocumentIdx`. The first-non-PDF-document scan is restored to the inline loop `main` already had (those lines leave the diff entirely) and the relocation walk tests the predicate directly in its `while`.

- **Behaviour-preserving, verified:** differential run of the pre- and post-refactor `_inject_cache_point` over **14,007 message shapes** × 3 cache configs → identical mutated messages *and* identical log records.
- **New coverage:** with the predicate no longer behind a name, a mutant deleting the restored loop's `break` (find the *last* non-PDF document instead of the first, placing the cache point directly after a document — the ordering Bedrock rejects) survived every test. `test_inject_cache_point_before_the_first_of_several_non_pdf_documents` and its TS mirror close it. Battery over the inlined paths: **9/9 Python, 6/6 TypeScript killed**.
- **Gates:** Python 193 passed, `ruff check` + `ruff format --check` clean; TypeScript 141 files / 3953 passed with `Type Errors: no errors`, `lint` / `format:check` / `type-check` clean.
- **Pre-commit hook bypassed** (`--no-verify`): its `test:coverage` step ENOSPCs on the 1 GB sandbox. Its five steps were run manually instead — build, full `unit-node` suite with coverage redirected to `/tmp`, lint, format:check, type-check — all green. CI should exercise the hook's own path.
- **Review loop:** 1 round, fresh-context reviewer with no write tools on `93f61d379` → **APPROVE, 0 actionable findings**. It re-derived the negation in both languages (including the `undefined` case `noUncheckedIndexedAccess` forces), found zero stale references to the four deleted names, and rebuilt the `break` mutant itself to confirm exactly one test fails per SDK. One non-blocking tradeoff left open for a human: the "is a non-PDF document" check now exists twice per SDK rather than once behind a name, so a future second exempted format must be changed in lockstep at both sites.

## Checklist
- [ ] I have reviewed and understand every line of code in this PR, including any generated by AI tools, and I can explain why it works — **this PR was written by an agent; a human needs to own this box before merge**

----


strands-py/src/strands/models/bedrock.py
```diff
@@ -406,8 +406,61 @@ def _build_tools_cache_point(self) -> list[dict[str, Any]]:
...
+            del content[placed_idx]
+            if target_idx == 0:
+                # Nothing precedes the documents, so there is no prefix to cache.
+                logger.debug("msg_idx=<%s> | dropped cache point ahead of a leading document", msg_idx)
+                return
+            content.insert(target_idx, placed)
+            logger.debug("msg_idx=<%s>, block_idx=<%s> | relocated caller-placed cache point", msg_idx, target_idx)
+            return
+
+        logger.debug("msg_idx=<%s>, block_idx=<%s> | honored caller-placed cache point", msg_idx, placed_idx)
```
agent-of-mkmeral, line 477: 🔴 **This block ships a request Bedrock rejects when `placed_idx == 0`.**

`target_idx == placed_idx` when the caller put the point first, so control falls straight to the `honored` log and the point stays at index 0. Bedrock rejects that — verified live on `us.anthropic.claude-sonnet-4-5`, with two prior turns present:

```
[cachePoint, text] -> ValidationException: There is nothing available to cache.
[text, cachePoint] -> OK, write=2639 then read=2639   (what main emits for the same input)
```

The rule is intra-message: a cache point needs ≥1 content block *before it in the same message*, and earlier messages do not count. That also means your `target_idx == 0` drop is correct — I tested keeping the point at 0 and Bedrock refuses it — but the justification comment ("no prefix to cache") undersells it, and the same condition needs to apply when the caller places at 0 directly.

Hoisting the zero-check ahead of the relocation branch covers both, and returning a bool lets the caller fall through to automatic placement (see my comment on the call site) so `[cachePoint, text]` keeps working as it does on `main`:

```suggestion
        if target_idx == 0:
            # Bedrock rejects a cache point with no content before it in the same message:
            # "There is nothing available to cache." Earlier messages do not count.
            del content[placed_idx]
            logger.warning("msg_idx=<%s> | cannot honor a cache point with nothing before it", msg_idx)
            return False

        if target_idx != placed_idx:
            del content[placed_idx]
            content.insert(target_idx, placed)
            logger.debug("msg_idx=<%s>, block_idx=<%s> | relocated caller-placed cache point", msg_idx, target_idx)
            return True

        logger.debug("msg_idx=<%s>, block_idx=<%s> | honored caller-placed cache point", msg_idx, placed_idx)
        return True
```

The signature on line 415 becomes `-> bool`. Verified: 193 tests still pass (including `test_inject_cache_point_drops_an_honored_point_when_a_document_leads_the_message`), index-0 violations across my 6,219-shape sweep go 1035 → 30, and the live request becomes `OK write=0 read=2639`.

strands-py/src/strands/models/bedrock.py
```diff
@@ -437,6 +494,20 @@ def _inject_cache_point(self, messages: list[dict[str, Any]]) -> None:
...
+                # would spend the provider's cache-point budget for nothing.
+                for extra_idx in reversed(placed_idxs[1:]):
+                    del content[extra_idx]
+                    logger.warning(
+                        "msg_idx=<%s>, block_idx=<%s> | stripped existing cache point (auto mode manages cache points)",
+                        last_user_idx,
+                        extra_idx,
+                    )
+                self._honor_placed_cache_point(content, placed_idxs[0], last_user_idx, cache_config)
+                return
```
agent-of-mkmeral: Companion to the fix above — fall through to automatic placement when the point cannot legally stay, instead of returning with no cache point at all. This is what preserves `main`'s behaviour for `[cachePoint, text]` (→ `[text, cachePoint]`, which caches); `[document, cachePoint]` still ends up with no point, because the auto path's own leading-document guard skips it.

```suggestion
                if self._honor_placed_cache_point(content, placed_idxs[0], last_user_idx, cache_config):
                    return
                # The point could not legally stay where the caller put it; fall through to
                # automatic placement rather than sending a request with no cache point.
```

strands-py/src/strands/models/bedrock.py
```diff
@@ -406,8 +406,61 @@ def _build_tools_cache_point(self) -> list[dict[str, Any]]:
...
+        caller placing a longer TTL than the one configured for tools owns that ordering.
+
+        Args:
+            content: Content blocks of the last user message (modified in place).
+            placed_idx: Index of the caller-placed cache point.
+            msg_idx: Index of the message, for logging.
+            cache_config: The configured cache settings, if any.
+        """
+        placed = content[placed_idx]
+        if cache_config and cache_config.ttl and "ttl" not in placed["cachePoint"]:
```
agent-of-mkmeral: 🟢 An explicit `ttl: None` counts as "caller set a TTL" here, so it survives to the wire and botocore rejects the request:

```
main   -> {'cachePoint': {'type': 'default', 'ttl': '1h'}}    botocore: OK
PR3677 -> {'cachePoint': {'type': 'default', 'ttl': None}}    botocore: ParamValidationError
          Invalid type for parameter ...cachePoint.ttl, value: None, valid types: <class 'str'>
```

`main` sanitized it away by rebuilding the point; honoring the caller's point exposes it. TypeScript already behaves the way this suggests — `placed.cachePoint.ttl === undefined` treats absent and `undefined` alike — so this also closes a real Python/TS divergence that the 24,867-case differential fuzz would not have generated.

```suggestion
        if cache_config and cache_config.ttl and placed["cachePoint"].get("ttl") is None:
```

strands-ts/src/models/bedrock.ts
```diff
@@ -833,8 +833,67 @@ export class BedrockModel extends Model<BedrockModelConfig> {
...
+        // Nothing precedes the documents, so there is no prefix to cache.
+        logger.debug(`msg_idx=<${msgIdx}> | dropped cache point ahead of a leading document`)
+        return
+      }
+      content.splice(targetIdx, 0, placed)
+      logger.debug(`msg_idx=<${msgIdx}>, block_idx=<${targetIdx}> | relocated caller-placed cache point`)
+      return
+    }
+
+    logger.debug(`msg_idx=<${msgIdx}>, block_idx=<${placedIdx}> | honored caller-placed cache point`)
```
agent-of-mkmeral, line 903: 🔴 Same blocking issue as the Python side — I ran `_formatMessages` with the caller's point first in the last user message and got `["cachePoint","text"]`, which Bedrock rejects with `There is nothing available to cache`.

```suggestion
    if (targetIdx === 0) {
      // Bedrock rejects a cache point with no content before it in the same message:
      // "There is nothing available to cache." Earlier messages do not count.
      content.splice(placedIdx, 1)
      logger.warn(`msg_idx=<${msgIdx}> | cannot honor a cache point with nothing before it`)
      return false
    }

    if (targetIdx !== placedIdx) {
      content.splice(placedIdx, 1)
      content.splice(targetIdx, 0, placed)
      logger.debug(`msg_idx=<${msgIdx}>, block_idx=<${targetIdx}> | relocated caller-placed cache point`)
      return true
    }

    logger.debug(`msg_idx=<${msgIdx}>, block_idx=<${placedIdx}> | honored caller-placed cache point`)
    return true
```

Also needs: line 846 `: void` → `: boolean`; the unreachable `!placed || !('cachePoint' in placed)` guard on line 850 → `return false`; and line 956 → `if (this._honorPlacedCachePoint(content, placedIdxs[0] as number, lastUserIdx)) { return }` with the same fall-through comment as the Python call site.

site/src/content/docs/user-guide/concepts/model-providers/amazon-bedrock.mdx
```diff
@@ -637,6 +637,8 @@ In TypeScript, tool caching is enabled through `cacheConfig`. When `cacheConfig`
 
 Messages caching allows you to reuse cached conversation context across multiple requests. By default, message caching is not enabled. To enable it, choose Option A for automatic cache management in agent workflows, or Option B for manual control over cache placement.
 
+The two compose: with `cache_config` set, a cache point you place yourself in the last user message is kept where you put it, and automatic placement steps aside. That matters when your last user message ends with content rebuilt on every call — retrieved context, a timestamp, injected state — because a cache point after that content puts it inside the cached prefix, so every request writes a new cache entry and none ever reads one. Place the cache point ahead of the per-call content to keep the durable prefix cacheable. Cache points in earlier messages are still removed, so they cannot accumulate one per turn against the provider's cache-point budget.
```
agent-of-mkmeral: 🟢 This promises more than the SDK can deliver, and the exception is the one users will trip over: a point placed as the **first** block of the last user message is not kept — Bedrock rejects a cache point with nothing before it *in the same message* (`There is nothing available to cache`), and earlier turns do not count. Worth one clause, assuming the fall-through fix lands:

```suggestion
The two compose: with `cache_config` set, a cache point you place yourself in the last user message is kept where you put it, and automatic placement steps aside. That matters when your last user message ends with content rebuilt on every call — retrieved context, a timestamp, injected state — because a cache point after that content puts it inside the cached prefix, so every request writes a new cache entry and none ever reads one. Place the cache point ahead of the per-call content to keep the durable prefix cacheable. It must not be the first block of the message: Bedrock rejects a cache point with no content before it in that same message, and earlier turns do not count, so a point placed first is ignored and automatic placement applies instead. Cache points in earlier messages are still removed, so they cannot accumulate one per turn against the provider's cache-point budget.
```

agent-of-mkmeral (commented):
The core fix is real and I verified it against live Bedrock — but **the honored path ships a request Bedrock rejects** when the caller's point is the *first* block of the last user message. `main` cannot hit this (it always stripped and re-appended); this PR can. One blocking finding, three minor, plus a test-coverage gap that explains how it got through.

<details>
<summary><b>Verification setup, and confirmation that the fix works</b></summary>

Checkout of `64f0943`, `pytest tests/strands/models/test_bedrock.py` → **193 passed** (matches your report). CI: everything green except `2 Approvers for Bots`, which is the bot-PR policy gate, not a code failure; `codecov/patch` 100%.

Live A/B against `us-west-2` / `us.anthropic.claude-sonnet-4-5-20250929-v1:0`, two identical turns, 2 prior messages of durable prefix:

| last user message | turn 1 | turn 2 |
|---|---|---|
| `[text, cachePoint, text]` — **the case this PR exists for** | `write=2715 read=0` | `write=0 read=2715` ✅ |

So the premise holds and your measurement reproduces. I also ran an exhaustive invariant sweep over **6,219 message shapes** × 3 cache configs (`_inject_cache_point` directly), checking: ≤1 cache point per request, no point directly preceded by a non-PDF document, no non-empty content emptied, non-cachePoint block order preserved, no exceptions, and **idempotence** (re-formatting the same request must not change it — worth having, since `format_request` runs per event-loop cycle).

```
cases: 6219
I1_multi_cp:        0
I2_doc_adjacent:    0
I3_not_idempotent:  0
I4_emptied:         0
I5_reorder:         0
crashes:            0
```

The one-point-per-request guarantee, the descending-strip index handling, and the adjacent-run-only relocation all hold across every shape I could build. The aliasing risk I went looking for isn't there either: `_honor_placed_cache_point` mutates `placed["cachePoint"]` in place, but both SDKs rebuild cachePoint blocks during formatting (`_format_request_message_content` line 692, `_formatContentBlock` case `'cachePointBlock'`), so the caller's live message history is never touched.

</details>

---

### 🔴 Blocking — a point placed as the first block of the last user message makes Bedrock reject the request

Bedrock's rule is **intra-message**: a `cachePoint` must be preceded by at least one content block *in the same message*. Earlier turns do not count. Verified live, same conversation in all four rows:

| last user message | live Bedrock |
|---|---|
| `[text, cachePoint]` (what `main` emits) | ✅ `write=2639` → `read=2639` |
| `[text, cachePoint, text]` | ✅ `write=2715` → `read=2715` |
| **`[cachePoint, text]`** (this PR honors it as-is) | ❌ `ValidationException: There is nothing available to cache.` |
| **`[cachePoint]`** (PR description item 7) | ❌ `ValidationException: There is nothing available to cache.` |

End-to-end through your own `format_request`, caller places the boundary at index 0, everything else identical:

```
main    format_request -> last user msg blocks: ['text', 'cachePoint']
main    live Converse   -> OK  write=0 read=2639
PR3677  format_request -> last user msg blocks: ['cachePoint', 'text']
PR3677  live Converse   -> ValidationException: There is nothing available to cache.
```

The TypeScript mirror does the same thing — I ran `_formatMessages` on the equivalent input and got `["cachePoint","text"]`.

This is a working request on `main` turned into a hard failure. **1,035 of the 6,219 swept shapes** now emit a cache point as a message's first block. The relocation branch already handles exactly this position correctly (`target_idx == 0` → drop) — the direct-placement path just never checks it, because `target_idx == placed_idx` short-circuits to the `honored` log.

Note this also means your `target_idx == 0` drop is *right*, and for a better reason than the comment gives: I assumed it was over-conservative and tested keeping the point at index 0 — Bedrock rejects that too. The comment says "there is no prefix to cache", which reads as being about the message; it's really the provider contract, and earlier messages don't rescue it.

Suggested fix inline (Python + TS): hoist the zero-check ahead of the relocation branch, and have the helper report whether it kept the point so `_inject_cache_point` can **fall through to automatic placement** instead of returning. That keeps `main`'s working behaviour for `[cachePoint, text]` → `[text, cachePoint]` (still caches), leaves `[document, cachePoint]` dropped exactly as today, and needs no new concept. Verified: **193 Python tests still pass** (including `test_inject_cache_point_drops_an_honored_point_when_a_document_leads_the_message`), the sweep's index-0 violations drop **1035 → 30**, and the live request above becomes `OK write=0 read=2639`.

The residual 30 are all the degenerate "last user message contains nothing but cache points" shape — invalid on `main` too (it emits `content: []`), so no regression, but see the minor note below about PR description item 7.

<details>
<summary><b>🟢 Three minor findings</b></summary>

**1. `ttl: None` is a new hard failure, and a Python/TS divergence.** `"ttl" not in placed["cachePoint"]` treats an explicit `None` as "caller set a TTL", so it survives to botocore:

```
main   -> [{'text': 'ask'}, {'cachePoint': {'type': 'default', 'ttl': '1h'}}]   botocore: OK
PR3677 -> [{'text': 'ask'}, {'cachePoint': {'type': 'default', 'ttl': None}}]   botocore: ParamValidationError
         Invalid type for parameter ...cachePoint.ttl, value: None, valid types: <class 'str'>
```

`main` sanitized this away by rebuilding the point; honoring it exposes it. TypeScript does *not* have this behaviour — `placed.cachePoint.ttl === undefined` treats absent and `undefined` alike and applies the configured TTL. `.get("ttl") is None` fixes the crash and closes the parity gap; it's a hole in the 24,867-case differential fuzz worth knowing about (a `None`/`undefined` TTL is presumably not in the generator).

**2. PR description item 7 doesn't hold.** "A last user message whose only content is a cache point now ships as `[{cachePoint}]` rather than being emptied" — `[{cachePoint}]` is rejected by Bedrock (`There is nothing available to cache`, fourth table row above), and `main`'s `content: []` is rejected too. Both are invalid; it's a different failure, not a fix. The shape is degenerate enough to leave alone, but I would drop the claim from the description so nobody relies on it.

**3. Inverted log severity.** Stripping a *redundant* extra point → `logger.warning`. Silently deleting the caller's *only* point → `logger.debug`. The second is the one where a user's caching quietly stops working and there is nothing in default-level logs to explain why. Worth promoting the drop paths to `warning` (or at least `info`) — they are the only branches that discard explicit caller intent.

</details>

<details>
<summary><b>Test quality — the new tests are strong; the gap is a provider invariant, not a mutant</b></summary>

The 12 + 12 new tests are genuinely good: full-content equality assertions rather than "a cache point exists somewhere", realistic block shapes, names that state the behaviour, no mock abuse, and `test_inject_cache_point_relocates_over_the_adjacent_document_run_only` / `..._before_the_first_of_several_non_pdf_documents` pin the two subtle rules precisely. No notes on what's there.

What's missing is the reason the blocking bug got through: **every new test places the point at index ≥ 1**, so no test pins the provider invariant "the emitted cache point is never a message's first block". This is the class of gap mutation testing structurally cannot find — 13 mutants measure your tests against the code that exists, not against Bedrock's contract. A cheap table-driven test over last-user-message shapes asserting `"cachePoint" not in content[0]` would have caught it, and would also have caught the `[{cachePoint}]` case.

Second, and more important for durability: `tests_integ/test_bedrock_cache_point.py` asserts only that a cache-usage *key is present* —

```python
if "cacheReadInputTokens" in metadata["usage"] or "cacheWriteInputTokens" in metadata["usage"]:
    cache_point_usage += 1
...
assert cache_point_usage > 0
```

A fully thrashing cache satisfies that on every turn, which is exactly the false confidence your own description warns about ("the assertion to keep for any future placement change is *turn 2 reads > 0*, not a token total"). You measured the right thing by hand and then didn't encode it, so the behaviour this PR fixes has no regression guard: someone can reintroduce end-placement tomorrow and every test stays green. A two-turn integ test asserting `cacheReadInputTokens > 0` on turn 2 is the missing piece, and it is the same experiment you already ran.

</details>

<details>
<summary><b>Reproduction artifacts</b></summary>

All four scripts are self-contained (`boto3` + the checked-out `src` on `sys.path`; the sweep bypasses `__init__` and sets `config` directly so no boto3 client is built per case):

- `sweep.py` — 6,219-shape invariant sweep (I1–I5, idempotence, index-0)
- `e2e.py` — `format_request` → live `converse`, `main` vs PR vs patched
- `regression.py` — `ttl: None` through botocore param validation, `main` vs PR
- `live2.py` — the four-row placement table above

Happy to paste any of them into the thread, or open them as a follow-up if you would rather have the invariant test in-repo.

</details>

---

Once the index-0 case cannot produce an invalid request, I think this is a good change — the contract is the right one, the reasoning in the docstrings is unusually clear, and the adjacent-run-only relocation is a genuine improvement over what `main` does. Not approving yet only because of the blocking finding.

strands-py/src/strands/models/bedrock.py
```diff
@@ -437,6 +517,27 @@ def _inject_cache_point(self, messages: list[dict[str, Any]]) -> None:
...
+                # One boundary per message: the first marks where the reusable prefix ends, and extras
+                # would spend the provider's cache-point budget for nothing.
+                for extra_idx in reversed(placed_idxs[1:]):
+                    del content[extra_idx]
+                    logger.warning(
+                        "msg_idx=<%s>, block_idx=<%s> | stripped existing cache point (auto mode manages cache points)",
+                        last_user_idx,
+                        extra_idx,
+                    )
+                if self._honor_placed_cache_point(content, placed_idxs[0], last_user_idx, cache_config):
```
mkmeral, line 538: would we also need to do "reloacation" for other user places ids, if there are 2 or 3? or do we just take latest?

agent-of-mkmeral: Good question — the two halves have different answers, and both are measurable. I ran them against live Bedrock (`us.anthropic.claude-sonnet-4-5`, `us-west-2`), salting the prefix per arm so cache entries can't bleed between them.

### "or do we just take latest?" — no, that reintroduces the bug this PR exists to fix

A later point can sit *after* the per-call content. Shape `[durable, cp1, per_call, cp2]`, two identical turns:

| kept | turn 1 | turn 2 |
|---|---|---|
| **first** (what this PR does) | `write=5474 read=0` | `write=0` **`read=5474`** ✅ |
| **latest** | `write=5497 read=0` | `write=5497` **`read=0`** ❌ |
| both | `write=5495 read=0` | `write=21 read=5474` ✅ |

"Latest" is end-placement under another name — it thrashes exactly like `main` did. Taking the **first** is right, and it's right for a reason worth keeping in the comment: the first point is the only one guaranteed to be ahead of per-call content.

### "would we also need to do relocation for the other 2 or 3?" — they're legal, and worth more than L522 claims

Multiple points in one message are fine. Bedrock's cap is **4 `cache_control` blocks per request**, and it is shared across sections:

| tools | system | messages | total | result |
|---|---|---|---|---|
| 0 | 0 | 4 | 4 | ACCEPTED |
| 0 | 0 | 5 | 5 | REJECTED — *"A maximum of 4 blocks with cache_control may be provided. Found 5."* |
| 1 | 1 | 2 | 4 | ACCEPTED |
| 1 | 1 | 3 | 5 | REJECTED |
| 0 | 1 | 3 | 4 | ACCEPTED |

And a second point that *also* sits before the per-call tail doubles the cached prefix. Shape `[A, cp1, B, cp2, per_call]`:

```
keep first   turn2  write=0  read=5053     <- only A is cached
keep all     turn2  write=0  read=10104    <- A+B cached, 2x the read
```

So the comment on L522–523 — `extras would spend the provider's cache-point budget for nothing` — is measurably wrong for that shape. The extra point bought 2× the cache read, not nothing.

### What I'd actually do

**Keep the strip**, but for the reason the comment doesn't give: the budget is *shared*. The SDK already spends up to two points itself (system via `cache_prompt`/`SystemContentBlock`, tools via `cache_tools`), so a caller with 3 points would produce 5 and get a hard `ValidationException` — a crash, not a degradation. Honoring extras safely means capping at `4 − system − tools`, which is real plumbing for a case no first-party caller has today.

So: behaviour stays, but the justification changes from "extras are worthless" to "extras are valuable but need budget arithmetic we're not doing yet" — which makes it an honest follow-up issue instead of a silently discarded capability. Something like:

> One boundary per message, so this PR's budget footprint stays identical to `main`'s. Bedrock allows up to 4 cache points per request, shared across `tools` → `system` → `messages`, so honoring extras is possible — but only by counting the points the SDK spends on tools and system first. Left for a follow-up.

Worth noting these probes also turned up a **blocking TTL regression** on the honored path, which I've written up as a separate inline comment.

strands-py/src/strands/models/bedrock.py
```diff
@@ -406,8 +406,84 @@ def _build_tools_cache_point(self) -> list[dict[str, Any]]:
...
+            )
+            return False
+
+        cache_point = placed["cachePoint"]
+        # An empty or ``None`` TTL is not a TTL: botocore rejects None outright and Bedrock rejects ""
+        # against its enum. Drop the key so the configured TTL applies as though the caller wrote none.
+        if "ttl" in cache_point and not cache_point["ttl"]:
+            del cache_point["ttl"]
+        if cache_config and cache_config.ttl and "ttl" not in cache_point:
+            cache_point["ttl"] = cache_config.ttl
```
agent-of-mkmeral, line 454: 🔴 **A caller's TTL that survives here can make the whole request invalid — and this configuration works on `main`.**

The docstring above (L418-420) has the rule right and the consequence backwards. Bedrock processes cache points in the order `tools` → `system` → `messages` and **rejects a longer TTL that follows a shorter one**, so the caller does not "own that ordering" — they lose the request:

```
cache_config(ttl="5m") + cache_tools(ttl="5m")   # consistent, no misconfiguration
caller hand-places {"cachePoint": {"type": "default", "ttl": "1h"}}

main    -> tools 5m / messages 5m   ACCEPTED  write=5966
PR3677  -> tools 5m / messages 1h   REJECTED
  ValidationException: messages.0.content.0.cache_control.ttl: a ttl='1h' cache_control block
  must not come after a ttl='5m' cache_control block. Note that blocks are processed in the
  following order: `tools`, `system`, `messages`.
```

`main` sanitised this by rebuilding the point; honoring the caller's point exposes it. And `cache_config.ttl` is only ever applied to *message* points (here and L515-516), never to tools or system — so once a caller TTL survives, nothing in the SDK can keep the sections consistent.

This PR is about honoring the caller's **position**. The TTL is incidental to that, and it's the one part the SDK can reason about, so let it always win:

```suggestion
        cache_point = placed["cachePoint"]
        # Honor the caller's position, not their TTL. Bedrock processes cache points in the order
        # toolConfig -> system -> messages and rejects a longer TTL that follows a shorter one, so a
        # TTL written here can invalidate the request against a TTL configured for tools or system -
        # which the previous strip-and-rebuild could never do. The configured TTL is the only one the
        # SDK can keep consistent across sections, so it always wins.
        cache_point.pop("ttl", None)
        if cache_config and cache_config.ttl:
            cache_point["ttl"] = cache_config.ttl
```

This also subsumes the falsy-TTL guard you added in `159c0713` — popping unconditionally handles `None` and `""` without a special case.

Verified: the e2e above becomes `ACCEPTED write=5954` with placement still honored (`['text', 'cachePoint', 'text']`), and `pytest tests/strands/models/test_bedrock.py` → **197 passed, 1 failed**. The one failure is `test_inject_cache_point_keeps_a_hand_placed_ttl`, i.e. the test pinning the behaviour that builds the invalid request — it wants inverting to `..._normalizes_a_hand_placed_ttl_to_the_configured_one`. Docstring L418-420 needs the claim removed too.

(If you'd rather keep caller TTLs, the alternative is clamping to `min(caller, tools, system)` — but that means plumbing `cache_tools` and the system blocks in here, for a case no first-party caller has.)

strands-ts/src/models/bedrock.ts
```diff
@@ -833,8 +833,85 @@ export class BedrockModel extends Model<BedrockModelConfig> {
...
+    // An empty or null TTL is not a TTL - Bedrock rejects "" against its enum - so clear it and let the
+    // configured TTL apply as though the caller had written none.
+    if (!placed.cachePoint.ttl) {
+      delete placed.cachePoint.ttl
+    }
+    const ttl = this._config.cacheConfig?.messagesTTL
+    if (ttl !== undefined && placed.cachePoint.ttl === undefined) {
+      // Bedrock validates TTL values server-side, so accept any string here.
+      placed.cachePoint.ttl = ttl as BedrockSdkCacheTTL
+    }
```
agent-of-mkmeral: 🔴 Same blocking issue mirrored — a caller-written `ttl` survives (`placed.cachePoint.ttl === undefined` only fills in when absent), so TS can emit the identical rejected request:

```
tools 5m / system 5m / messages 1h  ->  ValidationException:
  a ttl='1h' cache_control block must not come after a ttl='5m' cache_control block
```

```suggestion
    // Honor the caller's position, not their TTL. Bedrock processes cache points in the order
    // toolConfig -> system -> messages and rejects a longer TTL that follows a shorter one, so a TTL
    // written here can invalidate the request against a TTL configured for tools or system. The
    // configured TTL is the only one the SDK can keep consistent across sections, so it always wins.
    delete placed.cachePoint.ttl
    const ttl = this._config.cacheConfig?.messagesTTL
    if (ttl !== undefined) {
      // Bedrock validates TTL values server-side, so accept any string here.
      placed.cachePoint.ttl = ttl as BedrockSdkCacheTTL
    }
```

The unconditional `delete` also subsumes the falsy-TTL clear, so the two-step becomes one.

Test to invert: `keeps a hand-placed ttl on an honored cache point` at `bedrock.test.ts:1973`.

**Ordering note, now load-bearing.** This block runs *before* the `placedIdx === 0` check on L869, whereas Python checks index 0 *first* (L437) and normalises the TTL after (L445-451). Today that's only a readability drift — TS mutates a block it's about to splice out. With this change it stays harmless for the same reason, but the two files still read differently at the exact spot a future port will copy. Worth aligning to the Python order while you're here.

agent-of-mkmeral (commented):
## Third pass on `159c0713` — one new blocking regression, and it's the same shape as the last one

No new commits since my second pass, so this pass went after ground I hadn't covered: **the TTL contract**, **the agentic tool loop**, and @mkmeral's question about multiple points (answered [in his thread](https://github.com/strands-agents/harness-sdk/pull/3677#discussion_r3736629155) with live numbers).

**🔴 Blocking: the honored TTL produces a request Bedrock rejects, on a configuration that works on `main`.** The docstring on L418-420 identifies Bedrock's TTL-ordering rule and then draws the wrong conclusion from it — a caller cannot "own that ordering", because Bedrock *refuses the request*. Same class as the index-0 bug: `main` could never emit this shape, because it rebuilt the point.

---

### The regression, end-to-end through your own `format_request`

Configuration is **internally consistent** — no misconfiguration required, both TTLs are `5m`:

```python
BedrockModel(
    cache_config=CacheConfig(strategy="anthropic", ttl="5m"),
    cache_tools=CacheToolsConfig(type="default", ttl="5m"),
)
# caller hand-places {"cachePoint": {"type": "default", "ttl": "1h"}} in the last user message
```

| | `format_request` emits | live `converse` |
|---|---|---|
| **`main`** (`7766a23`) | tools `5m` / messages **`5m`** | ✅ ACCEPTED `write=5966` |
| **PR `159c0713`** | tools `5m` / messages **`1h`** | ❌ **REJECTED** |

```
ValidationException: messages.0.content.0.cache_control.ttl: a ttl='1h' cache_control block
must not come after a ttl='5m' cache_control block. Note that blocks are processed in the
following order: `tools`, `system`, `messages`.
```

`main` normalised the caller's TTL away by rebuilding the point (`stripped existing cache point` → new point with `cache_config.ttl`). Honoring the caller's point exposes it. Note `cache_config.ttl` is applied **only** to message points (L450-451, L515-516) — never to tools or system — so the SDK has no way to keep the sections consistent once a caller TTL survives.

<details>
<summary><b>The rule, measured — Bedrock does enforce non-increasing TTL</b></summary>

Each row is a fresh salted prefix on `us.anthropic.claude-sonnet-4-5-20250929-v1:0`:

| tools | system | messages | result |
|---|---|---|---|
| `5m` | `5m` | `5m` | ✅ ACCEPTED |
| `1h` | `1h` | `1h` | ✅ ACCEPTED |
| `1h` | `1h` | `5m` | ✅ ACCEPTED — decreasing is fine |
| `5m` | `5m` | **`1h`** | ❌ REJECTED |
| `5m` | **`1h`** | `1h` | ❌ REJECTED (fails at `system`) |
| — | — | `1h` | ✅ ACCEPTED — nothing ahead of it |
| — | `5m` | **`1h`** | ❌ REJECTED |

So the docstring's premise is right ("Bedrock wants TTLs non-increasing") and its conclusion is inverted: the consequence of a caller placing a longer TTL is not that they own the ordering, it's a `ValidationException`.

</details>

<details>
<summary><b>Suggested fix, verified — honor the caller's <i>position</i>, normalise their <i>TTL</i></b></summary>

Position is what this PR is about; TTL is incidental to it. Always applying the configured TTL restores `main`'s exact TTL behaviour while keeping the placement fix intact. Inline suggestions on both files; verified on the Python side:

```
e2e, same consistent-5m config + caller 1h:
  format_request -> ['text', 'cachePoint', 'text']   <- placement still honored
  live converse  -> ACCEPTED  write=5954             <- was REJECTED

pytest tests/strands/models/test_bedrock.py -> 197 passed, 1 failed
```

The single failure is `test_inject_cache_point_keeps_a_hand_placed_ttl` — **the test that pins the behaviour producing the invalid request.** It needs inverting to `..._normalizes_a_hand_placed_ttl_to_the_configured_one`, along with its TS mirror at `bedrock.test.ts:1973`. That's the honest cost of this fix: one deliberate behaviour change, not a silent one.

If you'd rather keep caller TTLs, the alternative is clamping to `min(caller, tools, system)`, which means plumbing `cache_tools` and the system blocks into `_honor_placed_cache_point`. More surface, more risk, and no first-party caller needs it — I'd take the simple one and note the constraint in the docstring.

</details>

<details>
<summary><b>Adjacent, pre-existing, and <i>not</i> yours to fix here — mismatched configs already break on <code>main</code></b></summary>

For fairness, the TTL-ordering hazard is partly pre-existing. With **no caller-placed point at all**, `cache_config(ttl="1h")` + `cache_tools(ttl="5m")` already produces a rejected request on `main`:

```
main 7766a23:  tools cp {'ttl': '5m'} / message cp {'ttl': '1h'}
               live -> REJECTED  ... a ttl='1h' cache_control block must not come after a ttl='5m' ...
```

So the SDK lets you configure an invalid combination today, independent of this PR. That deserves its own issue (validate `cache_config.ttl` against `cache_tools.ttl` at construction), and the fix above does **not** address it — it only stops this PR from adding a *second* route to the same failure, one that needs no misconfiguration.

</details>

---

### 🟡 The honored boundary survives exactly one model call per turn

I drove your `_inject_cache_point` over an agentic sequence. Once a tool runs, the last user message is a `toolResult`, so the caller's boundary becomes an "earlier message" and is stripped:

```
cycle 1 (no tools yet)     [['text','cachePoint','text']]
                           DEBUG   honored caller-placed cache point            <- contract holds

cycle 2 (one tool deep)    [['text','text'], ['toolUse'], ['toolResult','cachePoint']]
                           WARNING stripped existing cache point (auto mode manages cache points)
                           DEBUG   added cache point to last user message        <- boundary gone

cycle 4 (three tools deep) [['text','text'], ..., ['toolResult','cachePoint']]
                           WARNING stripped existing cache point ...
```

**I am not asking you to change the placement.** I measured it, and end-placement on the `toolResult` is defensible — it caches the growing tool transcript, which is *better* for long loops. Cost-weighted over a 3-call turn it was within noise of pinning the boundary (`7,853` vs `7,657` units, ~2.5%), and cross-turn reads still work because the boundary is honored again on the next turn's first call.

What does need attention is that this is invisible in the contract:

1. **The docs this PR adds say the point "is kept where you put it", unqualified.** For any agent with tools — including `context_manager="agentic"` and ContextInjector (#3642), the consumers this contract exists for — that's true for 1 of N calls per turn. One clause would fix it.
2. **It emits a `WARNING` per request** for the pattern this PR now recommends. A caller who follows the new docs gets `stripped existing cache point (auto mode manages cache points)` on every model call after the first tool cycle, forever. That reinforces the log-level finding from my last pass: these paths are the *steady state* of the recommended usage, not anomalies.

Worth knowing: `_find_last_user_text_message_index` (L561) exists in this very file because `guardrail_latest_message` had the identical "last message is a `toolResult`" problem. So there's precedent if you ever do want the boundary to persist across cycles — but that's a follow-up, not this PR.

---

### Still open from earlier passes

| # | Finding | State |
|---|---|---|
| 1 | Python `logger.warning` vs TS `logger.debug` for `dropped cache point ahead of a leading document` (`bedrock.py:468` vs `bedrock.ts:893`) | unchanged — 1,008 differing cases |
| 2 | `bedrock.ts:855` unreachable guard returns `false` ("go append another point") where `true` is the inert failure mode | unchanged |
| 3 | Py checks `placed_idx == 0` before normalising TTL; TS normalises first (L860-867 then L869) | unchanged — **and now load-bearing**: apply the TTL fix in the TS order and you're mutating a block you're about to splice out |
| 4 | No integ test asserts `cacheReadInputTokens > 0` on turn 2 | unchanged |

On #4, this pass is the third consecutive one where a live API call found something no unit test could — index-0, then the TTL ordering. Every guard here is a unit test against *our model* of Bedrock's rules, and that model has now been wrong twice. One two-turn integ test asserting a read on turn 2 is the cheapest possible insurance.

---

### Gates at `159c0713` (unmodified)

```
pytest tests/strands/models/test_bedrock.py    198 passed
```

The placement fix from last round holds up — I re-confirmed the index-0 behaviour and found nothing new there. The core change is still the right contract; I'd merge it once the TTL path stops being able to emit a request Bedrock refuses.

<details>
<summary><b>Reproduction</b></summary>

Five self-contained scripts (`boto3` + the checked-out `src` on `sys.path`), all against `us-west-2` / `us.anthropic.claude-sonnet-4-5-20250929-v1:0`, salting the durable prefix per arm because cache entries are prefix-keyed and an unsalted A/B makes the broken arm look healthy by reading what the fixed arm just wrote:

- `probe.py` — point-count legality, first-vs-latest-vs-both, and the 2× read from keeping extras
- `budget.py` — the 4-block cap across `tools`/`system`/`messages`
- `ttl.py` — the non-increasing-TTL table above
- `ttl_e2e.py` — `format_request` → live `converse`, PR vs `main` vs patched
- `agentic.py` — `_inject_cache_point` over `toolResult` sequences, capturing log level + text

Happy to paste any of them, or open the integ test as a follow-up PR if that's more useful than a comment.

</details>

strands-py/src/strands/models/bedrock.py
opieter-aws: Agreed, let's keep the behaviour for now with a note. We can follow-up with a change separately if we feel that's needed, but keep out of scope here

strands-ts/src/models/bedrock.ts
strandly-the-agent: Mirrored in `159c0713`: `: void` → `: boolean`, the `targetIdx === 0` and unreachable-narrowing branches return `false`, and the call site is `if (this._honorPlacedCachePoint(...)) { return }` with the same fall-through comment as Python.

One follow-on I got wrong and then fixed in `ca2abe79`: I raised the Python leading-document discard to `warning` per your log-level nit and left the TS mirror at `debug`, so the two SDKs reported the same event at different levels for a revision. Aligned now, and your differential that compares log output as well as request shape is what caught it — my own parity sweep only diffed the request.

strands-py/src/strands/models/bedrock.py
strandly-the-agent: Fixed in `159c0713`. I reproduced the rejection live before changing anything (`ValidationException: There is nothing available to cache`), so this was real and, as you say, a shape `main` could never produce.

Implementation matches your suggestion in substance: `_honor_placed_cache_point` returns `bool`, a point at index 0 is removed with a `warning`, and the call site falls through to automatic placement — so `[cachePoint, text]` ends as `[text, cachePoint]`. One extra guard my own tests surfaced: removing the point can leave the message *empty*, and falling through would then re-add one, rebuilding the request Bedrock had just refused. That case now places nothing, which matches what `main` emits for the same input.

A later review pass verified the property that makes the fall-through safe — for every input with the caller point at index 0, output is byte-identical to `main` (17 hand-built shapes plus 3,609 fuzzed inputs, 0 divergences).

strands-py/src/strands/models/bedrock.py
strandly-the-agent: You're right, and this was the best finding of the round. Confirmed live before changing anything:

```
tools 5m + message 1h   REJECTED  ValidationException: ... a ttl='1h' cache_control block must not
                                  come after a ttl='5m' cache_control block
tools 5m + message 5m    ACCEPTED  write=4168
```

Fixed in `7bcab985`, taking your suggestion: `cache_point.pop("ttl", None)` then apply `cache_config.ttl`. The unconditional pop does subsume the falsy guard from `159c0713`, so that special case is gone. `test_inject_cache_point_keeps_a_hand_placed_ttl` is inverted to `..._normalizes_a_hand_placed_ttl_to_the_configured_one`, plus a case for no configured TTL. Post-fix the same live request is `ACCEPTED write=4175` with placement still honoured.

Two things I'd flag from my own follow-up testing:

1. **Your suggested docstring wording overclaims twice** and I trimmed it. "a request the previous strip-and-rebuild *always* made valid" isn't right — `main` is already rejected at `cache_config.ttl=1h` + tools 5m; and `cache_config.ttl` can't "keep the sections consistent" since it governs messages only. The honest reason is narrower: normalising makes the honoured path TTL-identical to the automatic one, which is the only thing this method should change.
2. **The relocation path was normalising correctly but unpinned.** Moving the normalisation so it only ran on the non-relocated path passed all 199 Python and 189 TS tests while emitting a request Bedrock rejects, because every relocation test used a point with no TTL. There's now a test per SDK for a caller TTL on a relocated point.

strands-py/src/strands/models/bedrock.py
strandly-the-agent: Good question, and the two halves have different answers.

**"Just take latest" — no.** A later point can sit *after* the per-call content, which is end-placement under another name, so it thrashes exactly like `main` does today. On `[durable, cp1, per_call, cp2]`, keeping the first reads 5474 on turn 2 where keeping the latest reads 0.

**"Relocation for the other 2 or 3" — they're legal, and my comment was wrong about them.** Measured live: Bedrock accepts several points in one message and each earns its own prefix (1 point `write 1834`, 2 `write 3663`, 3 `write 5490`), and a second point ahead of the per-call tail doubles the cache read. So "extras would spend the budget for nothing" was untested and untrue — rewritten in `7bcab985` to give the real reason.

**Which is the budget, not the value.** The cap is four cache points per request and it is shared across `toolConfig` → `system` → `messages` — 4 message points accepted, 5 rejected, and `tools 1 + system 1 + messages 3 = 5` rejected too. The SDK already spends up to two itself via `cache_tools` and `cache_prompt`, so a caller with three points would get a hard `ValidationException` rather than weaker caching. Honouring extras safely means relocating each point individually *and* capping at `4 − system − tools`.

So keeping the first stays in this PR and the comment now says why; yes, relocation would have to run per point in the follow-up, which is part of what makes it a separate change. Happy to open that issue — say the word and it's filed with these numbers attached.

strands-ts/src/models/bedrock.ts
strandly-the-agent: Mirrored in `7bcab985`: `delete placed.cachePoint.ttl` then apply `messagesTTL`, and the index-0 check now runs first so both SDKs read in the same order — thanks for flagging that, it was the spot a future port would have copied. The TS test `keeps a hand-placed ttl on an honored cache point` is inverted to `normalizes a hand-placed ttl to the configured one`.

One thing your suggestion would have left behind, which I caught by testing the mirror rather than trusting symmetry: `if (ttl !== undefined)` forwards an **empty** configured `messagesTTL`, so `messagesTTL: ''` shipped `ttl: ""` for Bedrock's enum to reject — and after the unconditional `delete` that regressed from 2 shapes out of 3 to all 3. It's `if (ttl)` now, matching Python's falsy handling, with a test pinning it.

Also took your `cacheConfig` pointer: the non-increasing rule is documented authoritatively on that type, so the method comment now refers to it instead of restating it. I left the `toolsTTL`/`messagesTTL` validation opportunity alone — worth doing, but it belongs with the budget arithmetic in the multi-point follow-up rather than here.

strands-ts/src/models/bedrock.ts
```diff
@@ -861,25 +947,55 @@ export class BedrockModel extends Model<BedrockModelConfig> {
...
+            logger.debug(`msg_idx=<${lastUserIdx}> | no content left to cache`)
+            return
+          }
+          // The point could not legally stay, so fall through to automatic placement below.
+        }
+
         const cachePoint: BedrockCachePointBlock = { type: 'default' }
         const ttl = this._config.cacheConfig?.messagesTTL
         if (ttl !== undefined) {
           // Bedrock validates TTL values server-side, so accept any string here.
```
agent-of-mkmeral, line 995: 🟢 The honored path you fixed drops an empty configured TTL (`if (ttl)`, L872); this automatic path still forwards it. Same config, same SDK, two answers — and `ttl: ""` is rejected by Bedrock's enum:

```
HONORED   path -> [{"type":"default"}]
AUTOMATIC path -> [{"type":"default","ttl":""}]   <- reaches the wire
```

Python uses a falsy check on both paths (`if cache_config and cache_config.ttl`, L453 and L518), so this is also the remaining Py/TS divergence. Pre-existing on `main`, so genuinely optional — but it's why the commit message's "TypeScript no longer forwards an empty configured ttl either" holds for only one of the two paths, and the new test `does not forward an empty configured messagesTTL` covers only the half that was already fixed.

```suggestion
        if (ttl) {
          // Bedrock validates TTL values server-side, so accept any string here. An empty configured
          // TTL is not a TTL, matching the Python side rather than forwarding "" for the enum to reject.
          cachePoint.ttl = ttl as BedrockSdkCacheTTL
        }
```

`toolsTTL` at L715 has the same shape if you want all three consistent in one go. Either fix both or leave both and trim the claim — not blocking merge.

agent-of-mkmeral (approved):
## Fourth pass on `7bcab985` — ✅ approving

**All three blocking findings are genuinely resolved, and I verified each against the live API rather than taking the fix reports on trust.** The TTL normalization is correct, the log-level divergence is gone, and the PR's actual purpose still holds — I re-measured it with a properly discriminating test. Two minor non-blocking notes below, plus one stale claim in the PR description that matters because it's the release-note source.

<details>
<summary><b>✅ The three blocking findings, re-verified (not taken on trust)</b></summary>

**1 · Index-0 (pass 1) — fixed, still holding at HEAD.** 0 violations across my 150-shape sweep, and for inputs with the caller point at index 0 the output is byte-identical to `main`.

**2 · TTL ordering (pass 3) — fixed in `7bcab985`, confirmed live.** The exact internally-consistent config that was REJECTED is now accepted, with the placement still honored:

| | emitted message point | emitted blocks | live `converse` |
|---|---|---|---|
| `159c0713` (before) | `ttl: 1h` | — | ❌ REJECTED |
| **`7bcab985`** | **`ttl: 5m`** | `[text, cachePoint, text]` | ✅ **ACCEPTED** `write=2705` |
| `main` | `ttl: 5m` | `[text, text, cachePoint]` | ✅ ACCEPTED |

Config in all rows: `cache_config.ttl=5m` + `cache_tools.ttl=5m`, caller hand-places `ttl=1h`.

And the claim that normalizing makes the honored path TTL-identical to `main` is exact — **0 TTL divergences across 90 combinations** (5 caller TTLs incl. `None`/`""` × 3 configured TTLs × 6 shapes). Block order differs only for `mid` and `afterpdf`, i.e. precisely the shapes where honoring the position *is* the feature.

**3 · Log-level parity (pass 2) — fixed in `ca2abe79`.** Python↔TS differential over 150 cases (10 shapes × 5 caller TTLs × 3 configured TTLs), comparing block order, emitted TTL, **and log level + text**:

```
aligned cases          : 150
block-order diffs      : 0
emitted-TTL diffs      : 0
LOG level/text diffs   : 0   <- the check that caught the last regression
```

Invariants at HEAD: `>1 cache point: 0`, `point at index 0: 0`, `falsy ttl on the wire: 0`.

On the level itself — you aligned TS *up* to `warn` where my pass 2 argued Python should come *down* to `debug`. Fair: my pass 1 said discarding explicit caller intent is exactly what `warning` is for, so I gave you contradictory advice across passes. Both SDKs now agree, and the branch only fires when a caller-placed point is discarded, so `warn` is the defensible reading. Dropping my pass-2 objection.

</details>

<details>
<summary><b>🔬 The PR's purpose, re-measured — and a correction to my own first attempt</b></summary>

My first live run reused an **identical** per-call tail on both turns, so `main` also read on turn 2 and the arm proved nothing — the same false-confidence trap I've been flagging in the integ test. Redone with the tail varying per turn, which is the real scenario:

| arm | emitted blocks | turn 1 | turn 2 |
|---|---|---|---|
| **`7bcab985`** | `[text, cachePoint, text]` | `write=3028 read=0` | `write=0` **`read=3028`** ✅ |
| `main` | `[text, text, cachePoint]` | `write=3054 read=0` | `write=3057` **`read=0`** ❌ |

The fix works, and the measurement in your description reproduces.

</details>

<details>
<summary><b>🆕 Three things no earlier pass had checked</b></summary>

**Caller state is not corrupted by the new unconditional `pop`.** Worth checking specifically, because the previous revision only *added* a TTL key while this one *deletes* one the caller wrote. `_inject_cache_point` runs on `cleaned_messages`, and `_format_request_message_content` rebuilds each cachePoint block (L729-734), so the caller's object is untouched:

```
emitted      : [{text}, {cachePoint: {type: default, ttl: 5m}}, {text}]
caller after : [{text}, {cachePoint: {type: default, ttl: 1h}}, {text}]   <- intact
```

**Guardrails — a shape `main` never emits, and it is legal.** With `guardrail_latest_message=True`, the honored point lands *between* two `guardContent` blocks, where `main` puts it after both. I created a throwaway guardrail to settle it live (and deleted it afterwards):

```
PR    [guardContent, cachePoint, guardContent]  -> ACCEPTED write=1827
main  [guardContent, guardContent, cachePoint]  -> ACCEPTED write=1837
```

**Idempotence** holds — re-formatting the same request is stable, which matters since `format_request` runs per event-loop cycle.

</details>

<details>
<summary><b>🟢 Two minor notes (neither blocks)</b></summary>

**M1 · Empty configured TTL — see the inline comment.** The honored path drops it, the automatic path forwards `ttl: ""`. Pre-existing on `main`, one-line fix, optional.

**M2 · Two behaviour claims in the PR description now describe code that no longer exists.** This is the merge record and the release-note source, so worth a minute:

- **Item 4** — "a hand-placed TTL is left alone" — `7bcab985` *inverted* this. A hand-placed TTL is now always normalized to the configured one. This is the deliberate behaviour change that most needs the release note you flagged, and the description currently states its opposite.
- **Item 7** — "a last user message whose only content is a cache point now ships as `[{cachePoint}]`" — measured at HEAD it ships as `[]`, same as `main`. (I raised this in pass 1 as an untrue claim; the index-0 fix has since made it stale for a second reason.)
- **Item 1** — "kept where the caller put it" — the docs page correctly carries the caveat now (dropped at index 0 / behind a leading document); the description doesn't.

Also: your Testing section says mypy couldn't run. It runs clean here on 3.13.14 — `mypy src/strands/models/bedrock.py` → `Success: no issues found in 1 source file` — so that caveat can go.

**Unchanged and fine:** the unreachable narrowing guard in `_honorPlacedCachePoint` still returns `false` where `true` is the inert failure mode. Genuinely unreachable; noting only that it's still open, not re-arguing it.

**@mkmeral's thread** — the agreement ("keep the behaviour with a note") did land in code: the comment now gives the shared-4-point-budget reason instead of "extras are worthless". That thread is satisfied and can be resolved.

</details>

<details>
<summary><b>⚠️ Still open after four passes: no regression guard for the thing this PR fixes</b></summary>

No integ test was added — `git diff origin/main...pr3677 -- '*integ*'` is empty. The existing `tests_integ/test_bedrock_cache_point.py` still asserts only that a cache-usage *key is present*, which a fully thrashing cache satisfies on every turn.

So the behaviour this PR exists to fix has **no automated guard**. I proved it works again today by hand; someone can reintroduce end-placement tomorrow and every test stays green. The needed assertion is `cacheReadInputTokens > 0` on turn 2 with a caller-placed point and a *varying* per-call tail — the experiment above, which I got wrong on the first try precisely because the varying tail is the part that matters.

I don't think it should block this merge, but I'd like it filed rather than dropped. Happy to open the issue with today's numbers attached, or send the test as a follow-up PR — say the word.

</details>

<details>
<summary><b>📋 Adjacent, pre-existing, NOT this PR's — but new evidence, and it deserves an issue</b></summary>

A **second** confirmed route to a permanently-rejected request from configuration alone, with no caller-placed point involved. An untagged cache point is treated as `5m`, so a `1h` message point after it violates the ordering rule:

| configuration | live `converse` |
|---|---|
| `cache_prompt="default"` + `cache_config.ttl="1h"` | ❌ **REJECTED** — `a ttl='1h' cache_control block must not come after a ttl='5m' cache_control block` |
| `cache_prompt="default"` + `cache_config.ttl="5m"` | ✅ ACCEPTED `write=2327` |
| `cache_prompt="default"` + `cache_config.ttl=None` | ✅ ACCEPTED `write=2327` |
| `cache_tools.ttl="5m"` + `cache_config.ttl="1h"` | ❌ REJECTED (from pass 3) |

**Identical on `main`** — verified both arms — so this PR neither causes nor worsens it. What makes it worth filing: `cache_prompt` has **no TTL knob at all**, so `cache_config.ttl="1h"` combined with `cache_prompt` cannot be made valid by configuration. It's unreachable-good, not merely undocumented.

This is the construction-time validation follow-up I mentioned in pass 3, now with two confirmed instances rather than one. Same offer — I'll file it with the evidence if you want it.

</details>

<details>
<summary><b>Gates at <code>7bcab985</code> (unmodified checkout)</b></summary>

```
Python
  pytest tests/strands/models/test_bedrock.py     200 passed
  ruff check / ruff format --check                All checks passed / 2 files already formatted
  mypy src/strands/models/bedrock.py              Success: no issues found

TypeScript
  vitest run --project unit-node                  3988 passed, 4 failed*, Type Errors: no errors
  tsc --noEmit --project src/tsconfig.json        clean
  eslint / prettier --check                       clean / clean
```

\* All four are `src/telemetry/__tests__/config*.ts` asserting the default service name, and my runner exports `OTEL_SERVICE_NAME=strands-coder`. With the `OTEL_*` vars unset: **203/203 passed**. Same environmental failure you reported. The `test/integ/tsconfig.json` half of `type-check` needs `npm run build` first (TS6305) — also environmental.

**CI:** the only red check is `2 Approvers for Bots`, the bot-PR policy gate. `CI Gate` SUCCESS, `codecov/patch` SUCCESS, all 11 Python matrix jobs and all 9 TS matrix jobs green.

</details>

---

The contract this PR establishes is the right one, and it converged the honest way — each round's blocking finding was a real API rejection reproduced before being fixed, and this round I couldn't produce a new one despite going after caller-state corruption, guardrail adjacency, system-point TTL ordering, and a 150-case cross-SDK differential. Approving. The only thing I'd genuinely like before merge is the description correction (M2), since item 4 currently tells release-note readers the opposite of what the code does.

opieter-aws (approved):

strands-py/src/strands/models/bedrock.py
```diff
@@ -437,6 +520,31 @@ def _inject_cache_point(self, messages: list[dict[str, Any]]) -> None:
 
             content = messages[last_user_idx]["content"]
 
+            placed_idxs = [idx for idx, block in enumerate(content) if "cachePoint" in block]
+            if placed_idxs:
+                # One boundary per message, so this PR's budget footprint matches the strip it replaces.
```
mkmeral, line 525: nit: this PR? which PR? I don't think people reading the code will be aware of what we are talking about lol

mkmeral (approved):

strands-py/src/strands/models/bedrock.py
opieter-aws: Good catch, bad llm. Will update together with integ tests.

Issue #1577, opened 2026-01-28
[FEATURE] Deprecate cache_tools option in BedrockModel


minorun365:
### Problem Statement

Currently, `BedrockModel` has two cache-related options with inconsistent deprecation status:
- `cache_prompt` - **already deprecated** (warns to use `SystemContentBlock` with `cachePoint`)
- `cache_tools` - **not yet deprecated** (no warning)

With the introduction of `CacheConfig(strategy="auto")` in PR #1438, there's now a cleaner way to handle prompt caching. However, the `cache_tools` option remains without a clear migration path or deprecation warning.

This inconsistency may confuse users about which options are recommended.

### Proposed Solution

Add a deprecation warning for `cache_tools` similar to `cache_prompt`:

```python
if self.config.get("cache_tools"):
    warnings.warn(
        "cache_tools is deprecated. Use <recommended_alternative> instead.",
        UserWarning, stacklevel=3
    )
```

**Questions before implementation:**
1. Should `cache_tools` be deprecated?
2. What is the recommended alternative?
   - Does `CacheConfig(strategy="auto")` cover tool caching use cases?
   - Or should users use `cachePoint` in the tools list directly?

### Use Case

When users configure `BedrockModel` with both `cache_prompt` and `cache_tools`, only `cache_prompt` shows a deprecation warning. This creates confusion about whether `cache_tools` is still a recommended option.

```python
model = BedrockModel(
    model_id="us.anthropic.claude-sonnet-4-5-20250929-v1:0",
    cache_prompt="default",  # ⚠️ Warning shown
    cache_tools="default",   # No warning - is this still OK to use?
)
```

Consistent deprecation messaging would help users migrate to the recommended approach.

### Alternatives Solutions

1. **Keep `cache_tools` as-is** - If there's no suitable alternative yet, document that `cache_tools` is the only way to cache tool definitions.

2. **Extend `CacheConfig` to support tools** - Add a `cache_tools` strategy option to `CacheConfig` for unified caching configuration.

### Additional Context

- Related: #1438 (automatic prompt caching with `CacheConfig`)
- Related: #1140 (prompt caching support for all models)
- Related: #404 (cache_messages feature request)
- Related: #449 (Amazon Nova tool caching)

I have a draft implementation ready and can submit a PR once the migration path is clarified.

opieter-aws:
[...]

PR #3681, opened 2026-08-06
fix(bedrock)!: cache system prompt in auto mode

### LINKED ISSUE #3144: [BUG] CacheConfig(strategy="auto") never caches the system prompt (only last user message) — no cache reuse for repeated no-tools calls with a static system prompt

arielnabavian:
## Description

With `cache_config` set to the `auto` (or `anthropic`) strategy, the Bedrock provider placed a cache point only on the last user message — never on the system prompt. Prefix-cache reads only hit when a request's prefix matches a previously-written boundary, and a boundary on the *varying* last user message never matches the next call. So the system prompt — usually the largest, most static prefix — was written to cache on nearly every call and read back on nearly none: a fleet of fresh agents sharing one byte-identical system prompt (multi-user server, batched extraction / structured-output loops) got zero cross-call reuse. Hand-placing a `cachePoint` in the system prompt was the only workaround; this makes `auto` do it automatically.

The Python Anthropic provider couldn't cache the system prompt at all — `stream`/`format_request` dropped `system_prompt_content` and sent `system` as a plain string, leaving nowhere to attach `cache_control` (the TypeScript provider already rendered block form).

### Public API Changes

`CacheConfig` gains a system-prompt knob in both SDKs, shaped identically: Python `system_prompt_ttl: bool | str` mirrors TypeScript `systemPromptTTL?: boolean | CacheTTL`, alongside the existing per-section TTL fields. With caching already configured on an anthropic-strategy model, the system prompt is now cached **by default**:

```python
# Python — new system_prompt_ttl field on CacheConfig
CacheConfig(strategy="auto", ttl="1h")                    # system prompt cached by default, derives 1h
CacheConfig(strategy="auto", ttl="1h", system_prompt_ttl="30m")  # system prompt gets its own 30m duration
CacheConfig(strategy="auto", system_prompt_ttl=False)            # leave the system prompt uncached
```

```typescript
// TypeScript — new systemPromptTTL field, a peer of toolsTTL / messagesTTL
{ ttl: '1h' }                    // system prompt cached by default, derives 1h
{ ttl: '1h', systemPromptTTL: '30m' }  // system prompt gets its own 30m duration
{ systemPromptTTL: false }             // leave the system prompt uncached
```

`True` (or an omitted value) derives the section's duration from the shared `ttl`; a TTL string sets its own duration and is honored verbatim; `False` disables system caching. A hand-placed system cache point is always honored and suppresses auto-injection (deduplicated rather than doubled), so the four-cache-point budget is respected. Tool and message caching are unaffected.

On Bedrock a *derived* system point (from `True`) carries no TTL of its own — the existing normalizer fills it and stands it down behind a shorter tools checkpoint, since Bedrock rejects a longer TTL later in the toolConfig → system → messages chain. An explicit `system_ttl` string skips that stand-down and is written as-is, matching the TS behavior.

### Semi-breaking change

Marked `!` because this changes a default: an existing `auto`/`anthropic` config that cached nothing on the system prompt now writes a system cache point without any code change. The first call of a workload therefore incurs a cache-*write* cost it didn't before.

The one case that pays the write without saving on cache is a system prompt that changes on every request. There the added write is a small standalone cost, mitigated by opting out with `system_prompt_ttl=False` / `systemPromptTTL: false`; the docs call this out under System Prompt Caching. 

## Related Issues

Resolves #3144
Resolves #2160

## Documentation PR

Included 

## Type of Change

Bug fix

## Testing

Ran both SDKs' unit suites plus the provider model tests, covering default-on injection, the `system_ttl=False` / `systemTTL: false` opt-out, an explicit `system_ttl` string honored verbatim (including behind a differing tools TTL), the derived-TTL stand-down behind a shorter tools checkpoint, honoring a hand-placed system point (no doubling; the deprecated `cache_prompt` path stays single), and the ≤4 cache-point budget. Also exercised `format_request`/`_formatRequest` as pure dict builders via a script asserting the emitted `system` shape per case. The cross-session cache-*read* smoke (two calls, fresh agent each, identical system prompt, non-zero `cacheReadInputTokens` on the second) needs Bedrock credentials and runs in CI's provisioned integ job, not locally.

## Checklist

----


site/src/content/docs/user-guide/concepts/model-providers/amazon-bedrock.mdx
```diff
@@ -627,17 +631,17 @@ print(f"Cache read tokens: {response2.metrics.accumulated_usage.get('cacheReadIn
...
 | Configuration | Caches |
 |---------------|--------|
-| `{}` | tool definitions and conversation |
-| `{ messagesTTL: false }` | tool definitions only |
-| `{ toolsTTL: false }` | conversation only |
-| `{ ttl: '1h' }` | both, with a one-hour TTL |
-| `{ ttl: '1h', messagesTTL: '5m' }` | tool definitions at one hour, conversation at five minutes |
+| `{}` | tool definitions, system prompt, and conversation |
+| `{ messagesTTL: false }` | tool definitions and system prompt |
+| `{ toolsTTL: false }` | system prompt and conversation |
```
liramon2: Is there a reason this table doesn't have a row for `{ systemTTL: false}`?

site/src/content/docs/user-guide/concepts/model-providers/amazon-bedrock.mdx
```diff
@@ -627,17 +631,17 @@ print(f"Cache read tokens: {response2.metrics.accumulated_usage.get('cacheReadIn
 </Tab>
 <Tab label="TypeScript">
 
-Setting `cacheConfig` appends a cache point after the tool definitions in each request. The same `cacheConfig` also caches the conversation, and `toolsTTL` and `messagesTTL` control the two independently:
+Setting `cacheConfig` appends a cache point after the tool definitions in each request. The same `cacheConfig` also caches the system prompt and the conversation, and `toolsTTL`, `systemTTL`, and `messagesTTL` control the three independently:
```
zastrowm: > Default-on is scoped to when cache_config is already set.

This can lead to a cost-increase if the system prompt is changing all the time; any way to mitigate that and/or backwards compatibility concerns?

strands-py/src/strands/models/anthropic.py
```diff
@@ -383,6 +383,8 @@ def format_request(
         messages: Messages,
         tool_specs: list[ToolSpec] | None = None,
         system_prompt: str | None = None,
+        *,
```
zastrowm, line 388: This is a breaking change for tool_choice and dynamic_trailling_blocks, no?

strands-py/src/strands/models/model.py
```diff
@@ -143,10 +143,13 @@ class CacheConfig:
             checkpoint TTLs to be non-increasing across toolConfig, system and messages, and rejects a
             longer TTL that follows a shorter one. This TTL also fills in for a cache point that carries
             none of its own.
+        inject_system_cache_point: Whether to auto-inject a cache point at the end of the system prompt so
+            repeated calls with the same static system prefix hit the cache.
     """
 
     strategy: Literal["auto", "anthropic"] = "auto"
     ttl: str | None = None
+    inject_system_cache_point: bool = True
```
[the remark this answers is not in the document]
zastrowm: Oh hey, that's what I said :) 

We should debate the merits of both approaches; if we go forward to this we should make that apparent in the PR description of the repercussions (and also add a ! to the PR description so it's indicating a semi-breaking change)

zastrowm: Should we just go full-throttle parity with TypeScript and do a cache-object as well - deprecating the old way?  (if so, maybe defer python changes to a separate PR)

strands-ts/src/models/model.ts
```diff
@@ -97,6 +97,15 @@ export interface CacheConfig {
...
   toolsTTL?: boolean | CacheTTL
 
+  /**
+   * Cache the system prompt, auto-injecting a cache point at its end so repeated calls with the same
+   * static system prefix hit the cache. A TTL sets this section's duration; `false` disables it. A
+   * hand-placed system cache point is honored rather than doubled.
+   *
+   * @defaultValue true
+   */
+  systemTTL?: boolean | CacheTTL
```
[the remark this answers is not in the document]
zastrowm: LOL, this review is good.  Also flagged above.  I lean towards aligning python with an explicit CacheControl object - preferring the field if set and deprecating the old way.

zastrowm (changes requested):

strands-py/src/strands/models/model.py
zastrowm: Discussed offline.  The nuance here is that for tool-caching in python we previously deciding to align Python Bedrock & Anthropic at the cost of inconsistency with TypeScript.

However, systemTtl changes that a bit. 

We decided:

 - in Python, align the shape of systemTtl caching with that of python
 - going forward, when we enable caching by default, we'll:
   - add a toolsTtl to cache config - matching TypeScript;
   - we'd default `toolsTtl` to true (see next bullet point)
   - we'll deprecate `cache_tools`, but continue to use it if toolsTTL is unset by the caller

opieter-aws: Updated! 

@strandly-the-agent can you create an issue to add `tools_ttl` to cacheconfig in python and deprecate `cache_tools`?

strandly-the-agent: Done — filed #3893 capturing the decision above (add `tools_ttl: bool | str` to Python's `CacheConfig` matching TS `toolsTTL`, deprecate `cache_tools` but keep honoring it when `tools_ttl` is unset, default `True` once caching goes default-on).

Note: #1577 already asked to deprecate `cache_tools` but was blocked on the migration path — #3893 cross-references it as the answer; you may want to close #1577 as consolidated once triaged.

strands-py/src/strands/models/anthropic.py
opieter-aws: Thanks for catching!

strands-ts/src/models/model.ts
zastrowm: did we decide against `systemPromptTTL` because of verbose-ness.  `system` by itself is a little ambiguous.

strands-ts/src/models/model.ts
```diff
@@ -97,6 +97,15 @@ export interface CacheConfig {
    */
   toolsTTL?: boolean | CacheTTL
 
+  /**
+   * Cache the system prompt, auto-injecting a cache point at its end so repeated calls with the same
+   * static system prefix hit the cache. A TTL sets this section's duration; `false` disables it. A
+   * hand-placed system cache point is honored rather than doubled.
```
zastrowm:
```suggestion
   * Cache the system prompt, auto-injecting a cache point at its end so repeated calls with the same
   * static system prefix hit the cache. A TTL sets this section's duration; `true` (the default) reads
   * the value from `ttl`; `false` disables systemPrompt cache injection.
```

strands-ts/src/models/model.ts
opieter-aws: I didn't really entertain the option of `systemPromptTTL`, not opposed! Will update

site/src/content/docs/user-guide/concepts/model-providers/amazon-bedrock.mdx
opieter-aws: Added explicit sentence in ` System Prompt Caching` section above

strands-ts/src/models/__tests__/bedrock.test.ts
```diff
@@ -3028,6 +3029,125 @@ describe('BedrockModel', () => {
...
+
+      const call = mockConverseStreamCommand.mock.lastCall?.[0]
+      expect(call?.system).toStrictEqual([
+        { text: 'static prompt' },
+        { cachePoint: { type: 'default', ttl: '1h' } },
+        { text: 'trailing instructions' },
+      ])
+    })
+
+    it('stands the system cache point TTL down when the tools point carries a different TTL', async () => {
```
zastrowm, line 3114: > stands the system cache point TTL down when the tools point carries a different TTL

What does **stands** mean in this context?

zastrowm (approved):

strands-ts/src/models/__tests__/bedrock.test.ts
opieter-aws: System ttl stands down if tools ttl is also provided

zastrowm: FWIW, "tools cache takes priority over" or something is a bit more clear

Issue #2970, opened 2026-06-25
[FEATURE] Default Agents to turn on caching


notowen333:
### SDK Language

Both / Not language-specific

### Problem Statement

Caching is straightforward to configure and leads to substantial (~10x) cost savings on cache hits. In Strands, if I specify an Agent with a given model provider, caching is only turned on if I explicitly provide a cache config. In the default experience, caching is turned off.

### Proposed Solution

We default cache configs to a sensible `'auto'` mode in every provider where we vend an `'auto'` mode. In parallel, we build `'auto'` mode for all model providers.

For a provider like Bedrock, we can ship this very quickly because 1) we already have `auto` mode and 2) Pending more verification, I believe there is no IAM permissions necessary for caching which is not necessary for standard invocation.

For API key based providers, permissions would not pose an issue. Implementing `auto` and accepting the new default behavior would be the only concern. We do not currently support an `auto` config for all providers.

This would be technically a "breaking change", but I think it's worth the tradeoff. 

### Use Case

Arbitrary use cases can/will benefit from savings. All model calls with cache hits will reduce costs and cache misses will lead to the same outcomes as if caching was not enabled.

### Alternatives Solutions

We can push this off until V2, but I think it's possible to make a path for V1 support

PR #3682, opened 2026-08-06
feat(litellm): route cache_config per underlying provider

arielnabavian:
## Summary

LiteLLM wraps many provider APIs behind one interface, so an SDK-level `cache_config` has to route to whatever the underlying provider actually supports:

- **Anthropic-family** (`anthropic/`, `bedrock/anthropic.`, `vertex_ai/claude`, `vertex_ai_beta/claude`) accept `cache_control` markers. Auto mode injects a `cachePoint` on the last stable system block and the last user message; the existing translator turns them into `cache_control: ephemeral` markers before dispatch.
- **OpenAI-compatible** providers (`openai/`, `deepseek/`, `xai/`, `azure/`) cache automatically server-side. `cache_config` is accepted for portability and is a documented no-op.
- **Unrecognized** underlying providers warn once per resolution and skip injection.

Cache breakpoints on earlier turns are stripped before injection so a long conversation stays within Anthropic's 4-breakpoint request budget.

## Stack

3/9 — stacks on top of #3681. Related: #2970.

## Test plan

- [x] 62 Python litellm tests pass (10 new)
- [x] mypy, ruff clean
- [ ] Real-API smoke test on `anthropic/claude-sonnet-4-6`, `bedrock/anthropic.claude-sonnet-4-6`, `openai/gpt-4o`, `gemini/gemini-2.5-flash`

### DIFF
--- a/strands-py/src/strands/models/bedrock.py
+++ b/strands-py/src/strands/models/bedrock.py
@@ -288,6 +288,13 @@ class BedrockModel(Model):
             )
             system_blocks.append({"cachePoint": {"type": cache_prompt}})
 
+        # Auto-inject a cachePoint at the end of the system prompt so repeated calls with
+        # the same static system prefix hit the cache. Anthropic docs describe the
+        # tools → system → messages prefix chain; caching only messages leaves the
+        # (usually largest and most static) system prefix uncached.
+        if self._should_cache_system(system_blocks):
+            system_blocks.append(self._build_system_cache_point())
+
         return {
             "modelId": self.config["model_id"],
             "messages": self._format_bedrock_messages(messages),
@@ -386,6 +393,50 @@ class BedrockModel(Model):
             return {}
 
         return {"additionalModelRequestFields": additional_fields}
+
+    def _should_cache_system(self, system_blocks: list[SystemContentBlock]) -> bool:
+        """Whether to auto-inject a cache point at the end of the system prompt.
+
+        True only when auto caching is enabled for this model, ``system_blocks`` has cacheable
+        content, and no cachePoint has already been placed at the end of the system prefix.
+
+        Args:
+            system_blocks: The system content blocks that will be sent to Bedrock.
+
+        Returns:
+            True if a cache point should be appended.
+        """
+        cache_config = self.config.get("cache_config")
+        if not cache_config:
+            return False
+
+        resolved: str | None = cache_config.strategy
+        if resolved == "auto":
+            resolved = self._cache_strategy
+        if resolved != "anthropic":
+            return False
+
+        if not system_blocks:
+            return False
+
+        if "cachePoint" in system_blocks[-1]:
+            return False
+
+        return True
+
+    def _build_system_cache_point(self) -> SystemContentBlock:
+        """Build the cache point block appended to the system prompt.
+
+        Uses ``cache_config.ttl`` when set; falls back to Bedrock's ``default`` (5m).
+
+        Returns:
+            The cache point block.
+        """
+        cache_point: dict[str, Any] = {"type": "default"}
+        cache_config = self.config.get("cache_config")
+        if cache_config and cache_config.ttl:
+            cache_point["ttl"] = cache_config.ttl
+        return cast(SystemContentBlock, {"cachePoint": cache_point})
 
     def _build_tools_cache_point(self) -> list[dict[str, Any]]:
         """Build the cache point block appended to ``toolConfig.tools`` if ``cache_tools`` is configured.
--- a/strands-py/src/strands/models/litellm.py
+++ b/strands-py/src/strands/models/litellm.py
@@ -22,10 +22,19 @@ from ..types.exceptions import ContextWindowOverflowException
 from ..types.streaming import MetadataEvent, StreamEvent
 from ..types.tools import ToolChoice, ToolSpec, ToolUse
 from ._validation import validate_config_keys
-from .model import BaseModelConfig
+from .model import BaseModelConfig, CacheConfig
 from .openai import OpenAIModel
 
 logger = logging.getLogger(__name__)
+
+# Underlying providers whose native APIs support Anthropic-style cache_control markers.
+# LiteLLM already translates cachePoint → cache_control for these on the system prompt;
+# when cache_config resolves to "anthropic" we also mark the last user message.
+_ANTHROPIC_CACHE_LITELLM_PREFIXES = ("anthropic/", "bedrock/anthropic.", "vertex_ai/claude", "vertex_ai_beta/claude")
+
+# Underlying providers that cache automatically on the server side. cache_config is
+# accepted for portability but the SDK doesn't need to inject markers.
+_SERVER_AUTO_CACHE_LITELLM_PREFIXES = ("openai/", "deepseek/", "xai/", "azure/")
 
 # Separator used by LiteLLM to embed thought signatures inside tool call IDs.
 # See: https://ai.google.dev/gemini-api/docs/thought-signatures
@@ -47,11 +56,17 @@ class LiteLLMModel(OpenAIModel):
                 For a complete list of supported parameters, see
                 https://docs.litellm.ai/docs/completion/input#input-params-1.
             stream: Whether to use streaming. Defaults to True.
+            cache_config: Configuration for prompt caching. Use CacheConfig(strategy="auto") to
+                let LiteLLM route caching to the underlying provider. Behaviour depends on the
+                model_id prefix: Anthropic/Bedrock-Anthropic/Vertex-Claude get cache markers on
+                the system prompt and last user message; OpenAI-compatible providers cache
+                automatically server-side and this config is a no-op; other providers warn.
         """
 
         model_id: str
         params: dict[str, Any] | None
         stream: bool
+        cache_config: CacheConfig | None
 
     def __init__(self, client_args: dict[str, Any] | None = None, **model_config: Unpack[LiteLLMConfig]) -> None:
         """Initialize provider instance.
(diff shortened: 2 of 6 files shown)


strands-py/src/strands/models/bedrock.py
```diff
@@ -387,6 +394,50 @@ def _get_additional_request_fields(self, tool_choice: ToolChoice | None) -> dict
 
         return {"additionalModelRequestFields": additional_fields}
 
+    def _should_cache_system(self, system_blocks: list[SystemContentBlock]) -> bool:
```
arielnabavian: **concern** — this commit already landed. `6383aa65 fix(bedrock): cache system prompt in auto mode` is a precursor to #3681, which merged as `697c833d` with a different API. The branch is 200 commits behind `main`.

What `main` has today:

- `_should_cache_system` gates on `cache_config.system_prompt_ttl is False`, and skips when **any** system block already carries a cachePoint;
- TTL is applied separately by `_apply_system_cache_ttl`, not baked into the appended point;
- `CacheConfig` carries `system_prompt_ttl: bool | str` and `cache_key`. This branch's `CacheConfig` has only `strategy` and `ttl`.

So if this version survives the rebase it silently reverts two shipped behaviours: `CacheConfig(system_prompt_ttl=False)` stops disabling system caching, and the "caller already placed a point anywhere" check narrows to "…at the end of the list".

Suggest dropping this commit and rebasing so the PR is just the litellm change.

strands-ts/src/models/bedrock.ts
```diff
@@ -153,6 +153,9 @@ export interface BedrockCacheConfig extends CacheConfig {
 
   /** TTL applied to the auto-injected cache point appended to the last user message. */
   messagesTTL?: BedrockCacheTTL
+
+  /** TTL applied to the auto-injected cache point appended to the system prompt. */
+  systemTTL?: BedrockCacheTTL
```
arielnabavian: **concern** — `main` already has `systemPromptTTL?: boolean | CacheTTL` on this interface, landed with #3681. Adding `systemTTL` leaves two public knobs for one thing, and whichever one loses becomes dead config that still type-checks.

Same root cause as the `bedrock.py` note: this commit predates #3681.

strands-py/src/strands/models/litellm.py
```diff
@@ -236,6 +289,53 @@ def _format_system_messages(
...
+                        if ttl := block["cachePoint"].get("ttl"):
+                            cache_control["ttl"] = ttl
+                        cache_markers.append((msg_idx, len(new_content) - 1, cache_control))
+                    continue
+                new_content.append(cast(dict[str, Any], block))
+            stripped_messages.append({"role": message["role"], "content": new_content})
+
+        formatted = super()._format_regular_messages(cast(Messages, stripped_messages), **kwargs)
+
+        # Reattach cache_control to each marked (message, block) position. This works because
```
arielnabavian: **concern** — the index assumption in this comment does not hold, and the injected cache point is silently lost or misplaced as soon as any message carries a `toolResult`.

`OpenAIModel._format_regular_messages` appends **1 + one per `toolResult` + one per extracted image** per input message (openai.py:436-452). So `formatted[msg_idx]` stops matching input `msg_idx` at the first tool result — the splits do not all land after the messages you mark.

I ran the real code, this PR's `_format_regular_messages` and `_inject_messages_cache_point` spliced onto `main`'s parent:

**Last message is a `toolResult`** — every tool round-trip request:

```
[0] user       [{'text': 'what is the weather', 'type': 'text'}]
[1] assistant  tool_calls=1
[2] user       None
[3] tool       'sunny'
--> cache_control landed on: NOTHING
```

The marker for input index 2 hits `formatted[2]`, which has no `content` key at all. And even when it pointed at the tool message, `format_request_tool_message` returns `content` as a plain **string** for an all-text result (openai.py:268), so `isinstance(formatted_content, list)` is False and the marker is dropped.

**A new user turn after one tool round-trip:**

```
[4] assistant  [{'text': 'It is sunny.', ..., 'cache_control': {'type': 'ephemeral'}}]
[5] user       [{'text': 'and tomorrow?', 'type': 'text'}]
--> cache_control landed on: [(4, 'assistant', 'It is sunny.')]
```

`_inject_messages_cache_point` put the point on the last user message; the remap wrote it to the assistant message one slot earlier. With more tool round-trips the drift grows, so it will land on a `tool` message or past the end.

Neither case raises, so it fails quiet — you pay cache writes and get the wrong prefix, or no breakpoint at all. An agent loop is the workload this feature exists for.

Index arithmetic is the wrong coupling. `format_request_message_content` returns a fresh dict per block (openai.py:199), so marking by block identity would work — or have the parent hand back the input→output mapping — so the two functions can't drift apart.

Please add a test with a `toolUse`/`toolResult` round-trip. Codecov already flags 12 uncovered lines in this file, and these are them.

strands-py/src/strands/models/litellm.py
```diff
@@ -236,6 +289,53 @@ def _format_system_messages(
         # Create single system message with content array rather than mulitple system messages
         return [{"role": "system", "content": system_content}] if system_content else []
 
+    @override
+    @classmethod
+    def _format_regular_messages(cls, messages: Messages, **kwargs: Any) -> list[dict[str, Any]]:
```
arielnabavian: **concern** — this override is a `@classmethod`, so it can't read `self.config`. It rewrites `cachePoint` → `cache_control` for **every** LiteLLM model, whether or not `cache_config` is set.

Ran it: `openai/gpt-4o`, **no `cache_config` at all**, one caller-placed message `cachePoint`:

```
[0] user [{'text': 'big document ...', 'type': 'text', 'cache_control': {'type': 'ephemeral'}}]
```

Before this PR the parent dropped the block and said so — `"cachePoint content block is not supported by OpenAI | skipping"` (openai.py:421-423). So this quietly starts sending an Anthropic-only field to openai / deepseek / xai / azure / gemini, on a path unrelated to the feature.

`test_format_request_auto_noop_for_openai_underlying` only asserts that *auto-injection* doesn't happen; it never places a cachePoint itself, so it can't catch this.

If the translation is meant to be Anthropic-only it needs the strategy check, which means it can't stay a `classmethod`.

strands-py/src/strands/models/litellm.py
```diff
@@ -22,11 +22,20 @@
...
-from .model import BaseModelConfig
+from .model import BaseModelConfig, CacheConfig
 from .openai import OpenAIModel
 
 logger = logging.getLogger(__name__)
 
+# Underlying providers whose native APIs support Anthropic-style cache_control markers.
+# LiteLLM already translates cachePoint → cache_control for these on the system prompt;
+# when cache_config resolves to "anthropic" we also mark the last user message.
+_ANTHROPIC_CACHE_LITELLM_PREFIXES = ("anthropic/", "bedrock/anthropic.", "vertex_ai/claude", "vertex_ai_beta/claude")
```
arielnabavian: **concern** — `"bedrock/anthropic."` misses cross-region inference profiles, which is how most Claude-on-Bedrock traffic is addressed now. Measured against this code:

| model_id | `_cache_strategy` |
|---|---|
| `bedrock/anthropic.claude-sonnet-4-5-…` | `anthropic` |
| `bedrock/us.anthropic.claude-sonnet-4-5-…` | **None** |
| `bedrock/eu.anthropic.claude-sonnet-4-5-…` | **None** |
| `bedrock/converse/us.anthropic.claude-…` | **None** |
| `openrouter/anthropic/claude-sonnet-4.5` | **None** |

`None` means warn and inject nothing, so the most common ids get no caching while looking configured.

`BedrockModel._cache_strategy` solves the same problem with a substring test — `"claude" in model_id or "anthropic" in model_id` (bedrock.py:319 on `main`). A prefix allowlist here means the two providers disagree about what an Anthropic model is, and each new region prefix or router is another silent miss. Reusing the substring rule, or a shared helper, closes the whole class.

strands-py/src/strands/models/litellm.py
```diff
@@ -318,6 +418,117 @@ def format_chunk(self, event: dict[str, Any], **kwargs: Any) -> StreamEvent:
...
+
+        cache_point: dict[str, Any] = {"type": "default"}
+        cache_config = cast(CacheConfig | None, self.config.get("cache_config"))
+        if cache_config and cache_config.ttl:
+            cache_point["ttl"] = cache_config.ttl
+
+        cleaned: list[dict[str, Any]] = []
+        last_user_idx: int | None = None
+        for idx, message in enumerate(messages):
+            new_content = [block for block in message["content"] if "cachePoint" not in block]
```
arielnabavian: **concern** — this strips *every* caller-placed `cachePoint`, not just stale auto-injected ones, so a hand-pinned breakpoint is deleted.

Ran it with a point pinned after a large document in turn 1:

```
in:  user      ["<80k token contract>", cachePoint, "summarise clause 4"]
     assistant [...]
     user      ["now clause 5"]

out: [0] user  '<80k token contract>', 'summarise clause 4'   <- point gone
     [1] assistant
     [2] user  'now clause 5'  cache_control                  <- only marker
```

The document stops being a cache boundary, which is the only reason anyone pins a point there.

Two things make this look unintended rather than chosen. `_inject_system_cache_point` right above deliberately does the opposite — it returns early on a caller-placed point and keeps its ttl. And merged Bedrock honours hand-placed points too.

The 4-breakpoint budget in the docstring is a real constraint but this is the wrong lever: Anthropic allows four, so keeping the caller's point *and* adding yours is legal and strictly better. If the budget is the worry, count existing points and skip injection when full.

strands-py/src/strands/models/litellm.py
```diff
@@ -318,6 +418,117 @@ def format_chunk(self, event: dict[str, Any], **kwargs: Any) -> StreamEvent:
         # For all other cases, use the parent implementation
         return super().format_chunk(event)
 
+    @override
+    def format_request(
```
arielnabavian: **concern** — after the rebase this override collides with `_openai_cache.apply_cache_config`, which `OpenAIModel.format_request` now calls unconditionally (openai.py:536).

`main` has already decided how LiteLLM consumes `cache_config`. The `LiteLLMConfig` docstring says "Consumed by LiteLLM's OpenAI-compatible request path exactly as by `OpenAIModel`" (litellm.py:50-51), and `apply_cache_config` acts on that:

- it warns `"openai caches prompt prefixes automatically server-side; cache_config.strategy and system_prompt_ttl have no effect and will be ignored"` whenever `strategy != "auto"`;
- it sets `prompt_cache_key` / `prompt_cache_retention` on the request.

This PR makes `strategy` meaningful for LiteLLM. So `super().format_request(...)` will warn that the user's setting is ignored while this code is honouring it, and will put OpenAI-only request keys on a request bound for Anthropic.

Worth deciding which layer owns `cache_config` for LiteLLM and updating the other one, before this shape is public.

strands-py/src/strands/models/litellm.py
```diff
@@ -89,6 +104,44 @@ def get_config(self) -> LiteLLMConfig:
...
+        """
+        cache_config = cast(CacheConfig | None, self.config.get("cache_config"))
+        if not cache_config:
+            return None
+
+        resolved: str | None = cache_config.strategy
+        if resolved == "auto":
+            resolved = self._cache_strategy
+            if resolved is None:
+                logger.warning(
```
arielnabavian: **improvement** — this fires once per **request**, not "once per resolution" as the description says: `_resolve_cache_strategy` is called from `format_request`. A 30-turn loop on `gemini/…` with `cache_config` set prints it 30 times.

The repo already has an idiom for this — `model.py:327-328` guards with `getattr(self, "_utilization_limit_warned", False)`.

PR #3689, opened 2026-08-06
feat(bedrock): default cache_config to auto

### LINKED ISSUE #2970: [FEATURE] Default Agents to turn on caching

arielnabavian:
## Summary

**Flips the default:** BedrockModel now constructs with `CacheConfig(strategy="auto")` instead of no caching. For models that support Anthropic-style caching (Claude family, `us./eu./global.anthropic.*` cross-region profiles), this places cache points on the system prompt and last user message on every request; tools caching still requires the explicit `cache_tools` knob.

For non-Anthropic models on Bedrock (Amazon Nova, DeepSeek, Meta Llama, Mistral, Amazon Titan, etc.) the runtime silently skips injection — `_shouldEnableCaching` / `_cache_strategy` returns null — and logs a "model does not support automatic caching" warning at request time. **No request payload changes for those models.**

Callers who want to opt out pass `cache_config=None` (Python) or `cacheConfig: undefined` (TypeScript) explicitly.

## Compatibility

Behavior change is technically breaking, but a cache miss falls back to normal input pricing so the worst case is neutral, and the best case is ~90% off input on repeated prefixes. Recommend a release-note callout.

The AnthropicModel default flip is intentionally deferred until upstream #3571 lands (adds `cache_config` to AnthropicModel); this stack composes on top.

## Stack

**10/10 — the finale.** Closes #2970.

## Test plan

- [x] 189 Python bedrock tests pass (2 new)
- [x] 178 TS bedrock tests pass (extensive snapshot updates)
- [x] Full Python suite: 4805 tests
- [x] Full TS suite: 3940 tests
- [x] mypy, ruff, prettier, eslint clean
- [ ] Real-API smoke test on Claude Sonnet 4.6 (default cache write on turn 1, cache read on turn 2)
- [ ] Real-API smoke test confirming Nova/DeepSeek/Llama/Mistral models still work with the new default (warning + no cachePoint injection)
- [ ] Verify `Agent(model=BedrockModel())` and `Agent()` (default model) both cache with no config

### DIFF
--- a/strands-py/src/strands/models/bedrock.py
+++ b/strands-py/src/strands/models/bedrock.py
@@ -288,6 +288,13 @@ class BedrockModel(Model):
             )
             system_blocks.append({"cachePoint": {"type": cache_prompt}})
 
+        # Auto-inject a cachePoint at the end of the system prompt so repeated calls with
+        # the same static system prefix hit the cache. Anthropic docs describe the
+        # tools → system → messages prefix chain; caching only messages leaves the
+        # (usually largest and most static) system prefix uncached.
+        if self._should_cache_system(system_blocks):
+            system_blocks.append(self._build_system_cache_point())
+
         return {
             "modelId": self.config["model_id"],
             "messages": self._format_bedrock_messages(messages),
@@ -386,6 +393,50 @@ class BedrockModel(Model):
             return {}
 
         return {"additionalModelRequestFields": additional_fields}
+
+    def _should_cache_system(self, system_blocks: list[SystemContentBlock]) -> bool:
+        """Whether to auto-inject a cache point at the end of the system prompt.
+
+        True only when auto caching is enabled for this model, ``system_blocks`` has cacheable
+        content, and no cachePoint has already been placed at the end of the system prefix.
+
+        Args:
+            system_blocks: The system content blocks that will be sent to Bedrock.
+
+        Returns:
+            True if a cache point should be appended.
+        """
+        cache_config = self.config.get("cache_config")
+        if not cache_config:
+            return False
+
+        resolved: str | None = cache_config.strategy
+        if resolved == "auto":
+            resolved = self._cache_strategy
+        if resolved != "anthropic":
+            return False
+
+        if not system_blocks:
+            return False
+
+        if "cachePoint" in system_blocks[-1]:
+            return False
+
+        return True
+
+    def _build_system_cache_point(self) -> SystemContentBlock:
+        """Build the cache point block appended to the system prompt.
+
+        Uses ``cache_config.ttl`` when set; falls back to Bedrock's ``default`` (5m).
+
+        Returns:
+            The cache point block.
+        """
+        cache_point: dict[str, Any] = {"type": "default"}
+        cache_config = self.config.get("cache_config")
+        if cache_config and cache_config.ttl:
+            cache_point["ttl"] = cache_config.ttl
+        return cast(SystemContentBlock, {"cachePoint": cache_point})
 
     def _build_tools_cache_point(self) -> list[dict[str, Any]]:
         """Build the cache point block appended to ``toolConfig.tools`` if ``cache_tools`` is configured.
--- a/strands-py/src/strands/models/bedrock.py
+++ b/strands-py/src/strands/models/bedrock.py
@@ -101,7 +101,10 @@ class BedrockModel(Model):
             additional_request_fields: Additional fields to include in the Bedrock request
             additional_response_field_paths: Additional response field paths to extract
             cache_prompt: Cache point type for the system prompt (deprecated, use cache_config)
-            cache_config: Configuration for prompt caching. Use CacheConfig(strategy="auto") for automatic caching.
+            cache_config: Configuration for prompt caching. Defaults to CacheConfig(strategy="auto"),
+                which places cache points on the system prompt, tools (when tool caching is
+                separately enabled via cache_tools), and last user message when the model
+                supports Anthropic-style caching. Pass cache_config=None to disable caching.
             cache_tools: Cache point type for tools. Pass a string (e.g. "default") for the default 5m TTL,
                 or a CacheToolsConfig instance to set both type and TTL (e.g. "1h").
             guardrail_id: ID of the guardrail to apply
@@ -193,6 +196,7 @@ class BedrockModel(Model):
         self.config = BedrockModel.BedrockConfig(
             model_id=BedrockModel._get_default_model_with_warning(resolved_region, model_config),
             include_tool_result_status="auto",
+            cache_config=CacheConfig(strategy="auto"),
         )
         self.update_config(**model_config)
 
--- a/strands-py/src/strands/models/litellm.py
+++ b/strands-py/src/strands/models/litellm.py
@@ -22,10 +22,19 @@ from ..types.exceptions import ContextWindowOverflowException
 from ..types.streaming import MetadataEvent, StreamEvent
 from ..types.tools import ToolChoice, ToolSpec, ToolUse
 from ._validation import validate_config_keys
-from .model import BaseModelConfig
+from .model import BaseModelConfig, CacheConfig
 from .openai import OpenAIModel
 
 logger = logging.getLogger(__name__)
+
+# Underlying providers whose native APIs support Anthropic-style cache_control markers.
+# LiteLLM already translates cachePoint → cache_control for these on the system prompt;
+# when cache_config resolves to "anthropic" we also mark the last user message.
+_ANTHROPIC_CACHE_LITELLM_PREFIXES = ("anthropic/", "bedrock/anthropic.", "vertex_ai/claude", "vertex_ai_beta/claude")
+
+# Underlying providers that cache automatically on the server side. cache_config is
+# accepted for portability but the SDK doesn't need to inject markers.
+_SERVER_AUTO_CACHE_LITELLM_PREFIXES = ("openai/", "deepseek/", "xai/", "azure/")
 
 # Separator used by LiteLLM to embed thought signatures inside tool call IDs.
 # See: https://ai.google.dev/gemini-api/docs/thought-signatures
@@ -47,11 +56,17 @@ class LiteLLMModel(OpenAIModel):
                 For a complete list of supported parameters, see
                 https://docs.litellm.ai/docs/completion/input#input-params-1.
             stream: Whether to use streaming. Defaults to True.
+            cache_config: Configuration for prompt caching. Use CacheConfig(strategy="auto") to
+                let LiteLLM route caching to the underlying provider. Behaviour depends on the
+                model_id prefix: Anthropic/Bedrock-Anthropic/Vertex-Claude get cache markers on
+                the system prompt and last user message; OpenAI-compatible providers cache
+                automatically server-side and this config is a no-op; other providers warn.
         """
 
         model_id: str
         params: dict[str, Any] | None
         stream: bool
+        cache_config: CacheConfig | None
 
     def __init__(self, client_args: dict[str, Any] | None = None, **model_config: Unpack[LiteLLMConfig]) -> None:
         """Initialize provider instance.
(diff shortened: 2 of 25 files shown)


strands-ts/src/models/bedrock.ts
```diff
@@ -682,6 +694,22 @@ export class BedrockModel extends Model<BedrockModelConfig> {
       }
     }
 
+    // Auto-inject a cachePoint at the end of the system prompt so repeated calls with the
+    // same static system prefix hit the cache. Bedrock (Anthropic) documents the prefix
+    // chain as tools → system → messages; caching only messages leaves the (usually largest
+    // and most static) system prefix uncached.
+    if (request.system && request.system.length > 0 && this._shouldEnableCaching()) {
```
arielnabavian: `_shouldEnableCaching()` calls `logger.warn` — not `warnOnce` — every time it resolves `auto` against a model that doesn't support caching (`bedrock.ts:499`). This new call site makes that the third invocation per request: system (here), tools (739), messages (856).

With `cacheConfig: { strategy: 'auto' }` now the constructor default, `new BedrockModel({ modelId: 'amazon.nova-pro-v1:0' })` with a system prompt emits 2–3 identical WARNING lines on **every** request, for the life of the process. Same for Titan, Llama, Mistral, DeepSeek. The user never asked for caching, and the only way to silence it is to discover `cacheConfig: undefined`.

`warnOnce` is already imported in this file (line 64) and used at 445 and 585. Either wrap the warn at 499 in it, or resolve the strategy once per request and hand the result to the three call sites.

strands-py/src/strands/models/bedrock.py
```diff
@@ -193,6 +196,7 @@ def __init__(
         self.config = BedrockModel.BedrockConfig(
             model_id=BedrockModel._get_default_model_with_warning(resolved_region, model_config),
             include_tool_result_status="auto",
+            cache_config=CacheConfig(strategy="auto"),
```
arielnabavian: Same warn-per-request problem Python-side. With this default, `_format_bedrock_messages` reaches `logger.warning("... does not support automatic caching")` at `bedrock.py:628` on every request for every non-Claude Bedrock model. It was unreachable before, because `cache_config` was unset by default and that block is guarded by `if cache_config:`.

The test change in this PR is the tell:

```python
-        warning_records = [r for r in caplog.records if r.levelno == logging.WARNING]
-        assert len(warning_records) == 1
+        matching_records = [... "bedrock:CountTokens permission denied" in r.message ...]
```

That exact-count assertion was loosened because a second, unrelated warning started firing. Worth reading as a signal rather than test noise — warn once per `model_id`, or drop it to `debug`, since under an opt-out default the user never asked for caching in the first place.

strands-ts/src/models/__tests__/bedrock.test.ts
```diff
@@ -799,7 +820,10 @@ describe('BedrockModel', () => {
     })
 
     it('preserves ttl on user-supplied cache point blocks in messages', async () => {
-      const provider = new BedrockModel()
+      // Use a non-Anthropic model so auto caching is a no-op and the user-supplied cache point
+      // (including its TTL) passes through untouched. Auto mode strips and re-injects cache
+      // points, dropping caller-provided TTL.
+      const provider = new BedrockModel({ modelId: 'amazon.titan-text-express-v1' })
```
arielnabavian: This test (and the one at 873) used to run against the default Claude model and guarded that a caller's `CachePointBlock` TTL survives into the request. Re-pointing both at Titan means neither crosses the auto-caching path any more — so the behaviour they existed to protect is now untested for exactly the models where cache TTL matters.

The comment is honest about why: auto mode "strips and re-injects cache points, dropping caller-provided TTL." That's `_injectCachePoint` (`bedrock.ts:885-906`) deleting every caller cache point and rebuilding from `cacheConfig.messagesTTL`. Fine as opt-in. As the new default:

```ts
new BedrockModel({ modelId: 'us.anthropic.claude-sonnet-4-6' })  // no cacheConfig passed
// message content: [TextBlock('...'), CachePointBlock({ cacheType: 'default', ttl: '1h' })]
// -> request carries { cachePoint: { type: 'default' } } — ttl silently gone
```

Someone who deliberately placed a 1h breakpoint now gets 5m, plus a warning telling them their cache point was stripped. Either carry the TTL already on the block when rebuilding, or leave caller-placed points alone and only inject when there are none — and keep a test on a Claude model that proves it.

strands-py/src/strands/models/litellm.py
```diff
@@ -236,6 +289,53 @@ def _format_system_messages(
...
+        formatted = super()._format_regular_messages(cast(Messages, stripped_messages), **kwargs)
+
+        # Reattach cache_control to each marked (message, block) position. This works because
+        # _format_regular_messages preserves 1-to-1 mapping between input messages and the
+        # primary formatted message; tool-message splits append AFTER, so indices stay stable
+        # for the user/assistant messages we mark.
+        for msg_idx, block_idx, cache_control in cache_markers:
+            if msg_idx >= len(formatted):
+                continue
+            formatted_content = formatted[msg_idx].get("content")
```
arielnabavian: This remap breaks on any conversation that has used a tool.

The comment above says the parent "preserves 1-to-1 mapping between input messages and the primary formatted message; tool-message splits append AFTER." They append after *that message*, inside the loop — not after the whole list. `OpenAIModel._format_regular_messages` (`openai.py:443-454`) appends the primary message, then one message per `toolResult` block, then any image-carrying user messages, before moving to the next input message. So every `toolResult` earlier in the conversation shifts all later indices.

```python
messages = [
    {"role": "user",      "content": [{"text": "hi"}]},
    {"role": "assistant", "content": [{"toolUse": ...}]},
    {"role": "user",      "content": [{"toolResult": ...}]},  # -> 2 formatted messages
    {"role": "user",      "content": [{"text": "thanks"}]},   # cache point here, msg_idx=3
]
```

`formatted` is `[user, assistant, user(empty), tool, user(thanks)]`. `formatted[3]` is the **tool** message. LiteLLM and Anthropic both accept `cache_control` there, so nothing raises — the breakpoint just sits in the wrong place and the prefix you meant to cache isn't the one cached.

`block_idx` has the same shape of bug: it's recorded against the unformatted content array, but the parent filters out `toolUse` / `toolResult` / `reasoningContent` and location-source blocks before formatting, so offsets diverge. There the bounds guard turns it into a silent no-cache instead.

Better to carry the marker on the block itself through formatting, or have the parent return the index mapping, than to recompute position after the fact. Either way this needs a test with a `toolResult` turn — `test_format_request_auto_strips_stale_cache_points_across_turns` is text-only, so it passes today.

strands-py/src/strands/models/litellm.py
```diff
@@ -22,11 +22,20 @@
...
-from .model import BaseModelConfig
+from .model import BaseModelConfig, CacheConfig
 from .openai import OpenAIModel
 
 logger = logging.getLogger(__name__)
 
+# Underlying providers whose native APIs support Anthropic-style cache_control markers.
+# LiteLLM already translates cachePoint → cache_control for these on the system prompt;
+# when cache_config resolves to "anthropic" we also mark the last user message.
+_ANTHROPIC_CACHE_LITELLM_PREFIXES = ("anthropic/", "bedrock/anthropic.", "vertex_ai/claude", "vertex_ai_beta/claude")
```
arielnabavian: `"bedrock/anthropic."` misses the ids people actually use. LiteLLM routes cross-region Bedrock Claude as `bedrock/us.anthropic.claude-sonnet-4-6`, `bedrock/eu.anthropic....`, and also `bedrock/converse/us.anthropic....`. None of those start with `bedrock/anthropic.`.

So `LiteLLMModel(model_id="bedrock/us.anthropic.claude-sonnet-4-6", cache_config=CacheConfig(strategy="auto"))` resolves to `None`, injects nothing, and logs "the underlying provider is not recognized for caching" — about a Claude model. `bedrock.py` in this same PR handles the `us.` / `eu.` / `global.` prefixes, so the two providers now disagree about what a Bedrock Claude id looks like.

`openrouter/anthropic/claude-*` is missing too. Matching `anthropic` or `claude` anywhere after the provider segment — which is what `BedrockModel._cache_strategy` already does — covers all of these.

strands-py/tests/strands/models/test_bedrock.py
```diff
@@ -3892,3 +3895,106 @@ def test_format_request_cache_tools_string_backward_compat(model, messages, mode
...
+    tru_request = model.format_request(messages, system_prompt_content=[{"text": "static"}])
+
+    assert tru_request["system"] == [{"text": "static"}]
+
+
+def test_format_request_anthropic_strategy_appends_system_cache_point(bedrock_client, messages):
+    """Explicit anthropic strategy also appends a system cache point, mirroring auto."""
+    model = BedrockModel(
+        model_id="arn:aws:bedrock:us-east-1:123:application-inference-profile/abc",
+        cache_config=CacheConfig(strategy="anthropic"),
```
arielnabavian: This test uses an inference-profile ARN but pins `strategy="anthropic"` explicitly, so it never exercises `auto` against an ARN — and `auto` doesn't handle it.

`_cache_strategy` (`bedrock.py:234-237`) matches on `"claude" in model_id or "anthropic" in model_id`. An ARN like `arn:aws:bedrock:us-east-1:123:application-inference-profile/abc` contains neither. So a Claude model reached through an application inference profile — the standard setup when you need per-team cost allocation — gets no caching under the new default, plus the per-request "does not support automatic caching" warning.

Silently losing caching for those users is the opposite of what flipping the default is for. Either resolve the profile, or document that ARN callers must set `strategy="anthropic"` by hand — and add an `auto` + ARN test so the gap is visible instead of implied.

strands-py/src/strands/models/openai_responses.py
```diff
@@ -149,12 +149,36 @@ class OpenAIResponsesConfig(BaseModelConfig, total=False):
...
+                through to the request verbatim.
         """
 
         model_id: str
         params: dict[str, Any] | None
         stateful: bool
         use_native_token_count: bool
+        cache_config: CacheConfig | None
+        prompt_cache_key: str | None
+        prompt_cache_retention: str | None
```
arielnabavian: These two are typed wider than their TypeScript twins, and they're the fields it will hurt most to narrow later.

| | Python | TypeScript |
|---|---|---|
| retention | `str \| None` | `'in_memory' \| '24h'` |
| options | `dict[str, Any] \| None` | `{ mode: 'implicit' \| 'explicit' }` |

Both are closed sets and both are forwarded verbatim, so `prompt_cache_retention="24hr"` type-checks, passes every test here, and fails at the vendor at request time. `Literal["in_memory", "24h"]` and a `TypedDict` for options would catch it at the boundary.

Widening a `Literal` later is free; narrowing a `str` is a breaking change. Worth settling before these ship, since they're brand-new public config keys.

strands-py/src/strands/models/mistral.py
```diff
@@ -51,13 +51,25 @@ class MistralConfig(BaseModelConfig, total=False):
...
+                similar requests for cache lookups. Requests sharing the same key and a
+                common prefix hit the cache.
         """
 
         model_id: str
         max_tokens: int | None
         temperature: float | None
         top_p: float | None
         stream: bool | None
+        cache_config: CacheConfig | None
```
arielnabavian: Accepting `cache_config` and doing nothing with it is the one no-op in this set that costs the user money.

Elsewhere the no-op is safe because the vendor caches on its own — OpenAI over 1024 tokens, Gemini 2.5 implicit, Llama on stable prefixes. Mistral doesn't: as the docstring right above says, caching only happens once `prompt_cache_key` is set. So `MistralModel(model_id=..., cache_config=CacheConfig(strategy="auto"))` gets zero caching and zero feedback, while the identical call against Bedrock or LiteLLM caches.

`cache_config` exists for portability, so the portable call should either work or say why it doesn't. A warning — "cache_config has no effect on Mistral without prompt_cache_key" — is enough. The same argument applies more weakly to Llama and the Responses API, where `prompt_cache_key` only improves hit rate rather than enabling caching at all.

strands-ts/src/models/bedrock.ts
```diff
@@ -264,7 +267,12 @@ export interface BedrockModelConfig extends BaseModelConfig {
   stopSequences?: string[]
 
   /**
-   * Configuration for prompt caching.
+   * Configuration for prompt caching. Defaults to `{ strategy: 'auto' }`, which places
+   * cache points on the system prompt and the last user message for models that support
+   * Anthropic-style caching (Claude family, `us.anthropic.*` / `eu.anthropic.*` /
+   * `global.anthropic.*` cross-region profiles). Unsupported models silently skip
+   * injection. Pass `cacheConfig: undefined` (via a spread with the field omitted) or an
+   * empty object with `strategy` unset to disable caching.
```
arielnabavian: This sentence contradicts itself: "Pass `cacheConfig: undefined` (via a spread with the field omitted)". Omitting the field is exactly what gives you the default — the constructor spreads `modelConfig` over `{ cacheConfig: { strategy: 'auto' } }` at `bedrock.ts:438-442`, so an absent key keeps auto. Only an explicitly present `cacheConfig: undefined` overrides it.

The second suggested opt-out, "an empty object with `strategy` unset", isn't obvious either: `_shouldEnableCaching` reads `cacheConfig.strategy`, `{}` leaves it `undefined`, and that falls through to `return strategy === 'anthropic'` → false. It works, but by accident rather than design.

This is the documented escape hatch from a breaking default, so it should be exact and it should be tested. Python has `test_format_request_explicit_cache_config_none_disables_caching`; there's no TypeScript equivalent.

PR #3704, opened 2026-08-07
feat: add injected content behind cache points

opieter-aws:
## Description

This PR appends injected context behind the latest cache point, so cached prompts keep token savings with injection.

`ContextInjector` prepended its per-call text into the latest user message, ahead of the stable conversation. Since that text is regenerated every call, the cached prefix diverged at the injection site and no request ever read a cache entry back. 

The fix needs a boundary, and only the middleware that appends the temporary content knows where it belongs. A provider can't re-derive it. So producers now report a count of trailing blocks on the middleware context, and providers that place prompt-cache breakpoints keep theirs ahead of that count. Nothing is written into `content`, so the providers that reject an in-band `cachePoint` are unaffected.

Two consequences to note:
* Injected text is now appended rather than prepended 
* The count is deliberately end-relative: an absolute index goes stale when a provider's own content cleaning drops earlier blocks.

## Public API Changes

A new optional field on the invoke-model middleware context, mirrored onto `StreamOptions` so it reaches a provider. Both are exported from the TypeScript barrel, so this wants an API review.

Producers add to it:

```python
# Middleware that appends content regenerated on every call reports how many trailing
# blocks of the last user message it added.
return replace(context, messages=folded, per_call_trailing_blocks=context.per_call_trailing_blocks + 1)
```

Providers that place cache breakpoints consume it; every other provider ignores it:

```python
# BedrockModel / AnthropicModel read it from the existing **kwargs tail, so no provider
# signature changes: the reusable prefix ends where the trailing blocks begin.
kwargs.get("per_call_trailing_blocks", 0)
```

```typescript
interface InvokeModelContext {
  /** How many trailing blocks of the last user message are rebuilt on every call. */
  readonly perCallTrailingBlocks?: number
}

interface StreamOptions {
  perCallTrailingBlocks?: number
}
```

Defaults to `0`, and is omitted from the provider call entirely when zero, so an existing `stream()` implementation sees exactly the arguments it saw before. It states a fact about the content, not a caching directive: with no `cache_config` set, a non-zero count emits no breakpoint at all.

## Related Issues

N/A

## Documentation PR

N/A — no user-facing surface for the injector or agentic mode changed. The new field is a middleware-to-provider contract.

## Type of Change

Bug fix

## Testing

Verified end to end against real Bedrock (`claude-sonnet-4-5`), with a unique nonce per run so no stale cache entry could flatter the result. Billed units below are `uncached + 1.25×write + 0.10×read`; note that total token counts are near-identical whether the cache reads or thrashes, which is why the assertions target `cacheReadInputTokens` directly.

| scenario | before | after |
|---|---|---|
| Python, agentic + `cache_config` | 21,769 — 0 reads | 8,511 — reads every turn |
| TypeScript, injector active | 13,925 — 0 reads | 5,545 — reads every turn |

Also exercised the placement rules offline across both providers: multi-producer accumulation, an all-per-call message with no durable prefix, non-PDF document adjacency, and TTL propagation. Mutation-checked the wiring by breaking each producer, each consumer, and the forwarding step in turn; every one fails a test.

## Risks / Callouts

**Injected text moves from the front of the user message to the back.** Anyone relying on injected context preceding the user's own words will see the order flip. This is required for the fix — leading text invalidates the cache from the first block onward — but it is observable.

**Agentic mode starts reading cache entries where it previously wrote them.** For `context_manager="agentic"` with Bedrock `cache_config`, this is a pure win (measured above), but the per-turn cost profile changes shape from "write the whole prefix" to "read the prefix, write the tail".

**A provider that ignores the count is silently correct but unimproved.** Only Bedrock and Anthropic consume it. A third provider adding breakpoint placement later has to honor it, or it will reintroduce this bug with no test failure to warn it — the reason the offline suite asserts block *positions* rather than token totals.

## Checklist
- [ ] Any dependent changes have been merged and published — blocked on #3677 and #3571

----


strands-py/src/strands/_middleware/stages.py
```diff
@@ -39,6 +39,7 @@ class InvokeModelContext:
     invocation_state: dict[str, Any]
     model: Model
     projected_input_tokens: int | None = None
+    per_call_trailing_blocks: int = 0
```
mkmeral: trailing ephemeral blocks maybe? places where you pass these things are per-call anyway, so i'd focus a bit more on ephemerality of those blocks for the naming

strands-py/src/strands/injection/_message_injection.py
```diff
@@ -164,27 +164,29 @@ def _fold_into_last_user_message(messages: Messages, text: str) -> Messages:
...
+    content = [*target["content"], injected]
 
     folded: Message = {"role": target["role"], "content": content}
     if "metadata" in target:
         folded["metadata"] = target["metadata"]
 
     result = list(messages)
     result[target_index] = folded
-    return result
+    # A provider only places its cache point in the last message, so a fold elsewhere reports nothing.
```
mkmeral, line 191: > so a fold elsewhere reports nothing

does that mean, we'll break cache if we fold somewhere else? is that okay?

strands-py/src/strands/injection/_message_injection.py
```diff
@@ -164,27 +164,29 @@ def _fold_into_last_user_message(messages: Messages, text: str) -> Messages:
         text: The text to fold into the most recent user message.
 
     Returns:
-        A new list with the folded message, or the input list when there is no user message.
+        The folded conversation and how many trailing blocks of its last message are per-call
+        content: 1 when the fold landed on the last message, 0 otherwise (including when there is
```
mkmeral, line 168: if I call context injector multiple times, do we add it all to the last user message, meaning 1 is technically okay, we don't have to keep increasing?

strands-py/src/strands/_middleware/stages.py
opieter-aws: ephemeral is already Anthropic’s value for a cache breakpoint (`cache_control: {"type": "ephemeral"}`). Naming the field `trailing_ephemeral_blocks` would make “ephemeral” mean both “is a cache point” and “must not be cached” in the same file. 

If `per_call` is not the ephemerality naming you like, I'm open to something like `transient` or `dynamic`

strands-py/src/strands/injection/_message_injection.py
opieter-aws: Every injected context block is its own block trailing behind the last user message. So using the context injector twice for 2 different things will make the trailing_blocks count go to 2

strands-py/src/strands/injection/_message_injection.py
opieter-aws: Yes indeed. That's expected behaviour. Hand-placed cache points are honored, so a user can still set up caching manually in a different way if they want.

strands-py/src/strands/_middleware/stages.py
opieter-aws: Settled on `dynamic_trailing_blocks`, will update

mkmeral (approved):

PR #3708, opened 2026-08-07
feat: mcp v2 compatible changes under a flag

poshinchen:
## Description

`mcp` 2.0 renamed and relocated several public names, so `strands` will fail when `mcp>=2` is installed. This PR adds an internal compatibility module, `strands.tools.mcp._compat`, so the package imports and runs on both major lines. The dependency pin stays `<2` and nothing changes for current installs. This is groundwork for adopting the MCP 2026-07-28 spec (#1659) under an internal feature flag.

All version-dependent names now resolve through `_compat`:

- `MCP_V2` tells the rest of the codebase which version of MCP is installed. It probes for `ClientSession.discover` instead of parsing the version string, so pre-releases and backports resolve by capability. It replaces the version probe in `mcp_instrumentation.py`, and follow-up work (lifecycle, tasks, native OTel) will branch on this same flag.
- `MCPError` aliases the exception class, which 1.x spells `McpError`.
- `streamable_http_transport(url, headers)` keeps the 1.x call shape on both lines. 2.x dropped the `headers` kwarg in favor of a pre-configured HTTPX client, so on 2.x the adapter creates that client itself and closes it when the transport closes.
- `GetSessionIdCallback` falls back to a plain callable alias on 2.x, which removed protocol sessions. It is only used for typing.
- `ProgressFnT` needs no compat entry; its import location exists on both lines down to the 1.23 pin floor.

A new `mcp-v2-compat` CI job force-installs `mcp==2.0.*` over the pin, verifies `import strands` succeeds, and runs the compat tests against the real 2.x package. The rest of CI keeps testing 1.x only. Without this job, nothing would catch a new 1.x-only import breaking 2.x again.

## Related Issues

#1659

## Documentation PR

No documentation changes needed; the compat module is internal.

## Type of Change

New feature

## Testing

- Full MCP unit suite passes on `mcp` 1.x.
- The compat tests pass against `mcp==2.0.0`, both locally and in the new CI job.

## Checklist

----

### DIFF
--- a/.github/workflows/python-test-lint.yml
+++ b/.github/workflows/python-test-lint.yml
@@ -88,6 +88,49 @@ jobs:
         uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f # v7
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+  mcp-v2-compat:
+    name: MCP 2.x Compat
+    # The dependency pin is `mcp<2`, so the unit-test matrix only ever
+    # exercises the 1.x line and the 2.x branches of `_compat` run against
+    # mocks there. This job force-installs the 2.x line over the pin to verify
+    # against the real package that `import strands` succeeds and the compat
+    # layer resolves the 2.x names. Scoped to the compat tests; the full suite
+    # is not expected to pass on 2.x while the pin holds. Pinned to `2.0.*` so
+    # upstream 2.x releases cannot break unrelated PRs; bump deliberately.
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: read
+    defaults:
+      run:
+        working-directory: strands-py
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+        with:
+          ref: ${{ inputs.ref }}
+          persist-credentials: false
+
+      - name: Set up Python
+        uses: actions/setup-python@v7.0.0
+        with:
+          python-version: '3.10'
+
+      - name: Install package, then force mcp 2.x over the pin
+        run: |
+          pip install --no-cache-dir -e . \
+            "pytest>=9.0.0,<10.0.0" \
+            "pytest-asyncio>=1.0.0,<1.5.0" \
+            "pytest-timeout>=2.0.0,<3.0.0" \
+            "moto>=5.1.0,<6.0.0"
+          pip install --no-cache-dir "mcp==2.0.*"
+
+      - name: Verify import and version flag
+        run: python -c "import strands; from strands.tools.mcp import _compat; assert _compat.MCP_V2"
+
+      - name: Run compat tests
+        run: pytest tests/strands/tools/mcp/test__compat.py -vv
+
   lint:
     name: Lint
     runs-on: ubuntu-latest
--- a/strands-py/src/strands/tools/mcp/_compat.py
+++ b/strands-py/src/strands/tools/mcp/_compat.py
@@ -44,7 +44,7 @@ except ImportError:
 
 
 def streamable_http_transport(
-    url: str, headers: dict[str, Any] | None = None, auth: httpx.Auth | None = None
+    url: str, headers: dict[str, str] | None = None, auth: httpx.Auth | None = None
 ) -> AbstractAsyncContextManager[Any]:
     """Open a streamable HTTP client transport on either `mcp` major line.
 
--- a/strands-py/src/strands/tools/mcp/_compat.py
+++ b/strands-py/src/strands/tools/mcp/_compat.py
@@ -1,0 +1,67 @@
+"""Compatibility layer over the `mcp` 1.x and 2.x lines.
+
+The official `mcp` package renamed and relocated several public names in 2.0.
+Import any version-dependent name from this module instead of from `mcp`
+directly so the rest of the codebase stays version-agnostic. Branch on
+`MCP_V2` only where behavior differs between the two lines; pure renames are
+resolved here once.
+"""
+
+from contextlib import AbstractAsyncContextManager
+from typing import Any
+
+import httpx
+from mcp import ClientSession
+
+__all__ = ["MCP_V2", "GetSessionIdCallback", "MCPError", "streamable_http_transport"]
+
+# Feature-probed rather than version-parsed so pre-releases and backports
+# resolve by capability: `ClientSession.discover` is the 2.x replacement for
+# the removed initialize handshake.
+MCP_V2: bool = hasattr(ClientSession, "discover")
+
+try:
+    from mcp.shared.exceptions import MCPError
+except ImportError:
+    # mcp 1.x spells the class McpError
+    from mcp.shared.exceptions import McpError as MCPError  # type: ignore[attr-defined, no-redef]
+
+try:
+    from mcp.client.streamable_http import GetSessionIdCallback
+except ImportError:
+    # mcp 2.x removed protocol sessions, so its transports never yield a
+    # session-id callback; the alias survives only to type 1.x transports.
+    from collections.abc import Callable
+
+    GetSessionIdCallback = Callable[[], str | None]  # type: ignore[misc, assignment]
+
+
+def streamable_http_transport(
+    url: str, headers: dict[str, Any] | None = None, auth: httpx.Auth | None = None
+) -> AbstractAsyncContextManager[Any]:
+    """Open a streamable HTTP client transport on either `mcp` major line.
+
+    `mcp` 2.x replaced `streamablehttp_client(url, headers=..., auth=...)` with
+    `streamable_http_client(url, http_client=...)`, which takes a
+    pre-configured HTTPX client instead of loose header and auth kwargs. This
+    adapter keeps the 1.x-style call shape for both lines.
+
+    The imports are resolved at call time because each name exists on only
+    one major line.
+
+    Args:
+        url: The MCP server endpoint URL.
+        headers: Optional HTTP headers to send with each request.
+        auth: Optional HTTPX authentication handler for each request.
+
+    Returns:
+        An async context manager yielding the transport's read/write streams.
+    """
+    if MCP_V2:
+        from mcp.client.streamable_http import create_mcp_http_client, streamable_http_client
+
+        return streamable_http_client(url=url, http_client=create_mcp_http_client(headers=headers, auth=auth))
+
+    from mcp.client.streamable_http import streamablehttp_client
+
+    return streamablehttp_client(url=url, headers=headers, auth=auth)
--- a/strands-py/src/strands/tools/mcp/_compat.py
+++ b/strands-py/src/strands/tools/mcp/_compat.py
@@ -5,9 +5,16 @@ Import any version-dependent name from this module instead of from `mcp`
 directly so the rest of the codebase stays version-agnostic. Branch on
 `MCP_V2` only where behavior differs between the two lines; pure renames are
 resolved here once.
+
+mypy note: only one `mcp` line is installed at a time, so each try/except
+branch below is an `attr-defined` error when checked against the other line.
+The branch-level ignores cover whichever line mypy runs under, and the
+per-module `warn_unused_ignores` override in pyproject.toml silences the
+ignores the installed line doesn't need.
 """
 
-from contextlib import AbstractAsyncContextManager
+from collections.abc import AsyncIterator
+from contextlib import AbstractAsyncContextManager, asynccontextmanager
 from typing import Any
 
 import httpx
@@ -21,13 +28,13 @@ __all__ = ["MCP_V2", "GetSessionIdCallback", "MCPError", "streamable_http_transp
 MCP_V2: bool = hasattr(ClientSession, "discover")
 
 try:
-    from mcp.shared.exceptions import MCPError
+    from mcp.shared.exceptions import MCPError  # type: ignore[attr-defined]
 except ImportError:
     # mcp 1.x spells the class McpError
     from mcp.shared.exceptions import McpError as MCPError  # type: ignore[attr-defined, no-redef]
 
 try:
-    from mcp.client.streamable_http import GetSessionIdCallback
+    from mcp.client.streamable_http import GetSessionIdCallback  # type: ignore[attr-defined]
 except ImportError:
     # mcp 2.x removed protocol sessions, so its transports never yield a
     # session-id callback; the alias survives only to type 1.x transports.
@@ -58,10 +65,25 @@ def streamable_http_transport(
         An async context manager yielding the transport's read/write streams.
     """
     if MCP_V2:
-        from mcp.client.streamable_http import create_mcp_http_client, streamable_http_client
+        from mcp.client.streamable_http import (  # type: ignore[attr-defined]
+            create_mcp_http_client,
+            streamable_http_client,
+        )
 
-        return streamable_http_client(url=url, http_client=create_mcp_http_client(headers=headers, auth=auth))
+        # `streamable_http_client` closes an HTTPX client only when it created
+        # it (`client_provided` check in mcp 2.x), so a caller-provided client
+        # must be closed by the caller: enter both context managers together
+        # so the client's lifetime is bound to the transport's.
+        @asynccontextmanager
+        async def _owned_client_transport() -> AsyncIterator[Any]:
+            async with (
+                create_mcp_http_client(headers=headers, auth=auth) as http_client,
+                streamable_http_client(url=url, http_client=http_client) as transport_streams,
+            ):
+                yield transport_streams
 
-    from mcp.client.streamable_http import streamablehttp_client
+        return _owned_client_transport()
 
-    return streamablehttp_client(url=url, headers=headers, auth=auth)
+    from mcp.client.streamable_http import streamablehttp_client  # type: ignore[attr-defined]
+
+    return streamablehttp_client(url=url, headers=headers, auth=auth)  # type: ignore[no-any-return]
--- a/strands-py/tests/strands/tools/mcp/test__compat.py
+++ b/strands-py/tests/strands/tools/mcp/test__compat.py
@@ -1,0 +1,54 @@
+"""Unit tests for the mcp 1.x/2.x compatibility layer."""
+
+from unittest.mock import MagicMock, patch
+
+from strands.tools.mcp import _compat
+from strands.tools.mcp._compat import MCPError, streamable_http_transport
+
+
+def test_mcp_v2_flag_matches_discover_capability():
+    """Test that the flag reflects whether ClientSession has the 2.x discover API."""
+    from mcp import ClientSession
+
+    assert _compat.MCP_V2 is hasattr(ClientSession, "discover")
+
+
+def test_mcp_error_resolves_to_installed_exception():
+    """Test that MCPError is the mcp package's error type regardless of its spelling."""
+    import mcp.shared.exceptions as mcp_exceptions
+
+    installed = getattr(mcp_exceptions, "MCPError", None) or mcp_exceptions.McpError
+    assert MCPError is installed
+
+
+def test_streamable_http_transport_v1_call_shape():
+    """Test that the 1.x transport receives url, headers, and auth as loose kwargs."""
+    headers = {"Authorization": "Bearer token"}
+    auth = MagicMock()
+
+    with (
+        patch.object(_compat, "MCP_V2", False),
+        patch("mcp.client.streamable_http.streamablehttp_client", create=True) as mock_client,
+    ):
+        result = streamable_http_transport("https://example.com/mcp", headers=headers, auth=auth)
+
+        mock_client.assert_called_once_with(url="https://example.com/mcp", headers=headers, auth=auth)
+        assert result is mock_client.return_value
+
+
+def test_streamable_http_transport_v2_call_shape():
+    """Test that the 2.x transport receives a pre-configured HTTP client carrying the headers and auth."""
+    headers = {"Authorization": "Bearer token"}
+    auth = MagicMock()
+    http_client = MagicMock()
+
+    with (
+        patch.object(_compat, "MCP_V2", True),
+        patch("mcp.client.streamable_http.streamable_http_client", create=True) as mock_client,
+        patch("mcp.client.streamable_http.create_mcp_http_client", create=True, return_value=http_client) as mock_http,
+    ):
+        result = streamable_http_transport("https://example.com/mcp", headers=headers, auth=auth)
+
+        mock_http.assert_called_once_with(headers=headers, auth=auth)
+        mock_client.assert_called_once_with(url="https://example.com/mcp", http_client=http_client)
+        assert result is mock_client.return_value
--- a/strands-py/tests/strands/tools/mcp/test__compat.py
+++ b/strands-py/tests/strands/tools/mcp/test__compat.py
@@ -2,20 +2,28 @@
 
 The branch-specific tests patch real attributes of the installed `mcp`
 package (never `create=True`, which would invent missing names and pass
-against any spelling), so each test runs only on the line whose names exist
-and is skipped on the other.
+against any spelling), so each test is gated on the names it patches
+actually existing on the installed line. Late 1.x releases backport the
+2.x transport names, so the 2.x-branch transport test runs there too with
+the `MCP_V2` flag forced.
 """
 
+import importlib
+from collections.abc import Callable
 from contextlib import asynccontextmanager
 from unittest.mock import MagicMock, patch
 
+import mcp.client.streamable_http as streamable_http_module
 import pytest
 
 from strands.tools.mcp import _compat
 from strands.tools.mcp._compat import MCPError, streamable_http_transport
 
 requires_mcp_v1 = pytest.mark.skipif(_compat.MCP_V2, reason="exercises the mcp 1.x branch")
-requires_mcp_v2 = pytest.mark.skipif(not _compat.MCP_V2, reason="exercises the mcp 2.x branch")
+requires_v2_transport_names = pytest.mark.skipif(
+    not hasattr(streamable_http_module, "streamable_http_client"),
+    reason="installed mcp line lacks the 2.x transport names",
+)
 
 
 def test_installed_line_exposes_expected_transport_names():
@@ -42,6 +50,22 @@ def test_mcp_error_resolves_to_installed_exception():
     assert MCPError is installed
 
 
+def test_get_session_id_callback_falls_back_to_plain_callable(monkeypatch):
+    """Test that `GetSessionIdCallback` degrades to a plain callable alias when the name is absent.
+
+    Reloads `_compat` with the name deleted from the installed transport
+    module to exercise the 2.x fallback on either line, then reloads again so
+    later tests see the module as built against the real environment.
+    """
+    monkeypatch.delattr(streamable_http_module, "GetSessionIdCallback", raising=False)
+    try:
+        reloaded = importlib.reload(_compat)
+        assert reloaded.GetSessionIdCallback == Callable[[], str | None]
+    finally:
+        monkeypatch.undo()
+        importlib.reload(_compat)
+
+
 @requires_mcp_v1
 def test_streamable_http_transport_v1_call_shape():
     """Test that the 1.x transport receives url, headers, and auth as loose kwargs."""
@@ -55,15 +79,16 @@ def test_streamable_http_transport_v1_call_shape():
         assert result is mock_client.return_value
 
 
-@requires_mcp_v2
+@requires_v2_transport_names
 @pytest.mark.asyncio
-async def test_streamable_http_transport_v2_owns_client_lifecycle():
+async def test_streamable_http_transport_v2_owns_client_lifecycle(monkeypatch):
     """Test that the 2.x transport closes the HTTPX client it creates.
 
     Guards https://github.com/strands-agents/harness-sdk/pull/3708: 2.x's
     `streamable_http_client` only closes a client it created itself, so the
     adapter must bind the client's lifetime to the transport's.
     """
+    monkeypatch.setattr(_compat, "MCP_V2", True)
     lifecycle_events = []
     transport_streams = MagicMock()
     auth = MagicMock()
--- a/strands-py/tests/strands/tools/mcp/test__compat.py
+++ b/strands-py/tests/strands/tools/mcp/test__compat.py
@@ -1,16 +1,37 @@
-"""Unit tests for the mcp 1.x/2.x compatibility layer."""
+"""Unit tests for the mcp 1.x/2.x compatibility layer.
 
+The branch-specific tests patch real attributes of the installed `mcp`
+package (never `create=True`, which would invent missing names and pass
+against any spelling), so each test runs only on the line whose names exist
+and is skipped on the other.
+"""
+
+from contextlib import asynccontextmanager
 from unittest.mock import MagicMock, patch
+
+import pytest
 
 from strands.tools.mcp import _compat
 from strands.tools.mcp._compat import MCPError, streamable_http_transport
 
+requires_mcp_v1 = pytest.mark.skipif(_compat.MCP_V2, reason="exercises the mcp 1.x branch")
+requires_mcp_v2 = pytest.mark.skipif(not _compat.MCP_V2, reason="exercises the mcp 2.x branch")
 
-def test_mcp_v2_flag_matches_discover_capability():
-    """Test that the flag reflects whether ClientSession has the 2.x discover API."""
-    from mcp import ClientSession
 
-    assert _compat.MCP_V2 is hasattr(ClientSession, "discover")
+def test_installed_line_exposes_expected_transport_names():
+    """Test that the names each branch imports exist on the line the flag selects.
+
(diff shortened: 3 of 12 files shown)


strands-py/src/strands/tools/mcp/_compat.py
```diff
@@ -0,0 +1,63 @@
...
+
+__all__ = ["MCP_V2", "GetSessionIdCallback", "MCPError", "streamable_http_transport"]
+
+# Feature-probed rather than version-parsed so pre-releases and backports
+# resolve by capability: `ClientSession.discover` is the 2.x replacement for
+# the removed initialize handshake.
+MCP_V2: bool = hasattr(ClientSession, "discover")
+
+try:
+    from mcp.shared.exceptions import MCPError
```
strandly-the-agent: 🔴 **This is what's turning CI red, and it blocks merge regardless of the `mcp` pin.**

`Python / Lint` and `CI Gate` are both failing on `8aed1e85` while `main`'s equivalent job is green. mypy is wired into that job — `pyproject.toml:203-207` `prepare` → `test-lint` → `hatch fmt --linter --check` → `pyproject.toml:139-149` `lint-check = ["ruff check", "mypy ./src"]`. I reproduced it locally against `mcp==1.23.0` (the floor, so what CI's env sees); `ruff check` passes, and the only mcp-related errors are these 5, all in this file:

```
_compat.py:23: error: Module "mcp.shared.exceptions" has no attribute "MCPError"  [attr-defined]
_compat.py:26: error: Unused "type: ignore[attr-defined, no-redef]" comment  [unused-ignore]
_compat.py:35: error: Unused "type: ignore[assignment]" comment  [unused-ignore]
_compat.py:57: error: Module "mcp.client.streamable_http" has no attribute "streamable_http_client"  [attr-defined]
_compat.py:59: error: Returning Any from function declared to return "AbstractAsyncContextManager[Any, bool | None]"  [no-any-return]
```

The merge-base is clean, so all 5 are net-new. Root cause is structural rather than a typo: mypy only ever sees **one** installed `mcp`, so the non-matching branch of each `try/except` is a hard `attr-defined` with no covering ignore (`:23` and `:29` have none; the deferred imports at `:57`/`:61` have none either), and `warn_unused_ignores = true` (`pyproject.toml:220`) then makes the ignores that *do* exist "unused" on the other line. I checked both directions — there is no single mypy invocation, on either supported major, where this file passes clean as written.

No `suggestion` block, because the two halves need different fixes and I'd rather not hand you a guess: the `attr-defined`/`unused-ignore` half wants per-branch `# type: ignore[attr-defined]` plus a per-module `warn_unused_ignores = false` override (there's precedent at `pyproject.toml:227-233`), while the `no-any-return` at `:59` is fixed for free by giving the function a real return type — see my comment on the 2.x branch below.

strands-py/src/strands/tools/mcp/_compat.py
```diff
@@ -0,0 +1,63 @@
...
+        url: The MCP server endpoint URL.
+        headers: Optional HTTP headers to send with each request.
+
+    Returns:
+        An async context manager yielding the transport's read/write streams.
+    """
+    if MCP_V2:
+        from mcp.client.streamable_http import create_mcp_http_client, streamable_http_client
+
+        return streamable_http_client(url=url, http_client=create_mcp_http_client(headers=headers))
```
strandly-the-agent: 🟡 **The HTTPX client built here is never closed — one leaked connection pool per connect.** This is a definitive answer to the open bot thread on this line, and its other question (`headers=None`) is fine: that's the real 2.0.0 default, no bug.

`streamable_http_client` deliberately does *not* take ownership of a client you hand it — from the installed `mcp==2.0.0`:

```python
# mcp/client/streamable_http.py:664
client_provided = http_client is not None
...
async with contextlib.AsyncExitStack() as stack:
    # Only manage client lifecycle if we created it
    if not client_provided:            # :677-678
        await stack.enter_async_context(client)
```

Since `:59` always passes `http_client=`, `client_provided` is always `True`, so the client is never entered *or* closed — and `_compat.py` keeps no reference to close it itself. Verified by monkeypatching the lifecycle methods and driving a full `async with`: `{'aenter': 0, 'aexit': 0, 'aclose': 0}`, `is_closed=False` after exit, 50/50 clients still open across a run. There's no `__del__` on `httpx2.AsyncClient`, so nothing reclaims it. `create_mcp_http_client`'s own docstring says it must be used as a context manager. Note this is a genuine 1.x/2.x asymmetry — 1.x owns its client internally, so the adapter inherits a responsibility that didn't exist before.

**Reachability:** needs the pin relaxed *and* a `"streamable-http"` server entry (`mcp_client.py:1726`, via `MCPClient.load_servers`). 🟡 rather than 🔴 only because `pyproject.toml:37` still pins `<2.0.0` — it becomes 🔴 the day the pin widens, which is this series' stated goal.

What I verified works (plain block, not a `suggestion`, deliberately — see below):

```python
    if MCP_V2:
        from mcp.client.streamable_http import create_mcp_http_client, streamable_http_client

        @asynccontextmanager
        async def _owned_transport() -> AsyncIterator[Any]:
            async with (
                create_mcp_http_client(headers=headers) as http_client,
                streamable_http_client(url=url, http_client=http_client) as streams,
            ):
                yield streams

        return _owned_transport()
```

with `from collections.abc import AsyncIterator` and `asynccontextmanager` added at the top. Confirmed: client now closes (`is_closed=True`), ruff clean, and it also clears the `no-any-return` mypy error from the comment above (5 errors → 4), since the function stops forwarding an untyped `Any`.

**The reason it isn't a committable suggestion:** any ownership fix necessarily wraps the transport, and `test__compat.py:52` asserts `result is mock_client.return_value` — the raw transport object. I tried both a lazy and an eager variant and each fails that test. So the test as written would reject *any* correct fix here, which is why this needs a paired change rather than a one-click commit.

strands-py/tests/strands/tools/mcp/test__compat.py
```diff
@@ -0,0 +1,52 @@
...
+    with (
+        patch.object(_compat, "MCP_V2", True),
+        patch("mcp.client.streamable_http.streamable_http_client", create=True) as mock_client,
+        patch("mcp.client.streamable_http.create_mcp_http_client", create=True, return_value=http_client) as mock_http,
+    ):
+        result = streamable_http_transport("https://example.com/mcp", headers=headers)
+
+        mock_http.assert_called_once_with(headers=headers)
+        mock_client.assert_called_once_with(url="https://example.com/mcp", http_client=http_client)
+        assert result is mock_client.return_value
```
strandly-the-agent: 🟡 **These two branch tests verify the mock rather than the contract, so "4 passed" doesn't mean the 2.x branch works.**

`create=True` on both patches (`:45`, `:46`) tells `mock.patch` to *invent* the attribute when it's missing, so the test passes identically whether or not the real name exists on the installed line. Mutation-proven: I repointed `_compat.py`'s 2.x branch at `totally_bogus_transport` / `totally_bogus_factory` — names that exist on **neither** `mcp` line — and still got **4 passed on both venvs**. A rename or a moved symbol in a future `mcp` release is exactly what this file exists to catch, and it can't.

Worth pairing with the previous comment: `:52`'s `assert result is mock_client.return_value` pins the adapter to returning the raw transport, which is precisely what makes closing the HTTPX client impossible. So this test doesn't just miss the leak — it would fail any fix for it.

Something that would actually bite, since it's checked against the real installed module rather than a mock:

```python
import mcp.client.streamable_http as sh

assert hasattr(sh, "streamable_http_client") is _compat.MCP_V2
assert hasattr(sh, "streamablehttp_client") is not _compat.MCP_V2
```

For the same reason, two nearby items are worth knowing about: `test_mcp_v2_flag_matches_discover_capability` (`:9-13`) restates the implementation and only discriminates when a real `mcp>=2` is installed (which no CI leg does — I confirmed it fails there if `MCP_V2` is hardcoded `False`, and is silent on 1.x); of the four tests here, only `test_mcp_error_resolves_to_installed_exception` has teeth under the current pin. Separately, this PR's own edit to `test_mcp_client.py:1417,1449,1476,1491,1504,1528` swapped `McpError`→`MCPError` but kept the **1.x-only constructor** — 2.x is `MCPError(code, message, data=None)`, not `MCPError(error=…)`, so those 6 sites raise `TypeError` on 2.0.0. (`.error.code`/`.error.data` do resolve on both lines, so `mcp_client.py:963` itself is fine — only construction differs.) Not proposing a mechanical rewrite; the right fix spans six call shapes plus how these branch tests get version-gated.

strandly-the-agent (commented):
Reviewed `8aed1e85` end to end. Short version: **the code that ships is correct — but CI is red because of this PR, and the description claims more than the code delivers.**

The useful thing I could do here that a normal review can't: `mcp==2.0.0` is a real release and installable, so I built venvs on both lines and actually ran the 2.x branch instead of reasoning about it.

| Check | Result |
|---|---|
| `pytest tests/strands/tools/mcp/` on **mcp 1.23.0** (pin floor) | ✅ **217 passed** |
| Same suite on **mcp 2.0.0** | 🔴 **64 failed** / 153 passed |
| `Python / Lint` + `CI Gate` on this head | 🔴 **failure** (`main` is green) |
| `ruff check` locally | ✅ All checks passed |
| `mypy ./src` locally, mcp 1.23.0 | 🔴 **5 errors, all in `_compat.py`** — reproduces the CI failure |
| HTTPX leak on the 2.x branch | 🔴 confirmed by execution (`aclose()` never called) |
| `MCP_V2` probe, `MCPError` alias, transport names vs real 2.0.0 | ✅ all correct |
| OTel 1.x-only gating | ✅ real (`connect` wrapped on 1.23.0, untouched on 2.0.0) |

**The one blocker is mundane:** mypy runs inside `hatch fmt --linter --check` (`pyproject.toml:139-149` → `lint-check = ["ruff check", "mypy ./src"]`), which `hatch run prepare` and the gating lint job both call. It's failing on `_compat.py`. That also means the checklist's "I ran `hatch run prepare`" can't be accurate for this commit.

**The framing question is the one that matters** (in Questions below): is this PR meant to make 2.x *work*, or only *import* so the pin can move later? The body says "imports and runs cleanly on both major lines" and "the pin can later widen to `<3` with no code changes" — I measured both and neither holds today. If the answer is "import-only groundwork," the code is basically right and only the wording needs fixing. I've filed nothing; the 2.x functional gaps are pre-existing and listed below for you to decide on.

3 inline comments. Nothing here re-litigates the resolved threads on #3611.

<details><summary><b>Questions</b> (the framing one is blocking)</summary>

**Blocking**

1. **Is the goal "2.x works" or "2.x imports"?** This decides how everything else is tiered. What ships makes `strands.tools.mcp` *importable* on 2.x and consolidates the version probe — both real and both correct. But `MCPClient` is not functional on 2.x: `start()` raises unconditionally, ~15 camelCase field accesses break, and `tasks_config=` raises `ImportError` at construction. If that's intentional staging, could the description say so? As written the next reader will believe the pin can move.

2. **Merge order** — this is stacked on #3611, which is still open, and includes its commit. GitHub's Files-changed tab merges both into one diff, so anyone approving from that tab is implicitly approving #3611's content without it clearing its own gate. Is bottom-up (land #3611 first, let this shrink to its own commit) the intent?

**Non-blocking, but worth settling before more code branches on the pattern**

3. One boolean is standing in for at least six independently-probeable axes (property-vs-method, tasks surface, transport takes-a-client, transport yields-session-id, field casing, task types importable) — I verified each discriminates on its own. `_compat.py` already uses per-name `try/except` for `MCPError`/`GetSessionIdCallback` but the global flag for the transport. Since the body says lifecycle/tasks/native-OTel will all branch on this same flag, would per-capability constants age better than one flag a future 3.x can't name correctly? (`hasattr(ClientSession, "discover")` reads as "≥2" while `MCP_V2` reads as "==2".)

4. `MCPTransport` is public (`mcp_types.py:47`, exported in `__all__`) and quietly gets a second meaning here — on 2.x `GetSessionIdCallback` degrades to a structural `Callable[[], str | None]` and the 3-tuple arm is simply dead. Worth deciding what `MCPTransport` should *mean* on 2.x before more code depends on the current shape? Relatedly, the PR carries no `api/needs-review` / `api/review-complete` label — `check-api-review-label` is green only because it skips when no label is present.

5. Should HTTPX ownership be a written rule for future adapters here — *the adapter owns what it creates, never what it's given* — rather than a per-call-site decision?

</details>

<details><summary><b>Appendix — non-blocking (6)</b></summary>

- ⚪ `_compat.py:17-19` — the comment calls `ClientSession.discover` "the 2.x replacement for the removed initialize handshake," but `hasattr(ClientSession, "initialize")` is `True` on **both** installed lines. The spec removed the handshake; the Python client hasn't. Wording only.
- ⚪ `_compat.py:28-35` catches only `ImportError` and its `except` branch doesn't re-import from `mcp`, so an unrelated import failure in `mcp.client.streamable_http` gets swallowed and you get a fake alias plus a worse stack trace later. The `MCPError` sibling at `:22-26` doesn't have this gap because its `except` re-imports from the same module. No caller triggers it today.
- ⚪ Anything that adds a `discover` attribute to `ClientSession` on a real 1.x (subclass, backport, test double) flips `MCP_V2` and sends the transport down the 2.x path. Contrived — same root cause as Question 3.
- ⚪ `test_mcp_instrumentation.py:426` asserts `register_post_import_hook` is called, which the new gating correctly skips on 2.x, so the suite can't be green there. **Inherited from #3611's commit, not this one.**
- ⚪ No CI leg installs `mcp>=2`, so every 2.x branch is mock-only. Smallest thing that would work today and passes right now: a `@pytest.mark.skipif(not MCP_V2, …)` import-smoke test plus one single-Python job doing `pip install "mcp>=2.0.0,<3.0.0" --no-deps` and running just `test__compat.py`. That alone would have caught the vacuous-mock problem in comment 3.
- ⚪ Codecov reports 3 uncovered lines in `_compat.py` — exactly the 2.x/`except ImportError` branches that need a real `mcp>=2` install to execute.

</details>

<details><summary><b>Pre-existing — needs an issue, not a fix here (4)</b></summary>

I verified with `git diff 898cc0ef 8aed1e85` that this PR touches none of these lines. They're unchanged from `main` and only become reachable if the pin widens — so they are **not** regressions here, and I'm not treating them as blockers. Flagging because they're what stands between this PR and its stated goal. Happy to file them if you want.

1. **`MCPClient.start()` raises on 2.x** — `mcp_client.py:1063` calls `session.get_server_capabilities()`; 2.x made it a `server_capabilities` property. Unconditional bootstrap, every transport. Repro against a real in-process server: `AttributeError: 'ClientSession' object has no attribute 'get_server_capabilities'` → `MCPClientInitializationError`.
2. **~15 camelCase→snake_case field sites** — worst is `mcp_client.py:1007` `.isError`, where the `AttributeError` is swallowed into an error `ToolResult`, so **every successful 2.x tool call reports as failed** with the raw `AttributeError` as the tool's output to the model. Wrong-answer shaped, not crash shaped. Also `mcp_agent_tool.py:77,82,83` and `mcp_client.py:622,1015,1149,1179-1199`.
3. **`tasks_config=` raises a bare `ImportError` at construction on 2.x** (`mcp_client.py:299`) and isn't aliasable — `ClientSession.experimental` doesn't exist on 2.0.0, so `:1259,1541,1584,1625` have no 2.x surface. Needs a design call (clear "unsupported on 2.x" error vs. port to the extension), and #1659 already defers tasks pending the tasks-extension GA.
4. **`tests_integ/mcp/*` and the separate `strands-mcp` package are 2.x-broken** (`FastMCP`, `mcp.server.experimental.task_context`; `strands-mcp` has its own `<2.0.0` pin). Out of scope, worth tracking so it isn't rediscovered piecemeal.

</details>

---
*Automated review by `strandly-the-agent` — ran on both `mcp` lines with real installs. Solid work for a human to approve, not a gate; push back where I'm wrong.*

strandly-the-agent (commented):
**Good to approve on the code.** All three findings from my last review are fixed, and I re-verified each by execution rather than taking the diff's word for it. One non-code ask below.

Reviewed the delta `8aed1e85..b00237b` (4 commits). Not re-raising the three now-resolved threads.

| My finding at `8aed1e85` | Status |
|---|---|
| 🔴 mypy red → `Python / Lint` + `CI Gate` red | ✅ **Fixed.** `Success: no issues found` on **both** the 1.23.0 pin floor **and** 2.0.0 |
| 🟡 HTTPX client leaked once per connect | ✅ **Fixed.** Verified against the **real** `httpx2.AsyncClient`, not just the test's fakes |
| 🟡 `create=True` mocks verified the mock, not the contract | ✅ **Fixed.** `create=True` is gone; skipifs gate on real attribute existence |

Two of my open questions also resolved without needing to re-ask: **#3611 merged first** (the stack order I suggested), and the 2.x CI leg now exists as `Python / MCP 2.x Compat` — green.

**The one ask, and it's wording not code:** the description still says the package "imports **and runs cleanly** on both major lines" and that "the pin can later widen to `<3` **with no code changes**." On real `mcp==2.0.0` the MCP suite is **58 failed / 159 passed** (improved from 64 at my last review — the new version-aware `make_mcp_error()` helper fixed the `MCPError(error=…)` landmine). *Imports* cleanly is true, and is a genuinely useful milestone worth claiming; *runs* cleanly and *no code changes* aren't. Suggested edit in the collapsed section.

<details><summary><b>How I verified each fix</b></summary>

**mypy** (`_compat.py:30,33,36,64,83,85` + `pyproject.toml:240-244`) — the fix is the structural one: branch-level `# type: ignore[attr-defined]`, a `no-any-return` ignore on the 1.x return, and a per-module `warn_unused_ignores = false`. I ran it on both installed lines, since the earlier bot pass verified 1.29.0 and only *asserted* the floor was covered:
```
mcp 1.23.0 (pin floor): Success: no issues found in 1 source file
mcp 2.0.0             : Success: no issues found in 1 source file
```

**HTTPX leak** (`_compat.py:63-81`) — the shipped test proves ordering with fake context managers, so it does not prove the fix works against the real package. I drove `streamable_http_transport` with the real `create_mcp_http_client` and the real client class, instrumenting its lifecycle:
```
client class: httpx2.AsyncClient
AFTER exit -> lifecycle calls: {'aenter': 1, 'aexit': 1, 'aclose': 0}
clients created: 1 | is_closed after: [True]
RESULT: no leak - every created client is closed
```
(`aclose` is 0 because `__aexit__` closes it — `is_closed=True` is the thing that matters. Pre-fix this was `{'aenter': 0, 'aexit': 0, 'aclose': 0}` with `is_closed=False`.)

**Tests** — clean mirror-image behaviour, 4 passed / 1 skipped on each line, with the opposite-line test skipping each time. The leak guard genuinely *runs* rather than skipping: `test_streamable_http_transport_v2_owns_client_lifecycle` PASSED on 2.0.0, and it also runs on the 1.x matrix because 1.29.x backports `streamable_http_client` (I checked: 1.29.0 has all four transport names, `MCP_V2` still `False`). Full MCP suite on 1.23.0: **217 passed, 1 skipped**.

**CI** — green on `b00237b`: `Python / Lint`, `CI Gate`, `Python / MCP 2.x Compat`, all 11 unit-test matrix legs. I also ran the new job's exact two commands locally against real 2.0.0; both succeed.

</details>

<details><summary><b>Suggested description edit</b></summary>

> ~~"so the package imports and runs cleanly on both major lines"~~
> → "so the package **imports** cleanly on both major lines; running an `MCPClient` end-to-end against a 2.x server is not yet supported (tracked as follow-up)"

> ~~"and the pin can later widen to `<3` with no code changes"~~
> → "and the pin can widen to `<3` once `start()`, the camelCase field accesses, and the tasks paths are updated for 2.x (tracked separately)"

The Testing section's "`import strands` succeeds (fails at import today)" has the same top-level-vs-submodule conflation — on base it's `import strands.tools.mcp` that fails, not `import strands`. Minor, same fix.

</details>

<details><summary><b>Notes on the new delta — all non-blocking (3)</b></summary>

- ⚪ The new `mcp-v2-compat` job is well-built — deliberate `2.0.*` pin, `permissions: contents: read`, `persist-credentials: false`, `timeout-minutes: 10`, and I confirmed it's covered by `ci-gate` transitively (the `python` reusable-workflow caller fails if any inner job fails; `alls-green`'s `allowed-skips` only tolerates *skipped*, not failed). Action pinning (`@v7`, `@v7.0.0`) matches its siblings in the same file, so no inconsistency. Two trivia: `moto` in the pip install isn't needed for `test__compat.py`, and scoping to just that file is the right call given the 58 remaining 2.x failures — worth a comment saying so if it isn't already obvious to the next reader.
- ⚪ On the declared pin **floor** (1.23.0) the leak-guard test skips, because 1.23.0 lacks `streamable_http_client` entirely. It runs on 1.29.x and 2.0.0, which is what CI actually resolves, so real coverage is fine — just noting the floor itself is unguarded for that one behaviour.
- ⚪ `warn_unused_ignores = false` is scoped per-module, which is the right precedent (`pyproject.toml:227-233`), but it's module-wide: a genuinely stale ignore added to `_compat.py` later won't be flagged. Fine tradeoff for this file, worth knowing.

</details>

<details><summary><b>Unchanged from last time — pre-existing, not this PR's job</b></summary>

The four 2.x functional gaps I listed last review are unchanged and still latent behind the `<2.0.0` pin: `MCPClient.start()`'s `get_server_capabilities()`, the ~15 camelCase field sites (worst `mcp_client.py` `.isError`), `tasks_config=` having no 2.x surface, and `tests_integ`/`strands-mcp`. They're the bulk of the 58 failures. Still happy to file these as issues if useful — they're what stands between this PR and a pin widen.

</details>

---
*Scope note, in fairness: I ran the verification above myself in warm context since it's checking my own findings. I also kicked off an independent adversarial pass to attack the fixes for newly-introduced bugs (mutation-testing the new tests and CI job, and the `importlib.reload` + xdist ordering risk in `test__compat.py:53-66`); it hadn't finished when I posted, so it is **not** folded in. Nothing in it would change the ledger above — it could only add a new finding. I'll follow up only if it surfaces something real.*

strands-py/tests/strands/tools/mcp/test__compat.py
strandly-the-agent: ✅ **Fixed on `b00237b` — resolving from my side.**

`create=True` is gone from both branch tests, and the rewrite goes further than what I suggested: the module docstring now states the rule explicitly ("never `create=True`, which would invent missing names and pass against any spelling"), and `test_installed_line_exposes_expected_transport_names` asserts directionally against the real installed module, which is the check I proposed. The `requires_mcp_v1` / `requires_v2_transport_names` skipifs gate on real attribute existence rather than inventing it.

I also confirmed the specific thing I warned about here — that `:52`'s `assert result is mock_client.return_value` would reject any ownership fix. That assertion is now scoped to the 1.x call-shape test only, and the 2.x path gets `test_streamable_http_transport_v2_owns_client_lifecycle`, which asserts the full teardown ordering as one equality (`client_enter, transport_enter, transport_exit, client_exit`). That's a stronger contract than the identity check it replaced, and it's what let the leak fix land.

Verified by execution on both lines: 4 passed / 1 skipped each, mirror-image skips, and the lifecycle test genuinely *runs* on real `mcp==2.0.0` in the new `Python / MCP 2.x Compat` job rather than skipping. The related `MCPError(error=…)` 1.x-only constructor I flagged in this same comment is fixed too, via the version-aware `make_mcp_error()` helper in `conftest.py:13-17`.

One small note, not a request: on the declared pin floor (1.23.0) the lifecycle test skips, since 1.23.0 lacks `streamable_http_client` entirely. CI resolves 1.29.x where it's backported, so actual coverage is fine.

arielnabavian (dismissed):

strands-py/src/strands/tools/mcp/_compat.py
```diff
@@ -0,0 +1,89 @@
...
+from typing import Any
+
+import httpx
+from mcp import ClientSession
+
+__all__ = ["MCP_V2", "GetSessionIdCallback", "MCPError", "streamable_http_transport"]
+
+# Feature-probed rather than version-parsed so pre-releases and backports
+# resolve by capability: `ClientSession.discover` is the 2.x replacement for
+# the removed initialize handshake.
```
Unshure, line 27: What does this mean by "resolve by capability"? Shouldnt we realy on a major version parsing to understand which api shape we should expect?

strands-py/src/strands/tools/mcp/_compat.py
```diff
@@ -0,0 +1,89 @@
...
+MCP_V2: bool = hasattr(ClientSession, "discover")
+
+try:
```
Unshure, line 30: Could we check major version, and import accordingly, instead of doing this try, catch?

strands-py/src/strands/tools/mcp/_compat.py
```diff
@@ -0,0 +1,89 @@
...
+except ImportError:
+    # mcp 2.x removed protocol sessions, so its transports never yield a
+    # session-id callback; the alias survives only to type 1.x transports.
+    from collections.abc import Callable
+
+    GetSessionIdCallback = Callable[[], str | None]  # type: ignore[misc, assignment]
+
+
+def streamable_http_transport(
+    url: str, headers: dict[str, Any] | None = None, auth: httpx.Auth | None = None
```
Unshure: Does this need to be any type? Could it be union?

strands-py/src/strands/tools/mcp/_compat.py
poshinchen: The mcp package doesn't expose a `__version__`, and version numbers turned out to be not reliable predictor of the actual API surface (1.29.x backported the 2.x streamable_http_client), so checking for the name we need is both simpler and can't mispredict.

strands-py/src/strands/tools/mcp/_compat.py
poshinchen: same as above!

strands-py/src/strands/tools/mcp/_compat.py
poshinchen: changed!

strands-py/src/strands/tools/mcp/_compat.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+
+        # `streamable_http_client` closes an HTTPX client only when it created
+        # it (`client_provided` check in mcp 2.x), so a caller-provided client
+        # must be closed by the caller: enter both context managers together
+        # so the client's lifetime is bound to the transport's.
+        @asynccontextmanager
+        async def _owned_client_transport() -> AsyncIterator[Any]:
+            async with (
+                create_mcp_http_client(headers=headers, auth=auth) as http_client,
```
strandly-the-agent, line 80: 🟡 **The `auth` threading breaks the "mypy clean on both mcp lines" property this PR established — and the type error is telling the truth about runtime.**

On a 2.x env, mypy now reports exactly one error in this file:

```
_compat.py:80: error: Argument "auth" to "create_mcp_http_client" has incompatible type
"httpx._auth.Auth | None"; expected "httpx2._auth.Auth | None"  [arg-type]
```

It's real, not stub noise — verified on `mcp==2.0.1`: `httpx2.AsyncClient` rejects a v1 `httpx.Auth` instance at construction (`TypeError: Invalid "auth" argument`). Reached via `MCPClient(url=..., auth_provider=<custom httpx.Auth>)` once the pin widens; the built-in `auth={client_id, client_secret}` path is unaffected on both lines, because mcp's own `ClientCredentialsOAuthProvider` follows the installed line (it subclasses `httpx2.Auth` on 2.x — verified). CI stays green because its mypy runs on the 1.x env — the mirror image of the failure geometry we fixed here three weeks ago.

```suggestion
                create_mcp_http_client(headers=headers, auth=auth) as http_client,  # type: ignore[arg-type]
```

Suggestion: verified on this head — with it, `mypy` is `Success` on **both** 1.23.0 and 2.0.1, and ruff check/format stay clean (the per-module `warn_unused_ignores` override already absorbs the side unused on 1.x). Cost: like the file's other ignores it silences the checker rather than resolving the union, so the custom-`auth_provider`-on-2.x limitation stays — that part can't be fixed here (no object satisfies both `httpx.Auth` and `httpx2.Auth`), so it belongs on the pin-widen checklist, plus optionally one docstring line on `auth_provider` noting that a 2.x install will need an `httpx2`-compatible auth object.

strandly-the-agent (commented):
Re-reviewed at `53aa5a1`, including the comment threads. **Still good to approve — one new 🟡 in the delta, with a one-line verified fix inline.**

The real delta since the head I last verified (`b00237b`) is threading main's new OAuth support through the compat layer: `streamable_http_transport` gained `auth: httpx.Auth | None`, passed to 1.x as a kwarg and on 2.x into `create_mcp_http_client`. I verified it end to end: floor-safe (1.23.0's `streamablehttp_client` accepts `auth` — checked its real signature), both call-shape tests updated, and mutation-proven (dropping `auth` from either branch fails the suite). The rest is rebase.

**The one new finding** (inline at `_compat.py:80`): the auth threading regressed the *"type-checks on both mcp lines"* property this PR's own fix commit established — mypy on a 2.x env now reports one `arg-type` error there, and it points at a real runtime fact I proved on `mcp==2.0.1`: `httpx2.AsyncClient` **rejects a v1 `httpx.Auth` object at construction** (`TypeError: Invalid "auth" argument`). The built-in `auth={client_id, client_secret}` path is fine on both lines — mcp's own `ClientCredentialsOAuthProvider` follows the installed line (it's an `httpx2.Auth` subclass on 2.x, verified). Only a *custom* v1 `httpx.Auth` via `auth_provider=` breaks, only on the 2.x branch, and only once the pin widens — so 🟡, not a blocker; CI can't see it because its mypy runs on the 1.x env.

**On `Unshure`'s open thread** (capability probe vs version parsing): the author's answer is correct, and I have measured data backing it — name matrix from real installs in the collapsed section. A major-version parse would have needed a special case for the 1.29.x backports; the per-name probe can't mispredict by construction.

**My standing description ask from both prior reviews: addressed.** "Runs cleanly" and "pin can widen with no code changes" are gone; the new text scopes the claim honestly and documents the CI job. Thanks for that.

<details><summary><b>Evidence ledger (all measured on this head)</b></summary>

| Check | Result |
|---|---|
| `test__compat.py` on mcp **1.23.0** (floor) / **2.0.1** | 4 passed, 1 skipped — each line, mirror-image skips |
| Full `tests/strands/tools/mcp/` on 1.23.0 | **256 passed, 1 skipped** |
| Same on 2.0.1 | 63 failed / 193 passed — the known pre-existing 2.x gaps (`start()`, camelCase fields, tasks), unchanged, latent behind the pin |
| mypy `_compat.py` on 1.23.0 | ✅ Success |
| mypy `_compat.py` on **2.0.1** | 🟡 1 error: `:80` `arg-type` (`httpx._auth.Auth \| None` vs expected `httpx2._auth.Auth \| None`) |
| Runtime probe on 2.0.1 | `create_mcp_http_client(auth=<v1 httpx.Auth>)` → `TypeError` at construction; `auth=<ClientCredentialsOAuthProvider>` → OK |
| Mutation: drop `auth` from 2.x branch / from 1.x branch | lifecycle test fails / v1 call-shape test fails — both mutants killed |
| Prior findings (leak guard, no-`create=True` tests, `make_mcp_error`) | intact at this head |
| `import httpx` at `_compat.py:20` | safe on any resolve — `httpx>=0.28.1,<1.0.0` is a *direct* strands dep (`pyproject.toml:35`), not just transitive via mcp 1.x |
| CI on `53aa5a1` | `Python / Lint`, `CI Gate`, `Python / MCP 2.x Compat`, all unit matrices ✅ |

**Name matrix for the capability-probe thread** (real installs):

| | `streamablehttp_client` | `streamable_http_client` | `create_mcp_http_client` | `ClientSession.discover` |
|---|---|---|---|---|
| mcp 1.23.0 (floor) | ✅ | ❌ | ❌ | ❌ |
| mcp 1.29.x | ✅ | ✅ (backport) | ✅ (backport) | ❌ |
| mcp 2.0.x | ❌ | ✅ | ✅ | ✅ |

</details>

<details><summary><b>Two smaller notes</b></summary>

- **Integ CI**: `Run integration tests - Python 3.10 - all` is red on this head with an async-teardown signature (`Event loop is closed` ×7, no MCP-specific annotation), and `main`'s own recent integ runs (Aug 18–19) also show failures. Likely unrelated to this PR — but I couldn't read the job log to fully attribute it, so worth confirming it's the known flake before merge; integ isn't part of `CI Gate`.
- **Self-correction**: in my Aug 11 review I called `moto` in the `mcp-v2-compat` job's pip install unnecessary. It's now load-bearing — a parent conftest imports it, and the compat tests fail to *collect* without it. Good call not taking that nit.

</details>

---
*Follow-up re-review by `strandly-the-agent` — delta-scoped, run against real installs of mcp 1.23.0 / 1.29.x / 2.0.1 on this exact head.*

.github/workflows/python-test-lint.yml
```diff
@@ -88,6 +88,49 @@ jobs:
         uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f # v7
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+  mcp-v2-compat:
```
pgrayy, line 91: For clarity, I would recommend naming this `unit-test-mcp-v2`.

pgrayy (approved):

.github/workflows/python-test-lint.yml
poshinchen: Will do it in the 2nd PR!

PR #3722, opened 2026-08-08
fix(openai): classify Bedrock Mantle context-overflow errors

### LINKED ISSUE #3718: [BUG] classify_openai_error() doesn't recognize Bedrock Mantle's context-overflow message, so ContextWindowOverflowException / reduce_context() never fires
### Checks

### SDK Language

Python

### Strands Version

1.50.2

### Language Runtime Version

Python 3.14.7

### Operating System

MacOS 26.5.2

### Installation Method

pip

### Steps to Reproduce

Using Bedrock Mantle's OpenAI-compatible endpoint (`bedrock_mantle_config` on `OpenAIResponsesModel`), send a
prompt that exceeds the model's real context window for `openai.gpt-5.6-terra` (confirmed live limit:
1,050,000 tokens).

Minimal reproduction:

```python
from strands import Agent
from strands.models import BedrockMantleConfig
import boto3

model = OpenAIResponsesModel(
    model_id="openai.gpt-5.6-terra",
    bedrock_mantle_config=BedrockMantleConfig(
        boto_session=boto3.Session(), region="us-east-1"
    ),
    params={"max_output_tokens": 16},
)
agent = Agent(model=model, tools=[], callback_handler=None)

# ~8.6M chars / ~1.2M tokens -- well over Terra's 1,050,000-token window
big_input = "quantum entangled photon lattice resonance " * 200_000
agent(big_input)
```

1. Run the snippet above against a live Bedrock Mantle endpoint with valid AWS credentials.
2. Mantle returns HTTP 400 with body:
   `{"error": {"code": "validation_error", "message": "prompt tokens (1200008) exceed model maximum
   (1050000) for openai.gpt-5.6-terra", "param": null, "type": "invalid_request_error"}}`
3. Observe the exception that propagates out of `agent(...)`.

Root cause, confirmed by reading the installed package
(`strands/models/_openai_errors.py`):

```python
_CONTEXT_WINDOW_OVERFLOW_PATTERNS = (
    "maximum context length",
    "context_length_exceeded",
    "too many tokens",
    "context length",
    "input is too long for requested model",
    "input length and `max_tokens` exceed context limit",
    "too many total text bytes",
    "exceed customer model maximum",   # <-- has "customer"
    "exceeds the max_model_len",
)
```

Mantle's real message is `"...exceed model maximum..."` — no `"customer"` in it, and its error `code` is
`"validation_error"`, not `"context_length_exceeded"`. No pattern in the list is a substring of the real
message, so `classify_openai_error()` returns `None` for every genuine Terra context overflow.

In `openai_responses.py`'s `stream()`:

```python
except (openai.APIError, _OpenAIResponsesStreamError) as error:
    error_kind = classify_openai_error(error)
    if error_kind == "throttling":
        ...
    if error_kind == "context_overflow":
        raise ContextWindowOverflowException(str(error)) from error
    raise  # <-- this is what actually happens for Terra
```

Since `error_kind` is `None`, the original error is re-raised unclassified instead of being wrapped in
`ContextWindowOverflowException`.

`Agent._execute_event_loop_cycle`'s only auto-recovery path is:

```python
except ContextWindowOverflowException as e:
    self.conversation_manager.reduce_context(self, e=e)
    ...  # retry
```

Because the exception is never `ContextWindowOverflowException`, `reduce_context()` (and the whole
retry-after-trim flow) never runs — even though `agent.conversation_manager` is correctly a
`SlidingWindowConversationManager` (the default; we never override it). The conversation-manager wiring
is correct and present; only the error classification is missing the Mantle-specific string.

### Expected Behavior

A prompt that exceeds a Bedrock Mantle model's real context window should be classified as
`"context_overflow"` by `classify_openai_error()`, causing `ContextWindowOverflowException` to be raised,
which triggers `conversation_manager.reduce_context()` and an automatic retry with a trimmed conversation
— the same recovery behavior that already works correctly for providers whose error text matches an
existing pattern (e.g. plain OpenAI's `"maximum context length"`).

### Actual Behavior

The raw, unclassified error (`_OpenAIResponsesStreamError` / `openai.BadRequestError`, message
`"prompt tokens (N) exceed model maximum (M) for <model_id>"`) propagates straight out of `agent(...)`.
No `ContextWindowOverflowException` is raised, `reduce_context()` never runs, and the turn hard-fails with
no auto-trim-and-retry — even though a working `SlidingWindowConversationManager` is attached to the
agent and would recover correctly if the exception were classified.

In our application, this is caught by a broad top-level exception handler, so the end user sees a generic
error rather than a crash — but the intended graceful-degradation behavior (trim context, retry) never
engages, and the failure is indistinguishable from any other unrelated error.

### Additional Context

- Confirmed live against Bedrock Mantle in `us-east-1` on 2026-08-07 (Mantle's advertised context window
  for `openai.gpt-5.6-terra` is 1,050,000 tokens as of the 2026-08-03 long-context expansion).
- Verified the installed package's `_CONTEXT_WINDOW_OVERFLOW_PATTERNS` list directly
  (`site-packages/strands/models/_openai_errors.py`) matches the current `main` branch of
  `strands-agents/harness-sdk` — this is not fixed on a newer unreleased version as of filing.
- Also reproduced the same gap via `client.responses.create()` directly against Mantle with the plain
  `openai` Python SDK (bypassing Strands entirely) to confirm the message text originates from Mantle
  itself, not from any Strands-side wrapping.
- We reach `OpenAIResponsesModel` via Bedrock Mantle's `bedrock_mantle_config` (not plain OpenAI), so this
  is specifically about Mantle's error phrasing, not a general OpenAI regression.
- The identical gap likely also affects `OpenAIModel`'s Chat-Completions-based `stream()`/`structured_output()` for
  non-`gpt-5.*` Mantle models, since both call the same shared `classify_openai_error()`.

### Possible Solution

Add Mantle's exact phrasing to `_CONTEXT_WINDOW_OVERFLOW_PATTERNS` in
`strands/models/_openai_errors.py`, e.g.:

```python
_CONTEXT_WINDOW_OVERFLOW_PATTERNS = (
    ...,
    "exceed customer model maximum",
    "exceed model maximum",  # Bedrock Mantle's phrasing (no "customer")
    ...,
)
```

Since `"exceed customer model maximum"` is itself a superset-match risk, a more robust long-term fix might
be a single regex like `r"exceed(?:s|ed)?(?: customer)? model maximum"` or matching on `"model maximum"`
alone, to absorb minor future wording variants from any OpenAI-compatible provider without needing another
patch release per provider.

### Related Issues

_No response_

teddytennant:
## Description

`classify_openai_error()` matched Bedrock Mantle's context-overflow message against a list of literal substrings that only contained the OpenAI phrasing `"exceed customer model maximum"`. Mantle drops the `customer` qualifier:

```
prompt tokens (1200008) exceed model maximum (1050000) for openai.gpt-5.6-terra
```

and it reports `code="validation_error"`, so neither the code check nor any substring matched. The classifier returned `None`, `OpenAIResponsesModel.stream()` re-raised the raw vendor error instead of a `ContextWindowOverflowException`, and `Agent._execute_event_loop_cycle`'s `reduce_context()` recovery path never ran. A conversation that the default `SlidingWindowConversationManager` could have trimmed and retried hard-failed instead.

Reproducer on `main`:

```python
from strands.models._openai_errors import classify_openai_error

class FakeError(Exception):
    pass

msg = "prompt tokens (1200008) exceed model maximum (1050000) for openai.gpt-5.6-terra"
print(classify_openai_error(FakeError(msg)))  # None
```

Rather than appending one more literal, the `"model maximum"` family is now matched with a regex that absorbs the qualifier and the verb inflection (`exceed` / `exceeds` / `exceeded`, with or without `customer`). This is the phrase most likely to keep drifting between OpenAI-compatible providers, and a per-provider literal means a patch release every time the wording moves. The other literals are untouched and still matched by plain substring containment, so no existing classification changes.

The regex deliberately still requires the `exceed…` verb — matching on `"model maximum"` alone would classify unrelated validation errors (for example `"model maximum output tokens must be a positive integer"`) as context overflow and trigger a spurious context trim on a failure that trimming cannot fix. There is a test pinning that.

## Related Issues

Fixes https://github.com/strands-agents/harness-sdk/issues/3718

## Documentation PR

No docs change needed — `_openai_errors.py` is internal and the pattern list is not documented.

## Type of Change

Bug fix

## Testing

`tests/strands/models/test_openai_errors.py` gains the real Mantle message to the overflow parametrization, and `test_classify_openai_error_unknown` is parametrized with two adjacent-but-unrelated provider messages so an over-greedy pattern is caught. Verified fail-before/pass-after against the stashed source change (the new case fails with `assert None == 'context_overflow'` without the fix), and mutated the regex both ways — widening it to `model maximum` fails the negative case, narrowing it to `exceed model maximum` fails the existing `exceed customer model maximum` case.

The full unit suite passes (4846 tests), and `hatch fmt --linter --check` is clean (`ruff check` plus `mypy ./src`, no issues in 234 source files).

- [ ] I ran `hatch run prepare` — ran `hatch fmt --formatter --check`, `hatch fmt --linter --check`, and `hatch test`; skipped the `--all` matrix since I only have one interpreter locally. `hatch fmt --formatter --check` flags 13 files, none of them touched here (pre-existing on `main` with the ruff version that resolves locally).

## Deliberately not changed

`strands-ts/src/models/openai/errors.ts` carries the same literal list and the same gap. Keeping this PR to one sub-project per the contribution guidelines; happy to open the TypeScript counterpart separately.

## Checklist

----

### DIFF
--- a/strands-py/tests/strands/models/test_openai_errors.py
+++ b/strands-py/tests/strands/models/test_openai_errors.py
@@ -22,6 +22,9 @@ class OpenAICompatibleError(Exception):
         "too many total text bytes",
         "exceed customer model maximum",
         "the engine prompt length exceeds the max_model_len",
+        # Bedrock Mantle omits the "customer" qualifier
+        # (https://github.com/strands-agents/harness-sdk/issues/3718)
+        "prompt tokens (1200008) exceed model maximum (1050000) for openai.gpt-5.6-terra",
     ],
 )
 def test_classify_openai_error_context_overflow_message(message):
@@ -61,8 +64,16 @@ def test_classify_openai_error_throttling_precedes_context_overflow():
     assert classify_openai_error(error) == "throttling"
 
 
-def test_classify_openai_error_unknown():
-    assert classify_openai_error(OpenAICompatibleError("unrelated failure")) is None
+@pytest.mark.parametrize(
+    "message",
+    [
+        "unrelated failure",
+        "model maximum output tokens must be a positive integer",
+        "requested model openai.gpt-5.6-terra is not available in this region",
+    ],
+)
+def test_classify_openai_error_unknown(message):
+    assert classify_openai_error(OpenAICompatibleError(message)) is None
 
 
 def test_classify_openai_error_non_string_code():
--- a/strands-py/tests/strands/models/test_openai_errors.py
+++ b/strands-py/tests/strands/models/test_openai_errors.py
@@ -22,8 +22,6 @@ class OpenAICompatibleError(Exception):
         "too many total text bytes",
         "exceed customer model maximum",
         "the engine prompt length exceeds the max_model_len",
-        # Bedrock Mantle omits the "customer" qualifier
-        # (https://github.com/strands-agents/harness-sdk/issues/3718)
         "prompt tokens (1200008) exceed model maximum (1050000) for openai.gpt-5.6-terra",
     ],
 )
--- a/strands-py/src/strands/models/_openai_errors.py
+++ b/strands-py/src/strands/models/_openai_errors.py
@@ -1,5 +1,6 @@
 """Shared error classification for OpenAI model providers."""
 
+import re
 from typing import Literal
 
 OpenAIErrorKind = Literal["context_overflow", "throttling"]
@@ -14,9 +15,11 @@ _CONTEXT_WINDOW_OVERFLOW_PATTERNS = (
     "input is too long for requested model",
     "input length and `max_tokens` exceed context limit",
     "too many total text bytes",
-    "exceed customer model maximum",
     "exceeds the max_model_len",
 )
+# Providers phrase the token-budget overflow around "model maximum" differently: OpenAI says
+# "exceed customer model maximum", Bedrock Mantle drops the qualifier ("exceed model maximum").
+_CONTEXT_WINDOW_OVERFLOW_REGEX = re.compile(r"exceed(?:s|ed)?(?: customer)? model maximum")
 _RATE_LIMIT_PATTERNS = ("rate_limit_exceeded", "rate limit", "too many requests")
 
 
@@ -33,7 +36,11 @@ def classify_openai_error(error: BaseException) -> OpenAIErrorKind | None:
     ):
         return "throttling"
 
-    if code == "context_length_exceeded" or any(pattern in message for pattern in _CONTEXT_WINDOW_OVERFLOW_PATTERNS):
+    if (
+        code == "context_length_exceeded"
+        or any(pattern in message for pattern in _CONTEXT_WINDOW_OVERFLOW_PATTERNS)
+        or _CONTEXT_WINDOW_OVERFLOW_REGEX.search(message)
+    ):
         return "context_overflow"
 
     return None


strands-py/tests/strands/models/test_openai_errors.py
```diff
@@ -22,6 +22,9 @@ def __init__(self, message: str, *, code: object = None, status_code: int | None
         "too many total text bytes",
         "exceed customer model maximum",
         "the engine prompt length exceeds the max_model_len",
+        # Bedrock Mantle omits the "customer" qualifier
+        # (https://github.com/strands-agents/harness-sdk/issues/3718)
```
liramon2: Nit: this comment can be removed. The implementation already describes the issue.

teddytennant: Removed, thanks. Pushed in 5a79d7f.

liramon2 (approved):

PR #3725, opened 2026-08-09
fix(bidi): emit gemini usage metadata alongside content events

mehtarac:
## Description

Gemini's Live API sends token usage in a usageMetadata field that sits outside the messageType union. Per the Gemini Live Api reference: 
> "Server messages may have a usageMetadata field but will otherwise include exactly one of the other fields." So usage can accompany audio, a tool call, or aresumption update.

`BidiGeminiLiveModel._convert_gemini_live_event`'s implementation was such that any message with both content and usage lost the usage. The loss is silent and cumulative:
  `_BidiAgentLoop._run_model` accumulates session token counters solely from BidiUsageEvent, so every dropped event undercounts the session with no error. This restructures the converter to accumulate into a list. Control messages stay exclusive content messages append, and usage is evaluated unconditionally at the end. 
  
  The list-accumulating shape also unblocks emitting multiple events per message, needed for the missing BidiResponseStartEvent / BidiResponseCompleteEvent for google gemini live model provider change left to a follow-up.
## Type of Change

Bug fix

## Testing

## Checklist

----

### DIFF
--- a/strands-py/src/strands/experimental/bidi/models/gemini_live.py
+++ b/strands-py/src/strands/experimental/bidi/models/gemini_live.py
@@ -19,7 +19,7 @@ from typing import Any, AsyncGenerator, cast
 
 from google import genai
 from google.genai import types as genai_types
-from google.genai.types import LiveConnectConfigOrDict, LiveServerMessage
+from google.genai.types import LiveConnectConfigOrDict, LiveServerContent, LiveServerMessage, UsageMetadata
 
 from ....types._events import ToolResultEvent, ToolUseStreamEvent
 from ....types.content import Messages
@@ -219,6 +219,9 @@ class BidiGeminiLiveModel(BidiModel):
         - modelTurn text: Text response from the model
         - usageMetadata: Token usage information
 
+        `usageMetadata` sits outside the `messageType` union, so it can accompany any other field and
+        is collected independently of the content events.
+
         Returns:
             List of event dicts (empty list if no events to emit).
 
@@ -230,93 +233,31 @@ class BidiGeminiLiveModel(BidiModel):
                 message.go_away.model_dump_json(), live_session_handle=self._live_session_handle
             )
 
+        events: list[BidiOutputEvent] = []
+
         if message.session_resumption_update:
             resumption_update = message.session_resumption_update
             if resumption_update.resumable and resumption_update.new_handle:
                 self._live_session_handle = resumption_update.new_handle
                 logger.debug("session_handle=<%s> | updating gemini session handle", self._live_session_handle)
-            return []
 
-        # Handle interruption first (from server_content)
-        if message.server_content and message.server_content.interrupted:
-            return [BidiInterruptionEvent(reason="user_speech")]
-
-        # Handle input transcription (user's speech) - emit as transcript event
-        if message.server_content and message.server_content.input_transcription:
-            input_transcript = message.server_content.input_transcription
-            # Check if the transcription object has text content
-            if hasattr(input_transcript, "text") and input_transcript.text:
-                transcription_text = input_transcript.text
-                logger.debug("text_length=<%d> | gemini input transcription detected", len(transcription_text))
-                return [
-                    BidiTranscriptStreamEvent(
-                        delta={"text": transcription_text},
-                        text=transcription_text,
-                        role="user",
-                        # TODO: https://github.com/googleapis/python-genai/issues/1504
-                        is_final=bool(input_transcript.finished),
-                        current_transcript=transcription_text,
-                    )
-                ]
-
-        # Handle output transcription (model's audio) - emit as transcript event
-        if message.server_content and message.server_content.output_transcription:
-            output_transcript = message.server_content.output_transcription
-            # Check if the transcription object has text content
-            if hasattr(output_transcript, "text") and output_transcript.text:
-                transcription_text = output_transcript.text
-                logger.debug("text_length=<%d> | gemini output transcription detected", len(transcription_text))
-                return [
-                    BidiTranscriptStreamEvent(
-                        delta={"text": transcription_text},
-                        text=transcription_text,
-                        role="assistant",
-                        # TODO: https://github.com/googleapis/python-genai/issues/1504
-                        is_final=bool(output_transcript.finished),
-                        current_transcript=transcription_text,
-                    )
-                ]
+        if message.server_content:
+            events.extend(self._convert_server_content(message.server_content, has_audio=bool(message.data)))
 
         # Handle audio output using SDK's built-in data property
-        # Check this BEFORE text to avoid triggering warning on mixed content
         if message.data:
             # Convert bytes to base64 string for JSON serializability
             audio_b64 = base64.b64encode(message.data).decode("utf-8")
-            return [
+            events.append(
                 BidiAudioStreamEvent(
                     audio=audio_b64,
                     format="pcm",
                     sample_rate=cast(AudioSampleRate, self.config["audio"]["output_rate"]),
                     channels=cast(AudioChannel, self.config["audio"]["channels"]),
                 )
-            ]
+            )
 
-        # Handle text output from model_turn (avoids warning by checking parts directly)
-        if message.server_content and message.server_content.model_turn:
-            model_turn = message.server_content.model_turn
-            if model_turn.parts:
-                # Concatenate all text parts (Gemini may send multiple parts)
-                text_parts = []
-                for part in model_turn.parts:
-                    # Check if part has text attribute and it's not empty
-                    if hasattr(part, "text") and part.text:
-                        text_parts.append(part.text)
-
-                if text_parts:
-                    full_text = " ".join(text_parts)
-                    return [
-                        BidiTranscriptStreamEvent(
-                            delta={"text": full_text},
-                            text=full_text,
-                            role="assistant",
-                            is_final=True,
-                            current_transcript=full_text,
-                        )
-                    ]
-
-        # Handle tool calls - return list to support multiple tool calls
         if message.tool_call and message.tool_call.function_calls:
-            tool_events: list[BidiOutputEvent] = []
             for func_call in message.tool_call.function_calls:
                 tool_use_event: ToolUse = {
                     "toolUseId": cast(str, func_call.id),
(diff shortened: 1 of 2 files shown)


strands-py/src/strands/experimental/bidi/models/gemini_live.py
```diff
@@ -230,152 +233,153 @@ def _convert_gemini_live_event(self, message: LiveServerMessage) -> list[BidiOut
...
+            server_content: Server content to convert.
+            has_audio: Whether the enclosing message carries audio output. Text from `model_turn` is
+                skipped when it does, since the two represent the same response in different modalities.
+
+        Returns:
+            List of events derived from the server content.
+        """
+        events: list[BidiOutputEvent] = []
+
+        if server_content.interrupted:
```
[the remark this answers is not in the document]
mehtarac: Yes, intentional. Per the Live API reference, interrupted, outputTranscription, and modelTurn are all fields on the same BidiGenerateContentServerContent object, not mutually exclusive message types — the old early-return dropped whichever fields came after the first match.

strands-py/src/strands/experimental/bidi/models/gemini_live.py
```diff
@@ -230,152 +233,153 @@ def _convert_gemini_live_event(self, message: LiveServerMessage) -> list[BidiOut
...
+            events.append(BidiInterruptionEvent(reason="user_speech"))
```
strandly-the-agent, line 292: 🟡 **A barge-in can now replay the stale audio it was supposed to discard.**

`BidiInterruptionEvent` is appended here, inside `_convert_server_content`, which runs *before* the audio append at `:248-256`. Downstream, `io/audio.py:259` clears the playback buffer on the interruption and `io/audio.py:254` then re-buffers the audio that arrived in the same message — so the clear is immediately undone. Verified with real `google.genai` types:

```
input: LiveServerContent(interrupted=True, model_turn=Content(parts=[Part(inline_data=<pcm>)]))
  base 11ad6366: ['BidiInterruptionEvent']                         # audio dropped
  head 3bfda5ec: ['BidiInterruptionEvent', 'BidiAudioStreamEvent']
```

An end-to-end run through the real `_BidiAudioOutput`/`_BidiAudioBuffer` confirms it: 0 bytes queued for playback on base, 12 bytes queued at head.

I could **not** confirm the Gemini wire actually puts `interrupted` in the same `serverContent` as `modelTurn` audio — the docs narrate interruption as its own message (*"the server **then** sends a `BidiGenerateContentServerContent` message to report the interruption"*) yet never say the fields are exclusive, and the field's own doc says it is *"a good signal to stop and empty the current playback queue"*, which co-arriving audio defeats. So this is 🟡, not 🔴 — but if the shape occurs at all, clear-then-refill is never the intent. Your reply on the earlier thread justified co-emission on protocol *shape* grounds, which I buy; this is the separate question of whether post-interruption **audio** should be surfaced.

Cheapest fix — emit audio first, so a downstream buffer clear wins:

```suggestion
            events.append(BidiInterruptionEvent(reason="user_speech"))
```
<sub>(no change on this line — the move is in `_convert_gemini_live_event`: hoist the `if message.data:` audio block above the `_convert_server_content` call, keeping `has_audio` as-is.)</sub>

Either way, worth a test that pins the order: mutating the emission so audio comes first leaves all 27 tests green today, and no test combines `data=` with `interrupted=True`, so whichever ordering you decide on is currently unprotected.

strands-py/src/strands/experimental/bidi/models/gemini_live.py
```diff
@@ -230,152 +233,153 @@ def _convert_gemini_live_event(self, message: LiveServerMessage) -> list[BidiOut
...
+                    text=output_transcript.text,
+                    role="assistant",
+                    # TODO: https://github.com/googleapis/python-genai/issues/1504
+                    is_final=bool(output_transcript.finished),
+                    current_transcript=output_transcript.text,
+                )
+            )
+
+        # Reading model_turn parts directly avoids the mixed-content warning raised by message.data
+        if not has_audio and server_content.model_turn and server_content.model_turn.parts:
```
strandly-the-agent, line 324: 🟡 **`output_transcription` + `model_turn` text now appends the same assistant turn to history twice.**

Both branches emit `role="assistant"`, and this one hardcodes `is_final=True` (`:334`). `loop.py:319-322` appends a `Message` to `agent.messages` for *every* final transcript with no de-dup, so one utterance lands twice. `has_audio` can't help: a text part carries no `inline_data`, so `message.data` is `None`. Verified with real types:

```
input: LiveServerContent(output_transcription=Transcription(text="Paris.", finished=True),
                         model_turn=Content(parts=[Part(text="Paris.")]))
  base 11ad6366: [Transcript(assistant, final, 'Paris.')]
  head 3bfda5ec: [Transcript(assistant, final, 'Paris.'), Transcript(assistant, final, 'Paris.')]
```

An end-to-end run through the real `_BidiAgentLoop` gives one assistant message in `agent.messages` on base and two identical ones at head. Consequence is worse than the interruption ordering: duplicated history is re-sent to the model on every subsequent turn. Reachability is again unproven on the wire — `outputAudioTranscription` is on by default (`:113-117`), so it needs a text-only `modelTurn` to co-arrive, most plausibly a TEXT-modality session or a thinking-enabled model.

If `output_transcription` and `model_turn` text are two renderings of one response — the same reasoning the `has_audio` guard already encodes — the guard can just cover this case too:

```suggestion
        if (
            not has_audio
            and not (output_transcript and output_transcript.text)
            and server_content.model_turn
            and server_content.model_turn.parts
        ):
```

strands-py/tests/strands/experimental/bidi/models/test_gemini_live.py
```diff
@@ -553,6 +567,160 @@ async def test_event_conversion(mock_genai_client, model):
...
+
+
+@pytest.mark.asyncio
+async def test_usage_metadata_modality_details(mock_genai_client, model, live_message, usage_metadata):
+    """Prompt and response token details merge into per-modality usage."""
+    _, _, _ = mock_genai_client
+    await model.start()
+
+    prompt_detail = unittest.mock.Mock()
+    prompt_detail.modality = "AUDIO"
```
strandly-the-agent, line 639: 🟡 **This fixture makes the assertion pass on a value the real API never produces.**

`modality` is set to the plain string `"AUDIO"`, so production's `str(detail.modality).lower()` (`gemini_live.py:357`) coincidentally yields `"audio"` and the test goes green. On the wire `detail.modality` is a `MediaModality` enum, and `str(MediaModality.AUDIO).lower()` is `"mediamodality.audio"` — which violates `ModalityUsage.modality: Literal["text","audio","image","cached"]` (`types/events.py:460`). So the one new test covering modality details certifies the wrong string and can't catch the real bug.

The production bug is pre-existing (see the appendix on the summary comment), so fixing it isn't your obligation here — but the fixture is new, and it's a two-character fix in a line this PR already moves:

```suggestion
    prompt_detail.modality = genai_types.MediaModality.AUDIO
```

With the real enum, `"modality": "audio"` only passes once `_convert_usage_metadata` uses `detail.modality.value.lower()`. Same for `response_detail` at `:643`. If you'd rather keep this PR scoped to the usage fix, that's fine — just leave the fixture out of it rather than pinning a string the wire can't produce.

strands-py/tests/strands/experimental/bidi/models/test_gemini_live.py
```diff
@@ -553,6 +567,160 @@ async def test_event_conversion(mock_genai_client, model):
...
+    await model.stop()
+
+
+@pytest.mark.asyncio
+async def test_usage_metadata_emitted_alongside_session_resumption(
+    mock_genai_client, model, live_message, usage_metadata
+):
+    """Session resumption tracks the handle and still emits co-attached usage metadata.
+
+    Guards https://github.com/strands-agents/harness-sdk/issues/3745 — this branch previously
```
strandly-the-agent, line 605: 🟡 **Dangling issue link (×3), and this docstring narrates the diff.**

`https://github.com/strands-agents/harness-sdk/issues/3745` doesn't resolve — I checked via the API, and it's neither an issue nor a PR (the newest issue in the repo is ~3721). Same link at `:577` and `:675`, and the PR's own "Related Issues" section is empty. `docs/TESTING.md:82` draws the line as: a regression test for a *discovered bug* links the issue it guards; a test written as part of feature development carries **no** issue reference. These ship in the same PR as the fix, so unless there's a real issue to point at, dropping the links is the rule-conformant move.

Separately, *"this branch previously returned early, discarding usage outright"* narrates how the code changed, which root `AGENTS.md` explicitly forbids ("Never narrate how the code changed or what it used to be") — that reasoning belongs in the PR description.

```suggestion
    """Session resumption tracks the handle and still emits co-attached usage metadata.

    usageMetadata sits outside the messageType union, so it can accompany any other field,
    including a session-resumption update.
    """
```

strandly-the-agent (commented):
**Fresh review at `3bfda5ec`** — verdict: **comment**. No blocker. The usage fix is right, and the restructure quietly recovers two *other* silent drops (an interrupted turn's `inputTranscription`, and audio riding with an `outputTranscription`). Two **new** co-emission orderings deserve a decision, and one premise deserves confirming.

**Verified on this branch** (real `google.genai.types`, not mocks; differential vs base `11ad6366`):
- ✅ `27 passed` — `python -m pytest tests/strands/experimental/bidi/models/test_gemini_live.py -q` (google-genai 2.17.0)
- ✅ the fix works: `serverContent(audio) + usageMetadata` → `['BidiAudioStreamEvent', 'BidiUsageEvent']` at head vs `['BidiAudioStreamEvent']` on base
- ✅ floor is safe: `LiveServerContent`/`UsageMetadata` and every field read exist at `google-genai==1.32.0`
- ✅ shape matches repo convention (`openai_realtime.py:562-609` already accumulates a list with usage appended last)
- 🟡 2 new behaviors reproduced end-to-end through the real `_BidiAudioOutput` / `_BidiAgentLoop` (inline)
- 🟡 mutation check: emitting audio **before** the `server_content` events still passes all 27 tests → event order is unpinned

### ❓ Question (worth settling before merge)

Is Gemini Live's `usageMetadata` a **per-message delta** or a **running total**? `loop.py:340-344` sums every `BidiUsageEvent`. If it's cumulative, then emitting *more* usage events — precisely what this PR does — would make session accounting **over**count rather than fix an undercount, inverting the premise. I couldn't settle it: the Live API reference documents the fields but never says which; the official capabilities-guide example doesn't sum, it prints each message standalone as `"Used {usage.total_token_count} tokens in total"` under the comment *"the server will periodically send messages that include UsageMetadata"*; and several developer-forum reports describe `promptTokenCount` as cumulative/growing per turn. One captured live session settles it — do you have one handy? Happy to be wrong here.

<details>
<summary>Coverage gaps (non-blocking) — 3</summary>

- No test pairs `usage_metadata` with a **`tool_call`** or with a **transcript**, though the new docstring's own claim is that usage "can accompany any other field" — and usage landing with a tool-call message is arguably more common than with a session-resumption update.
- `test_usage_metadata_modality_details` uses `AUDIO` for both prompt and response details, so only the merge branch (`gemini_live.py:369-370`) runs; the new-modality append branch (`:372-374`, e.g. audio-in/text-out) is never exercised.
- `if detail.modality and detail.token_count:` (`:354`, `:365`) silently drops a legitimately reported `token_count == 0`. Intentional?

</details>

<details>
<summary>Appendix — non-blocking (5)</summary>

**Two pre-existing bugs, verified byte-identical on base — they belong in issues, not this PR.** I tried to file them and my token lacks `issues:write` here, so they're written up below for someone with permissions (or say the word and I'll draft them as an issue body you can paste).

1. ⚪ **`modality_details` emits `"mediamodality.audio"`, violating its own type contract.** `gemini_live.py:357,367` do `str(detail.modality).lower()`; on the wire `detail.modality` is a `MediaModality` enum with no `__str__` override, so:
   ```
   >>> str(t.MediaModality.AUDIO).lower()   -> 'mediamodality.audio'
   >>> t.MediaModality.AUDIO.value.lower()  -> 'audio'
   ```
   Real-types run against this branch emits `{'modality': 'mediamodality.audio', 'input_tokens': 7, 'output_tokens': 0}`, but `ModalityUsage.modality` is `Literal["text","audio","image","cached"]` (`types/events.py:460`) — any consumer branching on it silently fails to match. Same at floor 1.32.0 and at 2.17.0. Fix is `detail.modality.value.lower()`; `openai_realtime.py:608` may share the pattern.

2. ⚪ **`part.thought` isn't filtered, so model reasoning can leak into transcripts and history.** `gemini_live.py:326` takes every part with `.text`; google-genai's own `.text` property explicitly skips thought parts. Real-types repro: parts `[Part(text="reasoning: they want the capital", thought=True), Part(text="Paris.")]` → transcript `"reasoning: they want the capital Paris."`, where `message.text` gives `"Paris."`. Pre-existing, but the change in the second inline comment widens how often the path fires.

3. ⚪ `message.data` is now evaluated **twice** per message (`:245` and `:248`). It's a computed property that runs `part.model_dump()` per part and logs the "non-data parts" warning for mixed parts — at the 1.32 floor there's no once-per-process dedup, so a mixed message warns twice, which sits oddly beside the comment at `:323`. Measured cost is negligible (3.9 µs/chunk), so this is purely a `has_data = bool(message.data)` tidy-up.

4. ⚪ `test_event_conversion` is still one 113-line multi-scenario test — pre-existing, and the PR re-pointed it at the new fixtures without splitting it, even though it introduces exactly that per-scenario pattern next door.

5. ⚪ Pre-existing ruff `UP035` on `gemini_live.py:18` (`AsyncGenerator` from `typing`) — also on base, not yours.

</details>

<details>
<summary>How this review ran</summary>

Staged pipeline: routing triage → context build → parallel specialist passes (correctness/safety, adversarial-repro, test-quality) → aggregation with de-dup and a reachability gate → this post. Each pass was an independent fresh-context agent; the adversarial pass ran warm (no nested spawn available) against its own role rubric. I re-verified every load-bearing finding myself with real `google.genai` types before posting. Two threads from the earlier automated review are deliberately **not** re-raised: the exclusive→cumulative shift (you answered it) and the whole-object `BidiUsageEvent` assertions (fixed in `3bfda5ec`).

AI review — solid work for a human to approve, not a gate.

</details>

strands-py/src/strands/experimental/bidi/models/gemini_live.py
strandly-the-agent: **Withdrawing this** — the WebSockets reference disproves the reachability.

`generationComplete`: *"When model is interrupted while generating there will be no 'generation_complete' message in interrupted turn, it will go through 'interrupted > turn_complete'."* That describes those flags arriving as **their own messages in sequence**, and the guide's interruption text agrees (*"the server **then** sends a `BidiGenerateContentServerContent` message to report the interruption"*). No documented shape puts `interrupted` in the same `serverContent` as `modelTurn` audio.

By my own bar that makes this ⚪ defensive hardening with no trigger, not 🟡 — so: no action needed, sorry for the detour. The ordering argument would still hold *if* a capture ever shows the shape, but I'm not asking you to design against an undocumented one.

strands-py/src/strands/experimental/bidi/models/gemini_live.py
strandly-the-agent: **Withdrawing this too** — and the reference answers it in your favour.

`outputTranscription`: *"Output audio transcription. The transcription is sent independently of the other server messages and there is no guaranteed ordering, in particular not between serverContent and this outputTranscription."* (`inputTranscription` says the same.) Transcriptions arrive in their **own** messages, so the `output_transcription` + `model_turn`-text co-occurrence I was worried about isn't a documented shape either → ⚪, no action.

Worth saying explicitly: that same sentence is the strongest argument *for* this PR's design. Fields that are sent independently with no guaranteed ordering are precisely what an early-return chain gets wrong, so the accumulate-into-a-list shape is the documented-correct one — not just a convenience for the follow-up.

pgrayy (approved):

PR #3739, opened 2026-08-10
docs: add bidi observability documentation

mehtarac:
## Description

This PR adds a reference doc covering observability for BidiAgent. It documents the session-rooted trace structure (bidi_session with bidi_connect, bidi_response, execute_tool, and bidi_connection_restart as child spans), the attributes on each span, and why interruptions are recorded as span events rather than their own spans. It also covers provider-specific differences in response lifecycle and usage reporting across Nova Sonic, OpenAI Realtime, and Gemini Live, along with logging, system-prompt redaction, sampling, best practices, and common issues. 

## Related Issues

#1726 

## Type of Change

Documentation update

## Testing

## Checklist

----


site/src/content/docs/user-guide/concepts/bidirectional-streaming/observability.mdx
```diff
@@ -0,0 +1,517 @@
...
+
+`BidiAgent` observability is currently available for Python only.
+
+## Enabling Tracing
+
+:::note[Install OpenTelemetry Dependencies]
+To enable OTEL exporting, install Strands Agents with the `otel` optional dependency:
+
+```shell
+pip install 'strands-agents[otel]'
```
strandly-the-agent: 🟡 The install note only mentions the `otel` extra, but the very next code block imports `BidiNovaSonicModel`, which lives behind the **`bidi`** extra. Verified — with only `[otel]` installed:

```
$ python -c "from strands.experimental.bidi.models import BidiNovaSonicModel"
ModuleNotFoundError: No module named 'aws_sdk_bedrock_runtime'
```

`otel` is *only* the OTLP exporter (`pyproject.toml:63`); `aws_sdk_bedrock_runtime` comes from `bidi` (`:79`). Every sibling bidi page uses `[bidi]` / `[bidi,bidi-io]` / `[bidi-all]` — this is the only one with a bare `[otel]`. A reader working top-to-bottom hits it on the page's first sample. (`[bidi-all]` would also work; it already bundles `otel`.)

````suggestion
To enable OTEL exporting, install Strands Agents with the `bidi` and `otel` optional
dependencies:

```shell
pip install 'strands-agents[bidi,otel]'
````

**Review by strandly-the-agent — COMMENTED** [review]
**TL;DR** — nothing blocking in the page itself: 6 small 🟡 accuracy/wording fixes, all inline below. CI is red because of the `site/package-lock.json` regeneration (already flagged above) — reverting it to `main`'s lockfile should clear both failing checks.

The notable result is what *didn't* break. I attacked the page empirically rather than just reading it: both hand-written console-exporter JSON blocks are byte-accurate against real exporter output — same keys, same order, same values, same parent/child ids — and all six redaction recipes behave exactly as written, including the empty-value token and the documented comma pitfall. Hand-written span JSON in docs doesn't usually survive that. Nice page.

<details>
<summary>What I actually verified (branch <code>21505251a</code>)</summary>

✅ Site build: `npm run build` → 884 pages, exit 0, `[astro-broken-links-checker] No broken links detected`, new page rendered. `npm test` → 528 passed / 2 skipped. `npm run typecheck` → exit 0. Mermaid block parses (with a negative control proving the parser actually validates). Frontmatter validated through the real content schema; all 17 links on the page resolve via the site's own `resolveHref`/`resolveApiShorthand`, including `@api/python/...` and the `#session-level-attributes` self-anchor.

✅ Every code sample **executed**, not just parsed — installed `strands-py` and drove the documented flows against a stubbed `BidiModel` with an in-memory + console exporter. Reproduced and held: zero-valued usage attributes omitted while non-zero kept; `time_to_first_audio` recorded once, only on the response span, absent on a text-only turn; Nova Sonic's all-but-last-`interrupted` behaviour with TTFA re-baselining; tool spans parented to the session; response/tool span overlap (measured 50.4 ms); connect failure closing both spans ERROR and surfacing the exception; restart span always carrying `gen_ai.error.message`, status `OK` ⟺ reconnected (and ERROR → raises to `receive()`); all 7 DEBUG lines with exact text in exact order.

✅ Doc ↔ implementation checked line-for-line against `_telemetry.py`, `loop.py`, `types/events.py`, `hooks/events.py`, `tracer.py` and the three providers. Nothing the implementation emits is undocumented; nothing documented is unimplemented.

🔴 Could **not** verify: no live provider network calls, so the **Provider Differences table (`:478-482`) is source-verified only**, not run — worth a human eye if you have a session against each provider. Also untested: `setup_otlp_exporter()` against a real collector, the Jaeger `docker run` recipe (no docker), and `npm ci` (the lockfile issue). `typecheck:snippets` does not cover this page — it only globs `src/content/docs/**/*.ts`.

Counts: 3 specialist passes → 15 raw findings → 6 posted (🔴 0 · 🟡 6 · ⚪ 0), 9 deduped/suppressed/routed.
</details>

### Questions (both non-blocking)

❓ #1726 asks for "Otel **and metrics**", and both this PR and #3282 ("feat(bidi): add telemetry and metrics") cover spans only — `grep -rn "metrics" strands-py/src/strands/experimental/bidi/` returns zero hits, versus a substantial non-streaming analogue (`AgentResult.metrics`, plus `MetricsClient` emitting real instruments in `telemetry/metrics.py:549-575`). The page is honest about this at `:308`. Since the body says "Related Issues" rather than "Closes", merging won't auto-close #1726 — but would it be worth re-scoping or splitting that issue so the metrics half doesn't quietly read as done?

❓ `observability-evaluation/observability.mdx` (the generic landing page) has no pointer to any concept-specific observability doc — this is the first of its kind, so there's no convention either way. Is discovery via the bidi folder's nav intentional, or would a one-line link forward help a reader who starts from the generic entry point?

<details>
<summary>Appendix — non-blocking (5)</summary>

- **`:503`** "No spans at all → Install the `otel` extra" is slightly imprecise: the console exporter works without it, only OTLP needs it. The row already says "and configure an exporter", so this is cosmetic.
- **`:172-176`** `incomplete` has a second cause the table omits — a provider stream that simply *ends* with a response open (session still open) also yields it. Contrived reach: all three bundled providers' generators are `while True:` with no natural exit, so this needs a custom provider.
- **Completeness (no line)** — the page doesn't mention how to correlate a persisted `session_id` (`FileSessionManager`/`S3SessionManager`) with the OTel session span; no such attribute exists in `_telemetry.py` to point at. A one-liner would close the loop, but nothing here is wrong.
- **Pre-existing, not this PR — `gen_ai.error.message` is a mangled tuple repr.** `BidiModelTimeoutError.__init__` does `super().__init__(self, message)` (`strands-py/src/strands/experimental/bidi/models/model.py:133`), passing `self` as an arg, so `str(err)` — which `loop.py:245` writes to the attribute — renders as `"(BidiModelTimeoutError(...), 'connection idle for 8 minutes')"`. Reproduced. The doc's claims at `:183`/`:508` are true in substance; the reader just sees junk in the value.
- **Pre-existing, not a doc defect — `StopReason` is missing `"incomplete"`.** `loop.py:351` emits it at runtime; `types/events.py:83` types it `Literal["complete","error","interrupted","tool_use"]`, and `end_response_span(stop_reason: str | None)` (`_telemetry.py:100`) hides the gap from the type checker. The doc's table is right about behaviour.

I found three code-level bugs while verifying this page — the mangled timeout message above, the dead `interrupted_response_id` field (see the inline comment), and the `StopReason` gap. None is this PR's job. Happy to file them as separate issues if you want; I didn't want to spray issues on your docs review unasked.
</details>

---
<sub>Automated review by `strandly-the-agent` — an experimental Strands-built agent. Solid work for a human to approve, not a gate; I can be wrong, so push back where I am.</sub>

File: site/src/content/docs/user-guide/concepts/bidirectional-streaming/observability.mdx [hunk 2]
```diff
@@ -0,0 +1,517 @@
...
+export OTEL_EXPORTER_OTLP_ENDPOINT="http://collector.example.com:4318"
+
+# Set default OTLP headers
+export OTEL_EXPORTER_OTLP_HEADERS="key1=value1,key2=value2"
+```
+
+## Trace Structure
+
+The session span is the root span and opens before the connection is established.
+Response and tool spans are children of the session, not of each other:
```
strandly-the-agent: 🟡 The session span is only a root span when nothing else is active. `start_session_span` calls `tracer._start_span(...)` with no `parent_span` and no `force_root` (`_telemetry.py:45`), and `_start_span` falls back to `trace_api.get_current_span()` (`tracer.py:237-238`) — so it inherits whatever is current. Repro: starting a session inside `with tracer.start_as_current_span("POST /voice")` gives

```
bidi_session A    parent=POST /voice     <-- nested, not root
POST /voice       parent=ROOT
```

That's the realistic shape for a voice service behind an instrumented FastAPI/Starlette handler, and it quietly breaks root-span-based queries ("list my sessions"). See also the paired comment on `:470`, where the sampling rationale rests on this claim.

```suggestion
The session span is the root of its own trace unless an OpenTelemetry span is already
active when the agent starts, in which case it nests under that span. It opens before
the connection is established. Response and tool spans are children of the session,
not of each other:
```

site/src/content/docs/user-guide/concepts/bidirectional-streaming/observability.mdx
```diff
@@ -0,0 +1,517 @@
...
+    "end_time": "2026-08-10T13:27:30.900000Z",
+    "status": {
+        "status_code": "OK"
+    },
+    "attributes": {
+        "gen_ai.event.start_time": "2026-08-10T13:27:25.500001+00:00",
+        "gen_ai.operation.name": "bidi_session",
+        "gen_ai.system": "strands-agents",
+        "gen_ai.agent.name": "Voice Assistant",
+        "gen_ai.request.model": "amazon.nova-sonic-v1:0",
```
strandly-the-agent: 🟡 Model-id mismatch with the worked example. The sample at `:53` builds `BidiNovaSonicModel()` with no override, which defaults to `amazon.nova-2-sonic-v1:0` (`nova_sonic.py:123`), but this span JSON — same agent name, "Voice Assistant" — shows the older `amazon.nova-sonic-v1:0`. Every sibling page shows v2 (`agent.mdx:222`, `models/nova_sonic.mdx:57`, `events.mdx:118`); this is the only outlier in the docs tree. A reader who runs the sample and diffs their own console output against this block will wonder what they misconfigured.

```suggestion
        "gen_ai.request.model": "amazon.nova-2-sonic-v1:0",
```

site/src/content/docs/user-guide/concepts/bidirectional-streaming/observability.mdx
```diff
@@ -0,0 +1,517 @@
...
+
+Then open the Jaeger UI at http://localhost:16686. Any OpenTelemetry-compatible backend
+works — see [Visualization and Analysis](../../observability-evaluation/traces.md#visualization-and-analysis)
+for other options.
+
+## Session Metrics
+
+`BidiAgent` does not produce a metrics summary object. Interruption and reconnect
+counts are available through hooks, and token usage and latency are available through
+the event stream.
```
strandly-the-agent: 🟡 "...and latency are available through the event stream" promises something that isn't there. No `Bidi*Event` in `types/events.py` carries a latency or timing field; `time_to_first_audio` is computed in `loop.py:305-307` purely as a local variable feeding the span attribute, never emitted on an event. The section then delivers a full worked subsection for token usage and nothing for latency, so a reader goes looking for a field that doesn't exist.

```suggestion
`BidiAgent` does not produce a metrics summary object. Interruption and reconnect
counts are available through hooks, and token usage is available through the event
stream. Per-turn latency is not carried on any event; `gen_ai.server.time_to_first_audio`
on the response span is the only ready-made latency figure.
```

site/src/content/docs/user-guide/concepts/bidirectional-streaming/observability.mdx
```diff
@@ -0,0 +1,517 @@
...
+        registry.add_callback(BidiInterruptionEvent, self.on_interruption)
+        registry.add_callback(BidiAfterConnectionRestartEvent, self.on_restart)
+
+    def on_interruption(self, event: BidiInterruptionEvent) -> None:
+        self.interruptions += 1
+        logger.info(
+            "reason=<%s>, response_id=<%s> | user interrupted the model",
+            event.reason,
+            event.interrupted_response_id,
+        )
```
strandly-the-agent: 🟡 `event.interrupted_response_id` is always `None` here, on every bundled provider. The *stream* event can't carry it — its constructor takes `reason` only:

```
BidiInterruptionEvent.__init__(self, reason: Literal['user_speech', 'error'])
  -> {'type': 'bidi_interruption', 'reason': 'user_speech'}
```

`loop.py:336` populates the hook field via `event.get("interrupted_response_id")` from that dict, and all five provider construction sites (`nova_sonic.py:621,645`, `openai_realtime.py:247,535`, `gemini_live.py:242`) pass only `reason=`. Running the example as written logs `response_id=<None>`. The page already gives the working alternative at `:205` (look for the `interrupted` finish reason), so the example probably shouldn't advertise a field that can't carry data yet. (Underlying SDK gap — happy to file it separately.)

```suggestion
        logger.info("reason=<%s> | user interrupted the model", event.reason)
```

site/src/content/docs/user-guide/concepts/bidirectional-streaming/observability.mdx
```diff
@@ -0,0 +1,517 @@
...
+standard OpenTelemetry variables:
+
+```bash
+# Example: Sample 50% of traces
+export OTEL_TRACES_SAMPLER="traceidratio"
+export OTEL_TRACES_SAMPLER_ARG="0.5"
+```
+
+Because the session is the root span, a sampling decision keeps or drops the entire
+session rather than individual spans within it.
```
strandly-the-agent: 🟡 Paired with the comment on `:81` — the conclusion is right but the premise isn't. `traceidratio` samples off the trace id, so what guarantees all-or-nothing is that the whole session is *one trace*, which holds whether or not the session span is the root.

```suggestion
Because the whole session is one trace, a sampling decision keeps or drops all of its
spans together rather than individual spans within it.
```

site/src/content/docs/user-guide/concepts/bidirectional-streaming/observability.mdx
```diff
@@ -0,0 +1,517 @@
...
+| Gemini Live | None | Emits no response lifecycle events, so a session produces no `bidi_response` spans. Interruptions, token usage including cache reads, and the modality breakdown are still reported. |
+
+## Best Practices
+
+1. **Monitor time to first audio over session duration**: a long session can be
+   healthy; a slow first audio chunk is user-facing latency.
+2. **Separate connect latency from response latency**: use `bidi_connect` to isolate
+   handshake cost from model response time.
+3. **Track reconnects with hooks**: each reconnect is a window during which input was
+   rejected, so monitor the rate rather than treating individual reconnects as errors.
```
strandly-the-agent: 🟡 "rejected" mischaracterizes the behaviour — `_BidiAgentLoop.send()` does `await self._send_gate.wait()` while the gate is closed (`loop.py:190-192`): the call blocks and then proceeds once the reconnect completes. Nothing raises, nothing is dropped. "Rejected" implies an error callers must handle, and it contradicts the page's own more careful wording at `:185` ("the window during which the agent could not accept input").

```suggestion
3. **Track reconnects with hooks**: each reconnect is a window during which input is
   queued rather than sent, so monitor the rate rather than treating individual
   reconnects as errors.
```

site/package-lock.json
lizradway, line 1: why are we updating this in this pr? remove probably?

lizradway: oh i see to attempt to fix CI? looks like its still failing and no docs deploy

mehtarac: yes! removing it in the next iteration. Have to push the changes

site/src/content/docs/user-guide/concepts/bidirectional-streaming/observability.mdx
```diff
@@ -0,0 +1,517 @@
...
+`BidiAgent` sessions differ from request/response agents in structure: a single
+connection spans multiple model responses, tool calls, and possibly reconnects, and a
+response can end because the user interrupted it rather than because the model
+finished. This page covers the traces, metrics, and logs specific to bidirectional
+streaming. For general tracing, metrics, and logging concepts shared by every Strands
+agent, see [Traces](../../observability-evaluation/traces.md),
+[Metrics](../../observability-evaluation/metrics.md), and
+[Logs](../../observability-evaluation/logs.md).
+
+`BidiAgent` observability is currently available for Python only.
```
lizradway: shouldn't this just be the language support header instead of inline?

site/src/content/docs/user-guide/concepts/bidirectional-streaming/observability.mdx
```diff
@@ -0,0 +1,517 @@
...
+carrying an `interruption.reason` attribute, since an interruption is a point in time
+rather than an operation with a duration. The response that was cut off separately
+closes with `gen_ai.response.finish_reason` set to `interrupted`.
+
+To count interruptions, query the `bidi_interruption` event on the session span. To
+find which response was cut off, look for the `interrupted` finish reason. Recording
+the event on the session rather than the response also captures barge-ins that arrive
+between responses. For how barge-in is detected, see [Interruptions](interruption.md).
+
+## Inspecting Spans Locally
```
lizradway, line 182: is this any different than the generic SDK observability jaeger instructions? can we just point there?

site/src/content/docs/user-guide/concepts/bidirectional-streaming/observability.mdx
```diff
@@ -0,0 +1,517 @@
...
+Two span names carry a suffix identifying what they cover: `bidi_session <agent name>`
+and `execute_tool <tool name>`. The response ID is an attribute rather than part of the
+span name to avoid a distinct span name per turn, which fragments aggregation in most
+tracing backends.
+
+Response spans and tool spans can overlap in wall-clock time, since the model may keep
+streaming audio while a tool call is in progress. Tool spans are children of the
+session rather than of the response that requested the tool.
+
+## Captured Attributes
```
lizradway, line 116: wondering if all of these need to be headers? it makes the page really hard to navigate.

site/src/content/docs/user-guide/concepts/bidirectional-streaming/observability.mdx
lizradway: yes pls

site/src/content/docs/user-guide/concepts/bidirectional-streaming/observability.mdx
```diff
@@ -0,0 +1,517 @@
+---
+title: Bidirectional Streaming Observability
+description: 'Trace, measure, and log BidiAgent sessions: connection lifecycle,
+  time to first audio, interruptions, per-modality token usage, and reconnects.'
+sidebar:
+  label: "Observability"
+experimental: true
```
lizradway, line 8: is this supposed to be experimental? slash are you graduating it in docs all at once at the end?

site/src/content/docs/user-guide/concepts/bidirectional-streaming/observability.mdx
lizradway: we also might want to make some of these tables /code collapsible to reduce information overload/bulk?

site/src/content/docs/user-guide/concepts/bidirectional-streaming/observability.mdx
mehtarac: yes planning on graduating all at once

lizradway: ok nice!

site/src/content/docs/user-guide/concepts/bidirectional-streaming/observability.mdx
mehtarac: Addressed in the newest commit

lizradway (approved):

PR #3741, opened 2026-08-10
docs: add prescriptive guidance for keeping cognitive complexity low

yonib05:
## Description

Adds prescriptive, public guidance for the `complexity/*` label, structured so each document carries only what its readers need:

**`team/COMPLEXITY.md` (new)** is the canonical document, placed in `team/` because that is where this repository keeps its reasoning. It covers:

- **Why the metric is public**: it is the "Simple at any scale" tenet made measurable; it lets maintainers triage scarce review attention across ~200 open PRs; a reproducible number depersonalizes the "please split this" conversation; and authors see the signal locally before any reviewer does.
- **The scoring model**: one point per break in linear flow, plus one per nesting level, so nesting is the multiplier. Calibrated against this repo (median function scores 1 to 2; `low` covers ~90% of existing functions).
- **Guidance at three altitudes**, not just flattening:
  - *Design it simple*: one function one job, separate deciding from doing, I/O at the edges, variants as data, tests as a pressure gauge.
  - *Write it flat*: guard clauses, extracted helpers, lookup tables over branch ladders, uniform named booleans, errors at the edges, sequential await.
  - *Keep the change simple*: preparatory refactors as their own PR ("make the change easy, then make the easy change"), extract rather than deepen when landing in an already-complex function, and test generously since `size/*` excludes tests.
- **When high is honest**: protocol converters and state machines may correctly stay complex; say so in the description instead of contorting code to move a label.

**CONTRIBUTING.md** stays small: what the labels are, one paragraph on why, how to run the check locally, and a pointer to the team doc.

**AGENTS.md** gets one Writing Code bullet so agents see the expectation before writing, linking to the team doc.

## Type of Change

Documentation update

## Testing

The techniques are verified against the real analyzer rather than asserted. A deliberately nested `process()` scores **23** with complexipy; rewritten with exactly the documented techniques (guard clauses, a dispatch table, one extraction) and unchanged behavior, the maximum touched function scores **4**:

```
nested.py    process      complexity=23
flat.py      process      complexity=4
flat.py      _convert     complexity=3
```

Calibration numbers come from measuring both SDKs during threshold selection (strands-py p50=1, p90=12; strands-ts p50=2, p90=9). All cross-document links and in-document anchors verified against rendered heading slugs. Docs-only change; no code paths to exercise.

## Checklist

----

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -219,6 +219,44 @@ splitting, not a rule; sometimes a complex function is the honest solution.
 
 A docs-only or test-only PR touches no SDK source and gets no complexity label.
 
+#### Keeping complexity low
+
+Two things drive the score, and knowing them is most of the battle:
+
+1. Every break in linear flow costs one point: `if`, `else`, loops,
+   `catch`/`except`, ternaries, `switch`/`match`, recursion, and each run of
+   mixed boolean operators.
+2. **Nesting multiplies the cost.** Each structure also pays one point per
+   level of nesting it sits inside. An `if` at the top of a function costs 1;
+   the same `if` three levels deep costs 4.
+
+Because nesting is the multiplier, flattening is the highest-leverage move:
+
+- **Return early.** Guard clauses (`if not valid: return`) remove a nesting
+  level from everything that follows. Avoid `else` after a `return`.
+- **Extract nested logic into named helpers.** The nesting penalty resets to
+  zero inside the helper, the name documents intent, and the label keys on the
+  most complex single function you touched, so splitting one deep function into
+  three shallow ones directly lowers it.
+- **Prefer lookup tables over branch ladders.** A dict or `Map` from value to
+  handler replaces an `if`/`elif` or `switch` chain with zero branches.
+- **Keep boolean sequences uniform and name the pieces.** `a and b and c` costs
+  one point; mixing `and`/`or` costs more. `is_retryable = ...` beats a long
+  inline condition twice over: cheaper and self-describing.
+- **Handle errors at the edges.** One `try` around a coherent block costs less
+  than error handling threaded through nested branches.
+- **In async code, await sequentially** instead of nesting callbacks or `then`
+  chains.
+
+For calibration: the median function in both SDKs scores 1 to 2, and
+`complexity/low` (10 or less) covers roughly nine out of ten existing
+functions, so the thresholds leave generous room for ordinary logic. Some code
+is honestly complex, though. Protocol event converters and state machines often
+land at `high` because the domain has that many cases, and flattening them
+further would hurt readability. When that is your situation, keep the function
+and say so in the PR description rather than contorting the code to move a
+label.
+
 ## Using AI Tools
 
 We love AI. We build with coding agents every day, and you're welcome to use them too — they're a great way to move fast and explore a codebase.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -219,43 +219,11 @@ splitting, not a rule; sometimes a complex function is the honest solution.
 
 A docs-only or test-only PR touches no SDK source and gets no complexity label.
 
-#### Keeping complexity low
-
-Two things drive the score, and knowing them is most of the battle:
-
-1. Every break in linear flow costs one point: `if`, `else`, loops,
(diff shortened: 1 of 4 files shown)


team/COMPLEXITY.md
```diff
@@ -0,0 +1,137 @@
+# Code Complexity
+
+Every PR is labeled with the [cognitive complexity](https://www.sonarsource.com/docs/CognitiveComplexity.pdf)
+of the most complex function it touches: `complexity/low` (10 or less),
+`complexity/medium` (11 to 25), or `complexity/high` (above 25). This document
+is the reasoning behind that label and the practices that keep code under it.
+
+## Why we measure it, publicly
```
lizradway, line 8: i really like this section

CONTRIBUTING.md
```diff
@@ -219,6 +219,12 @@ splitting, not a rule; sometimes a complex function is the honest solution.
 
 A docs-only or test-only PR touches no SDK source and gets no complexity label.
 
+The labels exist because review attention is the scarcest resource this
+project has: they let maintainers triage which PRs need an unhurried senior
+read, and they give you the same signal locally before a reviewer sees it. The
+reasoning, the scoring model, and the practices that keep code under the
+thresholds live in [team/COMPLEXITY.md](./team/COMPLEXITY.md).
```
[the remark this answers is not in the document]
yonib05: Fixed in `fa62a466c`. CONTRIBUTING.md keeps the commands, since its role is what the labels are, why they exist, and how to run the check; `team/COMPLEXITY.md`'s **Check before you push** section now points there instead of carrying a verbatim copy. One source for the commands, no drift.

CONTRIBUTING.md
```diff
@@ -219,6 +219,44 @@ splitting, not a rule; sometimes a complex function is the honest solution.
...
+  inline condition twice over: cheaper and self-describing.
+- **Handle errors at the edges.** One `try` around a coherent block costs less
+  than error handling threaded through nested branches.
+- **In async code, await sequentially** instead of nesting callbacks or `then`
+  chains.
+
+For calibration: the median function in both SDKs scores 1 to 2, and
+`complexity/low` (10 or less) covers roughly nine out of ten existing
+functions, so the thresholds leave generous room for ordinary logic. Some code
+is honestly complex, though. Protocol event converters and state machines often
```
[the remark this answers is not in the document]
yonib05: Agreed with your follow-up: the restructure scoped "often land at `high`" to protocol converters, state machines, and exhaustive format mappings in **When high is honest**, which is exactly the subset where scores above 25 are plausible. Leaving the wording as is.

team/COMPLEXITY.md
```diff
@@ -0,0 +1,137 @@
...
+   mixed boolean operators.
+2. **Nesting multiplies the cost.** Each structure also pays one point per
+   level of nesting it sits inside. An `if` at the top of a function costs 1;
+   the same `if` three levels deep costs 4.
+
+For calibration: the median function in both SDKs scores 1 to 2, and
+`complexity/low` covers roughly nine out of ten existing functions. The
+thresholds leave generous room for ordinary logic.
+
+## Design it simple
```
lizradway: are the "design it simple" and "write it flat" sections necessary? feels like this is proper code etiquette that we don't need to rewrite, but maybe could link an existing standard somewhere?

lizradway: this document is for humans, not agents right?

lizradway: if this is "for-agents" rather than a team tenet maybe we should just have a contributing skill that outlines al of this alongside our CONTRIBUTING.MD?

just thinking out loud.

team/COMPLEXITY.md
yonib05: Thanks! That section is the heart of it: the label only works if the reasoning behind it is public too.

team/COMPLEXITY.md
yonib05: Both, deliberately. AGENTS.md describes `team/` as where the 'why' lives, and this doc is linked from both CONTRIBUTING.md (humans) and the AGENTS.md Writing Code bullet (agents). The reasoning is the same for either audience; only the entry points differ.

lizradway (approved):

mehtarac (approved):

PR #3747, opened 2026-08-10
test(models): retry inconclusive Mantle routing probes

yonib05:
## Description

The Mantle routing drift tests can fail on transient model outages even after request-level retries, blocking unrelated PRs. Use each SDK's standard integration retry policy for one additional sweep while preserving misrouted and unserved failures; the live catalog sweep also replaces the redundant hard-coded regression model cases.

## Related Issues

Follow-up to #3691. Observed in #3677.

## Documentation PR

Not required; this is a test-only reliability fix.

## Type of Change

Bug fix

## Testing

- `ruff format --check`, `ruff check`, and `python3 -m py_compile` for the Python Mantle integration test
- TypeScript build, lint, formatting, type-check, browser bundle, and 3,997 unit tests
- The full pre-push command was blocked by local infrastructure: Rust 1.92 cannot build the current `litellm` dependency, and the configured CodeArtifact registry returns 404 for `npm audit`

## Checklist

----

### DIFF
--- a/strands-py/tests_integ/models/test_mantle_routing.py
+++ b/strands-py/tests_integ/models/test_mantle_routing.py
@@ -3,10 +3,11 @@
 Mantle serves each model from exactly one base path (``/v1`` or ``/openai/v1``), rejects
 the other with HTTP 400, and exposes no API that reports the routing, so
 :data:`strands.models._openai_bedrock._OPENAI_PATH_MODEL_PREFIXES` goes stale whenever Mantle
-onboards a model. For every model in the live catalog, this test probes the resolved path
-first and the alternate on failure. A 200 from the resolved path or a definitive rejection
-from the alternate confirms the mapping; other outcomes are retried and fail as
-undetermined when neither path supplies definitive routing evidence.
+onboards a model. For every model in the live catalog, this test asserts that the
+resolved path serves it, probing the other path only on failure to distinguish misrouted
+from unserved ids. Only HTTP 200 and 400 count as answers; any other status is retried
+and, if it persists, fails the test as undetermined. The integration-test retry policy
+gives transient external-service failures one more sweep before they hold the gate.
 
 Failure means the table needs updating, not that the SDK is broken for existing models.
 """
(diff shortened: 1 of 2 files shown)


strands-py/tests_integ/models/test_mantle_routing.py
```diff
@@ -155,6 +155,10 @@ def check(model_id: str) -> tuple[str, str, str]:
         on_other = _serves(other, model_id, _mint_token())
         if on_other:
             return model_id, "misrouted", resolved
+        # A definitive rejection from the alternate path confirms the routing table even
+        # when model invocation on the resolved path is transiently unavailable.
+        if on_resolved is None and on_other is False:
```
mkmeral: why don't we add retries? I'm not a fan of this approach. 

> Having an error in different API endpoint confirms the API endpoint we didn't actually call is successful? 

I don't want to depend on it, we can add retries? and we don't have to integ test all the model ids explicitly

yonib05: Fair, but the point of this test is to notify us early of missing routings since this could sit silent for a while as it did.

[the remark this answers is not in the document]
yonib05: Addressed. The alternate-path inference is removed, so an inconclusive resolved probe can no longer hide an unserved model. Only a definitive rejection from both paths produces the unserved verdict; inconclusive models receive one targeted full routing retry before the sweep fails.

yonib05: Addressed. Both SDK sweeps now retry only model ids whose complete resolved/alternate probe remains undetermined, and the redundant hard-coded model-specific integration cases are removed.

yonib05: Follow-up: I replaced the custom targeted loop with the repository conventions. Python now uses `retry_on_flaky` for only the undetermined-routing assertion, and TypeScript relies on the existing Vitest integration-project retry policy. The hard-coded model cases remain removed.

JackYPCOnline (approved):

PR #3772, opened 2026-08-12
fix(models/bedrock): redact blocked content when guardrail trace is disabled

### LINKED ISSUE #3612: [BUG] guardrailConfig.trace: 'disabled' silently disables all guardrail redaction — blocked content stays in conversation history
### Checks

### SDK Language

TypeScript

### Strands Version

monorepo `main` @ `c0776c008be7d16367f7d5cca792a635fa0b92b1`

### Language Runtime Version

Node.js v24.14.0

### Operating System

Amazon Linux 2023.12.20260710 (Linux 6.1.161-18.298.amzn2023.aarch64)

### Installation Method

git clone

### Steps to Reproduce

Setting `guardrailConfig.trace: 'disabled'` silently disables **all** guardrail redaction, so content a Bedrock guardrail blocked stays in `agent.messages` (and in whatever a `SessionManager` persists). `stopReason` is still `guardrailIntervened`, so the interception itself works — only the redaction is lost.

The whole redaction path is gated on the *response* carrying a `trace` field:

```ts
// strands-ts/src/models/bedrock.ts:1377-1383  (non-streaming)
if (event.trace) {
  metadataEvent.trace = event.trace
  if (this._config.guardrailConfig && event.trace.guardrail && stopReasonRaw === 'guardrail_intervened') {
    for (const redactionEvent of this._generateRedactionEvents(event.trace.guardrail)) { events.push(redactionEvent) }
  }
}

// strands-ts/src/models/bedrock.ts:1564-1570  (streaming) — same shape, `data.trace`
```

With `trace: 'disabled'` Bedrock returns no trace, so `if (event.trace)` / `if (data.trace)` is false, `_generateRedactionEvents` is never reached, no `modelRedactionEvent` is emitted, `result.redaction` is never set at `models/model.ts:517-519`, and `Agent._redactLastMessage` (`agent/agent.ts:1988-1989`) is never called.

`trace` is a documented, public config field (`bedrock.ts:189`: `trace?: 'enabled' | 'disabled' | 'enabled_full'`) and the SDK defaults the request to `'enabled'` (`bedrock.ts:764`), so this only bites users who explicitly opt out of trace — plausibly to reduce log volume or avoid sending assessment detail, with no reason to expect it turns off redaction too.

**Repro** — real `Agent` + real `BedrockModel` with only the HTTP transport (`_client.send`) stubbed, so `_mapStreamedBedrockEventToSDKEvent`, `_generateRedactionEvents`, `Model.streamAggregated` and `Agent._redactLastMessage` all execute for real. No AWS credentials needed. Both cases send the same blocked prompt and receive `stopReason: 'guardrail_intervened'`; the only difference is whether the response carries a guardrail trace.

```
================ A. trace: 'enabled'  (control) ================
request.guardrailConfig = {"guardrailIdentifier":"gr-1","guardrailVersion":"1","trace":"enabled"}
response.stopReason    = guardrailIntervened

--- agent.messages after the intervention (2 messages) ---
[0] role=user
      (0) textBlock text="[User input redacted.]"
[1] role=assistant
      (0) textBlock text="Sorry, I can't help with that."

blocked content still in history? no
redaction applied?                YES

================ B. trace: 'disabled' (the bug) ================
request.guardrailConfig = {"guardrailIdentifier":"gr-1","guardrailVersion":"1","trace":"disabled"}
response.stopReason    = guardrailIntervened

--- agent.messages after the intervention (2 messages) ---
[0] role=user
      (0) textBlock text="my SSN is 123-45-6789 BLOCKED_MARKER_7c1e"
[1] role=assistant
      (0) textBlock text="Sorry, I can't help with that."

blocked content still in history? YES
  -> messages[0].content[0] (textBlock, role=user)
redaction applied?                NO
```

The repro script is not committed anywhere; happy to attach it or open a PR with it as a unit test if that's useful.

### Expected Behavior

When a Bedrock guardrail blocks and `redaction.input` is enabled (the default), the blocked user content should be redacted from conversation history regardless of the `guardrailConfig.trace` setting. Trace controls *diagnostic reporting*; it shouldn't silently change the safety behavior.

If redaction genuinely cannot work without a trace, then either (a) `trace: 'disabled'` combined with redaction enabled should warn loudly at construction time, or (b) the `trace` doc comment at `bedrock.ts:189` and the guardrails user guide should state the dependency explicitly.

### Actual Behavior

Nothing is redacted. `stopReason` is `guardrailIntervened`, the model's reply is the guardrail's canned refusal, and the blocked content remains verbatim in `agent.messages` — and therefore in session storage and in every subsequent request's history. There is no warning or log line indicating redaction was skipped; the only hint is the absence of the `redacting input due to guardrail` debug log at `bedrock.ts:1752`.

### Additional Context

**Not a regression** — verified this reproduces identically on `main` (`c0776c008`) and on the branch of #3608, since neither touches the `if (event.trace)` / `if (data.trace)` guards. Found while reviewing #3608 and split out here rather than commented inline, because it's independent of that PR.

Same gate exists on both the streaming (`:1564`) and non-streaming (`:1377`) paths, so `streamProcessingMode` makes no difference.

One consequence worth calling out: because the blocked content is never removed, it is resent on every following turn, so a guardrail that blocks deterministically will keep blocking — the conversation is stuck. That is precisely the failure mode the guardrails user guide says redaction exists to prevent (`site/src/content/docs/user-guide/safety-security/guardrails.mdx`: *"This is done so that follow-up questions are not also blocked by the same questions"*).

### Possible Solution

Hoist the guardrail check out of the `trace` guard so it's driven by the stop reason rather than by trace reporting, e.g. at both `:1377` and `:1564`:

```ts
if (event.trace) {
  metadataEvent.trace = event.trace
}

if (this._config.guardrailConfig && stopReasonRaw === 'guardrail_intervened') {
  for (const redactionEvent of this._generateRedactionEvents(event.trace?.guardrail ?? {})) {
    events.push(redactionEvent)
  }
}
```

That needs `_generateRedactionEvents` to cope with an empty/absent assessment. Note this interacts with the design question in #3608, which is moving that function toward keying off assessment content — worth settling the two together so `trace: 'disabled'` doesn't end up fail-open by a different route. A test would fit in `strands-ts/src/models/__tests__/bedrock.test.ts` under `describe('redaction event generation')`.

The Python SDK has the same structural dependency (`strands-py/src/strands/models/bedrock.py:1038-1042` guards on `"trace" in response`), so a fix should probably be mirrored — `sdk-parity`.

### Related Issues

Found while reviewing #3608. Related but distinct: #2119 (redaction not persisted to `AgentCoreMemorySessionManager`), #1077 / #1080 (tool-result redaction breaking `toolUse`/`toolResult` pairing).

Unshure:
### Human written description

This change adds a fallback behavior to guardrail handling in the bedrock model provider. Originally, the model provider required checking the trace data to see if a guardrail fired. Now, we check the trace data, and fall back to checking the stop reason `guardrail_intervened` to see if a guardrail triggered.

Im not sure what the historic reason was to check guardrail trace instead of stop reason, but with this new change, all previous integ tests pass and the xfail python test also passes.

We could consider removing the trace check, and instead relying only on the stop reason. But again, since im not sure of the historic reason of checking trace instead of stop reason, im hesitant to remove it. If someone can give some background, I would be happy to update the implementation.

### Agent written description
The change is applied to **both** SDKs to keep them at parity (a hard repo rule):

- **`strands-ts`** — the SDK the issue was filed against. Both the streaming and
  non-streaming paths now check the stop reason and use `trace?.guardrail ?? {}`.
- **`strands-py`** — the parity change. A `_should_redact_guardrail_content(guardrail_data,
  stop_reason)` helper detects a block from either a blocked trace assessment
  (existing behavior) or a `guardrail_intervened` stop reason, wired into both
  `_stream` branches. A `redaction_emitted` guard ensures redaction fires at most
  once per stream.

## Related Issues

Fixes #3612

## Type of Change

Bug fix

## Testing

- **TypeScript**: added streaming + non-streaming regression tests asserting a
  `modelRedactionEvent` is emitted when `guardrail_intervened` is reported with no
  trace. Full unit suite (193 node tests) passes; `type-check`, `build`, `lint`,
  and `format:check` all clean.
- **Python**: added streaming + non-streaming regression unit tests for the
  no-trace case, and **un-xfailed** the `test_guardrail_input_intervention`
  integ case for `guardrail_trace="disabled"`, which now passes. Full unit suite
  (202 tests) passes; linter clean.

Note: the un-xfailed integ test requires real AWS Bedrock credentials and the
`test-guardrail-block-cactus` guardrail; it runs against pre-provisioned
resources in CI rather than locally.

## Checklist

----

### DIFF
--- a/strands-ts/src/models/bedrock.ts
+++ b/strands-ts/src/models/bedrock.ts
@@ -683,29 +683,39 @@ export class BedrockModel extends Model<BedrockModelConfig> {
         if (response.stream) {
           let lastStopReason: string | undefined
           let redactionEmitted = false
+          let sawGuardrailTrace = false
           for await (const chunk of response.stream) {
-            // Map Bedrock events to SDK events
             const result = this._mapStreamedBedrockEventToSDKEvent(chunk, lastStopReason)
             lastStopReason = result.stopReason
-            // Redaction keys off the stop reason (see _generateRedactionEvents), but the guardrail
-            // trace only arrives with the metadata event. Emit exactly once, on the first metadata
-            // event that follows a guardrail_intervened stop reason.
-            if (
-              !redactionEmitted &&
-              result.guardrailData &&
-              lastStopReason === 'guardrail_intervened' &&
-              this._config.guardrailConfig
-            ) {
-              redactionEmitted = true
-              yield* this._generateRedactionEvents(result.guardrailData)
+            // The guardrail trace arrives with the metadata event and is authoritative for
+            // BLOCKED vs ANONYMIZED (masking) — Bedrock reports guardrail_intervened for both
+            // but only BLOCKED content warrants SDK redaction. Wait for the metadata chunk to
+            // decide, and fall back to the stop reason only when no trace was carried.
+            if (!redactionEmitted && this._config.guardrailConfig && 'metadata' in chunk) {
+              const guardrailData = chunk.metadata?.trace?.guardrail
+              if (guardrailData !== undefined) {
+                sawGuardrailTrace = true
+                if (this._hasBlockedGuardrailPolicy(guardrailData)) {
+                  redactionEmitted = true
+                  yield* this._generateRedactionEvents(guardrailData)
+                }
+              } else if (lastStopReason === 'guardrail_intervened') {
+                redactionEmitted = true
+                yield* this._generateRedactionEvents({})
+              }
             }
             for (const event of result.events) {
               yield event
             }
           }
 
-          // Guardrail intervened but no metadata event carried a trace: still redact.
-          if (!redactionEmitted && lastStopReason === 'guardrail_intervened' && this._config.guardrailConfig) {
+          // Safety net: guardrail_intervened but no metadata event ever arrived.
+          if (
+            !redactionEmitted &&
+            !sawGuardrailTrace &&
+            lastStopReason === 'guardrail_intervened' &&
+            this._config.guardrailConfig
+          ) {
             yield* this._generateRedactionEvents({})
           }
         }
@@ -1610,11 +1620,16 @@ export class BedrockModel extends Model<BedrockModelConfig> {
       metadataEvent.trace = event.trace
     }
 
-    // Redaction keys off the stop reason, not the trace: `trace: 'disabled'` still reports
-    // `guardrail_intervened` but carries no assessment, and blocked content must still be redacted.
+    // Bedrock reports guardrail_intervened for both BLOCKED and ANONYMIZED (masking). When the
+    // trace is carried it is authoritative — only BLOCKED policies warrant redaction. When it
+    // isn't (typically guardrailConfig.trace='disabled'), fall back to the stop reason.
     if (this._config.guardrailConfig && stopReasonRaw === 'guardrail_intervened') {
-      for (const redactionEvent of this._generateRedactionEvents(event.trace?.guardrail ?? {})) {
-        events.push(redactionEvent)
+      const guardrail = event.trace?.guardrail
+      const shouldRedact = guardrail !== undefined ? this._hasBlockedGuardrailPolicy(guardrail) : true
+      if (shouldRedact) {
+        for (const redactionEvent of this._generateRedactionEvents(guardrail ?? {})) {
+          events.push(redactionEvent)
+        }
       }
     }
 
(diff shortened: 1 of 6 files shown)


strands-ts/src/models/bedrock.ts
```diff
@@ -1686,12 +1687,13 @@ export class BedrockModel extends Model<BedrockModelConfig> {
...
+        }
 
-          // Check for blocked guardrails in trace and emit redaction events
-          if (this._config.guardrailConfig && data.trace.guardrail && lastStopReason === 'guardrail_intervened') {
-            for (const redactionEvent of this._generateRedactionEvents(data.trace.guardrail)) {
-              events.push(redactionEvent)
-            }
+        // Redaction keys off the stop reason, not the trace: `trace: 'disabled'` still reports
+        // `guardrail_intervened` but carries no assessment, and blocked content must still be redacted.
+        if (this._config.guardrailConfig && lastStopReason === 'guardrail_intervened') {
```
gautamsirdeshmukh: Could we move this redaction handling into `stream()` so it is emitted exactly once? We could mark redaction as pending when `messageStop` reports `guardrail_intervened`, then `yield` the redaction events using the first metadata trace (or after the stream loop with {} if metadata never arrives)

As written, it looks like missing metadata would skip redaction and repeated metadata emits duplicates

gautamsirdeshmukh (approved):

PR #3783, opened 2026-08-12
ci(anthropic): re-enable anthropic integration tests

opieter-aws:
## Description

The Python Anthropic integration tests have been disabled for over a year. This restores the tests to the standard API-key gate every other cloud provider uses (`providers.anthropic.mark`), so they skip locally without a key and run in CI where the key is provisioned.

While the suite was disabled, three tests rotted: two pinned `claude-sonnet-4-20250514`, which the API now 404s, and `TestCountTokens` never actually hit the native count-tokens path because its fixture omitted `use_native_token_count=True`. All model IDs now flow from a single `MODEL_ID` constant, and the context-overflow test's payload is sized to `claude-sonnet-4-6`'s larger context window.

Verified against a live `ANTHROPIC_API_KEY`: all 10 tests pass; with no key set, all 10 skip via the mark rather than a module-level skip.

## Related Issues

N/A

## Documentation PR

N/A

## Type of Change

CI

## Testing

## Checklist

----

### DIFF
--- a/strands-py/tests_integ/models/test_model_anthropic.py
+++ b/strands-py/tests_integ/models/test_model_anthropic.py
@@ -9,17 +9,12 @@ from strands.agent import NullConversationManager
 from strands.models.anthropic import AnthropicModel
 from strands.types.content import ContentBlock, Message
 from strands.types.exceptions import ContextWindowOverflowException
+from tests_integ.models import providers
 
-"""
-These tests only run if we have the anthropic api key
+# these tests only run if we have the anthropic api key
+pytestmark = providers.anthropic.mark
 
-Because of infrequent burst usage, Anthropic tests are unreliable, failing tests with 529s.
-{'type': 'error', 'error': {'details': None, 'type': 'overloaded_error', 'message': 'Overloaded'}}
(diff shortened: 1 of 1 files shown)


yonib05 (approved):

strands-py/tests_integ/models/test_model_anthropic.py
```diff
@@ -9,17 +9,12 @@
...
+pytestmark = providers.anthropic.mark
 
-Because of infrequent burst usage, Anthropic tests are unreliable, failing tests with 529s.
-{'type': 'error', 'error': {'details': None, 'type': 'overloaded_error', 'message': 'Overloaded'}}
-https://docs.anthropic.com/en/api/errors#http-errors
-"""
-pytestmark = pytest.skip(
-    "Because of infrequent burst usage, Anthropic tests are unreliable, failing with 529s", allow_module_level=True
-)
+MODEL_ID = "claude-sonnet-4-6"
```
lizradway, line 17: non blocking: should we update this as follow up?

opieter-aws: You mean to opus 4.8 or 5.0? Not opposed

Issue #3785, opened 2026-08-12
[BUG] AnthropicModel silently drops DocumentBlocks with unsupported formats (docx/xls/xlsx) from the request


yonib05:
### Checks

- [x] I have updated to the lastest minor and patch version of Strands
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched [./issues](./issues?q=) and there are no duplicates of my issue

### SDK Language

TypeScript

### Strands Version

1.12.0

### Language Runtime Version

Node.js 24.15.0

### Operating System

macOS 26.4.1

### Installation Method

npm

### Steps to Reproduce

1. Construct a `DocumentBlock` with a format outside the Anthropic provider's supported set (`pdf` plus the text-file list in `src/models/anthropic.ts:30`), for example `docx`, and pass it to an `Agent` configured with `AnthropicModel`:

```typescript
import { Agent } from '@strands-agents/sdk'
import { AnthropicModel } from '@strands-agents/sdk/anthropic'
import { readFileSync } from 'node:fs'

const agent = new Agent({ model: new AnthropicModel({ modelId: 'claude-sonnet-5' }) })
await agent.invoke([
  { type: 'textBlock', text: 'Summarize the attached file.' },
  {
    type: 'documentBlock',
    name: 'report',
    format: 'docx',
    source: { type: 'documentSourceBytes', bytes: readFileSync('report.docx') },
  },
])
```

2. The request succeeds and the model responds as if no file was attached.

The drop can be observed directly in the formatted request: `_formatContentBlock` (`src/models/anthropic.ts:457-495`) returns `undefined` for the block after a `logger.warn`, and `_formatMessages` (`src/models/anthropic.ts:384-392`) skips `undefined` blocks, so the content block count in the outgoing request is one less than the input.

### Expected Behavior

Passing a document the provider cannot deliver to the Anthropic API should fail loudly with a catchable error, the same way the `imageBlock` case in the same file already does (`src/models/anthropic.ts:440` throws `Unsupported image format for Anthropic: ...`).

### Actual Behavior

The document is silently omitted from the request. The only signal is a `logger.warn` (`format=<docx> | unsupported document format or source for anthropic`); no error is thrown and the caller has no programmatic way to detect the drop. The model answers as if no file was attached.

### Additional Context

Affected inputs are wider than just `docx`:

[...]

PR #3790, opened 2026-08-12
fix(anthropic): raise on unsupported document and image formats instead of sending undeliverable requests

### LINKED ISSUE #3789: [BUG] Python AnthropicModel sends undeliverable document formats to the API, failing with an opaque 400
### Checks

### SDK Language

Python

### Strands Version

1.52.0

### Language Runtime Version

Python 3.11.9

### Operating System

macOS 26.4.1

### Installation Method

git clone

### Steps to Reproduce

Pass a non-PDF, non-txt document to an `Agent` configured with `AnthropicModel`:

```python
agent = Agent(model=AnthropicModel(model_id="claude-sonnet-5", max_tokens=1024))
agent([
    {"text": "Summarize the attached file."},
    {"document": {"format": "docx", "name": "report", "source": {"bytes": docx_bytes}}},
])
```

`_format_request_message_content` (`strands-py/src/strands/models/anthropic.py:126-140`) resolves the media type with `mimetypes.types_map.get(f".{format}", "application/octet-stream")` and forwards every document as a base64 `document` source unless the resolved type is exactly `text/plain`.

### Expected Behavior

Formats Anthropic can actually accept are delivered; formats it cannot accept fail fast client-side with a clear, catchable error, matching the TypeScript provider after #3785.

### Actual Behavior

Anthropic's Messages API only accepts `application/pdf` base64 sources and `text/plain` text sources for document blocks, so only `pdf` and `txt` survive. Everything else in the `DocumentContent` format literal is sent in a shape the API rejects with an opaque 400:

- `doc`, `docx`, `xls`, `xlsx`: base64 with `application/msword`, `application/vnd.ms-excel`, or `application/octet-stream`
- `csv`, `html`: base64 with `text/csv` / `text/html`, even though the content is plain text the API could take
- `md`: base64 with `application/octet-stream` on Python < 3.13 (`mimetypes` has no `.md` entry there)

The error surfaces as a raw `anthropic.BadRequestError` after a network round trip, with no indication that the SDK constructed an undeliverable request.

### Additional Context

Counterpart to #3785 (TypeScript), which fixed the same underlying problem in the other direction: TypeScript silently dropped undeliverable documents and now throws client-side. The TypeScript provider also delivers text-file formats (`csv`, `html`, `md`, ...) as plain text rather than sending them in a rejected shape, so Python currently rejects inputs that work on TypeScript.

Widening the `DocumentContent` format literal itself (e.g. adding `json`/`xml`) is tracked separately in #2204.

The image path has the same defect: formats outside `_IMAGE_MEDIA_TYPES` fall back to a `mimetypes` lookup defaulting to `application/octet-stream` and are forwarded in a shape the API rejects, where the TypeScript provider raises. The fix should cover both content types.

### Possible Solution

In the Anthropic provider, deliver `pdf` as a base64 document source, deliver the text formats in the literal (`csv`, `html`, `md`, `txt`) as `text/plain` text sources, and raise a `TypeError` for the formats Anthropic cannot accept (`doc`, `docx`, `xls`, `xlsx`), mirroring the TypeScript provider and the existing `TypeError` contract of `_format_request_message_content`.

### Related Issues

#3785, #2204

yonib05:
## Description

The Python Anthropic provider guessed media types with `mimetypes` and forwarded content to the API even when the resulting request shape was one Anthropic always rejects, so undeliverable content failed as an opaque `anthropic.BadRequestError` after a network round trip instead of a clear client-side error.

**Documents.** Anthropic's Messages API only accepts `application/pdf` base64 sources and `text/plain` text sources for document blocks, so only `pdf` and `txt` ever survived: `csv` and `html` were sent base64 with `text/csv`/`text/html`, `md` as `application/octet-stream` on Python < 3.13, and `doc`/`docx`/`xls`/`xlsx` in shapes the API always rejects. Document handling now matches the TypeScript provider: `pdf` is delivered as a base64 document source (unchanged), the text formats in the `DocumentContent` literal (`csv`, `html`, `md`, `txt`) are delivered as `text/plain` text sources so they actually work, and the binary Office formats raise a `TypeError` client-side.

**Images.** Formats outside `_IMAGE_MEDIA_TYPES` fell back to a `mimetypes` lookup defaulting to `application/octet-stream`, another doomed request. They now raise the same client-side `TypeError`, matching the TypeScript provider's unsupported-image-format behavior. Supported formats are unchanged.

`TypeError` is already the documented error contract of `_format_request_message_content` and matches the Mistral provider's unsupported-format precedent.

A note on the parity claim, since the failure modes only align once both PRs land: TypeScript on `main` still warns-and-skips undeliverable documents; #3786 changes it to throw, and this PR gives Python the matching raise. One divergence is intentional and remains: TypeScript's delivered text-format list is wider (`json`, `xml`, `yaml`, ...) than Python's, because those formats are not in Python's `DocumentContent` literal; that gap is #2204's scope. Non-UTF-8 bytes in a text-format document also raise the documented `TypeError` (rather than an undocumented `UnicodeDecodeError`).

```python
# Before: docx forwarded as base64 application/octet-stream, API returns an opaque 400
# Before: csv forwarded as base64 text/csv, API returns an opaque 400
# Before: bmp image forwarded as base64 image/bmp, API returns an opaque 400

# After: csv delivered as a text/plain text source (works)
# After: docx and bmp raise client-side before any API call
# TypeError: content_type=<document>, format=<docx> | unsupported format
# TypeError: content_type=<image>, format=<bmp> | unsupported format
```

**Alternatives considered:**

- **Raise for everything except `pdf` and `txt`** (pure error-surface fix). Rejected: `csv`/`html`/`md` are plain text Anthropic can accept, and the TypeScript provider already delivers them, so raising would open a parity gap in the opposite direction.
- **Keep forwarding and let the API reject.** Rejected in #3786 for TypeScript for the same reason it fails here: the error is an opaque provider 400 after a network round trip instead of a clear client-side message.
- **Widen the delivered text formats to TypeScript's full list** (`json`, `xml`, `yaml`, ...). Deferred: those formats are not in Python's `DocumentContent` format literal; widening the literal is tracked in #2204.

**Callout:** intentional behavior change on a public provider. Requests that previously failed with a provider 400 for undeliverable documents and images now raise `TypeError` client-side, and `csv`/`html`/`md` documents that previously failed now succeed.

## Related Issues

Resolves #3789

Related: #3785, #3786 (TypeScript counterpart), #2204 (format literal gaps)

## Documentation PR

No documentation changes needed: the set of formats a user can pass is unchanged, only delivery and the failure mode.

## Type of Change

Bug fix

## Testing

Ran `hatch run format`, `hatch run lint`, `hatch run test-lint` (mypy), and the full unit suite on Python 3.11 (5166 tests passing, including new regression tests for the raising document and image formats and a csv delivery case).

- [x] I ran `hatch run prepare` (format, lint, test-lint, and the unit suite on 3.11; the full version matrix runs in CI)

## Checklist

----

### DIFF
--- a/strands-py/src/strands/models/anthropic.py
+++ b/strands-py/src/strands/models/anthropic.py
@@ -6,7 +6,6 @@
 import base64
 import json
 import logging
-import mimetypes
 from collections.abc import AsyncGenerator
 from typing import Any, TypeVar, cast
 
@@ -164,13 +163,13 @@ class AnthropicModel(Model):
 
         if "image" in content:
             image_format = content["image"]["format"]
+            if image_format not in _IMAGE_MEDIA_TYPES:
+                raise TypeError(f"content_type=<image>, format=<{image_format}> | unsupported format")
+
             return {
                 "source": {
                     "data": base64.b64encode(content["image"]["source"]["bytes"]).decode("utf-8"),
-                    "media_type": _IMAGE_MEDIA_TYPES.get(
-                        image_format,
-                        mimetypes.types_map.get(f".{image_format}", "application/octet-stream"),
-                    ),
+                    "media_type": _IMAGE_MEDIA_TYPES[image_format],
                     "type": "base64",
                 },
                 "type": "image",
--- a/strands-py/src/strands/models/anthropic.py
+++ b/strands-py/src/strands/models/anthropic.py
@@ -37,6 +37,9 @@ _IMAGE_MEDIA_TYPES = {
     "webp": "image/webp",
 }
 
+# Anthropic document sources accept only pdf (base64) and plain text; these formats are delivered as text.
+_TEXT_FILE_FORMATS = frozenset({"csv", "html", "md", "txt"})
+
 # Anthropic accepts ``cache_control`` on these block types only; any other block is rejected.
 # https://docs.claude.com/en/docs/build-with-claude/prompt-caching
 _CACHEABLE_BLOCK_TYPES = frozenset({"document", "image", "text", "tool_result", "tool_use"})
@@ -133,20 +136,28 @@ class AnthropicModel(Model):
             Anthropic formatted content block.
 
         Raises:
-            TypeError: If the content block type cannot be converted to an Anthropic-compatible format.
+            TypeError: If the content block type or document format cannot be converted to an
+                Anthropic-compatible format.
         """
         if "document" in content:
-            mime_type = mimetypes.types_map.get(f".{content['document']['format']}", "application/octet-stream")
+            document_format = content["document"]["format"]
+            if document_format != "pdf" and document_format not in _TEXT_FILE_FORMATS:
+                raise TypeError(f"content_type=<document>, format=<{document_format}> | unsupported format")
+
             return {
-                "source": {
-                    "data": (
-                        content["document"]["source"]["bytes"].decode("utf-8")
-                        if mime_type == "text/plain"
-                        else base64.b64encode(content["document"]["source"]["bytes"]).decode("utf-8")
-                    ),
-                    "media_type": mime_type,
-                    "type": "text" if mime_type == "text/plain" else "base64",
-                },
+                "source": (
+                    {
+                        "data": base64.b64encode(content["document"]["source"]["bytes"]).decode("utf-8"),
+                        "media_type": "application/pdf",
+                        "type": "base64",
+                    }
+                    if document_format == "pdf"
+                    else {
+                        "data": content["document"]["source"]["bytes"].decode("utf-8"),
+                        "media_type": "text/plain",
+                        "type": "text",
+                    }
+                ),
                 "title": content["document"]["name"],
                 "type": "document",
             }
--- a/strands-py/src/strands/models/anthropic.py
+++ b/strands-py/src/strands/models/anthropic.py
@@ -140,23 +140,29 @@ class AnthropicModel(Model):
         """
         if "document" in content:
             document_format = content["document"]["format"]
-            if document_format != "pdf" and document_format not in _TEXT_FILE_FORMATS:
+            if document_format == "pdf":
+                source: dict[str, Any] = {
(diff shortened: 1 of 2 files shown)


strands-py/src/strands/models/anthropic.py
```diff
@@ -133,33 +135,47 @@ def _format_request_message_content(self, content: ContentBlock) -> dict[str, An
...
+                    raise TypeError(
+                        f"content_type=<document>, format=<{document_format}> | document is not valid utf-8 text"
+                    ) from decode_error
+                source = {
+                    "data": text_data,
+                    "media_type": "text/plain",
+                    "type": "text",
+                }
+            else:
+                raise TypeError(f"content_type=<document>, format=<{document_format}> | unsupported format")
```
Unshure, line 162: Should we raise, or should we warn and pass?

Looks like we do not have a consistent behavior across the sdk's on this. I would prefer warn and pass on content that is not supported on one model, but it is on others. I could go either way on this though, curious to hear your thoughts

yonib05: Quite failing drops the doc, and the user has no idea; that seems to me like an anti-pattern. I'd rather fail loudly than quietly drop context.

Yes, it is inconsistent across the SDK; this, along with 2 other changes merged yesterday, is moving us to unifying around failing loudly. If you strongly disagree, I should probably revert those as well.

See: https://github.com/strands-agents/harness-sdk/pull/3786 & https://github.com/strands-agents/harness-sdk/pull/3792


Also, this is not model level this is API level.

JackYPCOnline: I'd recommand fail loud since removing part of the user's request will cause senmantics changes sliently. Especially document is one of the most importatnt context.

Unshure: I prefer consistency most of all, and if most people prefer to fail loudly, then im fine with that 👍

Unshure (approved):

PR #3799, opened 2026-08-13
fix(bedrock): omit falsy cache point TTLs

### LINKED ISSUE #3759: A falsy ttl on a caller-placed cache point still reaches the wire when caching is off (both SDKs)
Found while fixing the system-prompt half of this in #3571. Filing rather than folding in, because it is pre-existing on `main` in **both** SDKs and outside that PR's scope.

## Problem

A falsy `ttl` on a caller-placed cache point is dropped on the *managed* message path but sails through on the *unmanaged* one. botocore rejects `None` and Bedrock rejects `""` against its enum, so either one hard-fails the request.

With caching off (no `cache_config` / `cacheConfig`), a caller-placed message cache point keeps its empty TTL:

```python
# strands-py, no cache_config
messages = [{"role": "user", "content": [
    {"text": "hi"},
    {"cachePoint": {"type": "default", "ttl": ""}},
]}]
# emits  [{'text': 'hi'}, {'cachePoint': {'type': 'default', 'ttl': ''}}]   <- rejected
```
```ts
// strands-ts, no cacheConfig
new Message({ role: 'user', content: [new TextBlock('hi'), new CachePointBlock({ cacheType: 'default', ttl: '' })] })
// emits  [{"text":"hi"},{"cachePoint":{"type":"default","ttl":""}}]        <- rejected
```

Both reproduce on `main`. The two SDKs agree with each other, so this is not a parity bug — just a shared gap.

## Why it is inconsistent

The SDK already knows this hazard and handles it everywhere else:

| path | falsy caller TTL |
|---|---|
| managed message point (`_honor_placed_cache_point` / `_honorPlacedCachePoint`) | dropped ✅ |
| system point (after #3571) | dropped ✅ |
| config-derived section TTLs | treated as unconfigured ✅ |
| **unmanaged message point (caching off)** | **sent ❌** |

`_honor_placed_cache_point` even says why: *"Dropping it unconditionally also handles a caller TTL of None or "", which botocore and Bedrock respectively reject before the request can even be judged on placement."*

## Suggested fix

Normalize at the single point where a caller's cache point becomes wire shape, so no future guard can reintroduce it:

- **TypeScript** — one line in `_formatContentBlock` (`bedrock.ts`, `case 'cachePointBlock'`): `if (block.ttl !== undefined)` → `if (block.ttl)`. This covers system *and* message points on every config.
- **Python** — the equivalent in `_format_request_message_content`, so the message path matches the system path that #3571 just fixed.

Doing both keeps the two SDKs in step; doing only the TS one-liner would create a divergence, which is why #3571 did neither.

Low severity — it needs a caller to write an empty or null TTL explicitly, which is most plausible out of environment config (`ttl: process.env.CACHE_TTL ?? ''`). Cheap to close, and it removes the last way a caller's own TTL can hard-fail a request.

steven-ji:
## Description

Caller-placed message cache points could forward empty or null TTLs when Bedrock caching was disabled, causing botocore or Bedrock to reject an otherwise valid request. Normalize those values at the wire-format boundary in both SDKs while preserving non-empty TTLs.

## Related Issues

Resolves #3759

## Documentation PR

Not required; this corrects internal request normalization without changing public API or documented behavior.

## Type of Change

Bug fix

## Testing

- Python official check: `hatch run prepare` — format, lint, and type checks passed; Python 3.14–3.11 each passed 5167 tests, and Python 3.10 passed 5162 with 5 skipped
- Python coverage suite: `hatch test tests --cover` — 5167 passed
- TypeScript pre-commit suite: build, lint, format, type-check, and `npm run test:coverage` — 4104 passed

## Checklist

----

### DIFF
--- a/strands-py/src/strands/models/bedrock.py
+++ b/strands-py/src/strands/models/bedrock.py
@@ -814,7 +814,7 @@ class BedrockModel(Model):
         if "cachePoint" in content:
             cache_point = content["cachePoint"]
             result: dict[str, Any] = {"type": cache_point["type"]}
-            if "ttl" in cache_point:
+            if cache_point.get("ttl"):
                 result["ttl"] = cache_point["ttl"]
             return {"cachePoint": result}
 
--- a/strands-py/tests/strands/models/test_bedrock.py
+++ b/strands-py/tests/strands/models/test_bedrock.py
@@ -2717,6 +2717,30 @@ def test_format_request_preserves_cache_point_ttl(model, model_id):
     assert cache_point_block["ttl"] == "1h"
 
 
+# https://github.com/strands-agents/harness-sdk/issues/3759
+@pytest.mark.parametrize("ttl", [None, ""])
+def test_format_request_omits_falsy_cache_point_ttl(model, ttl):
+    """Falsy caller TTLs are omitted before Bedrock validates the request."""
+    messages = [
+        {
+            "role": "user",
+            "content": [
+                {
+                    "cachePoint": {
+                        "type": "default",
+                        "ttl": ttl,
+                    }
+                },
+            ],
+        }
+    ]
+
+    tru_cache_point = model.format_request(messages)["messages"][0]["content"][0]["cachePoint"]
+    exp_cache_point = {"type": "default"}
+
+    assert tru_cache_point == exp_cache_point
+
+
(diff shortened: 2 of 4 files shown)


yonib05 (approved):
Looks great, thank you!

PR #3804, opened 2026-08-13
fix(build): include strandly workspace in root build and type-check

### LINKED ISSUE #3768: [BUG] `strandly` workspace is never built or type-checked in CI
Filed as a follow-up from the review of #3734. Pre-existing gap; that PR is just what exposed it.

### Checks

### SDK Language

TypeScript

### Strands Version

`main`

### Language Runtime Version

Node.js 22.14.0

### Operating System

Amazon Linux 2023 (aarch64)

### Installation Method

git clone

### Steps to Reproduce

1. Introduce a type error in the `strandly` workspace (or change its `typescript`/`@types/node` versions, as #3734 does).
2. Open a PR and watch CI.

### Expected Behavior

A change that breaks `strandly`'s build or types should fail CI.

### Actual Behavior

CI passes regardless. Nothing builds or type-checks `strandly`:

- Root `package.json:16-31` — the `build` / `lint` / `format:check` / `type-check` / `check` scripts are **all** `npm run <x> -w strands-ts`, never `-w strandly`.
- `strandly/package.json` defines its own `build` (`tsc`) and `type-check` (`tsc --noEmit`), but nothing invokes them.
- `grep -rn strandly .github/workflows/*.yml` — `strandly` appears only as a path-filter trigger (`ci.yml:40`, `typescript-pr-and-push.yml:9`, `typescript-integration-test.yml:8`), deciding *whether* the TypeScript workflow runs, never as a build or test target inside any job. `typescript-ts-check.yml` runs the root scripts, which resolve to `-w strands-ts` per above.

So a `strandly`-only change can trigger the TypeScript workflow and still never be compiled.

### Additional Context

Concrete illustration from #3734: it bumps `strandly`'s `typescript` `^6.0.3` → `^7.0.2` and `@types/node` `^25` → `^26`. I ran `cd strandly && npx tsc --noEmit` manually — 0 errors, so that PR is safe. But that's incidental: had the major bump broken `strandly`, CI would have been green anyway.

### Possible Solution

Either add `-w strandly` to the root `build`/`type-check`/`lint` scripts (or make them run across all workspaces), or add a small `strandly` job to the TypeScript workflow that invokes the scripts the workspace already defines. The second keeps failure attribution clearer.

### Related Issues

Found during review of #3734

maxmilian:
Closes #3768

## What

`strandly` is declared in the root `package.json` `workspaces` array, but every
root script targets `-w strands-ts` only. So `npm run build` and
`npm run type-check` — the exact commands CI's TypeScript code-quality job runs —
never compiled the workspace. This adds `-w strandly` to those two scripts.

## A correction to the issue's diagnosis

The issue reads as "CI is never triggered for `strandly`". That isn't quite what
happens, and the real story is a bit worse:

- `.github/workflows/ci.yml` **does** list `strandly/**` in the `typescript`
  paths-filter, and `typescript-pr-and-push.yml` **does** list it in its `push`
  paths.
- So a change under `strandly/` **does** trigger the TypeScript workflows.
- `typescript-ts-check.yml` then runs `npm run build`, `npm run lint`,
  `npm run format:check`, `npm run type-check`, `npm run check:browser-bundle` —
  **all of which resolve to `-w strands-ts`**.

In other words the jobs run, go green, and report nothing — the path filter
actively creates the impression that the workspace is guarded when the commands
behind it never touch it. A silent green is harder to notice than no run at all.

## Verification

`strandly` already ships its own `build` / `type-check` scripts and a strict
`tsconfig.json`; nothing there needed changing, and it type-checks cleanly today
(so this PR turns the check on without dragging in a backlog of pre-existing
errors).

To prove the checks now actually cover the workspace, I appended a deliberate
type error to `strandly/src/cli.ts` and ran the root scripts with the old and
new `package.json` against identical sources:

**Before (current `main` scripts) — silently passes:**

```
$ npm run type-check ; echo exit=$?
> tsc --noEmit --project src/tsconfig.json && tsc --noEmit --project test/integ/tsconfig.json
exit=0

$ npm run build ; echo exit=$?
> tsc --project src/tsconfig.json
exit=0
```

**After (this PR) — caught by both:**

```
$ npm run type-check ; echo exit=$?
src/cli.ts(123,7): error TS2322: Type 'string' is not assignable to type 'number'.
npm error workspace @strands-agents/strandly@0.0.1
exit=2

$ npm run build ; echo exit=$?
src/cli.ts(123,7): error TS2322: Type 'string' is not assignable to type 'number'.
npm error workspace @strands-agents/strandly@0.0.1
exit=2
```

The deliberate error was reverted; it is not part of this diff.

Full local run of the CI code-quality sequence on the final tree, all green:
`build`, `lint`, `format:check`, `type-check`, `check:browser-bundle`, plus
`npm test` (145 files / 4102 tests passed).

## Scope

Deliberately limited to `build` and `type-check`:

- `lint` / `format` / `format:check` stay `-w strands-ts` — `strandly` has no
  eslint config and no `format` script, so wiring those up means introducing new
  tooling config, which belongs in its own change.
- `check` stays `-w strands-ts` — `strandly` has no `check` script, and CI runs
  `build` / `type-check` as separate steps anyway.
- Kept explicit `-w` names rather than switching to `--workspaces`, so behaviour
  for `strands-ts` is bit-for-bit unchanged.

Happy to fold any of the above in if maintainers would prefer a broader change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "scripts": {
     "dev": "strandly",
     "prepare": "husky",
-    "build": "npm run build -w strands-ts",
+    "build": "npm run build -w strands-ts -w strandly",
     "test": "npm run test -w strands-ts",
     "test:coverage": "npm run test:coverage -w strands-ts",
     "test:all": "npm run test:all -w strands-ts",
@@ -26,7 +26,7 @@
     "lint": "npm run lint -w strands-ts",
     "format": "npm run format -w strands-ts",
     "format:check": "npm run format:check -w strands-ts",
-    "type-check": "npm run type-check -w strands-ts",
+    "type-check": "npm run type-check -w strands-ts -w strandly",
     "check": "npm run check -w strands-ts",
     "check:browser-bundle": "npm run check:browser-bundle -w strands-ts",
     "complexity": "node .github/scripts/pr-metrics/run-analysis.mjs",


yonib05 (approved):
Thanks for the contribution, and for closing the gap from #3768. This makes the merge gate honest about the strandly workspace, and the two-line scope is exactly right.

Heads up: we're planning a follow-up PR that removes the strandly workspace entirely, since its remaining commands only wrap existing root npm scripts and nothing in CI or the docs invokes it anymore. That doesn't diminish this fix; until that lands, main is correct to type-check what it ships.

opieter-aws (approved):

PR #3632, opened 2026-08-04
feat(ts): add Background Tasks mechanism

gautamsirdeshmukh:
## Description

Introducing `backgroundTasks`, a new opt-in Agent capability that lets the model dispatch concurrent background tool calls and continue reasoning while they run; results are automatically delivered at safe loop boundaries. 

This capability is **disabled** by default, preserving existing Agent behavior. Developers can set `backgroundTasks: true` to enable model-driven backgrounding with defaults, or provide a `BackgroundTasksConfig` to pin specific tool behavior and override other settings (all shown below). Normal tools, agents-as-tools and those exposed by MCP clients can all be backgrounded.

See [design doc](https://github.com/strands-agents/harness-sdk/pull/3531/changes) for a deeper walkthrough of the feature, including detailed diagrams, code snippets and rationale.

## Public API Changes

The only new named exports from `@strands-agents/sdk` are `BackgroundTasksConfig`, `BackgroundTask` and a couple error types:

```typescript
export interface BackgroundTasksConfig {
  readonly waitForCompletion?: boolean
  readonly agentic?: readonly (Tool | '*')[]
  readonly always?: readonly (Tool | '*')[]
  readonly never?: readonly (Tool | '*')[]
  readonly maxConcurrency?: number
  readonly timeout?: number
}

export interface BackgroundTask {
  readonly taskId: string
  readonly toolUseId: string
  readonly toolName: string
  readonly status: 'queued' | 'working' | 'paused' | 'completed' | 'failed' | 'cancelled'
  readonly createdAt: string
  readonly updatedAt: string
  readonly result?: {
    readonly content: readonly Serialized<ToolResultContentData>[]
  }
  readonly error?: {
    readonly type: 'toolError' | 'executionError' | 'timeout' | 'recoveryError'
    readonly message: string
  }
  readonly interrupts?: readonly Readonly<Interrupt>[]
}

export class BackgroundTaskNotFoundError extends Error {}

export class BackgroundTasksTimeoutError extends Error {
  readonly timeoutMs: number
}
```

Background Tasks also extends the following existing public APIs:

```typescript
type AgentConfig = {
  backgroundTasks?: boolean | BackgroundTasksConfig
}

class Agent {
  readonly backgroundTasks: {
    readonly get: (taskId: string) => Promise<BackgroundTask | undefined>
    readonly list: () => Promise<readonly BackgroundTask[]>
    readonly cancel: (taskId: string) => Promise<BackgroundTask>
[...]

PR #3807, opened 2026-08-13
feat(ts): add execution-scoped cancellation to ToolContext

gautamsirdeshmukh:
## Description

Tool cancellation is currently coupled to the Agent invocation through `context.agent.cancelSignal`. Tool executors need an execution-scoped signal decoupled from the Agent to enable alternate execution lifetimes (i.e. background tasks) without changing existing behavior.

SDK-managed tool and middleware contexts now receive an executor-supplied `context.cancelSignal`. Foreground and direct calls receive `agent.cancelSignal` as their execution signal, while `AgentAsTool`, sequential cancellation, and retry suppression use the same executor-supplied signal. 

### Public API Changes

```diff
export interface ToolContext extends Interruptible {
  toolUse: ToolUse
  agent: LocalAgent
  invocationState: InvocationState
+ cancelSignal: AbortSignal
}
```

## Related Issues

Parent Background Tasks PR (extracted change from it): #3632

## Documentation PR

Documentation updates are included in this PR under `site/`.

## Type of Change

New feature

## Testing

- [x] Focused TypeScript unit tests
- [x] `npm run build -w strands-ts`
- [x] `npm run lint -w strands-ts`

## Checklist

----

### DIFF
--- a/site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
+++ b/site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
@@ -180,7 +180,7 @@ The `agent.cancel()` method provides a way to stop the loop from outside, such a
 
   #### Cancellation within tool execution
 
-  The SDK automatically checks for cancellation before and between tool invocations (see checkpoints above). However, once a tool callback is running, cancellation is **cooperative** — only the tool itself can respond mid-execution. Tools can participate by forwarding the signal to APIs that accept `AbortSignal`, or by polling `cancelSignal.aborted` between steps. If a tool does neither, it runs to completion and the agent resumes cancellation handling after the tool returns.
+  The SDK automatically checks for cancellation before and between tool calls (see checkpoints above). Once a tool callback is running, cancellation is **cooperative**: only the tool itself can respond mid-execution. Tools can participate by forwarding `context.cancelSignal` to APIs that accept `AbortSignal`, or by polling `context.cancelSignal.aborted` between steps. If a tool does neither, it runs to completion, and the agent resumes cancellation handling after the tool returns.
 
   ```typescript
   --8<-- "user-guide/concepts/agents/agent-loop.ts:cancel_signal"
--- a/site/src/content/docs/user-guide/concepts/tools/executors.mdx
+++ b/site/src/content/docs/user-guide/concepts/tools/executors.mdx
@@ -119,7 +119,7 @@ Cancellation works identically in both modes. Call `agent.cancel()` to request c
 <Tab label="TypeScript">
 
 - **Pre-launch cancel**: set `BeforeToolsEvent.cancel` on the batch-level hook, or call `agent.cancel()` before tools start, to produce error results for every tool in the batch.
-- **Mid-flight cancel** in sequential mode short-circuits not-yet-started tools. In concurrent mode, all tools have already launched, so each in-flight tool must cooperatively observe `agent.cancelSignal` to stop early.
+- **Mid-flight cancel** in sequential mode short-circuits not-yet-started tools. In concurrent mode, all tools have already launched, so each in-flight tool must cooperatively observe `context.cancelSignal` to stop early.
 
 </Tab>
 </Tabs>
--- a/strands-ts/src/agent/__tests__/agent-as-tool.test.ts
+++ b/strands-ts/src/agent/__tests__/agent-as-tool.test.ts
@@ -5,10 +5,12 @@ import { MockMessageModel } from '../../__fixtures__/mock-message-model.js'
 import { collectGenerator } from '../../__fixtures__/model-test-helpers.js'
 import { createMockContext } from '../../__fixtures__/tool-helpers.js'
 import { ToolValidationError } from '../../errors.js'
-import { Tool, ToolStreamEvent } from '../../tools/tool.js'
+import { BeforeModelCallEvent } from '../../hooks/events.js'
+import { createErrorResult, Tool, ToolStreamEvent } from '../../tools/tool.js'
 import { DELEGATION_DESCRIPTION_SUFFIX } from '../agent-as-tool.js'
 import { ToolResultBlock } from '../../types/messages.js'
 import { SessionManager } from '../../session/session-manager.js'
+import type { Plugin } from '../../plugins/plugin.js'
 import type { SnapshotStorage } from '../../session/storage.js'
 
 describe('AgentAsTool', () => {
@@ -130,6 +132,67 @@ describe('AgentAsTool', () => {
           text: 'Agent response',
         })
       )
+    })
+
+    it('returns an error when the execution signal cancels the wrapped agent', async () => {
+      const model = new MockMessageModel().addTurn({ type: 'textBlock', text: 'Agent response' })
+      const agent = new Agent({ model, name: 'test-agent', printer: false })
+      const tool = new AgentAsTool({ agent })
+      const controller = new AbortController()
+      agent.addHook(BeforeModelCallEvent, () => controller.abort())
+
+      const context = createMockContext({
+        name: 'test-agent',
+        toolUseId: 'tool-1',
+        input: { input: 'Hello agent' },
+      })
+      context.cancelSignal = controller.signal
+
+      const { result } = await collectGenerator(tool.stream(context))
+
+      expect(result).toEqual(createErrorResult("Agent 'test-agent' cancelled", 'tool-1'))
+    })
+
+    it('does not persist cancellation when the signal aborts during initialization', async () => {
+      let signalInitializationStarted!: () => void
+      const initializationStarted = new Promise<void>((resolve) => {
+        signalInitializationStarted = resolve
+      })
+      let releaseInitialization!: () => void
+      const initializationReleased = new Promise<void>((resolve) => {
+        releaseInitialization = resolve
+      })
+      const plugin: Plugin = {
+        name: 'delayed-initialization',
+        async initAgent(): Promise<void> {
+          signalInitializationStarted()
+          await initializationReleased
+        },
+      }
+      const model = new MockMessageModel().addTurn({ type: 'textBlock', text: 'Agent response' })
+      const agent = new Agent({ model, name: 'test-agent', plugins: [plugin], printer: false })
+      const tool = new AgentAsTool({ agent, preserveContext: true })
+      const controller = new AbortController()
+      const context = createMockContext({
+        name: 'test-agent',
+        toolUseId: 'tool-1',
+        input: { input: 'Hello agent' },
+      })
+      context.cancelSignal = controller.signal
+
+      const execution = collectGenerator(tool.stream(context))
+      await initializationStarted
+      controller.abort()
+      releaseInitialization()
+      const { result } = await execution
+
+      expect({
+        result,
+        messages: agent.messages,
+      }).toEqual({
+        result: createErrorResult("Agent 'test-agent' cancelled", 'tool-1'),
+        messages: [],
+      })
     })
 
     it('yields ToolStreamEvents wrapping sub-agent events', async () => {
--- a/strands-ts/src/agent/agent-as-tool.ts
+++ b/strands-ts/src/agent/agent-as-tool.ts
@@ -175,7 +175,7 @@ export class AgentAsTool extends Tool {
   }
 
   async *stream(toolContext: ToolContext): ToolStreamGenerator {
-    const { toolUse } = toolContext
+    const { toolUse, invocationState, cancelSignal } = toolContext
     const toolUseId = toolUse.toolUseId
 
     // Concurrency guard: loadSnapshot + agent.stream() must not overlap.
@@ -195,7 +195,10 @@ export class AgentAsTool extends Tool {
       // Stream the sub-agent, forwarding the outer invocation's state so
       // mutations in the inner agent's hooks/tools are visible to the outer
       // agent's downstream callbacks and final AgentResult.
-      const gen = this._agent.stream(input, { invocationState: toolContext.invocationState })
+      const gen = this._agent.stream(input, {
+        invocationState,
+        cancelSignal,
+      })
       let next = await gen.next()
       while (!next.done) {
         const event = next.value
@@ -208,6 +211,10 @@ export class AgentAsTool extends Tool {
         next = await gen.next()
       }
       const result = next.value
+
+      if (result.stopReason === 'cancelled') {
+        return createErrorResult(`Agent '${this.name}' cancelled`, toolUseId)
+      }
 
       // Build the tool result
       if (result.structuredOutput !== undefined) {
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -141,7 +141,7 @@ export type ToolList = (Tool | McpClient | Agent | ToolList)[]
  * - `'sequential'` — runs tool calls one at a time
  *
  * Cancellation works identically in both modes: {@link Agent.cancel} flips
- * {@link Agent.cancelSignal} and tools must observe it cooperatively to stop early.
+ * {@link Agent.cancelSignal}, which SDK-managed tool contexts expose as `context.cancelSignal`.
  * In concurrent mode, prompt batch-wide cancellation requires every in-flight tool
  * to honor the signal.
  */
@@ -972,7 +972,7 @@ export class Agent implements LocalAgent, InvokableAgent {
   /**
    * The cancellation signal for the current invocation.
    *
-   * Tools can pass this to cancellable operations (e.g., `fetch(url, { signal: agent.cancelSignal })`).
+   * SDK-managed tool contexts receive this as `context.cancelSignal` for cancellable operations.
    * Hooks can check `event.agent.cancelSignal.aborted` to detect cancellation.
    */
   get cancelSignal(): AbortSignal {
@@ -989,7 +989,7 @@ export class Agent implements LocalAgent, InvokableAgent {
    * - At the top of each agent loop cycle
    *
    * If a tool is already executing, it will run to completion unless
-   * the tool checks {@link LocalAgent.cancelSignal | cancelSignal} internally.
+   * the tool checks `context.cancelSignal` internally.
    *
    * Hook callbacks can check `event.agent.cancelSignal.aborted` to detect
    * cancellation and adjust their behavior accordingly.
@@ -1726,7 +1726,9 @@ export class Agent implements LocalAgent, InvokableAgent {
           role: 'assistant',
           content: [new TextBlock('Cancelled by user')],
         })
-        yield this._appendMessage(cancelMessage, invocationState)
+        if (this._hasOpenUserTurn()) {
+          yield this._appendMessage(cancelMessage, invocationState)
+        }
 
         result = new AgentResult({
           stopReason: 'cancelled',
@@ -1756,14 +1758,16 @@ export class Agent implements LocalAgent, InvokableAgent {
       throw error
     } finally {
       // If cancelled but the catch block was bypassed (generator terminated
-      // via .return() when the consumer breaks out of for-await), append an
-      // assistant message so the agent can be reinvoked with a new user prompt.
+      // via .return() when the consumer breaks out of for-await), close an
+      // existing user turn so the agent can be reinvoked.
       if (!caughtError && !result && this.isCancelled) {
         const cancelMessage = new Message({
           role: 'assistant',
           content: [new TextBlock('Cancelled by user')],
         })
-        yield this._appendMessage(cancelMessage, invocationState)
+        if (this._hasOpenUserTurn()) {
+          yield this._appendMessage(cancelMessage, invocationState)
+        }
       }
 
       this._tracer.endAgentSpan(agentSpan, {
@@ -2284,6 +2288,7 @@ export class Agent implements LocalAgent, InvokableAgent {
             middlewareRegistry: this._middlewareRegistry,
             tracer: this._tracer,
             meter: this._meter,
+            cancelSignal: this._abortSignal,
           },
           {
             toolUseBlocks,
@@ -2434,6 +2439,11 @@ export class Agent implements LocalAgent, InvokableAgent {
   private _appendMessage(message: Message, invocationState: InvocationState): MessageAddedEvent {
     this.messages.push(message)
     return new MessageAddedEvent({ agent: this, message, invocationState })
+  }
+
+  private _hasOpenUserTurn(): boolean {
+    const lastMessage = this.messages[this.messages.length - 1]
+    return lastMessage?.role === 'user'
   }
 }
 
--- a/strands-ts/src/tools/executors/__tests__/executor.cancel-signal.test.ts
+++ b/strands-ts/src/tools/executors/__tests__/executor.cancel-signal.test.ts
@@ -1,0 +1,156 @@
+import { describe, expect, it } from 'vitest'
+
+import { Agent } from '../../../agent/agent.js'
+import { MockMessageModel } from '../../../__fixtures__/mock-message-model.js'
+import { createMockTool } from '../../../__fixtures__/tool-helpers.js'
+import { AfterToolCallEvent } from '../../../hooks/events.js'
+import { ExecuteToolStage } from '../../../middleware/stages.js'
+import { MiddlewareRegistry } from '../../../middleware/registry.js'
+import { Meter } from '../../../telemetry/meter.js'
+import { Tracer } from '../../../telemetry/tracer.js'
+import { Message, TextBlock, ToolResultBlock, ToolUseBlock } from '../../../types/messages.js'
+import { SequentialToolExecutor } from '../sequential.js'
+
+import type { ToolExecutorOptions } from '../executor.js'
+import type { AgentStreamEvent } from '../../../types/agent.js'
+
+function createOptions(
+  agent: Agent,
+  cancelSignal: AbortSignal,
+  middlewareRegistry = new MiddlewareRegistry()
+): ToolExecutorOptions {
+  return {
+    agent,
+    middlewareRegistry,
+    tracer: new Tracer(),
+    meter: new Meter(),
+    cancelSignal,
+  }
+}
+
+async function runExecutor(
+  executor: SequentialToolExecutor,
+  options: ToolExecutorOptions,
+  toolUseBlocks: ToolUseBlock[],
+  onEvent?: (event: AgentStreamEvent) => void
+): Promise<ToolResultBlock[]> {
+  const toolResultBlocks: ToolResultBlock[] = []
+  const assistantMessage = new Message({ role: 'assistant', content: toolUseBlocks })
+  const generator = executor.execute(options, {
+    toolUseBlocks,
+    toolResultBlocks,
+    invocationState: {},
+    assistantMessage,
+  })
+
+  let next = await generator.next()
+  while (!next.done) {
+    onEvent?.(next.value)
+    next = await generator.next()
+  }
+
+  return toolResultBlocks
+}
+
+describe('ToolExecutor cancellation signal', () => {
+  it('keeps middleware and tools on the supplied execution signal', async () => {
+    const executionController = new AbortController()
+    const replacementSignal = new AbortController().signal
+    let middlewareSignal: AbortSignal | undefined
+    let toolSignal: AbortSignal | undefined
+    const tool = createMockTool('probe', (context) => {
+      toolSignal = context.cancelSignal
+      return 'ok'
+    })
+    const agent = new Agent({ model: new MockMessageModel(), tools: [tool], printer: false })
+    const middlewareRegistry = new MiddlewareRegistry()
+    middlewareRegistry.add(ExecuteToolStage, async function* (context, next) {
+      middlewareSignal = context.cancelSignal
+      return yield* next({ ...context, cancelSignal: replacementSignal })
+    })
+
+    const results = await runExecutor(
+      new SequentialToolExecutor(),
+      createOptions(agent, executionController.signal, middlewareRegistry),
+      [new ToolUseBlock({ name: 'probe', toolUseId: 'probe-1', input: {} })]
+    )
+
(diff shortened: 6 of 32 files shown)


strands-ts/src/agent/agent-as-tool.ts
```diff
@@ -195,7 +195,10 @@ export class AgentAsTool extends Tool {
       // Stream the sub-agent, forwarding the outer invocation's state so
       // mutations in the inner agent's hooks/tools are visible to the outer
       // agent's downstream callbacks and final AgentResult.
-      const gen = this._agent.stream(input, { invocationState: toolContext.invocationState })
+      const gen = this._agent.stream(input, {
+        invocationState: toolContext.invocationState,
+        cancelSignal: toolContext.cancelSignal ?? toolContext.agent.cancelSignal,
```
strandly-the-agent: 🔴 **A cancelled sub-agent comes back as a *successful* tool result — and under `delegate: true`, its `"Cancelled by user"` text becomes the user's final answer.**

Forwarding the signal here is the PR's one behavioral delta (main forwarded none), but the result mapping at `:213-228` is unchanged and converts **every** `AgentResult` to `status: 'success'` via `result.toString()`. A cancelled sub-agent has `stopReason: 'cancelled'` and `lastMessage = "Cancelled by user"` (`agent.ts:1721-1738`), so:

- `delegate: true` → `AgentDelegation` promotes it to the turn's final content: the outer result is `stopReason: 'endTurn'` with answer `"Cancelled by user"` — app code checking `result.stopReason === 'cancelled'` cannot detect it.
- non-delegate → outer history records `{"status":"success","content":[{"text":"Cancelled by user"}]}` for a sub-agent that never answered, which the model reads back on resume-after-cancel (a supported pattern, `agent.cancel.test.ts:269-312`).

Reachable via `someAgent.asTool({ delegate: true })` (documented public API) + `agent.cancel()` at any moment while the sub-agent runs — i.e. a stop button. A/B-verified branch-only against a build of merge-base `8c3b2102b`.

<details><summary>Repro + A/B output</summary>

```js
const inner = new Agent({ model: mock, name: 'researcher', description: 'd', printer: false })
const outer = new Agent({ model: outerMock, tools: [inner.asTool({ delegate: true, preserveContext: true })], printer: false })
inner.addHook(BeforeModelCallEvent, () => outer.cancel()) // sub-agent already running; hook only makes timing deterministic
const res = await outer.invoke('go')
```
```
branch : stopReason: endTurn | final answer: "Cancelled by user"
main   : stopReason: endTurn | final answer: "inner answer"
```
</details>

Fix shape — branch on the sub-agent's `stopReason` before building the success block (`'cancelled'` is in the `StopReason` union, `types/messages.ts:709`; string form matches the existing `createErrorResult` call at `:183`):

```ts
if (result.stopReason === 'cancelled') {
  return createErrorResult(`Agent '${this.name}' cancelled`, toolUseId)
}
```

(The blanket-success mapping also swallows `maxTokens`/`interrupt` — pre-existing for those; only `'cancelled'` becomes reachable via this PR.)

strands-ts/src/agent/agent-as-tool.ts
```diff
@@ -195,7 +195,10 @@ export class AgentAsTool extends Tool {
       // Stream the sub-agent, forwarding the outer invocation's state so
       // mutations in the inner agent's hooks/tools are visible to the outer
       // agent's downstream callbacks and final AgentResult.
-      const gen = this._agent.stream(input, { invocationState: toolContext.invocationState })
+      const gen = this._agent.stream(input, {
```
strandly-the-agent, line 198: 🟡 **An already-aborted signal at entry corrupts a `preserveContext: true` sub-agent's own history.**

`_throwIfCancelled()` runs at the top of the agent loop (`agent.ts:1485`) *before* the input user message is appended (`:1509-1515`), so when the forwarded signal is already aborted as the sub-agent's loop starts, the cancel handler (`agent.ts:1721-1731`) appends `assistant("Cancelled by user")` with no matching user message:

```
branch : inner roles: [ 'assistant' ]           content: ["Cancelled by user"]
main   : inner roles: [ 'user', 'assistant' ]   content: ["first question","inner answer 1"]
```

With `preserveContext: true` this persists (a `SessionManager` would save it), and the **next** delegation sends `[ 'assistant', 'user', 'assistant' ]` — a shape this repo documents as invalid (`sliding-window-conversation-manager.ts:485-491`; providers reject assistant-first history — not verified live). Default `preserveContext: false` is saved by the `loadSnapshot` at `:191-193`.

The window is real under the default executor: `ConcurrentToolExecutor` primes every tool generator up front (`concurrent.ts:70-74`) and never re-checks, so an abort landing during the sub-agent's startup reaches its loop already aborted (`SequentialToolExecutor` is immune via `sequential.ts:51`).

Cheapest fix at this PR's level: resolve the signal once and, if already `aborted`, return an error result **without entering the sub-agent** — which also covers the 🔴 above for this case. The underlying cancel-before-append ordering in `agent.ts` is pre-existing and reachable directly via `InvokeOptions.cancelSignal` — issue-ready writeup in the review appendix (I couldn't file it: the API token here can't create issues).

strands-ts/src/agent/__tests__/agent-as-tool.test.ts
```diff
@@ -132,6 +132,28 @@ describe('AgentAsTool', () => {
       )
     })
 
+    it('forwards the tool execution signal to the wrapped agent', async () => {
```
strandly-the-agent: 🟡 **The PR's only new behavior is pinned by argument-passing alone.**

`expect(streamSpy).toHaveBeenCalledWith(...)` proves the signal is *handed over*; nothing ever aborts it, so nothing proves the inner agent stops — and nothing covers entry with an already-aborted signal. That's exactly the gap hiding the two defects flagged on `agent-as-tool.ts`: all 296 tests in the affected files pass with both present. Two cases would pin both:

- abort mid-inner-stream → assert the inner loop terminates **and** assert the resulting `ToolResultBlock` (status + content), not just that `stream` was called;
- pre-aborted signal at entry → assert the result block shape and that the inner agent's `messages` stay user-first.

For contrast, the rest of the new coverage is load-bearing: reverting the signal source at `executor.ts:259`, `:203`, `sequential.ts:51`, or `:148` each fails a test in `executor.cancel-signal.test.ts`. This file is the outlier.

strands-ts/src/tools/executors/executor.ts
```diff
@@ -253,6 +256,7 @@ export abstract class ToolExecutor {
           },
           agent: options.agent,
           invocationState,
+          cancelSignal: options.cancelSignal,
```
strandly-the-agent, line 259: 🟡 **A middleware that substitutes `cancelSignal` is silently ignored.**

The terminal callback at `:219-224` forwards `middlewareContext.tool`, `.toolUse`, and `.invocationState`, but both the middleware context (`:203`) and this tool context read `cancelSignal` from `options`. So `next({ ...context, cancelSignal: tighter })` — the documented substitution pattern, exercised for `toolUse` at `agent-middleware.test.ts:1022-1031` — type-checks and no-ops while substituting its neighbours works. Silent no-op is the worst failure mode: no error, no warning.

It matters because **narrowing** is the single most obvious middleware use of a cancel signal: a per-tool-timeout middleware substituting `AbortSignal.any([ctx.cancelSignal, AbortSignal.timeout(n)])` compiles and does nothing today. For calibration, `agent` is options-sourced too (`:199` vs `:258`) and always has been — "identity fields come from options" is a defensible contract — but nothing says so, and a `readonly` optional field on a substitutable context invites the attempt.

Would you rather honour it (thread `middlewareContext.cancelSignal` into `_executeToolCore`) or reject it (state in the `stages.ts:113` TSDoc that substitution has no effect, like `agent`'s escape-hatch note at `:106`)? Either way, one test pinning the chosen contract would stop a future "consistency fix" from silently handing middleware control of tool lifetimes — it's currently untested in both directions.

strands-ts/src/tools/tool.ts
```diff
@@ -33,6 +33,9 @@ export interface ToolContext extends Interruptible {
    * deep-copied on read/write.
    */
   invocationState: InvocationState
+
+  /** Cancellation signal scoped to this tool execution; SDK-managed foreground tools receive the Agent signal. */
```
strandly-the-agent: 🟡 **This TSDoc describes today's implementation, not the contract that has to survive #3632.** (Distinct from the wording-drift thread — this is about what the contract promises.)

1. *"receive the Agent signal"* tells the reader "this is just `agent.cancelSignal`" — the exact conclusion that keeps them on the old field, against the PR's decoupling goal.
2. The field's most surprising property is undocumented: it's a **snapshot**. Captured once per `execute()` (`agent.ts:2285`) while `_abortSignal` is replaced per invocation (`:1097-1100`, `:1292-1294`) — a tool that stores this and polls later tracks the *finished* execution, unlike `agent.cancelSignal`. That's the intended semantics; it needs saying.
3. Scope mismatch: `executor.ts:34` says "this executor invocation", this says "this tool execution" — one signal object is shared by the whole batch today (`:203`/`:259` both read `options`). If #3632 hands out per-task signals, readers of the per-tool wording will have built on the wrong guarantee — say which one it is.

Also: "foreground" is undefined — it appears nowhere else in `site/` or `.agents/references/terminology.md`, and every SDK-built context gets the field, including direct `agent.tool.x()` calls (`tool-caller.ts:230`). Same term in `agent/agent.ts:144` and `:975`; `:144` also dropped the "must observe it cooperatively" requirement the old wording carried.

```suggestion
  /** Cancellation signal for this tool call, captured when the executor starts the batch — it does not track later agent invocations. Prefer this over `agent.cancelSignal`; SDK-managed contexts always provide it. */
```

site/src/content/docs/user-guide/concepts/tools/executors.mdx
```diff
@@ -119,7 +119,7 @@ Cancellation works identically in both modes. Call `agent.cancel()` to request c
 <Tab label="TypeScript">
 
 - **Pre-launch cancel**: set `BeforeToolsEvent.cancel` on the batch-level hook, or call `agent.cancel()` before tools start, to produce error results for every tool in the batch.
-- **Mid-flight cancel** in sequential mode short-circuits not-yet-started tools. In concurrent mode, all tools have already launched, so each in-flight tool must cooperatively observe `agent.cancelSignal` to stop early.
+- **Mid-flight cancel** in sequential mode short-circuits not-yet-started tools. In concurrent mode, all tools have already launched, so each in-flight tool must cooperatively observe `context.cancelSignal` (falling back to `context.agent.cancelSignal` for manually constructed contexts) to stop early.
```
strandly-the-agent: 🟡 **"manually constructed contexts" leaks an internal contract into user docs.** Readers of this page configure `toolExecutor`; they never construct a `ToolContext`, so the parenthetical is noise that also teaches the `??` idiom this PR is trying to retire.

```suggestion
- **Mid-flight cancel** in sequential mode short-circuits not-yet-started tools. In concurrent mode, all tools have already launched, so each in-flight tool must cooperatively observe `context.cancelSignal` to stop early.
```

Same leak in `strands-ts/src/vended-tools/sleep/README.md:5` (suggest: "It attaches a one-shot listener to `context.cancelSignal`, so cancelling the agent aborts the sleep immediately…"). And the new field currently appears in `site/` *only* inside this parenthetical — one sentence naming `context.cancelSignal` in `custom-tools.mdx` (where tool authors actually learn what `context` offers) and at `agent-loop.mdx:183` would land the guidance where readers look.

strandly-the-agent (changes requested):
**Verdict: changes-requested** — the one behavioral delta in this "no behavior change" PR (forwarding the cancel signal into `AgentAsTool`) makes a cancelled sub-agent come back as a **successful** tool result whose text is `"Cancelled by user"`; under `delegate: true` that becomes the user's final answer with `stopReason: 'endTurn'` (🔴 inline, with repro, A/B-verified against a merge-base build).

**API process:** two publicly exported interfaces gain a field with no `api/needs-review` label — +1 to the existing thread, with one addition: #3632 already carries that label and declares this same field **required**, so the optionality decision is best made once, there, *before* this merges. Landing optional+`??` first quietly pre-decides it: the `??` idiom ships in the two canonical TSDoc examples tool authors copy.

**Verified:** ✅ head `b1b45d86` vs merge-base `8c3b2102b` (30 files, +311/−39) · ✅ 452/452 tests in the 15 touched test files (`vitest run --project unit-node`) · ✅ `tsc --noEmit` (src) clean · ✅ every branch-only correctness claim A/B-verified by running the same repro against builds of both trees · ✅ adversarial repros re-run against main's files (their findings are pre-existing → appendix). 🔴 Not verified: browser suite (no chromium in my sandbox — the new `.test.ts` is browser-safe by inspection only); live-provider rejection of assistant-first history (reasoned from `sliding-window-conversation-manager.ts:485-491`).

**Questions**
1. *(blocking)* Was the `AgentAsTool` forwarding meant to land in this PR at all? It's the only behavioral delta — moving it (plus cancelled-result handling) to #3632 would make this PR genuinely inert, matching its stated scope.
2. *(blocking)* What should a cancelled sub-agent produce — an error tool result, or an outer `stopReason: 'cancelled'` even under delegation? (Delegation masking a cancel is pre-existing; the `"Cancelled by user"`-as-answer damage is new.)
3. *(non-blocking)* Is required-now actually off the table? Widening `cancelSignal?: T` → `cancelSignal: T` in #3632 is the same compile break for context-constructors whenever it happens — cheapest now, while the field has zero released readers. If optional-forever is the decision instead, stating it (and keeping the helper from the other thread as permanent API) beats arriving there by default.
4. *(non-blocking)* Should middleware be able to narrow a tool's cancel lifetime (inline at `executor.ts:259`)? Relatedly, should `ConcurrentToolExecutor` check the execution signal between `next()` steps once background lifetimes exist? It never has — cooperation is enough today.

<details><summary>Per-pass breakdown · appendix — non-blocking (13) · process notes</summary>

**Pass verdicts:** correctness **changes-requested** (2 A/B-verified defects, 5 leads refuted) · API bar-raiser **needs-changes** (moderate scale → informal API-reviewer chat + label; no full meeting) · adversarial **survived** (nothing broken *by this PR*; pre-existing edges below) · test-quality **gaps-found** (new executor test file is load-bearing — reverting `executor.ts:259`/`:203`, `sequential.ts:51`, or `executor.ts:148` each fails a test; the `AgentAsTool` spy test is the outlier) · docs-accuracy **docs-issues** (no wrong claim; wording findings inline + below).

**Nits (⚪):**
- `executor.ts:127` — surviving mutant: no test drives hook-cancel + `retry = true` + aborted execution signal; reverting to `options.agent.cancelSignal` stays green. Harmless today, matters once #3632 supplies a divergent signal (~10 lines reusing `createOptions`/`runExecutor`).
- `agent.concurrent.test.ts:60` — `ctx.cancelSignal ?? ctx.agent.cancelSignal` in `GatedTool` made the batch-cancellation test source-agnostic; `ctx.cancelSignal!` would pin the concurrent path for free.
- `src/__fixtures__/tool-helpers.ts:27,36,40` — fixture assigns the **same** signal to `context.cancelSignal` and `agent.cancelSignal`; blinds nothing today, but two separate controllers would make future fixture-based tests catch source-of-signal regressions.
- `executor.cancel-signal.test.ts:151-153` — three scalar asserts where `TESTING.md:218-271` asks for one whole-object `toEqual` (tests 1–2 comply).
- `http-request.test.ts:237-275` (new in this PR) — the fetch mock rejects with a fabricated `DOMException('AbortError')` instead of `signal.reason`, so reason propagation is untested; rejecting with `signal.reason` matches real undici.
- `make-sleep.ts:3-5,37-39` — three invented names ("the execution's `AbortSignal`", "the execution signal", "the Agent signal") for two concrete fields; `types/agent.ts:326` vs `:338` name the same value `cancelSignal` then `signal` — pick one. Same "manually constructed contexts" leak as the mdx inline appears in `sleep/README.md:5`; `custom-tools.mdx` and `agent-loop.mdx:183` never name `context.cancelSignal` at all, which is where tool authors would look.
- PR checklist: both docs boxes unchecked though `site/` changes are included — tick or explain.

**Pre-existing (would still be true with this PR closed — issue-ready; I attempted to file both but the API token here can't create issues, so lifting these two blocks into the bug template is a copy-paste):**
- ⭐ `agent.ts:1485` `_throwIfCancelled()` precedes the user-message append at `:1509-1515`, so `await agent.invoke('x', { cancelSignal: AbortSignal.abort() })` appends `assistant("Cancelled by user")` with **no user message** → next invocation sends assistant-first history, violating the invariant at `agent.ts:1062-1065` (and `sliding-window-conversation-manager.ts:485-491`). Reachable via the documented `InvokeOptions.cancelSignal` (`types/agent.ts:124-136` shows `AbortSignal.timeout(5000)` — an elapsed timeout hits it). A/B-identical on merge-base. Root cause under the 🟡 inline on `agent-as-tool.ts:198`.
- Non-`Error` abort reasons degrade in vended tools (identical on main): `make-sleep.ts:74-76` rethrows the raw reason → `normalizeError` (`errors.ts:145`) renders `"Error: [object Object]"` to the model; `http-request.ts:80-85` labels only `name === 'AbortError'`, but real fetch rejects with `signal.reason` → an `Error` reason loses the "Request cancelled" context, a string reason is thrown raw (not an `Error`).
- `make-sleep.ts:98-102` — the "re-check after attaching" branch is unreachable (no task boundary between the pre-flight check at `:73-76` and the attach; `Promise` executor bodies run synchronously), yet its comment asserts the opposite.

**Checked and clean (refutations, so nobody re-chases them):** snapshot-vs-live is **not** a divergence within a batch — `_abortSignal` is assigned only at `agent.ts:1097-1100` and `:1293-1294`, and `executeTools` runs strictly between them, pinned by `agent.cancel.test.ts:441` · required `ToolExecutorOptions.cancelSignal` is no runtime hole (`@internal`, single construction site `agent.ts:2280-2286`, module not in `exports`) · `ToolCaller` signal not stale (live getter read; no delta vs main) · middleware doesn't clone the signal away · retry suppression records the **real** tool result (ran once, no retry, `stopReason: 'cancelled'`) · sleep listener lifecycle leak-free (12 sequential → 0 listeners; 15 concurrent peak → 0) · `mcp-tool` already-aborted at entry → clean error result via the MCP SDK's `throwIfAborted()`, no unhandled rejection · `http-request` already-aborted execution signal labels "Request cancelled", not "timed out", and `AbortSignal.any` propagates reasons through real fetch · docs claims all check out (every SDK path sets the field; `context?.cancelSignal ?? context?.agent.cancelSignal` parses as intended; site snippets type-check against the branch; no stale primary-signal guidance remains).

**Suppressed as already-raised** (existing `github-actions` threads — not re-filed): the `resolveCancelSignal()` helper for the 4× duplicated `??`; the missing `api/needs-review` label + optional-vs-always-populated tension + `tool.ts:37`↔`stages.ts:113` wording drift. The `tool.ts:37` inline below adds only new substance (snapshot lifetime, batch-vs-per-tool scope, undefined "foreground"); the sequencing point above adds only the #3632 ordering.

**Process:** review ran as five independent fresh-context passes (correctness, API bar-raiser, adversarial, test-quality, docs-accuracy) + an independent aggregation pass. The API and adversarial passes first ran on a deeper-analysis tier and timed out at the harness's 900s ceiling; both were re-run successfully on the standard tier with narrowed scope. As always: solid work for a human to approve, not gospel.
</details>

strands-ts/src/tools/mcp-tool.ts
```diff
@@ -51,7 +51,7 @@ export class McpTool extends Tool {
 
     try {
       const rawResult: unknown = await this.mcpClient.callTool(this, input as JSONValue, {
-        signal: toolContext.agent.cancelSignal,
+        signal: toolContext.cancelSignal ?? toolContext.agent.cancelSignal,
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Updated `cancelSignal` to be required on ToolContext, all fallbacks are now removed. Resolved.

strands-ts/src/tools/tool.ts
```diff
@@ -33,6 +33,9 @@ export interface ToolContext extends Interruptible {
...
+  cancelSignal?: AbortSignal
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Discussed offline, decided to make `cancelSignal` required on ToolContext. Updated TSDoc accordingly, removed fallback behavior.

site/src/content/docs/user-guide/concepts/tools/executors.mdx
gautamsirdeshmukh: Removed the manually-constructed-context guidance

strands-ts/src/tools/tool.ts
gautamsirdeshmukh: Updated to describe it as an execution-scoped cancellation signal, removed wording about Agent/foreground

strands-ts/src/agent/__tests__/agent-as-tool.test.ts
gautamsirdeshmukh: Replaced the argument-passing test with behavioral coverage for cancellation during child execution and an already-aborted signal at entry

strands-ts/src/agent/agent-as-tool.ts
gautamsirdeshmukh: Fixed, `AgentAsTool` now checks the execution signal before invoking the child Agent and returns an error result immediately when it is already aborted

strands-ts/src/agent/agent-as-tool.ts
gautamsirdeshmukh: Fixed, `AgentAsTool` now converts a child result with `stopReason: cancelled` into an error tool result, preventing false reporting as a success

strands-ts/src/tools/executors/executor.ts
gautamsirdeshmukh: The tool and executor need to use the same cancellation signal. If middleware replaced it only for the tool, the tool could stop while the executor still considers/assumes the overall execution active and continues processing. So, middleware can read the cancel signal but not replace it (it can still prevent a tool from being run anyway). Added documentation/tests accordingly.

strandly-the-agent (approved):
**Verdict: approve** — all six findings from my previous review are fixed, verified against head `3907f316`, and one fix is better than what I asked for: instead of the `AgentAsTool` pre-abort check I suggested, the cancel-message append in `agent.ts` is now guarded on an open user turn (`agent.ts:1729-1731` and the `finally` variant) — which fixes the root cause, including the pre-existing cancel-at-entry bug from my appendix (`invoke()` with an already-aborted signal now leaves `messages` untouched, pinned by the updated `agent.cancel.test.ts:34`).

**Previous findings — status:**
1. 🔴 cancelled sub-agent → success result: **fixed** (`agent-as-tool.ts:216-218` maps `stopReason: 'cancelled'` → error result). I checked the delegate path specifically: `AgentDelegation` clears its mark on error results (`agent-delegation.ts:190-193`) and re-verifies in `_onAfterTools` (`:219`), so no `endTurn` promotion — the outer loop continues and an outer cancel surfaces as `stopReason: 'cancelled'`.
2. 🟡 pre-aborted entry corrupts `preserveContext` history: **fixed at the root** (above). Attacked the guard at the other cancel timings — mid-model-stream (deferred append ⇒ last is `user` ⇒ cancel message appended) and post-tool-results (last is the `user` toolResult message ⇒ appended) both keep alternation; the skipped-append case still returns a consistent `AgentResult` (`lastMessage` is the cancel message object, `agent.ts:1733-1735`).
3. 🟡 spy-only test: **fixed** — replaced with behavioral coverage (mid-run cancel → asserts the error `ToolResultBlock`; abort during `initAgent` with `preserveContext: true` → asserts error result **and** `messages: []`, `agent-as-tool.test.ts:137-196`).
4. 🟡 middleware-substituted `cancelSignal`: **decided and pinned** — executor-owned/read-only, documented (`stages.ts:113-114`) and tested (substituting a replacement signal is asserted ignored, `executor.cancel-signal.test.ts:56-88`). The rationale in the thread (tool stopping while the executor considers the execution live) is sound.
5. 🟡 TSDoc contract: **addressed** — "execution-scoped" wording, "foreground"/"Agent signal" phrasing removed across `agent.ts:144/975/992`.
6. 🟡 docs leak: **fixed** — `executors.mdx:122-124` parenthetical gone; `sleep/README.md:5` names `context.cancelSignal`.

The delta also lands the bigger call from the threads: **`cancelSignal` is now required** on `ToolContext` (`tool.ts:37-38`) and `ExecuteToolContext` (`stages.ts:113-114`), with every `??` fallback removed — I grepped `src/` and `site/`: zero remain, TSDoc examples updated (`types/agent.ts:327,337`), and `GatedTool` now pins the execution signal directly. That's the shape #3632 declares, taken while the field has zero released readers — the cheapest time to take it.

**Verified:** ✅ `npm ci` + 514/514 tests across the 16 touched test files (`vitest run --project unit-node`, 10.9s) · ✅ `tsc --noEmit` (src) clean · ✅ each fix checked against the diff, not the commit message · ✅ fix-attack pass on the new `agent.ts` guard and the delegation error path found nothing. Ran as a self-run follow-up verification (fixes-only delta), no specialist fan-out.

<details><summary>Non-blocking residuals (3)</summary>

- **Label trail:** the optional→required decision was made offline per the thread — if that was with an API reviewer, adding `api/needs-review` + `api/review-complete` would let the CI gate record it. Maintainer's call; not re-arguing the decision itself.
- **Snapshot lifetime still undocumented** (optional): `context.cancelSignal` is captured per `execute()` while `_abortSignal` is replaced per invocation — a tool that stores it and polls later tracks the finished execution. The new wording no longer implies otherwise, so this is a nice-to-have, not drift.
- The pre-existing abort-reason degradations (`make-sleep.ts` `"[object Object]"`, `http-request.ts` labeling) from my previous appendix remain open and untouched by this PR — still issue-ready there; the token here still can't file issues.
</details>

strands-ts/src/agent/agent-as-tool.ts
```diff
@@ -186,6 +186,7 @@ export class AgentAsTool extends Tool {
     this._busy = true
     try {
       const { input } = toolUse.input as { input: string }
+      const { cancelSignal } = toolContext
```
liramon2: Nit: this is inconsistent with how `invocationState` is accessed. You can inline this as well or add `invocationState` to the destructuring. If the destructuring doesn't need to be in the try block, you could also combine this with `const { toolUse } = toolContext` above.

site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
```diff
@@ -180,7 +180,7 @@ The `agent.cancel()` method provides a way to stop the loop from outside, such a
 
   #### Cancellation within tool execution
```
liramon2, line 181: `strands-ts/AGENTS.md` line 211 still references `agent.cancelSignal`, but not `context.cancelSignal`. Might be worth adding.

site/src/content/docs/user-guide/concepts/tools/executors.mdx
```diff
@@ -119,7 +119,7 @@ Cancellation works identically in both modes. Call `agent.cancel()` to request c
 <Tab label="TypeScript">
 
 - **Pre-launch cancel**: set `BeforeToolsEvent.cancel` on the batch-level hook, or call `agent.cancel()` before tools start, to produce error results for every tool in the batch.
-- **Mid-flight cancel** in sequential mode short-circuits not-yet-started tools. In concurrent mode, all tools have already launched, so each in-flight tool must cooperatively observe `agent.cancelSignal` to stop early.
+- **Mid-flight cancel** in sequential mode short-circuits not-yet-started tools. In concurrent mode, all tools have already launched, so each in-flight tool must cooperatively observe `context.cancelSignal` to stop early.
```
mkmeral, line 122: nit: `toolContext.cancelSignal`.I don't think we tell users what context this is, it'd be nice to just be a bit more explicit

strands-ts/src/tools/executors/__tests__/executor.cancel-signal.test.ts
```diff
@@ -0,0 +1,156 @@
+import { describe, expect, it } from 'vitest'
```
mkmeral, line 1: nit: file is too specific, should it just be part of other executor tests?

strands-ts/src/agent/agent.ts
```diff
@@ -2433,6 +2438,11 @@ export class Agent implements LocalAgent, InvokableAgent {
     this.messages.push(message)
     return new MessageAddedEvent({ agent: this, message, invocationState })
   }
+
+  private _hasOpenUserTurn(): boolean {
```
mkmeral, line 2444: wouldn't we want a cancellation message regardless? I mean the task is explicitly cancelled, right? 🤔

gautamsirdeshmukh: Callers always receive a cancellation message (`AgentResult.lastMessage` is always "Cancelled by user"). This guard only controls whether that synthetic assistant message is persisted in conversation history. If cancellation occurs before the invocation’s user message is appended, persisting it would create assistant-first history which breaks the contract

site/src/content/docs/user-guide/concepts/tools/executors.mdx
gautamsirdeshmukh: Agreed, updating to `toolContext.cancelSignal`

strands-ts/src/agent/agent-as-tool.ts
gautamsirdeshmukh: Good callout, updated

strands-ts/src/tools/executors/__tests__/executor.cancel-signal.test.ts
gautamsirdeshmukh: Renaming to `executor.test.ts` (there actually isn't any other executor test file)

site/src/content/docs/user-guide/concepts/agents/agent-loop.mdx
gautamsirdeshmukh: Updated to mention `ToolContext.cancelSignal`

liramon2 (approved):

mkmeral (approved):

Issue #3042, opened 2026-07-01
[FEATURE] Background Tasks — Implementation


yonib05:
## Summary

Implement async / background tasks in the agent loop, following the Background Tasks proposal (#2496) and the POC. Tool calls can be backgrounded and run concurrently without blocking the main event loop; the agent is notified/invoked when a backgrounded call completes. Align with the MCP spec for async tool execution.

Sub-issue of #2496.

## Deliverables

- [ ] Async tool invocation API (dispatch to background, obtain a handle).
- [ ] Progress/status update callbacks (coordinate with typed events, #242).
- [ ] Task cancellation and timeout handling.
- [ ] Async-aware event-loop integration (resume reasoning when a background result arrives).
- [ ] Alignment with MCP async patterns (#1812).

## Correctness hazards to account for

Concurrency in the loop already has sharp edges — the background-task work must not regress these, and should add coverage:
- #3021 — invocation lock has no ownership tracking (`release_lock()` is owner-blind).
- #2266 — `BedrockModel.stream` leaks inner task on outer cancellation.
- #2112 — `ConcurrentToolExecutor` collected tool results in completion order, breaking pairing.

Related: #1388 (async support in ConversationManager), #2234 (async SessionRepository / non-blocking session writes), #457 (tool consent in async frameworks), #2156 (`agent.cancel()` noop when idle), #2771 (`background=True` + stream resume on OpenAI Responses).

Note: Python's synchronous entrypoint is bridged to async via `run_async` (`strands-py`); the background-task model must work through that bridge as well as native TS async.

## Code pointers (pinned to `6ab71c6`)

- Tool executors to extend: TS [agent.ts#L2261](https://github.com/strands-agents/harness-sdk/blob/6ab71c61fa4e7a475b2723d6e369ebd948276f67/strands-ts/src/agent/agent.ts#L2261); Python [concurrent.py#L19](https://github.com/strands-agents/harness-sdk/blob/6ab71c61fa4e7a475b2723d6e369ebd948276f67/strands-py/src/strands/tools/executors/concurrent.py#L19).
[...]

Issue #3021, opened 2026-06-30
Invocation lock has no ownership tracking: release_lock() is owner-blind (root cause of #2918 only partially addressed by #2954)


strandly-the-agent:
**Learning** (run learning-run-pr2954, from #2954)

The per-Agent invocation lock in `_ConcurrencyController` has no ownership tracking. `release_lock()` releases the lock whenever it is held, regardless of which invocation acquired it. PR #2954 fixed only the `stream_async` symptom (by gating its `finally` release on `mode == THROW`), leaving the owner-blind primitive — the actual root cause described in #2918 — in place.

**Details**

`_ConcurrencyController.release_lock()` releases `_invocation_lock` whenever `_invocation_lock.locked()` is true, with no check that the *calling* invocation is the one that acquired it. This means correctness depends entirely on every caller self-disciplining its release. Today the callers disagree on how:

- `tools/_caller.py` does it correctly: it records the real acquisition result (`acquired_lock = ... try_acquire_lock()`) and releases only `if acquired_lock`. Release is conditioned on *proven ownership*.
- `agent/agent.py` `stream_async` (after #2954) conditions its release on `self._concurrency.mode == ConcurrentInvocationMode.THROW` — a *proxy* for ownership, re-deriving the same decision `begin()` already made when it chose whether to acquire.

#2918 suggested the robust fix: gate the release on actual acquisition and have `_ConcurrencyController` *track and enforce* acquire/release pairing. That was not done; #2954 shipped the narrower mode-based gate instead.

A related trap makes the straightforward ownership-based fix surprising: `_BeginResult.lock_acquired` is initialized to `True` and only overwritten in THROW mode (`_concurrency.py:124-128`). So in `UNSAFE_REENTRANT` mode `begin()` returns `lock_acquired=True` even though no lock was acquired. Gating the release on `begin.lock_acquired` (the intuitive fix) would therefore *not* fix the bug — the field name does not mean "this invocation holds the lock" in all modes.

Suggested direction: make acquire/release ownership-aware inside `_ConcurrencyController` (e.g. return an acquisition handle/token that only its owner can release, or at minimum set `lock_acquired=False` in `UNSAFE_REENTRANT` and gate every release on proven acquisition, mirroring the `_caller.py` pattern). This removes the latent footgun for any future caller of `release_lock()` and makes the misleading `lock_acquired` field honest.

**Other affected locations**

[...]

PR #3819, opened 2026-08-14
docs(designs): 0018 — Invocation Queueing (concurrentInvocationMode: 'enqueue')

agent-of-mkmeral:
## Summary

Design proposal for **invocation queueing**: extend the Agent's invocation lock into a mode — `concurrentInvocationMode: 'throw' | 'enqueue'` (TS first, name parity with the existing Python knob) — where `enqueue` turns the lock into a FIFO of **fresh invocations**. A second `invoke()`/`stream()` on a busy agent waits its turn and runs as its own invocation (own `AgentResult`, own hook pair, own fresh `AbortController`) instead of throwing `ConcurrentInvocationError`.

On top of the core mode:
- **Per-call `ifBusy: 'throw' | 'enqueue' | 'interrupt'`** — `interrupt` composes with the existing `agent.cancel()` (control plane stays separate from the data plane; no text heuristics).
- **Introspection**: `agent.pendingInvocations` + `agent.cancelPending(id)` — shape parity with `agent.backgroundTasks`.
- **Mid-run visibility**: a small vended plugin over the existing `ContextInjector` (`trigger: 'everyTurn'`) renders the pending queue into every model call of the running invocation — ephemeral, never in durable history — so the model can wrap up early when a pending request supersedes its current work.
- **Queue-aware completion wait**: the one integration contract with the Background Tasks feature (in review) — skip the lock-held `waitForCompletion` wait when requests are queued ("serve the queue, wait at the end"), so queued requests interleave with background execution instead of starving behind it. Loss-free because delivery is state-driven, not wait-driven.

## Why

Today a second request while busy has exactly one behavior: an exception. Every serious host hand-rolls the same stack in front of the agent (lock → queue → visibility → interruption) and hits the same bugs in the same order. We implemented this exact stack in **strandly**, our autonomous agent, and its two hardest bugs — a drain race at the queue/lock boundary, and a cancelled-turn resume poisoned by the still-armed cancel signal — are both artifacts of building *outside* the lock. Built into the lock, neither can exist.

The design also explains **why fresh invocations, not continuations**: continuations are intra-invocation, die with their invocation (including the poisoned-signal-after-cancel path), and are `@internal` — a queued external request needs its own result and must survive cancellation, because it is frequently the *reason* for the cancel.

## Notes for reviewers

- Doc only — no code changes. Development plan in the doc phases P1 (core mode) → P2 (visibility plugin) → P3 (`ifBusy`/`interrupt`) → P4 (Python parity via `ENQUEUE` on the existing enum).
- Prior art survey (LangGraph multitask strategies, Claude Code's queued-input vs `interrupt()` split, actor mailboxes, Temporal signals) is in Appendix A.
- All load-bearing SDK claims are verified against code (Appendix C); all 5 sequence diagrams validated with mermaid-cli.
- Numbered 0018 to leave room for the in-flight 0017 (subagents) design.

## Related

- #3042 (async/background deliverables), #3021 (owner-blind invocation lock)
- The Background Tasks feature (in review) — §"Interaction with Background Tasks" is the integration contract; the queue-aware wait is ~3 lines in whichever side lands second.

### DIFF
--- a/team/designs/0018-invocation-queueing.md
+++ b/team/designs/0018-invocation-queueing.md
@@ -1,0 +1,311 @@
+# Invocation Queueing: `concurrentInvocationMode: 'enqueue'`
+
+**Status**: Proposed
+
+**Date**: 2026-08-13
+
+**Issue**: TBD
+
+**Scope**: TypeScript SDK first (where the Background Tasks feature lives); Python parity port after, extending the existing `ConcurrentInvocationMode` enum.
+
+**Related**:
+- **Background Tasks** (TS) — design and implementation currently in review; §[Interaction with Background Tasks](#interaction-with-background-tasks) below is the integration contract between the two features
+- [#3042: async/background deliverables](https://github.com/strands-agents/harness-sdk/issues/3042), [#3021: owner-blind invocation lock](https://github.com/strands-agents/harness-sdk/issues/3021) — adjacent hazards; this design adds the missing lock semantics
+
+## Context
+
+An agent serving real traffic gets requests while it is busy. A GitHub bot gets a second mention twenty minutes into a review; a chat UI user double-texts; a Slack command arrives mid-task. Today the SDK's answer is fail-fast: the TS `Agent` throws `ConcurrentInvocationError` from `acquireLock()`, and Python offers `ConcurrentInvocationMode.THROW` (default) or `UNSAFE_REENTRANT` (explicitly unsafe). Every serious host therefore hand-rolls the same thing in front of the agent — a lock, then a queue, then visibility, then interruption — and gets the same bugs in the same order.
+
+We know because we did it: we implemented this exact stack in **strandly**, our autonomous agent — a session mailbox in front of the agent (a state machine with an atomic drain-or-idle transition), a pending-messages plugin inside it (mid-run visibility via `ContextInjector` plus delivery via `AfterInvocationEvent.resume`), and control-plane cancellation via `agent.cancel()`. It works and it is well tested, and its two hardest bugs — a drain race at the queue/lock boundary, and a cancelled-turn resume being poisoned by the still-armed cancel signal — are both artifacts of building *outside* the lock. Built into the lock, neither can exist.
+
+Meanwhile the Background Tasks feature adds tools that run detached from the invocation, with results delivered as continuation passes. Background tasks make the busy-agent problem *more* common (invocations legitimately stay open longer waiting for task results) and simultaneously provide the machinery that makes interleaving solvable. The two features need a designed contract, not an accident of merge order.
+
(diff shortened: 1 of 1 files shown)


team/designs/0018-invocation-queueing.md
```diff
@@ -0,0 +1,311 @@
...
+| `waitForCompletion: false` | **Interleaving works today, zero new code.** Invocation N returns after foreground work → lock released → queued request runs *while the task executes* → result delivered at the queued invocation's next `BeforeModelCall`. |
+| `waitForCompletion: true` (default), as written | Starvation: `_onAfterInvocation` holds the lock across `_waitForTaskResult` and the delivery passes. The queued request waits behind the whole background train. |
+| `waitForCompletion: true` + **queue-aware wait** (this design) | "Serve the queue, wait at the end": skip the wait when someone is queued; the last invocation's wait catches stragglers. |
+
+The fix is the shape of the task manager's existing early-return:
+
+```ts
+if (!this._config.waitForCompletion || …) return
+if (agent.pendingInvocations.length > 0) return   // queue-aware: they'll receive the results
+await this._waitForTaskResult(cannotContinue)
```
mkmeral, line 179: we also need to check if any other messages come while waiting for the task result. otherwise there's a gap that can create starvation

team/designs/0018-invocation-queueing.md
```diff
@@ -0,0 +1,311 @@
...
+
+**Issue**: TBD
+
+**Scope**: TypeScript SDK first (where the Background Tasks feature lives); Python parity port after, extending the existing `ConcurrentInvocationMode` enum.
+
+**Related**:
+- **Background Tasks** (TS) — design and implementation currently in review; §[Interaction with Background Tasks](#interaction-with-background-tasks) below is the integration contract between the two features
+- [#3042: async/background deliverables](https://github.com/strands-agents/harness-sdk/issues/3042), [#3021: owner-blind invocation lock](https://github.com/strands-agents/harness-sdk/issues/3021) — adjacent hazards; this design adds the missing lock semantics
+
+## Context
```
cagataycali, line 15: I would suggest to consider real concurrency instead of queue here.

Experience can be delivered with forking the history from point of time and absorb the history when agent finish

team/designs/0018-invocation-queueing.md
```diff
@@ -0,0 +1,311 @@
+# Invocation Queueing: `concurrentInvocationMode: 'enqueue'`
```
arielnabavian, line 1: Some things we need to discuss 

Would like to see Durable execution and how it fits into this. This only applies to statefully hosted strands agents. I would lean towards a new primitive for queueing with durable or in memory queues.

Issue #3802, opened 2026-08-13
[PORT] Add GuardContent to Python SystemContentBlock


tummon:
### Source SDK

strands-ts

### Target SDK

strands-py

### Related Files

```text
strands-ts/src/types/messages.ts
  strands-ts/src/types/__tests__/messages.test.ts
  strands-ts/src/models/bedrock.ts
  strands-ts/src/models/__tests__/bedrock.test.ts
```

### Confirmation

- [x] The change is complete in the source SDK, and every related artifact is listed above (source, tests, and docs, if the change has any).

PR #3820, opened 2026-08-14
feat(py): support guard content in system prompts and guard images

tummon:
## Description
Bring the Python guard content types to parity with the TypeScript SDK.

SystemContentBlock now accepts guardContent, matching the TypeScript SystemContentBlockData union. Bedrock's Converse API already accepts guardContent in system blocks and the provider forwards system blocks verbatim, so only the type needed widening; providers that do not support guardrails already skip blocks they do not recognize.

GuardContent now accepts an image alongside text, mirroring GuardContentBlockData. The Bedrock formatter read guard["text"] unconditionally, so an image block raised KeyError; the branch moves into _format_guard_content, which formats either member and rejects a block carrying neither.

The guardrail_latest_message wrap now skips images whose source is not raw bytes. GuardrailConverseImageSource accepts only bytes, so wrapping an S3-sourced image built a request Bedrock rejects, failing an otherwise valid message. The image is now sent unwrapped with a warning, matching _applyGuardBlocks in the TypeScript provider.

## Related Issues

#3802 

## Type of Change

New feature

## Testing

## Checklist

----


strands-py/src/strands/types/content.py
```diff
@@ -29,13 +29,43 @@ class GuardContentText(TypedDict):
...
 
+    Only one of `image` or `text` should be specified.
+
     Attributes:
+        image: Image within content block to be evaluated by the guardrail.
         text: Text within content block to be evaluated by the guardrail.
     """
 
+    image: GuardContentImage
     text: GuardContentText
```
kimnamu, line 69: Bedrock rejects an image in a system-prompt guard block no matter the guardrail, format or region (probes in the review comment), so system blocks need a text-only shape. Putting it next to `GuardContent` keeps both variants in one place:

```suggestion
    image: GuardContentImage
    text: GuardContentText


class SystemGuardContent(TypedDict):
    """Content block in a system prompt to be evaluated by guardrails.

    Bedrock rejects an image in a system prompt guard content block, so only text is accepted here.

    Attributes:
        text: Text within content block to be evaluated by the guardrail.
    """

    text: GuardContentText
```

strands-py/src/strands/types/content.py
```diff
@@ -111,10 +141,12 @@ class SystemContentBlock(TypedDict, total=False):
 
     Attributes:
         cachePoint: A cache point configuration to optimize conversation history.
+        guardContent: Contains the content to assess with the guardrail.
         text: A system prompt for the model.
     """
 
     cachePoint: CachePoint
+    guardContent: GuardContent
```
kimnamu, line 149: Then point `SystemContentBlock` at the narrow one, so mypy catches the image variant instead of Bedrock rejecting the request on every turn:

```suggestion
        guardContent: Contains the text to assess with the guardrail.
        text: A system prompt for the model.
    """

    cachePoint: CachePoint
    guardContent: SystemGuardContent
```

This also needs the `model.py:106` loop-variable rename from the review comment to stay mypy-clean. With all three edits `mypy src/strands` matches this branch's baseline exactly (`Found 57 errors in 24 files`) and the bedrock/model/agent/types tests give `610 passed`.

kimnamu (commented):
Thanks for doing this one — guard content in system prompts has been missing for a while, and the bytes-source gate you noticed on the way through is a good catch on its own.

I'm not a maintainer, just a Bedrock user who has spent time in this file. I checked out the branch, ran it against live Bedrock (my own throwaway guardrail, since the image path needs a real one), and re-ran everything cold before writing this. One thing I'd change before merge, then two non-blocking notes.

### What I checked that holds up

- Widening a shared type is the risk here and it's clean. Every non-Bedrock consumer of `system_prompt_content` already key-guards: `anthropic.py:482`, `litellm.py:228`, `openai.py:385-389`, `model.py:107`. Both internal `guardContent` readers are already defensive too (`model.py:78-81` checks both keys, `routing/classifier_strategy.py:238-246` is `isinstance`-guarded). No `KeyError` fallout of the #4018 / #3582 kind.
- The new types match botocore exactly: `GuardrailConverseImageFormat.enum == ['png', 'jpeg']`, `GuardrailConverseImageSource` has only `bytes`, and `GuardrailConverseImageBlock` has no `qualifiers` member. `GuardImageFormat` / `GuardImageSource` / `GuardContentImage` line up 1:1.
- The bytes gate is correct and the image path really works: a message-level `guardContent.image` is accepted live (png and jpeg) with a guardrail whose content filters have `inputModalities: [TEXT, IMAGE]`, and the same bytes as a plain `image` block is accepted as a control.
- Grounding works end to end, which is the interesting part of the system-block half. With a guardrail carrying a contextual grounding policy, the system `grounding_source` block is genuinely consumed: `contextualGroundingPolicyUnits: 1`, `GROUNDING score 0.61 / threshold 0.75 → BLOCKED`, `stopReason: guardrail_intervened`.
- A second bytes-carrying block type doesn't break persistence — `types/session.py:28-34` base64-encodes bytes recursively, so no special-casing was needed.

### 1. `SystemContentBlock` + `guardContent.image` type-checks, but Bedrock rejects it every time

The two type changes together make this legal:

```python
system: list[SystemContentBlock] = [
    {"text": "You are a helpful assistant."},
    {"guardContent": {"image": {"format": "png", "source": {"bytes": png_bytes}}}},
]
```

mypy is clean on that on this branch. On `main` it isn't (`error: Extra key "guardContent" for TypedDict "SystemContentBlock"`), so the combination is new here — each half is fine on its own. Bedrock refuses it unconditionally:

```
ValidationException: An error occurred (ValidationException) when calling the Converse operation:
The guard content field in system prompt cannot contain images. Please remove the image and try again.
```

Single-variable probes against live Bedrock:

| request | result |
|---|---|
| system `guardContent.text`, guardrail set | ACCEPTED |
| system `guardContent.image` png, guardrail set | ValidationException above |
| system `guardContent.image` png, **no** `guardrailConfig` | same error, and it fires *before* the "guardrail configuration is missing" one |
| system list containing only the guard image block | same error |
| system `guardContent.image` **jpeg** (the only other legal format) | same error |
| message `guardContent.image` png / jpeg, guardrail set | ACCEPTED (control) |
| message plain `image` block, identical bytes | ACCEPTED (control) |

Same rejection on `us.amazon.nova-lite-v1:0`, `us.anthropic.claude-haiku-4-5` and `us.anthropic.claude-sonnet-4-5`, in `us-west-2` / `us-east-1` / `eu-west-1`, on `converse` and `converse_stream`. So it isn't a policy gap in my guardrail or a model quirk. botocore's schema does allow it (`SystemContentBlock.guardContent` → `GuardrailConverseContentBlock`, which has an `image` member), so nothing local catches it — the ParamValidator passes and the request goes out.

End to end on this branch with a real `Agent` + `BedrockModel`:

```
[system guardContent.image]         emitted system=[['text'], ['guardContent']]
    -> ValidationException: The guard content field in system prompt cannot contain images.
[system guardContent.text (control)] emitted system=[['text'], ['guardContent']]
    -> OK: 'The capital of France is Paris. If you have any more questio'
```

The fix I'd suggest is the same move you already made with `GuardImageFormat`: narrow the type rather than let the service be the validator. Two inline suggestions below add a text-only `SystemGuardContent` and point `SystemContentBlock` at it.

One more edit is needed outside this diff, because `model.py:106` reuses `block` for a `SystemContentBlock` and then for a `ContentBlock` at `:113`, so mypy pins the variable to the first type. Renaming the loop variable keeps the tree clean and drops the accidental coupling:

```python
    if system_prompt_content:
        for system_block in system_prompt_content:
            if "text" in system_block:
                total += _heuristic_estimate_text(system_block["text"])
```

With all three edits: `mypy src/strands` gives `Found 57 errors in 24 files (checked 272 source files)`, byte-identical to this branch's baseline; `pytest tests/strands/models/test_bedrock.py tests/strands/models/test_model.py tests/strands/agent/test_agent.py tests/strands/types` gives `610 passed`; `ruff format --check` and `ruff check` clean. And the guard bites where it should:

```
# image in a system block, after
content.py: error: Missing key "text" for TypedDict "SystemGuardContent"  [typeddict-item]
content.py: error: Extra key "image" for TypedDict "SystemGuardContent"  [typeddict-unknown-key]

# system guardContent.text + message guardContent.image, after
(clean)
```

The trade-off is yours to weigh: narrowing means a `ContentBlock` value is no longer assignable to a `SystemContentBlock` slot, which is exactly how `model.py:106` surfaced. `one: SystemContentBlock = b` where `b: ContentBlock` passes on this branch and errors after (list-level assignment was already blocked by invariance, so the exposure is element-level only). If you'd rather keep the types 1:1 with botocore, the other route is to warn and skip the block in the Bedrock system path at format time, mirroring what `_format_bedrock_messages` already does for an unsupported guard image format. I lean toward the type because it costs nothing at runtime and this file already encodes guardrail-specific narrowings, but either one closes the hole.

### 2. Not blocking: two things a user following the example will hit

Both are live, both are one docstring line rather than code:

```
# system guardContent block, but no guardrail_id on the model
ValidationException: The guardrail can't assess the content in the guardContent field.
The guardrail configuration is missing. Specify the guardrailConfig field and try again.

# system grounding_source block alone, guardrail has a contextual grounding policy
ValidationException: The provided request does not contain the query. Grounding source,
query and content to guard are required for Guardrails contextual grounding policy evaluation.
```

The second one matters because `qualifiers: ["grounding_source"]` in a system block is the shape your test uses, and on its own it 400s. Adding a `query`-qualified block to the user message makes the request valid and gets the grounding policy actually scoring (that's the trace numbers above). Worth a sentence on the `guardContent` attribute or in the docs PR.

### 3. Not blocking, TypeScript

`strands-ts/src/types/messages.ts:810-811` has the same union (`SystemContentBlockData` includes `{ guardContent: GuardContentBlockData }`, and `GuardContentBlockData` carries `image`), so the same typed-legal-but-rejected combination exists there. Your side is faithful parity, so this looks like a separate follow-up rather than scope creep here.

Happy to re-run any of these probes or share the script if it helps. Other than the system/image combination this reads clean and tightly scoped, and I'd like to see it land.

PR #3821, opened 2026-08-14
feat(mcp): added server/discover for SEP-2575

poshinchen:
## Description

Another work for `mcp` 2.x in the Python SDK. 

`mcp` 2.0 renamed and moved several things we import, and the new spec (SEP-2575) replaced the `initialize` handshake with `server/discover`. Instead of scattering version checks around the codebase, this PR adds one module, `strands.tools.mcp._compat`, that handles the differences. Everything else imports from it and doesn't care which version is installed.

What's in `_compat`:

- `MCP_V2` tells you which major version is installed. It checks `hasattr(ClientSession, "discover")` instead of parsing the package version, so pre-releases and backports still work.
- `initialize_session()` sets up the connection on either version. On 2.x it uses the client's own `negotiate_auto`, which tries `server/discover` first and falls back to the old handshake for older servers. On 1.x it calls `session.initialize()` like before.
- `streamable_http_transport()` keeps the old `(url, headers, auth)` signature on both versions. The 2.x replacement takes a pre-built HTTPX client and only closes clients it created itself, so the wrapper builds the client and closes it when the transport closes. Otherwise we'd leak connections. Follow-up to #3708.
- Plain renames: `MCPError` (1.x spells it `McpError`) and `GetSessionIdCallback` (gone in 2.x, so it becomes a plain callable type there).

`mcp_instrumentation.py` now uses the shared `MCP_V2` flag instead of its own version check (`_is_mcp_v1`). The server-side OTEL patches still only apply on 1.x.

## Related Issues

Part of #1659.

## Type of Change

New feature

## Testing

- New `tests/strands/tools/mcp/test__compat.py` covers both versions. The version-specific tests patch real attributes on the installed package (no `create=True`) and skip when the installed version doesn't have them.
- New `mcp-v2-compat` CI job installs `mcp==2.0.*` over the pin and checks against the real package: `import strands` works, `MCP_V2` is true, and the compat tests pass. It pins `2.0.*` so new `mcp` releases can't break unrelated PRs.
- New `make_mcp_error` helper in conftest builds the error type for whichever version is installed, since 2.x changed the constructor. The elicitation tests use it.

## Checklist

----

### DIFF
--- a/strands-py/src/strands/tools/mcp/_compat.py
+++ b/strands-py/src/strands/tools/mcp/_compat.py
@@ -21,7 +21,7 @@ import httpx
 from mcp import ClientSession
 from mcp.types import ServerCapabilities
 
-__all__ = ["MCP_V2", "GetSessionIdCallback", "MCPError", "initialize_session", "streamable_http_transport"]
+__all__ = ["MCP_V2", "GetSessionIdCallback", "MCPError", "negotiate_session", "streamable_http_transport"]
 
 # Feature-probed rather than version-parsed so pre-releases and backports
 # resolve by capability: `ClientSession.discover` is the 2.x replacement for
@@ -44,14 +44,17 @@ except ImportError:
     GetSessionIdCallback = Callable[[], str | None]  # type: ignore[misc, assignment]
 
 
-async def initialize_session(session: ClientSession) -> tuple[str | None, ServerCapabilities | None]:
+async def negotiate_session(session: ClientSession) -> tuple[str | None, ServerCapabilities | None]:
     """Negotiate the connection on an entered session, on either `mcp` major line.
 
-    The 2026-07-28 spec replaced the mandatory `initialize` handshake with a
-    stateless `server/discover` probe (SEP-2575). `negotiate_auto` is the
-    official 2.x client's connect-time policy: it probes `server/discover`
-    first and falls back to the legacy handshake for pre-2026 servers, so
-    either server era works. It lives in a private module, so the forced-2.x
+    The 2026-07-28 spec (SEP-2575) removed the `initialize` handshake and
+    moved negotiation into a per-request `_meta` envelope that the 2.x client
+    stamps itself. `server/discover` is the optional up-front probe for the
+    server's supported versions and capabilities. `negotiate_auto` is the
+    official 2.x client's connect-time policy over it: it probes
+    `server/discover` first and falls back to the legacy handshake for
+    pre-2026 servers, so either server era works. It is importable from
+    `mcp.client.client` but not exported in an `__all__`, so the forced-2.x
     CI job is what catches a relocation.
 
     Args:
@@ -61,13 +64,13 @@ async def initialize_session(session: ClientSession) -> tuple[str | None, Server
         The server's instructions (if any) and advertised capabilities.
     """
     if MCP_V2:
-        from mcp.client._probe import negotiate_auto  # type: ignore[import-not-found]
+        from mcp.client.client import negotiate_auto  # type: ignore[import-not-found]
 
         await negotiate_auto(session)
         return session.instructions, session.server_capabilities  # type: ignore[attr-defined]
 
     init_result = await session.initialize()
-    return init_result.instructions, session.get_server_capabilities()
+    return init_result.instructions, session.get_server_capabilities()  # type: ignore[attr-defined]
 
 
 def streamable_http_transport(
--- a/strands-py/src/strands/tools/mcp/_compat.py
+++ b/strands-py/src/strands/tools/mcp/_compat.py
@@ -19,8 +19,9 @@ from typing import Any
 
 import httpx
 from mcp import ClientSession
+from mcp.types import ServerCapabilities
 
-__all__ = ["MCP_V2", "GetSessionIdCallback", "MCPError", "streamable_http_transport"]
+__all__ = ["MCP_V2", "GetSessionIdCallback", "MCPError", "initialize_session", "streamable_http_transport"]
 
 # Feature-probed rather than version-parsed so pre-releases and backports
 # resolve by capability: `ClientSession.discover` is the 2.x replacement for
@@ -41,6 +42,32 @@ except ImportError:
     from collections.abc import Callable
 
     GetSessionIdCallback = Callable[[], str | None]  # type: ignore[misc, assignment]
+
+
+async def initialize_session(session: ClientSession) -> tuple[str | None, ServerCapabilities | None]:
+    """Negotiate the connection on an entered session, on either `mcp` major line.
+
+    The 2026-07-28 spec replaced the mandatory `initialize` handshake with a
+    stateless `server/discover` probe (SEP-2575). `negotiate_auto` is the
+    official 2.x client's connect-time policy: it probes `server/discover`
+    first and falls back to the legacy handshake for pre-2026 servers, so
+    either server era works. It lives in a private module, so the forced-2.x
+    CI job is what catches a relocation.
+
+    Args:
+        session: An entered `ClientSession` that has not yet negotiated.
+
+    Returns:
+        The server's instructions (if any) and advertised capabilities.
+    """
+    if MCP_V2:
+        from mcp.client._probe import negotiate_auto  # type: ignore[import-not-found]
+
+        await negotiate_auto(session)
+        return session.instructions, session.server_capabilities  # type: ignore[attr-defined]
+
+    init_result = await session.initialize()
+    return init_result.instructions, session.get_server_capabilities()
 
 
 def streamable_http_transport(
--- a/strands-py/tests/strands/tools/mcp/test__compat.py
+++ b/strands-py/tests/strands/tools/mcp/test__compat.py
@@ -19,7 +19,7 @@ import mcp.client.streamable_http as streamable_http_module
 import pytest
 
 from strands.tools.mcp import _compat
-from strands.tools.mcp._compat import MCPError, initialize_session, streamable_http_transport
+from strands.tools.mcp._compat import MCPError, negotiate_session, streamable_http_transport
 
 requires_mcp_v1 = pytest.mark.skipif(_compat.MCP_V2, reason="exercises the mcp 1.x branch")
 requires_mcp_v2 = pytest.mark.skipif(not _compat.MCP_V2, reason="exercises the mcp 2.x line")
@@ -71,14 +71,14 @@ def test_get_session_id_callback_falls_back_to_plain_callable(monkeypatch):
 
 @requires_mcp_v2
 def test_installed_v2_line_exposes_negotiate_auto():
-    """Test that the private negotiation policy `initialize_session` imports exists on the 2.x line."""
-    from mcp.client._probe import negotiate_auto
+    """Test that the negotiation policy `negotiate_session` imports exists on the 2.x line."""
+    from mcp.client.client import negotiate_auto
 
     assert callable(negotiate_auto)
 
 
 @pytest.mark.asyncio
-async def test_initialize_session_v1_runs_the_initialize_handshake(monkeypatch):
+async def test_negotiate_session_v1_runs_the_initialize_handshake(monkeypatch):
     """Test that the 1.x handshake is initialize(), with instructions read from its result."""
     monkeypatch.setattr(_compat, "MCP_V2", False)
     server_capabilities = MagicMock()
@@ -86,7 +86,7 @@ async def test_initialize_session_v1_runs_the_initialize_handshake(monkeypatch):
     session.initialize = AsyncMock(return_value=MagicMock(instructions="use the tools"))
     session.get_server_capabilities = MagicMock(return_value=server_capabilities)
 
-    instructions, capabilities = await initialize_session(session)
+    instructions, capabilities = await negotiate_session(session)
 
     session.initialize.assert_awaited_once_with()
     assert instructions == "use the tools"
@@ -94,25 +94,26 @@ async def test_initialize_session_v1_runs_the_initialize_handshake(monkeypatch):
 
 
 @pytest.mark.asyncio
-async def test_initialize_session_v2_negotiates_and_reads_session_properties(monkeypatch):
+async def test_negotiate_session_v2_negotiates_and_reads_session_properties(monkeypatch):
     """Test that the 2.x handshake delegates to negotiate_auto and reads the session properties.
 
-    The `mcp.client._probe` module exists only on the 2.x line, so a stub
+    The `mcp.client.client` module exists only on the 2.x line, so a stub
     module stands in for it to exercise this branch under the 1.x pin.
     """
     monkeypatch.setattr(_compat, "MCP_V2", True)
     negotiate_auto = AsyncMock()
-    probe_module = ModuleType("mcp.client._probe")
-    probe_module.negotiate_auto = negotiate_auto  # type: ignore[attr-defined]
-    monkeypatch.setitem(sys.modules, "mcp.client._probe", probe_module)
+    client_module = ModuleType("mcp.client.client")
+    client_module.negotiate_auto = negotiate_auto  # type: ignore[attr-defined]
+    monkeypatch.setitem(sys.modules, "mcp.client.client", client_module)
     server_capabilities = MagicMock()
     session = MagicMock()
     session.instructions = "use the tools"
     session.server_capabilities = server_capabilities
 
-    instructions, capabilities = await initialize_session(session)
+    instructions, capabilities = await negotiate_session(session)
 
     negotiate_auto.assert_awaited_once_with(session)
+    session.initialize.assert_not_called()
     assert instructions == "use the tools"
     assert capabilities is server_capabilities
 
--- a/strands-py/tests/strands/tools/mcp/test__compat.py
+++ b/strands-py/tests/strands/tools/mcp/test__compat.py
@@ -9,17 +9,20 @@ the `MCP_V2` flag forced.
 """
 
 import importlib
+import sys
 from collections.abc import Callable
 from contextlib import asynccontextmanager
-from unittest.mock import MagicMock, patch
+from types import ModuleType
+from unittest.mock import AsyncMock, MagicMock, patch
 
 import mcp.client.streamable_http as streamable_http_module
 import pytest
 
 from strands.tools.mcp import _compat
-from strands.tools.mcp._compat import MCPError, streamable_http_transport
+from strands.tools.mcp._compat import MCPError, initialize_session, streamable_http_transport
 
 requires_mcp_v1 = pytest.mark.skipif(_compat.MCP_V2, reason="exercises the mcp 1.x branch")
+requires_mcp_v2 = pytest.mark.skipif(not _compat.MCP_V2, reason="exercises the mcp 2.x line")
 requires_v2_transport_names = pytest.mark.skipif(
     not hasattr(streamable_http_module, "streamable_http_client"),
     reason="installed mcp line lacks the 2.x transport names",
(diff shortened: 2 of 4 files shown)


strands-py/src/strands/tools/mcp/_compat.py
```diff
@@ -0,0 +1,116 @@
...
+        The server's instructions (if any) and advertised capabilities.
+    """
+    if MCP_V2:
+        from mcp.client._probe import negotiate_auto  # type: ignore[import-not-found]
+
+        await negotiate_auto(session)
+        return session.instructions, session.server_capabilities  # type: ignore[attr-defined]
+
+    init_result = await session.initialize()
+    return init_result.instructions, session.get_server_capabilities()
```
strandly-the-agent: 🟡 **The 1.x branch is the only cross-line attribute access in this module without an `attr-defined` ignore, so `_compat` itself doesn't type-check on the 2.x line** — which contradicts the invariant the docstring states at `_compat.py:9-13` ("the branch-level ignores cover whichever line mypy runs under").

Reproduced on `mcp 2.0.0` (mypy 1.20.2, repo config):
```
_compat.py:70: error: "ClientSession" has no attribute "get_server_capabilities"; maybe "server_capabilities"?  [attr-defined]
```

```suggestion
    return init_result.instructions, session.get_server_capabilities()  # type: ignore[attr-defined]
```

Verified: with the fix, on 2.0.0 this error is gone (the only remaining one is the pre-existing `:107` `httpx2` `auth` arg-type from #3708); on 1.29.0 `mypy ./src/strands/tools/mcp/` is still `Success` — the new per-module `warn_unused_ignores = false` (`pyproject.toml:240-244`) absorbs the now-unused ignore, as it already does for six others here. Nothing catches this today: the `mcp-v2-compat` job runs an import check + `test__compat.py`, no mypy.

**Reach:** no runtime impact on any user (pin holds, `MCP_V2=False` on 1.29.0). Hits the next person in this workstream — anyone bumping the pin or running mypy on 2.x — which is the audience this module exists for.

strands-py/tests/strands/tools/mcp/test__compat.py
```diff
@@ -0,0 +1,172 @@
...
+    probe_module.negotiate_auto = negotiate_auto  # type: ignore[attr-defined]
+    monkeypatch.setitem(sys.modules, "mcp.client._probe", probe_module)
+    server_capabilities = MagicMock()
+    session = MagicMock()
+    session.instructions = "use the tools"
+    session.server_capabilities = server_capabilities
+
+    instructions, capabilities = await initialize_session(session)
+
+    negotiate_auto.assert_awaited_once_with(session)
```
strandly-the-agent, line 115: 🟡 **The one behavior this commit is about — discover-first with legacy fallback — isn't pinned by a test.** This test injects an `AsyncMock` stub for `mcp.client._probe`, so it proves *strands calls* `negotiate_auto`, not that negotiation happens; and because `session` is a bare `MagicMock`, a stray `await session.initialize()` leaking into the 2.x branch would be silently absorbed with every assertion still green.

Cheap half — pins the two branches as mutually exclusive:
```suggestion
    negotiate_auto.assert_awaited_once_with(session)
    session.initialize.assert_not_called()
```

<details><summary>Valuable half — a real 2.x negotiation test (~40 lines, no network, &lt;1s, passes on 2.0.0 today)</summary>

Behind the existing `@requires_mcp_v2` it would run in the `mcp-v2-compat` job. I ran exactly this against real `mcp 2.0.0`:

```python
import anyio
from mcp import ClientSession
from mcp.server.mcpserver import MCPServer
from mcp.server.runner import serve_dual_era_loop, serve_loop
from mcp.shared.memory import create_client_server_memory_streams
from strands.tools.mcp._compat import initialize_session

async def run(server_driver, label):
    srv = MCPServer(name="probe-server", version="1.2.3", instructions="use the calculator")
    @srv.tool()
    def add(a: int, b: int) -> int:
        """Add two numbers."""
        return a + b
    low = srv._lowlevel_server
    async with create_client_server_memory_streams() as ((cr, cw), (sr, sw)):
        async with anyio.create_task_group() as tg:
            async def serve():
                async with low.lifespan(low) as state:
                    await server_driver(low, sr, sw, lifespan_state=state)
            tg.start_soon(serve)
            async with ClientSession(cr, cw) as session:
                instructions, caps = await initialize_session(session)
                print(label, instructions, session.protocol_version,
                      session.discover_result is not None, session.initialize_result is not None)
                res = await session.call_tool("add", {"a": 2, "b": 3})
                print(label, res.is_error, res.content[0].text)
            tg.cancel_scope.cancel()
```
Output:
```
A dual-era     'use the calculator' 2026-07-28 discover=True  init=False   add=5
B handshake    'use the calculator' 2025-11-25 discover=False init=True    add=5
```
One more assertion worth having: that the request `_meta` carries `io.modelcontextprotocol/{protocolVersion,clientInfo,clientCapabilities}` — that's the actual SEP-2575 mandate, pinned in a line. Related, for the record: `mcp_client.py:1081` is never exercised with `MCP_V2=True` — the 2.x client path exists only in isolation against mocks.

</details>

**Reach:** no user; a future regression in the 2.x branch would ship green. Worth taking now because this file *is* the guard the pin-lift will lean on.

strands-py/src/strands/tools/mcp/_compat.py
```diff
@@ -0,0 +1,116 @@
...
+
+async def initialize_session(session: ClientSession) -> tuple[str | None, ServerCapabilities | None]:
+    """Negotiate the connection on an entered session, on either `mcp` major line.
+
+    The 2026-07-28 spec replaced the mandatory `initialize` handshake with a
+    stateless `server/discover` probe (SEP-2575). `negotiate_auto` is the
+    official 2.x client's connect-time policy: it probes `server/discover`
+    first and falls back to the legacy handshake for pre-2026 servers, so
+    either server era works. It lives in a private module, so the forced-2.x
+    CI job is what catches a relocation.
```
strandly-the-agent: 🟡 **You asked whether this matches SEP-2575 — the code does (ledger in the review), but this docstring sentence doesn't, and it's the sentence that ships.** Per SEP-2575, servers *MUST* implement `server/discover` while clients *MAY* call it — it's a discovery/probe convenience, not the handshake's successor. What replaces the handshake is the per-request `_meta` envelope (`protocolVersion`/`clientCapabilities` REQUIRED, stamped by the library); `server/discover` is one of *two* negotiation flows (the other being optimistic-send-then-retry-on-`-32022`).

```suggestion
    The 2026-07-28 spec (SEP-2575) removed the `initialize` handshake and moved
    negotiation into a per-request `_meta` envelope the 2.x client stamps
    itself. `server/discover` is the optional up-front probe for the server's
    supported versions and capabilities; `negotiate_auto` is the official
    client's connect-time policy over it -- probe first, fall back to the legacy
    handshake for pre-2026 servers -- so either server era works. It lives in a
    private module, so the forced-2.x CI job is what catches a relocation.
```

The phrasing is inherited from #1659's own table, so it isn't a code problem — but the docstring is where the next person on that tracker reads it. Same shorthand is in the commit message ¶1 and PR body ¶1 if you want them consistent. (Deliberately not re-raising the `initialize_session` rename thread at `:47`.)

**Reach:** documentation only, zero runtime effect.

strandly-the-agent (commented):
**Verdict: comment** — no blockers, no changes requested. It matches SEP-2575; three 🟡 polish items (one type-ignore, one test, one docstring sentence) worth taking before merge.

**Does it match SEP-2575? Yes.** The code delegates negotiation to the official client's `negotiate_auto` — probe `server/discover` first, fall back to the legacy `initialize` handshake for pre-2026 servers — and reads era-unifying session properties. I verified it end to end against a real `mcp==2.0.0` server in *both* eras, plus a cross-era fallback to a real 1.29.0 legacy stdio server. strands' `src/` calls none of the RPCs SEP-2575 removes, and the per-request `_meta` mandate (`protocolVersion`/`clientCapabilities`) is satisfied by the library itself — nothing for strands to do. What overreaches is the *prose*, not the code (inline on the docstring; question 2 on the pin-lift claim).

✅ **Verified** — branch `0efc1395` · `mypy`/`ruff` clean + 259 passed/1 skipped on `mcp 1.29.0` (the pinned line; `MCP_V2=False`, behavior identical to `main`) · E2E on real `mcp 2.0.0`, both server eras · cross-era fallback against a real 1.29.0 legacy server.

Scope is clean: the genuinely new commit `0efc1395` is 3 files / +86−10, fully in-contract for #1659 row 1. (The wider 12-file / +381−96 diff is 4 of 5 commits belonging to the still-open stacked #3708, reviewed there.)

<details><summary>SEP-2575 conformance ledger (what I checked)</summary>

| SEP-2575 requirement | This PR | Verified |
|---|---|---|
| `initialize` handshake removed; negotiation moves to per-request `_meta` (`protocolVersion` + `clientCapabilities` REQUIRED) | Nothing strands-side — the 2.x lib stamps `protocolVersion`/`clientInfo`/`clientCapabilities` + the `MCP-Protocol-Version` header on every request once negotiated | ✅ sniffed on `server/discover`, `tools/list`, `tools/call` |
| Servers MUST implement `server/discover`; clients MAY call it | Delegated to `negotiate_auto` (`mcp/client/_probe.py`), not reimplemented; probe-first is spec-sanctioned (STDIO SHOULD, HTTP MAY) | ✅ read `_probe.py` |
| `-32601`/legacy fallback for pre-2026 servers | Covered — `negotiate_auto`'s denylist is a superset of the required fallback | ✅ source + cross-era E2E |
| Dual-era client ↔ legacy server | strands on `mcp 2.0.0` ↔ real `mcp 1.29.0` stdio server → fell back to `initialize()`, `protocol_version=2025-11-25`, instructions intact | ✅ E2E |
| Modern path negotiates | strands ↔ real 2.x dual-era server → `server/discover`, `protocol_version=2026-07-28`, caps + `tools/call` fine | ✅ E2E |
| No behavior change on the pinned 1.x line | `MCP_V2=False` on 1.29.0; 259 passed/1 skipped; the reordered `get_server_capabilities()` is a pure accessor, so moving it is neutral | ✅ |

</details>

- **Correctness/safety** — approve, no blockers. Clean on the pinned line (mypy, ruff, 259 tests); the 1.x/2.x E2E matrix above; the two new type-ignores confirmed *not* unused on 1.x. Found the missing `attr-defined` ignore (①) and the mock-can't-discriminate gap (②).
- **Issue-alignment / SEP-2575** — aligned, no scope creep. Built two venvs (1.29.0 / 2.0.0) and ran strands' own `initialize_session` against a real 2.x server in both eras. Confirmed `session.instructions` / `session.server_capabilities` are era-unifying properties, so the 2.x branch reads correctly for both.
- **Adversarial** and **test-quality** — see the disclosure at the bottom.

</details>

### Questions (none blocking)

1. **`_probe` privacy — two new datapoints for the existing thread at `_compat.py:64`** (not a new ask): (a) there's genuinely *no* public alternative — `Client(mode="auto")`'s `__aenter__` is literally `await negotiate_auto(session)` (`mcp/client/client.py:457`), and `ClientSession.discover()` deliberately has no fallback — so reusing `_probe` is right *if* it's a short-lived bridge; (b) I measured the relocation failure mode: simulating the module moving on 2.0.0 yields `MCPClientInitializationError: … unhandled errors in a TaskGroup (1 sub-exception)`, with the `ModuleNotFoundError` swallowed. Worth deciding deliberately — opaque-and-loud, or catch `ImportError` and re-raise naming the module? Is `MCPClient` eventually rebuilt on `Client`, or should we ask upstream for a public "run the auto policy on this session" hook?
2. **"lifting the pin is a one-line change later"** — the first half (pin stays `<2`) is right; the second isn't yet. On `mcp 2.0.0`, `MCPClient(transport, tasks_config={…})` raises `ImportError: cannot import name 'TaskExecutionMode'` at construction (`mcp_client.py:327`, because `_is_tasks_enabled()` at `:1431` is config-driven not feature-probed), and the first RPC after a 2.x connect dies at `mcp_client.py:623` on `list_tools(cursor=…)` (kwarg removed in 2.x); plus 65 unit tests fail on 2.0.0. All pre-existing, none of it this PR's job — but either soften the sentence or enumerate the remainder on #1659.
3. **Merge order with #3708?** It's still open and is this branch's base. Should #3708 land first so this diff shrinks to the 3 genuinely-new files, or does #3821 supersede it?
4. **Should `mcp-v2-compat` run `mypy … _compat.py`?** It's the one check that would have caught ①. Not green today even with ① fixed (the pre-existing `httpx2`-vs-`httpx` `auth` arg-type at `_compat.py:107`, already noted on #3708, remains) — so "worth wiring once that's settled", not now.
5. **Who unwinds the daemon-thread lifecycle?** #1659 names it as the strands-side work; nothing here touches it (still one long-lived `ClientSession` per `MCPClient`). Deliberately out of scope and fine — file a child issue so the tracker's row 1 isn't read as done when this lands?

<details><summary>Appendix — non-blockers &amp; nits (3)</summary>

- **⚪ Task support silently degrades on the modern era, and the commit message reads as reassurance that it doesn't.** `ServerTasksCapability` is documented "(2025-11-25 only)" (`mcp_types/_types.py:477`) and 2.x servers advertise tasks under `capabilities.extensions` (SEP-2663) instead, so on any 2026-era discover result `caps.tasks is None` → `_server_task_capable=False` silently (`mcp_client.py:1090-1096`). Measured on the 2.x E2E. Unreachable today (pin holds), correct-by-degradation — but I'd drop that sentence from the commit message.
- **⚪ "The rest of the client lifecycle needs no fork" holds for the lifecycle, not the first RPC after it** (see question 2's `list_tools(cursor=…)`). Just so "server/discover works" isn't read as "`MCPClient` works on 2.x".
- **⚪ Title nit:** "added server/discover" oversells slightly — no strands code sends `server/discover`; `negotiate_auto` does. Something like *"negotiate the connection era on both `mcp` lines"* matches what landed. The body's "delegates to `negotiate_auto`" is accurate.

</details>

<details><summary>⚠ Pipeline disclosure — two passes did not deliver</summary>

This review ran four specialist passes; the **adversarial** and **test-quality** passes **timed out** and produced no output. Their core surfaces were substantially covered by the two that completed — both ran real E2E repros on `mcp 1.29.0` and `2.0.0` (cross-era legacy-server fallback, `_probe`-relocation simulation, mypy on both lines) — but depth on test quality and hostile-input edges is thinner than a full run. Read the verdict as "nothing found by two thorough passes", not "four passes found nothing".

</details>

---
This is agentic review — solid work for a human to approve, not a gate. I didn't re-raise the two open `github-actions` threads on `_compat.py` (`:47` naming, `:64` `_probe` privacy); question 1 adds evidence to the latter.

strandly-the-agent (commented):
**Re-reviewed `d5e3b168` — all three of my findings are fixed, nothing new introduced. Good to approve.**

Status of my previous findings (verified against the actual diff + fresh runs, not the commit message):

| Finding | Status |
|---|---|
| ① `_compat.py:70` missing `attr-defined` ignore | ✅ **Fixed** (`_compat.py:72`). mypy on `mcp 2.0.0`: the error is gone — the only remaining one is the pre-existing `:109` `httpx2` `auth` arg-type already noted on #3708. mypy on 1.29.0: still `Success: no issues found in 7 source files` (the per-module `warn_unused_ignores=false` absorbs it, as expected). |
| ② v2 test can't discriminate negotiate-only | ✅ **Fixed** (`test__compat.py:116`). I mutation-tested it: injected a stray `await session.initialize()` into the 2.x branch → the test now **fails**; reverted → passes. The assertion earns its keep. |
| ③ docstring "`server/discover` replaces the handshake" | ✅ **Fixed** (`_compat.py:50-57`) — now correctly frames the per-request `_meta` envelope as the handshake's replacement and `server/discover` as the optional probe. PR body ¶1 updated to match. |

✅ **Verified** — head `d5e3b168` · full MCP suite on `mcp 1.29.0`: 259 passed / 1 skipped (unchanged) · `test__compat.py` on real `mcp 2.0.0`: 7 passed / 1 skipped · mypy both lines as above · delta is fixes-only (2 files, +10/−7), so I reviewed it directly without re-running the full pipeline.

The PR body edit also resolves my question 2 — it now says lifting the pin is *not* yet one-line and names the remainder (tasks, `list_tools` pagination, unit suite) as tracked on #1659. Thanks for folding that in.

**Still open, none blocking** (answers welcome here, per your note):
1. **Merge order with #3708** — it's still open and is this branch's base (4 of 5 commits). Land #3708 first so this shrinks to the 2 genuinely-new commits, or does #3821 supersede it?
2. The **real-2.x negotiation test** (the collapsed "valuable half" in my ② thread) remains optional-open — the in-memory server test I posted passes on 2.0.0 today and would give the `mcp-v2-compat` job a behavioral guard, not just a call-shape one. Fine as a follow-up.
3. The **`_probe` relocation error-message decision** stays with the existing `_compat.py:64` thread (incl. `github-actions`' point that the `2.0.*` job pin doesn't guard later minors) — not re-raising it here, just noting it's the one thread left with a decision in it.

strands-py/src/strands/tools/mcp/_compat.py
pgrayy, line 1: Can we support mcp v2 integ tests?

strands-py/src/strands/tools/mcp/_compat.py
```diff
@@ -0,0 +1,118 @@
...
+    """Negotiate the connection on an entered session, on either `mcp` major line.
+
+    The 2026-07-28 spec (SEP-2575) removed the `initialize` handshake and
+    moved negotiation into a per-request `_meta` envelope that the 2.x client
+    stamps itself. `server/discover` is the optional up-front probe for the
+    server's supported versions and capabilities. `negotiate_auto` is the
+    official 2.x client's connect-time policy over it: it probes
+    `server/discover` first and falls back to the legacy handshake for
+    pre-2026 servers, so either server era works. It lives in a private
+    module, so the forced-2.x CI job is what catches a relocation.
```
pgrayy: You say here that the initializes falls back to the legacy handshake. So does that mean we need to continue supporting mcp v1 in code here? Or can we fully upgrade to the mcp v2 client and let it handle backwards compatibility?

strands-py/src/strands/tools/mcp/_compat.py
```diff
@@ -0,0 +1,118 @@
...
+
+    Args:
+        session: An entered `ClientSession` that has not yet negotiated.
+
+    Returns:
+        The server's instructions (if any) and advertised capabilities.
+    """
+    if MCP_V2:
+        from mcp.client._probe import negotiate_auto  # type: ignore[import-not-found]
```
pgrayy: We should avoid operating on private members.

strands-py/src/strands/tools/mcp/_compat.py
pgrayy: From Joe:

Option A: rewrite MCPClient in pure 2.x style and require mcp>=2. The code gets cleaner (no flag, no _compat), but it's the "just bump" option from before: every downstream environment that still has anything pinned mcp<2 (including our own strands-agents-mcp-server since fastmcp broke on 2.0) can no longer install or upgrade strands. That's a hard break shipped in a minor release.

Option B: update the client to speak the new API through _compat, keep 1.x working, widen the pin to >=1.23,<3. This is your current stack. #3708 makes the imports version-proof, #3821 adopts the new lifecycle (negotiate_auto / server/discover), #3984 moves list pagination to the form both lines accept. After the pin widens, users who install mcp 2.x get the full new behavior including 2026-07-28 servers, and users stuck on 1.x lose nothing.

pgrayy (approved):

strands-py/src/strands/tools/mcp/_compat.py
poshinchen: Discussed offline, will make another PR for the integ tests

PR #3826, opened 2026-08-14
fix(a2a): reject NaN and Infinity at the JSON-RPC boundary

marcinmarzeta:
## Description

`_jsonable` guards the A2A executor's JSON-RPC boundary, but a non-finite float (`NaN`, `Infinity`) passes its `json.dumps` probe while pydantic's response serialization silently turns it into `null` (`ser_json_inf_nan` default) — the reason vanishes with no error, far from the interrupt that produced it. Guard non-finite float scalars explicitly and degrade them to their string form (`"nan"`, `"inf"`).

The guard is scalar-only on purpose: a structured reason keeps its JSON shape, so a dict with one non-finite leaf still arrives as a dict (the leaf becomes a `null` leaf on the wire, unchanged from today) rather than collapsing to an unparseable repr string.

Also adds a wire-level test pinning what a client actually receives (`model_dump_json` output, not the in-memory dict), finite-float coverage proving ordinary numbers pass unchanged, and a docs line on the degradation behavior.

Split out of #3817, which routes streaming-tool yields through the same helper and should not carry an interrupt-path behavior change.

## Related Issues

Split out of #3817.

## Documentation PR

None. One sentence on the degradation behavior added to `agent-to-agent.mdx` in this PR.

## Type of Change

Bug fix

## Testing

## Checklist

----

### DIFF
--- a/strands-py/src/strands/multiagent/a2a/executor.py
+++ b/strands-py/src/strands/multiagent/a2a/executor.py
@@ -84,7 +84,9 @@ def _jsonable(value: Any) -> Any:
     surface only while serializing the response, long after the interrupt details are assembled.
     """
     try:
-        json.dumps(value)
+        # allow_nan=False rejects NaN and Infinity, which json.dumps otherwise emits as bare
+        # literals that a strict JSON-RPC client refuses to parse.
+        json.dumps(value, allow_nan=False)
     except (TypeError, ValueError):
         return str(value)
     return value
--- a/strands-py/src/strands/multiagent/a2a/executor.py
+++ b/strands-py/src/strands/multiagent/a2a/executor.py
@@ -12,6 +12,7 @@ import asyncio
 import base64
 import json
 import logging
+import math
 import mimetypes
 import uuid
 import warnings
@@ -83,10 +84,13 @@ def _jsonable(value: Any) -> Any:
     An interrupt reason is arbitrary user data. A value the transport cannot encode would otherwise
     surface only while serializing the response, long after the interrupt details are assembled.
     """
+    if isinstance(value, float) and not math.isfinite(value):
+        # A non-finite float is not valid JSON. pydantic's response serialization silently turns it
+        # into null, so the reason vanishes with no error; degrade just this scalar to a string and
+        # leave a structured reason's shape intact.
+        return str(value)
     try:
-        # allow_nan=False rejects NaN and Infinity, which json.dumps otherwise emits as bare
-        # literals that a strict JSON-RPC client refuses to parse.
-        json.dumps(value, allow_nan=False)
+        json.dumps(value)
     except (TypeError, ValueError):
         return str(value)
     return value
--- a/strands-py/tests/strands/multiagent/a2a/test_executor.py
+++ b/strands-py/tests/strands/multiagent/a2a/test_executor.py
@@ -2694,16 +2694,30 @@ async def test_interrupt_advertises_every_pending_id(mock_strands_agent, mock_re
     assert tru_ids == exp_ids
 
 
+class _UnserializableReason:
+    """A reason no JSON encoder accepts."""
+
+    def __str__(self):
+        return "<custom reason>"
+
+
 @pytest.mark.asyncio
+@pytest.mark.parametrize(
+    ("reason", "exp_reason"),
+    [
+        (_UnserializableReason(), "<custom reason>"),
+        # json.dumps emits NaN and Infinity as bare literals that a strict JSON-RPC client refuses.
+        (float("nan"), "nan"),
+        (float("inf"), "inf"),
+        # The whole reason degrades together, so one bad leaf collapses the dict around it.
+        ({"limit": float("inf"), "note": "cap"}, "{'limit': inf, 'note': 'cap'}"),
+    ],
+)
 async def test_interrupt_reason_that_is_not_json_falls_back_to_text(
-    mock_strands_agent, mock_request_context, mock_event_queue
+    reason, exp_reason, mock_strands_agent, mock_request_context, mock_event_queue
 ):
     """A reason the transport cannot encode degrades to its string form rather than failing late."""
     from strands.interrupt import Interrupt
-
-    class Unserializable:
-        def __str__(self):
-            return "<custom reason>"
 
     executor = StrandsA2AExecutor(mock_strands_agent)
     await _run_until_interrupt(
@@ -2711,12 +2725,11 @@ async def test_interrupt_reason_that_is_not_json_falls_back_to_text(
         mock_strands_agent,
         mock_request_context,
         mock_event_queue,
-        [Interrupt(id="int-1", name="approval", reason=Unserializable())],
+        [Interrupt(id="int-1", name="approval", reason=reason)],
     )
 
     data_parts = [p.root.data for p in _input_required_message(mock_event_queue).parts if isinstance(p.root, DataPart)]
     tru_reason = data_parts[0]["interrupts"][0]["reason"]
-    exp_reason = "<custom reason>"
     assert tru_reason == exp_reason
 
 
--- a/strands-py/tests/strands/multiagent/a2a/test_executor.py
+++ b/strands-py/tests/strands/multiagent/a2a/test_executor.py
@@ -1,6 +1,7 @@
 """Tests for the StrandsA2AExecutor class."""
 
 import base64
+import json
 from typing import Any
 from unittest.mock import AsyncMock, MagicMock, patch
 
@@ -2639,7 +2640,7 @@ async def test_interrupt_advertises_ids_in_data_part(mock_strands_agent, mock_re
         mock_strands_agent,
         mock_request_context,
         mock_event_queue,
-        [Interrupt(id="v1:tool_call:tu-1:abc", name="approve_campaign", reason={"name": "spring"})],
+        [Interrupt(id="v1:tool_call:tu-1:abc", name="approve_campaign", reason={"name": "spring", "limit": 1.5})],
     )
 
     data_parts = [p.root.data for p in _input_required_message(mock_event_queue).parts if isinstance(p.root, DataPart)]
@@ -2647,7 +2648,11 @@ async def test_interrupt_advertises_ids_in_data_part(mock_strands_agent, mock_re
     exp_data = [
         {
             "interrupts": [
-                {"interruptId": "v1:tool_call:tu-1:abc", "name": "approve_campaign", "reason": {"name": "spring"}}
+                {
+                    "interruptId": "v1:tool_call:tu-1:abc",
+                    "name": "approve_campaign",
+                    "reason": {"name": "spring", "limit": 1.5},
+                }
             ]
         }
     ]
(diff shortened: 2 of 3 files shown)


strands-py/src/strands/multiagent/a2a/executor.py
```diff
@@ -84,7 +84,9 @@ def _jsonable(value: Any) -> Any:
     surface only while serializing the response, long after the interrupt details are assembled.
     """
     try:
-        json.dumps(value)
+        # allow_nan=False rejects NaN and Infinity, which json.dumps otherwise emits as bare
+        # literals that a strict JSON-RPC client refuses to parse.
+        json.dumps(value, allow_nan=False)
```
strandly-the-agent: 🟡 **A container reason with one non-finite leaf now loses its whole structure, where `main` shipped valid JSON.**

Measured on this branch vs `origin/main` (both through `DataPart(...).model_dump_json()`, a2a-sdk 0.3.26 / pydantic 2.13.4):

| `interrupt.reason` | `main` on the wire | this PR on the wire |
|---|---|---|
| `float("nan")` | `null` | `"nan"` ✅ better |
| `{"limit": inf, "note": "cap"}` | `{"limit": null, "note": "cap"}` | `"{'limit': inf, 'note': 'cap'}"` ⚠️ |

The second row is a Python `repr` — single quotes, bare `inf` — parseable by no client. A client doing `reason["note"]` works on `main` and breaks here. *Who hits it:* any caller whose `tool_context.interrupt(reason=...)` payload is a dict/list containing a computed non-finite float (a rate/ratio that hit a zero denominator), via the documented `StrandsA2AExecutor` interrupt path — and exposure grows if #3817 routes structured streaming-tool payloads through this same helper.

Guarding the scalar keeps the win and drops the regression — a nested non-finite float falls back to `main`'s null-leaf behavior (still lossy, but not worse than today):

```suggestion
    if isinstance(value, float) and not math.isfinite(value):
        # A non-finite float is not valid JSON. pydantic's response serialization silently turns it
        # into null, so the reason vanishes with no error; degrade just this scalar to a string and
        # leave a structured reason's shape intact.
        return str(value)
    try:
        json.dumps(value)
```

Two follow-ups if you take it: add `import math` to the stdlib import block (beside `import json`, line 13), and the dict param at `test_executor.py:2713` becomes `({"limit": float("inf"), "note": "cap"}, {"limit": float("inf"), "note": "cap"})`. I applied exactly this and ran the file: 94 pass, only that one param (which asserts the collapse) needs the update.

The rewritten comment also fixes a second thing: **the mechanism the current comment describes can't happen here.** `json.dumps` never runs on the response — pydantic does, and it emits `null` for a non-finite float (verified on all four a2a-sdk response paths; `DataPart.model_config` doesn't override `ser_json_inf_nan`); starlette's `JSONResponse.render` also uses `allow_nan=False`, so it would 500 rather than emit a bare literal. gRPC is unreachable — `A2AServer` only builds Starlette/FastAPI apps (`server.py:250,275`). Worth keeping the code comment on the real failure (silent `null`) so the next reader doesn't inherit the wrong model. Same wording applies to the parametrize comment at `test_executor.py:2709` and the docstring at `:2719` ("a reason the transport cannot encode" — it *can* encode `nan`/`inf`, it encodes them wrongly).

If whole-value collapse is deliberate even with the corrected facts, that's a defensible call — but then the comment should say why a repr string beats a null leaf.

strands-py/tests/strands/multiagent/a2a/test_executor.py
```diff
@@ -2694,29 +2694,42 @@ async def test_interrupt_advertises_every_pending_id(mock_strands_agent, mock_re
...
         mock_request_context,
         mock_event_queue,
-        [Interrupt(id="int-1", name="approval", reason=Unserializable())],
+        [Interrupt(id="int-1", name="approval", reason=reason)],
     )
 
     data_parts = [p.root.data for p in _input_required_message(mock_event_queue).parts if isinstance(p.root, DataPart)]
     tru_reason = data_parts[0]["interrupts"][0]["reason"]
-    exp_reason = "<custom reason>"
     assert tru_reason == exp_reason
```
strandly-the-agent, line 2738: 🟡 **Nothing here pins what a client actually receives** — this asserts on the in-memory `DataPart.data` dict, one serialization step short of the bug. Pre-fix, `reason=float("nan")` reached the wire as `"reason": null` on both response paths; the observable defect was that silent `null`, and no test states it.

I ran this one: it passes on this branch and fails with `assert None == 'nan'` once `allow_nan=False` is reverted — exactly the guarantee the PR is buying. (Note the local `import json`; the module has no top-level `json` import.)

```python
@pytest.mark.asyncio
async def test_interrupt_reason_survives_wire_serialization(
    mock_strands_agent, mock_request_context, mock_event_queue
):
    """The reason a client actually receives is the degraded text, not a null."""
    import json

    from strands.interrupt import Interrupt

    executor = StrandsA2AExecutor(mock_strands_agent)
    await _run_until_interrupt(
        executor,
        mock_strands_agent,
        mock_request_context,
        mock_event_queue,
        [Interrupt(id="int-1", name="approval", reason=float("nan"))],
    )

    wire = json.loads(_input_required_message(mock_event_queue).model_dump_json(exclude_none=True))
    data_parts = [p for p in wire["parts"] if p["kind"] == "data"]
    tru_reason = data_parts[0]["data"]["interrupts"][0]["reason"]
    exp_reason = "nan"
    assert tru_reason == exp_reason
```

strandly-the-agent (changes requested):
**changes-requested** — the scalar case is a clear win (`null` → `"nan"`), but for a *container* reason with one non-finite leaf this is a regression vs `main`, and the new comment documents a failure mode that can't happen on this transport. Both are fixed by the same few lines (inline suggestion).

**Verified** ✅ head `efb2523` (diff vs `origin/main` == diff vs merge-base `8fd742a3`) · ✅ `python -m pytest tests/strands/multiagent/a2a/test_executor.py -q` → **95 passed** (py3.13, a2a-sdk 0.3.26, pydantic 2.13.4) · ✅ mutation-checked: reverting `allow_nan=False` fails the new nan/inf/dict cases · ✅ wire path traced empirically on all four a2a-sdk response paths (`jsonrpc_app.py:557`/`:572`, `rest_adapter.py:110`, `base_push_notification_sender.py:60`) · ✅ suggested fix applied and run · ✅ proposed wire test run both ways (passes at head, `assert None == 'nan'` with the fix reverted).

**One correction that changes how you read this PR:** the pre-fix symptom in the body and commit message ("bare literals that a strict JSON-RPC client refuses to parse") does not occur on this stack. pydantic serializes a non-finite float in `DataPart.data` to `null` on every response path (`ser_json_inf_nan` default, not overridden by `DataPart.model_config`), and starlette's `JSONResponse.render` itself uses `allow_nan=False`, so a survivor would 500 rather than emit `NaN`. The real pre-fix bug is *silent data loss*: the reason arrives as `null` with no error. Still worth fixing — but it turns the dict case from an obvious win into a genuine trade (repr string vs one `null` leaf), which is worth re-deciding on the real facts.

No public Python API surface change — no `needs-api-review` label needed.

### Questions

❓ **(blocking, same as inline)** With the corrected mechanism, is whole-value collapse still what you want, or should the guard be scalar-only? Pre-fix a client got `{"limit": null, "note": "cap"}` (structure kept); post-fix it gets the Python repr `"{'limit': inf, 'note': 'cap'}"`, which nothing can parse — and `agent-to-agent.mdx:402` documents `reason` as an object. Either answer is fine; the PR body argued the trade against "the response is unparseable", which isn't the alternative.

❓ **(non-blocking)** Should the degraded wire value be specified once across SDKs? `strands-ts`'s a2a executor has no interrupt publisher yet, so no divergence today — but `JSON.stringify` is spec-bound to emit `null` for non-finite numbers, so when the TS interrupt port lands the two SDKs would put different values in the same field (py `"nan"` vs ts `null`). Worth a line on #3585 or a tracking note; the scalar-only fix narrows the gap.

<details><summary>Per-pass breakdown (6 passes)</summary>

| Pass | Verdict | What it contributed |
|---|---|---|
| Triage | accept | routed correctness, issue-alignment, adversarial, test-quality; skipped api-bar-raiser (no public surface change), llm-context, docs-accuracy |
| Context build | ✓ | `reason` is unconstrained `Any` from the public HITL entry point (`types/interrupt.py:82,107`); in-SDK producers all pass JSON-safe values, so non-finite floats only arise from user code; `reason` survives session persistence, so a restored agent re-publishes it |
| Correctness / safety | changes-requested | the container-collapse regression and the wrong-mechanism comment (inline on `executor.py:86-89`) |
| Issue alignment / scope | aligned | fix is right and minimal, split from #3817 is clean, nothing out of contract; the PR narrative carries the wrong pre-fix symptom (see above) |
| Test quality | gaps-found | mutation check **passed**; missing wire-level assertion (inline on `test_executor.py:2733`); confirmed the module-level `_UnserializableReason` hoist is *required* — parametrize args evaluate at import |
| Adversarial / repro | fix holds | first (deep) run timed out; a focused retry completed: no blockers. Probed gate/wire mismatches for exotic dict keys, throwing `__str__`, circular refs, float subclasses, Decimal, big ints — see appendix |

**Counts:** 6 passes → 14 raw findings → 2 inline (🔴 0 · 🟡 2) + 2 questions + appendix; duplicates across passes merged (the container trade and the wrong-narrative cluster each surfaced 2–3 times), 0 dropped as ungrounded.

</details>

<details><summary>Appendix — non-blocking (6)</summary>

- ⚪ **No finite-float coverage.** Nothing proves tightening the encoder left ordinary numbers alone. Cheapest: extend the whole-object assertion at `test_executor.py:2642-2653` to `reason={"name": "spring", "limit": 1.5}`. (Confirmed finite floats/str/bool are byte-identical across main / this PR / the suggested fix.)
- ⚪ **Docs.** `agent-to-agent.mdx:388-456` doesn't mention degradation, as the PR body notes. One line near `:402` — "a reason that isn't JSON-encodable arrives as its string form" — would be proportionate; #3817 already teaches the same expectation for tool yields.
- ⚪ **NaN/Infinity in *key* position now over-rejects** (new, contrived): `json.dumps` stringifies non-finite keys (`{"NaN": ...}`, valid JSON), but `allow_nan=False` rejects them, collapsing the dict to its repr. Pre-fix the wire silently rewrote that key to `"None"` anyway, so no validity regression — same shape as the main container finding; the scalar-only guard sidesteps it.
- ⚪ **Pre-existing gate/wire mismatches for non-str dict keys** (unchanged by this PR): `{1: "a", "1": "b"}` passes the gate but pydantic emits `{"1": "b"}` — one entry silently dropped; `{None: "z"}` → gate predicts `"null"`, wire says `"None"`. `_jsonable` is a validity gate, not a fidelity gate; widening it is out of scope here.
- ⚪ **Throwing `__str__`:** for any *truthy* reason, `executor.py:449` (`f": {interrupt.reason}"`) already crashed pre-PR, before `_jsonable` ran. This PR newly converts one contrived combo (falsy dict-subclass + throwing `__str__` + non-finite leaf) from `input_required` into a failed, unresumable task — three coincident conditions, negligible. General hardening would wrap `:449` and the fallback together. Also pre-existing at `:449`: `if interrupt.reason:` means a `0.0`/`{}`/`""` reason is dropped from the human-readable TextPart while `nan` (truthy) is included.
- ⚪ **Round-trip is clean:** `reason` is never parsed back — `_extract_interrupt_responses` reads only `interruptResponse.interruptId`/`.response` (`executor.py:694-731`), and `_InterruptState.resume()` keys on `interruptId`. Degrading it can't break resumption.

</details>

---
*I'm `strandly-the-agent`, an experimental AI reviewer — treat this as input for a human decision, not a gate. Note the deep adversarial pass ran in a reduced (retried) form; coverage there is good but not exhaustive.*

strands-py/src/strands/multiagent/a2a/executor.py
marcinmarzeta: Applied as suggested in 76034b4 — scalar-only guard with `math.isfinite`, comment reworded to the silent-null mechanism, and the dict param now asserts the preserved shape.

strands-py/tests/strands/multiagent/a2a/test_executor.py
marcinmarzeta: Added in 76034b4, with `import json` moved to the top-level import block.

PR #3833, opened 2026-08-15
feat(anthropic): route AnthropicModel through the Bedrock Mantle endpoint

### LINKED ISSUE #3832: [FEATURE] Route AnthropicModel through the Amazon Bedrock Mantle endpoint
### SDK Language

Python

### Problem Statement

Amazon Bedrock serves Claude models through the Anthropic Messages API on the `bedrock-mantle` endpoint (`https://bedrock-mantle.{region}.api.aws/anthropic`), billing through the caller's AWS account rather than an Anthropic API key. `AnthropicModel` has no configuration for it.

You can get partway there today by hand-building the endpoint in `client_args`:

```python
AnthropicModel(
    model_id="anthropic.claude-sonnet-5",
    max_tokens=1028,
    client_args={
        "api_key": "<BEDROCK_API_KEY>",
        "base_url": "https://bedrock-mantle.us-east-1.api.aws/anthropic",
    },
)
```

That reaches the endpoint, but it authenticates only with a static API key string. Bedrock signs requests with SigV4 from the standard AWS credential chain, which is what makes instance roles, SSO sessions, and assumed roles usable. There is no way to select that path through `AnthropicModel`: `client_args` is forwarded to `anthropic.AsyncAnthropic`, and the SigV4-capable client is a different class (`anthropic.AsyncAnthropicBedrockMantle`, added in `anthropic` 0.91.0). The provider does not accept a pre-built client either, so the class cannot be swapped from outside.

`OpenAIModel` and `OpenAIResponsesModel` already expose `bedrock_mantle_config` for the OpenAI-compatible side of the same endpoint. The Anthropic-compatible side has no equivalent.

### Proposed Solution

Add a `bedrock_mantle_config` option to `AnthropicModel`, matching the name and shape the OpenAI providers already use:

```python
AnthropicModel(
    model_id="anthropic.claude-sonnet-5",
    max_tokens=1028,
    bedrock_mantle_config={"region": "us-east-1"},
)
```

When set, the provider builds `anthropic.AsyncAnthropicBedrockMantle` instead of the direct client. That client derives the base URL from the region, sends the required `anthropic-version` header, and signs every request with SigV4, so the SDK never has to mint or refresh a bearer token. The config accepts `region`, `profile`, and `api_key`; supplying none of them falls back to the standard AWS credential chain.

Since the region is interpolated into the endpoint URL, it should go through the same `validate_region()` guard the OpenAI path already uses.

### Use Case

- Running Claude models against an AWS account's existing Bedrock spend and quotas instead of a separate Anthropic API key.
- Agents on EC2/ECS/Lambda authenticating with the instance or task role, with no long-lived credential to distribute or rotate.
- Local development against an SSO profile.
- Deployments that need in-region inference for data residency.

### Alternatives Solutions

- Document the `client_args` base URL workaround instead of adding config. No code change, but it stays static-API-key-only and asks every user to assemble the endpoint URL by hand.
- Accept a pre-built client on `AnthropicModel`. More general, but a larger API change than this needs, and it pushes the region, base URL, and version-header details onto every caller.

### Additional Context

Endpoint reference: https://docs.aws.amazon.com/bedrock/latest/userguide/inference-messages-api.html

The Mantle catalog uses `anthropic.`-prefixed model IDs that differ from the direct API (`anthropic.claude-sonnet-5`, `anthropic.claude-opus-4-8`, `anthropic.claude-haiku-4-5`, and others), and the available set varies by region.

TypeScript needs the same option, tracked separately as a port.

moritalous:
## Description

Amazon Bedrock serves Claude models through the Anthropic Messages API on the `bedrock-mantle` endpoint, billing through the caller's AWS account rather than an Anthropic API key. Reaching it from `AnthropicModel` today means assembling the region-specific base URL by hand in `client_args` and authenticating with a static API key string; the SigV4 path that makes instance roles, SSO sessions, and assumed roles usable is unreachable, because `client_args` is forwarded to `anthropic.AsyncAnthropic` and the SigV4-capable client is a different class.

This adds `bedrock_mantle_config` to `AnthropicModel`, matching the option `OpenAIModel` and `OpenAIResponsesModel` already expose for the OpenAI-compatible side of the same endpoint.

```python
from strands import Agent
from strands.models.anthropic import AnthropicModel

model = AnthropicModel(
    model_id="anthropic.claude-sonnet-5",
    max_tokens=1028,
    bedrock_mantle_config={"region": "us-east-1"},
)

agent = Agent(model=model)
print(agent("What is 2+2?"))
```

`BedrockMantleConfig` accepts `region`, `profile`, and `api_key`, all optional. Requests are signed with SigV4 unless `api_key` is supplied, using the named profile or the standard AWS credential chain. Omitting `region` resolves it through boto3, so a profile-configured region is picked up. Anything else — `timeout`, `max_retries`, `default_headers`, `http_client`, explicit static credentials — passes through `client_args` unchanged; the three arguments this config derives (`aws_region`, `aws_profile`, `api_key`) are rejected there so the two sources cannot disagree.

Two implementation notes worth a reviewer's attention:

**This deliberately does not share code with `_openai_bedrock.py`.** The OpenAI pathway mints and refreshes a bearer token because the `openai` package knows nothing about AWS. The `anthropic` package ships `AsyncAnthropicBedrockMantle`, which derives the base URL, sends the required `anthropic-version` header, and signs each request itself, so there is no token to mint and the client is still built once in `__init__`. The only genuinely common piece is `validate_region()`, which is already shared from `_validation.py`. The two `_resolve_region` helpers look similar but resolve different config shapes (`boto_session` versus a profile name).

**Region validation is a security guard, not a convenience.** The region is interpolated into the endpoint URL, so a malformed value containing `@`, `:`, `/`, or `#` could re-point a signed request at a non-AWS host. It goes through the same `validate_region()` the OpenAI path uses, on both the explicit and the boto3-resolved value.

The `anthropic` extra floor moves from `>=0.21.0` to `>=0.91.0`, the release that introduced the Mantle client. Users pinned below that will be upgraded by `strands-agents[anthropic]`.

## Related Issues

Resolves #3832

## Documentation PR

The provider page gains an "Amazon Bedrock (Mantle)" section under `site/`, documenting
this option, the model IDs the endpoint serves, and how to pick an authentication mode.

The sibling PR adds the same section for the other SDK. Both introduce the section
independently so either can merge first; whichever merges second resolves the overlap by
keeping both language tabs in the one section.

## Type of Change

New feature

## Testing

Unit tests cover the client swap, the argument mapping, `client_args` composition, the rejection of derived arguments, region resolution from an explicit value / the boto3 chain / a named profile, and rejection of a malformed region from either source.

Integration tests in `tests_integ/models/test_model_mantle.py` exercise the live `bedrock-mantle.us-east-1.api.aws/anthropic` endpoint: an agent invocation, tool-based structured output, and the native `count_tokens` path. All three pass against `anthropic.claude-sonnet-5`.

I also ran the documented example verbatim against the live endpoint, and confirmed the request reaches `POST /anthropic/v1/messages` with `anthropic-version: 2023-06-01`.

Not exactly: `hatch run prepare` starts with `hatch run format`, which reformats the whole package and dirties 13 files unrelated to this change. I ran the stages individually instead — `hatch run lint`, `hatch run test-lint`, and `hatch test` (5261 passed) — plus `.agents/skills/pre-push/run-checks.sh`, which reports all local checks passing. `hatch test --all` could not complete locally: the Python 3.14 environment fails to import `charset_normalizer`, which is unrelated to this change.

## Checklist

----

### DIFF
--- a/site/src/content/docs/user-guide/concepts/model-providers/anthropic.mdx
+++ b/site/src/content/docs/user-guide/concepts/model-providers/anthropic.mdx
@@ -5,6 +5,7 @@ tags: [structured-output, caching]
 integrationType: model-provider
 sourceLinks:
   - path: strands-py/src/strands/models/anthropic.py
+  - path: strands-py/src/strands/models/_anthropic_bedrock.py
   - path: strands-ts/src/models/anthropic.ts
 ---
 
@@ -67,6 +68,44 @@ print(response)
 ```
 </Tab>
 </Tabs>
+
+### Amazon Bedrock (Mantle)
+
+Claude models also run on
+[Amazon Bedrock's Anthropic-compatible endpoint](https://docs.aws.amazon.com/bedrock/latest/userguide/inference-messages-api.html)
+powered by Mantle, which bills through your AWS account instead of an Anthropic API key.
+Give the provider a region and it signs every request with SigV4 from the standard AWS
+credential chain.
+
+Model IDs on Mantle differ from the direct API. The endpoint serves
+`anthropic.claude-sonnet-5`, `anthropic.claude-opus-4-8`, and `anthropic.claude-haiku-4-5`,
+among others.
+
+<Tabs>
+<Tab label="Python">
+
+```python
+from strands import Agent
+from strands.models.anthropic import AnthropicModel
+
+model = AnthropicModel(
+    model_id="anthropic.claude-sonnet-5",
+    max_tokens=1028,
+    bedrock_mantle_config={"region": "us-east-1"},
+)
+
+agent = Agent(model=model)
+response = agent("What is 2+2?")
+print(response)
+```
+</Tab>
+</Tabs>
+
+Omit the region to resolve it from the `AWS_REGION` environment variable. To authenticate
+with a named profile or a
+[Bedrock API key](https://docs.aws.amazon.com/bedrock/latest/userguide/api-key-management.html)
+instead of the default credential chain, add `profile` or
+<Syntax py="api_key" ts="apiKey" /> to the same config.
 
 ## Configuration
 
--- a/strands-py/src/strands/models/_anthropic_bedrock.py
+++ b/strands-py/src/strands/models/_anthropic_bedrock.py
@@ -1,0 +1,107 @@
+"""Internal helpers for routing the Anthropic client to Bedrock Mantle.
+
+Converts a ``bedrock_mantle_config`` dict into the keyword arguments that
+``anthropic.AsyncAnthropicBedrockMantle`` consumes. That client derives the Mantle base URL
+from the region, sends the required ``anthropic-version`` header, and signs every request
+with SigV4 on its own, so nothing here mints or caches a credential.
+"""
+
+from __future__ import annotations
+
+from typing import Any, TypedDict
+
+import boto3
+
+from ._validation import validate_region
+
+_MANTLE_DOCS_URL = "https://docs.aws.amazon.com/bedrock/latest/userguide/inference-messages-api.html"
+
+# Client kwargs this module derives from the config. A caller that also accepts raw client
+# args must reject these, or the two sources would silently disagree.
+MANTLE_DERIVED_CLIENT_ARGS = ("aws_region", "aws_profile", "api_key")
+
+
+class BedrockMantleConfig(TypedDict, total=False):
+    """Config for routing the Anthropic client through Bedrock Mantle.
+
+    Attributes:
+        region: AWS region hosting the Bedrock Mantle endpoint. If omitted, resolved from
+            ``profile`` (if provided) or the standard boto3 chain (``AWS_REGION`` /
+            ``AWS_DEFAULT_REGION`` / active profile / EC2 metadata). A :class:`ValueError`
+            is raised if none resolve.
+        profile: Optional AWS profile to authenticate with. Selects SigV4 authentication.
+        api_key: Optional Amazon Bedrock API key. When set, requests carry it as a bearer
+            token instead of being signed with SigV4. Omit to use the standard AWS
+            credential chain.
+    """
+
+    region: str
+    profile: str
+    api_key: str
+
+
+def _resolve_region(config: BedrockMantleConfig) -> str:
+    """Resolve the AWS region, preferring explicit config then falling back to boto3.
+
+    The resolved region is validated before it is returned, since
+    ``AsyncAnthropicBedrockMantle`` interpolates it into the Mantle endpoint URL.
+
+    Args:
+        config: The Bedrock Mantle config to resolve the region from.
+
+    Returns:
+        The resolved AWS region.
+
+    Raises:
+        ValueError: If no region can be resolved from the config, the named profile, or the
+            standard boto3 credential chain, or if the resolved region is not a well-formed
+            AWS region identifier.
+    """
+    region = config.get("region")
+    if region:
+        return validate_region(region)
+
+    # ``boto3.Session()`` reads ``AWS_REGION`` / ``AWS_DEFAULT_REGION``, the active profile,
+    # and EC2 instance metadata. The Anthropic client's own fallback reads only the two env
+    # vars, so a profile-configured region would otherwise go unseen.
+    profile = config.get("profile")
+    session = boto3.Session(profile_name=profile) if profile else boto3.Session()
+    if session.region_name:
+        return validate_region(str(session.region_name))
+
+    raise ValueError(
+        "Could not resolve an AWS region for Bedrock Mantle. Pass 'region' in "
+        "bedrock_mantle_config, configure a region on the named profile, or set AWS_REGION "
+        f"in the environment. See {_MANTLE_DOCS_URL} for supported regions."
+    )
+
+
+def resolve_bedrock_client_args(
+    config: BedrockMantleConfig, client_args: dict[str, Any] | None = None
+) -> dict[str, Any]:
+    """Resolve a ``BedrockMantleConfig`` (plus optional ``client_args``) into Mantle client kwargs.
+
+    Callers are expected to validate that ``client_args`` carries none of
+    :data:`MANTLE_DERIVED_CLIENT_ARGS` before calling this function (typically at
+    ``__init__`` time for fail-fast behavior).
+
+    Args:
+        config: Config for routing the Anthropic client through Bedrock Mantle.
+        client_args: Additional arguments for the underlying Anthropic client.
+
+    Returns:
+        Keyword arguments for ``anthropic.AsyncAnthropicBedrockMantle``.
+
+    Raises:
+        ValueError: If no region can be resolved.
+    """
+    resolved: dict[str, Any] = dict(client_args or {})
+    resolved["aws_region"] = _resolve_region(config)
+
+    # Only forward keys the user set; the Anthropic client reads auth precedence off which
+    # arguments are not None, so passing an explicit None would change the selected mode.
+    if "profile" in config:
+        resolved["aws_profile"] = config["profile"]
+    if "api_key" in config:
+        resolved["api_key"] = config["api_key"]
+    return resolved
--- a/strands-py/src/strands/models/anthropic.py
+++ b/strands-py/src/strands/models/anthropic.py
@@ -21,6 +21,7 @@ from ..types.event_loop import Usage
 from ..types.exceptions import ContextWindowOverflowException, ModelThrottledException
 from ..types.streaming import StreamEvent
 from ..types.tools import ToolChoice, ToolChoiceToolDict, ToolSpec
+from ._anthropic_bedrock import MANTLE_DERIVED_CLIENT_ARGS, BedrockMantleConfig, resolve_bedrock_client_args
 from ._defaults import resolve_config_metadata
 from ._validation import _has_location_source, validate_config_keys
 from .model import BaseModelConfig, CacheConfig, CacheToolsConfig, Model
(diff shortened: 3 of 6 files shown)


strands-py/src/strands/models/_anthropic_bedrock.py
```diff
@@ -0,0 +1,107 @@
...
+
+import boto3
+
+from ._validation import validate_region
+
+_MANTLE_DOCS_URL = "https://docs.aws.amazon.com/bedrock/latest/userguide/inference-messages-api.html"
+
+# Client kwargs this module derives from the config. A caller that also accepts raw client
+# args must reject these, or the two sources would silently disagree.
+MANTLE_DERIVED_CLIENT_ARGS = ("aws_region", "aws_profile", "api_key")
```
arielnabavian: **concern** — `base_url` belongs on this list. It is the one thing `bedrock_mantle_config` exists to produce, and the Mantle client only produces it when the caller passes nothing: `_resolve_mantle_config` in `anthropic/lib/bedrock/_mantle.py` builds `https://bedrock-mantle.{region}.api.aws/anthropic` **only** `if base_url is None`.

So this constructs without error:

```python
AnthropicModel(
    model_id="anthropic.claude-sonnet-5",
    max_tokens=1028,
    client_args={"base_url": "https://not-aws.example"},
    bedrock_mantle_config={"region": "us-east-1"},
)
```

The derived endpoint is discarded, `use_sigv4` is still `True`, and `_prepare_request` then SigV4-signs every request with the caller's real credentials and sends it to `not-aws.example`. That is precisely the failure `validate_region()`'s own docstring is written to prevent — "can re-point a signed request to a non-AWS host, exfiltrating credentials" — reached through a door the guard doesn't cover. If a region is untrusted enough to need a regex, a `base_url` from the same source is too.

`_openai_bedrock.py` already rejects `base_url` next to `api_key`, so this is a divergence from the sibling path as well as a hole.

`skip_auth=True` has the same shape (silently stops signing), though at least it says what it does on the tin.

The constant's name is what's in the way: `base_url` isn't derived *here*, it's derived by the Anthropic client. Renaming to something like `MANTLE_REJECTED_CLIENT_ARGS` and adding `base_url` makes the name match what the check is actually for.

strands-py/src/strands/models/anthropic.py
```diff
@@ -88,21 +89,51 @@ class AnthropicConfig(BaseModelConfig, total=False):
...
         """Initialize provider instance.
 
         Args:
             client_args: Arguments for the underlying Anthropic client (e.g., api_key).
                 For a complete list of supported arguments, see https://docs.anthropic.com/en/api/client-sdks.
+                May be combined with ``bedrock_mantle_config``; when both are set, the config
+                derives the AWS arguments, which must not appear in ``client_args``.
+            bedrock_mantle_config: Route requests through Amazon Bedrock's Mantle
+                (Anthropic-compatible) endpoint. See :class:`BedrockMantleConfig` for accepted
+                keys. Requests are signed with SigV4 unless the config supplies an API key,
```
arielnabavian: **concern** — "signed with SigV4 unless the config supplies an API key" isn't what the client does. `resolve_auth_mode()` in `anthropic/lib/aws/_credentials.py` has a rung between the config and the credential chain:

1. `api_key` arg → bearer token
2. `aws_access_key`/`aws_secret_key` → SigV4
3. `aws_profile` → SigV4
4. **`AWS_BEARER_TOKEN_BEDROCK` or `ANTHROPIC_AWS_API_KEY` in the environment → bearer token**
5. default AWS credential chain → SigV4

`AWS_BEARER_TOKEN_BEDROCK` is the env var AWS documents for Bedrock API keys, so plenty of users already have it exported. With it set, `bedrock_mantle_config={"region": "us-east-1"}` authenticates with that token and never touches SigV4 — and if the token is expired or scoped to a different account, they get a 401 with nothing in the docs to explain it. The docs page repeats the claim (`anthropic.mdx:77-78`), as does the PR description.

The client offers no way to force SigV4 short of passing `aws_profile`, so stating the real precedence is probably the honest fix rather than trying to override it.

Two knock-ons:

- `profile: ... Selects SigV4 authentication` (`_anthropic_bedrock.py:32`) is also wrong when the same config sets `api_key` — rung 1 wins and `aws_profile` is silently ignored. Either reject that pair or document which one loses.
- `tests_integ/models/test_model_mantle.py:5-6` says "Credentials come from the ambient AWS credential chain; no explicit API key is passed by the user," but nothing pins that. On a machine with `AWS_BEARER_TOKEN_BEDROCK` exported, the new tests pass over the bearer-token path and the SigV4 path this PR is *for* is never exercised. A `monkeypatch.delenv` on both vars in the fixture would make the tests mean what the docstring says.

strands-py/src/strands/models/_anthropic_bedrock.py
```diff
@@ -0,0 +1,107 @@
...
+    resolved: dict[str, Any] = dict(client_args or {})
+    resolved["aws_region"] = _resolve_region(config)
+
+    # Only forward keys the user set; the Anthropic client reads auth precedence off which
+    # arguments are not None, so passing an explicit None would change the selected mode.
+    if "profile" in config:
+        resolved["aws_profile"] = config["profile"]
+    if "api_key" in config:
+        resolved["api_key"] = config["api_key"]
+    return resolved
```
arielnabavian: **concern** — nothing validates the keys of `config`. Anything that isn't `region`, `profile`, or `api_key` is dropped without a warning, and `TypedDict` gives you no runtime protection. That's a quiet failure with an expensive shape here, because this config's keys are disjoint from the same-named config on the OpenAI side (`region`, `boto_session`, `credentials_provider`, `expiry`):

```python
# valid on OpenAIModel; a silent no-op on AnthropicModel
AnthropicModel(
    model_id="anthropic.claude-sonnet-5",
    max_tokens=1028,
    bedrock_mantle_config={"boto_session": boto3.Session(profile_name="prod")},
)
```

No error, no warning. `_resolve_region` never sees a profile, so it falls through to bare `boto3.Session()` and the request is signed with the **default** profile's credentials in the **default** region. Wrong account, wrong region, no signal. An ordinary typo — `profile_name=`, `aws_profile=` — does the same thing.

The repo already has the tool, and `__init__` uses it on `model_config` a few lines earlier:

```python
validate_config_keys(bedrock_mantle_config, BedrockMantleConfig)
```

The OpenAI path has the same gap, so you could call it pre-existing. It bites harder here: there a missing key means "no session," here it means "someone else's credentials."

strands-py/src/strands/models/_anthropic_bedrock.py
```diff
@@ -0,0 +1,107 @@
...
+from ._validation import validate_region
+
+_MANTLE_DOCS_URL = "https://docs.aws.amazon.com/bedrock/latest/userguide/inference-messages-api.html"
+
+# Client kwargs this module derives from the config. A caller that also accepts raw client
+# args must reject these, or the two sources would silently disagree.
+MANTLE_DERIVED_CLIENT_ARGS = ("aws_region", "aws_profile", "api_key")
+
+
+class BedrockMantleConfig(TypedDict, total=False):
```
arielnabavian: **consider** — this is the decision that gets expensive after release. There are now two public `BedrockMantleConfig` TypedDicts, reached through the same `bedrock_mantle_config=` kwarg on providers people swap between, sharing exactly one key:

| key | OpenAI | Anthropic |
|---|---|---|
| `region` | ✓ | ✓ |
| `boto_session` | ✓ | — |
| `credentials_provider` | ✓ | — |
| `expiry` | ✓ | — |
| `profile` | — | ✓ |
| `api_key` | — | ✓ |

The description says this "matches the option `OpenAIModel` and `OpenAIResponsesModel` already expose." It matches the name, not the shape. And `tests_integ/models/test_model_mantle.py:23-25` shares one `bedrock_mantle_config` fixture across all four models — which only works because `region` happens to be the intersection, so the tests can't see the divergence either.

The pair that will hurt is `profile` vs `boto_session`: both mean "authenticate as somebody other than the default," and once shipped you support both spellings in both places forever.

I don't think `boto_session` here is obviously right — `AsyncAnthropicBedrockMantle` takes `aws_profile` or static keys, not a session, so supporting it means reading `region_name` off the session and passing `session.get_credentials()` through as `aws_access_key`/`aws_secret_key`/`aws_session_token`, which snapshots the credentials and loses the refresh this design otherwise gets for free. That's a real argument for `profile`.

So: fine if the answer is "they stay different" — but then say it in the docstring instead of leaving users to infer parity from the shared name and the shared kwarg.

strands-py/tests_integ/models/test_model_mantle.py
```diff
@@ -103,3 +107,53 @@ def test_reasoning_content_multi_turn(bedrock_mantle_config):
...
+
+    class Weather(pydantic.BaseModel):
+        time: str
+        weather: str
+
+    agent = Agent(model=anthropic_model, callback_handler=None)
+
+    result = agent("The time is 12:00 and the weather is sunny", structured_output_model=Weather)
+
+    assert result.structured_output == Weather(time="12:00", weather="sunny")
```
arielnabavian: **concern** — `cache_config` / `cache_tools` go untested over Mantle, and they're the provider's other headline feature (`anthropic.mdx` frontmatter: `tags: [structured-output, caching]`). The three new tests all cover the plain path, but the first thing someone does after pointing an existing `AnthropicModel` at `bedrock_mantle_config` is find out whether their prompt caching survived.

This isn't hypothetical: the docstring right above notes that Mantle rejects `output_config.format`, which is itself evidence that Mantle's feature surface diverges from the direct API. `cache_control: {"type": "ephemeral"}` blocks either work or come back as a 400, and right now neither this file nor the new docs section says which.

One more test with `cache_config=CacheConfig(strategy="auto")` settles it. If caching isn't supported on Mantle, the docs section should say so — quietly 400-ing on a documented feature is worse than a known gap.

site/src/content/docs/user-guide/concepts/model-providers/anthropic.mdx
```diff
@@ -68,6 +69,44 @@ print(response)
...
+)
+
+agent = Agent(model=model)
+response = agent("What is 2+2?")
+print(response)
+```
+</Tab>
+</Tabs>
+
+Omit the region to resolve it from the `AWS_REGION` environment variable. To authenticate
```
arielnabavian: **nit** — "resolve it from the `AWS_REGION` environment variable" undersells `_resolve_region` and misleads the exact case this sentence goes on to describe. It builds a `boto3.Session`, with `profile_name` when the config names one, so `AWS_DEFAULT_REGION` and a region configured on the profile both work.

As written, someone using `profile` reads this and concludes they must also export `AWS_REGION` — which is the opposite of why the profile branch exists (`_anthropic_bedrock.py:64-68`).

PR #3837, opened 2026-08-15
feat(ts): support multiple continuation inputs

gautamsirdeshmukh:
## Description

Adds an internal continuation mechanism that lets multiple hooks contribute ordered input without competing for `AfterInvocationEvent.resume`.

Contributions can be added after an invocation for the next agent pass or before a model call for the current request. Each contribution is incorporated into agent history or abandoned independently, while existing public `resume` behavior remains unchanged.

## Related Issues

None.

## Documentation PR

Not applicable. This change is internal.

## Type of Change

New feature

## Testing

- [x] TypeScript unit tests and repository validation checks

## Checklist

----

### DIFF
--- a/strands-ts/src/agent/__tests__/agent.continuation.test.ts
+++ b/strands-ts/src/agent/__tests__/agent.continuation.test.ts
@@ -1,6 +1,8 @@
 import { describe, expect, it, vi } from 'vitest'
 import { MockMessageModel } from '../../__fixtures__/mock-message-model.js'
+import { createMockTool } from '../../__fixtures__/tool-helpers.js'
 import { AfterInvocationEvent, BeforeModelCallEvent, MessageAddedEvent } from '../../hooks/events.js'
+import { InterruptResponseContent } from '../../types/interrupt.js'
 import { Message, TextBlock, ToolResultBlock, ToolUseBlock } from '../../types/messages.js'
 import { Agent } from '../agent.js'
 import { continuations } from '../continuation.js'
@@ -59,6 +61,59 @@ describe('Agent continuation input', () => {
     expect(abandoned).toHaveBeenCalledWith(
       expect.objectContaining({ message: 'Continuation input must contain a complete message sequence' })
     )
+  })
+
+  it('retains follow-up input through failed resume attempts', async () => {
+    const model = new MockMessageModel()
+      .addTurn({
+        type: 'toolUseBlock',
+        name: 'confirmTool',
+        toolUseId: 'tool-1',
+        input: {},
+      })
+      .addTurn({ type: 'textBlock', text: 'resumed' })
+      .addTurn({ type: 'textBlock', text: 'continued' })
+    const requests = captureRequests(model)
+    const appended = vi.fn()
+    const abandoned = vi.fn()
+    const tool = createMockTool('confirmTool', (context) => {
+      const response = context.interrupt<string>({ name: 'confirm', reason: 'Confirm?' })
+      return `confirmed:${response}`
+    })
+    const agent = new Agent({ model, tools: [tool], printer: false })
+    let added = false
+
+    agent.addHook(AfterInvocationEvent, (event) => {
+      if (added) return
+      added = true
+      continuations.addInput(event, {
+        args: 'pending',
+        onAppended: appended,
+        onAbandoned: abandoned,
+      })
+    })
+
+    const interruptResult = await agent.invoke('start')
+
+    expect(interruptResult.stopReason).toBe('interrupt')
+    expect(appended).not.toHaveBeenCalled()
+    expect(abandoned).not.toHaveBeenCalled()
+
+    await expect(agent.invoke('invalid resume')).rejects.toThrow('Agent is in an interrupted state')
+    expect(appended).not.toHaveBeenCalled()
+    expect(abandoned).not.toHaveBeenCalled()
+
+    const finalResult = await agent.invoke([
+      new InterruptResponseContent({
+        interruptId: interruptResult.interrupts![0]!.id,
+        response: 'yes',
+      }),
+    ])
+
+    expect(finalResult.stopReason).toBe('endTurn')
+    expect(textOf(requests[2]!.at(-1)!)).toBe('pending')
+    expect(appended).toHaveBeenCalledOnce()
+    expect(abandoned).not.toHaveBeenCalled()
   })
 
   it('preserves an unrecognized stop reason instead of continuing', async () => {
--- a/strands-ts/src/agent/__tests__/agent.continuation.test.ts
+++ b/strands-ts/src/agent/__tests__/agent.continuation.test.ts
@@ -61,6 +61,37 @@ describe('Agent continuation input', () => {
     )
   })
 
+  it('preserves an unrecognized stop reason instead of continuing', async () => {
+    // Unknown provider stop reasons must remain terminal (#3837).
+    const stopReason = 'providerSpecificStop'
+    const model = new MockMessageModel()
+      .addTurn({ type: 'textBlock', text: 'partial' }, { stopReason })
+      .addTurn({ type: 'textBlock', text: 'unreachable' })
+    const abandoned = vi.fn()
+    const agent = new Agent({ model, printer: false })
+    let added = false
+
+    agent.addHook(AfterInvocationEvent, (event) => {
+      if (added) return
+      added = true
+      continuations.addInput(event, { args: 'pending', onAbandoned: abandoned })
+    })
+
+    const result = await agent.invoke('start')
+
+    expect({
+      stopReason: result.stopReason,
+      modelCalls: model.callCount,
+      messages: agent.messages.map(textOf),
+      abandoned: abandoned.mock.calls.length,
+    }).toEqual({
+      stopReason,
+      modelCalls: 1,
+      messages: ['start', 'partial'],
+      abandoned: 1,
+    })
+  })
+
   it('appends a complete tool exchange contributed before the model call', async () => {
     const model = textModel('final')
     const requests = captureRequests(model)
--- a/strands-ts/src/agent/__tests__/agent.continuation.test.ts
+++ b/strands-ts/src/agent/__tests__/agent.continuation.test.ts
@@ -1,23 +1,9 @@
 import { describe, expect, it, vi } from 'vitest'
 import { MockMessageModel } from '../../__fixtures__/mock-message-model.js'
-import { createMockTool } from '../../__fixtures__/tool-helpers.js'
-import {
-  AfterInvocationEvent,
-  BeforeInvocationEvent,
-  BeforeModelCallEvent,
-  MessageAddedEvent,
-} from '../../hooks/events.js'
-import { AgentStreamStage, InvokeModelStage } from '../../middleware/stages.js'
-import { AgentResult } from '../../types/agent.js'
+import { AfterInvocationEvent, BeforeModelCallEvent, MessageAddedEvent } from '../../hooks/events.js'
 import { Message, TextBlock, ToolResultBlock, ToolUseBlock } from '../../types/messages.js'
-import { logger } from '../../logging/logger.js'
 import { Agent } from '../agent.js'
-import { continuations as continuationManager } from '../continuation.js'
-
-import type { InvokeArgs } from '../../types/agent.js'
-import type { AgentStreamContext, AgentStreamResult } from '../../middleware/stages.js'
-import type { StopReason } from '../../types/messages.js'
-import type { ContinuationInput } from '../continuation.js'
+import { continuations } from '../continuation.js'
 
 function textOf(message: Message): string {
   return message.content.flatMap((block) => (block.type === 'textBlock' ? [block.text] : [])).join('')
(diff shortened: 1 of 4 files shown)


strands-ts/src/agent/continuation.ts
```diff
@@ -0,0 +1,227 @@
...
+  result: AgentResult,
+  invocationState: InvocationState
+): Promise<readonly MessageAddedEvent[]> {
+  if (!invocation || invocation.settled) return []
+  if (CONTINUATION_REJECTION_STOP_REASONS.has(result.stopReason)) {
+    await rejectInvocation(invocation, new Error(`Continuation stopped by ${result.stopReason}`))
+    return []
+  }
+
+  const events = [...(await publish(agent, invocation, undefined, invocationState))]
```
strandly-the-agent: 🔴 **Deferred publish corrupts history: `result.lastMessage` gets re-pushed when the splice reorders the tail** — and it reproduces with the public API alone (`resume` + a `BeforeModelCallEvent.cancel` hook), no `_continueWith` needed.

On a follow-up pass that never reaches the provider, `_stream` appends the input message and then the assistant deny message; `publish` only runs afterwards (from `settleResult`) and splices the input to the **end** of `agent.messages` — behind the assistant reply — so `at(-1)` ≠ `lastMessage` and the same `Message` object is pushed a second time:

```
PR  : ['start','initial','model denied','public follow-up','model denied']   ← duplicate trackingId
main: ['start','initial','public follow-up','model denied']
```

The duplicate is then resent to the provider on the next pass, and duplicated `trackingId`s are exactly what `buildRequest`/`integrateMessages`' own filtering and session snapshots key on. Verified: the repro below fails on this branch and passes on `origin/main`. Same corruption via the internal channel, and a second instance via an `AgentStreamStage` middleware short-circuiting with an older `lastMessage`.

Fix direction: guard on membership (`trackingId` already present anywhere in history) instead of `at(-1)`, and avoid reordering when publish is deferred. A regression test asserting no duplicate `trackingId`s after a Before-cancel on a follow-up pass would pin it — nothing currently covers this path, and `settleResult`'s `settled` guard is itself untested (a mutant deleting it survives 54/54).

<details><summary>Repro (drop into src/agent/__tests__)</summary>

```ts
import { describe, expect, it } from 'vitest'
import { MockMessageModel } from '../../__fixtures__/mock-message-model.js'
import { AfterInvocationEvent, BeforeModelCallEvent } from '../../hooks/events.js'
import { Message } from '../../types/messages.js'
import { Agent } from '../agent.js'

const textOf = (m: Message) => m.content.flatMap((b) => (b.type === 'textBlock' ? [b.text] : [])).join('')

it('keeps history intact when the follow-up pass is model-cancelled', async () => {
    .addTurn({ type: 'textBlock', text: 'initial' })
    .addTurn({ type: 'textBlock', text: 'unreachable' })
  const agent = new Agent({ model, printer: false })
  let armed = false
  let modelCalls = 0
  agent.addHook(BeforeModelCallEvent, (event) => {
    if (++modelCalls === 2) event.cancel = 'model denied'
  })
  agent.addHook(AfterInvocationEvent, (event) => {
    if (armed) return
    armed = true
    event.resume = 'public follow-up'
  })

  await agent.invoke('start')

  expect(agent.messages.map(textOf)).toEqual(['start', 'initial', 'public follow-up', 'model denied'])
  const ids = agent.messages.map((m) => m.trackingId)
  expect(ids.length - new Set(ids).size).toBe(0) // PR head: 1
})
```
</details>

strands-ts/src/agent/continuation.ts
```diff
@@ -0,0 +1,227 @@
...
+  }
+}
+
+function buildRequest(history: readonly Message[], invocation: InvocationContinuation | undefined): Message[] {
+  if (!invocation || invocation.settled) return [...history]
+
+  const publicIds = new Set(invocation.publicMessages.map((message) => message.trackingId))
+  const baseMessages = history.filter((message) => !publicIds.has(message.trackingId))
+  const internalMessages = invocation.internal?.flatMap((entry) => entry.messages) ?? []
+  return [...baseMessages, ...internalMessages, ...invocation.publicMessages]
```
strandly-the-agent: 🟡 **The composed request can violate provider role-alternation — in exactly the multi-contribution case this mechanism exists for.** Every contribution normalizes to ≥1 `user`-role message (agent.ts:1983-2032) and history ends with the assistant message, so N contributions (+ `resume`) yield N(+1) consecutive `user` messages; nothing coalesces or validates them (`MockMessageModel` can't catch it, and the test at agent.continuation.test.ts:86-95 locks the shape in). The repo treats alternation as a hard constraint elsewhere: `message-injection.ts:100-105` folds injected text to “keep role alternation valid”, and `sliding-window-conversation-manager.ts:485-491` preserves it “for providers that require it (#2085/#2087)”.

First time two Background Tasks finish during one invocation → next request ends `…assistant, user, user` → Bedrock `ValidationException`, every contribution `onRejected`, and the completed first-pass answer is lost through the throw path. Same with one contribution + GoalLoop's `resume`. Only the single-contribution-no-resume case is legal today.

Suggestion: coalesce the composed tail into a single user message (the `foldIntoLastUserMessage` approach), or validate/coalesce here in `buildRequest`.

strands-ts/src/agent/agent.ts
```diff
@@ -1129,59 +1152,110 @@ export class Agent implements LocalAgent, InvokableAgent {
...
-        // Only emit AgentResultEvent on the final iteration (not on resumed ones).
-        yield await this._invokeCallbacks(
-          new AgentResultEvent({
-            agent: this,
-            result: result!,
-            invocationState,
-          })
-        )
+        const limitStopReason =
+          pendingContinuations.length > 0 && passRanAgentLoop ? this._checkLimits(currentOptions) : undefined
```
strandly-the-agent: 🟡 **Pending contributions are gated on limits here, but not on the pass's stop reason** — `interrupt` / `cancelled` / `checkpoint` launch a follow-up pass off a terminal state.

Verified probe: an interrupting tool + one `_continueWith` in an After hook makes `invoke()` throw `TypeError: Agent is in an interrupted state…` (from the follow-up pass, agent.ts:1541) instead of returning the `stopReason: 'interrupt'` result — the `interrupts` payload the caller needs is gone and the agent is left activated-but-unaddressable. Any HITL setup hits this the moment one plugin contributes on an invocation another plugin interrupted; the two need not know about each other. (`cancelled` merely burns a superfluous extra pass + full hook round; a checkpoint/interrupt-response `resume` plus a contribution turns a *valid* resume into a thrown TypeError via continuation.ts:61-68.)

Suggested fix mirrors this gate: `if (CONTINUATION_REJECTION_STOP_REASONS.has(completedResult.stopReason))` → `rejectInputs(pendingContinuations, …)` and fall through to the normal return.

strands-ts/src/agent/continuation.ts
```diff
@@ -0,0 +1,227 @@
...
+  updateArgs,
+}
+
+async function prepareInvocation(
+  inputs: readonly ContinuationInput[],
+  resume: InvokeArgs | undefined,
+  normalizeInput: NormalizeInput
+): Promise<InvocationContinuation> {
+  const internal = await prepareInputs(inputs, normalizeInput)
+  const publicMessages = resume === undefined ? [] : normalizeInput(resume)
```
strandly-the-agent: 🟡 **Settlement leak: if `normalizeInput(resume)` throws, the already-prepared contributions are dropped with *neither* callback firing.** `prepareInputs` has already run; this line executes outside any try. Verified probe (malformed resume content block): `invoke()` rejects with the normalize error and the contribution's settle log is empty — `stream()` has zeroed `pendingContinuations` by then and the outer `finally`'s `cleanup` no-ops on the previous, settled pass. A `resume` writer (e.g. GoalLoop) handing bad data thus breaks every pending contributor's only cleanup hook.

Fix: wrap this in the same try/reject pattern `prepareInputs` uses — `rejectInputs(internal.map((e) => e.input), error)` before rethrowing.

strands-ts/src/agent/continuation.ts
```diff
@@ -0,0 +1,227 @@
...
+  if (inputs.length === 0) return undefined
+  const ordered = inputs
+    .map((input, index) => ({ input, index }))
+    .sort((left, right) => (left.input.order ?? 0) - (right.input.order ?? 0) || left.index - right.index)
+    .map(({ input }) => input)
+
+  try {
+    const entries = ordered.map((input) => ({ input, messages: normalizeInput(input.args) }))
+    if (entries.some((entry) => entry.messages.length === 0)) {
+      throw new TypeError('Continuation must append at least one message')
```
strandly-the-agent: 🟡 **Blast radius: one malformed contribution destroys the completed result and all sibling contributions.** This `TypeError` escapes through agent.ts:1212, outside the pass's try/finally. Observed in a probe: `AgentResultEvent` count 0, the *valid* sibling rejected (never consumed), and `invoke()` rejects — a user's successful turn is discarded because one internal producer passed bad args. Same shape for args that normalize to `[]` (interrupt-response/checkpoint-like), and a `toolResult`-shaped contribution isn't caught here at all — it becomes an orphan toolResult after `endTurn` and fails at the provider instead (content is never shape-checked).

Since producers are plugins that don't know about each other, would it be better to reject the offending input in isolation (settling the siblings normally) and let the completed result stand?

strands-ts/src/agent/agent.ts
```diff
@@ -1129,59 +1152,110 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+        if (limitStopReason) {
+          await continuations.rejectInputs(
+            pendingContinuations,
+            new Error(`Continuation rejected by ${limitStopReason}`)
+          )
+          pendingContinuations = []
+          if (afterInvocationEvent.resume === undefined) {
+            completedResult = this._withStopReason(completedResult, limitStopReason)
```
strandly-the-agent: 🟡 **An invocation that genuinely finished (`endTurn`) is re-reported as `limitTurns` because a rejected contribution existed.** Sharpest in the common config `limits: { turns: 1 }`: the model answers in one cycle, some plugin contributed → the caller sees `limitTurns` for a completed answer, and retry-on-limit wrappers retry a finished turn. Output-phase `AgentStreamStage` middleware already returned with the *un-rewritten* `endTurn` result, so middleware and caller now disagree about the same invocation. The contributor already learns via `onRejected('Continuation rejected by limitTurns')` — would it be simpler to leave `completedResult` untouched and keep the signal on the rejection side?

strands-ts/src/agent/agent.ts
```diff
@@ -1129,59 +1152,110 @@ export class Agent implements LocalAgent, InvokableAgent {
...
-        return result!
+        const selectedContinuations = pendingContinuations
+        pendingContinuations = []
+        const hasInternalContinuation = selectedContinuations.length > 0
+        continuation = await continuations.prepareInvocation(
+          selectedContinuations,
+          afterInvocationEvent.resume,
+          (continuationArgs) => this._normalizeInput(continuationArgs)
+        )
+        if (!hasInternalContinuation) {
```
strandly-the-agent: 🟡 **A mixed pass (contributions *and* `resume`) silently gets internal semantics on both axes.** `hasInternalContinuation` ignores `resume`, so a pass carrying both takes drawn-down limits here and — via `continuation.internal !== undefined` at :1109 — also skips the abort-controller reset, inheriting the previous pass's possibly-aborted signal. Both contradict the PR's “public `resume` keeps fresh signal/limits”, which the new goal test only pins for the resume-only shape (`goal/__tests__/plugin.test.ts:103-126`). Concretely: GoalLoop's next iteration runs on leftover budget (dies early) or is dead-on-arrival if the user cancelled during the After hooks — purely because an unrelated plugin also contributed.

Suggestion: key both decisions on `resume === undefined` instead of `hasInternalContinuation`. (Whichever way: an alternating internal→resume chain resets to `resolvedOptions` on each resume pass — `limits: { turns: 2 }` produced 5 model calls in a probe. Consistent with base `resume` semantics, but surprising next to the shared internal budget; worth a deliberate line in the docs.)

strands-ts/src/agent/agent.ts
```diff
@@ -2111,6 +2207,8 @@ export class Agent implements LocalAgent, InvokableAgent {
     // cannot affect modelState at any point (before or after next()).
     const modelStateSnapshot = this.modelState.getAll()
     let tempModelState: StateStore | undefined
+    let modelRequestMessages = context.messages
+    let continuationEvents: readonly MessageAddedEvent[] = []
```
strandly-the-agent: 🟡 **Retry middleware — an explicitly documented pattern — drops the continuation's `MessageAddedEvent`s.** `continuationEvents` is *assigned* by the terminal callback; on the second `next()` (middleware/README.md:9,44: “a retry middleware that calls `next()` three times produces one hook pair”) `publish` no-ops on `settled` and returns `[]`, overwriting attempt 1's events — and `_invokeModel`'s own publish at :2110 is a no-op by then too. Net: `integrateMessages` has already spliced the messages into `agent.messages`, but no `MessageAddedEvent` is ever emitted for them — SessionManager, conversation managers, memory extraction and the printer never see messages that are in history, exactly the pairing `_appendMessage` (:2555) exists to guarantee.

Fix: accumulate — `continuationEvents = [...continuationEvents, ...(await …publish(...))]` — or hoist the publish out of the terminal callback.

strands-ts/src/hooks/events.ts
```diff
@@ -172,6 +172,7 @@ export class AfterInvocationEvent extends HookableEvent {
   readonly type = 'afterInvocationEvent' as const
   readonly agent: LocalAgent
   readonly invocationState: InvocationState
+  readonly #continuations: ContinuationInput[] = []
```
strandly-the-agent: 🟡 **This `#private` field flips `AfterInvocationEvent` to nominal typing in the shipped `.d.ts` — a type-level break the PR body says doesn't exist — and the channel doesn't have to live on the class at all.** Verified by building declarations: `#private;` lands in `dist/src/hooks/events.d.ts`, so any external structural stand-in (`const fake: AfterInvocationEvent = {…}` — the obvious way to unit-test a hook without booting an `Agent`) or `implements` stops compiling. Nothing in-repo breaks (every site uses `new`), which is exactly how this ships unnoticed and surfaces as a user upgrade failure. `_continueWith` / `_takeContinuations` / `ContinuationInput` are likewise visible on the public class in `.d.ts`/autocomplete, three lines from the documented `resume` slot — with no first-party caller yet (`_continueWith` appears only in tests).

A `WeakMap` in `agent/continuation.ts` keyed on the event instance gives identical mechanics with a declaration output byte-identical to main (in-repo precedent for the pattern: `goal/plugin.ts:225`; it also ports cleanly to Python, whose `AfterInvocationEvent._can_write` guard would reject new fields):

```ts
const pending = new WeakMap<AfterInvocationEvent, ContinuationInput[]>()

/** @internal */
export function continueWith(event: AfterInvocationEvent, input: ContinuationInput): void {
  const queue = pending.get(event) ?? []
  queue.push(Object.freeze({ ...input }))
  pending.set(event, queue)
}

/** @internal */
export function takeContinuations(event: AfterInvocationEvent): readonly ContinuationInput[] {
  return pending.get(event)?.splice(0) ?? []
}
```

That keeps `team/DECISIONS.md`'s “hooks as low-level primitives” intact and moots the `api/needs-review` question flagged in the summary.

strands-ts/src/hooks/events.ts
```diff
@@ -204,6 +215,22 @@ export class AfterInvocationEvent extends HookableEvent {
...
+ *
+ * @internal
+ */
+export interface ContinuationInput {
+  /** Input to include in the next agent pass. */
+  readonly args: InvokeArgs
+  /** Relative ordering. Lower values are included first; equal values preserve submission order. */
+  readonly order?: number
+  /** Runs after the contribution reaches the model invocation. */
+  readonly onConsumed?: (modelRequestMessages: readonly Message[]) => void | Promise<void>
```
strandly-the-agent: 🟡 **The settlement contract should say what it actually guarantees before Background Tasks builds on it.** Three specifics:

1. *“Reaches the model invocation” is stronger than the code.* An `InvokeModelStage` middleware that short-circuits with a cached result still fires `onConsumed` (`modelRequestMessages` defaults to the pre-middleware list at agent.ts:2210; the test at agent.continuation.test.ts:211-241 locks this in). Reasonable semantics — but then this doc should promise delivery-into-the-request, not a model call, or a delivery-accounting consumer will mark work done that no model saw.
2. *Error policy is asymmetric and undocumented.* A throwing `onRejected` is logged and swallowed (continuation.ts:156-158); a throwing `onConsumed` propagates — from inside `_invokeModel`'s try, *after* the provider succeeded — so it surfaces as `AfterModelCallEvent.error`, gets attributed to the model, and can trigger a real (billed) re-invocation via `errorEvent.retry`, with attempt 1's usage never metered. Log-and-continue on both would treat these as the notifications they are.
3. The actual invariant — *exactly one of `onConsumed`/`onRejected` runs, exactly once* — held under every hostile probe I ran, but it's written nowhere. It's the whole point of the contract; worth stating here.

strands-ts/src/agent/__tests__/agent.continuation.test.ts
```diff
@@ -0,0 +1,429 @@
...
+      streamStageCalls += 1
+      return streamStageCalls === 2 ? { ...context, args: 'rewritten public' } : context
+    })
+    agent.addHook(BeforeModelCallEvent, (event) => {
+      projectedInputTokens.push(event.projectedInputTokens ?? 0)
+    })
+
+    await agent.invoke('start')
+
+    expect(requests[2]?.map(textOf)).toEqual([
```
strandly-the-agent: 🟡 **This asserts `requests[2]` — the follow-up pass's *second* call, after settlement, when `buildRequest` just returns plain history — so the headline invariant (internal before public in the request the provider sees) is unpinned.** Mutation evidence: swapping internal/public in `buildRequest` (continuation.ts:85) survives the whole suite; swapping in `integrateMessages` (history side) is caught. `requests[1]` is where the contributions actually flow; one added line pins it (value confirmed at this head):

```ts
expect(requests[1]?.map(textOf)).toEqual(['start', 'initial', 'first', 'late-a', 'late-b', 'rewritten public'])
```

(keep the `requests[2]` assertion — they pin different code paths.)

Two cheap companions while in here: **(a)** no test asserts any rejection *reason* — the code builds five distinct reasons (continuation.ts:109,130,167; agent.ts:1130,1195) and all five could collapse into one generic `Error` with the suite green; upgrading the `onRejected` count checks to `toHaveBeenCalledWith(expect.objectContaining({ message: … }))` makes them real. **(b)** the `BeforeInvocationEvent.cancel` short-circuit's two new `reject*` calls (agent.ts:1128-1131, :1140-1143) have zero coverage — a mutant deleting either passes 54/54.

strandly-the-agent (changes requested):
**TL;DR — requesting changes (draft-calibrated): one verified data-corruption bug in the deferred-publish path (🔴, reproduces via public `resume` + `BeforeModelCallEvent.cancel` alone), plus contract gaps worth closing before the first real contributor lands on this mechanism. Direction looks right.**

**Previous findings (03:08 bot review) — status at 89d7c80:** ✅ `_withStopReason` field-drop → fixed (spread rebuild) · ✅ inline stop-reason denylist → fixed (typed `CONTINUATION_REJECTION_STOP_REASONS`, `checkpoint` added + test) · ✅ nested ternary → fixed (if/else). The `PreparedContinuation` flattening in the same delta is a clean simplification.

**Verified at this head:** ✅ src type-check clean · ✅ 54/54 tests in the two changed test files (`vitest --project unit-node`); 1163 tests / 43 files in the touched areas pass · ✅ the 🔴 repro fails on this branch and passes on `origin/main` (worktree control) · ✅ settlement exactly-once held under hostile probes (throwing callbacks, hook throws, stream break, Before-cancel) · ✅ limits draw-down cannot underflow (swept — the gate trips first, contra my own context notes) · ✅ continuation × stateful-model and × sliding-window probes held · 🧪 11 targeted mutants: 4 survived (see inline test comment + appendix).

Findings are inline: 1×🔴, 10×🟡.

**Questions (non-blocking):**
1. End-state of `resume` vs contributions: is `resume` eventually sugar for a single contribution, or do both survive with different limits/cancellation semantics forever? A paragraph in the PR body (or a `team/designs/` doc — none exists for continuations/Background Tasks) would anchor review of the stacked work.
2. `ContinuationInput` shape: drop `order` for v1 (FIFO is already the tie-break; an open numeric priority space invites magic-number wars) and instead add `signal?: AbortSignal` so an async producer can withdraw an input that went stale before the follow-up pass? The latter looks load-bearing for Background Tasks.
3. Internal continuation messages are persisted/extracted indistinguishably from user input (session snapshots replay them as user speech; memory extraction will record a task nudge as a user fact — `model-extractor.ts:56-61`). Intentional, or should the message/`MessageAddedEvent` carry provenance?
4. “Related to #3632” doesn't resolve via the API — could you link the actual Background Tasks reference?

**Process:** as written this PR changes the shipped `.d.ts` of a root-exported class (see the `events.ts:175` comment), so per `team/API_BAR_RAISING.md` it needs **`api/needs-review`** — unless the WeakMap variant lands, which makes the declaration output byte-identical to main and keeps the “no public API” claim true.

<details><summary>How this review was produced</summary>

Full pipeline — first `strandly-the-agent` review of this PR (the 03:08 assessment was the CI bot on a previous head): triage → context build → five independent specialist passes (correctness, adversarial, API/DevX, test-quality, LLM-context) → aggregation. Operational note: the first fan-out died on infra timeouts; all five passes were re-run time-boxed on the default model tier and completed. Aggregation was done by the orchestrator. All repro/probe scratch files were removed; the tree was left clean.
</details>

<details><summary>Appendix — non-blocking (10)</summary>

- ⚪ `throw undefined` from a hook/middleware still slips the `caughtError !== undefined` guard and resolves `invoke()` with `undefined` (agent.ts:1174→1181). Pre-existing — base swallowed *all* falsy throws, so the widening here is a real fix; a `caught` boolean would close the last hole.
- ⚪ Non-finite provider usage would make `_remainingOptions` produce `NaN` and trip `_validateLimits` as a raw TypeError (theoretical; no provider observed doing it).
- ⚪ `_invokeModel`'s error path lost its two invariant comments (“we guarantee the Before/After pair”, “let CancelledError propagate — no retry”); the evergreen-comment rule keeps invariants.
- ⚪ `continuation.ts`'s module functions have no TSDoc; three throw as part of their contract (`prepareInvocation`, `prepareInputs`, `notifyConsumed` → `@throws`), and `ModelInvocationOutcome` (agent.ts:407) lacks property docs.
- ⚪ The token estimate shown to the model is history-only and pre-middleware — under-counts memory injection / trailing blocks (pre-existing; now feeds compression gates + agentic context more visibly). Worth a doc comment.
- ⚪ The composed tail is exempt from in-pass context reduction; on overflow-retry the still-unsettled contributions are re-appended after the reduced history, so one oversized contribution makes every retry overflow identically (conversation-manager.ts:169 path).
- ⚪ LIFO rejection ordering (`[...inputs].reverse()`, continuation.ts:154) looks deliberate but is untested and uncommented (a mutant flipping it survives).
- ⚪ Test gaps beyond the inline comment: `it.each` over all six `CONTINUATION_REJECTION_STOP_REASONS` (three unexercised), the multi-error `AggregateError` branch of `notifyConsumed`, a `test-d.ts` pinning `ContinuationInput` isn't root-reachable, and continuation × stateful-model / × sliding-window integration tests (probes held; nothing pins them).
- ⚪ Cross-SDK parity: no tracked issue needed (no new hook event), but Python's `AfterInvocationEvent._can_write` guard means a port can't just add a field — one more argument for the off-class channel.
- ⚪ Stray double blank line at agent.ts:2153.
</details>

strands-ts/src/agent/continuation.ts
```diff
@@ -0,0 +1,411 @@
...
+    ...(await publish(
+      agent,
+      invocation,
+      undefined,
+      invocationState,
+      passRanAgentLoop ? result.lastMessage.trackingId : undefined
+    )),
+  ]
+  const existingIndex = agent.messages.findIndex((message) => message.trackingId === result.lastMessage.trackingId)
+  if (existingIndex < 0) {
```
strandly-the-agent: 🟡 **New in this revision: the public-`resume` + short-circuiting-middleware path now writes to history where `origin/main` didn't** — on a plain `event.resume` follow-up (no contributions anywhere) whose `AgentStreamStage` middleware returns a cached/fabricated result without calling `next()`, this push splices both the resume input *and* the middleware's fabricated `lastMessage` into `agent.messages`, each with a dispatched `MessageAddedEvent`:

- `origin/main`: `['start', 'initial']` — short-circuited pass leaves history untouched
- `55d9251`: `['start', 'initial', 'public', 'fabricated']`

Verified both sides with the same probe (adversarial pass; cache-style middleware, second pass short-circuits). Persisting the resume input is arguably an improvement over main silently swallowing it — but a middleware-fabricated assistant message now lands in the caller's conversation and session snapshots, on a path the PR body describes as unchanged public behaviour. Your test asserting `'cached'` in history suggests this is deliberate; if so it deserves (a) a line in the PR description, and (b) a test pinning the *resume-only* short-circuit case — or, if not deliberate, gating the push on `invocation.internal.length > 0` restores main's semantics for the public path.

strandly-the-agent (approved):
**TL;DR — approving at 55d9251: my 🔴 and 9 of the 10 🟡s are fixed and held up under a fresh adversarial pass on the new machinery. Two 🟡s remain open — role-alternation (replied on its existing thread) and one new finding on the public-resume short-circuit path (inline) — both worth settling before this leaves draft, neither blocking.**

This is a **fixes-delta follow-up review**: I re-verified my own 11 findings against the delta (89d7c80 → 55d9251) with my own repros and gates, plus one fresh independent adversarial pass attacking the fix machinery itself. All evidence below is mine, not the CI bot's ledger.

**Verified at 55d9251:** ✅ src type-check clean · ✅ 40 files / 1106 tests pass (agent, goal, hooks, middleware, session, conversation-manager suites) · ✅ all five of my prior repros re-run and now pass (🔴 dup repro, contribution+Before-cancel, settlement leak, real-interrupt gate, mixed-pass semantics) · ✅ `git diff origin/main -- strands-ts/src/hooks/events.ts` is **empty** — the WeakMap move makes the shipped `.d.ts` byte-identical to main · ✅ adversarial probes on the new machinery all HELD (batch dispatch exactly-once under double-`next()` retry; insertBefore under Before-cancel / stale-lastMessage / fabricated-id; add/take lifecycle incl. the new BeforeModelCallEvent surface; reject-and-continue fires exactly once; real interrupting tool returns `interrupts` intact).

**Prior findings — ledger (11):**
1. 🔴 deferred-publish history corruption → **fixed** (membership guard + insert-before; my exact repro passes, three adversarial variants held)
2. 🟡 role-alternation in `buildRequest` → **still open** — replied on the existing thread, not re-filed
3. 🟡 terminal stop reasons not gating contributions → **fixed** (`isRejectionStopReason`; verified with a *real* interrupting tool, `interrupts` payload intact, no follow-up launched)
4. 🟡 settlement leak on resume-normalize throw → **fixed** (try/catch + `rejectPrepared`; probe: exactly one `onRejected`)
5. 🟡 malformed-contribution blast radius → **fixed** (per-input try/catch + `isCompleteMessageInput`, incl. orphan-toolResult; siblings and result survive)
6. 🟡 `endTurn`→`limitTurns` rewrite → **fixed** (`_withStopReason` deleted; true result returned, contributors informed via `onRejected`)
7. 🟡 mixed-pass semantics → **fixed** (signal + limits keyed on `publicResume`; probe: fresh limits, contribution consumed)
8. 🟡 retry middleware dropping `MessageAddedEvent`s → **fixed** (batch dispatch in `_streamCore` before yield; exactly-once under retry, correct fallback for untracked events)
9. 🟡 `#private`/`.d.ts` break → **fixed** (channel moved to WeakMaps in `continuation.ts`; declaration output identical to main — this also **moots the `api/needs-review` label question** and makes the “no public API” claim true)
10. 🟡 `onConsumed` contract → **fixed** (TSDoc now states exactly-once; both callbacks are log-and-continue — `notifyConsumed` no longer throws into the model path; wording now “retained in the request that produced a successful model-stage result”)
11. 🟡 test assertions → **fixed** (`requests[1]` pinned, rejection reasons via `objectContaining`, terminal-reason `it.each`, blast-radius + Before-cancel coverage)

**New:** one 🟡 inline (`continuation.ts:254`) — the public-`resume` + short-circuiting-middleware path now writes to history where main didn't; looks deliberate, needs a decision + pin.

**Prior questions:** api-label → mooted (see #9). Still open but non-blocking: the `resume`-vs-contributions end-state paragraph/design doc, message provenance for session/memory consumers, and a resolvable link for the Background Tasks work (#3632 doesn't resolve via the API).

<details><summary>Process + appendix — non-blocking (3)</summary>

Shape: fixes-delta follow-up per my review SOP — self-run verification (delta read, gates, five ported repros) plus one fresh independent adversarial spawn attacking the new machinery (WeakMap channel, batch dispatch, insertBefore, reject-and-continue, stop-reason gate, blast radius). Its verdict: 9/10 probes HELD, one new finding (posted inline). Scratch probes were deleted; tree left clean.

- ⚪ A contribution `add()`ed after `take()` ran (late hook ordering) is silently dropped — no settlement callback. Not a regression (identical to the previous `_continueWith` take-point), and it doesn't leak into the next invocation (verified), but “added too late” is indistinguishable from “consumed” for the producer. A `logger.warn` in `add()` when the event was already taken (a `taken` WeakSet) would close it cheaply.
- ⚪ Code-shape note: the drain loop (`agent.ts:1389-1396`) dispatches `MessageAddedEvent`s without the `takeMessageEventBatch` check. No reachable double-dispatch path found (the publish yields aren't inside `finally`, so `.return()` never re-surfaces them) — flagging only so a future refactor doesn't make it reachable.
- ⚪ The new reject-and-continue asymmetry (bad public resume rejects the internal inputs but the resume still runs with fresh limits) is defensible — the resume is the caller's explicit instruction — but worth one line in the `ContinuationInput` TSDoc.
</details>

strands-ts/src/agent/continuation.ts
strandly-the-agent: **Status at 55d9251: still open** — the only one of my 11 findings the revision doesn't address. `buildRequest` still concatenates `[...history, ...internal messages, ...publicMessages]` with no coalescing (now at continuation.ts:194-198), and the new `isComposableUserInput` guard actually guarantees the composed tail is all `user`-role — so N contributions (+ resume) still produce N(+1) consecutive `user` messages. Your own updated test pins the shape: the request-roles assertion at agent.continuation.test.ts:113 expects `['user','assistant','user','user','user','user']`, which Bedrock/Anthropic reject at the API level.

Still the same two acceptable outcomes from my side: fold the composed tail into a single user message (the `message-injection.ts` `foldIntoLastUserMessage` approach), or document it as a deliberate provider-specific constraint that Background Tasks must handle. Not blocking the draft — I've approved the revision — but the first consumer with two simultaneous contributions hits this.

strands-ts/src/agent/agent.ts
```diff
@@ -1083,21 +1093,26 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+      let currentOptions = resolvedOptions
       let currentArgs: InvokeArgs = args
 
       while (true) {
-        // Fresh AbortController per iteration, composed with any external signal.
-        this._abortController = new AbortController()
-        this._abortSignal = resolvedOptions?.cancelSignal
-          ? AbortSignal.any([this._abortController.signal, resolvedOptions.cancelSignal])
-          : this._abortController.signal
+        // Internal follow-ups share cancellation; public resume starts a fresh pass.
```
mkmeral: why are they different?

strands-ts/src/agent/agent.ts
```diff
@@ -1110,14 +1125,27 @@ export class Agent implements LocalAgent, InvokableAgent {
         yield await this._invokeCallbacks(beforeInvocationEvent)
 
         if (beforeInvocationEvent.cancel) {
+          await continuations.rejectInvocation(
```
mkmeral: what happens on reject? why did we bring this feature?

strands-ts/src/agent/agent.ts
```diff
@@ -1110,14 +1125,27 @@ export class Agent implements LocalAgent, InvokableAgent {
...
           yield this._appendMessage(message, invocationState)
           const afterEvent = new AfterInvocationEvent({ agent: this, invocationState })
-          await this._invokeCallbacks(afterEvent)
+          try {
+            await this._invokeCallbacks(afterEvent)
+          } catch (error) {
+            await continuations.rejectInputs(continuations.take(afterEvent), error)
+            throw error
+          }
+          await continuations.rejectInputs(
```
mkmeral: why are we rejecting here, isn't it successful?

strands-ts/src/agent/continuation.ts
```diff
@@ -0,0 +1,546 @@
...
+  for (const event of events) dispatchedMessageEvents.add(event)
+}
+
+function takeDispatchedMessageEvent(event: MessageAddedEvent): boolean {
+  if (!dispatchedMessageEvents.has(event)) return false
+  dispatchedMessageEvents.delete(event)
+  return true
+}
+
+function isRejectionStopReason(stopReason: StopReason): boolean {
```
mkmeral: Does this method add any value?

strands-ts/src/agent/continuation.ts
```diff
@@ -0,0 +1,546 @@
...
+  pendingInputs.set(event, inputs)
+}
+
+function take(event: ContinuationEvent): readonly ContinuationInput[] {
+  const inputs = pendingInputs.get(event) ?? []
+  pendingInputs.delete(event)
+  return inputs
+}
+
+function markMessageEventsDispatched(events: readonly MessageAddedEvent[]): void {
```
mkmeral: Does this method add any value?

strands-ts/src/agent/continuation.ts
```diff
@@ -0,0 +1,546 @@
...
+  for (const invocation of invocations) {
+    if (!invocation) continue
+    const group = matchRequestGroups(invocation, request).get(requestMessage.trackingId)
+    if (!group) continue
+    for (const sourceId of matchedSourceIds(group, serializedContent(requestMessage), true)) sourceIds.add(sourceId)
+  }
+  return sourceIds.size > 0 ? sourceIds : undefined
+}
+
+async function publish(
```
mkmeral: can we get some docstrings. what is `continuations.publish`, it's not intuitive for me, what they are doing

strands-ts/src/agent/continuation.ts
```diff
@@ -0,0 +1,546 @@
...
+  rejectInputs,
+  rejectInvocation,
+  requestSourceIds,
+  settleResult,
+  take,
+  takeDispatchedMessageEvent,
+  updateArgs,
+}
+
+function add(event: ContinuationEvent, input: ContinuationInput): void {
```
mkmeral: where/how do we actually add this?

strands-ts/src/agent/continuation.ts
```diff
@@ -0,0 +1,546 @@
...
+  internal: PreparedContinuation
+  readonly publicResume: boolean
+  publicMessages: readonly Message[]
+  requestGroups: readonly RequestGroup[]
+  settled: boolean
+}
+
+type ContinuationEvent = AfterInvocationEvent | BeforeModelCallEvent
+
+const pendingInputs = new WeakMap<ContinuationEvent, ContinuationInput[]>()
```
mkmeral: are these global? would this work with multiple agents, each with their own continuations?

strands-ts/src/agent/agent.ts
```diff
@@ -1132,13 +1139,18 @@ export class Agent implements LocalAgent, InvokableAgent {
...
         try {
-          result = yield* this._streamWithMiddleware(currentArgs, resolvedOptions, invocationState)
+          result = yield* this._streamWithMiddleware(currentArgs, resolvedOptions, invocationState, continuationEvent)
         } catch (error) {
           caughtError = error as Error
         } finally {
           // AfterInvocationEvent always fires — even on error or consumer break. Outside middleware.
           // Invoke hooks (so .resume can be set) but don't yield in finally (yields in finally
           // suspend the generator on consumer break instead of completing cleanup).
+          await continuations.abandon(
```
mkmeral, line 1149: is this supposed to be here? we also get into finally block in success right, so abandon would fire either case?

strands-ts/src/agent/agent.ts
```diff
@@ -1152,9 +1164,14 @@ export class Agent implements LocalAgent, InvokableAgent {
...
-        // Resume only on a clean invocation — errors propagate above.
-        if (afterInvocationEvent.resume !== undefined) {
-          currentArgs = afterInvocationEvent.resume
+        const hasContinuation =
+          (await continuations.prepare(afterInvocationEvent, (continuationArgs) =>
+            this._normalizeInput(continuationArgs)
+          )) !== undefined
+        continuationEvent = hasContinuation ? afterInvocationEvent : undefined
+
+        if (hasContinuation || afterInvocationEvent.resume !== undefined) {
```
mkmeral, line 1181: can we combine this with resume? maybe allow to set Continuation events there too? 🤔  I don't like we have these two things doing the exact same thing, at worst the logic should be combined to continuations imo

mkmeral (commented):
tbh I kind of like continuations and i want to make it public 😅  how can we combine this with resume and just make this logic public. I see this as resume + stacking + callbacks. so just an extension of existing feature

strands-ts/src/agent/agent.ts
gautamsirdeshmukh: Yes, `finally` also runs on success, but by then `markAppended` has already removed successfully added inputs, so `abandon` does nothing. This specific call handles cases where the stream closes or the agent otherwise failes before those inputs are added.

strands-ts/src/agent/agent.ts
gautamsirdeshmukh: Discussed offline, keeping continuations internal for now (retaining `event.resume` usage) but will revisit for a proper devX review before making public. 

To this particular callout, this PR does not replace or wrap `resume`. event.resume` can only hold one input, so if two hooks set it, the second overwrites the first. `continuations` preserve all inputs and combine them for the next invocation. 

So in other words, the existing `resume` behavior remains unchanged; this change simply allows multiple hooks to contribute without overwriting one another.

strands-ts/src/agent/agent.ts
```diff
@@ -1152,9 +1164,28 @@ export class Agent implements LocalAgent, InvokableAgent {
...
-          currentArgs = afterInvocationEvent.resume
+        const stopReason = result!.stopReason
+        const continuationStopped = [
+          'cancelled',
+          'checkpoint',
+          'interrupt',
+          'limitOutputTokens',
+          'limitTotalTokens',
+          'limitTurns',
+        ].includes(stopReason)
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Updated to endTurn / stopReason allowlist

strands-ts/src/agent/agent.ts
```diff
@@ -1152,9 +1164,28 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+        if (continuationStopped) {
+          await continuations.abandon(afterInvocationEvent, new Error(`Continuation abandoned after ${stopReason}`))
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Added test

strandly-the-agent (approved):
**TL;DR — re-approving at 8d1ada5: the interrupt regression I reported this morning is fixed and verified; nothing else blocking from my side.**

Fixes-delta follow-up (self-run). Verified at this head with my own probes:

- ✅ **Real tool interrupt + pending contribution** now returns `stopReason: 'interrupt'` with the `interrupts` payload intact; the contribution is abandoned exactly once with `'Continuation abandoned after interrupt'`, no follow-up pass (was: thrown `TypeError`, result lost).
- ✅ **Interrupt-resume still works alongside an abandoned contribution** — an After hook answering via `event.resume` with `InterruptResponseContent` completes normally (`endTurn`, tool finishes).
- ✅ **`stopSequence` still allows continuation**; the allowlist (`endTurn`/`stopSequence`, agent.ts:1168) is the right shape — unknown/custom provider stop reasons now fail safe to terminal, pinned by the new `providerSpecificStop` test.
- ✅ Gates: src type-check clean; 27 files / 748 tests pass (agent, goal, hooks, middleware suites). Scoped diff vs current main is exactly the three PR files.

**“Anything else?”** — nothing blocking. Three non-blocking items already on record, for completeness: `mkmeral`'s make-it-public thread is the #3632-level API discussion my first review asked for — and the current WeakMap-internal shape is the right position for it, since wrapping it in a public API later is purely additive; internal continuation messages still carry no provenance marker for session/memory consumers; and `onAppended` means “in history”, not “the model saw it” (details in my failure-modes comment above) — worth remembering when Background Tasks does delivery accounting.

mkmeral (approved):

PR #3845, opened 2026-08-16
fix(python): suppress delegation end_turn placeholder in MessageAdded…

### LINKED ISSUE #3808: [BUG] Delegation end_turn placeholder corrupts MessageAddedEvent stream for subscribers
### Checks

### SDK Language

Both

### Strands Version

1.52.0

### Language Runtime Version

3.10.0

### Operating System

macOS 26.5.2

### Steps to Reproduce

When a delegated tool fires, the delegation middleware's handling of `MessageAddedEvent` for the real content diverges depending on whether a session manager is attached.

With a session manager attached, `MessageAddedEvent` never fires for the real delegated content. Only the placeholder reaches subscribers like `MemoryManager`. The event is suppressed to avoid creating duplicate records in `RepositorySessionManager`, which uses incremental appends and would persist both the placeholder and the real content if the event fired.

Without a session manager, `MessageAddedEvent` fires for both the placeholder and the real content, so subscribers see two assistant messages for what is a single logical turn, a phantom extra message.

Both scenarios share the same root cause: the event loop unconditionally fires `MessageAddedEvent` for the placeholder before the delegation middleware runs, and there's no way to prevent or retract that first event under the current end_turn design. Firing the real event as well duplicates the message and produces a phantom turn (no session manager) or corrupts session restore (session manager attached, hence the suppression); not firing it loses data for memory extraction and other subscribers.
```python
import asyncio
import tempfile

from strands import Agent
from strands.hooks import MessageAddedEvent
from strands.session import FileSessionManager


async def run(session_manager=None):
    sub = Agent(name="billing", system_prompt="Always respond with exactly: REAL_ANSWER_42", callback_handler=None)
    orch = Agent(
        name="orch",
        system_prompt=(
            "You must call the billing tool immediately for any balance question. "
            "Never ask the user for clarification. Never respond directly."
        ),
        tools=[sub.as_tool(delegate=True)],
        session_manager=session_manager,
        callback_handler=None,
    )

    texts = []
    orch.add_hook(
        lambda e: texts.extend(b["text"] for b in e.message.get("content", []) if "text" in b)
        if e.message.get("role") == "assistant"
        else None,
        MessageAddedEvent,
    )

    await orch.invoke_async("What is my balance using the billing tool?")
    return texts


async def main():
    session_manager = FileSessionManager(session_id="repro", storage_dir=tempfile.mkdtemp())

    print("With session manager:", await run(session_manager))
    print("Without session manager:", await run())


if __name__ == "__main__":
    asyncio.run(main())

# Output:
# With session manager: ['Turn ended early by hook after tool execution']
# Without session manager: ['Turn ended early by hook after tool execution', 'REAL_ANSWER_42']
```

### Expected Behavior

`MessageAddedEvent` fires exactly once per delegation turn, carrying the real delegated content (REAL_ANSWER_42), regardless of whether a session manager is attached.

### Actual Behavior

With session manager: `MessageAddedEvent` fires once, but with the internal placeholder text ("Turn ended early by hook after tool execution") instead of the real content. The real answer never reaches subscribers.

Without session manager: `MessageAddedEvent` fires twice: once with the placeholder, once with the real content for what is a single logical turn.

### Additional Context

This also affects the TS version of delegation. The `MessageAddedEvent` fires twice with or without the session manager.

### Possible Solution

`AfterToolsEvent.end_turn` would gain a sentinel value (e.g. SUPPRESS_MESSAGE) alongside its existing `bool | str` type, meaning "stop the loop but don't append or persist any message." The delegation plugin's `_on_after_tools` sets this sentinel instead of True. The event loop checks for it in the `end_turn` branch and skips the `_append_messages` call entirely, so no placeholder is ever written to `agent.messages` and no `MessageAddedEvent` fires for it.

The delegation middleware in `_handle_stream` then becomes the sole appender: since no placeholder exists, it always appends the real delegation message unconditionally and fires `MessageAddedEvent` once, on both the session-manager and no-session-manager paths. This removes the `_session_manager` special case and the `redact_latest_message` call entirely — there's nothing to redact since nothing wrong was ever written — fixing both the data-loss and phantom-turn variants at the root.

### Related Issues

_No response_

AnnasMazhar:
…Event stream

The event loop unconditionally appended the end_turn placeholder and fired MessageAddedEvent for it before the delegation middleware ran. With a session manager attached, the real delegated content never reached subscribers (only the placeholder did); without one, subscribers saw both the placeholder and the real message (a phantom duplicate assistant turn).

Introduce a SUPPRESS_MESSAGE sentinel on AfterToolsEvent.end_turn: the delegation plugin sets it instead of True, the event loop skips the placeholder append/persist (still halting via EventLoopStopEvent), and the delegation middleware becomes the sole appender, firing MessageAddedEvent exactly once with the real delegated content on both paths.

Fixes #3808

## Type of Change

Bug fix
New feature
Breaking change
Documentation update
Other (please describe):

## Testing

## Checklist

----

### DIFF
--- a/strands-py/src/strands/agent/_agent_delegation.py
+++ b/strands-py/src/strands/agent/_agent_delegation.py
@@ -6,11 +6,6 @@ When a tool is configured with ``delegate=True``, this plugin ensures:
 2. The agent loop exits immediately after a successful delegation (via end_turn)
 3. The AgentResult is transformed with the delegation tool's content as the last message
 4. Streaming events from the delegate agent are surfaced natively in the parent stream
-
-The delegation plugin sets ``AfterToolsEvent.end_turn`` to ``SUPPRESS_MESSAGE`` so the
-event loop stops without appending or persisting a placeholder. The final delegation
-message is produced in middleware after the core loop exits, and ``MessageAddedEvent``
-fires exactly once with the real delegated content, regardless of session manager presence.
 """
 
 from __future__ import annotations
@@ -355,9 +350,6 @@ class AgentDelegation(Plugin):
         delegation_message: MessageType = {"role": "assistant", "content": delegation_content}
         _ensure_tracking_id(delegation_message)
 
-        # No placeholder was ever appended (end_turn was set to SUPPRESS_MESSAGE),
-        # so append the real delegation message and fire MessageAddedEvent exactly
-        # once, regardless of session manager presence.
         messages.append(delegation_message)
         await context.agent.hooks.invoke_callbacks_async(
             MessageAddedEvent(agent=context.agent, message=delegation_message)
--- a/strands-py/src/strands/agent/_agent_delegation.py
+++ b/strands-py/src/strands/agent/_agent_delegation.py
@@ -7,10 +7,10 @@ When a tool is configured with ``delegate=True``, this plugin ensures:
 3. The AgentResult is transformed with the delegation tool's content as the last message
 4. Streaming events from the delegate agent are surfaced natively in the parent stream
 
-The final delegation message is produced in middleware after the core loop exits.
-``MessageAddedEvent`` subscribers will see the end_turn placeholder followed by a second
-event with the real delegation content (no session manager), or no event for the real
-content at all (session manager attached).
+The delegation plugin sets ``AfterToolsEvent.end_turn`` to ``SUPPRESS_MESSAGE`` so the
+event loop stops without appending or persisting a placeholder. The final delegation
+message is produced in middleware after the core loop exits, and ``MessageAddedEvent``
+fires exactly once with the real delegated content, regardless of session manager presence.
 """
 
 from __future__ import annotations
@@ -25,6 +25,7 @@ from typing import TYPE_CHECKING
 from .._middleware.stages import AgentStreamContext, AgentStreamStage, ExecuteToolContext, ExecuteToolStage
 from .._middleware.types import MiddlewareNext
 from ..hooks import (
+    SUPPRESS_MESSAGE,
     AfterToolCallEvent,
     AfterToolsEvent,
     BeforeModelCallEvent,
@@ -240,7 +241,7 @@ class AgentDelegation(Plugin):
             )
             return
 
-        event.end_turn = True
+        event.end_turn = SUPPRESS_MESSAGE
         state.end_turn_via_delegation = True
 
     # --- Middleware ---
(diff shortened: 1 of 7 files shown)


strands-py/src/strands/agent/_agent_delegation.py
```diff
@@ -7,10 +7,10 @@
 3. The AgentResult is transformed with the delegation tool's content as the last message
 4. Streaming events from the delegate agent are surfaced natively in the parent stream
 
-The final delegation message is produced in middleware after the core loop exits.
-``MessageAddedEvent`` subscribers will see the end_turn placeholder followed by a second
-event with the real delegation content (no session manager), or no event for the real
-content at all (session manager attached).
+The delegation plugin sets ``AfterToolsEvent.end_turn`` to ``SUPPRESS_MESSAGE`` so the
```
liramon2: You can remove this point entirely from the docstring because the edge case it describes is gone.

strands-py/src/strands/agent/_agent_delegation.py
```diff
@@ -354,21 +355,13 @@ async def _handle_stream(
...
-            else:
-                await context.agent.hooks.invoke_callbacks_async(
-                    MessageAddedEvent(agent=context.agent, message=delegation_message)
-                )
-        else:
-            messages.append(delegation_message)
-            await context.agent.hooks.invoke_callbacks_async(
-                MessageAddedEvent(agent=context.agent, message=delegation_message)
-            )
+        # No placeholder was ever appended (end_turn was set to SUPPRESS_MESSAGE),
```
liramon2: The comment describes an edge case that no longer exists, so you can remove the comment.

strands-py/src/strands/event_loop/event_loop.py
```diff
@@ -913,7 +919,10 @@ async def _handle_tool_execution(
             else "Turn ended early by hook after tool execution"
         )
         end_turn_message: Message = {"role": "assistant", "content": [{"text": end_turn_text}]}
-        await agent._append_messages(end_turn_message)
+        # SUPPRESS_MESSAGE: halt without appending or persisting any message.
```
liramon2: This can be inferred from the code, so remove the comment.

strands-py/src/strands/hooks/events.py
```diff
@@ -169,6 +169,45 @@ def _interrupt_id(self, name: str) -> str:
         return f"v1:before_tools:{uuid.uuid5(uuid.NAMESPACE_OID, name)}"
 
 
+class SuppressMessage:
+    """Sentinel type for :attr:`AfterToolsEvent.end_turn`.
+
+    Setting ``end_turn`` to :data:`SUPPRESS_MESSAGE` halts the loop without
+    appending or persisting any assistant message, leaving production of the
+    turn's final message entirely to the component that set the sentinel
+    (e.g. the delegation plugin).
```
liramon2: No need to mention delegation plugin, since doing so adds coupling to this component.

strands-py/src/strands/hooks/events.py
```diff
@@ -169,6 +169,45 @@ def _interrupt_id(self, name: str) -> str:
         return f"v1:before_tools:{uuid.uuid5(uuid.NAMESPACE_OID, name)}"
 
 
+class SuppressMessage:
```
liramon2: This belongs above all the events. This can also be simplified with something like:
```
class _SuppressMessage:
    """Halts the loop without appending or persisting any message."""
    __slots__ = ()
    def __repr__(self) -> str: return "SUPPRESS_MESSAGE"
    def __bool__(self) -> bool: return True

SUPPRESS_MESSAGE = _SuppressMessage()
```

liramon2 (changes requested):

